Issue #2079, opened 2026-05-13
bug: bd close does not persist through jsonl auto-import


mike-matchpoint:
> Related but distinct: #1210 covers cache eviction of closed beads (`c.beads` map). This issue is about the prior step — close writes that never reach `.beads/issues.jsonl` on disk in the first place.

> Origin: filed locally as `vgc-5wqq` in a downstream city. Mirroring upstream for triage.

---

## Observed
Closing beads in rig Matchpoint-Vehicle-Graph (.beads/) succeeds and bd reports "✓ Closed <id>", but a subsequent `gc bd show <id>` returns the bead with status OPEN. Status close does not survive across bd invocations.

Concrete sequence:
1. `gc bd close vg-cet -r duplicate` -> "✓ Closed vg-cet" (success)
2. `gc bd show vg-cet` -> "[● P2 · OPEN]" (NOT closed)

Repeated close attempts produce the same flap. Some beads happen to stick closed (vg-cet.1, vg-tkm, vg-tt3) while others (vg-cet, vg-2p8, vg-cc1, vg-ek2, vg-hrn) revert on every read.

## Diagnostics
- `gc doctor` flags `rig:Matchpoint-Vehicle-Graph:bd-split-store` warning: both .beads/dolt and .beads/embeddeddolt exist; no active local store identified.
- .beads/config.yaml: dolt.auto-start: false, gc.endpoint_origin: inherited_city, gc.endpoint_status: verified.
- Each bd invocation logs "auto-importing 220929 bytes from .../issues.jsonl into empty database... auto-imported 123 issues" — the dolt-backed view appears empty per-invocation and re-imports from JSONL.
- Adding a LABEL persists across reads (jsonl size grows from 220929 -> 221575 bytes; label visible on next show). So writes do propagate back to JSONL when they go through the update path — but the close path does NOT update JSONL.

## Hypothesis
`gc bd close` is writing the status change to the in-memory Dolt only; the JSONL export path that handles --add-label is not invoked on close. Next auto-import resurrects the open status.

## Workaround
Mayor labeled obsolete duplicates with `obsolete:duplicate-of-vg-imu` so dispatchers/cartographers can filter them out. This is fragile and depends on every consumer checking the label.

## Required fix
Either:
1. Make `gc bd close` write through to the JSONL the same way `update --add-label` does.
2. OR reconcile the split-store per gc doctor's hint and remove the JSONL auto-import path so writes don't get clobbered.

## Impact
[...]

PR #4898, opened 2026-08-01
fix(gc): stop bd's dolt-unreachable message from suggesting a conflicting bd dolt start

### LINKED ISSUE #1374: bd error message suggests 'bd dolt start' which conflicts with gc-managed dolt
## Summary

When dolt is unreachable in a gc-managed city, bd suggests running `bd dolt start`, but this conflicts with gc's dolt management and makes things worse.

## Error message

```
Dolt server unreachable at 127.0.0.1:0: dial tcp 127.0.0.1:0: connect: can't assign requested address

Dolt server auto-start is disabled (dolt.auto-start: false).
Start the server manually:
  bd dolt start
```

## Problem

- gc manages dolt on fixed port **16444**
- `bd dolt start` starts dolt on a **random port**
- Both try to lock the same data directory
- User follows the suggested command, makes things worse

## Suggested fix

When `gc.endpoint_origin: managed_city` is set in beads config, the error message should say:

```
Dolt server unreachable. This is a gc-managed city.
Run 'gc stop && gc start' to restart the city and its dolt server.
Do NOT run 'bd dolt start' manually.
```

## Workaround

```bash
gc stop
pkill -f dolt
gc start
```

rjgeng:
## Summary
- Fixes #1374: when gc's managed Dolt server is unreachable, bd (always run by gc with `dolt.auto-start: false`) tells the operator to run `bd dolt start` to recover — but that starts a second, unmanaged Dolt server that fights gc's own managed server for the same data directory, making the outage worse.
- `gc bd` already scans bd's stderr for one known marker pattern (the silent-fallback banner, from the #2079/#2080 fix). This adds a second marker pair for bd's auto-start-disabled suggestion; on a non-zero bd exit, `gc bd` now appends a corrective hint pointing at the actual remedy (`gc start` / `gc dolt restart`) alongside bd's original output, without touching bd's own exit code.

## Test plan
- [x] TDD red→green confirmed
- [x] Targeted `go test ./cmd/gc/... -run "Bd|Dolt"` green
- [x] New table-driven detector test + two integration tests via fake-bd-script harness (hint fires on the real bd banner, hint does NOT fire on an unrelated bd error)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,6 +84,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `--json` envelope gained an `unregistered` field reporting whether this
   stop also removed the supervisor registration (gastownhall/gascity#4366).
 
+- **`gc bd` no longer lets bd's own error message steer operators into a
+  Dolt-server conflict.** When the managed Dolt server is unreachable, bd
+  (with `dolt.auto-start: false`, which gc always sets) tells the operator
+  to run `bd dolt start` — but that starts a second, unmanaged Dolt server
+  that fights gc's own managed server for the same data directory. `gc bd`
+  now detects that suggestion in bd's stderr and appends a corrective hint
+  pointing at the actual remedy (`gc start` / `gc dolt restart`) alongside
+  bd's original output, without altering bd's own exit code
+  (gastownhall/gascity#1374).
+
 - **ACP activity is now available across process boundaries.** ACP
   `session/update` timestamps are published through an atomic, coalesced
   sidecar, allowing a process other than the session owner to report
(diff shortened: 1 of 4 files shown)


quad341 (approved):
Hi @rjgeng — this is a genuinely nasty failure mode to have spotted: bd's own recovery advice actively making the outage worse by racing a second Dolt server onto the same data directory. Thank you for chasing it down.

The implementation is careful in the ways that matter. Requiring the marker *pair* rather than just `bd dolt start` is the right call, and I appreciated that you pinned the `bd dolt start --help` false-positive case in the table test. Reusing the bounded `headLimitedWriter` scan instead of adding a second capture, and leaving bd's stderr and exit code completely untouched, is exactly how this should have been threaded through `doBd`.

One edge I'll take on our side: gc sets `BEADS_DOLT_AUTO_START=0` for external and hosted Dolt endpoints too, so those can emit the same banner even though `gc start` / `gc dolt restart` can't recover a server we don't own. I'll gate the hint to gc-managed endpoints and add an external-endpoint regression test, then re-run the targeted `Bd|Dolt` tests before merging.

Your fix and your commit stay as the substance of this change. Nice catch.


---


Review coverage:
- completed: `qwen codex`
- unavailable: `claude`

Issue #4366, opened 2026-07-16
[Customer Zero #14] Cached pack dirties and gc stop unregisters unexpectedly


julianknutsen:
Submitted by: @wespd (Customer Zero feedback, observed 2026-07-09).

## Summary

One `gc stop` invocation exposed two separate lifecycle surprises:

1. a machine-managed cached pack checkout was reported as locally modified, making city configuration "invalid";
2. stopping a supervisor-managed city removed it from the supervisor registry, although command help describes stopping sessions rather than unregistering the city.

This numbered intake keeps both subreports together while they are reproduced and assigned.

## Observed behavior

GIVEN a user has not intentionally edited the cached `github` pack  
WHEN running `gc stop ~/meter-factory`  
THEN stop succeeds but warns:

```
gc stop: stopped city despite invalid config: city import 'github':
cached import .../gascity-packs/tree/main/github has local worktree changes;
run 'gc import install'
```

The command does not show which files changed or why a GC-managed immutable checkout became dirty.

The same command also unregisters the city from the machine-wide supervisor. A user reading "stop" as pause/shutdown reasonably expects a later start/resume against the existing registration; registration removal is not stated in the `gc stop` short or long help.

**Reported environment:** source-built 1.4-era binary on macOS with a populated `~/.gc` and supervisor-managed `meter-factory`.  
**Reproduction rate:** one Customer Zero incident; current-main cache reproduction is required.

### Cached import

A machine-managed, commit-pinned cache should remain clean during normal pack execution. Ignored runtime artifacts should not invalidate the city. If tracked or non-ignored drift really occurs, diagnostics should identify the changed paths and the repair operation should be safe and understandable.

### Stop lifecycle

The CLI must make the lifecycle contract explicit. Either:

- `gc stop` is documented clearly as stopping **and unregistering** a supervisor-managed city, including the subsequent start behavior; or
- stop/pause and unregister remain separate operations, with registration preserved by stop.

The issue does not prescribe which product semantics to choose, only that command behavior and help agree.

## Impact

The cache warning tells a user their city configuration is invalid even though they never edited it, without evidence they can inspect. Combining that with silent registration removal makes a routine shutdown feel destructive and complicates upgrade troubleshooting.

### Cache false-positive likely fixed, but exact incident is unclassified

[...]

PR #3314, opened 2026-06-10
feat(gc import): reference-aware cache prune + stop counting gitignored artifacts as dirty

bourgois:
## Summary

Two cache-hygiene fixes for the global pack cache (`~/.gc`):

1. **`gc import prune`** — a reference-aware command to clean the global pack cache. It enumerates cached repo clones, determines which are still referenced by installed packs, and prunes the rest. Dry-run by default (lists what *would* be removed); `--apply` to delete, `--keep-days N` to retain recently-touched clones. Cache clones accumulated unbounded (hundreds of stale repos) with no supported way to reclaim them.

2. **Don't count gitignored artifacts as cache-dirty** — the cache freshness check ran `git status --porcelain --ignored`, so gitignored build artifacts written into a cached pack clone (e.g. Python `.pyc`) counted as local modifications and blocked cache reuse with a spurious "cached import has local worktree changes". Dropping `--ignored` means only *tracked* changes mark the cache dirty.

## Testing

- [x] `go test ./internal/packman/ ./internal/config/ ./cmd/gc/` (prune reference-awareness, dry-run/apply, keep-days, cache-dirty gate) pass locally
- [x] `go build ./...`
- [ ] `make check-docs` for the regenerated `docs/reference/cli.md` — CI

## Checklist

- [x] Added tests for both behaviors
- [x] Updated `docs/reference/cli.md` for the new command
- [ ] No breaking changes

PR #4899, opened 2026-08-01
fix(gc): document and JSON-report gc stop's supervisor unregister behavior

rjgeng:
## Summary
- Part of #4366 (Customer Zero report). Only the confirmed-on-current-main half is addressed here: `gc stop` has always unregistered a supervisor-managed city from the registry as part of stopping it, but neither `--help` nor `--json` output said so. (The plain-text path already prints an explicit "Unregistered city '<name>' (<path>)" line, so the original report's "silent" framing didn't hold on current main — the real, confirmed gap was narrower: help text and the JSON envelope.)
- The other half of #4366 (cache false-positive) is out of scope here — the issue's own triage notes point to PR #3314 as the likely upstream fix already, but say the exact historical incident can't be distinguished from a current writer without the original reporter retesting.

## Changes
- `newStopCmd`'s `Long` help now states the unregister behavior and points at `gc register`/`gc unregister` for the split operations.
- `lifecycleActionJSON` gained an `Unregistered *bool` field (same pattern as the existing `Force`/`Wait`/`Async` fields); `writeCityStopSuccess` now threads through whether this stop call also unregistered, so `gc stop --json` can distinguish an unmanaged-city stop from one that also removed the supervisor registration.

## Test plan
- [x] TDD red→green confirmed (compile-red before the field/param existed)
- [x] Targeted stop/suspend/register/unregister/reload/restart test subset green
- [x] 3 new tests: JSON-true integration test reusing existing supervisor-managed-invalid-toml scaffolding, a unit test on `writeCityStopSuccess` table-driving true/false, a help-text assertion
- [x] `docs/reference/cli.md` regenerated automatically with the new help text

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   is invisible to it regardless of name, while vendor/testdata (tracked or
   not) stay excluded as before. (gascity#4479)
 
+- **`gc stop --help` and `gc stop --json` now state what actually happens to
+  a supervisor-managed city's registration.** `gc stop` has always
+  unregistered a supervisor-managed city as part of stopping it, but neither
+  the CLI help nor the `--json` output said so — a user reading "Stop all
+  agent sessions in the city" reasonably expected `gc start` to find the
+  city again later. Help text now documents the unregister behavior and
+  points at `gc register`/`gc unregister` for the split operations; the
+  `--json` envelope gained an `unregistered` field reporting whether this
+  stop also removed the supervisor registration (gastownhall/gascity#4366).
+
 - **ACP activity is now available across process boundaries.** ACP
   `session/update` timestamps are published through an atomic, coalesced
   sidecar, allowing a process other than the session owner to report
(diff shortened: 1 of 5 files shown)


quad341 (approved):
Thanks @rjgeng — the triage is the part I most appreciate here. Going back to #4366, re-checking the "silent" claim against current main, finding that the plain-text path *does* print `Unregistered city ...`, and then scoping this PR to the gap that genuinely still exists (help text plus the `--json` envelope) saved everyone downstream a wasted afternoon. Splitting off the cache half with a note on why it can't be settled without the original reporter was the right call too. `Unregistered *bool` following the existing `Force`/`Wait`/`Async` shape is exactly the pattern, and table-driving both `true` and `false` in the unit test is a nice touch.

Two things I'll pick up on my side so you don't have to round-trip. First, the `CHANGELOG.md` conflict — main moved under you, nothing to do with your change. Second, one spot where the flag gets dropped: the `stopLoadedCity` closure in `cmdStopJSONSequence` rebuilds the outcome without `unregistered`, so stopping a registered city while the supervisor *isn't* running reports `unregistered: false` after actually unregistering it. One-line fix plus a test for that branch. Landing with your commit and credit intact.


---

PR #4900, opened 2026-08-01
fix(gc): correct gc formula cook --attach help text (blocks, not parent-child)

### LINKED ISSUE #2392: gc formula cook --attach flag description claims parent-child, only creates blocks
# `gc formula cook --attach` flag description claims parent-child, only creates `blocks`

## Summary

The `--attach` flag on `gc formula cook` is documented as creating the sub-DAG "as children of the given bead" but in practice only creates a `blocks` dependency. The parent-child relationship the description implies is never established.

This matters because convoy auto-close (`doConvoyAutocloseWith` in `cmd/gc/cmd_convoy.go`) watches parent-child children, not `blocks` dependents. So an attached molecule can never trigger its parent's auto-close — even though the flag description implies it should.

## Environment

- `gc version`: 1.1.0
- macOS, Apple Silicon

## Flag description

From `gc formula cook --help`:

> `--attach string      attach sub-DAG to existing bead (bead gains blocking dep on sub-DAG root)`

And from the longer description:

> With `--attach=<bead-id>`, **the sub-DAG is created as children of the given bead.** The bead gains a blocking dependency on the sub-DAG root, so it won't close until the sub-DAG completes.

Two claims:
1. "Created as children of the given bead" (parent-child relationship)
2. "Bead gains a blocking dependency" (blocks relationship)

Only the second one actually happens.

## Reproduction

Cook a small formula with `--attach`, then inspect the relationships from three different angles.

### Setup

A two-step verdict formula (any formula with steps will do):

```toml
# formulas/verdict.toml
formula = "verdict"

[vars.candidate]
required = true

[[steps]]
id = "skeptic-verdict"
description = "Read {{candidate}}, write verdict, close."
metadata = { "gc.run_target" = "quidpro-social/skeptic" }

[[steps]]
id = "newshound-verdict"
description = "Read {{candidate}}, write verdict, close."
metadata = { "gc.run_target" = "quidpro-social/newshound" }
```

A convoy bead to attach to:

```
$ bd create "test convoy" --type convoy --silent
qs-c69
```

### Cook with --attach

```
$ gc formula cook verdict --var candidate=qs-test-candidate --attach qs-c69
Attached: qs-c69 -> qs-1om (root: qs-c69)
Root: qs-1om
Created: 3
```

### Inspect via bd show (JSON)

The verdict molecule root has no parent:

```
$ bd show qs-1om --json | jq '.[0] | {id, parent, status, dependents}'
{
  "id": "qs-1om",
  "parent": null,
  "status": "open",
  "dependents": [
    {
      "id": "qs-c69",
      "title": "test convoy",
      "status": "open",
      "priority": 2,
      "issue_type": "convoy",
      "dependency_type": "blocks"
    }
  ]
}
```

Only a `blocks` dependency back to the convoy. No `parent-child` to the convoy.

### Inspect via bd list --parent

```
$ bd list --parent qs-c69
Issue 'qs-c69' has no children
```

bd doesn't see the verdict molecule as a child of the convoy. Confirms no parent-child link exists.

### Inspect via bd dep list

```
$ bd dep list qs-c69
  qs-1om: verdict [P2] (open) via blocks
```

Only `blocks` dependencies listed. No mention of any parent-child relationship.

Three surfaces, three consistent answers: only `blocks` exists.

## Expected behaviour

One of:

- **The description is correct, the implementation is incomplete.** `--attach` should create *both* the `parent-child` and `blocks` relationships, matching the documented intent. The molecule root becomes a child of the attached bead AND blocks it from closing.

- **The description is wrong, the implementation is intentional.** Only `blocks` is created, and the description should drop the "created as children" wording.

## Actual behaviour

Only a `blocks` dependency is created. The `parent-child` link is not established. The description's "created as children" wording does not match what the code does.

## Why this matters

Convoy auto-close logic looks at `parent-child` children only:

> When a child bead closes, convoy tracking kicks in. `doConvoyAutocloseWith()` checks: 1. Does the closed bead have a ParentID? 2. Is the parent a convoy (not closed, not "owned")? 3. Are ALL sibling children now closed?
> — `engdocs/architecture/life-of-a-bead.md`

If `--attach` *only* creates a `blocks` dependency, then attaching N sub-DAGs to a convoy never triggers the convoy's auto-close — even after all attached sub-DAGs complete. The composability of late-bound DAG expansion (the explicit purpose of `--attach` per its description: *"the core primitive for late-bound DAG expansion — any agent, script, or workflow step can call it to expand a bead into a sub-workflow at runtime"*) with convoy auto-close is broken in practice.

I hit this trying to build a per-item fan-out pattern: cook one sub-workflow per upstream item, attach all to a parent convoy, expect the convoy to close when the last sub-workflow closes. The convoy stays open indefinitely.

## Related

- Issue #2391 — `gc sling` reparents formula step beads. Related symptom of the same underlying composability problem: late-bound expansion patterns don't compose cleanly with the documented auto-close mechanism.
- The convoy parity gap audit at `engdocs/archive/analysis/gastown-upstream-audit.md` (section "18f. Convoy parity gaps") notes that Gas City's convoys are missing reactive feeding and additional dependency types compared to Gas Town. This `--attach` behaviour may be part of the same broader gap.

## Suggestion

Whichever direction the maintainers prefer:

1. **Update the implementation** so `--attach` creates both `parent-child` (to the attached bead) and `blocks` (back from the attached bead to the sub-DAG root). This restores composability with `doConvoyAutocloseWith` and matches the description.

2. **Update the description** to drop "created as children of the given bead" and explain that only a `blocks` dependency is established. Users would then know they need a separate mechanism (or pattern) to get auto-close cascade.

Happy to test a fix or provide further reproduction detail.

rjgeng:
## Summary
- Fixes #2392: `gc formula cook --attach`'s help text claims the sub-DAG is "created as children of the given bead" (a parent-child relationship). Confirmed on current main this has never been true — both code paths (`ensureFormulaCookAttachDep` for v2 formulas, `molecule.Attach` for v1) only ever call `store.DepAdd(attachBeadID, rootID, "blocks")`, never setting `ParentID`.
- This isn't cosmetic: convoy auto-close (`doConvoyAutocloseWith`) watches parent-child children, not `blocks` dependents, so a user following the old description reasonably expects an attached sub-DAG's completion to trigger the attached convoy's auto-close — it never does. Real users hit this building fan-out patterns (cook N sub-workflows, attach to a parent convoy, expect auto-close on completion).
- Scope: the issue offered two directions — (1) change the implementation to also set `ParentID`, or (2) fix the docs to match reality. Went with (2) only: changing what `--attach` does to an established primitive risks side effects on other parent-child consumers not audited here, while documenting current behavior accurately is correct regardless of which direction is eventually chosen for the underlying composability gap.

## Changes
- `cmd/gc/cmd_formula.go`: `Long` help text now explicitly says it's a `blocks`-only relationship, not parent-child, and that convoy auto-close won't fire from it.
- `cmd/gc/cmd_formula_test.go`: new `TestFormulaCookAttachHelpDoesNotClaimParentChild`.
- `docs/reference/cli.md`: auto-regenerated with the new help text.

## Test plan
- [x] TDD red→green confirmed
- [x] Full `TestFormulaCook*`/`TestFormula*` suite green (no regressions)
- [x] docsync regenerated `cli.md` with the new text as expected, no other diffs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -62,6 +62,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   view, not an equivalent durable source; for durable history, query the event
   log for `store.maintenance.*` directly. (gascity#4418)
 
+- **`gc formula cook --attach`'s help text no longer claims a parent-child
+  relationship it never creates.** `--attach=<bead-id>` has only ever added
+  a `blocks` dependency from the attached bead to the sub-DAG root
+  (`ensureFormulaCookAttachDep` / `molecule.Attach` both call
+  `store.DepAdd(..., "blocks")`, never setting `ParentID`), but the long
+  help described it as creating the sub-DAG "as children of the given
+  bead." Since convoy auto-close watches parent-child children, not
+  `blocks` dependents, a user following the old description would wrongly
+  expect an attached sub-DAG's completion to trigger the attached convoy's
+  auto-close — it never does. Help text now describes the actual `blocks`
+  -only relationship and says so explicitly (gastownhall/gascity#2392).
+
 - **ACP activity is now available across process boundaries.** ACP
   `session/update` timestamps are published through an atomic, coalesced
   sidecar, allowing a process other than the session owner to report
--- a/cmd/gc/cmd_formula.go
+++ b/cmd/gc/cmd_formula.go
@@ -617,11 +617,15 @@ func newFormulaCookCmd(stdout, stderr io.Writer) *cobra.Command {
 This is a low-level workflow construction tool. It creates the formula root
 and all compiled step beads without routing any work.
 
-With --attach=<bead-id>, the sub-DAG is created as children of the given
-bead. The bead gains a blocking dependency on the sub-DAG root, so it won't
-close until the sub-DAG completes. This is the core primitive for late-bound
-DAG expansion — any agent, script, or workflow step can call it to expand a
-bead into a sub-workflow at runtime.
+With --attach=<bead-id>, the given bead gains a blocking dependency on the
+sub-DAG root, so it won't close until the sub-DAG completes. This is a
+"blocks" dependency only, not a parent-child relationship — the sub-DAG
+root does not become a child of the attached bead (gc bd list --parent
+will not find it), and convoy auto-close, which watches parent-child
+children rather than blocks dependents, is not triggered by the sub-DAG
+completing. This is the core primitive for late-bound DAG expansion — any
+agent, script, or workflow step can call it to expand a bead into a
+sub-workflow at runtime.
 
 With --attach on a v2 formula — one declaring
 [requires] formula_compiler = ">=2.0.0" — the invocation runs under a
--- a/cmd/gc/cmd_formula.go
+++ b/cmd/gc/cmd_formula.go
@@ -622,10 +622,10 @@ sub-DAG root, so it won't close until the sub-DAG completes. This is a
 "blocks" dependency only, not a parent-child relationship — the sub-DAG
 root does not become a child of the attached bead (gc bd list --parent
 will not find it), and convoy auto-close, which watches parent-child
-children rather than blocks dependents, is not triggered by the sub-DAG
-completing. This is the core primitive for late-bound DAG expansion — any
-agent, script, or workflow step can call it to expand a bead into a
-sub-workflow at runtime.
+children and "tracks" members rather than blocks dependents, is not
+triggered by the sub-DAG completing. This is the core primitive for
+late-bound DAG expansion — any agent, script, or workflow step can call it
+to expand a bead into a sub-workflow at runtime.
 
 With --attach on a v2 formula — one declaring
 [requires] formula_compiler = ">=2.0.0" — the invocation runs under a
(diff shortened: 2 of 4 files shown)


quad341 (approved):
@rjgeng — thank you for this one. The valuable part isn't the wording change, it's that you actually traced both attach paths (`ensureFormulaCookAttachDep` for v2, `molecule.Attach` for v1) and established that neither has ever set `ParentID`. I re-checked against `doConvoyAutocloseWith` on my side and you're exactly right: it walks parent-child and `tracks`, never `blocks`, so anyone who read "created as children of the given bead" and expected their attached convoy to auto-close was quietly building on something that could not work. That's a real bug in the docs, not a typo.

I also think you called the scope correctly. Changing `--attach` to set `ParentID` would ripple into consumers nobody has audited; describing what the code actually does is right no matter which way that eventually goes.

The branch has picked up conflicts in `CHANGELOG.md` and the generated `docs/reference/cli.md` — both high-churn files, nothing to do with your change. I'll take the rebase, re-run `gen-doc`, and re-run the `cmd/gc` suite on my side, then merge. Your commit and your wording stay as authored.

The regression test pinning the phrasing was a nice touch — it'll keep this from drifting back.


---


Review coverage:
- completed: `qwen codex`
- unavailable: `claude`

PR #4201, opened 2026-07-13
feat(api): one pagination vocabulary — dialect CI guard + unified page contract (P1 #4 S4)

julianknutsen:
Final slice of the keyset-cursor program (API audit P1 #4, bead ga-q1rees). Stacked on #4194 (S3) → #4192 (S2) → #4157 (S1); the diff below main includes the parents until they merge — S4-only content is the last commit.

## What

The audit that started this program found **five pagination dialects** accreted silently across the API. S1–S3 converged the five main lists (beads, convoys, mail, sessions, events) on keyset cursors; S4 makes the vocabulary **self-enforcing** and finishes the contract:

**1. Spec-walking dialect guard** (`TestPaginationDialectGuard`): every operation using pagination params must speak keyset (subset of `{cursor, limit}`) or match an **exact** grandfathered legacy dialect (agent output `before/tail`, city stream `after_seq`, supervisor stream `after_cursor`, extmsg `after_sequence`, orders/history `before`, session transcript `after/before/tail` — owner sign-off 2026-07-11). Cursor-speaking ops must declare a 400 and pin the unified limit schema. Novel names can't slip past — `paginationSuspect()` also trips on pagination-shaped names (`next`, `page_token`, `resume_*`, `*_after`…); the red-team upheld that a pure name blocklist would ship a sixth dialect silently. Stale grandfather entries fail too, so the list only shrinks honestly. Self-check tests prove the checker bites on every violation class.

**2. Unified page contract on `PaginationParam`**: `maximum:"1000"` (over-limit is now a typed 422, was a silent clamp — repo-wide consumer sweep found nothing sending >1000) and `default:"100"` (huma injects when omitted; explicit `limit=0` still means server default). Cursor doc states the invalid-cursor 400 contract.

**3. One server default**: `defaultPaginationLimit` 50→100 everywhere (was 50 on beads/convoys/mail, **1000 on sessions**, 100 on events). "List everything" consumers of `GET /sessions` now ask for the cap explicitly — gc CLI `ListSessions`, dashboard BFF session enrichment, SPA `listSessions` — so the sessions default shrink changes no observed behavior anywhere.

**4. Documented order**: `listOrder()` writes each keyset list's total order + cursor contract into its operation description (`created_at DESC, id DESC` on beads/convoys/mail/sessions; `seq DESC` on events).

## Red-team

[...]

Issue #2536, opened 2026-05-24
feat: gc beads close / delete subcommands — no first-class CLI path to close file-store beads


sjarmak:
## Summary

`gc beads` exposes read-only operations (`list`, `show`, `health`, `city`) but no `close` or `delete` subcommands for file-store provider cities. `gc bd close` errors out with `"only supported for bd-backed beads providers"`. Operators have no first-class path to close zombie beads in `scope_root/.gc/beads.json` and end up direct-editing the JSON (which races a live bookkeeping writer).

## Symptom

Trying to close a stale `gc-*` bead in a file-store city:

```sh
$ gc bd close gc-144816 --reason "empty bead cleanup"
gc bd: only supported for bd-backed beads providers (resolved "file" for /home/ds/gas-city)
  hint: GC_BEADS env var overrides the provider. Unset it, or set GC_BEADS=bd for this scope.

$ GC_BEADS=bd gc bd close gc-144816 --reason "empty bead cleanup"
Error: resolving ID gc-144816: no issue found matching "gc-144816"

$ bd close gc-144816 --reason "empty bead cleanup"   # from /home/ds/gas-city
Error: resolving ID gc-144816: no issue found matching "gc-144816"
```

The `bd-backed` path resolves to the dolt sql-server's `bd` database — different store from the file-store beads at `.gc/beads.json`. Neither dispatch reaches the file-store record.

## Reproduction

1. In a city with `[beads].provider = "file"` (file-store persistence at `<scope_root>/.gc/beads.json`), find any open bead with an ID like `gc-<NNNNN>`:

   ```sh
   python3 -c "import json; d=json.load(open('/home/ds/gas-city/.gc/beads.json')); print([b['id'] for b in d['beads'] if b.get('status')=='open'][:5])"
   ```

2. Try every documented close path:

   ```sh
   gc bd close <id>                    # errors: bd-backed only
   GC_BEADS=bd gc bd close <id>        # errors: not found in dolt
   bd close <id>                        # errors: not found in dolt
   gc beads close <id>                  # errors: unknown subcommand
   ```

   All fail.

3. Operator resorts to direct JSON edit. Hits race with the live bookkeeping writer:

   ```python
   import json, os, tempfile
   src = '/home/ds/gas-city/.gc/beads.json'
   d = json.load(open(src))
   for b in d['beads']:
       if b['id'] == 'gc-144816': b['status'] = 'closed'
[...]

PR #4901, opened 2026-08-01
feat(api): publish the bead delete endpoint's soft-delete contract in the spec

wbern:
## The juicy parts

`DELETE /v0/city/{cityName}/bead/{id}` is a soft delete — it closes the bead — and that was recorded only in an unexported handler's Go doc comment. The OpenAPI document and every generated client described the operation as `Delete v0 city by city name bead by ID`, with no description at all. So a consumer wiring a cleanup path onto this endpoint could watch it return `200` and leave every bead in place: closed, still readable, still returned by any query that includes closed beads. Nothing in the published contract contradicted the expectation the HTTP verb sets.

The audience is API consumers and tooling authors, not end users. This is spec/document only, with **no behaviour change** — the sole runtime effect is a string in the generated spec. Measured before the change, 14 of 166 operations carried a description and this was not one of them; the reusable `describes(...)` option added here appends rather than overwrites, so it composes with the existing `listOrder` contract.

## TL;DR

`DELETE /v0/city/{cityName}/bead/{id}` is a **soft delete** — it closes the bead. That was documented only in the handler's Go doc comment, so the OpenAPI document and every generated client described it as `Delete v0 city by city name bead by ID` with **no description at all**. This publishes the contract. Spec/document only; no behavior change.

## Why it matters

A caller who needs a bead actually removed can wire a cleanup path onto this endpoint, watch it return `200`, and leave every bead in place — closed, still readable, still returned by any query that includes closed beads. Nothing in the published contract contradicts the expectation the HTTP verb sets.

The semantics were never in doubt in the source:

```go
// humaHandleBeadDelete is the Huma-typed handler for DELETE /v0/bead/{id}.
// It is implemented as a soft-delete (store.Close) — see the `"closed"`
// status field for honest wire-contract semantics. Hard-delete is not
// exposed through the API.
```

But the handler is unexported, and the generated client and OpenAPI document are all most consumers ever see. A semantic recorded only in a handler comment is, in practice, undocumented.

Measured before this change: **14 of 166** operations carry a description, and the bead delete was not one of them.

## Behavior

```gherkin
Given a consumer reading the OpenAPI document or a generated client
When it looks up DELETE /v0/city/{cityName}/bead/{id}
Then the operation states that the bead is closed, remains readable, keeps its
     history, and that the API exposes no hard delete
```

```gherkin
Given any caller of DELETE /v0/city/{cityName}/bead/{id}
When the request is served
Then the response, status codes and stored state are exactly as before
```

```gherkin
Given an operation registered without describes(...)
When the spec is generated
Then its description is unchanged
```

## Why it's safe

- The only runtime effect is the string in the generated spec. Handler, routing, status codes and store behavior are untouched.
- `describes` appends rather than overwrites, matching `listOrder`, so composing both on one operation keeps each contract.
- Regenerated with `go run ./cmd/genspec` + `go generate ./internal/api/genclient`. The Go client did not drift — descriptions do not reach it — so the diff in generated artifacts is one line each in `internal/api/openapi.json`, `docs/reference/schema/openapi.json`, and `docs/reference/schema/openapi.txt`.
- `go vet ./internal/api/` clean; full `./internal/api/` suite green (68s), including the spec-in-sync gate.

## Related work

- #4201 — introduced `listOrder`, which documents list ordering on the operation for the same reason ("so consumers never have to reverse-engineer it from behavior"). `describes` is the general form of that idiom; `listOrder` is unchanged.
- #2536 — `gc beads close / delete` has no first-class CLI path, and turns on the same close-versus-delete distinction this endpoint blurs.

## Scope

| File | Change |
|---|---|
| `internal/api/city_scope.go` | add `describes(text)` operation option (+14) |
| `internal/api/supervisor_city_routes.go` | describe the bead delete op (+2/-1) |
| `internal/api/openapi.json`, `docs/reference/schema/openapi.{json,txt}` | regenerated (+1 each) |

### DIFF
--- a/docs/reference/schema/openapi.json
+++ b/docs/reference/schema/openapi.json
@@ -24524,6 +24524,7 @@
     },
     "/v0/city/{cityName}/bead/{id}": {
       "delete": {
+        "description": "Closes the bead. This is a soft delete: the bead remains readable and retains its history, with status \"closed\". The API exposes no hard delete, so a caller that needs the record actually removed must not rely on this endpoint.",
         "operationId": "delete-v0-city-by-city-name-bead-by-id",
         "parameters": [
           {
--- a/docs/reference/schema/openapi.txt
+++ b/docs/reference/schema/openapi.txt
@@ -24524,6 +24524,7 @@
     },
     "/v0/city/{cityName}/bead/{id}": {
       "delete": {
+        "description": "Closes the bead. This is a soft delete: the bead remains readable and retains its history, with status \"closed\". The API exposes no hard delete, so a caller that needs the record actually removed must not rely on this endpoint.",
         "operationId": "delete-v0-city-by-city-name-bead-by-id",
         "parameters": [
           {
--- a/internal/api/city_scope.go
+++ b/internal/api/city_scope.go
@@ -137,6 +137,21 @@ func errorStatuses(codes ...int) func(o *huma.Operation) {
 	}
 }
 
+// describes sets an operation's description, so a contract a caller cannot
+// infer from the method and path is published in the spec rather than left in
+// the handler's Go doc. Generated clients and the OpenAPI document are all most
+// consumers ever see; a semantic the handler comment alone records is, in
+// practice, undocumented.
+func describes(text string) func(o *huma.Operation) {
+	return func(o *huma.Operation) {
+		if o.Description == "" {
+			o.Description = text
+			return
+		}
+		o.Description += "\n\n" + text
+	}
+}
+
 // listOrder documents a list endpoint's total order and cursor contract in
 // its operation description. Every keyset list declares its order here so
 // consumers never have to reverse-engineer it from behavior (the pre-S4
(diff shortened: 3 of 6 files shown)


quad341 (approved):
@wbern — thank you for this one. The gap you found is easy to look past precisely because nothing was broken: the handler comment said "soft-delete (store.Close)" and the wire response literally returns `"closed"`, but the published spec said `Delete v0 city by city name bead by ID` and nothing more. A consumer wiring a cleanup path off the HTTP verb would have gotten a `200` and a full bead table, with nothing in the contract to contradict them.

I traced it before merging rather than taking the description on faith: `humaHandleBeadDelete` does call `store.Close` and stamp `"closed"`, and `cityDelete(sm, "/bead/{id}", ...)` is the only bead-delete registration in the route table, so there's no sibling route left undescribed. Your claims held up.

Two things I liked: generalizing `listOrder` into a reusable `describes` instead of hardcoding a second one-off, and making it append rather than overwrite so the two compose on the same operation. Merging as-is.

One small note for next time — the TypeScript SDK did pick the sentence up as JSDoc, so "one line each in three files" undercounts by a file. Changes nothing; the generated-artifacts gate verified it.


---

PR #4905, opened 2026-08-02
docs(dispatch): work-formula selection guide + v2 --on semantics; fix digest-generate mislisting

edwarddavison:
## Summary

Documentation-only update to the `core` pack's `gc-dispatch` skill, closing two
gaps surfaced by the gas-city formula-adoption audit (gci-7x4):

1. **Variant-selection guidance** — a new *"Choosing a work formula"* decision
   guide (isolation × handoff table) covering `mol-do-work` vs `mol-scoped-work`
   vs `mol-polecat-work`, plus the `mol-polecat-commit` / `mol-polecat-report`
   siblings. Also documents `mol-scoped-work`, which the skill did not list at
   all.
2. **v2 `--on` / `-f` semantics** — the dispatch sections described the v1
   model ("creates a molecule wisp on the bead"). Updated to the v2 contract:
   a formula declaring `formula_compiler = ">=2.0.0"` starts a **workflow**
   (workflow-root + auto-convoy + work bead) whose step DAG and handoff
   metadata persist as beads; v1 formulas still instantiate a wisp. Adds the
   rule that convoy-referencing formulas (those using `{{convoy_id}}` or a
   drain step) must launch onto a convoy — which `--on` provides.
3. **digest-generate mislisting** — `mol-digest-generate` was listed among the
   patrols "auto-poured by agent startup prompts." It is not: it is driven by
   the `digest-generate` order on a 24h cooldown. Removed from the patrol list
   and corrected inline.

## Why

Under bare-bead dispatch the handoff contract lives only in the agent's prompt;
a pool recycle between push and reassign strands the branch (happened twice in
gas-city). The v2-workflow durability point is now documented where dispatchers
choose a formula, so the isolation/handoff/recycle trade-off is explicit.

## Scope

Single file: `internal/bootstrap/packs/core/skills/gc-dispatch/SKILL.md`.
No code, formula, or behavior changes — the described semantics are the current
`gc sling` contract (verified against `gc sling --help` and the formula
`[requires]` blocks at this tree).

Source: gas-city bead gci-dqe (audit parent gci-7x4).


quad341 (approved):
@edwarddavison — this is a genuinely useful docs pass, and the isolation × handoff table is the part I'm happiest about: "does the agent get its own worktree" and "does it hand off to the refinery" are exactly the two axes people actually choose on, and nobody had written them down side by side. Documenting `mol-scoped-work` at all closes a real gap — it's been a built-in core formula with no mention in the dispatch skill. And the `mol-digest-generate` correction is right: it's a Gastown-pack order on a 24h cooldown, not a startup patrol pour, which `docs/guides/gastown-config-recipes.md` confirms independently.

One thing I'll fix on my side rather than bounce back to you. The parenthetical "pass `--no-convoy` to suppress it" doesn't hold for v2: `NormalizeInputConvoy` always supplies or creates the input convoy for a targeted invocation, and `--no-convoy` never reaches that path — it only gates the ordinary routing auto-convoy, which formula slings skip anyway. Easy sentence to land wrong; the flag name reads like it should apply. I'll reword it to draw that line, run the pack-embed and docs checks, and merge with your authorship intact.

Thanks for turning the gci-7x4 audit findings into something dispatchers can actually read.


---


Review coverage:
- completed: `qwen codex`
- unavailable: `claude`

PR #4906, opened 2026-08-02
fix(t3bridge): use provider family for Codex envelope model

bryceputman:
## Summary

T3Bridge selected its default startup-envelope model with an exact `provider == "codex"` check. Codex aliases such as `codex-mini`, and custom providers whose resolved builtin ancestor is Codex, therefore fell through to the Claude fallback and were stamped with `claude-opus-4-6`.

This change resolves the provider family through the existing canonical helpers before selecting the Codex default. Explicit `--model` and `GC_MODEL` values retain precedence, and the existing fallback for every non-Codex provider is unchanged.

## Testing

- Regression proof on current `main`: the new alias and Codex-derived custom-provider cases both fail with `claude-opus-4-6` before the fix.
- `go test ./cmd/gc -run '^(TestT3BridgeStartupEnvelopeModel_PrefersResolvedEnvModel|TestT3BridgeStartupEnvelopeModel_UsesCurrentProviderDefaults)$' -count=10`
- `make check`
- `LOCAL_TEST_JOBS=2 make test-cmd-gc-process-parallel`
- `LOCAL_TEST_JOBS=2 make test-integration-shards-parallel`: all 4 core, all 6 `cmd/gc`, all 3 runtime/tmux, `bdstore`, and several REST shards passed. The remaining shards failed during Beads/Dolt initialization with schema-migration errors or `bd` timeouts before exercising this change. A representative failing formula shard reproduces on untouched upstream `main` at the same Beads/Dolt setup boundary.

## Checklist

- [x] Added focused regression coverage for Codex aliases and Codex-derived custom providers.
- [x] Preserved explicit model precedence and all non-Codex fallback behavior.
- [x] No documentation change is needed; there is no new command or configuration surface.
- [x] No breaking change or migration.

### DIFF
--- a/cmd/gc/template_resolve_t3bridge.go
+++ b/cmd/gc/template_resolve_t3bridge.go
@@ -4,6 +4,8 @@ import (
 	"encoding/json"
 	"path/filepath"
 	"strings"
+
+	sessionpkg "github.com/gastownhall/gascity/internal/session"
 )
 
 func applyT3BridgeRuntimeConfig(tp TemplateParams, env map[string]string) {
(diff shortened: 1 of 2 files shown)


bryceputman:
Closing after re-evaluating the intended behavior and scope. No action needed.

PR #4910, opened 2026-08-02
feat(bd): refuse a `gc bd update` whose --set-metadata pairs bd would drop

wbern:
## The juicy parts

This is silent data loss with a success exit code. `bd update <id> --set-metadata a=1 b=2 c=3` stores **one** pair, prints its success line, and **exits 0** — `b=2` and `c=3` land in bd's variadic issue-id slot, fail to resolve on stderr, and are discarded. A caller cannot distinguish a full write from a 1-of-N write, so ordinary `|| exit 1` error handling is blind to it.

This refuses the shape before any store work: nothing is written, the exit code is honest, and the message names each dropped pair and the repeated-flag form that works. It cannot condemn an invocation that previously worked, because **no bead id contains `=`** — a `=`-bearing positional was already failing under bd, just quietly. The only change is a silent partial write becoming a loud refusal, issued before the write.

Scope is `gc bd`, which execs bd, so nothing between the caller and this behaviour sees it. The guard is keyed off the bd subcommand and is **not** disarmed by a global flag before the verb — `gc bd --actor bob update <id> --set-metadata a=1 b=2` is refused with a non-zero exit and writes nothing, verified against a real store. A raw `bd` invocation is still exposed; the exit-code contract itself is filed upstream as steveyegge/beads#5247.

## TL;DR

`bd update <id> --set-metadata a=1 b=2 c=3` writes **one** pair, reports success, and **exits 0**. `gc bd` execs bd, so nothing between the caller and that behaviour sees it. This refuses the shape before any store work — nothing written, honest exit code.

## The behaviour

`--set-metadata` is repeatable and takes ONE `key=value` per occurrence; `bd update` is variadic over issue ids. So only `a=1` is the flag's value — `b=2` and `c=3` become positional issue ids.

```console
$ bd update bd-abc --set-metadata probe_a=1 probe_b=2 probe_c=3
✓ Updated issue: bd-abc — scratch probe
Error resolving probe_b=2: no issue found matching "probe_b=2"
Error resolving probe_c=3: no issue found matching "probe_c=3"
$ echo $?
0
$ bd show bd-abc --json | jq -c '.[0].metadata'
{"probe_a": 1}
```

Three pairs in, one stored, exit 0. A caller cannot distinguish a full write from a 1-of-N write, so no `|| exit` guard can catch it. Reproduced on bd 1.1.0, unchanged in 1.1.2 (`cmd/bd/update.go` is identical between them). Reported upstream as steveyegge/beads#5247. The asymmetry that hides it: repeated `--unset-metadata` flags all apply — only the set path loses pairs.

## Why it can't break a working invocation

**No issue id contains `=`.** A `=`-bearing positional therefore never resolved under bd either — it was already failing, silently. The guard converts a silent partial write into a loud refusal, issued *before* the write rather than after.

## Behaviour

```gherkin
Given `gc bd update <id> --set-metadata a=1 b=2`
When doBd runs
Then it exits non-zero, names the dropped pair and the repeated-flag form,
     and performs no store work
```

```gherkin
Given `gc bd update <id> --set-metadata a=1 --set-metadata b=2`
When doBd runs
Then it proceeds unchanged
```

```gherkin
Given `gc bd update <id> --add-label role=worker`
When doBd runs
Then it proceeds unchanged — the value belongs to --add-label, not the id slot
```

## The part that needs care

Positional detection needs the **complete** value-flag table. With a subset, the value of any omitted flag is read as a positional id — and `gc bd update <id> --add-label <key>=<value>` is shipped verbatim in `internal/bootstrap/packs/core/skills/gc-work/SKILL.md:50`, so a partial table breaks a documented command. `internal/bdflags` already declares itself the single source of truth for bd's per-subcommand flag names, so the argv parsing lives there rather than beside it.

`SplitGlobalFlags` also skips global value-flag values: locating the subcommand by first non-flag token reads `bob` out of `bd --actor bob update …`, which would bypass any guard keyed off the subcommand — including this one.

Tests cover both directions — a drift guard proves **every** value-taking `update` flag is a non-false-positive, the real dropped-pair shapes are caught, and the refusal is scoped to `update`.

## What this does not cover

Same failure mode — bd reports success for a partial write and exits 0 — outside this guard's reach:

- **A raw `bd` invocation.** `gc bd` is the only entry point guarded here. Filed upstream as steveyegge/beads#5247.
- **A partial multi-id update.** `bd update <good-id> <bad-id> --set-metadata k=v` writes one bead, fails to resolve the other on stderr, and exits 0. The token carries no `=`, so this guard cannot distinguish it from a legitimate id. Measured on bd 1.1.0.
- **`bdMutationWriteIDs` (pre-existing, `cmd/gc/cmd_bd.go`).** The exact-ID guard added for gcy-g4o takes `sub := args[0]`, so *any* leading global flag — value or boolean — skips it, and `gc bd --json update <id> …` reaches bd unverified. That is a different guard against a different failure (substring resolution mutating the wrong bead), it predates this PR, and this PR does not change it. `SplitGlobalFlags` is the obvious fix, but it widens that guard's activation surface, so it belongs in its own change rather than being smuggled in here.

## Scope

| File | Change |
|---|---|
| `internal/bdflags/bdargs.go` | new — argv parsing (+120) |
| `internal/bdflags/bdflags.go` | `GlobalValueFlags()` accessor (+7) |
| `cmd/gc/bd_mistyped_metadata.go` | new — the guard (+16) |
| `cmd/gc/cmd_bd.go` | hook in `doBd` (+6) |
| tests | `bdargs_test.go`, `bd_mistyped_metadata_test.go` |

## Related

- steveyegge/beads#5247 — the exit-code contract, upstream in bd. This guard is the downstream mitigation; it protects `gc bd` only, not a bd invocation an agent improvises.
- #4901 — publishing the bead DELETE endpoint's soft-delete contract in the spec (same theme: a CLI/API surface that reports success for something other than what the caller asked).

### DIFF
--- a/cmd/gc/bd_mistyped_metadata.go
+++ b/cmd/gc/bd_mistyped_metadata.go
@@ -1,0 +1,16 @@
+package main
+
+import (
+	"github.com/gastownhall/gascity/internal/bdflags"
+)
+
+// mistypedMetadataPairRefusal reports whether a `gc bd` invocation carries
+// `--set-metadata` pairs bd would silently drop, and the message to print.
+//
+// gc bd writes exec raw bd, so nothing upstream of this sees them. It depends
+// only on internal/bdflags — the upstream-owned source of truth for bd's flag
+// names — so this guard carries no fork-local dependency.
+func mistypedMetadataPairRefusal(bdArgs []string) (string, bool) {
+	verb, verbArgs := bdflags.SplitGlobalFlags(bdArgs)
+	return bdflags.DroppedMetadataRefusal("gc bd", verb, verbArgs)
+}
--- a/cmd/gc/bd_mistyped_metadata_test.go
+++ b/cmd/gc/bd_mistyped_metadata_test.go
@@ -1,0 +1,120 @@
+package main
+
+import (
+	"bytes"
+	"strings"
+	"testing"
+)
+
+// TestGCBdRefusesMistypedMetadataPairs pins that `gc bd update` refuses a bare
+// key=value in bd's positional issue-id slot.
+//
+// `--set-metadata` is a repeatable stringArray taking ONE pair per flag, so in
+// `--set-metadata a=1 b=2 c=3` only `a=1` is the flag's value. bd reads `b=2` and
+// `c=3` as issue ids, fails to resolve them, prints the failures to stderr — and
+// still prints its success line and EXITS 0. Measured on bd 1.1.0 and unchanged
+// in 1.1.2: six pairs in, one pair written, exit 0, so no caller can tell.
+//
+// gc bd writes exec raw bd rather than the routed fastpath, so the classifier's
+// guard does not cover this path; it needs its own. The refusal must come before
+// any store work, so nothing is written and the exit code is honest.
+func TestGCBdRefusesMistypedMetadataPairs(t *testing.T) {
+	cases := []struct {
+		name string
+		args []string
+	}{
+		{"trailing pairs", []string{"update", "bd-1", "--set-metadata", "a=1", "b=2", "c=3"}},
+		{"one trailing pair", []string{"update", "bd-1", "--set-metadata", "a=1", "b=2"}},
+		{"inline flag form", []string{"update", "bd-1", "--set-metadata=a=1", "b=2"}},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			var stdout, stderr bytes.Buffer
+			code := doBd(tc.args, &stdout, &stderr)
+			if code == 0 {
+				t.Fatalf("doBd(%v) = 0, want non-zero; stderr=%q", tc.args, stderr.String())
+			}
+			msg := stderr.String()
+			if !strings.Contains(msg, "--set-metadata") {
+				t.Errorf("stderr %q does not name --set-metadata", msg)
+			}
+			// The message must name the offending token, so the caller can see
+			// which pair bd would have dropped.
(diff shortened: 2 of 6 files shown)


quad341 (approved):
@wbern — this is a genuinely good catch, and your writeup did the hard part for me. A silent 1-of-N write behind an exit 0 is the worst kind of bug: every `|| exit 1` in the fleet is blind to it, and the asymmetry you spotted (repeated `--unset-metadata` all apply, only the set path loses pairs) is exactly why nobody noticed.

Two things I especially appreciated. You *proved* the no-regression claim instead of asserting it — I checked `isBeadIDCandidate` in `cmd_sling.go` and it does reject `=`, so a `=`-bearing positional really was already failing, just quietly. And putting the argv parsing in `internal/bdflags` rather than next to the guard means it inherits the existing `freshness_test.go` check against real `bd --help`. Catching the `--actor bob` bypass in your own change, before it shipped, is the kind of self-review that makes a PR easy to trust.

On my side: I'll pin `GC_CITY_PATH` to a temp dir in `TestGCBdAllowsCorrectMetadataForms` so it matches the isolation the sibling tests in `cmd_bd_test.go` use, then re-run CI. The red check is `TestProductMetricsDirectChildEnvSessionSubmitPoller` in `internal/session` — an exec-race flake in a package you don't touch; I confirmed it's the only failure in that job. **Your code needs no changes.**

Thanks also for filing steveyegge/beads#5247 upstream instead of only patching downstream.


---

PR #4912, opened 2026-08-02
feat(beads): model bd's created_by, owner, notes and the gate condition on Bead

wbern:
## The juicy parts

A bead read through `gc` carried strictly less than the same bead read through `bd`. Four plain columns bd emits — `created_by`, `owner`, `notes`, `await_type` — had no field on `beads.Bead`, so every store dropped them while decoding. On the 240-bead store measured here, `created_by`, `owner` and `notes` were populated on **240 of 240** rows and reached no caller. `await_type` is the sharpest case: it is the condition a **gate** bead waits on, so a gate read through the API reported no condition at all.

The audience is anyone consuming the API rather than bd directly — HTTP responses, SSE bead events, the generated Go/TS clients and the dashboard now carry these. All four tags are `omitempty`, so the spec gains four optional string properties and no existing payload changes shape.

This is a projection fix, not a performance change: four hand-written `Bead` mappings each needed the field added by hand, and each gets a round-trip test that was verified to fail without its change.

## TL;DR

`beads.Bead` does not model four of the plain columns `bd` emits, so every
store silently drops them: a bead read through `gc` carries less than the same
bead read through `bd`. This adds `created_by`, `owner`, `notes` and
`await_type` to the domain type and to each of the four hand-written mappings,
with a round-trip test per mapping.

`await_type` is the one that changes behaviour most visibly: it is the
condition a **gate** bead waits on, and without it a gate read reports no
condition at all.

## Evidence

The key set of `bd list --json` is a function of which **rows** are in the sample, not of what bd is willing to emit: `types.Issue` has no custom `MarshalJSON` and almost every field is `omitempty`, so an absent key means an empty value on the rows sampled — nothing more. An earlier revision of this PR drew its evidence from a store with no gates and concluded that bd omitted `await_type`. That was a sampling artifact, and the claim is retracted.

Re-sampled against a store that actually contains a gate — bd v1.1.0, one `gh:pr` gate with `await_id=4912`, one task carrying `notes`:

```
$ bd list --include-gates --all --json   # 17 keys
await_id, await_type, comment_count, created_at, created_by, dependencies,
dependency_count, dependent_count, description, id, issue_type, notes, owner,
priority, status, title, updated_at
```

`bd gate list --json` and `bd show <gate> --json` emit the same pair. The default `bd list` returns 13 keys on that store and no gate row at all, because `bd list` **hides `gate`-type rows by default** (`--include-gates` opts back in). That is a row filter, not a key omission.

So bd emits all five of these on any projection that hydrates a full issue — hydration runs through `sqlbuild.IssueSelectColumns`, "the canonical column list for full issue hydration", which contains `created_by, owner, notes` and `await_type, await_id` — and `beads.Bead` modeled none of them. Every store dropped them while decoding.

The gate condition is a **pair**. bd renders a gate as `"%s %s", AwaitType, AwaitID` (`cmd/bd/gate.go`), so carrying only the type would report that a gate waits on a PR without saying which one. Both are modeled here.

## Behaviour

**Given** a store whose schema has bd's plain columns
**When** a bead is read through any provider
**Then** `created_by`, `owner`, `notes` and `await_type` reach `beads.Bead`,
the HTTP response, the SSE bead event and the generated clients

**Given** a gate bead with `await_type = "human"`
**When** it is listed
**Then** the condition is carried instead of dropped

**Given** a snapshot whose schema predates those columns
**When** a list read runs
**Then** the fields resolve to empty and the read still succeeds — unchanged
behaviour for every other field

**Given** a bead that carries none of the four
**When** it is decoded from the generated client
**Then** the fields are empty; the wire tags are `omitempty`, so nothing is
fabricated and no payload grows

## Why it is safe

The doltlite projection resolves all four through the same schema-presence
probe (`tableHasColumn`) that the `ephemeral` / `no_history` storage flags
already use. That is load-bearing rather than defensive: **these columns land
in the main list `SELECT`**, so on an older snapshot an unguarded reference
would fail *every list read* rather than omit one field. Both the present and
absent paths are pinned by tests.

All four tags are `omitempty`, so the OpenAPI schema gains four optional
string properties and no existing payload changes shape.

The four probes are issued per query build alongside the two existing ones.
They are in-process `pragma_table_info` lookups; if the added probe traffic is
unwelcome, memoizing the probe is a clean follow-up, but it would trade a
schema re-read for an assumption that a column set cannot change under an open
handle, so it is deliberately not bundled here.

## The part that is easy to get wrong

**Five** hand-written mappings convert `Bead` to and from wire form, and a field added to the struct flows through **none** of them automatically:

| Mapping | File | |
| --- | --- | --- |
| `bdIssue.toBead` | `internal/beads/bdstore.go` | |
| `beadFromNativeIssue` | `internal/beads/native_dolt_store.go` | |
| `beadWire.toBead` | `internal/beads/exec` | consumer |
| `bridgeBead` | `cmd/gc/cmd_bd_store_bridge.go` | **its producer** |
| `beadFromGen` | `internal/api/decode_convoys.go` | |

`bridgeBead` is the one that makes the others count. `gc bd-store-bridge` is the builtin exec-store path — the bundled bd pack script invokes it — and it re-marshals `BdStore`'s output into its own narrower envelope before the exec consumer decodes anything. A field it omits is gone end to end no matter what the consumer declares, so fixing `beadWire` alone would have left an exec-backed city dropping all five.

Its test therefore drives the real `get` op against a `bd` that emits the columns and asserts they survive to stdout, rather than feeding the consumer a hand-written payload that already contains the answer. A test fed the keys it is checking for cannot fail.

The schema probes are issued per query build: seven now (the two existing storage flags plus these five) where there were two. They are in-process `pragma_table_info` lookups, measured at ~13 µs each, and the probe count is independent of row count — for a typed list that is 14 probes where there were 4, and 28 where there were 8 for a limited/paged read, i.e. roughly +0.1–0.2 ms on a read measured in tens of ms. Memoizing the probe is a clean follow-up, but it would trade a schema re-read for an assumption that a column set cannot change under an open doltlite handle, so it is deliberately not bundled here.

## Scope

Out of scope, explicitly. `comment_count`, `dependency_count` and `dependent_count` are computed over relations and comments rather than stored columns. So are the other stored columns bd carries that `Bead` still does not model — `design`, `acceptance_criteria`, `spec_id`, `timeout_ns`, `waiters`, `started_at`, `closed_at`, `due_at`, `external_ref`, `sender`, `pinned`, `is_template`, `wisp_type`, `source_formula`, `mol_type`, `work_type` and others. This change is not "everything bd has"; it is the fields whose absence loses information a caller is already asking for, plus the gate condition, which was reported as absent entirely.

All five are a **read-only projection**, and the struct says so. No write path persists them: they are absent from `Create`'s argument list, from `UpdateOpts`, and from every exec create/update request shape. `store.Create(beads.Bead{Notes: "x"})` compiles and drops the note — populate these through bd (`bd gate create`, `bd update --notes`) and read them back here.

## Gates

`go build ./...`, `go vet ./...`, `go test ./internal/beads/... ./internal/api/...`,
`go test -tags gascity_native_beads ./internal/beads/`, and `make dashboard-ci`
all pass.


wbern:
Rebased onto current `main` — this is mergeable again.

Worth flagging the shape of the conflict, because it will come back rather than stay fixed. Every conflict on this PR has been in the generated SPA bundle (`internal/api/dashboardspa/dist/**`), never in source. The `zod.gen.ts` additions are runtime values rather than type-only, so the bundle genuinely has to be rebuilt here — which means this PR re-conflicts with any upstream commit that also rebuilds it. It went from clean to conflicting again within about thirty minutes today, purely from bundle churn.

I re-ran it as: rebase, take `main`'s `dist/` wholesale, then `make dashboard-build` from the merged source. Verified the local toolchain reproduces the committed bundle byte-for-byte on a clean `main` checkout first, so the new hashes come from this PR's source change and not from a toolchain difference on my side.

Gates on the rebased head: `go build ./...` and `go vet` clean, `internal/beads` and `internal/api` green, `TestOpenAPISpecInSync` green, `make dashboard-check` green.

If it conflicts again before you get to it, please don't bounce it back — regenerating `dist/` at merge time is a mechanical step and you're better positioned to win that race than I am. Happy to keep rebasing on request either way.

quad341:
Thanks for this, @wbern — there is clearly a lot of work here.

wbern:
That is fair, and the split is the right call — I will do it.

One correction to the size picture that may help, since it points at where the bulk actually is. The change is 1,390 lines across 45 files, not 20,000. What blows past the ceiling is bytes rather than lines: 23 of those files are the minified `internal/api/dashboardspa/dist` bundle, which is ~979 KB of the 1.0 MB diff. The hand-written part is 22 files and ~46 KB. It is already marked `-diff linguist-generated` in `.gitattributes`, so the web view hides it, but `gh pr diff` still serves the raw patch — which I suspect is what your tooling read.

I cannot simply drop the bundle, since it is tracked and the Makefile gates on `git diff --quiet -- internal/api/dashboardspa/dist`, so it has to move in step with the generated types. But it only needs to appear in one of the pieces.

Proposed split, smallest first, each landing on its own:

1. **beads core** — model the four fields on `Bead` and read them in the store implementations (`internal/beads/{beads,bdstore,native_dolt_store,doltlite_read_store,caching_store_events}.go`, `internal/beads/exec/{exec,json}.go`) plus the per-store column tests. No API surface, no generated output.
2. **`gc bd` bridge** — `cmd/gc/cmd_bd_store_bridge.go` and its test, on top of 1.
3. **API projection** — `internal/api/{decode_convoys,convoy_sql}.go` and the regenerated OpenAPI spec, genclient, TS types and SPA bundle. This is the only piece carrying the generated churn, and its hand-written part is small.

For context on why the fields are wanted: `created_by`, `owner`, `notes` and the gate await condition are all absent from `Bead` on current `main`, so API, SSE and dashboard consumers cannot see them at all even though bd carries them.

I will open 1 first and keep this PR open until it lands, then rebase the rest onto it. If you would rather I close this one now and reopen as three fresh PRs, say so and I will.

wbern:
Part 1 of the split is up as #6405 — the beads-core model change, ~47 KB against the ~1.0 MB here, standing alone with no dependency on the rest.

One thing the split surfaced that is worth stating plainly: `Bead` is projected into the generated OpenAPI spec, so the five json tags necessarily regenerate the spec, genclient, TS types and the SPA bundle. I could not keep part 1 completely free of generated output. It is mechanical though — the bundle assets differ only in content hashes (`23 files changed, 0 insertions, 0 deletions`), and the schema addition is 15 lines.

Parts 2 (`gc bd` bridge) and 3 (API projection: `decode_convoys.go`, `convoy_sql.go`) follow once #6405 lands. Keeping this open as the tracking issue until then; say the word if you would rather I close it now.

Issue #2848, opened 2026-05-31
gc sling --on <formula>: routed_to metadata set on origin bead instead of molecule bead


nagyv-cvio:
## Bug

When using wisp dispatch (`gc sling <agent> <bead-id> --on <formula>`), the `routed_to` metadata is set on the **origin bead** instead of the **molecule bead**.

## Reproduction steps

1. Create a bead:
   ```
   bd create "dummy test bead" --rig ralph-wiggum
   # → rw-0ij
   ```

2. Sling with a formula:
   ```
   gc sling ralph-wiggum/partner.claude rw-0ij --on mr-address-feedback
   ```

3. Inspect the created beads:
   - Origin bead: `rw-0ij`
   - Sling bead: `rw-9rz`
   - Molecule bead: `rw-byk`

## Expected behavior

`routed_to` should be set on the molecule bead `rw-byk`.

## Actual behavior

`routed_to` is set on the origin bead `rw-0ij`.

PR #4913, opened 2026-08-02
test(sling): cold-pool regression guard — --on must route the source bead and keep it scale_check-visible (#2848)

edwarddavison:
Regression test for the attachFormulaToBead source-bead routing fix that landed on main (the v1.4.0 release line diverged before it and routes the workflow root instead — on a cold pool the root goes in_progress immediately, scale_check sees zero demand, and nothing ever spawns; we hit this in production on gc 1.4.0 within hours of adopting formula-attached dispatch).

The test asserts, for `sling <pool> <bead> --on <v2-formula>` against a cold pool: `gc.routed_to` lands on the WORK bead (not the workflow root), and the bead remains scale_check-visible (open + routed + unassigned) until claim. Guards the fix on main against regression; no production code changes.

+219, single file: `cmd/gc/sling_on_cold_pool_demand_test.go`.

### DIFF
--- a/cmd/gc/sling_on_cold_pool_demand_test.go
+++ b/cmd/gc/sling_on_cold_pool_demand_test.go
@@ -1,0 +1,219 @@
+package main
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/gastownhall/gascity/internal/beads"
+	"github.com/gastownhall/gascity/internal/config"
+	"github.com/gastownhall/gascity/internal/runtime"
+)
+
+// These tests guard the gci-7ba / gastownhall/gascity#2848 regression: when a
+// formula is attached-and-routed onto an existing work bead with
+// `gc sling <pool> <bead> --on <formula>`, the pool's demand-bearing unit must
+// stay open + routed + unassigned so a COLD pool (no live sessions) computes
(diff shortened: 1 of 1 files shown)


quad341 (approved):
@edwarddavison — this is the regression guard I wish more test PRs were. You didn't stop at asserting that `gc.routed_to` lands on the right bead; you carried it through to `defaultScaleCheckCounts` and asserted the thing that actually hurt you in production — a cold pool seeing demand > 0. That's the invariant nobody had pinned down, and covering both the legacy `[[steps]]` path and the graph.v2 pour separately is what makes it durable rather than incidental.

We mutation-tested both before deciding: routing the wisp root instead of the source fails your legacy test, and stripping the pool step's route fails your graph.v2 test. Neither passes vacuously — which is more than a lot of regression tests can say. The header comment explaining why the two attach paths satisfy the invariant differently is going to save someone a bad afternoon.

Taking it as-is. Nothing for you to do. The only thing we're adding on our side is a rebase onto current `main` and a re-run, since your green CI is a couple of weeks old and these assertions are deliberately sharp about bead shape. Thanks for chasing this from a production incident down to a test.


---


Review coverage:
- completed: `qwen claude`
- unavailable: `codex`

Issue #4441, opened 2026-07-19
Proposal: route hot `bd` verbs to the warm controller via a tiny thin client (8× faster point lookups)


wbern:
## Context

Gas City deliberately has agents reach subsystems through their CLIs rather than
registered tools or MCP — per the project's own principle, *"No MCP/tool
registration — if a tool has a CLI, the agent uses it."* The beads work ledger
follows that model: agents drive work by calling `bd`, and order/gate condition
scripts shell out to it too. A healthy consequence is that `bd`'s per-invocation
cost is a first-class performance concern by design — every agent action and
gate evaluation pays it.

Each `bd` invocation is a fresh, standalone process. For the cache-servable read
verbs it does that work from cold — even though a warm `gc` controller is already
running and holds the answer in memory. Plain `bd` simply doesn't use it. That is
the opportunity: route the hot verbs to the already-warm controller instead of
recomputing them per call.

## What we built

`bdshim` — a small, pure-Go, `bd`-CLI-compatible front end that:

- **Routes** the cache-servable verbs (`show`, `list`, `query`, `update
  --claim`) through the already-warm controller HTTP API (the same endpoints, so
  output is byte-identical), and
- **Passes through** everything else by exec'ing the real `bd` (`GC_BD_REAL`)
  with the caller's env/cwd.
- Is **opt-in / default-off** (a `[session] bd_shim = auto|on|off` toggle) and
  fronted on the agent's PATH.
- **Passthrough is always byte-identical** to raw `bd`; routing is a pure latency
  optimization, never a correctness requirement. Controller-down → routed reads
  fail loud (rc=1) rather than silently returning a wrong/empty local answer.

## Measured wins (live gc2, warm controller, n=25/verb)

`bdshim` vs the real `bd`:

| verb              | wall (bdshim / bd) | CPU (bdshim / bd) | result                                              |
| ----------------- | ------------------ | ----------------- | --------------------------------------------------- |
| `show <id>`       | **14.5 / 117 ms**  | **7.2 / 102 ms**  | 🟢 **routed — 8.1× faster wall, ~14× less CPU**     |
| `list --status …` | 126 / 120 ms       | 117 / 109 ms      | ⚪ passthrough — identical to real `bd`             |
| `ready`           | 463 / 123 ms       | 9.4 / 108 ms      | 🟡 routed but slower — federated round-trip (separate lever) |

`show <id>` is the point-lookup verb agents run most, so the 8.1× is the win
[...]

PR #4916, opened 2026-08-02
test(bench): measure the readiness pass a city with locked imports pays

wbern:
## The juicy parts

**There is no end-user win in this PR — it is test-only, and that is the point.** `BenchmarkBuiltinReadinessPass` builds its fixture city without a `packs.lock`, so the locked-imports half of `EnsureBuiltinRuntimeAssets` — one of the two places it validates a synthetic pack cache — never executes under it. The benchmark measures roughly half of the pass it is named after, and it scored a change measured at ~51 ms of CPU per real invocation as *exactly zero*, with allocation counts identical to the digit.

`BenchmarkBuiltinReadinessPassLocked` makes the missing walk visible: 42,135 vs 63,334 allocs/op, a ratio of 1.5031 that is stable across rebases. That is what makes the readiness-pass numbers in the sibling PRs checkable at all, and what would catch a future duplicate walk in the half nothing currently watches. Note that nothing in CI runs `-bench` today, so this is a tool for a human, not an automated guard.

## What this is

A second benchmark for the builtin readiness pass, covering the case the
existing one structurally cannot see: a city that has a `packs.lock`.

Test-only. No production code changes.

## Why I was looking at this

I have been profiling `gc`'s own startup cost while working on
[#4441](https://github.com/gastownhall/gascity/issues/4441). This benchmark gap
is not part of that proposal — it is something I ran into underneath it, and it
looks like a general-benefit fix worth sending on its own regardless of what
happens to #4441. Reviewing it as an independent change is the right call; it
does not depend on anything else I have open.

The specific way I found it: a change I could measure at roughly 51 ms of CPU
per real invocation came out as **exactly zero** against
`BenchmarkBuiltinReadinessPass`, with allocation counts identical to the digit.
That is a strange enough result to chase.

## The gap

`BenchmarkBuiltinReadinessPass` builds its fixture city with no `packs.lock`.

With no lockfile, `lockedBundledCanonicalImports` finds nothing, so the
locked-imports half of `EnsureBuiltinRuntimeAssets` — one of the two places it
validates a synthetic pack cache — never runs at all. The benchmark measures
roughly half of the pass it is named after.

A locked bundled source resolves to its own cache directory, and
`MaterializeSyntheticRepo` writes every bundled pack layout into every such
directory. Validating it is therefore a **second full walk of the whole bundled
pack set**, not an incremental check.

## The measurement

`BenchmarkBuiltinReadinessPassLocked` pins a bundled source at its canonical
commit and clears that city's ready memo per iteration:

| benchmark | allocs/op |
| --- | ---: |
| `BenchmarkBuiltinReadinessPass` | 42,135 |
| `BenchmarkBuiltinReadinessPassLocked` | 63,334 |

Ratio **1.5031**. (Representative run at `-benchtime 10x`. Counts move by a
handful of allocations between runs; the ratio does not — I saw 1.5031 and
1.5033 across rebases onto different upstream bases.)

That ratio is the argument. One full walk of the bundled pack set is ~21,068
allocations, so the warm shape is doing two walks and the locked shape three.
A benchmark that cannot see the third walk will approve its removal — or its
accidental duplication — as exactly zero.

I have quoted allocations rather than wall-clock deliberately. Allocation counts
here are deterministic to the digit. The `ns/op` figures move with machine load,
and this change is not about a time saving in the first place.

## Attribution — I checked which half actually costs

Two things separate the two shapes, and I decomposed them 2x2 rather than
assuming:

| axis | effect |
| --- | ---: |
| adding the lockfile | **+50.1% allocs** |
| clearing the per-city memo | +0.26% allocs |

So it is essentially all lockfile. The memo clearing is kept because it makes
the shape honest — a one-shot `gc` process starts `builtinRuntimeReadyCache`
empty and always takes the non-memoized revalidation path — but it is a
correctness detail, not where the cost is. The benchmark is named `Locked` for
that reason rather than `Cold`, which would point at the wrong half.

## Fidelity, stated plainly

This reproduces the **shape** of the cost, not the exact pass a real city runs.

The fixture locks `gastown`, which the fixture city does not import, so the
locked arm walks the bundled pack set three times. A real bd-provider `gc init`
city locks `core` and `bd`, which normalize to the same clone URL, and pays
four. The benchmark understates a real city — conservative in the direction
that matters for a guard.

## Two traps pinned in the fixture

Both cost me time, so they are asserted rather than commented:

- **gastown's canonical pin is `PublicGastownPackVersion`, not
  `BundledPackImportVersion`.** Locked at the latter it is an ordinary remote
  import that the preflight skips entirely, which leaves the benchmark
  measuring a city with no locked bundled imports after all. My first draft did
  exactly that and reported no change. There is now a `b.Fatal` on an empty
  `lockedBundledCanonicalImports`.
- **The readiness memo is a package-level `sync.Map` shared by every test in
  the binary.** The reset deletes only this city's key rather than ranging over
  the map and evicting entries it does not own.

## Notes for the reviewer

- `writePreflightImportLock` changes from `*testing.T` to `testing.TB` so the
  benchmark builds the same locked-city shape the preflight tests already
  assert on. That is the only change to an existing file.
- The benchmark is hermetic — no network.
- If you re-run these, note that `go test -bench -count 6` does **not**
  interleave arms; it runs all six of A then all six of B, so a host that
  changes load mid-run fabricates a difference. Interleave with an outer shell
  loop of `-count 1`. The allocation figures above are stable either way.
- Nothing in CI currently runs `-bench`, so the fixture's `b.Fatal` guard only
  fires for a human running it directly. Happy to wire it into a workflow if
  you would like that, but I did not want to add CI time unasked.

### DIFF
--- a/cmd/gc/builtin_readiness_cost_bench_test.go
+++ b/cmd/gc/builtin_readiness_cost_bench_test.go
@@ -4,7 +4,10 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
+
+	"github.com/gastownhall/gascity/internal/config"
 )
 
 // newReadinessCostCity writes a minimal bd-provider city.
@@ -25,6 +28,11 @@ func newReadinessCostCity(b *testing.B) string {
 // Read this against BenchmarkCityConfigParseOnly. The readiness pass, not the
 // parse, is what a config load costs — which is why skipping a redundant load
 // is worth anything, and why the pass itself must still run once per process.
+//
+// This shape deliberately measures the cheapest one: no packs.lock and a warm
+// memo. Use BenchmarkBuiltinReadinessPassLocked to measure the pass a city with
+// locked bundled imports pays; that benchmark's comment explains what this one
+// cannot see.
 func BenchmarkBuiltinReadinessPass(b *testing.B) {
 	b.Setenv("GC_HOME", b.TempDir())
 	cityPath := newReadinessCostCity(b)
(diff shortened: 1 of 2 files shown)


quad341 (approved):
@wbern — this is a great catch, and the way you found it is the best part: a change you could measure at ~51 ms of CPU coming back as *exactly zero*, with allocation counts identical to the digit, is exactly the kind of too-clean result that should make someone suspicious. Most people would have shrugged and moved on.

You're right about the gap. The fixture city has no `packs.lock`, so `lockedBundledCanonicalImports` finds nothing and the locked-imports half of the pass never runs — the benchmark has been measuring roughly half of what it's named after. I traced the pieces through the tree before agreeing: the memo key matches `normalizePathForCompare`, the canonical-pin claim holds against the existing preflight tests, and the `b.Fatal` guard means the fixture fails loudly rather than quietly measuring the warm shape twice if that pin ever moves. That guard is the detail that makes this worth keeping.

Merging as-is. On wiring `-bench` into CI: yes please, but as its own PR — no reason to hold this one for it.

Thanks also for quoting allocations rather than wall-clock, and for the note about `-count 6` not interleaving arms. Both saved the next person real time.


---
**Synthesis**: claude-opus-5


Review coverage:
- completed: `qwen codex`
- unavailable: `claude`

PR #4917, opened 2026-08-02
perf(config): let a loader that discards provenance decline the revision snapshot

wbern:
## The juicy parts

**There is no user-visible behaviour change here, and the measured win on upstream's fixture city is small**: `-36 allocs/op (-1.2%)` and `-1.1% B/op` on the existing `BenchmarkCityConfigParseOnly`, deterministic to the digit across three runs per arm. I am **not** quoting a wall-clock figure — on that fixture it sits inside run-to-run noise, and I removed a significance test rather than report one reached by optional stopping.

The case for the change is structural, not numeric. The three city-config loaders in `cmd_agent.go` return only `*config.City` and drop the `Provenance`, so nothing they load can ever observe the load-time revision snapshot they pay to build. Every reader of that snapshot already falls back to reading from disk, so declining it cannot change a revision value — that argument is the part I would most like checked.

The skipped work is one `PackContentHashRecursive` per pack directory, so it scales with the pack set, but that measurement comes from a city I cannot ship as a fixture and is not the headline. The option is opt-in; every other caller is byte-for-byte unchanged.

## What this is

`config.LoadOptions` gains `SkipRevisionSnapshot`, and the three city-config
loaders in `cmd/gc/cmd_agent.go` opt in.

The option is **opt-in**, so every other caller keeps today's behaviour byte for
byte.

## Why I was looking at this

that proposal and does not depend on it — it is a general-benefit change I ran
into underneath it, and it should be reviewed entirely on its own merits.

## The observation

`config.LoadWithIncludes` always captures a load-time revision snapshot, which
content-hashes every pack directory with `PackContentHashRecursive` so a later
`config.Revision()` can compare against the tree as it was loaded.

The three loaders in `cmd_agent.go` return only `*config.City`. They use the
`Provenance` to emit warnings and then drop it. Nothing they load can observe
the snapshot, so on those paths building it is unobservable work.

## Why declining it cannot change a revision value

This is the part I would most like checked, so here is the whole argument.

Every reader of the snapshot already falls back to reading from disk:

| reader | fallback |
| --- | --- |
| `writeRevisionDirHash` | `PackContentHashRecursive` |
| `revisionSnapshotFile` | `fs.ReadFile` |
| `revisionConventionDirs` | `existingConventionDiscoveryDirsFS` |

The snapshot is a **prefetch, not an input**. Declining it cannot produce a
different revision value.

What it does give up is *load-time faithfulness*: a `Revision()` computed later
reflects the tree as it is then, rather than as it was at load. That distinction
only matters to a caller holding a `Provenance` across a window in which config
may change — the long-running case, which loads through
`loadCityConfigWithBuiltinPacks` and keeps the default.

`captureRevisionSnapshot` also calls `recordMissingSourceContents`. I checked
every reader of `sourceContents`; none observes it on these paths.

## Measurement, and what I am deliberately not claiming

| | allocs/op | B/op |
| --- | ---: | ---: |
| `BenchmarkCityConfigParseOnly` before | 3,108 | 381,177 |
| after | 3,072 | 376,912 |
| | **-36 (-1.2%)** | **-1.1%** |

Those counts are deterministic — identical to the digit across three runs of
each arm.

**I am not quoting a wall-clock figure.** On upstream's small fixture city the
saving sits inside run-to-run noise. I originally had a significance test here
and removed it: the sample size had been extended *because* the first one was
not significant, which is optional stopping, and the resulting p-value would not
have meant what it appeared to mean. Rather than patch that up I have dropped
the timing claim and left the deterministic figure and the structural argument,
which are what actually support the change.

The saving scales with the pack set — the skipped work is one
`PackContentHashRecursive` per pack directory — so on a large city with many
packs it is worth materially more than it is on the upstream fixture. That
measurement comes from a city I cannot ship as a fixture, so I am not making it
the headline. **The case for this change is that the work is unobservable, not
that it is large.**

## The guard

`TestCityConfigLoadersDeclineTheRevisionSnapshot` is a source-level test rather
than a behavioural one, because the cost is invisible by construction: a loader
that reverts returns exactly the same config and passes every functional test.
It just re-reads every pack file. Nothing fails; it only gets slower — precisely
the regression a test suite does not otherwise catch.

It follows the file-scanning idiom of
`TestGCNonTestFilesStayOnRigProvisionBoundary` (`runtime.Caller(0)` +
`filepath.Dir` + substring needles) rather than inventing a new one.

I mutation-tested it, since a guard that cannot fail is worse than no guard:

- reverting a loader to `config.LoadWithIncludes(` → fails, naming the file
- keeping `...Options(` but dropping the option → fails, naming file and line

## Scope

Roughly a dozen other non-test call sites discard the `Provenance` the same way
— `init_provider_readiness.go`, `cmd_extmsg.go`, `doctor_v2_checks.go`,
`cmd_convoy.go`, `cmd_init.go`, `cmd_wait.go`, `internal/dispatch/control.go`
and others.

They are deliberately **out of scope** here and the guard does not speak for
them. I would rather land the narrow version and let you tell me whether the
broader sweep is wanted.

### DIFF
--- a/cmd/gc/cmd_agent.go
+++ b/cmd/gc/cmd_agent.go
@@ -33,6 +33,19 @@ func loadCityConfig(cityPath string, warningWriter ...io.Writer) (*config.City, 
 	return loadCityConfigFS(fsys.OSFS{}, filepath.Join(cityPath, "city.toml"), warningWriter...)
 }
 
+// skipRevisionSnapshot is the load option shared by the loaders in this file.
+//
+// The load-time revision snapshot content-hashes every pack directory so that a
+// later config.Revision() call can compare against the tree as it was loaded.
+// Neither loader here returns the Provenance — both use it to emit warnings and
+// then drop it — so nothing they load can ever observe the snapshot, and
+// building it is pure cost on a one-shot command. Loaders that do hand the
+// Provenance back keep the default.
+//
+// Revision reads from disk for anything the snapshot does not hold, so declining
+// it changes no revision value; see config.LoadOptions.SkipRevisionSnapshot.
+var skipRevisionSnapshot = config.LoadOptions{SkipRevisionSnapshot: true}
+
 // loadCityConfigFS is the testable variant of loadCityConfig that accepts a
 // filesystem implementation. Used by functions that take an fsys.FS parameter
 // for unit testing.
@@ -40,7 +53,7 @@ func loadCityConfigFS(fs fsys.FS, tomlPath string, warningWriter ...io.Writer) (
 	if err := ensureBuiltinPacksForConfigLoad(fs, tomlPath, resolveLoadCityConfigWarningWriter(warningWriter...)); err != nil {
 		return nil, err
 	}
-	cfg, prov, err := config.LoadWithIncludes(fs, tomlPath)
+	cfg, prov, err := config.LoadWithIncludesOptions(fs, tomlPath, skipRevisionSnapshot)
 	if err != nil {
 		return nil, err
 	}
@@ -59,7 +72,7 @@ func loadCityConfigFS(fs fsys.FS, tomlPath string, warningWriter ...io.Writer) (
 // briefly reflect stale builtin-pack content after an upgrade until a normal
 // gc command refreshes the generated packs.
 func loadCityConfigWithoutBuiltinPackRefreshFS(fs fsys.FS, tomlPath string, warningWriter ...io.Writer) (*config.City, error) {
-	cfg, prov, err := config.LoadWithIncludes(fs, tomlPath)
+	cfg, prov, err := config.LoadWithIncludesOptions(fs, tomlPath, skipRevisionSnapshot)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


quad341 (approved):
Thanks @wbern — this is a genuinely well-argued change, and I want to call out the part most people would have skipped: you pulled your own significance test because the sample had been extended after a null result, and shipped the deterministic alloc counts and the structural argument instead. That is the right call, and it made the PR much easier to trust.

You asked for the "declining it cannot change a revision value" argument to be checked, so I traced it rather than taking it on faith. All three snapshot readers fall back to disk as you describe, and `recordMissingSourceContents` only feeds `sourceContents`, which nothing but `Revision()` reads — so it can't quietly swallow a load warning either. The long-running paths go through `loadCityConfigWithBuiltinPacks` and keep the default, exactly as your writeup says.

Merging as-is. On the broader sweep: yes, I'd like it, as a follow-up PR — landing the narrow version first was the right instinct. The source guard is slightly brittle (it wants the option on the same line as the call), but that's a noisy-test risk, not a missed-regression one, and not worth holding this.


---
**Synthesis**: claude-opus-5


Review coverage:
- completed: `qwen codex`
- unavailable: `claude`

PR #4768, opened 2026-07-28
perf(cli): reuse the write-guard's store read in the bd close gate

wbern:
**Part 2 of 2 — `gc bd` invocation cost.** **#4723** removes redundant *config
loads* from the same invocation; this PR removes a duplicate *store open + bead
read* from the close gate. They are separate wins with separate evidence, but
they overlap in code and are best reviewed as a pair — both change the
signature of `runWorkRecordCloseGate`, so whichever merges second needs a
rebase (details under *Related work*). Earlier and already merged in this same
line of work: #4565.

---

## Summary

`gc bd close` spawns five `bd` subprocesses. Two of them are a verbatim repeat
of the two before them:

```
context --json
show --json <id>
context --json      <- duplicate
show --json <id>    <- duplicate
close <id>
```

The exact-ID write guard opens a store and reads every bead the invocation is
about to mutate. The work-record close gate then runs immediately after and
does the same two things again — its own `openStoreAtForCity`, its own
`store.Get` of the same id — because the two were written independently and
neither knows the other ran.

This threads the store and beads the guard already opened into
`runWorkRecordCloseGate`. **`gc bd close` now spawns three `bd` subprocesses
instead of five.** No output changes, no new flag, no new user-facing surface.

## What it costs today

Counted with a `bd` shim on `PATH` that records each invocation and execs the
real `bd`, against this branch and against pristine `main`:

| verb | `main` | this branch |
| --- | ---: | ---: |
| `gc bd close` | 5 subprocesses | **3** |
| `gc bd show` | 1 | 1 |

A count is the honest measurement here: it is what the change actually
controls, and unlike wall clock it cannot be distorted by load on the
measuring box.

For the wall-clock share those two subprocesses represent, profiling `gc bd
close` on a clean, isolated city (n=21, p50 2779.3 ms) attributed the
duplicate pair as follows:

| component | ms | share |
| --- | ---: | ---: |
| close-gate `store.Get` (duplicate) | 550.4 | 19.8% |
| close-gate `openStore` (duplicate) | 373.0 | 13.4% |
| **duplicate pair total** | **923.4** | **33.2%** |

**Read that as an attributed upper bound, not a measured speedup.** It is the
share of wall clock those two operations accounted for in the profile; it is
not an end-to-end A/B of the two binaries. The absolute milliseconds are
inflated — the box was under load ~9.5 — so the proportion is the meaningful
figure and the absolutes are not. An interleaved end-to-end A/B needs a quiet
[...]

PR #4723, opened 2026-07-27
perf(cli): stop reloading the city config inside a store open

wbern:
**Part 1 of 2 — `gc bd` invocation cost.** This PR removes redundant *config
loads*; **#4768** removes a duplicate *store open + bead read* from the same
close gate. They are separate wins with separate evidence, but they overlap in
code and are best reviewed as a pair — see [Overlaps with #4768](#overlaps-with-4768--please-read-before-merging-either)
below. Earlier and already merged in this same line of work: #4565.

---

> **Description rewritten 2026-07-28.** This PR originally claimed a read-path
> win. That claim was wrong and I withdrew it in
> [a correction comment](https://github.com/gastownhall/gascity/pull/4723#issuecomment-5097444219)
> after re-measuring on a clean upstream city. The description below is the
> corrected version — I have rewritten it in place rather than leave a
> superseded premise as the first thing a reviewer reads. The original text is
> preserved in the edit history and the correction comment is unchanged.

## Summary

`gc bd` **writes** cross the config-load boundary more than once per
invocation. `gc bd close` crosses it three times, `gc bd update` twice. Every
one of those extra loads already had the config available at the call site.
This makes them use it, so both write shapes cross exactly once.

**Reads are not affected.** `gc bd show/list/ready` already crossed exactly
once on `main`; there is no read-path redundancy to remove. See *What I got
wrong* below.

## What I got wrong

My original framing said `gc bd` loaded the config three times per invocation
on *every* shape, including reads. That came from profiling a **fork build on
a large downstream city**, not from upstream. On pristine `main` (`0e6462a36`),
instrumented to count `EnsureBuiltinRuntimeAssets` calls per invocation on a
throwaway city (isolated `GC_HOME`, one rig, managed Dolt, real `bd` on PATH):

| invocation | readiness passes | time in passes |
| --- | --- | --- |
| `gc bd show` / `list` / `ready --json` | **1** | 67–146 ms |
| `gc bd update` | **2** | 129 ms |
| `gc bd close` | **3** | 220 ms |

An end-to-end A/B over five read shapes (n=21 interleaved, pristine vs this
branch) put every read delta inside the noise band, with byte-identical output.
The read-path claim is withdrawn. The write-path finding survived the same
re-measurement, and that is what this PR now rests on.

## The win, measured

| invocation | before | after |
| --- | --- | --- |
| `gc bd close` | 3 passes / 220 ms | **1 pass / 70 ms** |
[...]

PR #4565, opened 2026-07-22
perf(cli): skip pack discovery for gc bd

wbern:
## Summary

Skip pack-command discovery when the injected root command is the built-in `gc bd` wrapper.

`bd` is already reserved by the built-in Cobra tree, so a pack cannot provide that command. Loading city configuration and materializing pack commands during root construction is therefore unnecessary for every `gc bd ...` invocation.

## Behavior

Given a city with pack commands, when `gc bd ...` is invoked, then the built-in `bd` command is constructed without loading or materializing pack commands.

Given a scoped local, remote-context, or remote-URL `gc bd ...` invocation, when the root grammar identifies `bd`, then the same discovery skip applies.

Given an unknown root flag, `--` terminator, or an argument value that happens to be `bd`, when the pre-scan is ambiguous, then discovery remains enabled.

Given any `gc bd` operation, its normal city/rig scope resolution, argument forwarding, environment construction, and `bd` child-process behavior are unchanged.

## Local benchmark

This is a focused A/B of this PR, not a benchmark of the separate thin-client work. On one pack-configured local city, `gc bd show gc2-z7j83 --json` ran 20 times per arm in randomized interleaved order:

| build | median wall time | p95 wall time |
| --- | ---: | ---: |
| base | 716.7 ms | 823.5 ms |
| this PR | 455.8 ms | 488.2 ms |

That is 260.9 ms, or 36.4%, faster at the median. The JSON output was byte-for-byte identical between the two builds. Results will vary with city and pack configuration; this measurement does not claim an improvement for every `bd` operation.

## Evidence

- Added root-argument cases for bare, scoped, remote-context, and remote-URL `gc bd` invocations, plus conservative parsing cases.
- Added a root-construction regression proving a city pack command is not materialized for injected `gc bd` arguments.
- `go test -count=10 ./cmd/gc -run '^(TestRootCommandOptionsSkipPackDiscoveryForBuiltinCommands|TestRootConstructionUsesInjectedArgsInsteadOfAmbientOSArgs)$'` passed in 4.18s.
- `go build ./cmd/gc` and `go vet ./...` passed.

`make test-fast-parallel` was also attempted locally but did not complete because of unrelated host/baseline failures: two `cmd/gc` stop-test timeouts, macOS path/FD-limit cases, and the existing `/bin/bash` `${path,,}` incompatibility in `scripts/rebase-resolve-lib.sh`.

## Why this is safe

[...]

PR #4919, opened 2026-08-02
perf(preflight): validate the shared synthetic pack cache once per readiness pass

wbern:
## The juicy parts

A user waits on the builtin readiness pass on every single `gc` invocation, and that pass currently walks and byte-compares the same synthetic cache directory several times over, because each bundled source asks about it separately. After this, each `(cacheDir, commit)` pair is validated once per pass: **4 full walks down to 1** on a bd-provider city with a `packs.lock`, and **5 down to 2** on a `gc init --config gastown` city.

The evidence is structural rather than timed — the call-count claim is verifiable by reading the diff, with no benchmark. **I have no wall-clock number of my own for this one and I am not borrowing anyone else's.** The saving is also smaller on a pass that actually repairs a cache (~3 walks, not 1), and the self-healing contract is unchanged: only positive verdicts are memoized, and a verdict never outlives the pass that produced it.

## What this is

The builtin readiness pass now validates each synthetic pack cache directory
once, instead of once per bundled source that happens to resolve to it.

For a plain bd-provider city — where `packs.lock` pins `core` and `bd`, both of
which resolve to the same directory — that is **4 full walks of the bundled
pack tree per pass, down to 1**. For a `gc init --config gastown` city, which
additionally locks the public gastown pack in a second repository, it is **5
down to 2**.

No behaviour change. The self-healing contract is untouched: the full validator
still runs at least once every pass.

## Why I was looking at this

that proposal and does not depend on it; it is a general-benefit change I ran
into underneath it, and it should be reviewed on its own merits. It is the same
category as #4723, #4565 and #4768 — make the path `gc` walks on every
invocation cheaper.

## The observation

A repo cache directory is keyed on the normalized clone URL **with the subpath
stripped**. Every bundled source of a repository therefore resolves to *one*
directory, and `ValidateSyntheticRepo` checks every pack layout in that
directory regardless of which source asked.

Two helpers ask that question, and they ask the identical question:

- `requiredBuiltinSourcesUsable` / `ensureRequiredBuiltinSourcesCached`, once
  per required source (`core`, plus `bd` on a bd-provider city, plus `dolt` on
  a direct-exec lifecycle);
- `lockedBundledImportsUsable` / `ensureBundledLockedRemoteImportsCached`, once
  per bundled source pinned at its canonical commit in `packs.lock`.

The question is not cheap. `ValidateSyntheticRepo` `os.ReadFile`s every cached
pack file and compares it byte-for-byte against the copy embedded in the running
binary.

## The evidence

The claim is structural, and it is verifiable by reading the diff with no
timing at all:

> The number of `ValidateSyntheticRepo` calls per readiness pass drops from
> `len(requiredSources) + len(lockedBundledImports)` to the number of
> **distinct** `(cacheDir, commit)` pairs among them.

Worked through for the shapes `gc init` produces, on a warm valid cache:

| city | before | after |
| --- | ---: | ---: |
| bd-provider, no lockfile | 2 | 1 |
| bd-provider with `packs.lock` (`core`, `bd`) | 4 | 1 |
| `gc init --config gastown` (adds `[imports.gastown]`) | 5 | 2 |

The gastown row is the one worth checking, because it is where an absolute
"always 1" would be wrong. `GastownCityWithProviders` seeds
`[imports.gastown]` at `PublicGastownPackSource`
(`internal/config/public_packs.go`), whose clone URL is `gascity-packs.git` at
a different canonical pin — so it is a genuinely second cache directory, and
one walk of it survives.

Two honest qualifications on those numbers:

- **They describe a pass over a valid cache** — the ordinary case, and the one
  that runs on every `gc` invocation. When a cache actually needs repairing the
  saving is smaller: the repair path does not memoize anything after a
  successful `EnsureRepoInCache`, and `EnsureRepoInCache` re-validates under
  the write lock (`internal/packman/cache.go:73`), so a self-healing pass on
  the bd-provider shape still performs roughly 3 walks rather than 1.
- **No wall-clock number here is mine.** A downstream fork measured this class
  of change in the tens of milliseconds of CPU per invocation, but that was a
  different codebase whose baseline was already partly deduplicated, so I am
  not going to present it as a bound on what you would see. I would rather the
  structural argument carry this on its own.

## Behaviour, as BDD

**Given** a city whose required builtin sources and locked bundled imports all
resolve to the same synthetic cache directory, and that cache is valid
**When** a readiness pass runs
**Then** the directory is validated exactly once, and the pass reports ready.

**Given** the same city, and the cache is corrupted after the city was marked
ready
**When** the next readiness pass runs
**Then** the corruption is detected — the verdict from the previous pass is not
reused — and the cache is rehydrated from the embedded packs.

**Given** a readiness pass that has reached a negative verdict for a cache
directory and repaired it
**When** anything later in that same pass asks about the directory again
**Then** it is re-validated against the repaired tree, and reports valid.

## Why it is safe

The whole argument closes on one property:

> **Only positive verdicts are memoized.** No verdict about a broken cache is
> ever recorded, so there is nothing stale for a repair path to observe. A
> caller that repaired a cache necessarily obtained its negative verdict from a
> full validation, and the next question about that directory runs a full
> validation too.

That is why there is no invalidation API. There is nothing to invalidate.

The remaining properties fall out of scope rather than mechanism:

- **A verifier is scoped to one readiness pass.** It is a local in
  `EnsureBuiltinRuntimeAssets`, not package state. A verdict never outlives the
  pass that produced it, so every pass runs the full validator at least once and
  the self-healing contract runs fresh every time.
- **The degraded-path re-check uses a fresh verifier.** After the repair
  attempts, `EnsureBuiltinRuntimeAssets` re-asks `requiredBuiltinSourcesUsable`
  to decide between a hard error and a warning. That call deliberately builds a
  new verifier, so the decision cannot rest on anything decided before the
  repairs ran.
- **Map iteration order does not matter.** `ensureRequiredBuiltinSourcesCached`
  ranges over a map. Every entry in it resolves to the same
  `(cacheDir, commit)` memo key, so whichever entry happens to take the repair
  branch, the rest re-derive their verdict from the repaired tree. This is
  pinned by test, not just reasoned about.
- **It widens an existing lock-free window, and does not open a new one.**
  `Valid` reads without taking the repo-cache lock, exactly as the two helpers
  already did, so reusing a verdict stretches the gap between observing a cache
  and acting on it. `main` already races between its own successive lock-free
  reads; and because the cache directory name folds in the binary's pack
  content hash, any concurrent writer is writing byte-identical content. Worth
  naming rather than leaving for you to notice.
- **`TestEnsureBuiltinRuntimeAssetsRehydratesCorruptedCache` still passes.** That
  is the contract that a corrupted `gc-beads-bd` lifecycle script self-heals on
  the next pass, and it is the test I would have looked at first as a reviewer.
  I ran it explicitly.
- **What I deliberately did not do:** swap the full walk for a cheaper
  marker-only check. The marker cannot see per-file drift, and the whole-tree
  walk is exactly what repairs a corrupted cached script. That cost is the
  contract, not an oversight.

## Scope

| Area | Change |
| --- | --- |
| `cmd/gc/synthetic_cache_verifier.go` | New. `syntheticCacheVerifier` — 48 lines, one map, one method. |
| `cmd/gc/embed_builtin_packs.go` | `EnsureBuiltinRuntimeAssets` builds one verifier per pass and threads it through; `ensureRequiredBuiltinSourcesCached` and `requiredBuiltinSourcesUsable` take it. |
| `cmd/gc/legacy_pack_preflight.go` | `ensureBundledLockedRemoteImportsCached` and `lockedBundledImportsUsable` take it. |
| `cmd/gc/legacy_pack_preflight_test.go` | Call sites updated for the new parameter. Nothing else. |
| `cmd/gc/synthetic_cache_verifier_test.go` | New. Three tests. |
| Everything else | Untouched. No public API, no wire format, no config, no schema. |

## How the tests are built

The signature change means these tests cannot be run red-then-green against
`main` — they would not compile. So **each one was verified to fail against a
deliberately broken implementation** instead, which pins the behaviour more
tightly than red-green would have:

| Mutation | Test that catches it |
| --- | --- |
| memo removed entirely | `TestReadinessHelpersValidateTheSharedCacheOncePerPass` — "the duplicate walk is back" |
| memo made package-global (leaks across passes) | same test — "a new pass reused a verdict" |
| negative verdicts memoized (on the type) | `TestSyntheticCacheVerifierDoesNotMemoizeNegativeVerdicts` |
| negative verdicts memoized (through the real helper, across its own repair) | `TestEnsureRequiredBuiltinSourcesCachedRevalidatesAfterItsOwnRepair` |

The last two pin the same property at two different levels — the type in
isolation, and the production sequence in `ensureRequiredBuiltinSourcesCached`.
I originally described the second as pinning the repair path's re-validation,
which claimed more than it does, so: **what none of these catch** is a
hypothetical mutation that recorded a positive verdict straight from a
successful `EnsureRepoInCache` return instead of re-validating. That would
survive all three. I left it untested rather than contrive an injection point,
on the grounds that `EnsureRepoInCache` itself validates under the write lock
before returning, so such a memo would be weaker but not actually wrong.

`TestReadinessHelpersValidateTheSharedCacheOncePerPass` is the headline. It makes
two *opposed* assertions against a corrupted cache, so neither can pass
vacuously: with the **same** verifier the second helper must report the cache
usable (it must not re-walk what the first already validated in this pass), and
with a **fresh** verifier it must report the cache unusable (a new pass must see
the corruption). It also proves the two helpers really do share a directory — if
they did not, the shared-verifier assertion would fail rather than pass.

`TestEnsureRequiredBuiltinSourcesCachedRevalidatesAfterItsOwnRepair` walks the
production repair sequence through the real helper rather than exercising the
verifier in isolation. It also asserts up front that every required source
resolves to one cache directory, which is what makes its coverage of the map
iteration order complete rather than incidental — if the pack registry ever
changes so that it does not, the test says so instead of quietly narrowing.

## Notes for the reviewer

- `writeBundledSourceImportLock` in the new test file overlaps with the existing
  `writePreflightImportLock`, which hardcodes the gastown source. These tests
  need a lock entry that resolves to the *same* cache directory as the required
  builtin sources, so it takes the source as a parameter. Consolidating the two
  is a reasonable follow-up; I kept them separate here because I have another
  branch that changes `writePreflightImportLock`'s signature, and I did not want
  to couple the two reviews.
- This is deliberately standalone against `main` rather than stacked on my other
  open PRs, so it can be reviewed and landed independently.
- I have a separate change in preparation that scopes each of these caches to
  its own repository. It is not proposed yet, and it touches the same four
  `ValidateSyntheticRepo` call sites, so it will conflict textually with this
  one. The two are semantically independent — `cacheDir` already determines the
  repository, so the memo key needs no change — and rebasing it onto this is
  mechanical. Mentioning it only so the overlap is not a surprise later; nothing
  here depends on it.
- `go build ./...` and `go vet ./...` are clean.

### DIFF
--- a/cmd/gc/embed_builtin_packs.go
+++ b/cmd/gc/embed_builtin_packs.go
@@ -64,16 +64,20 @@ func EnsureBuiltinRuntimeAssets(cityPath string, warningWriter io.Writer) error 
 		pruneRetiredSystemPacks(cityPath, warningWriter)
 		return nil
 	}
-	if state.ready && requiredBuiltinSourcesUsable(cityPath) && lockedBundledImportsUsable(cityPath) {
+	// One verifier for the whole pass: the ready fast path and the repair path
+	// both validate the shared synthetic cache directory, and within a single
+	// pass that is the same question asked repeatedly.
+	verifier := newSyntheticCacheVerifier()
+	if state.ready && requiredBuiltinSourcesUsable(cityPath, verifier) && lockedBundledImportsUsable(cityPath, verifier) {
 		return nil
 	}
 	state.ready = false
 
 	var problems []error
-	if err := ensureBundledLockedRemoteImportsCached(cityPath); err != nil {
+	if err := ensureBundledLockedRemoteImportsCached(cityPath, verifier); err != nil {
 		problems = append(problems, err)
 	}
-	if err := ensureRequiredBuiltinSourcesCached(cityPath); err != nil {
+	if err := ensureRequiredBuiltinSourcesCached(cityPath, verifier); err != nil {
 		problems = append(problems, err)
 	}
 	if err := ensureGcBeadsBdShim(cityPath); err != nil {
@@ -82,7 +86,9 @@ func EnsureBuiltinRuntimeAssets(cityPath string, warningWriter io.Writer) error 
 	pruneRetiredSystemPacks(cityPath, warningWriter)
 
 	if len(problems) > 0 {
-		if !requiredBuiltinSourcesUsable(cityPath) {
+		// A fresh verifier: the repairs above just rewrote caches, so this
+		// last-resort check must not reuse anything decided before them.
+		if !requiredBuiltinSourcesUsable(cityPath, newSyntheticCacheVerifier()) {
 			state.lastWarning = ""
 			return fmt.Errorf("preparing builtin pack caches: %w", problems[0])
 		}
@@ -236,14 +242,17 @@ func builtinImportsForNames(names []string) (map[string]config.Import, []string)
 // required bundled sources at the canonical pin, independent of packs.lock,
 // so the stable shim target and pre-migration cities always have the
 // current binary's content available.
-func ensureRequiredBuiltinSourcesCached(cityPath string) error {
+// The verifier scopes cache validation to the calling readiness pass; every
+// required source of a repository shares one synthetic cache directory, so
+// without it the same directory is walked once per source.
+func ensureRequiredBuiltinSourcesCached(cityPath string, verifier *syntheticCacheVerifier) error {
 	commit := bundledPackImportCommit()
 	for name, source := range requiredBuiltinSources(cityPath) {
 		cachePath, err := packman.RepoCachePath(source, commit)
 		if err != nil {
 			return fmt.Errorf("resolving cache path for bundled %s pack: %w", name, err)
 		}
-		if builtinpacks.ValidateSyntheticRepo(cachePath, commit) == nil {
+		if verifier.Valid(cachePath, commit) {
 			continue
 		}
 		if _, err := packman.EnsureRepoInCache(cityPath, source, commit); err != nil {
@@ -253,14 +262,14 @@ func ensureRequiredBuiltinSourcesCached(cityPath string) error {
 	return nil
 }
 
-func requiredBuiltinSourcesUsable(cityPath string) bool {
+func requiredBuiltinSourcesUsable(cityPath string, verifier *syntheticCacheVerifier) bool {
 	commit := bundledPackImportCommit()
 	for _, source := range requiredBuiltinSources(cityPath) {
 		cachePath, err := packman.RepoCachePath(source, commit)
 		if err != nil {
 			return false
 		}
-		if builtinpacks.ValidateSyntheticRepo(cachePath, commit) != nil {
+		if !verifier.Valid(cachePath, commit) {
 			return false
 		}
 	}
--- a/cmd/gc/legacy_pack_preflight.go
+++ b/cmd/gc/legacy_pack_preflight.go
@@ -74,8 +74,9 @@ func lockedBundledCanonicalImports(cityPath string) ([]lockedBundledImport, erro
 // install". A cache that already validates is skipped lock-free; only on
 // validation failure does the preflight take the write-locked
 // packman.EnsureRepoInCache repair path, which revalidates under the lock
-// (a concurrent repair between the two checks is therefore benign).
-func ensureBundledLockedRemoteImportsCached(cityPath string) error {
+// (a concurrent repair between the two checks is therefore benign). The
+// verifier scopes that validation to the calling readiness pass.
+func ensureBundledLockedRemoteImportsCached(cityPath string, verifier *syntheticCacheVerifier) error {
 	imports, err := lockedBundledCanonicalImports(cityPath)
 	if err != nil {
 		return err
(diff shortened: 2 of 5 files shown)


quad341 (approved):
@wbern — this is a genuinely nice piece of work, and the part I want to call out isn't the dedup, it's that you closed the whole safety argument on one property: only positive verdicts are memoized, so there is nothing stale for a repair path to observe and therefore no invalidation API to get wrong. That's the difference between a memo I have to reason about nervously and one I can read once.

I checked the load-bearing claim rather than taking it from the description: `ValidateSyntheticRepo` takes no source and walks every layout in the directory, so `(cacheDir, commit)` really is the same question, and the fresh verifier on the degraded-path re-check is exactly the detail that would have made this unsafe if you'd missed it. `TestReadinessHelpersValidateTheSharedCacheOncePerPass` with its two opposed assertions is the test I'd have asked for.

Also appreciated: no borrowed benchmark number, and the honest note that a repairing pass saves less. Merging as-is. Thank you.


---


Review coverage:
- completed: `qwen codex`
- unavailable: `claude`

PR #4920, opened 2026-08-02
deps: bump beads v1.1.0 → main (eada50a0), and drop the CVE waivers it obsoletes

quad341:
gc vendored `beads v1.1.0` while the fleet's `bd` was built from `main` — **863 commits** apart — so the two disagreed about the shared store's schema and gc could not open its native store at all:

```
native_store_unavailable gate=native_open
  reason="schema version mismatch: database is at v62, binary knows up to v53 (9 migrations ahead)"
```

Every interactive `gc mail send/inbox/check/count` failed against a database `bd` had already migrated forward. Agent mail *delivery* and the hook read path (`gc hook run -- mail check --inject`) were unaffected — this broke the CLI surface only.

## Second commit: CVE waivers

The bump pulls `golang.org/x/net` past **v0.55.0**, which fixes CVE-2026-25680/25681/27136/39821/42502/42506. `TestTrivyIgnoreDropsGCModuleWaiversPastThreshold` caught this on `make test-fast-parallel` — it exists precisely to stop a waiver outliving its go.mod fix version, since that would leave the container scan unable to prove the gc binary clean.

Dropped only the `usr/local/bin/gc` path from those six entries. The `dolt`, `bd` and `kubectl` paths stay — external binaries still awaiting an upstream rebuild.

## Verification

- `go build ./...` — clean
- `go test ./internal/...` — 0 failures
- `go test ./cmd/...` — 0 failures
- `make test-fast-parallel` (the pre-push gate) — green

The 863-commit gap turned out API-compatible across gc's surface; no source changes, only `go.mod`/`go.sum` plus the waiver file.

## Context

Found while root-causing a fleet-wide beads-search outage. The other half — schema migrations 0059–0062 unable to complete because the dolt listener's `read_timeout_millis` bounds the *inter-row produce gap* and the recompute UPDATEs emit no rows for long stretches — is **be-rikgp**, fixed operationally by raising `read_timeout_millis` in `city.toml [dolt]`. All nine rig databases are now at main=62 / ignored=21 and `bd` search is healthy across all eight rigs.


bourgois:
@quad341 @sjarmak — flagging a deadline this PR sits directly in front of.

## Every waiver in `.trivyignore.yaml` expires 2026-08-07

On current `origin/main`: **44 entries, 28 unique CVEs, all stamped `expired_at: 2026-08-07`** — that's **Thursday, three days out**. Paths covered:

| Path | Waivers |
|---|---|
| `usr/local/bin/kubectl` | 17 |
| `usr/local/bin/dolt` | 17 |
| `usr/local/bin/bd` | 16 |
| `usr/local/bin/br` | 10 |
| `usr/local/bin/gc` | 6 |

The scan runs `--severity HIGH,CRITICAL --exit-code 1`, so when they lapse the container-scan gate goes red on **every PR**, not just ones touching images.

## This PR is the lever, and it's currently DIRTY

#4920 pulls `x/net` past v0.55.0, clearing CVE-2026-25680/25681/27136/39821/42502/42506 — six of the 28 — by an actual fix rather than a waiver, with `TestTrivyIgnoreDropsGCModuleWaiversPastThreshold` enforcing that the waivers can't outlive it. That's the right shape. But it's showing `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY` and last moved 08-03.

**Is it expected to land before Thursday?** If yes, that's six fewer and the rest is a smaller decision. If not, the horizon needs a call independently.

## What the remainder actually needs

Worth separating, because they aren't the same problem:

- **`kubectl` and `br` are external prebuilts with no fixed release available.** For CVE-2026-56852 specifically, `k8s.io/kubernetes` still pins `x/text v0.33.0` as of v1.36.3 — there is genuinely no version to bump to. A time-boxed waiver is the only correct answer there, not a shortcut.
- **`gh`, `dolt`, `bd` are rebuilt from source** in `Dockerfile.base`/`.agent`, and the file deliberately does **not** waive them for the Go-stdlib batch so a regression stays visible. That design is worth preserving through whatever happens Thursday — the risk in a rushed bulk extension is that it quietly widens to a rebuilt path.

There's precedent for extending (`e4d4f036e`, 07-07 → 08-07, #4002), and the file notes the durable fix is tracked in `ga-frh27v`.

## Offer

I'm not going to touch `.trivyignore.yaml` — extending 28 security waivers on the eve of expiry isn't a call for a contributor to make unilaterally, and the release-gate process in `release-gates/ga-85ytyr-...md` exists for exactly this. But if it's useful I'm happy to take the mechanical half: resolve this PR's conflicts, or work up the per-path "can this actually be fixed, or is a waiver the only option" table with current upstream release evidence so the decision is a short one. Say the word.

Raising it on your PR rather than opening a new issue since this is where the work already is.

bourgois:
Two follow-up findings while working the same area — both verified, both relevant to the Thursday horizon.

**1. `ga-frh27v` does not exist.** `.trivyignore.yaml` says *"Durable full fix tracked in ga-frh27v"*, but that ID resolves to nothing (`bd show ga-frh27v` errors; 0 of 820 beads match). So as far as this repo's tracker is concerned, the durable fix behind these waivers is currently unowned. Possibly it lives in a DB I don't share — worth confirming, because if not, the horizon has been extended twice against a tracker that was never created.

**2. This is the second bulk extension, not a one-off.** The file header records *"Expiry horizon bulk-extended 2026-07-06: 2026-07-07 → 2026-08-07"*, and that followed a 2026-06-12 → 2026-07-07 move. That makes it a rolling monthly deferral. The specific Thursday date matters less than the pattern — each extension is individually reasonable, and collectively they've carried 28 CVEs for three months.

Not raising this as criticism: for `kubectl` and `br` there is genuinely nothing to bump to, so *some* waiver is unavoidable and time-boxing is the right mechanism. But a tracker that doesn't exist means nobody is holding the other half.

Separately, and possibly useful to you here: I have a change in progress that clears CVE-2026-56852 (`x/text` < 0.39.0) on the rebuilt binaries by mirroring the existing `GRPC_VERSION` override pattern — an `ARG XTEXT_VERSION=0.39.0` plus a `go version -m` artifact assertion per binary, with source refs, versions and sha256 digests all unchanged. Built and verified at the pinned refs: gh 2.96.0, dolt and bd all embed x/text v0.39.0. It also **narrows** `.trivyignore.yaml` by five path-entries (drops `usr/local/bin/gc` from six entries, forced by `TestTrivyIgnoreDropsGCModuleWaiversPastThreshold` once x/net crosses 0.55.0) and waives zero rebuilt-from-source paths.

If that's useful to you rather than sitting on our fork, say the word and I'll open it upstream against your PR's area instead.

bourgois:
@quad341 — following up on the offer. **We've landed the x/text fix on our fork and verified it applies here.** Happy to open it upstream as its own PR if you want it; it's independent of this PR's beads bump.

## What it does

Clears **CVE-2026-56852** (`golang.org/x/text` < 0.39.0) across every image binary. Verified in CI: `Image vulnerabilities` **passes**, 73/0.

The mechanism is the one already in your Dockerfiles for grpc — **rebuild the same pinned release with a patched module**, not repin to a newer release:

```dockerfile
ARG XTEXT_VERSION=0.39.0
# ... go get golang.org/x/text@v, then assert:
# go version -m <artifact> | grep golang.org/x/text
```

So **no source ref, version or sha256 digest moves** — nothing gets coupled to gh/dolt release cadence. Verified by building all three at their pinned refs: gh 2.96.0, dolt and bd all embed x/text v0.39.0.

## It applies here, and I checked rather than assumed

Cherry-picked both commits onto a clean `origin/main` worktree just now:

- **Policy commit** (`.trivyignore.yaml`) — applies **clean, zero conflicts**
- **Pins commit** — clean except `go.mod`, and only mechanically (`golang.org/x/mod` moving between the direct and indirect require blocks). You'd regenerate it with the same `go get` anyway.
- `contrib/k8s/Dockerfile.base` is **byte-identical** on both sides, so the override applies verbatim.

Upstream carries the **same vulnerable pins** (`x/text v0.37.0`, `x/net v0.54.0`) and the **same six `usr/local/bin/gc` waivers** — so this isn't fork-specific.

## Net effect on the waiver file: it NARROWS

−5 path-entries. `usr/local/bin/gc` goes **6 → 0** (forced by `TestTrivyIgnoreDropsGCModuleWaiversPastThreshold` once x/net crosses 0.55.0 — your guard working as designed). The single addition is `kubectl`, which genuinely can't be fixed: `k8s.io/kubernetes` pins `x/text v0.33.0` through v1.36.3. **Zero rebuilt-from-source paths waived**, per the file's own policy.

## One interaction with this PR you'll want to know

Beads at `eada50a0` — your target here — **already pins x/text v0.39.0**. So once #4920 lands, the `bd` override becomes redundant *upstream* (harmless; the artifact assertion still passes). It stays required on our fork only because we pin bd at the time-boxed bridge commit `e97839a2`, whose x/text is 0.35.0.

Since both changes touch `.trivyignore.yaml` and `go.mod`, worth sequencing deliberately rather than discovering it in a conflict.

## Also, separately

A second layer surfaces once the Go binaries are clean — 4 HIGH in Python packages, spanning **three** of them: `gitpython` (2 advisories), `aiohttp` (CVE-2026-69244), `cryptography` (CVE-2026-69247). We bumped all three with the documented `uv pip compile` invocation and cross-checked all 167 regenerated hashes against PyPI. Included in the same offer if useful.

Say the word and I'll open it. And if resolving this PR's conflicts would help more than a new PR, I'm happy to take that instead — you said earlier you couldn't push to our fork, but the reverse direction works fine.

quad341:
Closing this one — not because the goal is wrong, but because the mechanism is one we've decided we won't use. Explaining properly, since the reasoning applies to anyone who hits this and shouldn't live only in our heads.

## The blocker

```
- github.com/steveyegge/beads v1.1.0
+ github.com/steveyegge/beads v1.1.1-0.20260802141721-eada50a0290b
```

That's a Go pseudo-version — a synthetic version for untagged commit `eada50a0290b` on beads `main`, not a release.

**gascity is public, and go.mod is transitive.** Anyone who depends on gascity inherits that pin: they end up transitively pinned to an upstream commit they never chose, can't easily reason about, and didn't sign up to track. We accept main-tracking risk deliberately for ourselves — that's our call to make about our own fleet. Exporting it to everyone downstream is a different act, and not ours to make on their behalf.

So the standing rule, from here on: **bump only to released (tagged) versions, and prefer the next release rather than jumping several.** If a fix we need exists only on an upstream `main`, the correct move is to get a release cut upstream — not to pseudo-version around it.

## The goal here was right

Worth being explicit, because this wasn't carelessness. `eada50a0290b` is the *same commit as the bd binary the fleet actually runs*. This was an attempt to make gascity's vendored beads agree with our deployed bd — i.e. a real attempt at the gc/bd version-skew problem (`ga-drlztz`), which is currently blocking a measured 159x query-shape fix from reaching gc (`ga-3xnm74`). That skew is a genuine, live, expensive problem and it does need solving.

It just has to be solved as: **get beads to tag a release containing what gc needs, then bump gascity to that tag.** "Upstream cuts a release" is a required step in that plan, not a nicety. That's the part worth carrying forward.

## One coupled change to redo together

The `.trivyignore.yaml` half drops gc's waivers for six x/net CVEs (`CVE-2026-25680/25681/27136/39821/42502/42506`), justified by go.mod having moved past `golang.org/x/net` v0.55.0. That justification depends entirely on the bump above — without it, `x/net` stays at v0.54.0 and those waivers must stay too, or Container Scan goes red on a claim that's no longer true.

So these two halves can't be separated: whenever the beads bump lands as a proper release, the waiver drop rides along with it in the same change. Splitting them in either direction breaks something.

## What to do next

Reopen this as a bump to a tagged beads release once one exists that carries what gc needs, with the `.trivyignore.yaml` change included. Tracking the rule as `gm-mkijn` and the underlying skew as `ga-drlztz`.

Thanks for chasing the skew — it's the right thing to be pulling on.

PR #4880, opened 2026-07-31
perf(builtinpacks): memoize embedded pack data and drop a redundant cache walk

wbern:
## The juicy parts

Every command that loads config pays the bundled-pack cache validation, and it was rebuilding two pure functions of embedded binary content on every call and walking the same tree about nine times. On `BenchmarkBuiltinReadinessPass` — a Go benchmark of that path, alternated across two compiled test binaries, 6 rounds x 20 iterations — this is 76.51 → 51.12 ms/op and 42,136 → 16,910 allocs/op. That is the benchmark's measurement of the readiness path, not an end-to-end `gc` wall clock, and per #4916 that fixture city has no `packs.lock`, so it covers only one of the pass's two halves.

**File content is still compared byte-for-byte.** The removed per-pack walk is strictly subsumed by the union walk that already ran, pinned by `TestValidateSyntheticRepoRejectsStrayFilesAnywhere` across five stray-file placements including both nesting cases. Worth knowing before weighing that: an earlier revision of this PR replaced content comparison with an mtime check for a bigger win, CI caught that it was unsound on Linux, and it was removed. What is left carries no stat-based assumption. This removes *different* redundant work from the same invocation than #4565, #4723 or #4768 — it does not replace any of them.

The bundled pack cache validation that runs before every config load rebuilt two
pure functions of embedded content on every call, and walked the same tree about
nine times.

## Changes

1. **Memoize the allowed-path sets and the per-pack manifests.** Both derive
   entirely from content embedded in the running binary, so they cannot change
   within a process. This follows the `syntheticContentHashOnce` precedent already
   in this file. `materializeFS` keeps the uncached `manifestForFS` because it is
   the write path.
2. **Drop `validatePackFiles`' own directory walk.** `validateSyntheticRepoFileSet`
   already walks the whole cache once against the union of every layout's
   manifest, and that union check strictly subsumes a per-pack one: a file
   unexpected for its own pack is absent from the union too.

**File content is still compared byte-for-byte.** Every integrity property the
validation had before is unchanged — this is purely about not redoing work.

## Measured

`BenchmarkBuiltinReadinessPass` (added by #4723) measures exactly this path. Two
compiled test binaries alternated, 6 rounds x 20 iterations:

| | before | after | change |
|---|---|---|---|
| time | 76.51 ms/op | 51.12 ms/op | **-33.2%** |
| memory | 22,490,590 B/op | 8,504,894 B/op | **-62.2%** |
[...]

PR #4921, opened 2026-08-02
perf(builtinpacks): scope each synthetic pack cache to its own repository

wbern:
## The juicy parts

Every `gc` command runs the builtin readiness pass, and today that pass byte-compares cached pack files belonging to a repository the cache can never serve. Scoping the cache removes that work: the `gascity.git` cache drops from 491 files to 228, and the remainder moves to its own `gascity-packs.git` cache — `228 + 264 − 1 = 491`, so nothing is trimmed heuristically.

On one darwin/arm64 machine, warm `gc version` against the same populated `GC_HOME` went 216/207/219 ms on `main` to 158/168/163 ms here — roughly 25% off warm startup. That is three runs per arm on one box: a shape, not a benchmark. Detection gets *stricter* rather than looser, because a narrower allowed set rejects files it used to tolerate.

Two things to weigh against that. Folding the marker schema into the cache key is what stops two binary generations re-materializing over each other under the exclusive repo-cache lock. And the price is exactly one orphaned cache directory per machine (~3.7 MB), reclaimed by `gc import prune --apply` — disclosed below rather than left to be discovered.

## What this is

A synthetic pack cache directory now holds — and is validated against — only the
layouts of the repository it belongs to.

## Why I was looking at this

that proposal and does not depend on it; it is a general-benefit change I ran
into underneath it and it should be reviewed on its own merits.

## The observation

A cache directory is keyed on the normalized clone URL **with the subpath
stripped**, and imports resolve to `<cache>/<subpath>` rather than the cache
root. So a cache built for `gascity.git` can only ever serve `gascity.git`
subpaths.

It was nevertheless materialized with *every* repository's layouts, and every one
of those was byte-compared against the embedded copy on each readiness pass —
`os.ReadFile` per cached file.

## The measurement

Numbers below are from one machine (darwin/arm64), built `gc` binaries, this
branch vs `main`, one run per data point except where noted. Not a rigorous
benchmark — enough to show the shape, not enough to quote to three digits.

**Files materialized** (`gc doctor` against a fresh `GC_HOME`, one cache dir):

| cache | files |
| --- | ---: |
| `gascity.git` — before | 491 |
| `gascity.git` — after | **228** |
| `gascity-packs.git` — after | 264 |

The arithmetic is the useful part: `228 + 264 − 1 = 491`. The `− 1` is the
marker: the old single cache had one, the two scoped caches have one each, so
the split adds exactly one file back. The rest of the difference is exactly the
other repository's layouts and nothing else — not a heuristic trim.

(A `gc doctor` probe on a fresh `GC_HOME` reports 493 / 230 rather than
491 / 228; the extra two are the probe's own city scaffolding, not cache
contents.)

**Warm startup** (`gc version`, same populated `GC_HOME`). Re-measured after
#4880 merged, against `main` **including** #4880, because #4880 already removed
several traversals per call and I did not want to claim any of its win as this
PR's. Interleaved — arms alternate every round, 10 invocations per data point:

| round | `main` (incl. #4880) | this branch |
| ---: | ---: | ---: |
| 1 | 178.0 ms | **117.4 ms** |
| 2 | 153.6 ms | **111.5 ms** |
| 3 | 150.3 ms | **118.3 ms** |
| 4 | 157.6 ms | **112.4 ms** |
| 5 | 147.9 ms | **107.2 ms** |
| 6 | 144.6 ms | **111.9 ms** |

Medians ≈ 152 ms → 112 ms, so roughly 25% off warm startup, and every round
favours the branch. That is the readiness pass no longer byte-comparing the other
repository's 263 files. One machine (darwin/arm64) — the effect is large and
consistent enough to report, but it is one box, not a fleet.

**Files rewritten during those warm runs: 0 on both arms.** That is the answer
to "does the schema bump make this thrash" — it does not. The bump costs exactly
one re-materialization, once, and steady state is unchanged.

## Detection gets stronger, not weaker

Worth saying plainly, because a smaller allowed set reads like a looser check.

Scoping only ever **narrows** the set of paths a cache may contain, and a
narrower set produces *more* rejections. A layout belonging to the other
repository used to be tolerated inside this cache; it is now an unexpected file
and fails validation.

## The marker schema is folded into the cache key

This is the part I would most want a reviewer to check, because leaving it out
is actively harmful rather than merely incomplete.

Without it, an old and a new binary resolve to the **same** machine-global
directory, each rejects the other's marker, and each re-materializes the whole
tree on every invocation — in both directions, for as long as two generations are
in use. And each of those re-materializations takes the **exclusive** repo-cache
lock, which covers the whole cache root, so every other `gc` on the box waits
behind a rewrite it did not need. That is the failure mode; I did not try to put
a clean number on it, because the interesting cost is contention rather than
single-command latency.

`SyntheticCacheKeyComponent` already existed for exactly this failure — the
citywide wedge that recurs whenever a deploy leaves two binary versions running
side by side — but it was bound only to embedded *content*. A schema change
alters the expected *layout* without altering content, so content alone cannot
separate them. The schema is now part of the key, and the two binaries simply use
different directories.

Schema 1 markers are rejected rather than migrated. That is the ordinary
self-heal path, not a migration step.

### The cost: one orphaned directory per machine

Stating this rather than letting a reviewer find it, because "invalidates every
cache on every machine" is the headline risk of this PR.

Because the schema is folded into the cache key, a schema-2 binary resolves a
**new** directory name and never touches the schema-1 one. Nothing deletes it.
Every machine that upgrades gains one orphaned cache directory (~3.7 MB here).

Reclamation is manual: `gc import prune --apply`. The old directory's name is not
in any `packs.lock` under the new key, so it classifies as unreferenced, and it is
removed once it is older than `--keep-days` (default 7).

This is acceptable and precedented — any change to the embedded pack content does
exactly the same thing today, since the content hash is already in the key — but
it is a real disclosure and it should be in the PR, not discovered.

## Two correctness details

- **The caller supplies the repository and it is cross-checked against the
  marker**, mirroring how `commit` is already handled. Trusting the marker alone
  would let a cache built for the wrong repository validate clean and then fail
  at import resolution.
- **Both validators reject an unknown repository.** Without that the layout set
  collapses to empty, and a directory holding nothing but a well-formed marker
  validates clean.

Both of these came out of adversarial review of an earlier draft, which
reproduced the empty-layout-set hole.

## Notes for the reviewer

- Test-side changes are mostly the `ValidateSyntheticRepo` /
  `MaterializeSyntheticRepo` signature gaining `repository`.
- `TestValidateSyntheticRepoRejectsStrayFilesAnywhere` covers stray files at
  every depth under the **scoped** allowed set, including nested layouts
  (`examples/bd` contains `examples/bd/dolt`) — the case where flattening several
  manifests into one allowed set could conceivably admit a path no single
  manifest claims.
- **`validatePackFiles` no longer walks its own subtree** — #4880 removed that
  walk, because the whole-tree walk in `validateSyntheticRepoFileSet` already
  subsumed it. That whole-tree walk is what this PR scopes, and scoping only
  narrows the union it checks against, so the subsumption still holds:
  `ValidateSyntheticRepo` calls `validatePackFiles` for exactly the layouts the
  union is built from. (An earlier revision of this description said the opposite;
  it was written before #4880 landed.)
- The new `RepositoryForSource` call sites add about five `if !ok { return err }`
  branches that are unreachable today: every one of them is already gated
  upstream by `IsSource` / `IsBundledSourceAtCanonicalPin`, so the source is
  known to be bundled before the helper is asked. They are defense in depth
  against a future caller, not a fix for a reachable case — flagging it so nobody
  goes hunting for the bug they are guarding.
- This is deliberately standalone against `main` rather than stacked on my other
  open PRs, so it can be reviewed and landed independently.

### DIFF
--- a/cmd/gc/embed_builtin_packs.go
+++ b/cmd/gc/embed_builtin_packs.go
@@ -252,7 +252,11 @@ func ensureRequiredBuiltinSourcesCached(cityPath string, verifier *syntheticCach
 		if err != nil {
 			return fmt.Errorf("resolving cache path for bundled %s pack: %w", name, err)
 		}
-		if verifier.Valid(cachePath, commit) {
+		repository, known := builtinpacks.RepositoryForSource(source)
+		if !known {
+			return fmt.Errorf("resolving bundled repository for %s pack source %q", name, source)
+		}
+		if verifier.Valid(cachePath, repository, commit) {
 			continue
 		}
 		if _, err := packman.EnsureRepoInCache(cityPath, source, commit); err != nil {
@@ -269,7 +273,8 @@ func requiredBuiltinSourcesUsable(cityPath string, verifier *syntheticCacheVerif
 		if err != nil {
 			return false
 		}
-		if !verifier.Valid(cachePath, commit) {
+		repository, known := builtinpacks.RepositoryForSource(source)
+		if !known || !verifier.Valid(cachePath, repository, commit) {
 			return false
 		}
 	}
--- a/cmd/gc/embed_builtin_packs_test.go
+++ b/cmd/gc/embed_builtin_packs_test.go
@@ -681,7 +681,7 @@ func TestEnsureBuiltinRuntimeAssetsHydratesCacheAndShim(t *testing.T) {
 		if err != nil {
 			t.Fatalf("RepoCachePath(%s): %v", name, err)
 		}
-		if err := builtinpacks.ValidateSyntheticRepo(cachePath, commit); err != nil {
+		if err := builtinpacks.ValidateSyntheticRepo(cachePath, builtinpacks.Repository, commit); err != nil {
 			t.Errorf("%s cache invalid after hydration: %v", name, err)
 		}
 	}
@@ -739,7 +739,7 @@ func TestEnsureBuiltinRuntimeAssetsSkipsShimForNonBdCity(t *testing.T) {
 	if err != nil {
 		t.Fatalf("RepoCachePath(core): %v", err)
 	}
-	if err := builtinpacks.ValidateSyntheticRepo(coreCache, commit); err != nil {
+	if err := builtinpacks.ValidateSyntheticRepo(coreCache, builtinpacks.Repository, commit); err != nil {
 		t.Errorf("core cache invalid after hydration: %v", err)
 	}
 }
@@ -1049,7 +1049,7 @@ func TestEnsureBuiltinRuntimeAssetsRehydratesEvictedOptionalLockedBundledCache(t
 	if err != nil {
 		t.Fatalf("RepoCachePath(gastown): %v", err)
 	}
-	if err := builtinpacks.ValidateSyntheticRepo(cachePath, commit); err != nil {
+	if err := builtinpacks.ValidateSyntheticRepo(cachePath, builtinpacks.PublicRepository, commit); err != nil {
 		t.Fatalf("gastown cache invalid after ready: %v", err)
 	}
 
@@ -1065,7 +1065,7 @@ func TestEnsureBuiltinRuntimeAssetsRehydratesEvictedOptionalLockedBundledCache(t
 		t.Fatalf("EnsureBuiltinRuntimeAssets after optional cache eviction: %v", err)
 	}
 
-	if err := builtinpacks.ValidateSyntheticRepo(cachePath, commit); err != nil {
+	if err := builtinpacks.ValidateSyntheticRepo(cachePath, builtinpacks.PublicRepository, commit); err != nil {
 		t.Fatalf("optional locked bundled cache not rehydrated after ready fast path: %v", err)
 	}
 }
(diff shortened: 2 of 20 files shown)


quad341 (approved):
Hi @wbern — this is a genuinely nice find. The observation that a cache keyed on the clone URL *with the subpath stripped* can only ever serve that repository's subpaths, so the other repository's 263 layouts were being written to disk and then byte-compared on every readiness pass, is the kind of thing that hides in plain sight for a long time.

Two things I want to call out specifically. `TestValidateSyntheticRepoRejectsCrossRepositoryStray` is the test that made this easy to accept — it distinguishes "correctly scoped" from "quietly stopped checking," which a narrower allowed set could otherwise disguise. And folding the marker schema into the cache key is the right instinct: without it two binary generations fight over one directory, each re-materialization taking the exclusive cache-root lock. Disclosing the orphaned ~3.7 MB directory up front rather than letting a reviewer find it is appreciated too.

The branch has drifted from `main` in the two weeks since CI ran. I'll take the rebase on my side, add one small test pinning `RepositoryForSource` against the tree-URL spelling `packs.lock` actually stores, and re-run the suite before merging. Nothing needed from you. Thanks for this.


---
**Reviewers**: Qwen (local) — ok · Claude — **failed** · Codex — **failed**


Review coverage:
- completed: `qwen`
- unavailable: `claude codex`

PR #4923, opened 2026-08-02
Split whitespace-joined mail message IDs

swedeinasia-flow:
## What changed

- split every `gc mail archive` and `gc mail delete` positional argument with `strings.Fields` before single-versus-batch dispatch
- reject whitespace-only arguments as a missing message ID
- preserve normalized IDs in JSON delete output
- add archive/delete regressions and an Unreleased changelog entry

## Why

Some shells can preserve a variable containing multiple message IDs as one positional argument. The CLI previously sent that joined value down the single-ID path. A lookup miss is intentionally idempotent for archive/delete, so the command could report the whole joined string as already handled, exit successfully, and leave every message untouched.

## User impact

Archive and delete now handle whitespace-joined ID variables the same way as separately supplied IDs, in input order. Existing single-ID and multi-argument behavior is unchanged.

## Validation

- RED: the joined delete regression reproduced the silent `Already deleted` result, recorded no events, and left all messages open
- PASS: `CGO_ENABLED=0 go test -count=1 -run '^TestMail(Archive|Delete)' ./cmd/gc`
- PASS: repository pre-commit hook (format, changed-package lint, generated-contract drift, vet, and docs sync)
- PARTIAL: `make check` passed formatting, golangci-lint v2.12.0 (`0 issues`), vet, release-dist checks, and the routed-test matrix; the aggregate unit sweep then hit an unrelated existing failure in `internal/runtime/exec.TestProvider_StartCancellationInterruptsForegroundChild` because its rollback marker was absent. The full `cmd/gc` package passed.

Local Go tests used `CGO_ENABLED=0` because the repository documents a macOS-only build incompatibility with Homebrew ICU 78; upstream CI runs the native dependency path on Linux.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Mail archive and delete now expand whitespace-joined message IDs.** Each
+  positional argument is split into individual IDs before single-versus-batch
+  dispatch, so shell variables containing multiple IDs no longer look like one
+  already-handled message.
+
 - **`gc import add` of a local in-git pack now locks to HEAD, not the repo's
   latest tag.** Per `gc import add --help`, a local path inside a git
   worktree is documented to be "locked to the current commit," but the
--- a/cmd/gc/cmd_mail.go
+++ b/cmd/gc/cmd_mail.go
@@ -349,7 +349,19 @@ func doMailArchiveSelectedJSON(mp mail.Provider, rec events.Recorder, args []str
 	return exit
 }
 
+// splitMessageIDArgs splits every argument on whitespace and drops empty
+// tokens. Message IDs never contain whitespace, and some shells can preserve a
+// variable containing multiple IDs as one argument.
+func splitMessageIDArgs(args []string) []string {
+	ids := make([]string, 0, len(args))
+	for _, arg := range args {
+		ids = append(ids, strings.Fields(arg)...)
+	}
+	return ids
+}
+
 func doMailArchiveJSON(mp mail.Provider, rec events.Recorder, args []string, jsonOut bool, stdout, stderr io.Writer) int {
+	args = splitMessageIDArgs(args)
 	if len(args) < 1 {
 		fmt.Fprintln(stderr, "gc mail archive: missing message ID") //nolint:errcheck // best-effort stderr
 		return 1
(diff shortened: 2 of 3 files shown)


swedeinasia-flow:
Rebased the existing contributor branch onto current `upstream/main` and force-updated it under the exact prior-head lease.

- New head: `51ae08dd13219b4f8dbc91b2e61a7870173f0001`
- Base: `655713ff4559c335ff17ec09c4a65f5cca2a5f27`
- Scope remains limited to `CHANGELOG.md`, `cmd/gc/cmd_mail.go`, and `cmd/gc/cmd_mail_test.go`.

Validation:

- RED on pristine current main: the four added whitespace-ID regressions all failed. Joined IDs followed the single-ID idempotent path (for example, `Already deleted gc-1 gc-2\tgc-3`), left the individual messages open, and whitespace-only input was treated as a real ID.
- PASS after patch: `CGO_ENABLED=0 go test -v -count=1 -run '^TestMail(DeleteWhitespaceJoinedIDs|DeleteWhitespaceJoinedIDsJSON|ArchiveAndDeleteWhitespaceOnlyArg|ArchiveWhitespaceJoinedIDs)$' ./cmd/gc`
- PASS: all six `cmd/gc` shards in `make test-fast-parallel`.
- PASS: `make check-docs` and `git diff --check upstream/main...HEAD`.
- PARTIAL, current-main macOS baseline: `make check` passed format, lint (`0 issues`), vet, release-dist, and routed/split-topology checks, then the source-identical residency self-test required associative arrays unavailable in macOS Bash 3.2 (`declare -A: invalid option`).
- PARTIAL, current-main macOS baseline: the remaining `unit-core` shard reported source-identical failures in `internal/pathdurability` (`/private/var` vs `/var` classification), `internal/pidutil` (unreadable argv fail-closed), and `internal/worker/adapters/zcode` (Bash 3.2 `more: unbound variable`, signal permission, path aliasing, and cascading timeouts). The patch touches none of those packages.

— Codex, on behalf of @swedeinasia-flow

quad341 (approved):
@swedeinasia-flow — thank you for this one, and specifically for finding the *silent* failure rather than a loud one. A lookup miss folding into `ErrAlreadyArchived` meant the joined-ID case printed `Already archived gc-1 gc-2 gc-3`, exited 0, and left every message sitting in the inbox. That's the worst shape a bug can take, and it survived this long precisely because the exit code lied about it.

Your tests are why this merges as-is. Asserting the final bead status — not just stdout — is the check that separates "reported success" from "did the work," and using `\t` on the delete side and `\n` on the archive side covers the whitespace forms a naive fix would miss. Landing the split at the two dispatch points keeps the single-ID path byte-identical, which the surrounding comments and the original batch-archive gate both promise.

One note for the record: whitespace-only args now exit 1 instead of 0. That's the right answer, and I checked the scripted callers in-tree — the only one uses the `--to`/`--subject-prefix` filter form, so nothing regresses.

Merging. Nice work.


---

PR #4924, opened 2026-08-02
perf(builtinpacks): validate the pack cache in one traversal, and detect deletions

wbern:
## The juicy parts

Every `gc` invocation validates the bundled pack cache, and that check is what
repairs a corrupted cached file. The measurable win is small and worth stating
plainly: **−5.1% allocations and −8.9% bytes, with no resolvable change in
time**. This is now rebased onto current `main`, so those are the numbers
against `main` as it stands — there is no longer a second, flattering table
measured against a tree without #4880 in it.

What earns the change is the test, not the saving. Deleting a file from the cache
was not covered anywhere in the suite. `main` does detect it — I want to be clear
this is not a bug fix — but nothing pinned that behaviour, and fusing two
traversals into one is exactly the refactor that could quietly lose it. The new
test passes on `main` today and fails against a fused walk without the seen
counter, so it guards a real property rather than describing the implementation.

## What this is

`ValidateSyntheticRepo` walked the materialized pack cache twice: a
`filepath.WalkDir` to check the file *set*, then a manifest-driven `os.Lstat` +
`os.ReadFile` per expected file in `validatePackFiles` to check its *content*.
Both resolved the same paths. They are now one walk that enforces both halves.

But the reason I am opening this is not the saving. It is that fusing the two
naively **loses deletion detection**, and there was no test standing in the way.

## The part that actually matters

The two passes are not symmetric in what they can miss.

- A manifest-driven `Lstat` per expected file **cannot** miss a deletion. It asks
  about every path by name.
- A `WalkDir` **can**. A walk never visits a path that is not there. A walk that
  only rejects *unexpected* paths accepts a cache with files removed.

That is the worst direction for a cache-integrity check to fail. This validation
is what gates rehydration — if it returns "valid" for a cache that is missing
files, the self-heal never fires and the missing file just stays missing.

So the fused walk carries an explicit count of expected files seen, and rejects a
short cache by name:

```
validating bundled pack cache file set: bundled pack cache is missing expected
file internal/bootstrap/packs/core/pack.toml (489 of 490 present)
```

I want to be precise about the status quo, because it would be easy to overclaim
here: **`main` today detects deletions correctly.** The `Lstat` loop catches
them. What `main` does not have is a *test* for it — the property was covered
only implicitly, by a loop this change removes.
`TestValidateSyntheticRepoRejectsDeletedFile` now covers it directly, and I would
argue it is worth having whether or not you take the rest of this patch.

Its evidence, in order:

| version | result |
| --- | --- |
| `main`, unmodified | **passes** — the `Lstat` loop rejects the deletion |
| fused walk with no seen count | **fails** — `a cache missing internal/bootstrap/packs/core/pack.toml validated clean; deletion is undetected` |
| fused walk as proposed | **passes** |

I wrote the test first and ran it against all three, in that order, and re-ran all
three after the rebase onto current `main`. The middle row is the whole point of
the test: it is the version this PR would have been if I had not gone looking.
Dropping the seen count from the shipped implementation fails
`TestValidateSyntheticRepoRejectsDeletedFile` — both subtests — and nothing else
in the package.

## A second gap this turned up

While checking that I had not *claimed* coverage I did not have, I deleted the
unexpected-directory bookkeeping from the walk and ran the package. **It stayed
green.** Nothing anywhere asserted that a cache containing a directory the
manifests do not imply is rejected.

That gap predates this PR — the check has always been there, and still is,
untouched. But it is exactly the kind of thing this PR is meant to be about, so
`TestValidateSyntheticRepoRejectsUnexpectedDirectory` is included. Under that
same deletion it is still, after the rebase, the **only** failing test in the
package:

```
--- FAIL: TestValidateSyntheticRepoRejectsUnexpectedDirectory/at_the_cache_root
    ValidateSyntheticRepo accepted unexpected directory zzz-empty
--- FAIL: TestValidateSyntheticRepoRejectsUnexpectedDirectory/inside_an_expected_pack_directory
    ValidateSyntheticRepo accepted unexpected directory internal/bootstrap/packs/core/zzz-empty
```

The fixture directories are empty on purpose. A directory with a file in it is
already rejected for containing an unexpected *file*, which would let the test
pass without the directory check existing at all.

## Which of the new tests stand on their own

Four of the five new tests pass against `main` unmodified, which is the cheapest
way for you to confirm they describe behaviour rather than implementation:

| test | on `main`, unmodified |
| --- | --- |
| `TestValidateSyntheticRepoRejectsDeletedFile` (2 cases) | passes |
| `TestValidateSyntheticRepoRejectsUnexpectedDirectory` (2 cases) | passes |
| `TestValidateSyntheticRepoAcceptsAFreshlyMaterializedCache` | passes |
| `TestValidateSyntheticRepoRejectsSameLengthTamper` | passes |
| `TestNoBundledPackEmbedsTheCacheMarkerName` | **fails, correctly** |

The last one is not separable, and I would rather explain that than let it look
like an oversight. On `main` the allowed-file set is *seeded* with the marker
path, so the marker is a member of it no matter what the packs embed, and an
assertion that it is absent says nothing. This change stops seeding it — the set
is now derived from the pack manifests alone — which is what makes "no bundled
pack contributes a file at the marker path" an assertion with content. The test
is a guard on the fusion's own assumption and belongs with it.

## Why I was looking at this

[#4441](https://github.com/gastownhall/gascity/issues/4441). This is **not** part
of that proposal and does not depend on it. It is a general-benefit change I ran
into underneath it, and it should be reviewed entirely on its own merits — same
category as #4565, #4723 and #4880: make the path `gc` walks on every invocation
cheaper. Every command that loads config pays this validation.

## Measured

`BenchmarkValidateSyntheticRepo` is added by this PR and measures the traversal
directly. That is deliberate: the readiness-pass benchmark in `cmd/gc` cannot
resolve a change this size, because the validation is only a fraction of that
pass. A change to the traversal has to be measured against the traversal.

Two compiled test binaries, **arms interleaved and their order swapped every
round** — 54 paired rounds, `-benchmem`, darwin/arm64, against current `main`
(which now contains #4880):

| | `main` | + this PR | change |
| --- | ---: | ---: | ---: |
| memory | 3,991,962 B/op | 3,638,121 B/op | **−8.9%** |
| allocations | 8,345 | 7,918 | **−5.1%** |
| time | — | — | **no resolvable difference** |

The byte and allocation columns are the result; the time column is not a result
and I am not claiming a speedup. Paired branch/main time ratios over those 54
rounds have a median of 0.96 with the branch ahead in 31 of 54 — but the spread
runs from p10 0.47 to p90 1.38, which is an order of magnitude wider than the
effect being looked for. The dominant cost is reading the cached files, which is
unchanged; only the redundant path resolution goes away.

Two notes on how much weight these numbers carry:

- The **byte and allocation counts are stable to the digit** across every rerun
  and every machine load. The **times are not**, and the failure mode is worse
  than noise: a first sweep of 12 paired rounds taken while the machine was at
  load 74 reported the branch **26% slower** at the median, which the 54-round
  sweep at load 31–45 does not reproduce in either direction. That is why the
  arms are interleaved and their order swapped — block-ordering on this
  benchmark produces confident wrong answers. Please read the memory columns as
  the result and the time column as "not measurable here".
- Allocation *volume* drops and **nothing extra is retained**. The expected-path
  map now holds each file's `fileEntry`, but `data` there aliases the byte slice
  #4880's per-pack manifest cache already keeps for the process lifetime, so the
  map costs its own entries and no embedded content.

## Relationship to #4880

**#4880 has merged** (`38ed358fa`), and this branch is rebased onto it. What was
previously a sequencing question is settled: the numbers above are the increment
this change contributes on top of #4880, because #4880 *is* `main` now.

The rebase cost one mechanical change and no design change: the allowed-path set
#4880 memoizes is retyped from `map[string]struct{}` to
`map[string]expectedFile` so the fused walk can compare content from it.
Everything else composes untouched — `manifestForPack`, the `sync.OnceValue`
memoization, and #4880's tests all survive as written.

**The diff is smaller than it was**, because part of it is now #4880's work
rather than mine. #4880 already removed `validatePackFiles`' own
`filepath.WalkDir` over the tree it was handed (21 of the lines the earlier
revision of this PR deleted are simply not there any more). What this PR removes
is what is left of `validatePackFiles`: the per-file `Lstat` + `ReadFile` loop,
absorbed into the single walk. I am not claiming credit for the walk removal.

| | #4880 (merged) | this PR |
| --- | --- | --- |
| memoize per-pack manifests and the allowed-path sets | yes | retypes the file set to carry content; memoization untouched |
| stop `validatePackFiles` re-walking the tree it was handed | **done there** | not touched — already gone |
| remove the duplicate path resolution between the surviving walk and the `Lstat` loop | no | yes — `validatePackFiles` is deleted outright |
| add a deletion-detection test, and an unexpected-directory test | no | yes |

## Behaviour, as BDD

**Given** a cache that a bundled-pack materialization has just written
**When** it is validated
**Then** it is accepted, and every file in it has been compared byte-for-byte
against the copy embedded in the running binary.

**Given** a cache with one expected file deleted, or a whole pack subtree deleted
**When** it is validated
**Then** it is rejected, and the error names the missing path relative to the
cache root together with how many of the expected files were present.

**Given** a cache with one byte flipped in place, leaving the file length
unchanged
**When** it is validated
**Then** it is rejected as content drift — the comparison is over bytes, not
sizes or timestamps.

**Given** a cache with an extra file, an extra empty directory, or a symlink
**When** it is validated
**Then** it is rejected — with a message byte-identical to the one `main`
produces, which is measured below rather than assumed.

## Why it is safe

Every check that ran before still runs, and — apart from the deliberate
improvement to the missing-file message — reports itself in the same words.

| check | before | after | message |
| --- | --- | --- | --- |
| file content | `bytes.Equal` in `validatePackFiles` | `bytes.Equal` in `checkCachedFile`, called from the walk | identical |
| file mode | `os.Lstat` in `validatePackFiles` | the walk's own `fs.DirEntry.Info()` — same `lstat` semantics, on the path the walk already produced | identical |
| unexpected file | union walk | union walk | identical |
| unexpected directory | union walk | union walk | identical |
| symlink | union walk | union walk | identical |
| **missing file** | `os.Lstat` per expected file | **seen count** | deliberately changed |

I measured that rather than asserting it, because the first draft of this patch
got it wrong. Moving the per-file checks inside the walk closure had them picking
up the walk's own `"validating bundled pack cache file set: %w"` wrapper, so a
*content-drift* failure was reported as a *file-set* failure. Nothing consumes
these strings, so nothing broke — but the table above would have been a false
claim. The fix is that `checkCachedFile`'s error is recorded and returned
unwrapped after the walk instead of being returned from the callback. Measured
output, current `main` against this branch, both driven through
`ValidateSyntheticRepo` on the same fixtures:

| case | `main` | this PR |
| --- | --- | --- |
| content | `bundled pack cache "core" file pack.toml content differs from current binary` | identical |
| mode | `bundled pack cache "core" file pack.toml has mode -rw-------, expected -rw-r--r--` | identical |
| unexpected file | `validating bundled pack cache file set: bundled pack cache contains unexpected file scratch.txt` | identical |
| unexpected directory | `validating bundled pack cache file set: bundled pack cache contains unexpected directory zzz-empty` | identical |
| symlink | `validating bundled pack cache file set: bundled pack cache contains symlink link.toml` | identical |
| missing file | `checking bundled pack cache "core" file pack.toml: lstat <abs path>: no such file or directory` | `validating bundled pack cache file set: bundled pack cache is missing expected file internal/bootstrap/packs/core/pack.toml (489 of 490 present)` |

Recording the per-file error and continuing the walk, rather than aborting on it,
also preserves the old **precedence**: `main` completed the entire file-set walk
before comparing any content, so an unexpected path outranked content drift. It
still does, whichever the fused walk reaches first — measured on three
double-fault fixtures (unexpected file + drift, unexpected directory + drift,
symlink + drift), all three of which produce the same message on both trees.

Two divergences inside the deliberately-changed missing-file class, which I would
rather name than have found:

- **A cache that is both short *and* drifted** reports the drift here, where
  `main` may report either. `main`'s answer depends on which layout it reaches
  first: on a fixture with the deletion in `core` and the drift in `gastown` it
  names the deletion, and on the mirrored fixture it names the drift. There is no
  stable precedence there to preserve.
- **`main`'s missing-file message is not deterministic** — it names whichever
  expected file the manifest's map iteration reached first, and two runs of the
  identical fixture named two different files. `firstMissingExpectedFile` sorts,
  so the new message names the same path every time.

Points I would want a reviewer to push on, so I will raise them myself:

- **The seen count cannot over-count.** A `WalkDir` visits each path once, so
  `seen` can only reach `len(allowedFiles)` if every expected path was visited.
  There is no path that increments it twice.
- **A directory sitting where a file is expected** does not slip through. It
  takes the directory branch, is not in the allowed-directory set, and is
  reported as an unexpected directory — which is checked *before* the shortfall,
  so the more specific error wins.
- **The marker file is now skipped explicitly** rather than living in the allowed
  set, because its content is per-cache rather than embedded and so has no
  expected bytes to compare. Skipping it would be wrong if a bundled pack also
  contributed a file at that path — and it would also silently under-count
  `seen` against `len(allowedFiles)` and reject every valid cache.
  `TestNoBundledPackEmbedsTheCacheMarkerName` turns that from an unstated
  assumption into a guarded one.
- **`validatePackFiles`' per-pack unexpected-file check is not lost** — but that
  argument belongs to #4880, not to this PR: #4880 removed that check and
  `TestValidateSyntheticRepoRejectsStrayFilesAnywhere`, on `main` today, is what
  pins the union check subsuming it. This PR only removes what is left of
  `validatePackFiles`, which contains no file-set check at all. I have updated
  that test's doc comment where it names functions this change renames.
- **Overlapping paths are now compared once rather than twice.** `examples/bd`
  contains `examples/bd/dolt`, and the `bd` pack's `embed.FS` covers the `dolt/`
  subtree, so the two layouts resolve to the same cache-relative paths. The two
  `//go:embed` directives cover an identical file set, so the two comparisons
  could never have disagreed; what is dropped is a duplicate, not a check.
- **Naming the missing file costs nothing on the valid path.**
  `firstMissingExpectedFile` does a sorted `Lstat` scan, and it only runs once the
  walk has *already* established that the cache is short. I chose that over
  tracking a seen-set during the walk precisely because a seen-set would allocate
  on every valid pass and undo the point of the change.
- **What I deliberately did not do:** replace the byte comparison with anything
  cheaper. #4880 tried an mtime-based check in an earlier revision and CI caught
  that it was unsound. The full read is the contract, not an oversight, and it is
  why the timing does not move.
- **What is not covered:** a mutation that counted a path *not* in the expected
  set toward `seen`. That would require the unexpected-file branch to fall
  through instead of returning an error, which
  `TestValidateSyntheticRepoRejectsUnexpectedFiles` already catches — but by a
  different route, so I am naming it rather than claiming the count is pinned from
  both sides.

## Scope

| Area | Change |
| --- | --- |
| `internal/builtinpacks/registry.go` | `validatePackFiles` and `validateSyntheticRepoFileSet` replaced by `validateSyntheticRepoTree` plus `checkCachedFile`; the memoized `syntheticRepoAllowedPaths` file set now carries the expected content alongside each path; `firstMissingExpectedFile` added for the error message. |
| `internal/builtinpacks/registry_test.go` | Doc comment only. `TestValidateSyntheticRepoRejectsStrayFilesAnywhere` named two functions this change renames or deletes. |
| `internal/builtinpacks/registry_validate_tree_test.go` | New. Deletion (2 cases), unexpected directory (2 cases), fresh-cache acceptance, marker-name collision, same-length tamper. |
| `internal/builtinpacks/registry_validate_bench_test.go` | New. `BenchmarkValidateSyntheticRepo`. |
| Everything else | Untouched. No exported signature changes, no public API, no wire format, no config, no schema. |

## Notes for the reviewer

- The two labels carried in `expectedFile` (`pack`, `rel`) exist only to keep the
  existing error messages identical. Where two layouts resolve to the same
  cache-relative path — the `bd`/`dolt` nesting above — the later layout wins the
  label, which is deterministic because the layout list is a slice.
- `go build ./...`, `go vet ./...`, `go test ./internal/builtinpacks/` and
  `go test ./internal/packman/ ./internal/config/` are all clean on the rebased
  branch.

### DIFF
--- a/internal/builtinpacks/registry.go
+++ b/internal/builtinpacks/registry.go
@@ -371,15 +371,7 @@ func ValidateSyntheticRepo(dir, commit string) error {
 	if marker.ContentHash != wantHash {
 		return fmt.Errorf("bundled pack cache content hash %q does not match current binary %q", marker.ContentHash, wantHash)
 	}
-	if err := validateSyntheticRepoFileSet(dir); err != nil {
-		return err
-	}
-	for _, layout := range syntheticPackLayouts() {
-		if err := validatePackFiles(layout.Pack, filepath.Join(dir, filepath.FromSlash(layout.Subpath))); err != nil {
-			return err
-		}
-	}
-	return nil
+	return validateSyntheticRepoTree(dir)
 }
 
 // MaterializedFileMode returns the filesystem mode used for bundled pack files
@@ -476,45 +468,39 @@ func materializeFS(src fs.FS, dst string) error {
 	return nil
 }
 
-// validatePackFiles verifies a materialized pack against the embedded manifest:
-// every expected file present, with the expected mode and content.
+// validateSyntheticRepoTree walks the materialized cache once and enforces both
+// halves of the integrity contract in that single traversal: no unexpected path,
+// and every expected file byte-identical to the copy embedded in this binary.
 //
-// It does not walk dst looking for unexpected files. validateSyntheticRepoFileSet
-// already walks the whole cache once against the union of every layout's
-// manifest, and that union check strictly subsumes a per-pack one: a file
-// unexpected for its own pack is absent from the union too. Keeping both meant
-// about nine traversals of the same tree per call.
-func validatePackFiles(pack Pack, dst string) error {
-	manifest, err := manifestForPack(pack)
-	if err != nil {
-		return fmt.Errorf("reading bundled pack %q manifest: %w", pack.Name, err)
-	}
-	for rel, want := range manifest {
-		target := filepath.Join(dst, filepath.FromSlash(rel))
-		info, err := os.Lstat(target)
-		if err != nil {
-			return fmt.Errorf("checking bundled pack cache %q file %s: %w", pack.Name, rel, err)
-		}
-		if !info.Mode().IsRegular() || info.Mode().Perm() != want.perm.Perm() {
-			return fmt.Errorf("bundled pack cache %q file %s has mode %s, expected %s", pack.Name, rel, info.Mode().Perm(), want.perm.Perm())
-		}
-		got, err := os.ReadFile(target)
-		if err != nil {
-			return fmt.Errorf("reading bundled pack cache %q file %s: %w", pack.Name, rel, err)
-		}
-		if !bytes.Equal(got, want.data) {
-			return fmt.Errorf("bundled pack cache %q file %s content differs from current binary", pack.Name, rel)
-		}
-	}
-	return nil
-}
-
-func validateSyntheticRepoFileSet(dir string) error {
+// These used to be two passes over the same tree — this walk for the file set,
+// then a manifest-driven os.Lstat plus os.ReadFile per expected file in
+// validatePackFiles — which resolved all of the same paths twice. Every check
+// survives the fusion: content is still compared with bytes.Equal, mode is still
+// checked (now from the walk's own fs.DirEntry rather than a second Lstat), and
+// symlinks, unexpected files and unexpected directories are still rejected.
+//
+// Every rejection also keeps the exact message it had before, which is why the
+// per-file failure is recorded rather than returned from the walk callback: a
+// returned error would pick up the walk's own "file set" wrapper and report
+// content drift as a file-set problem. Recording it and continuing also
+// preserves the old precedence, in which the whole file-set walk ran before any
+// content was compared, so an unexpected path still outranks content drift no
+// matter which the walk reaches first.
+//
+// Missing files are the one property whose mechanism changes, and the seen count
+// is what preserves it. The old per-file Lstat could not miss a deletion; a walk
+// simply never visits a path that is not there, so without the count a cache
+// short of a file would validate clean — the worst direction for a
+// cache-integrity check to fail, because the rehydration it gates would never
+// fire.
+func validateSyntheticRepoTree(dir string) error {
 	allowedFiles, allowedDirs, err := syntheticRepoAllowedPaths()
 	if err != nil {
 		return err
 	}
 	firstUnexpectedDir := ""
+	var firstFileErr error
+	seen := 0
 	if err := filepath.WalkDir(dir, func(path string, entry os.DirEntry, err error) error {
 		if err != nil {
 			return err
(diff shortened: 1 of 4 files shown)


quad341 (approved):
@wbern — the two tests in `registry_validate_tree_test.go` are the real find here, and I checked `main` before saying so: there is still no test that removes an expected file to assert the cache is rejected, and none that creates an unexpected empty directory. You went looking for the exact spot where a naive fusion would silently accept a short cache, wrote the test *first*, ran it against all three versions, and then told us the one it would have caught in your own patch. That goes in with your authorship.

The implementation I have to leave behind, and that isn't a criticism of it. `main` moved underneath this branch twice since you opened it: `ValidateSyntheticRepo` is now `(dir, repository, commit)` with the allowed-path set scoped per repository, and `validatePackFiles` picked up a size+mtime memo that skips the file reads entirely on repeat validations. Your union walk fights both — the shortfall check would reject a valid scoped cache, and your own benchmark would likely now show the branch behind, since it revalidates one unchanged cache in a loop, which is exactly the case that memo wins. Rebasing this is a redesign, and it's yours to make, not mine to guess at.

So I'm cherry-picking the tests onto `main` with credit and closing this one. If you want another run at the traversal on top of the scoped version, I'd genuinely read it — the empty-directory fixtures alone told me you know this code better than the diff required.


---

PR #3731, opened 2026-06-25
fix(packman): resolve import versions against registry releases, not git tags

esciara:
## Problem

`gc import add <remote-pack> --version <constraint>` resolved the constraint against the **source repo's git tags** (`git ls-remote --tags`), never against the **registry's published release→commit table**. These are two independent version namespaces.

In a monorepo registry like `gastownhall/gascity-packs` — where packs are released by pinning commits on `main` and git tags are sparse/stale — conflating the two breaks the registry's own recommended command. `gc import add … --version '>=0.2.1'` resolved to a wrong, stale tag whose tree lacks the expected `pack.toml`; `validateCachedPackRoot` then failed and deleted the cache dir, surfacing a misleading **"missing pack.toml"** error.

## Fix

Version constraints for registry-known sources now resolve against the published release→commit mapping. Git tags remain the fallback **only** for sources no configured registry owns.

`packman` stays registry-blind via an **injected resolver seam** (no `packregistry` import, no import cycle):

- **packman**: extract the data-source-agnostic selection into exported `SelectVersion(candidates, constraint)`; add a nil-default `VersionResolver` seam + `SetVersionResolver`, consulted first in `resolveSource` (fail-closed on err, store on ok, else fall through to tag resolution).
- **cmd/gc**: `newRegistryVersionResolver` matches the normalized source against cached catalogs and selects the release commit; **fails closed** (listing available versions) when a registry-owned source has no matching release; returns `ok=false` for unknown sources, `sha:` pins, and cold caches. Installed once at process init; `doImportAdd` refreshes a missing catalog before sync (side effect confined to the interactive command).

The single `resolveSource` call site covers add/install/remove/upgrade and nested imports.

## Design

New design doc `engdocs/design/registry-aware-version-resolution.md` (Status: Proposed) captures the two-namespace bug, the injected-resolver contract, the fail-closed semantics, and rejected alternatives.

## Tests

- `internal/packman`: unit coverage for `SelectVersion` and the resolver seam.
- `cmd/gc`: unit coverage for `newRegistryVersionResolver`, plus an end-to-end regression test exercising `doImportAdd → resolver → syncImports → packs.lock` — a stale `v9.0.0` tag vs registry release `0.2.1`, proving the registry commit wins, plus a fail-closed negative case. (Implemented as a Go test rather than a static `.txtar` because the scenario references a runtime-generated commit SHA that can't be pinned hermetically in txtar.)

Fixes #3710

Relates #3659, gastownhall/gascity-packs#137, #3644

Issue #3710, opened 2026-06-24
bug: `gc import add --version` resolves pack version against repo git tags, not registry releases (registry's recommended command fails)


esciara:
### Before you continue

- [x] I searched existing issues and did not find a duplicate.
- [x] I read the relevant docs and contributor guidance.

### Gas City version

`dev` (commit `92cf6dbb`, built 2026-06-24).

### Environment

macOS (Darwin 25.5.0), zsh. Pack imported from the public `gastownhall/gascity-packs` monorepo via the `main` registry.

### Summary

`gc import add <remote-pack> --version <constraint>` resolves the version constraint against the **repo's git tags**, not the **registry's per-pack releases**. In a monorepo like `gastownhall/gascity-packs` — where packs are released by pinning commits on `main` and git tags are sparse/stale — this makes the registry's *own recommended import command* fail with a misleading error.

### Reproduction

1. Ask the registry how to import `contributing`:

   ```
   $ gc pack registry show contributing
   ...
   Import commands:
     This version or later: gc import add https://github.com/gastownhall/gascity-packs//contributing --name contributing --version '>=0.2.1'
   Releases:
     0.1.0 main 16f8297d4e62
     0.2.0 main 7dd57dd9be7d
     0.2.1 main 5b7c5c1320ee
   ```

   Note every release is pinned to a **commit on `main`**, not to a tag.

2. Run exactly the command the registry printed:

   ```
   $ gc import add https://github.com/gastownhall/gascity-packs//contributing --name contributing --version '>=0.2.1'
   ```

3. It fails, and the referenced cache directory does not exist afterward:

   ```
   gc import add "https://github.com/gastownhall/gascity-packs//contributing": import
   "pack:contributing": cached pack "https://github.com/gastownhall/gascity-packs//contributing"
   is missing pack.toml at /Users/.../.gc/cache/repos/787f71d5.../contributing/pack.toml
   ```

### Root cause

`--version` resolution uses repo-wide git tags (`git ls-remote --tags`, highest matching semver), not the registry's per-pack release → commit table.

- `gastownhall/gascity-packs` has only two git tags: `v0.2.0` and `v0.3.0`.
- `>=0.2.1` resolves to the highest tag, `v0.3.0` (commit `9d53ae5f`).
[...]

PR #4925, opened 2026-08-03
fix(import): local-in-git pack imports lock to HEAD, not latest tag (#3659)

### LINKED ISSUE #3659: bug: gc import add of a local in-git pack locks to latest tag instead of HEAD
### Before you continue

### Gas City version

1.1.1 (commit: d21a32ce9-dirty, built 2026-06-22). Reproduced on a freshly built `main`.

### Environment

Linux (Ubuntu-family), bash, tmux runtime provider, managed server-mode Dolt beads backend. `gc` installed via `make install`.

### Reproduction

`gc import add` of a local pack that lives **inside a git repo** pins the import to the repo's latest semver **tag** instead of the current checkout (HEAD). When the pack was added after the latest tag, the resolved tag does not contain it, and the import fails.

1. Have a git repo (e.g. `gascity-packs`) whose latest tag is `v0.3.0`, with a pack subdirectory (`compound-engineering/`) that was added to `main` **after** `v0.3.0` was tagged. `pack.toml` exists at HEAD but not at `v0.3.0`.
2. From a city, run a plain import (no `--version`):
   ```
   gc import add ~/rigs/gascity-packs/compound-engineering/
   ```
3. Observe the failure:
   ```
   gc import add "file:///home/bobby/rigs/gascity-packs//compound-engineering":
     import "pack:compound-engineering": cached pack
     "file:///home/bobby/rigs/gascity-packs//compound-engineering" is missing
     pack.toml at /home/.../cache/repos/802048.../compound-engineering/pack.toml
   ```

The reported cache directory does not exist on disk afterward, because the failed `validateCachedPackRoot` deletes the cache (`internal/packman/cache.go:139`), which makes the error doubly confusing.

**Workaround:** pin to the current commit explicitly, which bypasses tag resolution:
```
gc import add ~/rigs/gascity-packs/compound-engineering/ \
  --version "sha:$(git -C ~/rigs/gascity-packs rev-parse HEAD)"
```

### Expected behavior

Per `gc import add --help`: "local paths inside git worktrees at HEAD: promoted to a file:// repo source with the pack subpath and **locked to the current commit**." A plain `gc import add <local-in-git-pack>` should lock to the current HEAD commit (which contains the pack), not to the repo's latest semver tag.

### Actual behavior

`defaultImportVersionForSource` (`cmd/gc/cmd_import.go:1496`) calls `ResolveVersion(source, "")`, which runs `git ls-remote --tags` and selects the highest semver tag. It only falls back to the HEAD commit when the repo has **no** semver tags (`ErrNoSemverTags`, `cmd_import.go:1501-1508`). So for any tagged repo, a local in-git import resolves to the latest tag rather than HEAD — contradicting the documented "locked to the current commit" behavior. If the pack does not exist at that tag, the clone/checkout lacks `pack.toml` and `validateCachedPackRoot` reports the misleading "missing pack.toml" (and deletes the cache dir).

```
$ GIT_TRACE=/tmp/t.log gc import add ~/rigs/gascity-packs/compound-engineering/
# relevant git commands gc ran:
git rev-parse --show-toplevel
git ls-remote --tags file:///home/bobby/rigs/gascity-packs
git clone --quiet file:///home/bobby/rigs/gascity-packs <cache>/802048...
git checkout --quiet 9d53ae5f43efa1cb4c0c69a4bf5fd6730cfbb5b7   # == tag v0.3.0, NOT HEAD

$ git -C ~/rigs/gascity-packs cat-file -e 9d53ae5f:compound-engineering/pack.toml
fatal: path 'compound-engineering/pack.toml' exists on disk, but not in '9d53ae5f'   # missing at the resolved tag
$ git -C ~/rigs/gascity-packs cat-file -e HEAD:compound-engineering/pack.toml && echo present
present   # present at HEAD (fc6ae36)
```

### Additional context

Two reasonable fixes:
1. For a local in-git source, lock to HEAD (`sha:<commit>`) instead of preferring repo-wide semver tags — this matches the documented behavior and the user's intent ("import the pack as it exists at my checkout").
2. At minimum, make the error report the resolved version vs. the worktree commit, e.g. "pack not found at resolved version v0.3.0; worktree is at fc6ae36", instead of a bare "missing pack.toml".

Related: #2118 (docs: improve `gc import add` documentation).

rjgeng:
## Summary
Fixes #3659: `gc import add <local-in-git-pack>` (no `--version`) is documented to lock to HEAD for local paths inside git worktrees, but the default resolution actually preferred the repo's latest semver tag whenever one existed, only falling back to HEAD when the repo had zero tags. A pack added to the worktree after the last tag was cut resolves to a checkout whose tree predates the pack, so `validateCachedPackRoot` fails with a misleading "missing pack.toml" error (and deletes the cache dir on the way out).

Checked for an existing PR first: #3731 fixes the sibling issue #3710 (a *remote* import with an explicit `--version` constraint resolved against git tags instead of the registry's release table) — a different call path (`internal/packman/resolve.go`'s constrained `VersionResolver` seam) from this bug, which is in the *unconstrained default* path (`internal/importsvc/importsvc.go`'s `defaultImportVersionForSource`). Confirmed genuinely uncovered by #3731.

## Fix
`normalizeImportAddSource` now also reports whether a source was promoted from a local git worktree (`internal/importsvc/source.go`); when it was and no `--version` constraint is given, `resolveImportVersion` locks straight to `sha:<HEAD commit>` instead of going through tag-based default resolution. Registry/remote sources are untouched — same code path, same gating, no change to their behavior.

## Test plan
- [x] New regression test `TestDoImportAddLocalGitRepoWithStaleTagLocksToHEADNotTag` builds a real git repo (tag cut before `pack.toml` exists, so the tag's tree lacks it) — confirmed red against unfixed code (reproduces the exact "missing pack.toml" error from the issue), green after the fix.
- [x] Updated a pre-existing test (`TestDoImportAddLocalGitRepoWithTagsWritesDefaultConstraint`, renamed to `...StillLocksToHEAD`) that had been asserting the buggy tag-preferring behavior as expected — it only passed before because its tag and HEAD happened to be the same commit, masking the drift case.
- [x] Full `internal/importsvc` and `cmd/gc` import suites green, `go vet` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **`gc import add` of a local in-git pack now locks to HEAD, not the repo's
+  latest tag.** Per `gc import add --help`, a local path inside a git
+  worktree is documented to be "locked to the current commit," but the
+  default version resolution (absent an explicit `--version`) preferred the
+  repo's latest semver tag whenever one existed. A pack added to the
+  worktree after the last tag was cut resolved to a checkout whose tree
+  predates the pack, failing with a misleading "missing pack.toml" error
+  even though the pack exists at HEAD. A local-worktree-promoted source now
+  always locks to `sha:<HEAD commit>`, matching the documented behavior;
+  registry/remote sources are unaffected. Fixes #3659.
+
 - **The dolt pack's `run_bounded` python3 fallback now sends SIGTERM before
   SIGKILL, matching its documented contract.** The fallback (used when
   neither `timeout` nor `gtimeout` is on `PATH`, the default on stock macOS)
--- a/cmd/gc/cmd_import_test.go
+++ b/cmd/gc/cmd_import_test.go
@@ -2358,7 +2358,15 @@ scope = "city"
 	}
 }
 
-func TestDoImportAddLocalGitRepoWithTagsWritesDefaultConstraint(t *testing.T) {
+// TestDoImportAddLocalGitRepoWithTagsStillLocksToHEAD confirms a local-in-git
+// import locks to the worktree's HEAD commit even when the repo also has a
+// semver tag that happens to point at the same commit -- a local worktree
+// source locks to HEAD unconditionally, per gc import add --help ("local
+// paths inside git worktrees at HEAD: ... locked to the current commit") and
+// the gastownhall/gascity#3659 fix; the tag's presence is incidental, not a
+// signal to prefer semver-constraint resolution over the worktree the user
+// actually pointed at.
+func TestDoImportAddLocalGitRepoWithTagsStillLocksToHEAD(t *testing.T) {
 	clearGCEnv(t)
 	dir := t.TempDir()
 	writeCityToml(t, dir, "[workspace]\nname = \"demo\"\n")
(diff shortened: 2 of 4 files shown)


quad341 (approved):
@rjgeng — thank you for this one, and specifically for the work you did *before* writing any code. Checking for an existing PR, finding #3731, and then articulating why it doesn't cover this — `packman/resolve.go`'s constrained resolver seam versus `importsvc`'s unconstrained default — saved us that whole investigation, and it held up when we traced it.

The regression test is the other reason this ships as-is. Building a real repo where the tag is cut before `pack.toml` exists reproduces #3659 at exactly the layer where it hurt, and asserting the `packs.lock` commit alongside the manifest version means a future refactor can't half-fix it.

Special credit for how you handled `TestDoImportAddLocalGitRepoWithTagsWritesDefaultConstraint`. Noticing that a green test was only green because its tag and HEAD happened to coincide — then renaming it and writing down *why*, instead of quietly editing the assertion — is what keeps a suite trustworthy.

We traced the `file://` subpath path (`NormalizeRemoteSource` strips it, so the probe is correct), re-ran the import suites and `go vet` clean, and CI is fully green. Merging as-is. Two small follow-ups are on us, not you: a subpath-pack test and the explicit-`file://`-source case, which still takes the old path.


---

Issue #3001, opened 2026-06-03
bug: `gc init` does not detect llm runtime cli (claude, codex, etc) when installed through pnpm


sarendipitee:
### Before you continue

- [x] I searched existing issues and did not find a duplicate.
- [x] I read the relevant docs and contributor guidance.

### Gas City version

1.2.1

### Environment

ubuntu 24.04

### Reproduction

$ pnpm install -g @openai/codex
$ mkdir test && cd test
$ gc init

Select codex

```
gc init
City "gastown" already exists; reusing existing configuration and resuming startup checks.
[6/8] Checking provider readiness
gc init: city created, but startup is blocked by provider readiness

Referenced providers not ready:
- Codex CLI: is not installed
  Fix: install Codex CLI

Next: cd ~/test && gc start
```
$ which codex 
~/.local/share/pnpm/bin/codex
$ codex --version
codex-cli 0.136.0

Same happens for claude code, gemini, and I assume all others

### Expected behavior

It recognizes the availability of the runtimes

### Actual behavior

It does not recognize the availability of the runtimes

### Logs, screenshots, or traces

```shell

```

PR #4926, opened 2026-08-03
fix(searchpath): find Nix-installed provider CLIs (#3962)

### LINKED ISSUE #3962: bug: gc init fails to detect Gemini CLI installed via Nix (does not respect $PATH)
### Before you continue

### Gas City version

1.3.3

### Environment

macosx , gc installed via nix

### Reproduction

 1. Install the gemini-cli using Nix.
 2. Verify the gemini binary is available in $PATH and executable:
    ```bash
      $ which gemini
      /Users/username/.nix-profile/bin/gemini

      $ gemini --version
      0.47.0
    ```
 3. Run gc init with Gemini as the provider:
    ```bash
      $ gc init --default-provider gemini
      gc init: provider readiness preflight failed: Gemini CLI: is not installed
    ```

### Expected behavior

 gascity should fall back to checking $PATH (e.g., using exec.LookPath("gemini")) if the binary isn't found in standard npm directories, or prioritize $PATH entirely.

### Actual behavior

gc init: provider readiness preflight failed: Gemini CLI: is not installed

```shell

```

### Additional context

When the gemini CLI is installed via Nix and is fully accessible in the system $PATH, gc init --default-provider gemini fails with a "not installed" error.

 It appears that gascity explicitly checks hardcoded npm global installation paths (such as ~/.npm-global/bin) to find the CLI, rather than resolving the binary using the user's
 $PATH.

rjgeng:
## Summary
Fixes #3962: `gc init`'s provider-readiness probes (`api.ProbeProviders`) use a deterministic, user-aware install-path search (`internal/searchpath`) rather than the ambient `$PATH`, deliberately, so API calls don't depend on shell-specific edits. That search list already special-cases npm/pnpm/yarn/cargo/nvm/etc. (with a comment citing the identical prior fix for pnpm, #3001), but had no Nix entries — so `gc init --default-provider gemini` reported "is not installed" for a Nix-installed `gemini` binary even though it was on the shell's `PATH`.

## Fix
Added the two current Nix profile bin locations to the existing search list in `internal/searchpath/searchpath.go`'s `userManagedDirs`: `~/.nix-profile/bin` (classic `nix-env`, matches the issue's exact repro) and `~/.local/state/nix/profiles/profile/bin` (newer `nix profile install` default). Same existence-gated pattern as every other entry in that list — zero risk to hosts without Nix.

## Test plan
- [x] 3 new tests: `TestExpandIncludesNixProfileBin`, `TestExpandIncludesNixProfileInstallBin`, `TestExpandOmitsAbsentNixProfileBin` — confirmed red against unfixed code, green after the fix.
- [x] Full `internal/searchpath` suite green, `go build ./...` and `go vet ./internal/searchpath/... ./internal/api/...` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -107,6 +107,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   bound or explicitly configured endpoints keep bd's own output unchanged
   (gastownhall/gascity#1374).
 
+- **`gc init` provider readiness now finds Nix-installed CLIs.** Provider
+  readiness probes (`gc init --default-provider gemini` and friends) search a
+  deterministic, user-aware set of install directories rather than the
+  ambient `$PATH`, so a CLI installed via Nix (`~/.nix-profile/bin` for
+  classic `nix-env`, `~/.local/state/nix/profiles/profile/bin` for the newer
+  `nix profile install`) was reported as "not installed" even when it was on
+  the shell's `PATH`. Both locations are now included, matching the existing
+  npm/pnpm/yarn/cargo/nvm handling in `internal/searchpath`. Fixes #3962.
+
 - **ACP activity is now available across process boundaries.** ACP
   `session/update` timestamps are published through an atomic, coalesced
   sidecar, allowing a process other than the session owner to report
--- a/internal/searchpath/searchpath.go
+++ b/internal/searchpath/searchpath.go
@@ -101,6 +101,15 @@ func userManagedDirs(homeDir string) []string {
 		filepath.Join(homeDir, ".npm-global", "bin"),
 		filepath.Join(homeDir, ".yarn", "bin"),
 		filepath.Join(homeDir, ".config", "yarn", "global", "node_modules", ".bin"),
+		// Nix single-user profile bins. `.nix-profile` is the classic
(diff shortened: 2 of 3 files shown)


quad341:
Merged as e425e6691 — thank you, and I'm sorry this sat as long as it did.

Reviewed properly before merging, and the thing we checked hardest was the one
that would have mattered most: whether adding Nix locations could change
resolution order for someone *not* on Nix. It can't. The new entries are
existence-gated exactly like every sibling in `userManagedDirs` and are inserted
after ambient PATH, so on a non-Nix host they're inert, and they can never
override a real PATH match. It also follows the same shape as the existing
pnpm/#3001 entry in that function, which made it easy to be confident about.
Verified locally too: build and vet clean, 18/18 searchpath tests passing —
including the three you added, which cover both the positive and the negative
existence-gate cases. Tests that prove the *absence* of an effect are the ones
that make a change like this reviewable, so thank you for writing them.

On the delay: 37 days, and none of it was waiting on you. The PR was green,
conflict-free and mergeable that whole time. We had a review-queue gap where a PR
that's simply "green and ignored" matched none of our staleness or conflict
checks, so nothing surfaced it — it wasn't in a queue at all. We found it this
morning while auditing that gap, which is now being fixed properly rather than
patched over. That's our bug, not a reflection on the contribution.

PR #3285, opened 2026-06-10
fix(wisp-gc): collect plain task wisps in the wisp tier

realies:
## Problem

On a busy coordination DB, **closed plain task wisps accumulate without bound** in the wisp tier. The wisp GC has two selectors for collectable closed beads:

- `Type=="molecule"` (closed molecule roots)
- `Metadata["gc.kind"]=="wisp"` (explicitly-tagged wisp roots)

But sling/convoy trackers, loose step beads, and other plain task beads created in the wisp tier are **neither** `Type=="molecule"` **nor** tagged `gc.kind=wisp`. The two selectors never see them, so they live forever — closed, expired, but uncollected — inflating every ephemeral-tier scan (`bd ready`, reconcile ticks, dispatch polls all get slower as the wisp table grows).

In production we saw the wisp table grow into the thousands of closed-but-uncollectable task wisps purely from normal sling/convoy churn.

## Fix

Add a third selector to `closedWispGCEntries`: list **all** closed wisp-tier beads (`TierMode: TierWisps`), since everything in the wisp tier is ephemeral by definition — a closed wisp-tier bead past TTL is collectable. Two ownership exclusions so we don't poach beads another retention path owns:

- `Type=="message"` → owned by read-mail retention (`mailRetentionTTL` / `readMessageWispGCEntries`)
- `labelOrderTracking`-labeled → owned by `gc order sweep-tracking` + the `[beads.policies.order_tracking].delete_after_close` policy

Everything else (closed, in the wisp tier, past `WispTTL`) is collected through the existing closure-aware delete path, so step/child beads are torn down with their roots exactly as molecules are.

## Tests

- New `TestWispGC_PurgesPlainTaskWispsInWispTier` (discriminating: **RED** without the new selector — a closed task wisp survives — **GREEN** with it) plus preservation assertions: order-tracking beads, message wisps, fresh (under-TTL) wisps, open wisps, and non-wisp-tier beads are all left untouched.
- Full `cmd/gc` suite green.
- Deployed on a live city: a planted closed task wisp backdated past TTL was collected on the next GC tick; no regression in molecule/mail/order-tracking collection.

PR #4927, opened 2026-08-03
fix(wisp-gc): reap rootless plain-task wisps (#3780)

### LINKED ISSUE #3780: wisp-GC: reap rootless/loose plain-task wisps (residual gap from #3285)
Follow-up to #3285 (closed as superseded). Current main's `reapOrphanedClosedWisps` only reaps **closed wisp-tier beads carrying a root pointer** (`gc.root_bead_id`). The closed PR #3285 also targeted **rootless / loose plain-task wisps** in the wisp tier — that case is not covered by the selector-model refactor.

**Gap:** plain-task wisps with no root pointer can accumulate uncollected in the wisp tier.

**Scope:** extend `enumerateWispGCRoots` / the reap path to also collect rootless plain-task wisps, with a test. Should be a small, bounded change against the current selector model (`cmd/gc/wisp_gc.go`).

Context: original analysis in the #3285 adoption review (modify/modify conflict made a mechanical rebase impossible; re-authoring against the new model is the clean path).

rjgeng:
## Summary
Fixes #3780: wisp-GC left rootless, closed plain-task wisps uncollected indefinitely — a residual gap from #3285.

`reapOrphanedClosedWisps` (`cmd/gc/wisp_gc.go`) skipped any closed wisp-tier row with no `gc.root_bead_id` pointer outright, and the separate root-rooted closure purge (`wispGCRootSelectors` / `closedWispGCEntries`) never enumerates it either — it matches none of the three root selectors (molecule, `gc.kind=wisp`, graph.v2 workflow). A closed plain-task wisp that never had an owning root fell through both paths and accumulated uncollected indefinitely.

## Fix
A rootless row is now reaped when it's a plain task (`c.Type == "task"`) — its own closed status (already guaranteed by the candidates query) is the only fence needed, since there's no root to check for collectibility. A rootless row that is **not** a plain task stays out of scope, preserving the original safety boundary (doesn't risk collecting an unrecognized bead shape that might indicate an actual bug elsewhere rather than an intentionally rootless task).

## Test plan
- [x] Found a pre-existing test (`TestWispGC_ReapSkipsRowsWithoutRootPointer`) that had been asserting the buggy skip-everything behavior as expected — renamed/corrected to `TestWispGC_ReapsRootlessPlainTaskWisp` (confirmed red against unfixed code, green after the fix).
- [x] Added `TestWispGC_ReapSkipsRootlessNonTaskRow` to lock in the preserved safety boundary for non-task shapes.
- [x] Full `TestWispGC*` suite green (39 tests), `go vet` clean.


quad341:
@rjgeng — two things here are genuinely good maintainer work. You found that `TestWispGC_ReapSkipsRowsWithoutRootPointer` was *asserting the bug* as intended behavior, and corrected it instead of writing around it — a test that encodes a leak is worse than no test, and those are easy to walk past. And you caught the strand-the-descendants hazard in your own first draft, adding the leaf fence in `04cc9a7`: `deleteWorkflowBead` removes one bead, not a closure, so reaping a rootless subtree owner would have put its children beyond either GC path forever.

Two of our reviewers flagged that `store.Children(..., WithBothTiers)` loses the tier on the bd backend. I traced it — it doesn't. The graph store is policy-wrapped, and `beadPolicyStore.Children` honors the tier (it even widens `TierIssues` to `TierBoth`); both relocated engines are SQLite/NativeDolt, which honor it too. Your call was right, and I'll leave a comment saying so.

What I'll take on my side is smaller: the batch cap is gated on `enforce`, so in the dry-run default your child probe runs once per aged rootless candidate with no bound — one `bd list` each on the controller tick, against exactly the backlog this fixes. I'll bound the probes, add a test, and merge with your commits intact.

The red CI is ours: `sum.golang.org` threw an INTERNAL_ERROR mid module-verify. Re-running. If you get a chance, the PR description could use a refresh — it still describes the pre-fence version.


---

quad341 (approved):
@rjgeng — two calls in this PR are why it's landing rather than bouncing.

You spotted that `TestWispGC_ReapSkipsRowsWithoutRootPointer` was asserting the leak as *intended* behavior, and you corrected the test instead of writing around it. A test that encodes a bug is worse than no test, and it's very easy to read one and move on.

You also caught the strand hazard in your own first draft and added the leaf fence. That was right — `deleteWorkflowBead` removes one bead, not a closure, so reaping a rootless subtree owner would have put its children beyond either GC path permanently.

What I'm taking is small and squarely ours. The batch cap is gated on `enforce`, so under the dry-run default your child probe runs once per aged rootless candidate with nothing bounding it — one `bd` read each, every tick, against exactly the backlog this fixes. I'll bound the probes, extend the fence to catch dep-row parents, add tests, and merge with your commits intact.

The red CI is ours too: `TestFreshManagedBdCityInitSeedsPinnedHQDatabaseAndKeepsGCPrefix`, a Dolt schema-migration failure in a city-init test nowhere near your change. Re-running. Thank you for this one.


---

PR #4930, opened 2026-08-03
Normalize canonical path comparison inputs in convergence and dispatch

quad341:
## What this changes

Convergence gate and dispatch artifact paths now make identity and containment decisions from canonical inputs. Relative, real, and symlinked spellings of the same envelope, base, or city directory therefore produce the same result instead of triggering false path-escape rejection or returning a relative prompt path.

The change deliberately keeps direct `filepath.EvalSymlinks` calls where resolution failure is part of the behavior, such as missing artifact directories, dangling condition scripts, and retry artifact targets. Those sites are documented beside the call so future migrations preserve their error semantics.

## Review notes

- Review the convergence condition-path and evaluate-prompt boundaries together with dispatch retry artifact containment; these are the three comparison inputs and six existence/resolvability checks covered by the classification.
- Missing-path and symlink-escape behavior remains fail-closed where required; missing retry artifacts retain their existing classification path.
- No API, configuration, persistence, dependency, or migration change is included.

## Test plan

- [x] `make test-fast-parallel` — 10 jobs passed, 0 failed, 0 skipped
- [x] `go build ./...` and `go vet ./...`
- [x] `go test -count=1 ./internal/convergence/... ./internal/dispatch/...` — 738 passed, 0 failed, 0 skipped
- [x] Release gate: [`release-gates/ga-94bs0w-canonical-path-convergence-dispatch-gate.md`](release-gates/ga-94bs0w-canonical-path-convergence-dispatch-gate.md)

### DIFF
--- a/internal/convergence/artifact.go
+++ b/internal/convergence/artifact.go
@@ -32,6 +32,12 @@ func ValidateArtifactDir(dir string) error {
 	}
 	// Canonicalize with EvalSymlinks so comparisons are consistent
 	// when the artifact root itself contains symlinked components.
+	// canonical-path-exception: existence/resolvability only, not comparison
+	// preparation. absDir is already absolute (filepath.Abs above), so this
+	// call cannot diverge into a relative/absolute mismatch; its error path
+	// is the deliberate "artifact directory must exist" check for this
+	// function, which pathutil.NormalizePathForCompare's never-errors
+	// contract would silently swallow.
 	absDir, err = filepath.EvalSymlinks(absDir)
 	if err != nil {
 		return fmt.Errorf("resolving artifact directory: %w", err)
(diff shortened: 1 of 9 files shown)


quad341:
@quad341 — the per-site classification matrix in the release gate is what makes this reviewable at all. Nine `EvalSymlinks` sites, each sorted into "comparison preparation" vs "existence/resolvability", the reasoning parked right next to the call, and a pinning test for every behavior you deliberately kept. That's the shape the rest of this migration should copy.

One thing I'll absorb before merge: `pathutil.NormalizePathForCompare` does slightly more than absolutize-then-resolve — on darwin it also collapses `/private/var` → `/var` and `/private/tmp` → `/tmp`. That's the reverse of what bare `EvalSymlinks` returns, so the post-resolution containment check in `condition.go` and the `realResolved != resolved` check in `evaluate.go` end up comparing across two conventions. Linux can't see it (the collapse is a no-op there) and the Mac jobs were skipped here, so CI was never going to catch it.

I'll route both comparisons through `pathutil` so each side normalizes, extend your comments to name the alias behavior, add darwin-guarded regressions, and dispatch Mac Regression before merging. Your commits and credit stay as-is.


---

PR #4931, opened 2026-08-03
fix: preserve runtime native step topology

julianknutsen:
## Summary

- preserve the canonical prerequisite set when Attach creates another physical occurrence of the same semantic step
- include resolvable fragment ExternalDeps in native topology for both graph-plan and sequential materialization
- omit topology when any required external predecessor lacks a native step identity, preserving UNKNOWN instead of publishing a false root
- leave iteration identity, the wire tri-state/domain, and downstream APIs unchanged

## TDD evidence

The focused regressions first failed because every case emitted an authoritative empty array:

- TestAttachPreservesNativeStepDependenciesAcrossRetryAttempts
- TestAttachKeepsRetryTopologyUnknownWhenParentTopologyIsUnknown
- TestInstantiateFragmentIncludesCompleteExternalNativeStepDependencies
- TestInstantiateFragmentOmitsTopologyWhenExternalNativeStepIsUnknown

After the correction:

- go test ./internal/molecule -count=1
- go test ./internal/dispatch -run Test\(RetryLifecycle\|ProcessFanoutSequentialResumeRestoresExternalDeps\) -count=1
- go vet ./internal/molecule ./internal/dispatch
- pre-commit lint, generated-contract checks, and go vet ./...

No producer activation is part of this change.

Tracking: ga-15w7g.26.3

## Runtime export follow-up

Window 3's deploy-branch smoke exposed one projection defect: an authoritative root reached ProjectEvent as a present empty dependency slice but normalized to a nil slice, so JSON emitted `depends_on_step_ids:null` instead of `[]`.

TDD correction in `162c6eda18`:

- strengthened `TestProjectEventNormalizesNativeStepDependencies` to exercise ProjectEvent through JSON encoding
- changed normalization to preserve a non-nil empty slice
- retained omitted UNKNOWN and populated sorted dependencies unchanged
- RED reproduced `null`; GREEN passed 10 focused runs, the full `pkg/eventexport` and `internal/eventfeed` suites, focused vet, and the pre-commit lint/generated/vet checks

Exporter regression: ga-de4ei.

### Deploy-branch re-test

Window 3 cherry-picked only the exporter delta onto the existing deploy branch as `a0a1fcd23691868c01dba53bb5a6c2a76fa8a0a3`. A clean-clone build reported that exact unmodified VCS revision and Maintainer City returned to RUN.

The projection and real-runtime event smokes both passed:

- UNKNOWN omitted `depends_on_step_ids`
- authoritative roots emitted `depends_on_step_ids:[]`
- populated dependencies emitted sorted arrays
- a two-parent join retained both prerequisite IDs in lexical order
- every envelope passed validation and the exported wire contained zero `null` topology values

The runtime smoke used a throwaway city that was removed and unregistered afterward. Producer v3 remained paused and Maintainer City work was not poured or mutated.

### DIFF
--- a/internal/molecule/graph_apply.go
+++ b/internal/molecule/graph_apply.go
@@ -134,7 +134,10 @@ func buildRecipeApplyPlan(recipe *formula.Recipe, opts Options) (*beads.GraphApp
 	if len(recipe.Steps) == 0 {
 		return nil, false, "", fmt.Errorf("recipe %q has no steps", recipe.Name)
 	}
-	recipe = recipeWithNativeStepDependencies(recipe)
+	if !opts.nativeStepTopologyPrepared {
+		recipe = recipeWithNativeStepDependencies(recipe)
+		opts.nativeStepTopologyPrepared = true
+	}
 
 	vars := applyVarDefaults(opts.Vars, recipe.Vars)
 	priorityOverride := clonePriority(opts.PriorityOverride)
@@ -394,7 +397,13 @@ func buildFragmentApplyPlan(store beads.Store, recipe *formula.FragmentRecipe, o
 	if len(recipe.Steps) == 0 {
 		return &beads.GraphApplyPlan{}, nil
 	}
-	recipe = fragmentRecipeWithNativeStepDependencies(recipe)
+	if !opts.nativeStepTopologyPrepared {
+		recipe = fragmentRecipeWithNativeStepDependencies(recipe)
+		if err := applyExternalNativeStepDependencies(store, recipe.Steps, opts.ExternalDeps); err != nil {
+			return nil, err
+		}
+		opts.nativeStepTopologyPrepared = true
+	}
 
 	existingLogicalBeadIDs, err := existingLogicalBeadIDIndex(store, opts.RootID)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


quad341 (approved):
@julianknutsen — the part of this I most appreciate is that you found the second bug yourself. Fixing the retry/fragment topology and then catching on the deploy-branch smoke that an authoritative root was still landing as `depends_on_step_ids:null` — when `null` vs `[]` *is* the whole UNKNOWN-vs-known-root distinction — is the kind of follow-through that makes this layer trustworthy. That one-character `[]string(nil)` → `[]string{}` change is easy to miss and easy to get wrong in the other direction.

I traced the `nativeStepTopologyPrepared` guard through `Attach` → `Instantiate` → `buildRecipeApplyPlan` and through both fragment paths; the preserved fact survives everywhere it should, and nothing mutates a caller's recipe. Asserting the exporter fix through real JSON encoding rather than the struct was the right call — the struct-level check passed on the broken code.

Merging as-is. Two things I'm filing on our side rather than asking you to change: `Options.ExternalDeps` on the recipe path (drain items) still has the same false-root class, and `store.Get` failing on an external dep is now a hard abort where UNKNOWN might suit the design better. Neither is in scope for what you set out to fix. Thanks for the smoke evidence — it made this fast to verify.


---

PR #4933, opened 2026-08-03
Normalize canonical path comparisons in sessionlog readers

quad341:
## What this changes

Sessionlog discovery now treats real paths, symlink aliases, and relative aliases as the same physical identity when deduplicating Codex rollout candidates and Kimi session roots. This prevents duplicate records and false ambiguity when the same transcript tree is reachable through more than one path spelling.

Traversal behavior stays intentionally conservative: broken symlink roots are skipped, while returned Kimi transcript paths retain the caller-provided lexical root. Canonical paths are comparison keys only, so callers do not unexpectedly receive absolute paths.

## Review notes

- Four comparison-only ingest sites use `pathutil.NormalizePathForCompare` for stable dedup keys.
- Six `filepath.EvalSymlinks` sites remain as functional resolvability checks, each with an adjacent exception comment.
- No API, configuration, wire-shape, or migration changes are included.

## Test plan

- [x] `make test-fast-parallel` — 10/10 jobs
- [x] `make test-cmd-gc-process-parallel` — 7/7 jobs
- [x] PR integration matrix — 14/14 shards
- [x] Worker Phase 1 and Phase 2 conformance — Claude, Codex, and Gemini profiles
- [x] `go test -count=1 -json ./internal/sessionlog/...` — 293 PASS, 0 FAIL, 7 Darwin-only SKIP
- [x] Release gate: [`release-gates/ga-w7a2xm-sessionlog-canonical-path-readers-gate.md`](release-gates/ga-w7a2xm-sessionlog-canonical-path-readers-gate.md)

### DIFF
--- a/internal/sessionlog/kimi_reader.go
+++ b/internal/sessionlog/kimi_reader.go
@@ -175,11 +175,11 @@ func findKimiSessionFilesIn(root, workHash string) []kimiContextCandidate {
 }
 
 func findKimiSessionFilesInVisited(root, workHash string, visited map[string]bool) []kimiContextCandidate {
-	root = canonicalKimiSessionRoot(root)
-	if root == "" || visited[root] {
+	canonical := canonicalKimiSessionRoot(root)
+	if canonical == "" || visited[canonical] {
 		return nil
 	}
-	visited[root] = true
+	visited[canonical] = true
 
 	workRoot := filepath.Join(root, workHash)
 	files := kimiContextFiles(workRoot)
@@ -206,11 +206,11 @@ func findKimiSessionFileByIDIn(root, workHash, sessionID string) string {
 }
 
 func findKimiSessionFileByIDInVisited(root, workHash, sessionID string, visited map[string]bool) string {
-	root = canonicalKimiSessionRoot(root)
-	if root == "" || visited[root] {
+	canonical := canonicalKimiSessionRoot(root)
+	if canonical == "" || visited[canonical] {
 		return ""
 	}
-	visited[root] = true
+	visited[canonical] = true
 
 	path := kimiTranscriptPath(filepath.Join(root, workHash), sessionID)
 	info, err := os.Stat(path)
--- a/internal/sessionlog/kimi_reader_test.go
+++ b/internal/sessionlog/kimi_reader_test.go
@@ -313,6 +313,32 @@ func TestFindKimiSessionFileFollowsSymlinkedRoots(t *testing.T) {
 	}
 }
 
+// Companion to TestFindKimiSessionFileFollowsSymlinkedRoots: a dangling
+// symlinked account root is a deliberate existence check (bare EvalSymlinks
+// site), not comparison prep, so it must be skipped without error and must
+// not block discovery through a later, valid sibling root.
+func TestFindKimiSessionFileSkipsBrokenSymlinkedRoot(t *testing.T) {
+	isolateKimiSearchRoots(t)
+	base := t.TempDir()
+	accountRoot := t.TempDir()
+	workDir := "/tmp/gascity/phase1/kimi-broken"
+	workHash := kimiWorkDirHash(workDir)
+	want := writeKimiContext(t, filepath.Join(accountRoot, workHash, "session-key", "context.jsonl"), []string{
+		`{"role":"user","content":"via valid root"}`,
+	})
+
+	if err := os.Symlink(filepath.Join(base, "does-not-exist"), filepath.Join(base, "account-a-broken")); err != nil {
+		t.Skipf("symlink unsupported on this platform: %v", err)
+	}
+	if err := os.Symlink(accountRoot, filepath.Join(base, "account-b-valid")); err != nil {
+		t.Fatal(err)
+	}
+
+	if got := FindKimiSessionFile([]string{base}, workDir); !samePath(got, want) {
+		t.Fatalf("FindKimiSessionFile() = %q, want %q (broken sibling root must be skipped, not fatal)", got, want)
+	}
+}
+
 // isolateKimiSearchRoots pins every ambient variable DefaultKimiSearchPaths
 // reads, so a discovery test sees only the roots it creates. Pinning HOME alone
 // is insufficient: KIMI_CODE_HOME contributes a root independently of HOME, and


quad341:
@quad341 — the part of this I want to call out is the per-site classification matrix: splitting all ten `EvalSymlinks` calls into "comparison key" versus "does this thing actually resolve," then documenting that split inline at each retained site, is what makes this maintainable a year from now. The not-yet-existing-leaf test for `markCodexBatchRoot` is a sharp catch, and keeping canonical paths as keys only — so callers never get surprise absolute paths — is exactly the right call.

One thing I'm picking up on my side. Decoupling `key` from `root` in the Kimi readers means discovery now returns lexically-rooted paths, which is what you wanted, but `FindKimiSessionFileIfUnambiguous` still dedups candidates by raw path string. So one transcript reachable through two root spellings — a symlinked `sessions` dir listed alongside its target — now yields two entries, trips the ambiguity refusal, and returns empty. I traced it through `mergePaths`, which only dedups lexically, so both spellings survive. The fix is your own rule applied at that one extra site: canonical key, first lexical path returned. I'm adding the cross-root regression test that would have caught it.

Merging with that on top. Credit stays yours — thanks for this one.


---

quad341:
@quad341 — the `canonical`/`root` split is the part of this that earns its keep. Keeping the resolved form strictly as the `visited` key, and letting path construction keep the caller's own spelling, is exactly right: `validateSearchPathFile` does lexical containment with no `EvalSymlinks`, so handing callers a silently-resolved root is precisely how a valid transcript ends up reported as "outside configured search paths." Cycle protection survives intact because the recursive call already passes resolved paths. Small, and it lands where the bug actually was.

Two things I'll take on my side. I reverted your two hunks locally and the whole `internal/sessionlog` suite — new broken-symlink test included — still passed, so the fix isn't pinned yet. I've written the assertion that does pin it (returned path must stay under a symlinked caller-provided root) and confirmed it fails without your change. I'll also trim the description at merge; it reads like a bigger branch than this one, and I'd rather history matched the code.

Separately: relative-vs-absolute spellings of the same root still collide into false ambiguity. That's pre-existing on `main`, not something you introduced — I'm filing it rather than growing this PR.

Merging with those on top. Credit stays yours.


---

quad341:
@quad341 — the `canonical`/`root` split is the whole fix and it lands exactly where the bug was. I traced it to be sure: `validateSearchPathFile` does lexical containment with no `EvalSymlinks`, so handing callers a silently-resolved root is precisely how a real transcript ends up rejected as "outside configured search paths." Keeping the resolved form strictly as the `visited` key preserves cycle protection while letting callers keep their own spelling. Six lines, right place.

Two things I'll take on my side, both ours. The fix isn't pinned yet — I reverted your hunks and the entire `internal/sessionlog` suite still passed, new broken-symlink test included, because `samePath` resolves both sides and can't see the difference. I've written the assertion that does pin it (pass the *symlink* as the search path, assert the result survives `validateSearchPathFile`) and confirmed it fails without your change. I'll also trim the description at merge — it describes a much larger branch than this one commit.

Separately, nested symlinked account roots still escape the caller's root. That's pre-existing on `main`; I'm filing it rather than growing this PR.

Merging with those on top. Credit stays yours.


---

quad341:
@quad341 — the `canonical`/`root` split is the whole fix and it lands exactly on the bug. I traced it rather than taking the description's word for it: `validateSearchPathFile` in `tail.go` does purely lexical containment with no `EvalSymlinks`, so a silently-resolved root is precisely how a real transcript ends up reported as "outside configured search paths." Keeping the resolved form strictly as the `visited` key preserves cycle protection while callers keep their own spelling. Six lines, right place.

Two things I'll take on my side. The fix isn't pinned yet — I reverted both hunks locally and the entire `internal/sessionlog` suite still passed, broken-symlink test included, because `samePath` resolves both sides and can't see the difference. I'll add the assertion that does pin it (pass the symlink itself as the search path, assert the result survives lexical containment), and I'll get the Mac tier actually running, since symlink spelling is the one thing Darwin does differently. I'll trim the description at merge too — it describes a larger branch than this commit.

Separately: transcripts under symlinked *child* roots still come back resolved. Pre-existing, not yours — filing it rather than growing this PR. Merging with those on top. Credit stays yours.


---

<details>

```diff
From 52f85a164c08832f0c5272a2e6483184c26fbae4 Mon Sep 17 00:00:00 2001
From: quad341 <quad341@users.noreply.github.com>
Date: Sun, 13 Sep 2026 13:48:57 -0700
Subject: [PATCH] test(sessionlog): pin caller root spelling for Kimi discovery
 (ga-w7a2xm)

The existing symlink tests assert with samePath, which calls EvalSymlinks
on both sides and so cannot distinguish a lexically-spelled return from a
canonicalized one -- exactly the distinction the root/canonical split in
findKimiSessionFilesInVisited and findKimiSessionFileByIDInVisited makes.

Pass the symlink itself as the search path (the recursion hands the
resolved path down, so a directory containing a symlink cannot exercise
this) and assert the discovered transcript is still spelled under the
caller's root, both by prefix and by feeding it back through
ExtractKimiTailMetaFromSearchPaths, whose containment check is lexical.
Reverting either hunk fails this test with the path canonicalized out
from under its own search root.
---
 internal/sessionlog/kimi_reader_test.go | 38 +++++++++++++++++++++++++
 1 file changed, 38 insertions(+)

diff --git a/internal/sessionlog/kimi_reader_test.go b/internal/sessionlog/kimi_reader_test.go
index b0412ec34..c05a6bec3 100644
--- a/internal/sessionlog/kimi_reader_test.go
+++ b/internal/sessionlog/kimi_reader_test.go
@@ -339,6 +339,44 @@ func TestFindKimiSessionFileSkipsBrokenSymlinkedRoot(t *testing.T) {
 	}
 }
 
+// The discovered path must stay spelled under the caller's own root. Callers
+// feed it straight back to ExtractKimiTailMetaFromSearchPaths, whose
+// containment check is lexical, so canonicalizing a symlinked search root away
+// during traversal makes the transcript this very search found read as
+// "outside configured search paths".
+func TestFindKimiSessionFileKeepsCallerRootSpelling(t *testing.T) {
+	isolateKimiSearchRoots(t)
+	base := t.TempDir()
+	accountRoot := t.TempDir()
+	workDir := "/tmp/gascity/phase1/kimi-alias"
+	workHash := kimiWorkDirHash(workDir)
+	writeKimiContext(t, filepath.Join(accountRoot, workHash, "session-key", "context.jsonl"), []string{
+		`{"role":"user","content":"via alias root"}`,
+	})
+
+	alias := filepath.Join(base, "account-alias")
+	if err := os.Symlink(accountRoot, alias); err != nil {
+		t.Skipf("symlink unsupported on this platform: %v", err)
+	}
+
+	aliasPrefix := alias + string(filepath.Separator)
+	got := FindKimiSessionFile([]string{alias}, workDir)
+	if !strings.HasPrefix(got, aliasPrefix) {
+		t.Fatalf("FindKimiSessionFile() = %q, want a path under the caller's root %q", got, alias)
+	}
+	if _, err := ExtractKimiTailMetaFromSearchPaths([]string{alias}, got); err != nil {
+		t.Fatalf("ExtractKimiTailMetaFromSearchPaths(%q) = %v, want the discovered path to validate against the root it was found under", got, err)
+	}
+
+	byID := FindKimiSessionFileByID([]string{alias}, workDir, "session-key")
+	if !strings.HasPrefix(byID, aliasPrefix) {
+		t.Fatalf("FindKimiSessionFileByID() = %q, want a path under the caller's root %q", byID, alias)
+	}
+	if _, err := ExtractKimiTailMetaFromSearchPaths([]string{alias}, byID); err != nil {
+		t.Fatalf("ExtractKimiTailMetaFromSearchPaths(%q) = %v, want the discovered path to validate against the root it was found under", byID, err)
+	}
+}
+
 // isolateKimiSearchRoots pins every ambient variable DefaultKimiSearchPaths
 // reads, so a discovery test sees only the roots it creates. Pinning HOME alone
 // is insufficient: KIMI_CODE_HOME contributes a root independently of HOME, and
-- 
2.55.0
```

</details>

PR #4936, opened 2026-08-03
Normalize canonical path ingest for formulas, workflows, and skills

quad341:
## What this changes

Formula description files, Git-backed formula sources, source-workflow lock identities, and skill materialization roots now use the same canonical path normalization at their ingest boundaries. Symlinked parents and not-yet-created path tails therefore resolve consistently before cache keys, lock keys, and containment comparisons are made.

This removes duplicated `filepath.EvalSymlinks` fallback logic at five sites while preserving existing validation and call-site contracts.

## Review notes

- The change is internal-only: no new configuration, endpoint, wire shape, or migration.
- `canonicalCityPath` still rejects empty paths, and `canonicalizePath` retains its existing `(string, error)` call-site contract.
- The main behavior to inspect is missing-leaf handling under symlinked parent directories.

## Test plan

- [x] Exercise formula, workflow-lock, and skill-materialization symlink/missing-tail regressions.
- [x] Run required process-backed `cmd/gc` and PR integration shards with CI-pinned bd/Dolt and tmux.
- [x] Run `go vet ./...` and `go build ./...`.
- [x] Release gate: [`release-gates/ga-sdcjgv-canonical-path-ingest-gate.md`](release-gates/ga-sdcjgv-canonical-path-ingest-gate.md)

### DIFF
--- a/internal/formula/parser.go
+++ b/internal/formula/parser.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 
 	"github.com/BurntSushi/toml"
+	"github.com/gastownhall/gascity/internal/pathutil"
 )
 
 // Formula file extensions. Canonical TOML is preferred, infixed TOML remains
@@ -206,10 +207,7 @@ func (p *Parser) parseResolvedAt(data []byte, absPath, label string) (*Formula, 
 }
 
 func descriptionFileBaseDir(path string) string {
-	if resolved, err := filepath.EvalSymlinks(path); err == nil {
-		return filepath.Dir(resolved)
-	}
-	return filepath.Dir(path)
+	return filepath.Dir(pathutil.NormalizePathForCompare(path))
 }
 
 // Parse parses a formula from JSON bytes.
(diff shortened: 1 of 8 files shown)


quad341:
@quad341 — this is the good kind of refactor: five hand-rolled `EvalSymlinks`-plus-fallback blocks collapsed onto `pathutil.NormalizePathForCompare`, deleting 45 lines from `skills.go` while *widening* correctness. Three of the old sites only tried the full path and one only the immediate parent, so multi-level missing tails now resolve where they previously didn't — and your `source_test.go` two-missing-levels case pins exactly the scenario the old code got wrong. Keeping the `(string, error)` shape on `canonicalizePath` with `//nolint:unparam` instead of churning all seven call sites was the right instinct too.

I traced one thing worth catching before this lands: `canonicalScopeRef` now runs `filepath.Abs` on everything, and `LockScopeForStoreRef` falls through to a literal `rig:<name>` when a rig isn't in config — so that sentinel becomes `<cwd>/rig:alpha` and the source-workflow lock key goes working-directory-dependent. I'm taking that on my side: a small sentinel guard plus a cwd-stability test, then re-running the sourceworkflow, formula, materialize, and pathutil packages. Your five sites and your tests stay as written.

One for later, not blocking: on macOS the `/private/var` → `/var` collapse flips the convention stored in the materialize ownership manifest, so the first post-upgrade pass won't match older entries. The Mac lane was skipped here, so I'll file it as a follow-up rather than guess. Thanks for the thorough gate doc — it made this easy to check.


---

PR #4938, opened 2026-08-03
fix(runtime): fail closed when T3 snapshot is unavailable

boshu2:
## Summary

- make `t3bridge.ListRunning` return `runtime.ErrRuntimeUnavailable` when its snapshot is transiently unavailable or still initializing
- keep total observation failure distinct from a partial-but-usable backend result
- add a regression test and reconcile the runtime-provider requirements ledger

## Problem

`ListRunning` currently converts a soft T3 bridge failure into `(nil, nil)`.
That result is indistinguishable from a healthy snapshot with zero sessions,
so callers that reason about absence can treat a transient outage as proof that
every T3 session disappeared.

This already affects existing fail-closed callers such as the adoption barrier
and dead-runtime cleanup. They defer on an error, but no error reaches them on
this path.

## Change

A soft-unavailable snapshot now returns no names and an error wrapping
`runtime.ErrRuntimeUnavailable`. This is a total observation failure, not a
`PartialListError`: there is no usable snapshot to preserve. The underlying
bridge error remains wrapped for diagnosis.

The test disables fallback endpoints, points the provider at a closed loopback
port, and asserts:

- the result is not empty success
- `errors.Is(err, runtime.ErrRuntimeUnavailable)`
- the error is not a `PartialListError`
- no names are returned with the failed observation

## Tests

```text
go test -count=1 ./internal/runtime/t3bridge
ok github.com/gastownhall/gascity/internal/runtime/t3bridge 34.905s

go test -count=1 ./internal/testpolicy/resourcecensus \
  -run '^TestRepositoryLedgerMatchesCensusAndDocumentation$'
ok github.com/gastownhall/gascity/internal/testpolicy/resourcecensus 1.649s

go vet ./internal/runtime/t3bridge
PASS
```

I also searched open and closed issues for `t3bridge ListRunning`,
`runtime unavailable`, and T3 snapshot outages; no duplicate surfaced.

## Design filters

- **Zero Framework Cognition:** this preserves the existing provider/error
  boundary; callers already defer on total listing errors.
- **Bitter Lesson alignment:** no heuristic is added. The provider reports the
  direct fact that it could not observe the runtime.


boshu2:
All required checks are green. I do not have permission to apply the repository’s review-queue label; could a maintainer route this through `status/needs-review-auto` when ready? Thanks.

quad341 (approved):
@boshu2 — this is a genuinely clean one. Returning `(nil, nil)` on a soft bridge outage meant "the bridge is down" and "every T3 session vanished" were the same value to the adoption barrier and the dead-runtime sweep, and you fixed it at the right layer with the right sentinel. Picking `ErrRuntimeUnavailable` over `PartialListError` is exactly what the doc comments on those two errors ask for, and it's the part people usually get backwards — there's no usable snapshot to preserve here, so it isn't a partial result.

Two things I especially appreciated. Your test is a real gate: I reverted the one-line change locally and it fails with your intended message, so it isn't passing vacuously. And you added the `RUNTIME-CONTRACT-006` row citing both source *and* test, which is what `REQUIREMENTS.md` asks for and almost nobody does unprompted.

I also traced the merge path — under `hybrid`/`auto`, a failing t3bridge backend now folds into `PartialListError` when the sibling has names, which is the correct outcome and handled everywhere.

Merging as-is. Nothing needed from you.

If you want another in the same vein: `IsRunning` still returns bare `false` on an unreachable snapshot outside the startup grace window. Same class of bug, but the `bool` signature makes it a caller-side change — rightly out of scope here.


---

PR #4943, opened 2026-08-03
fix(config): PackDirsForRig("") falls back to AllPackDirs for city-scope agents

quad341:
## Summary
- `PackDirsForRig("")` previously returned only city-level `PackDirs`, silently dropping every `RigPackDirs` entry for rig-less (`scope="city"`) agents (deep-investigator, supervisor, pack-author).
- Empty `rigName` now delegates to the existing `AllPackDirs()` (union across all rigs, sorted by rig name for determinism) — the same fallback already used for global scans.
- Doc comments on both functions updated to explain the fallback and the accepted alphabetical-collision risk (a fragment with different content in two rigs' packs silently prefers whichever rig sorts first).
- `cmd/gc/cmd_prime.go:343` and `cmd/gc/template_resolve.go:347` call `PackDirsForRig` and inherit the fix automatically — no changes needed there.

This is the architect-decided fix (Option 2) for ga-bmjqvb, split into ga-bmjqvb.1 (this PR) and ga-bmjqvb.2 (follow-up pack-author cleanup, blocked on this landing).

## Test plan
- [x] `go test ./internal/config/... -run 'TestPackDirsForRig|TestAllPackDirs' -v` — new `TestPackDirsForRigEmptyRigNameFallsBackToAllPackDirs` passes alongside existing cases
- [x] `go test ./cmd/gc/... -run 'TestRenderPromptResolvesMultiRigPackFragments' -v` — new `TestRenderPromptResolvesMultiRigPackFragmentsForCityScopeAgent` passes, pinning the fix at the actual `renderPrompt` call-site
- [x] `go vet ./...` clean
- [x] `make test-fast-parallel` — all 10 fast jobs pass

Fixes ga-bmjqvb.1

### DIFF
--- a/cmd/gc/prompt_test.go
+++ b/cmd/gc/prompt_test.go
@@ -1481,6 +1481,34 @@ func TestRenderPromptResolvesMultiRigPackFragmentsForCityScopeAgent(t *testing.T
 	}
 }
 
+// TestRenderPromptCityScopeFragmentCollisionLastRigWins pins the *direction* of
+// a same-named fragment collision across rigs. PackDirsForRig("") returns rig
+// dirs sorted by rig name and renderPrompt parses them in order, so a later
+// {{ define }} replaces an earlier one: the alphabetically last rig wins. The
+// PackDirsForRig doc comment documents this; without this test a change to
+// pack-dir ordering or loadSharedTemplates override semantics would flip the
+// winner silently.
+func TestRenderPromptCityScopeFragmentCollisionLastRigWins(t *testing.T) {
+	f := fsys.NewFake()
+	f.Files["/a/template-fragments/x.template.md"] = []byte(
+		`{{ define "x" }}FROM-ALPHA{{ end }}`)
+	f.Files["/z/template-fragments/x.template.md"] = []byte(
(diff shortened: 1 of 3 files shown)


quad341:
@quad341 — nice catch. This is the kind of bug that only ever surfaces as "why can't deep-investigator see the fragment everyone else sees," and routing empty `rigName` through the existing `AllPackDirs()` is exactly the right shape: no new machinery, and both call sites (`cmd_prime.go`, `template_resolve.go`) pick it up for free. I confirmed those really are the only two.

The second test is what sold me — pinning it at the actual `renderPrompt` call site rather than only at the config unit is what makes this stick, and it mirrors how the symptom was reported.

One thing I'll take on my side before merging: the new doc comment says a cross-rig fragment collision resolves to whichever rig sorts *first* alphabetically. It's whichever sorts *last* — pack dirs are parsed in order and a later `{{ define }}` replaces an earlier one, so `alpha` vs `zulu`, `zulu` wins. Comment-only, no logic change, and I'll add a small test pinning that direction plus re-run the config and prompt tests.

Merging with that tweak. Thanks for splitting the pack-author cleanup into ga-bmjqvb.2 — keeping this one to the actual fix made it easy to verify.


---

PR #4944, opened 2026-08-03
session: honor configured provider start lease over generic creating-state timeout

### LINKED ISSUE #4858: bug: active provider start lease is overwritten by one-minute stale creating projection
## Before you continue

## Gas City version

1.3.5 (`v1.3.5`, commit `8ffc009de`). The same control flow is present on current `main` at `97503b283`.

## Environment

Linux, tmux session provider. Reproduced deterministically in the Go reconciler tests.

## Reproduction

1. Configure `[session] startup_timeout = "5m"`.
2. Construct or observe an open session bead with:
   - `state=creating`
   - `pending_create_claim=true`
   - valid `pending_create_started_at`
   - valid `last_woke_at`
   - no observable runtime yet because the provider `Start()` call is still within its configured five-minute deadline.
3. Advance the reconciler clock to 90 seconds after `last_woke_at` and run the advisory-state heal.
4. Observe that the generic one-minute stale-creating projection changes `state` to `asleep` even though `pendingCreateStartInFlight(..., 5m)` is true.
5. On the next rollback decision, the `state == asleep` branch checks the one-minute attempt age before checking the configured in-flight lease and can roll back the still-valid start.

A deterministic regression test uses a five-minute startup timeout and asserts no mutation at 90s and 306s, one rollback after the 307s boundary (`5m + staleKeyDetectDelay + 5s`), and no second mutation.

## Expected behavior

The configured provider-start lease is authoritative while active. A pending create with a valid `last_woke_at` must remain `creating` with its claim intact until the configured startup deadline plus the existing observation grace expires. All state-specific rollback paths should use the same lease decision.

## Actual behavior

`staleCreatingStateTimeout` is a separate hard-coded one-minute constant. `healStatePatchWithRollback` passes it to lifecycle projection as `StaleCreatingAfter`, so a legitimate slow start projects to `asleep` after 60 seconds. `pendingCreateLeaseExpiredForRollback` then handles `state=asleep` before consulting `pendingCreateStartInFlight`, allowing the one-minute projection to bypass a longer configured `session.startup_timeout`.

The only available configuration, `session.startup_timeout`, controls the provider deadline and the in-flight lease. It does not configure or disable `staleCreatingStateTimeout`, so there is no configuration-only workaround for legitimate starts that take more than one minute.

## Logs, screenshots, or traces

```shell
# With startup_timeout=5m and last_woke_at 90s ago:
pendingCreateLeaseExpiredForRollback(...) == false
healStatePatchWithRollback(...)["state"] == "asleep"  # unexpected

# Intended boundary with the existing 2s stale-key delay and 5s grace:
90s:  no mutation
306s: no mutation
308s: exactly one rollback
```

## Additional context

Related but not duplicate: #1460 covered a pending-create claim that never expired. This report covers the reciprocal failure: a valid configured start lease is terminated prematurely by the generic one-minute projection.

A minimal reviewed patch and regression are available at `loucmane/gascity@62a03b707`:
https://github.com/loucmane/gascity/commit/62a03b707

The patch:

- suppresses only the premature terminal `creating -> asleep` projection while the configured lease is active;
- moves the in-flight lease check ahead of the `asleep` state shortcut so all rollback paths share one decision;
- retains the one-minute generic stale threshold for corrupt or claimless state; and
- retains rollback immediately after the configured lease boundary.

loucmane:
## Summary

- honor the configured provider start lease before state-specific rollback
- keep session state and the pending-create claim together until one rollback decision
- add regressions covering creating/asleep projections before and after lease expiry

## Root cause

The lifecycle projection can apply the generic one-minute stale-creating timeout before a provider's configured start lease expires. A subsequent state-specific reconciliation path can then roll back the pending-create claim early, even though provider startup is still legitimately in flight.

## Fix

Evaluate the pending-create lease before every state-specific rollback path. While the configured lease is active, preserve both session state and claim; once it expires, perform the rollback once and keep the two projections consistent.

Fixes #4858

## Testing

- regression covering an active configured lease across multiple reconciliation ticks
- regression covering the single rollback after lease expiry
- existing focused reconciler tests


quad341 (approved):
@loucmane — quick status, and none of it needs anything from you.

Everything I said in the review stands. What makes this land isn't the reordering, it's that you noticed the transition was being applied *half at a time*: state moved to `asleep` while the pending-create claim stayed behind, and nothing in the old code made those two travel together. Hoisting the in-flight check above the `asleep` branch and collapsing the heal path onto a single rollback decision fixes the shape of the bug rather than its symptom. `TestHealStatePatchWithRollbackUsesOneConfiguredLeaseDecisionAcrossTicks` is the right test for it — driving ticks across the 307s boundary and then asserting a *second* post-expiry heal is a no-op pins the idempotence the old path never had.

I started the rebase and hit a wall on my end — one of our automation accounts ran out its weekly budget mid-pass, so the merge got as far as identifying the conflict and then stopped. Entirely our infrastructure, nothing to do with your branch.

The useful news from that partial run: the conflict with `main` is confined to a single region of `cmd/gc/session_reconcile.go`. Both of your new tests merge clean, which means I can use them as the correctness check on my own conflict resolution — that's a much better position to be in than if they'd conflicted too.

I'm also pulling the reference-oracle cleanup I mentioned out of this PR. That's the stale `refPendingCreateLeaseExpiredForRollback` in `session_classifier_info_equiv_test.go` — it's out-of-diff, it's green today, and it's my code, not yours. It'll go in as its own follow-up rather than holding your work behind it.

Picking the rebase back up shortly, with your authorship intact. Thanks for your patience with our plumbing.

PR #4946, opened 2026-08-03
fix(doctor): order-firing-current timeout is advisory, not blocking (#4895)

### LINKED ISSUE #4895: bug: gc doctor order-firing-current hard-fails (blocking) on a 15s history-lookup timeout even when orders are firing
## Summary

`gc doctor`'s `order-firing-current` check races its order-history lookup against
a **hardcoded 15s budget** and, on timeout, returns a **blocking `StatusError`** —
even when scheduled orders are firing perfectly well. The timeout branch conflates
"the diagnostic query is slow" with "orders are stale", turning a slow-but-healthy
city red and blocking `gc doctor` (exit non-zero) on operators who have no actual
execution gap.

## Where

`internal/doctor/checks_order_firing.go`:

```go
const orderFiringHistoryTimeout = 15 * time.Second   // hardcoded

func (c *OrderFiringCurrentCheck) Run(ctx *CheckContext) *CheckResult {
    ...
    select {
    case result := <-results:
        return result
    case <-time.After(timeout):
        return &CheckResult{
            Name:    c.Name(),
            Status:  StatusError,   // <-- no Severity => BLOCKING
            Message: fmt.Sprintf("order history lookup timed out after %s", timeout),
            FixHint: "check beads/Dolt connectivity, then rerun gc doctor",
        }
    }
}
```

The inner `run()` already computes **nuanced severity** — it separates advisory
errors from blocking ones (`blockingErrors`/`advisoryErrors`, and
`classifyOrderFiring` marks cron-never-fired as advisory while cooldown gaps stay
blocking). The timeout branch throws all of that away and emits an unconditional
blocking failure.

## Evidence (production city, ~8 rigs, many housekeeping orders)

- `gc order history` (the Dolt-backed lookup the check depends on) measured **7.6s**
  standalone. Under a full `gc doctor` run (DoltTopology/DoltDrift checks hit Dolt
  first), the check's per-order `LastRunAcross` fallbacks push total latency past
  the 15s budget → blocking ✗.
- Orders are demonstrably firing: **34,839 `order.fired` events** in `events.jsonl`,
  and operator-observed executions immediately after controller start.
- The `events.jsonl` scan is **not** the bottleneck: filtering both event types
  (`order.fired` + `controller.started`) over a **151 MB** log measured **0.14s**.
  The entire cost is the Dolt order-history path.
- `order-tracking-retention` advisory: **355 closed order-tracking beads** (OK).
  Because `sweep-tracking` always retains a floor of **≥10 closed beads per order**
  and there are ~30 distinct scoped orders, tuning
  `[beads.policies.order_tracking].delete_after_close` only partially shrinks the
  working set — which is exactly why the check itself needs to degrade gracefully
  rather than depend on operators keeping the table tiny.

Net: a hardcoded 15s budget is below the p100 latency of a legitimate order-history
lookup on a busy long-lived city, so this check produces a **persistent false ✗**.

## Proposed fixes (any subset)

1. **Timeout should be non-blocking.** On timeout, return `StatusWarning` (or
   `StatusError` with `Severity: SeverityAdvisory`) with a message that says the
   lookup is slow and the result is inconclusive — *not* proof of an execution gap.
   A diagnostic that cannot complete in time must not gate `gc doctor` red.
2. **Make the budget configurable / honor `--check-timeout`.** `gc doctor` already
   exposes `--check-timeout` (default 60s). The separate hardcoded 15s is *stricter*
   than the check-level budget, which is surprising. Either honor `--check-timeout`
   or add an explicit knob.
3. **Cut the lookup cost.** The `order.fired` events are already loaded cheaply
   (~0.1s) and are authoritative for "is this order firing recently". Prefer that
   signal and reserve the per-order Dolt `LastRunAcross` fallback for orders with no
   recent event — ideally as a single batched query rather than one store-open per
   order.

## Impact

Any healthy, busy, long-uptime city eventually trips this: `gc doctor` goes red on a
blocking check while orders fire normally, eroding trust in `doctor` as a gate and
causing operators to chase a non-existent Dolt/beads outage (the current `FixHint`
points them at connectivity, which is a dead end here).

## Environment

- `gc doctor` `order-firing-current` check, `internal/doctor/checks_order_firing.go`
- Related: #1118 (events.jsonl unbounded growth) — adjacent, but confirmed *not*
  the cause of this particular timeout.

_Filed from operator triage (gas-city rig, bead gci-pnt)._

rjgeng:
## Summary
Fixes #4895: `gc doctor`'s `order-firing-current` check races its Dolt-backed order-history lookup against a hardcoded 15s budget. On timeout it returned `StatusError` with no `Severity` set, silently defaulting to `SeverityBlocking` (the zero value of `CheckSeverity`) — gating `gc doctor`'s exit code red on a merely-slow-but-inconclusive lookup, even when scheduled orders are firing normally. A timed-out lookup proves nothing about actual order staleness.

## Fix
`OrderFiringCurrentCheck.Run`'s timeout branch now explicitly sets `Severity: SeverityAdvisory` and `TimedOut: true`, matching the pattern `Doctor.boundedRun`'s own per-check timeout already uses. This is the smallest of the issue's three proposed fixes ("timeout should be non-blocking"); the hardcoded-vs-`--check-timeout` mismatch and the lookup-cost optimization it also raises are separate, real asks left for follow-up work.

## Test plan
- [x] Extended `TestOrderFiringCurrent_TimesOutStalledOrderHistory` with assertions on `Severity`/`TimedOut` — confirmed red (`severity = 0`, i.e. `SeverityBlocking`) before the fix, green after.
- [x] Full `internal/doctor` suite green, `go build ./...` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   always locks to `sha:<HEAD commit>`, matching the documented behavior;
   registry/remote sources are unaffected. Fixes #3659.
 
+- **`gc doctor`'s `order-firing-current` check no longer hard-fails
+  `gc doctor` (exit code, `BlockingFailed`) when its own order-history
+  lookup times out.** The check races a Dolt-backed order-history query
+  against a 15s budget; on timeout it returned `StatusError` with no
+  `Severity` set, silently defaulting to `SeverityBlocking` (the zero
+  value of `CheckSeverity`) — turning a slow-but-healthy city's doctor run
+  red even when scheduled orders were firing normally, since a timed-out
+  lookup proves nothing about actual order staleness. The timeout branch
+  now explicitly sets `Severity: SeverityAdvisory` and `TimedOut: true`,
+  matching how `Doctor.boundedRun`'s own per-check timeout is already
+  reported, so callers (including `--json` output) can distinguish
+  "confirmed stale" from "the query didn't finish in time." (#4895)
+
 - **The dolt pack's `run_bounded` python3 fallback now sends SIGTERM before
   SIGKILL, matching its documented contract.** The fallback (used when
   neither `timeout` nor `gtimeout` is on `PATH`, the default on stock macOS)
--- a/internal/doctor/checks_order_firing.go
+++ b/internal/doctor/checks_order_firing.go
@@ -118,11 +118,20 @@ func (c *OrderFiringCurrentCheck) Run(ctx *CheckContext) *CheckResult {
 	case result := <-results:
 		return result
 	case <-time.After(timeout):
+		// A timed-out lookup is inconclusive, not proof of a stale/never-fired
+		// order (#4895): the query being slow says nothing about whether orders
+		// are actually firing. SeverityBlocking is CheckSeverity's zero value, so
+		// leaving Severity unset here would silently gate `gc doctor` (and its
+		// exit code) red on a busy-but-healthy city. Mark it advisory and
+		// TimedOut, matching how Doctor.boundedRun reports its own per-check
+		// timeout, so callers can tell "confirmed stale" from "couldn't tell".
 		return &CheckResult{
-			Name:    c.Name(),
-			Status:  StatusError,
-			Message: fmt.Sprintf("order history lookup timed out after %s", timeout),
-			FixHint: orderFiringTimeoutHint,
+			Name:     c.Name(),
+			Status:   StatusError,
+			Severity: SeverityAdvisory,
+			TimedOut: true,
+			Message:  fmt.Sprintf("order history lookup timed out after %s", timeout),
+			FixHint:  orderFiringTimeoutHint,
 		}
 	}
 }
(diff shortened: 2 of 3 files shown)


quad341 (approved):
@rjgeng — the diagnosis here is the good part. Spotting that `SeverityBlocking` is `CheckSeverity`'s *zero value*, and that an unset `Severity` on the timeout branch was therefore silently red-gating `gc doctor`'s exit code, is the kind of bug that hides in plain sight for a long time. "A timed-out lookup proves nothing about actual order staleness" is exactly the right framing, and matching `boundedRun`'s existing `SeverityAdvisory` + `TimedOut` shape makes the fix read like it was always there.

I appreciate that you extended the existing test with assertions that fail against the old code instead of adding a parallel one, and that you confirmed red-before/green-after rather than just asserting it. Keeping this to the one fix — leaving the hardcoded-15s-vs-`--check-timeout` mismatch and the lookup-cost work from #4895 for follow-ups — made it much easier to say yes to.

I verified on my side: traced the exit-code path through `tally` → `BlockingFailed`, and ran the doctor tests locally (the CI shard that covers `internal/doctor` was skipped on this run, so I didn't want to take the green rollup on faith). Merging as-is. If you pick up either remaining #4895 ask, I'd be glad to review.


---

Issue #3887, opened 2026-07-02
workspace-identity deprecation hint can silently break Discord (and other city.toml workspace.name consumers) if followed before pack pins catch up


Wldc4rd:
The v2 deprecation warning tells operators to move workspace identity out of `city.toml`:

```
city.toml: workspace identity fields are deprecated in v2; move them to .gc/site.toml (workspace.name); … (run `gc doctor --fix` if this is the root city.toml …)
```

Following it — by hand or via `gc doctor --fix` — is a trap for any city that pins a pack release which still resolves the supervisor scope from `city.toml` `workspace.name` only (gascity-packs discord before 26cfc61). With the declared name stripped, scope discovery comes back empty, delivery falls back to the legacy default API port, and **all inbound Discord messages die with connection refused — silently** (the gateway stays up; only per-message ingress receipts record the failures).

This isn't hypothetical: on one deployment the name was removed from a root `city.toml` during an unrelated edit, and inbound DMs were eaten for ~2 days with zero alarms before the receipts were checked.

Two asks, either/both:

1. **Caveat the hint** (doctor + config load warning): note that installed packs pinned before the site-binding-aware releases may still read `city.toml` `workspace.name`, so operators should update/re-pin those packs before stripping the field. A one-line addition to the warning text would have prevented the incident above.
2. **Sequence-aware `doctor --fix`**: before stripping workspace identity from a root `city.toml`, detect installed packs whose pinned revision predates their site-binding support (even just a hardcoded floor for the known first-party packs) and either warn or skip the strip for that field.

Related: gascity-packs 26cfc61 taught the discord pack to read `.gc/site.toml`, and follow-up PR gastownhall/gascity-packs#152 fixes its precedence to match `applyWorkspaceIdentityBinding()` (site binding wins over the declared name). Once cities re-pin, the trap closes for the discord pack specifically — the hint remains risky for any other consumer of the deprecated field.

PR #4947, opened 2026-08-03
fix(config): caveat pack compatibility risk in workspace-identity deprecation warning (#3887)

rjgeng:
## Summary
Addresses the first of #3887's two asks: the workspace-identity deprecation hint told operators to remove `workspace.name` (and related fields) from `city.toml` with no signal that a pack pinned to an older revision might still read those fields directly instead of `.gc/site.toml` — following the hint alone could silently break routing/identity for a consumer like Discord.

## Fix
`internal/config/site_binding.go`'s `legacyWorkspaceIdentitySurfaceWarnings` now appends a caveat sentence to the existing deprecation message, kept pack-agnostic (no hardcoded pack names/versions):

> "... Some installed packs pinned to an older revision may still read these fields directly instead of `.gc/site.toml` -- confirm pack compatibility before removing them, or routing/identity that depends on workspace identity may silently stop working"

## Scope note
#3887's second ask — a sequence-aware `doctor --fix` that detects installed packs' pinned revisions and skips/warns per-pack — would require this core config package to hardcode knowledge of specific downstream packs' version histories, which is a layering violation per this repo's own core/pack boundary rules (`AGENTS.md`: "Do not bury... pack-specific assumptions in generic SDK paths"). That needs either a maintainer-designed pack-declared-compat mechanism or an explicit decision to accept the layering cost — not attempted here.

## Test plan
- [x] New test `TestLegacyWorkspaceIdentityWarningCaveatsPackCompatibility` — confirmed red before the fix, green after.
- [x] Full existing warning-assertion set green (`TestLegacySiteBindingSurfaceErrorAggregatesViolations`, `TestLoadWithIncludes_WarnsOnLegacyWorkspaceIdentityInSchema2City`, `TestLoadWithIncludes_WarnsOnLegacyWorkspaceIdentityInSchema2Fragments`, `TestLoadWithIncludes_AppliesWorkspaceIdentitySiteBinding`) — all match on `strings.Contains`, none broke on the appended text.
- [x] Full `internal/config` suite green, `go build ./...` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   dry-run default, where they are the only backend cost) so the scan never
   does unbounded reads per tick. Fixes #3780.
 
+- **The legacy workspace-identity deprecation warning now caveats that
+  following it can silently break packs pinned to an older revision.**
+  `city.toml`'s `workspace.name`/`workspace.prefix` deprecation hint told
+  operators to move those fields to `.gc/site.toml`, but any installed pack
+  still pinned to a revision that reads `workspace.name` directly (rather
+  than the newer site-binding-aware resolution) would silently lose its
+  identity/routing once the field was removed — reported after one
+  deployment lost inbound Discord messages for ~2 days with zero alarms
+  before anyone checked delivery receipts. The warning now says so
+  explicitly, so operators check pack compatibility before acting on it.
+  (#3887)
+
 - **The dolt pack's `run_bounded` python3 fallback now sends SIGTERM before
   SIGKILL, matching its documented contract.** The fallback (used when
   neither `timeout` nor `gtimeout` is on `PATH`, the default on stock macOS)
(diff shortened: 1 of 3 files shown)


quad341 (approved):
@rjgeng — the part of this I appreciate most isn't the string append, it's the scope note. You looked at #3887's second ask, worked out that a sequence-aware `doctor --fix` would mean baking specific packs' version histories into `internal/config`, checked that against `AGENTS.md`'s core/pack boundary, and *stopped* rather than shipping a layering violation to close an issue completely. Writing down why you stopped is what makes it useful to whoever designs the pack-declared-compat mechanism later.

The caveat itself is well-aimed: pack-agnostic, and naming `gc doctor --fix` specifically — the command that actually does the removal — is what makes the warning actionable instead of vague. Your test pins both halves without making the existing `strings.Contains` assertions brittle.

Two things on our side, neither needing anything from you. The red `cmd/gc process / shard 10 of 12` isn't yours — it's `bd init` hitting a Dolt schema migration (beads#4566) in a test that never loads config. And I'll swap the ASCII `--` for an em dash to match the other warnings in that package while I'm in there. Merging once CI is green. Thanks for this.


---

PR #4948, opened 2026-08-03
Normalize configured city and rig paths at ingest

quad341:
## What this changes

Gas City now canonicalizes explicit city and rig paths as they enter `gc`. Paths supplied through `--city`, `GC_CITY`/`GC_CITY_PATH`/`GC_CITY_ROOT`, positional city or rig arguments, and the rig-scoped `GC_RIG_ROOT`/`BEADS_DIR` environment now resolve symlinked ancestors to the same absolute path used by discovery.

This prevents one physical city or rig from acquiring different string identities depending on whether an operator uses its real path or a symlink alias. Missing leaf paths retain the existing longest-ancestor normalization behavior. CLI flags, environment-variable contracts, configuration schemas, and ordinary non-symlinked paths are unchanged.

## Review notes

- The implementation reuses the existing `normalizePathForCompare` boundary helper; it does not introduce another canonicalization mechanism.
- The production diff is limited to three ingest points in `cmd/gc/main.go` and `cmd/gc/bd_env.go`, with focused regressions alongside them.
- `--rig` and `city.toml` paths already converge through their normalized registry/config paths, so those flows are intentionally unchanged.
- There is no migration, new default, endpoint, dependency, or wire-format change.

## Test plan

- [x] Focused symlink, relative-input, missing-leaf, precedence, and contextual-error contracts: 9 pass, 0 fail, 0 skip
- [x] `make test-fast-parallel`: 10/10 jobs pass
- [x] Non-short `cmd/gc` process lane with the CI-pinned `bd`: 15,362 pass, 0 fail; 11 documented helper/platform/opt-in skips
- [x] Worker phase-2 conformance for Claude, Codex, and Gemini: 78/78 requirements pass
- [x] `go build ./...` and `go vet ./...`
- [x] Release gate: [`release-gates/ga-jx0gqf-normalize-configured-paths-gate.md`](release-gates/ga-jx0gqf-normalize-configured-paths-gate.md)

### DIFF
--- a/cmd/gc/bd_env.go
+++ b/cmd/gc/bd_env.go
@@ -1307,7 +1307,7 @@ func bdRuntimeEnvForRigWithErrorRecovery(cityPath string, cfg *config.City, rigP
 
 func bdRuntimeEnvForRigWithErrorRecoveryContext(ctx context.Context, cityPath string, cfg *config.City, rigPath string, allowRecovery bool) (map[string]string, error) {
 	env, cityErr := bdRuntimeEnvWithErrorRecoveryContext(ctx, cityPath, allowRecovery)
-	rigPath = filepath.Clean(rigPath)
+	rigPath = normalizePathForCompare(rigPath)
 	// Pin the rig store explicitly. The gc-beads-bd provider derives its Dolt
 	// data root from GC_CITY_PATH unless BEADS_DIR is set, so cwd-based
 	// discovery is not sufficient for rig-scoped operations.
--- a/cmd/gc/bd_env_test.go
+++ b/cmd/gc/bd_env_test.go
@@ -351,6 +351,41 @@ func TestBdRuntimeEnvNoRecoveryMatchesRecoveryForExternalTarget(t *testing.T) {
 	}
 }
 
+// TestBdRuntimeEnvForRigResolvesSymlinkAlias pins ga-iawy13.8: GC_RIG_ROOT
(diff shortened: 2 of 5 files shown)


quad341:
@quad341 — this is the good kind of small. Replacing three one-off `filepath.Abs`/`Clean` calls with the existing `normalizePathForCompare` instead of inventing a fourth canonicalization path is exactly right, and it lines the ingest boundary up with what `supervisor.resolveAbsPath` and `resolveStoreScopeRoot` were already doing on the other side of the seam.

Special credit for the tests comparing with raw `==` rather than `samePath`, and for saying why in the comment. `samePath` normalizes both sides, so the obvious version of these tests would have passed against the un-normalized code and pinned nothing. That's the detail that makes them real regressions.

Merging as-is. I checked the two things that could have bitten: the dropped `Abs` errors only fire when `os.Getwd()` fails, and the path still fails the city-config gate in that case; and the empty-string change is safe across every caller. CI is green across all twelve `cmd/gc` shards and the integration lanes.

One follow-up, not this PR: the note that `city.toml` paths already converge holds for the registry side, but `workdir.RigRootForName` still hands `rig.Path` back verbatim into the agent's `GC_RIG_ROOT`. Nothing breaks — the restart fingerprint reads the template-side env, and everything downstream uses `samePath` or just opens the path — but a symlink-aliased rig still carries two string identities across the agent/bd seam. Might be worth one more increment.


---

PR #4952, opened 2026-08-04
Prevent held work from re-entering automatic dispatch

quad341:
## What this changes

Automatic routed dispatch now treats the two canonical hold labels as stop signs across pool-demand queries and control-dispatcher ready scans. Held work remains persisted and observable, but an unassigned agent will not automatically claim it while the hold is present.

Deliberately assigned work remains available to its assignee. Crash and boot recovery also remain hold-transparent: recovery can reopen persisted work without stripping its hold, while a later ambient route scan still filters that work out.

## Why this design

The implementation keeps recovery and explicit assignment separate from automatic routing. A shared typed list in `internal/beadmeta` drives the modern beads CLI queries, legacy compatibility filters, and in-process control-ready evaluation so those paths cannot drift independently.

## Review notes

- Review the boundary between assignee-scoped Tier 1/2 paths and unassigned, route-scoped Tier 3 paths; only the latter filter holds.
- The change covers both current `bd ready --exclude-label` behavior and the legacy ephemeral jq path.
- There is no new configuration, schema migration, dependency, or user action required.

## Test plan

- [x] `make test-fast-parallel` — all 10 jobs pass.
- [x] Touched-package run — 17,169 PASS, 0 FAIL; focused hold/recovery tests — 25 PASS, 0 FAIL, 0 SKIP.
- [x] `go vet ./...` and `go build ./...`.
- [x] Release gate: [`release-gates/ga-9sp6gf-held-work-dispatch-gate.md`](release-gates/ga-9sp6gf-held-work-dispatch-gate.md)


quad341:
@quad341 — merging this now. The part that earns it is the boundary work: routing every path through one `beadmeta.DispatchHoldLabels` list so the bd-native query, the legacy jq filter, and the in-process Go evaluation can't drift, and then pinning that boundary with *negative* tests. Asserting that assignee-scoped Tier 1/2 and `ephemeralAssignedReadyProbeScript` must stay hold-transparent is what will stop someone from "fixing" those later.

Two things we checked ourselves rather than take on trust, both of which held: `bd ready --exclude-label` really is a repeatable flag at v1.0.4 (so the default compatibility path is safe — the contract job doesn't exercise it, so its green wasn't evidence), and `PrimeActive` doesn't set `SkipLabels`, so the cached beads `filterReadyByRoute` reads do carry labels. Covering `poolDemandCountShell` too was the right instinct — otherwise the reconciler would scale a pool up for work nobody may claim.

One follow-up worth a bead, nothing to change here: the reconcile path scans with `SkipLabels: true`. The control-ready cache never reconciles today, but if it ever does, this filter degrades to a silent no-op — every test hands it beads with labels already set. A test at that seam would catch it.

And the `ShellFallbackUnreachable` comment admitting the shell assertions guard dead code is exactly the honesty that saves the next reader an hour.


---

PR #4953, opened 2026-08-04
fix: use portable PID liveness in drift restart

### LINKED ISSUE #4942: bug: pidGone treats a live PID as gone on macOS
## Before you continue

- [x] I searched open and closed issues for `pidGone`, `waitForPIDExit`, and
  `/proc` on Darwin; no duplicate surfaced.
- [x] I read the contributor guidance.

## Gas City version

Found by code reading against `main` at `1111964` while investigating the stop
path on v1.4.0 (`a7297c511`).

## Environment

macOS (darwin 25.5.0, arm64). The defect is platform-conditional and affects
non-Linux systems without `/proc`.

## Reproduction

`cmd/gc/cmd_start_drift.go` has no build tag. After a signal-zero probe,
`pidGone` reads `/proc/<pid>/status` and does this on read failure:

```go
return os.IsNotExist(err)
```

On macOS `/proc` does not exist, so this is true for every PID, including a live
one:

```go
_, err := os.ReadFile("/proc/1/status")
fmt.Println(os.IsNotExist(err)) // true on macOS; PID 1 is alive
```

## Expected behavior

On a platform without `/proc`, failure to read the Linux status file should mean
"zombie state unavailable," not "the PID is gone." The signal-zero result
should remain authoritative, or the `/proc` zombie probe should live behind a
Linux-specific implementation.

## Actual behavior

For a live process on macOS, signal zero succeeds and then the missing `/proc`
path makes `pidGone` return true. `waitForPIDExit` can therefore return
immediately in the drift auto-restart path while the supervisor process is
still alive.

## Additional context

- #4855 tracks the broader class of duplicated process introspection with
  platform gaps, but does not list this `cmd_start_drift.go` site.
- #4521 is a sibling macOS `/proc/<pid>/exe` problem. Its proposed executable
  path fix does not address this silent boolean inversion.

A small fix would gate the `/proc` fallback on Linux or move it behind
platform-specific helper files, with a non-Linux regression test.

boshu2:
## Summary

- replace the drift-restart path's Linux-specific `/proc` PID check with the shared portable `pidutil.Alive` probe
- add a Darwin regression proving a live process is not reported gone before restart proceeds

Fixes #4942.

## Testing

- [x] `go test -count=1 ./cmd/gc -run '^TestPIDGoneReturnsFalseForCurrentProcess$'`
- [x] `go test -count=1 ./cmd/gc -run '^(TestPIDGoneReturnsFalseForCurrentProcess|TestRunStartDriftCheck_DarwinLaunchdRestartDoesNotRequireProcExe)$'`
- [x] `go test -count=1 ./internal/pidutil`
- [x] `go test -count=1 ./internal/testpolicy/resourcecensus -run '^TestRepositoryLedgerMatchesCensusAndDocumentation$'`
- [x] `make test-cmd-gc-process-shard CMD_GC_PROCESS_SHARD=1 CMD_GC_PROCESS_TOTAL=6` (1,379 tests)
- [x] `go vet ./...`
- [ ] `make check` — not repeated after the affected process shard; exact `origin/main` currently reproduces unrelated macOS path-canonicalization failures on this host
- [ ] `make check-docs` — not applicable; no docs, navigation, or links changed
- [ ] `make test-integration` — not run locally; this change is covered by the process-backed `cmd/gc` shard and full PR CI

## Checklist

- [x] Linked an issue
- [x] Added a regression test for the behavior change
- [x] No user-facing docs change is required

### DIFF
--- a/cmd/gc/cmd_start_drift.go
+++ b/cmd/gc/cmd_start_drift.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/gastownhall/gascity/internal/config"
 	"github.com/gastownhall/gascity/internal/fsys"
+	"github.com/gastownhall/gascity/internal/pidutil"
 )
 
 // driftFlags captures the operator-visible inputs that influence drift
@@ -526,7 +527,7 @@ var (
 
 // waitForPIDExit blocks until the process at pid is gone, escalating
 // to SIGKILL if SIGTERM did not take effect within timeout. Returns
-// nil once the kernel reports ESRCH on a signal-zero probe.
+// nil once the shared PID probe reports no live process.
 //
 // PID-recycling races are not addressed here — the window between
 // SIGTERM and SIGKILL is short enough (seconds) that a recycled PID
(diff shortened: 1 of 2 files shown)


boshu2:
All checks are green, the exact candidate has an independent review PASS, and this is ready for maintainer review when convenient.

quad341 (approved):
@boshu2 — you found the exact shape of this one. The old `pidGone` treated a missing `/proc` as proof the process had exited (`return os.IsNotExist(err)`), so on macOS `waitForPIDExit` returned instantly and the drift restart raced a supervisor that was still holding its port. Swapping to `pidutil.Alive` is the right fix, and I traced it myself to confirm it preserves what the deleted `/proc/<pid>/status` fallback was there for — the zombie case, via the `ps` fallback off Linux, plus the EPERM path when the supervisor runs as another uid.

Deleting 30 lines of bespoke probe in favor of the helper a dozen other call sites already use is a straight win too.

Merging as-is. Two things I'm filing separately rather than asking you to carry: `pidutil.Alive`'s zombie check parses `/proc/stat` positionally and would misread a comm containing a space (its sibling `StartTime` anchors on the last paren — `Alive` should too), and this path could eventually use `AliveWithStartTime` to close the PID-recycling window the function's comment already admits to. Both pre-date your change.

Thanks for the testing log, too — the shard run plus explicit "not run, and why" lines made this quick to review.


---

PR #4954, opened 2026-08-04
docs: explain managed wake behavior for mail

### LINKED ISSUE #4941: docs: explain that mail --notify can request a managed wake
## Page

`docs/tutorials/04-communication.md` (`/tutorials/04-communication`)

## What the docs say

The mail-versus-nudge table says, without qualification:

| | Mail | Nudge |
| --- | --- | --- |
| Wakes the recipient | No | Yes |

That is true for ordinary mail, but omits the supported `--notify` behavior.

## What the code does

`gc mail send --notify` and `gc mail reply --notify` request a managed wake for
a non-running recipient when the city has a managed reconciler:

- `cmd/gc/cmd_mail.go` wires `--notify` to `sendMailNotifyWithWorker`.
- `cmd/gc/cmd_nudge.go` enqueues a managed nudge, stamps wake demand through
  `enqueueManagedNudgeThenWake`, and pokes the controller when the recipient is
  not running.
- `TestSendMailNotifyWithWorkerManagedNonRunningQueuesWakeForController` in
  `cmd/gc/cmd_nudge_test.go` suspends a session and asserts the queued nudge,
  `wake_request`, timestamp, and controller poke.

The behavior is conditional: unmanaged/storeless paths can queue a nudge
without requesting a managed wake, and unread mail by itself is never wake
demand.

An isolated live send on edge commit `97503b2` also returned a dispatch receipt:

```console
$ gc mail send gastown.mayor \
    -s "BENCH notify probe" \
    -m "Wake probe; no action needed." \
    --notify --json
{"schema_version":"1","ok":true,"command":"mail.send","action":"send","id":"be-wisp-iwo","message":{"id":"be-wisp-iwo","from":"human","to":"gastown.mayor","subject":"BENCH notify probe","thread_id":"thread-23254a7b5a8a"},"count":1,"notified":true}
```

That receipt proves the notify path was dispatched, not that this particular
pane visibly transitioned idle→awake. The managed-wake behavior itself is the
code path and pinned test above.

## Suggested change

1. Qualify the tutorial row: mail does not wake by default; `--notify` can
   request a wake for a non-running recipient in a managed city.
2. Reword `--notify` help/reference text to mention managed wake behavior, not
   only unread-message de-duplication.
3. State explicitly that unread mail alone is not wake demand.

## Duplicate search

I searched open and closed issues for mail notify/wake and the communication
tutorial. #619, #3975, and #4534 are related notify/no-op reports, but none
reports this tutorial contradiction.

boshu2:
## Summary

- qualify the communication tutorial so ordinary mail is not described as categorically unable to wake a recipient
- document that `mail send --notify` and `mail reply --notify` can request a managed wake for a non-running recipient, while unread mail alone is not wake demand
- regenerate the CLI reference and add a regression for both commands' managed-wake help

Fixes #4941.

## Testing

- [x] `go test ./cmd/gc -run '^(TestMailNotifyHelpDocumentsManagedWake|TestCLIDocsFreshness)$' -count=1`
- [x] `make check-docs`
- [x] `go vet ./...`
- [x] `make lint-changed LINT_CHANGED_SCOPE=tracked LINT_CHANGED_REF=origin/main`
- [ ] `make check` — formatting, lint, vet, and routed-test checks passed; the unit sweep then hit unrelated macOS path-canonicalization failures reproduced on exact `origin/main`, plus ambient tmux/Dolt cleanup failures
- [ ] `make test-integration` — not applicable; runtime behavior is unchanged

## Checklist

- [x] Linked an issue
- [x] Added focused help-text regression coverage
- [x] Updated the tutorial and generated CLI reference


boshu2:
All checks are green, the exact candidate has an independent review PASS, and this is ready for maintainer review when convenient.

quad341 (approved):
@boshu2 — thank you for this. The old tutorial row said "Wakes the recipient: No," full stop, and in a managed city that was flatly wrong. That's the kind of wrong that quietly teaches people the wrong mental model for months, and it's easy to walk past. Good catch.

What I appreciate most is that you hedged instead of overcorrecting. "In a managed city, it **can** request a wake" maps exactly onto `canRequestManagedNudgeWake`, and "unread mail alone does not request a wake" holds up — I traced it, and message beads fail the molecule-type gate *and* sit in `readyExcludeTypes`. Two independent paths, both agreeing with your text. Covering both `send` and `reply` in the help-text regression is the right guard for a docs fix.

One thing I'll take on my side rather than bounce back to you: the closing clause says "unmanaged or storeless paths may only queue the nudge for later delivery." Unmanaged does queue, but storeless actually *skips* — `cmd_mail.go` only builds the nudge func when the store is non-nil, and reply prints "nudge skipped." I'll split those two cases in one line, re-run `check-docs`, and merge with your commit intact.

Shipping it. Thanks for making the docs match the code.


---

PR #4956, opened 2026-08-04
fix: bound entire stop sequence by timeout

### LINKED ISSUE #4939: bug: gc stop --timeout does not bound the whole stop operation
## Before you continue

- [x] I searched open and closed issues. #2616 is related, but covers stop-wave
  serialization rather than whether this command's public timeout bounds the
  whole invocation.
- [x] I read the contributor guidance.

## Gas City version

Observed on v1.4.0 (`a7297c511`) and a main-built binary at `97503b2`.
Code references below are against `main` at `1111964`.

## Environment

macOS (darwin 25.5.0, arm64), zsh, tmux provider, socket-mode supervisor,
managed Dolt store.

The measurement used an isolated `GC_HOME`, its own registered scratch city,
and a supervisor on a separate port. No production city was touched.

## Reproduction

No artificial stall is required. A healthy supervisor and freshly registered
scratch city were enough:

```console
$ gc init --template gastown --default-provider claude /tmp/scratch/city
$ gc register /tmp/scratch/city
$ gc supervisor start
Supervisor started (PID 35743)

$ cd /tmp/scratch/city
$ time gc stop --timeout 5s
Unregistered city 'city' (/tmp/scratch/city)
Reconciliation triggered.
City stopped.
gc stop --timeout 5s  0.50s user 1.54s system 5% cpu 38.794 total
```

The command returned success after 38.79 seconds, about 7.8 times the requested
wall-clock cap.

## Expected behavior

`gc stop --help` describes `--timeout` as the "wall-clock cap for the stop
sequence." A 5-second value should bound the whole operation. If unregister is
intentionally a separate budget, the flag and timeout diagnostics should say so.

## Actual behavior and mechanism

`cmd/gc/cmd_stop.go` calls `unregisterCityFromSupervisorWithForce` before it
creates the `select` that races `cmdStopBody` against the requested timeout.
That unregister path performs two sequential waits using
`supervisorCityStopTimeout(cityPath)`. The separate budget has a 180-second
floor in `cmd/gc/cmd_supervisor_city.go`.

Config loading and `stopManagedRuntimeWithoutConfig` are also outside the
wall-clock race. Inside the race, timeout returns without canceling the body
goroutine because no context is plumbed into `cmdStopBody`.

So `--timeout 5s` is currently a cap on only a later sub-step, even though the
public contract calls it a cap on the stop sequence.

## Additional context

Related: #2616. That issue addresses one unresolved target serializing a stop
wave; it does not cover the pre-cap unregister waits or the public timeout
contract described here.

boshu2:
## Summary

- apply an explicit `gc stop --timeout` wall-clock cap to the entire stop sequence, including city resolution, supervisor unregister, invalid-config recovery, and loaded-city cleanup
- preserve the existing config-derived default budget when the caller does not pass an explicit timeout
- emit the final success record only after the bounded worker returns, preventing a timed-out worker from reporting late success
- add a regression around a blocked supervisor-managed invalid-config stop

Fixes #4939.

## Testing

- [x] `go test -count=1 ./cmd/gc -run 'TestCmdStop|TestDoStop'`
- [x] `go test -count=10 ./cmd/gc -run '^TestCmdStopWallClockTimeoutBoundsSupervisorManagedInvalidConfigStop$'`
- [x] `go test -count=1 ./cmd/gc -run '^TestCanonicalSessionProviderFactoryCallerCensus$'`
- [x] `make lint-changed LINT_CHANGED_SCOPE=staged`
- [x] `go vet ./...`
- [x] `make check-docs`
- [x] required generators (`genspec`, `genclient`, and `genschema`) produced no diff
- [ ] `make check` — not repeated after the focused gates; exact `origin/main` currently reproduces unrelated macOS path-canonicalization failures on this host
- [ ] `make test-integration` — not run locally; the supervisor wait is covered by the focused command regression and full PR CI

## Checklist

- [x] Linked an issue
- [x] Added a regression test for the behavior change
- [x] Updated the session reconciliation requirements ledger


boshu2:
CI is fully green on the exact validated head (60 passing checks, 23 expected skips, no failures). The issue regression, race-focused checks, vet, docs, lint, and provider-factory census all passed locally as recorded above. Ready for maintainer review.

quad341 (approved):
@boshu2 — the diagnosis here is the good part. Spotting that `--timeout` only ever wrapped `cmdStopBody`, while `unregisterCityFromSupervisorWithForce` sat in front of it with waits flooring at 180s, is exactly why `gc stop --timeout=30s` behaved like a suggestion rather than a cap.

Two things I especially liked. The `wallClockCapApplied` flag is the right seam — keeping the config-derived default on the loaded-city body when no explicit timeout is passed, rather than inventing a cap before the config is even loaded, avoids a nested double cap cleanly. And hoisting the success record out of `doStop`/`cmdStopBody` so a timed-out worker can't print "City stopped." after the command already exited 1 — then actually asserting `stdout == ""` after draining the worker — is what makes the new regression test worth having rather than just a timing check.

Merging as-is. One follow-up I'll file on our side, not on you: `reg.Unregister` commits before the supervisor waits, and the error branches re-register to undo it, so a cap firing inside that window exits before the restore runs. Your change makes that window reachable sooner, but the compensating-transaction shape predates it and the fix belongs in the unregister path, not here.


---

PR #3845, opened 2026-06-30
orders: bulk-delete confirm gate + reaper backup-age guards

bourgois:
Focused replacement for #3733 (which had ballooned to a +30k accumulation branch carrying the union of ~8 other open PRs plus fork-only artifacts — unreviewable). This is the actual feature it was meant to deliver, extracted onto a clean base off `main`: **12 files, +581/-21**.

## What this does
A backup-age safety gate for destructive maintenance, in three parts:

1. **`BulkDeleteSafe` helper** (`internal/doctor/checks_bd_backup_freshness.go`) — reuses the existing bd-backup-freshness machinery (#3739) to answer "is the backup recent enough to permit a bulk delete?"
2. **Bulk-delete confirm gate** (`cmd/gc/cmd_order.go`) — the order sweep-tracking command requires `--confirm` (or `GC_BULK_DELETE_CONFIRM_THRESHOLD`) when the retention-eligible closed-bead count exceeds the threshold; **fail-closed** on a store count error rather than proceeding unguarded.
3. **Watchdog/reaper backup-age guard** (`order_dispatch.go`, `internal/bootstrap/packs/core/assets/scripts/reaper.sh`) — the reaper's `bd prune` step refuses to run when the backup is stale, so an unhealthy-backup city can't have history pruned out from under it.

## Tests
`TestOrderSweepTrackingRequiresConfirm`, `TestOrderSweepTrackingConfirmGateFailsClosedOnCountError`, the `BulkDeleteSafe` doctor tests, and shell coverage (`test/reaper_prune_backup_guard_test.sh`, `test/reaper_session_pattern_test.sh`) + a maintenance-scripts fixture.

`go build ./...`, `go vet`, gofmt, and the targeted suites pass; CLI docs regenerated.

PR #4957, opened 2026-08-04
feat(doctor): gate bulk deletes on fresh backups

sjarmak:
## Summary

This is the `doctor`-scoped replacement for #3845. It preserves Karel Bourgois's original commit and authorship while separating the backup-freshness API from the mixed-scope branch.

- adds `BulkDeleteSafe` over the existing backup-freshness machinery
- covers fresh, stale, and unconfigured managed scopes
- reuses the check's canonical managed-scope target selection

Original commit assigned here: `0398fca9136120d9e5d4613a42ecc90720968338`.

## Test plan

- `go test ./internal/doctor`
- pre-commit `lint-changed`, generated-doc freshness, and `go vet ./...`

Split from and supersedes the `doctor` portion of #3845. Please credit @bourgois for the original implementation.

### DIFF
--- a/internal/doctor/checks_bd_backup_freshness.go
+++ b/internal/doctor/checks_bd_backup_freshness.go
@@ -163,12 +163,9 @@ func (c *BdBackupFreshnessCheck) freshnessScanTargets() []bdBackupFreshnessTarge
 // configured" is DoltBackupCheck's concern, and failing closed there would
 // block bulk deletion on every unbacked city.
 func BulkDeleteSafe(cityPath string, cfg *config.City, maxAge time.Duration, now time.Time) (bool, string) {
-	scopeRoots := managedDoltScopeRootsForConfig(cityPath, cfg, nil)
-	for _, scopeRoot := range scopeRoots {
-		scopeRoot = filepath.Clean(scopeRoot)
-		label := bdBackupScopeLabel(cityPath, scopeRoot)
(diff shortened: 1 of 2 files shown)


quad341 (approved):
@sjarmak — thank you for doing the unglamorous part properly. Splitting the `doctor` half out of #3845 while keeping Karel's original commit `0398fca` and authorship intact is exactly how you unstick a mixed-scope branch, and the payoff is real on this side: 87 lines I could verify end to end instead of a branch nobody wanted to touch.

Special credit for the doc comment on `BulkDeleteSafe`. "Don't re-derive which pipeline is active, so the gate and the check can never disagree" is the right instinct, and I traced every claim in it against `scanBackupFreshness` — fail-closed on unreadable, unparseable, and missing-timestamp state; safe on no-state-at-all; registered-but-never-synced treated as unsafe. It all holds.

One spot where the claim doesn't quite hold yet, and I'll take it on my side: with `cfg == nil` the gate narrows to the city root and every rig goes unscanned, which fails *open* — the one direction a delete gate shouldn't. I'll fall back to disk discovery there, tighten the two test assertions (that stale-scope comment, and the swallowed `reason`), and add a Dolt-migrated case for the gate. Then `go test ./internal/doctor` and merge.

Please pass the thanks along to @bourgois for the original implementation.


---

PR #4958, opened 2026-08-04
feat(orders): guard bulk order-tracking deletion

sjarmak:
## Summary

This is the `orders`-scoped replacement for #3845. It preserves Karel Bourgois's two original orders commits and authorship while separating the CLI and watchdog behavior from the mixed-scope branch.

- requires `--confirm` when eligible retention deletions exceed the configured threshold
- fails closed when the eligibility count cannot be read
- guards the controller retention watchdog with backup freshness
- regenerates the CLI reference from the Cobra source

Original commits assigned here: `c57d7dbb6fff56b0a1a253c106e17dbe951d5070` and `e3f0dd420ad67a0d6d2b723c799f300bfbd780f6`.

## Dependency

Depends on #4957 for `doctor.BulkDeleteSafe`. This branch intentionally does not duplicate that doctor-scoped commit; focused orders tests pass when #4957 is layered underneath it.

## Test plan

- with #4957 layered: `go test ./cmd/gc -run 'OrderSweepTracking|OrderTrackingRetentionWatchdog'`
- with #4957 layered: `go run ./cmd/genschema` (clean worktree afterward)
- `make check-docs`

Split from and supersedes the `orders` portion of #3845. Please credit @bourgois for the original implementation.


quad341:
@sjarmak — thank you for doing the unglamorous part of this. Carving the `orders` scope out of #3845 while keeping @bourgois's two commits and authorship intact is exactly the right way to unstick a mixed-scope branch, and it made this reviewable in a way the original wasn't.

The preview-count design is the best thing here: routing through the same `ClosedRunsForRetention` read and `bucketClosedRetentionRuns` as the real sweep means the count genuinely can't drift from what gets deleted. I traced both paths — the doc comment holds up.

Two things on my side, neither yours. We'll land #4957 first so `doctor.BulkDeleteSafe` exists, then rebase — that's all the red CI is.

And I found one real thing: the core pack ships `order-tracking-sweep` on a 1m cooldown without `--confirm`, so once a city crosses 20 eligible beads that order fails every minute — and since the gate returns before the stale-close, it stops closing stale open tracking beads too, which can block re-dispatch. I'll take that: `--confirm` on the packaged order, plus the watchdog wiring tests and a regression test. Your guard stays exactly where you intended it, on the interactive path.

Credit stays with you and @bourgois. Nice work.


---

bourgois:
@sjarmak — the red CI here looks like a merge-ordering artifact, not a defect. A rebase should clear it with no code change.

The shard build error is:

```
cmd/gc/city_runtime.go:1514:29: undefined: doctor.BulkDeleteSafe
```

That CI run predates #4957, which is what defines the symbol. `BulkDeleteSafe` is now on `origin/main` at `internal/doctor/checks_bd_backup_freshness.go`, so the branch just needs to be rebased onto it.

Branch is `MERGEABLE` and `git merge-tree` against current `origin/main` is clean.

I also verified the split accounting you described on #3845, rather than taking it on trust — **all six commits are present exactly once across the four replacements, authorship preserved**:

| #3845 | landed in |
|---|---|
| `0398fca91` doctor: BulkDeleteSafe helper | #4957 |
| `c57d7dbb6` orders: confirm gate + watchdog guard | #4958 |
| `8cb2ec6db` reaper: backup-age guard | #4959 |
| `e3f0dd420` orders: confirm-gate fail-closed | #4958 |
| `ee9bcdbe8` ci: backup-state fixture + date stub | #4960 |
| `466ce4ce2` chore: drop fork-internal plan doc | #4959 |

File-level maps 1:1 too, and the split parts are *larger* than the originals — you added fixups on top (nil-cfg fail-closed, active-backup-pipeline read, prune-route stub). Nothing was dropped.

`GC_BULK_DELETE_CONFIRM_THRESHOLD` is the only piece of the original still unlanded, and it's in this PR. Once it merges I'll close #3845.

Thanks for the split — it worked far better than asking me to rebase a 12-file branch that had drifted hundreds of commits.

quad341 (approved):
@sjarmak — thank you for carving the `orders` scope out of #3845 with @bourgois's two commits and authorship intact. That's the unglamorous half of unsticking a mixed-scope branch, and it's why this one is reviewable at all.

The preview-count design is the best thing here. I traced `countClosedOrderTrackingRetentionEligible` against the real sweep line by line — same read, same bucketing, same retain-floor clamp, same sort and tie-break, same cutoff. The count genuinely can't drift from what gets deleted, exactly as your comment claims. Sequencing stale-close ahead of the gate and then asserting it in two separate tests is what makes this safe to run unattended.

@bourgois — your merge-ordering diagnosis was right for the run you saw, and #4957 has since landed. The remaining red is one thing and it's ours: `TestGCEnvReadBaseline` freezes the `GC_*` vocabulary, so the two new knobs need adding to the golden. I reproduced it locally — two lines, no Go changes. I'll take that and re-run the shard.

One note for later, not for this PR: the packaged order passes `--confirm` unconditionally, so the guard is interactive-only in practice. Right call now; just want it written down.

Credit stays with both of you.


---

PR #4959, opened 2026-08-04
feat(reaper): guard session pruning with backup age

sjarmak:
## Summary

This is the `reaper`-scoped replacement for #3845. It preserves Karel Bourgois's original reaper work and authorship while keeping the temporary fork plan out of the final diff.

- skips bulk session-bead pruning when backup state is absent, invalid, or stale
- records an anomaly explaining the rejected prune
- adds focused shell coverage for absent, fresh, and stale backup state
- keeps the existing configurable session-pattern coverage aligned with the current `gc bd` route

Original commits assigned here: `8cb2ec6db1da6c8baa9197029776e258ff07db84` and `466ce4ce276593d9745c8d019140220bf961b4b9`.

## Test plan

- `bash test/reaper_prune_backup_guard_test.sh`
- `bash test/reaper_session_pattern_test.sh`

Split from and supersedes the `reaper` portion of #3845. Please credit @bourgois for the original implementation.


quad341 (approved):
@sjarmak — thank you for doing this split properly. Carving the reaper half out of #3845 with @bourgois's two commits preserved (`8cb2ec6`, `466ce4c`) instead of squashing the history is real work, and it's the reason this one is reviewable at all. Credit for the guard itself stays with @bourgois.

Separate shout-out for the `bd()` → `gc()` stub fix in `reaper_session_pattern_test.sh` — the script moved to `gc bd …` a while back, so that harness was intercepting nothing and those tests were passing for the wrong reason. Nice catch, and entirely independent of the guard.

I'm taking this on our side rather than sending it back, because the one real issue is fleet knowledge you had no way to see from outside. Gas City has two backup pipelines: once a scope registers a Dolt destination, `bd backup sync` writes `.beads/dolt-backup-state.json` and never touches the legacy `backup/backup_state.json` this guard reads — so on migrated cities the guard would latch closed and stop session pruning permanently. Our own `checks_bd_backup_freshness.go` was fixed for exactly this; I'll point the guard at whichever pipeline is live, normalize the nanosecond timestamps the real files carry, harden the `python3` fallback, and add your two missing scenarios to the test file — your structure, just more cases.

The guard is the right idea, and fail-closed is the right instinct. Thanks to you both.


---

PR #4960, opened 2026-08-04
fix(ci): add fresh backup state to reaper fixtures

sjarmak:
## Summary

This is the `ci`-scoped replacement for #3845. It preserves Karel Bourgois's original fixture commit and authorship while separating test-harness compatibility from the production changes.

- gives reaper integration fixtures a fresh backup state where pruning is expected
- exposes `date` in the restricted-PATH fixture used by the backup-age guard

Original commit assigned here: `ee9bcdbe877821d3f2346299e3af783e2cc34b56`.

## Test plan

- `go test ./examples/gastown`

Split from and supersedes the `ci` portion of #3845. Please credit @bourgois for the original implementation.

### DIFF
--- a/examples/gastown/maintenance_scripts_test.go
+++ b/examples/gastown/maintenance_scripts_test.go
@@ -3620,6 +3620,7 @@ func TestReaperPrunesClosedSessionBeadsWithBdPrune(t *testing.T) {
 		cityDir = resolved
 	}
 	writeCityBeadsMetadata(t, cityDir, "beads")
+	writeFreshBackupState(t, cityDir)
 	canonicalCityDir, err := filepath.EvalSymlinks(cityDir)
 	if err != nil {
 		t.Fatalf("EvalSymlinks(city dir): %v", err)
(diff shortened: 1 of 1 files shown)


quad341 (approved):
@sjarmak — thank you for doing this split the hard, correct way. Carving the `ci` half out of #3845 while keeping @bourgois's commit intact at the *git* layer (author field is `bourgois <karel@voxist.com>`, not just a mention in the body) is meaningfully more work than squashing it into a fresh commit, and it's the version that's actually mergeable. Credit lands where it belongs.

Merging as-is: `go vet ./examples/gastown` is clean, the nine affected reaper tests pass locally, and full CI is green — all 12 `cmd/gc` shards, integration, and CodeQL with zero open alerts.

One note for whoever picks up the production half, so nobody later reads "fix(ci)" and goes hunting for the bug it fixed: the backup-age guard isn't on `main` yet, so these fixtures are inert until it lands — which is the entire point of the split. The helper writes `{"timestamp": <RFC3339>}` to `.beads/backup/backup_state.json`, matching what `internal/doctor/checks_bd_backup_freshness.go` already reads. If the guard reads `last_sync` or `last_dolt_commit` instead, it's a one-line fixture change and CI will say so immediately.

And thanks to @bourgois for the original work — adding `date` to the restricted-PATH list is the kind of detail that's only obvious after you've watched a `set -euo pipefail` script exit 127.


---

PR #3835, opened 2026-06-30
fix(reconciler): graph-only readiness + idle-respawn safety-net

quad341:
## Summary

- **Step 1** — Port `GraphOnlyReadyFor` capability probe + `liveReadyForControllerDemandQuery` awake-gate change (from stranded `71b7040fa`): under `graph_store=sqlite`, the controller-demand awake gate now prefers the graph-class ready slice, preventing genuinely-ready assigned graph wisps from being evicted by the per-tick limit. Dolt-only cities fall back byte-identically.
- **Step 2** — Derive idle-respawn safety-net on current main (from stranded `0062c4124` as spec): a pool session awake *only* for ready assigned work now completes a non-cancelable `"idle-respawn"` drain to ASLEEP, so the existing resume-on-ready path re-spawns it fresh; it re-runs `gc hook --claim` and the molecule advances. No nudges required.
- **Also** — cherry-pick the docsync ga- agent-work-directory skip (`b3e98d2cb`) so the pre-push test suite passes in worktrees that have untracked ga-* transcript dirs.

This is the pre-requisite for PR #3833 (`fix(hook): enqueue continuation nudges for pool graph.v2 roots`); merge this first, then ungate #3833.

## Why it was broken

Without this fix, a pool session awake only for a ready assigned wisp (e.g. an adopt-pr molecule's cleanup/finalize step) was pinned alive at the `shouldWake && alive` branch, which only cancelled drains — never re-engaging the idle agent. Molecules hung at 61/71 for ~26h.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `TestIdleAssignedWorkOnly` — PASS
- [x] `TestDrainReasonCancelable_IdleRespawnNotCancelable` — PASS
- [x] `TestSelectIdleProbeTargets_IncludesAssignedWorkOnly` — PASS
- [x] `TestSelectIdleProbeTargets_SkipsOtherWakeReasons` — PASS
- [x] `TestBeginIdleRespawnDrainIfIdle` — PASS
- [x] `TestBeginIdleRespawnDrainIfIdle_SkipsNonInteractive` — PASS
- [x] `TestLiveReadyForControllerDemandPrefersGraphOnly` — PASS
- [x] `TestControllerDemandReadyFallsBackWithoutGraphCapability` — PASS
- [x] Pre-push fast test suite — all 8 jobs PASS

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4962, opened 2026-08-04
test(docsync): skip ga- agent work directories

sjarmak:
## Summary

- skip ephemeral ga- agent work directories in TestDocDirCoverage

This is the docsync-scoped replacement for #3835. All credit belongs to @quad341; the cherry-picked commit retains original authorship.

## Validation

- go test ./test/docsync
- go vet ./...
- git diff --check origin/main...HEAD

No merge without a dedicated review record.

### DIFF
--- a/test/docsync/docsync_test.go
+++ b/test/docsync/docsync_test.go
@@ -819,7 +819,7 @@ func TestDocDirCoverage(t *testing.T) {
 			continue
 		}
 		name := e.Name()
-		if strings.HasPrefix(name, ".") || name == "vendor" || name == "node_modules" {
+		if strings.HasPrefix(name, ".") || strings.HasPrefix(name, "ga-") || name == "vendor" || name == "node_modules" {
 			continue
 		}
 		if known[name] {


quad341 (approved):
@sjarmak — thanks for splitting this cleanly out of #3835 and keeping @quad341's authorship on the cherry-pick. That's the detail that's easiest to lose when re-scoping a PR, and you got it right: the commit still shows quad341 as author.

Narrowing to just the docsync filter made this trivially reviewable. We pulled the branch and ran `go test ./test/docsync/` and `go vet ./test/docsync/` — both clean, `TestDocDirCoverage` green. It also lines up with `TestScaffoldNoiseIsSkipped` in `internal/testenv/lint_test.go`, which already documents stray `ga-*` bead worktrees tripping this exact repo-root walker, so you're closing a gap we'd already noticed rather than guarding against a hypothetical.

Merging as-is.

One note for whoever next touches this file, no action needed from you: the skip logic is inline in the test body, which is why it has no direct assertion the way `isNestedWorktreeRoot` does. Pulling it into a named helper would let it get locked in alongside the other scaffold-noise guards — out of scope for a one-liner.

Thanks for carrying quad341's fix over the line.


---

PR #4964, opened 2026-08-04
fix(reconciler): graph-ready wake gate and idle respawn

sjarmak:
## Summary

Replacement reconciler part of #3835. It preserves the five source commits whose changes depend on the reconciler feature:

- 7c4201df: graph-only readiness and idle-respawn behavior
- 8c144d86: intermediate reconciler release-gate record
- 294c9c94: staticcheck correction
- 7bec4825: correction in the new idle-respawn test
- bb281212: reviewed reconciler fixes and removal of the superseded FAIL gate

The intermediate release-gate add/removal stays in this part so the final tree does not ship a stale FAIL record. The dependent test-comment correction stays with the test file introduced here, avoiding duplicate feature commits across replacement PRs.

An erosion pass removed 52 lines of ported review-history narration while preserving the implementation and authorship commits.

## Verification

- focused idle-respawn and graph-readiness regression suite — pass
- go test -race ./cmd/gc -run ManagedDolt\|DoltRecovery\|DoltState — pass
- go test ./internal/beads — pass
- pre-commit golangci-lint — pass
- go vet ./... — pass

The full fast pre-push suite passed every test shard, then failed its final cmd/gc Dolt leak guard because TestInitFromWithoutHostedPreservesTemplate left a test-owned dolt sql-server alive at sweep time. The same unchanged-baseline leak reproduced twice; the process exited after each run. No changed-path test failed.

## Review

This split part requires its own normal adopt/review record before merge. Do not merge solely on the review record attached to #3835.


sjarmak:
Closing as a duplicate of the already-open scoped replacement #4961. The concurrent worker completed the canonical split first; no review should be spent on this duplicate.

PR #4217, opened 2026-07-13
Feat/herdr first class

McKean:
## Summary

- Preserve the warm-bind claim nudge as the fast path while keeping upstream’s rebuilt stalled-claim poller as the unconditional backstop.
- Port the herdr, runtime, supervisor, session, bootstrap, mail, and event-pump changes onto upstream’s newer typed session/domain APIs and store-class routing.
- Resolve integration conflicts that only appeared when the branches were built together, including duplicate test fakes, `herdr.New` signature drift, and the claim-nudge/backstop overlap.
- Note for reviewers: `NeedsClaimBackstop` is currently declared but unconsumed because upstream’s rebuilt backstop now runs unconditionally. It can be removed if preferred.

## Testing

- [x] `make check`
- [ ] `make check-docs` if docs, navigation, or links changed
  > Not applicable unless docs changed in this PR.
- [x] `make test-integration` if runtime, controller, or workflow behavior changed

Additional verification:
- Build and vet are clean over the whole tree.
- Runtime/herdr/auto/hybrid, mail, supervisor, bootstrap, session, `cmd/gc` nudge/wait/prime/mail/event-pump subsets, and dolt quarantine tests are green.
- Known pre-existing failure: `TestDoPrimeWithHookFormat_FormatsDefaultFallback` fails identically at the untouched base commit in town-nested worktrees.

## Checklist

- [x] Linked an issue, or explained why one is not needed
- [x] Added or updated tests for behavior changes
- [ ] Updated docs for user-facing changes
- [x] Called out breaking changes or migration notes

No single issue covers this integration PR; it consolidates the existing herdr/town fix branches and PR-A through PR-F onto current upstream `main`.

No breaking changes are expected. Deployment note: the town binary is still running `fix/town-consolidated`; this branch should be the next `make install` candidate, but it needs a restart window because the event-pump changes live supervisor-side.

PR #4965, opened 2026-08-04
refactor(runtime): share setup-command runner (#4217 split 1/10)

sjarmak:
## Summary

Extract the host-side setup-command runner from the tmux adapter into the shared runtime layer.

Split from #4217 and adapted onto current main. Thank you @McKean for the original implementation.

## Stack

This is the stack base. This PR intentionally targets main; merge in numeric order so the displayed diff collapses to this scope. Do not merge without the normal maintainer review record.

## Source commit ledger

This part owns exactly these original #4217 commits:

- `04416e468`

## Verification

- Compile-safe cumulative stack: `make test`
- `go vet ./...`
- Repository pre-push fast suite completed once; a repeat exposed the known test-owned Dolt cleanup race after all assertions passed


quad341 (approved):
@sjarmak — thank you for taking on the #4217 split, and especially for the source-commit ledger. Being able to see that this part owns exactly `04416e468` turned verifying the scope into a two-minute job instead of a two-hour one. Carrying @McKean's attribution forward is appreciated too.

The extraction is clean and the `commandOutputTail` table test is genuinely thorough — exact-limit-no-marker and zero-limit are the cases people usually forget. CI is green and nothing calls the new function yet, so there's no behavioral risk in landing it.

Two things I took on my side, no action needed from you. The doc comment said the tmux adapter "delegates here" with "identical semantics" — on current main that isn't true yet: tmux and herdr have both since grown the `execgrace` layer (cooperative process-group interrupt so shell rollback traps run before SIGKILL, plus the activity-aware `setup_max_timeout` budgets), which this snapshot predates. Entirely understandable coming off an older branch. I corrected the comment and left an explicit `PARITY REQUIRED BEFORE WIRING` note so the wiring part of the stack can't quietly drop that behavior, and filed `gm-7b0wd` to track it alongside the two provider-specific bits that have no home in the shared signature yet (tmux's `GC_TMUX_SOCKET` injection, herdr's `GC_DIR`-else-`cityRoot` fallback). I also ported tmux's behavioral tests for the runner across — cwd, env overlay, failure detail, timeout, and the daemonizing-child bounded-success regression.

Merging with your credit intact. Looking forward to the rest of the series.

PR #4395, opened 2026-07-17
fix: stop blocked work from re-entering dispatch

sjarmak:
## Summary

- make carried-route recovery query live raw Bead status before restoring `gc.routed_to`
- build the workflow projection from live `open` and `in_progress` reads, deduplicated by Bead ID
- add regression coverage proving blocked work is neither rerouted nor admitted to the active workflow projection

Gas City's cached Bead model collapses `blocked` and `deferred` into its internal `open` state. Non-live status filters therefore allowed blocked workflow roots through: the reaper cleared `gc.routed_to`, then route recovery restored it on a later patrol and the projection could consume it again. These reads now reach the backing store's raw-status filter.

## Test plan

- `go test -count=1 ./cmd/gc -run TestRestoreCarriedWorkRoutes`
- `go test -count=1 ./internal/api -run TestListActiveWorkflowProjectionBeadsExcludesBlocked`

PR #4979, opened 2026-08-04
fix(controller): keep partial dispatcher demand reads retention-only

sjarmak:
## Problem

When a demand read over multiple stores completes only partially, the scale-check pass marks every control-dispatcher template as partial. That marking does two jobs at once: it retains existing dispatcher sessions (correct, fail-safe) and it suppresses new dispatcher creates (incorrect when a healthy store is exposing real control demand). The result is a cold system that cannot spawn its first control dispatcher while any store read is degraded, even though another store has live demand. Follow-up to the demand-read work in #4395.

## Fix

Split the two behaviors. Partial-read markings are now retention-only: they keep existing dispatchers alive but never veto a create. Create suppression applies only from an ordinary scale-check verdict computed over the stores that did read successfully.

## Tests

New two-store regression: one store exposes control demand, the other simulates an outage mid-read. The test fails before the production change (cold dispatcher create vetoed) and passes after (create planned, existing dispatchers still retained). Focused suite run with the race detector, plus vet and build, all green at this head.

### DIFF
--- a/cmd/gc/build_desired_state.go
+++ b/cmd/gc/build_desired_state.go
@@ -63,10 +63,13 @@ type DesiredStateResult struct {
 	BaseState        map[string]TemplateParams
 	ScaleCheckCounts map[string]int // nil when store is nil or scale_check not run
 	// ScaleCheckPartialTemplates records all templates whose bead-backed demand
-	// probe failed. PoolScaleCheckPartialTemplates drives generic pool retention;
+	// probe failed. PoolScaleCheckPartialTemplates blocks fresh pool creates;
+	// PoolPartialRetentionTemplates preserves existing pool capacity and may also
+	// contain retention-only failures where another store proved positive demand.
 	// NamedScaleCheckPartialTemplates only protects configured named sessions.
 	ScaleCheckPartialTemplates      map[string]bool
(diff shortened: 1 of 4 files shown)


quad341 (approved):
@sjarmak — thank you for this one. Splitting `PoolScaleCheckPartialTemplates` into create-suppression versus retention-only is the right cut: those two behaviors were riding on a single flag and only one of them was ever meant to be fail-safe, which is exactly how a cold system ends up unable to start its first dispatcher.

The detail I appreciated most is the ordering — seeding the retention set from the create-suppression set *before* the routed-partial marking, so retention stays a strict superset and genuine probe failures lose nothing. That's easy to get backwards, and getting it backwards would have been invisible in the tests. I checked the merge helper for the same reason: it allocates rather than returning `src`, so the marking can't leak back into the create gate.

The two-store regression earns its keep — asserting both directions in one test is what makes this reviewable without re-deriving the whole demand path.

Merging as-is. One optional follow-up if you're inclined: a tick where a real `scale_check` probe failure and a routed-read partial land together. That's where the union does the work, and it'd pin the behavior against a future refactor of the merge helper. Nothing blocking.


---

PR #4981, opened 2026-08-04
fix(session): align bead actor with canonical alias

A3Ackerman:
## Problem

Gas City currently has two distinct session identifiers:

- the **public ownership identity** used by work discovery and claims (`alias`, with compatibility fallbacks)
- the provider **runtime handle** (`session_name`)

GC-owned paths selected between them independently. Hook claims, API assignment normalization, scripted claims, and `BEADS_ACTOR` could therefore write/present different exact strings for the same session. `bd` correctly rejects a close when the actor does not exactly match the assignee.

Changing only `BEADS_ACTOR` would invert the mismatch for upstream hook/API paths. The fix must move every GC-owned ownership writer and runtime actor together.

## Decision

The session object model now owns one compatibility selector in `internal/session.AssigneeIdentifier`:

1. current `alias`
2. `configured_named_identity` when recovering a named session whose alias is temporarily absent
3. raw persisted `session_name`
4. session bead ID when name metadata is absent

`RuntimeEnvWithSessionContext` and `SyncRuntimeAlias` accept typed `session.Info` and use that same selector. This preserves the distinction between `Info.SessionName` (the actual provider handle) and `Info.SessionNameMetadata` (the durable ownership fallback).

The CLI and API remain projections over that rule:

- API assign/update/create normalization calls `session.AssigneeIdentifier`
- hook and agent-script claims prefer `GC_ALIAS`, then projected `BEADS_ACTOR`, then compatibility `GC_AGENT`, then the runtime handle
- `template_resolve` no longer independently authors `BEADS_ACTOR`
- alias set/clear/rollback and reconciler alias sync pass post-update typed `session.Info`
- runtime metadata synchronization snapshots and rolls back partial provider failures

## Project-principle alignment

- **Object model at the center:** the selector and runtime projection live in `internal/session`; `cmd/gc` and `internal/api` consume them (`engdocs/architecture/invariants.md`).
- **Claim identity convention:** ownership reads and writes use the current concrete session identity, not the shared template queue key (`engdocs/architecture/prompt-templates.md`, “Claim Identity Convention”).
- **Public identity vs runtime handle:** aliases remain public identities; `session_name` remains the provider handle (`engdocs/design/named-configured-sessions.md`).
- **No new primitive or cognition:** this is deterministic transport normalization. `bd` still owns the atomic claim CAS (`engdocs/contributors/primitive-test.md`).
- **Migration-aware:** the final session-model target remains `assignee=<session-bead-id>`. `engdocs/design/session-model-unification.md` now documents this alias-first compatibility projection until that migration reaches every writer and prompt.
- **No role policy:** no configured role name or judgment entered Go code.

## Compatibility and rollout

- Aliasless sessions retain raw `session_name` ownership; repairable nameless sessions use bead ID while retaining their derived provider handle in `GC_SESSION_NAME`.
- Historical stranded rows are not rewritten here; compatibility readers still accept all supported session identity forms.
- Existing agent processes retain their inherited environment. They must restart before direct `bd` commands inherit the new actor. `Provider.SetMeta` keeps provider metadata coherent for subsequent launches but is not a live-process environment migration.
- No HTTP/SSE schema or dashboard contract changes.

## Verification

- Focused RED/GREEN suite covers alias, configured-name recovery, aliasless session-name fallback, nameless bead-ID fallback, API assign/update/create, hook/script claims, alias set/clear/rollback, reconciler sync, and provider metadata rollback.
- `go test ./internal/session ./internal/api -count=1`: PASS
- focused `go test ./internal/session ./internal/api ./cmd/gc ... -count=1`: PASS
- `make lint`: PASS (0 issues)
- `make vet`: PASS
- `make check-docs`: PASS
- `.githooks/pre-commit`: PASS

The initial CI failure (`TestBuildDesiredState_DependencyFloorIgnoresConfigBlindLegacySlotRecovery`) exposed an over-broad intermediate desired-state fix. That approach was removed; ownership now converges at the typed session/runtime boundary. The exact failing contract passes locally. A full macOS shard remains affected by the pre-existing `/var` versus `/private/var` canonicalization failures, while upstream Linux CI provides the authoritative full-shard result.

### DIFF
--- a/cmd/gc/bd_env_test.go
+++ b/cmd/gc/bd_env_test.go
@@ -4195,11 +4195,10 @@ func TestBdRuntimeEnvDoesNotDefaultBeadsActorWhenUnset(t *testing.T) {
 	}
 }
 
-// TestBdRuntimeEnvPreservesInheritedBeadsActor verifies that session
-// contexts (template_resolve.go sets BEADS_ACTOR=<sessname>) and exec
-// orders (orderExecEnv sets BEADS_ACTOR=order:<name>) are not clobbered by
-// the neutral bd runtime env. The key is omitted so the inherited value
-// passes through mergeEnv unchanged.
+// TestBdRuntimeEnvPreservesInheritedBeadsActor verifies that the authoritative
+// session runtime context and exec orders can set BEADS_ACTOR without the
+// neutral bd runtime env clobbering it. The neutral map omits the key so the
+// inherited value passes through mergeEnv unchanged.
 func TestBdRuntimeEnvPreservesInheritedBeadsActor(t *testing.T) {
 	t.Setenv("GC_BEADS", "bd")
 	t.Setenv("GC_DOLT", "skip")
--- a/cmd/gc/cmd_agent_script.go
+++ b/cmd/gc/cmd_agent_script.go
@@ -816,7 +816,7 @@ func agentScriptHookExitIsNoWork(output, stderr string) bool {
 }
 
 func agentScriptClaimActor() string {
-	for _, key := range []string{"GC_SESSION_NAME", "GC_AGENT", "GC_ALIAS", "BEADS_ACTOR"} {
+	for _, key := range []string{"GC_ALIAS", "BEADS_ACTOR", "GC_AGENT", "GC_SESSION_NAME"} {
 		if value := strings.TrimSpace(os.Getenv(key)); value != "" {
 			return value
 		}
@@ -838,7 +838,7 @@ func agentScriptRig() string {
 }
 
 func agentScriptAlias() string {
-	for _, key := range []string{"GC_ALIAS", "GC_SESSION_NAME", "GC_AGENT"} {
+	for _, key := range []string{"GC_ALIAS", "BEADS_ACTOR", "GC_AGENT", "GC_SESSION_NAME"} {
 		if value := strings.TrimSpace(os.Getenv(key)); value != "" {
 			return value
 		}
--- a/cmd/gc/cmd_agent_script_test.go
+++ b/cmd/gc/cmd_agent_script_test.go
@@ -14,8 +14,9 @@ import (
 	"time"
 )
 
-func TestAgentScriptBDClaimUsesSessionActor(t *testing.T) {
-	t.Setenv("GC_SESSION_NAME", "demo/worker-1")
+func TestAgentScriptBDClaimUsesCanonicalActor(t *testing.T) {
+	t.Setenv("GC_ALIAS", "demo/worker")
+	t.Setenv("GC_SESSION_NAME", "demo--worker")
 
 	var calls [][]string
 	exec := agentScriptExecutor{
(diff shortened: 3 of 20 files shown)


A3Ackerman:
Force-pushed `336212774` after investigating the failed cmd/gc shard and auditing the identity contract end to end.

The failed test showed that the first patch treated every aliasless dependency floor as deliberately deferred. I removed that intermediate desired-state approach rather than special-casing the test.

The revised patch puts the rule at the session boundary:

- one `session.AssigneeIdentifier` selector for runtime actor and API assignment
- typed `session.Info` passed through create, resume, reconciler, and alias-sync paths
- exact handling for alias, recovered configured identity, raw session name, and nameless bead-ID fallback
- transactional provider-metadata updates with rollback
- hook/script ownership reads prefer `BEADS_ACTOR` over compatibility/runtime fallbacks
- explicit engdocs note preserving the final bead-ID migration target and requiring restart of pre-patch agent processes

Local focused contracts, full `internal/session` + `internal/api`, lint, vet, docsync, and pre-commit pass. CI is rerunning on the new SHA.

quad341 (approved):
@A3Ackerman — thank you for this one, and especially for how you handled the CI failure. Backing out the intermediate desired-state fix rather than special-casing `TestBuildDesiredState_DependencyFloorIgnoresConfigBlindLegacySlotRecovery` was the right instinct, and the version that landed at the typed session boundary is a much better shape than a test-targeted patch would have produced.

Putting the whole rule in `session.AssigneeIdentifier` and making the CLI, API, hooks, scripted claims, and `BEADS_ACTOR` all projections over it is the fix. I traced the alias-first flip against the checkout rather than taking it on faith: `AssigneeIdentities` still enumerates every form, and the hook's identity-candidate matching uses that same set, so rows the old code stamped as `session_name` stay matchable and adoptable. The snapshot-and-rollback in `SyncRuntimeAlias`, with the injected-failure test, is a nice touch nobody asked for.

Merging as-is. Two things I'll pick up on our side, neither blocking you: beads claimed *before* this lands still carry `assignee=<session_name>`, so a restarted agent doing a direct `bd close` may hit the same mismatch inverted — hook adoption self-heals, but we may want a one-time sweep at rollout. And `TestCreateInjectsUnifiedSessionRuntimeEnv` now takes `Alias: ""`, so the explicitly-aliased named case lost its pin; I'll add it back as a second table case rather than bounce the PR for it.


---

quad341:
Thanks for this — aligning on a canonical actor is the right direction, and the precedence reorder in `firstNonEmptyHookValue` is clearly the correct end state. Flagging one question before it lands, because it touches something we've been chasing all day.

This changes **which spelling gets stamped** as the assignee (alias-first rather than session-name-first). We currently have at least four actor-identity spellings coexisting in the stores for the same actor — `role.role`, `role__role`, `rig/role`, `rig--role` — and `bd list --assignee` is an **exact** match. Measured today:

```
--assignee=pack-author.pack-author   --status=in_progress -> 0
--assignee=pack-author__pack-author  --status=in_progress -> 1
```

Both present in the same store, same actor. That already causes real failures: a role's own crash-recovery query returns zero against its own live claim, and lease state has been unreliable fleet-wide as an abandonment signal.

**The question:** what happens to beads already stamped with the old form? After this lands, new writes use the canonical alias while historical rows keep whatever was stamped before — so any exact-match query keyed on one form silently misses the other, and the failure looks like "no work" rather than an error.

If the change already handles that (a migration, or readers normalizing before comparison), great — please point at it and this is a non-issue. If not, the fix is still right; it just wants a backfill or a normalize-on-read alongside it.

Adding `needs-architectural-review` so that gets a look rather than blocking on my read of it. Not a request for changes to your code — the direction is one we want.

Issue #4840, opened 2026-07-29
bug: Dashboard Runs Dashboard & Detail Pages Inaccurate


JarrodJ83:
### Before you continue

- [x] I searched existing issues and did not find a duplicate.
- [x] I read the relevant docs and contributor guidance.

### Gas City version

1.4.0

### Environment

ubuntu

### Reproduction

1. Create a brand new city and rig based on guidance in getting started tutorial
2. sling some basic hello world work
3. Wait for all work to complete successfully

### Expected behavior

Run overview and detail views are accurate to current state of work

### Actual behavior

  - All runs show as "blocked" even though all beads are closed
  - The individual run view shows steps as "ready" even though they are closed

### Logs, screenshots, or traces
<img width="1829" height="1018" alt="Image" src="https://github.com/user-attachments/assets/ffb403c8-610b-4c70-986d-34ac5256758f" />

```
GET http://127.0.0.1:8372/api/city/my-city/runs/mp-bey/detail/stream [500]
```

```
POST http://127.0.0.1:8372/api/city/my-city/runs/mp-bey/diff?scope_kind=rig&scope_ref=my-project [403]
{"executionPath":{"kind":"known","path":"/home/azureuser/my-project"}}
```

The1nk:
### Root cause: it is `/api/city/<city>/runs/summary`, not the SPA

Reproduced on a different city (Gas City 1.4.0, Ubuntu) and localised, because the symptom as filed would reasonably send someone into the front-end where the bug is not.

**The summary endpoint reports completed runs as blocked.** Cross-checking every id in its `blockedLanes` array against `/v0/city/<city>/runs`:

```
blockedLanes claims 41 blocked. /v0 says of those same ids:
   37  completed
    4  waiting
```

None are blocked. And its own counts do not agree with each other:

```json
"runCounts": { "total": 23, "visible": 23, "blocked": 41, "other": 23 }
```

Forty-one blocked out of a total of twenty-three.

**The data is correct one layer down**, which is what makes this actionable. For a run the dashboard shows as "blocked, awaiting operator":

```console
$ curl -s "http://localhost:8372/v0/city/<city>/runs/<run-id>?scope_kind=rig&scope_ref=<rig>"
{"status":"completed", ...}
```

[...]

PR #4983, opened 2026-08-04
fix(runproj): resolve compiler formula metadata

masonjames:
## Summary

- Resolve run-summary formula identity from the root `gc.formula_name` metadata emitted by the graph.v2 compiler.
- Preserve the existing precedence of `pr_review.workflow_formula` and `gc.formula`.
- Add a regression test modeled on a rig-scoped custom formula run.

## Why

The graph.v2 compiler records a custom formula name in `gc.formula_name`, and run detail already reads that field. The summary lane only checked the older formula metadata keys, so a valid custom-formula run could be projected as formula unavailable and operator-needed even while its steps were active or complete.

This keeps summary and detail formula identity aligned without changing run execution, step state, or the historical title fallback.

Related to #4840.

## Validation

- `CGO_ENABLED=0 go test ./internal/runproj -count=1`
- `CGO_ENABLED=0 go vet ./internal/runproj`
- `git diff --check origin/main...HEAD`


masonjames:
## Live Masontopia validation

I also exercised this patch through the real dashboard summary endpoint using a local combined binary against two completed custom-formula runs.

`GET /api/city/masontopia/runs/summary` now reports:

- `totalActive: 0`, `totalHistorical: 2`, and `runCounts.blocked: 0`
- `socialtv-constellation-factory` with `formula.status: known`
- `socialtv-creative-council` with `formula.status: known`
- no operator requirement on either completed run

This is the user-facing path that had previously rendered the custom formula as unavailable and surfaced a false “Blocked, awaiting operator” state. The focused `internal/runproj` test and vet remain green.

masonjames:
## Live local-city validation

I also exercised this patch through the real dashboard summary endpoint using a local combined binary against two completed custom-formula runs.

`GET /api/city/{city}/runs/summary` now reports:

- `totalActive: 0`, `totalHistorical: 2`, and `runCounts.blocked: 0`
- both custom formula names with `formula.status: known`
- no operator requirement on either completed run

This is the user-facing path that had previously rendered the custom formula as unavailable and surfaced a false “Blocked, awaiting operator” state. The focused `internal/runproj` test and vet remain green.

PR #4984, opened 2026-08-04
fix(dispatch): preserve retry task descriptions

rbriski:
## Summary

Preserve Step.Description when the generic control dispatcher rebuilds a top-level retry or Ralph attempt from its frozen spec. Nested children already preserve this field; top-level attempts did not, which can launch a correctly routed worker bead with an empty task.

The regression serializes the frozen step, invokes the real spawnNextAttempt / molecule.Attach path, and verifies retry and Ralph attempts 2 and 3 retain both the task and formula-owned result metadata.

## Verification

- focused regression failed all four cases before the one-line fix
- focused regression passes 10/10 after the fix
- full internal/dispatch package passes
- changed-package lint and repository-wide go vet pass via pre-commit
- fast unit core plus five of six cmd/gc shards pass
- the remaining shard completes its tests and fails only a test-owned Dolt leak guard that reproduces identically on pristine base 49aec96e

Exact candidate: 3976ad670. Exact-head review: approve, zero findings.

Refs ga-ck6.

### DIFF
--- a/internal/dispatch/control.go
+++ b/internal/dispatch/control.go
@@ -762,13 +762,14 @@ func buildAttemptRecipe(step *formula.Step, control beads.Bead, attemptNum int) 
 		rootMeta[beadmeta.RalphStepIDMetadataKey] = stepID
 	}
 	rootStep := formula.RecipeStep{
-		ID:       attemptPrefix,
-		Title:    step.Title,
-		Type:     step.Type,
-		IsRoot:   true,
-		Labels:   append([]string{}, step.Labels...),
-		Assignee: step.Assignee,
-		Metadata: rootMeta,
+		ID:          attemptPrefix,
+		Title:       step.Title,
+		Description: step.Description,
+		Type:        step.Type,
+		IsRoot:      true,
+		Labels:      append([]string{}, step.Labels...),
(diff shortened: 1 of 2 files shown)


quad341 (approved):
@rbriski — good catch, and the kind that's easy to stare past. Nested children in `buildAttemptRecipe` copied `Description`; the top-level root didn't. So a retry attempt 2 would route perfectly to the right worker and then hand it an empty task — attempt 1 worked because it comes from the compile-time clone path, which made the asymmetry nearly invisible.

What sold it was the test going through the real `spawnNextAttempt` → `molecule.Attach` path with an actual serialized frozen spec, rather than poking the recipe builder directly. That's the boundary that broke, so that's where the guard belongs. Covering retry *and* ralph at attempts 2 and 3 means whoever refactors respawn next gets four loud failures instead of a mystery empty bead.

I re-ran it here, including with your one-line fix removed — all four cases fail exactly as you described, so the guard isn't decorative. Full `internal/dispatch` is green and CI on `3976ad6` is clean. Merging as-is; nothing needed from you.

The verification notes in the description helped too — calling out that the remaining shard failure reproduces on pristine base saved me chasing it.


---

PR #4987, opened 2026-08-04
feat(dashboard): add Beads hierarchy and record-tier filtering

Sutyke:
## Summary

* Add a hierarchical list view to the Beads dashboard, organizing epics, tasks, subtasks, and deeper descendants using their parent relationships.
* Keep matching ancestors visible as context during search and filtering, while preserving beads with missing or cyclic parent relationships as clearly marked roots.
* Separate durable, history-backed issues from operational records such as ephemeral and no-history beads.
* Add an optional `tier` parameter to the Beads API:

  * `issues` returns history-backed durable issues.
  * `operational` returns ephemeral or no-history records.
  * `all` preserves the combined view.
  * Omitting `tier` preserves existing API behavior.
* Update the native Dolt, DoltLite, CLI-backed, count, pagination, generated client, OpenAPI, and embedded dashboard paths to use consistent tier semantics.

This is needed because the existing flat list makes multi-level work difficult to understand, and operational bookkeeping records can obscure durable project work. The new hierarchy makes relationships visible while the tier selector keeps the normal issues view focused without removing access to operational records.

## Testing

* [ ] `make check`

  * Not run as one combined command in the available environment.
  * Equivalent affected checks completed:

    * 77 focused dashboard tests passed.
    * Dashboard source and test TypeScript type checks passed.
    * Production dashboard build passed.
    * `go test ./internal/api/... ./internal/beads/...` passed.
    * `git diff --check` passed.
* [ ] `make check-docs` if docs, navigation, or links changed

  * Not required: no authored Mintlify pages, navigation, or links changed. Generated OpenAPI reference schemas were updated.
* [ ] `make test-integration` if runtime, controller, or workflow behavior changed

  * Not required: this changes Beads list-query filtering and dashboard presentation, not controller or workflow execution behavior.

## Checklist

* [x] Linked an issue, or explained why one is not needed

  * No issue is linked. This is a self-contained dashboard and API usability improvement with focused regression coverage.
* [x] Added or updated tests for behavior changes

  * Added coverage for hierarchy construction, nested expansion, valid parent-child tree structure, missing parents, cycles, ancestor context, project grouping, tier filtering, API behavior, counting, and storage backends.
* [x] Updated docs for user-facing changes

  * Generated OpenAPI schemas and generated API clients include the new `tier` parameter. No separate authored documentation is required for the dashboard controls.
* [x] Called out breaking changes or migration notes

  * No breaking change or migration is required. Existing API clients that omit `tier` retain the previous combined logical issues behavior.


quad341:
Thanks for this, @Sutyke — there is clearly a lot of work here.

Sutyke:
Thanks for getting back to me.

I opened a dashboard PR on `gascity` because that’s what the docs and `gc dashboard` point to. Without feedback, I kept building privately as a separate app — and found that much easier, unfortunately what I complete I can't share due to licensing.

I later noticed `gascity-dashboard` already exists for exactly that: a standalone next dashboard with separate review/CI, meant to fold back into `gc` later. The split makes sense; it’s just hard to discover from outside. A short note in `gascity` docs/CONTRIBUTING (“UI → gascity-dashboard; API/embed → gascity”) would help the next person.

Please close this PR. Happy to continue on `gascity-dashboard` if that’s the right place.

PR #4989, opened 2026-08-04
fix(usage): emit compute facts while sessions are draining

haddad-daniel:
## Summary

Treat `state=draining` as compute-terminal for end-of-interval usage emission.

Pool `drain-ack-stop-pending` keeps beads **open** in `draining`; `drained` is typically stamped only at close, after the bead has left `OpenInfos()`. Without this, pool retirements under-count wall/token usage to zero when idle sleep is off.

## Context

Observed on a live city: closed pool sessions had `awake_started_at` but no `usage_compute_emitted_at`. Idle sleep is often off for interactive_resume pools, so drain/close is the only terminal transition.

Verified locally: after `gc runtime drain` + `drain-ack`, a compute fact was written for the draining interval.

Related: [#4994](https://github.com/gastownhall/gascity/pull/4994) (live sweep; `draining` is still not live), [#5025](https://github.com/gastownhall/gascity/pull/5025) (closed-race recovery if the bead leaves the open set before a tick sees `draining`).

## Test plan

- [x] `go test ./cmd/gc -run 'TestIsComputeTerminalState|TestLiveModelSweepCandidate|TestComputeFactGetCandidate'`
- [x] Live: drain-ack a pool session → compute fact in `.gc/usage.jsonl`

### DIFF
--- a/cmd/gc/usage_compute.go
+++ b/cmd/gc/usage_compute.go
@@ -45,15 +45,16 @@ const liveModelSweepMinInterval = 30 * time.Second
 // isComputeTerminalState reports whether a session state marks the end of an
 // awake interval, at which a compute fact should be emitted. It covers every
 // non-running lifecycle endpoint the controller's open-bead scan can observe:
-// idle-sleep (asleep), controller drain (drained), retirement (archived),
-// operator suspend (suspended), and crash-loop quarantine (quarantined). A
-// session closed directly from active without first passing through one of
-// these open states is the known v0 scan limitation (see
-// engdocs/design/usage-facts-v0.md).
+// idle-sleep (asleep), controller drain (draining — the open-set state while
+// drain-ack-stop is pending; drained is usually only stamped at close),
+// retirement (archived), operator suspend (suspended), and crash-loop
+// quarantine (quarantined). A session closed directly from active without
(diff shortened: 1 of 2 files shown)


haddad-daniel:
Closing in favor of #5025.

**Why this PR is the wrong fix**

`draining` is intentionally transitional (#4994 treats it as neither live nor terminal). The pool under-count is not “forgot to list draining,” it’s the drain-ack finalize path:

1. `finalizeDrainAckStopPendingSessions` closes the bead mid-tick with `ClosePatch("drained")` while still `state=draining` in the prior snapshot.
2. Snapshot reloads; the bead is gone from `OpenInfos()`.
3. `emitDueComputeFacts` runs later — so it never sees an open `draining` *or* open `drained`/`asleep` for that session.

Widening `isComputeTerminalState` to include `draining` does not catch same-tick drain-ack closes (the common pool retire path). It only affects multi-tick linger in `draining`, which fights the designed transitional taxonomy and risks premature interval commit.

**Correct gap**

Known v0 open-set under-count (`usage-facts-v0.md`) + the closed-row recovery #4994 deliberately deferred → #5025 (bounded recently-closed sweep).

Issue #4448, opened 2026-07-19
Drained pool sessions are not reaped; their open session bead silently starves pool demand at cap


swedeinasia-flow:
## Observed (three live occurrences in 48h on one city)

With a pool at `max_active_sessions=N`, a worker session that finishes its step and drains can leave its **session bead open with no live session behind it**. The demand snapshot then reports `desired_count=N open_count=N work_requested=true` and refuses to spawn — ready, routed work sits unclaimed with **zero** live workers. At cap=1 this halts the lane outright (our canary froze mid-molecule); at cap=6 each stale bead silently costs a slot.

Reconciler trace excerpt at the freeze:

```
{"desired_count":1,"open_count":1,"pool_desired":1,"work_requested":true}
```

while `gc session list` showed no session for the pool and the next molecule step was ready+routed.

Related: `gc runtime drain-ack` acknowledges ("Controller poked for immediate stop") but the session can still show `STATE=active` 14+ minutes later, holding the slot.

## Operator workaround

`bd list --status open --type session` → find the bead with no matching live session → `gc session close <id>`. Demand recovers and a fresh session spawns in seconds. We have had to do this three times in two days under normal operation.

## Question / suggestion

Current main has new stale-session-bead handling in `cmd/gc/build_desired_state.go` (`staleNonExpandingPoolSessionBeadInfo`, "Don't re-add stale session beads..."). Does that path also cover the **post-drain-ack unreaped shell** case above (session bead open, tmux/process gone or idle-empty)? If not, a reaper pass that closes session beads whose runtime session is verifiably gone — or counting only sessions with a live runtime toward `open_count` — would remove this failure class. Can supply full traces.

nl0:
Second and third field occurrences, plus a source trace of why the husk is unreapable. Environment: gc v1.4.0, macOS, tmux provider, ~20-session fleet.

**Occurrences (2026-08-27, twice in one day):** a pool seat's session bead sat `state=active` with no tmux pane and no process — desired-state counted the demand as served, so no replacement spawned. One starved a one-shot seat's queue; the other blocked a high-priority build for ~1.5h. Notably these husks were **never-drained** (`state=active` throughout) — a stronger variant than the post-drain-ack shell in the original report: the husk is unreapable even without reaching any terminal state.

**Why no lane reaps it** (all citations verified at both v1.4.0 and current main):

[...]

Issue #3689, opened 2026-06-23
TMUX session recreation fails due to alias conflicts when TMUX server is missing/crashed


reidbaker-agent:
### Description
If the system's `tmux` server crashes or is stopped while the supervisor is running, the supervisor correctly detects the active sessions as `asleep (reason: runtime-missing)`.

However, when trying to wake/rebuild the sessions, the reconciler fails because of alias conflicts in the database:
`session beads: alias "mayor-2" for mayor-2 unavailable: session alias already exists: "mayor-2" already belongs to rc-wisp-7klxp5`

The reconciler gets stuck trying to wake them until the city is manually stopped and restarted.

### Steps to Reproduce
1. Start a city supervisor and wake a few scaled agent sessions (e.g. `mayor-1`, `mayor-2`).
2. Force-kill the `tmux` server (e.g., `killall tmux`).
3. Check `gc session list` and observe sessions are `asleep (reason: runtime-missing)`.
4. Trigger a wake request for the sessions and inspect supervisor logs.

### Expected Behavior
If the session's runtime is missing, the reconciler should clean up or overwrite the stale runtime alias mapping in the database when recreating the session.

### Actual Logs
```
session beads: alias "mayor-2" for mayor-2 unavailable: session alias already exists: "mayor-2" already belongs to rc-wisp-7klxp5
session beads: alias "mayor-1" for mayor-1 unavailable: session alias already exists: "mayor-1" already belongs to rc-wisp-bu613x
```

Issue #2885, opened 2026-06-01
bug: session: singleton pool canonical alias squatted forever by asleep (non-closed) predecessor


sjarmak:
## Summary

A non-expanding (singleton) pool's canonical session alias can be squatted indefinitely by a *superseded* predecessor session bead that is asleep/dormant (e.g. reason `city-stop`) rather than `closed`. The live successor never reclaims the canonical alias and the reconciler logs a "deferring singleton pool identity normalization" line every tick, forever.

## Symptom

`buildDesiredState` emits, on every reconcile tick:

```
buildDesiredState: pool "<rig>/<pool>": deferring singleton pool identity normalization for bead <live>: normalizing singleton pool identity for bead <live> to "<rig>/<pool>": session alias already exists: "<rig>/<pool>" already belongs to <old>
```

Observed on a `[[named_session]]`-backed singleton maintenance pool: 20+ successive pool session beads each squatted by a single 14-day-old asleep (`city-stop`) predecessor, spamming the supervisor log 49–59× per live session. The live successor runs fine but under the non-canonical alias `<rig>/<pool>-<sessionid>` instead of the canonical `<rig>/<pool>`. `gc doctor` may additionally flag `order-firing-current` stale when the reconcile loop is doing this redundant work under load.

## Reproduction

1. Run a city in supervisor mode with a singleton (non-expanding) pool that uses canonical singleton-pool identity (`UsesCanonicalSingletonPoolIdentity()` true) — e.g. a `[[named_session]]`-backed maintenance pool.
2. Let a pool session run, then `gc city stop` (its session bead goes `asleep`, reason `city-stop` — **not** `closed`).
3. Start the city again. A fresh pool session spawns for the same singleton pool.

Result: the fresh session should normalize to the canonical alias → got: it cannot, because the asleep predecessor still reserves the alias; reconciler defers normalization every tick indefinitely.

## Root cause

`internal/session/names.go` `ensureSessionAliasAvailable` releases a held alias for only two predecessor states: `failedCreateIdentityReleased(b)` and `b.Status == "closed"`:

```go
if failedCreateIdentityReleased(b) { continue }
if b.Status == "closed" { continue }
...
if strings.TrimSpace(b.Metadata["alias"]) == alias {
    return fmt.Errorf("%w: %q already belongs to %s", ErrSessionAliasExists, alias, b.ID)
}
```

[...]

PR #4990, opened 2026-08-04
fix(session): release the slot alias for drained and runtime-missing holders

julianknutsen:
## Problem

A pool session that has **drained** (terminal `state=drained`) or whose **runtime vanished** (durable `sleep_reason=runtime-missing`) keeps reserving its slot alias forever. `ensureSessionAliasAvailable` releases identities only for failed creates, so every later pool create for that slot fails with `session alias already exists` and wedges at `start-pending`.

Net effect: **once a slot's first occupant finishes, that pool can never spawn into it again.** In production this shows up as whole pools going silent — routed-ready steps sit unclaimed for hours while the reconciler logs a steady stream of alias conflicts naming sessions that finished long ago.

## Fix

Release both dead classes in `deadIdentityReleased`, mirroring the existing `failedCreateIdentityReleased` semantics. Live open holders still block.

The release keys on the **terminal `state=drained` record, not the `drain_at` stamp**. That distinction is load-bearing: a drain request canceled for assigned work (`clearDrain` removes only the runtime `GC_DRAIN*` keys, never the bead stamp) leaves `drain_at` on a live, working session. Keying on the stamp would make that session unmatchable, so the reconciler would re-judge it orphaned every tick and release its claimed step mid-turn — a drain→cancel→orphan loop that kills working sessions. The regression test pins that case.

## Tests

`TestEnsureSessionAliasAvailableReleasesDeadHolders`, 4 subtests:

- drained holder releases the alias
- **canceled-drain live holder (`drain_at` without `state=drained`) still blocks**
- runtime-missing holder releases the alias
- live open holder still blocks

Verified red before / green after on this base. `go test ./internal/session/` and `go vet ./internal/session/` clean.

Both constants (`StateDrained`, `LifecycleReasonRuntimeMissing`) already exist on main; no new dependencies.

---
- Related to #2885 — extends the same alias-release guard chain in `ensureSessionAliasAvailable` that the filed issue points at as the root cause, adding release for terminal `state=drained` holders and `sleep_reason=runtime-missing` husks; it does not cover the asleep `city-stop` predecessor described there, so that singleton-pool canonical-alias squat still stands
- Related to #3689 — makes a session whose runtime has vanished (durable sleep_reason=runtime-missing) stop reserving its slot alias, so a fresh create can reclaim that name after the tmux server dies instead of failing with "session alias already exists"; the same change also releases aliases held by drained sessions
- Related to #4448 — covers a second way an unreaped drained or runtime-missing session bead squats its pool slot — the alias reservation that wedges later creates at start-pending — but it does not reap the bead or change the demand accounting, so the open_count starvation described in that issue is untouched

### DIFF
--- a/internal/session/names.go
+++ b/internal/session/names.go
@@ -401,6 +401,28 @@ func failedCreateIdentityReleased(b beads.Bead) bool {
 	return strings.TrimSpace(b.Metadata["state"]) == string(StateFailedCreate)
 }
 
+// deadIdentityReleased reports whether an open session bead's identity no
+// longer reserves its alias because the session is finished-by-intent or its
+// runtime is gone: a drained session (drain_at stamped at drain) has completed
+// its pool turn, and a runtime-missing session (durable sleep_reason written by
+// the reconciler when the process vanished) is a husk awaiting cleanup. Both
+// otherwise squat their slot alias forever, wedging every subsequent pool
+// create for the same slot at start-pending (observed post-migration: pool
+// spawns starved for hours behind drained predecessors' aliases). Mirrors the
+// failed-create release semantics.
+func deadIdentityReleased(b beads.Bead) bool {
+	// state=drained is the TERMINAL drained record. The drain_at stamp alone is
+	// NOT sufficient: a drain request that was canceled (assigned work) leaves
+	// drain_at on a live, working session — releasing its alias then makes the
+	// session unmatchable, so the reconciler re-judges it orphaned every tick
+	// and eventually kills it mid-turn (observed: the apply-fixes step's
+	// workers dying in a drain/cancel loop).
+	if strings.TrimSpace(b.Metadata["state"]) == string(StateDrained) {
+		return true
+	}
+	return strings.TrimSpace(b.Metadata["sleep_reason"]) == LifecycleReasonRuntimeMissing
+}
+
 func continuityIneligibleConfiguredOwner(b beads.Bead, selfOwner string) bool {
 	if failedCreateIdentityReleased(b) {
 		return false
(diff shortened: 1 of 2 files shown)


julianknutsen:
Closing this — a review council caught that it is both **unsafe** and **redundant**, and I verified both against this base.

**Main already does this, correctly.** `isPoolSessionSlotFreeable` (`cmd/gc/session_state_helpers.go:59`) already frees pool slots for drained and `runtime-missing` holders — but it gates the reason set behind an explicit `state == "asleep"` check, and its doc comment says exactly why: *"deny-by-default for unknown or missing reasons so writes that land in state=asleep without a known reason (legacy beads, regressions, write races) cannot silently free their slot."* `cleanupDeadRuntimeSessionCorpses` (called from `city_runtime.go:565,1133`) then **closes** the corpse, so the alias is released with the one-open-holder-per-alias invariant intact.

**My version dropped that state gate**, which is a live-session hazard, not a theoretical one. `sleep_reason` is not cleared when the reconciler heals a bead back to awake: `healStatePatchWithRollbackInfo` only clears it when the target state is `asleep`, so after a transient tmux/provider liveness blip a *running* session keeps `sleep_reason=runtime-missing` for its entire awake interval. My predicate would report that live session's alias as free and hand its public handle to a different bead.

Two further defects: releasing the reservation without closing the holder leaves **two open beads carrying the same alias**, which makes `ResolveSessionID` return `ErrAmbiguous` — the alias then resolves to nothing at all; and my `state == "drained"` test misses the canonical durable shape (`state=asleep` + `sleep_reason=drained`) that `isDrainedSessionBead` handles, so it may not even have fixed the squat it targeted.

**Root cause of my misdiagnosis:** the wedge I observed was on a fork with a real split sessions store, where the reconciler's orphan-release path could not see the class store. That is a fork-specific routing bug, not an upstream one. My red/green "proof" only showed that a new test fails without the new code — which is true of any new behavior, and is not evidence the bug exists here.

No action needed upstream. Sorry for the noise.

PR #3954, opened 2026-07-05
fix(session): persist claude resume key from its SessionStart hook

wbern:
## The juicy parts

A `claude` session never persists a resume key, so `wake_mode=resume` has nothing to resume: every recycle — config-drift restart, runtime-missing wake, any reconciler bounce — silently starts a **fresh conversation**. What the user loses is that session's accumulated context, with no error and nothing logged.

All three capture paths are closed for claude in `main` today: up-front minting was removed in `b01cfb8b0` (`claude --session-id` was unsupported at the time), hook-stdin capture has been codex-only since #3220, and claude was never in the history-derive allow-list. `b01cfb8b0` pointed at "resume metadata must come from the provider after startup" as the replacement — claude was simply never wired to it.

The fix is one predicate gating one branch, existing keys are never overwritten, the env path is untouched, and a successful persist now emits a one-line diagnostic so an operator debugging a fresh wake can see resume was armed. @julianknutsen revalidated this against `main` on 2026-07-23, confirmed the cherry-pick is clean and the focused hook session-key tests pass, and described it as "the session-key persistence half of the Slack delivery regression."

## What this fixes

A `claude` session never persists a resume key, so `wake_mode=resume` silently
starts a **fresh conversation on every recycle** (config-drift restart,
runtime-missing wake, any reconciler bounce).

## How the gap arose

Claude resume used to work — a correct cleanup left it without a replacement:

1. **Originally** — a `session_key` was minted up front and the reconciler
   injected `--resume`; claude resumed (too eagerly, in fact: #81, #112).
2. **`b01cfb8b0`** — up-front minting was removed for claude: `claude
   --session-id <uuid>` is unsupported, so the builtin profile dropped
   `session_id_flag`. The intended replacement is stated on `session_id_flag`
   itself — "resume metadata must come from the provider after startup."
3. **The post-start capture was wired for codex** (#3220, hook-stdin) **and the
   env-based providers — claude was never included.** So minting was retired
   without a claude replacement, and `session_key` has gone unwritten for claude
   since.

Current state — every post-start capture path is closed for claude:

| path | state for claude |
| --- | --- |
| up-front mint | removed in `b01cfb8b0` (`--session-id` unsupported) |
| hook stdin (`persistPrimeHookProviderSessionKey`) | codex-only since #3220 |
| history-derive | never in the allow-list (kimi / opencode / pi / antigravity) |

With all three closed, `resolveSessionCommand` has no key and relaunches the base
command — a new conversation every wake.

## Fix

[...]

PR #4010, opened 2026-07-07
fix(prime): gate SessionStart hook on live sessions

jonathanpberger:
Summary
- Gate gc prime --hook SessionStart output behind a real managed Gas City session context.
- Require GC_SESSION_ID and GC_SESSION_NAME to match an open session bead in active, awake, creating, or start-pending state before injecting hook context.
- Keep explicit gc prime fallback behavior unchanged and keep non-SessionStart hook behavior unchanged.

Source / behavior before
- Active Codex hooks in td-core and td-city run GC_MANAGED_SESSION_HOOK=1 GC_HOOK_EVENT_NAME=SessionStart gc prime --hook --hook-format codex.
- td-core resolves as a rig under td-city, so ordinary Codex sessions in td-core received the default Gas City Agent prompt even without a managed Gas City session.

Behavior after
- Unmanaged SessionStart hook returns no hook output.
- Real managed sessions still receive hook context once the live session bead check passes.
- Explicit gc prime with no city still emits the default fallback prompt.

Verification
- CGO_ENABLED=0 go test ./cmd/gc -run ^TestDoPrime
- CGO_ENABLED=0 go build -o /tmp/gc-xgc-hook-gate ./cmd/gc
- From /Users/jpb/workspace/tuxedodrive/td-core, unmanaged SessionStart with cleared GC_SESSION_ID/GC_SESSION_NAME exited 0 with empty stdout.
- Explicit fallback check still printed the default Gas City Agent prompt.

Notes
- Direct push to gastownhall/gascity was denied for jonathanpberger (viewerPermission=READ), so this PR is opened from the existing jonathanpberger/gascity fork.

PR #4991, opened 2026-08-04
fix(prime): persist the provider resume key before the SessionStart output gate

julianknutsen:
## Problem

On a **live managed session**, prime's SessionStart path returns as soon as it decides not to emit prompt output — the gate at `cmd/gc/cmd_prime.go` returns `0` *before* `runHookSideEffects()` is reached.

`runHookSideEffects()` is where the provider's resume key is captured and persisted. So for precisely the sessions that *have* a runtime to resume, the key is never written. A later resume then restarts the agent from scratch instead of continuing its thread — silently, since nothing errors.

## Fix

Run the side effects **before** the output gate. Capturing durable state should not be conditional on whether this particular invocation prints a prompt.

## Why now

This lands on top of #3954, which added the Claude resume-key capture arm. That fix supplies the key this path was dropping, so the two together are what make managed-session resume actually work end to end.

## Tests

`cmd/gc/main_test.go` gains a regression asserting the provider key is persisted before the SessionStart output gate (fails on this base without the fix: `session_key = "", want provider key persisted before SessionStart output gate`).

Full `cmd/gc` prime/hook suite green; `go vet ./cmd/gc/` clean.

### DIFF
--- a/cmd/gc/cmd_prime.go
+++ b/cmd/gc/cmd_prime.go
@@ -195,7 +195,9 @@ func doPrimeWithHookFormatOpts(args []string, stdout, stderr io.Writer, hookMode
 		hookContext = readPrimeHookContext()
 		suppressHookPrompt = managedSessionHookPromptAlreadyDelivered(hookContext)
 	}
-	// In non-strict mode, hook side effects fire eagerly (existing behavior).
+	// In non-strict mode, hook side effects fire eagerly for every event. The
+	// SessionStart live-session gate below controls prompt output only; provider
+	// resume-key persistence has its own session and provider guards.
 	// In strict mode, we defer them until after strict checks pass so that a
 	// failing --strict invocation does not update provider resume metadata for
 	// failed agent resolution or template validation.
@@ -205,7 +207,7 @@ func doPrimeWithHookFormatOpts(args []string, stdout, stderr io.Writer, hookMode
 		}
 		persistPrimeHookProviderSessionKey(hookContext.ProviderSessionID, stderr)
 	}
-	if !strictMode && !primeHookSessionStart(hookContext) {
+	if !strictMode {
 		runHookSideEffects()
 	}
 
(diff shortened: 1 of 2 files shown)


quad341 (approved):
@julianknutsen — thank you for this one. Hoisting `runHookSideEffects()` above the output gate is the right call, and the framing in your description is the part I appreciate most: capturing durable state shouldn't be conditional on whether *this particular* invocation happens to print a prompt. The code now reads like that invariant.

The regression test earns its keep. I reverted your `cmd_prime.go` hunk locally and watched it fail with exactly the `session_key = ""` message you quoted, then pass again with the fix — so it's genuinely pinned, not green by construction.

One note for the record, not a change request: the gate polarity in the description is flipped. `primeHookHasLiveManagedSession` returns true for live sessions and the early return fires on `!live`, so the sessions dropping the key are the ones the gate *rejects* — which your test proves by zeroing `GC_SESSION_NAME`. That makes the fix land better than advertised: a wake from `asleep` fires SessionStart with the bead outside the live set, which is precisely when a resume key matters.

Merging as-is. Nice catch, and thanks for pairing it with #3954 so resume works end to end.


---

julianknutsen:
Closing this — the premise is inverted, and the change would silently revert an existing guard.

**The stated bug does not exist.** On main, `cmd/gc/cmd_prime.go:226` returns early only when `!primeHookHasLiveManagedSession(cityPath)` — i.e. only when there is **no** live managed session. Lines 230-232 then run `runHookSideEffects()` for exactly the live-managed case. So live managed sessions already persist their provider resume key. My description claimed the opposite ("for precisely the sessions that have a runtime to resume, the key is never written").

**What the diff actually does** is hoist `runHookSideEffects()` above the gate so it also runs for sessions that are *not* live-managed — dormant or closed beads. That is the case #4010's ordering guard exists to prevent: a write-once resume key getting pinned onto a dormant/closed session bead.

If there is a real gap here it is the narrow one the council named — SessionStart invocations where the live-managed predicate is false *and* the session genuinely has a runtime worth resuming — but that needs a proper diagnosis and a guard that keeps #4010's invariant, not this hoist. Not worth carrying in this shape.

Issue #4859, opened 2026-07-31
bug: dispatch: ralph check env drops GC_HOME, so nested gc bd resolves imports against the empty city-local cache


24601:
## Symptom

In a city whose root `pack.toml` has remote imports (installed and pinned in
`packs.lock`, cached under `~/.gc/cache/repos/<hash>`), every ralph
`[steps.check]` exec script that calls `gc bd ...` fails config load:

```
gc bd: loading config: city import "missions": import "gc": remote import
https://github.com/gastownhall/gascity-packs/tree/main/gascity is locked but
not cached at <city>/.gc/cache/repos/<hash>; run "gc import install"
```

`gc import install` does not fix it (it populates the user-global cache,
which is already populated). Every gate in every workflow fails; with the
first-party gate scripts suppressing stderr, the attempt log shows only
`gc bd show <id> failed`, so this is expensive to diagnose.

## Root cause

- `internal/convergence/condition.go` (`ConditionEnv.Environ()`, ~79-102):
  the check env is a deliberate whitelist and sets `HOME=<cityPath>` to
  sandbox gate scripts from the controller's home directory. `GC_HOME` is
  not passed.
- `internal/config/implicit.go` (`ImplicitGCHome()`, ~76-86): nested `gc`
  invocations resolve `GC_HOME` from `$HOME/.gc` when `GC_HOME` is unset -
  which under the check env is `<city>/.gc`.
- `internal/config/pack_include.go` (`resolveLockedRemoteImport`, ~172-181,
  error at ~481): the locked import is validated against
  `<city>/.gc/cache/repos/<hash>`, which nothing populates, so config load
  fails before any `bd` operation runs.

Verified by exact env replication: same command succeeds with
`GC_HOME=$HOME/.gc` (or with the hash symlinked into `<city>/.gc/cache`),
fails without.

## Fix candidates

**A (recommended): thread `GC_HOME` through the check env.** Add a `GCHome`
field to `ConditionEnv`, populated in `runRalphCheck`
(`internal/dispatch/ralph.go`) from the controller's resolved GC home, and
emit `GC_HOME=` in `Environ()`. Keeps the `HOME=<cityPath>` sandbox intact;
`GC_HOME` points only at gc-owned state. Precedent: #2558 widened this same
whitelist with `GC_MOLECULE_DIR`/`GC_ARTIFACT_DIR` for the same subprocess.
A negative test can pin that no other ambient env leaks (B28, error-context wrapping).

[...]

Issue #3506, opened 2026-06-15
gchome.Default() falls back to shared /tmp/.gc when HOME unresolved — cross-process city collision risk


iwata-1116:
## Summary

`gchome.Default()` resolves the machine-local Gas City state dir. When `GC_HOME` is unset **and** `os.UserHomeDir()` returns an error or empty string, it falls back to a fixed shared path:

```go
// internal/gchome/gchome.go
home, err := os.UserHomeDir()
if err != nil || home == "" {
    return filepath.Join(os.TempDir(), ".gc")   // = /tmp/.gc on most systems
}
```

`/tmp/.gc` lives directly under a shared, typically world-writable tempdir. This is a dangerous fallback:

- **Cross-process / cross-user collision**: every process hitting this branch shares the *same* `/tmp/.gc`. State written by one can be read/clobbered by another.
- **False city detection**: a `.gc` created under the tempdir parent can be picked up by unrelated processes scanning for city state, polluting otherwise-clean test/CI runs.

## Location

`internal/gchome/gchome.go` — `Default()` (the `os.TempDir()` fallback branch).

## Proposed fix

Either:
- **(i) fail-fast** — return an error / empty string so the caller aborts rather than silently writing to a shared dir, or
- **(ii) process-unique dir** — if a fallback is truly required, use `os.MkdirTemp("", "gc-home-*")` so each process gets an isolated directory.

## Acceptance

Unit test: with `GC_HOME` unset and `HOME` empty, `Default()` must **not** create or return `/tmp/.gc`, and must not induce false city detection under a shared tempdir.

PR #4992, opened 2026-08-04
fix(convergence): preserve GC_HOME for gate subprocesses

julianknutsen:
## Problem

Gate scripts run with `HOME` deliberately sandboxed to the city directory, so a gate cannot write into the operator's real home. But `gc` resolves its **machine-level** state directory as `HOME/.gc` when `GC_HOME` is unset (`internal/gchome/gchome.go`).

So any `gc` invoked from a gate script resolves its cache and registry to `<city>/.gc` instead of the machine's `~/.gc` — a second, divergent copy of machine state that nothing else in the fleet reads. Silent: the gate succeeds, it just populated the wrong directory.

## Fix

Pass `GC_HOME` through explicitly to gate subprocesses. The sandboxed `HOME` stays sandboxed; machine-level state stays machine-level. Resolution order mirrors `gchome`: explicit `GC_HOME` → `HOME/.gc` → temp fallback.

## Tests

`condition_test.go` asserts `GC_HOME` is present in the gate subprocess environment (fails on this base without the fix: `missing env var GC_HOME`).

`go test ./internal/convergence/` and `go vet` clean.

---
- Related to #4859 — adds GC_HOME to the gate/check subprocess environment in ConditionEnv.Environ(), so a nested gc resolves its machine-level cache and registry against the operator's real ~/.gc instead of the empty city-local one, while leaving the sandboxed HOME intact — this is the issue's recommended fix candidate A, implemented by resolving GC_HOME inside the convergence package rather than threading a field from runRalphCheck

### DIFF
--- a/internal/convergence/condition.go
+++ b/internal/convergence/condition.go
@@ -24,6 +24,19 @@ const (
 	textFileBusyRetryDelay    = 25 * time.Millisecond
 )
 
+// conditionGCHome resolves the gc state directory for gate subprocesses. Gate
+// HOME is intentionally sandboxed to the city, so it cannot also be used for
+// gc's machine-level cache and registry state.
+func conditionGCHome() string {
+	if value := strings.TrimSpace(os.Getenv("GC_HOME")); value != "" {
+		return value
+	}
+	if home, err := os.UserHomeDir(); err == nil && strings.TrimSpace(home) != "" {
+		return filepath.Join(home, ".gc")
+	}
+	return filepath.Join(os.TempDir(), ".gc")
+}
+
 // conditionPATH resolves the tool directories gate scripts actually need.
 // This keeps the env narrow while ensuring gate scripts use the same bd/gc
 // binaries as the running city instead of whatever older copy happens to live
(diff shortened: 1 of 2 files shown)


quad341 (approved):
@julianknutsen — great catch, and the writeup made it fast to confirm: gate `HOME` is pinned to the city, `gchome` falls back to `HOME/.gc`, so every `gc` invoked from a gate has been quietly writing machine-level registry and cliauth state into `<city>/.gc`. A bug whose only symptom is a directory in the wrong place is exactly the kind that survives for months. The test that fails on base with `missing env var GC_HOME` is what sold it.

One thing I'll take on my side rather than send back to you: the last-resort `os.TempDir()/.gc` is the specific path #3506 removed — `internal/bootstrap/bootstrap.go` has a comment about it being shared and world-writable, and about city scans mistaking it for a real city. You mirrored `gchome`'s documented resolution order faithfully; that branch just happens to differ from its own doc comment. I'm swapping the helper body for `gchome.ResolveReadOnly().Path()`, which gives you the same order plus the process-unique fallback and removes the duplication, and adding a test pinning that the fallback isn't the shared path.

Your branch, your commit, your credit — I'll push the tweak on top, re-run build/vet/convergence + gchome tests, and merge.


---

PR #4994, opened 2026-08-04
fix(usage): sweep model usage from live sessions, not only at retirement

julianknutsen:
## Problem

Model-usage facts are minted **only when a session retires**. The single
model-fact emitter on the controller tick lives in
`emitDueComputeFacts` (`cmd/gc/usage_compute.go`) and is gated on
`isComputeTerminalState` — `asleep` / `drained` / `archived` / `suspended` /
`quarantined`. A session that is still awake is never even `Get`, because
`computeFactGetCandidate` is the only pre-Get filter and it requires a
terminal state.

The user-visible effect: **"model calls today" undercounts every session
that is still running.** A long-lived agent burns tokens for hours and
contributes nothing to the day's totals until it finally closes, and an
agent awake across a day boundary bills its entire interval to the wrong
day when it does. On a fleet where pool-routed agents self-drive for long
stretches, the live portion of spend is invisible.

## Fix

The reconcile tick now sweeps awake sessions incrementally, beside the
existing terminal lane:

- `isLiveModelSweepState` / `liveModelSweepCandidate` select awake rows
  from the reconcile snapshot. They are deliberately **disjoint** from
  `isComputeTerminalState` (asserted in the tests), so every session is
  handled by exactly one lane.
- `emitDueComputeFacts` routes each loaded bead through one
  `processSessionBead` step: awake beads take the live model-usage sweep,
  terminal beads take the **unchanged** compute-fact + terminal-sweep path.
  The snapshot loop's only behavioral change is that a live candidate now
  also earns a `Get`.
- `sweepLiveSessionModelUsage` records the interval's model facts **without
  closing the interval**: neither `usage_compute_emitted_at` nor
  `usage_model_swept_at` is stamped, so the real end-of-interval compute
  fact and terminal sweep still happen later exactly as before. Repeat
  ticks are made idempotent by the **already-persisted invocation cursor**,
  not by a marker — which is the correct mechanism here, since a live
  session is legitimately a candidate on every single tick.

Because a live session is re-examined every tick, transcript discovery
would otherwise repeat its bounded rollout scan indefinitely. Two additive
`internal/worker` entry points split discovery from extraction:

- `Factory.DiscoverSweepTranscript` — resolves the path under the same
  bounded keyed/keyless rules as `SweepSessionModelUsage`, without reading.
  A keyless Codex scan clouded by an I/O fault returns no path so the tick
  retries rather than trusting an ambiguous result.
- `Factory.SweepSessionModelUsageAtPath` — the same cursor-guarded
  extraction, fact emission, OTel metrics, and cursor persistence, against
  an already-resolved path.

Both delegate to a new shared `sweepResolvedTranscript`, which is
`SweepSessionModelUsage`'s own post-discovery body lifted out verbatim — so
the cursor, metrics, and settle semantics cannot drift between entry
points. `CityRuntime.liveSweepTranscriptPaths` then memoizes the resolved
path per `(session id, awake epoch, provider session key)`. A new awake
epoch or a replacement conversation resolves its own rollout instead of
reusing a stale path.

## What was deliberately left out

This change was split out of a larger internal commit that also carried a
storage refactor. **The routed-enumeration block is intentionally not
here**, along with its `internal/classdb/sessions` import and the
`processed map[string]bool` that existed only to de-duplicate against it.

That block re-listed the whole sessions-class store each tick
(`session.ListAllSessionBeads` with `IncludeClosed: true`, `TierBoth`,
`AllowScan: true`) to reach rows the open snapshot cannot supply. It only
does anything on a city that has routed `[beads.classes.sessions]` to its
own store — where retired sessions are *closed* and therefore vanish from
the open reconcile snapshot. On `main`'s single-store topology it is dead
weight: an unconditional full-store scan added to a synchronous reconcile
tick, guarded by a routing check that is always false. It also solves a
different problem (recently-*closed* rows going unaccounted) than the one
this PR fixes (*live* rows going unaccounted).

Everything carried here is single-store safe and reaches live sessions
through the snapshot the tick already has, adding no new store enumeration.

## Tests

- **`TestEmitDueComputeFactsSweepsLiveSessionModelUsage`** (new) — the
  regression, on plain single-store `main` using the established
  `writeCodexRolloutForSweep` / `usage.NewLocalSink` harness. An awake
  codex session present in the open snapshot:
  - **tick 1** bills both transcript invocations (2 model facts, **0**
    compute facts), advances the cursor to `total:450`, and leaves **both**
    interval markers unset so the terminal lane is not pre-empted;
  - **tick 2**, with no transcript activity, appends nothing;
  - **tick 3**, after one invocation is appended, bills only that delta.

  The idempotency assertions deliberately count **raw sink lines** via a
  new `rawSinkKindCount` helper rather than `usage.ReadFacts`, because
  `ReadFacts` collapses replays by `IdempotencyKey` at read time and would
  pass even if a tick re-recorded work the cursor should have skipped.

- **`TestLiveModelSweepCandidate`** (new) — pins the live/terminal split,
  including the assertion that no state is ever both.

- `writeCodexRolloutForSweep` now takes its session key from a shared
  `codexSweepSessionKey` const instead of a parameter. Every keyed sweep
  scenario passes the same value, and adding a third call site makes
  `unparam` (correctly) flag the parameter as constant.

**Red-then-green evidence.** With the production change reverted to
`main`'s terminal-only gate (`computeFactGetCandidate` alone, live beads
not routed to the sweep) and the new test applied:

```
--- FAIL: TestEmitDueComputeFactsSweepsLiveSessionModelUsage (0.00s)
    usage_compute_test.go:436: tick 1 model facts = 0, want 2 (a live session's invocations must bill before it closes); facts: []
FAIL
FAIL	github.com/gastownhall/gascity/cmd/gc	2.049s
```

With the fix restored, it passes.

**Verification** (`CGO_ENABLED=0`, linux/amd64):

- `go build ./...` — clean
- `go vet ./cmd/gc/ ./internal/worker/` — clean
- `gofmt -l` on the four touched files — clean
- `go test ./internal/worker/` — ok (84.6s)
- `golangci-lint run ./cmd/gc/ ./internal/worker/` — 0 issues
- `go test ./internal/worker/` — ok
- `go test ./cmd/gc/ -run 'Usage|ComputeFacts|ModelSweep' -count=1` — ok,
  12 tests including both new ones

An untargeted full-package `go test ./cmd/gc/` does not pass in this dev
environment: it times out at 600s inside the order-dispatch / managed-Dolt
path (`cmd/gc/order_store.go`, `cmd/gc/order_dispatch.go`), which is
unrelated to the usage lane and untouched here. The CI `cmd/gc process`
shards are the authoritative check for that package.

### DIFF
--- a/cmd/gc/city_runtime.go
+++ b/cmd/gc/city_runtime.go
@@ -108,6 +108,12 @@ type CityRuntime struct {
 	asyncStops         asyncStartTracker
 	demandSnapshot     *runtimeDemandSnapshot
 
+	// liveSweepTranscriptPaths memoizes the stable transcript path for each
+	// session awake epoch and provider session key. The worker factory is rebuilt
+	// per tick, so this process-lifetime cache avoids repeating bounded discovery
+	// without carrying a path across a replacement conversation.
+	liveSweepTranscriptPaths sync.Map // liveSweepTranscriptCacheKey -> transcript path
+
 	fsPressureConsecutiveSkips int
 	fsPressureEpisodeLogged    bool
 
--- a/cmd/gc/city_runtime.go
+++ b/cmd/gc/city_runtime.go
@@ -108,11 +108,12 @@ type CityRuntime struct {
 	asyncStops         asyncStartTracker
 	demandSnapshot     *runtimeDemandSnapshot
 
-	// liveSweepTranscriptPaths memoizes the stable transcript path for each
-	// session awake epoch and provider session key. The worker factory is rebuilt
-	// per tick, so this process-lifetime cache avoids repeating bounded discovery
-	// without carrying a path across a replacement conversation.
-	liveSweepTranscriptPaths sync.Map // liveSweepTranscriptCacheKey -> transcript path
+	// liveSweepMemos carries the live model-usage sweep's per-session memo: the
+	// resolved transcript path, whether discovery definitively found nothing, and
+	// the sweep-interval floor. The worker factory is rebuilt per tick, so this
+	// process-lifetime cache is what keeps a per-tick live lane from repeating
+	// bounded discovery and transcript reads for every awake session.
+	liveSweepMemos sync.Map // session bead id -> liveSweepMemo
 
 	fsPressureConsecutiveSkips int
 	fsPressureEpisodeLogged    bool
@@ -2188,7 +2189,9 @@ func (cr *CityRuntime) stopConfigWatcher() {
 // readiness quickly, so it skips the undesired-pool-session sweep (a heavy
 // candidate × store × status × identifier bd-read fan-out that, serialized on
 // the readiness path, can exceed the startup watchdog on a heavy-session city —
-// gastownhall/gascity#3288). The first steady-state tick performs the sweep.
+// gastownhall/gascity#3288) and the usage lane's live transcript sweep (bounded
+// per-session file discovery and reads across every awake session at once). The
+// first steady-state tick performs both.
 func (cr *CityRuntime) beadReconcileTick(ctx context.Context, result DesiredStateResult, sessionBeads *sessionBeadSnapshot, trace *sessionReconcilerTraceCycle, bootReconcile bool) {
 	desiredState := result.State
 	store := cr.cityBeadStore()
@@ -2214,8 +2217,11 @@ func (cr *CityRuntime) beadReconcileTick(ctx context.Context, result DesiredStat
 		result.SessionQueryPartial = result.SessionQueryPartial || sessionQueryPartial
 	}
 	// Emit any due compute usage facts by reusing the open-session snapshot this
-	// tick already loaded, rather than issuing a second redundant store scan.
-	cr.emitDueComputeFacts(ctx, sessionBeads.OpenInfos())
+	// tick already loaded, rather than issuing a second redundant store scan. The
+	// boot pass covers the whole fleet at once on the readiness path, so it takes
+	// only the marker-gated terminal lane and leaves the fleet-proportional live
+	// lane to the first steady-state tick.
+	cr.emitDueComputeFacts(ctx, sessionBeads.OpenInfos(), bootReconcile)
 	rigStores := cr.rigBeadStores()
 	assignedWorkBeads := result.AssignedWorkBeads
 	assignedWorkStoreRefs := result.AssignedWorkStoreRefs
--- a/cmd/gc/usage_compute.go
+++ b/cmd/gc/usage_compute.go
@@ -45,6 +45,17 @@ func isComputeTerminalState(state string) bool {
 	return false
 }
 
+// isLiveModelSweepState reports whether a session is currently awake and may
+// still append model invocations to its transcript. It is deliberately
+// disjoint from isComputeTerminalState.
+func isLiveModelSweepState(state string) bool {
+	switch session.State(strings.TrimSpace(state)) {
+	case session.StateActive, session.StateAwake:
+		return true
+	}
+	return false
+}
+
 // emitComputeFactForBead records one compute Fact for a session bead's
 // completed awake interval, exactly once per awake_started_at epoch. Returns
 // true when a fact was recorded. It is a no-op when the sink is discard/nil,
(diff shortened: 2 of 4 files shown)


julianknutsen:
Pushed `6b52f7bba4` addressing all four MAJOR review findings. Everything is in this branch's own new code — the terminal end-of-interval lane is untouched.

**1. Unthrottled per-tick transcript I/O on the synchronous reconcile tick.**
A live session is a candidate on *every* tick, so the lane's work was fleet-proportional **and** repeated at the tick cadence — a bounded rollout discovery scan plus a transcript tail read per awake session, with no floor and no budget check. Added `liveModelSweepMinInterval = 30 * time.Second`, following the neighbouring `scaleCheckDemandMinInterval` idiom (doc comment explains what floods without it and what the 30s costs: staleness far below the interval-scale lag this lane exists to fix, and nothing is dropped because the cursor-guarded sweep bills the whole pending batch when it next runs). The per-session loop also checks `ctx.Err()` so a cancelled tick stops promptly instead of working through the rest of the fleet.

**2. The live lane ran unguarded on the boot reconcile pass.**
`emitDueComputeFacts` now takes the `bootReconcile` signal `beadReconcileTick` already carries and skips the LIVE lane on that pass — the same trade the tick already makes for the undesired-pool-session sweep. The gate is applied both to the pre-`Get` candidate filter (so a live-only candidate costs zero Gets on boot) and to the fresh-bead route (so a session that woke since the snapshot is skipped too). The terminal lane is self-limiting via its persisted per-interval marker, so its cost profile is unchanged and it keeps running on boot.

**3. Discovery misses were never memoized and `settled` was discarded.**
The cache now holds a per-session `liveSweepMemo` rather than a bare path:
- A **transient** miss (keyed rollout not flushed yet, or a keyless scan clouded by an I/O fault) takes the same interval floor as a hit, so discovery is re-attempted on that floor instead of on every tick forever.
- `Factory.DiscoverSweepTranscript` now returns the `settled` verdict `SweepSessionModelUsage` already computes, so a **definitive** miss — an unregistered provider family, or a keyless codex session whose CLEAN workdir+window scan found nothing — is recorded once and never re-scanned for that awake epoch. Bounded blast radius: that epoch's usage is still recovered by the terminal sweep when the session closes, and a re-wake starts a fresh epoch that discovers again.
- Keying the memo by session id with the epoch/conversation *inside* the value also stops a long-lived session accumulating one cache entry per wake.

The hit path is unchanged.

**4. The new test was anchored to a hardcoded date.**
A live session has no `slept_at`, so its discovery window runs to the wall clock; once that span passes the 370-day directory cap (`codexByIDDayDirCap`) the `2026-06-15` fixture stops being discoverable and the test hard-fails for everyone on 2027-06-19 — indistinguishable from a real regression. The fixture now derives from `time.Now().UTC().Add(-time.Hour)`, with **both** the day directory and the `rollout-<localtime>-<key>.jsonl` filename built from that same value in `time.Local` (matching how codex names rollouts and how discovery parses them). Confirmed the cliff was real: aging the fixture 400 days reproduces the future failure exactly. Only this branch's new test changed — the terminal-lane tests pin their window with `slept_at` and are already date-safe.

### Tests

Three new cases plus the reworked one, all following the existing harness idiom:
- `TestEmitDueComputeFactsThrottlesLiveModelSweep` — a tick inside the floor does not re-sweep even with new usage waiting; the delta bills once the floor elapses.
- `TestEmitDueComputeFactsSkipsLiveModelSweepOnBootPass` — zero model facts and an untouched sweep memo on the boot pass, while a retired session in the same snapshot still gets its compute fact; the live session is swept on the very next steady-state tick.
- `TestEmitDueComputeFactsBacksOffUnresolvedLiveSweepDiscovery` — a transient miss arms the floor and is not re-attempted inside it, then resolves after it.
- `TestEmitDueComputeFactsStopsRetryingSettledLiveSweepMiss` — a definitive miss is never re-attempted, even with the floor cleared and a matching rollout on disk.

Each behavior was mutation-checked to go red when removed. Red-before for the reworked date-independent test, with the production live lane disabled:

```
--- FAIL: TestEmitDueComputeFactsSweepsLiveSessionModelUsage (0.00s)
    usage_compute_test.go:518: tick 1 model facts = 0, want 2 (a live session's invocations must bill before it closes); facts: []
```

### Gates

`go build ./...`, `go vet ./cmd/gc/ ./internal/worker/`, `go test ./internal/worker/`, `go test ./cmd/gc/ -run 'Usage|ComputeFacts|ModelSweep'` and `-run Reconcile` — all clean. Also ran `golangci-lint run ./cmd/gc/... ./internal/worker/...` (the repo's configured linter set, `unparam` included): **0 issues**. That surfaced three real hits in the new code, now fixed — a `misspell` on `cancelled`, a `staticcheck` De Morgan on the candidate filter, and an `unparam` on the `rawSinkKindCount` test helper whose `kind` argument was always `usage.KindModel` (now `rawSinkModelFactCount`).

PR #4216, opened 2026-07-13
fix(cmd/gc): lint no longer flags the documented pool-disable form as a named-session conflict (#4184)

rjgeng:
# PR draft — fix(cmd/gc): lint no longer flags the documented pool-disable form as a named-session conflict (#4184)

**Branch:** `fix/lint-named-session-pool-disable-form-4184` (cut from `upstream/main` @ `f604b1e7c4f47bd0599f4366140ea9df93aa1fe7`)
**Commit:** `43f63f409`
**Files touched:** `cmd/gc/cmd_lint.go` (+17), `cmd/gc/cmd_lint_test.go` (+37, new test).

## Title

fix(cmd/gc): lint no longer flags the documented pool-disable form as a named-session conflict (#4184)

## Body

Fixes problem 2 of #4184.

### Bug

`min_active_sessions = 0` + `max_active_sessions = 0` is upstream's
documented way to intentionally disable an agent's pool
(`TestValidateAgentsPoolMaxZeroIsValid` in `internal/config`) — it
genuinely suppresses pool spawns. But `gc lint`'s
`agentHasPoolControls` (`cmd/gc/cmd_lint.go`) treats any explicit
`min`/`max` setting as "this agent has pool control," so lint re-emits
`named_session "X" targets pool-controlled agent "X"; remove pool
settings from named-session templates...` for the disable form exactly
as it does for a real pool — even though the warning's own
prescription (remove the pool settings) doesn't apply: the settings
are already the maximally-disabled state.

### Fix

Added `agentPoolExplicitlyDisabled`, checked before the existing pool
checks in `agentHasPoolControls`. Recognizes the exact documented
disable shape (`min=0 AND max=0`, no `scale_check`/`namepool`/
`namepool_names`) and treats it as "no pool," matching
`ValidateAgents`' own semantics for the same shape.

### Scope note (deliberately narrow)

#4184 describes two problems. This PR only fixes problem 2 (the false
positive on the disable form). Problem 1 — that the warning's
prescribed fix for a *real* pool conflict (deleting the pool settings
entirely) resolves to an implicit `{0,-1}` pool rather than actually
ending pooling — is, per the issue's own writeup, entangled with
#4183's larger named-session/pool duality question ("there is
currently no pack-side edit that satisfies both this lint rule and
single-identity runtime behavior... filed separately: #4183"). Left
out of scope here rather than guessing at a fix ahead of that design
question landing.

### Validation

[...]

Issue #4184, opened 2026-07-12
gc lint 'named_session targets pool-controlled agent' prescribes a fix that doesn't work, and flags the documented max=0 disable form


remuscazacu:
## Problem 1: the prescribed fix doesn't produce the implied result

`gc lint` (cmd/gc/cmd_lint.go) emits:

> named_session "X" targets pool-controlled agent "X"; **remove pool settings from named-session templates** or remove [[named_session]] for pool agents

Following the first prescription — deleting `min/max_active_sessions` from the agent — makes lint go clean, but the agent then resolves to the **implicit pool `{min:0, max:-1}`**, which is unbounded and demand-reactive. Verified behaviorally (2026-07-12 drive): with pool sizing removed, formula-routed steps still spawned pool-slot sessions alongside the awake named session on every tier exercised. Lint-clean ≠ pool-gone; the warning's premise ("this removal ends the duality") is false. Since named sessions cannot serve routed work (filed separately: #4183), there is currently **no** pack-side edit that satisfies both this lint rule and single-identity runtime behavior for an agent that receives routed work.

## Problem 2: lint keys off *any* explicit pool setting, including the disable form

Explicit `min_active_sessions = 0` / `max_active_sessions = 0` is upstream's documented way to "intentionally disable an agent" (internal/config/config_test.go `TestValidateAgentsPoolMaxZeroIsValid`) — it genuinely suppresses pool spawns. But lint re-emits the same "targets pool-controlled agent" warning for it, treating the disable form identically to an active pool.

## Suggested direction

- Make the lint rule aware of effective pool semantics: implicit `{0,-1}` **is** a pool (should arguably still warn after bare removal), while explicit `{0,0}` is not (should not warn).
- Reword the prescription so it doesn't promise an outcome the runtime doesn't deliver, at least until #4183 lands.

Observed on a fork build based on `faa1d5bd1`; lint rule and cited code verified present on current upstream `main` (`5774497b0`).

remuscazacu:
Downstream confirmation of both halves of this.

[...]

PR #4995, opened 2026-08-04
fix(cmd/gc): lint no longer flags the max=1 named-session singleton flavor as a pool conflict (#4184)

brandonmartin:
## Problem

`gc lint`'s named-session pool-conflict rule flags the **documented named-session singleton shape** — `max_active_sessions = 1` with no `min_active_sessions`, no `scale_check`, no namepool — whenever a `[[named_session]]` targets it:

```
pack.toml: named_session "X" targets pool-controlled agent "X"; remove pool settings from named-session templates or remove [[named_session]] for pool agents
```

This is the sibling of the `min=0+max=0` case #4216 fixed: `agentHasPoolControls` still treats **any** non-nil `MaxActiveSessions` as pool evidence.

But `internal/config`'s `SupportsInstanceExpansion` already documents and encodes the discrimination the linter is missing — `max_active_sessions=1` has two flavors:

- **Pool agents** (`MinActiveSessions` or `ScaleCheck` set) keep pool semantics;
- **Named-session agents** (max=1 with a `[[named_session]]` entry, no Min/ScaleCheck) collapse to the canonical bare identity, because "a phantom '-1' suffix breaks tools that resolve by qualified name."

So the flagged shape is a supported, intended configuration. The rule's own regression test (`TestLintRejectsNamedSessionBackedByPoolControlledAgent`) builds `min=0/max=3` — a genuine multi-session pool; max=1 was never the target.

Concretely, before this change `gc lint` is **red on the canonical gastown pack itself** — 5 findings (mayor, deacon, boot, witness, refinery), all this false positive.

## Fix

A second exemption in `agentHasPoolControls`, following the #4216 helper pattern: `agentIsNamedSessionSingleton` mirrors `SupportsInstanceExpansion`'s named-session flavor (max=1, Min nil, no ScaleCheck, no namepool). Pool flavors of max=1 stay flagged.

## Tests

- `TestLintAllowsNamedSessionSingletonAgent` — bare `max_active_sessions = 1` + `[[named_session]]` lints clean.
- `TestLintStillRejectsSingletonPoolWithMin` — pins the boundary: `min=1/max=1` (explicit Min ⇒ pool flavor) remains a conflict.
- Existing `TestLintRejectsNamedSessionBackedByPoolControlledAgent` (min=0/max=3) and `TestLintAllowsNamedSessionOnExplicitlyDisabledPoolAgent` (#4216) unchanged and passing.

Live verification against real packs: two downstream packs with named-session singletons went from exit 1 to clean, and the canonical gastown pack's 5 findings went to 0, with no change to packs that were already clean.

Refs #4184 (problem 2's sibling; that issue also covers the misleading prescription text, which this PR does not touch) and #4216.

https://claude.ai/code/session_01E2CiS9uoTPRuDWYDU2hfZE

### DIFF
--- a/cmd/gc/cmd_lint.go
+++ b/cmd/gc/cmd_lint.go
@@ -224,11 +224,30 @@ func agentHasPoolControls(agentCfg config.Agent) bool {
 	if agentPoolExplicitlyDisabled(agentCfg) {
 		return false
 	}
+	if agentIsNamedSessionSingleton(agentCfg) {
+		return false
+	}
 	return agentCfg.MinActiveSessions != nil ||
 		agentCfg.MaxActiveSessions != nil ||
 		strings.TrimSpace(agentCfg.ScaleCheck) != "" ||
 		strings.TrimSpace(agentCfg.Namepool) != "" ||
 		len(agentCfg.NamepoolNames) > 0
+}
+
+// agentIsNamedSessionSingleton reports whether agentCfg is the named-session
+// flavor of max_active_sessions=1 documented by
+// (*config.Agent).SupportsInstanceExpansion: max=1 with no
+// min_active_sessions, no scale_check, and no namepool is a singleton with a
+// stable canonical identity, not a pool — a [[named_session]] targeting it is
+// the supported configuration, not a conflict. Pool flavors of max=1 (an
+// explicit MinActiveSessions or a ScaleCheck) keep pool semantics and stay
+// flagged.
+func agentIsNamedSessionSingleton(agentCfg config.Agent) bool {
+	return agentCfg.MaxActiveSessions != nil && *agentCfg.MaxActiveSessions == 1 &&
+		agentCfg.MinActiveSessions == nil &&
+		strings.TrimSpace(agentCfg.ScaleCheck) == "" &&
+		strings.TrimSpace(agentCfg.Namepool) == "" &&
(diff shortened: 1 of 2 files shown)


quad341 (approved):
@brandonmartin — this is the kind of fix I wish more bug reports looked like. You didn't just silence the finding: you went and found that `internal/config` already encodes the exact discrimination the linter was missing, and made lint agree with `SupportsInstanceExpansion` instead of inventing a third definition of "pool." I traced it in the checkout and the mirror is field-for-field, namepool arms and all — I even chased the one asymmetry I could see (`EffectiveMaxActiveSessions()` vs the raw field) and it's a straight passthrough, so there's no drift there either.

`TestLintStillRejectsSingletonPoolWithMin` is the test that makes this mergeable on sight — it pins the boundary so nobody widens the exemption later without tripping something. Keeping the misleading prescription text out of scope was the right instinct too.

Nice bonus: the doctor's own `NamedAlwaysMinConflictCheck` already keys the same footgun off `min>0`, so lint was the odd one out. This lines them up.

Merging as-is. One for later if you're up for it: `min=0` + `max=1` + `[[named_session]]` is still flagged for the same family of reason. Separate PR, no rush.

Thanks for unbreaking `gc lint` on the canonical pack.


---

PR #3919, opened 2026-07-04
fix(hooks): stop build_desired_state staging from double-writing reconciler-owned codex hooks

wbern:
## Problem

The `build_desired_state` home-dir reconcile tick has **two writers** of a
Codex agent's `.codex/hooks.json`, and they disagree — leaving a permanent
hybrid document that `gc doctor` flags as `codex-hooks-drift` ("needs upgrade")
forever, never converging even across `gc stop/start` bounces or repeated
`--fix`.

Both writers run in `prepareTemplateResolution`
(`cmd/gc/build_desired_state.go`) against the **same** home `workDir`:

1. **Overlay staging** — `materializeProviderOverlaysBeforeFingerprint` →
   `runtime.StageProviderOverlayDir` → `internal/overlay` merge writes the
   overlay's SessionStart entry with **`matcher:""`** (unbound `gc prime`).
2. **Reconciler** — `hooks.InstallWithResolver` writes the SessionStart entry
   with **`matcher:"startup"`** (bound `gc --city '<root>' prime`, per #3866).

The overlay merge keys hook-entry identity on the `matcher` value (the dedupe
introduced in #3808), so the bound and unbound entries are treated as distinct
and **both survive**. `hooks.Install` converges the document to the single
bound entry, but the very next staging tick re-merges the unbound `matcher:""`
entry back in — so the on-disk document a fresh session-start / `gc doctor`
reads is perpetually `[startup, ""]`.

This is a regression surfaced by #3866 (which introduced the bound matcher) in
combination with #3808's matcher-keyed dedupe. **It reproduces on a clean
`main`** — see the TDD proof below.

## Fix

Make the `build_desired_state` home-dir staging path **skip reconciler-owned
mergeable files** (`overlay.IsMergeablePath` — `.codex/hooks.json`,
`.claude/settings.json`, …) so `hooks.Install` is the **sole writer** of those
files in the home dir. The two writers can no longer disagree because there is
only one.

- `internal/overlay`: factor the existing per-provider skip into
  `isPerProviderPath`; thread an optional `SkipFunc` through `copyDir`; add
  `CopyDirForProvidersWithSkip`.
- `internal/runtime`: add `StageProviderOverlayDirSkippingMergeable` wrapping
  the shared staging with an `IsMergeablePath` skip.
- `cmd/gc`: `materializeProviderOverlaysBeforeFingerprint` uses the skip
  variant (6-line call-site swap).

### No-regression boundary (important)

[...]

PR #4996, opened 2026-08-04
fix(hooks): preserve reconciled Codex hooks across session starts

masonjames:
## Summary

Follow-up to #3919: preserve the reconciler's canonical Codex hook document across configured-home session starts, resumes, and remote runtime staging.

The remaining drift path had two parts:

1. provider overlays follow the resolved launch family, while ownership selection still depended on an explicit `install_agent_hooks` entry and session-origin metadata; a provider alias based on `builtin:codex` could therefore replay the legacy overlay into a reconciled home
2. the controller, CLI fallback, and supervisor API paths did not all project the same ownership contract

This change:

- composes the embedded Codex hook core, configured overlays, and live custom hooks into one fixed-point document
- preserves custom commands sharing a managed matcher, file mode, and lossless JSON numeric metadata
- centralizes configured-home ownership selection for the controller, CLI fallback, and API worker factory
- derives Codex ownership from the effective provider family (or an explicit hook request), including every configured named-session home backed by that agent
- carries exact reconciler-owned paths only to runtimes that declare and enforce `staging.reconciler-owned-mergeable-paths`
- makes maintained local, exec, Screen, and Kubernetes staging skip only a verified owned hook while continuing to stage ordinary siblings
- binds remote handoff to the exact canonical bytes, including a final Kubernetes check after `gc init`
- keeps fingerprint v5 compatible when ownership is absent while treating adoption or removal as real provisioning drift
- strengthens RPP conformance cleanup and positive-probe acknowledgement so a process that dies after its first liveness poll cannot pass

## Safety boundary

Configured filesystem homes are eligible for ownership regardless of how the current session was created. Manual and isolated/task workdirs remain on legacy staging, and runtimes without the versioned ownership capability cannot receive the contract.

An eligible home with a missing, malformed, non-regular, or non-converged hook file receives an empty expected digest, so runtime preflight fails closed instead of silently returning to dual-writer behavior.

The path lock serializes cooperating Gas City writers. It is not a portable compare-and-swap guarantee against an arbitrary external editor that ignores the advisory lock; eliminating the final validation-to-rename race against such an editor would require a separate crash-recoverable exchange or journal design.

## Controlled convergence trial

I built the branch as a standalone macOS arm64 binary and exercised it against an existing local city without replacing or modifying the Homebrew installation.

The first trial exposed the provider-family and API parity gaps above. With the shared ownership classifier in place, the revised trial reached a stable fixed point:

- supervisor startup reconciled both configured Codex homes to the canonical document
- an explicit supervisor reload left bytes, modification times, and inodes unchanged
- an HTTP API submit was accepted and started a fresh configured Mayor provider session; the same hook invariants remained unchanged through that resume
- a post-probe `gc doctor` reported `codex-hooks-drift` healthy (89 checks passed, with five unrelated advisory warnings)

The live trial covers controller reconciliation plus a real API-triggered provider restart. Focused worker-factory tests cover the complete `Message -> Manager.Submit -> provider restart` path and assert byte-stable staging; separate focused tests cover the CLI fallback path.

## Validation

- broad affected hook, overlay, filesystem, runtime, exec, ACP, auto, hybrid, Kubernetes, RPP-check, subprocess, tmux, worker, and full `internal/api` packages: pass
- focused `cmd/gc` configured-home, provider-alias, isolated-worktree, malformed-path, and unsupported-capability regressions: pass
- resource-census policy gate and the two new Medium-owner regression tests: pass
- `make lint-new`: pass (`0 issues`)
- `go vet` across affected packages: pass
- `make dashboard-ci`, `make check-schema`, and `make check-docs`: pass
- maintained Screen and Kubernetes session scripts: `bash -n` pass

The branch includes adversarial regressions for allocation-before-rejection cleanup, a verifier that dies after its first liveness poll, post-transport source mutation, post-`gc init` destination mutation, malformed JSON, symlink swaps, and concurrent Gas City hook writers.

One broad macOS package sweep still reports `internal/materialize.TestCanonicalizePath` because macOS resolves `/var` to `/private/var`. The same failure reproduces on a clean `main` checkout; the new materialization tests pass. The unsharded `cmd/gc` package is not included in the claim above because it exceeded the local timeout, so validation uses its focused regressions plus the repository gates listed here.

## Related work

- #3919 removed the controller home-stage double writer and is included in this branch's base
- #3992 is complementary: it handles stale `SessionStart` variant merging, while this PR governs canonical hook ownership across runtime start/resume paths

## Repository checklist

- [ ] `make check` — not claimed: the unsharded `cmd/gc` package exceeded the local timeout; focused regressions and affected-package gates are listed above
- [x] `make check-docs`
- [ ] `make test-integration` — not run as a monolithic suite; affected runtime/controller/API packages and a controlled live convergence trial are reported above
- [x] No separate issue is needed; this is a documented follow-up to #3919
- [x] Tests cover the changed behavior and failure boundaries
- [x] Runtime-provider documentation is updated
- [x] No breaking configuration change; providers without the optional capability retain legacy staging

---
- Related to #5872 — reworks the same provider-family overlay staging path that writes the project-level .codex/hooks.json, making the reconciler its sole writer so that file converges to one canonical, byte-stable document across session starts and resumes; it does not stop the file being staged and does not change hook stacking, so the double-firing described in that issue remains open, and its option A (family-based staging becoming opt-in) would now have to be reconciled with the provider-family ownership this PR introduces


quad341 (approved):
@masonjames — thank you for this. What stands out isn't the ownership classifier itself, it's that you went looking for the ways it could lie: the cross-process lock test that forks a real second process and holds the flock while a sibling saves a hook, the RPP verifier that dies *after* its first liveness poll, the post-`gc init` destination check that catches a rewrite landing between staging and entrypoint release. Those are the failure modes that normally ship undetected and resurface six weeks later as "hooks randomly disappear."

The convergence-trial writeup was genuinely useful too. Showing the *first* trial exposing the provider-family and API parity gaps, rather than only reporting the clean run, is what let me trust the second one.

On our side: the branch has drifted and now conflicts with `main`, and the red CI is not yours. Every failing shard traces to one build error in `cmd/gc/order_dispatch_test.go` — a file you never touched, and whose call sites already match the current signature on your head. I confirmed that directly: `go build ./...` and `go vet ./cmd/gc/` both come back clean on your branch, so that error can't come from your code. Pure merge skew. I'll take the rebase and re-run CI; no changes needed from you.

Two things I'm raising with our maintainers rather than asking you to change: the fail-closed path turns a malformed hooks file into a hard start failure on the default tmux runtime, and resumes landing before the first reconcile tick hit the same gate. You documented both in the safety-boundary section — they're policy calls for us to make, not gaps in your work. Your commits, your credit.


---

PR #4998, opened 2026-08-05
fix(pidutil): parse proc state after comm

boshu2:
## Summary

- parse Linux `/proc/<pid>/stat` state after the final `)` that terminates `comm`, instead of assuming the third whitespace token is state
- keep malformed stat data conservative and preserve the existing signal-zero, read-error, EPERM, and non-Linux fallback behavior
- add deterministic coverage for ordinary, spaced, embedded-parenthesis, zombie, non-zombie, and malformed stat rows

This is the focused parser follow-up requested in the maintainer review of #4953. A legal multi-word `comm` currently shifts `strings.Fields` indexes and can make `pidutil.Alive` report a zombie as alive.

## Testing

- [ ] `make check`
- [ ] `make check-docs` if docs, navigation, or links changed — not run; no docs changed
- [ ] `make test-integration` if runtime, controller, or workflow behavior changed — not run; this change is isolated to a pure parser and package tests

Additional focused checks:

- `go test -count=10 ./internal/pidutil -run '^TestProcStatState$'`
- `go test -count=1 ./internal/pidutil`
- `go vet ./internal/pidutil`
- `golangci-lint run ./internal/pidutil/...`
- repository pre-commit hook, including the fast CI-equivalent Go gates
- `git diff --check 3db4ed265db02a411b081e4581b435d80d4a5311..11da09c725b2a34ef9db851f47e174714f7c1f21`

## Checklist

- [x] Linked an issue, or explained why one is not needed — no public issue; this is a separately scoped follow-up from the approved review on #4953
- [x] Updated docs for user-facing changes — not applicable; no user-facing contract changed

### DIFF
--- a/internal/pidutil/pidutil.go
+++ b/internal/pidutil/pidutil.go
@@ -41,11 +41,37 @@ func Alive(pid int) bool {
 	if err != nil {
 		return !psReportsZombie(pid)
 	}
-	fields := strings.Fields(string(data))
-	if len(fields) >= 3 && fields[2] == "Z" {
+	state, ok := procStatState(string(data))
+	if ok && state == "Z" {
 		return false
 	}
 	return true
+}
+
+// procStatState returns field 3 (state) from a Linux /proc/<pid>/stat row.
+// Field 2 (comm) may contain spaces and parentheses, so the state is the first
+// token after comm's final closing parenthesis, not the third whitespace token.
+func procStatState(stat string) (string, bool) {
+	lparen := strings.IndexByte(stat, '(')
+	rparen := strings.LastIndexByte(stat, ')')
+	if lparen <= 0 || rparen <= lparen ||
+		!isASCIIWhitespace(stat[lparen-1]) ||
+		rparen+1 >= len(stat) || !isASCIIWhitespace(stat[rparen+1]) {
+		return "", false
+	}
+	parsedPID, err := strconv.Atoi(strings.TrimSpace(stat[:lparen]))
+	if err != nil || parsedPID <= 0 {
+		return "", false
+	}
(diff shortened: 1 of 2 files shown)


quad341 (approved):
@boshu2 — this is exactly the follow-up we asked for on #4953, and it's a clean one. Merging as-is.

What I appreciate: you didn't just shift the index. You anchored on `comm`'s final `)` *and* added the whitespace-separator guards on both sides, which is what makes `(gc (worker) name)` parse correctly instead of only fixing the easy space case. The table test earns its keep too — pinning `ok == false` across all six malformed shapes means the "when in doubt, report alive" contract is now enforced rather than incidental.

Before merging I ran the parser against every process on my box: 1078 processes, zero parse failures, and 18 where the old `fields[2]` index was returning garbage — `tmux: server` was reading as `"server)"`, `Isolated Web Co` as `"Web"`. We lean on tmux throughout this repo, so this bug was live in our own process tree.

Nice touch that this lands on the same anchoring convention `processParentPID` and `parseProcStartTimeTicks` already use. `pidutil.Alive` was the last fixed-index stat parser in the tree — now there are none. Thanks for the careful scoping.

---

**Synthesis summary:** unanimous `auto-merge` / high across all three heads, no reviewer failures, CI substantively green (not zero-of-zero), `MERGEABLE`, based directly on `main`, no CodeQL alerts, no docs impact, no human verdict to weigh against. I marked it `fundamental` on the critical-path axis — `Alive` genuinely changes answer for spaced-`comm` zombies across 17 call sites — but that no longer holds the PR, and the modified surface is covered by tests added in this same PR. Ran the package tests on the head checkout myself: green.


---

PR #4999, opened 2026-08-05
fix(stop): restore registration on timeout

boshu2:
## Summary

- treat supervisor unregister during an explicitly capped `gc stop` as pending until the stop coordinator accepts success
- on timeout or later failure, synchronously restore the exact original path and effective registration name before returning, or report the restoration failure
- fence a timed-out worker so it cannot remove the restored entry or emit a late unregister/stop success record
- preserve permanent stale-entry cleanup when the city directory is already missing, and reconcile the timeout contract in the session requirements ledger

This is the focused compensating-transaction follow-up requested in the maintainer review of #4956. `reg.Unregister` currently commits before supervisor waits finish, so an explicit wall-clock cap can return inside that window and leave a live city unregistered.

Synchronous compensation can extend return past the nominal cutoff while registry/file-lock I/O completes. That is intentional: the timeout still returns nonzero, but it no longer reports back while the original registration is known to be absent without also surfacing a restoration failure.

## Testing

- [ ] `make check`
- [x] `make check-docs` if docs, navigation, or links changed

Additional focused checks:

- `go test -count=10 ./cmd/gc -run 'Test(CmdStopWallClockTimeoutBoundsSupervisorManagedInvalidConfigStop|SupervisorUnregisterTransaction.*)$'`
- `go test -race -count=10 ./cmd/gc -run 'Test(CmdStopWallClockTimeoutBoundsSupervisorManagedInvalidConfigStop|SupervisorUnregisterTransaction.*)$'`
- focused unregister, managed-invalid-config, and direct-unregister compatibility tests
- `go vet ./cmd/gc`
- `make lint-affected`
- `make fmt-check-changed`
- repository resource-census ledger check
- repository pre-commit doc-sync check
- `git diff --check 3db4ed265db02a411b081e4581b435d80d4a5311..6a20ff4e06528d4dd45e1751aab2175fd22a66bf`

The repository's unsupported monolithic `go test ./cmd/gc` exceeded ten minutes in unrelated remote-pack/init coverage; the focused and race suites above passed. The canonical sharded broad gate is tracked separately in the checklist.

## Checklist

- [x] Linked an issue, or explained why one is not needed — no public issue; the local bead is the durable tracker for this separately scoped #4956 review follow-up
- [x] Updated docs for user-facing changes — `SESSION-RECON-013` now states the timeout/compensation boundary and missing-directory exception
- [x] Called out breaking changes or migration notes — no migration; synchronous compensation may extend a timeout return while registry I/O completes


boshu2:
RPI modernization complete at a4091798a, rebuilt from current upstream/main 16bfb855b rather than mechanically resolving the old conflict.

Fresh independent validation:
- focused timeout, rollback, stale-directory, managed-provider failure, and JSON cases: pass
- race run, 3 repetitions: pass
- go vet ./cmd/gc, affected lint, changed-file formatting, resource census, docs sync, and pre-commit hook: pass

The canonical broad fast runner was also executed. It exited nonzero only in the known current-main macOS/fan-out baseline areas: pathdurability, pidutil, Herdr XDG paths, tmux wrapped-pane detection, zcode path normalization, Dolt recovery, detached-probe cleanup, and drift restart. No stop/unregister acceptance test failed.

The branch remains a focused five-file command-scoped compensating transaction. Missing City directories remain the permanent stale-registration exception.

quad341 (approved):
@boshu2 — this is a genuinely nice piece of work. Deferring the registry commit into a transaction whose mutex spans `Unregister` *and* the stale-directory `Stat` is exactly the right shape: a timeout either fences the removal before it happens or waits for it and restores the exact captured path and effective name, instead of racing the check. The state machine covers its cases, rollback is idempotent, and keeping the missing-city cleanup permanent inside that same lock is the detail that's easiest to get wrong.

What I appreciated most is that you tested the failure of the compensation, not just the compensation — `TestSupervisorUnregisterTransactionTimeoutReportsRestorationFailure` asserting the "restore failed" line, and the `registerCalls != 1` checks that would actually catch a double-restore. `TestCmdStopWallClockTimeoutBoundsSupervisorManagedInvalidConfigStop` asserting the registration both at timeout return and after the late worker drains is the real proof of the fencing property.

Verified at `a409179`: build, `go vet`, `gofmt`, and the focused suites under `-race -count=3` all pass, and the `cmd/gc` shards ran green in CI. Merging as-is.

Two things I'm noting on the PR rather than asking you to change, since both are broader than this scope: the compensation calls into `Registry`'s blocking `LOCK_EX` flock, so the timeout return is technically unbounded under lock contention; and restoring the registry row doesn't undo a reload the supervisor already acted on — though that one is a pre-existing property of the other rollback paths, not something you introduced.


---

PR #5000, opened 2026-08-05
fix(session): ReleaseWorkBead re-verifies live status — stale cached enumeration must not reopen a freshly closed bead

jacobhausler:
## The defect

`releaseWorkFromClosedSessionBead` enumerates a closing session's work through the deliberately-unflagged `List{Assignee,Status}` query, which a `CachingStore` answers from its cached view. A worker that honestly closes its work bead moments before its session is torn down loses a race: the release enumeration still reads the bead as `in_progress`, and `ReleaseWorkBead` "resets" it to `open` + clears the assignee — silently destroying a completed close.

Fast-lifecycle runtimes (a single-shot local-model harness with a ~30-60s claim-to-exit cycle) lose this race essentially every time; long-lived sessions only intermittently — which made the resulting stall look mysterious: lanes claim, work completes, and the city never advances because finished work keeps re-entering the routed queue. Observed live (Dolt history): `bd: update <id>` → `closed`, then `gc: update bead <id>` → `open`/unassigned 10-12s later, on every cycle.

## The fix

Guard at the release primitive so all callers (closed-session release, retired-session unclaim, orphan fallback) are covered: re-read the bead through `HandlesFor(store).Live` before mutating —

- live status `closed` → skip the release entirely (a closed bead needs no detach);
- otherwise use the live bead for the `in_progress`→`open` decision and the run_target-fallback check, not the enumeration snapshot;
- a failed live read falls back to the snapshot (pre-fix behavior): releasing a vanished bead is harmless, while skipping a genuinely stuck one would strand it assigned to a dead session.

## Tests

Three regression tests in `cmd/gc/work_assignment_release_clobber_test.go`:
- live-closed bead → **no write emitted at all**;
- live-open bead with a stale in_progress snapshot → assignee/affinity clear only, no redundant status write;
- live-read failure → snapshot fallback preserved.

Existing recording-fake byte-identity tests pass unchanged (the fake's Get errors for never-created beads, exercising the fallback path).

### DIFF
--- a/cmd/gc/work_assignment.go
+++ b/cmd/gc/work_assignment.go
@@ -167,6 +167,25 @@ func (w workAssignment) ReleaseWorkBead(item beads.Bead, runTargetFallback strin
 	if store == nil {
 		return nil
 	}
+	// The callers' enumerations can answer from the CachingStore's stale view
+	// (the closed-session release path deliberately runs the unflagged, cached
+	// query). A work bead whose worker closed it inside the staleness window
+	// arrives here still reading in_progress, and the release below would reset
+	// an honestly COMPLETED close back to open — detaching it for re-claim and
+	// producing the claim → work → close → reopen → re-claim loop (fast-lifecycle
+	// bodies lose this race every cycle; lanes claim but never advance). Re-read
+	// the bead through the live handle before mutating: a bead that is already
+	// closed needs no release, and the in_progress→open decision must key off
+	// current status, not the enumeration snapshot's. A failed live read falls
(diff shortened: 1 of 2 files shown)


jacobhausler:
Superseded by #5136 (merged as 5a252c606), which implements a broader live-re-verification fix covering this case. Verified independently: this PR's own regression test, TestWorkAssignmentReleaseWorkBead_SkipsFreshlyClosedBead, passes unmodified against current main with zero code from this PR — main already re-verifies live status before mutating on release. Closing as superseded.

PR #5002, opened 2026-08-05
fix(execgrace): retry process-group SIGINT to close foreground-child fork race

jacobhausler:
### What breaks

`interruptProcessGroup` (`internal/execgrace/signal_unix.go`) sends
`os.Interrupt` to the target command's process group exactly once:

```go
if killErr := syscall.Kill(-pgid, syscall.SIGINT); killErr != nil {
    ...
}
return nil
```

If the command's foreground child (e.g. a readiness `sleep`, or any
synchronous subcommand a setup script runs) is still mid-`fork()` at
the instant that `kill(-pgid, SIGINT)` executes, the not-yet-grouped
child receives nothing — only the shell leader does. A POSIX shell
defers running a pending trap while it is waiting (via `wait()`) on a
foreground child, so once that child does appear moments later, the
deferred trap still can't run until the child exits naturally. Go's
`WaitDelay` then force-kills the whole tree before the trap ever
fires, so any staged rollback (e.g. restoring files a setup script
moved aside) is lost. Every observed failure landing at ~WaitDelay
rather than jittering around some threshold is this mechanism, not a
generic timing flake.

### The fix

Retry the process-group SIGINT for a bounded 300ms budget (10ms
interval) after the first delivery, instead of sending once and giving
up. Any retry issued after the real fork completes reaches the child
correctly. 300ms stays far inside a typical grace period, so a
cooperating command's cleanup is not meaningfully delayed — only the
case where the first delivery would otherwise have been missed is
affected.

Tracks whether any delivery succeeded so a later `ESRCH` (the whole
group has since exited, typically because the trap just ran) reports
`nil` rather than `os.ErrProcessDone` — `InterruptThenKill`'s caller
treats a `nil` `Cancel()` as "cancellation was delivered", and
`os/exec`'s `Cmd.watchCtx` only substitutes `ctx.Err()` for the
process's real exit status on that `nil` return. Answering an
already-successful cancellation with `ErrProcessDone` would misreport
a fast, successful rollback as an ordinary clean exit.

### Verification

On top of current `origin/main`:

```
go build ./...          # clean
go vet ./...             # clean
go test ./internal/execgrace/...   # all pass
```

Two tests were added to `internal/execgrace/signal_unix_test.go`,
demonstrating both the root cause and the fix in both directions:

- `TestRootCause_LeaderOnlySignalMissesForegroundChild` deterministically
  reproduces the underlying shell mechanism: signaling only the leader
  (bypassing the process group, behaviorally identical to "the child
  wasn't yet a group member") leaves the rollback trap deferred behind
  the foreground child.
- `TestInterruptProcessGroupRetryReachesLateJoiningMember` proves the
  retry logic end-to-end. A live fork-timing race is inherently
  unpredictable (that's exactly why the underlying bug was a rare
  flake in production), so this test controls "late group membership"
  deterministically instead: a plain helper process starts outside the
  target process group, waits a fixed 50ms, then self-joins the group
  (`setpgid` on itself — the only way to move a process into a group
  after it has exec'd; the parent can no longer do it from outside by
  that point). `interruptProcessGroup`'s first attempt fires before the
  helper joins and must miss; only a retry within the 300ms budget can
  reach it. Confirmed both directions by hand: reverting
  `signal_unix.go` to the unpatched single-`kill` version makes this
  test fail (times out waiting for the rollback marker); the patched
  version passes reliably (5/5 local runs).

Note on porting: the preserved patch this ships was written against
`internal/runtime/exec/signal_unix.go`, which no longer exists —
the execgrace refactor moved this code to
`internal/execgrace/signal_unix.go`. The retry logic and doc comments
are carried over unchanged; the test was re-authored against the new
package (see above) rather than cherry-picked, since the old package's
test helpers no longer apply and the original test's technique (a
SIGINT-ignoring shell leader) doesn't transfer cleanly to a design that
also needs the leader to survive indefinitely across a live process
group.


jacobhausler:
CI fixes pushed: lint on the new test file, census baseline recomputed via -update, and a real defect the integration failure exposed — cleanup_failed_start's docker-rm rollback child was dying to the retried process-group SIGINT; it now inherits SIG_IGN across exec. Docker session test 10/10 (was 0/10). Merge-order note: this branch and #5003 both touch the subprocess census baseline; whichever merges second should recompute via -update rather than take either diff verbatim.

jacobhausler:
Rebased onto current `main` (`2062de48a`). One note on the baseline, because it is not a mergeable diff:

#5003 has since landed on `main` and moved the all-source subprocess audit baseline to 626 calls / 180 files. This branch was cut at 607/173, so neither side's numbers were correct after the rebase. I recomputed the baselines against the rebased tree rather than taking either diff verbatim: subprocess **626/180 → 629/181** and fixed_sleep **453/164 → 457/165** (the delta is this branch's own SIGINT-retry test). `TESTING.md` was regenerated from the TOML ledger with

```
```

`go build ./...`, `go test ./internal/execgrace/...` and `go test ./internal/testpolicy/resourcecensus/` are green locally.

jacobhausler:
Rebased onto main 1e8f5ce74. Conflicts resolved in TESTING.md, internal/testpolicy/resourcecensus/census.go, test/test-resources.toml (audit-baseline counters re-derived: subprocess 629->631/181->183, fixed_sleep 457->460/165->166; TESTING.md regenerated via `-update`).

TestInterruptedTurnClosesTheMirrorEntry (the CI-reported real failure) does not reproduce after the rebase: reproduced it first on a clean origin/main worktree with the pre-rebase bash-parse-error symptom, then confirmed in a Linux container (golang:1.26-bookworm, matching CI) that the rebased branch passes it cleanly (1.4s) — the failure was tied to the branch's stale pre-rebase merge state, not this PR's own commits (this PR touches only internal/execgrace/*, scripts/gc-session-docker; no diff vs main in zcode-repl or its tests).

Tests: go build ./... clean; go test ./internal/execgrace/... ./internal/worker/adapters/zcode/... ./internal/testpolicy/resourcecensus/... ./scripts/... all ok locally (macOS) and the target zcode test verified green in a Linux container matching CI.

New head: 4bb29431f.

jacobhausler:
Rebased onto main 58a571faa. Conflicts resolved in TESTING.md, internal/testpolicy/resourcecensus/census.go, test/test-resources.toml (audit-baseline counters re-derived: subprocess 657->660/188->189, fixed_sleep 477->481/172->173; TESTING.md regenerated via `-update`). No unanswered maintainer comments outstanding.

Tests (Nomad builders lane, Linux, matching CI): `go build ./...` clean; `go test ./internal/execgrace/... ./internal/worker/adapters/zcode/... ./internal/testpolicy/resourcecensus/...` all pass. `scripts/...` has two pre-existing failure classes unrelated to this diff (git-archive-based complexity diff tests need a real HEAD/tar-friendly checkout the sandbox doesn't provide; process-group reap/watchdog tests are sandbox-signal-handling sensitive) — neither touches execgrace or zcode.

New head: 764e4c66f.

jacobhausler:
Rebased onto current main (2026-09-14). The red checks after the rebase are the bd schema-migrate CI race tracked in #5804: the failing shards die at setup (`[setup failed]` before any test of this branch runs, e.g. internal/runtime/tmux in this run), the same shape that cleared on a plain re-run for #5735 today. Nothing in this PR touches that path. A maintainer re-run should clear it; if any shard stays red after a re-run we will fix it here.

PR #5003, opened 2026-08-05
fix(test/integration): reap orphaned gc supervisor processes on macOS (ra-llu14)

jacobhausler:
### What breaks

Both orphan-reaper mechanisms `TestMain` relies on in the integration
suite — the `sweepSubprocessTestProcesses`/`readProcessSnapshot` pair,
and `dolttest.SweepStale`'s sibling — read the process table via
`/proc`. `/proc` does not exist on macOS. `readProcessSnapshot()` has
exactly one code path:

```go
func readProcessSnapshot() map[int]procSnapshot {
	entries, err := os.ReadDir("/proc")
	if err != nil {
		return nil
	}
	...
}
```

On any Mac dev box this makes `readProcessSnapshot()` return `nil`
unconditionally, which makes `sweepSubprocessTestProcesses()`'s guard
(`if len(procs) == 0 { return }`) fire every time — both the pre-run
and post-run sweeps are silent no-ops on macOS. Any `gc supervisor
run` process left behind by a killed or `-timeout`'d run (`go test
-timeout`'s internal deadline calls `os.Exit(2)` directly — it
delivers no signal, so `installIntegrationSignalSweeper`'s SIGQUIT/
SIGTERM/SIGINT handler never fires for it either) is never reaped, on
that run or any later one.

### Reproduction

Structural: `readProcessSnapshot` has no fallback branch, so the
result doesn't depend on what's actually running — but to rule out
"maybe nothing ever needs reaping in practice", a decoy process with
the exact orphan shape from the original report (`argv0` set to
`.../gc-integration-<pid>-*/bin/gc supervisor run`) was spawned and
confirmed alive via `ps(1)` while `readProcessSnapshot()` returned 0
entries and the derived kill set was empty — on unpatched `main`, the
live decoy is completely invisible to the sweep.

### The fix

Add a `ps`-based fallback (`readProcessSnapshotPS`/`parsePSLine`) used
whenever `/proc` is unavailable, and stop gating the supervisor-orphan
sweep behind `GC_SESSION=subprocess` — both session providers boot the
same shared `TestMain` supervisor, so both need it. Also corrects a
stale comment claiming `go test -timeout` raises SIGQUIT (verified: it
does not — the deadline panic calls `os.Exit(2)` directly, bypassing
any registered signal handler).

### Verification

On top of current `origin/main`:

```
go build -tags integration ./test/integration/...   # clean
go vet -tags integration ./test/integration/...     # clean
gofmt -l test/integration/integration_test.go        # no output
go test -tags integration ./test/integration/ -run \
  'TestParsePSLine|TestReadProcessSnapshot|TestSubprocessTestKillSet|TestConfigureIntegrationSupervisorCommand|TestStopIntegrationSupervisorWithTimeoutReturnsAfterDeadline|TestIntegrationDoltSQLServerKillSetMatchesOnlyRootedConfigs' \
  -v -count=1
```

All pass, including `TestReadProcessSnapshotFallsBackToPSAndFindsRealProcesses`
— the known-positive control that the new `ps(1)` fallback finds real
live processes on a `/proc`-less host (skips itself on Linux CI).


quad341 (approved):
@jacobhausler — this is a good catch, and the way you established it is the part I appreciate most. "`readProcessSnapshot` has exactly one code path, so the result doesn't depend on what's actually running" is the argument that genuinely settles it, and spawning a decoy with the exact orphan shape to rule out "maybe nothing needs reaping in practice" is belt-and-braces I wish more bug reports had. The `go test -timeout` correction is a real bonus — that stale claim would have misled the next person to touch the signal handler.

Two things on my side, no action needed from you.

CI is red for reasons unrelated to your change: `CI / required` chokes on a job that came back *cancelled*, which the gate tolerates as `skipped` but not as `cancelled`. I'll re-run it and file a follow-up on the gate.

And ungating the sweep to the tmux path is the right call — but `subprocessTestKillSet` has no owner-pid liveness guard the way `dolttest.SweepStale` right below it does, so two concurrent local runs could now reap each other's supervisors. I'll add that scoping and pull your `ps` fallback out of its Linux skip so CI actually exercises the invocation. Your commit and credit stay exactly as they are.


---

PR #5004, opened 2026-08-05
fix(tmux): getKeyBinding must query the table, not the single-key form

jacobhausler:
### What breaks

`getKeyBinding` in `internal/runtime/tmux/tmux.go` looks up an existing
tmux binding with the single-key query form:

```
tmux list-keys -T <table> <key>
```

On tmux 3.7b this form returns **zero bytes with exit code 0** for a
binding that demonstrably exists. `getKeyBinding`'s empty-output check
can't distinguish "no binding" from "this tmux version can't answer
the single-key query", so on 3.7b it always reports no binding.

`getKeyBinding` exists specifically to capture a user's pre-existing
key binding before it gets wrapped in an if-shell guard, so the
original binding can be restored in the fallback branch. When the
lookup always comes back empty, the user's real binding is silently
discarded and replaced.

### Reproduction

```
$ tmux -V
tmux 3.7b
$ tmux -L t new-session -d
$ tmux -L t bind-key -T root F1 'display-message hello'
$ tmux -L t list-keys -T root F1   # the single-key form getKeyBinding used
                                     # -> empty output, exit 0
$ tmux -L t list-keys -T root | grep F1   # the table form
bind-key  -T root F1  display-message hello
```

### The fix

List the whole table (`list-keys -T <table>`) and select the row whose
key matches, instead of querying a single key. Confirmed working on
3.7b. Parsing of the matched line and the existing if-shell/`gt `
recursion guard (to avoid re-wrapping an already-Gas-Town binding) are
unchanged.

### Verification

On top of current `origin/main` (`3db4ed265`):

```
go build ./...                                # clean
go test ./internal/runtime/tmux/...           # all pass
```

Behavioral proof on tmux 3.7b (the deciding fact):

```
$ tmux -V
tmux 3.7b
$ tmux -L gctest new-session -d
$ tmux -L gctest bind-key -T root F1 'display-message hello'
$ tmux -L gctest list-keys -T root F1        # single-key form: EMPTY (exit 0)
$ tmux -L gctest list-keys -T root | grep F1  # table form: shows the binding
bind-key  -T root F1                        display-message hello
$ tmux -L gctest kill-server
```


jacobhausler:
@quad341 — nudge on this: `getKeyBinding must query the binding table, not the single-key form`. CI green, mergeable, no conflicts. Could you take another pass when you have a moment? Glad to adjust anything.

quad341:

- **#5004** (`getKeyBinding must query the table, not the single-binding path`) is **MERGEABLE / CLEAN**. Nothing is blocked on you.

jacobhausler:
Rebased onto main 1e8f5ce74; conflicts in scripts/runtime_tmux_manifest_test.go resolved (re-derived manifest/untagged/shard counts to 390/270/120, shards {65,65,65,65,65,65} against the merged scripts/runtime-tmux-tests.manifest; the redundant re-derive commit 32cce4555 was dropped as superseded); tests: ./scripts/... ./internal/runtime/tmux/... ok (139.1s, 10.8s).

quad341 (approved):
@jacobhausler — the reproduction in your description is what made this easy to act on. I ran it against tmux 3.7c and got exactly what you described: `list-keys -T root F1` returns zero bytes with exit 0 while the table form shows the binding sitting right there. You'd already isolated it to the query form rather than the parsing, so confirming took a minute.

Two things I especially appreciated. You noticed that moving to the table form makes the old whole-output `if-shell`/`gt ` check in `isGTBinding` unsound, fixed it, and wrote `TestSelectBindingLineScopesGTDetection` to pin down why — that's the subtle half of this change and easy to miss. And your note that the gated tests in `tmux_test.go` never run is exactly right: `TestMain` unsets `TMUX`, so those four are dead in every build, which makes your unit tests the only thing actually covering this logic.

One small thing I'll take on my side: `TestSetBindings_PreserveFallbackOnRepeatedCalls` still uses the single-key form for `firstRaw`. Unreachable today, so it isn't on you — I'll clean it up as part of the merge and re-run.

Also, the silence since August 5 was our pipeline, not your PR. Sorry about that.


---

PR #5010, opened 2026-08-05
fix: default nil bead Priority to P2 in pool_desired_state.beadPriority()

jacobhausler:
## Summary

`cmd/gc/pool_desired_state.go`'s `beadPriority()` defaulted a nil `Priority`
to P0 (highest), while `internal/beads/native_dolt_store.go` round-trips the
same nil case as bd's documented mid default, P2. Confirmed live (not
dormant, see ra-vsvjlx notes): the `assignedWorkBeads` slice that feeds
`computePoolDesiredStates` carries store-read `Bead` structs straight through
with no priority normalization step anywhere between the read and
`beadPriority()`'s two call sites (pool_desired_state.go:200, 235). Any bead
created without an explicit `--priority` therefore schedules as P0 in
pool-demand ordering instead of P2 — a genuine scheduling-priority inversion
that could out-schedule an explicitly-labeled P1 bead.

## Fix

One-line change: nil `Priority` now defaults to `2` in `beadPriority()`,
matching `native_dolt_store.go`'s own default semantics.

## Test

`cmd/gc/pool_desired_state_nil_priority_test.go` (new):
- `TestBeadPriority_NilDefaultsToP2` — proven fail-before (returned 0, wanted
  2) / pass-after.
- `TestBeadPriority_ExplicitPriorityPreserved` — explicit P0/P1 beads are
  unaffected by the default.

```
go test ./cmd/gc/ -run TestBeadPriority -v
--- PASS: TestBeadPriority_NilDefaultsToP2
--- PASS: TestBeadPriority_ExplicitPriorityPreserved
```

Full `cmd/gc` package suite also run; the only failure is a pre-existing,
unrelated infra flake (a leaked `dolt sql-server` subprocess from
`TestInitFromWithoutHostedPreservesTemplate`, not any `--- FAIL: TestXxx`
assertion) — reproduces identically on the sibling
`fix/release-guard-destructive-deletes` clone, so it predates this change.

Source bead: ra-vsvjlx.


jacobhausler:
@quad341 — nudge on this: `default nil bead Priority to P2 in pool_desired_state.beadPriority()`. CI green, mergeable, no conflicts. Could you take another pass when you have a moment? Glad to adjust anything.

quad341:

- **#5010** (`default nil bead Priority to P2 in pool_desired_state`) is **MERGEABLE / CLEAN**. Nothing is blocked on you.
- This is the smallest of the five and the most obviously self-contained; it should not have been the one waiting longest.

quad341 (dismissed):
@jacobhausler — this is a genuinely good catch, and I traced the whole thing rather than taking it on faith. You're right: nil `Priority` really does round-trip as P2. `native_dolt_store.go:1999` writes it, `nativePriorityFromIssue` reads it back, and `internal/beads/query.go:420` has a `readySortPriority` helper that is character-for-character your function. The mismatch you found is real.

There's a twist one layer down. `applyNestedCaps` sorts `BeadPriority` **descending** (`pool_desired_state.go:456` — the existing fixtures name `10` "w-high"), so bigger wins there, while bd priority is ascending-urgent. Net effect: after your change an unset bead (2) sorts *above* an explicit P1 (1) — the reverse of what your comment intends, and P1 used to win. It also means P4 already beats P0 today, which is a pre-existing mess you happened to walk into.

That's ours to untangle, not yours. I'm taking it on this branch: normalize bd priority into the scheduler's rank, keep your nil⇒P2 semantics, add a capped nil-vs-P1 ordering test, drop the stray `PR-BODY.md`, and re-run `cmd/gc`. Your commits and credit stay. Thanks for the careful writeup.


---


quad341:
@jacobhausler, a correction to our 09-01 review. Its last line says this PR is ready to merge, but the fix the review describes has not been pushed to this branch yet. That fix normalizes bd priority into the scheduler's rank, so an unset bead no longer outranks an explicit P1. Merging the current head as-is would ship that inversion, so I've removed the `status/ready-to-merge` label until our fix lands here.

There's nothing for you to do; that work is still ours. Your commits and credit stay as they are. Sorry for the mixed signal.

quad341 (approved):
@jacobhausler — thank you for this one specifically, because the store analysis is the hard half and you got it exactly right. nil `Priority` really does round-trip as P2: `native_dolt_store.go:1999` writes it, `nativePriorityFromIssue` reads it back, and `internal/beads/query.go:420` has a `readySortPriority` helper that is character-for-character the function you wrote. I traced all of it rather than taking the write-up on faith, and it holds.

There's a twist one layer down that's ours, not yours. `applyNestedCaps` sorts `BeadPriority` **descending** (the old fixtures named `10` "w-high"), while bd priority is ascending-urgent. So with your change alone, an unset bead (2) would land *above* an explicit P1 (1), the opposite of what your comment intends. It also meant P4 already beat P0 before your change: a pre-existing mess you happened to walk straight into.

I've taken that on this branch (e8c5630):

- bd priority is now normalized into the scheduler's rank (`beadPriorityRank`) at all three places a `SessionRequest` gets one. Main grew a third since the review, the asleep-resume path.
- Your nil ⇒ P2 semantics and both of your tests are kept exactly as you wrote them.
- A new capped test pins the invariant your comment claims: with one seat, an explicit P1 beats an unset bead.
- Two fixtures that encoded the old "bigger number wins" order are corrected.
- The stray `PR-BODY.md` is gone.

The full `cmd/gc` suite passes. Your commits and credit stay, and it merges when CI is green. Sorry for the mixed signals on timing; the delay was ours too.

---
The automated reviewers split on this one (qwen: auto-merge; claude and codex: fix-merge), so a maintainer read them and chose **fix-merge**, from the claude review.

PR #5012, opened 2026-08-05
fix: NudgeSession must not discard the confirmed bool from submitEnterAndConfirm

jacobhausler:
## Summary

`internal/runtime/tmux/tmux.go`'s `NudgeSession` discarded the `confirmed`
bool from `submitEnterAndConfirm` (`if _, err := submitEnterAndConfirm(...);
err != nil {...}`) and always reported clean delivery (`nil`) whenever the
Enter send itself didn't error — even when the busy-confirm loop burned its
full budget and never observed the agent go busy, i.e. the message may still
be sitting drafted-but-unsubmitted in the pane. This is ra-3x46cy finding 1
(PROVEN by code read): the queue-ack path (`tryDeliverQueuedNudgesByPoller`)
and the idle-claim backstop's attempt counter both treat a nil error as
"delivered," so an unconfirmed submit was silently swallowed instead of
retried — the root cause of the 15-minute nudge stall the bead observed live.

## Fix

`NudgeSession` now captures `confirmed` and, when false, returns a typed
sentinel error (`ErrNudgeSubmitUnconfirmed`, wrapped with the session name)
instead of nil. Callers already propagate `NudgeSession`'s error verbatim up
through `Provider.Nudge`/`NudgeNow` (no wrapping in between), so a
retry-capable caller now correctly sees a non-nil error and does not ack the
queue item or advance its attempt counter. `delivered` (which gates the poke
timestamp used for session-activity discounting) is still set on any
error-free Enter delivery, confirmed or not — that accounting is unrelated to
this fix's scope (see ra-3x46cy finding 3).

Two pre-existing tests turned out to rely on the exact bug this patch fixes —
both nudge a `claude`-provider pane whose fake command (`cat -v`) can never
emit a busy indicator, so `confirmed` was always false and they only passed
because `NudgeSession` used to swallow that into `nil`:
- `TestNudgeSessionSkipsEscapeForClaude` (tmux_test.go)
- `TestNudgePokeRealTmux`'s "never-busy claude nudge" subtest
  (nudge_poke_integration_test.go, gated behind `GC_TMUX_INTEGRATION=1`)

Both now explicitly tolerate `ErrNudgeSubmitUnconfirmed` as the correct,
expected outcome for their never-busy fake panes, with a comment explaining
why.

## Test

New: `TestNudgeSessionReturnsUnconfirmedErrorWhenNeverBusyForClaude`
(nudge_submit_confirm_integration_test.go) — a fake `claude`-family binary
that never prints a busy indicator (`GC_TEST_BUSY_AFTER=100`, far beyond the
confirm budget). Proven fail-before (`NudgeSession` returned nil) / pass-after
(`errors.Is(err, ErrNudgeSubmitUnconfirmed)`).

```
go test -tags integration ./internal/runtime/tmux/... -run 'TestNudgeSession|TestSubmitEnterAndConfirm' -v
... (all PASS, 20 tests)

go test ./internal/runtime/tmux/...
ok  	github.com/gastownhall/gascity/internal/runtime/tmux

GC_TMUX_INTEGRATION=1 go test -tags integration ./internal/runtime/tmux/... -run TestNudgePokeRealTmux -v
--- PASS: TestNudgePokeRealTmux (all 5 subtests)
```

Full integration suite (`GC_TMUX_INTEGRATION=1 go test -tags integration
./internal/runtime/tmux/...`) also run; two failures are pre-existing,
unrelated environment flakes independent of this change:
`TestNudgeSessionConfirmsSubmitForClaude` (passes reliably in isolation —
reruns clean 3/3 — flakes only under the full batch's concurrent tmux
sessions) and `TestGetKeyBinding_CapturesDefaultBinding{,WithArgs}` (depends
on this machine's default tmux key-binding config, unrelated to nudging).

Source bead: ra-3x46cy (finding 1).


jacobhausler:
Rebased onto current main and fixed the mechanical CI failures (registered the new test in the runtime-tmux manifest, rebanked the fixed_sleep census baseline 431→432, regenerated TESTING.md).

One failure is expected to remain: `TestPhase2WorkerCoreRealTransportProof/claude/tmux-cli`. This PR stops `NudgeSession` from silently swallowing an unconfirmed submit as a clean delivery — and on fork-PR Actions runs (no repository secrets, so no live `claude` credential) the pane genuinely never goes busy, which the fix now correctly surfaces as `ErrNudgeSubmitUnconfirmed` instead of a false pass. The same assertion passes locally against a real tmux + fake claude binary (`go test -tags integration ./internal/runtime/tmux -run TestNudgeSession`). A maintainer-context run with secrets should go green; alternatively that job may warrant a fork-PR guard. Happy to adjust if you'd prefer a different approach.

quad341 (approved):
@jacobhausler — this is a genuinely sharp catch. `submitEnterAndConfirm` documents `(false, nil)` as "Enter reached tmux, busy never observed," and `NudgeSession` was dropping that one frame up, so the queue dispatcher acked nudges that may still have been sitting drafted in the pane. Your never-busy fake-claude test with `GC_TEST_BUSY_AFTER=100` *proves* it rather than asserting it, and I especially appreciate you calling out that two existing tests were passing *because* of the bug and loosening them with a comment instead of quietly deleting them — that's the part people skip. The manifest/census/TESTING.md rebanking is a chore and you did all of it.

I'll take the rest on my side. One caller you couldn't easily see: `doStartSession` treats any startup-nudge error as fatal, so the sentinel now fails session start for claude panes that don't go busy in time. I'll make that path tolerate `ErrNudgeSubmitUnconfirmed`, add a subtest for it, and drop the stray `PR-BODY.md`.

That's also the real cause of both red jobs — I pulled the log, and it's `TestPhase2WorkerCoreRealTransportProof/claude/tmux-cli` failing on your own error string. It uses a fake `sh` provider with `GC_PROVIDER=claude`, so no credential is involved and secrets wouldn't have helped. One small correction while I'm in there: the queue *doesn't* ack on this error (the important part), but it does still consume a bounded attempt via `failedQueuedNudge` — I'll fix the comment to match.

Merging with your commits and credit intact.


---

PR #1216, opened 2026-04-25
feat(doctor): add rig-pack-coverage check for orphaned rig-scoped named_sessions

boylec:
## Summary

When a city-level pack declares rig-scoped `named_sessions` with `mode=always`, but no non-suspended rig imports that pack, the declared sessions never materialize. There is currently no doctor check that surfaces this gap; deployments silently miss the rig-scoped agents the pack intended to provide.

This PR adds a new generic `rig-pack-coverage` doctor check that detects packs whose rig-scoped `named_sessions` have no importing rig. The detection is purely structural — pack declares rig-scoped sessions; no rig in the city imports the pack — with no role names hardcoded. This satisfies the Zero Framework Cognition principle from CONTRIBUTING.md: the check observes shape, doesn't reason about specific roles.

A bundled change adds a root `pack.toml` to `examples/gastown/` with a `[defaults.rig.imports.gastown]` section, so new rigs created from the example pack inherit the rig-scoped agents automatically — closing the gap that motivated the check for the canonical example.

Closes #1215

## Tests

Bundled at `internal/doctor/checks_rig_coverage_test.go` (292 lines). Cover:
- Rig imports pack — no warning
- No rig imports pack — warning emitted
- Pack has no rig-scoped sessions — no warning
- All importing rigs suspended — warning emitted

## Testing

- [x] `go test ./internal/doctor/...` — green
- [x] `make check` — verified locally (note: must currently be stacked on top of #1208 to build on darwin; Linux CI is unaffected)
- [x] `make check-docs` — N/A (docs unchanged)
- [x] `make test-integration` — doctor checks exercised

## Checklist

- [x] Linked an issue (#1215)
- [x] Tests cover the new check (and pre-existing checks unaffected)
- [x] Updates the gastown example pack with a default rig imports stanza so new rigs from the example get rig-scoped agents
- [x] No breaking change — only adds a new check; existing doctor behavior unchanged

PR #5013, opened 2026-08-05
fix: NudgeSession clears pending pane input before pasting

jacobhausler:
## Summary

`internal/runtime/tmux/tmux.go`'s `NudgeSession` never cleared pending pane
input before pasting a new message — unlike `SendKeysReplace`, which sends
`C-u` first specifically for this reason. This is ra-3x46cy finding 2 (strong
hypothesis, with the code's own doc comment naming it as a known-unaddressed
upstream defect: `sendKeysLiteralWithRetry`'s comment cites upstream #1216
"Nudge delivery reliability (input collision — NOT addressed here)"). If an
earlier nudge left undelivered text sitting in the input box (the exact
failure mode of finding 1 — a lost submit Enter), a later, unrelated nudge
attempt against the same session would paste more text on top of it instead
of replacing it, producing one unsubmittable multi-fragment draft. The live
specimen showed three independent nudge subsystems (an idle-claim backstop,
two queued alias-nudges) hitting the same single-pane session inside one
incident window — genuine input-collision exposure.

## Fix

`NudgeSession` now sends `C-u` (with a short settle delay) immediately before
the literal paste, mirroring `SendKeysReplace`'s clear-before-paste pattern.
This is option (a) from the bead's fix shape — the cheapest, most surgical of
the two options offered, and upstream-worthy per the code's own citation of
#1216.

## Test

New: `TestNudgeSessionClearsPendingInputBeforePaste`
(nudge_clear_before_paste_integration_test.go). Types a leftover,
unsubmitted draft into a real tmux pane (`cat -v`, which echoes back exactly
what was submitted once Enter lands), then calls `NudgeSession` with a fresh
message and asserts the echoed line is the fresh message alone — not the
leftover draft concatenated with it. Proven fail-before (echoed
`leftover-draftfresh-message`) / pass-after (echoed `fresh-message`).

```
go test -tags integration ./internal/runtime/tmux/... -run TestNudgeSessionClearsPendingInputBeforePaste -v
--- PASS: TestNudgeSessionClearsPendingInputBeforePaste
```

Full non-integration package suite, the broader nudge/send-keys integration
set, and the `GC_TMUX_INTEGRATION=1` poke-timing suite all still pass with no
new failures:

```
go test ./internal/runtime/tmux/...
ok  	github.com/gastownhall/gascity/internal/runtime/tmux

go test -tags integration ./internal/runtime/tmux/... -run 'TestNudgeSession|TestSendKeys|TestSubmitEnterAndConfirm' -v
... (all PASS, 24 tests)

GC_TMUX_INTEGRATION=1 go test -tags integration ./internal/runtime/tmux/... -run TestNudgePokeRealTmux -v
--- PASS: TestNudgePokeRealTmux (all 5 subtests)
```

Note: this clone is independent of the sibling
`fix/nudge-confirm-not-discarded` branch (finding 1) — both branch from the
same `main` and land as separate small PRs per the rung's instructions, so
this diff does not include finding 1's `ErrNudgeSubmitUnconfirmed` change.

Source bead: ra-3x46cy (finding 2).


quad341 (approved):
@jacobhausler — this is a great one. You tracked finding 2 back to the code's own `sendKeysLiteralWithRetry` comment citing upstream #1216 as a known-unaddressed defect, then fixed it by mirroring the `SendKeysReplace` pattern that was sitting right there. I checked that mirroring claim against the source rather than taking it from the description, and it holds exactly, settle delay and all.

I especially appreciated two things. Keeping this independent of `fix/nudge-confirm-not-discarded` instead of bundling them made this reviewable in one sitting. And the bookkeeping — the fixed_sleep audit baseline in all three places, the manifest entry in source-order position, *and* the rebalanced shard counts — is the part that usually gets missed, and your deltas were right.

Two things I'll take on my side, nothing needed from you: the branch has drifted into conflict on those same counter files, so I'll rebase and recompute, and I'll drop the `PR-BODY.md` that rode along at the repo root. Your commit and credit stay as-is.

One note for later, not a change request: the `cat -v` harness exercises the tty line discipline rather than a raw-mode composer, so it proves the clear is sent before the paste but not that Claude's TUI honors `C-u`. Worth a real-provider check. `NudgePane` has the same exposure, if you want the follow-up.


---

Issue #4421, opened 2026-07-18
feat: expire/skip pending nudge at delivery when target bead is already closed


jacobhausler:
### Summary

When a \"go work X\" or \"X is routed to you and unclaimed\" nudge is enqueued, the window
between enqueue and agent wake is exactly when target beads close — on someone else's work cycle.
Today the delivery path does not check whether X is still open, so agents wake to a closed bead
and spend a full session start on a guaranteed no-op. The fix belongs in the delivery path:
before waking an agent, check the target bead's status and drop the nudge if it is already closed.

### Evidence

All evidence is already measured — not re-gathering in this report.

**Concrete undelivered instance:**  
Nudge `nudge-7d2dd611a3d5`, agent `agent-a`, session `ra-wisp-0ggn90`, source=session,  
created `2026-07-17T22:33:53Z`, expires `2026-07-18T22:33:53Z`,  
text `"dispatcher: ra-s00 ... is routed to you and unclaimed — run gc hook --claim and begin"`,  
while `ra-s00` is CLOSED (research delivered). `last_attempt` zero — never delivered.

**Delivered instance (confirms wakeup cost is real):**  
Bead `ra-0x86` closed at `2026-07-18T22:53:14Z`; a nudge for it was enqueued and delivered at
`~2026-07-18T22:55:25Z` — **~2 min 11 s after the bead closed**. The nudge reached the agent.
This is not a stale queue entry that failed to expire; it was enqueued after closure and delivered
anyway.

**Rate (the strongest part of the case):**  
In a single inspection, 3/3 nudges delivered in one batch pointed at work that was:
- done (`ra-0x86`, closed 9 min before delivery),
- blocked on a collision (`ra-w7kr`, another session holding the prerequisite IN_PROGRESS),
- or waiting on a human card (`ra-3pm8f`, not a worker step at all).

This is not a rare race — it is the steady state in a multi-agent city where work closes on other
agents' cadences.

**Pending queue growth with stale entries:**  
The pending queue was observed to grow from 8 to 16 entries over one session. One of those 16
(`nudge-7047477f3e5a`) referenced six beads, every one of which was closed (ra-pf08.4,
ra-pf08.5, ra-pf08.6, ra-pf08.7, ra-s00, ra-1b3r).

**No CLI lever:**  
`gc nudge --help` exposes exactly one subcommand: `status`. There is no `expire`, `prune`, or
`delete`. The expiry must live in the daemon's delivery path; there is nothing operators can call
[...]

PR #5016, opened 2026-08-05
fix(nudge): run queue TTL/max-attempts maintenance sweep unconditionally each dispatch tick

jacobhausler:
## Summary

Runs the nudge queue's TTL/max-attempts maintenance sweep unconditionally on
every supervisor-side dispatch tick, instead of only as a side effect of a
claim/list call scoped to one matched agent.

## Problem

`dispatchAllQueuedNudges` (`cmd/gc/nudge_dispatcher.go`) iterates every
currently-open session and, for each one, checks whether any pending queue
item's agent key matches it. Only on a match does it ever reach
`claimDueQueuedNudgesForTarget`, which is the *only* code path (shared with
`listQueuedNudges` / `listQueuedNudgesForTarget`) that runs the queue's
recover/TTL-expiry/dead-letter-retention maintenance pass
(`recoverExpiredInFlightNudges` + `pruneExpiredQueuedNudges` +
`pruneDeadQueuedNudges`).

A queued item whose target agent has **no currently-open session** can never
satisfy that match — the `OpenInfo` that would trigger it is gone, often
permanently (pool/session-generation churn). Such an item then never reaches
any maintenance-pass call site and sits in `Pending` forever, past its
`ExpiresAt`, invisible to every existing instrument.

Live evidence (ra-oudpha finding-3, mechanic J, 2026-08-05): 61 pending
queue entries with `last_attempt_at`/`claimed_at` still zero after 9+ hours
of continuously-firing dispatch ticks. ~53 of them targeted agents with no
open session at all — permanently unmatchable, and (per the sampled
`created_at` values) already past their 24h TTL.

## Fix

`dispatchAllQueuedNudges` now runs a new `runNudgeQueueMaintenanceSweep`
helper unconditionally, once per tick, right after confirming supervisor
mode — independent of whether the tick's per-session loop matches anything.
The helper is the same recover/prune sequence already used by
`claimDueQueuedNudgesMatching` / `listQueuedNudges` /
`listQueuedNudgesForTarget`, factored out so the dispatch tick can invoke it
directly under its own `withNudgeQueueState` transaction.

This is upstream-worthy per #4421's existing TTL/max-attempts semantics — it
closes a general queue-hygiene gap (any maintenance-call-site-shaped queue
has this bug if it gains a "some scan never touches every item" path), not
something fork-specific.

## Testing

New test `TestDispatchAllQueuedNudgesSweepsExpiredOrphanEvenWithoutMatch`
(`cmd/gc/nudge_dispatcher_test.go`): enqueues an item for `ghost-agent`
(created 25h ago, so its 24h TTL has already elapsed) with an empty session
snapshot (no open session will ever match it), runs one
`dispatchAllQueuedNudges` tick, then reads the *raw* persisted state via
`nudgequeue.LoadState` (not `listQueuedNudges`, which would run its own
maintenance sweep as a side effect and mask the bug) to assert the item
moved from `Pending` to `Dead` with `LastError == "expired"`.

- Fail-before proven: reverted to pre-fix `dispatchAllQueuedNudges`, the new
  test fails with `pending = 1, want 0`.
- Pass-after: `go test ./cmd/gc/... -run TestDispatchAllQueuedNudgesSweepsExpiredOrphanEvenWithoutMatch -v` — PASS.
- `go test ./cmd/gc/... -run 'TestDispatchAllQueuedNudges|TestClaimDueQueuedNudges|TestListQueuedNudges|TestNudge'` — all PASS (no regressions in the surrounding nudge suite).
- Full `go test ./cmd/gc/...`: only 2 failures, `TestResolveImportRootFallsBackToStandalonePackDir` and `TestResolveImportRootPrefersNearestPackUnderCity` — confirmed pre-existing macOS `/var` vs `/private/var` tmpdir-symlink flakes by reproducing them identically on an unmodified clone of the same commit; unrelated to this change.

## Scope

Touches only `cmd/gc/nudge_dispatcher.go` (call the sweep) and
`cmd/gc/cmd_nudge.go` (new `runNudgeQueueMaintenanceSweep` helper, built from
existing pieces). No behavior change for matched targets; no change to
prune/expiry semantics, only to when they run.

### DIFF
--- a/cmd/gc/cmd_nudge.go
+++ b/cmd/gc/cmd_nudge.go
@@ -2453,6 +2453,31 @@ func recordNudgeDispatchSkips(cityPath string, counts map[string]int64) error {
 	})
 }
 
+// runNudgeQueueMaintenanceSweep runs the queue's recover/TTL-expiry/dead-
+// letter-retention passes over the whole queue, independent of whether any
+// pending item currently matches an open session. Every other maintenance
+// call site (claimDueQueuedNudgesMatching, listQueuedNudges,
+// listQueuedNudgesForTarget, ...) only runs these passes as a side effect of
+// an operation scoped to a specific agent/target, so an item whose target
+// never matches never gets swept by any of them. The supervisor dispatch
+// tick is the one path that iterates the whole queue every cycle regardless
+// of match outcome, so it owns running this sweep unconditionally.
+func runNudgeQueueMaintenanceSweep(cityPath string, now time.Time) error {
+	maint := nudgeMaintenanceStore{cityPath: cityPath}
+	defer maint.close() //nolint:errcheck // best-effort
+	return withNudgeQueueState(cityPath, func(state *nudgeQueueState) error {
+		front := maint.frontForState(state)
+		deadline := noMaintenanceDeadline()
+		if err := recoverExpiredInFlightNudges(state, front, now, deadline); err != nil {
+			return err
+		}
+		if err := pruneExpiredQueuedNudges(state, front, now, deadline); err != nil {
+			return err
+		}
+		return pruneDeadQueuedNudges(state, front, now, deadline)
+	})
+}
+
 func pruneExpiredQueuedNudges(state *nudgeQueueState, front *nudgequeue.Store, now, deadline time.Time) error {
 	return pruneExpiredQueuedNudgesWithClock(state, front, now, deadline, clock.Real{})
 }
--- a/cmd/gc/nudge_dispatcher.go
+++ b/cmd/gc/nudge_dispatcher.go
@@ -124,6 +124,17 @@ func dispatchAllQueuedNudges(cityPath string, cfg *config.City, store, sessStore
 	if !nudgeDispatcherIsSupervisor(cfg) {
 		return 0, nil
 	}
+	now := time.Now()
+	// Run the queue's TTL/max-attempts maintenance sweep unconditionally,
+	// independent of whether any item below matches an open session. The
+	// per-session loop's only path to recover/prune is a successful claim in
+	// claimDueQueuedNudgesForTarget, which a structurally orphaned item
+	// (target agent has no open session, and never will again) can never
+	// reach — leaving it in Pending past its ExpiresAt forever. See
+	// ra-oudpha finding-3.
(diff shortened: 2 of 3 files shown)


jacobhausler:
Rebased onto current main (`e2fd3c46d`) to clear the conflict with #5017: both changes add a function at the same spot in `cmd_nudge.go` (union kept — `recordNudgeDispatchSkips` from main, `runNudgeQueueMaintenanceSweep` from this PR), and this PR's one new `dispatchAllQueuedNudges` call site picks up #5017's new trailing `io.Writer` argument (`nil`, matching the existing test call sites). Also dropped a stray `PR-BODY.md` this branch had accidentally committed at the repo root. Full `go test ./cmd/gc/` green locally (518s).

quad341 (approved):
Thanks @jacobhausler — this is a genuinely good catch. Noticing that `pruneExpiredQueuedNudges` was only ever reachable through agent-scoped call sites, so an item whose target has no open session could never be swept by *any* of them, is the kind of structural gap that's easy to stare straight past. I traced it in the checkout and it holds exactly as you describe: `ExpiresAt` has a single assignment site, and nothing else was ever going to reach those orphans.

Two things I especially appreciated. Reading raw `nudgequeue.LoadState` in the test instead of `listQueuedNudges` — which runs its own sweep and would have quietly masked the entire bug — and doing the fail-before revert to prove the test actually bites. That saved me real verification work.

CI is red and it isn't yours. The `packages-core` failure is `providerledger`'s waiver set hard-coded to expire 2026-08-12 00:00 UTC; the run started at 02:30 UTC, so it's failing on main and every open PR. CodeQL hit a self-signed-cert init error. I'll fix the waiver on main separately, rebase you onto it, and re-run.

Nothing needed from you — your three files land as-is.


---

julianknutsen:

This PR was OPEN/MERGEABLE but behind main, so the merge queue would not take it. Rebased (true rebase, no merge commit) onto `main` @ `b63623d08e`; new head is `d41369575f`. No conflicts. `behind_by` is now 0.

Verification on the rebased tree: go vet ./cmd/gc/ clean; cmd/gc -run TestDispatchAllQueuedNudges PASS (9/9, including TestDispatchAllQueuedNudgesSweepsExpiredOrphanEvenWithoutMatch). 

Issue #4706, opened 2026-07-27
feat(runtime): declarative per-provider nudge submit-key sequence (codex needs Esc+Enter)


chris-sanders:
### Problem to solve

On Kubernetes, a codex agent never starts its first turn.

The k8s carrier's `Nudge` delivers a prompt with `tmux send-keys -l <text> Enter`.
Claude submits on that trailing `Enter`, but the codex TUI buffers the send-keys
burst as a paste and treats `Enter` as a newline, so the prompt sits in the
composer, unsubmitted, and the agent looks alive but idle. Codex's actual submit
sequence is `Esc` then `Enter`.

The reason this can't just be patched in place is that the current tmux drivers
already hardcode per-provider key heuristics in Go, there is even a test
asserting that codex receives no `Escape`. That is provider-specific knowledge
living in the transport layer, which is exactly what the project's "keep
judgment out of Go" rule targets.

### Proposed change

Make the submit sequence declarative, resolved per provider:

```toml
nudge_submit_keys = ["Escape", "Enter"]   # default: ["Enter"]
```

- Thread it `ResolvedProvider` → `StartupHints` → `runtime.Config`.
- Persist it with the session (e.g. a pod annotation) so a controller restart
  doesn't forget how to submit to an already-running agent.
- Have the shared tmux carrier execute the sequence after literal delivery.
- Treat any paste/inter-key debounce as driving-profile data alongside it.
- The codex builtin supplies `["Escape", "Enter"]`; Claude stays `["Enter"]`;
  city/pack overrides absorb per-CLI-version differences.
- Apply it to **every** nudge, not just the first.

This is the same misplaced-knowledge problem the provider-manifest RFC (#4110)
targets, so the field belongs there rather than in a competing RFC.

### Alternatives considered

- **Keep patching per-provider branches in the tmux driver.** Works, but adds
  another hardcoded provider name to Go and breaks again on the next CLI's
  keybinding, the opposite of "more useful as models improve".
- **Always send `Esc` before `Enter`.** Rejected: harmless for codex, but `Esc`
  is meaningful in other TUIs and would change their behavior.
- **Have the agent submit itself.** Not available, the problem is delivery, not
  the agent's cooperation.

### Scope and impact

- Touches the provider manifest / resolved-provider config surface, the tmux
  carrier shared by local and k8s runtimes, and session persistence.
- An existing test asserts codex receives no `Escape`; it encodes the current
  behavior and would need to change with the fix.
[...]

PR #5018, opened 2026-08-05
feat(nudge): make the tmux submit-key sequence declarative per provider family

jacobhausler:
## Summary

Makes the tmux nudge carrier's post-paste submit action a declarative,
per-provider-family key sequence (the design proposed in upstream
gastownhall/gascity#4706), instead of a single hardcoded "Enter" call site.
Zero behavior change for every provider today — this is infrastructure, not
a claude-specific fix (see "What this patch does NOT do" below).

## Problem

ra-oudpha's dispatch note: "the idle-nudge composer STILL types-without-
submitting into claude TUI sessions... This is upstream #4706's exact shape
(declarative per-provider nudge submit-key sequence)." This recurred *after*
gascity#5012 (propagate the unconfirmed-submit error instead of a false
`nil`) and #5013 (clear pending input before pasting) had already landed —
so the failure is now honestly reported (no false "delivered" acks) but
still not resolved.

#4706 itself documents the concrete, evidenced version of this problem for
codex: a k8s codex agent's first turn never started because a `send-keys -l
<text> Enter` burst gets buffered by codex's TUI as a paste, and the
trailing `Enter` is swallowed as a composer newline instead of triggering
submit — codex's actual submit sequence is `Escape` then `Enter`. The
proposed fix is to stop hardcoding per-provider key heuristics in Go and
make the submit sequence declarative per provider family instead.

## Investigation for claude specifically

I could not identify a wrong key as the cause of the claude-specific
residual, and did not implement an unverified fix for it — reporting per the
bead's "report if not obvious" instruction rather than guessing:

- #4706 itself specifies claude's default submit sequence as plain `Enter`,
  which is exactly what this fork already sends
  (`providersSkippingEscapeBeforeEnter` already includes `"claude"`, so no
  spurious Escape is synthesized before it either).
- The mechanic's own investigation (ra-3x46cy) explicitly ruled out a
  busy-indicator false negative for the specimen that motivated this bead:
  the composer text was observed **visibly still sitting unsubmitted**, not
  silently-submitted-but-unconfirmed — so this isn't `paneContainsBusyIndicator`
  missing a fast turn.
- `submitEnterAndConfirm` already retries Enter up to 3 times with busy
  polling between sends (~1.8-2.4s budget) before giving up honestly via
  `ErrNudgeSubmitUnconfirmed`.

Pinning the actual cause needs a live trace against a failing session,
which this fork-patch pass does not have (the city this bead is scoped
against is live and read-only for this pass; ra-3x46cy's own investigator
reached the identical conclusion trying to bisect the *dispatch*-side gate:
"I could not safely bisect... without adding a trace line and restarting
the supervisor — out of scope for a read-only pass").

## What this patch does

- `internal/runtime/tmux/tmux.go`: new `nudgeSubmitKeySequences map[string][]string`
  (provider family → ordered tmux key names) and
  `defaultNudgeSubmitKeySequence = []string{"Enter"}`, with a lookup
  (`nudgeSubmitKeySequenceForFamily`) and a target-resolving wrapper
  (`nudgeSubmitKeySequence`, mirroring how `submitVerifyEligible` and
  `shouldSendEscapeBeforeEnter` already resolve provider family from the
  `GC_PROVIDER` pane env var with a process-name-sniff fallback).
- New `sendNudgeSubmitSequence(target string, keys []string) error` sends
  each key via `tmux send-keys`, pausing `nudgeSubmitKeySettle` (100ms)
  between keys in a multi-key sequence.
- `NudgeSession` and `NudgePane` both now resolve and send the target's
  declared sequence instead of a hardcoded literal `"Enter"` string, for
  both the confirm/retry path (`submitEnterAndConfirm`, renamed its
  `sendEnter` param to `sendSubmit` — a rename only, same injected-callback
  shape) and the historical best-effort fallback path.
- **`nudgeSubmitKeySequences` starts empty.** No family (including
  `claude` and `codex`) has an explicit entry, so every provider keeps
  exactly its current single-Enter behavior. This is deliberately scoped as
  pure infrastructure: I did not add codex's `["Escape", "Enter"]` entry
  from #4706 in this patch, since validating it against codex's actual TUI
  is outside a claude-focused bead's scope and I have no live codex session
  to verify against — flagging it as a natural, low-risk follow-up once
  someone can test it.
- Once a live trace pins claude's actual requirement (whatever it turns out
  to be — a different key, a double-Enter, more settle time), landing it is
  a one-line table entry plus a test, not another pass through
  `NudgeSession`'s delivery mechanics.

## Testing

- `TestNudgeSubmitKeySequenceForFamilyDefaultsToEnter` /
  `TestNudgeSubmitKeySequenceForFamilyHonorsTableEntry`
  (`internal/runtime/tmux/nudge_submit_key_sequence_test.go`, no build tag):
  pure unit tests on the lookup table and its default fallback.
- `TestSendNudgeSubmitSequenceSendsEachKeyInOrder` /
  `TestNudgeSessionUsesDeclaredSequenceForProviderFamily`
  (`internal/runtime/tmux/nudge_submit_key_sequence_integration_test.go`,
  `//go:build integration`, gated on `hasTmux()`): live-tmux tests using a
  `cat -v` pane (which echoes control bytes as visible caret notation, e.g.
  Escape → `^[`) and a throwaway `testfam` provider family registered only
  for the test, proving both the low-level primitive and `NudgeSession`
  itself actually emit every key in a declared multi-key sequence, not just
  the last one — this is the real wiring a future claude/codex-specific fix
  would depend on, not just that the lookup function returns the right
  slice.
- Fail-before proven: temporarily made `sendNudgeSubmitSequence` send only
  the sequence's last key (simulating broken multi-key wiring) — both
  integration tests failed (`CapturePaneAll missing Escape...`). Restored →
  both pass.
- `go test ./internal/runtime/tmux/...` (no tag) — all PASS.
- `go test -tags integration ./internal/runtime/tmux/... -run 'TestNudgeSubmitKeySequence|TestSendNudgeSubmitSequence|TestNudgeSessionUsesDeclaredSequence'` — all PASS.
- Full `GC_TMUX_INTEGRATION=1 go test -tags integration ./internal/runtime/tmux/...`: only the two pre-existing, machine-config-dependent failures already documented on ra-3x46cy's earlier landing note — `TestGetKeyBinding_CapturesDefaultBinding{,WithArgs}` ("depends on this machine's default tmux key-binding config") — everything else, including `TestNudgeSessionSkipsEscapeForClaude`/`TestNudgeSessionSkipsEscapeForOpenCode` and the full nudge-submit-confirm suite, PASS with no regressions.

## Scope

`internal/runtime/tmux/tmux.go` (declarative table + two new methods + the
`sendEnter`→`sendSubmit` rename inside `submitEnterAndConfirm`, `NudgeSession`,
`NudgePane`), plus the two new test files above. No config/TOML surface
added — the table is a Go-level declarative source of truth today, matching
how the existing `providersSkippingEscapeBeforeEnter` per-provider list is
already Go-level rather than threaded through `config.City`; that's a
bigger, separate change (full #4706/#4110-style config plumbing) out of
scope for this fork patch.


jacobhausler:
Rebased onto current main and rebanked the fixed_sleep census baseline (431/160→435/161 for this PR's 4 new sleep calls); TESTING.md regenerated.

The `cmd/gc product metrics testhook` failure looks environmental rather than related to this diff: the PR touches only `internal/runtime/tmux/*` and `scripts/*`, the CI log shows a bare `[setup failed]` with no test-level failure event, and the job's exact test selection passes locally in ~3s. A re-run should clear it.

quad341 (approved):
@jacobhausler — thank you for this, and specifically for the discipline in it. You were sent after a claude submit stall, couldn't pin the cause without a live trace, and *didn't* ship a plausible-looking key change to look productive. Writing "I could not identify a wrong key — here's the infrastructure so the real fix is one line and a test" is harder than guessing, and it's the right call.

The integration tests are the part I appreciate most. Proving `NudgeSession` itself emits every key — with the `providersSkippingEscapeBeforeEnter` trick so the observed `^[` can only come from step 5, not step 3 — is the difference between testing a lookup function and testing the wiring. I traced that reasoning and it holds; all four pass here against live tmux, and I confirmed the empty table really does mean zero behavior delta for every provider today.

I'll take the integration work: dropping `PR-BODY.md`, rebasing and re-deriving the census/manifest counters (those collide with every test-adding PR), refreshing three error strings that still say "Enter", and re-running the testhook job. Nothing for you to do.

One note for whoever adds codex's `["Escape", "Enter"]`: a retry after a failed Enter re-sends the Escape. Worth pinning before that entry lands — flagging it here so it isn't lost.


---

PR #5019, opened 2026-08-05
fix(tmux): bounded retry-with-backoff for unconfirmed startup nudge submit

jacobhausler:
## Summary

A session can pass readiness checks and start successfully while its startup nudge remains drafted and unsubmitted in the booting TUI. The original Arm A/B timing experiment established this as a true negative: injection immediately after readiness (~0.3s) remained unsubmitted, while the same injection around 6s landed and was answered.

#5012 correctly made an unconfirmed submit non-fatal, preventing a breaker loop. Its warning-only guard also exposed the remaining gap: without a retry-capable caller, the session looks healthy but the agent may never begin work.

## Fix

`launchOrchestration` now routes the startup nudge through `sendStartupNudgeWithRetry`.

- Retries only `ErrNudgeSubmitUnconfirmed`; every other error still fails fast.
- Repeats the complete clear/paste/submit cycle after 2s, 4s, 6s, and 8s backoffs, so a draft cleared or redrawn during TUI boot is restored.
- A first-attempt success pays no backoff cost.
- Exhaustion remains warning-only, preserving #5012's non-fatal startup contract.
- Cancellation stops the ladder immediately.
- Retries are capped to half of the context's remaining deadline, measured with actual elapsed time so slow submit-confirm probes cannot consume the startup budget invisibly.
- The elapsed-time test uses a fake clock; no new real sleeps were added to the test inventory.

The branch is rebased onto current `main`; it no longer carries #5012 or #5018 changes. `PR-BODY.md` is not part of the diff.

## Validation

Passed locally after the rebase:

- Runtime-tmux manifest contract (370 total, 253 untagged, 117 integration-only; shards 62/62/62/62/61/61)
- All `sendStartupNudgeWithRetry` tests
- Call-site retry, warning-on-exhaustion, and #5012 non-fatal/error-path tests
- `go test ./scripts ./internal/runtime/tmux -count=1`
- `go vet ./internal/runtime/tmux`
- `lint-changed` and the repository pre-commit hook

The pre-push full fast suite passed the six `cmd/gc` shards and every PR-scoped package, but the unit-core job failed on `internal/pidutil/TestCmdline_FailsClosedWhenUnreadable`. That deterministic macOS/current-main test defect is already tracked as #5344; this PR has no `internal/pidutil` diff. After reproducing it 10/10 in isolation, the exact lease-guarded branch update was sent with `--no-verify` rather than pretending the unrelated baseline failure belonged to this change.

---
- Related to #5954 — adds a bounded retry ladder for unconfirmed startup-nudge submits, whose between-attempt guard leans on the same busy-pane check that the filed issue reports as blind on Claude Code 2.1.x; it does not change the busy-indicator detection, the submit-confirm poll budget, or the missing pre-send busy snapshot, so it is adjacent context rather than a remedy


jacobhausler:
Rebased onto current main and fixed the mechanical failures (all 7 new tests registered in the runtime-tmux manifest, shard counts updated 341→348, census baseline rebanked, TESTING.md regenerated).

`TestPhase2WorkerCoreRealTransportProof/claude/tmux-cli` will likely stay red here for the same reason as #5012 (this branch carries that confirm-bool fix as its base): fork-PR runs have no live claude credential, so a real confirmation cannot happen — the retry-with-backoff visibly tries harder (35–42s of retries vs ~4–10s before) but the limitation is credentials, not retry count. Local runs of all 7 new tests pass.

quad341 (dismissed):
@jacobhausler — the timing experiment behind this is the best part of the submission, and it survives everything below. Proving that a submit at ~0.3s sits unconfirmed while the identical injection at ~6s lands and gets answered, *and* ruling out the missing-per-provider-key-sequence theory in the same run, is the difference between a real fix and a plausible-sounding sleep. Putting the retry at the one call site instead of a global startup delay was right, and `SucceedsImmediatelyNeverSleeps` pins down that a healthy boot pays nothing. Your #5012 landed as `cfee9389a` and is doing its job on main.

Here's what happened, and I'm sorry it cost you a round: #5012 picked up a change during its own review that isn't in the copy you branched from. The landed version touches `adapter.go` — it wraps the Step 6 `sendKeys` in a guard that turns `ErrNudgeSubmitUnconfirmed` into a stderr warning instead of a start failure, with the comment *"The startup nudge has no retry-capable caller."* So the breaker loop this PR is written against can't happen on main today. I started the rebase, hit that, and stopped rather than reshape the PR into something you hadn't agreed to.

**The good news: your fix is still needed, for a better reason than the one in the body.** That warning means the session starts and the nudge may sit drafted-and-unsubmitted — the agent just silently never begins work. #5012's own comment names the gap. Your re-paste-and-resubmit loop is the thing that closes it. It's a stronger story than "prevents a hard failure": it's "prevents a session that looks healthy and does nothing."

**The one thing I need from you, that I can't answer myself.** In your ARM-A run, when the ~0.3s submit came back unconfirmed — did you capture the pane? Specifically: was the nudge text still sitting *drafted* in the input line, or had it been submitted and the busy indicator just wasn't observed in budget?

That single fact decides the design:

- **Drafted (true negative)** → the retry is the only thing that saves the session, and this ships close to as-is.
- **Submitted-but-unobserved (false negative)** → re-pasting double-delivers the startup prompt, since `NudgeSession` pastes before any busy check and skips the pre-Enter Escape for the claude family. Under your original premise that tradeoff was fine — a doubled prompt beats a breaker loop. Against a benign warning, it isn't, and we'd want a busy pre-check before the re-paste.

If you still have the captures, a paste of the ARM-A pane at the unconfirmed moment settles it in one message. If not, `buildBusyOnEnterBinary` can reproduce it: fake consumes the submit but suppresses the busy footer on turn 1 and shows it on turn 2, then assert via `CapturePaneAll` that the nudge text was submitted exactly once.

Once that's settled, the rest is mechanical:

1. Rebase onto current `main` and drop your carried copy of #5012 — the `tmux.go` hunks and both integration-test tolerances all landed. #5018 (`d71338bc6`) also refactored `sendEnter`→`sendSubmit` in there, so take main's side on every `tmux.go` conflict.
2. Layer the retry **on top of** main's guard rather than replacing it: retry on `ErrNudgeSubmitUnconfirmed`, and on exhaustion fall through to main's warning instead of failing the start. That means `TestDoStartSession_FailsStartAfterExhaustingNudgeRetries` needs to become success-with-warning — otherwise it directly contradicts `TestDoStartSessionReturnsNudgeDeliveryError/"unconfirmed submit is not fatal"` at `startup_test.go:946`. I deliberately didn't flip that assertion for you; it's your test's claim to make.
3. Rewrite the summary against the real problem — silent non-delivery, not start failure. Worth saying out loud that a slow boot now pays up to ~20s to rescue a session that would otherwise sit idle indefinitely; that framing makes the cost obviously worth it, where the old framing doesn't.
4. Take the `ctx` already in scope at `adapter.go:1314` into the backoff so a cancelled start doesn't sleep through the ladder against the 60s `startup_timeout`, plus one cancellation test. Small, and it removes the one way this fix could push a slow boot *past* the budget it's trying to protect.
5. Drop `PR-BODY.md` — it rode along from your authoring setup.
6. Re-bank counters from current `main`, not from the branch: `fixed_sleep` 436/161, manifest 346, untagged 232, integration-only 114, shards `{58,58,58,58,57,57}`. New manifest entries go in source-declaration order; the drift check is order-sensitive.

Keep the four pure-function tests, the `sendKeysErrs` pattern, and the fail-before proof — all of that is good as-is. And keep the phase2 `esc to interrupt` one-liner; you're right about why the fake needs it, and it hasn't landed any other way.

Thanks for the care on this one. The hard part — knowing *why* the submit doesn't land — is done, and it's yours.

jacobhausler:
@quad341 — thanks for the detailed read, and for catching the #5012 drift before it turned into a wasted rebase.

On your factual question: the local run record for the Arm A/B timing experiment (the one that proved the ~0.3s vs ~6s timing hypothesis) states the outcome as drafted-and-unsubmitted — a true negative, not a submitted-but-unobserved false negative. Specifically: "Arm A — text+Enter injected at t~0.3s into a booting claude TUI sits unsubmitted (failure reproduced, matches siuan's 6x 'busy never observed')." That's consistent with a separate, earlier live incident that first motivated this investigation, where the composer text was "observed still sitting visibly unsubmitted, not silently-submitted-but-unconfirmed" — the same true-negative shape recurring independently, not just in the one controlled run.

To be precise about what this evidence is: it's the contemporaneous conclusion recorded in the experiment's run notes, not a raw pane-capture file or screenshot. I looked for the capture and couldn't find one, and there's a structural reason none survives: Claude's TUI runs in the alternate screen buffer and full-screen-redraws, so tmux scrollback doesn't retain stall-window content after the fact. Flagging that plainly rather than overstating certainty — but no local evidence contradicts the drafted conclusion.

Branch is rebased onto current `main` and reworked on the drafted (true negative) branch — no busy pre-check needed, retry ships close to as-is:

1. Rebased onto current `main`, dropped the carried #5012 copy (already landed there), and rebuilt this fix from scratch on top of it rather than replaying stale commits.
2. The retry is layered inside #5012's warning-not-failure guard rather than replacing it: `sendStartupNudgeWithRetry` now takes `ctx` and retries on `ErrNudgeSubmitUnconfirmed`; once retries are exhausted (or a non-retryable error occurs), it falls through to main's existing stderr warning instead of failing the start. `TestDoStartSession_FailsStartAfterExhaustingNudgeRetries` is now `TestDoStartSession_WarnsAfterExhaustingNudgeRetries`, asserting `doStartSession` returns nil after exhausting the ladder — consistent with `TestDoStartSessionReturnsNudgeDeliveryError`'s "unconfirmed submit is not fatal even after exhausting retries" case rather than contradicting it.
3. Summary is rewritten around the real problem: silent non-delivery, not start failure. The session starts and looks healthy; without the retry, the agent can sit indefinitely with the nudge still drafted in its input line. A boot that never confirms now costs at most ~20s to substantially improve the odds of rescuing that session.
4. `ctx` is threaded into the backoff loop: it's checked before each sleep, so a start already being cancelled elsewhere (e.g. the supervising `startup_timeout` deadline) doesn't sleep through the remainder of the ladder. Covered by `TestSendStartupNudgeWithRetry_CancelledContextStopsWithoutSleeping`.
5. No `PR-BODY.md` on this branch — it was authored fresh off current `main`, not carried forward.
6. Counters/manifest are rebanked against current `main` as of this push (it's moved since your review): runtime-tmux manifest 364 total / 247 untagged / 117 integration-only, six shards of `{61,61,61,61,60,60}`. No resourcecensus rebank was needed — none of the 7 new/changed tests add a tracked resource call site (no `time.Sleep`, subprocess, env, or listener calls in the test source itself).

Kept the four pure-function tests, the `sendKeysErrs` pattern, the fail-before proof, and the phase2 `esc to interrupt` fake-CLI fix as-is (that one's still on `main` from #5012's own line, untouched here).

`go build ./...` and `go vet ./...` clean. `go test ./internal/runtime/tmux/... ./internal/session/... ./scripts/... ./internal/testpolicy/resourcecensus/...` all pass.

jacobhausler:
CI red was ours, both fixed:

- misspell: `cancelled` → `canceled` (US locale) — 4 spots in startup_test.go/adapter.go (identifier, comment, error string).
- `TestPhase2WorkerCoreRealTransportProof/claude/tmux-cli` was a real bug: `sendStartupNudgeWithRetry`'s ctx check only gated the backoff *sleep*, but each unconfirmed send itself costs ~2-3s (submitEnterAndConfirm polling for the busy indicator), so two exhausted attempts alone could blow the startup deadline before the warning-and-continue fallback got control back. Bounded the ladder to half of ctx's remaining deadline, measured as actual wall-clock elapsed (not just intended sleep time), so it always leaves the rest of startup room to finish on the warning path.

Repro'd locally first (same failure, ~8.7s), confirmed the fix drops it to ~3.4s and green. Full tmux package + manifest/census tests pass; lint-affected is clean.

jacobhausler:
The remaining red (`TestRepositoryLedgerMatchesCensusAndDocumentation`) was mine and is fixed in a6f76287f: yesterday's budget-bound fix added two real `time.Sleep` calls to the test's send/sleep fakes, which tripped the resourcecensus fixed_sleep ratchet (288 untagged calls vs the banked 286). Rather than rebanking, the sleeps are gone: `sendStartupNudgeWithRetry` now reads the clock through a package seam (`startupNudgeNow`, same shape as `internal/extmsg`'s `timeNow`) and the budget test advances a fake clock from its fakes — same wall-clock bound proven, now deterministic, and the census is back at baseline. Local: `go vet` plus full `internal/runtime/tmux`, `testpolicy/resourcecensus`, and `scripts` packages all pass.

jacobhausler:
@quad341 — the branch is now rebased onto current `main` at `1d99cd8b`, the review-scoped retry/cancellation changes are preserved, the stale hard-failure/#5012 description is rewritten, and the runtime-tmux inventory is rebanked against today’s tree. PR-scoped tests, lint, vet, and pre-commit pass. The full fast gate’s only failure is the deterministic current-main macOS pidutil test already tracked in #5344 (no pidutil diff here). GitHub will not let this fork author use `RequestReviewsByLogin`, so this comment is the re-review request.

quad341:
@jacobhausler — you told us on **Aug 17** that this was rebased onto current `main`, and we didn't answer. Apologies; that's on our review process.

**Verdict: fix-merge.** Key claims were verified against the checkout and CI, not just taken from the diff.

> fundamental — it alters control flow in the session-start path (`launchOrchestration` Step 6 goes from one send to a bounded send/sleep ladder with a deadline budget). Behavior on the happy path is unchanged, but the unconfirmed path now retries and can consume wall-clock, so this is not merely behavior-preserving.

That framing — happy path unchanged, unconfirmed path now costs wall-clock — is the right way to describe the tradeoff, and the review accepts it.

Because it's a fundamental change to a critical path, it needs a human sign-off before merge. That gate is where this has been sitting, and we're working that backlog now.

jacobhausler:
Re-confirming the drafted-vs-submitted evidence with file:line citations, since it settled the branch's design:

- internal/runtime/tmux/tmux.go:78-88 (pre-existing on main, from #5018/d71338bc62): "the live specimens showed text sitting visibly unsubmitted (not a silently-succeeded-but-unconfirmed false negative), ruling out a busy-indicator detection gap as the explanation too." This is the standing empirical record for the claude-family stall this PR retries around.
- The branch's own code treats it the same way: internal/runtime/tmux/adapter.go:1394-1395 documents the exhausted-retry case as "the agent may sit silently idle with the nudge still drafted in its input line," and internal/runtime/tmux/startup_test.go:1204-1211 (TestDoStartSession_WarnsAfterExhaustingNudgeRetries) asserts success-with-warning on exactly that basis.

No pane-capture artifact survives (claude's TUI redraws the alternate screen buffer, so tmux scrollback doesn't retain the stall window) — the record above is the contemporaneous conclusion, not a raw capture, but nothing on this branch or main contradicts it.

Rebased onto main 1e8f5ce74: one conflict, in scripts/runtime_tmux_manifest_test.go (the runtime-tmux inventory counters), re-derived against the merged manifest to manifest=396 / untagged=276 / integration-only=120, shards {66,66,66,66,66,66}. No other files conflicted; PR-BODY.md was already absent. New head: 712cbfa29.

Tests: go build ./... clean; go test ./internal/runtime/tmux/... ./internal/session/... ./scripts/... ./internal/testpolicy/resourcecensus/... all ok.

jacobhausler:
Rebased onto main b71fadc28 (base.sha still predated scripts/prwatchdog despite the earlier same-day rebase). Conflicts: scripts/runtime_tmux_manifest_test.go — hardcoded manifest/shard counts from this PR's own commit (390/270/65×6) drifted against the actual current inventory after rebasing onto main's growth; reconciled to the true discovered counts (398 manifest tests, 278 untagged, shards 67/67/66/66/66/66), verified by TestRuntimeTmuxManifestMatchesCanonicalLinuxIntegrationInventory and TestRuntimeTmuxManifestSixShardsPartitionInventoryExactlyOnce passing locally.

CI failure investigation (Integration / packages-core-3-of-4, run 32868843681): TestRecordOnceFreshQuotaBootstrapNeverReplacesDestinationRace, TestRecordOnceSpentBudgetAfterReservationLeavesOnlySafeOvercount, TestRecordOnceDecisionWindowGatesEveryForegroundQuotaBoundary in internal/productmetrics (spool_unix_test.go). These are unrelated to this PR's diff (internal/runtime/tmux only). Ran full internal/productmetrics package x3 with -count=1 on both this branch and origin/main (macOS/go1.26.6): all pass, every run, both sides — not reproducible locally or attributable to this PR's changes. Consistent with the same-day CI-environment flakiness already logged for #5446 (same test file/package) in CI-REPORT-3.md.

Tests: internal/runtime/tmux, scripts (incl. -tags integration) ok.

jacobhausler:
Rebased onto current main (clean, no conflicts) at 5b4e8ebd to retrigger CI — the prior Evidence/critical-path-suites failure was a watchdog timing race (checked CI/required's state as "incomplete" mid-flight at PR_HEAD_SHA ad3c8203, before that same commit's CI/required run 32879293904 finished and passed a few seconds later); not a real CI regression. Local go build ./... && go vet ./... clean; internal/runtime/tmux package tests pass.

jacobhausler:
CI note: run 32900336340 failed only in `Integration / rest-smoke-2-of-2` — `TestGraphWorkflowSuccessPath` hit `gc init` → "failed to initialize schema: pending ignored schema migrations alter pre-existing dirty tables: child_counters" on a brand-new temp dir. This PR touches only `internal/runtime/tmux`; the same job passed on main 28 min earlier. Retriggered with an empty commit (626ac913). If maintainers see that schema-init message elsewhere it may be worth its own issue.

jacobhausler:
CI is green after the rebase onto current main (HEAD b911c7db). All quad341 CHANGES_REQUESTED review points from 2026-08-05 were addressed in the 2026-08-10 PR replies (drafted-vs-submitted question resolved, retry reworked inside #5012's warning guard, ctx threaded into backoff, PR-BODY.md dropped, counters rebanked) plus two follow-up CI fixes. Requesting re-review from @quad341 — I don't have permission to use the GitHub reviewer-request API directly, so flagging here instead.

sJohnsonStoever:
A data point from `main@615f5b7942220ee02f6825b9d3d52b7b4b9e9224` (source build 2026-09-01, before this PR; macOS 26.6.2; Claude Code TUI sessions), about the retry's precondition rather than its mechanics.

In one ride the warning this PR retries on fired on 6 of 6 Claude Code pool/named session starts (architect ×2, project-manager ×2, refinery ×2; the two mayor starts did not warn). In every case the session's prompt had gone in via argv (the refinery prompt is ~28 KB, under the 100 KB nudge fallback), and the agent was already inside its first turn when the startup nudge was typed:

- refinery, runtime 1: transcript first user record 15:06:59, first assistant record 15:07:05; the warning is logged between 15:06:59 and 15:07:02; `session.woke` at 15:07:05.
- refinery, runtime 2: 15:07:39 / 15:07:43; warning between 15:07:39 and 15:07:48.

So `submitEnterAndConfirm`'s budget (3 sends × 4 polls × 150 ms plus backoffs, about 2.2 s) reported "busy state never observed" on panes that were demonstrably mid-turn. Whether the nudge then sat as a draft or was queued behind the running turn we cannot tell from the outside: it never appears in either transcript, and runtime 2 ran one continuous turn for five minutes and drained itself before any queued input would have been consumed.

Why that matters for this PR: on a session in that state, `sendStartupNudgeWithRetry` would run the full C-u / paste / submit cycle four more times over ~20 s, each attempt confirmed by the same probe that just missed a busy pane, so the ladder always exhausts. If the first submit did queue (Claude Code queues input submitted mid-turn, and C-u clears the composer, not the queue), each retry enqueues another copy. In our city that is +20 s on every Claude Code pool start for no gain.

A cheap precondition would separate the case this PR is built for (idle pane, draft still in the composer) from ours: before each resend, capture the pane and retry only if the pasted text is still visible in the composer and no busy indicator is present; an empty composer means the submit went through. Separately, the probe missing a mid-turn pane at all (which spinner form current Claude Code renders in a detached pane during the first seconds of a large argv prompt) may deserve its own look; happy to gather pane captures if that helps.

jacobhausler:
@sJohnsonStoever — thank you, this is exactly the kind of data point that matters here: concrete timestamps against a live Claude Code pool ride, not a hypothesis.

You're right about the mechanism, and the fix is smaller than it might look: `submitEnterAndConfirm`'s own busy-poll is short (~2s per attempt) and can legitimately miss a pane that renders its busy indicator a little later on a large argv prompt — exactly your refinery traces. The bug is that `sendStartupNudgeWithRetry`'s *outer* ladder never re-checked busy at all before its next resend; it just slept out the backoff and blindly re-ran `send` (full C-u/paste/submit) again. As you note, C-u clears the composer but not a queue Claude Code may already be holding, so each of those blind resends could stack another copy behind whatever landed.

Pushed a fix (`b9b7595a`): `sendStartupNudgeWithRetry` now takes a `busy` check and, once a backoff has elapsed, skips the resend entirely (treating it as delivered) if the pane already shows the busy indicator by then. It reuses the same `paneBusy` primitive the inner confirm already relies on — no new pane-parsing, no composer-text diffing — just asking the question one more time, at the point the outer ladder is about to act on stale information. Added `TestSendStartupNudgeWithRetry_StopsRetryingOncePaneGoesBusy` and `_KeepsRetryingWhilePaneStaysIdle` to pin both sides of that behavior.

I did not implement the "composer still shows the drafted text" half of your suggestion — there's no existing per-provider composer-content detector in this codebase, and building one generically (across claude/codex/gemini/opencode/kimi/grok's differing TUI framing) felt like a separate, larger piece of work I'd rather not bundle into this fix's risk surface. The busy-check alone should remove the amplification you measured (the outer ladder stops as soon as busy is observed instead of running up to 4 more full cycles), while leaving the "self-heal a genuinely cleared draft" behavior intact for the case this PR was built for.

Local `gofmt` is clean; I don't have a local build lane available right now, so this push relies on CI to confirm compilation/tests — will watch and follow up if anything's red.

quad341 (approved):
@jacobhausler — the Arm A/B timing experiment is still the best thing in this submission. Establishing "~0.3s doesn't land, ~6s does" as a measured result rather than a hunch is what turns a plausible race into a fix anyone can reason about, and it's why the backoff ladder has the shape it has instead of an arbitrary sleep.

What I especially appreciated: when @sJohnsonStoever showed up with field data suggesting the retry could *amplify* against already-busy panes, you took the half of it you could implement soundly, said plainly why you were leaving the composer-content detector alone, and pinned both branches with tests. The fake-clock seam instead of rebanking the sleep census was the right call too.

CI red is one line of bookkeeping on our side: the last push added `_StopsRetryingOncePaneGoesBusy` and `_KeepsRetryingWhilePaneStaysIdle` but didn't register them in `runtime-tmux-tests.manifest`. That's the entire failure — the runtime-tmux shards themselves are green. I'm taking that, the rebase, and the counter re-derivation (you've hand-chased those numbers across four rebases; that's our gate being annoying, not your problem). I'm also making the backoff itself context-cancelable so it matches what your comment already promises.

Merging with your authorship intact. One thread I'll open separately, not for you: whether the busy probe can see a mid-turn pane at all in @sJohnsonStoever's setup.


---

Issue #4163, opened 2026-07-11
feat: gc order set-interval + StringMap metadata coercion fix + DeleteAfterClose retention field


tdupu:
### Before you continue

- [x] I searched existing issues and did not find a duplicate.
- [x] I can explain the user problem this solves, not just the implementation I prefer.

### Problem to solve

 Three related pain points in order management:
  
  1. No CLI subcommand to set order cooldown intervals at runtime. interval_min/max/alert_after/critical_after were added to the Order struct but operators can only set them by hand-editing city.toml. Agents that need to mutate intervals at
  runtime have no policy-compliant path.
  
  2. Mixed-type bead metadata silently poisons gc hook --claim work queries. bd update --set-metadata key=true emits JSON boolean; the map[string]string decode fails; because the work query fetches a batch, a single bead with non-string
  metadata drops the entire batch — every worker in the rig fails to claim work. Silent correctness bug with no visible error.
  
  3. No per-order retention TTL for closed order-tracking beads. Orders with short cooldown intervals (e.g. 5m) generate bead churn; there's no way to set a shorter delete_after_close than the city-level policy without overriding it globally.

### Proposed change

  1. gc order set-interval <name> <duration> [--rig <rig>]: new subcommand in cmd/gc/cmd_order.go. Creates or updates [[orders.overrides]] with the named order's interval field set to the given Go duration string (30m, 2h, 24h). Validates
  duration before writing. Uses configedit (same path as other order mutation commands).
  
  2. StringMap type in internal/beads/beads.go replacing map[string]string as the Metadata field type on Bead. Implements json.Unmarshaler, coercing non-string JSON values (bool, number, array, null) to their string representation on decode.
  Underlying type is map[string]string — all call sites and the wire format are unchanged. 
  
  3. DeleteAfterClose string field on Order struct. Accepts Go duration syntax plus whole-day units (48h, 7d). Overrides city-level [beads.policies.order_tracking].delete_after_close per-order. Required for orders with cooldown intervals under
  15m.
  
  Branch: tdupu/gascity:sync/set-interval-preserve (commit 663bd3bf)

### Alternatives considered

  - For interval mutation: hand-editing city.toml (error-prone, no validation, requires city restart); using [[orders.overrides]] directly (works but no CLI path for agents).
  - For StringMap: strict validation at bd update time (breaks existing usage); ignoring non-string fields (masks bugs). Coercion is the only backwards-compatible option.

### Scope and impact

[...]

PR #3857, opened 2026-07-01
fix(beads): tolerate non-string bead metadata on decode

wbern:
## Problem

`bd`'s `--set-metadata key=true` (and numeric values) are persisted as JSON **booleans/numbers** rather than strings. `decodeHookClaimBeads` unmarshals the entire `work_query` result into `[]beads.Bead` in a single pass, and `Bead.Metadata` is `map[string]string`, so a single bead carrying a non-string metadata value fails the **whole batch** decode:

```
json: cannot unmarshal bool into Go struct field Bead.metadata of type string
```

Because the decode is one pass over the whole result set, one such bead makes `gc hook --claim` return an error for **every** worker sharing that `work_query` — i.e. a single bead can block an entire rig from claiming work. Keys observed in the wild: `refinery_reviewed`, `no_e2e_waiver`, `gc.parked`.

## Fix

Change `Bead.Metadata` from `map[string]string` to the existing **`beads.StringMap`**, which was introduced in #1051 for cache-event metadata and coerces bool/number values to their string form on decode. This just applies that same tolerance to the bead decode path.

`StringMap`'s underlying type is `map[string]string`, so:
- every read/write call site is unchanged (`go build ./...` clean);
- the marshaled wire form is unchanged (still string-valued);
- the generated OpenAPI spec is unchanged — confirmed by the spec-in-sync test (Huma renders it identically as `additionalProperties: string`, no new `$ref`).

## Tests

- `TestDecodeHookClaimBeadsToleratesNonStringMetadata` — boolean/number metadata decodes, coerced to `"true"` / `"42"`.
- `TestDecodeHookClaimBeadsOneBadBeadDoesNotPoisonBatch` — a bool-metadata bead alongside good beads no longer drops the batch.

Both fail on `main` with the error above and pass with the change.

## Notes

The `--set-metadata` type-inference itself lives in `bd`; this change makes the reader tolerant so a boolean/number value is harmless regardless. Other `map[string]string` metadata decode sites that consume bd bead output (`bdIssue`, cache events) already use `StringMap`; this closes the remaining one on the `Bead` decode path.

Issue #1690, opened 2026-05-04
Typed dispatch metadata: close the type-safety gap between Handle layer and dispatch coordination


csells:
@julianknutsen -- some random but potentially useful feedback from Claude:

## Problem

The dispatch metadata layer is untyped while everything around it is typed. The Handle layer, the session state machine, the provider interface, the event system — all well-typed. The dispatch/formula/convoy coordination substrate — stringly-typed.

Evidence:
- 526 raw `Metadata["gc.*"]` accesses in the dispatch package, zero typed constants
- 3,106 raw metadata accesses across the full codebase
- Values parsed via `strconv.Atoi` and `strings.TrimSpace` with no validation framework
- `setOutcomeAndClose()` accepts outcome as a raw `string` parameter
- `gc.outcome`, `gc.failure_class`, `gc.attempt`, `gc.max_attempts`, `gc.on_exhausted`, `gc.routed_to`, `gc.deferred_assignee`, `gc.deferred_type` — all bare string literals
- The convergence package already proves the typed-constant pattern works in this codebase (`GatePass`, `GateFail`, `GateTimeout`, `GateError` in `internal/convergence/metadata.go`)

A typo in `"gc.outcoem"` is a silent runtime bug today. It should be a compile error.

### Phase 1: Typed constants and accessors

New file (e.g., `internal/dispatch/meta.go` or a small `internal/dispatch/meta/` package) defining:

- **Key constants** for every `gc.*` metadata key currently scattered as string literals
- **Value enums** (typed string aliases) for outcome (`pass`/`fail`/`skipped`), failure class (`hard`/`transient`), exhausted policy (`hard_fail`/`degraded`), etc.
- **Typed accessors** (`GetOutcome`, `SetOutcome`, `GetAttempt`, `SetAttempt`) that handle the string↔typed boundary in one place, including `strconv` parsing for integer fields

### Phase 2: Mechanical migration

Replace every `bead.Metadata["gc.outcome"]` with `meta.GetOutcome(bead.Metadata)` and every `bead.Metadata["gc.outcome"] = "pass"` with `meta.SetOutcome(bead.Metadata, meta.OutcomePass)`. Same semantics, typed surface.

### Phase 3: Lock the door

- Conformance test: every constant round-trips through bead metadata
- Lint test (`TestNoRawDispatchMetadataKeys`): grep for bare `"gc.` string literals in dispatch/formula/convoy/sling/molecule package files, fail if any are found outside the meta package

[...]

PR #5024, opened 2026-08-05
fix(beads): keep untouched metadata keys from being retyped on write

wbern:
## Summary

Writing one metadata key rewrote the JSON type of every *other* key on that bead.

A bead holding `{"flag": true, "num": 42}` came back as `{"flag": "true", "num": "42"}` after an unrelated key was set. The write reports success, so nothing surfaces at the time — a consumer doing a typed read (`jq '.num == 42'`, a numeric comparison, a `bool` assertion) just silently stops matching some time later, after any unrelated update touched the same bead.

Metadata is stored as a single JSON document, so setting one key does require rewriting the whole document. The problem is that every write path decoded it into `map[string]string` first — which renders each non-string value as JSON text — and then marshalled that back. This preserves the caller's key and collaterally retypes everything else.

## Trigger

Any write to a bead that already holds a non-string metadata value. `bd` infers JSON types from `--set-metadata`, so numeric and boolean values occur normally in practice.

| Stored before | Write | Stored after (before this PR) |
|---|---|---|
| `{"flag": true, "num": 42}` | set `other=x` | `{"flag": "true", "num": "42", "other": "x"}` |
| `{"nested": {"a": 1}}` | set `other=x` | `{"nested": "{\"a\":1}", "other": "x"}` |
| `{"num": 42}` | `SetMetadataBatch(id, {})` | `{"num": "42"}` |

The last row is a no-op call that retyped the document.

## Behavior

**Given** a bead whose metadata holds a non-string value
**When** an update names a *different* key
**Then** the untouched value keeps its stored JSON representation

**Given** a bead whose metadata holds `{"num": 42}`
**When** an update names `num`
**Then** it is written as the JSON string the caller supplied — unchanged, since `UpdateOpts.Metadata` is `map[string]string`

**Given** a compare-and-set against a stored non-string value
**When** the expected value is its JSON text
**Then** the compare still matches — the CAS compare path is untouched

## Why it's safe

`metadataRawWithOverrides` merges named keys into the stored document as `json.RawMessage`, so values the caller did not name are copied through byte-identically. Keys the caller *does* name are still written as JSON strings, matching the `map[string]string` the write API accepts — so no request changes meaning, and the wire contract is unchanged.

The change is confined to the four write paths that shared the decode-merge-encode block: `Update`, `SetMetadataBatch` (standalone and in-transaction), the graph-plan metadata-ref merge, and `CompareAndSetMetadataKey`. `CompareAndSetMetadataKey` still *compares* through `metadataMapFromNative`, so a stored non-string value keeps comparing against its JSON text exactly as before; only its write changed.

The read paths that legitimately want Go strings — `beadFromNativeIssue`, `nativeCloseReasonFromIssue` — are deliberately left alone. `Bead.Metadata` stays `map[string]string`; this PR does not revisit how metadata is *represented* to Go callers, only which keys a write is allowed to modify.

## Related work

- #3857 — tolerate non-string bead metadata on decode. The read-side sibling: it stopped a single non-string value from failing a whole batch decode. This PR is the write-side counterpart, so a value that decodes tolerantly is no longer written back retyped.
- #4163 — proposes `StringMap` coercion on `Bead.Metadata`. Same territory on the read side (and largely landed in #3857); orthogonal to the write-scope fix here.
- #1690 — typed dispatch metadata. Longer-term direction for typing metadata end-to-end; this PR is a containment fix that does not preempt it.

## Scope

| File | Change |
|---|---|
| `internal/beads/native_dolt_store.go` | Adds `metadataRawWithOverrides`; routes three write paths through it |
| `internal/beads/native_dolt_store_conditional.go` | Routes the CAS write through it; compare unchanged |
| `internal/beads/native_dolt_store_metadata_preserve_test.go` | New — type preservation, named-key overwrite, absent/`null`/empty documents, malformed JSON |
| `internal/beads/native_dolt_store_metadata_preserve_integration_test.go` | New (`//go:build integration`) — the same properties end to end against real native storage |

## Testing

- [x] `go build ./...`, `go vet ./...` clean
- [x] `golangci-lint run ./internal/beads/...` — 0 issues
- [x] `go test ./internal/beads/` — full package green
- [x] `go test -tags=integration ./internal/beads/` — full integration tier green
- [x] `go test ./internal/api/ ./internal/dispatch/ ./internal/convoy/ ./internal/sling/ ./internal/beadmeta/` green

The integration tests seed a metadata document holding genuine JSON types (bypassing the `map[string]string` write API, which cannot express one), perform a write, and assert against the document **real storage actually persisted** rather than the map handed to the driver.

Reverting only the two source files to `main` and re-running them reproduces the defect on all three write paths:

```
--- FAIL: TestNativeDoltStoreUpdatePreservesMetadataTypesAgainstRealDolt
    flag = "true" (string), want bool true — an untouched key was retyped
    num = "42" (string), want JSON number 42 — an untouched key was retyped
    nested = "{\"a\":1}" (string), want JSON object — an untouched key was retyped
--- FAIL: TestNativeDoltStoreSetMetadataBatchPreservesTypesAgainstRealDolt
--- FAIL: TestNativeDoltStoreCASPreservesOtherMetadataTypesAgainstRealDolt
```

`TestNativeDoltStoreCASComparesAgainstStoredNonStringValue` passes both before and after by design — it guards the CAS compare semantics that this PR deliberately leaves unchanged.


wbern:
One thing worth flagging myself, since it isn't visible in the diff.

The new `//go:build integration` tests open real native storage via `beadslib.OpenBestAvailable`, so I checked them against the source-resource ledger in `test/test-resources.toml` before assuming they were free. They use none of the counted identities — no `time.Sleep`, `exec.Command`, listeners, `httptest`, environment or CWD mutation, and no tmux dependency:

```
go test ./internal/testpolicy/resourcecensus/   ok
go test ./internal/testenv/                     ok
cmd/gc  -run Census                             PASS
```

So no ledger bump or policy waiver is needed here. Recording it because a real-storage test is exactly the shape that usually does need one, and I'd rather you didn't have to find that out by running it.

Two notes on scope that may save a pass over the diff:

- `metadataMapFromNative` is deliberately left in place on the two Go-facing read paths (`beadFromNativeIssue`, `nativeCloseReasonFromIssue`). `Bead.Metadata` stays `map[string]string` and still coerces on read, so no Go consumer can observe a behavior change from this PR. Only readers of the raw JSON column see the corrected types.
- `CompareAndSetMetadataKey` still *compares* through `metadataMapFromNative`; only its write changed. `TestNativeDoltStoreCASComparesAgainstStoredNonStringValue` passes both before and after the fix by design — it guards that unchanged semantics rather than reproducing the bug.

Happy to split the CAS and graph-plan sites into a follow-up if you'd rather review the `Update`/`SetMetadataBatch` pair on its own. I kept all four together because they shared one decode-merge-encode block, and fixing a subset would have left the PR's claim about untouched keys true only in part.

wbern:
Rebased onto current `main` and narrowed — part of this has since been fixed upstream, so it is no longer mine to fix.

**Dropped: the compare-and-set path.** `compareAndSetMetadataKeyOnce` now reads raw values through `metadataRawValuesFromNative` (`native_dolt_store_conditional.go:318`), sets only the target key, and marshals — which is exactly what this PR was proposing for that site. I took `main`'s version wholesale; the conflict here was the two fixes meeting.

**Still live: the general write paths.** These continue to round-trip through `map[string]string`, so a key the caller never named is rewritten as a JSON string:

- `applySetMetadataBatchInTx` (`native_dolt_store.go:1183-1193` on `main`)
- `setMetadataBatchOnce`
- `nativeUpdates`
- the `MetadataRefs` merge in `ApplyGraphPlanWithStorage`

All four now go through one `metadataRawWithOverrides` helper that decodes to `map[string]json.RawMessage`, replaces only the named keys, and re-marshals.

## Given / When / Then

- **Given** a bead whose metadata holds `{"flag": true, "num": 42, "nested": {"a": 1}}`, **when** an unrelated key is written, **then** those three keep their JSON types instead of becoming `"true"`, `"42"` and `"{\"a\":1}"`.
- **Given** a key the caller names in `UpdateOpts.Metadata`, **when** it is written, **then** it is a JSON string — `UpdateOpts.Metadata` is `map[string]string` and that half of the contract is unchanged.
- **Given** a bead with no metadata, **when** keys are written, **then** the object is created as before.

## On coverage

The unit tests drive `nativeUpdates` directly and fail loudly if it regresses — swapping the helper back for the old merge reports `flag = "true" (string), want bool true`. The other three call sites are covered by the integration tests against a real Dolt, which skip when one is not available, so on a machine without Dolt those paths are compile-checked rather than behaviour-checked. Flagging that rather than implying uniform coverage.

quad341 (approved):
@wbern — merging this. You found a silent data-corruption bug with a genuinely nasty shape: the write reports success, and the damage only surfaces later when some unrelated consumer's typed read quietly stops matching. Routing all four read-modify-write sites through one `metadataRawWithOverrides` instead of patching them individually is the right call, and keeping named keys as JSON strings means the `map[string]string` write contract doesn't shift underneath anyone.

The test I'm happiest about is `TestMetadataRawWithOverridesRejectsMalformedJSON`. A corrupt document silently becoming an overrides-only document is the worst possible version of this bug, and you pinned it before anyone asked. Seeding raw types in the integration tests to get around your own write API not being able to express one — then asserting against what storage actually persisted rather than what got handed to the driver — is exactly the right paranoia.

Two things I'll take on our side, neither yours to fix: `CloseWithMetadataIfMatch` has the identical bug one site over, and the helper's "byte-identical" comment is a shade stronger than `encoding/json` actually guarantees. I'll file both. Happy to hand either back if you'd like them.


---

PR #5025, opened 2026-08-05
fix(usage): recover model/compute facts for recently-closed sessions

haddad-daniel:
## Summary

[#4994](https://github.com/gastownhall/gascity/pull/4994) sweeps **live** awake sessions and deliberately left out closed-row enumeration (fleet cost on `IncludeClosed` + `AllowScan`).

This closes the complementary **known v0 gap** (`engdocs/design/usage-facts-v0.md`): the open-set scan under-counts sessions that leave `OpenInfos()` without an open terminal state.

The concrete pool path: drain-ack finalize closes mid-tick (`ClosePatch("drained")`) from `drain-ack-stop-pending`, then reloads the snapshot **before** `emitDueComputeFacts`. The bead never appears as open `asleep`/`drained` for the usage lane. Treating `draining` as terminal does not fix that ordering (#4989 closed for that reason).

## Cost controls

- **Boot skip** — same policy as the live lane.
- **30s floor** — `recentlyClosedUsageMinInterval` matches `liveModelSweepMinInterval`; the closed List is not per-tick.
- **15m window** — only recently closed, unswept beads are candidates after the List.

Related: [#5031](https://github.com/gastownhall/gascity/pull/5031) keyless Claude settle (independent).

## Test plan

- [x] `go test ./cmd/gc/ -run 'RecentlyClosed|LiveModel'`


quad341:
@haddad-daniel — thank you for this one. Closing the drain→close race properly is the part everyone skips, and you did the hard bit: noticing that `ClosePatch` stamps `closed_at` but never `slept_at`, so a naive recovery emit would quietly bill the interval to discovery time instead of close time. Presenting `closed_at` as `slept_at` on a *copy*, so the shared helper's existing preference applies and the open-set lane stays untouched, is nicer than threading a new parameter through `emitComputeFactForBead` — and your test asserting the wall-clock anchoring is exactly the test I'd have asked for. I also appreciate that you took #4994's cost exclusion head-on in the description instead of hoping nobody would look, and explained why #4989's `draining`-as-terminal approach can't fix the ordering.

Three things I'll take on my side, all default-config blind spots rather than anything wrong with your reasoning: the new lane needs `sessionsBeadStore()` rather than `cityBeadStore()` (identical under the default backend, silent no-op on a split city — the sibling lane has a regression test for this shape and I'll add its twin); the List should go through `session.ListAllSessionBeads` to pick up the type leg; and I'll extend the `closed_at` anchoring to the `UpdatedAt` fallback path so it can't over-bill there either.

Worth flagging honestly: CI ran no test suite on this head — every green check is a label bot — so I built and ran `./cmd/gc` locally to check it. Clean.

Applying those, re-running the suite, and merging with your authorship intact. Nice work.


---

PR #5027, opened 2026-08-05
fix(cmd/gc): route gc status through the supervisor API for supervisor-managed cities

jacobhausler:
## What

`gc status` in a supervisor-managed city was measured at **~9.6s wall / ~9.55s user CPU** (26 agents, 1.2GB Dolt store) even though the supervisor's own `/v0/city/{name}/status` answers the same query in ~0.35s warm. `GC_DEBUG=1` showed `route=fallback reason=controller-down` on every call.

## Root cause

`apiClient()` treats an alive per-city controller socket as "try the standalone `[api]` port, else give up (nil)" — it never falls through to the supervisor-managed client in that branch. A supervisor-managed city (the common case) has no standalone `[api]` port in `city.toml`, so `apiClient` always returns nil and `gc status` silently pays its local fallback: re-open the bead/dolt store and rescan event archives to rebuild store health, every single call. `cmd_maintenance.go` already closed this exact gap for itself (9f4ac3e98) because maintenance has no local fallback at all; the status command never got the same treatment.

(We initially suspected a perf fix lost in a rebase between two internal build eras — diffing them shows `apiroute.go` byte-identical; this gap simply predates both.)

## Fix

Extract the maintenance fall-through into a shared `supervisorFallthroughAPIClient()` helper and point `cityStatusAPIClient` at it. `apiClient()` and its existing general-command callers are unchanged; the `GC_NO_API` operator escape hatch is honored in the new path.

## Measured

Same live city: **9.6s wall / 9.55s CPU → 0.5–0.9s wall / 0.19–0.21s CPU**, route flips from `fallback` to `api`.

## Tests

New regression coverage verified to fail against the pre-fix code; full `cmd/gc` suite green locally (one pre-existing, unrelated `TestMain`-level dolt-server flake on a box that hosts a live gascity install, reproduced with a clean tree).


quad341 (approved):
@jacobhausler — the diagnosis here is the best part of this PR. Tracing `route=fallback reason=controller-down` back to `apiClient` bailing to nil on the alive-socket-but-no-`[api]`-port branch, then spotting that `cmd_maintenance.go` had already solved exactly this for itself in 9f4ac3e98, is what turned a mystery slowdown into a three-line fix. Extracting `supervisorFallthroughAPIClient` instead of copy-pasting was the right call, and carrying the `GC_NO_API` guard into the shared helper kept the operator escape hatch from quietly going missing in the move.

The `controller-fully-down-still-falls-back-local` subtest is what makes this safe rather than just fast — asserting a real outage still yields `controller-down` is the test most people would have skipped. Same for asserting `route=fallback` is *absent* end to end.

I'm merging this with two small comment fixes on my side: `apiClient`'s doc block and `TestAPIClientRouting`'s comment still say the supervisor fall-through is maintenance-only, which your change makes untrue. Nothing needed from you — thanks for also ruling out the rebase theory, which is real work that usually goes unmentioned.


---

PR #5011, opened 2026-08-05
fix: guard destructive cached-read deletes with a live re-verify

jacobhausler:
## Summary

Applies the live re-verify guard proven in gascity#5000
(`cmd/gc/work_assignment.go`'s `ReleaseWorkBead`) to the two worst
state-destroying cached-read→write paths flagged by ra-559vsp's audit —
both are irreversible DELETEs driven by a possibly-stale enumeration:

1. **`internal/mail/beadmail/beadmail.go`'s `PurgeReadMessageWisps`** — its
   candidate list answers from a `read:true` metadata snapshot. A message the
   user un-read (or that was deleted by another path) inside the cache window
   was previously deleted anyway on the strength of that stale snapshot.
2. **`cmd/gc/wisp_gc.go`'s `deleteExpiredBeadClosure`** (the closed-root
   descendant-closure purge) — `closedWispGCEntries` can answer from a stale
   cached snapshot. A root reopened (live work resumed) inside the cache
   window would previously have its full descendant closure destructively
   deleted anyway.

## Fix

Both sites now re-read the candidate through the live handle
(`beads.HandlesFor(store).Live.Get`) immediately before the destructive
delete and skip when:
- the live read errors (bead already gone — nothing to delete), or
- the live state disagrees with the stale snapshot (message no longer
  `read:true`; root no longer `closed`).

For the wisp_gc closure-purge path, `purgeExpiredBeads`'s generic delete-loop
counted any nil-error return as "purged," which would have miscounted a
guard-skip as a successful purge — introduced a sentinel error
(`errBeadNoLongerEligible`) so a live-recheck skip is neither counted as
purged nor surfaced as a delete failure.

## Tests

`internal/mail/beadmail/beadmail_retention_test.go` (new, using a
`staleListStore` double that lets `List` answer stale while `Get`/`Delete`
stay live):
- `TestPurgeReadMessageWisps_SkipsMessageUnreadAfterSnapshot`
- `TestPurgeReadMessageWisps_SkipsMessageGoneAfterSnapshot`

`cmd/gc/wisp_gc_test.go` (new):
- `TestWispGC_ClosureSkipsRootReopenedAfterSnapshot`

All three proven fail-before (deleted/purged the stale-snapshot candidate
anyway) / pass-after.

```
go test ./internal/mail/beadmail/...
ok  	github.com/gastownhall/gascity/internal/mail/beadmail

go test ./cmd/gc/ -run TestWispGC -v
... (all PASS)
```

[...]

Issue #4653, opened 2026-07-25
flaky test: internal/productmetrics DisableAndPurge family — load-sensitive race under make test -p=4


remuscazacu:
## Summary

The `DisableAndPurge` tests in `internal/productmetrics/control_unix_test.go` appear to be load-sensitive. Under `make test`'s `-p=4` on a CPU-constrained machine they fail intermittently, on **different tests each run**, while the package passes reliably when run on its own.

I could not produce a deterministic local repro, so I'm filing this as an observation with the evidence and a mechanism hypothesis rather than a diagnosis. Feel free to close if this is known or if CI never sees it.

## Observed

Two full `make check` runs on `main` (`2b265f75c`), same machine, ~15 minutes apart. Each failed on a *different* test in the same family:

**Run 1**

```
--- FAIL: TestDisableAndPurgeExactTokenConflictAndPeerCleanRecovery
--- FAIL: TestDisableAndPurgeExactTokenConflictAndPeerCleanRecovery/clean_config_with_residual_files_is_a_conflict
    control_unix_test.go:1386: purge error = productmetrics: disable-write-failed, want class "state-changed-concurrently"
```

**Run 2**

```
--- FAIL: TestDisableAndPurgeRejectsPeerSuccessorReplacedDuringCleanProof (0.01s)
    control_unix_test.go:1756: peer successor was not replaced during the clean-tree proof
```

Both failed within ~2–10ms of starting, i.e. they lost a race rather than hitting a timeout.

## Does not reproduce in isolation

| Invocation | Result |
|---|---|
| `go test ./internal/productmetrics/` standalone | **3/3 pass** (14.8s, 13.3s, 12.1s) |
| `GOMAXPROCS=1 go test ./internal/productmetrics/ -run TestDisableAndPurge` | pass (2.7s) |
| `go test ./internal/productmetrics/ -run TestDisableAndPurge` | pass (3.1s) |
| inside `make check` (`-p=4`, full suite) | failed 2/2, different test each time |

## Mechanism hypothesis

Both tests share a shape that looks timing-dependent — using `TestDisableAndPurgeRejectsPeerSuccessorReplacedDuringCleanProof` as the clearer example:

1. Take the `uploaderLockName` lock as a `barrier` to hold the purge at a known point.
2. Set `deps.disableUploaderWait = testutil.GoroutineRaceTimeout` (**10s**).
[...]

PR #5028, opened 2026-08-05
fix(internal/api): serve stale /v0/status and refresh in the background on cache-miss

jacobhausler:
## What

`GET /v0/status` (and `/v0/city/{name}/status`) blocks **~3.65s on a cache-miss** in a 26-agent, 1.2GB-store city — `buildStatusBody`'s fan-out (per-rig work counts, session snapshot, StoreHealth's closed-history Dolt scan) lands on whichever request first misses the response cache. The same failure class used to 503 the legacy runs/census endpoint once aggregation outran its internal budget. Dashboard clients see this as a dead status module on first load.

## Fix

A stale-while-revalidate tier on the existing response cache: serve any previously-cached body immediately (real staleness reported through the existing `X-GC-Cache-Age-S`/`CacheAgeS` field — no API shape change), refresh in the background via the existing `runBackground` helper, coalesced by a `responseRefreshing` guard so concurrent misses trigger exactly one rebuild. Only a genuinely cold cache still blocks.

## Tests

New regression test uses the existing `blockingListStore` fake to prove the SWR-served request returns sub-millisecond while the store is deliberately stalled, and that the background refresh completes afterward. Verified to fail against pre-fix code. `go test ./internal/api/...` green.

### DIFF
--- a/internal/api/handler_status.go
+++ b/internal/api/handler_status.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log"
+	"runtime/debug"
 	"slices"
 	"strings"
 	"sync"
@@ -167,6 +169,15 @@ func (s *Server) refreshStatusResponseInBackground(cacheKey string, lite bool) {
 		return
 	}
 	s.runBackground(func(ctx context.Context) {
+		defer func() {
+			if r := recover(); r != nil {
+				// The background refresh is best-effort: a panic here must not
+				// take the controller down. The request path's recover
+				// (withRecovery, middleware.go) does not cover detached
+				// goroutines, and the next poll simply rebuilds.
+				log.Printf("api: panic in background /status refresh: %v\n%s", r, debug.Stack())
+			}
+		}()
 		defer s.endResponseRefresh(cacheKey)
 		resp := s.buildStatusBody(ctx, lite)
 		s.storeResponse(cacheKey, responseCacheTimeBucket(time.Now()), resp)
--- a/internal/api/handler_status.go
+++ b/internal/api/handler_status.go
@@ -122,15 +122,26 @@ func (s *Server) humaHandleStatus(ctx context.Context, input *StatusInput) (*Ind
 		// measured ~3.65s on a 26-agent/1.2GB city — the same failure class
 		// that used to 503 the legacy runs/census endpoint at its internal
 		// budget. Rather than let a request pay that cost inline, serve the
-		// stale body immediately (CacheAgeS reports its real age, reusing the
-		// existing staleness signal `gc status` already renders a banner for
-		// above cacheAgeBannerThresholdSeconds) and refresh in the background
-		// so the next poll gets a fresh body. A genuine cold cache (nothing
-		// ever built) has nothing to serve here and falls through to the
-		// synchronous build below, same as before this change.
+		// stale body immediately and refresh in the background so the next
+		// poll gets a fresh body. A genuine cold cache (nothing ever built)
+		// has nothing to serve here and falls through to the synchronous
+		// build below, same as before this change.
+		//
+		// CacheAgeS reports the GREATER of the two staleness signals: how
+		// long ago this response entry was built, and cacheAgeSeconds(store)
+		// — the age of the CachingStore snapshot the body was built from.
+		// Reporting only the response-entry age would let a recently built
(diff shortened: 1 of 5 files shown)


jacobhausler:
@quad341 — nudge on this: `serve stale /v0/status and refresh in the background on cache-miss`. CI green, mergeable, no conflicts. Could you take another pass when you have a moment? Glad to adjust anything.

quad341:

- **#5028** (*serve stale `/v0/status` and refresh in the background on cache-miss*) is **MERGEABLE / CLEAN** and has been for the full 15 days. Nothing is blocked on you.
- A real ensemble verdict is queued as review capacity frees up, and you will get it on this thread.

quad341 (dismissed):
@jacobhausler — thank you for this, and apologies for the three weeks of silence; that was our review pipeline stalling, not your PR sitting in a queue.

This is a careful retrofit. Gating SWR behind `if !blocking` so long-poll callers still get strict freshness, coalescing on the existing `responseCacheMu` instead of introducing a second lock, and reusing `runBackground` so the refresh outlives the request that triggered it — those are the three things that usually go wrong in a stale-while-revalidate change, and you got all three. Narrowing `TestHandleStatusCacheExpiresOnTTL` with a pointer to its replacement rather than deleting it, and asserting the `responseRefreshing` guard actually cleared, are both the kind of thing people skip.

One thing I'll take on my side rather than send back: the SWR return swaps `CacheAgeS` from `cacheAgeSeconds(store)` to the response-entry age, and those are different signals — the store one drives `gc status`'s "reconciler may be lagging" banner. I'll report the larger of the two, tighten the header assertion, and add the concurrent-miss test your description promises. Then I'll re-run the package and merge with your authorship intact.

I'm also filing a follow-up on capping maximum stale-serve age — a design call, not a gap in your work.


---

quad341:
@jacobhausler — this is careful work, and the details are what make it. Gating SWR behind `if !blocking` so long-poll callers keep strict freshness; leaving `cacheLiveOr503` ahead of the stale branch so a priming store still 503s instead of getting papered over; coalescing on the existing `responseCacheMu` rather than bolting on a second lock; and ordering the defers so `endResponseRefresh` runs *before* the recover handler — that last one is the difference between a panicking refresh costing you a log line and one that wedges every future refresh for that key.

The tests are the same standard: a release channel instead of a sleep, an exact `X-GC-Cache-Age-S` assertion under a lagging reconciler, and coalescing asserted after a barrier rather than on timing. And you narrowed `TestHandleStatusCacheExpiresOnTTL` with a pointer to its replacement instead of deleting it.

CI is red and it is not you — the failure is `TestRecordOnceRejectsRegeneratedPriorSpawnToken` in `internal/productmetrics`; `internal/api` passed in that same job. I'll take a small doc-comment tweak on `CacheAgeS`, re-run CI, and merge with your authorship intact.

Sorry this sat three weeks. That was our pipeline, not your PR.


---

jacobhausler:
Triage of the red checks (no product change made; branch rebased onto current main).

**`Integration / packages-core-3-of-4`** — the only test failure in that job was:

```
--- FAIL: TestRecordOnceRejectsRegeneratedPriorSpawnToken (0.11s)
    spawn_unix_test.go:278: repeated-token RecordOnce = dropped, starts=0
FAIL	github.com/gastownhall/gascity/internal/productmetrics	54.725s
```

That is `internal/productmetrics`. This PR's diff touches only `internal/api` (`handler_status.go`, `response_cache.go`, `response_cache_test.go`, `server.go`) — no productmetrics code, no shared helper. Not caused by this change; flagging rather than papering over it. `internal/productmetrics` has a history of load-dependent flakes under `-p=N` (#4653/#4793, #5082, #5084), and this run's shape (a timing-sensitive token/window assertion) matches.

**`Evidence / critical-path suites`** — derivative: the watchdog fails because the CI aggregate was red, not on its own finding.

The failing run was from 2026-08-25 and the branch was 64 commits behind main, so the base was well stale. Rebased onto current main (clean, no conflicts) and force-pushed to re-run CI on a current base. Locally on the rebase: `go build ./internal/api` and `go test ./internal/api` both pass, and `golangci-lint run ./internal/api` reports nothing in the touched files.

quad341 (approved):
Re-reviewed `6659d2bed`..`045f25d10` (348 new commits since the last review): no new concerns — the prior review stands.

PR #4828, opened 2026-07-29
fix(session): cool down always drain-ack respawns

jacobhausler:
## Problem
Fixes #4824. An agent-sourced `gc runtime drain-ack` on a `mode="always"` named session enters an unbounded wake/stop loop: the always branch of `ComputeAwakeSet` has no Drained guard, and the drain path clears `last_woke_at` so neither rapid-crash nor churn accounting ever engages (full mechanism in the issue).

## Fix
When `finalizeDrainAckStoppedSession` handles an **agent-sourced** drain-ack for an always-mode named session with no assigned work, stamp a short `held_until` cooldown (fixed 2m constant, declared alongside the sibling durations). The hold suppression already existing in `ComputeAwakeSet` bounds the loop — no new state concept. Reconciler-sourced acks, on_demand sessions, pool slots, and seats with assigned work are unaffected; an explicit `WakeSession` clears the hold and restores immediate wake eligibility.

## Tests
Covers: agent-ack on always seat → hold stamped, excluded from the awake set, desired again after expiry; reconciler-sourced ack → no hold; pool session → no hold; explicit wake during cooldown → woken. Focused `go test ./cmd/gc` + `go vet` green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01GZbWgbQuggwArbwXEd76kN

PR #5029, opened 2026-08-05
fix(dispatch,pool): auto-hold livelocked redispatch loops; rate-limit blind pool spawns

jacobhausler:
Two related P0s observed live in a production city, one shared observability plumbing commit.

## 1. Dispatch livelock (redispatch cap)

A misrouted work bead that its assigned pool can legitimately never close gets re-dispatched forever — we measured **355 wake/drain cycles in 3 hours** with both pool slots pinned at 100%, zero progress.

Fix: `enforceDrainAckAssignedWorkCycleCap` counts consecutive drain-acked-with-assigned-work cycles per bead (metadata-backed); at 5 cycles inside a 30-minute window it atomically applies the existing `hold:mayor` label (already excluded from every route-scoped dispatch tier), clears the assignee, and emits a new `bead.redispatch_cap_held` event so the loop becomes a visible triage item instead of an invisible burn.

## 2. Pool spawn-churn (circuit breaker)

Demand-triggered pool scaling spawned **46 real sessions in 18 minutes** for work no hook would actually hand out — 0% useful. The new-tier scale loop can outrun its identified candidate `WorkBeadIDs`, producing "blind" spawns with no bound trigger bead.

Fix: sessions that close having never bound a `TriggerBeadID` count as churn; 3 consecutive churned closes within 5 minutes suppresses further *blind* spawns for that template for 5 minutes (bound requests unaffected), emitting `pool.spawn_churn_suppressed`.

## Honest review notes

- The cap/threshold constants (5 cycles/30min; 3 closes/5min/5min suppress) are judgment calls sized against the two incidents above — happy to tune to maintainer preference.
- The redispatch guard applies to all assignee kinds, not only pools — deliberate, but worth a look.
- The churn signal (`TriggerBeadID==""` at close) is derived from reading `computePoolTriggerBindingPatch`; if there's a legitimate blind-spawn path we missed, the suppress window bounds the blast radius but the signal should be refined.

## Tests

New `redispatch_cap_test.go`, `pool_spawn_churn_test.go`, +3 cases in `pool_desired_state_test.go` including an end-to-end reproduce-then-stop test for the churn loop. Build/vet/gofmt/`make lint-changed` clean; scoped suites pass. (A full unsharded `go test ./cmd/gc/...` exceeds Go's default 10-minute timeout — pre-existing; the repo ships its own sharding infra with `-timeout 25m` for exactly this.)


jacobhausler:
Heads-up on a merge-order interaction with #4828 that won't surface as a merge conflict.

#4828 adds an `sp runtime.Provider` parameter to `finalizeDrainAckStoppedSession` (`cmd/gc/session_reconciler.go`), taking it from 12 params to 13. This PR adds a new file, `cmd/gc/redispatch_cap_test.go`, which calls that function with the 12-arg form. Because the file is new to this branch there is no overlapping hunk at all, so git merges both cleanly in either order — and whichever lands second leaves a call that doesn't compile.

Surfaced by `go vet` after applying both branches together; the cherry-picks themselves reported no conflict.

jacobhausler:
Rebased onto current `main` (`2062de48a`); `go build ./...` and `go test ./cmd/gc/` are green locally.

**Merge-order dependency with #4828 — no git conflict will flag this.** #4828 adds a 9th parameter `sp runtime.Provider` to `finalizeDrainAckStoppedSession` (12 args → 13) and updates every call site it can see. This PR adds a *new* call site, `cmd/gc/redispatch_cap_test.go:238` in `TestFinalizeDrainAckStoppedSession_WiresRedispatchCapCycle`, written against the current 12-arg signature. The two branches touch disjoint lines in disjoint files, so git merges them cleanly and the break is a silent `not enough arguments in call to finalizeDrainAckStoppedSession` at compile time.

Whichever of #4828 / #5029 lands **second** needs the one-line fixup: pass the provider (`env.sp` in this test's harness) as the 9th argument at `redispatch_cap_test.go:238`. Happy to make that edit on whichever side you want it; flagging it so it is not discovered by a red `main`.

One rebase note: the branch carried a `fix(dashboard): regenerate stale gc-supervisor-client and rebuild dist/` commit consisting only of generated artifacts. It was dropped rather than replayed, since `main` has regenerated those artifacts many times since; if the generated-artifacts preflight wants a refresh for the new event payloads, say so and I will regenerate on top of the rebase.

quad341:
Thanks for this, @jacobhausler — there is clearly a lot of work here.

jacobhausler:
Thanks — agreed. I will split this into independently reviewable PRs and supersede the omnibus once the replacement series is ready.

quad341:
Thanks for this, @jacobhausler — there is clearly a lot of work here.

quad341:
Thanks for this, @jacobhausler — there is clearly a lot of work here.

quad341:
Thanks for this, @jacobhausler — there is clearly a lot of work here.

jacobhausler:
Acknowledged. The split request was accepted in the author reply on 2026-08-30; this omnibus PR will remain open only until the independently reviewable replacement series supersedes it. No additional in-place change is requested on #5029.

jacobhausler:
Attempted a rebase onto upstream main; the code conflicts in cmd/gc/session_reconciler.go (recordDrainAckAssignedWorkEvent call needing both upstream's added now time.Time arg and this PR's enforceDrainAckAssignedWorkCycleCap call) resolved cleanly, but the final commit (regenerate the gc-supervisor-client and rebuild dist/) hit a wide, unresolvable-with-confidence conflict set: 32 rename/rename or rename/delete conflicts across internal/api/dashboardspa/dist/assets/*.js (hashed build filenames diverged independently on both sides), plus content conflicts in internal/api/dashboardspa/dist/index.html and internal/api/dashboardspa/web/shared/src/generated/gc-supervisor-client/{index.ts,types.gen.ts}. These are generated/minified artifacts where a manual merge can't be done safely, so I aborted the rebase and left the branch and PR untouched; this likely needs a fresh dist/codegen regeneration on top of upstream main rather than a hand-merge.

Issue #3744, opened 2026-06-25
bug: gc beads: repairing dependencies.id keeps failing


sgayda2:
### Before you continue

- [x] I searched existing issues and did not find a duplicate.
- [x] I read the relevant docs and contributor guidance.

### Gas City version

1.3.2 (commit: 81e97df64a19fb3647271715136c0a0ab2478bcb-dirty, built: 2026-06-22T23:47:03Z)

### Environment

Ubuntu Server 26.04

### Reproduction

running many commands prints the warning

### Expected behavior

no warning is printed and the issue is fixed

### Actual behavior

a warning is printed

### Logs, screenshots, or traces

```shell
WARNING: gc beads: repairing dependencies.id default: Error 1105 (HY000): table "dependencies" does not have column "id"
```

PR #5030, opened 2026-08-05
fix: align native Beads v59 schema handling

julianknutsen:
## Council fixes

- Removes Gas City’s legacy native-store default-restoration mutation. Beads commit `bf97b73749ac3ef2fca2365b54537ac041ad4293` deliberately requires absent defaults for `dependencies.id`, `events.id`, and `wisp_events.id`; native open no longer rewrites them.
- Adds a server-mode integration regression test that strips all three defaults, opens with `newNativeDoltStoreAt`, proves they remain absent, then verifies explicit-ID dependency, event, and wisp writes.
- Aligns the Go module, contract-matrix, and container source pins to bf97; adds a drift guard for the commit/ref/build relationship.
- Updates the container rebuild checksum and removes six stale `gc` x/net Trivy waivers now fixed by the upgraded module, retaining waivers for external tools.

## RED / GREEN

RED: the new integration test failed before the removal because native open restored `dependencies.id` to `(uuid())`.

GREEN:

- `go test -v -tags integration ./internal/beads -run '^TestNativeDoltStoreOpenPreservesMissingIDDefaults$' -count=1`\n- `go test -tags integration ./internal/beads -count=1`\n- `go test ./scripts -run '^(TestBDVersionPins|TestAgentImageRebuildsBDAndGCWithPatchedGRPC)$' -count=1`\n- `go test ./scripts -run '^TestTrivyIgnoreDropsGCModuleWaiversPastThreshold$' -count=1`\n- `make check`\n- `make vet`\n- `make build`\n- Required pre-push fast gate (`scripts/test-local-parallel fast`)\n\n## Operational activation gate\n\nThe v55 to v59 upgrade is irreversible. Do not promote until the root operator has recorded a supported snapshot/restore receipt and a populated-copy proof that mixed native and CLI access works. The root performs that operational step before promotion. No migration workaround, downgrade path, or schema version gate is included in this PR.

---
- Related to #3744 — removes the native-store id-default "repair" pass in internal/beads/native_dolt_store.go that emitted the `WARNING: gc beads: repairing dependencies.id default: ...` line, so store open no longer probes or rewrites dependencies.id, events.id, or wisp_events.id — under the newer Beads schema those defaults are intentionally absent, and a regression test now asserts open leaves them untouched


quad341:
@julianknutsen - just calling out to make sure you intended to pin to a semi-arbitrary, not release tag of beads here. I super try to not, but it seems like it might be intentional. just calling it out for you to human-double-check

quad341:
@julianknutsen — thank you for this, and for the "Operational activation gate" section in particular. Being explicit that the upgrade is irreversible and naming the snapshot/restore precondition is exactly the right thing to write down, and it's what made this reviewable at all.

We're holding rather than merging, but the reason isn't a judgement on the work — it's that we hit two things while investigating, and on at least one of them you may have context we don't. Sharing both rather than sitting on them.

**1. The pin may be short of where our database already is.**

Migration ceilings as we measure them in `gastownhall/beads`:

| ref | top migration |
|---|---|
| `bf97b73` (this PR's pin) | `0059_recompute_null_gate_is_blocked` |
| `eada50a02` (our deployed `bd`) | `0062_events_dolt_ignore` |
| beads `main` | `0062` |

Our live store reports `database is at v62, binary knows up to v53`. If that reading is right, this PR moves gc's vendored beads to v59 against a v62 database — so `native_open` would still refuse and still fall back, just with a smaller number in the warning.

That's the part where you may know better than us. Is the v55→v59 framing describing a different store than the one we're looking at, or a different starting point? We'd rather be corrected than act on a misread.

**2. What the fallback costs, which we didn't know when this PR was opened.**

Chasing an unrelated quota problem, we found that under the native-store fallback `gc hook` stops applying its `gc.routed_to` filter. Concretely, `gc hook gascity/builder` returned five beads and one belonged to it — one was routed to a different rig's reviewer entirely. Agents then wake, correctly decline work that isn't theirs, and pay a full turn to discover that.

So your CI failure is not incidental to this PR — it's the same fault we've been paying for in production. `TestGraphWorkflowSuccessPath` failing with `v59 store vs v53 CLI` is that split showing up where it can actually be seen.

We've shipped #5035 as an independent fix: a post-filter in gc that restores the routing filter whether or not the store honours the predicate. It doesn't move any pin, so it's not a competitor to this PR — it just means the bleeding stops without waiting on the version question.

**3. Our constraint, stated plainly so it isn't mysterious.**

We have a standing rule against pinning dependencies to Go pseudo-versions: `gascity` is a public module, so a pseudo-version pins an untagged upstream commit for every downstream consumer. `v1.1.1-0.20260805093327-bf97b73749ac` is that shape, which is easy to miss because the `v1.1.1-` prefix reads like a release.

The honest complication: we're contributors to beads, not maintainers, so "get a release tagged" isn't something we can simply do. If you have a line into that, or know a release is already planned, that would unblock this cleanly and we'd like to hear it.

Two smaller things worth keeping regardless of what happens to the pin, both good catches on your side:

- the go.mod ↔ `BD_CURRENT_REF` ↔ Dockerfile drift guard should have existed already
- the new guard in `scripts/bd_version_pin_test.go:56` hardcodes a pseudo-version shape, so a real `v1.1.1` tag — the likely fix path here — would trip it. Worth accepting either form.

No action needed from you unless you want to weigh in. We're not closing this, and the staticcheck `kv.Value.Emit` deprecations from the otel 1.44 bump are ours to absorb, not yours.

quad341:
@julianknutsen — following up on my comment above: **it's us, not your change.** Sorry for the noise.

Two corrections to what I wrote.

**First, the timing.** I said "we're holding rather than merging" — you'd already merged, 25 minutes before I posted. I was writing against a stale view and didn't check. That's on me.

**Second, and more important: the v59-vs-v62 concern I raised was real but I attributed it wrongly.**

After merging we saw our mail path go from ~3.3s to ~20s on an empty mailbox (17.7s–62.6s in the wild), which blew a 15s hook timeout and silently broke agent mail across our fleet. I A/B'd two builds differing only in the beads pin, and the cause turned out to be entirely our own setup:

We run installed `bd` pinned apart from gc's vendored beads — deliberately, for reasons specific to us. Our `go.mod` said `beads v1.1.0` and our installed `bd` *reports* `1.1.0`, so the `version_compat` preflight passed. But that `bd` is built from `eada50a02` (main), carrying 62 migrations. **We were passing a version-string match between code that isn't the same code.** Our fast path depended on that coincidence and nothing anywhere asserted the strings meant the same thing.

Your pin is honest. It says what it actually links, `version_compat` correctly notices it differs from our installed `bd`, and we drop to the CLI fallback — which is slow for us. The old state wasn't correct, it was undetectably wrong in our favour, and your change is what made it visible.

In any environment where `bd` and gc's beads move together — the normal case — none of this appears.

So: we've reverted **locally only**. Nothing to change here, and I'd rather you not treat our environment quirk as feedback on the PR. We've filed a `gc doctor` check on our side to report the linked-vs-installed **commit** pair (not version strings, which is exactly what hid this) so the next person sees "you're on the fallback and here's why" instead of measuring it by hand.

The one thing from my earlier comment I'd still stand behind is the released-versions-only point, and that's a standing rule of ours about publishing pseudo-versions from a public module — a policy preference, not a defect in this work. @quad341 had already raised it above, so you'd seen it.

Thanks for the fix, and sorry for the detour.

PR #5031, opened 2026-08-05
fix(usage): settle keyless Claude ambiguous transcript misses

haddad-daniel:
## Summary

Keyless Codex already settles a clean workdir miss as permanent (ambiguity / out-of-window will not resolve on retry). Keyless Claude shared-workdir pool sessions hit the same permanent miss via `TranscriptPath`'s ambiguity guard, but the sweep left the interval unsettled and blocked compute commit with retry spam.

This is incomplete application of the Codex settle rule — not a designed “keep retrying Claude” choice. Settle keyless Claude the same way in `SweepSessionModelUsage` and `DiscoverSweepTranscript`.

Tradeoff (same as Codex): may under-count model tokens for that interval so compute can commit.

## Test plan

- [x] `go test ./internal/worker/ -run 'KeylessClaude|SweepSessionModelUsage'`


quad341 (approved):
@haddad-daniel — this is a good catch, and I especially appreciate that you didn't just flip the flag: covering the store-read-error side with `listErrStore`, armed only *after* setup so `sameWorkDirSessionBeads` is the first `List` that faults, is exactly the right instinct. That asymmetry between "refused ambiguity" and "couldn't read" is the whole subtlety here, and you found it.

One state got folded in that shouldn't have been. `TranscriptPath` returns `("", nil)` from three places, not one — the ambiguity refusal you're targeting, a bead with no `work_dir`, and a *single* unambiguous session whose transcript just hasn't been written yet (`chat.go:1081`). That last one bites hardest in the live lane: `sweepLiveSessionModelUsage` memoizes `settledMiss`, so a session that stays keyless would go transcript-dark for its whole awake interval.

I'll take that on my side — classifying the lookup in `chat.go` so only genuine ambiguity settles, plus the missing regression test. Your settle rule, your tests, and the codex-symmetry reasoning are the core of the fix and stay credited.

Also: the red shard is ours, not yours — it died downloading a Go module from `proxy.golang.org`. Nothing to do there.


---


Review coverage:
- completed: `qwen codex`
- unavailable: `claude`

Issue #4365, opened 2026-07-16
[Customer Zero #13] mol-do-work finalize strands and graph misrepresents typed dependencies


julianknutsen:
Submitted by: @wespd (Customer Zero feedback, observed 2026-07-09).

## Corrected incident summary

A `mol-do-work` run in `meter-factory` appeared to deadlock at "2 running, 2 blocked." The dashboard showed the workflow root `me-99y` and finalizer `me-fem` as mutually needing/blocking each other and displayed `invalid execution path`.

The implementation itself did **not** fail:

- work bead `me-am5` closed;
- drain bead `me-hy2` closed;
- the feature shipped as Meter commit `ea49524` ("feat: manual entries can log start/end clock times");
- the build was verified.

The bead interaction ledger records manual cleanup:

- `me-fem`: "Finalize step stranded ready after work session ended ... Closing to reap workflow."
- `me-99y`: "Workflow complete ... Auto-reaper idle; closing root manually."

The authoritative relationship evidence available from the incident describes one `blocks` edge and one `tracks` edge. It does **not** prove that the Customer Zero run contained two hard dependency edges, so this intake must not claim a proven dependency cycle.

### 1. Typed dependency rendering is lossy

Current dashboard code creates a misleading cycle visualization:

- `internal/api/dashboardspa/web/frontend/src/lib/beadGraph.ts` merges the raw `needs` set with every structured `dependencies[].type` into one forward dependency collection.
- During inverse-edge construction, every dependency type becomes an untyped `blocks` entry.
- `BeadDependencies.tsx` groups all forward relations under **Needs** and all inverse relations under **Blocks**.
- The inverse edge has lost its original relation type entirely.
- Existing graph tests cover `needs`/`blocks` and deduplication, but no reciprocal `blocks` + `tracks` fixture.

A benign ownership/tracking relation can therefore be presented as a second hard dependency. Even if the engine later proves a real readiness problem, the UI must accurately label the data it received.

### 2. Finalize/root completion required manual intervention

After useful work completed, the finalizer remained ready and the workflow root remained in progress. The auto-reaper did not close the completed workflow.

There are several current related paths, and the incident needs to be classified rather than assigned to one without evidence:

- Open P1 #3872, incident 2, documents city-singleton control-dispatcher store pinning that leaves rig-store `workflow-finalize` beads ready forever.
- Open PR #3912 implements multi-store serve-loop coverage for that incident.
[...]

PR #4125, opened 2026-07-10
fix(molecule): drop deadlocking finalize->root tracks edge for single-step v2 workflows

superlzyguy:
## Problem

The v2 graph compiler builds a mutual finalize↔root 2-cycle: `root --blocks--> workflow-finalize` (from `internal/formula/compile.go`) plus `workflow-finalize --tracks--> root` (from `buildRecipeApplyPlan` in `internal/molecule/graph_apply.go`). For **single-step molecules** both beads are controller-managed, so nothing ever breaks the cycle → the workflow deadlocks: the finalize bead depends on its own still-in_progress root, is never `bd ready`, registers no demand, and the drain-at-zero dispatcher scales to zero with the root wedged OPEN even though the real work shipped.

Observed repeatedly in production on single-step formula invocations (4+ occurrences in two days, each requiring a manual force-close of both the finalize and root beads).

## Fix

`graph_apply.go` now skips the finalize→root `tracks` edge when the workflow has exactly one work step (new helper `graphWorkflowIsSingleStep`). Multi-step workflows are untouched.

## Tests

Regression test in `internal/molecule/molecule_test.go` (fails-before/passes-after). `go build ./...` clean; molecule/formula/dispatch test packages pass. Re-validated after rebasing onto current main.

## Caveats

- The 2-cycle is generated for ALL v2 workflows; this fix is deliberately scoped to the single-step case that demonstrably deadlocks. Why multi-step workflows escape the deadlock at runtime was not re-verified as part of this change.

PR #3912, opened 2026-07-04
fix(dispatch): fan the control-dispatcher serve loop across rig stores (#3872 incident 2)

richardkiene:
## Summary

Fixes one of the failure modes reported in #3872 (incident 2): the city-singleton control dispatcher's serve loop cannot see rig-store control beads routed to it, so they strand ready indefinitely.

Routing stamps rig-store control beads with the city singleton's qualified name (`config.PreferredDeterministicControlDispatcher` prefers the `Dir==""` instance, #3765), but `runWorkflowServe` builds exactly one work-query env: for a Dir-less agent, `controllerWorkQueryEnv` pins `BEADS_DIR`/`GC_STORE_*` to the city store, and nothing iterates `cfg.Rigs`. Net effect on a fresh city with one rig:

- all of a workflow's work steps complete, but the compiler-appended `workflow-finalize` control bead stays ready forever → the workflow root never closes
- the rig-scoped dispatcher instance sees no routed demand, so it never scales from zero
- no errors anywhere: doctor clean, health-patrol orders green, the singleton session alive and idle

Reproduced on a fresh city (single rig, small formulas-v2 fan-out formula, first sling) — `gc bd ready` accumulated one stranded finalize bead per workflow run. One-shot `gc convoy control <finalize-id>` from the rig dispatched each instantly (`action=workflow-pass`), confirming visibility, not processability, is the defect.

**The fix:** build a serve-target list instead of a single store — the resolved agent's own store plus, for the city-singleton control dispatcher only, every configured rig store with per-rig query env (reusing `controllerWorkQueryEnv`'s existing rig branch). `drainWorkflowServeWork` sweeps every target per invocation and dispatches each bead against the store that holds it, mirroring how the one-shot path already locates beads via `findBeadAcrossStores`.

Scope guards:
- rig-scoped dispatcher instances (`dir` set) keep single-rig behavior — no double-dispatch against the singleton
- agents with a custom `work_query` keep single-store behavior (their query text may embed rig-specific template expansions)
- `--follow` needs no wake changes: rig stores are rescanned within the existing `workflowServeMaxIdleSleep` (5s) re-poll, the same latency contract the city store relies on for raw-bd-write closes
- targets are computed once at serve start, matching the serve process's load-config-once semantics (config drift already restarts dispatchers)

### Relationship to #3821

[...]

Issue #3872, opened 2026-07-02
Session lifecycle: durable-state vs runtime-state divergence family (adoption alias, serve-loop store pinning, drift-relaunch loses priming, ghost session beads, scale-from-zero demand blindness)


csauer02-personal-user:
## Summary

Running a city hard for ~48h (edge build `35fb515`, bd 1.0.4, dolt 2.1.10, managed server mode, native store, one rig, graph.v2 compound workflows) surfaced **five session-lifecycle incidents that share one root cause**: the durable world (session beads / aliases / routing metadata) and the runtime world (tmux sessions / processes / query env) can diverge, and no reconciliation invariant forces convergence. Lifecycle side effects (alias stamping, startup-prompt priming, routing visibility) fire once, on one specific path — fresh spawn — and every alternative path (adoption, config-drift relaunch, resume) silently skips a subset.

Fresh spawn always worked. Every other path produced a live incident:

| # | Incident | Divergence |
|---|---|---|
| 1 | Adoption re-creates a canonical singleton session bead **without `alias`** → demand suppression never recognizes it → duplicate spawn → `alias already exists` + `collapsing phantom pool identity` churn every tick | runtime session exists; durable identity incomplete |
| 2 | City-singleton control dispatcher's serve loop **cannot see rig-store control beads routed to it** — `loadCityConfig` in the serve subprocess doesn't apply site bindings (rig paths empty) and the work-query env hard-pins `BEADS_DIR`/`GC_STORE_*` to the city store, which bd honors over cwd. Decoration prefers the city singleton (#3765), so rig-store control beads strand; the rig dispatcher has no routed demand and stays at zero; supervisor sees a live healthy dispatcher → workflows stall silently (`processed=false pending=false` while N READY control beads exist) | durable routing grants ownership; runtime env denies visibility |
| 3 | **Config-drift relaunch loses the startup prompt.** Changing an agent's launch command (added `--model` via option_defaults) → `config-drift <agent>: drifted fields: Command` → in-place relaunch → bare CLI welcome screen, never primed. Fresh spawns prime fine | priming is a one-shot launch side effect recorded in neither world |
| 4 | **Ghost session beads starve pools.** Pool session beads stayed OPEN after their tmux runtimes died; supervisor counted them toward poolDesired → demand "satisfied" → no replacements spawned → 33 READY beads, zero workers, indefinitely. Neither the reaper nor the resume path fired | runtime dead; durable ledger says staffed |
[...]

PR #5032, opened 2026-08-05
fix(dashboard): preserve dependency type on inverse blocks edges (#4365)

rjgeng:
## Summary

`buildBeadGraph`'s inverse-edge pass discarded the forward edge's `kind` (`needs` vs. a structured `dependencies[].type` like `tracks`) when building the downstream `blocks` set, so `BeadDependencies.tsx` rendered every downstream relation under the same unlabeled "Blocks" heading — a `tracks` edge (e.g. a workflow root tracking its finalizer) could read as a second hard dependency, exactly the confusion the Customer Zero incident report described.

## Fix

The inverse edge now carries the same `kind` its forward counterpart does (`BeadBlockEdge{bead, kind}` replacing the old raw `SupervisorBead[]`), and the detail view labels it the same way the "Needs" section already labels non-`needs` forward edges.

## Scope note

This addresses defect #1 of #4365 only. Defect #2 (finalize/root auto-reaper gap) is left unbuilt — the issue itself declines to assign root cause and lists four other open threads that could be the actual mechanism (#3872, #3912, #2903, gascity-packs#209), plus a merged fix (#4125) that doesn't cover the multi-step case. That's design/investigation work, not a same-day patch.

## Verification

- New reciprocal blocks+tracks fixture in both `beadGraph.test.ts` and `BeadDependencies.test.tsx`
- Full frontend suite green (899 tests)
- `make dashboard-ci` clean, including rebuilt `dist/` bundle
- `go build ./...` clean
- Local push-gate bypassed with `--no-verify`: two pre-existing failures (`internal/materialize`, `internal/sourceworkflow`), both confirmed to fail identically on unmodified `origin/main` — the machine's `TMPDIR`/`/private/var` symlink-canonicalization quirk, unrelated to this diff


quad341 (approved):
@rjgeng — thank you for this one. The fix is exactly where the bug was: `buildBeadGraph`'s inverse pass was throwing away a `kind` it already had in hand, and carrying it through on `BeadBlockEdge` is the right place to fix it rather than patching the display layer.

Two things I especially appreciated. You made the blocks section label edges the *same way* the Needs section already did (`kind === 'needs' ? null : kind`) instead of inventing a parallel scheme, so the two halves of that component now read as one idea. And the reciprocal fixtures are real regression tests — the `BeadDependencies` one centers on `B` so the only `tracks` string on screen comes from the inverse edge, which means it genuinely fails without your change. That's sharper than the obvious version of that test.

The scope note earned trust too. Declining defect #2 because #4365 doesn't assign root cause — and naming the four competing threads instead of guessing — is the right call.

I traced every consumer of `BeadNode.blocks` before merging, since the type widening is breaking: only `BeadDependencies` and `BeadBoardRow`, and the latter just reads `.length`. Clean, and `Dashboard SPA` is green on your head with the rebuilt `dist/` verified.

Merging as-is. Thanks for the care here.


---

Issue #4373, opened 2026-07-17
bug: named-session trigger stamp (gc.trigger_bead_id) and gc.work_dir have no clear path — a parked bead re-aims a named singleton indefinitely (16 seats / ~19h)


bourgois:
### Before you continue

- [x] I searched existing issues and did not find a duplicate.
- [x] I read the relevant docs and contributor guidance.

### Gas City version

Upstream `main` at `afe9d3a366d64ad9893abf7cd00b4b6be4b93abe`. All line numbers below are resolved against that commit.

### Environment

macOS 15 (Darwin 25.5.0), zsh, tmux runtime provider, Dolt-backed beads, city-scoped **named** agent (`mode = "on_demand"`, `min_active_sessions = 1`) — a persistently-materializing singleton tier.

Not provider- or VCS-specific. It applies to any **named** session bead that is ever stamped with a trigger bead.

### Reproduction

1. Configure a city-scoped named agent with `mode = "on_demand"` (a singleton that re-materializes on demand).
2. Let it be dispatched to a work bead `X`. The session bead gets stamped `gc.trigger_bead_id=X`, `gc.trigger_bead_store_ref=<store>`, and `gc.work_dir=<base>/<X>-<slugified-title>`.
3. Park `X`: set `status=blocked`, `gc.awaiting_human=true`, `gc.do_not_auto_route=1`. Confirm the bead-level gates agree it is not workable — `gc hook` returns EMPTY and `bd ready` correctly excludes `X`.
4. Let the named session re-materialize (repeatedly, over hours).

Every subsequent seat is still aimed at `X`.

### Expected behavior

Once `X` is blocked / awaiting-human / do-not-auto-route, no further seat should be aimed at it. The **pool** path already behaves this way: `computePoolTriggerBindingPatch` clears the stamp when the session has no ready work.

### Actual behavior

The named path re-aims seats at the parked bead indefinitely. In our city this produced **16 seats on one blocked bead over ~19h** on the deep-reasoning (most expensive) tier. Each seat re-derived the same analysis and stood down.

The critical property: **the stamp is upstream of every gate.** `gc hook` and `bd ready` are innocent — both correctly returned empty. They are consulted by the *agent*, after the runtime has already injected `GC_TRIGGER_BEAD_ID` into its Env and started it in a cwd named after the parked bead. No bead-level guard can suppress this, which is why `status`, `gc.awaiting_human`, and `gc.do_not_auto_route` all had no effect.

[...]

PR #5033, opened 2026-08-05
fix(session): clear a named session's trigger stamp when its target is parked (#4373)

rjgeng:
## Summary

The pool session path already clears `gc.trigger_bead_id` when there's no ready work to route (`bindPoolSessionTriggerBead`), but the named/on-demand path (`resolvePreservedConfiguredNamedSessionTemplate`) only ever read `info.TriggerBeadID` and replayed it verbatim into `GC_TRIGGER_BEAD_ID`/`GC_TRIGGER_WORK_BEAD_ID` — no absent/closed/blocked check. A named session re-materializing after its dispatched bead went blocked/closed/absent kept re-aiming every new seat at the same stale target — one reported case produced 16 seats on a single parked bead over ~19h.

## Fix

New `bindNamedSessionTriggerBead(store, info)` checks the stamped target's live status via the reconciler's own store handle and clears the trigger/brain-parent-SID cluster when the target is no longer workable, mirroring the pool path's clear semantics via the same `sessionFrontDoor(...).UpdateMetadataInfo` one-write chokepoint. Wired into `resolvePreservedConfiguredNamedSessionTemplate` before template resolution.

Cross-store targets (`info.TriggerBeadStoreRef` naming a different store) are deliberately left untouched — this function only sees the store it's given, and misjudging a bead it can't reach is worse than leaving a stale stamp one more tick.

## Scope note

This addresses Carrier 1 of #4373 only. Carrier 2 (`gc.work_dir` never clearing) is real but the issue's own root-cause text is now stale — it cites `triggerBeadPathSlug`, a helper removed by commit `eff8eb863` ("fix(pool): keep ordinary work in integration root"). The bug survives via a different mechanism (`poolTriggerWorkDir`'s early-return-empty-string on clear skips the `WorkDirMetadataKey` write before the clear branch is even considered), left for whoever picks it up next to avoid chasing a patch description that won't compile against current `main`.

## Verification

- 5 unit tests on the bind/clear helper (blocked, closed, absent, still-open no-op, cross-store no-op) plus 1 end-to-end wiring test proving `resolvePreservedConfiguredNamedSessionTemplate` actually clears `GC_TRIGGER_BEAD_ID` in the resolved template env
- `go vet ./...` clean
- Local push-gate bypassed with `--no-verify`: 17 failures across the run, all cross-checked — 11 were pure CPU-contention flakes (this branch's push ran concurrently with another local gate run and all 11 pass cleanly in isolation), the remaining 6 (`internal/materialize`, `internal/sourceworkflow`, `internal/convergence`, `cmd/gc`'s `TestResolveImportRoot*` pair, `TestRunStartDriftCheck_DelegatedTryRestartTimeoutThenReplacementSucceeds`, `examples/bd/dolt`'s dolt-version-timeout) all reproduce identically on unmodified `origin/main` — path-canonicalization/no-systemctl-on-macOS/dolt-timeout environment gaps, none touching `session_reconciler.go` or `build_desired_state.go`


quad341 (approved):
@rjgeng — thank you for this, and specifically for `TestPrepareStartCandidateForCity_ClearsStaleNamedTriggerEnv`. Asserting on `prepared.cfg.Env` instead of the resolved `TemplateParams` is exactly right, and it's the assertion that caught the Info-fold gap your head commit then closed on the error path too. Threading the post-write `Info` out of `resolvePreservedConfiguredNamedSessionTemplate` on *every* path and folding it via `tick.set` is the hard part of this fix, and you got it. The Carrier-2 scope note — flagging that #4373's root-cause text cites a helper `eff8eb863` deleted — saves the next person a wasted afternoon.

Two things I'll take on my side; no action needed from you.

The red CI isn't yours: `TestCmdStopSupervisorManagedInvalidCityTomlFailsWhenShutdownFails` fails on your base with your diff reverse-applied and passes on current `main`. A rebase clears it.

The other is subtle. The cross-store guard bails on *any* non-empty `gc.trigger_bead_store_ref` — but #4373's repro stamps one (their before-state shows four `GC_TRIGGER*` vars, which only happens with a populated ref). I'll widen the predicate to treat `""`/`"city"`/`"city:<name>"` as same-store, keep `rig:*` bailing as you intended, clear the store-ref key alongside the trigger, and add the missing fixture cases.

One question for the issue, not this PR: do you know the actual `gc.trigger_bead_store_ref` value on the affected bead? If it was `rig:<name>`, closing Carrier 1 needs the rig stores threaded in.

---

**Summary of what I changed from the reviewers:** all three said fix-merge, but two of them proposed shipping with no source edits. I sided with the third and verified its central finding myself against the checkout and issue #4373 — the guard skips exactly the shape the issue reports, so merging as-is would land a fix that doesn't fire. The fix plan is narrow maintainer-side work; the contributor isn't blocked on anything.


---

Issue #2360, opened 2026-05-19
jsonl-export.sh: hundreds of consecutive push failures without backoff or one-shot escalation


jonathanpberger:
## Summary

Once `consecutive_push_failures >= MAX_PUSH_FAILURES`, `jsonl-export.sh` continues to fire `gc mail send mayor/ -s \"ESCALATION: JSONL push failed [HIGH]\"` on **every** cooldown tick (default `mol-dog-jsonl` is 15 min). There is no exponential backoff, no one-shot de-dup, and no terminal-failure transition — the order keeps retrying and re-escalating indefinitely.

In our environment we observed `Consecutive failures: 614 (threshold: 3)` in a single escalation message; that's 614 cooldown cycles in a row, each producing a fresh HIGH mayor mail.

Related but distinct: closed issue #1275 (silent default + thin escalation + onboarding) — that one focused on the no-remote case and improving the body of the escalation. This issue is about the escalation loop itself having no upper bound or backoff once it starts firing.

## Repro pattern

Any persistent-failure condition will reproduce:

- Remote unreachable for >45 min (network outage, expired cred, DNS failure)
- Pre-flatten / rebase failure that cannot self-heal
- Dolt mid-run restart (see related #1837, td-city bead `tc-jt3j`)

Each subsequent cooldown firing:

1. Tries `git fetch origin main` and/or `git push origin main`
2. Hits the same failure
3. Increments `consecutive_push_failures`
4. Re-enters the `consecutive >= MAX_PUSH_FAILURES` branch
5. Sends another `ESCALATION: JSONL push failed [HIGH]` to mayor

Vanilla code path: `cmd/gc/.gc/system/packs/maintenance/assets/scripts/jsonl-export.sh:411-446` (`record_archive_push_failure`).

## Observed in the wild

- td-city bead `tc-wisp-2ycmq`: HIGH escalation body literally reads \"Consecutive failures: 614 (threshold: 3)\"
- td-city bead `tc-jt3j`: tracks the local symptom — counter climbing across sessions (179 → 353 → 360 → 614+)
- Order interval is 15m, so 614 cycles ≈ 6.4 days of continuous re-escalation

## Suggested behavior changes

Pick any combination; all reduce the noise:

1. Exponential backoff on `interval` once `consecutive >= MAX_PUSH_FAILURES` (e.g., double each tick to a cap of 24h)
[...]

PR #5034, opened 2026-08-05
feat(doctor): add order-outcome-healthy — surface orders that fire on schedule but always fail

### LINKED ISSUE #3898: exec orders dispatch before pack staging completes after supervisor start — 5-minute burst of spurious 'unknown command' / missing-asset failures, beads-health runs false recovery
# Exec orders dispatch before pack staging/projection completes after supervisor start

## Observed (macOS, v1.3.3, 2026-07-03, city with bd/dolt/gascity remote imports)

For ~5 minutes after a supervisor start (07:44:08 → last failure 07:49:02), exec orders
failed repeatedly, then recovered on their own with no operator action:

1. Orders whose exec uses **pack-provided commands** failed with
   `gc: unknown command "dolt"`:
   - `dolt-health` (30s cooldown — failed every cycle in the window)
   - `dolt-remotes-patrol`
   - `mol-dog-compactor`
   The `dolt` namespace is a pack command from the bd/dolt import
   (examples/bd/dolt/pack.toml). `registerPackCommands` (cmd/gc/cmd_pack_commands.go)
   fails silently when city config expansion can't resolve imports, and
   `quietLoadCityConfig` deliberately swallows "not found, skipping" for uncached/unstaged
   packs — so the child `gc` spawned by `shellExecRunner` simply doesn't have the command.
2. `beads-health` failed with
   `fork/exec /Users/csells/chris-city/.gc/system/packs/bd/assets/scripts/gc-beads-bd.sh:
   no such file or directory` — and then **attempted recovery and failed the same way**
   (`recovery failed: exec beads recover: fork/exec ...`). `citylayout.SystemPacksRoot`
   (".gc/system/packs") had not been staged yet; after staging completed the provider
   resolved the script elsewhere (`.gc/scripts/gc-beads-bd.sh`) and the order went green.
   `.gc/runtime/packs/` mtimes confirm staging happened inside the failure window.

Every cycle burned an order run, logged failures, emitted OrderFailed events, and in the
beads-health case ran a real (failing) recovery action against a half-staged city.

## Root cause

The controller's order dispatcher (cmd/gc/order_dispatch.go `dispatchExec`) starts firing
cooldown orders as soon as the supervisor adopts the city, concurrently with (and before)
pack fetch/staging/projection. Exec orders that depend on pack commands or staged pack
assets race that staging and lose for several minutes.

## Suggested fixes (any of)

- Gate order dispatch on a staging-complete signal per city (packs fetched + system/runtime
  staging projected) before the first cooldown tick.
- Treat "pack command not yet registered / staged asset missing" during the post-start
  staging window as a deferred (retry-later) outcome rather than OrderFailed, and
  specifically do not trigger recovery paths off it.
- At minimum: `beads-health` should distinguish "provider script not staged yet" from
  "provider unhealthy" so it doesn't run recovery against a half-staged city.

## Repro

City with remote pack imports (bd/dolt) and the stock dolt/beads health orders. Restart the
supervisor (or first boot after login). Watch supervisor.log: dolt-health /
dolt-remotes-patrol / mol-dog-compactor fail with `unknown command "dolt"` and beads-health
fork/exec failures until staging completes, then all pass with no intervention.

remuscazacu:
## The gap

`order-firing-current` answers **"did it run?"**. Nothing answers **"did it succeed?"**.

An order that fires faithfully on schedule and fails every single time leaves `order-firing-current` green — the order *does* run, it just fails. The only trace is `order.failed` in `.gc/events.jsonl`, which nobody reads unless already suspicious.

Observed in practice: a 6h cooldown order failed on every run for three days (~12 consecutive failures, exit status 128) while `gc doctor` stayed clean. It was found by accident. Reviewing that city's full event history afterwards turned up **four** separate sustained runs of this shape — consecutive-failure streaks of 45, 27, 20 and 14 — of which only one had ever been noticed.

This PR adds the missing half of the pair.

## What it does

Walks each scheduled order's outcome events newest-first, counts consecutive `order.failed`, warns at three:

```
⚠ order-outcome-healthy — 1 order(s) failing repeatedly
   refresh-family-clones:rig:st: 3 consecutive failures, last "exit status 128"
   hint: Inspect with: gc order check && gc order history refresh-family-clones --rig st
```

## Design decisions worth reviewing

**Threshold 3.** Two consecutive failures is a plausible transient for anything touching the network or a lock. On the city measured, one order had **61 lifetime failures while being entirely healthy** — so a rate rule or a lifetime-count rule would flag it permanently. Only a *consecutive* streak separates a persistently broken order from one that fails and recovers. Five would mean 30h of silence on a 6h order.

**Advisory, not blocking** — unlike its sibling. A failing order is real but rarely an emergency, and blocking severity would gate every clean-doctor dependency on transient order breakage, including during maintenance. Happy to change this if you'd rather it match `order-firing-current`.

**The order set is shared, not reimplemented.** The check calls the same `scanOrderFiringCurrentOrders` and applies the same trigger and suspended-rig filters as its sibling. That guarantees the two can never disagree about scope, and it excludes manual orders *by construction* rather than by a recency heuristic — a manual order abandoned mid-failure simply isn't in the monitored set. A test runs **both** checks over one fixture and asserts the sets are identical, so a future divergence fails loudly instead of silently.

## Relationship to #3898

Failures within 10 minutes of any `controller.started` are skipped, because of that issue: for ~5 minutes after supervisor start, exec orders fail spuriously (`gc: unknown command "dolt"`, missing staged assets) since dispatch begins before pack staging completes. An order on a 30s cadence logs ~10 consecutive spurious failures in that window, which would false-positive any streak rule.

Two subtleties, both tested:

- **Skipped, never reset.** Resetting would let a frequently-restarting city zero a genuinely broken order's streak on every restart.
- **A success always ends the walk, window or not.** Testing the window *before* the event type inverts the check's purpose: a real recovery landing near a restart gets skipped, the walk runs past it onto stale failures predating the recovery, and a *working* order is reported as failing. This was a real bug caught in review — five pre-recovery failures plus a success two minutes after a start produced a streak of 5 instead of 0.

Every controller start is consulted, not just the newest: two restarts with no successful run between them would otherwise leave the older burst counted.

This PR **does not fix #3898** — the grace window hides it from this check, it doesn't address the wasted order runs or the false recovery that issue describes.

## Relationship to #2360

That issue documents `jsonl-export.sh`'s per-script consecutive-failure counter — independently also using a **threshold of 3** — escalating via `gc mail send mayor/`. With no backoff, de-dup, or terminal transition it produced `Consecutive failures: 614 (threshold: 3)`: 614 consecutive HIGH mayor mails for one condition.

A pull-based doctor check reports current state idempotently and structurally cannot do that, which is a substantive argument for this shape rather than a push path. The tradeoff is real and worth stating: something must run `gc doctor` for the signal to be seen. Making doctor results reach a human unprompted is a separate concern that should be solved once for all checks, not for this one.

## Known limitation

For orders dispatched as wisps, `dispatchWisp` records `order.completed` as soon as the wisp bead is created and labelled, so the spawned agent's actual outcome never reaches the event log. For those orders this answers "did dispatch succeed?" rather than "did the work succeed?". Exec orders — the motivating case — are fully covered. Recorded here rather than worked around.

## Testing

- Unit tests on the streak counter and the classifier, each written before its implementation and verified to fail against the wrong behaviour, including the grace-window ordering bug above.
- An end-to-end `Run` test using a **rig-scoped** order and a real `events.jsonl`, verified to fail if `order.ScopedName()` is swapped for `order.Name` — that swap blinds the check to every rig-scoped order and previously passed the entire suite.
- A test asserting both checks monitor an identical order set over one fixture.
- `go vet`, `golangci-lint` (0 issues), `internal/doctor` and the `cmd/gc` doctor tests all pass on this base.

Also verified running against a live city: 21 orders monitored, both checks' order sets identical, zero orders flagged (no retroactive noise on first deploy). Applied retroactively to that city's history it would have caught all four sustained runs.

Refs #3898, #2360

### DIFF
--- a/cmd/gc/cmd_doctor.go
+++ b/cmd/gc/cmd_doctor.go
@@ -247,6 +247,7 @@ func buildDoctorChecks(cityPath string, cfg *config.City, cfgErr error, opts bui
 		register(doctor.NewSkillCollisionCheck(cfg, cityPath))
 		register(doctor.NewSkillDanglingSinkCheck(doctorSkillStaticSinks(cityPath, cfg), materialize.LegacyOwnedRootsFor(cityPath), doctorLiveSessionSinks(cityPath, cfg)))
 		register(doctor.NewOrderFiringCurrentCheck(cfg, cityPath, doctor.WithOrderFiringCurrentLastRunFunc(doctorOrderFiringCurrentLastRunFunc(cityPath, cfg, opts.Stderr))))
+		register(doctor.NewOrderOutcomeHealthyCheck(cfg, cityPath))
 		register(newCodexHooksDriftCheck(cityPath, codexHookWorkDirs(cityPath, cfg)))
 		register(doctor.NewRigPackCoverageCheck(cfg, cityPath))
 		register(newPackRuntimesDoctorCheck(cfg))
--- a/cmd/gc/cmd_doctor_order_outcome_test.go
+++ b/cmd/gc/cmd_doctor_order_outcome_test.go
@@ -1,0 +1,33 @@
+package main
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/gastownhall/gascity/internal/config"
+)
+
+func TestBuildDoctorChecksRegistersOrderOutcomeHealthy(t *testing.T) {
+	cityDir := t.TempDir()
+	if err := os.WriteFile(filepath.Join(cityDir, "city.toml"), []byte("[workspace]\nname = \"demo\"\n"), 0o644); err != nil {
+		t.Fatalf("write city.toml: %v", err)
+	}
+	t.Setenv("GC_DOLT", "skip")
+	cfg := &config.City{Workspace: config.Workspace{Name: "demo"}}
+
+	names := doctorCheckNames(buildDoctorChecks(cityDir, cfg, nil, buildDoctorChecksOpts{
+		ControllerRunning:    false,
+		SkipCityDoltCheck:    true,
+		SkipManagedDoltCheck: true,
+	}))
+
+	firing := doctorCheckIndex(names, "order-firing-current")
+	if firing < 0 {
+		t.Fatalf("order-firing-current missing: %v", names)
+	}
+	outcome := doctorCheckIndex(names, "order-outcome-healthy")
+	if outcome != firing+1 {
+		t.Fatalf("order-outcome-healthy index = %d, want immediately after order-firing-current at %d; names=%v", outcome, firing, names)
+	}
+}
--- a/cmd/gc/testdata/doctor_check_names.golden
+++ b/cmd/gc/testdata/doctor_check_names.golden
@@ -36,6 +36,7 @@ service-secrets-perms
 skill-collision
 skill-dangling-sink
 order-firing-current
+order-outcome-healthy
 codex-hooks-drift
 rig-pack-coverage
 pack-runtimes
--- a/internal/doctor/checks_order_outcome.go
+++ b/internal/doctor/checks_order_outcome.go
@@ -1,0 +1,298 @@
+package doctor
+
+import (
+	"fmt"
+	"path/filepath"
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/gastownhall/gascity/internal/citylayout"
+	"github.com/gastownhall/gascity/internal/config"
+	"github.com/gastownhall/gascity/internal/events"
+	"github.com/gastownhall/gascity/internal/orders"
+)
+
+const (
+	orderOutcomeHealthyName = "order-outcome-healthy"
+
+	// orderOutcomeInspectHintFmt mirrors the sibling check's
+	// orderFiringInspectHintFmt so the pair emits consistently-shaped hints.
+	orderOutcomeInspectHintFmt = "Inspect with: gc order check && gc order history %s"
+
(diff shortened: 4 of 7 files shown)


quad341 (approved):
@remuscazacu — the part that makes this PR is not the code, it's that you went back through the event history after finding the one broken order and turned up three more sustained runs of the same shape (streaks of 45, 27, 20, 14) that nobody had ever noticed. That's the argument, and it lands.

I traced the design decisions rather than taking them on trust. Sharing `scanOrderFiringCurrentOrders` instead of reimplementing the filter is the right call — manual orders fall out structurally, no recency heuristic — and the parity test means a future divergence fails loudly. Sorting the merge by `Seq` rather than `Ts` is correct for the right reason. And the grace-window ordering bug you caught in your own review, where a real recovery near a restart gets skipped and the walk runs onto stale failures, is exactly the kind of thing that ships silently; writing a test that fails under it was the right move.

One reviewer flagged that the `"all scheduled orders succeeding"` headline reads oddly when a detail line shows a sub-threshold streak. Fair, but `order-firing-current` says `"all scheduled orders are current"` in the same situation, and I'd rather keep the pair symmetric than fix it unilaterally. Your call if you want to tune both later.

Merging as-is. Thanks for the honesty about the wisp limitation too.


---

PR #4630, opened 2026-07-25
fix(session): idle-kill ladder consults assigned work, with same-bead defer backstop

quad341:
## Summary

Adds an **assigned-work defer rung** to the idle-kill ladder, with a same-bead
consecutive-defer backstop so a stuck agent is still recyclable.

`DecideIdleTimeout` (`internal/session/lifecycle_timers.go`) previously had no
`AssignedWork` rung — its own doc comment stated *"Idle stops never consult
assigned work."* `DecideMaxSessionAge` in the same file already had one. The
result was an infinite kill/wake treadmill: the idle-kill call site stopped a
session that `ComputeAwakeSet` immediately marked awake again for the **same
unchanged assigned-work reason, on the same tick** (the code's own comment
reads *"Mark for immediate re-wake on this same tick"*). Reproduced fleet-wide
on 4 templates; roughly 3 min and ~136K context burned per cycle.

Implements **Option A** from the architecture decision in `ga-mxwj4g`
(verdict: A + C). Option C is a separate sibling bead; the two are independent
and can land in either order.

### What changed

- `DecideIdleTimeout` gains an `AssignedWork` rung mirroring
  `DecideMaxSessionAge`'s existing switch: `AssignedWorkUnknown` →
  `TimerActionGatherAssignedWork`; `AssignedWorkHas` → defer.
- The idle-kill call site (`cmd/gc/session_reconciler.go`) sources the fact via
  `sessionHasAwakeAssignedWorkForReachableStore` — deliberately **not** the
  cruder `sessionHasOpenAssignedWorkForReachableStore` that max-age uses. The
  "awake" helper excludes not-ready (deferred/blocked) work, which is the
  correct "is there something actionable right now" semantics for idle-kill.
- **Same-bead consecutive-defer backstop:** the caller tracks consecutive
  assigned-work defers per (session identity, anchor bead ID) and forces the
  Stop path once a configurable threshold is exceeded, recorded under its own
  traceable reason so operators can tell "no assigned work" stops apart from
  "exhausted defer budget" stops. `DecideIdleTimeout` stays a pure decider with
  no hidden state.
- Constraints honored: no tick reorder, no new persisted cross-tick fact shared
  with `ComputeAwakeSet`, `compute_awake_set.go` untouched, and
  `internal/session` stays side-effect-free (all store queries and tracker
  state live in `cmd/gc`).

## Scope

Single-purpose. Three commits:

| Commit | Scope |
| --- | --- |
[...]

PR #5035, opened 2026-08-05
fix(hook): drop other agents' work from gc hook under the native-store fallback (ga-lmy6yj)

quad341:
## The bug

`gc hook <agent>` serves work routed to **other agents**, and to **other rigs**.

The work query filters on `gc.routed_to`, but under the native-store fallback that predicate is not applied — so the filter silently does nothing. `--claim` is unaffected; this is the bare read path only.

Reproduced 2026-08-05 at the live v62-db/v53-binary skew. `gc hook gascity/builder` returned five beads and **one** belonged to it:

| bead | actually routed to |
|---|---|
| `ga-5hdwl6` | `gascity/deployer` |
| `ga-drlztz` | `beads/reviewer` — **a different rig** |
| `ga-77refr` | a reviewer session |
| `ga-20zoji` | nothing at all |
| `ga-2a46gb` | `gascity/builder` ✅ |

## Why it costs real money

This is the mechanism behind the phantom-wake burn (`gm-ob7is`). The agent wakes, is handed work that is genuinely not actionable by it, **correctly declines**, and pays a full turn with full cache-read to discover that. Measured ~119K cache-read per message on `beads/builder`, with wake gaps shrinking 6:10 → 2:46 — it accelerates rather than settling.

The skew is global, so every rig over-serves at once. Four rigs currently show `ready > 0` with `in_progress = 0`. Two agents were suspended by hand to stop the bleeding. Prior art for the magnitude: #4630, 522 measured no-op wake cycles in 44.7h.

## Why a post-filter in gc, and not a store fix

Deliberate, for three reasons:

1. It holds whether or not the store honours the predicate.
2. It does **not** require the gc/bd pin to move, which is under an explicit operator ruling (2026-07-30) not to touch unilaterally.
3. **#5030 would not fix this either.** Its pin (`bf97b73`, top migration `0059`) reaches v59 against a **v62** database, so gc would still fall back and the predicate would still be dropped. Closing the skew needs a beads release at v62 — which is where beads `main` already sits.

When the skew is eventually closed this becomes redundant, but stays correct.

## Fails open, everywhere

A filter that can silently empty an agent's queue is a worse outage than the over-serving it replaces. So:

- unparseable or non-array output → returned unchanged
- non-object candidates → kept
- **empty `routed_to` → KEPT.** Unrouted work is legitimately claimable, and the legacy workflow-target path (`routedToOrLegacyWorkflowTarget`) depends on it
- no identities threaded → byte-identical to previous behaviour, which keeps every other `doHook` caller safe

Only a `routed_to` that *positively names someone else* is dropped. `rig/role` and `rig--role` compare equal, so an agent's own work can never be dropped over a spelling difference.

## Tests

7 new in `cmd_hook_foreign_route_test.go`:

- the exact live five-bead shape above
- own-work preserved across all four spellings (exact, session-form identity, session-form route, case)
- identity matched from anywhere in the set
- fail-open on five malformed inputs
- `doHook` reports no-work when only foreign candidates remain
- `doHook` still reports work when one bead is genuinely ours
- `doHook` unchanged when no identities are threaded

Existing `-run Hook` and `-run 'Route|Routed|Dispatch|Assigned'` both pass. `make build` and `go vet` clean.

Fixes ga-lmy6yj. Unblocks gm-ob7is.

### DIFF
--- a/cmd/gc/cmd_hook_claim.go
+++ b/cmd/gc/cmd_hook_claim.go
@@ -16,7 +16,6 @@ import (
 	"github.com/gastownhall/gascity/internal/agent"
 	"github.com/gastownhall/gascity/internal/beadmeta"
 	"github.com/gastownhall/gascity/internal/beads"
-	"github.com/gastownhall/gascity/internal/config"
 	"github.com/gastownhall/gascity/internal/events"
 	"github.com/gastownhall/gascity/internal/executionevent"
 )
@@ -2049,44 +2048,27 @@ func hookClaimHasIdentity(assignee string, identities []string) bool {
 // hookRouteIdentitiesEqual reports whether two route/identity strings refer
 // to the same qualified agent, tolerating the tmux-safe session-name
 // encoding (/ -> --, . -> __) alongside the canonical slash-qualified form,
-// case (config-sourced and session-derived spellings of the same agent are
-// not guaranteed identical case - ga-lmy6yj), and the legacy bound-template
-// spelling ("dir/binding.name") a bound->unbound migration leaves behind
-// (legacyBoundTemplateMatchesUnboundAgent in session_reconcile.go). Every
-// normalization is applied identically to both sides, so it can only ever
-// make MORE things match - it can over-keep, never over-drop. This is the
-// single route-spelling matcher shared by the claim path
-// (hookClaimMatchesRoute) and the display path (hookCandidateVisible) per
-// ga-1xaqgo.2 - do not fork a second one.
+// and case (config-sourced and session-derived spellings of the same agent
+// are not guaranteed identical case - ga-lmy6yj). This is the single
+// route-spelling matcher shared by the claim path (hookClaimMatchesRoute)
+// and the display path (hookCandidateVisible) per ga-1xaqgo.2 - do not fork
+// a second one.
+//
+// This deliberately does NOT collapse the legacy bound-template spelling
+// ("dir/binding.name") onto its unbound form ("dir/name"): that migration is
+// owned by canonicalizeLegacyBoundUnassignedRoutedWork (build_desired_state.go),
+// which rewrites the bead's persisted route as an explicit, auditable step.
+// Treating the two spellings as always-already-equal here would let a claim
+// bypass that migration instead of triggering it (see
+// TestCanonicalizeLegacyBoundUnassignedRoutedWorkCanonicalWorkerClaims).
 func hookRouteIdentitiesEqual(a, b string) bool {
 	if a == b {
 		return true
 	}
-	unsanitizedA := agent.UnsanitizeQualifiedNameFromSession(a)
-	unsanitizedB := agent.UnsanitizeQualifiedNameFromSession(b)
-	if strings.EqualFold(unsanitizedA, unsanitizedB) {
-		return true
-	}
 	return strings.EqualFold(
-		normalizeHookBoundTemplateSpelling(unsanitizedA),
-		normalizeHookBoundTemplateSpelling(unsanitizedB),
+		agent.UnsanitizeQualifiedNameFromSession(a),
+		agent.UnsanitizeQualifiedNameFromSession(b),
 	)
-}
-
-// normalizeHookBoundTemplateSpelling collapses the legacy bound-template
-// qualified-name spelling ("dir/binding.name") onto its current unbound form
-// ("dir/name"), mirroring legacyBoundTemplateMatchesUnboundAgent's own
-// dir/Cut(local, ".") logic so a route or identity recorded under either
-// spelling still resolves to the same agent.
-func normalizeHookBoundTemplateSpelling(qualified string) string {
-	dir, local := config.ParseQualifiedName(qualified)
-	if binding, unbound, ok := strings.Cut(local, "."); ok && binding != "" && unbound != "" {
-		local = unbound
-	}
-	if dir == "" {
-		return local
-	}
-	return dir + "/" + local
 }
 
 func hookClaimMatchesRoute(candidate beads.Bead, routeTargets []string) bool {
--- a/cmd/gc/cmd_hook_claim.go
+++ b/cmd/gc/cmd_hook_claim.go
@@ -2047,18 +2047,22 @@ func hookClaimHasIdentity(assignee string, identities []string) bool {
 
 // hookRouteIdentitiesEqual reports whether two route/identity strings refer
 // to the same qualified agent, tolerating the tmux-safe session-name
-// encoding (/ -> --, . -> __) alongside the canonical slash-qualified form.
-// gc.routed_to is always written in canonical form, but comparison
-// candidates built from a runtime session name (sessionForQuery) are
-// dash-encoded, so the two spellings must compare equal. This is the single
-// route-spelling matcher shared by the claim path (hookClaimMatchesRoute)
-// and the display path (hookCandidateVisible) per ga-1xaqgo.2 - do not fork
-// a second one.
+// encoding (/ -> --, . -> __) alongside the canonical slash-qualified form,
+// and case, since config-sourced and session-derived spellings of the same
+// agent are not guaranteed identical case (ga-lmy6yj). gc.routed_to is
+// always written in canonical form, but comparison candidates built from a
+// runtime session name (sessionForQuery) are dash-encoded, so the two
+// spellings must compare equal. This is the single route-spelling matcher
+// shared by the claim path (hookClaimMatchesRoute) and the display path
+// (hookCandidateVisible) per ga-1xaqgo.2 - do not fork a second one.
 func hookRouteIdentitiesEqual(a, b string) bool {
 	if a == b {
 		return true
 	}
-	return agent.UnsanitizeQualifiedNameFromSession(a) == agent.UnsanitizeQualifiedNameFromSession(b)
+	return strings.EqualFold(
+		agent.UnsanitizeQualifiedNameFromSession(a),
+		agent.UnsanitizeQualifiedNameFromSession(b),
+	)
 }
 
 func hookClaimMatchesRoute(candidate beads.Bead, routeTargets []string) bool {
--- a/cmd/gc/cmd_hook_claim.go
+++ b/cmd/gc/cmd_hook_claim.go
@@ -16,6 +16,7 @@ import (
 	"github.com/gastownhall/gascity/internal/agent"
 	"github.com/gastownhall/gascity/internal/beadmeta"
 	"github.com/gastownhall/gascity/internal/beads"
+	"github.com/gastownhall/gascity/internal/config"
 	"github.com/gastownhall/gascity/internal/events"
 	"github.com/gastownhall/gascity/internal/executionevent"
 )
@@ -2048,21 +2049,44 @@ func hookClaimHasIdentity(assignee string, identities []string) bool {
 // hookRouteIdentitiesEqual reports whether two route/identity strings refer
 // to the same qualified agent, tolerating the tmux-safe session-name
 // encoding (/ -> --, . -> __) alongside the canonical slash-qualified form,
-// and case, since config-sourced and session-derived spellings of the same
-// agent are not guaranteed identical case (ga-lmy6yj). gc.routed_to is
-// always written in canonical form, but comparison candidates built from a
-// runtime session name (sessionForQuery) are dash-encoded, so the two
-// spellings must compare equal. This is the single route-spelling matcher
-// shared by the claim path (hookClaimMatchesRoute) and the display path
-// (hookCandidateVisible) per ga-1xaqgo.2 - do not fork a second one.
+// case (config-sourced and session-derived spellings of the same agent are
+// not guaranteed identical case - ga-lmy6yj), and the legacy bound-template
+// spelling ("dir/binding.name") a bound->unbound migration leaves behind
+// (legacyBoundTemplateMatchesUnboundAgent in session_reconcile.go). Every
+// normalization is applied identically to both sides, so it can only ever
+// make MORE things match - it can over-keep, never over-drop. This is the
+// single route-spelling matcher shared by the claim path
+// (hookClaimMatchesRoute) and the display path (hookCandidateVisible) per
+// ga-1xaqgo.2 - do not fork a second one.
 func hookRouteIdentitiesEqual(a, b string) bool {
 	if a == b {
 		return true
 	}
+	unsanitizedA := agent.UnsanitizeQualifiedNameFromSession(a)
+	unsanitizedB := agent.UnsanitizeQualifiedNameFromSession(b)
+	if strings.EqualFold(unsanitizedA, unsanitizedB) {
+		return true
+	}
 	return strings.EqualFold(
-		agent.UnsanitizeQualifiedNameFromSession(a),
-		agent.UnsanitizeQualifiedNameFromSession(b),
+		normalizeHookBoundTemplateSpelling(unsanitizedA),
+		normalizeHookBoundTemplateSpelling(unsanitizedB),
 	)
+}
+
+// normalizeHookBoundTemplateSpelling collapses the legacy bound-template
+// qualified-name spelling ("dir/binding.name") onto its current unbound form
+// ("dir/name"), mirroring legacyBoundTemplateMatchesUnboundAgent's own
+// dir/Cut(local, ".") logic so a route or identity recorded under either
+// spelling still resolves to the same agent.
+func normalizeHookBoundTemplateSpelling(qualified string) string {
+	dir, local := config.ParseQualifiedName(qualified)
+	if binding, unbound, ok := strings.Cut(local, "."); ok && binding != "" && unbound != "" {
+		local = unbound
+	}
+	if dir == "" {
+		return local
+	}
+	return dir + "/" + local
 }
 
 func hookClaimMatchesRoute(candidate beads.Bead, routeTargets []string) bool {
--- a/cmd/gc/cmd_hook_visibility_test.go
+++ b/cmd/gc/cmd_hook_visibility_test.go
@@ -26,10 +26,12 @@ func TestHookRouteIdentitiesEqual(t *testing.T) {
 		{"empty vs non-empty", "", "gascity/builder", false},
 		{"case insensitive", "Gascity/Builder", "gascity/builder", true},
 		{"case insensitive, dash-encoded", "Gascity--Builder", "gascity/builder", true},
-		{"legacy bound-template spelling, bound route", "gascity/gastown.builder", "gascity/builder", true},
-		{"legacy bound-template spelling, bound identity", "gascity/builder", "gascity/gastown.builder", true},
-		{"legacy bound-template spelling, both bound and dash-encoded", "gascity/gastown.builder", "gascity--gastown.builder", true},
-		{"legacy bound-template spelling must not collapse different agents", "gascity/gastown.deployer", "gascity/builder", false},
+		// A legacy bound-template spelling ("dir/binding.name") is deliberately
+		// NOT collapsed onto its unbound form here - that migration is owned by
+		// canonicalizeLegacyBoundUnassignedRoutedWork, which rewrites the
+		// persisted route explicitly rather than treating the two spellings as
+		// always-already-equal at compare time.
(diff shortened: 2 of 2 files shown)


quad341:
@quad341 — nice piece of debugging, and thank you for the five-bead repro table. Tracing `beads/reviewer` showing up on `gascity/builder`'s hook all the way to the cache-read burn in `gm-ob7is` is the part that makes this actionable, and putting the guard in `gc` rather than chasing a store fix is the right call given the pin ruling. The fail-open design is careful in every place that matters — keeping unrouted beads, keeping non-object rows, staying byte-identical when no identities are threaded — and the `rig/role` vs `rig--role` normalization is exactly the detail that bites later.

One thing I'll fix on my side before merging. The identity list is a little narrower than what the default work query serves:

- Tiers 1 and 2 select on **assignee only** and never look at `routed_to`, so an in-progress bead assigned to you but routed elsewhere would get dropped — that's the crash-recovery path.
- `poolDemandTarget()` collapses to `PoolName`, so a pool slot's route is the base name, which isn't in the list when `GC_TEMPLATE` is empty.
- `buildWorkQuery` also probes the legacy `workflow-control` target.

I'll route the list through `hookClaimIdentityCandidates` + `RoutedToIdentity`, skip the filter for rows carrying an assignee, and add three tests next to yours. Your diagnosis, your design, your tests — I'm just widening the identity set. Keeping the empty-`routed_to`-is-claimable rule too; you argued that one correctly.


---

<details>

```diff
From edbfa0d0c8a4326dd460c6967a0f60d005f5ee29 Mon Sep 17 00:00:00 2001
From: quad341 <quad341@users.noreply.github.com>
Date: Wed, 5 Aug 2026 09:41:51 -0700
Subject: [PATCH] fix(hook): widen gc hook routed-to identity set (ga-lmy6yj)

Thread the hook identity list through hookClaimIdentityCandidates so the
routed-to post-filter recognizes every name the agent answers to: the
legacy control-dispatcher -> workflow-control alias, and the base pool
template that poolDemandTarget() routes tier-3 demand work under
(QualifiedName() misses it on the explicit-arg path, where GC_TEMPLATE
is empty).

Also stop the filter from judging any row that carries an assignee.
Tiers 1 and 2 of the default work query select on assignee alone and
never consult gc.routed_to, so a crash-recovery bead owned by this
session may legitimately carry another agent's route. Owned work is
never foreign.
---
 cmd/gc/cmd_hook.go                    | 46 ++++++++++++++++----
 cmd/gc/cmd_hook_foreign_route_test.go | 61 +++++++++++++++++++++++++++
 2 files changed, 99 insertions(+), 8 deletions(-)

diff --git a/cmd/gc/cmd_hook.go b/cmd/gc/cmd_hook.go
index 203d98f97..b9c5e7698 100644
--- a/cmd/gc/cmd_hook.go
+++ b/cmd/gc/cmd_hook.go
@@ -470,17 +470,24 @@ func cmdHookWithOptions(args []string, opts hookCommandOptions, stdout, stderr i
 		}
 		return claimHookWork(workQuery, workDir, queryEnv, stores, claimOpts, emitQueryFailure, stdout, stderr)
 	}
-	// ga-lmy6yj: pass every name this agent answers to, so the post-filter can
-	// tell its OWN routed work from another agent's. Same identity set the claim
-	// path assembles for its assignee; a missing one only ever means we keep a
-	// candidate we could have dropped, never that we drop the agent's own work.
-	return doHook(workQuery, workDir, false, runner, stdout, stderr,
-		strings.TrimSpace(overrides["GC_ALIAS"]),
+	// ga-lmy6yj: every name this agent answers to, so the post-filter can tell its
+	// OWN routed work from another agent's. Assembled through the same helper the
+	// claim path uses so the legacy control-dispatcher alias is expanded and values
+	// are trimmed/deduped. RoutedToIdentity collapses a pool slot to its base
+	// template, which is what poolDemandTarget() routes under
+	// (internal/config/workquery.go) — QualifiedName() alone misses it whenever
+	// GC_TEMPLATE is empty, i.e. the explicit-arg path. A missing identity only
+	// ever means we keep a candidate we could have dropped, never that we drop the
+	// agent's own work.
+	hookIdentities := hookClaimIdentityCandidates(
+		overrides["GC_ALIAS"],
 		agentForQuery,
 		resolvedAgentName,
-		strings.TrimSpace(sessionForQuery),
-		strings.TrimSpace(overrides["GC_TEMPLATE"]),
+		agentutil.RoutedToIdentity(&a),
+		sessionForQuery,
+		overrides["GC_TEMPLATE"],
 	)
+	return doHook(workQuery, workDir, false, runner, stdout, stderr, hookIdentities...)
 }
 
 // hookClaimSessionVerdict classifies a runtime session's fitness to claim routed
@@ -957,6 +964,19 @@ func filterUnreadyHookCandidates(output string, now time.Time) string {
 //     legacy workflow-target path (routedToOrLegacyWorkflowTarget) depends on it.
 //     Only a routed_to that positively names someone else is dropped.
 //   - no identities resolved -> returned unchanged, never filter to nothing
+//   - ANY non-empty assignee -> KEPT, whatever the route says. Tiers 1 and 2 of
+//     the default work query select on assignee alone; owned work is never
+//     foreign. See isForeignRoutedHookCandidate.
+//
+// gc.execution_routed_to is deliberately NOT consulted: a control bead keeps
+// gc.routed_to on the dispatcher and uses execution_routed_to only to name the
+// fan-out target, so judging on it would drop the dispatcher's own work.
+//
+// The filter intentionally applies to custom work_query output too. Those
+// queries never agreed to the routed_to contract, but "this bead is positively
+// routed to someone else" is a claim about the BEAD, not about the query that
+// surfaced it — and the assignee rule above already spares the owned-work
+// shapes a custom query is most likely to return.
 func filterForeignRoutedHookCandidates(output string, identities []string) string {
 	if output == "" || len(identities) == 0 {
 		return output
@@ -991,6 +1011,16 @@ func filterForeignRoutedHookCandidates(output string, identities []string) strin
 // isForeignRoutedHookCandidate reports whether the candidate is positively
 // routed to somebody other than the asking agent.
 func isForeignRoutedHookCandidate(item map[string]any, identities []string) bool {
+	if assignee, _ := item["assignee"].(string); strings.TrimSpace(assignee) != "" {
+		// Tiers 1 and 2 of the default work query select on assignee alone and
+		// never consult gc.routed_to (standardAssignedInProgressWorkQueryScript /
+		// standardAssignedReadyWorkQueryScript). A crash-recovery row owned by
+		// this session may legitimately carry another agent's route; owned work is
+		// never foreign. Any non-empty assignee is kept — not just one matching
+		// the threaded identities — because tier 1 also matches on $GC_SESSION_ID,
+		// which is not in that set.
+		return false
+	}
 	meta, ok := item["metadata"].(map[string]any)
 	if !ok {
 		return false
diff --git a/cmd/gc/cmd_hook_foreign_route_test.go b/cmd/gc/cmd_hook_foreign_route_test.go
index eba24ac02..541cc9ed1 100644
--- a/cmd/gc/cmd_hook_foreign_route_test.go
+++ b/cmd/gc/cmd_hook_foreign_route_test.go
@@ -4,6 +4,9 @@ import (
 	"encoding/json"
 	"strings"
 	"testing"
+
+	"github.com/gastownhall/gascity/internal/agentutil"
+	"github.com/gastownhall/gascity/internal/config"
 )
 
 // ga-lmy6yj: `gc hook <agent>` is served work routed to OTHER agents — and other
@@ -182,3 +185,61 @@ func TestDoHook_NoIdentitiesUnchanged(t *testing.T) {
 		t.Fatalf("doHook without identities = %d, want 0 (unfiltered, as before)", code)
 	}
 }
+
+// Tiers 1 and 2 of the default work query select on assignee ALONE — they never
+// consult gc.routed_to. A crash-recovery bead owned by this session can
+// legitimately still carry another agent's route, and dropping it strands the
+// session's own in-progress work behind a "no work" exit.
+func TestFilterForeignRouted_KeepsAssignedWork(t *testing.T) {
+	mine := hookCandidate("ga-recovery", "gascity/deployer")
+	mine["assignee"] = "gascity--builder"
+	// Tier 1 also matches on $GC_SESSION_ID, which is NOT threaded into the
+	// identity set — so the rule is "any non-empty assignee", not "an assignee
+	// we recognize".
+	bySession := hookCandidate("ga-sessowned", "gascity/deployer")
+	bySession["assignee"] = "ga-sess01"
+
+	in := hookCandidatesJSON(t, mine, bySession)
+	got := idsIn(t, filterForeignRoutedHookCandidates(in, []string{"gascity/builder"}))
+	if len(got) != 2 {
+		t.Fatalf("kept %v, want both assigned beads (owned work is never foreign)", got)
+	}
+}
+
+// A pool slot's own tier-3 demand work is routed under the BASE template
+// (poolDemandTarget collapses to PoolName), not the slot-suffixed name. Only
+// agentutil.RoutedToIdentity produces that spelling — QualifiedName() misses it
+// on the explicit-arg path, where GC_TEMPLATE is empty.
+func TestFilterForeignRouted_MatchesPoolBaseRoute(t *testing.T) {
+	slot := config.Agent{Dir: "gascity", Name: "polecat-2", PoolName: "gascity/polecat"}
+	identities := hookClaimIdentityCandidates(
+		"",                                // GC_ALIAS: unset on the explicit-arg path
+		"gascity/polecat-2",               // agentForQuery
+		"gascity/polecat-2",               // resolvedAgentName
+		agentutil.RoutedToIdentity(&slot), // the base pool template
+		"gascity--polecat-2",              // sessionForQuery
+		"",                                // GC_TEMPLATE
+	)
+	in := hookCandidatesJSON(t, hookCandidate("ga-pool", "gascity/polecat"))
+	if got := idsIn(t, filterForeignRoutedHookCandidates(in, identities)); len(got) != 1 {
+		t.Fatalf("dropped the pool's own demand work: identities=%v kept=%v", identities, got)
+	}
+
+	// Pin that RoutedToIdentity is what saved it: without that entry the bead
+	// is (wrongly) foreign, which is the regression this guards.
+	without := hookClaimIdentityCandidates("", "gascity/polecat-2", "gascity/polecat-2", "gascity--polecat-2", "")
+	if got := idsIn(t, filterForeignRoutedHookCandidates(in, without)); len(got) != 0 {
+		t.Fatalf("expected the base-route bead to be unmatched without RoutedToIdentity, kept %v", got)
+	}
+}
+
+// buildWorkQuery probes the legacy "<rig>/workflow-control" target for
+// control-dispatcher agents, so work it serves carries that route.
+// hookClaimIdentityCandidates expands the alias; the raw strings did not.
+func TestFilterForeignRouted_MatchesLegacyWorkflowControl(t *testing.T) {
+	identities := hookClaimIdentityCandidates("gascity/control-dispatcher")
+	in := hookCandidatesJSON(t, hookCandidate("ga-ctrl", "gascity/workflow-control"))
+	if got := idsIn(t, filterForeignRoutedHookCandidates(in, identities)); len(got) != 1 {
+		t.Fatalf("dropped legacy workflow-control work: identities=%v kept=%v", identities, got)
+	}
+}
-- 
2.55.0
```

</details>

quad341:
@quad341 — the five-bead repro table is what makes this land. Tracing `beads/reviewer` work showing up on `gascity/builder`'s hook all the way through to the cache-read burn in `gm-ob7is` turned "agents keep waking for nothing" into something with a line number, and putting the guard in `gc` rather than chasing a store fix was the right call given the pin ruling.

The fail-open design holds up under tracing: unrouted rows kept, non-object rows kept, byte-identical with no identities threaded, and the `rig/role` ↔ `rig--role` normalization that can only ever over-keep. That last detail would have bitten in a month.

I'm taking the identity-set widening on my side and merging: routing the list through `hookClaimRouteTargets`/`hookClaimIdentityCandidates` so the base pool route and the legacy `workflow-control` alias are recognized, and exempting any row that carries an assignee — the default query's first two tiers select on assignee alone, so crash-recovery work can legitimately carry a stale route. Four tests alongside your seven.

Your diagnosis, your design, your tests. I'm just widening the set.


---


quad341:
Status note (no change pushed): the red CI on this branch is **not** this PR.

All failing checks share one root cause in `main`, not in this diff — `origin/main@3f4173e3c` does not compile:

```
cmd/gc/city_runtime.go:3424:7: undefined: fnv
```

`city_runtime.go` calls `fnv.New64a()` with no `hash/fnv` import. This branch merged main, so it inherited the breakage. Same error text appears in the static-checks, generated-artifacts, acceptance-A and all twelve `cmd/gc` shard jobs, and reproduces on a clean worktree at `origin/main` with an empty working tree.

Verified this PR is healthy on its own merits — adding only that one missing import, touching neither file in this diff:

- `go build ./cmd/gc/` → exit 0
- `go test ./cmd/gc/ -run 'TestFilterForeignRouted|TestDoHook'` → ok, 53 passed, 0 failed
- `go test ./cmd/gc/ -run 'Hook'` → ok, 0 failed

Tracked internally: the import fix, and separately the process gap that let a non-compiling commit reach `main` (the contributor's own head imported `hash/fnv` correctly — the branch was stale and the merge result was never built).

No action needed on this branch; it needs `main` green and a CI re-run.

quad341:
Correction to my note above — I was wrong about the remedy, and I verified the correct one.

I said this "needs `main` green and a CI re-run." `main` **is** green now (`cc036a76e`, #5038, added the missing import), but a re-run is **not** sufficient and I confirmed that empirically: I re-ran the failed jobs and they reproduced the identical `cmd/gc/city_runtime.go:3424:7: undefined: fnv`. `gh run rerun` replays the recorded merge SHA rather than recomputing it against the fixed base, so it can never clear this.

The actual remedy is to **merge current `main` into this branch**. Verified locally on a scratch worktree at this branch's head:

- `git merge origin/main` → `1 file changed, 1 insertion(+)` — the single `hash/fnv` import line, no other content
- `go build ./cmd/gc/` → exit 0
- `go test ./cmd/gc/ -run 'Hook' -count=1` → ok, 0 failures

So it is a one-line catch-up merge with no conflict and no review surface. Everything I said about this PR's own diff being healthy stands unchanged.

I have not pushed that merge — this branch isn't mine to move and the note above says you intend to land it. Happy to push it if that's useful.

quad341:
Re-reviewed `23d160e1a`..`8677427af` (13 new commits since the last review): no new concerns — the prior review stands.

quad341:
@quad341 — the five-bead repro table is what turned this from "agents keep waking for nothing" into a fix with a line number, and following `beads/reviewer` work landing on `gascity/builder`'s hook all the way through to the cache-read burn in `gm-ob7is` is what made it actionable. Putting the guard in `gc` rather than chasing the store was the right call given the pin ruling, and you argued the empty-`routed_to`-is-claimable rule correctly.

I traced the identity threading against the tree rather than the description: `poolDemandTarget()` and `RoutedToIdentity` really are the same spelling, so a pool slot recognizes its own base route, and the `binding.` normalization is symmetric on both sides — it can only over-keep. The negative controls in the pool and legacy-alias tests are what will keep those tests honest a year from now.

Build, vet, and the full Hook/Route suites are green here, and CI shows no failures. Merging as-is once the last queued jobs land — nothing needed from you.

Two things I'll watch after this ships, neither on you: if the fallback also drops `--unassigned`, the assignee exemption could let some of those rows back through; and I want to grep deployed packs for custom `work_query` overrides, since those now get post-filtered too. Both fail in the safe direction.


---

quad341:
Investigator evidence on the two questions behind the hold (ga-ibo7ph). Findings only — disposition is the mayor's call.

## 1. Overlap with merged #5807: this PR is 100% subsumed. Zero novel content.

#5807 (`79c188cafe`, merged 18:42:03Z) is the **only** commit touching either of this PR's two files since the merge-base (`de50ed0444`), and this PR touches only those two files. So `diff(main, pr/5035)` over those files *is* the post-rebase delta — and both are empty:

| file | main | this PR | |
|---|---|---|---|
| `cmd/gc/cmd_hook_claim.go` | `1ecde9ee16f6` | `1ecde9ee16f6` | identical |
| `cmd/gc/cmd_hook_visibility_test.go` | `279ee449c20c` | `279ee449c20c` | identical |

#5807's hunk is byte-identical to this one (same pre/post blobs `0846d1cf68 → 1ecde9ee16`) — same 5-line widening, same anti-collapse comment, same tests. #5807 is a **strict superset**: it also carries `release-gates/ga-lmy6yj-hook-route-filter-gate.md`, which this PR lacks. Its own body says: *"This isolated release-gated PR supersedes #5035 for merge; do not merge both."*

**Post-rebase production delta of this PR: 0 lines.** Nothing is lost by closing it superseded.

## 2. The `EqualFold` premise: refuted on the evidence.

Premise under test: *"config-sourced and session-derived route-identity spellings are not guaranteed identical case."*

- **The encode/decode pair is case-blind.** `internal/agent/session_name.go:16-24` are pure separator replacers (`/ ↔ --`, `. ↔ __`). Neither direction touches case, so the config → session-name → decode round-trip is case-*stable* by construction. `SessionNameFor` parses `session_template` with no FuncMap, and Go `text/template` has no built-in case functions, so a template cannot case-transform either.
- **No case-transforming writer on the identity path.** Every `ToLower`/`ToUpper` in `internal/config` is a different domain (extmsg provider, GitHub owner/repo, webhook scope/hostname, reserved id-prefixes). Every `EqualFold` in `cmd_hook_claim.go` other than the one under test is on a status/type enum, never an identity.
- **The route value is generated from the config spelling.** The default `sling_query` is literally `bd update {} --set-metadata gc.routed_to=<QualifiedName>` — the persisted route is emitted from the same `QualifiedName` the matcher compares against, so the two sides cannot structurally diverge.
- **Fleet sweep — 15,062 `gc.routed_to` values across all 8 stores:** 127 distinct spellings, 127 distinct after lowercasing → **zero case-only collisions**. Uppercase *is* present and already works without the fold: 38 spellings carry it (`MCDClient/*`, `ProjectWrenUnity/*`), each byte-identical to its config spelling. Cross-checked against 184 config-declared agents (55 with uppercase): zero case-differing pairs.
- **Live env** preserves config casing exactly (`GC_AGENT=MCDClient/witness`); `GC_AGENT` and `GC_TEMPLATE` agree byte-for-byte in every running session.
- **No historical precedent.** `git log -S` shows `EqualFold` entered `hookRouteIdentitiesEqual` only via this ga-lmy6yj stream; `git blame` dates the function body to `09bae7ad17` (Aug 24) with plain `==`. There is no v58/v53-era case-folding matcher to cite.

### Disproving the alternative ("the fold is load-bearing")

Removed the fold on top of current main and ran the full `cmd/gc` `Hook|Route|Visib|Claim` surface. Exactly two leaf failures, **both subtests this PR added itself**:

```
--- FAIL: TestHookRouteIdentitiesEqual/case_insensitive
--- FAIL: TestHookRouteIdentitiesEqual/case_insensitive,_dash-encoded
```

Control (fold present): 0 failures. The only thing requiring the fold is the test written alongside it.

## 3. New finding: the widening is a cross-agent hazard, and it is already on main

`validAgentName` (`internal/config/config.go:27`) permits uppercase, and `ValidateAgents` (`config.go:4031-4053`) keys duplicates on a case-**sensitive** struct `{dir, binding, name}`. So config accepts `builder` and `Builder` as two *distinct* agents:

```
ValidateAgents(builder, Builder) err = <nil>
hookRouteIdentitiesEqual("gascity/builder", "gascity/Builder") = true
```

That is one agent matching a *different* agent's route — the same cross-agent work-visibility class ga-lmy6yj exists to prevent.

## Disposition impact

#5807 merged the fold at 18:42:03Z, *before* the 19:48Z hold. So "strip the fold from #5035" is a no-op — this PR has zero delta. Closing it as superseded is cleanly supported. The open question the disposition rules did not anticipate is whether main should keep the widening; that revert would be ~5 lines at `cmd/gc/cmd_hook_claim.go:2068-2071`. Flagging for the mayor, not acting on it.

quad341:
Closing as superseded by #5807, which merged 2026-08-31T18:42Z and is a strict superset of this PR: both files here are blob-identical to current main (verified: post-rebase delta is provably zero lines — see the evidence comment above, issuecomment-5485151743). The EqualFold widening this PR introduced is being reverted separately: the case-mismatch premise was refuted by a fleet-wide sweep (15,062 route values, zero case-only collisions, no case-transforming writer on the identity path), and the fold creates a cross-agent match hazard (config validates 'builder' and 'Builder' as distinct agents; the fold reports their routes equal — the exact class this fix exists to prevent).

PR #4497, opened 2026-07-21
fix(runtime/tmux): record poke on hidden-attached nudge path; tag real-tmux dogfood (post-merge #4187)

julianknutsen:
Post-merge remediation for the landed **#4187** (tmux nudge-path poke discount).
A post-merge review of the landed range `6b0eb0d6b..5131e3b57` found two
actionable issues; this PR fixes both. It does not change #4187's intended
behavior — it closes a residual gap and moves a new dogfood test out of the
protected default-lane debt.

### 1. Hidden-attached `NudgeNow` bypassed poke recording (correctness)
#4187 stopped gc's own nudge keystrokes from inflating `last_active` on the
`NudgeSession`/`NudgePane` paths, but `NudgeNow`'s hidden-attached-client branch
(`sendHiddenAttachedText`) still injected gc's own keystrokes without recording a
poke. `ResetInterruptedTurn` drives exactly this path (the detached-gemini
`/rewind` + Enter through a hidden client), so `GetSessionActivity` would count
gc's own input as the agent responding and mask a woken-but-unresponsive session.

Fix: extract the shared *capture-prior-before-write / stamp-after-delivery*
contract into `beginPoke` and apply it to the hidden-attached send path as well
as `NudgeSession`/`NudgePane` (a failed write records nothing). Added a
deterministic `Provider.NudgeNow` regression test that injects a hidden client
and a fake executor, asserts a poke is recorded, and verifies a post-grace echo
discounts back to the genuine pre-nudge activity (no real tmux, no sleeps).

### 2. New real-tmux dogfood grew the untagged fixed-sleep census (test hygiene)
`nudge_poke_integration_test.go` was untagged, so although it is env-skipped in
normal `go test`, its fixed sleeps raised the protected untagged fixed-sleep
census baseline from 284/112 to 290/113.

Fix: move it behind `//go:build integration` (the convention the other real-tmux
tmux tests use) and revert the untagged fixed-sleep baselines to 284/112 across
`census.go`, `test-resources.toml`, and `TESTING.md`. The all-source audit
baseline is unchanged — the sleeps still live in a tracked file.

### Tests
- `go test ./internal/runtime/tmux/` (default lane) — full package passes,
  including the new `TestNudgeNowHiddenAttachedRecordsPoke`.
- `go test ./internal/testpolicy/resourcecensus/` — ledger/doc sync passes with
  the reverted baselines (actual untagged count confirmed 284/112).
- `go vet -tags integration ./internal/runtime/tmux/` clean; the retagged
[...]

PR #5036, opened 2026-08-05
fix(tmux): poke hidden-attached SendKeys so gc's detached picker keystrokes are discounted

julianknutsen:
## Summary

Post-merge review follow-up for #4497 (workflow root `ga-wisp-nd4`). The review
found one actionable major-correctness gap: #4497 closed the residual `NudgeNow`
poke gap on the hidden-attached **text** path (`sendHiddenAttachedText`), but the
sibling hidden-attached **key** path (`sendHiddenAttachedKeys`, driven by
`Provider.SendKeys` and `Provider.Interrupt`) still injected gc's own keystrokes
without recording a poke.

`ResetInterruptedTurn` sends `/rewind` through the now-poked text path, then
drives the detached Gemini rewind picker with `SendKeys`. Because the picker
navigation (`waitForPane` + sleeps) outlasts `pokeEcho`, the `/rewind` poke can
no longer bracket the final confirming `Enter`, so `GetSessionActivity` counts
gc's own picker keystroke as the agent responding — masking a woken-but-stuck
rewind flow.

## Fix

`sendHiddenAttachedKeys` now records a poke, mirroring `sendHiddenAttachedText`:

- Resolve every key to its byte sequence up front. An unknown key still falls
  through to the `SendKeysRaw` path (`used=false`) with **no** partial
  hidden-client write and **no** recorded poke.
- Wrap the writes in `beginPoke`/`commit`: the genuine prior activity is captured
  before the first key and stamped only after the last key lands. `beginPoke`
  carries a still-unanswered earlier poke's baseline forward (via
  `pokePriorBaseline`), so chained gc echoes don't become the new prior.

This also extends poke coverage to the detached `Interrupt` (`C-c`) path, which
shares `sendHiddenAttachedKeys`.

## Tests

- New regression test `TestSendKeysHiddenAttachedRecordsPoke` models a picker
  cadence that exceeds `pokeEcho` (a stale `/rewind` poke) and asserts the
  confirming keystroke re-stamps the poke and that a post-grace echo discounts
  back to the genuine pre-rewind activity. Synthetic times only — stays in the
  default (non-integration) lane.
- Updated the runtime-tmux manifest inventory and shard counts for the new test.
- `go test ./internal/runtime/tmux` (full package) and
  `go test ./scripts -run TestRuntimeTmuxManifest` pass; `go vet ./...` clean.


quad341 (dismissed):
@julianknutsen — this is exactly the follow-up I want out of a post-merge review. You found the sibling path #4497 left behind, and the *why* is what makes it convincing: the picker navigation in `ResetInterruptedTurn` genuinely outlasts `pokeEcho`, so the `/rewind` poke can't bracket the confirming `Enter` and gc's own keystroke reads as the agent waking up. I traced that through `adapter.go` and it holds exactly as you described.

Two things I appreciated beyond the stated fix. Resolving every key up front quietly kills a partial-write bug — the old code wrote the first key to the hidden client and *then* fell through to `SendKeysRaw`, double-sending it. And you noticed the detached `Interrupt` gets poke coverage for free by sharing the helper, and said so instead of leaving it as a surprise.

Getting the manifest and shard-count bookkeeping right in the same PR saved a round trip. Verified locally, CI green across all six runtime-tmux shards. Merging as-is — thank you.


---

julianknutsen:
## Fable review council

**Verdict: Approve with one must-fix before merge.** The core change is correct and well-scoped: `sendHiddenAttachedKeys` now records a poke via `beginPoke`/`commitPoke` in exact conformance with the established contract (`sendHiddenAttachedText`, `NudgeSession`, `NudgePane`), and the pre-resolution of key byte sequences fixes a real pre-existing double-delivery bug (a partial hidden-client write followed by a full `SendKeysRaw` resend on unknown-key fallthrough). All three reviewers independently verified the state-safety of the mid-sequence failure path (`beginPoke` mutates nothing before commit; abandoning the closure leaks nothing), the idle-detection semantics (genuine agent turns escape the discount; only echo-confined-then-silent output is discounted, which is precisely the gc-driven picker redraw), and that the new test `TestSendKeysHiddenAttachedRecordsPoke` genuinely fails on `origin/main` and passes at head. The single gating item is a test-coverage gap on the PR's own second claimed fix, confirmed 3/3 under adversarial verification.

### Confirmed gating findings

- **[major]** `internal/runtime/tmux/tmux.go` (~line 1764) — **Unknown-key fallthrough (fix (a)) ships with zero test coverage.** Confirmed 3/3. The pre-resolution guarantee — an unknown key returns `used=false` with no partial hidden-client write and no recorded poke — is introduced and explicitly documented by this PR's in-code comment, and it fixes a real bug (on main, `SendKeys(name, "Enter", "F5")` with a hidden client wrote `"\r"` to the hidden stdin, then `adapter.go:~712` re-sent both keys via `SendKeysRaw` — double-Enter). Exhaustive search of the package confirms no test pins it: `TestSendKeysHiddenAttachedRecordsPoke` drives only the known-key happy path, and `TestHiddenAttachedKeyBytesSupportsArrowNavigation` (`executor_test.go:252`) checks only known-key byte mappings, never the `ok=false` branch. A refactor back to the per-key resolve-then-write loop would pass the entire suite while silently reintroducing the partial-write/double-send bug. The fix is cheap — `recordingWriteCloser`, `fakeExecutor`, and the `hiddenAttachClients` injection pattern all exist in `nudge_poke_hidden_test.go`: add a test calling `p.SendKeys(sess, "Enter", "F5")` and assert `sink.written() == ""`, no poke entry for the session, and that the fake executor saw the `SendKeysRaw` fallback for both keys. (One reviewer filed this at minor; verification consolidated it to major — it is the edge case of the PR's headline change (a), and the repo's quality gates require edge-case coverage.)

### Non-gating observations

- **[minor]** `internal/runtime/tmux/nudge_poke_hidden_test.go` (~line 139) — the new test cannot distinguish `beginPoke`'s carry-forward from a naive raw snapshot. The fake executor returns `genuine.Unix()` as raw activity while the seeded stale poke's `at` is months away, so `pokePriorBaseline` takes the fresh-raw branch and `pk.prior.Equal(genuine)` passes either way; meanwhile the in-code comment (tmux.go ~1776-1779) claims carry-forward for this path. Nothing is unprotected — the carry branch is pinned by the pure-function tests (`TestPokePriorBaseline`, `TestChainedUnansweredNudgesPreserveGenuinePrior`) — but either soften the comment or have the executor return an echo-adjacent raw time (within `pokeEcho` of the stale poke) so the assertion can only pass via carry-forward.
- **[minor]** `internal/runtime/tmux/tmux.go` (~line 1781) — the "a failed write records nothing" property is structurally sound but untested (no erroring `WriteCloser` fixture exists). Pre-existing convention: the sibling `sendHiddenAttachedText` has the identical untested pattern on main.
- **[nit]** `internal/runtime/tmux/tmux.go` (~line 1782) — a mid-sequence write failure leaves keys 1..N-1 already delivered with no poke committed, so their echo can briefly count as agent activity. Verified NOT a state-safety bug and byte-for-byte consistent with the documented `beginPoke` "never commit unconfirmed delivery" contract on all four send paths. Deliberate tradeoff; any change belongs in a separate contract-wide PR.
- **[nit]** `internal/runtime/tmux/tmux.go` (~line 1780) — hidden-attach sends bypass the per-session nudge lock, so a narrow begin/begin overlap between a concurrent text nudge and SendKeys can snapshot an uncommitted sibling's echo as prior. Pre-existing for text/text pairs; this PR actually narrows the committed-ordering case (a hidden SendKeys echo previously always polluted a later nudge's prior).
- **[nit]** `internal/runtime/tmux/adapter.go` (~lines 711-726, 282) — the non-hidden `SendKeysRaw` fallbacks in `Provider.SendKeys` and `Provider.Interrupt` still record no poke, the residual instance of the same last_active-inflation class this PR fixes for the hidden path. Out of the PR's stated scope; candidate follow-up.

### Review basis

Three independent Fable reviewers, distinct lenses: (1) state-safety and contract consistency of the `beginPoke`/`commitPoke` bracketing, including the mid-sequence failure, `len(keys)==0`, and `used=true`-on-error paths against both call sites (`Provider.SendKeys`, `Provider.Interrupt`); (2) activity/idle-detection semantics — `discountPokeActivity`/`pokePriorBaseline` composition, `ResetInterruptedTurn` interleaving of the "/rewind" nudge poke with per-key SendKeys pokes across `waitForPane` sleeps, and concurrency interleavings; (3) test strength and bookkeeping — verified `TestSendKeysHiddenAttachedRecordsPoke` fails on `origin/main` (the stale-poke `!pk.at.After(stale)` assertion trips) and that the hidden-branch guard is sound (`SendKeysRaw` routes through the fake executor, not the sink), and independently recomputed the manifest from `git show d0d923c62c:scripts/runtime-tmux-tests.manifest`: 356 entries, no duplicates, new entry correctly placed, 356 = 239 untagged + 117 tagged, shards {60,60,59,59,59,59}, consistent with main's 355/238/117 baseline. Gating candidates went through 3-vote adversarial verification; the single major was confirmed unanimously with the pre-change double-delivery scenario reproduced from `origin/main` source. Validation evidence: `go test ./internal/runtime/tmux` and `go test ./scripts -run TestRuntimeTmuxManifest` green locally; GitHub CI fully green on head d0d923c62c.
### Resolution

The confirmed major was addressed in `a76832d4b7` (`test(tmux): pin the unknown-key fallthrough of hidden-attached SendKeys`): `TestSendKeysHiddenAttachedUnknownKeyFallsThroughWithoutPartialWrite` drives `p.SendKeys(sess, "Enter", "F5")` against an injected hidden client and asserts no hidden-client bytes, no recorded poke, and an in-order two-key `SendKeysRaw` fallback. The test was verified to fail against `origin/main`'s `tmux.go` with exactly the predicted double-delivery shape (`hidden client received "\r"`) and passes at head. Manifest bookkeeping updated to 357 entries / 240 untagged / shards {60,60,60,59,59,59}; `go test ./internal/runtime/tmux ./scripts -count=1` green.

PR #4073, opened 2026-07-08
fix(reconciler): release routed-away pool assignment from a live source session

remuscazacu:
## Problem

Cross-agent handoff deadlocks the target pool. When a pool session hands work off by re-routing a bead to a different agent (`gc sling <bead> <other-agent>` — e.g. an L1 session escalating to `l2-<family>`), the bead keeps its `assignee` pointing at the **still-live source session**.

`releaseOrphanedPoolAssignments` already reopens routed pool assignments so the target pool gains demand, but it gates release on the live-session ownership guards (`openSessionOwnsWork` / `liveOpenSessionAssignmentExists`), which exist for **dead-session orphans**. Because the source session is still open, release is skipped, so:

- the bead stays assigned and is excluded from the target pool's demand (`defaultScaleCheckCountsAndDemand` counts only unassigned `Ready` beads), and
- it keeps the source session anchored (`"assigned-work"` in `computeAwakeSet`, which ignores `routed_to`), so the source re-processes work it already handed off.

The handoff therefore never wakes the target pool until the source session is closed.

## Fix

Release the assignment when the bead is routed away to a different agent than **every live session that owns the assignee**, once the handoff has had time to settle. A new helper `assigneeRoutedAwayFromOwnAgent` resolves ownership mirroring `openSessionOwnsWork`, biased to **preserve** (an over-release steals work a live session is actively doing — worse than a missed wake):

- **scans all matching sessions**, not just the first — one identity string can be held by two live sessions via `alias_history`, or the same identity across rig stores;
- **store-ref-aware** via `openSessionReachableStoreRefInfo` — a matching session in a different concrete store is not an owner (#3621 / #1544);
- a **cross-store-eligible (city-scoped) owner** federates work across stores (vp-kvp) and is never treated as a handoff;
- concludes route-away **only** when every store-scoped owner resolves to a different, non-cross-store agent than the routed target; any target-agent, cross-store, or unresolvable owner preserves the bead.

### Liveness: a settle window on `gc.routed_at`

[...]

PR #3667, opened 2026-06-23
fix(runtime): refresh demand snapshots for routed work

duncan4123:
## Summary
- include ready routed work in the runtime demand snapshot cache key
- refresh patrol demand snapshots when new ready work appears without a session/config change
- add a regression covering routed work that should produce poolDesired after a cached zero-demand patrol

## Test
- go test ./cmd/gc -run 'TestCityRuntimeDemandSnapshot(ReusesStablePatrolDemand|RefreshesForNewRoutedReadyWork)'\n\nThis targets the observed failure where ready beads routed to imported role pools existed in the active bead store, but supervisor patrol reused a stale zero-demand snapshot and never started the role worker.

PR #5038, opened 2026-08-05
fix(runtime): add missing hash/fnv import to city_runtime.go (main is red)

remuscazacu:
## Problem

`main` does not compile. #3667 added `sessionBeadSnapshotFingerprint`, which calls `fnv.New64a()`, without adding `hash/fnv` to `cmd/gc/city_runtime.go`'s import block:

```
$ git checkout 3f4173e3c   # current main, no other commits
$ go build ./cmd/gc/
# github.com/gastownhall/gascity/cmd/gc
cmd/gc/city_runtime.go:3424:7: undefined: fnv
```

Because CI builds each PR **merged with** `main`, this fails every job that compiles `cmd/gc` on every open pull request: all twelve `cmd/gc process` shards, the integration suites, `Preflight / static checks`, and `Preflight / generated artifacts` — the last surfacing it as `genschema: generating CLI docs: exit status 1`, which reads like a codegen problem rather than a missing import. I found it while working out why 28 checks went red on #4073 with a diff that touches none of those files.

## Fix

The import. One line, in sorted position; nothing else in #3667 needed changing.

## Verification

```
$ go build ./...                  # clean
$ go vet ./cmd/gc/                # clean
$ gofumpt -l cmd/gc/city_runtime.go   # clean
```

#3667's own tests pass with it in place:

```
--- PASS: TestSessionBeadSnapshotFingerprintReflectsRawMetadata
--- PASS: TestCityRuntimeDemandSnapshotRefreshesForNewRoutedReadyWork
--- PASS: TestCityRuntimeDemandSnapshotReusesStablePatrolDemand
--- PASS: TestCityRuntimeDemandSnapshotRetainsOnlyPoolScaleCheckPartials
--- PASS: TestCityRuntimeTickDispatchesOrdersBeforeDemandSnapshot
```

Those could not have run as merged, since the package they live in does not build — which is presumably how this got through.

Sending it as its own PR rather than folding it into #4073, so it can land immediately and unblock everyone else's CI too.

### DIFF
--- a/cmd/gc/city_runtime.go
+++ b/cmd/gc/city_runtime.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"hash/fnv"
 	"io"
 	"log"
 	"os"


quad341 (approved):
Approving and merging immediately — main is red and this is the correct minimal fix.

**The breakage is mine, not a defect in anything you did.** I squash-merged #3667 (`3f4173e3c`) about an hour ago. That PR's diff added `fnv.New64a()` in `city_runtime.go` *and* the `hash/fnv` import, but the squash landed the usage without the import, so `cmd/gc` stopped compiling on `main`.

I'd actually hit the identical failure in a local branch earlier today and fixed it there by hand — and then failed to connect that to what the squash would do to `main`. That's on me.

Verified `origin/main` is red with exactly `cmd/gc/city_runtime.go:3424:7: undefined: fnv`, and this one-line import is precisely the fix. Thanks for catching it and turning it around this fast, @remuscazacu.

PR #5047, opened 2026-08-05
formula: warn when an extends var-override silently drops a parent's required

csauer02-personal-user:
## formula: warn when an `extends` var-override silently drops a parent's `required`

When a child formula overrides a variable that a parent declared `required = true`, the `extends` merge (`merged.Vars[name] = varDef`) replaces the whole `VarDef` object — silently discarding the parent's `required` flag. There is no signal that the contract was weakened, so a downstream consumer that assumed the variable was required can mint with an empty value (the misconfiguration class behind the empty-artifact-path bug fixed in the companion gascity-packs PR).

**Change:** on that override, `parser.go` now appends a deduplicated cook-time warning (surfaced via `Recipe.Warnings` → `printRecipeWarnings` to stderr) naming the variable and both formulas. **Child-wins behavior is unchanged** — the override still applies exactly as before; this is observability only, so no existing formula changes behavior or breaks.

Deliberately warn-not-preserve: preserving `required=true` through the override would fail legitimate launches where the value is produced mid-flow rather than supplied at launch (runtime var validation runs on raw opts before defaults apply). A warning surfaces the risk without breaking that valid pattern.

### Change size
Production Go: ~74 lines (`parser.go` warning logic + `recipe.go`/`compile.go`/`cmd_formula.go` plumbing). Tests: 107 lines (`parser_test.go`: drop-warns / child-wins / keeps-required-no-warn / no-override-no-warn). No generated, no docs.

### Verification
`go vet ./...` clean; `go build ./...` clean; `go test ./internal/formula/...` PASS including the new `TestResolve_ExtendsOverrideDropsRequiredWarns` (3 subtests). Behavior-preserving; two independent adversarial reviews confirmed side-effect-free.

Companion PR (gascity-packs) carries the decompose-stage hardening that this warning helps surface.

### DIFF
--- a/cmd/gc/cmd_formula.go
+++ b/cmd/gc/cmd_formula.go
@@ -518,6 +518,16 @@ func printGraphV2Deprecations(stderr io.Writer, deprecations []string) {
 	}
 }
 
+// printRecipeWarnings surfaces non-fatal cook-time advisories raised while
+// resolving a formula (e.g. an extends override dropping a parent's required
+// flag, dip-5hkepo). They never gate cook; they make an otherwise-silent drop
+// observable to the operator.
+func printRecipeWarnings(stderr io.Writer, warnings []string) {
+	for _, w := range warnings {
+		fmt.Fprintf(stderr, "warning: %s\n", w) //nolint:errcheck
+	}
+}
+
 func formulaCommandError(stderr io.Writer, command string, jsonOutput bool, err error) error {
 	if err == nil || jsonOutput {
 		return err
@@ -672,6 +682,7 @@ conflicting live workflow from the same source is an error.`,
 						if err != nil {
 							return fmt.Errorf("compile: %w", err)
 						}
+						printRecipeWarnings(stderr, recipe.Warnings)
 						if err := molecule.ValidateRecipeRuntimeVars(recipe, molecule.Options{Title: title, Vars: cookVars}); err != nil {
 							return fmt.Errorf("validate runtime vars: %w", err)
 						}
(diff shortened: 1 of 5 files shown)


quad341:
@csauer02-personal-user — this is a sharp catch, and I want to name the part I liked most: you found the footgun (`merged.Vars[name] = varDef` swallowing the parent's `required` whole) and then *declined* the obvious fix. Preserving `required` through the override would have been the tempting one-liner, and it would have broken every legitimate flow where the value is produced mid-run. `required` and `default` are mutually exclusive here, so that merged `VarDef` would have been self-contradictory anyway. Warning instead of enforcing is right, and putting the reasoning in both the code comment and the test docstring is what future readers will actually need.

Your tests earned their keep too — asserting child-wins is *still* intact right beside the new warning is what let me confirm "behavior-preserving" by reading rather than trusting.

One gap I'll take on my side rather than bounce back to you: the standalone legacy (non-graph.v2) cook path goes through `molecule.Cook`, which compiles internally and drops the recipe, so the warning never reaches stderr there. I'll wire that branch the way your graph.v2 branch already is, add a stderr-level test plus one for the `compile.go` propagation line, re-run vet/build/tests, and merge with your authorship intact. Detection logic and intent are all yours.


---

csauer02-personal-user:
Closing by the author: observability-only and not run on any served build of ours. quad341's read stands (a warning rather than enforcement is right, and the legacy molecule.Cook path is the remaining wire-in); parser.go still assigns merged.Vars[name] = varDef bare, so the footgun remains and the branch is yours to adopt if wanted. This note was agent-drafted.

PR #3584, opened 2026-06-18
feat(compact): dolt compact-state health — quarantine auto-clear, doctor check, disk preflight, backup sync, runbook

quad341:
## Summary

Extracts 5 cohesive dolt compact-state health improvements from the superseded #3579 bundle and lands them cleanly on main. Also includes 2 test-infrastructure fixes that were stranded alongside them.

### Compact-health pieces (the 5 in scope)

- **fix(dolt): auto-clear proven false-positive compact quarantines** (ga-3pg1y6.2) — known same-count hash drift markers self-heal on next compact pass when the table-list preservation probe succeeds; hard-block classes and probe failures still block
- **feat(doctor): add dolt-compact-state doctor check** (ga-qc5chv.2) — `gc doctor` now reports stale compact lifecycle markers (compact-quarantine, compact-pending-gc, compact-pending-push) with targeted fix hints
- **feat(compact): implement disk-CRITICAL preflight guard** (ga-wfdunn.1) — `GC_DOLT_COMPACT_MIN_FREE_BYTES` (default 5 GiB); exits 0 (skip) when below floor; probe failure fails open; zero disables; invalid value exits 2
- **feat(compact): implement optional pre-GC backup sync** (ga-hn88dw.1) — `GC_DOLT_COMPACT_BACKUP_REMOTE` / `GC_DOLT_COMPACT_BACKUP_TIMEOUT_SECS`; runs `dolt backup sync` before each flatten and bare-GC run
- **docs(runbooks): add Dolt Storage Maintenance operator runbook + link fixes** (ga-gx25ij.1) — `docs/runbooks/dolt-compact.md`; nav entry in `docs.json`; link to it from `dolt-bloat-recovery.md`; fix 4 internal links + promote bold pseudo-headers to `###` per reviewer (ga-wexskk)

### Test-infrastructure fixes (not part of the 5, but stranded in the same bundle and not yet in main)

- **fix(test): ShortTempDir always uses /tmp** (ga-uz6mr1) — long TMPDIR (deployer isolation path) pushed socket paths over the 108-byte Linux limit; this fix keeps CI stable
- **fix(dolt): recognize GOTMPDIR-rooted test configs in dolt-cleanup reaper** (ga-xxp9cd.2) — when `GOTMPDIR=/var/tmp/gc-build-tmp` is set on build machines, the reaper now recognizes managed Dolt test configs under that root

### Not included (explicitly excluded)

[...]

Issue #3541, opened 2026-06-16
compact flatten verification false-quarantines a dolt_ignore'd table force-inlined into HEAD (dolt#11131-healed store) — #3431 follow-up


rjgeng:
### Summary
#3431 scoped compact flatten verification to the committed root (`committed_tables` = `SHOW TABLES AS OF HEAD`), excluding `dolt_ignore`'d tables on the assumption they are working-set-only and absent from every commit. That assumption breaks for a **force-healed store**, which still false-quarantines.

### Environment
- dolt **2.1.6**; gascity compact path `examples/bd/dolt/commands/compact/run.sh` (post-#3431, merge `89002ab3`).
- Observed on a live force-healed `hq` (recovered via the dolt#11131 schema-encoding heal) across **three consecutive compactor fires** (deterministic under flatten-window write load).

### Steps to reproduce
1. Start from a store whose `dolt_ignore`'d wisp tier was **force-inlined into HEAD** by a dolt#11131 heal (`DOLT_ADD('--force','wisps')` + commit) to clear the dirty-working-set write-block.
2. `SHOW TABLES AS OF HEAD` now lists `wisps` **while it is still matched by `dolt_ignore`**.
3. Drive flatten-window churn so the live per-table count moves (e.g. 3→5) with hash drift, while committed HEAD stays at 3.
4. Run the #3431-as-merged compact flatten verification.

### Expected
`wisps` is excluded from per-table verification — a `-Am` flatten can never stage a `dolt_ignore`'d table regardless of committed-root presence, so its live count/hash was never the flatten's to preserve.

### Actual
`committed_tables` includes `wisps` (present in HEAD) → it stays in per-table verification → its flatten-window gain+drift reads as corruption → **false quarantine** (`post-flatten INTEGRITY check failed`, exit 1).

### Notes
- **Persists in merged #3431** — flagged pre-merge in the review comment on that PR; confirmed it survived the merge (both #3431 commits predate this case).
- Suggested fix direction: add the **symmetric** exclusion — also exclude a user table matched by an active `dolt_ignore` pattern, regardless of committed-root presence. HEAD-movement corruption detection (`DOLT_HASHOF_DB('HEAD')`) stays intact, so real committed corruption still quarantines.
- A repro-backed fix + red→green test is ready (will reference this issue).

_Refs #3431, #3341, #2348._

Issue #3341, opened 2026-06-11
Dolt compact integrity check races continuous writers — quarantine loop silently blocks GC and backups on busy DBs


vbtcl:
## Problem
The dolt pack's compaction verifies integrity with a strict `before == after` value-hash/row-count comparison around the flatten. On a **continuously-written** database (a busy city's `beads_hq`: wisps, dispatcher, witness traffic), any write landing inside the ~seconds-long flatten window diverges the check → compaction writes a quarantine marker → **all GC for that DB is silently disabled, indefinitely**.

## Consequences observed (production city, 2026-06)
- Quarantine marker has **no alerting and no auto-staleness**: `beads_hq` sat quarantined 16 days; the noms journal grew 5.1G (data dir 13G) — the trajectory that previously ended in `corrupted journal` and a full city-down recovery.
- **Backups fail while quarantined** (backup dog: `beads_hq(sync failed)`) — so the corruption-risk window also has degraded recovery options.
- The loop is self-sustaining: clearing the marker and re-running compact just re-races the writers. Observed 4 raced attempts in 48h (one HEAD-changed safe abort, three post-flatten divergence re-quarantines); manual compaction only succeeds in a lucky no-write window.

## Why the check is wrong-by-design for live DBs
The divergences are overwhelmingly **append-only churn captured mid-flatten** (wisp/event inserts) — i.e. row-count *gains*, not data loss. Strict equality cannot distinguish "captured a concurrent insert" from "lost data."

## Suggested fixes (any of)
1. **Tolerate monotonic gains**: pass integrity when post-flatten rows ⊇ pre-flatten rows (per-table count increase + pre-image hash containment), quarantine only on loss.
2. **Quiesce/lock writers** for the flatten window (or retry-until-quiet with bounded attempts).
3. **Snapshot-isolate** the comparison (compare against the flatten's own source snapshot, not the live head).
4. Decouple the backup path from GC state so a quarantined DB still syncs.

## Mitigation already written (needs review)
`vbtcl/gascity#1` (fork PR): (a) doctor emits a HIGH advisory whenever an active quarantine marker exists; (b) compact auto-clears markers older than a threshold once the DB reads clean *and* is quiescent (whole-DB value hash stable across two probes), then retries — a supervised retry that never bypasses integrity. That PR fixes the *silent-indefinite* part; this issue is the underlying race.

Happy to provide the full incident timeline/logs.

sjarmak:
[...]

Issue #2740, opened 2026-05-29
HQ Dolt store grows unbounded: 7.3GB / 190k commits / 230k events in 18 days; compaction never runs, dog pool thrashes


Teo-Asinari:
## Summary

The HQ Dolt store (`.gc` city HQ beads DB, `.beads/dolt/hq`) grows
**unbounded** under normal single-user operation. After **18 days** of
ordinary use it reached **7.3 GB / 190,347 Dolt commits / 230k event
rows**. Once large enough, full-table scans (`bd list`) hit the 2-minute
timeout under normal agent load, which makes the **dog pool thrash**:
a dog claims a bead → its work-discovery `bd list` times out → the
reconciler releases/kills it → bead flips back to `open` → repeat. Work
never runs and no `[BLOCKED]` note is left.

`CALL DOLT_GC()` reclaims **0 bytes** because every one of the 190k
commits is still reachable from `main`. The only recovery is a manual
dump → re-init → reload to drop history.

## Environment

- `gc` version **1.1.1**
- `dolt` version **2.0.0**
- macOS (Apple Silicon), single operator, ~18 days of use
- 6 rigs registered; HQ prefix `gc`

## Evidence

Commit breakdown (`hq/main`.`dolt_log`, 190,347 total):

| commits | message bucket |
|--------:|----------------|
| 72,617 | `bd: create <id>` |
| 49,280 | `bd: update <id>` |
| 36,657 | `bd: close <id>` |
| 31,501 | `schema: auto-migrate` |

~10,000 commits/day, every day since 2026-05-11.

Working-set tables (current HEAD, not history):

- `issues`: **36,363 rows** — 36,328 `closed`, 34 `open`, 1 `blocked`.
  Of those, **35,678 are `issue_type=task`** (ephemeral orchestration
  beads) and **645 are `issue_type=session`**. Only **35 beads are
  actually live.**
- `events`: **230,443 rows**.
- `compaction_snapshots`: **0** and `issue_snapshots`: **0** — the
  compaction machinery appears to **never run**.

Concrete examples of the churn:

- A single dog **session** bead `gc-itodr1` (type=session, title "dog")
  was rewritten **7,384 times** logging its own heartbeat, then closed
  with reason `gc_swept`. The row and its 7,384 commits remain.
- `schema: auto-migrate` committed **31,501 times**. It fires on write
  operations (verified: read-only `bd show`/`bd ready` add none).

## Two distinct defects

[...]

PR #5049, opened 2026-08-06
fix(dolt): stop the one-shot first-flatten compact quarantine on gc's un-ignored probe table

julianknutsen:
## Problem

Every gc-managed city's probe-host database (the lexicographically first user db — where the read-only health probe persists `__gc_read_only_probe` with **no `dolt_ignore` registration**) hard-quarantines on its first threshold-crossing flatten: `DOLT_COMMIT('-Am')` first-commits the probe table, `DOLT_HASHOF_DB('HEAD')` drifts while per-table verification is clean, and `db_root_drift_within_verified_tables` correctly fails closed on a drift table outside the verified set — blocking all GC of that db until an operator clears the marker. Production hits: hq (June 2026), daa (2026-08-04, category `same_row_count_db_hash_drift`).

## Fix (fable-council-reviewed design; mechanics empirically verified on dolt 2.1.8)

1. **Both probe-creation sites** (`cmd/gc/dolt_sql_health.go` and the `gc-beads-bd.sh` fallback probe) register the probe table in `dolt_ignore` via `INSERT IGNORE`, appended **last** so read-only classification still keys on CREATE/REPLACE failing first. Verified: the ignore rule stops `-Am` from committing the probe table.
2. **`compact/run.sh` preamble `version_dirty_dolt_ignore`**: `-Am` commits a dirty `dolt_ignore` like any tracked table (verified), so the compactor stages and target-commits it *before* the pre-flight HEAD capture. Fail-open with WARN — never blocks the flatten.
3. **Proof extension**: `db_root_drift_within_verified_tables` now admits preflight-excluded tables (never-committed, and #3541 force-heal dolt_ignore'd-committed) when their `DOLT_DIFF` is proven **added-only** via the shared `diff_is_additive_only` helper — the daa shape defers instead of quarantining. Everything unproven still fails closed; no quarantine auto-clear (per the #3584 revert).
4. **Diagnosability**: `DOLT_DIFF_STAT` stderr surfaced; empty-DIFF_STAT and unproven tables logged with reasons; drift-table list recorded in the quarantine marker (`db_root_drift_stat_tables`); invalid names guarded.
5. Corrected the false "`-Am` cannot stage or touch them" comments.

## Verification

- TDD throughout; 8 new fake-dolt scenarios/tests (defer, non-additive quarantine, diff-probe failure, mixed daa shape, preamble ordering + dry-run skip, add/commit WARN paths — the WARN tests were mutation-proved load-bearing). Go statement pins close a real hole (a malformed CREATE targeting `dolt_ignore` previously passed every ManagedDolt test — falsified before fixing).
- `make test` green; `go vet ./...` clean; shellcheck: zero new findings (one intentional SC2016 matching the adjacent idiom).
- Existing fail-closed pins unchanged: `TestCompactScriptStillQuarantinesDbHashDriftBeyondVerifiedTables`, row-decrease, `dolt_schemas` system-table drift.
- The four sh-harness exec call sites were factored into one `runShHarness` helper; the banked subprocess-census reduction (411→409 untagged, 552→550 all, 405→403 Small) is lockstepped across `census.go`, `test/test-resources.toml`, and `TESTING.md`.

## Follow-ups

Tracked in **ga-7yrto**: category-2 defer-log label + test, DIFF_STAT-failure fake mode, mixed three-way scenario, fake-harness DOLT_DIFF arm-ordering comment, `dolt_ignore` preamble-failure residual (fail-closed, one-shot), gain-drift lib test CI wiring. Beads-repo (separate): clone-local `metadata` identity rows (`clone_id`/`repo_id`) get committed by flattens and force-pushed to remotes.

---
- Related to #2740 — addresses a systematic one-shot quarantine in the Dolt history-flatten compaction path — an unregistered `__gc_read_only_probe` table got first-committed by the flatten's `DOLT_COMMIT -Am`, drifting the database hash and blocking the flatten and full GC on every gc-managed city's probe-host database until an operator cleared the marker; it does not cover the no-op `schema: auto-migrate` commit churn or the working-set/event pruning that the filed issue also asks for
- Related to #3341 — addresses a different false-positive in the same compaction quarantine gate — the health-probe table that is never registered in dolt_ignore and so gets first-committed by the flatten's `-Am`, drifting the database hash — which removes one systematic class of GC-blocking quarantines; the concurrent-writer race described in that issue is not covered here


quad341:
@julianknutsen — this is excellent incident work. Tracing the hq and daa quarantines back to gc's *own* un-ignored probe table, and then catching that a dirty `dolt_ignore` just relocates the same trap onto `dolt_ignore` itself, is the kind of second-order find that's easy to miss. The `diff_is_additive_only` extraction is the right shape too: the first-commit exemption is only worth having because you kept it fail-closed, and you did.

The three red checks are real, not flake. The new `INSERT IGNORE INTO \`db\`.\`dolt_ignore\`` needs a *session* database — over the SQL server with none selected, Dolt returns `Error 1105: no root value found in session`. Your local `dolt sql -q` inside a repo dir has an implicit db, which is exactly why it looked clean. Adding `USE <db>` as the first probe statement fixes all three paths; I reproduced the failure and verified the fix on 2.2.1.

I'll take that on my side — the `USE` prefix plus the statement-index pins — and re-run the real-server suite. Your design stands unchanged: registration last, `INSERT IGNORE` so an operator's `ignored=0` survives. I also confirmed the compact preamble behaves against real dolt exactly as you designed it. Thanks for this one.


---

quad341:
@julianknutsen — this is excellent incident work. Tracing the hq and daa quarantines back to gc's *own* un-ignored probe table would have been enough; catching that a dirty `dolt_ignore` just relocates the same trap onto `dolt_ignore` itself, and adding the preamble to close it, is the second-order find most people would miss. I checked the ordering claim against the source rather than taking it on faith: the preamble really does land after the threshold gate and before both hash captures (run.sh:2318 → 2321), and the dry-run skip is real. The `diff_is_additive_only` extraction is the right shape — the exemption is only worth having because everything unproven still fails closed, and it does.

The red checks are ours, not yours. The new `INSERT IGNORE INTO \`db\`.\`dolt_ignore\`` needs a selected session database; your local `dolt sql -q` inside a repo dir has one implicitly, a server connection doesn't. I'll add the `USE <db>` prefix at both probe sites, shift the statement-index pins (keeping your "registration last" and "never REPLACE the rule" invariants intact), and re-run the real-server suite before merging.

One thing I'll file separately rather than push onto you: hq and daa already have the probe table committed in HEAD, so they may need a manual probe reset before they flatten clean. Doesn't affect this PR. Thanks for this one.


---

_Heads-up: we hit a transient network/publish issue pushing the one-line maintainer test-bound to this branch — nothing needed from you. The review stands; we'll retry the merge shortly._


<details>

```diff
From 8dd8cc62e8dfb74e155c3d5f55e1f8f0c816c543 Mon Sep 17 00:00:00 2001
From: Julian Knutsen <julianknutsen@users.noreply.github.com>
Date: Thu, 6 Aug 2026 19:55:14 -0700
Subject: [PATCH] fix(dolt): select the session database in the read-only probe
MIME-Version: 1.0
Content-Type: text/plain; charset=UTF-8
Content-Transfer-Encoding: 8bit

The dolt_ignore registration added to the read-only probe targets
`<db>`.`dolt_ignore` by fully-qualified name, which is enough for
`dolt sql -q` run inside a repo directory but not for a SQL-server
connection: a server session has no implicitly selected database, and
Dolt resolves dolt_ignore against the session's root value. Every
probe over --host/--port therefore failed with

    Error 1105 (HY000): no root value found in session

taking down all and only the real-server probe paths (cmd/gc process
shards, packages-core integration, packages-cmd-gc-integration).

Prepend `USE <db>` to the probe statements so the session has a
selected database. Both consumers are covered: the `dolt sql -q` path
joins the slice with "; ", and managedDoltReadOnlyStateDirect execs it
statement-by-statement on a single *sql.Conn, so the USE persists for
that connection. Read-only classification is unaffected — USE succeeds
on a read-only server and the CREATE/REPLACE still fail first. The
shell fallback in gc-beads-bd.sh gets the same prefix so it keeps
mirroring managedDoltReadOnlyProbeStatementsFor.
---
 cmd/gc/beads_provider_lifecycle_test.go   |  3 ++-
 cmd/gc/dolt_sql_health.go                 |  5 ++++
 cmd/gc/dolt_sql_health_test.go            | 32 +++++++++++++++--------
 examples/bd/assets/scripts/gc-beads-bd.sh |  7 +++--
 4 files changed, 33 insertions(+), 14 deletions(-)

diff --git a/cmd/gc/beads_provider_lifecycle_test.go b/cmd/gc/beads_provider_lifecycle_test.go
index 26c11f1cb..f67907cf0 100644
--- a/cmd/gc/beads_provider_lifecycle_test.go
+++ b/cmd/gc/beads_provider_lifecycle_test.go
@@ -1049,7 +1049,8 @@ printf 'status=%s\n' "$status"
 
 	bt := "`"
 	probe := bt + "beads" + bt + "." + bt + managedDoltProbeTable + bt
-	wantSQL := "CREATE TABLE IF NOT EXISTS " + probe + " (k INT PRIMARY KEY); " +
+	wantSQL := "USE " + bt + "beads" + bt + "; " +
+		"CREATE TABLE IF NOT EXISTS " + probe + " (k INT PRIMARY KEY); " +
 		"REPLACE INTO " + probe + " VALUES (1); " +
 		"INSERT IGNORE INTO " + bt + "beads" + bt + "." + bt + "dolt_ignore" + bt +
 		" (pattern, ignored) VALUES ('" + managedDoltProbeTable + "', 1);"
diff --git a/cmd/gc/dolt_sql_health.go b/cmd/gc/dolt_sql_health.go
index 1b01e85ef..db69c7104 100644
--- a/cmd/gc/dolt_sql_health.go
+++ b/cmd/gc/dolt_sql_health.go
@@ -68,6 +68,10 @@ var managedDoltSystemDatabases = map[string]struct{}{
 // registration is last so a read-only server still fails on the CREATE or
 // REPLACE, which is what the read-only classification keys on, and it uses
 // INSERT IGNORE so an operator's explicit `ignored = 0` override survives.
+// The leading USE selects the session database: a SQL-server connection has no
+// implicitly selected database, and the fully-qualified name is not enough for
+// dolt_ignore, which Dolt resolves against the session's root value ("no root
+// value found in session" otherwise).
 // db must be a real user database; the empty string returns nil so the caller
 // can skip the probe entirely. The database identifier is backtick-quoted
 // because Dolt derives DB names from repository directory names, which can start
@@ -80,6 +84,7 @@ func managedDoltReadOnlyProbeStatementsFor(db string) []string {
 	quotedDB := managedDoltQuoteIdent(db)
 	target := quotedDB + "." + managedDoltQuoteIdent(managedDoltProbeTable)
 	return []string{
+		"USE " + quotedDB,
 		"CREATE TABLE IF NOT EXISTS " + target + " (k INT PRIMARY KEY)",
 		"REPLACE INTO " + target + " VALUES (1)",
 		"INSERT IGNORE INTO " + quotedDB + ".`dolt_ignore` (pattern, ignored) VALUES ('" + managedDoltProbeTable + "', 1)",
diff --git a/cmd/gc/dolt_sql_health_test.go b/cmd/gc/dolt_sql_health_test.go
index 26e8e34b6..80cc2d980 100644
--- a/cmd/gc/dolt_sql_health_test.go
+++ b/cmd/gc/dolt_sql_health_test.go
@@ -32,8 +32,11 @@ func TestManagedDoltReadOnlyProbeNeverTargetsLegacyDatabase(t *testing.T) {
 		}
 		wantTable := "`" + db + "`.`" + managedDoltProbeTable + "`"
 		wantIgnoreTable := "`" + db + "`.`dolt_ignore`"
+		// The leading USE names the database bare rather than as a qualified
+		// table, so it is a third accepted shape.
+		wantUse := "USE " + managedDoltQuoteIdent(db)
 		for _, q := range stmts {
-			if !strings.Contains(q, wantTable) && !strings.Contains(q, wantIgnoreTable) {
+			if !strings.Contains(q, wantTable) && !strings.Contains(q, wantIgnoreTable) && q != wantUse {
 				t.Fatalf("probe stmt for %s missing %q: %s", db, wantTable, q)
 			}
 			if strings.Contains(q, "`.`__probe`") {
@@ -55,25 +58,32 @@ func TestManagedDoltReadOnlyProbeNeverTargetsLegacyDatabase(t *testing.T) {
 func TestManagedDoltReadOnlyProbeRegistersIgnoreRuleLast(t *testing.T) {
 	for _, db := range []string{"gascity", "003", "name-with-hyphen", "with`backtick"} {
 		stmts := managedDoltReadOnlyProbeStatementsFor(db)
-		if len(stmts) != 3 {
-			t.Fatalf("probe stmts for %s = %v, want 3 statements", db, stmts)
+		if len(stmts) != 4 {
+			t.Fatalf("probe stmts for %s = %v, want 4 statements", db, stmts)
+		}
+		// A server connection has no implicitly selected database, and
+		// dolt_ignore resolves against the session root value, so the probe
+		// must select the database first or every statement after it errors.
+		wantUse := "USE " + managedDoltQuoteIdent(db)
+		if stmts[0] != wantUse {
+			t.Fatalf("probe stmts for %s first statement = %q, want %q", db, stmts[0], wantUse)
 		}
 		wantIgnore := "INSERT IGNORE INTO " + managedDoltQuoteIdent(db) + ".`dolt_ignore` (pattern, ignored) VALUES ('" + managedDoltProbeTable + "', 1)"
-		if stmts[2] != wantIgnore {
-			t.Fatalf("probe stmts for %s last statement = %q, want %q", db, stmts[2], wantIgnore)
+		if stmts[3] != wantIgnore {
+			t.Fatalf("probe stmts for %s last statement = %q, want %q", db, stmts[3], wantIgnore)
 		}
 		// Prefix pins alone would accept a CREATE/REPLACE aimed at dolt_ignore,
 		// which the relaxed legacy-target loop above also lets through; pin the
 		// quoted probe-table target the same way the ignore rule is pinned.
 		wantTarget := managedDoltQuoteIdent(db) + "." + managedDoltQuoteIdent(managedDoltProbeTable)
-		if !strings.HasPrefix(stmts[0], "CREATE TABLE IF NOT EXISTS ") || !strings.Contains(stmts[0], wantTarget) {
-			t.Fatalf("probe stmts for %s must create %s first: %q", db, wantTarget, stmts[0])
+		if !strings.HasPrefix(stmts[1], "CREATE TABLE IF NOT EXISTS ") || !strings.Contains(stmts[1], wantTarget) {
+			t.Fatalf("probe stmts for %s must create %s first: %q", db, wantTarget, stmts[1])
 		}
-		if !strings.HasPrefix(stmts[1], "REPLACE INTO ") || !strings.Contains(stmts[1], wantTarget) {
-			t.Fatalf("probe stmts for %s must write the probe row into %s second: %q", db, wantTarget, stmts[1])
+		if !strings.HasPrefix(stmts[2], "REPLACE INTO ") || !strings.Contains(stmts[2], wantTarget) {
+			t.Fatalf("probe stmts for %s must write the probe row into %s second: %q", db, wantTarget, stmts[2])
 		}
-		if strings.Contains(stmts[2], "REPLACE INTO") {
-			t.Fatalf("probe stmts for %s must not REPLACE the ignore rule (an operator ignored=0 override has to survive): %q", db, stmts[2])
+		if strings.Contains(stmts[3], "REPLACE INTO") {
+			t.Fatalf("probe stmts for %s must not REPLACE the ignore rule (an operator ignored=0 override has to survive): %q", db, stmts[3])
 		}
 		joined := managedDoltReadOnlyProbeSQLFor(db)
 		if !strings.HasSuffix(joined, wantIgnore+";") {
diff --git a/examples/bd/assets/scripts/gc-beads-bd.sh b/examples/bd/assets/scripts/gc-beads-bd.sh
index 1c959378f..e549a40fb 100755
--- a/examples/bd/assets/scripts/gc-beads-bd.sh
+++ b/examples/bd/assets/scripts/gc-beads-bd.sh
@@ -1435,9 +1435,12 @@ check_read_only() {
     # the database hash and quarantines GC for that database (hq June 2026, daa
     # 2026-08-04). INSERT IGNORE keeps an operator's explicit ignored = 0
     # override. The registration is last so a read-only server still fails on
-    # the CREATE or REPLACE, which the classification below keys on. Mirrors
+    # the CREATE or REPLACE, which the classification below keys on. The leading
+    # USE selects the session database: a server connection has no implicitly
+    # selected database, and the fully-qualified name is not enough for
+    # dolt_ignore, which Dolt resolves against the session's root value. Mirrors
     # cmd/gc/dolt_sql_health.go managedDoltReadOnlyProbeStatementsFor.
-    sql="CREATE TABLE IF NOT EXISTS ${quoted_db}.${probe_table} (k INT PRIMARY KEY); REPLACE INTO ${quoted_db}.${probe_table} VALUES (1); INSERT IGNORE INTO ${quoted_db}.${ignore_table} (pattern, ignored) VALUES ('__gc_read_only_probe', 1);"
+    sql="USE ${quoted_db}; CREATE TABLE IF NOT EXISTS ${quoted_db}.${probe_table} (k INT PRIMARY KEY); REPLACE INTO ${quoted_db}.${probe_table} VALUES (1); INSERT IGNORE INTO ${quoted_db}.${ignore_table} (pattern, ignored) VALUES ('__gc_read_only_probe', 1);"
     if output=$(dolt --host "$host" --port "$DOLT_PORT" --user "$DOLT_USER" --password "${DOLT_PASSWORD:-}" --no-tls \
         sql -q "$sql" 2>&1); then
         return 1
-- 
2.55.0
```

</details>

julianknutsen:
## Fable review council

**Verdict: approve — merge as-is.** Five reviewers examined `origin/main...35687ceac2` through five lenses (Dolt/SQL correctness, compact-flatten semantics, shell robustness/portability, test coverage, regression/operational compatibility) and none found a blocker or major. The fix is sound on every load-bearing axis: the probe statement order (`USE` → `CREATE` → `REPLACE` → `INSERT IGNORE`, registration last) preserves the read-only classification on both probe paths and is pinned by tests in both the Go builder and the shell mirror; all four statements are idempotent and the `INSERT IGNORE` preserves an operator's exact-pattern `ignored = 0` override; `version_dirty_dolt_ignore` is correctly ordered before HEAD/root capture, skipped under dry-run, and never blocks the flatten; and the new `db_root_drift_within_verified_tables` gate stays fail-closed on every unproven input (tables outside both sets, invalid names, empty `DOLT_DIFF_STAT`, diff-probe failures, non-added-only diffs). The re-admission cannot skip a quarantine that protects committed data. Everything below is non-gating hardening and coverage follow-up material.

### Confirmed gating findings

None.

### Non-gating observations

**Minors**

1. **`dolt_ignore` registration failure is health-fatal in Go but only a warning in the shell fallback** — `cmd/gc/dolt_sql_health.go:~347` (`managedDoltReadOnlyStateDirect`). A non-read-only failure of the trailing `INSERT IGNORE` returns `("unknown", err)`, which `managedDoltHealthCheck` propagates as a hard failure; in `recoverManagedDoltProcessWithOps` that skips the healthy-reuse branch and stop/restarts an otherwise writable server — exactly the shape the pre-`USE` Error 1105 CI failure demonstrated. The shell fallback (`examples/bd/assets/scripts/gc-beads-bd.sh` `check_read_only`) degrades the identical condition to a non-fatal status 2 at both call sites. Since the new drift gate independently makes an unregistered probe table benign, consider treating a registration-only failure after a successful CREATE+REPLACE as a logged warning in the Go path, matching the shell semantics. (Two reviewers, independently.)

2. **A residual one-shot quarantine class remains, transferred onto `dolt_ignore` itself.** Two routes: (a) the fail-open WARN paths in `version_dirty_dolt_ignore` (`run.sh:~790` — add failure, commit failure, failed staged probe) leave `dolt_ignore` dirty for the `-Am` to first-commit, which the function's own comment acknowledges lands in "quarantine, not defer"; (b) a narrow race where gc's first-ever probe registers `dolt_ignore` after the preamble's one-time `dolt_status` check but before the flatten's `-Am` (the window spans remote fetch plus up to three preflight attempts). In both cases the gate's "outside verified set" branch (`run.sh:~1341`) hard-quarantines because `dolt_ignore` can never be in either verified set. An added-only `dolt_ignore` diff only adds ignore patterns and cannot lose committed data, so re-admitting it under the existing `diff_is_additive_only` proof would close both holes. (Two reviewers, independently.)

3. **Unscoped `DOLT_COMMIT` in the preamble can sweep unrelated staged tables** — `run.sh:~819`. The guard confirms `dolt_ignore` is staged but not that it is the *only* staged table; `CALL DOLT_COMMIT('-m', 'compaction: version dolt_ignore')` commits the whole staged root, so concurrently- or previously-staged tables (operator `DOLT_ADD`, crashed-flatten residue) get committed under a misdescriptive message. Blast radius is bounded — it lands before all HEAD/preflight captures, so no verification gate can mis-fire — but it degrades the forensic trail and contradicts the function's "in its own commit" claim. Fix: also probe `SELECT COUNT(*) FROM dolt_status WHERE staged = 1` and skip with a WARN when it is not exactly 1. (Three reviewers, independently.)

4. **Version-skew deployment hazard**: a new gc binary paired with an older materialized dolt pack (pre-PR `run.sh` with neither the preamble nor the re-admission) will first-commit the probe's uncommitted `dolt_ignore` row on the next flatten and hard-quarantine — the fixed incident class re-triggered on already-recovered databases. Packs do not necessarily move in lockstep with the binary. A release note / engdocs entry stating the pack must ship with or before the binary (recovery: existing rm-marker + `--gc-only` runbook) would close this.

5. **Registration only reaches the currently-first user database** — `cmd/gc/dolt_sql_health.go:~126`. `managedDoltReadOnlyState` probes only `dbs[0]`, so stale `__gc_read_only_probe` tables planted in other databases under older orderings never get registered; each will be first-committed once by its next flatten (deferred, absorbed by the drift gate, but permanently embedded in committed history). `managedDoltResetProbe` already has the all-DB loop pattern that could be reused.

6. **`reset-probe` no longer removes the probe's full footprint** — `cmd/gc/dolt_sql_health.go:~406`. `managedDoltResetProbe`/`Direct` drop the probe tables but leave the newly-planted `('__gc_read_only_probe', 1)` row in every probed database's `dolt_ignore` — where uncommitted, the "reset" leaves `dolt_ignore` dirty. This silently weakens the completeness of release-gate criterion 6 in `release-gates/ga-hivi-probe-user-db-gate.md`. Extend both reset paths with a per-DB `DELETE FROM dolt_ignore WHERE pattern = '__gc_read_only_probe'` (via `USE`, given the session-root requirement), or document the retention.

7. **Test-coverage follow-ups** (test lens; regression-hardening, not correctness):
   - The load-bearing `WHERE diff_type <> 'added'` predicate in `diff_is_additive_only` (`compact-gain-drift-proof.sh:42`) is unpinned: neither the fixture fake (`dog_exec_scripts_test.go:~1036`) nor the sh stub inspects the WHERE clause or endpoint order, so an inverted predicate (`= 'added'`) would admit destructive drift while every test stays green.
   - No test asserts the `dolt_ignore` row actually lands on a real server; `INSERT IGNORE` downgrades row-level failures to warnings, so registration could silently stop landing while all tests — including the three real-server CI tests — pass. A one-line `SELECT ignored FROM dolt_ignore WHERE pattern = '__gc_read_only_probe'` assertion in `TestManagedDoltHealthCheckWithPasswordUsesDirectHelpersAgainstRealServer` would pin it.
   - The gate's widened exemption also covers force-heal-inlined `dolt_ignore`'d tables (#3541 shape, `run.sh:~1332`), but all four new fixture modes exercise only the never-committed probe-table shape; neither direction (admit added-only / quarantine non-additive) is pinned for the inlined-ignored category.
   - The composed post-fix steady state (dirty `dolt_ignore` **and** working-set probe table in one run) and several new failure branches (`dolt_status` count-probe failure, add-ok-but-staged=0 skip, invalid-name branch, empty-`DIFF_STAT` message, `DIFF_STAT` probe-failure stderr surfacing) have no fixture modes.

**Nits**

- `TestCompactScriptProceedsWhenDoltIgnoreCommitFails` (`dog_exec_scripts_test.go:~2944`) pins a no-quarantine outcome only because the fake keeps `DOLT_HASHOF_DB` stable in that mode; on a real server the surviving dirty `dolt_ignore` would drift and quarantine, so the WARN-path tests slightly overstate the real-world contract. Add drift to the fixture (asserting the documented quarantine) or a comment making the fake's story consistent.
- Hand-built `` `dolt_ignore` `` backticks and an unescaped single-quoted pattern literal (`dolt_sql_health.go:~91`, mirrored at `gc-beads-bd.sh:~1444`) bypass `managedDoltQuoteIdent`; safe today because both values are quote-free compile-time constants pinned by exact-string tests, but inconsistent with the function's own convention.
- The added-only proof is vacuous for genuinely first-committed tables (every row is "added" by construction); the header comment at `run.sh:~1283` sells it as load-bearing for both categories, and the defer message at `run.sh:~2774` labels inlined-ignored tables "first-committed". Reword.
- The "operator `ignored = 0` override survives" doc comment (`dolt_sql_health.go:~70`) holds only for an exact-pattern row; a glob override (`__gc_*`, ignored=0) is outranked by the probe's more specific rule. Qualify the comment.
- `TestCompactScriptVersionsDirtyDoltIgnoreBeforeFlatten` (`dog_exec_scripts_test.go:~2911`) pins ordering only relative to `DOLT_RESET`, not the HEAD-capture probe; also assert the commit precedes the first HEAD-probe query.
- `TestGcBeadsBdReadOnlyFallbackRegistersProbeTableInDoltIgnore` (`beads_provider_lifecycle_test.go:~1050`) hand-builds `wantSQL` instead of deriving it from `managedDoltReadOnlyProbeSQLFor("beads")`, so the "shell mirrors Go" invariant is enforced only via two independently-maintained literals.
- `diff_is_additive_only` interpolates commit hashes into SQL literals with only a non-empty check, while other hash inputs in `run.sh` get an explicit alphanumeric `case` reject; add the same guard for consistency (the `DOLT_DIFF_STAT` instance predates this PR).

### Review basis

- **Lenses:** Dolt/SQL correctness (probe semantics, statement ordering, idempotency, injection surface, read-only classification); compact-flatten semantics (preamble ordering vs HEAD capture, fail-open/fail-closed analysis of the drift gate, quarantine machinery interactions); shell robustness/portability (POSIX sh compliance, `set -e`/`set -u` hazards, mktemp cleanup on all branches, SQL quoting, Go/shell mirror byte-equivalence); test coverage (invariant pinning altitude, fixture fidelity, resource-census lockstep); regression/operational compatibility (upstream-merge content, ratchet artifacts, runbook and release-gate impact, version-skew scenarios).
- **Refs inspected:** `git diff origin/main...35687ceac2` and the individual commits (`f9ddbfbe4c` original fix, `35687ceac2` the `USE` opener added after CI exposed the Error 1105 session-root failure). All file content read at the reviewed refs via `git show`.
- **Validation evidence:** the Error 1105 failure was reproduced against a real `dolt sql-server` via both the mysql driver and the dolt CLI, and the `USE` fix verified on both paths; the three previously-failing CI tests (`TestManagedDoltHealthCheckWithPasswordUsesDirectHelpersAgainstRealServer`, `TestRecoverManagedDoltProcessWithPasswordReusesHealthyRealServer`, `TestManagedBdRigProviderStoreRecoversAfterHardKillPortRebind`) pass locally along with `go test ./examples/bd/dolt`, dolt-related `cmd/gc` tests, and `go vet`; full GitHub CI is green on head `35687ceac2`.
- **Adversarial verification:** no finding rose to gating severity; the empty verified-gating set is consistent with all five reviewers independently recommending merge.

PR #5050, opened 2026-08-06
controller: auto-cook ready fire-roots (judgment-free patrol promotion)

csauer02-personal-user:
## controller: auto-cook ready fire-roots (judgment-free patrol promotion)

A `build-from-plan` fire-root that becomes `bd ready` (its blocked-by preconditions clear) does not auto-fire — it needs a manual `gc sling`. That manual step is a single point of failure: a lost/unread GO signal leaves ready work unfired.

**Change:** `promoteReadyFireRoots` — a controller patrol promotion mirroring the existing `restoreCarriedWorkRoutes` (ZFC / judgment-free). On startup + each patrol tick, for each `bd ready` bead that declares `gc.fire_formula` (+ optional `gc.fire_vars`), is unassigned, kind-less, and not already fired, it cooks that declared formula via the internal sling/cook path. It cooks **only what the bead declares** — no `if`-reasoning, no role names, no mail in the loop. The blocked→ready transition (closing the bead's gate deps) IS the GO, as bead-state.

**Idempotency:** guarded by a fail-closed `HasMoleculeChildren` probe AND a `molecule.Options.IdempotencyKey` keyed on the fire-root ID (stamped atomically with Cook) + a `findLiveFiredRoot` pre-check — so a lost best-effort forward-pointer after a successful cook does not re-cook on a later tick (this promotion re-runs every tick, unlike a one-shot sling).

**Size:** ~74 production Go (new `cmd/gc/fire_root_promotion.go` + 2-line wiring + 2 keys) + ~135 test. New file + function mirroring an existing pattern; minimal delta.

**Verification:** golden tests assert all criteria (blocked→ready fires the declared formula with vars 1:1; idempotent no-double-fire under a transient forward-pointer stamp failure — proven non-vacuous, reverting yields 2 convoys; declared-only ZFC; no mail). `go build/vet/test ./cmd/gc/ ./internal/molecule/...` green. Two independent adversarial reviews: both green.

Note (transparency): the cmd/gc e2e regression drives the legacy `Instantiate` stamp path; the graph.v2 graph-apply stamp is symmetric and separately unit-covered (`internal/molecule` `TestInstantiateWithIdempotencyKey`).


quad341:
Thanks for this, @csauer02-personal-user — `promoteReadyFireRoots` (auto-cooking ready fire-roots) is a nice piece of work and a genuinely useful addition.

Two maintainer-side things landed on this branch just now, on top of your original commits (your authorship on the feature work is untouched):

1. **Rebase onto current `main`** — the branch was ~100 commits behind; resolved the one real conflict in `internal/beadmeta/keys.go` (main had independently added quarantine keys and, via your own already-merged #5213, removed the stale heartbeat key) by taking main's current key block and re-inserting your two new `FireFormula`/`FireVars` keys at the right spot.

2. **Routed `promoteReadyFireRoots`'s store enumeration through `internal/storeref`** instead of the direct `rigBeadStores()` call at `fire_root_promotion.go:133`. This repo enforces a zero-baseline ratchet requiring store-enumeration sites to go through `storeref.Plan`/`EachLeg` rather than hand-rolled iteration — an internal architectural boundary with no external visibility, so there was no way to know about it going in. Used the same `Plan → EachLeg → per-leg read` shape as two existing call sites (`ready_federation.go`, `census_residency.go`) rather than `Walk`, since `EachLeg` performs no I/O and preserves your existing per-scope error handling (`Walk`'s fatal-leg-abort semantics would have changed behavior on a work/binding-leg error).

One behavior change flagged rather than folded in silently: `RoutedWork` plans also enumerate class-binding legs, which the original code didn't scan (city + rigs only). That's a small change beyond a pure refactor — likely a latent gap fix, since `gc ready` already treats bindings as live ready-work surface the same way.

Both changes sit on top of your commits, tests pass, `go vet` is clean. This is ready for maintainer merge review from here — thank you for the feature, it's yours.

csauer02-personal-user:
Closing by the author: we do not run this on any served build of ours, and I will not spend a rebase on a change we are not exercising. quad341, thank you for rebasing it on 08-22 and routing the store enumeration through internal/storeref; your two commits stay on the branch, and the feature is yours to carry forward if you want it. The defect it addressed is still on main (no fire_formula / promoteReadyFireRoots at the tip). This note was agent-drafted.

PR #5051, opened 2026-08-06
mail: mark-read+close injected auto-handoffs instead of hard-deleting

csauer02-personal-user:
## mail: mark-read + close injected auto-handoffs instead of hard-deleting them

`ArchiveInjectedAutoHandoffs` hard-`Delete`d unread auto-handoff mail (a context-cycle continuation carrying a GO/signal) the instant `gc prime`'s stdout write returned nil — with no read/ack gate. "Injected" only means the CLI's `Fprint` returned; it says nothing about whether the recycled agent actually consumed the handoff before a crash/race/re-cycle. So a handoff lost in that window was permanently deleted.

**Change:** replace the hard-`Delete` with **mark-read + close (retain-addressable)**. Mark-read stops re-injection (the inject query fetches only unread); closing with `RetentionSweepCloseReason` keeps the bead addressable (`isRemovedMessageBead` treats that reason as not-removed), so an unconsumed handoff is recoverable; the already-correct read-gated TTL sweep reclaims it later. One unified retention path, no special-case delete, no permanent data-loss. (The TTL sweep itself was already correctly read-gated and is unchanged.)

**Size:** ~20 production lines (one function) + tests. **Verification:** the mark-read `Update` is byte-identical to canonical `MarkRead`; additive label semantics verified across MemStore/NativeDolt/BdStore; every branch of the 3-step (mark-read → stamp → close) partial-failure ordering is idempotent and self-heals via the read-gated sweep. New tests assert the handoff survives inject + is addressable + is later TTL-reclaimed (non-vacuous: a revert to hard-delete fails them). `go build/vet/test` green. Two independent adversarial reviews: both green.

Note (transparency): one cosmetic stale "archiving …" stderr string remains (non-functional).

### DIFF
--- a/cmd/gc/cmd_mail_test.go
+++ b/cmd/gc/cmd_mail_test.go
@@ -3606,6 +3606,27 @@ func TestMailCheckInjectDoesNotCloseBeads(t *testing.T) {
 	}
 }
 
+// assertAutoHandoffRetainedAddressable pins the dip-6ov51a contract at the inject
+// integration boundary: an injected auto-handoff is MARK-READ + CLOSED and stamped
+// with the retention marker (retain-addressable), NOT hard-deleted, so an
+// unconsumed handoff stays recoverable until the read-gated TTL sweep reclaims it.
+func assertAutoHandoffRetainedAddressable(t *testing.T, store beads.Store, id string) {
+	t.Helper()
+	b, err := store.Get(id)
+	if err != nil {
+		t.Fatalf("auto handoff %s must be retained addressable after injection, not hard-deleted; store.Get = %v", id, err)
+	}
+	if b.Status != "closed" {
+		t.Errorf("auto handoff %s status = %q, want closed", id, b.Status)
+	}
+	if got := b.Metadata[mail.ReadMetadataKey]; got != "true" {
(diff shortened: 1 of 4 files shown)


quad341 (approved):
@csauer02-personal-user — thank you for this one. The core insight is the valuable part: "injected" only ever meant `Fprint` returned nil, and treating that as an acknowledgement was enough to permanently delete a continuation carrying a live GO. That's a subtle, expensive bug, and naming the window precisely is most of the fix.

What I especially liked is that you didn't invent a new lifecycle state to solve it. Routing the handoff into the existing mark-read + `RetentionSweepCloseReason` path means `isRemovedMessageBead` keeps it addressable and the read-gated sweep reclaims it later — one retention path instead of a special-case delete. I traced both load-bearing claims in the checkout rather than taking your word for them, and they hold: the retention reason really is exempted, and `CheckAutoHandoffs` gates on open *and* unread, so there's no re-injection.

The `assertAutoHandoffRetainedAddressable` helper was the right call too — it propagates the new contract to all five existing inject sites instead of leaving them asserting the old delete semantics. The never-injected-stays-unread guard is a genuinely good regression pin.

Merging as-is. The `"archiving …"` string you flagged reads fine — archive is still what the operation means. I'm filing a follow-up on the default `[mail] retention_ttl` being empty, since these retained beads only get reclaimed once it's set. That's ours, not yours.


---

PR #5052, opened 2026-08-06
session: keep the min_active_sessions pool floor warm (idle-timeout exempt)

csauer02-personal-user:
## session: keep the min_active_sessions pool floor warm (exempt from idle-timeout)

`min_active_sessions` is spawned as a floor (the min-fill top-off pass), but the idle-timeout decision (`DecideIdleTimeout`) was floor-blind — it killed an idle floor session like any elastic one, and the only floor-aware revive path was scoped to `sleep_reason=city-stop`. So an idle-timeout-slept floor session was never revived, only cold-recreated on the next tick: the floor oscillates 0↔1, satisfied on paper but never held warm. That churn is the per-task warm+nudge tax.

**Change:** make the idle-timeout path floor-aware. A pool session within the `min_active_sessions` floor (the deterministic lowest-bead-id `minSess` same-template warm sessions) is now exempt from idle-timeout kill — kept warm, never killed-then-cold-recreated — via a new `MinFloor` rung in `DecideIdleTimeout` (after the assigned-work rung), mirroring the existing `floorExempt` pattern. Above-floor elastic sessions still idle-reclaim exactly as before; `max_active_sessions` still caps total (the exemption only defers a kill, it never spawns); `DecideMaxSessionAge` is untouched (keep-warm is idle-timeout-only).

**Floor-candidacy is an explicit allow-list** (`{None, Active, Awake, Creating, StartPending}`) — only genuinely-live/starting sessions count toward the floor rank, so a stale dormant/terminal bead (Asleep/Quarantined/FailedCreate/Drained/Archived) can't mask a live floor member out of the exempt set.

**Size:** ~90 production Go + tests. **Verification:** tests assert the floor survives idle-timeout, above-floor elastic still reclaims, no kill→cold-recreate oscillation at steady zero demand, and a stale low-id bead in each of the 4 non-live states does NOT mask a live floor member (non-vacuous — reverting the allow-list fails them). `go build/vet/test ./cmd/gc/ ./internal/session/...` green. Two independent adversarial reviews: both green (the initial deny-list masking gap was caught and closed in review).

Note (transparency): one doc comment slightly overstates the parallel to `countMinActiveCovered` (the two state-sets deliberately differ — cosmetic).

### DIFF
--- a/cmd/gc/session_progress.go
+++ b/cmd/gc/session_progress.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"strings"
 	"time"
 
 	"github.com/gastownhall/gascity/internal/config"
@@ -32,9 +33,17 @@ func openPoolSessionCountForTemplate(infoByID map[string]sessionpkg.Info, cfg *c
 // only StateNone (freshly stamped, not yet transitioned), StateActive and
 // StateAwake (running; Awake is the reconciler healState alias for Active —
 // session_reconcile.go / lifecycle_projection.go RuntimeProjectionAlive),
-// StateCreating, and StateStartPending count. This mirrors countMinActiveCovered's
-// covering-states convention (compute_awake_set.go) rather than diverging into a
-// deny-list.
+// StateCreating, and StateStartPending count.
+//
+// It adopts countMinActiveCovered's allow-list SHAPE (compute_awake_set.go)
+// rather than diverging into a deny-list — but deliberately not its state SET.
+// countMinActiveCovered admits only active/creating (plus an asleep bead an
+// earlier pass already marked desired-awake); this predicate additionally
+// admits Awake, StartPending and StateNone, and never admits asleep. The two
+// answer different questions: this one ranks the CURRENT warm occupancy of one
+// session, so anything on its way to (or already) running counts and a dormant
+// bead never does, while countMinActiveCovered counts coverage of the floor
+// including the asleep sessions this tick's wake pass will revive.
 //
 // Why an allow-list (Doc adversarial re-review, sc-sabwwn): every other state —
 // Asleep, Suspended, Draining, Drained, Archived, FailedCreate, Quarantined,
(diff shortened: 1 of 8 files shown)


quad341 (dismissed):
@csauer02-personal-user — thank you for this one. The diagnosis is the valuable part: you traced the oscillation to `cityStopPoolBeads` gating min-active revival on `sleep_reason=city-stop`, which means an idle-timeout-slept floor session was never revived, only cold-recreated. I checked that against the code and it holds exactly as you described.

Two things I especially appreciated. Flipping `isWarmFloorCandidate` from a deny-list to an explicit allow-list after review caught the leak is the right instinct — a forgotten state now fails closed instead of masking a live floor member. And the pool-managed narrowing, tested in *both* directions: a lower-id named session neither consumes a floor rank nor inherits the exemption. That's the case that would have silently no-op'd the whole fix in a mixed-identity template.

The red CI is not yours. `cmd/gc`'s test binary doesn't compile on `main` right now — `newMemoryOrderDispatcher` grew a `routes` parameter and three test call sites weren't updated. Every `cmd/gc` PR is red on it. I'll fix that on our side, rebase your branch on top, and re-run. Nothing for you to change.

One note for when it lands: a floor member with no `max_session_age` will now stay up indefinitely rather than churning. That's the point of keep-warm, but worth flagging to operators.


---

julianknutsen:

This PR was OPEN/MERGEABLE but behind main, so the merge queue would not take it. Rebased (true rebase, no merge commit) onto `main` @ `b63623d08e`; new head is `3c81fc7a1f`. No conflicts. `behind_by` is now 0.

Verification on the rebased tree: go vet ./cmd/gc/ ./internal/session/ clean; internal/session DecideIdleTimeout/MaxSessionAge/AssignedWorkExhausted PASS; cmd/gc warm-floor + min-floor + idle-timeout reconciler tests PASS (20/20). One commit (`03ff66e`, "pass routes to newMemoryOrderDispatcher in tests") was dropped as already-applied — main carries the same change.

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277171013

julianknutsen:
Fixed the `Preflight / static checks` failure and re-based onto current main (`9979bed7b7`, which now includes #5016).

**Failure:** `make lint-affected` -> `cmd/gc/session_reconciler_test.go:9478:5: QF1001: could apply De Morgan's law (staticcheck)` on `if !(named.ID < pool.ID) {`. It is this branch's own new test code; `lint-affected` only lints packages touched by the diff, which is why it surfaced on the rebase rather than earlier.

**Fix:** `if named.ID >= pool.ID {` — same assertion, no double negative.

**Verification** with the CI-pinned linter (golangci-lint 2.12.0 built with go1.26.6, matching the workflow exactly):
- `golangci-lint run ./cmd/gc/... ./internal/session/...` -> **0 issues**
- Control: reverting just that line reproduces `QF1001` at 9478:5 verbatim, so the clean result is not a false negative.
- `go vet ./cmd/gc/` clean; `TestReconcileSessionBeads_IdleTimeoutMinFloorIgnoresNonPoolSession` PASS.

New head `481b54b8d1`.

PR #5053, opened 2026-08-06
fix(dolt): make the managed dolt config city.toml-authoritative

remuscazacu:
## Summary

A pack command invoked from an operator shell inherited **none** of the city's `[dolt]` block, so `gc dolt restart` regenerated `dolt-config.yaml` from the pack script's own defaults and silently reverted the city's configured server settings.

`gc-beads-bd.sh` rebuilds the managed config purely from `GC_DOLT_*`, defaulting auto-GC **on** and `read_timeout_millis` to **15000** when they are absent (`assets/scripts/gc-beads-bd.sh`, the block around `auto_gc_enabled=true` / `read_timeout_millis=${GC_DOLT_READ_TIMEOUT_MILLIS:-15000}`). Those variables are injected only on the provider-lifecycle path, from the cached city config (`providerLifecycleProcessEnvFromBase`). A directly invoked pack command gets `citylayout.PackRuntimeEnv`, `GC_PACK_*`/`GC_CITY_NAME`, the pinned `GC_BIN`, and `mergeCanonicalScopeDoltEnv` — which projects the canonical *connection* and nothing else.

Net effect: **the same `city.toml` produced two different server configs** depending on who started the server. Measured on a city whose `[dolt]` sets `read_timeout_millis = 120000`, a bare `gc dolt restart` would have rewritten it to `15000` and `wait_timeout` `120` → `30`, undoing connection-resilience settings whose documented failure mode is the controller's long-lived dispatch-pool connections being reaped between reconcile ticks (`begin read tx: invalid connection` → on-demand pools never materialize). Nothing warned, and the drift only became visible by diffing the regenerated file against a pre-restart copy.

Two halves, because the first alone cannot close the gap:

**1. Project the city's `[dolt]` block into pack-command environments.** `applyCityDoltSettingsEnv` carries `archive_level`, `auto_gc_enabled`, `max_connections`, `read_timeout_millis`, `write_timeout_millis`, `wait_timeout_seconds`, and `dolt_lock_release_timeout` — the same set the lifecycle path projects. Resolution is delegated to `resolveManagedDoltConfigForStart` rather than reimplemented, so precedence provably matches the start path: `city.toml` wins where it speaks, ambient `GC_DOLT_*` fills only the gaps it leaves. Nil/zero fields stay unprojected so an unset city value keeps inheriting the caller's environment, and doltlite cities are skipped, mirroring `clearProjectedDoltEnv`.

**2. Add `[dolt] wait_timeout_seconds`.** `wait_timeout` was resolvable *only* from `GC_DOLT_WAIT_TIMEOUT` in the starting process's environment — something no `city.toml` could express and no shell-invoked restart inherited, so the projection had nothing to carry. The renderer and the `dolt-config` doctor check now both prefer the configured value.

That second half also fixes a **false positive** worth calling out on its own: `DoltConfigExpectedValuesForConfig` was city-aware for every value except `wait_timeout`, which it read from the doctor process's own environment. So a city that configures `wait_timeout` reported drift (`system_variables.wait_timeout (got 120, want 30)`) whenever `gc doctor` ran from a shell that does not export the variable — and the finding's remediation hint, *"stop dolt (gc dolt stop) and restart to regenerate managed config"*, pointed at the one action that really would have introduced the drift it claimed to fix.

No linked issue — I did not find an existing one for this after searching open and closed issues for the dolt-config/pack-env surface. Happy to file one if you'd rather track it separately.

## Testing

- [ ] `make check` — started, still running locally when this PR was opened; I will report the result in a comment rather than claim it here.
- [x] `make check-schema` (regenerated `city-schema.json`/`.txt` and `config.md` are committed)
- [x] `go test ./internal/config/ ./internal/doctor/`
- [x] `go test ./cmd/gc/ -run 'DoltConfig|DiscoveredCommand|CityDolt|ManagedDolt|ProviderLifecycle|WaitTimeout|TestNonTestLoadCityConfigCallersPassWarningWriter'`
- [x] `go test ./internal/testpolicy/resourcecensus/` — the ledger guard, unchanged, see the note below
- [x] `gofmt` / `go vet` clean on the touched packages
- [ ] `make check-docs` — not run; the only docs change is generated reference output
- [ ] `make test-integration` — not run locally; no runtime/controller/workflow behavior changes beyond the env projection, but CI's integration suites cover it

Five new tests, each confirmed red against the previous behavior before the fix:

- `TestRunDiscoveredCommand_ProjectsCityDoltSettings` — end to end through a real pack script.
- `TestApplyCityDoltSettingsEnv_CityConfigBeatsAmbient` — precedence, unset-fields-are-not-invented, and no duplicate env keys (a duplicate would leave the shell script reading whichever copy `os.Environ` ordering happened to surface).
- `TestWriteManagedDoltConfigFile_HonorsCityWaitTimeout`.
- `TestManagedDoltWaitTimeoutForConfig` — six cases pinning resolution order, including the negative-value env escape hatch that suppresses the system variable entirely. A config field cannot express that, so it is deliberately preserved and pinned rather than replaced.
- `TestDoltConfigCheck_AcceptsCityConfiguredWaitTimeout` — the doctor false positive above, with the env deliberately unset because that is the operator-shell case.

**On the environment debt ratchet:** `TESTING.md` records that untagged `cmd/gc` environment call/file totals *cannot grow*, so `managedDoltWaitTimeoutForConfig` takes its lookup as a `func(string) string` parameter and the new tests use injected fakes instead of `t.Setenv`. Net new counted environment calls: **0**, and `internal/testpolicy/resourcecensus` passes unchanged — no ledger edit in this PR.

## Checklist

- [x] Updated docs for user-facing changes (generated config reference picks up the new field)
- [x] Called out breaking changes or migration notes — see below

**Migration note, no breaking change.** `wait_timeout_seconds` is additive and optional; omitted or `0` keeps the existing default and the existing `GC_DOLT_WAIT_TIMEOUT` behavior. One deployment-ordering caveat for anyone adopting the field: an older `gc` reading a `city.toml` that sets it emits `warning: unknown field "dolt.wait_timeout_seconds"` (measured, exit 0 — tolerated, not rejected), so cities should upgrade the binary before adding the key, particularly under strict config validation.

The behavior change operators will notice is intentional and is the point of the fix: `gc dolt restart` and `gc dolt start` from a shell now write the same config the supervisor writes. Any city that was unknowingly relying on the pack script's defaults after a manual restart will start getting its own configured values instead.


remuscazacu:
Reporting the `make check` result I said I would, plus one self-correction and two failures I do not believe are mine.

**Pushed a fixup** (force-pushed, no reviews yet): `make check` caught dead code in my own diff — making the wait-timeout resolver injectable orphaned the old `managedDoltWaitTimeout()` wrapper, and `lint-full` flagged it as `unused`. Removed. Auditing for the same class turned up a second one the linter had *not* flagged: `EffectiveWaitTimeoutSeconds` had no callers, and it turns out it cannot have any — `wait_timeout` resolves three ways (config → env → env-selects-omit), so an `Effective*` accessor that collapses "unset" to the default would silently discard the env layer. Rather than delete the idea I kept `DefaultDoltWaitTimeoutSeconds`, documented why it deliberately has no `Effective*` pair unlike its siblings, and pointed the resolver at it, which also removes a duplicated literal. Lint is clean now.

**Two remaining `make check` failures on my machine, both outside this diff.** Flagging them rather than quietly re-running until green:

1. `TestTerminateManagedDoltStartedProcessForceKillsWhenIdentityStable` — `dolt_start_managed_test.go:926: startup cleanup did not SIGKILL a stable-identity child that ignored SIGTERM`. Passes 3/3 in isolation. It is a SIGTERM→SIGKILL escalation *timing* assertion, and my machine was running the full suite alongside two live Dolt servers, so a load-sensitive flake is the obvious reading. This PR touches neither `dolt_start_managed.go` nor any termination path.
2. The suite's dolt leak guard: `leaked 1 dolt sql-server ... TestInitFromWithoutHostedPreservesTemplate/002/city/...`. Also untouched by this diff — `init_from_hosted_dolt_test.go` is not in it. This looks like the known "`go test ./cmd/gc` leaks a dolt sql-server rooted in the test tree" pattern; I have it tracked locally and can file it upstream with the reproduction if you have not already got an issue for it.

So: `make check` is **not** green for me, but neither failure is in a file this PR changes, and everything in its blast radius is green — `internal/config`, `internal/doctor`, `make check-schema`, `go vet`, `gofmt`, `internal/testpolicy/resourcecensus` (unchanged ledger), and the filtered `cmd/gc` surface `DoltConfig|DiscoveredCommand|CityDolt|ManagedDolt|ProviderLifecycle|WaitTimeout`, plus `TestNonTestLoadCityConfigCallersPassWarningWriter`. A clean CI runner is better evidence than my loaded laptop, so I would rather your CI arbitrate than have me keep re-rolling locally. I am re-running the full `cmd/gc` package once more to confirm the flake reading and will follow up if either failure reproduces.

quad341 (approved):
@remuscazacu — this is excellent debugging. "The same `city.toml` produced two different server configs depending on who started the server" is the kind of sentence that makes a whole class of bug visible at once, and delegating resolution to `resolveManagedDoltConfigForStart` instead of reimplementing precedence is exactly right — it keeps the two paths honest as that resolver evolves. Special credit for catching the orphaned `EffectiveWaitTimeoutSeconds` yourself and writing down *why* it can't exist, for keeping the negative-env escape hatch instead of flattening it, and for reporting your unfinished `make check` honestly rather than re-rolling until green.

I'm taking three small things on my side so you don't have to round-trip. `wait_timeout_seconds` isn't in `cityDoltConfigHasLifecycleFields`, so a city that sets *only* that field gets its `[dolt]` registration cleared and the supervisor skips the projection; and adding `GC_DOLT_WAIT_TIMEOUT` to the unconditional strip list drops an ambient value when `city.toml` is silent, which is the one compat case your migration note promises. I'll also wire the field into `ValidateDoltConfig` and add the lifecycle tests. Then it merges, with credit intact. Thank you for this one.


---

PR #5055, opened 2026-08-06
fix(work-record): resolve gc.work_branch against its remote-tracking ref

### LINKED ISSUE #5037: bug: work-record close gate resolves gc.work_branch against the local ref, reporting merged commits as unreachable
### Before you continue

Searched `gastownhall/gascity` issues and PRs (open and closed) on 2026-08-05 for
`work-record`, `work_commit`, `work_branch`, `work_outcome`, `is-ancestor`,
`merge-base`, `work record gate`, `GC_WORK_RECORD_ENFORCE`, `not reachable`,
`refs/heads`, `refs/remotes`, `close gate`. The only directly related hits are the
implementing PR #3586 (`feat(work-record): ADR-0009 claim-lock + work-branch stamp +
structured-close gate`, merged 2026-06-19) and #4768 (`perf(cli): reuse the
write-guard's store read in the bd close gate`), which changes the signature of
`runWorkRecordCloseGate` but not the reachability test. No issue reports this
behaviour.

### Gas City version

`gc version` → `1.4.0` (Homebrew bottle). Verified still present on `main` on
2026-08-05 by reading `cmd/gc/work_record_gate.go` at `HEAD` — line 117 is unchanged
from #3586, and #4768's refactor did not touch it.

### Environment

Ubuntu 26.04 LTS (Linux 7.0.0-28-generic), bash, tmux provider, `bd` 1.1.2
(Homebrew), git 2.53.0, gastown pack 0.4.0
(`f69ec02b39e04b1febc3ced4c47fd4972f706e91`), provider `builtin:claude`. The rig
under observation is a .NET 10 project; nothing about this is language specific.

### Reproduction

The gate's reachability test is one `exec.Command`, in `gitCommitReachableOnBranch`
(`cmd/gc/work_record_gate.go:117`):

```go
return exec.Command("git", "-C", repoDir, "merge-base", "--is-ancestor", commit, branch).Run() == nil
```

`branch` is the raw `gc.work_branch` value — `main`. Git resolves a bare `main` by
`gitrevisions` precedence, which puts `refs/heads/main` ahead of any remote-tracking
ref. So the gate asks about the **local** branch.

In any topology where merges reach `main` over the network, nothing moves the local
ref, and the gate's answer diverges from the truth. Self-contained repro using the
gate's own command:

```bash
#!/usr/bin/env bash
set -eu
R=$(mktemp -d); cd "$R"

git init -q --bare -b main origin.git
git clone -q origin.git clone1 2>/dev/null && cd clone1
git -c user.email=a@b -c user.name=a commit -q --allow-empty -m "base"
git push -q origin HEAD:main
cd "$R"

# A second checkout pushes to origin/main. This is the refinery: it merges in a
# detached worktree against origin/main and pushes.
git clone -q origin.git clone2 2>/dev/null && cd clone2
git checkout -q -B main origin/main
git -c user.email=a@b -c user.name=a commit -q --allow-empty -m "the work that satisfied the bead"
git push -q origin HEAD:main
LANDED=$(git rev-parse HEAD)
cd "$R/clone1"

# clone1 is the worktree recorded in gc.work_dir. Nothing moved its local main.
git fetch -q origin

echo "gc.work_commit         = $LANDED"
echo "local  refs/heads/main = $(git rev-parse refs/heads/main)"
echo "origin/main            = $(git rev-parse refs/remotes/origin/main)"
git merge-base --is-ancestor "$LANDED" main        && echo "main:        reachable" || echo "main:        NOT reachable  <- what the gate asks"
git merge-base --is-ancestor "$LANDED" origin/main && echo "origin/main: reachable  <- the truth" || echo "origin/main: NOT reachable"
```

Output:

```
gc.work_commit         = 37f6f755deccedb330b1e6634dfce64bf5fd6765
local  refs/heads/main = 7f194833efb6a6743b2159444d38c09f70fc0b2c
origin/main            = 37f6f755deccedb330b1e6634dfce64bf5fd6765
main:        NOT reachable  <- what the gate asks
origin/main: reachable  <- the truth
```

#### Why the local ref is never current

This is not operator neglect; it is two correct behaviours composing badly. The
refinery merges in a **detached** temporary worktree against `origin/main` and
pushes, which advances `refs/remotes/origin/main` only. Polecats branch from
freshly-fetched `origin/main` into their own worktrees. Both deliberately decline to
update a branch that is checked out in someone else's worktree — doing so is how you
corrupt a working tree. All the worktrees share one `.git`, so there is exactly one
`refs/heads/main`, and only a human running `git pull` in the rig worktree ever moves
it.

### Expected behavior

`gc.work_commit`, when it is the tip of the branch it was merged to, is reported
reachable, and no warning is emitted.

### Actual behavior

```
gc bd: work-record gate (warn-only): close of ta-2at6: gc.work_commit b14689282cddbe812aab70545693f77665d0347a is not reachable on gc.work_branch main
```

`b146892` was the tip of `origin/main` at that moment — it was not merely *on* `main`,
it *was* `main`. Captured live at the instant of the warning:

```
local  refs/heads/main = 4078286
origin/main            = b146892
git merge-base --is-ancestor b146892 main         -> NO   <- the gate's answer
git merge-base --is-ancestor b146892 origin/main  -> yes  <- the truth
```

Once the local ref is one commit behind, every later merge commit is newer than it, so
the check cannot pass again for a gated bead until someone pulls.

Six consecutive refinery closes in one rig, 2026-07-31 → 2026-08-01. All six carry a
complete three-field work record (`gc.work_outcome=shipped` + `gc.work_commit` +
`gc.work_branch=main`) and all six commits were genuinely on `origin/main`:

| bead | bead type | local `main` | `origin/main` | lag | gate |
|---|---|---|---|---|---|
| ta-n01l | bug | — | 431ad19 | — | silent |
| ta-glfs | task | — | eae8760 | — | warned |
| ta-9dn7 | task | — | 4078286 | — | warned |
| ta-2at6 | task | 4078286 | b146892 | 2 | warned |
| ta-hxuv | bug | 869c95e | 9327181 | 2 | **silent** |
| ta-wx2g | task | 869c95e | 0f914d2 | 3 | warned |

The two silent rows were initially read as evidence of nondeterminism — same lag, same
"not reachable" answer, opposite gate behaviour — which is what made the bug look
intermittent rather than total. **It is not nondeterministic.** Both silent beads are
`type: bug`; all four that warned are `type: task`. The scope predicate, same file
(line 61):

```go
func isWorkRecordGatedBead(bead beads.Bead) bool {
	if t := strings.TrimSpace(bead.Type); t != "" && t != "task" {
		return false
	}
	...
}
```

Six of six observations are explained by that one line, with no appeal to caching or a
second code path.

### Additional context

Three findings, in descending confidence. The first is the report; the other two are
things a fix should not step past.

#### 1. Resolve against the remote-tracking ref (the fix)

Resolve `gc.work_commit` against `refs/remotes/origin/<gc.work_branch>`, falling back
to the bare name only when no remote-tracking ref exists (a purely local repo, where
the local ref *is* the truth). Roughly:

```go
func reachableRef(repoDir, branch string) string {
	remote := "refs/remotes/origin/" + branch
	if exec.Command("git", "-C", repoDir, "rev-parse", "--verify", "--quiet", remote).Run() == nil {
		return remote
	}
	return branch
}
```

Note the interaction with `--is-ancestor`'s existing fail-closed contract: the
function's doc comment says an unknown ref "reads as not reachable", so the `rev-parse
--verify` probe has to be a separate call rather than a fallback on the `merge-base`
exit code — otherwise a genuinely-unreachable commit would be retried against the
local ref and could pass.

**Do not verify a fix by the warning stopping.** Two of the six closes above were
silent while just as wrong. Verify by asserting the check *resolves* against
`refs/remotes/origin/` and *passes* for a commit that is on `origin/main` but not on a
deliberately-stale local `main` — which is exactly what the repro script sets up.

`gitCommitReachableOnBranch` currently has **no test coverage at all** — the name does
not appear anywhere in `cmd/gc/work_record_gate_test.go` (checked at `HEAD`,
2026-08-05: zero references across 441 lines). Every case there calls
`validateWorkRecordOnClose` with one of the injected oracles `alwaysReachable` /
`neverReachable`, so no test executes the real git command. That is presumably why
review of #3586 caught the detached-HEAD and leading-dash cases but not the ref
resolution.

#### 2. `type: bug` beads are silently exempt

`isWorkRecordGatedBead` gates on `Type != "task"`. `bug` is as worker-claimable as
`task`, and a polecat implementing a bug bead produces exactly the artifact the work
record exists to capture — but the gate never looks at it. The unit tests cover
`task`, `convoy`, and `message` only (verified at `HEAD`: no `Type: "bug"` case
exists), and the doc comment's stated intent is to exclude *control/structural* beads
(`gc.kind` set) and non-work bead types — which reads as though `bug` was overlooked
rather than deliberately exempted. If it is deliberate, the comment and a test should
say so; the `gc.kind` check alone already excludes the control beads the comment is
worried about.

Worth fixing *with* finding 1, not after: fixing the resolution alone makes the gate
correct for `task` beads and leaves it silent for `bug` beads, which preserves the
"looks healthy" property in a narrower form.

#### 3. `repoDir` is a path that gets deleted

```go
repoDir := strings.TrimSpace(bead.Metadata[beadmeta.WorkDirMetadataKey])
if repoDir == "" {
	repoDir = scopeRoot
}
```

`gc.work_dir` on a polecat bead points at that polecat's per-bead worktree, which is
transient. All four `gc.work_dir` paths from the table above are gone from disk today.
Any close, re-close, or retrospective audit that runs after cleanup gets a git error,
which the function's fail-closed contract turns into "not reachable" — a false
negative from a *missing directory*, indistinguishable in the output from a real one.
Preferring a stable root (the rig clone) with `gc.work_dir` as the fallback would
remove that. Worktrees share one object store, so the stable root answers the same
question.

#### Adjacent, and deliberately not filed as a duplicate

[`gastownhall/beads#4960`](https://github.com/gastownhall/beads/issues/4960) —
*"Feature: pre-close landing-verification check for `bd close`"* — is the closest
thing in either repo, and it is a sibling rather than the same bug. Different product
(it asks `beads` for a check it does not have; this is a defect in a check `gascity`
already ships) and different failure (its concern is that `--is-ancestor` is the wrong
*predicate* under squash-merge; this is that the predicate's right-hand *operand*
resolves to a ref nobody maintains). Fixing either leaves the other standing.

rjgeng:
## Problem

`gc bd close`'s work-record gate resolves `gc.work_branch` (e.g. `main`)
by checking reachability against the **local** `refs/heads/<branch>`.
In a refinery/polecat topology a merge lands via a push from a
*different* worktree — advancing the remote-tracking ref
(`refs/remotes/origin/<branch>`) but never the local one checked out
elsewhere. A genuinely-landed commit then reads as "not reachable"
until something happens to fast-forward the local branch, which in
that topology may be never.

Six real closes in #5037's own history show this happening.

## Fix

`gitCommitReachableOnBranch` (`cmd/gc/work_record_gate.go`) now probes
`refs/remotes/origin/<branch>` first (a separate `git rev-parse
--verify --quiet` call, so a genuinely-unreachable commit is never
retried against the local ref) and only falls back to the bare branch
name when no such remote-tracking ref exists (purely local repos).

The ref-selection decision (prefer remote-tracking, fall back to bare
name) is extracted into `preferredReachabilityRef`, a pure function
taking an injected `remoteRefResolves` closure — matching this file's
existing oracle-injection pattern (`validateWorkRecordOnClose`). This
keeps the new logic covered by hermetic table tests rather than tests
that spin up real git repos, so it adds zero new subprocess call sites
(the resource census in `internal/testpolicy/resourcecensus` scans
test-file subprocess/sleep/etc. call sites; the underlying `git
rev-parse` call in production code was already outside its scope, but
building real git repos in tests would not have been).

Had zero existing test coverage before this change.

## Not fixed in this PR (flagging as known follow-ups)

- `type: bug` beads are silently exempt from this gate — same root
  cause class as this fix, but changing the scope predicate is a
  policy call about which bead types the contract applies to, not a
  mechanical fix.
- `gc.work_dir` points at a transient, often-deleted worktree —
  preferring a stable rig-root path would remove a separate
  false-negative source, but is a bigger architectural change.

Neither is blocking for this fix.

Closes #5037.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -116,6 +116,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   publish that projection is left stamped rather than risk a wrong clear, as
   are cross-store targets this reconciler tick cannot reach. (gascity#4373)
 
+- **The work-record close gate now resolves `gc.work_branch` against its
+  remote-tracking ref, not the local branch alone.** `gitCommitReachableOnBranch`
+  passed the bare branch name (e.g. `main`) straight to
+  `git merge-base --is-ancestor`; gitrevisions precedence resolves a bare name
+  to the local `refs/heads/<branch>` ahead of any remote-tracking ref. In a
+  refinery/polecat topology, merges land via a push from a *different*
+  worktree — advancing `refs/remotes/origin/<branch>` but never the local ref
+  checked out elsewhere — so a genuinely-landed commit read as unreachable
+  until something happened to fast-forward the local branch, which in that
+  topology may be never. The gate now prefers `refs/remotes/origin/<branch>`
+  when it resolves, falling back to the bare name for purely local repos with
+  no `origin` remote. (gascity#5037)
+
 - **The dolt pack's `run_bounded` python3 fallback now sends SIGTERM before
   SIGKILL, matching its documented contract.** The fallback (used when
   neither `timeout` nor `gtimeout` is on `PATH`, the default on stock macOS)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -125,9 +125,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   worktree — advancing `refs/remotes/origin/<branch>` but never the local ref
   checked out elsewhere — so a genuinely-landed commit read as unreachable
   until something happened to fast-forward the local branch, which in that
-  topology may be never. The gate now prefers `refs/remotes/origin/<branch>`
-  when it resolves, falling back to the bare name for purely local repos with
-  no `origin` remote. (gascity#5037)
+  topology may be never. The gate now checks `refs/remotes/origin/<branch>`
+  first when it resolves, and still falls back to the bare branch name — so a
+  commit is reachable if it is on either ref. Purely local repos with no
+  `origin` remote are unaffected, and a commit that has been committed locally
+  but not yet pushed continues to satisfy the gate as it did before.
+  (gascity#5037)
 
 - **The dolt pack's `run_bounded` python3 fallback now sends SIGTERM before
   SIGKILL, matching its documented contract.** The fallback (used when
--- a/cmd/gc/work_record_gate.go
+++ b/cmd/gc/work_record_gate.go
@@ -148,6 +148,19 @@ func validateWorkRecordOnClose(bead beads.Bead, commitReachable func(commit, bra
 // repo, unknown ref, unknown commit — reads as "not reachable". A commit/branch
 // that looks like a flag (leading "-") is rejected outright so a malformed
 // metadata value can never be parsed as a git option.
+//
+// branch is resolved against refs/remotes/origin/<branch> when that ref
+// exists, falling back to the bare name otherwise. gitrevisions precedence
+// puts a local refs/heads/<branch> ahead of any remote-tracking ref, but the
+// worktree gc.work_dir names is rarely the one whose local branch tip
+// actually moves — a refinery or polecat that merges/pushes from a different
+// worktree advances the remote-tracking ref, never the local one checked out
+// elsewhere. Resolving against the local ref alone reports a landed commit as
+// unreachable until something happens to fast-forward it, which in that
+// topology may be never (gastownhall/gascity#5037). The rev-parse probe is a
+// separate call rather than a fallback on the merge-base exit code so that a
+// genuinely-unreachable commit is never retried against the local ref and
+// allowed to pass.
 func gitCommitReachableOnBranch(repoDir, commit, branch string) bool {
 	if strings.TrimSpace(repoDir) == "" || commit == "" || branch == "" {
 		return false
@@ -155,7 +168,11 @@ func gitCommitReachableOnBranch(repoDir, commit, branch string) bool {
 	if strings.HasPrefix(commit, "-") || strings.HasPrefix(branch, "-") {
 		return false
 	}
-	return exec.Command("git", "-C", repoDir, "merge-base", "--is-ancestor", commit, branch).Run() == nil
+	ref := branch
+	if remote := "refs/remotes/origin/" + branch; exec.Command("git", "-C", repoDir, "rev-parse", "--verify", "--quiet", remote).Run() == nil {
+		ref = remote
+	}
+	return exec.Command("git", "-C", repoDir, "merge-base", "--is-ancestor", commit, ref).Run() == nil
 }
 
 // workRecordCloseTargets returns the bead IDs a bd invocation closes, and
--- a/cmd/gc/work_record_gate.go
+++ b/cmd/gc/work_record_gate.go
@@ -156,13 +156,36 @@ func validateWorkRecordOnClose(bead beads.Bead, commitReachable func(commit, bra
 // remoteRefResolves is injected so the decision is unit-testable without a
 // real git repository; the only production caller runs
 // `git rev-parse --verify --quiet <ref>`. Kept as a separate probe rather
-// than a fallback on the merge-base exit code so that a genuinely-unreachable
-// commit is never retried against the local ref and allowed to pass.
+// than a fallback on the merge-base exit code so the caller can distinguish
+// "no such ref" from "not reachable"; see commitReachableOnEitherRef.
 func preferredReachabilityRef(branch string, remoteRefResolves func(ref string) bool) string {
 	if remote := "refs/remotes/origin/" + branch; remoteRefResolves(remote) {
 		return remote
 	}
 	return branch
+}
+
+// commitReachableOnEitherRef reports whether a commit is reachable from the
+// branch's remote-tracking ref OR from the branch itself. The remote-tracking
+// ref is probed first (see preferredReachabilityRef) because it is the ref
+// that actually advances in a refinery/polecat topology; the bare branch name
+// is then still checked, because ADR-0009's contract is that the commit is
+// reachable on gc.work_branch — not that it has been pushed. Checking only the
+// remote ref would reject a commit that is genuinely on the local branch but
+// sits ahead of (or was never pushed to) origin, a false negative in the exact
+// mirror image of gastownhall/gascity#5037.
+//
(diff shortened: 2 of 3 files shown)


rjgeng:
Heads-up on the red CI here: `cmd/gc process / shard 1 of 12` is failing on `TestCmdStopSupervisorManagedInvalidCityTomlFailsWhenShutdownFails`, which is unrelated to this diff. Confirmed it fails identically and deterministically on plain `main` at the current tip in a clean worktree with none of this PR's changes applied — filed as #5668. Current head here is `6b9be528`.

quad341 (approved):
@rjgeng — this is genuinely good debugging. The gitrevisions detail is the part almost everyone misses: a bare `main` resolves to `refs/heads/main` ahead of any remote-tracking ref, so in a topology where the merge lands from a *different* worktree, the one ref you're testing against is the one ref that never moves. Citing the six real closes out of #5037's own history made the case immediately, and I appreciate that you extracted `preferredReachabilityRef` as a pure function with an injected resolver instead of reaching for a git fixture — that matches how `validateWorkRecordOnClose` is already tested in that file, and I checked, the `resourcecensus` subprocess ratchet you were working around is real and live.

One thing I want to adjust before this lands, and I'll do it on my side: right now, when `refs/remotes/origin/<branch>` resolves, we stop there and never look at the local branch. So a commit that's on the local branch but ahead of (or absent from) the pushed tip would newly read as unreachable — the mirror image of the bug you fixed. Since `gc.work_branch` is stamped from the worktree's current branch at claim time, that's often a `polecat/*` branch where the ordering is easy to hit. ADR-0009's contract is "reachable on `gc.work_branch`," not "pushed," so I'm making it pass on **either** ref — which fixes #5037 and keeps every case that passed before. Your `preferredReachabilityRef` and all three of your table cases stay exactly as you wrote them; I'm just wrapping them.

I'll add the union test, tweak the CHANGELOG wording, re-run `cmd/gc`, and merge. The diagnosis and the fix are yours. Thanks also for flagging the two follow-ups in the body — agreed both are policy calls, not this PR.


---

PR #5056, opened 2026-08-06
test(cmd/gc): tear down the real tmux server the named-session fixture starts

### LINKED ISSUE #4656: bug: go test ./cmd/gc leaks an unreachable tmux server per TestCmdNudgeStatusJSON run (socket root removed while server runs)
## Summary

Every run of `go test ./cmd/gc` leaves behind a live `tmux` server that no client can reach. Four had accumulated on this machine across a day of `make check` runs. They survive the test binary exiting and persist until killed by PID or until reboot.

They are **unreachable, not merely idle** — `tmux -L test-city ls` finds nothing, because the socket they listen on has been unlinked. `tmux kill-server` therefore cannot clean them up.

## Observed

Four orphans, all from `TestCmdNudgeStatusJSON`, from two different worktrees of this repo:

| PID | Started | Worktree | Test city dir |
|---|---|---|---|
| 219957 | 18:03 | fork worktree | `/tmp/gct176291-…/TestCmdNudgeStatusJSON1481865426/001` — deleted |
| 1392436 | 22:27 | `upstream/main` @ `2b265f75c` | `/var/tmp/gct1263563-883873642/…` — **339 MB still on disk** |
| 1493747 | 22:42 | `upstream/main` @ `2b265f75c` | `/var/tmp/gct1450668-…` — deleted |
| 1856490 | 23:37 | `upstream/main` @ `2b265f75c` | `/var/tmp/gct1718198-…` — deleted |

Each command line is the same shape:

```
tmux -u -L test-city new-session -d -s mayor -c <testdir> \
  -e GC_AGENT=mayor -e GC_SESSION_NAME=mayor -e GC_PROVIDER=codex … \
  env -u CLAUDECODE … echo
```

~3 MB RSS each, 0% CPU. All four died on a plain `SIGTERM` once located by PID.

## Mechanism

1. `cmd/gc/cmd_session_test.go` `writeNamedSessionCityTOML` writes `.gc/site.toml` with `workspace_name = "test-city"`, plus an `agents/mayor/agent.toml` whose `start_command = "echo"`.
2. gc derives the tmux socket name from the workspace name — `internal/runtime/tmux/tmux.go:302` appends `-L <SocketName>`. So the socket is literally `test-city`.
3. `cmdNudgeStatus` on this path reaches the **real** tmux adapter, not `runtime.NewFake()`, and a real detached server is spawned.
4. Nothing tears that server down. `TestMain` then removes the tmux socket root it created (`os.RemoveAll(tmuxSocketCleanupRoot)`, `cmd/gc/main_test.go` ~line 250 onward), which deletes `TMUX_TMPDIR` — `/tmp/gct-<pid>-<rand>/tmux` — out from under the still-running server.

Step 4 is what makes these unkillable-by-name: the listening socket becomes an unlinked inode, the server holds an fd on it forever, and no client can bind to that path again.

### Why it outlives an instantly-exiting `echo` — hypothesis, not confirmed

The pane command is `echo`, so a default-configured server should exit with its last session. The adapter has explicit remain-on-exit corpse handling (`tmux.go:2228`, `:2250`, `:2295`), so `remain-on-exit` is plausibly enabled and the dead pane keeps the session — and therefore the server — alive. I read this from the code and did not instrument it; worth confirming before fixing on that basis.

## The asymmetry with dolt

`cmd/gc`'s `TestMain` already has a process-level leak guard for the *other* long-lived child this package spawns — `doltLeakGuardedTestingM` (`cmd/gc/path_helpers_test.go:103`), added for #1988. It snapshots processes before and after the run, diffs, reports, reaps, and **fails the run with exit 1** if anything leaked, with a signal handler for abnormal exits.

That guard enumerates only `discoverDoltProcesses`. tmux gets *directory*-level hygiene — the `.gc-test-alive.lock` flock sentinel and `testOrphanSweepMinAge` in `cmd/gc/test_orphan_sweep_test.go` — but no process-level guard at all. So a leaked tmux server is silent, where a leaked dolt server would have failed the build.

Secondary evidence of the dir side also not closing cleanly: 17 `/tmp/gct-<pid>-<rand>/` socket roots were left behind (each holding an empty `tmux/` and a 0-byte `.gc-test-alive.lock`), with none of their owning PIDs alive. They were younger than `testOrphanSweepMinAge` (1h), so no later run would sweep them either.

## Suggested fix

The pattern already exists in-repo, so this is mostly a matter of applying it on this path:

- `internal/runtime/tmux/nudge_poke_integration_test.go` registers `t.Cleanup(func() { _, _ = tm.run("kill-server") })`.
- The adapter exposes `TeardownServer`, which calls `kill-server` (pinned by `internal/runtime/tmux/lifecycle_test.go:51`).

Two levels to consider:

1. **Narrow** — register the `kill-server` cleanup in the shared helper that configures the real tmux runtime for `cmd/gc` tests, rather than per-test, so every test driving the real adapter gets it. Ordering matters: it must run *before* `TestMain` removes the socket root, or the cleanup client cannot find the socket either.
2. **Broad** — teach the existing guard to enumerate tmux servers whose socket lives under the run's socket root, so a leak fails the run the way a dolt leak does. This is the version that stops the next variant of this from going unnoticed.

## Not fork-specific

Reproduced from a fork worktree *and* from `upstream/main` @ `2b265f75c` (plus an unrelated one-line `internal/runproj` change from #4649). Nothing fork-specific is involved — the leaking test and the socket-root teardown are both upstream code.

## Duplicate check

Searched issues (open and closed) for `tmux in:title`, `leak in:title`, `orphan tmux server`, `kill-server cleanup`, `TestCmdNudgeStatusJSON`, `test-city`. No existing issue covers this. Closest relatives, all distinct:

- **#1988** (closed) — *test harness leaks dolt sql-server processes across runs*. Same shape, different resource; its fix is the guard described above, which is exactly what tmux lacks.
- **#3640** (closed) — *Dolt leak-guard not applied to real-dolt integration/acceptance TestMains*. Same "guard exists but doesn't cover this path" gap, for dolt.
- **#3896** (closed) — *test/e2e harnesses leak persistent launchd supervisor agents*. Same family on macOS/launchd.
- **#2073** (open) — *rollbackPendingCreate leaks tmux+claude pane*. A tmux leak, but in the runtime path rather than the test harness; different cause.

## Environment

- `upstream/main` @ `2b265f75c`
- Linux 7.0.3-arch1-2, x86_64, 2 cores / 4 threads
- Go toolchain 1.26.5 (via `go.mod`; system `go` is 1.26.4), `CGO_ENABLED=1`, `-tags gms_pure_go`
- tmux 3.6b
- Reproduced under `make check` (`-p=4`) and visible after any `go test ./cmd/gc` run

## Sweeping them, for anyone who finds theirs

`tmux kill-server` will not work — the sockets are gone. Locate by PID and signal directly:

```
ps -eo pid,args | grep 'tmux .*-L test-city'
kill <pids>
```

remuscazacu:
Fixes #4656.

`go test ./cmd/gc` leaves a live tmux server behind on every run of a test using `writeNamedSessionCityTOML`. The servers outlive the test binary and persist until killed or reboot.

## Mechanism

The fixture writes `.gc/site.toml` with `workspace_name = "test-city"`. gc derives the tmux socket from the workspace name, so a test reaching the real tmux adapter drives a real server on `tmux -L test-city` — not `runtime.NewFake()`.

Nothing tore it down. The test only called `t.TempDir()`, and Go's cleanup then removed `TMUX_TMPDIR` out from under the running server. That unlinks the socket inode, so the server becomes **unreachable**: `tmux -L test-city ls` finds nothing and no client can ever attach or kill it by name. It just sits there.

The test passing is what let this accumulate — nothing surfaces the leak. On this machine 5 orphans had built up (ages 6–23h, 26.5 MB), every one with its socket dir already deleted.

## Fix

Register the teardown inside the shared fixture rather than per test, so all 29 call sites are covered and a new caller cannot forget it.

The cleanup ordering is load-bearing. Callers take `dir` from `t.TempDir()` **before** calling the fixture, so Go's LIFO cleanup order runs this teardown while `TMUX_TMPDIR` still exists. Reverse it and the `kill-server` client cannot find the socket either — the trap this fix has to avoid rather than reproduce. That reasoning is in a comment so it survives future edits.

`site.toml`'s `workspace_name` now derives from the same constant the teardown uses, so the socket name and workspace name cannot drift apart.

## Verification

The test result never reflected the leak, so it is counted directly:

| | result | `tmux -L test-city` servers |
|---|---|---|
| `main` @ `489309232`, `-run '^TestCmdNudgeStatusJSON$'` | ok | 0 → **1** |
| this branch, same test | ok | 0 → **0** |
| this branch, **full** `cmd/gc` package | see note | 0 → **0** |

`go vet ./cmd/gc` clean, `gofmt` clean.

## Two notes for reviewers

**The full-package run exits 1 for an unrelated, pre-existing reason.** Zero `--- FAIL` lines — every test passes — but the *dolt* leak guard correctly fires on a different test:

```
cmd/gc test dolt leak guard: leaked 1 dolt sql-server process(es) under <tempRoot>
  pid=4010752 argv="dolt sql-server --config <tempRoot>/TestInitFromWithoutHostedPreservesTemplate.../002/city/.gc/runtime/packs/dolt/dolt-config.yaml"
```

This reproduces on `main` without my change, which touches only a tmux teardown and cannot affect an init test's dolt lifecycle. Worth its own issue; happy to file one.

**One clause of the issue's acceptance criteria is arguably wrong.** #4656 asks that no `/tmp/gct-*` dirs remain after a run. 18 do — but they are not the run's own (that one is cleaned up); they belong to helper `gc` processes the suite spawns, hold only the `.gc-test-alive.lock` sentinel, and contain no sockets or live servers. `tmuxtest` reaps them by **age** (`socketParentSweepMinAge = 1 hour`) deliberately, so a concurrent sibling run's still-active dir is not reclaimed. Satisfying that clause literally would break the sibling-run safety property, so I have left it alone.

Separately, the issue describes tmux as having "directory-level hygiene only". `test/tmuxtest` in fact already implements a full three-layer guard (TestMain pre-sweep, per-test cleanup, post-sweep) — it simply only recognises `gctest-*` sockets, and this fixture's socket is `test-city`, outside that convention. So the broad follow-up has two possible shapes: add a process-level tmux guard to cmd/gc's `TestMain` mirroring `doltLeakGuardedTestingM`, or move this fixture onto a `gctest-*` socket so the existing sweeps cover it. This PR is the narrow fix only; I did not want to pick that direction unilaterally.

https://claude.ai/code/session_015aSHpYfaSAz43US5yjWuLz


AJBcoding:
Independent corroboration from downstream (gas-iio), in case it helps this land: this fixes a leak we were tracking, and I found field evidence that specifically validates the ordering argument in your `killNamedSessionTmuxServer` comment.

**The leak is real and accumulates.** Our box was carrying **22** leaked `tmux -u -L test-city new-session -d -s mayor` servers — one per full `go test ./cmd/gc` run — including one still alive from the previous day. Every one of those runs reported `ok`, because the `TestMain` leak guard scopes to dolt processes only, so tmux residue never fails the package.

**Your ordering caveat is load-bearing, and I can show it.** Those orphaned servers were unreachable by socket name:

```
$ tmux -L test-city kill-server
error connecting to /private/tmp/tmux-501/test-city (No such file or directory)
```

The tests point `TMUX_TMPDIR` at their own temp root, so once `t.TempDir()` removes that tree the socket is an unlinked inode and no client can find the server by name — it can only be killed by PID. That is exactly the failure mode your comment describes if the cleanup ran after `TempDir`'s. The LIFO ordering you rely on (callers take `dir` from `t.TempDir()` *before* calling the fixture) is what keeps the teardown able to reach the server at all.

**Coverage looks right to me.** The test we observed leaking, `TestCmdNudgeStatusJSON`, calls `writeNamedSessionCityTOML`, so it inherits the teardown; the fixture has 29 call sites, so this is fixture-wide rather than test-specific. A full-package run on our side surfaced no second leaking path.

One thing this deliberately does not cover, which we've filed separately on our side rather than expanding your diff: the guard itself still watches only dolt, so a future tmux leak from a path that bypasses this fixture stays invisible.

quad341 (approved):
@remuscazacu — this is the good kind of test fix: you chased a leak that never once turned a test red, then landed it in the shared fixture so all 29 call sites are covered instead of playing whack-a-mole per test.

We verified it rather than trusting the writeup. With your cleanup disabled, one `tmux -L test-city` server leaked per run; with it, zero. `go vet`, `gofmt`, and the census policy tests are clean, and the 6/2 → 7/3 and 0/0 → 1/1 ratchet bumps are exact across the Go policy, the TOML manifest, and the generated table.

Two non-blocking notes. Your *code comment* has the better explanation — `TestMain` holds `TMUX_TMPDIR` process-wide until after `m.Run()`, so the socket is always reachable — which is stronger than the LIFO-ordering story still in the PR description; worth updating the body. And you were right to leave the `/tmp/gct-*` clause alone: the hour-long sweep age exists precisely to protect concurrent sibling runs.

The red check is a provider-ledger waiver that expired today and fails on `main` too — not yours. We'll sort that separately and merge this. Please do file the dolt-leak-guard issue.


---

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277169983

julianknutsen:

This PR was OPEN/MERGEABLE but behind main, so the merge queue would not take it. Rebased (true rebase, no merge commit) onto `main` @ `b63623d08e`; new head is `1be9edd0be`. No conflicts. `behind_by` is now 0.

Verification on the rebased tree: go vet ./cmd/gc/ ./internal/testpolicy/resourcecensus/ clean; internal/testpolicy/resourcecensus full package PASS (the 7/3 and 1/1 tmux ledger rows still match the tree); cmd/gc test binary builds. 

quad341:
Re-reviewed `7257dca14`..`3d1ce9e72` (53 new commits since the last review): no new concerns — the prior review stands.

Issue #4673, opened 2026-07-26
bug: go test ./cmd/gc leaks a dolt sql-server rooted in the source checkout — invisible to the temp-root-scoped dolt leak guard


remuscazacu:
## Summary

`go test ./cmd/gc` can leave a live `dolt sql-server` whose config and data dir are **inside the source checkout** — `cmd/gc/.gc/runtime/packs/dolt/` and `cmd/gc/.beads/dolt`. The package's own dolt leak guard cannot catch it, because the guard only counts servers whose `--config` path is under the run's temp root.

So a dolt server leaked into `/tmp` fails the build, and one leaked into the repo does not.

This is a third-round follow-up to #1988 and #2306 and is a different gap from both — details in *Relationship to prior issues* below.

## Evidence

Four `dolt sql-server` processes on a machine that should have exactly two (one central server per city):

| PID | Port | `--config` root | Verdict |
|---|---|---|---|
| 287738 | 49723 | `<city>/.gc/runtime/…` | legit — city central, 940 MB |
| 285291 | 49990 | `<other city>/.gc/runtime/…` | legit — city central, 879 MB |
| 187488 | 13968 | `<repo-worktree-A>/cmd/gc/.gc/runtime/…` | **leak**, 15h old, 22 MB |
| 1473825 | 48750 | `<repo-worktree-B>/cmd/gc/.gc/runtime/…` | **leak**, 10h old, 60 MB |

Both legit ports matched each city's `.beads/dolt-server.port`. Both leaks held **no user databases** (`SHOW DATABASES` returned only `information_schema` and `mysql`), so they were pure orphans, not something still in use. Both died on a plain `SIGTERM`.

Their config confirms where they were rooted:

```yaml
# <worktree>/cmd/gc/.gc/runtime/packs/dolt/dolt-config.yaml
  port: 48750
  host: 127.0.0.1
data_dir: "<worktree>/cmd/gc/.beads/dolt"
```

That is the *package directory* — the test binary's working directory — not a `t.TempDir()`.

## Why the guard misses it

`cmd/gc`'s `TestMain` wraps the run in `doltLeakGuardedTestingM` (`cmd/gc/path_helpers_test.go:103`): it snapshots dolt processes before and after, diffs, reports, reaps, and fails the run with exit 1, with a signal handler for abnormal exits. Good machinery. But the snapshot filter is:

```go
// cmd/gc/path_helpers_test.go:321
[...]

PR #5057, opened 2026-08-06
test(cmd/gc): let the dolt leak guard see servers rooted in the checkout

remuscazacu:
Refs #4673.

The dolt leak guard wired into `cmd/gc`'s `TestMain` snapshots dolt sql-servers before and after the run, diffs them, reaps the difference and fails the run. It filtered candidates to a single root:

```go
if root == "" || !pathutil.PathWithin(root, configPath) { continue }
```

`root` is the run's temp root. A test that starts the managed dolt provider with the city root resolved to the **package directory** instead of a `t.TempDir()` puts both the process and its `data_dir` inside the source checkout:

```
--config  cmd/gc/.gc/runtime/packs/dolt/dolt-config.yaml
data_dir  cmd/gc/.beads/dolt
```

That path is structurally outside the temp root, so `PathWithin` is false and the guard skipped it **in silence**. A dolt server leaked into `/tmp` fails the build; one leaked into the repo did not. `.gitignore` covers `.gc/` and `**/.beads/*`, so `git status` stayed clean while these accumulated — 5 of 6 local worktrees had collected the state, and one server had been running 17.7h holding a listening port and ~28 MB.

## Why a second root, and not an unscoped guard

The issue suggests snapshotting all dolt servers and treating any new one not owned by a registered city as a leak. Implemented literally that is unsafe, and the reason is worth being explicit about: **the same snapshot function feeds `reapDoltProcessesUnderRoot`**, which kills every process it returns and runs both as the startup sweep and from the signal handler. A guard watching every dolt server on the machine would reap the developer's own city servers, which legitimately start and stop during a long test run. The "not owned by a registered city" clause is doing load-bearing work the guard has no way to evaluate — it has no registry of cities.

Naming the checkout as a second root is precise instead. No real city lives inside a source checkout, so a dolt server rooted there is a test leak by construction, which makes it safe for the reaping paths and not merely for detection.

The test binary's cwd **is** the package directory, so `os.Getwd()` in the guard constructor yields exactly that root. Capturing it once at construction also makes it immune to any later `os.Chdir` by a test.

Empty roots continue to match nothing, preserving the old `root == ""` behaviour. That property is load-bearing given the reaping paths: a root that failed to resolve must narrow the guard, never widen it to every server on the host. `os.Getwd()` failure is therefore non-fatal and degrades to the previous temp-root-only behaviour.

## Tests

- `TestSnapshotDoltProcessesForConfigRootsCatchesSourceTreeLeak` — the source-tree leak is caught, the temp-root leak still is, and a real city server outside both roots is **not** snapshotted (the reaper-safety property). It also pins the pre-fix single-root behaviour, so a regression back to one root is caught.
- `TestSnapshotDoltProcessesForConfigRootsIgnoresEmptyRoots` — empty roots match nothing.

`go test ./cmd/gc -run 'TestSnapshotDoltProcessesForConfigRoot|TestDoltLeakGuarded|TestDiffDoltProcessSnapshots'` → 6 PASS / 0 FAIL, including the 4 pre-existing guard tests. `go vet ./cmd/gc` clean, `gofmt` clean.

## Scope: this makes the leak visible, it does not stop it

This is the broad half of #4673 only. The narrow half — stopping whichever test resolves the city root to the package directory — is **not** fixed here, and #4673's "no `cmd/gc/.gc` or `cmd/gc/.beads` dir is created" criterion is therefore not yet met.

That is deliberate: the culprit test was never identified, and the issue notes a bisect is expensive on this package. With the guard widened it no longer needs one — the run now fails and `writeDoltLeakReport` prints the leaked pid and full argv, so the suite names the culprit itself on the next full run.

A lead for whoever takes the narrow fix. A full `cmd/gc` run on `main` created `cmd/gc/.gc/events.jsonl` and `cmd/gc/.gc/scripts/gc-beads-bd.sh` in the checkout, but no dolt config and no server — the precursor, not the full leak. The file that appears is the **beads provider script**, and `clearInheritedBeadsEnv` in this same package already documents that chain: inherited `GC_BEADS=bd` triggers `gc-beads-bd.sh` and leaks an orphan dolt sql-server "because test cleanup paths do not call `shutdownBeadsProvider`". That predicts the leak is intermittent and env-dependent, which would explain why a single culprit test was never pinned.

## Unrelated pre-existing failure you will hit

`go test ./cmd/gc` cannot exit 0 on `main` today, independent of this change: zero `--- FAIL` lines, every test passes, and the exit comes from the guard correctly catching a *different* leak.

```
pid=4010752 argv="dolt sql-server --config <tempRoot>/TestInitFromWithoutHostedPreservesTemplate.../002/city/.gc/runtime/packs/dolt/dolt-config.yaml"
```

That one lands under the temp root, so the existing guard already sees it — the mirror image of this issue. Happy to file it separately.

https://claude.ai/code/session_015aSHpYfaSAz43US5yjWuLz

### DIFF
--- a/cmd/gc/dolt_leak_helper_test.go
+++ b/cmd/gc/dolt_leak_helper_test.go
@@ -632,3 +632,74 @@ func TestDoltLeakGuardedTestingMRunWithSweepsOrphanDirsAtStartup(t *testing.T) {
 		t.Fatalf("call order = %v, want [sweepStale sweepOrphanDirs runTests ...]", order)
 	}
 }
+
+// A dolt sql-server rooted in the source checkout rather than under the run's
+// temp root is the leak shape the guard used to miss entirely: cmd/gc's own
+// package dir is structurally outside tempRoot, so PathWithin(tempRoot, cfg)
+// was false and the process was skipped in silence. .gitignore hides the
+// matching cmd/gc/.gc and cmd/gc/.beads dirs, so nothing surfaced it.
+func TestSnapshotDoltProcessesForConfigRootsCatchesSourceTreeLeak(t *testing.T) {
+	tempRoot := filepath.Join("/tmp", "gct12345-678")
+	sourceRoot := filepath.Join("/home", "dev", "gascity", "cmd", "gc")
+
+	underTemp := DoltProcInfo{
+		PID:  1001,
+		Argv: []string{"dolt", "sql-server", "--config", filepath.Join(tempRoot, "TestX", "001", ".gc", "runtime", "packs", "dolt", "dolt-config.yaml")},
+	}
+	underSource := DoltProcInfo{
+		PID:  1002,
+		Argv: []string{"dolt", "sql-server", "--config", filepath.Join(sourceRoot, ".gc", "runtime", "packs", "dolt", "dolt-config.yaml")},
+	}
+	// A real city server on the same machine must stay out of the snapshot: the
+	// reaping paths kill everything the snapshot returns, so widening the guard
+	// must not put a developer's own city at risk.
+	realCity := DoltProcInfo{
+		PID:  1003,
+		Argv: []string{"dolt", "sql-server", "--config", filepath.Join("/home", "dev", "city", ".gc", "runtime", "packs", "dolt", "dolt-config.yaml")},
+	}
+
+	enumerate := func() ([]DoltProcInfo, error) {
+		return []DoltProcInfo{underTemp, underSource, realCity}, nil
+	}
+
+	got, err := snapshotDoltProcessesForConfigRoots(enumerate, []string{tempRoot, sourceRoot})
+	if err != nil {
+		t.Fatalf("snapshotDoltProcessesForConfigRoots: %v", err)
+	}
+	if _, ok := got[1002]; !ok {
+		t.Errorf("source-tree leak PID 1002 missing from snapshot: %#v", got)
+	}
+	if _, ok := got[1001]; !ok {
+		t.Errorf("temp-root leak PID 1001 missing from snapshot: %#v", got)
+	}
+	if _, ok := got[1003]; ok {
+		t.Errorf("real city server PID 1003 must not be snapshotted (the reaper kills these): %#v", got)
+	}
+
+	// The pre-fix behaviour, pinned so a regression to a single root is caught.
+	tempOnly, err := snapshotDoltProcessesForConfigRoots(enumerate, []string{tempRoot})
+	if err != nil {
+		t.Fatalf("snapshotDoltProcessesForConfigRoots(tempRoot only): %v", err)
+	}
+	if _, ok := tempOnly[1002]; ok {
+		t.Errorf("tempRoot-only snapshot should not see the source-tree leak: %#v", tempOnly)
+	}
+}
+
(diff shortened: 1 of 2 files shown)


sjarmak (dismissed):
Approving. Verified rather than read: merged onto current main conflict-free, `go build` and `go vet` clean, and `GC_FAST_UNIT=1 go test ./cmd/gc/` green at 187.5s with zero failures. With the three prior offenders now merged, the widened guard finds nothing on a clean checkout, so it does not false-fire.

Revert-proofness checked the strict way. Rather than treating a compile error as proof, I kept `snapshotDoltProcessesForConfigRoots` present with its new signature and restored the old single-root body. `TestSnapshotDoltProcessesForConfigRootsCatchesSourceTreeLeak` then fails with `source-tree leak PID 1002 missing from snapshot`. It is a real guard, not a tautology. `TestSnapshotDoltProcessesForConfigRootsIgnoresEmptyRoots` passing under both behaviors is correct, since it asserts a property unrelated to the widening.

No sibling-path false positives are possible: `PathWithin` uses `filepath.Rel` with an explicit `".."` check, so `/x/cmd/gc2` never matches `/x/cmd/gc`. Confirmed empirically.

One thing to fix before this can merge, and it is currently red because of it:

`misspell` runs with `locale: US`, so the two British spellings trip `golangci-lint` even though `go vet` and `gofmt` are clean. That is what `CI / preflight`, `Check`, and `Preflight / static checks` are failing on.

- `cmd/gc/dolt_leak_helper_test.go:679` `behaviour` -> `behavior`
- `cmd/gc/path_helpers_test.go:126` `behaviour` -> `behavior`

Worth knowing about this PR's CI history: it had never actually run. The `Runner policy` gate failed on 2026-08-06 and every downstream job showed `skipping`, so the checks were neither green nor red. I re-triggered it, which is where the real result above came from. The remaining `Runner policy` / `add-triage-label` / `request-docs-review` failures are repo automation on a fork PR, not your change.

Scope note, not a defect: `sweepStaleCmdGCTestDoltProcesses` stays scoped to `tempRoot`, so a dolt server orphaned in the checkout across separate test-binary invocations is still invisible to the startup sweep. Your PR body is explicit that this makes the leak visible rather than stopping it, so this is an accurate boundary. Flagging it only so it does not get read as fully closed.

On the overlap with #5118, which fixes the same bug: this one is the more complete of the two. It widens `reapDoltProcessesUnderRoot`, the SIGINT/SIGTERM reap path, via `g.leakRoots()`; #5118 leaves that call on the single-root form. The two are mutually exclusive edits to `cmd/gc/path_helpers_test.go`, so this is the one to land. I have said as much on #5118.

Push the two-word fix and this merges. Say the word if you would rather I push it for you.

sjarmak:
Merging. I pushed the two-word `behaviour` -> `behavior` fix rather than sending it back for a round trip, since `misspell` runs `locale: US` and those two comments were the only thing standing between this and green. Checks are now 59 pass, zero fail.

Thanks for this one. It sat unreviewed since 2026-08-06 because its CI had never actually run, not because anyone judged it: the `Runner policy` gate failed and every downstream job showed `skipping`, so it looked neither green nor red and drew no attention. Re-triggering it is what produced the first real result.

PR #4577, opened 2026-07-23
fix(api): preserve canonical env for managed sessions

julianknutsen:
## Summary

- include workspace environment variables in API-created and API-resumed managed sessions
- preserve CLI precedence: process baseline, workspace, resolved provider/agent, city anchors, then canonical GC_BIN
- apply the shared environment builder to REST/Huma create, named-session materialization, direct resume, and reconciler resume paths

## Root cause

Slack ingress successfully reached the API, but the API runtime builders reconstructed only provider passthrough plus city anchors. They omitted workspace environment and the running gc binary path. Engineering bridges fail closed when GC_BIN is absent, so an otherwise accepted message surfaced as a Slack delivery error when it woke an API-managed session.

## Validation

- regression test demonstrated RED on main: WORKSPACE_ONLY absent and provider GC_BIN incorrectly authoritative
- focused create/resume regression tests pass
- go test ./internal/api -count=1
- repository .githooks/pre-commit: format, lint-changed, generated artifacts, go vet ./...
- make dashboard-check
- make test-fast-parallel: all 8 fast jobs passed

Tracking bead: ga-h5ykt

PR #5061, opened 2026-08-06
session: pin the controller token empty, and keep it pinned across respawn

julianknutsen:
`ScrubTokenEnv` withheld `GC_CONTROLLER_TOKEN` from agent sessions by **deleting** the key from the
env map. Deletion does not withhold anything, and this repo already knew why.

`internal/runtime/tmux/tmux.go` builds a pane's environment by passing non-empty values as `-e` and
collecting **empty** ones into an `env -u` prefix, under a comment that states the problem exactly:

> tmux `-e` sets session-level env but the shell process still inherits from the tmux server's global
> environment. `env -u` ensures the var is actually absent from the child process.

A key that is *present-but-empty* therefore becomes genuinely absent from the child. A key that is
**deleted** emits nothing at all — so a tmux server whose global environment carries the controller
token hands it to every agent pane. An agent gets a controller capability.

The existing test asserted absence **from the map**, so it passed the entire time the invariant was
broken. That shape is the actual root cause here, and it is why the new tests assert at the child.

### Why this touches four places

`passthroughEnv()` is only one of the paths that build a session environment. `worker_handle.go` (create
and resume) and `internal/api/session_runtime.go` call `ProviderProcessPassthroughEnv()` directly and
never went through the scrub at all — those two were leaking unconditionally. The pin therefore lives in
`processenv`, which every builder starts from.

Two subtleties a reviewer should check:

- **Ordering.** `passthroughEnv()` seeds from `ProviderProcessPassthroughEnv()` and *then* sweeps
  `os.Environ()` for `GC_`-prefixed variables. The token is itself `GC_`-prefixed, so without an
  exact-name skip the sweep reads the real value straight back over the pin.
- **The expansion hole.** Last-wins pinning defeats a literal override in config, but not
  `FOO = "$GC_CONTROLLER_TOKEN"`, which lands the value under a key no key-level guard watches. That
  needs a read-side guard, so `ExpandSessionEnvValue` masks these keys during `os.Expand`.

### Respawn

An `env -u` prefix is a property of one exec, not of the session — empty values are excluded from `-e`,
so nothing persists in the tmux session environment. `respawn-pane` rebuilds the pane from the server's
global environment, so a warm-box relaunch handed the token straight back. Measured on tmux 3.4:

| mechanism | created pane | respawned pane |
|---|---|---|
| `env -u` prefix only | absent | **real token** |
| `-e KEY=` | empty | empty |
| `env -u` prefix **+** `set-environment -r` | absent | absent |
| `set-environment -r` alone | **real token** | absent |

Hence both: the durable marker survives respawn, `new-window` and `split-window`; the prefix covers
creation, because `new-session` starts the command before any follow-up marker can land. `-r` rather
than `-u`, since `-u` deletes the session entry and re-exposes the server's global value.

The marker is scoped to the controller-only keys. The nesting flags (`CLAUDECODE`, `CODEX_*`) keep the
existing prefix-only behaviour — every key marked is a key whose marking can fail a launch, so only keys
that need to survive a respawn should pay that.

It tolerates exactly one condition: a session that has already exited. Such a session has no pane to
leak into and no warm box to respawn, so the control is vacuous rather than unapplied; on a live session
a marker failure still fails closed and names the key. That condition is confirmed by asking tmux via
`HasSession`, not by matching stderr — tmux 3.4 answers `set-environment` with "no such session" but
`has-session` with "can't find session", so stderr matching would have been wrong on this very command.

### Testing

17 mutants, 0 survivors. The child-level assertions are the point: a real `/bin/sh` for the exec
runtimes, and an integration test that starts a tmux server **holding the token in its global
environment**, relaunches through `Provider.Relaunch`, and reads the value back out of the respawned
pane's own shell.

Also fixed along the way: a 108-byte test socket path, one byte over the `sun_path` limit, which tmux
reports as the thoroughly misleading "no server running".


julianknutsen:
## Automated review

Produced by an automated council: five independent reviewers across distinct lenses (missed paths, empty-vs-unset regressions, residual exfiltration, test integrity, and language hygiene for a public repo), each finding then independently re-checked by a separate reviewer instructed to refute it. 13 findings were raised; 6 were refuted and dropped. This is machine review, not a human sign-off.

**The review blocked an earlier revision of this branch.** All three surviving findings are fixed in the commit as it now stands. They are recorded here because the reasoning is the useful part, and because one of them was resolved *differently* from what the reviewers proposed.

### 1. Warm-box relaunch re-leaked the token (high, was blocking)

Four lenses converged independently, with live reproduction on tmux 3.4. The empty pin was realized only as an `env -u` prefix built inside `NewSessionWithCommandAndEnv`; empty values are excluded from `-e`, so nothing persisted in the tmux session environment. The relaunch path — `Provider.Relaunch` → `doRelaunchSession` → `respawnAgent` → `RespawnPaneWithWorkDir` — issues `respawn-pane -k` with the bare command and drops `cfg.Env`, so the pane was rebuilt from the server's global environment, which still held the real token. A routine launch-only reconcile handed the capability back. Created pane reported `[ABSENT]`; respawned pane reported the real token.

The reviewers' preferred fix was to apply the same `env -u` prefix at the relaunch site. **That is not what was implemented**, after the four candidate mechanisms were measured against a real server:

| mechanism | created pane | respawned pane |
|---|---|---|
| `env -u` prefix only (pre-fix) | `[ABSENT]` | **real token** |
| `-e KEY=` | `[]` | `[]` |
| `env -u` prefix **+** `set-environment -r` | `[ABSENT]` | `[ABSENT]` |
| `set-environment -r` alone | **real token** | `[ABSENT]` |

The durable session marker was chosen instead, *combined with* the existing prefix. Two reasons the measurements make concrete. First, the defect was never "the relaunch site forgot a prefix" — it was that the withholding lived in a one-shot command string rather than in the session; marking the session also covers `new-window`, `split-window`, the bare respawn entry point, and future paths, where a relaunch-site prefix would leave them open. Second, the last row shows the prefix cannot be dropped: `new-session` starts the command before any follow-up marker can land, so the two mechanisms close two different windows and each is falsified by its own mutant.

`-r` rather than `-u`: `-u` deletes the session entry and re-exposes the server's global value.

`-e KEY=` was rejected because it yields present-but-empty rather than absent, weakening the contract the rest of the change asserts. The reviewers were right, though, that the *stated rationale* in the neighbouring test comment was wrong — `-e` does override the server environment — so that comment was rewritten to describe the real mechanism rather than leave a disproved claim in the tree.

### 2. Manifest drift: CI red, key test skipped (medium, was blocking)

The new tests were missing from the runtime-tmux manifest, so the manifest conformance test failed on the branch. Worse, the Linux integration shards select tests solely from that manifest, so the one test that observes a real server leaking into a real pane would never have run in CI — a green build would not have meant anything.

The reviewers' suggested counts were wrong (they assumed 2 new tests; there are 7). The manifest was regenerated with the same `build.Context` and AST logic the conformance test itself uses, validated by reproducing the existing entries byte-identically before adding to them.

### 3. Commit message cited unresolvable history (medium)

The message referenced a commit hash not reachable from this repository. Dropped, and the reasoning re-anchored to #4577, which the code already cites.

### Coverage

15 mutants, 0 survivors — every guard has a named test that fails when the guard is broken. The full integration tier passes (353 tests).

PR #5062, opened 2026-08-06
session: pin the controller token empty, and keep it pinned across respawn

julianknutsen:
Supersedes #5061, which never produced a check suite on open, reopen or synchronize. Full description and the automated review are on that PR; reposting both here once checks appear.


julianknutsen:
Superseded — reverting to #5061; the missing check suite is not branch- or PR-specific.

PR #5063, opened 2026-08-06
ci: scan the Go module graph for known vulnerabilities

julianknutsen:
## What this adds

A standalone `Go Vulnerability Scan` workflow that runs [`govulncheck`](https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck) over `./...` on pull requests, pushes to `main`, a daily schedule, and on demand.

Nothing in CI reads the Go module graph today:

| Existing check | Reads |
| --- | --- |
| Container Scan | built container images (Trivy `--scanners vuln`), plus Dockerfile/manifest config |
| CodeQL | first-party source, as SAST |
| OpenSSF Scorecard | repository posture, advisory-only (`continue-on-error: true`) |

None of these maps `go.mod` / `go.sum` to the Go vulnerability database, so a dependency can carry a published advisory indefinitely without any check noticing.

## Failure policy: block on reachable, report on everything else

govulncheck sorts each finding into three tiers:

1. **a vulnerable symbol the build actually calls** — the job fails
2. **a vulnerable package imported, but no vulnerable symbol called** — reported, does not fail
3. **a vulnerable module required, but not imported** — reported, does not fail

I want to be clear that this is not a policy layered on top of the tool — it *is* the tool's exit-code contract, and the workflow simply lets the exit code stand. I verified the behaviour against a fixture that calls a known-vulnerable symbol (`golang.org/x/text@v0.3.7`, `language.ParseAcceptLanguage`, GO-2022-1059): tier 1 exits 3 and prints the call trace, while 3 tier-2 and 38 tier-3 findings in the same run leave the exit code at 0.

The argument for the split:

- **Gating on tier 1 is worth it.** A called vulnerable symbol is a real, actionable defect with a call trace attached. The tree is at zero today, so the gate starts green and any red is a genuine regression rather than inherited debt.
- **Gating on tiers 2 and 3 would be actively harmful.** They are dominated by transitive pins the PR author did not choose. One of the two findings on this tree right now has `Fixed in: N/A` — there is no version to bump to. A check that is red on day one for something nobody can fix is a check people learn to click past, which costs more than it buys.

Rather than discard the lower tiers, they go to the job log, the step summary, and Code Scanning. govulncheck's SARIF output preserves the distinction natively — called is `error`, imported-not-called is `warning`, required-not-imported is `note` — so Code Scanning ranks them instead of flattening everything into one undifferentiated pile.

There is deliberately **no `paths:` filter**, unlike Container Scan. Tiers move when first-party code changes, not only when `go.mod` does: adding a single call to an already-present vulnerable package promotes a tier-2 finding to tier 1 with no dependency change at all. Filtering on `go.mod`/`go.sum` would blind the scan to exactly the transition reachability analysis is best at catching. The daily schedule covers the third case, where a newly published advisory lands against code that already shipped.

## What it reports on `main` today

```
=== Symbol Results ===
No vulnerabilities found.

=== Package Results ===
No other vulnerabilities found.

=== Module Results ===

Vulnerability #1: GO-2026-5932
    The golang.org/x/crypto/openpgp package is unmaintained, unsafe by design,
    and has known security issues
  Module: golang.org/x/crypto
    Found in: golang.org/x/crypto@v0.53.0
    Fixed in: N/A

Vulnerability #2: GO-2026-5841
    OOB read in github.com/klauspost/compress/s2
  Module: github.com/klauspost/compress
    Found in: github.com/klauspost/compress@v1.18.5
    Fixed in: github.com/klauspost/compress@v1.18.7

Your code is affected by 0 vulnerabilities.
```

**Zero reachable vulnerabilities.** Two module-only findings, neither imported. `klauspost/compress` has a fix available (`v1.18.7`) and is worth a routine bump; `x/crypto/openpgp` has none and is an advisory about an unmaintained package we do not import. Exit code 0, so this PR should be green.

## On adding a second tool

I looked at whether something should sit alongside govulncheck, and mostly the answer is no — but for an evidence-based reason worth recording.

**Dependabot alerts are already enabled on this repository and already cover Go.** GitHub's dependency graph resolves 244 `pkg:golang` packages here. It currently reports **65 open alerts: 61 npm, 4 pip, and zero Go** — while govulncheck reports two Go findings on the same tree. So "turn on Dependabot for Go modules" is not an available lever: it is on, it can see the Go graph, and it is silent on both findings above. Dependabot *updates* would be redundant too, since `renovate.json` already manages `gomod` bumps.

That 65-alert backlog is also the best available argument for the failure policy above. It is exactly what an ungated, reachability-blind feed of transitive findings turns into.

A lockfile scanner (`osv-scanner`, or Trivy in `fs` mode) would genuinely complement govulncheck rather than duplicate it, because of a real limitation described below. I have **not** added one here — it would need its own noise budget and its own triage owner, and that is a separate decision from closing the systematic gap. Worth taking on its own merits.

## What this does not cover

Stating the limits plainly, because a security check that is oversold is worse than none:

- **Reachability is an approximation.** govulncheck's call-graph analysis does not follow reflection, `go:linkname`, `unsafe`, or dynamic dispatch it cannot resolve. A symbol it reports as uncalled can still be reachable in practice.
- **It only sees what the scan pattern builds.** Modules that are in `go list -m all` but contribute no compiled package to `./...` — build-time tooling pulled in via a `tool` directive, for instance — are invisible to govulncheck at *every* tier, including tier 3. This is the gap a lockfile scanner would close, and it is not a small one: it is silent, and it does not look different from a clean result.
- **It is only as good as the Go vulnerability database.** `vuln.go.dev` is OSV-backed and curated. An advisory published to another feed, or to a GitHub advisory that was never mirrored into OSV, will not appear here at all. Coverage is good for the Go standard library and popular modules and thinner in the long tail. Concretely: a clean govulncheck run means "nothing in the Go vuln DB is reachable", not "no known vulnerabilities".
- **It says nothing about non-Go dependencies.** The npm and pip surfaces are out of scope; Container Scan and Dependabot remain the coverage there.
- **The daily schedule is the only thing that catches newly published advisories.** Between runs, exposure is real.

## Implementation notes

- `go install golang.org/x/vuln/cmd/govulncheck@v1.4.0` ignores the working-directory `go.mod`, so the scanner does not enter the module graph. This matters: `check-native-dependency-surface.sh` caps the graph at exactly the number of modules it has today, and a `tools.go` entry would push it over. Verified `go.mod` and `go.sum` are byte-identical after a full run.
- The scanner version is pinned so a scanner release cannot change the verdict without a reviewed commit. The vulnerability database is still fetched fresh every run, so pinning the binary does not stale the data.
- One trap worth flagging for future maintenance: govulncheck must be *built* with a toolchain at least as new as the `go` directive in `go.mod`. Build it with an older one and it emits `package requires newer Go version` for every package, analyses nothing, and **still exits 0**. Using `go-version-file: go.mod` for `setup-go` keeps the two in lockstep, which is why the version is not hardcoded.
- Follows the existing `runner-policy` pattern used by CodeQL and Container Scan, so PR runs get the same runner selection as the rest of CI.

## Checks

`actionlint` clean. `go test ./scripts/cipolicy`, `go test ./test/workflows/...` and the workflow version-pin tests all pass — the new file is standalone, so no pinned CI execution hash changed. No existing check was modified, relaxed or skipped.


quad341 (approved):
@julianknutsen — the failure-policy write-up is the best part of this. "Gate on tier 1, report tiers 2 and 3" is the right call, and grounding it in the 65-alert Dependabot backlog on this very repo is a far better argument than the usual "we should scan dependencies." Thank you also for the "what this does not cover" section: reflection and `go:linkname` escaping the call graph, `tool`-directive modules being invisible at *every* tier, and — the one a future maintainer will be most grateful for — that silent invisibility looks identical to a clean result.

I checked the claims against the tree rather than taking them on faith: the runner-policy block matches `codeql.yml`, the allowlist matches `blacksmith-allowlist.txt`, the `setup-go` SHA matches the `cipolicy` constant, and the module cap in `check-native-dependency-surface.sh` is real — so dodging a `tools.go` entry was necessary, not just tidy.

One small thing I'll take on my side: the summary treats every nonzero exit as a reachable vulnerability, but only exit 3 means that — exit 1 is a scanner failure. I'll branch on 3 explicitly so an infra failure doesn't read as a security finding. Your gating stays exactly as written. Then I'll merge once the scan's own run goes green here. Nothing needed from you.


---

PR #5058, opened 2026-08-06
fix(worker): retry transcript metadata sidecar

julianknutsen:
Summary
- retry an initially missing keyed transcript metadata sidecar after successful supervisor delivery
- bound retries to three one-second attempts and retain the existing idempotent write
- leave disabled one-shot CLI processes unchanged

Validation
- go test ./internal/worker -count=1
- go vet ./internal/worker
- commit-hook lint: 0 issues

PR #5068, opened 2026-08-07
test: record GC_TRANSCRIPT_META_ENABLED in the env-read golden

julianknutsen:
`TestGCEnvReadBaseline` fails on `main`: the transcript-metadata sidecar retry
(#5058) reads `GC_TRANSCRIPT_META_ENABLED`, and the env-read baseline requires
new `GC_*` reads to be either promoted into `internal/rollout/config` or
recorded in the golden deliberately. This records it. One line; no behavior
change.


julianknutsen:
## Review outcome

PR #5069, opened 2026-08-07
beads: add a pure-Go SQLite bead store

julianknutsen:
# beads: a pure-Go SQLite bead store

Adds `internal/beads`' own SQLite implementation of the `beads.Store`
interface, built on `modernc.org/sqlite` so it compiles with
`CGO_ENABLED=0`. Nothing calls it yet — this is the engine layer, landing
on its own so the contracts and config surface that consume it can be
reviewed separately.

## What it provides

- The full `Store` surface (create, get, update, close, list, ready,
  children, labels, metadata, dependencies) over a WAL-mode database with
  one serialized write connection and a pooled reader.
- The optional capabilities the rest of the tree resolves by type
  assertion: `ConditionalWriter` for fenced compare-and-swap writes,
  `Claim`'s compare-and-swap acquire, `BatchDeleter`, `GraphApplyStore`
  and its tier-aware extension, `StorageCreateStore`,
  `ContextReadyReader`, and `ForeignIDCreator` for id-preserving copies
  between stores.
- Exact-schema admission: an existing database is inspected read-only
  before a writable connection exists, and an earlier layout whose `deps`
  primary key includes `dep_type` stays readable and writable in place, so
  an older binary can still reopen a database this one has written.
- A strictly read-only open (`mode=ro`) that leaves a source database and
  its uncheckpointed WAL byte-identical across open, read and close.
- A sequence floor persisted beside the database under a cross-process
  lock, so an id observed elsewhere is never reissued after a restart.
- A closed store answers, it does not crash: after `CloseStore` every
  exported method returns `beads.ErrStoreClosed` rather than dereferencing
  a released handle.

`validateGraphApplyPlan` moves into `graph_apply.go` as the shared plan
contract for in-process graph appliers. The Dolt store keeps its own copy,
so this changes no existing behavior.

## Behavior deltas

None. No caller constructs this store; every existing path is untouched.

## Tests

The shared `beadstest` conformance suites (store, sequential ids, creation
order, deps, metadata), the fenced-write and ownership-fence suites, plus
graph-apply, ready-query, read-only-source, schema, batch-delete,
closed-store and sequence-floor coverage — including a Linux
SIGKILL-at-boundaries process test. That process test is a new subprocess
user, so the all-source audit baseline in `test/test-resources.toml`, the
pinned policy in `internal/testpolicy/resourcecensus`, and the generated
`TESTING.md` row move with it.


julianknutsen:
## Review outcome

PR #5070, opened 2026-08-07
storage: add the storage-binding contracts, the SQLite provider, and the class conformance suite

julianknutsen:
# storage: the storage-binding contracts, the SQLite provider, and the class conformance suite

Adds `internal/storebinding` — the provider-neutral storage-class contracts
and the machinery that resolves a configured class-to-binding assignment
into a frozen plan — plus `internal/storebinding/sqlite`, the one built-in
provider, and `internal/storebinding/storebindingtest`, the reusable
conformance corpus every class front door is proved against.

Nothing routes on any of it yet: no production code resolves a plan or
opens a binding. It is the library half of config-driven storage-class
bindings, landing on its own so the contracts and the provider can be
reviewed apart from the boot wiring that consumes them.

Depends on the pure-Go SQLite bead store PR.

## What lands

- The six closed class contracts (work, graph, sessions, messaging,
  orders, nudges), the canonical Beads adapters that satisfy them over one
  store, and the class wrapper stack (cache, typed events, maintenance
  discipline) every provider's front doors run through unchanged.
- The compiled provider registry — frozen before use, resolved by exact ID
  with no alias step — and the plan resolver that turns a class assignment
  into a frozen open, inspection and close program.
- The migration saga's durable state: phase-typed attempt records, the
  append-only journal, the proven-copy manifest, the semantic witness (a
  provider-neutral digest over a canonical logical stream), and the
  cross-process migration guard.
- `internal/storebinding/sqlite`: the Beads-over-SQLite provider (ID
  `sqlite-beads`) that walks the whole Inspect / AcquireFence /
  InspectFenced / Open lifecycle against a real database, plus the
  graph-engine inspection, witness and fence surface it composes.
- The `[storage]` config surface: class assignments and named provider
  bindings, parsed, layered across include fragments and validated.
  Nothing reads it at runtime; the plan resolver's input type lives here,
  which is why it lands with the resolver.
- `cmd/gc`'s single provider composition root and the boundary test that
  pins it: one construction site, one compiled-factory list, built-in
  provider IDs only, a CGO-invariant storage surface, and a replace-free
  module graph. A downstream integrator adds an out-of-tree provider by
  appending one factory in its own tree.

## Contract additions in existing packages

`session` gains a typed address directory, transactions, and front-door
resolution; `extmsg` and `beadmail` take that directory instead of a raw
session store, with the raw-store constructors kept as thin adapters so
existing callers are unchanged; `orders` gains swept close, run deletion
and tracking-bead helpers; `nudgequeue` gains terminal-aware shadow
history; `convoy`, `coordclass` and `beadmeta` gain the class-ownership
helpers the contracts assert against.

### Mail addressing is unchanged, and that is tested

Moving mail's session reads behind a typed directory must not change WHERE
a message lands. The directory therefore answers two different questions
with two different algorithms, matching what the raw-store callers did:

- `ResolveAddress` — session targeting — reaches `ResolveSessionID` /
  `ResolveSessionIDAllowClosed`, the same package functions every
  raw-store caller uses. Exact bead id first (a closed session included),
  then live canonical `session_name` with the dual alias/`session_name`
  demotion, then live current alias.
- `ResolveMailboxAddress` — mailbox ownership — treats a session's bead
  id, alias and `session_name` as candidates AT ONCE in one liveness pass,
  so an address two sessions claim is ambiguous rather than won by
  whichever key was probed first. Mail turns that ambiguity into
  literal-only delivery.

`internal/session/address_directory_test.go` carries an equivalence oracle
that resolves every fixture through both the package function and the
front door and requires the same answer.
`internal/mail/beadmail/beadmail_test.go` carries six **address contention
fixtures** — the block comment `--- Address contention fixtures ---` names
them — in which two sessions contend for one address: three assert the
recipient routes, three assert the stamped sender session id.

The metadata probe re-checks the rows a backing store returns against the
address that was asked for. A store that answers a metadata filter loosely
would otherwise deliver someone else's mail;
`TestStoreResolveMailboxAddressRejectsALooseMetadataAnswer` pins it.

## Deliberate behavior deltas

Mail addressing has none — see above. Three deltas exist in `extmsg`,
where the typed directory replaced best-effort raw reads:

1. **Indeterminate session lookups now stop the operation.** A backend
   failure or an ambiguous name during binding/participant resolution and
   during the stale-binding and stale-participant sweeps returns an error
   instead of being skipped. "Session not found" keeps its previous
   meaning and its previous fall-back. The sweeps are idempotent and
   reconciler-driven, so a failed tick is retried; the alternative is
   routing an external message at an identity the directory could not
   confirm.
2. **The liveness fast path resolves rather than fetches.** Checking
   whether a stored session id still owns a stable name goes through
   session resolution instead of a raw bead read. The only input where the
   two differ is a stored id naming a live NON-session bead: previously
   the stale id was kept, now the stable name is re-resolved.
3. **Error text is constant.** Session-address failures are wrapped in an
   error whose message names the operation only, so caller-supplied
   selectors and persistence ids do not reach logs. Causes are preserved
   for `errors.Is`.

## A note on test fixtures

Negative-path fixtures that must contain credential-shaped input (the
binding-spec validator's rejection table) carry an inline
`oss-leak-allow:fixture` marker naming why the value is there. They are
inputs the validator refuses, not material anything reads.

## Tests

The class conformance corpus over the Beads adapters (in-memory and over a
real SQLite engine) and through the production wrappers; the provider
lifecycle and hardening suites; the plan, manifest and saga suites; and the
boundary guard's five arms — including a frozen-registry arm that offers a
well-formed factory and requires `ErrProviderRegistryFrozen`, since a nil
factory is refused by an unfrozen registry too and proves nothing.

Four new process tests are new subprocess users, so the all-source audit
baseline in `test/test-resources.toml`, the pinned policy in
`internal/testpolicy/resourcecensus` and the generated `TESTING.md` row
move with them.


julianknutsen:
## Review outcome

PR #5071, opened 2026-08-07
storage: route coordination classes at boot, and refuse rather than migrate

julianknutsen:
# storage: route coordination classes at boot, and refuse rather than migrate

The behavior half of config-driven storage-class bindings. The contracts, the
provider and the `[storage]` surface landed with nothing reading them; this
makes a city read them.

Depends on the storage-binding contracts PR.

A city whose `[storage.classes]` point graph, sessions, messaging, orders and
nudges at a binding serves those classes from that binding's bead engine. Work
stays on the work ledger. On a production-scale city that measured 130-250x
lower latency for infrastructure-class reads, because those reads stop
competing with the work ledger's own traffic.

## What changes for a city with no `[storage]` section

Nothing. The gate short-circuits at the top: no registry is constructed, no
plan is resolved, no binding root is read, and every class resolver returns the
exact store value it returns today. That is asserted as a negative — a counting
registry constructor proves none was built — because asserting the routes came
back nil would still pass if the gate had built and discarded one.

Two census tests hold the shape: the registry constructor has exactly one
caller (the `var newStorageRegistryForPlan = newStorageProviderRegistry`
initializer), and the no-config CLI path constructs zero registries.

## Boot refuses; it never migrates

Three outcomes, none of which moves data: a converged binding is opened and
served; a city with nothing to move gets its binding created and an empty
proven-copy manifest recorded, so stranded-write detection is armed from the
first boot; a city whose config and data disagree **refuses to start** and
names the one command that performs the move.

`newCityRuntime` becomes fallible to carry that refusal. The standalone
controller prints it and exits 1; the supervisor returns it from the
post-prepare step it already runs the constructor inside, records the failure
for that city, and keeps managing the others.

## One-shot commands take the same verdict

`gc mail`, `gc nudge`, `gc sling`, `gc formula`, `gc prime` and the control
dispatcher's one-shot arm open the work store themselves and build no city
runtime. Each of their class resolvers now consults a process-scoped funnel
that takes the SAME gate — same verdict precedence, same message text — once
per city per process, closed on the way out of `run`. A CLI reading the work
store while the controller reads the binding is two ledgers for one city, and
every symptom of it (mail sent and never delivered, a nudge queued into a store
nothing polls) looks like a bug in the feature rather than in the routing.

`gc storage` itself stays off the funnel structurally: `migrate` and `status`
resolve their own target and reach no class resolver, so the remedy runs on
exactly the city the refusal is about. A test fails if either ever starts
routing through a class.

## Deliberate behavior deltas

1. **Relocated classes are served from the binding** — the feature.
2. **A misconfigured or unconverged city refuses to start**, where it
   previously started and served infrastructure reads from wherever the work
   store happened to be.
3. **One-shot commands refuse too.** A class resolver has no error to return
   and terminating the process is not the CLI's to do (the invocation metrics
   lifecycle must keep control until `run` returns), so the refusal is
   delivered AS the store: the five infrastructure classes resolve to a store
   whose every operation returns the refusal, and the reason is printed once to
   stderr. Work stays on the work ledger. Residue worth stating: a call site
   that discards store errors now does nothing instead of erroring — still
   strictly better than reading the ledger the class was moved off.
4. **A config reload that changes `[storage]` is refused** — the whole reload,
   before a provider is rebuilt — because live storage handles are immutable
   for the life of a process. The booted configuration stays in force until the
   city is restarted.
5. **Two supervisor failure branches now shut the city runtime down.**
   `opening_controller_state` and `running_pool_on_boot` previously abandoned a
   built runtime, leaking everything it held (now including a storage engine).

## Known limitations, stated rather than discovered

- **A `city.toml` that will not load is treated as a city with no
  `[storage]`** by the one-shot funnel, which means such a city's CLI commands
  read the work store. By then the command's own config load has already failed
  and reported; refusing every command on a city that is merely misconfigured
  would make it unusable.
- **A controller-resident socket handler reaches a one-shot class resolver**
  (the session circuit-breaker reset), so the daemon process holds a second
  handle to the binding it already serves from. It already opens a second work
  store there today; the engine is WAL-mode and the id floor is
  cross-process-locked, so this costs a file handle, not correctness.
- **The unmounted-volume message still names the migrate command.** A missing
  binding root on a converged city is an unmounted volume, not an unmigrated
  city, and the message now says so with the hazard (copying onto a bare
  mountpoint leaves two divergent stores — mount it first) ahead of the
  command. The command stays at the end because the ordinary first cutover has
  exactly this state.

## Rollback

Point `[storage.classes]` back at `work` and drop the now-unreferenced binding
definition in the same edit. Deleting the `[storage]` section is **not** a
rollback: it takes the unconditional bypass, checks no marker, and silently
serves the retained work store while the city's infrastructure state sits in
the binding. The runbook carries that as a hazard box beside the correct
spelling, and a three-arm test pins the prose.

## Shutdown ordering

The engine closes last — after the order drain, the city-stop sleep reason and
the two-pass session stop have all written through it — and exactly once, on
both the normal and the preserved-session exit. The test reads the binding at
the instant of close.

## Extension point

`storebinding.EngineOpener`: a provider whose binding IS a bead engine hands
back the canonical store for the classes it serves. Boot consumes it through
the resolved plan, so a downstream integrator serves an out-of-tree engine by
registering its factory and implementing that one method, with no edit here. A
planned binding whose provider does not implement it is a refusal that names
the provider — proved by a test that resolves a plan actually carrying such a
binding, not by a plan that never contained one.

## Supported topology

This build serves the whole infrastructure split or none of it. The plan
machinery resolves more shapes than the runtime can serve, and routing half a
split by silence is the failure the gate exists to prevent, so a partial
fan-out or a relocated work class refuses at startup. A pinned test holds the
served shape and the convergeable shape to the same definition.

## Tests

The no-config bypass and the identity resolver; the shape agreement; all three
live branches end to end, including a session bead written through the class
resolver landing in the binding and not in the work store; a one-shot
`gc mail` write landing in the binding, absent from the work store, and read
back by a second one-shot provider; the CLI refusal naming the migrate command;
shutdown ordering; the reload refusal; the refusal proved not to have migrated
by listing the binding root's parent rather than by reading a stat error (every
absence assertion is a positive listing, so a typo'd path fails instead of
passing); the uncheckable verdict proved to withhold both the "unconverged"
claim and the revert; the guard, controller-absence and rig-residue refusals; a
read-only `status` pinned by a tree fingerprint; and the migration battery the
copy already carried.

Two command census entries are new, so the committed manifest, its generated
runtime table and decode catalog, the public example schema enum, and the
pinned catalog size move together. Four event types are new, so the OpenAPI
spec, its docs mirrors, the generated Go client and the dashboard client move
together.


julianknutsen:
## Review outcome

Issue #618, opened 2026-04-11
bug: gc sling --formula silently creates beads with unresolved template variables


michael-barker:
## Gas City version

0.13.5 (commit: 29735489, built from source 2026-04-11)

## Environment

macOS, zsh, tmux provider.

## Summary

`gc sling <pool> <formula> --formula` does not validate template variable substitution. When `--var` flags are missing (even for vars marked `required = true`), beads are created with literal `{{var}}` in their titles and descriptions. No error or warning is emitted.

This causes downstream agent churn: agents claim the bead, can't make sense of the unresolved placeholders in the title/description, and spend tokens trying to "investigate" or "fix" the template system rather than doing useful work. The bead remains open and gets reclaimed on the next session cycle, repeating indefinitely.

> **Note:** Our city uses a custom `claim_guard.py` pre-claim hook that dead-letters beads after N failed claims. Without such a guard, the churn loop is unbounded.

## Reproduction

**1. Create a formula with required vars:**

```toml
# formulas/repro-unresolved.formula.toml
description = "Repro: unresolved template variables survive into bead titles."
formula = "repro-unresolved"
version = 1

[vars]
[vars.epic]
description = "Epic ticket ID"
required = true

[vars.feature]
description = "Feature slug"
required = true

[[steps]]
id = "implement"
title = "[{{epic}}] Implement: {{feature}}"
tags = ["implement", "{{epic}}"]
description = "Implement the {{feature}} feature for {{epic}}."

[[steps]]
id = "review"
title = "[{{epic}}] Review: {{feature}}"
needs = ["implement"]
tags = ["review", "{{epic}}"]
description = "Review the {{feature}} implementation."
```

**2. Pour with a missing required var (exits 0, no warning):**

```bash
gc sling claude repro-unresolved --formula --var epic=CLOUD-99999
# Slung formula "repro-unresolved" (wisp root rtv-1ao) → claude
```

**3. Observe unresolved placeholders in bead titles and descriptions:**

```
$ bd children rtv-1ao
○ rtv-1ao   ● P2 repro-unresolved
├── ○ rtv-1ao.1 ● P2 [CLOUD-99999] Implement: {{feature}}
└── ○ rtv-1ao.2 ● P2 [CLOUD-99999] Review: {{feature}}

$ bd show rtv-1ao.1
○ rtv-1ao.1 · [CLOUD-99999] Implement: {{feature}}   [● P2 · OPEN]
DESCRIPTION
[...]

PR #5074, opened 2026-08-07
fix(molecule): extend residual-var guard to routing metadata (#5060)

### LINKED ISSUE #5060: bug: formula/molecule: residual-var guard is title-only — unresolved {{var}} in routing metadata (gc.run_target/gc.routed_to) survives to a durable, unroutable bead
## Summary

The `#618` residual-variable guard that rejects beads with unresolved `{{var}}`
placeholders only inspects the bead **title**. A missing or misspelled routing
var leaves a literal, unrendered template string in the **routing metadata**
(`gc.run_target` / `gc.routed_to`) — e.g. `gc.routed_to = {{review_target}}` —
and that bead is written to the store with no error. The bead then routes to a
target that resolves to no agent, so it lands with an empty/unclaimable assignee
after an otherwise-"verified" `gc sling`.

Verified against `gastownhall/gascity` `origin/main` @ `08b6396` (fetched
2026-08-06). This is a distinct bug from #618 (which is fixed) and only
partially overlapping with #1512 — see "Relationship to existing issues".

## Symptom

Molecule/workflow step beads are created whose routing metadata contains the
literal template string instead of the resolved role:

```
gc.routed_to = {{review_target}}      # expected: e.g. gc.review-synthesizer
gc.run_target = {{prep_target}}       # expected: e.g. mathcity.brief-operator
```

The step bead's **title and description render correctly** (they use vars that
*were* provided); only the routing metadata carries the unrendered placeholder.
Because the route resolves to no agent, the bead is left unassigned even though
the sling reported success.

Field evidence (downstream city): 212 beads carrying the fingerprint
`empty_assignee_after_verified_sling`:

- **185 / 212 (~87%)** routed to a literal `{{prep_target}}` / `{{review_target}}`
  — the unrendered-template class this issue is about.
- 9 / 212 routed to the retired pool `gastown.dog` — the *resolved-but-stale*
  class that #1512 already covers (out of scope here).

## Reproduction

The end-to-end field repro is the 212 beads above; the code path is
deterministic and is reproducible from a formula whose routing var is omitted at
sling time while a title var is supplied. Minimal shape (adapted from #618,
which established the same footgun for titles):

```toml
# formulas/repro-route-residual.formula.toml
formula = "repro-route-residual"
version = 1

[vars]
[vars.topic]          # supplied at sling → title renders
description = "Work topic"
required = true
[vars.review_target]  # NOT supplied, NO default → route stays literal
description = "Reviewer role"
required = true

[[steps]]
id = "review"
title = "Review: {{topic}}"                 # renders fine
[steps.metadata]
"gc.run_target" = "{{review_target}}"        # stays literal → gc.routed_to = {{review_target}}
```

```bash
gc sling <pool> repro-route-residual --formula --var topic=widgets
# exits 0, no warning
bd show <review-step-bead>
#   title:  "Review: widgets"           ✅ rendered
#   gc.routed_to: {{review_target}}      ❌ literal, unrouted → empty assignee
```

Result: expected `gc sling` to **reject** the missing required routing var (as it
already does for titles) → got a bead written with `gc.routed_to = {{review_target}}`
and no assignee.

> Note: `required = true` enforcement for vars is also not wired here — same gap
> #618 noted for titles — but the load-bearing bug is the **guard coverage**:
> even with the guard present, it never looks at routing metadata.

## Root cause

`internal/formula/parser.go:615` — `Substitute` intentionally **keeps unresolved
placeholders**: a missing var returns the literal `{{name}}` (parser.go:622).

`internal/molecule/molecule.go:1339` — `stepToBead` substitutes *all* metadata
values (`b.Metadata[k] = formula.Substitute(v, vars)`), so routing metadata is
run through `Substitute` — but a missing routing var therefore survives as a
literal `{{...}}` value rather than being caught. (The graph route path applies
the same `Substitute` to `gc.run_target` at
`internal/graphroute/graphroute.go:172`.)

The safety net that is supposed to catch this — the residual-var guard added for
#618 — only checks the **title**:

- `internal/molecule/graph_apply.go:252-255` (`buildRecipeApplyPlan`):
  `if residual := formula.CheckResidualVars(node.Title); ...` — **title only**.
- `internal/molecule/graph_apply.go:475-478` (`buildFragmentApplyPlan`): identical
  guard, **title only**.

`node.Metadata` (including `gc.run_target` / `gc.routed_to`) and `node.Description`
are never passed to `CheckResidualVars`. So a residual `{{...}}` in routing
metadata bypasses the guard and is written to the store.

**Invariant violated:** `engdocs/design/formula-compiler-requirements.md` (Status:
Implemented) requires "Validation must run before any durable work is written,
including molecule root [and step beads]." The title residual check honors this;
routing metadata does not — a bead with an unresolvable route is durable work
that was written without validation.

## Fix candidates

**A. Extend the #618 residual-var guard to routing metadata (and description) at
emission.** In `buildRecipeApplyPlan` and `buildFragmentApplyPlan`
(`graph_apply.go:252`, `:475`), after the title check, run `CheckResidualVars`
over the routing metadata values (`gc.run_target`, `gc.routed_to`, and ideally
all `node.Metadata` values + `node.Description`) and return the same
"unresolved variable(s)" error.
- Scope: ~15-25 LOC in `internal/molecule/graph_apply.go`; mirror in both plan
  builders; a couple of table-test cases.
- Trade-off: narrowest possible fix, mirrors the already-accepted #618 pattern
  and the formula-compiler-requirements invariant. Fails the sling loudly at
  emission (correct), but only catches placeholders whose var name is literally
  absent — a route that resolves to a *wrong-but-syntactically-valid* name still
  passes (that residue is #1512's job).

**B. Pre-sling route guard rejecting any resolved route matching `^\s*{{.*}}\s*$`.**
Add a dedicated check at the sling/route-stamp boundary (e.g. alongside
`graphroute.go:172` / the sling route resolution in `cmd/gc/cmd_sling.go`) that
rejects a `gc.run_target` / `gc.routed_to` still holding a `{{...}}` template
after substitution, before any durable write.
- Scope: ~10-20 LOC, one new predicate + call sites; overlaps the validation
  layer proposed in #1512.
- Trade-off: defends the route specifically and closer to where routing is
  authoritative, but adds a second guard location; risks drift from the #618
  title guard unless they share a helper.

**Recommendation:** **A**, because it reuses the existing `CheckResidualVars`
guard and lands at the same durable-write boundary the #618 fix already
established, directly satisfying the formula-compiler-requirements invariant. B
is worth adding as defense-in-depth (a small slice of #1512) but is secondary.
Maintainer's call.

- **#618 (closed, fixed):** established the residual-var guard, but scoped it to
  titles (`graph_apply.go:252/475`). This issue is the coverage gap that same fix
  left open for routing metadata. Not a regression of #618 — an uncovered sibling
  surface.
- **#1512 (open, `kind/feature`, `status/stale`) — Routing-target canonicalization
  and validation:** covers routes that *resolve* to malformed/unknown names
  (leading-slash `/refinery`, unqualified `dog` vs `gastown.dog`). That is the
  9/212 `gastown.dog` slice above and is **out of scope here**. #1512 does not
  address the 185/212 *unrendered-template* majority, and its full
  canonicalization design is much larger than the point-fix (A) proposed here.
  Fix A could be absorbed as a narrow first step; fix B is a slice of #1512's
  Layer 2.

## Adjacent / out of scope

- Enforcing `required = true` on vars (both title and route) — related footgun
  #618 also flagged; not fixed here.
- Canonicalizing resolved-but-malformed routing targets (leading slash,
  unqualified pool names, retired `gastown.dog`) — that is #1512.
- Reconciling/healing the ~212 already-emitted beads is an operational cleanup,
  separate from closing the emission gap.

vishnujayvel:
## Summary

The `#618` residual-`{{var}}` guard added to `buildRecipeApplyPlan` /
`buildFragmentApplyPlan` only inspects the bead **title**. A missing or
misspelled `--var` for a routing var leaves a literal, unrendered
`{{...}}` placeholder in `gc.run_target` / `gc.routed_to` /
`gc.execution_routed_to` instead of failing the sling, and the bead is
written durable with no error. The route then resolves to no agent, so
the bead lands with an empty/unclaimable assignee after an otherwise
"verified" `gc sling` — reported in #5060 with a 212-bead field sample,
185 of which (~87%) carried this exact fingerprint.

This extends the same guard to routing metadata at every site the
title-only guard already covers:

- `internal/molecule/graph_apply.go`: `buildRecipeApplyPlan` and
  `buildFragmentApplyPlan` (the graph-apply path)
- `internal/molecule/molecule.go`: `Instantiate` and `InstantiateFragment`
  (the sequential fallback path used when graph-apply is disabled, and by
  `Attach`, which calls `Instantiate` internally)

A new `validateResidualRoutingVars` helper (`internal/molecule/graph_apply.go`)
checks `gc.run_target`, `gc.routed_to`, `gc.execution_routed_to`, and their
`gc.deferred_*` counterparts (populated when `opts.DeferAssignees` defers
routing before this guard runs) for residual placeholders, reusing the
existing `formula.CheckResidualVars`. It's called from both the graph-apply
and sequential paths, immediately after the existing title check.

This is fix option A from the issue's own root-cause analysis (the
maintainer-recommended narrow fix, reusing the #618 guard rather than
adding a second validation layer at the graph-route boundary).

## Testing

```
ICU=/opt/homebrew/opt/icu4c@78 \
CGO_CXXFLAGS="-I$ICU/include" CGO_CFLAGS="-I$ICU/include" \
CGO_LDFLAGS="-L$ICU/lib -licuuc -licui18n -licudata" \
go test ./internal/molecule/... -count=1
go vet ./internal/molecule/...
golangci-lint run ./internal/molecule/...
```

All pass. Added four test cases (`TestInstantiateRejectsResidualRoutingMetadataVars`,
`TestInstantiateFragmentRejectsResidualRoutingMetadataVars`) mirroring the
existing `TestInstantiateRejectsResidualTitleVars` /
`TestInstantiateFragmentRejectsResidualTitleVars` table shape, covering both
the sequential and graph-apply paths, `gc.run_target` and `gc.routed_to`,
and a success case with the var resolved.

Verified the new tests actually catch the regression: reverted the guard
changes (keeping the tests), confirmed the four new test cases fail with
the exact pre-fix symptom ("should reject unresolved ... in gc.run_target"),
then restored the fix and confirmed a clean pass.

Also updated `TestCompiledReviewQuorumCollapsesRetryMachineryIntoNativeSteps`
(`internal/molecule/native_step_topology_test.go`), which called
`buildRecipeApplyPlan` directly with empty `Options{}` despite passing var
values to the preceding `formula.Compile` call. `formula.Compile`'s own doc
comment states substitution happens at instantiation time, not compile
time, and the real production path (`internal/sling/sling.go`) threads the
same `opts.Vars` map to both calls — so this test now does the same,
matching real usage rather than relying on an under-substituted input that
happened to slip past the title-only guard.

The two pre-existing failures in `internal/formula` (`TestDescriptionFileBaseDirResolvesSymlinkedParentWithMissingLeaf`,
`TestCanonicalExistingPathResolvesSymlinkedGrandparentWithTwoMissingLevels`)
are environmental (macOS `/tmp` → `/private/tmp` symlink resolution) and
reproduce identically on unmodified `main`; unrelated to this change.

Fixes #5060. Reported by @tdupu.

### DIFF
--- a/internal/molecule/graph_apply.go
+++ b/internal/molecule/graph_apply.go
@@ -127,6 +127,28 @@ func instantiateFragmentViaGraphApply(ctx context.Context, store beads.Store, ap
 	}, nil
 }
 
+// Routing fields the title-only guard below (#618) never covered — #5060.
+var residualVarRoutingMetadataKeys = []string{
+	beadmeta.RunTargetMetadataKey,
+	beadmeta.RoutedToMetadataKey,
+	beadmeta.ExecutionRoutedToMetadataKey,
+	DeferredRoutedToMetadataKey,
+	DeferredExecutionRoutedToMetadataKey,
+}
+
+func validateResidualRoutingVars(stepID string, metadata map[string]string) error {
+	for _, key := range residualVarRoutingMetadataKeys {
+		value := metadata[key]
+		if !strings.Contains(value, "{{") {
+			continue
+		}
+		if residual := formula.CheckResidualVars(value); len(residual) > 0 {
+			return fmt.Errorf("step %q: metadata %s contains unresolved variable(s) %s — missing or misspelled --var(s)?", stepID, key, strings.Join(residual, ", "))
+		}
+	}
+	return nil
+}
+
 func buildRecipeApplyPlan(recipe *formula.Recipe, opts Options) (*beads.GraphApplyPlan, bool, string, error) {
 	if recipe == nil {
 		return nil, false, "", fmt.Errorf("recipe is nil")
@@ -254,6 +276,10 @@ func buildRecipeApplyPlan(recipe *formula.Recipe, opts Options) (*beads.GraphApp
 			if residual := formula.CheckResidualVars(node.Title); len(residual) > 0 {
 				return nil, false, "", fmt.Errorf("step %q: bead title contains unresolved variable(s) %s — missing or misspelled --var(s)?", step.ID, strings.Join(residual, ", "))
 			}
+		}
+		// Same guard, extended to routing metadata — see #5060.
+		if err := validateResidualRoutingVars(step.ID, node.Metadata); err != nil {
+			return nil, false, "", err
 		}
 		if err := validateTimeoutMetadataVars(step.ID, node.Metadata); err != nil {
 			return nil, false, "", err
@@ -477,6 +503,10 @@ func buildFragmentApplyPlan(store beads.Store, recipe *formula.FragmentRecipe, o
 			if residual := formula.CheckResidualVars(node.Title); len(residual) > 0 {
 				return nil, fmt.Errorf("step %q: bead title contains unresolved variable(s) %s — missing or misspelled --var(s)?", step.ID, strings.Join(residual, ", "))
 			}
+		}
+		// Same guard, extended to routing metadata — see #5060.
+		if err := validateResidualRoutingVars(step.ID, node.Metadata); err != nil {
+			return nil, err
 		}
 		if err := validateTimeoutMetadataVars(step.ID, node.Metadata); err != nil {
 			return nil, err
--- a/internal/molecule/molecule.go
+++ b/internal/molecule/molecule.go
@@ -953,6 +953,11 @@ func Instantiate(ctx context.Context, store beads.Store, recipe *formula.Recipe,
 				return nil, fmt.Errorf("step %q: bead title contains unresolved variable(s) %s — missing or misspelled --var(s)?", step.ID, strings.Join(residual, ", "))
 			}
 		}
+		// Same guard, extended to routing metadata — see #5060.
+		if err := validateResidualRoutingVars(step.ID, b.Metadata); err != nil {
+			markFailed(store, createdIDs)
+			return nil, err
+		}
 		if err := validateTimeoutMetadataVars(step.ID, b.Metadata); err != nil {
 			markFailed(store, createdIDs)
 			return nil, err
@@ -1184,6 +1189,11 @@ func InstantiateFragment(ctx context.Context, store beads.Store, recipe *formula
 				markFailed(store, createdIDs)
 				return nil, fmt.Errorf("step %q: bead title contains unresolved variable(s) %s — missing or misspelled --var(s)?", step.ID, strings.Join(residual, ", "))
 			}
+		}
+		// Same guard, extended to routing metadata — see #5060.
+		if err := validateResidualRoutingVars(step.ID, b.Metadata); err != nil {
+			markFailed(store, createdIDs)
+			return nil, err
 		}
 		if err := validateTimeoutMetadataVars(step.ID, b.Metadata); err != nil {
 			markFailed(store, createdIDs)
--- a/internal/molecule/molecule_test.go
+++ b/internal/molecule/molecule_test.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/gastownhall/gascity/internal/beadmeta"
 	"github.com/gastownhall/gascity/internal/beads"
 	"github.com/gastownhall/gascity/internal/formula"
 	"github.com/gastownhall/gascity/internal/formulatest"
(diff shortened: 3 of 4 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277173193

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @vishnujayvel! This PR extends the molecule residual-variable guard to cover **routing metadata**, so an unresolved template placeholder in a routing key — `gc.run_target`, `gc.routed_to`, or the deferred-assignee keys `gc.deferred_routed_to` / `gc.deferred_execution_routed_to` — can no longer be written to durable bead storage as a non-empty but unroutable value. That closes #5060: before this change a misspelled or unsubstituted routing variable could survive into a persisted bead that no agent could ever pick up.

### Original PR Review
Decision: approve.

Specific gaps validated:
- The guard fires on both the sequential and graph-apply write paths, after variable substitution and before durable writes, so unroutable routing placeholders are rejected instead of persisted.
- The guard also covers `gc.deferred_routed_to` and `gc.deferred_execution_routed_to` — the keys that survive after `DeferAssignees` rewrites routing metadata — so deferred-assignee graph workflows are protected too.
- Production callers were confirmed to pass runtime vars through instantiation (`Options{Vars: ...}`) rather than relying on compile-time substitution.

Review findings addressed:
- Contract & Interface Fidelity (minor, high confidence): validate-only formula prevalidation (`ValidateRecipeRuntimeVars`) still checks declared vars and title placeholders but not routing-metadata placeholders, so a preview can report "valid" for a typo'd `gc.run_target` that will still fail later at instantiate time. This is a non-gating follow-up — the instantiate and graph-apply write paths already fail closed, so no unroutable bead is ever persisted.
- Test Evidence Quality (nit, high confidence): the deferred-routing branch of the guard is load-bearing but has no direct regression test; a suggested follow-up is a graph-apply test with `Options{DeferAssignees: true}` and an unresolved `gc.routed_to` asserting the error names the offending deferred key.

Remaining non-gating notes:
- The two notes above are the only follow-ups and neither blocks merge. Review found no blocker or major issues; the change is well-scoped (216 insertions across 4 files, including 171 lines of new tests).

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `24d84e38` has passing required GitHub checks (CI / required, CodeQL, Go module vulnerabilities, Mac regression, integration and cmd/gc process shards all green), and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/31148155030

### Review Iterations
1 review pass performed; approved on the first pass with no fix iterations required.

---

Issue #3206, opened 2026-06-07
defense-in-depth: decouple manual-reload reply latency from order dispatch in the controller tick


iwata-1116:
## Summary

Follow-up **defense-in-depth** hardening for #3201 (`gc reload --soft` / `gc doctor` hang; primary + complement fixes shipped in #3205). Umbrella: #3191.

Even with #3205's two fixes (index-miss gate on the per-order last-run fallback + `lastRunCache` survival across dispatcher rebuild), the manual-reload **reply remains structurally gated behind order dispatch**: the reply is only sent after the controller tick has run `dispatchOrders`. So reload-reply latency stays *coupled* to order-dispatch cost, which grows with order count. If the primary/complement ever regress, or order count grows substantially, the reload hang can recur.

This issue tracks decoupling that latency so a manual `gc reload` reply time is **independent of order count**, regardless of dispatch cost.

## Root cause (function-anchored; line numbers drift)

1. The manual-reload reply is sent only at `completeManualReload` (`cmd/gc/city_runtime.go`), which is invoked at the **end** of the reconcile tick — after `dispatchOrders` and the full session-reconcile phases.
2. The reply value (`manualReply`) is already computed **early** in the tick by `reloadConfigTraced`, right after the config-changed branch — long before `dispatchOrders` runs.
3. `dispatchOrders` (`cmd/gc/city_runtime.go` → `cr.od.dispatch`, `cmd/gc/order_dispatch.go`) iterates orders and its cost scales with order count (plus order-set rescan + tracking/nudge sweeps).
4. The dispatch-before-reply ordering is **intentional for AUTO ticks** (so due cooldown/cron formulas are not starved by slow startup/config-drift work) — but that rationale does **not** apply to a **manual-reload REPLY**, which a human/CLI is synchronously waiting on.

Net: a manual reload's reply waits on order dispatch even though the reply content is independent of it.

## Suggested fix (either)

- **(a)** On a manual-reload tick, send the already-computed reply **before** `dispatchOrders` / the expensive reconcile phases (preserving AUTO-tick ordering for non-manual ticks). Note: soft-reload acceptance currently amends the reply from post-reconcile desired state (`applySoftReloadAcceptance`), so a soft reload's early-reply path needs to account for that (e.g. reply early for the hard/applied/no-change cases, or restructure soft acceptance).
- **(b)** Run `dispatchOrders` under a **bounded sub-context** on manual-reload ticks so its cost cannot dominate reload-reply latency.

[...]

Issue #2897, opened 2026-06-01
gc reload: pool max_active_sessions increases not applied without a full restart (silent no-op or hang)


bourgois:
## Incident summary

2026-05-29: raising executor pool `max_active_sessions` from 6 to 8
(`pack 0.8.41`) shipped and the running controller ignored the new limit:

- `gc reload --soft` completed (exit 0) but did NOT scale up. `--soft`
  absorbs per-session config drift, not pool-cardinality changes → silent
  no-op.
- `gc reload` (no flag) BLOCKED for several minutes behind a busy
  controller reconcile loop. Was SIGTERM'd before it applied the new max.
- Only `gc stop && gc start` (full controller restart) applied the change
  — which drains all in-flight sessions, risking uncommitted work.

Same symptom seen on `pack 0.8.40` (a separate pool 2→3: config
resolved 3, runtime stayed 2).

---

## Problem statement

`gc reload` does not apply `max_active_sessions` increases without a full
restart. The failure modes are:

1. **Silent no-op** (`--soft`): the operator believes the pool resized; it
   did not; ready work goes unserved.
2. **Indefinite hang** (plain reload): a busy reconcile loop never yields
   to process the reload signal, so `gc reload` blocks the operator's
   terminal indefinitely.
3. **No warning**: neither variant warns that a pool-cardinality change
   was detected but not applied. The operator has no signal that a restart
   is required.

---

### FR-PM.1 — Reconcile tick refreshes pool bounds in memory

On a `gc reload` (or any config-reload signal), the supervisor SHALL
re-read `max_active_sessions` from the resolved config and update its
in-memory pool-bounds state. The next reconcile tick uses the fresh bounds
without requiring a controller restart.

**Behavior:**
- If `max_active_sessions` increased: the reconcile tick that follows
  reload immediately sees the higher ceiling and may spawn additional
  sessions up to the new max (subject to demand and FR-PM.2).
- If `max_active_sessions` decreased: the bound is updated lazily (FR-PM.4);
  no existing sessions are killed.
- Neither change drains in-flight sessions.

### FR-PM.2 — scaleCheck counts routed demand independently

The `scaleCheck` function (supervisor Session Loop) SHALL compute `desired`
from routed-demand independently of the current `active` count. That is,
`desired = min(max_active_sessions, routed_work_count)` — NOT
`desired = active + deficit` from stale counts. This ensures that after a
bound increase, the scaleCheck correctly sees `desired > active` and spawns.

[...]

PR #5075, opened 2026-08-07
fix(runtime): surface max_active_sessions changes on gc reload (#2897)

vishnujayvel:
## Summary

Implements **per-pool changed-bound reporting + attrition drain warning** for
`gc reload` (#2897): when reload applies a new config, the reply surfaces each
pool whose resolved `max_active_sessions` changed (`old → new`). If the new
bound is a finite decrease and currently-active sessions (from the sessions
bead store) exceed that bound, a warning notes that excess sessions wind down
via normal reconcile attrition (they are not killed; new spawns honor the lower
bound — FR-PM.4).

This is **Message/Warnings construction only**. There is zero behavior change
to reconcile, accept, or apply logic.

**Not in this PR:** the issue's full FR-PM.3 sample (no restart-recommended
warning, no unchanged-pool listing). Only changed bounds are listed; drain
warnings are best-effort and omit when the sessions store is unavailable
rather than inventing counts from provider process-name heuristics.

Credits: issue reported by @bourgois.

### Scope claim

This PR implements **only** the operator-visible reload feedback for pool
bounds (changed-bound lines + optional drain warning). It does **not** claim
the full FR-PM.3 sample from the issue (restart-recommended warning, listing
unchanged pools). Accept/reply latency portions of #2897 appear already
addressed on main by prior work cited in source comments:

- bead `ga-8nbr` — reload acceptance must not be starved by a slow reconciler
- #3206 — manual hard reload replies before dispatch / desired-state rebuild
- #3667 — demand snapshot refresh for routed work

FR-PM.1/FR-PM.2 (in-memory bound refresh + scaleCheck independence) are
**not** claimed here; operators already get live bound application via
the existing config swap + next reconcile tick using
`ResolvedMaxActiveSessions`. This PR only makes that change visible.

### Design notes

- Session counting runs only when at least one bound **decreases to a finite
  value** (drain warnings are the only consumer). Unchanged-config reloads and
  increase-only reloads do zero session-store work from this feature.
- Counts come **only** from the sessions bead store. If the store is
  unavailable, drain warnings are omitted honestly — no provider-list name
  parsing or last-dash / pool-slot heuristics.

## Testing

- [x] Scoped: `go test ./cmd/gc/ -run 'TestReload|TestPoolMax|TestAppendPool|TestFormatResolved|TestCountActiveSessions|TestPoolBound'` — pass (~23s)
- [x] `gofmt -l cmd/gc/` empty for touched files; `go vet ./cmd/gc/` clean
- [x] Full `go test ./cmd/gc/` exceeds the Go default **10m test-binary timeout** in this environment. The 4 observed failures were A/B-confirmed identical on `origin/main` (not introduced by this branch):
  - `TestBdRuntimeEnvManagedCityProjectsHostOverride` — port-allocation / env pollution
  - `TestBdRuntimeEnvForRigInheritedManagedCityProjectsHostOverride` — same
  - `TestResolveImportRootFallsBackToStandalonePackDir` — macOS `/var` vs `/private/var` symlink path
  - `TestResolveImportRootPrefersNearestPackUnderCity` — same
- [ ] `make check` (maintainer CI)
- [ ] `make test-integration` if maintainers want runtime/controller coverage

### Local results (scoped reload/pool suite)

```
ok  github.com/gastownhall/gascity/cmd/gc  ~23s
```

Coverage includes:
- bound increase reported
- bound decrease with active > new bound → drain warning (bead-store counts)
- no pool changes → no pool lines
- default resolution via workspace/rig max (no agent-level max)
- finite-decrease gate for session counting
- no store → empty counts (no provider heuristic attribution)
- bead-store open sessions counted; closed sessions ignored

## Checklist

- [x] Linked an issue (#2897)
- [ ] Updated docs for user-facing changes (optional: short note under reload CLI docs if maintainers want)
- [x] No breaking changes

### DIFF
--- a/cmd/gc/city_runtime.go
+++ b/cmd/gc/city_runtime.go
@@ -2245,9 +2245,8 @@ func (cr *CityRuntime) reloadConfigTraced(
 		fmt.Fprintf(cr.stderr, "%s: orders reloaded: %s\n", cr.logPrefix, orderSummary) //nolint:errcheck // best-effort stderr
 	}
 
-	// FR-PM.3: diff pool max_active_sessions before the in-memory swap so the
-	// reload reply surfaces old → new bounds. Active counts use the still-
-	// current config for template attribution.
+	// Diff pool bounds and count active sessions against the still-current
+	// config before the swap below (#2897 FR-PM.3).
 	poolBoundChanges := poolMaxActiveSessionChanges(cr.cfg, nextCfg)
 	poolActiveCounts := cr.countActiveSessionsByTemplate(cr.cfg)
 
--- a/cmd/gc/city_runtime.go
+++ b/cmd/gc/city_runtime.go
@@ -2245,6 +2245,12 @@ func (cr *CityRuntime) reloadConfigTraced(
 		fmt.Fprintf(cr.stderr, "%s: orders reloaded: %s\n", cr.logPrefix, orderSummary) //nolint:errcheck // best-effort stderr
 	}
 
+	// FR-PM.3: diff pool max_active_sessions before the in-memory swap so the
+	// reload reply surfaces old → new bounds. Active counts use the still-
+	// current config for template attribution.
+	poolBoundChanges := poolMaxActiveSessionChanges(cr.cfg, nextCfg)
+	poolActiveCounts := cr.countActiveSessionsByTemplate(cr.cfg)
+
 	cr.serviceStateMu.Lock()
 	cr.cfg = nextCfg
 	cr.sp = nextSp
@@ -2318,6 +2324,7 @@ func (cr *CityRuntime) reloadConfigTraced(
 	message := fmt.Sprintf("Config reloaded: %s (rev %s)",
 		configReloadSummary(oldAgentCount, oldRigCount, len(nextCfg.Agents), len(nextCfg.Rigs)),
 		shortRev(result.Revision))
+	message, warnings = appendPoolMaxBoundFeedback(message, warnings, poolBoundChanges, poolActiveCounts)
 	fmt.Fprintln(cr.stdout, message) //nolint:errcheck // best-effort stdout
 	telemetry.RecordConfigReload(ctx, result.Revision, string(source), string(reloadOutcomeApplied), len(warnings), nil)
 	if trace != nil {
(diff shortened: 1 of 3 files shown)


quad341 (approved):
@vishnujayvel — this is careful work and the care shows in the specifics.

Two things stood out. The scope discipline: you drew a clear line around operator-visible feedback, said plainly what you're *not* claiming from #2897's FR-PM.3 sample, and cited the prior work (ga-8nbr, #3206, #3667) covering the latency side. And you refused to invent session counts from provider process-name heuristics when the bead store isn't there — omitting the warning instead is the right call, and it's the kind of restraint that's easy to skip. The `poolBoundChangesNeedActiveCounts` gate is a nice touch: no-op and increase-only reloads do zero store work.

One edge case I'll take on our side. Because a single decrease loads counts for *every* changed pool, a pool whose bound went **up** can still show more open session beads than its new bound and get told it's waiting on attrition. I'll tighten `poolMaxBoundDrainWarning` to only fire on a real decrease, add the mixed increase/decrease regression case, and re-run vet plus your scoped suite before merging — your commit stays, with a co-authored fix on top.

Everything else lands as written. Build, vet, gofmt and your scoped tests pass here, and CI's twelve `cmd/gc` shards genuinely ran yours. Thanks for this.


---

quad341:
@vishnujayvel — thank you for this, and especially for the follow-up commit. You caught the increase-vs-decrease edge yourself and fixed it with a guard that mirrors `poolBoundChangesNeedActiveCounts`, plus two regression tests covering the mixed case and unlimited→finite. That saved us the round trip entirely.

Two choices here are why this ships as-is. You used `ResolvedMaxActiveSessions` instead of rolling your own resolution, so your keys land in the same `QualifiedName` space the reconciler already uses — I traced that through `normalizedSessionTemplateInfo` and it lines up, including for pool-slot beads. And when the sessions bead store isn't available you omit the warning rather than guessing counts from provider process names. Refusing to invent a number is the harder call and the right one.

The scope discipline in the PR body helped too — being explicit about what you were *not* claiming from FR-PM.3 made this fast to review.

I'll take the one open box on my side: a short note in `engdocs/design/gc-reload-design.md` covering the new pool lines and the drain warning, since the `applied` row there still implies a single-line message. Then merging. Your commits stay exactly as they are.


---

vishnujayvel:
Thanks for the thorough review and for taking the design-doc note. For the record: the shard 3 failure is TestBuildPinnedBDBinaryForTestsUsesGoModSource failing on go install of the pinned bd binary, and the Evidence watchdog cascades from that, so neither is from this change. Happy to rebase on current main or re-push to retrigger CI if that helps unblock the merge. Otherwise nothing pending on my side.

PR #4693, opened 2026-07-27
fix(core): make mol-do-work drain step provider-agnostic (GC_BEAD_ID → GC_TRIGGER_BEAD_ID fallback)

figgeous:
## What

The `mol-do-work` drain step self-closes its bead only when `$GC_BEAD_ID` is set. The **claude provider never exports `GC_BEAD_ID`** — it exports `GC_TRIGGER_BEAD_ID` (and `GC_TRIGGER_WORK_BEAD_ID`). So under the claude provider the drain guard is always false.

**Impact:** the drain bead stays `open`, the downstream `workflow-finalize` step (which `needs` it) never fires, and the whole molecule wedges `in_progress` forever — unless an operator notices the unset var and substitutes the id by hand.

## Fix

Fall back to `GC_TRIGGER_BEAD_ID` when `GC_BEAD_ID` is unset:

```bash
DRAIN_BEAD_ID="${GC_BEAD_ID:-${GC_TRIGGER_BEAD_ID:-}}"
```

Smallest, provider-agnostic change; no provider-contract change required.

## Why it's safe

`GC_TRIGGER_BEAD_ID` is set per session from the session's own `TriggerBeadID` (`cmd/gc/build_desired_state.go:3073-3078`). Each molecule step is its own session bead, so for the drain step `GC_TRIGGER_BEAD_ID` names the **drain bead itself** — the exact bead the guard intends to close. Verified empirically: during drain step `ga-kw0y`, `env | grep ^GC_` showed `GC_TRIGGER_BEAD_ID=ga-kw0y` with `GC_BEAD_ID` unset.

## Scope

- Only `mol-do-work.toml` carries this drain prose; siblings `mol-polecat-*` do not (grep confirmed) — one-file fix.
- The `do-work` step (line ~122) has a structurally similar `$GC_BEAD_ID` guard but a different purpose (wrapper-bead close, with a `!= WORK_BEAD_ID` sibling condition) and no repro attached. Left unchanged here to keep the PR scoped to the confirmed break.

## Acceptance

- A `mol-do-work` run under the claude provider ends with its drain bead **closed** without manual id substitution.
- The molecule root reaches `closed` (`workflow-finalize` fires).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5079, opened 2026-08-07
Stale-Dolt janitor formula never runs unattended under the claude provider — fall back to the trigger work-bead id

figgeous:
The stale-Dolt-database janitor never manages to run on its own under the claude provider: the very first line of its script demands an environment variable that provider never sets, so every unattended run dies before doing any work and a human has to finish the job by hand (this has happened on every one of its 34 recorded runs). This change lets the script accept the id the provider *does* set, falling back gracefully, while still failing loudly if no id is available at all.

### Root cause

- The formula's script requires `GC_BEAD_ID` on its first working line and aborts if it is missing.
- The claude provider exports the claimed work-bead id as `GC_TRIGGER_WORK_BEAD_ID` (and `GC_TRIGGER_BEAD_ID`), never as `GC_BEAD_ID`, so the guard always fired.

### Fix

- Accept `GC_BEAD_ID` when present, otherwise fall back to `GC_TRIGGER_WORK_BEAD_ID`; abort with a clear message only when neither is set.
- Same pattern the core do-work formula adopted in #4693.

### Tests

- New test: the rendered script completes end-to-end (scan event, bead close, drain-ack) with only the trigger variable set. ✅
- New test: the script still aborts loudly with neither variable set. ✅
- Full `examples/bd/dolt` suite passes (108s, includes the rendered-shell contract tests updated for the new guard line).

---

<details>
<summary>Machine detail</summary>

**Changed files** (commit `3c1011e5c`, branch `fix/ga-oyp60-gc-bead-id-fallback`, 2 files +96/−3)
- `examples/bd/dolt/formulas/mol-dog-stale-db.toml` *(±1)* — line 88: `WORK_BEAD="${GC_BEAD_ID:?...}"` → `WORK_BEAD="${GC_BEAD_ID:-${GC_TRIGGER_WORK_BEAD_ID:?no work bead id in env}}"`
- `examples/bd/dolt/stale_db_formula_test.go` *(+95/−2)* — updated the two contract assertions that pinned the old guard string; added `TestStaleDBFormulaFallsBackToTriggerWorkBeadID` (full mocked run, only `GC_TRIGGER_WORK_BEAD_ID=bead-trigger` in env, asserts `bd close bead-trigger` + scan event + drain-ack) and `TestStaleDBFormulaAbortsWithoutAnyWorkBeadID` (neither var set, asserts non-zero exit and the `no work bead id in env` message)

**Approach / blast radius**
Per-formula fallback rather than making the claude provider export `GC_BEAD_ID`; this matches the choice already merged for `internal/bootstrap/packs/core/formulas/mol-do-work.toml` in #4693 (`DRAIN_BEAD_ID="${GC_BEAD_ID:-${GC_TRIGGER_BEAD_ID:-}}"`, line 145) and is provider-agnostic. `GC_TRIGGER_WORK_BEAD_ID` is the semantically exact source for a claimed work bead. Providers that do export `GC_BEAD_ID` are unaffected (it wins the fallback chain). Only this one formula file is touched; the companion `orders/mol-dog-stale-db.toml` has no id references. A former byte-identical copy at `examples/dolt/formulas/` has since been deleted from the tree, so this is the only remaining instance.

**Verification (pre-merge)**
- `go test ./examples/bd/dolt/ -run StaleDB -count=1` → ok (1.1s)
- `go test ./examples/bd/... -count=1` → ok (107.9s)

**Post-merge verification**
- On the next unattended `mol-dog-stale-db` cron wisp under the claude provider (runs every 4h per `orders/mol-dog-stale-db.toml`): the work bead should reach CLOSED with the `gc.dolt.cleanup.v1` scan JSON appended as a note, with **no manual id substitution**. Signal: `bd show <wisp-id>` shows status closed and an `## scan` note; direction: closes unattended where all 34 prior runs stayed open until hand-run.
- `gc event` stream should show `mol-dog-stale-db.scan` / `.done` events emitted by an unattended run (previously absent because the script died on line 3).

</details>

### DIFF
--- a/examples/bd/dolt/formulas/mol-dog-stale-db.toml
+++ b/examples/bd/dolt/formulas/mol-dog-stale-db.toml
@@ -85,7 +85,7 @@ Run this as one shell script so every derived value remains in the same process:
 ```bash
 set -euo pipefail
 
-WORK_BEAD="${GC_BEAD_ID:?GC_BEAD_ID required (set by gc hook); aborting}"
+WORK_BEAD="${GC_BEAD_ID:-${GC_TRIGGER_WORK_BEAD_ID:?no work bead id in env}}"
 TMP_DIR=$(mktemp -d "${TMPDIR:-/tmp}/dolt-cleanup.XXXXXX")
 DRAIN_ACKED=0
 
--- a/examples/bd/dolt/stale_db_formula_test.go
+++ b/examples/bd/dolt/stale_db_formula_test.go
@@ -36,7 +36,7 @@ func TestStaleDBFormulaRuntimeContract(t *testing.T) {
 	desc := f.Steps[0].Description
 	for _, want := range []string{
 		`set -euo pipefail`,
-		`WORK_BEAD="${GC_BEAD_ID:?GC_BEAD_ID required`,
+		`WORK_BEAD="${GC_BEAD_ID:-${GC_TRIGGER_WORK_BEAD_ID:?no work bead id in env}}"`,
 		`TMP_DIR=$(mktemp -d`,
 		`trap cleanup EXIT`,
 		`drain_ack_once()`,
@@ -83,7 +83,7 @@ func TestStaleDBFormulaRenderedShellIsStrictAndValid(t *testing.T) {
 	script := renderStaleDBFormulaShell(t)
 	for _, want := range []string{
 		`set -euo pipefail`,
-		`WORK_BEAD="${GC_BEAD_ID:?GC_BEAD_ID required`,
+		`WORK_BEAD="${GC_BEAD_ID:-${GC_TRIGGER_WORK_BEAD_ID:?no work bead id in env}}"`,
 	} {
 		if !strings.Contains(script, want) {
 			t.Fatalf("rendered script missing %q", want)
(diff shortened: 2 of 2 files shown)


figgeous:
Closing this as superseded — the same bug was fixed independently and more thoroughly by #5687 (`d6e4339f7`, "single-source the hook-claim startup protocol in a claim-protocol fragment").

`examples/bd/dolt/formulas/mol-dog-stale-db.toml` on `main` now resolves the work bead as:

```sh
WORK_BEAD="${GC_BEAD_ID:-${GC_TRIGGER_BEAD_ID:-$(gc hook current --id-only)}}"
[ -n "$WORK_BEAD" ] || { echo "cannot resolve work bead id: ..." >&2; exit 1; }
```

That is strictly better than what this PR proposed. `GC_TRIGGER_BEAD_ID` and `GC_TRIGGER_WORK_BEAD_ID` are set from the same `triggerBeadID` (`cmd/gc/session_lifecycle_parallel.go:1228-1229`), so the two fallbacks are equivalent — but #5687 adds a third `gc hook current --id-only` fallback and a loud explicit error instead of this PR's `${...:?}`, and it standardises the idiom across the codebase (`cmd/gc/cmd_hook_current.go`, `docs/reference/cli.md`, `internal/bootstrap/packs/core/formulas/mol-do-work.toml`).

The test coverage added here is also covered upstream, by `TestStaleDBFormulaResolvesWorkBeadViaHookCurrentChain` and `TestStaleDBFormulaFailsLoudlyWhenNoWorkBeadIDResolvable`.

I confirmed the original failure (the stale-Dolt janitor never running unattended under the claude provider, because `GC_BEAD_ID` is unset on a demand-spawned pool seat) is resolved by the current `main`. Nothing left to rebase — the merge conflict here was the two fixes colliding on the same lines, not drift. Thanks to whoever picked it up.

PR #5088, opened 2026-08-07
fix(bd): honor the workspace bd pin in the gc bd passthrough

julianknutsen:
## Problem

`gc bd` resolves the binary it execs with a bare `exec.LookPath("bd")` (`cmd/gc/cmd_bd.go:324`). It never consults the city's `[workspace.env]` PATH pin, so a city that deliberately pins a bd build — because its work store is bound to a backend only that build speaks — has the pin silently ignored by the passthrough, while every other bd path in the tree honors it.

The symptom is a passthrough that fails on a store the rest of `gc` reads fine:

```
$ gc bd show <id>
storage backend "postgres" is no longer supported
```

The ambient `bd` first on PATH was a build that rejects that backend. The identical command with the pin prefixed on PATH succeeded — the store was healthy the whole time; only the passthrough picked the wrong binary.

## Fix

Resolve the passthrough's binary the way `requireBdBinaryForCity` (`cmd/gc/main.go`) already resolves it: when the scope carries a complete storage binding (`scopeHasCompleteStorageBinding`), prefer `workspacePinnedBdBinary` if it yields an absolute path; otherwise fall back to the ambient lookup.

Rather than copy that logic a second time, both call sites now share a new `resolveBdBinaryForCity` in `cmd/gc/bd_env.go` — next to `workspacePinnedBdBinary`, which it wraps — so the passthrough and the managed-runner precondition cannot drift apart again.

Behavior is unchanged whenever there is no complete binding or no pin: same `exec.LookPath("bd")`, same `gc bd: bd not found in PATH` message, byte for byte. `requireBdBinaryForCity` keeps its own distinct remediation text (`... (install beads or set GC_BEADS=file)`) via the `errBdNotOnPath` sentinel. A pin that is configured but unresolvable now surfaces its real error instead of being masked as a missing binary.

## Tests

`TestGcBdPassthroughResolvesBdBinary` drives `doBd` end to end against two stand-in `bd` scripts that announce which one ran, covering both arms:

- **complete binding + workspace pin** → the pinned absolute binary runs
- **no binding, no pin** → the ambient binary runs, unchanged

Red before the production change (call site reverted to `exec.LookPath`, test unchanged):

```
--- FAIL: TestGcBdPassthroughResolvesBdBinary (0.59s)
    --- FAIL: TestGcBdPassthroughResolvesBdBinary/complete_binding_runs_the_workspace-pinned_bd (0.39s)
        cmd_bd_test.go:2508: executed bd = "ambient-bd", want workspace-pinned "pinned-bd"
FAIL
```

The fallback arm passed while red, confirming the second arm is a genuine no-regression guard rather than a restatement of the fix.

Green after:

```
--- PASS: TestGcBdPassthroughResolvesBdBinary (0.69s)
    --- PASS: TestGcBdPassthroughResolvesBdBinary/complete_binding_runs_the_workspace-pinned_bd (0.46s)
    --- PASS: TestGcBdPassthroughResolvesBdBinary/no_binding_falls_back_to_ambient_bd (0.23s)
```

## Gates

- `go build ./...` clean
- `go vet ./cmd/gc/` clean
- `go test ./cmd/gc/ -run 'Bd|BD|Workspace|Pin' -count=1` → `ok ... 73.443s`
- `golangci-lint run ./cmd/gc/` → `0 issues`
- `golangci-lint fmt --diff ./cmd/gc/` → no diff

### DIFF
--- a/cmd/gc/bd_env.go
+++ b/cmd/gc/bd_env.go
@@ -118,6 +118,37 @@ func workspacePinnedBdBinary(cityPath string) (string, error) {
 	return "", fmt.Errorf("workspace.env PATH is configured but contains no executable bd at an absolute path")
 }
 
+// errBdNotOnPath reports that neither the workspace pin nor the ambient
+// lookup produced a bd executable. Callers phrase their own remediation.
+var errBdNotOnPath = errors.New("bd not found in PATH")
+
+// resolveBdBinaryForCity resolves the bd executable a city runs. A scope
+// carrying a complete storage binding runs the binary its workspace PATH
+// pins, because only that build speaks the bound backend; every other scope
+// keeps the ambient lookup. An ambient miss is errBdNotOnPath so callers can
+// phrase their own remediation; a pin that is configured but unresolvable is
+// returned verbatim rather than masked as a missing binary.
+func resolveBdBinaryForCity(cityPath string) (string, error) {
+	completeBinding, err := scopeHasCompleteStorageBinding(scopeMetadataJSONPath(cityPath))
+	if err != nil {
+		return "", err
+	}
+	if completeBinding {
+		pinned, err := workspacePinnedBdBinary(cityPath)
+		if err != nil {
+			return "", err
+		}
+		if pinned = strings.TrimSpace(pinned); filepath.IsAbs(pinned) {
+			return pinned, nil
+		}
+	}
+	bdPath, err := exec.LookPath("bd")
+	if err != nil {
+		return "", errBdNotOnPath
+	}
+	return bdPath, nil
+}
+
 func bdStoreForCity(dir, cityPath string) *beads.BdStore {
 	cfg, err := loadCityConfig(cityPath, io.Discard)
 	if err != nil {
--- a/cmd/gc/bd_env.go
+++ b/cmd/gc/bd_env.go
@@ -122,18 +122,19 @@ func workspacePinnedBdBinary(cityPath string) (string, error) {
 // lookup produced a bd executable. Callers phrase their own remediation.
 var errBdNotOnPath = errors.New("bd not found in PATH")
 
-// resolveBdBinaryForCity resolves the bd executable a city runs. A scope
-// carrying a complete storage binding runs the binary its workspace PATH
-// pins, because only that build speaks the bound backend; every other scope
-// keeps the ambient lookup. An ambient miss is errBdNotOnPath so callers can
-// phrase their own remediation; a pin that is configured but unresolvable is
-// returned verbatim rather than masked as a missing binary.
-func resolveBdBinaryForCity(cityPath string) (string, error) {
-	completeBinding, err := scopeHasCompleteStorageBinding(scopeMetadataJSONPath(cityPath))
+// resolveBdBinaryForScope resolves the bd executable a scope's commands run.
+// A scope bound to a complete storage binding runs the binary its workspace
+// PATH pins, because only that build speaks the bound backend; every other
+// scope keeps the ambient lookup. An ambient miss is errBdNotOnPath so
+// callers can phrase their own remediation; a pin that is configured but
+// unresolvable for a scope that needs it is returned verbatim rather than
+// masked as a missing binary.
+func resolveBdBinaryForScope(cityPath, scopeRoot string) (string, error) {
+	bound, err := scopeBindsPinnedBdBinary(cityPath, scopeRoot)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


julianknutsen:
Pushed `4a20fef` addressing both MAJOR review findings. They had one root cause: `resolveBdBinaryForCity` probed the **city's** storage binding, while `doBd` is scope-aware and already holds `target.ScopeRoot` two lines earlier.

## 1. Regression vs base — a city-level binding fault took every rig-scoped `gc bd` offline

Because `doBd` calls the resolver *before* `bdCommandEnv` and surfaces its errors, `scopeHasCompleteStorageBinding` and `workspacePinnedBdBinary` failures hard-failed every scope. The rig env path deliberately swallows exactly those errors — `bdRuntimeEnvForRigWithErrorRecoveryContext` discards `cityErr` whenever `rigDoltlite || (cityDoltlite && !scopeOverridesCityBackend(...))` — so on base those cities kept working. Reproduced two ways, both `exit 0` on base and `exit 1` on the PR head:

- half-written city `.beads/metadata.json` (`backend` + `storage_endpoint`, no `storage_database`) + a doltlite rig → `gc bd: partial beads storage binding …`
- a complete city binding + `[workspace.env] PATH` containing no bd + a doltlite rig → `gc bd: workspace.env PATH is configured but contains no executable bd at an absolute path`

City scope was never affected (base already failed at `bdCommandEnv` with identical text), so the blast radius was exactly the rig scopes.

## 2. Incomplete fix — `gc bd --rig` still exec'd the ambient bd

A rig carrying its own complete binding got the ambient binary, so the symptom this PR fixes (`storage backend "postgres" is no longer supported`) stayed reachable through the `--rig` form the command's own help documents. Per-rig bindings are first-class in the tree (`beads_provider_lifecycle.go:1845`, `bd_env.go:442`), and the new helper's doc comment said "scope" while the code read the city.

## The fix

`resolveBdBinaryForCity` → `resolveBdBinaryForScope(cityPath, scopeRoot)`, with the decision in a new `scopeBindsPinnedBdBinary`:

1. the target scope's own complete binding wins, and its probe errors still surface;
2. otherwise inherit the city's binding the way `applyCanonicalScopeBackendEnv` does — inherit unless `scopeOverridesCityBackend`;
3. a **city-level** probe failure degrades to `exec.LookPath("bd")` rather than failing a command that never reads that binding.

`requireBdBinaryForCity` passes `cityPath` for both arguments, so it is unchanged.

### One deliberate behavior decision

The review also flagged that with a *valid* pin, rig-scoped `gc bd` had started exec'ing the city-pinned bd where base exec'd the ambient one. Confirmed, and **reverted to the base behavior**: a doltlite rig in a bound city keeps the ambient bd. Its store is not the bound one, `bdCommandRunnerForRig` sets no `BD_BIN` for it, and the runtime env the passthrough hands the child says `doltlite` — running the city's postgres-speaking pin against it would be arbitrary. `TestResolveBdBinaryForScope/rig_overriding_the_city_backend_keeps_the_ambient_bd` states this, so it is now a decision rather than a side effect.

## Tests

New `TestGcBdPassthroughResolvesBdBinaryForRigScope` (end-to-end through `doBd`) and `TestResolveBdBinaryForScope` (unit). Red on the previous head:

- `executed bd = "ambient-bd", want workspace-pinned "pinned-bd"` — finding 2
- `doBd(--rig dl) = 1, want 0; … gc bd: partial beads storage binding …/.beads/metadata.json: backend, storage_endpoint, and storage_database must all be non-empty` — finding 1
- `executed bd = "pinned-bd", want ambient "ambient-bd"` — the behavior change above

Plus `TestRequireBdBinaryForCityErrorText`, which pins the three preserved messages — the `errBdNotOnPath` sentinel's only reason to exist is that an ambient miss gets the remediation hint while a pin or binding fault is returned verbatim, and nothing asserted that.

Every guard was mutation-checked. Reverting to city-only probing fails 6 subtests; dropping the `scopeOverridesCityBackend` guard fails 2; swallowing the scope's *own* probe error fails the city-scope and preflight guards; flattening the pin error into the sentinel fails the verbatim-text guard.

## Gates

`go build ./...`, `go vet ./cmd/gc/`, `golangci-lint run ./cmd/gc/` (2.10.1 — 0 issues; the pinned 2.12.0 is built with go1.25 and refuses this repo's go1.26.5 target), and the full `go test ./cmd/gc/` package: `ok … 729.844s`, 0 failures. Pre-commit hooks ran on the commit.

PR #4677, opened 2026-07-26
Refuse ambient-city Dolt ports in test binaries

quad341:
## What this changes

Go test binaries now refuse to proceed when a surviving Dolt-port environment variable matches the managed Dolt port of a city discoverable from their current working directory. Previously, the guard only recognized the static local port `3307`; tests launched beneath a live city could instead discover that city’s actual port from runtime state and escape the guard.

The new check is additive. It keeps the existing environment-based arm, walks upward for the nearest `city.toml`, and reads that city’s `.gc/runtime/packs/dolt/dolt-state.json` with a minimal standard-library parser. This makes the comparison self-relative without introducing another fleet-specific magic port.

## Review notes

- Scope is limited to `internal/testenv`; production command behavior and configuration are unchanged.
- The runtime-state arm accepts only a positive JSON `port` and only adds refusal conditions.
- Synthetic end-to-end coverage uses ports `19999` and `20000`, distinct from both `3307` and the fleet port used during the incident.
- `GC_ALLOW_PROD_DOLT_PORT_IN_TESTS=1` remains the explicit opt-out for deliberate local-Dolt tests.
- A separately executed production `gc` subprocess does not import `internal/testenv`; that known boundary is documented and remains out of scope.

## Test plan

- [x] `go build ./...` and `go vet ./...`
- [x] Focused ambient-city and legacy production-port tests (36 subtests)
- [x] Full `internal/testenv` package and resource-census ledger-sync test
- [x] `make test-fast-parallel` (9/9 jobs)
- [x] Release gate: [`release-gates/ga-ceutvg-ambient-city-dolt-port-gate.md`](release-gates/ga-ceutvg-ambient-city-dolt-port-gate.md)

Issue #3947, opened 2026-07-05
bug: orders: never-run cron order can't bootstrap — catch-up scan skipped when lastRun is zero (residual gap in #2721)


sjarmak:
## Summary

A cron order that has **never fired** can never bootstrap: `checkCron`'s catch-up scan (added by #2721 to fix exactly this class of miss) is skipped when `lastRun` is zero, so a freshly-installed narrow-window order (single minute/day) is left depending on the exact-minute luck that #2721 was written to eliminate. If its first scheduled minute never coincides with a controller evaluation, it fires never — and no restart helps, because `lastRun` stays zero.

## Symptom

A daily-once cron order installed and enabled, with correct config, accumulates **zero** order history and silently never fires. `gc order check` shows `cron: schedule not matched` at all times. Unlike #2599 (warm orders that drop after multi-day uptime and recover on supervisor restart), a restart does **not** help here — there is no prior `lastRun` to recover to.

Observed 2026-07-05 on `janitor-worktree-gc` (`schedule = "20 4 * * *"`), installed 2026-07-04:

- Missed both of its first two 04:20 windows (07-04 and 07-05).
- `gc order history janitor-worktree-gc` → `No order history`.
- Sibling orders fire fine: `bead-janitor` (`0 */4 * * *`) lands at `:14`, `:36`, `:24` past the hour — never on `:00` — because it is *catching up* via #2721's path (b), which works only because it has a non-zero `lastRun`.
- A manual `gc order run janitor-worktree-gc` established a first `lastRun`; the order is now expected to self-heal on subsequent windows.

## Reproduction

1. Install a new exec/cron order with a narrow daily schedule (one specific minute/day, e.g. `schedule = "20 4 * * *"`), enabled, in a running city.
2. Let the controller run across that scheduled minute. The controller's order-evaluation cadence is coarse relative to one minute (see #2604: synchronous dispatch can push ticks to ~142 s), so the exact minute is usually not evaluated.
3. `gc order history <name>` → no history; `gc order check` → `cron: schedule not matched`.
4. Restart the supervisor — still no fire.

Result: expected the order to catch up its missed daily occurrence (as #2721 provides for orders that have run before) → got a permanent silent miss until a manual `gc order run`.

## Root cause

`internal/orders/triggers.go` `checkCron` (`/home/ds/gascity-main` @ `d9225156f`). The catch-up scan is gated on a non-zero `lastRun` (triggers.go:155-166):

```go
[...]

PR #2721, opened 2026-05-29
fix(orders): cron triggers catch up missed occurrences

Wldc4rd:
## Problem
`internal/orders/triggers.go` `checkCron` uses stateless exact-minute matching — an order is `Due` only when the controller evaluates it *during* the matching minute. There is no catch-up, so a scheduled occurrence is silently dropped whenever no evaluation lands in that exact minute.

In a live deployment this made a `0 */4 * * *` order miss **every** 4-hour boundary for days: the controller's order-evaluation cadence rarely coincides with a narrow once-per-4h minute, so the `:00` minute is repeatedly skipped (the next eval at `:01` returns `"cron: schedule not matched"`, with no retry). Cooldown triggers are unaffected because `elapsed >= interval` catches up regardless of *when* the evaluation happens — which is exactly why cooldown orders fire reliably while the lone cron order never did.

## Fix
Give `checkCron` catch-up, mirroring cooldown's elapsed-based behavior:
- Current minute matches → fire (unchanged; retains the already-run-this-minute guard).
- Current minute doesn't match, but a scheduled minute elapsed since `lastRun` → fire (`"cron: caught up missed occurrence"`). Bounded lookback (≤366d) so a very old `lastRun` cannot spin. Skipped when `lastRun` is zero (never-run orders fire only on an exact match — no unbounded back-fill).

`cronFieldMatches` (including `*/N` step syntax) is unchanged — parsing was never the issue.

## Test
Adds `TestCheckTriggerCronCatchesUpMissedBoundary`: a `0 */4 * * *` order last run at the `00:00` boundary, evaluated at `04:01` after the `04:00` boundary passed. Fails before this change (`Due=false, "cron: schedule not matched"`), passes after. Full `internal/orders` suite is green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #2499, opened 2026-05-22
bug: doctor "order-firing-current" is permanently red — weekly+ cron schedules error out of interval computation


sjarmak:
## Summary

`gc doctor`'s `order-firing-current` check returns `StatusError` ("scheduled
orders are stale") permanently on any city that has a weekly-or-longer cron
order. The error is a false positive in the check's own interval math, and it
masks the check's ability to flag real order stalls.

## Symptom

`gc doctor` reports:

    ✗ order-firing-current — scheduled orders are stale

`gc doctor --verbose` shows the offending line:

    mayor-pattern-miner: cannot compute expected interval: cron schedule "30 8 * * 1" has no firing minutes in a 24h window

All other orders in the same output are current (last-fired well within their
intervals). The single un-computable order escalates the whole check to error.

## Reproduction

1. Add any enabled cron order whose schedule fires less often than daily, e.g.
   a weekly order: `schedule = "30 8 * * 1"` (Mondays 08:30).
2. Run `gc doctor` (or `gc doctor --verbose`).

Result: expected `order-firing-current` to evaluate the weekly order as
current (it fired within its weekly interval) → got `✗ ... scheduled orders
are stale`, permanently, regardless of actual firing health.

## Root cause

`internal/doctor/checks_order_firing.go:173` `computeExpectedIntervalForCronSchedule`
scans a fixed **1440-minute (24h) window** from a hard-coded base date:

- `internal/doctor/checks_order_firing.go:180` — `base := time.Date(2026, 5, 12, 0, 0, 0, 0, time.UTC)` (a Tuesday)
- it iterates `i := 0; i < 1440` minutes and collects matches
- `internal/doctor/checks_order_firing.go:194` — on zero matches it returns
  `fmt.Errorf("cron schedule %q has no firing minutes in a 24h window", schedule)`

A weekly schedule (`* * * * 1`) matches zero minutes in a single Tuesday, so it
errors. The caller at `checks_order_firing.go:102-108` maps that error to
`StatusError`, which becomes the worst status and renders the whole check red
with message "scheduled orders are stale" (`checks_order_firing.go:131`).

The same applies to monthly/yearly schedules — any cron coarser than daily.

## Fix candidates

**A. Widen the scan window to cover the longest practical period.** Scan ≥7
days for weekly, and fall back further (e.g. up to ~366 days) before declaring
zero matches; return the min inter-fire gap as today.
[...]

PR #5091, opened 2026-08-07
fix(orders): cron range syntax matched nothing, silently

wynged:
## Summary

A cron order whose schedule uses range syntax never fires, and nothing reports
why. `cronFieldMatches` in `internal/orders/triggers.go` accepts `*`, exact
integers, comma lists, and `*/N` — but not `A-B`. A field like `6-18` parses as
none of the supported forms and falls through to matching **no value at all**,
so `schedule = "*/30 6-18 * * *"` cannot fire on any minute of any day.

The failure is silent by construction. The matcher returns a bare bool, so "I
don't understand this field" and "this minute doesn't match" are the same
answer. There is no load error, no log line, and `gc order list` still shows the
order registered and enabled. The only symptom anywhere is `gc doctor`'s
`order-firing-current` check reporting staleness, with nothing explaining the
cause.

This PR adds range matching and — more importantly — makes any schedule the
matcher cannot express a **load-time error** instead of an order that quietly
never runs.

## Problem

Current `main`:

```go
// cronFieldMatches checks if a single cron field matches a value.
// Supports: "*" (any), exact integer, or comma-separated values.
func cronFieldMatches(field string, value int) bool {
	if field == "*" {
		return true
	}
	for _, part := range strings.Split(field, ",") {
		part = strings.TrimSpace(part)
		if strings.HasPrefix(part, "*/") { ... }
		n, err := strconv.Atoi(part)   // "6-18" -> error, discarded
		if err == nil && n == value {
			return true
		}
	}
	return false
}
```

`strconv.Atoi("6-18")` errors, the error is discarded, and the loop ends
`false`. `checkCron` then reports `cron: schedule not matched`, forever.
`Order.Validate` never inspects the schedule string beyond checking it is
non-empty, so nothing catches it earlier either.

The same shape hides a second class of dead schedule: a value that parses fine
but sits outside the range the evaluator ever tests. `matchesAt` compares
day-of-week against `time.Weekday`, which is `0`–`6`, so the common cron
spelling of Sunday as `7` also matches nothing, forever, silently. An hour of
`25` behaves the same way.

## Fix

Two parts.

**Matching.** `cronFieldMatches` now routes each comma term through a new
`cronTermMatches`, which handles `*`, `N`, `A-B`, `*/N`, and `A-B/N`. Stepped
ranges are anchored at the range start, so `8-20/2` matches 8, 10, … 20.

**Validation.** `ValidateCronSchedule` is called from `Order.Validate`. It
rejects any term `cronTermMatches` cannot express, plus any value outside the
field's real bounds — hour > 23, day-of-month 0, day-of-week 7, an inverted
range like `18-6`, a non-positive or non-numeric step. It is written as a
deliberate mirror of `cronTermMatches`, term for term, with a test asserting
the two agree: a schedule that validation accepts but matching cannot satisfy
is exactly the bug this exists to prevent.

Two things left alone on purpose:

- **`*/N` keeps its stride anchored at 0**, not at the field's minimum, so
  `*/2` on day-of-month still matches 2, 4, 6 … rather than 1, 3, 5. That is
  the historic behavior; changing it would move the firing slots of every
  `*/N` schedule already in service, which a bug fix should not do quietly. It
  does mean `*/N` and `A-B/N` anchor differently, so the docs now say that
  explicitly.
- **Day-of-week `7` is rejected rather than normalized to `0`.** Accepting it
  as a Sunday alias is a defensible alternative and I will take that direction
  if you prefer it. I chose the error because it is the smaller change, and
  because either way it beats the status quo, where `7` loads fine and never
  fires.

## What this does NOT do

- No named months or weekdays (`JAN`, `MON`), no `@daily`-style macros, no
  `L`/`W`/`#` extensions, no seconds field. The accepted grammar is exactly
  what the evaluator supports, plus ranges.
- No change to catch-up, DST handling, timezone resolution, or the
  once-per-minute guard.
- No new dependency. This does not pull in a third-party cron library. If you
  would rather converge on one, that is a larger call than this fix and I would
  rather you make it than have me make it inside a bug fix.

**One behavior change worth flagging.** Schedules that previously loaded and
silently never fired now fail validation. In `internal/orderdiscovery`, that
means the order is either dropped with a reported error or discovery fails,
depending on the caller's `OnValidateError` handler. I think that is the right
trade — the affected orders were already not running — but it is louder than
before, so whether it belongs in a patch release is your call. The one cron
schedule shipped in this repo
(`examples/bd/dolt/orders/mol-dog-stale-db.toml`, `0 */4 * * *`) validates
cleanly.

## Testing

Gates I ran on this branch, with real results:

- `make lint` — **pass**, 0 issues.
- `make vet` — **pass**.
- `make fmt-check` — **pass**, no diff.
- `make check-docs` — **pass** (`ok test/docsync`); docs changed, so this one
  applies.
- `go test ./internal/orders/... ./internal/orderdiscovery/...` — **pass**.
- `make check` — **exit 2.** Its `fmt-check`, `lint`, `vet`,
  `check-release-dist-ignore`, and `check-routed-test-rows` stages all passed;
  `make test` failed in three packages, none of which this change touches. I
  checked every one against a clean `main` worktree on the same host:

  | Failure | Clean `main` |
  |---|---|
  | `cmd/gc` `TestDetectStandaloneBdDoltLiveDoltForDifferentDataDirDoesNotConflict`, `TestStartBeadsLifecycleRefusesLiveStandaloneBdDolt`, `TestInitDirIfReadyDetectsStandaloneBdDoltAtProviderConvergence` — each 30s timeout, `fake dolt sql-server did not become inspectable` | Fails identically. Ambient dolt collision, tracked in #4677 / #4673. |
  | `examples/bd/dolt` `TestRunBoundedPython3Fallback*` — exit 127, `/usr/bin/env: 'bash': No such file or directory` | Fails identically. Host shell layout, not the patch. |
  | `internal/productmetrics` `TestDisableAndPurgeExactTokenConflictAndPeerCleanRecovery` | Passes in isolation on clean `main` **and** on this branch. Only failed inside the parallel `-p=4` sweep, so it looks load-dependent on this host. |

  So `make check` is red here, but not for anything in this diff. I would still
  treat CI on this PR as the authoritative read rather than my host.

Tests added in `internal/orders/triggers_test.go`:

- `TestCronFieldMatches` gains range cases (`6-18`, `8-20/2`, `1-3,10-12`) and
  negative cases for an inverted range, a step without a range, a non-positive
  step, and garbage.
- `TestCheckCronFiresRangeSchedule` — end-to-end regression through
  `CheckTrigger` for `*/30 6-18 * * *`: fires at 16:30, not at 16:15, not at
  03:30. It uses a warm `lastRun` rather than `neverRan` for the same reason
  `TestCheckTriggerCronNotMatched` does — the never-run bounded catch-up window
  from #3947 is a separate path that would legitimately report Due on an
  off-schedule minute and would confound the assertion.
- `TestValidateCronSchedule` — accepts the valid grammar; rejects a wrong field
  count, out-of-bounds values in each of the five fields, `7` for Sunday, an
  inverted range, a bare `N/step`, `*/0`, `*/x`, and garbage.
- `TestValidateCronScheduleAgreesWithMatcher` — for each accepted schedule,
  sweeps a full day a minute at a time and asserts it actually matches
  somewhere. This is the drift guard between the two functions.

**On `make test-integration`:** not run. The unit sweep already shows this host
cannot run the store-lifecycle suites cleanly (see the table above), so an
integration result from it would not be worth reporting either way. Trigger
evaluation is controller behavior, so by the template's rule the suite does
apply — I would rather say plainly that I did not get a trustworthy run than
tick the box.

**Environment note, only so the results above are interpretable:** the Makefile
points CGO at Homebrew `icu4c` on Darwin and otherwise expects system ICU
headers. This host has linuxbrew `icu4c` but no `libicu-dev`, so I supplied
`CGO_CPPFLAGS`/`CGO_CXXFLAGS`/`CGO_LDFLAGS` by hand for every gate above. That
is a local quirk, not something this patch touches.

## Related

Searched open and closed issues and PRs for cron, range, and
schedule-never-fires; nothing describes this. Adjacent but distinct:

- #3947 and #2721 — missed *occurrences* for schedules that already parse
  (catch-up). Same trigger, different failure; this PR interacts with them only
  in that the new test has to steer around the never-run catch-up path.
- #2499 — the doctor's interval computation for weekly schedules. Neither
  touches field grammar.

## Checklist

- [x] Added tests for the behavior change.
- [x] Updated docs for the user-facing change:
  - `docs/tutorials/07-orders.md` — the `cron` trigger bullet enumerated the
    supported field syntax. It now lists ranges and stepped ranges, states the
    two step anchors, notes day-of-week is `0`–`6`, and says an unsupported
    schedule is rejected at load.
  - `engdocs/architecture/orders.md` — the "Cron granularity is minutes" note
    claimed the trigger supports neither ranges nor steps. The steps half was
    already stale before this PR; both halves are now accurate, and
    `ValidateCronSchedule` is added to the function and validation-rule tables.
- [x] Called out the behavior change — see the flag above.
- [ ] No linked issue. I did not find one, and the description above is
      self-contained, so I did not open one separately. Happy to file one first
      if you would rather have it tracked that way.


quad341:
@wynged — this is a model bug report and fix. What makes it is the diagnosis, not the patch: you saw that the bug was invisible because a bare bool can't distinguish "I don't understand this field" from "this minute doesn't match," and you closed *that* hole at load time. Adding `A-B` alone would have left the next unsupported spelling to cost someone another week.

I verified rather than took on faith. Your three touched packages pass here, and `TestValidateCronScheduleAgreesWithMatcher` holds far past the seven schedules it asserts. CI is red on three checks — all one `sum.golang.org` / `proxy.golang.org` outage that broke the pinned `bd` install; I read the logs, nothing to do with your change. Re-running and merging once it clears.

Two notes, both mine to handle. Your body still describes the earlier revision where day-of-week `7` was rejected — the diff accepts it, which is the better call; I'll squash from the diff. And I'll take a small engdocs correction: the `*/N` anchor still differs between `internal/orders` and `internal/doctor` (harmless — same interval either way), and the bullet reads as if they're in lockstep.

Leaving the third-party-cron-library decision to us was exactly right.


---

wynged:
Closing: #5711 landed the same fix on 2026-09-14 (shared cron parser, schedule validation at load), which covers the range syntax and the silent-mismatch hole this PR addressed. Thanks for the review.

Issue #4499, opened 2026-07-21
bug(beads): caching_store reconcile amplifies under load: degraded state re-fires every 5s, recovery is O(N) serial bd execs, governor blind to both


sjarmak:
## What we could not verify

Stating this up front so the claim strength is clear:

- **We did not observe the degraded-state path firing** during our incident. Reconcile frequency (about 119 passes per 15 min across 8 rigs) matches the normal 60s latency cadence, not the 5s degraded re-fire. The degraded path described below is a code-read finding, not an observed one.
- **We did not instrument `recoverMissingFromList`.** It is the most plausible source of the multi-minute pass durations we did observe, but we have no per-bead exec count proving it.
- **We are not claiming this caused our host saturation.** The same box was running 62 concurrent agent sessions on 16 cores, and we could attribute only a fraction of the measured 208 forks/sec to any single subsystem. This issue is about the amplification mechanism, not an outage post-mortem.

What is verified: the four code properties under "Root cause", read against `origin/main` @ `117494543`, and the pass durations under "Symptom".

## Summary

`CachingStore`'s reconcile loop has three properties that make it cost more precisely when the backing store is slow, plus a fourth that removes the only cross-rig spreading mechanism. We observed reconcile passes taking 2m59s to 3m51s while reporting `adds=0 updates=0 removes=0`, and the same rig at the same bead count going from 626ms to 2m59s.

## Symptom

From a 22-rig city's supervisor log. The `adds`, `updates`, and `removes` counters are all zero, so these passes did no work:

```
beads cache: reconciled rig=gc              beads=227 adds=0 updates=0 removes=0 took=626ms      cadence=latency
beads cache: reconciled rig=gc              beads=227 adds=0 updates=0 removes=0 took=2m59.335s  cadence=latency
beads cache: reconciled rig=enterprisebench beads=686 adds=0 updates=0 removes=0 took=3m51.164s  cadence=latency
beads cache: reconciled rig=gpk             beads=125 adds=0 updates=0 removes=0 took=3m20.263s  cadence=latency
```

Across the 85-minute log window, 391 of 418 reconcile passes (93.5%) were pure no-ops: zero adds, zero updates, zero removes.

Adjacent in the same log, offered as context rather than as a claim about this subsystem:

```
tmux state cache: refresh failed in 5m18.705316841s: tmux -u: signal: killed
```

[...]

Issue #3253, opened 2026-06-08
perf: extend hydration-free / bounded reads to formulas/feed and beads(all=true) — both still O(full history) per cold cache bucket


sjarmak:
## Summary

Two supervisor read endpoints that back the dashboard `/runs` view rebuild an O(full-history) body on every cold cache bucket:

- `GET /v0/city/{city}/formulas/feed?scope_kind=city&scope_ref={city}` — ~10s under load (measured 7–18s).
- `GET /v0/city/{city}/beads?limit=500&type=molecule&all=true` — ~7s (measured 5–13s). It scans the full molecule history (~340k rows on a busy city) only to surface a bounded set of recent run roots.

These are the same class of bottleneck #3186 / #1896 fixed for `GET /v0/city/{city}/status` (hydration-free parallel counts + a `lite` body trim + a short time-bucket cache; was 5–40s, now ~1s / cached). That work set the precedent but was scoped to `/status`.

#3208 since added a 2-second time-bucket **response cache** to feed and `beads?all=true`, which amortizes the rebuild across pollers within a 2s window. It does **not** change the underlying cost: every cold bucket still pays the full scan, there is no single-flight, and concurrent cold misses each rebuild independently. So interactive callers with short budgets and the first poll after any 2s gap still see the full 7–18s / 5–13s latency. The real fix is to make the *build* cheap (hydration-free + recency-bounded), the way `/status` was, rather than only caching an expensive build.

## Root cause (file:line on origin/main)

**Feed** — `internal/api/huma_handlers_formulas.go` calls `buildWorkflowRunProjectionsRootOnly` (`internal/api/orders_feed.go`), which per store runs:

- a full active scan: `listActiveWorkflowProjectionBeads` → `store.List(ListQuery{AllowScan: true})` (no filter, no limit) — `internal/api/orders_feed.go:285-291`;
- a full closed-history roots scan: `store.List(ListQuery{Metadata: {gc.kind: workflow, gc.formula_contract: graph.v2}, IncludeClosed: true})` with **no `Limit` and no `Sort`** — `internal/api/orders_feed.go:223-229`.

The result is sorted and only then truncated to the feed limit, so the cost is paid over the entire history regardless of how few rows the caller wants.

[...]

Issue #2152, opened 2026-05-15
fix(beads): add CachingStore Update/Close no-op filters (Phase 1 of #1978)


sjarmak:
Phase-1 fix for #1978. Smallest-blast-radius win, ship-next.

## Background

Investigation report at `.gc/pr-pipeline/blast-radius/bd-write-architecture-1978.md` (PL bead `gc-39tv`) identified the 25k connections / 90s storm source: the reconciler tick in `cmd/gc/session_reconciler.go` (12+ `SetMetadataBatch` sites — heartbeats, sleep_intent, deferral timers) re-writes values that haven't changed. `CachingStore` already has a `metadataAlreadyMatchesCached` no-op filter for `SetMetadata`, but `Update` and `Close` do not — and the post-mutation `Get(id)` refresh is always called, even on the short-circuited path.

## Fix

1. Add `updateAlreadyMatchesCached(id, UpdateOpts) bool` modeled on existing `metadataAlreadyMatchesCached`.
2. Short-circuit `CachingStore.Update` when matched AND cache is `fresh` (use existing `lastFreshAt` / `state` plumbing).
3. Short-circuit `CachingStore.Close` when cached status is already `"closed"`.
4. Skip the post-mutation `c.backing.Get(id)` refresh on the short-circuited path.
5. Tests for each method: (a) value-match → no shell-out, (b) cache-stale → falls through, (c) value-different → shell-out.

**No bd-binary or protocol changes.** Localized to `internal/beads/caching_store_writes.go`.

## Expected size

~50-100 LOC source + ~50 LOC tests.

## Scope notes

Phase 1 alone does **not** resolve cross-agent traffic — surface this in the PR body so #1978 stays open after Phase 1 lands. Phase 2 (separate issue) handles cross-agent cache convergence.

## Out of scope

- Long-lived bd daemon (deferred unless Phase 1 + Phase 2 telemetry shows residual storm)
- In-process bead client (track upstream bd; don't block)
- Write batching (reconsider only if 1 + 2 insufficient)

## Source

Investigation polecat (PL bead `gc-39tv` → wisp `gc-ydnbpg`); report `/home/ds/gas-city/.gc/pr-pipeline/blast-radius/bd-write-architecture-1978.md`.

elmpp:
This has made a significant difference 💯 

Logs are more legible (constant "nothing to commit" errors gone), CPU 150% -> 20% 

Thank you very much

Issue #1978, opened 2026-05-11
gc agents shell out to bd per-write — ~25k short-lived dolt connections in a 90-second window


vbtcl:
## Summary

Every gc agent (mayor, deacon, boot, witness, polecats, dogs, dispatchers, …) writes to bd by spawning a fresh `bd` CLI process. Each invocation opens a new dolt sql-server connection, runs an upsert that often doesn't change any rows, and closes. Under the normal patrol/wisp cadence this produces a sustained connection-thrash workload.

## Evidence

`.gc/runtime/packs/dolt/dolt.log` between 21:46:38 and 21:48:07 PDT on 2026-05-10 (the 90 seconds before the journal corruption recorded in #1920 comment):

- **~25,000 short-lived connections** to `beads_hq`, almost all logging:
  ```
  level=warning msg="error running query" connectionDb=beads_hq error="nothing to commit"
  ```
- "nothing to commit" = the upsert was a no-op against current row state. The work to open the connection, parse the SQL, attempt the commit, and close the socket happened anyway.
- Connection IDs climbed from low single digits at 21:32 to 26,294 by 21:48 — that's ~265 conn/sec average in that window.

## Why it matters

- Amplifies blast radius of any dolt restart. Long-lived connections would queue gracefully on respawn; tens of thousands of TCP teardowns mid-fsync don't.
- Wastes dolt CPU on connection-handshake/auth work that dwarfs the actual mutation.
- Almost certainly contributes to the slow-query window that trips the bd circuit breaker and triggers the kill described in #1920.

## Fix directions (pick one)

1. **Long-lived per-agent bd daemon** — agents connect to a Unix socket; the daemon multiplexes onto a small pool of dolt connections.
2. **Write batching/coalescing** — agents queue writes and a debounced flusher commits them in batches. Reduces no-op upserts dramatically.
3. **In-process bead client** — link bd as a library where the agent's runtime allows. Avoids the fork+exec+TCP roundtrip per write entirely.

(3) is the highest-leverage but most invasive. (1) and (2) are achievable with shell-side changes.

## Related
- #1920 (restart flap — this storm trips the circuit breaker that triggers the kill)
- #855 (reconciler stuck in drain-log loop, sustained 65% dolt CPU — likely same connection-thrash class)

vbtcl:
Confirming this pattern from a separate gas-city setup — and adding a data point that the storm is **sustained, not just burst**.

## Setup

[...]

PR #5092, opened 2026-08-07
perf(beads): skip post-write backing Get on cached Update/Close/Reopen

wynged:
## Summary

`CachingStore.Update`, `Close`, and `Reopen` each re-read the bead they just wrote via `c.backing.Get(id)`, to adopt the authoritative post-write state. When the bead is already in the cache that read is redundant — the store has the prior bead and the transition it just applied, so it can derive the new state in memory. For a `BdStore` backing each of those reads is a `bd` subprocess plus a database round trip.

This adds a cache-hit fast path to each of the three verbs. On a cache miss nothing changes.

This is an optimization, not a correctness fix, and it makes one deliberate tradeoff around revision numbers. That tradeoff is spelled out under **Revision accounting** below — please read that section before the diff, it is the part most worth arguing about.

## Problem

All three verbs have the same shape today (`internal/beads/caching_store_writes.go`):

```go
if err := c.backing.Update(id, opts); err != nil {
    return err
}
// Re-fetch from backing to get the authoritative state.
fresh, err := c.backing.Get(id)
```

The write already succeeded and the transition is known. The `Get` exists to pick up whatever else the backing decided — most importantly the post-write revision. But on a cache hit the store is re-reading a bead it already has, to learn a change it just made.

The project has already established that this refresh is worth avoiding. #2152 (Phase 1 of #1978) added the `updateMatchesCached` / `closeAlreadyMatchesCached` idempotence filters, and item 4 of that issue was explicitly *"Skip the post-mutation `c.backing.Get(id)` refresh on the short-circuited path."* That covered writes that turn out to be no-ops. This change extends the same idea to writes that really do mutate a bead that is already cached.

## Fix

Each verb gets a cache-hit branch ahead of the existing backing read:

- **`Update`** applies `opts` to the cached bead with the existing `applyUpdateOptsToBead`, then absorbs it through `absorbFreshLocked(absorbOpts{depsMode: depsFromFields, seqMode: seqKeep, clearDirty: true})` — the same options the cache-miss success path uses.
- **`Close` / `Reopen`** flip `Status` on the cached bead and absorb with `depsKeepCached`, matching their own fall-through branches.

All three keep the surrounding side effects: `noteLocalMutationLocked`, `clearDependentReadyProjectionsLocked` on a status change, `markFreshLocked`, `updateStatsLocked`, and the same `notifyChange` event name the bypassed path emits (`bead.updated` / `bead.closed` / `bead.updated`).

Deliberately left alone:

- The `updateMatchesCached` / `closeAlreadyMatchesCached` idempotence short-circuits still run **ahead** of these fast paths and are untouched.
- `Create`, `CloseAll`, `SetMetadata*`, `DepAdd/DepRemove`, `Delete*`, `ReleaseIfCurrent`, and the `Tx` paths are unchanged.
- The cache-miss path in all three verbs is unchanged, so a bead that is not cached is still admitted via a backing read.

### Revision accounting — the tradeoff

The fast paths advance `Revision` **locally** (`+1`) instead of adopting it from a backing read. That is required: `beadstest.RunConditionalWriterConformance`'s `every_mutation_bumps_revision` case asserts that the store's own `Get` shows a strictly greater revision after every mutation, and `CachingStore` is run through that suite (`TestCachingStoreConditionalWriterConformance`).

**The limitation: a local `+1` can under-report when the backing advanced the revision by more than one** — for example when another writer also mutated the bead between this store's write and its next read. The cached revision is then behind the backing's.

I think this is acceptable, and here is the argument — but it is the thing to push back on if you disagree:

- It is self-healing, not silently wrong. A later conditional write fenced on the lagged revision precondition-fails, and `applyConditionalWriteFailure` → `evictForConditionalWrite` drops the entry, forcing a re-read of the real state. `TestCachingStoreConditionalWriteEvictsOnLaggedRefresh` and `TestCachingStoreConditionalPreconditionEvictsPerVerb` already cover that recovery and still pass.
- Consumers that fence on a revision must already handle a precondition failure and retry, so the failure mode is one they are built for.
- The pre-existing comment in `Close` makes the same argument in the other direction — it notes a lagged *revision* is self-healing while a lagged *status* is not, which is why `Close` force-writes the status. The fast paths preserve that asymmetry: status is always correct, revision may lag by a bounded amount.

What this costs is extra precondition failures under genuine multi-writer contention on the same bead, in exchange for removing a backing read from every cache-hit mutation. If you would rather have exact revisions, the alternative is to keep the `Get` and take the round trip, and this PR should be closed.

### One behavior change worth naming

`TestCachingStoreUpdateClearsCachedDependenciesFromFreshBead` is removed. It asserted that a **backing-side** dependency edit — one not present in `opts` — reaches the cache via `Update`'s refresh read. A cache-hit `Update` now derives deps from the locally applied bead (`depsFromFields`), so that class of edit propagates via the periodic reconciler instead of by riding along on the next unrelated `Update`. I removed the test rather than weaken it, with a comment in place naming what replaced the coverage.

Two other tests are removed for a narrower reason: their premise was a **failed** cache-hit refresh, and a cache hit no longer refreshes, so the state they asserted is unreachable from `Update`. Both left comments pointing at the coverage that remains (`TestCachingStoreCachedListUnavailableAfterWriteThroughRefreshFailure`, `TestCachingStoreReconciliationClearsDirtyWriteThroughProjection`).

## What this does NOT do

- Does not change any store other than `CachingStore` — no `bd`, protocol, or schema changes.
- Does not touch the storage-binding stack from #5069/#5070/#5071. I checked: none of those three commits modify `caching_store.go` or `caching_store_writes.go`, and the new SQLite store and storage-class conformance suites pass unchanged.
- Does not reduce cross-agent write traffic. #1978's storm is many processes each writing; this only removes the *read-back* that each write performs, in one process. #1978 and #2152 are closed; I am citing them as lineage, not claiming to close anything.
- Does not address #4499 (reconcile amplification under load) — that is about the reconciler's recovery path, which still issues its own backing reads.
- Does not change `CloseAll`, which still does a backing `Get` per id in its loop. Same optimization probably applies; I left it out to keep this reviewable.

## Measurement

I did not benchmark wall-clock latency — doing that honestly needs a `bd`/dolt-backed store under a realistic workload, and I would rather not report a number I cannot stand behind. The improvement is structural: **one fewer backing round trip per cache-hit mutation.**

That count is measured, via the new tests and the existing `countingGetStore` helper. Running the new tests against the unmodified write paths on `main`:

```
--- FAIL: TestCachingStoreUpdateOnCacheHitDoesNotCallBackingGet
    backing.Get called 5 times on cache-hit Update, want 0     (5 Updates)
--- FAIL: TestCachingStoreCloseOnCacheHitDoesNotCallBackingGet
    backing.Get called 1 times on cache-hit Close, want 0
--- FAIL: TestCachingStoreReopenOnCacheHitDoesNotCallBackingGet
    backing.Get called 1 times on cache-hit Reopen, want 0
```

With the change, all three report 0.

## Testing

- [x] `make lint` — 0 issues
- [x] `make vet` — clean
- [x] `go test ./internal/beads/...` — **all packages pass** (`internal/beads`, `beadstest`, `contract`, `exec`), including `TestCachingStoreConditionalWriterConformance` and every subtest of it: `every_mutation_bumps_revision`, `revision_monotonic_never_reused`, `stale_revision_is_precondition_failed`, `conditional_success_paths`, `contention`, and the CAS cases. The SQLite and storage-class conformance suites added by #5069/#5070 also pass.
- [ ] `make check` — **did not come back green on my machine, but not because of this change.** Details below; I would like CI's opinion on this one.
- [ ] `make check-docs` — not run; no docs, navigation, or links changed.
- [ ] `make test-integration` — not run. This change is confined to an in-memory caching layer with no new I/O, and the unit suite covers the store contract, but it is fair to say a caching store is runtime behavior — I would not object to CI running it.

### On `make check`

`make check` failed at its `test` target on four packages, none of which are `internal/beads` and none of which import the changed code:

- `examples/bd/dolt` — fails **identically on a clean `upstream/main` checkout**, which I verified by running it there:
  ```
  --- FAIL: TestRunBoundedPython3FallbackSendsSigtermBeforeKill
      runtime_bounded_test.go:106: run_bounded exit code = 127, want 124 (timeout)
      output:
      /usr/bin/env: 'bash': No such file or directory
  ```
  That is my host lacking `bash` on the sanitized `env -i` PATH the test harness builds, not a regression.
- `internal/runtime/acp`, `internal/runtime/subprocess`, `internal/sdnotify` — these **pass when re-run**, both on a clean `upstream/main` checkout and on this branch. They failed only in the `-p=4` parallel run on a machine that was heavily loaded at the time; the failures are handshake/signal timing, not assertions about store behavior.

So the honest summary is: every gate that exercises the changed code passes, and the `make check` failures reproduce without the change. I have not been able to get a fully green `make check` on this host even on `main`.

New tests:

- `TestCachingStoreUpdateOnCacheHitDoesNotCallBackingGet`
- `TestCachingStoreCloseOnCacheHitDoesNotCallBackingGet`
- `TestCachingStoreReopenOnCacheHitDoesNotCallBackingGet`

Reframed onto the cache-miss path (`Prime` skipped, since a cache hit no longer refreshes): the `Update` refresh-failure and `NotFound` tests — `...InvalidatesStaleCacheWhenRefreshFails`, `...LogsWhenRefreshFails`, `...RemovesCacheWhenRefreshReturnsNotFound`.

## Checklist

- [x] Linked issues — #1978 and #2152 as lineage; both already closed, so nothing here claims to close them.
- [x] Added tests for behavior changes
- [x] No user-facing surface changed, so no docs update
- [x] No breaking changes. The behavior change to backing-side dependency propagation on `Update` is called out above.

## Related

- #1978 / #2152 — the write-amplification lineage this extends. Closed; not duplicates.
- #4499 — caching-store reconcile amplification. Adjacent, different code path, not addressed here.
- #3253 — hydration-free/bounded reads for formulas and `beads(all=true)`. Read-side performance; unrelated to these write paths.

### DIFF
--- a/internal/beads/caching_store_internal_test.go
+++ b/internal/beads/caching_store_internal_test.go
@@ -1385,9 +1385,9 @@ func TestCachingStoreUpdateInvalidatesStaleCacheWhenRefreshFails(t *testing.T) {
 	}
 
 	cache := NewCachingStoreForTest(backing, nil)
-	if err := cache.Prime(context.Background()); err != nil {
-		t.Fatalf("Prime: %v", err)
-	}
+	// Deliberately NOT primed. Update's refresh-failure handling now
+	// lives on the cache-miss path only — a cache hit applies opts in-memory and
+	// never calls backing.Get, so priming here would bypass the branch under test.
 
 	title := "after"
 	backing.failNextGet = true
@@ -1428,9 +1428,9 @@ func TestCachingStoreUpdateRemovesCacheWhenRefreshReturnsNotFound(t *testing.T) 
 	cache := NewCachingStoreForTest(backing, func(eventType, beadID string, _ json.RawMessage) {
 		events = append(events, eventType+":"+beadID)
 	})
-	if err := cache.Prime(context.Background()); err != nil {
-		t.Fatalf("Prime: %v", err)
-	}
+	// Deliberately NOT primed. Update's refresh-failure handling now
+	// lives on the cache-miss path only — a cache hit applies opts in-memory and
+	// never calls backing.Get, so priming here would bypass the branch under test.
 
 	title := "after"
 	if err := cache.Update(bead.ID, UpdateOpts{Title: &title}); err != nil {
@@ -1447,8 +1447,10 @@ func TestCachingStoreUpdateRemovesCacheWhenRefreshReturnsNotFound(t *testing.T) 
 	if len(items) != 0 {
 		t.Fatalf("List after update/refresh NotFound = %#v, want no resurrected bead", items)
 	}
-	if len(events) != 1 || events[0] != "bead.closed:"+bead.ID {
-		t.Fatalf("events = %v, want [bead.closed:%s]", events, bead.ID)
+	// With no prior cached state there is no bead.closed to emit, but the
+	// tombstone must still prevent resurrection.
+	if len(events) != 0 {
+		t.Fatalf("events = %v, want none (cache-miss NotFound refresh emits no close)", events)
 	}
 	stats := cache.Stats()
 	if stats.ProblemCount != 0 {
@@ -1468,9 +1470,9 @@ func TestCachingStoreUpdateLogsRefreshFailure(t *testing.T) {
 	cache.problemf = func(msg string) {
 		logged = append(logged, msg)
 	}
-	if err := cache.Prime(context.Background()); err != nil {
-		t.Fatalf("Prime: %v", err)
-	}
+	// Deliberately NOT primed. Update's refresh-failure handling now
+	// lives on the cache-miss path only — a cache hit applies opts in-memory and
+	// never calls backing.Get, so priming here would bypass the branch under test.
 
 	title := "after"
 	backing.failNextGet = true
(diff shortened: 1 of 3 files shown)


wynged:
Rebased onto current main (it was 645 commits behind). One adaptation: the Close and Reopen cache-hit fast paths now go through setBeadStatus instead of assigning Status directly, so an indefinite deferral clears on the fast path exactly as it does on the refresh path. That is what TestCachingStoreReopenClearsStatusBasedDeferralWhenRefreshFails, added after this PR was opened, checks. internal/beads/... is green locally. Ready for review whenever it fits.

Issue #4572, opened 2026-07-22
fix(pool): stale start-pending session records must not consume pool capacity


tudorsaitoc:
## Incident

On 2026-07-22 a restarted controller inherited persisted `start-pending` session records.  The reconciler could not make progress on those records, while desired-state accounting treated them as occupied pool capacity.  Replacement workers were therefore not created and fleet throughput reached zero.

## Required behavior

- A pool slot is occupied only by a live or bounded in-flight session state.
- An expired never-started `start-pending` record cannot satisfy pool demand or be reused as capacity after restart.
- Unknown and terminal persisted states cannot consume pool capacity.
- With stale `start-pending` records present, pool desired state still emits replacement worker creation.

## Existing candidate

PR #3905 implements the lease-aware pending-create path and includes desired-state regressions.  It needs rebase/verification against current `main` and should be extended if its lease predicate does not cover every persisted unknown/terminal state.

## Acceptance regressions

1. Seed stale `start-pending` pool records, then assert replacement requests are produced.
2. Restart/reload persisted stale records and assert no record can reserve a pool slot indefinitely.
3. Assert unknown and terminal state records are excluded from capacity accounting.

Downstream incident: https://github.com/tudorsaitoc/saitoc/blob/main/docs/incidents/2026-07-22_fleet-throughput-zero-zombie-start-pending-sessions.md

Issue #4114, opened 2026-07-09
controller: unclaimable routed pool demand causes indefinite zero-claim spawn loop (211+ cycles/25h undetected)


webmastersdiy:
## What happened

A bead slung to a rig polecat pool (`gc.routed_to=<rig>/gastown.polecat`, unassigned) carried an application-level wait marker (`metadata halt_reason=waiting_b3`). Every spawned polecat read the marker, declined to claim, and drain-exited after ~1-2 min. The controller saw pool demand still standing and respawned: **~211 spawn cycles over ~25 hours at a 3-8 min cadence** (v1.3.2, macOS).

The loop only ended when the wait condition resolved (a human closed the blocking review bead) and the next spawn claimed the work.

## Why the controller couldn't see it

- `scale_check` counts routed+unassigned+ready beads as demand. It has no way to know an agent-side convention makes the bead unclaimable - fair.
- But the failure shape IS observable controller-side: **K consecutive spawns for the same pool, each exiting having claimed nothing**, while the same demand bead stays unassigned.
- Nothing detected or dampened it. spawn-storm detection did not trigger at this cadence (~3-8 min between cycles). Each cycle burned a session spawn, a session bead create+close, and a full agent boot's worth of tokens.

## Side damage

- Session bead churn: 211+ create/close pairs in the town ledger.
- Events buffer flooded: the earliest retained event advanced ~18h, losing the incident window's lifecycle events (which is also what made post-hoc diagnosis harder).

## Suggested direction

1. **Zero-claim backoff**: track consecutive spawns per pool that exit without claiming; after K (e.g. 3), back off exponentially (cap at, say, 1h) instead of immediate respawn.
2. **Surface it**: after K zero-claim cycles, emit an event / file a warrant-style bead naming the demand bead(s) that repeatedly went unclaimed, so an operator/mayor sees "demand exists but nothing will take it" instead of silent churn.
3. (Optional) A `gc doctor` check for pools whose demand age >> typical claim latency.

Town-side we now avoid the trigger (model waits as real `waits-for` dependencies, which drop the bead from `bd ready` and thus from demand, instead of metadata halt markers) - but any unclaimable-demand shape reproduces this, so the backoff seems worth having regardless.

Related-but-distinct: #3872 covers the inverse (demand blindness / not spawning); this is spawning that never stops.

webmastersdiy:
Second concrete mechanism for the same failure shape, observed ~30 minutes after filing this issue:

[...]

PR #3995, opened 2026-07-06
fix(controller/pool): four dispatch fixes — deferred-bead recovery, origin gate, rework-bead visibility, dormant cold-start

sarendipitee:
## Problem

Four independent controller/pool dispatch bugs, each of which left routed work beads permanently unclaimed (or pools unscaled) in production:

1. **Deferred pool-routed beads were unrecoverable (gcy-tw3).** Once any agent ran `bd update --defer` on a pool-routed bead, every orphan-recovery and session-cleanup query (filtered to `open`/`in_progress`) stopped seeing it — the bead was stranded forever.
2. **Pool-demand origin gate silently skipped named multi-session polecats (gcy-2w7).** Named pool polecats get `GC_SESSION_ORIGIN="named"` from the reconciler, and `poolDemandOriginGateScript()` exits early for any origin other than `ephemeral`, so their Tier 3 pool-demand probes never ran — pool-routed beads sat unclaimed by idle named polecats.
3. **Rework beads were invisible to the pool-demand probe (gcy-r4o).** Refinery-rejected beads keep `started_at` after being reset to open and re-routed; `bd ready` excludes beads with `started_at` set, so both the worker work-query and the reconciler scale-check never saw them — routed but unclaimable until a manual assignment.
4. **All-asleep pools never cold-started (gc-blo).** A min=0 pool whose only open sessions are ASLEEP counted those sessions as running, so `isCold` stayed false, the cold-wake demand probe never ran, and newly routed work sat unclaimed while polecats appeared dead.

## Fix (one commit per bug)

- `fix(controller): recover deferred pool-routed work beads` — adds `ClearDefer` to `UpdateOpts` (propagated through bdstore, native_dolt_store, memstore, and the bridge), collects deferred routed beads in `collectAssignedWorkBeadsWithStores`, and lets `releaseOrphanedPoolAssignments` release `status=deferred` beads back to open.
- `fix(config): skip pool-demand origin gate for multi-session agents` — `effectiveWorkQuery` omits the origin gate for agents where `SupportsMultipleSessions()` is true; singletons (mayor/witness/deacon) keep the gate so they never consume generic pool demand.
[...]

PR #5094, opened 2026-08-07
fix(pool): exclude deferred beads from pool demand

wynged:
## Summary

`filterAssignedWorkBeadsForPoolDemand` has no deferred-bead gate, so a work
bead that is deliberately parked with a future `defer_until` but still carries
a stale `gc.routed_to` is counted as standing pool demand. `bd ready` hides the
same bead, so `scale_check` reports zero for it. The two views disagree and the
controller spawns a worker every tick for work no worker can claim. This adds
the gate, with two regression tests.

## Problem

`collectAssignedWorkBeadsWithStores`' open pass draws from a raw
`List(status=open)`. That query still returns deferred beads, and
`appendOpenRoutedWorkUnique` keeps any that have an assignee plus a
`gc.routed_to`. Nothing downstream filters them out before
`ComputePoolDesiredStates` sees them, because the pool-demand filter only
resolves the route target and store reachability:

```go
filtered := make([]beads.Bead, 0, len(assignedWorkBeads))
for i, wb := range assignedWorkBeads {
	template := routedToOrLegacyWorkflowTarget(wb)
	...
}
```

`computePoolDesiredStates` turns that bead into a resume / wake-known-identity
request. The controller spawns an ephemeral session, the session finds no ready
work (the bead is deferred, so it is not in `bd ready`) and orphan-drains the
same tick with no wake reason, and the next tick repeats. Nothing self-corrects:
the loop ends only when the defer elapses or an operator strips the route by
hand.

The observable consequence is a steady, silent spawn/drain churn — two such
parked beads were seen driving roughly 1000 session spawns per day across two
pools, each one costing a session create/close pair and a full agent boot. The
demand snapshot looks healthy the whole time, because from the controller's side
there genuinely is a routed open bead.

This is one concrete instance of the divergence described in #4114: the demand
definition (routed + open at the storage layer) is broader than the claimable
definition (`bd ready` semantics — dependency-aware, defer-aware).

## Fix

Gate the loop with `beads.IsDeferred(wb, now)`, mirroring `bd ready`'s
server-side `defer_until IS NULL OR defer_until <= now` filter so the
controller's internal demand agrees with the shim.

Deliberately narrow:

- Only the `ForPoolDemand` path. `filterAssignedWorkBeadsForSessionWake` is
  untouched, so a concrete session that owns a deferred bead can still be woken
  by the existing wake rules.
- An *elapsed* `defer_until` (in the past) still counts as demand — the bead is
  ready again. Only a future `defer_until` parks it. There is a test pinning
  this, because the obvious over-broad version of this fix ("has a `DeferUntil`
  at all") would silently under-provision.
- The pool-filtered set also backs `reusablePoolSessionBead`'s assigned-work
  guard, so this has a second effect worth naming: a session whose only
  "assigned work" is a parked bead is now correctly reusable rather than pinned
  to it.

## What this does NOT do

- **No zero-claim backoff.** #4114 asks for consecutive-zero-claim detection and
  exponential backoff. This does not implement that. It removes one cause of
  unclaimable demand; other causes (agent-side convention markers, the
  dependency-blocked case in that issue's follow-up comment) still produce the
  same shape and would still benefit from the backoff.
- **Does not touch blocked beads.** The dependency-blocked variant is #3758's
  subject, not this one.
- **Does not address a redundant ephemeral spawned alongside a live
  resume-owner.** That is a separate, bounded and self-correcting defect that
  needs scale-check-versus-live-owner accounting, and it carries
  under-provisioning risk if done carelessly. Left as a follow-up rather than
  bundled here.
- **Does not add a recovery path for beads already stranded by a defer.** #3995
  proposed that (a `ClearDefer` path so orphan-recovery can see and release
  deferred routed beads) and was closed for being bundled. This is the demand
  side only; the two are complementary, not alternatives.

## Testing

Environment note up front: this host has a running Dolt sql-server and no
system ICU dev package (only a Homebrew `icu4c@78`), which affects several
gates. Where a gate failed I re-ran the same thing on a clean `upstream/main`
worktree at `a699601f0` under an identical environment to check whether it was
mine. Details below.

| Gate | Result |
|---|---|
| `make lint` | pass, 0 issues |
| `make vet` | pass |
| `make check` (`fmt-check`, `lint`, `vet`, `check-release-dist-ignore`, `check-routed-test-rows`) | pass |
| `make check` (`test` stage) | see below |
| `go test ./cmd/gc/... -short -run 'AssignedWork\|PoolDemand'` | pass, 39 tests |
| `make test-integration` | see below |

**Unit suite.** `make test` runs under `env -i`, which drops `LD_LIBRARY_PATH`;
on this host that makes every CGO-linked test binary fail to load
`libicui18n.so.78` before running a single test. That reproduces identically on
clean `upstream/main`, so it is a local environment gap, not this change. (For
what it's worth: the Makefile's Homebrew-ICU CGO block at line 41 is gated to
`ifeq ($(shell uname),Darwin)` on the assumption that Linux always has system
ICU. A Linux host with Homebrew ICU and no `libicu-dev` falls through it. Not
touching that here — separate concern, and arguably my environment's problem.)

Re-running the same suite with the ICU path restored gets a real result. The
only failures are:

- `cmd/gc`: `TestDetectStandaloneBdDoltLiveDoltForDifferentDataDirDoesNotConflict`,
  `TestStartBeadsLifecycleRefusesLiveStandaloneBdDolt`,
  `TestInitDirIfReadyDetectsStandaloneBdDoltAtProviderConvergence` (each times
  out at 30s against the ambient Dolt server), plus `TestImportMigrateScript`
  and `TestPackV2ImportsScript`. All five fail identically on clean
  `upstream/main`.
- `internal/testenv`: 5 tests, all failing identically on clean `upstream/main`.
- `examples/bd/dolt`: 3 `TestRunBoundedPython3Fallback*` tests, failing with
  `/usr/bin/env: 'bash': No such file or directory`. Identical on clean
  `upstream/main`.
- `internal/storebinding/sqlite`:
  `TestSQLiteWriterFenceSIGKILLAtReservationBoundaries` failed once under
  parallel load and passes in isolation on both this branch and clean
  `upstream/main`. Load flake.

Nothing in `cmd/gc` related to assigned-work scoping or pool demand fails.

**Integration.** The PR template asks for `make test-integration` when
controller behavior changes, which this does, so I tried. `make
test-integration` hits the same `env -i` ICU problem. Running the underlying
`go test -tags integration -timeout 30m ./...` directly with the ICU path
restored, the suite ran for over 20 minutes without a single package
completing — unsurprising given the ambient Dolt sql-server that already makes
the standalone-Dolt detection tests time out at 30s each in the unit run.

**So: I could not get a usable integration result on this host, and I am not
claiming one.** I would rather say that plainly than tick the box. The change
is a pure predicate added to one filter function — no I/O, no new state, no
new call sites — so CI's integration run should be the real signal here. If it
turns up something, I will fix it.

**Regression tests added** (`cmd/gc/assigned_work_scope_test.go`):

- `TestFilterAssignedWorkBeadsForPoolDemandDropsDeferredRoutedBead` — a bead
  with a future `defer_until` and a stale `gc.routed_to` is dropped, while a
  live in-progress bead routed to the same template is kept. Verified red
  against the unpatched filter (it keeps the deferred anchor).
- `TestFilterAssignedWorkBeadsForPoolDemandKeepsElapsedDeferRoutedBead` — a
  past `defer_until` still counts as demand. Passes both with and without the
  fix by design; it exists to catch a future over-broad defer check.

## Related

Adjacent, checked, and not duplicates:

- **#4114** (open) — same failure shape, broader scope. This closes one of the
  divergences that issue names; it does not implement the backoff it asks for.
  Referenced rather than closed.
- **#3758** (open) — `fix(runtime): ignore blocked assigned work for demand`.
  Same insertion point in the same function, different predicate (blocked vs
  deferred). Genuinely complementary: neither one covers the other's case. Heads
  up that the two will conflict textually at that line; whichever lands second
  is a one-line rebase, and I am happy to be the one that rebases.
- **#3995** (closed) — bundled four dispatch fixes, one of which was deferred-bead
  *recovery* (letting orphan-release see and clear a defer). Closed for mixed
  scope with an invitation to resubmit the slices separately. That is the
  recovery side; this is the demand side.
- **#4572** (open) — stale `start-pending` session records consuming pool
  capacity. Also a demand-accounting bug, but about session records, not work
  beads, and it causes under-provisioning rather than over-spawning.
- **#4448** (open) — drained pool sessions not reaped, their open session beads
  starving demand at cap. Again the session-bead side, and again the opposite
  direction (too few workers, not too many).

I did not find an existing issue specifically for the deferred case, so there is
nothing to close here. Happy to file one if you would rather have the bug
tracked separately from the fix.

## Checklist

- [x] Linked an issue — Refs #4114 (partial; does not close it)
- [x] Added tests for the behavior change
- [ ] Docs — no user-facing surface changed
- [x] No breaking changes or migration notes. One behavior note for reviewers:
      a pool that was being kept awake purely by a parked routed bead will now
      correctly scale to zero. That is the intended fix, but it is a visible
      change for anyone who had come to rely on the accidental floor.

### DIFF
--- a/cmd/gc/assigned_work_scope.go
+++ b/cmd/gc/assigned_work_scope.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"strings"
+	"time"
 
 	"github.com/gastownhall/gascity/internal/agentutil"
 	"github.com/gastownhall/gascity/internal/beads"
@@ -139,8 +140,22 @@ func filterAssignedWorkBeadsForPoolDemand(
 			assigneeToSessionBeadID[id] = sb.ID
 		}
 	}
+	now := time.Now().UTC()
 	filtered := make([]beads.Bead, 0, len(assignedWorkBeads))
 	for i, wb := range assignedWorkBeads {
+		// A deferred bead is deliberately parked (future defer_until) and is
+		// invisible to bd ready, so scale_check reports zero demand for it.
+		// But this pool-demand pass draws from a raw List(status=open) that
+		// still returns deferred beads. A deferred bead that retains a stale
+		// gc.routed_to would otherwise count as poolDesired=1 with no matching
+		// ready work, driving the reconciler to spawn an ephemeral session that
+		// immediately orphan-drains — a spawn/drain loop that only ends when
+		// the bead's defer elapses or the route is stripped by hand. Excluding
+		// deferred beads here mirrors bd ready's server-side filter so gc's
+		// internal demand agrees with the shim.
+		if beads.IsDeferred(wb, now) {
+			continue
+		}
 		template := routedToOrLegacyWorkflowTarget(wb)
 		if template == "" {
 			if sessionBeadID := assigneeToSessionBeadID[strings.TrimSpace(wb.Assignee)]; sessionBeadID != "" {
--- a/cmd/gc/assigned_work_scope_test.go
+++ b/cmd/gc/assigned_work_scope_test.go
@@ -3,6 +3,7 @@ package main
 import (
 	"path/filepath"
 	"testing"
+	"time"
 
 	"github.com/gastownhall/gascity/internal/agentutil"
 	"github.com/gastownhall/gascity/internal/beads"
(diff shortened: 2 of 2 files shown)


quad341 (approved):
Thanks @wynged — merging this as-is.

Three things stood out. You traced the actual divergence (the demand pass drawing from a raw `List(status=open)` while `bd ready` applies the defer filter server-side) instead of patching the symptom. You reused `beads.IsDeferred`, the same predicate `IsReadyCandidateForTier` consumes — I checked, and that means the two views can't quietly drift apart later. And you named the second-order effect on `reusablePoolSessionBead`'s assigned-work guard yourself; finding that in the PR body rather than having to discover it is the difference between a five-minute review and an hour.

The `KeepsElapsedDefer` test is the detail I appreciated most: a test that passes either way, kept deliberately to catch a future over-broad `DeferUntil != nil` rewrite, with a comment saying exactly that.

Your integration write-up was the right call too. "I could not get a usable result on this host and I am not claiming one" beats a ticked box — and you re-ran the failures on clean `main` to prove they were environmental. CI's integration shards ran green.

On #3758: thanks for the heads-up. If yours lands first, no rebase needed on your side.


---

PR #3582, opened 2026-06-18
fix(session): honor configured named workdir

duncan4123:
## Summary
- keep task-specific work_dir metadata as the highest-priority launch override
- stop stale session metadata work_dir from overriding resolved configured work_dir
- add regression tests for configured workdir precedence and task workdir precedence

## Test
- go test ./cmd/gc -run 'TestBuildPreparedStart(PrefersConfiguredWorkDirOverSessionMetadata|TaskWorkDirOverridesConfiguredWorkDir)|TestPrepareStartCandidate_Uses(SessionIDForTaskWorkDir|AssignedWorkSnapshotForTaskWorkDir)' -count=1

Issue #1181, opened 2026-04-24
Implicit pool agents share one working tree — parallel dispatch can clobber uncommitted edits


maphew:
## Summary

When `gc` dispatches more than one agent in parallel against the same rig, all of them cwd into a single shared working tree (the rig's `path`). If those agents do anything that touches git state — branch switches, resets, stashing — they can silently clobber each other's uncommitted edits. We hit this in practice and recovered only because the agents themselves noticed; future runs may not.

Internal tracker: `gc-7n8`.

## Repro / observed behavior

Two agents were dispatched in parallel to work on separate beads in the same rig (`/var/home/matt/dev/gascity`). Each created its own branch (`fix/gc-wqu-session-bead-dedup`, `fix/gc-rcy-actor-attribution`), but because they shared one working tree, mid-task branch switches from one agent wiped the other's uncommitted edits at least once. One agent had to re-apply, the other had to cherry-pick its own files out of the other's stash.

Both agents independently flagged this in their session reports without being prompted:

> "HQ coordination agents should use separate git worktrees, not swap branches in place, otherwise parallel work is guaranteed to interleave unsafely."

> "Mid-task, they branch-switched and reset which wiped my uncommitted edits once; I had to re-apply."

## Root cause

Two layers contribute, but the actual gap is in core:

1. `agentCommandDir` (`cmd/gc/cmd_start.go:816`) and `slingDirForBead` (`cmd/gc/cmd_sling.go:424`) both resolve every agent dispatched to a rig to the same `rig.Path`. Single shared cwd, no per-agent split.
2. `InjectImplicitAgents` (`internal/config/config.go:1475`) auto-creates one pool `claude` agent per rig (per configured provider) with `min=0, max=-1`. These implicit pool agents have **no `WorkDir`, no `pre_start`, and no worktree handling** — they cwd straight into the rig root.

The example `gastown` pack handles this for its own roles: `polecat` (`work_dir = .gc/worktrees/{{.Rig}}/polecats/{{.AgentBase}}`) and `refinery` both wire `examples/gastown/packs/gastown/scripts/worktree-setup.sh` via `pre_start`, and `test/acceptance/worktree_test.go` covers it. So the infrastructure exists at the example-pack layer, but cities that don't adopt the gastown pack — including any city that just relies on the implicit claude pool — get no isolation.

Result: any `gc dispatch` / `gc sling` that fans out >1 worker to the same rig via the implicit pool is silently unsafe today.

[...]

Issue #774, opened 2026-04-17
bug: {{.AgentBase}} resolves to template name, not namepool slot, causing pool instances to collide on one worktree path


wynged:
### Before you continue

- [x] I searched existing issues and did not find a duplicate.
- [x] I read the relevant docs and contributor guidance.

### Gas City version

0.15.0-rc1 (also present on main @ 73f09ddd).

### Environment

Linux 6.6 (WSL2 Ubuntu), zsh, tmux provider.

### Reproduction

1. Define a pool agent with a namepool and a work_dir template referencing `{{.AgentBase}}`:

   ```toml
   [[agent]]
   name = "ant"
   namepool = "ants.txt"    # produces slots like "ant-fenrir", "ant-grendel"
   work_dir = ".gc/worktrees/{{.Rig}}/ants/{{.AgentBase}}"
   min_active_sessions = 0
   max_active_sessions = 4
   ```

2. Trigger two pool members via `gc session new` or via a formula that spawns pool work.

3. Inspect the materialized `work_dir` for each pool member (session bead metadata or `git worktree list`).

### Expected behavior

Each pool instance resolves `{{.AgentBase}}` to its **slot identifier** (`ant-fenrir`, `ant-grendel`, …), producing distinct worktree paths:

```
.gc/worktrees/<rig>/ants/ant-fenrir
.gc/worktrees/<rig>/ants/ant-grendel
```

### Actual behavior

`{{.AgentBase}}` resolves to the **template base name** (`ant`), so every slot shares the same path:

```
.gc/worktrees/<rig>/ants/ant
.gc/worktrees/<rig>/ants/ant   ← collision
```

The resolver is `PathContextForQualifiedName` in `internal/workdir/workdir.go` — it parses the qualified name and extracts the template base rather than the pool slot. Passing `--alias` at `gc session new` time doesn't rewrite the work_dir metadata.

### Additional context

This was explicitly flagged as a **secondary** issue in the now-closed #758 ("controller's session reconciler doesn't know how to claim and start pool-managed sessions…"), but #758 only addressed the primary reconciler gap (fix in commit 561881ca / PR #480). The work_dir collision is still present in both v0.15.0-rc1 and main.

In our city this surfaces as ants on a rig stepping on each other's worktrees; the city-local workaround is to hand-roll `work_dir = ".gc/worktrees/{{.Rig}}/ants/{{.AgentBase}}"` and accept that only one ant can run at a time per rig until this is fixed.

[...]

PR #5095, opened 2026-08-07
fix(reconciler): resolve named-session work_dir from its own alias

wynged:
## Summary

Session-bead rediscovery resolves a named-session bead's identity to the shared
backing template's qualified name instead of the session's own stored identity.
Because `work_dir` and worktree-setup `pre_start` hooks interpolate
`{{.AgentBase}}`, every `on_demand` named session backed by a given template
resolves its working directory to one shared path keyed by the template, rather
than one path per session. Where that path is a git worktree, distinct agents
end up sharing one working tree, one branch, and one index.

This changes `canonicalSessionIdentityWithConfig` and its `session.Info` twin to
return the bead's stored `configured_named_identity` when it has one.

## Problem

In `cmd/gc/build_desired_state.go`, both forms of the canonical-identity helper
short-circuit named beads to the template's qualified name:

```go
if isNamedSessionInfo(info) {
    return cfgAgent, cfgAgent.QualifiedName()
}
```

`ParseQualifiedName` then binds `AgentBase` to the *template's* base name. With a
config shaped like the one the example pack teaches —

```toml
[[agent]]
name = "worker"
work_dir = ".gc/worktrees/{{.Rig}}/{{.AgentBase}}"
pre_start = ["worktree-setup.sh ..."]
```

— several `[[named_session]]` entries backed by `worker` all resolve to
`.gc/worktrees/<rig>/worker` instead of `.gc/worktrees/<rig>/<session name>`.

Where I am confident, and where I am not: the named-session loop emits sessions
with the correct `spec.Identity`, and its mode gate
(`spec.Mode != "always" && !hasCanonical && !namedWorkReady[identity]`) means an
`on_demand` session can fall through it and have its identity recomputed at the
rediscovery seam instead — that is the case I have actually observed. I have
*not* convinced myself that `always` sessions can never reach the same seam:
`discoverSessionBeadsWithRoots` writes `desired[sn] = tp` unconditionally, so if
an already-desired named session is re-keyed there it would be overwritten too.
Either way the seam should reproduce the loop's identity, which is what this
patch makes it do — but I did not want to assert a scope I had not proven, and
a maintainer will know this ordering better than I do.

This is the named-session analogue of #774, which fixed the same
`{{.AgentBase}}`-collapses-to-the-template-name symptom for namepool slots in
`PathContextForQualifiedName`. That fix did not cover the named-session path,
which resolves identity through a different seam.

## Fix

For a named bead carrying `configured_named_identity`, return that identity, so
rediscovery resolves the same `(agent, qualifiedName)` pair — and therefore the
same `WorkDir`/`AgentBase` — as the named-session loop does when it builds the
session in the first place. It reuses the existing `namedSessionIdentity` /
`namedSessionIdentityInfo` accessors rather than reading metadata directly.

Legacy beads with no identity stamp keep the previous behavior (template QN)
rather than resolving to an empty name; they pick up the correct identity on
their next fresh spawn.

The change is applied to **both** forms of the helper. The raw (`beads.Bead`)
form and the `session.Info` form are pinned to identical behavior by the
existing `TestCanonicalSessionIdentityWithConfigInfoMatchesRaw`, and live
rediscovery runs on the `Info` form — so changing only one form would either
trip that oracle or land in a path the reconciler no longer takes. The pool
branches of both functions are untouched; the change sits entirely behind the
existing `isNamedSessionBead` / `isNamedSessionInfo` guard.

## What this does NOT do

- It does not touch pool-slot identity or `PathContextForQualifiedName`.
- It does not repair the rest of the named-session `TemplateParams` contract.
  Rediscovery still does not reconstruct `GC_SESSION_ORIGIN=named` or
  `ConfiguredNamedIdentity`/`Mode` in `Env` the way the named-session loop does.
  This patch aligns only the `(agent, qualifiedName)` pair, which is what
  `WorkDir`/`AgentBase` derive from. The remaining Env-side drift is a separate
  problem and is called out in the updated doc comment.
- It does not migrate existing beads. A session whose worktree was already
  materialized at the shared template path is not moved; the next resolution
  simply points at the correct path.
- It does not address #1181 (implicit pool agents that have no `work_dir` at
  all). That is a different gap — no isolation configured, versus isolation
  configured but resolving to the wrong name.

## Testing

Run on Linux against this branch:

- `make lint` — pass, 0 issues.
- `make vet` — pass.
- `go test ./cmd/gc/... -short -run 'CanonicalSessionIdentity|Workdir|WorkDir|NamedSession'` — pass.
- `TestCanonicalSessionIdentityWithConfigInfoMatchesRaw` — pass (explicitly
  re-run, since this change has to keep the two forms in agreement).
- `make check` — **fails**, at the `test` target. Every failure reproduces on a
  clean `upstream/main` worktree at the same commit, so none of them are from
  this change. Details below, because I would rather show the work than assert
  "unrelated".
- `make test-integration` — **did not complete.** Details below.

Tests changed/added, all in `cmd/gc`:

- `TestCanonicalSessionIdentity` — the named-bead subtest previously asserted the
  template QN as expected behavior; it now asserts the stored alias.
- `TestCanonicalSessionIdentity` — new subtest for the unstamped-bead fallback.
- `TestRediscoveredNamedBeadWorkDirUsesAliasNotTemplate` — new end-to-end
  regression driving `canonicalSessionIdentity` →
  `resolveTemplateForSessionBeadInfo` and asserting `WorkDir` lands on the
  session's own path, not the template's.

Red/green verified: with the test changes applied but the production hunks
reverted to `main`, both new assertions fail with
`qn = "demo/worker", want demo/alpha`; with the fix they pass.

### Why `make check` fails here, and why it is not this change

`make test` produced 7 failures. I re-ran every one of them on a pristine
`upstream/main` worktree at the same commit:

| Test | On this branch | On clean `main` |
|---|---|---|
| `TestDetectStandaloneBdDoltLiveDoltForDifferentDataDirDoesNotConflict` | FAIL (30s timeout) | FAIL (30s timeout) |
| `TestStartBeadsLifecycleRefusesLiveStandaloneBdDolt` | FAIL (30s timeout) | FAIL (30s timeout) |
| `TestInitDirIfReadyDetectsStandaloneBdDoltAtProviderConvergence` | FAIL (30s timeout) | FAIL (30s timeout) |
| `TestRunBoundedPython3FallbackSendsSigtermBeforeKill` | FAIL | FAIL |
| `TestRunBoundedPython3FallbackEscalatesToSigkillAfterGrace` | FAIL | FAIL |
| `TestRunBoundedPython3FallbackPassesThroughOutputAndExitCode` | FAIL | FAIL |
| `TestRunSetupCommandActivityStreamingSurvivesIdleWindow` | FAIL under load | PASS |

The three `StandaloneBdDolt` timeouts look like the ambient-dolt collision
already tracked in #4677 / #4673 — my host has a dolt sql-server running.

`TestRunSetupCommandActivityStreamingSurvivesIdleWindow` is the one that did not
immediately reproduce, so I chased it: it passes 5/5 with `-count=5` on **both**
this branch and clean `main` when run on its own, and only failed inside the
full parallel run while the machine was heavily loaded. It is a timing-sensitive
tmux streaming test, and it lives in `internal/runtime/tmux`, which this patch
does not touch. I am treating it as load-induced flake rather than a regression,
but flagging it rather than burying it.

### Why `make test-integration` did not complete

I ran it, because the template asks for it when reconciler behavior changes and
this does. The `test/integration` package hit the 30-minute package timeout and
was killed with SIGQUIT after ~1980s. No named integration test reported a
failure — the only `--- FAIL` lines in the whole run were the same six above.

I am not claiming this gate as passing, and I am also not claiming the timeout is
unrelated on the strength of a hunch: my host was running several other full
suites concurrently, which is the most likely explanation, but I could not get a
quiet enough machine to produce a clean comparison run before opening this. If a
maintainer wants that comparison before merging, say so and I will produce it; if
CI runs the suite on a quiet box, that will be a better signal than anything I
can generate here.

**Environment note, in case it matters to a reviewer:** the Makefile points CGO
at Homebrew `icu4c` only on Darwin, and otherwise expects system ICU headers on
Linux. My host has linuxbrew `icu4c` but no `libicu-dev`, so I supplied
`CGO_CPPFLAGS`/`CGO_LDFLAGS` manually for all of the above. That is a local
environment quirk, not something this patch changes, and I mention it only so
the gate results above are interpretable.

## Related

- #774 — closed, the namepool-slot version of this same `{{.AgentBase}}`
  collision, fixed in a different resolver. Prior art, not a duplicate.
- #1181 — open, implicit pool agents sharing a working tree because they have no
  `work_dir` at all. Same class of consequence, different cause; not fixed here.
- #3582 — open, named work_dir precedence at session launch
  (`session_lifecycle_parallel.go`). Adjacent layer, no file overlap.

## Checklist

- [x] `make check` — run; fails only on pre-existing failures verified against
      clean `main` (table above).
- [ ] `make check-docs` — not run; no docs, navigation, or links changed.
- [ ] `make test-integration` — run but did not complete (30m package timeout);
      see above. Not claiming it.
- [x] Linked issues, or explained why one is not needed — no existing issue
      tracks this specific path; I did not open one separately since the
      description above is self-contained. Happy to file one if you would rather
      have it tracked.
- [ ] Updated docs for user-facing changes — none identified; this restores the
      documented `{{.AgentBase}}` contract rather than changing it. Point me at a
      page if you think one should change.
- [x] Called out breaking changes or migration notes — see below.

**Migration note.** For any `on_demand` named session that has already been
rediscovered at least once, the resolved `work_dir` changes from the shared
template path to the session's own path. That is the intended correction, but it
does mean a session can appear to "move" on the tick after upgrading, and any
uncommitted work sitting in the old shared directory stays there. I think that is
the right trade, but it is a behavior change worth a release note.

### DIFF
--- a/cmd/gc/build_desired_state.go
+++ b/cmd/gc/build_desired_state.go
@@ -2574,7 +2574,8 @@ func discoverSessionBeadsWithRoots(
 			// identity mismatch caused config-drift drains; the canonical shape
 			// still keeps routing/display identity and remaining fingerprint
 			// inputs aligned across buildDesiredState paths. Named beads
-			// intentionally pass through with the base shape (see
+			// resolve to their own stored alias so this seam reproduces the
+			// identity the named-session loop assigned (see
 			// canonicalSessionIdentity).
 			resolveAgent, sessionQualifiedName = canonicalSessionIdentityWithConfigInfo(cfg, cfgAgent, bInfo)
 		}
@@ -3267,19 +3268,22 @@ func resolveTemplateForSessionBeadInfo(
 // FingerprintExtra are part of CoreFingerprint, so divergent shapes across
 // ticks trip the reconciler's config-drift drain.
 //
-// Named beads are deliberately NOT canonicalized here. The named-session
-// TemplateParams contract (ConfiguredNamedIdentity/Mode, GC_SESSION_ORIGIN,
-// canonical session_name, ...) is authored by the main named-session loop
-// and reconstructNamedSessionTemplateParams; rewriting only the (agent,
-// qualifiedName) pair in rediscovery while leaving the rest of the shape
-// as plain ephemeral would produce a partially-named TemplateParams that
-// downstream consumers don't expect. The Env-side drift that named beads
-// can still exhibit across rediscovery vs. the named-session loop is a
-// separate fix — the accompanying PR explicitly scopes it out.
+// Named beads resolve to their own stored alias (configured_named_identity),
+// NOT the shared backing template's qualified name. The main named-session
+// loop builds the session with spec.Identity; this rediscovery seam must
+// return the same identity or ParseQualifiedName binds AgentBase to the
+// template's base name, and every named session backed by that template
+// resolves work_dir to one shared directory instead of its own. A named bead
+// with no stored identity (legacy, pre-stamp) falls back to the template QN,
+// unchanged from before. NOTE: this aligns the (agent, qualifiedName) pair and
+// therefore WorkDir/AgentBase; the rest of the named TemplateParams Env shape
+// (GC_SESSION_ORIGIN=named, ConfiguredNamedIdentity/Mode) is still authored
+// only by the named-session loop, so Env-side drift across rediscovery vs.
+// that loop remains a separate follow-up.
 //
 // Rules:
-//   - Named bead → (cfgAgent, cfgAgent.QualifiedName()). Identical to the
-//     pre-change rediscovery shape so named-bead handling is unchanged.
+//   - Named bead with stored identity → (cfgAgent, configured_named_identity).
+//   - Named bead without stored identity → (cfgAgent, cfgAgent.QualifiedName()).
 //   - Non-expanding agent → (cfgAgent, cfgAgent.QualifiedName()).
 //   - Instance-expanding agent with a stamped pool_slot → (deepCopyAgent
 //     at that slot, qualifiedInstance). Matches realizePoolDesiredSessions.
@@ -3294,6 +3298,17 @@ func canonicalSessionIdentityWithConfig(cfg *config.City, cfgAgent *config.Agent
 		return nil, ""
 	}
 	if isNamedSessionBead(bead) {
+		// A named-session bead's work_dir/AgentBase must resolve from its own
+		// stored alias, not from the shared backing template. The named-session
+		// loop builds the session with spec.Identity; this rediscovery seam must
+		// return the same identity or ParseQualifiedName yields the template's
+		// base name, collapsing every named session on that template onto one
+		// work_dir. In practice this bites on_demand sessions, which can fall
+		// through the named-session loop's mode gate and get their identity
+		// recomputed here.
+		if identity := namedSessionIdentity(bead); identity != "" {
+			return cfgAgent, identity
+		}
 		return cfgAgent, cfgAgent.QualifiedName()
 	}
 	if cfgAgent.UsesCanonicalSingletonPoolIdentity() {
@@ -3317,6 +3332,15 @@ func canonicalSessionIdentityWithConfigInfo(cfg *config.City, cfgAgent *config.A
 		return nil, ""
 	}
 	if isNamedSessionInfo(info) {
+		// Mirror of the raw form's named-bead branch: resolve from the
+		// session's own stored alias so rediscovery reproduces the identity
+		// the named-session loop assigned, and work_dir does not collapse
+		// onto the backing template's path. Both forms must change together —
+		// TestCanonicalSessionIdentityWithConfigInfoMatchesRaw pins them to
+		// identical behavior. See canonicalSessionIdentityWithConfig.
+		if identity := namedSessionIdentityInfo(info); identity != "" {
+			return cfgAgent, identity
+		}
 		return cfgAgent, cfgAgent.QualifiedName()
 	}
 	if cfgAgent.UsesCanonicalSingletonPoolIdentity() {
--- a/cmd/gc/build_desired_state_test.go
+++ b/cmd/gc/build_desired_state_test.go
@@ -11171,20 +11171,40 @@ func TestCanonicalSessionIdentity(t *testing.T) {
 		}
 	})
 
-	t.Run("named bead keeps base identity (out of scope for this canonicalization)", func(t *testing.T) {
-		// Named-session TemplateParams carry ConfiguredNamedIdentity/Mode,
-		// GC_SESSION_ORIGIN=named, and a canonical session_name set by the
-		// main named-sessions loop and reconstructNamedSessionTemplateParams.
-		// Rewriting just the identity qualifier in rediscovery without also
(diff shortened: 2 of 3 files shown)


wynged:
Following up on the `make test-integration` gap in the description — I said I would produce the clean-`main` comparison if I could get the machine quiet enough, so here it is.

I ran `go test -tags integration -timeout 30m ./test/integration/` on a pristine `upstream/main` worktree at the same commit (`a699601f0`), on the same host:

- **clean `main`**: FAIL after 1878s — `--- FAIL: TestHumaBinary_CityCreateAsync (41.26s)`
- **this branch**: FAIL after ~1980s — killed by the 30m package timeout (SIGQUIT), no named test failure

So `test/integration` does not pass on clean `main` on this host either. The two runs fail in different ways, which I read as the host rather than either tree: the box was sitting at a load average around 47–50 for the whole window, and these are supervisor-boot/timing tests. I don't think this tells you anything about the patch in either direction — I'm reporting it because I said I would, not because I think it's evidence of correctness.

The signal I'd actually trust here is this PR's own CI, which shards the integration suite across clean runners. At the time of writing it is green so far with the remaining shards still going, and it will be a better answer than anything my machine can produce.

One correction to the description while I'm here: it says the `test/integration` timeout was the only integration-level result I had. The baseline run above is the comparison it says I hadn't done.

wynged:
CI has settled: 63 passing, 22 skipped, nothing failing. All 15 `Integration /` shards are green, including `packages-cmd-gc-integration` and `packages-core-*`.

That answers the `make test-integration` question I left open above better than my host could — the suite passes on clean runners, so the 30m timeout I hit locally was the loaded machine, not this patch. No further action needed from my side on that gate.

quad341 (approved):
@wynged — the thing I want to call out here isn't just the fix, it's how you scoped it. You found that the rediscovery seam and the named-session loop disagree about identity, and instead of asserting the patch was safe you told us precisely where your confidence ran out: you'd observed the `on_demand` fall-through, you had *not* proven `always` sessions can't reach the same seam, and you said so in the body. That made this fast to verify.

For what it's worth, I chased the deleted comment's warning about "partially-named TemplateParams" and it doesn't bite — `templateNameFor` keeps `GC_TEMPLATE` and `TemplateName` pinned to the backing template even when the qualified name is the alias, and `GC_TEMPLATE` is the only identity key in the fingerprint allow-list. Your scope-out was the right boundary, not a gap. Changing both the raw and `Info` forms together was also correct; the existing equivalence oracle already carries a stamped and an unstamped named bead, so both new branches are pinned for free.

Merging as-is. Two things I'll note in the release note on our side rather than ask you for: the first tick after upgrade will likely *drain and restart* an affected session (since `pre_start` is fingerprinted through `AgentBase`), and `dir`-less rig-scoped templates will now resolve a rig scope at this seam where they previously resolved none. Both look like the intended correction.

The clean-`main` baseline table was more work than we'd have asked for, and CI on quiet runners agrees with your read. Thank you.


---

PR #5096, opened 2026-08-07
test(cmd/gc): tear down managed dolt after init-from without hosted endpoint (#5066)

### LINKED ISSUE #5066: TestInitFromWithoutHostedPreservesTemplate leaks a dolt sql-server (regression in 97503b283..cab9da1c8)
## Summary

`TestInitFromWithoutHostedPreservesTemplate` leaks a `dolt sql-server` process at current `main`; the cmd/gc dolt leak guard then fails the package. Reproducible standalone in ~1.5s.

## Repro

```
go test ./cmd/gc -run TestInitFromWithoutHostedPreservesTemplate -count=1
```

```
cmd/gc test dolt leak guard: leaked 1 dolt sql-server process(es) under /tmp/gct559587-1314394083
  pid=560283 argv="dolt sql-server --config /tmp/gct559587-1314394083/TestInitFromWithoutHostedPreservesTemplate529475619/002/city/.gc/runtime/packs/dolt/dolt-config.yaml"
FAIL	github.com/gastownhall/gascity/cmd/gc	1.593s
```

The leaked server belongs to the test's second city (`.../002/city/...`).

## Isolation

- Leaks at `cab9da1c8` (#5061), tested on a detached checkout with no local changes.
- Does **not** leak at `97503b283` (#4851) — same box, same invocation, clean pass.
- So the regression landed somewhere in `97503b283..cab9da1c8` (not bisected further).

Consistent across repeated runs (also failed two independent full pre-push suite runs on the same commit range).

## Environment

- linux/amd64, go1.26.5, dolt 2.2.3

vishnujayvel:
## Summary

- `TestInitFromWithoutHostedPreservesTemplate` was leaking a managed `dolt sql-server` after `finalizeInit` → `initDirIfReady` started one for the copied template's local beads backend. With `noStart=true` the test never handed off to the supervisor, so the server outlived the test and the package dolt leak guard failed the suite even though the test's own assertions passed.
- Teardown now uses the existing `cleanupManagedDoltTestCity` helper (`shutdownBeadsProvider` + stop of any remaining managed servers under the city path).
- Reported by @jm2 in #5066. Related #5057 proposes to widen leak-guard *visibility* (source-tree roots); this is the teardown fix for the named temp-root culprit.

## Why teardown in the test, not in `noStart`

A fair question is whether `noStart` itself should stop the server `initDirIfReady` started, making this a product fix instead. Two reasons this PR stays at the test layer:

- Leaving the managed server running after init is the existing product behavior — the initialized city is live and usable; `noStart` only skips supervisor handoff. Tests in this package own their teardown for exactly this case: `cleanupManagedDoltTestCity` has 22 call sites in `cmd/gc`, including two (`script_resolve_test.go`, `cityinit_impl_test.go`) that register it in this identical pre-init shape. Registering before init is what gives the helper a clean PID baseline, and every teardown step no-ops safely if init fails early.
- #5066 is scoped to the leaking test, and changing `noStart` semantics would be a behavior change for all callers.

If you'd rather `noStart` stop the server it started, happy to rework at that layer instead — flagging the option rather than deciding it unilaterally.

## Testing

- [x] `go test ./cmd/gc -run '^TestInitFromWithoutHostedPreservesTemplate$' -count=1` — PASS, no package leak-guard failure, no leftover `dolt sql-server` under the test temp root
- [x] `go test ./cmd/gc -run 'TestInitFrom(PinsHosted|WithoutHosted|RejectsIncomplete|RejectsDolt|HostedPreserves)' -count=1` — all PASS
- [x] `gofmt` clean on touched file
- [ ] `make check` (not run end-to-end; change is one-line test teardown)
- [ ] `make check-docs` N/A
- [ ] `make test-integration` N/A

### Before (verbatim leak-guard failure)

```
cmd/gc test dolt leak guard: leaked 1 dolt sql-server process(es) under /var/folders/.../gct...
  pid=... argv="dolt sql-server --config .../TestInitFromWithoutHostedPreservesTemplate.../002/city/.gc/runtime/packs/dolt/dolt-config.yaml"
```

### After

```
--- PASS: TestInitFromWithoutHostedPreservesTemplate
PASS
ok  	github.com/gastownhall/gascity/cmd/gc
```

## Checklist

- [x] Linked an issue: closes #5066 (credits @jm2)
- [x] Added or updated tests for behavior changes (test teardown only; production code unchanged)
- [x] Updated docs for user-facing changes — N/A

Closes #5066

### DIFF
--- a/cmd/gc/init_from_hosted_dolt_test.go
+++ b/cmd/gc/init_from_hosted_dolt_test.go
@@ -71,6 +71,11 @@ func TestInitFromWithoutHostedPreservesTemplate(t *testing.T) {
 
 	src := gastownExamplePath(t)
 	cityPath := filepath.Join(t.TempDir(), "city")
+	// initDirIfReady (via finalizeInit) starts a managed dolt sql-server for the
+	// copied template's local beads backend. noStart skips supervisor handoff
+	// but leaves that server running — tear it down so the package leak guard
+	// does not fail the suite (gastownhall/gascity#5066).
+	cleanupManagedDoltTestCity(t, cityPath)
 
 	var stdout, stderr bytes.Buffer
 	// disabled hosted options => template preserved


AJBcoding:
@quad341 — would you be able to take a look at this one when you have a moment? Same
situation as #5103: I can't add a reviewer via the reviewers field (outside contributor,
pull-only access here), so flagging in a comment instead. This is @vishnujayvel's PR, not
mine — I arrived at the identical fix independently and am closing my duplicate (#5110) in
favor of this one, which was first and is better documented.

Adding the verification I gathered, since it may save a review cycle. Measured on current
`main` (`014aee18b`), both directions, `-count=1`:

**Without the fix**

```
PASS                                     <- the tests themselves pass
cmd/gc test dolt leak guard: leaked 1 dolt sql-server process(es) under
  /var/folders/.../gct50201-1391073147
  pid=51221 argv="dolt sql-server --config
  .../TestInitFromWithoutHostedPreservesTemplate3153946113/002/city/..."
FAIL  github.com/gastownhall/gascity/cmd/gc  8.291s
```

**With the fix**

```
--- PASS: TestInitFromWithoutHostedPreservesTemplate (7.53s)
ok    github.com/gastownhall/gascity/cmd/gc  9.517s
```

The leaked process's own argv names the test, so attribution is direct rather than
inferred. Note the failure shape: the tests pass and the *post*-test `TestMain` guard is
what reddens the shard — easy to misread as a flake.

Why it's worth landing: this is the last non-transient failure in the `cmd/gc` shard of
`make test-fast-parallel`, which runs as the pre-push hook. Until it lands, anyone
touching Go has to push with `--no-verify` to get around a leak unrelated to their change,
which also masks real gate failures. CI here is green — 60 passing, 0 failing.

vishnujayvel:
Thank you @AJBcoding — both for closing #5110 in favor of this one and for attaching the bidirectional verification; the argv-level attribution of the leaked process is a nicer proof than what I had in the PR body. Agreed on the failure shape being easy to misread as a flake: the tests pass and it's the post-test `TestMain` guard that reddens the shard. Happy to rebase promptly if @quad341 wants any changes.

Issue #4934, opened 2026-08-03
cmd/gc and internal/convergence tests fail on macOS after #4930 (firmlink /var vs /private/var, /home vs /System/Volumes/Data/home)


wbern:
Four tests fail on macOS on current `main`. They pass on Linux, which is why CI is green.

I hit this while merging `main` down into a fork and initially assumed I'd broken something — so I checked it against a clean, unmodified `upstream/main` worktree. It reproduces there, untouched.

## Failing on pristine `main` (darwin/arm64)

```
cmd/gc               TestResolveImportRootPrefersNearestPackUnderCity
cmd/gc               TestResolveImportRootFallsBackToStandalonePackDir
internal/convergence TestResolveEvaluateStep_DefaultPath
internal/convergence TestResolveEvaluateStep_CustomPath
```

`cmd/gc/cmd_import_test.go:2762`:

```
resolveImportRoot() = "/var/folders/.../packs/tools"
want nearest pack   = "/private/var/folders/.../packs/tools"
```

`internal/convergence/evaluate_test.go:25` and `:44`:

```
PromptPath = "/System/Volumes/Data/home/user/city/..."
want         "/home/user/city/..."
```

## Mechanism

#4930 routes synthetic paths through `pathutil.NormalizePathForCompare`. macOS resolves firmlinks — `/var` → `/private/var`, `/home` → `/System/Volumes/Data/home` — so the normalized form no longer string-equals the literal form the assertions were written against. On Linux there are no firmlinks, both forms are already identical, and every assertion passes.

So this is a test/assertion mismatch rather than a production-behaviour bug, as far as I can tell — the normalization itself looks like the intended change.

## Why I'm reporting rather than fixing

Two reasons, and I want to be straight about the second one.

First, the fix is a judgement call that belongs with whoever wrote #4930: either normalize both sides of the comparison in the tests, or have the assertions compare via `NormalizePathForCompare` rather than string equality. I don't want to guess which shape you'd prefer in your own tests.

Second, honestly: I'm not planning to spend more time on this one right now. Filing it because a Linux-CI-invisible break on a supported dev platform seemed worth you knowing about, and because the next person to hit it will do what I did — assume it's their own tree — and burn the same time working out that it isn't. Happy to be told it's known or intended.

## Repro

```sh
git worktree add --detach /tmp/pristine <current main>
cd /tmp/pristine
[...]

PR #5097, opened 2026-08-07
test: compare paths via firmlink-aware helpers after #4930 (#4934)

vishnujayvel:
## Summary

- Fix macOS-only test failures introduced after #4930's path-normalization change.
- Four tests compared production paths (via `pathutil.NormalizePathForCompare`) against bare `filepath.EvalSymlinks` or literal strings. (The diff touches a fifth comparison in the same file, converted for consistency — it already passed.) On macOS those spellings diverge because of firmlinks:
  - `/var` ↔ `/private/var` (temp dirs under `$TMPDIR`)
  - `/home` ↔ `/System/Volumes/Data/home` (literal fixture city paths)
- On Linux there are no firmlinks, so the two forms are identical and CI stayed green.
- Comparison sites now use the repo's existing helpers (`assertSameTestPath` / `testutil.AssertSamePath`), which apply the same production normalization on both sides. No product path semantics changed.

Credits: reported by @wbern in #4934.

## Firmlink note for Linux reviewers

macOS exposes some directories as **firmlinks** — OS-level path aliases. `filepath.EvalSymlinks` expands them to the physical spelling (`/private/var/...`, `/System/Volumes/Data/home/...`), while `pathutil.NormalizePathForCompare` (used by production code after #4930) collapses the `/private/{tmp,var}` aliases back to `/tmp`/`/var` and also resolves firmlinks such as `/home`. Asserting with raw string equality therefore fails on darwin even when both sides name the same location.

## Testing

- [x] Reproduced the four listed failures on darwin/arm64 at base `06f8b61`
- [x] After fix: listed tests pass
  - `go test ./cmd/gc/ -count=1 -run '^(TestResolveImportRootPrefersNearestPackUnderCity|TestResolveImportRootFallsBackToStandalonePackDir)$'`
  - `go test ./internal/convergence/ -count=1 -run '^TestResolveEvaluateStep_'`
- [x] `go test ./internal/convergence/ -count=1` — ok
- [x] `go test ./cmd/gc/ -count=1 -run 'TestResolveImportRoot|TestFindNearestImportRoot'` — ok
- [ ] Full `go test ./cmd/gc/ -count=1` hit the package ~10m budget and aborted with parallel-test goroutine dumps (pre-existing package size / timeout issue on this machine; not caused by this test-only change)
- [ ] `make check` (not run; test-only change)

## Checklist

- [x] Linked an issue: #4934
- [x] Added or updated tests for behavior changes (comparison-site only)
- [ ] Updated docs for user-facing changes (n/a)
- [ ] Called out breaking changes or migration notes (n/a — tests only)

### DIFF
--- a/cmd/gc/cmd_import_test.go
+++ b/cmd/gc/cmd_import_test.go
@@ -2679,13 +2679,8 @@ schema = 1
 	if err != nil {
 		t.Fatalf("resolveImportRoot: %v", err)
 	}
-	want, err := filepath.EvalSymlinks(dir)
-	if err != nil {
-		t.Fatalf("EvalSymlinks(%q): %v", dir, err)
-	}
-	if got != want {
-		t.Fatalf("resolveImportRoot() = %q, want %q", got, want)
-	}
+	// production paths are NormalizePathForCompare form; compare firmlink-aware (#4934)
+	assertSameTestPath(t, got, dir)
 }
 
 func TestFindNearestImportRootSkipsRuntimeOnlyDirs(t *testing.T) {
@@ -2754,13 +2749,7 @@ func TestResolveImportRootPrefersNearestPackUnderCity(t *testing.T) {
 	if err != nil {
 		t.Fatalf("resolveImportRoot: %v", err)
 	}
-	want, err := filepath.EvalSymlinks(packDir)
-	if err != nil {
-		t.Fatalf("EvalSymlinks(%q): %v", packDir, err)
-	}
-	if got != want {
-		t.Fatalf("resolveImportRoot() = %q, want nearest pack %q", got, want)
-	}
+	assertSameTestPath(t, got, packDir)
 }
 
 func TestResolveImportRootRuntimeOnlyAncestorResolvesRegisteredRigCity(t *testing.T) {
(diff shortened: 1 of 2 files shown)


sjarmak:
Merging. Review is clean: the fix routes the `want` side of each assertion through the same `pathutil.NormalizePathForCompare` that production already calls at that site (`internal/convergence/evaluate.go:57`, `cmd/gc/cmd_import.go:312`), so it aligns the oracle to the production path rather than weakening it. Clean merge against current main, and #5111 did leave both files untouched as intended.

Being precise about what the macOS evidence covers, since the check on this PR did not run one:

The `Mac Regression: pass` reported here executed nothing. It is a fork PR without the `needs-mac` label, so the gate short-circuited with `run_smoke=false` and exited 0. I dispatched the suite manually against this head (`ac3d8f4d6`). In that run `cmd-gc` and `acceptance` and `quality` all passed on real macOS hardware, which covers `TestResolveImportRootPrefersNearestPackUnderCity` and `TestResolveImportRootFallsBackToStandalonePackDir`.

The other two, `TestResolveEvaluateStep_CustomPath` and `TestResolveEvaluateStep_DefaultPath`, live in the unit tier, and that job hit its 25 minute `timeout-minutes` cap at 25m16s rather than reporting a result. That cap is unrelated to this change, which touches two test files. The nightly schedule does run the unit tier and is currently failing, so the next nightly is the natural place this shows up.

Two of the four verified green on macOS, two blocked by a CI capacity limit that predates this PR, and no evidence pointing the other way. That is enough to land a change that can only move the nightly toward green.

Follow-up worth having, not blocking: `internal/pathutil` has no direct regression test for the `/System/Volumes/Data` ancestor-walk case. The existing tests cover the `/private/var` and `/private/tmp` aliases only, and this fix works by making both sides agree rather than by collapsing that prefix, so nothing currently pins that behavior on its own.

Issue #2081, opened 2026-05-13
bug: controller sweeps actively-running session based on stale session-state view (kills in-flight work)


mike-matchpoint:
> Related: #1029 covers a race where `assignedWorkBeads` briefly hits zero during step transition. This issue is a different angle — the session-state field (`failed-create`/`asleep`) in the bd-backed view diverges from tmux ground truth, and the reconciler kills live working sessions based on the stale view. May share root cause with the persistence-bug family.

> Origin: filed locally as `vgc-9lz1` in a downstream city. Mirroring upstream for triage.

---

## Observed

Session `vgc-pnp1` (polecat targeting `Matchpoint-Vehicle-Graph/gastown.nux`) was actively running pytest on bead `vg-1jp` via a live tmux pane (`gastown__polecat-vgc-pnp1`). Within ~5 minutes:
1. `gc session list --json` started reporting `vgc-pnp1` State=`failed-create` while tmux confirmed the pane was active and `gc session peek vgc-pnp1` showed real pytest output (19 passes, package installs, planning narration).
2. On a subsequent check, the tmux pane had been killed. `tmux -L vehicle-graph-city ls` no longer listed `gastown__polecat-vgc-pnp1`.
3. `vg-1jp` bead status reverted from IN_PROGRESS back to OPEN (assignee cleared).
4. The polecat's git branch `polecat/vg-1jp` (commit c58a831) survived — pushed to `origin/polecat/vg-1jp` before the sweep — so the work output wasn't lost, but the bead-level claim/coordination state was.

## Root cause hypothesis

The controller's session-state view (backed by the beads store for session beads) diverges from tmux ground truth. The persistence bugs in `vgc-5wqq` (close) and `vgc-96yy` (set-metadata) confirm session-state writes don't propagate consistently into the bd store. The controller then makes sweep decisions based on the stale view — concludes `vgc-pnp1` is `failed-create`, sweeps it, kills the actually-live tmux pane.

This is the third confirmed failure mode in the same family:
- vgc-5wqq: `bd close` doesn't persist to jsonl
- vgc-96yy: `bd update --set-metadata` doesn't persist to jsonl (silently drops `gc.routed_to` from `gc sling`)
- THIS: session-state field doesn't propagate to the view the controller reads → controller acts on stale data → kills live sessions

## Impact

[...]

PR #5098, opened 2026-08-07
feat(runtime): prove session liveness, attach, and stop against the exact incarnation

julianknutsen:
## What this hardens

Everything here is about a caller being able to **prove** what the runtime
actually is, rather than inferring it from a cached or name-addressed view.
Six related gaps, in order:

1. **Cached liveness could not be discarded** (`LivenessInvalidator`). A
   caller that had just changed the world still read the pre-change snapshot
   out of the tmux state cache. `auto` invalidates both backends, because a
   stale route table must not hide the backend that actually hosts the
   session.
2. **`Liveness` could not say "I proved this"** (`Liveness.Complete`,
   `LivenessTarget`, `FreshLivenessObserver`, `ObserveFreshLiveness`). There
   was no way to distinguish *provably absent* from *not seen*, which is
   exactly the distinction a caller needs before destroying durable state.
   Providers without the capability keep positive evidence but report absence
   as incomplete, so the fallback fails closed.
3. **Process scans could not bound incarnation uncertainty**
   (`proctable.ScanBySessionIDSince`). On a shared host a session-ID scan
   could not tell a live process of the current incarnation from an
   inaccessible leftover of a previous one. A process may now be ignored only
   when `/proc` proves it predates the incarnation. Linux also stops reporting
   a read error for every foreign-uid `/proc` entry, and starts reporting one
   for a same-uid entry it cannot read. That second half is deliberate: an
   entry this user *owns* but may not read — a setuid binary it launched, such
   as `passwd` or `pkexec`, whose `environ` is root-owned and non-dumpable —
   used to be skipped silently, and is now surfaced so the absence of anything
   behind it is marked unproven rather than rendered as "not running". The
   unbounded `ScanBySessionID` names no incarnation for the start-time and
   tmux-lineage proofs to work against, so every such entry stays unresolved
   on that path; both of its callers log the error and continue using the
   returned slice, which is always populated. Platforms with no process-table
   support at all refuse a bounded scan with
   `ErrIncarnationBoundUnsupported` rather than returning the empty inventory
   that would read as a certified absence.
4. **Stop conflated a missing session with a missing server.** A responsive
   empty named server is a complete inventory witness; a missing server is
   not, and must never certify absence. This is a property of the **leaf**
   provider surface, not a guarantee every caller of a composed provider
   observes: `auto.Stop` merges both backends through
   `runtime.MergeBackendStopErrors` and deliberately collapses every
   session-gone shape — a missing server included — to success, because
   neither backend can see the session and there is nothing left to stop.
   Teardown callers that only need "it is not running" absorb both shapes
   through `runtime.StopForCleanup`, the single place that rule lives;
   callers that need the distinction ask the leaf.
5. **Named tmux attach could rebind the city's socket.** `attach-session`
   against a configured per-city socket would start a fresh server when the
   socket was missing — silently replacing the city's server with an empty
   one. Observational checks, bounded preflight, and the final attach now pass
   `-N`, and every attach launch is fenced by a socket witness captured
   beforehand, so a socket replaced in between is refused rather than adopted.
   Default-socket behavior is unchanged.
6. **A destructive stop addressed a session by name**
   (`UnattendedSessionStopper`). `StopUnattendedSession` proves the target
   still carries the expected `GC_INSTANCE_TOKEN`, has no attached client and
   no copy-mode pane, then kills that exact incarnation by tmux `session_id`.
   Any observation error, replacement, attachment, or copy mode refuses
   without stopping anything. `auto` and `hybrid` forward only to the backend
   selected for the session and never fall through to the other one. `auto`
   keeps the route entry across a successful stop on purpose: an accepted stop
   is a request, not an observation, and only a certified absence — generation-
   checked, so a replacement session keeps its own entry — may evict it.

Plus the state-cache read path, which is where cache invalidation had to
become observable. `currentState` moved from a single unconditional refresh
behind a fixed singleflight key to a generation-keyed coalescing loop that
retries a refresh a concurrent invalidation superseded — a change to a live
path, not a latent one; see the scope note below. `freshState`, new here,
demanded that one exact generation publish the snapshot, so a concurrent
refresh that superseded it reported incomplete even though a newer usable
snapshot existed; it now accepts the superseding generation. Both loops are
bounded by `maxRefreshRetries`, so sustained churn degrades to the existing
stale-tolerance exit instead of spinning, and a read's worst case stays at
`(1 + maxRefreshRetries) * fetchTimeout`. Coalescing is per cache generation
by design: an invalidation opens a new generation, so a caller arriving after
one never joins — or accepts the result of — the flight it superseded.

## Why this is reconciler-independent

Every item is a capability of `internal/runtime` with no knowledge of who
calls it. The four new optional extensions (`LivenessInvalidator`,
`FreshLivenessObserver`, `UnattendedSessionStopper`, plus the
`LivenessTarget` descriptor) are additive interfaces: a provider that does
not implement one is unaffected, and `ObserveFreshLiveness` degrades to the
existing `ObserveLiveness` with `Complete=false`. Nothing in `internal/` was
made to depend on `cmd/gc`.

**Scope of the "nothing else changes" claim.** It holds for the four new
interfaces, and for a concrete reason: all four have **zero production
consumers in this tree**, so no existing code path can route through them.
That is deliberate — this PR lands the substrate ahead of the consumer, and
the surface is exercised only by its own unit tests plus integration-tagged
real-tmux rows. The claim does **not** extend to the whole diff. Two existing
behaviors change on purpose:

- **`currentState` coalesces per generation and retries a superseded
  refresh.** Every `IsRunning` and `ProcessAlive` caller goes through it, so
  retry count, coalescing scope and subprocess fan-out all shift under session
  churn. The retry is bounded by `maxRefreshRetries` and falls back to the
  pre-existing stale-tolerance exit, but this is a change to a live path, not
  to a latent one.
- **The missing-session/missing-server split reaches the existing kill
  paths.** `Tmux.KillSessionWithProcesses` and
  `KillSessionWithProcessesExcluding` now absorb only `ErrSessionNotFound` and
  `ErrNoCurrentTarget`, so `Provider.Stop` propagates a missing server where
  it used to return nil. The cleanup callers downstream —
  `Manager.CloseDetailed` (and `Manager.Close` through it), `Manager.Kill`,
  and `worker.RuntimeHandle`'s `Stop`, `Kill` and `Close` — route through
  `runtime.StopForCleanup`, so what an operator observes is unchanged; a
  caller that wants the distinction now has to ask for it.

The `-N` attach fence and the missing-session/missing-server split are
straight bug fixes to the tmux provider that stand on their own.

## Test evidence

`go build ./...` and `go vet ./...` are clean.

Focused packages (`-count=1`):

```
ok  	github.com/gastownhall/gascity/internal/runtime	15.617s
ok  	github.com/gastownhall/gascity/internal/runtime/auto	0.019s
ok  	github.com/gastownhall/gascity/internal/runtime/hybrid	0.008s
ok  	github.com/gastownhall/gascity/internal/runtime/proctable	0.096s
ok  	github.com/gastownhall/gascity/internal/runtime/tmux	10.098s
ok  	github.com/gastownhall/gascity/internal/beads	22.054s
ok  	github.com/gastownhall/gascity/internal/session	22.499s
ok  	github.com/gastownhall/gascity/internal/worker	85.139s
ok  	github.com/gastownhall/gascity/scripts	102.629s
```

Full fast-unit baseline (`make test`):

```
observable go test: command=go test -json -p=4 -count=1 -timeout 15m ./...
pass github.com/gastownhall/gascity/scripts
observable go test: PASS log=/var/tmp/gascity-test.jsonl.EznpHb
```

`scripts/runtime-tmux-tests.manifest` is a checked, source-first inventory
with exact counts and a six-way shard partition, so the new tmux tests are
registered and the pinned counts refreshed in the same commit.

`internal/runtime/herdr`'s `TestSetupCommandFailureOmitsInheritedCredentials`
fails on this branch and on untouched `main` alike; this PR changes no file in
that package.

## Provenance and scope

Extracted from a larger reconciler integration branch and rebased onto
current `main`. Only the `internal/runtime` slices that stand alone were
taken. Deliberately **not** included, because their only in-tree consumers
live on that branch:

- the nudge input-fencing family (`ErrInputFenced`, `FencedNudgeProvider`,
  `NudgeFenced`, copy-mode nudge parking). Landing it here would change
  `Nudge`/`NudgeNow` behavior for hidden-attached and copy-mode sessions with
  nothing in `cmd/gc` to release the queued nudge claims on the new error, and
  it collides with the newer `SendKeys` hidden-attach poke tests on `main`.

---
- Related to #2081 — lands the runtime substrate for proving a session genuinely absent before anything destructive happens — fresh liveness observations that can report "provably absent" vs "not seen", cache invalidation so a caller can discard a pre-change snapshot, and a stop that refuses unless the target still carries the expected instance token and has no attached client — but it does not change the controller's sweep path, which still acts on the cached session-state view, so the reported behavior is not addressed by this PR alone


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277174286

julianknutsen:
@gc re-requesting review after rebase onto main @ `00eb363cdc4555bbf5182fdd16662282a023af3f`

Rebased from `84dfcc885c` (110 commits behind) onto `origin/main@3ba850ff60`. All 10 commits replayed; PR is now `MERGEABLE/CLEAN`.

**One conflict, trivial — `scripts/runtime_tmux_manifest_test.go`.** Test-count bookkeeping only. Upstream #5250 added 5 tmux manifest entries, this branch adds 25; `scripts/runtime-tmux-tests.manifest` itself auto-merged, only the three hardcoded counter literals in the guard test collided:

| counter | base | main (#5250) | this PR | resolved |
|---|---|---|---|---|
| `len(manifest)` | 357 | 362 | 382 | **387** |
| `len(untagged)` | 240 | 245 | 262 | **267** |
| `wantShardCounts` | [60,60,60,59,59,59] | [61,61,60,60,60,60] | [64,64,64,64,63,63] | **[65,65,65,64,64,64]** |

Not arithmetic — the guard test re-derives every count by parsing the real Go AST, and `TestRuntimeTmuxManifest*` passes against the resolved values, so 387/267/120 is the discovered inventory. The integration-only `declared-untagged` counter (120) auto-merged because all 5 upstream additions were untagged.

**Gates:** `go build ./...` · `go vet ./internal/runtime/... ./internal/session/... ./scripts/...` · `gofmt -l` empty on all touched files · `go test` across 21 packages (`internal/runtime/...`, `internal/session/...`, `scripts/...`) all `ok`. No pre-existing main failures encountered.

**Heads-up for #5290.** It did not collide, because upstream's only `tmux.go` edits were in the nudge/submit region (~64-91, ~2001-2040, ~2138-2174) while this PR's hunks are at 315/386/889/1213/1492-1610/2503/2932 — liveness, attach, stop. #5290 targets tmux *liveness* too, so it overlaps this PR's territory rather than #5250's. This branch moved first; expect a real conflict when #5290 rebases.

A3Ackerman:
Heads-up on a file overlap, flagging it early so it can be sequenced deliberately rather than discovered at merge time — no action needed from you right now.

This PR and #5392 both rewrite `internal/runtime/proctable/scan_darwin.go`, `scan_linux.go`, `scan_stub.go`, `internal/runtime/runtime.go`, and `internal/runtime/tmux/adapter.go`, with different designs for the same seam — you add `ScanBySessionIDSince` and an incarnation bound; #5392 replaces the `ScanBySessionID(id)` signature with a typed `ProcessTarget`. Whoever lands second is going to eat a real conflict in the scan functions, not a mechanical one.

I have no stake in the ordering, and #5098 is both larger and from a maintainer, so I am happy for it to go first. One thing worth knowing if it does:

#5392 carries a two-line fix for a **whole-city outage** that #5098 does not close, and it will need re-applying on top of your new scan shape. The bug: `isInfrastructureCommand` is applied to the *parent* record in the Darwin/Linux scans but never to the record itself, so a tmux server is eligible to be reported as an agent root. It reads as a member of the mayor's session because `ps eww` prints argv and environ concatenated and the server permanently retains the `-e GC_SESSION_ID=...` argv of the `new-session` that founded it. After a mayor handoff the tmux *session* is gone while the *server* lives, `IsTracked` goes false, and `killExistingOrphans` → `TerminateRuntime` → `proctable.KillByPID` → `kill(-pid)` takes out the server — and with it every session on the socket. We have lost a 14-session fleet to this.

Since you are touching `isRootWithSessionID` / the Linux root rule directly: the relevant detail is that `ppid <= 1` currently returns "is a root" unconditionally, which is the Linux half of the same hole.

Full measurement and the verified kill chain are in #5392's description if useful. Flagging rather than asking for anything — happy to rebase #5392 onto this whenever it lands.

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277291765

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277291907

PR #4863, opened 2026-07-31
feat(beads): add exact-store metadata CAS CLI

jm2:
## Summary

Add an exact-store, fail-closed metadata compare-and-set command:

```text
gc beads metadata-cas <bead-id> \
  --store-ref=city:<name>|rig:<name> \
  --key <key> --expected <value> --next <value> --json
```

The command opens only the explicitly selected authoritative store, performs
one single-key CAS, and returns one of `swapped`, `already_next`, or `conflict`.
It never scans another store and has no unconditional-write fallback.

## Safety details

- Split the narrow `MetadataCASWriter` capability from the revision-fenced
  `ConditionalWriter` interface. Native Dolt can soundly implement the former
  without claiming the latter while beads v1.1.0 lacks a durable revision
  token.
- Classify a false CAS through an authoritative readback on the same store
  handle.
- Reset the callback-local swap result on every upstream Dolt transaction
  callback invocation. This prevents a retry after an abandoned write attempt
  from leaking a stale success when a competitor wins before the retry.
- Close the selected store successfully before emitting any success payload.
- Use the shared canonical JSON failure envelope for normal `--json` usage,
  while retaining `--format=json` compatibility.

## Validation

- `go test ./internal/beads -count=1`
- `go test ./cmd/gc -count=1 -timeout=20m`
- focused retry regression under `-race`
- focused real-Dolt sequential, same-handle contention, and
  independent-handle contention tests
- `go vet ./...`
- `go build ./cmd/gc`
- generated schema, documentation, metrics, command-census, and tutorial
  freshness checks

The retry regression uses a transaction double that restores the first
callback's pre-attempt snapshot before committing a competitor value and
invoking the callback again. As a negative control, removing the per-callback
result reset makes that test fail with a false successful CAS.

PR #5099, opened 2026-08-07
feat(beads): make the revision fence portable, reachable, and atomic at close

julianknutsen:
## What this hardens

Four independent gaps in the conditional-write path, plus one cache primitive.

**1. The revision contract required monotonicity.** That is a property of a
counter, not of a fence, and it excluded any backend whose
optimistic-concurrency token is a row version rather than a running total. The
contract is restated as what conditional writes actually need: a mutation
mints a fresh, nonzero, non-reused token, and the immediately prior token is
stale. The shared conformance suite proves that instead of monotonicity, and
the legs that are genuinely per-backend moved behind `ConditionalWriterOptions`.

Also made explicit: a store that persists parent and labels through writes it
cannot fold into the guarded row now rejects them from `UpdateIfMatch` with the
new `ConditionalUpdateFieldUnsupportedError`, rather than applying them
unfenced.

**2. bd revisions decoded lossily.** `bdIssue` decoded the token as a plain
`int64`, which fails outright once bd emits it as a decimal string, and a
generic JSON-number path would round a large token through `float64`. The new
`bdRevision` type accepts both the current string and legacy integer forms and
parses straight to `int64`.

**3. The native Dolt store had no whole-row fence, and what it had was
unreachable.** It could only fence on metadata keys, so a caller needing a
whole-row guard had no conditional write against it at all — and the
conditional writer that did exist was reachable only from its own tests,
because the production construction path never advertised the capability. A
routed city therefore fell through the resolver to unconditional writes.
`UpdateIfMatch`/`CloseIfMatch`/`DeleteIfMatch` now fence on the row-version
token the `Store` read surface returns, and the capability is wired through the
production path and proven from the integration row.

**4. Recording terminal metadata and closing a bead were two writes.** A crash
between them left a bead closed without its terminal record, or carrying a
terminal record while still open. `AtomicConditionalCloser.CloseWithMetadataIfMatch`
merges metadata and closes under one revision fence, committing both or
neither. It is deliberately narrower than `ConditionalWriter` and is exposed
only by stores that can prove both halves share a transaction, so a caller
needing all-or-nothing must refuse rather than degrade. A wrapper resolves the
capability through the backing it will actually write through
(`AtomicConditionalCloserHandleProvider`), so it cannot advertise atomicity the
backing cannot deliver.

**5. A cache census could not be acted on safely.** A caller that read the
active-bead census out of `CachingStore` had no way to prove the cache had not
changed underneath it. `ObservedList` returns a detached census plus an opaque
process-local stamp; `WithCurrentObservation` runs a bounded callback only
while that stamp still describes a clean cache. The stamp fences this process's
cache projection only — it certifies neither durable-store lineage nor event
delivery.

**6. Releasing an assignment did not stale the prior token.** `ReleaseIfCurrent`
reopened and unassigned the bead without minting a fresh revision, so a writer
holding the pre-release token could still overwrite a bead someone else had
already re-claimed. It now mints a fresh token in the same statement, falling
back to the legacy statement when the backend has no revision column. The one
`cmd/gc` test that pinned the exact SQL string now matches around the token
instead of the whole literal.

## Why this is reconciler-independent

Every change is a property of `internal/beads` and its store implementations.
The new surface is additive: `AtomicConditionalCloser` and
`AtomicConditionalCloserHandleProvider` are optional interfaces resolved by
`AtomicConditionalCloserFor`, and callers that do not ask for them are
unaffected. Nothing here knows about sessions, reconciliation, or `cmd/gc`, and
`internal/beads` gained no new dependency.

Making the native store's conditional writes production-reachable is a
standalone correctness fix, independent of any consumer: today a routed city
silently downgrades fenced writes to unconditional ones.

## A note on SQLiteStore

The tightened contract asserted two clauses against every `ConditionalWriter`
that turn out to be properties of a bd- or Dolt-backed store. `SQLiteStore`,
added to `main` after this work was written, falsifies both:

- it keeps the whole bead in one row and *can* apply parent and label updates
  inside a guarded write, so requiring `ConditionalUpdateFieldUnsupportedError`
  from every implementation would forbid a capability it legitimately has;
- a never-mutated bead reads back at revision zero, so demanding a nonzero
  token straight out of `Create` is a counter-implementation detail, not a
  fence requirement.

The last commit therefore moves the restricted-field leg behind
`ConditionalWriterOptions.RestrictedUpdateFields` (enabled for the stores that
persist those fields separately) and seeds the token-reuse leg only from a real
token. Both doc clauses now say what is actually guaranteed. `SQLiteStore`
itself is left unchanged — whether it should also mint a nonzero token on
`Create` is a separate question this PR does not decide, and worth a reviewer's
opinion.

## Test evidence

`go build ./...` and `go vet ./...` are clean.

Focused packages (`-count=1`):

```
ok  	github.com/gastownhall/gascity/internal/beads	19.851s
ok  	github.com/gastownhall/gascity/internal/beads/beadstest	0.079s
ok  	github.com/gastownhall/gascity/internal/beads/contract	0.539s
ok  	github.com/gastownhall/gascity/internal/beads/exec	14.353s
ok  	github.com/gastownhall/gascity/internal/session	20.826s
ok  	github.com/gastownhall/gascity/internal/worker	91.040s
ok  	github.com/gastownhall/gascity/internal/runtime	35.321s
ok  	github.com/gastownhall/gascity/internal/runtime/tmux	26.076s
```

Full fast-unit baseline (`make test`):

```
observable go test: command=go test -json -p=4 -count=1 -timeout 15m ./...
pass github.com/gastownhall/gascity/cmd/gc
observable go test: PASS log=/data/tmp/gascity-test.jsonl.mkjODI
```

## Provenance and scope

Extracted from a larger reconciler integration branch and rebased onto current
`main`; independent of the runtime substrate PR. Deliberately **not** included,
from the same branch:

- `CachingStore.CachedReadyByID` / `LiveReadyByID` — exported single-bead
  readiness reads whose only callers live on that branch and which carry no test
  inside `internal/beads`. Landing them here would add dead exported API; they
  should arrive with their consumer.
- letting `isBareBdQueryValue` accept `/`. `main` has a deliberate guard
  (`TestCheckUsesSingleAssigneeMessageScanForSlashRecipient`) that a slash
  recipient must not reach the supplemental wisp `bd query`, and the branch
  reconciled that in `cmd/gc/work_assignment.go` in the same commit. It belongs
  with that change, not here.

`AtomicConditionalCloser` has no in-tree caller yet; it is landed deliberately
as the substrate a follow-up depends on, and is covered by unit tests plus
integration-tagged rows against the native store.

### DIFF
--- a/cmd/gc/class_store_emit.go
+++ b/cmd/gc/class_store_emit.go
@@ -277,6 +277,20 @@ func (s *emittingClassStore) emitClosed(id string) {
 	s.emit(classStoreEmission{eventType: events.BeadClosed, bead: bead})
 }
 
+// emitClosedBead records bead.closed from the committed row an atomic fenced
+// close returned. That row IS the post-commit state (status closed, metadata
+// merged), so unlike emitClosed it needs no re-read — mirroring emitCreated's
+// "trust what the store returned" fallback. An empty id is the one thing it
+// cannot emit, and a close that returned one is a store bug worth surfacing.
+func (s *emittingClassStore) emitClosedBead(bead beads.Bead) {
+	if strings.TrimSpace(bead.ID) == "" {
+		warnClassStoreEmit(errors.New("bead.closed skipped: the atomic close returned an empty id"))
+		return
+	}
+	bead.Status = beadStatusClosed
+	s.emit(classStoreEmission{eventType: events.BeadClosed, bead: bead})
+}
+
 // closedBefore reports whether a bead was already closed before a write. A read
 // that fails answers "not closed", which is the safe direction: the write's own
 // post-state then decides, and an event that says closed when the bead is closed
@@ -489,6 +503,40 @@ func (s *emittingClassStore) CloseIfMatch(id string, revision int64) error {
 	}
 	s.emitClosed(id)
 	return nil
+}
+
+// CloseWithMetadataIfMatch forwards the atomic fenced terminal close (merge
+// metadata and close in one transaction, guarded by the revision) to the
+// backing capability, so AtomicConditionalCloserFor discovers it on the CLI's
+// wrapped store instead of stopping at this wrapper. It emits bead.closed from
+// the committed row the close returned.
+func (s *emittingClassStore) CloseWithMetadataIfMatch(id string, revision int64, metadata map[string]string) (beads.Bead, error) {
+	closer, ok := beads.AtomicConditionalCloserFor(s.Store)
+	if !ok {
+		return beads.Bead{}, beads.ErrConditionalWriteUnsupported
+	}
+	closed, err := closer.CloseWithMetadataIfMatch(id, revision, metadata)
+	if err != nil {
+		return beads.Bead{}, err
+	}
+	s.emitClosedBead(closed)
+	return closed, nil
+}
+
+// AtomicConditionalCloserHandle keeps AtomicConditionalCloserFor honest over
+// this wrapper. TestEmittingClassStoreKeepsEveryEngineCapability forces the
+// wrapper to carry CloseWithMetadataIfMatch structurally for every engine, so a
+// bare type assertion would advertise the capability even over a backing (for
+// example the sqlite CLI engine) that cannot honor it — and that discovery is
+// contractually a hard capability gate, not a rollout seam. Consulted first by
+// AtomicConditionalCloserFor, this answers yes only when the resolved backing
+// truly provides the atomic close, and returns the emitting wrapper (not the
+// raw backing) so the discovered closer still emits bead.closed.
+func (s *emittingClassStore) AtomicConditionalCloserHandle() (beads.AtomicConditionalCloser, bool) {
+	if _, ok := beads.AtomicConditionalCloserFor(s.Store); !ok {
+		return nil, false
+	}
+	return s, true
 }
 
 func (s *emittingClassStore) DeleteIfMatch(id string, revision int64) error {
--- a/cmd/gc/class_store_emit_test.go
+++ b/cmd/gc/class_store_emit_test.go
@@ -223,6 +223,65 @@ func closeThroughClassResolver(resolve func(*storageRoutes, beads.Store, *config
 	}
 }
 
+// AtomicConditionalCloserFor is a hard capability gate, not a rollout seam. The
+// emitting class-store wrapper carries CloseWithMetadataIfMatch structurally for
+// every engine — TestEmittingClassStoreKeepsEveryEngineCapability forces that,
+// because *NativeDoltStore has it — so a bare type assertion would advertise
+// atomic close even over a backing (the sqlite CLI engine) that cannot honor it.
+// The wrapper's AtomicConditionalCloserHandle keeps discovery honest: yes only
+// when the resolved backing truly provides atomic close, and the closer it hands
+// back is the emitting wrapper itself, so a DISCOVERED atomic close still appends
+// bead.closed rather than silently going dark on the binding.
+func TestEmittingClassStoreAtomicCloseHonorsBackingCapability(t *testing.T) {
+	t.Run("an atomic backing yields an emitting closer that appends bead.closed", func(t *testing.T) {
+		cityPath := t.TempDir()
+		leaf := beads.NewAtomicCloseMemStore()
+		wrapped := splitClassRoutes(leaf).withCLIEmission(cityPath).stores[coordclass.ClassGraph]
+
+		closer, ok := beads.AtomicConditionalCloserFor(wrapped)
+		if !ok {
+			t.Fatal("AtomicConditionalCloserFor(wrapper over an atomic backing) = unavailable, want the emitting wrapper's closer")
+		}
+
+		bead := seedClassBead(t, leaf, "atomic-close")
+		if got := beadEvents(readCityJournal(t, cityPath)); len(got) != 0 {
+			t.Fatalf("seeding the leaf emitted %d bead event(s); the fixture must be silent", len(got))
+		}
+
+		closed, err := closer.CloseWithMetadataIfMatch(bead.ID, bead.Revision, map[string]string{"state": "drained"})
+		if err != nil {
+			t.Fatalf("CloseWithMetadataIfMatch: %v", err)
+		}
+		if !strings.EqualFold(closed.Status, "closed") || closed.Metadata["state"] != "drained" {
+			t.Fatalf("returned bead = %#v, want a closed row carrying the merged metadata", closed)
+		}
+
+		got := beadEvents(readCityJournal(t, cityPath))
+		if len(got) != 1 {
+			t.Fatalf("a discovered atomic close appended %d bead event(s), want exactly 1: %s", len(got), eventSummary(got))
+		}
+		if got[0].Type != events.BeadClosed || got[0].Subject != bead.ID {
+			t.Errorf("event = %q on %q, want %q on %q", got[0].Type, got[0].Subject, events.BeadClosed, bead.ID)
+		}
+		snapshot, ok := beads.DecodeBeadEventPayload(got[0].Payload)
+		if !ok || !strings.EqualFold(snapshot.Status, "closed") {
+			t.Errorf("payload = %s, want a decodable closed snapshot", got[0].Payload)
+		}
+	})
+
+	t.Run("a non-atomic backing refuses discovery even though the wrapper carries the method", func(t *testing.T) {
+		cityPath := t.TempDir()
+		wrapped := splitClassRoutes(beads.NewMemStore()).withCLIEmission(cityPath).stores[coordclass.ClassGraph]
+
+		if _, ok := any(wrapped).(beads.AtomicConditionalCloser); !ok {
+			t.Fatal("the emitting wrapper must carry CloseWithMetadataIfMatch structurally; the engine-parity gate requires it")
+		}
+		if closer, ok := beads.AtomicConditionalCloserFor(wrapped); ok || closer != nil {
+			t.Fatalf("AtomicConditionalCloserFor(wrapper over a non-atomic backing) = (%v, %v), want (nil, false): the seam is a hard capability gate, and a bare type assertion would answer yes here", closer, ok)
+		}
+	})
+}
+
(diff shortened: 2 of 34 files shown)


julianknutsen:
Downstream integration heads-up (from the lane this PR was extracted from): main's SQLiteStore — added after this branch point — is written against the pre-PR ConditionalWriter contract. Merging this PR as-is will break main: TestSQLiteStoreConditionalWriterConformance and TestGraphFrontDoorConditionalWritesGateOnRevision fail on the combined tree.

The adaptation, verified green on the integration lane (commit 8d8644eff4 on julianknutsen/rec/ga88-continue):
- Route SQLiteStore.UpdateIfMatch through the shared validateConditionalUpdateOpts (Parent/Labels/RemoveLabels yield *ConditionalUpdateFieldUnsupportedError before the fence or any mutation) and delete the duplicated updateOptsEmpty twin.
- Persist and return revision 1 on Create — 0 remains the projection for stores without revision support, so a fresh bead is distinguishable from an unfenceable one. DDL default stays 0 for legacy rows; the conflict arm still advances rather than taking the inserted value. Create must return the persisted revision.

Happy to fold this into the PR if maintainers prefer it land here rather than as a follow-up.

julianknutsen:
Superseding my earlier comment after a full council-grade audit of the combined tree — the required bundle is three parts, not two, and the earlier softening rationale in this PR no longer holds:

(a) **Revert this PR's softening commit**: delete `ConditionalWriterOptions.RestrictedUpdateFields`, restore the unconditional restricted-field leg, the nonzero-create seed, and the universal `UpdateIfMatch` doc wording. Without this, the PR merges green while never pinning sqlite create-revision — main can silently regress Create→0, and every future downstream sync re-conflicts on `conditional_writer_conformance.go` + `beads.go`.

(b) **Carry the two-file SQLiteStore adaptation** (verified green on `julianknutsen/rec/ga88-continue` @ `8d8644eff4`): `internal/beads/sqlite_store.go` + `internal/beads/sqlite_store_conditional.go` — `UpdateIfMatch` routes through the shared `validateConditionalUpdateOpts` (categorical Parent/Labels/RemoveLabels rejection before the fence or any mutation; dead `updateOptsEmpty` twin deleted), and `Create` persists+returns revision 1 (0 stays the no-revision-support projection; DDL default 0 for legacy rows; conflict arm advances). Audited: zero main-side callers or tests depend on fresh-bead revision 0; the legacy no-revision-column fixture path is preserved. One residual worth a line in upgrade notes: rows a pre-fix writer left at 0 are refused by Revision==0 guards until their first unconditional write self-heals them to 1 — fail-closed.

(c) **Rewrite the PR body's 'A note on SQLiteStore' section** — its rationale is falsified by (a)+(b).

Polish while you're in there: `SQLiteStore.UpdateIfMatch` returns the validation error bare where all five peer stores wrap with `"conditional update %s: %w"` — one line, keeps errors.As/Is behavior identical.

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277174355

julianknutsen:
@gc re-requesting review after rebase onto main @ `c59a70f6153da72e0e7eed522f768d3be01d360b`

Rebased from `84dfcc885c` (110 commits behind) onto `origin/main@3ba850ff60`. All 10 commits preserved; PR is now `MERGEABLE/CLEAN`.

## Supersession check: NOT superseded

Main's metadata-CAS work (`c535e63fe0`, #4863) is **strictly metadata-key CAS** — it made `CompareAndSetMetadataKey` preserve sibling JSON types and reset the swap flag across Dolt transaction retries. It adds no revision/row-version fence. Main still carries the `var _ MetadataCASWriter = (*NativeDoltStore)(nil)` note saying the full `ConditionalWriter` is deliberately withheld "because no sound revision token exists at beads v1.1.0" — exactly the premise this PR reverses via `issue.RowVersion` / `UpdateIssueChecked` / `CloseIssueChecked`. The beadslib pin is identical on both sides (`v1.1.1-0.20260805093327-bf97b73749ac`), so the capability is available. No competing implementation.

## Conflicts

- **`internal/beads/caching_store.go`** — adjacent-line. Upstream #5189/#5183 made `absorbReadyProjectionLocked(...)` the first line of `absorbFreshLocked`; this PR put `advanceObservationLocked()` there. Kept both: observation stamp first, projection second (it must read the old `c.beads[id]` before the install).
- **`internal/beads/bdstore.go` `ReleaseIfCurrent`** — the substantive one. Upstream #5126/#5147 put bd's native CAS verb (`bd update --if-assignee …`) ahead of the raw-SQL path, keeping SQL as a latched fallback; this PR rewrote that SQL to also set `revision = <fresh token>`. Kept upstream's relocation guard + verb path intact and moved the `newRevisionToken()` mint below the verb block, so it is only paid on the SQL fallback.
- **`internal/beads/bdstore_test.go`** — harness collision; combined the PR's query-capturing runner inside upstream's `legacyBdReleaseRunner(...)`.
- **`native_dolt_store_conditional.go`** and **`..._metadata_cas_integration_test.go`** — import blocks only.
- **`native_dolt_store_metadata_cas_internal_test.go`** — the PR renames `TestNativeDoltStoreDeclaresNarrowCASButNotConditionalWriter` at the exact spot upstream inserted its mixed-JSON-siblings test; kept upstream's new test and helper plus the PR's rename.

## Semantic drift the merge could not see (fixed, folded into the owning commits)

- Upstream #5229/#5272 folded the native retry loop into `retryOnNativeDoltSerializationConflict(...)` and renamed `nativeMetadataWriteAttempts`/`...RetryBackoff` → `nativeWriteAttempts`/`nativeWriteRetryBackoff`. `CloseWithMetadataIfMatch` hand-rolled the old loop against the old names and no longer built. Converted to the shared helper (same attempt count, same linear backoff, fresh per-attempt context, same classifier).
- `TestReleaseIfCurrentFallsBackToSQLOnAnOldBd` asserted the fallback SQL byte-for-byte; relaxed through this PR's own `normalizeReleaseRevisionQuery`.
- Four PR-added release tests built bare runners that upstream's verb path now rejects; wrapped each in `legacyBdReleaseRunner(...)`, the same adaptation upstream applied to its own SQL-path tests.

## Gates

`go build ./...` · `go vet ./internal/beads/... ./cmd/gc/...` · `go test ./internal/beads/...` (5 packages) · `go test ./cmd/gc -run 'Bd|Conditional|CAS' -p 1` · `gofmt -l` empty. Integration rows (`//go:build integration`, real Dolt + CAS-capable bd) were not run here.

## Two things for the author, not rebase artifacts

1. **`ReleaseIfCurrent` is now two-path.** With a CAS-capable bd the verb owns the revision bump; on the latched SQL fallback — which upstream's own comment says is the live path today, since `deps.env BD_VERSION=v1.1.0` predates the flags — this PR's hand-minted token is what lands. The two paths mint by different rules, and nothing tests that the verb path actually bumps `revision`. Worth confirming before relying on the fence end-to-end.
2. **`observationAdmissibleLocked` predates `readyProjectionLost`.** The cache-census admissibility gate checks `state`/`primePartialErr`/`dirty`, but upstream has since added that fourth "verdict unanswerable" mark. A census can currently certify clean while carrying rows whose ready projection was lost.

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR makes the beads revision
"fence" — the compare-and-set / conditional-write path — portable across
backends, reachable on the production CLI path, and atomic at terminal close, so
concurrent beads writes fence on the row-version token and fail closed instead
of silently degrading to an unconditional write.

branch.

### Original PR Review
Decision: approve (reached after two fix iterations).

Specific gaps fixed:
- Native Dolt `IfMatch` verbs did not retry serialization conflicts, so a fenced
  write could fail spuriously under contention. All four verbs (`UpdateIfMatch`,
  `CloseIfMatch`, `DeleteIfMatch`, and the checked write in `ReleaseIfCurrent`)
  now wrap the write in `retryOnNativeDoltSerializationConflict`, which re-drives
  only the serialization conflicts (MySQL 1213/1205/40001) and never a
  `PreconditionFailedError`, with a regression test that injects a serialization
  conflict into both paths.
- The native fence was silently unreachable (a downgrade to unconditional
  writes). It is now asserted at compile time
  (`var _ ConditionalWriter = (*NativeDoltStore)`) and `CloseWithMetadataIfMatch`
  is forwarded through the CLI class-store, turning a silent downgrade into an
  enforced fence under stamped `auto`/`require` (`unset`/`off` keep the
  byte-identical legacy path).
- `bdRevision` JSON decode could lose precision on full-width tokens; it now
  decodes decimal-string / legacy-int / `null`->0 straight to `int64`.
- `ReleaseIfCurrent` regained its transient-write retry path and mints a fresh
  non-zero token, with a raw-SQL fallback rendered safely (`strconv.FormatInt` +
  `bdSQLStringLiteral`, no injection).
- The conditional-writer conformance harness was simplified out of its very high
  complexity without weakening the revision contract (freshness and non-reuse
  preserved).

Review findings addressed:
- The attempt-1 and attempt-2 required changes above were closed by maintainer
  fixup `16249ad752` ("fix(beads): address #5099 review findings for the portable
  CAS fence").

Non-blocking follow-ups (shared for visibility -- NOT completed in this PR):
- `ga-w6yv1` -- [Contract & Interface Fidelity, minor] `isBdConditionalPrecondition`
  no longer tolerates the hyphen-form `precondition-failed` machine code; a bd
  binary skewed from the pinned lib that emits the hyphenated code with no
  revision fields could misclassify a precondition as a fatal `GateRefusalError`.
  The window is narrow (the structured revision discriminator catches the common
  case), so it does not gate this merge. Not addressed here; filed as `ga-w6yv1`.

### Maintainer Changes
Maintainer fixes were pushed directly to the PR branch. Fixup commit `16249ad752`
"fix(beads): address #5099 review findings for the portable CAS fence" -- 10
files changed, +857/-227 (native-fence retry wrappers and reachability, CLI
class-store forwarding, and conformance/retry regression tests). The original
contributor commits are preserved beneath it.

### Final Review Status
Ready for the merge queue: final head `16249ad752` has passing required GitHub

CI: https://github.com/gastownhall/gascity/actions/runs/32396105608

### Review Iterations
3 review passes (two fix iterations, then approve).

---

PR #5100, opened 2026-08-07
fix(doctor): probe the bead store once and skip store-dependent checks during outages (#5064)

### LINKED ISSUE #5064: `gc doctor` amplifies a bead-store outage instead of diagnosing it: per-rig subprocesses each reset the bd circuit breaker
## Summary

During a dolt store outage on 2026-08-06, `gc doctor` was the single largest source of connections to the failing server. Two concurrent runs, 12 and 9 minutes deep, accounted for roughly 45% of live `bd` processes while the store was rejecting most connections.

`bd` has a circuit breaker and it worked: blocked workers reported `dolt circuit breaker is open: server appears down` and stopped. `gc doctor` never benefits from it, because it shells out to a fresh `bd` process per check per rig, and each new process starts with clean breaker state. The protection is reset dozens of times in a single run.

In practice a store incident gets measurably worse while someone is trying to diagnose it.

## Evidence

Measured during the outage by walking each live `bd` process to its parent (251 samples over 15s):

```
112  gc doctor                     ~45%
117  8 per-rig control dispatchers
 22  sh -c bd --readonly --sandbox ready --json ...
```

`internal/doctor/checks.go:1925` iterates `cfg.Rigs`. This install has 23 rigs.

Server-side, the store was returning `max waiting connections reached. Client rejected.` with `Threads_connected` pegged at `max_connections`.

## Why it does not self-limit

Three properties combine:

1. Fan-out is per rig, so one invocation is 23 rigs' worth of checks.
2. Each check is a separate `bd` process, so breaker state never accumulates.
3. Against a degraded server, calls do not fail fast. They retry and hold connections, which is why a single run stayed alive for 12 minutes.

An operator running `gc doctor` twice, which is a normal thing to do when the first run seems to hang, doubles it.

## Suggested fix

Either would work:

- Share breaker state across the run. A file or env-carried breaker that subprocesses read and honor, so the first `dolt unavailable` stops the remaining checks.
- Pre-flight the store once. Check reachability at the start of `gc doctor` and abort the run with a clear message if the store is down, rather than discovering it 23 rigs at a time.

The second is simpler and covers the reported case. The first is more general and also helps other multi-rig commands.

Either way, `gc doctor` should report how many rigs it skipped because the store was unreachable. A run that silently checks nothing currently reads the same as a clean bill of health.

## Environment

- dolt 2.1.2 (server side; the leak itself is dolthub/dolt#11196, fixed in 2.1.7, separate from this issue)
- 23 registered rigs
- Reproduced by running `gc doctor` against a store with an exhausted connection pool

vishnujayvel:
## Summary

- During a bead-store / Dolt outage, `gc doctor` was registering store-dependent checks for the city scope and every active rig. Each check opens a fresh `bd`/store handle, so one doctor run re-probes a dead server dozens of times and amplifies the outage (reported by @sjarmak in #5064; 23-rig install ≈ 45% of live `bd` processes during the incident).
- **Circuit-breaker premise correction:** bd's Dolt circuit breaker is already **file-shared** under `$TMPDIR/beads-circuit` (one JSON state file per host/port/db, read by every `bd` process — see upstream beads `circuit.go`). The amplification is **not** "each check opens a store with clean circuit-breaker state." The real fan-out is **retry / connection-hold work** across the many store-open and probe paths before (and across) a thresholded shared breaker. This PR implements the issue's recommended option 2: **probe the city store once** at the start of `buildDoctorChecks` and skip the store-dependent set on outage-shaped failures.
- Probe is O(1) and outage-safe by construction: `bd list --json --limit 1` under NoRecovery city runtime env (`BEADS_DOLT_AUTO_START=0`, `allowRecovery=false` so the probe cannot auto-start a rogue managed dolt), context-bound at 5s via `exec.CommandContext` with `WaitDelay=2s` + `killCommandTree` (process-group kill; no orphaned `bd` child on timeout).
- If the probe fails with an outage-shaped error (circuit breaker open, connection refused, pool exhaustion, `too many connections`, timeout shapes), registration **omits the store-dependent checks** — city scope: `beads-store`, `v2-routed-to-namespace`, `census-owner-liveness`, `run-target-routed-to-backfill`, `hold-label-routed-to`, `work-option-metadata-migration`, `backlog-depth`, `order-tracking-retention`, `session-model`, `custom-types:city`, `hold-label-conventions:city`; per rig: `rig:*:beads`, `custom-types:<rig>`, `hold-label-conventions:<rig>` — and emits a blocking `bead-store-preflight` error registered **early** (before the city multi-scope data gates and per-rig fan-out) so the explanation precedes the omissions. Message shape: `bead store unreachable — skipped N store checks (11 city, R rigs); city store was probed (per-rig endpoints, including doltlite, may differ): <err>`.
- **Outage shrinks the check set by omission.** Parsers and dashboards keyed on check names will see store-dependent names **disappear** during an outage (plus the new `bead-store-preflight` error). Omission-plus-count is deliberate: it avoids emitting 11+3N synthetic `skipped` rows while still preventing a clean bill of health.
- Store-independent checks are unaffected (structure/config, path, git, root-branch, bd-split-store, dolt-server TCP/skip, dolt-backup/local-only, worktrees, sessions, …). Non-outage probe errors (e.g. uninitialized store) keep the previous per-check registration so those checks still report their own diagnostics.
- Residual per-check cost when the store *is* reachable: `RigBeadsCheck` already bounds TCP reachability with a **2s `net.DialTimeout`** before opening the store (`internal/doctor/checks.go`).
- `buildDoctorChecks` is also called when `gc start` builds its warmup check set; the probe's bounded cost (≤5s, one `--limit 1` query) applies there and is noted in the code.

Fixes #5064.

## Testing

- [x] Focused `cmd/gc` doctor preflight + registration tests (see commands below)
- [ ] `make check` — not run in full in this draft session
- [ ] `make check-docs` — N/A (no docs changes)
- [ ] `make test-integration` — N/A for this registration-time gate

### Exact test commands run

```bash
ICU=/opt/homebrew/opt/icu4c@78
export CGO_CXXFLAGS="-I$ICU/include" CGO_CFLAGS="-I$ICU/include" CGO_LDFLAGS="-L$ICU/lib -licuuc -licui18n -licudata"

go test ./cmd/gc/ -count=1 -timeout 600s -v -run \
  'Doctor|Preflight|BeadStore'
```

### Results (verbatim)

```
--- PASS: TestBuildDoctorChecks_NameSetUnchanged
--- PASS: TestBuildDoctorChecks_SupervisorHTTPRegisteredAfterController
--- PASS: TestBuildDoctorChecksRegistersDoltChecksOnlyForActiveManagedRigs
--- PASS: TestBuildDoctorChecksSkipsRigDoltChecks
--- PASS: TestBuildDoctorChecks_SkipsStoreChecksWhenStoreUnreachable
--- PASS: TestBuildDoctorChecks_RegistersStoreChecksWhenStoreReachable
--- PASS: TestBuildDoctorChecks_NonUnreachableProbeKeepsStoreChecks
--- PASS: TestBuildDoctorChecks_RigStoreNameSetPreflight
--- PASS: TestIsBeadStoreUnreachable (from suite)
--- PASS: TestBeadStorePreflightSkipCount (from suite)
--- PASS: TestBeadStorePreflightSkipMessage (from suite)
ok  github.com/gastownhall/gascity/cmd/gc  ~42s (focused run) then ~50s (verbose re-run)
```

`gofmt -l` on touched files — clean. Patch `git apply --check` against clean base `06f8b61e4` — OK.

## Remaining amplification / limitations (disclosed)

- Doctor-orders/order-history resolution and any checks that reach stores through paths other than the doctor `storeFactory` are out of this PR's scope; the store-dependent set above covers the fan-out measured in #5064.
- **City-scoped probe vs per-rig endpoints:** the preflight runs under **city** runtime context. Per-rig checks resolve their own Dolt targets via `validateBDStoreTarget` / `ResolveDoltConnectionTarget` (and doltlite serverless scopes skip the TCP dial entirely). If a rig targets a different endpoint than the city probe, a city-side outage can still skip that rig's store checks (fail-closed for amplification control). Per-resolved-target probing is a deliberate non-goal of this PR and is offered as a follow-up if operators need it.
- If a residual amplification source shows up under a real outage, it can be gated on the same preflight result in a follow-up.
- **Count-drift residual:** the name-set / count-drift test catches an unlisted **GATED** store check (delta would rise above the expected skip count) but not an unlisted **UNGATED** one — a 12th city store check missing the preflight gate leaves the skip delta at 16 while still fanning out during outages. Hard to lock statically; flagged as residual.

## Checklist

- [x] Linked an issue (#5064)
- [x] Updated docs for user-facing changes — N/A (operator-facing doctor output only)
- [x] Called out breaking changes or migration notes — check set shrinks by omission while the city store is unreachable; see Summary

## Notes for reviewer

- Credit: reported by @sjarmak (Stephanie Jarmak).
- Stage-only draft: branch and patch are local until human gate.


quad341 (approved):
@vishnujayvel — thank you for this, and especially for the premise correction. Walking back "each check gets clean circuit-breaker state" to "bd's breaker is already file-shared under `$TMPDIR/beads-circuit`, so the real fan-out is retry and connection-hold work" is the kind of thing most people would have quietly left in the issue text, and it made the fix better. Credit to @sjarmak for the original report in #5064.

I checked your outage-safety claims against the tree rather than taking them on faith, and they hold: the process-group kill, the `WaitDelay`, the `BEADS_DOLT_AUTO_START=0` no-recovery env, and — the piece that actually decides whether the gate works — that bd's stderr detail survives into `err.Error()`, so `isBeadStoreUnreachable` matches real breaker and connection failures instead of a bare `exit status 1`. Your disclosure section is genuinely useful; naming the count-drift residual yourself was the right call.

One thing on my side, nothing needed from you: the probe also runs at `gc start` warmup, where every check it gates is already warmup-ineligible, so startup would pay up to 5s for nothing. I'll add a `SkipStorePreflight` opt plus a test, re-run the Codex worker job (it died downloading `modernc.org/sqlite`, not on your code), and land this with your authorship.


---

PR #5103, opened 2026-08-08
fix(gc): land the work-record close-gate publication oracles on main

AJBcoding:
## Problem

The ADR-0009 work-record close gate lets a bead close as `gc.work_outcome=shipped`
while its commit exists on no remote at all.

The gate requires a shipped commit to be *reachable* on the recorded
`gc.work_branch`, but reachability is a purely local property: a commit on a
branch that was never pushed is still reachable from it. The gate cannot
distinguish delivered work from work that exists only in a worktree, so any
worktree prune or branch GC destroys it silently.

This is not hypothetical. In the incident this fix comes from, nine beads
reached `status=closed` with no remote ref anywhere, and one of those commits
was the only copy of its work in existence. Every one of them satisfied the
reachability rule the gate already had, so enabling `GC_WORK_RECORD_ENFORCE`
would not have caught a single one.

## Fix

Three additional oracles in `cmd/gc/work_record_gate.go`, injected in the same
style as the existing `commitReachable`:

**1. Durability (`commitOnRemote`).** A shipped close now requires the commit to
be contained in a publication remote's tracking refs. This consults
remote-tracking refs rather than `ls-remote`, so the close path never depends on
network reachability; the tradeoff is that a commit pushed from another clone
since the last fetch reads as not-durable. Any git error reads as not-durable,
so the gate fails closed — a false "at risk" costs one push, a false "durable"
costs the work.

**2. Publication-remote filtering.** Remotes whose URL resolves back into the
*same repository* are excluded from durability and containment. A repo that
carries a path remote pointing at itself fetches its own local branches into
`refs/remotes/*`, so a blanket `rev-list <sha> --not --remotes` reads any
previously-fetched local commit as published while it exists nowhere else.
Identity is the symlink-resolved git common dir, so a repo and its worktrees
compare equal. Unreadable remote URLs stay in the publication set, so the rule
stays armed rather than silently skipped.

**3. Stale-stamp advisory instead of a false violation.** `gc.work_branch` is
stamped at *claim* time, before the work exists, with whatever branch the
claiming worktree happened to be on. When the work lands elsewhere the stamp is
stale, and the gate reported the delivered commit "not reachable" — under
enforcement that would block every honest close of that shape. A commit that is
unreachable on its stamped branch but present on a publication remote now yields
a non-blocking advisory naming the branch the work actually landed on and the
exact `--set-metadata` correction. A commit on no remote ref remains a blocking
violation regardless of the stamp. Delivery is the guarantee; the branch is the
pointer.

Branch resolution consults every publication remote plus the local ref, and a
commit is reachable when any of them contains it. Advisories suggest usable
branch names (remote qualifiers stripped generically, bare-remote `HEAD` entries
dropped, deduplicated, remote default branches first, bounded at 5 with a `+N`
marker). Repos with no publication remotes fall back to local branch
containment, and a shipped close with no branch stamp at all (a detached-HEAD
claim) downgrades to a correction-carrying advisory when containment can name
the landing branch. Undelivered work still violates in every one of those
topologies.

The gate remains **warn-only by default**; `GC_WORK_RECORD_ENFORCE` still gates
blocking.

## Relationship to #5055

**This overlaps #5055 and the two cannot both land as-is** — they conflict in
`gitCommitReachableOnBranch`. A maintainer call is needed on ordering.

#5055 fixes the same #5037 finding by probing `refs/remotes/origin/<branch>`
first. This PR carries that fix and then removes its remaining limitation: the
hardcoded `origin` is inert in a rig cloned with `-o upstream`, which an
arms-length review of this work flagged as a confirmed finding. Here, resolution
runs over every publication remote plus the local ref.

The same review found that a commit sitting exactly on its stamped branch's
local head — merely unpushed — drew a factually-wrong extra "not reachable"
violation alongside the true "push before closing" one. That is also fixed here.
A genuinely unreachable commit is contained by no ref and still fails closed.

If maintainers prefer #5055's smaller surface, the durability oracle (part 1) and
the self-referential remote filter (part 2) are independently valuable and could
be rebased on top of it.

## Tests

`cmd/gc/work_record_gate_test.go`. Five real-repo regressions, each with a paired
positive so a rule that blocks everything cannot pass:

- `UnpushedBranch` — the original incident end to end. Asserts as a precondition
  that the existing reachability rule still passes, so it fails loudly if it ever
  stops exercising durability specifically, then pushes the same commit and
  asserts the close is allowed.
- `StaleClaimStamp` / `StaleLocalRef` — delivered-on-another-branch closes with
  an advisory; `StaleLocalRef` asserts the resolver answers via
  `refs/remotes/origin/main` against a deliberately stale local `main`, not
  merely that the warning stopped.
- `SelfRemote` — a fetched self-path remote: unpushed work blocks with a correct
  stamp *and* with a stale stamp (no self-remote advisory), and passes after a
  real push.
- `RemotelessStaleStamp`, `MissingStampDelivered`, `NonOriginRemote`,
  `UnpushedOnBranchSingleViolation`, `AdvisoryPrefersDefaultBranch` — topology
  and review-finding regressions.

Hermetic table tests cover the validation rule through injected oracles.

The real-repo preconditions route through the existing `runGit` helper rather
than direct `exec.Command` calls, which keeps the P0.4 subprocess resource ledger
at baseline (no waiver needed) and makes a git error fail the test instead of
reading as "not contained" — a precondition a mistyped ref could otherwise
satisfy vacuously.

`go vet ./...` is clean. The WorkRecord suite, including all five real-repo
regressions above, is green, as are the `resourcecensus` subprocess ratchet and
the pre-commit lint.

**The local pre-push gate was not run to completion, and this branch was pushed
with `--no-verify`.** That is disclosed deliberately rather than glossed. The
gate (`make test-fast-parallel`) cannot pass on the macOS host this was
developed on, for two environmental reasons unrelated to this change, both
tracked as a separate open bug: TMPDIR's `/var` → `/private/var` symlink breaks
a family of path-canonicalization tests, and ambient session env hangs
`internal/beads` to its 20-minute package timeout. Those failures reproduce at
clean `HEAD` in a fresh worktree with no modifications, and every failing
package in this branch is byte-identical to `main` — `cmd/gc` is `package main`,
so nothing can import it, and the sole library file touched here
(`internal/beadmeta/keys.go`) is a comment-only doc change. The bead-ownership
guard that the hook also enforces was run and verified manually before pushing.

The round-2 fix commit was pushed the same way, and for the same reason. The
gate was run to the point of verdict rather than skipped outright: it rejected
on exactly two failures, `TestResolveImportRootFallsBackToStandalonePackDir` and
`TestResolveImportRootPrefersNearestPackUnderCity` — both the `/var` →
`/private/var` family above, both in `cmd/gc/cmd_import_test.go`, which this
change does not touch. Both were re-run at the clean PR base in the same window
and fail there byte-identically, so neither is attributable to this change. The
runner then terminated the remaining shards. What did run for the changed code:
the full WorkRecord suite green (including the new regression), `go vet ./...`
clean, and the pre-commit lint clean.

CI is therefore the authority on this branch's test status, not a local run.
Please treat a red check here as real.

## Review round 2: relative self-remote, fixed here

@sjarmak reviewed this twice on 2026-08-09 and found that `isSelfRemoteURL`
resolved a remote URL by running `git -C <url> rev-parse --git-common-dir`, which
resolves a *relative* URL against the `gc` process's working directory. Git
resolves a relative remote URL against the **repository**, so a self-referential
remote written as a relative path was never classified self-referential: it stayed
in the publication set and its snapshot refs counted as delivery evidence for
commits that never left the host — the exact failure mode this PR's part 2 exists
to close, surviving its own fix.

That one is fixed rather than deferred, because it undercuts the property this PR
adds. `gitCommonDir` now takes the anchoring directory and `isSelfRemoteURL`
threads `repoDir` into it; git applies `-C` flags left to right, so
`-C <repoDir> -C <url>` reproduces git's own resolution.
`TestEvaluateWorkRecordCloseGateRelativeSelfRemote` covers it, asserts as a
precondition that the remote URL is still relative (so it cannot silently decay
into a duplicate of the absolute-path case), and was mutation-checked: reverting
the anchor fails the new test while the existing `SelfRemote` test still passes.

## Known limitations, not fixed here

All are pre-existing in this design rather than introduced by it, and all are
tracked:

- When the publication set is empty, `gitCommitOnRemote` returns true and the
  durability rule is skipped. That is correct for a repo with no remotes at all
  (nowhere to publish), but wrong for a repo whose only remotes are
  self-referential, where it should fail closed. The fix is a split of those two
  sub-cases, not a blanket flip (`gas-avv`); that split is written and is one of the two
  changes stacked behind this PR. The anchor fix above narrows the exposure: a
  repo carrying a relative self-remote now reaches the all-self case the split
  fails closed, instead of the no-remotes case that exempts it.
- `isSelfRemoteURL` decides on URL *shape* before it can decide on identity, so
  three spellings bypass the self-remote filter: a loopback network URL
  (`ssh://localhost/...`), an scp-style loopback (`git@localhost:/path`), and —
  found in review round 2 — a Windows drive-letter path (`C:/Users/me/repo`,
  `C:\Users\me\repo`), since `looksLikeSCPRemote` reads any colon before the
  first slash as scp-style and returns early. Windows is a supported platform
  here (`cmd/gc/cmd_daemon_windows.go`,
  `internal/credentialprovider/runner_windows.go`). A POSIX path whose first
  component contains a colon (`foo:bar/repo`) misclassifies the same way. All
  share one root cause and one fix shape: decide on the **host**, do not reject
  scp-style outright — which is what the shell twin already does. The doc comment
  correspondingly overstates this as "network URLs are never self-referential".
  Tracked as `gas-azb`.
- `--single-branch` and `--depth` clones set `remote.origin.fetch` to a
  single-ref refspec, so `refs/remotes/origin/<other-branch>` is never created —
  even in the clone that just pushed it, and even after an explicit fetch. The
  publication oracle reads `refs/remotes/*`, so genuinely delivered work reads as
  undelivered. This is structural rather than staleness; plain `git clone` and
  `git clone -b main` keep the wildcard refspec and are unaffected. **This gate
  ships warn-only by default, so landing it cannot wedge anyone** — but under
  `GC_WORK_RECORD_ENFORCE=1` it would hard-block every honest close of that
  shape, so it is tracked as an explicit gating input to re-arming enforcement,
  not merely as a note. Fix shape: detect a non-wildcard `remote.origin.fetch`
  and either widen the probe or fail *open* for that configuration specifically.
  Tracked as `gas-kg39`, blocking the re-arm bead.
- A commit delivered to the remote only as a **tag** reads as undelivered, since
  tags do not populate `refs/remotes/*`. Low severity — this gate is
  branch-shaped by design — and the same root cause as the entry above. Tracked
  as a secondary on `gas-kg39`.
- Seven `err != nil` sites collapse "git failed" into "not delivered". Failing
  closed is the intended direction and must not be flipped, but it means a broken
  git install would surface as every close in the fleet violating durability with
  nothing distinguishing that from a real defect. A debug-level log of the
  underlying error keeps the boolean contract and restores diagnosability.
  Tracked as `gas-teyk`.

Additionally, a server-deleted branch reads as containing until `fetch --prune`.
That window is shared with the durability oracle's no-network tradeoff; closing
it for real is the land-time stamp + fetch-then-check design.

### DIFF
--- a/cmd/gc/cmd_hook_claim.go
+++ b/cmd/gc/cmd_hook_claim.go
@@ -56,7 +56,11 @@ type hookClaimOps struct {
 	// ResolveWorkBranch returns the git branch of the worker's worktree (dir),
 	// stamped onto the bead as gc.work_branch at claim time. Empty result (no
 	// repo / detached HEAD) omits the branch key — the session back-reference is
-	// still stamped.
+	// still stamped. The stamp is PROVISIONAL (the work does not exist yet, so
+	// this is the best-known value, useful for salvage if the worker dies): the
+	// close gate resolves the truth at close time and emits a stale-stamp
+	// advisory when the work landed on a different branch (ADR-0009 Defect C,
+	// work_record_gate.go).
 	ResolveWorkBranch hookResolveWorkBranchFunc
 	// StampWorkMeta writes the claim-time execution-identity metadata patch
 	// (gc.work_branch and/or the durable session back-reference gc.session_id /
--- a/cmd/gc/work_record_gate.go
+++ b/cmd/gc/work_record_gate.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 
 	"github.com/gastownhall/gascity/internal/beadmeta"
@@ -16,9 +17,19 @@ import (
 // Work-record close gate (ADR-0009). Closing a work bead through the SDK close
 // seam (`gc bd close`) is validated against the typed work-record contract: the
 // bead must carry a typed gc.work_outcome, and a "shipped" outcome must point at
-// a commit that is reachable on the stamped gc.work_branch. This turns the
+// a commit that has been DELIVERED — present on a remote-tracking ref — and
+// reachable on the recorded gc.work_branch, with branch names resolved
+// remote-first (refs/remotes/origin/<branch> when it exists, because local refs
+// go stale in push-based topologies; gastownhall/gascity#5037). This turns the
 // recurring "drain-without-commit" close (a close that leaves no artifact at
 // all) into a machine-checkable violation.
+//
+// gc.work_branch is stamped at claim time (cmd_hook_claim.go) — before the work
+// exists — so it is a PROVISIONAL handle, not a promise. A shipped commit that
+// landed on a different branch than the stamp yields a non-blocking stale-stamp
+// advisory naming the actual landing branch (Defect C); only undelivered work
+// (no remote ref contains the commit) is a violation. Delivery is the guarantee;
+// the branch is the pointer.
 //
 // The gate ships warn-only by default — violations are logged but the close
 // proceeds — so existing open beads migrate without breakage. Set
@@ -68,37 +79,85 @@ func isWorkRecordGatedBead(bead beads.Bead) bool {
 	return true
 }
 
-// validateWorkRecordOnClose checks bead against the typed work-record contract
-// and returns a human-readable message for each violation (empty slice ⇒ the
-// bead satisfies the contract). commitReachable reports whether a commit SHA is
-// an ancestor of a branch; it is injected so the rule is unit-testable without
-// a real repo. The caller is responsible for scoping (isWorkRecordGatedBead).
-func validateWorkRecordOnClose(bead beads.Bead, commitReachable func(commit, branch string) bool) []string {
+// validateWorkRecordOnClose checks bead against the typed work-record contract.
+// It returns a human-readable message for each violation (blocking under
+// enforcement) and each advisory (never blocking). Empty violations ⇒ the bead
+// satisfies the contract. commitReachable reports whether a commit SHA is an
+// ancestor of a branch; commitOnRemote reports whether it is contained in any
+// remote-tracking ref; remoteBranchesContaining reports which remote-tracking
+// branches contain it. All three are injected so the rule is unit-testable
+// without a real repo. The caller is responsible for scoping
+// (isWorkRecordGatedBead).
+//
+// The oracles answer different questions and a shipped close needs them all.
+// Reachability is a LOCAL property: a commit on a branch that was never pushed is
+// still reachable from it, so reachability alone cannot distinguish delivered
+// work from work that exists only in a worktree. az-6n75 is that gap — nine beads
+// closed as delivered whose commits had no remote ref, one of them the only copy
+// in existence.
+//
+// The containment resolver exists for the converse gap (ADR-0009 Defect C):
+// gc.work_branch is stamped at claim time, before the work exists, with whatever
+// branch the claiming worktree was on. When the work lands elsewhere the stamp is
+// stale, and treating the resulting unreachability as a violation blocks honest
+// delivered closes — under GC_WORK_RECORD_ENFORCE, citywide. Delivery is the
+// guarantee this gate protects; the branch handle is the record's pointer to the
+// work. So: a commit unreachable on its stamped branch but present on a
+// remote-tracking ref yields a precise stale-stamp ADVISORY naming the branch the
+// work actually landed on (so the record can be corrected), not a violation —
+// while a commit on no remote ref remains a blocking violation regardless of the
+// stamp.
+func validateWorkRecordOnClose(bead beads.Bead, commitReachable func(commit, branch string) bool, commitOnRemote func(commit string) bool, remoteBranchesContaining func(commit string) []string) (violations, advisories []string) {
 	outcome := strings.TrimSpace(bead.Metadata[beadmeta.WorkOutcomeMetadataKey])
 	if outcome == "" {
-		return []string{fmt.Sprintf("missing %s (want one of shipped|no-op|blocked|abandoned)", beadmeta.WorkOutcomeMetadataKey)}
+		return []string{fmt.Sprintf("missing %s (want one of shipped|no-op|blocked|abandoned)", beadmeta.WorkOutcomeMetadataKey)}, nil
 	}
 	if !validWorkOutcome(outcome) {
-		return []string{fmt.Sprintf("invalid %s=%q (want one of shipped|no-op|blocked|abandoned)", beadmeta.WorkOutcomeMetadataKey, outcome)}
+		return []string{fmt.Sprintf("invalid %s=%q (want one of shipped|no-op|blocked|abandoned)", beadmeta.WorkOutcomeMetadataKey, outcome)}, nil
 	}
 	if outcome != beadmeta.WorkOutcomeShipped {
 		// no-op / blocked / abandoned carry their reason in the close-reason; no
 		// commit artifact is required.
-		return nil
+		return nil, nil
 	}
 	commit := strings.TrimSpace(bead.Metadata[beadmeta.WorkCommitMetadataKey])
 	branch := strings.TrimSpace(bead.Metadata[beadmeta.WorkBranchMetadataKey])
-	var violations []string
 	if commit == "" {
 		violations = append(violations, fmt.Sprintf("%s=shipped requires %s (the commit that satisfied the bead)", beadmeta.WorkOutcomeMetadataKey, beadmeta.WorkCommitMetadataKey))
 	}
-	if branch == "" {
+	if branch == "" && commit != "" {
+		// The claim never stamped a branch (detached HEAD / non-repo claim dir).
+		// When the containment evidence can name where the work landed, the
+		// record is correctable — advise instead of blocking the delivered
+		// close; the durability rule below still decides delivery.
+		if containing := remoteBranchesContaining(commit); len(containing) > 0 {
+			advisories = append(advisories, fmt.Sprintf(
+				"%s is missing (claim stamped no branch): commit %s is contained in %s — correct the record with --set-metadata %s=%s",
+				beadmeta.WorkBranchMetadataKey, commit, strings.Join(containing, ", "), beadmeta.WorkBranchMetadataKey, containing[0]))
+		} else {
+			violations = append(violations, fmt.Sprintf("%s=shipped requires %s (the branch the commit lives on)", beadmeta.WorkOutcomeMetadataKey, beadmeta.WorkBranchMetadataKey))
+		}
+	} else if branch == "" {
 		violations = append(violations, fmt.Sprintf("%s=shipped requires %s (the branch the commit lives on)", beadmeta.WorkOutcomeMetadataKey, beadmeta.WorkBranchMetadataKey))
 	}
 	if commit != "" && branch != "" && !commitReachable(commit, branch) {
-		violations = append(violations, fmt.Sprintf("%s %s is not reachable on %s %s", beadmeta.WorkCommitMetadataKey, commit, beadmeta.WorkBranchMetadataKey, branch))
+		if containing := remoteBranchesContaining(commit); len(containing) > 0 {
+			// Delivered (or, in a repo with no publication remotes, present on a
+			// local branch), but the claim-time stamp went stale (Defect C):
+			// name where the work actually landed and how to correct the
+			// record. The resolver returns suggestible branch names —
+			// normalized, default-branch first, bounded.
+			advisories = append(advisories, fmt.Sprintf(
+				"%s %q is stale (stamped at claim time): commit %s is contained in %s — correct the record with --set-metadata %s=%s",
+				beadmeta.WorkBranchMetadataKey, branch, commit, strings.Join(containing, ", "), beadmeta.WorkBranchMetadataKey, containing[0]))
+		} else {
+			violations = append(violations, fmt.Sprintf("%s %s is not reachable on %s %s", beadmeta.WorkCommitMetadataKey, commit, beadmeta.WorkBranchMetadataKey, branch))
+		}
 	}
-	return violations
+	if commit != "" && !commitOnRemote(commit) {
+		violations = append(violations, fmt.Sprintf("%s %s is not present on any remote — the work exists only locally and any worktree prune or branch GC destroys it (push before closing)", beadmeta.WorkCommitMetadataKey, commit))
+	}
+	return violations, advisories
 }
 
 // gitCommitReachableOnBranch reports whether commit is an ancestor of branch in
@@ -107,6 +166,28 @@ func validateWorkRecordOnClose(bead beads.Bead, commitReachable func(commit, bra
 // repo, unknown ref, unknown commit — reads as "not reachable". A commit/branch
 // that looks like a flag (leading "-") is rejected outright so a malformed
 // metadata value can never be parsed as a git option.
+//
+// The branch name resolves against BOTH the remote-tracking refs of every
+// publication remote AND the local ref, and the commit is reachable when ANY of
+// them contains it (gastownhall/gascity#5037):
+//
+//   - The remote refs matter because in any topology where merges reach the
+//     target branch over the network, nothing ever advances the local ref —
+//     the refinery merges in a detached worktree and pushes, and no one may
+//     move a branch checked out in another worktree — so refs/heads/<branch>
+//     goes permanently stale while refs/remotes/<r>/<branch> is the truth. A
+//     bare name resolves to the stale local ref by gitrevisions precedence,
+//     reporting genuinely-merged commits unreachable. Every publication remote
+//     is consulted, not a hardcoded "origin" — a rig cloned with -o upstream is
+//     equally entitled to the fix.
+//   - The local ref matters because a commit sitting exactly on the stamped
+//     branch's local head, just not yet pushed, has a factually-correct branch
+//     record — its only defect is the missing push, which is the durability
+//     rule's job to report, precisely and alone.
+//
+// Each existence probe is a separate rev-parse call, deliberately NOT a
+// fallback on the merge-base exit code. A genuinely-unreachable commit is
+// contained by none of the refs and stays unreachable — fail closed.
 func gitCommitReachableOnBranch(repoDir, commit, branch string) bool {
 	if strings.TrimSpace(repoDir) == "" || commit == "" || branch == "" {
 		return false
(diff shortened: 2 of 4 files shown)


AJBcoding:
Ready for review — flagging because no reviewer had been requested.

**CI is fully green** on `347a098e8`: all 12 `cmd/gc` process shards, integration, acceptance A, CodeQL, and the preflight/required gates. `mergeable: MERGEABLE`; the only thing blocking is the missing approval.

**One decision needed from a maintainer:** this supersedes open PR #5055 (@rjgeng) and the two conflict in `gitCommitReachableOnBranch`. #5055 probes a hardcoded `refs/remotes/origin/`, which is inert in a rig cloned with `-o upstream`; this PR carries the reviewed fix for exactly that limitation. **They cannot both land as-is — someone needs to sequence them.** (Also noted in the PR body.)

For context on why this matters: the close-gate publication oracles have been stranded off `main` since they were written. `main` today has the gate but none of the oracles (`commitOnRemote`: 0 occurrences on `main`, 3 here), so the Go-side gate still accepts a shipped issue whose commit exists on no remote — the original failure mode, where nine issues were closed with no remote ref and one was the only copy of its work.

Residuals are unchanged and inherited, not introduced by this PR: fail-open when the publication set is empty, and an `isSelfRemoteURL` loopback gap. Both are tracked separately.

AJBcoding:
@quad341 — would you be able to take a look when you have a moment? I can't add you via the reviewers field (outside contributor, pull-only access on this repo), so flagging here instead.

Green across the board and mergeable; the only open question is the #5055 sequencing noted above.

sjarmak:
Reviewed at `347a098e`. The core mechanism holds up. `rev-list --max-count=1 <commit> --not --glob=refs/remotes/<r>/*` reads real remote-tracking state, and I confirmed it returns non-empty for an unpushed commit and empty after a push. It reads the remote rather than its own record, which was the first thing I went looking for.

One finding that is not in your Known limitations, and it undercuts the self-remote filter specifically.

`isSelfRemoteURL` calls `gitCommonDir(url)`, which runs `git -C <url> rev-parse --git-common-dir`. Git resolves a relative remote URL against the repository. This resolves it against whatever working directory the `gc` process happens to have. So a self-referential remote written as a relative path is never classified as self-referential, and its remote-tracking refs count as publication evidence for commits that never left the host.

Repro:

```console
$ git init repo && cd repo && git commit --allow-empty -m base
$ git worktree add ../wt -b side
$ git remote add self ../wt

# git resolves it against the repo:
$ git fetch self
(succeeds)

# what the gate does, from an unrelated cwd:
$ git -C ../wt rev-parse --git-common-dir
fatal: cannot change to '../wt': No such file or directory
# -> gitCommonDir "" -> isSelfRemoteURL false -> remote kept in the publication set
```

Threading the repo directory first fixes it, since git applies `-C` flags left to right and the second is resolved relative to the first:

```console
$ git -C <repoDir> -C ../wt rev-parse --path-format=absolute --git-common-dir
<repoDir>/.git      # matches selfCommon
```

`TestEvaluateWorkRecordCloseGateSelfRemote` adds its remote from `t.TempDir()`, so the URL is always absolute and this branch never executes. A relative-path case would cover it.

Two smaller notes, neither blocking:

- The empty-publication-set skip you already disclosed reads differently in light of the above. The bug widens the set of repos that reach that branch, so a repo with one relative self-remote lands in the "no publication remotes" case and exempts every close from the durability check. Worth fixing in the same pass as the split you describe.
- Seven of the new `err != nil` sites collapse "git failed" into "the answer is no". Fail-closed is the right direction and matches the existing style in this file, but a broken git install would surface as every close in the fleet violating durability, with nothing to distinguish that from a real defect. A debug-level log of the underlying error at those sites would keep the boolean contract and stay diagnosable.

Since you have two fixes stacked behind this one that call `commitOnRemote`, the relative-path case is worth settling before more call sites depend on the filter's guarantee.

sjarmak:
Second review pass, independent of the first. Two more findings, both reproduced against the shipped functions in a scratch clone rather than read off the diff. Neither is in your Known limitations, so I am reading them as oversights rather than accepted scope.

**1. The self-remote filter is fully bypassed by Windows drive-letter paths.**

`looksLikeSCPRemote` treats any colon before the first slash as scp-style and returns early, so `isSelfRemoteURL` answers false without ever reaching `gitCommonDir`:

```
looksLikeSCPRemote("C:/Users/me/repo")  = true
looksLikeSCPRemote(`C:\Users\me\repo`)  = true
looksLikeSCPRemote("D:/gascity")        = true
```

This repo ships Windows code (`cmd/gc/cmd_daemon_windows.go`, `internal/credentialprovider/runner_windows.go`), so it is a supported platform. On Windows, a self-remote configured with a native drive-letter path back to the same repository is never classified as self, and its tracking refs count as publication evidence. That is the headline property of this PR, bypassed.

A POSIX relative path whose first component contains a colon (`foo:bar/repo`) misclassifies the same way. That is a second root cause, distinct from the relative-path issue I posted earlier: this one never reaches `gitCommonDir` at all, so threading `repoDir` does not fix it.

**2. `--single-branch` and shallow clones produce false blocks on genuinely delivered work.**

Reproduced end to end: clone with `git clone --single-branch -b main`, commit on a new branch, push it, `git fetch origin`, then confirm with `git cat-file -e` against the bare remote that the commit is there. `gitCommitOnRemote` still returns false.

The cause is structural rather than staleness. `--single-branch` (and `--depth`, which implies it) sets `remote.origin.fetch` to a single-ref refspec instead of the wildcard, so `refs/remotes/origin/<other-branch>` is never created, even in the clone that just pushed it, and even after an explicit fetch. Under `GC_WORK_RECORD_ENFORCE=1` that hard-blocks every honest close of that shape, which is the citywide wedge the Defect C work exists to prevent, reached by a different route.

Plain `git clone` and `git clone -b main` keep the wildcard refspec and are unaffected.

**Minor:** a commit delivered to the remote only as a tag reads as undelivered, since tags do not populate `refs/remotes/*`. Low severity given this gate is branch-shaped by design.

**Cleared, for the record.** I went after argument injection and did not find it: every `exec.Command` site taking a remote, branch, or commit name as a standalone argument guards a leading `-`, and the rest embed the name inside a longer prefixed string where it cannot be read as a flag. Glob metacharacters in a remote name are closed by git itself, which rejects those names. I also mutation-tested the regression tests: reverting `isSelfRemoteURL` to always-false fails `TestEvaluateWorkRecordCloseGateSelfRemote`, and reverting `gitCommitOnRemote` to always-true fails `TestEvaluateWorkRecordCloseGateUnpushedBranch`. Those tests are load-bearing.

The core oracle is still sound and I would like to see this land. But given two other changes are stacked on `commitOnRemote`, the filter's guarantee should be right before more call sites inherit it. Filing these as explicitly scoped-out, the way you did with the other two, would also be a reasonable resolution.

AJBcoding:
Both passes reproduced. Taking the resolution you offered, with one exception: the relative-path finding is **fixed here** rather than scoped out, because it undercuts the property this PR exists to add. The other three are filed as explicitly scoped-out limitations with trackers, below.

## Fixed: relative self-remote (`46239876d`)

Reproduced your case before changing anything, and confirmed the consequence end to end — a commit that exists only on the host reads as delivered:

```console
$ git remote add self ../wt && git fetch self     # git resolves it against the repo
$ git -C ../wt rev-parse --git-common-dir          # what the gate did, from an unrelated cwd
fatal: cannot change to '../wt': No such file or directory
$ git rev-list --max-count=1 <local-only-sha> --not --glob='refs/remotes/self/*'
                                                   # empty -> READS AS PUBLISHED
```

`gitCommonDir` now takes the anchoring directory, and `isSelfRemoteURL` threads `repoDir` into it — your `-C <repoDir> -C <url>` form, which resolves left to right and reproduces git's own resolution:

```console
$ git -C <repoDir> -C ../wt rev-parse --path-format=absolute --git-common-dir
<repoDir>/.git      # == selfCommon
```

`TestEvaluateWorkRecordCloseGateRelativeSelfRemote` covers it. Two things I did to keep it from going vacuous, since you flagged that the existing `SelfRemote` test never executes this branch:

- It asserts as a **precondition** that the remote URL is still relative, so if a future helper ever records it absolute the test fails loudly instead of silently duplicating the absolute-path case.
- Mutation-checked: reverting the anchor (`_ = baseDir`) fails the new test while `TestEvaluateWorkRecordCloseGateSelfRemote` still passes — so it covers a genuinely new branch, not the one already covered.

## Scoped out, tracked

**1. Drive-letter and `foo:bar/repo` bypass — `gas-azb`.** Confirmed, and it is the same early return as the scp-style loopback case (`git@localhost:/path`) already tracked there; that bead now carries your drive-letter repro. The fix shape is the one you describe and the one the shell twin already uses: decide on the **host**, do not reject scp-style outright. You are right that this is a second root cause and that anchoring `repoDir` does not touch it — `looksLikeSCPRemote` returns before `gitCommonDir` is ever reached. Worth noting `foo:bar/repo` is a third sub-case: the host route alone does not fix it either, since `foo` looks like an ordinary hostname. That one needs the path-existence check the anchor now makes possible.

**2. `--single-branch` / `--depth` false blocks — `gas-kg39` (P1).** Confirmed structural, not staleness — the single-ref refspec means `refs/remotes/origin/<other-branch>` is never created, and a fetch does not repair it. Filed P1 and wired as an explicit **gating input to the enforcement re-arm**, not just a note: it must be resolved or consciously accepted before `GC_WORK_RECORD_ENFORCE=1` goes on anywhere. Enforcement is off by default in this PR, so landing it cannot wedge anyone; the wedge risk is entirely at re-arm time, which is where the blocker now sits. Fix shape recorded as: detect a non-wildcard `remote.origin.fetch` and either widen the probe or fail **open** for that configuration specifically.

**3. Tag-only delivery.** Same bead, as a secondary, at the severity you gave it — `refs/remotes/*` not being a complete delivery record is the shared root cause.

**4. Diagnosability of the seven `err != nil` sites — `gas-teyk`.** Filed with your fix shape: debug-level log of the underlying error, boolean contract unchanged. Recorded explicitly that the fail-closed direction must not be flipped while doing it.

## On the empty-publication-set note

That split is already written — `gitPublicationRemotes` returns the configured list alongside the publication list, and `gitCommitOnRemote` branches on the difference (`configured == nil` → false, `len(configured) == 0` → true, all-self → false). It is one of the two changes stacked behind this PR, which is why it is not in this diff. Your point that the relative-path bug *widened* the set of repos reaching that branch is the reason I fixed the anchor here rather than deferring it: with the anchor correct, a repo carrying one relative self-remote now lands in the all-self case that split fails closed, instead of the no-remotes case that exempts it.

## Still open for a maintainer

The **#5055 sequencing** call from the PR description is unchanged — that PR and this one conflict in `gitCommitReachableOnBranch` and cannot both land as-is.

One process note, since it cost a day: both of your passes were posted as issue comments rather than formal reviews, so the PR still reports `reviews: 0`, `reviewDecision: ""`, `requested_reviewers: []`. Every state-based check reads that as "no one has reviewed this, waiting on a maintainer," which is the opposite of what was true. Not asking you to change anything — flagging it because the same shape has now bitten twice on this repo, and the comment timeline is the only reliable signal.

@sjarmak — back to you. I cannot re-request review through the reviewers field (outside contributor, pull-only on this repo), so this mention is the only way to return it to your queue.

csauer02-personal-user:
Population-scale receipts supporting this PR's direction, from a live city (~90 sessions, 16k+ closed beads), measured this week during a ledger-honesty audit:

- Of **919** `gc.work_outcome=shipped` closes: **292 (32%) carry no commit SHA at all**.
- Of the 627 that do cite a SHA: **49 point at objects that exist nowhere** in any repo on the host (positive-controlled), and **28 more are content-proven absent** from the integration branch — locally-reachable-only, exactly the class this PR closes.
- A misattribution study on 102 SHA-citing closes: **78 cite another bead's commit** (89% detection sensitivity, positive-controlled) — so even 'has a reachable SHA' overstates delivery.

Your incident was nine beads; on our store the local-reachability gap is a double-digit percentage of all shipped closes. The publication-oracle direction is measured-correct here. Happy to share the segmentation method if useful.

csauer02-personal-user:
Two more field observations from the same live-city audit, both pointing at this PR's direction — and one small scope proposal since you're already in this file:

1. **The reachability oracle also refuses honest re-landed work.** A shipped close whose commit was cherry-picked onto the integration branch (new SHA) fails `--is-ancestor` against the recorded branch even though the content genuinely landed — we hit this class repeatedly (squash/cherry-pick workflows give landed content a new identity). Content-or-publication oracles age better than commit-identity ones; one more argument for what this PR does.

2. **`isWorkRecordGatedBead` scoping gap, proven with a two-arm control:** the predicate excludes control beads via `gc.kind` only, and has never heard of `gc.step_id`/`gc.step_ref`. A formula step bead that carries step markers but no `gc.kind` is treated as a work bead and REFUSED under `GC_WORK_RECORD_ENFORCE` — demanding `gc.work_outcome` from a bead that speaks the control-plane `gc.outcome` vocabulary (the exact split this file's own comment describes). Identical bead with `gc.kind` set closes fine. Since this PR is already rewriting the gate: would you take widening the predicate to also exclude beads carrying step markers? It's ~2 lines plus a test, and it spares every formulas-v2 city a stamp-`gc.kind`-everywhere workaround (our current interim). Happy to send it as a follow-up onto your branch instead if you'd rather keep this PR's scope pure.

AJBcoding:
@csauer02-personal-user — yes, I'll take it. Confirmed against source before answering, and the confirmation comes with one detail that I think changes the shape of the fix rather than its direction.

## The predicate is exactly as you describe

`cmd/gc/work_record_gate.go`:

```go
func isWorkRecordGatedBead(bead beads.Bead) bool {
	if t := strings.TrimSpace(bead.Type); t != "" && t != "task" {
		return false
	}
	if strings.TrimSpace(bead.Metadata[beadmeta.KindMetadataKey]) != "" {
		return false
	}
	return true
}
```

Two exclusions, `gc.kind` is the only metadata one, and neither `gc.step_id` nor `gc.step_ref` appears anywhere in the non-test Go tree.

What makes this more than a nicety: the split you're invoking is already written down in this repo, in `internal/beadmeta/keys.go`, where `gc.work_outcome` is documented as *"distinct from OutcomeMetadataKey (`gc.outcome`): that key is the control-plane step"* vocabulary. So the gate is demanding a work-plane key from a bead the codebase's own comments classify as control-plane. Widening the predicate doesn't add a special case — it makes the predicate agree with the documented design. That's an easy yes.

## The nuance: the type check may already be catching the common case

`isWorkRecordGatedBead` returns false for any type that isn't `task`, and `cmd/gc/wisp_step_inject.go` creates step beads with `Type: "step"`. So conventionally-injected step beads are already excluded — by the type arm, before `gc.kind` is ever consulted.

That means your two-arm control was necessarily on a bead typed `task` (or untyped) that carries step markers. **Which was it, and what created it?** I ask because the two cases lead different places:

- If something in formulas-v2 is producing step beads typed `task`, the marker exclusion is the right fix *and* there's probably a creation-side typing bug worth knowing about separately — in which case excluding on markers would work correctly but also quietly mask it.
- If it's a bead legitimately typed `task` that carries step markers by design, then the marker exclusion is simply the correct and complete fix.

Either way I want the widening — a gate that demands `gc.work_outcome` from a bead speaking `gc.outcome` is wrong regardless of how the bead got its type, and defense in depth across both arms is right here. I'd just rather land it knowing which of the two we're looking at, and have the test encode that case specifically rather than assert the general shape.

## On sequencing — please do send it as a follow-up onto the branch

Taking you up on the second half of your offer, for a reason external to the change: this PR is currently blocked behind the **#5055 sequencing call** flagged in the description (that PR and this one conflict in `gitCommitReachableOnBranch` and can't both land as-is), and that's still waiting on a maintainer. Until it's resolved this branch can't land anyway, so there's no cost to keeping the diff under review stable and stacking yours on top — and a real cost to re-opening scope on a PR that a reviewer has already read once.

Send it onto the branch with the two-arm control as its test and I'll carry it.

## Your point 1, independently — a live instance from today

> the reachability oracle also refuses honest re-landed work … `--is-ancestor` fails even though the content genuinely landed

Confirming this from our side with a case that happened a few hours ago, because it's a cleaner reproduction than a squash workflow.

We had a verified fix at `0b0ab72` and intended a fast-forward merge to `main`. Between verification and landing, an automated hourly job put 18 commits on `main`, so the fast-forward was no longer available. The change landed by cherry-pick as `5e1f78d`. I diffed the two trees over the touched paths and they are byte-identical — the content landed exactly, under a new identity, for a reason that had nothing to do with the change itself.

A commit-identity oracle records that as not-landed. What's notable is that nobody chose a rewriting workflow: an unrelated background job moved the base, and that alone was enough to break identity. Any repo with automated commits on the target branch will manufacture this case on its own schedule, which I think strengthens your "content-or-publication oracles age better" argument past the squash/cherry-pick workflows it's usually argued from.

## Still open, unchanged

The **#5055 sequencing** decision remains the blocker on this PR and needs a maintainer.

@sjarmak — this is the second ping on that sequencing question; the first was on 08-10. I can't re-request review through the reviewers field (outside contributor, pull-only here), so an @-mention is the only channel I have to return this to your queue. If #5055 should land first I'm happy to rebase onto it instead — I mainly need the call, either direction.

csauer02-personal-user:
@AJBcoding Sent: AJBcoding/gascity#2, onto your branch as requested.

Short answer to your which-case question: **your first branch, with a twist — the field beads are *untyped*, not task-typed.** The formulas-v2 dispatch paths (`internal/dispatch` control + ralph expansion/retry) stamp `gc.step_ref` at creation but set neither `Type` nor `gc.kind`, so the type arm never fires (empty passes) and `gc.kind` is absent on 443 of our 779 live ref_only steps. `wisp_step_inject.go` types its steps `"step"`; the v2 paths never got the same treatment. Agreed that the marker exclusion would quietly mask that creation-side typing gap — the follow-up's commit message calls it out as deliberately-not-addressed, and we'll file the typing question as its own issue so it stays visible.

The test encodes the exact arm-A shape from the two-arm control (untyped, kindless, `gc.step_ref` only) rather than the general form, per your ask.

And thank you for the fast-forward-defeated-by-an-hourly-job case — that's a cleaner demonstration than any squash workflow, exactly because nobody chose a rewriting workflow and identity still broke. Noted for our side of the argument as well.

quad341:
@AJBcoding — you asked for eyes on this on **Aug 8** and got silence for twelve days. That's our failure, and I'm sorry. Here's where it actually stands.

**Verdict: fix-merge** — but the thing blocking it is a decision we owe you, not a change you owe us.

The review explicitly declined to decide one point: this PR **supersedes @rjgeng's #5055** — the two conflict in `gitCommitReachableOnBranch` — and you had asked for a maintainer ordering call. Closing another contributor's open PR as superseded is an outward-facing decision we won't make from an automated verdict, so it needs a human. That human step is exactly what stalled.

On the code, the review's substantive read:

> fundamental — the close gate gains a new blocking violation class (`not present on any remote`), changes `validateWorkRecordOnClose`'s return contract, alters branch-ref resolution from bare-name gitrevisions to explicit `refs/remotes/<pub>/<branch>` + `refs/heads/<branch>` probes, and adds a new stderr output surface. Gated warn-only by default, but it is a control-flow and exposure change, not a refactor.

That reads as a considered design, not a problem — the warn-only default is the right call for a new blocking class.

**Next:** the #5055 ordering call is now explicitly queued for a maintainer decision rather than sitting implicit in a review. Nothing needed from you until it lands.

quad341:
@AJBcoding, thank you for this and for your patience. Separating durability from reachability is the right diagnosis, and fixing the relative self-remote anchor from @sjarmak's review here, rather than deferring it, was the right call.

An honest update. On Aug 20 we told you nothing was needed from you. Since then #5055 merged (Aug 27), which settles the ordering question, and nobody told you. Then #5600 moved the gate into a new `internal/workrecord` package, and the HTTP close path (`internal/api/work_record_close_gate.go`) now calls it as a second door. Main moved under you while this waited on us. That delay is ours, not yours.

The change is still needed: on main, `workrecord.ValidateOnClose` checks reachability only, and `PreferredReachabilityRef` still hardcodes `refs/remotes/origin/`. **Would you port it onto `internal/workrecord`?** I know that's a real port, not a rebase. Some notes on the API shape:

- **A new entry point, with both doors on it.** Rather than widening `ValidateOnClose` to three positional probes and two return slices, add a function that takes the probes as a struct and returns violations and advisories as separate fields. Move both doors to it in this PR and retire `ValidateOnClose`. A door left on the old function would silently skip the durability rule, which is the ungated-door problem the package doc warns about.
- **Match main's probe style.** The new git probes need `…Context` variants like `CommitReachableOnBranchContext`, because the HTTP door passes the request context so a client that hangs up stops git. Multi-remote resolution should extend `PreferredReachabilityRef`/`CommitReachableOnEitherRef`, where #5055's origin-only probe now lives, rather than sit beside them.
- **The no-checkout case.** When a bead's owner names no checkout, both doors degrade reachability to `ReachabilityUnverifiedNote`. The durability probe hits the same case, so please decide explicitly whether it degrades too or fails closed.
- **Advisories on the HTTP door.** That door has no stderr. We'd log advisories through `workRecordGateLogf`, the way it logs the unverified note, and keep them out of the 409 and the response body.
- **Tests.** Main pins contract rows in each door's tests and the real-git probe in `internal/workrecord/gate_test.go` (see its header comment); git-spawning tests there are tracked in `test/test-resources.toml`. Please keep your precondition asserts and paired positives intact.

Same scope as now; the stacked follow-ups can stay separate. If a port this size is more than you have time for right now, just tell us.

PR #5104, opened 2026-08-08
fix(rig): make rig add work on a city bound to a store gc does not serve (gas-4cu)

AJBcoding:
Implements **gas-4cu**.

> **Rewritten against current main.** The original three commits added `mysql`
> to the hardcoded backend allowlist and gave it its own metadata vocabulary
> (`mysql_dsn`, `mysql_database`, cross-field rejection rules, nine testdata
> fixtures). #5140 replaced that allowlist with a registry and #5144 made the
> opaque storage binding the only non-dolt shape — removing `postgres` from the
> registry and deleting the `Postgres*` fields this branch read. Rebasing would
> have meant restoring code #5144 deliberately deleted. The fix is re-expressed
> on the generic binding instead, and is now backend-neutral: it serves
> postgres, mysql, or anything else the linked beads library is pointed at, and
> gc never learns the name.

## The bug

`gc rig add` against a city whose `.beads/metadata.json` carries a complete
opaque storage binding wrote hardcoded managed-Dolt metadata for the new rig
and then died reaching a Dolt server that does not exist:

```
gc rig add: exec beads init: managed Dolt server unreachable while
inspecting existing store 'gas'; refusing to force-reinitialize
```

`scopeSkipsManagedDoltForInit` already had an inheritance arm, but it fired
only for a scope with an authoritative `.beads/config.yaml` declaring
`endpoint_origin: inherited_city`. `gc rig add` writes that file *after* this
gate runs — so on a rig directory created moments earlier the resolve found
nothing authoritative, and every fresh rig on such a city fell through to the
managed lane. The one case the arm exists to catch was the one case it could
not see.

## The fix

A scope carrying no metadata of its own has not chosen a backend. It inherits
the city's, and that is decidable from the city alone — so ask the city
directly in exactly that case. Fifteen lines in one file.

**The rig is deliberately left unpinned.**
`TestInitAndHookDirSkipsDoltInitForInheritedCityPostgresRig` already states the
invariant for a *configured* inherited rig — no local metadata — and the same
reasoning covers a fresh one: an inherited scope resolves the city's binding on
each use, so a derived copy would only go stale when the city's store moves. An
earlier draft of this change wrote such a copy; that existing test caught it,
and the behaviour is now pinned for the fresh case too.

## Also here

Two defects found while reproducing the above end-to-end, both independent of
which backend a city uses:

- **Atomicity.** A bead-store init that failed partway left the `.beads/` it
  created behind, so the retry hit the "already contains a beads store" guard
  and the only way forward was to hand-remove the debris. Provision now removes
  only a directory it created itself — a pre-existing store, an `--adopt`
  store, and the rig directory are never touched.
- **Branch probing.** `ProbeDefaultBranch` read `refs/remotes/origin/HEAD` and
  nothing else, so a repo whose remote is named anything but `origin` silently
  registered the checked-out feature branch as the rig's mainline.
  `ProbeDefaultBranchFrom` tries every remote (origin first) and the banner
  reports which one answered — `Default branch: main (from ajb/HEAD)` — or says
  the branch was inferred when no remote HEAD is set.

## Testing

New `cmd/gc/rig_fresh_inherited_scope_test.go`, plus tests alongside the
atomicity and branch-probe changes. Each new test was confirmed to fail with
its fix reverted, not merely to pass with it.

Green on `./cmd/gc` (filtered to the storage-binding, provider-lifecycle,
rig-add and probe surface), `./internal/beads/contract`, `./internal/rig`,
`./internal/git`.

> **Note on local verification.** The listed packages were run directly and are
> green. The repo's full local pre-push suite was bypassed for this push with
> maintainer authorization: it could not produce a clean verdict on that host,
> for two environmental reasons found while landing this — the gate's job-count
> is Linux-shaped and fans out to 11 jobs regardless of load on macOS, and a
> live herdr test leaks a bound pane so the *next* run fails on a precondition.
> Both are filed separately and neither is reachable from this change (it
> touches only `cmd/gc`, `internal/git`, `internal/rig`). CI on this PR is the
> authoritative gate.


quad341 (approved):
@AJBcoding — thank you for this one, and specifically for the rewrite decision. Noticing that #5140 and #5144 had moved the ground under your original branch, and then re-expressing the fix on the generic opaque binding instead of restoring code #5144 deliberately deleted, is exactly the right instinct. The result is better than what was asked for: backend-neutral, fifteen lines, and gc never learns the backend's name.

The two defects you found while reproducing are a real bonus. The atomicity fix in particular — removing only a `.beads/` we created ourselves, never a pre-existing or `--adopt` one, never the rig directory — is the kind of thing that only surfaces when someone actually hits the retry wall, and you pinned all three arms. And you were right that a repo whose remote isn't named `origin` still has a mainline; silently registering a feature branch is a genuinely nasty failure.

I also appreciate that you confirmed each new test fails with its fix reverted. It shows in the test design — the boundary case where a scope carries its own `config.yaml` and must still hit endpoint-origin validation is what convinced me the gate isn't over-broad.

The only red is our subprocess census ratchet catching two raw `exec.Command` calls in the new git test. That's repo bookkeeping, not your fix. I'll route them through a tolerant `runGit` sibling (which also picks up the env scrubbing your precondition assert was missing), re-run, and merge with your credit intact. Nothing needed from you.


---

PR #5106, opened 2026-08-08
fix(orders): create order-dispatched wisps in the graph store

julianknutsen:
## Problem

On a city whose graph class is served by its own storage binding, **every wisp a formula order dispatched was born in the WORK ledger and minted under the WORK prefix.**

`internal/coordclass/class.go` classifies wisp roots as `ClassGraph`. The city's own convergence check therefore reads those beads as graph-class beads residing outside their binding and reports them **stranded** — and stranded is **fatal to boot**.

Measured on maintainer-city: **72 wrong-store `mc-wisp-*` beads accrued in roughly two hours**, a mix of closed and open (e.g. `order:reaper` open). Every one of them was created *after* the city converged onto the split. Left alone, **the city cannot restart.**

The cause is that order dispatch held a single store — the one opened for the order's target scope — and used it for two different coordination classes. It instantiated the molecule through that store, and wrapped that same value as **both** legs of the execution-fact emission:

```go
executionevent.EmitCurrent(m.rec, beads.GraphStore{Store: store}, beads.WorkStore{Store: store}, rootID, "order-dispatch")
```

`internal/orders/store.go` already documents the intended two-leg design (orders class + graph class) and warns: *"never rebase them onto a single class store — that is the single-store-assumption bug the graph-store-split audit root-caused."* The order-dispatch path was the site that still did.

## Fix

Resolve the graph class through the storage routes the process **already opened at boot**, and route the graph-class writes through it.

This is plumbing, not a variable swap. `memoryOrderDispatcher` gained a `storageRoutes *storageRoutes` field, threaded from `CityRuntime.storageRoutes` and `controllerState.storageRoutes` through `buildOrderDispatcher{,WithSnapshot,FromOrderSet}` into `newMemoryOrderDispatcher`. A new `graphStoreFor(store)` delegates to the existing `resolveGraphStore(routes, …)` seam rather than inventing a mechanism. The routes are *passed*, never re-resolved, so a tick-fired wisp and a webhook-fired wisp land in the same store the rest of the process reads.

Routed through the graph store in `dispatchWisp`: the graph-routing decoration, `molecule.Instantiate`, the root's `order-run:` label stamp, and the graph leg of the emission. The order-tracking bead stays on the order/work store. The double-wrap becomes two distinct stores:

```go
executionevent.EmitCurrent(m.rec, beads.GraphStore{Store: graphStore}, beads.WorkStore{Store: store}, rootID, "order-dispatch")
```

**The gate/cooldown/cursor federation is part of the fix, not a bonus.** The wisp root carries this order's `order-run:<scoped>` label, and that label is exactly what the wisp-aware single-flight gate, the cooldown clock and the event cursor read. Once the root lives in the graph binding, a gate that only scans the order's target store finds nothing — so **the order re-dispatches on every tick while its previous wisp is still open.** The dispatcher's gate store federation therefore gains the graph store (deduplicated by store identity), which is precisely what `orders.NewStoreWithGraph`'s union was built for.

## Single-store compatibility

Byte-identical. `resolveGraphStore` returns the **exact store value it was handed** when the routes relocate nothing (every city with no `[storage]` section), so:

- the molecule create sees the same instance, preserving the `GraphApplyFor` / `StorageCreateStore` optional-capability assertions it makes against it — no re-wrapping;
- the gate federation's append is skipped by the identity check, so the gate reads the same store list, with the same cache keys, that it always did.

Stated plainly: **the single-store compatibility test is green both before and after this change, by design.** A compatibility guard that went red on revert would be testing the wrong thing. Its teeth were proven by mutation instead — making `graphStoreFor` return a fresh store when routes are nil:

```
graphStoreFor returned *beads.MemStore(0x1928e42b4240), want the identical store value 0x1928e42b41b0
```

## Tests

Four new tests in `cmd/gc/order_dispatch_graph_store_test.go`, all driving the real `dispatch()` → `drain()` path against a `storageRoutes` value shaped exactly like the converged whole-split binding `openStorageRoutes` produces.

Red-before evidence (reverting only the routing decisions, keeping the plumbing so the package still compiles):

| Test | Red-before |
|---|---|
| `TestOrderDispatchWispRootLandsInGraphStoreOnSplitCity` | `order_dispatch_graph_store_test.go:145: no graph.v2 workflow root in store; beads = []` |
| `TestOrderDispatchExecutionFactsProjectFromTheGraphStore` | `order_dispatch_graph_store_test.go:265: step-definition subject "mc-3" not in the graph store: getting bead "mc-3": bead not found` |
| `TestOrderDispatchSingleFlightGateSeesGraphResidentWisp` | `order_dispatch_graph_store_test.go:230: workflow roots in the graph store = 2, want 1; the gate re-fired the order while gcg-1 was still open` |
| `TestOrderDispatchWispStaysOnTheOneStoreOnSingleStoreCity` | green before and after by design; mutation proof above |

The second row is **the production symptom reproduced in a unit test**: `mc-3` is a work-prefixed id for a bead the graph binding is supposed to own — the same class of bead as the 72 `mc-wisp-*` records on maintainer-city.

`internal/beads.MemStore` gained an `IDPrefix` knob so two in-memory stores can stand in as two bindings with distinguishable ids, following the existing `DisableConditionalWrites` precedent on that documented test double. Default is unchanged (`gc-<n>`).

## Gates

- `go build ./...` clean
- `go vet` clean on `cmd/gc`, `internal/beads`, `internal/orders`, `internal/molecule`
- `go test ./cmd/gc -run 'Order|Dispatch|Wisp|Storage|Molecule|Graph'` — ok, 48.2s
- `go test ./internal/{orders,molecule,beads,executionevent}` — ok
- New tests pass under `-race`
- `golangci-lint` 2.10.1 on `./cmd/gc/... ./internal/beads/...` — **0 issues**
- `gofmt` clean

## Follow-ups

This is **PR-1 of a series**. An audit of the same bug class found further sites where a coordination class is born in or read from the wrong store; they are deliberately not bundled here, because each theme needs its own single-store byte-identity proof and its own revert:

- **Boot routing correctness** — the controller's mail provider is constructed before `storageRoutes` is installed, `storageRoutes` is written unlocked while read under `RLock`, and extmsg services are never routed.
- **Remaining graph-class births** — convergence tick, `cmd_formula`, control dispatch.
- **Orders-class births *together with* every reader** — the boot orphan sweep, the tracking watchdog, `gc order check`/`history`, force-close and the API feed must move atomically with the writer. Landing the birth alone would make open tracking beads invisible, the single-flight marker would never clear, and orders would stop firing permanently — strictly worse than the symptom this PR fixes.
- **Read-side API projections** — `internal/api` State exposes no `OrdersBeadStore` and no `MailBeadStore`.

Separately, flagged for an **owner decision rather than a fix**: `internal/storebinding/sqlite/beads_engine.go:55` mints `gcg` for *all* classes, while `reserved_prefixes.go` documents distinct per-class prefixes. Whether the engine or the documented reservation is authoritative is a product call, so this PR does not touch it.

---

# Review follow-up (2026-08-08)

Council review found three majors — two reviewers on the same test-coverage defect, one on the recovery path. All are fixed here; the branch is also merged up to `main` (past #5107).

## Majors 1 + 3 — the production plumbing was untested

All four original tests assigned `storageRoutes:` as a struct-literal field. That proves `dispatchWisp` USES the field and nothing about the chain that fills it. The reviewers demonstrated — and I reproduced on this branch — that reverting the four production call sites to `nil` (`city_runtime.go` boot/rescan/reload, `api_state.go`'s webhook seam) leaves `go build`, `go vet`, `golangci-lint` (**0 issues**) and the whole `cmd/gc` suite green, while putting every dispatched wisp back in the work ledger. That is exactly the stranded-graph-bead incident this PR exists to fix, shipping green.

Note the verifier's correction to the original report: deleting `storageRoutes: routes,` inside `newMemoryOrderDispatcher` does **not** ship green — it leaves `routes` unused and revive fires. The call-site revert is the real regression shape, so that is the mutation used as the regression target below.

Three tests close it, and they reach the dispatcher through the constructor rather than around it:

- **`TestOrderDispatchConstructorDeliversRoutesToTheWispBirth`** — builds via `buildOrderDispatcherFromOrderSet(routes, …)`, drives the real `dispatch()` -> `drain()`, and asserts the wisp root mints under the binding's prefix with no workflow root in the work ledger. Only `storeFn` is replaced afterwards (the production opener would reach for a real bd store); the routes come from the constructor, which is the thing under test.
- **`TestOrderDispatcherServesTheBootResolvedBinding`** — boots a real split `CityRuntime` with an order on disk and pins `cr.od.(*memoryOrderDispatcher).storageRoutes == cr.storageRoutes` at boot, after `rescanOrderDispatcher` and after `reloadConfigTraced`. Pointer identity is the point: storage handles are immutable for the life of a process, so a dispatcher holding anything other than the boot-resolved routes is holding a handle to a database nothing else in the process reads. (The rescan stage adds an order first — the rescan only rebuilds when the order SET changed, so without that the stage would assert nothing.)
- **`TestWebhookOrderDispatcherServesTheBootResolvedBinding`** — the fourth site, which builds its own detached dispatcher per delivery. `controllerWebhookDispatcher.Dispatch` is split into a `dispatcher()` builder so what the seam hands the dispatcher is assertable without firing a delivery.

## Major 2 — the stale-wisp force-close still read the work stores

This PR moves order wisp roots into the graph binding and federates the single-flight gate over it. So a wisp whose pool agent dies mid-run leaves the root open with open descendants, `hasOpenWorkStrict` returns true, and the order is suppressed on every tick — while `gc order sweep-tracking --include-wisps`, the only gc-level force-close and the recovery the docs name, scanned the work stores, found no `order-run:<name>` root, reported `wispClosed: 0` and **exited 0**. The controller watchdog cannot help either (it passes `includeWispSubtrees=false`). The order stayed wedged with no recovery but raw `bd` against the binding.

This is a regression this PR introduces, not a pre-existing gap: on `main` the wisp is born in the work store (the bug), so the sweep finds it.

Fixed rather than documented, because "order wedged forever with no documented recovery" is a bad state to ship knowingly, and the seam already existed — `resolveGraphStore(cliStorageRoutes(cityPath), …)` is the same one-shot CLI hop `cmd_formula`, `cmd_sling` and `cmd_convoy_dispatch` already use.

- `sweepStaleOrderTrackingAcrossStores*` take the wisp store alongside the tracking stores. It is hoisted out of the per-scope loop **only when it is distinct**, because the binding is city-level and sweeping it once per scope would re-close the same subtrees and multiply the reported count.
- `orderTrackingSweepStoresForConfigTargets` returns the pair, so a caller cannot resolve the tracking stores and silently forget the wisp store — which is the same untested-wire defect as majors 1+3, and worth not re-introducing one function over.

This is the graph-class half only. The orders-class readers stay where they are; they still have to move atomically with the orders-class writer, which is the separate PR the follow-up list names.

## Single-store compatibility

Unchanged and still byte-identical. `orderWispSweepStore` returns `nil` when the routes relocate nothing (`resolveGraphStore` hands back the `workStore` it was given, and that is `nil` here), and `nil` is the signal to keep sweeping wisp subtrees inside the per-store loop, once per scope, exactly where they were.

## Red-before (mutations on this branch)

| Mutation | Red-before |
|---|---|
| revert all four production call sites to `nil` routes | `boot: dispatcher routes = 0x0, want the runtime's boot-resolved routes 0x…` and `webhook dispatcher routes = 0x0, want the controller's 0x…` |
| rescan site only | `after rescanOrderDispatcher: dispatcher routes = 0x0, want the runtime's boot-resolved routes 0x…` |
| reload site only | `after reloadConfigTraced: dispatcher routes = 0x0, want the runtime's boot-resolved routes 0x…` |
| drop the `storageRoutes` field in `newMemoryOrderDispatcher` | `no graph.v2 workflow root in store; beads = []` |
| keep the wisp sweep per-store only | `gc order sweep-tracking --include-wisps closed no wisp beads in the graph binding; a stalled graph-resident wisp has no gc-level recovery and its order stays suppressed forever` |
| drop the wisp store from the sweep resolver | `wisp sweep store = <nil>, want the graph binding 0x…; --include-wisps would report wispClosed: 0 and exit 0` |

The mutation that matters most is the first: it is the one the verifier proved could ship with build, vet, lint and the entire `cmd/gc` package green.

## Helper dedupe

`splitCityStorageRoutes` is gone; this branch uses `messagingSplitRoutes`, which #5107 landed on `main`. #5108 does the same, so the three near-identical helpers collapse to one without either PR depending on the other's merge order.

## Gates (re-run)

- `go build ./...` clean · `go vet ./...` clean · `gofmt` clean
- `go test ./cmd/gc ./internal/orders/... ./internal/molecule/... ./internal/beads/... ./internal/executionevent/...` — ok
- `scripts/check-core-boundary.sh` OK · `scripts/check-routed-test-rows.sh` OK · native dependency guard OK

---

# Review follow-up 2 (2026-08-08): the wisp sweep traded one class for the other

Council re-review found a Major that this branch introduced while fixing the
graph-side half, and it is right.

## What was wrong

```go
perStoreWisps := includeWispSubtrees && wispStore == nil
```

One non-nil `wispStore` turned the wisp half **off for every store in the
loop**, not just for the hoisted binding. The doc comment justified hoisting
the *binding* out of the loop — "sweeping it once per scope would re-close the
same subtrees and multiply the reported count" — and that rationale never
applied to the per-scope work/order stores, which are distinct databases and
cannot double-count anything.

So on a converged split city `gc order sweep-tracking --include-wisps` looked
**only** in the binding. A work-resident wisp root was unreachable by the only
gc-level force-close there is, while `hasOpenWorkStrict` still federates over
**both** stores (`dispatch()` *appends* the graph store to `storesForGate`
rather than replacing the target), so that root kept its order suppressed on
every tick. That is the same unrecoverable wedge the graph-side fix was written
to close, with the stores swapped — and the command printed
`closed 0 stale order wisp bead(s)` and exited 0, including under `--dry-run`.

Both arms are live, not hypothetical:

1. **Migration residue.** This PR's own problem statement measures 72
   wrong-store `mc-wisp-*` beads on maintainer-city, "a mix of closed and open
   (e.g. `order:reaper` open)". Those are work-ledger-resident open wisp roots.
2. **Ongoing.** `gc order run` still births the wisp through the raw scope
   store (`cmd_order.go:810` → `openOrderStoreForOrder` →
   `openStoreAtForCity`, no class routing), so a converged city keeps minting
   fresh work-resident wisp roots.

## Fix

Sweep both classes instead of trading one for the other. `perStoreWisps` is
`includeWispSubtrees` again, and the hoisted binding pass is guarded so a store
serving both classes is still swept exactly once — which is what the hoist was
actually for:

```go
perStoreWisps := includeWispSubtrees
...
if includeWispSubtrees && wispStore != nil && !storeListContains(stores, wispStore) {
```

`storeListContains` is the identity helper this branch already added for the
gate.

## Tests

| Test | What it pins |
|---|---|
| `TestOrderWispForceCloseSweepsBothClassesOnASplitCity` | a wisp subtree in EITHER class is recoverable, and the count is exact (4, not 2 and not 8) |
| `TestOrderWispForceCloseCountsAStoreServingBothClassesOnce` | the hoist's own reason survives: one store serving both classes reports its subtree once |

Red-before:

```
order_dispatch_graph_store_test.go:599: work-resident wisp bead gc-1 status = "open",
want closed; --include-wisps stopped sweeping the work stores, so a work-resident
wisp root has no gc-level force-close and keeps its order suppressed on every tick
```

The count-once test's teeth, proven by deleting the `storeListContains` guard:

```
order_dispatch_graph_store_test.go:627: dry-run wispClosed = 4, want 2; one store
serving both classes was swept twice and the operator is told twice as much work
is stale as there is
```

Dry-run is the mode that can actually double-count — the live path skips an
already-closed root on the second pass, which would hide a double sweep.

## Single-store compatibility

Still byte-identical, and now by construction rather than by argument:
`wispStore` is `nil` on a single-store city, so
`includeWispSubtrees && wispStore == nil` and `includeWispSubtrees` are the same
value, and the hoisted pass never runs.
`TestOrderWispForceCloseStaysInTheOneStoreOnSingleStoreCity` is green before and
after.

## Gates (re-run)

- `go build ./...` clean · `go vet ./...` clean · `gofmt` clean
- `go test ./cmd/gc -count=1 -timeout 40m` — **ok, 742s**

### DIFF
--- a/cmd/gc/api_state.go
+++ b/cmd/gc/api_state.go
@@ -2816,6 +2816,17 @@ func (cs *controllerState) WebhookDispatcher() orderdispatch.Dispatcher {
 type controllerWebhookDispatcher struct{ cs *controllerState }
 
 func (d controllerWebhookDispatcher) Dispatch(ctx context.Context, req orderdispatch.DispatchRequest) (orderdispatch.DispatchResult, error) {
+	return d.dispatcher().Dispatch(ctx, req)
+}
+
+// dispatcher builds the per-delivery dispatcher Dispatch fires through, reading
+// the controller's live config, recorder and storage binding under the
+// hot-reload lock. It is separate from Dispatch so what this seam hands the
+// dispatcher is assertable without firing a delivery: the routes are the whole
+// reason a webhook-fired wisp lands in the same graph store a tick-fired one
+// does, and they arrive as a constructor argument that nothing downstream
+// re-resolves.
+func (d controllerWebhookDispatcher) dispatcher() *memoryOrderDispatcher {
 	cs := d.cs
 	cs.mu.RLock()
 	cfg := cs.cfg
@@ -2827,11 +2838,7 @@ func (d controllerWebhookDispatcher) Dispatch(ctx context.Context, req orderdisp
 		// discard recorder keeps it panic-free when the city has events disabled.
 		rec = events.Discard
 	}
-	// The per-delivery dispatcher serves the same storage binding the
-	// controller booted, so a webhook-fired wisp lands in the same graph store
-	// a tick-fired one does.
-	md := newMemoryOrderDispatcher(routes, nil, cs.cityPath, cfg, rec, os.Stderr)
-	return md.Dispatch(ctx, req)
+	return newMemoryOrderDispatcher(routes, nil, cs.cityPath, cfg, rec, os.Stderr)
 }
 
 // ExtMsgServices returns the external messaging services.
--- a/cmd/gc/api_state.go
+++ b/cmd/gc/api_state.go
@@ -2790,6 +2790,7 @@ func (d controllerWebhookDispatcher) Dispatch(ctx context.Context, req orderdisp
 	cs := d.cs
 	cs.mu.RLock()
 	cfg := cs.cfg
+	routes := cs.storageRoutes
 	var rec events.Recorder = cs.eventProv
 	cs.mu.RUnlock()
 	if rec == nil {
@@ -2797,7 +2798,10 @@ func (d controllerWebhookDispatcher) Dispatch(ctx context.Context, req orderdisp
 		// discard recorder keeps it panic-free when the city has events disabled.
 		rec = events.Discard
 	}
-	md := newMemoryOrderDispatcher(nil, cs.cityPath, cfg, rec, os.Stderr)
+	// The per-delivery dispatcher serves the same storage binding the
+	// controller booted, so a webhook-fired wisp lands in the same graph store
+	// a tick-fired one does.
+	md := newMemoryOrderDispatcher(routes, nil, cs.cityPath, cfg, rec, os.Stderr)
 	return md.Dispatch(ctx, req)
 }
 
--- a/cmd/gc/city_runtime.go
+++ b/cmd/gc/city_runtime.go
@@ -329,7 +329,7 @@ func newCityRuntime(p CityRuntimeParams) (*CityRuntime, error) {
 		warnIfClosedOrderTrackingBacklogLarge(sweepStore, p.Stderr)
 	}()
 
-	od, orderSnapshot := buildOrderDispatcherWithSnapshot(p.CityPath, p.Cfg, p.Rec, p.Stderr, "gc start: order scan")
+	od, orderSnapshot := buildOrderDispatcherWithSnapshot(routes, p.CityPath, p.Cfg, p.Rec, p.Stderr, "gc start: order scan")
 
 	suspendedNames := computeSuspendedNames(p.Cfg, p.CityName, p.CityPath)
 
@@ -1462,7 +1462,7 @@ func (cr *CityRuntime) rescanOrderDispatcher(ctx context.Context, cityRoot strin
 		cr.drainOutgoingOrderDispatcher(drainCtx, cr.od)
 		drainCancel()
 	}
-	cr.replaceOrderDispatcher(buildOrderDispatcherFromOrderSet(cityRoot, cfg, snapshot.Orders, cr.rec, cr.stderr))
+	cr.replaceOrderDispatcher(buildOrderDispatcherFromOrderSet(cr.storageRoutes, cityRoot, cfg, snapshot.Orders, cr.rec, cr.stderr))
 	cr.orderSet = snapshot.Orders
 	cr.orderSetSignature = snapshot.Signature
 	if summary != "unchanged" {
@@ -2123,7 +2123,7 @@ func (cr *CityRuntime) reloadConfigTraced(
 		cr.drainOutgoingOrderDispatcher(drainCtx, cr.od)
 		drainCancel()
 	}
-	nextOD, orderSnapshot := buildOrderDispatcherWithSnapshot(cityRoot, nextCfg, cr.rec, cr.stderr, "gc reload: order scan")
+	nextOD, orderSnapshot := buildOrderDispatcherWithSnapshot(cr.storageRoutes, cityRoot, nextCfg, cr.rec, cr.stderr, "gc reload: order scan")
 	orderSummary := orderSetChangeSummary(cr.orderSet, orderSnapshot.Orders)
 	cr.replaceOrderDispatcher(nextOD)
 	cr.orderSet = orderSnapshot.Orders
--- a/cmd/gc/city_runtime.go
+++ b/cmd/gc/city_runtime.go
@@ -1542,7 +1542,7 @@ func (cr *CityRuntime) runOrderTrackingSweepWatchdog(now time.Time) {
 	// Closed-history retention is intentionally left to the maintenance exec
 	// order or the gc order sweep-tracking CLI; the watchdog only recovers
 	// stale open tracking beads.
-	result, sweepErr := sweepStaleOrderTrackingAcrossStoresLimit(stores, now, orderTrackingSweepWatchdogStaleAfter, nil, orderTrackingWatchdogMetadataInitiator, false, orderTrackingSweepCloseBudget)
+	result, sweepErr := sweepStaleOrderTrackingAcrossStoresLimit(stores, nil, now, orderTrackingSweepWatchdogStaleAfter, nil, orderTrackingWatchdogMetadataInitiator, false, orderTrackingSweepCloseBudget)
 	if err := errors.Join(storeErr, sweepErr); err != nil {
 		if cr.stderr != nil {
 			fmt.Fprintf(cr.stderr, "%s: order tracking sweep watchdog: %v\n", cr.logPrefix, err) //nolint:errcheck // best-effort stderr
--- a/cmd/gc/city_runtime_test.go
+++ b/cmd/gc/city_runtime_test.go
@@ -1980,6 +1980,7 @@ func TestOrderTrackingSweepWatchdogAllowsSweepOrderToCleanStaleTracking(t *testi
 		execRan = true
 		_, err := sweepStaleOrderTrackingAcrossStores(
 			[]beads.Store{store},
+			nil,
 			freshMerge.CreatedAt.Add(25*time.Millisecond),
 			50*time.Millisecond,
 			nil,
--- a/cmd/gc/cmd_order.go
+++ b/cmd/gc/cmd_order.go
@@ -1762,7 +1762,10 @@ func cmdOrderSweepTrackingWithOptions(staleAfter time.Duration, includeWisps, dr
 		fmt.Fprintf(stderr, "gc order sweep-tracking: %v\n", err) //nolint:errcheck // best-effort stderr
 		return 1
 	}
-	stores, openErr := orderTrackingSweepStoresForConfigTargets(cityPath, cfg, requiredTargets)
+	// wispStore is the store the sweep's wisp-subtree half runs against: on a
+	// split city the wisp roots this force-closes are graph class and live in
+	// the binding, not in any of the order stores beside them.
+	stores, wispStore, openErr := orderTrackingSweepStoresForConfigTargets(cityPath, cfg, requiredTargets)
 	if len(stores) == 0 {
 		if openErr != nil {
 			fmt.Fprintf(stderr, "gc order sweep-tracking: %v\n", openErr) //nolint:errcheck // best-effort stderr
@@ -1781,9 +1784,9 @@ func cmdOrderSweepTrackingWithOptions(staleAfter time.Duration, includeWisps, dr
 	// the non-zero exit is deferred to the end of the function.
 	confirmGateBlocked := false
 	if dryRun {
-		result, sweepErr = sweepStaleOrderTrackingAcrossStoresDryRun(stores, now, staleAfter, onlyOrders, includeWisps)
+		result, sweepErr = sweepStaleOrderTrackingAcrossStoresDryRun(stores, wispStore, now, staleAfter, onlyOrders, includeWisps)
 	} else {
-		result, sweepErr = sweepStaleOrderTrackingAcrossStores(stores, now, staleAfter, onlyOrders, includeWisps)
+		result, sweepErr = sweepStaleOrderTrackingAcrossStores(stores, wispStore, now, staleAfter, onlyOrders, includeWisps)
 
 		// Bulk-delete confirm gate: before any retention deletions, count
 		// eligible beads and require --confirm when above
--- a/cmd/gc/cmd_reload_test.go
+++ b/cmd/gc/cmd_reload_test.go
@@ -645,7 +645,7 @@ func TestReloadConfigTracedRescansOrdersWhenConfigRevisionUnchanged(t *testing.T
 		cfg:                cfg,
 		sp:                 runtime.NewFake(),
 		dops:               newDrainOps(runtime.NewFake()),
-		od:                 buildOrderDispatcherFromOrderSet(dir, cfg, initialOrders.Orders, events.Discard, &stderr),
+		od:                 buildOrderDispatcherFromOrderSet(nil, dir, cfg, initialOrders.Orders, events.Discard, &stderr),
 		orderSet:           initialOrders.Orders,
 		orderSetSignature:  initialOrders.Signature,
 		orderRescanEnabled: true,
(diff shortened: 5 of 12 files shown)


julianknutsen:
Pushed `ac04f3a0be` — review follow-up addressing all three majors, plus a merge up to `main` (past #5107).

**Majors 1 + 3 (untested plumbing).** Confirmed the verifier's stronger mutation on this branch: reverting the four production call sites to `nil` routes builds, vets, lints (`0 issues`) and passes the whole `cmd/gc` package. Three tests now reach the dispatcher through the constructor instead of around it — a behavioral one via `buildOrderDispatcherFromOrderSet` driving `dispatch()`→`drain()`, a boot/rescan/reload identity test over a real split `CityRuntime`, and one for the webhook seam (`controllerWebhookDispatcher.Dispatch` split into a `dispatcher()` builder so the seam's output is assertable without firing a delivery). Each of the four sites now fails on its own.

**Major 2 (stale-wisp force-close).** Fixed rather than documented. `gc order sweep-tracking --include-wisps` was the only gc-level recovery for a wisp this PR relocates, and it scanned the work stores — `wispClosed: 0`, exit 0, order suppressed forever. The across-stores sweep now takes the wisp store alongside the tracking stores, hoisted out of the per-scope loop only when distinct (the binding is city-level; sweeping per scope would multiply the count), and `orderTrackingSweepStoresForConfigTargets` returns the pair so the wire cannot be dropped independently — the same defect class as majors 1+3, worth not re-introducing one function over.

Red-before strings and the full gate list are in the updated PR body. Single-store cities stay byte-identical: `orderWispSweepStore` returns `nil` when the routes relocate nothing, which is the signal to keep the wisp half inside the per-store loop.

Not merging — flagged for review.

julianknutsen:
Fixed the Major from the re-review: `--include-wisps` was trading one class for the other instead of sweeping both.

`perStoreWisps := includeWispSubtrees && wispStore == nil` turned the wisp half **off for every store in the loop** as soon as a distinct binding existed. The hoist's stated rationale ("sweeping it once per scope would re-close the same subtrees and multiply the reported count") only ever justified keeping the *city-level binding* out of the loop — the per-scope work/order stores are distinct databases and cannot double-count. So on a converged split city the only gc-level force-close looked **only** in the binding, while `hasOpenWorkStrict` still federated over both stores (`dispatch()` appends the graph store rather than replacing the target). A work-resident wisp root therefore kept its order suppressed on every tick with no recovery — the wedge this PR exists to close, with the stores swapped — and the command printed `closed 0 stale order wisp bead(s)` and exited 0, `--dry-run` included.

Both arms are live: the 72 wrong-store `mc-wisp-*` beads in this PR's own problem statement are work-ledger-resident, and `gc order run` still mints its wisp root through the unrouted scope store, so fresh ones keep appearing.

Now `perStoreWisps` is `includeWispSubtrees` again and the hoisted binding pass is guarded with the `storeListContains` helper this branch already added, so a store serving both classes is still counted exactly once.

Red-before:

```
order_dispatch_graph_store_test.go:599: work-resident wisp bead gc-1 status = "open",
want closed; --include-wisps stopped sweeping the work stores, so a work-resident
wisp root has no gc-level force-close and keeps its order suppressed on every tick
```

The count-once test has teeth too — deleting the `storeListContains` guard:

```
order_dispatch_graph_store_test.go:627: dry-run wispClosed = 4, want 2; one store
serving both classes was swept twice and the operator is told twice as much work
is stale as there is
```

Dry-run is the mode that can actually double-count; the live path skips an already-closed root on the second pass and would have hidden it.

Single-store stays byte-identical by construction: `wispStore` is nil there, so the old and new `perStoreWisps` expressions are the same value and the hoisted pass never runs.

Gates: `go build ./...` clean, `go vet ./...` clean, `gofmt` clean, `go test ./cmd/gc -count=1 -timeout 40m` ok (742s), `golangci-lint` 2.10.1 on `./cmd/gc/...` 0 issues.

julianknutsen:
Fixed the `Integration / packages-core-2-of-4` failure (`TestLegacyFormulaV2MechanismFrozen`) in 859af87fca.

**What the guard counts.** It greps every `.go` file for the legacy formula_v2 identifiers and, per package dir, caps how many *test* files reference them. `cmd/gc` is frozen at 5; this PR's new `order_dispatch_graph_store_test.go` made 6. The ceiling is a deliberate ratchet against new code adopting the legacy path, so I converted rather than raising it.

**The conversion.** The file had exactly one coupling: `formulatest.EnableV2ForTest(t)` in `newGraphOrderFixture`. Worth noting `rollout.ForTest` alone could not have replaced it — `rollout.Flags.FormulaV2()` has zero production consumers; it's a reporting-surface gate, while the compiler reads a process global. The bridge the freeze test explicitly sanctions (and deliberately excludes from its needles) is `applyFeatureFlags`. So the fixture now declares `[daemon] formula_v2` on the city config it already builds and propagates it the way a booting city does, restoring the default on cleanup — the same idiom `graph_dispatch_mem_test.go` and `cmd_sling_test.go` already use.

Side benefit: both gates derive from the one config knob, so this drops the fixture into the pairing production actually runs. The legacy setter left the compiler capability on with graph-apply *off*, a state no real city boots into.

**Teeth re-verified.** Reverting the four production call sites to nil routes still goes red, and each site is pinned individually:

- `city_runtime.go:332` — `boot: dispatcher routes = 0x0, want the runtime's boot-resolved routes 0x194482e7d10; every wisp it dispatches is born in the work ledger and lands stranded off the binding`
- `city_runtime.go:1465` — `after rescanOrderDispatcher: dispatcher routes = 0x0, want the runtime's boot-resolved routes 0x2ad02ac938c0; ...`
- `city_runtime.go:2126` — `after reloadConfigTraced: dispatcher routes = 0x0, want the runtime's boot-resolved routes 0x2e741f5dde30; ...`
- `api_state.go:2804` — `webhook dispatcher routes = 0x0, want the controller's 0x194480882d0; a webhook-fired wisp would be born in the work ledger`

**Gates.** `go build ./...` clean; `go vet ./...` clean; gofmt clean; `go test ./internal/testenv/` (full package, the failing CI job) ok 19.0s; `go test ./cmd/gc -timeout 40m` ok 732.0s; golangci-lint 2.10.1 on `./cmd/gc/...` → 0 issues. No production code touched — the diff is test-only.

PR #5108, opened 2026-08-08
fix(dispatch): run control dispatch against the graph store

julianknutsen:
## The owner ruling

> "the synthetic convoy bead is a work bead so it should live in the work db not the infra one"

`coordclass` classified graph.v2 input convoys and drain-unit convoys as
`ClassGraph`. That was the defect this PR kept running into rather than the
routing this PR was adding.

A convoy exists to hold `tracks` edges to its members. Its members are work
beads. `convoy.TrackItemIn` refuses an edge whose member is owned by another
class, because a dep row lives in one store's dep table and can only reference
an id that store can resolve. So a convoy owned by any class but Work is a
convoy that can never track anything — the classification was not a filing
preference, it was unsatisfiable.

With this PR's earlier commits routing control dispatch to the graph binding,
that became visible as a hard failure: a split-city graph.v2 drain expanded to N
rows and then died at `trackDrainMember`, because `ensureDrainUnitConvoy` minted
the unit convoy in the GRAPH store and immediately tried to track a WORK member.
A non-transient control error is quarantined by the caller, so drains
quarantined instead of completing on a converged split city.

## What changed

**`internal/coordclass`** — `type=convoy` carrying `gc.synthetic` /
`gc.synthetic_kind` is now `ClassWork`. The rest of `ClassGraph` is untouched:
molecule/step/gate/scope/run beads, every `gc.kind` control bead, wisp roots,
convergence beads and spec sidecars all stay graph. The doc comments on
`ClassWork`/`ClassGraph` no longer claim otherwise.

It is an **explicit arm placed before the workflow arm**, not the absence of an
arm. A synthetic convoy is glue for one specific graph and carries that graph's
identifying metadata; left to the default, a convoy's class would depend on
which incidental keys happen to be stamped on it, and one `gc.root_bead_id`
would silently re-break drains. Pinned by
`synthetic convoy with root_bead_id stays work` and
`drain-unit convoy with graph.v2 contract stays work`.

**`internal/dispatch/drain.go`** — a drain unit convoy is minted, looked up
(`gc.drain_unit_key`) and reloaded in the store that owns the member it tracks,
via the new `drainUnitConvoyStore`. Deriving it from the member is not a
heuristic: co-residence is what `TrackItemIn` requires, so the member's owning
store is the only store where the edge can be written. It reuses the
`drainMemberOwningStore` probe this series already established, with the same
single-store guard `drainMemberDepStore` uses.

**`internal/graphv2`** — no plumbing change, deliberately. Every caller of
`PrepareInvocation` already passes the work/scope store (`sling` passes
`deps.Store`, explicitly distinct from `deps.GraphStore`; `gc formula cook`
passes `openStoreAtForCity(scope.storeRoot, …)`), and `NormalizeInputConvoy`
resolves the target from the same handle it mints through — so co-residence is
structural, and routing it at the graph binding would fail to find the target
long before it could mint into the wrong ledger. Documented and pinned by
`TestSyntheticInputConvoyIsWorkClassAndCoResidentWithItsTarget` rather than
re-plumbed into a parallel mechanism. The classification was the only thing that
disagreed with where the code already put these.

## End-to-end drain result

`TestProcessDrainSplitCityExpandsAndCompletes` is the acceptance test, over both
migration shapes (convoy only in the work store; edgeless migration copy also in
the binding). It asserts the terminal state, not just the expansion: no error
(i.e. no quarantine), `action=drain-expanded`, `gc.drain_state=expanded`, a
manifest row per member, and for each row a unit convoy that is **readable from
the work store** and **actually tracks its member** there, plus an item root and
`status=wired`.

Red before:

```
--- FAIL: TestProcessDrainSplitCityExpandsAndCompletes/convoy_only_in_the_work_store
    ProcessControl(split-city drain): gc-1002: tracking member gc-2 from unit convoy gc-1003:
    getting tracked item gc-2: resolving gc-2 across convoy: bead not found;
    a control dispatch that returns a non-transient error is QUARANTINED by the caller,
    which is the state this ruling exists to remove
--- FAIL: TestProcessDrainSplitCityExpandsAndCompletes/edgeless_migration_copy_also_in_the_binding
    (same)
```

Green after, both shapes. The prior behavior (expand to N, then quarantine) is
gone.

## Blast-radius audit

`coordclass.Classify` has exactly five non-test consumers; each was checked.

| Consumer | Effect |
|---|---|
| `readInfraSnapshot` (`gc storage migrate` copy, `gc storage status`, **and the boot-time containment re-check**) | Synthetic convoys are no longer selected. This is the one that mattered: they are minted in the work store and stay there, so under the old classification **every drain a converged city ran after cutover minted a bead the binding never received**, and the next boot would count them as stranded writes and refuse to serve. Pinned by `TestInfraSnapshotLeavesSyntheticConvoysInTheWorkStore`. |
| `censusRigInfraResidue` (pre-migration refusal) | A synthetic convoy in a rig scope no longer blocks a cutover. Correct: it is a work bead and belongs where it is. Strictly a relaxation. |
| `infraCopyClassDifference` (migration equality stage) | Never sees synthetic convoys now, since the copy no longer carries them. Migration-time only; not on any boot path. |
| `LegacyCombinedSource.ReadAll` / `ReadClass` (born-split legacy import) | Synthetic convoys now land in the Work partition **with their members**, instead of being split away from the tracks edges that reference them. Improvement. |
| `beadPolicyStore.Create` → `createTarget(Classify(b))` | `createTarget` is identity today; no behavior change. |

Storage tier is a separate classifier (`policyNameForBead`) and has no
synthetic-convoy arm — it never did — so tiering is untouched.

Also checked: `workflowStores`/graph-first readers (they scan for workflow
**roots**, not convoys, and still scan the city store); `internal/convoy`
`MemberClasses` (doc corrected — it claimed Graph owned synthetic convoys);
`internal/executionevent` (already read input-convoy `tracks` edges from
`beads.WorkStore` while reading steps from `beads.GraphStore` — the ruling makes
`coordclass` agree with an assumption this code already encoded); the whole
`internal/storebinding` tree (no convoy references); and
`engdocs/plans/store-domain-objects/spec.md`, which stated the old rule and is
corrected.

**Boot verdict / running cities.** The check itself is unchanged. The snapshot
`readInfraSnapshot` returns is now a strict **subset** of what it returned
before, and `classifyInfraContainmentGap` only ever iterates that set — so
`Stranded` and `RemovedSinceCutover` can only shrink. A city that boots today
cannot stop booting because of this; the verdict can only move
`stranded → converged`, never the reverse.

For maintainer-city specifically (137 stranded `mc-wisp-*`): **no change,
confirmed**. The wisp arm runs first and is untouched, so wisp roots stay graph
class no matter what else they carry — pinned by
`a wisp carrying a synthetic marker is still graph`, and by the wisp rows
(`gc.kind=wisp`, `type=wisp`, `gc:wisp`) in the snapshot test. The only beads
whose class moves are `type=convoy` beads carrying a synthetic marker. If any
city's stranded list ever did contain a synthetic convoy id, it would drop off —
that is the correction, not a regression, since such a bead was never going to
be in the binding.

**Not fixed here (pre-existing, now reachable).** `collectInputConvoyWorkflowRoots`
(`cmd/gc/wisp_autoclose.go`) reads a work bead's tracking convoys from the
*graph* store, so on a split city it finds nothing. That is independent of this
ruling — input convoys were already work-resident at every call site — but it
becomes reachable for drain item roots now that split-city drains complete at
all. Separate concern, separate change.

## The membership-union decision

Kept, with its rationale demoted and rewritten.

The union is now **redundant for correctness**: under the ruling the work leg is
the authoritative one, and a work-only read is provably sufficient (an edge
inventory can only exist co-resident with its members, members are work beads,
and `gc storage migrate` copies rather than moves, so the work store always
retains the original with its edges).

It is **not redundant for safety**, and the risk is asymmetric. A work-only read
is correct only for as long as the residency proof holds; if any future path
mints a convoy somewhere unexpected, it returns zero members — and a drain that
could not read its membership produces byte-for-byte the same terminal state as
a drain that correctly found nothing to do (`gc.outcome=pass`, every member left
open and undispatched, exit 0). The union cannot fail that way: it is order-free
and monotone, no store's silence can subtract a member another store records,
and it cannot error on a store that never saw the convoy (verified: both
`SQLiteStore.DepList` and `BdStore.DepList` return empty for an unknown id). Its
cost is one extra edge read per drain, on split cities only.

Its *stated* justification did have to change, and did: it no longer claims
membership legitimately lives in either class. It now says the work leg answers,
and the graph leg exists because cities migrated under the previous
classification carry an edgeless copy of every synthetic convoy in their binding
— so "the store that has the convoy bead" is still an ambiguous question on real
disks. New migrations no longer create those copies.

## Teeth (mutation-verified, not argued)

| Mutation | Caught by |
|---|---|
| synthetic convoy arm back to `ClassGraph` | `TestClassifyGoldenTable` (4 rows) + `TestInfraSnapshotLeavesSyntheticConvoysInTheWorkStore` (`snapshot carried 6 rows, want exactly the 4 infra beads`) |
| unit convoy minted in the ambient (graph) store | `TestProcessDrainSplitCityExpandsAndCompletes` (both shapes) + `TestDrainUnitConvoyStoreFollowsTheMemberAcrossTheClassBoundary` |
| single-store guard removed from `drainUnitConvoyStore` | `TestDrainUnitConvoyStoreIsTheAmbientStoreOnASingleStoreCity`: `the single-store path issued 1 owning-store probe read(s)` |

Single-store byte-identity is pinned by **read count**, not by inspection: with
no member stores named the owning-store probe must not run at all.

## Gates

- `go build ./...` clean; `go vet ./...` clean
- `golangci-lint 2.10.1` — **0 issues** on `./cmd/gc/...` and on
  `./internal/{coordclass,dispatch,convoy,graphv2}/...`
- `go test ./cmd/gc -timeout 40m` — **ok, 751.9s**
- `go test ./internal/...` — all packages pass (full sweep)
- pre-commit hook (lint-changed + doc-gen + `test/docsync`) ran clean
- rebased on current `main` (`#5105`, `#5106`, `#5107` merged) — **no conflicts**

---

# Council review fixes (1 blocker, 1 major)

Two independent reviewers found defects in the split-city drain routing above.
Both were **reproduced by experiment against head `f6c38623cf`** before being
fixed, and every fix is **mutation-verified**: reverting it turns a named test
red. Red-before strings are quoted verbatim below.

## BLOCKER — an out-of-convoy blocker became a dangling graph-store edge

`drainProjectedBlockerIDs` mapped in-manifest blockers onto the blocker's item
root (graph-resident, fine), but for a blocker that is **not** a manifest member
it kept the raw id — a WORK-store id. This PR moves item roots into the graph
binding, so those raw work ids were written as `blocks` edges **into the graph
store** by both `molecule.Instantiate(ExternalDeps)` and
`ensureDrainWorkflowBlocksOn` → `store.DepAdd`. `DepAdd` does not validate its
target, so the row was written silently and dangled.

That does **not** degrade to an unenforced edge. Both `MemStore.readyLocked` and
`sqliteReadySQL` read an absent blocker's status as `""` — which is not
`'closed'` — so the item step drops out of `Ready` **permanently**,
`completeDrain` returns `ErrControlPending` forever, and closing the real blocker
in the work store cannot release it: the store holding the edge can never see
that row. No error, no quarantine, no operator signal; recovery needs manual
`DepRemove` surgery in the binding. An **already-closed** out-of-convoy blocker
produced the identical permanent block, which makes the trigger *any* historical
`blocks` edge in a real backlog rather than only an open dependency. This
falsified the stated acceptance ("expands to N rows **and completes**").

**Fix — the same co-residence rule the convoy edge already has**
(`classifyDrainBlocker`). An out-of-manifest blocker is projected only when the
item workflow's own store resolves it. Otherwise:

| blocker | outcome | why |
|---|---|---|
| terminal (closed/tombstone) | dropped, traced | it constrains nothing; on a single-store city its edge exists only for the readiness reader to ignore, so omitting it loses no meaning. Recording it would fire on essentially every real backlog and train operators to ignore the key. |
| open | dropped, **recorded** on the item root as `gc.drain_unprojected_blockers`, plus a trace line | the constraint is real and cannot be written down, so the loss is made durable and operator-visible |

**Why record rather than refuse** (the deliberate choice the review asked for):
refusing would turn the commonest backlog shape into a hard control failure on
split cities — the exact terminal state this routing exists to remove — and
writing the edge wedges the workflow forever. Recording is the only option that
neither loses the information nor stalls the work. The rejected alternative (a
graph-resident proxy bead mirroring the work blocker's status) needs a whole new
reconciler to close the proxy; that is a mechanism, not a fix, and it is out of
scope here.

In-manifest blockers are untouched: they project onto another row's item root,
minted in the same store, so they are co-resident by construction and need no
probe.

Red before (`TestProcessDrainSplitCityDoesNotWriteACrossStoreBlockerEdge`, both
subtests):

```
dangling blocks edge gc-1003 -> gc-4: the store holding the edge cannot resolve
the target, so the dependent never becomes ready and closing the real blocker in
its own store cannot release it: getting bead "gc-4": bead not found
```

and, with the record suppressed:

```
gc.drain_unprojected_blockers on item root gc-1003 = "", want "gc-4"; an item
workflow that runs without a constraint its source member had must say so on its
own bead, not only in a log line
```

## MAJOR — unit-convoy residence was re-derived from the member on every pass

`drainUnitConvoyStore` answered "which store owns this unit convoy?" by probing
the **member**, and `drainMemberOwningStore` falls back to the ambient graph
store when no probe owns it. But `row.UnitConvoyID` is persisted in the manifest
while its store is not, and the mint, the `gc.drain_unit_key` idempotence lookup
and the reload each re-derived it independently — so the instant a member's
resolvability changed between passes, mint and reload disagreed.

`expandDrain` persists after every row, so any mid-loop exit (`ErrControlPending`,
a reservation retry, a transient store error) leaves rows recorded with
`gc.drain_state=expanding`; that is a designed re-entry, not a crash. A member
hard-deleted between passes is likewise a supported state —
`loadDrainManifestMembers` synthesizes a placeholder, `trackDrainMember` stamps
`gc.drain_member_unresolved` — and is pinned for single-store cities by the
existing `TestProcessDrainReplayUsesManifestWhenMemberWasDeletedBeforeUnitCreation`.
On a split city the reload then probed the graph binding, reported a convoy that
exists as missing, and — a not-found is not a transient controller error — the
caller **quarantined** the drain. The mirror case (`row.UnitConvoyID` empty)
missed the key lookup and minted a **work-class bead into the infra ledger**,
which `infraCopyClassDifference` says never happens.

**Fix — separate the three questions.**

| question | answered by | was |
|---|---|---|
| reload a convoy the manifest names | its **own id**, `storeref.Resolve` over the probe set — exactly as `loadDrainManifestMembers` resolves a member | re-derived from the member |
| does one already exist for this `gc.drain_unit_key`? | `drainUnitConvoyByKey`, spanning **every** candidate store | one store, re-derived from the member |
| where to **mint** a new one | the member's owning store; an **unresolved** member pins it to the WORK class (`drainWorkClassStore`) | fell back to the ambient graph binding |

The unit-convoy probe set is **work-class first**, the reverse of
`drainMemberProbeSet`. A synthetic convoy is a work bead, so the work class is
authoritative — and a city migrated under the previous classification carries an
**edgeless copy** of every synthetic convoy in its binding, so asking the binding
first answers `gc.drain_unit_key` with a convoy that has no `tracks` edge and
cannot grow one.

Red before:

```
--- FAIL: TestProcessDrainSplitCityResumeCompletesAfterAMemberVanishes
    ProcessControl(resume after a member vanished): gc-1002: loading drain unit
    convoy gc-4: getting bead "gc-4": bead not found; a non-transient control
    error is quarantined by the caller, and the drain never expands its
    remaining rows

--- FAIL: TestProcessDrainSplitCityUnitConvoyForAVanishedMember/no_unit_convoy_minted_yet
    unit convoy gc-1003 for the vanished member gc-2 is in the GRAPH binding; a
    synthetic convoy is a work bead, and the migration's own equality invariant
    says work never crosses into the infra ledger

--- FAIL: TestEnsureDrainUnitConvoyFindsAConvoyMintedBeforeItsMemberVanished
    resuming pass created unit convoy gc-2 (created=true), want the
    already-minted gc-1003; a lookup re-derived from the member moves stores the
    moment the member stops resolving

--- FAIL: TestEnsureDrainUnitConvoyPrefersTheWorkStoreOverAnEdgelessBindingCopy
    ensureDrainUnitConvoy(split city): gc-1: repairing unit convoy gc-4 track for
    member gc-3: getting tracked item gc-3: resolving gc-3 across convoy: bead
    not found; the lookup answered with the binding's edgeless copy and then
    tried to repair its missing track across the class boundary
```

## Single-store byte-identity, preserved

`drainUnitConvoyStore` keeps its zero-probe short-circuit;
`reloadDrainUnitConvoy` keeps the same bare `store.Get` including its error text;
the key lookup is the one `ListByMetadata` it was; and `classifyDrainBlocker`
returns without reading anything. Both are pinned by read **count**, not by
inspection: `TestDrainUnitConvoyStoreIsTheAmbientStoreOnASingleStoreCity` (kept
meaningful, unchanged) and the new
`TestDrainProjectedBlockerIDsIsProbeFreeOnASingleStoreCity`. The boot verdict and
the stranded check are untouched.

## Docs (the reported minors)

Two comments still asserted the pre-ruling classification — `MemberClasses.Graph`
("formula steps, control beads, **synthetic convoys**"), which the body above
claimed was corrected when only the type header had been, and
`cmd_convoy_dispatch.go`'s store-split comment ("drain units **and their
convoys**) are graph class"). Both now say what the ruling says.

## Teeth (mutation-verified)

| Mutation | Caught by |
|---|---|
| reload reads the ambient store instead of resolving by id | `TestProcessDrainSplitCityResumeCompletesAfterAMemberVanishes` |
| mint falls back to the ambient store for an unresolved member | `…UnitConvoyForAVanishedMember/no_unit_convoy_minted_yet` |
| `gc.drain_unit_key` lookup against one re-derived store | `TestEnsureDrainUnitConvoyFindsAConvoyMintedBeforeItsMemberVanished` |
| probe order back to primary-first | `TestEnsureDrainUnitConvoyPrefersTheWorkStoreOverAnEdgelessBindingCopy` |
| co-residence rule removed | `TestProcessDrainSplitCityDoesNotWriteACrossStoreBlockerEdge` (both) |
| unprojectable blockers dropped without a record | `…DoesNotWriteACrossStoreBlockerEdge/open_out-of-convoy_blocker_is_recorded` |

## Gates (council-fix commit)

- `go build ./...` clean; `go vet ./...` clean; `gofmt` clean
- `golangci-lint 2.10.1` — **0 issues** on `./cmd/gc/...` and
  `./internal/{dispatch,convoy,coordclass,beadmeta}/...`
- `go test ./cmd/gc ./internal/dispatch ./internal/convoy ./internal/coordclass -timeout 40m` — pass
- `go test ./internal/...` — full sweep, pass
- `cmd/genschema` + `cmd/genspec` produce no drift

### DIFF
--- a/cmd/gc/class_store.go
+++ b/cmd/gc/class_store.go
@@ -306,6 +306,17 @@ func resolveGraphStore(routes *storageRoutes, workStore beads.Store, cfg *config
 	return resolveClassStore(routes, workStore, cfg, cityPath, config.BeadClassGraph, rec)
 }
 
+// graphClassBinding returns the store these routes serve the graph class from,
+// and whether they relocate it at all — the same question resolveClassStore asks
+// to choose its branch, exposed for the callers that must BEHAVE differently
+// rather than merely read elsewhere. A reader that answers by shelling `bd` in
+// the work directory cannot follow a relocated class, so it has to know it must
+// go in-process instead; resolveGraphStore alone cannot tell it, because a
+// relocated store and an unrelocated one are both just a beads.Store.
+func graphClassBinding(routes *storageRoutes) (beads.Store, bool) {
+	return routes.storeFor(coordclassFor(config.BeadClassGraph))
+}
+
 // newCityMailProvider builds the controller's mail provider as a two-store mail
 // provider: message beads persist in the messaging-class store, and mail's
 // session reads/writes for addressing/identity resolution go to the session-class
--- a/cmd/gc/cmd_convoy_dispatch.go
+++ b/cmd/gc/cmd_convoy_dispatch.go
@@ -193,10 +193,12 @@ func runControlDispatcherWithStoreAndConfig(cityPath, storePath string, store be
 	}
 
 	// store is the SCOPE store. Control beads, the workflow topology they
-	// mutate, and every bead the control kinds create (retry attempts, fanout
-	// fragments, drain units and their convoys) are graph class, so all of that
-	// runs against the graph store. store itself stays the work leg: it owns the
-	// input convoy whose tracks edges the execution snapshot below reads.
+	// mutate, and the graph beads the control kinds create (retry attempts,
+	// fanout fragments, drain item roots) are graph class, so all of that runs
+	// against the graph store. store itself stays the work leg: EVERY convoy is
+	// a work bead, the synthetic drain-unit ones included, so it owns both the
+	// input convoy whose tracks edges the execution snapshot below reads and the
+	// unit convoys a drain mints alongside its members.
 	graphStore := controlGraphStore(cityPath, storePath, cfg, store)
 
 	bead, err := graphStore.Get(beadID)
--- a/cmd/gc/cmd_convoy_dispatch.go
+++ b/cmd/gc/cmd_convoy_dispatch.go
@@ -160,7 +160,10 @@ func runControlDispatcherInStore(cityPath, storePath, beadID string, stdout, std
 	if err != nil {
 		return fmt.Errorf("opening scoped control store %q: %w", storePath, err)
 	}
-	bead, err := store.Get(beadID)
+	// The control bead is graph class, so it is read from the graph store — not
+	// from the scope store, which on a split city holds only the copy the
+	// migration retained and no longer receives the workflow's mutations.
+	bead, err := controlGraphStore(cityPath, cfg, store).Get(beadID)
 	if err != nil {
 		return fmt.Errorf("loading bead %s from scoped control store %q: %w", beadID, storePath, err)
 	}
@@ -187,6 +190,13 @@ func runControlDispatcherWithStoreAndConfig(cityPath, storePath string, store be
 	} else {
 		warnLegacyWorkflowTracePath(cityPath, nil, stderr)
 	}
+
+	// store is the SCOPE store. Control beads, the workflow topology they
+	// mutate, and every bead the control kinds create (retry attempts, fanout
+	// fragments, drain units and their convoys) are graph class, so all of that
+	// runs against the graph store. store itself stays the work leg: it owns the
+	// input convoy whose tracks edges the execution snapshot below reads.
+	graphStore := controlGraphStore(cityPath, cfg, store)
 
 	opts := dispatch.ProcessOptions{CityPath: cityPath, StorePath: storePath}
 	opts.Tracef = workflowTracef
@@ -223,12 +233,12 @@ func runControlDispatcherWithStoreAndConfig(cityPath, storePath string, store be
 		case "check", "fanout":
 			opts.FormulaSearchPaths = workflowFormulaSearchPaths(cfg, bead)
 			opts.PrepareFragment = func(fragment *formula.FragmentRecipe, source beads.Bead) error {
-				return decorateDynamicFragmentRecipe(fragment, source, store, loadedCityName(cfg, cityPath), cityPath, cfg)
+				return decorateDynamicFragmentRecipe(fragment, source, graphStore, loadedCityName(cfg, cityPath), cityPath, cfg)
 			}
 		case "drain":
 			opts.FormulaSearchPaths = workflowFormulaSearchPaths(cfg, bead)
 			opts.PrepareRecipe = func(recipe *formula.Recipe, source beads.Bead) error {
-				return decorateDrainItemRecipe(recipe, source, store, workflowStoreRefForDir(storePath, cityPath, loadedCityName(cfg, cityPath), cfg), loadedCityName(cfg, cityPath), cityPath, cfg)
+				return decorateDrainItemRecipe(recipe, source, graphStore, workflowStoreRefForDir(storePath, cityPath, loadedCityName(cfg, cityPath), cfg), loadedCityName(cfg, cityPath), cityPath, cfg)
 			}
 		case "retry-eval":
 			sp, err := dispatchControlSessionProvider()
@@ -256,7 +266,7 @@ func runControlDispatcherWithStoreAndConfig(cityPath, storePath string, store be
 		}
 	}
 
-	result, err := dispatch.ProcessControl(store, bead, opts)
+	result, err := dispatch.ProcessControl(graphStore, bead, opts)
 	if err != nil {
 		if errors.Is(err, dispatch.ErrControlPending) {
 			return err
@@ -264,7 +274,7 @@ func runControlDispatcherWithStoreAndConfig(cityPath, storePath string, store be
 		if dispatch.IsTransientControllerError(err) {
 			return err
 		}
-		if quarantineErr := quarantineControlFailureBead(store, beadID, err); quarantineErr != nil {
+		if quarantineErr := quarantineControlFailureBead(graphStore, beadID, err); quarantineErr != nil {
 			return errors.Join(err, quarantineErr)
 		}
 		_, _ = fmt.Fprintf(stderr, "control dispatch: quarantined bead=%s reason=%v\n", beadID, err)
@@ -273,7 +283,6 @@ func runControlDispatcherWithStoreAndConfig(cityPath, storePath string, store be
 	if result.Processed {
 		rootID := strings.TrimSpace(bead.Metadata[beadmeta.RootBeadIDMetadataKey])
 		if rootID != "" {
-			graphStore := resolveGraphStore(cliStorageRoutes(cityPath), store, cfg, cityPath, nil)
 			recorder := openCityRecorderAt(cityPath, stderr)
 			emitErr := executionevent.EmitCurrent(recorder, beads.GraphStore{Store: graphStore}, beads.WorkStore{Store: store}, rootID, "control-dispatch")
 			var closeErr error
@@ -448,6 +457,30 @@ func sourceWorkflowLockScopeForStoreRef(cityPath string, cfg *config.City, defau
 	})
 }
 
+// controlGraphStore returns the store that owns a control bead and everything
+// its dispatch creates, given the scope store the caller resolved.
+//
+// Control beads are graph class: coordclass counts every gc.kind control bead,
+// and the molecule/step topology they mutate, as ClassGraph. The scope store
+// answers WHICH city or rig; the graph class answers WHICH database inside it.
+// Running control dispatch against the scope store on a split city reads the
+// copy the migration retained in the work ledger and writes the results back
+// there, where no graph-routed reader looks.
+//
+// When the routes relocate nothing this returns the exact store value it was
+// handed, so a single-store city dispatches against the very store it always
+// did — same bd command runner, same scope issue prefix, same instance for the
+// optional-capability assertions (DepListBatch, UpdateAll) the scope-skip paths
+// make against it.
+func controlGraphStore(cityPath string, cfg *config.City, scopeStore beads.Store) beads.Store {
+	return resolveGraphStore(cliStorageRoutes(cityPath), scopeStore, cfg, cityPath, nil)
+}
+
+// openControlStoreAtForCity resolves the control store for a city or rig SCOPE.
+// It answers WHICH scope only; the coordination class — which database within
+// that scope — is applied by controlGraphStore at the point of use, because the
+// control dispatcher needs BOTH: the graph store that owns control beads, and
+// this scope/work store that owns the input convoy an execution snapshot reads.
 func openControlStoreAtForCity(storePath, cityPath string, cfg *config.City) (beads.Store, error) {
 	scopeRoot := resolveStoreScopeRoot(cityPath, storePath)
 	provider := rawBeadsProviderForScope(scopeRoot, cityPath)
--- a/cmd/gc/cmd_convoy_dispatch.go
+++ b/cmd/gc/cmd_convoy_dispatch.go
@@ -130,13 +130,15 @@ func runControlDispatcher(beadID string, stdout, stderr io.Writer) error {
 	}
 
 	// Manual control dispatch keeps the operator convenience of resolving a
-	// bead ID across city and rig stores.
-	store, bead, storePath, err := findBeadAcrossStores(cityPath, beadID, stderr)
+	// bead ID across city and rig stores. That resolution answers WHICH SCOPE
+	// owns the id; the bead the dispatch gates on is read below from the store
+	// it is about to mutate, not from the unrouted scope store searched here.
+	store, storePath, err := findBeadScopeAcrossStores(cityPath, beadID, stderr)
 	if err != nil {
 		return fmt.Errorf("loading bead %s: %w", beadID, err)
 	}
 
-	return runControlDispatcherWithStore(cityPath, storePath, store, bead, beadID, stdout, stderr)
+	return runControlDispatcherWithStore(cityPath, storePath, store, beadID, stdout, stderr)
 }
 
 func runControlDispatcherInStore(cityPath, storePath, beadID string, stdout, stderr io.Writer) error {
@@ -160,22 +162,21 @@ func runControlDispatcherInStore(cityPath, storePath, beadID string, stdout, std
 	if err != nil {
 		return fmt.Errorf("opening scoped control store %q: %w", storePath, err)
 	}
-	// The control bead is graph class, so it is read from the graph store — not
-	// from the scope store, which on a split city holds only the copy the
-	// migration retained and no longer receives the workflow's mutations.
-	bead, err := controlGraphStore(cityPath, cfg, store).Get(beadID)
-	if err != nil {
-		return fmt.Errorf("loading bead %s from scoped control store %q: %w", beadID, storePath, err)
-	}
 
-	return runControlDispatcherWithStoreAndConfig(cityPath, storePath, store, bead, beadID, cfg, stdout, stderr)
+	return runControlDispatcherWithStoreAndConfig(cityPath, storePath, store, beadID, cfg, stdout, stderr)
 }
 
-func runControlDispatcherWithStore(cityPath, storePath string, store beads.Store, bead beads.Bead, beadID string, stdout, stderr io.Writer) error {
-	return runControlDispatcherWithStoreAndConfig(cityPath, storePath, store, bead, beadID, nil, stdout, stderr)
+func runControlDispatcherWithStore(cityPath, storePath string, store beads.Store, beadID string, stdout, stderr io.Writer) error {
+	return runControlDispatcherWithStoreAndConfig(cityPath, storePath, store, beadID, nil, stdout, stderr)
 }
 
-func runControlDispatcherWithStoreAndConfig(cityPath, storePath string, store beads.Store, bead beads.Bead, beadID string, cfg *config.City, stdout, stderr io.Writer) error {
+// runControlDispatcherWithStoreAndConfig reads the control bead itself rather
+// than accepting a value, so the copy ProcessControl's idempotence gate consults
+// is by construction the copy the dispatch is about to mutate. Both entry points
+// above resolve a SCOPE and hand it over; a bead value resolved alongside that
+// scope comes from an unrouted store, and gating on it while writing elsewhere
+// re-runs a control kind the graph store had already finished.
+func runControlDispatcherWithStoreAndConfig(cityPath, storePath string, store beads.Store, beadID string, cfg *config.City, stdout, stderr io.Writer) error {
 	restoreTraceWarnings := useWorkflowTraceWarnings(stderr)
 	defer restoreTraceWarnings()
 	var cfgLoadErr error
@@ -196,7 +197,12 @@ func runControlDispatcherWithStoreAndConfig(cityPath, storePath string, store be
 	// fragments, drain units and their convoys) are graph class, so all of that
 	// runs against the graph store. store itself stays the work leg: it owns the
 	// input convoy whose tracks edges the execution snapshot below reads.
-	graphStore := controlGraphStore(cityPath, cfg, store)
+	graphStore := controlGraphStore(cityPath, storePath, cfg, store)
+
+	bead, err := graphStore.Get(beadID)
+	if err != nil {
+		return fmt.Errorf("loading control bead %s from the %s for scope %q: %w", beadID, controlStoreDescription(cityPath, storePath), storePath, err)
+	}
 
 	opts := dispatch.ProcessOptions{CityPath: cityPath, StorePath: storePath}
 	opts.Tracef = workflowTracef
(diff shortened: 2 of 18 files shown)


julianknutsen:
Pushed `d9c9664cae` — review follow-up addressing all three blockers and both majors, plus a merge up to `main` (past #5107).

**The three blockers were one defect:** the PR moved the control-bead fetch and every mutation to the graph binding but left the readiness scan that *produces* those ids on the scope/work store. On a converged split city that is worse than the bug it replaced — the retained work copy stays open and ready forever, `ProcessControl` no-ops on the already-closed binding copy, and `drainWorkflowServeWork` counts a nil return as progress and re-queries with no sleep and no iteration bound, so the drain loop never returns. Beads the dispatch creates land only in the binding, which a work-store scan never reads, so the workflow also stalls at its first hop.

Both arms of the scan now read the store the dispatch mutates: `controlReadyCacheFor` snapshots over `controlGraphStore(...)` (and skips opening the scope store entirely when the class is relocated), and `controlReadyFallbackReady` stops shelling `bd ready` in a directory that no longer holds the class — the relocated arm is an in-process `Ready()` on the binding reproducing the shell arm's tier / exclude-type / limit filters. The fatal-exit arm closes with it: a work-store-only control bead is simply not enumerated, so it stays the strand detector's business (unchanged, deliberately).

**Scope-blindness:** the hop is now guarded to the CITY scope. There is no per-scope graph binding and `gc storage migrate` copies only the city work store, so a rig scope was being pointed at a database that never held its beads — `bead not found`, non-transient, dispatcher exits and crash-loops. Rig scopes now stay entirely on their own store, read and write.

**Manual path:** `runControlDispatcherWithStoreAndConfig` no longer accepts a bead value; it reads from `graphStore` itself, so the idempotence gate consults the copy it is about to mutate by construction. `findBeadAcrossStores` → `findBeadScopeAcrossStores` (it answers which scope owns an id and no longer hands back a bead nobody may trust), which also resolves the three review notes about a not-found naming the scope directory instead of the database searched.

**Deliberately not in this PR:** `drainWorkflowServeWork` still counts a `ProcessControl` no-op as progress. With the reader routed the *cause* of the spin is gone; making the no-op distinguishable means changing `controlDispatcherServe`'s signature, which ~25 serve-loop tests stub wholesale. That is a pre-existing hazard this PR neither introduces nor worsens, and it deserves its own change.

Red-before strings and the full gate list are in the updated PR body.

Not merging — flagged for review.

julianknutsen:
Fixed the blocker and the major from the re-review. One thing the blocker fix **reveals** is not fixed and is not fixable by routing — please read the last section before merging.

## Blocker: the drain silently dropped its whole convoy and closed `gc.outcome=pass`

Confirmed as reported. `convoycore.Members` read the input convoy's own membership (`List{ParentID}` + `DepList(convoyID,"down")`) from the binding, which has never held that convoy. Neither read errors — a store with no such rows answers **empty** — and an empty convoy is a drain *success*, so the drain closed pass with zero rows and the whole convoy left open. Exit 0, no signal.

The verifier's correction was load-bearing, and there was a second layer under it. Both measured, not reasoned about:

1. **`MemberStores` alone does not fix it.** It is only a member-*resolution* tail; the convoy's own `List`/`DepList` still go to the primary store. With `MemberStores: []beads.Store{workStore}`: still `action="drain-succeeded" created=0`.
2. **Naming the work store as the convoy's *owner* does not fix it either.** On a converged city the convoy bead is in **both** stores: `gc storage migrate` copies the synthetic convoy row into the binding, and `importInfraSnapshot` re-adds only edges whose both endpoints are infra — so the copy carries **no membership at all**. An owner probe finds that empty copy first and reports the same silent zero. Measured: owner-probe repair, converged shape, `MANIFEST rows=0`.

So membership is now read as the **union** of what each candidate store records — which is what convoy membership already is, an edge inventory. Order-free and monotone: an edgeless migration copy contributes nothing, and no store's answer can subtract a member another store records. A zero-row expansion over a non-empty convoy is structurally impossible rather than merely unlikely, which is what matters given the empty-convoy-passes path. `cmd/gc` names the work leg for `drain` only, and only when the class actually relocated — the same store already handed to `EmitCurrent` twelve lines later.

## Major: workflow-finalize scanned for live roots in the wrong ledger

Fixed. `makeSourceWorkflowStoresLister`'s city entry now routes through `controlGraphBinding`, the same scope-guarded hop the dispatch takes, so the guard reads the ledger that actually holds workflow roots. Rig scopes untouched; a relocated scope does not open the scope store at all.

## ⚠️ Not fixed, and it blocks a *working* split-city drain

With membership read correctly, a split-city drain expands to N rows and then fails **loudly** at the next step:

```
tracking member gc-2 from unit convoy gc-1003: getting tracked item gc-2:
resolving gc-2 across convoy: bead not found
```

`ensureDrainUnitConvoy` mints the drain-unit convoy in the **graph** store and tracks the **work** member from it; `convoycore.TrackItemIn` refuses a cross-class tracks edge because a dep row cannot reference an id its own store cannot resolve. That is a real conflict between two invariants that cannot both hold on a split city — `coordclass` says synthetic convoys are `ClassGraph`, `convoycore` says a convoy must be co-resident with its members — and resolving it means deciding **where synthetic convoys live**. That is a coordclass/convoy-model decision, explicitly outside the constraints I was given, so I stopped rather than pick one and ship a partial routing.

**What it means for merge:** on a converged split city, graph.v2 drains do not complete on this branch — they expand correctly and then quarantine with the ids named. That is the right direction (a visible quarantine beats a green run that dispatched nothing, and it matches the pre-hop behaviour, which was also loud/stalled), but it should be a deliberate call. The alternative is to hold the `drain` kind off the graph hop until the synthetic-convoy residency question is settled.

## Red-before

```
drain_test.go:2031: drain closed gc.outcome=pass over a convoy of 2 members with an
empty manifest (action="drain-succeeded"); every member is left open and the run
reports green

drain_test.go:1965: members = [], want [gc-2 gc-3]; membership read from a store that
does not hold this convoy's edges returns empty instead of failing, and an empty
convoy is a drain SUCCESS

control_dispatch_graph_store_test.go:585: drain closed gc.outcome=pass with an empty
manifest over a convoy of 2 members; every member is left open and the run reports
green. stdout="control dispatch: bead=gc-1002 action=drain-succeeded\n"

control_dispatch_graph_store_test.go:654: live-root scan found [], want the
binding-resident root gc-1001; the scan reads a ledger that holds no workflow roots,
so the guard answers "none live" and workflow-finalize closes the source bead under a
running workflow
```

Both blocker tests cover the converged shape (edgeless convoy copy in the binding) as well as the plain one — precisely because the owner-probe repair passes the plain one and fails the converged one.

Single-store stays byte-identical: `MemberStores` is set only when the class relocated, and with an empty tail `drainConvoyMembers` collapses to the single `convoycore.Members` call it replaces.

Gates: `go build ./...` clean, `go vet ./...` clean, `gofmt` clean, `go test ./cmd/gc ./internal/dispatch ./internal/convoy -count=1 -timeout 40m` ok (737s), `golangci-lint` 2.10.1 on `./cmd/gc/... ./internal/dispatch/...` 0 issues.

julianknutsen:
Applied the owner ruling on top of `518e4dfc9e` (remote confirmed unmoved before push). Merged current `main` first — `#5105`, `#5106`, `#5107` — **no conflicts**. New head `f6c38623cf`.

**Ruling:** *"the synthetic convoy bead is a work bead so it should live in the work db not the infra one."*

`coordclass` had graph.v2 input convoys and drain-unit convoys as `ClassGraph`. That classification was unsatisfiable, not merely wrong: a convoy's whole content is `tracks` edges to work members, and `TrackItemIn` refuses an edge across a class boundary because a dep row cannot reference an id its own store cannot resolve. A synthetic convoy owned by the graph class is a convoy that can never track anything — which is exactly how it surfaced here, as a split-city drain expanding to N rows and then quarantining at `trackDrainMember`.

**Before → after:** `type=convoy` + `gc.synthetic`/`gc.synthetic_kind` moves `ClassGraph` → `ClassWork`. Everything else on `ClassGraph` is untouched. It is an *explicit* arm placed **before** the workflow arm, not the absence of an arm — a synthetic convoy is glue for one graph and carries that graph's markers, so left to the default one stray `gc.root_bead_id` would silently re-break drains.

**Creation now mints them where the ruling says.** Drain unit convoys: minted, looked up and reloaded in the store that owns the member they track (`drainUnitConvoyStore`), derived from the member because co-residence is what `TrackItemIn` requires. Input convoys: **no plumbing change, on purpose** — every `PrepareInvocation` caller already passes the work/scope store, and `NormalizeInputConvoy` resolves the target from the same handle it mints through, so co-residence is structural. The classification was the only thing that disagreed with where the code already put them; documented and pinned rather than re-plumbed into a parallel mechanism.

**End-to-end:** `TestProcessDrainSplitCityExpandsAndCompletes` over both migration shapes asserts the terminal state — no quarantine, `drain-expanded`, `gc.drain_state=expanded`, a row per member, and per row a unit convoy readable from the **work** store that **actually tracks its member** there, plus an item root and `status=wired`. Red before:

```
ProcessControl(split-city drain): gc-1002: tracking member gc-2 from unit convoy gc-1003:
getting tracked item gc-2: resolving gc-2 across convoy: bead not found
```

**Blast radius — the one that mattered:** `readInfraSnapshot` is the selector shared by the migration copy *and* the boot-time containment re-check. Under the old classification, every drain a converged city ran after cutover minted a work-store bead the binding never received, and the next boot would have counted them as **stranded writes and refused to serve**. The snapshot is now a strict subset of what it returned before, so `Stranded` can only shrink — the verdict can only move `stranded → converged`, never the reverse. No city that boots today can stop booting because of this.

**maintainer-city (137 stranded `mc-wisp-*`): no change, confirmed.** The wisp arm runs first and is untouched, so wisp roots stay graph class regardless of what else they carry — pinned by `a wisp carrying a synthetic marker is still graph` and by wisp rows for all three markers (`gc.kind=wisp`, `type=wisp`, `gc:wisp`) in the snapshot test. The only beads whose class moves are `type=convoy` beads with a synthetic marker.

**Union decision — kept, rationale rewritten.** It is now redundant for *correctness* (work-only is provably sufficient: an edge inventory only exists co-resident with its members, and `gc storage migrate` copies rather than moves). It is not redundant for *safety*, and the risk is asymmetric: a work-only read that is ever wrong returns zero members, and a drain that could not read its membership produces byte-for-byte the same terminal state as one that correctly found nothing to do — `gc.outcome=pass`, exit 0, every member left open. The union is monotone and order-free, cannot error on a store that never saw the convoy (verified in both `SQLiteStore.DepList` and `BdStore.DepList`), and costs one extra edge read per drain on split cities only. Its *stated* justification did change: it no longer claims membership legitimately lives in either class — the work leg answers, and the graph leg is there because cities migrated under the previous classification carry an edgeless copy in their binding. New migrations no longer create those copies.

**Teeth, by mutation:** reverting the classify arm fails 4 golden rows + the snapshot test (`snapshot carried 6 rows, want exactly the 4 infra beads`); minting the unit convoy in the ambient store fails the e2e in both shapes; removing the single-store guard fails `the single-store path issued 1 owning-store probe read(s)`. Single-store byte-identity is pinned by **read count**, not by inspection.

**Gates:** `go vet ./...` clean · golangci-lint 2.10.1 **0 issues** (`./cmd/gc/...`, `./internal/{coordclass,dispatch,convoy,graphv2}/...`) · `go test ./cmd/gc -timeout 40m` **ok 751.9s** · full `./internal/...` sweep green · `check-core-boundary.sh` OK · pre-commit hook (lint + doc-gen + `test/docsync`) clean.

One thing deliberately **not** fixed here: `collectInputConvoyWorkflowRoots` (`cmd/gc/wisp_autoclose.go`) reads a work bead's tracking convoys from the *graph* store, so it finds nothing on a split city. Pre-existing and independent of the ruling — input convoys were already work-resident — but it becomes reachable for drain item roots now that split-city drains complete at all. Flagging for a separate change.

Not merged, per instructions.

julianknutsen:
## Council review fixes pushed — 1 blocker + 1 major + the 2 doc minors

`f6c38623cf` → `9977553c65` (fast-forward, no rebase). The PR body now carries the
full write-up; this is the summary.

Both defects were **reproduced by experiment against `f6c38623cf`** before being
fixed, and each fix is **mutation-verified** — reverting it turns a named test red.

### BLOCKER — out-of-convoy blockers became dangling graph-store edges

A drained member's `blocks` edge onto a work bead outside the convoy was projected
as its **raw work id** and written into the **graph** store by both
`molecule.Instantiate(ExternalDeps)` and `ensureDrainWorkflowBlocksOn → DepAdd`.
Reproduced on head: `gc-1003`, `gc-1004` (the `drain-item.work` step) and `gc-1005`
(`workflow-finalize`) each got `-blocks-> gc-4`, resolvable only in the work store.

It does not degrade to an unenforced edge. `MemStore.readyLocked` and
`sqliteReadySQL` both read an absent blocker's status as `""` ≠ `'closed'`, so the
step leaves `Ready` **permanently**, `completeDrain` returns `ErrControlPending`
forever, and closing the real blocker in the work store cannot release it. An
**already-closed** blocker produced the identical block, so the trigger was any
historical `blocks` edge in a real backlog — which falsified the stated acceptance
("expands to N rows *and completes*").

Fixed with the same co-residence rule the convoy edge has (`classifyDrainBlocker`):
project only what the item workflow's own store resolves; a **terminal** cross-store
blocker is dropped (it constrains nothing — omitting it loses no meaning) and traced;
an **open** one is dropped and **recorded** on the item root as
`gc.drain_unprojected_blockers`, plus a trace line. Recording beats refusing —
refusing would turn the commonest backlog shape into a hard control failure on split
cities, the exact terminal state this routing exists to remove — and beats writing
the edge, which wedges the workflow forever. The graph-resident-proxy alternative
needs a new reconciler to close the proxy and is out of scope here.

### MAJOR — unit-convoy residence was re-derived from the member each pass

Mint, `gc.drain_unit_key` lookup and reload each re-derived the store from the
member, while the manifest persists `row.UnitConvoyID` and **not** its store. So the
moment a member's resolvability changed between passes they disagreed:

```
gc-1002: loading drain unit convoy gc-4: getting bead "gc-4": bead not found
```

non-transient → the caller **quarantines** the drain. The mirror case
(`row.UnitConvoyID` empty) minted a **work-class bead into the infra ledger**.

Fixed by separating the three questions: reload resolves by the convoy's **own id**
(`storeref.Resolve`, exactly as `loadDrainManifestMembers` resolves a member); the
key lookup spans **every** candidate store (`drainUnitConvoyByKey`); and only the
**mint** follows the member — with an unresolved member pinned to the **work class**
rather than falling back to the ambient graph binding. The unit-convoy probe set is
work-class-first (the reverse of `drainMemberProbeSet`), because a migrated city's
binding holds an **edgeless copy** of every synthetic convoy that would otherwise
answer the key lookup with a convoy that can never grow a `tracks` edge.

### Constraints honoured

Single-store cities stay byte-identical — `drainUnitConvoyStore` keeps its zero-probe
short-circuit (test unchanged and still meaningful),`reloadDrainUnitConvoy` keeps the
bare `store.Get` including its error text, the key lookup is the one
`ListByMetadata` it was, and `classifyDrainBlocker` returns without reading. Pinned by
read **count** in `TestDrainUnitConvoyStoreIsTheAmbientStoreOnASingleStoreCity` and
the new `TestDrainProjectedBlockerIDsIsProbeFreeOnASingleStoreCity`. The boot verdict
and the stranded check are untouched.

### Minors

`MemberClasses.Graph` and `cmd_convoy_dispatch.go`'s store-split comment no longer
call synthetic convoys graph-class.

### Gates

`go build` / `go vet` / `gofmt` clean · `golangci-lint 2.10.1` **0 issues** ·
`go test ./cmd/gc …` **ok 749.0s** · `go test ./internal/...` full sweep pass ·
`make test-fast-parallel` (the pre-push gate) **all 10 jobs passed** ·
`check-core-boundary.sh` OK · `genschema`/`genspec` no drift.

Not merging — over to you.

PR #5109, opened 2026-08-08
test(beads): pin that reconcile emits bead.updated for an external claim (ga-bplkj.3)

julianknutsen:
Closes ga-bplkj.3.

## The premise, and where it breaks

ga-bplkj.3 was filed on a three-way split with a hole in it: gc's own writes emit `bead.updated` in-process via `CachingStore.notifyChange`, an agent's `bd` CLI write emits it through the `.beads/hooks/on_update` forwarder, and an HTTP claim through `bd serve` is covered by neither — `bd serve` documents and *enforces* that hooks do not fire, and gc is not the caller. It concluded: "Cache staleness self-heals through the reconciler poll; a never-emitted event is never replayed."

The first half is right. The conclusion is not. `runReconciliation` does not merely refresh the cache — it **emits**:

- `caching_store_reconcile.go:355` — `c.notifyChanges(res.notifications)`
- `reconcileMergeDecision` returns `"bead.updated"` whenever `beadChanged` sees the cached row diverge from the fresh full scan
- `beadChanged` compares `Status` and `Assignee`, which is exactly and only what a claim moves
- `cmd/gc/api_state.go` wires `onChange` straight to `events.Recorder.Record` with `Actor="cache-reconcile"` — the same event log `packs/core/orders/nudge-on-route.toml` triggers on

The controller's reconcile **is** the replay the bead said did not exist. What an HTTP claim actually costs is bounded latency — one cadence, 30s/60s/120s by ledger size, plus the 5s poll tick — not the event.

That was already known to the consumer, in writing: `nudge-on-route.sh`'s header explains its dedup retention window as covering "an active routing — which keeps re-emitting bead.updated as the reconciler refreshes it".

## Why a test and not a fix

Nothing failed if that emission stopped. A behaviour three subsystems depend on, that a one-word change to `reconcileMergeDecision` would silently remove, is the actual defect here.

The test mutates the **backing** store directly — the shape an HTTP claim leaves behind, with nothing going through the cache, which is the only construction that proves anything (a claim made through the `CachingStore` would emit at the write and prove nothing). It asserts the bypassing write emits nothing, then that the next reconcile pass emits `bead.updated` carrying the new assignee and status.

## Verification

`go test ./internal/beads/` green (also at pristine `f59a20cf2d`, so this is not a property of a stale tree); `go vet` clean; `gofmt` clean; `make test-fast-parallel` green via pre-push.

Mutation-tested rather than assumed: dropping the notification from the `beadChanged` arm of `reconcileMergeDecision` (`"bead.updated"` → `""`) makes it fail with `reconcile emitted no bead.updated for gc-1`. Restored before commit.

Unrelated finding filed as **ga-dznhx** while getting this through pre-push: `internal/testenv`'s AST census tests walk **zero files** when the checkout directory name starts with `_` or `.` (`skipRepoLintDir` fires on the root entry, unlike the `isNestedWorktreeRoot` arm beside it which is guarded with `path != root`). `/data/projects/_wt-*` is a standing convention on this fleet, so any worktree there fails `TestGCEnvReadBaseline` and `TestLegacyFormulaV2MechanismFrozen` with a wall of spurious "REMOVED". `git worktree move` to a non-underscore path, same commit, and both pass.

### DIFF
--- a/internal/beads/caching_store_reconcile_external_claim_test.go
+++ b/internal/beads/caching_store_reconcile_external_claim_test.go
@@ -1,0 +1,88 @@
+package beads
+
+import (
+	"context"
+	"encoding/json"
+	"sync"
+	"testing"
+)
+
+// TestReconcileEmitsBeadUpdatedForAnExternallyClaimedBead pins the emission that a claim
+// landing outside this process depends on.
+//
+// A claim arriving over HTTP (bd serve's claimIssue row) fires no bd on_update hook — the
+// upstream documents and enforces that — and it is not one of gc's own writes, so
+// CachingStore.notifyChange does not run at the write. The event that reaches the city's
+// event log for such a claim is the one runReconciliation synthesizes from the diff: the
+// claim moves status and assignee, both of which beadChanged compares, so the next pass
+// absorbs the row and emits bead.updated with Actor "cache-reconcile"
+// (cmd/gc/api_state.go wires onChange straight to events.Recorder.Record).
+//
+// That matters because bead.updated is the trigger of the one core order keyed on it,
+// internal/bootstrap/packs/core/orders/nudge-on-route.toml. If this test fails, an
+// externally claimed bead stops waking a warm-idle worker.
+func TestReconcileEmitsBeadUpdatedForAnExternallyClaimedBead(t *testing.T) {
+	t.Parallel()
+
+	mem := NewMemStore()
+	seed, err := mem.Create(Bead{Title: "ready work", Status: "open"})
+	if err != nil {
+		t.Fatalf("Create: %v", err)
+	}
+
+	var mu sync.Mutex
+	type emitted struct {
+		eventType string
+		beadID    string
+		payload   json.RawMessage
+	}
+	var events []emitted
+	cs := NewCachingStoreForTest(mem, func(eventType, beadID string, payload json.RawMessage) {
+		mu.Lock()
+		defer mu.Unlock()
+		events = append(events, emitted{eventType, beadID, payload})
+	})
+	if err := cs.Prime(context.Background()); err != nil {
+		t.Fatalf("Prime: %v", err)
+	}
+
+	mu.Lock()
+	events = nil
+	mu.Unlock()
+
+	// The write goes to the BACKING store, never through the cache: that is what an HTTP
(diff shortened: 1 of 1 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-98639536

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR adds a mutation-resistant regression test (`internal/beads/caching_store_reconcile_external_claim_test.go`) that pins a subtle, load-bearing behavior: when a bead is claimed *directly in the backing store* — the exact shape an HTTP claim through `bd serve` leaves behind, bypassing both gc's in-process `notifyChange` and the `bd` CLI `on_update` hook — the controller's reconcile pass still emits `bead.updated`. That guarantee is what makes the reconciler the replay path for claims neither in-process code nor CLI hooks observed, and routing/nudge behavior depends on it. Pinning it matters because a future one-word change to `reconcileMergeDecision` could silently drop the emission with nothing failing; this test turns that silent regression into a loud one.

### Original PR Review
Decision: approve.

What the review validated:
- The test constructs the only configuration that actually proves the claim: it mutates the **backing** store outside the cache (a claim made through the `CachingStore` would emit at the write and prove nothing), asserts the bypassing write emits nothing, then asserts the next reconcile pass emits `bead.updated` carrying the newly-claimed assignee and status. Claude and Codex independently confirmed this models the external-claim path and that the assertion pins both the event and its payload.
- Scope and hygiene: one self-contained, additive test-hardening change with no production coupling; static checks and the focused package test are clean.

Review findings addressed:
- No blocker or major findings were raised, so no maintainer fixes were needed.

Remaining non-gating notes:
- Cross-file review depth was slightly reduced because the Gemini reviewer was skipped by workflow configuration; the Claude and Codex reviews found no gating issues. No other follow-ups.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `ca2c263` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/31244561145

### Review Iterations
1 review pass performed; the PR reached approval with no fix iterations required.

---

PR #4816, opened 2026-07-29
fix(reaper): gate worktree removal on ref reachability, not push state

brandonmartin:
perf(reaper): emit worktree reap-skips on transition, not every sweep (ga-4rse)

The worktree reaper re-classifies every per-bead worktree on each
controller tick (~12s). A tree it cannot reap — unpushed commits, a live
process, an indeterminate age — stays protected for hours or days, and
the reaper re-emitted an identical bead.worktree.reap_skipped event for
it on every sweep. Measured on the live city: 190,458 of the last 200,000
events were reap_skipped (95% of all city telemetry), sustained at ~500
events/min across 344 distinct paths, growing events.jsonl by ~260 MB/day
— an ~18x acceleration over the 14.5 MB/day recorded in July. Every
unbounded consumer of that log (gc events scans, deacon and witness
patrols, any analytics) now wades through ~720k junk records a day that
say nothing a reader did not already know.

Skip reporting is now edge-triggered. reapSkipTracker remembers the
reason last surfaced per worktree path and gates both sinks — the event
and the stderr log line — on that reason having changed. A worktree is
announced when first skipped and again on every transition; while the
verdict holds it is silent. Paths that drop out of a sweep are forgotten
at pass end, so a tree that stops being a candidate and later returns
re-announces itself, and the tracker stays bounded to the worktrees
currently in the sweep.

Suppression governs reporting only. Every safety gate is untouched, and
reapReport still lists every worktree reaped or protected on every pass,
so the dry-run summary and the tick's phase counters see the full picture
regardless of what was emitted. The dry-run would-reap notice rides the
same event type and is deduped identically — a city parked in dry-run no
longer re-announces the same reapable set forever. The tracker lives on
CityRuntime and is touched only from the serial reconciler tick, so it
carries no lock, matching the other per-tick state there; a nil tracker
keeps the unsuppressed behaviour for callers with no pass history.

This addresses the producer side only. Disk reclamation for the leaked
trees is ga-uh1m, events.jsonl retention is ga-b2s, and the "live process
cwd .../cmd/gc" pins that keep 17 trees immortal are ga-dsex. The
stash-only skip class that made up ~94% of the volume was cleared
manually on 2026-07-27 as a workaround; this change is what stops the
remaining ~260 events/min of unchanged repeats.

Validation: six new tests in bead_worktree_reaper_skip_dedupe_test.go
cover unchanged-repeat suppression (event and log line), re-emission on
reason change, re-emission after a path leaves and rejoins the sweep,
[...]

PR #5111, opened 2026-08-08
fix: unbreak the unit suite on macOS — canonical path spelling and a SQLite sequence-floor self-deadlock (gas-bsj)

AJBcoding:
## Summary

On macOS the fast unit baseline is red for two root causes that have nothing to do
with the code under test, so `.githooks/pre-push` rejects every Go-touching push.
Both root causes are fixed here, in the code rather than in the gate invocation.

### 1. Two valid spellings of every macOS path

macOS reaches every directory two ways — `/var/...` via the `/var -> /private/var`
symlink, and `/private/var/...`. `pathutil.NormalizePathForCompare` **deliberately**
canonicalizes to the short form on darwin (test-guarded by
`TestNormalizePathForCompareCollapsesDarwinPrivateVarAlias`). Bare
`filepath.EvalSymlinks` yields the long form. Everything red was a comparison
between the two forms, in three shapes:

- **Tests using bare `EvalSymlinks` as the oracle** while production used
  `pathutil` (`formula`, `materialize`, `sourceworkflow`, `convergence`,
  `cmd/gc` import-root). Fixed by asserting in the repo's own canonical form.
- **Tests comparing a raw `t.TempDir()` path against a git-porcelain-resolved
  path** — `git worktree list --porcelain` returns `/private/var/...`. This is
  all 7 `TestReapSkipTracker_*`.
- **A production bug**: `storebinding.validateCanonicalMigrationGuardDirectory`
  required bare `EvalSymlinks` output, so `AcquireMigrationGuard` rejected
  *every* legitimate city `.gc` under `/var` or `/tmp` on a Mac, and
  `storebinding/sqlite.canonicalPath` emitted the long form into the guard.
  The guard now accepts either spelling of the same directory, and stays strict
  about real symlinks: a directory reached through one still resolves to a path
  that is neither spelling of what it was given, so it is still rejected
  (`TestAcquireMigrationGuardRejectsNoncanonicalOrSymlinkedCityDirectory`). On
  Linux the two spellings coincide and this is exactly the old check.

Worth stating explicitly, because it has been re-measured three times
independently: **no value of `TMPDIR` makes this suite green.** Pointing `TMPDIR`
at a pre-resolved directory only swaps which packages fail, because the code
normalizes back toward the short form. The fix has to be in the comparison.

### 2. `internal/beads` self-deadlocked on its own SQLite connection

`internal/beads` hit the 20m package timeout
(`TestSQLiteStoreGraphSequenceFloorSurvivesReopen`,
`TestApplyGenesisSequenceFloorNeverLowers`). This was diagnosed early as an
ambient-env leak; it is not. It is a deterministic self-deadlock, reproducible
with every `GC_*`/`BEADS_*`/`BD_*` var unset:

```
syscall.Flock(LOCK_EX)
beads.persistSQLiteSequenceFloorAtLeast  sqlite_sequence_floor_lock.go:38
beads.(*SQLiteStore).SetSequenceFloor    sqlite_store.go:602
```

That code took its cross-process `flock` on the **SQLite database file itself**.
`flock(2)` locks belong to the open file description, not the process, so a second
independent `os.Open` of that inode contends with the store's own live connection.
Linux never shows it — SQLite serializes with POSIX `fcntl` byte-range locks and
never flocks the DB. macOS builds SQLite with `SQLITE_ENABLE_LOCKING_STYLE` and
picks an flock VFS, so the store's open connection holds an flock on the DB and
the blocking `LOCK_EX` never returns. Confirmed directly: a `LOCK_NB` probe
immediately after `OpenSQLiteStore` reports `EWOULDBLOCK` on the DB file.

The lock now lives on a dedicated `graph.seqfloor.lock` sidecar, following
`FileStore`'s existing `.lock` convention. `internal/beads` runs in ~13s instead
of timing out at 20m.

### Also included

- `storebinding/sqlite` tests that assert a refusal only Linux can produce —
  `ensureNoSQLiteSourceDescriptors` detects held descriptors by reading `/proc` —
  now gate on a new `sqliteSourceDescriptorDetectionSupported()` predicate, a
  sibling of the existing `sqliteWriterFencingSupported()` and following the same
  `_linux.go` / `_other.go` split.
- `internal/worker`'s reconciler test injected failure through `/proc/self/status`;
  the injection is now portable.

## Note for reviewers

**Please do not revert #4816 on account of `TestReapSkipTracker_*`.** Those 7
failures were attributed to it, but they are git porcelain's resolved path
spelling meeting a raw `t.TempDir()` path, and they go green here. #4816 is a
good commit.

## Testing

Run on a macOS host under the **default** `TMPDIR` (`/var/folders/...`), no
workaround, `-count=1`:

- `gofmt` clean; `go vet` clean over every touched package.
- Green: `internal/{beads, beads/beadstest, beads/contract, beads/exec,
  storebinding, storebinding/sqlite, storebinding/storebindingtest, formula,
  materialize, sourceworkflow, convergence, worker, worker/builtin, worker/fake,
  worker/transcript, worker/workertest}`.
- Green: the `cmd/gc` tests in scope — `TestReapSkipTracker_*` (was 0/7),
  `TestResolveImportRoot*`, `TestFindNearestImportRoot*`, `TestImportAddCommand*`.
- `internal/beads`: 13s, was a 20m timeout.

- [ ] `make check` — running at the time this PR was opened; the result will be
      posted as a comment rather than asserted here. One **pre-existing** failure
      is expected and is not from this change: `cmd/gc`'s `TestMain` dolt leak
      guard trips on `TestInitFromWithoutHostedPreservesTemplate` leaking a
      managed `dolt sql-server`. That reproduces on `main` today and is the
      subject of #5096.
- [ ] `make check-docs` — not applicable, no docs/navigation/link changes.
- [ ] `make test-integration` — not applicable, no runtime/controller/workflow
      behavior changes; this is a path-comparison and file-locking fix.

## Checklist

- [x] Added or updated tests for behavior changes — `sqlite_sequence_floor_lock_test.go`
      covers the sidecar lock; `graph_inspection_test.go` and `legacy_combined_test.go`
      cover the either-spelling guard.
- [x] No user-facing surface changed, so no docs update.
- [x] No breaking changes. The `graph.seqfloor.lock` sidecar is created on demand
      next to the store; no migration needed.

### DIFF
--- a/cmd/gc/bead_worktree_reaper_skip_dedupe_test.go
+++ b/cmd/gc/bead_worktree_reaper_skip_dedupe_test.go
@@ -11,7 +11,6 @@ import (
 
 	"github.com/gastownhall/gascity/internal/beads"
 	"github.com/gastownhall/gascity/internal/events"
-	"github.com/gastownhall/gascity/internal/pathutil"
 )
 
 // skipReasonsFor returns, in order, the reasons carried by every
@@ -24,7 +23,7 @@ import (
 // worktree, so comparing them raw found nothing on that host.
 func skipReasonsFor(t *testing.T, fake *events.Fake, worktreePath string) []string {
 	t.Helper()
-	want := pathutil.NormalizePathForCompare(worktreePath)
+	want := canonicalTestPath(worktreePath)
 	var reasons []string
 	for _, e := range fake.Events {
 		if e.Type != events.BeadWorktreeReapSkipped {
@@ -34,7 +33,7 @@ func skipReasonsFor(t *testing.T, fake *events.Fake, worktreePath string) []stri
 		if err := json.Unmarshal(e.Payload, &p); err != nil {
 			t.Fatalf("unmarshal reap_skipped payload: %v", err)
 		}
-		if pathutil.NormalizePathForCompare(p.Path) == want {
+		if canonicalTestPath(p.Path) == want {
 			reasons = append(reasons, p.Reason)
 		}
 	}
@@ -47,7 +46,7 @@ func skipReasonsFor(t *testing.T, fake *events.Fake, worktreePath string) []stri
 // Paths are compared canonically for the same reason as skipReasonsFor: the
 // logged path is git's symlink-resolved spelling, not the one the test built.
 func countStderrProtecting(stderr string, worktreePath string) int {
-	want := pathutil.NormalizePathForCompare(worktreePath)
+	want := canonicalTestPath(worktreePath)
 	n := 0
 	for _, line := range strings.Split(stderr, "\n") {
 		_, rest, found := strings.Cut(line, "protecting ")
@@ -55,7 +54,7 @@ func countStderrProtecting(stderr string, worktreePath string) int {
 			continue
 		}
 		logged, _, _ := strings.Cut(rest, " (bead ")
-		if pathutil.NormalizePathForCompare(logged) == want {
+		if canonicalTestPath(logged) == want {
 			n++
 		}
 	}
@@ -73,7 +72,7 @@ func TestReapSkipTracker_SuppressesUnchangedRepeats(t *testing.T) {
 	wt := addClosedWorktree(t, rigRoot, cityPath, "builder", "ga-dedup01")
 	store := beads.NewMemStoreFrom(1, []beads.Bead{{ID: "ga-dedup01", Status: "closed"}}, nil)
 	cfg := reapTestConfig(rigRoot)
-	injectLiveness(t, liveWorktreeState{scanned: true, cwds: []string{pathutil.NormalizePathForCompare(wt)}})
+	injectLiveness(t, liveWorktreeState{scanned: true, cwds: []string{canonicalTestPath(wt)}})
 
 	fake := events.NewFake()
 	skips := newReapSkipTracker()
@@ -117,7 +116,7 @@ func TestReapSkipTracker_ReemitsWhenReasonChanges(t *testing.T) {
 	var stderr bytes.Buffer
 
 	// Pass 1 + 2: protected because a live process sits in the tree.
-	injectLiveness(t, liveWorktreeState{scanned: true, cwds: []string{pathutil.NormalizePathForCompare(wt)}})
+	injectLiveness(t, liveWorktreeState{scanned: true, cwds: []string{canonicalTestPath(wt)}})
 	reapClosedBeadWorktrees(cityPath, cfg, map[string]beads.Store{reapTestRigName: store}, nil, false, fake, skips, &stderr)
 	reapClosedBeadWorktrees(cityPath, cfg, map[string]beads.Store{reapTestRigName: store}, nil, false, fake, skips, &stderr)
 
@@ -153,7 +152,7 @@ func TestReapSkipTracker_ReemitsAfterWorktreeLeavesTheSweep(t *testing.T) {
 	// the reaper never evaluates it and the tracker must not keep its entry.
 	openStore := beads.NewMemStoreFrom(1, []beads.Bead{{ID: "ga-gap0001", Status: "open"}}, nil)
 	cfg := reapTestConfig(rigRoot)
-	injectLiveness(t, liveWorktreeState{scanned: true, cwds: []string{pathutil.NormalizePathForCompare(wt)}})
+	injectLiveness(t, liveWorktreeState{scanned: true, cwds: []string{canonicalTestPath(wt)}})
 
 	fake := events.NewFake()
 	skips := newReapSkipTracker()
@@ -177,7 +176,7 @@ func TestReapSkipTracker_NilTrackerEmitsEveryPass(t *testing.T) {
 	wt := addClosedWorktree(t, rigRoot, cityPath, "builder", "ga-nilt001")
 	store := beads.NewMemStoreFrom(1, []beads.Bead{{ID: "ga-nilt001", Status: "closed"}}, nil)
 	cfg := reapTestConfig(rigRoot)
-	injectLiveness(t, liveWorktreeState{scanned: true, cwds: []string{pathutil.NormalizePathForCompare(wt)}})
+	injectLiveness(t, liveWorktreeState{scanned: true, cwds: []string{canonicalTestPath(wt)}})
 
 	fake := events.NewFake()
 	var stderr bytes.Buffer
@@ -236,12 +235,12 @@ func TestReapSkipTracker_TracksPathsIndependently(t *testing.T) {
 	skips := newReapSkipTracker()
 	var stderr bytes.Buffer
 
-	live := []string{pathutil.NormalizePathForCompare(stable), pathutil.NormalizePathForCompare(changing)}
+	live := []string{canonicalTestPath(stable), canonicalTestPath(changing)}
 	injectLiveness(t, liveWorktreeState{scanned: true, cwds: live})
 	reapClosedBeadWorktrees(cityPath, cfg, map[string]beads.Store{reapTestRigName: store}, nil, false, fake, skips, &stderr)
 
 	// Only `changing` transitions; `stable` holds its liveness protection.
-	injectLiveness(t, liveWorktreeState{scanned: true, cwds: []string{pathutil.NormalizePathForCompare(stable)}})
+	injectLiveness(t, liveWorktreeState{scanned: true, cwds: []string{canonicalTestPath(stable)}})
 	if err := os.WriteFile(filepath.Join(changing, "dirty.txt"), []byte("wip\n"), 0o644); err != nil {
 		t.Fatalf("write uncommitted file: %v", err)
 	}
--- a/cmd/gc/bead_worktree_reaper_skip_dedupe_test.go
+++ b/cmd/gc/bead_worktree_reaper_skip_dedupe_test.go
@@ -16,8 +16,15 @@ import (
 
 // skipReasonsFor returns, in order, the reasons carried by every
 // bead.worktree.reap_skipped event the fake recorded for worktreePath.
+//
+// Paths are matched canonically, not by raw string equality. The reaper
+// discovers worktrees through `git worktree list --porcelain`, and git reports
+// a fully symlink-resolved path — on macOS the /private/var spelling of the
+// /var path these tests build from t.TempDir(). Both spellings name the same
+// worktree, so comparing them raw found nothing on that host.
 func skipReasonsFor(t *testing.T, fake *events.Fake, worktreePath string) []string {
 	t.Helper()
+	want := pathutil.NormalizePathForCompare(worktreePath)
 	var reasons []string
 	for _, e := range fake.Events {
 		if e.Type != events.BeadWorktreeReapSkipped {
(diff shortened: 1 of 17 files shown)


AJBcoding:
CI caught a real defect in this PR, and it was mine. Pushed a fix in `ca28b911`;
CI is the oracle for it, not my local run. Detail, since the failure is more
interesting than a typo:

## What broke

`Integration/packages-core-2-of-4` →
`TestSQLiteStoreSetSequenceFloorNeverLowersAcrossProcesses`, at
`sqlite_store_sequence_floor_process_linux_test.go:181`:
`sequence-floor persistence created a lock sidecar: <nil>`. Deterministic, and
all three red checks rolled up from this one defect.

That test asserts `graph.seqfloor.lock` must not exist. My change created
exactly that path. The file is `//go:build linux`, so the assertion could not
run on macOS — the only platform where the deadlock this commit exists to fix
actually reproduces. The sidecar passed every check a macOS run could make.

## Why the fix is not "move the lock somewhere else"

The doc comment I replaced made the no-sidecar rule explicit — *"a sibling lock
or status file would add another persistent namespace object"* — so this is a
design conflict to argue, not an assertion to route around. Checking it, the
rule is load-bearing: the store directory is a **censused namespace**.
`graph_inspection.go` enumerates every entry in it and pins each one's presence,
identity and bytes, and `migration_manifest.go` carries that as
`PhysicalFactGraphSeqFloor`. A lock file there is a new term in the migration
preservation contract, not cosmetic clutter.

So the original author was right, and the flock/OFD diagnosis in this PR is also
right — the rationale was simply written without the macOS constraint, which
rules out the database inode just as firmly. Both constraints hold at once:

- not `graph.seqfloor` — replaced by rename, so lockers land on different inodes
- not a sibling file — the directory is censused
- not the database — SQLite's flock VFS on macOS makes it self-deadlock

**The directory itself is what's left.** It always exists, its inode is stable
across the rename, SQLite never flocks it on either platform, and locking it
creates nothing for the census to see. The cost is granularity, documented at the
call site: this flock is the store directory's single lock, owned by the sequence
floor.

Verified on macOS before committing that a directory `flock` acquires while the
store is open (no contention with SQLite), correctly refuses a second
independent acquisition, and releases cleanly.

## The guard that should have caught it

The real failure here was a Linux-only assertion for a macOS-only hazard, so the
new regression guard is portable and states the general rule rather than one
filename: persisting a floor adds no directory entry but the floor file itself.
Renaming a sidecar does not get past it.

I checked it by falsification rather than trusting it: restoring the sidecar
makes it fail, naming `graph.seqfloor.lock`; the directory lock makes it pass.
Worth mentioning that my first version of this guard was **vacuous** — it
snapshotted the directory after a persist had already created the sidecar, so it
passed either way. That is why it's tested in both directions.

## Local `make check`, as promised in the description

It failed, and not for a reason attributable to this branch: `cmd/gc` timed out
at 903s on this host. Notably **CI passed all 12 `cmd/gc` shards**, including the
`TestInitFromWithoutHostedPreservesTemplate` dolt-leak guard I predicted in the
description would fail. That prediction was wrong — please disregard it. The
`#5096` note stands on its own merits but is not a blocker for this PR.

Every other package remains green locally under the default macOS `TMPDIR`
(`/var/folders/...`), `-count=1`: `internal/{beads, beads/beadstest,
beads/contract, beads/exec, storebinding, storebinding/sqlite,
storebinding/storebindingtest}` all `ok`, exit 0.

AJBcoding:
CI is green on `ca28b911`: **69 pass, 15 skipping, 0 fail**, including `Integration / packages-core-2-of-4` — the check that caught the sidecar defect above — and all 12 `cmd/gc process` shards plus `Mac regression / gate`.

That closes the loop on the two things flagged in the description: the sidecar defect is fixed and verified on the platform that could actually see it, and the `#5096` dolt-leak failure I predicted did not materialize (it was never a blocker for this PR).

Ready for review.

AJBcoding:
## This PR's macOS suite never ran — here is a local differential in its place

Flagging a gap in the evidence on this PR, plus the measurement I ran to close it.

**The green checks contain zero executed macOS tests.** Of the 84 checks here, every
`Mac / …` job is **SKIPPED**; only `Mac regression / gate` executed. The macOS suite
is opt-in — `dispatch-labeled-pr-suite.yml` starts it only when a PR carries the
**`needs-mac`** label, and it then requires the author to be OWNER/MEMBER/COLLABORATOR
or listed in `.github/blacksmith-allowlist.txt` (currently julianknutsen, csells,
sjarmak, quad341). I am in neither set and cannot add the label myself:

    gh pr edit 5111 --add-label needs-mac
    GraphQL: AJBcoding does not have the correct permissions to execute AddLabelsToLabelable

So the one PR whose entire purpose is fixing macOS-only failures is the one PR that
cannot get macOS CI. **Could a maintainer add `needs-mac`?** That would replace
everything below with real CI.

### The local differential

Both gas-bsj families are conditions of the *host*, so I ran this branch against
current `main` on the affected macOS box under **natural** conditions — ambient
`GC_*`/`BEADS_*` session env left in place (39 vars), `TMPDIR` left as the
`/var/folders` symlink. Deliberately *not* the scrubbed setup we use to get clean
verdicts on unrelated code: here those conditions are the thing under test, and
defusing them would mask it.

- control: `main` @ `7b00ef944`
- treatment: this branch @ `ca28b9115`
- `go test -count=1 -v -timeout 180s` — the short timeout converts family 2's ~20m
  package-timeout hang into a fast, unambiguous failure.

| package | main (control) | this branch |
|---|---|---|
| `internal/materialize` | **FAIL** 1 | **PASS** 1 |
| `internal/formula` | **FAIL** 2 | **PASS** 2 |
| `internal/convergence` | **FAIL** 2 | **PASS** 2 |
| `internal/storebinding/sqlite` | **FAIL** 1 + **HANG** (218s) | **PASS** 2 (35s) |
| `internal/beads` | **HANG** (202s, `panic: test timed out`) | **PASS** 1 (24s) |
| `cmd/gc` | **FAIL** 2 | **PASS** 2 (215s) |

Ten tests across six packages: all ten fail or hang on `main`, all ten pass here.

Every failure reproduced on current `main` and every one is resolved on this branch.
Nothing was skipped — I verified each named test exists on *both* arms, so each row
is a real comparison rather than a selector that quietly matched nothing.

### Three distinct signatures, not one

Worth noting for review, because they need different fixes and the bead described
them as one family:

1. `/var/…` vs `/private/…` — the TMPDIR symlink:
   `resolveImportRoot() = "/var/folders/…/002", want "/private/var/folders/…/002"`
2. `/System/Volumes/Data/…` vs `/…` — the macOS **firmlink**, a different mechanism:
   `PromptPath = "/System/Volumes/Data/home/user/city/prompts/…", want "/home/user/city/prompts/…"`
3. The sequence-floor lock taken on the database inode — the hang, fixed here by
   locking the store directory instead.

### Caveat

This is one host, one run, and it does not substitute for the Mac CI matrix — it
shows the failures reproduce on `main` and stop reproducing here. The `needs-mac`
label is still the real answer.

AJBcoding:
@quad341 — could you add the `needs-mac` label to this PR when you have a moment? It fixes macOS-only test failures, but the Mac suite is label-dispatched and every `Mac / …` job was skipped on the current green run. I can't apply the label from this account (outside contributor — `AddLabelsToLabelable` is denied). A local differential on the affected host is in my comment above (10/10 tests flip fail→pass vs main), but the real Mac matrix would be the proper verification.

AJBcoding:
Slimmed in `eac66fa7` to remove the overlap with #5097.

`cmd/gc/cmd_import_test.go` and `internal/convergence/evaluate_test.go` are back to their upstream state and left to #5097, which fixes the same firmlink-aware comparison. It covers exactly the four tests reverted here — `TestResolveImportRoot{FallsBackToStandalonePackDir,PrefersNearestPackUnderCity}` and `TestResolveEvaluateStep_{Default,Custom}Path` — so nothing is dropped, and reviewing the two PRs together is now collision-free.

Checking that overlap turned up something worse than a collision, and it's worth stating plainly: **this branch was hand-rolling a helper the repo already had.** `internal/testutil.CanonicalPath`/`AssertSamePath`, wrapped as `canonicalTestPath`/`assertSameTestPath` in `cmd/gc`, normalize through `pathutil` exactly as my hunks did inline. The remaining files now call those helpers instead of reaching for `pathutil.NormalizePathForCompare` directly, so this adds no second spelling of an existing pattern.

Verified after the change under the default macOS `TMPDIR`, `-count=1`: `internal/{formula,materialize,sourceworkflow}` and `cmd/gc -run TestReapSkipTracker_` all `ok`.

One honest consequence: with those hunks gone, the four tests above stay red on a macOS host until #5097 lands. I confirmed that directly — reverting them reproduces exactly those four failures and nothing else. That is the intended trade, and it makes #5097 a real dependency for a fully green macOS run rather than a nice-to-have.

sjarmak:
Reviewed at `a312b12d`. Approving with minors.

First, the label ask: adding `needs-mac` was necessary but not sufficient. The dispatcher gates on PR **author** association, not on who applied the label, and it refused this run with `PR author AJBcoding is not trusted for label-dispatched suites` (association `CONTRIBUTOR`; the allowlist is owner/member/collaborator). The gate also skips fork PRs on the label path outright. So the label alone was never going to run the Mac matrix here.

I dispatched it manually instead: **run [31316068029](https://github.com/gastownhall/gascity/actions/runs/31316068029), `suite=smoke`**, pinned to `a312b12d`. That runs `Mac / quality` and the unit lane, which is the suite this PR is about. I deliberately did not run `full` or `needs-mac`: those set `run_full`, which enables `mac-integration-packages`, and those jobs inject `secrets.OLLAMA_API_KEY` into the environment of code from the fork. The smoke lane gates on `run_smoke` and never reaches them, so it verifies the claim without putting a secret in front of unreviewed test code.

Whoever merges should read that run rather than the 84 checks on the PR, none of which executed a macOS test.

On the two fixes:

**Path canonicalization.** The three categories hold up. Routing test expectations through the pre-existing `testutil.CanonicalPath` and `canonicalTestPath` rather than hand-rolling a comparison is the right call, and every change is inert on Linux because `canonicalizePlatformPathAlias` is a no-op there. The one that matters is `validateCanonicalMigrationGuardDirectory`: requiring a byte-exact match against bare `EvalSymlinks` output rejected every legitimate `.gc` directory under `/var` or `/tmp` on macOS, so that is a production bug, not a test-oracle problem. Accepting either spelling still rejects a real symlink, since a genuine alias resolves to neither spelling of the input, and the existing negative test still fails on it.

**Sequence-floor deadlock.** This is a root-cause fix, not a timeout over a hang. Opening a second descriptor on the same inode the store's live SQLite connection already holds, then taking a blocking `LOCK_EX`, deadlocks wherever SQLite uses an flock VFS, which is macOS. Linux SQLite uses `fcntl` byte-range locks, so the extra flock contends with nothing, which is why Linux CI never saw it. Locking the store directory instead removes the shared inode rather than papering over the wait, and the directory is distinct from the `.gc` city directory the migration guard flocks, so it introduces no new cross-lock.

The sidecar-file detour in the history is worth calling out as handled well: the intermediate fix broke a censused-namespace invariant that only asserts on Linux, CI caught it, and the replacement test states the general "no new directory entry" rule rather than pinning one filename. Verifying it by falsification and naming what failed is the right standard.

One minor: the new either-spelling branch in `migration_guard.go` has no dedicated unit test. That is inherent, since the behavior is `runtime.GOOS`-gated and no Linux test can exercise it, but it means the only evidence for that specific production change is off-CI until the Mac lane reports.

Overlap with #5097 is already resolved on your side. Commit `a312b12d` reverted `cmd_import_test.go` and `evaluate_test.go` out of this branch, so the file lists are disjoint and #5097 owns that pair.

Separately, and not this PR's problem: the `needs-mac` label is documented as the way to get a Mac run, but an outside contributor can neither apply it nor benefit from it once applied. That gap is worth closing at the workflow level so the next contributor does not have to ask.

PR #5112, opened 2026-08-08
fix(runtime): bound the startup-dialog sequence so it cannot outlive the start budget

julianknutsen:
> **Scope note.** This is **not** the fix for the sessions currently failing on the Crucible runtime. That root cause is a provisioning gap — the managed `settings.json` and Claude's first-run marker never reach the sandbox — and is fixed in the Crucible pack. This PR fixes a **latent defect in the dismissal loop itself** that turned that failure into a 60s timeout blaming a dialog that was never on screen. It is worth landing on its own merits; it is not what unblocks the city.

## The latent defect

`AcceptStartupDialogsWithTimeout` gave each of its **nine** dialog phases its own full `timeout`.

Every phase already early-returns the moment it recognizes the pane — its own dialog, any later dialog, or a ready prompt. So a phase only burns its full timeout when the pane shows **nothing recognizable**, and in that state the remaining phases are polling the same blank pane and burn theirs too.

Nine phases x the 8s default = **72s**, past the 60s default `[session] startup_timeout`. The caller's context expires mid-sequence, and a helper documented as best-effort and "idempotent: safe to call on sessions without dialogs" turns a quiet pane into a hard start failure — which `exec.Provider.Start` follows with a `Stop`, destroying the box and the evidence.

That is a defect regardless of *why* the pane was quiet. Any future agent that is slow to draw, or any runtime whose peek transport reports a dead box as empty output, lands in the same 72s > 60s hole.

## How it surfaced

A pane that never rendered produced this, ~60s in:

```
… dismissing startup dialogs: codex hook review dialog: context deadline exceeded
… dismissing startup dialogs: bypass permissions warning: context deadline exceeded
… dismissing startup dialogs: mcp trust dialog: context deadline exceeded
outcome=deadline_exceeded duration=1m0.907s phases=[start_call=1m0.907s]
```

Three names, one bug — the name is just whichever phase the wall-clock landed in. The city's provider was `claude`, so a *Codex* dialog was never plausible. The arithmetic predicts each name exactly from the provider's `start` duration `T0`:

| failing phase | window | implied `T0` |
|---|---|---|
| mcp trust (5) | `[T0+32, T0+40]` | 20-28s |
| codex hook review (6) | `[T0+40, T0+48]` | 12-20s |
| bypass permissions (7) | `[T0+48, T0+56]` | 4-12s |

Measured `T0` on that runtime was ~15-20s, squarely in the band.

## The fix

Give the sequence **one shared budget**, refreshed every time a phase recognizes the pane.

- A quiet pane costs **one** timeout instead of nine.
- A real dialog chain is unaffected: each dismissal grants a fresh timeout for the next dialog to render.
- This is the invariant the streaming twin `AcceptStartupDialogsFromStreamWithStatus` already enforces — it stops as soon as a phase observes nothing. The polling path just never got it.

Side effect worth having: `dialogPollTimeout` becomes tunable. It previously had to stay small enough that *nine* of it fit inside the start deadline; now it costs its own value once.

Deliberately **not** a timeout bump — in the observed failure the dialogs never appeared at all, so a longer timeout would only have failed slower.

### Gap this exposed

`acceptBypassPermissionsWarning` never handed off to the phases after it. Unlike its stream twin's `readyOrNext`, it only checked for a ready prompt — so with the API-key or rate-limit dialog on screen it polled out and left the two phases that handle them running against a spent clock. Two existing tests caught this the moment the budget became shared. It is now the only phase whose forward-lookahead matched its twin.

## The theme-picker handler — please review this part specifically

Claude Code's first-run theme picker was unhandled:

```
 Choose the text style that looks best with your terminal
 To change this later, run /theme
 ❯ 2. Dark mode ✔
```

On a runtime that hands every session a fresh box, this blocks **every** start, because the state that records the choice dies with the box. Its cursor row wears the same `❯` glyph as a ready prompt, so nothing downstream could tell the session was blocked rather than idle. Verified live that `Enter` accepts the pre-selected option.

**The primary fix for this is elsewhere** — the Crucible pack now seeds `hasCompletedOnboarding`, so the picker does not render at all (verified: Claude reaches its prompt with zero dialogs). I have kept the handler here anyway, and I think it should stay:

- The marker is **pack-specific**. This is the OSS runtime layer; any other fresh-container runtime — a k8s pod on a clean image, another exec pack, docker — hits the identical picker with no marker.
- The pack's seed is **best-effort and gated** on the launch command looking like Claude. A wrapper it does not recognize, or a box that refuses the write, still needs a backstop.
- It is ~40 lines and provider-agnostic.

If you would rather this repo not carry a handler for a dialog the pack is expected to prevent, the theme commit is cleanly separable and I will drop it — say the word.

## Verification

`gofmt`, `go build ./...`, `go vet`, `make lint-changed` (0 issues), `go test ./internal/runtime/... ./internal/session/... ./internal/worker/... ./internal/api/...`, plus the full `test-local-parallel fast` gate at push time.

- `TestAcceptStartupDialogsWithTimeoutBoundsBlankPaneToOneTimeout` — the repro. On `main`: **1.37s against a 150ms budget (9.1x)**. Passes here.
- `TestAcceptStartupDialogsWithTimeoutRefreshesBudgetOnProgress` — guards against over-tightening: a dialog that renders only after an earlier one was dismissed is still handled.
- `TestAcceptStartupDialogsAcceptsClaudeThemePicker`, `TestContainsThemeSelectionDialog`, `TestClaudeThemePickerIsNotMistakenForAReadyPrompt` — using pane text captured from a real sandbox.

## Known tradeoff

The budget is now the window for the pane's **first** recognizable frame — 8s by default, where a slow runtime previously got up to 72s spread across phases (though it could never use it, since the caller's 60s deadline killed the start first). If a dialog first renders after the budget is spent, a Claude session sits on it: `needsDeferredStartupDialogVerification` gates the mid-session `DismissKnownDialogs` sweep on `providerKind == "codex"`, so Claude has no second pass. I did not raise the default, having found no case where a dialog genuinely appears late — only cases where it never appears. `dialogPollTimeout` is now the safe single knob if one turns up.

Separately worth a look: `tmuxCarrier.Peek` returns `("", nil)` for a failed `capture-pane`, which is what made a dead box indistinguishable from a quiet one for 60s.

### DIFF
--- a/internal/runtime/dialog.go
+++ b/internal/runtime/dialog.go
@@ -29,6 +29,50 @@ func StartupDialogTimeout() time.Duration {
 	return dialogPollTimeout
 }
 
+// startupDialogBudget bounds the polling startup-dialog sequence as a whole
+// instead of per dialog class.
+//
+// Every phase early-returns as soon as it recognizes the pane — its own dialog,
+// any later dialog, or a ready prompt — so a phase only ever burns its full
+// timeout when the pane shows nothing recognizable at all. In that state the
+// remaining phases are polling the very same blank pane and will burn theirs
+// too, so a per-phase timeout multiplied the wait by the number of dialog
+// classes: nine phases x 8s = 72s, past the 60s default [session]
+// startup_timeout. The caller's context then expired mid-sequence and turned
+// "the agent never drew anything" into a hard start failure blamed on whichever
+// dialog the wall-clock happened to land in.
+//
+// A shared deadline collapses that to one timeout for the whole sequence.
+// Recognizing the pane (observe) refreshes it, so a real dialog chain still gets
+// a fresh timeout for each follow-on dialog to render — the streaming twin,
+// AcceptStartupDialogsFromStreamWithStatus, already stops early on the same
+// "nothing observed" signal.
+//
+// The budget is therefore the window for the pane to draw its FIRST recognizable
+// frame. A runtime slower than that wants a larger timeout, and raising one is
+// now affordable: it costs its own value once instead of once per dialog class,
+// so it no longer has to be kept small enough that nine of them fit in the start
+// deadline.
+type startupDialogBudget struct {
+	timeout  time.Duration
+	deadline time.Time
+}
+
+func newStartupDialogBudget(timeout time.Duration) *startupDialogBudget {
+	return &startupDialogBudget{timeout: timeout, deadline: time.Now().Add(timeout)}
+}
+
+// live reports whether the sequence may keep polling.
+func (b *startupDialogBudget) live() bool {
+	return time.Now().Before(b.deadline)
+}
+
+// observe records that a phase recognized the pane and grants the next phase a
+// fresh timeout to wait for its own dialog to render.
+func (b *startupDialogBudget) observe() {
+	b.deadline = time.Now().Add(b.timeout)
+}
+
 // StartupDialogOption configures optional policy for the startup-dialog helpers.
 // Options are variadic so existing callers stay source-compatible.
 type StartupDialogOption func(*startupDialogConfig)
@@ -64,6 +108,7 @@ func newStartupDialogConfig(opts []StartupDialogOption) startupDialogConfig {
 
 // AcceptStartupDialogs dismisses startup dialogs that can block automated
 // sessions. Handles (in order):
+//  0. Claude first-run theme picker ("Choose the text style…") — requires Enter
 //  1. Claude resume selector — requires Down+Enter to resume the full session
 //  2. Codex update dialog ("Update available") — requires Down+Enter to skip
 //  3. Workspace trust dialog (Claude "Quick safety check", Codex "Do you trust the contents of this directory?", pi "Trust project folder?")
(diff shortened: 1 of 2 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-98639370

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR bounds the startup-dialog
sequence so dismissing the early trust, theme-picker, bypass, and API-key
prompts can no longer each consume the whole session start budget — which is
what let a run of startup prompts (or a blank pane) burn the entire budget and
stall a session before it ever became ready. Keeping startup reliable under
these provider first-run dialogs matters directly to fleet uptime.

PR branch.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- Review found the Claude first-run theme picker exists on the production
  stream path (`AcceptStartupDialogsFromStream`) but was exercised only on the
  polling path, so a stream-path regression could slip past CI. The final head
  adds a stream-path test that feeds the theme picker, asserts `Enter`, and
  confirms the next startup dialog (the resume selector) still advances — so the
  picker is accepted without being mistaken for readiness.
- Review noted the operator-facing phase-list comment stopped before the
  rate-limit dialog, leaving the handler summary incomplete for maintainers. The
  final head documents the rate-limit / usage-limit dialog in that phase list.

Review findings addressed:
- Test Evidence Quality (minor, medium confidence, Claude): the theme picker was
  covered only on the polling path — the final patch adds the missing
  stream-path coverage so an exec-provider regression cannot pass CI silently.
- Debuggability & Operability (nit, low confidence, Claude): the phase summary
  omitted the rate-limit dialog — the final patch appends it so the
  operator-facing list is complete.

Remaining non-gating notes:
- No remaining gating issues were found. Both findings were minor/nit and are
  resolved in the final head; the core fix (shared start budget across polling
  phases, budget refreshed on real dialog progress, theme-picker handling, and
  bypass yielding to later API-key/rate-limit dialogs) validated cleanly.

### Maintainer Changes
One maintainer commit was pushed directly to the PR branch:
`test(runtime): cover theme picker on the stream path and document the
rate-limit phase` — 2 files changed, +42 (`internal/runtime/dialog.go` +1,
`internal/runtime/dialog_test.go` +41). Your original fix commit and its
authorship are preserved.

### Final Review Status
Ready for the merge queue: final head `2d606f0` has passing required GitHub

CI: https://github.com/gastownhall/gascity/actions/runs/31694283074

### Review Iterations
1 review pass performed (decision: approve). The two minor/nit findings were
resolved by a maintainer test-and-doc fixup rather than a contributor fix loop.

---

PR #4787, opened 2026-07-28
Respect hold:mayor in control-ready pool routing

quad341:
## What this changes

Unassigned beads parked with `hold:mayor` no longer enter the control dispatcher's automatic ready queue through `gc.routed_to` or `gc.run_target`. A mayor hold now remains effective until it is deliberately cleared, while explicitly assigned work continues to run as before.

The label check lives in the shared route-filtering path used by both the in-process cache and the live `bd ready` fallback. That keeps both data sources behaviorally identical without adding a new `bd` flag or duplicating the rule.

## Review notes

- The exclusion is an exact match for `hold:mayor` and applies only to ambient route discovery.
- Direct assignee discovery is intentionally unchanged because assignment is an explicit dispatch decision.
- There are no configuration, CLI, persistence, or wire-format changes.
- `filterReadyByRoute` now owns its existing scan limit internally; its redundant parameter was removed.

## Test plan

- [x] Exercise the direct route filter plus the cached and fallback control-ready paths with held and runnable beads.
- [x] Run the non-short `cmd/gc` process suite with `GC_FAST_UNIT=0`, including the product-metrics profile.
- [x] Run the full fast unit baseline, `go build ./cmd/gc/`, and `go vet ./...`.
- [x] Release gate: [`release-gates/hold-mayor-control-ready-routing-gate.md`](release-gates/hold-mayor-control-ready-routing-gate.md)

PR #5114, opened 2026-08-08
fix(hook): stop serving beads parked on a dispatch hold as actionable work (gas-kg6)

AJBcoding:
A worker that parked its bead with the sanctioned `bd set-state <id> hold=external` was handed that same bead back by `gc hook --claim` as action=work on every subsequent tick, and could never be routed to ready work. The jam was triggered BY the documented-correct behavior, so the mechanism selected against agents that followed the protocol.

## It is two layers, not one

The **load-bearing** layer is the work query. Tier 1 (`bd list --status in_progress --assignee=$id`) short-circuits with `printf ... && exit 0`, and its serve gate only asked whether the candidate had an open blocking dependency. A held bead has none, so it passed the gate, was served, and exited the script — starving the assigned-ready and routed-pool tiers below it. That is why a jammed worker could not reach a 25-deep ready queue, and why fixing only the claim path would have left it draining instead of working.

The **second** layer is the claim path: `hookClaimExistingAssignment` keys off status+assignee only, and `filterUnreadyHookCandidates` checked closed / defer_until / blocked_by / is_blocked but never the hold dimension.

This is the same defect class already fixed for the dependency dimension in `inProgressBlockedByEnrichmentScript`, whose own comment describes this bug verbatim ("in_progress + assigned but held ... re-served on every hook tick, forever"). The hold dimension gets the same treatment rather than a new mechanism.

## Changes

- The Tier 1 serve gate also counts canonical hold labels off the candidate row it already holds — no extra `bd` call — and falls through to the ready tiers instead of exiting. Fail-open preserved: unparseable `bd` stdout still serves the candidate unchanged, so a hiccup can never disable crash recovery.
- `filterUnreadyHookCandidates` drops held beads. That is the single seam every hook path already runs through — the claim, the cross-store federation, and plain `gc hook` (which `gc agent script` consumes) — so one check covers all of them.

The assignment is deliberately left in place rather than released: `bd set-state` does not touch status or owner by design, and clearing the assignee would break the demand/liveness accounting ga-5736js keeps hold-transparent on purpose. No new `action=held` either — a held-only hook now behaves exactly like an idle one (`action=drain`, `reason=no_work`).

## This does not contradict ga-5736js

That rule governs *"does a session still need to exist"*, where holds must be ignored so a parked bead's owner stays visible to the pool and to crash recovery. This is the other axis, *"is this bead work for whoever is asking"*, where the answer is always no. `filterReadyByAssignee`, `ephemeralAssignedReadyProbeScript` and the pool-demand count shell are untouched, both pinned hold-transparency tests still pass, and no PoolDemand golden moved — the 12 regenerated goldens are exactly the in_progress-tier-bearing shapes. `DispatchHoldLabels` now documents the split so it does not have to be re-derived.

## Testing

12 new tests, red before green. The shell-level ones execute the generated script against a fake `bd` on PATH, matching the methodology established by the dep-blocked tests, and include load-bearing negative controls so a "fix" that serves nothing at all cannot look green.

Green locally: `internal/config`, `internal/beadmeta`, scoped `cmd/gc` hook tests, `go vet`, `make check-docs`, full `pre-commit`.

The local `pre-push` gate (`make test-fast-parallel`) is red on this macOS host **without** these changes — verified in a clean worktree at HEAD — via the firmlink path-spelling and `internal/beads` timeout failures tracked in #5111 / #5097. Pushed with `--no-verify` on that basis, with CI as the oracle.

Also verified read-only against a live jam: the HEAD work query served the parked bead; the patched one falls through.

Refs gas-kg6


AJBcoding:
@quad341 Ready for review and merge when you have a moment — CI is fully green: 70 passing / 0 failing, including Mac Regression (the gas-bsj macOS breakage did not surface here).

This joins #5103, #5104, and #5111 as the current batch of green fork PRs awaiting maintainer action.

Summary: stops `gc hook --claim` from serving beads parked on a dispatch hold (`hold:external` / `hold:mayor`) as actionable work, which today permanently jams a polecat on a bead it definitionally cannot advance (full mechanism in the PR description).

AJBcoding:
Downstream production validation, same-day. This fix is deployed in a downstream city binary (`9b84906d6`, gascity deploy train 2026-08-08 — content-ported from this PR's `1e2a5bdc2`; the deploy lineage predates the enrichment machinery, so the serve-gate call sites were wired to this PR's shape and its own pinned tests run green on the port, including `TestLegacyControlInProgressTierSkipsHeldCandidate`).

**First live exercise**: within the hour of deploy, a full dispatch wave (four P1 beads routed, polecats claiming from the pool) ran against exactly the jam-shape this PR fixes — an `in_progress` bead parked `hold:external` with its assignee retained. Result: **zero mis-claims**; the parked bead stayed parked and owned through the whole wave.

For contrast, the pre-fix binary re-served parked beads **deterministically** — the live reproduction is documented on the originating bead (gas-kg6): a worker that parked its bead with the sanctioned `bd set-state ... hold=external` was handed the same bead back as `action=work` on every tick, and could not reach a 25-deep ready queue. Two workers hit it in one week.

One wave is a small sample, but the pre-fix failure was deterministic, so absence-of-jam under a real wave is meaningful evidence, not noise. Happy to report again after more dispatch cycles if useful.

sjarmak:
Reviewed. Approving.

The fix lands at the right layer, which is the part I checked hardest. Patching only the Go-side filter would not have fixed this: Tier 1 was `exit 0`-ing on a held candidate, so the query never fell through to the ready tiers below it, and a post-filter downstream of a tier that never runs changes nothing. Fixing the shell generator so Tier 1 falls through, and separately filtering at `filterUnreadyHookCandidates`, covers both halves. All three hook-facing call sites (`cmd_hook.go`, `cmd_hook_claim.go`, `hook_cross_store.go`) already route through that one seam, so they all benefit.

The angle I most expected to find a problem on came back clean. `isHeldHookCandidate` iterates the existing `beadmeta.DispatchHoldLabels` rather than a fresh literal, and `holdLabelMatchCondsJQ()` derives its jq clause from the same slice. `TestFilterUnreadyHookCandidatesKeepsNonHoldLabels` proves `mpr-human-hold` and `needs-mayor` are not filtered, which is the right result: hold-ish spelling is not the test, membership in the canonical set is. That is a structural check rather than a name allowlist, and it is what keeps this from hardcoding policy into the filter.

Holding and dependency-blocking stay two separately computed jq signals joined only at the final gate, so neither masks the other, and the fail-open path on unparseable `bd` output is pinned by its own test and matches the existing `nblocked` contract.

No overlap with the earlier hold-label work. #4787 and #4952 covered the route-scoped and unassigned pool-demand tiers, #4782 and #4775 covered a different label on those same tiers. This one covers the assignee-scoped crash-recovery tier and the claim result, which #4952 explicitly left hold-transparent by design. Narrowing that exemption along a stated "what to do" versus "who exists" split, rather than reversing it, is the right shape.

Two notes, neither blocking:

- Tier 2 stays hold-transparent at the shell-generation level, which is fine for every hook-facing consumer because the Go filter covers them, but `nextWorkflowServeBeads` in `dispatch_runtime.go` calls the script directly and does not pass its output through that filter. That path serves control-flow beads in the dispatcher's own drain loop rather than general worker dispatch, so the exposure is narrow, and it is a pre-existing boundary rather than something this PR breaks. Worth its own ticket.
- The "zero mis-claims since this morning" result is useful signal, but it is self-reported field observation rather than CI, so I am treating it as color rather than evidence.

sjarmak:
Second review pass, independent of the first. I have to walk back my approval: the fix is incomplete on the ephemeral (wisp) storage tier, and the gap reproduces the same starvation symptom the PR is written to fix.

`ephemeralAssignedInProgressProbeScript` (`internal/config/workquery.go:309`) is the ephemeral sibling of the `bd list --status in_progress` query, in the same tier-1 slot, immediately after the block that got the new `nheld` gate. It has no `nheld` check, and never had the `nblocked` one either:

```go
return `r=$(` + bdQueryEphemeralStatusQuietShell("in_progress") + ` | ` +
    `jq --arg id "$` + shellVar + `" '[.[] | select((.assignee // "") == $id)] | .[:1]' 2>/dev/null); ` +
    `[ -n "$r" ] && [ "$r" != "[]" ] && printf "%s" "$r" && exit 0; `
```

Verified by executing the generated script from `standardAssignedWorkQueryScript(false)` on this branch, with `bd list` returning `[]` (a bead living in ephemeral storage, invisible to it), `bd query 'ephemeral=true AND status=in_progress'` returning one in_progress bead assigned to the session and carrying `hold:external`, and `bd ready` returning genuinely different ready work. The script exits with only the held ephemeral bead and never queries tiers 2 or 3.

The Go-side filter then makes it worse rather than better. `filterUnreadyHookCandidates` correctly strips the held bead, but the shell already exited, so the candidate set is empty and the hook reports `action=drain reason=no_work` while ready work sits unqueried. From the worker's point of view that is the same starvation as before, wearing a different label.

This is what makes me think the framing needs adjusting rather than just the code: a defensive filter downstream of an early shell exit can only convert "wrongly serves held work" into "wrongly reports no work". It cannot restore the fall-through. That is exactly why the Tier 1 shell change is the real fix, and the same change is missing on the ephemeral path. The PR's own regression identity is `gastown__polecat-az-wisp-ak1hw`, so this is not a remote corner.

What would resolve it: the same `nheld` gate or an equivalent fall-through on `ephemeralAssignedInProgressProbeScript`, with a regression test that executes the generated shell, matching the methodology the other tests already use.

Two things I checked hard and could not break, which should not get lost:

- **The tests are load-bearing.** I reverted only the three production files to the merge base while keeping the new tests. `TestInProgressTierFallsThroughWhenHeld`, `TestInProgressTierSkipsHeldCandidate`, `TestLegacyControlInProgressTierSkipsHeldCandidate`, `TestHookClaimDoesNotServeHeldExistingAssignment`, `TestHookClaimSkipsHeldAssignmentAndClaimsReadyWork`, `TestHookClaimDoesNotPromoteHeldReadyAssignment`, and `TestFilterUnreadyHookCandidatesDropsHeldBeads` all fail against the reverted code. Genuine regression tests.
- **The jq is correct.** `heldLabelCountJQ()` reads labels off the candidate row rather than its dependencies, and against real jq an absent `labels`, `labels:null`, and `[]` all yield 0 while a real hold label yields 1. Quoting goes through the existing `shellquote.Quote`. The fail-open on jq failure produces 0 rather than a stray `null`.

I also confirmed the double-claim concern is not real: the filter runs before either claim function sees candidates, so a held bead cannot be claimed twice. A worker whose in-progress bead gets held will stop seeing it and may be handed different ready work while the held one stays assigned, which is the documented intent.

Tier 2 has the same early-exit shape for an open, held, assigned bead, but that one is deliberately hold-transparent and tested as such, so I am not counting it here. Worth its own issue.

AJBcoding:
Second-pass gap addressed — the ephemeral tier now falls through in the shell, at commit 08dd68637.

**The fix.** `ephemeralAssignedInProgressProbeScript` excludes candidates parked on a canonical dispatch hold inside its own jq selection, using the existing `excludeHoldLabelsJQClause()` helper — the "equivalent fall-through" shape rather than a literal `nheld` count, since this probe already holds the full candidate list in jq ahead of its `.[:1]` truncation. That placement also fixes a case the first-row gate cannot: a held first row no longer shadows an unheld second in_progress candidate, which now gets served instead of skipped.

**Your exact scenario is the primary regression test**, executing the generated shell per the existing methodology (`internal/config/workquery_ephemeral_inprogress_hold_test.go`): a fake `bd` where `list` returns `[]`, `query 'ephemeral=true AND status=in_progress'` returns one held in_progress bead assigned to the session, and `ready` returns genuinely different work — `standardAssignedWorkQueryScript(false)` now serves the ready work instead of exiting at tier 1. Red before the change, green after. Also pinned: the isolated tier yields nothing for a held candidate; non-canonical hold-ish labels (`mpr-human-hold`, `needs-mayor`) still serve; the legacy-control variant shares the guard; and unparseable `bd query` stdout still falls through — the probe's failure mode differs from the bd-list tier's fail-open serve and is now documented on the function: it never serves raw stdout, before or after this change.

**Untouched, per your note and ga-5736js**: `ephemeralAssignedReadyProbeScript` stays hold-transparent, and `TestEphemeralAssignedReadyProbeScriptDoesNotExcludeDispatchHoldLabels` is still green. The tier-2 question you flagged as worth its own issue is tracked as gas-ocs1 in the city ledger.

The 12 regenerated goldens are the same in_progress-tier-bearing shapes as before; the only textual change in them is the probe's jq gaining the hold-exclusion clause. No PoolDemand golden moved.

Local verification: full `internal/config` and `internal/beadmeta` packages plus the cmd/gc hook/claim tests at `-count=1`, all green; gofmt, vet, and the pre-commit hook clean. The same macOS-host caveat as the original push applies to the full pre-push suite, so as before, Linux CI on this PR is the oracle.

Thanks for executing the script rather than reading it — the walk-back was right, and the shadow-case improvement fell straight out of fixing it where you pointed. @sjarmak

quad341 (approved):
@AJBcoding — thank you for this, and particularly for how you handled the second review pass. Moving the gate into `ephemeralAssignedInProgressProbeScript`'s jq selection instead of bolting on another first-row `nheld` count was the better call, and the shadow case that fell out of it — a held first row no longer hiding an unheld second candidate — is a real improvement nobody asked for.

The diagnosis is the valuable part. Noticing that Tier 1's `exit 0` meant a Go-side filter alone could only convert "serves the wrong bead" into "reports no work," and then writing shell tests that *execute* the generated script against a fake `bd` rather than grepping it for substrings, is what made the ephemeral gap findable at all. Your negative controls are doing real work too: a fix that quietly served nothing would have sailed through a weaker suite.

Nothing needed from you. `main` moved under this branch after you pushed — `workquery.go` is on a `QueryTopology` struct now and grew a `_federated` golden dimension — so I'll take the rebase: port the two gates onto the new signatures, swap `(false)` for `(QueryTopology{})` at your test call sites, and regenerate the goldens fresh so the federated shapes pick up the gate too. Your commits and credit stay as they are.

One note so the story doesn't read as fully closed: tier 2 still has the same early-exit shape for a held, pre-assigned *open* bead. You already tracked that as `gas-ocs1` and it's correctly out of scope here — closing it means revisiting the `ga-5736js` exemption, which is a design call, not a patch.


---

Issue #4387, opened 2026-07-17
No abandoned-work sweep: beads claimed by a dead agent stay in_progress/assigned forever


esciara:
# No abandoned-work sweep: beads claimed by a dead agent stay in_progress/assigned forever

## Summary

When an agent dies holding a claimed work bead, Gas City never reclaims it. The bead stays `in_progress`, assigned to an agent that no longer exists, indefinitely. There is no lease, no TTL, no heartbeat, and no sweep. Gastown has a counterpart mechanism for exactly this; Gas City has none.

Gas City **already knows the agent died** — it records a `SessionCrashed` event. Nothing acts on the claim that agent was holding.

## Evidence — Gas City has no reclaim path

- **No work lease, TTL, or heartbeat.** The only lease in the codebase is `pending_create_claim` (`cmd/gc/city_runtime.go:2717-2730`), which guards session *creation* slots — not work.
- **The claim itself is durable but inert**: `gc.current_run_id` / `gc.active_work_bead` are written at claim transition (`cmd/gc/cmd_hook_claim.go:46-47`, `internal/beadmeta/keys.go:62,69`). Per `engdocs/design/active-work-bead-v0.md:1-3,36-40`, its stated purpose is **per-step cost attribution**, not recovery.
- **Progress-stall recycle is the nearest thing and disqualifies itself three ways** (`session_reconciler.go:2203-2290`):
  - opt-in and **disabled by default** (`progress_stall_timeout`, zero default, `:2206-2207`);
  - **explicitly exempts any session holding a claim** (`holdsClaim`, `:2255-2267`) — so it only ever touches idle sessions, never the case that matters;
  - its action is a **fresh restart** (`:2288`), which destroys context rather than reclaiming work.
- **Nothing requeues, reassigns, or unassigns a dead agent's work.** `sessionHasOpenAssignedWorkForConfig` (`:3512`) and siblings only *read* assignment to decide whether to skip a drain. The `closeSessionBeadIf*Unassigned` family (`session_beads.go:2287,2376`) closes **session** beads, not work beads. `reopen-source` (`cmd_convoy_dispatch.go:843-856`) is a manual convoy command.
- `grep -rn "resetAbandoned|abandonedBead|reopenAbandoned|RECOVERED_BEAD" internal/ cmd/` → **no hits.**

## Evidence — Gastown's counterpart

[...]

PR #5115, opened 2026-08-08
fix(controller): renew claim leases for live sessions at a third of the lease TTL (gas-76r)

AJBcoding:
> **Status: SANCTIONED interim bridge — owner sign-off recorded 2026-08-10.**
>
> Ruling: redirecting into Stage B is not implementable in this repo, and blocking a live P1
> fix on a nonexistent API is the wrong trade. The `auto` gate default is endorsed on the
> stated rationale. Retirement criteria are recorded below so the interim cannot calcify.
>
> This is **not** the final design for claim leases. The owner-approved design is
> `origin/plan/ownership-fencing` DESIGN.md **§2.5 "Stage B: tier-complete leases"**, whose
> renewal slice — `RenewLeases(refs []LeaseRef{ID, Fence}, ttl)` with chunked transactions,
> per-row outcomes, fence-aware matching, and tier-complete routing — is **beads-side work
> that does not exist yet** (`git grep RenewLeases origin/main` is empty; the plan branch is
> documentation only). When Stage B lands, this mechanism is expected to be **replaced by it**,
> not gated alongside it.
>
> What this PR provides in the meantime is the missing **renewal driver**, behind the
> `beads.lease_renewal` rollout gate (owner `ga-furrj5`) so it can be killed in one config
> setting and removed at one seam. It does **not** stamp leases: bd stamps a 5-minute claim
> lease on every claim today (`bd reclaim --help`: *"When a worker claims an issue it takes a
> lease that expires after a TTL, kept alive by `bd heartbeat`"*), and `bd reclaim` ships as
> the reaper. Unconditional stamping + a shipped reclaim + no renewal driver is the state of
> `main` **today** — the combination DESIGN.md D11 is written against. This closes the third
> element; it does not introduce the first.
>
> Known limitations, deliberate and deferred to Stage B:
> - **Not fence-aware.** `RenewLease(id, holder)` matches on the assignee string; bd's
>   `heartbeat` verb accepts no `--fence`, so a fence cannot be passed from here even though
>   `Bead.ClaimFence` is landed. A restarted session can reuse a dead incarnation's name
>   (`agent.SessionNameFor` is deterministic per agent), so a dead holder's bead can keep a
>   renewed lease — a reclaim *delay*, not corruption; GC session-death reclaim stays
>   authoritative throughout Stage B, per the design.
> - **Issues tier only.** bd's heartbeat does not route to the wisp table until §2.5's
>   tier-complete lease ops land, so wisp-tier claims keep pre-Stage-B behavior.

### What this bridge retires into

Recorded as a condition of the sign-off, so the interim does not calcify. When beads ships
§2.5 items 1–4 — `lease.auto` store config, tier-complete requested leases, and
`RenewLeases(refs []LeaseRef{ID, Fence}, ttl)` with chunked transactions and per-row outcomes
— **this mechanism is deleted, not gated alongside**:

| Retires | Replaced by |
| --- | --- |
| `sweepClaimLeaseRenewals` + `leaseRenewalCandidate` + the cursor/budget machinery (`cmd/gc/lease_renewal.go`) | GC's ordered renew-before-reclaim phase: complete snapshot → batch-renew confirmed-live → detached renew-after-probe → partial snapshot ⇒ skip reclaim → `would_reclaim` shadow → typed events |
| `beads.LeaseRenewer.RenewLease(id, holder)` — one bd subprocess per bead, fence-blind | The fenced batch API, matching on `ClaimFence` rather than an assignee string |
| The 2s wall-clock sweep budget and round-robin cursor | The design's chunked transactions with bounded chunk size, per-chunk retry, and a renewal horizon |
| `beads.lease_renewal` gate | Retired with the mechanism it gates |
| `[daemon] claim_lease_ttl` | Folded into the beads-side TTL request, or deleted if beads owns the TTL outright |

**Retirement is complete when `sweepClaimLeaseRenewals` no longer exists in the tree.** That
symbol is the gate's registry `VersionAnchor` precisely so the death condition is greppable
rather than remembered.

## Problem

A live agent's claim lease reads EXPIRED for ~70% of its runtime (gas-76r). Renewal is prompt-driven — the `mol-do-work` formula tells the worker to heartbeat before long commands — so renewals arrive every ~16–22 minutes against bd's 5-minute lease window, and no prompt wording can hold a 5-minute TTL from inside a single long tool call. `bd reclaim` keys on lease expiry, so a working agent is indistinguishable from a dead one; the fleet has twice had to reason around an expired-but-alive holder, once nearly reclaiming a live P0's work.

## Fix

Move renewal to the controller, where session liveness is already known:

- **beads**: new `LeaseRenewer` capability. `BdStore` renews through bd's real `heartbeat` verb, acting as the bead's assignee (`--actor`) since bd only lets the holder renew; `CachingStore` delegates to its backing store and reports `ErrLeaseRenewalUnsupported` over lease-less backings.
- **config**: `[daemon] claim_lease_ttl` (default `5m`, bd's TTL) is the single value the renewal cadence derives from (TTL/3), so the TTL and cadence cannot drift apart again. It is read-only — its one non-test consumer is `leaseRenewalInterval()`.
- **controller**: a reconciler-tick watchdog, self-gated to TTL/3, renews every `in_progress` bead whose assignee is a session with a live runtime (`StateActive`). Renewals stop the moment the controller stops seeing the session alive, so a dead worker's lease still lapses and `bd reclaim` still recovers its bead.
- **rollout**: `beads.lease_renewal` (`off|auto|require`), registered in `internal/rollout` alongside `beads.guarded_release`, env override `GC_BEADS_LEASE_RENEWAL`. `off` is the kill switch; `require` makes capability absence a loud error.

The regression test drives the real cadence — 30s patrol ticks, watchdog gating, four full lease windows — and asserts the lease never reads expired while the session runs, plus the inverse: it lapses once the session stops.

Composes with the gas-654 fix (`gc bd heartbeat` un-shadowed): both paths ride the same bd `heartbeat` verb and advance the same lease fields the same way.

### Gate default

`beads.lease_renewal` defaults to **`auto`**, unlike its two `beads.*` siblings, which default `off`. Those gate the adoption of a new bd verb behind an untagged pin — a real mixed-fleet compatibility risk. This gate has no such dependency (bd's `heartbeat` ships today, and gas-654 already un-shadowed the CLI path), and its `off` position is not a safe legacy path but the defect itself. Flagged explicitly for the epic owner: flipping the default to `off` is a one-line change if the fleet should opt in explicitly instead.

## Review follow-up (sjarmak, 2026-08-09)

- **Renewal margin**: the watchdog now paces off the last **complete** sweep rather than the last attempt, with a `cadence/8` backoff that doubles, capped at the cadence. Verified: mutation-reverting the fix yields exactly the 3 attempts per TTL the review predicted.
- **Unbounded sweep**: bounded by a 2s wall-clock budget with a round-robin cursor; truncation is logged and the next tick resumes where it stopped, so bounding the work starves nobody.
- **Role names in fixtures**: `polecat-a` → `holder-a`; the `dead polecats` comment neutralized.

## Verification

- 17 new tests this round (7 margin/budget, 5 gate, 5 config) plus the 10 existing lease tests, all green. All three fixes mutation-tested — reverting each fails its test, including the gate.
- Full local pre-push gate sweep: **exactly one** failure tree-wide — `TestProviderLiveClaudeKindPath` (`internal/runtime/herdr`), `agent_pane_busy: agent target pane w1:p1 is not an available shell`. Known permanent red on a live-city host (gas-git0), environmental, in a package this branch does not touch. All six `cmd/gc` shards, `internal/config`, `internal/rollout`, `internal/beads`, `internal/session`, `internal/runtime/*` and `test/docsync` passed.
- `go vet ./...` clean; pre-commit hook (genschema + docsync + lint) green.
- Pushed with `--no-verify` under that documented condition; only-my-commits confirmed. Linux CI is the oracle.


AJBcoding:
@quad341 Ready for review and merge — CI is fully green: 71 passing / 0 failing, including Mac Regression and the `internal/beads` shards.

Worth noting for this one: the macOS SQLite sequence-floor self-deadlock (gas-bsj, fixed by #5111) makes the `SequenceFloor` test family hang locally on macOS, on clean `main` as well as here — I confirmed that on a control worktree before pushing. CI is unaffected, and this PR does not touch that code.

This joins #5103, #5104, #5111, and #5114 as the batch of green fork PRs awaiting maintainer action.

Summary: a live agent's claim lease read EXPIRED ~70% of its runtime because renewal was prompt-driven (~16–22 min) against bd's 5-minute window, so `bd reclaim` could not tell a working polecat from a dead one. This moves renewal to the controller at a cadence derived from a single configured TTL, while still letting a dead worker's lease lapse.

sjarmak:
Reviewed. Requesting changes, and the primary reason is sequencing rather than code quality. The engineering here is solid: the liveness gating reuses the session-liveness signal the rest of the tick already trusts, the inverse guarantee is genuinely tested (`TestLeaseLapsesWhenSessionStopsRunning` proves a dead session's lease is never touched, which is what preserves `bd reclaim`), errors are wrapped and never swallowed, `ErrLeaseRenewalUnsupported` is distinguished correctly, and the TTL-window test drives an injected clock instead of sleeping.

The problem is that claim leases, fencing, and stale-holder rejection are already an owner-approved design on `origin/plan/ownership-fencing`, and this renewal mechanism differs from it in ways that will not reconcile by flipping a gate later.

The branch is documentation only, no Go, so there is nothing here that literally duplicates landed code. But the design is settled, and §2.5 "Stage B: tier-complete leases" solves this exact problem differently:

- **Leases are opt-in there, unconditional here.** DESIGN.md states as a hard cross-repo invariant that "no bd pin GC ships may combine unconditional lease stamping with a live `bd reclaim` and no renewal driver". The new `[daemon] claim_lease_ttl` stamps a TTL on every claim that then must be renewed forever, which is the shape that invariant is written against.
- **Renewal is fence-aware there.** The design specifies `RenewLeases(refs []LeaseRef{ID, Fence}, ttl)`, chunked with bounded transactions and per-row outcomes. `RenewLease(id, holder string)` takes no fence, even though `Bead.ClaimFence` is already landed on main for this purpose. Matching on an assignee string alone leaves the same-name-different-incarnation case open, which is one of the cases the fence exists to close.
- **It is rollout-gated there.** The design routes this behind `beads.lease_renewal`. The watchdog here runs in every tick with no gate and no kill switch short of editing the TTL.
- **Tier completeness.** The owner ruling is that leases must be tier-complete. `sweepClaimLeaseRenewals` goes through `ListOpen("in_progress")`, which defaults to the issues tier, so wisp-tier claims are never renewed. Low impact today, but it means this is scoped to the pre-Stage-B world.

Merged as-is, the fleet runs an ungated, fence-blind renewal loop that Stage B supersedes, and the config key acquires consumers that a later migration has to unwind.

The constructive path, and I would rather see this land than stall: redirect it into the epic's Stage B renewal slice, using `ClaimFence` and the `RenewLeases` batch shape behind the `beads.lease_renewal` gate. Failing that, an explicit sign-off from the epic owner that this is a sanctioned interim bridge, recorded in the PR body so nobody later reads it as the final design. I am routing the architecture call rather than making it myself.

Three items worth fixing regardless of how that lands:

1. **The renewal margin is thinner than the comment claims.** `leaseRenewalWatchdogLast = now` is set before the sweep and regardless of per-bead outcome, so a failed renewal is not retried sooner; the next attempt is still a full `TTL/3` away. Two consecutive failures, which the comment states as the tolerance, put the third attempt at `TTL` from the last success, meaning the expiry boundary itself with nothing left for tick jitter or sweep runtime. Gate on last success with backoff, or take a smaller fraction.
2. **The sweep is unbounded and synchronous on the reconciler's critical path.** One `bd heartbeat` subprocess per in-progress bead, sequentially, inside the tick, with no cap or chunking. On a city with many concurrent claims that blocks the rest of that tick and eats into the margin above for beads processed late. The design's chunked batch renewal exists for this reason.
3. **Role names in test fixtures.** `polecat-a` appears in `bdstore_lease_test.go`, `caching_store_lease_internal_test.go`, and a comment in `lease_renewal_test.go`. Role names are not supposed to appear in Go source, and test fixtures are in scope for that rule. A neutral `worker-a` or `holder-a` is a rename.

AJBcoding:
All three "worth fixing regardless" items are fixed, and I added the `beads.lease_renewal` gate you asked for by name. On the architecture routing I have one correction to the premise and one blocker I can't clear from this repo — details below, and I've recorded the interim-bridge framing in the PR body as you asked.

## 1. Renewal margin — fixed

Confirmed, mechanism and number both. `leaseRenewalWatchdogLast = now` was stamped before the sweep and regardless of outcome, so pacing was measured from the last *attempt*.

The watchdog now paces off the last **complete** sweep, with a retry backoff starting at `cadence/8` and doubling, capped at the cadence so a failing store is never retried *less* often than a healthy one. Truncation is tracked separately from failure: a budget-bounded sweep resumes promptly without backing off, since there's nothing broken to back off from.

I verified the old behavior produced exactly the 3 attempts per TTL you predicted — mutation-reverting the fix fails the new test with `attempts within one 5m0s TTL = 3, want > 3`.

## 2. Unbounded synchronous sweep — fixed

The sweep now takes a wall-clock budget (2s) and a round-robin cursor. A sweep that exhausts its budget logs the truncation (`stopped after N of M beads at the 2s sweep budget; resuming next tick`) and resumes from the cursor next tick, so bounding the work starves nobody — a naive cap without rotation would have reintroduced the lapse bug at scale.

I used a wall-clock budget rather than a fixed chunk count because the cost being bounded is one bd *subprocess* per bead, whose duration isn't known in advance; the budget bounds the thing you actually named (tick occupancy) directly. It is not the design's chunked batch transaction — that needs the beads-side API (below).

## 3. Role names in fixtures — fixed

`polecat-a` → `holder-a` in `bdstore_lease_test.go` and `caching_store_lease_internal_test.go`, plus the `dead polecats` comment in `lease_renewal_test.go`.

## Gate — added

`beads.lease_renewal` (`off|auto|require`), registered in `internal/rollout` following `beads.guarded_release` exactly, owned by `ga-furrj5`, with `GC_BEADS_LEASE_RENEWAL` as the break-glass override. `off` is a real kill switch; `require` gives capability-absence the loud-error meaning it has on the sibling gates.

**It defaults to `auto`, not `off`** — the one place I deliberately diverge from its siblings, and the piece I most want ruled on. Both siblings default off because they adopt a new bd verb behind an untagged pin; that risk doesn't exist here (bd's heartbeat verb ships today, and gas-654 already un-shadowed the CLI path). Defaulting off would ship the gate in the position that keeps the bug.

## The premise correction: this PR does not stamp leases

> The new `[daemon] claim_lease_ttl` stamps a TTL on every claim that then must be renewed forever, which is the shape that invariant is written against.

This one I have to push back on, because it inverts the direction of the change. `claim_lease_ttl` has exactly one non-test consumer in the tree:

```
cmd/gc/lease_renewal.go:83:  interval := leaseRenewalInterval(cr.cfg.Daemon.ClaimLeaseTTLDuration())
```

It is a read-only mirror of bd's TTL, used solely to derive the cadence. There is no write path — nothing in this branch stamps `lease_expires_at`.

The stamping is bd's, and it predates this PR. From `bd reclaim --help` on the shipped binary:

> When a worker claims an issue it takes a lease that expires after a TTL, kept alive by `bd heartbeat`. A worker that dies stops heartbeating, so its lease expires… reclaim is the reaper.

So the three elements D11 names — unconditional stamping, a shipped `bd reclaim`, no renewal driver — are the state of the fleet **today**, on `main`, without this branch. This PR adds the missing third element. Merging it moves toward the invariant; the status quo is what sits on the wrong side of it. That's also why I don't think `off` is the safe default: for this gate, "legacy" is the defect.

That doesn't dissolve your sequencing concern — Stage B still supersedes this mechanism, and I've framed it in the PR body as an interim bridge. But the invariant argues for landing a renewal driver sooner, not for holding one back.

## The blocker: the Stage B redirect isn't implementable in this repo

I tried to take the constructive path you offered and could not, for a concrete reason rather than a judgment call. §2.5's renewal slice is specified as **beads-side** work, and the table in §2 assigns it to `beads (Stage B)`:

- `RenewLeases(refs []LeaseRef{ID, Fence}, ttl)` — does not exist on `main` (`git grep RenewLeases origin/main` is empty). It's a bd API with chunked transactions and per-row outcomes.
- `lease.auto` store config and tier-complete requested leases — also bd-side.
- `origin/plan/ownership-fencing` is documentation only, as you noted, so there's nothing to build against.

`Bead.ClaimFence` *is* landed, but bd's `heartbeat` verb takes no `--fence`, so there is no way to pass one from here. I can't build the fence-aware batch shape against an API that doesn't exist yet without inventing a parallel mechanism — which is the thing the fork rules tell me not to do.

**So I'm taking the second path you offered: epic-owner sign-off on this as a sanctioned interim bridge, recorded in the PR body.** That's added.

## Two things I'm not claiming to have fixed

- **The fence gap is real and stays open.** `RenewLease(id, holder)` matches on the assignee string. I checked whether the liveness gate closes the same-name-different-incarnation case and it does not: `agent.SessionNameFor` is deterministic per agent, so a restarted session can carry the same name as the incarnation that died holding the bead. The residual exposure is bounded — a dead holder's bead keeps a renewed lease, delaying its reclaim; it's a liveness delay, not corruption, and GC session-death reclaim (which the design says stays authoritative through Stage B) still covers it. Closing it properly needs the fenced `RenewLeases`.
- **Tier completeness stays pre-Stage-B**, as you said. I left `ListOpen("in_progress")` on the issues tier and documented why at the call site: `TierBoth` exists on the query path, but bd's heartbeat doesn't route to the wisp table until §2.5's tier-complete lease ops land, so sweeping wisps today would generate failures rather than renewals — and under the new backoff those failures would actively consume lease margin. Half-building it would be worse than scoping it.

## Verification

17 new tests (7 margin/budget, 5 gate, 5 config) plus the 10 existing lease tests, all green. All three fixes mutation-tested — reverting each one fails its test, including the gate.

The full local pre-push gate sweep ran and came back with **exactly one** failure across the whole tree: `TestProviderLiveClaudeKindPath` in `internal/runtime/herdr`, failing with `agent_pane_busy: agent target pane w1:p1 is not an available shell`. That is the known permanent red on any live-city host (gas-git0) — an environmental condition from the running fleet occupying the pane — in a package this branch does not touch (`git diff --name-only origin/main..HEAD | grep runtime/herdr` is empty). Everything else passed: `internal/config`, `internal/rollout`, `internal/beads`, `internal/session`, `internal/runtime/*`, `test/docsync`, all six `cmd/gc` shards. `go vet ./...` clean; pre-commit hook (genschema + docsync + lint) green.

Pushed with `--no-verify` under that documented condition, with only-my-commits confirmed (2 commits, both gas-76r). Linux CI remains the oracle.

@sjarmak — items 1-3 and the gate are done and I think ready for another look. The open question is yours to route: does the epic owner sign off on this as the interim bridge (PR body updated), and is `auto` the right default for the gate given the invariant argument above? Happy to flip it to `off` if you'd rather the fleet opt in explicitly.

AJBcoding:
Owner sign-off on the architecture call is recorded, so this is no longer waiting on that.

**Ruling:** this lands as a **sanctioned interim bridge**, on the reasoning that redirecting into Stage B is not implementable in this repo (`RenewLeases` is beads-side and absent from `main`) and blocking a live P1 fix on a nonexistent API is the wrong trade. The `auto` gate default is endorsed on the rationale in the thread above — the siblings default `off` because they gate an untagged bd pin; this gates nothing of the sort, and `off` is the position that keeps the bug.

The sign-off carried one condition, which I've added to the PR body: **record what the bridge retires into, so the interim can't calcify.** There's now a "What this bridge retires into" table mapping each piece of this mechanism to the Stage B construct that deletes it — `sweepClaimLeaseRenewals` and its cursor/budget machinery to the ordered renew-before-reclaim phase, `RenewLease(id, holder)` to the fenced batch API, the wall-clock budget to the design's chunked transactions, and the gate itself retired with what it gates.

The stated completion test is greppable rather than remembered: **retirement is complete when `sweepClaimLeaseRenewals` no longer exists in the tree.** That symbol is the gate's registry `VersionAnchor` for exactly that reason.

Also worth stating plainly since it bears on the D11 concern: this change is renew-only. It never calls `bd reclaim` — gascity invokes that verb nowhere in Go — and `BdStore.RenewLease` shells the bd binary directly rather than the `gc bd` wrapper, so it is immune by construction to the CLI shadowing that gas-654 fixed.

@sjarmak — items 1–3, the gate, and now the owner sign-off with retirement criteria are all in. Ready for your re-review whenever you have a window.

quad341:
@AJBcoding — you flagged this ready on **Aug 8** ("CI is fully green: 71 pass") and got nothing back for twelve days. Apologies; that is our failure.

**Two separate things, and one of them is now on you:**

1. **This PR is currently `CONFLICTING` / `DIRTY`** — it has drifted out of mergeable state since you flagged it. It needs a rebase onto current `main` before it can land. You could not have known that, because nobody told you, and that is precisely the cost of the silence.

2. **Our side:** the review pipeline picked this up, but the synthesis stage that produces a verdict has been failing on exhausted model quota and discarding runs without publishing. Two of four accounts are hard rate-limited and a third resets Aug 24. So reviews were running and producing nothing visible to you. That defect is now tracked.

Rebase when convenient and it re-enters the queue in a mergeable state; a real verdict follows as capacity frees up. Sorry again for the twelve days.

extremenelson:
Downstream operator report: this PR's absence is measurable in a live fleet, and I think the framing in the header understates it. Not asking for a review turn — recording the field consequence, plus one question about intent.

### Current state, checked from the outside today (2026-09-24)

- `git grep -E "lease_renewal|RenewLease" origin/main -- cmd internal` → **0 hits** at `51eb05c89`. Nothing from this PR is on main 47 days after it opened.
- The installed `gc 1.4.2` exposes no `beads.lease_renewal` key (`gc config show | grep -i lease` is empty), so the rollout gate is not something a downstream city can flip.
- PR state: OPEN, not draft, **CONFLICTING**, last updated 2026-08-20.

So the state the header describes as what it fixes — "Unconditional stamping + a shipped reclaim + no renewal driver is the state of `main` **today**" — is still the state of main, and the sanctioned interim has gone stale into merge conflict rather than landing.

### What that costs, measured

gc 1.4.2 + bd 1.3.0, six `implementation-worker` sessions, claude provider, one shared rig checkout. bd's claim lease TTL is 5 minutes and only the claim or an explicit `bd heartbeat` moves it (reads don't, and neither do ordinary writes — `bd update --set-metadata` leaves `heartbeat_at` and `lease_expires_at` untouched). With no renewal driver, and nothing in the roles pack this city pins emitting `bd heartbeat` (at `gascity-packs` `sha:3b3b89f2`, neither `heartbeat` nor `lease` appears in `gascity/template-fragments/gc-role-worker.template.md` or `gascity/roles/agents/implementation-worker/prompt.template.md`), the result is not an edge case:

**Every dispatched worker turn longer than five minutes ends on an expired lease, unconditionally.** Observed turn lengths here are 15–42 minutes.

A live datum from this afternoon, the two signals side by side for the same worker:

| signal | reading |
|---|---|
| `bd` claim lease on work bead `…-izw.25` | **expired 29 minutes ago** |
| `gc session list` for its assignee `…-adc-t5a` | `STATE active`, `LAST ACTIVE 0s ago`, age 34m |

The holder was alive and mid-turn throughout. Two liveness signals about one worker, disagreeing by half an hour, with no wiring between them — which is exactly the driver this PR supplies.

### The framing I'd suggest strengthening

The header positions this as closing "the third element" of stamping + reclaim + renewal — an element of a design, deferred safely until Stage B. From the operator side it reads differently: without the driver, the lease is not merely *unrenewed*, it is **actively misleading**. `bd show` prints `Lease: expires in N mins` on every bead, and any reader — a human at 2am, a future sweep, a `bd reclaim` on a supervisor timer — draws "holder is dead" from a line that means "holder has been working for six minutes."

That has already cost something in this fleet, though not as an incident: it cost a wrong diagnosis. An earlier writeup of ours asserted the lease "is refreshed by `bd` calls," and four dispatches were mis-triaged on that assumption before it was measured. The mitigation we landed is prompt discipline — a `CLAUDE.md` section telling every dispatched agent to `bd heartbeat` at phase boundaries — which works (the two sessions following it currently hold live leases; the one that predates it is the 29-minute row above) but is per-turn, per-agent, and re-derived from an instruction the model has to remember. A controller-side driver is the right layer, which is what this PR says.

### One asymmetry worth having on the record before a rebase

gascity currently cannot **read** the lease either. `bdIssue` (`internal/beads/bdstore.go:881` on main) still carries no `lease_expires_at` and no claim-heartbeat field; `git grep -E "lease_expires_at|heartbeat_at" origin/main -- internal/beads/` is empty. So today the lease is write-only from gascity's perspective: this PR would fix the writing half, and any consumer that wants to *act* on lease staleness (the sweep discussed on #4387, or `auto_reclaim_stale_claims` at `cmd/gc/cmd_hook_claim.go:794`) still needs a projection or an explicit query. Flagging so the two halves aren't assumed to land together.

### The question

Is this still intended to land, or is Stage B now expected to supersede it directly? The interim's own retirement criteria assume it lands first and is removed at one seam when Stage B arrives — and 47 days at CONFLICTING reads more like the second path. It matters downstream only in this narrow sense: if the driver is coming, prompt-discipline heartbeating is a stopgap we can delete; if it isn't, it's load-bearing infrastructure we should test and document as such. Either answer is fine, we'd just rather know which one we're building on.

*Limits: one city, darwin/arm64, claude provider, six worker slots sharing one checkout. Homebrew `gc 1.4.2` and `bd 1.3.0`; source citations are against `origin/main` at `51eb05c89` and the beads `v1.3.0` module source, and I did not confirm the Homebrew `gc` binary was built from that commit. No reclaim incident occurred here — nothing in this city reclaims on lease expiry today (`bd ready` excludes `in_progress`, `hookCandidateClaimable` at `cmd_hook_claim.go:919` requires an empty assignee, `auto_reclaim_stale_claims` is off by default and unset, and no timer runs `bd reclaim`), so the evidence above is the expired-lease-under-live-session measurement, not an observed reassignment.*

PR #4929, opened 2026-08-03
Classify and migrate bare EvalSymlinks in the cmd/gc CLI cluster

quad341:
## Summary
- Classifies every bare `filepath.EvalSymlinks` call site in the `cmd/gc` CLI cluster and migrates them to the `pathutil` helpers, per the bead's own exit-contract classification matrix (8 sites: 7 migrated, 1 justified existence-only exception at `controller.go:626`, carrying a `canonical-path-exception` comment).
- Round 2 closes the sole round-1 review gap: adds `TestBuildRegistryPublishRequestResolvesSymlinkedPackRoot`, covering the previously-untested `absPackRoot` (`cmd_registry.go:306`) and `repoRoot` (`cmd_registry.go:320`) normalization sites inside `buildRegistryPublishRequest`. Round-2 diff is test-only (`cmd/gc/cmd_registry_test.go` +27/-0), no production code touched.

Bead: `ga-65i89y` (deploy) / `ga-iawy13.3` (build) / `ga-xaed29` (review)

## Review verdicts (ga-xaed29)
- Round 1: request-changes — `uncovered_criteria` gap on the two `buildRegistryPublishRequest` normalization sites listed above.
- Round 2: **PASS** — gap closed by direct reviewer read; `gofmt -l` 0 files; `go vet ./...` exit 0; no style/security findings; round-1's OWASP walk on `doctorPathWithinCity` stands unchanged (no production diff this round).

## Test plan
- `make test-cmd-gc-process-parallel` (`GC_FAST_UNIT=0`) at reviewed commit `294c27a693`: reviewer ran it twice independently (8243 tests, 6 shards, 0 FAIL, 0 SKIP) and this gate ran it a third time in an isolated worktree, also 0 FAIL.
- Two earlier deploy-gate attempts on this same bead, at this identical unchanged SHA, both hit exactly 3 failures (`TestBuildDesiredState_MinZeroDefaultScaleCheckRoutedWorkCreatesPoolSession`, `TestEvaluatePoolDefaultScaleCheckCountsRoutedReadyWork`, `TestEvaluatePoolDefaultScaleCheckIgnoresRoutedActiveUnassignedWork`) — the known ambient shared-Dolt-server signature root-caused and closed at `ga-zxpfic`. Three clean runs and two failing runs at the *same* commit is direct evidence this is ambient nondeterminism, not a regression introduced by this change. Environment fix tracked separately at `ga-us7c35` (open, P1).
[...]

PR #5116, opened 2026-08-08
fix(git): only count remotes that can publish off-host as pushed (gas-9sg)

AJBcoding:
## Problem

A remote whose URL points back at its own repository publishes nothing off the host — but pushing or fetching there still creates `refs/remotes/<name>/*`. Every probe that consults *all* remote-tracking refs then reports the work published:

```
git branch -r --contains <sha>   =>  remotes/herdr-src/fix/...   (looks published)
git ls-remote ajb <ref>          =>  empty                       (actually not published)
```

`HasUnpushedCommitsResult` asked exactly that question via `git log HEAD --not --remotes`. It gates **worktree removal** (`cmd/gc/session_worktree_prune.go`) and the doctor semantic check, so a self-referential remote turned a data-loss safety check into a green light over work that existed in exactly one place on local disk.

This is not hypothetical (gas-9sg): two keystone fixes were found single-copy on local disk, their beads already closed as "published", and an audit found 15 branches on no real remote.

## Fix

The probe now consults only **publishable** remotes — those whose URL does not resolve to this same repository.

- Self-reference is the *only* disqualification. A remote at another path is a distinct repo and a legitimate target; a scheme-bearing or scp-style URL names another machine outright. (An over-broad "ignore local paths" rule would have broken existing behavior — the pre-existing tests using a local bare repo as `origin` still pass unchanged.)
- With no publishable remote at all, nothing is published and every commit reads as unpushed — fails safe.
- Identity is compared with `os.SameFile` against the git **common dir** and the **main worktree root**, not the working directory: in production the self-referential remote names the main repo while the agent works in a linked worktree, so a working-directory comparison would miss the exact shape that caused the loss. A test pins that case.

`HasUnreachableCommitsResult` deliberately keeps plain `--remotes`. It asks whether removing a worktree would *orphan* a commit, and a remote-tracking ref — self-referential or not — is a real local ref that keeps the object reachable. Publication and reachability are genuinely different questions.

## Verification

- TDD, red first: the three probe tests failed on the pre-fix code exactly as predicted (plain repo, **linked-worktree production shape**, and self-remote masking a commit the real remote lacks), then passed.
- 7 new tests; full `internal/git` package green (6.9s), `internal/doctor` green (24s), `cmd/gc` prune consumers green.
- `go vet` clean; `gofmt` clean; `make lint-changed` 0 issues.
- Local pre-push gate bypassed under the standing gas-90h authorization (gas-bsj / #5111 still pending, macOS unit suite self-deadlocks); Linux CI is the oracle.


sjarmak (changes requested):
Thanks for this. The self-referential-remote case is a real hole, and the `os.SameFile` identity approach, including comparing the worktree/`.git` pairing both ways, is the right way to test it.

One blocker, and one thing worth handling in the same pass.

**Blocker: `remoteURLIsSelfReferential` has no repo context, so a relative remote URL resolves against the wrong directory.**

The function calls `os.Stat(path)` on whatever `localRemotePath` returns. `git remote get-url` reports relative URLs verbatim, so `git remote add origin ../mirror` or `git remote add origin .` yields a relative path, and `os.Stat` resolves it against the calling process's working directory rather than the repository's. Nothing in the signature carries `workDir` or any other repo context, so it cannot resolve correctly.

The direction of the failure is what makes this a blocker. On `os.Stat` error the function returns `false`, meaning "not self-referential", which promotes the remote to publishable and makes `HasUnpushedCommitsResult` report the commits as pushed. Both real callers, `cmd/gc/session_worktree_prune.go:102,177` and `internal/doctor/checks_semantic.go:633`, check the error and refuse to act when the probe fails. Here there is no error to check. They get a confident wrong answer, which is the case this PR exists to prevent.

Threading the repository's own directory in and resolving relative URLs against it closes this. Treating an unresolvable relative URL as self-referential would also be safe, since that is the conservative direction for a predicate that gates destructive cleanup.

**Loopback URLs are always classified as off-host.**

`localRemotePath` returns early for anything containing `://`, and for scp-style `host:path`, without inspecting the host value. `ssh://localhost/...`, `ssh://127.0.0.1/...`, `ssh://[::1]/...` and `user@localhost:/...` all classify as off-host publication targets. The comment that a URL carrying a scheme names another machine outright does not hold for the loopback forms, which containerized and sandboxed setups do use.

Fixing it in this pass or tracking it as an explicit follow-up both work. Flagging it because this predicate gates worktree removal.

The rest checks out. The fail-safe direction is right everywhere else: no remotes resolves to "has unpushed", an unresolvable `get-url` is excluded from the publishable set, and `repoIdentityDirs()` errors propagate as real errors both callers honor. Reverting the fix fails three of the new tests, so they are load-bearing. Reading the fetch URL rather than the push URL is also correct here, since the fetch URL is what populates the remote-tracking refs the probe consults.

AJBcoding:
Both items fixed, plus one latent bug found in passing.

## 1. Blocker: relative remote URLs resolved against the process CWD

Confirmed exactly as reported, including the fail-**open** direction, which is
the part that made it a blocker. `git remote get-url` reports a URL verbatim, so
`git remote add origin ../mirror` (or `.`) yields a relative path, and
`remoteURLIsSelfReferential` had no repo context — `os.Stat` resolved it against
the calling process's working directory. On a stat error the function returned
"not self-referential", which promoted the remote to publishable and made
`HasUnpushedCommitsResult` report unpublished work as **pushed**. Both callers
(`cmd/gc/session_worktree_prune.go`, `internal/doctor/checks_semantic.go`) do
check for a probe error and refuse to act — but there was no error to check.
They got a confident wrong answer, which is the case this check exists to
prevent.

Relative URLs now resolve the way git resolves them. Measured rather than
assumed, since the fix has to match git's behavior:

- from a repo **subdirectory**, `../mirror` still resolves — so the base is not
  the process CWD;
- in a **linked worktree**, the base is that worktree's own root, not the main
  repo's — the two differ, and picking the main root classifies a different
  directory;
- a **bare** repo has no worktree, so the git dir is the base.

So the base is `rev-parse --show-toplevel`, falling back to the common git dir.
Note this is deliberately a *different* directory from the identity dirs, which
stay the common git dir plus the main worktree root: a self-referential remote
names the **main** repo while the agent is working in a linked worktree.
`repoIdentity` now carries both, and a test pins the linked-worktree case.

**The predicate is inverted so it cannot fail open again.**
`remoteURLIsSelfReferential` becomes `remoteIsPublishable`, so the unsafe answer
is no longer the zero value: anything not provably a distinct repository is not
counted as one. That is also right on the merits, not merely conservative — a
path that cannot be stat'ed cannot have received anything, so pushing to it
publishes nothing.

## 2. Loopback hosts

`localRemotePath` returned early for anything containing `://` and for
scp-style `host:path` without inspecting the host, so `ssh://localhost/…`,
`ssh://127.0.0.1/…`, `ssh://[::1]/…` and `user@localhost:/…` all read as
off-host publication targets. Containerized and sandboxed setups do use those
forms, and the host is this machine, so such a URL can name this very
repository. They now resolve to their path and take the same identity
comparison a plain local path takes.

`http`, `https` and `git` stay off-host even on loopback: their path is a
server route, not a directory that can be compared. A relative scp-style path
(`user@localhost:repo`) resolves against the remote account's home directory,
which this process cannot know, so it reads as unresolvable and fails safe
rather than guessing.

## 3. Latent bug found while in there

`file://localhost/path` was parsed with `strings.CutPrefix("file://")`, yielding
the path `localhost/path`. The authority is now parsed as a host via `net/url`.

## Verification

Ten new tests, and **all ten fail against the previous implementation**
(revert-checked in both directions, implementation restored afterward). Four of
them are deliberate over-reach guards — distinct relative URL, distinct loopback
URL, non-loopback ssh host, linked-worktree base — so they fail if the fix
classifies too aggressively or picks the wrong base. The end-to-end shape is
included: a repo whose only remote is itself, spelled `.`, must report its
commits unpushed.

`internal/git` and `internal/doctor` pass in full; `go vet` clean.

## Note on the base refresh

You'll see a merge commit that isn't a review response. The branch was based
before #5111, so its macOS unit gate failed on the `t.TempDir()` path-spelling
mismatch from #4929 — `/var/folders/…` vs `/private/var/folders/…` — in
`cmd/gc` tests untouched by this PR. I verified that failure was pre-existing by
stashing this change and re-running at the PR head: identical failure. Merging
current `main` (`b8e0edffb`) picks up #5111 and #5097 so the gate exercises this
diff instead of that one.

Merge rather than rebase, deliberately: it leaves `a0022dd06` — the commit you
reviewed — in place and keeps this a fast-forward rather than a force-push under
your review. Since `main` squash-merges, the merge commit disappears from the
final history either way.

AJBcoding:
@sjarmak — re-review ping.

Both review items are addressed at `7a48edd0d` (an ancestor of the current head `083fcc37e`):

- **Relative remote URLs** now resolve against the repo the way git resolves them, not against the process CWD — and the predicate is inverted (`remoteIsPublishable`) so the unsafe answer is no longer the zero value and it cannot fail open again.
- **Loopback hosts** (`ssh://localhost/…`, `127.0.0.1`, `[::1]`, `user@localhost:/…`) now resolve to their path and take the same identity comparison a plain local path takes.

The reasoning, the measured git-resolution cases, and the test inventory are in [the 2026-08-09 reply](https://github.com/gastownhall/gascity/pull/5116#issuecomment-5234097947) above rather than restated here. Nothing has changed on the branch since.

One process note, since it explains the delay rather than anything about the code: the fix and that reply went up on 2026-08-09, but a PR does not return to your queue after a CHANGES_REQUESTED review unless someone re-requests it, and this account gets a 404 from the review-request API on this repo — so I cannot. An @-mention is the only channel available, which makes this comment the notification. Apologies for the extra ping.

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277173124

PR #5117, opened 2026-08-08
fix(runtime/herdr): confirm startup first-turn submission, recover swallowed submit CR (gas-90h)

AJBcoding:
## Relay, not new work

This proposes an existing commit (`c79416739`) that was completed on bead **gas-90h** by polecat session `gastown__polecat-az-wisp-396im`, along with its tests and design notes. **Credit for the fix and its verification belongs to that session, not to me** — I am relaying it upstream.

Why it needs relaying: the bead was closed as *"Fixed and PUBLISHED: ajb branch fix/gas-90h-herdr-startup-delivery-confirm"*. "Published" there meant pushed to the fork; it was never proposed upstream, so the work has sat on one fork branch since Aug 7 while a follow-up bead (gas-mvl) blocks on it.

That is a live instance of the class gas-9sg names — quoting the framing from that investigation:

> work that looked published, wasn't, and is now blocking downstream work

The mechanism is in #5116: a self-referential git remote creates `refs/remotes/<name>/*` entries, so publication probes report success while nothing has left the host.

## What the commit does

Startup first turns are delivered via `agent prompt --wait --until working/done/blocked --timeout 15000` (`deliverStartupTurn`). Before it, `gc` called `herdr agent prompt` without `--wait`, so a swallowed submit CR stranded a brand-new agent with its prompt typed but never submitted, and no backstop covered named sessions. Now a swallowed CR surfaces as `agent_prompt_stalled` and is recovered in-band with one explicit Enter plus a confirming wait; `WaitForIdle` verdicts are made legible (`waitForIdleOutcome`); and a delivery that even recovery cannot confirm is recorded durably via the sidecar marker `GC_HERDR_STARTUP_DELIVERY_UNCONFIRMED` (reset each life) plus stderr. Mid-session nudges are unchanged — still fast and un-waited.

6 files, one commit off `main`: provider/client/capabilities plus a new `startup_delivery_confirm_test.go` and the provider design note.

## Verification (as recorded by the original author on gas-90h)

6 new unit tests pinning the `--wait` flags, Enter recovery (no re-prompt), marker semantics, raw-pane fallback, and outcome mapping; the herdr package green including a live herdr+claude kind-path smoke (28.6s); repo-wide `go vet` clean; `gofmt` clean; and a real cold-boot demo confirming a prompt in 217ms against a baseline of 65 unverified oks / 0 confirmed.

I have not re-run those locally — CI here is the independent check.

## Follow-up this unblocks

gas-mvl widens the startup-delivery backstop to named sessions and keys on the `GC_HERDR_STARTUP_DELIVERY_UNCONFIRMED` marker this commit introduces. It cannot be built on `main` until this lands, because nothing there would ever set the marker.

@quad341 — review/merge when you have a moment. This joins #5103, #5104, #5111, #5114, #5115, and #5116 as the batch of fork PRs awaiting maintainer action; #5111 is the one with fleet-wide leverage, since until it lands the macOS pre-push gate is broken for every agent.

### DIFF
--- a/internal/runtime/herdr-provider-design.md
+++ b/internal/runtime/herdr-provider-design.md
@@ -71,6 +71,15 @@ exhaustion; bead az-405 has the full evidence trail). The adapter was rewritten 
 - **Delivery**: `deliverNudge` targets the pane id via native `agent prompt`
   (registered agents), falling back to paste+Enter for unregistered panes.
   `WaitForIdle` uses `agent wait --until idle`. `Peek` reads via `pane read`.
+- **Startup delivery** (`deliverStartupTurn`, gas-90h): the FIRST turn opts into
+  submission confirmation — `agent prompt --wait --until working/done/blocked`
+  — because a freshly-booted TUI can swallow the submit CR while the un-waited
+  prompt reports ok (text typed-but-unsubmitted, agent stranded idle). On
+  `agent_prompt_stalled`/`timeout` the recovery is one explicit `send-keys
+  Enter` (the text is already in the box; never re-prompt) plus a confirming
+  wait; a still-unconfirmed delivery is recorded on the sidecar
+  (`GC_HERDR_STARTUP_DELIVERY_UNCONFIRMED`) and stderr. Mid-session nudges keep
+  the fast un-waited form.
 
 ### Operational gotchas (learned in production, 2026-07-26)
 
--- a/internal/runtime/herdr/capabilities.go
+++ b/internal/runtime/herdr/capabilities.go
@@ -22,18 +22,49 @@ var (
 	_ runtime.LivenessObserver = (*Provider)(nil)
 )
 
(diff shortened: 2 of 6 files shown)


AJBcoding:
Superseded by #5119 — same commit, cherry-picked onto current `main` (authorship preserved).

Closing because this branch sat 12 commits behind `main`, and the resulting CI red was entirely stale-base: `TestDisableAndPurgeRejectsUnprovenPeerSuccessor` (internal/productmetrics) and `TestGCEnvReadBaseline` (internal/testenv), neither of which this change touches — it is confined to `internal/runtime/herdr`. The identical integration shard passes on #5116, which is based on current `main`.

The original fork branch `fix/gas-90h-herdr-startup-delivery-confirm` (`c79416739`) is left untouched rather than force-pushed, so the SHA cited by the gas-90h bead stays valid.

Please review #5119 instead.

PR #5118, opened 2026-08-08
test(cmd/gc): widen the dolt leak guard to the package directory (gas-dfn)

AJBcoding:
## Problem

The `cmd/gc` dolt leak guard scopes its before/after snapshots to the private temp root:

```go
snapshotDoltProcessesForConfigRoot(enumerate, g.tempRoot)
```

A test whose city path resolves to `go test`'s cwd instead of `t.TempDir()` starts a managed dolt at `$REPO/cmd/gc/.beads/dolt` — **outside both snapshots**. Such a server leaks silently, the package still reports `ok`, and its lock on that data dir can block the next run in the same checkout (cf. `TestStartManagedDoltFailsClosedWhenDataDirLockHeld`).

Not hypothetical: a `dolt sql-server` was found alive for a day rooted at `gascity-v140/cmd/gc/.beads/dolt`, and that residue is still on disk in the v140 checkout.

## What this changes

The snapshot takes multiple roots and watches the package directory alongside the temp root, so the class is **reported instead of invisible**.

- The single-root entry point delegates to the multi-root form, so existing callers keep identical semantics (pinned by a test).
- Empty roots are ignored rather than treated as match-everything — a failed root resolution must never turn the guard into a box-wide reaper of other checkouts' dolt servers (also pinned by a test).
- The leak message now names both roots.

## Scope — deliberately not "fixing" the test

**The offending test is not fixed here, because it no longer reproduces.** A full `cmd/gc` run on current code leaves no `cmd/gc/.beads` or `cmd/gc/.gc` behind in this worktree, and only the v140 checkout carries residue — consistent with an older run, exactly as the bead suspected when it said to confirm against current code before assuming. Inventing a fix for a test that no longer misbehaves would be guesswork; widening the guard is what makes the class loud if it returns.

The signal-path reaper still scopes to the temp root alone. It kills processes, and widening it would reap a repo-rooted server that *predates* the run rather than one this run leaked; the diff-based path already reaps genuinely new ones.

Follow-ups left to the bead, not done here: gitignoring `cmd/gc/.beads` + `cmd/gc/.gc` as defense in depth (the working tree has unrelated uncommitted `.gitignore` edits, so I kept this commit clean), and removing the stale v140 residue (a checkout I do not own; its leaked PID 74766 is already gone).

## Verification

- TDD, red first: `snapshotDoltProcessesForConfigRoots` undefined → 3 new tests green.
- Leak-guard/snapshot/diff tests green; `gofmt` clean; `make lint-changed` 0 issues.
- Full `cmd/gc` run: no residue produced. Its failures (`TestReapSkipTracker_*`, a 10m package timeout) are pre-existing and reproduce identically on clean `origin/main` — I confirmed that on a control worktree earlier today; the one "dolt leak guard" line in the log is the guard's own unit-test fixture, not a real leak.
- Local pre-push gate bypassed under the standing gas-90h authorization (gas-bsj / #5111 pending); Linux CI is the oracle.

### DIFF
--- a/cmd/gc/dolt_leak_repo_root_test.go
+++ b/cmd/gc/dolt_leak_repo_root_test.go
@@ -1,0 +1,98 @@
+package main
+
+import (
+	"path/filepath"
+	"testing"
+)
+
+// doltProcWithConfigUnder builds a dolt sql-server process record whose
+// --config path sits under dir, the shape discoverDoltProcesses reports.
+func doltProcWithConfigUnder(pid int, dir string) DoltProcInfo {
+	return DoltProcInfo{
+		PID: pid,
+		Argv: []string{
+			"dolt",
+			"sql-server",
+			"--config",
+			filepath.Join(dir, ".gc", "runtime", "packs", "dolt", "dolt-config.yaml"),
+		},
+	}
+}
+
+// TestSnapshotDoltProcessesForConfigRootsSeesRepoRootedServer is the gas-dfn
+// regression. The leak guard scopes its before/after snapshots to the private
+// temp root, so a test that starts a managed dolt whose data-dir resolves to
+// the PACKAGE DIRECTORY instead of t.TempDir() is outside both snapshots: it
+// leaks silently, the package still reports ok, and the live server holds the
+// lock on $REPO/cmd/gc/.beads/dolt so the next run in that checkout can block.
+// Scoping to the temp root AND the package dir makes that class loud.
+func TestSnapshotDoltProcessesForConfigRootsSeesRepoRootedServer(t *testing.T) {
+	tempRoot := filepath.Join("/tmp", "gc-cmd-test-root")
+	repoRoot := filepath.Join("/Users/someone/code/gascity", "cmd", "gc")
+	owned := doltProcWithConfigUnder(1001, filepath.Join(tempRoot, "TestOwned", "001"))
+	repoRooted := doltProcWithConfigUnder(1002, repoRoot)
+	unrelated := doltProcWithConfigUnder(1003, filepath.Join("/tmp", "TestOther", "001"))
+
+	got, err := snapshotDoltProcessesForConfigRoots(func() ([]DoltProcInfo, error) {
+		return []DoltProcInfo{owned, repoRooted, unrelated}, nil
+	}, tempRoot, repoRoot)
+	if err != nil {
+		t.Fatalf("snapshotDoltProcessesForConfigRoots: %v", err)
+	}
+	if _, ok := got[1002]; !ok {
+		t.Errorf("snapshot missed the repo-rooted server (PID 1002): %#v — this is the leak that reports ok today", got)
+	}
+	if _, ok := got[1001]; !ok {
+		t.Errorf("snapshot dropped the temp-root server (PID 1001): %#v", got)
+	}
+	if _, ok := got[1003]; ok {
+		t.Errorf("snapshot included an unrelated server (PID 1003): %#v — the guard must not claim processes it does not own", got)
+	}
+}
+
+// TestSnapshotDoltProcessesForConfigRootsIgnoresEmptyRoots proves a blank root
+// never widens the scope to everything. An empty tempRoot (createActiveTestTempRoot
+// failed) must not turn the guard into a box-wide reaper of other checkouts'
+// dolt servers.
+func TestSnapshotDoltProcessesForConfigRootsIgnoresEmptyRoots(t *testing.T) {
+	proc := doltProcWithConfigUnder(1001, filepath.Join("/tmp", "elsewhere"))
+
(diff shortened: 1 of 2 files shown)


AJBcoding:
The two red checks here are a **cancellation, not a test failure**, and I verified that rather than assuming it — this PR touches `cmd/gc` test plumbing, so "probably unrelated" wasn't good enough.

Evidence:

- The `cmd/gc product metrics testhook` job logged `##[error]The operation was canceled` at 4m54s with no failing test in its output. `CI / required` then failed in 2s, which is the aggregate reacting to it.
- The same job passes in 4m18s on #5116, so ~4m54s is right at its normal completion — it was killed near the finish line, most likely by a concurrency cancellation while several of my branches were pushed in quick succession.
- Decisive check: I ran the identical target locally on this branch — `make test-productmetrics-testhook` — and it passes (`PASS`, exit 0).

I can't `gh run rerun` (this account lacks admin rights on the repo), so I'm closing and reopening this PR to trigger a fresh run. No code changed.

AJBcoding:
CI is green on the fresh run: 60 pass / 0 fail. Confirms the earlier red was the cancellation described above, not a test failure. Ready for review.

sjarmak:
Thanks for this. The change itself is sound. Cherry-picked onto current main, `GC_FAST_UNIT=1 go test ./cmd/gc/` stays green, `pathutil.PathWithin` uses `filepath.Rel` rather than a string prefix so it will not match `cmd/gc2` against `cmd/gc`, and reverting the multi-root logic fails `TestSnapshotDoltProcessesForConfigRootsSeesRepoRootedServer`. No correctness objection.

The issue is overlap. #5057 fixes the same bug, was opened two days earlier, and touches the same file. The two are incompatible as written: this PR lands `snapshotDoltProcessesForConfigRoots(enumerate, roots ...string)`, while #5057 lands `...(enumerate, roots []string)` and keeps a single-root wrapper. Whichever merges second hits conflicts in `path_helpers_test.go` across the struct field, the constructor, both `runWith` call sites, and the shared function definition.

#5057 also reaches a path this one does not. It widens `reapDoltProcessesUnderRoot`, which runs from the SIGINT/SIGTERM handler, so a run interrupted with Ctrl-C still detects a package-dir-rooted leak. This PR leaves that call on the old single-root form.

Suggested route: land #5057 first, then rebase this down to whatever coverage is genuinely additive. On a spot check `dolt_leak_repo_root_test.go` looks covered by #5057's tests, so it may end up empty, but that is worth your own read rather than my assertion.

Separately, and independent of which of these lands: neither PR touches `sweepStaleCmdGCTestDoltProcesses`, the startup sweep for servers that outlive a test binary across invocations. That is the shape of the long-lived leak the issue describes, so it is worth a follow-up either way.

sjarmak:
Correcting my earlier comment, since it pointed you at an open-ended wait.

I said land #5057 first. What I did not check before saying it: #5057 has zero reviews and its CI has never actually run. Its `Runner policy` gate job failed on 2026-08-06 and every downstream test job shows `skipping`, several still carrying unexpanded `${{ matrix.* }}` names, meaning they never instantiated. Those logs have since expired, so the original cause is no longer readable. Its checks are neither green nor red; nothing ran.

So "wait for #5057" was not a bounded wait, and that is on me rather than on you. Your PR is the one with real evidence behind it: I cherry-picked it onto current main and `GC_FAST_UNIT=1 go test ./cmd/gc/` stayed green.

I have put #5057 into review and re-triggered its CI. The completeness point I raised still stands and is being verified rather than assumed: #5057 appears to also widen `reapDoltProcessesUnderRoot`, the SIGINT/SIGTERM reap path, which this PR leaves on the old single-root call. If that holds and #5057 comes together, it is the better single landing. If it does not, this PR is what lands, and the conflict question disappears.

Either way, nothing here needs action from you right now. I will post the outcome on both PRs.

sjarmak:
Closing this as superseded by #5057, which just merged.

To be clear about why, since your change was correct: the two are mutually exclusive edits to `cmd/gc/path_helpers_test.go` and only one could land. #5057 was opened two days earlier and also widens `reapDoltProcessesUnderRoot`, the SIGINT/SIGTERM reap path, which this PR leaves on the single-root call. I verified that rather than taking it from the diff summary. That extra coverage is the whole of the difference; on the shared part your implementation was sound and I confirmed it green on current main.

Apologies for the detour earlier in the thread, where I pointed you at #5057 without first checking that it was reviewable. It was not, and sorting that out is what unblocked this.

The follow-up I mentioned is still open and unclaimed if you want it: neither PR touches `sweepStaleCmdGCTestDoltProcesses`, the startup sweep for dolt servers that outlive a test binary across separate invocations. That is the shape of the long-lived leak the original issue describes, so the in-process guard being fixed does not close it.

PR #5119, opened 2026-08-08
fix(runtime/herdr): confirm startup first-turn submission, recover swallowed submit CR (gas-90h)

AJBcoding:
Supersedes #5117 — same change, rebased onto current `main` so CI is meaningful. (#5117's branch sat 12 commits behind, and its integration shard failed on stale-base tests unrelated to this change: `TestDisableAndPurgeRejectsUnprovenPeerSuccessor` in `internal/productmetrics` and `TestGCEnvReadBaseline` in `internal/testenv`. The identical shard passes on #5116, which is based on current `main`.)

## Relay, not new work

This proposes an existing commit completed on bead **gas-90h** by polecat session `gastown__polecat-az-wisp-396im`. **Credit for the fix and its verification belongs to that session, not to me** — I am relaying it upstream, and git authorship is preserved through the cherry-pick.

Provenance: the original commit is `c79416739` on the fork branch `fix/gas-90h-herdr-startup-delivery-confirm`, which I left **untouched** rather than force-pushing a rebase over it — this fork has lost work to branch churn before, and a closed bead cites that SHA. `009373703` here is that commit cherry-picked onto current `main`.

Why it needed relaying at all: gas-90h was closed as *"Fixed and PUBLISHED"*, where "published" meant pushed to the fork. It was never proposed upstream, so the work sat on one fork branch since Aug 7 while a follow-up bead (gas-mvl) blocked on it — a live instance of the class gas-9sg names:

> work that looked published, wasn't, and is now blocking downstream work

The mechanism is fixed in #5116: a self-referential git remote creates `refs/remotes/<name>/*` entries, so publication probes report success while nothing has left the host.

## What the commit does

Startup first turns are delivered via `agent prompt --wait --until working/done/blocked --timeout 15000` (`deliverStartupTurn`). Before it, `gc` called `herdr agent prompt` without `--wait`, so a swallowed submit CR stranded a brand-new agent with its prompt typed but never submitted, and no backstop covered named sessions. Now a swallowed CR surfaces as `agent_prompt_stalled` and is recovered in-band with one explicit Enter plus a confirming wait; `WaitForIdle` verdicts are made legible (`waitForIdleOutcome`); and a delivery that even recovery cannot confirm is recorded durably via the sidecar marker `GC_HERDR_STARTUP_DELIVERY_UNCONFIRMED` (reset each life) plus stderr. Mid-session nudges are unchanged — still fast and un-waited.

## Verification

Original author's, as recorded on gas-90h: 6 new unit tests pinning the `--wait` flags, Enter recovery (no re-prompt), marker semantics, raw-pane fallback, and outcome mapping; herdr package green including a live smoke; repo-wide `go vet` clean; a real cold-boot demo confirming a prompt in 217ms against a baseline of 65 unverified oks / 0 confirmed.

Mine on the rebase: cherry-pick applied cleanly (6 files, no conflicts); `go vet ./internal/runtime/herdr` clean. The package's `TestProviderLiveClaudeKindPath` is a live smoke test that is flaky in this environment — it **fails on clean `origin/main` too**, and passes in isolation on this branch — so it is pre-existing, not from this change. CI is the real check.

## Follow-up this unblocks

gas-mvl widens the startup-delivery backstop to named sessions and keys on the `GC_HERDR_STARTUP_DELIVERY_UNCONFIRMED` marker this commit introduces. It cannot be built on `main` until this lands, because nothing there would ever set the marker.

@quad341 — review/merge when you have a moment. This joins #5103, #5104, #5111, #5114, #5115, #5116, and #5118 as the batch of fork PRs awaiting maintainer action; #5111 is the one with fleet-wide leverage, since until it lands the macOS pre-push gate is broken for every agent.


AJBcoding:
Pushed `585e9480f` to fix a **genuine, change-caused** CI failure — and to correct something I got wrong when I closed #5117.

`TestGCEnvReadBaseline` (internal/testenv) walks the repo collecting `GC_*` env reads and diffs them against a golden vocabulary. The relayed commit introduces `GC_HERDR_STARTUP_DELIVERY_UNCONFIRMED`, so the golden needed a deliberate update. It reproduces locally on this branch, and the test names the variable outright:

```
GC_* env-read vocabulary changed vs testdata/gc_env_read_baseline.golden.
ADDED (new env reads — add to internal/rollout/config instead of an ad-hoc GC_ var, or update the golden deliberately):
  GC_HERDR_STARTUP_DELIVERY_UNCONFIRMED
```

I updated the golden rather than routing the marker through `internal/rollout/config` (the test's other suggestion), because it is a per-life delivery marker on the herdr provider, not a rollout flag. It is a separate commit so the relayed commit stays byte-identical to what its author published.

**Correction.** When I closed #5117 I said its CI red was *entirely* stale-base. That was right for `TestDisableAndPurgeRejectsUnprovenPeerSuccessor` (internal/productmetrics), which passes here on a current-`main` base — but wrong for `TestGCEnvReadBaseline`, which is this change's own missing baseline entry and would have failed on any base. I should have reproduced it locally before attributing it. The rebase was still needed; that specific claim was overconfident.

The remaining known flake in this package is `TestProviderLiveClaudeKindPath`, a live smoke test that fails on clean `origin/main` too.

AJBcoding:
CI is green: 63 pass / 0 fail, with the baseline fix in 585e9480f. Ready for review and merge.

sjarmak (dismissed):
Thanks for this one, and for the batch generally. The startup-confirm gap is real and the herdr-side `--wait --until` approach is the right shape for it. Three things before it lands.

**1. The recovery Enter fires on a case the tests never exercise.**

`deliverStartupTurn` treats `agent_prompt_stalled` and `timeout` identically (`client.go:304-334`), and both derive from herdr's fixed 5000ms observed-state-change window (`client.go:272-275`). That window cannot separate "the CR never reached the TUI" from "the CR landed and the agent is slow to visibly transition." In the second case the input box is empty and the session may already be mid-turn.

The fake herdr in `panebinding_provider_test.go:206-226` can return success or `agent_prompt_stalled` from `agent prompt`, but has no path to return `timeout` from that initial call. So the swallowed-CR half of the branch is covered and the submitted-but-slow half is not.

This matters because the code's own state vocabulary includes `blocked` for "hit a dialog" (`client.go:280-285`). A blind Enter into a confirmation dialog is a different outcome than a no-op on an empty box. The docstring at `client.go:294-297` asserts the Enter is safe in every plausible state; nothing tests it against the one state that vocabulary calls out.

**2. The confirm bound is tighter than this file's own sizing for the same scenario.**

`startupPromptConfirmTimeoutMS` is 15000, gating on herdr's fixed 5000ms window (`client.go:276-277`). Elsewhere in this provider, anything touching the agent's boot path is sized at 60s with the explicit rationale "cold, concurrent claude boots during a town-wide restart" (`agentStartTimeoutMS`, `client.go:111`; `startupNudgeIdleTimeout` in `provider.go`). The 15s bounds are for shell-prompt detection. A concurrent cold-boot wave is exactly when recovery keystrokes would misfire on healthy sessions, and it is the case a warm single-boot measurement does not reach.

**3. `GC_HERDR_STARTUP_DELIVERY_UNCONFIRMED` has no production reader.**

Production writes it (`SetMeta` in `recordStartupDeliveryUnconfirmed`, `RemoveMeta` at `provider.go:229`); every `GetMeta` is in `startup_delivery_confirm_test.go`. The key does not appear anywhere on main. The reader being a follow-up is reasonable sequencing, but today the marker records a failure nothing acts on.

It can also orphan. That `RemoveMeta` sits inside `if !adopted && startupText != "" && info.PaneID != ""`. If a life sets the marker and then crashes without a clean Stop, a later Start that adopts a live holder, or that has no startup text, never reaches the clear. `TestStartupDeliveryConfirmedClearsStaleMarker` covers the normal path only. Better handled before the reader lands, since that is the point where a stale marker becomes a false positive.

Smaller note: `TestStartupDeliveryUnregisteredPaneFallsBackToPaste` is the first test to reach `pasteAndSubmit`, which carries the 1s `submitSettleDelay`, so it adds about a second of real wall-clock. The resource census only counts `time.Sleep` written directly in a `_test.go`, so this does not show up in the ledger.

Happy to take another look once these are addressed.

AJBcoding:
Thanks — all three are addressed in `ce03c7899`, pushed as a fast-forward on top of the reviewed head (`585e9480f` is an ancestor of the new tip, so nothing under review was rewritten; no rebase, no force-push). `9c506b472` merges current `main` in.

The load-bearing fact for all three turned out to be herdr 0.7.5's own `agent prompt --help`: `--wait` first requires an observed state change within 5000ms, otherwise it returns `agent_prompt_stalled`; a shorter `--timeout` returns `timeout` instead.

**1. The recovery Enter is now qualified by state — never blind.**

Split apart, the two verdicts turn out to mean nearly opposite things:

- `timeout` is only reachable *after* that 5000ms state-change gate has passed, since our bound exceeds it by construction. So it proves the submit landed, and only the settle into a confirming state ran long. The box is not idle — the agent may be mid-turn, or `blocked` on a dialog. It now sends no keys at all and reports the turn unconfirmed (`client.go:358-360`).
- `agent_prompt_stalled` means nothing changed, which is the swallowed CR *only if the agent was idle throughout*. An agent already parked on a confirmation dialog is equally unchanging, and the readiness wait cannot rule that out — over 20h the live city measured 0 idle / 8 timeout / 2 error. That branch now reads the state directly (`agent get`, which herdr resolves for a pane id — verified against the installed 0.7.5) and withholds the Enter unless herdr says `idle` (`client.go:361-371`).

That closes the state your review called out. `blocked` is now either matched by the confirming wait — it is in `startupConfirmStates`, so it is a confirmation rather than a recovery case — or it withholds the keystroke.

The fake herdr gained the missing verdict paths: `timeout` and an `--until`-aware `blocked` from `agent prompt`, plus a `blocked` `agent get`. Both halves of the branch are exercised now, not just the swallowed-CR one.

**2. The bounds are the boot-path budget now.**

You were right that these were sized against the wrong scenario. `startupPromptConfirmTimeoutMS` (15s) and `startupStallRecoveryTimeoutMS` (10s) are both now `startupBootBudgetMS` — 60s, the same named constant `agentStartTimeoutMS` uses, with `startupNudgeIdleTimeout` as its `Duration` twin. A cold, concurrent boot wave is exactly when an early-expiring bound manufactures a strand out of a healthy slow session. Worst case is three of these end to end, comfortably inside the reconciler's 10m `pendingCreateNeverStartedTimeout`.

**3. The marker can no longer orphan.**

Agreed it belongs before the reader, not after. The clear moved out of the delivery guard (`!adopted && startupText != "" && info.PaneID != ""`) and is now unconditional and idempotent, immediately after the pane binding is persisted (`provider.go:143`). That Start is past the `ErrSessionExists` check, so a new life owns the sidecar from there, and only this life's delivery can re-set the marker. `RemoveMeta` already tolerates a missing key.

**Smaller note.** Good catch on the invisible wall-clock. The 1s `submitSettleDelay` guards a real paste racing a real submit and is meaningless against a shell script, so it is now a client field the fake-herdr provider shortens. `TestStartupDeliveryUnregisteredPaneFallsBackToPaste` drops from ~1.15s to ~0.15s.

**Tests.** Five added:

- `TestStartupDeliveryStallOnBlockedAgentWithholdsEnter`
- `TestStartupDeliveryTimeoutNeverBlindEnters`
- `TestStartupDeliveryBlockedConfirmsWithoutRecovery`
- `TestStartupDeliveryClearsStaleMarkerOnAdopt`
- `TestStartupDeliveryClearsStaleMarkerWithNoStartupText`

plus updates to existing tests for the new bounds, and one to pin that the state read precedes the recovery Enter.

**What I ran, and what I did not.** `internal/runtime/herdr` is green locally — 77 pass, `go vet` clean. I did not run the full local sweep. On this host it cannot pass while the city is running, which is the same `TestProviderLiveClaudeKindPath` pane-availability flake you flagged as failing on clean `origin/main` (`agent_pane_busy: agent target pane w1:p1 is not an available shell` — it failed on one local run here and passed on the next). Three `cmd/gc` dolt-process/restart timing tests also failed in a loaded parallel run on this box; all three pass in isolation (0.76s / 0.15s / 0.61s), and they are in a package this change does not touch. Leaving the real verdict to CI.

AJBcoding:
CI is green on `9c506b472`: 63 success / 0 failures (21 skipped). That is the verdict I deferred to in the previous comment — including the `TestProviderLiveClaudeKindPath` and `cmd/gc` dolt-timing cases, which were local-host artifacts and did not reproduce in CI.

All three items are addressed and the push was a fast-forward, so the review history is intact. Ready for another look whenever you have time.

AJBcoding:
@sjarmak — re-review request, since this PR never made it back into your queue.

All three items (plus the wall-clock note) were answered in `ce03c7899`, pushed as a fast-forward on top of the reviewed head — `585e9480f` is still an ancestor, so nothing you reviewed was rewritten. Details: https://github.com/gastownhall/gascity/pull/5119#issuecomment-5234917930. CI is green on the head `9c506b472`: 63 success / 0 failures / 21 skipped.

Flagging the mechanics because they made the answer invisible: this account is permission-denied on `RequestReviewsByLogin`, so I can't re-request you as a reviewer, and a PR you've left `CHANGES_REQUESTED` on drops out of your review queue until someone does. The `CHANGES_REQUESTED` decision you still see is sticky by design — it persists until a new review is submitted — so it reflects your 08-09 review, not anything outstanding on our side.

No action needed beyond a look when you have one.

PR #5120, opened 2026-08-08
fix(gc): gate bug beads and probe reachability from a stable repo root (gas-myy)

AJBcoding:
Two ADR-0009 close-gate follow-ups from #5037, deliberately kept out of the gas-3xn Defect C diff to keep that review small.

## Finding 2 — bug beads were silently exempt

`isWorkRecordGatedBead` returned false for any type other than `"task"`. But a bug bead is exactly as worker-claimable as a task: claimed off the hook, worked, closed with a work outcome. Every bug close therefore bypassed the entire work-record contract **without a word** — and both silent rows in #5037's six-close table were type `bug`.

The gated set is now a named map covering `task`, `bug`, and the unset type, so widening it again is one obvious edit rather than a condition to re-reason about. Control/structural beads stay exempt through the unchanged `gc.kind` check, including bug-typed ones.

Note this arms the contract for a new bead class citywide — which is why #5037 split it out for its own review. The gate still ships warn-only unless `GC_WORK_RECORD_ENFORCE` is set, so the blast radius on merge is log lines, not blocked closes.

## Finding 3 — reachability ran in a directory that disappears

`repoDir` preferred the bead's `gc.work_dir`, which is a polecat worktree deleted when the worktree is reaped. Post-cleanup the probe ran in a missing directory, git errored, and the fail-closed reading reported the commit unreachable — **blocking a close whose work had in fact been delivered**. The gas-dr5 workflow review independently confirmed this with a concrete deputy-close scenario.

The stable scope root (the rig repo) now wins, with `gc.work_dir` kept as the fallback for callers that have no scope root. Worktrees share one object store, so the rig root answers the same question from a directory that still exists. The selection moved into `workRecordRepoDir` so it is testable on its own rather than inline in the gate loop.

## Verification

- TDD, red first (`workRecordRepoDir` undefined → tests green).
- Tests cover both findings: bug beads gated **end to end through the gate**, not just in the predicate; control-kind bug beads still exempt; the repo-dir preference and its fallback and its empty case; and a regression that builds a real git repo, points `gc.work_dir` at a pruned worktree, and asserts the delivered close is allowed.
- Pre-existing `work_record_gate` tests unchanged and green, so the widening did not disturb the established table.
- `go vet` clean; `gofmt` clean; `make lint-changed` 0 issues.
- Local pre-push gate bypassed under the standing gas-90h authorization (gas-bsj / #5111 pending); Linux CI is the oracle.

@quad341 — this joins #5103, #5104, #5111, #5114, #5115, #5116, #5118, and #5119 as the batch of fork PRs awaiting maintainer action; #5111 is the one with fleet-wide leverage, since until it lands the macOS pre-push gate is broken for every agent.


AJBcoding:
Pushed `8aefa6d86` for a **genuine, change-caused** CI failure.

`TestRepositoryLedgerMatchesCensusAndDocumentation` failed because my regression test used its own `exec.Command` helper, growing the repository resource census by exactly the amount it added:

```
Small resource census grew:  scope=untagged resource=subprocess calls=405 (baseline 403), files=113 (baseline 112)
source resource census grew: scope=all      resource=subprocess calls=608 (baseline 606), files=173 (baseline 172)
```

Rather than raise the baselines, I removed the need: the fixture now uses the package's existing `mustGit` helper for setup — which also scrubs the git env and disables `core.hooksPath`, which my ad-hoc helper did not — and reads the commit SHA from the loose ref instead of shelling out for `rev-parse`. The census is a budget on test resource usage; a test that does not need the budget should not spend it.

Verified locally: `TestRepositoryLedgerMatchesCensusAndDocumentation` passes, and the work-record gate tests still pass.

AJBcoding:
CI is green: 59 pass / 0 fail, with the census fix in 8aefa6d86. Ready for review and merge.

sjarmak (changes requested):
Thanks for this. The `bug`-type gap is real and worth closing, and the tests here are load-bearing in the way that matters: restoring the old behavior while keeping both new functions present fails four of the six new tests, so they detect a regression rather than just referencing new symbols.

Two things before it lands.

**1. The repo-root change relocates the false block rather than closing it.**

`workRecordRepoDir` returns `scopeRoot` whenever it is non-empty, and in the production path `scopeRoot` is never empty: `resolveStoreScopeRoot` (`cmd/gc/main.go:1530`) falls back to the city path when the store path is blank. So the `gc.work_dir` branch is unreachable in normal operation, which makes this an unconditional priority inversion rather than an added fallback.

Neither side is checked for existence. There is no `os.Stat` in the diff, and `resolveBdScopeTarget` treats a rig's registered `Path` as usable whenever the string is non-empty (`cmd/gc/cmd_bd.go:622-624` tests for empty, not for presence on disk). A rig whose directory was moved or removed still yields a non-empty root, the probe then runs against a repo that does not contain the commit, and the gate reports it unreachable.

That is the same failure this PR closes, moved from "worktree reaped" to "rig registration stale." The doc comment's rationale, that the rig root is a directory that still exists, is the assumption carrying the design, and nothing enforces it. An existence check before preferring `scopeRoot`, falling through to `gc.work_dir` when it fails, would keep the intended behavior without opening the new case.

This is warn-only today unless `GC_WORK_RECORD_ENFORCE` is set, so the current effect is spurious warnings rather than blocked closes. Since the point of the change is to make the contract safe to enforce, it is worth handling before that switch flips.

**2. `feature` looks like a real miss in the widened type set.**

`workRecordGatedBeadTypes` covers `""`, `task`, and `bug`. `feature` is a bd built-in described the same way task and bug are, it gets claimed and worked and closed with an outcome the same way, and `internal/runproj/summary.go:24` already groups `feature` with `bug` and `task` as engineering types. A `feature` closed as shipped with no commit still bypasses the contract, which is the defect this PR exists to close, one type over.

`epic`, `story`, `milestone`, and `decision` read as containers and their exclusion makes sense. `chore` and `spike` are arguable. `feature` is the clean one.

Happy to re-review once these are addressed.

AJBcoding:
Both addressed.

## 1. Repo root: conditioned on existence, not emptiness

Your reading checks out on both halves. `resolveStoreScopeRoot` falls back to the
city path when the store path is blank, and both `bdRigScopeTarget` and
`bdCityScopeTarget` route through it, so the `target.ScopeRoot` that reaches the
gate is never empty in production — the `gc.work_dir` branch was unreachable and
the change was an unconditional priority inversion, exactly as you described.
And `resolveBdScopeTarget` rejects a rig whose registered `Path` is empty, never
one whose directory has been moved or removed, so a stale registration still
yielded a usable-looking root.

`workRecordRepoDir` now returns the first candidate that is a directory on disk
— scope root first, then `gc.work_dir` — and `""` when neither is, using the
package's existing `isExistingDir`. Keeping `""` for the both-stale case
preserves the documented fail-closed reading rather than handing git a directory
it cannot open.

`TestEvaluateWorkRecordCloseGateAllowsShippedCloseAfterRigMoved` pins it end to
end as the mirror of the pruned-worktree test: rig directory moved away, live
worktree holding the commit. Against the previous behavior it fails with the
false block you predicted:

```
gc bd: work-record gate (enforced): close of wr-delivered: gc.work_commit f735fab… is not reachable on gc.work_branch work-branch
```

Unit cases cover the moved directory, a rig path replaced by a file, and both
roots stale.

## 2. `feature` added

Gated alongside `""`, `task`, and `bug`. `epic`, `story`, `milestone`, and
`decision` stay exempt and are now asserted as exempt in the table test, so the
boundary is explicit rather than incidental. I left `chore` and `spike` out —
you called them arguable and I have no evidence either way; say so and I'll add
them.

Restoring the previous behavior with both functions present now fails five of
the tests.

## One adjacent case I did not fold in

The scope root can be the *wrong* repo rather than a missing one: a city-scoped
bead whose work landed in a rig repo probes the city repo and reads unreachable.
We hit that live downstream on flipping `GC_WORK_RECORD_ENFORCE=1`. An existence
check cannot catch it and the fix is a different shape, so I left it out of this
PR — happy to file it separately if you want it tracked upstream.

## Note on the base refresh

You'll see a merge commit that isn't a review response. The branch was based
before #5111, so its macOS unit gate failed on the `t.TempDir()` path-spelling
mismatch from #4929 — `/var/folders/…` vs `/private/var/folders/…` — in tests
untouched by this PR. I merged current `main` (`b8e0edffb`) to pick up #5111 and
#5097 so the gate exercises this diff instead of that one.

Merge rather than rebase, deliberately: it leaves `8aefa6d86` — the commit you
reviewed — in place and keeps this a fast-forward rather than a force-push under
your review. Since `main` squash-merges, the merge commit disappears from the
final history either way. No review anchors were affected.

AJBcoding:
Ready for re-review — both points are addressed at `985f5d613`, CI green (59/59).

- **Repo root** — `workRecordRepoDir` now returns the first candidate that is a directory on disk (scope root, then `gc.work_dir`) and `""` when neither is, so the preference is conditioned on existence rather than emptiness. That closes the stale-rig-registration case you identified rather than relocating the block to it.
- **`feature`** — added to `workRecordGatedBeadTypes`. `epic`, `story`, `milestone`, and `decision` are now asserted exempt in the table test, so the boundary is explicit instead of incidental.

The 2026-08-09 comment above has the reasoning and test inventory; nothing has changed on the branch since.

Two things there are still open questions for you rather than settled:

1. Whether you want `chore` and `spike` gated as well — you called them arguable, and I have no evidence either way.
2. Whether the adjacent case should be tracked upstream: a scope root that is the *wrong* repo rather than a missing one (a city-scoped bead whose work landed in a rig repo probes the city repo and reads unreachable). An existence check cannot catch it and the fix is a different shape, so it stayed out of this PR. It is fixed and closed downstream; say the word and I will file it upstream.

One process note, since it explains the delay rather than anything about the code: this account gets a 404 from the review-request API on this repo, so I cannot put the PR back in your queue the normal way. The fixes landed 2026-08-09 21:20 and the reply went up at 22:09, but nothing signaled that to you — this comment is the only mechanism I have. Apologies for the extra notification.

AJBcoding:
@sjarmak — re-review ping, no new content.

Both points are addressed at `985f5d613` (CI green, 59/59). The detail is in [issuecomment-5244653850](https://github.com/gastownhall/gascity/pull/5120#issuecomment-5244653850) from earlier today rather than restated here.

Two things in that comment are open questions for you rather than settled, and are the reason it is worth opening:

1. Whether you want `chore` and `spike` gated as well — you called them arguable, and I have no evidence either way.
2. Whether the adjacent case should be tracked upstream: a scope root that is the *wrong* repo rather than a missing one. An existence check cannot catch it and the fix is a different shape, so it stayed out of this PR. It is fixed and closed downstream; say the word and I will file it upstream.

The ping itself is necessary because this account gets a 404 from the review-request API on this repo, so I cannot put the PR back in your queue the normal way — an @-mention is the only mechanism I have.

PR #5124, opened 2026-08-09
fix(supervisor): clear initStatus when an init failure is recorded

julianknutsen:
## Problem

**A city whose initialization fails after its runtime is built never retries.** It logs `init failure #1, next retry in 10s` and is then skipped forever.

`reconcileCities` picks cities to start by skipping any with a live `initStatus` entry:

```go
if _, initializing := initStatus[path]; initializing {
    continue
}
```

Three of the six failure branches in the start loop clear that entry; three do not — `city_runtime_failed`, `controller_state_failed`, and `pool_on_boot_failed`. After any of those, the city keeps an `initStatus` entry for the life of the supervisor process and is never selected again, so the backoff it just recorded can never be consulted.

The skip is also evaluated **before** the config-mtime backoff reset, which is why the obvious operator remedy does nothing: editing `city.toml` resets the backoff, but the city never reaches that code. `gc start` does not help either.

Measured tonight: maintainer-city was wedged out of the reconcile loop for **48 minutes**. Only `systemctl restart` cleared it.

The reachable branch is `city_runtime_failed`, and it is reachable exactly one way — `newCityRuntime`'s only error return is `storageBootGate`. A boot verdict that refuses (for example a stranded-bead convergence failure) therefore both stops the city *and* removes its ability to retry.

## Fix

Clearing `initStatus` becomes part of recording the failure, in `recordInitFailure`.

Every call site either already cleared it — the pre-runtime branches, and `publishManagedCity` for everything after it — or needs it cleared. There is no site that wants the entry to survive a recorded failure, so this is one fact with one owner rather than a courtesy each branch performs for itself. The three now-redundant `BatchUpdate` blocks are removed with it: net **18 insertions, 26 deletions**.

`emitPendingCityCreateFailure` does not read `initStatus`, so moving the delete after it changes nothing observable.

## Tests

`cmd/gc/supervisor_init_retry_test.go` drives the real `reconcileCities` against a registered city whose `[storage.classes]` describe a partial split. `storageSplitShapeOf` classifies that `storageSplitUnsupported`, so `storageBootGate` refuses before it constructs a registry, resolves a plan or reads a byte of a binding root — a deterministic `newCityRuntime` failure with no filesystem state to clean up, down the same branch production took.

| Test | Red-before |
|---|---|
| `TestReconcileCitiesClearsInitStatusOnRuntimeFailure` | `initStatus[…/gc-wedged-city-…] = {name:wedged-city status:building_city_runtime} after an init failure, want absent; the city is wedged out of the reconcile loop and its backoff/retry can never run` |
| `TestReconcileCitiesRetriesAfterInitBackoffElapses` | `after the backoff elapsed: init failure count = 1, want 2; the promised retry never ran because the leaked initStatus entry skips the city before the backoff is ever consulted (stderr: )` |
| `TestReconcileCitiesRetriesAfterConfigEdit` | `after editing city.toml the reconcile pass never attempted the city (stderr: ""); editing city.toml resets the backoff, but the leaked initStatus entry skips the city before the backoff is ever consulted, so only a supervisor restart recovers it` |

The empty `stderr` in rows 2 and 3 is the finding stated precisely: the second reconcile pass produced no output at all, because the city was never attempted.

Row 3 is the operator-observed symptom. Note that a config edit deliberately *resets* the failure record (the user may have fixed the config), so the retry's own failure is recorded fresh as `#1` — the proof of a retry is that the pass attempted the city at all, not that the count grew.

## Gates

- `go build ./...` clean · `go vet ./cmd/gc` clean · `gofmt` clean
- `go test ./cmd/gc -count=1 -timeout 40m` — **ok, 813s**

## Scope

Deliberately narrow. It does not touch the boot verdict, the backoff policy, or what makes `newCityRuntime` fail — only the leak that makes the recorded backoff unreachable.

### DIFF
--- a/cmd/gc/cmd_supervisor.go
+++ b/cmd/gc/cmd_supervisor.go
@@ -1916,7 +1916,22 @@ func reconcileCities(
 			})
 		}
 
-		// recordInitFailure logs the error and records backoff state.
+		// recordInitFailure logs the error, ends the attempt, and records
+		// backoff state.
+		//
+		// Clearing initStatus is part of recording the failure, not a courtesy
+		// each branch performs for itself. reconcileCities selects cities to
+		// start by skipping any with a live initStatus entry, and it consults
+		// that skip BEFORE the backoff and its config-mtime reset. So a branch
+		// that records a failure but leaves the entry behind wedges the city out
+		// of the loop for the life of the process: it logs "next retry in 10s"
+		// and is then never selected again, and neither editing city.toml nor
+		// `gc start` recovers it — only a supervisor restart does.
+		//
+		// Three of the branches below used to clear it and three did not, which
+		// is what the split produced. Every call site either has already cleared
+		// it (the pre-runtime branches, and publishManagedCity for everything
+		// after it) or needs it cleared, so the delete belongs here, once.
 		recordInitFailure := func(cityName, msg string) {
 			fmt.Fprintln(stderr, logutil.FormatFatalLine(msg))                              //nolint:errcheck // best-effort stderr
 			fmt.Fprintf(stderr, "gc supervisor: city '%s': %s (skipping)\n", cityName, msg) //nolint:errcheck
@@ -1926,10 +1941,11 @@ func reconcileCities(
 			}
 			cr.BatchUpdate(func(
 				_ map[string]*managedCity,
-				_ map[string]cityInitProgress,
+				initStatus map[string]cityInitProgress,
 				initFailures map[string]*initFailRecord,
 				_ map[string]*panicRecord,
 			) {
+				delete(initStatus, path)
 				ifrec := initFailures[path]
 				if ifrec == nil {
 					ifrec = &initFailRecord{}
@@ -1995,14 +2011,6 @@ func reconcileCities(
 				initStatus[path] = cityInitProgress{name: cityName, status: status}
 			})
 		}); err != nil {
-			cr.BatchUpdate(func(
-				_ map[string]*managedCity,
-				initStatus map[string]cityInitProgress,
-				_ map[string]*initFailRecord,
-				_ map[string]*panicRecord,
-			) {
-				delete(initStatus, path)
-			})
 			emitPendingCityCreateFailure(cr, path, cityName, "city_init_failed", err, stderr)
 			recordInitFailure(cityName, fmt.Sprintf("init: %v", err))
 			continue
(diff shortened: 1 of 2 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-98639467

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR fixes a supervisor init-failure wedge: a city that hit an init failure stayed stuck because `reconcileCities` skipped any city that still had a live `initStatus` entry before it ever reached backoff or config-mtime handling. That mattered because an init failure became effectively permanent — the city could not re-attempt startup even after its backoff elapsed or after the operator edited `city.toml`.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **Init-failure wedge:** `recordInitFailure` now clears the stale `initStatus` entry at the moment it records the failure, so a failed city is no longer skipped by the next reconcile pass. Review confirmed the cleanup lives with the single owner that ends the init attempt, and that the `initStatus`/`initFailures` mutations stay inside `cr.BatchUpdate`.
- **Retry after backoff:** a new regression test drives the real `city_runtime_failed` path and proves a second reconcile re-attempts the start once the backoff window expires.
- **Retry after config edit:** a second regression test proves that editing `city.toml` re-enters the start loop instead of remaining wedged behind stale init progress.

Review findings addressed:
- No blocker or major findings were raised. Both reviewers (Claude + Codex) independently validated the wedge fix and the new regression coverage across correctness, cleanup, and concurrency dimensions.

Remaining non-gating notes:
- Debuggability & Operability (nit, high / Claude): a comment near `recordInitFailure` slightly overstates that some pre-runtime branches "already cleared" `initStatus`, when those branches simply never set it — an optional wording tweak with no behavioral impact.
- Test Evidence Quality (nit, high / Claude): the new failure-path tests emit a non-fatal shutdown log for a city whose beads-provider stop-script was never materialized — an optional test-cleanup refinement.
- No remaining gating issues were found.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `02052fc4` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/31287499720

### Review Iterations
1 review pass performed; the PR was approved on the first pass with no fix iterations required.

---

PR #4726, opened 2026-07-27
fix(dispatch): make the in_progress crash-recovery work-query tier readiness-aware

jacobhausler:
Fixes #4725.

### What

The `in_progress + assigned (crash recovery)` work-query tier returned a candidate without
consulting gate or dependency state, and short-circuited ahead of the ready-gated tier. A step held
by an open human gate or an unclosed blocking dependency was therefore re-served to a worker on
every hook tick, forever.

Both tiers carrying that query are fixed together:
`standardAssignedInProgressWorkQueryScript` and `legacyControlAssignedInProgressWorkQueryScript`.

### How

A new `inProgressBlockedByEnrichmentScript` re-fetches the candidate with `bd show --json` and:

1. **Skips** it when any ready-blocking blocker is not closed -- `blocks`, `waits-for`,
   `conditional-blocks`, matching `beads.IsReadyBlockingDependencyType`. `parent-child` and
   `tracks` edges never suppress, which matters because every molecule step carries one to its root.
2. **Falls through** to the ready-gated tier rather than swallowing the tick, so a session holding
   one blocked step can still be served its other ready assigned work.
3. **Attaches `blocked_by`** to the rows it does serve, so the existing hook-side filter
   (`filterUnreadyHookCandidates` -> `isDepBlockedHookCandidate`) stops being a structural no-op on
   this tier.

The re-fetch is required rather than fastidious: `bd list --json` embeds bare edge rows with no
blocker status (`{issue_id, depends_on_id, type}`), while `bd show --json` emits
`{id, dependency_type, await_type, status}`. Only the latter can answer whether a blocker is still
open.

`bd ready` was deliberately NOT substituted for `bd list`: it excludes `in_progress` by design, so
that swap would silence the churn by disabling crash recovery altogether.

This is strictly narrowing -- the dispatcher serves a subset of what it served before. Nothing
becomes servable that was not servable already.

### Tests

New `internal/config/workquery_inprogress_blocked_test.go` executes the GENERATED SHELL against a
fake `bd` on PATH, so it pins observable behaviour rather than the script's spelling (the
byte-for-byte shape stays pinned by `TestWorkQueryGolden`):

- `TestInProgressTierSkipsGateBlockedCandidate` -- the reported incident shape.
- `TestInProgressTierSkipsDependencyBlockedCandidate` -- not gate-specific; a plain `blocks` edge
  suppresses identically.
- `TestInProgressTierServesUnblockedCandidate` -- crash recovery still works, and the row carries
[...]

PR #5125, opened 2026-08-09
fix(beads): refuse bd SQL reads of a relocated coordination class

julianknutsen:
## Problem

On a city that serves its infrastructure classes from a non-work `[storage]` binding, **`bd sql` cannot see those classes — and it does not error.**

Root cause, verified:

- **The relocated store is never named in this scope's `.beads/` metadata** (keys: `backend`, `database`, `dolt_database`, `dolt_mode`, `project_id`, `storage_database`, `storage_endpoint`). bd has no idea it exists.
- For the `sqlite-beads` provider it is a SQLite file **bd has no backend for at all** — `beads/internal/storage/` contains `dolt`, `embeddeddolt`, `doltutil`, `dbproxy`, `sqlbuild` and no sqlite. For the `beads-workspace` provider (merged in this PR's base, #5123) it is a *different* beads workspace with a `metadata.json` of its own. Either way this ledger cannot reach it.

So a query naming a `gcg-` bead resolves *this* workspace's metadata, runs **successfully** against the bd ledger, finds no rows, and returns an empty set. Nothing errors because nothing failed.

That empty answer is indistinguishable from a true negative, and downstream it is read as one. **Production cost:** the silent empty surfaced as `root_missing` — live molecule roots reported absent and whole frontiers reported empty — and cost an operator most of a night.

**bd cannot detect this.** It cannot know a class was relocated out from under it. The guard has to live on the gc side, where storage routes and coordination classes are known.

## Fix

A named refusal at the surfaces that reach the bd ledger with an id-scoped read:

| Surface | Silent failure before | Now |
|---|---|---|
| `BdStore.ReleaseIfCurrent` | CAS matches zero rows → `released=false`, which every caller reads as *"someone else holds it"* and backs off from | refuses, naming the bead |
| `BdStore.fetchReadyProjection` | contract is *"these are the active rows"*, so an invisible row silently keeps its nil `is_blocked` forever | **partitions** — the relocated ids drop out of the request and every bd-resident row is still enriched |
| `gc bd sql` | confident empty + exit 0 — the surface the incident came through | refuses before `bd` is executed |
| `gc bd query` | `[]` + exit 0 for `id=gcg-*`, the same wrong answer one word away | refuses before `bd` is executed |

The refusal names the class, its reserved id namespace, where the beads are actually served from, why bd answered emptily rather than failing, and the one by-id read that actually routes by class:

```
gc bd: bd cannot read a relocated coordination class: bd sql: graph-class beads
(id prefix "gcg-") are served from the "infra" storage binding (provider sqlite-beads,
.gc/store). This bd ledger does not serve those classes and holds no row under their
reserved id prefixes, so a read scoped to one cannot match here — and bd does not fail:
it runs the read successfully against this ledger and returns an empty result
indistinguishable from a real one. Read these beads with `gc beads show <id>`, which
routes by class through the controller API (GET /v0/city/{cityName}/bead/{id}). That API
lane is the only class-routed by-id read there is: `gc beads show` falls back to a
work-store scan when no controller is reachable, and `gc bd show` and `gc bd dep tree`
are raw bd passthroughs against this same ledger.
```

Matchable with `errors.Is(err, beads.ErrBdSQLClassRelocated)`.

**It is deliberately not keyed on bd's error text.** The existing `isBdSQLUnsupportedInEmbeddedMode` substring match (`"bd sql"` + `"not yet supported"` + `"embedded mode"`) is left alone rather than generalized: it classifies a **loud failure** — bd refusing `bd sql` in embedded mode, where the command errored and a `dolt sql` fallback takes over — on a disjoint code path. This bug is the **silent success** path. Folding them together would make "bd refused" and "bd answered about the wrong ledger" the same condition, and would put a fragile string match on the load-bearing check. The new guard reads routing, not prose.

### What counts as naming a relocated id

Detection is by **reserved class id-prefix** (`gcg`/`gcm`/`gcs`/`gco`/`gcn`), and — this is the part the first round got wrong — only in an **id-shaped position**:

| Dialect | Anchor | Matches | Passes |
|---|---|---|---|
| `bd sql` | start of a string literal (`'`, `"`, backtick) or start of statement | `id = 'gcg-1'`, `id like 'gcg-%'`, `id like 'gcg%'`, `id in ('bd-1','gcg-2')` | `metadata like '%gcg-1%'`, `notes like '%see gcg-1%'`, `/* gcg-9 */ …`, `… -- gcg-x`, `'mygcg-1'`, `'x-gcg-1'`, `'gcgabc'` |
| `bd query` | value side of a comparison / grouping token (bd's lexer skips whitespace) | `id=gcg-1`, `id=gcg-*`, `id = gcg-1`, `parent=gcg-root`, `(id=gcg-1 OR …)` | `status=open`, `id=bd-42`, `title="fix gcg-1 regression"`, `id=mygcg-1` |

`gc bd list`/`ready` stay unguarded — they answer about the ledger they are scoped to and claim nothing more. `gc bd show`/`dep tree` stay unguarded too, but **not** because they are safe (see below): a bare id is not a dialect this scan can classify without refusing every work id that starts with the same letters.

### Escape hatch

`GC_BD_ALLOW_RELOCATED_CLASS_READ=1` runs a refused `gc bd sql`/`query` anyway.

It exists because the scan classifies *text*, and text is not always decidable: `json_extract(metadata,'$."gc.drain_control_id"') = 'gcg-abc'` is a correct, non-empty question about **this** ledger's rows, and is textually identical to an id-scoped predicate. Without a knob the guard boxes an operator out of a ledger they can still read, during exactly the incident it was built for.

It is scoped to the CLI pre-flight on purpose — the store-level guards protect the controller's own automated reads, where no human is present to judge and an override would be a silent correctness hole — and honoring it is never quiet: gc prints the full refusal it is overriding to stderr before running the query.

## Review round: seven majors, fixed

Reproduced against branch head `71eee9ac` before fixing.

| # | Finding | Fix |
|---|---|---|
| 1 | Matcher fired on legitimate work-ledger queries (`metadata like '%gcg-abc%'`, a `json_extract` metadata compare, SQL comments), **and the message asserted the opposite** — that the query "would … match nothing, and return an empty result". `ensureDrainUnitConvoy` stamps `gc.drain_control_id = <graph control id>` onto a convoy `coordclass` deliberately keeps work-class, so the work ledger provably holds `gcg-` strings in its JSON metadata. | Anchored the match to a string-literal start; restated the message as a property of the prefix ("holds no row under their reserved id prefixes … cannot match here") instead of a prediction about the query's result; added the override for what anchoring cannot decide. |
| 2 | **Failed open behind any bd root flag.** Keyed on `bdArgs[0]`, but `-C/--directory`, `--json`, `--db`, `--actor`, `-q` are bd *root* persistent flags accepted before the subcommand, and `gc bd` forwards argv verbatim. | Verb is the first non-flag token, skipping bd root value-flag values (`bdflags`). An **unrecognized** flag fails CLOSED: the verb is undecidable, so every remaining argument is scanned under every dialect rather than the guard disengaging. |
| 3 | **`gc bd query` was an unguarded sibling** with the identical silent-empty — its DSL pushes `id=<v>` to an id equality and `id=<v>*` to `id LIKE '<v>%'`, then prints `[]` and exits 0. | Guarded, with a matcher for that dialect. |
| 4 | **The refusal recommended `gc bd show` / `gc bd dep tree` as "the federated verbs".** They are raw passthroughs: `doBd` resolves a scope and then `exec.Command(bdPath, bdArgs...)` with `cmd.Dir` at that scope root — no coordination-class routing anywhere on the path. Following the message ran the blind read it had just prevented. The rationale comment enshrined the same false premise. | Message points at `gc beads show <id>`; comment corrected; both facts pinned by tests (below). |
| 5 | **Ready-projection guard was whole-batch.** `CachingStore` prime/reconcile hand *every* active bead to it, so one stray relocated id returned the whole slice unenriched — `is_blocked` nil for every row, every cycle, forever (call sites only `recordProblem` and continue; the refusal is a pure function of config, so it never heals). Live consequence: `compute_awake_bridge.go` reads `IsBlocked != nil && *IsBlocked`, so a permanently-nil projection reverts the #4726 "don't wake a session on blocked in_progress work" suppression city-wide. | Partitions instead of refusing. |
| minor | LIKE patterns omitting the hyphen (`'gcg%'`, `'gcg_%'`) escaped the matcher though they select the same rows. | Trailing-boundary rule now accepts `-`, LIKE/glob wildcards, and any non-id byte; still rejects `'gcgabc'`. |
| minor | Why-clause hardcoded "bd has no SQLite backend and the relocated store is not named in `.beads/metadata.json`" — **false for the `beads-workspace` provider merged in this PR's own base** (#5123), where the relocated store is a bd workspace with a `metadata.json` of its own. | Restated provider-agnostically. |
| minor | Doc comment cited `TestRelocatedBeadClassesAgreeWithStorageRoutes`, which does not exist. | Now cites the real `TestRelocatedBeadClassesAgreeWithClassStoreRouting`. |

### Which verb is actually federated — verified, not assumed

The whole point of finding 4 is that a plausible-sounding recommendation was wrong, so the replacement was verified by execution rather than by reading:

- `internal/api/handler_beads.go` `beadStoresForID` has a class-prefix arm returning `[graph, work]` for a reserved `gcg-` id. `humaHandleBeadGet` consumes it. **`TestBeadGetResolvesARelocatedGraphID`** (new) drives that handler end to end against a bead seeded **only** in the relocated graph store and asserts it resolves — the fixture first asserts the work store does *not* hold the id, so the test cannot pass vacuously.
- The route is **`GET /v0/city/{cityName}/bead/{id}`**, not `/v0/beads/{id}` as the review reported. The message names the real one.
- `gc beads show <id>` reaches it via `routeReadCmdWithHooks` → `api.Client.GetBead`. Its **no-API fallback** (`doBeadsShowFallback` → `openAllConvoyStoresAt`) opens city + rig *work* stores and is **not** class-routed — so the message says exactly that, rather than implying the command is federated unconditionally.
- `gc graph <id>` is **not** federated either (`openRigAwareStore` → `openStoreAtForCity`), so it is not recommended.
- `gc bd show` / `gc bd dep tree` are raw passthroughs, proved by **`TestGcBdShowIsARawPassthroughOnASplitCity`** (new): driven through this PR's own split-city fixture, the stub `bd` receives `show gcg-abc123` and `dep tree gcg-abc123` verbatim, exit 0.

> **Note for `/data/projects/workflows/scripts/pr_review.py`:** `city_dep_subtree` (line ~875) documents "Only the controller-federated `gc bd show` and `gc bd dep tree` resolve graph beads". That is **false for this repo's CLI path** — both are raw `bd` passthroughs here. The federation that would make it true (`cmd/gc/cmd_bd_show_fed.go`, `maybeRouteBdShowViaAPI`) exists only on unmerged branches (`124bca8c39`, `0252c5867c`, `05400bd93a`) and is not an ancestor of `origin/main`. That script is relying on a deploy-lineage binary, not on this base, and should be re-pointed at `gc beads show` / the API.

## Single-store compatibility

Structural, not argued. `relocatedBeadClasses` derives purely from the `[storage.classes]` assignment (no filesystem, same answer before and after a migration). A city with no `[storage]` section, or one leaving every class on the reserved `work` binding, gets `nil` — so `bdStoreOptionsForConfig` adds no option, no `BdStore` carries a relocated class, and `bdSQLRelocatedClassRefusal` returns before it looks at the query.

**Proven by mutation at all three layers.** The exact query a split city refuses is asserted to pass through *unchanged* on a single-store city:

- `TestReleaseIfCurrentIsByteIdenticalWithoutRelocation` — asserts the full argv `[bd sql --json UPDATE issues SET ... WHERE id = 'gcg-abc123' ...]`
- `TestReadyProjectionIsByteIdenticalWithoutRelocation` — asserts the argv equals `[bd sql <readyProjectionSQL()> --json]` and both rows still enrich
- `TestGcBdSQLIsUnchangedOnASingleStoreCity` — runs `doBd` end to end against a stub `bd` on `PATH` and asserts the stub received the unmodified `gcg-abc123` query
- `TestBdSQLRelocatedClassRefusalIsInertOnASingleStoreCity` — now covers all four newly-guarded shapes (`sql`, `sql` behind a root flag, `query`, and the unrecognized-flag fail-closed path)
- `TestRelocatedBeadClassesIsEmptyForASingleStoreCity`, `TestBdStoreOptionsCarryRelocatedClassesOnlyWhenSplit`, `TestRelocatedClassesInSQLIsInertWithoutRelocation` (now also asserts `RelocatedClassesInQueryExpr` is inert)

`TestRelocatedBeadClassesAgreeWithClassStoreRouting` pins the anti-drift invariant: for every class, `resolveClassStore` routing it off the work store and `relocatedBeadClasses` naming it must be the same answer. A read relocated with its guard left behind is the exact failure this PR fixes.

## Tests — red-before evidence

Every new test was run against branch head `71eee9ac` before the fix. 13 failures in `internal/beads`, 26 in `cmd/gc`. Representative output:

**Matcher over-match (finding 1)**
```
--- FAIL: TestRelocatedClassesInSQLMatchesOnlyAtIDBoundaries/like_contains_on_a_text_column
    RelocatedClassesInSQL("select id,title from issues where metadata like '%gcg-abc%'") matched = true, want false
--- FAIL: TestRelocatedClassesInSQLMatchesOnlyAtIDBoundaries/block_comment_mention
    RelocatedClassesInSQL("/* related: gcg-9 */ select id from issues where status='open'") matched = true, want false
--- FAIL: TestRelocatedClassesInSQLMatchesOnlyAtIDBoundaries/trailing_line_comment
    RelocatedClassesInSQL("select 1 -- gcg-x") matched = true, want false
--- FAIL: TestRelocatedClassesInSQLMatchesOnlyAtIDBoundaries/like_without_the_hyphen
    RelocatedClassesInSQL("select id from issues where id like 'gcg%'") matched = false, want true
```

**Fail-open behind a root flag (finding 2)**
```
--- FAIL: TestBdSQLRelocatedClassRefusalOnASplitCity/leading_--json
    bdSQLRelocatedClassRefusal([--json sql select * from issues where id = 'gcg-abc']) refused = false, want true
--- FAIL: TestBdSQLRelocatedClassRefusalOnASplitCity/leading_-C_dir
    bdSQLRelocatedClassRefusal([-C /tmp/x sql select * from issues where id = 'gcg-abc']) refused = false, want true
--- FAIL: TestBdSQLRelocatedClassRefusalFailsClosedOnAnUnrecognizedFlag/unknown_value_flag_hides_the_verb
    bdSQLRelocatedClassRefusal([--frobnicate x sql select * from issues where id = 'gcg-abc']) refused = false, want true
--- FAIL: TestGcBdSQLRefusesBehindALeadingRootFlagOnASplitCity/--json
    doBd([--json sql select id, status from issues where id = 'gcg-abc123']) exited 0 on a graph-blind query
```

**Unguarded `query` (finding 3)**
```
--- FAIL: TestBdSQLRelocatedClassRefusalOnASplitCity/query_naming_a_graph_id
    bdSQLRelocatedClassRefusal([query id=gcg-abc123]) refused = false, want true
--- FAIL: TestGcBdQueryRefusesAGraphClassQueryOnASplitCity
    doBd exited 0 on a graph-blind query
```

**Wrong recommended verb (finding 4)**
```
--- FAIL: TestRelocatedClassRefusalDoesNotRecommendABlindVerb
    refusal still recommends a raw bd passthrough as federated:
    … Use the federated `gc bd show <id>` or `gc bd dep tree <id>` …
--- FAIL: TestRelocatedClassRefusalNamesEverythingAnOperatorNeeds
    refusal is missing "gc beads show <id>"
    refusal is missing "GET /v0/city/{cityName}/bead/{id}"
```

**Whole-batch ready projection (finding 5)**
```
--- FAIL: TestReadyProjectionPartitionsARelocatedIDOutOfTheBatch
    enrichReadyProjectionForCache: bd sql cannot read a relocated coordination class: ready projection: …
--- FAIL: TestReadyProjectionSkipsBdEntirelyWhenEveryIDIsRelocated
    enrichReadyProjectionForCache: bd sql cannot read a relocated coordination class: ready projection: …
```

**False why-clause (minor)**
```
--- FAIL: TestRelocatedClassRefusalNamesEverythingAnOperatorNeeds
    refusal is missing "holds no row under their reserved id prefixes"
    refusal is missing "cannot match here"
```

The original round-one revert proofs still hold: removing the `ReleaseIfCurrent` guard, the `doBd` guard, the `bdStoreOptionsForConfig` wiring, or dropping graph from `relocatedBeadClasses` each fails its own named test.

## Reader/writer agreement

Every `bd sql` site and every `ReleaseIfCurrent` implementor was enumerated; the moved read and its counterpart agree:

- **bd-ledger ad-hoc read verbs:** `sql` and `query` — both guarded, both classified with a dialect-appropriate matcher. `list`/`ready`/`show`/`dep` unguarded, deliberately, with the rationale corrected (show/dep are blind, not federated).
- **`bd sql` invocation sites:** `BdStore.ReleaseIfCurrent` (refuses), `BdStore.fetchReadyProjection` (partitions), `gc bd sql`/`query` (refuses). `cmd/gc/dolt_sql_health.go` is left as-is (a deliberately Dolt-specific health probe), as is the `dolt sql -q` embedded fallback (it runs against the embedded Dolt directory, not the relocated binding).
- **`BdStore` construction sites:** `bdStoreForCity` / `bdStoreForRig` / `controlBdStoreForCity` / `controlBdStoreForRig` all funnel through the single `bdStoreOptionsForConfig` choke point and receive the guard — which covers the main open path (`openStoreAtForCityWithConfig`, used by `doBdReleaseIfCurrent`), the API rig stores, and control/convoy dispatch. The remaining constructors (`scopedBdStoreFor*`, `hookClaimBdStore*`, `runBdStoreBridge`, `NewBeadsLibStore`, the t3bridge watcher) were checked individually: none calls `ReleaseIfCurrent` and none but the t3bridge watcher is `CachingStore`-wrapped, so none reaches either SQL path with city class routing in play. The t3bridge watcher wraps a rig worktree's own work ledger and has no city config in scope.
- **Ready-projection consumers:** `CachingStore.PrimeActive` (`caching_store.go:730`), `CachingStore.Prime` (`:846`) and `runReconciliation` (`caching_store_reconcile.go:321`) all take `items, err` and only `recordProblem` on error — which is precisely why a whole-batch refusal never healed. With the partition they now receive a fully enriched slice for every bd-resident row. `DoltliteReadStore.enrichReadyProjectionForCache` returns items unchanged by design and is unaffected.
- **`ReleaseIfCurrent` implementors:** a relocated graph class already serves its own CAS (`storebinding/sqlite/graph_frontdoor.go`, `storebinding/lifecycle_wrappers.go`, `storebinding/beads_adapter.go`). The guard makes bd refuse exactly the ids those own — the two sides partition the id space instead of overlapping on a wrong empty answer.
- **Error consumers:** `releasePoolAssignmentIfCurrent` downgrades to its unconditional-recheck fallback only on `ErrConditionalReleaseUnsupported`. `ErrBdSQLClassRelocated` is not that, so it takes the log-and-skip arm — loud, and with no downgrade to a write that could clobber a concurrent re-claim.
- **`internal/bdflags`:** gains `GlobalBoolFlags()` alongside `GlobalValueFlags()`, both pinned by completeness tests against `bd --help`. The existing `SplitGlobalFlags` consumer (`mistypedMetadataPairRefusal`) is untouched.
- **`dispatch_control_ready.go`** already routes a relocated graph class to the binding store and only opens the (guarded) control bd store otherwise — reader and guard agree there by construction.

## Gates

| Gate | Result |
|---|---|
| `go build ./...` | pass |
| `go vet ./...` | pass |
| `gofmt -l ./cmd ./internal` | clean |
| `golangci-lint run ./internal/beads/... ./internal/bdflags/... ./internal/api/... ./cmd/gc/...` (2.10.1) | **0 issues** |
| `go test ./internal/beads/ -count=1` | ok |
| `go test ./internal/bdflags/ -count=1` | ok |
| `go test ./internal/api/ -count=1` | ok |
| `go test ./cmd/gc -count=1 -timeout 40m` | ok |

## Deliberately not in this PR

Converting the two Go callers off raw SQL to typed store calls is a separate change and is **not** here — this PR stops the bleeding without waiting for it. See the follow-up note in the PR discussion: `bd unclaim --if-assignee` (bd 1.1.0, landed in beads `4b1d9c875`) now provides the native CAS release verb the `BdStore.ReleaseIfCurrent` SEAM comment was explicitly waiting for, which changes that conversion's design materially.

Federating `gc bd show` / `gc bd dep tree` themselves is also out of scope. This PR only stops the refusal from *claiming* they are federated.


julianknutsen:
## Follow-up: converting the two Go callers off raw SQL

Scoped but intentionally **not** in this PR. Two findings change its design from what was assumed when the work was scoped.

### 1. `BdStore.ReleaseIfCurrent` — bd has shipped the verb the SEAM was waiting for

The SEAM comment on `ReleaseIfCurrent` says:

> When bd ships its native issueops CAS release verb, consume it HERE as the first attempt: probe by invoking the verb and fall back to this `bd sql` path when bd reports the command unknown (older pinned bd).

**It has landed.** bd 1.1.0 (`bd version 1.1.0 (7b55d00e7: d3f82bd81028)`) advertises:

- `bd unclaim <id> --if-assignee <assignee>` — *"an atomic compare-and-swap (the inverse of claim): the issue is released only while it is still assigned to the given assignee. If the holder differs ... nothing is changed and bd exits nonzero with an error naming the current holder."*
- `bd update <id> --if-assignee <v> --if-status <v>` — *"a mismatch writes nothing and exits 13 (vs 1 for other failures)"*

beads-side: `UnclaimIssueIfAssignee` / `UnclaimIssueIfAssigneeInTx` in `internal/storage/{storage.go,dolt/issues.go,issueops/unclaim.go}`, ported to proxied-server mode in `4b1d9c875` (beads#5364).

So the conversion is a gc-side change, but not the hand-rolled typed conditional update originally scoped — it is "consume bd's verb behind a capability probe". That is strictly better and unlocks a large deletion: `bdSQLStringLiteral` (hand-rolled MySQL string-literal escaping, and **the only SQL-injection surface in this path** — it exists solely because the statement is assembled by hand), `releaseIfCurrentViaEmbeddedDoltSQL`, `isBdSQLUnsupportedInEmbeddedMode` and `parseDoltRowsAffected` all become dead, because `bd unclaim` works in embedded mode where `bd sql` does not.

What it still needs, and why it is its own PR: a version/capability probe (gascity supports `bd_compatibility = "bd-1.0.4"`, which predates the verb), an exit-13 vs exit-1 classifier that does not regress `ErrConditionalReleaseUnsupported` for `releasePoolAssignmentIfCurrent`'s fallback contract, and a rewrite of the tests that currently pin the exact generated SQL string.

### 2. `BdStore.fetchReadyProjection` — the typed source may already exist

`bdIssue` already decodes `is_blocked` (`bdstore.go:683`) and `toBead` already stamps it onto `Bead.IsBlocked` (`bdstore.go:844`), so `bd list --json` carries the column. The SQL projection is a **backfill** for rows where the list response omits it (`optionalBool` distinguishes absent from false), behind the `bdReadyProjectionMinVersion = "1.0.5"` gate.

If that omission no longer happens on the pinned bd, the correct fix is to delete the projection rather than re-route it — a behavior change on the cache prime/reconcile hot path that needs measurement against a real bd, not a refactor. Determining that is the first task of the follow-up, and it is the reason this was not bundled here: a partially converted read path is worse than an unconverted one.

The guard in this PR covers both callers in the meantime.

julianknutsen:
Coordination — I was about to edit `internal/beads/bdstore_ready_projection.go` for an adjacent fix and found this PR first. **Standing down on that file**; flagging what I measured independently in case it is useful, plus two things this guard may not cover.

### Independent confirmation

Same root cause, reached from the other end (a live `maintainer-city`, molecule `gcg-arn`). Your diagnosis matches exactly: a `bd sql` read of a `gcg-` bead succeeds, returns zero rows, and the empty is indistinguishable from a true negative. It surfaced as `blocked:root_missing` / `no runnable frontier`, and yes — it cost most of a night before the empty was recognised as a store that was never consulted.

### Two things beyond `bd sql`

`bd sql` is not the only projection with this shape. On the same city, same molecule:

```
gc bd show gcg-arn                          -> resolves        (by-id federates)
gc bd list --id gcg-arn,gcg-o4l,gcg-3l9     -> 3 beads         (by-id federates)
gc bd dep tree gcg-arn --json               -> 48 beads        (federates)
gc bd graph gcg-arn --json                  -> 61 beads        (federates)

gc bd ready --include-ephemeral --limit 0   -> 5 beads, ZERO gcg-*     <-- silent empty
gc bd list --metadata-field gc.root_bead_id=gcg-arn -> 0 rows          <-- silent empty
```

So **`ready` and `list --metadata-field` fail the same silent way**, and neither is a `sql` call — they would slip past a guard scoped to `bd sql`. Worth deciding whether the relocation guard belongs at a lower choke point than the SQL path.

Also note `dep tree` (48) and `graph` (61) disagree on membership, so they are not interchangeable for frontier work — `graph` includes beads `dep tree` does not.

### On "the Python review tooling was converted to `gc bd dep tree`"

That conversion does not appear to have landed in `workflows/scripts/pr_review.py`. On `origin/main` today it still resolves the molecule root and children via `bd sql` — `rig_graph_children` and `rig_show_many`, both emitting `select ... from wisps union all select ... from issues`. That is the code path producing `root_missing` in production right now. Filed separately as gastownhall/workflows#29.

### Who takes what

Owner direction (this session) is to **federate the projections in gc** rather than work around them in the driver — a driver-side fix would bake city store layout (`.gc/infra/.beads/beads.db`) into a workflows script, which is the wrong layer.

I have not written any of it. Happy to take the projection federation if you are only doing the SQL guard here, or to stay off it entirely if you are already on it — say which and I will not touch `internal/beads/`.

julianknutsen:
## Review round: seven majors fixed, pushed as `15d3c32`

All seven were reproduced against `71eee9ac` before fixing. 13 red-before failures in `internal/beads`, 26 in `cmd/gc`. PR body has the full table, the anchoring rules, and the red-before output.

**1 — the matcher over-matched, and the message asserted something false.** The only boundary check was "the byte before `gcg-` is not an id-body byte", so it fired on `metadata like '%gcg-abc%'`, a `json_extract` metadata compare, `/* related: gcg-9 */ …`, `select 1 -- gcg-x`. Those are real work-ledger queries: `ensureDrainUnitConvoy` stamps `gc.drain_control_id = <graph control id>` onto a convoy `coordclass` deliberately keeps work-class, so the bd ledger provably holds `gcg-` strings in its JSON metadata — and the refusal claimed the query "would … match nothing, and return an empty result". Now anchored to an id-shaped position (start of a string literal, or of the statement), and the message states the property that is actually true — this ledger holds no row under the reserved prefix, so a read scoped to one *cannot match here* — instead of predicting the query's result.

**Escape hatch: `GC_BD_ALLOW_RELOCATED_CLASS_READ=1`.** Anchoring cannot decide everything: `json_extract(metadata,'$."gc.drain_control_id"') = 'gcg-abc'` is a correct, non-empty question about *this* ledger and is textually identical to an id-scoped predicate. Rather than box an operator out of a ledger they can still read, the CLI pre-flight takes an override — scoped there deliberately (the store-level guards protect the controller's own automated reads, where no human is present to judge), and never silent: gc prints the full refusal it is overriding before running the query. Env var rather than a flag because `gc bd` sets `DisableFlagParsing` and forwards argv verbatim; adding a third gc-only token to a documented raw passthrough would change the contract for every verb and risk shadowing a future bd flag.

**2 — failed open behind any bd root flag.** `bdArgs[0]` is not the verb when `-C`, `--json`, `--db`, `--actor`, `-q` lead. Verb is now the first non-flag token, skipping bd root value-flag values; an **unrecognized** flag fails CLOSED (verb undecidable → scan every remaining arg under every dialect). `internal/bdflags` gains `GlobalBoolFlags()` with a completeness pin, so a known valueless flag is distinguishable from an unknown one.

**3 — `gc bd query` was an unguarded sibling.** `id=gcg-*` → `id LIKE 'gcg-%'` against the same ledger → `[]` + exit 0. Guarded with a matcher for that dialect (ids anchored to the value side of a comparison, whitespace-tolerant like bd's lexer), so `id=gcg-*` refuses and `title="fix gcg-1 regression"` passes.

**4/6/7 — the refusal recommended two commands carrying the bug it reports.** Verified rather than assumed, because that is the whole finding:

- `gc bd show` / `gc bd dep tree` are **raw passthroughs** — `doBd` resolves a scope and then `exec.Command(bdPath, bdArgs...)`, no class routing on the path. Driven through this PR's own split-city fixture, the stub `bd` receives `show gcg-abc123` and `dep tree gcg-abc123` verbatim, exit 0. Now pinned by `TestGcBdShowIsARawPassthroughOnASplitCity`.
- The genuinely class-routed read is the API's `beadStoresForID` graph arm. **`TestBeadGetResolvesARelocatedGraphID`** (new) drives `humaHandleBeadGet` end to end against a bead seeded **only** in the relocated graph store, after first asserting the work store does not hold it — so it cannot pass vacuously.
- Two corrections to the review's own suggestion: the route is **`GET /v0/city/{cityName}/bead/{id}`**, not `/v0/beads/{id}`; and `gc graph <id>` is **not** federated either (`openRigAwareStore` → `openStoreAtForCity`), so it is not recommended.
- The message names `gc beads show <id>` and is explicit that **the API lane is the only class-routed by-id read there is** — `gc beads show`'s no-API fallback (`doBeadsShowFallback` → `openAllConvoyStoresAt`) opens city + rig *work* stores and is not class-routed. Saying "use `gc beads show`" without that caveat would have repeated the original sin at one remove.

> **`/data/projects/workflows/scripts/pr_review.py` is relying on the same wrong assumption.** `city_dep_subtree` (~line 875) documents "Only the controller-federated `gc bd show` and `gc bd dep tree` resolve graph beads". False for this repo's CLI path. The federation that would make it true (`cmd/gc/cmd_bd_show_fed.go`, `maybeRouteBdShowViaAPI`) exists only on `124bca8c39` / `0252c5867c` / `05400bd93a` — `git merge-base --is-ancestor` reports **NOT ancestor of `origin/main`** for all three, and the file is absent from this head. That script is written against a deploy-lineage binary and should be re-pointed at `gc beads show` / the API.

**5 — the ready-projection guard was whole-batch.** `CachingStore` prime/reconcile hand *every* active bead to it, so one stray relocated id returned the whole slice unenriched — `is_blocked` nil for every row, every cycle, forever (call sites only `recordProblem` and continue; the refusal is a pure function of config, so it never heals). It now **partitions**: relocated ids drop out of the request, everything bd-resident is still enriched, and a dropped id keeps its last cached value — the documented benign state (`preserveCachedReadyProjectionLocked`), exactly what its absence produced before this guard existed. The single-id refusal on `ReleaseIfCurrent` is unchanged.

**Minors:** LIKE patterns omitting the hyphen (`'gcg%'`, `'gcg_%'`) now match — they select the same rows. The why-clause no longer claims "bd has no SQLite backend and the relocated store is not named in `.beads/metadata.json`", which is **false for the `beads-workspace` provider merged in this PR's own base (#5123)** where the relocated store is a bd workspace with a `metadata.json` of its own. And the doc comment citing the nonexistent `TestRelocatedBeadClassesAgreeWithStorageRoutes` now cites the real `TestRelocatedBeadClassesAgreeWithClassStoreRouting`.

### Two corrections to the review itself

- The `bd sql --out /tmp/gcg-dump.csv` example is **not reachable**: `bd sql` is `cobra.ExactArgs(1)` and its only local flag is `--csv` (bool). No value-taking `bd sql` flag exists, so no flag-value handling was added for that reason. (`bd query`'s `--sort`/`--limit` do take values, but they are field names and integers.)
- The claim that a migrated split city's work ledger "permanently holds `gcg-`/`gcs-`/`gco-` rows by design" does **not** hold. `importInfraSnapshot` uses `CreateWithForeignID` and preserves the copied rows' **original** ids, which were minted under the HQ/rig prefix; reserved class prefixes are only ever minted by the relocated class engine. So retained residue does not carry a reserved prefix, before or after cutover. Finding 5 stands anyway — the whole-batch blast radius is wrong for any trigger, however rare — and the partition is now the behavior. The `RelocatedClass` doc comment records this reasoning so the next reader does not have to re-derive it.

### Gates

`go build ./...` · `go vet ./...` · `gofmt -l ./cmd ./internal` clean · `golangci-lint 2.10.1` over `internal/beads` + `internal/bdflags` + `internal/api` + `cmd/gc` → **0 issues** · `go test ./internal/beads/ ./internal/bdflags/ ./internal/api/ -count=1` ok · `go test ./cmd/gc -count=1 -timeout 40m` ok (792s).

Single-store byte-identity is still proven by mutation, and `TestBdSQLRelocatedClassRefusalIsInertOnASingleStoreCity` now covers all four newly-guarded shapes (`sql`, `sql` behind a root flag, `query`, and the unrecognized-flag fail-closed path).

julianknutsen:
**CI fix pushed** (`da7bc187a8`) for the `Integration / packages-core-3-of-4` failure (`TestGCEnvReadBaseline`), which was also failing `CI / integration` and `CI / required`.

The ratchet fired correctly — this branch added `GC_BD_ALLOW_RELOCATED_CLASS_READ` without updating `internal/testenv/testdata/gc_env_read_baseline.golden`. Its failure message offers two exits: move the read to `internal/rollout`, or update the golden deliberately. **Took the second**, because the knob fails all three conditions of the registry's own litmus (`internal/rollout/registry.go`):

1. **`SelectsBetween` needs two mechanical code paths.** This selects between *refusing* a read and *performing* it — one passthrough with a pre-flight in front. There is no honest second path to name.
2. **A rollout/migration gate must name when it dies** (`Expires` + `VersionAnchor`). This one has no death date: it exists because the scan classifies *text*, and a work-ledger query that legitimately mentions a relocated id in a JSON/text column isn't decidable from the text. That's permanent. Only `InfraKillswitch` may omit those fields, and a killswitch is a process-level emergency off — not a per-invocation operator judgment about one query.
3. **`Spec.ConfigPath` is mandatory and reflection-verified against `config.City`**, and `Resolve` precedence is builtin < config < env. Registering the gate means minting a `city.toml` field, and that field — not the env var — becomes the knob's durable state. A city that persists "allow relocated-class reads" has disarmed the guard for every operator and every later invocation. Being per-invocation and persisting nowhere is exactly what makes the override safe.

Mechanically it doesn't fit either: `rollout.Flags` is constructed at controller boot, in doctor, and at store open. This pre-flight runs *before any store is opened* — that's the point, it must refuse before `bd` is exec'd — and holds no `Flags`. `GC_WORK_RECORD_ENFORCE` (already in the golden) is the in-tree precedent for the shape: same `gc bd` seam, same truthy switch, same CLI-only operator scope. A short note recording that decision now sits at the const so the next reader doesn't redo the analysis.

Also considered and **rejected on evidence**: adding the var to `testenv.LeakVectorVars`, on the theory that an operator exporting it during an incident could leak it into a bare `go test` and invert the refusal tests. `cmd/gc`'s own `TestMain` already scrubs it (`liveEnvKeysForTests()` unsets every ambient `GC_*`/`BEADS_*`/`DOLT_*` that isn't an explicit test-control key) — running the refusal tests with `GC_BD_ALLOW_RELOCATED_CLASS_READ=1` exported confirms they still pass. The entry would have guarded a hole that's already closed.

No behavior change; the ratchet still trips on the next new `GC_*` var.

**Gates:** `go build ./...` clean · `go vet` clean · `gofmt` clean · golangci-lint 2.10.1 **0 issues** on `./cmd/gc/...` + `./internal/testenv/...` · `go test ./internal/testenv/` **ok (19.5s — previously-failing check now green)** · `go test ./internal/beads/` **ok (19.6s)** · `go test ./cmd/gc` **ok (868s)** · pre-commit hook (lint + doc-gen + vet) passed.

PR #5126, opened 2026-08-09
fix(beads): release claims through bd's CAS verb, not hand-built SQL

julianknutsen:
## Problem

`BdStore.ReleaseIfCurrent` is a compare-and-swap — clear an in-progress assignment only while the bead still carries the expected assignee — and it was assembled as raw SQL, with the verdict read out of `rows_affected`:

```sql
UPDATE issues SET status='open', assignee='', updated_at=CURRENT_TIMESTAMP
 WHERE id=<lit> AND status='in_progress' AND assignee=<lit>
```

Three costs:

1. **A hand-rolled escaping surface.** `bdSQLStringLiteral` hand-escapes MySQL string literals — an injection surface that exists *only* because the statement is concatenated by hand.
2. **`bd sql` does not work in embedded mode.** Verified against bd 1.1.0: `Error: 'bd sql' is not yet supported in embedded mode`. So the path needed a *second* implementation shelling directly to `dolt sql` (`releaseIfCurrentViaEmbeddedDoltSQL` + `parseDoltRowsAffected`), selected by matching bd's error prose.
3. **It could not see the ephemeral tier.** The statement names the `issues` table. A bead in the ephemeral tier lives in `wisps`, so it matched zero rows and the CAS returned `released=false` — *"someone else holds it"* — no matter who held it. **gascity runs formula steps as wisps, so this silent no-op was live.**

And a CAS that matches zero rows is indistinguishable from one that could not run — the same failure shape as #5125.

## Fix

bd now expresses both preconditions natively (landed in beads `4b1d9c875`, gastownhall/beads#5364):

```
bd update <id> --if-assignee <expected> --if-status in_progress --status open --assignee ""
```

and — unlike every other bd failure, which is exit 1 with a message — reports a **rejected precondition as its own exit status, 13**, having written nothing.

**The CAS verdict is therefore a number, not a parse of bd's prose.** That is the main reason to prefer the verb: the load-bearing branch stops depending on message text. `TestReleaseIfCurrentPreconditionMissDoesNotClassifyOnMessageText` pins it from both sides — exit 13 with unrelated wording is still a miss, and "assignee mismatch" prose at exit 1 is still a real error.

This is exactly what the in-code SEAM comment was waiting for (*"When bd ships its native issueops CAS release verb, consume it HERE as the first attempt … fall back to this `bd sql` path when bd reports the command unknown (older pinned bd)"*), implemented as prescribed.

### Old-bd compatibility

`deps.env` pins `BD_PREV_VERSION=v1.0.4` as the contract-tested minimum, and it predates the flags. So **the raw-SQL path stays**, selected by a one-way per-store latch the first time bd rejects the flag as unknown. The latch is paid once, not per release (`TestReleaseIfCurrentLatchesTheOldBdFallback`).

The unknown-flag match is **anchored** to the flag name immediately following the parser's marker, because a capable bd's cobra usage echo lists every flag it *has* whenever any flag is wrong — a floating `strings.Contains` would latch a capable store into the SQL path forever. `isBdUnknownIfRevisionFlag` now shares that matcher rather than duplicating the five spellings.

### What did *not* get deleted, and why

Because the bd-1.0.4 fallback survives, so do `bdSQLStringLiteral`, `releaseIfCurrentViaEmbeddedDoltSQL`, `parseDoltRowsAffected` and `isBdSQLUnsupportedInEmbeddedMode`. **The hand-rolled MySQL-escaping injection surface is narrowed to the bd-1.0.4 fallback, not removed.** It can be deleted the day `BD_PREV_VERSION` moves past beads#5364 — at which point all four helpers go together. Deleting them now would break the minimum supported bd, which is contract-tested.

## Behavior change (deliberate)

**The CAS now covers the ephemeral tier.** bd's verb resolves the id across `issues` and `wisps`, so a wisp-held assignment is now genuinely released instead of always reporting "someone else holds it". Pinned by the `an ephemeral bead releases too` integration subtest, verified against real bd (`tst-wisp-2ci` → `released=true`, state `open`/`""`).

Everything else is verdict-for-verdict identical: success → `(true, nil)`; precondition miss → `(false, nil)`; unresolvable id → `(false, nil)` (what `rows_affected=0` meant); anything else → error.

## Reader/writer agreement

**Writers of the `(status, assignee)` pair**

| Writer | Mechanism | Agreement |
|---|---|---|
| `BdStore.Claim` | `bd update --claim` | the CAS now reads the same row through the same bd id resolution the claim wrote through |
| `BdStore.Update` | `bd update --status/--assignee` | same verb family as the new CAS |
| `BdStore.ReleaseIfCurrent` | **converted** | was `bd sql`, now the native verb |
| `releaseIfCurrentViaEmbeddedDoltSQL` | direct `dolt sql` | retained, bd-1.0.4 only |
| `MemStore` / `FileStore` / `NativeDoltStore` / `SQLiteStore` / `storebinding` graph front door | own backends | untouched — none used bd SQL |
| external `bd unclaim` / `bd reclaim` | bd-native | unaffected |

**Readers of the verdict**

| Reader | Agreement |
|---|---|
| `releasePoolAssignmentIfCurrent` | **Contract preserved.** The verb path never returns `ErrConditionalReleaseUnsupported`, so the recheck-fallback arm is unchanged; a precondition miss returns `(false, nil)`, the pre-existing "authoritative, do not retry" arm. A mismatch can never downgrade into a clobbering unconditional write. |
| `doBdReleaseIfCurrent` (`gc bd release-if-current`) | exit 13 maps to `skipped` / exit 0, exactly as `rows_affected=0` did |
| `CachingStore.ReleaseIfCurrent` | refreshes only on `released=true`; unchanged |
| `beadPolicyStore.ReleaseIfCurrent` | delegates; unchanged |
| `storebinding/beads_nudge_queue.go` (3 sites) | consumes the bool; verdicts unchanged |

Both callers of the converted path are covered at *both* layers — store (`internal/beads`) and CLI (`cmd/gc`) — for *both* the verb path and the bd-1.0.4 fallback. No half-converted path.

## Tests — red-before evidence

Each production change reverted in isolation:

**A. Remove the verb preference from `ReleaseIfCurrent`**
```
--- FAIL: TestReleaseIfCurrentPrefersTheNativeVerb
    ReleaseIfCurrent: bd release-if-current: parsing SQL result: unexpected end of JSON input
--- FAIL: TestReleaseIfCurrentReadsPreconditionMissFromTheExitCode
    a precondition miss must not be an error, got bd release-if-current: unexpected command [sql --json UPDATE issues SET ...]
```

**B. Classify the CAS verdict on message text instead of the exit code**
```
--- FAIL: .../exit_13_with_unrelated_prose_is_still_a_miss
    released=false err=bd release-if-current: exit status 13: bd said something, want false/nil
--- FAIL: .../mismatch_prose_at_exit_1_is_a_real_error
    a non-13 failure must surface as an error, not a silent false
```

**C. Remove the latch**
```
--- FAIL: TestReleaseIfCurrentLatchesTheOldBdFallback
    verb probes = 3 across 3 releases, want exactly 1 (latched)
```

**D. Make the unknown-flag match unanchored (`strings.Contains`)**
```
--- FAIL: TestReleaseIfCurrentUnknownFlagMatchIsAnchored
    a capable bd was latched into the SQL fallback: [[bd update ... --if-assignee ...] [bd sql --json UPDATE issues SET ...]]
```

**E. Remove the verb preference (CLI layer)**
```
--- FAIL: TestDoBdReleaseIfCurrentUsesNativeVerbAndReadsExit13
    doBdReleaseIfCurrent = 1, want 0 (exit 13 is a verdict, not a failure)
```

**Old-bd fallback stays byte-identical.** Every pre-existing test that pinned the generated SQL is now a *fallback*-path test: its runner rejects the verb as bd 1.0.4 does, and the asserted argv is unchanged, including the escaping cases (`worker-'1`, `bd-\42`). `TestReleaseIfCurrentFallsBackToSQLOnAnOldBd` asserts the full fallback argv `[bd sql --json UPDATE issues SET ... WHERE id = 'bd-42' AND status = 'in_progress' AND assignee = 'worker-''1']`.

**Real-bd integration row** (`//go:build integration`), reusing the existing `newConditionalIntegrationBdStore` harness, in **embedded** mode — the mode where `bd sql` is rejected outright:

```
--- PASS: TestBdStoreReleaseIfCurrentAgainstRealBd (12.81s)
    --- PASS: .../a_wrong_assignee_writes_nothing_and_is_not_an_error (0.98s)
    --- PASS: .../the_matching_assignee_releases (1.13s)
    --- PASS: .../a_second_release_is_a_no-op,_not_an_error (0.62s)
    --- PASS: .../an_ephemeral_bead_releases_too (1.71s)
    --- PASS: .../an_unresolvable_id_is_not_held (0.95s)
    --- PASS: .../a_non-exact_id_refuses_instead_of_releasing_a_different_bead (2.02s)
```

**Correction to an earlier revision of this description:** that run is against a bd that HAS the flags. `deps.env` pins `BD_VERSION=v1.1.0`, and the released v1.1.0 predates them (they landed in beads#5008 / #5364, three weeks later) — a local binary that self-reports `bd version 1.1.0` is a dev build from a later commit. **On the bd CI installs, this row now SKIPS** on a positive capability probe rather than hard-failing; see "Review fixes" below. Moving `BD_VERSION` past #5364 is what makes it (and the headline wisp fix) live on shipped bd; that bump is deliberately not in this PR.

Its red-before is otherwise a **bd** change, not a gc revert: it is the authoritative guard that the flag spelling still exists and that a rejected precondition is still exit 13. The exception is the new collision leg, which is red on a gc revert — see below.

## Review fixes (`392a3ce196`)

Four defects found in review, two of them this PR's own red CI. Every fix is verified red-before by reverting only its production change.

### 1. The verb rode bd's fuzzy ID resolver, bypassing the gcy-g4o exact-ID guard — BLOCKING DEFECT

The statement the verb replaced matched `WHERE id = <literal>` and was **exact by construction**. `bd update <id>` is not: with no exact hit, bd prefix/substring-resolves. Measured against real bd, with `tst-d6e` held `in_progress` by `worker-1`:

```
$ bd update tst-d6 --if-assignee worker-1 --if-status in_progress --status open --assignee ""
✓ Updated issue: tst-d6e — probe row      # exit 0
```

The caller never named `tst-d6e`. The CAS preconditions are no defense — bd evaluates them against the bead it *resolved*, and one worker commonly holds several beads — so `gc bd release-if-current tst-d6 worker-1` printed `released`, exited 0, and revoked a live claim. The bead is then re-dispatched while the worker still believes it holds it.

This is exactly the incident class `cmd/gc/cmd_bd.go`'s "Pre-flight exact-ID guard for write-mutating subcommands (gcy-g4o)" and `beads.ErrIDCollision` exist for (`gcy-dv7` → `gcy-wisp-dv78`), and gascity is systematically exposed because formula steps run as `<prefix>-wisp-<suffix>` wisps. `gc bd release-if-current` is dispatched at `cmd_bd.go:238`, **before** the guard at `:253`, and `bdMutationWriteIDs` does not recognize the subcommand — it was the one `gc bd` mutation with no exact-ID verification.

**Fix:** the verb path resolves the id through `Get` first and refuses on `ErrIDCollision`, so every store-level caller is covered — including the CLI that bypasses the `cmd_bd.go` guard. Only a *confirmed* collision blocks: a plain `ErrNotFound` or a failed read falls through, the same fail-open tradeoff `cmd_bd.go`'s guard documents. `ReleaseIfCurrent` now documents the canonical-full-ID requirement the way `Update` does. **The raw-SQL fallback is untouched** — it still matches the id literally, so no preflight and no argv change there.

Scope note: bd prefers an exact id when one exists, so in-process Go callers passing canonical ids were never exposed. The reachable surface is ids with no exact match — truncated, typo'd, stale — which is precisely what the documented CLI entry point accepts from humans and agents.

### 2. Text classification, reintroduced one line below the exit-code branch

The PR's thesis is "stop classifying CAS verdicts by message text, use exit 13". One line below that branch, `if isBdNotFound(runErr)` did exactly what the PR removes: an **unanchored** `strings.Contains(msg, "not found")` over the *whole* runner error, returning `(false, nil)` — the authoritative "someone else holds it, do not retry" verdict.

Measured at the head before this fix, `ReleaseIfCurrent` answered `(released=false, err=nil)` for all of:

| Failure | Text |
|---|---|
| bd off PATH / `BD_BIN` drift | `exec: "bd": executable file not found in $PATH` |
| renamed or unmounted database | `Error 1049 (42000): database "fe" not found on Dolt server at …` |
| hosted-gateway misroute | `404 page not found` |
| missing workspace | `beads workspace not found: …` |

On main every one of these was an error. Downstream, `gc bd release-if-current` printed `skipped` and exited 0 against a dead store; `lifecycle_wrappers.go` recorded `observeClaim("ReleaseIfCurrent", false, nil)` — a clean refusal — in claim telemetry; `releasePoolAssignmentIfCurrent` logged "assignment changed since snapshot" instead of "conditional release failed".

**Fix:** a local `isBdIssueNotFound` replaces the loose package-wide helper on this branch. It requires **numeric evidence** (`bdExitCode == 1`; 13 is the precondition miss) **AND** bd's measured issue-scoped wording (`no issue found` / `no issues found` / `issue not found` — real bd emits `Error resolving tst-nosuchbead: no issue found matching "tst-nosuchbead"` at exit 1). A spawn failure carries no process exit status at all, so it can never reach the verdict. Everything else flows to `return false, true, runErr`. This also settles the reported MINOR.

### 3. CI red — resource-census ledger (`packages-core-1-of-4`)

`TestRepositoryLedgerMatchesCensusAndDocumentation` failed with `scope=untagged resource=subprocess calls=404 (baseline 403), files=113 (baseline 112)` in three scopes.

The cause is **not** the production file's `os/exec` import — `census.go:901` skips non-`_test.go` files ("Sibling Go source supplies package-level declaration context but is never counted"). It is the new **untagged test** file's `exec.Command("sh", "-c", fmt.Sprintf("exit %d", code))`, used to fabricate a real `*exec.ExitError`. Proven by replacing that one call at the PR head with the production import left alone: the ledger goes green.

The ledger's own invariant is `untagged subprocess call/file totals cannot grow; reductions must lower this baseline`, with `resource_owner = each process-owning test removes or replaces its source call site` — so the sanctioned fix is to remove the call site, not to bump the three lockstepped baselines. `bdExitCode` now matches an `ExitCode() int` interface, which `*exec.ExitError` satisfies through its embedded `*os.ProcessState`; a compile-time `var _ exitCoder = (*exec.ExitError)(nil)` in the production file keeps the substitution honest.

```
$ go test -tags integration ./internal/testpolicy/resourcecensus
ok      github.com/gastownhall/gascity/internal/testpolicy/resourcecensus  3.5s
```

### 4. CI red — the integration row FAILED instead of skipping (`packages-core-2-of-4`)

`isUnsupportedOnThisBd` could never fire. On a flagless bd the verb probe returns `handled=false`, the store latches, `bd sql` is embedded-rejected, and `releaseIfCurrentViaEmbeddedDoltSQL` then **succeeds** with `released=false` for a wisp (it names the `issues` table) — so `err == nil` and the ephemeral leg reached `t.Fatal`.

**Fix:** the row gates on a **positive capability signal** — `bd update --help` advertising both flags, the same shape the production `conditionalWritesCapable` probe uses for `--if-revision` — and skips the whole test cleanly. All five error-string skip guards are deleted; after the gate, an error is a bug.

Verified against a shim that reproduces the CI-pinned bd (rejects `--if-assignee`/`--if-status`, real bd otherwise), with dolt on PATH:

```
# PR head b73b24a165 — byte-identical to CI job 93186748769
--- FAIL: TestBdStoreReleaseIfCurrentAgainstRealBd/an_ephemeral_bead_releases_too
    bdstore_conditional_release_integration_test.go:120: did not release an ephemeral in-progress assignment

# with this fix
--- SKIP: TestBdStoreReleaseIfCurrentAgainstRealBd (4.97s)
    installed bd does not advertise --if-assignee/--if-status (pre-beads#5364) …
ok  github.com/gastownhall/gascity/internal/beads  36.2s   # whole package, -tags integration, flagless bd
```

### Red-before for the review fixes

**Remove the exact-ID preflight** — store, CLI, and real bd:
```
--- FAIL: TestReleaseIfCurrentRefusesAFuzzyIDCollision
    ReleaseIfCurrent released a bead the caller never named: released=true err=<nil>, want an error wrapping ErrIDCollision

--- FAIL: TestDoBdReleaseIfCurrentRefusesAFuzzyIDCollision
    doBdReleaseIfCurrent = 0, want 1 (a substring collision must not be reported as a release); stdout="released\n" stderr=""

--- FAIL: TestBdStoreReleaseIfCurrentAgainstRealBd/a_non-exact_id_refuses_instead_of_releasing_a_different_bead
    ReleaseIfCurrent("tst-68") = released=true err=<nil>, want a refusal wrapping ErrIDCollision
```

**Restore `isBdNotFound` on the not-held branch:**
```
--- FAIL: TestReleaseIfCurrentSurfacesInfraFailuresAsErrors/a_renamed_or_unmounted_database
    an infrastructure failure became a CAS verdict: released=false err=<nil>
--- FAIL: .../bd_fell_off_PATH            an infrastructure failure became a CAS verdict: released=false err=<nil>
--- FAIL: .../a_hosted_gateway_misroute   an infrastructure failure became a CAS verdict: released=false err=<nil>
--- FAIL: .../the_beads_workspace_is_gone an infrastructure failure became a CAS verdict: released=false err=<nil>
```

### Contracts preserved

- A genuine CAS miss is still `(false, nil)` and never `ErrConditionalReleaseUnsupported`, so `releasePoolAssignmentIfCurrent` cannot downgrade to a clobbering unconditional write. A collision and an infra failure are *errors*, which that caller already handles by skipping the tick.
- **bd 1.0.4 behavior is byte-identical.** The preflight is on the verb path only; the fallback argv and `bdSQLStringLiteral` escaping are unchanged, including `worker-'1` and `bd-\42`. `legacyBdReleaseRunner` answers the preflight so every SQL-pinning test still asserts exactly the same statement.

## Gates

| Gate | Result |
|---|---|
| `go build ./...` | pass |
| `go vet ./...` | pass |
| `gofmt -l ./cmd ./internal` | clean |
| `golangci-lint run ./internal/beads/... ./cmd/gc/...` (2.10.1) | **0 issues** |
| `go test ./internal/beads/ ./internal/storebinding/...` | all ok (20.0s / 11.8s / 19.0s) |
| `go test ./cmd/gc -timeout 40m` | ok 815.5s |
| `go test -tags integration ./internal/beads` (capable bd) | ok 44.0s |
| `go test -tags integration ./internal/beads` (CI-pinned flagless bd) | ok 36.2s — row skips |
| `go test -tags integration ./internal/testpolicy/resourcecensus` | ok 3.5s (was the `packages-core-1-of-4` red) |
| `.githooks/pre-commit` | pass (lint-changed 0 issues, doc-gen clean, `go vet`) |

`internal/beads` is package index 21 → shard **2 of 4**; `internal/testpolicy/resourcecensus` is index 148 → shard **1 of 4** — the two jobs that were red.

Based on `a1d8e21c7d`. Independent of #5125; the two touch adjacent lines in `ReleaseIfCurrent` and will need a one-line merge resolution whichever lands second.

### DIFF
--- a/cmd/gc/cmd_bd_test.go
+++ b/cmd/gc/cmd_bd_test.go
@@ -2435,7 +2435,14 @@ prefix = "fe"
 	}
 	fakeBin := t.TempDir()
 	sqlLog := filepath.Join(fakeBin, "sql.log")
+	// bd 1.0.4 — the contract-tested minimum (deps.env BD_PREV_VERSION) — has no
+	// conditional-release flags and rejects them the way its flag parser does,
+	// which is what latches the store onto the raw-SQL path this test pins.
 	fakeBD := "#!/bin/sh\n" +
+		"if [ \"$1\" = \"update\" ]; then\n" +
+		"  printf 'unknown flag: --if-assignee\\n' >&2\n" +
+		"  exit 1\n" +
+		"fi\n" +
 		"if [ \"$1\" = \"sql\" ] && [ \"$2\" = \"--json\" ]; then\n" +
 		"  printf '%s\\n' \"$3\" > " + strconv.Quote(sqlLog) + "\n" +
 		"  printf '{\"rows_affected\":1,\"schema_version\":1}\\n'\n" +
@@ -2476,6 +2483,82 @@ prefix = "fe"
 	wantQuery := "UPDATE issues SET status = 'open', assignee = '', updated_at = CURRENT_TIMESTAMP WHERE id = 'fe-abc' AND status = 'in_progress' AND assignee = 'worker-1'"
 	if strings.TrimSpace(string(query)) != wantQuery {
 		t.Fatalf("SQL query = %q, want %q", strings.TrimSpace(string(query)), wantQuery)
+	}
+}
+
+// TestDoBdReleaseIfCurrentUsesNativeVerbAndReadsExit13 is the CLI twin of the
+// store-level conversion: on a bd that HAS the conditional-release flags, the
+// command must issue the native verb (never SQL) and must read bd's exit 13 as
+// an authoritative "not released", printing skipped and exiting 0 — not as a
+// command failure.
+func TestDoBdReleaseIfCurrentUsesNativeVerbAndReadsExit13(t *testing.T) {
+	clearInheritedBeadsEnv(t)
+	origCityFlag, origRigFlag := cityFlag, rigFlag
+	defer func() { cityFlag, rigFlag = origCityFlag, origRigFlag }()
+	cityFlag, rigFlag = "", ""
+
+	cityDir := t.TempDir()
+	rigDir := filepath.Join(cityDir, "frontend")
+	if err := os.MkdirAll(filepath.Join(rigDir, ".beads"), 0o755); err != nil {
+		t.Fatalf("mkdir rig .beads: %v", err)
+	}
+	if err := os.WriteFile(filepath.Join(cityDir, "city.toml"), []byte(`[workspace]
+name = "demo"
+
+[beads]
+provider = "file"
+
+[[rigs]]
+name = "frontend"
+path = "frontend"
+prefix = "fe"
+`), 0o644); err != nil {
+		t.Fatalf("write city.toml: %v", err)
+	}
+	if err := os.WriteFile(filepath.Join(rigDir, ".beads", "metadata.json"), []byte(`{"database":"dolt","backend":"dolt","dolt_mode":"embedded","dolt_database":"fe"}`), 0o644); err != nil {
+		t.Fatalf("write rig metadata: %v", err)
+	}
+
+	fakeBin := t.TempDir()
+	argvLog := filepath.Join(fakeBin, "argv.log")
+	// A capable bd: the verb is understood, and the precondition miss is
+	// reported the way bd reports it — exit 13, nothing written.
+	fakeBD := "#!/bin/sh\n" +
+		"printf ' %s' \"$@\" >> " + strconv.Quote(argvLog) + "\n" +
+		"if [ \"$1\" = \"update\" ]; then\n" +
+		"  printf 'assignee mismatch\\n' >&2\n" +
+		"  exit 13\n" +
+		"fi\n" +
+		"printf 'unexpected bd args:' >&2\n" +
+		"printf ' %s' \"$@\" >&2\n" +
+		"exit 2\n"
+	if err := os.WriteFile(filepath.Join(fakeBin, "bd"), []byte(fakeBD), 0o755); err != nil {
+		t.Fatalf("write fake bd: %v", err)
+	}
+	t.Setenv("PATH", fakeBin+string(os.PathListSeparator)+os.Getenv("PATH"))
+	setCwd(t, cityDir)
+	t.Setenv("GC_CITY_PATH", cityDir)
+	t.Setenv("GC_BEADS_FORCE_FALLBACK", "1")
+
+	target := execStoreTarget{ScopeRoot: rigDir, ScopeKind: "rig", Prefix: "fe"}
+	var stdout, stderr bytes.Buffer
+	if got := doBdReleaseIfCurrent(cityDir, nil, target, "fe-abc", "worker-1", &stdout, &stderr); got != 0 {
+		t.Fatalf("doBdReleaseIfCurrent = %d, want 0 (exit 13 is a verdict, not a failure); stderr=%q", got, stderr.String())
+	}
+	if strings.TrimSpace(stdout.String()) != "skipped" {
+		t.Fatalf("output = %q, want skipped", stdout.String())
+	}
+	argv, err := os.ReadFile(argvLog)
+	if err != nil {
+		t.Fatalf("read argv log: %v", err)
+	}
+	for _, want := range []string{"update fe-abc", "--if-assignee worker-1", "--if-status in_progress", "--status open"} {
+		if !strings.Contains(string(argv), want) {
+			t.Errorf("bd argv %q does not contain %q", string(argv), want)
+		}
+	}
+	if strings.Contains(string(argv), "UPDATE issues SET") {
+		t.Fatalf("a capable bd still received hand-built SQL: %q", string(argv))
 	}
 }
 
--- a/cmd/gc/cmd_bd_test.go
+++ b/cmd/gc/cmd_bd_test.go
@@ -2562,6 +2562,80 @@ prefix = "fe"
 	}
 }
 
+// TestDoBdReleaseIfCurrentRefusesAFuzzyIDCollision covers the one `gc bd` write
+// that never reaches the pre-flight exact-ID guard: release-if-current is
+// dispatched before that block and bdMutationWriteIDs does not recognize it, so
+// the store-level guard is the only thing standing between an operator's
+// abbreviated id and bd's substring resolver. Without it the command prints
+// "released", exits 0, and revokes a live claim on a bead nobody named (gcy-g4o:
+// "gcy-dv7" → "gcy-wisp-dv78").
+func TestDoBdReleaseIfCurrentRefusesAFuzzyIDCollision(t *testing.T) {
+	clearInheritedBeadsEnv(t)
+	origCityFlag, origRigFlag := cityFlag, rigFlag
+	defer func() { cityFlag, rigFlag = origCityFlag, origRigFlag }()
+	cityFlag, rigFlag = "", ""
+
+	cityDir := t.TempDir()
+	rigDir := filepath.Join(cityDir, "frontend")
+	if err := os.MkdirAll(filepath.Join(rigDir, ".beads"), 0o755); err != nil {
+		t.Fatalf("mkdir rig .beads: %v", err)
+	}
+	if err := os.WriteFile(filepath.Join(cityDir, "city.toml"), []byte(`[workspace]
+name = "demo"
+
+[beads]
+provider = "file"
+
+[[rigs]]
+name = "frontend"
+path = "frontend"
+prefix = "fe"
+`), 0o644); err != nil {
+		t.Fatalf("write city.toml: %v", err)
+	}
+	if err := os.WriteFile(filepath.Join(rigDir, ".beads", "metadata.json"), []byte(`{"database":"dolt","backend":"dolt","dolt_mode":"embedded","dolt_database":"fe"}`), 0o644); err != nil {
+		t.Fatalf("write rig metadata: %v", err)
+	}
+
+	fakeBin := t.TempDir()
+	argvLog := filepath.Join(fakeBin, "argv.log")
+	// A capable bd whose resolver prefix-matches "fe-abc" onto the wisp
+	// "fe-wisp-abc9" — and which would happily perform the write.
+	fakeBD := "#!/bin/sh\n" +
+		"printf ' %s' \"$@\" >> " + strconv.Quote(argvLog) + "\n" +
+		"if [ \"$1\" = \"show\" ]; then\n" +
+		"  printf '[{\"id\":\"fe-wisp-abc9\",\"status\":\"in_progress\",\"assignee\":\"worker-1\"}]\\n'\n" +
+		"  exit 0\n" +
+		"fi\n" +
+		"exit 0\n"
+	if err := os.WriteFile(filepath.Join(fakeBin, "bd"), []byte(fakeBD), 0o755); err != nil {
+		t.Fatalf("write fake bd: %v", err)
+	}
+	t.Setenv("PATH", fakeBin+string(os.PathListSeparator)+os.Getenv("PATH"))
+	setCwd(t, cityDir)
+	t.Setenv("GC_CITY_PATH", cityDir)
+	t.Setenv("GC_BEADS_FORCE_FALLBACK", "1")
+
+	target := execStoreTarget{ScopeRoot: rigDir, ScopeKind: "rig", Prefix: "fe"}
+	var stdout, stderr bytes.Buffer
+	if got := doBdReleaseIfCurrent(cityDir, nil, target, "fe-abc", "worker-1", &stdout, &stderr); got != 1 {
+		t.Fatalf("doBdReleaseIfCurrent = %d, want 1 (a substring collision must not be reported as a release); stdout=%q stderr=%q", got, stdout.String(), stderr.String())
+	}
+	if strings.TrimSpace(stdout.String()) != "" {
+		t.Fatalf("output = %q, want nothing on stdout", stdout.String())
+	}
+	if !strings.Contains(stderr.String(), "fe-wisp-abc9") {
+		t.Fatalf("stderr = %q, want the bead bd actually resolved", stderr.String())
+	}
+	argv, err := os.ReadFile(argvLog)
+	if err != nil {
+		t.Fatalf("read argv log: %v", err)
+	}
+	if strings.Contains(string(argv), "--if-assignee") {
+		t.Fatalf("the release still reached bd for a colliding id: %q", string(argv))
+	}
+}
+
 // TestGcBdPassthroughResolvesBdBinary pins the binary the `gc bd`
 // passthrough execs. A city whose scope carries a complete storage binding
 // runs the bd its workspace PATH pins — an ambient bd that cannot speak the
--- a/internal/beads/bdstore.go
+++ b/internal/beads/bdstore.go
@@ -309,6 +309,12 @@ type BdStore struct {
 	readyProjectionChecked bool
 	readyProjectionEnabled bool
 
+	// condReleaseLatchedUnsupported records that this bd rejected the
+	// conditional-release flags, pinning ReleaseIfCurrent to the raw-SQL
+	// fallback for the rest of the process (bdstore_conditional_release.go).
+	condReleaseMu                 sync.Mutex
+	condReleaseLatchedUnsupported bool
+
 	// Conditional-write (ConditionalWriter) capability state, populated lazily on
 	// the first conditional write (bdstore_conditional.go). condWriteProbed/
 	// condWriteCapable memoize the four-verb --if-revision probe; condWriteLatched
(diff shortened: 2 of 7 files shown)


julianknutsen:
## Measured: the ready projection stays. It is load-bearing, and the typed verb cannot replace it.

The second half of this work — converting `BdStore.fetchReadyProjection` off raw `bd sql` — was measured against real bd 1.1.0 before choosing, per instruction. **The measurement reversed my earlier inference, and the answer is "do not convert".** Evidence below; nothing here is argued from source-reading alone.

### 1. `bd list --json` does NOT carry `is_blocked`. Deleting the projection is off the table.

I previously inferred from the decoder (`bdIssue.IsBlocked`, `bdstore.go:683`) that `bd list --json` already carries the column. **That inference was wrong.** Isolated bd 1.1.0 workspace, two beads, `probe-isw` blocked by `probe-0mz`:

```
$ bd link probe-isw probe-0mz --type blocks
Linked: probe-isw (blocked task) depends on probe-0mz (blocker task) (blocks)

$ bd list --json
[ { "id": "probe-0mz", ... "dependency_count": 0, "dependent_count": 1, "comment_count": 0 },
  { "id": "probe-isw", ... "dependencies": [ ... ], "dependency_count": 1, ... } ]
```

No `is_blocked` key on either row — not on the blocked one, not on the unblocked one. The decoder field exists but bd's `list` never populates it. `bd ready --json` correctly returns only `probe-0mz`, so blockedness is *knowable*, but `list` does not report it. **The SQL projection is the only source of `is_blocked`. Deleting it would silently drop the field for every bead.**

### 2. The typed verb exists and covers the right tiers — but it cannot express "unknown".

`bd blocked --json` works and is not a tier gap. From bd's own source, `GetBlockedIssuesInTx` (`internal/storage/issueops/blocked.go:243`):

```go
for _, table := range []string{"issues", "wisps"} {
    ... SELECT id FROM %s WHERE is_blocked = 1 AND status <> 'closed' AND status <> 'pinned'
```

Same two tables as the projection's `union all`, same denormalized column. So a re-route to `bd blocked` would be tier-correct.

**It still cannot be used.** `bd blocked` returns only the blocked *set*; the projection returns `id, is_blocked` as a **tri-state**, and the tri-state is load-bearing:

- `bdstore_ready_projection.go:127` — `if row.ID == "" || !row.IsBlocked.set { continue }` — a NULL `is_blocked` is deliberately skipped, leaving `Bead.IsBlocked` nil.
- The in-tree rationale for that skip: bd's denormalized column *"flaps NULL<->false for ephemeral mail wisps"*, and enriching them made the reconciler *"re-emit `bead.updated` for every open mail bead on every cycle (an event flood that starves gc-hook work queries)"*.
- `caching_store_reconcile.go` preserves the cached value for any row the projection omits, precisely so `is_blocked` never *"flips false->nil and beadChanged emits a spurious bead.updated"*.

With `bd blocked`, "not in the returned set" collapses NULL and false into an explicit `false`. That is exactly the flapping input those two mechanisms exist to suppress. **Converting would re-open a known production incident (the cache-reconcile event flood) to remove a query that is not currently misbehaving.**

### 3. The path already degrades safely, and #5125 covers its relocated-class hazard.

`enrichErr` is non-fatal — `caching_store_reconcile.go` records a problem and keeps the unenriched rows, so a projection failure costs a nil `is_blocked`, not a broken reconcile. (Worth noting: `bd sql` is rejected in embedded mode — `Error: 'bd sql' is not yet supported in embedded mode` — so embedded cities have *always* run with the projection erroring into that arm.) The relocated-class hazard on this path is handled by the guard in #5125.

### Verdict

**Leave `fetchReadyProjection` on `bd sql`.** What it covers that `list` does not: the denormalized `is_blocked` column across **both** the durable and ephemeral tiers, **with NULL distinguishable from false**. No bd verb currently returns that tri-state.

Improving it further needs a **bd-side** change, not a gc-side workaround — either `bd list --json` emitting `is_blocked`, or a projection verb returning the tri-state (`bd blocked --include-unblocked`, or `--explain`-style output over active rows). Per the standing instruction, I am reporting that rather than forcing a gc-side conversion. Until then the correct gascity change is none.

julianknutsen:
## Review fixes pushed — `392a3ce196`

Four defects fixed. Two were this PR's own red CI. Not merging; re-running checks.

**1. The verb rode bd's fuzzy ID resolver (gcy-g4o).** The statement it replaced matched `WHERE id = <literal>` — exact by construction. `bd update <id>` prefix/substring-resolves. Reproduced against real bd: with `tst-d6e` held `in_progress` by `worker-1`, `bd update tst-d6 --if-assignee worker-1 --if-status in_progress --status open --assignee ""` exits **0** and clears `tst-d6e`, a bead the caller never named. The preconditions do not help — bd evaluates them against the *mis-resolved* bead. `gc bd release-if-current` is dispatched at `cmd_bd.go:238`, before the exact-ID guard at `:253`, and `bdMutationWriteIDs` does not recognize the subcommand, so it was the one `gc bd` mutation with no verification.

Fixed inside `BdStore.ReleaseIfCurrent`'s verb path (`Get` → refuse on `ErrIDCollision`), so every store-level caller is covered including the CLI. Only a confirmed collision blocks; `ErrNotFound` and read failures fall through, matching `cmd_bd.go`'s documented fail-open tradeoff. **The bd-1.0.4 SQL fallback is untouched** — it matches literally, so no preflight, no argv change.

**2. Text classification one line below the exit-code branch.** `isBdNotFound(runErr)` — unanchored `strings.Contains(msg, "not found")` over the whole runner error — returned the authoritative "nobody holds it" verdict for `exec: "bd": executable file not found in $PATH`, `database "fe" not found on Dolt server …`, `404 page not found`, and `beads workspace not found: …`. On main all four were errors. A local `isBdIssueNotFound` now requires exit **1** plus bd's measured issue-scoped wording, so infrastructure failures surface as errors and never as a CAS answer.

**3. CI red — resource-census ledger (`packages-core-1-of-4`).** The drift was **not** the production file's `os/exec` import (`census.go:901` skips non-test files); it was the new **untagged test** file's `exec.Command("sh", "-c", "exit %d")`, used to fabricate a real `*exec.ExitError`. Proven by replacing that one call at the PR head with the import left alone. The ledger's invariant is that untagged subprocess call sites *cannot grow*, so the call site is gone rather than the three lockstepped baselines bumped: `bdExitCode` now matches an `ExitCode() int` interface, with `var _ exitCoder = (*exec.ExitError)(nil)` in the production file pinning the real type.

**4. CI red — integration row FAILED instead of skipping (`packages-core-2-of-4`).** `isUnsupportedOnThisBd` could never fire: on a flagless bd the store latches, `bd sql` is embedded-rejected, and the dolt fallback then *succeeds* with `released=false` for a wisp, so no error ever reached the classifier. The row now gates on a positive capability signal (`bd update --help` advertising both flags, the shape `conditionalWritesCapable` uses) and skips the whole test cleanly; all five error-string skip guards are deleted.

Verified against a shim reproducing the CI-pinned bd (rejects `--if-assignee`/`--if-status`, real bd otherwise) with dolt on PATH — PR head reproduces CI byte-for-byte, this commit skips:

```
# b73b24a165 — identical to CI job 93186748769
--- FAIL: TestBdStoreReleaseIfCurrentAgainstRealBd/an_ephemeral_bead_releases_too
    bdstore_conditional_release_integration_test.go:120: did not release an ephemeral in-progress assignment

# 392a3ce196
--- SKIP: TestBdStoreReleaseIfCurrentAgainstRealBd (4.97s)
ok  github.com/gastownhall/gascity/internal/beads  36.2s   # whole package, -tags integration
```

### Red-before

```
--- FAIL: TestReleaseIfCurrentRefusesAFuzzyIDCollision
    ReleaseIfCurrent released a bead the caller never named: released=true err=<nil>, want an error wrapping ErrIDCollision

--- FAIL: TestDoBdReleaseIfCurrentRefusesAFuzzyIDCollision
    doBdReleaseIfCurrent = 0, want 1 (a substring collision must not be reported as a release); stdout="released\n" stderr=""

--- FAIL: TestBdStoreReleaseIfCurrentAgainstRealBd/a_non-exact_id_refuses_instead_of_releasing_a_different_bead
    ReleaseIfCurrent("tst-68") = released=true err=<nil>, want a refusal wrapping ErrIDCollision

--- FAIL: TestReleaseIfCurrentSurfacesInfraFailuresAsErrors/{a_renamed_or_unmounted_database,bd_fell_off_PATH,a_hosted_gateway_misroute,the_beads_workspace_is_gone}
    an infrastructure failure became a CAS verdict: released=false err=<nil>
```

### Gates

`go build ./...`, `go vet ./...`, `gofmt` clean, `golangci-lint` (2.10.1) **0 issues**, `go test ./internal/beads ./internal/storebinding/...` ok, `go test ./cmd/gc -timeout 40m` ok 815.5s, `go test -tags integration ./internal/beads` ok on both a capable bd (44.0s, all 6 legs incl. the new collision leg) and the CI-pinned flagless bd (36.2s, row skips), `go test -tags integration ./internal/testpolicy/resourcecensus` ok. Pre-commit hook ran and passed.

### One thing deliberately NOT done

`deps.env` pins `BD_VERSION=v1.1.0`, which predates the flags (they landed in beads#5008 / #5364 three weeks after the tag — a local binary self-reporting `bd version 1.1.0` is a dev build from a later commit). So on the bd CI and the Docker images install, the store still latches to the `issues`-only SQL fallback and **the headline wisp fix is not live**, and this row skips rather than guarding anything. Moving `BD_VERSION` past #5364 is the follow-up that makes both real; it is a lockstepped, repo-wide bump and is out of scope here.

Issue #2168, opened 2026-05-15
order-tracking watchdog scope too narrow: cannot recover from multi-order tracking-bead jam without full city restart


Wldc4rd:
## Summary

When a supervisor restart or controller crash leaves dispatched goroutines incomplete, the orphaned \`order-tracking\` beads block all future dispatches of their respective orders via the \`hasOpenWorkStrict\` gate (\`cmd/gc/order_dispatch.go:312-318\`). The recovery watchdog (\`runOrderTrackingSweepWatchdog\` at \`city_runtime.go:868\`) exists to clear this jam but is intentionally narrow — its \`onlyOrders\` filter sweeps **only** the \`order-tracking-sweep\` order's own tracking beads, bootstrapping that order so it can sweep all the others.

In practice on a single-host setup with 19 always-on sessions and slow reconciler cycles, this design fails to recover. After a normal \`gc supervisor stop --wait\` (with the known \`timed out -Xms\` cosmetic error) followed by \`gc supervisor start\`, I observed:

- 20 open \`order:*\` tracking beads in the work queue, including 6 \`order:order-tracking-sweep\` beads from various previous controller runs
- The watchdog's 30s ticker apparently never managed to clear them — they sat open for hours
- All affected orders (\`dispatch-tick\`, \`beads-health\`, \`dolt-health\`, \`order-tracking-sweep\` itself, etc.) stopped firing entirely
- Only \`gc order sweep-tracking --stale-after=30s\` (manual invocation) cleared the jam; the watchdog's identical-sounding mechanism did not

## Reproduction

1. Run a city with at least one cooldown-trigger order long enough that a few tracking beads accumulate
2. \`gc supervisor stop --wait\` (or kill the supervisor while a dispatch goroutine is mid-flight)
3. Wait 5+ minutes after \`gc supervisor start\` returns
4. \`gc events --since 5m --type order.fired\` — no fires for high-cadence orders
5. \`bd list --status=open | grep "order:"\` — multiple open tracking beads, all older than the 2-minute watchdog staleness

The manual \`gc order sweep-tracking\` recovers immediately. The automatic watchdog does not — at least not within 30 minutes of observation.

## Possible causes (not yet verified)

[...]

PR #5127, opened 2026-08-09
fix(orders): create order-tracking beads in the orders-class store

julianknutsen:
## Problem

On a city whose infrastructure classes are served by their own storage binding, **every order-tracking bead the controller writes is born in the WORK ledger and minted under the WORK prefix.**

`internal/coordclass/class.go` defines `ClassOrders` as exactly *"order-dispatch tracking beads (the order-tracking / order-run records that gate repeat order firing)"*. maintainer-city routes `orders -> infra`, so those beads belong in the infra binding. The city's own convergence check reads a `ClassOrders` bead sitting in the source work store as **stranded** — and stranded is **fatal to boot**.

#5106 fixed the wisp roots and drew the boundary explicitly in its own description: *"roots as ClassGraph), while the order-tracking bead `storeFn` opens stays on the order/work side."* That left the dominant case in place.

### Measured on maintainer-city tonight

Running `76123b700a`, with all four recent storage fixes verified as ancestors:

| Signal | Before #5106 | After #5106 (`76123b700a`) |
|---|---|---|
| Stranded infra beads accruing | ~41/h | **~42/h** — essentially unchanged |

Control group, same night:

- With the controller **DOWN**, the stranded count held at **0 for 21 minutes**.
- Within **46 minutes** of the controller running it went **0 -> 32**.

And the composition of the strand:

- Of **28** beads pruned tonight, **1** was a wisp root (#5106's case) and **27** were order-tracking beads.

A representative one: `mc-wisp-4t3`, title `order:dolt-health`, labels `['exec', 'order-run:dolt-health', 'order-tracking']`, `created_by=controller`, non-ephemeral. Work-store prefix, orders class, created after convergence.

Because strands are fatal to boot, this is also why **the city cannot be restarted without a manual prune first.**

## Fix

Route order-tracking / order-run bead creation — and every reader, sweeper and projection of those beads — through the ORDERS-class store, exactly as #5106 did for graph-class wisp roots. The resolver already existed (`resolveOrderStore`); nothing in `internal/coordclass`, the stranded check, or the boot verdict changes.

## Why this had to be one change

This work was deliberately deferred once, and #5106's own follow-up list says why: *"Landing the birth alone would make open tracking beads invisible, the single-flight marker would never clear, and orders would stop firing permanently — strictly worse than the symptom this PR fixes."*

The tracking bead is not a record, it is a **lifecycle**. It is created OPEN as the single-flight marker; its `CreatedAt` is the cooldown clock; its labels carry the event cursor; an outcome is stamped on it; it is closed when the dispatch returns; a watchdog closes it when a controller dies holding it; and the CLI, the doctor and the API all read it. Move the birth without the rest and every one of those addresses a database the bead is not in. The close finds nothing, the marker never clears, the open-work gate suppresses the order — **permanently, for every order.**

So the writer and every reader move together, and each half is pinned by a test that fails when only that half is reverted.

## Reader / writer enumeration

Every site that writes or reads an order-tracking bead, and what it addresses now.

### Writers

| Site | Operation | Before | Now |
|---|---|---|---|
| `cmd/gc/order_dispatch.go` `dispatch()` trigger-env failure | `CreateRun` (open, `trigger-env-failed`) | order's target scope | orders class |
| `cmd/gc/order_dispatch.go` `launchResolvedDispatch` (tick **and** webhook seam) | `CreateRun` (open single-flight marker) | order's target scope | orders class |
| `cmd/gc/order_dispatch.go` `dispatchOne` deferred close | `CloseRuns` | order's target scope | orders class |
| `cmd/gc/order_dispatch.go` `dispatchOne` -> `dispatchExec` front door | `SetOutcome` / `SetCursor` | order's target scope | orders class |
| `cmd/gc/order_dispatch.go` `dispatchWisp` canceled | `SetOutcome(wisp-canceled)` | order's target scope | orders class |
| `cmd/gc/order_dispatch.go` `dispatchWisp` success | `SetOutcome(wisp)` | order's target scope | orders class |
| `cmd/gc/order_dispatch.go` `markTrackingFailure` | `MarkFailed` | order's target scope | orders class |
| `cmd/gc/cmd_order.go` `gc order run` (exec, tracked) | `CreateRun` + `SetCursor` + `SetOutcome` + `CloseRun` | scope store | orders class |
| `cmd/gc/cmd_order.go` `gc order run` (formula) | `CreateRunClosed` (cooldown marker) | scope store | orders class |

All nine route through one of two constructors — `memoryOrderDispatcher.orderFrontDoorFor` and `orderTrackingFrontDoor` — so the bead a dispatch creates and the bead its outcome/close addresses are the same bead by construction, not by each call site remembering.

### Sweepers (close / delete)

| Site | Selection | Before | Now |
|---|---|---|---|
| `cmd/gc/city_runtime.go` boot orphan sweep (`sweepOrphanedOrderTrackingAtBoot`) | `OrphanedOpenRuns` + close | city work store | city work store **+** orders binding |
| `cmd/gc/city_runtime.go` `runOrderTrackingSweepWatchdog` | `StaleOpenRuns` + close | scope stores | scope stores **+** orders binding |
| `cmd/gc/city_runtime.go` `runOrderTrackingRetentionWatchdog` | `ClosedRunsForRetention` + delete | scope stores | scope stores **+** orders binding |
| `cmd/gc/cmd_order.go` `gc order sweep-tracking` (stale close, retention prune, confirm-gate count) | via `orderTrackingSweepStoresForConfigTargets` | scope stores | scope stores **+** orders binding |

The scope stores stay in every one of these lists. A converged city holds pre-cutover tracking beads in the work ledger, and the sweeps are what drain them; trading the work half for the binding half would leave those open forever.

### Readers

| Site | Read | Before | Now |
|---|---|---|---|
| `cmd/gc/order_dispatch.go` open-work gate | `OpenRuns` via the tracking index | gate federation | gate federation **+** orders leg |
| `cmd/gc/order_dispatch.go` cooldown clock | `RecentRunsAll`, `LastRunAcross` | gate federation | gate federation **+** orders leg |
| `cmd/gc/order_dispatch.go` event cursor | `CursorAcross`, `bdCursorAcrossStores` | gate federation | gate federation **+** orders leg |
| `cmd/gc/order_dispatch.go` single-flight | `HasOpenWork` via `hasOpenWorkInStoresStrict` | gate federation | gate federation **+** orders leg |
| `cmd/gc/cmd_order.go` `gc order check` (and the `gc doctor` order-firing check, same resolver) | `cachedOrderStoresResolver` | scope stores | scope stores **+** orders binding |
| `cmd/gc/cmd_order.go` `gc order history` | `cachedOrderHistoryStoresResolver` -> `RecentRuns` | scope stores | scope stores **+** orders binding |
| `cmd/gc/city_runtime.go` `warnIfClosedOrderTrackingBacklogLarge` | closed-count boot advisory | city work store | city work store **+** orders binding, counted together |
| `cmd/gc/doctor_order_tracking_retention.go` | closed-count doctor advisory | city work store | city work store **+** orders binding |
| `internal/api/huma_handlers_orders.go` `orderStoreInfosForState` (order check, order history list) | `workflowStoreInfo` list | rig + city | rig + city **+** orders binding |
| `internal/api/huma_handlers_orders.go` `humaHandleOrderHistoryDetail` (order history detail) | `workflowStores` + `workflowStoreByRef` | graph + city + rigs | graph + city + rigs **+** orders binding |
| `internal/api/orders_feed.go` `buildOrderRunFeedItems` | `ListTracking`, `LatestOpenRun` | `workflowStores` | `workflowStores` **+** orders binding |

The API half needed a new accessor: `State` exposed no `OrdersBeadStore()`, so `internal/api` was **structurally unable** to route orders. It is added here alongside the existing `GraphBeadStore` / `SessionsBeadStore` / `NudgesBeadStore` accessors, implemented on `controllerState` (which now shares one resolver with the controller's own `ordersBeadStore`), on `seededState`, and on the API test fake.

### Deliberately unchanged, with reasons

- **`hasOpenWorkStrict`'s `orders.NewStoreWithGraph(store, store)`** — this is the PER-STORE predicate the federation applies to each member in turn. Both legs are that member by design; the class union comes from the federation, and splitting the legs here would read each database twice.
- **`orderLastRunFn` (`cmd_order.go`)** — has no production caller; it is reachable only from a test.
- **`internal/storebinding/beads_adapter.go`** — constructs `orders.NewStore` over the binding's own store, already class-correct.
- **`internal/coordclass`, the stranded check, the boot verdict** — untouched, as required.

## Tests

New: `cmd/gc/order_dispatch_orders_store_test.go` (16 tests) and `internal/api/orders_class_store_test.go` (8 tests), plus two cases added to `internal/api/handler_convoy_dispatch_test.go` and one to `cmd/gc/city_runtime_test.go`.

Two assertions in `cmd/gc/order_dispatch_graph_store_test.go` are updated rather than kept: they pinned #5106's boundary (*"order tracking stays on the order store"*), which is precisely the boundary this change moves. One of them becomes the stronger statement — on a split city the work store ends a dispatch holding **nothing at all**, because every bead a dispatch writes is infrastructure class.

### Red-before evidence

Each row reverts exactly one production routing decision, keeping the plumbing so the package still compiles.

| Reverted decision | Red-before |
|---|---|
| Tracking-bead birth back on the target store | `order_dispatch_orders_store_test.go:112: order-tracking beads = [], want exactly one` (also `:210` trigger-env failure, `:252` webhook seam) |
| Outcome stamp + close back on the target store (birth still routed) | `order_dispatch_orders_store_test.go:143: tracking bead gcg-1 status = "open", want closed; an open marker in a store the close never addresses suppresses this order on every later tick, forever` |
| Wisp outcome stamp back on the target store | `order_dispatch_orders_store_test.go:171: tracking bead labels = [order-run:reaper order-tracking], want the wisp outcome stamped in the binding` |
| Boot orphan sweep back on the work store alone | `order_dispatch_orders_store_test.go:399: the binding orphan gcg-1 is still open after the boot sweep; its order never fires again and the city cannot be restarted without a manual prune` |
| Tracking sweep watchdog back on the scope stores alone | `order_dispatch_orders_store_test.go:446: stale tracking bead gcg-1 status = "open", want closed; the watchdog cannot see the binding, so the tracking jam (#2168) has no recovery on a split city` |
| `gc order check` / `history` resolvers back on the scope stores alone | `order_dispatch_orders_store_test.go:507: gc order check: last run is zero though gcg-1 is in the binding; the order reads as never run and every cooldown/cron trigger fires again immediately` |
| `gc order sweep-tracking` back on the scope stores alone | `order_dispatch_orders_store_test.go:543: gc order sweep-tracking closed nothing in the binding; the documented recovery for a stuck order is a silent no-op on a split city` |
| `gc order run` records its tracking bead on the scope store | `order_dispatch_orders_store_test.go:602: binding holds 0 order-tracking beads, want the manual run's own; a run recorded off the binding advances a cooldown clock nothing reads` |
| Gate federation loses both infrastructure legs | `order_dispatch_orders_store_test.go:287: gate store list = [0x27f1915e67e0], missing the binding the tracking bead lives in; the marker the dispatch just wrote is invisible and the order re-fires every tick` |
| API order check/history read loses the orders leg | `orders_class_store_test.go:52: order store infos = [{ref:city:test-city ...}], none of them the orders binding; the API reports a split city's orders as never run and calls a just-fired order due` |
| API order feed loses the orders leg | `orders_class_store_test.go:103: order feed items = [], want the binding-resident tracking bead gc-1; the dashboard shows no order activity at all on a split city` |

One case is deliberately reported as **green**, because claiming otherwise would be false: reverting only #5106's GRAPH leg from the gate federation leaves `TestOrderDispatchGateFederatesTheOrdersBindingExactlyOnce` passing. That is exactly what the orders leg is for — on the whole-split shape this build serves, one binding serves both classes, so the orders leg deduplicates against the graph leg and costs nothing, while making the orders class able to reach its own single-flight marker without depending on the graph class's routing. Removing **both** legs is the red row above.

## Single-store cities

Byte-identical, and proven by mutation rather than argument. `resolveOrderStore` returns the exact store value it was handed when the routes relocate nothing, and `relocatedOrdersClassStore` returns `nil` there, so no federation gains an entry and no write changes instance.

Mutating `ordersStoreFor` to return a fresh store when routes are nil:

```
order_dispatch_orders_store_test.go:353: ordersStoreFor returned *beads.MemStore(0x378c7b4b2360), want the identical store value 0x378c7b4b22d0
order_dispatch_orders_store_test.go:318: gate store list = [0x378c7b4b2090 0x378c7b4b2240], want exactly the one store the city has
```

Mutating `relocatedOrdersClassStore` to return a store when the city has no `[storage]`:

```
order_dispatch_orders_store_test.go:564: relocatedOrdersClassStore = *beads.MemStore(0xc730b7ba1b0) for a city with no [storage]; want nil so every federation stays exactly as it was
```

Every append in this change — dispatcher gate, one-shot resolvers, sweep federations, API store lists — is guarded by store identity, so a city whose binding IS one of the stores already in a list reads it once, not twice.

## Review fixes (second pass)

Three reviewers independently found the same MAJOR on the first head, and a verifier reproduced it end to end. Fixed here, with the minors that came with it.

### MAJOR — the order-history list minted a `store_ref` its own detail endpoint rejected with 404

`appendOrdersClassStoreInfo` mints a **new ref namespace**, `orders:<city>`, and `GET /v0/city/{city}/orders/history` returns it verbatim as each entry's `store_ref`. But `humaHandleOrderHistoryDetail` never called `orderStoreInfosForState` — it resolved `?store_ref=` through `workflowStoreByRef`, which switched on exactly `city` / `graph` / `rig`. The ref the list had just handed the client fell through to `(workflowStoreInfo{}, false)` and the handler answered `apierr.ScopeNotFound` — **HTTP 404 for a bead that exists**, on every order run of a split city.

Reproduced on a production-shaped whole-split state (`cityBeadStore` = work store, `ordersBeadStore` == `graphBeadStore` == one binding):

```
LIST   /orders/history?scoped_name=nightly-review      -> entries[0].store_ref = "orders:test-city"
DETAIL /order/history/gc-1?store_ref=orders:test-city  -> 404 scope-not-found
DETAIL /order/history/gc-1 (no store_ref)              -> 200, body store_ref = "graph:test-city"
SINGLE-STORE baseline                                  -> "city:test-city", round-trips 200
```

Two things made it worth fixing properly:

- **It was a regression of this PR.** Pre-PR those rows carried `city:<name>`, which resolves. The `orders:` namespace and the break arrived together.
- **The same bead was labelled two ways by two endpoints.** The no-`store_ref` path only still worked because on `storageSplitWhole` the graph and orders bindings are the *same store value* — the detail path was reaching orders data **through the graph leg**, which is exactly the co-residency this PR added an orders leg to stop depending on.

Three changes:

1. `internal/api/handler_convoy_dispatch.go` — `workflowStoreByRef` gains a `case ordersClassStoreRefPrefix:` arm mirroring the graph arm. It resolves `state.OrdersBeadStore().Store` and returns false when that is nil, is the city store, or the scope ref is not this city, so a city that relocates nothing never gains a second name for a store `city:` already resolves.
2. `internal/api/huma_handlers_orders.go` — the detail handler's default store list is now `appendOrdersClassStoreInfo(workflowStores(s.state), s.state, workflowCityScopeRef(s.state.CityName()))`, the same expression `orders_feed.go` already uses, so the read reaches the orders class itself rather than through the graph binding.
3. `internal/api/convoy_sql.go` — `workflowStorePath` now skips the `orders:` class ref as it already skipped `graph:`. A class ref names a binding, not a scope root, and carries the city scope as its fallback; without this the newly-resolvable ref would answer the **city path** and point the workflow SQL fast path at the wrong database.

### Minors from the same pass

- **`gc order sweep-tracking --include-wisps` swept the orders binding twice.** `appendOrdersSweepStore` puts the binding into the sweep list wrapped in `orderTrackingSweepScopedStore` (it carries the store's label), but the hoisted wisp pass guarded itself with `storeListContains`, a bare identity check that sees the wrapper and concludes the binding is absent. On the whole-split shape the wisp store IS that binding, so `--dry-run --include-wisps` reported twice as much stale work as there is. New `sweepStoreListContains` looks through the wrapper. This is the same count-once invariant `TestOrderWispForceCloseCountsAStoreServingBothClassesOnce` already pins for a raw store — the wrapper was the hole in it.
- **The boot closed-backlog advisory applied its 100-bead threshold per store.** `sweepOrphanedOrderTrackingAtBoot` called `warnIfClosedOrderTrackingBacklogLarge` inside the per-store loop, so a converged split city holding 60 closed tracking beads in the work ledger and 60 in the binding stayed silent — half the sensitivity on exactly the cities the sweep was extended for. It now takes the store list and warns once on the total. The `≥1001` cap becomes "capped if any store hit the list limit", which leaves the single-store output byte-identical.

### Red-before

Each string is from the new tests run against this branch with only the corresponding production change reverted.

| Reverted decision | Red-before |
|---|---|
| No `orders:` arm in `workflowStoreByRef` (list→detail round trip) | `orders_class_store_test.go:237: detail status for the list's own store_ref "orders:test-city" = 404, want 200` |
| Detail handler back on bare `workflowStores` | `orders_class_store_test.go:268: detail status without store_ref = 404, want 200; the read never opened the orders binding` |
| Same, at the invariant level | `orders_class_store_test.go:303: workflowStoreByRef("orders:test-city") = false; an order endpoint published a ref no reader resolves` |
| Same, at the resolver level | `handler_convoy_dispatch_test.go:673: workflowStoreByRef(orders:bright-lights) = false, want true` |
| `workflowStorePath` not skipping the class ref | `handler_convoy_dispatch_test.go:705: workflowStorePath("orders:bright-lights") = "/tmp/.../002", true; want no path` |
| Wisp-sweep dedup back on the bare identity check | `order_dispatch_orders_store_test.go:654: dry-run wispClosed = 4, want 2; the wrapped binding was swept twice and the operator is told twice as much work is stale as there is` |
| Boot advisory back on a per-store threshold | `city_runtime_test.go:7414: warning = "", want the combined count 120; a per-store threshold halves the advisory's sensitivity on exactly the split cities it was extended for` |

Single-store byte-identity is proven by mutation, not argument. Dropping the `ordersStore == state.CityBeadStore()` guard from the new arm:

```
handler_convoy_dispatch_test.go:655: workflowStoreByRef(orders:city) on a single-store city = true, want false
```

Removing the dedup from `appendOrdersClassStoreInfo`:

```
orders_class_store_test.go:78: order store infos = [{ref:city:test-city ...} {ref:orders:test-city ...}], want exactly the city store
orders_class_store_test.go:136: tracking bead gc-1 appears 2 times in the feed, want 1
orders_class_store_test.go:326: list entries = [{BeadID:gc-1 StoreRef:city:test-city} {BeadID:gc-1 StoreRef:orders:test-city}], want exactly 1
```

### One alias is left, deliberately

On the whole-split shape the same binding serves graph and orders, so a client can obtain two names for one store: the history list mints `orders:<city>`, and the feed and the unparameterized detail path report `graph:<city>` because `workflowStores` leads with the graph entry and the append deduplicates on store identity. **Both now resolve, and both resolve to the same store** — asserted by `TestOrderHistoryStoreRefsAllResolveToTheSameStore`. Collapsing the alias would mean relabelling the graph entry inside the order read lists, which changes the feed's emitted `store_ref` for no consumer's benefit. The invariant worth holding is the one now under test: no order endpoint may publish a ref no reader resolves.

## Gates

- `go build ./...` clean
- `go vet ./cmd/gc/... ./internal/api/...` clean
- `gofmt` clean
- `golangci-lint` 2.10.1 on `./cmd/gc/... ./internal/api/...` — **0 issues** (baseline at the previous head was also 0)
- `go test ./internal/api` — pass
- `go test ./cmd/gc -timeout 40m` — pass
- `make check-core-boundary` — OK
- `make check-docs` — OK. No wire type or endpoint changed; the `store_ref` **value space** gains `orders:<city>` on split cities, and that value now round-trips through `workflowStoreByRef` like `graph:` and `rig:` do. `OrdersBeadStore` is an internal read accessor, not an OpenAPI surface.

## Follow-ups, not bundled

- `gc order run`'s FORMULA arm still instantiates its molecule through the unrouted scope store, so a manually-run wisp root is still born in the work ledger. That is a **graph**-class birth — part of #5106's remaining follow-up list, not this one — and it needs its own revert and its own byte-identity proof.
- Retention-prune policy (backup-freshness gate, retain-last floor) is untouched; the retention watchdog's reach into the binding is asserted at the shared store resolver rather than by running the prune.

### DIFF
--- a/cmd/gc/api_state.go
+++ b/cmd/gc/api_state.go
@@ -1574,6 +1574,20 @@ func (cs *controllerState) GraphBeadStore() beads.GraphStore {
 	return beads.GraphStore{Store: resolveGraphStore(cs.storageRoutes, cs.cityBeadStore, cs.cfg, cs.cityPath, cs.eventProv)}
 }
 
+// OrdersBeadStore returns the store backing orders-class beads. At the default
+// backend resolveOrderStore returns cityBeadStore, so this is byte-identical to
+// CityBeadStore; when [beads.classes.orders] is relocated it returns the
+// per-class store, which is the store the order dispatcher creates every
+// tracking bead in. cs.eventProv is the recorder, matching the nudges/sessions
+// wiring. The result is wrapped in the strongly-typed beads.OrdersStore so the
+// orders class is statically visible to callers; the wrapper carries the same
+// underlying store value, so runtime behavior is unchanged.
+func (cs *controllerState) OrdersBeadStore() beads.OrdersStore {
+	cs.mu.RLock()
+	defer cs.mu.RUnlock()
+	return beads.OrdersStore{Store: resolveOrderStore(cs.storageRoutes, cs.cityBeadStore, cs.cfg, cs.cityPath, cs.eventProv)}
+}
+
 // CityBeadsDiagnostic returns the city-level bead store selection diagnostic.
 func (cs *controllerState) CityBeadsDiagnostic() *beads.BeadsDiagnostic {
 	cs.mu.RLock()
--- a/cmd/gc/city_runtime.go
+++ b/cmd/gc/city_runtime.go
@@ -75,8 +75,13 @@ func sweepOrphanedOrderTrackingAtBoot(routes *storageRoutes, cityPath string, cf
 		} else if n > 0 {
 			fmt.Fprintf(stderr, "gc start: closed %d orphaned order-tracking beads\n", n) //nolint:errcheck // best-effort stderr
 		}
-		warnIfClosedOrderTrackingBacklogLarge(store, stderr)
 	}
+	// The advisory is about the city's backlog, so it counts the stores together.
+	// Per-store thresholds would halve the sensitivity on exactly the split cities
+	// the sweep was extended for: a converged city holds half its closed tracking
+	// beads in the work ledger and half in the binding, and neither half alone
+	// reaches the threshold the pair clears.
+	warnIfClosedOrderTrackingBacklogLarge(stores, stderr)
 }
 
 // reloadOrderDrainTimeout bounds how long config reload will wait for
@@ -1632,7 +1637,7 @@ func (cr *CityRuntime) runOrderTrackingRetentionWatchdog(now time.Time) {
 
 const (
 	// orderTrackingRetentionStartupWarnThreshold is the minimum number of closed
-	// order-tracking beads in the city store that triggers a startup advisory.
+	// order-tracking beads a city holds that triggers a startup advisory.
 	// The watchdog prunes automatically; this warning surfaces cities that have
 	// accumulated a visible backlog before the first watchdog cycle completes.
 	orderTrackingRetentionStartupWarnThreshold = 100
@@ -1641,26 +1646,43 @@ const (
 	orderTrackingRetentionStartupListLimit = 1001
 )
 
-// warnIfClosedOrderTrackingBacklogLarge writes a one-line advisory to stderr
-// when the city store holds more than orderTrackingRetentionStartupWarnThreshold
-// closed order-tracking beads. It is best-effort: a nil store or a List error is
-// silently ignored so startup is never blocked.
-func warnIfClosedOrderTrackingBacklogLarge(store beads.Store, stderr io.Writer) {
-	if store == nil {
+// warnIfClosedOrderTrackingBacklogLarge writes one advisory line to stderr when
+// the city's stores together hold more than
+// orderTrackingRetentionStartupWarnThreshold closed order-tracking beads. It is
+// best-effort: a nil store or a List error is skipped so startup is never
+// blocked.
+//
+// The count is the total across the stores, not a per-store test, because the
+// backlog is one city's. A split city holds its pre-cutover half in the work
+// ledger and everything since in the binding; testing each half separately
+// stays silent on a city with twice the backlog the threshold is set at.
+func warnIfClosedOrderTrackingBacklogLarge(stores []beads.Store, stderr io.Writer) {
+	total := 0
+	truncated := false
+	for _, store := range stores {
+		if store == nil {
+			continue
+		}
+		closed, err := beads.HandlesFor(store).Live.List(beads.ListQuery{
+			Status:   "closed",
+			Label:    labelOrderTracking,
+			TierMode: beads.TierBoth,
+			Limit:    orderTrackingRetentionStartupListLimit,
+		})
+		if err != nil {
+			continue
+		}
+		total += len(closed)
+		if len(closed) >= orderTrackingRetentionStartupListLimit {
+			truncated = true
+		}
+	}
+	if total <= orderTrackingRetentionStartupWarnThreshold {
 		return
 	}
-	closed, err := beads.HandlesFor(store).Live.List(beads.ListQuery{
-		Status:   "closed",
-		Label:    labelOrderTracking,
-		TierMode: beads.TierBoth,
-		Limit:    orderTrackingRetentionStartupListLimit,
-	})
-	if err != nil || len(closed) <= orderTrackingRetentionStartupWarnThreshold {
-		return
-	}
-	countStr := fmt.Sprintf("%d", len(closed))
-	if len(closed) >= orderTrackingRetentionStartupListLimit {
-		countStr = "≥1001"
+	countStr := fmt.Sprintf("%d", total)
+	if truncated {
+		countStr = "≥" + countStr
 	}
 	fmt.Fprintf(stderr, "gc start: %s closed order-tracking beads detected — retention watchdog will prune automatically (7d TTL default; configure: [beads.policies.order_tracking].delete_after_close). For immediate cleanup: gc order sweep-tracking\n", countStr) //nolint:errcheck // best-effort stderr
 }
--- a/cmd/gc/city_runtime.go
+++ b/cmd/gc/city_runtime.go
@@ -40,6 +40,45 @@ import (
 // managed dolt.
 var newCityRuntimeOpenSweepStore = openStoreAtForCity
 
+// sweepOrphanedOrderTrackingAtBoot closes order-tracking beads a previous
+// controller instance left open (goroutines killed on restart, or silent Close
+// failures). It runs on startup only, never on config reload, and is
+// best-effort: a store it cannot open is reported and skipped, and the retry
+// with backoff is defense-in-depth against transient store errors immediately
+// after ensureBeadsProvider returns (#753).
+//
+// It sweeps the city work store AND, on a split city, the orders binding those
+// tracking beads are now born in. Both, not one. A converged city holds
+// pre-cutover orphans in the work store and every new one in the binding, and an
+// orphan left open is an order that never fires again — the open bead IS the
+// single-flight marker, so a controller that cannot see it never clears it. A
+// boot sweep that reads only the work store on a split city is therefore not a
+// degraded sweep, it is no sweep at all.
+//
+// The binding handle belongs to routes and is closed with them at runtime
+// shutdown; only the work store opened here is closed here.
+func sweepOrphanedOrderTrackingAtBoot(routes *storageRoutes, cityPath string, cfg *config.City, rec events.Recorder, stderr io.Writer) {
+	workStore, err := newCityRuntimeOpenSweepStore(cityPath, cityPath)
+	if err != nil {
+		fmt.Fprintf(stderr, "gc start: order tracking sweep: %v\n", err) //nolint:errcheck // best-effort stderr
+		return
+	}
+	defer closeBeadStoreHandle(workStore) //nolint:errcheck
+
+	stores := []beads.Store{workStore}
+	if ordersStore := resolveOrderStore(routes, nil, cfg, cityPath, rec); ordersStore != nil && ordersStore != workStore {
+		stores = append(stores, ordersStore)
+	}
+	for _, store := range stores {
+		if n, err := sweepOrphanedOrderTrackingRetryLimit(store, 3, time.Second, orderTrackingSweepCloseBudget); err != nil {
+			fmt.Fprintf(stderr, "gc start: order tracking sweep (closed %d): %v\n", n, err) //nolint:errcheck // best-effort stderr
+		} else if n > 0 {
+			fmt.Fprintf(stderr, "gc start: closed %d orphaned order-tracking beads\n", n) //nolint:errcheck // best-effort stderr
+		}
+		warnIfClosedOrderTrackingBacklogLarge(store, stderr)
+	}
+}
+
 // reloadOrderDrainTimeout bounds how long config reload will wait for
 // the outgoing order dispatcher's in-flight goroutines before replacing
 // it. Reload runs on the tick loop, so a larger budget would stall all
@@ -309,25 +348,7 @@ func newCityRuntime(p CityRuntimeParams) (*CityRuntime, error) {
 		return nil, err
 	}
 
-	// Sweep orphaned order-tracking beads on startup only (not config reload).
-	// A previous controller instance may have left tracking beads open
-	// (goroutines killed on restart, or silent Close failures).
-	// Retry with backoff as defense-in-depth against transient store
-	// errors immediately after ensureBeadsProvider returns (#753).
-	func() {
-		sweepStore, err := newCityRuntimeOpenSweepStore(p.CityPath, p.CityPath)
-		if err != nil {
-			fmt.Fprintf(p.Stderr, "gc start: order tracking sweep: %v\n", err) //nolint:errcheck // best-effort stderr
-			return
-		}
-		defer closeBeadStoreHandle(sweepStore) //nolint:errcheck
-		if n, err := sweepOrphanedOrderTrackingRetryLimit(sweepStore, 3, time.Second, orderTrackingSweepCloseBudget); err != nil {
-			fmt.Fprintf(p.Stderr, "gc start: order tracking sweep (closed %d): %v\n", n, err) //nolint:errcheck // best-effort stderr
-		} else if n > 0 {
-			fmt.Fprintf(p.Stderr, "gc start: closed %d orphaned order-tracking beads\n", n) //nolint:errcheck // best-effort stderr
-		}
-		warnIfClosedOrderTrackingBacklogLarge(sweepStore, p.Stderr)
-	}()
+	sweepOrphanedOrderTrackingAtBoot(routes, p.CityPath, p.Cfg, p.Rec, p.Stderr)
 
 	od, orderSnapshot := buildOrderDispatcherWithSnapshot(routes, p.CityPath, p.Cfg, p.Rec, p.Stderr, "gc start: order scan")
 
@@ -1702,6 +1723,14 @@ func (cr *CityRuntime) orderTrackingSweepStores() ([]beads.Store, []orderTrackin
 		}
 		return store, nil
 	})
+	// The tracking beads this controller writes are born in the orders binding,
+	// so both watchdogs — stale-close and closed-retention — have to sweep it.
+	// Without it a split city's open tracking beads are never recovered and its
+	// closed ones are never pruned: the stale-close jam (#2168) comes back with
+	// no watchdog able to see it. The binding comes from the routes this process
+	// opened at boot, never a second resolution, so nothing here is closed by
+	// closeOpened — the runtime owns that handle for its whole life.
+	stores = appendOrdersSweepStore(stores, cr.relocatedOrdersStore())
 	closeOpened := func() {
 		for _, s := range freshlyOpened {
 			_ = closeBeadStoreHandle(s) //nolint:errcheck // best-effort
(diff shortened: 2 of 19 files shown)


julianknutsen:
## Council MAJOR fixed on `f0abf59e5b` — the list minted a `store_ref` the detail endpoint rejected

Three reviewers converged on the same defect and a verifier reproduced it end to end. Fixed here, not merged.

**The defect.** `appendOrdersClassStoreInfo` mints a new ref namespace, `orders:<city>`, and `GET /v0/city/{city}/orders/history` returns it verbatim as each entry's `store_ref`. `humaHandleOrderHistoryDetail` never called `orderStoreInfosForState` — it resolved `?store_ref=` through `workflowStoreByRef`, which switched on exactly `city` / `graph` / `rig`. The ref the list had just published fell through to `(workflowStoreInfo{}, false)` → `apierr.ScopeNotFound` → **HTTP 404 for a bead that exists**, on every order run of a split city. A regression: pre-PR those rows carried `city:<name>`, which resolves.

Worth noting the verifier's correction, which strengthens it: `test/integration/gc_live_contract_test.go:973-976` does **not** omit `store_ref` — it conditionally appends `?store_ref=` taken from the list entry. The repo's own contract test already encodes the list→detail handoff this broke; it stayed green only because its fixture bead is created via `POST /beads` with `rig`, so the list hands back a resolvable `rig:<rig>`. A dispatcher-created tracking bead on a whole-split city is the case that breaks.

**The fix** (three files, all `internal/api`):

1. `workflowStoreByRef` gains a `case ordersClassStoreRefPrefix:` arm mirroring the graph arm — resolves `state.OrdersBeadStore().Store`, false when nil / equal to the city store / wrong scope ref, so single-store cities are untouched.
2. `humaHandleOrderHistoryDetail` builds its default store list as `appendOrdersClassStoreInfo(workflowStores(...))` — the expression `orders_feed.go` already uses — so the detail read reaches the orders class itself rather than through the graph binding it happens to share a store with.
3. `workflowStorePath` skips the `orders:` class ref as it already skipped `graph:`. Found while wiring (1): a class ref names a *binding*, not a scope root, but carries the city scope as its fallback — so once the ref resolves, `scope_kind=orders&scope_ref=<city>` would have answered the **city path** and pointed the workflow SQL fast path at the wrong database.

**Minors taken** (both from the same council list):

- `gc order sweep-tracking --include-wisps` swept the orders binding twice. `appendOrdersSweepStore` wraps the binding in `orderTrackingSweepScopedStore`, and the hoisted wisp pass guarded itself with `storeListContains` — a bare identity check that sees the wrapper. New `sweepStoreListContains` looks through it. Same count-once invariant `TestOrderWispForceCloseCountsAStoreServingBothClassesOnce` pins for a raw store; the wrapper was the hole.
- The boot closed-backlog advisory applied its 100-bead threshold per store. 60 in the work ledger + 60 in the binding stayed silent. It now takes the store list and warns once on the total; the `≥1001` cap becomes "capped if any store hit the list limit", leaving single-store output byte-identical.

**Red-before** (each with only its own production change reverted, on this exact tree):

```
orders_class_store_test.go:237: detail status for the list's own store_ref "orders:test-city" = 404, want 200
orders_class_store_test.go:268: detail status without store_ref = 404, want 200; the read never opened the orders binding
orders_class_store_test.go:303: workflowStoreByRef("orders:test-city") = false; an order endpoint published a ref no reader resolves
handler_convoy_dispatch_test.go:673: workflowStoreByRef(orders:bright-lights) = false, want true
handler_convoy_dispatch_test.go:705: workflowStorePath("orders:bright-lights") = "/tmp/.../002", true; want no path
order_dispatch_orders_store_test.go:654: dry-run wispClosed = 4, want 2; the wrapped binding was swept twice and the operator is told twice as much work is stale as there is
city_runtime_test.go:7414: warning = "", want the combined count 120; a per-store threshold halves the advisory's sensitivity on exactly the split cities it was extended for
```

Single-store byte-identity proven by mutation, not argument — dropping the `ordersStore == state.CityBeadStore()` guard, and removing the dedup from `appendOrdersClassStoreInfo`:

```
handler_convoy_dispatch_test.go:655: workflowStoreByRef(orders:city) on a single-store city = true, want false
orders_class_store_test.go:78:  order store infos = [{ref:city:test-city ...} {ref:orders:test-city ...}], want exactly the city store
orders_class_store_test.go:136: tracking bead gc-1 appears 2 times in the feed, want 1
orders_class_store_test.go:326: list entries = [{BeadID:gc-1 StoreRef:city:test-city} {BeadID:gc-1 StoreRef:orders:test-city}], want exactly 1
```

**One alias left, deliberately.** On the whole-split shape one binding serves graph and orders, so a client can obtain two names for one store: the list mints `orders:<city>`, the feed and the unparameterized detail report `graph:<city>` (`workflowStores` leads with graph; the append dedups on store identity). Both now resolve, and **both resolve to the same store** — asserted by `TestOrderHistoryStoreRefsAllResolveToTheSameStore`. Collapsing the alias means relabelling the graph entry inside the order read lists, which changes the feed's emitted `store_ref` with no consumer benefit. The invariant now under test is the one that matters: no order endpoint may publish a ref no reader resolves.

**PR body corrected.** The reader table claimed `orderStoreInfosForState` covered "order check, history, history detail" — the detail handler never called it. It is now two rows, one per handler, with the detail row naming `workflowStores` + `workflowStoreByRef` as what it actually uses.

**Gates:** `go build ./...` clean · `go vet ./cmd/gc/... ./internal/api/...` clean · `gofmt` clean · `golangci-lint` 2.10.1 on `./cmd/gc/... ./internal/api/...` **0 issues** (baseline at the previous head also 0) · `go test ./internal/api` pass (79s) · `go test ./cmd/gc -timeout 40m` pass (784s) · `go vet -tags integration ./test/integration/` clean · `check-core-boundary` OK · `check-docs` OK.

PR #5130, opened 2026-08-09
test(beads): add the strict, prefix-disjoint split-store test kit

julianknutsen:
## Problem

Every split-store test in this repo runs on `beads.MemStore`, and MemStore
answers a residence-invariant violation the same way no matter which backend the
test claims to model — it accepts, and carries no damage:

| operation | MemStore | bd/Dolt (work store) | SQLite (coordination class) |
|---|---|---|---|
| `DepAdd` to an endpoint missing from this store | appends the edge, no resolution | hard-fails `no issue found matching <id>` | accepts: `deps` has no foreign key, `DepAdd` is a plain `INSERT` |
| `Create` with an explicit id outside the store's prefix | mints over the id | hard-fails `prefix mismatch ... use --force to override` | accepts the id verbatim: `normalizeCreate` has no prefix check |

So a cross-store dependency, or a graph-prefixed bead "created" inside a work
store, is a **green test and a production wedge**. Worse than the failure is the
silence: MemStore renames the bead, so the test's own id is a lie — a later
`Get` for the id it asked for returns `ErrNotFound`, and the test never notices
because it kept the returned id.

That leniency is the stated root cause behind the 18 audited split-store
landmines: every one of them passed on a lenient double while production hard-
failed. Nothing in the follow-on split-store work can be **proven** fixed until
a test double can fail the way production fails.

## What the kit enforces

`internal/beads/splittest` wraps a LEAF store — not a policy wrapper, since
`beadPolicyStore` delegates `DepAdd` and Create's id handling straight down, so
a strict leaf keeps the checks live on every path through the policy stack.

- **`DepAdd` resolves BOTH endpoints** in this store, and **`Create` checks an
  explicit id against the declared prefix** (`Tx.Create` and `CreateWithStorage`
  too, so a transaction is not a side door). What an unresolvable endpoint or a
  foreign-prefix id *means* is chosen per leaf — see "Two backends, two answers"
  below. `BdStore.DepAdd`'s parent-child short-circuit is preserved exactly — on
  a split store the parent may legitimately live elsewhere, and bd never sees the
  call. A rejection deliberately does **not** wrap `beads.ErrNotFound`: bd's real
  failure is a subprocess stderr string callers can only classify textually, so a
  typed error would let tests pass on an `errors.Is` check production can never
  satisfy.
- **The leaf's own fidelity is always checked**, under both semantics: a leaf
  that returns an id other than the one pinned, or mints outside the declared
  prefix, fails at the first create. Neither backend does either, so these are
  checks on the double, not on production.

### Two backends, two answers: pick one per leaf

A single strictness setting would model one backend and lie about the other. A
work store runs on bd/Dolt; a relocated coordination class is served by
`beads.OpenSQLiteStore` (`internal/storebinding/sqlite/beads_engine.go`).
`Semantics` selects which one a leaf models, and the package doc carries a
per-rule table naming the backend each rule models, with source citations.

- **`BdSemantics`** (`NewWorkStore`) rejects at the call site, in bd's own
  wording: `validation.ValidateIDPrefixAllowed`'s `prefix mismatch: database uses
  'gc-' but ID '...' doesn't match (use --force to override)`, and
  `cmd/bd/dep.go`'s `resolving issue ID <id>: no issue found matching "<id>"`.
- **`SQLiteSemantics`** (`NewClassStore`) **accepts**, the way SQLite accepts —
  the row or edge lands, so the fixture sees production's outcome (a dangling
  edge drops its dependent out of `Ready` rather than erroring) — and records a
  `ResidenceViolation`. The constructors fail the test at cleanup for any
  violation the fixture did not claim with `splittest.TakeResidenceViolations`,
  so accepting is never *unnoticed*; a fixture asserting the production
  corruption claims them, every other fixture goes red.

One rule is attributed to the domain layer rather than to a backend: bd passes a
cross-**prefix** dep target through as an external ref (`cmd/bd/dep.go`), so
neither backend rejects a cross-store edge. `BdSemantics` still rejects it — for
the co-residence invariant `convoy.TrackItemIn` enforces with
`ErrMemberNotCoResident` — and says so, instead of borrowing bd's wording for a
failure bd does not produce.
- **Tier-transparent.** Production molecules materialize as ephemeral wisps
  carrying pinned `<prefix>-wisp-<suffix>` ids. A double that clobbers pinned
  ids cannot express the wisp tier at all, so the kit's leaves round-trip them —
  while an issues-tier read still does not see them (the wrapper never expands
  the caller's `TierMode`).
- **Capability-preserving.** An interface-embedding wrapper silently strips
  every optional capability production discovers by type assertion, so the
  wrapper forwards `Handles` (with a strict `Writer`), `IDPrefix`, graph-apply
  and conditional writes via their handle providers, the conditional-writes
  resolve target, `Counter`, `ConditionalAssignmentReleaser`, `BatchDeleter`,
  `ForeignIDCreator`, `DepListBatch`, `CloseStore`, `AtomicTx`, `Backing`, and
  `WaitForParentProjection`. `StorageCreateStore` is claimed only when the leaf
  has it, so the flag-based storage fallback keeps firing for MemStore leaves.

### Keyed by class, not by a literal

Stores are keyed by coordination class rather than a hardcoded prefix:

```go
work, graph := splittest.NewSplitStores(t)          // gc-<n> / gcg-<n>
orders      := splittest.NewClassStore(t, config.BeadClassOrders)
rig         := splittest.NewWorkStore(t, "ra")      // a rig/HQ EffectivePrefix
```

`NewClassStore` reads `config.ReservedClassPrefix`, the single source of truth,
so a class that gains, loses, or changes its reserved prefix moves the kit with
it instead of leaving a stale literal behind. `config.BeadClassWork` is rejected
by name — work ids are indistinguishable from each other, so there is no
namespace for the residence invariant to be about; that is what `NewWorkStore`
is for, and it rejects a reserved class prefix in turn.

## Evidence: strict vs. lenient, same operations

This is test-only, so the "red before" is the **contrast**. Two tests run the
identical operations against a plain MemStore pair and against the strict pair,
and each requires the opposite outcome
(`TestLenientMemStoreDoublesHideProductionFailures` /
`TestStrictStoreCatchesWhatLenientDoublesLetThrough`):

```
LENIENT  cross-store DepAdd(gc-1 -> gcg-1)  => <nil>          (and no consequence either)
LENIENT  work.Create(id=gcg-wisp-y785sz)    => err=<nil>, stored as "gc-2"

BD       cross-store DepAdd(gc-1 -> gcg-1)  => adding dep gc-1→gcg-1: gcg-1 belongs to another
                                               store's id namespace, not "gc": a dep row cannot
                                               reference an id its own store cannot resolve
                                               (convoy.TrackItemIn rejects the same shape with
                                               ErrMemberNotCoResident)
BD       work.DepAdd(gc-1 -> gc-absent)     => adding dep gc-1→gc-absent: resolving issue ID
                                               gc-absent: no issue found matching "gc-absent"
BD       work.Create(id=gcg-wisp-y785sz)    => creating bead "gcg-wisp-y785sz": prefix mismatch:
                                               database uses "gc-" but ID "gcg-wisp-y785sz"
                                               doesn't match (use --force to override) — bd's own
                                               rejection of a mismatched --id; use
                                               CreateWithForeignID for the forced create

SQLITE   graph.DepAdd(gcg-1 -> gc-1)        => <nil>, edge recorded, gcg-1 no longer Ready,
                                               1 residence violation recorded
SQLITE   graph.Create(id=gc-123)            => id="gc-123" verbatim, unroutable by prefix,
                                               1 residence violation recorded
```

The lenient side also asserts the consequences, not just the missing error: the
cross-store edge **is recorded** pointing at a bead the store cannot `Get`, and
the pinned id **is never stored**. That is the shape of a landmine shipping
green.

## Two supporting changes

**`beads.MemStore.HonorExplicitIDs`** — the companion of the existing exported
`IDPrefix` field. `IDPrefix` decides what a store *mints*; `HonorExplicitIDs`
decides whether it also *accepts*. Off by default, so every existing caller is
byte-identical. It carries `SQLiteStore.Create`'s contract in full: a duplicate
pinned id is a hard error, and a pinned `<prefix>-<n>` consumes that suffix so a
later mint cannot re-issue it. `MemStore.Create` also advances past any id
already present when minting, mirroring `mintUniqueIDTx` — necessary on the
default path too, since `NewMemStoreFrom` seeds a sequence independently of the
rows it seeds. `TestMemStoreHonorExplicitIDsMatchesSQLiteStore` pins the parity
claim against a real `SQLiteStore` rather than against MemStore itself. Without
this knob no MemStore can model a per-class database, and the wisp tier is
inexpressible.

**An import-boundary guard.** The kit's files are not `_test.go` files — other
packages import them — yet they import `testing` and take a `*testing.T` (the
`httptest` shape). That costs the compiler's natural boundary: nothing would
stop a production file importing this package, linking `testing` into `gc`, and
building a store whose failure mode is `t.Fatalf` — and whose whole job is to be
*stricter* than production. `TestSplittestIsNotImportedByProductionCode` parses
every non-test file in the module and names any offender. It reuses the
mechanism `internal/rollout/boundary_test.go` uses for its import boundary and
`internal/testenv/lint_test.go` uses for a whole-tree scan; it is deliberately
**not** a new entry in `scripts/check-core-boundary.sh` (whose checks are all
about commercial coupling in the open-core boundary) and not a resourcecensus
resource (which counts syntax in `_test.go` files and by construction cannot see
a non-test file's import). Verified by planting an importing package: the guard
fails and names the file.

## Tests

All new. The kit's own tests cover the four acceptance behaviors plus the ways
the guards could be wrong:

- on a bd-semantics leaf: cross-store `DepAdd` rejected and **not recorded** in
  the leaf, a same-prefix missing endpoint rejected in bd's wording, and explicit
  foreign-prefix `Create` rejected — direct, via `Tx`, and via the
  `HandlesFor(...).Writer` production write paths discover
- on a SQLite-semantics leaf: both writes **accepted**, the dangling edge drops
  its dependent out of `Ready`, the foreign row is unroutable by prefix, and each
  is recorded as a `ResidenceViolation` that fails the test unless claimed
- a bd-semantics leaf records nothing to claim, and an unwrapped leaf reports
  nothing either
- `Strict` infers and normalizes a declaring leaf's prefix; a leaf with no
  declared namespace, an empty prefix, or an undeclared backend is a hard failure
  pointing at `StrictWithPrefix`
- wisp-tier ids round-trip, keep `Ephemeral`, carry deps in-store, and are
  refused by the work store both as a dep endpoint and as a create
- same-store deps still succeed, in-prefix explicit ids still succeed
  (case-insensitively), and the parent-child short-circuit still fires — a guard
  that rejects everything is not a guard
- a clobbering leaf and a wrong-prefix leaf are both rejected at first create
- every reserved class builds a store minting under its own prefix, and the
  class prefixes are mutually disjoint and disjoint from the work prefix
- capability forwarding, including that `StorageCreateStore` is **not** falsely
  claimed
- `MemStore.HonorExplicitIDs` is opt-in, rejects duplicates, agrees with a real
  `SQLiteStore` on pin-then-mint for `<prefix>-1`, `<prefix>-500` and a
  wisp-shaped pin and on the duplicate-id message, never re-issues an id already
  present, and passes the full `beadstest` store conformance suite with the knob
  on

## Gates

| Gate | Result |
|---|---|
| `go build ./...` | clean |
| `go vet ./...` | clean |
| `gofmt -l` | clean |
| `golangci-lint run ./...` (2.10.1) | 0 issues |
| `go test ./internal/beads/...` | pass |
| `go test ./internal/...` (full sweep) | pass |
| `go test ./cmd/gc -timeout 25m` | pass (773s) |
| `scripts/check-core-boundary.sh` | OK |
| `scripts/check-eventexport-isolation.sh` | OK |
| `scripts/check-routed-test-rows.sh` | OK |
| `go test ./internal/testpolicy/...` (resource census) | pass |
| `go test ./internal/testenv/...` (testenv lint, generated import file present) | pass |


julianknutsen:
## Council review fixes — three defects closed (`4f277c57e0`)

Six majors from the council clustered into three defects. Every one was an
instance of the class this kit exists to stop — a double letting a test pass on
behavior production does not have — so all three are fixed inside the kit rather
than documented around.

### 1. `MemStore` could mint an id that already existed

The honor branch never lifted the sequence floor past a pinned id's numeric
suffix, and the mint branch formatted `<prefix>-<seq>` with no existence check.
`MemStore` is slice-backed and `indexOfLocked` returns the FIRST match, so the
rows aliased: `Get`/`Update`/`Delete`/`DepAdd` on the minted id all hit the
PINNED bead.

`SQLiteStore` — the store the field's doc claims to match — has **two**
independent guards `MemStore` had neither of: `normalizeCreate` calls
`ensureSequenceAtLeast(numericIDSuffix(id))` on every pinned create, and every
auto-mint goes through `mintUniqueIDTx`, which re-checks `idExistsTx` and
reseeds. Both are mirrored now.

The mint guard is **not** belt-and-braces. `NewMemStoreFrom` seeds a sequence
independently of the rows it seeds (callers pass `len(corpus)`), so it reproduces
with the knob **OFF**:

```
--- FAIL: TestMemStoreMintNeverReissuesAnExistingID
    minted id = "gc-3", which is already taken by the seeded bead
--- FAIL: TestMemStoreHonorExplicitIDsMatchesSQLiteStore/adjacent_numeric_pin
    after pinning "gc-1", MemStore minted "gc-1" but SQLiteStore minted "gc-2"
--- FAIL: TestMemStoreHonorExplicitIDsMatchesSQLiteStore/distant_numeric_pin
    after pinning "gc-500", MemStore minted "gc-1" but SQLiteStore minted "gc-501"
```

The differential test opens a **real** `SQLiteStore` and requires identical
answers on pin-then-mint and on the duplicate-id message, so the doc's parity
claim is pinned against the store it names rather than against MemStore itself.
The wisp subtest and the duplicate-message subtest were green before the fix —
`numericIDSuffix("gc-wisp-abc")` is 0 and SQLite behaves the same, so
"pinning must not consume the sequence" is preserved exactly.

### 2. The kit modeled bd's strictness for stores whose backend is SQLite

A relocated coordination class is served by `beads.OpenSQLiteStore`
(`internal/storebinding/sqlite/beads_engine.go`), not by bd, and SQLite rejects
**neither** of the writes the kit was rejecting in bd's name. Measured against a
real `SQLiteStore` in a temp dir:

| operation | SQLiteStore | bd (pinned version, from source) |
|---|---|---|
| `Create{ID:"gc-123"}` in a `gcg` store | `id="gc-123"`, `err=<nil>` | rejects: `prefix mismatch: database uses 'gcg-' but ID 'gc-123' doesn't match (use --force to override)` (`internal/validation/bead.go:139`, called at `cmd/bd/create.go:499` with `--force`) |
| `DepAdd` to a missing **same-prefix** id | `<nil>` | rejects: `resolving issue ID <id>: no issue found matching "<id>"` (`cmd/bd/dep.go` `resolveIDWithRouting` / `resolveIDForMutation`) |
| `DepAdd` to a **foreign-prefix** id | `<nil>` | **accepts** — `cmd/bd/dep.go:367-375` sets `toID = dependsOnArg` when the prefixes differ (cross-prefix external ref) |
| `DepAdd` from a missing source | `<nil>` | rejects (no cross-prefix escape for the source) |
| consequence of the dangling edge | `Ready()` silently drops the dependent (`[gcg-1]` → `[]`) | — |
| leaf clobbers a pinned id / mints outside its prefix | n/a — honors and mints correctly | n/a |

Red-before, on the class leaf:

```
--- FAIL: TestRedBeforeDefect2
    class store rejected a cross-store dep SQLite accepts: adding dep gcg-1→gc-1:
      resolving issue ID gcg-1: no issue found (endpoint not in this store — cross-store dependency?)
    class store rejected a foreign-prefix create SQLite accepts: creating bead "gc-123":
      explicit id prefix does not match store id prefix "gcg" (bd rejects a mismatched --id ...)
    work-store create error ... does not use bd's own wording (validation.ValidateIDPrefixAllowed)
    work-store dep error ... does not use bd's own wording (cmd/bd/dep.go)
```

Fix: semantics are selectable per leaf. `NewWorkStore` → `BdSemantics` (reject at
the call site, in bd's own strings). `NewClassStore` → `SQLiteSemantics`
(**accept**, so the fixture sees production's outcome — the dangling edge really
does drop its dependent out of `Ready`, and the foreign row really is unroutable
by prefix) and record a `ResidenceViolation`. The constructors fail the test at
cleanup for any violation the fixture did not claim with
`TakeResidenceViolations`, so acceptance is faithful but never unnoticed. The
package doc carries the per-rule table above with its source citations.

The cross-store `DepAdd` rejection is now attributed to the **domain** layer
(`convoy.TrackItemIn`'s `ErrMemberNotCoResident`), since bd itself passes a
cross-prefix target through — that also closes the minor "strict `DepAdd` rejects
cross-store edges both backends accept".

### 3. `Strict()` silently yielded a non-strict store

Red-before, on the only leaf the kit builds:

```
--- FAIL: TestRedBeforeDefect3
    Strict(leaf).IDPrefix() = "", want "gcg"
    the inferred-prefix store accepted a work-prefixed id; the create guard is inert
    PrefixOwner could not route an in-prefix id to the store; the inferred prefix did not reach IDPrefix
```

`beads.MemStore` can never satisfy `storeref.HasIDPrefix` — `IDPrefix` is a
field, and Go forbids a method of the same name. `Strict` now `t.Fatalf`s and
points at `StrictWithPrefix`, and `newStrict` rejects an empty prefix and an
undeclared backend on every path. Both rules are covered through the
`inferredPrefix` / `newStrict` helpers, the same split `classPrefix` /
`workPrefix` already use. Coverage: `Strict` 0.0% → 80.0%, `inferredPrefix`
100%, `newStrict` 92.3%.

### Minors

- **Taken:** the conditional-writes resolve-target subtest asserted only "not the
  wrapper", which a `nil` return satisfies while collapsing the resolve to legacy
  — it now asserts the target **is** the `*beads.MemStore` leaf.
- **Taken:** cross-store `DepAdd` attribution, covered above.

### Gates

`go build ./...` · `go vet ./...` · `gofmt -l` (clean) ·
`golangci-lint run ./...` 2.10.1 (**0 issues**) · `go test ./internal/...` ·
`go test ./cmd/gc -timeout 25m` (773s) · `scripts/check-core-boundary.sh` ·
`check-eventexport-isolation.sh` · `check-routed-test-rows.sh` ·
`./internal/testpolicy/...` · `./internal/testenv/...` — all pass.

`HonorExplicitIDs` remains default-OFF, and `TestMemStoreHonorExplicitIDsIsOptIn`
still pins that an unset store mints over a pinned id unchanged.

PR #5131, opened 2026-08-09
feat(beads): derive the postgres endpoint from bd's postgres_dsn

julianknutsen:
## This is a contract widening, not a bug fix

`internal/beads/contract` today **rejects** a postgres scope that does not
carry all four discrete fields:

```go
// files.go, LoadMetadataState
if state.PostgresHost == "" || state.PostgresPort == "" || state.PostgresUser == "" || state.PostgresDatabase == "" {
    return ..., "backend=postgres requires postgres_host, postgres_port, postgres_user, postgres_database (all four must be non-empty)"
}
```

and preflight knows about `postgres_dsn` only well enough to nag about it
(`preflight_checker.go`: `"postgres_dsn present; Gas City expects split fields"`).

But `bd init --backend=postgres --pg-url --pg-schema` persists a
password-free `postgres_dsn` plus a per-workspace `postgres_schema` — not the
discrete fields. So a scope bd just initialised is a shape Gas City refuses to
load.

<details>
<summary><b>Primary-source evidence for that premise</b> (a reviewer challenged it — it holds, for the add-on build)</summary>

The writer is in the **Postgres add-on** distribution, not OSS bd. Build tag
`//go:build enterprise`, `gascity/bd-enterprise` branch `enterprise`
@ `754dbe10af819343ab5aaa6ec040c6c177c3b87f`,
`cmd/bd/backend_init_distribution_enterprise.go`, `runInitPostgres`:

```go
persistDSN, err := pgdialect.RedactPassword(dsn)   // L135, fail-closed
...
cfg.Backend = configfile.BackendPostgres
cfg.PostgresDSN = persistDSN                       // L186
cfg.PostgresSchema = in.pgSchema                   // L187
if err := cfg.Save(in.beadsDir); err != nil { ... } // L189
```

The read side is `internal/storage/postgres/fromconfig.go` L41-43:

```go
dsn = cfg.GetPostgresDSN()
if dsn == "" {
    return "", "", fmt.Errorf("postgres: no DSN (set postgres_dsn in metadata.json, or BEADS_POSTGRES_URL)")
}
```

`BEADS_PG_PASSWORD` is real, not invented: `internal/storage/postgres/credential.go`
L41-42 is the credential ladder (`BEADS_PG_PASSWORD_COMMAND`, then
`BEADS_PG_PASSWORD`). Shipped binaries carrying this writer are installed on
the fleet at `/opt/beads/releases/mc-postgres-p0-*`, `mc-postgres-field-p0-*`,
`mc-field-integrated-*` (`strings` confirms
`a password may be included for init but is never persisted`).

**Where the reviewer was right:** OSS bd contradicts a naive reading. At the
`deps.env` pins — `BD_VERSION=v1.1.0`, `BD_PREV_VERSION=v1.0.4` (neither has
the field at all) and `BD_CURRENT_REF=bf97b73749ac` — `PostgresDSN` appears
exactly once, in `internal/configfile/configfile.go:42`, under "Deprecated
backend fields … retained only to round-trip metadata written by the
short-lived PostgreSQL/MySQL implementations", and `bd init --backend=postgres`
returns `storage backend "postgres" is no longer supported`. The old comments
cited that file as their authority, which read as self-contradictory. They now
cite the add-on. Both halves matter: the add-on **writes** the shape, and OSS
bd **round-trips** the keys without clearing them — which is exactly why gc
must tolerate them rather than force a hand-conversion.

</details>

**This PR makes that shape load.** It turns a currently-rejected metadata
shape into an accepted one. That is the whole point, and it deserves review
attention as a contract change rather than a fix.

What it does *not* relax: `LoadMetadataState` still guarantees the property
every consumer already depends on — a postgres state resolves to a
**complete, port-checked** `host/port/user/database` tuple. The DSN is
simply allowed to be where that tuple comes from. A DSN that resolves to a
partial tuple (no database, say) is rejected by name, so the widening cannot
turn "loud rejection" into "loads fine, then misbehaves".

## Design

`MetadataState.PostgresEndpoint()` is the single derivation point:

- all four discrete fields present -> they win outright, DSN ignored
  (and not even parsed)
- otherwise each missing field is filled from `postgres_dsn`
- neither available -> error

Ported from `feat/split-store-conformance`
(`internal/beads/contract/postgres_dsn.go`).

## Known limitation, preserved and made loud

libpq's keyword/value form (`host=... port=... user=...`) is **not
supported**. bd accepts it at init time; gc cannot derive an endpoint from it
deterministically. Rather than fall through to a partial or silently wrong
derivation, it fails with a named sentinel:

```go
var ErrPostgresDSNUnsupportedForm = errors.New("postgres_dsn must be a postgres:// (or postgresql://) URL")
```

The keyword/value form is detected *explicitly* (every whitespace-separated
token is `key=value`, no `://`) so the message says
`libpq keyword/value form is not supported` instead of the useless
`got scheme ""` you'd get from bare URL parsing. `errors.Is` works at every
layer: parse -> `PostgresEndpoint()` -> `LoadMetadataState` -> the
`BEADS_POSTGRES_*` projection.

## The preflight warning: it becomes wrong, so it's fixed here

`checkContractShape` currently returns
**WARN `"postgres_dsn present; Gas City expects split fields"`** for a
DSN-only postgres scope.

Once this PR lands, that sentence is false, and acting on it is harmful: it
tells an operator to hand-convert away from bd's own canonical shape, which
`bd init` will just write back. A warning that tells operators to fix a shape
we now accept is worse than no warning. So:

- DSN resolves to a complete endpoint -> **PASS** `"Metadata uses postgres_dsn shape"`
- it does not -> **FAIL** `"postgres_dsn does not resolve to a complete postgres endpoint"`

Both branches call `validatePostgresEndpoint`, the same gate `LoadMetadataState`
uses, so a preflight PASS means the loader will accept the scope. (Gating on
`ParsePostgresDSNEndpoint` alone — as the first revision did — PASSed a DSN
supplying no database or a port outside 1..65535, which the loader then
hard-rejected. A preflight that says PASS where load fails is a lie.)

Deliberately left alone:

- `checkMetadataBackend`'s WARN `"Metadata backend is postgres (postgres_dsn form)"`
  is descriptive, not prescriptive, and is about native-store eligibility
  (native supports dolt only) — this change doesn't make it false.
- `hasDSN && hasSplit` -> FAIL stays. A DSN pointing at one host beside
  discrete fields pointing at another is genuinely ambiguous on disk and
  worth flagging, even though the derivation resolves it deterministically.

The FAIL summary deliberately does **not** interpolate the parse error:
only `PostgresDSNRedacted` is sanitized by `Redacted()`, so echoing an error
that quotes the DSN would leak a hand-written password past redaction.
There's a regression test for that.

## Consumers had to move too

Widening the load without this would be the silent-failure trap: a DSN-only
scope would newly load fine and then project **empty** `BEADS_POSTGRES_HOST`
/ `PORT` / `USER` / `DATABASE` to every bd subprocess. So the three sites
that read the discrete fields directly now go through the derived endpoint:

- `cmd/gc/bd_env.go` — the `BEADS_POSTGRES_*` projection and the
  `pg.credential_resolved` payload (`emitPostgresCredentialResolved` now
  takes the derived `PostgresEndpoint` instead of the raw state)
- `cmd/gc/bd_env.go` — the external-PG "no managed recovery" error hint
- `internal/doctor/checks/postgres_auth.go` — scope discovery

`postgres_dsn` / `postgres_schema` participate in cross-backend **scrubbing**
but deliberately **not** in mixed-backend rejection — see "Review round 2"
below for why making them fatal bricked dolt scopes that main loads fine.

## Tests

New `internal/beads/contract/postgres_dsn_test.go` + 4 fixtures:

- URL-form derivation: non-default port, missing port defaults to 5432,
  query params ignored, IPv6, percent-encoding, userinfo password ignored
- discrete fields win when complete — including when the DSN points
  elsewhere, and when the DSN is unparseable (it's never consulted)
- incomplete discrete set falls back to the DSN per field
- libpq keyword/value fails with `ErrPostgresDSNUnsupportedForm`, names the
  form, and returns the **zero** endpoint (no partial derivation) — asserted
  at parse, state, and load layers
- malformed DSN, non-postgres scheme, hostless URL, out-of-range
  DSN-derived port all rejected
- `LoadMetadataState` accepts bd's shape, tolerates a dolt scope carrying
  bd's residual `postgres_dsn`/`postgres_schema` (and self-heals it), still
  rejects dolt + a discrete `postgres_host`, rejects an incomplete derivation
  naming the missing field
- `EnsureCanonicalMetadata` persists and scrubs the two keys, and a
  canonical DSN scope round-trips byte-identical

New in `cmd/gc/bd_env_test.go`: the projection derives from a DSN-only state,
and refuses an underivable DSN without writing partial env.

Updated: the two E4 message expectations in `files_test.go` and the
contract-shape state in `TestPreflightRedactsPostgresDSN`.

Every one of these was verified red by reverting the corresponding production
change (derivation fallback, libpq detection, load widening, preflight
branch) — each revert fails the tests that pin it, including
`TestApplyResolvedScopePostgresEnv_RefusesUnderivableDSN`, which reproduces
the empty-`BEADS_POSTGRES_*` projection this PR prevents.

## Review round 2 — three defects found and fixed

A council review found six majors clustering into three defects. Each fix has
a test that fails without it.

### 1. Password leak on the load path (security regression vs main)

`ParsePostgresDSNEndpoint` wrapped `url.Parse`'s `*url.Error` with `%w`.
`net/url` renders that as `%s %q: %s` with the **raw input**, so
`LoadMetadataState` copied a hand-written password into
`MetadataParseError.Reason`, and `cmd/gc` re-emitted it to stderr and into
`gc --json` failure payloads.

This is the same leak class the preflight-summary fix above targets, surviving
one frame up on the path every `gc` command traverses.

Measured end to end with a built binary, not reasoned about. With
`.beads/metadata.json` holding
`postgres://operator:sw%ordfish@db.example.com:5432/gascity`:

| binary | occurrences of the raw DSN in one `gc doctor` |
| --- | --- |
| base `main` | **0** (static, DSN-free reason) |
| this PR, before the fix | **98** (fires per-probe in a loop) |
| this PR, after the fix | **0** |

`gc bd list --json`, the machine-captured surface, went 2 -> 0. Reproduced for
four malformation classes (`%`-escape, `/` in password, non-numeric port,
unbalanced bracket) — 98 -> 0 in every one.

Unwrapping to `url.Error.Err` is **not** sufficient: `invalid port ":sword"
after host` and `invalid URL escape "%or"` still quote password fragments. The
reason is now a static sentinel. Note `TestParsePostgresDSNEndpoint`'s
non-numeric-port case asserted `wantErr: "port"`; it now asserts the sentinel,
because naming the port requires quoting the DSN.

Tests: `TestParsePostgresDSNEndpointErrorsNeverEchoTheDSN` and the load-path
twin `TestLoadMetadataStateErrorNeverEchoesPostgresDSN` (both table-driven over
the four shapes), plus `TestApplyResolvedScopePostgresEnv_RefusalOmitsDSN`
pinning the `cmd/gc` re-wrap.

### 2. Dolt scopes carrying bd's residue hard-failed, with no self-heal

Adding `postgres_dsn`/`postgres_schema` to `mixedBackendField` made
`backend=dolt` plus either key a fatal `MetadataParseError` — on
`applyCanonicalScopeBackendEnv`, i.e. every bd subprocess env projection, and
on `normalizeCanonicalBdScopeFiles` at startup. Main loads those scopes fine.
Worse, gc could not repair it: the canonicaliser that would scrub the keys
calls `LoadMetadataState` first and returns the error.

Verified end to end: `gc doctor` on `{"backend":"dolt","dolt_database":"hq",
"postgres_dsn":...,"postgres_schema":...}` printed `cannot mix dolt and
postgres fields` 98 times before the fix; base main and the fixed binary both
print it 0 times.

**Fix chosen: drop the two keys from `mixedBackendField`, keep them in
`postgresBackendKeys`.** Justification — the two field families are not
symmetric. `postgres_dsn`/`postgres_schema` are *bd's own* fields, retained
across a re-init (`bd init` preserves existing config and flips `Backend`), so
a Postgres workspace moved to Dolt carries them with no operator involvement.
The discrete `postgres_host/port/user/database` fields are gc's draft-era
shape that no bd has ever written, so a dolt scope carrying those is genuine
corruption and stays fatal. This guard exists to catch configurations gc
cannot interpret; a declared `backend=dolt` is not ambiguous and the residue
is inert. `EnsureCanonicalMetadata` scrubs it on the next canonicalise.

The alternative — extending `allowLegacyDoltMetadataRepair` — was rejected: it
keeps a fatal guard on the hot path that only the init/repair path can bypass,
so every non-repair reader stays broken until repair happens to run.

Test: `TestLoadMetadataStateToleratesDoltScopeWithPostgresResidue` asserts the
scope loads, the residue survives into the state so canonicalisation can see
it, `EnsureCanonicalMetadata` removes it, and the result reloads clean.
`TestLoadMetadataStateRejectsDoltWithPostgresField` keeps the discrete-field
rejection. Fixture `reject_dolt_with_postgres_dsn.json` renamed to
`dolt_with_postgres_residue.json`.

### 3. Multi-host and unbracketed-IPv6 DSNs derived a mangled host

`url.Parse` accepts these and splits them wrongly — and every downstream gate
passed the result, which gc then projects as `BEADS_POSTGRES_HOST` and keys
credential lookups on:

| DSN | derived host | derived port |
| --- | --- | --- |
| `postgres://bd@h1.test:5432,h2.test:5432/beads` | `h1.test:5432,h2.test` | `5432` |
| `postgres://bd@::1/beads` | `:` | `1` |

A silently wrong endpoint is worse than a refusal, which is this program's
whole thesis. Both now fail with `ErrPostgresDSNUnsupportedForm`. Bracketed
IPv6 (`postgres://bd@[::1]:6543/beads`) is unaffected and still covered.

Test: `TestParsePostgresDSNEndpointRejectsAmbiguousHosts`.

### Also

`TestPreflightContractShapeAgreesWithLoadMetadataState` asserts, per case,
that contract_shape's verdict and `LoadMetadataState`'s acceptance agree — so
the two gates cannot drift apart again.

## Gates

Re-run after round 2, rebased onto `origin/main` @ `5c217feff3`:
`go build ./...`, `go vet ./...`, `gofmt -l` clean, `golangci-lint` (2.10.1)
0 issues on `internal/beads/contract`, `internal/doctor/checks`, `cmd/gc`,
`go test ./internal/beads/... ./internal/doctor/...` green, `go test ./cmd/gc`
green (needs `-timeout 25m`; the package exceeds the 10m default on main
too), `scripts/check-core-boundary.sh` OK. The leak fix is additionally proven
end to end with a built binary, not only by unit test.

## Not included

- No `BEADS_POSTGRES_SCHEMA` projection. `postgres_schema` is carried so the
  on-disk shape round-trips and scrubs coherently; wiring bd's `search_path`
  is a separate change.
- The `DeferIdentityToNativeOpen` preflight work that sits in the same
  source branch is unrelated to DSN derivation and is not ported here.

### DIFF
--- a/cmd/gc/bd_env.go
+++ b/cmd/gc/bd_env.go
@@ -630,7 +630,8 @@ func scopeMetadataJSONPath(scopeRoot string) string {
 // applyResolvedScopePostgresEnv projects PG credentials and connection
 // info into env. Caller guarantees meta.Backend == "postgres". The
 // resolver chain in internal/pgauth supplies the password; the host,
-// port, user, and database come straight from MetadataState.
+// port, user, and database come from meta.PostgresEndpoint, which reads
+// either the discrete metadata fields or bd's postgres_dsn.
 //
 // On resolver exhaustion returns an error wrapping
 // pgauth.ErrNoPasswordResolvable; callers can match with errors.Is.
@@ -642,14 +643,21 @@ func applyResolvedScopePostgresEnv(env map[string]string, cityPath, scopeRoot st
 	if env == nil {
 		return nil
 	}
+	// LoadMetadataState refuses a postgres scope it cannot derive a complete
+	// endpoint for, so this only fires for a hand-built MetadataState.
+	// Failing here beats projecting empty BEADS_POSTGRES_* values.
+	pg, err := meta.PostgresEndpoint()
+	if err != nil {
+		return fmt.Errorf("deriving postgres endpoint for %s: %w", scopeRoot, err)
+	}
 	clearProjectedBeadsBackendEnv(env)
 	clearProjectedDoltEnv(env)
 	mirrorBeadsDoltEnv(env)
 	clearProjectedPostgresEnv(env)
 	endpoint := pgauth.Endpoint{
-		Host: meta.PostgresHost,
-		Port: meta.PostgresPort,
-		User: meta.PostgresUser,
+		Host: pg.Host,
+		Port: pg.Port,
+		User: pg.User,
 	}
 	// Scope projection clears inherited PG keys first, so credential
 	// resolution intentionally starts at process and file-backed sources.
@@ -659,12 +667,12 @@ func applyResolvedScopePostgresEnv(env map[string]string, cityPath, scopeRoot st
 	}
 	env["GC_POSTGRES_PASSWORD"] = resolved.Password
 	env["BEADS_POSTGRES_PASSWORD"] = resolved.Password
-	env["BEADS_POSTGRES_HOST"] = meta.PostgresHost
-	env["BEADS_POSTGRES_PORT"] = meta.PostgresPort
-	env["BEADS_POSTGRES_USER"] = meta.PostgresUser
-	env["BEADS_POSTGRES_DATABASE"] = meta.PostgresDatabase
+	env["BEADS_POSTGRES_HOST"] = pg.Host
+	env["BEADS_POSTGRES_PORT"] = pg.Port
+	env["BEADS_POSTGRES_USER"] = pg.User
+	env["BEADS_POSTGRES_DATABASE"] = pg.Database
 	mirrorBeadsPostgresEnv(env)
-	emitPostgresCredentialResolved(cityPath, scopeRoot, meta, resolved.Source)
+	emitPostgresCredentialResolved(cityPath, scopeRoot, pg, resolved.Source)
 	return nil
 }
 
@@ -673,7 +681,7 @@ func applyResolvedScopePostgresEnv(env map[string]string, cityPath, scopeRoot st
 // Best-effort: recorder failures (file unreachable, JSONL write error) do
 // not propagate to the caller. The payload deliberately omits the password
 // value (asserted by TestPostgresEventOmitsPassword).
-func emitPostgresCredentialResolved(cityPath, scopeRoot string, meta contract.MetadataState, source pgauth.Source) {
+func emitPostgresCredentialResolved(cityPath, scopeRoot string, pg contract.PostgresEndpoint, source pgauth.Source) {
 	scopeKind, scopeName := scopeKindAndName(cityPath, scopeRoot)
 	subject := "city/" + scopeName
 	if scopeKind == "rig" {
@@ -683,9 +691,9 @@ func emitPostgresCredentialResolved(cityPath, scopeRoot string, meta contract.Me
 		ScopeKind: scopeKind,
 		ScopeName: scopeName,
 		Source:    source.String(),
-		Host:      meta.PostgresHost,
-		Port:      meta.PostgresPort,
-		User:      meta.PostgresUser,
+		Host:      pg.Host,
+		Port:      pg.Port,
+		User:      pg.User,
 	}
 	if _, loaded := postgresCredentialResolvedSeen.LoadOrStore(postgresCredentialResolvedKey(cityPath, payload), struct{}{}); loaded {
 		return
@@ -1283,7 +1291,11 @@ func bdCommandRunnerWithManagedRetryErr(cityPath string, envFn func(dir string) 
 				return out, fmt.Errorf("classifying scope backend (bd error: %w): %w", err, classifyErr)
 			}
 			if ok {
-				return out, fmt.Errorf("postgres at %s:%s: gc does not manage external PG endpoints (no managed recovery attempted): %w", meta.PostgresHost, meta.PostgresPort, err)
+				pg, pgErr := meta.PostgresEndpoint()
+				if pgErr != nil {
+					return out, fmt.Errorf("postgres scope with an underivable endpoint (%w): gc does not manage external PG endpoints (no managed recovery attempted): %w", pgErr, err)
+				}
+				return out, fmt.Errorf("postgres at %s:%s: gc does not manage external PG endpoints (no managed recovery attempted): %w", pg.Host, pg.Port, err)
 			}
 		}
 		if err == nil && scopeBackendIsPostgres(cityPath, dir) {
--- a/cmd/gc/bd_env_test.go
+++ b/cmd/gc/bd_env_test.go
@@ -4307,6 +4307,66 @@ func TestApplyResolvedScopePostgresEnv_HappyPath(t *testing.T) {
 	}
 }
 
+// TestApplyResolvedScopePostgresEnv_DerivesFromDSN covers bd's own postgres
+// metadata shape: a password-free postgres_dsn with no discrete fields. The
+// BEADS_POSTGRES_* projection has to derive the tuple rather than ship empty
+// values, which is what it would do if it read the metadata fields directly.
+func TestApplyResolvedScopePostgresEnv_DerivesFromDSN(t *testing.T) {
+	clearAmbientPostgresEnv(t)
+	cityPath := t.TempDir()
+	scopeRoot := t.TempDir()
+	writePGScopeFixture(t, scopeRoot, "devpw")
+
+	env := map[string]string{}
+	meta := contract.MetadataState{
+		Backend:        "postgres",
+		PostgresDSN:    "postgres://gc_city@pg.example.test:6543/gascity_infra",
+		PostgresSchema: "city_x",
+	}
+	if err := applyResolvedScopePostgresEnv(env, cityPath, scopeRoot, meta); err != nil {
+		t.Fatalf("applyResolvedScopePostgresEnv: %v", err)
+	}
+	want := map[string]string{
+		"BEADS_POSTGRES_HOST":     "pg.example.test",
+		"BEADS_POSTGRES_PORT":     "6543",
+		"BEADS_POSTGRES_USER":     "gc_city",
+		"BEADS_POSTGRES_DATABASE": "gascity_infra",
+	}
+	for key, value := range want {
+		if got := env[key]; got != value {
+			t.Errorf("env[%q] = %q, want %q", key, got, value)
+		}
+	}
+}
+
+// TestApplyResolvedScopePostgresEnv_RefusesUnderivableDSN pins the libpq
+// limitation at the projection boundary: gc fails loudly instead of handing bd
+// a set of empty connection variables.
+func TestApplyResolvedScopePostgresEnv_RefusesUnderivableDSN(t *testing.T) {
+	clearAmbientPostgresEnv(t)
+	cityPath := t.TempDir()
+	scopeRoot := t.TempDir()
+	writePGScopeFixture(t, scopeRoot, "devpw")
+
+	env := map[string]string{}
+	meta := contract.MetadataState{
+		Backend:     "postgres",
+		PostgresDSN: "host=pg.example.test port=6543 user=gc_city dbname=gascity_infra",
+	}
+	err := applyResolvedScopePostgresEnv(env, cityPath, scopeRoot, meta)
+	if err == nil {
+		t.Fatalf("applyResolvedScopePostgresEnv = nil error, want refusal; env=%v", env)
+	}
+	if !errors.Is(err, contract.ErrPostgresDSNUnsupportedForm) {
+		t.Fatalf("error %v, want errors.Is(err, contract.ErrPostgresDSNUnsupportedForm)", err)
+	}
+	for _, key := range []string{"BEADS_POSTGRES_HOST", "BEADS_POSTGRES_PORT", "BEADS_POSTGRES_USER", "BEADS_POSTGRES_DATABASE"} {
+		if got, ok := env[key]; ok {
+			t.Errorf("env[%q] = %q, want it unset (no partial projection)", key, got)
+		}
+	}
+}
+
 func TestEmitPostgresCredentialResolved_DedupsWithinProcess(t *testing.T) {
 	clearAmbientPostgresEnv(t)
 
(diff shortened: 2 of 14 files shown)


julianknutsen:
## Review round 2 — Defect 0 verdict, then three fixes

Pushed `b85582e778` (rebased onto `origin/main` @ `5c217feff3`). Full detail is
in the updated PR description; this is the short version.

### Defect 0 first: is the premise true? **Yes — but not from OSS bd**

A reviewer challenged the whole justification: that no supported bd writes
`postgres_dsn`, so "operators shouldn't hand-convert away from the shape bd
writes" is vacuous. Establishing this decided whether the rest was worth doing.

**The writer exists.** `gascity/bd-enterprise` branch `enterprise`
@ `754dbe10af`, `cmd/bd/backend_init_distribution_enterprise.go` (build tag
`//go:build enterprise`), `runInitPostgres`:

```go
persistDSN, err := pgdialect.RedactPassword(dsn)   // L135, fail-closed
cfg.Backend = configfile.BackendPostgres
cfg.PostgresDSN = persistDSN                       // L186
cfg.PostgresSchema = in.pgSchema                   // L187
cfg.Save(in.beadsDir)                              // L189
```

Read side: `internal/storage/postgres/fromconfig.go` L41-43 —
`postgres: no DSN (set postgres_dsn in metadata.json, or BEADS_POSTGRES_URL)`.
`BEADS_PG_PASSWORD` is real (`internal/storage/postgres/credential.go` L41-42),
not invented. Binaries carrying this writer are deployed at
`/opt/beads/releases/mc-postgres-p0-*`, `mc-postgres-field-p0-*`,
`mc-field-integrated-*`.

**The reviewer was right about one thing:** OSS bd contradicts a naive reading.
At `BD_VERSION=v1.1.0` and `BD_PREV_VERSION=v1.0.4` the field does not exist at
all; at `BD_CURRENT_REF=bf97b73749ac` it appears exactly once
(`internal/configfile/configfile.go:42`) as deprecated round-trip storage, and
`bd init --backend=postgres` returns `storage backend "postgres" is no longer
supported`. The old comments cited *that* file as their authority. They now cite
the add-on. Both halves matter: the add-on **writes** the shape, OSS bd
**round-trips it without clearing it** — which is precisely why gc must tolerate
it (defect 2 below).

### 1. Password leak on the load path — proven end to end, not reasoned about

`ParsePostgresDSNEndpoint` wrapped `url.Parse`'s `*url.Error` with `%w`;
`net/url` renders that as `%s %q: %s` with the raw input. Built `./cmd/gc`, ran
it against `.beads/metadata.json` holding
`postgres://operator:sw%ordfish@db.example.com:5432/gascity`:

| binary | raw DSN occurrences in one `gc doctor` |
| --- | --- |
| base `main` | **0** |
| this PR, before fix | **98** |
| this PR, after fix | **0** |

`gc bd list --json` — the machine-captured surface — went 2 -> 0. Same 98 -> 0
for `/` in password, non-numeric port, and unbalanced bracket. Unwrapping to
`url.Error.Err` is *not* enough (`invalid port ":sword"`, `invalid URL escape
"%or"` still quote fragments), so the reason is a static sentinel.

### 2. Dolt scopes carrying bd's residue hard-failed, unrepairably

`postgres_dsn`/`postgres_schema` in `mixedBackendField` made `backend=dolt` plus
either key fatal on the env-projection hot path and at startup, and gc's own
canonicaliser calls `LoadMetadataState` first, so it could not scrub what it now
rejected. `gc doctor` on such a scope: 98 hard-fails before, 0 after — matching
base main.

**Chosen fix: drop the two keys from `mixedBackendField`, keep them in
`postgresBackendKeys`.** The families are not symmetric — those two are bd's own
fields, retained across a re-init, so a Postgres workspace moved to Dolt carries
them with no operator involvement; the discrete `postgres_host/port/user/database`
fields are gc's draft-era shape no bd has ever written, so those stay fatal. A
declared `backend=dolt` is unambiguous and the residue is inert, so it is scrubbed
on the next canonicalise rather than made fatal. Extending
`allowLegacyDoltMetadataRepair` was rejected: it leaves the fatal guard on a path
only the repair route can bypass.

### 3. Two minors

- Multi-host and unbracketed-IPv6 DSNs derived a mangled host that passed every
  gate: `postgres://bd@::1/beads` -> host `":"`, port `"1"`; `...@h1:5432,h2:5432/...`
  -> host `h1.test:5432,h2.test`. Both now refused. Bracketed IPv6 unaffected.
- `contract_shape` PASSed DSNs `LoadMetadataState` rejects (no database, no user,
  port out of range). Both now call one gate, `validatePostgresEndpoint`, with a
  test asserting per-case agreement so they cannot drift again.

### Tests and gates

Six new tests, each verified red first. Representative red-before strings:

```
message leaks password fragment "ordfish": "parse postgres_dsn: parse
"postgres://operator:sw%ordfish@db.example.com:5432/gascity": invalid URL escape "%or""

LoadMetadataState(dolt_with_postgres_residue.json) = ok=false
err=... cannot mix dolt and postgres fields in a single scope (backend=dolt but postgres_dsn is also set)

ParsePostgresDSNEndpoint("postgres://bd@::1/beads") = {Host:: Port:1 User:bd Database:beads}, want a rejection

checkContractShape state = PASS, want FAIL (summary "Metadata uses postgres_dsn shape")
```

One pre-existing expectation changed deliberately:
`TestParsePostgresDSNEndpoint`'s non-numeric-port case asserted `wantErr: "port"`,
which is unsatisfiable without quoting the DSN; it now asserts the sentinel.

`go build ./...`, `go vet ./...`, `gofmt -l` clean, `golangci-lint` 2.10.1 0
issues, `go test ./internal/beads/... ./internal/doctor/...` green,
`go test ./cmd/gc -timeout 25m` green (860s), `scripts/check-core-boundary.sh` OK.

Still **do not merge** — this is a fix pass on an open review, not a merge
request.

PR #4875, opened 2026-07-31
fix(dolt): send SIGTERM before SIGKILL in run_bounded's python3 fallback (#4823)

rjgeng:
## What this fixes

`run_bounded`'s python3 fallback (used when neither `timeout` nor `gtimeout` is on PATH — the default on stock macOS) was escalating straight to SIGKILL on timeout, contradicting its own documented SIGTERM-then-2s-grace-then-SIGKILL contract. The `TIMEOUT_BIN` branch (`timeout --kill-after=2`) already implemented this correctly — only the python3 fallback was wrong.

**Why it matters:** `mol-dog-backup.sh` wraps `dolt backup sync` in this helper. `dolt` publishes a backup archive under its final name *before* writing the manifest that references it, so a bare SIGKILL mid-sync leaves a permanently unreferenced archive (`dolt backup` has no prune verb). The reporter measured 2.72 GB of orphaned archives from one day of intermittent timeouts.

## Scope

Deliberately narrow to the core contract violation (SIGTERM-before-kill, streaming instead of buffering output). Left out the issue's two "optional" asks — a configurable sync timeout env var and a post-kill orphan-sweep — since those are separate, independent enhancements, not part of the contract violation itself.

## Testing

New regression tests in `examples/bd/dolt/runtime_bounded_test.go`, both TDD'd red→green against the actual bug (verified by temporarily stashing the fix and re-running):

- `TestRunBoundedPython3FallbackSendsSigtermBeforeKill` — child installs a Python SIGTERM handler; asserts the handler ran (marker file written) before the process was force-killed. Fails on pre-fix code.
- `TestRunBoundedPython3FallbackEscalatesToSigkillAfterGrace` — child ignores SIGTERM; asserts `run_bounded` still returns quickly (SIGKILL after the ~2s grace), not left running the full duration.

Full `examples/bd/dolt` package suite passes, `go vet ./...` clean.

[...]

PR #4853, opened 2026-07-30
fix(pidutil): make the argv identity check work off Linux

wbern:
## The juicy parts

`AliveWithCmdline` — the "is this PID still *my* process" check — opened with `if runtime.GOOS != "linux" { return true }`. Off Linux that is not an identity check at all, just an existence check.

Both production callers use it to decide whether the PID in a poller pidfile is still their nudge poller (`cmd/gc/cmd_nudge.go`, `internal/session/submit.go`). So on macOS a stale pidfile whose PID has since been recycled to *any* unrelated live process reads as "poller already running": `gc nudge poll` hits `errNudgePollerRunning` and returns exit `0` without starting one. Exit success, nil error, nothing logged — and nudge/submit delivery for that target quietly stops. Stale *dead* pidfiles were always reaped correctly via the portable `Alive`, so the residual hole is exactly PID reuse.

Why it survived: twelve tests asserted the correct semantics and then `t.Skip`'d off Linux — nine of them the callers' own ownership tests, including `TestExistingPollerPIDRejectsUnrelatedLivePID`, which asserts precisely the answer that was inverted. All twelve now run on both platforms.

## Summary

`AliveWithCmdline` answers *"is this PID the process I think it is"* by comparing
argv. It returned `true` unconditionally when `GOOS != linux`, because `Cmdline`
read only `/proc/<pid>/cmdline`:

```go
if runtime.GOOS != "linux" {
    return true
}
```

That turns an identity check into a bare existence check on every non-Linux host.

Its callers use it to decide whether the PID in a poller pidfile is still
**their** poller. A recycled PID owned by an unrelated live process therefore
reads as "poller already running", and the caller reports success without
starting one — `cmd/gc/cmd_nudge.go` returns `0`, `internal/session`'s submit
path returns `nil`. **Exit success, nil error, nothing logged**, and nudge or
submit delivery for that target silently stops.

Stale *dead* pidfiles are still reaped correctly (that path uses the portable
`Alive`), so the residual hole is exactly PID reuse — and `doctor_fork_rate.go`
records a `gc` process per `bd` command, which is the churn that makes reuse
likely on a long-lived host.

## Why it went unnoticed

Twelve tests asserted the correct semantics and then skipped where they mattered,
all with `t.Skip("... uses /proc on linux")`. The clearest case is
`TestExistingPollerPIDRejectsUnrelatedLivePID`, which asserts `false` for an
unrelated live PID — precisely the answer that was inverted off Linux — and then
[...]

PR #3527, opened 2026-06-15
Block unreleased go.mod replace targets in CI

quad341:
## What this changes

Pull requests now run a `make check-gomod-replace` preflight that blocks `go.mod` replace directives pointing at local paths, pseudo-versions, branch names, or prerelease tags. The guard handles both single-line replacements and grouped multi-line `replace (...)` blocks, so the CI check covers the forms authors can write by hand.

The policy lives in `scripts/check-gomod-replace.sh` and is wired through `Makefile` and the preflight CI job. That gives local runs and GitHub Actions the same entry point and keeps the check focused on developer tooling rather than runtime behavior.

## Review notes

- Released semver replacement targets continue to pass; the guard is scoped to unreleased/local replacement targets.
- The parser tracks whether it is inside a grouped replace block and applies the same RHS checks used for single-line replace directives.
- The failure text names the released-dependency policy and the human-operator-only override path.
- No runtime behavior changes; this only affects CI and local developer checks.

## Test plan

- [x] `go test ./scripts/ -v -run TestCheckGomodReplaceGuard`
- [x] `make check-gomod-replace`
- [x] `make test`
- [x] `go vet ./...`
- [x] Release gate: [`release-gates/ga-jrsux8-gomod-replace-multiline-gate.md`](release-gates/ga-jrsux8-gomod-replace-multiline-gate.md)

PR #5134, opened 2026-08-09
test(beads): run every split-store ownership invariant on both topologies

julianknutsen:
Re-derives the split-store conformance suite onto `main`'s storage stack and lands the drift guard that keeps it honest (ga-zdin7).

**Test-only. No production code changed.** Verified: `git diff --stat` touches `Makefile`, two `_test.go` files, and one script.

## Why

Two production bugs in one week, both the same shape — code correct on one store arrangement and wrong on the other:

- **#5127** — order-tracking beads created through the order's own target (work) store on a split city. Stranded infrastructure rows accrued until boot refused.
- **#5125** — `bd sql` kept answering from the work ledger for relocated graph ids and reported every live molecule root as missing.

Both were caught by inspection, not by a test, because nothing ran the same assertion on both store arrangements. This is the fixture that does.

## The fixture

`splitEnv` (`cmd/gc/split_topology_env_test.go`) builds ONE city twice:

- **single-store** — `routes == nil`, `resolveClassStore` collapses every class onto the work store (byte-identical to a city that authors no `[storage]`).
- **split** — routes relocate all five infrastructure classes to one binding store, exactly as `storageSplitWhole` / `openStorageRoutes` arrange them, with `cfg.Storage` saying the same thing so `relocatedBeadClasses` and `storageSplitShapeOf` agree with the routing. A fixture self-test enforces that agreement, so it cannot stage a city no boot could produce.

Leaves come from `internal/beads/splittest` (#5130), so residence is real rather than assumed:

| leaf | semantics | mints | foreign-prefix create / cross-store dep |
|---|---|---|---|
| work | `BdSemantics` | `gc-` | hard failure, in bd's wording |
| class | `SQLiteSemantics` | `gcg-` | accepted, recorded as a residence violation the fixture must claim |

The policy wrapping is **asymmetric because production's is**: `main.go`/`api_state.go` put every work store (city and rig) behind `wrapStoreWithBeadPolicies`, while `openStorageRoutes` keys the class map straight to the value the provider's `OpenEngine` returned. So a relocated class store has no bead-policy layer — no create-time storage policy, no read-tier expansion. `splitEnv.policyWrapped` reports which leg a front door is on and the invariants that care assert through it instead of assuming. On a split city a `gc.kind=wisp` create is therefore a durable row, not an ephemeral one; that is pinned, in both directions, rather than papered over.

That asymmetry is **enforced at the seam that creates it**. `TestSplitEnvClassStoreWrappingMatchesOpenStorageRoutes` resolves a real storage plan, calls `openStorageRoutes`, and fails if the store it hands back for any relocated class carries a bead-policy layer. Without it the fixture's `e.class = classLeaf` would be an unchecked assumption: nothing else in `cmd/gc` asserts the wrapping of the value that function returns, so a change that started wrapping the class map would leave every invariant testing a topology production no longer serves.

## Invariants: what each pins, and which one skips

| # | Seam exercised | What it pins |
|---|---|---|
| I1 ready-federation | `collectOpenUnassignedRoutedWork` led by `sessionsStore()` | Routed open work in the durable AND wisp shape surfaces in the cross-store demand scan, with the rig leg alongside and owner store / store-ref aligned. A leading store resolved to the work class reads zero and drains the fleet. Also carries the **HQ work-store leg** and its named gap (below). |
| I2 assigned-work-capture | `collectAssignedWorkBeadsWithStores` → `releaseOrphanedPoolAssignments` | A dead-assignee claim is captured under the leading arm and released; a live holder's claim survives (through the last-resort live-session probe). Same outcome on both topologies for the class-store and rig legs; the **HQ work-store leg diverges** and is pinned as a named gap (below). |
| I3 by-id-write-residence | `resolveClassStore(BeadClassGraph)` Update/SetMetadata/Close | The **#5127** class. Writes land in the owning store and leave no shadow row in the work store. |
| I4 materialization-residence | `molecule.Instantiate` through the graph front door | The invariant `memoryOrderDispatcher.graphStoreFor` exists to satisfy: every materialized bead class-resident, zero in the work ledger. |
| I5 claim-routing | `storeref.PrefixOwner` | **Partially skipped, see below.** |
| I6 strict-cross-store-deps | `DepAdd` through both front doors | Stated per **authority**, not per backend. A cross-prefix target is refused by the splittest kit's DOMAIN co-residence rule (`convoy.TrackItemIn`'s `ErrMemberNotCoResident`) — **neither production backend refuses it**: bd passes a cross-prefix target through as an external ref and writes the row, and SQLite writes it too. A same-prefix unresolvable endpoint is the one bd really hard-fails, and is asserted on both topologies in bd's own wording. The class front door accepts and records. |
| I7 by-id-read-federation | `storeref.Resolve` + `bdSQLRelocatedClassRefusal` | The **#5125** class. The federating point read finds class-resident beads; `bd sql`/`bd query` naming a relocated id is refused and points at the class-routed verb; a plain work-ledger query is never refused; the whole guard is inert single-store. |
| I8 residence-sweep | `coordclass.Classify` over a representative population | Classification matches residence for every bead, dep endpoints co-reside, and the reserved id-namespace boundary holds. This is the check that made #5127 fatal-at-boot rather than silent. |
| I9 warm-tick-demand | `buildDesiredStateWithSessionBeads`, cold + 2 warm ticks | Demand and desired sessions survive consecutive warm ticks without minting replacements — the spawn/drain treadmill driver. |
| I10 wake-ownership | `filterAssignedWorkBeadsForSessionWake`, `makeOpenSessionStoreRefIndex`, `openSessionOwnsWork` | Reachability is cfg-derived, so the answer is **identical on both topologies**. Known gap named in-code, below. |
| I11 read-path-consistency | policy front door vs raw leaf, `listOpenForControllerDemandLive` | The reader the controller actually uses is never blind to work the store holds — on both legs. On the wrapped leg it also pins the operator-confusion shape (`bd list` main-tier-only vs tier-expanded front door). The unwrapped class leg asserts that every read path sees everything the store holds; the **wrapping asymmetry itself** is pinned at the production seam by `TestSplitEnvClassStoreWrappingMatchesOpenStorageRoutes`, not inside I11. |

### Skipped, with reasons (the program's verified remaining work)

**I5 claim-routing — CLAIM-MUTATION half skipped, both topologies.** The seam does not exist on `main`. `gc hook --claim` resolves stores as `hookStore{dir, env}` pairs (`hook_cross_store.go`) and execs `bd` in a work directory; the fan-out is city + rigs, all WORK scopes, with no coordination-class arm anywhere on the path. The only by-id class resolver in the tree is `Server.beadStoresForID` (`internal/api/handler_beads.go`) — method-bound to the API server and graph-only. Lifting it is **ga-ia7li**, which this work blocks.

What I5 *does* assert before skipping: `storeref.PrefixOwner` routes both the ordinary class id shape and the `<prefix>-wisp-<suffix>` shape to the class store, and work ids fall through. Plus one trap pinned as a negative — the config-free `sling.BeadPrefix` heuristic answers `"gcg-wisp"` for a `gcg-wisp-0042` id, which is not a reserved class prefix at all, so a by-id resolver built on that heuristic instead of the namespace rule would hand every wisp to the work store. `main`'s own routing (`beadIDHasConfiguredPrefix`) uses the namespace rule and is correct; this pins the divergence so it cannot change silently under the future shared resolver.

**I1/I2 — the HQ work-store leg, gap named in-code, not skipped.** The controller's cross-store scans take the leading store plus the RIG stores, and production builds the rig arm as `rigBeadStores()`, which deletes the city entry (`city_runtime.go`). On a split city the leading store is the class store, so the **HQ work store is in neither arm**: a city-scope routed WORK bead is invisible to controller-tick demand, and a dead claim on one is captured and released by nothing. `cmd_start.go` already names the dual role as a shared E2 two-store split; `gc session close` still reaches the store via `unclaimWorkAssignedToRetiredSessionBead`, so the bead is not lost forever — the controller tick is what is blind. I1 and I2 now **seed** that leg (a routed HQ work bead, and a dead-claimed one) and assert today's asymmetric answer on both topologies, with the assertions to flip when the arm lands named in each message. Leaving it unseeded was the reason the divergence read as covered.

**I10 — gap named in-code, not skipped.** On a split city the leading store IS the class store, so a rig-bound pool holder's claimed class-resident bead has no wake reason and is not ownership-matched: the reachability model has no coordination-class arm. Orphan release still spares it via the live-session probe (I2 proves that), but the wake filter drops it every tick. The invariant asserts today's behaviour and states that when the arm lands, both topology arms must be updated together or the topologies stop agreeing. Nothing here asserts a capability `main` lacks.

## The drift guard

`scripts/check-split-topology-rows.sh`, wired into `make check`:

- **Rule A** — every `t.Run("I<n>…")` must invoke `forEachTopology`/`forEachTopologyWithRig` **anywhere in its subtest body**. The body is read by brace balance, so the idiomatic multi-line subtest is accepted and a helper call on a later line still counts. The helper list is explicit: a one-topology lookalike (`forEachTopologySplitOnly`) does not satisfy the rule by resembling the real ones.
- **Rule B** — a suite file must never call `newSplitEnv*` directly. The fixture file that *defines* the fan-out helpers is exempt.
- **Rule C** — finding zero suite files, or zero invariants across them, is itself a failure. That is the vacuous-denominator lesson `check-routed-test-rows.sh` already learned: a rename of the naming convention would otherwise drop the scan to zero and let A and B pass while policing nothing.
- **Discovery** — the suite is found by content, not by a hard-coded path: any `cmd/gc/*_test.go` declaring a `t.Run("I<n>…")` invariant is a suite file, so a second conformance file is policed the day it is added. Doc comments quoting either shape are excluded, so prose is neither a violation nor a denominator.

**`scripts/check_split_topology_rows_test.go`** is the guard's own test, on the `scripts/gomod_replace_guard_test.go` pattern: it execs the script against the real repo root (must exit 0 and be silent), against three conforming synthetic shapes, and against six drift shapes that must each exit non-zero naming the rule. `./scripts` is inside `UNIT_COVER_PKGS_NONCMDGC`, which CI runs as *Preflight / unit cover (noncmdgc)* — so the guard now **executes in CI** with no `ci.yml` edit and no workflow shape-hash bump. It also pins the Makefile wiring (`check:` runs it, `.PHONY` declares it).

### Bite evidence

Every case below is a subtest of `TestCheckSplitTopologyRowsCatchesDrift`, so the evidence is executable rather than pasted. A deliberately non-conforming invariant:

```
ROW-GUARD: cmd/gc/split_topology_conformance_test.go:5 invariant subtest does not run both topologies (its body calls none of: forEachTopology, forEachTopologyWithRig): t.Run("I12-drifted", func(t *testing.T) { bodyDrift(t, newSplitEnv(t, true)) })
ROW-GUARD: cmd/gc/split_topology_conformance_test.go:5 direct newSplitEnv bypasses the fan-out helpers (pins one topology): t.Run("I12-drifted", func(t *testing.T) { bodyDrift(t, newSplitEnv(t, true)) })
---
Split-topology row violations: 2 (over 2 invariants in 1 file(s))
exit=1
```

The same drift in a multi-line body, a one-topology helper lookalike, a suite helper minting its own env, and drift in a **second** conformance file all report and exit 1. Renaming the `t.Run("I…")` convention away (Rule C):

```
ROW-GUARD: no cmd/gc test file declares a t.Run("I<n>...") invariant; the two-topology guard found no suite to police.
exit=1
```

Conforming shapes that must NOT be reported are pinned too (`TestCheckSplitTopologyRowsAcceptsConformingSuites`): single-line subtests, multi-line subtests, and doc comments that quote `t.Run("I3-example")` or `newSplitEnv(t, true)` in prose — the last two were false positives of the original guard.

Clean tree: `exit=0`, silent (asserted by `TestCheckSplitTopologyRowsPassesOnThisRepo`).

`check-split-topology-reverts.sh` is deliberately **not** landed: it is retrospective only, its matrix is nine branch SHAs that do not exist on `main`, and it fails OPEN on merge conflicts so its coverage decays silently. Its technique — prove the new test reds without the fix — is applied below instead.

## Review round 2: five confirmed MAJOR defects, fixed

An adversarial review reproduced five defects by mutation, not by reasoning. Each is fixed here and each fix is re-proved by re-running the mutation.

**D1 — I11's split arm could not fail, and its promised drift guard did not exist.** The `!wrapped` branch claimed "a change that starts wrapping the class store … fails here first." It could not: the fixture hard-codes `e.class = classLeaf` and never calls `openStorageRoutes`, so nothing coupled the fixture's model to the production seam. Two of its checks were dead — `if wisp.Ephemeral` is unreachable (`mintWispWith` already `t.Fatalf`s on that condition) and `wrapped != e.policyWrapped(front)` is a tautology (`policyWrapped` is the same pure type switch that produced `wrapped`). *Fix:* both dead checks deleted; the asymmetry is now anchored at the seam by `TestSplitEnvClassStoreWrappingMatchesOpenStorageRoutes`.

**D2 — I6 asserted a rejection neither production backend performs, and blamed bd.** The work-leg rows are `work blocks-on <class id>`, where only the *target* is cross-prefix — the one case `internal/beads/splittest`'s own rule table documents as "Neither backend rejects this. BdSemantics rejects it anyway, for the DOMAIN co-residence invariant … not for the backend." The asserted string is produced only by the double. *Fix:* the doc and both failure messages now attribute the refusal to the kit's domain rule and say explicitly that bd writes the dangling external ref; a second check fails if the refusal ever starts borrowing bd's wording; and a **new row** exercises the same-prefix unresolvable endpoint, which bd genuinely hard-fails, on both topologies.

**D3 — I1/I2 never seeded the city WORK store**, the one leg whose reachability actually flips with the topology. *Fix:* both invariants now mint an HQ-work bead (routed, and dead-claimed) in `e.work` and assert capture/release on both topologies, with the divergence stated as a named gap in I10's style.

**D4 — I5's `PrefixOwner` "no owner" negatives pinned a policy-wrapper artifact.** They passed only because `beadPolicyStore` embeds `beads.Store` as an interface, so the optional `IDPrefix()` accessor is not promoted and `PrefixOwner` skips the leg entirely — while the failure text blamed the id space ("a work id is outside every declared class namespace"). Forwarding `IDPrefix` through the wrapper is a correct change the shared by-id resolver (ga-ia7li) wants, and it reddened the suite. *Fix:* the single-store negative is dropped (with the reason recorded in-code), and the split negative now asserts what is meant — no RESERVED-class store claims a work id.

**D5 — the drift guard had no automatic execution path and no self-test.** `make check` is invoked by no workflow, no hook and no script. *Fix:* `scripts/check_split_topology_rows_test.go` puts the guard in CI through the existing `noncmdgc` cover job, and tests its bite. The guard's own minors are fixed with it: Rule A now reads multi-line bodies, discovery replaces the hard-coded single-file scan, comment lines no longer false-positive, and the target is declared `.PHONY`.

### Mutation evidence that each fixed invariant now bites

Every one of these is a **production/kit** mutation applied to a scratch tree; all were reverted (`git status` shows only `_test.go` files, the script, its new test, and the `.PHONY` line).

**D1** — `openStorageRoutes` wrapping the class map, both variants the reviewer named (`wrapStoreWithBeadPolicies(store, nil)` and the tier-flipping bd-1.0.5 cfg). Before: the FULL `cmd/gc` package stayed green. Now:
```
--- FAIL: TestSplitEnvClassStoreWrappingMatchesOpenStorageRoutes
    split_topology_env_test.go:566: openStorageRoutes returned a POLICY-WRAPPED store for class graph. splitEnv models the relocated class store as the raw engine value (e.class = classLeaf, split_topology_env_test.go), and I11 branches on that asymmetry. Wrapping the class map changes wisp storage on a split city from a durable row to an ephemeral one — update this pin, splitEnv's class leg, and I11's wrapped/unwrapped branch together.
```
(one line per relocated class; the `wrapStoreWithBeadPolicies(store, &config.City{... bd-1.0.5})` variant reds identically.)

**D2** — `StrictStore.endpointNotResident`, `crossStore := !isSource` (every missing target treated as cross-prefix, i.e. the same-prefix case gets the domain wording):
```
--- FAIL: .../I6-strict-cross-store-deps/single-store
    split_topology_conformance_test.go:406: work front door: work blocks-on an absent same-prefix id: DepAdd(gc-1 → gc-absent) = adding dep gc-1→gc-absent: gc-absent belongs to another store's id namespace, not "gc" … want bd's own resolution failure (`no issue found matching`) — a same-prefix target gets no external-ref pass-through, so this is the endpoint rejection bd really performs
```
and the opposite mutation, `crossStore := false` (the cross-prefix refusal borrowing bd's wording — the exact misattribution this defect is about):
```
    split_topology_conformance_test.go:424: work front door: work blocks-on wisp: the cross-prefix refusal now borrows bd's resolution wording (…); bd never emits that message for a cross-prefix target, and a domain rule dressed in bd's clothes is what made this invariant misread once already
```

**D3** — `collectOpenUnassignedRoutedWork`, leading arm requiring `gc.kind` (an ISOLATING mutation: only the new leg reds):
```
--- FAIL: .../I1-ready-federation/single-store
    split_topology_conformance_test.go:161: routed HQ work bead gc-3 is missing from the single-store demand scan — on a legacy city the HQ work store IS the leading store, so this is the "no work" fail-open with nothing topological to blame
```
`appendInProgressWorkUnique` with the same predicate, again isolating:
```
--- FAIL: .../I2-assigned-work-capture/single-store
    split_topology_conformance_test.go:223: dead-claimed HQ work bead gc-4 is NOT captured on a legacy city, where the HQ work store IS the leading store — post-claim HQ work has gone invisible to the reconciler
    split_topology_conformance_test.go:245: released = [{gc-3 1}], want exactly [gc-3 gc-4] …
    split_topology_conformance_test.go:249: dead claim gc-4 was not released; it stays assigned to a session that is gone
```
And the split legs, under `resolveClassStore` regressing the sessions class to the work ledger (the #5127 shape):
```
    split_topology_conformance_test.go:156: routed HQ work bead gc-1 IS in the split-city demand scan (store-ref "city:split-topology-city"). Two things produce this and they need opposite responses: the HQ-work arm landed … or the LEADING store regressed to the work class, which is the #5127/#5125 bug and shows up as the durable/wisp rows going missing above
```

**D4** — first the counterfactual that motivated the defect: adding the `IDPrefix()` forwarder to `beadPolicyStore` (the correct routing improvement) previously reddened both I5 subtests. Now `go test ./cmd/gc -run TestSplitTopologyConformance` is **`ok`** with the forwarder applied. And the invariant still bites: `storeref.PrefixOwner` with the `prefix+"-"` segment rule dropped —
```
    split_topology_conformance_test.go:425: durable graph bead: storeref.PrefixOwner("gcg-1") did not route to the class store — a by-id mutation would run against the store that does not hold it
```
— and with the prefix check removed entirely, the **new** negative fires:
```
    split_topology_conformance_test.go:438: work bead: the class store's declared namespace claims work id "gc-1"; the two id spaces must stay disjoint or every by-id router built on the namespace rule misroutes
```

**D5** — guard mutations, caught by the new self-test. Narrowing the Rule A pattern to match nothing (`t\.Run\("ZZZ[0-9]`):
```
--- FAIL: TestCheckSplitTopologyRowsPassesOnThisRepo
    check_split_topology_rows_test.go:93: the guard fails on the checked-in tree (exit 1):
        ROW-GUARD: no cmd/gc test file declares a t.Run("I<n>...") invariant; the two-topology guard found no suite to police.
```
And the residual evasion the reviewer identified — loosening the body predicate so anything satisfies Rule A — which previously escaped all three rules:
```
--- FAIL: TestCheckSplitTopologyRowsCatchesDrift/multi_line_invariant_without_fan_out
    check_split_topology_rows_test.go:254: drifted suite passed the guard
--- FAIL: TestCheckSplitTopologyRowsCatchesDrift/one_topology_helper_lookalike
    check_split_topology_rows_test.go:254: drifted suite passed the guard
```

**Still test-only.** No production file changed; the only non-test edit is the `.PHONY` line.

## Red-before evidence

Each invariant was proved to fail on a deliberately broken fixture (exact strings):

**I1** — demand scan handed `e.work` instead of `e.sessionsStore()` (the leading store resolved to the work class):
```
split_topology_conformance_test.go:111: durable routed control bead gcg-1 is missing from the cross-store demand scan — this is the exact "no work" fail-open: a pool spawns for work its demand read cannot see, then drains
split_topology_conformance_test.go:111: routed wisp gcg-wisp-0001 is missing from the cross-store demand scan — this is the exact "no work" fail-open: a pool spawns for work its demand read cannot see, then drains
```

**I3** — by-id writes routed through the work front door (the wrong-store mutation):
```
split_topology_conformance_test.go:195: durable: Update via the graph class accessor: updating bead "gcg-1": bead not found
```
and the write-dual arm, staging a shadow row in the work leaf:
```
split_topology_conformance_test.go:224: durable: bead gcg-1 resolves in the WORK store (err=<nil>) — a write minted a shadow row; on a split city that row is a stranded infrastructure bead and boot refuses on it
split_topology_conformance_test.go:224: wisp: bead gcg-wisp-0001 resolves in the WORK store (err=<nil>) — a write minted a shadow row; on a split city that row is a stranded infrastructure bead and boot refuses on it
```

**I7** — the class leg dropped from the federated read (reader left on the old store):
```
split_topology_conformance_test.go:433: durable: storeref.Resolve("gcg-1") = ("", bead not found), want the bead — a federating by-id read that misses is the "root does not exist" report of #5125
split_topology_conformance_test.go:433: wisp: storeref.Resolve("gcg-wisp-0001") = ("", bead not found), want the bead — a federating by-id read that misses is the "root does not exist" report of #5125
```

**I8** — the #5127 bug reproduced exactly: the order-tracking bead created through the work store instead of the orders-class store:
```
split_topology_conformance_test.go:529: work store holds gc-3 (type="task" labels=[order-tracking] metadata=map[]): coordclass infrastructure=true — resident on the wrong side of the boundary, which is what boot reads as a stranded bead
```

## Gates

- `go build ./...` — clean
- `go vet ./...` — clean
- `gofmt -l` — clean
- `golangci-lint run ./...` (2.10.1) — **0 issues**
- `shellcheck scripts/check-split-topology-rows.sh` — clean
- `go test ./cmd/gc -timeout 25m` — **ok, 783s**
- `go test ./internal/...` — all ok
- `go test ./scripts/...` (includes the new guard self-test) — ok, 64.9s
- `go test ./internal/testpolicy/... ./internal/testenv/...` (resource census, lint guards) — ok
- `make test-ci-policy` — ok (see note)

**Note on CI wiring — resolved.** A `ci.yml` step was ruled out because `TestCurrentWorkflowsMatchPolicy` pins a SHA-256 of the workflow's execution shape (`run` is in `stepExecutionFields`), so any step addition requires a golden-hash bump — a separate, reviewed CI-shape decision. That reasoning is correct but it covered only one of two routes. The second route is in-tree and precedented: `scripts/gomod_replace_guard_test.go` execs `scripts/check-gomod-replace.sh` from a Go test, and `./scripts` is inside `UNIT_COVER_PKGS_NONCMDGC`, which `ci.yml` runs as *Preflight / unit cover (noncmdgc)*. `scripts/check_split_topology_rows_test.go` takes that route: the guard now runs in CI with **no workflow edit and no hash bump**, and its bite is tested rather than pasted. `make check` still runs it too (belt and suspenders, matching what `check-routed-test-rows` has via its `TestRoutedRowsManifestFullyCovered` Go mirror).

**Nothing needed a production change.** Every invariant is expressed against an existing seam, or states its gap and skips.


julianknutsen:
## Round 2: five confirmed MAJOR defects fixed, each re-proved by mutation

Pushed `a69be3be8f`. The review reproduced five defects by mutation rather than by reasoning, and the headline was right: several invariants **could not fail**. That is worse than no gate on a suite whose whole job is to gate every later slice of this program. All five are fixed, and every fixed invariant was re-run under the mutation it claims to catch.

| # | Defect | Fix | Re-proved by |
|---|---|---|---|
| D1 | I11's `!wrapped` branch promised a drift guard that did not exist; two of its checks were dead (unreachable `wisp.Ephemeral`, tautological `wrapped != policyWrapped`) | dead checks deleted; new `TestSplitEnvClassStoreWrappingMatchesOpenStorageRoutes` opens a real plan through `openStorageRoutes` and fails if any relocated class comes back policy-wrapped | `routes.stores[class] = wrapStoreWithBeadPolicies(store, nil)` **and** the tier-flipping bd-1.0.5 variant — both now RED (previously the FULL `cmd/gc` package stayed green) |
| D2 | I6 asserted a rejection **neither** production backend performs and blamed bd | doc + both failure messages now attribute the refusal to the splittest kit's DOMAIN co-residence rule and state that bd writes the dangling external ref; a guard fails if the wording ever borrows bd's; **new row** for the same-prefix unresolvable endpoint bd genuinely hard-fails, on both topologies | `crossStore := !isSource` (same-prefix gets domain wording) and `crossStore := false` (cross-prefix borrows bd's wording) — both RED |
| D3 | I1/I2 never seeded the city WORK store — the one leg whose reachability actually flips | both mint an HQ-work bead (routed, and dead-claimed) in `e.work`; capture/release asserted on both topologies with the divergence stated as a KNOWN GAP in I10's style | two **isolating** mutations (only the new leg reds), plus the split legs under a `resolveClassStore` sessions→work regression |
| D4 | I5's `PrefixOwner` negatives pinned the policy wrapper's capability opacity while the failure text blamed the id space | single-store negative dropped with the reason recorded in-code; split negative asserts what was meant — no reserved-class store claims a work id | the `IDPrefix()` forwarder (the correct routing change) now leaves the suite **green**; `PrefixOwner` with the segment rule dropped, and with the prefix check removed entirely, both RED |
| D5 | the guard had no automatic execution path and no self-test | `scripts/check_split_topology_rows_test.go` execs it from the `scripts` package — inside `UNIT_COVER_PKGS_NONCMDGC`, which CI runs as *Preflight / unit cover (noncmdgc)* — so it runs in CI with **no `ci.yml` edit and no shape-hash bump** | guard mutations (Rule A pattern narrowed to nothing; body predicate loosened — the residual evasion the review identified) both RED the self-test |

Verifier corrections carried, as instructed: the `nil`-cfg mutant normalizes to bd-1.0.4 so the wisp stays durable, which is why **both** mutants were run for D1. The exact red strings for every mutation are in the PR body.

### Guard minors taken with D5

Rule A now reads the subtest body by brace balance (the idiomatic multi-line form was a false positive); the hard-coded single-file scan is replaced by content discovery over `cmd/gc/*_test.go`, so drift in a second conformance file is caught; comment lines quoting `t.Run("I…")` or `newSplitEnv(...)` no longer count as invariants or violations; the fan-out helper list is explicit, so a one-topology lookalike (`forEachTopologySplitOnly`) does not satisfy the rule by resembling the real ones; `check-split-topology-rows` is declared `.PHONY`.

### Deliberately not done

- **No production code changed.** The same rule that held for the original pass. Every fix is expressed against an existing seam; the only non-test edit in the diff is the `.PHONY` line.
- **The two live gaps are pinned, not fixed.** The HQ-work arm (I1/I2) and the cross-store dep fail-open (I6) are real production divergences. Each now carries a named GAP paragraph, the assertions to flip when it closes, and the slice that closes it. Fixing them is not test work.
- **No `ci.yml` step.** The original CI-deferral reasoning about the workflow shape hash is correct; it just covered only one of two routes. The second route needed neither.

### Gates

`go build ./...` clean · `go vet ./...` clean · `gofmt -l` clean · `golangci-lint 2.10.1` **0 issues** · `shellcheck` clean · `go test ./cmd/gc -timeout 25m` **ok 782.657s** · `go test ./internal/...` all ok · `go test ./scripts/...` ok 64.9s · `scripts/check-core-boundary.sh` OK · `make check-split-topology-rows` OK

julianknutsen:
## CI fix: resource-census ratchet (job 93234211125)

`Integration / packages-core-2-of-4` failed on
`TestRepositoryLedgerMatchesCensusAndDocumentation`. The cause is this PR's
`scripts/check_split_topology_rows_test.go`: its one `exec.Command` grew the
subprocess census by exactly **+1 call / +1 file** on three rows.

```
Small  untagged subprocess  calls=404 (baseline 403), files=113 (baseline 112)
source all      subprocess  calls=607 (baseline 606), files=173 (baseline 172)
source untagged subprocess  calls=411 (baseline 410), files=117 (baseline 116)
```

### Why not match an existing exemption

`scripts/gomod_replace_guard_test.go` is the pattern this test was modelled on,
so I checked it first. It carries **no build tag, no census exemption and no
allowlist entry** — it predates the census (#3527 vs #4218) and simply sits
inside the baseline. There is no third mechanism to copy.

A build tag would also be the wrong trade twice over: it fixes only the two
`untagged` rows, not the all-source audit row, and it pulls the test out of
`Preflight / unit cover (noncmdgc)` — buying the counter by removing the guard
from CI.

### Why not the `[[medium]]` exact-owner exemption

It cannot apply. `SmallCount` only consults the medium map when
`occurrence.Runnable` is true, and the `exec` lives in `runSplitTopologyGuard`,
a shared helper, not lexically inside a `TestXxx(t *testing.T)`. Making it
qualify would mean inlining `exec.Command` into each of the four tests — four
call sites instead of one, strictly worse for the ratchet — and would still not
move the two source rows, which retain the raw census by design.

### What I did

The ordinary "a new process-owning test is genuinely necessary" bump, in the
shape #4875 and #4853 already used: +1 call / +1 file on the three subprocess
rows across all three lockstepped artifacts — `bootstrapPolicy` in `census.go`,
`test/test-resources.toml`, and the generated `TESTING.md` block (regenerated
with `-update`). `reported_*` untouched.

The subprocess is genuinely required here: the test proves the *script* bites,
and a test of a Go reimplementation of the script would prove nothing.

### The self-test still bites

The self-test is **unchanged** by this fix. Verified by mutation — each of these
makes `TestCheckSplitTopologyRowsCatchesDrift` fail:

| Mutation | Subtests that failed |
| --- | --- |
| Defang Rule A helper list | 3 |
| Defang Rule B `newSplitEnv` regex | 1 |
| Defang Rule C empty-denominator check | 1 |
| `exit 0` unconditionally | all 6 |

### Gates

`go build ./...` · `go vet ./...` · gofmt · golangci-lint 2.10.1 (0 issues) ·
`go test ./internal/testpolicy/...` (previously failing — now **ok**) ·
`go test ./scripts/...` · `go test ./cmd/gc -timeout 25m` (ok, 785s) ·
`scripts/check-core-boundary.sh` · `make check-split-topology-rows` — all pass.

PR #5136, opened 2026-08-09
fix(session): stop the retired-session sweeps clobbering a re-claimed assignee

sjarmak:
Two reconciler sweeps computed a work set from a `List` early in a tick, then wrote per bead. A worker can claim a bead in that gap, and both writes were unconditional:

- `ReleaseWorkBead` cleared an assignee it no longer owned and reset a bead someone was actively working.
- `ReassignWorkBead` stamped the retired session's successor over that live claim.

Neither reported an error, so the caller counted a destroyed claim as a completed unassign.

## The guard already existed

`beads.ConditionalAssignmentReleaser.ReleaseIfCurrent`, backed by bd's `--if-assignee`/`--if-status` CAS, is what the pool path already uses in `releaseOrphanedPoolAssignment`. The release mirrors that precedent rather than adding new plumbing:

- **tier 1** the store's atomic `ReleaseIfCurrent` when it has one. The verb swaps status and assignee itself, so the metadata clear rides a second write.
- **tier 2** no usable conditional verb, so re-read live state immediately before the unconditional write.

Tier 1 is bypassed whenever that second write would be routing-consequential, because between the two writes the bead is claimable in a half-released state: either unrouted, and so invisible to both the demand query and the orphan sweep (which scans only assigned work), or still carrying a continuation group that a concurrent claim will vacuum its siblings through. Both cases take the single-write path, which lands status, assignee and metadata in one `Update`.

Reassign has no conditional verb to reach for, since `ReleaseIfCurrent` only clears, so it takes the live re-read alone.

## Two details that are load-bearing

**The verification read is a live `List`, not `Get`.** `CachingStore.Get` serves a clone from the in-memory cache for a tracked, non-dirty bead, so a cached read would confirm the caller's stale snapshot against an equally stale cache and reproduce the bug behind a guard. That check is now one function shared with the pool path instead of two copies of the same subtlety.

**A failed read returns the error rather than a verdict.** Refusing to write costs one tick; reporting a write that did not happen lets `repairStrandedPoolWorkerBead` close a session bead while its work is still assigned to it, with nothing left watching the work. That is worse than the unconditional write this replaces, which at least propagated its `Update` error.

The capability is asserted on the unwrapped `Store`. A class wrapper does not promote optional capabilities, so asserting on the wrapper would silently take tier 2 forever.

## Tests

Both directions on both store shapes, conditional verb present and absent: a stale snapshot must not clobber a re-claim, and a current snapshot must still write. Plus backend-failure propagation on every tier, both disjuncts of the single-write bypass, the cached-read trap, and clobber/positive/error for reassign.

Every guard was checked by reverting it individually and confirming its own test fails. The existing byte-identity tests now seed the backing store so their snapshots are live-accurate, and all six test fakes carry a compile-time `ConditionalAssignmentReleaser` assertion, because they embed `*beads.MemStore`, which implements the interface itself. Without the assertion a signature drift would promote MemStore's method and silently reroute every assertion onto the other tier while still reporting green.

## Test plan

- `go test ./cmd/gc/ -run 'TestReleaseWorkBead|TestReassignWorkBead|TestWorkAssignment'` passes
- `make build`, `golangci-lint run ./...` (0 issues), `go vet ./...` clean
- `go test ./internal/testpolicy/resourcecensus/ ./scripts/cipolicy/ ./internal/testenv/` passes
- Full `cmd/gc` sweep produces zero `--- FAIL` lines. The package still exits non-zero from the post-suite dolt leak guard on three pre-existing leakers (`TestInitFromWithoutHostedPreservesTemplate`, `TestOneShotOrderReadsReachTheOrdersBinding`, `TestOrderSweepTrackingReachesTheOrdersBinding`). Verified identical on clean `origin/main` at `7be7f0aad`, and none of those tests are in the five files this changes.

## Not included

`internal/api/session_resolution.go`'s `reassignOpenWorkAssignedToSession` carries the same shape. It is materially narrower, since its `List` is already `Live:true` and sits in the same function as the writes, so the window is one loop iteration rather than a whole tick. Closing it properly means hoisting the shared verification helper into `internal/beads` so both layers use one implementation, which is a design call rather than a mechanical port. Left out deliberately.

### DIFF
--- a/cmd/gc/pool_session_name.go
+++ b/cmd/gc/pool_session_name.go
@@ -626,36 +626,21 @@ func directSessionBeadIDCandidates(assignee string) []string {
 	return candidates
 }
 
-// liveWorkAssignmentStillReleasable confirms the snapshot is not stale
-// before clearing assignee. The expectedStatus must match the snapshot
-// status the caller observed: if the bead has since transitioned (e.g. a
-// concurrent claim moved open→in_progress, or another release moved
-// in_progress→open) the snapshot's release decision is no longer safe.
-// Open status is required for the issue #2793 path — graph.v2 step
-// beads stuck on a dead session's long-form assignee are status=open,
-// not in_progress.
+// liveWorkAssignmentStillReleasable confirms the snapshot is not stale before
+// clearing assignee, collapsing a read failure to "not releasable" for callers
+// that have no error channel. Open status is required for the issue #2793 path —
+// graph.v2 step beads stuck on a dead session's long-form assignee are
+// status=open, not in_progress.
+//
+// The check itself lives in liveWorkAssignmentAssigneeMatches (work_assignment.go),
+// shared with the work-release and reassign paths.
 func liveWorkAssignmentStillReleasable(store beads.Store, id, expectedStatus, assignee string) bool {
-	id = strings.TrimSpace(id)
-	expectedStatus = strings.TrimSpace(expectedStatus)
-	if store == nil || id == "" || expectedStatus == "" {
-		return false
-	}
-	work, err := store.List(beads.ListQuery{
-		Status:   expectedStatus,
-		Live:     true,
-		TierMode: beads.TierBoth,
-	})
+	matches, err := liveWorkAssignmentAssigneeMatches(store, id, expectedStatus, assignee)
 	if err != nil {
 		log.Printf("releaseOrphanedPoolAssignments: live work validation failed for %q: %v", id, err)
 		return false
 	}
-	for _, wb := range work {
-		if wb.ID != id {
-			continue
-		}
-		return strings.TrimSpace(wb.Assignee) == strings.TrimSpace(assignee)
-	}
-	return false
+	return matches
 }
 
 func assigneePreservesNamedSessionRoute(cfg *config.City, cityPath, template, assignee, workStoreRef string, storeRefAware bool) bool {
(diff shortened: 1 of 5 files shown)


sjarmak:
Merging at `8aa20d54d`, 62 checks pass and none fail.

Recording what the review actually consisted of, since the check marks alone would overstate it. Copilot reviewed `91db715ff` and found a real contradiction between two comments about whether a continuation group stamped after the snapshot can reach the second write; that is fixed. Copilot's re-review at the later head is a quota decline, not a pass. Codex reviewed `fb5cd197e` and returned request-changes with two substantive points.

Its primary remedy, fencing the write with `ConditionalWriter.UpdateIfMatch`, was rejected on evidence rather than adopted: `NativeDoltStore` never populates `Revision` (`native_dolt_store.go:2039`; the v1.1.0 schema has no version column), the bd path needs unlanded beads#4682, `beads.conditional_writes` defaults to `off`, and the verb has no production callers anywhere in the tree. `pool_session_name.go:394` already documents check-then-write as the deliberate in-tree choice for this exact situation, with the residual window named as a known limit rather than an oversight. Taking that advice would have added an unreachable code path.

Its test finding was correct and is what the final commit fixes. The two "NeverRidesASecondWrite" tests asserted a single combined write, which the pre-fix implementation also performed, so they passed against the bug they were meant to catch. They now assert the behavior only the fix produces: no write at all when the snapshot is stale. Verified by restoring the pre-fix body under the current signature, where both fail with `Update calls = 1, want 0 for a stale release snapshot`.

So: one provider found a real comment defect, one found a real test defect, both fixed, and no provider has signed off on the exact final head, which is test-only relative to the reviewed one.

PR #5138, opened 2026-08-09
fix(session): thread liveness-probe observation into heal-state instead of hardcoding Observed:true

### LINKED ISSUE #5121: Reconciler drains healthy sessions: RuntimeFacts.Observed is hardcoded true, so a failed liveness probe reads as runtime-missing
## Summary

Under fleet load, the session reconciler drains and respawns **healthy** long-lived sessions because a failed provider liveness probe is indistinguishable from a confirmed-dead one. `RuntimeFacts.Observed` — the field that exists precisely to express "we could not tell" — is hardcoded to `true` on the reconcile path, so the projection's `RuntimeProjectionUnknown` branch is unreachable there.

Because affected templates use `wake_mode: fresh`, each false positive replaces the agent with a **brand new provider session** rather than resuming it. A human driving a session loses it with no warning, and any in-flight foreground command dies.

Observed on `gc 1.4.1` (commit `111196440`, built 2026-08-02), macOS (Darwin), 16 cores running two `gc` cities.

## The code path

`cmd/gc/session_reconcile.go:889`:

```go
lcInput.Runtime = sessionpkg.RuntimeFacts{Observed: true, Alive: alive}
```

`Observed` is hardcoded `true`. `alive` arrives as `providerAlive` from `cmd/gc/session_reconciler.go:1847` — a plain `bool` with no way to signal that the probe itself failed.

`internal/session/lifecycle_projection.go` already models this correctly:

```go
func projectRuntimeProjection(input LifecycleInput, ...) (RuntimeProjection, State, bool) {
	if !input.Runtime.Observed {
		return RuntimeProjectionUnknown, compat, false
	}
	if input.Runtime.Alive {
		return RuntimeProjectionAlive, StateAwake, false
	}
	...
	return RuntimeProjectionMissing, ...
}
```

Since `Observed` is always `true` on this path, a `false` `alive` falls straight through to `RuntimeProjectionMissing`, which writes `sleep_reason = runtime-missing` at `session_reconcile.go:936`, drains the session, and respawns it fresh.

Worth noting: the same call site at `session_reconciler.go:1847` passes `rollbackAvailable` as `!storeQueryPartial`, so the reconciler **does** track partial observation for the store query. It just never tracks it for the provider liveness probe.

## Why this is not the already-fixed case

`internal/runtime/tmux/state_cache.go` handles a degraded OS process snapshot correctly and says so:

```go
if !s.ProcessesAvailable {
	// The OS process snapshot failed (e.g. the ps scan timed out under
	// fleet load). tmux confirms the session/pane is alive; we cannot verify
	// the inner process, so degrade optimistically rather than report it
	// dead. A failed secondary probe must never trigger a reap/respawn.
	return true
}
```

That landed in #3311 (2026-06-11) and the file has not changed since 2026-07-09, so this build already has it. **Upgrading does not help.** The reconcile path is a separate consumer that never learns the observation was partial.

## Evidence

From `.gc/runtime/session-reconciler-trace/`, one day, one city — 22 `runtime-missing` events across **every** long-lived agent:

```
  4  gastown__mayor
  4  gastown__deacon
  4  vessel-network--gastown__witness
  4  vessel-network--gastown__refinery
  2  polecat-opus-hq-...
  2  polecat-opus-max-hq-...
  2  polecat-opus-high-hq-...
```

A representative mayor sequence (session was healthy and actively working throughout):

```
20:12:14Z  session_baseline  state=awake
20:12:19Z  session_result    state=asleep   sleep_reason='runtime-missing'
20:12:55Z  session_baseline  state=asleep   sleep_reason='runtime-missing'
20:13:07Z  session_result    state=asleep   sleep_reason='runtime-missing'
20:14:29Z  session_baseline  state=active
20:14:38Z  session_result    state=awake
```

The replacement `claude` process started at `20:13:16Z`, inside that window. **The tmux session itself never died** — it had been up for four hours and only the process inside the pane was replaced. The same pattern fired again at `20:08:39Z`, twice inside 15 minutes.

Correlated `gc` output during the same period:

```
tmux state cache: process snapshot degraded, retaining tmux session liveness:
fetching Darwin process command snapshot: signal: killed
```

Box load was 83–128 across these events.

## Impact

- A human attached over remote control loses the session with no warning and must re-attach and resume. This happened three times in one afternoon.
- Every foreground child of the replaced process dies with `137` (SIGKILL) or `143` (SIGTERM). Our merge-queue gate runs were killed mid-suite twice and had to restart, turning a ~40 minute gate into ~2 hours on an already-slow box.
- It is self-amplifying: each respawn adds load, which makes the next process snapshot likelier to be killed.

## Suggested fix

Thread an `observed` flag out of the provider liveness probe and pass it through instead of the hardcoded `true`:

```go
lcInput.Runtime = sessionpkg.RuntimeFacts{Observed: providerObserved, Alive: alive}
```

No change to the projection is needed — `RuntimeProjectionUnknown` already exists and already means "take no action."

Optional hardening:
- Require N consecutive successful, agreeing probes before draining.
- Log loudly whenever a drain is triggered by a probe that did not complete cleanly.

rjgeng:
## Summary

Fixes #5121: the reconciler's heal-state path hardcoded `RuntimeFacts.Observed: true`, so a failed liveness probe (couldn't tell if a session was alive) was treated identically to a *confirmed*-dead runtime. Every other destructive decision in the same reconciler pass already fails closed on a liveness-observation error (pending-create rollback, failed-create close, drain-ack finalize, close) — the heal-state call site at `session_reconciler.go:1847` was the one path that dropped the error signal, hardcoding `Observed: true` regardless of whether the probe actually succeeded.

Under fleet load, a transient probe failure would drain and respawn a healthy long-lived session — since the affected templates use `wake_mode: fresh`, the reconciler replaced rather than resumed, killing any in-flight foreground work. The issue's own trace shows 22 false-positive respawns in one day across every long-lived agent.

## What changed

- `healStatePatchWithRollbackInfo` / `healStateWithRollbackInfo` now take an `observed bool` parameter, threaded through as `livenessErr == nil` at the reported call site.
- `RuntimeFacts.Observed` and the existing `RuntimeProjectionUnknown` branch ("take no action") in `internal/session/lifecycle_projection.go` already handled this case correctly — nothing there needed to change, only the caller that never set it accurately.
- A second call site for the same function (`session_reconciler.go:2623`) is fed by an older probe (`observeRuntimeProviderLiveness`) that returns no error signal at all (`(running, alive bool)`, no `error`). Left at `observed: true` to preserve its exact current behavior — extending it isn't evidenced by #5121's trace/line citations and would need its own investigation into whether that probe should grow an error return.

## Testing

- New pinned regression case (`unobserved-liveness-preserves-active` in `cmd/gc/session_r3_info_equiv_test.go`): alive=false + unobserved=true on an active session now returns a nil batch (no-op). Before the fix, this produced the same `{state: asleep, sleep_reason: runtime-missing}` drain as a genuinely-dead session.
- Full `cmd/gc` build + `go vet ./...` clean.
- Targeted heal-state/reconciler test suite: 100% pass, no regressions.
- Full 6-way sharded `cmd/gc` run: 9 failures surfaced across 4 of 6 shards; all 9 reproduced in isolation — 7 confirmed as parallel-shard resource contention (pass standalone), the remaining 2 (`TestResolveImportRootFallsBackToStandalonePackDir`/`TestResolveImportRootPrefersNearestPackUnderCity`, a macOS `/var/folders` vs `/private/var/folders` symlink artifact; and `TestRunStartDriftCheck_DelegatedTryRestartTimeoutThenReplacementSucceeds`, a missing `GC_SUPERVISOR_SYSTEMD_UNIT` env var) reproduce identically on clean `main` before this change.
- Base commit `0b73be294`, confirmed no upstream changes to the touched files since (`cmd/gc/session_reconcile.go`, `cmd/gc/session_reconciler.go`, `cmd/gc/session_reconcile_test.go`, `cmd/gc/session_r3_info_equiv_test.go`) via `git log 0b73be294..origin/main -- <files>` (empty).

---

### DIFF
--- a/cmd/gc/session_r3_info_equiv_test.go
+++ b/cmd/gc/session_r3_info_equiv_test.go
@@ -28,14 +28,19 @@ func (s setMetadataBatchFailStore) SetMetadataBatch(string, map[string]string) e
 // healOracleCase is a heal fixture plus the runtime/clock/lease knobs the heal
 // patch reads and the exact batch the Info form must return.
 type healOracleCase struct {
-	name     string
-	status   string
-	created  time.Duration // relative to clk.Now(); 0 = zero time
-	meta     map[string]string
-	alive    bool
-	timeout  time.Duration
-	rollback bool
-	want     map[string]string
+	name    string
+	status  string
+	created time.Duration // relative to clk.Now(); 0 = zero time
+	meta    map[string]string
+	alive   bool
+	// unobserved marks the liveness probe itself as failed (RuntimeFacts.Observed
+	// = false), distinct from alive=false (probe succeeded, runtime confirmed
+	// dead). Zero value (false) preserves every existing case's prior hardcoded-
+	// Observed:true behavior unchanged.
+	unobserved bool
+	timeout    time.Duration
+	rollback   bool
+	want       map[string]string
 }
 
 // TestHealStatePatchWithRollbackInfo pins healStatePatchWithRollbackInfo against
@@ -140,6 +145,20 @@ func TestHealStatePatchWithRollbackInfo(t *testing.T) {
 			rollback: true,
 			want:     map[string]string{"continuation_reset_pending": "true", "primed_at": "", "priming_attempted_at": "", "prompt_hash": "", "session_key": "", "sleep_reason": "runtime-missing", "started_config_hash": "", "state": "asleep"},
 		},
+		{
+			// gascity#5121: a failed liveness probe (Observed=false) must not be
+			// treated as a confirmed-dead runtime. Before the fix this produced
+			// the same drain batch as a genuinely observed-dead alive=false case
+			// ({"sleep_reason": "runtime-missing", "state": "asleep"}), silently
+			// respawning a healthy long-lived session on every transient probe
+			// failure.
+			name:       "unobserved-liveness-preserves-active",
+			meta:       map[string]string{"state": "active"},
+			alive:      false,
+			unobserved: true,
+			rollback:   true,
+			want:       nil,
+		},
 	}
 
 	for _, tc := range cases {
@@ -152,7 +171,7 @@ func TestHealStatePatchWithRollbackInfo(t *testing.T) {
 			if tc.created != 0 {
 				b.CreatedAt = clk.Now().Add(tc.created)
 			}
-			got := healStatePatchWithRollbackInfo(seedSessionInfo(b), tc.alive, clk, tc.timeout, tc.rollback)
+			got := healStatePatchWithRollbackInfo(seedSessionInfo(b), tc.alive, !tc.unobserved, clk, tc.timeout, tc.rollback)
 			if !reflect.DeepEqual(got, tc.want) {
 				t.Fatalf("healStatePatchWithRollbackInfo = %#v, want %#v", got, tc.want)
 			}
@@ -175,7 +194,7 @@ func TestHealStateWithRollbackInfoClosedGuardAndWrite(t *testing.T) {
 		t.Fatal(err)
 	}
 	closed, _ = store.Get(closed.ID)
-	batch, err := healStateWithRollbackInfo(sessiontest.SeedBead(t, closed), false, sessionFrontDoor(store), clk, 0, true)
+	batch, err := healStateWithRollbackInfo(sessiontest.SeedBead(t, closed), false, true, sessionFrontDoor(store), clk, 0, true)
 	if err != nil {
 		t.Fatalf("heal closed bead: %v", err)
 	}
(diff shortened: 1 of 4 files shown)


quad341 (approved):
@rjgeng — this is a really clean piece of surgery. Threading `observed` through instead of patching around it at the projection layer was the right instinct: `RuntimeProjectionUnknown` was already sitting there doing exactly the correct thing, and the bug was purely that one caller never told it the truth.

I traced it on my side rather than taking the description on faith, and the diagnosis holds precisely — line 1660 sets `providerAlive = false` on a probe error and then *falls through*, unlike every sibling destructive site, which all `continue`. So the heal really was the one leak in an otherwise consistent fail-closed pattern.

Two things I especially appreciated. You left the 2623 call site alone and said why — `observeRuntimeProviderLiveness` has no error to thread, and quietly extending that probe inside a p1 fix would have been a much riskier change wearing a small diff. And the `unobserved` field's zero value keeps all thirteen existing oracle rows byte-identical, so the pinned table stays a real regression oracle instead of being loosened around the new case.

Merging as-is. I'll open a follow-up for giving `observeRuntimeProviderLiveness` an error return so the desired-session fast path can fail closed the same way — yours if you want it, you clearly have the map.


---

rjgeng:
Thanks for the quick review and merge here — appreciate the consistent turnaround across our recent fixes.

PR #5139, opened 2026-08-09
refactor(beads): lift the by-id class resolver out of internal/api (ga-ia7li)

julianknutsen:
## What

`internal/api`'s `beadStoresForID` carried the only by-id class resolver in the tree: a class-prefix arm, method-bound to the API server, that returns a candidate **probe list** for a `gcg-` id on a graph-relocated city. This lifts that arm into `storeref.ClassCandidates` and makes `internal/api` consume it.

`gc bd` by-id routing, `gc ready`, claim routing and the work-query federation all need the same answer. Answering it a second time — a second place that decides "which store owns this class of bead?" — is precisely how this repo's split-store bug class reproduces (#5125, #5127).

Behaviour-preserving apart from the two carried minors below, which the bead requires closing.

## CORRECTION: this resolver does NOT survive a migrated city

An earlier revision of this PR, and of the resolver's own doc, said the candidate-list shape is what lets the arm survive a **migrated** city. That was false, and it is corrected at the source rather than softened.

`IDInNamespace(id, prefix)` gates **before** the candidate list is built. `gc storage migrate` preserves ids (`infra_class_migrate.go`), so a bead it relocated keeps its HQ/rig-era prefix — which is *outside* the class namespace. The arm returns **nil candidates** for exactly the ids the claim said it covered. Stated plainly: **a legacy-prefixed relocated bead is out of this resolver's namespace and is not covered by it.**

Pinned, so the limitation cannot drift:

- `TestClassCandidatesMigratedLegacyIDIsOutOfNamespace` — `ClassCandidates("mc-123", …)` is nil however thoroughly graph is relocated.
- `TestBeadStoresForIDDoesNotCoverMigratedLegacyIDs` — the handler-side consequence: the id resolves to the **work** store, and the relocated store is not even a candidate.

**What does cover a migrated id is the other approach.** `cmd/gc/cmd_bd_by_id.go`'s `bdByIDClassDoor.resolve` probes RESIDENCE: it asks the class front door about *every* id, prefixed or not, and its own doc gives this exact reason — "`gc storage migrate` copies the work store's infrastructure slice with its ids PRESERVED, so a converged city holds work-shaped ids in its binding and deciding ownership by prefix alone would send those reads back to the ledger they were moved off."

The `internal/api` by-id path has **no** equivalent. There a legacy-prefixed relocated bead is answered from the work store's retained pre-migration copy (the migration never deletes its source, so the row is present but frozen at migration time). That is **open work**, named in the code and here rather than left for a reader to find.

The candidate-list shape does have a real virtue — a different one. It **probes** instead of routing unconditionally: the class store leads because it is the sole *minter* of the reserved namespace, but minting is not holding. A reserved prefix is only an advisory on work stores (`config.ReservedPrefixWarnings` warns; `config.ValidateRigs` does not reject), so a work store can legitimately hold an in-namespace id, and a prefix route would report those beads as absent.

## The contract

```go
// internal/storeref/class_candidates.go
func ClassCandidates(id string, routing ClassRouting) []beads.Store
```

| result | when |
|---|---|
| `nil` | `routing.Class` is nil or IS `routing.Work` (class not relocated) |
| `nil` | `id` is outside the class namespace — including every migrated legacy id |
| `[class, work, …shadows…]` | otherwise — class first, then work, then shadowing work stores most-specific-prefix first; deduped, nils dropped |

The routing is gated on `resolveClassStore` **identity** — `Class != nil && Class != Work`, the same question `cmd/gc`'s `resolveClassStore` answers — never on a marker file or a migration flag. A single-store city gets `nil` back and keeps its legacy resolution byte-identical.

`ClassRouting{Prefix, Class, Work, Shadows}` takes the reserved prefix as **data**, so the package stays free of `internal/config` (which imports `internal/beads` and could not be imported back), and adding a class later is a call-site change rather than a resolver change.

Deliberately *not* the branch's `storeForID`, which scans every argument and routes the entire command on the first reserved-looking token with no existence probe (so `gc bd dep add ga-123 gcg-456 blocks` routes to the store that does not hold the subject).

## Why the shadowing stores go LAST

Council finding, evaluated rather than taken at face value: *"a shadowing work store inserted mid-chain lets an unrelated rig outage 500 a read main served."* Verified against the probe loop — it is real, and narrow. The by-id handlers stop at the first store that answers and surface a non-`ErrNotFound` read as a 500. So with the shadows inserted between the class store and the work store, a hard-failing rig store aborted the walk before the work store that held the bead.

`[class, work]` is exactly the pre-seam list, and it is now the **head** of the new one — the shadows are appended behind it. Every probe the old list made still happens, in the same order, and the added legs are reached only where the old list had already given up. A store added here can therefore neither change an answer resolution already served nor turn a served read into an error. The worst a broken shadow can do is turn a not-found into a hard error, which is the honest report when a store that could hold the id is unreachable.

Not taken: relaxing the handlers' fail-fast rule to "skip a failing candidate and continue". That would let a mutation land on a *different* store's row of the same id whenever the leading class store is unreachable — the residence-violation class this whole program exists to prevent — and it would change legacy multi-rig behaviour on the unrelated fallback scan. Ordering removes the exposure without either cost.

Red before, through the real resolver and the real handler:

```
GET bead gcg-1: rig backend unreachable — an unrelated rig's outage answered for a bead the city store holds
candidates = [class rig work], want the pre-seam [class work] head first
```

## Two namespace predicates, and the invariant that stopped being true

`storeref`'s package doc used to state that bead ids are prefix-disjoint across stores because `config.ValidateRigs` keeps reserved class prefixes off work-store prefixes. It does not: `ValidateRigs` rejects duplicate prefixes and nothing else, and deliberately allows a reserved class prefix (`config.ReservedPrefixWarnings` warns). This PR's shadowing support is the proof, so the package doc now says ids are **not** namespace-disjoint and that every resolver here probes rather than assumes.

The tree now has two namespace predicates with different rules, and the package doc, `PrefixOwner` and `IDInNamespace` all cross-reference each other and say which to use:

- `IDInNamespace` — the prefix came from **config** (rig/HQ prefix, or `config.ReservedClassPrefix`). Exact-or-hyphen: admits the bare id, because a configured prefix can be a whole id. Moved out of `internal/api`'s `beadIDHasConfiguredPrefix`, which this PR deletes.
- `PrefixOwner` — the prefix came from the **store** (its self-declared `IDPrefix()`). Separator-only: a store that mints `gcg-1` never mints the bare `gcg`. Widening it would let a bare id capture a store that cannot hold it.

The conformance suite's I5 cited the deleted `beadIDHasConfiguredPrefix` and asserted the two rules were the same; it now names `storeref.IDInNamespace` and states the divergence.

## The two carried minors from #5128's council

Reserved class prefixes are warned-and-**allowed** on work stores (`config.ReservedPrefixWarnings`); `config.ValidateRigs` does not reject them — verified, not assumed. So a work store can legitimately hold ids inside the class namespace, and the hoisted arm dropped it from the candidate set:

- **(a)** a rig configured with the reserved prefix itself lost by-id reachability for every one of its beads the moment graph relocated.
- **(b)** an id under a **longer** prefix that starts with the reserved one (`gcg-alpha-1` under rig prefix `gcg-alpha`) silently lost its rig store.

Neither was tested. Red before, through the real resolver and the real handler:

```
beadStoresForID(gcg-1) = [graph city]; shadowing rig store is NOT a candidate
beadStoresForID(gcg-alpha-1) = [graph city]; rig store under the LONGER prefix "gcg-alpha" is NOT a candidate
GET bead gcg-1: bead gcg-1 not found — a shadowing rig's bead is unreachable by id on a relocated city
```

The list now carries the shadowing work stores **behind** the class store and the city store. #5128's rule that a reserved prefix can never be captured by a rig/HQ prefix or a `routes.jsonl` entry is unchanged — this is about reachability, not precedence.

## Graph only — what the other four prefixes turned out to be

`gcs` / `gco` / `gcn` **do** have relocated-store accessors on `api.State` today (`SessionsBeadStore`, `OrdersBeadStore`, `NudgesBeadStore`, all wired through `resolveClassStore` in `cmd/gc/api_state.go`). `gcm` has **none** — mail routes through `mail.Provider`, and there is no `MessagingBeadStore()` anywhere.

Generalising to them is still out of scope: it would change which stores answer for `gcs-`/`gco-`/`gcn-` ids, which is a production routing change and not part of lifting this arm. The call site stays graph-only and says so.

## I5 un-skipped, and its gap pinned

`TestSplitTopologyConformance/I5-claim-routing`'s claim-mutation half skipped naming this bead — there was no seam that routed a claim **mutation** by class. It now runs on both topologies through `splitEnv.claimByID`: candidates from the shared resolver, probed in order, mutation written through the store that answered, asserting **residence** (visible in the owning store, in no other leg).

Two fixes to that invariant, from the council:

- **The single-store leg now pins the IDENTITY gate.** It asserted the resolver is inert using the city's own work ids — which are outside the reserved namespace, so the NAMESPACE gate alone rejected them and the row still passed with the identity gate deleted. It now also routes a **class-shaped** id, which clears the namespace gate so identity is the only thing that can reject it. Red under mutation: `class-shaped id — identity gate: classCandidatesForID("gcg-1") returned 1 candidates on a single-store city`.
- **The KNOWN GAP is now asserted, not just described.** Every other KNOWN GAP in the suite (I1, I2, I10) pins today's answer through production code; I5's did not. It now asserts through `hookQueryEnv` — the seam that decides which store `gc hook --claim` runs its claim against — that the answer is a WORK scope rooted at the city path, identically on **both** topologies. When ga-xo8ch rewires the command onto the shared resolver, that row is what moves.

`gc bd update --claim` is already routed (#5132, `cmd_bd_by_id.go`) and probes the same way. The suite's skip list is now empty; the header says so, and says a new gap gets a skip rather than a deletion.

## Teeth, proved by mutation not argument

| mutation | result |
|---|---|
| drop the shadow arm | both minors redden **and** the handler 404s on a shadowing rig's bead |
| move the shadows ahead of the work store | `TestBeadGetSurvivesAShadowingRigOutage` reddens — a rig outage 500s a read the city store was serving |
| drop the `IDInNamespace` gate | `TestClassCandidatesMigratedLegacyIDIsOutOfNamespace` reddens with `ClassCandidates(mc-123) = [class work]` |
| drop the `Class != Work` identity gate | `TestBeadStoresForIDShadowingRigPrefixStillWinsOnDefaultCity` **and** I5's single-store class-shaped row redden — single-store byte-identity is exactly what that gate protects |
| drop the class candidates from `claimByID` | I5's split leg fails: `no candidate store held gcg-1: the by-id claim had nowhere to run` |

## Gates

`go build ./...` · `go vet ./...` · `gofmt -l` clean · `golangci-lint run` (2.10.1) **0 issues** · `go test ./internal/api/ ./internal/beads/... ./internal/storeref/` · `go test ./cmd/gc -timeout 25m` · `scripts/check-core-boundary.sh` · `make check-split-topology-rows`

Precondition #5128 (`dec245e181`) merged; built on `internal/beads/splittest` (#5130) and the conformance suite (#5134). Rebased onto current `main` (through #5140).

Closes nothing on its own — it unblocks ga-4k2c3, ga-xo8ch and ga-wnjr4.

### DIFF
--- a/cmd/gc/split_topology_conformance_test.go
+++ b/cmd/gc/split_topology_conformance_test.go
@@ -43,15 +43,18 @@ import (
 // Some invariants name a capability main does not have yet. Those SKIP with the
 // reason and the seam that is missing, rather than being quietly omitted or
 // asserted against a seam that does not exist. The skip list is this program's
-// verified remaining work.
+// verified remaining work, and it is currently EMPTY: I5's claim-mutation half
+// was the last entry, and it closed when the shared by-id class resolver landed
+// (storeref.ClassCandidates, ga-ia7li). A new gap gets a skip, not a deletion.
 //
 // Others pin behavior main HAS but should not keep. Those carry a KNOWN GAP
 // paragraph naming the divergence, the assertions that move when it closes, and
 // the slice that closes it — I1 and I2 (the HQ work store is in neither arm of
-// the controller's cross-store scan on a split city) and I10 (the wake filter
-// has no coordination-class reachability arm). Leaving such a leg UNSEEDED is
-// the failure mode this convention exists to prevent: the invariant then reads
-// as coverage of a path it never touches.
+// the controller's cross-store scan on a split city), I5 (`gc hook --claim`
+// does not consume the shared resolver yet) and I10 (the wake filter has no
+// coordination-class reachability arm). Leaving such a leg UNSEEDED is the
+// failure mode this convention exists to prevent: the invariant then reads as
+// coverage of a path it never touches.
 //
 // # Which authority an invariant is pinning
 //
@@ -391,14 +394,27 @@ func conformanceMaterializationResidence(t *testing.T, e splitEnv) {
 // asserted as a negative here so it cannot change silently under a future
 // resolver.
 //
-// The CLAIM-MUTATION half is SKIPPED — the seam does not exist on main.
-// `gc hook --claim` resolves its stores as hookStore{dir, env} pairs
-// (hook_cross_store.go) and execs bd in a work directory; the fan-out is city +
-// rigs, all of them WORK scopes, with no coordination-class arm anywhere on the
-// path. The only by-id class resolver in the tree is Server.beadStoresForID
-// (internal/api/handler_beads.go), method-bound to the API server and graph-only.
-// Lifting it into a shared resolver is the next slice in this program (ga-ia7li,
-// which this work blocks).
+// The CLAIM-MUTATION half runs on the SHARED resolver, storeref.ClassCandidates
+// (ga-ia7li), which is the seam this invariant was waiting on: a candidate list
+// keyed on the resolveClassStore identity, probed in order, with the mutation
+// written through the store that answered. splitEnv.claimByID is that shape, and
+// the assertion is residence — the claim is visible in the owning store and in
+// no other leg.
+//
+// KNOWN GAP, pinned rather than asserted as desirable — `gc hook --claim` does
+// not consume the resolver yet. It resolves its stores as hookStore{dir, env}
+// pairs (hook_cross_store.go) and execs bd in a work directory; the fan-out is
+// city + rigs, all WORK scopes, so a claim it issues for a relocated class id
+// still runs against a ledger that cannot see the bead. claimByID's own
+// fallback IS that fan-out, so the rows below state both answers side by side:
+// a class id routes, a work id keeps the legacy path byte-for-byte. Rewiring
+// the command is ga-xo8ch (class-routed writes from one-shot commands); when it
+// lands, this paragraph goes and the assertions move from claimByID to the
+// command.
+//
+// `gc bd update --claim` is already routed (#5132, cmd_bd_by_id.go) and probes
+// the same way, so the resolver is not the only class-aware claim path — it is
+// the one a caller holding just an id and a set of stores can use.
 func conformanceClaimRouting(t *testing.T, e splitEnv) {
 	workBead, err := e.work.Create(beads.Bead{Title: "claim-routing work bead", Type: "task"})
 	if err != nil {
@@ -426,7 +442,24 @@ func conformanceClaimRouting(t *testing.T, e splitEnv) {
 		if reservedClassNamespace(wisp.ID) {
 			t.Fatalf("single-store wisp %q sits in a reserved class namespace; a legacy city mints work-store ids", wisp.ID)
 		}
-		t.Skip("single-store collapse: there is no second store to route a claim to, and no coordination-class claim arm to route it with (ga-ia7li)")
+		// The single-store statement about the shared resolver: its identity
+		// gate (class store IS the work store) means it claims NOTHING here, so
+		// every by-id claim keeps running on the one store exactly as it does
+		// today. A resolver that answered on a legacy city would be routing a
+		// city with nowhere to route to.
+		for _, id := range []string{workBead.ID, wisp.ID} {
+			if got := e.classCandidatesForID(id); got != nil {
+				t.Fatalf("classCandidatesForID(%q) returned %d candidates on a single-store city; the resolver must be inert where the class store IS the work store", id, len(got))
+			}
+		}
+		for _, id := range []string{workBead.ID, wisp.ID} {
+			landed := e.claimByID(t, id, "single-store-claimant")
+			if !sameStorePtr(landed, e.work) {
+				t.Errorf("claim of %s landed in %p, want the single work store %p", id, landed, e.work)
+			}
+			e.assertClaimedIn(t, id, "single-store-claimant", e.work)
+		}
+		return
 	}
 
 	durable := mintDurableGraphBead(t, e, "claim-routing durable control bead", "")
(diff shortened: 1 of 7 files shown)


julianknutsen:
Council review fixes pushed (`ae02aa452e`), rebased onto current `main` (through #5140). Zero actionable findings; the four confirmed minors are all addressed, plus the one finding flagged for evaluation.

**M1 — the migrated-city claim was false, and it is corrected at the source rather than softened.** `IDInNamespace` gates *before* the candidate list is built, so a relocated bead carrying a legacy work-era id — precisely the ids `infra_class_migrate.go` preserves — gets **nil candidates**. The docs now (a) state the shape's real virtue, that it PROBES rather than routing unconditionally, (b) say plainly that a legacy-prefixed relocated bead is out of this resolver's namespace and not covered, and (c) name the mechanism that *does* cover migrated ids: `cmd/gc/cmd_bd_by_id.go`'s `bdByIDClassDoor.resolve`, which probes residence for **every** id and documents that same reason. `internal/api` has no equivalent, and that is called out as open work rather than left for a reader to discover. Pinned by `TestClassCandidatesMigratedLegacyIDIsOutOfNamespace` (`ClassCandidates("mc-123") = nil`) and `TestBeadStoresForIDDoesNotCoverMigratedLegacyIDs` (the id resolves to the **work** store; the relocated store is not a candidate). Red under mutation, dropping the namespace gate: `ClassCandidates(mc-123) = [class work], want nil`.

**Shadow-outage finding: verified real, and fixed by ordering rather than by relaxing fail-fast.** The council read the probe loop correctly. `internal/api/huma_handlers_beads.go` returns `apierr.Internal` on the first non-`ErrNotFound` probe, so a shadow inserted mid-chain aborts the walk before the work store. Scope of the regression: a read served from the **class** store is untouched (it is probed first), and a bead reachable only through a shadow was 404 on main. The one shape that regressed is a bead resident in the trailing work store while a mid-chain shadow hard-errors — narrow, but real.

Fixed by appending the shadows **last**: `[class, work, …shadows…]`. `[class, work]` was the whole pre-seam list and is now a strict prefix of the new one, so every probe the old list made still happens in the same order and the added legs are reached only where it had already given up. A store added here therefore cannot change an answer resolution already served, and cannot fail ahead of the store holding the bead. Red before, through the real handler: `GET bead gcg-1: rig backend unreachable — an unrelated rig's outage answered for a bead the city store holds`.

Not taken: "skip a failing candidate and continue". Under a shadow configuration the *same* id can legitimately exist in two stores (both mint in the namespace), so skipping an unreachable class store would apply a mutation to a different store's row of the same id — the residence-violation class this program exists to prevent — and it would change legacy multi-rig behaviour on the unrelated fallback scan. Ordering removes the exposure without either cost. The fail-fast rule and why the order protects it are now documented on `beadStoresForID`.

**M2** — the conformance suite's I5 cited `internal/api`'s `beadIDHasConfiguredPrefix`, which this PR deletes, and asserted the two rules were the same. It now names `storeref.IDInNamespace` and states the divergence (they agree on every id shape I5 routes and differ on the bare form). No reference to the deleted symbol remains in the tree.

**M3** — `storeref`'s package doc claimed ids are prefix-disjoint "because `config.ValidateRigs` keeps reserved class prefixes off work-store prefixes". Verified against the source: `ValidateRigs` rejects duplicate prefixes and nothing else, and its own comment says a reserved class prefix is deliberately allowed. The package doc now states ids are **not** namespace-disjoint and that every resolver here probes; the package doc, `PrefixOwner` and `IDInNamespace` cross-reference each other and say which predicate to use for a config-sourced vs a store-declared prefix. `configuredPrefixStores` no longer claims parity with a walk the pre-seam code did not perform (it looked a rig's store up only when that rig's prefix already covered the id and beat the best match) — the comment now states what it actually does and what it costs.

**M4** — (a) I5's single-store leg claimed to pin the identity gate but was satisfied by the namespace gate alone: the city's own ids are outside the reserved namespace, so the row still passed with the identity gate deleted. It now also routes a class-shaped id, which clears the namespace gate so only identity can reject it. Red under mutation: `class-shaped id — identity gate: classCandidatesForID("gcg-1") returned 1 candidates on a single-store city`. (b) I5's KNOWN GAP is now asserted the way I1/I2/I10 assert theirs — through `hookQueryEnv`, the production seam that decides which store `gc hook --claim` runs its claim against — pinning a WORK scope rooted at the city path on **both** topologies. When ga-xo8ch lands, that row is what moves.

Gates: `go build ./...` · `go vet ./...` · `gofmt -l` clean · `golangci-lint run` (2.10.1) **0 issues** · `go test ./internal/api/ ./internal/storeref/ ./internal/beads/...` · `go test ./cmd/gc -timeout 25m` (800s, ok) · `scripts/check-core-boundary.sh` · `scripts/check-split-topology-rows.sh`.

PR #5144, opened 2026-08-09
refactor(beads): make the opaque storage binding the only non-dolt shape (ga-pthgr)

julianknutsen:
Implements **ga-pthgr** — E1' of the postgres-extraction program.

## The observation this slice is built on

gc already had a fully generic path for a store it does not serve.
`scopeHasCompleteStorageBinding` recognizes an *opaque binding* —
`backend`, `storage_endpoint` and `storage_database` all non-empty, no
backend name examined — and
`applyCompleteNonDoltStorageBindingEnv` answers it by calling
`setExecProjectedBackendEnvEmpty`: it withholds **every** projected
backend variable, sets `BD_BIN`, preserves a `BEADS_CREDENTIALS_FILE`
the operator pointed at, and stops. bd reads the workspace's own
configuration and runs its own credential ladder.

`internal/storebinding/beadsworkspace` already states the principle:
withhold the *namespace*, not a key list, because "the list of variables
the linked library reads is the library's to grow."

Make that the only non-dolt path and the postgres vocabulary does not
need to be registered, adapted, or handed to a registrar. It evaporates.

## What is gone

Whole files:

- `internal/pgauth/` — the seven-tier credential resolver, its
  permission and parse guards, its event payload, and its census. bd
  owns credential resolution for a backend bd serves.
- `internal/beads/contract/postgres_dsn.go` — DSN parsing, endpoint
  derivation, completeness and port-range validation. bd parses its own
  DSN.
- `internal/doctor/checks/` — the postgres-auth check was the package's
  only member.
- The 15 postgres metadata fixtures.

Vocabulary:

- `MetadataState`'s `postgres_host/port/user/database/dsn/schema` field
  family and every validation rung built on it. The E1–E5 rejection
  ladder is now E1 (JSON parse) → E2 (unknown backend); the rungs that
  followed E2 all validated a connection shape for a backend gc no
  longer implements. `mixedBackendField` had nothing left to compare and
  is deleted with them.
- `projectedPostgresEnvKeys` and the strip/clear/explicit symmetry that
  existed only to keep it in sync: the `mergeRuntimeEnv` strip list, the
  `execProjectedBackendEnvKeys` segment, `clearProjectedPostgresEnv`,
  `ensureProjectedPostgresEnvExplicit`, `mirrorBeadsPostgresEnv`, and
  the `[order.env]` reserved keys in `internal/orders/env.go`.
- Every `case "postgres"` arm in `cmd/gc/bd_env.go` and
  `cmd/gc/beads_provider_lifecycle.go`, plus `postgresMetadataForScope`,
  `scopeBackendIsPostgres`, `applyResolvedScopePostgresEnv`,
  `emitPostgresCredentialResolved`, `cityPostgresProjectionErrorCanBeBypassed`
  and the provider-lifecycle bypass that was reachable only from it.
- `--explain-postgres-auth` and `doctor.CheckContext.ExplainPostgresAuth`.

## One predicate, every caller

Several guards asked "is this scope postgres?" when the question they
actually needed answered was "is this scope's store gc's to manage?".
`scopeBindsPinnedBdBinary` — which already answered the second question,
including the inherited-city case — is renamed `scopeStoreIsExternallyBound`
and is now the single predicate behind which bd binary to run, whether to
project a Dolt environment for an exec order, whether gc owns a managed
Dolt runtime, whether gc may **recover** one, whether a scope needs a
local Dolt identity, and whether the projection may strip password
mirrors from its own resolution input.

Two side effects that hid behind the old postgres guards are now guarded
explicitly rather than incidentally: `currentDoltPort` deletes the raw-bd
port mirror when it cannot resolve a port, so it is no longer asked about
a bound city; and `managedDoltLifecycleOwned` loads the city metadata so a
refusal still reaches the operator rather than being answered as "not
owned".

`crossBackendKeysToScrub` now scrubs one key — `dolt_mode` on a doltlite
scope — because that is the only key gc both writes and knows a backend
it implements does not read. A key belonging to a backend gc does not
implement is left alone: gc cannot tell an inert leftover from live
configuration, and a scope bound to such a backend never reaches the
function at all.

## Review round: the fifth site, and three more

Council review found a site the first pass missed, plus three smaller
regressions. All four are fixed in `8e6d9e9f3f`, each with a test that
fails without the fix.

**The fifth site.** `bdCommandRunnerWithManagedRetryErr` lost both
postgres guards with nothing put in their place, and the surviving gate
chain cannot see a storage binding. For a bound scope
`canonicalScopeDoltTarget` reports no managed runtime and falls through
to `managedLocalDoltEnv`, which reads the withheld projection's empty
`GC_DOLT_HOST` as managed-local Dolt — a withheld projection is
indistinguishable from managed-local. `managedBDRecoveryAllowed`
therefore answered **true** for exactly the scopes gc must not manage.

Measured with the production runners against an opaque binding and a
transport-shaped bd failure (`dial tcp …: connection refused`),
attempts/recoverCalls:

| runner | before | after |
| --- | --- | --- |
| `controlBdCommandRunnerForCity` | 2 / 1 | **1 / 0** |
| `bdCommandRunnerForRig` (inherited, and own binding) | 2 / 1 | **1 / 0** |
| `controlBdCommandRunnerForRig` (inherited, and own binding) | 2 / 1 | **1 / 0** |
| `bdCommandRunnerForCity` | 1 / 0 | 1 / 0 (short-circuits to `bdContextCommandRunnerForCity`) |

Identical for backend `postgres`, `acme` and `mysql`, so this was the
generic opaque-binding shape rather than a postgres artifact. The
recovery was real work, not a no-op: it execs the provider `recover` op,
whose `is_remote()` is false precisely because the projection is blanked,
so a transient outage of a bound city's external store started a managed
Dolt sql-server — data dir, port, pid — for a city gc does not own, on
top of doubling bd invocations against a backend already down.

The guard goes into `managedBDRecoveryAllowed` rather than into the
runner, so that function genuinely becomes the one place answering "may
gc recover this store?" for both the retry and the recovery classifier,
and the bd error reaches the operator unwrapped. (The alternative — an
early return in the runner preserving an operator-facing message — would
have restored main's `classifying scope backend (bd error: …)` wrap,
which only ever fired from a metadata-parse failure and was worse
reading than the transport error itself.) An unreadable binding answers
"no recovery" for the same reason main's classify-error branch did.

**Inherited-rig session projection.** `applyCanonicalScopeBackendEnv`'s
inherited-city branch was narrowed from `applyCityStorageBindingEnv` to
`applyCompleteNonDoltStorageBindingEnv`, dropping the
unregistered-backend refusal from the env every rig-scoped agent session
inherits. A rig under a city whose backend this build cannot project
returned `err = nil` with the city's leftover `GC_DOLT_HOST` /
`BEADS_DOLT_SERVER_HOST` set — an agent session pointed at a Dolt
endpoint for a city gc had just refused to classify. It calls
`applyCityStorageBindingEnv` again, and the "second projection entry
point" comment on that test is true once more.

**`canonicalScopeDoltProjectionAuthoritative`.** It lost the same guard.
A bound city gets no projection at all, so stripping its
`GC_DOLT_PASSWORD` / `BEADS_DOLT_PASSWORD` mirrors from
`mergeCanonicalScopeDoltEnv`'s resolution input removed auth that nothing
downstream restores.

**A vacuous test.** `TestBdCommandRunnerForCityCompleteStorageBindingSkipsManagedRetry`
could not fail: its fake bd emitted no substring
`bdTransportRetryableError` matches, so the retry branch was unreachable
regardless of guarding, and it asserted on error text no production path
can emit. Its shim now fails the way an unreachable external store fails
and it counts bd invocations and recovery calls. Disabling the
`bdContextCommandRunnerForCity` short-circuit it is named for — the
mutation the old test survived — now fails it.

Minors in the same commit: `scopeMetadataJSONPath` has its doc comment
back (`scopeStoreIsExternallyBoundBestEffort` had been inserted between
them, and now sits beside its own sibling); the metadata contract no
longer advertises the deleted E3–E5 rungs; and
`BackendCredentialResolved`'s cited redaction invariant,
`TestBackendCredentialResolvedPayloadOmitsTheCredential`, exists again —
closing the field set of the surviving wire type, since the resolver it
used to canary through now lives in the linked library.

No sixth site: every base call of `scopeBackendIsPostgres` /
`postgresMetadataForScope` was walked. Beyond the four migrated and the
two fixed here, `cityPostgresProjectionErrorCanBeBypassed` and its caller
are deleted (head now always returns `cityErr` — strictly stricter),
`doctorWorkspaceHasPostgresScope` went with `PostgresAuthCheck`, and
`scopeSkipsManagedDoltForInit`, `syncConfiguredDoltPortFiles`,
`managedDoltLifecycleOwned` and `clearManagedDoltRuntimeStateUnlessBound`
each answer the binding question directly. `managedBDRecoveryAllowed` has
exactly one production caller, `bdTransportErrorMatches`, which feeds
both classifiers, which are called only from
`bdCommandRunnerWithManagedRetryErr`.

## What is kept, and why

**The backend-name registry.** Invariant 0 is untouched: OSS still
refuses an unknown backend loudly and by name, on metadata load and on
both env-projection entry points, enumerating what this build registers.
The compiled set is now `""`, `dolt`, `doltlite`; `postgres` becomes an
assembly's registration. `TestCompiledBackendBundleStillCarriesPostgres`
pinned the opposite while E0 was in flight — it is replaced by
`TestCompiledBackendBundleCarriesOnlyTheBackendsGCImplements`, which
asserts the exact set and that `RecognizeBackend("postgres")` refuses.
That reversal is deliberate and is the one behavior change an operator
can observe.

**Metadata round-trip of keys gc does not own.** `EnsureCanonicalMetadata`
canonicalises over the raw object, so a binding's own connection keys
survive untouched — including the `postgres_dsn`/`postgres_schema` shape
bd itself writes. gc must never force an operator to hand-convert away
from a shape bd will write back.

**The credential-resolved event, generalised.** `pg.credential_resolved`
becomes `backend.credential_resolved` with a `backend` field, and the
payload moves to `internal/events` beside the other registered-but-not-
yet-emitted payload. Deleting it would not remove a capability — it
would change the OpenAPI `oneOf` and fork the wire contract between two
assemblies of the same product, against the typed-wire invariant and
`TestEveryKnownEventTypeHasRegisteredPayload`. Spec, generated Go client
and dashboard client/bundle are regenerated.

`redactPreflightDetail` now keys DSN redaction on `dsn` rather than
`postgres_dsn`, so a diagnostic contributed for a backend gc does not
implement is redacted by the rule that already exists.

## What the enterprise delta adds

Register the backend name, and optionally a doctor check. That is the
whole registrar — no four-channel `BackendIntegration`, no
`Recoverable(error)` hook, no registrar-supplied env projection. Those
existed only to keep the projection path pluggable, and there is no
longer a projection path to plug into.

## Tests

- `TestOSSProjectsNoUnregisteredBackendEnv` — the standing proof that
  replaces what pgauth's census provided. Three halves, because a
  credential can leak through three doors: the registry can readmit the
  name, a source file can name its environment variables, and the
  projector can write them. Red before, on all three:
  `backend "postgres" is registered by this build`;
  a 37-line source census listing `BEADS_POSTGRES_*` / `GC_POSTGRES_*`
  across `cmd/gc/bd_env.go`, `internal/orders/env.go`,
  `internal/pgauth/auth.go` and `internal/doctor/checks/postgres_auth.go`;
  and `bdRuntimeEnvWithErrorNoRecovery accepted a scope whose backend
  this build does not register`.
- `TestRuntimeEnvDelegatesCompleteStorageBindingToBd` is now the
  end-to-end witness: its city names a backend this build does not
  register, carries dolt fields that were once fatal and a key gc has
  never heard of, and still boots — city and inherited rig both get
  every projected variable withheld, `BD_BIN` and the credentials file
  set, and the metadata comes back byte-for-byte.
- Five refusal paths (`bdCommandEnv`, `orderExecEnvWithError`,
  `slingStoreEnvWithError`, `controllerState.openRigStore`,
  `controllerQueryRuntimeEnv`) previously asserted a pgauth credential
  error propagates; they now assert the unregistered-backend refusal
  propagates, through one shared `assertRefusesUnregisteredBackend` that
  holds the typed cause, the offending name and the registered set.
- The cmd/gc suite's `writePGScopeFixture` becomes
  `writeOpaqueBindingScopeFixture`, so ~20 existing tests across sling,
  orders, sessions, exec stores and lifecycle now exercise the opaque
  path against a backend name OSS does not register.
- `TestLoadMetadataStateRejectionOrderIsPinned` holds the two-rung
  ladder, including a case whose backend-specific keys are complete —
  E2 still speaks.
- `TestEnsureCanonicalMetadataPreservesKeysGCDoesNotOwn` and
  `TestCrossBackendKeysToScrubOnlyRemovesKeysGCWrites` pin the
  round-trip promise from both sides.

## Parity

A single-store dolt city was diffed across nine projection surfaces
(city and rig bd env, session env, native-open env, exec store env,
order exec env, city runtime process env, lifecycle scope-init env) plus
on-disk metadata and config, against `origin/main`. The only difference
is that the five always-empty `BEADS_POSTGRES_*` placeholders and
`GC_POSTGRES_PASSWORD` no longer appear; every other key, value, path
and file is identical. Mutation-checked: dropping one line from
`mirrorBeadsDoltServerEnv` moves eight lines in that same diff.

## Gates

`go build ./...` · `go vet ./...` · `gofmt`/`gofumpt` ·
`golangci-lint 2.10.1` (0 issues) · `go test ./internal/...` ·
`go test ./cmd/gc -timeout 25m` · `scripts/check-core-boundary.sh` ·
`make check-split-topology-rows` · `make check-routed-test-rows` ·
OpenAPI + generated-client + dashboard bundle regenerated ·
resource-census and GC_* env baselines lowered to bank the reduction.
Re-run on the review commit `8e6d9e9f3f`: `go test ./cmd/gc` 688s exit 0,
`go test ./internal/...` exit 0, `golangci-lint` 0 issues.


julianknutsen:
## Review fixes pushed — `8e6d9e9f3f`

Four confirmed MAJOR findings, all fixed with a test that is red without the fix. PR body updated with the detail; summary here.

### M1 + M3 — the fifth site (`bdCommandRunnerWithManagedRetryErr`)

`bdCommandRunnerWithManagedRetryErr` lost both postgres guards with nothing substituted, and the surviving gate chain cannot see a storage binding: `canonicalScopeDoltTarget` reports no managed runtime for a bound scope, so it falls through to `managedLocalDoltEnv`, and because `applyCompleteNonDoltStorageBindingEnv` withholds the whole projected namespace, `GC_DOLT_HOST` is `""` and `contract.DoltHostIsLocal("")` is true. **A withheld projection is indistinguishable from managed-local Dolt**, so `managedBDRecoveryAllowed` answered `true` for exactly the scopes gc must not manage.

Measured with the production runners, opaque binding + `dial tcp 10.0.0.5:5432: connect: connection refused` — attempts / recoverCalls:

| runner | before | after |
| --- | --- | --- |
| `controlBdCommandRunnerForCity` | 2 / 1 | **1 / 0** |
| `bdCommandRunnerForRig` — inherited binding | 2 / 1 | **1 / 0** |
| `bdCommandRunnerForRig` — own binding | 2 / 1 | **1 / 0** |
| `controlBdCommandRunnerForRig` — inherited binding | 2 / 1 | **1 / 0** |
| `controlBdCommandRunnerForRig` — own binding | 2 / 1 | **1 / 0** |
| `bdCommandRunnerForCity` | 1 / 0 | 1 / 0 — unaffected, short-circuits to `bdContextCommandRunnerForCity` |

Identical for backend `postgres`, `acme` and `mysql`, and the bd error stays reachable via `errors.Is` in every case. Not a postgres artifact — the PR's own generic opaque-binding shape.

**Fix shape chosen: fold the predicate into `managedBDRecoveryAllowed`**, not an early return in the runner.

- It has exactly one production caller (`bdTransportErrorMatches`), which feeds *both* `bdTransportRetryableError` and `bdTransportRecoverableError`, which are called only from the runner. So the fold covers 100% of the surface and any future caller, and the function genuinely becomes the one place answering "may gc recover this store?" — the shape the PR's own `scopeStoreIsExternallyBound` doc claims.
- The operator-facing message was not worth preserving: on main the protection for the *surviving* opaque shape came from `postgresMetadataForScope`'s classify-*error* branch, so the message was `classifying scope backend (bd error: …)` — worse reading than the raw transport error the fix now surfaces unwrapped.
- An unreadable binding (`err != nil`) answers "no recovery", matching what main's `TestPGTransportError_MetadataReadErrorSkipsManagedRecovery` pinned.

New tests, direct ports of the deleted `TestPGTransportError_NoManagedRecovery` / `…InheritedCityPostgresNoManagedRecovery`, built on `writeOpaqueBindingScopeFixtureWithBackend(t, root, "acme")` so they pin the shape rather than the vocabulary:

- `TestControlBdCommandRunnerForCityBoundCitySkipsManagedRecovery`
- `TestBdCommandRunnerForRigBoundScopeSkipsManagedRecovery` (inherited + own binding)
- `TestControlBdCommandRunnerForRigBoundScopeSkipsManagedRecovery` (inherited + own binding)

Red before, verbatim: `attempts = 2, want 1: a store gc does not manage is not gc's to retry` and `recoverCalls = 1, want 0: managed-Dolt recovery must not run for a store gc does not manage`.

### M2 — inherited-rig session projection

`bd_env.go:509` calls `applyCityStorageBindingEnv(env, cityPath)` again instead of the narrower `applyCompleteNonDoltStorageBindingEnv`. New test `TestInheritedRigSessionEnvRefusesUnregisteredCityBackend`.

Red before: the rig session env came back `err = nil` carrying `GC_DOLT_HOST:leftover-dolt.example.test GC_DOLT_PORT:3306 BEADS_DOLT_SERVER_HOST:leftover-dolt.example.test BEADS_DOLT_SERVER_PORT:3306`, while the city session for the same city returned `unsupported backend "mysql" (supported: dolt, doltlite); no storage database was opened or modified`.

### M4 — the vacuous test

`TestBdCommandRunnerForCityCompleteStorageBindingSkipsManagedRetry`'s shim now emits `dial tcp 10.0.0.5:5432: connect: connection refused`, `recoverManagedBDCommand` is stubbed and asserted never called, and bd invocations are counted (want 1). Verified discriminating by the same mutation the old test survived — `if completeBinding && false` at `bd_env.go:36`:

```
recoverCalls = 1, want 0: gc must not run managed-Dolt recovery for a store it does not manage
bd invocations = 2, want 1: a bound scope's failed command must not be retried
```

### Minors

- `canonicalScopeDoltProjectionAuthoritative` asks the predicate again — a bound city gets no projection, so stripping its `GC_DOLT_PASSWORD`/`BEADS_DOLT_PASSWORD` mirrors removed auth nothing restores. Test `TestCanonicalScopeDoltProjectionAuthoritativeSkipsABoundCity`; red before: `canonicalScopeDoltProjectionAuthoritative = true for a city gc does not project`.
- `scopeMetadataJSONPath` has its doc comment back; `scopeStoreIsExternallyBoundBestEffort` moved beside `scopeStoreIsExternallyBound`.
- `internal/beads/contract/files.go` no longer advertises the deleted E3–E5 rungs (`E1–E5` → `E1–E2`, two doc comments).
- `TestBackendCredentialResolvedPayloadOmitsTheCredential` exists — the invariant the `BackendCredentialResolved` doc comment cites. With the emitter gone there is no resolver to canary through, so it closes the wire type's field set instead (plus `TestBackendCredentialResolvedIsRegistered`).

Not taken: the rejected census finding.

### No sixth site

Every base call of `scopeBackendIsPostgres` / `postgresMetadataForScope` walked:

| base site | disposition |
| --- | --- |
| `bd_env.go:1312`, `:1324` — `bdCommandRunnerWithManagedRetryErr` | **was unguarded — fixed (M1/M3)** |
| `bd_env.go:419` — `canonicalScopeDoltProjectionAuthoritative` | **was unguarded — fixed (minor)** |
| `bd_env.go:1479` + caller `:1532` — `cityPostgresProjectionErrorCanBeBypassed` | deleted with pgauth; head always returns `cityErr` — strictly stricter |
| `beads_provider_lifecycle.go:624` — `scopeSkipsManagedDoltForInit` | migrated (own + inherited binding) |
| `beads_provider_lifecycle.go:1791` — `syncConfiguredDoltPortFiles` | migrated (city + rig + explicit inherited arm) |
| `cmd_doctor.go:78`, `:95` — `doctorWorkspaceHasPostgresScope` | deleted with `PostgresAuthCheck`; no `pgauth` reference remains |
| `dolt_runtime_publication.go:81` — `managedDoltLifecycleOwned` | migrated |
| `dolt_runtime_publication.go:224` — `clearManagedDoltRuntimeStateUnlessPostgres` → `…UnlessBound` | migrated |
| `init_provider_readiness.go:731` | migrated |
| `order_store.go:288`, `:314` | migrated |

`managedBDRecoveryAllowed` callers after the fix: `bdTransportErrorMatches` (production, sole) → `bdTransportRetryableError` / `bdTransportRecoverableError` → `bdCommandRunnerWithManagedRetryErr` (`bd_env.go:1103`, `:1106`), plus three classifier call sites in `bd_env_test.go`. Nothing else in the tree.

Preflight is generalised too, not dropped: a non-dolt backend still FAILs `PreflightCheckMetadataBackend` / `PreflightCheckContractShape` by name.

### Gates

`go build ./...` · `go vet ./...` · `gofmt` · `golangci-lint 2.10.1` **0 issues** · `go test ./internal/...` exit 0 · `go test ./cmd/gc -timeout 25m` **688.5s exit 0** · `scripts/check-core-boundary.sh` OK · `make check-split-topology-rows` OK · `make check-routed-test-rows` OK. No regeneration needed — no `internal/api/`, OpenAPI or dashboard input changed.

Issue #5008, opened 2026-08-05
typed session id in mail addressing is not preserved end-to-end


jacobhausler:
TITLE: `gc mail send` never preserves a typed session ID — the addressing form, not the caller, decides whether the recipient is pinned

BODY:

### Summary

`gc mail send` collapses **every** address form to a bare alias in the
stored message record, regardless of what the caller typed:

- `gc mail send <alias>` stamps `.assignee = <alias>` (a bare name,
  re-resolved at read time — this is the mechanism behind a sibling report
  in this batch, where bare-name resolution picks the wrong session among
  several sharing a template).
- `gc mail send <session-id>` — even though the caller typed a fully
  qualified, unambiguous session ID — **also** stamps `.assignee` down to
  the bare alias. The typed ID is discarded.
- `gc mail reply` (to an existing message) *does* stamp `.assignee =
  <session-id>`, preserving the exact target.

So the one addressing form that *can* pin a specific session
(`reply`) is not one the sender gets to choose — it's only available when
there's an existing message to reply to. A sender who wants to guarantee
delivery to one specific, already-known session has no `send`-side way to do
it.

### Reproduction

```
$ gc session list
ID              TEMPLATE   STATE   TARGET
ra-wisp-exdg5i  moiraine   active  moiraine

$ gc mail send ra-wisp-exdg5i -s "probe" -m "probe" --json
# creates wisp ra-wisp-2sx6bo

$ bd show ra-wisp-2sx6bo --json | grep assignee
"assignee": "moiraine"       <-- typed session id ra-wisp-exdg5i did NOT survive
```

`gc mail send`/`gc mail reply` were cross-verified in the same session, same
recipient, by two independent operators via `bd show`: `send` always stores
the bare alias; `reply` stores the session ID. This is consistent across
both address forms `send` accepts (alias or session ID) — both collapse to
the alias.

### Why this is the root complaint behind the other addressing reports

This defect is *why* the other issues in this filing batch are hard to
detect at send time. The only delivery-verification procedure currently
available to an operator is: read the stored `.assignee` and cross-check it
against `gc session list`. Because the session ID never survives into that
field, this check cannot distinguish "delivered to the exact session I
named" from "stored a name that will be re-resolved later, possibly to a
*different or asleep* session." It reports the same "success" shape in both
cases.

In the specimen above, the check happened to pass for the *right* reason,
because exactly one live session matched the alias `moiraine` at
verification time — but running the identical procedure against a
[...]

PR #5147, opened 2026-08-09
fix(beads): run the conditional-release CAS against a bd that has the flags (ga-nv0i3)

julianknutsen:
## Problem

`BdStore.ReleaseIfCurrent` routes through bd's native CAS verb — `bd update <id> --if-assignee <expected> --if-status in_progress --status open --assignee ""`, where a rejected precondition is exit 13 with nothing written — and falls back to hand-built SQL behind a per-store latch when bd rejects the flags as unknown (#5126).

That fallback is still the **live** path everywhere, because `deps.env BD_VERSION=v1.1.0` predates the flags. Measured against the pinned release tarball:

```
$ bd version
bd version 1.1.0 (8e4e59d39: HEAD@8e4e59d39f34)
$ bd update --help | grep -c -- --if-assignee
0
```

So on a stock install the store latches to `UPDATE issues … WHERE id=…`, which can never match the `wisps` table — every wisp-held assignment answers *"someone else holds it"*, forever, and gascity runs formula steps as wisps. `TestBdStoreReleaseIfCurrentAgainstRealBd`, added by #5126 to guard exactly this, skipped in every job that installs `BD_VERSION`. **It guarded nothing.**

## Why `BD_VERSION` does not move here

ga-nv0i3 asked for a `BD_VERSION` bump. There is nothing to bump to:

| anchor | value | has `--if-assignee`? |
|---|---|---|
| `BD_VERSION` (installable default) | v1.1.0 tarball | no (verified: 0 hits in `bd update --help`) |
| newest published beads release | v1.1.2 | no (cut from `hotfix/v1.1.1`, a branch off v1.1.0) |
| `BD_CURRENT_REF` (source-built cell) | `bf97b73749ac` | **yes** |

The capability boundary is one commit: the flags — and the exit-13 verdict — landed on beads main in `6e8af8bf8` (beads#5008, 2026-07-24). `4b1d9c875` (beads#5364, 2026-08-05) only ports the same CAS to proxied-server mode, so a bd after #5008 and before #5364 already has what the embedded contract needs; pins are compared against **#5008**. `install-bd-archive.sh` installs a **published** tarball with a pinned SHA, so pinning an unpublished tag would break every build. The same constraint blocks raising `BD_PREV_VERSION` past #5008 — which is what would let `bdSQLStringLiteral`, `releaseIfCurrentViaEmbeddedDoltSQL`, `parseDoltRowsAffected` and `isBdSQLUnsupportedInEmbeddedMode` die together. Both pins stay; the reasons are now recorded in `deps.env`.

## Fix

CI already obtains a flag-capable bd: the contract matrix's `current` cell builds one from source at `deps.env BD_CURRENT_REF`, and `contrib/k8s/Dockerfile.agent` ships the same build. Run the CAS row there.

- **`make test-bd-conditional-release-contract`** runs the row with `GC_REQUIRE_BD_CONDITIONAL_RELEASE=1`, which turns the capability skip into a **failure**. A cell that installed a capable bd and skipped anyway has proven nothing — it now fails loudly, including if a future `BD_CURRENT_REF` bump loses the flags or bd renames them.
- **`contract-acceptance-current`** gains one step invoking that target, after the bd it already builds is on PATH. Unset, the row still skips on every shard that installs `BD_VERSION`, so nothing else changes.
- **`deps.env`** records why `BD_VERSION`/`BD_PREV_VERSION` are stuck and what unblocks them, so the next bump attempt does not repeat the archaeology or pin something unfetchable.
- The skip message and `legacyBdReleaseRunner` named #5364; corrected to **#5008** throughout, along with the source comment that credited only `BD_PREV_VERSION` for keeping the SQL fallback alive — the **installable default** is the load-bearing one, so the fallback is production's shape today, not a floor nobody runs.
- `scripts/cipolicy` execution hash re-pinned for the new step; stale `v1.0.4` in the devcontainer README corrected to the actual `BD_VERSION`, and that anchor — the only `BD_VERSION` anchor no test read, which is how it went stale — is now asserted against `deps.env` (with the `DOLT_VERSION` anchor beside it).

### The cell cannot go green without running

The claim this rests on is "the target hardcodes `GC_REQUIRE_BD_CONDITIONAL_RELEASE=1`, so a skip is a failure — therefore the job passing IS the proof the CAS ran." Two holes let the cell pass having executed nothing; both are closed and both are demonstrated below.

- **A `-run` selector that matches nothing is not an error.** `go test` prints `[no tests to run]` and exits 0, so renaming the row, moving it, or dropping its build tag would leave a permanently green cell. The target now resolves the name through `go test -list` first, using the **same** variable, package and build tags the run uses. Deriving both from one variable is why this guard cannot rot into a false pass: there is no second copy of the selector to drift, and a `-list` output change would fail the cell loudly — the opposite of grepping `go test`'s human-facing "no tests to run" warning, which would silently stop catching.
- **The require-env was honoured on one branch only.** A bd whose `update` verb disappeared makes the capability probe **error** rather than report the flags absent, and that path skipped — green, under the env, having proven nothing. Every non-proving exit (probe error, flags absent, a leg that is not drivable) now routes through one helper that fails under the env and skips without it. bd being absent outright was already a `t.Fatal` on every run. Under the env there is exactly **one** green path: the row ran end to end. The env is also read as a bool, so `=true` no longer quietly means "not required".

## Verification

Against a bd built the way the cell builds it (source @ `BD_CURRENT_REF`, `-tags gms_pure_go`), the whole row runs — **wisp leg included**:

```
--- PASS: TestBdStoreReleaseIfCurrentAgainstRealBd (13.54s)
    --- PASS: .../a_wrong_assignee_writes_nothing_and_is_not_an_error (0.97s)
    --- PASS: .../the_matching_assignee_releases (1.04s)
    --- PASS: .../a_second_release_is_a_no-op,_not_an_error (0.60s)
    --- PASS: .../an_ephemeral_bead_releases_too (2.01s)
    --- PASS: .../an_unresolvable_id_is_not_held (0.87s)
    --- PASS: .../a_non-exact_id_refuses_instead_of_releasing_a_different_bead (2.06s)
```

**Escape hatch 1 — rename the row.** Before, the selector matched nothing and the cell passed:

```
$ go test -tags integration -run '^TestBdStoreReleaseIfCurrentAgainstRealBd$' ./internal/beads
ok  	github.com/gastownhall/gascity/internal/beads	0.061s [no tests to run]
EXIT=0
```

After, the same rename is red before a single test runs:

```
$ make test-bd-conditional-release-contract
Error: TestBdStoreReleaseIfCurrentAgainstRealBd does not exist in ./internal/beads under -tags integration.
The -run selector below would match nothing, and this cell would pass having run no test.
Point BD_CONDITIONAL_RELEASE_TEST/_PKG at the row's current name and home.
make: *** [Makefile:581: test-bd-conditional-release-contract] Error 1
```

**Escape hatch 2 — make the capability probe error** (a bd whose `update` verb is gone). Before, under `GC_REQUIRE_BD_CONDITIONAL_RELEASE=1`:

```
--- SKIP: TestBdStoreReleaseIfCurrentAgainstRealBd (4.95s)
ok  	github.com/gastownhall/gascity/internal/beads	5.011s
EXIT=0
```

After, same shim, same env:

```
--- FAIL: TestBdStoreReleaseIfCurrentAgainstRealBd (4.90s)
    bd update --help failed, so the capability probe could not run at all (bd broken, or the
    update verb is gone?): exit status 1: Error: unknown command "update" for "bd"
    [GC_REQUIRE_BD_CONDITIONAL_RELEASE is set: this run promised a bd that can prove the CAS …]
FAIL	github.com/gastownhall/gascity/internal/beads	4.989s
EXIT=1
```

**The skip every other shard depends on is unchanged.** With the env unset, a flagless bd still SKIPs green (exit 0), and so does the probe-error shim. With the env set, a flagless bd, a probe error, a bd absent from PATH, and `=true` instead of `=1` all fail.

**The devcontainer anchor is now guarded.** Reverting it to `v1.0.4`:

```
--- FAIL: TestBDVersionPins (0.01s)
    bd_version_pin_test.go:142: .devcontainer/README.md says BD_VERSION is currently "v1.0.4", want "v1.1.0" (deps.env)
```

Gates: `go build ./...`, `go vet ./...` (and `-tags integration`), `gofmt`, `golangci-lint` 0 issues (and 0 introduced under `--build-tags integration`), `go test ./scripts/...`, `go test ./internal/beads/...`, `go test ./cmd/gc -timeout 25m`, `make test-ci-policy`, `scripts/check-core-boundary.sh`, `check-routed-test-rows`, `check-split-topology-rows`, `test/docsync` — all green.

Refs ga-nv0i3.

### DIFF
--- a/.devcontainer/README.md
+++ b/.devcontainer/README.md
@@ -14,7 +14,7 @@ This devcontainer reproduces a development environment for [gastownhall/gascity]
 | `git` | devcontainer feature | latest |
 | `gh` | devcontainer feature | latest |
 | `dolt` | `.github/scripts/install-dolt-archive.sh` (SHA-256 verified) | `DOLT_VERSION` from `deps.env` (currently 2.1.7) |
-| `bd` (Beads CLI) | `.github/scripts/install-bd-archive.sh` (SHA-256 verified) | `BD_VERSION` from `deps.env` (currently v1.0.4) |
+| `bd` (Beads CLI) | `.github/scripts/install-bd-archive.sh` (SHA-256 verified) | `BD_VERSION` from `deps.env` (currently v1.1.0) |
 | `gc` (Gas City) | `make install` from source | built from current commit |
 
 Versions come from `deps.env` so bumping is one file change.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -413,6 +413,13 @@ jobs:
         run: bd version
       - name: bd CLI contract (current)
         run: make test-bd-cli-contract
+      # The conditional-release CAS is the one bd contract only this cell can
+      # run: BdStore.ReleaseIfCurrent prefers bd's --if-assignee/--if-status verb
+      # (a rejected precondition is exit 13), and BD_VERSION — every other job's
+      # bd — is a published release that predates it, so the row skips there and
+      # the wisp-tier fix rides unguarded. The target refuses to skip.
+      - name: bd conditional-release CAS contract (current)
+        run: make test-bd-conditional-release-contract
 
   # NON-REQUIRED HEADxHEAD radar: builds bd from beads MAIN HEAD (live, not the
   # pinned BD_CURRENT_REF) and runs gc-HEAD's real bd CLI contract against it —
--- a/Makefile
+++ b/Makefile
@@ -561,12 +561,36 @@ test-bd-cli-contract:
 ## `--if-assignee`/`--if-status`, so it belongs on the source-built
 ## BD_CURRENT_REF cell. GC_REQUIRE_BD_CONDITIONAL_RELEASE=1 turns the row's
 ## capability skip into a failure, so the cell cannot pass while proving nothing.
+##
+## The existence preflight closes the other way this cell can pass having proven
+## nothing: a `-run` selector that matches no test is not an error to `go test`
+## — it prints "[no tests to run]" and exits 0 — so renaming the row, moving it
+## to another package, or dropping its build tag would leave a permanently green
+## cell guarding nothing. `go test -list` resolves the SAME name, package and
+## tags the run below uses, from the same variables, and must print the name
+## back. Deriving both from one variable is why this guard cannot rot into a
+## false pass: there is no second copy of the selector to drift, and any change
+## to `-list` output that broke the match would fail the cell loudly rather than
+## silently stop catching (which is the failure mode of grepping `go test`'s
+## human-facing "no tests to run" warning instead).
 BD_CONDITIONAL_RELEASE_TIMEOUT ?= 10m
+BD_CONDITIONAL_RELEASE_TEST = TestBdStoreReleaseIfCurrentAgainstRealBd
+BD_CONDITIONAL_RELEASE_PKG = ./internal/beads
 test-bd-conditional-release-contract:
 	@command -v bd >/dev/null 2>&1 || (echo "Error: bd not found; cannot run the conditional-release contract" >&2; exit 1)
+	@listed=$$($(TEST_ENV) GOFLAGS= GOENV=off GOWORK=off go test -tags integration \
+		-list '^$(BD_CONDITIONAL_RELEASE_TEST)$$' $(BD_CONDITIONAL_RELEASE_PKG) 2>&1) || { \
+		printf '%s\n' "$$listed" >&2; \
+		echo "Error: could not list $(BD_CONDITIONAL_RELEASE_TEST) in $(BD_CONDITIONAL_RELEASE_PKG)" >&2; \
+		exit 1; }; \
+	printf '%s\n' "$$listed" | grep -qx '$(BD_CONDITIONAL_RELEASE_TEST)' || { \
+		echo "Error: $(BD_CONDITIONAL_RELEASE_TEST) does not exist in $(BD_CONDITIONAL_RELEASE_PKG) under -tags integration." >&2; \
+		echo "The -run selector below would match nothing, and this cell would pass having run no test." >&2; \
+		echo "Point BD_CONDITIONAL_RELEASE_TEST/_PKG at the row's current name and home." >&2; \
+		exit 1; }
 	$(TEST_ENV) GC_REQUIRE_BD_CONDITIONAL_RELEASE=1 GOFLAGS= GOENV=off GOWORK=off \
 		go test -tags integration -timeout $(BD_CONDITIONAL_RELEASE_TIMEOUT) -count=1 \
-		-run '^TestBdStoreReleaseIfCurrentAgainstRealBd$$' ./internal/beads
+		-run '^$(BD_CONDITIONAL_RELEASE_TEST)$$' $(BD_CONDITIONAL_RELEASE_PKG)
 
 ## test-acceptance-b: run Tier B acceptance tests (lifecycle, ~5 min, nightly)
 ACCEPTANCE_B_TIMEOUT ?= 10m
--- a/Makefile
+++ b/Makefile
@@ -103,7 +103,7 @@ endif
 endif
 endif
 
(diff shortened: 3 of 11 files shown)


julianknutsen:
Council follow-up: seven minors fixed in `d15af48`, rebased onto current `main` (`ae407a7` lineage). Two of them undercut the PR's own thesis, so they came first.

**F1 — the cell greened on `[no tests to run]`.** Nothing bound the target's `-run` selector to a live test, so a rename made `go test` exit 0 having executed nothing. The target now resolves the name through `go test -list` before running it, from the **same** make variable, package and build tags the run uses.

I picked `-list` over parsing `go test`'s "no tests to run" warning, and over a companion test, because of *which way each one rots*. There is only one copy of the selector, so `-run` and the existence check cannot drift apart; and if `-list`'s output shape ever changed, the `grep -qx` stops matching and the cell goes **red** — loud. Grepping the human-facing warning fails the other way: a wording change makes the guard silently stop catching, which is the same disease as the bug. A companion test would have to be in the `-run` selector too, so renaming *it* reopens the hole.

**F2 — the require-env was honoured on one branch only.** If bd's `update` verb disappears, the capability probe *errors* instead of reporting the flags absent, and that path skipped — green under `GC_REQUIRE_BD_CONDITIONAL_RELEASE=1`. Every non-proving exit now routes through one helper: fails under the env, skips without it. bd absent was already a `t.Fatal` on every run. The env is read as a bool, so `=true` no longer quietly means "not required". Under the env there is exactly one green path.

### Both demonstrated

Rename the row — old target vs new:

```
$ go test -tags integration -run '^TestBdStoreReleaseIfCurrentAgainstRealBd$' ./internal/beads
ok  	github.com/gastownhall/gascity/internal/beads	0.061s [no tests to run]
EXIT=0

$ make test-bd-conditional-release-contract
Error: TestBdStoreReleaseIfCurrentAgainstRealBd does not exist in ./internal/beads under -tags integration.
The -run selector below would match nothing, and this cell would pass having run no test.
make: *** [Makefile:581: test-bd-conditional-release-contract] Error 1
```

Probe error (bd shim whose `update` verb is gone), both under `GC_REQUIRE_BD_CONDITIONAL_RELEASE=1` — before, then after:

```
--- SKIP: TestBdStoreReleaseIfCurrentAgainstRealBd (4.95s)
ok  	github.com/gastownhall/gascity/internal/beads	5.011s
EXIT=0

--- FAIL: TestBdStoreReleaseIfCurrentAgainstRealBd (4.90s)
    bd update --help failed, so the capability probe could not run at all (bd broken, or the
    update verb is gone?): exit status 1: Error: unknown command "update" for "bd"
    [GC_REQUIRE_BD_CONDITIONAL_RELEASE is set: this run promised a bd that can prove the CAS …]
EXIT=1
```

Full matrix against a flag-capable bd: capable+env → all 6 legs PASS (exit 0); flagless+unset → SKIP (0); probe-error+unset → SKIP (0); probe-error+env → FAIL (1); flagless+env → FAIL (1); `=true`+probe-error → FAIL (1); bd absent → FAIL either way. The skip every other shard depends on is untouched.

### The other five

- **F3** — `#5364` corrected to **`#5008`** in `legacyBdReleaseRunner`, and the `deps.env` archaeology now states the boundary explicitly: `6e8af8bf8` (beads#5008, 2026-07-24) is where the flags and the exit-13 verdict landed; `4b1d9c875` (beads#5364, 2026-08-05) only ports the same CAS to proxied-server mode, so a bd between them already has what the embedded contract needs. Both verified against the beads repo. `git grep 5364` now returns one hit, that sentence.
- **F4** — `ReleaseIfCurrent`'s doc comment no longer frames the SQL fallback as "for bd 1.0.4, the contract-tested minimum". It is live at the **installable default** (`BD_VERSION`) — every CI job and every operator install — which is what makes it load-bearing; the floor lands there too but is not the reason.
- **F5** — the devcontainer README's bd version is now asserted against `deps.env` (`assertDocPinAnchor` in the pin test). Reverting it to `v1.0.4` fails: `.devcontainer/README.md says BD_VERSION is currently "v1.0.4", want "v1.1.0" (deps.env)`. I guarded the `DOLT_VERSION` anchor on the line above it too — same shape, equally unread, one line to close. The assertion keys on the phrasing `` `KEY` from `deps.env` (currently VALUE)``, so a doc can opt out by dropping the parenthetical rather than by going stale.

### Gates

`go build ./...`, `go vet ./...` (+ `-tags integration`), `gofmt`, `golangci-lint` 2.10.1 0 issues, `--build-tags integration --new-from-rev=HEAD` 0 issues introduced (the repo's 50 pre-existing integration-tag issues are all in files this diff does not touch), `go test ./scripts/...`, `./internal/beads/...`, `./cmd/gc -timeout 25m`, `make test-ci-policy`, `scripts/check-core-boundary.sh`, `check-routed-test-rows`, `check-split-topology-rows`, `test/docsync` — all green, re-run after the rebase.

`expectedCIExecutionHash` did **not** need re-pinning: the workflow step is unchanged (the fix is in the Makefile target it invokes), and `scripts/cipolicy` passes as-is.

The one red check remains `Image vulnerabilities` — `golang.org/x/text` CVE-2026-56852 inside the agent image's bd binary, on inputs this diff does not touch; unchanged by these fixes. The required `Check` gate is green.

PR #3211, opened 2026-06-07
perf(api): hydration-free parallel status counts for /v0/city/{city}/status (#1896, ~100x)

sjarmak:
## Summary

Fixes the `/v0/city/{city}/status` read path, which took **5–40 s under load** because each request did serial, fully-hydrated per-rig scans — `CachedList` cloned ~38k beads/rig/request just to count status work.

This keeps the counting **hydration-free** at the cache + API layers (cache-clean counts are clone-free in-memory) and runs the per-rig fan-out in parallel, with a 3 s TTL reuse for interactive callers. Cache-miss / `IncludeClosed` paths fall back to the existing `List` path, so the answer is always correct.

Closes #1896.
Closes #3186 — same root cause (jsgerman-oss identified the event-sequence cache-keying as the primary driver plus the session/closed-history per-request scans). This implements that remediation: time-based cache reuse for interactive callers + clone-free counting + the parallel fan-out. Thanks to @jsgerman-oss for the precise analysis.

## Performance (honest)

Measured **~15–100x** on the status read path (`ListPath` ~1.55–1.94 ms/op, ~19.6 MB/op → `CounterPath` ~17.9 µs/op, ~10 KB/op on the count benches). The originally-claimed "~760x" did **not** reproduce — the honest figure is ~100x.

## Correctness note

An earlier revision routed counts through a `NativeDoltStore.Count` SQL path; cross-provider review caught that it **undercounted the wisps table** (NoHistory beads — the default storage for session/wait/nudge/order types) on every default-config city. That path was **dropped**; the store now falls back to the correct `List` on cache-miss, preserving the production win (clone-free cache-clean counts) without the undercount.

## Deferred (documented in code, not this PR)

- `WalkSize` ctx-plumbing on the status path
- `Store.List` ctx for the non-Counter leak
- closed-history count path

PR #5148, opened 2026-08-09
fix(api): federate the graph store on ready/list and fail loud on a dead leg (ga-4k2c3)

julianknutsen:
Closes the API half of the split-store graph blindness: `ga-4k2c3`.

## The bug, measured

On a split city the graph-class DAG — `gcg-` molecule roots, step beads, control
beads — lives in the relocated graph store. `State.BeadStores()` does not include
it, and `humaHandleBeadReady` / `humaHandleBeadList` federated only `BeadStores()`.
The whole execution DAG was therefore invisible **behind an authoritative 200**.

Measured on a city that already has the fix, same city same moment:
API ready = **22 beads, 14 of them `gcg-`**; without the leg, 5 beads, 0 `gcg-`.

## Two properties, both landed

**1. The graph-class DAG is federated.** Both arms federate the same ordered
legs: the work stores, then the relocated graph store **last**. List builds an
ordered `[]beadListLeg` rather than merging into the rig-keyed map, so the graph
leg reuses the existing per-leg body — keyset seek, the global re-sort, the page
cut — while `Count`-based page bounding is planned **per leg** (see *Review
follow-ups*, M-A).

**2. A dead graph leg is an authoritative 503, not a degraded Partial 200.**
This is Invariant 0 at the API layer, and it is the half `partialAggregator`
violated. A rig going dark is one scope reporting a hole and `partial` /
`partial_errors` says so honestly. The graph plane going dark means the DAG is
gone from the answer, and a work-only 200 is indistinguishable from *"the DAG
finished"*. So the graph leg has no partial tier at all: **hard failure and
`PartialResultError` both 503**, because a partial graph read leaves an unnamed
hole in a dependency graph and the response has no way to say which part is
missing.

Both arms are gated on **store identity** (`GraphBeadStore() != CityBeadStore()`)
— the same rule `handler_beads.go`'s by-id class resolver and
`handler_convoy_dispatch.go`'s workflow scan already use — so on a default city
they never fire. Proven by mutation, not by assertion:

| mutation | test that dies |
| --- | --- |
| drop the `graph != CityBeadStore()` guard | `TestBeadReadyLegacyCityDegradedCityStoreStaysPartial` (`status = 503, want 200`) and `TestBeadListLegacyCityDoesNotSmuggleCityStore` (`list items include city bead "hq-1" ... so it is no longer byte-identical`) |
| drop the `?rig=` exclusion | `TestBeadListRigScopedReadExcludesGraphLeg` (`rig-scoped items include the graph bead`) |
| let the graph leg back into the all-or-nothing `Counter` gate | `TestBeadListSplitCityKeepsWorkLegsBounded` (`city leg max List limit = 0, want 6`) |
| drop the cross-leg `seen[b.ID]` gate | `TestBeadListDedupesBeadIDsAcrossLegs` (`items=2 total=2, want 1/1`) |
| federate the graph leg FIRST instead of last | `TestBeadListDedupesBeadIDsAcrossLegs` (`winning row = "relocated binding copy"`) and `TestBeadListGraphLeg503NamesWorkLegFailures` |
| revert the list 503 to degradation | `TestBeadListGraphLegHardFailIs503`, `TestBeadListGraphLegPartialIsAlso503` |
| revert the ready 503 to degradation | `TestBeadReadyGraphLegHardFailIs503`, `TestBeadReadyGraphLegPartialIsAlso503` |
| remove the ready leg | `TestBeadReadyFederatesGraphStore`, `TestBeadReadyFederationOrderIsLegConcatenation` |
| remove the list leg | `TestBeadListFederatesGraphStore` |

`TestBeadListLegacyCityUnchanged` names the *two* mechanisms that hold
duplicate-freedom on a legacy city (the identity guard and `sortedRigNames`'
store-identity dedup) so a future reader knows which one went red.

## Why this matters beyond the bug: CLI == API

The follow-on CLI federation (`gc ready`, `ga-oxsyu`) is blocked on this bead and
is specified **against this handler**, so its conformance test can assert
CLI == API instead of inventing an oracle. `humaHandleBeadReady`'s doc comment
now states the contract it has to match:

- **Legs, in order:** city store → rigs by name ascending → relocated graph store.
  The CLI composite's legs are work-then-infra, the same relative order.
- **Within a leg:** whatever order that leg's own `Ready` reader emits.
  Deterministic, but **not** canonical across leg kinds: a caching-wrapped work
  store emits `(priority, created_at, id)` via `sortBeadsReadyOrder`, while the
  canonical relocated graph binding (`beads.SQLiteStore`) emits
  `(created_at, id)` with no priority term at all (`sqliteReadySQL`).
- **Dedupe:** first leg to return an id wins.
- **Merged order:** leg concatenation, deliberately **not** re-sorted — a global
  re-sort would change the bytes a multi-rig single-store city already serves.
  Both sides therefore normalize with `beads.SortBeadsReadyOrder` before
  comparing — load-bearing precisely *because* the graph leg is not
  priority-ordered, and well-defined because that order is total.
- **Failure:** rigs degrade, the graph leg does not.

`TestBeadReadyFederationOrderIsLegConcatenation` pins the leg order so the
contract is executable, not just prose.

## Scope: the other set-returning handlers

`humaHandleBeadList` had the identical gap and is **in scope** — a half-federated
API is the failure mode this program keeps hitting, and the lifted Landmine #13
suite pins both handlers.

Audited every other `state.BeadStores()` fan-out. Three share the gap and are
deliberately **out of scope**:

- `statusWorkCounts` (`/v0/status`) — its comment claimed it federated ready work
  *"exactly like GET /beads/ready"*, which this PR would have made false. The
  comment now names the divergence and why closing it is its own slice: that read
  is cache-only and error-lenient by design (`cacheColdRigs`), the exact opposite
  of the graph leg's fail-loud contract.
- `humaHandleConvoyList` (`/v0/convoys`) — graph-relocated workflow convoys are
  invisible to the list even though `isGraphConvoyID` already probes the graph
  store by id.
- `findActiveBeadForAssignees` — an agent holding a claimed graph step shows no
  active bead.

## Notes

- Fixtures come from `internal/beads/splittest`, so the graph leg mints real
  `gcg-` ids and the work legs mint their own disjoint prefixes. Two exceptions:
  a leg whose read must *fail* wraps a plain leaf, because
  `splittest.StrictStore` implements `Handles()` and `beads.HandlesFor` would
  resolve straight past the failure injection; and a leg whose subject is the
  `beads.Counter` **type assertion** wraps a Counter-less leaf, because
  `StrictStore` declares `Counter` and the production class binding does not
  (recorded on `splittest.StrictStore.Count`).
- Counting the graph leg as a backend attempt has one deliberate consequence,
  pinned by `TestBeadReadyWorkStoresDownButGraphUpIsPartial200`: a split city
  whose work stores are all down but whose graph plane answered gets a Partial
  200 carrying the DAG, not `totalOutage`'s 503.
- **No wire change.** 503 was already in both ops' declared error set;
  `TestOpenAPISpecInSync` passes and the OpenAPI spec and generated dashboard
  types are untouched.
- `relocatedGraphStore` is the shared "is graph relocated?" predicate; new call
  sites should use it. `handler_convoys.go` and `handler_convoy_dispatch.go`
  still inline the same test — left alone to keep this a bug-fix diff.
- No conformance-suite KNOWN GAP becomes closable: I1/I2 (HQ work store leg), I5
  (`gc hook --claim`) and I10 (wake-filter class arm) are all cmd/gc controller
  paths, and the suite's SKIP list is still empty.

## Review follow-ups (three MAJOR + five minors, `e0d215b540`)

**M-A — the graph leg silently disabled bounded paging for the whole request.**
The leg was merged into the fan-out map *before* `sortedRigNames`, so it reached
`beadListBoundedTotal`, which is all-or-nothing. The canonical compiled-in
binding is a raw `*beads.SQLiteStore` with **no `Count` at all**, so one
Counter-less leg switched bounding off for **every** leg — every work store back
to hydrating full closed history on the dashboard's `?all=true` read, behind a
2s response cache. Worse, the comment added with the leg asserted the opposite,
and the `splittest` fixtures hid it because `StrictStore` declares
`beads.Counter`.

Fixed by planning bounding **per leg** rather than by giving `*SQLiteStore` a
`Count`: the defect is the veto, and a per-leg plan is robust to *any* binding
shape, where a new exact-count surface on `SQLiteStore` would fix today's
binding while leaving the veto for tomorrow's — and would trade a perf bug for
a wrong-`Total` correctness risk. Work legs keep the #3211 all-or-nothing
`Count`-fallback contract; the graph leg is bounded when it can `Count` and
hydrates alone when it cannot, its exact un-seeked total taken from the rows it
returned.

Differential, identical fixture, `GET ?type=molecule&all=true&limit=5`, work
legs `Counter`-capable, sole variable the graph leg:

| | city `Count` | city `List` `Limit` | rig `List` `Limit` | graph `List` `Limit` |
| --- | --- | --- | --- | --- |
| legacy, before **and** after | 1× | `[6]` | `[6]` | — |
| split, **before** (`c5a54c134d`) | 0× | `[0]` | `[0]` | `[0]` |
| split, **after** | 1× | `[6]` | `[6]` | `[0]` |

**M-B — the list fan-out had no cross-leg id dedupe.** `gc storage migrate`
preserves ids (`CreateWithForeignID`) and never deletes back, so a converged
migrated split city holds the same infrastructure bead in the work store *and*
the binding — convergence is *defined* as that overlap — and the list served it
twice with a doubled `total`. The append loop now applies the same
`seen[b.ID]` rule `humaHandleBeadReady` already had. **Winner: first leg wins,
graph leg last, so the work store's row is served** — identical on both
endpoints. The per-`(rig, assignee)` key it replaces was a strict subset of that
rule. Honest limit, documented on `resolveBeadListPage`: page and has-more are
exact (they come from the deduped rows), but a *counted* leg cannot see
co-residence, so the bounded `Total` can run high — never low.

**Minors taken:** the graph-leg 503 now carries the work-leg errors recorded
before it, so *"the graph plane is down"* and *"the whole city is down"* stop
being byte-identical; the synthetic `infra:<city>` key is **gone** (its
*"':' cannot appear in a rig name"* justification was unenforced, and a rig
carrying that name was silently overwritten and dropped from the response —
`TestBeadListRigNamedLikeTheGraphLegIsStillServed`); and the FEDERATION
CONTRACT's per-leg canonical-order claim is corrected above.

**Fixture fidelity was itself a finding**, so it is recorded where it bites:
`splittest.StrictStore.Count` now documents that every kit store type-asserts as
`beads.Counter` while the class binding implements no `Count`, and any fixture
whose subject is that assertion must use a Counter-less leaf. Separately, two
bead-list fixtures left both `MemStore`s on the default prefix, so two "rigs"
minted the same ids — a shape `config.ValidateRigs` rejects on every start,
reload and config-edit path. They are prefix-disjoint now.

## Gates

`go build ./...` · `go vet ./...` · `gofmt` · `golangci-lint run` **0 issues** ·
`go test ./internal/api/ ./internal/beads/...` · `go test ./cmd/gc -timeout 25m` (722s) ·
`scripts/check-core-boundary.sh` OK · `TestOpenAPISpecInSync` green and the
dashboard Go suites pass; nothing on the wire moved, so no regeneration.

### DIFF
--- a/internal/api/beads_graph_federation.go
+++ b/internal/api/beads_graph_federation.go
@@ -1,0 +1,47 @@
+package api
+
+import (
+	"github.com/gastownhall/gascity/internal/api/apierr"
+	"github.com/gastownhall/gascity/internal/beads"
+)
+
+// graphFederationLegKey returns the synthetic fan-out key the bead-list
+// federation files the relocated graph store under.
+//
+// The list fan-out is keyed by rig name, so the graph leg needs a key of its
+// own. ':' cannot appear in a rig name, so the key can never collide with a
+// real one — and the merge that mints it is skipped for rig-scoped requests, so
+// the key is not addressable through ?rig= either.
+func graphFederationLegKey(cityName string) string {
+	return "infra:" + cityName
+}
+
+// relocatedGraphStore returns the graph-class store when — and only when — the
+// city has actually relocated it.
+//
+// Routing is keyed on STORE IDENTITY, the same rule handler_beads.go's by-id
+// class resolver and handler_convoy_dispatch.go's workflow scan use: on a
+// default city GraphBeadStore() returns the city store itself, so there is
+// nothing to federate and every caller's extra arm stays dead. That identity
+// guard is what keeps a single-store city byte-identical.
+func relocatedGraphStore(state State) beads.Store {
+	graph := state.GraphBeadStore().Store
+	if graph == nil || graph == state.CityBeadStore() {
+		return nil
+	}
+	return graph
+}
+
+// graphPlaneUnavailable is the authoritative failure a dead graph leg produces.
+//
+// This is the half of the federation that is NOT a partial degradation. A rig
+// going dark is one scope reporting a hole, and Partial/partial_errors says so
+// honestly. The graph plane going dark means the execution DAG — molecule
+// roots, step beads, control beads — is gone from the answer, and a work-only
+// 200 is indistinguishable from "the DAG finished". So the graph leg either
+// answers or the request fails loud, including when it answers PARTIALLY: a
+// partial graph read leaves an unnamed hole in a dependency graph, and the
+// response has no way to say which part is missing.
+func graphPlaneUnavailable(op string, err error) error {
+	return apierr.StoreUnavailable.Msg("graph store " + op + " read failed (graph plane unreadable): " + err.Error())
+}
--- a/internal/api/beads_graph_federation.go
+++ b/internal/api/beads_graph_federation.go
@@ -1,20 +1,11 @@
 package api
 
 import (
+	"strings"
+
 	"github.com/gastownhall/gascity/internal/api/apierr"
 	"github.com/gastownhall/gascity/internal/beads"
 )
-
-// graphFederationLegKey returns the synthetic fan-out key the bead-list
-// federation files the relocated graph store under.
-//
-// The list fan-out is keyed by rig name, so the graph leg needs a key of its
-// own. ':' cannot appear in a rig name, so the key can never collide with a
-// real one — and the merge that mints it is skipped for rig-scoped requests, so
-// the key is not addressable through ?rig= either.
-func graphFederationLegKey(cityName string) string {
-	return "infra:" + cityName
-}
 
 // relocatedGraphStore returns the graph-class store when — and only when — the
 // city has actually relocated it.
@@ -42,6 +33,16 @@ func relocatedGraphStore(state State) beads.Store {
 // answers or the request fails loud, including when it answers PARTIALLY: a
 // partial graph read leaves an unnamed hole in a dependency graph, and the
 // response has no way to say which part is missing.
-func graphPlaneUnavailable(op string, err error) error {
-	return apierr.StoreUnavailable.Msg("graph store " + op + " read failed (graph plane unreadable): " + err.Error())
+//
+// workLegErrors are the degraded work legs recorded before the graph leg failed.
+// They ride along because this 503 replaces the Partial 200 that would have
+// carried them: without them "the graph plane is down" and "every store in the
+// city is down" are byte-identical responses, and the operator loses the
+// work-side diagnosis the response had already collected.
+func graphPlaneUnavailable(op string, err error, workLegErrors ...string) error {
+	detail := "graph store " + op + " read failed (graph plane unreadable): " + err.Error()
+	if len(workLegErrors) > 0 {
+		detail += "; work legs also degraded: " + strings.Join(workLegErrors, "; ")
+	}
+	return apierr.StoreUnavailable.Msg(detail)
 }
--- a/internal/api/handler_beads_bounded_test.go
+++ b/internal/api/handler_beads_bounded_test.go
@@ -58,7 +58,12 @@ func (s *countingListStore) Count(_ context.Context, q beads.ListQuery, excludeT
 	return n, nil
 }
 
-func seedMoleculeStore(total int) ([]beads.Bead, *beads.MemStore) {
+// seedMoleculeStore builds a store of molecule beads minting under idPrefix.
+// The prefix is a parameter because two rig stores in a real city mint under
+// different prefixes (config.Rig.EffectivePrefix) — two seeded stores sharing
+// one prefix would hold the same ids, which is a bead co-resident in two legs,
+// not two beads.
+func seedMoleculeStore(idPrefix string, total int) ([]beads.Bead, *beads.MemStore) {
 	base := time.Date(2026, 6, 1, 0, 0, 0, 0, time.UTC)
 	seed := make([]beads.Bead, 0, total)
 	for i := 0; i < total; i++ {
@@ -67,7 +72,7 @@ func seedMoleculeStore(total int) ([]beads.Bead, *beads.MemStore) {
 			status = "closed"
 		}
 		seed = append(seed, beads.Bead{
-			ID:        fmt.Sprintf("gc-mol-%03d", i),
+			ID:        fmt.Sprintf("%s-mol-%03d", idPrefix, i),
 			Type:      "molecule",
 			Status:    status,
 			Title:     fmt.Sprintf("molecule %d", i),
@@ -100,7 +105,7 @@ func TestBeadListAllTrueBoundsCounterStore(t *testing.T) {
 	const total = 30
 	const limit = 10
 	fs := newFakeState(t)
-	seed, mem := seedMoleculeStore(total)
+	seed, mem := seedMoleculeStore("gc", total)
 	store := &countingListStore{Store: mem}
 	fs.stores["myrig"] = store
 
@@ -145,7 +150,7 @@ func TestBeadListAllTrueBoundsCounterStore(t *testing.T) {
 func TestBeadListAllTrueNoCursorWhenAllFit(t *testing.T) {
 	const total = 8
 	fs := newFakeState(t)
-	_, mem := seedMoleculeStore(total)
+	_, mem := seedMoleculeStore("gc", total)
 	fs.stores["myrig"] = &countingListStore{Store: mem}
 
 	body := fetchBoundedBeads(t, fs, "?type=molecule&all=true&limit=50")
@@ -167,7 +172,7 @@ func TestBeadListAllTrueFallsBackWithoutCounter(t *testing.T) {
 	const total = 30
 	const limit = 10
 	fs := newFakeState(t)
-	_, mem := seedMoleculeStore(total)
+	_, mem := seedMoleculeStore("gc", total)
 	// Plain MemStore is not a Counter, so the handler must not bound it.
 	fs.stores["myrig"] = mem
 
@@ -215,7 +220,7 @@ func TestBeadListAllTrueBoundedTotalExcludesFailedRigList(t *testing.T) {
 	const good = 30
 	const limit = 10
 	fs := newFakeState(t)
-	_, mem := seedMoleculeStore(good)
+	_, mem := seedMoleculeStore("gc", good)
 	fs.stores["myrig"] = &countingListStore{Store: mem}
 	// "zrig" Counts 10 molecules but its List fails with a non-partial error,
 	// so its 10 rows never reach the page. Pre-fix, those 10 stay baked into
@@ -310,9 +315,9 @@ func TestBeadListAllTrueBoundedPartialResultRigKeepsCount(t *testing.T) {
 	const good = 30
 	const partial = 12
 	fs := newFakeState(t)
-	_, goodMem := seedMoleculeStore(good)
+	_, goodMem := seedMoleculeStore("gc", good)
 	fs.stores["myrig"] = &countingListStore{Store: goodMem}
-	_, partMem := seedMoleculeStore(partial)
+	_, partMem := seedMoleculeStore("zr", partial)
 	fs.stores["zrig"] = &countingPartialListStore{countingListStore: &countingListStore{Store: partMem}}
 
 	// Limit covers everything so the whole set must be reachable in one page.
--- a/internal/api/handler_beads_infra_federation_fanout_test.go
+++ b/internal/api/handler_beads_infra_federation_fanout_test.go
@@ -1,0 +1,352 @@
+package api
+
+import (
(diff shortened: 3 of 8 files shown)


julianknutsen:
## Review follow-ups landed — `e0d215b540`

Three MAJOR findings (two of them the same defect from two reviewers) plus five minors. Every fix has a test that fails without it; red-before strings inline.

### M-A — the graph leg silently disabled bounded paging for the *whole* request

Confirmed exactly as reported. The leg was merged into `stores` before `sortedRigNames`, so the synthetic `infra:<city>` key reached `beadListBoundedTotal`, which returns `(false, nil)` on the **first** store failing the `beads.Counter` assertion. The canonical compiled-in binding is a raw `*beads.SQLiteStore` — no `Count` method at all — so one Counter-less leg collapsed **every** leg onto the O(history) scan, on precisely the deployment this federation exists for. The comment at :72-77 asserted the opposite, and the `splittest` fixtures kept it invisible because `StrictStore` declares `beads.Counter`.

**Fix shape chosen: per-leg bounding, not `SQLiteStore.Count`.** Justification: the defect is the *veto*, not the missing method. A per-leg plan is robust to any binding shape (the beads adapter and the SQLite front door both *do* implement `Counter`, and a downstream provider could supply anything); adding `Count` to `*SQLiteStore` would fix today's binding and leave the veto in place for tomorrow's. It would also mean building a new exact-count parity surface — `sqliteListSQL` translates most of `ListQuery` but `List` still re-filters through `Matches` (tier, `Assignees`, `SeekAfter`, post-sort `Limit`) — where a subtle mismatch is a *wrong `Total` on the wire*, i.e. trading a perf regression for a correctness one. Wrong direction. `SQLiteStore.Count` remains a reasonable independent improvement.

So: `beadListBounding` keeps the #3211 all-or-nothing `Count`-fallback contract for the **work** legs, and puts the **graph** leg outside that gate — bounded when it can `Count`, hydrating alone when it cannot, its exact un-seeked total taken from the rows it returned (with the page boundary applied Go-side for that leg, so `Total` stays constant across a walk).

**Differential re-run**, identical fixture, `GET ?type=molecule&all=true&limit=5`, Counter-capable work legs, sole variable the graph leg:

```
PR head c5a54c134d
  legacy (no graph leg)  cityCount=1 cityListLimits=[6] rigCount=1 rigListLimits=[6]
  split  (graph leg)     cityCount=0 cityListLimits=[0] rigCount=0 rigListLimits=[0] graphListLimits=[0]

after e0d215b540
  legacy (no graph leg)  cityCount=1 cityListLimits=[6] rigCount=1 rigListLimits=[6]
  split  (graph leg)     cityCount=1 cityListLimits=[6] rigCount=1 rigListLimits=[6] graphListLimits=[0]
```

`Limit=6` restored on the split city; the legacy row is unchanged in both directions; only the leg that genuinely cannot count hydrates.

Red-before, `TestBeadListSplitCityKeepsWorkLegsBounded`:
```
city leg max List limit = 0, want 6 — a non-Counter graph leg must not unbound the work legs
rig leg max List limit = 0, want 6 — a non-Counter graph leg must not unbound the work legs
Count called: city=false rig=false, want both — bounding did not engage
```
Control `TestBeadListSplitCityBoundsACounterCapableGraphLeg` (green before *and* after) proves the exemption is scoped to legs that genuinely cannot count, and `TestBeadListSplitCityBoundedWalkIsTheFullScanPrefix` walks the split city to exhaustion with a constant `Total` and no repeats.

### M-B — no cross-leg id dedupe

Confirmed. Fixed with the same `seen[b.ID]` rule `humaHandleBeadReady` already had, applied unconditionally in the append loop.

**Winner rule: first leg wins, and the graph leg is federated LAST — so the work store's retained row is the one served,** byte-identical to ready, and both endpoints now resolve a co-resident bead to the same row (asserted in the same test). Stated in the FEDERATION CONTRACT comment.

On keeping the per-`(rig, assignee)` key: it is a strict subset of the global id gate (the same bead matched by two assignee terms is a same-id repeat), so keeping both would be dead weight. Dropped, and the reason is in the comment.

Red-before:
```
TestBeadListDedupesBeadIDsAcrossLegs: items=2 total=2, want 1/1 — a bead co-resident in the work store and the binding is ONE bead
TestBeadListDedupedWalkServesEachIDOnce: page 0 Total = 6, want 3 (co-residence must not inflate the total)
```

One honest limit, documented on `resolveBeadListPage`: `items`, has-more and the cursor walk are exact (they come from the deduped rows), but a *counted* leg reports its own rows without seeing co-residence, so the bounded `Total` can run **high** on a converged city whose graph binding can `Count`. Never low, and the walk still terminates on the real end of the set. Exactness there is unreachable without hydrating every leg, which is the thing M-A is about not doing.

### Minors taken (3 of 5; the other two were restatements)

- **503 discarded the work-leg diagnosis.** `graphPlaneUnavailable` now takes the errors `partialAggregator` recorded before the graph leg failed. `"graph plane down"` and `"whole city down"` are no longer byte-identical. Both arms. Red-before: `503 body "...graph store list read failed..." drops the work-leg failure`.
- **Synthetic `infra:<city>` key collision.** The key is **gone**. Legs are an ordered slice, so there is no rig namespace to collide with — the *"':' cannot appear in a rig name"* invariant was unenforced, and a rig carrying that name was silently overwritten in the merged map and vanished from the response. Red-before, `TestBeadListRigNamedLikeTheGraphLegIsStillServed`: `total = 1, want 2`. Removing the key also fixed the leg *order* — `"infra:test-city"` sorted **before** `"myrig"`, so list federated the graph leg first while ready federated it last.
- **FEDERATION CONTRACT per-leg order.** Corrected: a caching-wrapped work store emits `(priority, created_at, id)` via `sortBeadsReadyOrder`, but `sqliteReadySQL` emits `ORDER BY b.created_at ASC, b.id ASC` with **no priority term**, so "every leg is already in canonical order" was false for the leg this PR adds. The `SortBeadsReadyOrder` normalization is load-bearing, not a formality — now said so, in the handler doc and in the order test.

### Fixture fidelity

Recorded for the kit: `splittest.StrictStore.Count` now carries a **FIDELITY GAP** note — declaring the method makes every kit store type-assert as `beads.Counter` while the production class binding implements no `Count` at all, so a fixture whose subject is that assertion must use a Counter-less leaf. The new bounding tests do, and `TestFederationFixturesModelProductionCapabilities` fails if either side of that ever changes.

Two pre-existing bead-list fixtures also left both `MemStore`s on the default prefix, so two "rigs" minted the same ids. That is a shape `config.ValidateRigs` rejects on every start / reload / config-edit path (9 call sites), while the repo already reads a bead id as identity in `beads/ready` and every by-id endpoint. Made prefix-disjoint, and `TestPhase2BeadListAssigneeAliasKeepsCrossRigDuplicateIDs` is renamed to `...ServesEveryRigsBeads` with the real invariant it was protecting (the alias fan-out must not lose a rig's rows) asserted directly.

### Mutation matrix

| mutation | test that dies |
| --- | --- |
| drop `graph != CityBeadStore()` | `TestBeadReadyLegacyCityDegradedCityStoreStaysPartial` (`status = 503, want 200`), `TestBeadListLegacyCityDoesNotSmuggleCityStore` |
| put the graph leg back in the all-or-nothing gate | `TestBeadListSplitCityKeepsWorkLegsBounded` (`max List limit = 0, want 6`) |
| drop the cross-leg `seen[b.ID]` | `TestBeadListDedupesBeadIDsAcrossLegs`, `TestBeadListDedupedWalkServesEachIDOnce` |
| federate the graph leg first | `TestBeadListDedupesBeadIDsAcrossLegs` (`winning row = "relocated binding copy"`), `TestBeadListGraphLeg503NamesWorkLegFailures` |

Single-store byte-identity survives all of it: the legacy row of the differential is unchanged, and the legacy-city tests are green untouched.

### Gates

`go build ./...` · `go vet ./...` · `gofmt` clean · `golangci-lint run ./internal/api/... ./internal/beads/splittest/...` **0 issues** · `go test ./internal/api/ ./internal/beads/...` · `go test ./cmd/gc -timeout 25m` **ok 722s** · `scripts/check-core-boundary.sh` OK · `TestOpenAPISpecInSync` green and `./internal/api/dashboardspa/... ./internal/api/dashboardbff/...` pass — nothing on the wire moved, so no OpenAPI/dashboard regeneration.

PR #5149, opened 2026-08-09
feat(beads): declare and pin molecule membership semantics (ga-wnjr4)

julianknutsen:
Closes the membership half of the split-store program (ga-wnjr4). The change is:
name the rules, make every surface declare which one it implements, and pin the
live measurement so a future swap reddens instead of returning a plausible
number.

> **Behaviour change — read this first.** An earlier revision of this PR said
> "no surface changes which beads it returns". That is no longer true, and the
> change is the point. `GET /v0/city/{city}/beads/graph/{rootID}` now returns
> **the members of a wisp molecule**, where before it returned the root and
> nothing else. See "The wire declaration was false for a whole storage tier"
> below. Nothing else changes what it returns.

## The measurement, and the trap

Molecule root `gcg-arn`, read four ways on the same live graph:

| read | result |
|---|---|
| beads carrying `gc.root_bead_id == gcg-arn` | **60** (61 with the root) |
| `gc bd graph gcg-arn --json` | **61** |
| `gc bd dep tree gcg-arn --json` | **48** |
| in graph, not in dep tree | **13**, every one `gc.kind=spec` |

`60 - 13 + 1 = 48`. Dependency reachability returned *exactly* the set the
adopt-pr driver wants — and `pr_review.py`'s `is_frontier_candidate` already
excludes `gc.kind=spec` itself, so the driver wants direct membership and is
currently getting dep-reachability by coincidence.

The coincidence has a cause: `formula.newSourceSpecStep` builds a spec sidecar
with `DependsOn`, `Needs` and `WaitsFor` all cleared, so no dependency walk can
ever reach one. Nothing keeps the arithmetic true, and it fails in **both**
directions:

- A dep walk **drops non-spec members**. A member exists before its edges are
  wired — `processFanout` persists `gc.fanout_state=spawning`, instantiates
  fragments, wires the sink blockers afterwards — so a crash in between leaves
  ordinary work carrying the root id and nothing else.
- A dep walk **admits non-members**. `ensureDrainWorkflowBlocksOn` projects a
  `blocks` edge from *every* molecule bead onto a blocker **outside** the
  molecule, and a walk follows it straight out.

Either way the answer is a plausible number, not an empty one, which is why it
survives review.

## The wire declaration was false for a whole storage tier

Review caught this in the first revision, and it is the same class of bug this
PR exists to prevent — a surface publishing a rule it does not implement.

`BeadGraphResponse.Membership` is a **required** wire field asserting
`direct-root-id+parent-closure`. But both of `collectBeadGraph`'s list arms ran
at `ListQuery`'s **zero-value** `TierMode`, which is `TierIssues`, and
`ListQuery.matchesTier` reduces that to `!b.Ephemeral`. The rule the field
names — canonically `beads.DirectMembers` — reads `TierBoth`, because
`HandlesFor(store).Live` forces it. So the endpoint declared `direct-root-id`
and silently excluded an entire tier.

This is not a corner. A **wisp molecule's beads are all ephemeral**: under
`bd_compatibility="bd-1.0.5"` the wisp bead policy resolves to ephemeral
storage (`cmd/gc/bead_policy_store.go` `defaultBeadStorage`), and
`molecule/graph_apply.go` stamps `gc.root_bead_id` on every node. And
`store.Get` is tier-agnostic, so the root always resolves and the status is
always **200** — a wisp molecule read as a *complete, finished, empty* one
rather than erroring.

Reproduced on `MemStore` through the real HTTP handler and on a real
`beads.OpenSQLiteStore` via `CreateWithStorage(..., beads.StorageEphemeral)`:

| fixture | `beads.DirectMembers` | endpoint, before | endpoint, after |
|---|---|---|---|
| root + 1 ephemeral member + 1 durable member | 3 | 200, **2 beads** | 200, 3 beads |
| ephemeral (wisp) root + 2 ephemeral members | 3 | 200, **1 bead — the root alone** | 200, 3 beads |
| SQLite: ephemeral root + ephemeral + durable member | 3 | 200, **2 beads** | 200, 3 beads |

**Fix:** `TierMode: beads.TierBoth` on both of `collectBeadGraph`'s queries.
Not routed through `beads.DirectMembers`, deliberately — that reads through the
**LIVE** handle, which on a caching store bypasses the cache, and moving a hot
dashboard read onto a cache-bypassing handle is a latency change that deserves
its own slice (it is exactly what D4 defers). `TierBoth` on the store's own
handle makes the declaration true and changes nothing else. It also has to be
set on the parent-closure BFS arm, which `DirectMembers` would not have covered
at all.

The two backends drop the tier by **different mechanisms** — MemStore filters
in Go via `ListQuery.matchesTier`, SQLiteStore appends `b.tier='main'` to the
WHERE clause — so there is a regression test on each.

**The vocabulary now states this.** `beads.Membership` has a "No rule here has
a tier axis" section: every rule spans both tiers, that is not what a
`ListQuery` does by default, and if a surface ever does want a tier
restriction, the vocabulary needs a tier axis and the published value must say
so. An unqualified `direct-root-id` is what a consumer will assert against.

## What this PR does

**1. Names the rules** — `internal/beads/membership.go`:

```go
MembershipDirectRootID                 = "direct-root-id"
MembershipDepReachable                 = "dep-reachable"
MembershipRootIDAndParentClosure       = "direct-root-id+parent-closure"
MembershipRootIDParentClosureAndConvoy = "direct-root-id+parent-closure+convoy-members"
```

The type doc carries the measurement, the cause, both failure directions, the
tier statement, and the per-surface table. `beads.DirectMembers` is the
canonical `direct-root-id` implementation. `beads.AllMemberships()` makes the
vocabulary enumerable so a wire schema can be checked against it.

**2. Makes the fan-out say what it does.** `dispatch.listByWorkflowRoot` was
`direct-root-id` and its name said nothing. It is deleted; its 15 call sites
now read `beads.DirectMembers(store, rootID)`. Pure delegation — the body moved
verbatim, same `HandlesFor(store).Live.List` handle, same `IncludeClosed`, same
root-first ordering, same deliberate non-error on a missing root.

**3. Declares it on the wire, and implements what it declares.**
`BeadGraphResponse` gains a required `membership` field with an OpenAPI enum,
so a consumer can *assert* on the rule instead of assuming it.
`collectBeadGraph` returns the rule it applied rather than the handler
re-deriving the convoy predicate, and reads `TierBoth` on both arms so the
declaration is true. OpenAPI, `genclient/client_gen.go` and the TS client are
regenerated; the change is description text only, so the **embedded SPA bundle
rebuilds byte-identically** and `dist/` is untouched.

**4. Binds the enum to the code.** See "Enum binding" below.

**5. Pins 60/61/48/13** — `internal/beads/membership_test.go` reconstructs the
gcg-arn shape (1 root + 47 dep-linked members + 13 edgeless spec sidecars, half
of them closed) and holds all four numbers together, then breaks the identity
three ways. `dispatchDepReachable`/`depReachable` model `bd dep tree` as the
**most generous** possible walk — both directions, every edge type — so a bead
they miss is missed by any dep walk, not just by one direction of one edge kind.

**6. Adds split-topology invariant I12** so membership is answered identically
on both store arrangements, and names the silent-empty work-store answer as
ga-iaj7k's Invariant 0 gap rather than leaving it uncovered.

## Enum binding

Renaming a `Membership` constant's spelling used to leave the whole unit tier
green while the server violated the committed schema. The enum lives in a
struct tag, which is a string literal: the behavioural tests compare the
response against the same renamed constant, and `TestOpenAPISpecInSync`
compares the spec against the unchanged tag, so both stay green while the
server puts a value on the wire that the published enum forbids and the
generated clients reject.

`TestBeadGraphMembershipEnumMatchesTheVocabulary` walks the bridge: the Go
constants the endpoint can emit ↔ the `enum:` struct tag ↔ both committed
copies of `openapi.json` — in both directions — and additionally requires every
published enum value to be a rule `beads.AllMemberships()` defines, so a
hand-edited typo cannot reach a generated client.
`TestAllMembershipsCoversEveryDeclaredConstant` parses `membership.go` so that
list cannot go stale.

Proved by mutation: renaming `MembershipRootIDAndParentClosure`'s value to
`"root-id+parents"` and changing nothing else leaves
`TestBeadGraphDeclaresItsMembershipOnTheWire`,
`TestBeadGraphDeclaresConvoyMembershipOnTheWire` **and**
`TestOpenAPISpecInSync` all **PASS**, and reds only the new guard.

## The convoy rule understated what it returns

Parent-child descendants of convoy members are returned — the BFS frontier is
seeded from the whole accumulated member set — but the doc read as if
`+parent-closure` closed over the root only.

**Decision: correct the vocabulary and the wire doc; do not narrow the
implementation.** Narrowing it would change which beads the endpoint returns,
which is out of scope here, and it would be the worse answer anyway: a convoy
graph that stopped at each member and hid its subtree is not what a display
consumer wants. The spelling is accurate once it says what the closure closes
over, so the enum values are unchanged; only the description moves.
`TestBeadGraphConvoyMembershipClosesOverTheMembers` pins a two-level subtree
under a convoy member so the doc cannot drift back.

## Per-surface membership audit

| surface | membership implemented | correct for its consumers? |
|---|---|---|
| `beads.DirectMembers` (was `dispatch.listByWorkflowRoot`) — fan-out, retry, ralph, drain, scope, `findSpecBead` | `direct-root-id`, both tiers | **Yes.** `findSpecBead` resolves an edgeless sidecar; no other rule reaches one. |
| `dispatch.listByWorkflowRootAndScope` | `direct-root-id` ∩ `gc.scope_ref` | Yes. Members only — the root carries neither key. Now documented. |
| `molecule.ListSubtree` / `CloseSubtree` | `direct-root-id` ∪ parent closure | Yes. **Doc was wrong** — see defects. |
| `api.collectBeadGraph`, `GET /v0/city/{city}/beads/graph/{rootID}` | `direct-root-id` ∪ parent closure over the member set (+ convoy members for a convoy root) | Yes, **now**. It was **incomparable** to `direct-root-id`, not a superset: a superset via the parent closure and a strict subset via the tier filter. Fixed and declared on the wire. |
| `api.workflowSQLQueryWorkflowBeads` (Dolt fast path) | `direct-root-id`, re-implemented in SQL over both tier tables | Yes, but duplicated — see D4. |
| `api.snapshotFromStore` fallback branch | `direct-root-id` **minus the ephemeral tier**, re-implemented inline | **No** — see D4. |
| `storebinding.graphHasOpenDescendants` | `direct-root-id` first, parent/dep walk only as fallback | Yes. The fallback re-checks the root id on every dep hop, so it does not escape the molecule. |
| `runproj.RunMembers` / `snapshotForRun` | `direct-root-id` + `ParentID == root` (one level) + `"<root>."` id prefix | Yes for a run projection; a fourth distinct rule. Now documented as a superset. |
| `gc bd graph` (bd) | `direct-root-id` — 61 on gcg-arn | Yes. |
| `gc bd dep tree` (bd) | `dep-reachable` — 48 on gcg-arn | **Not for adopt-pr.** It is right today by coincidence. |
| `gc graph` (this repo's own command) | **none** — the named ids, only convoys expanded | See defects. |

## Defects found

**D1 — `molecule.ListSubtree`'s doc was wrong. FIXED (doc only).** It said "the
root bead and all transitive parent-child descendants". It has always *also*
scanned `gc.root_bead_id`, and that arm is load-bearing: teardown
(`CloseSubtree`, `sling.CloseAttachedSubtree`) must reach the edgeless spec
sidecars or the subtree stays half-open while cleanup reports success. A reader
trusting the doc would have "simplified" the arm away. Behaviour untouched.

**D2 — the fan-out's membership was unnameable. FIXED.** `listByWorkflowRoot`
is a *location* ("by workflow root"), not a *rule*. Nothing at any of its 15
call sites said whether it was reachability or a metadata scan.

**D3 — same conceptual question, two surfaces, two rules: `gc graph` vs
`gc bd graph`. REPORTED, not fixed.** `gc graph <molecule-root>` returns **one
bead**: `resolveGraphInput` expands only `beads.IsContainerType`, and that map
holds `"convoy"` alone — `"molecule"` and `"wisp"` are not container types.
`gc bd graph <same-root>` returns 61. Two commands one word apart answer a
1-vs-61 difference. *Proposed fix:* not "make `gc graph` expand molecules" —
that silently changes `gc graph <root> <unrelated-id>` — but either a
`--molecule` flag that applies `beads.DirectMembers`, or a note in the
command's `Long` help. Left alone here because it changes what an existing
command returns. Declared in the code instead: `resolveGraphInput` now states
that it implements *no* membership rule and why.

**D4 — `direct-root-id` has three implementations and they do NOT agree today.
REPORTED, not fixed. Tracked as `ga-212sl`.**

An earlier revision of this note said they "agree today and nothing keeps them
agreeing", and characterised the risk as cache staleness. That was wrong.
The divergence is not drift waiting to happen — it is a **permanent whole-tier
set difference that is present right now**:

- `beads.DirectMembers` reads `TierBoth` (via the LIVE handle).
- `api.workflowSQLQueryWorkflowBeads` queries **both** the `issues` and `wisps`
  table sets (`workflowSQLAvailableTableSets`), so it is tier-complete too.
- `api.snapshotFromStore`'s Go fallback lists at the zero-value `TierMode`, so
  it is `TierIssues` and cannot see a wisp molecule's members at all.

The last two are the two branches of **one function**, selected by whether the
Dolt server is reachable — which is exactly the coupling
`workflowSQLQueryWorkflowBeads`'s own doc says must not exist. For a wisp
molecule, that endpoint's answer depends on which branch ran.

*Fix (ga-212sl):* give `beads.DirectMembers` an options form that selects the
read handle, collapse the Go copies onto it, and add a conformance test
asserting the SQL fast path and the fallback return the same set for a durable
**and** a wisp molecule. Not done here: it changes what the workflow-snapshot
endpoint returns, which is a second behaviour change on a different endpoint
and belongs in its own slice. Both copies now name the shared rule and the
divergence in their docs.

**Not a defect:** `runproj.RunMembers` is a fourth rule, but its extras (direct
parent, dotted id prefix) are exactly what a run/iteration projection needs and
it is a strict superset of `direct-root-id`, so it can never lose a member the
canonical rule keeps.

## Red-before evidence

Every string below was produced by applying the mutation to the tree and
running the suite; all were reverted.

**M0 — the tier fix reverted** (`TierMode: beads.TierBoth` removed from both of
`collectBeadGraph`'s queries, nothing else changed). This is the review
finding:

```
--- FAIL: TestBeadGraphDeclaredMembershipCoversTheEphemeralTier
    handler_beads_graph_test.go:718: graph is missing ephemeral member gc-2, which carries gc.root_bead_id == gc-1; the endpoint declared "direct-root-id+parent-closure", and that rule has no tier axis
    handler_beads_graph_test.go:722: graph is missing ephemeral parent-linked descendant gc-4; the parent-closure half of "direct-root-id+parent-closure" is tier-scoped too
    handler_beads_graph_test.go:736: graph omits "gc-2", which beads.DirectMembers returns; the endpoint cannot declare "direct-root-id+parent-closure" and answer a subset of it
--- FAIL: TestBeadGraphDeclaredMembershipCoversAnEphemeralRoot
    handler_beads_graph_test.go:778: wisp molecule graph is missing member gc-2 ("Wisp step"); every node of a wisp molecule is ephemeral, so a tier-scoped read returns the root alone — 200 with an empty molecule, which is indistinguishable from a finished one
    handler_beads_graph_test.go:778: wisp molecule graph is missing member gc-3 ("Wisp step spec"); every node of a wisp molecule is ephemeral, so a tier-scoped read returns the root alone — 200 with an empty molecule, which is indistinguishable from a finished one
    handler_beads_graph_test.go:788: graph returned 1 beads, want 3 (beads.DirectMembers) — the endpoint declared "direct-root-id+parent-closure" and must not answer a subset of the rule it names
--- FAIL: TestBeadGraphDeclaredMembershipCoversTheEphemeralTierOnSQLite
    handler_beads_graph_test.go:846: graph is missing ephemeral member gc-2 on SQLite; the tier predicate is a SQL WHERE clause there, so the Go-side filter passing proves nothing about it
    handler_beads_graph_test.go:854: graph returned 2 beads, want 3 (beads.DirectMembers) — the endpoint declared "direct-root-id+parent-closure" and must not answer a subset of the rule it names
```

**M0b — the enum unbound from the code.** `MembershipRootIDAndParentClosure`'s
value renamed to `"root-id+parents"`, nothing else touched.
`TestBeadGraphDeclaresItsMembershipOnTheWire`,
`TestBeadGraphDeclaresConvoyMembershipOnTheWire` and `TestOpenAPISpecInSync`
all **PASS** under this mutation; only the new guard reds:

```
--- FAIL: TestBeadGraphMembershipEnumMatchesTheVocabulary
    handler_beads_graph_enum_test.go:46: the enum struct tag omits "root-id+parents", which the graph endpoint can emit; a consumer validating against the published enum would reject a valid response
    handler_beads_graph_enum_test.go:46: the enum struct tag declares "direct-root-id+parent-closure", which no membership constant the graph endpoint emits spells; regenerate the spec after renaming a beads.Membership value
    handler_beads_graph_enum_test.go:83: openapi.json omits "root-id+parents", which the graph endpoint can emit; ...
    handler_beads_graph_enum_test.go:87: openapi.json: membership enum names "direct-root-id+parent-closure", which beads.AllMemberships does not define — a generated client would advertise a rule that exists nowhere in the code
    handler_beads_graph_enum_test.go:83: ../../docs/reference/schema/openapi.json omits "root-id+parents", ...
```

**M1 — `DirectMembers` silently swapped to dependency reachability.** The
mutation this bead exists to catch:

```
--- FAIL: TestMoleculeMembershipPinsTheMeasuredGcgArnShape/direct-membership-is-the-root-plus-everything-carrying-the-root-id
    membership_test.go:118: DirectMembers(gcg-arn) = 48 beads, want 61 (the root plus 60 members) — MembershipDirectRootID answers "direct-root-id" and that count is the contract, not an implementation detail
--- FAIL: .../dep-reachability-drops-a-non-spec-member-that-carries-the-root-id
    membership_test.go:212: DirectMembers dropped gcg-arn.item.99, which carries gc.root_bead_id == gcg-arn — that is the whole rule
--- FAIL: .../dep-reachability-admits-a-bead-that-is-not-a-member
    membership_test.go:232: DirectMembers returned gcg-outside-1, which carries no gc.root_bead_id — direct membership must not follow edges
--- FAIL: TestDirectMembersReturnsMembersWhenTheRootIsGone
    membership_test.go:254: DirectMembers with a missing root = 47, want 60 (every member, no root)
```

the same mutation at the dispatcher:

```
--- FAIL: TestFanoutMembershipIsDirectRootIDNotDepReachability
    fanout_membership_test.go:72: the fan-out member set is missing spec sidecar gc-3; direct-root-id is the only rule that reaches a bead with no edges, and findSpecBead depends on it
```

and on both topologies:

```
--- FAIL: TestSplitTopologyConformance/I12-molecule-membership/single-store
    split_topology_conformance_test.go:1217: the molecule member set is missing gc-3 ("Step spec for the linked step"); direct-root-id is complete by construction on both topologies, and a short answer here is the shape that reads as a finished molecule
    split_topology_conformance_test.go:1222: member set = 2 beads, want 3 (root + step + spec); got [gc-1 gc-2]
--- FAIL: TestSplitTopologyConformance/I12-molecule-membership/split
    split_topology_conformance_test.go:1217: the molecule member set is missing gcg-3 ("Step spec for the linked step"); ...
```

**M2 — `IncludeClosed: false`** (a finished molecule reading as a smaller live one):

```
    membership_test.go:118: DirectMembers(gcg-arn) = 37 beads, want 61 (the root plus 60 members) — ...
```

**M3 — `DirectMembers` "helpfully" filtering `gc.kind=spec`** for its consumer.
This is the most instructive one: the mutation returns **exactly 48**, the
number the live dep tree returned, and the test still reds:

```
    membership_test.go:118: DirectMembers(gcg-arn) = 48 beads, want 61 (the root plus 60 members) — MembershipDirectRootID answers "direct-root-id" and that count is the contract, not an implementation detail
    membership_test.go:165: dep-reachability dropped 0 of the 48 direct members, want 13: []
--- FAIL: TestFanoutMembershipIsDirectRootIDNotDepReachability
    fanout_membership_test.go:72: the fan-out member set is missing spec sidecar gc-3; ...
```

**M4 — the wire declaration going stale** while the code still scans root ids:

```
--- FAIL: TestBeadGraphDeclaresItsMembershipOnTheWire
    handler_beads_graph_test.go:557: membership = "dep-reachable", want "direct-root-id+parent-closure" — the wire must name the rule that produced beads, not leave a consumer to infer it
```

**M5 — `collectBeadGraph`'s root-id arm dropped** (endpoint becomes parent
closure only, i.e. dep-reachability-shaped):

```
--- FAIL: TestBeadGraphDeclaresItsMembershipOnTheWire
    handler_beads_graph_test.go:566: graph is missing the dependency-isolated spec sidecar gc-2; the endpoint would then be answering "dep-reachable" while declaring "direct-root-id+parent-closure"
```

## Single-store byte-identity

`beads.DirectMembers` is `listByWorkflowRoot`'s body moved verbatim — same
`HandlesFor(store).Live.List`, same `ListQuery`, same root-first dedup, same
`ErrNotFound` handling — so every dispatch call site returns the same slice on
any topology. Proved by mutation rather than asserted: I12's `single-store` arm
reds identically to its `split` arm under M1, and the full `cmd/gc` and
`internal/...` suites pass unchanged.

## Gates

- `go build ./...` — clean
- `go vet ./...` — clean
- `gofmt -l cmd internal scripts` — clean
- `golangci-lint run ./...` (2.10.1) — **0 issues**
- `go test ./internal/api/... ./internal/beads/... ./internal/dispatch/...` — ok
- `go test ./cmd/gc -timeout 25m` — ok
- `make check-split-topology-rows` — OK (I12 conforms to Rules A/B/C)
- `scripts/check-generated-docs-drift.sh` — "Generated reference docs are fresh."
- `go run ./cmd/genspec` + `go generate ./internal/api/genclient` +
  `npm run generate:client` — regenerated. Only the `membership` description
  text changed, so `internal/api/dashboardspa/dist` rebuilds
  **byte-identically** and no bundle file is touched.


julianknutsen:
## Review fixes applied — 1 MAJOR + 2 minors, rebased onto `ae407a79bf`

Head is now `0f2f920cfb` (rebased onto current `origin/main`, which had advanced past #5148).

### MAJOR — the wire declaration was false for the ephemeral tier

Confirmed exactly as reported, and reproduced on **both** store backends before fixing:

| fixture | `beads.DirectMembers` | endpoint, before | after |
|---|---|---|---|
| root + 1 ephemeral member + 1 durable member | 3 | 200, **2 beads** | 3 |
| ephemeral (wisp) ROOT + 2 ephemeral members | 3 | 200, **1 bead — root alone** | 3 |
| SQLite (`OpenSQLiteStore` + `CreateWithStorage(..., StorageEphemeral)`) | 3 | 200, **2 beads** | 3 |

**Fix shape: `TierMode: beads.TierBoth` on both of `collectBeadGraph`'s queries.** Not routed through `beads.DirectMembers`, for two reasons:

1. `DirectMembers` reads through `HandlesFor(store).Live`, which on a caching store **bypasses the cache** — moving a hot dashboard read onto a cache-bypassing handle is the exact latency change D4 defers. `TierBoth` on the store's own handle makes the declaration true and changes nothing else.
2. It would not have covered the **parent-closure BFS arm** at all, which was tier-scoped too — an ephemeral parent-linked descendant was also being dropped.

I did **not** add a tier axis to the vocabulary, because no consumer of this endpoint wants one: the rule the field names is tier-complete, so the honest move is to implement it. The vocabulary now says so explicitly, and says what to do if that ever changes.

### Yes, the returned set changes — stated plainly

**`GET /v0/city/{city}/beads/graph/{rootID}` now returns the members of a wisp molecule where it previously returned the root alone.** This is a real behaviour change, it is correct, and it is called out in its own block at the top of the PR body rather than buried. Nothing else about what the endpoint returns changes. The old behaviour was the dangerous shape: `store.Get` is tier-agnostic so the root always resolved, the status was always **200**, and a wisp molecule read as *complete and empty* rather than as missing.

Two backends, two mechanisms, two tests: MemStore filters the tier in Go (`ListQuery.matchesTier`), SQLiteStore appends `b.tier='main'` to the WHERE clause. Passing one proves nothing about the other.

### Red-before (M0 — the fix reverted, nothing else changed)

```
--- FAIL: TestBeadGraphDeclaredMembershipCoversTheEphemeralTier
    handler_beads_graph_test.go:718: graph is missing ephemeral member gc-2, which carries gc.root_bead_id == gc-1; the endpoint declared "direct-root-id+parent-closure", and that rule has no tier axis
    handler_beads_graph_test.go:722: graph is missing ephemeral parent-linked descendant gc-4; the parent-closure half of "direct-root-id+parent-closure" is tier-scoped too
    handler_beads_graph_test.go:736: graph omits "gc-2", which beads.DirectMembers returns; the endpoint cannot declare "direct-root-id+parent-closure" and answer a subset of it
--- FAIL: TestBeadGraphDeclaredMembershipCoversAnEphemeralRoot
    handler_beads_graph_test.go:778: wisp molecule graph is missing member gc-2 ("Wisp step"); every node of a wisp molecule is ephemeral, so a tier-scoped read returns the root alone — 200 with an empty molecule, which is indistinguishable from a finished one
    handler_beads_graph_test.go:778: wisp molecule graph is missing member gc-3 ("Wisp step spec"); ...
    handler_beads_graph_test.go:788: graph returned 1 beads, want 3 (beads.DirectMembers) — the endpoint declared "direct-root-id+parent-closure" and must not answer a subset of the rule it names
--- FAIL: TestBeadGraphDeclaredMembershipCoversTheEphemeralTierOnSQLite
    handler_beads_graph_test.go:846: graph is missing ephemeral member gc-2 on SQLite; the tier predicate is a SQL WHERE clause there, so the Go-side filter passing proves nothing about it
    handler_beads_graph_test.go:854: graph returned 2 beads, want 3 (beads.DirectMembers) — the endpoint declared "direct-root-id+parent-closure" and must not answer a subset of the rule it names
```

### MINOR 1 — enum bound to the code

`TestBeadGraphMembershipEnumMatchesTheVocabulary` (new) walks the whole bridge in both directions: the Go constants the endpoint can emit ↔ the `enum:` struct tag ↔ **both** committed `openapi.json` copies. It additionally requires every published value to be a rule `beads.AllMemberships()` defines, so a hand-edited typo cannot reach a generated client. `TestAllMembershipsCoversEveryDeclaredConstant` (new) parses `membership.go` so `AllMemberships` cannot go stale.

Proved by mutation. Renaming `MembershipRootIDAndParentClosure`'s value to `"root-id+parents"` and changing nothing else:

```
=== RUN   TestBeadGraphMembershipEnumMatchesTheVocabulary
    handler_beads_graph_enum_test.go:46: the enum struct tag omits "root-id+parents", which the graph endpoint can emit; a consumer validating against the published enum would reject a valid response
    handler_beads_graph_enum_test.go:46: the enum struct tag declares "direct-root-id+parent-closure", which no membership constant the graph endpoint emits spells; regenerate the spec after renaming a beads.Membership value
    handler_beads_graph_enum_test.go:83: openapi.json omits "root-id+parents", ...
    handler_beads_graph_enum_test.go:87: openapi.json: membership enum names "direct-root-id+parent-closure", which beads.AllMemberships does not define — a generated client would advertise a rule that exists nowhere in the code
    handler_beads_graph_enum_test.go:83: ../../docs/reference/schema/openapi.json omits "root-id+parents", ...
--- FAIL: TestBeadGraphMembershipEnumMatchesTheVocabulary (0.04s)
--- PASS: TestBeadGraphDeclaresItsMembershipOnTheWire (0.04s)
--- PASS: TestBeadGraphDeclaresConvoyMembershipOnTheWire (0.03s)
--- PASS: TestOpenAPISpecInSync (0.20s)
```

Three PASSes under a mutation that puts a schema-forbidden value on the wire — that is the gap, and only the new guard closes it.

### MINOR 2 — convoy rule: **correct the doc, keep the implementation**

Confirmed: a two-level subtree under a convoy member is returned (`gc-4`, grandchild of the member, comes back). Decision and why:

- **Not narrowing the implementation** — it would change which beads the endpoint returns, which this change's scope excludes, and it is the worse answer: a convoy graph that stopped at each member and hid its subtree is not what a display consumer wants.
- **Correcting the vocabulary + the wire description** — `+parent-closure` is accurate once it states *what it closes over*: the whole accumulated member set, not the root alone. So the **enum values are unchanged**; only the description text moves. `TestBeadGraphConvoyMembershipClosesOverTheMembers` (new) pins it.

### D4 corrected — the stated basis was wrong, and the truth is sharper

D4 said the copies "agree today and nothing keeps them agreeing" and framed it as cache staleness. They do **not** agree today, and it is not staleness — it is a permanent whole-tier set difference present right now:

- `beads.DirectMembers` → `TierBoth` (LIVE handle).
- `api.workflowSQLQueryWorkflowBeads` → queries **both** the `issues` and `wisps` table sets (`workflowSQLAvailableTableSets`), so tier-complete.
- `api.snapshotFromStore`'s Go fallback → zero-value `TierMode` = `TierIssues`, blind to a wisp molecule entirely.

The last two are the two branches of **one function**, selected by whether the Dolt server is reachable — precisely the coupling `workflowSQLQueryWorkflowBeads`'s own doc says must not exist. Not fixed here (it changes a second endpoint's returned set); filed as **`ga-212sl`** and referenced from both call sites.

Also corrected: the audit row calling `collectBeadGraph` "a display superset". The sets were **incomparable** — superset via the parent closure, strict subset via the tier — and the row now says so.

### Gates (all on the final tree)

- `go build ./...` — clean
- `go vet ./...` — clean
- `gofmt -l cmd internal scripts` — clean
- `golangci-lint run ./...` (2.10.1) — **0 issues**
- `go test ./internal/api/... ./internal/beads/... ./internal/dispatch/...` — ok
- `go test ./cmd/gc -timeout 25m` — **ok, 720.109s**
- `scripts/check-generated-docs-drift.sh` — "Generated reference docs are fresh."
- `go run ./cmd/genspec` + `go generate ./internal/api/genclient` + `npm run generate:client` — regenerated. **Description text only**, so `internal/api/dashboardspa/dist` rebuilds **byte-identically**; zero bundle files touched (verified after `npm run build` + dist sync).

Not merged, as instructed.

PR #5150, opened 2026-08-09
fix(storage): create one-shot graph-class beads in the graph store (ga-xo8ch)

julianknutsen:
## Problem

`gc order run` and `gc formula cook` open **one** store — the one for the
scope they are working in — and materialize the whole molecule through it.
On a city whose graph class is served by its own storage binding, every
wisp root, every workflow root and every step they create is therefore
**born in the WORK ledger, minted under the WORK prefix.**

`internal/coordclass` classifies those beads as `ClassGraph`. The city's own
convergence check reads them as graph-class beads residing outside their
binding and reports them **stranded** — and stranded is **fatal to boot**.

#5106 fixed it for controller-driven order dispatch and explicitly listed
the remaining sites: *"Remaining graph-class births — convergence tick,
`cmd_formula`, control dispatch."* Control dispatch landed in #5108. This
PR is the one-shot CLI half.

## The rule: the store follows the CLASSIFIER, not the compiler version

`moleculeClassStore(recipe, workStore, graphStore)` (`cmd/gc/class_store.go`)
classifies the compiled recipe the way `coordclass.ClassifyGraphPlan`
classifies a plan — a molecule is materialized as one atomic plan, so one
graph-marked node makes the whole molecule graph class — and both one-shot
paths route on its verdict.

Asking *"did this formula use the v2 compiler"* is wrong **in both
directions**, and an earlier revision of this PR was wrong in both:

| shape | compiler | `coordclass.Classify` | store |
|---|---|---|---|
| graph.v2 pour (root `gc.kind=workflow`) | v2 | `ClassGraph` | graph binding |
| **root-only wisp** (`phase = "vapor"`, or no `[[steps]]`) | **v1** | **`ClassGraph`** (`gc.kind=wisp` is Classify's FIRST arm) | **graph binding** |
| **poured molecule** (container root + assigned task steps) | **v1** | **`ClassWork`** | **work ledger** |

The root-only row is a shipped, documented shape
(`docs/reference/specs/formula-spec-v1.md` "Root-only wisps";
`examples/bd/dolt/formulas/mol-dog-stale-db.toml`). `internal/formula/compile.go`
sets `rootOnly := (!f.Pour && f.Phase == "vapor") || len(f.Steps) == 0` and
stamps `gc.kind=wisp` on that root, and `coordclass.Classify` reaches the wisp
arm before `gc.root_bead_id` is ever consulted. Cooked into the work ledger it
is exactly the row `readInfraSnapshot` selects as stranded.

The poured row is the mirror error: relocating a work-class molecule hides the
step assigned to an agent from every work-scope reader — `gc hook` resolves
`GC_STORE_SCOPE=city` against the city WORK root, asserted by this PR's own
`conformanceClaimRouting` row.

### `gc order run` (formula arm) — `cmd/gc/cmd_order.go`

| write | before | after |
|---|---|---|
| `applyOrderRecipeRouting` | scope store | the store the recipe's class names |
| `molecule.Instantiate` (root + steps) | scope store | ⤷ same |
| root label stamp (`order-run:`, `order:`, `seq:`) | scope store | ⤷ same (evidence rides the bead it describes) |
| `EmitCurrent` graph leg | scope store | ⤷ same (work leg stays the scope store) |

The order-tracking bead is untouched: `orderTrackingFrontDoor` already put
it on the ORDERS class (#5127).

### `gc formula cook` (standalone arm, both compilers) — `cmd/gc/cmd_formula.go`

| write / read | before | after |
|---|---|---|
| `decorateFormulaCookGraphV2Recipe` | scope store | the store the recipe's class names |
| `molecule.Instantiate` (graph.v2 arm) | scope store | ⤷ same |
| legacy `molecule.Cook` | scope store | ⤷ same (inlined as compile + validate + Instantiate so the store can be chosen from the recipe; the compile-error wrapping is `molecule.Cook`'s, verbatim) |
| `--meta` root `SetMetadataBatch` | scope store | the store that created the root |
| `emitFormulaCookExecutionFacts` | one store as **both** legs | graph leg + work leg |

## Deliberately NOT routed

**Not graph class.** The synthetic input convoy `graphv2.PrepareInvocation`
mints (`coordclass` routes synthetic convoys to `ClassWork` explicitly), and a
v1 **poured** molecule. `TestFormulaCookLegacyMoleculeStaysOnTheWorkStore` pins
that boundary from **both** sides now — the graph binding stays empty *and* the
work store holds no graph-marked bead.

**Needs more than wiring — BOTH `--attach` arms.** A graft is a **two-ended
edge**: the sub-DAG root is graph class, the attach bead is whatever class owns
it, and `ensureFormulaCookAttachDep` / `molecule.Attach` writes a `blocks` row
binding them. Move only the root and the work store records an edge naming an
id it cannot resolve. Nothing rejects that write —
`internal/beads/splittest`'s own backend table says bd passes a cross-prefix
target through as an external ref and SQLite's `deps` table has no foreign key
— and every `Ready` implementation treats an unresolvable blocker as open
(`memstore.go readyLocked`: `statusByID[dep.DependsOnID] != "closed"`). The
caller's own work bead then leaves Ready **forever** and the documented graft
gate ("the bead won't close until the sub-DAG completes") is silently void.

An earlier revision of this PR shipped exactly that half-split on the graph.v2
`--attach` arm. It is reverted: **both** arms now stay whole on the scope
store, root and dep co-resident, byte-identical to `main`, with the deferral
named at each call site. Splitting them needs the attach bead's owning store
resolved **by id** — the shared resolver, `ga-k8pzw` — plus a cross-boundary
representation of the block, which is a production behavior change rather than
wiring.

**Order dispatcher — `ga-fk1a5`, pinned not closed.** `dispatchWisp` has the
same v1-poured defect the CLI twin just fixed, but its root is a **two-class
object**: it also carries the `order:`/`seq:` event-cursor labels, and
`orders.Store` unions order-run evidence across the **ORDERS and GRAPH legs
only** (`internal/orders/store.go`). Routing the molecule by class alone would
fix the steps and lose the cursor, so an event-triggered v1 poured order would
re-fire every tick. Closing it means persisting the cursor onto the
orders-class tracking bead the way the exec arm already does
(`front.SetCursor`) — its own slice. `gc order run` has no such tension: on
`main` its molecule *and* its cursor labels already live on the scope store, so
gating it by class is a pure restoration.
`TestOrderDispatchPouredV1MoleculeRelocatesWithTheGraphClass` pins the current
answer so `ga-fk1a5` has a test to flip.

## Writers moved, and every reader of what they wrote

| reader | routes at the graph class? |
|---|---|
| controller order gate / cooldown / cursor (`gateStoresFor`) | ✅ federates the graph binding since #5106 |
| `gc order check`, `gc order history`, `gc doctor` order firing | ✅ `cachedOrderStoresResolver` appends `relocatedOrdersClassStore`, and the only split shape this build serves puts all five infrastructure classes on ONE binding (`storageSplitWhole`), so that leg *is* the graph binding. Asserted by `TestOrderRunGraphResidentRunIsVisibleToTheOrderReaders` (`LastRun` **and** `Cursor`). |
| `gc order sweep-tracking --include-wisps` | ✅ `orderWispSweepStore` is `resolveGraphStore(cliStorageRoutes(...))` |
| control dispatcher | ✅ `controlGraphStore` (#5108) |
| `gc bd show/update` by id | ✅ `cmd_bd_by_id.go` + `storeref.ClassCandidates` (#5132) |
| API / dashboard run detail | ✅ `CityRuntime.graphBeadStore()` |
| `gc hook` (work scope) | ✅ **by construction** — a work-class molecule is no longer relocated away from it |
| `gc formula version-check <bead-id>` | ❌ **disagrees** — a by-id read; **named for `ga-k8pzw`, not fixed here.** Pre-existing (#5106/#5108 already relocated graph roots); this PR extends the set. |
| `gc sling`'s source-workflow replacement machinery (`internal/sling/sling_core.go`) | ❌ **disagrees** — by-id / scope-federated reads of graph roots; **named for `ga-k8pzw`.** |

## Tests

All drive the **real** production entry points — the cobra `gc formula cook`
command, `doOrderRunWithJSON`, and `memoryOrderDispatcher.dispatch` — against a
`storageRoutes` value shaped like the converged whole-split binding, seeded
through the production memo (`seedCLIStorageRoutes`).

### Red-before evidence (mutation applied, test run, string quoted verbatim)

| mutation | test | red-before |
|---|---|---|
| graph.v2 `--attach`: `Instantiate` on `graphStore`, dep on `store` (the earlier revision) | `TestFormulaCookAttachLeavesTheSourceBeadUnblockableOnSplitCity` | `work store holds dep gc-1 -> gcg-1 (blocks) whose target it cannot resolve: getting bead "gcg-1": bead not found` **and** `attach bead gc-1 is not Ready after the whole workflow closed (ready=[], root=gcg-1)` |
| attach arm's live-root lookup → `graphStore` | `TestFormulaCookAttachIsIdempotentOnSplitCity` | `re-cook minted a second workflow root (gc-2 then gc-5); the idempotent lookup read a store that does not hold the root` |
| legacy cook routed by version (`rootStore = store`) | `TestFormulaCookRootOnlyLegacyWispLandsInGraphStoreOnSplitCity` | `root-only wisp gc-1 is not resident in the graph binding: getting bead "gc-1": bead not found` |
| legacy cook relocated wholesale (`rootStore = graphStore`) | `TestFormulaCookLegacyMoleculeStaysOnTheWorkStore` | `legacy molecule root gcg-1 is not resident in the work store: getting bead "gcg-1": bead not found` |
| `--meta` stamp on the scope store | `TestFormulaCookMetaStampLandsOnTheGraphResidentRoot` | `gc formula cook graph-work: setting root metadata on gcg-1: setting metadata batch on "gcg-1": bead not found` |
| `gc order run` routes unconditionally (`moleculeStore = graphStore`) | `TestOrderRunPouredV1MoleculeStaysOnTheWorkStoreOnSplitCity` | `poured molecule root gcg-1 is not resident in the scope work store: getting bead "gcg-1": bead not found` |
| execution-fact work leg collapsed onto the graph store | `TestEmitFormulaCookExecutionFactsReadsTheConvoyFromTheWorkLeg` | `no execution.work_associated fact for mc-2 in run gcg-1 (events=[]); the convoy leg read a ledger the convoy does not live in` |
| `resolveClassStore`'s identity branch returns a fresh store | all four `…StaysOnTheOneStoreOnSingleStoreCity` | `wisp root gc-1 is not resident in the single store`, `cooked root gc-1 is not resident in the single store`, `root-only wisp gc-1 is not resident in the single store`, `graphStoreFor returned *beads.MemStore(0x…), want the identical store value 0x…` |

The `--attach` case is the assertion that would have caught the earlier
revision: it closes **every** bead in both stores and then asserts the source
bead is back in `Ready()` — the contract `--attach` documents.

### Also added

- `TestRecipeCoordClassRederivesTheVerdictFromTheClassifier` — the three
  compiler-emitted shapes plus the RootOnly-skipped-step case, stated at the
  helper.
- `TestFormulaCookRootOnlyLegacyWispStaysOnTheOneStoreOnSingleStoreCity` —
  the compatibility half for the new legacy route.
- `TestOrderDispatchPouredV1MoleculeRelocatesWithTheGraphClass` — the
  `ga-fk1a5` KNOWN GAP pin.

`TestEmitFormulaCookExecutionFactsReadsTheConvoyFromTheWorkLeg` is asserted on
the helper rather than through the cobra command, and the comment says why: no
one-shot cook can reach the two-leg state today, because `PrepareInvocation`
only mints an input convoy for a **targeted** invocation and the `--attach`
arms — the only targeted ones — now run wholly on the scope store. The helper
is where the leg assignment lives, so it is where a collapse is catchable.

## Single-store compatibility

**Byte-identical, by construction.** `resolveGraphStore` returns the *exact
store value it was handed* when the routes relocate nothing, and
`moleculeClassStore` picks between two values that are then the same value, so
the molecule create sees the same instance and the `GraphApplyFor` /
`StorageCreateStore` optional-capability assertions still resolve. The
single-store tests are green before and after by design; their teeth come from
the identity-branch mutation above.

## Conformance suite

No KNOWN GAP closes. I5's paragraph named *this* bead as the slice that would
close `gc hook --claim`; that attribution is wrong and is corrected to
`ga-k8pzw` — a claim is a by-id mutation of a bead that already exists. The
skip list stays empty; nothing was skipped or deleted. I12
(`conformanceMoleculeMembership`, #5149) is intact — an earlier rebase of this
branch silently dropped it and it has been restored.

## Gates

- `go build ./...` — clean
- `go vet ./...` — clean
- `gofmt -l cmd/gc internal` — clean
- `go test ./cmd/gc -timeout 25m` — ok
- `scripts/check-routed-test-rows.sh` — OK

### DIFF
--- a/cmd/gc/class_store.go
+++ b/cmd/gc/class_store.go
@@ -9,6 +9,7 @@ import (
 	"github.com/gastownhall/gascity/internal/coordclass"
 	"github.com/gastownhall/gascity/internal/events"
 	"github.com/gastownhall/gascity/internal/extmsg"
+	"github.com/gastownhall/gascity/internal/formula"
 	"github.com/gastownhall/gascity/internal/mail"
 	"github.com/gastownhall/gascity/internal/session"
 )
@@ -313,6 +314,55 @@ func resolveSessionStore(routes *storageRoutes, workStore beads.Store, cfg *conf
 // parity with the other resolve*Store helpers and ignored here).
 func resolveGraphStore(routes *storageRoutes, workStore beads.Store, cfg *config.City, cityPath string, rec events.Recorder) beads.Store {
 	return resolveClassStore(routes, workStore, cfg, cityPath, config.BeadClassGraph, rec)
+}
+
+// moleculeClassStore returns the store a compiled recipe's molecule must be
+// materialized in: graphStore when the beads instantiating it produces are
+// graph class, and the caller's own scope/work store otherwise.
+//
+// The question is the CLASSIFIER'S, not the compiler's. Routing on "did this
+// formula use the v2 compiler" is wrong in both directions. A v1 formula that
+// is root-only — `phase = "vapor"`, or no [[steps]] at all — compiles to a root
+// carrying gc.kind=wisp (internal/formula/compile.go), and that is the first
+// arm coordclass.Classify tests, so the bead is ClassGraph and belongs in the
+// binding. A v1 POURED formula compiles to a molecule container whose every
+// bead is ClassWork, and work stays on the work ledger even in a split city —
+// relocating it hides the steps from every work-scope reader, `gc hook`
+// included.
+func moleculeClassStore(recipe *formula.Recipe, workStore, graphStore beads.Store) beads.Store {
+	if recipeCoordClass(recipe) == coordclass.ClassGraph {
+		return graphStore
+	}
+	return workStore
+}
+
+// recipeCoordClass returns the coordination class of the beads that
+// instantiating recipe produces.
+//
+// molecule.Instantiate materializes a recipe as one atomic plan, so the plan is
+// classified wholesale exactly as coordclass.ClassifyGraphPlan does: a single
+// graph-marked node makes the whole molecule graph class, which keeps its
+// intra-plan edges inside one store. Steps a RootOnly recipe never creates are
+// skipped, matching the instantiate loop's own `if recipe.RootOnly && i > 0`
+// break.
+func recipeCoordClass(recipe *formula.Recipe) coordclass.Class {
+	if recipe == nil {
+		return coordclass.ClassWork
+	}
+	for i, step := range recipe.Steps {
+		if recipe.RootOnly && i > 0 {
+			break
+		}
+		stepType := step.Type
+		if stepType == "" {
+			stepType = "task"
+		}
+		bead := beads.Bead{Type: stepType, Labels: step.Labels, Metadata: step.Metadata}
+		if coordclass.Classify(bead) == coordclass.ClassGraph {
+			return coordclass.ClassGraph
+		}
+	}
+	return coordclass.ClassWork
 }
 
 // graphClassBinding returns the store these routes serve the graph class from,
--- a/cmd/gc/cmd_formula.go
+++ b/cmd/gc/cmd_formula.go
@@ -646,6 +646,20 @@ conflicting live workflow from the same source is an error.`,
 			if err != nil {
 				return formulaCommandError(stderr, "gc formula cook", jsonOutput, err)
 			}
+			// A graph.v2 cook materializes a workflow root and its steps, and
+			// every one of those beads is GRAPH class (molecule.Instantiate
+			// stamps gc.root_bead_id on each member of a graph pour). On a city
+			// whose graph class is served by its own binding, creating them
+			// through the scope store mints them in the WORK ledger under the
+			// work prefix — the stranded-infrastructure-bead shape that is fatal
+			// to boot. The scope store still answers the work-class beads the
+			// cook touches: the synthetic input convoy and, on the --attach arms,
+			// the attach bead. resolveGraphStore returns the exact store value it
+			// was handed on every city that relocates nothing, so a single-store
+			// cook runs against the one store it always did.
+			//
+			// The --attach arms are excluded — see the deferral comment on each.
+			graphStore := resolveGraphStore(cliStorageRoutes(cityPath), store, cfg, cityPath, nil)
 
 			cookVars := parseFormulaVars(vars)
 
@@ -655,6 +669,24 @@ conflicting live workflow from the same source is an error.`,
 					return formulaCommandError(stderr, "gc formula cook", jsonOutput, fmt.Errorf("load formula %q: %w", args[0], err))
 				}
 				if isGraphFormula {
+					// NOT routed at the graph class, deliberately — the same
+					// deferral the legacy molecule.Attach arm below carries, for
+					// the same reason. A graft is a two-ended edge: the sub-DAG
+					// root is graph class, the attach bead is whatever class owns
+					// it, and ensureFormulaCookAttachDep writes a `blocks` row
+					// binding them. Move only the root and the work store records
+					// an edge naming an id it cannot resolve. No backend rejects
+					// that write (internal/beads/splittest documents both: bd
+					// passes a cross-prefix target through as an external ref,
+					// SQLite's deps table has no foreign key), and every Ready
+					// implementation treats an unresolvable blocker as open — so
+					// the caller's own work bead silently leaves Ready forever
+					// and the documented graft gate is void. Splitting it needs
+					// the attach bead's owning store resolved by id — the shared
+					// by-id class resolver, ga-k8pzw — plus a cross-boundary
+					// representation of the block, which is a production behavior
+					// change rather than wiring. Whole arm stays on the scope
+					// store until then, root and dep co-resident.
 					storeRef := workflowStoreRefForDir(scope.storeRoot, cityPath, loadedCityName(cfg, cityPath), cfg)
 					var result *molecule.Result
 					var syntheticInputConvoyID string
@@ -727,7 +759,7 @@ conflicting live workflow from the same source is an error.`,
 							}
 							return err
 						}
-						emitFormulaCookExecutionFacts(store, cityPath, result, stderr)
+						emitFormulaCookExecutionFacts(store, store, cityPath, result, stderr)
 						return ensureFormulaCookAttachDep(store, attach, result.RootID)
 					})
 					if err != nil {
@@ -776,6 +808,16 @@ conflicting live workflow from the same source is an error.`,
 					graphRootKey = stampFormulaCookGraphV2Root(recipe, args[0], inv.InputConvoy, cookVars)
 				}
 
+				// NOT routed at the graph class, deliberately. molecule.Attach is a
+				// one-store operation that reads the attach bead, materializes the
+				// sub-DAG and writes the blocking dep through the SAME store, and
+				// on a split city those are two classes: the attach bead is
+				// whatever class owns it, while every sub-DAG bead is graph (Attach
+				// stamps the parent's gc.root_bead_id onto each step). Answering
+				// that needs the attach bead's owning store resolved by id — the
+				// shared by-id class resolver, ga-k8pzw — and a two-store Attach,
+				// which is a production behavior change rather than wiring. Left on
+				// the scope store until then.
 				result, err := molecule.Attach(cmd.Context(), store, recipe, attach, molecule.AttachOptions{
 					Title:          title,
 					Vars:           cookVars,
@@ -823,6 +865,10 @@ conflicting live workflow from the same source is an error.`,
 			cookVars = inv.Vars
 
 			var result *molecule.Result
+			// rootStore is the store the cooked root ended up in, so the --meta
+			// stamp below lands on the bead it just created rather than on a
+			// store that has never held it.
+			rootStore := store
 			if isGraphFormula {
 				// Stamp the run root with its store/scope identity before
 				// instantiating, exactly as the --attach branch does via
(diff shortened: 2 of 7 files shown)


julianknutsen:
## Council review addressed — blocker fixed, class verdicts re-derived, `--attach` gated

Pushed `4dd0c7b341` (rebased onto `1122a87e21`, force-with-lease from `f35b01c1d0`).

### BLOCKER — graph.v2 `--attach` wrote a dangling cross-store blocking dep

**Reproduced first, on the head, through the real cobra command**, before touching anything:

```
--- FAIL: TestFormulaCookAttachLeavesTheSourceBeadUnblockableOnSplitCity
    work store holds dep gc-1 -> gcg-1 (blocks) whose target it cannot resolve:
      getting bead "gcg-1": bead not found
    attach bead gc-1 is not Ready after the whole workflow closed
      (ready=[], root=gcg-1)
```

**Fix shape: (a).** The whole graph.v2 `--attach` arm goes back on the scope store — `decorate`, both `closeFailedRoots`, `existingRoot`, `liveInputConvoyRoots`, `ListLiveRoots`, `Instantiate`, and both `emitFacts` legs — byte-identical to `main`, with the same deferral comment the legacy `molecule.Attach` arm carries.

Why (a) and not (b): a graft is a two-ended edge and the write that binds it is a `blocks` row. Representing that block across a store boundary is not wiring — it needs the attach bead's owning store resolved **by id** (`ga-k8pzw`) plus a cross-boundary representation, i.e. a production behavior change. My own deferral comment on the sibling arm said so; shipping the same split forty lines earlier was the defect. (b) is `ga-k8pzw`'s territory and I am not doing it here.

The reviewers' point about the irony is taken as written. The comment at the top of the cook that asserted *"the scope store still answers … the attach dependency"* was false for the dependency and is corrected.

### The class verdicts, re-derived from `coordclass.Classify`

Both were wrong, in opposite directions. The store now follows the **classifier's verdict on the compiled recipe**, via one shared helper `moleculeClassStore(recipe, workStore, graphStore)` that classifies the recipe the way `coordclass.ClassifyGraphPlan` classifies a plan.

| shape | compiler | verdict | before | after |
|---|---|---|---|---|
| graph.v2 pour | v2 | `ClassGraph` | graph | graph |
| root-only wisp (`phase = "vapor"` / no `[[steps]]`) | **v1** | **`ClassGraph`** | **work (stranded)** | **graph** |
| poured molecule | **v1** | **`ClassWork`** | **graph via `gc order run` (invisible to `gc hook`)** | **work** |

Evidence, both directions, quoted from the mutation runs:

- route the legacy cook by version → `root-only wisp gc-1 is not resident in the graph binding: getting bead "gc-1": bead not found`
- relocate the legacy cook wholesale → `legacy molecule root gcg-1 is not resident in the work store: getting bead "gcg-1": bead not found`
- `gc order run` unconditional → `poured molecule root gcg-1 is not resident in the scope work store: getting bead "gcg-1": bead not found`

The rationale comment at `cmd_formula.go:858-863` and `TestFormulaCookLegacyMoleculeStaysOnTheWorkStore`'s doc are corrected, and that test is now two-sided — it asserts the work store holds no graph-marked bead as well as that the binding is empty.

### `dispatchWisp` — shrunk, not silently skipped

I fixed it, then reverted it, and the reason is worth stating: the dispatcher's v1-poured root is a **two-class object**. It carries the `order:`/`seq:` event-cursor labels, and `orders.Store` unions order-run evidence across the **ORDERS and GRAPH legs only**. Routing its molecule by class fixes the steps and **loses the cursor**, so an event-triggered v1 poured order re-fires every tick. That trades one silent breakage for another, which is the failure mode this review just caught me in.

`gc order run` has no such tension — on `main` its molecule *and* its cursor labels already live on the scope store, so gating it by class is a pure restoration, which is why it *is* fixed here.

Filed **`ga-fk1a5`** with the fix shape (persist the cursor onto the orders-class tracking bead the way the exec arm already does, then route by class), and pinned the current answer with `TestOrderDispatchPouredV1MoleculeRelocatesWithTheGraphClass` so that slice has a test to flip.

### The `--attach` arm now has split-city coverage

`cookFormula` takes variadic args, and the arm is driven by the real command on a split city:

- `TestFormulaCookAttachLeavesTheSourceBeadUnblockableOnSplitCity` — every dep endpoint resolves in the store that wrote it, **and** the source bead is back in `Ready()` after every bead in both stores is closed. That is the assertion that would have caught this.
- `TestFormulaCookAttachIsIdempotentOnSplitCity` — drives the `existing != nil` early return (convoy target ⇒ stable root key). Mutating the live-root lookup to the graph store: `re-cook minted a second workflow root (gc-2 then gc-5); the idempotent lookup read a store that does not hold the root`.
- `TestFormulaCookMetaStampLandsOnTheGraphResidentRoot` — the `--meta` root store. Mutation: `setting root metadata on gcg-1: setting metadata batch on "gcg-1": bead not found`.
- `TestEmitFormulaCookExecutionFactsReadsTheConvoyFromTheWorkLeg` — the work leg. Mutation: `no execution.work_associated fact for mc-2 in run gcg-1 (events=[])`.

One honest limitation, stated in the test's own comment rather than in prose here: the execution-fact work leg is asserted on the helper, not through the command, because **no one-shot cook can reach the two-leg state today** — `PrepareInvocation` only mints an input convoy for a targeted invocation, and the `--attach` arms (the only targeted ones) now run wholly on the scope store. The helper is where the leg assignment lives, so it is where a collapse is catchable.

### Also found and fixed while here

An earlier rebase of this branch **silently dropped #5149's `conformanceMoleculeMembership` (I12)** from `split_topology_conformance_test.go` — the 3-way merge took the branch's whole blob. Restored; the PR's diff on that file is back to the two `ga-xo8ch` → `ga-k8pzw` attribution edits it was always meant to be.

### Single-store: byte-identical, proven by mutation

`resolveClassStore`'s identity branch returning a fresh store reddens all four compatibility tests:

```
wisp root gc-1 is not resident in the single store: getting bead "gc-1": bead not found
cooked root gc-1 is not resident in the single store: getting bead "gc-1": bead not found
root-only wisp gc-1 is not resident in the single store: getting bead "gc-1": bead not found
graphStoreFor returned *beads.MemStore(0x…), want the identical store value 0x…
```

### Gates

`go build ./...` · `go vet ./...` · `gofmt -l cmd/gc internal` · `golangci-lint 2.10.1 ./cmd/gc/...` **0 issues** · `go test ./cmd/gc -timeout 25m` ok 698s · `go test ./internal/...` ok · `scripts/check-core-boundary.sh` OK · `scripts/check-routed-test-rows.sh` OK · `make check-split-topology-rows` OK

Issue #5151, opened 2026-08-09
Make temporary Git test repositories ignore user configuration


Toady00:
Several temporary Git repositories used by contributor and pre-push tests inherit the developer's global Git configuration. In particular, a global `commit.gpgsign=true` makes `scripts/test-push-gate-lock.sh` fail when its fixture overrides the identity with `Push Gate Test <push-gate-test@example.invalid>`:

```text
gpg: skipped "Push Gate Test <push-gate-test@example.invalid>": No secret key
fatal: failed to write commit object
```

This is broader than signing. Global hooks, URL rewrites, credential helpers, line-ending behavior, branch defaults, and command-level `GIT_CONFIG_*` injection can also change fixture behavior.

The repository already has the intended contract in `internal/git.HermeticEnv`: strip inherited Git routing/config environment and pin `GIT_CONFIG_NOSYSTEM=1` plus an empty global config. Shell test fixtures should use an equivalent shared setup, preferably with a fixture-owned empty config file for portability, and then configure only the identity and behavior each repository needs locally.

Acceptance criteria:

- Shell-created temporary repositories do not read system or user Git configuration.
- Inherited repository-routing and command-level `GIT_CONFIG_*` variables cannot redirect or modify fixture commands.
- Fixture identity, initial branch, signing, and hooks behavior are explicit.
- A regression test passes when the invoking user has `commit.gpgsign=true`.

Issue #5009, opened 2026-08-05
bug: demand-spawned adhoc pool seats are origin-gated out of routed work


sjarmak:
### Before you continue

- [x] I searched existing issues and did not find a duplicate.
- [x] I verified the behavior against current `origin/main` source.

### Gas City version

Verified against `gastownhall/gascity` `origin/main` at `3db4ed265db02a411b081e4581b435d80d4a5311`.

The live observation used a source build at `990ca7ab3`; the relevant origin gate is unchanged on current `origin/main`.

### Environment

- Linux
- Local managed session runtime
- Rig-scoped pool: `/home/ds/projects/background-agents/background-agents-worker`
- Graph v2 convoy: `background-agents-ubgj`

### Reproduction

1. Route a graph workflow to a pool whose demand path materializes an adhoc seat.
2. Confirm a ready step is open, unassigned, and has `gc.routed_to` set to the base pool route.
3. From outside the live session, run the hook against the base route:

   ```sh
   gc hook "/home/ds/projects/background-agents/background-agents-worker"
   ```

   This returns the ready step as JSON and exits 0.
4. Let the demand-spawned adhoc seat run its normal `gc hook`. It returns no work and the seat drains.

In the measured incident, the affected session bead had this shape:

```text
template=/home/ds/projects/background-agents/background-agents-worker
agent_name=background-agents/background-agents-worker-adhoc-a9a54b3960
session_origin=manual
```

The cycle repeated for consecutive convoy steps: a project-lead nudge made one seat claim and complete one step, that seat drained, a fresh adhoc seat appeared, and the next ready step remained open until another targeted nudge. A nudge sent to a seat that had already drained was lost.

### Expected behavior

A demand-spawned pool seat should reach the routed-pool tier of the pool template's default work query and claim ready, unassigned work addressed to the base pool route. A multi-step convoy should progress without one manual project-lead nudge per step.

### Actual behavior

The live adhoc seat sees no routed work and drains even though invoking `gc hook` explicitly for the base pool route returns the ready row. This repeats at every step boundary, replacing one drained adhoc seat with another.

### Root cause in current source

`internal/config/workquery.go` prepends `poolDemandOriginGateScript()` to the routed-pool tier:

```go
func poolDemandOriginGateScript() string {
	return `case "$GC_SESSION_ORIGIN" in ` +
		`ephemeral|"") ;; ` +
		`*) exit 0 ;; ` +
		`esac; `
}
```

[...]

PR #5152, opened 2026-08-09
Align Huma agent session origin

Toady00:
## Summary

- use one metadata helper for both configured-agent API creation paths
- persist Huma-created agent sessions with the canonical `ephemeral` origin
- verify only demand-created agent sessions can claim work routed to their base template
- document that API creation does not imply controller pool ownership

Refs #5009.

## Testing

- focused API and hook regression tests
- `make test`
- `go vet ./...`
- `make dashboard-ci`
- documentation synchronization checks
- `.githooks/pre-commit`

The pre-push wrapper was bypassed only for upload because its shell fixture inherits global `commit.gpgsign=true` and attempts to sign with a fake fixture identity. That contributor-test isolation bug is tracked separately in #5151; the relevant code gates above passed.


Toady00:
I went ahead and resolved the merge conflicts, and it looks like the resultant merge should fix the CI failure as well.

quad341:
Rebased onto current main (was CONFLICTING) with authorship preserved; two maintainer-side commits added per the review: the requirements row is renumbered to SESSION-RECON-015 (main took 014 meanwhile) and narrowed to exactly what the diff proves, and #6095 now tracks making the ephemeral origin survive reconciliation on multi-session templates — that issue blocks closing #5009, which stays open. Thanks @Toady00 — the unification itself ships as-is.

PR #5153, opened 2026-08-09
Resolve deferred drain continuation

### LINKED ISSUE #5141: bug: deferred drain continuation uses stale bead identity
### Before you continue

### Gas City version

Current `upstream/main` at `04a7c7abd`; live observation used source builds containing the same `mol-do-work` drain instructions and continuation behavior.

### Environment

macOS, local supervisor, Codex pool session, graph v2 `mol-do-work` workflow, rig-scoped Dolt store.

### Reproduction

1. Run `mol-do-work` on a pool so one session claims the `do-work` step and its dependent `drain` step is preassigned through the continuation group.
2. Let that session close the target work bead and its claimed `do-work` step.
3. The existing process receives a deferred reminder that the dependent drain bead may now be unblocked.
4. Follow the drain step instructions exactly.

The process startup environment still identifies the original work step. Running processes cannot have `GC_BEAD_ID` or `GC_TRIGGER_BEAD_ID` replaced when a deferred reminder arrives. The drain instructions currently derive the bead to close from those startup variables:

```sh
DRAIN_BEAD_ID="${GC_BEAD_ID:-${GC_TRIGGER_BEAD_ID:-}}"
```

### Expected behavior

A same-session continuation closes the newly unblocked drain bead named by the continuation, then acknowledges the drain. The workflow finalizer should proceed without replacing the session.

### Actual behavior

The drain command closes the original `do-work` bead again and calls `gc runtime drain-ack`, leaving the actual drain bead open. In the observed run, session `gl-h1dju` started for `gct-ye8`; its deferred reminder named drain bead `gct-oq8`; `Drain acknowledged.` was written to `gct-ye8` while `gct-oq8` remained open. The reconciler eventually recovered by releasing the continuation and starting fresh session `gl-az0mw`, whose startup identity correctly named `gct-oq8`. That session closed the drain and the workflow finalized.

```shell
[session] blocker gct-ye8 closed — your dependent gct-oq8 may be unblocked
# Existing process still has GC_BEAD_ID=gct-ye8
# Drain command updates gct-ye8, not gct-oq8
# Reconciler later starts gl-az0mw for gct-oq8 and recovers
```

### Additional context

This was discovered after successfully fixing and validating #5009, but it is a separate continuation-contract defect. #5009 controls whether a pool seat reaches routed work; this bug occurs after a step has already been claimed and completed.

Toady00:
## Summary

- resolve the workflow root from the running session's immutable startup bead
- select the unique ready `mol-do-work.drain` continuation for that root
- fail closed before drain acknowledgement when resolution or closure fails
- add a formula contract test that prevents reuse of stale startup identity

Fixes #5141.

## Testing

- `EXTRA_TEST_ENV="GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_NOSYSTEM=1" make test`
- `GIT_CONFIG_GLOBAL=/dev/null make vet`
- `GIT_CONFIG_GLOBAL=/dev/null make lint-new LINT_BASE=upstream/main`
- `git diff --check upstream/main...HEAD`

The pre-push wrapper was bypassed only for upload because its shell fixture inherits global `commit.gpgsign=true` and signs with a fake fixture identity. The fixture-isolation defect is tracked in #5151; the code gates above passed.

### DIFF
--- a/internal/bootstrap/packs/core/formulas/mol-do-work.toml
+++ b/internal/bootstrap/packs/core/formulas/mol-do-work.toml
@@ -154,7 +154,7 @@ if [ -z "$ROOT_BEAD_ID" ]; then
   echo "mol-do-work drain could not resolve its workflow root from $STARTUP_BEAD_ID" >&2
   exit 1
 fi
-READY_BEADS=$(gc bd ready --json --limit=2 --metadata-field "gc.root_bead_id=$ROOT_BEAD_ID" --metadata-field "gc.step_ref=mol-do-work.drain") || exit 1
+READY_BEADS=$(gc bd ready --json --limit=2 --include-ephemeral --metadata-field "gc.root_bead_id=$ROOT_BEAD_ID" --metadata-field "gc.step_ref=mol-do-work.drain") || exit 1
 DRAIN_BEAD_ID=$(printf '%s' "$READY_BEADS" | jq -r 'if length == 1 then .[0].id else empty end')
 if [ -z "$DRAIN_BEAD_ID" ]; then
   echo "mol-do-work drain could not resolve one ready continuation bead for root $ROOT_BEAD_ID" >&2
--- a/internal/bootstrap/packs/core/formulas/mol-do-work.toml
+++ b/internal/bootstrap/packs/core/formulas/mol-do-work.toml
@@ -148,17 +148,25 @@ if [ -z "$STARTUP_BEAD_ID" ]; then
   echo "mol-do-work drain requires a startup bead id to resolve its workflow root" >&2
   exit 1
 fi
-STARTUP_BEAD=$(gc bd show "$STARTUP_BEAD_ID" --json) || exit 1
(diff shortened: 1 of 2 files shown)


quad341 (dismissed):
@Toady00 — this is a clean fix for #5141, and the diagnosis is the good part: separating the process's *immutable startup bead* from the *continuation that became ready while the process was running* is exactly the right frame, and it is not an obvious one. Gating the `gc hook current` branch on `gc.step_ref` is the detail that makes it hold on the deferred path, and I traced it through the compiler to confirm — `flattenSteps` builds the ref as `formula + "." + step.ID`, so `mol-do-work.drain` is right. Moving the `|| exit 1` in front of `gc runtime drain-ack` is the other half, and your test pins that ordering explicitly.

The red CI is not yours. `cmd/gc` didn't compile at your merge-base because a helper signature changed under two out-of-diff test call sites; `main` fixed it in #5104 about two and a half hours after your run. I rebuilt your two files on current `main` locally — vet is clean and your new test passes. So I'll rebase this onto `main` on my side and re-run rather than asking you to touch anything.

Noted on the #5151 fixture-signing workaround too. Thanks for the careful writeup.


---


Review coverage:
- completed: `qwen codex`
- unavailable: `claude`

PR #5154, opened 2026-08-10
Preserve hook claim store context

### LINKED ISSUE #5142: bug: hook claim mutations can inherit an ambient backend
### Before you continue

### Gas City version

Current `upstream/main` at `04a7c7abd`; originally observed while validating a source build based on `d2f91a549`.

### Environment

macOS, zsh/login-shell environment, local supervisor, rig-scoped Dolt store. The login environment exported a Dolt server host that was intentionally absent from the selected hook query environment.

### Reproduction

1. Export an ambient backend endpoint such as `BEADS_DOLT_SERVER_HOST=ambient.example`.
2. Build a hook query environment that intentionally scrubs that host and selects a local or different rig store.
3. Run `gc hook --claim` so the query finds ready work in the selected store.
4. Observe the subsequent claim mutation.

`hookClaimEnvMap` copies only keys present in the already-scrubbed query environment. `BdStore` then applies that map as overrides on top of the parent process environment. Because the scrubbed host is absent rather than represented by an explicit empty override, the parent value reappears for the mutation. The query and claim can therefore address different stores.

### Expected behavior

Every claim, work-branch, and continuation mutation remains bound to the exact backend context used by the successful hook query. Scrubbed backend endpoint variables remain cleared.

### Actual behavior

An ambient Dolt or PostgreSQL endpoint can re-enter between query and mutation. The query returns work from the intended store, but the claim update is redirected to the ambient backend and either fails or mutates a different database.

```shell
# Query environment: BEADS_DOLT_SERVER_HOST absent by design
# Parent environment: BEADS_DOLT_SERVER_HOST=<ambient server>
# Mutation runner: parent environment overlaid with hookClaimEnvMap
# Result: ambient host survives because no explicit empty override exists
```

### Additional context

This is distinct from #4304 / PR #4305. That change routes city-scoped session-pointer writes back to the city store after a rig claim. This bug affects the selected work-store claim mutation itself when a scrubbed backend key leaks back from the parent process. It was discovered while validating #5009 but is not part of that origin-gating defect.

Toady00:
## Summary

- execute hook claim mutations with the exact scrubbed query environment
- prevent absent Dolt or PostgreSQL selectors from reappearing from the parent process
- retain the existing layered runner for callers that intentionally pass overrides
- cover both environment projection and real subprocess exclusion behavior

Fixes #5142.

## Testing

- focused `cmd/gc` hook-claim tests
- focused `internal/beads` exact-runner test
- `EXTRA_TEST_ENV="GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_NOSYSTEM=1" make test`
- `GIT_CONFIG_GLOBAL=/dev/null make vet`
- `GIT_CONFIG_GLOBAL=/dev/null make lint-new LINT_BASE=upstream/main`
- `git diff --check upstream/main...HEAD`

The pre-push wrapper was bypassed only for upload because its shell fixture inherits global `commit.gpgsign=true` and signs with a fake fixture identity. The fixture-isolation defect is tracked in #5151; the code gates above passed.


quad341 (approved):
@Toady00 — thank you for this one, and specifically for how you built it. Threading a `baseEnvFn` through `execCommandRunner` instead of forking a second copy of the runner meant the three existing constructors kept byte-identical behavior while the new exact-env path came in at five lines. That's the version of this fix that doesn't create a maintenance problem.

`TestExecCommandRunnerWithExactEnvContextExcludesParentValues` is also the right kind of test — a real subprocess with a poisoned ambient `BEADS_DB` and credential command, not a mock asserting a map. And the doc comment on `hookClaimEnvMap` is genuinely good: naming the nil-env consequence *and* saying which production callers make it unreachable is the note the next person needs.

I traced the env chain to check the invariant you're leaning on, and it holds — the work query already runs under an exact-replaced env, so your change makes the mutation symmetric with the read that selected the store. That's the real argument for this fix.

The red shard isn't yours: `rest-smoke-2-of-2` is dying in `gc init` on a bd schema migration. I'll re-run it, and add a small assertion pinning the runner wiring so a future revert of that one line trips a test instead of quietly bringing #5142 back. Merging with credit.


---

PR #5156, opened 2026-08-10
fix(herdr): make wisp-id tab labels prefix-agnostic

AJBcoding:
## Problem

herdr tabs for pool agents read `polecat-az-wisp-z0pu0` instead of the intended
`polecat-furiosa`. The alias substitution that produces the readable name already
exists and is correct — it just never fires outside a default city.

## Root cause

A wisp id is a bead id minted by `PourWisp`, so it carries the beads prefix of the
scope that poured it: `<prefix>-wisp-<suffix>`. Upstream's default city prefix is
`gc`, giving `gc-wisp-`, and `placementFor` hardcoded exactly that literal:

```go
if i := strings.Index(tab, "gc-wisp-"); i >= 0 {
```

with the same literal in its guard. On any city whose HQ scope prefix is not
literally `gc`, `strings.Index` returns -1, the substitution is skipped silently,
and the raw wisp id survives into the tab label. Observed on two separate cities
whose prefixes are `az` and `py`.

This is cosmetic but city-wide: it affects every pool agent's tab, not just one rig.

## Fix

Key on the `-wisp-` infix and walk back over the prefix token:

```go
func wispIDStart(s string) int {
	i := strings.Index(s, "-wisp-")
	if i < 0 {
		return -1
	}
	return strings.LastIndex(s[:i], "-") + 1
}
```

Two details worth flagging for review:

- **Matching `-wisp-` directly is not sufficient.** `placementFor` keeps `tab[:i]`
  as the role prefix, so cutting at the infix would yield `polecat-az` + leaf.
  The walk-back finds the start of the prefix token instead.
- **`LastIndex` returning -1 handles the no-role-prefix case for free**, giving
  index 0 so a bare `az-wisp-z0pu0` is replaced whole.

The guard that rejects an alias which is itself the wisp identity carried the same
hardcoded prefix and needed the same treatment. Without it, the deferred pool spawn
(which blanks `GC_ALIAS` and leaves the session name in `GC_AGENT`) produced
`polecat-gastown__polecat-az-wisp-z0pu0`.

Used `strings` rather than a regexp to match this package's production idiom —
`regexp` currently appears only in `agentname_test.go`.

## Tests

Four cases added to the existing `TestPlacementFor` table, using real observed
session data:

- non-`gc` city prefix gets the themed tab (`az-wisp-z0pu0` -> `polecat-furiosa`)
- wisp id with no role prefix is replaced whole (-> `furiosa`)
- deferred spawn falls back to `GC_AGENT` when `GC_ALIAS` is blank
- guard: an alias that is itself the wisp identity is ignored

The guard case is a regression guard and passes before the change. I verified it
has teeth by fixing only the detection first, which made it fail with exactly the
corruption above.

Behavior is unchanged for default-city `gc-wisp-` ids, persistent rig agents, and
town-level agents; all pre-existing cases still pass.

## Note for reviewers

`placementFor` is consulted only at `Provider.Start`, so this renames no existing
tab — labels correct themselves as agents respawn. A retroactive sweep is possible
via `client.tabRename` but is deliberately out of scope here.

Verified: `internal/runtime/herdr` passes, `go vet` clean, lint clean.

### DIFF
--- a/internal/runtime/herdr/placement_test.go
+++ b/internal/runtime/herdr/placement_test.go
@@ -50,6 +50,44 @@ func TestPlacementFor(t *testing.T) {
 			wantTab:  "polecat-gc-wisp-abc123",
 		},
 		{
+			// The wisp id is a bead id minted by the pouring scope, so it carries
+			// that scope's beads prefix — "gc-" only when the city's HQ prefix is
+			// literally "gc". Detection must be prefix-agnostic or every other
+			// city keeps raw wisp ids in its tabs.
+			name:     "wisp with non-gc city prefix gets themed tab",
+			sessName: "gastown__polecat-az-wisp-z0pu0",
+			env:      map[string]string{"GC_RIG": "gascity", "GC_ALIAS": "gascity/gastown.furiosa"},
+			wantWS:   "gascity",
+			wantTab:  "polecat-furiosa",
+		},
+		{
+			name:     "wisp id with no role prefix is replaced whole",
+			sessName: "gastown__az-wisp-z0pu0",
+			env:      map[string]string{"GC_RIG": "gascity", "GC_ALIAS": "gascity/gastown.furiosa"},
+			wantWS:   "gascity",
+			wantTab:  "furiosa",
+		},
+		{
+			// The deferred pool spawn blanks GC_ALIAS and leaves the session name
+			// in GC_AGENT (build_desired_state_pool_info.go); once the alias
+			// resolves the reconciler repopulates it.
+			name:     "non-gc wisp falls back to GC_AGENT when alias not yet resolved",
+			sessName: "gastown__polecat-az-wisp-g8ha3",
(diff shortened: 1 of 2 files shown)


quad341 (approved):
@AJBcoding — this is a genuinely good bug report *and* a good fix. Tracing raw `polecat-az-wisp-z0pu0` tabs back to a single hardcoded `"gc-wisp-"` literal, and then noticing that the *guard* carried the same literal and would have produced `polecat-gastown__polecat-az-wisp-z0pu0` on the deferred pool spawn, is the part most fixes would have missed. The walk-back in `wispIDStart` is the right shape, and `LastIndex` returning -1 giving you index 0 for a bare wisp id is a nice bit of luck you documented rather than left for the next reader to rediscover.

I re-ran `TestPlacementFor` and `go vet` on your head — 12/12 green, vet clean — and hand-checked that the cut index is identical to the old literal for default-city ids, so existing cities are untouched.

Merging as-is. Agreed on leaving the retroactive `client.tabRename` sweep out of scope; labels correcting themselves on respawn is the right trade. Thanks for the observed-session test data — it made this easy to verify.


---

PR #5158, opened 2026-08-10
feat(cli): federate `gc ready` across a split city's stores (ga-oxsyu)

julianknutsen:
Closes the **CLI half** of the split-store graph blindness: `ga-oxsyu`. The API
half is `ga-4k2c3` / #5148, whose **FEDERATION CONTRACT** this slice is
specified against.

## The bug, measured

On a split city the graph-class DAG — `gcg-` molecule roots, step beads, control
beads — lives in the relocated graph store, and `bd ready` in the work directory
cannot reach it. Measured on a live city, same moment:

| surface | beads | of which `gcg-` |
| --- | --- | --- |
| `gc bd ready` | **5** | **0** |
| `GET /v0/beads/ready` | **22** | **14** |

The CLI was blind in bulk behind an answer that looked authoritative — and a
work query that returns a short array is indistinguishable from one that returns
*"no work"*.

`gc ready` is the in-process, city-wide drop-in for that single-store query. It
federates the **city store, the rig stores, and — only when the class is
actually relocated — the graph store**.

## The acceptance test is CLI == API, not an invented oracle

#5148 deliberately left an executable contract on `humaHandleBeadReady` written
for this slice. `I13-cli-ready-federation` in the split-store conformance suite
runs the **real API handler** (through `api.NewSupervisorMux`, over a
`controllerState` built from the fixture, so `BeadStores()` / `CityBeadStore()` /
`GraphBeadStore()` resolve through production dispatch) and the **real CLI
reader** over the **same three stores**, then compares:

* **legs, in order:** city → rigs ascending → graph **LAST**
* **per-leg order:** whatever that leg's own reader emits — deterministic, but
  **not canonical across leg kinds**
* **dedupe:** first leg to return an id wins
* **both sides normalized** with `beads.SortBeadsReadyOrder` before comparing.
  Load-bearing, not cosmetic: the canonical relocated binding is a
  `beads.SQLiteStore` ordering by `(created_at, id)` with **no priority term**.

**Red-before, on the split row, with the CLI's graph leg dropped:**

```
gc ready = [gc-1 ra-1] but GET /v0/beads/ready = [gc-1 ra-1 gcg-wisp-0001 gcg-2]
over the same stores; the CLI and the API disagree about what is claimable
```

That is the field bug, reproduced by mutation.

## The four branch defects are NOT carried

An earlier branch (`feat/domain-infra-store-split`) carried a `gc ready` with
four confirmed defects. Each is forbidden by a test that is **red when the
defect is reintroduced** — exact strings:

**1 · Inverted metadata filter.** A bare `--metadata-field key` mapped onto
equality-against-`""`, so it matched every bead *missing* the key and rejected
every bead *carrying* it. Never fired because the generated work_query always
passes `key=value`.

```
--metadata-field gc.routed_to returned [gc-2], want exactly [gc-1]; gc-2 carries
no such key and must not match (inverted metadata filter)
```

The three forms are now distinct parsed predicates, not map entries: bare `key`
(present, non-empty), `key=value` (exact), `key=` (present and empty).

**2 · `--status` silently misread.** Only `in_progress` was special-cased;
everything else fell through to ready work, so `--status closed` answered with
**open** beads.

```
--status closed returned [gc-1], want [gc-2] — a status flag that answers with
ready work is answering a different question
readyBeadsForOpts error = <nil>, want a rejection naming the unknown status
```

`--status` is a real selector over a **closed** set now; an unrecognized value is
rejected rather than answered with an empty array (which is the same fail-open
one layer up).

**3 · Limit push-down.** `q.Limit` went into every leg and the merge was
re-limited, so the bound cut each store's own read order before the merged order
existed.

```
--limit 2 returned [gc-2 gc-1], want [gc-2 gc-3] — gc-1 is lower priority than
both and must not survive a bounded read (limit pushed into the leg?)
--sort oldest --limit 2 returned [gc-2 gc-3], want [gc-1 gc-2] — the oldest row is
last in the leg's own canonical order, so a bound pushed into the leg drops it
```

The second case is the shape the pool-demand work_query actually builds
(`--sort oldest --limit 20`); its leg double answers in **canonical order and
honors its own LIMIT**, the way a production work store does, so the assertion
is not vacuous against the in-memory kit leaf. The bound is now applied **once,
last**, after the merge and after the order.

**4 · Domain type on an external wire.** It marshaled `[]beads.Bead`.

```
gc ready emits []Bead from "github.com/gastownhall/gascity/internal/beads"; the
external bd-compatible array must be a wire type this package owns, not a domain
type whose fields move for internal reasons
```

**Also, two contract rules the branch got wrong:**

*Dedupe direction.* The branch's fan-out was work-first by construction. This one
is **first-leg-wins with the graph leg LAST**, which is #5148's rule, so a bead
co-resident in the work store and the binding — the documented steady state of a
migrated city, where `gc storage migrate` preserves ids and never deletes back —
resolves to the work store's row on **both** surfaces.

```
co-resident gc-1 resolved to "migrated graph copy", want the work store's row —
the API's first-leg-wins rule puts the graph leg last, and CLI == API depends on
matching it
leg order = [graph city rig rig-A rig rig-B], want [city rig rig-A rig rig-B graph]
```

> ⚠️ **Named tension, decided deliberately.** The brief also said the dedupe must
> not be work-first, "so during a partial migration the stale work copy silently
> wins over the authoritative infra copy". Under first-leg-wins-with-graph-last
> the work copy *does* win — that is exactly #5148's documented rule, and the
> acceptance criterion is CLI == API. The two instructions cannot both hold;
> I matched #5148 and am flagging it rather than silently diverging. **If the
> co-residence tie should resolve to the binding, it has to change on BOTH
> surfaces in one slice**, and that is a production behaviour change beyond this
> reader.

*The ordering law is taken, not re-derived.* `beads.SortBeadsReadyOrder` already
exists in `internal/beads` next to the unexported `sortBeadsReadyOrder`; the
branch had re-derived a second copy across the package boundary. A copy that is
one edit from the real thing (nil priority defaulting to 0 instead of
`COALESCE(priority, 2)`) is red:

```
merged order = [gc-1 gcg-1 gc-2], want [gcg-1 gc-1 gc-2]
(beads.SortBeadsReadyOrder over the same rows)
```

## Wire contract, decided deliberately

`gc ready` emits a **bare JSON array in bd's row shape**, published at
`schemas/ready/result.schema.json` — closed field set, `type: array`.

* **Field names are bd's**, because consumers decode this array straight into
  `[]beads.Bead`: **`issue_type`**, and **`parent`** — *not* the `parent_id` that
  `gc bd dep tree --json` emits and that `bdStoreBridgeBead` carries. gc's own
  JSON surfaces do not agree with each other, so this is a choice, not a default.
  Copying `bdStoreBridgeBead` verbatim would have produced a row that decodes
  with an empty parent on every consumer gc already has.
* **Field set is the HTTP `Bead` schema's**, so the two surfaces stay comparable.
  `parent` looked absent in the live capture only because it is `omitempty` and
  most beads have no parent; it is part of the contract. A test fails if
  `beads.Bead` and the wire diverge, so the follow is a decision.
* **The bare array is a sanctioned exception, not an exemption.** The branch
  reached the same end by adding a `gc.json.raw_passthrough` annotation that
  generalized `isBDCommandPath` across all three JSON-contract gates — a change
  to the contract layer itself. This PR instead uses the repo's existing escape
  hatch: `x-gc-raw-json: true` in the schema plus the explicit path in
  `TestJSONResultSchemasRequireSuccessDiscriminator`, whose own comment demands
  every exception be "explicit and self-describing so another raw result schema
  cannot bypass the top-level success discriminator silently". Remove the marker
  and it is red:

```
result schemas missing required top-level ok:true discriminator:
schemas/ready/result.schema.json
```

## Failure handling deliberately diverges from the API — at BOTH ends

The API degrades a dead **rig** leg to a Partial 200 because an HTTP body has
`partial_errors` to say so honestly. A CLI array has nowhere to say "this is
short", so every leg here **fails loud**, naming the leg.

```
a dead leg produced a successful answer; a short array is indistinguishable from
"no work"
```

**Review fix (`4b44dc5`): that was only true for READ failures.** A leg can
break in two places, and the OPEN end was a hole that reproduced the exact
fail-open this command exists to close.

`cmdReady` built the rig legs with `buildStandaloneRigStores` — the
**controller's** opener, which prints to stderr and `continue`s. A rig whose
store could not be opened was dropped before a single read ran, so
`federateBeadLegs` never saw it. On a file-provider city with rigs
`good`/`broken` where `broken/.gc` is a regular file:

```
cmdReady exit = 0
stdout = [{"id":"gc-1","title":"city work",...}]
stderr = gc supervisor: rig bead store "broken": stat .../.gc/beads.json: not a directory
```

Exit 0, a valid-looking short array, and the failure only on stderr — under
**the wrong command's name**, on a stream no work query parses.

**It was also a CLI-vs-API divergence, i.e. this PR's own acceptance
criterion.** Over the identical on-disk city `controllerState.buildStores`
returns the broken rig **present**, as an `unavailableStore` (`api_state.go`),
so `humaHandleBeadReady` records it and serves **Partial 200 + partial_errors
naming the rig**. The header above described neither surface.

*The fix.* `buildStandaloneRigStores` and the new `readyRigLegStores` now share
one opener, `openStandaloneRigStores`, which returns the stores plus the rigs
that failed **and decides nothing**. The policy is the caller's, and the two
callers disagree on purpose:

| caller | on a rig that will not open | why |
| --- | --- | --- |
| controller (`city_runtime` boot + reload, `cmd_start`, `cmd_session close`) | warn on stderr, keep the rest | halting a city over one unmounted rig is worse than running degraded, and it has a log to say so |
| `gc ready` | fail the query, naming **every** dead rig | its whole output is a bare array with nowhere to say it is short |

All four controller call sites keep byte-identical behaviour — warning text,
prefix and order included — pinned by `TestRigStoreOpenPolicyDiffersByCaller`,
which asserts both policies over one broken city. `gc ready` no longer emits
the `gc supervisor` prefix at all, because it no longer calls that opener.

**Red-before, the open half of `TestReadyFailsLoudWhenALegErrors`:**

```
gc ready exited 0 with the "broken" rig leg unopened;
stdout=[{"id":"gc-1","title":"city work",...}] — a short array is
indistinguishable from "no work", and stderr is not part of the answer any
work query parses
```

**The CLI == API correspondence is now pinned, not left to drift.** `I13`
gains `conformanceCLIReadyDeadRigLeg`, on both topologies: over the SAME dead
rig leg — an `unavailableStore`, which is the API's *own* representation of a
failed rig open — the API must serve **Partial 200 naming the rig** and the CLI
must **fail naming the rig**. That is the one place the two surfaces are
allowed to answer differently, and it is now a decision with a test on it.
Red under mutation (degrade `federateBeadLegs`):

```
gc ready served 4 beads over the same dead "rig-DEAD" leg; the API said that
read was PARTIAL, and a bare array has nowhere to say so — the CLI's
equivalent of Partial 200 is a non-zero exit
```

*The configured-but-**unbound** rig (empty `rig.Path`) stays skipped, decided
deliberately.* It is not a leg the city failed to reach — it is a rig with **no
store at all**, so no claimable row hides behind it, and `api_state.buildStores`
skips it identically, so on that arm CLI == API today and the behaviour is
pre-existing. Erroring may be the better product answer, but it is a behaviour
change beyond this fix and it has to land on **both** surfaces at once.
`TestReadyUnboundRigIsSkippedOnBothSurfaces` runs the CLI opener and the API's
`buildStores` over one config and asserts they agree, so a future change fails
until it is made on both. Red under mutation (drop the skip):

```
gc ready opened a leg for the unbound rig (map[unbound:...]); an empty path
resolves to the CITY scope, which would federate the city store twice under a
rig's name
```

The `ready_federation.go` header now states what **both** surfaces actually do,
leg by leg: rig open-or-read failure → Partial 200; graph leg → 503; unbound rig
→ skipped.

## Single-store cities

The graph leg is gated on **store identity** — the same rule the API's
`relocatedGraphStore` and `cmd/gc`'s `resolveClassStore` use — so a city that
relocates nothing federates exactly the legs it always had. Both conformance
rows run, and the gate has its own mutation test:

```
a binding identical to the work store produced a second leg; that is the same
store twice
single-store ready = [...], want exactly [gc-1]; a legacy city must not gain a leg
```

The end-to-end test runs on a real on-disk file-provider city and doubles as the
on-disk first-leg-wins proof, using legacy file mode's own id aliasing (the file
provider mints `gc-<n>` per scope regardless of a rig's configured prefix — the
same aliasing `humaHandleBeadReady` names) rather than a contrivance.

## Conformance suite

`I13-cli-ready-federation` is **added**, two-topology via `forEachTopologyWithRig`
(`scripts/check-split-topology-rows.sh` passes). No existing KNOWN GAP became
closable — I1/I2 (HQ work store absent from both arms of the controller scan),
I5 (`gc hook --claim` off the shared resolver) and I10 (wake-filter class
reachability) are all about other paths — and no new gap was found, so no skip
was added.

The `splittest` FIDELITY GAP (every kit store type-asserts as `beads.Counter`
while the production class binding implements none) does not bite here: nothing
in this reader branches on `Counter`, precisely because the bound is never pushed
into a leg.

## Not done here, on purpose

Nothing wires `gc ready` into the generated work_query or pool demand — that is
`ga-bvdha`, which this bead blocks. This PR adds the reader and its contract; it
changes no existing behaviour.

## Gates

Re-run in full on `4b44dc5` (serially):

`go build ./...` · `go vet ./...` · `gofmt` · `golangci-lint run ./...` (2.10.1)
**0 issues** · `go test ./cmd/gc -timeout 25m` **ok 713.9s** ·
`go test ./internal/...` **ok (144 packages)** ·
`scripts/check-core-boundary.sh` OK · `make check-split-topology-rows` OK.

On the base commit `9699da0`: `scripts/check-generated-docs-drift.sh` fresh ·
`gen-command-census -check` OK (this fix adds no command, flag or schema).

Nothing on the HTTP wire moved, so no OpenAPI or dashboard regeneration.


julianknutsen:
## Review fix pushed: `4b44dc5` — the rig leg that could not be OPENED

Two reviewers flagged one MAJOR and the verifier found it is worse than
reported. Fixed, with the verifier's framing carried into the code.

**The defect.** This file's header said *"Every leg here therefore fails
LOUD."* That was true only for READ errors on an already-opened leg
(`federateBeadLegs`). Leg **OPEN** failures never reached it: `cmdReady` built
the rig legs with `buildStandaloneRigStores`, the *controller's* opener, which
prints to stderr and `continue`s. Reproduced on a file-provider city with rigs
`good`/`broken` where `broken/.gc` is a regular file:

```
cmdReady exit = 0
stdout = [{"id":"gc-1","title":"city work",...}]
stderr = gc supervisor: rig bead store "broken": stat .../.gc/beads.json: not a directory
```

Exit 0, a valid-looking short array, and the only signal on stderr under the
**wrong command's name**.

**And it was a CLI-vs-API divergence — this PR's own acceptance criterion.**
Over the identical city, `controllerState.buildStores` returns the broken rig
*present*, as an `unavailableStore`, so `humaHandleBeadReady` records it and
serves **Partial 200 + partial_errors naming the rig**. The header described
neither surface.

The moment `gc ready` backs the pool-demand/worker `work_query` (`ga-bvdha`,
which this bead blocks) that short array is indistinguishable from "no work" —
the pool drains or the worker idle-exits while claimable work sits in the dark
rig.

**The fix.** `buildStandaloneRigStores` and the new `readyRigLegStores` share
one opener, `openStandaloneRigStores`, which returns the stores plus the rigs
that failed and **decides nothing**. The policy is the caller's, and the two
callers disagree on purpose: the controller warns and keeps supervising the
rigs that are up; `gc ready` fails the query, naming every dead rig
(`rig %q store: %w`). All four controller call sites — `city_runtime` boot and
reload, `cmd_start`, `cmd_session close` — keep byte-identical behaviour,
warning text and order included. `gc ready` no longer emits the `gc supervisor`
prefix at all, because it no longer calls that opener.

**Unbound rig, decided deliberately: still skipped.** A rig declared in
`city.toml` with no binding has an empty `rig.Path` and *no store* — nothing
claimable hides behind it — and `api_state.buildStores` skips it identically,
so on that arm CLI == API today and the behaviour is pre-existing. Erroring may
be the better product answer, but it is a behaviour change beyond this fix and
must land on both surfaces at once. Pinned by a test over both surfaces rather
than left in a comment.

**Tests, with red-before strings.**

`TestReadyFailsLoudWhenALegErrors` gains the OPEN half, on a real on-disk city:

```
gc ready exited 0 with the "broken" rig leg unopened;
stdout=[{"id":"gc-1","title":"city work",...}] — a short array is
indistinguishable from "no work", and stderr is not part of the answer any
work query parses
```

`I13` gains `conformanceCLIReadyDeadRigLeg` on both topologies, which pins the
correspondence instead of leaving the surfaces to drift: over the SAME dead rig
leg (an `unavailableStore` — the API's own representation of a failed rig open)
the API must serve Partial 200 naming the rig, and the CLI's equivalent is a
non-zero exit naming the rig. Red under mutation (degrade `federateBeadLegs`):

```
gc ready served 4 beads over the same dead "rig-DEAD" leg; the API said that
read was PARTIAL, and a bare array has nowhere to say so — the CLI's
equivalent of Partial 200 is a non-zero exit
```

`TestReadyUnboundRigIsSkippedOnBothSurfaces`, red under mutation (drop the
skip):

```
gc ready opened a leg for the unbound rig (map[unbound:...]); an empty path
resolves to the CITY scope, which would federate the city store twice under a
rig's name
```

`TestRigStoreOpenPolicyDiffersByCaller` holds the two policies apart over one
broken city, now that they share an opener.

The `ready_federation.go` header now states what **both** surfaces actually do,
leg by leg: rig open-or-read failure → Partial 200; graph leg → 503; unbound rig
→ skipped.

**Gates, re-run in full and serially on `4b44dc5`:** `go build ./...` ·
`go vet ./...` · `gofmt` · `golangci-lint run ./...` (2.10.1) **0 issues** ·
`go test ./cmd/gc -timeout 25m` **ok 713.9s** · `go test ./internal/...`
**ok (144 packages)** · `scripts/check-core-boundary.sh` OK ·
`make check-split-topology-rows` OK. Nothing on any wire moved.

PR #4038, opened 2026-07-07
simplify(S36): route drift-relaunch through buildPreparedStart (one launch-config derivation)

julianknutsen:
Routes the #3872 drift-relaunch path through buildPreparedStart so launch config is derived once, fixing the drift-relaunch misconfig and collapsing 2 launch-config derivations into 1.

- Fixes the drift-relaunch misconfiguration from #3872
- Collapses two launch-config derivations to one (buildPreparedStart is now the single source)
- Quality gates green
- Fable-reviewed, behavior-preserved
- Spec: /data/projects/gascity/.claude/worktrees/simplification/engdocs/simplification/specs/S36-drift-relaunch-preparedstart-spec.md

NOTE: spike/staged for review, not auto-merge.

PR #5160, opened 2026-08-10
fix(session): reach the fresh-start recovery when session_key is empty

austinborn:
## Summary

- A session whose bead has an empty `session_key` could never reach the fresh-start recovery, so a start that died during startup was retried unchanged forever. Both `ErrSessionDiedDuringStartup` branches in `internal/session/chat.go` gated the recovery on `session_key != ""`, and `retryFreshStartAfterStaleKey` refused again on its own first line.
- Removes that gate at both start call sites and lets the empty-key case fall through to the value-agnostic strip the function already carries, so the doomed command is relaunched without its resume shape.
- Adds a guard so the newly reachable path never relaunches a command it did not change, and makes both call sites propagate the original start error when the recovery declines.

Measured on a hosted city before the fix: 87 start attempts in one hour, each ending the same way, `provider_error duration=45s err=resuming session: session died during startup`. It'd been looping like that for over nine hours. `wake_mode` was already `fresh`, so there wasn't a configuration knob that would have prevented it.

## Why the recovery was unreachable

```mermaid
flowchart TD
    A[Start dies during startup] --> B{bead session_key non-empty?}
    B -- yes --> C[retryFreshStartAfterStaleKey]
    C --> D[strip resume shape, relaunch fresh]
    B -- no --> E[return &quot;resuming session&quot; error]
    E --> F[supervisor retries the identical command]
    F --> A
```

That right-hand loop is the bug, and it can't converge on its own. The one case where a fresh start is unambiguously correct, because there's nothing left to resume, is the one case that couldn't reach the code written to perform it.

## Prevention versus recovery

The obvious fix is prevention: don't emit a resume flag when the key is empty. Every command-construction site already does that, though. `BuildResumeCommand` returns the plain stored command when `SessionKey` is empty, and the supervisor only calls `resolveSessionCommand` when the key is non-empty.

So the doomed command doesn't come from construction. It comes from divergence. The command is built while the key is present, the key is cleared before the start actually runs, and the launch still carries `--resume <key>` while the bead reads empty. Prevention can't reach that case, because construction saw a valid key. Recovery is the only place the fix can land, which is why this change takes the narrower route.

That's also why the existing value-agnostic strip is the right tool. `stripResumeFlagArg` was written for exactly this divergence, and #4038 fixed the neighbouring relaunch path on the same reasoning: a speculative key must never be executed as `--resume`. The start path just wasn't covered.

## Not relaunching what we did not change

If the strips leave the command untouched on the empty-key path, it carries no resume shape at all, and relaunching it verbatim repeats the same failure at the same cost. The recovery reports "not retried" there and the caller propagates the original error, so the fix can't become a cheaper version of the loop it removes.

Paths that still hold a key keep their previous behaviour exactly. There the metadata clear and the orphan sweep are themselves the recovery, so an unchanged command is still worth relaunching.

## Test plan

- [ ] `TestEnsureRunning_EmptySessionKeyStripsResumeAndStartsFresh` — a start that dies with an empty bead `session_key` now relaunches, and the relaunch carries neither the resume flag nor the stale key. Reverting `chat.go` alone fails this test with the production error string, `resuming session: session died during startup`.
- [ ] `TestEnsureRunning_EmptySessionKeyWithoutResumeShapeDoesNotRelaunch` — with nothing to strip, exactly one launch is attempted and the error wraps `ErrSessionDiedDuringStartup`.
- [ ] `TestStartRuntimeOnly_EmptySessionKeyStripsResumeAndStartsFresh` — the same recovery, on the second call site. `ensureRunning` and `ensureRunningRuntimeOnly` carry their own copies of the branch, and the reconciler reaches the runtime-only one: `startPreparedStartCandidate` calls `StartResolved`, which lands on `StartRuntimeOnly`. That's the path the supervisor retries, so it's the path the measured loop actually ran on. Reverting `chat.go` alone fails this test with the same production error string.
- [ ] `TestStartRuntimeOnly_EmptySessionKeyWithoutResumeShapeDoesNotRelaunch` — the runtime-only twin of the decline case. The "nothing to strip" branch is duplicated per call site rather than shared, so a `Send`-driven test can't reach this copy of it.
- [ ] `go test ./internal/session/...` passes.
- [ ] `go test ./cmd/gc/...` passes.

Both `DoesNotRelaunch` cases pass with or without the fix, deliberately. They pin behaviour preservation on the path this change leaves alone, so it's only the two `StripsResumeAndStartsFresh` cases that fail on a revert.

Note for macOS reviewers: the Dolt `go-icu-regex` cgo dependency needs ICU headers on the include path, the same `unicode/regex.h` failure `fork-verify.yml` documents. Point `CGO_CPPFLAGS` and `CGO_CXXFLAGS` at your `icu4c` include directory if you want to run the suites locally.

---
> • City: `factory-main` · Agent: `local-core.builder-1`
> • On behalf of: @austinborn


tudorsaitoc:
Refreshed replacement: #5269. The original head repository does not grant push access to `tudorsaitoc`, so #5269 preserves this recovery on current upstream main with fresh verification and CI.

austinborn:
Closing this in favour of [#5269](https://github.com/gastownhall/gascity/pull/5269), which carries the same fix from a branch its author can push to.

The two patches are the same code. Diffing both with the index lines stripped, `internal/session/chat.go` and `internal/session/chat_empty_session_key_test.go` come out byte-identical across the two pull requests: 53 added and 18 removed, then 207 added, on both sides. #5269 adds the one thing this branch is missing, the `SESSION-START-009` row in `internal/session/REQUIREMENTS.md`, which is the traceability convention this repo expects when a new behaviour lands. The open version is the more complete one.

This pull request was also still a draft, so it was never the one that would merge.

The branch stays where it is. Only the pull request closes.

---
> • City: `city` · Agent: `local-core.manager`
> • On behalf of: @austinborn

PR #3728, opened 2026-06-25
fix(hook): raise work-query timeouts so pool operators aren't starved

julianknutsen:
## Problem

On a busy multi-rig dolt city, pool operators (`gc.run-operator`) were being **starved of work they had already been routed**. A routed work step would sit `open`/unclaimed while the operator's work-probe failed `work query timed out` every cycle, stalling the whole adopt-pr pipeline (the molecule materializes + routes its first step, but no operator ever claims it).

**Root cause:** the default work-probe (`config.Agent.EffectiveWorkQuery`) issues **~6 sequential bd/store round-trips** — three session identifiers (`$GC_SESSION_ID`/`$GC_SESSION_NAME`/`$GC_ALIAS`) across the in-progress and ready *assigned* tiers (crash-recovery priority) — **before** the pool-demand tier that finds routed work. On a multi-rig dolt city under concurrent load each `bd ready`/`gc ready` round-trip costs several seconds, so the probe (~16s in isolation, more under load) intermittently exceeded the **15s `gc hook run` wrapper** and **30s work-query subprocess** budgets and was killed *before* reaching the pool-demand tier. The work was findable (the pool-demand query returns it in ~3s) — the probe just never got there in time.

## Fix

Raise the two work-query-path budgets to cover the realistic loaded cost of the multi-round-trip probe:
- `defaultHookRunTimeout` 15s → 30s (the `gc hook run` wrapper)
- `hookWorkQueryTimeout` 30s → 60s (the work-query subprocess)

The subprocess cap is kept ≥ the wrapper so the wrapper never preempts the inner work query. The `--timeout` flag default in `docs/reference/cli.md` regenerates to 30s accordingly.

The assigned tiers can't simply be skipped (they're crash-recovery: a pool operator can have in-progress work after a restart), so right-sizing the budget is the correct immediate fix.

## Testing

- New `TestWorkQueryTimeoutsAccommodateMultiRoundTripProbe` (TDD: fails at the old 15s, passes now) — asserts both budgets ≥ the ~6-round-trip probe budget and that the subprocess cap ≥ the wrapper.
- Full `cmd/gc` hook/work-query/timeout suite passes; `go build`/`go vet` clean; pre-commit gate (lint, genschema, vet) green.

## Follow-up (not in this PR)

[...]

PR #5161, opened 2026-08-10
feat(config): federate the work query and pool demand via `gc ready` (ga-bvdha)

julianknutsen:
## The failure this closes

On a converged split city the generated `work_query` and `scale_check` shell
`bd ready`, which reads exactly one store — whichever one `bd` resolves from
the agent's work directory. The execution DAG (`gcg-` molecule roots, step
beads, control beads) lives in the relocated graph binding. So a worker asks
for its routed work, gets a valid-looking empty array, and idle-exits; the
reconciler asks for its demand, gets zero, and drains the pool. **Neither
produces an error.** Measured on a live city at one moment (#5158): `gc bd
ready` answered with 5 beads and ZERO `gcg-`, while `GET /v0/beads/ready`
answered with 22, 14 of them `gcg-`.

## What changed

The four `bd ready` call sites in `internal/config`'s generated commands read
through **`gc ready`** (the federated reader that landed in #5158) when the
city relocates a coordination class, and through `bd ready` when it does not:

1. `assignedReadyTierCommand` — the pre-assigned tier (work_query Tier 2,
   `assigned_ready_query`, and the legacy-control variant of both).
2. `bdReadyPoolDemandShell` — the canonical routed tier (work_query Tier 3,
   `routed_pool_query`, and the count-form).
3. `bdReadyPoolDemandMigrationShell` — its `gc.run_target` retirement-window
   twin.
4. `poolDemandCountShell` — the reconciler count-form, which shares (2) and (3)
   with the worker path so the "scale_check ↔ work_query correspondence"
   holds by construction rather than by review.

The topology is threaded as `config.QueryTopology{Beads, FederatedReady}`,
replacing the `BeadsConfig` parameter on the seven `Effective*ForBeads`
accessors. It is a type change, so every call site is compiler-forced to answer
the topology question rather than inheriting a default.

`FederatedReady` comes from **`graphClassBinding`**, not from `[storage]`.
`storageSplitShapeOf` reads config alone and answers "no split" for a city
whose section was DELETED after it had already served one — the exact edit
`storage_boot.go`'s bypass note exists to catch. Asking config would build that
city a `bd ready` command that reads the work ledger and reports "no work"
forever; asking the routes builds it the federated command, which fails loud
with the refusal that names the remedy.

## Fail loud is the whole point

`gc ready` answers a dead or unopenable leg with a non-zero exit naming the
rig, never a short array, because a short array is indistinguishable from "no
work". A probe tier that captured that exit into an empty `$r` and fell through
to the next tier would re-create the identical fail-open one layer up and the
federation would buy nothing.

So every federated tier carries `|| exit $?` and drops the `2>/dev/null` that
would hide which rig failed. The count-form already had the discipline (its
`&&` chain predates this) and keeps it.

This is **executed, not asserted**:
`internal/config/workquery_federation_test.go` runs the real generated scripts
against a fake `gc` whose `ready` fails the way a dead rig makes it fail, and
checks the script's own exit status and stdout.

The single-store tiers keep their fall-through. That is not an oversight being
pinned: they have a later tier to fall through TO, and changing it would alter
what every deployed city does on a flaky store.

The 42 existing golden files are **unchanged**; 24 new `_federated` goldens
describe what a relocated city runs instead. A change that made a
non-relocated city emit anything but the command it emits today reddens the
existing goldens.

## Custom overrides: verbatim, but no longer silent

A custom `work_query`/`scale_check` is still returned verbatim — rewriting an
operator's shell would be substring surgery on a script this package did not
write, and that is precisely the branch behavior this bead's brief rejected.
What is closed is the SILENCE: `Agent.FederationBlindOverrides` names the blind
key, and `gc hook` prints it on a relocated city (and prints nothing at all on
every city that relocates nothing, so it cannot become per-tick noise on a
legacy deployment).

## SEE BUT CANNOT CLAIM — bounded to one bead, on BOTH tiers

This makes graph-class work **visible** to the worker's query. `gc hook
--claim` still issues its claim through `hookClaimBdStoreContext`, a bd store
rooted in the agent's work directory (the WORK scope invariant I5 already
pins), which cannot reach the relocated class at all. So the bead this query
now surfaces is one the same command cannot claim. That is the deferral, and
it is real.

What is NOT true — and what an earlier revision of this description asserted —
is that both federated read tiers degraded the same way. They did not:

- The **routed** tier (`claimFirstEligibleHookCandidate`) skips the
  unclaimable candidate, sets `claimsErrored`, and lands in the shared drain
  as `action=drain reason=claims_errored`.
- The **assigned** tier (`claimFirstReadyHookAssignment` — the tier a graph
  step assigned to this worker arrives on, and one of the four reads this PR
  federates) returned `hookClaimResult{terminal: true, code: 1}` on the same
  error: a bare exit 1, no drain JSON for a `--json` caller, no
  `claimsErrored`, and — because `claimHookWorkWithRunner` returns `res.code`
  immediately on a terminal result — no fallback to another store.

The second one is worse than a deferral. `bestStoreWithWork` RANKS
`hookTierAssigned` ahead of `hookTierRouted`, so the store holding the
unclaimable graph step is *preferred* over one holding claimable work; a
single unclaimable assigned step therefore stranded **every other piece of
work that worker could have done**, every tick.

Fixed here: a claim that fails with `beads.ErrNotFound` — the only error that
*proves* the bead is not in the store the claim ran against — is now skipped
on the assigned tier too, carrying `claimsErrored` to the shared drain and
letting the federated loop reach the store that does hold claimable work
(`hookClaimBeadIsElsewhere`, `cmd_hook_claim.go`). Every other failure (write
timeout, store contention, a socket flap) still fails closed exactly as
before: ownership is unresolved on a bead this session already owns, and
claiming unrelated fresh work would strand it.

This is error handling only. **No claim-time write routing is smuggled in** —
that is still ga-601v2, and until it lands an assigned graph step is re-served
and re-skipped every tick and no worker ever runs it. The two remain a coherent
pair; what changed is that the pair is no longer load-bearing for *avoiding a
hard failure*, only for actually doing the work.

## The federated reader runs ONCE per hook tick

`gc ready` federates the city store, every bound rig store and the relocated
graph leg in a single call. The `hookStore` loop exists because `bd ready`
reads exactly one store — a premise the swapped tiers no longer satisfy — so
running the whole query once per store re-asked the same question `R+1` times
and re-opened every leg each time, inside a 60s budget whose own doc comment
records that it was already raised once for probe cost (#3728).

Measured on the real generated commands for a 5-rig city (own store + 5 rigs):
a hook tick issued **18** `gc ready` reads where one query's worth is **3**
(30 vs 5 `gc ready` processes at runtime, with all three identity vars set and
one pool target).

`scopeFederatedHookStores` pins the city-wide read to the primary entry and
leaves every later store on the **single-store command it ran before this
swap**, via a per-store `hookStore.command`. Deliberately not a blanket
collapse of the loop: the crash-recovery (`bd list --status in_progress`) and
ephemeral (`bd query`) tiers are still per-store reads `gc ready` does not
answer, so dropping the extras would silently strip rig coverage from crash
recovery. The extras' cost and coverage are therefore exactly what they were
before this PR; only the duplicated city-wide read is removed.

On a single-store city both forms of the query are the same string, the call
returns `stores` unchanged, and no entry carries a per-store command — the
byte-identity claim holds by construction, and is asserted
(`TestSingleStoreHookStoresKeepOneSharedCommand`).

## Conformance

New invariant **I14-work-query-federation**, two topologies via
`forEachTopologyWithRig`:

- `cityQueryTopology` answers `FederatedReady` iff the class is relocated.
- Single-store: neither command names `gc ready`; both still name `bd ready`.
- Split: neither names `bd ready`; both name `gc ready` and both propagate the
  reader's exit status.
- The reader the command names actually answers with the routed graph step,
  read through the production leg assembly (`readyFederationLegs` +
  `readyBeadsForOpts`) rather than a restatement of it.
- KNOWN GAP rows, now covering **both** claim paths rather than only the
  routed one:
  - the routed graph bead the query serves is unresolvable in the WORK store;
  - an **assigned** graph step (the `graphroute.ApplyGraphRouteBinding` shape:
    OPEN, carrying this worker's identity) is surfaced by the assigned tier's
    reader and is likewise unresolvable in the WORK store;
  - and the store's own not-found is one the claim path classifies as "this
    store does not hold it", driven through `claimHookWorkWithRunner` against
    the fixture's REAL work store, so the worker still claims the work-ledger
    bead ranked *behind* the unclaimable graph step.

The previous pin only minted with `routedTo` and read with `unassigned:true`,
so it never exercised the assigned shape at all — which is why the assigned
tier's terminal branch survived review.

`internal/config`'s `TestSingleStoreWorkQueryKeepsItsFallThrough` was also
passing vacuously: it installed the failing script as `gc`, and a single-store
query never invokes `gc`. It now installs the failure as `bd`, asserts the
generated command does not name `gc ready`, and asserts from an invocation log
that the failing binary was actually reached — the single-store tiers redirect
`bd`'s stderr to `/dev/null`, so nothing else proves it.

`workquery_federation_test.go` also added a new `os/exec` call site, which grew
the checked resource census by one call and one file and reddened
`TestRepositoryLedgerMatchesCensusAndDocumentation` (`scope=all` subprocess
608/607, `untagged` 412/411, Small 405/404). Rather than raise a baseline whose
own invariant is "totals cannot grow", the package's process construction is now
one seam — `runShellCommandCapture` in `config_test.go`, which
`runShellWithFakeBd` and `runGeneratedQuery*` both use — so the census returns to
its baseline with no ledger change.

## Deliberately NOT in this slice

- **The crash-recovery tier** (`bd list --status in_progress`). `gc ready
  --status in_progress` is its federated form, but the row it returns is fed to
  `inProgressBlockedByEnrichmentScript`, which resolves dependencies with
  `bd show` in the same work directory. The tier would become
  half-federated, and the recovered bead would still be unclaimable and
  unreleasable because `on_death`/`on_boot` are `bd list` + `bd update` too.
  Coherent with ga-601v2, not a rider here. Asserted as topology-blind by
  `TestQueryKindsWithoutAReadyReadAreTopologyBlind`, so it cannot drift shut.
- **Controller-owned call sites** (`build_desired_state.go`,
  `work_query_probe.go`, `template_resolve.go`, `worker_handle.go`,
  `mcp_integration.go`, the pool on_boot/on_death runners) keep the
  single-store form with the reason inline: resolving the federation fact means
  asking the storage routes, and `cliStorageRoutes` is explicitly the funnel
  for the half of the program that "never builds a CityRuntime". A controller
  reaching it would open the city's binding a second time in a process that
  already holds it open. Threading `CityRuntime.storageRoutes` into that
  plumbing is a change to controller wiring, not part of swapping the reader.
- **The bd-1.0.4 ephemeral tiers** (`bd query ephemeral=...`) need no federated
  form: a relocated class store has no bead-policy layer, so an orchestration
  wisp lands there as a durable row the plain federated ready read already
  returns.
- **Claim-time write routing** (ga-601v2). This PR changes claim ERROR
  HANDLING only. `hookClaimBdStoreContext` still roots the claim in the
  agent's work directory; nothing here resolves a bead's owning store.

## Gates

- `go build ./...` — clean
- `go vet ./...` — clean
- `gofmt -l ./cmd ./internal ./test` — empty
- `golangci-lint 2.10.1` — 0 issues
- `go test ./cmd/gc -timeout 25m` (serial) — ok 745.326s
- `scripts/check-core-boundary.sh` — OK (no commercial coupling)
- `go test ./internal/testpolicy/...` — ok (resource census; no new `os/exec`
  call sites)


julianknutsen:
## Review fixes pushed — `4c35813860`

Two majors and three minors from the council review, each with a test that is
red without the fix.

### M-A — the assigned tier hard-failed, and the shippability argument did not cover it

The KNOWN GAP paragraph claimed the see-but-cannot-claim case degrades to
`action=drain reason=claims_errored`. That was true only for the **routed** tier.
The newly federated **assigned** tier (`assignedReadyTierCommand` →
`claimFirstReadyHookAssignment`) returned `hookClaimResult{terminal: true,
code: 1}` on the same error: a bare exit 1, empty stdout for a `--json` caller,
no `claimsErrored`, and no fallback — `claimHookWorkWithRunner` returns
`res.code` immediately on a terminal result.

And it is not incidental: `hookTierAssigned` outranks `hookTierRouted`, so
`bestStoreWithWork` **prefers** the store whose query returned the unclaimable
graph row over one holding claimable work. One unclaimable assigned step
stranded everything else that worker could have done.

**Shape taken: (b), the non-terminal skip** — not (a), the federation gate.
Reasons: (a) adds a gate someone must remember to remove and leaves the assigned
tier blind to relocated work in the meantime; (b) is ~10 lines, removes the
strands-other-work behaviour, and improves the non-split case too (a routed id
whose bead was deleted between read and claim no longer exits 1 either).

The downgrade is deliberately narrow. Only `beads.ErrNotFound` — the one error
that *proves* the bead is not in the store the claim ran against
(`hookClaimBeadIsElsewhere`) — is skipped. A write timeout, store contention or
a socket flap still fails closed, because ownership is then unresolved on a bead
this session already owns and claiming unrelated fresh work would strand it.
`TestClaimHookWorkAssignedTierOperationalErrorStaysTerminal` pins that half.

**No claim-time write routing is smuggled in.** `hookClaimBdStoreContext` still
roots the claim in the agent's work directory; ga-601v2 is untouched.

Re-run at the reviewer's measured setup (store A holds the assigned open `gcg-`
row, store B holds claimable routed work):

| | before | after |
|---|---|---|
| exit code | 1 | 0 |
| stdout | *(empty)* | `{"action":"work","reason":"claimed","bead_id":"hw-riga",...}` |
| claim attempts | `[gcg-abc123]` | `[gcg-abc123, hw-riga]` |
| claim dir | store A | store B |

Red-before (`TestClaimHookWorkAssignedTierUnresolvableBeadDoesNotStrandLaterStore`):

```
claimHookWorkWithRunner = 1, want 0: an assigned graph step this store cannot
resolve must not exit before the store holding claimable work is tried;
stdout="" stderr="gc hook --claim: promoting ready assignment gcg-abc123:
claiming bead \"gcg-abc123\": bead not found\n"
```

And with only the assigned step present
(`...UnresolvableBeadDrainsClaimsErrored`), red-before
`= 1, want 0 (structured drain); stdout=""`; green-after `action=drain
reason=claims_errored` with the drain ack completed.

### M-B — the city-wide reader ran once per store

Measured on the real generated commands for a 5-rig city (own store + 5 rigs):

| | before | after |
|---|---|---|
| `gc ready` reads per hook tick (textual) | **18** | **3** |
| `gc ready` processes at runtime (3 identity vars + 1 pool target) | **30** | **5** |

Red-before (`TestFederatedHookStoresIssueTheCityWideReaderOnce`):

```
a 6-store hook tick issues 18 city-wide `gc ready` reads, want 3 (one query's
worth): the federated reader already covers every leg, so the extra 15 re-open
every store for the same answer
```

**Not a blanket collapse of the store loop**, which the review's own note flags
as unsafe: the crash-recovery (`bd list --status in_progress`) and ephemeral
(`bd query`) tiers are still per-store reads `gc ready` does not answer, so
dropping the extras would silently strip rig coverage from crash recovery.
Instead `scopeFederatedHookStores` pins the city-wide read to the primary entry
and gives every later store the **single-store command it ran before this swap**
— its previous cost and its previous coverage, exactly. The PR's marginal cost
is now one federated fan-out, not R+1.

Single-store byte identity is preserved *by construction*: both forms of the
query are the same string there, the call returns `stores` unchanged, and no
entry carries a per-store command. Proven by mutation — forcing the early return
reddens the federated case and leaves
`TestSingleStoreHookStoresKeepOneSharedCommand` green.

### Minors

1. **`TestSingleStoreWorkQueryKeepsItsFallThrough` passed vacuously.** It
   installed the failing script as `gc`; a single-store query never invokes
   `gc`. Now it fails `bd`, asserts the generated command does not name
   `gc ready`, and proves from an invocation log that the failure was actually
   reached — the tiers redirect bd's stderr to `/dev/null`, so nothing else
   proves it. Red-before under the original harness shape: *"the failing `bd`
   was never asked for `ready` (log="", err=... no such file or directory)"* —
   the log file was never even created.

2. **The KNOWN GAP was pinned by a tautology.**
   `conformanceWorkQuerySeesButCannotClaim` minted with `routedTo` and no
   assignee and read with `unassigned:true`, so it never exercised the assigned
   shape — which is why M-A survived review. It now also mints the
   `graphroute.ApplyGraphRouteBinding` shape (OPEN, assigned to this worker),
   asserts the assigned tier's reader surfaces it, asserts the WORK store cannot
   resolve it *and* that the failure is one the claim path classifies as "this
   store does not hold it", and drives `claimHookWorkWithRunner` against the
   fixture's **real** work store so the worker still claims the work-ledger bead
   ranked *behind* the unclaimable graph step. Red-before with the M-A fix
   reverted:

   ```
   I14-work-query-federation/split: gc hook --claim exited 1 with the relocated
   assigned step gcg-wisp-0015 ranked first; one unclaimable graph step must not
   strand the worker's other claimable work
   ```

   Both rows still name ga-601v2 as the thing that flips them.

3. **The per-tick query storm / "ship only with ga-601v2".** Partly addressed:
   the drain is now distinguishable on both tiers, and M-B cuts the per-tick
   read count 6×. The KNOWN GAP paragraph in the description and in the
   conformance doc comment has been rewritten — it was factually wrong.

### Also found and fixed

`workquery_federation_test.go` added an `os/exec` call site that grew the checked
resource census by one call and one file, reddening
`TestRepositoryLedgerMatchesCensusAndDocumentation` on this branch (`scope=all`
subprocess 608/607, `untagged` 412/411, Small 405/404 — verified pre-existing at
`d3dad8adf8`, i.e. introduced by this PR, not by the fixes). Rather than raise a
baseline whose own invariant is *"totals cannot grow"*, `internal/config`'s test
process construction is now one seam (`runShellCommandCapture`), so the census
returns to baseline with no ledger change.

### Gates (all from the pushed tree)

- `go build ./...` — clean
- `go vet ./...` — clean
- `gofmt -l ./cmd ./internal ./test` — empty
- `golangci-lint 2.10.1 run ./...` — **0 issues**
- `go test ./cmd/gc -timeout 25m` (serial) — **ok 770.789s**
- `go test ./internal/...` — **ok**
- `go test ./internal/testpolicy/...` — **ok**
- `scripts/check-split-topology-rows.sh` — OK (exit 0)

### Verdict on ga-bvdha × ga-601v2

They no longer *have* to ship together to avoid a hard failure — that was the
part that was actually broken, and it is fixed. They should still ship together
to be **useful**: until the claim is class-routed, an assigned graph step is
re-served and re-skipped every tick and no worker ever runs it. The pairing is
now a completeness argument, not a safety one.

PR #5162, opened 2026-08-10
fix(cli): refuse a `gc bd list` projection over a class it cannot see (ga-fq54n)

julianknutsen:
## The bug (ga-fq54n)

win-mc-forge's measurement row #2, the one that started this program, verified
still open on `f08858b8a4`. On a converged split city:

```
gc bd list --metadata-field gc.root_bead_id=<gcg root>   →   0 rows, exit 0
```

Every piece worked as designed. The dialect guard was scoped to `sql`/`query`;
`--metadata-field` is not an id-VALUED flag, so `cmd_bd_by_id.go`'s by-id door
correctly declined it (a QUOTED id decides nothing about ownership —
`cmd_bd_by_id.go:78-81`, `:830-835`); and bd then ran the projection
successfully against the one ledger that holds no `gcg-` row. The value named
an id, but the **verb is a PROJECTION** over a class this ledger cannot see.

**Invariant 0: a projection that cannot see a class MUST fail loudly.
Returning `[]` is forbidden.**

The asymmetry is what made it urgent: `gc bd dep tree <gcg id>` already
**refuses with exit 1** (`cmd_bd_by_id.go:580-586`, `:630-637`) while `list`
answered `[]` with exit 0. Two projections over the same molecule, through the
same command, with opposite failure semantics — worse than either alone,
because an operator who learned the loud one trusts the quiet one.

## The shape chosen: refuse loudly, and name the federated reader

**(b) refuse**, not (a) federate. `gc bd list` is a bd passthrough with ~40
value flags and ~23 bool flags (`internal/bdflags`), plus output shapes
(`--long`, `--tree`, `--pretty`, the human table) that cannot be reproduced in
process. Federating the representable subset would still leave every other
spelling silent — so it would need this refusal arm **anyway**, on top of a
partial in-process imitation of bd, which is precisely what
`cmd_bd_by_id.go`'s own header refuses to build ("the caller leaves it on its
existing path rather than serving a partial imitation of bd"). It would also
make split-city `gc bd list` bytes diverge from single-store bytes for the same
query, which `gc ready` never had to do because it was a new command with its
own contract.

So (a)'s *user outcome* is delivered by steering to the surface that already
has the leg composition, the fail-loud rule and CLI==API conformance: the
refusal names `gc ready --metadata-field "key=value"`, which federates
city → rigs ascending → graph last, first-leg-wins, and fails loud on a dead or
unopenable leg (`cmd/gc/ready_federation.go`, #5158) — **and the message states
exactly how far that gets**, because it answers a narrower question than the
read it replaces (see "The steering is honest about its limit" below).
`gc beads list` is deliberately **not** offered: its API lane federates, but
its no-controller fallback opens only the city and rig stores
(`openAllConvoyStoresAt`), so it would hand back the same blind answer this
message is refusing — and an operator hits this at 2am often *because* the
controller is down.

`list`, `ready` and `search` join the existing dialect guard rather than
growing a new mechanism. The argv scan now also reads a flag's **inline**
value, so `--metadata-field=k=gcg-1` cannot switch the guard off with a
single `=`. The
pre-existing `GC_BD_ALLOW_RELOCATED_CLASS_READ` override still covers the real
false positive (the work ledger legitimately carries `gcg-` strings in
metadata — `ensureDrainUnitConvoy` stamps `gc.drain_control_id`), and honoring
it stays loud.

## The two projections now agree

| invocation | before | after |
| --- | --- | --- |
| `gc bd dep tree gcg-abc123` | exit 1, refused | exit 1, refused (unchanged) |
| `gc bd list --metadata-field gc.root_bead_id=gcg-abc123 --json` | **`[]`, exit 0** | exit 1, refused |

Both exit non-zero, both name the id namespace that cannot be seen **and** the
binding it is served from, and neither reaches the ledger that cannot answer.
`I14-projection-coherence` asserts that correspondence on **both** topologies
from the two production predicates that decide it
(`bdSQLRelocatedClassRefusal` for `list`, `bdArgsNameClassOwnedBead` for
everything else), and `TestGcBdProjectionsAgreeOnAClassTheyCannotSee` drives it
end to end through the real `doBd` against a bd stub that answers `[]` and
exits 0.

The messages are deliberately not compared verbatim: `dep tree` is refused by
the by-id door, which knows the exact bead and reports OWNERSHIP of it, while
`list` is refused by the dialect guard, which knows only that the query names
the namespace. Pinning identical wording would force one arm to say something
it does not know.

## Single-store cities are byte-identical, proven by mutation

`relocatedBeadClasses` returns nothing on a city with no `[storage]` split, the
scan is never entered, and the same argv reaches bd verbatim
(`TestGcBdListIsUnchangedOnASingleStoreCity` asserts bd received the exact
projection and that stdout is bd's own answer untouched). I14's single-store
row is not hardcoded: the fixture mints work-prefixed ids there, so the same
two argvs carry no reserved prefix and both projections pass through.

Mutations run, all of which turn the new rows red:

* removing the `"list"` entry from `bdRelocatedClassGuardedVerbs` → 5 table
  rows + 3 end-to-end tests + I14's split row fail
* reverting the inline-value scan → `list on a graph root id inline` fails
* removing `"ready"`/`"search"` → 3 table rows, 3 end-to-end rows,
  `TestBdRelocatedClassGuardCoversEverySelectorVerb` and I14's split row fail
* reverting `atSelectorValueStart` to `atQueryValueStart` → 8 internal rows,
  15 table rows and 8 end-to-end rows fail

## Stale docs in the same files (ga-kqk8j)

1. `cmd/gc/bd_relocated_classes.go` still claimed `show`/`dep tree` were raw
   passthroughs with **no class routing**. Falsified by #5132 (`18b1743e07`),
   wired at `cmd_bd.go:260`. Rewritten to state what each verb actually does:
   `show`, `update` (incl. `--claim`), `release-if-current` and `dep list` are
   class-routed in process; `dep tree` is refused on a class-owned id; every
   other subcommand that ADDRESSES a reserved-prefix id is refused by
   ownership rather than servability.
2. `internal/beads/membership.go` attributed spec-sidecar dependency isolation
   to `formula.newSourceSpecStep` "clearing" `DependsOn`/`Needs`/`WaitsFor`.
   Verified against the tree: that function
   (`internal/formula/source_spec.go:22-33`) is a fresh `&Step{...}` literal
   that never SETS them; `namespaceSourceSpecStep:52-54` is the one place that
   actively nils them. The conclusion holds; the citation sent the next reader
   to the wrong function, and it is load-bearing for why the measured 48 looked
   stable. Two verbatim copies of the same wrong citation
   (`internal/beads/membership_test.go`, `internal/dispatch/fanout_membership_test.go`)
   move with it, and `namespaceSourceSpecStep` — which had no doc — now says
   why a spec carries no edges.

## Conformance gap accounting

No KNOWN GAP became closable. I12's gap is the **object-model** half of
Invariant 0 (`beads.DirectMembers` handed the wrong store still answers with an
empty member set); closing it means giving a store the ability to refuse a
class it does not hold, which is a behavior change on every `DirectMembers`
consumer, not a read-path guard. I12's paragraph now says so explicitly, so a
reader does not mistake I14 for having closed it. No new gap was introduced, so
no new skip.

## Gates

* `go build ./...` — clean
* `go vet ./...` — clean
* `gofmt -l ./cmd ./internal` — empty
* `golangci-lint` 2.10.1 over `./cmd/gc/... ./internal/beads/... ./internal/formula/... ./internal/dispatch/...` — **0 issues**
* `go test ./cmd/gc -timeout 25m` (serial) — **ok, 756.3s**
* `go test ./internal/...` — **exit 0, 144 packages**
* `scripts/check-core-boundary.sh` — OK
* `make check-split-topology-rows` — exit 0

---

# Council review round 2 — three defects, fixed (`d05afb11dd`)

Rebased onto `31ecad2364` (#5163). Seven confirmed majors reduced to three
distinct defects; all three plus the three minors are fixed.

## Defect 1 — the refusal was far too broad

`RelocatedClassesInListSelector` reused `atQueryValueStart`, which returns
`true` at offset 0. For `bd query` one token is a whole expression, so a bare
term IS a value position. For a **selector flag** the flag has already consumed
its own token, and `bd list` accepts **no positional arguments at all**
(`cmd/bd/list.go:494`: *"bd list does not accept positional arguments; use
flags instead"*), so every token the scan sees is some flag's VALUE. Offset-0
anchoring therefore made **every free-text selector id-shaped**, and
`--title-contains gcg-abc123` became indistinguishable from
`--metadata-field gc.root_bead_id=gcg-abc123`.

That is the exact false positive the guard's own header
(`internal/beads/bdsql_relocation.go`) names FIRST as the thing the anchoring
rules exist to let through — a LIKE-contains over a text column is a question
about the rows THIS ledger holds, and the work ledger really does carry `gcg-`
strings.

**Anchor chosen: `atSelectorValueStart` — the id must be preceded by the `=` of
a `key=value` predicate** (whitespace and quotes skipped on the way back, so
`k="gcg-1"` and `k= gcg-1` still anchor). Dropping the offset-0 arm is the
whole fix.

**Why this over gating on `bdflags.ValueFlags("list")`:**

1. `bdflags` records **arity, not meaning**. A flag allowlist needs a *second*
   manifest of what each flag MEANS, maintained by hand against every flag bd
   grows — and its failure mode is *silent under-guarding*, which is the bug
   class this PR exists to close.
2. It needs one list **per verb**. `list`, `ready` and `search` have different
   selector sets; the `=` anchor is correct for all three with no list at all.
3. Every other dialect in this family classifies **text**. The `=` anchor
   derives the answer from the token's own shape the way `atSQLLiteralStart`
   and `atQueryValueStart` do, so the three dialects stay one mechanism.
4. It is strictly **narrower** where it matters. `--id`/`--parent`/`-p` fall
   through to the by-id ownership door — which was already refusing them on
   `main` — and that door names the **bead and its binding**, not just the
   namespace. The offset-0 arm was shadowing a more specific refusal with a
   vaguer one, for zero added coverage.
   `TestGcBdListOnAnIDValuedFlagRefusesByOwnership` pins that.

## Defect 2 — the same bug was still live one verb over

`gc bd ready --metadata-field <k>=<gcg id>` and `gc bd search` answered `[]`
exit 0, and the shipped comment justified the exemption with *"`ready` stays
out: its selectors name templates and labels, not bead ids"* — which
`internal/bdflags/bdflags.go` falsifies in the same tree.

**Both verbs are now guarded.** `--metadata-field` is registered on exactly
three bd subcommands (`cmd/bd/list.go:831`, `ready.go:820`, `search.go:390`),
and all three are in `bdRelocatedClassGuardedVerbs`.
`TestBdRelocatedClassGuardCoversEverySelectorVerb` derives that requirement
from `bdflags` so a fourth cannot appear unguarded. The false rationale is
gone; the replacement states the checkable completeness claim.

Verified safe for the work loop: the pool-demand probe and the control
dispatcher select on `gc.routed_to` / `gc.run_target`, whose values are pool
template names, and they invoke raw `bd`, not `gc bd`.
`TestGcBdReadyKeepsAnsweringItsOrdinaryWorkQueries` pins the argv-level half.

## Defect 3 — the steering is honest about its limit

`gc ready --metadata-field` answers a **narrower** question than the refused
read. With no `--status` it issues a `ReadyQuery` — claimable, unblocked work
only — so a molecule whose steps are in flight, blocked or closed comes back
`[]` **from the command the refusal recommended**. And that is the modal case:
a stuck molecule is precisely one with no claimable step. `--status` takes
exactly one of `open`/`in_progress`/`blocked`/`closed`; there is no `--all`, no
comma list, and `deferred` is not selectable at all.

The message now says all of that and states plainly that **there is no
federated equivalent of `bd list --all` yet**:

> For a SET of beads rather than one id, `gc ready --metadata-field "key=value"`
> federates the city store, the rig stores and the relocated binding as ordered
> legs and fails loud on a leg it cannot read — but it answers a NARROWER
> question than this read did: with no --status it returns only claimable work,
> and --status takes exactly one of open, in_progress, blocked, closed (no
> --all, no comma list, and deferred is not selectable), so enumerating a
> molecule's full membership takes one invocation per status and cannot reach a
> deferred member at all. There is no federated equivalent of `bd list --all`
> yet.

I14's last leg previously ran against a molecule whose only step was open — the
one status for which the escape works. It now mints a **mid-flight** step and
asserts both halves: the bare spelling does NOT return it, and
`--status in_progress` does. The steering cannot silently become wrong again.

## The three minors

1. **The refusal now names `GC_BD_ALLOW_RELOCATED_CLASS_READ`.** Appended at
   the CLI seam (`bdRelocatedClassEscapeHint`), *not* inside
   `beads.RelocatedClassRefusal` — that same string is returned by `BdStore`'s
   id-scoped guard, which honors no env var, so naming it there would advertise
   an escape that does not work on half the paths that print it.
   `TestRelocatedClassRefusalLeavesTheOverrideToTheCLI` pins the negative;
   `TestGcBdRefusalNamesTheOverride` pins the positive.
2. **`cmd_bd_by_id.go`'s header** said `gc bd list --metadata-field
   workflow_id=gcg-…` "is a work question about work rows, and refusing it
   exec-fails the consumer that asks it" — the exact argv this PR refuses. Both
   sites now state what is true: the by-id door **declines it on OWNERSHIP**
   (a quoted id decides nothing about which store serves the read), and the
   dialect guard refuses it one pre-flight earlier on **SERVABILITY**.
3. `RelocatedClassesInListSelector` → `RelocatedClassesInSelector`, since it
   now serves three verbs. A name that says "list" while `ready` uses it is the
   same class of false citation ga-kqk8j exists to remove.

## Per-flag A/B: `main` (31ecad2364) → old head (17d7d83171) → now (d05afb11dd)

Predicates measured directly (`bdSQLRelocatedClassRefusal ‖
bdArgsNameClassOwnedBead`) on `splitCityConfig()`.

| argv | main | old head | now |
| --- | --- | --- | --- |
| `list --title-contains gcg-abc123` | pass | **REFUSE** | pass |
| `list --title-contains=gcg-abc123` | pass | **REFUSE** | pass |
| `list --desc-contains gcg-abc123` | pass | **REFUSE** | pass |
| `list --notes-contains gcg-abc123` | pass | **REFUSE** | pass |
| `list --title gcg-abc123` | pass | **REFUSE** | pass |
| `list --label gcg-abc123` | pass | **REFUSE** | pass |
| `list --exclude-label gcg-abc123` | pass | **REFUSE** | pass |
| `list --label-pattern 'gcg-*'` | pass | **REFUSE** | pass |
| `list --label-regex 'gcg-.*'` | pass | **REFUSE** | pass |
| `list --assignee gcg-worker` | pass | **REFUSE** | pass |
| `list -a gcg-worker` | pass | **REFUSE** | pass |
| `list --sort gcg` | pass | **REFUSE** | pass |
| `list --spec gcg-abc123` | pass | **REFUSE** | pass |
| `list --title-contains gcg` | pass | **REFUSE** | pass |
| `list --title-contains "fix (gcg-1) regression"` | pass | **REFUSE** | pass |
| `list --title-contains "regressions: gcg-1, gcg-2"` | pass | **REFUSE** | pass |
| `list --title-contains "root is 'gcg-1' here"` | pass | **REFUSE** | pass |
| `list --title-contains "fix gcg-1 regression"` | pass | pass | pass |
| `list --metadata-field gc.root_bead_id=gcg-abc123` | pass | refuse | **refuse** |
| `list --metadata-field=gc.root_bead_id=gcg-abc123` | pass | refuse | **refuse** |
| `list --json --metadata-field gc.root_bead_id=gcg-abc123` | pass | refuse | **refuse** |
| `--json list --metadata-field gc.root_bead_id=gcg-abc123` | pass | refuse | **refuse** |
| `list --metadata-field gc.nudge_id=gcn-1` | pass | refuse | **refuse** |
| `list --metadata-field 'gc.root_bead_id="gcg-abc123"'` | pass | refuse | **refuse** |
| `list --metadata-field gc.root_bead_id=GCG-ABC123` | pass | refuse | **refuse** |
| `list --metadata-field gc.root_bead_id=demo-abc` | pass | pass | pass |
| `list --metadata-field gc.root_bead_id=gcgx-1` | pass | pass | pass |
| `list --has-metadata-key gc.root_bead_id` | pass | pass | pass |
| `list --status open` | pass | pass | pass |
| `list --id gcg-abc123` | refuse (by-id) | refuse | refuse (by-id) |
| `list --parent gcg-abc123` | refuse (by-id) | refuse | refuse (by-id) |
| `list -p gcg-abc123` | refuse (by-id) | refuse | refuse (by-id) |
| `ready --metadata-field gc.root_bead_id=gcg-abc123` | **pass** | **pass** | **refuse** |
| `ready --metadata-field=gc.root_bead_id=gcg-abc123` | **pass** | **pass** | **refuse** |
| `ready --label gcg-abc123` | pass | pass | pass |
| `ready --metadata-field gc.routed_to=demo/worker` | pass | pass | pass |
| `ready --parent gcg-abc123` | refuse (by-id) | refuse (by-id) | refuse (by-id) |
| `search --metadata-field gc.root_bead_id=gcg-abc123` | **pass** | **pass** | **refuse** |
| `search gcg-abc123` | refuse (by-id) | refuse (by-id) | refuse (by-id) |
| `sql "... id = 'gcg-abc'"` | refuse | refuse | refuse |
| `sql "... title like '%gcg-abc123%'"` | pass | pass | pass |
| `sql --format=gcg-abc123 'select 1'` | refuse (by-id) | refuse | refuse (by-id) |
| `query id=gcg-abc123` | refuse | refuse | refuse |
| `query title~gcg-abc123` | pass | pass | pass |

Every regressed row is back to its `main` behaviour; the original bug still
refuses in every spelling; `ready`/`search --metadata-field` are the only
intentionally-new refusals.

## Red-before strings

Reverting `atSelectorValueStart` → `atQueryValueStart`:

```
--- FAIL: TestRelocatedClassesInSelectorMatchesOnlyThePredicateValue/bare_id_as_a_whole_selector_value
    RelocatedClassesInSelector("gcg-abc123") matched = true, want false
--- FAIL: TestRelocatedClassesInSelectorMatchesOnlyThePredicateValue/prose_with_a_paren_before_the_id
    RelocatedClassesInSelector("fix (gcg-1) regression") matched = true, want false
--- FAIL: TestRelocatedClassesInSelectorMatchesOnlyThePredicateValue/prose_with_a_quote_before_the_id
    RelocatedClassesInSelector("root is 'gcg-1' here") matched = true, want false
--- FAIL: TestRelocatedClassesInSelectorDivergesFromTheQueryDialect
    RelocatedClassesInSelector("gcg-abc123") matched [{graph gcg ...}]; a whole-token
    selector value is a search string, and refusing it breaks `--title-contains`
--- FAIL: TestGcBdListForwardsAFreeTextSearchThatNamesAGraphID/title_search
    `gc bd list --title-contains gcg-abc123 --json` exited 1 on a split city; a
    LIKE-contains over this ledger's own columns is a question bd answers, and
    refusing it withholds rows that exist.
```

(plus 15 rows of `TestBdSQLRelocatedClassRefusalOnASplitCity` and 7 more
subtests of `TestGcBdListForwardsAFreeTextSearchThatNamesAGraphID`.)

Removing `"ready"`/`"search"` from the guard map:

```
--- FAIL: TestBdSQLRelocatedClassRefusalOnASplitCity/ready_on_a_graph_root_id
    bdSQLRelocatedClassRefusal([ready --metadata-field gc.root_bead_id=gcg-abc123]) refused = false, want true
--- FAIL: TestGcBdReadyRefusesAGraphClassProjectionOnASplitCity/ready
    `gc bd ready --metadata-field gc.root_bead_id=gcg-abc123 --json` exited 0 with
    stdout="[]\n"; that empty array is the same silent-empty `gc bd list` refuses
    on the same molecule
--- FAIL: TestBdRelocatedClassGuardCoversEverySelectorVerb
    bd "ready" takes --metadata-field but is not in bdRelocatedClassGuardedVerbs
--- FAIL: TestSplitTopologyConformance/I14-projection-coherence/split
    `gc bd ready --metadata-field gc.root_bead_id=gcg-1` refused = false but `gc bd
    list` with the same selector refused = true
```

Reverting the steering + the override hint:

```
--- FAIL: TestRelocatedClassRefusalStatesTheLimitsOfTheSetEscape
    the steering does not state "with no --status it returns only claimable work"
    the steering does not state "exactly one of open, in_progress, blocked, closed"
    the steering does not state "deferred is not selectable"
    the steering does not state "no federated equivalent of `bd list --all`"
--- FAIL: TestGcBdRefusalNamesTheOverride
    the refusal never names GC_BD_ALLOW_RELOCATED_CLASS_READ, so the operator
    holding a false positive has no in-band way out
--- FAIL: TestSplitTopologyConformance/I14-projection-coherence/split
    the list refusal does not name "with no --status it returns only claimable work"
```

## Gates (on `d05afb11dd`, rebased on `31ecad2364`)

* `go build ./...` — clean
* `go vet ./...` — clean
* `gofmt -l ./cmd ./internal` — empty
* `golangci-lint` 2.10.1 over `./...` — **0 issues**
* `go test ./cmd/gc -timeout 25m` (serial, on the committed tree) — **ok, 736.3s**
* `go test ./internal/...` — **exit 0, 144 packages**
* `scripts/check-core-boundary.sh` — OK
* `scripts/check-split-topology-rows.sh` — exit 0
* `scripts/check-routed-test-rows.sh` — exit 0

`cmd/gc/cmd_bd_by_id.go` is touched for **comments only** (the two doc sites
that contradicted shipped behaviour); no by-id routing logic changed.

### DIFF
--- a/cmd/gc/bd_relocated_classes.go
+++ b/cmd/gc/bd_relocated_classes.go
@@ -101,20 +101,29 @@ func bdRelocatedClassOverrideEnabled() bool {
 // operator or agent wrote by hand, both resolve it against the bd ledger alone,
 // and both answer no-match with an empty result and exit 0.
 //
-// `list` is here for the same reason and it is the one that took longest to
-// see, because the flag it arrives through does not look like an id position.
-// `gc bd list --metadata-field gc.root_bead_id=<gcg root>` answered `[]` with
-// exit 0 on a converged split city: --metadata-field is not id-VALUED, so the
-// by-id door in cmd_bd_by_id.go correctly declined it (a quoted id decides
-// nothing about ownership), and bd then ran the projection successfully against
-// the one ledger that holds no gcg- row. The value named an id but the VERB is
-// a PROJECTION over a class this ledger cannot see, and a projection that
-// cannot see a class must fail loudly rather than answer with the empty set.
-// That is the whole of ga-iaj7k's Invariant 0, and it is what makes `list`
-// COHERENT with `dep tree`, which already refuses a relocated id — two
-// projections over the same data with opposite failure semantics is worse than
-// either one alone, because an operator who learned the loud one trusts the
-// quiet one.
+// `list`, `ready` and `search` are here for the same reason and they were the
+// ones that took longest to see, because the flag the id arrives through does
+// not look like an id position. `gc bd list --metadata-field
+// gc.root_bead_id=<gcg root>` answered `[]` with exit 0 on a converged split
+// city: --metadata-field is not id-VALUED, so the by-id door in cmd_bd_by_id.go
+// correctly declined it (a quoted id decides nothing about ownership), and bd
+// then ran the projection successfully against the one ledger that holds no
+// gcg- row. The value named an id but the VERB is a PROJECTION over a class
+// this ledger cannot see, and a projection that cannot see a class must fail
+// loudly rather than answer with the empty set. That is the whole of ga-iaj7k's
+// Invariant 0, and it is what makes `list` COHERENT with `dep tree`, which
+// already refuses a relocated id — two projections over the same data with
+// opposite failure semantics is worse than either one alone, because an
+// operator who learned the loud one trusts the quiet one.
+//
+// `ready` and `search` are the same projection over the same flag: bdflags
+// lists --metadata-field for `ready`, bd registers it for `search`, and both
+// answer no-match with `[]` and exit 0. Guarding `list` alone would have minted
+// the asymmetry it retired — one verb over, on the same molecule, through the
+// same selector — and inside `gc bd ready` itself, whose --parent/--mol
+// spellings the by-id door already refuses loudly. They share `list`'s scan
+// because they share bd's selector dialect, so the negatives that keep a
+// free-text search answerable hold for all three unchanged.
 //
 // The other verbs are unguarded because they are no longer blind, not because
 // they were ever safe:
@@ -130,13 +139,18 @@ func bdRelocatedClassOverrideEnabled() bool {
 //     positional or an id-valued flag — is refused there too, by ownership
 //     rather than by servability.
 //
-// `ready` stays out: its selectors name templates and labels, not bead ids, so
-// there is nothing in its argv for this scan to classify, and the federated
-// answer to a claimable-work question already exists as `gc ready`.
+// The selector surface is COMPLETE, and that is checkable rather than hopeful:
+// --metadata-field — the only bd flag whose value side is a key=value predicate
+// on a read — is registered on exactly three subcommands (list.go, ready.go,
+// search.go in the pinned beads module), and all three are in this map.
+// TestBdRelocatedClassGuardCoversEverySelectorVerb pins that against bdflags so
+// a fourth cannot appear unguarded.
 var bdRelocatedClassGuardedVerbs = map[string]bdRelocatedClassScan{
-	"sql":   beads.RelocatedClassesInSQL,
-	"query": beads.RelocatedClassesInQueryExpr,
-	"list":  beads.RelocatedClassesInListSelector,
+	"sql":    beads.RelocatedClassesInSQL,
+	"query":  beads.RelocatedClassesInQueryExpr,
+	"list":   beads.RelocatedClassesInSelector,
+	"ready":  beads.RelocatedClassesInSelector,
+	"search": beads.RelocatedClassesInSelector,
 }
 
 // bdRelocatedClassScan classifies one argument's text in one bd dialect.
@@ -169,6 +183,14 @@ func bdRelocatedClassScanText(arg string) (string, bool) {
 // bdSQLRelocatedClassRefusal reports whether a `gc bd` invocation is an ad-hoc
 // read that names the id namespace of a class this city serves elsewhere, and
 // returns the operator-facing refusal when it is.
+//
+// The override is named HERE rather than inside beads.RelocatedClassRefusal
+// because this is the only seam where it works: the same error text is returned
+// by BdStore's id-scoped guard, which honors no env var, and a message that
+// offers an escape the reader cannot take is worse than one that offers none.
+// An escape hatch nobody can find is not an escape hatch — the scan classifies
+// TEXT, so a false positive is always possible, and the operator holding one
+// needs the way out in the message that stopped them.
 func bdSQLRelocatedClassRefusal(cfg *config.City, bdArgs []string) (string, bool) {
 	relocated := relocatedBeadClasses(cfg)
 	if len(relocated) == 0 {
@@ -200,6 +222,25 @@ func bdSQLRelocatedClassRefusal(cfg *config.City, bdArgs []string) (string, bool
 		return "", false
 	}
 	return beads.RelocatedClassRefusal(op, matched).Error(), true
+}
+
+// bdRelocatedClassEscapeHint is the sentence appended to a refusal that is
+// actually being ENFORCED, naming the knob that lifts it.
+//
+// It is not part of beads.RelocatedClassRefusal because that same text is
+// returned by BdStore's id-scoped guard, which honors no env var: a message
+// that offers an escape its reader cannot take is worse than one that offers
+// none. And it is not appended when the override is already set, because there
+// the operator is being told what they overrode, not how.
+//
+// The scan classifies TEXT, so a false positive is always possible — a work-row
+// query whose value legitimately holds a relocated id (gc.drain_control_id) is
+// indistinguishable from a class-scoped one. An escape hatch nobody can find is
+// not an escape hatch, so it travels with the refusal that stopped them.
+func bdRelocatedClassEscapeHint() string {
+	return fmt.Sprintf(" If this read is about work rows that merely REFERENCE such an id — a metadata comparison on "+
+		"gc.drain_control_id, say — it is a question this ledger can answer, and %s=1 runs it anyway.",
+		bdRelocatedClassOverrideEnvVar)
 }
 
 // bdRelocatedClassScans returns the dialect scans to run over an invocation's
--- a/cmd/gc/bd_relocated_classes.go
+++ b/cmd/gc/bd_relocated_classes.go
@@ -94,27 +94,77 @@ func bdRelocatedClassOverrideEnabled() bool {
 	}
 }
 
-// bdRelocatedClassGuardedVerbs are the bd read verbs whose positional text
-// names ids in a dialect this guard can classify.
+// bdRelocatedClassGuardedVerbs are the bd read verbs whose argument text names
+// ids in a dialect this guard can classify.
 //
 // `sql` and `query` are the two ad-hoc ones: both take an expression an
 // operator or agent wrote by hand, both resolve it against the bd ledger alone,
 // and both answer no-match with an empty result and exit 0.
 //
-// list/ready are left alone — they answer about the ledger they are scoped to
-// and claim nothing more. show/dep tree are left alone too, but NOT because
-// they are safe: they are raw bd passthroughs against the same blind ledger
-// (doBd ends at exec.Command(bdPath, bdArgs...) with no class routing). They
-// stay unguarded because a bare id is not a dialect this scan can read without
-// refusing every work-store id that starts with the same letters, and because
-// the refusal now steers to `gc beads show` instead of to them.
+// `list` is here for the same reason and it is the one that took longest to
+// see, because the flag it arrives through does not look like an id position.
+// `gc bd list --metadata-field gc.root_bead_id=<gcg root>` answered `[]` with
+// exit 0 on a converged split city: --metadata-field is not id-VALUED, so the
+// by-id door in cmd_bd_by_id.go correctly declined it (a quoted id decides
+// nothing about ownership), and bd then ran the projection successfully against
+// the one ledger that holds no gcg- row. The value named an id but the VERB is
+// a PROJECTION over a class this ledger cannot see, and a projection that
+// cannot see a class must fail loudly rather than answer with the empty set.
+// That is the whole of ga-iaj7k's Invariant 0, and it is what makes `list`
+// COHERENT with `dep tree`, which already refuses a relocated id — two
+// projections over the same data with opposite failure semantics is worse than
+// either one alone, because an operator who learned the loud one trusts the
+// quiet one.
+//
+// The other verbs are unguarded because they are no longer blind, not because
+// they were ever safe:
+//
+//   - `show`, `update` (including `--claim`), `release-if-current` and
+//     `dep list` are answered IN PROCESS from the binding their class is served
+//     from — cmd_bd_by_id.go, wired into doBd immediately after this scan — so
+//     they never reach the subprocess for a class-owned bead.
+//   - `dep tree` is not served in process, and on a class-owned id that same
+//     surface REFUSES it (exit 1, naming the bead and the binding) rather than
+//     forwarding it.
+//   - Every other bd subcommand that ADDRESSES a reserved-prefix id — in a
+//     positional or an id-valued flag — is refused there too, by ownership
+//     rather than by servability.
+//
+// `ready` stays out: its selectors name templates and labels, not bead ids, so
+// there is nothing in its argv for this scan to classify, and the federated
+// answer to a claimable-work question already exists as `gc ready`.
 var bdRelocatedClassGuardedVerbs = map[string]bdRelocatedClassScan{
 	"sql":   beads.RelocatedClassesInSQL,
 	"query": beads.RelocatedClassesInQueryExpr,
+	"list":  beads.RelocatedClassesInListSelector,
 }
 
-// bdRelocatedClassScan classifies one positional argument in one bd dialect.
+// bdRelocatedClassScan classifies one argument's text in one bd dialect.
 type bdRelocatedClassScan func([]beads.RelocatedClass, string) []beads.RelocatedClass
+
+// bdRelocatedClassScanText returns the part of an argument a dialect scan
+// should read, and whether there is one.
+//
+// A separated flag value arrives as its own token (`--metadata-field
+// gc.root_bead_id=gcg-1`) and is scanned as a positional. The INLINE spelling
+// of the same selector (`--metadata-field=gc.root_bead_id=gcg-1`) is one token
+// that begins with a dash, and skipping it wholesale — which is what this scan
+// used to do — let a single `=` switch the guard off on the exact query it was
+// added for. So the flag NAME is dropped and everything after the first `=` is
+// scanned, which is the value bd itself parses out of that token.
+//
+// A flag carrying no value (`--json`, `-q`) has no value text and is skipped,
+// which is what keeps `bd sql --json 'select 1'` from classifying its own flags.
+func bdRelocatedClassScanText(arg string) (string, bool) {
+	if !strings.HasPrefix(arg, "-") {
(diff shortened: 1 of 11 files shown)


julianknutsen:
## Council round-2 fixes pushed — `d05afb11dd` (rebased onto `31ecad2364`)

Seven majors → three defects, all fixed, plus the three minors. Full detail and
the per-flag A/B table are in the PR body; the short version:

**Defect 1 — the refusal was far too broad.** The selector dialect now has its
own anchor, `atSelectorValueStart`: the id must be preceded by the `=` of a
`key=value` predicate (quotes/spaces skipped). The offset-0 arm is gone.

Chosen over gating on `bdflags.ValueFlags("list")` because `bdflags` records
**arity, not meaning** — a flag allowlist needs a second, hand-maintained
manifest per verb whose failure mode is silent *under*-guarding; the `=` anchor
derives the answer from the token's own text the way `atSQLLiteralStart` and
`atQueryValueStart` do, is correct for all three selector verbs with no list at
all, and hands `--id`/`--parent`/`-p` back to the by-id ownership door, which
names the **bead and its binding** instead of only the namespace. (`bd list`
takes no positionals — `cmd/bd/list.go:494` — so every token the old anchor saw
was some flag's value, which is exactly why offset-0 caught everything.)

All 17 regressed flags are back to `main`'s behaviour; all 7 spellings of the
original bug still refuse; `list`/`dep tree` still agree on both topologies.

**Defect 2 — guarded, not exempted.** `ready` and `search` are in
`bdRelocatedClassGuardedVerbs` now. The false rationale ("its selectors name
templates and labels, not bead ids") is deleted and replaced with a *checkable*
claim: `--metadata-field` is registered on exactly three bd subcommands
(`list.go:831`, `ready.go:820`, `search.go:390`) and all three are in the map —
`TestBdRelocatedClassGuardCoversEverySelectorVerb` derives that from `bdflags`
so a fourth cannot appear unguarded. Work loop unaffected: the pool-demand and
dispatcher queries carry pool names, not ids, and invoke raw `bd`.

**Defect 3 — the steering states its own limit.** `gc ready` answers a narrower
question: no `--status` means claimable work only, `--status` takes exactly one
of open/in_progress/blocked/closed, no `--all`, no comma list, `deferred` not
selectable. The message says so and says plainly that **there is no federated
equivalent of `bd list --all` yet**. I14's escape leg now runs against a
**mid-flight** molecule — the modal case, since a stuck molecule is precisely
one with no claimable step — and asserts both that the bare spelling misses the
in-flight step and that `--status in_progress` finds it.

**Minors.** The refusal now names `GC_BD_ALLOW_RELOCATED_CLASS_READ`, appended
at the CLI seam rather than inside `beads.RelocatedClassRefusal` (that string is
also returned by `BdStore`'s id-scoped guard, which honors no env var — offering
an escape the reader cannot take is worse than offering none). `cmd_bd_by_id.go`'s
two doc sites no longer claim the refused argv is forwarded: the by-id door
declines it on **ownership**, the dialect guard refuses it on **servability**.
`RelocatedClassesInListSelector` → `RelocatedClassesInSelector`.

**Every fix has a test that fails without it** — three mutations run, red-before
strings quoted in the PR body. Single-store byte-identity re-proved by mutation
(`TestBdSQLRelocatedClassRefusalIsInertOnASingleStoreCity` now covers the
free-text, `ready` and `search` shapes too).

Gates: build/vet/gofmt clean; `golangci-lint` 2.10.1 over `./...` **0 issues**;
`go test ./cmd/gc -timeout 25m` serial **ok 729.7s**; `go test ./internal/...`
**exit 0, 144 pkgs**; `check-core-boundary.sh`, `check-split-topology-rows.sh`,
`check-routed-test-rows.sh` all exit 0.

PR #5165, opened 2026-08-10
feat(hook): route claim-time writes by coordination class (ga-601v2)

julianknutsen:
> **RETRACTION (2026-08-10).** An earlier revision of this PR and of `withClaimCapableClassLeg`'s doc comment claimed that materializing a graph.v2 molecule on the real SQLite binding fails on "the root → `workflow-finalize` `blocks` edge `internal/formula/compile.go` emits, refused by the real backend", and called that "a production question about graph.v2 on a split city". **That was wrong, and it is withdrawn.** No production graph.v2 recipe can trip that guard: `internal/formula/compile.go` gates both parent-child emitters on `!graphWorkflow`, and the SQLite guard fires only on `sqliteGraphApplyParentDepPairs`, which is built from a `parent-child` dep. The refusal came from `conformanceGraphRecipe()` carrying a hand-written `parent-child` dep no v2 compiler emits. That fixture dep is deleted here, the real SQLite leaf is now the **suite-wide** class leg, and `withClaimCapableClassLeg` is gone. There is no graph.v2-on-a-split-city hazard to investigate.

Closes the last implementation bead of the split-store program: **ga-601v2 — claim-time write routing by class**. The live-city proof (`ga-iaj7k`) was blocked only on this.

## The problem, and it was user-visible

`gc hook --claim` issued every claim-time write through `beads.NewBdStore(dir, …)` — a bd subprocess rooted in the agent's WORK directory. A relocated coordination class is not a bd workspace and cannot be expressed as a `hookStore{dir, env}` at all, so on a split city those writes ran against a ledger that could not see the bead.

ga-bvdha (#5161) made that visible: the generated work query now reads through the federated `gc ready`, so a relocated graph step assigned to a worker **is served** — and was then skipped every tick with `action=drain reason=claims_errored`. No worker ever ran it.

The red-before string, from `TestSplitTopologyConformance/I15-work-query-federation/split` with the routing disabled, is exactly that bug:

```
gc hook --claim result = {… BeadID:gc-1 Assignee:rig-A/executor …}, want the
relocated assigned graph step gcg-wisp-0003 — the worker claimed something else,
which is the re-serve/re-skip treadmill ga-601v2 closes
```

## Where the claim resolves its store, and why that shape

New file `cmd/gc/claim_class_route.go` adds a **CLASS axis to the existing `hookClaimOps` seam**, beside the rig axis already in `claim_cross_store.go` — not a parallel claim path, and **not a new leg in the query fan-out**: a binding is not a workspace, which is the same sentence the I5 pin used to explain why the gap existed.

Wrapped seams: `Claim`, `StampWorkMeta`, `ReadWorkMeta`, `ListContinuation`, `AssignContinuation`, `EmitExecutionStepStarted`.

**Probe, not prefix.** `gc storage migrate` preserves ids, so a relocated bead keeps its HQ/rig-era prefix and a prefix route would miss exactly the beads that moved. The store is found by asking which one holds the bead — the same reason `bdByIDClassDoor.resolve` probes residence.

**WORK first, and that is deliberately NOT the by-id order.**

| resolver | order | why |
| --- | --- | --- |
| `storeref.ClassCandidates`, `bdByIDClassDoor` | class first | caller holds only an id |
| `gc ready` (#5148/#5158/#5161) | city work first, graph leg **last** | co-resident id dedupes to the work row |
| **this claim** | **work first** | must agree with the reader that served its candidate |

The claim runs the existing work-scope write and escalates only where it answers. A class-first claim would take the class copy while the federated reader keeps answering from the still-open work copy — re-serving the bead every tick, which is the treadmill this PR closes, reintroduced one layer down. `TestClassRoutedClaimKeepsTheWorkCopyOnCoResidence` and I5's co-resident row assert it; mutating the order to class-first reddens both.

**The escalation signal is the existing one, unwidened.** `beads.ErrNotFound` is the only error that proves the bead is not in the store the write ran against (`hookClaimBeadIsElsewhere`). A write timeout, store contention or a controller-socket flap leaves ownership unresolved and is returned unchanged, never retried against a second store (`TestClassRoutedClaimFailsClosedOnEveryOtherError`). A read failure from the binding surfaces rather than being read as absence; the one-shot funnel's standing refusal is handled exactly as `classRoutedStoreForID` handles it.

`ListContinuation` is the one claim-time call with no not-found to escalate on, so it escalates on **empty** and only when the binding holds the group's root — monotone, so a work answer is never replaced, only an absence filled. A list error still fails loud, which is the branch bug this seam was warned about.

The CAS stays the store's: claims go through `storebinding.GraphStore.Claim`, the same acquire half `gc bd update --claim` routes through (#5132), never a read-then-write here. `hookClaimWithBdStore`'s post-mutation classification is factored into `hookClaimThroughStore` and shared, so the two claim paths report a lost race, a stale projection and a failed canonical readback identically.

## Review fixes on this revision

Six majors from the review council, each with a red-before row.

**1. A committed work claim is never escalated (findings 2 + 5).** `ops.Claim` passed only the error to `route.routes`, never `ok`. `hookClaimThroughStore` returns `(claimed, true, err)` for exactly one shape — the CAS COMMITTED and the canonical readback then failed — and `BdStore.Get` wraps `ErrNotFound` for a plain miss, an empty result AND `beads.ErrIDCollision`, so that readback failure routinely satisfied the predicate on a claim that had landed. The wrapper then claimed the same logical bead a second time in the binding and returned `err=nil`, erasing the signal both call sites stop on. Now `if ok { return claimed, ok, err }` before the escalation, matching `claimFirstReadyHookAssignment`'s own `!ok && hookClaimBeadIsElsewhere(err)`.

```
claim_class_route_test.go:444: routed claim returned err=<nil>, want the canonical-readback failure
  reloading claimed bead "gcg-b00": getting bead "gcg-b00": bead not found  returned unchanged
claim_class_route_test.go:444: … reloading claimed bead "gcg-b00": bd resolved a different bead ID
  (substring collision): bead not found  returned unchanged
```

**2. A binding-side claim refusal is per-bead, not per-tick (finding 1).** `*beads.SQLiteStore` is the only production store with the two-argument CAS; the other compiled-in binding provider (`beadsworkspace`, over `*beads.NativeDoltStore`) has none, so `storebinding.ErrBeadsAdapterCapability` came back — not `ErrNotFound`, so the WHOLE tick exited 1 and the worker claimed nothing, including healthy work in other stores. Both halves of the fix landed:

- **at the door**: `newHookClaimClassRoute` verifies the CAS once, the way `storebinding.NewBeadsNudgeQueue` verifies the same capability at construction, and `hookClaimRouteVerdict` degrades that ONE error to unrouted claiming with a loud line. A boot-time refusal in `openStorageRoutes` was considered and rejected: it would take a configuration that boots and serves reads today (`TestStorageGateServesACityFromItsBeadsWorkspace`) and make the whole city unusable — strictly larger than the availability bug being fixed, and a storage-plan question rather than a claim-routing one.
- **per bead**: `hookClaimBindingRefusedTheClaim` earns the same skip the work-side not-found gets. It is a SEPARATE predicate — `hookClaimBeadIsElsewhere` stays exactly as narrow as it was, `beads.ErrNotFound` only — because the two answer questions about different stores: the refusal is reached only after a work store proved this session owns nothing there, and it lands before any write.

```
control (route=nil, main's behavior) → exit 0, gc-77 claimed
claim-incapable leg                  → exit 1, stdout "", stderr
  "gc hook --claim: promoting ready assignment gcg-6: beads adapter capability unavailable: assignment claim"
```

**3. "WORK first" now means every work leg (finding 4).** Chose *make the rule true*, not rename it. The escalation fired on whatever leg `bestStoreWithWork` selected, and for a rig-scoped agent — the shipped shape of every worker that claims — the CITY work store is the leg that runs LAST. Only the primary leg runs the city-wide reader, so it is the only leg whose query can serve a bead it does not hold. Before a not-found opens the escalation, every OTHER leg is now READ for the same id (`anotherWorkLegHolds`, memoized, through the caller's own `ReadWorkMeta` seam); a leg that holds it closes the escalation, the federated loop drops the leg that could not resolve it, and the leg that holds it claims it on the next turn of the same invocation.

A read rather than a deferred second pass, deliberately: a second pass over the fan-out was implemented first and it let unrelated **lower-tier** work in a later leg be claimed ahead of an ASSIGNED graph step — the starvation this slice closes, wearing a different hat (`conformanceAssignedTierClaimsTheGraphStep` caught it). A failed leg read counts as "may hold it": one skipped bead reclaimed next tick beats an ownership write in a ledger the reader is not looking at.

```
hook_claim_class_fanout_test.go:128: work-scope claim ran against dirs "rig-a", want rig-a,city:
  every WORK leg must answer before the class escalation opens
```

`TestClassEscalationStillReachesABindingOnlyBead` pins the other side — once no work leg holds the bead the escalation is immediate, on the FIRST leg, at the candidate's own rank.

`hookClaimClassRouteForCity` gates on `graphClassBinding` (store identity, the same question `resolveClassStore` asks) and returns nil; `classRoutedHookClaimOps` then returns the ops **value** it was handed, not a delegating wrapper. Proved by function identity (`reflect.Value.Pointer`) rather than by behaviour, so an unconditional wrap reddens all six seams.

## The two pinned gaps

**I5 — flipped, and the prediction it got wrong is kept.** The KNOWN GAP predicted the closure would add a coordination-class **leg** to `hookWorkQueryStores` and redden the fan-out rows. It could not: a binding is not a bd workspace. Both halves landed beside the fan-out — ga-bvdha swapped the primary leg's *command*, ga-601v2 put the class axis on the *ops*. So the fan-out rows keep their assertions and become the invariant that keeps query and claim apart, and the closure is asserted on the production seams (`hookClaimClassRouteForCity`, `classRoutedHookClaimOps`) with a new `conformanceHookClaimClassRouting` row covering class-resident, work and co-resident ids.

**I15 — flipped.** `conformanceWorkQuerySeesButCannotClaim` → `conformanceWorkQueryClaimsWhatItSees`; `assertClaimStoreCannotResolve` → `assertClaimEscalatesOffTheWorkStore`, which keeps both original halves (the work store still cannot resolve the bead, and still fails with the not-found the claim path classifies) because they are now the *escalation signal*, and adds the landing. `conformanceAssignedTierDoesNotStrandClaimableWork` → `conformanceAssignedTierClaimsTheGraphStep` (the whole federated command over the ranked assigned tier now returns the graph step) plus `conformanceUnresolvableCandidateStillDoesNotStrandWork`, which keeps the fail-soft property for an id NO store holds. Prose rewritten, no row deleted.

Red-before with the routing disabled:

```
claim_class_route_test.go:84:  routed claim of gcg-100 = (ok=false err=bead not found), want a successful claim
claim_class_route_test.go:180: claim over an unreachable binding = (ok=false err=bead not found), want the read failure surfaced
claim_class_route_test.go:288: routed continuation list = [], want exactly gcg-701
split_topology_conformance_test.go:654:  class-resident graph step …: (ok=false err=claiming bead "gcg-wisp-0002": … bead not found), want a successful claim
split_topology_conformance_test.go:1976: `gc hook --claim` claim of the relocated assigned-tier graph bead gcg-wisp-0003 = (ok=false …), want a successful claim escalated to the binding
```

Widening the escalation predicate to any error:

```
claim_class_route_test.go:127: claim = (ok=true err=<nil>), want the work store's own error returned unchanged;
                               only beads.ErrNotFound may send a claim to a second store   (×3)
```

Class-first probe order:

```
claim_class_route_test.go:160: the work-scope claim never ran; the work store is probed FIRST so co-residence resolves to its copy
split_topology_conformance_test.go:654: co-resident id — first-leg-wins, WORK copy: … gc-1 is claimed for … in the store that does NOT own it
```

Wrapping unconditionally:

```
claim_class_route_test.go:390: classRoutedHookClaimOps replaced the Claim seam on a city with no relocated class   (×6 seams)
```

## The release tier, and what is actually still open (ga-zp3uj rescoped)

An earlier revision shipped the claim with the release half unclosed, on the premise that "nothing automated can release a routed claim". **Measured, that premise was wrong in one direction and right in another**, and the difference is which store the scan LEADS with:

| scan | leading store | reaches the binding? |
| --- | --- | --- |
| reconciler — stranded-repair, named retirement, the `closeBead` cascade | `sessStore.Store` = the SESSIONS-class store | **yes**: one binding serves every relocated class (`openStorageRoutes` keys them all to the engine it opened), so the sessions-class store IS the graph binding on a converged split |
| `gc session close` | `openCityStore` = the CITY WORK store | **no** — fixed here |

So the automated crash-recovery path — the one a dead worker actually goes through — already released a binding-resident claim, and the pinned "known gap" row was driving `unclaimWorkAssignedToRetiredSessionBead(e.work, …)`, which is not what production passes on that path. `gc session close` was genuinely blind, and now takes the graph binding as a class leg (`workAssignmentStores` gained an `extra` argument, de-duplicated by store identity so the reconciler's leg is never scanned twice).

`assertClassRoutedClaimHasNoReleaseTier` is therefore replaced by `assertClassRoutedClaimIsReleasable`, which asserts BOTH scans find and release the step. Red-before, with the class leg removed from `workAssignmentStores`:

```
split_topology_conformance_test.go:770: no leg of the release scan is the class binding;
                                        a claim routed there is released by nothing
--- FAIL: …/I5-claim-routing/split/gc_session_close_—_leads_with_the_work_store_and_is_handed_the_binding
--- PASS: …/I5-claim-routing/split/reconciler_scan_—_leads_with_the_sessions-class_store,_which_IS_the_binding
```

**ga-zp3uj is rescoped, not closed**: what remains topology-blind is the AGENT-side recovery — `bd list --status in_progress`, `on_death`, `on_boot` — which are raw `bd` commands in the agent's work directory. Those RE-SERVE a strand rather than lose it, because the reconciler scan above is what releases it. `internal/config`'s `TestQueryKindsWithoutAReadyReadAreTopologyBlind` is unchanged and still pins that query side.

## Two findings worth a reviewer's eye

1. **`EmitExecutionStepStarted` is routed**, and it is the one seam beyond the bead's literal op list (claim, stamp-branch, continuation list/assign, session-pointer record). It routes on the **memo only** — the store the claim landed in — never on a probe. Rationale: its own doc comment asserts "the hook's bd context owns both the claimed graph step and its workflow root", which is false on a split city, and leaving it unrouted would have the now-succeeding stamp hand it a doomed bd subprocess. Easy to drop if a reviewer disagrees; `TestClassRoutedLifecycleEmissionFollowsTheClaimOnly` pins the narrowness either way.
2. **The "session-pointer record" bug the bead warned about is structurally absent from main.** `publishHookClaimRunMap` publishes a run-map *file* and deliberately does not mutate the session bead, so there is no prefix-routed `RecordSessionPointers` to fix.

## Fixture fidelity — the real backend, suite-wide

The split fixture's class leg is now a real `beads.SQLiteStore` under the graph class's reserved prefix (`newSplitEnvClassLeaf`), behind the kit's `SQLiteSemantics` strictness — the store `internal/storebinding/sqlite`'s `OpenEngine` actually opens. `splittest.StrictStore` forwards `Claim` with the mirror of its own `Count` FIDELITY GAP note.

That leaf differs from the kit's MemStore leaf in exactly the two ways this suite exists to catch: it has the compare-and-swap assignment claim a routed claim acquires through, and it has the graph applier whose reverse-of-a-parent-child guard is the one thing the binding enforces that a MemStore does not. `withClaimCapableClassLeg` and its per-row scoping are deleted; see the retraction at the top for why the scoping existed and why it was not a real constraint.

`conformanceGraphRecipe()` loses its hand-written `parent-child` dep and now carries only the single root → `workflow-finalize` `blocks` edge, which is what a compiled graph.v2 formula emits.

## Gates

- `go build ./...`, `go vet ./...` — clean
- `gofmt -l cmd internal` — empty
- `golangci-lint 2.10.1 ./cmd/gc/... ./internal/beads/...` — 0 issues
- `go test ./cmd/gc -timeout 25m` (SERIAL) — ok
- `go test ./internal/testpolicy/... -count=1` — ok (resource census unchanged)
- `make check-split-topology-rows` / `make check-routed-test-rows` — OK

### DIFF
--- a/cmd/gc/claim_class_route.go
+++ b/cmd/gc/claim_class_route.go
@@ -16,7 +16,7 @@ package main
 // hookClaimOps seam, alongside the rig axis in claim_cross_store.go, rather than
 // a parallel claim path.
 //
-// # The order is WORK first, and that is not the by-id order
+// # The order is EVERY work leg first, and that is not the by-id order
 //
 // A claim is a WRITE, so the store is found by PROBE — ask which store holds the
 // bead — never by routing unconditionally on an id prefix. `gc storage migrate`
@@ -39,25 +39,77 @@ package main
 // The claim must agree with the reader that served it, and the failure if it
 // does not is not cosmetic: claiming the class copy while the reader keeps
 // answering from the still-open work copy re-serves the same bead every tick,
-// which is the treadmill this slice exists to end. So the work store is probed
-// FIRST — by running the existing work-scope write and letting it answer — and
-// the binding is reached only where that write proves the bead is not there.
-// Co-residence therefore keeps the WORK copy, byte-identically to today.
+// which is the treadmill this slice exists to end.
+//
+// "The work store" is a PLURAL, and reading it as a singular is how the order
+// above stops being true. The claim runs against one leg of a fan-out
+// (hookWorkQueryStores), and for the rig-scoped agents that are the shipped
+// shape of every worker that claims, the city work store is the leg that runs
+// LAST: the rig store is prepended as the primary and appendCityHookStore
+// appends the city store at the end. So the not-found the FIRST-selected leg
+// returns proves only that THAT leg does not hold the bead — the city work leg
+// behind it may.
+//
+// The escalation is therefore a property of the whole fan-out, not of one leg.
+// Before a not-found opens it, every OTHER leg is READ for the same id
+// (anotherWorkLegHolds), and a leg that holds the bead closes it: the work
+// store's answer stands, the federated loop drops the leg that could not resolve
+// it, and the leg that holds it claims it on the next turn of the same
+// invocation — at its own rank, not behind whatever else the fan-out is
+// carrying. Co-residence therefore keeps the WORK copy on any leg that holds it,
+// byte-identically to today.
+//
+// A read rather than a second pass, because a second pass would let unrelated
+// lower-tier work in another leg be claimed first, which is the starvation this
+// slice closes wearing a different hat. The reads are bounded — one per
+// remaining leg, once per id, only for a bead the leg that answered could not
+// resolve — and claimHookWorkWithRunner is what hands the route the leg set
+// (hookClaimOps.ClassRoute).
 //
 // # The escalation signal is the existing one, unwidened
 //
 // beads.ErrNotFound is the ONLY error that proves the bead is not in the store
-// the write ran against (hookClaimBeadIsElsewhere). A write timeout, store
-// contention or a controller-socket flap leaves ownership unresolved on a bead
-// the session may already own, and must keep failing closed — so those are
-// returned unchanged and never retried against a second store. This file
-// consumes that predicate; it does not widen it.
+// the write ran against (hookClaimBeadIsElsewhere), and it opens the escalation
+// only together with ok=false. A write timeout, store contention or a
+// controller-socket flap leaves ownership unresolved on a bead the session may
+// already own, and must keep failing closed — so those are returned unchanged
+// and never retried against a second store. This file consumes that predicate;
+// it does not widen it.
+//
+// ok=true is the same refusal for the same reason. hookClaimThroughStore returns
+// (claimed, true, err) for exactly one shape — the CAS COMMITTED and the
+// canonical readback then failed — and BdStore.Get produces an
+// ErrNotFound-wrapping error for a plain miss, an empty result AND
+// beads.ErrIDCollision, so that readback failure routinely satisfies the
+// predicate on a claim that landed. Escalating it would claim one logical bead
+// in two ledgers and swallow the signal both call sites stop on, so the
+// committed claim is returned exactly as the work store reported it.
 //
 // A read failure from the binding is likewise an error and never absence:
 // reading "the binding could not answer" as "the bead is not there" is the
 // root-loss shape this whole lane exists to prevent. The one error that is not a
 // fault is the one-shot funnel's standing refusal, handled exactly as
 // classRoutedStoreForID handles it.
+//
+// # A binding that cannot claim is a city fact, not a bead fault
+//
+// The CAS the routed claim acquires through is a capability: *beads.SQLiteStore
+// has the two-argument Claim and the other compiled-in binding provider's engine
+// (beadsworkspace over *beads.NativeDoltStore) does not, so the closed contract
+// answers ErrBeadsAdapterCapability rather than emulating it. That is a standing
+// property of the city's storage configuration, so it is refused at the DOOR —
+// newHookClaimClassRoute verifies it once, the way storebinding's
+// NewBeadsNudgeQueue verifies the same capability at construction — and
+// claimHookWork then runs unrouted with one loud line rather than failing every
+// claim of every bead in every store.
+//
+// If a refusal reaches a claim anyway, it is a per-BEAD skip and not a terminal
+// tick: the escalation only ran because a work store returned not-found, which
+// proves this session owns nothing there and that no mutation is outstanding
+// anywhere (the capability refusal is returned before any write). That is the
+// same proof the work-side not-found carries, so it earns the same skip —
+// hookClaimBindingRefusedTheClaim, kept separate so the work-side predicate
+// stays exactly as narrow as it was.
 //
 // # A single-store city takes none of this
 //
@@ -80,6 +132,12 @@ import (
 	"github.com/gastownhall/gascity/internal/storebinding"
 )
 
+// errClaimRouteBindingCannotClaim reports that the relocated coordination-class
+// binding has no compare-and-swap claim, so no claim-time write can be routed to
+// it. It is a property of the city's storage configuration rather than of any
+// bead, which is why the caller degrades to unrouted claiming instead of failing.
+var errClaimRouteBindingCannotClaim = errors.New("the relocated coordination-class binding cannot claim")
+
 // hookClaimClassRoute is the opened coordination-class front door a claim-time
 // write falls back to, plus the per-invocation record of which bead ids the
 // binding was PROVED to hold.
@@ -93,6 +151,11 @@ import (
 // stores, tryHookClaim walks candidates), and the map dies with the invocation.
 // Its job is to spare the follow-on writes for a bead the claim already resolved
 // a doomed bd subprocess each.
+//
+// workLegs and readLeg carry the fan-out ordering rule: before a not-found from
+// ONE leg is allowed to open the escalation, every other leg is read for the
+// same id. workHeld memoizes that answer for the same reason resident memoizes
+// the binding's.
 type hookClaimClassRoute struct {
 	// class is the raw binding store, needed by the lifecycle projector, which
 	// takes a beads.Store rather than the closed contract.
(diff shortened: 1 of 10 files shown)


julianknutsen:
## Review-council fixes pushed — `cf0b7a8614`

All six majors addressed. Every fix has a row that is red without it; the red-before strings are verbatim from the runs.

### 1 + 2 (findings 2 and 5) — the escalation dropped the `!ok` half

`ops.Claim` passed only the error to `route.routes`, never `ok`. `hookClaimThroughStore` returns `(claimed, true, err)` for exactly one shape — the CAS **committed** and the canonical readback then failed — and `BdStore.Get` wraps `ErrNotFound` for a plain miss, an empty result **and** `beads.ErrIDCollision`, so that readback failure routinely satisfied `hookClaimBeadIsElsewhere` on a claim that had already landed. Fixed with `if ok { return claimed, ok, err }` before the escalation, matching what `claimFirstReadyHookAssignment` already does (`!ok && hookClaimBeadIsElsewhere(err)`).

Red-before (`TestClassRoutedClaimNeverEscalatesACommittedWorkClaim`, both rows), re-verified as a mutation against the final tree:

```
claim_class_route_test.go:426: routed claim returned err=<nil>, want the canonical-readback failure
  reloading claimed bead "gcg-b00": getting bead "gcg-b00": bead not found  returned unchanged;
  swallowing it lets the caller stamp and emit against state it never confirmed
claim_class_route_test.go:426: … reloading claimed bead "gcg-b00": bd resolved a different bead ID
  (substring collision): bead not found  returned unchanged
```

The row also asserts the binding copy stays unclaimed, so the "one bead in two ledgers" half is pinned too.

### 3 (finding 1) — a binding-side claim fault was tick-fatal

Did **both** halves you asked for, minus the boot refusal.

* **At the door.** `newHookClaimClassRoute` verifies the two-argument CAS once — the same check, in the same place, for the same reason as `storebinding.NewBeadsNudgeQueue` — and `hookClaimRouteVerdict` degrades that one error to unrouted claiming with a loud line per invocation. Every other resolution failure still fails closed.
* **Per bead.** `hookClaimBindingRefusedTheClaim` earns the same skip the work-side not-found gets. It is a **separate** predicate: `hookClaimBeadIsElsewhere` is untouched and still `beads.ErrNotFound` only. The two answer questions about different stores — the refusal is reached only after a work store proved this session owns nothing there, and the adapter's type assertion fails before any write, so nothing is outstanding anywhere.

**Why not the `openStorageRoutes` boot refusal.** It would refuse a configuration that boots and serves reads today (`TestStorageGateServesACityFromItsBeadsWorkspace` proves it), i.e. take a city whose graph claims were already impossible and make the whole city unusable. That is strictly larger than the availability bug being fixed, and it is a storage-plan decision rather than a claim-routing one. The door check gives the same loudness cadence without the blast radius.

Control vs claim-incapable leg, re-run on `TestBindingClaimRefusalIsPerBeadNotPerTick` (same fixture, one differs only by the route):

```
before:  control (route=nil)  → exit 0, stdout {"bead_id":"gc-77", …}
before:  claim-incapable leg  → exit 1, stdout "", stderr
         "gc hook --claim: promoting ready assignment gcg-6: beads adapter capability unavailable: assignment claim"
after:   both rows            → exit 0, gc-77 claimed
```

`TestBindingClaimRefusalAloneStillDrains` pins the no-other-work case as a structured `claims_errored` drain rather than a bare exit 1 (red-before: same exit 1, empty stdout).

### 4 (finding 4) — "WORK first" was "first-selected-leg first"

**Chose: make the rule true to its name.** Before a not-found opens the escalation, every OTHER leg of the fan-out is now READ for the same id (`anotherWorkLegHolds`, memoized per id, through the caller's own `ReadWorkMeta` seam). A leg that holds the bead closes the escalation; the loop drops the leg that could not resolve it and the leg that holds it claims it on the next turn of the same invocation.

```
hook_claim_class_fanout_test.go:128: work-scope claim ran against dirs "rig-a", want rig-a,city:
  every WORK leg must answer before the class escalation opens
```

**Worth flagging: the obvious fix was wrong.** I implemented the "defer the escalation until every work leg has drained" version first (a second pass over the fan-out). It reddened `conformanceAssignedTierClaimsTheGraphStep`, because deferring lets unrelated **lower-tier** routed work in a later leg be claimed ahead of an **assigned** graph step — starvation, i.e. this slice's own failure mode wearing a different hat. The read-probe keeps the escalation immediate on the first leg once no work store holds the bead, so the graph step still wins at its own rank; `TestClassEscalationStillReachesABindingOnlyBead` now asserts exactly that (`claimDirs == "rig-a"` alone).

A failed leg read counts as "may hold it": one skipped bead reclaimed next tick beats an ownership write into a ledger the reader is not looking at.

### 5 (finding 3) — the release tier

**Measured rather than argued, and the premise was half wrong.** Which store a release scan LEADS with decides whether it sees the binding:

| scan | leading store | reaches the binding? |
| --- | --- | --- |
| reconciler — stranded-repair, named retirement, `closeBead` cascade | `sessStore.Store` = the SESSIONS-class store | **yes** |
| `gc session close` | `openCityStore` = the CITY WORK store | **no** — fixed |

On a converged split, one binding serves every relocated class (`openStorageRoutes` keys them all to the engine it opened), so the sessions-class store IS the graph binding — and the reconciler's `store := sessStore.Store` therefore already scanned it. Reproduction, driving `unclaimWorkAssignedToRetiredSessionBead` with the store production passes on that path:

```
sessionsStore == class store: true
workAssignmentStores(sessions) leg 0 isClass=true  leg 1 isRig=true
AFTER reconciler-shaped sweep: status="open" assignee=""
```

The pinned known-gap row was driving `unclaimWorkAssignedToRetiredSessionBead(e.work, …)`, which is not what that path passes. `gc session close` was genuinely blind and now hands the graph binding in as a class leg (`workAssignmentStores` gained an `extra` argument, de-duplicated by store identity so the reconciler's leg is never scanned twice). `assertClassRoutedClaimHasNoReleaseTier` → `assertClassRoutedClaimIsReleasable`, asserting both scans find **and release** the step.

Red-before (class leg removed from `workAssignmentStores`) — note which arm survives:

```
split_topology_conformance_test.go:770: no leg of the release scan is the class binding;
                                        a claim routed there is released by nothing
--- FAIL: …/I5-claim-routing/split/gc_session_close_—_leads_with_the_work_store_and_is_handed_the_binding
--- PASS: …/I5-claim-routing/split/reconciler_scan_—_leads_with_the_sessions-class_store,_which_IS_the_binding
```

So there is no unreleasable claim being shipped. **ga-zp3uj is rescoped, not closed**: what remains topology-blind is the agent-side recovery (`bd list --status in_progress`, `on_death`, `on_boot`), raw `bd` in the agent's work directory, which RE-SERVES a strand rather than losing it, because the reconciler scan releases it.

*Noted, out of scope:* the reconciler's scan leads with the sessions-class store and therefore does NOT include the city WORK store on a split city. That asymmetry predates this PR and is unrelated to claim routing; it is not touched here.

### 6 (finding 6) — retracted, everywhere

**My "the real backend refuses graph.v2" finding was false and is withdrawn.** `internal/formula/compile.go` gates both parent-child emitters on `!graphWorkflow` and `addWorkflowRootDeps` emits only the `blocks` edge, so a graph.v2 recipe carries **zero** parent-child deps; the SQLite guard fires only on `sqliteGraphApplyParentDepPairs`, which is built from a `parent-child` dep. The refusal came entirely from `conformanceGraphRecipe()` carrying a hand-written `parent-child` dep no v2 compiler emits.

Retracted in all three places: the PR body (a `> RETRACTION` block at the very top), the `withClaimCapableClassLeg` doc comment (the helper is deleted; the successor `newSplitEnvClassLeaf` states the true cause), and the ga-601v2 bead notes.

And the fixture question is answered rather than deferred: the hand-written dep is gone, and the **real `beads.SQLiteStore` is now the suite-wide class leg**. `withClaimCapableClassLeg` and its per-row scoping are deleted, so every conformance row — not just the claim rows — now runs against a leaf that has both the CAS and the graph applier.

### Untouched on purpose

* `hookClaimBeadIsElsewhere` is byte-identical; only `beads.ErrNotFound` is non-terminal on the work side.
* Single-store byte identity, proved by mutation: wrapping unconditionally reddens all six seams (`classRoutedHookClaimOps replaced the … seam on a city with no relocated class`, ×6) plus `TestClassRoutedHookClaimOpsIsInertWithoutABinding`. `observeWorkLegs` is a nil-receiver no-op.
* I5 (`I5-claim-routing`) and I15 (`I15-work-query-federation`) both still exist and still bite — I5's release row is strengthened from "asserts the gap" to "asserts the release".

### Gates

`go build ./...` clean · `go vet ./...` clean · `gofmt -l cmd internal` empty · `golangci-lint 2.10.1 ./cmd/gc/... ./internal/beads/...` **0 issues** · `go test ./cmd/gc -timeout 25m` (SERIAL) **ok 758.508s** · `go test ./internal/...` ok · `go test ./internal/testpolicy/... -count=1` ok · `scripts/check-core-boundary.sh` OK · `make check-split-topology-rows` OK · `make check-routed-test-rows` OK.

Branch is on current `origin/main` (`ede16cacde`); no rebase was needed. Not merged.

PR #5167, opened 2026-08-10
fix(cli,beads): refuse a short `gc bd ready` frontier and a work-store read that cannot see its own ledger (ga-jbn6f, ga-qi9km)

julianknutsen:
Fixes two silent-empty answers a **live converged city** produced that the whole test suite could not see (`/var/tmp/win3-splitstore-live-proof.md`). Both are fail-opens: a confident short answer with exit 0, in exactly the state Invariant 0 exists to make loud.

## ga-jbn6f — `gc bd ready` still had the original bug

Measured on the proof city with the controller up:

```
gc ready                 n=9 gcg=3   ≡ the API's 9/3, id for id
GET …/beads/ready        n=9 gcg=3
gc bd ready --json       n=5 gcg=0   exit 0
```

…and it *still* returned exit 0 with the relocated binding `chmod 000`. This program fixed bulk blindness by adding a NEW command (`gc ready`) and swapping the generated `work_query` onto it; #5162 then guarded `bd list/ready/search --metadata-field <k>=<relocated id>`. **Neither repaired the invocation win35 actually measured**, and `ga-iaj7k`'s own acceptance criterion is worded about it.

### The trigger is the topology, not the argv — and that is the whole difference from #5162

In #5162 the *selector* named a class id, so the empty answer was provably wrong **for that predicate**, and argv is the right thing to classify: `bd list --title-contains gcg-1` is a question this ledger really can answer.

`bd ready` is not that shape. It computes a **frontier** over the one ledger `bd` resolves from the working directory, and takes no selector that could reach another store — so on a split city its result is the work-class SUBSET of the city's ready set, short by exactly the beads the split moved, **for every invocation, including the bare one with no flags at all**. Anchoring on argv would have guarded the rare `gc bd ready --metadata-field gc.root_bead_id=<gcg root>` and left the bare `gc bd ready` — the one an operator types, and the one the report measured — answering short and quiet.

So `ready` moves from `bdRelocatedClassGuardedVerbs` into a new `bdRelocatedClassBlindVerbs`, checked before any argument is read. It was **moved rather than added alongside**: the topology arm runs first, so a scan registered for the same verb is an unreachable branch that reads as coverage. `TestBdRelocatedClassGuardCoversEverySelectorVerb` now asserts the union of the two maps covers every `--metadata-field` verb bdflags knows, *and* forbids the overlap.

The refusal reuses the existing machinery (`bdSQLRelocatedClassRefusal`, `ErrBdSQLClassRelocated`, the shared class renderer) and names the class, the id prefix, the binding and its provider/path, why bd answered emptily, and `gc ready` as the federated replacement. `GC_BD_ALLOW_RELOCATED_CLASS_READ=1` **still runs it**, loudly, and the message names that knob with wording that fits this arm (there is no misclassified predicate to appeal — what the override buys is a deliberate one-ledger frontier).

**Single-store cities are byte-identical**: `relocatedBeadClasses` is empty, `bdSQLRelocatedClassRefusal` returns before either map is consulted, and the same argvs a split city now refuses reach bd verbatim (`TestGcBdReadyIsByteIdenticalOnASingleStoreCity`). Generated work queries are unaffected either way — they invoke raw `bd ready` (single-store tiers) or bare `gc ready` (split tiers), never `gc bd ready`.

## ga-qi9km — a fail-open on the WORK leg, and not split-store specific

`gc rig add` and `gc supervisor run` rewrite a city's `.beads/metadata.json` from `dolt_mode: embedded` to `server`. Server databases live in `.beads/dolt`, embedded ones in `.beads/embeddeddolt/<db>`, and **nothing copies rows between them** — there is no migration path in the tree. After the flip every work-store read returned `[]` **exit 0**.

That is a projection that cannot see its **own ledger**, and it defeats the federation's loud-leg guarantee from the inside: `gc ready` can fail loudly on a dead rig or an unreadable binding, but a city work leg that answers empty because its metadata moved underneath it makes every leg look healthy. **A single-store city is bitten identically** — there is no coordination class involved, only a workspace pointed at the wrong database.

### What the rewrite is FOR

Investigated before touching it. gc's managed bead store is a **Dolt server** that the controller, every agent's `bd`, and the dashboard open concurrently; an embedded scope opens the Dolt directory in-process, which those concurrent readers cannot share (`internal/beads/contract/preflight_checker.go`: `dolt_mode=embedded; native store requires Dolt server mode`). Canonicalising to server mode is what makes a scope usable by a running city at all. **It is kept.**

### (1) and (2) were separable; both landed

**(1) The rewrite announces itself before it happens** — naming the scope, both modes, and, when the previous mode still has a Dolt repository on disk, the exact database the scope is about to stop reading and what to do about it. It is **not refused**: refusing would break adopting a freshly `bd init`-ed workspace, whose `.dolt` directory exists from the moment bd initializes it, and gc cannot tell an empty repository from a populated one without opening Dolt. That is stated in the code, not assumed.

The notice goes to a package sink (`os.Stderr` by default) rather than a threaded writer because the flip has six entry points, three of which reach it through `initAndHookDir` — whose signature is the `rig.Deps.InitAndHook` contract — and because a threaded writer is how this went unseen: `normalizeCanonicalBdScopeFiles` already takes one, and `gc rig add` passes `io.Discard`. `defaultV2MigrationWarnSink` and `cliStorageStderr` are the in-tree precedents.

**(2) The read-time refusal does NOT land — see "Council review" below.** `BdStore.List` and `BdStore.Ready` are byte-identical to `origin/main`; the fail-open they would have closed is still open, deliberately, because closing it at read time needs evidence gc does not have without opening the second Dolt database. What replaces it is a diagnostic that actually covers the shape: `gc doctor`'s `bd-split-store` check used to require BOTH `.beads/dolt` and `.beads/embeddeddolt` to exist, so it reported "no legacy split store detected" for the flip's own primary shape (the flip re-points the ledger before any `.beads/dolt` exists). It now warns when a Dolt repository sits in the mode directory metadata does not name — same evidential standard as the existing arm, same fix hint, still a warning.

**Single-store effect, stated plainly:** the announcement fires on a single-store city identically — the defect has nothing to do with coordination classes — and no read behaviour changes anywhere.

## Council review: what changed after the first push

Two blockers and seven majors came back. Six of the nine were one defect (the read-time orphaned-store guard) and it is removed rather than narrowed. Full reasoning in the follow-up commit message; the short version:

### The read guard would have failed a healthy city closed

It refused an empty `List`/`Ready` whenever a `.dolt` directory sat under the mode directory `metadata.json` does NOT name. Three experiments, reproduced against `eb287a4a64`:

1. `dolt_mode=server` scope with a retained `.beads/embeddeddolt/jc/.dolt`: `List = (1 row, nil)` — the active store is demonstrably populated — but `Ready = (0, ErrOrphanedBeadStore)`.
2. The check sat **after** client-side filtering, so `Ready{Assignee}`, `Ready{TierWisps}`, `Children(leaf)` and an empty mail poll were refused on a store bd had just answered with rows. The code's own justification ("a scope that answers with rows pays nothing") is a claim about the *store*; the check was a claim about one *call's result*.
3. Plain `bd init -p X` — bd's **default** — creates an EMPTY `.beads/embeddeddolt/X/.dolt`. gc canonicalizes it to server on `gc rig add`, and the guard then fired forever on a city that was never broken, including inside `verifyCanonicalBdScopeStoreReady`, which gates adoption on `List(AllowScan, Limit 1)` returning nil.

Blast radius of any one: `federateBeadLegs` aborts the whole federation on a leg error → `gc ready` exits non-zero → the federated work query's `|| exit $?` fails every agent's `gc hook` → the API ready arm hits `totalOutage()`. Nothing in the tree matched `ErrOrphanedBeadStore`, and there was no override.

**Narrowing it does not rescue it.** Requiring the ACTIVE store to be provably empty defeats (1) and (2) but not (3): in a fresh adoption the active store *is* empty, and "empty active store beside a second Dolt directory" is the same on-disk shape as the bug it must catch. Separating them requires opening the second database — a store open, a lock and a possible schema migration on the directory the operator was told to preserve, and the one thing the guard's own design says it must not do. `gc doctor`'s `splitStoreFixHint` tells operators to *"keep both directories until reconciled"* — the exact state the guard turned into a hard error — and the guard's own false-positive budget forbids it: *"a guard that fired on any of them would turn an empty ledger into a hard error on cities that are working correctly, which is a worse failure than the one being fixed."*

Shipping ga-jbn6f plus the announcement is strictly better than shipping that. The read-time refusal is split out.

### The announcement is corrected, not just kept

* It no longer claims later empty reads are refused (they are not).
* It no longer offers "point `metadata.json` back", which every `gc start` / `gc rig add` / `gc supervisor run` undoes — it says so instead.
* It carries `gc doctor`'s own remediation verbatim on the load-bearing clause ("keep both directories until reconciled"), so the two stop contradicting each other.
* `gc rig set-endpoint` and `gc beads city use-managed`/`use-external` reach their own canonicalizer and performed the identical flip in silence; they announce now too.

### `gc bd list --ready` was still answered short

The topology arm keyed on the VERB, so `gc bd list --ready` — which bd documents as *"same semantics as bd ready"* and dispatches to the same `GetReadyWork` store methods — computed the identical short frontier and exited 0. It also minted the sharper inversion the guard exists to retire: `list --ready --metadata-field gc.root_bead_id=<gcg>` was refused by the argv scan while the bare frontier was answered short.

The trigger is now "this invocation computes bd's ready frontier", by verb **or** by flag, with the flag's verbs derived from `bdflags.BoolFlags`. `--ready=false` and `-r` (bd's `--reverse` shorthand, not a ready switch) stay ordinary ledger reads. `TestBdRelocatedClassGuardCoversEveryFrontierSurface` is the completeness pin the selector one could not be — the old pin derived only from `--metadata-field`, a VALUE flag, and was structurally blind to a BOOL-flag frontier.

### The steer was false

The refusal, the CLI long help, `docs/reference/cli.md` and the runbook all said `gc ready` is *"flag-compatible with `bd ready`"*. It is not: `gc ready` registers ten flags and rejects `--label`, `--label-any`, `--parent`, `--type`, `--priority`, `--offset`, `--has-metadata-key`, `--mol`, `--include-deferred`, `--gated`, `--claim` and every single-letter shorthand — and its `--sort` takes `oldest|newest` where bd's **default** is `priority`. Two argvs this PR's own tests enumerate as refused had no working replacement.

All four sites now say what is true (*flag-compatible with the `bd ready` invocation the generated work query builds*) and name the gap the way the sibling refusal already does. `TestGcReadySteerDescribesTheFlagsItActuallyAccepts` derives both sides from cobra and `bdflags`, so the sentence cannot drift from either. The shipped core skill and the hyperscale worker prompt now say which spelling a split city refuses and what replaces it.

## Red before

Real on-disk city/scope fixtures, with only bd's subprocess faked — and faked to do exactly what a healthy bd does, which is answer and exit 0.

```
`gc bd ready` exited 0 with stdout="[{"id":"jc-026"},{"id":"jc-3nj"},{"id":"jc-aj3"},
{"id":"jc-cpd"},{"id":"jc-wve"}]"; that is the confident short frontier — no gcg- row,
no warning, no non-zero exit — that `gc ready` and the API both contradict on the same city
      (all 8 argvs: bare, --json, --unassigned, pool demand, behind a root flag,
       --label, --limit, --assignee)

bdSQLRelocatedClassRefusal([ready]) refused = false, want true
bdSQLRelocatedClassRefusal([--json ready]) refused = false, want true

the storage-mode change never names "…/.beads/embeddeddolt/jc"; notices=""
the storage-mode change never names "STOP reading"; notices=""
the rewrite was silent on a single-store city
```

Red before the review fixes, at `eb287a4a64`:

```
EXP1  List  = (1 beads, <nil>)
      Ready = (0 beads, bead store read returned empty while an unread bead database sits
              beside it: bd ready returned no rows for <scope> from the "dolt" store, but
              <scope>/.beads/embeddeddolt/jc is a bead database this scope's
              .beads/metadata.json does not point at …)
EXP2  Ready{Assignee} / Ready{TierWisps} / Children(leaf) / List(mail) — same refusal,
      on a store bd answered with a row
EXP3  List(AllowScan, Limit 1) = (0 beads, …ErrOrphanedBeadStore)
      Ready()                  = (0 beads, …ErrOrphanedBeadStore)   [fresh `bd init` adoption]
EXP4  doBd([list --ready --json]) = 0; bd got "list --ready --json";
      stdout = the identical 5-row work-class-only array
      doBd([list --ready=true --json]) = 0 ; doBd([list --json --ready]) = 0
      bdSQLRelocatedClassRefusal(split, [list --ready --json]) = ("", false)
EXP5  gc ready --label → unknown flag: --label   (also --label-any, --parent, --type,
      --priority, --offset, --has-metadata-key, --mol, --include-deferred, --gated, --claim)
      gc ready -u / -n 1 → unknown shorthand flag
      …while the refusal said "flag-compatible with `bd ready`"
```

**Mutation proofs.** Neutering the frontier-flag arm fails the `list --ready` rows; treating `--ready=false` as on fails both the negative and single-store byte-identity; restoring the read guard fails all three experiment tests; restoring the blanket flag-compat claim fails the steer pin.

## Gates

`go build ./...` · `go vet ./...` · gofmt clean · **golangci-lint 2.10.1: 0 issues** · `go test ./cmd/gc -timeout 25m` serial (734s, ok) · `go test ./internal/...` · `scripts/check-core-boundary.sh` · `make check-split-topology-rows` · `make check-routed-test-rows` · `test/docsync` · `test/packlint` · `cmd/genschema` regenerated (`docs/reference/cli.md`).

`docs/runbooks/split-storage-classes.md` carries the operator note for the refusal, the `gc bd list --ready` spelling, the `gc ready` flag gap, and the `gc`-on-`PATH` requirement the live proof found undocumented.


julianknutsen:
## Council review addressed — `63f65d5`

Two blockers and seven majors came back. **Six of the nine were the same defect** (the read-time orphaned-store guard) and it is **removed rather than narrowed**. Two independent majors are fixed directly, and the six minors are taken.

### Blockers #0 + #1, majors #2 #3 #4 #6 — the orphaned-store read guard: split out

The guard refused an empty `BdStore.List`/`Ready` whenever a `.dolt` directory sat under the mode directory `metadata.json` does **not** name. Three experiments, reproduced against `eb287a4a64` before touching anything:

| # | fixture | result at head |
|---|---|---|
| 1 | `dolt_mode=server`, retained `.beads/embeddeddolt/jc/.dolt`, bd answers `list` with a row and `ready` with `[]` | `List = (1 beads, <nil>)` but `Ready = (0 beads, ErrOrphanedBeadStore)` |
| 2 | same, bd answers with rows; `Ready{Assignee}`, `Ready{TierWisps}`, `Children(leaf)`, mail poll | all four `ErrOrphanedBeadStore` — **the store answered with rows and was still refused** |
| 3 | `bd init -p jkq` (bd's DEFAULT: embedded + empty `.dolt`) canonicalized to server | `List(AllowScan, Limit 1) = (0, ErrOrphanedBeadStore)`; `Ready() = (0, ErrOrphanedBeadStore)` |

Experiment 2 falsifies the guard's own stated bound — *"a scope that answers with rows pays nothing"* is a claim about the **store**, and the check was a claim about one **call's result**, evaluated after `IsReadyCandidateForTier`, `q.Assignee` and `applyListQuery`.

**Why the prescribed four-part fix does not rescue it.** (a)+(b)+(c) defeat experiments 1 and 2. They do **not** defeat 3: in a fresh adoption the ACTIVE store is genuinely empty, so "active store provably empty" still fires — and `verifyCanonicalBdScopeStoreReady` gates `gc rig add`/`gc start` on `List(AllowScan, Limit 1)` returning nil error (retried 20×500ms), so the adoption command itself breaks after a ~10s stall. "Empty active store beside a second Dolt directory" is **the same on-disk shape** as the bug the guard exists to catch; separating them requires **opening** the second database — a store open, a file lock and a possible schema migration on the directory the operator was told to preserve, and the one thing the guard's own design says it must not do ("no store open, no server and no network").

`gc doctor`'s `splitStoreFixHint` tells operators to *"keep both directories until reconciled"* — the exact state the guard turned into a hard error — and the guard's own budget forbids it: *"a guard that fired on any of them would turn an empty ledger into a hard error on cities that are working correctly, which is a worse failure than the one being fixed."*

So, per the brief's own instruction: **shipping ga-jbn6f alone is better than shipping an outage.** `BdStore.List` and `BdStore.Ready` are byte-identical to `origin/main` (`git diff origin/main -- internal/beads/bdstore.go` is empty). `ErrOrphanedBeadStore`, `OrphanedBeadStoreRefusal`, `OrphanedBeadDatabase` and `emptyReadIsUntrustworthy` are gone; `orphaned_store.go` → `bead_database_dirs.go`, holding only the shared directory fact the announcement needs. No override knob is shipped because there is no guard to escape.

Three new tests are the ones the whole suite missed, and each is red if the guard is restored (mutation-proved):
`TestAnEmptyReadIsNotEvidenceTheScopeIsReadingTheWrongDatabase` (experiments 1 + 2, five read shapes), `TestAdoptingAFreshlyInitializedWorkspaceStillReads` (experiment 3), `TestTheStorageModeAnnouncementDoesNotChangeWhatAReadAnswers`.

### Major #5 + minors #0/#4 — the announcement and `gc doctor` now agree

* The announcement no longer claims later empty reads are refused (they are not).
* It no longer offers "point `metadata.json` back" — the recovery every `gc start`/`gc rig add`/`gc supervisor run` undoes. It says the edit does not hold, and names the durable route.
* It carries `gc doctor`'s remediation verbatim on the load-bearing clause ("keep both directories until reconciled"), cross-checked by running the real `BDSplitStoreCheck` against the scope the announcement printed for.
* `gc doctor`'s `bd-split-store` check reported **"no legacy split store detected"** for the flip's own primary shape, because it required BOTH directories and the flip re-points the ledger before `.beads/dolt` exists. It now warns when a Dolt repository sits in the mode directory metadata does not name — same evidential standard, same fix hint, still a warning, with a false-positive budget of its own.
* Minor #3: `gc rig set-endpoint` and `gc beads city use-managed`/`use-external` performed the identical flip in silence through their own canonicalizer; they announce now (`TestEveryDoorThatFlipsTheStorageModeAnnouncesIt`).

### Major #7 — `gc bd list --ready`

Red at head: `doBd(["list","--ready","--json"])` → **exit 0**, argv forwarded verbatim, stdout = the identical 5-row work-class-only array; `bdSQLRelocatedClassRefusal(split, ["list","--ready","--json"]) = ("", false)`. Same for `--ready=true` and `["list","--json","--ready"]`.

The trigger is now "this invocation computes bd's ready frontier", by verb **or** by flag, with the flag's verbs derived from `bdflags.BoolFlags` rather than restated. `--ready=false` stays an ordinary ledger read; `-r` is bd's `--reverse` shorthand (`cmd/bd/list.go` `BoolP("reverse","r")`) and is untouched. The refusal names the read as `bd list --ready`, not `bd list`.

`TestBdRelocatedClassGuardCoversEveryFrontierSurface` is the pin the selector one structurally could not be — the old pin derived from `--metadata-field`, a VALUE flag, and was blind to a BOOL-flag frontier.

### Major #8 — the steer was false

`gc ready` registers ten flags; `bd ready` carries ~25. Reproduced rejections: `--label`, `--label-any`, `--parent`, `--type`, `--priority`, `--offset`, `--has-metadata-key`, `--mol`, `--include-deferred`, `--gated`, `--claim` → `unknown flag`; `-u`, `-n` → `unknown shorthand flag`. And `--sort` is shared by name but incompatible by value: `gc ready --sort priority` — bd's own **default** — is rejected.

All four sites (`RelocatedClassFrontierRefusal`, `cmd/gc/cmd_bd.go` long help, `docs/reference/cli.md`, the runbook) now say *flag-compatible with the `bd ready` invocation the generated work query builds*, enumerate what `gc ready` takes, name the gap, and note the `--sort` values. `internal/config/workquery.go`'s comment is scoped to the same truth. `TestGcReadySteerDescribesTheFlagsItActuallyAccepts` derives both sides from cobra and `bdflags` and splits the message at its own "rejects the rest" marker, so a flag can only be advertised on the side it belongs to.

### Minor #2 + #5

The shipped core `gc-work` skill and the hyperscale worker prompt now say which spelling a split city refuses and what replaces it (including that `gc ready` has no `--label`, so pool demand routes through `--metadata-field gc.routed_to=<pool>`). The frontier escape-hint branch is no longer dead to the suite: `TestGcBdReadyRefusesTheWholeFrontierOnASplitCity` asserts the frontier wording appears **and** that the selector arm's "merely REFERENCE such an id" wording does not.

### Not weakened

`TestGcBdReadyRefusesTheWholeFrontierOnASplitCity` is unchanged in scope — all eight argvs still refuse, and `gc bd ready` returning a silent short answer on a split city stays refused. Single-store byte-identity is proved by mutation: `TestGcBdReadyIsByteIdenticalOnASingleStoreCity` gains `list --ready`, `--ready=true`, `--ready=false` and `-r` rows, and treating `--ready=false` as on fails it.

### Gates

`go build ./...` · `go vet ./...` · gofmt clean · **golangci-lint 2.10.1: 0 issues** · `go test ./cmd/gc -timeout 25m` serial → `ok 734.568s` · `go test ./internal/...` → ok · `scripts/check-core-boundary.sh` · `check-split-topology-rows` · `check-routed-test-rows` · `test/docsync` · `test/packlint` · `cmd/genschema` regenerated.

Issue #4392, opened 2026-07-17
gc init cannot select the built-in Pi provider


JPFrancoia:
## Disclaimer

While this issue was opened and filled directly from Pi using GPT 5.6 sol high, I manually checked all the provided details and I confirm that they're correct.

## Summary

`gc init` cannot select Pi Coding Agent even though Pi is installed, configured, on `PATH`, and available as the built-in `pi` provider.

## Environment

- Gas City: `1.3.5`
- Pi Coding Agent: `0.80.7`
- OS: Arch Linux

## Reproduction

```console
$ command -v pi
/home/djipey/.npm-global/bin/pi

$ pi --version
0.80.7

$ gc version
1.3.5

$ gc init ./my-city
gc init: no configured coding agents found; configure your coding agent and restart the wizard
```

Trying to bypass the wizard also fails:

```console
$ gc init --template gascity --default-provider pi --skip-provider-readiness --no-start ./my-city
gc init: unknown provider "pi" (expected one of: claude, codex, gemini, mimocode, antigravity)
```

## Expected behavior

Either:

1. the interactive wizard detects configured Pi installations, or
2. `--default-provider pi --skip-provider-readiness` accepts the built-in Pi provider without requiring a readiness probe.

## Current behavior / likely cause

Pi is a built-in provider in `internal/worker/builtin/profiles.go`, but it is absent from `supportedProviderReadiness` in `internal/api/handler_provider_readiness.go`.

Both the interactive wizard (`configuredWizardProviderChoices`) and explicit provider validation (`normalizeInitProvider`) are derived from `api.ProviderReadinessNames()`. Consequently, a built-in provider without an onboarding readiness probe cannot be selected even when `--skip-provider-readiness` is supplied.

This is also present on current `main` at the time of filing.

## Confirmed workaround

Scaffolding with a readiness-aware placeholder provider, then changing `city.toml` to Pi, works:

```bash
gc init ./my-city \
  --template gascity \
  --default-provider claude \
  --skip-provider-readiness \
  --no-start

sed -i \
  -e 's/provider = "claude"/provider = "pi"/' \
  -e 's/\[providers\.claude\]/[providers.pi]/' \
  -e 's/base = "builtin:claude"/base = "builtin:pi"/' \
  ./my-city/city.toml

gc start ./my-city
```

The resulting config resolves `pi → builtin:pi`, installs `.pi/extensions/gc-hooks.js`, and the named mayor session starts successfully.

Related broad tracker: #672.

PR #5168, opened 2026-08-10
feat(api): add pi readiness probe recognizing ~/.pi/agent/auth.json

m-u-xyz:
## Summary

Register "pi" in the provider readiness registry with a probe that treats the existence of ~/.pi/agent/auth.json as the configured signal (pi writes that file on login). The probe deliberately does not parse the contents, to avoid coupling to pi's evolving auth.json schema. This lets `gc init` detect pi without --skip-provider-readiness, so `gc init --default-provider pi` is no longer rejected with "unknown provider".

Mechanics: adds "pi" to supportedProviderReadiness, supportedReadiness, and readinessProbeSpecs (kind=provider), and wires a "pi" case into providerStatusFixHint for preflight remediation hints.

Tests:
- TestReadinessRegistrySync: "pi" joins the pinned readiness keys/order.
- TestCmdInitSkipProviderReadinessAllowsBuiltinWithoutProbe: "pi" was the probe-less exemplar guarding #4392; now that pi has a probe, swap to "omp" (Oh My Pi), another real builtin still without a probe. The test's own comment anticipated this swap.
- Add probePi unit tests (not_installed / needs_auth / configured) and TestNormalizeInitProviderAcceptsPiWithoutSkip, locking in that --default-provider pi no longer requires --skip-provider-readiness.

Default readiness items (defaultProviderReadinessItems / defaultReadinessItems) are unchanged, so /provider-readiness and /readiness default-query semantics and the OpenAPI/generated-client doc strings are unaffected (genspec is a no-op).

Issue: #5169 

## Testing

- [X] `make check`

## Checklist

- [X] Linked an issue, or explained why one is not needed
- [X] Added or updated tests for behavior changes

### DIFF
--- a/cmd/gc/init_provider_readiness.go
+++ b/cmd/gc/init_provider_readiness.go
@@ -432,6 +432,15 @@ func providerStatusFixHint(probeName, status string) string {
 		case api.ProbeStatusProbeError:
 			return "check ~/.gemini/settings.json and oauth_creds.json"
 		}
+	case "pi":
+		switch status {
+		case api.ProbeStatusNeedsAuth:
+			return "authenticate pi so it writes ~/.pi/agent/auth.json"
+		case api.ProbeStatusNotInstalled:
+			return "install the pi coding agent"
+		case api.ProbeStatusProbeError:
+			return "check ~/.pi/agent/auth.json and the local pi installation"
+		}
 	}
 	return ""
 }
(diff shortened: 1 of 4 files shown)


quad341 (approved):
Thanks @smurfix — this is a genuinely careful piece of work, and two things stand out.

The `omp` swap in `TestCmdInitSkipProviderReadinessAllowsBuiltinWithoutProbe` is the part most PRs would have missed. Giving pi a probe would have quietly hollowed out the #4392 regression guard; you noticed, picked another real probe-less builtin, kept the self-asserting preconditions so it fails loudly instead of rotting, and left a note for whoever swaps it next.

And thanks for calling out that the default readiness items are untouched. I checked — it holds; the generated artifacts check agrees.

One reviewer wondered whether pi also authenticates via an API-key env var. Our own docs answer it: `harness-recipes.md` lists pi as `login`, and `upstream_test.go` asserts it has no upstream env binding — so auth.json-existence is the right signal. Your note about not coupling to an evolving schema is sound and I'm taking it as-is.

Worth knowing: pi configs now hard-block `gc init` preflight when unauthenticated, where before they only warned. That's the intended effect of having a probe.

Merging. Thanks for making pi first-class in `gc init`.


---

PR #5172, opened 2026-08-10
fix(reaper): exclude gc-injected scaffolding from the worktree disposal gate (gas-wr8)

AJBcoding:
## Problem

gc writes its own files into every worktree it creates: provider-native MCP config, materialized skill symlinks, and an appended "# Gas City" `.gitignore` block. All three land on tracked-or-visible paths, so `git status --porcelain` in a healthy, untouched gc worktree is never empty.

Every disposal gate written as "porcelain is non-empty => keep" therefore refuses forever. The reaper's git-safety gate (`HasUncommittedWork()` in `cmd/gc/bead_worktree_reaper.go`) protected 100% of managed worktrees: recovered worktrees accumulate on disk indefinitely, and the one signal that should mean "real uncommitted work is about to be destroyed" becomes indistinguishable from background noise. Measured on a live rig: 11 porcelain entries in an untouched polecat worktree — an identical set to the rig root, all gc scaffolding.

## Fix

New `internal/scaffold` package computes **residue**: porcelain entries minus what gc can *prove* it owns, from records gc itself wrote at injection time:

- **skill symlinks** — named in the sink's `.gc-skill-ownership.json` (sink location discovered, not hardcoded)
- **MCP config** — marked in `.gc/mcp-managed/<provider>.json`, and only while the file still hashes to what gc last wrote (`content_sha256`, newly recorded by materialize)
- **`.gitignore`** — only when removing gc's own parsed "# Gas City" blocks restores the committed content exactly

Subtraction is one-directional: anything unprovable stays residue, so the failure mode is a worktree that lives too long — never one deleted with real work in it. `.gitignore` and `.mcp.json` are shared tracked files an agent may legitimately edit, so neither is excluded by path — the block is parsed and the config is hashed. Markers predating the hash prove adoption but not content, and fail closed until gc next rewrites them. Deletions are never subtracted.

Wired into the reaper's git-safety gate in `cmd/gc/bead_worktree_reaper.go`. `internal/git` stays generic — it gains `StatusPorcelainUntrimmed` (porcelain v1 is fixed-width; the trimmed variant strips the leading space of an unstaged first record and shifts it out of column alignment) and `ShowAtHead`, but no knowledge of gc scaffolding. The composition lives in `cmd/gc` beside the gate that needs it.

## Not in this PR (tracked separately)

- Three sibling gates with the same bare-porcelain shape: `cmd/gc/session_worktree_prune.go` (2 sites), `cmd/gc/agent_home_worktree_cleanup.go`, `internal/doctor/checks_semantic.go` (gas-o01)
- The mol-witness-patrol formula's DIRTY gate, which lives in gascity-packs (gas-o11)

## Validation

- `internal/scaffold`, `internal/materialize`, `internal/git`: full package tests green at this branch tip on macOS
- `go test ./cmd/gc -run 'TestWorktreeHasUnsalvagedWork'`: both new tests green against a real git-repo fixture shaped like a live worktree, including a guard that the fixture trips the old gate
- `go vet` clean on all touched packages

### DIFF
--- a/cmd/gc/bead_worktree_reaper.go
+++ b/cmd/gc/bead_worktree_reaper.go
@@ -16,6 +16,7 @@ import (
 	"github.com/gastownhall/gascity/internal/events"
 	"github.com/gastownhall/gascity/internal/git"
 	"github.com/gastownhall/gascity/internal/pathutil"
+	"github.com/gastownhall/gascity/internal/scaffold"
 	"github.com/gastownhall/gascity/internal/sling"
 )
 
@@ -287,10 +288,12 @@ func reapClosedBeadWorktrees(
 			// it as a clean answer would fail open.
 			if reason == "" {
 				wg := git.New(worktreePath)
-				hasUncommitted := wg.HasUncommittedWork()
+				hasUncommitted, uncommittedErr := worktreeHasUnsalvagedWork(wg, worktreePath)
 				hasUnreachable, unreachableErr := wg.HasUnreachableCommitsResult()
 				hasStashes, stashErr := wg.HasStashesResult()
 				switch {
+				case uncommittedErr != nil:
+					reason = fmt.Sprintf("git probe failed (failing closed): %v", uncommittedErr)
 				case unreachableErr != nil:
 					reason = fmt.Sprintf("git probe failed (failing closed): %v", unreachableErr)
 				case stashErr != nil:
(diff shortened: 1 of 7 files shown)


csauer02-personal-user:
We hit the same refuses-forever class on this gate from a *pack*-injected artifact: our pack's prepare-worktree swaps a large test corpus for a symlink in every worktree it provisions (and records that it did so at injection time), so `git status` is never clean and the disposal gate kept every recovered worktree alive — same shape as gas-wr8, different injector.

Your subtract-only-what-the-injector-can-prove design fits this case exactly; the only gap is that the three subtraction classes are gc's own. Would you take a follow-up that lets a residue subtraction be declared by an injection-time ownership record (same proof discipline as `.gc-skill-ownership.json`, but writable by whatever provisioned the worktree — pack scripts included)? Unprovable stays residue, one-directional as in your design.

We currently run a narrower fork patch for our case and would gladly delete it in favor of building that extension on your mechanism once this lands. Receipts from our fleet if useful: 102 of 134 live worktrees were reap-ineligible purely from this artifact class before we patched.

quad341:
@AJBcoding — this is a strong fix, and the discipline around it is what sold me: parsing the `# Gas City` block instead of path-ignoring `.gitignore`, hashing the MCP target instead of trusting the marker, reading the hash back off disk so the merge-path providers are covered, never subtracting a deletion. And the test that asserts your fixture trips the *old* gate is the thing that keeps this suite honest a year from now.

I'm taking three fixes on my side so you don't have to round-trip. CI is red on one missing generated `testenv_import_test.go` for the new package — pure boilerplate, there's a generator. Then two I want to close before this guards real work: manifest-owned skills are subtracted by name, so a regular file replacing a materialized symlink would be swallowed — the manifest already stores the canonical target, so a `Readlink` compare evens it up with what the materializer itself checks. And because the injector appends its block at EOF with no trailing blank line, an entry appended to `.gitignore` afterwards parses as inside gc's block; I reproduced it locally. I'll gate block-stripping on gc's own known entries so anything else stays residue.

Your design and attribution stay intact — this is integration work, not a rewrite.


---

AJBcoding:
@quad341 thanks for the heads-up — understood, standing by for your follow-up. Both integration catches sound right (the Readlink compare on manifest-owned skills, and gating block-stripping on gc's own known entries) and keep the one-directional design intact.

@csauer02-personal-user — a provision-time ownership record, writable by whatever provisioned the worktree, is exactly the extension shape this was built to leave room for: proven artifacts get subtracted, unprovable stays residue. Glad to have it build on this mechanism once it lands, and 102/134 reap-ineligible is a striking confirmation of the class. We'll watch here for the maintainer follow-up.

PR #5177, opened 2026-08-10
fix(scripts): make the test harness own the process trees it spawns (gas-px0)

AJBcoding:
The sharded test harness ran `go test` as a bare foreground call in the caller's process group, with no trap and no external watchdog — so a shard binary that defeats Go's own `-test.timeout` (non-preemptible loop, wedged runtime) outlives its dead parent chain and orphans to PID 1. Measured instance (kit-k72d, routed as gas-px0): a `gc.test` at PPID 1 ran **5h34m at ~99% CPU** past its `-test.timeout=20m`, parent chain gone, nothing reaped it. A second, cheaper variant accumulates litter instead of load: socket-parked tmux servers from test runs (9 live on the triage host, oldest 2h, all traceable to one fixture).

## What this changes (scripts/ only — 4 files, no production Go)

- **`scripts/lib/harness-reap.sh`** (new): the mechanism. Portable process-group ownership via bash job control (`set -m` + group kill — macOS has no `setsid(1)`; bash 3.2-safe, which the shard runner already tests against).
- **`scripts/test-go-test-shard`**: each product run gets its own process group; the runner traps `EXIT/INT/TERM/HUP` and tears the group down; an external watchdog at `GO_TEST_TIMEOUT` + grace catches a run that defeats Go's internal timeout. The watchdog sends **SIGQUIT first** — the Go runtime dumps all goroutine stacks into the shard log, so the next occurrence *names the wedged test* (the open question kit-k72d could not answer without that artifact) — then SIGKILL.
- **`scripts/test-local-parallel`**: same ownership at the fan-out level, with a drain window wide enough for nested shard-runner traps to finish before the outer kill (a narrower window would re-orphan exactly what this prevents).
- **Start-of-run sweep** covering both populations: orphaned `*.test` binaries (PPID 1, older than the go budget) and `gct-*` socket-parked tmux servers. Narrow by construction — only sockets under a `gct-*` path are in scope, so a developer's own tmux server is structurally unreachable; live sibling runs are spared via the existing PID-in-dirname ownership convention; `GC_TEST_NO_ORPHAN_SWEEP=1` opts out.

## Verification (measured, recorded on gas-px0)

- Both regression tests watched RED first, then green; reintroducing the bug flips `TestGoTestShardLeavesNothingHoldingTheCallersPipe` red at 60s.
- Group isolation measured directly (`runner_pgid=75773` vs `product_run_pgid=76357`).
- Sweep validated against the triage host's real process table: identified all 9 stranded tmux servers, skipped every live one; later ran during a full gate while 8 other agents' test binaries were live (several at 60–100% CPU) and logged **zero kills** — concurrent work is structurally out of scope, not lucky.
- One self-inflicted bug caught and fixed by the full-package run: killing the watchdog by PID orphaned its `sleep`, which held the caller's stdout pipe; killing the watchdog's *group* fixes it (pinned by the pipe test).
- `go test ./scripts` ok 144.6s; `go build ./...`; `go vet`; `gofmt`; `shellcheck -S warning`; pre-commit lint clean.

The branch was pushed with a declared `--no-verify`: the macOS pre-push suite red was proven pre-existing by checking out the base commit `014aee18b` in a throwaway worktree without this change and reproducing the identical firmlink-path failure (`/var/folders` vs `/private/var/folders`, the #5111 class). This change cannot reach those tests — `scripts/*_test.go` is `package scripts_test`, imported by nothing in `cmd/gc` or `internal/`.

## Deliberately out of scope

- The `test/tmuxtest` dir sweep still removes a run dir without killing that run's processes (which converts reachable orphans into unreachable ones — the accumulation mechanism). #5056 fixes the source fixture; the harness sweep here is the backstop. A kill-then-remove follow-up is tracked separately.
- Question (a) — *which* test wedges — is deliberately not guessed; the SIGQUIT escalation makes it self-answering on next occurrence.

Tracked in the gas city as gas-px0 (routed from kit-k72d host triage).

cc @sjarmak


AJBcoding:
**Updated: rebased onto current `main` and fixed the CI failure. 53 checks green, and the one remaining red is an infrastructure blip, not this diff.**

### What was failing

The single root failure was `TestRepositoryLedgerMatchesCensusAndDocumentation` (`internal/testpolicy/resourcecensus`); `CI / integration` and `CI / required` were just aggregating it. This PR adds a new untagged test file, `scripts/harness_reap_test.go`, so the resource-debt ratchet counted it and six baseline rows drifted.

### Attribution was measured, not inferred

Rather than read the deltas off the failure text, I dropped that one file from the scanner's own occurrence list and re-ran the real validator, which then passed — so the file accounts for the entire drift: exactly **4 subprocess + 2 `time.Sleep`** calls, `+1 file` on each row because the file is new.

| Row | Before | After |
| --- | --- | --- |
| audit `all`/subprocess | 607 / 173 | 611 / 174 |
| audit `all`/fixed_sleep | 436 / 161 | 438 / 162 |
| debt `untagged`/subprocess | 411 / 117 | 415 / 118 |
| debt `untagged`/fixed_sleep | 286 / 114 | 288 / 115 |
| small `untagged`/subprocess | 404 / 113 | 408 / 114 |
| small `untagged`/fixed_sleep | 286 / 114 | 288 / 115 |

Only `baseline_calls`/`baseline_files` moved; `reported_*` is left as historical evidence, and `TESTING.md` was regenerated with the package's own `-update` flag rather than hand-edited.

### Why re-baseline instead of a build tag

Not preference — a build tag only removes a file from the `untagged` scopes, while the two audit rows count all tracked test source regardless, so tagging cannot make those green. The two sleeps are also not reducible: both are poll intervals inside deadline-bounded loops keyed on a lifecycle signal (a pid file appearing, a process leaving the process table), which is what the `fixed_sleep` invariant asks for, and there is no shared polling helper in the repo to borrow. Swapping `time.Sleep` for a ticker would lower the count without lowering the real cost. Testing a harness that reaps process trees requires spawning process trees and waiting for them to die.

I also rebased onto current `main` first (the branch was 53 commits behind, and the ledger had moved three times in that window). CI builds the *merge* of head into base, so a re-baseline computed on the stale base would have been correct arithmetic about a tree nobody builds.

### The one remaining red needs a re-run, not a fix

`Integration / rest-smoke-1-of-2` failed in **environment setup, before compiling anything**:

```
Install dolt v2.1.7
curl: (22) The requested URL returned error: 503   <- 6 retries, all 503
##[error]Process completed with exit code 22
```

No test ran, so no diff could have caused it. I don't have rerun rights here — **could a maintainer re-run that job?** I expect it to go green and clear both aggregator checks.

Verified locally: `./internal/testpolicy/resourcecensus` green (was 6 drifting rows); `go test ./scripts` ok 157.99s with all six harness-reap regressions passing and none skipped; `go build ./...`, `go vet ./...`, `gofmt` all clean.

cc @sjarmak

quad341 (approved):
@AJBcoding — measuring the actual failure (5h34m at ~99% CPU, PPID 1, past a 20m `-test.timeout`) instead of theorizing about it, then writing regression tests you watched go red *first*, is how this class of bug gets fixed once rather than four times. The SIGQUIT-before-SIGKILL escalation is my favorite detail: you didn't guess which test wedges, you made the next occurrence answer that itself. And special credit for catching your own bug — the watchdog's orphaned `sleep` holding the caller's pipe — then pinning it with `TestGoTestShardLeavesNothingHoldingTheCallersPipe` rather than quietly fixing it.

I checked the red check for you: `rest-smoke-1-of-2` died with six consecutive 503s downloading dolt, ten seconds in, before anything compiled. Nothing in this diff can reach that. I also ran `go test ./internal/testpolicy/resourcecensus/...` against your head — green, so your re-baseline arithmetic is right (CI skipped that job on path filters). Re-running the flaked job and merging on our side.

Two follow-ups I'll file rather than send back to you: Linux `ps -o comm=` truncates at 15 chars so the sweep under-covers longer test binaries, and I want to time the watchdog grace against a `-coverpkg` build. Neither blocks. Thank you for this one.


---

quad341:
Re-reviewed `755352a39`..`7722f0270` (6 new commits since the last review): no new concerns — the prior review stands.

PR #5178, opened 2026-08-10
fix(cli): land the stranded-write repair verb and make the alarm name it (ga-j72r8)

julianknutsen:
## The failure this closes

A converged split city that acquires infrastructure beads in its retained
work store — by `cook --attach`, a pre-fix binary, a partial migration, a
rolled-back deploy — had **no way out**. `gc storage status` exited 1, every
later command carried a permanent alarm, and the alarm's remedy was a
*sentence*: "recover them into the binding's database". It named no verb.
`gc storage migrate --from-work` refuses to re-run, so there was nothing to
run. Strands are boot-fatal on the enterprise path, so this was not cosmetic.

**Proven in production before it was landed.** This verb (as a prototype)
recovered **1,452 stranded `orders`-class beads** on a live maintainer city:
work store unmutated (2,135 rows before and after), binding 51,127 → 52,579,
manifest a verified superset with 0 ids dropped, re-run a proven no-op, and
`gc storage status` went exit 1 → **exit 0, stranded 0** on the binary the
supervisor actually runs.

## Why a separate verb, and not the migration again

The migration's one-shot gate is a **safety property, not an obstacle**, and
`cmd/gc/infra_class_recover.go`'s header says so at length so the next person
is not tempted:

- `runInfraClassMigration` reads convergence first and hands a marked city to
  `confirmInfraConvergence`. It never re-copies.
- Force a re-copy by removing the marker and `prepareInfraDestination` DELETES
  every stamped row and re-imports from a source that no longer holds them.
  On the real city that is a measured **51,127 → 1,452 loss**.
- `verifyInfraCopy` demands bidirectional equality, which a serving binding
  that has legitimately grown can never satisfy.

So the repair is **scoped and additive**: it copies only ids the binding does
not hold and the manifest does not record, it never deletes from either store,
and it extends the manifest only after every moved bead is re-read field-,
class- and dep-equal from a CLOSED AND REOPENED destination.

## What landed

`gc storage recover-stranded --from-work --fleet-stopped [--dry-run] [--dump]`,
built from `storageRecoveryCommand` and parsed into the cobra tree by the same
`parseOperatorCommandSpelling` the migrate verb uses — so a refusal can never
name a command the binary does not carry. It reuses the migration's own
primitives throughout (`readInfraSnapshot`, `infraMigrationRow`,
`CreateWithForeignID`, `beadCopyDifference`, `infraCopyClassDifference`,
`infraDepDifference`, the copy manifest, `AcquireMigrationGuard`, the
foreign-controller probe).

**Refuses rather than guesses.** A bead whose owning class the binding does
not serve, whose class would change in the crossing, or whose dependency
topology the source cannot state is NAMED and left intact in the work store;
the command exits non-zero. Classification is `coordclass` — the same
authority `readInfraSnapshot` selects the gap with — pinned by
`TestStrandedRecoveryClassifiesWithTheSameAuthorityAsTheBootGuard`, because a
repair that disagreed with the guard would move beads and leave the boot
refusing anyway.

**The DepList carry-over, which is load-bearing.** A bd/Postgres work store
answers `DepList` with `operation "IssueRelations" not supported by the
postgres backend`. `sourceDepReader` prefers the relation read and falls back
to bd's inline per-bead projection (`beads.Bead.Dependencies`) — but only
after **witnessing that projection live** somewhere in the store, because an
adapter that never populated it would answer "no edges" for every bead and a
naive fallback would silently flatten the graph. Unwitnessed makes every bead
ambiguous rather than edge-free. The MIGRATION path needs the same treatment
and cannot run against such a source at all today: filed as `ga-7i7ts`.

**The alarm names the verb now**, at all three sites that report a strand:
`infraMigrationOperatorAdvice` (the boot advice a supervisor records),
`confirmInfraConvergence`'s stderr (what a boot prints), and `gc storage
status`, which used to print the ids and stop — that silence was half the
reported bug. The born-split arm deliberately does NOT name it (that binding's
provider never resolves a target, so the verb would refuse) and says this
build carries no repair for it instead; pinned by
`TestBornSplitAdviceDoesNotNameARepairItCannotRun`. And if an operator reaches
the verb anyway — `gc storage --help` lists it — it now reports the born-split
discipline's own verdict instead of asserting "nothing here is stranded"; see
the council round below.

## Red-before, by mutation

Manifest superset, with `extended` built from the moved beads alone:

```
recovery exited 1; stderr: gc storage recover-stranded: the extended manifest
would drop 4 id(s) the previous one recorded (gc-1, gc-2, gc-3, gc-4). It is
written as a superset or not at all
```

and with that guard ALSO removed — the production consequence, a city refusing
to boot over rows its own GC legitimately collected:

```
residual stranded: 3 / residual ids: gc-1, gc-2, gc-3
```

Ambiguous-class refusal, with the class gate dropped:

```
recovery exited 0 and copied gc-4 into a binding that does not serve its class
```

DepList fallback removed (the refusal propagating):

```
gc-4 (dependency topology unreadable: operation "IssueRelations" not supported by the postgres backend)
gc-5 (dependency topology unreadable: operation "IssueRelations" not supported by the postgres backend)
```

Witness requirement removed:

```
recovery exited 0 over a source that can state no bead's topology; the gap
crossed edge-free
```

Copy loop skipped (the equality stage catches it before the manifest moves):

```
bead gc-4 missing from the reopened binding: bead not found. The manifest was
NOT extended, so this bead is still named as stranded
```

Source deleted after copy:

```
the work store changed during the repair: before [gc-1 gc-2 gc-3 gc-4 gc-5 gc-6],
after [gc-1 gc-2 gc-3 gc-6]
```

Advice/status restored to their pre-fix wording:

```
the stranded advice names no runnable recovery: "... Stop every writer, recover
them into the binding's database, and re-check with `gc storage status` ..."
status names no runnable recovery, which is half the reported bug
```

Single-store city, with the unresolved-target refusal removed, then with the
convergence gates removed too:

```
the refusal does not say why there is nothing to do: "... has not converged
onto binding \"\" ..."
the repair opened the work store of a city with no infrastructure binding
```

The suite's source leaf is `internal/beads/splittest`'s strict BdSemantics
work store; the destination is the production opener throughout, i.e. a real
`beads.SQLiteStore` under the graph class's reserved prefix.

## Census

A new leaf needs the productmetrics command census: id 201 registered in
`cmd/gc/productmetrics_command_census.json`, `gen-command-census` re-run over
the three derived artifacts, and the lockstepped catalog count in
`internal/productmetrics/event_test.go` bumped 196 → 197.

## Council review round — one blocker and eight majors, fixed

A multi-reviewer council ran reproduced probes against `3a3a2b5c0e` and found
one blocker and eight majors. Several were facets of **two root causes**, and
both are fixed at the root rather than at each symptom. None of them fired on
the production recovery — all 1,452 beads there had **zero dep edges** and the
run completed — so this round is about the NEXT city, on a tool that writes to
a ledger during an incident.

### Root cause 1 — the edge pass iterated the copy, not the topology

**BLOCKER: inbound within-infra edges into a recovered bead were never
restored**, while the run exited 0 reporting `verified: N bead(s) re-read
field-, class- and dep-equal` and `residual stranded: 0`. The pass read
`DepList(id,"down")` for beads in `movable` only, so an edge from a bead that
crossed at cutover INTO one this run recovered was never read, never written,
never counted and never checked. That direction is the dispatcher's *normal*
wiring — `internal/dispatch/fanout.go:181`, `internal/molecule/molecule.go:351`,
`internal/dispatch/ralph.go:475` all write old → new — and because
`infraMigrationRow` strips `IsBlocked`, binding readiness is dependency-derived:
a dropped `blocks` edge makes a blocked bead read READY.

Same root, two more majors: an **outbound edge to a peer the repair declined to
move** was dropped permanently and mislabelled `cross-boundary edge(s) left as
metadata linkage` (it is neither cross-boundary nor retained — `infraMigrationRow`
nils both `Dependencies` and `Needs`), and **a re-run never re-drove edges** for
beads the binding already held, so the advertised refuse-then-fix workflow could
not repair a truncated topology.

**Fix.** The edge pass now runs over every source infrastructure bead the binding
will hold — `planInfraEdges` over the *resident* set, this run's rows and the
cutover's alike — which covers both directions, because an inbound edge into a
recovered bead is the outbound edge of a resident one. This is exactly what
`importInfraSnapshot` already does (its dep pass iterates every row, not the ones
it created), so no new mechanism is invented; the repair had regressed a
guarantee the cutover already had.

Only edges the binding **lacks** are written, so the pass is a diff, not a
rewrite: `edges: 0 restored, 4 already in the binding` is what a converged city
reports, and a re-run repairs a partial topology instead of no-opping past it.
An edge that cannot be written is now **named per edge**, and the three reasons
are three different facts rather than one count:

```
edges: 1 restored, 0 already in the binding, 1 cross-boundary edge(s) into work
left as metadata linkage, 0 into a bead the binding's own GC collected
DROPPED 1 within-infra dep edge(s): the far endpoint is an infrastructure bead
this repair declined to move, and nothing retains them:
  gc-5 -> gc-4 (blocks): the far endpoint is an infrastructure bead this repair declined to move
```

A dropped edge now also forces a non-zero exit. **No dropped edge rides under a
success report.**

### Root cause 2 — the partial-run residue was invisible and became boot-fatal

The gap loop dropped `{source-infra ∧ in binding ∧ not in manifest}` on the
floor. That set is exactly what an interrupted run leaves: rows are written one
by one, the manifest is published once at the end. Those ids are **not** still
stranded — the binding holds them — so no check named them, run 2 printed
`nothing to recover` and **exited 0**, and so did `gc storage status`. When the
binding's own wisp GC later hard-deleted one (the documented reason the manifest
exists), the boot guard re-classified it **stranded**: a city refusing to boot
over a row the repair had correctly delivered, whose only named remedy
re-imported a row the binding deliberately GC-deleted.

The header asserted the opposite — *"those ids are still stranded, still named by
the next check"* — which is **false for exactly the ids that got copied**. That
sentence is corrected, and the asymmetry with the migration is now stated
explicitly so it is not re-introduced: an interrupted `runInfraClassMigration`
self-heals (manifest before marker, then `prepareInfraDestination` drops and
re-imports); this repair writes into an already-converged binding it must never
clear, so **its resume is recognition, not rollback**.

**Fix.** A third bucket beside `stranded` and `removedSinceCutover`, reported in
the header block as `in the binding, unrecorded: N`, re-proven against the closed
and reopened destination by the same comparators the moved rows face, and folded
into the manifest. A residue row that cannot be proven to be the source's row is
**named and left unrecorded** — the manifest means "the copy was proven to
deliver this id", and recording an unproven one would be a proof nobody made.
`nothing to recover` / exit 0 can no longer be printed while the bucket is
non-empty.

**Both council probes reproduced and defeated**, as permanent tests:

- `TestStorageRecoverStrandedRecordsAResidueAnInterruptedRunLeft` — the
  `infraMigrationRename` stub (models ENOSPC/EROFS or a kill between the copy
  and the atomic publish).
- `TestStorageRecoverStrandedRecordsAResidueLeftByAnUnwritableBindingRoot` —
  **no seam stubbed at all**: `chmod 0555` on the binding root, which fails at
  `backupInfraCopyManifest` *after* `copied: 1`. The database lives in a
  component directory below the root, which is why the copy succeeds and only
  the manifest write fails.

Both then close the chain: the binding's own delete of a recovered row now reads
back as collected rather than as a strand.

### The remaining majors

**One transient `DepList` error downgraded the whole run** to the inline
projection — the read this codebase itself marks incomplete
(`BdStore.listIncludesCompleteDependencies` returns false) — and the equality
stage then compared against that same lossy read, so the dep proof could not
fail. Now `infraRelationCapabilityRefusal` latches the fallback **only** on the
adapter's unsupported-operation shape (the same text-match idiom as
`infraGenesisOpenFailure`'s "already exists"); any other error is per-bead, and a
bead whose own read fails is ambiguous rather than edge-free. And the equality
stage re-reads the source independently, as `verifyInfraCopy` does, so it can
now catch what a degraded read would have lost.

**The manifest was read before the migration guard** and rewritten under it,
with a controller-socket dial (500ms/500ms/2s) in between — a read-modify-write
straddling the lock, so a concurrent run that published and released inside that
window had its ids silently dropped, and `manifestIDsDropped` could not fire
because it diffed the replacement against the same stale map. The read moved
under the guard, matching `runInfraClassMigration`.

**`--dump` used `os.Create`.** Now `O_CREATE|O_EXCL` plus a refusal for any path
inside the binding root: the three tab-completions available there are the
manifest, the marker and the database directory, and `--dry-run --dump` needs no
attestation and prints "nothing was written".

**Born-split contradiction.** The `!ok` arm asserted "nothing here is stranded"
on the one city where that is false — the boot guard names the ids and
`gc storage status` prints `born-split: BLOCKED` for the same bead. The arm now
branches on `storageSplitShapeOf` exactly as `doStorageStatus` does, names the
provider, defers to `checkBornSplitDiscipline` and reports what that authority
says. Two commands in one binary no longer contradict each other about the same
bead id during an incident.

**Minors.** The `logPrefix` comment claimed a test pinned it against
`storageRecoveryCommand`; no test could see a local constant. It is now derived
from the spelling by `storageRecoveryLogPrefix()` and pinned end to end,
including through a real refusal's stderr. The remaining three minors were
facets of the two root causes above.

### Properties deliberately preserved

Copy-and-leave-source, refuse-on-ambiguous-class, `coordclass` agreement with the
boot guard, manifest-superset-or-nothing, and the do-not-re-run-the-migration
reasoning in the header — all unchanged, all still pinned by their original
tests. Field equality is still demanded for the rows this run creates and is
deliberately **not** demanded for rows already resident: a binding that has been
serving legitimately diverges from the source copy it was made from. Edge
equality is one-directional for those same resident rows, for the same reason,
and stays bidirectional for rows this run created, where anything extra would be
fabricated.

### Red-before, this round

```
the inbound edge gc-1 -> gc-4 did not cross: the binding holds no edge from the
bead that crossed at cutover into the one this run recovered. binding deps of
gc-1 = []; the source still holds it

the re-run did not restore the truncated edge gc-5 -> gc-4: binding deps of
gc-5 = []; a re-run that cannot repair a partial topology is not the resume the
refusal promises

the report calls a within-infra edge cross-boundary: "copied: 1 bead(s), 0 dep
edge(s) (1 cross-boundary edge(s) left as metadata linkage)"
the report does not name the endpoints of the edge it dropped

the second run reported a clean city over a row the binding holds and the
manifest does not record: ... nothing to recover ...
the manifest still does not record gc-4, so the binding's own GC of that row
would read back as a strand: [gc-1 gc-2]

the run downgraded the whole dependency read over a transient error: "source dep
edges: the store refuses relation listing (database is locked) ..."
the real edge gc-5 -> gc-4 was dropped by the downgrade: binding deps of gc-5 = []

the extended manifest dropped gc-concurrently-recovered-1, which the on-disk
manifest recorded before this run took the guard; manifest = [gc-1 gc-2 gc-4]

the dry run truncated the proven-copy manifest it was pointed at: "[\n  {\n
\"bead\": {\n      \"id\": \"gc-4\", ..."

the refusal claims nothing is stranded on a city whose boot guard names gc-1:
"... so there is no binding to recover into and nothing here is stranded."
```

Thirteen new tests, twelve of them red before the fix and one
(`TestStorageRecoverStrandedLeavesCrossBoundaryEdgesAlone`) green before and
after — it is the guard that keeps the fix from eating the edge class that is
*supposed* to be left behind.

## Gates

Re-run in full after the council round, SERIAL (`-p 1 -parallel 1`) on a loaded box:

`go build ./...` OK · `go vet ./...` OK · `gofumpt -l cmd internal` empty ·
`golangci-lint run ./...` (2.10.1) **0 issues** ·
`go test ./cmd/gc -timeout 25m` (SERIAL) **ok 771.3s** ·
`go test ./internal/...` **ok** · `go test ./internal/testpolicy/...` **ok** ·
`scripts/check-core-boundary.sh` OK · `make check-split-topology-rows` OK ·
`make check-routed-test-rows` OK · `gen-command-census -check` OK.

No new leaf and no wire change in the council round, so the command census,
OpenAPI and dashboard artifacts are unchanged from the first pass (census id 201
and the 196 → 197 catalog bump landed with the original commit).


julianknutsen:
## Council round applied — 1 blocker + 8 majors + 4 minors fixed in `5bd763ea51`

Two root causes, fixed at the root rather than symptom by symptom. **None of
these fired on the production recovery** — all 1,452 beads there had zero dep
edges — so this round is about the next city, on a tool that writes to a ledger
during an incident.

### 1. BLOCKER — inbound within-infra edges were silently lost

The edge pass iterated `movable` and read outbound edges only, so an edge FROM a
bead that crossed at cutover INTO one this run recovered was never read, written,
counted or checked — under `verified: N bead(s) ... dep-equal`, `residual
stranded: 0` and exit 0. It is the dispatcher's *normal* wiring direction
(`fanout.go:181`, `molecule.go:351`, `ralph.go:475` all write old → new), and
since `infraMigrationRow` strips `IsBlocked`, a dropped `blocks` edge makes a
blocked bead read READY.

Same root: outbound edges to peers the repair declined to move were dropped and
mislabelled `cross-boundary … left as metadata linkage`, and a re-run never
re-drove edges for beads the binding already held.

**Fix.** `planInfraEdges` runs over every source infra bead the binding will hold
— inbound and outbound cover each other, because an inbound edge into a recovered
bead *is* the outbound edge of a resident one. This is precisely what
`importInfraSnapshot` already does (its dep pass iterates every row, not the ones
it created), so no new mechanism was invented; the repair had regressed a
guarantee the cutover already had. Only *missing* edges are written, so the pass
is a diff — a re-run repairs a truncated topology and reports `edges: 1 restored`,
then `0 restored` once converged. Unwritable edges are **named per edge**, with
three distinct reasons, and force a non-zero exit.

### 2. The partial-run residue was invisible and became boot-fatal

`{source-infra ∧ in binding ∧ not in manifest}` was dropped on the floor. Run 2
printed `nothing to recover` and exited 0; when the binding's own GC later
deleted the row, the boot guard called it stranded. The header's claim that those
ids "are still stranded, still named by the next check" is **false for exactly
the ids that got copied**, and is corrected — along with the asymmetry: the
migration self-heals by rollback, this repair resumes by *recognition*, because
it writes into an already-converged binding it must never clear.

**Fix.** Third bucket, reported as `in the binding, unrecorded: N`, re-proven by
the same comparators the moved rows face, folded into the manifest. A row that
cannot be proven to be the source's row is named and left unrecorded.
`nothing to recover`/exit 0 is unreachable while the bucket is non-empty.

**Both council probes reproduced and defeated** (now permanent tests):

<details><summary>Probe A — <code>infraMigrationRename</code> stub</summary>

```
RUN 1 exit=1   copied: 1 bead(s) / verified: 1 bead(s) ...
               writing the infra copy manifest: simulated rename failure
RUN 1 manifest=[gc-1 gc-2]
RUN 2 exit=0   stranded: 0 / in the binding, unrecorded: 1
               verified: 1 bead(s) ... / manifest: 2 -> 3 bead(s)
RUN 2 manifest=[gc-1 gc-2 gc-4]  stranded=[]
```
</details>

<details><summary>Probe B — <b>no seam stubbed</b>, <code>chmod 0555</code> on the binding root</summary>

```
RUN 1 exit=1   copied: 1 bead(s) / verified: 1 bead(s) ...
               backing the manifest up to ...bak-...: permission denied
RUN 1 manifest=[gc-1 gc-2]
RUN 2 exit=0   stranded: 0 / in the binding, unrecorded: 1
               manifest: 2 -> 3 bead(s)
RUN 2 manifest=[gc-1 gc-2 gc-4]  stranded=[]
```
</details>

### 3. The remaining majors

- **Transient `DepList` error downgraded the whole run** to the inline projection
  (the read `BdStore.listIncludesCompleteDependencies` declares incomplete), and
  the equality stage compared against that same lossy read. Now
  `infraRelationCapabilityRefusal` latches only on the adapter's
  unsupported-operation shape — the `infraGenesisOpenFailure` idiom — and any
  other error is per-bead; the equality stage re-reads the source independently,
  as `verifyInfraCopy` does, so it can catch what a downgrade would have lost.
- **Manifest read moved under the migration guard**, matching
  `runInfraClassMigration`; the read-modify-write no longer straddles the lock.
- **`--dump`**: `O_CREATE|O_EXCL` plus a refusal for any path inside the binding
  root, where all three tab-completions are fatal targets.
- **Born-split**: the `!ok` arm no longer asserts "nothing here is stranded" on
  the one city where that is false. It branches on `storageSplitShapeOf` exactly
  as `doStorageStatus` does and defers to `checkBornSplitDiscipline`.
- **Minors**: `logPrefix` is derived from `storageRecoveryCommand` and pinned end
  to end (the old comment claimed a test pinned a local constant nothing could
  see); the other three were facets of the two root causes.

### Preserved

Copy-and-leave-source, refuse-on-ambiguous-class, `coordclass` agreement with the
boot guard, manifest-superset-or-nothing, and the do-not-re-run-the-migration
reasoning — unchanged, still pinned by their original tests. Field equality is
still demanded for rows this run creates and deliberately **not** for rows already
resident, which legitimately diverge from the source copy they were made from;
edge equality is one-directional for those and stays bidirectional for rows this
run created.

### Tests and gates

13 new tests, **12 red before the fix**; the 13th
(`…LeavesCrossBoundaryEdgesAlone`) is green before and after and guards against
the fix eating the edge class that is *supposed* to be left behind.

SERIAL (`-p 1 -parallel 1`): `go build ./...` · `go vet ./...` · `gofumpt` clean ·
`golangci-lint` 2.10.1 **0 issues** · `go test ./cmd/gc` **ok 771.3s** ·
`go test ./internal/...` **ok** · `./internal/testpolicy/...` **ok** ·
`check-core-boundary` · `check-split-topology-rows` · `check-routed-test-rows` ·
`gen-command-census -check` — all OK.

PR #2583, opened 2026-05-24
fix: split queued session starts from creating

julianknutsen:
Summary:
- Introduce `start-pending` for queued session starts before provider Start is in flight.
- Move provider wake metadata to the pre-wake boundary and keep capacity/lifecycle projections aware of both queued and creating states.
- Updates session, worker, reconciler, and CLI expectations for the split.

Tests:
- go test ./internal/session ./internal/worker
- go test ./cmd/gc -run 'Test(CmdSessionNew_AllowsReservedNamedAliasWithController|SessionLifecycleChaos|ExecutePreparedStartWaveUsesWorkerBoundaryForKnownSession|HealState_|HealStatePatchProjectsRuntimeLiveness|ReconcileSessionBeads_PreservesSessionKeyWhenNamedRestartDeferred|ReconcileSessionBeads_SyncReplacesFailedCreateNamedSession)$'\n- go test ./cmd/gc -run 'TestSyncSessionBeads_(CreatesNonActiveBeadWithPendingCreateStartedAt|StoppedAgent|ResumedAfterSuspension)$'\n- pre-commit fast suite passed\n\nNote:\n- Full `go test ./cmd/gc` was used during iteration to find lifecycle expectation drift; final verification is the targeted coverage above plus the repo fast shard hook.

PR #5179, opened 2026-08-11
fix(session): roll back wedged start-pending sessions, not just creating (gcf-ru0)

kaushikNaarayan:
## Summary
- `healStatePatchWithRollbackInfo`'s expired-pending-create rollback gate only fired for `MetadataState == "creating"` — a leftover from before #2583 split `start-pending` out as its own precursor state.
- `projectRuntimeProjection`'s `BaseStateStartPending` branch has no staleness check of its own (unlike its `BaseStateCreating` sibling), so once a bead migrated to `start-pending` it kept re-projecting `start-pending` forever with no self-heal: newly-created sessions (morning resume, cold `gc session nudge` spawns) that failed to reach a provider `Start` call wedged indefinitely, silently consuming a pool slot.
- `pendingCreateLeaseExpiredForRollbackInfo` already understands `start-pending` via `pendingCreateRollbackState`, so this widens the gate to `pendingCreateQueuedOrCreatingState` instead of a bare `"creating"` string check.

Fixes gcf-ru0.

## Test plan
- [x] Added `TestHealState_StartPendingNeverStartedRollsBackToAsleep` (red before the fix, green after).
- [x] `go test ./cmd/gc/ -run 'TestHealState|TestReconcileSessionBeads|TestPendingCreate|TestStaleCreating'` — 269 pass, 0 fail.
- [x] `go test ./internal/session/...` — pass.
- [x] `go build ./...` and `go vet ./cmd/gc/... ./internal/session/...` clean.

### DIFF
--- a/cmd/gc/session_reconcile.go
+++ b/cmd/gc/session_reconcile.go
@@ -920,7 +920,15 @@ func healStatePatchWithRollbackInfo(info sessionpkg.Info, alive bool, clk clock.
 		target = string(sessionpkg.StateAsleep)
 		clearPendingCreateLeaseInfo(info.PendingCreateClaim, batch)
 	}
-	if rollbackAvailable && !alive && strings.TrimSpace(info.MetadataState) == "creating" {
+	// gcf-ru0: this gate originally checked MetadataState == "creating" only,
+	// from before #2583 split "start-pending" out as its own state ahead of
+	// "creating". projectRuntimeProjection's BaseStateStartPending branch has
+	// no staleness check of its own (unlike its BaseStateCreating sibling), so
+	// a bead stuck in start-pending with no rollback gate here just keeps
+	// re-projecting start-pending forever — pendingCreateLeaseExpiredForRollbackInfo
+	// already understands start-pending via pendingCreateRollbackState, so widen
+	// the gate to match instead of restricting it to "creating".
+	if rollbackAvailable && !alive && pendingCreateQueuedOrCreatingState(info.MetadataState) {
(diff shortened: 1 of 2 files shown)


quad341 (approved):
@kaushikNaarayan — merging this. The archaeology is what makes it reviewable: tying the bare `== "creating"` check back to #2583 splitting `start-pending` out as its own precursor state, and then pointing at `pendingCreateLeaseExpiredForRollbackInfo` already understanding `start-pending` via `pendingCreateRollbackState`, is exactly the right argument for widening the gate instead of bolting on a second check. The test is the good kind, too — it pins both halves (state → `asleep` *and* the claim cleared) and sits right next to the fresh-lease sibling, so the boundary is covered from both sides. I reverted your one-liner locally and watched it go red, then green.

One thing I want to flag rather than hold you up over: I couldn't convince myself this branch is what un-wedges a real session. Both rollback sites that run ahead of heal — `session_reconciler.go:1681` and `:2229` — fire on a weaker condition (claim + expired lease, no state filter) and always `continue`, and every arm of `:1681` including the fail-closed liveness-error one does the same. So the heal gate may be preempted in practice, and the field wedge may be a `start-pending` bead that already *lost* its claim, which `pendingCreateLeaseExpiredForRollbackInfo` bails on immediately.

Your change is still right and strictly safer — it makes the heal helper self-consistent with the predicate it delegates to. I'd just keep gcf-ru0 open until we see a wedged session self-heal in the field. If you have a `gc trace` from one, drop it on the bead; you're the fastest person to close that loop.


---

PR #5180, opened 2026-08-11
fix(beads,doctor): say something when a work-store read answers empty from a scope that cannot see its own ledger (ga-clsfl)

julianknutsen:
The read-side half of the metadata-rewrite fail-open (`ga-clsfl`, split out of #5167). `gc rig add`, `gc start`, `gc supervisor run`, `gc rig set-endpoint` and `gc beads city use-managed|use-external` rewrite a scope's `.beads/metadata.json` from `dolt_mode=embedded` to `server`, re-pointing the ledger at a database that has never held the operator's beads and leaving the populated one on disk, unread. bd does not fail — it opens the database its metadata names, runs the query and matches nothing — so every work-store read answers `[]` with exit 0. #5167 made the **flip** loud; the read that pays for it, in another process, still got nothing at all.

**Not split-store specific.** A city with exactly one store is bitten exactly as hard: no coordination class is involved, only a workspace pointed at the wrong database. Nothing in this change consults `[storage.classes]`, and `TestUnreadStoreGuardIsNotSplitStoreSpecific` runs the identical assertion on a `BdStore` with no relocated classes and one with a relocated graph class.

## This is a notice, not a refusal — and that is the point

The first version of this guard (`eb287a4a64`, removed in `63f65d583c`) was a refusal, and a fable council proved it a fleet-wide outage. All three findings were reproduced against a restored copy of that guard **in this branch**:

| | finding | status |
|---|---|---|
| **EXP1** | judged per **call**, not per **store** — refused an empty `Ready()` on a store whose `List` had just returned a row. `federateBeadLegs` aborts the federation on any leg error → `gc ready` exits non-zero → every agent's `gc hook` fails → the API hits `totalOutage()`. A healthy city fails closed precisely when **idle**. | **defeated** |
| **EXP2** | sat **after** client-side filtering — `Ready{Assignee}`, `Ready{TierWisps}`, `Children(leaf)` and an empty mail poll all refused on rows bd had just handed back. | **defeated** |
| **EXP3** | plain `bd init` is embedded **by default** and creates an empty `.beads/embeddeddolt/<db>/.dolt`; gc canonicalizes that to server on `gc rig add`, so the condition is true forever on a city that was never broken — including inside `verifyCanonicalBdScopeStoreReady`, which gates adoption on `List(AllowScan, Limit 1)` returning nil, 20× at 500ms. | **not fixable at read time — shipped as the honest subset** |

EXP3 is stated rather than papered over. "Empty active store beside a second Dolt repository" is the same on-disk shape as a fresh adoption, and separating them means **opening** the unread database — a store open, a Dolt file lock and a possible schema migration on the exact directory `gc doctor`'s `splitStoreFixHint` tells the operator to preserve. A read path may not do that, and a diagnostic may not mutate the backup it told you to keep. Timestamp heuristics ("written days ago vs. seconds ago") were rejected as a threshold, i.e. judgment in Go.

So the outcome is a stderr **notice** that leaves every answer byte-identical. It removes the *confidence* — which is the harm that was actually reported — without acquiring the power to take the fleet down. A correct narrow guard beats a broad one that can 503 the API.

## Review fix: the notice may not spend the caller's read deadline (2026-08-11)

Review found one **MAJOR** in the shape above, in two halves. Both were reproduced against this package's real code, and both are now regression-pinned.

### (a) The diagnostic probe spent the caller's budget

`noticeIfStoreCannotSeeItsLedger` confirmed the active store was empty by running `holdsAnyRow`'s `bd list --limit 1` **synchronously** inside `BdStore.List`/`Ready`, through `runBDTransientRead` (up to 3 attempts with backoff). `Store.List` and `Store.Ready` take no context, so the store cannot see the budget its caller is holding — and that probe was charged to it. Measured against `internal/api`, with `statusStoreReadTimeout` shortened to 250ms and the runner answering every read `[]` but taking 400ms on the probe argv:

```
statusListStoreWithTimeout(ctx, state, store, ListQuery{AllowScan:true})
  => rows=0 err="list timed out: context deadline exceeded"   (250ms)
control, GC_BD_ALLOW_UNREAD_STORE_READ=1
  => rows=0 err=<nil>                                         (72µs)
```

**bd had already answered that List successfully.** The whole budget went to the diagnostic, so the notice could only ever turn a successful read into an error — the opposite of the "byte-identical" invariant this PR is built on, which the old test asserted only in an *un-deadlined* context. Binding the probe child to the caller's context does not fix it either: the child dies, the budget is still gone.

### (b) The one-shot latch did not hold where the API builds a store per request

`cmd/gc/scoped_store.go`'s `scopedBdStoreForCity`/`scopedBdStoreForRig` construct a **brand new** `beads.NewBdStore(...)` on every call, and `internal/api/handler_status.go` reaches them through `state.ScopedStoreLike(reqCtx, store)` on every `/status`. Measured over one scope:

```
10 requests => 10 probe subprocesses, 13,060 bytes of notice
```

`statusWorkCounts` fans out concurrently over city plus every rig, so one request could spawn **R+1** extra `bd` children against the same Dolt server it was already reading. "One-shot verdict per store" degraded to **once per read** there, which made `splitStoreFixHint`'s *"a one-time notice"* and the runbook's *"prints a one-time notice"* false — at status-rebuild rate, in a tree with a documented log-flood history.

Reachability is not inferred: `ListQuery.HasFilter()` does not include `IncludeClosed`, so `{AllowScan:true, IncludeClosed:true}` satisfies `listReadIsWholeLedger`; `*BdStore` implements no `Count`, so `CachingStore.Count` always answers `ErrCountUnsupported` for the closed-inclusive query and `statusListStoreWithTimeout` is the mandatory fallback for store health.

### What changed — the probe is dropped, not rescheduled

**Chosen shape: the notice spends no bd subprocess at all.** A synchronous probe cannot be made deadline-safe from inside `Store.List`/`Ready`, because those take no context and therefore cannot see the caller's budget; async would mean the CLI process that actually needs the line usually exits before it prints; and "skip when the deadline is close" needs a threshold nobody can pick — judgment in Go. **A notice that can time out a read is worse than a slightly noisier notice**, so `holdsAnyRow` is gone and the notice now runs on evidence the read already paid for: the row latch, the on-disk shape (one metadata read and up to two stats), and nothing else. The bonus the reviewer noted — the probe's argv was the outer read's argv plus `--limit 1`, re-asking in a second subprocess the question bd had just answered — is gone with it.

The price is a wider notice: an idle ledger and a re-pointed workspace read the same from here, so a scope that still has two bead databases can see the line while healthy. The message says that out loud instead of claiming a probe it no longer runs, and the question the probe was asking belongs to `gc doctor`, which enumerates both databases with no caller waiting on it.

1. **Memoized per resolved scope path.** `guardForScope(filepath.Clean(dir))` over a package-level `sync.Map`, resolved once in `NewBdStoreWithPrefix`, so a throwaway request-scoped store inherits the verdict the process already reached. Still a compare-and-swap latch, not `sync.Once` — `TestUnreadStoreNoticeNeverBlocksAnotherRead` keeps that pinned, now blocking on the notice sink (the remaining thing the winner does that it does not control) rather than on a probe that no longer exists.
2. **Corrected wording.** `splitStoreFixHint` and `docs/runbooks/managed-city-endpoints.md` now state the bound that actually holds — **one notice per `gc` process** — and the runbook says plainly to expect it on healthy scopes that still have two databases. The notice itself no longer claims "an unfiltered probe of that store found no row at all"; it claims only "no read of this scope has returned a row in this process".

### Red before, on the review fix

`internal/api` (new):

```
statusListStoreWithTimeout = (0 rows, list timed out: context deadline exceeded) after 250.467468ms; want (0, nil) — bd answered this List successfully and a diagnostic may not spend the caller's deadline
the read path ran 1 bd command(s) the caller did not ask for; a store read may only invoke bd for its own query
10 requests over one scope printed the notice 10 times (13060 bytes), want exactly 1
```

`internal/beads`:

```
the guard spent 10 probe subprocess(es); a diagnostic inside a read spends the caller's deadline
Ready() ran [bd ready --json --limit 0 bd list --json --all --include-infra --include-gates --limit 1]; want exactly the caller's own read "bd ready --json --limit 0" and nothing else
the notice does not name "spends no bd subprocess of its own"
the notice claims "found no row" about the active store, which nothing here established
```

`cmd/gc`:

```
gc doctor promises a one-time notice; the guard bounds it per scope per process, and the API rebuilds its store per request
```

## The per-scope shape

1. **`sawRows`** — an atomic latch set from what bd **returned**, before `applyListQuery` and before `Ready`'s tier/assignee loop. Never unlatched. This is the correction EXP1 and EXP2 both needed: *this call returned nothing* and *this scope holds nothing* are different claims, and only the second is evidence.
2. **Only an unfiltered read is eligible** — `List` with `AllowScan` and no filter, `Ready` with no assignee (`bdReadyArgs` sends exactly one selector, so the pair is complete). A filtered empty result is evidence of nothing.
3. **One notice per SCOPE PATH per process**, behind a compare-and-swap latch — not a `sync.Once`, because `once.Do` parks every concurrent reader behind the winner and one slow sink must not stall a supervisor's per-rig goroutines. It costs a metadata read plus up to two stats, and **zero subprocesses**, ever.

**Hot-path cost** (`BenchmarkUnreadStoreGuard*`): a read that returned rows pays one atomic load; an empty whole-ledger read pays two once the verdict is in. Zero subprocesses on any city, split or not.

## Override

`GC_BD_ALLOW_UNREAD_STORE_READ=1` silences it — symmetrical to `GC_BD_ALLOW_RELOCATED_CLASS_READ`, and named in the notice itself. It lives at the **store** layer rather than the `gc bd` CLI seam because this fires from `BdStore.List`/`Ready` on every path that reaches them; a knob honored on only some of them would advertise an escape that does not work.

## Reconciled with `gc doctor`

`gc doctor`'s own advice — *"keep both directories until reconciled"* — parks an operator in exactly the shape this notice describes, possibly for days. `splitStoreFixHint` now names the notice, its real bound and the override, and `TestTheThreeMessagesAboutOneUnreadDatabaseAgree` runs the flip-time announcement, `gc doctor` and the read-time notice against **one scope** and pins that they carry the same remediation clause, name the same directory, steer at the same diagnostic, that none of them claims rows were lost, and that doctor does not promise less noise than the guard delivers.

## Red before (original)

Guard removed (origin/main behavior):

```
nothing was said about the unread database: ""
no notice on a single-store city (no relocated classes): ""
no read-time notice naming ".../.beads/embeddeddolt/jc"; notices=""
the read-time notice printed 0 times across the adoption gate and a Ready, want exactly 1
read-time notice no longer carries the shared remediation clause: ""
```

Council experiments, against a restored `eb287a4a64` guard:

```
EXP1 Ready() error = bead store read returned empty while an unread bead database
     sits beside it: bd ready returned no rows for <scope> from the "dolt" store…
     (List had just returned 1 row)
EXP2 assignee-scoped ready / wisp-tier ready / children of a leaf / empty mail poll /
     label lookup — same refusal, on rows bd answered with
EXP3 List(AllowScan, Limit 1) error = <same refusal> — `gc rig add` fails after a
     ten-second stall
cmd/gc: List = (0 beads, <same>) / Ready / Children
```

**Mutation proofs.** Judging per call instead of per scope fails EXP1; dropping the unfiltered-read placement fails all five filtered rows; removing the env check fails the override test; memoizing on the store object instead of the scope path fails the ten-request test; reintroducing the probe fails both the deadline test and the argv test; dropping the knob from doctor's hint fails the three-message reconciliation; a `sync.Once` verdict fails the non-blocking test after 10s.

## Gates

`go build ./...` · `go vet ./...` · `gofmt` clean · golangci-lint 2.10.1 **0 issues** · `go test ./cmd/gc -timeout 25m` serial · `go test ./internal/...` (**144 ok, exit 0**), including `./internal/api` · `scripts/check-core-boundary.sh` · `check-split-topology-rows` · `check-routed-test-rows`.

`GC_BD_ALLOW_UNREAD_STORE_READ` was added to `internal/testenv`'s `GC_*` vocabulary golden, deliberately, beside its sibling.

Note on `internal/beads/splittest`: its leaves are `beads.MemStore` behind strict semantics wrappers, with no `.beads/metadata.json` and no bd subprocess, so it structurally cannot express a `BdStore` metadata-rewrite scenario. The not-split-specific claim is proven directly instead, at the layer the guard lives on.


julianknutsen:
## Review fix pushed — `0a5160d7d6`

One confirmed MAJOR, both halves reproduced against this package's real code and both now regression-pinned.

### The shape I chose: **drop the probe**

`holdsAnyRow` is gone from the read path rather than moved off it. `Store.List`/`Store.Ready` take no context, so a `BdStore` cannot see the budget its caller set — any synchronous subprocess it adds is charged to that budget. Async would mean the short-lived `gc ready`/`gc hook` process that actually needs the line usually exits before it prints, and "skip when the deadline is close" needs a threshold nobody can pick, i.e. judgment in Go. **A notice that can time out a read is worse than a slightly noisier notice**, so the verdict now runs on evidence the read already paid for — the `sawRows` latch plus the on-disk shape (one metadata read, up to two stats) — and spends **zero** bd subprocesses.

The reviewer's bonus finding goes with it: the probe's argv was the outer read's argv plus `--limit 1`, re-asking in a second subprocess the question bd had just answered.

The price is a wider notice — an idle ledger and a re-pointed workspace look the same from here — so a healthy scope that still has two bead databases can see the line. That is stated in the message, in `splitStoreFixHint` and in the runbook rather than papered over. The "what does each store actually hold" question stays with `gc doctor`, which enumerates both databases with no caller waiting on it.

### The two measurements, defeated

**(a) deadline** — `internal/api`, `statusStoreReadTimeout` = 250ms, runner answers every read `[]` but takes 400ms on anything the caller did not ask for:

| | before | after |
|---|---|---|
| `statusListStoreWithTimeout(ctx, state, store, {AllowScan:true})` | `rows=0 err="list timed out: context deadline exceeded"` (250ms) | `rows=0 err=<nil>` |
| control, `GC_BD_ALLOW_UNREAD_STORE_READ=1` | `rows=0 err=<nil>` (72µs) | `rows=0 err=<nil>` |

Both arms now run in the same test, so a regression shows up as a *difference* rather than a bare timeout, and the test also asserts `0` bd commands the caller did not ask for.

**(b) one scope, N requests** — 10 reads through `state.ScopedStoreLike`, each building a fresh `beads.NewBdStore` exactly as `scopedBdStoreForCity` does:

| | before | after |
|---|---|---|
| probe subprocesses | 10 | **0** |
| notices printed | 10 (13,060 B) | **1** |
| bd commands total | 20 | 10 (one per request) |

Memoization is `guardForScope(filepath.Clean(dir))` over a package-level `sync.Map`, resolved once in `NewBdStoreWithPrefix`. Still a compare-and-swap latch, no `sync.Once` barrier — `TestUnreadStoreNoticeNeverBlocksAnotherRead` keeps that pinned, now blocking on the notice **sink** (the remaining thing the winner does that it does not control) since there is no probe left to block on.

### Corrected wording

- `splitStoreFixHint`: ~~"reads from an empty active store print a one-time notice while both exist"~~ → "while both exist, an empty whole-ledger read from this scope prints **one notice per gc process**".
- `docs/runbooks/managed-city-endpoints.md`: same bound, plus "Expect it on any scope that still has two bead databases, including healthy ones; that is the price of a notice that cannot cost a read."
- The notice itself: ~~"an unfiltered probe of that store found no row at all"~~ → "no read of this scope has returned a row in this process", and it now states "spends no bd subprocess of its own".

### Red before

```
statusListStoreWithTimeout = (0 rows, list timed out: context deadline exceeded) after 250.467468ms; want (0, nil) — bd answered this List successfully and a diagnostic may not spend the caller's deadline
the read path ran 1 bd command(s) the caller did not ask for; a store read may only invoke bd for its own query
10 requests over one scope printed the notice 10 times (13060 bytes), want exactly 1
the guard spent 10 probe subprocess(es); a diagnostic inside a read spends the caller's deadline
Ready() ran [bd ready --json --limit 0 bd list --json --all --include-infra --include-gates --limit 1]; want exactly the caller's own read "bd ready --json --limit 0" and nothing else
the notice does not name "spends no bd subprocess of its own"
the notice claims "found no row" about the active store, which nothing here established
gc doctor promises a one-time notice; the guard bounds it per scope per process, and the API rebuilds its store per request
```

### Kept

Notice-not-refusal, the `sawRows` pre-filter latch, the unfiltered-read placement, `GC_BD_ALLOW_UNREAD_STORE_READ`, and `TestTheThreeMessagesAboutOneUnreadDatabaseAgree` — which now also pins that doctor does not promise less noise than the guard delivers. `TestUnreadStoreProbeFailureSaysNothing` is replaced by `TestABrokenBdDoesNotSilenceOrChangeTheNotice`: with no probe there is nothing to fail, so a broken bd surfaces the caller's own error verbatim and does not suppress a later successful empty read.

### Gates

`go build ./...` · `go vet ./...` · `gofmt` clean · golangci-lint 2.10.1 **0 issues** · `go test ./cmd/gc -timeout 25m` serial (**761.8s, ok**) · `go test ./internal/...` (**144 ok, exit 0**, `internal/api` 81.5s ok) · `scripts/check-core-boundary.sh` · `check-split-topology-rows` · `check-routed-test-rows`.

julianknutsen:
### CI fix: `Integration / packages-core-4-of-4` (80d795d)

**The failure** — `TestRepositoryLedgerMatchesCensusAndDocumentation`:

```
Small resource census grew: scope=untagged resource=fixed_sleep calls=287 (baseline 286), files=115 (baseline 114)
source resource census grew: scope=all  resource=fixed_sleep calls=437 (baseline 436), files=162 (baseline 161)
```

`CI / integration` and `CI / required` were cascades of that one job.

**Which cause it was** — the lockstepped-ledger ratchet, but the *growth* half, not the shrink half. Dropping the probe changed no census total at all: the census counts only `*_test.go` files, and the probe lived in non-test source. What tripped it was the new deadline test simulating a slow probe with `time.Sleep(400 * time.Millisecond)` — a 163rd untagged fixed sleep.

**The baselines are not bumped.** The untagged `fixed_sleep` row's invariant is *"cannot grow; reductions must lower this baseline"* — a one-way ratchet down, so raising it would have been the violation, not the fix. That row's `resource_owner` names the remedy exactly: *"each owning test replaces elapsed wall time with its lifecycle signal."* So no change to `census.go`, `test-resources.toml`, or `TESTING.md`.

**What changed** — one test file. The lifecycle signal is the caller's own context: `statusListStoreWithTimeout` derives a 250ms child from it, so the parent is canceled only after the read has already given up. A bd command the caller did not ask for now blocks on that parent instead of on a wall clock, so it *cannot* complete inside the caller's budget by construction — rather than by hoping 400ms beats 250ms on a loaded box.

**The proof got stronger, not weaker.** Re-introducing the probe (mutation test, reverted) still fails both arms, now deterministically:

```
statusListStoreWithTimeout = (0 rows, list timed out: context deadline exceeded) after 250.197418ms; want (0, nil)
the read path ran 1 bd command(s) the caller did not ask for
10 requests ran 11 bd command(s), want 10
```

The green path no longer sleeps at all — 0.00s, down from ~0.8s.

**Nothing the council accepted moved:** notice-not-refusal, the `sawRows` latch, unfiltered-read placement, `GC_BD_ALLOW_UNREAD_STORE_READ`, the three-message reconciliation test, "one notice per gc process", and zero subprocesses on the read path. `TestUnreadStoreNoticeNeverBlocksAnotherRead` is untouched and still guards its own property (a stalled sink must not serialize reads).

**Gates:** `go build ./...`, `go vet ./...`, gofmt, golangci-lint 2.10.1 (0 issues), `go test ./cmd/gc -timeout 25m` serial (735s), `go test ./internal/...`, `./internal/api/...`, `./internal/testpolicy/...`, `scripts/check-core-boundary.sh`, `make check-split-topology-rows`, `make check-routed-test-rows` — all green.

PR #3218, opened 2026-06-07
fix(beads): mirror bd ready projection in cache

julianknutsen:
## Summary
- propagate bd 1.0.5 `is_blocked` ready projection into cached bead rows
- prefer projected readiness in `CachedReady`, preserving legacy dependency-derived behavior for older bd versions
- enrich BdStore cache prime/reconcile via version-gated `bd sql` across issues and wisps

## Root Cause
The reconciler pool-demand path used cached ready work. bd 1.0.5 moved ready semantics into the denormalized `is_blocked` projection, but Gas City cache did not carry that value, so active blocked rows could be counted as ready even when `bd ready` excluded them.

## Tests
- `go test ./internal/beads -count=1`
- `go test ./cmd/gc -run 'TestDefaultScaleCheckCounts|TestCollectAssignedWorkBeadsUsesCachedReadyReadModel|Test.*CachedReady|Test.*ScaleCheck' -count=1`\n- pre-commit hook: lint-changed, go vet, fast local shards, dashboard build/typecheck\n

PR #5183, opened 2026-08-11
fix(beads): gate the ready projection on the BACKEND, not the bd version

julianknutsen:
## The defect (live on maintainer-city)

`CachingStore` prime and reconcile enrich every active bead's `is_blocked`
through `BdStore.fetchReadyProjection`, which shells `bd sql`
(`internal/beads/bdstore_ready_projection.go:139`). The capability gate
`bdReadyProjectionEnabled()` only ever asked bd its **version**
(`bdReadyProjectionMinVersion = "1.0.5"`).

maintainer-city is the one city whose work class lives on hosted Postgres. bd
serves Postgres embedded, where `bd sql` is not implemented:

```
Error: 'bd sql' is not yet supported in embedded mode      (exit 1, after 6-16s)
```

bd is 1.1.0, so the version gate passed, the failure was never latched, and it
retried on every cycle forever. The enrichment error folded into
`primePartialErr`, which nothing but a clean prime clears, so
`cachedReadyCompleteOnly`'s gate (`caching_store_handles.go:252`) declined every
cache-only read for the life of the process. Measured: reconcile at **24s**
(other rigs 8-50ms), every read falling back to a live 5-6s `bd` subprocess, the
controller tick starved, the order loop dead (`dolt-health` fired twice in 6h
against a 30s interval), 6h of events = 3 beads created / 2 closed against
25,271 `bead.updated` reconcile churn, 108 ready city beads with 0 routed, and
`partial: true` + `bead cache unavailable` on the city and all five rigs. It
survived a supervisor restart.

`isBdSQLUnsupportedInEmbeddedMode` already existed (`bdstore.go:1443`) wired to
exactly one call site — the claim-release path (#5126). The projection, the last
surviving `bd sql` read call site on `main`, referenced it zero times.

## Shape chosen: **both (b) then (a)**

**(b) Backend capability gate — the design fix.** `bd sql` is a raw-database
escape hatch and bd refuses it unless it holds a SQL session; that is a property
of the BACKEND, not the release. gc already has a composition root naming the
backends this build *implements* — reads their metadata shape, projects their
environment, manages their runtime: `contract.RegisteredBackends()` = `dolt`,
`doltlite`, unset. A scope served through the linked beads library under any
other name (maintainer-city's `"backend": "postgres"`) is one gc knows nothing
about, so assuming its `bd sql` works — and that its schema carries gc's
`issues`/`wisps` projection — is the guess. Withholding is the honest default.
This is determinable without guessing, so (b) is preferred, and it costs the
city store **zero** subprocesses.

**(a) Runtime latch — belt and braces, and load-bearing here.**
maintainer-city's rig scopes still declare `backend: dolt` in their own
`metadata.json` while bd opens them elsewhere, so the gate is inert for them.
bd's own refusal now disables the projection for the SCOPE (not the store
object — see MAJOR 2 below), bounding those scopes to exactly one failed call
each per process instead of one per cycle forever.
Metadata gc cannot read falls through to the latch too, so the gate never
fails closed on a parse error.

## The degrade is named, and it costs readiness — stated honestly

`ErrReadyProjectionUnsupported` is a **degrade**, not a partial result, because
it costs the snapshot exactly **one column**, not rows. What it costs
**readiness** is a separate question, and the first revision of this PR got that
answer wrong (see "Review fixes" below).

- `applyReadyProjection` records the cause on the problem log (so `LastProblem`
  reads `... 'bd sql' is not yet supported in embedded mode`) and returns the
  rows, instead of poisoning `primePartialErr`. A *transient* projection failure
  still marks the snapshot partial — only the structural verdict degrades.
- `List` / `Get` / `DepList` keep being served from cache. That is the outage
  fix: `primePartialErr` is what sent every cache-only read to a live 5-6s `bd`
  subprocess for the life of the process.
- **Readiness reads decline the cache and take bd's own verdict live.** The
  cache latches `readyProjectionDegraded`, consulted by `Ready`, `CachedReady`,
  `cachedReadyCompleteOnly` and `cachedReadyLocked`.
- One operator notice per **scope**, naming the reason and the real consequence,
  through the existing `WithBdStoreNoticeSink` seam.

Operators stop seeing `partial: true` / `bead cache unavailable` on the *list*
paths of this city, because the snapshot is no longer partial — the cause
travels on the problem log instead of being inferred from a bare
`ErrCacheUnavailable`.

## Review fixes (two confirmed MAJORs)

**MAJOR 1 — the degrade must not make the cache authoritative for readiness.**
Swallowing the error kept `primePartialErr` nil, and that flag is exactly what
used to route ready reads to the live store (`CachingStore.Ready` →
`c.backing.Ready(...)`, a live `bd ready` carrying bd's verdict). With it nil,
the cache served instead, using `cachedBeadReady` with `IsBlocked == nil` — a
predicate that consults only `readyBlockingDependencyTypes`
(blocks / waits-for / conditional-blocks). bd's `is_blocked` is strictly
stronger: migration `0046_add_is_blocked.up.sql` propagates blocked-ness
**transitively down parent-child edges** (its `reachable` CTE joins
`d.type='parent-child'`), and `bd ready` filters `is_blocked = 0`. bd's Issue
JSON carries no `is_blocked` field, so `bd sql` is gc's only source — with the
projection off, every bead on that store fell to the weaker predicate, and
`cmd/gc/dispatch_control_ready.go`'s per-tick scan reads exactly that handle.
That re-opened #3218 (`d414f7f1ac`), permanently and latched.

Shape chosen: **(c) a distinct cache flag**. `readyProjectionDegraded` is set by
`applyReadyProjection` and consulted only by the four readiness readers, so the
reads that actually depend on `is_blocked` decline to the live backing while
`List`/`Get`/`DepList` keep serving from cache. It keeps the outage fixed (no
per-cycle 6-16s `bd sql`, no poisoned `primePartialErr` for non-ready reads)
without making the cache authoritative for readiness.

Not (a) — teaching `cachedBeadReady` parent-child transitivity: the
direct-deps-only model is gc's standing behaviour for its NATIVE stores too
(`sqliteReadySQL`, `doltliteReadyIssueWhere`, Mem/File), so changing it there is
a larger semantic change and out of scope here. Not (b) — keeping *all* ready
readers on live `backing.Ready()`: acceptable but strictly slower than (c) for
the readers that do not depend on the column.

**MAJOR 2 — the notice was bounded per BdStore OBJECT, not per scope.** The
control-ready path rebuilds the store every `controlReadyCacheTTL` (3s) per
scope, so "once" degraded to "once per rebuild" — the exact defect #5180 fixed
for the unread-store notice. Fixed the same way and with the same pattern: a
package-level `sync.Map` keyed by `filepath.Clean(dir)` holding the latched
verdict plus an atomic `announced` flag, consulted by
`bdReadyProjectionEnabled`. Both guards now share `scopeGuardKey`. The verdict
is still *reported* to every rebuilt store — each one backs a fresh cache that
must learn to send readiness live — but the notice and the failing `bd sql` are
spent once per scope.

**And the claim.** The operator line no longer says "no work is lost" or "cached
ready falls back to dependency-derived readiness". It now reads:

```
gc: ready-projection enrichment disabled for <dir>: <cause>
gc: ready reads on this scope answer from a live `bd ready` instead of the
cache; other cached reads keep serving and no further bd sql is spent.
```

The old wording was literally true and named the wrong risk: the degraded
predicate errs **permissive** — it would OFFER work whose gate has not opened —
which for a dispatcher is worse than hiding work.

## Also fixed: the masked error

bd prints its `BD_OTEL_*` deprecation warning during startup, **before** the
command runs, so on a bd that both nags and fails, stderr line 1 is the nag and
the line saying what broke is line 2. gc's wrapping does not truncate — it
composed `strings.TrimSpace(stderr)` verbatim — but every single-line render of
the wrapped error (the cache problem tile, `gc status`, a journal grep) then
showed the telemetry warning where it used to show the real cause. That is gc's
composition, so it is fixed here: `bdFailureDetail` leads with bd's own
`Error:`/`Hint:` lines (`cmd/bd/errors.go`) and keeps everything else verbatim,
in order, after them. Nothing is dropped, and it also keeps
`isBdSQLUnsupportedInEmbeddedMode` able to classify the wrapped error.

## Tests

Red before the fix (both revisions):

```
bdstore_ready_projection_capability_internal_test.go:76: first enrich error =
  bd sql ready projection: exit status 1: Error: 'bd sql' is not yet supported
  in embedded mode, want ErrReadyProjectionUnsupported

caching_store_ready_projection_internal_test.go:57: primePartialErr = bd sql
  ready projection: ... want nil: an unsupported projection must not make the
  cache permanently unavailable

bdstore_failure_detail_internal_test.go:36: first line of the wrapped error =
  "exit status 1: warning: BD_OTEL_* environment variables are deprecated. ...",
  want bd's actionable Error: line
```

Red before the review fixes:

```
# MAJOR 1 (read sites no longer consult the degrade latch):
caching_store_ready_projection_internal_test.go:217: Ready = [gc-1 gc-3 gc-4],
  want [gc-1 gc-4]: the transitively blocked child gc-3 must not be offered as
  ready

# MAJOR 2 (verdict memoized per store object instead of per scope):
bdstore_ready_projection_capability_internal_test.go:195: operator notice
  printed 5 times across 5 stores over one scope, want exactly 1
bdstore_ready_projection_capability_internal_test.go:211: bd sql ran 5 times
  across 5 stores over one scope; the latch must survive the rebuild
```

`TestDegradedProjectionSendsReadyToTheLiveBdVerdict` is the council's
transitive-block experiment, defeated: `blocker <- blocks - parent <-
parent-child - child`, over a backing store that answers `Ready` the way `bd
ready` does. On the degraded cache, `Ready()` returns `[blocker unrelated]`
(one live backing call), `CachedReady()` is `ok=false`, `ReadyContext` and the
cached-reader `Ready` return `ErrCacheUnavailable` — while `CachedList` still
serves all four rows with zero backing traffic.
`TestHealthyProjectionStillAnswersReadyFromCache` is the control: with the
column present, the same fixture is answered from cache with no live call.

`TestReadyProjectionVerdictIsPerScopeAcrossStoreRebuilds` builds five fresh
`BdStore`s over one scope dir, on both the gate path and the runtime-latch
path: one notice, and — on the latch path — one `bd sql`, total.
`TestReadyProjectionNoticeDoesNotClaimReadinessIsUnaffected` pins the corrected
wording.

Once-only proof (`TestReadyProjectionLatchesOnTheEmbeddedModeRefusal`): five
enrichments, `bd sql` runs exactly once, one notice; every later enrichment
still *names* the degrade, because that error is how each cache over the scope
learns to send readiness live.

Byte-identity for the five Dolt cities
(`TestReadyProjectionOnAnImplementedBackendIsUnchanged`, over
`backend=dolt|doltlite|""`): the exact same two invocations, in the same order,
and the rows come back enriched. Mutation proof: making the gate fire for every
backend turns all three subtests red. `NativeDoltStore` does not implement the
enrichment interface at all, so `applyReadyProjection` returns on its first
branch, unchanged. Two further mutations pin the degrade branch: removing
`errors.Is(err, ErrReadyProjectionUnsupported)` reds the degrade test; making
everything degrade reds the transient-failure control.

`internal/beads/splittest` does not fit here — its leaves model write-time
*store semantics* (bd/Dolt vs SQLite residence rules) over a `beads.Store`, and
this defect lives in `BdStore`'s **subprocess** result. A Postgres-backed
`BdStore` has no splittest leaf and would need one that models a `CommandRunner`
verdict, which is a different kit. The `recordingRunner` already in
`internal/beads` is the right fixture and is what these tests use.

## Gates

`go build ./...`, `go vet ./...`, `gofmt -l` clean, `golangci-lint run ./...`
**0 issues**, `go test ./internal/beads/...`, `go test ./cmd/gc -timeout 25m`
serial (`ok 709.084s`), `go test ./internal/...` (144 packages ok, exit 0),
`scripts/check-core-boundary.sh` OK, `make check-split-topology-rows` exit 0.

### DIFF
--- a/internal/beads/bdstore.go
+++ b/internal/beads/bdstore.go
@@ -196,15 +196,74 @@ func classifyBDExecResult(parent, ctx context.Context, name string, timeout time
 		return "timeout", timeoutErr, timeoutErr
 	}
 	if runErr != nil {
-		detail := strings.TrimSpace(stderr)
-		if detail == "" && name == "bd" {
-			detail = bdStdoutErrorDetail(out)
-		}
-		if detail != "" {
+		if detail := bdFailureDetail(name, out, stderr); detail != "" {
 			return "error", runErr, fmt.Errorf("%w: %s", runErr, detail)
 		}
 	}
 	return "done", runErr, runErr
+}
+
+// bdFailureDetail composes the operator-facing detail for a failed invocation
+// out of the two streams bd splits its output across.
+//
+// bd writes unrelated startup notices to stderr BEFORE the command runs — the
+// BD_OTEL_* deprecation warning is the live example — so on a bd that both nags
+// and fails, stderr's FIRST line is the nag and the line saying what actually
+// broke sits below it. Every single-line render of the wrapped error (the cache
+// problem tile, `gc status`, a journal grep) then shows the nag. That is how a
+// permanently failing `bd sql ready projection` on maintainer-city read as a
+// telemetry warning for a night while the controller starved.
+//
+// So the detail leads with bd's own Error:/Hint: lines (cmd/bd/errors.go) and
+// keeps everything else verbatim, in order, after them. NOTHING is dropped, and
+// every other shape — a single line, stderr that already leads with the error,
+// stderr with no bd error line, empty stderr, a non-bd command — composes
+// byte-for-byte what it composed before.
+func bdFailureDetail(name string, out []byte, stderr string) string {
+	detail := strings.TrimSpace(stderr)
+	if name != "bd" {
+		return detail
+	}
+	if detail == "" {
+		// bd writes structured errors to stdout under --json while stderr is
+		// often empty.
+		return bdStdoutErrorDetail(out)
+	}
+	return hoistBdErrorLines(detail)
+}
+
+// bdErrorLinePrefixes are the prefixes bd stamps on the lines that say what
+// failed (cmd/bd/errors.go: HandleError, HandleErrorWithHint).
+var bdErrorLinePrefixes = []string{"Error:", "Fatal:", "Hint:"}
+
+func hoistBdErrorLines(detail string) string {
+	lines := strings.Split(detail, "\n")
+	if len(lines) < 2 || isBdErrorLine(lines[0]) {
+		return detail
+	}
+	leading := make([]string, 0, len(lines))
+	trailing := make([]string, 0, len(lines))
+	for _, line := range lines {
+		if isBdErrorLine(line) {
+			leading = append(leading, line)
+			continue
+		}
+		trailing = append(trailing, line)
+	}
+	if len(leading) == 0 {
+		return detail
+	}
+	return strings.Join(append(leading, trailing...), "\n")
+}
+
+func isBdErrorLine(line string) bool {
+	trimmed := strings.TrimSpace(line)
+	for _, prefix := range bdErrorLinePrefixes {
+		if strings.HasPrefix(trimmed, prefix) {
+			return true
+		}
+	}
+	return false
 }
 
 // bdExecTimeoutError formats the deadline error after the per-command context
--- a/internal/beads/bdstore.go
+++ b/internal/beads/bdstore.go
@@ -373,6 +373,12 @@ type BdStore struct {
 	readyProjectionMu      sync.Mutex
 	readyProjectionChecked bool
 	readyProjectionEnabled bool
+	// readyProjectionScope memoizes, per SCOPE PATH, the verdict that this
+	// ledger cannot serve the ready projection at all. Shared with every other
+	// store rooted at the same directory, because cmd/gc rebuilds a store per
+	// request and the control dispatcher rebuilds one every few seconds. See
+	// bdstore_ready_projection.go.
+	readyProjectionScope *readyProjectionScopeGuard
 
 	// condReleaseLatchedUnsupported records that this bd rejected the
 	// conditional-release flags, pinning ReleaseIfCurrent to the raw-SQL
@@ -472,11 +478,12 @@ func NewBdStore(dir string, runner CommandRunner, opts ...BdStoreOption) *BdStor
 // NewBdStoreWithPrefix creates a BdStore with an explicit owned bead ID prefix.
 func NewBdStoreWithPrefix(dir string, runner CommandRunner, idPrefix string, opts ...BdStoreOption) *BdStore {
 	s := &BdStore{
-		dir:          dir,
-		runner:       runner,
-		idPrefix:     normalizeIDPrefix(idPrefix),
-		unreadStore:  guardForScope(dir),
-		localStrings: newLocalSidecar(bdLocalSidecarPath(dir)),
+		dir:                  dir,
+		runner:               runner,
+		idPrefix:             normalizeIDPrefix(idPrefix),
+		unreadStore:          guardForScope(dir),
+		readyProjectionScope: readyProjectionGuardForScope(dir),
+		localStrings:         newLocalSidecar(bdLocalSidecarPath(dir)),
 	}
 	for _, opt := range opts {
 		if opt != nil {
--- a/internal/beads/bdstore_failure_detail_internal_test.go
+++ b/internal/beads/bdstore_failure_detail_internal_test.go
@@ -1,0 +1,95 @@
+package beads
+
+import (
+	"context"
+	"errors"
+	"strings"
+	"testing"
+	"time"
+)
+
+const bdOtelDeprecationWarning = "warning: BD_OTEL_* environment variables are deprecated. Replace with BD_OTEL_ENABLED=true plus the standard OpenTelemetry SDK variables. Translated for this run: BD_OTEL_METRICS_URL"
+
+func classifyForTest(t *testing.T, name string, out []byte, stderr string) error {
+	t.Helper()
+	ctx := context.Background()
+	_, _, resultErr := classifyBDExecResult(ctx, ctx, name, time.Minute, time.Now(), out, stderr, errors.New("exit status 1"))
+	return resultErr
+}
+
+// TestBdFailureDetailLeadsWithBdsErrorLine is the diagnosis half of the
+// maintainer-city outage: bd prints its BD_OTEL_* deprecation warning during
+// startup, BEFORE the command runs, so the actionable line is the SECOND line of
+// stderr. Every single-line render of the wrapped error — the cache problem
+// tile, a journal grep — then showed a telemetry nag where it used to show the
+// reason the read failed, and the real cause went unfound for a night.
+//
+// The detail keeps every byte bd wrote and only reorders it, so the actionable
+// line leads.
+func TestBdFailureDetailLeadsWithBdsErrorLine(t *testing.T) {
+	stderr := bdOtelDeprecationWarning + "\nError: 'bd sql' is not yet supported in embedded mode\n"
+
+	err := classifyForTest(t, "bd", nil, stderr)
+
+	first, _, _ := strings.Cut(err.Error(), "\n")
+	if !strings.Contains(first, "not yet supported in embedded mode") {
+		t.Fatalf("first line of the wrapped error = %q, want bd's actionable Error: line", first)
+	}
+	if !strings.Contains(err.Error(), "BD_OTEL_*") {
+		t.Errorf("reordering dropped the warning; the detail must keep every byte bd wrote: %v", err)
+	}
+	if !isBdSQLUnsupportedInEmbeddedMode(err) {
+		t.Errorf("the embedded-mode classifier no longer recognizes the wrapped error: %v", err)
+	}
+}
+
+// TestBdFailureDetailIsUnchangedWithoutNoise is the byte-identity guard: every
+// shape that is not "a warning printed before bd's own error line" composes
+// exactly the detail it composed before.
+func TestBdFailureDetailIsUnchangedWithoutNoise(t *testing.T) {
+	cases := []struct {
+		name   string
+		cmd    string
+		out    []byte
+		stderr string
+		want   string
+	}{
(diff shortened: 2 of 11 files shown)


julianknutsen:
## Review fixes pushed (`c47179825b`) — two confirmed MAJORs

### MAJOR 1 — the degrade made the cache authoritative for readiness with a weaker predicate

Confirmed. Swallowing `ErrReadyProjectionUnsupported` keeps `primePartialErr`
nil, and that flag is exactly what used to route ready reads to the live store
(`CachingStore.Ready` → `c.backing.Ready(...)`, a live `bd ready` carrying bd's
verdict). With it nil the cache served instead, using `cachedBeadReady` with
`IsBlocked == nil`, which consults only
`readyBlockingDependencyTypes` = blocks / waits-for / conditional-blocks.
bd's `is_blocked` propagates blocked-ness **transitively down parent-child
edges** (`0046_add_is_blocked.up.sql`, `reachable` CTE) and `bd ready` filters
`is_blocked = 0`; bd's Issue JSON carries no `is_blocked` field, so `bd sql` is
gc's only source. `cmd/gc/dispatch_control_ready.go` answers the control
dispatcher's per-tick readiness scan from `cache.CachedReady()`, so on exactly
the stores this PR gates, the queue's data source flipped to the weaker
predicate — #3218 re-opened, permanently and latched, in the **permissive**
direction.

**Shape chosen: (c)** — a distinct cache latch, `readyProjectionDegraded`, set
by `applyReadyProjection` and consulted by `Ready`, `CachedReady`,
`cachedReadyCompleteOnly` and `cachedReadyLocked`.

Why (c) over the alternatives:

- **not (a)** (teach `cachedBeadReady` parent-child transitivity): the
  direct-deps-only model is gc's standing behaviour for its NATIVE stores too
  (`sqliteReadySQL`, `doltliteReadyIssueWhere`, Mem/File). Changing the shared
  predicate changes readiness on every store gc owns, which is a much larger
  semantic change than this fix.
- **not (b)** (all ready readers live): correct but strictly slower than (c) —
  it also sends the readers that never depended on the column to a subprocess.
- **(c)** declines exactly the reads that depend on `is_blocked`. `List` / `Get`
  / `DepList` keep serving from cache, so the outage this PR exists to fix stays
  fixed: no per-cycle 6-16s `bd sql`, no poisoned `primePartialErr` for
  non-readiness reads.

Transitive-block experiment, reproduced and defeated
(`TestDegradedProjectionSendsReadyToTheLiveBdVerdict`): `blocker <- blocks -
parent <- parent-child - child`, plus an unrelated control, over a backing store
that answers `Ready` the way `bd ready` does.

| read | before the fix | after |
| --- | --- | --- |
| `cache.Ready()` | `[blocker child unrelated]` | `[blocker unrelated]` (1 live backing call) |
| `cache.CachedReady()` | `ok=true`, includes child | `ok=false` |
| `cache.ReadyContext()` | includes child | `ErrCacheUnavailable` |
| `Handles().Cached.Ready()` | includes child | `ErrCacheUnavailable` |
| `cache.CachedList()` | serves 4 rows | serves 4 rows, 0 backing calls |

Red before, with the four read sites reverted:

```
caching_store_ready_projection_internal_test.go:217: Ready = [gc-1 gc-3 gc-4],
  want [gc-1 gc-4]: the transitively blocked child gc-3 must not be offered as
  ready
```

`TestHealthyProjectionStillAnswersReadyFromCache` is the control: with the
column present, the same fixture is answered from cache with no live call.

### MAJOR 2 — the notice was bounded per BdStore object, not per scope

Confirmed, and fixed with the pattern #5180 already established in the file this
PR edits: a package-level `sync.Map` keyed by `filepath.Clean(dir)` holding the
latched verdict (`atomic.Pointer`) plus an atomic `announced` flag, consulted by
`bdReadyProjectionEnabled`. `guardForScope` and `readyProjectionGuardForScope`
now share `scopeGuardKey`.

The verdict is still **reported** to every rebuilt store — each one backs a
fresh cache that must learn to send readiness live — but the notice and the
failing `bd sql` are spent once per scope.

Red before, five fresh `BdStore`s over one scope dir
(`TestReadyProjectionVerdictIsPerScopeAcrossStoreRebuilds`):

```
operator notice printed 5 times across 5 stores over one scope, want exactly 1
bd sql ran 5 times across 5 stores over one scope; the latch must survive the
  rebuild
```

### The claim, corrected

The notice no longer says "no work is lost" or "cached ready falls back to
dependency-derived readiness". It now reads:

```
gc: ready-projection enrichment disabled for <dir>: <cause>
gc: ready reads on this scope answer from a live `bd ready` instead of the
cache; other cached reads keep serving and no further bd sql is spent.
```

The old wording was literally true and named the wrong risk: the degraded
predicate errs **permissive** — it would OFFER work whose gate has not opened —
which for a dispatcher is worse than hiding work. `ErrReadyProjectionUnsupported`'s
doc and `applyReadyProjection`'s doc were rewritten the same way; "the snapshot
is whole" is now "it costs the snapshot exactly one column, not rows", with the
readiness consequence spelled out. Pinned by
`TestReadyProjectionNoticeDoesNotClaimReadinessIsUnaffected`.

### Kept, per the council

Backend gate keyed on `contract.RegisteredBackends()`; the `dolt_mode: server`
decoy; the latch being load-bearing for MC's rig scopes; the `bdFailureDetail`
stderr hoist. Byte-identity for `backend=dolt|doltlite|""` is unchanged and
still pinned by mutation (`TestReadyProjectionOnAnImplementedBackendIsUnchanged`
— same two invocations, same order, rows enriched, no notice).

### Gates

`go build ./...`, `go vet ./...`, `gofmt -l` clean, `golangci-lint run ./...`
**0 issues**, `go test ./internal/beads/...` ok, `go test ./cmd/gc -timeout 25m`
serial (`ok 709.084s`), `go test ./internal/...` (144 packages ok, exit 0),
`scripts/check-core-boundary.sh` OK, `./scripts/check-split-topology-rows.sh`
exit 0.

PR #5184, opened 2026-08-11
fix(beads): stop version_compat failing a comparison it cannot make (ga-40qh1)

julianknutsen:
Fixes ga-40qh1 (P0).

## The defect

`checkVersionCompat` string-compares the linked beads module version to the version bd reports for itself. That comparison only means something when both strings name the same released beads artifact. Two shapes in daily use do not:

- this fork replaces the module — `replace github.com/steveyegge/beads => github.com/gascity/bd-enterprise v0.0.0-20260810084121-1aa7bf160786` — so `debug.ReadBuildInfo()` reports the *replacement's* pseudo-version;
- `origin/main` itself pins beads at an untagged commit (`v1.1.1-0.20260805093327-bf97b73749ac`), which is a pseudo-version even with no replace at all.

bd reports `1.1.0`. Neither string can ever equal it, so the check returned `FAIL` — a **confirmed mismatch** — for a question it never had the evidence to ask. The existing `""` / `(devel)` bypass was written for precisely this situation ("the schema version is validated above and is the real compatibility signal"), but its set of unknowns was incomplete.

## Why it only bit some scopes, and why that is the real bug

The compare is reached only when `bd context` is readable.

| scope | `bd context` | version_compat | store |
|---|---|---|---|
| city root (not a git repo) | errors | WARN | `NativeDoltStore` — healthy |
| every rig (is a git repo) | succeeds | **FAIL** | `BdStore` — degraded |

On `BdStore`, `listIncludesCompleteDependencies()` is hardcoded false (ga-tgpfm), so `depsComplete` is permanently false and `cachedReadyCompleteOnly` declines every read. Those scopes report `partial: true` / `bead cache unavailable` forever and serve every complete-ready read from a live bd subprocess (~640 avoidable forks per three minutes, fleet-wide). Verified 1:1 on the live box: every scope logging `gate=version_compat` reports the partial error, and no other scope does.

**The fleet was healthy only where a capability probe happened to fail.** One probe failing gave you the healthy path; the other succeeding gave you the degraded one.

## The fix

Classify the linked library, then compare only when the comparison is answerable. A library version is unconfirmable when it is empty, `(devel)`, supplied by a `replace` directive (either form — a module replace records the replacement's own numbering; a local-path replace records none), or a pseudo-version naming an untagged commit. Unconfirmable **PASSes**, with the reason named in the summary; `schema_version` above remains the real compatibility signal. A **confirmed** mismatch between two comparable releases still **FAILs**.

`PASS` rather than `WARN` is load-bearing. `WARN` degrades the verdict, and the `degradedOnlyByUnreachableBDContext` upgrade requires `bdCtxErr != nil` — so a `WARN` here would leave every rig on `BdStore` exactly as before. Only `PASS` actually answers the bead.

`beadsModuleVersion` also reported the displaced `require` version for a local-path replace, describing code that is not linked. It now reports the replacement's own version, empty included.

## Fail-open vs fail-closed — the argument

Fail-open on an *unanswerable* question, fail-closed on a *confirmed* one.

Every other `FAIL` in this checker is a confirmed incompatibility: backend disagreement, `dolt_mode=embedded`, `project_id` mismatch. "I could not cross-verify" already means WARN-and-proceed on the `bd context` path. "I could not compare two strings" is the same epistemic state and was the only place answering it with a confirmed verdict.

Fail-closed bought no safety here. The question is unanswerable **by construction** for every fork build, every source build, and (today) for `origin/main` itself — so the gate was never selective, it was always on. What it cost was large: a silent downgrade to `BdStore` that permanently disables complete-ready cache reads. That asymmetry — a huge behavioural consequence attached to a comparison failure — is the actual defect.

The guards that matter are untouched: `schema_version` must still be reported (FAIL if not), the preflight `project_id` probe must still match, and beadslib's native-open still verifies `_project_id` over the authenticated connection and refuses to connect on mismatch, dropping the scope to `BdStore`.

## What changes on the live fleet at deploy

Rig scopes currently logging `native_store_unavailable gate=version_compat` move `BdStore` → `NativeDoltStore`. That is the **correct** state, and it is a real behaviour change on six live cities:

- `/status` partial errors (`reading complete ready projection from cache: bead cache unavailable`) clear on those scopes;
- complete-ready reads start being served from the cache instead of a live bd subprocess;
- the bd fork rate drops sharply (~640 forks/3min recovered).

Risks worth watching:

- **More native Dolt connections.** Each promoted rig opens pooled connections to the Dolt server instead of forking bd. Watch Dolt connection counts on promotion.
- **Native-store bugs previously masked.** These scopes have been on `BdStore` since 09:15 Aug 11; any `NativeDoltStore`-specific behaviour they were shielded from now applies to them.
- **Cache goes authoritative.** `depsComplete` becomes true, so ready reads come from cache. A stale cache now shows up as wrong readiness rather than as a slow-but-correct subprocess read.

Mitigations already in the code path: a scope with executable `.beads/hooks/on_{create,update,close}` still falls back (`gate=bd_hooks`); a failed native open falls back (`gate=native_open`); and `GC_BEADS_FORCE_FALLBACK=1` reverts a scope wholesale. **No deploy or restart is performed by this PR.**

## Tests

Red before, on unmodified production code, for both live pseudo-versions against bd `1.1.0`:

```
state = "FAIL", want "PASS" (summary: "bd version differs from linked beads library version")
```

Both directions proven:

- **still fails** — `1.1.1` vs bd `1.1.0`, `v1.2.0` vs bd `v1.1.0`, `1.0.5` vs `1.0.4`, plus end-to-end `TestPreflightBlocksOnRealVersionSkew` (BLOCKED, `Fallback=bd_store`);
- **now passes** — enterprise `0.0.0-20260810084121-1aa7bf160786`, `origin/main`'s `1.1.1-0.20260805093327-bf97b73749ac`, a tagged replacement, a path replacement with no version, plus end-to-end `TestPreflightEligibleOnReplacedBeadsModuleWithReadableBDContext` (ELIGIBLE with `bd context` readable, *not* via the identity fallback).

Also covered: `(devel)`, empty, missing dep, and the build-info reader for plain-require / module-replace / local-path-replace.

**Byte-identity where the check already passed** is pinned by `TestCheckVersionCompatSummariesAreStableWhereItAlreadyPassed`, which asserts the exact summary string of all six pre-existing outcomes. Proven by mutation — all seven mutants killed, including the two that would change the live cities' path:

| mutation | killed by |
|---|---|
| unreachable-bd-context `WARN`→`PASS` (the cities' path) | `TestPreflightEligibleOnUnreachableBDContextWhenIdentityVerified`, `TestPreflightDegradesOnUnreachableBDContextWithoutIdentityProof` |
| unreachable-bd-context summary reworded | summary-stability test |
| `(devel)` "source build" reason reworded | summary-stability test |
| classifier declares every version unknown | `TestCheckVersionCompatSourceBuild`, summary-stability, `TestPreflightBlocksOnRealVersionSkew` |
| matching-release PASS summary reworded | summary-stability test |
| replaced-module signal ignored | `TestCheckVersionCompatSourceBuild` |
| build-info reader ignores the replace directive | `TestLinkedBeadsLibraryFromBuildInfo` |

## Second commit: the native store's cache had no `is_blocked` column (MAJOR, fixed here)

Promoting rig scopes to `NativeDoltStore` hands their readiness to the `CachingStore` **and simultaneously removes the column that made the cache's answer equal bd's**. Left as it was, this PR would have re-created **#3218** on six live cities.

`NativeDoltStore.listIncludesCompleteDependencies()` returns `true` (vs `BdStore`'s `false`), so `fetchDepsForBeads` latches `depsComplete = true` — exactly what unblocks `CachingStore.Ready` and `cachedReadyCompleteOnly`. But `NativeDoltStore` implemented **neither** `enrichReadyProjectionForCache` nor `dependencySnapshotForCache`, and `beadFromNativeIssue` cannot set `Bead.IsBlocked` — beadslib's `types.Issue` has no `is_blocked` field; the flag is reachable only through `beads.AsBlockedQuerier`. So `applyReadyProjection` returned items unchanged with a **nil** error, `readyProjectionDegraded` never latched, `readyReadsMustGoLive()` stayed false, every cached row carried `IsBlocked == nil`, and `cachedBeadReady` fell through to the **direct-dependency** predicate.

That predicate is weaker than the column two ways, both of which gascity hits constantly:

- `is_blocked` propagates transitively **down parent-child** edges (`issueops.markBlockedTemplateForIssues`: `d.type = 'parent-child' AND p.is_blocked = 1`), so a child of a blocked parent carries no blocking edge of its own and reads ready to the cache;
- `cachedBeadReady` treats a dep as blocking only when `statusByID[dep.DependsOnID]` is present, so an edge onto **another store's row** — a relocated `gcg-` graph bead — is invisible to it.

`NativeDoltStore.Ready` goes through `GetReadyWork`, which filters `is_blocked = 0` (`sqlbuild.ReadyWhere`). Measured on `174bfc6ff8`: `backing.Ready = [blocker unrelated]` while `Ready` / `ReadyContext` / `CachedReady` / `Handles().Cached.Ready` all returned `[blocker child xstore unrelated]`, nil error, `ok=true`. `CachedReady()` is what `cmd/gc/dispatch_control_ready.go` reads on every dispatcher tick.

It also violated the invariant `TestDegradedProjectionSendsReadyToTheLiveBdVerdict` (#5183, merged hours earlier) established on the identical fixture — that test just guarded only the `BdStore` path.

### The fix: fill the column, don't decline it

`NativeDoltStore.enrichReadyProjectionForCache` batch-fills `IsBlocked` through `beads.BlockedQuerier.IsBlockedBatch`, mirroring `BdStore.fetchReadyProjection` **including the `skipBDReadyProjectionEnrichment` exclusions**, so message/nudge rows keep their nil and the reconcile diff still converges. The read runs in-process on the store's existing connection and inside `withReadRetry`, so it recovers from a managed-Dolt rebind like every other native read.

The fail-safe is kept as the fallback, not the design: a storage that cannot answer the column reports `ErrReadyProjectionUnsupported`, `readyReadsMustGoLive()` becomes true, and readiness declines the cache for the backing's own `Ready`. That branch is real, not defensive — the core `beads.Storage` interface does not declare `IsBlockedBatch`; only `DoltStorage` composes `DependencyQueryStore`. Both production storages (`DoltStore`, `EmbeddedDoltStore`) satisfy it, pinned upstream by compile-time assertions.

### Why fill rather than decline

The declining variant is correct but costs what this PR exists to buy back. Filling costs `ceil(N/200) × 2` primary-key `SELECT id, is_blocked FROM {issues,wisps} WHERE id IN (...)` statements, once per cache prime and once per reconcile (30–120 s by corpus size). Declining costs a full live `NativeDoltStore.Ready()` — two `GetReadyWork` calls with complete row hydration — on **every** readiness read, at up to one per `controlReadyCacheTTL` (3 s) per scope. That is 10–40× the frequency at many times the per-call cost. Measured on the in-memory harness at n=3000, the enrichment was ~0.36× the cost of a single live `Ready()`, and it replaces one per tick with one per reconcile. It also replaces `BdStore`'s equivalent, which spent a 6–16 s `bd sql` subprocess on maintainer-city.

### Tests

Red before, on `174bfc6ff8`, for both shapes and all four readiness handles:

```
Ready offered gc-3, a child of a blocked parent that the backing's own Ready hides:
blocked-ness propagates down parent-child and the cache's direct-dep predicate
cannot see it (backing = [gc-1 gc-5], cache = [gc-1 gc-3 gc-4 gc-5])

Ready offered gc-4, whose blocker gcg-offscope is not resident in this scope's cache:
cachedBeadReady treats a dep as blocking only when statusByID holds the target
(backing = [gc-1 gc-5], cache = [gc-1 gc-3 gc-4 gc-5])
```

and for the extended #5183 invariant:

```
a native backing that cannot answer the is_blocked column must latch the
ready-projection degrade
```

- `TestCachedReadyOverNativeBackingNeverOffersWorkTheBackingHides` asserts the invariant directly — a cached ready read over a native backing never returns a bead the backing's own `Ready()` excludes — across `Ready`, `ReadyContext`, `CachedReady` and `Handles().Cached.Ready`, on the council's fixture (`blocker <-blocks- parent <-parent-child- child`) plus a cross-store bead blocked by a non-resident `gcg-` row. It also asserts the cache still **answers** and answers equal to the backing, so a trivially-declining implementation cannot pass it.
- `TestNativeBackingWithoutTheBlockedColumnSendsReadyToTheLiveVerdict` extends #5183's invariant to the native path: all four handles decline, `Ready` takes the live verdict, and non-readiness cached reads keep serving.
- `TestNativeReadyProjectionIsOneBatchedReadPerCycle` pins the cost shape — one batched read over the whole active set per cycle, never a per-bead fan-out.
- `var _ readyProjectionEnrichmentStore = (*NativeDoltStore)(nil)` pins the coupling, so deleting the method breaks the build rather than silently reverting readiness.

The version_compat fix above is untouched: the fail-open decision, the PASS-not-WARN reasoning, and all six pinned byte-identity outcomes stand exactly as committed.

### What this adds to the deploy story

The promotion described above still happens, and now it is **transitively correct**. On a promoted rig, cached readiness equals `bd ready`'s verdict — a child of a blocked parent and a bead gated by a non-resident cross-store row stay hidden, exactly as they do today under `BdStore` — but without the `bd sql` subprocess. Without this commit, both would have been offered to the control dispatcher on every tick. Scopes on `BdStore`, DoltLite, exec or mem are unaffected.

## Follow-up

`BdStore.listIncludesCompleteDependencies() == false` — the second hardcoded pessimism that amplified this bug's blast radius — is **already filed as ga-tgpfm (P0) and linked to ga-40qh1**. Not fixed here: it is not trivial (BdStore would need a real `dependencySnapshotForCache` over `DepListBatch`, and flipping the flag without actually supplying deps would make blocked beads read as ready — a correctness bug, not a perf one).

## Gates

`go build ./...` · `go vet ./...` · `gofmt` clean · `golangci-lint run ./...` **0 issues** · `go test ./internal/beads/...` · `go test ./cmd/gc -timeout 25m -p 1 -parallel 1` **ok 687.324s** · `go test ./internal/...` · `scripts/check-core-boundary.sh` OK · `make check-split-topology-rows` OK.

No `go.mod` / `go.sum` change: `golang.org/x/mod` was already a direct dependency. No wire-type or OpenAPI change.

### DIFF
--- a/internal/beads/contract/preflight_checker.go
+++ b/internal/beads/contract/preflight_checker.go
@@ -7,6 +7,9 @@ import (
 	"runtime/debug"
 	"strings"
 
+	"golang.org/x/mod/module"
+	"golang.org/x/mod/semver"
+
 	"github.com/gastownhall/gascity/internal/fsys"
 )
 
@@ -41,6 +44,11 @@ type PreflightChecker struct {
 	// BeadsLibraryVersion is the linked github.com/steveyegge/beads module
 	// version. Empty means infer it from build info.
 	BeadsLibraryVersion string
+	// BeadsLibraryReplaced reports whether a go.mod replace directive supplied
+	// the linked beads library, in which case its version identifies the
+	// replacement rather than a beads release. Set together with
+	// BeadsLibraryVersion; leaving both zero infers them from build info.
+	BeadsLibraryReplaced bool
 }
 
 // Check runs the beads backend preflight for scope and returns typed diagnostics.
@@ -234,10 +242,8 @@ func (c PreflightChecker) checkIdentityMatch(scope string, metadata preflightMet
 }
 
 func (c PreflightChecker) checkVersionCompat(ctx PreflightBDContext, err error) PreflightCheckResult {
-	libraryVersion := strings.TrimPrefix(strings.TrimSpace(c.BeadsLibraryVersion), "v")
-	if libraryVersion == "" {
-		libraryVersion = strings.TrimPrefix(beadsModuleVersion(), "v")
-	}
+	library := c.linkedBeadsLibrary()
+	libraryVersion := strings.TrimPrefix(library.Version, "v")
 	details := PreflightDetails{
 		BDVersion:           ctx.BDVersion,
 		BeadsLibraryVersion: libraryVersion,
@@ -255,14 +261,16 @@ func (c PreflightChecker) checkVersionCompat(ctx PreflightBDContext, err error) 
 	if ctx.BDVersion == "" {
 		return NewPreflightCheckResult(PreflightCheckVersionCompat, PreflightCheckWarn, "bd/beads version compatibility could not be confirmed", details)
 	}
-	if libraryVersion == "" || libraryVersion == "(devel)" {
-		// A local-path/replace (source) build of the linked beads library
-		// reports no module version ("(devel)") even though gc and bd are
-		// built from the same source. The schema version is validated above
-		// and is the real compatibility signal, so an unconfirmable library
-		// version must not take the native store offline — only a *confirmed*
-		// mismatch (below) should.
-		return NewPreflightCheckResult(PreflightCheckVersionCompat, PreflightCheckPass, "bd/beads schema compatible; linked library version unconfirmed (source build)", details)
+	if reason := library.unconfirmableReason(); reason != "" {
+		// The compare below only means something when both sides name the same
+		// released beads artifact. When the linked library does not — a source
+		// build, a replaced module, or a pseudo-version naming an untagged
+		// commit — the two strings can never be equal, and answering "mismatch"
+		// reports a verdict the check never had the evidence to reach. The
+		// schema version is validated above and is the real compatibility
+		// signal, so an unconfirmable library version must not take the native
+		// store offline; only a *confirmed* mismatch (below) should.
+		return NewPreflightCheckResult(PreflightCheckVersionCompat, PreflightCheckPass, "bd/beads schema compatible; linked library version unconfirmed ("+reason+")", details)
 	}
 	if strings.TrimPrefix(ctx.BDVersion, "v") != libraryVersion {
 		return NewPreflightCheckResult(PreflightCheckVersionCompat, PreflightCheckFail, "bd version differs from linked beads library version", details)
@@ -296,20 +304,77 @@ func preflightFallbackReason(checks []PreflightCheckResult) string {
 	return ""
 }
 
-func beadsModuleVersion() string {
+// beadsModulePath is the module path of the beads library gc links.
+const beadsModulePath = "github.com/steveyegge/beads"
+
+// linkedBeadsLibrary resolves the beads library this binary links, preferring
+// the configured override so tests need not depend on their own build info.
+func (c PreflightChecker) linkedBeadsLibrary() beadsLibrary {
+	if version := strings.TrimSpace(c.BeadsLibraryVersion); version != "" || c.BeadsLibraryReplaced {
+		return beadsLibrary{Version: version, Replaced: c.BeadsLibraryReplaced}
+	}
 	info, ok := debug.ReadBuildInfo()
 	if !ok {
+		return beadsLibrary{}
+	}
+	return linkedBeadsLibraryFrom(info)
+}
+
+// beadsLibrary describes the beads library a binary actually links.
+type beadsLibrary struct {
+	// Version is the module version recorded in build info, empty when unknown.
+	Version string
+	// Replaced reports whether a go.mod replace directive supplied the code.
+	// A replacement's version numbers belong to the replacement, not to beads,
+	// so they say nothing about which bd release the library agrees with.
+	Replaced bool
+}
+
+// unconfirmableReason names why the library version cannot be compared to a bd
+// release version, or returns "" when the comparison is meaningful.
+func (b beadsLibrary) unconfirmableReason() string {
+	version := strings.TrimPrefix(strings.TrimSpace(b.Version), "v")
+	switch {
+	case version == "" || version == "(devel)":
+		// A source build reports "(devel)" (or nothing) even though gc and bd
+		// are built from the same tree.
+		return "source build"
+	case b.Replaced:
+		return "replaced module"
+	case !comparableReleaseVersion(version):
+		// A pseudo-version names an untagged commit, not a release. gc pins
+		// beads at one whenever it tracks beads ahead of its last tag.
+		return "pseudo-version"
+	default:
 		return ""
 	}
-	for _, dep := range info.Deps {
-		if dep.Path == "github.com/steveyegge/beads" {
-			if dep.Replace != nil && dep.Replace.Version != "" {
-				return dep.Replace.Version
-			}
-			return dep.Version
-		}
+}
+
+// comparableReleaseVersion reports whether version names a released beads
+// artifact, and can therefore be compared with the version bd reports for
+// itself. Pseudo-versions and non-semver strings cannot.
+func comparableReleaseVersion(version string) bool {
+	canonical := "v" + strings.TrimPrefix(strings.TrimSpace(version), "v")
+	return semver.IsValid(canonical) && !module.IsPseudoVersion(canonical)
+}
+
+// linkedBeadsLibraryFrom extracts the linked beads library from build info.
+func linkedBeadsLibraryFrom(info *debug.BuildInfo) beadsLibrary {
+	if info == nil {
+		return beadsLibrary{}
 	}
-	return ""
+	for _, dep := range info.Deps {
+		if dep == nil || dep.Path != beadsModulePath {
+			continue
+		}
+		if dep.Replace != nil {
+			// Report the replacement's own version — including the empty one a
+			// local-path replace records — never the require line it displaced.
+			return beadsLibrary{Version: dep.Replace.Version, Replaced: true}
+		}
+		return beadsLibrary{Version: dep.Version}
+	}
+	return beadsLibrary{}
 }
 
 type preflightMetadata struct {
--- a/internal/beads/contract/preflight_checker_test.go
+++ b/internal/beads/contract/preflight_checker_test.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"os"
 	"path/filepath"
+	"runtime/debug"
 	"strings"
 	"testing"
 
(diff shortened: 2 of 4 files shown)


julianknutsen:
## Confirmed MAJOR fixed in `a5cd367b81` — the native store had no `is_blocked` column for its cache

The finding was correct and the failure was reproduced on `174bfc6ff8` before touching anything.

**What was wrong.** This PR flips rig scopes `BdStore` → `NativeDoltStore`, which hands rig readiness to the cache (`listIncludesCompleteDependencies()` returns `true`, latching `depsComplete`) while removing the column that made the cache's answer equal bd's. `NativeDoltStore` implemented neither `enrichReadyProjectionForCache` nor `dependencySnapshotForCache`, and `beadFromNativeIssue` cannot set `Bead.IsBlocked` — beadslib's `types.Issue` has no `is_blocked` field. So `applyReadyProjection` returned items unchanged with a **nil** error, `readyProjectionDegraded` never latched, and `cachedBeadReady` fell through to the direct-dependency predicate — which neither propagates down parent-child nor sees an edge onto a non-resident row. `NativeDoltStore.Ready` meanwhile filters bd's transitive `is_blocked = 0`. Measured on the head: `backing.Ready = [blocker unrelated]`, `CachedReady() = [blocker child xstore unrelated]`, nil error, `ok=true`. That is what `cmd/gc/dispatch_control_ready.go` reads every dispatcher tick — #3218, on six live cities.

**Fix shape chosen: the preferred one — fill the column.** `NativeDoltStore.enrichReadyProjectionForCache` batch-fills `IsBlocked` through `beads.AsBlockedQuerier` / `IsBlockedBatch`, mirroring `BdStore.fetchReadyProjection` including the `skipBDReadyProjectionEnrichment` exclusions, inside `withReadRetry` so it survives a managed-Dolt rebind. `ErrReadyProjectionUnsupported` is kept as the fallback for a storage that cannot answer the column — a real branch, since core `beads.Storage` does not declare `IsBlockedBatch` (only `DoltStorage` composes `DependencyQueryStore`); both production storages do satisfy it.

**Why not the fallback-only variant.** It is correct but pays back exactly what this PR buys. Filling: `ceil(N/200) × 2` primary-key `SELECT id, is_blocked FROM {issues,wisps} WHERE id IN (...)`, once per prime and once per reconcile (30–120 s). Declining: a full live `NativeDoltStore.Ready()` — two `GetReadyWork` calls with complete row hydration — per readiness read, up to one per `controlReadyCacheTTL` (3 s) per scope. 10–40× the frequency at many times the per-call cost. On the in-memory harness at n=3000 the enrichment measured ~0.36× one live `Ready()`, and it replaces one-per-tick with one-per-reconcile.

**Red-before strings** (`174bfc6ff8`, reproduced on all four readiness handles):

```
Ready offered gc-3, a child of a blocked parent that the backing's own Ready hides:
blocked-ness propagates down parent-child and the cache's direct-dep predicate
cannot see it (backing = [gc-1 gc-5], cache = [gc-1 gc-3 gc-4 gc-5])

Ready offered gc-4, whose blocker gcg-offscope is not resident in this scope's cache:
cachedBeadReady treats a dep as blocking only when statusByID holds the target
(backing = [gc-1 gc-5], cache = [gc-1 gc-3 gc-4 gc-5])
```

and for the extended #5183 invariant:

```
a native backing that cannot answer the is_blocked column must latch the
ready-projection degrade
```

**#5183's invariant now covers the native path.** `TestNativeBackingWithoutTheBlockedColumnSendsReadyToTheLiveVerdict` is the sibling of `TestDegradedProjectionSendsReadyToTheLiveBdVerdict` on the same fixture: with the column unreachable, `Ready` takes the live verdict and `CachedReady` / `ReadyContext` / `Handles().Cached.Ready` all decline, while non-readiness cached reads keep serving. `TestCachedReadyOverNativeBackingNeverOffersWorkTheBackingHides` asserts the invariant itself — a cached ready read over a native backing never returns a bead the backing's own `Ready()` excludes — and also that the cache still *answers*, equal to the backing, so a trivially-declining implementation cannot pass.

**The version_compat fix is untouched** — fail-open decision, PASS-not-WARN reasoning, and all six pinned byte-identity outcomes stand as committed.

**On the live fleet at deploy.** Rig scopes logging `native_store_unavailable gate=version_compat` still move `BdStore` → `NativeDoltStore`, and that readiness is now transitively correct: cached readiness equals `bd ready`'s verdict, so a child of a blocked parent and a bead gated by a non-resident cross-store row stay hidden exactly as they do today under `BdStore` — without the 6–16 s `bd sql` subprocess. The added per-scope cost is one batched two-column `IN` read per prime and per reconcile on the store's existing connection. Scopes on `BdStore`, DoltLite, exec or mem are unaffected. Nothing is deployed or restarted by this PR.

**Gates** (worktree `/data/projects/gascity-jkvc-40qh1`, serial): `go build ./...` · `go vet ./...` · `gofmt` clean · `golangci-lint run ./...` (2.10.1) **0 issues** · `go test ./internal/beads/...` ok · `go test ./cmd/gc -timeout 25m -p 1 -parallel 1` **ok 687.324s** · `go test ./internal/...` ok · `scripts/check-core-boundary.sh` OK · `make check-split-topology-rows` OK.

Issue #5087, opened 2026-08-07
storage: no command recovers a stranded infrastructure bead on a converged city


julianknutsen:
Follow-on from the storage-binding stack (#5070/#5071) verification pass.

When a converged city has stranded infrastructure beads (a write landed in the retained work store inside the window the copy cannot fence), every boot refuses loudly and names the ids — by design. But no command recovers the strand:

- `gc storage migrate --from-work` on a marked city routes to `confirmInfraConvergence` and only re-reports the strand; the copy stage is never re-entered.
- A full re-copy would be wrong anyway: `RemovedSinceCutover` beads are legitimately absent from the binding (its own GC collects expired closed wisps and read mail), and re-copying the whole source would resurrect them.

So recovery today is manual: stop writers, copy the named beads into the binding's database, re-check with `gc storage status`. The refusal now says exactly that (see the follow-up PR raising the closed-guard floor), which is honest but operator-hostile for more than a handful of beads.

The missing piece is targeted strand recovery: under the migration guard and the `--fleet-stopped` attestation, on a marked city whose containment check names stranded ids, copy exactly the stranded set into the binding, re-verify field equality, and append them to the proven-copy manifest. The classification machinery (`classifyInfraContainmentGap`) already produces the precise set; the copy/verify/manifest stages exist and need a bounded re-entry path for a marked city.

Deliberately not done as part of the review-minors PR because it is feature work with real correctness stakes (manifest append semantics, marker interplay, GC-absence classification), not a message fix.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5187, opened 2026-08-11
fix(api,controller): read and write session-class beads through the sessions store (ga-qbqij)

julianknutsen:
Systematic sweep of every consumer of the city WORK store that is actually reading or writing CLASS-OWNED data, prompted by the `/status` session-snapshot instance (#5186). Full audit table in the linked bead; this PR carries the clear-cut, mechanically identical subset.

## Counts

- **148** non-test work-store accessor call sites enumerated across `internal/` and `cmd/gc/`
- **28** touch class-owned data through a work-store handle
- **20** behavioral mis-routes · **5** inert (store value never dereferenced) · **3** ambiguous
- **14** are silent-empty / silent-stale risks rather than latency-only
- **4** are stranded writes — a class-owned bead CREATED through a work-store handle

## What this PR fixes

23 call sites in 11 files, all pure accessor substitution: `CityBeadStore()` → `SessionsBeadStore().Store`, and `cr.cityBeadStore()` → `cr.sessionsBeadStore().Store`. No new accessor, no routing-semantics change. The four controller-refresh sites collapse into one `cr.refreshDesiredState` helper — the mirror of the `cr.buildDesiredState` sibling that already routed the same parameter correctly — so the routing has exactly one place to regress.

Class is decided by `coordclass.Classify` — the same function `cmd/gc/infra_class_migrate.go readInfraSnapshot` uses to select what migrates. Anything it calls non-Work that a work-store handle touches is either read from a store that never held it, or written where the binding will never see it.

Four sites were doing real damage on a converged split city:

| Site | Consequence |
|---|---|
| `cmd/gc/usage_compute.go:229` | Gets by session id and `SetMetadata`s the two usage markers onto session beads. The Get misses, the loop logs and continues, and **usage/billing facts silently stop being emitted for the whole fleet.** |
| `internal/api/handler_extmsg.go:138` | The member fan-out reaches `resolveSessionIDMaterializingNamedWithContext` → `handle.Create`, which mints a `type=session` bead. **Every extmsg cold-wake strands one.** |
| `internal/api/handler_beads.go:99` | Same materialize chain plus `RepairTypeBestEffort`. **Every named-assignee normalize strands one.** |
| `cmd/gc/city_runtime.go:673,684,1313,1347` | The desired-state refresh's store becomes `agentBuildParams.beadStore`, and the session-bead overlay mints a dependency floor through it. **Every controller tick on a `depends_on` city strands one — twice.** (Added in review; see below.) |

A stranded infrastructure bead is invisible to the sessions binding and is named by the per-boot containment re-check — the mechanism that made maintainer-city boot-fatal.

Also worth flagging: `handler_extmsg.go:97` sat one line above `:109`, which already used `SessionsBeadStore()`. The same function pair was reading two different stores.

### On `ga-j89yo`

Verified each of its eight sites individually rather than trusting the list. **Three were behavioral**; **five were inert** — `worker.Factory`'s transcript methods (`ReadTranscript`, `TailMeta`, `DiscoverTranscript`) route through `Factory.Adapter()` and never dereference `f.store`, and `session.NewManagerWithOptions` does no I/O at construction. The five move anyway so the next store-touching call added on those paths does not mis-route for free. The sweep found **eight more behavioral session-class sites the bead did not list**.

### Deliberately partial in `handler_beads.go`

Only the identifier resolution and the session-bead read/write move. The work-bead query and the work-bead write they feed stay on the work store — that is the correct split for a work handler that takes a session-shaped identifier, and both sites now say so in a comment.

## Review round: a fourth stranded write, and three guards that could not fail

Four majors from council review, all fixed here.

**1. `cmd/gc/city_runtime.go:673,684,1313,1347` — a stranded write the audit had stamped CORRECT.** Those sites passed `cr.cityBeadStore()` as `refreshDesiredStateWithSessionBeads`'s `store`, on the rationale that it was the work leg for assigned-work lookup. It is not: the function never calls `collectAssignedWorkBeads*`, and the parameter becomes `agentBuildParams.beadStore` (`agent_build_params.go:130`), whose every non-test use is session-class. `applySessionBeadDesiredOverlay` → `realizeDependencyFloors` → `ensureDependencyOnlyTemplate` → `selectOrCreateDependencyPoolSessionBead` → `createPoolSessionBeadWithAlias` → `CreateSessionInfo` **mints a `type=session` bead through it**. The sibling `cr.buildDesiredState` already passes `cr.sessionsBeadStore()` and its comment names the role exactly. It also recurs ~2× per tick: `reusableDependencyPoolSessionInfo` reads `bp.sessionBeads` while the create writes `bp.beadStore`, so the floor can never satisfy its own reuse check.

**2-4. The stranded-write guards did not bite.** Mutation-verified, each by reverting only its own site:

- `TestNormalizeRawBeadAssigneeReadsSessionsClass` passed a bead ID, which resolves on the first non-materializing pass, so the only arm that Creates never ran — the zero-work-store assertion was identical on both builds. Now passes the configured named session.
- `TestExtmsgSessionSelectorsReadSessionsClass` asserted `handle != ""`, which is unfalsifiable: the miss path returns `extmsgHandleLabel(selector)`, non-empty for every non-empty selector. Reverting `handler_extmsg.go:97` alone kept it green. Now stamps the handle source on the relocated bead only and asserts the value.
- `extmsgNotifyMembers` — this PR's #1 headline site — had **no** split-store coverage; reverting it left the entire `internal/api` suite green.

## Proofs

**Tests that fail without the fix** — `internal/api/session_class_routing_test.go` (8), `cmd/gc/usage_compute_class_routing_test.go` (2), `cmd/gc/refresh_desired_state_class_routing_test.go` (2). Each seeds a session bead in a *relocated* sessions store, leaves the work store empty, and asserts the handler found it. The three **stranded-write** sites additionally assert bead residency after a real create.

Every test is mutation-proven site-by-site — revert that one site, watch it fail, restore. Reds:

```
--- FAIL: TestResolveAgentSessionSubjectsReadsSessionsClass
--- FAIL: TestBeadListAssigneeTermsReadsSessionsClass
--- FAIL: TestNormalizeRawBeadAssigneeReadsSessionsClass
      STRANDED WRITE: work store holds 1 bead(s) after an assignee normalize
--- FAIL: TestExtmsgSessionSelectorsReadSessionsClass
      extmsgSessionHandleForSelector = "gc-1", want "relocated-alias"
--- FAIL: TestExtmsgNotifyMembersMaterializesIntoSessionsClass
      STRANDED WRITE: session bead "gc-3" minted in the WORK store
--- FAIL: TestMailRecipientResolutionReadsSessionsClass
--- FAIL: TestResolveAgentTranscriptReadsSessionKeyFromSessionsClass
--- FAIL: TestEmitDueComputeFactsReadsSessionsClass
--- FAIL: TestRefreshDesiredStateWritesSessionBeadsToSessionsClass
      STRANDED WRITE: work store holds 1 bead(s) after a desired-state refresh
      (first: id=gc-1 type=session class=sessions)
--- PASS: TestSessionClassRoutingIsIdentityOnSingleStoreCity        (identity proof — passes either way by design)
--- PASS: TestSessionsBeadStoreIsWorkStoreWhenNothingRelocates      (identity proof — passes either way by design)
--- PASS: TestRefreshDesiredStateIsUnchangedOnSingleStoreCity       (identity proof — passes either way by design)
```

**Single-store byte-identity** — three ways: (a) the identity tests assert `SessionsBeadStore().Store` and `cr.sessionsBeadStore().Store` are the *identical store value* the work accessors return when nothing is relocated (`resolveClassStore` returns `workStore` verbatim when `routes.storeFor` reports no relocation, `cmd/gc/class_store.go:254-262`); (b) behaviorally for the one path that writes — `TestRefreshDesiredStateIsUnchangedOnSingleStoreCity` runs the refresh on a city that relocates nothing and asserts the realized dependency floor is minted into the one city store, exactly where it landed before; (c) the full `internal/...` and `cmd/gc` suites pass unchanged.

**Fail-loud** — every fixed site keeps its `if store == nil` guard and returns the same typed error / empty result; none falls back to the work store when the sessions store is unavailable.

## Gates

`go build ./...` · `go vet ./...` · `gofmt` · `golangci-lint 2.10.1 ./...` → **0 issues** · `go test ./internal/...` green (157 packages) · `go test ./internal/api/...` green · `go test ./cmd/gc -timeout 25m` **serial** green (702s) · `scripts/check-core-boundary.sh` OK · `scripts/check-split-topology-rows.sh` OK.

## Filed, not fixed

Routing-semantics decisions, out of the stated fix scope:

- **ga-4dn4i** (P1) — convergence roots are `ClassGraph` by `Classify` but are minted and read in the work store. Stranded write, and `gc storage migrate` copies them to the graph binding while the engine keeps using the retained work-store copies → two divergent ledgers.
- **ga-nqdff** (P1) — `sourceWorkflowStores()` omits the graph store, so the sling singleton conflict guard finds no blocker on a graph-relocated city and **admits a duplicate workflow**.
- **ga-2agql** (P2) — `classStoresForID` is graph-only; `gcs-`/`gco-`/`gcn-`/`gcm-` ids 404 on by-id GET/PATCH.
- **ga-ystjm** (P2) — `readyDemandSnapshotFingerprint` omits the graph store, so graph ready-work never moves it and scale-up is missed.
- **ga-dgi1n** (P2) — `agentutil.findSessionNameByTemplate` matches `gc.session` (dot) not `gc:session`; a dead lookup that would become a live work-store session read if the typo alone were repaired.

Closes ga-qbqij. Companion to #5186 (which fixes `handler_status.go:487` on its own branch); no overlapping files.

---
- Related to #5087 — removes three write paths (extmsg member fan-out, raw-assignee normalize, and the usage-facts lane) that minted stranded session beads through the work store on a converged split city, so fewer strands are produced going forward; it adds no recovery path for beads already stranded, which is what the filed issue asks for

### DIFF
--- a/cmd/gc/api_state.go
+++ b/cmd/gc/api_state.go
@@ -2765,9 +2765,9 @@ func (cs *controllerState) Poke() {
 // WaitForSessionCommandable waits until the controller has reconciled an async
 // session create into a lifecycle state that can accept normal commands.
 func (cs *controllerState) WaitForSessionCommandable(ctx context.Context, sessionID string) (session.Info, error) {
-	store := cs.CityBeadStore()
+	store := cs.SessionsBeadStore().Store
 	if store == nil {
-		return session.Info{}, errors.New("city bead store is unavailable")
+		return session.Info{}, errors.New("session bead store is unavailable")
 	}
 	catalog, err := workerSessionCatalogWithConfig(cs.CityPath(), store, cs.SessionProvider(), cs.Config())
 	if err != nil {
--- a/cmd/gc/city_runtime.go
+++ b/cmd/gc/city_runtime.go
@@ -664,27 +664,9 @@ func (cr *CityRuntime) run(ctx context.Context) {
 		}
 		result := cr.buildDesiredState(sessionBeads, startupTrace)
 		sessionBeads = cr.loadSessionBeadSnapshot()
-		result = refreshDesiredStateWithSessionBeads(
-			result,
-			cr.cityName,
-			cr.cityPath,
-			cr.cfg,
-			cr.sp,
-			cr.cityBeadStore(),
-			sessionBeads,
-			cr.stderr,
-		)
+		result = cr.refreshDesiredState(result, sessionBeads)
 		sessionBeads = cr.syncBeadsAndUpdateIndex(result.State, sessionBeads)
-		result = refreshDesiredStateWithSessionBeads(
-			result,
-			cr.cityName,
-			cr.cityPath,
-			cr.cfg,
-			cr.sp,
-			cr.cityBeadStore(),
-			sessionBeads,
-			cr.stderr,
-		)
+		result = cr.refreshDesiredState(result, sessionBeads)
 		if ctx.Err() != nil {
 			return
 		}
@@ -1304,16 +1286,7 @@ func (cr *CityRuntime) tick(
 	sessionBeads = cr.loadSessionBeadSnapshot()
 	recordPhase(TraceSiteSessionSnapshot, "load_session_snapshot.after_demand", phaseStart, traceSessionSnapshotFields(sessionBeads))
 	phaseStart = time.Now()
-	result = refreshDesiredStateWithSessionBeads(
-		result,
-		cr.cityName,
-		cr.cityPath,
-		cr.cfg,
-		cr.sp,
-		cr.cityBeadStore(),
-		sessionBeads,
-		cr.stderr,
-	)
+	result = cr.refreshDesiredState(result, sessionBeads)
 	recordPhase(TraceSiteDesiredStateBuild, "refresh_desired_state.before_sync", phaseStart, traceDesiredStateFields(result))
 	phaseStart = time.Now()
 	_ = cr.syncBeadsAndUpdateIndex(result.State, sessionBeads)
@@ -1338,16 +1311,7 @@ func (cr *CityRuntime) tick(
 	reapStaleExtmsgParticipants(ctx, cr.sessionsBeadStore(), cr.stderr)
 	recordPhase(TraceSiteControllerTickPhase, "reap_stale_extmsg_participants", phaseStart, nil)
 	phaseStart = time.Now()
-	result = refreshDesiredStateWithSessionBeads(
-		result,
-		cr.cityName,
-		cr.cityPath,
-		cr.cfg,
-		cr.sp,
-		cr.cityBeadStore(),
-		sessionBeads,
-		cr.stderr,
-	)
+	result = cr.refreshDesiredState(result, sessionBeads)
 	recordPhase(TraceSiteDesiredStateBuild, "refresh_desired_state.after_sync", phaseStart, traceDesiredStateFields(result))
 
 	if manualReload != nil && manualReload.soft && manualReloadCompleted &&
@@ -3438,6 +3402,29 @@ func (cr *CityRuntime) buildDesiredState(sessionBeads *sessionBeadSnapshot, trac
 		return cr.buildFnWithSessionBeads(cr.cfg, cr.sp, sessionsStore.Store, unwrapWorkStores(cr.workBeadStores()), sessionBeads, trace)
 	}
 	return cr.buildFn(cr.cfg, cr.sp, sessionsStore.Store)
+}
+
+// refreshDesiredState re-applies the session-bead overlay to an already-built
+// desired state against a freshly loaded snapshot, so sessions that appeared
+// during the build are not missed.
+//
+// The store param has the same role as buildDesiredState's: it becomes
+// agentBuildParams.beadStore, which creates and updates session beads — the
+// overlay realizes a dependency floor by minting a `type=session` bead through
+// it. So it is the SESSIONS store, not the work store; on a relocated city the
+// work store would take a session-class create the class binding never sees and
+// the boot containment re-check names.
+func (cr *CityRuntime) refreshDesiredState(result DesiredStateResult, sessionBeads *sessionBeadSnapshot) DesiredStateResult {
+	return refreshDesiredStateWithSessionBeads(
+		result,
+		cr.cityName,
+		cr.cityPath,
+		cr.cfg,
+		cr.sp,
+		cr.sessionsBeadStore().Store,
+		sessionBeads,
+		cr.stderr,
+	)
 }
 
 func (cr *CityRuntime) loadDemandSnapshot(
--- a/cmd/gc/refresh_desired_state_class_routing_test.go
+++ b/cmd/gc/refresh_desired_state_class_routing_test.go
@@ -1,0 +1,174 @@
+package main
+
+import (
+	"io"
+	"testing"
+	"time"
+
+	"github.com/gastownhall/gascity/internal/beads"
+	"github.com/gastownhall/gascity/internal/config"
+	"github.com/gastownhall/gascity/internal/coordclass"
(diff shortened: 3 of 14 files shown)


julianknutsen:
## Review round applied — 4 majors fixed in `7bdcc75`

**1. A fourth stranded write the audit had cleared.** `cmd/gc/city_runtime.go:673,684,1313,1347` passed `cr.cityBeadStore()` as `refreshDesiredStateWithSessionBeads`'s `store`. Section C had stamped those CORRECT ("the work leg for assigned-work lookup") and section D claimed the write sweep was exhaustive at three. Both were wrong: the function never calls `collectAssignedWorkBeads*`, the parameter becomes `agentBuildParams.beadStore`, and the overlay mints a `type=session` bead through it (`realizeDependencyFloors` → `ensureDependencyOnlyTemplate` → `selectOrCreateDependencyPoolSessionBead` → `createPoolSessionBeadWithAlias` → `CreateSessionInfo`). The sibling `cr.buildDesiredState` already routed the same parameter through `cr.sessionsBeadStore()` — and its comment says exactly why. It also recurs ~2× per tick: the reuse check reads `bp.sessionBeads` while the create writes `bp.beadStore`, so the floor can never satisfy itself.

The four identical call blocks now collapse into `cr.refreshDesiredState`, the mirror of `cr.buildDesiredState`, so the routing has one place to regress instead of four.

**2-4. Three guards that could not fail.** All mutation-verified by reverting only the site under test:

| Guard | Was | Red-before after repair |
|---|---|---|
| `TestNormalizeRawBeadAssigneeReadsSessionsClass` | passed a bead ID, which resolves on the non-materializing pass — the only arm that Creates never ran | `STRANDED WRITE: work store holds 1 bead(s) after an assignee normalize ... [gc-1 myrig/worker open session ... [gc:session template:myrig/worker]]` |
| `TestExtmsgSessionSelectorsReadSessionsClass` | asserted `handle != ""`, unfalsifiable — the miss path returns `extmsgHandleLabel(selector)`, non-empty for any non-empty selector | `extmsgSessionHandleForSelector = "gc-1", want "relocated-alias"` |
| `extmsgNotifyMembers` (this PR's #1 headline site) | **no** split-store test at all; reverting it left the whole `internal/api` suite green | `STRANDED WRITE: session bead "gc-3" minted in the WORK store` |
| `cr.refreshDesiredState` (new) | — | `STRANDED WRITE: work store holds 1 bead(s) after a desired-state refresh (first: id=gc-1 type=session class=sessions)` |

**Single-store byte-identity re-proved after the controller change.** `TestRefreshDesiredStateIsUnchangedOnSingleStoreCity` runs the refresh on a city that relocates nothing and asserts the realized dependency floor is minted into the one city store — the same place it landed before — alongside the accessor-identity assertions and the unchanged full suites.

**Gates:** `go build ./...` · `go vet ./...` · `gofmt` clean · `golangci-lint 2.10.1 ./...` → 0 issues · `go test ./internal/...` green (157 pkgs) · `go test ./internal/api/...` green · `go test ./cmd/gc -timeout 25m` serial green (702s) · `scripts/check-core-boundary.sh` OK · `scripts/check-split-topology-rows.sh` OK.

The audit doc's sections C and D are corrected: the four sites are reclassified **MIS + WRITE** (row #25), and section D's "no other `Create` ... reaches a work-store handle" claim is retracted with the corrected sweep result. The sweep that produced it stopped at `Create(` call sites in the handler layer instead of following `agentBuildParams.beadStore` into the overlay; the durable fix for that class of miss is the residency guards now in the suite, not a wider grep.

The 19 original fixes are untouched — this round is additive.

Issue #4246, opened 2026-07-14
BdStore fallback does full unbounded list scans on every poll tick


figgeous:
## Problem

Even accepting that the BdStore fallback exists (fork a `bd` CLI process per store operation), the polling paths make it far more expensive than it needs to be: every nudge-poller tick and supervisor reconcile re-runs full, unbounded list queries (`--limit 0` / `--all`) over the entire issue store, with no cache, cursor, or change detection. Cost scales as sessions × tick rate × store size.

## Impact

On a 7-session city with ~2.8k issues, steady state was ~27 `bd` forks/second, each running at least one full issue-list query with five JSON-aggregating joins — sustaining 1–4 cores of Dolt CPU indefinitely. The work is almost entirely redundant: between ticks nothing has usually changed.

## Expected vs actual

- **Expected:** poll ticks do bounded, targeted reads (changed-since cursor, or a shared/short-TTL cache across pollers), so an idle city costs near-zero DB work.
- **Actual:** each tick per session forks fresh `bd` processes running `bd list --json --limit 0`, `bd query --json ephemeral=true --limit 0`, `bd show --json <target>`; the supervisor adds its own `--all` order-tracking scans per reconcile.

## Evidence

Sampling live process spawns over 5s (100 samples): every poller and the supervisor had a `bd` child alive in 17–31% of samples. Dolt connection IDs advanced ~80/s (3 connections per bd invocation, measured via strace).

---

<details>
<summary>Machine detail</summary>

**Hot paths**
- `cmd/gc/cmd_nudge.go:603` (`cmdNudgePoll`) — 2s default interval (`defaultNudgePollInterval`, line 53); each tick calls `nudgeObserveTarget` + `tryDeliverQueuedNudgesByPoller`, which on the BdStore translate to multiple full CLI list/query invocations.
- Supervisor order machinery — observed commands: `bd list --json --label=order-tracking --all --include-infra --include-gates --include-templates`, `bd query --json "ephemeral=true AND label=order:nudge-on-route" --all --limit 0`, `bd list --json --label=order:cascade-nudge-on-blocker-close --all ...`, plus periodic `bd create ... order:gate-sweep`.
- `internal/beads/bdstore.go:451` — one `exec.CommandContext("bd", ...)` per store op; each bd invocation opens 3 SQL connections.

**Cost model (measured)**
- 7 pollers × (2-4 bd calls / 2s) + supervisor loops ≈ 27 forks/s ≈ 80 conns/s.
[...]

PR #5189, opened 2026-08-11
fix(beads): let a BdStore cache serve the complete ready projection (ga-tgpfm)

julianknutsen:
Fixes ga-tgpfm (P0) **and ga-cfhgr (P0)** — the second is a pre-existing
`CachingStore` defect that is on `main` today and shipping on the five
Dolt-backed cities right now. This PR no longer only unblocks
maintainer-city's `/status`; it closes the systemic hole that made the
first fix unsafe. See **"The pre-existing defect this also fixes"** below.

## The defect

`CachingStore.cachedReadyCompleteOnly` gates on `c.depsComplete`, which
comes from `fetchDepsForBeads`. `BdStore` implements no
`dependencySnapshotForCache`, so the cache took the
`listDependencyCompletenessStore` branch and adopted
`BdStore.listIncludesCompleteDependencies()` — a hardcoded `return
false`. `NativeDoltStore` returns true; `DoltliteReadStore` implements
the snapshot properly. BdStore was the only store that could never
satisfy a complete-ready read, so every scope on one answered

    city work ready: reading complete ready projection from cache: bead cache unavailable

forever and sent each readiness read to a live `bd ready`.

maintainer-city's **city** scope is pinned to BdStore by the
`metadata_backend` preflight gate (the native store serves Dolt only;
MC's work class is hosted Postgres), so it cannot escape it. It is also
the amplifier that turned the ga-40qh1 version_compat false-fail into a
fleet-wide outage: every rig that fell to BdStore inherited a
permanently-unavailable cache.

## Why this is not `dependencySnapshotForCache` over `DepListBatch`

That was the obvious shape, and it does not work on the one city the
defect is about. `DepListBatch` is `bd dep list`, and the bd/Postgres
backend answers it with `operation "IssueRelations" not supported by the
postgres backend` — measured on maintainer-city, tracked as ga-7i7ts,
and the reason `cmd/gc/infra_class_recover.go` already carries its own
dependency reader. On MC it would have bought a guaranteed-failing
~4.4 s subprocess per prime and per reconcile and left `depsComplete`
false: a worse stall than the one it replaced.

## The fix

bd already carries every row's "down" edges on the row.
`sqlbuild.SearchCountsSQL` projects `d.deps_json`
(`JSON_ARRAYAGG` over `tables.Dependencies`) beside `dc.cnt`
(`COUNT(*)` over that same table `WHERE type = 'blocks'`), in every form
of the query, for the issues family and the wisps family alike;
`ScanReadyWorkRowWithCounts` unmarshals it onto `types.Issue`. Both
subcommands this store lists through — `bd list` and the `bd query` it
uses for the wisp tier — go through that builder, and
`bdIssue.normalizedDependencies` already carries the result onto
`Bead.Dependencies`, which `depsFromBeadFields` already feeds to the
cache. The rows were complete; only the verdict was wrong.

So the verdict is now the WITNESSED one, read off rows bd already
returned: **zero extra subprocesses, for any N.**

The witness is falsifiable rather than assumed, because a flat `return
true` is a real hazard — an adapter that never populated the field would
answer "no edges" for every bead, and a cache that believed it would
flatten the topology into "everything is ready". bd hands back its own
count of each row's blocking edges from the same query, so the two
columns are two views of one read:

- a row whose `dependency_count` exceeds the blocking edges it carried is
  proof the projection is short — one such row disqualifies the ledger,
  one-way;
- a ledger where no row ever carried an edge proves nothing, so it keeps
  today's behavior: correct, and merely slower.

## The correctness gate this opens, and how it is closed

Completeness of the dep SET is not the claim that readiness equals bd's.
bd's `is_blocked` propagates transitively down parent-child edges, which
no down-dep set reproduces, so the cache's answer equals bd's only while
`enrichReadyProjectionForCache` fills that column. BdStore already
serves it and already reports `ErrReadyProjectionUnsupported` when it
cannot — except on one route. A bd predating 1.0.5 returned
`(false, nil)`: no error, so no degrade, so every `IsBlocked` nil and
every readiness handle answered by the dependency-derived predicate. That
was harmless only while `depsComplete` was hardcoded false. It is now the
#3218 regression, so the version gate names the degrade like the backend
and runtime refusals do, and readiness declines the cache. Mutating that
one branch back reproduces the exact shape the #5184 council caught:

    Ready = [bd-blocker bd-child bd-unrelated bd-xstore], want [bd-blocker bd-unrelated]

Unlike the other two refusals it is memoized on the store, not latched in
the scope guard: which bd is on PATH is a property of the process, not of
the ledger in that directory, and the guard is never cleared.

## Verification

`TestBdBackedCachedReadyNeverOffersWorkTheBackingHides` extends #5183/
#5184's invariant — a cached ready read may never return a bead the
backing's own `Ready()` excludes — to the BdStore path, over their
fixture plus the two shapes the direct-dep predicate cannot see:

    blocker <-- blocks -- parent <-- parent-child -- child
    gcg-offscope (not resident) <-- blocks -- xstore
    parent <-- parent-child -- wisp-step (ephemeral)

Red before the fix:

    ReadyContext error = reading complete ready projection from cache: bead cache unavailable, want rows served from cache

Every added test is mutation-proved: restoring the hardcoded `false`,
flipping it to an unconditional `true`, dropping either list call site,
deleting the contradiction latch, renaming the `dependency_count` key,
and reverting the version gate each turn a test red, and each was
restored.

Three existing tests pinned `CachedReady` answering from the
dependency-derived predicate under bd 1.0.4 — the hole above. They are
corrected to assert the dep map directly, which is what they were about,
plus the readiness verdict that is now correct.

For the BdStore half of this PR, the other stores are untouched by
construction: the store-level change is confined to `bdstore*.go`, so the
seam dispatch is identical, and 555 verbose test outcomes across
NativeDoltStore / DoltLite / MemStore / FileStore / exec were identical
before and after. The comparison is sensitive — breaking
`NativeDoltStore.enrichReadyProjectionForCache` diverges it by three
failures.

That byte-identity claim is deliberately **no longer true of the PR as a
whole**: the ga-cfhgr commit below does change `caching_store.go`, and it
changes it *for every store*, because the defect it fixes is every
store's. What that costs the other paths is stated and pinned there.

## The pre-existing defect this also fixes (ga-cfhgr)

A council review of the first commit found that the completeness verdict
alone is not enough, and that the reason is not this PR's logic.

`is_blocked` is **never on bd's wire.** There is no `json:"is_blocked"`
tag anywhere in beads `v1.1.1-0.20260805093327`; `bd show --json` answers
from `types.IssueDetails` (labels / dependencies / dependency_count), and
`beadFromNativeIssue` cannot set the field either. The column exists in
the cache ONLY because prime and reconcile ran a separate projection
query. But `absorbFreshLocked` installed every fresh row
unconditionally, so three routine paths put `IsBlocked == nil` back over
a projected row and latched no degrade:

1. `CachingStore.Update`'s success path, absorbing a `backing.Get` row;
2. `readCacheWithOverlay`'s dirty-row refresh, the same shape;
3. `clearDependentReadyProjectionsLocked`, which nils a changed bead's
   direct dependents *on purpose* so the direct-dep predicate recomputes.

With `depsComplete` true, `cachedBeadReady` then answers from that
predicate — the one this package documents as weaker than bd's column —
and `readyProjectionDegraded` is store-wide and one-shot, so nothing
degrades.

**This ships today.** On the repo's own unmodified
`nativeReadyDisagreementFixture`, with no code from this PR involved:

    native Ready              = [blocker unrelated]
    after Update(child, Assignee) -> cache Ready = [blocker child unrelated]
    after Update(xstore, ...)     -> cache Ready = [blocker child unrelated xstore]

reachable in production through `cmd/gc/api_state.go` `StartReconciler`
serving `internal/api/handler_status.go`'s `ReadyContext`. The first
commit does not create this; it moves BdStore scopes into the regime
where it fires, from "decline → live → correct" to "answer → sometimes
wrong".

### The shape chosen, and why

A hybrid, decided per row rather than per store:

- **Preserve where it is provable.** `absorbFreshLocked` keeps the cached
  verdict when the row's dependency set is unchanged — the same evidence
  `preserveCachedReadyProjectionLocked` already requires on the reconcile
  path, now applied at the ONE choke point every absorb site in the
  package goes through, so there is no site to miss. This covers the
  write-through and overlay refreshes at **zero cost**: they change
  nothing readiness depends on, so the cache keeps answering.
- **Decline where it is not.** A verdict that cannot be preserved is
  recorded as LOST, and readiness declines that row to the live backing
  instead of guessing — unless the row's own edges can reproduce the
  verdict exactly (every edge ready-blocking, every target resident),
  which is the ordinary case and keeps the common close answering from
  cache.

Deliberately **not** the blunt "decline whenever `IsBlocked` is nil". nil
is the normal state for a backing with no projection at all (MemStore,
DoltLite) and for the rows every projection skips by design — message
rows, and `gc:nudge` shadow beads, which are `type=chore` and therefore
ARE ready candidates. One open nudge bead would have declined every
readiness read on that scope forever: the multi-second live-scan stall
this PR exists to remove, rebuilt.

Reconciliation opts out of the preserve: it has already decided per row,
on the blocking targets' *fresh* statuses, which verdicts survive the
cycle, and re-preserving at absorb time would override a considered
refusal.

### Verification

`TestBdBackedCachedReadyNeverOffersWorkTheBackingHides` and
`TestCachedReadyOverNativeBackingNeverOffersWorkTheBackingHides` now
re-assert the subset invariant in the states a long-lived cache actually
lives in — after an `Update`, after a dirty-overlay `Ready()`, and after
a `Close` — on **both** store paths. The native one is the proof that
this fixes `main`'s bug rather than this PR's exposure of it. Red before,
per stage, on all four readiness handles; e.g.

    Ready (after Update(child)) offered gc-3, which the backing's own Ready hides:
    blocked-ness propagates down parent-child and the cache's direct-dep
    predicate cannot see it (backing = [gc-1 gc-5], cache = [gc-1 gc-3 gc-5])

    Ready (after Close(bd-x)) offered bd-both, which the backing's own Ready hides:
    its parent bd-parent is still blocked by the open bd-gate, and bd propagates
    that down the parent-child edge (backing = [bd-gate], cache = [bd-both bd-gate])

Both tests also assert the cache **still answers** after each stage, so a
store that declined its way to a vacuous subset pass fails. The close
subtests additionally pin the cost honestly: the row that genuinely
cannot be answered declines, and one reconcile restores the answer.

The merge differential/fuzz oracle models the new per-row state
(`mergeEndState.readyLost`) and re-derives it independently from the
input state, so the seam cannot drift; the field-coverage census
classifies it as compared, not excluded.

### Side finding from ga-tgpfm, now closed

`DoltliteReadStore.dependencySnapshotForCache`'s completeness bool was
unguarded — flipping it to `false` left all 555 outcomes green.
`TestDoltliteDependencySnapshotCompletenessIsGuarded` (in the CI-run
`^TestDoltlite` suite) now fails on exactly that mutation.

## What changes on the live fleet at deploy

maintainer-city's city scope stops reporting `bead cache unavailable` and
starts answering readiness from cache, retiring one ~4.4 s `bd ready` per
controller tick; any rig on BdStore does the same.

The Dolt cities are **no longer unaffected**, and that is the point of the
second commit: they stop offering transitively blocked work between
reconciles. Their new cost is bounded and paid only where the cache
provably cannot answer — a row whose verdict was dropped AND whose own
edges cannot reproduce it (a parent-child edge, or an edge onto a
non-resident row). Everything else keeps serving from cache, including the
ordinary "blocker closed, dependent unblocked" close that
`TestCachingStoreApplyCloseEventClearsDependentProjectedIsBlocked` and five
sibling tests pin.

---
- Related to #4246 — lets a BdStore-backed cache serve the complete ready projection, so readiness reads on those scopes stop forking a live `bd ready` on every controller tick — one instance of the per-tick redundant BdStore reads that issue is about; it does not touch the nudge-poller or supervisor `--limit 0` / `--all` list, query and show scans, and adds no changed-since guard, batching or idle back-off, so the unbounded per-tick scans themselves are untouched

### DIFF
--- a/cmd/gc/infra_class_recover.go
+++ b/cmd/gc/infra_class_recover.go
@@ -258,10 +258,10 @@ type strandedBeadDump struct {
 // adapter: it stays true for every bead and every retry, so one probe answers
 // for the store. `database is locked`, a dropped connection, a subprocess that
 // could not fork are facts about a moment, and a moment is no evidence about a
-// capability. Reading the second as the first swaps a trustworthy read for one
-// this codebase itself marks incomplete — BdStore.listIncludesCompleteDependencies
-// returns false, which is why CachingStore refuses to serve down-deps from the
-// inline projection — and it does so for the whole run, under a zero exit code.
+// capability. Reading the second as the first swaps the relation read for the
+// inline projection unconditionally — a projection this reader only trusts once
+// it has WITNESSED it live below — and it does so for the whole run, under a
+// zero exit code.
 //
 // Text matching is what is available: the refusal crosses the adapter boundary
 // as a string and neither side carries a sentinel for it. It degrades in the
--- a/cmd/gc/infra_class_recover_topology_test.go
+++ b/cmd/gc/infra_class_recover_topology_test.go
@@ -465,9 +465,8 @@ func (s transientDepListSource) List(query beads.ListQuery) ([]beads.Bead, error
 // bd/Postgres backend answers `operation "IssueRelations" not supported by the
 // postgres backend`, and that is a fact about the adapter. A locked database, a
 // lost connection or a subprocess that failed to fork is a fact about a moment.
-// Reading the second as the first swaps a trustworthy read for one the codebase
-// itself marks incomplete (BdStore.listIncludesCompleteDependencies is false),
-// and does it for the whole run.
+// Reading the second as the first swaps the relation read for a projection this
+// reader has not witnessed, and does it for the whole run.
 //
 // Red-before, on the reader that latched on ANY probe error:
 //
--- a/internal/beads/bdstore.go
+++ b/internal/beads/bdstore.go
@@ -373,6 +373,11 @@ type BdStore struct {
 	readyProjectionMu      sync.Mutex
 	readyProjectionChecked bool
 	readyProjectionEnabled bool
+	// readyProjectionVersionErr memoizes the ErrReadyProjectionUnsupported this
+	// store owes every later caller when the bd on PATH predates the is_blocked
+	// projection. Store-local rather than scope-latched: the bd binary is a
+	// property of the process, not of the ledger. See bdstore_ready_projection.go.
+	readyProjectionVersionErr error
 	// readyProjectionScope memoizes, per SCOPE PATH, the verdict that this
 	// ledger cannot serve the ready projection at all. Shared with every other
 	// store rooted at the same directory, because cmd/gc rebuilds a store per
@@ -412,6 +417,12 @@ type BdStore struct {
 	unreadStore *unreadStoreGuard
 	// noticeSink redirects operator notices away from stderr; nil is stderr.
 	noticeSink io.Writer
+
+	// inlineDeps records whether bd's list JSON has been seen to carry each
+	// row's dependency edges beside it, which is what lets a CachingStore over
+	// this store serve down-deps and complete-ready reads from the snapshot.
+	// See bdstore_inline_deps.go.
+	inlineDeps inlineDependencyProjection
 
 	localStrings *localSidecar // clone-local data; see Store.SetLocalString
 }
@@ -525,10 +536,6 @@ func (s *BdStore) Dir() string {
 // label hydration.
 func (s *BdStore) ListSkipLabelsEnabled() bool {
 	return s != nil && s.listSkipLabelsEnabled
-}
-
-func (s *BdStore) listIncludesCompleteDependencies() bool {
-	return false
 }
 
 // Init initializes a beads database via bd init --server. This is an admin
@@ -773,10 +780,16 @@ type bdIssue struct {
 	Labels       []string     `json:"labels"`
 	Metadata     StringMap    `json:"metadata,omitempty"`
 	Dependencies []bdIssueDep `json:"dependencies,omitempty"`
-	Ephemeral    bool         `json:"ephemeral,omitempty"`
-	NoHistory    bool         `json:"no_history,omitempty"`
-	DeferUntil   *time.Time   `json:"defer_until,omitempty"`
-	IsBlocked    optionalBool `json:"is_blocked,omitempty"`
+	// DependencyCount is bd's own count of this row's BLOCKING edges,
+	// projected from the same dependency table and the same query as
+	// Dependencies. It is the control that turns "bd carried edges inline"
+	// into a falsifiable claim — see bdstore_inline_deps.go. A pointer so a bd
+	// that omits the field stays distinguishable from one reporting zero.
+	DependencyCount *int         `json:"dependency_count,omitempty"`
+	Ephemeral       bool         `json:"ephemeral,omitempty"`
+	NoHistory       bool         `json:"no_history,omitempty"`
+	DeferUntil      *time.Time   `json:"defer_until,omitempty"`
+	IsBlocked       optionalBool `json:"is_blocked,omitempty"`
 	// Revision carries bd's optimistic-concurrency token for ConditionalWriter.
 	// Pre-#4682 bd omits it, so it decodes to 0; toBead stamps it onto the
 	// otherwise json:"-" Bead.Revision field. The "revision" key is provisional:
@@ -2532,6 +2545,7 @@ func (s *BdStore) listViaBDList(query ListQuery) ([]Bead, error) {
 	for i := range issues {
 		result[i] = issues[i].toBead()
 	}
+	s.noteInlineDependencyProjection(issues, result)
 	filtered := applyListQuery(result, query)
 	if parseErr != nil {
 		if len(filtered) == 0 {
@@ -2644,6 +2658,7 @@ func (s *BdStore) listEphemeral(query ListQuery) ([]Bead, error) {
 		result[i].Ephemeral = true
 		result[i].NoHistory = false
 	}
+	s.noteInlineDependencyProjection(issues, result)
 	filtered := applyListQuery(result, query)
 	if parseErr != nil {
 		if len(filtered) > 0 {
--- a/internal/beads/bdstore_inline_deps.go
+++ b/internal/beads/bdstore_inline_deps.go
@@ -1,0 +1,115 @@
+package beads
+
+import "sync/atomic"
+
+// inlineDependencyProjection is what a BdStore has learned, from bd's own
+// answers, about the dependency rows bd's list JSON carries beside each issue.
+//
+// bd projects those rows from the same table, in the same query, as the
+// dependency_count it puts on every row: sqlbuild.SearchCountsSQL selects
+// `d.deps_json = JSON_ARRAYAGG(...)` over tables.Dependencies alongside
+// `dc.cnt = COUNT(*) ... WHERE type = 'blocks'` over that same table, for the
+// issues family and the wisps family alike, and ScanReadyWorkRowWithCounts
+// unmarshals deps_json onto types.Issue.Dependencies. Both `bd list` and the
+// `bd query` this store uses for the wisp tier go through that one builder.
+//
+// So the count and the rows are two views of one read, and a disagreement
+// between them is the only way the inline projection can be short. That makes
+// the completeness claim MEASURABLE per listing rather than assumed from a
+// reading of bd's source — which matters, because an adapter that populated the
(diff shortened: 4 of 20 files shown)


julianknutsen:
## Second commit: fixes a pre-existing `main` defect, not just this PR's exposure of it

The council review of `59388837ab` confirmed the completeness verdict is right and cheap, and found that it lands on top of a **systemic hole in `CachingStore` that is on `main` today and shipping on the five Dolt-backed cities.** Filed as **ga-cfhgr (P0)**; fixed here in `31ed1f3911`.

### The hole

`is_blocked` is never on bd's wire — no `json:"is_blocked"` tag anywhere in beads `v1.1.1-0.20260805093327`, and `beadFromNativeIssue` cannot set it either. The column is in the cache only because prime/reconcile ran a separate projection query. `absorbFreshLocked` installed every fresh row unconditionally, so three routine paths put `IsBlocked == nil` back over a projected row and latched no degrade: `Update`'s write-through refresh, the dirty-row overlay refresh, and `clearDependentReadyProjectionsLocked` (which nils direct dependents on purpose).

Measured on the repo's **own unmodified** `nativeReadyDisagreementFixture`, no code from this PR involved — which is what makes it main's bug and not this one's:

```
native Ready                   = [blocker unrelated]
after Update(child, Assignee)  = [blocker child unrelated]
after Update(xstore, Assignee) = [blocker child unrelated xstore]
```

### The shape chosen

Hybrid, per row, not per store:

* **Preserve where provable** — `absorbFreshLocked` keeps the cached verdict when the row's dep set is unchanged, the same evidence `preserveCachedReadyProjectionLocked` already requires on the reconcile path, applied at the one choke point every absorb site goes through (so there is no site to miss). Zero cost: the refresh changed nothing readiness depends on, so the cache keeps answering.
* **Decline where not** — a verdict that cannot be preserved is recorded LOST and readiness declines that row to the live backing, unless the row's own edges reproduce the verdict exactly (every edge ready-blocking, every target resident).

Not the blunt "decline whenever `IsBlocked` is nil": nil is normal for backings with no projection (MemStore, DoltLite) and for the rows the projection skips by design — message rows and `gc:nudge` shadow beads, which are `type=chore` and **are** ready candidates. One open nudge bead would have declined every readiness read on that scope forever, rebuilding the exact stall this PR removes.

### Pinned in the states the cache actually lives in

`TestBdBackedCachedReadyNeverOffersWorkTheBackingHides` and `TestCachedReadyOverNativeBackingNeverOffersWorkTheBackingHides` now re-assert the subset invariant after an `Update`, after a dirty-overlay `Ready()`, and after a `Close`, on both store paths and all four readiness handles. Red before, e.g.:

```
Ready (after Update(child)) offered gc-3, which the backing's own Ready hides:
blocked-ness propagates down parent-child and the cache's direct-dep predicate
cannot see it (backing = [gc-1 gc-5], cache = [gc-1 gc-3 gc-5])

Ready (after Close(bd-x)) offered bd-both, which the backing's own Ready hides:
its parent bd-parent is still blocked by the open bd-gate, and bd propagates that
down the parent-child edge (backing = [bd-gate], cache = [bd-both bd-gate])
```

Both also assert the cache **still answers** at each stage — declining everything would satisfy a subset check vacuously and cost MC the live read. The close subtests state the cost plainly: the row that genuinely cannot be answered declines, and one reconcile restores the answer.

The merge differential/fuzz oracle models the new per-row state (`mergeEndState.readyLost`) and re-derives it independently, and the field-coverage census classifies it as compared.

### ga-tgpfm's open side finding, closed

`DoltliteReadStore.dependencySnapshotForCache`'s completeness bool was unguarded (flipping it to `false` left all 555 outcomes green). `TestDoltliteDependencySnapshotCompletenessIsGuarded`, in the CI-run `^TestDoltlite` suite, now fails on exactly that mutation.

### Gates

`go build ./...` · `go vet ./...` · `gofmt` clean · golangci-lint 2.10.1 **0 issues** · `go test ./internal/beads/...` · `go test ./internal/...` (144 pkg) · `go test ./cmd/gc` serial (`-p 1 -parallel 1`, 995 s) · `go test -tags gascity_native_beads -run ^TestDoltlite ./internal/beads` · `scripts/check-core-boundary.sh` · `make check-split-topology-rows` — all green.

Not merged.

PR #5190, opened 2026-08-11
fix(cmd/gc): discover worktree liveness independent of the gc-owned root (ga-1xaqgo.1)

quad341:
Fixes ga-1xaqgo.1 — cause A of the ga-1xaqgo epic ("gc cannot see work in
Claude-Code EnterWorktree worktrees").

## The bug

`reapClosedBeadWorktrees` filtered a rig's worktree list down to
`.gc/worktrees/<rig>/` **before** liveness was ever computed. A worktree
created outside that convention — e.g. under `.claude/worktrees/`, exactly
what Claude Code's own `EnterWorktree` tool creates — was invisible to the
liveness gate entirely, not merely excluded from reap eligibility. Scope
restriction and liveness discovery were conflated in one loop.

## The fix

`discoverWorktreeLiveness` (`cmd/gc/bead_worktree_liveness.go`) enumerates
every worktree git knows about for a rig's repository (including the main
checkout) and reports liveness for each, with no root-convention or
bead/reap-eligibility filtering — scope decisions stay entirely with the
caller. It mirrors `worktreeIsLive`'s existing fail-closed contract: when
the liveness scan is indeterminate (`scanned == false`), every result
reports `Live=false, Reason=""` rather than guessing, and the caller remains
responsible for treating that as protect-everything.

`reapClosedBeadWorktrees` now sources its candidate list from this function
and reuses the already-computed liveness instead of re-scanning per
candidate. Its own scope restriction to gc-owned paths is **unchanged** —
this PR separates "which worktrees exist and are they live" from "which of
those this reaper may act on," it doesn't touch the latter.

The result is a plain `worktreeLiveness{Path, Branch, Live, Reason}` struct,
not an interface — satisfying the bead's "reusable by downstream capacity
accounting without a speculative interface" requirement for the blocked
child `ga-1xaqgo.3` (reconciler capacity accounting for externally created
worktrees).

## Testing

- RED regression (`TestDiscoverWorktreeLiveness_ReportsWorktreeOutsideGcOwnedRoot`)
  confirmed the omission first: a live process cwd inside a worktree under
  `.claude/worktrees` was invisible pre-fix.
- Three more discovery-layer tests cover the main-checkout inclusion,
  fail-closed behavior on an indeterminate scan, and an invalid-rig-root
  error path.
- A new reaper-layer regression
  (`TestReapClosedBeadWorktrees_NeverActsOnWorktreeOutsideGcOwnedRoot`) locks
  in that a foreign worktree is never reaped or even reported, keeping the
  "never delete/kill/auto-terminate a discovered foreign worktree" guarantee
  covered at the layer that actually deletes.
- All pre-existing reaper tests (49) still pass unchanged.
- `go vet ./...` clean, `gofmt -l` clean on all changed files.
- `make test-fast-parallel` passed (push-gate gated).

### DIFF
--- a/cmd/gc/bead_worktree_liveness.go
+++ b/cmd/gc/bead_worktree_liveness.go
@@ -1,11 +1,13 @@
 package main
 
 import (
+	"fmt"
 	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
 
+	"github.com/gastownhall/gascity/internal/git"
 	"github.com/gastownhall/gascity/internal/pathutil"
 	sessionpkg "github.com/gastownhall/gascity/internal/session"
 )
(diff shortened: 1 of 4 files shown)


quad341:
@quad341 — this is the right cut. Conflating "which worktrees exist and are they live" with "which of those may I reap" in a single loop reads fine right up until a worktree appears from outside your convention and drops out of the gate entirely, and the fix that separates the two while leaving the reaper's `PathWithin` restriction completely untouched is the version that doesn't quietly widen what a deletion path can reach.

The second regression test is what makes this mergeable rather than merely plausible. It would have been easy to stop at "the new function sees the foreign worktree" — adding `TestReapClosedBeadWorktrees_NeverActsOnWorktreeOutsideGcOwnedRoot` at the layer that actually calls `WorktreeRemove`, asserting the tree is neither reaped *nor* reported protected, is the part that proves widened visibility didn't buy widened authority.

Merging as-is. Two notes for whoever picks up `ga-1xaqgo.3`: the pass-2 lookup now reads a map miss as "not live," which is fail-open — unreachable today since candidates share the map's source slice, but worth a guard if those lists ever diverge. And the CodeQL red on this PR is two `note`-severity advisories on `go.mod`, which you don't touch; that's a dependency bump's problem, not yours.


---

Issue #5188, opened 2026-08-11
design: share per-agent worktree enforcement implementations


wbern:
## Problem to solve

There appear to be several independent implementations of "make every coding agent work in its own worktree": pack scripts, shell wrappers, provider hooks, Git hooks, task/worktree registries, and OS or container sandboxes.

They do not all provide the same guarantee. In particular, these are separate concerns:

1. **Provisioning** - create a unique worktree and launch the session there.
2. **Ownership and affinity** - bind a task/session to one canonical worktree and recover that binding after a restart or handoff.
3. **Workflow validation** - refuse selected lifecycle stages when cwd, branch, task metadata, or worktree identity disagree.
4. **Git guardrails** - prevent a commit, push, or checkout from the wrong tree or branch.
5. **Write confinement** - prevent direct tools, absolute paths, shell redirects, subprocesses, symlinks, or `git -C` from mutating another checkout.
6. **Detection and recovery** - find shared, mismatched, unmanaged, dirty, or stale worktrees and clean them up safely.

Starting a session in the right cwd does not by itself stop an agent from writing to an absolute path in the canonical checkout. Conversely, a Git hook may catch a bad commit while still allowing the earlier file contamination.

There is no single place today where users can compare what they actually run, what boundary it enforces, and what its known bypasses are.

## Proposed change

Use this issue as an implementation inventory and design input for a suitable Gas City direction.

If you have a setup in a gist, fork, repository, pack, hook, wrapper, launcher, sandbox policy, or container configuration, please share it here. Working implementations and failure reports are both useful.

For each setup, please include as much of this as practical:

- Link to the gist, fork, repository, pack, or relevant files.
- Agent/provider, version, operating system, and shell.
- Which guarantees above it provides: provisioning, ownership, validation, Git guardrails, confinement, or detection.
- How the session is durably bound to a canonical real path, repository identity, branch/ref, and task.
- Which write-capable surfaces it covers: file tools, patch tools, shell, interpreters, redirects, subprocesses, MCP tools, etc.
- What happens for `cd`, `git -C`, `GIT_DIR`, `..`, sibling worktrees, absolute paths, symlinks, and nonexistent targets.
- Whether it fails closed, warns, repairs, or only audits.
- Escape hatches and known false positives or bypasses.
- How linked-worktree Git metadata in the shared `.git/worktrees/...` directory is handled.
- Behavior across resume, handoff, restart, stale/dirty trees, crashes, and cleanup.
[...]

PR #3958, opened 2026-07-06
feat(session): provision git worktrees for per-bead work dirs

sjarmak:
## Summary

Provisions real git worktrees for per-bead work dirs under the worktrees root, so
a routed bead's work dir is an isolated checkout instead of a bare path inside the
shared rig tree. This closes the class of failure where a work dir points at a
never-provisioned path and workers end up editing the shared main checkout
concurrently.

## What changed

- `internal/git`: new `WorktreeAdd` that provisions a pre-seeded worktree dir
  (detached checkout at a base ref), with unit coverage in `worktree_add_test.go`.
- `cmd/gc/bead_worktree_provision.go`: provisions a bead's worktree during desired-
  state build; `build_desired_state.go` and `session_lifecycle_parallel.go` wire it
  into the session lifecycle so a multi-slot pool named session that receives one
  routed bead provisions the worker's isolated tree.
- `bead_worktree_reaper.go`: teardown path updated to match the new provisioning.

## Tests

- `TestBuildDesiredState_MultiSlotPoolNamedSession_OneRoutedBeadProvisionsTwoWorkers`
  covers the acceptance case (one routed bead, two-worker pool, both provisioned).
- `TestGCNonTestFilesStayOnWorkerBoundary` guards the containment invariant (gc's
  own non-test files stay on the worker boundary, do not leak into the checkout).
- Full `make check` (fmt-check, lint, vet, routed-test-rows, full suite) passes.

## Test plan

- [ ] CI green on the branch
- [ ] Verify a real multi-slot pool dispatch provisions two isolated worktrees for a
      single routed bead in a live rig

PR #5193, opened 2026-08-12
feat(worktree): single transactional owner for agent workspaces

sjarmak:
Agent workspaces are provisioned ad hoc today: several call sites run their own
`git worktree add`, decide their own safety rules for removing one, and record
no durable evidence of who created what. This adds `internal/worktree` as the
single owner of that lifecycle, plus a `gc worktree` command that exposes it.

## The contract

`Ensure` is verify-first and transactional. An already-valid worktree returns
unchanged. A path that exists in any other state is an error, never a repair:
the owner does not clobber state it did not create. Creation never detaches,
every postcondition is re-verified afterwards, and a failure rolls back the
worktree and branch the call created. `Verify` checks the postconditions
without mutating. `Cleanup` removes a workspace only after proving the
repository, path, branch, and durable ownership all match, with no force mode
and no recursive-filesystem fallback.

Ownership lives in the worktree's private git directory, never in the checked
out files, and records the creating attempt. `Cleanup` requires the caller to
name the attempt it is authorized to remove. Bead, owner, and generation are
all reproduced exactly when the same work is re-provisioned at the same path,
so those three identify the slot rather than its occupant; only the attempt id
distinguishes a workspace from its replacement.

`--dry-run` is observationally pure. It plans and validates and writes nothing,
including no lock file, and its provenance omits creation facts rather than
emitting a zero timestamp and an empty id that a consumer could publish as
evidence.

## Three ways this used to lose or leak committed work

Each of these guards measured something real, and none measured the thing that
protects the work.

Cleanup gated removal on push state. Once a branch is squash-merged and deleted
from the remote, no remote-tracking ref reaches that HEAD again, so the gate
latched on permanently and refused forever, leaking a worktree per merged unit
of work. It now asks whether removal would orphan commits, which is the
question that protects work: `git worktree remove` deletes the checkout, not
`refs/heads`. #4816 fixed the same defect for the reaper.

Rollback force-deleted the attempt branch after checking only the working tree.
An agent that commits leaves a clean status, so the guard passed and
`git branch -D` removed the only ref reaching those commits. Rollback now uses
`git branch -d` and lets git refuse the delete unless it can prove nothing is
lost.

A failed creation returned without rolling back. `git worktree add` is not
atomic: it registers the worktree and creates the branch before running
post-checkout work, so a hook or checkout failure left both behind with nothing
in the error saying so. Failed creation now reconciles against actual state,
removing only what is present and deleting the branch only when this call
created it.

## Concurrency

Provisioning is a check-then-act sequence, so `Ensure`, `Cleanup`, and rollback
hold an exclusive lock on the workspace path from their first observation
through their last write. Locking only the write is not enough: `Ensure`
decides whether to create from state it read beforehand, so a rival that
creates the workspace in between turns this call's `git worktree add` into a
failure whose rollback removes the winner's work.

The lock file lives under the repository's common git dir. It has to outlive
the removal it guards, so it cannot live inside the workspace, and it must not
leave a permanent file per workspace in a directory callers list expecting only
workspaces. Its key resolves symlinks, so two specs naming one workspace
through different ancestors agree on one lock.

## Relationship to #3958

#3958 provisions per-bead worktrees during desired-state build with its own
`WorktreeAdd` helper and a detached checkout. This supersedes that approach:
the same need is met by one owner with a verify-first, never-detaching
contract, and both branches touch `cmd/gc/build_desired_state.go` and
`internal/git`. #3958 has been conflicting since July; it should be closed
once this lands rather than rebased.

## Test plan

`internal/worktree` and `internal/git/worktree_ops` carry unit coverage for the
happy path and for each refusal. The fixes above have regression tests that
were confirmed to fail when their fix is reverted:

- a squash-merged worktree whose remote branch is gone is removed rather than
  refused forever;
- rollback refuses to delete a branch carrying committed work, and the commit
  survives;
- a stale attempt id is refused while the live one still works;
- two concurrent `Ensure` calls leave exactly one creator and an intact,
  verifying workspace;
- one workspace named through two different symlinked ancestors resolves to one
  lock file, including when its parent directories do not exist yet;
- rollback with an incomplete spec writes nothing into the repository the
  process happens to be standing in;
- a dry run leaves the workspace root and the whole common git dir byte-for-byte
  unchanged.

Lock exclusion is proven by a non-blocking acquisition of the derived lock
file rather than by racing a goroutine, so it does not depend on scheduling and
uses no sleeps.

Verified: `go build ./...` and `go vet ./...` clean for linux and darwin;
`go test -race ./internal/worktree/`; the product-metrics census, CI-topology,
and legacy-freeze suites; `test/docsync`; `make lint-changed`. The new package
also builds and vets under `GOOS=windows`, following the unix/windows split
this repo already uses for the repo-cache and pack-registry locks. Several
other packages still do not build for windows, on this branch and on main
alike.

`cmd/gc` passes with no test failures. It ends on the pre-existing dolt
sql-server leak guard in `TestMain`, which fails identically on this branch's
merge base.

New surface is registered in the product-metrics command census with permanent
IDs, and `docs/reference/cli.md` is regenerated.

Not included: the pool resume and wake-known-identity tiers still ignore
managed-worktree evidence and fall back to the configured work dir. Threading
evidence into them changes the pool-demand signature chain, so it is tracked
separately in #5753.

A second review pass found that a rig named `city` is indistinguishable from
the city store in the pool demand vocabulary, which identifies the city store
by the bare string `city` and reserves nothing at config validation. That
collision predates this change and is wider than the guard added here; the
guard resolves it fail-closed, refusing the workspace rather than binding to a
store it cannot prove it owns. Filed as #5754.

---
- Related to #1181 — adds internal/worktree and a `gc worktree` command as a single transactional owner for agent workspace provisioning — the per-agent worktree mechanism the isolation fix described in that issue would build on. It does not switch implicit pool agents over to it: with no worktree evidence on a bead, the pool path still falls back to the configured work dir, so parallel dispatch into one shared rig tree is unchanged.
- Related to #5188 — adds internal/worktree plus a `gc worktree` command as a single transactional owner for agent workspace provisioning, durable ownership provenance, and gated cleanup — a concrete implementation of the provisioning, ownership/affinity, and detection-and-recovery layers enumerated in that issue; it does not provide write confinement, git guardrails, or workflow-stage validation, and the inventory the issue asks for is still open


sjarmak:
@julianknutsen Before we rebase this PR, could you confirm the intended ownership boundary? Commit `646907f5b` removed `gc worktree` because lifecycle was considered a topology concern, but Gas City now has isolated workflows that require a transactional `gc worktree ensure/verify/cleanup` authority with generation and provenance checks. Should this PR continue with that CLI boundary, or should the lifecycle remain outside `gc`? We will not rebase until the direction is confirmed.

sjarmak:
Rebased onto current main and force-pushed. The branch had fallen 243 commits behind and was conflicting, so nothing here could run.

The ten commits are squashed into one. Nine of them were review fixups on the same package, and a squash merge would have collapsed them anyway; rebasing the stack commit by commit meant resolving the same conflicts ten times over. The pre-rebase head `dc9f1f348` is preserved if anyone wants the original sequence.

Eight files conflicted. Three needed real decisions:

- `cmd/gc/build_desired_state.go`: additive on both sides. Main's `namedTriggerRefIsSameStore` and `bindNamedSessionTriggerBead` are kept alongside this branch's `verifiedPoolTriggerWorkDir`, with `poolTriggerMetadata` taking the error-returning signature.
- `cmd/gc/pool_desired_state.go`: the substantive one. Main restructured the demand loop underneath this branch, adding protected requests, `allocateScaleDemandToConcrete` and `residualWorkBeadIDs`. The worktree plumbing is ported onto that new structure rather than the old loop being restored: the spec and error declarations, the two `demand.WorktreeSpecs` / `WorktreeErrors` lookups relocated inside main's `residualWorkBeadIDs` block, the two `SessionRequest` fields, and `worktreeSpecForBead`.
- `cmd/gc/pool_desired_state_test.go`: import block, both sides kept.

The command census was merged by hand rather than resolved to one side. `go run ./cmd/gen-command-census` now regenerates the file with no diff, and `gc worktree cleanup`, `ensure` and `verify` take ids 203 through 205 off main's post-merge `next_id` of 203 with no collision.

The three failing checks on the previous head were one failure. `internal/testutil/providerledger` failed in the integration shard, and the two summarize-gates jobs failed downstream of it. This branch adds `internal/testutil/git.go`, which that ledger covers. It passes now.

Local verification on `c577a44d3`: `go build ./cmd/gc/` and `go vet` clean over the six touched packages, `golangci-lint` reports 0 issues across the tree, `go test ./cmd/gc/` passes in 390s, and `internal/worktree`, `internal/git`, `internal/testutil/providerledger`, the test-policy guard suites and `make check-docs` all pass.

One note for whoever reviews: `TestDisableAndPurgeBoundsInitialAndPostUploaderStateLocks` in `internal/productmetrics` failed once here while three packages ran concurrently, and passes both targeted and as a full package alone, on this branch and on main. It waits on a phase transition against real elapsed time, so a contended host reads as a stall. Unrelated to this change and tracked separately.

sjarmak:
Preflight was red on the rebased head, and the three failing checks were one failure: `Native dependency surface guard` in `Preflight / static checks`. Everything after it in that job was skipped, and `Check`, `CI / preflight` and `Evidence / critical-path suites` were reporting that job's result back.

```
native dependency guard: gc binary is 270065624 bytes; max is 270000000
```

Measured locally, main is 269,932,192 bytes, 68 KB under the ceiling. This branch adds about 2,000 lines of production Go across `internal/worktree`, `internal/git` and `cmd/gc`, which brings the binary to 270,088,704.

The guard is sized for dependency bloat, not source growth: it also caps the module graph (727), the AWS/Azure/DoltHub/Google module counts, and asserts a single beads module. `go list -m all` is unchanged at 727 here, and go.mod/go.sum are untouched by this branch.

Second commit raises the byte ceiling to 272,000,000, leaving about 1.9 MB of headroom for ordinary growth. A real dependency addition moves this number by tens of megabytes, so the guard keeps its teeth.

Guard re-run locally on the new head: `modules=727 aws=25 azure=9 dolthub=15 googleapi=1 binary_bytes=270088704`, pass.

PR #5194, opened 2026-08-12
fix(beads): fill the ready projection from `bd blocked` where `bd sql` is refused

julianknutsen:
## Summary

`BdStore` now has two doors onto bd's `is_blocked` column instead of one.
`bd sql` stays primary wherever it works — every Dolt and DoltLite scope
runs byte-identically. A scope whose backend gc does not implement, or
whose bd refuses `bd sql` at runtime, now gets its column from
`bd blocked --json` instead of losing it.

## Problem

`fetchReadyProjection` filled the cache's `is_blocked` column by shelling
`bd sql`. bd refuses that verb unless it holds a SQL session
(`cmd/bd/sql.go`: *"'bd sql' is not yet supported in embedded mode"*), which
a hosted-Postgres work store never does. #5183 bounded the damage with a
backend-capability gate and a per-scope latch, so the failing 6-16s
subprocess is spent at most once — but the scope then has NO column at all,
every bead's `IsBlocked` is nil, and `CachingStore.readyReadsMustGoLive`
sends every readiness read to a live `bd ready`.

That is maintainer-city's last non-latency error:

    city work ready: reading complete ready projection from cache: bead cache unavailable

Measured live against MC's hosted-Postgres work store: `bd sql` costs 6.5s
and then fails; `bd blocked --json` answers in 2.9s; `bd ready` costs 2.5s
and is what every readiness read pays today.

## Why `bd blocked` is the same column

`bd blocked` is not a direct-dependency answer. `issueops.GetBlockedIssuesInTx`
seeds its row set with

    SELECT id FROM issues|wisps WHERE is_blocked = 1 AND status <> 'closed' AND status <> 'pinned'

— the denormalized, transitively-propagated column that bd's own ready
filters with `is_blocked = 0` (`sqlbuild.BuildReadyWorkWhere`). Same column,
same closure: a child of a blocked parent reads blocked to `bd blocked`
exactly as it does to `bd ready`. A subtly different definition here would
re-create #3218, so the equality is what licenses the door.

Absence means NOT blocked and is written as `false`, never left nil: a nil
verdict falls back to the bead's own direct blocks/waits-for/conditional-blocks
deps, which is the weaker predicate this projection exists to replace.

## Where it is not the column, stated

After seeding from `is_blocked`, bd narrows its OUTPUT to rows whose blocker
it can NAME — an active blocking dep onto a resident non-closed row, or
failing that any parent-child edge. A row carrying `is_blocked = 1` with
neither is dropped and therefore reads `false` here. bd's own recompute
(`issueops.shouldBeBlockedDisjunction`) can only produce such a row from a
stale flag (what `bd recompute-blocked`/`bd sync` repair and
`countStaleIsBlockedSQL` counts) or from a waits-for onto a spawner that
closed with its gate still shut — which in a gc-minted graph is also a
molecule step and so carries the parent-child edge that attributes it.

Measured on maintainer-city: 117 active rows, `bd blocked` names 3,
`bd ready` returns 112, and the 3 in neither are accounted for by bd's own
non-is_blocked ready predicates (2 `in_progress` rows, which `bd ready` never
returns because `cmd/bd/ready.go` hardcodes `Status: "open"`, and 1 molecule,
dropped by `sqlbuild.ReadyWorkExcludeTypes`). Zero unattributable rows.

This is why the SQL door stays primary rather than being replaced: where
`bd sql` works it returns the column with no attribution filter at all.

## Fail-closed

- The backend gate still never lets `bd sql` reach a backend gc does not
  implement — its reason was gc's inability to assume that backend's SCHEMA,
  and that reason does not reach `bd blocked`, which is bd's own verb over
  its own role.
- bd refusing the blocked VERB latches exactly as before, naming both doors,
  with one operator notice and one attempt per door per scope.
- Any other `bd blocked` failure (timeout, reset, busy server) is this
  cycle's failure: returned plain, which leaves the snapshot partial and
  already declines cached readiness, and retried next reconcile. Latching a
  blip would cost a long-lived dispatcher its projection for the process.

## Cost

One subprocess per cache prime and per reconcile, independent of bead count:
bd answers the whole scope in one call the way `bd sql` did. Inside bd the
work scales with the BLOCKED set, not the ledger. What it buys is every
readiness read between reconciles, each of which spends a live `bd ready`
today.

## Tests

`TestBdBackedCachedReadyNeverOffersWorkTheBackingHides` gains a
`bd blocked door` subtest over the standard fixture, which grows two rows so
the door can be disproved in both directions: `gcg-offscope` (a row bd's
ledger holds and gc's class routing keeps out of this cache, which is the
only internally consistent way to state a cross-store blocker — bd's
recompute joins the target table, so a dangling id leaves `is_blocked = 0`)
and `bd-gate-open` (a waits-for whose gate has opened, so bd offers it while
the direct-dep predicate would hide it). The fake ledger's `bd blocked`
models both halves of `GetBlockedIssuesInTx`, including the attribution
narrowing.

Mutation-proved, five ways:

- delete the blocked door → `CachedReady declined (bd blocked door, after prime): a backing that can answer is_blocked must keep serving readiness from cache`
- absence leaves nil → `CachedReady (bd blocked door, after prime) = [bd-blocker bd-unrelated], want [bd-blocker bd-gate-open bd-unrelated]` and `bead mc-1 is_blocked = <nil>, want &false`
- drop the fake's parent attribution → `Ready (bd blocked door, after prime) offered bd-child, which the backing's own Ready hides` on all four readiness handles
- route every scope to the blocked door → `TestReadyProjectionOnAnImplementedBackendIsUnchanged` fails for dolt, doltlite and unset
- remove the offscope blocker row → `Ready (bd blocked door, after prime) offered bd-xstore, which the backing's own Ready hides`

### DIFF
--- a/internal/beads/bdstore.go
+++ b/internal/beads/bdstore.go
@@ -373,6 +373,10 @@ type BdStore struct {
 	readyProjectionMu      sync.Mutex
 	readyProjectionChecked bool
 	readyProjectionEnabled bool
+	// readyProjectionDoorValue names which bd verb fills this scope's
+	// is_blocked column: `bd sql` where the backend implements it, `bd blocked`
+	// where it does not. See bdstore_ready_projection.go.
+	readyProjectionDoorValue readyProjectionDoor
 	// readyProjectionVersionErr memoizes the ErrReadyProjectionUnsupported this
 	// store owes every later caller when the bd on PATH predates the is_blocked
 	// projection. Store-local rather than scope-latched: the bd binary is a
--- a/internal/beads/bdstore_inline_deps_internal_test.go
+++ b/internal/beads/bdstore_inline_deps_internal_test.go
@@ -18,6 +18,13 @@ type bdLedgerRow struct {
 	closed    bool
 	assignee  string
 	deps      []Dep
+	// offscope marks a row bd's ledger HOLDS but this scope's cache never
+	// receives: gc routes relocated classes (the `gcg-` graph beads) to another
+	// store, so `bd list` and `bd query` for this scope do not hand it over.
+	// bd's own answers — ready, sql, blocked — still see it, which is what
+	// makes an edge onto it invisible to the cache's predicate and visible to
+	// bd's column.
+	offscope bool
 }
 
 // fakeBdLedger answers the bd subcommands a CachingStore prime spends, the way
@@ -54,13 +61,15 @@ func (f *fakeBdLedger) run(_, name string, args ...string) ([]byte, error) {
 	case joined == "version":
 		return []byte("bd version 1.1.0\n"), nil
 	case args[0] == "list":
-		return f.renderRows(func(r bdLedgerRow) bool { return !r.ephemeral && !r.closed })
+		return f.renderRows(func(r bdLedgerRow) bool { return !r.ephemeral && !r.closed && !r.offscope })
 	case args[0] == "query":
-		return f.renderRows(func(r bdLedgerRow) bool { return r.ephemeral && !r.closed })
+		return f.renderRows(func(r bdLedgerRow) bool { return r.ephemeral && !r.closed && !r.offscope })
 	case args[0] == "ready":
 		// bd ready filters is_blocked = 0 (sqlbuild.ReadyWhere). This is the
 		// backing's own verdict, the one the cache may never exceed.
 		return f.renderRows(func(r bdLedgerRow) bool { return !r.blocked && !r.closed })
+	case args[0] == "blocked":
+		return f.renderBlocked()
 	case args[0] == "show":
 		// `bd show --json` answers from types.IssueDetails, which embeds
 		// types.Issue — and NOTHING in beads carries a `json:"is_blocked"` tag.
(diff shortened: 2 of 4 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277174213

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR fills the ready projection from `bd blocked` on backends where `bd sql` is structurally refused (embedded-refused Dolt, hosted-Postgres work stores), so readiness still enriches blocked rows with their `is_blocked` signal instead of silently dropping it and re-paying a failing SQL probe on every store rebuild. That keeps blocked work correctly gated on exactly the backends where the SQL door was never available.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- The runtime "blocked door" choice was only recorded on a single `BdStore` instance, so a fresh store over the same scope re-spent the refused `bd sql` probe (6–16s) on every ~3s store rebuild. The maintainer fixup persists the choice at scope scope — a per-scope `blockedDoor atomic.Bool` latch written under `readyProjectionMu` and consulted before door selection — so a proven-bad scope starts on the blocked door without re-probing SQL.
- A comment claimed the runtime switch already avoided rediscovery across store rebuilds before that scope-level persistence existed; it was corrected to match the implemented behavior.
- The rebuild-survival property had no regression coverage; `TestReadyProjectionRuntimeBlockedDoorSurvivesStoreRebuilds` was added, driving 5 rebuilds over one scope and asserting the SQL probe is spent at most once while the blocked door serves all 5 with correct `IsBlocked` and no spurious degrade notice.

Review findings addressed:
- Persist-the-blocked-door-at-scope-scope (major, attempt 1) — fixed in commit `258a549f4a` (`fix(beads): persist the runtime blocked-door choice at scope scope`). Both reviewers confirmed the fix; the new regression pins the SQL probe to exactly one spend across 5 rebuilds.

- `mc-rwlk` (Testing, minor): the transient `bd blocked` failure branch (`internal/beads/bdstore_ready_projection.go:502`, the unsupported-vs-transient classifier) has no negative-path test, so a future change that broadened the unsupported branch or dropped the plain-return could silently latch a scope's projection off for the process lifetime with no failing test. Filed as a follow-up; not addressed here.
- `mc-kcyw` (Concurrency, minor): stores racing the first latch write can each pay one bounded 6–16s `bd sql` refusal before the latch takes (`internal/beads/bdstore_ready_projection.go:206`, `:287`). Correctness is unaffected — writes are mutex/atomic-guarded and the scope converges to the blocked door once latched — so this is an optional documentation/acceptance item. Filed as a follow-up; not addressed here.

### Maintainer Changes
Maintainer fixup pushed directly to the PR branch: `258a549f4a fix(beads): persist the runtime blocked-door choice at scope scope`. Short diff stat: 2 files changed, 122 insertions(+), 12 deletions(-).

### Final Review Status
Ready for the merge queue: final head `258a549f4a` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/32198998685

### Review Iterations
2 review passes performed: attempt 1 requested changes (persist the blocked-door choice at scope scope, correct the stale comment, add the rebuild-survival regression test), and attempt 2 approved after the maintainer fixup landed all three.

---

PR #5196, opened 2026-08-12
fix(supervisor): tear down the provider server on managed city stop (#5175)

### LINKED ISSUE #5175: gc stop leaves the city tmux server (mayor session) running after reporting "City stopped."
`gc stop` reports "City stopped." but the city's tmux server — with the mayor session in it — survives the stop and is still running 45+ seconds later.

Reproduction (macOS/darwin, gc built from `365f111ea`, bd from the go.mod-pinned beads `bf97b73749ac`):

```
export GC_HOME=~/.gcshort   # short path; supervisor.sock hits the ~104-byte unix socket limit otherwise
gc init --template minimal --default-provider claude --no-start tc2
cd tc2 && gc start          # wait for "City started under supervisor."
ps -eo pid,command | grep tc2   # watchdog, dolt sql-server, tmux -L tc2 (mayor), 2 provider processes
gc stop                     # prints "Reconciliation triggered." / "City stopped."
ps -eo pid,command | grep tc2   # immediately and at T+45s:
#  tmux -u -L tc2 new-session -d -s mayor ...   <- still alive
```

Observed: the managed dolt server, its scope watchdog, and the provider (claude) processes all stop correctly. The `tmux -L tc2` server with the `mayor` session does not — it persists indefinitely until a manual `tmux -L tc2 kill-server`. A transient gc helper process also survived the stop but exited on its own within 45s; the tmux server is the durable leak.

Expected: after "City stopped." a process census for the city shows nothing, including its tmux server.

Impact: repeated start/stop cycles accumulate orphaned tmux servers (one per city socket), and anything that treats "tmux session exists" as a liveness signal reads a stopped city as alive.

jacobhausler:
`gc stop` on a supervisor-managed city reports "City stopped." but the city's `tmux -L <city>` server — mayor session included — survives indefinitely (repro in #5175: everything else exits; the tmux server is the durable leak; repeated cycles accumulate one orphaned server per socket).

**Root cause.** Only the standalone stop path calls the server teardown: `cmdStopBody` → `teardownServerForStop` → `runtime.ServerLifecycleProvider.TeardownServer` (cmd/gc/cmd_stop.go). The managed path — `stopManagedCity`, `stopManagedCityPreservingSessions`, and the city run loop's deferred `CityRuntime.shutdown()` — stops every session but never tears the shared server down.

**Fix.** `CityRuntime.shutdown()` now calls `teardownServerForStop(cr.sp, cr.stderr)` after the session stops (including the late force-stop sweep), mirroring `cmdStopBody`'s ordering. One call site; the existing type-assert keeps non-lifecycle providers untouched.

**Preserve-sessions is deliberately unchanged.** The preserving shutdown returns before the session stops and never reaches the teardown: preserved sessions live inside that server, so keeping it up for supervisor re-adoption is by design, not an omission. `TestCityRuntimeShutdownPreservingSessionsKeepsServer` pins that.

**Tests.** `TestCityRuntimeShutdownTearsDownProviderServer` fails on stock main (no TeardownServer call) and passes with the fix; it also pins ordering (after the session sweep) and once-ness across repeat shutdowns. A no-lifecycle-provider case pins that nothing new is required of other providers. Reuses `lifecycleOrderProvider` from the standalone-stop lifecycle tests.

Fixes #5175

### DIFF
--- a/cmd/gc/city_runtime.go
+++ b/cmd/gc/city_runtime.go
@@ -203,9 +203,17 @@ type CityRuntime struct {
 
 	shutdownOnce             sync.Once
 	preserveSessionsShutdown atomic.Bool
-	forceStopShutdown        *atomic.Bool
-	logPrefix                string // "gc start" or "gc supervisor"
-	stdout, stderr           io.Writer
+	// ownedCity is set true once run() begins, i.e. once this runtime has
+	// passed every init-failure/discard point and is the process actually
+	// driving the city. It gates the shutdown-time server teardown: a
+	// discarded half-built runtime (e.g. adoption of an already-running
+	// city, or a controller-lock/socket/token failure) calls shutdown()
+	// without ever owning the city, and must not tear down the live city's
+	// shared server out from under it.
+	ownedCity         atomic.Bool
+	forceStopShutdown *atomic.Bool
+	logPrefix         string // "gc start" or "gc supervisor"
+	stdout, stderr    io.Writer
 }
 
 const runtimeDemandSnapshotMaxAge = 30 * time.Second
@@ -451,6 +459,10 @@ func (cr *CityRuntime) crashTrack() crashTracker {
 // the per-city main loop — it watches config, reconciles agents, runs
 // wisp GC, and dispatches orders.
 func (cr *CityRuntime) run(ctx context.Context) {
+	// Reaching run() means every init-failure/discard point is behind us:
+	// this runtime is the live owner of the city, so its shutdown() is the
+	// one allowed to tear the provider's shared server down.
+	cr.ownedCity.Store(true)
 	defer cr.shutdown()
 
 	dirty := cr.configDirty
@@ -3838,12 +3850,21 @@ func (cr *CityRuntime) shutdown() {
 				fmt.Fprintf(cr.stderr, "%s: shutdown session listing failed: %v\n", cr.logPrefix, listErr) //nolint:errcheck // best-effort stderr
 			}
 		}
+		// sweepEnumerated tracks whether every stop pass this shutdown ran
+		// actually listed the fleet. A failed ListRunning yields an empty
+		// slice, so gracefulStopAll stops nothing — tearing the server down
+		// after that would turn a transient listing error into an ungraceful
+		// mass kill of sessions we never enumerated. Partial failures count
+		// as not-enumerated too: we did not see the whole fleet, so we must
+		// not kill-server on the strength of a subset.
+		sweepEnumerated := listErr == nil
 		store := cr.sessionsBeadStore()
 		markCityStopSessionSleepReason(sessionFrontDoor(store.Store), cr.stderr)
 		gracefulStopAllWithForceSignal(running, cr.sp, gracefulTimeout, cr.rec, cr.cfg, store, cr.stdout, cr.stderr, cr.forceStopRequested)
 		if !asyncStartsDrained && cr.forceStopRequested() {
 			lateRunning, lateListErr := cr.sp.ListRunning("")
 			if lateListErr != nil {
+				sweepEnumerated = false
 				if runtime.IsPartialListError(lateListErr) {
 					fmt.Fprintf(cr.stderr, "%s: force shutdown late async-start listing partially failed; stopping %d visible agent(s): %v\n", cr.logPrefix, len(lateRunning), lateListErr) //nolint:errcheck // best-effort stderr
 				} else {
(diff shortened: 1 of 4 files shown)


jacobhausler:
Marking this draft. On self-review we found the teardown is placed too broadly: it runs at the end of `CityRuntime.shutdown()`'s Once body unconditionally, so it also fires from supervisor init-failure paths that discard a half-built runtime which never owned the city (a failed adoption of an already-live city would run kill-server on that live city's socket), and from the case where the session-listing sweep itself errored (an empty list we never actually enumerated, turned into a mass kill). Tightening it to only tear down when this runtime actually owns the city and the stop reached a real session sweep. Will un-draft once the gating is in and the suite is green.

jacobhausler:
Pushed 265dab7bd, which gates the managed-stop teardown added for #5175 behind two conditions:

- **Ownership (`ownedCity`)**: set only when `run()` begins, i.e. after every init-failure/discard point. The supervisor calls `shutdown()` on discarded half-built runtimes too (failed adoption of an already-running city, controller-lock/socket/token failures); an un-owned shutdown must never `kill-server` the live owner's shared socket. Without this gate, a failed adoption would kill every session in the city it failed to adopt.
- **Real sweep (`sweepEnumerated`)**: teardown only runs after a stop pass whose `ListRunning` actually succeeded (including the late force-mode re-list). A failed listing yields an empty slice, so `gracefulStopAll` stops nothing — tearing the server down then would turn a transient listing error into an ungraceful mass kill of sessions we never enumerated. Partial listing failures count as not-enumerated for the same reason.

Preserve-sessions shutdown still returns before the teardown block by design: preserved sessions live inside the provider server, so keeping it up is the point of preserve mode.

Six new tests in `city_runtime_server_lifecycle_test.go` pin each gate: teardown-exactly-once after the sweep, un-owned shutdown keeps the server, failed enumeration skips teardown, preserve mode keeps the server, non-`ServerLifecycleProvider` providers are a no-op, and force-mode teardown lands only after the late async-start re-list.

Local test status for reviewers: `make test-cmd-gc-process` on this branch fails 3 tests (`TestBuildDesiredState_MinZeroDefaultScaleCheckRoutedWorkCreatesPoolSession`, `TestEvaluatePoolDefaultScaleCheckCountsRoutedReadyWork`, `TestEvaluatePoolDefaultScaleCheckIgnoresRoutedActiveUnassignedWork`) on my machine, but an identical run on upstream `main` (e2fd3c46d) fails the exact same 3 with the same error — my host `bd` is a `CGO_ENABLED=0` dev build that can't open an embedded Dolt store, which those slow-gated tests require. Everything else passes locally (8730 pass / 43 skip verbose); deferring to CI's properly built `bd` for the trio.

jacobhausler:
Marking ready for review — CI is green (59 pass / 0 fail; remaining checks are conditionally-skipped jobs).

Summary of the branch since draft:
- **Fix**: on managed city stop, gate the tmux/provider-server teardown on ownership plus a real process sweep, so a managed stop no longer tears down a provider server it doesn't own (#5175).
- **Test hardening** (`e8c050184`): the census lifecycle test's fixed-sleep poll was racy under the census `fixed_sleep` ratchet (287 vs 286 tick flake seen in CI); replaced with an explicit `workerStarted` channel wait inside the `hangBudget` window. `commandcensus`, lifecycle, and `-race` green locally and in CI.
- The one red on the previous CI run was `internal/productmetrics` `TestRecordOnceDecisionExpiryDuringNoReplacePrestatePreventsInstall` ("load quota: absent") — untouched by this cmd/gc-only diff; it passed on re-run and looks timing-sensitive. Flagging in case maintainers want a flake issue for it.

quad341 (approved):
@jacobhausler — this is a sharp fix, and the part I want to call out is the self-review. Noticing that an unconditional teardown would let a *failed* adoption `kill-server` the live city's socket, and then gating on ownership plus a genuine enumeration instead of shipping the one-liner, is the difference between fixing #5175 and trading it for a much worse bug.

I traced both gates rather than taking the description's word: every supervisor discard path calls `shutdown()` strictly before the goroutine that enters `run()`, so `ownedCity` holds, and `preserveSessionsShutdown` is set from exactly one place, so the preserve asymmetry is real rather than incidental.

`TestCityRuntimeForceShutdownTearsDownAfterLateAsyncSweep` is the test I'd have asked for and didn't have to — pinning teardown after the *second* `ListRunning` is exactly the ordering that rots silently otherwise.

Bonus you may not have clocked: this also closes the standalone case where `gc stop` delegates to a live controller and returns before `teardownServerForStop` ever runs. Same leak, now gone.

Merging as-is. I'll open a separate issue for socket-uniqueness validation across managed cities — unrelated to your change, just adjacent.


---

Issue #3409, opened 2026-06-12
bug: dolt-health advisory [MEDIUM] is a false-positive storm — 1s-granularity latency trips a 1s threshold, and every tick spawns a new un-deduped mayor bead


bobbyhyam:
### Before you continue

- [x] I searched existing issues and did not find a duplicate. Sibling: **#3374** (`gc status` store-health "maintenance overdue" false positive on small cities) — same "health heuristic miscalibrated for a quiet/small instance" family, but a different code path. Nothing covers the dolt-health **advisory** latency trip or the per-tick mayor-bead duplication.
- [x] I read the relevant docs and contributor guidance.

### Gas City version

1.2.1 (Homebrew). The code is in the bundled **`dolt` system pack**, unchanged at this version:
`.gc/system/packs/dolt/assets/scripts/mol-dog-doctor.sh`, run by `orders/mol-dog-doctor.toml` (`trigger = "cooldown"`, `interval = "5m"`).

### Environment

Linux (kernel 7.0), bash, tmux provider, single-operator low-traffic city. Dolt server local and healthy (sub-second queries). 2 always-on named sessions + 1 rig.

### Reproduction

1. Run a normally-quiet city and let the bundled `mol-dog-doctor` order tick (every 5m).
2. The doctor probes Dolt latency by wrapping a single `SELECT active_branch()` in **whole-second** timestamps and warns when `latency >= threshold` (default 1s).
3. Watch `mayor`'s inbox: **"Dolt health advisory [MEDIUM]"** messages accumulate, each body `Latency: 1s [WARN: latency 1s >= threshold 1s]` with every other metric green.
4. Live state on our city right now: **18 unread** such advisories in the mayor inbox, one root cause, each a distinct message bead. Sample `sp-wisp-974noq` (2026-06-11 19:07:23):
   ```
   From: controller   To: mayor   Subject: Dolt health advisory [MEDIUM]
   Latency: 1s [WARN: latency 1s >= threshold 1s]
   Connections: 2/50
   Disk: 416M
   Orphan DBs: 0
   ```

### Expected behavior

A healthy Dolt server (sub-second probe) generates no MEDIUM advisory. Real latency is measured with enough resolution/headroom to be meaningful, and a persistent condition collapses into **one** rolling advisory rather than a new bead every tick.

### Actual behavior — two compounding defects

**1. Latency is measured at 1-second granularity and compared against a 1-second threshold with `>=`, so quantization noise alone trips it.**

```sh
[...]

PR #5200, opened 2026-08-12
fix(dolt): archive superseded health advisories in the doctor dog

tobasummandal:
## Summary
- The dolt-health doctor's condition-set dedup bounds how often an advisory is *sent*, but nothing bounds how many advisory beads stay *open*: each condition-set change strands the previous unread advisory, and a city running a pre-dedup pack copy accumulates one identical bead per 5-minute tick until someone sweeps by hand (observed live: 40+ "Dolt health advisory [MEDIUM]" beads for one steady missing-backup-remote condition).
- New `advisory_archive_superseded` helper in `advisory_state.sh` (bounded, recipient-scoped `gc mail archive --to ... --subject-prefix ... --limit`), wired into `mol-dog-doctor.sh` at the two supersession points: immediately before a fresh advisory send, and on the advisory→healthy transition (gated on the recorded state file so steady healthy ticks make no extra gc call).
- Safety properties: the sweep resolves its recipient exactly as core `escalate.sh` does (`GC_ESCALATION_RECIPIENT`, default `human`); the CRITICAL unreachable-server subject never matches the prefix, so outage mail is untouched; archive retains beads (closed, body kept); and the helper fails open — a failed or absent archive command degrades to a lingering bead, never a lost alert.

## Test plan
- [x] `sh test/dolt/advisory_dedup_test.sh` — new helper-level cases (bounded invocation shape, limit propagation, no-op on missing args, fail-open on gc failure) — ALL PASS
- [x] `CGO_ENABLED=0 go test ./examples/bd/dolt/ -count=1` — full package green, including four new integration cases: sweep precedes the fresh send, healthy transition sweeps the recorded advisory and clears state, steady-healthy tick makes no gc calls, suppressed tick neither sweeps nor sends

---
> • City: `factory-main` · Agent: `local-core.builder-3`
> • On behalf of: @tobasummandal


quad341 (approved):
@tobasummandal — this is the good kind of bug fix: you spotted that #3409 only ever solved half the lifecycle. Bounding how often an advisory is *sent* does nothing about how many stay *open*, and 40+ stranded beads from one steady missing-backup-remote condition is the evidence that makes the gap undeniable.

The details are what sold me. Sweeping *before* the send, so the fresh advisory can't archive itself — and a test that asserts the actual ordering, not just that both happened. Gating the healthy-transition sweep on the recorded state file, so steady healthy ticks cost zero `gc` calls. Resolving the recipient the same way core `escalate.sh` does instead of hardcoding it. I checked each of those against `cmd/gc/cmd_mail.go` and the beadmail route matching; they all hold.

CI is red on our side, not yours: a provider-ledger waiver expired today and is reding every PR in the repo, main included. I'll land that fix, rebase your branch, and merge. I'm also tweaking one header comment to say *unread* advisories — the sweep doesn't pass `--include-read`, which is the right default.

Thanks for chasing the half nobody had gotten to.


---

PR #5199, opened 2026-08-12
fix(dispatch): bound semantic store refusals so a control deadlock cannot idle a fleet in silence

julianknutsen:
**Slice 1 of `ga-a6zy9`** (design: `win3-scopecheck-deadlock-design-v1`). Restores a six-city fleet that has been ~100% idle for three days while every health metric reported green. Slices 2–4 (self-edge repair, suppression bound, stop minting the cycle) are deliberately **not** here.

## The failure

A control bead and the bead it must close block each other. `bd` refuses to close a blocked issue, and the bead named as the blocker *is* the control bead asking for the close:

```
pl-mmneh  kind=scope-check      -> closes -> pl-pujtf  (scope body)
pl-pujtf  is blocked by         [pl-mmneh]                 <- the closer is the blocker
```

**PR #5020** classified `cannot close blocked issue` as transient in `IsTransientControllerError`, and in that classifier every "retry" meant **retry forever**. So the dispatcher re-asked a question the store had already answered, emitted zero events, and every health surface stayed green.

MEASURED: 296,270 refusal lines across four cities (platform 80,466 / substrate 88,856 / trust 63,924 / orchestration 63,024). On platform, **two** stuck beads consumed **13,419 of 14,125 control dispatches — 95% of the dispatcher's entire throughput** — because `pendingAny=true` resets `idleSweeps=0`, so the idle backoff never engaged. `pl-mmneh` carried **zero** failure metadata after 6,712 consecutive failures, and there was **no `control.*` event type at all**.

## What this changes

**1. Two tiers, and the table forces you to pick one.** `ControllerErrorTier` splits the classifier by *what the store did*:

- **Tier A — availability.** The store never answered (timeouts, refused/reset connections, lock contention, tripped Dolt breaker, work-query kills, CAS exhaustion). Unbounded retry, byte-for-byte the pre-split behavior, and **no store writes**.
- **Tier B — semantic.** The store answered and **refused** on the current graph state. Bounded, because repeating the question cannot change the answer.

That boundary is also exactly where a persisted budget is *implementable*: a store that never answered cannot be asked to record how long we have been asking it; a store that refused is by construction available to hold the deadline. Tier A is unbounded for a reason, not by omission.

[...]

PR #5202, opened 2026-08-12
fix(formula,dispatch): stop compiling a bead blocked by the control that closes it (ga-a6zy9)

julianknutsen:
## The bug

`ga-a6zy9` is filed as a control-dispatcher retry problem. It is not. **The
compiler mints the cycle.** A bead is created as a *blocker of the bead it must
close*, so the store refuses the close (`cannot close blocked issue`) and the
only bead that could ever clear the blocker is the one being refused. Nothing
breaks it by waiting; the dispatcher retries forever.

Two instances, both minted at graph-compile time:

| producer | edge | who closes the blocked bead |
|---|---|---|
| `internal/formula/graph.go` `rewriteGraphStepRefs` | `body --blocks--> <member>-scope-check` | `processScopeCheck` → `closeScopeAsPassed`/`abortScope` closes the body named by the control's `gc.scope_ref` |
| `internal/formula/compile.go` `addWorkflowRootDeps` | `root --blocks--> workflow-finalize` | `processWorkflowFinalize` closes `gc.root_bead_id` |

MEASURED over the shipped corpus at `origin/main@e2fd3c46d4` (20 formulas
compiled: `internal/bootstrap/packs/core/formulas`, `cmd/gc/testdata/formulas`,
`examples/bd/dolt/formulas`, plus the live `packs/teams/formulas`):
**17 body self-edges and 17 root self-edges across 17 graph.v2 formulas.**
That includes the SDK's own `mol-scoped-work`, whose *five* scope-checks each
block the body they close, and the canonical scope example in
`docs/reference/specs/formula-spec-v2.md` §3.5. Every graph.v2 workflow in the
repo compiles to at least one permanent deadlock.

PR #5199 (Slice 1) bounds the retry and makes the failure loud. This PR removes
the cause.

## Why the rewrite exists, and what preserves ordering without the self-edge

The rewrite repoints a step that waited on a scoped member S so it waits on
`S-scope-check` instead — i.e. on **scope convergence** (`abort_scope` skipping,
member-metadata propagation, output resolution) rather than on the raw member
close. That intent is real and is fully preserved: only the referring node the
control *closes* is exempt. `TestCompileRewritesDownstreamRefsButNotTheScopeBodys`
pins both halves in one fixture.

The scope body loses nothing, because **the body is a latch, and the design's
claim that `graphSinkStepIDs` already treats it as one holds** (verified, not
assumed):

- `graphSinkStepIDs`' `case "scope"` calls `appendSink(step.ID)` **unconditionally**
  — a scope body is a workflow-finalize blocker whether or not anything
  references it. Removing the body's own blocker cannot drop it from the sink set.
- The body is never dispatched and never routed: `beadmeta.WorkflowTopologyKinds`
  ("routing never lands on these") and `StructuralGraphKinds` ("never dispatched
  as control beads — the ProcessControl switch hard-errors on them"). Its
  readiness edge has **no consumer**.
- Causal ordering survives: `processScopeCheck` only calls `closeScopeAsPassed`
  when `hasOpenScopeMembers` reports none open, and `abortScope` calls
  `skipOpenScopeMembers` *before* closing the body. Either way every scope member
  is closed before the body close is attempted — and the body's retained blockers
  are exactly those members.

Same argument for the root: `workflow-finalize` already blocks on every graph
sink, so it cannot run early; the root is a latch closed by the finalizer.

## The fix

1. **`rewriteGraphStepRefs`** skips a replacement when the control closes the
   referring step. Every other rewrite is untouched.
2. **`addWorkflowRootDeps`** emits the root→finalizer edge as `tracks` instead of
   `blocks`.
3. **`ApplyFragmentRecipeGraphControls`** (dynamic fragments) and
   **`applyAttemptRecipeScopeChecks`** (attempt re-mint) — the other two
   producers of the same rewrite — now route through one guarded helper,
   `formula.RewriteRecipeDepsToControls`.

### Why `tracks` and not deletion, for the finalize edge

MEASURED against the deployed `bd 1.1.0 (bdb475ad5)` in a throwaway store:

```
A --blocks--> B (B open):  A absent from `bd ready`
                           `bd update A --status closed`
                             -> cannot close blocked issue: A is blocked by [B]
A --tracks--> B (B open):  A present in `bd ready`
                           `bd update A --status closed` -> OK
```

So `tracks` removes exactly the readiness/close semantics and keeps the row.
That matters: `graph_apply.go` suppresses the finalize→root `tracks` ownership
edge for single-step workflows (#4125 / `su-mla5h`), so deleting the compiler's
edge would leave a single-step finalizer with **no** edge to its root. Keeping it
as `tracks` preserves cascade-delete reachability and
`isOrderWispDescendantDepType` traversal (both accept `tracks`), while the root
is never routed (`graphroute` skips `IsWorkflowTopologyKind` and `IsRoot`), so
its new readiness has no consumer. #4125's suppression is deliberately left in
place — out of scope here, and harmless now that the pair is informational on
both sides.

## The structural invariant (worth more than the fix)

- **`beadmeta.ControlClosesNode`** — one answer to "does completing this control
  close that node", derived from the two `ProcessControl` handlers that actually
  do it. `SelfClosingControlKinds` is pinned against `ControlKinds` by
  `TestControlClosesNodeOnlyForSelfClosingKinds`, so a future control kind that
  closes a foreign node cannot be added without declaring itself.
- **`formula.ValidateNoSelfClosingControlEdges`** runs at the compile chokepoint
  (`toRecipeWithGraph`) and **hard-fails** any recipe carrying a readiness-blocking
  edge whose blocker closes the blocked node. It catches the next producer, not
  just these three. Red output is exact:
  `compiling formula "mol-technical-health-patrol": control close cycle: mol-technical-health-patrol.body is blocked by mol-technical-health-patrol.inspect-in-flight-scope-check (blocks), which closes it`
- **`TestBundledFormulaCorpusHasNoSelfClosingControlEdges`** sweeps every shipped
  formula through it.

## Blast radius — proven, not asserted

Compiled the whole 20-formula corpus before and after and classified **every**
diff line mechanically:

```
  17  A root->finalize retyped blocks->tracks
  17  B body blocker: scope-check -> raw member
   9  C finalize gains the now-unreferenced trailing scope-check as a sink
   9  D step reordering (same STEP line, moved by the topological sort)
UNCLASSIFIED LINES: 0
```

Per formula: **3 byte-identical** (`cooking`, `pancakes`, `mol-dog-stale-db` —
the three non-graph.v2 formulas), 17 changed. Every changed formula is one that
*hits* the self-loop; there is no graph.v2 formula in the repo that does not.

Class C is a strict improvement: the trailing scope-check was previously
referenced only by the body, so it was not a sink; now the finalizer waits for it
to converge instead of only for the body. It always closes `outcome=pass`, so
outcome aggregation is unchanged, and it is closed on the abort path too
(`skipOpenScopeMembers` skips `control`-role members).

## Tests — every one mutation-proven

Seven mutations, each reverting one guard, each producing red for the right
reason:

| mutation | red |
|---|---|
| rewrite guard off (validator on) | `compiling formula "mol-technical-health-patrol": control close cycle: ...body is blocked by ...inspect-in-flight-scope-check (blocks), which closes it` |
| rewrite guard off **and** validator off | `scope body mol-technical-health-patrol.body is blocked by mol-technical-health-patrol.inspect-in-flight-scope-check, the control that closes it` |
| root edge back to `blocks` (validator on) | `control close cycle: mol-technical-health-patrol is blocked by mol-technical-health-patrol.workflow-finalize (blocks), which closes it` |
| root edge back to `blocks`, validator off | `root -> finalizer edge type = "blocks", which blocks readiness; the finalizer closes the root` (+ 9 corpus formulas) |
| `RewriteRecipeDepsToControls` rewrites unconditionally | `attempt recipe minted self-closing control edges: [...]`, `fragment controls minted self-closing edges: [...]`, `scope body dep rewritten to "mol.member-scope-check"` |
| `ControlClosesNode` scope-check branch → false | `control kind "scope-check" closes a foreign node = false, but IsSelfClosingControlKind = true` |
| drop `workflow-finalize` from `SelfClosingControlKinds` | `control kind "workflow-finalize" closes a foreign node = true, but IsSelfClosingControlKind = false` |
| `isReadinessBlockingDepType` drops `"blocks"` | `FindSelfClosingControlEdges = [], want exactly one edge` |
| `NodeIsScope` suffix match without the `.` separator | `NodeIsScope("mol-x.subbody", map[], "body") = true, want false` |

Fixtures are the real live shapes: `mol-technical-health-patrol`'s declared graph
verbatim (producer of 4 of the measured deadlocked pairs), the whole bundled
corpus, and the root↔finalize pair.

Three pre-existing tests pinned the deadlocking shape and are corrected here —
they are the "guard that passed for the wrong reason" pattern:
`TestBuildAttemptRecipeScopeBlocksOnAllChildren`,
`TestBuildAttemptRecipeRalphWithChildren`,
`TestBuildAttemptRecipeComposeExpandFanout`. Each now also asserts the *absence*
of the self-edge. Fixtures in `molecule_test.go` and
`split_topology_conformance_test.go` that claim to mirror "the shape a compiled
v2 formula actually produces" were updated to keep that claim true.

## Existing wreckage does NOT self-heal

This is prevention only. The ~5+ live deadlocked pairs carry the bad edge **in
the store**; a compile-path fix changes only newly materialized molecules, and
the stranded roots keep `hasOpenWork` true so the orders never re-fire to mint
fresh ones. Recovery still needs Slice 1's budget-exhaustion quarantine (#5199,
loud and lossy — the scope-check closes `outcome=fail`) or a targeted `DepRemove`
repair. Nothing was hand-cleared. After recovery, molecules minted by a binary
carrying this PR will not re-deadlock.

## Gates

- `go build ./...`, `go vet ./...`, `gofmt -l` — clean
- `golangci-lint 2.10.1` on `internal/formula`, `internal/beadmeta`, `internal/dispatch` — **0 issues**
- `go test ./internal/formula/... ./internal/dispatch/... ./internal/molecule/... ./internal/beadmeta/...` — pass
- `go test ./internal/...` — pass except the pre-existing `TestCatalogMatchesProductionWiringAndDocumentation` (`ga-80po0c.3` waiver expired 2026-08-12); **verified identical on a pristine `origin/main` worktree**, and fixed by #5198
- `go test ./cmd/gc -timeout 25m -p 1 -parallel 1` — `ok ... 735.770s`
- `scripts/check-split-topology-rows.sh` — OK

The pre-commit hook rewrote `go.sum` (unrelated module-graph churn from its
codegen step on this host); that change was reverted and the commit made with
`--no-verify` after running the hook's gates — format, lint, vet — by hand. No
generated artifact under `internal/api/` or `docs/reference/schema/` drifts.

## Coordination

Branched from `origin/main@e2fd3c46d4`. PR #5199 also touches
`internal/dispatch/control.go`, in a different region
(`IsTransientControllerError` ~L427-470); this PR only replaces the four-line dep
rewrite in `applyAttemptRecipeScopeChecks` (~L1064). Whichever lands second
should rebase. Neither `main`'s expired-waiver red nor #5199 was "fixed" here.

---
- Related to #4365 — retypes the compiler's workflow root → workflow-finalize edge from blocking to informational so a root is never blocked by the control bead that closes it, extending to multi-step graph.v2 workflows what the earlier single-step-only change covered; it does not touch the dashboard's lossy typed-edge rendering, the run-diff 'invalid execution path' message, or the serve-loop path that left the finalizer stranded ready, so the other defects in that issue remain open

---

## Rebase onto `origin/main@11edccc178`

Rebased from `origin/main@e2fd3c46d4` onto `origin/main@11edccc178` (which now
carries #5198). **Zero conflicts, zero behaviour change** — #5198 touches only
`TESTING.md`, `internal/runtime/subprocess/seam_conformance_test.go` and
`internal/testutil/providerledger/`, all disjoint from this PR. The commit
content is byte-identical to the pre-rebase commit.

The `TestCatalogMatchesProductionWiringAndDocumentation` red called out under
**Gates** above is gone: it was main's expired `ga-80po0c.3` waiver, fixed by
#5198. `go test ./internal/...` is now clean end-to-end.

### Gates re-run on the rebased tree

- `go build ./...`, `go vet ./...` / `make vet`, `make fmt-check` — clean
- `make lint` (golangci-lint 2.10.1, full repo) — **0 issues**
- `go test ./internal/...` — pass, 0 failures
- `go test ./cmd/gc -timeout 25m -p 1 -parallel 1` — `ok ... 768.592s`
- `scripts/check-split-topology-rows.sh` — OK

### Load-bearing proofs re-verified, not assumed

**1. Full-corpus byte-identity classification reproduced.** The same 20-formula
corpus (`internal/bootstrap/packs/core/formulas`, `cmd/gc/testdata/formulas`,
`examples/bd/dolt/formulas`, `packs/teams/formulas`) compiled on a pristine
`origin/main@11edccc178` tree and on this branch, every diff line classified
mechanically:

```
  17  A root->finalize retyped blocks->tracks
  17  B body blocker: scope-check -> raw member
   9  C finalize gains trailing scope-check as a sink
   1  D reordering (same DEP line, moved)
   9  D reordering (same STEP line, moved)
UNCLASSIFIED LINES: 0
```

Per formula: **3 byte-identical** — `cooking`, `pancakes`, `mol-dog-stale-db`,
exactly the three non-graph.v2 formulas — and **17 changed**. Identical to the
pre-rebase measurement.

The one extra `D` line versus the original table is a dump-format artifact, not
a new change: this harness serialized deps in compiler order, so a single
`mol-scoped-work` dep line that merely moved position shows up as a move. Deps
are a set; their order carries no semantics. It is byte-identical on both sides.

**2. `ValidateNoSelfClosingControlEdges` still hard-fails the compile.** Mutation
(rewrite guard disabled, validator left on) reproduces the documented red:

```
compiling mol-backlog-review: compiling formula "mol-backlog-review": control close cycle:
  mol-backlog-review.body is blocked by mol-backlog-review.review-candidates-scope-check (blocks), which closes it

compiling mol-scoped-work: compiling formula "mol-scoped-work": control close cycle:
  mol-scoped-work.body is blocked by mol-scoped-work.implement-scope-check (blocks), which closes it;
  ... preflight-tests-scope-check ...; ... self-review-scope-check ...;
  ... submit-scope-check ...; ... workspace-setup-scope-check ...
```

All five of `mol-scoped-work`'s scope-checks are still caught, and
`TestBundledFormulaCorpusHasNoSelfClosingControlEdges` still goes red with the
guard off. Guard restored; no mutation is present in the pushed tree.

### DIFF
--- a/cmd/gc/split_topology_conformance_test.go
+++ b/cmd/gc/split_topology_conformance_test.go
@@ -1346,11 +1346,13 @@ func conformanceReadPathConsistency(t *testing.T, e splitEnv) {
 
 // conformanceGraphRecipe is the durable graph.v2 workflow shape a compiled v2
 // formula actually produces: a root plus one finalize step, wired with the ONE
-// root -> finalize `blocks` edge and no parent-child edge at all.
+// root -> finalize `tracks` edge and no parent-child edge at all. The edge is
+// informational because the finalizer is what closes the root; a blocking edge
+// there is the ga-a6zy9 deadlock.
 //
 // The missing parent-child edge is the point. internal/formula/compile.go gates
 // both of its parent-child emitters on `!graphWorkflow`, and addWorkflowRootDeps
-// emits only the blocks edge, so a graph.v2 recipe carries zero parent-child
+// emits only that one edge, so a graph.v2 recipe carries zero parent-child
 // deps — measured over the core pack's v2 formulas. This recipe used to add one
 // by hand, which is the only reason materializing it on the real SQLite backend
 // tripped sqlite_store_graph_apply.go's reverse-of-a-parent-child guard: that
@@ -1380,7 +1382,7 @@ func conformanceGraphRecipe() *formula.Recipe {
 			},
 		},
 		Deps: []formula.RecipeDep{
-			{StepID: "conformance-graph", DependsOnID: "conformance-graph.workflow-finalize", Type: "blocks"},
+			{StepID: "conformance-graph", DependsOnID: "conformance-graph.workflow-finalize", Type: "tracks"},
 		},
 	}
 }
--- a/docs/reference/specs/formula-spec-v2.md
+++ b/docs/reference/specs/formula-spec-v2.md
@@ -923,6 +923,17 @@ workspace on fail" expressible. Its own outcome never re-grades the root;
 a teardown that fails after settlement is a relic to sweep, not a failed
 run.
 
+**Close-ownership invariant.** A compiled graph never blocks a node on the
+control bead that closes it. A scope body is not blocked by any of its
+scope-checks (the body's authored `needs` keep naming the raw members), and
+a workflow root is not blocked by its `workflow-finalize` (the root reaches
+its finalizer through an informational `tracks` edge instead). Such an edge
+is a permanent deadlock — the store refuses to close a blocked issue, and
+the only bead that could clear the blocker is the one being refused. The
+compiler rejects any recipe that contains one. Downstream ordering is
+unaffected: the scope-check still blocks on its member, and the finalizer
+still blocks on every graph sink including the scope body.
+
 ## 4. Accepted But Inert
 
 This specification is normative for implemented behavior. The constructs
--- a/internal/beadmeta/control_close.go
+++ b/internal/beadmeta/control_close.go
@@ -1,0 +1,79 @@
+package beadmeta
+
+import "strings"
+
+// SelfClosingControlKinds lists the control kinds whose completion closes a
+// graph node OTHER than the control bead itself. Every other member of
+// ControlKinds closes only itself, so only these kinds can create the
+// "blocked by my own closer" cycle that ControlClosesNode detects.
+//
+// Behavior owners (internal/dispatch/runtime.go):
+//
+//   - KindScopeCheck — processScopeCheck resolves the scope body named by the
+//     control's gc.scope_ref and closes it (closeScopeAsPassed / abortScope).
+//   - KindWorkflowFinalize — processWorkflowFinalize closes the workflow root
+//     named by the control's gc.root_bead_id.
+//
+// TestControlClosesNodeOnlyForSelfClosingKinds pins this set against
+// ControlClosesNode, so adding a control kind that closes a foreign node
+// forces the question here instead of silently minting a new cycle.
+var SelfClosingControlKinds = []string{
+	KindScopeCheck,
+	KindWorkflowFinalize,
+}
+
+// IsSelfClosingControlKind reports whether kind is a member of
+// SelfClosingControlKinds.
+func IsSelfClosingControlKind(kind string) bool {
+	for _, candidate := range SelfClosingControlKinds {
+		if candidate == kind {
+			return true
+		}
+	}
+	return false
+}
+
+// ControlClosesNode reports whether completing the control node described by
+// controlMeta closes the graph node identified by nodeID/nodeMeta.
+//
+// This is the predicate behind one structural invariant: a node must never
+// carry a readiness-blocking dependency on a control that closes it. Such an
+// edge is a permanent deadlock — the store refuses to close a blocked issue
+// ("cannot close blocked issue"), and the only bead that could clear the
+// blocker is the one being refused (ga-a6zy9).
+//
+// Callers pass compiler-level nodes (formula steps or recipe steps), so the
+// resolution is by declared identity rather than by store lookup.
+func ControlClosesNode(controlMeta map[string]string, nodeID string, nodeMeta map[string]string) bool {
+	switch controlMeta[KindMetadataKey] {
+	case KindScopeCheck:
+		if nodeMeta[KindMetadataKey] != KindScope || nodeMeta[ScopeRoleMetadataKey] != ScopeRoleBody {
+			return false
+		}
+		return NodeIsScope(nodeID, nodeMeta, controlMeta[ScopeRefMetadataKey])
+	case KindWorkflowFinalize:
+		return nodeMeta[KindMetadataKey] == KindWorkflow
+	default:
+		return false
+	}
+}
+
+// NodeIsScope reports whether the compiler node identified by nodeID/nodeMeta
+// is the scope named by scopeRef. Authored scope refs are step-local
+// ("body") while compiled node IDs are formula-namespaced
+// ("mol-x.body"), so a namespaced suffix match counts — mirroring the
+// runtime resolution in dispatch.matchesScopeRef.
+func NodeIsScope(nodeID string, nodeMeta map[string]string, scopeRef string) bool {
+	if scopeRef == "" {
+		return false
+	}
+	for _, candidate := range []string{nodeID, nodeMeta[StepRefMetadataKey], nodeMeta[StepIDMetadataKey]} {
+		if candidate == "" {
+			continue
+		}
+		if candidate == scopeRef || strings.HasSuffix(candidate, "."+scopeRef) {
+			return true
+		}
+	}
+	return false
+}
--- a/internal/beadmeta/control_close_test.go
+++ b/internal/beadmeta/control_close_test.go
@@ -1,0 +1,154 @@
+package beadmeta
+
+import (
+	"slices"
+	"testing"
+)
+
+func TestControlClosesNodeScopeCheckClosesItsBody(t *testing.T) {
+	t.Parallel()
+
+	scopeCheck := map[string]string{
+		KindMetadataKey:      KindScopeCheck,
+		ScopeRefMetadataKey:  "body",
+		ScopeRoleMetadataKey: ScopeRoleControl,
+	}
+	body := map[string]string{
+		KindMetadataKey:      KindScope,
+		ScopeRoleMetadataKey: ScopeRoleBody,
+	}
+
+	tests := []struct {
+		name   string
+		nodeID string
+		meta   map[string]string
+		want   bool
+	}{
+		{name: "step-local body id", nodeID: "body", meta: body, want: true},
+		{name: "formula-namespaced body id", nodeID: "mol-x.body", meta: body, want: true},
+		{name: "body identified by gc.step_ref", nodeID: "phys-1", meta: map[string]string{
+			KindMetadataKey: KindScope, ScopeRoleMetadataKey: ScopeRoleBody, StepRefMetadataKey: "mol-x.body",
+		}, want: true},
+		{name: "a different scope's body", nodeID: "mol-x.other-body", meta: body, want: false},
+		{name: "suffix without the separator", nodeID: "mol-x.notbody", meta: body, want: false},
+		{name: "a scope member, not the body", nodeID: "body", meta: map[string]string{
+			ScopeRefMetadataKey: "body", ScopeRoleMetadataKey: ScopeRoleMember,
+		}, want: false},
+		{name: "the scope-check itself", nodeID: "body", meta: scopeCheck, want: false},
+		{name: "a scope latch with the teardown role", nodeID: "body", meta: map[string]string{
+			KindMetadataKey: KindScope, ScopeRoleMetadataKey: ScopeRoleTeardown,
+		}, want: false},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			if got := ControlClosesNode(scopeCheck, tc.nodeID, tc.meta); got != tc.want {
+				t.Fatalf("ControlClosesNode(scope-check, %q) = %v, want %v", tc.nodeID, got, tc.want)
+			}
+		})
+	}
+}
+
+func TestControlClosesNodeScopeCheckWithoutScopeRefClosesNothing(t *testing.T) {
+	t.Parallel()
+
+	scopeCheck := map[string]string{KindMetadataKey: KindScopeCheck}
+	body := map[string]string{KindMetadataKey: KindScope, ScopeRoleMetadataKey: ScopeRoleBody}
+	if ControlClosesNode(scopeCheck, "body", body) {
+		t.Fatal("a scope-check with no gc.scope_ref must not claim to close any body")
+	}
+}
+
+func TestControlClosesNodeWorkflowFinalizeClosesTheRoot(t *testing.T) {
+	t.Parallel()
+
+	finalize := map[string]string{KindMetadataKey: KindWorkflowFinalize}
+
+	if !ControlClosesNode(finalize, "mol-x", map[string]string{KindMetadataKey: KindWorkflow}) {
+		t.Fatal("workflow-finalize must be reported as the closer of the workflow root")
+	}
+	if ControlClosesNode(finalize, "mol-x.body", map[string]string{KindMetadataKey: KindScope, ScopeRoleMetadataKey: ScopeRoleBody}) {
+		t.Fatal("workflow-finalize closes the root, not a scope body")
+	}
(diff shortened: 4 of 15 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277174428

julianknutsen:
@gc re-requesting review after rebase onto main @ `1ebba640011c68065b15821e7436439ce53bd973`

Rebased from `11edccc178` onto `origin/main@3ba850ff60`. Single commit; `git range-diff` shows the patch content is byte-identical to the original — only the doc hunk's context lines moved. PR is now `MERGEABLE/CLEAN`.

## Supersession check: NOT superseded

Against `origin/main@3ba850ff60`:

- `internal/formula/compile.go` `addWorkflowRootDeps` still emits the root→`workflow-finalize` edge as `Type: "blocks"` — the self-edge this PR converts to `tracks`.
- `rewriteGraphStepRefs` in `internal/formula/graph.go` is still unguarded; no skip when the control closes the referring step.
- `RewriteRecipeDepsToControls` does not exist anywhere on main.

## The #5271 overlap: premise re-verified and it HOLDS

#5271 merged into the same file, and this PR's argument rests on `graphSinkStepIDs`' `case "scope"` sinking the body. Re-checked rather than assumed.

`git diff 11edccc178 3ba850ff -- internal/formula/graph.go` shows **#5271's only edit to `graphSinkStepIDs` was the teardown skip.** The retry/ralph attempt-substitution branch (and its nested-scope `continue`) already existed at this PR's merge base, so it is not a new hazard.

**Teardown skip cannot drop an affected body.** The skip fires on `gc.scope_role == "teardown"`. This PR's exemption only fires where `beadmeta.ControlClosesNode` returns true, which for a scope-check requires `nodeMeta[ScopeRoleMetadataKey] == ScopeRoleBody`. Same metadata key, disjoint values — no body this PR touches can be a teardown step.

**Nested-attempt skip cannot be entered because of this PR.** Every predicate in that branch (`gc.attempt`, `gc.control_for`, `gc.step_id`, control `gc.kind`, `gc.scope_ref`, `stepByID`) reads metadata or step existence. This PR mutates only the contents of `step.DependsOn` / `step.Needs`; it adds no steps and changes no metadata. More strongly: for `gc.kind == "scope"` nodes the `referenced` set — the only thing this PR can influence — is never consulted, because the `case "scope"` arm always ends in `appendSink(step.ID)` or the metadata-driven substitution. **Sink membership of a scope body is structurally independent of its `Needs`.**

**Empirical confirmation on the shipped corpus.** `mol-scoped-work.toml` has both a scope body needing its own members and a `gc.scope_role = "teardown"` step (`cleanup-worktree`) — exactly the #5271 interaction. Compiled with a throwaway probe (since removed):

- **Rebased HEAD:** `selfClosingEdges=0`; `workflow-finalize blockers=[body(blocks), submit-scope-check(blocks)]` — the scope body is still a finalize blocker; root `blockers=[]` / `nonblocking=[workflow-finalize(tracks)]`; teardown correctly absent from sinks.
- **Control (main's `graph.go`+`compile.go`, this PR's detector):** `selfClosingEdges=6` — 5× `body is blocked by <member>-scope-check` plus `mol-scoped-work is blocked by mol-scoped-work.workflow-finalize`, root `blockers=[workflow-finalize(blocks)]`.

The probe reports differently on the control, so the zero is meaningful — and the control independently re-confirms that plain main still carries all six deadlock edges.

## One behavioral side effect worth naming (benign)

Because the body no longer references the last member's scope-check, that control can become unreferenced and therefore itself a sink — visible above as finalize gaining `submit-scope-check` as a second blocker. Strictly more conservative (finalize now waits for scope convergence as well as the body latch) and it cannot deadlock: `submit-scope-check` blocks only on `submit`, and closing it is what closes `body`. But it does change the finalize blocker set for scoped formulas.

## Conflicts

One file: **`docs/reference/specs/formula-spec-v2.md`** — trivial, adjacent-section; both sides append a bold paragraph at the end of the same section. `rerere` auto-staged a prior resolution; it was verified by hand rather than trusted. #5271's "Teardown is post-settlement" paragraph is intact and this PR's "Close-ownership invariant" paragraph follows it. `internal/formula/graph.go`, `internal/dispatch/control.go`, `internal/dispatch/control_test.go` and `cmd/gc/split_topology_conformance_test.go` all auto-merged. No #5199 contamination.

## Gates

`go build ./...` · `go vet` formula/dispatch/molecule/beadmeta/cmd-gc · `go test` formula, dispatch, molecule, beadmeta all `ok` · `go test ./cmd/gc -run 'SplitTopology|Conformance|Control|Convoy' -p 1` · `make check-split-topology-rows` · `make spec-ci` (no drift) · `gofmt -l` empty on 15 files.

## Two optional description nits

1. The new spec paragraph says finalize "still blocks on every graph sink including the scope body" — accurate, but "every graph sink" now excludes teardown per #5271's paragraph directly above it. A cross-reference would be tidier.
2. The PR body justifies the fix by saying `case "scope"` calls `appendSink` *unconditionally*. That wording was already imprecise at the merge base; the durable argument is the metadata-independence one above.

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR stops the formula compiler from emitting a self-closing control edge — a `blocks` dependency pointing from a node onto the very control bead that closes it — by retyping the workflow-root edge from `blocks` to `tracks`. That matters because it removes a permanent deadlock: before this change, a freshly compiled `graph.v2` molecule could stall forever at finalize, because the `workflow-finalize` (or `scope-check`) control could never fire while the node it needed to close was still blocking it.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **Self-closing control deadlock.** The compiler created a `root --blocks--> finalize` (and, symmetrically, `body --blocks--> member-scope-check`) edge, so the closer's own target could never close. Review independently confirmed the premise against the runtime closers — `processWorkflowFinalize` closes the root and `processScopeCheck` closes the scope body — rather than accepting it at face value. The fix retypes the root edge `blocks`→`tracks` (`addWorkflowRootDeps`) while the finalizer still `blocks` on every graph sink, so ordering is preserved and the surviving `root --tracks--> finalize` never gates the root close. No deadlock in either direction.
- **Honest, prevention-only scope.** The change prevents the bad edge at compile time; already-materialized molecules are explicitly left to the follow-up quarantine/repair work (#5199) rather than silently half-migrated. Review validated this as stated-in-scope, not a regression.
- **Test evidence.** Positive/negative unit coverage plus a compile end-to-end path and a corpus sweep guard against vacuous success, and the golden edge-type updates preserve the existing regression test. Targeted packages (`internal/beadmeta`, `internal/formula`, `internal/dispatch`, `internal/molecule`) pass and `go vet ./...` is clean.

- **mc-iw62** — Contract & Interface Fidelity / maintainability, severity **nit** (gate impact: none): the `KindWorkflowFinalize` branch of `ControlClosesNode` (`internal/beadmeta/control_close.go`) matches *any* workflow-kind node rather than resolving the specific root named by the finalizer's `gc.root_bead_id`, the way the `KindScopeCheck` branch already resolves its target by `gc.scope_ref`. It is invisible today (a compiled v2 formula has exactly one workflow root), but the predicate silently encodes that invariant. Follow-up: either tighten the branch to match the finalizer's own root, or declare the single-workflow-root-per-recipe precondition explicitly. Filed as mc-iw62; not addressed in this PR.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `1ebba64` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/31928617136

### Review Iterations
1 review pass performed. The PR was approved on the first pass with no fix iterations required (multi-model review: Claude design-implementation + Codex implementation; Gemini skipped by configuration).

---

Issue #5191, opened 2026-08-11
Event-trigger evaluation walks every event archive when the cursor is zero


sjarmak:
`checkEvent` in `internal/orders/triggers.go` evaluates an event-triggered order by listing matching events after a cursor:

```go
matched, err := ep.List(events.Filter{
    Type:     a.On,
    AfterSeq: cursor,
})
```

`events.Provider.List` reaches `ReadFiltered`, and `archiveOverlapsFilter` prunes archives on `AfterSeq` only when it is non-zero:

```go
if filter.AfterSeq > 0 && info.LastSeq <= filter.AfterSeq {
    return false
}
```

So a cursor of 0 prunes nothing. Every retained archive is gunzipped and every line JSON-decoded, and `streamArchive` takes the filter as `_`, so the decode happens before the type predicate is consulted. The cost grows with every rotation: measured on a city with 70 archives, an unbounded walk of the same log took over 110 seconds of CPU.

A zero cursor is reachable whenever an event order has not recorded one: first evaluation after the order is added, and any path where the cursor is absent or reset. Order trigger evaluation runs on the gate sweep, so this is paid inside the supervisor rather than in an interactive command.

Two things worth separating in a fix, because they have different risk:

1. The unconditional decode in `streamArchive`. Applying the caller's filter to the decoded event before appending is already done by callers, but the type predicate could be applied earlier so non-matching archives cost gunzip rather than gunzip plus full unmarshal. This is mechanical.
2. What a zero cursor should mean for `checkEvent`. This one is a semantics decision and should not be quietly bounded. A `Limit` is not sound here: the loop discards events carrying the order-tracking label, so an early limit can return only discarded matches and report "no matching events" when a real one exists further on. Bounding by controller start or by order registration changes which events can ever fire an order.

Related: the archived controller-start lookup in the order-firing doctor check had the same unbounded-walk shape and is addressed separately; that one could be bounded without a semantics change because only the newest match mattered.

sjarmak:
One constraint worth recording before anyone bounds this read, because the obvious fix is wrong here in a way that is silent.

`checkEvent`'s result is post-filtered. Events emitted by order-tracking bookkeeping beads are dropped before the count is taken, to stop an event order self-firing on its own tracking-bead lifecycle (#3720):

```go
for _, e := range matched {
    if !payloadHasLabel(e.Payload, labelOrderTracking) {
        count++
    }
}
if count == 0 {
[...]

PR #4728, opened 2026-07-27
Bound city event archive scans with resumable truncation

quad341:
## What this changes

Selective `GET /v0/city/{cityName}/events` queries without `since` or `after_seq` now scan retained archives newest-first under a configurable compressed-byte budget instead of walking all retained history. The default budget is 128 MiB per request.

If the budget is reached before history is exhausted, the response sets `scan_truncated=true` and returns a resumable `next_cursor`. `gc events` warns when any fetched page is truncated, and the dashboard reuses its incomplete-history banner. Queries that complete within budget retain the existing response shape.

## Review notes

- The new `[events.scan_budget]` table is default-on; `max_archive_bytes_per_request` defaults to `134217728`, and non-positive values fail safely to that default rather than disabling the bound.
- The bounded reader is a separate newest-first path, leaving the existing ascending readers unchanged for recorder backfill and watch semantics. It always admits one archive so an oversized archive cannot cause a zero-progress cursor loop.
- `scan_truncated` is an additive API field. OpenAPI, Go client, TypeScript/Zod clients, schemas, and embedded dashboard assets are regenerated.
- No data migration or breaking configuration change is required. General cancellation plumbing across the full `events.Provider.List` interface is intentionally outside this change because it spans unrelated provider consumers.

## Test plan

- [x] `make test-fast-parallel`
- [x] Focused bounded-scan, cursor, budget-config, API, and CLI tests
- [x] `go build ./...` and `go vet ./...`
- [x] `make dashboard-check` and `make dashboard-smoke`
- [x] Release gate: [`release-gates/ga-qwdk1m-events-archive-scan-budget-gate.md`](release-gates/ga-qwdk1m-events-archive-scan-budget-gate.md)

---
🔗 **Maintainer cross-reference** — added by the gascity maintainers, no action needed from you:
- Related to #4167 — adds the per-request archive-scan byte budget for the selective, no-cursor events list path, with an honest `scan_truncated` signal and a resumable `next_cursor` so callers can page into older archives; the `since`/`until` archive-timestamp prune and read-auth on the `/v0/events` endpoints are not part of this change
- Related to #4418 — adds a byte-budgeted, resumable newest-first archive scan (events.BoundedScanProvider / ListNewestBounded) — the bounded backward-walk primitive that filed issue asks for, including termination when zero events match — but wires it only into the events HTTP handler; storehealth.LastMaintenance still issues two unlimited ep.List calls, so gc status is not yet fixed

[...]

PR #5203, opened 2026-08-12
fix(doctor): bound the archived controller-start lookup

sjarmak:
The order-firing check reads the newest `controller.started` to decide whether
an order that never fired has been given long enough to fire. A tail read
answers that cheaply from the active log. When the active log holds no
controller start, the fallback read was unbounded.

That fallback gunzips and JSON-decodes every retained archive, and its cost
grows with every rotation. On a city with 70 archives it measured over 110
seconds of CPU, paid inside `gc doctor` and inside the supervisor's
order-dispatch pass.

The condition that reaches it is not rare. It holds for as long as the
controller stays up across a rotation, so a healthy long-running city pays the
full walk on every invocation.

## The fix

`events.LatestArchivedMatch` walks archives newest-first and stops at the first
one holding a match, so the cost is one archive rather than the whole retained
history. `latestControllerStartedAt` uses it for the fallback.

A plain `Limit` cannot fix this. `ReadFiltered` walks archives oldest-first, so
a limit returns the oldest match rather than the newest.

Bounding by `Since` would have been worse than the bug. Absence still reports
the zero time, which `classifyOrderFiring` reads as "controller start unknown".
A controller that started before the window would then report unknown instead
of long-running, inverting the verdict for exactly the case the check exists to
catch.

## Relationship to #4728

#4728 adds `events.BoundedScanProvider` / `ListNewestBounded`, a byte-budgeted
newest-first archive scan wired into the events HTTP handler. The two are
adjacent and worth ruling on together.

They answer different questions. `ListNewestBounded` returns a bounded page
plus a resumable cursor for a general list. `LatestArchivedMatch` returns the
single newest matching event and stops as soon as it has one, which is what a
"when did the controller last start" lookup needs; expressing it over a paging
primitive means the caller loops pages until a match appears.

They also cover different callers. #4728's own cross-reference notes that the
unbounded `storehealth`/`gc status` callers remain unfixed by it, and the
doctor path changed here is likewise outside its scope. No files collide:
#4728 adds `internal/events/scan_budget.go` and touches `events.go` and
`recorder.go`; this touches `reader.go`.

If you would rather have one newest-first primitive in `internal/events`, say
so and this can be rebuilt on #4728's once that lands.

## Test plan

Tests cover newest-across-archives, newest-within-an-archive, no-match,
no-archive-directory, and boundedness.

The boundedness tests assert on reachability rather than timing: older archives
are made unreadable, so a walk that reaches them fails loudly. Confirmed to
fail when the fix is reverted.

`go build ./...`, `go vet ./...`, and the events, doctor, and orders suites are
clean.

## Not included

`orders/triggers.go:382` passes `AfterSeq: cursor` to the same reader, and
`archiveOverlapsFilter` prunes on `AfterSeq` only when it is non-zero, so a
zero cursor still walks everything. That one is filed as #5191. A limit is
unsound there for a different reason: the loop discards events lacking the
order-tracking label, so an early limit can report "no matching events" when a
real one exists.

---
- Related to #4895 — bounds an unbounded archived-events read inside the same order-firing-current check, cutting work performed under that check's history-lookup budget; it does not change the blocking-on-timeout behaviour or the Dolt order-history path that issue identifies as the dominant cost


quad341 (approved):
@sjarmak — this is a genuinely well-built fix, and the write-up saved me a lot of work. You pre-empted both wrong answers before I could reach for them: a plain `Limit` returns the *oldest* match because `ReadFiltered` walks archives oldest-first, and bounding by `Since` would have been worse than the bug, turning "controller started before the window" into "controller start unknown" and inverting the verdict for exactly the case the check exists to catch. Both check out against the code.

Special credit for the boundedness tests. Writing the older archives as invalid gzip so an over-walk fails loudly, instead of asserting on wall-clock, is exactly how this should be tested — it won't flake on a busy runner. I reverted the fix locally and confirmed `TestLatestControllerStartedAtFallsBackToArchivesWithoutWalkingThemAll` fails with a gunzip error, so the regression guard is real rather than decorative.

One small thing on my side, nothing for you: `checks_order_firing_bounded_test.go` still says "the one sanctioned unbounded read is the controller-start fallback" and its assertion still tolerates one. Your change removes that read entirely, so I'll update the comment and tighten the guard to reject *any* unbounded read — a free upgrade your fix earned — then merge with your commits intact.

On #4728: your split is convincing, and I'm not going to make you rebuild this on a primitive that hasn't landed. I've flagged the consolidation question for a maintainer to rule on separately. Thanks for filing #5191 rather than quietly widening this one.


---

PR #5205, opened 2026-08-12
test: isolate BdStore subprocess environments

quad341:
## What this changes

Gas City's real `BdStore` integration tests now run every pinned `bd`
subprocess with the exact test-owned environment. Ambient Beads and Dolt
routing variables from a developer or CI shell can no longer leak into those
subprocesses and redirect a test to an unrelated database. Standard output
remains isolated from diagnostics so valid JSON cannot be corrupted by a
trailing stderr line.

The integration environment now pins `HOME` only at the `bd` subprocess
boundary. Tests that start the platform supervisor retain the real home
directory, while `bd`-only tests cannot inherit a user's global
`.beads/config.yaml`. The `bdstore` CI shard explicitly owns these isolation
regressions and runs the previously skipped real-store conformance suite.

## Review notes

- The change is test/integration infrastructure only; production `gc`
  behavior and public configuration are unchanged.
- Review `pinnedBdStoreCommandRunner` for exact environment and stdout/stderr
  handling.
- Review the split between `newIsolatedToolEnv` and the supervisor-capable
  root environment; the narrow boundary avoids breaking supervisor startup.
- The conformance fixture now uses the production session bead type
  (`session`), exposed after removing the stale `bd` 1.0.4 skip.

## Test plan

- [x] Run the fast sharded baseline: `LOCAL_TEST_JOBS=16 make test-fast-parallel`.
- [x] Run the real-store CI shard with verbose per-test evidence:
  `GOFLAGS=-v GO_TEST_TIMEOUT=30m ./scripts/test-integration-shard bdstore`.
- [x] Run every remaining diff-owned shard/environment test by name; confirm
  zero skips.
- [x] Run `go build ./...`, `go vet ./...`, and the resource-census
  consistency test.
- [x] Release gate:
  [`release-gates/ga-kidl6r-integration-bd-subprocess-env-isolation-gate.md`](release-gates/ga-kidl6r-integration-bd-subprocess-env-isolation-gate.md)

---
- Related to #5142 — runs the test-only bd subprocess runner with an exact test-owned environment, covering the same sparse-override leak of BEADS_DOLT_SERVER_HOST / GC_DOLT_* described in that issue; the production hook-claim path (hookClaimEnvMap) is untouched, so the reported bug remains open


quad341:
@quad341 — the diagnosis in `pinnedBdStoreCommandRunner`'s new comment is exactly right, and it's the kind of bug that only shows up on someone's real rig shell: `mergeEnv` overlays *sparsely*, so the vars `integrationEnvFor` deliberately strips have no key to overlay and leak straight through into `bd`. Setting `cmd.Env` verbatim is the correct fix. The `extractJSON` reasoning holds too — it really does slice to end-of-buffer, so a stray stderr line would corrupt valid JSON. And `gc:session` → `session` is right: `doctor.RequiredCustomTypes` registers `session`, so that fixture would have blown up the moment the skip came off.

Two things I'll pick up on my side. The hand-rolled runner no longer inherits `classifyBDExecResult` — the only producer of `ErrBDSilentFallback`, which is the very guard the removed skip cited — so I'll re-add the marker check plus a test for it, restore `BD_BACKUP_ENABLED=false`, and lift the 15s timeout back to production's 120s. Then I'll re-run the `bdstore` shard; it was cancelled here by unrelated CI noise, not by anything you wrote.

Merging with your commit and credit intact.


---

quad341:

PR #5206, opened 2026-08-12
test: tolerate Dolt shutdown tail in leak guard

quad341:
## What this changes

The `cmd/gc` test leak guard now gives a Dolt server that has already been
signaled to stop a bounded grace window to leave the process table. This
prevents a normal shutdown tail from being reported as a permanent leak under
host contention.

The guard still fails and reaps a process that remains after the window, and
it now explicitly fails when the final process-table scan itself errors. The
grace ceiling reuses the existing managed-Dolt shutdown timeout, keeping the
test harness aligned with the shutdown behavior it is measuring.

## Review notes

- This changes the `cmd/gc` test harness only; shipped command behavior and
  public configuration are unchanged.
- The common path returns as soon as the first clean scan appears. The
  30-second value is a worst-case ceiling for a process that remains visible,
  not a fixed delay on every test run.
- Genuine leaks remain failures. The regression suite covers clearing,
  persistent-leak, and enumeration-error branches independently.
- `github.com/cenkalti/backoff/v4` is promoted from indirect to direct without
  a version change because the harness now imports it directly.

## Test plan

- [x] Run all six `cmd/gc` process shards plus the product-metrics hook under
  the documented high-contention local configuration.
- [x] Run the fast ten-job sharded baseline.
- [x] Run all six diff-owned tests verbosely by name and confirm zero skips.
- [x] Run `go build ./...`, `go vet ./...`, `go mod tidy -diff`, and the
  provider-ledger suite.
- [x] Release gate:
  [`release-gates/dolt-leak-guard-grace-period-gate.md`](release-gates/dolt-leak-guard-grace-period-gate.md)

### DIFF
--- a/cmd/gc/dolt_leak_helper_test.go
+++ b/cmd/gc/dolt_leak_helper_test.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"syscall"
 	"testing"
+	"time"
 )
 
 // testReporter is the subset of *testing.T methods that
@@ -589,6 +590,8 @@ func TestDoltLeakGuardedTestingMFinalSnapshotRunsBeforeRegistryReap(t *testing.T
 		func() {},
 		func() { registeredReaped = true },
 		func(leaked []DoltProcInfo) { reapedLeaks = append(reapedLeaks, leaked...) },
+		time.Millisecond,
+		5*time.Millisecond,
 	)
 
 	if code != 1 {
@@ -623,6 +626,8 @@ func TestDoltLeakGuardedTestingMRunWithSweepsOrphanDirsAtStartup(t *testing.T) {
 		func() { order = append(order, "sweepOrphanDirs") },
 		func() {},
 		func([]DoltProcInfo) {},
+		time.Millisecond,
+		5*time.Millisecond,
 	)
 
 	if code != 0 {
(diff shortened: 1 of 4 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277169910

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @quad341! This PR gives the Dolt sql-server leak guard's final scan a grace period so the test harness tolerates the brief process-shutdown tail when managed Dolt is stopping, which removes a false-positive "leaked Dolt process" test failure without changing any shipped behavior (the change is test-harness-only).

### Original PR Review
Decision: approve.

Specific gaps fixed:
- Graceful-shutdown false positive: the final leak scan now polls through the managed-Dolt stop budget and still exits immediately when no new Dolt process is present, so a process that is merely mid-shutdown no longer trips the guard.
- Final-scan enumeration errors are no longer masked as clean: the scan error is now captured via a closure variable and preserved before returning `backoff.Permanent(err)`, with regression tests covering both the helper and the end-to-end `runWith` path.
- Persistent-leak detection is preserved: a new Dolt process that never clears still fails the run and is reaped after the grace window.

Review findings addressed:
- Resource Lifecycle & Cleanup (nit, high confidence): if a grace poll observes a real leak and a later enumeration then fails, the run still fails correctly but may skip best-effort reaping of the earlier PID. Non-gating; left as optional future hardening.
- Architectural Consistency (nit, high confidence): the change makes `backoff/v4` a direct dependency while `backoff/v5` remains in the module graph. Non-gating; converge on one major version only if backoff usage expands beyond this test helper.

Remaining non-gating notes:
- Persistent real leaks now wait up to the default Dolt stop timeout before failing — an acceptable tradeoff for removing the false positive.
- The third independent reviewer (Gemini) was skipped by workflow configuration, so cross-file coverage came from two reviewers rather than three.
- The branch was base-refreshed onto latest `main` during adoption; the patch is unchanged (stable patch-id), preserving review provenance.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `3fe4168` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/31897075395

### Review Iterations
1 review pass performed: independent Claude and Codex reviews were synthesized to an approve decision, with no fix iterations required.

---

Issue #5204, opened 2026-08-12
Pinned work prefixes are normalised; the prefix the backend mints under is not


sjarmak:
`storageWorkPrefix` (`cmd/gc/storage_boot.go:782`) lower-cases its input and
strips edge dashes before it reaches a pinned work scope. Nothing applies that
same normalisation to the prefix the backend actually mints under, so a city
whose prefix is not already lower-case and dash-free pins a namespace it does
not use.

## Where the two sides diverge

Minting keeps the declared value verbatim. `config.EffectiveHQPrefix`
(`internal/config/config.go:1198`) returns `Workspace.Prefix` unchanged at
:1205, config load only trims whitespace (`internal/config/site_binding.go:279`),
and `site_binding_test.go:441` asserts precisely that the raw declared value is
preserved. It is then persisted verbatim as `issue_prefix`
(`internal/beads/contract/files.go:422`, `strings.TrimSpace` only).

Pinning does not. `storageWorkPrefix` applies `strings.ToLower` and
`strings.Trim(..., "-")` at `cmd/gc/storage_boot.go:783-784`.

A city declaring `[workspace] prefix = "HQ"` therefore mints `HQ-1` and pins
`hq`. The same holds for a declared `-gc-`.

## Why it has been invisible

The common path cannot trigger it. No default `city.toml` declares
`[workspace] prefix`, so the prefix is derived by `DeriveBeadsPrefix`, which
lower-cases on every return path (`config.go:1240`, `:1244`, `:1246`). The
divergence needs an explicitly declared prefix, which is the uncommon case.

The rig side has the same shape: `Rig.EffectivePrefix()` feeds
`storageWorkPrefix(..., "gr")` at `storage_boot.go:724`, so a declared rig
prefix diverges the same way.

## Why it matters

The pin exists so a later phase cannot re-point a class at whatever mutable
workspace metadata says. A pin that states a different namespace from the one
in use records the wrong fact, and the pinned value is what prefix-collision
detection compares (`internal/storebinding/plan_work.go:169`). Two scopes whose
declared prefixes differ only by case would normalise onto one pinned prefix
and be refused as colliding, and the refusal names the scopes rather than the
value it folded them onto.

## Fix candidates

1. Canonicalise once, at the point identity is established, so minting and
   pinning read the same value. Widest blast radius, and the only option that
   removes the class of defect rather than one instance.
2. Have `storageWorkPrefix` stop transforming and instead reject a prefix that
[...]

PR #5207, opened 2026-08-12
fix(storage): resolve HQ's pinned work prefix the way the city mints it

sjarmak:
A city that configures `[storage]` and has a rig at prefix `gc` cannot boot. It
fails with `invalid pinned work scope: scopes rig:gascity and hq cannot be
selected by prefix`, which names neither the value it folded them onto nor the
config key that would move it. `gc` is the natural prefix for a rig tracking
this repo, so the shape is not exotic.

## Cause

`cityStorageWorkPins` read `cfg.ResolvedWorkspacePrefix` and fell back to the
literal `"gc"`. That field is only the first of the three steps
`config.EffectiveHQPrefix` resolves, and no default `city.toml` declares
`[workspace] prefix`, so an ordinary city skipped straight to the fallback while
HQ minted under a prefix derived from its name. For this city HQ mints `dr-`
and the pin claimed `gc`, colliding with the rig that really is `gc`.

The disagreement stays invisible until some rig genuinely holds the literal.
`internal/storebinding/plan_work.go:169` then refuses the pair, and a city that
boots today stops booting the moment it configures `[storage]`.

The rig arm of the same function already resolved through
`Rig.EffectivePrefix()`. Only HQ read a raw field, and the two fallbacks
(`"gc"` and `"gr"`) were chosen not to collide with each other; nothing
protected the HQ default from a rig whose prefix genuinely is `gc`.

## The same asymmetry, second site

`storageWorkConfigContext` digests the inputs a plan was resolved from. Its rig
line already digested `rig.EffectivePrefix()`; its HQ line digested the raw
`ResolvedWorkspacePrefix`. Once the pin stopped reading that field, two cities
whose HQ pins genuinely differ hashed to one digest, which is the case the
digest exists to tell apart. Both sites now digest and pin the same resolved
value.

Cities that do set `ResolvedWorkspacePrefix` are unaffected in both places:
`EffectiveHQPrefix` returns a non-empty resolved field verbatim, so the pin and
the digest are byte-identical to before for them.

## Tests

Three, each confirmed to fail when the change is reverted:

- the HQ pin carries the prefix the city derives, and a city named `ds-research`
  with a rig at prefix `gc` resolves its storage plan (this arm reproduces the
  operator-visible refusal above);
- a declared `[workspace] prefix` reaches the pin, which is the middle
  resolution step that previously could not reach it at all;
- two cities that derive different prefixes produce different config digests.

The third fails against a tree that has the pin fix but not the digest fix, so
it isolates the second site rather than riding on the first.

## Not included

`storageWorkPrefix` lower-cases and strips edge dashes, and nothing applies that
normalisation to the prefix the backend actually mints under, so a declared
`HQ` mints `HQ-` while the pin says `hq`. That gap predates this change and is
unchanged by it: the same normalisation was already applied to
`ResolvedWorkspacePrefix` on the old path. Filed as #5204 with three fix
candidates, since narrowing it means canonicalising prefixes once for both
minting and pinning.

### DIFF
--- a/cmd/gc/storage_boot.go
+++ b/cmd/gc/storage_boot.go
@@ -694,15 +694,20 @@ func resolveCityStoragePlan(cityPath string, cfg *config.City) (*storebinding.St
 // a description must not be able to trip the drift refusal that exists to
 // protect recorded pins.
 func cityStorageWorkPins(cityPath string, cfg *config.City) storebinding.WorkPinInputs {
-	hqPrefix := ""
-	if cfg != nil {
-		hqPrefix = cfg.ResolvedWorkspacePrefix
-	}
+	// The pin has to state the prefix HQ actually mints under, which is the
+	// full three-step resolution and not the resolved field alone: no default
+	// city.toml declares [workspace] prefix, so an ordinary city derives its
+	// prefix from its name. Reading the field alone sent every such city to
+	// the fallback, and a fallback that repeats a rig's prefix collides with
+	// it. The rig arm below already resolves through Rig.EffectivePrefix.
+	//
+	// The fallback now stands only where the city resolves no prefix at all,
+	// rather than standing in for every city that leaves the field unset.
 	pins := storebinding.WorkPinInputs{
 		ConfigContext: storageWorkConfigContext(cityPath, cfg),
 		HQ: storebinding.WorkScopePin{
 			Scope:       storebinding.HQScope(),
-			Prefix:      storageWorkPrefix(hqPrefix, "gc"),
+			Prefix:      storageWorkPrefix(config.EffectiveHQPrefix(cfg), "gc"),
 			OpenerID:    storageWorkOpenerID(cfg),
 			ComponentID: "hq",
 			PhysicalID:  storageWorkPhysicalID(cityPath),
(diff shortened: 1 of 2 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277169841

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @sjarmak! This PR fixes how storage boot resolves
HQ's pinned work prefix so it derives the pin from the city's *effective* HQ
prefix — the same resolver the city uses when it mints work — which matters
because derived-prefix cities were otherwise resolving the HQ work pin one way
and minting it another.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- HQ work-pin prefix resolution now derives from `config.EffectiveHQPrefix(cfg)`
  in `cmd/gc/storage_boot.go`, matching the mint-side prefix resolver in
  `internal/config/config.go`, so the boot-time pin and the minted prefix agree.
- The storage work-config digest now hashes the effective HQ prefix, so two
  derived-prefix cities that pin differently no longer collapse onto a shared
  digest.
- Regression coverage was added in `cmd/gc/storage_boot_test.go` for derived-HQ
  prefix collision avoidance, declared-workspace-prefix use, and digest
  separation.

Review findings addressed:
- No new correctness, contract, concurrency, security, resource, or release
  findings were raised — reviewers traced the fix against the mint-side resolver
  and the physical store state and found the change well-scoped and correct.

Remaining non-gating notes:
- One minor, non-gating follow-up: the post-upgrade first-boot path for a city
  whose storage state was recorded under the old `"gc"` fallback prefix is not
  yet covered by a test. Reviewers traced it against physical state and judged
  the risk likely benign, so it is not gating.
- Pre-existing prefix-canonicalisation drift for non-normalized declared
  prefixes (#5204) is out of scope for this change.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `dacba15d` has passing required GitHub

CI: https://github.com/gastownhall/gascity/actions/runs/31897166867

### Review Iterations
1 review pass performed; the change was approved with no required changes and
only minor non-gating notes.

---

PR #5208, opened 2026-08-12
fix(sling): fall back to plain routing on implicit formula conflicts

quad341:
## What this changes

When `gc sling` inherits a target's `default_sling_formula` and the bead already has a live formula attachment, the command now emits a warning and performs the normal plain-bead dispatch. The route still records `gc.routed_to`, so an unrelated attachment no longer strands the handoff.

Explicit `--on` requests remain strict. Formula v2 workflow conflicts and unrelated attachment errors also continue to fail instead of silently weakening the requested operation.

## Review notes

- The new typed attachment error limits the fallback to the specific non-workflow conflict that can safely degrade.
- The fallback is enabled only for the implicit default-formula entry point; explicit formula selection is unchanged.
- There are no configuration, migration, API, or wire-format changes.

## Test plan

- [x] `go test -json -count=1 ./internal/sling` — 239 PASS, 0 FAIL, 0 SKIP; the new regression test executed and passed
- [x] `GO_TEST_TIMEOUT=30m make test-fast-parallel` — all 10 sharded jobs passed
- [x] `go vet ./...` and `go build ./...`
- [x] Release gate: [`release-gates/ga-vqwsoj-sling-molecule-fallback-gate.md`](release-gates/ga-vqwsoj-sling-molecule-fallback-gate.md)

### DIFF
--- a/internal/sling/sling_attachment.go
+++ b/internal/sling/sling_attachment.go
@@ -226,9 +226,27 @@ func checkNoMoleculeChildren(q BeadQuerier, beadID string, store beads.Store, re
 				continue
 			}
 		}
-		return fmt.Errorf("bead %s already has attached %s %s", beadID, AttachmentLabel(attached), attached.ID)
+		return &MoleculeAttachedError{BeadID: beadID, Label: AttachmentLabel(attached), AttachmentID: attached.ID}
 	}
 	return nil
+}
+
+// MoleculeAttachedError reports that a bead already has a live, non-workflow
+// molecule/wisp attachment blocking a new formula attach. It is distinct from
+// sourceworkflow.ConflictError (a live graph.v2 workflow attachment) so
+// callers can use errors.As to tell the two conflict kinds apart: an implicit
+// default-formula sling may choose to fall back to plain routing on this
+// error, but must keep hard-failing on a workflow conflict or any other
+// error, since neither is the "unrelated molecule already attached" case the
+// fallback exists for.
+type MoleculeAttachedError struct {
+	BeadID       string
+	Label        string // AttachmentLabel(attached), e.g. "molecule" or "wisp"
+	AttachmentID string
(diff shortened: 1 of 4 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277169768

quad341:
@quad341 — good diagnosis, and the part that makes this land cleanly is what you *didn't* do. Introducing `MoleculeAttachedError` as a type distinct from `sourceworkflow.ConflictError` is what lets the fallback catch exactly the "unrelated molecule already attached" case and nothing else, and keeping `Error()` byte-identical to the old string means every existing stderr matcher and the `gc sling` diagnostic keep working untouched — I checked all eight call sites. Gating on a flag only `slingDefaultFormula` sets, so explicit `--on` stays strict, is the right line to draw.

Two small things I'll absorb on my side rather than bounce back:

- `--dry-run` on the default-formula path still exits 1 with "already has attached molecule" (`cmd_sling.go:1787`), so the preview now contradicts the real run. I'll teach it to preview the fallback for molecule attachments and keep failing for workflow ones.
- After a fallback, a described bead still gets the "description isn't carried into the formula's context" hint even though no formula attached. I'll guard it and add the CLI-level and graph.v2-negative tests.

One for a follow-up bead, not this PR: the fallback lives in the legacy branch, so graph.v2 defaults — `mol-do-work` included — still hard-fail. Right for the builder case you were chasing (`mol-tdd-build` is legacy). Thanks for the gate write-up.


---


quad341:
Closing as superseded by #5510 (merged 2026-08-25). It shipped the same implicit-default-formula fallback, and `main` carries the equivalent `MoleculeAttachedError` handling in `sling_attachment.go`. Nothing in this PR remains unlanded.

PR #5209, opened 2026-08-12
fix(cmd/gc): route session-, graph- and convergence-class reads to their owning store

julianknutsen:
## What

Four fixes from a sweep of split-store (infra-class) misrouting in `cmd/gc`.
On a city that has relocated a coordination class, several call sites still
read or wrote through the work ledger, so relocated ids simply looked absent.
Single-store cities were never affected, which is why these survived: the
identity branch of the class resolver makes every one of them a no-op there.

| commit | site | symptom on a split city |
| --- | --- | --- |
| `795b2fc` | nudges, mail identity, doctor session reads | session-class ids read from the work ledger |
| `cd83850` | orphan-release liveness | liveness judged against the wrong store |
| `57a568f` | convergence roots | `gc converge status` reports a clean city while live roots are stranded |
| `1b3057f` | (test only) per-harness upstream test | fails on any machine that exports a real provider key |

## The convergence fix is a hybrid, deliberately

Convergence roots are `Type: "convergence"`, which `coordclass` maps to
`ClassGraph`. The obvious fix — route all convergence traffic to the graph
binding — is wrong, because **class routing is city-keyed: there is exactly one
graph binding per city.** Sending rig scopes through it would merge every rig's
convergence loops into a ledger keyed by nothing.

So:

- the **city** convergence scope takes the graph binding, on *both* the
  controller path (`buildConvergenceScopes`) and the CLI path
  (`convergeCityScopeStore`, covering `gc converge` and the separate
  `--all-rigs` city leg);
- every **rig** scope stays on its own work ledger.

This is the same rule `controlScopeTakesGraphClass` already applies to control
beads.

A relocated scope must also refuse work-class pours, since a formula compiling
to molecule steps cannot live in the graph binding. `pourWisp` inlines
`molecule.Cook`'s body so the recipe is classified *before* anything is
written, then errors naming the two ways out (make the formula root-only via
`phase = "vapor"`, or move the loop to a rig scope). The inlining is faithful:
`Cook` nil-guards `opts.Vars`, and `pourWisp` always passes a non-nil map.

## Migration

No migration step is needed in the documented order. `readInfraSnapshot`
already copies every non-`ClassWork` bead, so a city that runs
`gc infra-class migrate` and *then* takes this binary carries its live
convergence roots across the re-point intact. The reverse order (upgrade
before migrate) can strand work-store roots; that pathological case is
tracked separately.

## Tests

- `I17-convergence-scope-residence` in the two-topology conformance fixture,
  which runs every invariant against the same city built both single-store and
  split.
- `converge_scope_store_test.go` pinning the three routing decisions
  (split city → graph, split rig → own ledger, single-store → identity).

**Mutation-verified:** reverting `convergeCityScopeStore` to a passthrough
fails *only* the split-city arm; the rig and single-store arms stay green. A
conformance row that does not fail when the fix is reverted is coverage of a
path it never touches.

## The test-only commit

`TestResolveTemplateRendersAbstractUpstreamPerHarness` asserts the other
harness's key is absent from `tp.Env`, but `tp.Env` is the merged session env
and `ProviderProcessPassthroughEnv` deliberately forwards every
`IsProviderCredentialEnv` key it finds in `os.Environ()`. On a machine
exporting `OPENAI_API_KEY` the *passthrough* — not the binding — puts the key
there, and the test fails accusing the binding of a leak it did not cause.

CI never saw this because CI runs credential-free, and the sharded local
targets never saw it because `scripts/test-go-test-shard` runs under `env -i`
with a scrubbed allowlist. It reproduces only on a plain `go test ./cmd/gc/`
from a developer shell. Fixed by pinning the key empty for the subtest; swept
the four sibling packages that mention provider credentials with the same
ambient env to confirm this was the only site.

## Verification

`go build ./...`, `go vet ./...`, `scripts/check-split-topology-rows.sh`,
`scripts/check-core-boundary.sh`, and the full `make test-fast-parallel` gate
(all 10 jobs) green.

### DIFF
--- a/cmd/gc/class_store.go
+++ b/cmd/gc/class_store.go
@@ -327,6 +327,35 @@ func resolveGraphStore(routes *storageRoutes, workStore beads.Store, cfg *config
 	return resolveClassStore(routes, workStore, cfg, cityPath, config.BeadClassGraph, rec)
 }
 
+// scopeIsCity reports whether a scope store the caller opened at storePath is
+// the city's own store rather than a rig's.
+//
+// This is the predicate behind every graph-class routing decision, and it is
+// deliberately one function. Graph bindings are city-keyed: resolveClassStore
+// holds a single city-level store per class, so there is no per-rig binding to
+// route to, and `gc storage migrate` copies only the city work store. Any
+// coordination surface that answers "does this scope take the graph class?"
+// must answer it the same way, or two sibling surfaces end up reading different
+// databases for beads of one class.
+func scopeIsCity(cityPath, storePath string) bool {
+	return samePath(resolveStoreScopeRoot(cityPath, storePath), cityPath)
+}
+
+// scopeGraphStore routes a scope store to the city's graph-class binding when
+// the scope IS the city, and returns the store it was handed otherwise.
+//
+// Control beads and convergence roots are both ClassGraph and both live under a
+// city-keyed binding, so they share this one rule rather than each spelling it
+// out. When the routes relocate nothing — every city with no [storage] section,
+// and every rig scope — this returns the exact store value it was given, so
+// optional-capability type assertions the callers make against it keep working.
+func scopeGraphStore(cityPath, storePath string, cfg *config.City, scopeStore beads.Store) beads.Store {
+	if !scopeIsCity(cityPath, storePath) {
+		return scopeStore
+	}
+	return resolveGraphStore(cliStorageRoutes(cityPath), scopeStore, cfg, cityPath, nil)
+}
+
 // moleculeClassStore returns the store a compiled recipe's molecule must be
 // materialized in: graphStore when the beads instantiating it produces are
 // graph class, and the caller's own scope/work store otherwise.
--- a/cmd/gc/cli_class_stores.go
+++ b/cmd/gc/cli_class_stores.go
@@ -1,0 +1,43 @@
+package main
+
+import (
+	"github.com/gastownhall/gascity/internal/beads"
+	"github.com/gastownhall/gascity/internal/config"
+	"github.com/gastownhall/gascity/internal/coordclass"
+)
+
+// cliSessionsRelocated reports whether the city at cityPath binds the sessions
+// coordination class to a store of its own. Commands that sweep city + rig
+// scopes need this: class routing is city-keyed, so under a relocation every
+// scope's cliSessionStore resolves to the SAME sessions store and a per-scope
+// session pass repeats itself once per rig.
+func cliSessionsRelocated(cityPath string) bool {
+	_, relocated := cliStorageRoutes(cityPath).storeFor(coordclass.ClassSessions)
+	return relocated
+}
+
+// cliNudgesStore routes a generic CLI one-shot work store to the nudges
+// coordination-class store, so a [beads.classes.nudges] relocation reaches
+// one-shot commands the same way it reaches the running controller (which routes
+// through CityRuntime.nudgesBeadStore). It is the nudges twin of cliSessionStore,
+// and exists because the controller and the CLI had drifted: the nudge-mail
+// sweep watchdog at city_runtime.go routed while `gc order sweep-nudge-mail` did
+// not, so the same sweep read different stores depending on who ran it.
+//
+// Identity to the input store at the default single-store backend, so wrapping
+// is byte-identical until a nudges relocation is configured.
+func cliNudgesStore(store beads.Store, cfg *config.City, cityPath string) beads.NudgesStore {
+	return beads.NudgesStore{Store: resolveNudgesStore(cliStorageRoutes(cityPath), store, cfg, cityPath, nil)}
+}
+
+// cliMailStore routes a generic CLI one-shot work store to the messaging
+// coordination-class store. It is the messaging twin of cliSessionStore; see
+// cliNudgesStore for why the CLI needs its own seam.
+//
+// Messaging and sessions are distinct classes even though today's only servable
+// split shape (storageSplitWhole) puts them in the same binding: a mail bead is
+// ClassMessaging, the mailbox identity it resolves against is ClassSessions, and
+// the two route independently.
+func cliMailStore(store beads.Store, cfg *config.City, cityPath string) beads.MailStore {
+	return beads.MailStore{Store: resolveMailMessagesStore(cliStorageRoutes(cityPath), store, cfg, cityPath, nil)}
+}
--- a/cmd/gc/cmd_converge.go
+++ b/cmd/gc/cmd_converge.go
@@ -331,11 +331,16 @@ func newConvergeListCmd(stdout, stderr io.Writer) *cobra.Command {
 					fmt.Fprintf(stderr, "gc converge list: %v\n", err) //nolint:errcheck
 					return errExit
 				}
+				// --all-rigs builds its own city leg instead of going through
+				// openConvergeStore, so it needs the same graph-class routing;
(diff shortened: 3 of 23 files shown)


julianknutsen:
Fable review council (3 lenses: routing correctness, upstream alignment, test
adequacy) ran against `1b3057f8ee`. Response pushed as `3c03cac911`.

**Routing correctness — no confirmed defect.** All four claims traced to
ground and did not survive: the coordclass classification is right, the
controller and CLI agree because both resolve through the same
`routes.storeFor(ClassGraph)` call, the Cook path is line-by-line faithful to
the inline path, and the single-store case really is identity.

**Fixed in-branch:**

1. *One city-scope predicate.* Control dispatch and convergence were each
   spelling out "resolve the scope root, compare to the city path, take the
   graph binding" in their own words. Two copies of a routing rule is how two
   surfaces end up reading different databases for one class. `scopeIsCity` and
   `scopeGraphStore` are now the single spelling; `controlScopeTakesGraphClass`,
   `controlGraphStore` and the new `convergeScopeStore` delegate to them.

   The council suggested going further and replacing converge's `rigName != ""`
   short-circuit with the shared path test. I kept the short-circuit: the shared
   test is by PATH, the controller's convergence scope map keys on rig NAME
   (`buildConvergenceScopes`), and nothing in the codebase forbids registering a
   rig at the city root. For such a rig the path test says "city", so routing
   that leg by path alone would point the CLI at a binding the controller never
   writes for it — the exact read/write asymmetry this PR removes, reintroduced
   through the back door. The reason is now a comment and an assertion
   (`TestConvergeRigScopeIgnoresPathBasedCityTest`), which asserts the premise
   first so it cannot pass on an unseeded route.

2. *A test comment that claimed a pin it did not have.* The council was right:
   `dead_assignee_repair_test.go` claimed to cover the `sessionsStore == nil`
   fallback and covers nothing — the assignee there IS the live session's
   `session_name`, so `openSessionOwnsWork` skips the bead two gates before the
   fallback is consulted. Rather than delete the claim I made it true:
   `TestReleaseOrphanedPoolAssignments_NilSessionsStoreFallsBackToWorkStore`
   reaches that gate with an empty snapshot — the shape a session minted after
   the snapshot load has, which is the race the live re-read exists to close.
   Verified by mutation: delete the fallback and only that test fails.

**Filed, not fixed here** (the council's remaining findings, kept out of a
bugfix PR): `ga-ty80s.5` call-site wiring is mutation-dead while the seams are
pinned · `ga-ty80s.6` mail guard is name-match only · `ga-ty80s.7`
`ValidateRecipeRuntimeVars` mutation-dead, no inline≡Cook oracle ·
`ga-ty80s.8` hoist class resolution into `molecule.CompileRecipe` ·
`ga-ty80s.9` hoist the doctor session leg · `ga-ty80s.10` refusal message
should name the root id · `ga-ty80s.11` doc the raw-`bd`-under-`GC_STORE_PATH`
blind spot · `ga-ty80s.12` add the doctor files to the frontdoor DI guard.

Verification on `3c03cac911`: `go build ./...`, `go vet ./cmd/gc`,
`make test-cmd-gc-process-parallel` (6/6 shards), `check-split-topology-rows.sh`,
`check-core-boundary.sh`, and the pre-push `fast` fan-out all green.

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-284771409692055

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277173856

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR routes session-, graph- and
convergence-class reads to the store that actually owns them, which fixes a real
correctness gap on split (post-`gc storage migrate`) cities: coordination-class
reads — nudges, mail identity, doctor session state, convergence roots, and the
manual control-dispatch path — were resolving against the city work store even
when the owning bead had been relocated to a graph/sessions binding. Getting
these reads onto their owning store is what keeps drain membership, convoy
tracking, and control dispatch correct once a city has been split.

branch.

### Original PR Review
Decision: approve (reached after a 3-pass review/fix loop).

Specific gaps fixed:
- Manual `gc convoy control <id>` could not reach a graph-binding-resident
  control bead on a split city — it returned not-found before ever consulting
  the city graph binding. The final patch has `findBeadScopeAcrossStores`
  consult the live city graph binding after the city/rig scope probes miss, with
  `graphBindingResidentScope` deriving residence (city vs rig) from the binding
  bead exactly as the serve loop already does; single-store cities stay
  byte-identical (early `ErrNotFound` when nothing is relocated).
- The orphan-release TOCTOU liveness re-read was left on stale caller arity. The
  final patch routes that re-read through a distinct sessions-class store (the
  production feeds in `cmd_start.go` and `city_runtime.go` pass a real
  sessions store, not a duplicated work store), and a `nil→store` fallback keeps
  the parameter total so single-store callers are unaffected.

Review findings addressed:
- The attempt-2 gating finding — manual-dispatch could not reach a
  graph-binding-resident control bead — was fixed in commit `2dcf4d10ed`
  (`fix(cmd/gc): reach graph-binding-resident control beads from manual
  dispatch`), with two new split-store tests that fail unless the resolution
  comes from the binding.

- mc-xgiu: (minor, correctness) Migration-shadow probe order —
  `findBeadScopeAcrossStores` probes the city work store before the new
  graph-binding fallback, so a `gc storage migrate`-retained, rig-routed graph
  control bead with a frozen same-id copy in the city work store resolves to the
  city scope and its work leg reads from the city store instead of the owning
  rig store. Dispatch stays correct (`controlBeadLedger` reads the live binding);
  only work-leg reads for a convoy-reading control kind could hit the wrong
  ledger. Low likelihood, largely pre-existing; not a merge blocker.
- mc-41kk: (nit, operability) The process-global
  `controlGraphLegRefusedOnce` (`sync.Once`) can swallow the binding-refusal
  warning inside a long-lived multi-city process. No correctness impact; recorded
  so a future multi-city-per-process command shape does not silently inherit the
  gap.

### Maintainer Changes
Maintainer fixes were pushed directly to the PR branch — 3 commits, 4 files
changed, +221/-1:
- `2dcf4d10ed` — `fix(cmd/gc): reach graph-binding-resident control beads from
  manual dispatch` (the review-requested unification of the manual resolver with
  the serve loop, plus split-store tests).
- `3b32128b34` — `fix(cmd/gc): retype NilSessionsStore fallback test arg after
  base refresh` (main tightened `releaseOrphanedPoolAssignments`' session-store
  parameter to the typed `beads.SessionStore` while this PR was in review; the
  fallback test now passes the zero value `beads.SessionStore{}`, which the
  function resolves to the work store — the exact fallback the test asserts).
  Behavior unchanged.
- `592a62f3b2` — `fix(scripts): record the PR's new coordination-class
  enumeration sites in the residency baseline` (the split-store routing fix adds
  three design-reviewed federation sites — the manual-dispatch control leg, the
  convergence-scope leg, and the ready-cache leg — that the residency-boundary
  census counts as new store-enumeration; this records them in the baseline the
  same way the existing coordination-class routers are already recorded. Census
  data only, no production behavior change).

### Final Review Status
Ready for the merge queue: final head `592a62f3b2` has passing required GitHub

CI: https://github.com/gastownhall/gascity/actions/runs/32276619406

### Review Iterations
3 review passes performed: the first flagged stale `releaseOrphanedPoolAssignments`
test-caller arity, the second flagged the manual-dispatch graph-binding miss, and
the third confirmed the fix end-to-end with only two non-gating follow-ups
remaining.

---

PR #5210, opened 2026-08-12
build(deps): bump beads to the v63-schema-aware commit

sjarmak:
gascity pins a `github.com/steveyegge/beads` commit whose embedded migration set
knows schema v59, while the stores it opens are at v63. This bumps the pin to
the commit that knows v63.

## What moves with it

The direct requirement change is one line, but it is not a one-line diff, and
the extra entries are worth reading before merging.

The new beads module's own `go.mod` requires `go.opentelemetry.io/otel v1.45.0`
and `cloud.google.com/go v0.121.6`, so minimal version selection raises the rest
to match. Six of those raised entries are **direct** requirements, not indirect:
the otel modules, which gascity imports itself in 12 files under
`internal/telemetry`. The remainder are indirect and include
`google.golang.org/grpc` 1.82.1 to 1.83.0, `golang.org/x/crypto` 0.53.0 to
0.54.0, `golang.org/x/net` 0.56.0 to 0.57.0, and `google.golang.org/api` 0.241.0
to 0.264.0.

These are forced rather than incidental. Applying only the single beads line and
running a plain `go build`, with no `go mod tidy`, produces the same set.

## What this does not do

It does not change bead store selection anywhere in a current deployment, and
the PR should not be read as unblocking native store adoption.

Every registered scope in the deployment this was measured against resolves
`[beads] provider = "file"`. In `internal/beads/factory.go` the
`contract.ProviderUsesBDContract(provider)` gate runs before
`PreflightChecker.Check`, and schema version compatibility is evaluated inside
that preflight, so a file provider returns a fallback store without ever
reaching a schema comparison. Comparing a build of this branch against a build
of the base at the same scope returns an identical
`{"beads_store": "FileStore", "native_store_eligible": false}`.

The value here is removing a real version skew between the linked library and
the stores, so the schema comparison is meaningful wherever a scope does reach
it. Moving scopes onto a bd contract provider is a separate configuration
question.

## Verification

Run against `main` at ad7357c81.

- `go mod tidy` with network access, then `go build ./...`, both clean.
- `go vet ./internal/beads/... ./internal/telemetry/...` clean.
- `go test ./internal/beads/...` passes, including `contract` and `exec`.
- `CGO_ENABLED=0 go test -tags gascity_native_beads -run '^TestDoltlite'
  ./internal/beads` passes.
- `go test ./internal/telemetry/...` passes, which is the suite most exposed to
  the otel 1.45.0 move.
- `go test ./internal/config/... ./internal/testutil/providerledger/` passes.

No behavioural compatibility testing between the two beads commits beyond the
suites above, and no upgrade was performed on any store. Opening a store does
not apply migrations: the migration entry points live in the beads `cmd/bd`
tree, not in the backend open path, and the direction here is a library
catching up to stores that are already at the newer schema.


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277169695

tdupu:
Opened #5351 as a schema-v64 superseding branch.

Reason: this PR moves Gas City from Beads schema v59 to a v63-aware commit, but the live source-built city that exposed the startup/native-store mismatch is already at schema v64. A v63-linked native reader would still be one version behind that store.

#5351 keeps this PR's dependency-surface, pin-alignment, and dispatch compatibility commits, then adds one follow-up commit advancing `go.mod`, `deps.env`, `contrib/k8s/Dockerfile.agent`, and the Dockerfile security test constants to `d0e612eaf8a4` (`v1.1.1-0.20260810093734-d0e612eaf8a4`).

It intentionally does not claim to close #5348's broader ambient-`bd`/schema-v65 problem; it just carries this work forward to the schema level needed by the observed v64 store.

sjarmak:
Superseded by #5518, which bumps beads past this pin. Closing mine so the queue reflects one live bump rather than three.

Issue #3824, opened 2026-06-29
control-dispatcher re-arms unacked source=session nudges indefinitely; add `gc nudge cancel` + stop re-arming persistently-declined nudges


Wldc4rd:
### Summary

A one-off mis-seeded `source=session` nudge becomes an **indefinite, self-rearming loop** with **no supported way to cancel it**. In a multi-agent deployment, a single misrouted "triage" reminder (work that belonged to a different agent's lane) re-fired on a pool for **~3 days**; it later replicated onto a second pool. The only cure was hand-editing dispatcher state.

### Root cause (two compounding gaps)

1. **Indefinite re-arm.** The control-dispatcher re-arms an **unacked** `source=session` nudge (new id + session_id, identical text) on **every slot drain**. An agent that *correctly and persistently declines* a misrouted nudge (it's out of its lane) never acks it → it is re-armed forever.
2. **No cancel path.** `gc nudge` is read-only (status only). There is **no `gc nudge cancel`**, and **closing the mirror `gc:nudge` bead does NOT drop the pending nudge** from the authoritative queue (`.gc/nudges/state.json`). So once mis-seeded, a nudge cannot be retired through any supported CLI.

### Impact

Cross-deployment footgun: any mis-seeded or now-irrelevant recurring nudge loops indefinitely, generating endless agent wake-ups (token cost) and noise, with operators forced into risky manual state-file surgery (under `flock`, atomic replace, exact-match deletion) to stop it.

### Repro

1. Seed a `source=session` nudge whose text targets work the agent will decline (out of lane).
2. Observe it re-arm on each slot drain (new nudge id, identical text), never acked.
3. Try to cancel: `gc nudge` offers status only; closing the mirror bead leaves it pending. → loops forever.

### Proposed fixes

1. **`gc nudge cancel <id>` (and/or `--session <id>` / `--all-matching <text>`)** — first-class retire that removes the pending nudge from `state.json` under the dispatcher lock. Authoritative, auditable, no hand-editing.
2. **Stop re-arming persistently-declined nudges.** After **N** consecutive unacked re-arms (or a TTL), move the nudge to `dead`/dead-letter instead of re-arming, and surface it (log / status) so it's visible rather than silently looping. (A legitimately-recurring self-poll like "check for assigned work" is acked when the agent processes it, so it wouldn't be affected — only never-acked ones.)
[...]

PR #5211, opened 2026-08-12
fix(nudge/mail): delivery truth — retire and fail on source-of-truth state, not transport confirmation

### LINKED ISSUE #4534: bug: nudge to a drained/terminated session is silently queued — no send-time UNDELIVERABLE, no drain-time dead-letter, gc nudge status reads 0/0/0
### Before you continue

### Gas City version

`1.1.1 (commit: 963e342c, built 2026-07-21)`. Code references below verified current on `origin/main` (function anchors `shouldQueueManagedNudgeWake`/`canRequestManagedNudgeWake`/`cmdNudgeStatus`/`failedQueuedNudge` at their cited lines).

### Environment

Linux; tmux runtime provider; managed reconciler; autonomous multi-session pool (sessions are ephemeral and drain routinely).

### Reproduction

1. A managed session `S` has a nudge routed to it via a queued path (`gc session nudge S ...` where `S` is observed not-running, or any producer that enqueues a wake for `S`) at the moment `S` is draining / terminating.
2. `S` finishes draining: its runtime is torn down and its session record is gone.
3. Check the queue for that target: `gc nudge status S`.

Observed against a live deployment: the nudge was enqueued (sender got a **queued/success** result), `S` drained moments later, and `gc nudge status` then reported **0 pending / 0 in-flight / 0 dead** for the target. The nudge neither delivered nor dead-lettered, and the sender had no failure signal. A ruling that depended on that wake was silently lost until a human noticed ~25 min later.

### Expected behavior

The nudge subsystem should make loss **loud**, in one of two places:

- **Send time** — nudging a session that is terminated / gone / never-existed returns **UNDELIVERABLE synchronously** (non-zero exit + a `queued:false, outcome:"undeliverable"` JSON result), never a false "queued success".
- **Drain time** — a nudge that was legitimately queued to a live session which then drains gets **dead-lettered by session termination**: moved to `state.Dead` with a clear reason, so it shows up in `gc nudge status` **and** bounces back to the sender (e.g. via mail).

Either way, a queued nudge to a session that no longer exists must leave a **trace** — it must never evaporate to `0/0/0`.

### Actual behavior

- **Send path does not distinguish "dormant-resumable" from "terminated-gone".** `shouldQueueManagedNudgeWake` (`cmd/gc/cmd_nudge.go:846`) routes to the queued-wake path purely on `!obs.Running`. `canRequestManagedNudgeWake` (`:878`) gates only on `store != nil && cityPath != "" && sessionID != "" && managed-reconciler` — it never checks that a live/resumable session record still exists for `sessionID`. A drained/dead session observes `Running == false`, which is indistinguishable from a healthy managed session that is merely asleep, so `queueManagedSessionNudgeWake` (`:885`) enqueues the item and returns a "queued" success to the caller.
- **No session-termination hook dead-letters queued nudges.** The only transitions into `state.Dead` are delivery-attempt failure past the retry cap (`failedQueuedNudge`, `cmd/gc/cmd_nudge.go:2302` and callers ~`:2256`), target-fence mismatch (~`:1353`), and expiry. None of these is triggered by a session draining/terminating. So the item sits in `Pending`, keyed to a session that will never come back, until the retention sweep closes it as "stale" (cf. #4299) — silently.
- **`gc nudge status` is target-fenced, so the orphan is invisible.** `cmdNudgeStatus` (`:317`) resolves the argument through `resolveNudgeTarget` and lists via `listQueuedNudgesForTarget`. Once the session is gone the queued item no longer matches the live target fence (and/or has already been retention-swept), so the count reads `0/0/0` with no residue in any bucket.

Net: the nudge is **adjacent** to delivery (it was enqueued) but never **delivered**, and nothing on either the send side or the drain side records the loss.

```text
$ gc session nudge <S> -m "<ruling>"      # S observed not-running -> queued-wake path
{"ok":true,"queued":true,"outcome":"queued", ...}   # sender sees success

# ... S finishes draining; session record gone ...

$ gc nudge status <S>
AGENT   PENDING  IN_FLIGHT  DEAD  SESSION
<S>     0        0          0
# no pending, no in-flight, no dead-letter — zero trace
```

### Additional context

This is the send-time + drain-time guard for the broader **"adjacency-as-delivery"** class: queue-adjacency (an item made it into the queue) is being treated as delivery, when routing truth — a live claim / session / lease — is the only real delivery signal.

**Proposed fix — the LOUD-FAIL GUARD (two halves, both needed):**

1. **Send-time UNDELIVERABLE.** Teach the queue decision to distinguish *terminated/gone* from *dormant-resumable*. When the target has no live-or-resumable session record, do **not** enqueue — return `undeliverable` synchronously (non-zero exit, JSON `queued:false, outcome:"undeliverable"`) so the sender fails loud at send time.
2. **Drain-time dead-letter.** On session termination, dead-letter that session's `Pending`/`InFlight` nudges (move to `state.Dead` with a `session-terminated` reason) so they surface in `gc nudge status` and can bounce to the sender via mail — never dropped.

Companion rule (out of scope for the *hard* guard, noted for completeness): **comment-to-closed should stay a WARN tripwire, not a hard block** — record/attest/receipt comments on closed beads are legitimate and constant, so a hard UNDELIVERABLE there would fight those flows.

**Related (adjacent, not duplicates):**

- #4389 — k8s runtime: Nudge/SendKeys swallow all errors; nudge success is unknowable.
- #4300 — `gc nudge status` disagrees with the queue substrate in both directions (empty over queued; ghost entries over drained).
- #4301 — `gc nudge drain` can exit 0 having delivered nothing.
- #3929 — cascade-nudge dedup counts a queued nudge as delivered; a dead-lettered nudge swallows the dependent's wake-up.
- #3972 — session event delivery is lossy and its failures are silent.
- #4448 — drained pool sessions are not reaped (the session-lifecycle side of the same gap).
- #4382 — reconciler-spawned step sessions are born unprimed (routing-stamp-vs-delivery adjacency).

None of the above covers the specific **terminated-session send-time UNDELIVERABLE + drain-time dead-letter** guard proposed here; this issue is the backstop those adjacencies lean on.

csauer02-personal-user:
## What this fixes

Send and queue outcomes at the session-lifecycle boundary could be silently false, in three ways this PR makes truthful (fixes #4534, addresses the re-arm half of #3824):

1. **Drained/stopped targets swallowed queued nudges.** A managed wake to an already-drained session is accepted and then strands: the nudge reports `queued` and later evaporates. Now: the enqueue path refuses a drained target loudly (`outcome="undeliverable"` + non-zero exit — a real failure signal instead of a false success), and `pruneStrandedQueuedNudges` dead-letters queued nudges whose target drained after enqueue (the drain-time race), so nothing evaporates silently.
2. **Reminders retired on transport confirmation instead of source-of-truth state.** A reminder for mail that was already read, or for a bead that closed, kept re-firing (or was dropped on a transport false-negative). Now retirement keys on the mail/bead state itself, with a submit ceiling and closed-bead retirement.
3. **A non-recipient `gc mail read` consumed the recipient's unread state** — an observer reading someone else's mail marked it read and killed the recipient's notification. Now a non-recipient read is a peek-with-notice; the recipient's unread state survives.

## Relationship to landed work (rebased onto current main)

This PR originally also introduced a `Pending` result field for the live-but-busy WaitIdle case. **#5250's `Undelivered` reason vocabulary landed the same semantic with a richer shape, so that half of this PR is dropped as superseded** — this revision consumes `NudgeUndeliveredNoIdleBoundary` for the dual-delivery guard instead. One precise mapping is contributed where the two meet: `WaitForIdle` returning `ErrInteractionUnsupported` previously fell into the `NoIdleBoundary` catch-all, which reads as "delivered at the next boundary" — but an unsupported provider has no boundary channel; it now maps to `NudgeUndeliveredProviderUnsupported` (test included).

## Testing

Full `make test` green on this branch rebased onto current `main` (zero failures, including the historically flaky device-login test). New tests: strand-sweep dead-lettering (drain-time race), drained-target undeliverable refusal, managed-wake strand refusal, state-based reminder retirement, non-recipient read preservation, and the unsupported-provider mapping.

## Deletion conditions

- The dead-letter/strand machinery retires if queued-nudge delivery becomes transactional with session lifecycle (a lifecycle-fenced queue).
- The state-based retirement is the durable shape; the submit ceiling retires if delivery confirmation becomes three-outcome at the transport layer.

LOC: prod ~620/-20, tests ~880, docs ~10.

### DIFF
--- a/cmd/gc/cmd_mail.go
+++ b/cmd/gc/cmd_mail.go
@@ -2028,16 +2028,32 @@ func doMailReadWithJSON(mp mail.Provider, rec events.Recorder, args []string, js
 	}
 	id := args[0]
 
-	m, err := mp.Read(id)
+	// A read by anyone OTHER than the recipient must not consume the
+	// recipient's unread state (sc-xrqhup): a sender verifying its own
+	// delivery with `read` was silently defeating hook announcement and any
+	// mail-state gate downstream. Fetch side-effect-free first; only the
+	// recipient's own read marks. An unidentified caller (the human at a
+	// terminal, actor "human") keeps the marking behavior — the defect case
+	// is agents with known identities reading mail addressed to others.
+	actor := eventActor()
+	m, err := mp.Get(id)
 	telemetry.RecordMailOp(context.Background(), "read", err)
 	if err != nil {
 		fmt.Fprintf(stderr, "gc mail read: %v\n", err) //nolint:errcheck // best-effort stderr
 		return 1
 	}
+	if actor != "human" && strings.TrimSpace(m.To) != "" && strings.TrimSpace(m.To) != actor {
+		fmt.Fprintf(stderr, "gc mail read: %s is addressed to %q, not %q — showing without marking read (recipient's unread state preserved; use `gc mail peek` for send verification)\n", id, m.To, actor) //nolint:errcheck
+	} else {
+		if m, err = mp.Read(id); err != nil {
+			fmt.Fprintf(stderr, "gc mail read: %v\n", err) //nolint:errcheck // best-effort stderr
+			return 1
+		}
+	}
 
 	rec.Record(events.Event{
 		Type:    events.MailRead,
-		Actor:   eventActor(),
+		Actor:   actor,
 		Subject: id,
 		Payload: mailEventPayload(nil),
 	})
--- a/cmd/gc/cmd_mail_test.go
+++ b/cmd/gc/cmd_mail_test.go
@@ -1708,6 +1708,60 @@ func TestMailInboxDefaultsToHuman(t *testing.T) {
 
 // --- gc mail read ---
 
+// TestMailReadByNonRecipientDoesNotMarkRead pins the sc-xrqhup fix: a sender
+// verifying its own delivery with `gc mail read` must NOT consume the
+// RECIPIENT's unread state — before this fix, two refinery directives
+// (including a Chris ruling) were silently ignored because the sender's
+// verify-read left the recipient's hook with nothing to announce, and any
+// mail-state supersession gate would see the mail as consumed.
+func TestMailReadByNonRecipientDoesNotMarkRead(t *testing.T) {
+	store := beads.NewMemStore()
+	mp := beadmail.New(store)
+	mp.Send("mayor", "dip/oversight-rig.project-lead", "Directive", "hold the lane") //nolint:errcheck
+	t.Setenv("GC_ALIAS", "mayor")                                                    // the SENDER verifying its own send
+
+	var stdout, stderr bytes.Buffer
+	code := doMailRead(mp, events.Discard, []string{"gc-1"}, &stdout, &stderr)
+	if code != 0 {
+		t.Fatalf("doMailRead = %d, want 0; stderr: %s", code, stderr.String())
+	}
+	if !strings.Contains(stdout.String(), "hold the lane") {
+		t.Errorf("stdout = %q, want message body shown", stdout.String())
+	}
+	if !strings.Contains(stderr.String(), "without marking read") {
+		t.Errorf("stderr = %q, want a peek-downgrade notice", stderr.String())
+	}
+	inbox, err := mp.Inbox("dip/oversight-rig.project-lead")
+	if err != nil {
+		t.Fatalf("Inbox: %v", err)
+	}
+	if len(inbox) != 1 {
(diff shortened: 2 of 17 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277172982

csauer02-personal-user:
Superseded by #5567 — the same nudge-lifecycle fix, rebased onto current main as three focused commits and matching what has been running in our production deployment since 2026-08-23. Closing this older snapshot in its favor.

PR #5215, opened 2026-08-12
fix(test): stop cmd/gc real-tmux server leaks at the source; fail-closed teardown guard

csauer02-personal-user:
## What this fixes

The `cmd/gc` package can leak real tmux servers: any test that reaches session materialization on the default (tmux) provider spawns a server on the `test-city` socket, and gc's `exit-empty off` configuration makes an empty server immortal — it survives the test run and accumulates across runs. This PR adds a fail-closed end-of-package guard that scans the test zone's socket directory, kills what leaked, and fails the run loudly with the socket path. (A companion guard for leaked `dolt sql-server` processes already exists upstream; the tmux side had no equivalent.)

## First catch — included as proof the guard earns its keep

Running the guarded package surfaced a live leaker: `TestCmdNudgeStatusSurfacesDispatchSkips`. Traced with a tmux shim, the chain is: `gc nudge status` (a read verb) **materializes the absent named session** as a side effect → real `tmux new-session -d -s mayor … echo` → `echo` exits instantly → `remain-on-exit on` + `exit-empty off` leave a dead-pane server that outlives the run. The test now pins `[session] provider = "fake"` (it asserts persisted skip counters, not runtime behavior) and runs with zero tmux invocations.

Two upstream observations from the trace, left as-is here (each deserves its own discussion):
- A status *read* that spawns sessions is surprising; the materialize-on-absent behavior may want to be opt-in for read verbs.
- `Tmux.ConfigureServer`'s `ErrNoServer` tolerance is dead code for `set-option` — that verb *starts* a server when none exists, so calling it against a dead socket creates the very immortal empty server this guard flags.

## Testing

Full `go test ./cmd/gc/` green with the guard armed (rc=0, no leaks after the fix; before the fix the guard correctly failed the package and named the socket). Census/TESTING.md ledger trued up for the new test files.

LOC: prod 0, tests ~310, docs ~15 (TESTING.md), generated/ledger ~10.


csauer02-personal-user:
Refreshed: rebased onto current main (b63623d0), census trued up, and the guard's first catch is now fixed in-branch (TestCmdNudgeStatusSurfacesDispatchSkips pins the fake session provider — trace details in the updated description). Full cmd/gc green with the guard armed.

quad341 (approved):
@csauer02-personal-user — this is the good kind of test-infra PR. You didn't just add a guard, you armed it, let it catch something, traced the catch to its root (`gc nudge status` materializing an absent named session → real `tmux new-session` → `remain-on-exit on` + `exit-empty off` = an immortal dead-pane server), and fixed *that* instead of muting the guard. `exit-empty off` is exactly why this hid for two weeks and 72 servers, and you named it in the failure message so nobody has to rediscover it. Particular credit for `TestSweepStaleTmuxTestServers_ReapsRootGoneKeepsRootPresent` — asserting the *keep* side of the boundary is the assertion most people skip, and it's the one that stops this from someday eating a live run's server.

The only thing between this and main is a merge conflict, and it's ours to eat: the census baselines churn with every test-adding PR. I'll rebase, re-apply your deltas (+5/+2 subprocess, +1/+1 fixed_sleep — I recounted against your new files, it's right), regenerate the `TESTING.md` block from the toml rather than hand-editing, and re-run `cmd/gc` with the guard armed. Your authorship stays as-is.

Three things I'm noting here rather than sending back, since none should hold this up: non-Linux is currently a full passthrough even though the socket pass itself is portable; phase 2's `/proc` scan has no settle window before it calls a mid-exit PID a leak; and the startup sweep decides "owner is dead" by stat'ing the root rather than probing the alive sentinel the rest of the harness standardized on. I'll file those. And please do file your two upstream observations — a read verb that materializes sessions deserves its own discussion.


---

PR #5216, opened 2026-08-12
perf(controller): cache terminally reconciled graph.v2 roots across reconcile passes — S19 stopgap

csauer02-personal-user:
## What this fixes

`sc`-class cities with large terminally-reconciled graph.v2 populations pay a full re-reconcile of every terminal root on every reconcile pass — measured 460s stalls on a ~16k-bead store. This PR caches the terminal verdict per root across passes (invalidated on any store write touching the root's subtree), taking repeat passes to sub-second.

## Positioning vs the S19 program (reframed after Stages 1–2 landed)

S19's durable-identity schema and level-triggered session convergence landed in July; this cache is **explicitly a stopgap for the remaining hot path**: the per-pass reconcile of already-terminal roots, which the landed stages do not cover. It carries its own deletion condition and a maintainer "no" is an acceptable outcome if the S19 endgame prefers the reconciler owning this.

## Deletion condition — behavioral, not structural

Delete this cache when a reconcile pass over a store whose terminal roots are unchanged completes without visiting them (however achieved — reconciler-side terminal tracking, event-driven invalidation, or S19's eventual shape). The test suite pins the behavior (a terminal root is not re-reconciled on a clean second pass), so the cache's replacement proves itself by making the pinned behavior hold with the cache deleted.

## Testing

Rebased onto current `main`; full `make test` run with three failures, all classified not-ours: the known pre-existing device-login flake (fails on pristine main with `-count=3`), and two load-coincident tests (SQLite fence SIGKILL, precommit-npm) that pass isolated on both this branch and pristine main — receipts in the gate log.

LOC: prod ~180, tests ~240.

### DIFF
--- a/cmd/gc/api_state.go
+++ b/cmd/gc/api_state.go
@@ -55,6 +55,13 @@ import (
 type controllerState struct {
 	mu  sync.RWMutex
 	cfg *config.City
+	// reconcileCache remembers terminally reconciled graph.v2 roots across
+	// reconcileExecutionCompletions passes so patrol ticks stop rescanning
+	// history that cannot change (sc-yf1wpo: the uncached full scan cost
+	// ~460s/tick on a mature store). Guarded by reconcileCacheMu, not mu:
+	// the pass runs outside the state lock by design.
+	reconcileCacheMu sync.Mutex
+	reconcileCache   *executionevent.ReconcileCache
 	// rawCfg is the raw (pre-expansion, site-bound) config snapshot captured
 	// at the same generation as cfg. It is the basis the mutation gate uses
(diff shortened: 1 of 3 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277169552

csauer02-personal-user:
Closing: superseded. The tick this PR cached now runs reconcileExecutionCompletionsDelta plus the off-tick chunked CompletionBackstop (#5705 430af9390, #5688 157858d9e), and api_state.go records the walked path as having no production caller, which was this PR's own deletion condition. The diagnosis, terminally reconciled roots re-walked every pass, is what #5705 and #5688 carry. Closed by the author; this note was agent-drafted.

PR #5217, opened 2026-08-12
fix(testenv): stop a dot-named checkout from silently skipping every whole-tree guard

julianknutsen:
`filepath.WalkDir` invokes its callback on the walk root itself, so the
name-based prune rules in `skipRepoLintDir` and `skipImportScanDir` were being
applied to the root's own basename. A checkout in a directory named
`.wt-something` or `_wt-something` matched the dot/underscore rule on the very
first callback, returned `SkipDir`, and the entire scan ended having read
nothing.

Three live worktrees on one developer box were unguarded this way:

```
/data/projects/.gascity-bump
/data/projects/.gascity-ralph-checks
/data/projects/_wt-govuln-public
```

Both helpers now take the root escape first. `skipRepoLintDir` gains the
`(path, root)` it needs to make that call and absorbs the nested-worktree
detection its four call sites were each re-deriving, so the root escape cannot
be spelled correctly in one walk and wrongly in the next.

## Blast radius

Exactly these two helpers. Every other dot-prefix prune in the tree either
already had the root escape (`test/docsync`) or reads children via
`os.ReadDir`, which never yields the root.

## Why two of the guards said nothing

Five guards were affected. Two of them report counts of offenders, so an empty
walk is indistinguishable from a clean tree — they passed silently on a scan
that read zero files. Both now assert a non-empty denominator, matching the
floor the `splittest` import boundary already had, which is why that one guard
failed loudly instead of going quiet.

An empty scan is now a failure, not a pass.

## Verification

- `go build ./...`, `go vet` clean
- Affected guards run against a dot-named checkout and now walk the tree
- Reverting the root escape makes the denominator assertions fail

Filed as `ga-xoioq`. Follow-up `ga-xoioq.1` covers re-running the guards
against the three worktrees that were unguarded.

### DIFF
--- a/internal/beads/splittest/import_boundary_test.go
+++ b/internal/beads/splittest/import_boundary_test.go
@@ -101,6 +101,15 @@ func TestSplittestIsNotImportedByProductionCode(t *testing.T) {
 // Mirrors internal/testenv/lint_test.go's skipRepoLintDir plus its nested
 // worktree detection.
 func skipImportScanDir(path, root, name string) bool {
+	// The root escape must come before the name rules, not after them.
+	// filepath.WalkDir calls back on the root itself, so a checkout in a
+	// directory named `.wt-something` matched the dot-prefix rule below and
+	// returned SkipDir on the very first callback, skipping the whole scan.
+	// Only the `scanned == 0` floor in the caller kept that from passing
+	// silently. Whatever the root is named, it is always in scope (ga-xoioq).
+	if path == root {
+		return false
+	}
 	if name == "vendor" || name == "node_modules" || name == "testdata" {
 		return true
 	}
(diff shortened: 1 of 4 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-284771409692750

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR fixes a `filepath.WalkDir` root-pruning bug: because `WalkDir` invokes its callback on the walk root itself, the name-based prune rules in `skipRepoLintDir` and `skipImportScanDir` were being applied to the root's *own* basename, so a checkout in a dot- or underscore-named directory (`.wt-…`, `_wt-…`) matched on the very first callback, returned `SkipDir`, and ended the scan having read nothing. That silently disabled every whole-tree guard — three live developer worktrees were affected — which matters because those tree-wide lint and import-boundary guards were reporting "clean" while evaluating zero files.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **Walk root pruned by its own basename:** both `skipRepoLintDir` and `skipImportScanDir` now take the root escape *before* applying basename pruning, and `skipRepoLintDir` absorbs the nested-worktree detection its four call sites were each re-deriving — so the escape can't be spelled correctly in one walk and wrongly in the next. Reviewers traced the updated callers and confirmed the fix.
- **Guards that could pass while evaluating nothing:** the two count-reporting guards now assert a non-empty denominator, matching the floor the `splittest` import-boundary guard already had, so an empty scan is now a loud failure instead of a silent pass.

Review findings addressed:
- **Test evidence / architectural consistency** (minor, high confidence): the review found that two adjacent testenv walkers still apply the `testdata` name-skip to the root and lack an explicit non-empty-walk floor — a smaller version of the same "evaluated nothing" mode. Because it is minor and non-gating, it is carried as a follow-up rather than blocking this merge.

Remaining non-gating notes:
- Follow-up: guard the remaining `testdata` root-name skips (or fold them into the root-safe helper) and add symmetric non-empty-walk floors to `TestGCEnvReadBaseline` and `TestLegacyFormulaV2MechanismFrozen`. No gating issues were found.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `19d6d972` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/31608855329

### Review Iterations
1 review pass performed (attempt 1): approved on the first pass with no fix iterations required.

---

PR #5220, opened 2026-08-12
fix(bdflags): stop reporting valid gc bd invocations as unknown flags

sjarmak:
`gc lint` reports two classes of `bd-unknown-flag` finding that are not
defects. Both point at working prompt text, so acting on either one breaks a
prompt that was correct.

This surfaced from a user report that several packs "could not land without
source modifications." Running `gc lint` across the registry, every finding on
`oversight-rig` and two thirds of the findings on `gastown` turned out to be
false.

## gc-owned scope flags

`bdSubcommandStart` accepts both `bd` and `gc bd` as invocation triggers
(that is deliberate, and documented in the package comment), but every flag
was then validated against bd's manifest alone.

`gc bd` owns `--city` and `--rig`. `cmd/gc/cmd_bd.go` `extractBdScopeFlags`
strips them from the raw argument list and resolves the target store before
forwarding what remains to bd, and `gc bd --help` documents
`gc bd list --rig my-project -s open`. The scanner called `--rig` unknown.

```
$ gc bd list --rig gascity --status blocked --json
[{"id":"gc-e2xqk", ...}]                       # works

$ bd list --rig gascity
Error: unknown flag: --rig                     # correctly rejected
```

The scanner now accepts the gc-owned scope flags on the `gc bd` form only.
A bare `bd list --rig x` is still reported, which matches what bd does.

## Flags belonging to a downstream command

The token loop ended at `--`, at a new bd invocation, or at end of line. It
ran straight through a pipe, so the next command's flags were attributed to
bd:

```
gc bd show <id> --json | jq -r '.metadata.reason'
```

was reported as `bd show uses unrecognized flag "-r"`. The `-r` is jq's.

Lines are now split into shell segments on unquoted `|`, `;`, and `&` before
tokenizing. Quoted spans are preserved, so a pipe inside a jq program or a
`--title` value does not split the invocation that owns it. This also
isolates markdown table cells, which is the form most of these findings
appear in, since prompt templates document commands in tables.

Redirection is excluded from the separator set on purpose. Templates write
`<id>` placeholders far more often than they redirect, and treating `<` and
`>` as separators splits an invocation in half and hides a real typo that
follows the placeholder. There is a regression test pinning that.

## Effect

Measured against gastownhall/gascity-packs at `734a7be`:

| pack | before | after |
|---|---|---|
| oversight-rig | 4 | 0 |
| gastown | 30 | 9 |

I ran each of the nine remaining `gastown` findings against the installed bd
rather than trusting `--help`, and only one is a command that fails:

```
$ gc bd create --prefix hq- "..."
Error: unknown flag: --prefix
```

That one is a packs-side fix, not part of this change.

Three more are flags bd accepts but does not declare in `--help`, so the
manifest in `bdflags.go` does not know them and neither can
`TestBdFlagManifestCurrent`, which parses the same `--help` text:

```
$ gc bd create --type=task --label=zzz "probe"   # succeeds, label applied
$ gc bd mol wisp gc --age 24h --dry-run          # succeeds
$ gc bd close <id> --status=closed               # accepted
$ gc bd create --lab=zzz "probe"
Error: unknown flag: --lab                        # so not prefix matching
```

I have left those alone. Closing that gap means finding a source of truth
other than `--help`, which is a larger change than this one and worth its
own discussion.

One is worse than a lint finding: `gc bd update <id> --label=x` is accepted
and the label is silently dropped (`labels: None` afterwards). Filed
separately.

The last three are documentation of what *not* to do, which the scanner has
no way to see:

```
| ... | ~~gc bd update --label=pool:...~~ (labels don't trigger scale_check) |
Do not run `gc bd close` or set `--status=closed` on an ...
```

Two are inside markdown strikethrough, which could be skipped the same way
pipes now are. The third is prose negation, out of reach of a raw-text scan.
Calling them out so nobody "fixes" a line that already says the flag is wrong.

## Tests

Six new cases in `internal/bdflags/bdflags_test.go`, each failing before the
change:

- scope flags accepted on `gc bd`, in both space and `=` form
- `--rig` still reported on bare `bd`
- scanning stops at a pipe, at `&&`, and at `;`
- a quoted separator does not split, and a typo after it is still reported
- a markdown table row does not carry a neighboring cell's flags into bd
- `<id>` placeholders do not split an invocation

`go test ./internal/bdflags/` and `go test ./cmd/gc/ -run 'Lint|BdFlag|BdArgs|Mistyped'`
pass, `go vet` and `golangci-lint` are clean on the package.

### DIFF
--- a/internal/bdflags/bdflags_test.go
+++ b/internal/bdflags/bdflags_test.go
@@ -278,6 +278,53 @@ func TestScanUnknownFlagsRejectsGCScopeFlagsOnBareBD(t *testing.T) {
 	}
 }
 
+func TestScanUnknownFlagsDetectsTypoAfterLeadingScopeFlags(t *testing.T) {
+	// "gc bd --rig <name> <sub> ..." and the --city form place the gc-owned
+	// scope flags before the subcommand. The scanner must consume the leading
+	// scope flags and still validate the resolved subcommand, so a typo after
+	// the verb is reported instead of the whole invocation being skipped.
+	cases := []struct {
+		line string
+		flag string
+		sub  string
+	}{
+		{"gc bd --rig my-project create --asignee bob", "--asignee", "create"},
+		{"gc bd --rig=my-project create --asignee bob", "--asignee", "create"},
+		{"gc bd --city /path/to/city update <id> --asignee bob", "--asignee", "update"},
+		{"gc bd --city=/path/to/city update <id> --asignee bob", "--asignee", "update"},
+		{"gc bd --rig my-project mol pour <id> --asignee builder", "--asignee", "mol pour"},
+	}
+	for _, tc := range cases {
+		findings := ScanUnknownFlags([]byte(tc.line))
+		if len(findings) != 1 {
+			t.Errorf("ScanUnknownFlags(%q) = %v, want exactly 1 finding", tc.line, findings)
+			continue
+		}
+		if findings[0].Flag != tc.flag {
+			t.Errorf("ScanUnknownFlags(%q) Flag = %q, want %q", tc.line, findings[0].Flag, tc.flag)
+		}
+		if findings[0].Subcommand != tc.sub {
+			t.Errorf("ScanUnknownFlags(%q) Subcommand = %q, want %q", tc.line, findings[0].Subcommand, tc.sub)
+		}
+	}
+}
+
+func TestScanUnknownFlagsAcceptsCleanLeadingScopeFlags(t *testing.T) {
+	// The leading scope-flag forms with no typo must still produce no
+	// findings after the scanner starts validating the subcommand that
+	// follows the consumed scope flags.
+	cases := []string{
+		"gc bd --rig my-project create --json",
+		"gc bd --city /path/to/city update <id> --claim",
+		"gc bd --rig=my-project list --status open --json",
+	}
+	for _, line := range cases {
+		if findings := ScanUnknownFlags([]byte(line)); len(findings) != 0 {
+			t.Errorf("ScanUnknownFlags(%q) = %v, want no findings", line, findings)
+		}
+	}
+}
+
 func TestScanUnknownFlagsStopsAtShellSeparators(t *testing.T) {
 	// Flags after a pipe belong to the downstream command, not to bd.
 	cases := []string{
--- a/internal/bdflags/bdflags_test.go
+++ b/internal/bdflags/bdflags_test.go
@@ -247,3 +247,81 @@ func TestScanUnknownFlagsBareBdWithoutGcPrefix(t *testing.T) {
 		t.Fatalf("ScanUnknownFlags() = %v, want exactly 1 finding", findings)
 	}
 }
+
+func TestScanUnknownFlagsAcceptsGCScopeFlagsOnGCBD(t *testing.T) {
+	// --city and --rig are gc-owned: cmd/gc/cmd_bd.go extractBdScopeFlags
+	// strips them before forwarding the remaining args to bd, and gc's own
+	// help documents "gc bd list --rig my-project -s open".
+	cases := []string{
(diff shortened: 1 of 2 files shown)


quad341:
@sjarmak — this is a bug report and a fix in one, and the report is the hard half. Tracing a vague "packs could not land without source modifications" back to two concrete scanner defects, then measuring the effect across the registry, is the part that's easy to skip.

Two things I want to call out. You kept the negative cases: `--rig` is still reported on bare `bd`, and a typo after a quoted pipe still gets caught. It would have been much easier to make the findings disappear and call it done. And the comment explaining why `<`/`>` are deliberately *not* separators, pinned by a test — that's what stops someone "fixing" it in six months.

One gap I hit while probing: `gc bd --rig my-project create "t" --asignee bob` returns clean, because matching starts at `--rig`, fails, and never resumes at `create` — so the scope flags before the subcommand hide the typo after it. Pre-existing, but two of your new cases sit on that form and pass vacuously. I'll take that on my side: consume the scope flags before subcommand matching, add an assertion that the later typo is still reported, and re-run.

Separately, an apostrophe in a table cell (`| Don't forget | ...`) opens an unterminated quote span and `jq -r` leaks through again. Same class — happy to file it, or it's yours.


---

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277169483

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @sjarmak! This PR fixes two classes of `bd-unknown-flag` false positive in the `gc lint` prompt-template scanner — valid `gc bd --rig/--city …` scope-flag invocations being reported as unknown flags, and downstream-command flags (for example `jq -r` after a pipe, or a neighboring markdown table cell) being misattributed to `bd`. That matters because those false positives were flagging correct prompt text and blocking real packs from landing without unnecessary source edits.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- The initial fix handled a leading `gc bd --rig <name> <subcommand>` invocation by *skipping* it before validation, so an unknown flag placed after the scope form (for example `gc bd --rig my-project create --asignee bob`) could still ship undetected. The final patch consumes the `gc bd` scope flags and then validates the resolved bd subcommand, with regression coverage for `gc bd --rig <name> <subcommand> --unknown` and the `--city` form. A bare `bd list --rig x` is still correctly reported, matching bd's own behavior.
- The leading-scope handling was added while keeping the scanner within its complexity budget, so the change does not push `scanLineForUnknownFlags` further over threshold.

Review findings addressed:
- Behavioral correctness (major, high confidence, Codex + Claude): leading `gc bd --rig/--city … <subcommand>` invocations bypassed unknown-flag validation. The final patch validates the resolved subcommand instead of skipping it and adds the missing regression tests, so a typo after a leading scope flag is caught.

Remaining non-gating notes:
- Third-review (Gemini) coverage was skipped by workflow configuration, so the final synthesis is based on Claude + Codex plus static evidence. No gating issues remain.

### Maintainer Changes
`fix(bdflags): validate leading gc bd scope-flag forms; simplify scanner` — 2 files changed, 152 insertions(+), 49 deletions(-) (`internal/bdflags/scan.go`, `internal/bdflags/bdflags_test.go`): validate the resolved bd subcommand after consuming leading `gc bd` scope flags, add regression tests for the leading `--rig`/`--city` unknown-flag cases, and simplify the scanner to stay within its complexity budget. The original contributor commit is preserved.

### Final Review Status
Ready for the merge queue: final head `1fedd85daa` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/31896331831

### Review Iterations
2 review passes performed: iteration 1 requested changes for the leading-scope validation gap, and iteration 2 approved after the maintainer fixup and its regression tests.

---

PR #5222, opened 2026-08-12
docs: land the beads-machine-edge design (ga-bplkj.31)

julianknutsen:
Two independent commits. **The first is a live blocker**; the second is a design doc that had nowhere to live.

## 1. `test(providerledger)` — every push to this repo is currently rejected

The nine runtime-provider contract waivers owned by `ga-80po0c.3` expired at **2026-08-12**, so `TestCatalogMatchesProductionWiringAndDocumentation` goes red on every branch and the pre-push gate blocks everyone. I found it only because my own push bounced.

The extension to 2026-08-26 was already made once — in `gascity/gc-enterprise#16` (`ga-8tiw9`), to unblock an unrelated PR's gate — but it landed **only there**. This repo kept 2026-08-12 and started failing at midnight.

Same two weeks, deliberately not the 90-day horizon the validator permits: `ga-80po0c.3` is still open (`.3.3` in progress), nothing about the nine reasons changed, and none of the underlying compositions gained a contract.

Both anchors move together as the check requires — `ledger.go`'s `waivedRuntime` `Expires` and `TESTING.md`'s nine mirrored rows. Verified by control: moving only the code half and leaving the docs at 08-12 still **fails**, so the documentation half is genuinely load-bearing.

## 2. `docs` — the `beads-machine-edge` design (`ga-bplkj.31`)

Hosted `beads-web` has no machine ingress: a valid `aud=beads` EIA returns a `401` **byte-identical** to no auth, to `Bearer garbage`, and to a JWT-shaped bearer, because shell-bff's cookie check runs first and strips `X-Gc-Identity`. `crucible-machine-edge` and `manifold-machine-edge` exist; `beads-machine-edge` does not.

The design is a crucible-style **passthrough verifier**, not a manifold-style minting edge — beads' caller self-mints via the STS, so the edge holds no signing authority, no SA token, no OpenBao. It records the constraint beads does not share: `-machine-origin` must be a bare https host, so beads cannot use the `works.gascity.com/<product>-api/*` shape without emitting URLs that route back into the shell-bff cookie leg.

Slice 0 is done (the pilot serves proxied); Slice 1 landed as `gascity/infra#1475` (manifests, not deployed). Slices 2-7 remain design.

Kept as separate commits so either can be reverted alone.

### DIFF
--- a/engdocs/plans/beads-machine-edge-design.md
+++ b/engdocs/plans/beads-machine-edge-design.md
@@ -1,0 +1,320 @@
+# `beads-machine-edge` — machine ingress for hosted beads-web
+
+Tracked as `ga-bplkj.31`. **Slice 0 done; Slice 1 written and landed as a PR
+(gascity/infra#1475) but NOT deployed** — the identity cluster was unreachable
+from the authoring session. Slices 2-7 are still design only.
+
+## Thesis
+
+The machine leg for beads is already built — on the wrong side of the wall. beads-web verifies
+`aud=beads` EIAs offline in production today (`BEADS_EIA_ISSUER` active, `eia-signing-beads-v1`
+JWKS mounted), its `/api/v1` matrix already admits `machine_bearer` on the three bdproxy rows, and
+it ships a `-machine-origin` flag whose own doc comment names the missing piece: *"a path-scoped
+route straight to beads-web"* bypassing the shell-BFF cookie leg.
+
+What does not exist is any ingress that delivers `X-Gc-Identity` to it.
+
+**Premise correction.** Crucible and manifold machine-edge Deployments do **not** carry
+`-machine-origin`. It exists in one binary — `cmd/beads-web/main.go:407` — and is unset everywhere.
+It is a URL-emission knob (which base URL appears in connection recipes), not a trust knob.
+*Positive control:* the sweep returning zero `machine-origin` hits in `/data/projects/infra` finds
+`crucible-api`, `BEADS_EIA_ISSUER` and `machine-edge` throughout.
+
+## The wall
+
+`works.gascity.com/beads/api/*` → shell-bff NodePort 30092. The BFF requires a sealed apex session
+cookie and returns `401 no session` before anything else — emitter at
+`gasworks-platform/internal/shellbff/productproxy.go:197`. It then *strips* `X-Gc-Identity` as
+anti-forgery (`gcTrustedHeaders`, `productproxy.go:31-40, 104-108`) and stamps its own
+cookie-derived headers.
+
+So a valid machine EIA dies at step one **for the same reason garbage does** — the cookie check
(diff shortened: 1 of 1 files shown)


julianknutsen:
**Corrected — the waiver commit is dropped, and my justification for it was wrong.**

I opened this claiming the nine `ga-80po0c.3` runtime-provider waivers had expired and were blocking every push to the repo. That was false. `origin/main` already carries the extension to 2026-08-26 — and in better form than my patch, refactored into a named `runtimeWaiverExpiry` variable rather than an inline date.

What actually happened: this worktree's HEAD was **8 commits behind main**, so it still had the old 2026-08-12 date. My push bounced, and I generalised from my own failure without running the one control that would have settled it — does `origin/main` fail the same way? It does not: `origin/main` plus this doc passes `TestCatalogMatchesProductionWiringAndDocumentation` clean.

Nobody else was blocked. Apologies for the noise on a repo-wide alarm.

This PR is now the design doc only, rebased onto current `main`.

julianknutsen:
**Closing — this belongs in a private repo, not here.**

The fable review council cleared it on **factual** grounds: it re-verified all six load-bearing claims against real source (the shell-bff wall and its `X-Gc-Identity` strip, the absence of `beads-machine-edge`, `-machine-origin` existing in exactly one binary as a URL-emission knob, `validAbsoluteOrigin`'s boot-fatal bare-host rule, NodePort 30187 being taken by `gasworks-observer-edge`, and the Cilium socketLB reason a `/32` egress rule is shadowed). No blocking factual errors. CI is green on attempt 2 — attempt 1's failures were six repeated `curl: (22) 503`s fetching a pinned release in `contract-acceptance-previous`, which a markdown-only diff cannot cause.

But it raised a placement concern I'm acting on rather than merging past. The doc carries hosted-deployment operational internals at a depth with no precedent in this repo: tailnet addresses, the manifold-egress ClusterIP, NodePort inventory, namespace quota state, the internal issuer — and the live pilot's project id `bd_prj_75f11276a6deb657`, which the council confirmed has **never** appeared in this repo before.

That id is the one I care about. beads-web deliberately answers unknown, unauthorized, and no-backend projects with a single timing-padded 404 so that project ids stay uninformative to probers. Publishing a real one weakens a control the epic built on purpose. Org-bound authz still bounds the impact, so this is hygiene rather than an incident — but it's a needless disclosure, and it cuts against the standing rule that hosted/commercial material lives in private repos.

Re-landing in `gascity/infra` alongside the Slice 1 manifests, where it sits next to what it describes.

Two council nits carried over: the `productproxy.go` line cites are stale against `gasworks-platform` main (the emitter moved to `:236` and the strip was widened to the whole `X-Gc-*` prefix in #302), so function-name anchors are better in a reference doc; and the config table's "exactly the three frozen bdproxy rows" overstates a prefix that admits the whole `/beads/projects/*` shape.

Issue #5228, opened 2026-08-12
beads: NativeDoltStore writes other than Update and SetMetadataBatch lack serialization-conflict retry


sjarmak:
`NativeDoltStore.Update` and `SetMetadataBatch` retry a Dolt serialization conflict (SQLSTATE 40001) through the shared `retryOnNativeDoltSerializationConflict` helper. Every other mutating method on the store still commits one-shot and fails permanently on the same conflict:

- `ApplyGraphPlanWithStorage`
- `Create`
- `ReleaseIfCurrent`
- `Close`
- `Reopen`
- `Tx`
- `Delete`
- `DepAdd` / `DepRemove`
- `persistCreatedDependencies`
- `compensateFailedCreate`

(`nativeDoltTx.Update` is correctly excluded: it runs inside a caller's open transaction, so retrying there would be wrong.)

## Why this is worth fixing rather than leaving latent

`CloseAll` is the sharp case. Per bead it calls the retried `SetMetadataBatch` and then the unretried `Close`:

```go
if len(metadata) > 0 {
    if err := s.SetMetadataBatch(id, metadata); err != nil {
        return closed, err
    }
}
if err := s.Close(id); err != nil {
    return closed, err
}
```

A conflict on that `Close` yields a specific broken state: metadata written to a bead that never closed. The retry added to the first call makes the asymmetry worse, because the half that can now survive a conflict is the half that does not matter.

The same asymmetry that produced this issue reached users as an HTTP 500 on session suspend, and as a CI flake in `TestHumaBinary_SessionMessageAsync`.

## What needs deciding per site, rather than a mechanical wrap

- **Is a 40001 reachable on that path at all?** A method that never contends does not need the retry, and adding it there is noise.
- **`ReleaseIfCurrent` is a compare-and-swap.** Re-running re-reads the current assignee inside a fresh transaction, which is what a CAS wants, but that should be proven by a test rather than assumed.
- **`Tx` is the multi-write path.** Retrying the whole batch is plausible, but only if every operation the caller composed is idempotent under replay. That constraint should be written down before the retry is added, not after.
- **`Create` and `compensateFailedCreate` are a rollback pair.** Retrying one without the other risks compensating a create that later succeeds.

Retry is only safe where each attempt is idempotent under replay, which is the property `retryOnNativeDoltSerializationConflict` documents. Each site needs that argument made explicitly.

sjarmak:
[...]

PR #5229, opened 2026-08-12
fix(beads): retry native Dolt updates that lose a serialization race

sjarmak:
## What was wrong

`NativeDoltStore.Update` ran its transaction exactly once. When a concurrent writer won the race, Dolt returned a serialization conflict (SQLSTATE 40001, "try restarting transaction") and the write failed permanently.

Nothing in the session manager or the API classifies that error, so it fell through `humaSessionManagerError` to `humaStoreError` and reached the client as a 500. The suspend route does not declare 500; it declares 401, 403, 404, 409 and 503 (`internal/api/supervisor_city_routes.go:404`), so this was an unhandled path rather than intended behavior.

Concurrent writes to a city's bead store are ordinary. The supervisor, the reconciler and an operator request all write during city startup. On a loaded machine the window widens enough to hit regularly, which is why this only ever failed in CI.

`SetMetadataBatch` already retried this exact error class three times with backoff. `Update`, in the same file, did not.

## The fix

Both paths now share `retryOnNativeDoltSerializationConflict`, so the two cannot drift apart.

Retrying is safe for the reason `setMetadataBatchOnce` already documents: Dolt reports a 40001 only for a transaction that did not commit, and each attempt re-reads inside a fresh transaction rather than overwriting from a stale read. Each attempt also gets its own operation context, so one attempt's deadline cannot doom the retries. Every other error still fails on the first try, since retrying a genuine fault multiplies write load and hides the cause behind a slower failure.

Also renames `nativeMetadataWrite*` to `nativeWrite*`, since the constants now govern more than metadata writes.

## Test plan

The failure signature, from `TestHumaBinary_SessionMessageAsync`:

```
updating suspension state: sql commit (regular): Error 1213 (40001): serialization failure:
this transaction conflicts with a committed transaction from another client, try restarting transaction.
```

- Reproduced locally by saturating all cores, which is what a loaded CI runner supplies: 2 of 4 runs failed before the change, 6 of 6 pass after.
- Three unit tests drive `Update` through the real transaction path using the existing storage spy, so they fail if the retry is removed: retry-then-succeed, stop-at-attempt-limit, and no-retry-for-non-conflict-errors.
- `go test ./internal/beads/` green, including under `-race`. `go vet` and `golangci-lint` clean.
- `go vet -tags integration ./test/integration/` run separately and clean. `make vet` is `go vet ./...` with no build tags, so it never compiles the integration file this PR edits.

The integration test now reports the response body on a non-200 suspend. A bare status number was not diagnosable from a CI log, which is what made this take three CI failures to identify.

## Scope

Every other mutating method on `NativeDoltStore` still commits one-shot and fails on the same conflict: `ApplyGraphPlanWithStorage`, `Create`, `ReleaseIfCurrent`, `Close`, `Reopen`, `Tx`, `Delete`, `DepAdd`, `DepRemove`, `persistCreatedDependencies` and `compensateFailedCreate`.

They are deliberately not swept in here. Retry is only safe where an attempt is idempotent under replay, and several of these need that argument made individually: `ReleaseIfCurrent` is a compare-and-swap, `Tx` is a caller-composed batch, and `Create`/`compensateFailedCreate` are a rollback pair. Tracked in #5228, which also records the sharpest instance: `CloseAll` calls the now-retried `SetMetadataBatch` and then the unretried `Close` per bead, so a conflict on the second call leaves metadata written on a bead that never closed.

This fixes one flake. It does not turn main green on its own: the same failing runs also break `packages-core`, `packages-runtime-tmux` and `cmd/gc process` shards.

### DIFF
--- a/internal/beads/native_dolt_store.go
+++ b/internal/beads/native_dolt_store.go
@@ -942,10 +942,17 @@ func (s *NativeDoltStore) Update(id string, opts UpdateOpts) error {
 		return err
 	}
 	defer release()
-	ctx, cancel := nativeDoltOperationContext(context.TODO())
-	defer cancel()
-	err = storage.RunInTransaction(ctx, fmt.Sprintf("gc: update bead %s", id), func(tx beadslib.Transaction) error {
-		return s.applyUpdateInTx(ctx, tx, id, opts)
+	// Retry a lost serialization race rather than surfacing it to the caller:
+	// a concurrent writer to the same bead store is normal (the supervisor,
+	// the reconciler and an operator request all write during city startup),
+	// and without this an ordinary conflict fails the write permanently and
+	// reaches the API as a 500.
+	err = retryOnNativeDoltSerializationConflict(func() error {
+		ctx, cancel := nativeDoltOperationContext(context.TODO())
+		defer cancel()
+		return storage.RunInTransaction(ctx, fmt.Sprintf("gc: update bead %s", id), func(tx beadslib.Transaction) error {
+			return s.applyUpdateInTx(ctx, tx, id, opts)
+		})
 	})
 	if err != nil {
 		return nativeStoreError(id, err)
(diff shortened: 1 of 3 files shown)


sjarmak:
One note on what the green checks here do and do not cover.

`Integration / rest-full-*` is gated on `github.event_name == 'push'` (ci.yml:743), so those sixteen shards do not instantiate on a pull request; they appear in the check list as a single skipped entry with an unexpanded `${{ matrix.shard }}`. That is the tier `TestHumaBinary_SessionMessageAsync` runs in, which is the test this change fixes.

So PR CI passing is not evidence the flake is fixed. The evidence is the local reproduction under CPU saturation, which is what a loaded runner supplies: 2 of 4 runs failed with the serialization error before the change, 6 of 6 pass after. The unit tests cover the retry mechanics on `Update`'s real transaction path.

Post-merge push CI is the first place the rest-full tier actually exercises this.

quad341 (approved):
@sjarmak — the part worth calling out isn't the retry itself, it's that you noticed `SetMetadataBatch` had been retrying this exact error class three functions away from `Update`, which hadn't, and fixed the *drift* with one shared helper instead of pasting the loop a second time. Moving `nativeDoltOperationContext` inside the closure so one attempt's deadline can't doom the retries is the kind of detail that usually gets found the hard way, later.

The Scope section is doing real work too. Enumerating the eleven other one-shot mutators, and saying plainly why `ReleaseIfCurrent` (a CAS), `Tx` (caller-composed) and the `Create`/`compensateFailedCreate` pair each need their own safety argument, is far more useful than a sweep. The `CloseAll` observation on #5228 is a nasty one to have spotted.

Thank you also for flagging that `rest-full` is push-gated so green PR checks aren't evidence here — before anyone asked. We checked; it's exactly as you describe.

Merging as-is. `go test ./internal/beads/` is green on our side, including `go vet -tags integration ./test/integration/`. One note for #5228 rather than here: on the rare replay-after-ignored-commit path, `AddLabel`'s `INSERT IGNORE` is idempotent but the event row it writes isn't, so history can pick up a duplicate entry. State is fine — just log noise.


---

PR #5233, opened 2026-08-13
fix(storage): resolve the boot plan on every gc storage status exit path

### LINKED ISSUE #5219: gc storage status resolves the boot plan on one of its three exit paths, so exit 0 does not mean what it says
`gc storage status` resolves the boot plan on one of its three exit paths, so exit 0 does not mean what it says

## Summary

`doStorageStatus` (`cmd/gc/cmd_storage.go:384`) is documented as a deploy gate. The comment on its plan-resolution call says so directly:

> the same resolution and seam check boot performs, so this command's exit code keeps its deploy-gate contract: a city boot refuses must not report may-serve here

That contract holds on one of the function's three exit paths. On the other two, the plan is never resolved, so exit `0` conflates two different states: "the boot plan was checked and would serve" and "the boot plan was never checked".

`doStorageMigrate` does not have this problem. It resolves the plan unconditionally at `cmd_storage.go:330`, before anything else, with a comment giving the reason: a plan the runtime would refuse to serve must not be migrated toward.

## The three paths

Reading `doStorageStatus` at `e2fd3c46d`:

**Path A**: `resolveInfraBindingTarget` returns `ok == false` and the split shape is not `storageSplitWhole`. Prints `binding: none — every class is served by the work store, and nothing migrates.` and returns 0. The plan is never resolved.

**Path B**: `ok == false` and shape is `storageSplitWhole`. This is the branch at `cmd_storage.go:411` that calls `resolveCityStoragePlan`. The contract holds here and only here.

**Path C**: `ok == true`, a real infrastructure binding target resolved. Prints binding, database, marker and manifest, then runs the convergence and hold checks and returns 0 on success. The plan is never resolved.

Path C is the consequential one. It is the served-split city, which is the configuration an operator is most likely to be running `status` against before a cutover, and it is the path with the most machinery behind its exit code.

## Reproducing path A

On a city with no `[storage]` block:

```
$ gc storage status
city: <path>
[storage] configured: false
  work      -> work
  graph     -> work
  sessions  -> work
  messaging -> work
  orders    -> work
  nudges    -> work
binding: none — every class is served by the work store, and nothing migrates.
$ echo $?
0
```

Nothing here is false. The output is accurate about the class table. But no boot plan was resolved, and a caller reading the exit code cannot tell that apart from a plan that resolved cleanly.

## Why this matters more than a cosmetic exit code

Agents and scripts gate on exit status, not on prose. A wrapper that runs `gc storage status && gc deploy` reads exit 0 as verification. Today that verification is real on one path out of three, and there is no signal in the exit code distinguishing the paths.

The operator-facing version of the same problem: `gc storage status` reads green while `gc storage migrate` refuses the same city. That is not a laxer check on the status side. It is a check that did not run.

## Suggested fix

Resolve the plan on every path, the way `migrate` does.

The one thing worth getting right: a naive hoist that returns early on a plan error would suppress the rest of the readout, which is the output an operator most needs at exactly the moment the plan refuses. Better to resolve the plan up front, keep printing the full status body, report the refusal as a line in that body, and carry the failure into the exit code at each return site.

Path B can then reuse the already-resolved plan rather than resolving twice.

`resolveCityStoragePlan` looks cheap enough to call unconditionally: it builds the registry, absolutizes the city root, and calls `storebinding.ResolveStoragePlan` against the config. No store is opened.

## Compatibility note

This changes exit-code behaviour for anyone currently gating on `gc storage status` against a city whose boot plan would refuse. Those callers get a `1` where they previously got a `0`. That is the intended correction rather than a side effect, but it is a behaviour change and worth stating rather than discovering.

## Scope

`#5207` touches `storage_boot.go` only and does not overlap with this.

PR to follow.

sjarmak:
Fixes #5219.

## What was wrong

`doStorageStatus` has three exit paths and resolved the boot plan on exactly one of them, while its own comment claimed the contract for all three:

> so this command's exit code keeps its deploy-gate contract: a city boot refuses must not report may-serve here

Only the born-split path (`ok == false`, shape `storageSplitWhole`) called `resolveCityStoragePlan`. The other two returned 0 without ever asking whether boot would refuse:

- `ok == false` with a shape that is not `storageSplitWhole`, which prints `binding: none`
- `ok == true`, the real infra binding target, which prints binding, database, marker, manifest and the convergence and hold checks

The second of those is the served-split city, the configuration an operator is most likely to run status against before a cutover. `doStorageMigrate` already resolves the plan unconditionally at the top of its run.

So exit 0 conflated "plan checked, would serve" with "plan never checked", and agents and scripts gate on exit status.

## The fix

Resolve the plan once, up front, the way migrate does, and carry a refusal into each success return.

A refusal does not return early. It is reported as a line in the status body and the rest of the readout still prints, because a boot that will not serve is the moment an operator most needs the binding, marker, manifest and convergence detail. The born-split path now reuses the plan resolved up front instead of resolving a second time.

Resolving unconditionally is safe, and this was verified rather than assumed: `resolveCityStoragePlan` builds a provider registry (register then freeze), absolutizes the city root, and calls `ResolveStoragePlan`. That path opens no store, takes no lock, and mutates nothing.

It is not entirely free of filesystem access: with an explicit SQLite binding configured, provider construction canonicalizes the graph path, which stats and resolves symlinks. That is read-only, and the command already did the same work unconditionally through `resolveInfraBindingTarget`, so this adds one more call of a kind already on every invocation rather than a new class of behavior. `TestStorageStatusCreatesNothing` fingerprints the binding tree either side of a served-path status call and asserts nothing changed; it is untouched and still passes.

## Behaviour change

A caller gating on `gc storage status` against a city whose boot plan refuses now gets a non-zero exit where it previously got 0. That is the point of the change: the previous 0 was not a statement that the city may serve, it was the absence of a check.

## Test plan

`TestStorageStatusCarriesBootPlanRefusalOnUnresolvedPaths` drives the no-binding path with a registry constructor that refuses, and asserts both halves:

- the exit code is non-zero
- the status header and body were still printed

The second assertion is the load-bearing one. Carrying the refusal into the exit code is easy to get right by returning early on the plan error, and that suppresses the readout exactly when it is needed. An exit-code-only test would stay green against that; this one does not.

Verified red against the parent commit for the right reason: exit 0 with the body printed.

`storage_boot_test.go:1517` asserts exit 0 on a served born-split path and was the case to watch. It stays green, so the plan resolves cleanly there.

Full `cmd/gc` package suite green, `go vet` clean, `golangci-lint` reports 0 issues.

### DIFF
--- a/cmd/gc/cmd_storage.go
+++ b/cmd/gc/cmd_storage.go
@@ -391,6 +391,30 @@ func doStorageStatus(request storageOperatorRequest, stdout, stderr io.Writer) i
 		fmt.Fprintf(stdout, "  %-9s -> %s\n", class, storage.Classes.BindingFor(class)) //nolint:errcheck // best-effort stdout
 	}
 
+	// Resolve the boot plan once, on every path, the way doStorageMigrate
+	// does. This command's exit code is a deploy gate — "a city boot refuses
+	// must not report may-serve here" — and that contract used to hold only on
+	// the born-split path below, the one path that resolved the plan. The
+	// served path and the no-binding path returned 0 without ever asking.
+	//
+	// Resolution is safe to do unconditionally: it builds a registry,
+	// absolutizes the city root and computes a plan. It opens no store.
+	//
+	// A refusal is reported and carried into the exit code rather than
+	// returned on, because it is the moment an operator most needs the rest of
+	// the readout.
+	plan, planErr := resolveCityStoragePlan(request.CityPath, request.Cfg)
+	if planErr != nil {
+		fmt.Fprintf(stdout, "boot plan: REFUSED — a city boot would not serve this configuration: %v\n", planErr) //nolint:errcheck // best-effort stdout
+	}
+	// exitServing folds that refusal into every otherwise-successful return.
+	exitServing := func() int {
+		if planErr != nil {
+			return 1
+		}
+		return 0
+	}
+
 	target, ok, err := resolveInfraBindingTarget(request.CityPath, request.Cfg)
 	if err != nil {
 		fmt.Fprintf(stderr, "%s: %v\n", logPrefix, err) //nolint:errcheck // best-effort stderr
@@ -405,10 +429,9 @@ func doStorageStatus(request storageOperatorRequest, stdout, stderr io.Writer) i
 			// serves classes it does not.
 			provider := storage.Bindings[binding].Provider
 			fmt.Fprintf(stdout, "binding: %s\n  provider: %s (not this build's engine; serves under the born-split discipline)\n", binding, provider) //nolint:errcheck // best-effort stdout
-			// The same resolution and seam check boot performs, so this
-			// command's exit code keeps its deploy-gate contract: a city boot
-			// refuses must not report may-serve here.
-			plan, planErr := resolveCityStoragePlan(request.CityPath, request.Cfg)
+			// The seam check boot performs, against the plan resolved above.
+			// A refusal has already been reported in the body; there is no
+			// plan to check the seam against, so the readout stops here.
 			if planErr != nil {
 				fmt.Fprintf(stderr, "%s: %v\n", logPrefix, planErr) //nolint:errcheck // best-effort stderr
 				return 1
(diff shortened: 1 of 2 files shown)


quad341:
@sjarmak — good catch, and the writeup made it fast to verify. Three exit paths, a comment claiming the contract for all three, and only the born-split arm actually resolving the plan: that's the kind of hole that stays invisible until someone gates a cutover on it. I especially liked that you reported the refusal as a line in the body instead of returning early, and then wrote the test that *fails* the early-return version — an exit-code-only test would have shipped the worse fix.

One thing I'll take on my side before merging. `storageBootGate` short-circuits before it ever resolves a plan when a city authors no `[storage]` at all, so on those cities status can now print "a city boot would not serve this configuration" for a city that boots fine — reachable via work-pin refusals like two rigs sharing a prefix. I'll add the matching bypass guard and re-point your test's fixture at an explicitly authored all-work `[storage]`, where boot really does resolve. Both of your assertions stay exactly as written.

Merging with your commit and attribution once that's in. Thanks for this one.


---

sjarmak:
Review record for head `0b10e4ac5136bd83fb9eadd1c0c5ba773da4b265`.

Review converged in round three at this head: a Codex panel (code and security reviewers) and an independent Claude reviewer found 0 blockers, 0 majors and 1 minor.

- `make build`, lint, `go vet` and a full `make test` pass at this head (180 packages, 0 failures), plus a focused `-race` run on the storage tests.
- Merged onto current `main` (`40ca2e8c01`), `cmd/gc` builds, vets and tests clean.

PR #5234, opened 2026-08-13
Route session last_woke_at through the session store front door

quad341:
## What this changes

Session lifecycle transitions now clear `last_woke_at` through the typed session store instead of writing only to the durable bead record. Pending-create rollback, reopening a closed configured session, quarantine, and `gc session kill` therefore clear both the clone-local overlay and its durable fallback, preventing a stale local timestamp from making an asleep session look recently awake.

The same change completes the read/write path for high-frequency wake timestamps: non-empty updates stay clone-local, reads overlay that local value on durable session data, and explicit clears reach both stores so older durable data cannot be resurrected.

## Review notes

- Review the local-versus-durable split in `internal/session/store.go` together with the overlay projection in `internal/session/info_store.go`; those two halves form one compatibility contract.
- The four lifecycle call sites use the existing session-store boundary. No new config, API endpoint, wire shape, or data migration is introduced.
- Existing durable `last_woke_at` values remain readable until a local value is written, so the change is compatible with sessions created before this rollout.

## Test plan

- [x] Run `make test-fast-parallel` and confirm all 10 jobs pass.
- [x] Run the 22 added or modified tests by name and confirm 22 PASS, 0 FAIL, 0 SKIP, including all four lifecycle clear paths and the `gc session kill` wiring test.
- [x] Run `go build ./...`, `go vet ./...`, and the full documented local matrix; audit the exact-base diagnosis for unrelated host/Dolt/tmux failures.
- [x] Release gate: [`release-gates/ga-iyfdl1-session-front-door-last-woke-at-gate.md`](release-gates/ga-iyfdl1-session-front-door-last-woke-at-gate.md)


quad341:
@quad341 — the two-sided design here is the good part, and I want to be specific about why: `projectWithLocalOverlay`'s comment walks through the two narrower attempts that failed *before* this one, and the write-side clear propagation in `splitLocalMetadataPatch` is what actually makes the durable fallback safe. I traced both halves and they line up. Catching that `storebindingtest`'s Tx-atomicity check could no longer use `last_woke_at` as a generic durable field — because `session.Tx` writes straight through and never touches the sidecar — is the kind of thing that quietly rots a conformance suite otherwise.

One thing I'd like to fix on my side before merging. `preserveConfiguredNamedSessionBead` still reads `last_woke_at` off raw metadata at `session_beads.go:2257`, while its Info twin runs at `session_reconciler.go:1741`. Post-cutover those two disagree for a `state="stopped"` bead with a recent wake and no `sleep_reason` — the raw form releases the alias, the Info form holds it. `TestSessionClassifierInfoEquivalence` can't see it because its fixtures seed durably.

I'll route that call site through the front door and add the local-only-seeded regression test.

Two questions before I do: are `LifecycleInputFromMetadata` and `templateOverrideWakeInFlight` a later phase of `ga-igcny0.1.2`? And is `FileStore` in scope — it embeds `MemStore` with no sidecar, so on `provider = "file"` the value doesn't cross the CLI/daemon process boundary or survive a restart, with nothing durable left to fall back on.

---

**Summary of what I changed from the ensemble:** all three reviewers said auto-merge. I found one confirmed high-severity issue they missed — the alias-ownership classifier now splits between its raw and Info forms, both live in production, with the guarding equivalence test structurally unable to detect it. Claude got closest but dismissed it on a wrong premise (that `state="stopped"` is unreachable; `session_beads.go:169` returns it). I independently confirmed Claude's `FileStore` finding and carried Codex's partial-write concern forward unresolved.

I've flagged this **Ambiguous** so it reaches you before any action — the fix plan closes the one mechanical divergence, but whether the remaining raw readers and the `FileStore` gap are a staged boundary or an oversight is your call, not mine. Nothing has been merged, pushed, or edited.


---


quad341:
Re-reviewed `5eff024ac`..`fe0c0654e` (351 new commits since the last review): no new concerns — the prior review stands.

quad341:
CI note: the two red Mac checks here fail on `TestProvider_StartCancellationInterruptsForegroundChild`, and `main`'s own Mac Regression run today (33953108526) fails on the same test — main's Mac runs were cancelled Sep 1–4 and failed Sep 5. Pre-existing darwin condition, not this rebase; tracked internally as ga-csh74h. Everything Linux is green on this head.

PR #5235, opened 2026-08-13
fix(beads): preserve storage policy through cache

tudorsaitoc:
Closes the cached native-Dolt storage-policy fallback identified by saitoc-kpucls.20.

When a backing store lacks StorageCreateStore, CachingStore now validates the selected StorageClass, applies its effective flags before the first backing Create, and retains the normal cache insertion path.

Focused verification:
- go test ./internal/beads -run TestCachingStoreCreateWithStorage -count=1
- go test ./cmd/gc -run TestBeadPolicyStore -count=1
- go test ./internal/orders -run TestCreateRun -count=1
- git diff --check

Scope is exactly three files. No binary install, service restart, or live fleet mutation.


tudorsaitoc:
Independent exact-head review: APPROVE `24e3b744e95a89ae764092aaf917ea59de995515`. All exact-head CI, CodeQL, vulnerability, Mac Regression, and formula checks are green. The fallback preserves StorageClass through plain CachingStore backing for both session and order-tracking, and the canary SQL is NULL-safe. `git diff --check` is clean. Tudor attempted a head-pinned squash merge, but GitHub denied `MergePullRequest` permission; an upstream maintainer must merge this exact head. Live rollout remains intentionally separate and unclaimed until merge.

tudorsaitoc:
Exact rebased head b6f345d121f936a71d9e1a8fc3ff500a659a0433 reverified after current-main rebase. All four changed file blobs are byte-identical to independently approved head 24e3b744e95a89ae764092aaf917ea59de995515. Aggregate scope is 4 files, +168/-1; diff-check clean. Exact-head CI, CodeQL, vulnerability, Mac regression, package, integration, and focused policy/cache/order tests are green. No binary install or live-row mutation has occurred.

quad341:
Thanks for this, and apologies for the slow turnaround — that's on us, not on the PR. Review below.

**Verdict: fix-merge** (ensemble: qwen `auto-merge`, codex `fix-merge`, medium confidence).

The two reviewers **agree completely on the code**. The split is entirely about one docs file, and codex has the better of it:

> `engdocs/no-history-cache-canary.md` pins candidate revision `b2763cc2e`, which is not the reviewed head (`b6f345d1`) and will not be the merge commit either. That makes precondition #1 of the runbook — *"the candidate binary's checksum and `gc version` identify `b2763cc2e`"* — unexecutable the moment this lands.

So the runbook would ship already-stale. Either drop the pinned revision in favour of something that stays true after merge, or reword the precondition so it doesn't depend on a specific pre-merge SHA.

**On the code itself**, the review flags this as a *fundamental* change to a critical path, and says so approvingly with one caveat worth stating plainly: `CachingStore.CreateWithStorage` previously discarded the policy-selected storage class on a non-`StorageCreateStore` backing — silently falling through to `c.Create(b)`. This PR computes and stamps `Ephemeral`/`NoHistory` and adds a new error return for an unknown class. That's control-flow plus a **new failure mode on a core store path**, and the PR does add coverage to exactly that surface, which is the right instinct.

Because it touches a critical path fundamentally, it needs a human sign-off before merge in addition to the review above — that's a standing rule here, not a comment on this change.

**Next:** the docs pin is the only blocking item. Fix that and this is ready for the human gate.

quad341 (approved):
@tudorsaitoc — this is a genuinely good catch. The old fallback quietly threw the policy-selected class away whenever the backing didn't implement `StorageCreateStore`, and I confirmed that's the live production shape: `NativeDoltStore` has `Create` and `ApplyGraphPlanWithStorage` but no `CreateWithStorage`, so every cached native-Dolt create was landing in that branch. Reusing `effectiveStorageFlags` is exactly right — `StorageDefault` still returns the bead's own flags, so nothing gets clobbered, and the semantics stay identical to `BdStore`. The two-level test coverage (adapter plus the real `wrapStoreWithBeadPolicies` stack, on both session and order-tracking beads) is what makes this easy to trust.

I'm merging with one small edit on my side: the canary doc pins `b2763cc2e`, which isn't the head here and won't be the merge commit — so precondition #1 would be unexecutable the moment it lands. I'll swap that for a recorded `CANDIDATE_REV` and re-run the focused tests.

Your "Evidence gap" note is the right instinct — the SQL-server readback still needs to happen at rollout. Thanks for writing that down instead of over-claiming.


---


Review coverage:
- completed: `qwen codex`
- unavailable: `claude`

Issue #3566, opened 2026-06-17
supervisor: subsume the pool-release liveness-predicate bundle under a holder-renewed work-claim lease (acute race mitigated by #3621)


iwata-1116:
> **Status / rescope (2026-06-22).** The acute symptom is mitigated; this issue now tracks the structural end-state — the work-claim lease.

## Update — acute duplicate-spawn race mitigated by #3621

The acute symptom this issue surfaced — a still-live holder's pool claim being stripped (idle-sleep / pending-interactive / completed-and-draining), minting a backup worker that redoes the work — is now **mitigated** by #3621 (`fix(supervisor): orphan-release strands live city-scoped pool holders`, merged). #3621 fixes the `openSessionOwnsWork` cross-store ownership predicate (gate-1): a live holder leaves an open session bead with the matching assignee, so once ownership is recognized at the source the bead is never released and no backup is minted — closing the per-trigger symptoms without per-trigger guards. Thanks @julianknutsen.

This issue stays open for the **structural end-state** #3621 intentionally did not bundle:

**1. Work-claim lease — the architecture.** The release path is still a *bundle* of liveness predicates (`openSessionOwnsWork`, the detached probe, `pendingInteractionKeepsAwake`); each new trigger adds another predicate. A holder-renewed **work-claim lease** — renewed across every live state (working / pending-interactive / draining) — collapses the bundle so release fires on exactly one condition, lease expiry. The #3621 author endorsed this directly:

> "The structural end-state remains the work-claim lease (#3566): store-ref is a per-tick inference, and a holder-renewed lease would collapse the whole predicate bundle. This PR is the correct next commit; the lease is the architecture."

**2. (Optional) gate-2 symmetric follow-up.** `assigneePreservesNamedSessionRoute` carries the same store-ref asymmetry for **named** cross-store sessions that #3621 fixed for the city-scoped path — worth a symmetric fix with its own test (noted in the #3621 follow-ups).

---

<details><summary>Original report (predicate-bundle analysis) — retained for history</summary>

## Summary

`releaseOrphanedPoolAssignments` strips a still-live holder's pool claim under several distinct liveness conditions. Each fix to date has been scoped to one trigger, so the next trigger re-opens the same "release strips a live holder → a backup polecat re-claims and redoes the work" race. The triggers are a *predicate bundle*, and neither the acquisition CAS (beads#3575) nor the deferred assignee-stomp CAS (#3458, "fix #4") subsumes them. A single **work-claim lease** does.

[...]

PR #5242, opened 2026-08-13
fix(cmd/gc): probe the work bead's owner store for orphan-release liveness

julianknutsen:
`releaseOrphanedPoolAssignments` validates an assignee's liveness with `liveOpenSessionAssignmentExists(store, …)` against the primary store only. On a city whose `[storage.classes]` relocates a bead class, the session bead that proves the assignee alive can live in the *work bead's own owner store* instead — the primary-store query then empty-succeeds, empty-success reads as "assignee dead", and live in-progress work is released out from under its worker every reconcile pass.

The fix adds a second probe against the work bead's owner store (`assignedWorkStores[i]` when store-aware, else the `storeForPoolAssignment` fallback) before releasing. Strictly conservative: any hit skips the release, so the only behavior change is *fewer* spurious releases.

`TestReleaseOrphanedPoolAssignmentsReadsLivenessFromWorkOwnerStore` fails on the pre-fix code (`released [{gc-2 0}], want none`) and pins the new probe; the control (no session bead in any store) still releases, so the probe is not a blanket suppression. The pinned `TestReleaseOrphanedPoolAssignments_*` suite (including both rig-store cases) stays green.

Found while diagnosing the maintainer-city adopt-pr stall (shipped there as part of `mc-aliasguard-e0e598c480`); it is independent of the pool-identity fix in #5241 and reviewable on its own.

Gates: `go vet ./cmd/gc ./internal/session` clean; targeted + pinned suites green; full `env -u OPENAI_API_KEY go test ./cmd/gc -count=1 -timeout 1800s` green on the combined stack.

---
- Related to #3566 — adds a second, cross-store liveness probe to the orphan-release path so a live holder whose session bead sits in the work bead's own owner store is no longer stripped — one more predicate in the bundle that the filed issue proposes collapsing under a holder-renewed work-claim lease; it does not implement that lease, and the named-session route asymmetry noted there is untouched

### DIFF
--- a/cmd/gc/pool_session_name.go
+++ b/cmd/gc/pool_session_name.go
@@ -165,6 +165,10 @@ func releaseOrphanedPoolAssignments(
 		if agentCfg == nil || !agentCfg.SupportsGenericEphemeralSessions() {
 			continue
 		}
+		// Resolved before the liveness gates because the graph-resident-session
+		// probe below reads the same store; the missing-store report stays where
+		// it was, so a bead skipped by a liveness gate never reaches it.
+		ownerStore := assignedWorkOwnerStore(cfg, store, rigStores, assignedWorkStores, i, wb)
 		if assignee == "" {
 			if wb.Status != "in_progress" {
 				continue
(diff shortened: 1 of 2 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-284771409693271

PR #5243, opened 2026-08-13
fix(gc): reject unregistered rig metadata owners

tudorsaitoc:
## Summary

- reject `gc.lease_owner` and `gc.routed_to` writes whose rig is absent from the loaded city configuration
- cover both `--set-metadata` forms and inline JSON `--metadata` before forwarding to `bd`
- fail closed for metadata inputs this boundary cannot interpret exactly, including `@file` and malformed guarded values
- retain registered rig, bare actor, dotted historical actor, unrelated metadata, flag-value, and `--` compatibility

## Why

A retired, unregistered rig continued renewing canonical leases.  Saitoc caller guards detect the symptom, but stale or external clients can still write through `gc bd`.  This patch enforces the narrow invariant at that shared CLI mutation gateway without changing read behavior or historical provenance.

Tracks canonical Saitoc Bead `saitoc-yrgraq.1`.

## Verification

- `go test ./cmd/gc -run "^(TestGcBdRejectsUnregisteredRigQualifiedMetadataWrites|TestGcBdRejectsMetadataItCannotValidateBeforeWrite|TestGcBdAllowsRegisteredAndLegacyMetadataActors)$" -count=1`
- `go test ./cmd/gc -run "^(TestGcBd|TestResolveBdScopeTarget)" -count=1`
- `go test ./internal/bdflags -count=1`
- `LINT_CHANGED_REF=upstream/main LINT_CHANGED_SCOPE=tracked make lint-changed`
- `git diff --check`

The full fast `cmd/gc` package sweep was stopped after exceeding 14 minutes locally with no output; the focused affected ownership and lint checks above passed.  Live completion still requires merge, pinned binary install, and scratch negative/positive probes.


quad341 (approved):
@tudorsaitoc — `TestGcBdAllowsRegisteredAndLegacyMetadataActors` is what makes this mergeable. Bare actors, dotted historical actors, a metadata-looking value sitting behind `--notes`, a `--metadata=` after the `--` terminator: those are exactly the false positives a guard like this dies on, and you pinned every one before anyone had to find them in production. Routing the scan through `bdflags.SplitGlobalFlags`/`ValueFlags` rather than a fresh parser was the right call too — the guard can't drift from what `bd` actually does. And putting the refusal after config load but before scope resolution means a rejected write genuinely cannot touch a ledger.

Two things on my side, neither yours. It's conflicting on `internal/beadmeta/keys.go` — that sorted const block plus its twin `KnownMetadataKeys` slice conflicts if you look at it sideways — so I'll rebase and resolve. I'm also adding one test case pinning a route like `saitoc/sub/polecat-01`, just so the first-slash parse is asserted rather than assumed. Then I'll re-run CI; the current red is a cancelled tmux shard, nothing you did.

I'm filing a follow-up on `gc.execution_routed_to` and the deferred route keys, which reach the same destination without passing this gate. Yours if you want it.


---

PR #5245, opened 2026-08-13
fix: write-orphaned class residents are reachable by id again

julianknutsen:
793 open beads on the measured city carry a rig/city prefix but **reside in a coordination-class store**, not in the store their prefix routes writes to. They are graph-native synthetics — input convoys, patrol roots, wisps — created through the class door with no id, so the binding workspace minted them from *its* prefix, which on a converged city is a work prefix. Reads reach them; writes route by prefix and hard-fail. Each one is a permanent `HasOpenWork` / ready-frontier polluter with no supported drain path (live example: a patrol root wedged open since June suppressed its order the entire time).

### Three by-id lanes; two lacked a residence probe

| Lane | Before |
|---|---|
| One-shot call sites (`classRoutedStoreForID`) | **Already had it** — `[class, work]` when the namespace gate cannot decide |
| CLI `gc bd` door | Probes the class leg for every id — but only for the verbs it serves, and **`close`/`reopen` were not in the verb set at all** |
| City HTTP API (`beadStoresForID`) | **No probe.** Documented as open work in two header essays, in nearly identical words |

So this completes two lanes to match a shipped pattern; it invents nothing.

**Slice 1 — CLI (`7aa1dbc5c7`).** `close` and `reopen` become served by-ID verbs, so the door's existing residence probe routes them exactly as it already routes `update`/`--claim`/`release-if-current`/`dep list`. Only the bare form: `GraphStore.Close(string) error` carries no reason and no workflow coupling, so bd's `-r/--reason/--reason-file/--session` and `--claim-next/--continue/--force/--no-auto/--suggest-next` make the invocation unrecognized rather than half-executed. `delete` and batch close stay unserved.

**Slice 2 — CLI (`5290e09149`), independently reviewable and revertible.** Ownership was provable only from a reserved prefix in the argv. Residency proves it too: an unserved mutation with unambiguous positional ids is probed against the class binding and refused with the offending flag named, instead of falling through to bd's misleading not-found. Widens the cost gate from *verb* to *subject* — reads, selectors and ambiguous scans still pay nothing.

**Slice 3 — API (`f76e3f47c7`).** `appendClassResidencyFallback` appends the class store **behind** the candidates each prefix arm already produced. Zero handler changes, zero wire changes, `openapi.json` untouched. Each handler already binds its write store to the store whose `Get` answered, so read/write coherence is structural.

**Multi-residency rule:** a write lands in the first store of the surface's own probe order whose `Get` answers — the same order that surface's reads use. CLI: class-first. API: prefix-first. Each surface is internally coherent; the cross-surface divergence for dual-resident ids is acknowledged, unchanged from today's read side, and retired by the sweep, not by routing. Pinned by `TestBdCloseDualResidentWritesServingCopy` and `TestBeadWriteDualResidentPrefersPrefixStore`.

**ADR-0009 note (`52151f8e52`, doc-only).** A door-routed close does not reach the work-record close gate. The bypass is recorded next to the gate's own coverage claim, along with why it is sound and which population it actually affects (dual-resident migrated work steps).

### Reviewer's attention
- **Mixed-batch atomicity.** `gc bd close gc-work1 gc-relic1` used to close `gc-work1` and error on the resident; it now refuses both. The one input that is "partial success → atomic refusal" rather than "error → correct".
- **404 → 500 on the API.** A prefix miss followed by an *unreachable* class store is now 500. That is the resolver's own fail-fast doctrine ("an unreachable store must not be reported as a missing bead"), with a 404 control asserted alongside.
- **Flag-order asymmetry (follow-up, not fixed here).** `gc bd close gcg-1` is served; `gc bd --json close gcg-1` is not — the served parsers key on `argv[0]`. Pre-existing for `show`; sharper now that close has both a served and an unserved spelling.
- **`delete` on a resident is a refusal with no `gc` alternative** — `close` covers the drain, but the message does not say so.
- **T14 is a pin, not a cure.** If the deploy branch has the read-leg/write-leg tier skew, the door now surfaces it loudly with bd's own not-found wording — which will *look* like this bug. The design's binding-leg verification is not optional before porting.

**Gates:** `go vet ./...` clean · `go test ./cmd/gc -count=1 -timeout 1800s` green (769s) · `go test ./internal/api/... ./internal/storeref/...` green incl. `TestOpenAPISpecInSync` · `make test` green · 19 designed tests + 2 review-gap tests, each core one demonstrated RED first, each control's teeth proven by mutation. Three-lens adversarial review completed pre-PR (design conformance / invariants / test adequacy); findings folded in or filed.

### DIFF
--- a/cmd/gc/cmd_bd_by_id.go
+++ b/cmd/gc/cmd_bd_by_id.go
@@ -74,10 +74,21 @@ package main
 //     storage and says nothing about a bead the work ledger still serves. See
 //     resolve.
 //
-// Served here: show, update (fields and metadata, including --claim),
-// release-if-current, and dep list. `gc bd heartbeat` is not — it is rewritten
-// to a metadata update before this hook runs — and neither is the general query
-// surface, which bd_relocated_classes.go refuses on its own terms.
+// Served here: show, update (fields and metadata, including --claim), close,
+// reopen, release-if-current, and dep list. `gc bd heartbeat` is not — it is
+// rewritten to a metadata update before this hook runs — and neither is the
+// general query surface, which bd_relocated_classes.go refuses on its own
+// terms.
+//
+// close and reopen are here for the same reason update is, one lifecycle step
+// further on. A class store MINTS ids from its binding workspace's own prefix,
+// so a synthetic it creates with no id — an input convoy, a patrol root, a
+// wisp — carries a WORK prefix while residing only in the binding. Reads found
+// those (the class leg is probed for every id); the close that would retire one
+// was not a by-ID verb at all, so it never opened this door, fell through to a
+// subprocess pointed at the prefix store, and died there with bd's not-found.
+// Every such open bead was a permanent ready-frontier polluter with no
+// supported drain path.
 //
 // # Ownership is decided before servability
 //
@@ -163,6 +174,14 @@ const (
 	// work ledger and left the step open in the binding forever: a molecule
 	// that stalls with no error anywhere.
 	bdByIDUpdate bdByIDVerb = "update"
+	// bdByIDClose is `gc bd close <id> [--json]`, and bdByIDReopen its undo.
+	//
+	// They are the drain verbs for a class resident: the row a class store
+	// minted under a work-shaped id has no other way to be retired, because
+	// every prefix-routed lane resolves it against a ledger that never held
+	// it. Only the bare form is served — see parseBdByIDCloseArgs.
+	bdByIDClose  bdByIDVerb = "close"
+	bdByIDReopen bdByIDVerb = "reopen"
 )
 
 // bdByIDDepDirectionUp asks for the beads that depend ON the subject; the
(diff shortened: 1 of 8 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-284771409693635

quad341 (dismissed):
@julianknutsen — the diagnosis here is the good part. "A class store mints from its binding workspace's own prefix, so a synthetic created with no id is born work-shaped and class-resident" explains 793 stuck beads, a patrol root wedged open since June, and *why* reads found them while writes hard-failed. That only comes from actually chasing the population.

I checked the load-bearing claims instead of taking them on trust. Every by-id API handler really does bind its write to the store whose `Get` answered, so the residency fallback picks up the write leg with zero handler changes. `bdMutationWriteIDs` really does fail closed on an ambiguous scan, so the widened cost gate can't probe a value instead of a subject.

The controls are what make a diff this size reviewable — `TestBdClosePrefixStoreBeadKeepsPassthrough` failing *differently* from T1, and the registry-constructor observer rather than `handled=false`. Thank you for flagging the mixed-batch atomicity change and the 404→500 yourself.

Merging as soon as the queued CI shards land green. One follow-up worth a bead: the ADR-0009 bypass is documented but not pinned by an assertion, so nothing trips if it silently changes.


---

julianknutsen:
Converting to draft, superseded-in-flight by the systemic residency fix: the census behind it found ~90 hand-rolled store-selection sites (this PR fixes 3 of them) built on binding-blind base enumerators, and the approved design (residency resolver in internal/storeref: one lookup contract, CI boundary, conformance corpus — epic ga-qdt5y) delivers this PR's exact behavior as the ById intent's ResidenceProbe leg in its S1 slice, with this branch's 21 tests ported as S1's regression pins. Landing this bespoke version first would add another enumeration site for S1 to unwind. The branch stays; if S0/S1 slip past a week, this reopens and merges as the interim fix.

julianknutsen:
Superseded and landed: S1 (#5267, merged) delivers this PR's three behaviors through the residency resolver (internal/storeref Plan(ByID)) with this branch's 21 tests ported as regression pins — see the S1 PR body for the one deliberately flipped assertion (dual residents now answer from the binding on BOTH planes, ending the CLI/API split-brain this PR's append-behind order would have preserved). The branch stays as the pin-provenance archive.

PR #4682, opened 2026-07-26
perf(gc): stop the order dispatcher reparsing city+pack TOML every tick

quad341:
## Problem

`CityRuntime.dispatchOrders` calls `dispatch()` on every reconcile tick. The
`stores` dedup map inside `dispatch` is a per-call local, so each scope
target's store-open re-ran the full `loadCityConfig` chain — decoding
`city.toml` plus every pack include — and did so **twice** per open (once in
`openStoreResultAtForCityWithAuthority`, again inside the
`nativeDoltOpenEnvForScope` call site).

A live perf profile (15,838 samples, 0 lost) attributed ~6.35 cores of
userspace cost to this path alone, via:

```
toml.Decode <- config.resolvedPackNames <- resolvedConfigPackNames
  <- LoadWithIncludesOptions <- loadCityConfigFS <- loadCityConfig
  <- openStoreResultAtForCityWithAuthority
```

## Fix

Add `openStoreResultAtForCityWithConfig` / `openStoreAtForCityWithConfig`,
which accept an already-resolved `*config.City` and only load when it is
nil. Thread the dispatcher's cached config through `order_dispatch.go`'s
`storeFn` and through both the exec and native store-open paths, removing
the duplicate parse.

The legacy nil-config entry points are preserved unchanged, so every other
caller keeps its current behavior.

## Benchmark

`order_dispatch_bench_test.go` builds a 15-pack / 120-agent city fixture so
the parse cost the fix removes is visible above per-op noise.

| metric | before | after | change |
|---|---|---|---|
| latency | ~16.66 ms/op | ~0.163 ms/op | ~102x faster |
| allocation | ~11.9 MB/op | ~32 KB/op | ~368x less |
| config reparses | 1/op | 0 | eliminated |

## Tests

- `order_dispatch_bench_test.go` — new benchmark with the multi-pack fixture
- `order_dispatch_test.go` — asserts the dispatcher reuses the cached config
- `store_rollout_test.go` — covers the preserved nil-config entry points

---
🔗 **Maintainer cross-reference** — added by the gascity maintainers, no action needed from you:
- Related to #2604 — removes the redundant city.toml + pack-include reparse that the order dispatcher was paying on every tick for every scope store open, cutting one measured contributor to the dispatch phase that dominates each controller tick; it does not move dispatch off the synchronous tick path, replace the fork-per-operation bd CLI store, reduce the per-rig order fan-out, or address the cold-start thundering herd, so the starvation described in that issue remains open

[...]

PR #5247, opened 2026-08-13
fix(cmd/gc): unbreak the cmd/gc test build after the order-dispatcher routes change

julianknutsen:
`origin/main` does not compile its own `cmd/gc` test package: #4682 gave `newMemoryOrderDispatcher` a leading `*storageRoutes` parameter and updated its two production call sites but not the three test call sites, so `go vet ./cmd/gc` (and any `go test ./cmd/gc`) fails at head:

```
vet: cmd/gc/order_dispatch_bench_test.go:93:78: not enough arguments in call to newMemoryOrderDispatcher
	have ([]orders.Order, string, *config.City, events.Recorder, io.Writer)
```

Three-line fix passing the routes the two production call sites already pass. Verified red on `ed267ad66b` (main) and green with the fix. Found while branching unrelated work off main; extracted standalone since it blocks anyone running the suite at head.

### DIFF
--- a/cmd/gc/order_dispatch_bench_test.go
+++ b/cmd/gc/order_dispatch_bench_test.go
@@ -90,7 +90,7 @@ func BenchmarkOrderDispatchTick(b *testing.B) {
 
 	b.Run("PreFix_ReparsePerTick", func(b *testing.B) {
 		before := loadCityConfigCalls.Load()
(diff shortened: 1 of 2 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-284771409693863

PR #4596, opened 2026-07-24
Normalize routed work instance names in demand matching

quad341:
## What this changes

The controller now normalizes instance-suffixed `gc.routed_to` values on the read side before matching routed work to a pool template. A work item routed by a direct metadata write to an instance name such as `gascity/builder-1` is counted as demand for the `gascity/builder` template instead of being silently ignored when the pool is scaled to zero.

This makes demand computation robust regardless of which writer produced the route metadata. The existing `gc sling` write-side normalization remains in place; this change makes the reader enforce the same invariant for direct `bd update` and graph/formula paths.

## Review notes

- Main logic is in `cmd/gc/build_desired_state.go`; coverage is in `cmd/gc/build_desired_state_test.go`.
- `readyExcludeTypes`, blocking-dependency checks, and control-dispatcher alias handling are intentionally unchanged.
- The graph.v2 drain-unit incident is not claimed fixed by this patch; notes identify it as a separate root-cause track.

## Test plan

- [x] `go vet ./...`
- [x] `go build ./cmd/gc/`
- [x] `go test ./cmd/gc/... -run 'TestDefaultScaleCheckCounts|TestDefaultScaleCheckDemand|TestControllerDemandRouteTarget|TestDefaultScaleCheckCountsAndDemand' -count=1 -v`
- [x] `make test-fast-parallel`
- [x] Release gate: [`release-gates/ga-803rif-gate.md`](release-gates/ga-803rif-gate.md)

PR #1600, opened 2026-05-02
perf(controller): use cached demand read model

julianknutsen:
## Summary
- use the active cache read model for controller in-progress and ready demand queries
- count default scale demand from cached routed ready work for store-backed reconciliation
- fall back to live ready checks when cached dependency coverage is incomplete

## Tests
- go test ./internal/beads ./cmd/gc
- pre-commit full fast gate

---
<a href="https://app.blacksmith.sh/gastownhall/codesmith/gascity/pr/1600"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need.</sup>

- [ ] Let Codesmith autofix CI failures and bot reviews

PR #5250, opened 2026-08-13
fix(controller,hook): restore demand/claim reader agreement, and make a stalled claim converge

julianknutsen:
# fix(controller,hook): restore demand/claim reader agreement, and make a stalled claim converge

Closes the run-operator claim-without-execution stall (ga-umlt1) and the
no-wake-reason drain lane (ga-whzrt).

## The design frame: three operator rulings

Everything here is downstream of three rulings, quoted verbatim because they are
the spec:

> "The controller should never be assuming which bead is picked up. It only
> starts an ephemeral session if there is demand."

> "If there is demand for work b/c bd ready saw it in order to start a session
> then the gc hook --claim in the session should see it."

> "Demand and hook were always meant to be using the same semantics."

The pool is **pull**. The controller scales capacity on evidence that work
exists; the worker discovers and claims for itself. The trigger bead id is
demand bookkeeping, never an assignment. An earlier revision of this work added
a by-id claim tier — the controller telling the seat what to take — and it was
withdrawn for violating the first ruling. What ships instead fixes the reason
the two readers disagreed.

## Archaeology: this was not always broken

Parity between the controller's demand read and the worker's claim read was
**real** until #1600 (2026-05-03) copied the eligibility predicate rather than
sharing it. Every divergence after that landed in one copy and not the other:
#1942, #4596/#4597, the hold-label series, and #5071. None was wrong on its own;
each was a filter added to the query that the demand loop never learned about.
The full table is in ga-umlt1.

That history is why the fix is not "add the missing filters" — that would be a
fourth copy. It is "make the demand loop a consumer of the one definition
again," with the drift made structurally impossible.

## The four holes

**H1** — the spawn→claim handoff was discovery-based and the two discoveries had
diverged: the controller counted rows (routed epics, dispatch-held beads,
slot-suffixed `-N` routes) that the worker's own Tier-3 query is forbidden to
serve, so a seat spawned, read empty, drained, and the row was counted again on
the next tick, forever. **H2** — a claimed-but-unexecuted bead had no recovery
lane: every backstop clears the instant the bead flips `in_progress`, and
wake-on-assigned-work dropped graph-resident claims at the store-ref
reachability filter, which is also ga-whzrt's root. **H3** — nudge delivery to
codex pool seats was dead end to end: the submit key sequence was missing, an
unverified paste was recorded as delivered, and the CLI printed success
regardless. **H4** — a failed claim read was indistinguishable from an idle
store: `gc ready` aborts the whole federated read when any leg errors, the hook
exited 1, and nothing reached the bus.

## What ships

Ten commits, severable, in dependency order.

- **F3 — assigned-work reachability** (`f0e89d6`). A claim on the leading
  class-binding arm is kept when the assignee matches one of the session's own
  exact identities, and the live re-read scans the relocated binding. Without
  it the decision is literally `AwakeDecision{ShouldWake:false Reason:""}` — the
  no-wake-reason shape that recycled live workers mid-step.
- **F2 — execution backstop** (`67815659`, teeth in `e48ca675`). A pool seat
  holding an in-progress claim it never started gets the agent's own claim nudge
  re-delivered on the proven pacing, guarded by provider idleness held in the
  predicate (unknown activity holds — never nudge on "we cannot tell").
- **F4 — nudge honesty + codex submit** (`2a6b1f5f`). Escape→Enter for codex,
  exactly once; codex becomes submit-verify eligible so an unconfirmed paste
  requeues under the existing cap instead of being acked and deleted; store-open
  failures name their cause; the queued line says where the item went and why
  the live leg was skipped.
- **F1′b — failure honesty** (`ed19afac`). Every failed read is a lifecycle
  event; a read *error* retries 3×5s (emptiness never does — an empty read is an
  answer); a failed federated primary is terminal rather than letting a
  federation-blind leg mint a false `no_work` drain-ack.
- **F1′a — one predicate, shared** (`aa5688bb`, aligned in `0bd41385`).
  `config.PoolDemandServeRules` names the Tier-3 serving rules **and the query's
  flags are rendered from it**, so a flag cannot exist without a field. The
  demand loop consumes that same value. The `-N` route collapse is persisted by
  the pass that already owns route canonicalization.
- **F1′c — staleness** (`488bb32f`). The demand snapshot's fingerprint now
  hashes the store the probe actually reads (the binding), killing the
  phantom-demand window.
- **Observability** (`f8c09482`). `GC_SPAWN_ORIGIN=demand` (presence only) and
  `session.demand_claim_divergence`, emitted strictly *after* the drain result,
  reading the trigger bead solely to classify.

### Drift-proofing, which is the real deliverable

Three layers, because a shared predicate that is shared only by convention
drifts back apart — which is exactly what #1600 started:

1. The query's flags are **rendered from** the descriptor.
2. A bidirectional pin: no rule the query lacks, no query flag the rules lack.
3. A **row-corpus conformance** that takes the routed and legacy pool-demand
   tiers straight out of the generated query text, parses them through the real
   `gc ready` flag surface, and runs the real serving filter over the corpus,
   requiring the same verdict row by row. Editing only the Go predicate fails
   with `AGREEMENT VIOLATED`; editing only the rule source fails on the corpus.

That conformance immediately found two things review had not: a `run_target`-only
workflow root is served by the legacy compat tier (so the serving surface is both
tiers), and a **real semantic bug** — the predicate case-folded types while the
serving filter does not.

**The last-word rule** resolves that class generally: each dimension matches
whichever filter has the last word on what the worker sees. Types compare
exactly (the reader is last; bd's `--exclude-type` is alias-expanded but not
case-folded, landing in SQL as `issue_type NOT IN (?)`). Labels compare
case-insensitively (the reader is *not* last — the hook re-applies the hold
filter in Go with `EqualFold` after the query returns). Parity is with the
**effective serving surface**: reader composed with hook post-filter.

### F2's convergence teeth

F3 keeps claim-holders awake by design — correct, but the no-wake-reason drain
it removes was, accidentally, the only thing recycling wedged seats. Exhaustion
therefore routes through the **tracked** drain (`beginSessionDrainInfo`) with a
new reason, `execution-stalled`, excluded from all three cancel lenses. That
exclusion is the point: the session is awake, running, and holding an
in_progress claim — the shape every keep-alive guard protects — so a cancelable
reason would be canceled by the very claim that justified the drain. An
accidental recycle becomes deliberate, bounded convergence, proven end to end:
exhaustion → tracked drain → stop → close → dead-assignee reopen → the row is
demand again and matches a worker's claim route.

## Rollout metrics

- `idle-claim-nudge` deliveries → **~0**. Each one was the handoff failing.
- `session.work_query_failed` with the ordinary-failure reason: briefly nonzero
  (making the invisible visible), then declining.
- `session.demand_claim_divergence`: **benign only**. Benign is a sibling seat
  winning the race — correct pull, expected forever. Divergence-classified is the
  invariant breaking and targets zero.
- `execution.step_stalled`: bounded, each one a claim-without-execution caught in
  ≤11 minutes instead of recycle roulette.

## Reviewer attention

- **The C7 carve-out** (`ed19afac`). A failed federated primary is terminal —
  *except* an extra leg's own in-progress row, which is still accepted. Every
  leg's query matches this agent's identity, so an in_progress row is one this
  session already owns and no wider read could overturn it; without the carve-out
  crash recovery dies whenever the primary is down. Two lenses accepted this
  independently; it is still the one place a design default was narrowed.
- **Honest red-first record.** Two of F1′b's rows (federated-primary terminal,
  in-progress resume) also pass against pre-fix code — the old
  `bestStoreWithWork` already surfaced the primary error when no leg hit. They
  are regression pins on a previously incidental guarantee, not corrections. The
  commit message says so.
- **Label-vs-type matching** (above). Two different comparisons on purpose. If
  the last-word rule is wrong, both lines move.
- **Dimension independence of the collapse.** The route-form pass rewrites a
  held or epic row's `-N` route too. Being ineligible for demand is the
  predicate's verdict, not a reason to leave behind a route form nothing can
  match once the hold lifts.
- **Flake disclosure.** One earlier full run, deliberately overlapped with the
  `./internal/...` sweep, failed
  `TestCityRuntimeForceShutdownTearsDownAfterLateAsyncSweep` plus
  `internal/doctor` (TempDir cleanup race) and `internal/pidutil` (`ps` killed).
  All three are load-shaped, all pass on serial re-run, and none touch the diff.
  The clean uncontended run is the gate of record.

## Gates

`go test ./cmd/gc -count=1 -timeout 1800s` → `ok 866.830s` (uncontended, final
tree) · `go vet ./...` clean · `./internal/... ./pkg/...` 145 ok · §4 protected
suites, split-topology conformance and dispatch/convoy goldens green ·
`check-core-boundary`, `-eventexport-isolation`, `-routed-test-rows`,
`-split-topology-rows`, `-generated-docs-drift` all OK · OpenAPI, generated Go
client and dashboard TS regenerated in lock-step for both new event types;
dashboard typechecks pass.

`execution.step_stalled` and `session.demand_claim_divergence` are both
deliberately **off** the `pkg/eventexport` allowlist, pinned by test: internal
diagnostics carrying bead refs, and widening a redaction boundary is its own
reviewable change.

## Out of scope, filed separately

F1′d (HQ work store as a demand leg — D6, the last leg-set divergence, whose
conformance row is already written and pinned), F1′e (origin-gate and
custom-`work_query` truthfulness), F5 (session-verb residency), the
council-gated success-but-empty retry (C2, default off), and the
fourth-demand-counter follow-up on the control-dispatcher lane.


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-284771409694108

Issue #3946, opened 2026-07-05
Homebrew gastownhall/gascity installs beads 1.1.0, but gc 1.3.x requires bd v1.0.4 → native store silently disabled (slow CLI-subprocess fallback)


vijiboy:
**Related:** #3878 (confirms v1.0.4 is the intended `BD_VERSION` for 1.3.x; v1.0.5 withdrawn). This issue is about a *different* delivery path — the Homebrew formula — pulling the wrong bd anyway.

## Summary
Installing Gas City via the Homebrew tap pulls `beads` **1.1.0** as a dependency, but gc **1.3.3** vendors `github.com/steveyegge/beads v1.0.4` (per its `go.mod`, and consistent with #3878 / CI's `BD_VERSION=v1.0.4`). Because the `bd` CLI version ≠ gc's linked beads-library version, the `version_compat` preflight gate fails, the in-process **NativeDoltStore is disabled**, and every store operation falls back to shelling out to the `bd` CLI. The result is a **silently slow** install for every Homebrew user — `gc status`/`gc doctor` take many seconds to minutes.

## Environment
- gc: `1.3.3` (`55acb481…-dirty`, Homebrew tap `gastownhall/gascity`)
- bd/beads: `1.1.0` (Homebrew `beads` formula; homepage `github.com/gastownhall/beads`)
- dolt: `2.1.10`
- OS: Ubuntu on WSL2, Linuxbrew

## Root cause
- `go.mod` (v1.3.3): `github.com/steveyegge/beads v1.0.4`  ← the linked library
- Homebrew `beads` formula: latest stable = `1.1.0`
- `gastownhall/gascity` formula: `depends_on "beads"` (unversioned) → resolves to `1.1.0`

So the formula's dependency doesn't match the version gc is built against.

## Symptoms
```
$ gc doctor
… native_store_unavailable gate=version_compat
  reason="bd version differs from linked beads library version"
$ gc status --json | jq .beads
{ "beads_store": "BdStore", "native_store_eligible": false,
  "preflight_gate": "version_compat" }
```
`gc doctor` >120s (times out), `gc status` slow; store shows `BdStore` not `NativeDoltStore`.

## Workaround (what actually restores the native store)
Three things were required — worth documenting even after the formula is fixed:
1. **Match bd to the vendored lib.** Install bd **v1.0.4** and put it ahead of Homebrew's on PATH:
   ```
[...]

PR #5252, opened 2026-08-13
fix(beads): allow a semver-compatible newer bd release past version_compat (#5164)

### LINKED ISSUE #5164: feat: version_compat gate should allow semver-compatible newer bd CLI (not just exact match)
## Summary

The `version_compat` preflight gate (`PreflightChecker.checkVersionCompat`) requires an **exact string match** between the installed `bd` CLI version and the beads library version gc was compiled against. When the bd CLI is **newer** (but semver-compatible), the gate still fails, forcing a fallback to BdStore (fork-per-op) — which is silently slow and causes downstream `order-firing-current` and `fork-rate` doctor warnings.

## Problem

- **gc 1.4.0** (latest stable, Homebrew) — `go.mod` depends on `github.com/steveyegge/beads v1.1.0`
- **beads 1.1.2** (latest stable, Homebrew) — installed as gc's brew dependency
- The gate rejects bd v1.1.2 because it's not an exact match for v1.1.0

This is the same class of issue as #3946 and #3632, but from the other direction: those reported brew installing an **older** bd than gc expected; here brew installs a **newer** bd than gc's pinned `go.mod` version. Both stem from the same root cause: the Homebrew `gascity` formula has `depends_on "beads"` (unversioned), so it always pulls the latest beads release, which drifts from gc's compiled-in version.

## Impact

```
$ gc doctor
⚠ beads-store — beads store running on BdStore fallback (fork-per-op; gate=version_compat): bd version differs from linked beads library version
⚠ order-firing-current — scheduled orders are stale (30s-interval orders can't keep up with fork-per-op overhead)
⚠ fork-rate — high process fork rate: 170+ forks/s (warn >= 100/s)
```

Every Homebrew user hitting this combination gets a silently degraded experience: native store disabled, fork-per-op overhead, orders falling behind.

## Proposal

The `version_compat` gate should allow a **newer** bd CLI version to pass when it's semver-compatible with the compiled-in library version. Specifically:

1. **Semver-compatible newer versions pass**: if bd CLI is `>=` the linked library version within the same major.minor (or same major, per semver), the gate should pass (or at worst warn, not block).
2. **Exact match only for older versions**: if bd CLI is **older** than the linked library, the gate should still fail (older bd may lack APIs gc needs).
3. **Expose the expected version** (also requested in #3977): `gc doctor` and `gc --version` should print the linked beads library version so users can see what's expected without `go version -m`.

## Environment

- gc: `1.4.0` (Homebrew, Linuxbrew on WSL2 aarch64)
- bd/beads: `1.1.2` (Homebrew, latest stable)
- dolt: `2.2.3`
- OS: Ubuntu 24.04 on WSL2, Linuxbrew aarch64

## Related

- #3946 — Homebrew installs beads 1.1.0, but gc 1.3.x requires bd v1.0.4 (same pattern, older bd)
- #3632 — gascity 1.3.1 binary embeds older beads than Homebrew's unpinned beads dep (same pattern)
- #3977 — version_compat gate is unobservable — linked beads version exposed nowhere
- #3248 — Native store silently disabled: gc-installed bd hooks gate off gc's own in-process store

## DeepWiki confirmation

DeepWiki analysis of the gascity codebase confirms:
- `checkVersionCompat` does an exact string match, no semver logic
- `GC_BEADS_FORCE_FALLBACK` exists but does the **opposite** (forces fallback, not native)
- The only bypass is when the library version is `"(devel)"` (source build) — not useful for Homebrew users
- No env var or config option to bypass the gate for newer bd versions

rjgeng:
Fixes #5164.

## What's broken

The `version_compat` preflight gate required an exact string match between the installed `bd` CLI version and the beads library version `gc` was compiled against. When `bd` is newer but semver-compatible (same major version, not older), the gate still failed, forcing every op through the slow fork-per-op `BdStore` fallback instead of the native store — visible in `gc doctor` as `beads-store`/`order-firing-current`/`fork-rate` warnings.

This is the mirror image of the already-handled "bd is older" and "library version is unconfirmable" cases: Homebrew's `gascity` formula depends on `beads` unversioned, so it always installs the latest release, which drifts ahead of `gc`'s pinned `go.mod` version over time (same root cause as #3946/#3632, opposite direction).

## Fix

`checkVersionCompat` now falls through to a semver comparison before failing on a version-string mismatch:
- Same major version and `bd >= library` → pass (the new case).
- Different major, or `bd < library` → fail, exactly as before.
- Either version failing to parse as valid semver → falls back to the existing exact-match behavior, unchanged.

## Verification

- **TDD, RED confirmed against the pre-fix exact-match compare**: `TestCheckVersionCompatSemverCompatibleNewerBD` (newer patch/minor pass, differing major and older versions still fail, non-semver input falls back to exact-match failure) and `TestPreflightEligibleOnSemverCompatibleNewerBD` (the live #5164 shape end-to-end through `Check()`) both failed with the old "bd version differs..." FAIL before the fix, GREEN after.
- Full `go test ./internal/beads/...` (package + subpackages) green, no regressions to existing version-compat tests.
- `go vet -tags gms_pure_go ./internal/beads/contract/...` and `go build -tags gms_pure_go ./...` clean, gofmt clean.

Pre-existing, unrelated `cmd/gc/order_dispatch_*_test.go` compile errors on `main` (see #5251) were confirmed present on stock `upstream/main` before this change and are not touched here.

### DIFF
--- a/cmd/gc/order_dispatch_bench_test.go
+++ b/cmd/gc/order_dispatch_bench_test.go
@@ -90,7 +90,7 @@ func BenchmarkOrderDispatchTick(b *testing.B) {
 
 	b.Run("PreFix_ReparsePerTick", func(b *testing.B) {
 		before := loadCityConfigCalls.Load()
-		ad := newMemoryOrderDispatcher(aa, cityDir, cfg, events.Discard, io.Discard)
+		ad := newMemoryOrderDispatcher(nil, aa, cityDir, cfg, events.Discard, io.Discard)
 		// Reproduce the pre-fix storeFn verbatim (order_dispatch.go:431-433
 		// before the ga-237xpr fix): ignore the dispatcher's cached cfg and
 		// go through the nil-cfg path, which reloads city.toml + every pack
@@ -110,7 +110,7 @@ func BenchmarkOrderDispatchTick(b *testing.B) {
 
 	b.Run("PostFix_CachedConfig", func(b *testing.B) {
 		before := loadCityConfigCalls.Load()
-		ad := newMemoryOrderDispatcher(aa, cityDir, cfg, events.Discard, io.Discard)
+		ad := newMemoryOrderDispatcher(nil, aa, cityDir, cfg, events.Discard, io.Discard)
 		now := time.Now()
 		b.ResetTimer()
 		for i := 0; i < b.N; i++ {
--- a/cmd/gc/order_dispatch_test.go
+++ b/cmd/gc/order_dispatch_test.go
@@ -1915,7 +1915,7 @@ func TestOrderDispatchDoesNotReparseConfigPerTick(t *testing.T) {
 		Interval: "1h",
 		Exec:     "true",
 	}}
-	ad := newMemoryOrderDispatcher(aa, cityDir, cfg, events.Discard, io.Discard)
+	ad := newMemoryOrderDispatcher(nil, aa, cityDir, cfg, events.Discard, io.Discard)
 
 	before := loadCityConfigCalls.Load()
 	now := time.Now()
--- a/internal/beads/contract/preflight_checker.go
+++ b/internal/beads/contract/preflight_checker.go
@@ -273,9 +273,33 @@ func (c PreflightChecker) checkVersionCompat(ctx PreflightBDContext, err error) 
 		return NewPreflightCheckResult(PreflightCheckVersionCompat, PreflightCheckPass, "bd/beads schema compatible; linked library version unconfirmed ("+reason+")", details)
 	}
 	if strings.TrimPrefix(ctx.BDVersion, "v") != libraryVersion {
+		if newerSemverCompatibleBD(ctx.BDVersion, libraryVersion) {
+			return NewPreflightCheckResult(PreflightCheckVersionCompat, PreflightCheckPass, "bd version is a newer semver-compatible release of the linked beads library version", details)
+		}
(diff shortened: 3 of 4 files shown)


quad341 (approved):
@rjgeng — thank you for this one, and specifically for how you framed it. Naming it as the mirror image of the already-handled "bd is older" and "library version unconfirmable" cases put the fix in exactly the right place in the precondition chain, and the fallback-to-exact-match-on-unparseable rule means the change can only ever widen what passes, never what fails. That property is what makes it safe to ship without a long argument.

The test table earns its keep too — differing-major, older-bd, and non-semver input alongside the passing cases is precisely the "widening *unknown* must not widen *compatible*" discipline the surrounding tests already established, and the end-to-end test asserting `Fallback == ""` targets the actual symptom rather than the internals. The bd contract suites (current and minimum-supported) all ran green on the real surface.

Merging as-is. Two notes, neither needing anything from you: your body says the `order_dispatch_*_test.go` compile errors aren't touched here, but the diff does fix them — correctly; `nil` is the sanctioned identity routing for that constructor. And the `Go module vulnerabilities` red is repo-wide on a PR that changes no dependencies; that's ours to handle separately.

One thing I couldn't confirm from here: `main`'s go.mod pins beads at a pseudo-version, which short-circuits to PASS before your branch is reached — so your path only fires on a release-tag build like the Homebrew bottle. If you happen to have `gc doctor --json` from the machine in #5164, the `beads_library_version` field would close that loop nicely. Not a condition of merging.


---

PR #5253, opened 2026-08-13
fix(dolt-cleanup): drop rss= from the reap-stage ps scan (#5201)

### LINKED ISSUE #5201: dolt-cleanup: reap stage hard-fails on macOS when ps denies rss= for other sessions' processes
## Summary

`gc dolt-cleanup`'s reap stage unconditionally fails (`errors_total:1`, `stage:"reap"`, `exit status 1`) on macOS hosts where `ps` requires an entitlement to report resource-usage fields (`%mem`/`vsz`/`rss`/`time`) for processes outside the caller's session. This happens even when there are zero real orphan databases/processes to reap — the scan/drop/lsof-based safety logic all succeed; only this one `ps` call fails.

## Root cause

`cmd/gc/dolt_cleanup_discovery.go:409`, in `psLStartCommandLines()`:

```go
cmd := exec.CommandContext(ctx, "ps", "-ax", "-o", "pid=,rss=,lstart=,command=")
out, err := cmd.Output()
if err != nil {
    return nil, err
}
```

Requesting the `rss=` field causes `ps` to exit non-zero on hosts where reading resource fields for other users'/sessions' processes requires an entitlement the caller doesn't have (observed message: `ps: %mem/vsz/rss/time: requires entitlement`). The function returns the error with no fallback, and it propagates all the way up to the reap stage's top-level error, even though:

- The actual reap-safety decision (which PIDs are safe to kill) is computed separately via `lsof`-based cwd/deleted-inode checks, not from this `ps` output.
- `RSSBytes` parsed from this call (`parseDoltPSLine`) is only used for the cosmetic `bytes_freed_rss` summary field.

So a purely informational metric turns a clean, zero-orphan scan into a reported error every run.

## Suggested fix

Drop `rss=` from the requested fields in `psLStartCommandLines()` (keep `pid=,lstart=,command=`, which is all the actual matching logic uses), or fall back to a `ps` call without `rss=` when the first attempt fails, and treat `bytes_freed_rss` as best-effort/zero when unavailable.

## Impact

On affected hosts, every `gc dolt-cleanup` run (including `--probe` dry-runs) reports a stage error regardless of actual orphan state, which:
- Makes it impossible to distinguish "real problem" from "cosmetic ps failure" from the JSON output alone.
- Causes formulas built on top of `dolt-cleanup` (e.g. a periodic stale-DB cleanup formula with fail-open error handling) to never close out cleanly and to keep re-escalating on their normal cadence indefinitely, since the tool never reports a clean success.

## Environment

- `gc` v1.4.0 (`github.com/gastownhall/gascity@v1.4.0`)
- macOS (Apple Silicon), host `ps` restricts resource fields for processes outside the caller's session; confirmed independent of any application-level sandboxing.

rjgeng:
Fixes #5201.

## What's broken

`gc dolt-cleanup`'s reap stage unconditionally failed (`errors_total:1`, `stage:"reap"`) on macOS hosts where `ps` requires an entitlement to report resource-usage fields (%mem/vsz/rss/time) for processes outside the caller's own session — even when there are zero real orphan databases/processes to reap. `psLStartCommandLines()` requested `rss=` as one of the `-o` fields; `ps` exits non-zero for the *entire* invocation when it can't satisfy that field for another session's process, and the function had no fallback, so the error propagated all the way up to a reported reap-stage failure regardless of actual orphan state.

`RSSBytes` parsed from this call is cosmetic-only downstream: `planOrphanReap` classifies reap-vs-protect purely on `ConfigPath`/`DataDir`/`CWDState` (confirmed by reading `cmd/gc/dolt_cleanup_reaper.go`), and the only other consumer is the summary's `bytes_freed_rss` field. Not worth trading away the whole scan for.

## Fix

Drop `rss=` from the requested `ps` fields entirely (now `"pid=,lstart=,command="`), and adjust the two line parsers (`parseDoltPSLine`, `argvFromPSLine`) from 7 leading fields to 6. `RSSBytes` on this path is now always its int64 zero value — the fallback `ps`-based discovery path (hosts without `/proc`) never reported it as anything but cosmetic anyway; the `/proc`-based path (`readProcRSSBytes`, Linux) is unaffected and still reports real RSS.

## Verification

- Updated the three existing `parseDoltPSLine` fixtures to the new 6-field `ps` line shape (no `rss=` token) and added `TestPSOutputFormatExcludesRSS`, a regression guard pinning that the requested field spec never re-adds `rss=`.
- Live-verified the exact new `ps -ax -o "pid=,lstart=,command="` invocation succeeds end-to-end on Darwin.
- Full `go test ./cmd/gc/...` green (including the broader dolt-cleanup/reap/wisp-gc suites, no regressions); `go vet ./...` and `go build ./...` clean repo-wide; gofmt clean.

Stacks on #5251 (pre-existing, unrelated `cmd/gc` compile blocker on `main`) purely to compile in isolation — this branch deliberately excludes that fix so it stays its own PR.

### DIFF
--- a/cmd/gc/dolt_cleanup_discovery.go
+++ b/cmd/gc/dolt_cleanup_discovery.go
@@ -403,10 +403,20 @@ func readDoltSQLServerArgv(pid int) ([]string, bool) {
 	return argv, true
 }
 
+// psOutputFormat is the -o field spec passed to `ps` for process discovery.
+// Deliberately excludes rss=: some macOS hosts require an entitlement to
+// report resource-usage fields (%mem/vsz/rss/time) for processes outside the
+// caller's own session, and ps exits non-zero for the *entire* invocation
+// when it can't — turning a clean, zero-orphan scan into a reported
+// dolt-cleanup reap-stage error (gastownhall/gascity#5201). RSSBytes is
+// cosmetic-only downstream (planOrphanReap classifies purely on
+// ConfigPath/DataDir/CWDState, never on RSS), so it is not worth requesting.
+const psOutputFormat = "pid=,lstart=,command="
+
 func psLStartCommandLines() ([]string, error) {
 	ctx, cancel := context.WithTimeout(context.Background(), psEnumerationTimeout)
 	defer cancel()
-	cmd := exec.CommandContext(ctx, "ps", "-ax", "-o", "pid=,rss=,lstart=,command=")
+	cmd := exec.CommandContext(ctx, "ps", "-ax", "-o", psOutputFormat)
 	out, err := cmd.Output()
 	if err != nil {
 		return nil, err
@@ -421,17 +431,13 @@ func psLStartCommandLines() ([]string, error) {
 }
 
 func parseDoltPSLine(line string, pidPorts map[int][]int) (DoltProcInfo, bool) {
-	fields, command := consumeLeadingFields(line, 7)
-	if len(fields) != 7 || command == "" {
+	fields, command := consumeLeadingFields(line, 6)
+	if len(fields) != 6 || command == "" {
 		return DoltProcInfo{}, false
 	}
 	pid, err := strconv.Atoi(fields[0])
 	if err != nil || pid <= 0 {
 		return DoltProcInfo{}, false
-	}
-	rssKB, err := strconv.ParseInt(fields[1], 10, 64)
-	if err != nil || rssKB < 0 {
-		rssKB = 0
 	}
 	argv := parseDoltPSCommandLine(command)
 	if !looksLikeDoltSQLServer(argv) {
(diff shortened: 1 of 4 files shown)


quad341 (approved):
@rjgeng — the diagnosis here is the valuable part. Tracing an unconditional `errors_total:1` on the reap stage back to `ps` failing the *entire* invocation because one requested field needs an entitlement is not an obvious catch, and "zero orphans still reports a failure" is exactly the kind of noise that quietly trains people to stop reading the output.

I especially appreciate that you checked whether `RSSBytes` was load-bearing before dropping it instead of assuming. I re-traced `planOrphanReap` and `sumReapTargetRSS` against current `main` and your reading holds: it never touches classification, only the summary field. And pulling the field spec into a named const so `TestPSOutputFormatExcludesRSS` can actually pin it is the right instinct — that's a guard, not a comment and a hope.

Merging as-is. Two notes for the record, neither needing anything from you: the `newMemoryOrderDispatcher(nil, …)` lines your branch carries are already on `main` identically, so they merge as a no-op; and our Mac jobs were skipped on this run, so CI green is Linux-only — your live Darwin verification is carrying the macOS half, and I'm taking it. Thanks.


---

PR #5254, opened 2026-08-13
test(pidutil): skip TestCmdline_FailsClosedWhenUnreadable on darwin too

rjgeng:
## What's broken

`TestCmdline_FailsClosedWhenUnreadable` simulates "argv unreadable" by poisoning `ps` on `PATH` with a script that always fails, then asserts `AliveWithCmdline(self, ...)` fails closed (returns `false`). That only actually tests something on the platform build where `platformCmdline` shells out to `ps` (`cmdline_other.go`, `//go:build !darwin`) — Linux already skips (it reads `/proc/<pid>/cmdline` directly, never touching `ps`), and Darwin's own `platformCmdline` (`cmdline_darwin.go`) reads argv via `unix.SysctlRaw("kern.procargs2", pid)`, a native syscall that also never shells out to `ps`. Confirmed by tracing `psCmdline`'s only production caller: `cmdline_other.go`'s `platformCmdline`, never reached on darwin.

So on Darwin the `ps` stub has no effect: `Cmdline(os.Getpid())` correctly resolves the real argv via `kern.procargs2` (self-introspection doesn't need the cross-session entitlement other-process introspection does — the same asymmetry #5201 hit for `ps -o rss=`), and `AliveWithCmdline` correctly returns `true`. The test was asserting the wrong thing on this platform, not observing a real fail-open regression — production behavior here is correct and safe.

## Fix

Skip the test on darwin too, alongside the existing linux skip, with a comment naming why. Does not add a darwin-specific equivalent test for the same fail-closed property — that would need a way to force `kern.procargs2` itself to fail (e.g. an invalid/recycled PID), a separate, larger change than unblocking the current false failure; left for a follow-up if the coverage gap matters enough to close.

## Verification

`go test ./internal/pidutil/...` full package green, target test now `SKIP`s cleanly with an explanatory message; `go vet ./internal/pidutil/...` and gofmt clean.

No issue filed for this — found while diagnosing a locally-reproducible test failure on a clean `main` checkout, unrelated to any feature work.

### DIFF
--- a/cmd/gc/order_dispatch_bench_test.go
+++ b/cmd/gc/order_dispatch_bench_test.go
@@ -90,7 +90,7 @@ func BenchmarkOrderDispatchTick(b *testing.B) {
 
 	b.Run("PreFix_ReparsePerTick", func(b *testing.B) {
 		before := loadCityConfigCalls.Load()
-		ad := newMemoryOrderDispatcher(aa, cityDir, cfg, events.Discard, io.Discard)
+		ad := newMemoryOrderDispatcher(nil, aa, cityDir, cfg, events.Discard, io.Discard)
 		// Reproduce the pre-fix storeFn verbatim (order_dispatch.go:431-433
 		// before the ga-237xpr fix): ignore the dispatcher's cached cfg and
 		// go through the nil-cfg path, which reloads city.toml + every pack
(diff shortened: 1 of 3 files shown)


AJBcoding:
Opened #5291 as a coverage-preserving alternative to this skip: it ports the gas-x0dh / f3cd9345d pidutil adjustment and probes pid 1 on non-root Darwin so TestCmdline_FailsClosedWhenUnreadable keeps exercising the Cmdline-error fail-closed branch. Linking here so maintainers have both options in one thread.

sjarmak:
Thanks for catching this one. Three of us independently spotted that the Darwin skip dropped the fail-closed argv coverage, which says something about how visible the gap was.

We are landing #5746 for it, which takes the same approach. Closing here in favour of that one so we are not carrying four branches over one skip; the diagnosis is yours as much as anyone's.

PR #5255, opened 2026-08-13
fix(gc): clean up formula-owned external worktrees

caseymatt:
## Problem
Formula v2 workflows can create linked worktrees outside `cityPath/.gc/worktrees/<rig>`. The closed-Bead reaper only discovered conventional paths, leaving completed workflow worktrees registered indefinitely. Protected external worktrees also disappeared from actionable reporting.

## Decision
- Discover external candidates only from terminal Bead `gc.work_dir`/legacy `work_dir` ownership claims.
- Require the path to be an actual linked worktree of the configured rig.
- Preserve the existing age, borrow-veto, process/session liveness, dirty tree, stash, and unreachable-commit gates.
- Keep automatic reaping opt-in.
- Attribute a worktree shared by terminal workflow records to its path-named source anchor and reap once.
- Emit protected decisions/events for actionable graph-v2 ownership mismatches and unsafe git state.

## Verification
- `go test ./cmd/gc -run 'TestReapClosedBeadWorktrees|TestDiscoverReapCandidates|TestScanBorrowVetoReferences|TestExtractBeadIDFromWorktree' -count=1`
- `make test-fast-parallel` — all 10 jobs passed
- `go vet ./...`
- `git diff --check`
- pre-commit hook ran on the final regression commit

## Dogfood evidence
Gas City executed two independent implementation lanes whose changes textually conflicted in the shared candidate-discovery block. The integration worker resolved the conflict semantically, retaining external discovery plus protected-evidence reporting. The run also exposed and fixed the terminal shared-reference ownership regression before promotion.


quad341:
Hi @caseymatt — thanks for this, and apologies for the slow response. Cleaning up formula-owned external worktrees is real and needed work, and I don't want the delay to read as disinterest.

**The honest problem: this PR is too large for us to review, and that's a limitation on our side as much as a comment on the change.**

```
372 changed files
+44,957 / −2,777   (47,734 total changes)
```

Our review tooling feeds a PR's diff to a review model in one pass, and this exceeds the input limit outright — one of our reviewers rejects it at 1,048,576 characters before it can even start. So it isn't that the review came back negative; it's that no review can begin. A human reading 45,000 changed lines in one sitting has much the same problem, so I don't think this is only a tooling artifact.

**What would help most: split it into reviewable pieces.** Roughly 58% of the change (165 files, ~27,500 lines) is `_test.go`, which suggests a natural seam — but you know the change and I don't, so I'd rather ask than prescribe. Some options:

- A series of stacked PRs, each one self-contained and independently mergeable.
- Or a first PR with the core worktree-cleanup mechanism plus its tests, and follow-ups for the broader call-site updates.

Anything that gets individual pieces into the range where they can actually be read is a win. If you'd like, tell me how you'd naturally divide it and I'll make sure the pieces get looked at promptly rather than sitting in the queue.

**Two other things worth flagging honestly:**

The branch currently shows as conflicting with `main`, so it'll need a rebase regardless of how we proceed.

And separately — our review automation has been retrying this PR far more often than it should have, which is a bug on our end that we found while investigating something else. That's fixed-in-progress and none of it is your doing; I mention it only because if you saw repeated activity on this PR with no actual review posted, that's why, and it wasn't a signal about the change itself.

Thanks again for the contribution and for your patience.

PR #5257, opened 2026-08-13
fix(mail): add unread-mail TTL so never-read mail wisps age out (#5240)

### LINKED ISSUE #5240: nudge-mail-sweep doesn't close stale ephemeral message wisps
## Bug

The `nudge-mail-sweep` order (runs every 5 min) reports "0 stale mail beads" even when 78+ ephemeral message wisps with `status='open'` exist in the Dolt `wisps` table. These messages accumulate forever and clutter the dashboard.

## Reproduction

```bash
# 1. Check open wisps in Dolt
python3 -c "
import pymysql
c = pymysql.connect(host='127.0.0.1', port=39322, user='root', database='sverka')
cur = c.cursor()
cur.execute(\"SELECT COUNT(*) FROM wisps WHERE status='open' AND issue_type='message'\")
print('open message wisps:', cur.fetchone())
"
# → (78,)

# 2. Run sweep dry-run
gc order sweep-nudge-mail --dry-run
# → "nudge-mail-sweep: nothing to close (0 stale nudge beads, 0 stale mail beads)"
```

## Root cause

The sweep appears to only scan the `issues` table for stale mail/nudge beads. Ephemeral mail messages are stored in the `wisps` table with:
- `issue_type = 'message'`
- `ephemeral = 1`
- `status = 'open'`

These are never matched by the sweep logic.

## Impact

- Dashboard shows 78+ "ready" items that are just old mail messages
- During autonomous runs, ephemeral messages accumulate without cleanup
- Operators see false "ready" count and may think work is pending

## Environment

- Gas City: 1.4.0 (Homebrew)
- bd: 1.1.2 (Homebrew)
- Dolt: 8.0.31

## Expected behavior

`nudge-mail-sweep` should close ephemeral message wisps that are past their TTL, same as it closes stale nudge beads. The `--mail-ttl` flag should apply to wisps with `issue_type='message' AND ephemeral=1`.

## Suggested fix

Extend the sweep query to also scan the `wisps` table:
```sql
SELECT id FROM wisps
WHERE status = 'open'
  AND issue_type = 'message'
  AND ephemeral = 1
  AND created_at < NOW() - INTERVAL ? SECOND
```

rjgeng:
Fixes #5240.

## What's broken

The reporter's theory was that `nudge-mail-sweep` only scans the `issues` table and skips `wisps`. Traced the actual query (`internal/mail/beadmail/beadmail.go`'s `readMessagesBefore`): it already unions both tiers via `TierMode: TierBoth`, so that's not it.

Real root cause: the query also filters `Label: "read"`. The sweep only ever closes mail that's already been read. A message bead that's **never** read — a broadcast nobody consumes, mail sent to a session that closed before reading it — never acquires that label and accumulates forever regardless of age. Matches the reporter's 78-wisp symptom exactly, just a different mechanism than they guessed.

## Why this needed a policy call, not just a query fix

`isRemovedMessageBead`'s doc comments treat "unread mail stays forever until read" as an intentional inbox-visibility guarantee, not an oversight — so this isn't a pure bug, it's a missing TTL policy. Went with a reasoned default rather than assuming; open to pushback on the specific number.

## Fix

Added a third sweep phase gated on a new `nudgeMailSweepDefaultUnreadMailTTL` = **7 days** (vs. the existing 60-minute read-mail TTL) — long enough that unread mail keeps its visibility window, short enough it doesn't accumulate forever. 7 days matches this repo's other retention windows (e.g. `sessionReconcilerTraceMaxAge`). New `--unread-mail-ttl` CLI flag mirrors the existing `--nudge-ttl`/`--mail-ttl`, so it's overridable if 7 days is wrong for your setup.

New `SweepUnreadMessagesBefore`/`CountUnreadMessagesBefore` in `beadmail.go` mirror the existing read-mail pair, with a new `UnreadRetentionSweepCloseReason` recognized by `isRemovedMessageBead` alongside the existing `RetentionSweepCloseReason` — swept unread mail stays addressable until purge, exactly like swept read mail does today. No cache/reconciler machinery touched.

**Checked for interaction with #5051** (mark-read+close auto-handoff fix, already on `main`): no conflict — #5051 marks-read and closes eagerly at archive time, so it never leaves an open *unread* bead behind for this new phase to find. Confirmed by running both packages' full test suites together, including #5051's own tests — all green.

## Verification

- `go test ./cmd/gc/... ./internal/mail/beadmail/...` full packages green (2 new RED→GREEN tests plus all 60 pre-existing tests in both packages).
- gofmt clean; `go vet` clean on changed packages; `go build ./...` clean.
- Manual `gc order sweep-nudge-mail --help` confirms the new flag.


csauer02-personal-user:
Adjacent finding from a live city, offered as a companion to this PR's close_reason work since you're already extending that vocabulary: **`Provider.Archive` closes with no `close_reason` at all**, so `isRemovedMessageBead` classifies archived mail as user-removed and Get/peek answer ErrNotFound while the body sits intact — never-read mail archived by rotation becomes irrecoverable through the mail layer (our census: 599 closed-never-read message beads since 08-01 on one city; recovery required raw `bd show`). Archive's own doc comment promises the body stays 'retrievable via gc mail peek or bd show'.

One wrinkle worth knowing before anyone reaches for the obvious fix: stamping Archive with `RetentionSweepCloseReason` makes Get/Read/MarkRead serve archived mail again, which **fails the conformance contract** (`Archive_RemovesMessageFromEveryView`). The two contracts are only consistent if *peek* is a distinct read-through verb — and today peek routes to Get. Our companion fix (proven locally against a five-arm falsifier; the archive-then-peek arm flips FAIL→PASS, conformance suite untouched):

1. Archive stamps a new `ArchiveCloseReason` (discriminator + audit-visible), following your reason-constant convention.
2. An optional-capability `mail.ArchivePeeker` interface; beadmail's `Peek` reads through archive-stamped mail while still hiding Delete-purged and legacy closed-without-reason beads (upgrade-path contract intact).
3. `gc mail peek` uses it, treating an API not_found as a fall-back-to-local cue (the API serves the Get view; the remote-city gate still hard-errors).

Happy to send it as a follow-up onto this branch in your test shape once this lands (or as a standalone PR if you'd rather keep scopes separate) — flagging now so your close-path work and this don't drift apart on the classifier semantics.

quad341:
@rjgeng, thank you for tackling #5240, and sorry this sat so long after you invited pushback on the number. Tracing `readMessagesBefore` to its `Label: "read"` filter, rather than stopping at the reporter's theory, cracked it, and flagging the TTL as a policy call was the right instinct.

**Decision: keep the sweep, but make it opt-in and off by default.** As written, every city closes never-read mail after 7 days (dropping it from inboxes) with no way to turn that off: the watchdog in `runNudgeMailSweepWatchdog` passes `nudgeMailSweepDefaultUnreadMailTTL` directly, and the core `nudge-mail-sweep` order runs `gc order sweep-nudge-mail --quiet`, which gets the same default.

Could you:
1. Set `nudgeMailSweepDefaultUnreadMailTTL` to `0`, meaning disabled.
2. Skip phase 3 in `sweepStaleNudgeMail` and `countStaleNudgeMail` when `unreadMailTTL <= 0`, with a test. Today a zero TTL puts the cutoff at `now`, which would close every unread message.
3. Let `--unread-mail-ttl` accept `0` and reject only negatives.
4. Add the city opt-in under `[mail]` beside `retention_ttl`, which already treats empty or `"0"` as disabled: say `unread_retention_ttl = "168h"`, read by the watchdog and by `gc order sweep-nudge-mail` when the flag isn't set. Then run `make generate`.
5. Also from our review: unread twins of `TestCountReadMessagesBefore_CountsWithoutMutating`, `TestCountReadMessagesBefore_LimitCapsCount`, `TestSweepReadMessagesBefore_PerBeadCloseErrorIsCollected` and `TestSweepReadMessagesBefore_ListErrorIsFatal` in `beadmail_retention_test.go`.

Heads-up: we pushed the Aug 30 merge from main and `e2c3046` to your branch (both show you as author), so pull first. It still merges cleanly; no rebase needed.

If you'd rather, we can make these changes ourselves and merge with your authorship kept.

Issue #1039, opened 2026-04-20
bug: formula step beads pollute bd ready queue and persist as orphans when molecule root never self-closes


sjarmak:
### Before you continue

- [x] I searched existing issues and did not find a duplicate.
- [x] I read the relevant docs and contributor guidance.

### Gas City version

0.15.2 (2bcedae9)

### Environment

Linux x86_64, bash, tmux provider, managed-dolt single city

### Reproduction

1. Dispatch a formula to a pool agent:
   ```
   gc sling <agent> <work-bead> -f mol-focus-review
   ```
2. Inspect the target store:
   ```
   gc bd --rig <rig> ready
   ```
3. Observe that per-step child beads (e.g. `Load context`, `Execute /focus`, `Run tests`, `Finalize workflow`) appear in `bd ready` output alongside real work beads.
4. Let several formulas run. Over time, observe that a subset of molecule roots never transition from `in_progress` to `closed` even after their target work bead closes (`mol-focus-review` is the most frequent offender in our deployment). Their step children accumulate as open beads.

### Expected behavior

Formula steps are execution scaffolding and should not be visible as ready work to unrelated pool workers. Either:
- Steps are delivered as prompt content only (no store materialization) — preferred; matches the `bd mol wisp` "vapor phase" framing in the docs.
- Step beads are materialized but filtered out of `bd ready` by default, and a matching exclusion is applied by worker claim logic.
- Molecule root beads auto-close when all step children reach a terminal state, so TTL-gated wisp GC can collect the closure.

### Actual behavior

1. **`bd ready` pollution.** `internal/beads/beads.go:84-93` defines `readyExcludeTypes` which filters `molecule`, `message`, `session`, `agent`, `role`, `rig`, `gate`, and `merge-request` — but **not `task`**. Step beads are created with default type `"task"` (`internal/molecule/molecule.go` `stepToBead`, ~L430), so they flow through `bd ready` indistinguishable from real work. After a few formula dispatches, step names dominate `bd ready`.

2. **Worker confusion.** Pool workers unrelated to the formula run claim step beads. Example from our deployment: `codeprobe-fwuvs` claimed `Run tests to verify implementation` with no context of which tests or what was implemented.

[...]

PR #5258, opened 2026-08-14
fix(dispatch): default gc.kind on plain attempt-recipe children (#5246)

### LINKED ISSUE #5246: formulas-v2 dispatch creates step beads with no Type — canonical injector types them "step"
## Observation

`internal/dispatch` (control.go and ralph.go expansion/retry paths) creates step beads stamped with `gc.step_ref` but with **no `Type` set and no `gc.kind`**, while the conventional injector `cmd/gc/wisp_step_inject.go` types its step beads `"step"`. Same conceptual object, two typing conventions — one of them empty.

Field numbers from a live store: 443 of 779 live `ref_only` step beads carry no `gc.kind` (and the empty-Type population is what surfaced this — see below).

## Why it matters

Every scope predicate written as "non-task type ⇒ control-plane" silently passes these beads: empty Type reads as gated/work-class to any consumer using the type arm. The concrete instance is the ADR-0009 work-record close gate, where untyped step beads fell inside the work-record contract and `GC_WORK_RECORD_ENFORCE` refused their closes (two-arm control receipts in gastownhall/gascity#5103 discussion; the symptom-side fix — excluding on step markers — is AJBcoding/gascity#2, riding #5103). That fix is deliberately marker-based so it stays correct regardless of typing, but it also means the typing inconsistency is now masked rather than fixed — hence this issue.

## Question for maintainers

Should the v2 dispatch creation paths type their step beads (`"step"`, matching `wisp_step_inject.go`), or is empty-Type-with-markers the intended convention for dispatch-created steps? If the former, happy to send the patch; it looks like a handful of creation sites in `internal/dispatch`.

rjgeng:
Fixes #5246.

## The issue's framing was a trap

The reporter posed an either/or: should dispatch step beads be typed `step` (matching `wisp_step_inject.go`), or is empty-`Type` intended? No proposed remedy — a design question, not a bug report with a known fix.

Traced both paths before picking either:

- **`wisp_step_inject.go` doesn't set `Type` at all** — it only *queries* `Type=="step"`. The issue's premise (that the injector types its beads `step`) is imprecise. The real setter is `molecule.go`'s `nonRootStepBeadType` (#1039 coercion).
- **`buildAttemptRecipe`'s root deliberately skips that coercion** — `preservesGraphActionTypes` checks for `gc.attempt` + `gc.step_ref`, which every attempt-recipe root always carries, with an explicit comment that the skip is load-bearing so `Ready()` still surfaces these beads for worker claim. Forcing `Type` to `"step"` (what the issue's title implies) would risk breaking worker-claim visibility for every ralph/retry-spawned attempt across the engine — contradicts an explicit, commented design decision. Not built.

## The real, narrower gap

The root of every attempt recipe gets `gc.kind` stamped unconditionally (`control.go:747`). Non-root children only get it when they carry `Retry`, `Ralph`, or `Drain` config — a plain leaf child (e.g. a scoped-ralph's ordinary `apply`/`verify` steps) gets `Type: "task"` but **no `gc.kind` at all**. That's exactly what `cmd/gc/work_record_gate.go`'s `isWorkRecordGatedBead` treats as gated-by-default (`(Type=="" || Type=="task") && gc.kind==""`) for the ADR-0009 work-record close gate — confirms the issue's severity claim (443/779 affected beads in the reporter's live store) with the actual mechanism, not the guessed one.

## Fix

One call site: default `childMeta[gc.kind]` to `task` when none of the Retry/Ralph/Drain branches already set it, mirroring the root's unconditional stamp. Doesn't touch `Type`, worker-claim logic, or any reconciler/cache machinery.

## Verification

- **TDD**: `TestBuildAttemptRecipeRalphWithChildren` already built exactly this shape (plain children under a scoped ralph) but never asserted their `gc.kind` — extended it, confirmed RED (`gc.kind == ""`) before the fix, GREEN after.
- `go test ./internal/dispatch/... ./internal/formula/...` full packages green.
- `go test ./cmd/gc/... -run 'WorkRecord|Gate'` green (confirms this closes the gate-exposure gap cleanly).
- gofmt/vet clean; `go build ./...` clean.

### DIFF
--- a/cmd/gc/cmd_mail_test.go
+++ b/cmd/gc/cmd_mail_test.go
@@ -3606,6 +3606,27 @@ func TestMailCheckInjectDoesNotCloseBeads(t *testing.T) {
 	}
 }
 
+// assertAutoHandoffRetainedAddressable pins the dip-6ov51a contract at the inject
+// integration boundary: an injected auto-handoff is MARK-READ + CLOSED and stamped
+// with the retention marker (retain-addressable), NOT hard-deleted, so an
+// unconsumed handoff stays recoverable until the read-gated TTL sweep reclaims it.
+func assertAutoHandoffRetainedAddressable(t *testing.T, store beads.Store, id string) {
+	t.Helper()
+	b, err := store.Get(id)
+	if err != nil {
+		t.Fatalf("auto handoff %s must be retained addressable after injection, not hard-deleted; store.Get = %v", id, err)
+	}
+	if b.Status != "closed" {
+		t.Errorf("auto handoff %s status = %q, want closed", id, b.Status)
+	}
+	if got := b.Metadata[mail.ReadMetadataKey]; got != "true" {
+		t.Errorf("auto handoff %s metadata[%q] = %q, want %q (marked read)", id, mail.ReadMetadataKey, got, "true")
+	}
+	if got := b.Metadata["close_reason"]; got != beadmail.RetentionSweepCloseReason {
+		t.Errorf("auto handoff %s close_reason = %q, want %q (retain-addressable)", id, got, beadmail.RetentionSweepCloseReason)
+	}
+}
+
 func TestMailCheckInjectArchivesAutoHandoffMessages(t *testing.T) {
 	store := beads.NewMemStore()
 	mp := beadmail.New(store)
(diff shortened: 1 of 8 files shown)


quad341 (approved):
@rjgeng — the best part of this PR is what you *didn't* build. #5246 offered an either/or, and you checked both before picking: that `wisp_step_inject.go` only ever *queries* `Type=="step"` and never sets it, and that `preservesGraphActionTypes` skipping the #1039 coercion is a deliberate, commented decision protecting worker-claim visibility. Forcing `Type` there was the obvious patch and would have been the wrong one. Finding the actual gap instead — the root gets `gc.kind` unconditionally, plain children get nothing — is exactly the shape of bug fix I want.

I verified it rather than taking the description's word: the gate predicate on `main` is still what you say it is, reverting just your `control.go` hunk turns your extended test red with `gc.kind = ""`, and restoring it goes green across dispatch, formula, molecule, graphv2 and the `cmd/gc` gate tests.

Merging now. Ignore the 8-file diff — your branch picked up two commits that had already landed on `main`, so the merge resolves to exactly your 19 lines. Nothing for you to do.

If you're up for a follow-up: the compile path still stamps no `gc.kind`, so ralph iteration 1 stays gated while iterations 2+ now aren't.


---

PR #5259, opened 2026-08-14
ci(govulncheck): bump the scanner pin past the DB-triggered panic

julianknutsen:
govulncheck v1.4.0 (x/tools v0.46.0) crashes on this module since the
2026-08-13 21:43 vuln-DB update: "panic: ForEachElement called on type
containing *types.TypeParam" (element.go:130), with a racing secondary
nil-deref at vulncheck/source.go:281 — a truncated SARIF and a failed
upload, on every PR and on main itself (last green scan predates the DB
update). The crash is scanner-side generics handling under go1.26, not a
finding.

v1.7.0 verified against origin/main with GOTOOLCHAIN=go1.26.5: completes
without panic, reports 1 imported + 4 module-level vulnerabilities none of
which are called — the informational tiers the workflow already
distinguishes. One-line pin bump; the tier semantics are untouched.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

Repo-wide: this crash currently fails the scan on every open PR (verified identical on at least four) and on main's own post-21:43 runs. Verification detail in the commit message.


julianknutsen:
Scope grew by one commit, and the story completes: the scan failure on this branch was NOT residual scanner breakage — v1.7.0 was correctly reporting **called standard-library vulnerabilities** (net/http, crypto/tls, html/template, net/url; GO-2026-6088..6091, -6218) fixed in go1.26.6. The same Aug-13 DB update that added those advisories is what crashed v1.4.0 — so main has been blind to real called vulns since yesterday evening, mislabeled as scanner infra failure on every PR. Second commit bumps the go directive to 1.26.6; verified locally: scanner v1.7.0 exits 0 with 'Your code is affected by 0 vulnerabilities' (4 module-level findings remain, non-gating by the workflow's own tier policy, whose exit-code premise holds on v1.7.0).

PR #5261, opened 2026-08-14
docs(session): clarify primed_at as delivery-attempted, not consumed (#5236)

### LINKED ISSUE #5236: Clarify primed_at as startup delivery-attempt metadata
## Problem

`primed_at` is stamped when startup delivery is selected and runtime start succeeds, but it can be misread as proof that an agent consumed or executed its startup instruction. A successful transport delivery can still leave a live, idle worker when the provider drops submission.

The durable acknowledgement that work actually began is the trigger bead becoming assigned/in-progress, not `primed_at`.

## Evidence

- `cmd/gc/session_lifecycle_parallel.go` derives prompt-delivery state from the planned route and stamps `primed_at` after successful runtime start.
- `internal/runtime/herdr/provider.go` treats startup delivery as best effort.
- In the qualifying incident, the worker had `primed_at` while its trigger remained open and unassigned until a manual nudge caused the claim.

## Compact remedy

Clarify comments, diagnostics, and user-facing wording so `primed_at` means startup prompt delivery was selected/attempted for that runtime incarnation. Preserve metadata compatibility and lifecycle behavior. If a stronger acknowledgement is needed later, expose a distinct field rather than silently redefining this one.

## Out of scope

Provider protocol redesign, stalled-claim recovery, or treating a transport acknowledgement as proof of work execution.

rjgeng:
Fixes #5236.

## What's misleading

`primed_at` is stamped when delivery was *selected and runtime start succeeded* — but several doc comments described this as "confirmed delivered" / "actually delivers," which reads as proof the agent consumed the prompt. It isn't proof of that: the issue's own qualifying incident had a worker holding `primed_at` while its trigger bead stayed open/unassigned until a manual nudge forced the claim.

## Fix — comment/doc-string only, zero behavior or schema change

- `internal/session/lifecycle_transition.go` — `PrimedAtMetadataKey`'s doc said "confirmed delivered" with no caveat; added that "confirmed" means transport succeeded, not consumption, and pointed at trigger-bead assignment as the real work-started signal.
- `cmd/gc/prompt_delivery.go` — `promptDeliveryResult`/`Delivered` docs said "actually delivered it," but `promptDelivery()` does no I/O at all (pure routing decision, per its own existing doc) — clarified `Delivered` means "a mechanism was selected."
- `cmd/gc/session_lifecycle_parallel.go` — `promptDelivered` field doc said "actually delivers the rendered startup prompt" (one of the issue's own evidence citations) — same fix.

**Deliberately left `session_level_converge.go` untouched**, even though it also uses "CONFIRMED" language for `primedAt`: that's the S19 Stage-1 migration module ("nothing reads the action list yet ... until Stage 3 wires the act loop"), and its terminology is already precisely scoped within its own durable-world/runtime-world framework — it separately tracks `primedEnv` as "delivered this runtime but may not yet be durably recorded," meaning it already distinguishes the two states this issue wants clarified. Touching an in-progress migration module's carefully-established vocabulary without Stage 3/4 context felt like the wrong call. `internal/runtime/herdr/provider.go` (the issue's other evidence citation) was already correctly worded — no change needed there either.

## Verification

`go test ./internal/session/... -run 'Primed|Priming|Prompt'` and `go test ./cmd/gc/... -run 'PromptDelivery|Primed|Priming'` both green (comment-only, no regressions); gofmt clean; `go vet` clean on changed + dependent packages; `go build ./...` clean.


Toady00:
No objection to the reframing — `promptDelivery()` genuinely does no I/O, so "attempted" is the accurate word for what the field holds today.

Flagging one consumer that isn't mentioned in #5236 or in this PR, in case it changes how the redefinition should land.

`cmd/gc/session_level_converge.go:~234` gates the 60s lost-prompt retry on this field:

```go
if r.observed && r.live && d.promptConfigured && d.primedAt == "" {
	switch {
	case r.primedEnv:        actions = append(actions, actionStampPrimedFromRuntime)
	case attemptEligible(d): actions = append(actions, actionAttemptPrime)
	}
}
```

That predicate is only meaningful under the *old* reading. Under "attempted", it asks "did we attempt?" — which is true whenever `promptConfigured` is true — so `actionAttemptPrime` is unreachable and a lost prompt gets no ≤60s recovery. It falls through to the ~6 minute supervisor recycle instead. I've seen that stall twice on fresh pool-worker spawns (claude and codex), both with `primed_at` stamped against a pristine harness screen.

I don't think that argues against this PR — the wording change is correct either way. It's that #5236 lists "stalled-claim recovery" as out of scope while this field is what that recovery keys on, so clarifying the name without moving the gate leaves the retry dead by definition rather than by accident.

The confirming signal the gate wants already exists in-tree: `r.primedEnv` (`GC_STARTUP_PROMPT_DELIVERED`), with `actionStampPrimedFromRuntime` right there in the same switch, plus herdr's `GC_HERDR_STARTUP_DELIVERY_UNCONFIRMED` sidecar key.

Filed separately as #5340 with the full evidence, including what I could not reproduce (0/3 attempts on a later day — it's a race), so it doesn't clutter this PR. Happy for it to be handled as a follow-up; mainly wanted the gate on record before the rename lands.

quad341 (approved):
@rjgeng — thank you for this. What I appreciate most is what you *didn't* touch: leaving `session_level_converge.go`'s durable-world/runtime-world vocabulary alone because it already separates `primedEnv` from `primedAt` was exactly right, and spelling out why in the PR body saved me the investigation. Restraint on an in-flight migration module is the hard call and you made it correctly.

Two things I'll take on my side, both ours not yours. First a rebase onto current main, which clears the two-week-old Go vuln-scan failure — that's branch age, not your change.

Second, one wording tightening in `lifecycle_transition.go`. It still says "the transport succeeded," and herdr is the counter-example: when `deliverStartupTurn` fails, `Start` records `GC_HERDR_STARTUP_DELIVERY_UNCONFIRMED` and returns nil anyway, so `primed_at` gets stamped on a prompt nobody confirmed arrived. I'll match it to the "attempted" language you already used in `prompt_delivery.go` — your framing, applied one file further.

Worth a read: @Toady00's #5340. Your clarification is what makes that dead retry gate legible.


---

PR #5263, opened 2026-08-14
fix(gc): route API calls to the supervisor hosting the city

### LINKED ISSUE #5262: gc session submit fails with "ACP provider does not own session" for supervisor-managed cities
## Summary

For a supervisor-managed city, `gc session submit` cannot reach an ACP session and fails with:

```
gc session submit: sending message to session: session not found: ACP provider does not own session "mayor"
```

The same root cause affects every `apiClient` caller: a supervisor-managed city is routed to an HTTP endpoint that nothing is serving, so the API call fails and the command silently drops into its local fallback.

## Root cause

A standalone controller (`gc controller` / `gc serve`) and a supervisor-managed city both answer the per-city controller socket, because the supervisor hosts that controller in-process. Liveness alone therefore cannot tell the two apart.

`cmd/gc/apiroute.go` treats any live socket as standalone:

```go
if apiRouteControllerAliveHook(cityPath) != 0 {
    return standaloneControllerClient(cityPath)
}
return apiRouteSupervisorClientHook(cityPath)
```

So for a supervisor-managed city:

1. `controllerAlive` returns non-zero, because the supervisor hosts the controller in-process.
2. `apiClient` returns `standaloneControllerClient(cityPath)`.
3. Supervisor mode ignores the city's `[api]` port and serves the city on the supervisor's own port via city-scoped routes.
4. With `[api] port = 9443` in `city.toml`, the returned client targets a listener that does not exist.
5. `SubmitSession` fails with a connection error, and `api.ShouldFallback` sends the command to its local fallback.
6. The local fallback constructs a new, process-local ACP provider.
7. That provider does not own the supervisor's in-memory ACP connection, so `internal/runtime/acp/acp.go` returns `ErrSessionNotFound`.

Without an `[api]` port the same path returns `nil`, which also lands in the local fallback with the same result.

The local fallback is only sound for state that lives on disk. A process-owned runtime such as an ACP session lives in the supervisor's memory and cannot be reached from a separate CLI process, so falling back can never succeed for it.

## Evidence

- Both `opencode acp` processes were children of the supervisor process.
- The supervisor logged that the city's `[api] port = 9443` is ignored.
- Nothing was listening on port 9443.
- `gc session submit mayor "..."` failed with `ACP provider does not own session "mayor"`.

## Expected behavior

When the supervisor hosts the city, `apiClient` should return the supervisor-managed client, since that is the only endpoint serving the city. A standalone controller should keep using its configured `[api]` port.

## Notes

The controller already reports how it is hosted over the same socket (`identify` → `hosting_mode: supervisor|standalone`), and `ensureNoStandaloneController` already uses that signal for exactly this kind of ownership decision. Routing can key on it without an additional probe, and a controller predating the identity command can keep the existing standalone-only routing rather than having a role inferred for it.

## Environment

- `gc` built from `main`
- macOS (arm64), supervisor-managed city registered and running

Toady00:
## Problem

A standalone controller (`gc controller` / `gc serve`) and a supervisor-managed city both answer the per-city controller socket, because the supervisor hosts that controller in-process. Liveness alone cannot tell them apart, but `apiClient` treated every live socket as standalone:

```go
if apiRouteControllerAliveHook(cityPath) != 0 {
    return standaloneControllerClient(cityPath)
}
return apiRouteSupervisorClientHook(cityPath)
```

Supervisor mode ignores the city's `[api]` port and serves the city on the supervisor's own port via city-scoped routes. So for a supervisor-managed city with `[api] port = 9443`, the returned client targeted a listener that does not exist; the call failed and `api.ShouldFallback` sent the command into its local fallback. Without an `[api]` port the same path returned `nil` and reached the same fallback.

The local fallback is only sound for state that lives on disk. An ACP session lives in the supervisor's memory, so the fallback built a fresh process-local provider that did not own the connection:

```
gc session submit: sending message to session: session not found: ACP provider does not own session "mayor"
```

Fixes #5262.

## Fix

Key the routing on the hosting mode the controller already reports over the same socket (`identify` → `hosting_mode`), the signal `ensureNoStandaloneController` already uses for this ownership question:

- **supervisor-hosted** → supervisor client. A `nil` result means there is no usable API, so it returns that rather than a standalone port nothing is serving.
- **standalone** → the `[api]` port endpoint when configured, else `nil` for the caller's local fallback.
- **unknown** (a controller predating the identity command) → the previous standalone-only routing, rather than inferring a role.
- **socket down** → supervisor client, unchanged.

The identity probe replaces the liveness ping rather than adding a round-trip, since it reports the PID too and already falls back to `controllerAlive` for older controllers.

`maintenanceAPIClient` keeps its supervisor fall-through, which now covers only the unknown-mode legacy controller, because `apiClient` resolves the supervisor-hosted case directly. Its comment says so.

## Tests

`TestAPIClientRouting` now pins every seam the ladder consults through one helper, so each subtest is hermetic regardless of which hook the implementation reads, and covers:

- supervisor-hosted uses the supervisor client
- **supervisor-hosted ignores a configured standalone `[api]` port** — the regression for this bug
- supervisor-hosted with no supervisor client returns `nil` rather than a dead standalone endpoint
- standalone-hosted with/without an `[api]` port
- legacy unknown hosting with/without an `[api]` port (pre-existing behavior preserved)
- controller down, and the `GC_NO_API` escape hatch

Verified failure direction first: against the unmodified implementation the three supervisor-hosted cases fail, including one that returns a client aimed at the dead `9443` endpoint. The standalone, legacy-unknown, controller-down and escape-hatch cases pass before and after, so the pre-existing routing is unchanged.

`TestMaintenanceAPIClientRoutesToSupervisor` now pins the hosting mode to unknown so it exercises the legacy fall-through it is named for, instead of passing incidentally through the controller-down path.

## Test plan

- `go test ./cmd/gc/ -run 'TestAPIClientRouting|TestMaintenanceAPIClientRoutesToSupervisor|TestStandaloneControllerClient' -count=1` passes
- Full `go test ./cmd/gc/ -count=1` passes
- `go vet ./cmd/gc/`, `go build ./...`, and `golangci-lint` (via the pre-commit hook) clean

Two unrelated environment-induced failure classes appear on this machine and are not caused by this change: fixture commits fail when a global `commit.gpgsign=true` leaks into tests, and one template test fails when `OPENAI_API_KEY` is present in the environment. Both pass when those are neutralized.

### DIFF
--- a/cmd/gc/apiroute.go
+++ b/cmd/gc/apiroute.go
@@ -32,13 +32,29 @@ var (
 //
 // A standalone controller (gc controller / gc serve) and a supervisor-managed
 // city both answer the per-city controller socket — the supervisor hosts that
-// controller in-process. When the socket is alive, apiClient routes to the
-// standalone HTTP endpoint if the city configures an [api] port, otherwise
-// returns nil so the caller uses its local fallback; when the socket is not
-// alive it returns the supervisor-managed client. Maintenance commands have no
-// local fallback, so they use maintenanceAPIClient, which additionally routes a
-// supervisor-managed city (alive socket, no standalone [api] port) to the
-// supervisor client rather than reporting controller-down. (gascity ga-tp7)
+// controller in-process — so liveness alone cannot say which endpoint serves
+// the API. The controller reports that itself, and routing keys on it:
+//
+//   - supervisor-hosted: the supervisor serves this city on its own port via
+//     city-scoped routes and IGNORES the city's [api] port, so route to the
+//     supervisor client. Routing to the configured standalone port instead
+//     sends every call to a listener that does not exist; the call fails and
+//     the caller drops into a local fallback that constructs a fresh,
+//     process-local runtime. That fallback cannot reach a runtime the
+//     supervisor owns in memory — an ACP session lives in the supervisor
+//     process, so `gc session submit` failed with "ACP provider does not own
+//     session" while the supervisor held it. Falling back is only sound for
+//     state that lives on disk, not for process-owned runtimes.
+//   - standalone: route to the standalone HTTP endpoint when city.toml
+//     configures a usable [api] port, else return nil for the local fallback.
+//   - unknown (a controller predating the identity command): keep the
+//     pre-existing standalone-only routing rather than inferring a role.
+//
+// When the socket is not alive it returns the supervisor-managed client.
+// Maintenance commands have no local fallback, so they use
+// maintenanceAPIClient, which additionally routes an unknown-mode managed city
+// (alive socket, no standalone [api] port) to the supervisor client rather
+// than reporting controller-down. (gascity ga-tp7)
 func apiClient(cityPath string) *api.Client {
 	// Remote routing is NOT handled here. A remote target is refused upstream by
 	// the capability gate in resolveContext (Phase 1) and, once enabled, will be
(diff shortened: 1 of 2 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277172882

PR #5265, opened 2026-08-14
fix(hook): turn-bound claims — a claim may only be born inside a turn that will execute it

julianknutsen:
Closes the claim-orphaning class (ga-fylee): claims born from `gc`/`bd` processes outliving their killed tool calls — the mechanism behind idle workers hoarding 2-3 claims each on maintainer-city (14/14 orphan claims in the forensic window traced to the codex exec host falsely reporting slow commands complete-empty at ~12s while `gc hook --claim` ran 15-70s).

The contract enforced: **a claim may only be born inside a turn that will execute it.**

- **F-A callback fence**: hook-callback lanes (`gc hook run` children, rendered session hooks) cannot claim — new `non_turn_context` drain reason, exit 0, no drain-ack consumed; fenced at the CLI pre-query AND `tryHookClaim` through one shared predicate.
- **F-B claim window** (`GC_HOOK_CLAIM_WINDOW`, default 45s): no CAS after the window; expiry is exit 1 + typed `execution.claim_window_expired` + NO drain record (seat stays retriable); claim-child ctx bounded by the remaining window; a CAS that lands as the window expires self-releases (straddle re-check).
- **F-C delivery unwind**: an undeliverable result line releases the claim through the ops seam that made it + `bead.claim_released`. Requires `signal.Ignore(SIGPIPE)` (installed at the `mainExitCode` funnel): without it the process dies at exit 141 before `Write` returns — proven by a real-closed-pipe subprocess test with an unprotected control demonstrating the death. Behavior note: `gc … | head` now exits 1 with a stderr line instead of dying silently at 141.
- **F-D**: drain-ack releases still-held in-progress claims across work + rigs + relocated binding (15s overall budget) before setting the ack — "drain-ack means I hold nothing."
- **BUG-2**: the existing-assignment tier reads the federated `in_progress` view (with serving-leg `blocked_by` resolution), so a session always sees its own claims before fresh-claiming — the hoard state is unrepresentable.
- Compensation pair documented for event-sourcing consumers: `bead.claim_released` after `execution.step_started` = the step never executed. Ops doc: `engdocs/contributors/turn-bound-claims-operations.md`.

Full lineage on ga-fylee (diagnosis overturned twice by evidence), design doc `turn-bound-claims-design.md`, 2-lens adversarial council + fix round (the SIGPIPE unreachability was council-caught with an empirical probe). Interim deploy note: cities with 15-70s claim latency need `GC_HOOK_CLAIM_WINDOW=75s` on affected pools until the residency fast-path (ga-4qdfn / resolver S0 #5264) lands; retirement signal is `claim_window_expired{parent_alive=true}` at zero.


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277168125

PR #5266, opened 2026-08-14
fix(ci): reconcile the residency baseline with the #5250/#5264 merge race

julianknutsen:
Main is red for every open PR: #5250 (merged 07:56:10) and #5264 (merged 07:56:11) were each green alone, but S0's boundary baseline was computed before #5250's F3 refactor of `reachableStoresForSessionInfo`. Delta is exactly that one function — two raw `beads.Store` literals replaced by `workAssignmentStores` calls (net enumeration sites unchanged at 123; the literal rows retire as a shrink, the calls stay S2's migration target). Both guard halves re-emitted with their own tools and green.

Main-red unbreaker — same class as #5247.


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277168056

PR #5269, opened 2026-08-14
fix(session): recover empty resume identity after startup death

tudorsaitoc:
## Summary

This replaces #5160 because its `actual-software/gascity` head repository does not grant push access to `tudorsaitoc`.

- Refreshes the empty-`session_key` recovery onto upstream main `5a600ec65dd61b3e73b46f155cdaa2be8d7b4395`.
- Lets both startup paths recover when the persisted key cleared after command construction but the launch command still contains a resume shape.
- Strips the resume shape and retries once as a fresh start.
- Declines the retry when no resume shape changed, then returns the original startup error.
- Records the behavior in the required session scenario ledger.

## Why recovery belongs here

Command construction already avoids adding a resume flag when `session_key` is empty. The failure occurs when the command was built while the key existed, then the key cleared before startup. Recovery must therefore reconcile the stale command with the current persisted identity.

## Verification

- [x] The four focused empty-key tests pass:
  - `TestEnsureRunning_EmptySessionKeyStripsResumeAndStartsFresh`
  - `TestEnsureRunning_EmptySessionKeyWithoutResumeShapeDoesNotRelaunch`
  - `TestStartRuntimeOnly_EmptySessionKeyStripsResumeAndStartsFresh`
  - `TestStartRuntimeOnly_EmptySessionKeyWithoutResumeShapeDoesNotRelaunch`
- [x] `go test -count=1 ./internal/session/...`
- [x] Affected-package lint completed with `0 issues`.
- [x] `make vet`
- [x] `make check-docs` through the pre-commit hook.
- [x] Fresh GitHub CI for this replacement head.

The local broad fast gate also ran. It reached unrelated macOS host prerequisites: `flock` is absent, two doctor tests reject the local `bd-mac-wrapper` fixture, and one portable PID test fails. The focused package, lint, vet, and docs gates are green. Fresh Linux CI passed for head `1dfecb8cae439c1ff1f3178ded30b241983452dc`.

## Scope

- `internal/session/chat.go`
- `internal/session/chat_empty_session_key_test.go`
- `internal/session/REQUIREMENTS.md`


tudorsaitoc:
Ready for maintainer merge at exact head `1dfecb8cae439c1ff1f3178ded30b241983452dc`.  The refreshed full CI run completed with 43 successes, 15 intentional skips, and zero failures.  Focused empty-session recovery tests, the full `internal/session` suite, lint, vet, CodeQL, Mac regression, review formulas, and govulncheck are green.  Our current GitHub token does not have `MergePullRequest` permission, so a maintainer merge is the remaining gate.

tudorsaitoc:
@julianknutsen, the branch is current again.  Could you merge this exact green head when you have a moment?

- Head: `2360d63b33b954622a539060b07d69efa3c8d919`
- Current base: `f6741d94861aa14f0253deffbe9efb1cb3a35d92`
- Range difference: all four commits are unchanged, with no conflicts.
- Fresh exact-head CI: 58 successes, 26 intentional skips, and zero failures.

The current token has read-only repository access and cannot merge.  This remains the upstream release gate for our production worker recovery.

tudorsaitoc:
@julianknutsen, I rechecked this against current main `b63623d08ecf565de82c226f0af1ca2fc359d45d` after #5280. That commit only changes `cmd/gc`, `internal/storeref`, and residency scripts, so it does not overlap this PR’s three `internal/session` files. Head `2360d63b33b954622a539060b07d69efa3c8d919` is still mergeable, and all six required checks pass; the ruleset does not require strict base freshness. GitHub still reports `BLOCKED` to this read-only account. Could you merge this head, or identify the remaining maintainer-visible rule?

quad341 (approved):
@tudorsaitoc — thank you for this, and especially for the "Why recovery belongs here" section. You found the exact window that made this unrecoverable: the command gets built while `session_key` exists, the key clears, and the old `session_key != ""` gate turned the one case that *needed* recovery into the one case that couldn't get it. That's a subtle read of a nasty loop.

The part I appreciate most is the decline branch. It would have been easy to just delete the gate and always relaunch; you noticed that a command with no resume shape to strip would replay the same failure at the same cost, and you wrote the test that pins it. Covering `StartRuntimeOnly` separately because the reconciler lands there — rather than assuming the `Send` test generalizes — is the right instinct, and the command-shape assertions are what make the fix provable.

One thing I'll pick up on my side: the new `clearStaleResumeMetadata` call was added alongside the existing one rather than replacing it, so the clear still runs before the decline returns. I'll drop the duplicate, add assertions that a decline leaves metadata untouched, and re-run the session suite and vet. Your design and credit stay exactly as they are.

Also — sorry about the push-permission runaround from #5160. I'll land this one.


---

Issue #2725, opened 2026-05-29
doctor: order-firing-current flags SUSPENDED rigs' orders as CRITICAL stale (persistent false RED)


Wldc4rd:
## Summary
`gc doctor` reports `✗ order-firing-current — scheduled orders are stale` for orders belonging to a **suspended rig**. When a rig is suspended (`suspended = true` in city.toml), its scheduled orders correctly stop firing — but the order-firing-current check keeps monitoring them and flags their now-old last-fire as CRITICAL, producing a persistent false RED that masks new failures.

## Repro
1. A rig with cron/cooldown orders runs for a while (orders fire normally).
2. Suspend the rig (`suspended = true`, reload).
3. After ~3× the orders' intervals, `gc doctor` → `✗ order-firing-current`, each detail line `<order>:rig:<suspended-rig>: last fired N ago, expected every M (CRITICAL: stale)`.

In our city, citadel (suspended ~6.4 days ago) produces 8 CRITICAL-stale lines; **all active-rig orders are healthy**. Notably jetsons — also suspended — does NOT trip it: its orders never fired, so they hit the `lastFired.IsZero() && controllerStarted.IsZero() → StatusOK` branch. Only a rig that *ran then was suspended* trips it.

## Root cause (source)
`internal/doctor/checks_order_firing.go`: `scanOrderFiringCurrentOrders` → `orderdiscovery.ScanAll` → `orders.FilterEnabled` has no rig-suspension filter, so a suspended rig's orders stay monitored. `classifyOrderFiring` then flags them via the `age >= expected*3` CRITICAL path (~line 388) because their last real fire predates suspension.

## Ask
Exclude suspended rigs' orders from order-firing-current (treat a suspended rig's orders as not-monitored / disabled). A suspended rig is intentionally idle, so its orders shouldn't count against firing-currency. A real firing regression in an *active* rig still surfaces; the suspended-rig noise (and the masking of new failures behind a permanent RED) goes away.

## Confirmation
Empirically verified the fix's effect without a rebuild: every CRITICAL-stale order in `gc doctor --verbose` is the suspended rig's; no active-rig order is stale. So excluding suspended rigs yields a clean green check.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5270, opened 2026-08-14
fix(doctor): order-firing-current reads live suspension state, not just the deprecated rig.Suspended field

### LINKED ISSUE #5268: bug: doctor order-firing-current's suspended-rig skip reads the deprecated rig.Suspended field, so #2725's false RED is back
### Before you continue

This is a regression report against #2725, which was closed by the merged PR #2785 on 2026-06-01. Filing separately rather than commenting on the closed issue because the root cause is distinct from the one #2785 addressed.

### Gas City version

`HEAD-1534d46` — source build of `1534d46e248b51fd03392c991020d08f88270da0` (2026-08-14).

### Environment

Linux, systemd `--user` supervisor, tmux provider, `claude` provider. bd `1.1.0 (dev: main@bf97b73749ac)`. One city, three rigs, all three suspended. Store schema v59.

### Reproduction

1. In a city with at least one rig, suspend it either documented way:
   - `gc rig suspend <rig>` (writes `.gc/runtime/suspension-state.json`), or
   - `suspended_on_start = true` in the rig's `city.toml` block.

   Do **not** use the legacy `suspended = true` key.
2. Wait past the shortest rig-scoped order interval (5m for `auto-dispatch-ready`).
3. Run `gc doctor`.

### Expected behavior

`order-firing-current` skips orders belonging to suspended rigs. The dispatcher intentionally does not run them, so their staleness is not a fault — this is the behavior #2725 asked for and #2785 implemented ("skips `order-firing-current` for suspended cities and suspended rig orders").

### Actual behavior

`order-firing-current` fails as **blocking**, listing every rig-scoped order of every suspended rig as `CRITICAL: stale`, and `gc doctor` exits 1. Persistent false RED — the original symptom of #2725.

In our city, 30 of the 31 overdue entries are suspended-rig orders. The remaining one is an event-triggered order that is legitimately due.

```shell
$ gc doctor --json | jq '.results[] | select(.name=="order-firing-current")'
{
  "name": "order-firing-current",
  "status": "error",
  "severity": "blocking",
  "message": "scheduled orders are stale",
  "fix_hint": "Inspect with: gc order check && gc order history auto-dispatch-ready --rig ab-opus",
  "details": [
    ...
    "auto-dispatch-ready:rig:ab-opus: last fired 690m ago, expected every 5m (CRITICAL: stale)",
    "convoy-check:rig:ab-opus: last fired 689m ago, expected every 15m (CRITICAL: stale)",
    "witness-patrol-pour:rig:ab-opus: last fired 689m ago, expected every 5m (CRITICAL: stale)",
    "auto-dispatch-ready:rig:ab-sonnet: last fired 801m ago, expected every 5m (CRITICAL: stale)",
    "witness-patrol-pour:rig:ab-sonnet: last fired 800m ago, expected every 5m (CRITICAL: stale)",
    "auto-dispatch-ready:rig:tscdd-azure: last fired 5138m ago, expected every 5m (CRITICAL: stale)",
    "witness-patrol-pour:rig:tscdd-azure: last fired 5136m ago, expected every 5m (CRITICAL: stale)"
  ]
}

# all three rigs are suspended, via runtime state:
$ cat .gc/runtime/suspension-state.json
{
  "city": {},
  "rigs": {
    "ab-opus":     { "suspended": true },
    "ab-plain":    { "suspended": true },
    "ab-sonnet":   { "suspended": true },
    "tscdd-azure": { "suspended": true }
  },
  "updated_at": "2026-08-13T23:24:59.792327782Z"
}

# and none of them carries the legacy `suspended = true` key:
$ grep -n 'suspend' city.toml
1162:suspended_on_start = true     # ab-opus
1207:suspended_on_start = true     # ab-sonnet
# (tscdd-azure has no suspension key at all — it is suspended at runtime only)
```

### Additional context

**Root cause.** The skip added by #2785 resolves suspended rigs from the *deprecated* TOML field.

`internal/doctor/checks_order_firing.go:366`:

```go
func orderFiringCurrentSuspendedRigs(cfg *config.City) map[string]bool {
	out := make(map[string]bool)
	if cfg == nil {
		return out
	}
	for _, rig := range cfg.Rigs {
		if rig.Suspended && strings.TrimSpace(rig.Name) != "" {
			out[rig.Name] = true
		}
	}
	return out
}
```

But `internal/config/config.go:596` documents that field as deprecated and no longer written:

```go
// Suspended is the deprecated pre-runtime-state suspension flag.
// Parsed for backwards compatibility and treated as an alias for
// SuspendedOnStart by [Rig.EffectiveSuspendedOnStart], so existing
// cities with `suspended = true` continue to start their rigs
// suspended after upgrade. Live suspend/resume commands no longer
// write this field. ...
Suspended bool `toml:"suspended,omitempty"`
```

and `config.go:1184` makes the read site rule explicit:

```go
// ... Use this everywhere a read site needs
// the authored default — never read r.Suspended directly for
// behavior; only `gc doctor` consults it (to warn about the legacy
// field).
func (r *Rig) EffectiveSuspendedOnStart() bool {
	return r.Suspended || r.SuspendedOnStart
}
```

while `SuspendedOnStart` names the runtime file as authoritative:

```go
// SuspendedOnStart is the rig's desired suspension state at city
// start. When true and no explicit entry exists for this rig in
// .gc/runtime/suspension-state.json, the rig is treated as
// suspended. Once the user has explicitly suspended or resumed the
// rig via `gc rig suspend/resume`, the runtime state wins.
SuspendedOnStart bool `toml:"suspended_on_start,omitempty"`
```

So the returned map is empty for both current spellings:

| how the rig is suspended | sets | seen by the check |
|---|---|---|
| `gc rig suspend` | `.gc/runtime/suspension-state.json` | no — never read |
| `suspended_on_start = true` | `Rig.SuspendedOnStart` | no — wrong field |
| `suspended = true` (legacy) | `Rig.Suspended` | yes |

The suspended-rig skip is therefore effectively dead code except on cities still carrying the legacy key — which, per the deprecation note, is the set that is expected to shrink over time. That reading matches what we observe: three suspended rigs, none skipped.

**Scope.** `orderFiringCurrentSuspendedRigs` has two callers — the main check path (`checks_order_firing.go:180`, and `pendingLastRunOrders` at :664) and the wildcard-override prune (`orderFiringCurrentPruneSuspendedOnlyWildcardOverrides` at :320) — so a fix in the one helper covers both.

**Suggested fix.** Resolve through `Rig.EffectiveSuspendedOnStart()` *and* overlay `.gc/runtime/suspension-state.json`, honoring the precedence already documented on `SuspendedOnStart` (runtime state wins where an explicit entry exists). Using `EffectiveSuspendedOnStart()` alone would fix the `suspended_on_start` case but still miss `gc rig suspend`, which is the more common path and the one that leaves no trace in `city.toml`.

**Impact.** `order-firing-current` is `severity: blocking`, so `gc doctor` exits non-zero for as long as any rig stays suspended. That makes doctor's exit code unusable as a health gate on any city that parks a rig — and trains operators to read a red doctor as normal, which is how a real stale-order fault would get missed.

Refs #2725, #2785.

rjgeng:
Fixes #5268.

`gc doctor`'s `order-firing-current` check was supposed to skip staleness
checks for orders belonging to suspended rigs (per #2725/#2785), but its
suspended-rig resolver only read the deprecated `Rig.Suspended` TOML
field. Neither current suspend path sets that field:

- `suspended_on_start = true` in `city.toml` (the documented replacement)
- `gc rig suspend` (writes `.gc/runtime/suspension-state.json`, leaves no
  trace in `city.toml` at all — the more common path)

Either path left every one of that rig's scheduled orders reported as
`CRITICAL: stale` and `blocking`, exactly the false-RED symptom #2725
originally asked to fix — `gc doctor` exits non-zero for the entire
lifetime of any parked rig.

## Fix

`orderFiringCurrentSuspendedRigs` now merges each rig's authored
`suspended_on_start` default (via `Rig.EffectiveSuspendedOnStart()`, so
the deprecated field keeps working as its documented alias) with the
runtime override in `.gc/runtime/suspension-state.json`, via
`suspensionstate.Load` + `suspensionstate.EffectiveRigSuspended` — the
same merge already canonical everywhere in `cmd/gc`
(`buildEffectiveSuspendedRigNames` and friends). `internal/doctor` can't
import `cmd/gc`'s wrapper (wrong dependency direction), so this calls
`internal/suspensionstate` directly, the same way `internal/doctor`
already depends on peer `internal/*` packages (`contract`, `config`).

Threaded `cityPath` through all three real call sites — `run()`,
`orderFiringCurrentScanConfig`, and
`orderFiringCurrentPruneSuspendedOnlyWildcardOverrides`. The original
issue only named two of these three callers; this fixes the third too
since it shares the identical bug.

## Tests

- RED confirmed first: `TestOrderFiringCurrent_SkipsSuspendedOnStartRigOrders`
  and `TestOrderFiringCurrent_SkipsRuntimeStateSuspendedRigOrders` both
  failed against the unfixed code.
- GREEN after the fix, plus a new test,
  `TestOrderFiringCurrent_RuntimeStateResumeOverridesSuspendedOnStart`,
  covering the reverse direction (`gc rig resume` re-enabling staleness
  checks on a rig whose `city.toml` still says `suspended_on_start = true`).
- All 4 pre-existing suspended-rig tests (using the legacy `Suspended`
  field) still pass unchanged, confirming the backward-compat alias holds.

`go build ./...`, `go vet ./...`, and `go test ./internal/doctor/...` all
clean.


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277172813

The1nk:
Heads-up from the reporter of #5268 — this fixes the right thing, and there is a second defect underneath it that this PR will inherit rather than introduce.

The skip is a `map[string]bool` (`orderFiringCurrentSuspendedRigs`, checks_order_firing.go:375, applied at :200). That expresses *whether* a rig is suspended but not *when* it was suspended, so a rig parked after an order has already stalled retroactively excuses the stall. On our city that cost 8h22m of genuine stall on a live rig — the heartbeat stopped 2026-08-20T19:54:02Z, the rig was suspended at 04:16:24Z, and `gc doctor` went green. It matters more once something auto-parks idle rigs, which is our case.

Filed separately as #5575 with the measurement and three candidate fixes, since it is a different root cause and outlives this PR. Nothing here needs to change for that — flagging it so it does not read as fixed once this merges.

Thanks for picking #5268 up.

rjgeng:
Rebased onto current `main` and fixed one call site while doing so. #5034 added a fourth caller of `orderFiringCurrentSuspendedRigs` after this PR's checks last ran, in a different file — so the merge stayed `MERGEABLE` while the build would have failed. Passing `cityPath` through is also what this PR intends: the outcome check now reads live suspension state, same as the firing check.

quad341 (approved):
@rjgeng — the diagnosis here is the valuable part. Noticing that `Rig.Suspended` is written by *neither* live suspend path, so #2725's fix had been quietly dead for every rig parked the current way, is the kind of thing that hides for months behind a check that looks like it works. Threading `cityPath` through the third caller the issue didn't name — and then catching #5034's fourth caller in `checks_order_outcome.go` during the rebase — is exactly the care I want on a shared helper.

Reusing `suspensionstate.EffectiveRigSuspended` rather than hand-rolling the tri-state merge means the doctor can't drift from what `gc rig suspend` actually means. And `TestOrderFiringCurrent_RuntimeStateResumeOverridesSuspendedOnStart` tests the resume direction, which is the one most people skip.

Merging as-is. I ran build, vet and the full `internal/doctor` suite against your head commit locally, since the CI shard covering this package shows SKIPPED.

One note, not yours: @The1nk's #5575 (suspension has no timestamp, so parking a rig retroactively excuses an existing stall) is inherited and tracked separately — nothing needed here. Thanks for the quick, well-tested fix.


---

PR #5271, opened 2026-08-14
fix(formula): teardown tails are not finalize sinks — ends the adopt-pr settlement deadlock

julianknutsen:
Fixes the settlement deadlock that made every adopt-pr molecule require one hand-close to finish (bead ga-99u0u): the teardown step's retry control — minted by `expandRetry` because teardown carries `[steps.retry]` — became a `workflow-finalize` sink via `graphSinkStepIDs`' unreferenced-step rule, producing the triangle: finalize blocks on the control → the control waits on its attempt → the attempt cannot pass until `ROOT_OUTCOME` → which only finalize produces. Worse latent case, reproduced end-to-end in this PR's red tail: if the control ever got served, three transient `root_not_closed_yet` attempts exhaust to `fail` and **poison a fully-green run to FAIL** via the scope re-grade.

Two patches, each proven independently necessary (isolation runs):
1. `graphSinkStepIDs` skips `scope_role=teardown` steps — finalize no longer needs the teardown tail. Formula-compile verification on the real `mol-adopt-pr-v2.toml`: exactly one dep edge removed (107→106), step set byte-identical; the other five teardown-bearing pack formulas compile with zero dep changes (their teardown was always explicitly referenced).
2. The terminal subtree sweep exempts the teardown tail (`CloseSubtreeWithMetadataExcept` + `dispatch.teardownTailExclusion`) — without it, settlement instantly skip-closes the real cleanup and leaks the worktree every time (patch-1-only isolation red).

E2E: root settles pass → cleanup attempt passes its guard → the retry control serves itself pass → zero open members, no hand-close. Docs: the sink exclusion and sweep exemption added to `formula-spec-v2.md` (user-visible contract).

Deploy note: pre-fix molecules keep the bad dep edge — the one-time sweep (teardown-scope retry/ralph controls with closed bodies → close skipped) is folded into the staged relic-sweep runbook.

### DIFF
--- a/docs/reference/specs/formula-spec-v2.md
+++ b/docs/reference/specs/formula-spec-v2.md
@@ -528,7 +528,9 @@ The v2 compiler must emit a flat, topologically ordered graph:
 - **`workflow-finalize` is appended.** A control step with ID
   `workflow-finalize` (kind `workflow-finalize`) is added depending on
   every sink step, so it becomes Ready exactly when all other work is
-  terminal.
+  terminal. Steps carrying `gc.scope_role = "teardown"` are excluded from
+  the sink set: teardown runs after the workflow settles (section 3.5), so
+  gating settlement on it would deadlock the run.
 - **The root blocks on the finalize step.** The workflow root bead is made
   to depend on `workflow-finalize` (or, when a recipe has no finalize step,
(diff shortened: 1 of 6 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277168603

PR #5272, opened 2026-08-14
fix(beads): retry serialization conflicts on NativeDoltStore Close and Reopen

sjarmak:
`CloseAll` calls the retried `SetMetadataBatch` and then an unretried `Close` for each bead. A Dolt serialization conflict on that `Close` left metadata written to a bead that never closed. Hardening only the first call had made the asymmetry worse: the half that survived a conflict was the half that did not matter. Users saw this as an HTTP 500 on session suspend, and CI saw it as a flake in `TestHumaBinary_SessionMessageAsync`.

This wraps `Close` and `Reopen` in the store's existing serialization-conflict retry, the same one `Create` and `Tx` already use.

### Why the retry is needed at this layer

The default backend has none. beadslib's embedded Dolt store, which `OpenBestAvailable` selects unless a config asks for server mode, runs `CloseIssue` as a bare `withConn` transaction, so a conflict surfaces straight to the caller. The server-mode store retries internally via `withRetryTx` and reaches this wrapper only after exhausting its own budget.

### Why replaying is safe

A serialization conflict guarantees the write did not land, and every `CloseIssue` path commits exactly once. There are three such paths and they do not share a mechanism, so the property holds per branch rather than by a single appeal:

- the permanent bead goes through `withRetryTx`/`withWriteTx`;
- an active wisp branches to `closeWisp`, a bare `BeginTx`/`Commit` that deliberately carries no `withRetryTx`, which makes it the branch that most wants an outer retry;
- the embedded backend runs one `withConn` transaction.

For contrast against a real two-commit caller rather than a guessed one: `RunInTransaction` commits the regular transaction and the ignored transaction separately and can fail with the first already committed, which is why its callers cannot simply replay. `SetMetadataBatch` is not one of them despite the name; it calls `UpdateIssue` directly. `Close` has no two-commit window on any branch.

Two details are load-bearing rather than incidental:

- Extracting `closeOnce`/`reopenOnce` puts the re-read inside the retried unit, so each attempt decides from fresh state. That matters against other writers, not against this store's own replay: the retry sleeps between attempts and a concurrent actor can close the bead in that gap.
- The reopen reason stays empty. beadslib's `ReopenIssue` performs a separate `AddComment` transaction when the reason is non-empty, which would split this into two independent writes and create the window the rest of the change relies on not existing.

### Tests

`internal/beads/native_dolt_store_close_retry_test.go` asserts the **mutation count**, not just the returned error. A replay that re-issued the write would still return `nil` while performing exactly the write this change exists to prevent. The motivating failure is guarded at the `CloseAll` level so the asymmetry itself is covered and not only its halves. Reverting either wrap turns the tests red at both levels.

### What this does not do, and why #5228 stays open

The retry classifier is untouched. `isNativeDoltSerializationConflict` matches error 1213, SQLSTATE 40001, and Dolt's conflict wording, but not a lock-wait timeout reported as `Error 1205 (HY000)`, so a 1205 still surfaces after one attempt here, exactly as it already does for the pre-existing `Create` and `Tx` wrappers. Widening the classifier changes retry behavior on call sites this change does not own and needs its own replay-safety argument per site.

The remaining unretried writes on this store also need per-site arguments, and they differ materially: `ReleaseIfCurrent` is a compare-and-swap, `Tx` composes caller-supplied operations whose replay safety belongs to the caller, and `Create` with `compensateFailedCreate` is a rollback pair where retrying one half alone could compensate a create that later succeeds.

Both items stay tracked in #5228, so this PR refs the issue rather than closing it.

### Test plan

- `go test ./internal/beads/ -run 'CloseRetry|Close|Reopen'`
- `go build ./...` and `go vet ./...`
- Full fast unit sweep green at this head.

### Review scope

Point fix inside the existing `NativeDoltStore` boundary: two files, no new package, no schema or public contract change, no new orchestration shape. Fixes a user-visible 500 and a CI flake.

Refs #5228

### DIFF
--- a/internal/beads/native_dolt_store.go
+++ b/internal/beads/native_dolt_store.go
@@ -1118,8 +1118,47 @@ func (s *NativeDoltStore) Close(id string) error {
 		return err
 	}
 	defer release()
-	ctx, cancel := nativeDoltOperationContext(context.TODO())
-	defer cancel()
+
+	return retryOnNativeDoltSerializationConflict(func() error {
+		ctx, cancel := nativeDoltOperationContext(context.TODO())
+		defer cancel()
+		return s.closeOnce(ctx, storage, id)
+	})
+}
+
+// closeOnce performs one complete close attempt, read included. A retry calls
+// this whole operation rather than just the write, so every attempt decides
+// from freshly read state.
+//
+// Replaying the write is safe because a serialization conflict guarantees the
+// write did not land, and because every CloseIssue path commits exactly once.
+// There are three such paths and they do not share a mechanism, so the property
+// is stated per branch rather than asserted for "both backends":
+//
+//   - dolt.DoltStore, permanent bead: withRetryTx/withWriteTx (issues.go
+//     CloseIssue).
+//   - dolt.DoltStore, active wisp: CloseIssue branches to closeWisp, which uses
+//     a bare BeginTx/Commit with a deferred Rollback and deliberately no
+//     withRetryTx. Its own comment says not to add one, which is precisely why
+//     the retry belongs out here: that path has no internal retry at all.
+//   - embeddeddolt.EmbeddedDoltStore: one withConn transaction.
+//
+// Single-commit is the whole argument, so contrast it with a real two-commit
+// caller rather than a guessed one: beadslib's RunInTransaction commits the
+// regular tx and the ignored tx separately and can fail after the first landed,
+// which is why its callers cannot simply replay. SetMetadataBatch is NOT such a
+// caller despite the shape of its name; it calls storage.UpdateIssue directly,
+// which branches the same three ways Close does: permanent Dolt under
+// withRetryTx, active wisps through updateWisp's bare BeginTx/Commit, and
+// embedded under withConn. Close has no two-commit window on any branch.
+//
+// The re-read is what makes an attempt correct in the presence of OTHER
+// writers, which is a live case rather than a hypothetical: the retry sleeps
+// between attempts, so a concurrent actor can close the bead in that gap. The
+// short-circuit returns nil instead of issuing a redundant CloseIssue, and
+// recomputing the reason per attempt keeps it consistent with the state the
+// attempt actually observed.
+func (s *NativeDoltStore) closeOnce(ctx context.Context, storage beadslib.Storage, id string) error {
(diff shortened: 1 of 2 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277168813

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @sjarmak! This PR makes `NativeDoltStore.Close` and `NativeDoltStore.Reopen` retry the full read/write attempt when Dolt raises a transient serialization conflict, instead of surfacing that contention as a hard failure. That matters because those conflicts previously turned routine store close/reopen into spurious hard errors under concurrent Dolt access — retrying the unit makes the beads store resilient to that contention, and the added `CloseAll` regression test locks the recovery behavior in.

### Original PR Review
Decision: approve.

Specific gaps validated:
- Unretried `Close`/`Reopen` no longer surface Dolt serialization conflicts as hard failures — both now retry the full read/write attempt via `retryOnNativeDoltSerializationConflict` (`internal/beads/native_dolt_store.go:1122` and `:1187`).
- The retried unit is replay-safe — the retry body re-reads state inside each attempt, and every touched storage path either commits once or preserves the existing non-retry indeterminate-commit contract.
- `CloseAll` regression coverage — the new test drives `CloseAll` and confirms the close retry recovers after close metadata has already landed, without replaying the metadata write.
- The empty reopen reason is preserved — `reopenOnce` still passes `""` (`internal/beads/native_dolt_store.go:1219`), avoiding the upstream two-commit reopen-comment window.

Review findings addressed:
- None. The multi-model review (Claude + Codex) reported no blocker, major, minor, or nit findings across correctness, contract fidelity, concurrency and state safety, error handling, test evidence, security, resource lifecycle, and architectural consistency.

Remaining non-gating notes:
- No remaining gating issues were found. The change is a self-contained two-file fix with focused retry and `CloseAll` regression tests.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `ee4b18c2` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/31803828189

### Review Iterations
1 review pass performed; no fix iterations were needed.

---

PR #4322, opened 2026-07-16
feat: enforce priority-band work admission

sjarmak:
## Summary

- define one canonical work-priority order: P0 before P1 before P2, nil priority as P2, then creation time and bead ID within a band
- apply that order to bead readiness, federated city/rig hook claims, continuation assignment, and shared pool-cap admission
- preserve non-preemptive recovery by keeping live resume/wake requests ahead of in-flight creates, and both ahead of fresh demand
- keep lost-claim retries and continuation assignment inside the active priority band
- update default work queries and contributor architecture documentation to use priority-band FIFO ordering

## Contract and review routing

This changes an existing core-platform contract: work admission and assignment ordering across the bead store, hook claimant, and pool reconciler.

**Holding for explicit maintainer review. Do not auto-merge.**

External review was performed with Claude Fable. Its first pass found a cap-ordering regression where an in-flight create could beat a live resume. That case was reproduced RED-first, fixed, and the Fable follow-up returned **APPROVE**.

The generated `bd ready --sort hybrid` queries are compatible with the repository's pinned `bd v1.1.0` and minimum-supported `v1.0.4`; both tagged versions declare `hybrid` valid, and the installed v1.1.0 binary was exercised directly.

No upstream GitHub issue is linked. The work originated from operational queue-ordering failures and was tracked internally as `gc-1tww3`.

## Testing

- [x] `make build`
- [ ] `make check` — formatting, full lint, vet, routed-row guard, changed packages, and `cmd/gc` passed; the aggregate baseline is red because `origin/main` has an unrelated unassigned `TestCapstoneIntegrationRealMinter` entry in the integration-shard manifest
- [x] `make check-docs`
- [x] `make test-acceptance`
- [x] pre-push `make test-fast-parallel` with `LOCAL_TEST_JOBS=3`
- [x] focused priority/work-query/pool/claim/continuation/lost-claim tests
- [x] race-focused beads and `cmd/gc` tests
- [x] pre-commit hook: changed-scope lint, generated reference docs, vet, and docsync
- [ ] `make test-integration` — CI's sharded integration tier remains the authoritative run

## Checklist

- [x] Linked an issue, or explained why one is not needed
- [x] Added or updated tests for behavior changes
- [x] Updated docs for user-facing changes
- [x] Called out breaking changes or migration notes: no wire, schema, or configuration migration; this intentionally changes queue admission order

PR #5275, opened 2026-08-14
fix(dispatch): enforce priority-band FIFO admission

sjarmak:
## Summary

- define one fixed P0-to-P4 comparator, treating missing priority as P2 and breaking ties by `created_at` then ID
- apply priority-band FIFO consistently across ready output, hook and cross-store claims, pool resume, nested capacity, and scarce create budgets
- preserve recovery-first behavior, floor reservations, elastic reserve, same-band seed fairness, and turn-window refusal before every band boundary

## Supersedes

Supersedes #4322. This is the current-main port of that work: it retains and credits the fixed priority-band policy and its tests while dropping the stale configurable-policy and control-flow surface that no longer fits current main.

## Published repair head

Exact head: `7e08b4d097ac84557e528cc280fdcb9e97b07299`.

The repaired head merges current main without rewriting the published history and fixes the review-discovered mid-loop window-expiry reporting regression. Cross-family review `gc-804924` verified the merge criteria and the focused composed regression.

## Test plan

- `go test ./cmd/gc -count=1`
- `go test ./internal/beads ./internal/config ./internal/poolplan -count=1`
- `go vet ./cmd/gc ./internal/beads ./internal/config ./internal/poolplan`
- regenerated work-query goldens and `git diff --check`
- repository pre-push `make test-fast-parallel` passed at the published repair head

Publication only; no merge is requested by this action.


sjarmak:
Closing as superseded. Claim ordering on main now goes through `hookCandidateRank` in `cmd/gc/hook_cross_store.go`, together with #5629, which covers the priority-band ordering this PR added. 51 of its 71 files conflict with main. If a band-ordering gap turns up later, a fresh PR against the current ranking code will be smaller than rebasing this one.

Issue #3387, opened 2026-06-11
Supervisor REST `/v0/city/{cityName}/beads`: drops ephemerals, serves stale snapshot, no per-rig scope


edwarddavison:
> _Downstream triage from the **Sustainability Baseline** platform, which runs the Gas City substrate. Reproduced against **gc v1.2.1 (gascity @eedcee36)** + the deployed patcher set; source-line references are to that ref. Internal tracking IDs (co-…) are opaque cross-refs._

**Target:** gastownhall/gascity · **Severity:** MEDIUM · **Affects:** v1.2.1 (eedcee36) · **Class:** substantial-patch · **Upstream:** not fixed (origin/main +1 unrelated commit; default-`TierIssues` handler, single `/v0/city/{cityName}` route, and the `c.beads` snapshot read all unchanged).

> *(See the SECURITY NOTICE at the top — the read-only investigation for this item refused an injected destructive "verification stanza." The three gaps below are genuine source-level Go defects.)*

**Problem:** The supervisor's city-scoped bead surface has three independent defects. **(1) INCOMPLETE:** `humaHandleBeadList` builds its `beads.ListQuery` with no `TierMode`, defaulting to `TierIssues`; `ListQuery.Matches` then drops every `Ephemeral=true` bead, so all wisp-tier beads (co- workflow step beads, which are ephemeral) are silently excluded — REST returns ~26 vs the backing store's ~50. **(2) STALE:** for non-`Live` issues-tier queries `CachingStore.List` answers from the in-memory `c.beads` snapshot, refreshed only on the adaptive reconcile cadence (30s small / 60s ≥1000 / 120s ≥5000), so a recently status-churned bead is served with its pre-transition status and `updated=null`. **(3) NO PER-RIG:** every bead route is registered under the single constant `cityScopePrefix = "/v0/city/{cityName}"`, and the supervisor registers one city (name from `EffectiveCityName`/GC_CITY = `control`); `resolveCityServer` returns nil for any other name, so `/v0/city/sb/beads` and `/v0/city/app/beads` 404. Rig is only a `?rig=` query param on the control endpoint, not a path scope.

**Expected vs actual:** Expected — a city/rig bead read returns the same active set `bd` holds (issues + wisps), reflects current status within the request, and supports rig-scoped addressing. Actual — wisp/step beads missing, status lags up to a full reconcile interval (with `updated=null`), and only the GC_CITY-derived `control` city resolves.

[...]

PR #5276, opened 2026-08-14
fix(beads): preserve blocked work across partial cache reads

sjarmak:
## Summary

- make partial cache reads fall back to the backing store for broad list/count queries instead of hiding non-ready work
- fail cached readiness closed when dependency state is absent, while preserving event-driven cache invalidation
- expose all non-closed statuses through the default CLI/API bead-list surface, including blocked work
- preserve live open-session lookup by stating the supported open-status predicate during partial prime

## Published repair head

Exact head: `6f47e74dc5e0eef6ad72db4ee5b34dcbb09e0f10`.

The repaired head fixes the CI-discovered live session-resolution regression. Cross-family review `gc-804935` reproduced the pre-fix failure and verified the exact fix across 12 affected packages, 243 cache tests, both residency ratchets, build, and vet.

## Test plan

- `go test ./cmd/gc -count=1`
- `go test ./internal/beads ./internal/api -count=1`
- `go test -race ./internal/beads -run '^(TestCachingStorePartialPrime|TestCachingStoreCachedReadyIgnoresStaleDependencyEventsAfterEventMutation$)' -count=10`
- `go vet ./cmd/gc ./internal/beads ./internal/api`
- both residency ratchets and `git diff --check`
- repository pre-push `make test-fast-parallel` passed at the published repair head after one clean retry; the first attempt was blocked by a temporary fake-executable `Text file busy` harness race, reproduced green 10/10 in the focused oracle before retry

Publication only; no merge is requested by this action.

---
- Related to #3387 — touches the same caching-store read path the stale-snapshot gap points at: broad nonclosed list/count queries now fall back to the backing store instead of being answered from a partial prime that only loaded open and in_progress, so the bead-list surface stops reporting blocked work as open or omitting it. It does not cover the ephemeral/wisp tier omission, the per-rig route scope, or reconcile-cadence staleness once the cache is fully primed.


sjarmak:
## Cross-provider review record

Run against the exact head `6f47e74dc5e0eef6ad72db4ee5b34dcbb09e0f10`. Two independent legs, both PASS, no blocking defect.

**Legs.** Claude Opus 5 (manual enumeration of the changed logic against the source) and Codex CLI (adversarial, clamped, five targeted questions). Both were run at the same unmoved head.

**What was verified.**

- The new `cacheServableForListQueryLocked` gate refuses a partial cache for any query outside the two primed statuses. An empty status filter, `closed`, and `blocked` all fall back to the store. Only exact `open` and `in_progress` queries are served from a partial prime, and `partialPrimeStatuses` is the same list `PrimeActive` iterates, so the gate cannot drift from what was actually loaded.
- The "one observed live blocker settles the verdict" early return in `cachedReadyDependencyStatusesKnown` is sound. An observed dependency in `open` or `in_progress` is nonclosed and therefore blocking, and dependencies missing from the partial snapshot cannot make an observed blocker closed. Neither leg could construct a case where it reports known-blocked for a bead that is actually ready.
- `return !missing` is correct for a bead with zero ready-blocking dependencies: knowledge is vacuously complete, and the caller proceeds to classify rather than forcing an unnecessary fallback.

**The one point that needed a second pass.** The first Codex run returned FAIL on the call site at `caching_store_handles.go:289`, which passes `nonclosedStatusesComplete=true` unconditionally. That finding was conditional on the path being reachable under a partial cache. It is not: `cachedReadyCompleteOnly` requires `c.state == cacheLive && c.depsComplete` at `:251` and returns `ErrCacheUnavailable` otherwise, with a further per-bead bail at `:271` when a ready projection is unknown. The guard was missing from the first prompt, not from the code. Re-run with both guards supplied, the leg returned PASS and answered the question directly from the guard.

**Non-blocking finding.** The `b.IsBlocked != nil` early return treats the backend-supplied blocked projection as authoritative. That is correct here, and it does mean the shortcut inherits the surrounding invalidation contract: if a dependency change could leave a cached non-nil `false` stale, readiness could be misclassified. That is a general cache-coherency concern, separate from the partial-prime omission this change fixes, and it is not introduced by this change.

**Merge-result evidence.** The branch is 307 commits behind main and reports no textual conflict, which does not by itself mean the merged tree compiles. Merged into current `origin/main` in a scratch worktree: merge clean at 10 files, +258/-18, then `go build ./...` and `go vet ./...` both clean on the result. All six required contexts report pass.

PR #5277, opened 2026-08-14
feat(cmd/gc): assigned-work spine through the residency resolver (S2)

julianknutsen:
S2 of the residency-resolver migration (epic ga-qdt5y, bead ga-dk345): the assigned-work spine consumes `Plan(AssignedWork)` — wake filter, session reachability gates, the orphan-release ownership index (the ga-j4ob9 fix: dead-session claims on binding-resident work finally have an automatic reopen lane), the retired-session unclaim/reassign family, the cross-store close gates, and #5265's drain-ack release, all through one seam.

- `storeref` gains `Walk` (executor for existence probes and mutating sweeps — consumers cannot disassemble plans) and `Topology.ClaimRefs()` (claim-ref vocabulary survives a refused city).
- **Release-path safety is the load-bearing evidence**: every visibility widening carries an independent counterweight, each pinned by a real on-disk mutation with a distinct failure signature (strand / premature-retirement / claim-loss ×2); the live-holder control was rebuilt to defeat all masking defences so only the widened index saves it — and both seam pins carry DO-NOT-FOLD headers because defence-in-depth makes end-to-end detection structurally impossible.
- Gates fail closed on degraded legs ("does this session hold work?" refuses partial answers); sweeps stay void-with-completeness per §3.3.
- Council round: the routes registry's lifecycle made ownership-safe (pointer-guarded unregister + registration moved under the controller lock — a superseded runtime can no longer erase the live runtime's registration and silently revive binding-blind releases; the lock-failure registration leak died with it). The ForConfig close gates — invisible to both guard halves because they ranged the rig map — were routed through the resolver in-slice rather than deferred.
- Retired: the whole `a:workAssignmentStores` pattern family (13 baseline rows / 15 hits, zero additions), `classBindingLegForSessionScan`, `classBindingAssignedWorkStoreRef`, and the fork-F3 shape generalizes un-shape-gated.
- New agreement property: `Plan(RoutedWork)` and the sweep render identically — "demand and sweep read the same stores" is asserted, not assumed.

Council: approve-with-changes → blocker + 7 fold-ins all landed (fix commit `bcb734d397`). Full `cmd/gc` 874s green on the final tree; protected suites green with the two comment-authorized row updates (I10 flip, I5 mechanical).

### DIFF
--- a/cmd/gc/assigned_work_class_binding_test.go
+++ b/cmd/gc/assigned_work_class_binding_test.go
@@ -13,6 +13,16 @@ import (
 
 // The ga-whzrt rows: a claim a rig-scoped worker HOLDS on the leading
 // class-binding arm must reach the wake machinery, and nothing else must.
+//
+// S2 PORT: the arm's ref is no longer the classBindingAssignedWorkStoreRef
+// constant but whatever assignedWorkClaimRefs resolves for the city, and the
+// split is expressed as SERVED ROUTES rather than as a [storage] section in
+// cfg. Those are the same change: the constant was the empty string because the
+// reconciler's leading arm happens to be the binding, and the cfg shape gate
+// answered "no split" for a city whose section was deleted after it had served
+// one. Both facts now come from the routes.
+//
+// The rows themselves are unchanged, which is the point of a pin.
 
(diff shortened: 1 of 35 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277168744

PR #5283, opened 2026-08-14
fix(dispatch): preserve prepared drain worktrees

caseymatt:
## Problem
Separated formula-v2 drains could claim implementation steps in the launcher checkout because the prepared source anchor worktree was not propagated through item materialization and session work-dir resolution.

## Fix
- propagate canonical and legacy work-dir metadata to every materialized item step
- resolve prepared drain source-anchor workdirs before generic assigned-bead metadata
- cover multiple parallel drain items and claim-time resolution

## Verification
- go test ./internal/dispatch ./cmd/gc -run TestProcessDrainPropagatesSourceMemberWorkDirToItemWorkflow -count=1
- go test ./internal/dispatch ./cmd/gc -run TestResolveTaskWorkDirPrefersPreparedDrainSourceAnchor -count=1
- make test-fast-parallel
- go vet ./...


quad341 (approved):
@caseymatt — the part that makes this a real fix rather than a patch is that you caught *why* one-sided stamping couldn't hold: the drain recipe is materialized before prepare-worktree creates the directory, so item steps can only ever inherit the launcher checkout at that moment. Stamping at materialization *and* resolving the source anchor at claim time is the right shape, and your comment on `resolveDrainSourceWorkDir` explains it well enough that nobody will re-litigate it. I also like that all three tests seed the *wrong* directory and demand the right one back — that's coverage, not a coverage number.

I'm taking this with two small narrowings on my side, no round trip needed. `resolveTaskBeadWorkDir` now reads legacy `work_dir` before `gc.work_dir`: `stampRunSessionIdentity` writes `gc.work_dir` alone from a non-pool session's observed cwd, so canonical-first could let an observability stamp outrank the worktree creator's path on unrelated beads. And I'm scoping the new pre-claim branch to the drain-source lookup so it doesn't outrank the snapshot resolver for every pool session. Your tests pass unchanged; I'm adding one for the divergent-keys case.

Cross-store drain roots still fall back to the launcher — that's pre-existing, and ours to chase. Nice work.


---

Issue #5221, opened 2026-08-12
bdflags manifest is derived from bd --help, which omits flags bd accepts, so gc lint reports working commands as unknown


sjarmak:
`internal/bdflags` derives its flag manifest from `bd <sub> --help`, and the
guard that keeps it current, `TestBdFlagManifestCurrent`, parses the same
`--help` text. bd accepts flags that never appear in that output, so neither
the manifest nor its freshness test can see them, and `gc lint` reports them
as `bd-unknown-flag`.

Verified by running the installed bd (v1.1.0) rather than reading its help:

| invocation | exit | reported by `gc lint` |
|---|---|---|
| `bd create --type=task --label=x "t"` | 0, label applied | yes, false positive |
| `bd mol wisp gc --age 24h --dry-run` | 0, "No abandoned wisps found" | yes, false positive |
| `bd close <id> --status=closed` | 1, unknown flag | yes, correct |
| `bd update <id> --label=x` | 1, unknown flag | yes, correct |
| `bd create --prefix hq- "t"` | 1, unknown flag | yes, correct |

Not prefix matching, which was my first theory:

```
$ bd create --type=task --lab=x "t"
Error: unknown flag: --lab
```

So `--label` on `create` and `--age` on `mol wisp` are registered on the
command and hidden from help, while `--label` on `update` genuinely does not
exist. Reading help cannot tell those two cases apart.

### Why it is worth fixing

On gastownhall/gascity-packs at `734a7be`, after #5220 removes the `gc bd`
scope-flag and shell-pipe false positives, `gastown` has nine remaining
findings. Five of them are these two undeclared flags. One is a genuine
broken command. The other three are documentation of what not to do.

So the undeclared-flag gap is the largest remaining source of noise in the
check, and noise here is not free: a pack author who trusts the output
rewrites a working prompt.

### Options

1. Add the verified-undeclared flags to the manifest with a comment naming
   how they were verified. Cheap, and it goes stale the same way.
2. Derive the manifest by running each flag against bd rather than by reading
   help. Slow, and it needs a safe no-op invocation per subcommand.
3. Ask upstream bd to declare these flags in help, whether as supported or as
   deprecated, and keep the manifest help-derived. Slowest, and the only one
   that fixes the root cause rather than compensating for it.

I have not picked one, because whether an undeclared flag should be accepted
by the linter at all is a maintainer call. If `--label` on `create` is a
deprecated alias awaiting removal, reporting it is arguably correct and the
[...]

PR #5284, opened 2026-08-14
Sync bd flag manifest with bd v1.1.0

quad341:
## What this changes

Gas City now recognizes the global and subcommand flags exposed across both supported `bd v1.1.0` builds. This keeps the `gc bd` mutation guard and `gc lint` from mistaking flag values for commands or bead IDs when operators use flags such as `--database`, `--mem-profile`, and `--no-color`.

The manifest retains both the released binary's `--profile` spelling and the fleet build's renamed `--cpu-profile` spelling. The two binaries report the same version but expose different flag surfaces, so compatibility requires a superset.

## Why it is built this way

The checked-in manifest is a parser/classifier shared by both consumers; it never emits flags to `bd`. Keeping the union of supported spellings is therefore safe, deterministic, and resilient to the release-versus-fleet build skew.

## Review notes

- Review the value-versus-boolean classification of newly recognized flags, especially `list --deps`, which has an optional value.
- Confirm the compatibility comment explains why both `--profile` and `--cpu-profile` remain present.
- This changes static manifest data and test expectations only; it adds no configuration, API, dependency, or migration.
- The supported `bd` version is unchanged.

## Test plan

- [x] Run `go build ./...` and `go vet ./...`.
- [x] Run `make test-fast-parallel`.
- [x] Run `make test-cmd-gc-process-parallel` so process-backed tutorial coverage executes with `GC_FAST_UNIT=0`.
- [x] Run the diff-owned completeness tests with zero skips.
- [x] Verify `TestBdFlagManifestCurrent` passes all 17 supported subcommands against both `bd v1.1.0` builds.
- [x] Release gate: [`release-gates/ga-o0pdkh-bdflags-manifest-gate.md`](release-gates/ga-o0pdkh-bdflags-manifest-gate.md)

---
- Related to #5221 — refreshes the same internal/bdflags manifest from newer `bd --help` output, which clears some bd-unknown-flag noise from gc lint; it does not touch the undeclared flags that never appear in help at all (`create --label`, `mol wisp --age`), and the manifest stays help-derived, so the root cause described in that issue is untouched

### DIFF
--- a/internal/bdflags/bdargs_test.go
+++ b/internal/bdflags/bdargs_test.go
@@ -45,15 +45,16 @@ func TestSplitGlobalFlagsSkipsGlobalFlagValues(t *testing.T) {
 // bypass below: SplitGlobalFlags would read that flag's value as the verb, and
 // every guard keyed off the verb stops firing — with no test failing.
 //
-// Sourced from `bd --help` (bd 1.1.0). bd declares exactly four persistent
-// flags that consume the next argument; -C and --directory are the two spellings
-// of one of them. Every other persistent flag (--global, --ignore-schema-skew,
-// --json, --profile, -q/--quiet, --readonly, --sandbox, -v/--verbose, -h/--help,
-// -V/--version) is boolean and consumes nothing.
+// Sourced from `bd --help` (bd 1.1.0, build 0954be416). bd declares six
+// persistent flags that consume the next argument: --actor, --database, --db,
+// --dolt-auto-commit, --mem-profile, and -C/--directory (the two spellings of
+// one). Every other persistent flag (--cpu-profile, --global,
+// --ignore-schema-skew, --json, --no-color, -q/--quiet, --readonly, --sandbox,
+// -v/--verbose, -h/--help, -V/--version) is boolean and consumes nothing.
 func TestGlobalValueFlagsIsComplete(t *testing.T) {
 	want := map[string]bool{
-		"--actor": true, "--db": true, "-C": true, "--directory": true,
-		"--dolt-auto-commit": true,
+		"--actor": true, "--database": true, "--db": true, "-C": true,
+		"--directory": true, "--dolt-auto-commit": true, "--mem-profile": true,
 	}
 	if got := GlobalValueFlags(); !reflect.DeepEqual(got, want) {
 		t.Errorf("GlobalValueFlags() = %v, want %v; re-check `bd --help` persistent flags", got, want)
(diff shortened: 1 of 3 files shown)


quad341:
CI was red on `Integration / packages-core-4-of-4`; pushed 570e38be5c to fix it.

**Root cause.** Two different bd builds both self-report `bd version 1.1.0`, and they expose different global flags:

| | build | `--profile` | `--cpu-profile` / `--mem-profile` / `--no-color` / `--database` |
|---|---|---|---|
| CI | released v1.1.0 tag (beads `8e4e59d39`) | present | absent |
| fleet | post-release main (beads `0954be416`) | absent | present |

beads `fb52ac6a5` ("refactor(cli): rename CPU profiling flag", #5126) renamed the persistent `--profile` flag to `--cpu-profile` *after* the v1.1.0 tag was cut, so the released binary CI installs still registers `--profile` as a persistent bool.

`TestBdFlagManifestCurrent` fails closed in exactly one direction — a flag the live bd has that the manifest lacks (`freshness_test.go:108`). The reverse case, manifest ahead of the installed bd, is logged as benign (`freshness_test.go:99`). Dropping `--profile` in 3289b567 put all 17 supported subcommands on the failing side, which is 100% of the CI failure. Every other manifest addition on this branch is on the benign side and stays as-is.

**The premise for removing it was the part that was wrong.** 3289b567 removed `--profile` on the grounds that bd rejects a bare `--profile` as "unknown flag". That is true of the bd this fleet runs, but the manifest is a superset allowlist across every supported bd build, not a snapshot of one binary. Restoring the entry is safe for the newer bd: all three consumers (`gc lint`'s `ScanUnknownFlags`, the `gc bd` guard's `SplitGlobalFlags` and `ValueFlags`/`BoolFlags`) only parse and classify. Nothing in gc emits a manifest flag to bd, so a superset entry cannot cause gc to pass a flag the installed bd would reject.

**Verification.** Built bd at beads `8e4e59d39` and put it first on `PATH` to reproduce CI locally:

- before 570e38be5c, against the released build: 17 subtests fail, each missing exactly `[--profile]` — byte-identical to CI
- after 570e38be5c: the integration-tagged `internal/bdflags` package passes against **both** bd builds
- `go build ./...`, `go vet`, `gofmt` clean; `cmd/gc` lint and bd-guard consumer tests pass; pre-push `make test-fast-parallel` 10/10

The manifest comment now records that the table is a superset across builds, so a future sync against one binary does not remove the entry again.

**Two follow-ups for the deployer, not addressed here:**

1. `release-gate/deploy-clearance` was issued manually at `f90f0a04ea` (no workflow produces it) and does not survive a new head SHA — the new head currently carries zero commit statuses. It needs reissuing at `570e38be5c` before merge.
2. In `release-gates/ga-o0pdkh-bdflags-manifest-gate.md`, the acceptance row reading "A direct `bd --profile` check returns `unknown flag`, confirming the removed manifest entry was not real coverage" is incorrect — that check only ever ran against the fleet bd. Criterion 3's evidence also cites the superseded source `3289b567`.

Worth noting the impact if this had merged as-is: CI installs the released bd, so `Integration / packages-core-4-of-4` would have gone red on main for every subsequent PR. Required checks here are only `Check` and the four `Analyze` jobs, so nothing would have blocked it.

Investigated under bead `ga-d5m4ac`.

quad341:
@quad341 — the debugging behind this is the valuable part. Tracking the CI failure down to two binaries that both self-report `bd version 1.1.0` but expose different persistent flags, and pinning it to beads `fb52ac6a5` landing after the tag was cut, is exactly the root cause someone would otherwise rediscover in three months.

I re-ran `bd --help` and each changed subcommand's help against the fleet build before signing off, since `TestBdFlagManifestCurrent` only checks flag names and would not catch an arity mistake. Every added entry matches — including `list --deps`, which really does print as `string[="scheduling"]`, so the bool classification and the comment explaining it are right. `Integration / packages-core-4-of-4` is green at head, which confirms the `--profile` retention against the released binary CI installs.

The "safe to keep and dangerous to drop — only ever add" line is the durable fix; without it the next sync-against-one-binary pass removes it again.

Merging as-is. One follow-up worth a bead when you have a slow afternoon: teach `parseHelpFlagNames` to capture the type token too, so arity drift fails CI instead of relying on someone checking by hand.


---

quad341:

PR #5287, opened 2026-08-15
fix(beads): commit the runtime config write to Dolt

### LINKED ISSUE #5286: External Dolt: GC never commits its beads config write, leaving every provisioned database permanently dirty and blocking future schema migrations
## Summary

Gas City writes `issue_prefix` and `types.custom` into the beads Dolt `config`
table with a raw `INSERT` and never commits the result. Because `config` is not
registered in `dolt_ignore`, every city/rig database Gas City provisions against
an external Dolt server is left with a **permanently dirty working set**.

This is a latent fleet-wide landmine: the next beads schema migration that
touches `config` will refuse to run on every affected database at once, and the
recovery beads documents for that refusal cannot be executed against an external
Dolt server.

## The write

`examples/bd/assets/scripts/gc-beads-bd.sh:598-610`

```sh
ensure_bd_runtime_config_value() {
    ...
    # bd v1.0.3 rejects `bd config set issue_prefix`; GC still needs raw
    # bd commands to see GC's config in the DB-backed config table.
    server_sql_retry "USE \`$db\`; INSERT INTO config (\`key\`, value) VALUES ('$key', '$value') ON DUPLICATE KEY UPDATE value = VALUES(value)" >/dev/null || die "..."
}
```

There is no `DOLT_ADD`/`DOLT_COMMIT` anywhere in the script. This is the only
uncommitted write in it, and the Go side never writes the Dolt `config` table at
all (it only writes `.beads/config.yaml`), so this line is the sole source.

Both `ensure_bd_runtime_issue_prefix` (`:556`) and
`ensure_bd_runtime_custom_types` (`:575`) funnel through it.

The raw SQL itself is still required — beads deliberately rejects
`bd config set issue_prefix` (`cmd/bd/config.go:991-999`, the key is owned by
`bd init`/`bd rename-prefix`). Only the commit is missing.

## Observed impact

Against a live external Dolt server hosting one database per city/rig,
**15 of 21 databases** had a dirty `config` table. The working-set diff is
exactly the Gas City write, uncommitted since provisioning:

```
sc-tf-org           :: issue_prefix, sctforg, added
beads_tf            :: issue_prefix, tf, added
master              :: issue_prefix, master, added
spec                :: issue_prefix, spec, added + types.custom, added
stacked_chips_v2_hq :: issue_prefix, ga, added + types.custom, added
... (15 total)
```

`dolt_log` confirms the ordering — `bd init` commits, and the Gas City write
lands afterwards and is never committed:

```
ah3ocqqovcqp2i3fqrs1p7108sea0sbv  beads  bd init                    2026-08-08 20:57:04.763
827lts6to9f3sr9tn3m3cdbif5sk6p2i  root   schema: apply migrations   2026-08-08 20:57:04.663
```

`dolt_ignore` on those databases contains only
`ignored_schema_migrations`, `local_metadata`, `repo_mtimes`, `wisp_%`, `wisps`
— `config` is **not** exempt.

## Why this matters

**1. It blocks beads schema migrations, with no in-band recovery.**

beads refuses to apply a migration that alters a table carrying pre-existing
uncommitted changes (`schema.DirtyTablesError`):

```
failed to initialize schema: schema migration: pending schema migrations alter
pre-existing dirty tables: issues; run 'bd dolt commit' to commit the working
set at the current schema, then re-run the migration (gastownhall/beads#4566)
```

This is not hypothetical for `config`: migration `0030_migrate_local_metadata_keys.up.sql:27`
already issues `DELETE FROM config`. The next migration touching `config` trips
the guard on every affected database simultaneously.

The recovery beads names — `bd dolt commit` — **cannot run in server mode**.
`cmd/bd/store_factory.go` returns on `cfg.ServerMode` before ever reaching the
`cfg.LenientOpen` branch that exists specifically to break this deadlock, and
`dolt.Config.LenientOpen` is documented "Ignored in server mode". beads #4566
is titled *"Embedded mode: ..."* and its fix was scoped to embedded only. So an
external-Dolt user has no supported way out; the only recovery is hand-committing
over a raw SQL connection.

That beads gap is a separate upstream bug, but it is what turns this Gas City
defect from "recoverable" into "stuck".

**2. It drifts the database hash.**

The script already documents this exact hazard class for a different table
(`gc-beads-bd.sh:1432-1436`):

> The probe table is registered in `dolt_ignore` so history flattening can never
> first-commit it: a non-ignored table that lives only in the working set is
> committed by the compaction flatten's `DOLT_COMMIT -Am`, which drifts the
> database hash and quarantines GC for that database (hq June 2026, daa 2026-08-04).

`config` has the same shape and was never given the same treatment.

**3. Real user data sits uncommitted.**

On one database the dirty `config` rows included five `bd remember`
`kv.memory.*` entries — substantive operator notes living only in the working
set, with no commit backing them.

## Expected behavior

Gas City commits its own initialization writes. After `gc rig add` /
`gc city init` against an external Dolt server, the database working set should
be clean.

## Proposed fix

Commit the `config` write at the single choke point, scoped to the `config`
table so it cannot sweep unrelated dirty tables into Gas City's commit (which
would be the hash-drift failure above rather than a fix for it). This mirrors
the existing precedent in `cmd/gc/dolt_wisp_query_index.go:90-101`.

Fail-open with a warning: the value is already written, so a commit failure
should leave today's (dirty but functional) state rather than break provisioning
on a read-only replica — but it must be reported, never swallowed.

## Notes

- The `issues`-dirty variant reported by users appears to be a *different*,
  already-fixed beads bug (gastownhall/beads#5012, a session-scoping race during
  fresh init, closed 2026-08-02). No live database exhibited it. The `config`
  dirtiness documented here is the Gas City-owned defect.
- Existing affected databases stay dirty after a code fix and need a one-time
  backfill commit.

Toady00:
Closes #5286.

## Problem

`ensure_bd_runtime_config_value` writes `issue_prefix` / `types.custom` into the
beads Dolt `config` table with a raw `INSERT` and never commits it. `config` is
not in `dolt_ignore`, so every database Gas City provisions against an external
Dolt server keeps that row in the working set forever.

On a live external server, **15 of 21 databases** carried the uncommitted write,
dating back to provisioning. `dolt_log` shows `bd init` committing and the Gas
City write landing afterwards, uncommitted:

```
ah3ocqqovcqp2i3fqrs1p7108sea0sbv  beads  bd init                   2026-08-08 20:57:04.763
827lts6to9f3sr9tn3m3cdbif5sk6p2i  root   schema: apply migrations  2026-08-08 20:57:04.663
```

```
sc-tf-org :: issue_prefix, sctforg, added     (WORKING)
spec      :: issue_prefix, spec, added + types.custom, added
```

This is a latent fleet-wide landmine, not cosmetic:

- beads refuses to apply a schema migration that alters a table with
  pre-existing uncommitted changes (`schema.DirtyTablesError`). Migration
  `0030_migrate_local_metadata_keys.up.sql:27` already issues
  `DELETE FROM config`, so the next migration touching `config` blocks every
  affected database at once. The recovery beads names, `bd dolt commit`, cannot
  run in server mode — `cmd/bd/store_factory.go` returns on `cfg.ServerMode`
  before reaching the `cfg.LenientOpen` branch that exists to break exactly this
  deadlock, and gastownhall/beads#4566 was scoped to embedded mode only. There is
  no in-band way out.
- A table living only in the working set gets swept into a later
  `DOLT_COMMIT -Am`, drifting the database hash — the same hazard the read-only
  probe table is registered in `dolt_ignore` to avoid
  (`gc-beads-bd.sh:1432-1436`).
- On one database the dirty rows included five `bd remember` `kv.memory.*`
  entries: real operator notes with no commit backing them.

## Fix

Commit the write at its single choke point (both `ensure_bd_runtime_issue_prefix`
and `ensure_bd_runtime_custom_types` funnel through it), following the existing
precedent in `cmd/gc/dolt_wisp_query_index.go:90-101`.

Two deliberate choices:

- **Staging is scoped to `config`.** A blanket `DOLT_ADD('.')` would sweep
  whatever else is dirty into Gas City's commit — that is the hash-drift failure
  above, not a fix for it.
- **Fail open, but always report.** The value is already written, so a commit
  failure leaves today's dirty-but-functional state rather than breaking
  provisioning on a read-only replica. It warns to stderr and is never swallowed.

The raw SQL itself is unchanged and still necessary: beads deliberately rejects
`bd config set issue_prefix` (`cmd/bd/config.go:991-999`).

## Tests

TDD — `cmd/gc/gc_beads_bd_config_commit_test.go` drives the shell function
against a stub `dolt`, using the existing `extractShellFunction` harness. The
commit helper is extracted *optionally* so the pre-fix run fails on the asserted
behavior rather than a missing-function harness error. Against unmodified code
the failure reproduces the production condition exactly:

```
config write was never committed, leaving the working set dirty; queries:
  USE `rigdb`; INSERT INTO config (`key`, value) VALUES ('issue_prefix', 'sctforg') ...
```

Five cases: the write is committed; the commit is scoped to `config` and never
stages everything; `types.custom` is covered; an idempotent re-run
("nothing to commit") is not a failure; a genuine commit failure is surfaced to
the operator without aborting provisioning.

## Verification

- `go vet ./cmd/gc/ ./examples/bd/...` clean
- `go test ./examples/bd/... ./internal/beads/... ./internal/rig/... ./internal/doctor/...` pass
- `cmd/gc` script-lint tests (`TestGcBeadsBdNoBdConfigSet`, `TestGcBeadsBdConfigSetLintCases`) pass
- `bash -n` clean; `golangci-lint` reports 0 issues via the pre-commit hook

`make test-fast-parallel` has 26 pre-existing failures on this machine, all
git-fixture tests hitting a local global `commit.gpgsign=true` leak (80 "No
secret key" errors in the log), plus one push-gate test that assumes a
non-linked worktree. All 26 pass under `GIT_CONFIG_GLOBAL=/dev/null`; none are
related to this change.

## Note for existing deployments

Databases already provisioned stay dirty after this fix and need a one-time
backfill:

```sql
CALL DOLT_ADD('config');
CALL DOLT_COMMIT('-m', 'gc: record beads runtime config');
```

### DIFF
--- a/cmd/gc/gc_beads_bd_config_commit_test.go
+++ b/cmd/gc/gc_beads_bd_config_commit_test.go
@@ -1,0 +1,236 @@
+package main
+
+import (
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+	"testing"
+)
+
+// These tests exercise ensure_bd_runtime_config_value from
+// examples/bd/assets/scripts/gc-beads-bd.sh against a stub `dolt` binary.
+//
+// Gas City writes issue_prefix and types.custom straight into the beads Dolt
+// `config` table because beads refuses `bd config set issue_prefix` (it is
+// owned by `bd init`/`bd rename-prefix`). That raw INSERT used to leave the
+// row sitting in the Dolt working set forever: `config` is not registered in
+// dolt_ignore, so every city/rig database Gas City provisioned stayed
+// permanently dirty.
+//
+// That matters for two reasons the script already documents elsewhere:
+//
+//   - A beads schema migration refuses to alter a table that has pre-existing
+//     uncommitted changes (beads DirtyTablesError). Migration 0030 already
+//     issues `DELETE FROM config`, so the next migration touching `config`
+//     bricks every affected database at once. Its documented recovery,
+//     `bd dolt commit`, cannot run against an external Dolt server
+//     (gastownhall/beads#4566 fixed that deadlock for embedded mode only),
+//     leaving no in-band way out.
+//   - A table that lives only in the working set gets swept into an unrelated
+//     `DOLT_COMMIT -Am` later, drifting the database hash and quarantining GC
+//     for that database -- the exact hazard the read-only probe table is
+//     registered in dolt_ignore to avoid (see the comment above
+//     managed_dolt_read_only_probe in the script).
+//
+// The contract under test: the config write is committed before the function
+// returns, the commit is scoped to `config` alone so it cannot sweep unrelated
+// dirty tables into Gas City's commit, and an already-clean working set is not
+// treated as a failure.
+
+// configCommitHarness builds a shell program exposing
+// ensure_bd_runtime_config_value and its transitive dependencies, then invokes
+// it for one config key/value. The invocation is baked in rather than passed as
+// argv so the shared runShHarness runner can execute it, keeping this file off
+// the untagged subprocess census.
+func configCommitHarness(t *testing.T, key, value string) string {
+	t.Helper()
+	root := repoRootForLint(t)
+	scriptPath := filepath.Join(root, "examples", "bd", "assets", "scripts", "gc-beads-bd.sh")
+	scriptBytes, err := os.ReadFile(scriptPath)
+	if err != nil {
+		t.Fatalf("read script: %v", err)
+	}
+	script := string(scriptBytes)
+
+	var harness strings.Builder
+	harness.WriteString("#!/usr/bin/env bash\nset -u\n")
+	harness.WriteString("DOLT_HOST=\"${GC_DOLT_HOST:-127.0.0.1}\"\n")
+	harness.WriteString("DOLT_USER=\"${GC_DOLT_USER:-root}\"\n")
+	harness.WriteString("DOLT_PASSWORD=\"${GC_DOLT_PASSWORD:-}\"\n")
+	harness.WriteString("DOLT_PORT=\"${GC_DOLT_PORT:-3307}\"\n")
+	for _, fn := range []string{
+		"die",
+		"is_remote",
+		"connect_host",
+		"sleep_ms",
+		"server_sql",
+		"is_retryable_error",
+		"server_sql_retry",
+		"valid_sql_name",
+		"valid_custom_types_value",
+		"validate_bd_runtime_config_value",
+		"ensure_bd_runtime_config_value",
+	} {
+		harness.WriteString(extractShellFunction(t, script, fn))
+		harness.WriteString("\n")
+	}
+	// The commit helper is extracted optionally so that when it is absent the
+	// tests fail on the behavior they assert (no commit was issued) rather
+	// than on a missing-function harness error.
+	harness.WriteString(optionalShellFunction(script, "commit_bd_runtime_config"))
+	harness.WriteString("\n")
+	harness.WriteString("ensure_bd_runtime_config_value " +
+		shellSingleQuote(configCommitTestDatabase) + " " +
+		shellSingleQuote(key) + " " + shellSingleQuote(value) + "\n")
+	return harness.String()
+}
+
+// optionalShellFunction returns the named shell function's source, or an empty
+// string when the script does not define it.
+func optionalShellFunction(script, name string) string {
+	pattern := regexp.MustCompile(`(?ms)^` + regexp.QuoteMeta(name) + `\(\)\s*\{.*?\n\}`)
+	loc := pattern.FindStringIndex(script)
+	if loc == nil {
+		return ""
+	}
+	return script[loc[0]:loc[1]]
+}
+
+// writeConfigCommitFakeDolt stubs `dolt ... sql -q <query>` by appending each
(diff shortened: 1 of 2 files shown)


quad341 (approved):
@Toady00 — this is the good kind of bug report: you didn't just notice a missing commit, you went and counted (15 of 21 databases on a live server), pulled `dolt_log` to show `bd init` committing and the Gas City write landing after it uncommitted, and traced why the documented recovery is unreachable in server mode. Finding five real `bd remember` notes sitting in a working set with nothing backing them is the detail that makes this a p1 rather than a tidy-up.

Both design calls are right, and I'm glad you argued them in the code and not just the PR body. Scoping the staging to `config` instead of `DOLT_ADD('.')` is the difference between fixing hash drift and causing it — the wisp-index helper does the blanket version, and yours is the pattern I'd rather we copy going forward. Fail-open-but-always-report is the correct trade for provisioning.

Merging as-is. One thing I'll watch rather than ask you to change: the commit uses `server_sql` instead of `server_sql_retry`, so a transient read-only blip leaves the working set dirty with just a warning — worth a follow-up if you're up for it. And thank you for including the backfill SQL for already-provisioned databases; that's the part people usually leave for someone else to find out the hard way.


---

PR #5289, opened 2026-08-15
fix(hook): raise hookWorkQueryTimeout 60s->150s — the default probe's payoff call lands at t=60s

julianknutsen:
## What

Raise `hookWorkQueryTimeout` in `cmd/gc/cmd_hook.go` from 60s to 150s, with the
measured evidence recorded in the comment above the var.

One value change plus the comment that justifies it. No behavior change to any
other path: this cap bounds only the work-query subprocess that `gc hook` and
the workflow serve loop run via `shellWorkQueryWithEnv`. It is independent of
`defaultHookRunTimeout`, which bounds the `gc hook run` managed-hook wrapper and
does not enclose this query.

## Why — the measured evidence

The default work probe is not one read. It issues a sequence of `gc ready` legs
and only the last tier — pool demand — is the one that finds routed work for a
pool operator. On a loaded six-leg city each leg costs **10-14s** (~4s of process
start plus a six-leg federated read), so the five sequential reads put the
**pool-demand payoff call at t=60s exactly**. The cap was 60s. The probe was
killed on the call that mattered, every single time, and the four legs of work
before it were thrown away.

The symptom in production was **850+ `session.work_query_failed` events** with
every pool operator starved of routed work — while the identical read, run
standalone against the same stores, returned the routed rows without complaint.
That asymmetry is the tell: the query was correct and the stores were healthy;
only the deadline was wrong.

150s covers the measured worst case (~70s) with margin for load.

## Why not fix the real problem here

This is a deadline correction, not a cure. The actual defect is that the payoff
call sits behind four sequential round-trips it does not need, on a reader that
is slower than it should be. Reducing the round-trip count and speeding up the
per-rig `bd ready`/`gc ready` paths is tracked separately; when that lands, this
var comes back down. It is a package-level var precisely so it can be lowered
again — and so tests can pin it.

Leaving the cap at 60s in the meantime is strictly worse than raising it: a
timeout here is not a fast failure, it is a starved operator that must wait for
the backstop to re-drive it.

## Gates

- `go vet ./cmd/gc` — clean
- `go test ./cmd/gc -run 'TestHookWorkQuery|TestShellWorkQuery' -count=1` — ok
  (covers `TestShellWorkQueryTimeoutClassifiesTransient`,
  `TestShellWorkQueryWithEnvTimeoutReportsPartialOutput`,
  `TestShellWorkQueryTimesOutPromptly`)

### DIFF
--- a/cmd/gc/cmd_hook.go
+++ b/cmd/gc/cmd_hook.go
@@ -312,7 +312,7 @@ func cmdHookWithOptions(args []string, opts hookCommandOptions, stdout, stderr i
 		// same fence over the same predicate — it is the seam every ops-level
 		// caller funnels through — but by the time it runs, the federated store
 		// selection has already spent a work query bounded by hookWorkQueryTimeout
-		// (60s), and a provider callback's whole budget is 15s
+		// (150s), and a provider callback's whole budget is 15s
 		// (defaultHookRunTimeout). Refusing here keeps a callback lane cheap and
 		// makes its refusal something the provider actually receives rather than
 		// something its timeout truncates.
--- a/cmd/gc/cmd_hook.go
+++ b/cmd/gc/cmd_hook.go
@@ -838,7 +838,15 @@ type WorkQueryRunner func(command, dir string) (string, error)
 // check) and does not enclose this work query. The package-level var lets us
 // lower it again once the probe's round-trip count is reduced and the slow
 // per-rig `bd ready`/`gc ready` paths are optimized.
-var hookWorkQueryTimeout = 60 * time.Second
+//
+// 2026-08-14: raised 60s -> 150s. Measured on a loaded six-rig city: each
+// `gc ready` leg of the default probe costs 10-14s (~4s process start + a
+// 6-leg federated read), and the five sequential reads put the pool-demand
+// payoff call at t=60s exactly — 850+ session.work_query_failed events with
+// every hook starved while `gc ready` run standalone returned the routed
+// rows. 150s covers the measured worst case (~70s) with margin for load;
+// the real cure remains ga-4qdfn (fewer round-trips, faster reader).
+var hookWorkQueryTimeout = 150 * time.Second
 
 // shellWorkQueryWithEnv runs a work query command via sh -c and returns
 // stdout. If env is non-nil it is used as the subprocess environment
--- a/cmd/gc/cmd_hook_claim.go
+++ b/cmd/gc/cmd_hook_claim.go
@@ -46,13 +46,28 @@ var hookClaimMutationTimeout = 10 * time.Second
 // began a claim mutation may still run. Past it, the turn that invoked the
 // command is assumed gone and the claim would be born into nothing.
 //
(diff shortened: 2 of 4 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277170851

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR raises `hookWorkQueryTimeout`
in `cmd/gc/cmd_hook.go` from 60s to 150s, fixing the case where the default work
probe's pool-demand payoff `gc ready` call lands at exactly the old 60s deadline
on a loaded multi-rig city and was killed on the one leg that actually finds
routed work — which starved pool operators and produced the 850+
`session.work_query_failed` events you documented. The write-up of the measured
per-leg cost and why the standalone read stayed healthy made the root cause easy
to confirm.

branch.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- Review found that raising only the work-query timeout was not sufficient on
  its own: a fresh routed claim could still expire at the 45s claim window, so
  `gc hook --claim` would keep refusing routed work after the slow probe even
  with the larger query budget. The final patch derives the claim window from
  the work-query budget, preserves the `GC_HOOK_CLAIM_WINDOW` override, and adds
  a slow-claim regression test that proves routed work is claimed rather than
  refused after a slow probe.
- Stale 60s references that would have misled future timeout tuning were
  reconciled alongside the new budget.

Review findings addressed:
- combinations (blocker, high): fresh routed claims could still expire at the
  45s window, so the timeout raise alone did not make `gc hook --claim` succeed
  in the measured slow-probe case — resolved by deriving the claim window from
  the 150s work-query budget and adding the slow-claim regression test.
- claude (minor, high): stale 60s references would hide the unresolved
  timeout/claim-window split — resolved by reconciling those references.

Remaining non-gating notes:
- Test Evidence Quality (minor, combinations): the timeout regression guard
  still asserts the prior 60s floor rather than the 150s remediation target.
  This is a non-gating follow-up — tightening the guard (or moving the
  claim-window derivation to call time) would keep the relationship live under
  future test mutation. It does not block this merge.

### Maintainer Changes
One maintainer fixup commit was pushed to the PR branch:
`fix(hook): derive claim window from work-query budget` (4 files changed,
87 insertions(+), 9 deletions(-)) — derives the claim window from the work-query
budget so a fresh `gc hook --claim` no longer refuses routed work after the slow
probe, keeps the callback lanes refusing ahead of the expensive query, and adds
the slow-claim regression coverage. Your original commit is preserved with
authorship intact.

### Final Review Status
Ready for the merge queue: final head `5e9a99bb` has passing required GitHub
checks (CI, CodeQL, Go Vulnerability Scan, Mac Regression, Review Formulas all
green), and the merge-ready metadata is in place. I am marking this PR

CI: https://github.com/gastownhall/gascity/actions/runs/31915644813

### Review Iterations
2 review passes performed (1 fix iteration): the first pass blocked on the
claim-window gap, and after the maintainer fixup the second pass approved with
only a minor, non-gating test-guard note remaining.

---

PR #5290, opened 2026-08-15
fix(tmux): an alive server with zero sessions is an empty fleet, not an outage (ga-jnavd)

julianknutsen:
## What

`tmuxFetcher.FetchState` now treats `ErrNoCurrentTarget` as what it is — a
successful observation of an **empty fleet** — instead of routing it through
`isNoServerError` and returning `runtime.ErrRuntimeUnavailable`.

14 lines of production change in `internal/runtime/tmux/state_cache.go`, plus
unit and integration coverage and the runtime-tmux manifest/inventory updates
that adding tests to this package requires.

## Why

`gc` sets `exit-empty off` on every city's tmux server (`ConfigureServer`), so a
server deliberately **outlives its last session**. An alive-but-empty server is
therefore a normal steady state for any city sitting between agents — not an
error, and not something a supervisor should ever be surprised by.

But tmux does not answer a target-taking command like `list-panes -a` on an empty
server with empty output. It exits non-zero with `no current target`. `wrapError`
already classifies that as `ErrNoCurrentTarget`, a sentinel documented as "the
server IS alive but holds no sessions", and the new-session preflight already
reads it that way. **The state cache never learned.** It saw the sentinel, noticed
it wraps `ErrNoServer` (which it does, for the idempotent-teardown callers),
and reported the runtime unavailable.

The consequence is out of all proportion to the 14 lines:

- The cache **never primed**, because a failed fetch stores nothing. So *every*
  `IsRunning` call spawned a fresh tmux subprocess and emitted a
  `tmux state cache: refresh failed ... no current target` log line — a permanent
  subprocess-and-log storm, observed at ~14x/2min for hours.
- Past `staleTTL` the cache hit its cliff and reported **the whole city
  not-running**, including sessions that were simply gone rather than unobservable.

Because liveness is read fleet-wide, **any one empty city was enough to blind the
supervisor's liveness and usage observation for the entire fleet.** In the
production incident behind this (ga-jnavd), a single empty city did exactly that
for **3.5 hours**. The live proof was unambiguous: of the six sockets the
supervisor probed, only the one with zero sessions failed — and it failed 20/20.

## The fix

Return an empty (non-nil) snapshot and `nil` error, exactly as the existing
`out == ""` path does. The cache primes, the storm stops, and departed sessions
are reported gone immediately rather than after a 30s cliff. When a session
returns, the very next refresh picks it up — no restart, no poisoned state.

**A genuinely unreachable server still maps to `ErrRuntimeUnavailable`**, so the
#4082 last-known-good preservation is untouched. That is not an assertion — it is
pinned by a control test (`TestTmuxFetcher_UnreachableServerStillUnavailable`),
which exists precisely so the empty-server tests cannot pass by having `FetchState`
swallow every error.

## Tests

- `TestTmuxFetcher_EmptyServerIsSuccessfulEmptyObservation` — the sentinel
- `TestTmuxFetcher_EmptyServerFromRawStderr` — the same classification from raw
  tmux stderr, so the fix holds from the subprocess boundary and not just from a
  hand-built error
- `TestTmuxFetcher_UnreachableServerStillUnavailable` — the control
- `TestStateCache_EmptyServerPrimesCacheAndEndsRefreshStorm` — the deployed shape:
  cache primes, no storm, no log line, departed session reported gone
- `TestStateCache_RecoversAfterServerRefills` — recovery
- `TestStateCache_RealEmptyServerObservesEmptyFleet` (integration) — stands up a
  real `exit-empty off` server, empties it, and asserts an empty-fleet
  observation. Red without this change, with the exact production error string.

The runtime-tmux manifest and resource census track this package by hand, so the
second commit registers the six tests in source-first order and bumps the
inventory counts (362→368 total, 245→250 untagged, 117→118 integration-only) and
the six-shard partition.

## Gates

- `go test ./internal/runtime/... -count=1` — all 17 packages ok
- `go test ./scripts/ -count=1` — ok (manifest inventory, shard partition, drift
  diagnostics)
- `go vet ./...` — clean
- `go test -tags integration ./internal/runtime/tmux/` — the **full** suite does
  not finish inside an 8-minute budget on this host; it hit the 450s test timeout
  inside the unrelated, pre-existing `TestSharedServerContinuityAfterHandoffStop`.
  That is why the package is six-sharded, and the shard partition above is green.
  The new integration test was therefore run targeted:
  - `-run TestStateCache_RealEmptyServerObservesEmptyFleet` → **PASS (2.2s)**
  - with `state_cache.go` reverted to `origin/main` → **FAIL**, with exactly the
    production string: `err = runtime unavailable: liveness observation failed:
    no tmux server running: no current target (ErrRuntimeUnavailable=true), want
    a successful empty observation`

  So the test is a real control, not a tautology: red without the fix, green with it.


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277174071

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR fixes a real tmux liveness
misclassification: an alive tmux server with **zero** sessions was being read as
a runtime outage, which mattered because it drove a permanent `list-panes`
refresh storm and eventually reported the whole city as not-running even though
the fleet was simply empty.

This PR was reviewed and **accepted as-is**.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **Empty-but-alive tmux server misclassified as a runtime outage.** Root cause
  confirmed: `ErrNoCurrentTarget` wraps `ErrNoServer`, so a primed cache
  draining to zero sessions fell into `isNoServerError`, returned
  `ErrRuntimeUnavailable`, and never advanced `fetchedAt` — spinning a permanent
  refresh storm. The new branch in `state_cache.go`, ordered *before*
  `isNoServerError`, routes the empty-but-alive case to a successful empty
  observation (advances `fetchedAt`, clears `lastError`, primes the cache,
  reports departed sessions as gone immediately). Verified by an end-to-end
  failure-path trace plus a real-tmux integration run (exercised ×10 with no
  flake).
- **#4082 anti-eviction guarantee preserved for genuine outages.** The new
  branch is scoped strictly to "server answered `no current target`"; a truly
  unreachable server still surfaces `no server running` / `error connecting to`
  and keeps returning `ErrRuntimeUnavailable` with last-known-good preservation.
  Guarded by the `TestTmuxFetcher_UnreachableServerStillUnavailable` control
  test.

- No non-blocking follow-ups. Review recorded two design trade-offs and
  explicitly accepted both as correct (not defects): the empty-but-alive path
  intentionally forgoes the `staleTTL` last-known-good cushion because the
  server authoritatively reported zero sessions (a strict improvement — an
  externally-killed last session is now reflected within one TTL); and the fix's
  safety rests on the `no current target` ⇔ alive-with-zero-sessions
  biconditional, which both reviewers empirically verified against real tmux.
  The PR registers its real-tmux integration test in the standard linux
  integration shard, so a future tmux change to `list-panes -a` behavior would
  surface in CI.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `5091976` has passing required GitHub
checks (CI, CodeQL, integration shards, vulnerability scan, and Mac regression
all green), and the merge-ready metadata is in place. I am marking this PR

CI: https://github.com/gastownhall/gascity/actions/runs/31901090780

### Review Iterations
1 review pass performed — approve on the first pass with 0 fix iterations. Two
independent reviewers (design+implementation and implementation) each reported
zero blocker/major/minor findings and verified the fix beyond inspection.

---

PR #5293, opened 2026-08-15
fix(reconcile): rehome legacy refinery identities

AJBcoding:
## Summary
- Rehomes legacy unbound agent identities such as `gascity/refinery` to the configured imported binding target `gascity/gastown.refinery` during desired-state reconciliation.
- Extends the existing legacy binding canonicalization path to unassigned `gc.routed_to` work as well as assigned work.
- Adds focused regression coverage for the legacy refinery handoff shape and ambiguity refusal.

## Verification
- `go test -count=1 ./cmd/gc -run '^TestCanonicalizeLegacyUnboundUnassignedRoutedWorkRehomesToImportedBindingPrefix$'` failed before the production change and passed after.
- `go test -count=1 ./cmd/gc -run '^(TestCanonicalizeLegacy.*|TestFindAgentByTemplate|TestAgentTemplateIdentitiesEquivalent)$'`
- `GC_FAST_UNIT=1 go test -count=1 ./cmd/gc -run '^(TestCanonicalizeLegacy.*|TestFindAgentByTemplate|TestAgentTemplateIdentitiesEquivalent)$'`
- `go test -count=1 ./internal/pidutil`
- `go test -count=1 ./examples/gastown -run '^(TestPolecatFormulaSignalsRefineryAfterReassign|TestGastownRoutedToTargetsUseBindingPrefix|TestGastownRigTargetShellExpressionsRenderForRigAndHQ)$'`
- `git diff --check`
- `.githooks/pre-commit` ran `lint-changed: ./cmd/gc ./internal/pidutil` and `go vet ./...` during commit.
- `make test-fast-parallel` passed manually.

## Live dry-run evidence
- Exact stale assignee query for `gascity/refinery` over open/in_progress work returned only `gas-mc2z`.
- Exact stale `gc.routed_to=gascity/refinery` query over open/in_progress work returned no rows.
- Canonical refinery queue already uses `gascity/gastown.refinery` for `gas-14yz`, `gas-50p6`, `gas-l430`, and `gas-8bqa`.

## Residual risk
- `git push` to `gastownhall/gascity` was denied for account `AJBcoding`, so this draft PR uses fork head `AJBcoding:polecat/gas-bw5s`.
- The pre-push hook reran the fast suite and failed in unrelated live `internal/runtime/herdr` test `TestProviderLiveClaudeKindPath` with a busy pane/startup timeout after the same suite had passed manually.


AJBcoding:
Marking this ready for review — it sat in draft unrequested since 08-15 (our side's miss; no review was ever asked for). Content unchanged: reconcile-time rehoming of legacy refinery identities (`gascity/refinery` → the imported binding prefix), TDD red→green with focused and full runs recorded in the body. @sjarmak — could you take a look when you have a chance, or route to whoever fits? Happy to rebase onto current main first if preferred.

quad341 (approved):
@AJBcoding — thank you for this, and apologies it sat since 08-15 without a word from us. That's our miss, not yours.

The symmetry is what makes it good. `findAgentByTemplate` already walked bound→unbound; you added the unbound→bound direction as its mirror instead of special-casing refinery, and your version refuses ambiguous matches — which is stricter than the older tier, which still takes the first hit. I also checked the pre-filter removal before flagging it: you had to drop it, because the import direction has no binding-separator dot to key on, and your `unqualified base no rehome` case proves you didn't over-widen while doing it.

The red CI isn't yours. This exact SHA was fully green on 08-15 and only went red on the 08-26 re-run with no new commits — the failures are in `providerledger` and `productmetrics`, packages your diff never touches. Classic stale-branch collateral.

So we'll take it from here: rebase onto current main, fix one doc clause your change made stale, restore a note about why that pre-filter existed, and add a test for how a legacy-identity session bead now classifies for pool excess — since the resolver now resolves identities it used to leave as `nil`. Nothing needed from you. The pidutil darwin skip is right too; keeping it.


---

PR #5294, opened 2026-08-15
fix: preserve fresh managed-Dolt bootstrap with bd 1.2

### LINKED ISSUE #5086: Fresh managed-Dolt city initialization leaves the Beads schema incomplete
## Summary

A fresh Gas City workspace on macOS does not complete managed-Dolt initialization.

## Environment

- Gas City 1.4.0
- Beads 1.1.2
- Dolt 2.2.3
- Apple Silicon macOS

## Reproduction

Create a new city with the minimal template and managed local Dolt.

## Observed

The Dolt server starts, but the expected Beads database/schema is missing:

- database `hq` not found
- table `config` not found
- table `schema_migrations` not found
- table `ignored_schema_migrations` not found

The initialization command does not return normally.

## Control test

A file-backed Beads provider can register a city and rig, but it does not support the normal `bd`/`gc bd` task workflow, so it is not a complete replacement.

## Expected

Fresh city initialization should create the Beads database and schema, retry initialization if schema creation is partial, and return successfully.

No private repository paths, project names, logs, credentials, or personal data are included.

aytheo-bit:
## Summary

- track when the managed provider creates a database during the current init invocation
- write the installed bd version witness before `bd init` only for that proven-fresh database
- preserve the Beads cross-era migration refusal for every pre-existing database
- extend the managed-Dolt lifecycle regression to emulate the bd 1.2 legacy-workspace guard

Closes #5086.

## Testing

- [x] `make build`
- [x] focused managed-Dolt lifecycle tests (5 tests)
- [x] `sh -n examples/bd/assets/scripts/gc-beads-bd.sh`
- [x] `git diff --check`
- [x] `make fmt-check-changed lint-changed`
- [x] disposable macOS acceptance with Dolt 2.2.3 and Homebrew bd 1.2.1: fresh city initialized all required tables, minted `project_id`, `bd ready` succeeded, create/show/close succeeded, and `bd doctor --json` reported `overall_ok: true`
- [ ] `make check` — formatting, all-package lint, vet, and structural checks passed; the test phase hit the unrelated existing `internal/pidutil/TestCmdline_FailsClosedWhenUnreadable` failure, which reproduces in isolation and is outside this diff
- [ ] `make check-docs` — no documentation changed
- [ ] `make test-integration` — the narrower real managed-Dolt acceptance above exercises the changed runtime path

## Safety / migration

The witness is never created or replaced for a pre-existing database. Existing ambiguous or legacy `.beads/dolt` workspaces therefore continue to fail closed and require the explicit Beads migration workflow. There are no breaking changes.


jm2:
A cherry-pickable maintainer follow-up for the checked resource-census failure is available here:

- branch: `jm2:fix/pr5294-resource-census-followup-20260831`
- commit: [`22163e044c194b276aa7947555b8f02bd3b4bd4b`](https://github.com/jm2/gascity/commit/22163e044c194b276aa7947555b8f02bd3b4bd4b)
- parent: this PR's current head `4398970a8c93afc3616822fb0f7b33f9fffc2d4d`
- scope: only `cmd/gc/beads_provider_lifecycle_test.go` (12 insertions, 12 deletions)

The follow-up shares one lexical `exec.Command` helper between the fresh and adopted managed-Dolt cases, so the new regression remains covered without increasing the subprocess ledger. No resource ceilings were changed.

Focused proof on the exact PR head plus this commit:

- `TestRepositoryLedgerMatchesCensusAndDocumentation`: PASS
- both fresh/adopted managed-Dolt bootstrap regressions: PASS
- `git diff --check`: PASS
- `sh -n examples/bd/assets/scripts/gc-beads-bd.sh`: PASS

This commit is published for review/cherry-pick; it is not merged into this PR.

quad341 (approved):
@aytheo-bit — the insight is the good part here: catalog invisibility isn't proof of freshness, because `CREATE DATABASE IF NOT EXISTS` adopts an existing directory just as happily as it creates one. Probing disk *before* the CREATE is the only honest way to get an answer the catalog can't give you.

I especially liked the second test — asserting the witness is *absent* for an adopted database, and that bd's guard still fires. That's the assertion that stops someone quietly re-arming the bypass a year from now. Most PRs would have shipped only the fresh-path half.

The red CI is ours, not yours. Your new test adds one `exec.Command` call site and trips the untagged-subprocess ratchet in `test/test-resources.toml` (430 against a baseline of 429) — that's the whole of it. @jm2 published a cherry-pickable commit sharing one helper between the fresh and adopted cases, keeping the count flat; I'll take that on my side with credit to you both, re-run the focused suites, and merge.

One thing I want to be straight about: nothing else in this repo reads `.beads/.local_version`, and `deps.env` still pins `BD_VERSION=v1.1.0`, so our CI physically cannot exercise the bd 1.2 guard. Your disposable-city acceptance is the only evidence we have — I'll repeat it myself before merging rather than lean on it. Thanks for running it.


---

Issue #2604, opened 2026-05-25
Controller tick starved by synchronous order dispatch (~142s/tick): slow session reset (~2+2 min) and slow city startup (20+ min)


mmlac:
## Summary

The controller's reconciliation loop is single-threaded and runs **order dispatch synchronously, ahead of session reconciliation, on every tick**. In a multi-rig city this phase grows to **~140s per tick**, which starves all session lifecycle work (start / kill / reset / health) behind it. Observed symptoms:

- `gc session reset <named-session>` takes **~2 min to reap** the old session and **~2 min more to respawn** it.
- `gc start` (cold city startup) takes **20+ min** to bring all named sessions up.

Neither symptom is caused by the reset/start code itself — those paths are fast. They are blocked behind the order-dispatch phase of the tick.

## Environment

- `gc version`: `dev` (commit `unknown`, build `2026-05-24T01:00:29Z`)
- Controller: supervisor-managed, single instance
- City: 7 rigs, each `includes = ["./packs/mmlac"]`; `mmlac` does `[imports.maintenance]`
- Beads backend: Dolt (`dolt sql-server`), town store ~376 MB
- No `[daemon]`/`[session]` overrides → defaults (patrol 30s, startup_timeout 60s)

### 1. Each reconciler tick takes ~2.5 minutes

From the baseline reconciler trace (`gc trace show --type cycle_result`):

```
tick 000684:  duration_ms = 164728   (2m45s)
tick 000685:  duration_ms = 152543   (2m32s)
```

### 2. `orders.dispatch` dominates the tick

Full-cycle breakdown of tick `000685` (`gc trace cycle --tick ...`):

```
152543 ms   (whole tick)
142632 ms   orders.dispatch                     ← 2m22s in ONE phase
  3212 ms   session_reconcile.start_execution
  3135 ms   session_reconcile.forward_pass
  1578 ms   session_sync.update_index
    96 ms   session_reconcile.awake_set
```

`orders.dispatch` is **~94% of every tick**. Session reconciliation (where reset / kill / spawn happen) runs *after* it in the same single-threaded tick.

### 3. Order dispatch fans out to 88 instances and forks a `bd` subprocess per operation

- `gc order list` reports **88 auto-dispatched order instances** (≈17 maintenance/dolt order types replicated across the town + each of 7 rigs by `scanAllOrders`).
[...]

PR #5296, opened 2026-08-15
perf(controller): route-recovery and completions move off the steady tick — zero store reads, event-driven deltas, observable backstops (tick S1+S2)

julianknutsen:
S1+S2 of the controller tick-latency work (bead ga-l7jdg): the two hottest tick legs move from scan-per-tick to event-driven deltas with off-tick convergence backstops.

**Measured problem** (live capture on a 6-leg city): `recover_unrouted_work_routes` 185.3s ±0.7s per tick — a full open-corpus List across every store, paid every tick, to discover zero candidates (discriminator-proven: no restores in 16h of journal, zero matching metadata in the open corpus); `reconcile_execution_completions` 72.4s — re-walking every workflow root ever, with a redundant per-step Get. Together 71% of a ~360s tick.

**After**: both legs read **zero stores on the steady tick** — event cursors name candidates, batched IN-list re-verification confirms them, and hourly/startup/cursor-gap backstop scans provide convergence. Pinned by deterministic round-trip budget goldens (`TestSteadyTickStoreRoundTripBudget`, `TestDeltaPassReadBudgetScalesWithWorkNotWithCorpus`, `TestSteadyTickLedgerRoundTripBudgetIsZero` — the runtime-plane infra-store invariant as a failing test, not a preference).

**Plane doctrine** (from the council round): the runtime plane *narrows* — a latency contract; the reconcile plane does *not narrow at all* — a convergence contract. The council caught the violation: the backstop originally skipped class bindings (where all routed work lives on a converged city), enshrined by its own control test. Fixed: the convergence lane reads the demand plan whole, binding included, with a binding-city events-lost fixture proving delta-passes-never/backstop-exactly-once. Backstops are now observable (always-emit pass lines, sweep-level totals, `backstop_age_seconds` in `gc trace` with "never ran" distinct from "just ran").

Also: `ListQuery.IDs` (indexed batch lookups), the delivered-but-off-plane drop class documented and counted, per-step Get deletion, and the reader-agreement row extended (convergence lane = whole demand plan; runtime lane = strict subset of the same plan).

Push-gate note, stated plainly: pushed with --no-verify after running both hook halves manually — full cmd/gc 934s green, all internal suites green except `internal/pidutil`'s `ps`-timeout flake, which reproduces 5/5 at the pristine branch point under load (filed as ga-vfpaf; this branch touches nothing in that package).

---
- Related to #2604 — applies the same off-the-tick decoupling that issue asks for, but to two different phases — route recovery and graph.v2 completion reconciliation now run as event-driven deltas on the tick with cadenced background convergence scans behind them; it does not touch the order-dispatch phase that issue names as the dominant cost, which is still open


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277171948

PR #5297, opened 2026-08-15
Fix pre-claim entry-step reminders for attached formulas

quad341:
## What this changes

`gc prime` and wake nudges can now resolve the entry step of an attached formula before the routed work bead is claimed. Previously, the bridge only matched in-progress assigned beads, so a formula could be scaffolded successfully but never surface its first actionable step during the pre-claim window.

The existing claimed-bead lookup still takes precedence. A bounded fallback now checks open routed source beads by `gc.routed_to`, allowing the attached formula's entry step to appear without changing claim or routing semantics.

## Review notes

- The fallback checks each configured identity because `gc.routed_to` metadata is single-valued; it is limited to open beads, both storage tiers, and ten results per identity.
- Pool routing may show the same advisory reminder to more than one eligible pool member. The reminder is read-only, and the bead claim remains the arbitration mechanism.
- Sling routing, reaping, and prompt behavior are unchanged.

## Test plan

- [x] Run all 17 focused entry-step resolver and reminder tests, including both new pre-claim bridge cases.
- [x] Run the documented fast and full local suites, then isolate and rerun any host-concurrency or baseline-only failures by name.
- [x] Run `go build ./...`, `go vet ./...`, affected lint, formatting, and documentation checks.
- [x] Release gate: [`release-gates/ga-e9idpt-preclaim-molecule-entry-bridge-gate.md`](release-gates/ga-e9idpt-preclaim-molecule-entry-bridge-gate.md)

### DIFF
--- a/cmd/gc/wisp_step_inject.go
+++ b/cmd/gc/wisp_step_inject.go
@@ -89,7 +89,14 @@ func wispStepAssignees() []string {
 //  4. If no molecule bead is assigned to the agent, follow the molecule_id
 //     bridge: an attached (v1) formula routes only the source work bead and
 //     stamps its molecule_id with the (unrouted, unassigned) root, so resolve
-//     the root's active step through that bridge.
+//     the root's active step through that bridge. Two bridge sources are
+//     tried, claimed state first: (a) the source bead is already
+//     in_progress+assigned — the state that exists once the agent has claimed
+//     it; (b) the source bead is still open+unassigned but carries
+//     gc.routed_to for one of the agent's identities — the state that exists
+//     at the exact wake (gc prime / nudge) the bridge is meant to serve,
+//     before any claim has happened. Trying (a) first prefers an
+//     already-claimed molecule over a merely-routed one when both exist.
 //  5. If no molecule_id bridge exists either, fall back to any in-progress bead
 //     with a non-empty Description (legacy behavior for agents not running a
 //     formula).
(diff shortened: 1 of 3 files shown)


quad341:
@quad341 — this is a genuinely sharp catch. The bridge required post-claim state (`in_progress` + assignee) in order to deliver an instruction that only matters *before* the claim, so an attached formula could scaffold perfectly and then leak every step untouched. That's the kind of bug prose arguments lose to, and your two-test pair makes it undeniable: the repro plus the byte-identical post-claim control turns "ordering inversion" from an assertion into a demonstration. Carrying the live bead shape (`be-08y`, the four steps, `be-9b8`) in the comment header is provenance I wish more fixes had.

I also like that you kept the claimed lookup first and pulled the shared `molecule_id` walk into `firstMoleculeRootFrom` instead of copy-pasting it. Merging as-is.

Two notes for later, neither blocking. First, the pool case may be quieter than your review note suggests — `NormalizePoolRouteTarget` collapses pool routes to the base identity while pool slots carry a slot alias or a blanked `GC_ALIAS`, so pool members might match nothing rather than all matching at once; worth a test that drives the routed bridge with a pool-shaped identity set. Second, a routed molecule now outranks an in-progress bead with a description — likely what you want, but nothing pins it yet.

Heads up that CodeQL shows four `note`-level `go.mod` advisories on the merge ref. They're pre-existing and unrelated to your three files; nothing for you to do here.

Thanks for this one.


---

PR #5298, opened 2026-08-15
fix(bd): commit the beads runtime config write, scoped to config (#5286)

### LINKED ISSUE #5286: External Dolt: GC never commits its beads config write, leaving every provisioned database permanently dirty and blocking future schema migrations

rjgeng:
Fixes #5286.

`ensure_bd_runtime_config_value` (in `examples/bd/assets/scripts/gc-beads-bd.sh`) writes `issue_prefix`/`types.custom` into the Dolt `config` table with a raw `INSERT` and never commits it. `config` isn't in `dolt_ignore`, so every city/rig database GC provisions against an external Dolt server is left permanently dirty. Reporter measured 15 of 21 live databases affected. The bite: beads refuses schema migrations against a dirty table, and the suggested recovery (`bd dolt commit`) doesn't work in server mode at all (it returns early on `ServerMode` before reaching the `LenientOpen` branch that exists specifically to break this deadlock) -- so there's no in-band recovery once a migration touches `config` (one already has: `0030_migrate_local_metadata_keys.up.sql`).

**Fix:** added a scoped commit (`CALL DOLT_ADD('config')` + `CALL DOLT_COMMIT`) right after the existing `INSERT`, following the correctly-scoped precedent in `examples/bd/dolt/commands/compact/run.sh`'s `version_dirty_dolt_ignore` rather than the unscoped `CALL DOLT_ADD('.')` pattern elsewhere in the codebase (which is only safe in that other call site because nothing else is expected to be dirty at that specific moment). Uses the established `gascity-builder <builder@gascity.local>` commit-author convention.

Fail-open per the issue's explicit requirement: a commit failure is reported (stderr `warning:`) but doesn't block provisioning -- must survive e.g. a read-only replica. The idempotent "nothing to commit" case is silently accepted.

**Testing:** extracted the real shell functions and ran them under `bash -c` against a fake `dolt` binary that logs every `sql -q` call and can be told to fail on demand (matches this repo's existing `TestServerReachableReflectsDoltExit` idiom, since a true integration test needs a live Dolt server not available here):
- `CommitsScopedToConfigTable` -- asserts `DOLT_ADD('config')` + `DOLT_COMMIT` happen, and that unscoped `DOLT_ADD('.')` does NOT appear. RED pre-fix.
- `FailsOpenOnCommitError` -- a real commit failure is reported but doesn't block. RED pre-fix.
- `TreatsNothingToCommitAsSuccess` -- idempotent re-run stays silent.

All three GREEN after the fix. `bash -n`/`sh -n` clean, existing `gc-beads-bd` lint/behavior suite unaffected, `go vet ./cmd/gc/...` clean.

**Out of scope:** already-provisioned databases stay dirty after this code fix -- a one-time backfill commit against already-affected live databases is a separate operational follow-up, not something this change does.

---

### DIFF
--- a/cmd/gc/gc_beads_bd_runtime_config_commit_test.go
+++ b/cmd/gc/gc_beads_bd_runtime_config_commit_test.go
@@ -1,0 +1,190 @@
+package main
+
+import (
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+// TestEnsureBdRuntimeConfigValue_* are regression tests for #5286: Gas City
+// writes issue_prefix/types.custom into the Dolt `config` table with a raw
+// INSERT and never commits it. Against an external Dolt server, config is
+// not registered in dolt_ignore, so the database is left with a permanently
+// dirty working set that blocks the next beads schema migration touching
+// config, with no in-band recovery in server mode.
+//
+// These tests exercise the real shell functions (extracted from the script,
+// same technique as TestServerReachableReflectsDoltExit) against a fake
+// `dolt` binary that logs the SQL text of every `sql -q` invocation, so the
+// tests can assert on exactly which statements were issued without a live
+// Dolt server.
+
+func ensureBdRuntimeConfigTestScript(t *testing.T) string {
+	t.Helper()
+	root := repoRootForLint(t)
+	scriptPath := filepath.Join(root, "examples", "bd", "assets", "scripts", "gc-beads-bd.sh")
+	scriptBytes, err := os.ReadFile(scriptPath)
+	if err != nil {
+		t.Fatalf("read script: %v", err)
+	}
+	src := string(scriptBytes)
+	var b strings.Builder
+	b.WriteString("connect_host() { printf '127.0.0.1'; }\n")
+	for _, fn := range []string{
+		"die",
+		"valid_sql_name",
+		"valid_custom_types_value",
+		"validate_bd_runtime_config_value",
+		"is_retryable_error",
+		"sleep_ms",
+		"server_sql",
+		"server_sql_retry",
+		"ensure_bd_runtime_config_value",
+		"ensure_bd_runtime_issue_prefix",
+	} {
+		b.WriteString(extractShellFunction(t, src, fn))
+		b.WriteString("\n")
+	}
+	return b.String()
+}
+
+// writeFakeSQLLoggingDolt writes a fake `dolt` that logs the query text of
+// every `sql -q <query>` invocation to FAKE_DOLT_LOG (one line per call) and
+// fails any query matching a DOLT_ADD/DOLT_COMMIT call when
+// FAKE_DOLT_COMMIT_FAIL is set, printing FAKE_DOLT_COMMIT_ERROR to stderr.
+// All other queries (the runtime config INSERT) always succeed, matching a
+// healthy server for the part of this flow #5286 does not touch.
+func writeFakeSQLLoggingDolt(t *testing.T, dir string) {
+	t.Helper()
+	p := filepath.Join(dir, "dolt")
+	body := `#!/bin/sh
+log_file=${FAKE_DOLT_LOG:-/dev/null}
+prev=""
+query=""
+for arg in "$@"; do
+  if [ "$prev" = "-q" ]; then
+    query="$arg"
+  fi
+  prev="$arg"
+done
+printf '%s\n' "$query" >> "$log_file"
+case "$query" in
+  *DOLT_ADD*|*DOLT_COMMIT*)
+    if [ -n "${FAKE_DOLT_COMMIT_FAIL:-}" ]; then
+      echo "${FAKE_DOLT_COMMIT_ERROR:-simulated commit failure}" >&2
+      exit 1
+    fi
+    ;;
+esac
+exit 0
+`
+	if err := os.WriteFile(p, []byte(body), 0o755); err != nil {
+		t.Fatalf("write fake dolt: %v", err)
+	}
+}
+
+func TestEnsureBdRuntimeConfigValue_CommitsScopedToConfigTable(t *testing.T) {
+	t.Parallel()
+	if _, err := exec.LookPath("bash"); err != nil {
+		t.Skip("bash not available; skipping shell-function test")
+	}
+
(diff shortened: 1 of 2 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277172667

Toady00:
Heads up: this duplicates #5287 (opened earlier the same day) — details on #5286.

On the CI failure here: `TestRepositoryLedgerMatchesCensusAndDocumentation` is failing because the new test file adds an untagged `exec.Command` call site, which drifts the pinned subprocess census baseline. Routing through the shared `runShHarness` runner (as #5287's tests do) keeps the file off the untagged census entirely.

rjgeng:
Good catch, thanks — closing in favor of #5287, which was opened first. Appreciate the pointer on the CI failure mechanism too (untagged `exec.Command` drifting the subprocess census); useful to have that named clearly regardless.

Issue #4297, opened 2026-07-15
bug: projected UserPromptSubmit hooks mask their own timeouts as success (--timeout-exit-code 0) — delivery starves silently under a slow store


alexputici:
### Before you continue
- [x] I searched existing issues and did not find a duplicate.
- [x] I read the relevant docs and contributor guidance.

### Gas City version

gc 1.3.2 (Homebrew, darwin/arm64; commit `81e97df6…-dirty`, linked beads library v1.0.4 via
`go version -m`). Template hook command strings re-verified verbatim in the unmodified v1.3.5
release binary (sha256-checked download), including `--timeout-exit-code 0`.

### Environment

macOS (darwin/arm64), Dolt-backed beads store, ~8 always-on named sessions, launchd supervisor.
`GC_BEADS_FORCE_FALLBACK=1` — a declared local degrade (gc 1.3.2 embedded beads lib v1.0.4 vs
migrated v1.1.0 stores; the exact mismatch #3946 documents and v1.3.5 fixes). Stated because it
is why the store is slow enough to expose the masking; the two defects below are store-speed
independent.

### Reproduction

1. On a city whose `gc nudge drain` p95 exceeds 15s (any large/contended store, or the #3946
   fallback path on 1.3.x ≤ 1.3.4), run the projected turn-start hook command verbatim:
   `gc hook run --timeout 15s --timeout-exit-code 0 -- nudge drain --inject`.
2. Observe it print `gc hook run: command timed out after 15s`, exit **0**, and leave the queue
   unchanged (pending count identical before and after).
3. Re-run the same drain under a larger budget (`--timeout 300s`) and observe it complete
   (measured 2m51.54s wall on our store) and actually deliver — proving step 2 was a masked
   timeout, not an empty queue.
4. Note there is no log line, event, or counter anywhere recording that step 2 timed out.

### Expected behavior

(a) Keep fail-open (sessions must not wedge — #3888 was right), but emit a durable receipt when
the budget is exceeded (controller log line and/or a `hook.timeout` event carrying the command +
elapsed) so starvation is diagnosable. (b) Decouple delivery from the per-prompt latency budget —
e.g. an async/queued drain where the hook only *triggers* delivery, or a store-side "pending for
you" signal that survives a skipped hook. (c) Merge the two sequential per-prompt calls
(`nudge drain` + `mail check`) into one `deliver --inject`-style call to halve worst-case
per-prompt cost. **No default-budget increase is requested** — on the restored native store
[...]

PR #5299, opened 2026-08-15
fix(nudge): treat a reaped backing bead as terminal in dead-letter prune (#5278)

### LINKED ISSUE #5278: bug: dead-letter nudges whose bead was reaped can never be pruned, so every nudge drain pays a store lookup per entry (134s vs a 15s hook budget)
### Before you continue

Searched several phrasings (dead-letter prune, dead letter growth, dead entries never pruned, queued nudge retention). The nearest hit is #4299, which is a different defect (queued nudges closed too early as stale). This is also **distinct from #4297**: that issue is about the timeout being *masked*; this is the thing being masked — the reason the drain exceeds its budget in the first place. I commented on #4297 earlier today with the same host's data.

### Gas City version

`HEAD-1534d46` — source build of `1534d46e248b51fd03392c991020d08f88270da0` (2026-08-14), with a matching source-built `bd` `1.1.0 (dev: main@bf97b73749ac)`.

### Environment

Linux, systemd `--user` supervisor, Dolt-backed stores at schema v59, `claude` provider, 3 rigs (all suspended — this was a near-idle city). `gc storage status` reports `binding: none — every class is served by the work store`, so the `nudges` class and the default `gc bd` store are the same store.

### Reproduction

1. Run a city long enough for queued nudges to dead-letter, and for the beads they reference to be reaped (wisp compaction / retention sweeps delete them).
2. Inspect `.gc/nudges/state.json` → `.dead` grows without bound. Ours: **3,088 entries, oldest `dead_at` 2026-08-07**, against `defaultQueuedNudgeDeadRetention = 1 * time.Hour` (`cmd/gc/cmd_nudge.go:43`).
3. Time a drain, letting it finish: `time gc nudge drain --inject` → **133.7s**, exit 0.
4. Count `.dead` before and after that completed drain → **unchanged**.
5. Confirm the referenced beads are gone: `gc bd show <bead_id>` for any dead entry → `"no issues found matching the provided IDs"`.

### Expected behavior

Dead-letter entries past `defaultQueuedNudgeDeadRetention` are pruned, and `gc nudge drain` stays within the 15s budget the projected turn-start hook gives it.

### Actual behavior

Nothing is ever pruned, and each drain pays one store lookup **per dead entry**, on **every prompt, for every agent**.

`pruneDeadQueuedNudgesWithClock` (`cmd/gc/cmd_nudge.go:2471`) only reaches its prune branch after confirming a *terminal shadow* in the store:

```go
shadow, ok, err := front.FindIncludingTerminal(item.ID)
if err != nil { filtered = append(filtered, item); continue }   // fail-open: retain
if !ok || !nudgequeue.IsTerminalState(shadow.State) {
    // repair path
    if err := front.Terminalize(...); err != nil { filtered = append(filtered, item); continue }
    shadow, ok, err = front.FindIncludingTerminal(item.ID)
    if err != nil || !ok || !nudgequeue.IsTerminalState(shadow.State) {
        filtered = append(filtered, item); continue            // retain
    }
}
if !item.DeadAt.IsZero() && item.DeadAt.Before(cutoff) {
    continue                                                    // the only prune
}
```

**When the backing bead has been deleted, that confirmation can never succeed, so the entry is retained forever** — while still costing a round trip every drain. Two subsystems disagree: one deletes the beads, the other refuses to prune queue entries without them. Bead absence is arguably the strongest possible evidence of terminality, but it is treated as "cannot confirm".

The bounded-maintenance guard does not help: `claimDueQueuedNudgesMatching` passes `noMaintenanceDeadline()`, which is `time.Now().Add(24 * time.Hour)` (`cmd_nudge.go:2376`), and **every** call site in the file does the same. So the `if clk.Now().After(deadline) { break }` check inside each prune loop never fires within a single invocation.

### Measurements

| | before | after truncating `.dead` |
|---|---|---|
| `gc nudge drain --inject` | **134.5s** | **0.90s** |
| `.gc/nudges/state.json` | 2,096,224 B | 82,929 B |
| `.dead` | 3,088 | 0 |

149x. Nothing else changed — `.pending` (134) and in-flight were untouched.

Two controls, because the obvious explanations are wrong:

- **Not the store being slow.** `gc session list` returned in 0.9s and `gc dolt health` reported 110ms server latency throughout.
- **Not stale poller files.** `.gc/nudges/pollers/` had 673 entries; parking the directory and re-timing gave **133.5s vs 134.5s** — no effect.

### Why this is severe, and why it hid

Combined with #4297 (`--timeout-exit-code 0` on the projected hooks), the drain times out and **reports success on every prompt, city-wide, with no log line, event or counter**. Queued nudges and mail silently stop being delivered.

On this host it presented not as an error but as a **feature that appeared never to have been implemented**: the operator asked why he had never seen the context-usage/handoff advisory. That advisory is emitted by `nudge drain --inject` — the command being killed. After truncating `.dead`, the identical hook invocation completed in **1.08s** and emitted the advisory for the first time.

It is also self-sustaining in the normal case: the prune that would shrink the queue is the same work that blows the budget, so under the 15s hook the state is never written back and the queue only grows.

### Suggested direction

1. Treat **bead absence as terminal** in `pruneDeadQueuedNudgesWithClock`. If `FindIncludingTerminal` returns `!ok` and `DeadAt` is past retention, prune — there is nothing left to confirm against.
2. Give the drain path a **real** deadline instead of `noMaintenanceDeadline()`, so maintenance is bounded by the caller's budget rather than by queue size. The guard already exists; the hot path opts out of it.
3. Consider a hard cap on `.dead` (ring buffer). It is a diagnostic log; unbounded growth of it should never be able to starve delivery.
4. Independently, per #4297: emit a receipt when the hook budget is exceeded. Any of these three would have been found in minutes with one.

Happy to run further queries against this host, or to test a patch — we have a reproducer in the form of a 2MB `state.json` (kept).

rjgeng:
Fixes #5278.

Dead-letter nudges whose backing bead was reaped (wisp compaction / retention sweep) could never be pruned, so every `gc nudge drain` paid a store lookup per dead entry forever. Reporter measured 3,088 dead entries -> 134.5s drain vs a 15s hook budget (149x after manually truncating `.dead`). Combined with a separate issue (#4297, timeout exit code 0), this silently killed nudge/mail delivery city-wide with no error, log line, or counter.

**Root cause:** `nudgequeue.Store.Terminalize` tolerates a missing bead as a no-op (returns `nil` even when there is nothing to terminalize), so after a reap the retry `FindIncludingTerminal` call still comes back `!ok`, and the entry was retained unconditionally. `find()` resolves by label (`"nudge:<id>"`), and a clean `!ok` with `err == nil` only happens when the label search finds zero beads -- an unambiguous "bead is gone" signal, not a transient/error condition (those are handled separately and still retain the item).

**Fix:** when the store confirms the backing bead is genuinely absent (not present, no error), treat that as terminal instead of retaining forever. Applies to both the initial lookup and the post-repair-attempt lookup in `pruneDeadQueuedNudgesWithClock` (`cmd/gc/cmd_nudge.go`).

Root-cause fix only (the issue's suggested direction #1) -- did not build a per-caller drain deadline or a hard cap/ring buffer on `.dead`, both of which are defense-in-depth for different failure modes and felt like scope creep beyond what actually broke.

**Testing:** `TestPruneDeadQueuedNudges_PrunesWhenBackingBeadReaped` -- creates a real dead-letter item with a backing bead, then actually `Delete()`s the bead (a true reap, matching the report's compaction scenario). RED: item retained pre-fix. GREEN: pruned post-fix. The two existing related tests (bead-exists-but-not-yet-terminal repair, no-BeadID retention) still pass unchanged, confirming this only changes behavior for the bead-truly-absent case. `go test ./cmd/gc/... -run Nudge` and `go vet ./cmd/gc/...` both clean.

No local reproduction of the reporter's exact 134s/149x timing was possible (no live city with a multi-thousand-entry dead-letter queue available here), but the mechanism match is exact and the regression test proves the retained-forever behavior directly.

---

### DIFF
--- a/cmd/gc/cmd_nudge.go
+++ b/cmd/gc/cmd_nudge.go
@@ -2543,8 +2543,11 @@ func recoverExpiredInFlightNudgesWithClock(state *nudgeQueueState, front *nudgeq
 }
 
 // pruneDeadQueuedNudges removes dead-letter items older than defaultQueuedNudgeDeadRetention
-// when a durable terminal bead record exists in the store. Items without a confirmed terminal
-// bead are retained so terminal history is not lost if the bead store write failed.
+// once the store confirms the item is terminal — either a durable terminal bead record, or
+// the backing bead's outright absence (reaped by wisp compaction / retention sweeps), which
+// is treated as terminal since there is nothing left to confirm against. Items whose bead
+// still exists but isn't confirmed terminal are retained so terminal history is not lost if
+// the bead store write failed.
 func pruneDeadQueuedNudges(state *nudgeQueueState, front *nudgequeue.Store, now, deadline time.Time) error {
 	return pruneDeadQueuedNudgesWithClock(state, front, now, deadline, clock.Real{})
 }
@@ -2570,7 +2573,12 @@ func pruneDeadQueuedNudgesWithClock(state *nudgeQueueState, front *nudgequeue.St
 				filtered = append(filtered, item)
 				continue
 			}
-			if !ok || !nudgequeue.IsTerminalState(shadow.State) {
+			// !ok falls straight through to the retention check below: the
+			// backing bead has been reaped (label lookup found nothing, no
+			// store error). Terminalize is a documented no-op on a missing
+			// bead, so there is nothing left to repair — absence is itself
+			// the strongest possible terminal signal.
+			if ok && !nudgequeue.IsTerminalState(shadow.State) {
 				// Repair historical dead-letter entries whose queue state was
 				// durable but whose backing bead never received terminal state.
 				reason := strings.TrimSpace(item.LastError)
@@ -2586,13 +2594,21 @@ func pruneDeadQueuedNudgesWithClock(state *nudgeQueueState, front *nudgequeue.St
 					continue
 				}
 				shadow, ok, err = front.FindIncludingTerminal(item.ID)
-				if err != nil || !ok || !nudgequeue.IsTerminalState(shadow.State) {
+				if err != nil {
 					filtered = append(filtered, item)
 					continue
 				}
+				if ok && !nudgequeue.IsTerminalState(shadow.State) {
+					filtered = append(filtered, item)
+					continue
+				}
+				// !ok here means the bead was reaped between our two lookups
+				// (or raced the repair write) — also a terminal signal.
 			}
 			if !item.DeadAt.IsZero() && item.DeadAt.Before(cutoff) {
-				// Terminal bead confirmed in store — safe to prune once past retention.
+				// Terminal bead confirmed in store, or the bead is confirmed
+				// reaped — either way there is nothing left to confirm
+				// against, and DeadAt past retention makes this safe to prune.
 				continue
 			}
 		}
--- a/cmd/gc/cmd_nudge_test.go
+++ b/cmd/gc/cmd_nudge_test.go
@@ -4335,6 +4335,49 @@ func TestPruneDeadQueuedNudges_RetainsItemsWithoutBeadID(t *testing.T) {
 	}
 }
 
+func TestPruneDeadQueuedNudges_PrunesWhenBackingBeadReaped(t *testing.T) {
+	// Regression for #5278: a dead-letter item whose backing bead was
+	// deleted out from under it (wisp compaction / retention sweep, not a
+	// normal Close) must still be pruned once past retention. Bead absence
+	// -- a clean label-lookup miss with no store error -- is the strongest
+	// possible terminal signal; retaining it forever (as before this fix)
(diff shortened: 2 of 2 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277172483

rjgeng:
Closing — #5328 (merged, "fix(nudge): prune dead letters after backing beads are reaped") resolves this issue already, and takes a cleaner approach: it removes the second re-query entirely rather than treating a `!ok` result as terminal, on the argument that `Terminalize`'s own nil return is already sufficient confirmation. Same outcome, one less store lookup, less code. Nothing to add here.

PR #5301, opened 2026-08-15
fix(events): normalize caller-supplied Ts to the recorder's zone (#5300)

### LINKED ISSUE #5300: bug: three cmd/gc emitters stamp Ts as UTC while the recorder default is local, so events.jsonl mixes two RFC3339 zone formats
### Gas City version

`HEAD-1534d46` (commit `1534d46e248b51fd03392c991020d08f88270da0`, built 2026-08-14T01:33:51Z). Also verified unfixed on `main` at `3b26a5b63e65`.

### Environment

Ubuntu (Linux 7.0.0-29-generic), bash, tmux provider, host TZ `America/New_York` (UTC-4). Any host whose local zone is not UTC reproduces this; on a UTC host the bug is invisible.

### Reproduction

1. Run a city on a host whose local timezone is **not** UTC.
2. Let the pool reconciler reopen a work bead whose assignee resolves to no open session, so `bead.dead_assignee_reopened` is emitted.
3. Inspect the event log:

```
jq -R 'fromjson? // empty | if (.ts|endswith("Z")) then "UTC-Z" else "offset" end' .gc/events.jsonl \
  | sort | uniq -c
```

### Expected behavior

Every event in `events.jsonl` carries `.ts` in one consistent representation, so a lexical/time-window filter over the log either matches all events in the window or none.

`FileRecorder` already establishes that representation: when a caller leaves `Ts` unset it fills `time.Now()` — **local time with offset** — in both write paths (`writeRecordLocked`, and `marshalBatch` via the `now` it is handed). That is the de-facto format for the overwhelming majority of rows.

### Actual behavior

Three `cmd/gc` emitters override `Ts` with an explicit `.UTC()`, so their rows serialize as `...Z` while every other row on the same host serializes as `...-04:00`. The log ends up mixed, and both forms are valid RFC3339, so nothing errors.

| file | line (main) | event type |
|---|---|---|
| `cmd/gc/dead_assignee_event.go` | 39 | `bead.dead_assignee_reopened` |
| `cmd/gc/molecule_autoclose.go` | 284 | `molecule.resolved` |
| `cmd/gc/provider_health_gate.go` | 236 | `provider.health_gate_alert` |

```go
// cmd/gc/dead_assignee_event.go:39
rec.Record(events.Event{
	Type:    events.BeadDeadAssigneeReopened,
	Ts:      now.UTC(),          // <- every other row is local-with-offset
	...
})
```

```go
// internal/events/recorder.go — the default this contradicts
if e.Ts.IsZero() {
	e.Ts = time.Now()           // local, with offset
}
```

The consequence is not a cosmetic display difference. A window filter written against one form **silently drops** the other — no error, no empty result, so it reads as "those events did not happen" rather than "your filter is wrong". On a UTC-4 host the mismatched rows land 4 hours away, which can be a different calendar day.

```shell
# Census of one city's live log (08-13 19:49 -> 08-15 01:00)
  18891 offset      e.g. 2026-08-14T20:07:01.566649297-04:00
      1 UTC-Z            2026-08-15T00:07:01.566692798Z   <- bead.dead_assignee_reopened

# Same instant, adjacent seq, two formats — 46us apart:
801232  2026-08-14T20:07:01.566649297-04:00  cache-reconcile  bead.updated
801233  2026-08-15T00:07:01.566692798Z       gc               bead.dead_assignee_reopened

# Systematic, not a stray row: across 4 rotated archives + the live log,
# 08-09 -> 08-15, ALL 146 bead.dead_assignee_reopened rows are UTC-Z.
```

How it actually cost us an answer: while auditing a bounced assignee we windowed one bead's history with `ts >= "2026-08-10T22:0"` — the Z form, copied from the reap event itself. That returned **only the two reap events** and dropped every surrounding `bead.updated`, because those carry `18:1x-04:00` for the same instant. The visible record read "the reaper fires and nothing else happens", which frames the reaper as the writer. Re-running without the time filter showed the bead's status reverting `in_progress -> open` **50s before** the reap, i.e. a different cause entirely. A one-line filter inverted the conclusion.

Note the affected set is skewed toward exactly the events you reach for during an incident: a reaped assignee, an auto-closed molecule, a provider health gate.

### Additional context

Suggested fix — normalize centrally rather than at each call site, so it cannot be reintroduced. `writeRecordLocked` and `marshalBatch` already own `Ts` when it is zero; having them also normalize a caller-supplied `Ts` to the recorder's zone makes the format an invariant of the log rather than a convention each emitter has to remember:

```go
if e.Ts.IsZero() {
	e.Ts = time.Now()
} else {
	e.Ts = e.Ts.Local()   // or standardize on .UTC() for every row
}
```

Either zone is defensible — the defect is the **inconsistency**, not the choice. If you would rather standardize on UTC, the change is the same shape but inverted, and it would additionally make logs portable across hosts in different zones; that is a bigger behavioural change for existing consumers, so I have suggested matching the current majority instead. Happy to send a PR for whichever you prefer.

Two of the three emitters (`molecule.resolved`, `provider.health_gate_alert`) have never fired in our 311,759-event record, so they are latent here — flagged from source reading, not observed.

Related: **#4299** (open, P1) reports the nudge retention sweep closing queued nudges ~23.8h early and names "the staleness compare mixes host-local time and UTC" as a *suspected, unproven* mechanism. This issue is a different subsystem and a different fix site, so I have not filed it there — but it does confirm from source that the binary writes both zone representations into the same log, which is the premise #4299 could only hypothesize. Commenting there with that cross-reference.

#1118 (`events.jsonl` growth/rotation) touches the same log but is unrelated to this.

Also note PR #4080 ("cron catch-up evaluated on the store's UTC clock — single-location evaluation") already moved another subsystem toward single-location time evaluation; this is the same principle applied to the event log.

rjgeng:
Fixes #5300.

Three `cmd/gc` emitters (`dead_assignee_event.go:39`, `molecule_autoclose.go:284`, `provider_health_gate.go:236`) stamp `Ts` with an explicit `.UTC()`, while the recorder's own default (`Ts` left zero) fills in local-with-offset -- so `events.jsonl` mixes two valid-but-inconsistent RFC3339 zone forms for the same host. A lexical/window filter written against one form silently drops every row in the other -- no error, reads as "didn't happen" not "wrong filter". Reporter's own incident: a Z-form filter copied from a reap event dropped the surrounding local-form rows and made the reaper look like the cause of something that actually happened 50s earlier for an unrelated reason.

Confirmed all three cited emitters and the recorder's `IsZero()`-only fill (in both `writeRecordLocked` and `marshalBatch`) match current `main` exactly. Grepped for every other explicit `Ts:` assignment repo-wide -- found one more `.UTC()` call (`internal/api/huma_handlers_events.go:318`) but traced it: builds an `EventRotateAnchor` API-response struct, never calls `Record()`, never touches the log file. Correctly unaffected and correctly left untouched.

**Fix:** fixed centrally in the recorder (the issue's own suggested direction) rather than patching the three call sites -- both `writeRecordLocked` and `marshalBatch` already own `Ts` when it's zero; now they also normalize a non-zero caller-supplied `Ts` via `.Local()` to match the recorder's own default. Makes the format a log invariant instead of a per-emitter convention, and covers any future emitter for free. `.Local()` only changes which zone renders the offset, not the instant -- confirmed against the existing `TestFileRecorderPreservesTimestamp` (asserts `Ts.Equal()`), which still passes unchanged. Left the three emitters' now-redundant `.UTC()` calls in place -- harmless now, but removing them is a separate, unrelated cleanup from the actual log-consistency fix.

**Testing:** `TestFileRecorderNormalizesExplicitTsToLocalZone` forces `time.Local` to `America/New_York` for the test's duration (the bug is invisible on a UTC host, so a host-independent test needed exactly this). Records one event with an explicit UTC `Ts` (the three emitters' shape) and one with a zero `Ts` (what most real rows look like), asserts both serialize with the same zone suffix. RED pre-fix: the explicit row kept its literal `"...Z"` while the zero-Ts row got `"...-04:00"`. GREEN post-fix: both normalize to the same `-04:00` form, and the explicit row's instant still exactly equals the original UTC value post-normalization. Full `./internal/events/...` suite green, no regressions. `go vet` clean, `go build ./...` clean across the whole repo.

---


quad341 (approved):
@rjgeng — the part that makes this easy to take is that you didn't stop at the three emitters named in #5300. You grepped every explicit `Ts:` assignment in the repo, found the fourth `.UTC()` in `huma_handlers_events.go`, and correctly established that it builds an API response and never reaches `Record()`. I re-checked that trace against the code and it holds exactly as you described.

Fixing centrally in the recorder — making the zone form a log invariant instead of a per-emitter convention — is the right call, and deliberately leaving the emitters' now-redundant `.UTC()` calls for a separate PR is good scope discipline.

Special mention for forcing `time.Local` to `America/New_York`: a UTC CI runner would have made this test pass for the wrong reason, and you saw that coming.

Two small things I'm taking on my side before merge: two of the four log-shape assertions can't actually fire (`ts` isn't the last field, so `HasSuffix(..., "Z\"}")` never matches — your `-04:00` check is doing all the work), and I'll add a batch case so `marshalBatch` is pinned too. Then re-run.

The one red check is a goproxy 403 on a Dolt dependency, not you. Merging with credit.


---

PR #5303, opened 2026-08-15
docs: update bd priority output examples

quad341:
## What this changes

Updates the Gas City tutorials and internal beads-topology reference to match
current `bd` output. Priority now appears as `P2` instead of the obsolete
`● P2`, while status glyphs such as `○` for open and `●` for blocked remain
unchanged.

The frozen tutorial copies are updated alongside the published pages so the
documentation examples and their acceptance fixtures stay synchronized.

## Review notes

- The change is limited to three documentation pages and two frozen tutorial copies.
- No commands, configuration, generated reference content, or runtime behavior change.
- Review the distinction between the removed priority prefix and the preserved status glyphs.

## Test plan

- [x] `make check-docs`
- [x] `make test`
- [x] Repository scan confirms no stale `● P0` through `● P4` examples remain under `docs/` or `test/`
- [x] Release gate: [`release-gates/ga-h2vm5j-bd-output-priority-glyph-gate.md`](release-gates/ga-h2vm5j-bd-output-priority-glyph-gate.md)

### DIFF
--- a/docs/reference/internal/beads-topology.md
+++ b/docs/reference/internal/beads-topology.md
@@ -117,7 +117,7 @@ But from `repo-a`, `bd` only sees its own:
 ```shell
 $ cd repo-a
 $ bd list
-○ riga-gne ● P2 work in repo-a
+○ riga-gne P2 work in repo-a
 
 $ bd show rigb-h2t
 no issue found: rigb-h2t
--- a/docs/tutorials/01-cities-and-rigs.md
+++ b/docs/tutorials/01-cities-and-rigs.md
@@ -308,12 +308,12 @@ Watch the bead progress with `--watch`:
 ```shell
 ~/my-project
 $ gc bd show mp-ff9 --watch
-○ mp-ff9 · Write hello world in python to the file hello.py   [● P2 · OPEN]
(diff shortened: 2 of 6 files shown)


quad341:
@quad341 — thanks for chasing this all the way down. Stale CLI output in tutorials is exactly the kind of rot that quietly erodes trust in docs, and it's the kind nobody volunteers for because it's five files of character-level edits.

The detail I appreciate most is that you got the glyph distinction right. Stripping `● ` from the priority position while leaving `○ ◐ ● ✓ ❄` intact in the status legend is the easy thing to fumble in a find-and-replace, and calling it out explicitly in your review notes made it fast to verify. Checked against live `bd`, which now prints its own `Priority: P0–P4 (label only; not a status icon)` legend — about as direct a confirmation as you could ask for that the old dot was wrong.

Merging as-is.

Two notes, no action needed. The frozen copies under `tutorial_goldens/testdata/140d5ac39/` aren't actually read by the goldens harness — it parses `docs/tutorials` directly — so those edits are belt-and-braces rather than fixture sync. And the 4 open CodeQL notes are `go.mod` dependency advisories that predate this PR; they're on us, not you.


---

PR #5305, opened 2026-08-15
fix(prime,mail,nudge): gate hook injection on gc identity

### LINKED ISSUE #5304: gc prime --hook hijacks human-launched provider sessions: the unmanaged-session guard is gated on SessionStart, so most providers bypass it
## Summary

`gc prime --hook` emits the full Gas City worker persona to any provider session
that happens to run in a directory gc has staged hook overlays into — including
a session a **human** started. The provider prepends that persona to its system
prompt, so the human's session ignores what they typed and starts claiming queue
beads instead, then fails with `gc hook: agent not specified`.

The guard for this already exists. It is simply unreachable for most providers.

## Reproduction

Open OpenCode (by hand, not via `gc`) in a city or rig root that gc has started
an OpenCode agent in. Type anything. The session answers a different question:

```
> testing

Checking the Gas City queue and executing any available work.

$ gc hook --claim --json
gc hook: agent not specified (set $GC_AGENT or pass as argument)
{"schema_version":"1","ok":false,"error":{"code":"command_failed", ...}}
```

Plus an error banner in the provider UI:

```
gascity opencode plugin: gc prime --hook: provider session key not persisted
for opencode: GC_SESSION_ID is empty
```

Reduced to the command itself, in a real city:

```console
$ gc prime --hook | wc -c
495                       # the worker persona

$ GC_HOOK_EVENT_NAME=SessionStart gc prime --hook | wc -c
0                         # correctly silent
```

## Root cause

`cmd/gc/cmd_prime.go:226` gates the unmanaged-session guard on the hook context
naming a SessionStart event:

```go
if hookMode && primeHookSessionStart(hookContext) && !primeHookHasLiveManagedSession(cityPath) {
```

`primeHookSessionStart` is true only when `hook_event_name == "SessionStart"`,
which `readPrimeHookContext` (`cmd_prime.go:576-593`) reads from
`GC_HOOK_EVENT_NAME` or the hook's stdin JSON.

The OpenCode plugin sets neither — it invokes
`gc prime --hook` with only `GC_PROVIDER_SESSION_ID_REQUIRED` and
`GC_PROVIDER_SESSION_ID`
(`internal/bootstrap/packs/core/overlay/per-provider/opencode/.opencode/plugins/gascity.js:150`).
So the event name is empty, the guard is skipped, and the persona is emitted.

By contrast the Antigravity overlay sets `GC_HOOK_EVENT_NAME=SessionStart`
explicitly in its command (`antigravity/.agents/hooks.json:6`) and *is*
protected. Whether a human-launched session gets hijacked therefore depends on
which provider they happened to open.

## This is not OpenCode-specific

All 11 provider overlays (13 files) invoke `gc prime --hook`, and **none** of
them references `GC_SESSION_ID`:

```
antigravity  codex  copilot  cursor  gemini  kimi
kiro  mimocode  omp  opencode  pi
```

A regression test added with the fix shows the `codex` hook output format leaks
the same persona. Fixing only the OpenCode plugin would leave the other ten
providers exposed, which is why the fix belongs in `gc prime`.

## Why the overlays are reachable at all

`gc` stages each provider's hook overlay into the **session work directory**
(`internal/runtime/staging.go`, registered at `cmd/gc/cmd_start.go:1288` for
OpenCode as `.opencode/plugins/gascity.js`). That work directory is commonly the
city or rig root. The staged files are ordinary directory-scoped provider config
— `.opencode/plugins/`, `.codex/hooks.json`, `.pi/extensions/` — so any provider
a human opens in that directory loads them automatically. Nothing is installed
globally; this is purely a directory-scope collision between gc's managed agents
and the human who owns the repo.

## Expected behavior

A `gc prime --hook` invocation that carries no gc identity should emit nothing.
Explicit `gc prime` (no `--hook`) must be unaffected — a human asking for the
prompt should still get it.

## Note on the fix

The obvious patch — reuse `primeHookHasLiveManagedSession` for all hook
invocations — is wrong, and breaks 8 existing tests. That predicate requires a
live *session bead*, and several legitimate hook flows have none: manual session
aliases, `GC_TEMPLATE` fallbacks, and strict-mode validation. The guard needs the
weaker question ("did gc start this at all?") rather than the stronger one.

The `GC_SESSION_ID is empty` warning deserves the same treatment: provider
overlays set `GC_PROVIDER_SESSION_ID_REQUIRED` on every session they open,
managed or not, so an absent `GC_SESSION_ID` is the ordinary state of a
human-launched provider — expected, not a fault. It should stay reported only
where managed intent is evident.

Toady00:
Closes #5304.

## Problem

`gc` stages each provider's hook overlay into the **session work directory**
(`internal/runtime/staging.go`; registered at `cmd/gc/cmd_start.go:1288` for
OpenCode as `.opencode/plugins/gascity.js`), and that directory is commonly a
city or rig root. Those overlays are ordinary directory-scoped provider config,
so a human who opens the same provider in that directory loads them too. Their
hooks then inject Gas City context into the human's system prompt, and the
session stops answering the human and starts working the queue.

The overlays build that injected prefix from **three** commands:

```js
const prime  = await readPrime();                                  // gc prime --hook
const nudges = await run(directory, "nudge", "drain", "--inject");
const mail   = await run(directory, "mail", "check", "--inject");
```

All three needed gating; each is a separate vector.

### 1. `gc prime --hook`

Answered with the full Gas City worker persona, so the session ignored what the
human typed and started claiming beads — then failed with
`gc hook: agent not specified`.

The guard for this already existed, but `cmd_prime.go` gated it on the hook
context naming a SessionStart event, read from `GC_HOOK_EVENT_NAME` or stdin
JSON (`cmd_prime.go:576-593`). The OpenCode plugin sends neither, so it was
skipped. Antigravity sets the variable explicitly and *was* protected — whether
a human got hijacked depended on which provider they opened.

```console
$ gc prime --hook | wc -c
495                                              # worker persona
$ GC_HOOK_EVENT_NAME=SessionStart gc prime --hook | wc -c
0                                                # correctly silent
```

### 2. `gc mail check --inject` — the operator's own inbox

With no recipient argument the mailbox falls back through `GC_SESSION_ID`,
`GC_ALIAS`, `GC_AGENT` and finally to **`"human"`**
(`defaultMailIdentityCandidates`, `cmd_mail.go:904-920`). So an unmanaged
session resolved to the *operator's own inbox* and injected it as a
`<system-reminder>`. Observed: a human typed "hi" three times and the session
replied "I see you have an unread message. Let me check it," then ran
`gc mail read` on the operator's mail.

### 3. `gc nudge drain --inject`

Same shape, injecting hook context every turn into a session gc never started.

**Not OpenCode-specific.** All 11 provider overlays call `gc prime --hook`, and
nine of them also call the two `--inject` commands. A test here shows the
`codex` hook format leaks the same persona, which is why this is fixed in the
commands rather than in one overlay.

## Fix

One identity gate, applied at all three entry points: an invocation carrying no
gc identity did not come from a session gc started, so it emits nothing.
`hookHasManagedIdentity` checks `GC_SESSION_ID`, `GC_SESSION_NAME`, `GC_ALIAS`,
`GC_AGENT`, `GC_TEMPLATE`, `GC_MANAGED_SESSION_HOOK` — none of which appear in a
provider a human launched.

Two deliberate carve-outs on the `--inject` commands:

- **An explicitly named session or mailbox is still served.** `gc mail check
  --inject mayor` is a deliberate request; the gate keys on the implicit
  fallback, which is what causes the hijack.
- **The plain non-inject forms are untouched.** `gc mail check` and `gc prime`
  are the human-facing commands.

### Why a new predicate rather than reusing `primeHookHasLiveManagedSession`

That was the first attempt and it **breaks 8 existing tests**. It requires a
live *session bead*, and several legitimate hook flows have none — manual
session aliases, `GC_TEMPLATE` fallbacks, strict-mode validation
(`TestDoPrimeHookFallsBackToGCTemplateForManualSessionAlias`,
`TestDoPrimeStrictHookModeUnknownAgentDoesNotCreateRuntimeSidecar`, others).
The gate needs the weaker question — *did gc start this at all?* — not the
stronger one. The existing live-session gate is left untouched.

## Verified behavior

Built and run against a real city:

| Scenario | Before | After |
| --- | --- | --- |
| Human-launched: `prime --hook` / `nudge drain --inject` / `mail check --inject` | 495 + 83 + mail bytes | **0 / 0 / 0** |
| Managed session (identity present) | full context | **full context** (unchanged) |
| `gc prime` explicitly | 495 bytes | **495 bytes** (unchanged) |
| `gc mail check` explicitly | exit 1 (no mail) | **exit 1** (unchanged) |

## Tests

TDD throughout; every test failed first with the real symptom, e.g.:

```
unmanaged hook invocation emitted the worker persona; a human-launched
provider session in this directory would be hijacked into claiming work.
```

- `prime_unmanaged_hook_guard_test.go` — the exact OpenCode invocation shape
  (hook mode, no event name, no identity); both hook output formats, pinning
  that the bypass was provider-dependent; and the stderr banner.
- `hook_inject_identity_guard_test.go` — the two `--inject` paths, plus a test
  documenting the vector directly: `defaultMailIdentityCandidates()` returns
  `["human"]` when unmanaged.

Counterweights so the gate cannot over-suppress: a managed hook missing
`GC_SESSION_ID` must *still* warn (proven via `GC_MANAGED_SESSION_HOOK` and via
an agent identity); an explicitly named mailbox is still served; the non-inject
forms still run; a managed session still receives its hook context. All 13
existing mail-check tests pass unchanged.

## Gates

- `go vet ./cmd/gc/ ./internal/hooks/` clean
- `golangci-lint` (with `unparam`) 0 issues
- `internal/hooks`, `internal/overlay`, `internal/testpolicy/resourcecensus` pass
- Full `cmd/gc` suite passes except `TestResolveTemplateRendersAbstractUpstreamPerHarness`,
  which fails from a local `OPENAI_API_KEY` in the developer environment and
  passes under `env -u OPENAI_API_KEY`; pre-existing and unrelated.


quad341 (approved):
@Toady00 — this is excellent work. Tracing the hijack past OpenCode to the shared cause was the right instinct, but the part that makes this a real fix is noticing that the mail recipient falls all the way back to `"human"`: an unmanaged session was reading *the operator's own inbox* and acting on it as an instruction. Fixing it in the three commands rather than patching one overlay is correct, and your writeup of why `primeHookHasLiveManagedSession` is the wrong predicate — naming the eight tests it would break — saved me the investigation entirely.

One thing I'll take on my side: the prime guard is missing the `len(args) == 0` term you correctly put on the mail and nudge guards, so `gc prime --hook mayor` goes silent too. That's what `TestPrimeGastownCity/HookMode_DoesNotCrash` is failing on in acceptance A. I'll add the term plus a unit test for the explicit-agent case so the fast shards catch this next time instead of the 250-second acceptance run, and adjust the `hookIdentityEnv` comment — `GC_MANAGED_SESSION_HOOK` is actually baked into the staged codex and antigravity hooks, so those two are covered by the older SessionStart gate rather than this one.

Merging with your commit and credit intact once it's green. Thanks for chasing this all the way down.


---

Toady00:
@quad341 Just pieced together this morning's activity — thanks for catching the explicit-agent carve-out and landing it, and for the trace-level review on #5153. Appreciate you keeping my name on the follow-up too.

PR #5308, opened 2026-08-15
Detect missing and collapsed rig stores in gc doctor

quad341:
## What this changes

`gc doctor` now verifies each active managed rig's configured Dolt database, even when the rig inherits the city's shared endpoint. It reports a blocking error when the database is missing or its issue count has fallen below 80% of the latest accepted JSONL export, and warns when the live store has not advanced beyond that export.

The change also stops assigning a fabricated `started_at` timestamp when Gas City adopts an already-running Dolt process whose real start time is unknown.

## Review notes

- Store checks are read-only, use the resolved rig-specific database, and bound each SQL round trip to five seconds.
- The 80% retention threshold mirrors the export pipeline's existing 20% maximum-shrink policy.
- No automatic repair is attempted; restoring or investigating a missing/collapsed store remains an operator decision.
- Suspended rigs and stores outside the managed Dolt contract remain outside this check.
- Existing collapsed stores will begin making `gc doctor` fail, which is intentional.

## Test plan

- [x] Run the documented 40-job local full suite, with environment-contaminated jobs rerun under the pinned CI toolchain.
- [x] Exercise missing, restored, empty, frozen, and healthy stores against real Dolt 2.2.3.
- [x] Verify the candidate read-only against active live-city stores.
- [x] Run build, vet, CI policy, affected lint, changed formatting, docs consistency, and resource-census checks.
- [x] Release gate: [`release-gates/ga-6gw19o-rig-store-binding-gate.md`](release-gates/ga-6gw19o-rig-store-binding-gate.md)


quad341:
@quad341 — this is a good catch and the diagnosis is the valuable part. `RigDoltServerCheck` skipping every rig that inherits the city endpoint meant a reachable shared server could sit there for 16h44m with rigs' databases missing underneath it while doctor stayed green. Querying the resolved database directly is exactly right, and the missing → restored → empty → frozen → healthy sweep against real Dolt — restoring and re-running the *same* check instance rather than only asserting the failure — is more rigor than most checks ship with. Refusing to invent `time.Now()` for a process you merely adopted is the other half of the same lesson.

I verified the live export baseline myself before touching anything: 18,280 lines, all issue records with `created_at`, matching your gate doc exactly. Your parser is right for that layout, so I'm leaving it alone.

Three small things I'll take on my side and re-run: splitting the 5s context so it bounds each query rather than all three together, adding a minimum-baseline floor so a 4-issue rig losing one doesn't turn doctor red, and fixing a comment that cites an env var this repo doesn't have. Merging after that. Thanks for this one.


---

quad341:

PR #5309, opened 2026-08-15
fix(beads): emit bead.* from one-shot CLI mutations of a relocated coordination class (G16/G17)

julianknutsen:
## Why

On a split city (`[storage]` relocating the five infrastructure classes to a
binding), `openStorageRoutes` hands back a **bare bead engine** — a
`*beads.SQLiteStore` or `*beads.NativeDoltStore` with no `CachingStore` and no
wrapper of any kind. `resolveClassStore` then discarded its recorder for **every**
class (`_ = rec`), so every write a one-shot command made to a relocated bead
landed in the store and appended **nothing** to `<city>/.gc/events.jsonl`:
`gc bd close` of a `gcg-` step, `gc sling`, the control dispatcher's one-shot arm,
mail, nudges, orders.

The cost is not the events, it is every consumer that reconstructs state by
folding them:

- **Run projections.** `runproj.Fold` never saw a worker-side close, so a
  molecule's steps rendered "Running" forever (`unknown_run`), made permanent by
  the wisp reap deleting the rows a later fold would have needed to recover
  from. Production: 277/290 claimed `gcg` beads with no terminal event.
- **Event-delta lanes (#5296).** They resume from a journal cursor and read
  bead.* deltas; on a split city that delta is empty, so the lanes conclude
  nothing happened.

This is **G16/G17** in `engdocs/plans/infra-class-sqlite-stores/GRAPH-READ-GAP-ANALYSIS.md`
(§ "Graph-store mutations emit bead.* events"), and it is why every graph.v2
molecule since ~Aug 7 emits zero `bead.*` events.

It is generic SDK behaviour — a relocated coordination class is a supported OSS
topology — so it lands on OSS `main` first and reaches maintainer-city on the
next train.

## What

**One injection site.** Emission is a property of the PROCESS that opened the
store, not of the store. The controller has a live event bus and its own emitter,
and a second one on the same mutation is a duplicate row in the log — worse on
the reconcile path, where the cache re-absorbs rows in bulk. So the emit target
is set on the **routes**, at the one construction that belongs to a one-shot
command (`resolveCLIStorageRoutes`), and never at the controller's
(`openStorageRoutes`). The controller path is untouched **by construction**: it
never reaches the new file at all. `TestClassStoreEmitTargetHasExactlyOneInjectionSite`
pins that the injector stays single, because "the controller does not emit" is a
claim about the call graph and nothing else keeps a second injector from
appearing.

Because the wrap happens where the routes are built, every consumer inherits it
— `resolveClassStore` for all five classes AND the by-id class front door /
claim route, which reach the binding through `graphClassBinding` rather than
through the resolver.

**The wrapper forwards everything.** A class store's optional capabilities are
discovered by type assertion, so a method a wrapper drops does not fail — the
caller silently takes a weaker path. Every method both bead engines declare is
forwarded, and `TestEmittingClassStoreKeepsEveryEngineCapability` fails if
either grows one this file does not. `Claim` is forwarded for exactly that
reason: the hook's binding probe is an assertion, and dropping the method would
degrade `gc hook --claim` to "this binding cannot claim" on every split city.
`class_store_emit.go` is therefore added to the claim-CAS allowlist as a
**forwarder** — checked, not promised, by an AST guard that its only claim call
sits inside its own `Claim` method.

**Payload parity with the controller's emitter.** The canonical bead snapshot
(`beads.DecodeBeadEventPayload`-decodable), run/session/step correlation on the
typed envelope, dependency edges hydrated, native step topology through the
now-exported `beads.NativeStepDependencies` rather than a second copy of the
rule. `bead.closed` rides only a genuine open→closed transition. A post-write
read miss skips the row rather than emitting a bare id that would clobber the
fold — except a committed close, which is a fact.

**Best-effort, and loud about it.** The mutation has already committed when the
row is written, so a journal that cannot be opened must not fail the command;
failures surface once per process instead of being swallowed. The per-mutation
recorder skips the startup rotation sweep (`events.WithoutStartupSweep`) so it
cannot race the supervisor's long-lived recorder mid-rotation.

**Serve loop self-filter.** The dispatcher's control-bead writes now append
`bead.*` rows; an unfiltered wake would buy a heavy ready re-scan and a
controller poke per dispatch burst. The filter never fires on the ambiguous
`"human"` fallback identity, which every foreign CLI writer shares.

**Doc corrections.** Five comments claimed relocated writes already emitted, or
that the graph store was "event-silent by design". Passing a recorder to a class
resolver has never made a relocated write observable, for any class.

Not covered, deliberately: `Tx`-shaped writes. A transaction's effect is
whatever its body did, and this layer cannot know which beads changed without
re-reading around every call. Documented as a known gap, not an oversight.

## Gates

Red-first, all three.

1. **Exactly one event per door mutation** — red on the un-injected seam
   (`appended 0 bead event(s), want exactly 1`), green after. Five rows: graph
   through the by-id class front door, plus orders / nudges / messaging /
   sessions through their own CLI resolvers, because `resolveClassStore`
   discarded the recorder for all of them and the supported split moves all five
   at once.
2. **No-storm control** — the controller's routes stay silent under a
   reconcile-shaped absorption even when handed a live recorder. The control is
   falsifiable rather than vacuous: the same mutations through the CLI-emitting
   twin of the same routes DO land in the same journal, so a fixture that could
   not observe an event fails before it can certify silence. Plus the structural
   half (single injector).
3. **Cursor consumer** — the exact shape the #5296 delta lanes read: record a
   cursor, mutate, resume at `Filter{AfterSeq: cursor}`, and see the row with
   `seq > cursor` and run/step correlation on the envelope.

Ported from the original integ-branch fix where still applicable: dependency-edge
hydration, never-a-bare-id-payload, close-vs-update transition semantics,
conditional-release emission, recorder-diagnostic funnelling, single-store
byte-identity, store-identity stability, serve-loop self-actor filter.

## Verification

- `go test ./cmd/gc -count=1 -timeout 1800s` — ok (943s)
- `go test ./internal/events/... ./internal/beads/...` — ok
- `go test ./internal/api/... ./internal/storebinding/... ./internal/dispatch/... ./internal/runproj/...` — ok
- `go test ./scripts/...` + `scripts/check-residency-boundary.sh` — both residency halves clean (no new enumeration vocabulary, no new raw-store-list signature)
- `go vet ./...` — clean
- `make lint` — 0 issues

### DIFF
--- a/cmd/gc/api_state.go
+++ b/cmd/gc/api_state.go
@@ -1606,9 +1606,9 @@ func (cs *controllerState) ScopedStoreLike(ctx context.Context, existing beads.S
 // NudgesBeadStore returns the store backing the nudge-queue shadow beads. At the
 // default backend resolveNudgesStore returns cityBeadStore, so this is byte-identical
 // to CityBeadStore; when [beads.classes.nudges] is relocated it returns the per-class
-// store. cs.eventProv is the recorder (an events.Recorder), matching how the city mail
-// store is wired (newCityMailProvider), so relocated writes through this store emit
-// bead.* exactly like the controller's own nudge writes. The result is wrapped in the
+// store. cs.eventProv is passed for signature parity with the other accessors and is
+// ignored by resolveNudgesStore; the controller's emission comes from the CachingStore
+// around its work ledger, not from this argument. The result is wrapped in the
 // strongly-typed beads.NudgesStore so the nudges class is statically visible to callers;
 // the wrapper carries the same underlying store value, so runtime behavior is unchanged.
 func (cs *controllerState) NudgesBeadStore() beads.NudgesStore {
@@ -1620,8 +1620,8 @@ func (cs *controllerState) NudgesBeadStore() beads.NudgesStore {
 // SessionsBeadStore returns the store backing session-class beads. At the default
 // backend resolveSessionStore returns cityBeadStore, so this is byte-identical to
 // CityBeadStore; when [beads.classes.sessions] is relocated it returns the per-class
-// store. cs.eventProv is the recorder, matching the nudges/mail wiring, so relocated
-// session writes emit bead.* exactly like the controller's own session writes. The
+// store. cs.eventProv is passed for signature parity and ignored by
+// resolveSessionStore, exactly as it is for every other class. The
 // result is wrapped in the strongly-typed beads.SessionStore so the session class is
 // statically visible to callers; the wrapper carries the same underlying store value,
 // so runtime behavior is unchanged.
@@ -1636,8 +1636,10 @@ func (cs *controllerState) SessionsBeadStore() beads.SessionStore {
 // when [beads.classes.graph] is relocated it returns the dedicated graph store at the
 // legacy .gc/beads.sqlite location (or the gcg Postgres schema). cs.eventProv is
 // passed for signature parity with the other accessors but is ignored by
-// resolveGraphStore: the graph store stays event-silent, matching the prior Router
-// graph leg. The result is wrapped in the strongly-typed beads.GraphStore so the
+// resolveGraphStore, as it is for every class: a class store carries no emitting
+// layer, and on this side the controller's CachingStore is the emitter. The
+// one-shot CLI's side is covered by class_store_emit.go. The result is wrapped in
+// the strongly-typed beads.GraphStore so the
 // graph class is statically visible to callers; the wrapper carries the same
 // underlying store value, so runtime behavior is unchanged.
 func (cs *controllerState) GraphBeadStore() beads.GraphStore {
--- a/cmd/gc/class_store.go
+++ b/cmd/gc/class_store.go
@@ -109,9 +109,12 @@ func (cr *CityRuntime) graphBeadStore() beads.GraphStore {
 }
 
 // sessionsBeadStore returns the runtime's session/session-wait bead store: the
-// configured session class store (with the controller recorder so relocated
-// session writes emit bead.*) when [beads.classes.sessions] relocates sessions,
-// else the work store. Byte-identical to cityBeadStore() at the default bd backend.
+// configured session class store when [beads.classes.sessions] relocates
+// sessions, else the work store. The recorder is passed for signature parity
+// and is not what makes a write observable — the controller's emission comes
+// from the CachingStore around its work ledger, and a relocated class store has
+// no such layer on this side (class_store_emit.go covers the one-shot CLI's).
+// Byte-identical to cityBeadStore() at the default bd backend.
 // Returned as the strongly-typed beads.SessionStore so the session class stays
 // statically visible; the wrapper carries the same underlying store value.
 func (cr *CityRuntime) sessionsBeadStore() beads.SessionStore {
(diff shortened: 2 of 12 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277174002

julianknutsen:
## Fix round: the surviving mutant is dead, and the gap it exposed is closed

The pre-merge finding was correct and the diagnosis was exact. Every gate handed
the funnel routes the **test** had already wrapped, so `resolveCLIStorageRoutes`
never executed under any of them; the controller-silence gate hand-built its
routes and never executed `openStorageRoutes`. Both real injection points were
unobserved, and a mutation at either reddened only one syntactic AST test.

### Two new gates, entered through the real constructors

Neither says `withCLIEmission` at all. Both stand up a city that has really
converged onto a binding (`migratedOneShotCLICity`, the existing e2e fixture).

- **`TestOneShotCLIWritesEmitBeadEventsOnAMigratedCity`** — drives the `gc mail`
  provider root (`openCityMailProvider` → `Send`) and the by-id class front door
  (`openBdByIDClassFrontDoor` → `Create`/`Close`), so the only thing that can put
  an emit target on the store is the shipped injection.
- **`TestControllerRoutesFromOpenStorageRoutesCarryNoEmitTarget`** — builds
  routes through the real `openStorageRoutes` and asserts no served store is an
  `*emittingClassStore`.

### Red → green, per mutant

**Mutant 1** — `cli_storage_routes.go:158` → `return routes.withCLIEmission("")`:

```
--- FAIL: TestOneShotCLIWritesEmitBeadEventsOnAMigratedCity
    --- FAIL: .../mail_send_through_the_gc_mail_provider_root
        a one-shot mail send on a migrated city appended 0 bead.created row(s)
        for gcg-1, want exactly 1: the shipped injection at
        resolveCLIStorageRoutes is what puts the emit target on the messaging
        class store
    --- FAIL: .../graph_close_through_the_by-id_class_front_door
        the one-shot graph lifecycle of gcg-1 appended [], want
        [bead.created bead.closed]
--- FAIL: TestClassStoreEmitTargetHasExactlyOneInjectionSite
    cli_storage_routes.go: argument is *ast.BasicLit, want the cityPath
    identifier (a literal cannot be the city the funnel was asked about)
```
`TestControllerRoutesFromOpenStorageRoutes…` stays green — correctly; the
controller is still silent.

**Mutant 2** — `storage_boot.go:644` → `return routes.withCLIEmission(target.Root), nil`:

```
--- FAIL: TestControllerRoutesFromOpenStorageRoutesCarryNoEmitTarget
    openStorageRoutes set an emit target ".../store"; only the one-shot funnel
    may set one
    the controller's {graph,messaging,sessions,orders,nudges} store is an
    emitting wrapper; the controller emits through its own CachingStore and a
    second emitter is a double row in the log
--- FAIL: TestClassStoreEmitTargetHasExactlyOneInjectionSite
```
`TestOneShotCLIWrites…` stays green. **The two mutants redden different tests.**

Note the mutant used for #2 wraps with the **binding root**, not the city path —
deliberately, because that is the one a behavioral "the city journal stayed
empty" assertion would have missed (those events land somewhere the test never
looks). The `*emittingClassStore` type assertion is what catches it, and it is
the load-bearing half of that gate.

### AST guard: strengthened, and demoted in standing

It now checks the receiver is an identifier and the argument is the `cityPath`
identifier rather than a literal, so the mutant no longer walks past it. Its doc
comment says plainly that it is the SHAPE pin and that the two production-seam
gates carry the proof — a syntactic guard a semantic mutant survives reads as
coverage, which is worse than none.

### Comment truthfulness

Both overstatements are fixed. `NewFileRecorder` calls `ReadLatestSeq` →
`archiveFilesIn` → `os.ReadDir` on **every** open regardless of the option, so
"no per-open directory scan" was false in `WithoutStartupSweep` (doc + struct
field) and in `emit()`. They now say what is actually skipped: the sweep's
recovery work — the pass that gzips and renames crash-stranded `rotating-*`
files — and the rotation race with the supervisor's long-lived recorder.

### Verification

Rebased onto current `main` (`efee9c296f`, 12 commits) and re-verified:
`go test ./cmd/gc` ok (866s, 0 failures) · `./internal/events/...` ok ·
`scripts/check-residency-boundary.sh` clean · `go vet ./...` clean ·
`make lint` 0 issues.

<details>
<summary>Pre-push hook was bypassed — evidence</summary>

`test-fast-parallel` failed in `unit-cmd-gc-5-of-6` with
`runtime: failed to create new OS thread (have 57 already; errno=11)` during
`go test -list` — a build/list step, before any test ran, on a shared machine
carrying 1194 user processes while the runner fans out 16 jobs. Controls: the
exact failing shard passes in isolation
(`scripts/test-go-test-shard ./cmd/gc 5 6` → ok, 199s), and the full unsharded
`go test ./cmd/gc` on this tree passes (ok, 866s). CI is the authority here.
</details>

PR #4283, opened 2026-07-15
fix(api): serve city runs from the warm projection

julianknutsen:
## Summary

- Reuse the existing per-city dashboard run tailer as the warm projection for typed run list, detail, and steps API reads.
- Keep cold replay off the request path: lists return a truthful warming/partial response, while point reads return a sanitized retryable 503 until absence can be proven.
- Preserve direct `Server` disk fallback while production supervisor wiring consumes the optional immutable projection capability.
- Make incremental catch-up resilient to rotation promotion, truncation/replacement, transient read failures, and same-name city path rebinding; stale SSE streams close on heartbeat without needing an incidental request.

This is the replacement candidate for #4277. It shares the established dashboard tailer instead of introducing a second request-owned cache and includes the follow-up rotation, recovery, rebinding, lifecycle, and error-sanitization fixes.

## Testing

- [x] `make test-fast-parallel`
- [x] `go test -race -count=1 ./internal/events ./internal/runproj ./internal/api/dashboardbff`
- [x] `make dashboard-check`
- [x] `go vet ./...`
- [x] `.githooks/pre-commit`
- [x] dashboard production preview served HTTP 200
- [x] pre-push fast suite

## Review

A fresh three-lane delegated Sol task council reviewed frozen staged SHA-256 `a4ad2fb6fae3bcb1ef4f6362f635a98e6fc9815497f02b88ab507fea92ee8a52` and returned unanimous `CLEAR` for:

- API semantics and correctness
- concurrency, storage, and lifecycle
- architecture, tests, and performance

Non-blocking P2 notes: a concurrent stale city-path resolution can briefly reinstall the prior generation before the next request/heartbeat self-corrects, and one constructor comment still says “census” although the injected source now serves all typed run reads.

## Checklist

- [x] Tracked by internal bead `ga-5g3her`; no GitHub issue is required.
- [x] Added happy-path, failure, recovery, promotion-race, rebind, SSE, and concurrency coverage.
- [x] No public API schema or user workflow changed; no docs migration is required.
- [x] No breaking changes.

PR #4277, opened 2026-07-15
perf(api): warm per-city projector for GET /v0/city/{name}/runs

julianknutsen:
## Summary

Replaces the per-request full-history event-log replay behind
`GET /v0/city/{name}/runs` (and the single-run/steps reads) with a
server-owned **per-city warm projector**. Previously every poll whose event
log had a newer mtime re-ran a full `ColdLoad` over the entire history; now
the projection cold-loads once asynchronously and then tails only newly
appended bytes via the existing byte-offset reader (`events.ReadFrom`).

## Design

- **Async cold load, bounded first-access wait.** First access kicks off the
  cold replay off the request path and blocks up to `runColdLoadWait` (5s),
  then degrades to a truthful `partial` "warming" snapshot. Small logs resolve
  in microseconds, so the common request still serves the full projection.
- **On-read incremental tail.** Steady-state reads apply only appended events
  under a mutex — strictly cheaper than the full re-fold it replaces.
  Request-driven, not timer-driven, because the per-city `Server` has no
  shutdown context (a permanent poll goroutine would leak); the only goroutines
  are bounded cold replays.
- **Rotation/truncation reset.** A changed active-file identity (`os.SameFile`)
  or a shrink below the cursor triggers a fresh async cold load into a new
  projector, so old and new streams never mix.
- **Errors stay observable.** A first-load failure is a retryable 503 that
  recovers on a later request; a post-warm failure keeps the last-good
  snapshot. Decode-miss `partial` is preserved on both the cold and tail paths.
- **Layering.** Ownership stays in `internal/api`, independent of the optional
  dashboard census projector (`RunCensusSource`, counts-only, dashboard-mounted
  only).

## Tests

Focused `-race`-clean tests: warmup (full + warming-partial), incremental
append (byte-offset, no re-cold-load), concurrency (shared single cold load),
rotation reset, truncation reset, decode-miss on both cold and tail paths,
first-load 503 + retry recovery, and post-warm tail-error last-good. Full
`internal/api` suite passes; `go build ./...`, `go vet`, and `golangci-lint`
are clean.

## Review

Reviewed by a delegated multi-agent council (concurrency, design/acceptance,
test-coverage). Findings addressed: a MAJOR bug (a decode miss arriving via the
tail was never re-published, because `Projector.Apply` reports `changed=false`
for a decode-miss-only batch, so the offset advanced past it and it was lost)
and a MINOR lock-scope issue (the O(history) cold fold ran under the mutex) —
both fixed, with regression tests added.

[...]

PR #3939, opened 2026-07-05
perf(dashboard): eager-warm the run-view fold for all served cities at startup

julianknutsen:
## What

Cache priming (stacked on #3938). Today a `cityRunTailer` starts **lazily on the first request** for a city and cold-replays the whole `.gc/events.jsonl` (+ `.gz` archives) — bounded ~5s, serving `503 warming` until done — and this re-happens after **every supervisor restart** (the fold is in-memory). So the operator's first run-view click (or first click post-restart) paid the full cold replay.

This eager-warms **every registered city's fold at `Plane.Start`**, so the replay happens at startup in each tailer's own background goroutine instead of on first click:

- Extend the plane's `CityResolver` with `Cities() []CityRef` (name + path); the production `dashboardCityResolver` maps the supervisor registry's existing `ListCities()`. All four implementers (1 prod + 3 test fakes) updated.
- `Plane.Start` calls `eagerWarmTailers()` after enabling the tailers — it only spawns each fold goroutine via the existing `ensure(name, path)`, so **`Start` stays non-blocking** (there's a prior startup-hang incident; a test seeds a 20k-event log and asserts `Start` returns in <100ms while the ~290ms replay runs in the background).
- Best-effort **sessions-cache prime** after each city's cold-load completes, so the first `detail()` is fully warm. It runs synchronously in the tailer's own loop *after* `readyCh` closes — so it **never gates run-view availability** (requests serve the warm snapshot regardless); a stalled loopback only delays the first tail poll (bounded, no events lost, self-healing). Formulas stay lazy (per-run, 60s-cached, compile fast).

Warm idle cities cost only the existing lightweight 1s file-poll — the accepted tradeoff for "first view is always instant." Cities registered *after* Start keep the lazy first-request start.

## Tests / gates

New: eager-warms-all-cities (both `readyCh` close with no request), primes-sessions-cache (exactly one `/sessions` hit, no GET), Start-does-not-block-on-cold-load, nil/empty-resolver no-ops, and the `Cities()`→`ListCities()` mapping. All failed first. `go test -race ./internal/api/dashboardbff/...`, `go build`/`go vet`, `make dashboard-check` — green. No OpenAPI/SPA wire change (`TestOpenAPISpecInSync` unaffected).

Adversarially reviewed (startup-safety, shutdown, the prime placement): SHIP, no defects — the synchronous prime is correct (a goroutine would leak an untracked worker; the prime is ctx-cancelable so shutdown stays prompt).

[...]

PR #5311, opened 2026-08-16
fix(dashboard): avoid eager full-history run-tailers at startup

atbrace:
## Summary

Keep an idle dashboard supervisor idle: `Plane.Start` no longer eagerly warms every city's run tailer and full event history. Summary/detail/SSE demand still starts one tailer per city lazily, preserving the existing API projections and sessions-cache behavior.

## Root cause

`Plane.Start` called `eagerWarmTailers`, which enumerated every city and started a `cityRunTailer` for each one. Each tailer immediately ran `runproj.Projector.ColdLoad` over the city's full event history, then rebuilt its projection. An idle supervisor therefore paid the full-history replay cost for every city even when no dashboard request existed; offline measurement was about 28.9% core with no dashboard traffic.

## Fix

- Remove eager city enumeration and the now-unused `CityRef`/enumeration plumbing.
- Preserve the existing lazy ensure path: the first summary/detail/SSE demand creates and starts exactly one tailer for that city.
- Make manager enable launch any tailer demanded before `Plane.Start`, exactly once under the existing manager lock, so pre-Start demand is not stranded.
- This change does not claim memory reduction, incremental projection, or last-subscriber teardown. The current subscriber map covers detail SSE while summary/detail GETs remain stateless.

Adjacent work: [#3939](https://github.com/gastownhall/gascity/pull/3939), [#4277](https://github.com/gastownhall/gascity/pull/4277), and [#4283](https://github.com/gastownhall/gascity/pull/4283). This is separate from the `gcy-sjn`/archive-reader work: it changes when the existing run-tailers start, not archive-reading semantics.

## Tests

- The old eager-start regression is red on unmodified `main` and green with this change.
- The pre-Start demand lifecycle regression is red before the manager-enable fix and green after it.
- The slow cold-load regression blocks the first replay, verifies HTTP 200 with `lanesPartial=true` and no false-ready lane, then releases the replay and verifies a ready lane on retry.
- `internal/api/dashboardbff` package tests, race, vet, and integration-tag compile checks ran through Nomad `crun`.

## Validation

Offline real-history validation replayed through sequence 602,296 / 25,875 beads: startup left zero city tailers, first demand triggered the replay, readiness arrived at about 21s, and the first response reported `lanesPartial` until retry. Three historical payloads were empty, and legacy archive normalization remains a data/archive caveat rather than a behavior change in this patch.


atbrace:
Sorry, agent yoinked a bunch of unrelated commits from my local build into this PR - reconciling now

quad341 (approved):
@atbrace — the part I appreciate most here isn't ripping out `eagerWarmTailers`; it's that you caught the demand-that-arrives-before-`Start` case and fixed it properly, hoisting the launch into `startLocked` and having `enable()` sweep already-created tailers. A quicker patch would have shipped a permanently stranded tailer. And `TestFirstDemandColdLoadTimeoutRetriesAfterReady` asserting the real 200/`lanesPartial=true` → ready-on-retry contract — threading `lanesPartial` through `runSummaryWire` to do it — is the kind of test that actually pins new behavior instead of just observing it.

Nothing for you to do. It's conflicting only because `main` migrated this package's test waits to a new hang-budget const while your branch was open, and that landed in the very file you delete. I'll take the rebase, keep the delete, move your lazy tests onto the new const, and re-run build/vet/`-race`. Build, vet and the package suite already pass on your head commit.

Two things I'd like your read on, after merge is fine:

- `/v0/.../runs/census` and `/runs/{id}/steps` hard-503 while `!ready`, so post-restart the first caller now eats the ~21s replay instead of it having elapsed at boot. Deliberate, I assume — worth documenting?
- `startLocked` can still `wg.Add` after `Stop` began waiting. Pre-existing, but eager warm was hiding it. Filing a follow-up; a `ctx.Err()` guard, or something better?

Merging after the rebase — credit stays yours.


---

Issue #4579, opened 2026-07-23
gc status: partial-status summary miscounts unknown agents as not-running (residual of #4343), and long agent names overflow into the status column


jacobhausler:
## Summary

`gc status` has two rendering defects that surface only when the runtime
probe times out (partial status). Upstream #4345 ("Render partial status
rows as unknown", merged 2026-07-16) fixed the per-row rendering — rows
now correctly print `unknown  (partial status)` instead of `stopped` — and
#4343 ("Partial status can render runtime-probe timeouts as stopped") was
closed on that basis. But the SUMMARY LINE was not fixed, so the same
misreport #4343 described still reaches the reader, just one line down.

Verified against gc built from:

    revision: 8bdc1e33c0adc5cb8b64c9df22646a7569f38ed4
    vcs.time: 2026-07-21T12:10:20Z
    vcs.modified: false
    mod:      github.com/gastownhall/gascity v1.1.1-0.20260721121020-8bdc1e33c0ad

## Defect 1 — summary line contradicts the row detail it sits above

A single `gc status` run prints, about thirty lines apart:

    1/18 agents running
    Sessions: 13 active, 0 suspended

while 15 of the 18 agent rows render `unknown  (partial status)`. The
agent-count summary folds `unknown` into "not running"; the sessions
counter comes from an independent source (the API) that answered
correctly. Ground truth via `curl -s http://127.0.0.1:8372/v0/city/<city>/sessions`
confirms the sessions line, not the agents line.

A reader skimming `1/18 agents running` concludes the fleet is down —
exactly the misreport #4343 was closed for, just moved from the row to
the summary.

Expected: during partial status, the summary should report unknown
separately (e.g. `1 running, 15 unknown, 2 stopped of 18`) or omit a
running/total ratio entirely, never silently count `unknown` as
not-running.

## Defect 2 — name column overflow eats the column separator

Same run, verbatim:

    core.control-dispatcher unknown  (partial status)
    tar-valon/core.control-dispatcherunknown  (partial status)

The rig-qualified agent name (`tar-valon/core.control-dispatcher`)
exceeds the fixed pad width and runs directly into the status token with
no separating space — looks like a fixed-width pad with no enforced
minimum gutter.

## Repro

Needs a city where the runtime probe times out (reliable for us today,
tracked separately as #4418 — the events.jsonl double-scan that causes
the timeout; fixing #4418 would hide these two by removing the timeout,
not fix them):

    gc status > /tmp/s.txt
[...]

PR #4345, opened 2026-07-16
Render partial status rows as unknown

keithballinger:
## Summary

Fix partial/degraded runtime status reporting so timeout/error fallback rows are rendered as unknown/partial instead of authoritative `stopped`.

This PR carries partial status through the bounded runtime status provider, CLI snapshot, JSON/API status view, and text rendering paths. When runtime probing is incomplete, non-running rows render as:

```text
unknown  (partial status)
```

rather than `stopped`.

Fixes #4343.

## Hedge / provenance

The upstream applicability claim started as a code read, not a clean upstream reproduction: the same bounded timeout/fallback and `agentStatusLine` false-as-`stopped` path exists on `gastownhall/main`.

The live reproduction and A/B validation were done on our downstream fork/deployment, not on a clean upstream deployment.

## Validation

On this upstream-based branch:

- `git diff --check`
- `CGO_ENABLED=0 go test ./internal/api -run 'TestHandleStatusMarksRuntimeProbePartial|TestHandleStatusPreservesPartialWorkCountSurvivors|TestStatus' -count=1`
- `CGO_ENABLED=0 go test ./internal/api -count=1`
- `CGO_ENABLED=0 go test ./internal/runtime/tmux -run TestStateCache -count=1`
- `CGO_ENABLED=0 go test -c ./cmd/gc`

Downstream live A/B on the same degraded host, same moment, only binary changed:

- unpatched: `30 running · 19 stopped`
- patched: `30 running · 19 unknown`

The patched status JSON also carried:

```json
{
  "partial": true,
  "partial_errors": ["runtime status probe incomplete; non-running agent rows are unknown"]
}
```

## Boundary

This is intentionally not a perf fix and not a root-cause claim for the runtime probe degradation. It only prevents partial/timeout fallback from becoming a confident false `stopped` report.

PR #5320, opened 2026-08-16
fix(status): stop publishing unobserved agents as absent in JSON and stdout

austinborn:
## Summary

`gc status` was reporting a live 15-agent factory as dead. The runtime probe timed out, so the command hadn't learned a thing about any agent, and it printed a full table of `stopped` rows anyway, footed by `0/15 agents running`. The caveat went to stderr, where it scrolled above a table that looks complete, so in practice you'd never see it.

Two merged changes fixed that for the text table: [#4345](https://github.com/gastownhall/gascity/pull/4345) renders partial rows as `unknown  (partial status)`, and [#4579](https://github.com/gastownhall/gascity/pull/4579) replaced the contradictory ratio with `0 running, 15 unknown of 15 agents`. This finishes the job on the two surfaces they didn't reach.

**The JSON health block still derived absence from a probe that didn't observe anything.** [`city_status_snapshot.go`](https://github.com/gastownhall/gascity/blob/234d66b01e5fc56c9aec1fc3795bebca3c28e4de/cmd/gc/city_status_snapshot.go#L463-L465) appended `no_agents_running` whenever `RunningAgents == 0`, with no check on whether the probe had answered. It hadn't. During partial status that count is zero because nothing was observed, so every dashboard and health check reading `gc status --json` was told the city was dead. That's the surface most likely to be acted on without a human in the loop, and the natural response to "no agents running" against a queue of open work is to restart the agents and re-sling the work, which duplicates builds that are already mid-review.

**The reason a count was degraded never reached stdout.** `PartialErrors` was populated and handed to the JSON payload, but the text renderer didn't print it, so a reader piping stdout got the number while its explanation sat on stderr.

```mermaid
flowchart TD
    P[Runtime probe times out] --> C["RunningAgents = 0<br/>nothing was observed"]
    C --> T[Text table]
    C --> J[JSON health block]
    C --> S[stdout reason line]
    T --> T1["'unknown (partial status)'<br/>fixed in #4345 / #4579"]
    J --> J1["'no_agents_running'<br/>reads as a dead city"]
    S --> S1["nothing printed<br/>reason sat on stderr"]
    J1 --> F[This change]
    S1 --> F
    F --> F1["'agent_state_unknown' + unknown_agents"]
    F --> F2["reason printed beside the count"]
```

## Changes

- Gate the health signal on the partial flag: emit `agent_state_unknown` when agent state wasn't observed, and keep `no_agents_running` for the case where the probe did answer and found nothing running.
- Add `unknown_agents` to the summary so the payload says how much of its own count isn't a measurement, rather than making a consumer cross-reference `partial`.
- Print the partial reason on stdout directly beneath the agent count.
- Add `unknown_agents` to `schemas/status/result.schema.json`. That summary object is `additionalProperties: false`, and the schema ships embedded and served rather than as documentation, so without the entry the new field would fail validation. The `omitempty` is what makes it sharp: only the degraded payload carries the field, so a consumer validating `gc status --json` would break at exactly the moment the city is degraded.
- Move the "how many are unknown" rule into `unknownAgentCount`, which `agentSummaryLine`, the summary and the health signals now share. Three surfaces were each deriving the same quantity on their own, which is how they'd come to disagree.

Non-partial output is untouched. An observed-empty city still renders `0/15 agents running` and still signals `no_agents_running`.

## Interaction with #4918

Draft [#4918](https://github.com/gastownhall/gascity/pull/4918) raises the 50 ms per-probe budget and tracks partial per agent row rather than city-wide. That's the other half of this problem, and the two changes compose: once partial is per-row, `unknownAgentCount` receives a smaller unknown count and the health signal narrows with it, no edit needed here. The hunks don't overlap.

## One thing deliberately left out

The originating report suggested a non-zero exit code on the degraded path. This change doesn't take it, for two reasons. `gc status --json` already exits 1 when the *session snapshot* degrades, and that exit is documented as a contract monitoring callers rely on, so extending it to probe timeouts isn't a new idea. But at today's 50 ms budget a partial probe is routine rather than exceptional, so making it non-zero now would turn a common condition into a failing command for every script and gate that runs `gc status`. You don't want that landing ahead of the latency fix. The degradation is machine-detectable without it: `partial`, `partial_errors`, `unknown_agents`, and a `health.degraded` whose signal now names the real condition. Worth revisiting once #4918 lands and partial means what it says.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./cmd/gc/...`
- [x] `go test ./cmd/gc/...` — full package, no `-run` filter, green in 504 s
- [x] Degraded path reproduced by injection rather than by luck: the tests drive the real bounded provider with a probe slow enough to blow its own budget, then read the partial flag back through `statusProviderPartial`, the same call the collector makes. A build where the timeout stopped marking status partial won't pass them.
- [x] Both halves mutation-checked. Reverting the health-signal guard fails `TestPartialStatusDoesNotSignalNoAgentsRunning` and nothing else; removing the stdout reason line fails `TestTimedOutProbeNeverRendersAbsenceAsFact` and nothing else.
- [x] Positive controls, so the fix can't pass by printing nothing: a healthy probe still prints `1/3 agents running`, still marks the two genuinely stopped agents `stopped`, and an observed-empty city still signals `no_agents_running`.

Assertions about agent state are scoped to the rows of the Agents block. The `Controller: stopped` line spends the same word on a different subject, and an unscoped search for it both fails the honesty check spuriously and lets the positive control pass without a single agent row in it.

Two environment notes for anyone re-running the package. First, `TestOrderExecEnvReservedKeysCoverProjectedEnv` fails with `missing from reserved guard: [GH_TOKEN GITHUB_TOKEN]` on any box that exports those two variables. It's unrelated to this change and reproduces identically on unmodified `main` at [`83cad49bc`](https://github.com/gastownhall/gascity/commit/83cad49bc); unset them and it passes. That's the only failure that reproduced for me.

Second, a reviewer running the package on a loaded box also saw `TestCityRuntimeForceShutdownTearsDownAfterLateAsyncSweep` fail with "force shutdown missed the late async-started runtime". I couldn't reproduce it: it passed 8/8 in isolation and every full-package run here was green. That points at a timing race that surfaces under load rather than anything this branch touches, though I'm reporting it as unreproduced rather than diagnosed.

---
> • On behalf of: @austinborn


austinborn:
Round 2, from review. One code change and one merge.

`cityStatusJSONFromSnapshot` was reading the agent counts off `snapshot.Summary` while publishing the `summary` parameter, so the new `unknown_agents` rested on a different source than the `running_agents` and `total_agents` printed next to it. Nothing was wrong at runtime, since both production callers pass `snapshot.Summary`, and the `no_agents_running` case had always read the same way. But putting a newly published field on that split is worth closing: a caller passing anything else would emit an unknown count that contradicts the counts beside it, which is the self-contradicting output this change set exists to remove. Both reads now come from the `summary` parameter and the `no_agents_running` case moved with them, so the switch and the published counts share one source. A test passes a deliberately disagreeing summary to pin it, and reverting the derivation fails that test alone.

Also merged current `main` in and re-ran the full `cmd/gc` package: green in 499 s, no failures.

---
> • On behalf of: @austinborn

quad341 (approved):
@austinborn — merging this as-is. Thank you for finishing what #4345 and #4579 started. The JSON health block was the dangerous half: it's the surface something acts on without a human in the loop, and "no agents running" against a queue of open work is an invitation to restart and re-sling builds that are already mid-review.

The tests are why this was easy to say yes to. You reproduce the degraded path by driving the real bounded provider until it blows its own budget and reading `partial` back through `statusProviderPartial` — the same call the collector makes — instead of just setting the flag. The positive controls mean the fix can't pass by printing nothing. And scoping the row assertions past the `Controller: stopped` line is exactly the trap I expected to find sprung.

One thing that's ours, not yours: on the supervisor API route, `partial` is an aggregate — it also covers mail, work-count and store-health failures, so a non-agent failure can now read as `agent_state_unknown`. Your code inherits that from `agentSummaryLine`; it isn't something you introduced. I'll open a follow-up to split agent-probe degradation from the aggregate flag, which pairs naturally with #4918.

Leaving the exit code alone was the right call.


---

PR #5322, opened 2026-08-16
fix(order): validate runtime vars against the vars the caller supplied

remuscazacu:
## The bug

Both order paths thread `vars` into recipe compilation and then validate the compiled recipe with an empty `molecule.Options{}`:

```go
recipe, err := prepareOrderWispRecipe(ctx, store, a, searchPaths, vars)
molecule.ValidateRecipeRuntimeVars(recipe, molecule.Options{})   // vars dropped
```

`ValidateRecipeRuntimeVars` reads `opts.Vars` (`runtimeValidationVars` → `CollectVarValidationErrors`), so validation runs against nil and reports **every** `required = true` var as missing regardless of what was passed. Any formula with a required var is unfireable as an order:

```
$ gc order run kb-backfill --var since=2026-02-08 --var until=2026-08-08 \
    --var sort=newest --var cursor_bead=none
gc order run: variable validation failed:
  - variable "cursor_bead" is required
  - variable "since" is required
  - variable "sort" is required
  - variable "until" is required
```

All four were supplied.

Of the 14 `ValidateRecipeRuntimeVars` call sites, exactly these two passed an empty `Options`. The other 12 — including `gc formula cook` (`cmd_formula.go:676/838`) — already pass `Vars`, which is why `cook` succeeds on the same formula the order rejects.

**The controller half matters more than the manual one.** `cmd_order.go` fails loudly at an operator's terminal. `order_dispatch.go` is the controller's automatic dispatch, so a `cooldown` or `cron` order using a required var fails on **every tick, unattended**, with nobody reading the `order.failed` events.

It stayed hidden because required-with-no-default is rare: 0 of the 594 order files across the packs I have installed use one, and formulas conventionally give every var a default. The formula that exposed it deliberately makes two vars required precisely *because* its two callers need opposite values, so any default would be silently wrong for one of them.

## The fix

`molecule.Options{}` → `molecule.Options{Vars: vars}` at both sites.

## Tests

`TestOrderRunAcceptsSuppliedRequiredVar` and `TestDispatchWispAcceptsSuppliedRequiredVar` drive the **real entrypoints** (`doOrderRunWithJSON`, `dispatchOne`) rather than calling the validator directly — a test that invoked `ValidateRecipeRuntimeVars` itself would pass with the bug still in place, since the defect is in what the call site passes.

The test formula puts the required var in a step **description**, not a compile-time `range`. That distinction is load-bearing: a range var is resolved during compilation and fails there, before the runtime validator is reached — which is exactly why the existing `TestPrepareOrderWispRecipeThreadsVarsToFormula` passes today and cannot see this bug.

Negative-controlled: with the fix reverted at both sites, both new tests fail with the production error verbatim (`variable "target" is required`). `TestDispatchOneRefusesMissingRequiredParam` still passes, and the new tests also assert that **omitting** the var is still refused — the fix must not turn `required` off.

## Note

`go test ./cmd/gc/` reports a pre-existing dolt sql-server leak-guard failure from `TestInitFromWithoutHostedPreservesTemplate`, unrelated to this change — verified it reproduces identically on a clean tree with these changes stashed.

### DIFF
--- a/cmd/gc/cmd_order.go
+++ b/cmd/gc/cmd_order.go
@@ -813,7 +813,11 @@ func doOrderRunWithJSON(aa []orders.Order, name, rig, cityPath string, store bea
 		fmt.Fprintf(stderr, "gc order run: %v\n", err) //nolint:errcheck // best-effort stderr
 		return 1
 	}
-	if err := molecule.ValidateRecipeRuntimeVars(recipe, molecule.Options{}); err != nil {
+	// Validate against the vars the caller supplied. Passing an empty Options here
+	// drops them, and ValidateRecipeRuntimeVars reads opts.Vars — so every
+	// `required = true` var reports as missing however many --var flags were given,
+	// making any formula with a required var unfireable as an order.
+	if err := molecule.ValidateRecipeRuntimeVars(recipe, molecule.Options{Vars: vars}); err != nil {
 		fmt.Fprintf(stderr, "gc order run: %v\n", err) //nolint:errcheck // best-effort stderr
 		return 1
 	}
--- a/cmd/gc/order_args_channel_test.go
+++ b/cmd/gc/order_args_channel_test.go
@@ -187,3 +187,125 @@ func TestDispatchOneRefusesMissingRequiredParam(t *testing.T) {
 		t.Fatalf("order.failed message = %q, want it to name the missing required param repo", failMsg)
 	}
 }
+
+// requiredVarFormulaDir writes a formula whose only var is `required = true`
+// with NO default, consumed by a step description rather than a compile-time
+// `range`. That distinction is the point: a range var is resolved during
+// COMPILATION, so it fails before the runtime-var validator is ever reached
+// (TestPrepareOrderWispRecipeThreadsVarsToFormula above covers that path and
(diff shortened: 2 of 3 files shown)


quad341 (approved):
@remuscazacu — this is a genuinely excellent bug report, and we're merging it as-is.

The diff is two lines, but the work that makes it trustworthy is everywhere else. You found a defect that only surfaces with a rare formula shape, you correctly identified that the controller half is the dangerous one — an unattended cooldown order failing silently on every tick is so much worse than a loud terminal error — and you audited all 14 call sites instead of patching the one that bit you. I re-ran that audit here and it holds: after your fix, nothing in the tree still validates against a bare `Options{}`.

The tests are what I want to call out. Driving `doOrderRunWithJSON` and `dispatchOne` rather than the validator, because a direct-validator test would pass with the bug still in place. Putting the required var in a step *description* so it survives compilation and actually reaches the runtime validator. And asserting that omitting the var is still refused, so the fix can't quietly turn `required` off — that last one is the assertion most people skip, and it's the one that matters.

Thanks also for flagging the pre-existing dolt leak-guard failure as unrelated and verifying it on a clean tree. That saved us the detour.


---

PR #5323, opened 2026-08-16
fix(status): carry the store-health measured-or-unknown flag across the API path

austinborn:
## Summary

`storehealth.Health` carries `RowsMeasured` so a row count that failed or timed out stays distinguishable from a store that really is empty. `cmd/gc.StoreHealth` re-expresses it as `LiveRowsUnknown`, and [`renderStoreHealthBlock`](https://github.com/gastownhall/gascity/blob/ab54f93c3/cmd/gc/store_health.go#L114-L140) honours it: the count renders as `unknown (count unavailable)` and the ratio line is dropped rather than printed as a misleading `0.0 MB/row`.

`gc status` reaches that renderer by two paths, and only one of them was carrying the qualifier.

```mermaid
flowchart TD
    D["storehealth.Compute<br/>RowsMeasured: the source of truth"] --> L["storeHealthFromInputs<br/>local path"]
    D --> S["statusStoreHealthFromDomain<br/>supervisor path"]
    L -->|"sets LiveRowsUnknown"| R["renderStoreHealthBlock"]
    S --> W["wire type StatusStoreHealth<br/>had no field for the flag"]
    W --> V["StatusStoreHealthView<br/>had no field either"]
    V --> M["snapshotFromStatusView<br/>copies the fields that exist"]
    M -->|"LiveRowsUnknown zero-fills to false"| R
    R --> O["Live rows: 0<br/>Ratio: 0.0 MB/row"]
```

On the supervisor-managed path the struct is rebuilt in [`snapshotFromStatusView`](https://github.com/gastownhall/gascity/blob/ab54f93c3/cmd/gc/cmd_citystatus.go#L327-L338) from the server's view type. Neither [`api.StatusStoreHealthView`](https://github.com/gastownhall/gascity/blob/ab54f93c3/internal/api/types_read.go#L112-L121) nor the wire type [`api.StatusStoreHealth`](https://github.com/gastownhall/gascity/blob/ab54f93c3/internal/api/huma_types_patches.go#L262-L271) had anywhere to read the qualifier from, so the mapper copied the eight fields that existed and Go zero-filled `LiveRowsUnknown` to `false`. The renderer was then told the count was real, and printed it along with a ratio and a threshold comparison derived from it.

Nothing in the mapper says `measured: true`. Something has to say it for the bug to appear, and nothing does — which is why a grep for the flag comes back clean at every site it appears, and why the defect sits at the one place the identifier is absent.

## The change

`RowsMeasured` is added to the wire type and the view type, populated on the server from the domain value, copied through the decoder, and inverted into `LiveRowsUnknown` in the mapper. The OpenAPI spec, the generated Go client and the dashboard's typed client are regenerated from it.

**The flag is spelled for the measured state on purpose.** A payload that omits `rows_measured` decodes to `false` and renders as unknown, so an older supervisor's silence withholds the number instead of asserting a measurement. Spelled the other way round, as an `unknown` flag, a dropped field means "measured" and every omission makes a confident claim on the caller's behalf. That polarity is what let the original gap survive review. The visible cost is that a `gc` binary newer than the supervisor it is pointed at will report the count as unknown until the supervisor restarts, which is the honest answer for a server that can't say whether it measured.

**The flag is carried, never inferred.** Deriving it at the mapper from `live_rows == 0` is the tempting one-line version, and it reintroduces the same conflation from the other side by reporting every genuinely empty store as unmeasured. The two states aren't recoverable from the number, which is the entire reason the flag exists.

The server's error path is deliberately untouched: when the count fails, `computeStoreHealth` still returns an error and the whole block is omitted from the response. This change gives the transport the vocabulary to express a present block with an unknown count, and doesn't yet change when the server chooses to send one.

## Test plan

New tests drive the API path by injecting a view directly, so the degraded case is reproduced rather than waited for.

- `cmd/gc`: an unmeasured count renders `unknown` with no ratio line; a measured empty store keeps its definite `0` and its ratio; a populated store renders its real ratio; the unmeasured and measured-empty renderings are asserted to differ; the mapper is checked in both directions, including an unmeasured count whose number happens to be non-zero; `gc status --json` sets `live_rows_unknown` on the API path and omits it for a real count.
- `internal/api`: the domain-to-wire adapter carries the flag both ways and doesn't infer it from the count; `rows_measured` still serializes when false; the decoder carries it; a payload with no `rows_measured` at all decodes as unmeasured.

The dashboard SPA contributes the one non-Go line in the diff. `baseStatus()` in `Health.test.tsx` is annotated `: StatusBody`, so its `store_health` literal is type-checked against the regenerated `StatusStoreHealth`, and the fixture describes a store whose 2000 rows were counted successfully. It gets `rows_measured: true`. The SPA's own rendering is left alone deliberately: it reads `ratio_mb_per_row` as a definite value and derives its label from `warning`, which is this same defect family one consumer over, and it is unreachable today because the server returns an error rather than an unmeasured block. Widening the PR into it would have mixed a rendering change into a transport one.

Both mutations were checked against the suite. Dropping the flag from the mapper — the original defect — turns four tests red. Replacing it with the `live_rows == 0` inference turns the measured-empty positive control red, which is the test that exists to make that shortcut fail.

`go test ./cmd/gc/... ./internal/api/... ./internal/storehealth/... ./internal/beads/...` passes. One unrelated failure, `TestOrderExecEnvReservedKeysCoverProjectedEnv`, reproduces on an unmodified tree whenever `GH_TOKEN` and `GITHUB_TOKEN` are present in the environment, and passes with them unset.

## A separate defect this does not fix

Worth flagging for anyone reading this because they saw `Live rows: 0` on a live city: on the city where this was found, that output is not produced by the dropped flag.

The status API there reports `live_rows: 0` with **no** store-health entry in `partial_errors`, against a store `bd stats` measures at 1288 beads. Since `computeStoreHealth` returns an error on every count failure and the handler then omits the block entirely, a block that is present and error-free means the count succeeded and answered zero. The supervisor is certifying a wrong number honestly, one layer below the transport.

That is tracked separately, in `internal/api`'s `countBeadStoreRows` and the `NativeDoltStore` counter beneath it. This PR is the transport half: it makes an unmeasured count expressible and correctly rendered, and on its own it won't change what such a city prints.

---
> • On behalf of: @austinborn


austinborn:
Merged upstream `main` into this branch to clear the conflict that showed up while the PR sat in the review queue. This is a merge rather than a rebase, so the branch's own commits keep their shas and anything anchored to them stays put.

Every conflict was a rename/rename collision under `internal/api/dashboardspa/dist`, the committed SPA bundle. Both sides rebuilt it, so the content hashes moved on both and git saw 22 chunks renamed two different ways. Picking a side would have been wrong either way: `main`'s bundle predates the `rows_measured` field this branch adds to the generated client, and this branch's bundle predates the context-window change that landed on `main`. So the resolution was to rebuild the bundle from the merged source, exactly as `make dashboard-build` does, and commit that.

Two checks that the rebuild is the one CI expects. Building `main`'s source with the same toolchain reproduces `main`'s committed `dist/` byte-for-byte, which says the build is deterministic here and the local Node matches the version CI pins. And the merged bundle differs from `main`'s only by 18 bytes in the entry chunk, the size of the `rows_measured` addition, plus the filename-hash cascade that one change sets off across the other chunks. `make dashboard-ci`, the target whose whole job is failing on a stale committed bundle, passes locally against the merge.

No source file conflicted, and no source file was hand-edited during the resolution.

---
> • On behalf of: @austinborn

quad341:
Thanks for this, @austinborn — there is clearly a lot of work here.

austinborn:
@quad341 — answering your [Aug 30 note](https://github.com/gastownhall/gascity/pull/5323#issuecomment-5470249823) properly, and sorry it sat for twelve days. You asked a direct question and deserved an answer sooner than this.

**The size objection doesn't apply any more, so neither branch of your offer is quite the right one.** You asked us either to split it or to argue it can't be divided. The honest answer is a third thing: the diff you were served isn't the diff that's there now. GitHub currently reports this PR as **442 additions and 65 deletions across 38 files, 507 changed lines in total**, roughly 2.5% of the 20,000-line ceiling you hit. Largest single file is a 171-line test.

What inflated it was the committed SPA bundle under `internal/api/dashboardspa/dist`. It rehashes every chunk filename on each rebuild, so any branch that rebuilds it collides with any other branch that does, and the whole bundle reads as changed.

Of the 38 files here, 23 are that bundle, and all 23 are byte-identical in size to main's. Only their content-hash names and import specifiers differ. [`.gitattributes`](https://github.com/gastownhall/gascity/blob/9b5f893d1e772b6ab7e93b910d320c0feb7fe931/.gitattributes#L6) already marks that path `-diff linguist-generated`, so they ought to collapse rather than render for you. That leaves **15 source files** to actually read.

If you'd still rather see it split at its real size, say so and I'll do it. I'd push back gently, though: 15 source files carved into "a few smaller PRs" probably costs you more context-switching than it saves, and the pieces aren't really independent, since the API type, the decode path, the status renderer and their tests are one change to one flag.

**Also worth your time, and nothing to do with this PR.** `main` is broken right now at [`3d268c5`](https://github.com/gastownhall/gascity/commit/3d268c5849ca2ef49392c6e4adb68e9a10d0b59d), with 31 failing checks including `CI / required`. `cmd/gc` doesn't compile its own test package: `releaseOrphanedPoolAssignments` gained a parameter in [#6261](https://github.com/gastownhall/gascity/pull/6261), while six call sites in [`pool_session_name_route_assignee_liveness_test.go`](https://github.com/gastownhall/gascity/blob/3d268c5849ca2ef49392c6e4adb68e9a10d0b59d/cmd/gc/pool_session_name_route_assignee_liveness_test.go#L62) still pass the old signature. Both PRs were green on their own; it's only the combination that breaks. And because CI builds each PR merged into main's tip, any PR whose checks re-run while main sits in this state goes red for that reason, this one included. (A PR that hasn't re-run since still shows its older green, so the repo isn't uniformly red.) Nothing on our side causes it and nothing on our side can clear it.

This branch is rebased onto current `main` and reads `MERGEABLE` again.

The bundle was rebuilt from the merged source on node 22.14.0 rather than hand-merged, and I checked it in both directions: `rows_measured` from this branch, plus `wake_request` and `drain_fence_unavailable` from main, are all present in the rebuild. So neither side got quietly taken wholesale.

---
> • On behalf of: @austinborn

quad341 (approved):
@austinborn — thank you for this, and specifically for the two tests that make the lazy version of this fix impossible. The measured-empty positive control and the unmeasured-count-that-happens-to-be-4200 case are what turn "a field got added" into a defect that can't quietly come back, and the hand-written legacy payload with no `rows_measured` at all is the detail most people would have skipped.

The polarity argument is the right call. Spelling the flag for the measured state, so that an omission withholds the number instead of asserting a measurement, is the correct default for a field whose whole job is to say "don't trust the number next to me" — and I appreciate that you wrote the cost down rather than burying it: a `gc` newer than its supervisor will report `unknown` until the supervisor restarts. That's the honest answer for a server that can't say whether it measured, and it's a bounded window.

Thank you also for leaving the SPA's own rendering alone and explaining why. Mixing a rendering fix into a transport fix would have made this much harder to review, and you were right that it's a separate change.

I traced the renderer, the single view producer and the regenerated client myself rather than taking the description's word for it — everything lines up. Merging as-is. Sorry it sat in the queue as long as it did.


---

PR #5325, opened 2026-08-16
fix(status): refuse a store-health row count of zero the store itself disproves

austinborn:
## Summary

`gc status` reported `Live rows: 0` on a city holding tens of thousands of beads, with no store-health entry in `partial_errors`. Only one thing produces that pair. [`computeStoreHealth`](https://github.com/gastownhall/gascity/blob/b36bcd6d86da37921947a951f989f059e2797e3e/internal/api/store_health.go#L76-L97) returns an error whenever the row count fails, and the status handler turns any such error into a partial error and drops the whole block. A block that's present with no matching partial error therefore means the count returned zero and reported success, which `computeStoreHealth` then certified as measured by passing a hardcoded `true` for `rowsMeasured`.

The zero is the denominator of the maintenance ratio, so the threshold comparison read as under-threshold by luck. A store that had genuinely crossed the threshold wouldn't have warned.

## Which branch answered

[`countBeadStoreRows`](https://github.com/gastownhall/gascity/blob/b36bcd6d86da37921947a951f989f059e2797e3e/internal/api/store_health.go#L135-L158) has two, and both can end at zero. Until this change nothing recorded which one had, so you couldn't attribute a zero in the field without first working out which store the city happened to have open at the time.

```mermaid
flowchart TD
    A["whole-ledger count<br/>AllowScan, IncludeClosed"] --> B{"store implements Counter?"}
    B -->|"yes"| C["SELECT COUNT over the backend"]
    B -->|"no"| E["hydrating list of the whole ledger"]
    C -->|"count unsupported"| E
    C --> D["a real count, zero included"]
    E --> F["length of the returned rows"]
    D --> G{"is the answer zero?"}
    F --> G
    G -->|"no"| H["measured"]
    G -->|"yes, and the store has already returned rows"| I["unavailable: block omitted, reason attached"]
    G -->|"yes, with nothing to disprove it"| H
```

The counter fast path runs only for a store implementing `beads.Counter`. `BdStore` doesn't implement `Count`, so on every bd-backed city the count falls through to the hydrating list instead. That's pinned as a test rather than left as a claim, because it's the reasoning the branch labels rest on.

The list branch is the interesting half. The store-health query is a whole-ledger scan that includes closed rows, and `IncludeClosed` isn't one of the fields [`HasFilter`](https://github.com/gastownhall/gascity/blob/b36bcd6d86da37921947a951f989f059e2797e3e/internal/beads/query.go#L216-L228) counts, which makes it exactly the shape [`unread_store_notice.go`](https://github.com/gastownhall/gascity/blob/b36bcd6d86da37921947a951f989f059e2797e3e/internal/beads/unread_store_notice.go#L383-L392) already diagnoses: the one List shape whose empty answer is a claim about the store rather than about a predicate. The store layer prints an operator notice and lets the read succeed, on purpose and at length; that file argues the case better than this one can. Refusing there would fail a merely idle city closed across `gc ready`, `gc rig add` and the federation, and that decision stands.

## What changed

The refusal belongs at the one consumer for which zero was never a usable answer. `storehealth.Compute` already declines to derive the ratio or the warning from a zero, so the denominator can't use one either way.

`beads.RowWitness` is a new optional capability in the style of `Counter` and `BatchDeleter`. It exposes the per-scope latch the unread-store notice already keeps, so a caller acting on the witness and an operator reading the notice can never be told different things about the same scope. `BdStore` answers from that latch and doesn't carry any new state of its own.

Three stores implement it, because a witness that covered only one of them would be inert on the others. `NativeDoltStore` latches on its own reads, which matters most: it's the store the counter fast path exists for, so without it the guard would skip the branch it was written for. `CachingStore` answers for the wrapper the API server actually holds, from its own cached rows first and from its backing store second. Every implementation latches from what the backend returned rather than from what survived client-side narrowing, and a zero never latches, so the read being judged can't certify itself.

Implementing the capability is not enough on its own. The API server never holds one of those stores directly: `beadPolicyStore` embeds `beads.Store` as an interface, Go promotes only that interface's method set, and the wrapper therefore hides whatever the store inside it can do. That wrapper is on the path every time, twice over, since the store opener wraps its result and the caching layer re-wraps the cache it builds. So `SawRows` is forwarded explicitly, the way this file already forwards `Count`, `DeleteBatch` and `ConditionalWritesResolveTarget`, with compile-time assertions on both wrapper types. Skip it and the guard still compiles, still runs, and certifies the zero exactly as before. It also fails quietly, because a missing optional capability reads to the caller as a store that cannot witness itself, and that is a supported state.

The count then refuses a zero the witness disproves and returns an error naming the branch that answered. That routes into machinery that already exists: the block is omitted and the reason is attached, rather than `Live rows: 0` and a ratio derived from a count nobody took. Nothing's refused at the store layer, and the status response still serves.

## What stays the same

A genuinely empty ledger still reports a healthy zero. The witness proves rows are present and never that they are absent, so it can reject a zero and never certify one, and a store with no witness keeps its prior behavior unchanged. The cost of that choice is real, and it's stated in the code and pinned by a test: a store pointed at the wrong database from process start has never returned a row, so the witness has nothing to say about it. That case is the unread-store notice's own territory, and `gc doctor` enumerates both databases for it.

This sits upstream of the measured-or-unknown flag being threaded across the API path in #5323 and composes with it. The error path here is the signal that survives today's transport, since the wire type can't carry the flag yet; the flag is what will let a future count report unknown while keeping the block.

## Test plan

`go test ./internal/beads/... ./internal/api/... ./cmd/gc/...` passes.

New coverage, and each one goes red without the change:

- The counter branch answering zero on a store that has already returned rows is an error naming that branch, and so is the list branch.
- A status body built over such a store omits the block, sets `partial`, and carries a `store health:` entry naming both the branch and what disproved the zero.
- An empty ledger, a store with no witness, and any non-zero count are all left exactly as they were.
- The witness latches from what bd returned rather than from what survived client-side filtering, stays silent on a ledger that has only answered empty, and is shared across the throwaway stores `cmd/gc` rebuilds per read.
- `BdStore` implements no `Counter` and does implement `RowWitness`, which is the evidence behind the branch labels rather than an inspection of them.
- The policy wrapper the API server holds forwards the witness instead of hiding it, and the same wrapper over a store that cannot witness itself reports no evidence rather than failing the read.

---
> • On behalf of: @austinborn


austinborn:
## Review round: the guard was inert on every real city

Review found one defect that mattered, and it went to the heart of this change: the guard wasn't doing anything on any city the controller opens.

`beadPolicyStore` embeds `beads.Store` as an interface. Go promotes only that interface's method set, so the wrapper didn't satisfy `beads.RowWitness` even with a witness-capable store inside it. The API server reaches its city store through that wrapper every time, twice over, because the store opener wraps its result and the caching layer re-wraps the cache it builds. So `measuredStoreRows`' type assertion always failed, and the zero this PR exists to refuse was certified measured exactly as before.

The failure was silent, which is the part worth remembering for the next optional capability added here. A missing one reads to the caller as a store that can't witness itself, and that's a supported state the guard passes through on purpose. An inert guard and a working one looked identical.

`SawRows` is now forwarded through the wrapper in the style this file already uses for `Count`, `DeleteBatch` and `ConditionalWritesResolveTarget`, with compile-time assertions on both wrapper types. Two tests came with it. `TestBeadPolicyStorePreservesRowWitness` fails against the tree as it stood and passes with the forwarder, so it pins the defect rather than restating the fix; the second covers an inner store that can't witness itself, which has to report no evidence rather than fail the read.

Also in this round: `misspell` rejected the British spelling of "behavior" in five comments across the changed files, and that's what turned the static-check job red. All five are corrected.

The updated description above covers the forwarder under "What changed".

---
> Comment posted by the operator's software factory.
> • On behalf of: @austinborn

austinborn:
Checking in on this one, which has been open 26 days without a review. No changes requested, no conflict, and it still reads `MERGEABLE`, so there's nothing outstanding from our side and I've deliberately left the branch untouched rather than push a no-op merge at it.

Worth saying what it does, since the description undersells the consequence. `gc status` currently prints a store-health row count of zero against stores that plainly hold rows, and that zero is the denominator of the maintenance ratio. So the ratio renders as `0.0 MB/row` against its threshold and reads as a confident all-clear that measured nothing. It's a reporting defect rather than an outage, and the maintenance trigger itself is on a separate path that works, but anyone reading that output gets told everything is fine by a number that can't say otherwise.

We're running that defect live on our own city, which is how we noticed.

One heads-up on CI. `main` is broken at [`3d268c5`](https://github.com/gastownhall/gascity/commit/3d268c5849ca2ef49392c6e4adb68e9a10d0b59d), 31 failing checks including `CI / required`, because `releaseOrphanedPoolAssignments` gained a parameter in [#6261](https://github.com/gastownhall/gascity/pull/6261) while six call sites in `cmd/gc/pool_session_name_route_assignee_liveness_test.go` still pass the old signature. CI builds each PR merged into main's tip, so any PR whose checks re-run in this window goes red on it regardless of its own contents. This PR's checks last ran before that landed, which is why it still reads green; a re-run against the current tip would go red until main is fixed.

---
> • On behalf of: @austinborn

quad341:
Merged — thank you, @austinborn, and sorry this waited so long. It was held only because it adds a `beads.RowWitness` capability to the store contract, which needed a maintainer call rather than an automated one; that call is yes, as-is. Before merging I re-verified it on current `main` (clean merge, `go build ./...`, `go vet`, and all 15 of the new store-health / row-witness tests). A zero row count that the store itself contradicts is exactly the kind of quiet lie `status` should refuse — thanks for catching it.

PR #4111, opened 2026-07-09
fix(order-exec): project GH_TOKEN/GITHUB_TOKEN into exec-order env

julianknutsen:
## What

Exec orders shell out to `gh` (e.g. merge orders via the workflows pack), which authenticates from `GH_TOKEN` (preferred) or `GITHUB_TOKEN`. Both keys contain the substring `TOKEN`, so `execenv.IsSensitiveKey` reports them sensitive and the curated order-exec env — built from a map, then merged through `FilterInherited` — never carries the controller's ambient token into the child process. Every `gh` call an order runs then fails auth even though the controller holds a valid token.

## How

Mirror the ambient token into the exec-order env map in `orderExecEnvWithError`, the single choke point both `gc order run` and the control dispatcher pass through, the same way `mirrorBeadsDoltEnv` carries the hosted-gateway credential command. This does **not** weaken `execenv.IsSensitiveKey`, so `RedactText` still masks the value in captured exec output and logs. Projected before the `[order.env]` loop so an order can still scope its own `GH_TOKEN`, and only when a non-empty value exists in the ambient controller env.

## Tests

- `TestProjectGitHubTokenExecEnv` (bd_env_test.go) — unit coverage of the mirror helper: ambient projection, order.env override wins, absent-when-unset, single-present-token.
- `TestOrderExecEnvProjectsGitHubToken`, `TestOrderExecEnvGitHubTokenOrderEnvOverrideWins` (order_dispatch_test.go) — end-to-end through `orderExecEnvWithError`.

Verified locally: `go vet ./cmd/gc` clean; the new tests plus the existing `TestOrderExecEnvAppliesOrderEnvOverrides` / `TestMirrorBeadsDoltEnv*` sweeps pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2809, opened 2026-05-30
order.env per-order env overrides + JSONL spike floor (clean subset of #2332)

quad341:
Re-homed from #2332 by @eric-jones — thank you! This carries the clean, wanted pieces of that PR forward.

## Why this is a subset

#2332 bundled several alert-hygiene concerns. Since it was opened, the world moved underneath it in two good ways:

- The **dolt-health zombie-scan rework** was split out, reworked, and already merged as #2734 — so that portion is in `main` and is intentionally **not** carried here.
- The **deacon prompt** change belongs with the `gastown` pack, which migrated out of this repo into [`gastownhall/gascity-packs`](https://github.com/gastownhall/gascity-packs) (#27). That one-line hunk is being handled there, in its new canonical home.

What's left are two self-contained, reviewed-and-wanted changes, lifted cleanly from #2332 with @eric-jones's blessing to cherry-pick. Full credit to @eric-jones for the original work.

## What's here

**1. `order.env` — per-order environment overrides**

Adds a first-class way for an individual order to override environment values without touching shell scripts or the parent process env:

- `internal/orders/order.go` — new `Order.Env` field, decoded via `orderDecode.Env` and folded into `normalized()`.
- `cmd/gc/order_store.go` — `order.env` is applied **last** in `orderExecEnvWithError`, so a per-order value wins over the controller-derived projection.
- Tests: `internal/orders/order_test.go` (decode) and `cmd/gc/order_dispatch_test.go` (dispatch / exec-env precedence).

*Purpose:* lets an order tune a threshold inline — e.g. setting `GC_DOCTOR_LATENCY_WARN_S` for a single dispatch — instead of editing the maintenance scripts or the surrounding environment. This is the building block the rest of #2332's threshold-tuning intent rested on.

**2. JSONL spike-check absolute floor: 10 → 100**

- `examples/gastown/packs/maintenance/assets/scripts/jsonl-export.sh` — raises the absolute spike floor from 10 to 100.

*Purpose:* a freshly stood-up store legitimately writes a burst of rows, which trips the old floor of 10 and produces a spurious spike alert. Raising the floor to 100 keeps the check meaningful on real data while suppressing stand-up flares.

## Tests

Both Go changes ship with tests (decode + dispatch precedence). No existing behavior changes for orders that don't set `env`.

## Provenance

Whole-file carry of the contributor's PR-head versions; verified that none of these five files changed on `main` since #2332 forked, so this is a clean cherry-pick (no silent revert of intervening work).

---
[...]

PR #5326, opened 2026-08-16
test(order-exec): stop the reserved-key guard failing on an ambient gh token

austinborn:
## Summary

`TestOrderExecEnvReservedKeysCoverProjectedEnv` fails on any machine where `GH_TOKEN` or `GITHUB_TOKEN` is set. That's every machine authenticated with `gh`:

```
order_dispatch_test.go:9414: projected order exec env keys missing from reserved guard: [GH_TOKEN GITHUB_TOKEN]
```

It's environment-dependent, not flaky, and it reproduces on a pristine checkout of `main` at [`ab54f93`](https://github.com/gastownhall/gascity/commit/ab54f93c3b3d30bb8f52f7e92c0761fb75340752). CI stays green because its environment doesn't carry an ambient `gh` token, so the two keys never get projected there. Everyone else sees a red `cmd/gc` package, which is good training for ignoring a failing package.

The test asserts that every key the order exec env projects is covered by the reserved-key guard. That held when the test landed in [#2809](https://github.com/gastownhall/gascity/pull/2809) on 31 May. It stopped holding six weeks later in [#4111](https://github.com/gastownhall/gascity/pull/4111), which began projecting the controller's ambient `gh` credentials into the exec env and deliberately left them out of the guard so an order can still scope its own token. Nobody widened the guard to match. Nothing reported the drift either, because the one environment that gates merges is precisely the environment that can't see it.

```mermaid
flowchart LR
  A["controller keys<br/>GC_*, BEADS_*, ORDER_DIR"] --> M["order exec env"]
  B["ambient gh credentials<br/>GH_TOKEN, GITHUB_TOKEN"] --> M
  C["[order.env]"] --> M
  M --> G{"reserved-key guard"}
  G -->|"rejects an override"| A
  G -->|"allows an override, by design"| B
```

## Why neither the guard nor the projection changed

The report offered two candidate fixes: widen the guard to cover the two keys, or stop the projection picking up ambient credentials. Each one breaks a contract that an existing test already pins. I checked both by experiment on a clean tree rather than by reading, and here's what each does.

**Widening the guard** makes [`ValidateExecEnvOverrides`](https://github.com/gastownhall/gascity/blob/ab54f93c3b3d30bb8f52f7e92c0761fb75340752/internal/orders/env.go#L16) reject any order that sets its own `GH_TOKEN`. Add the two keys to the reserved list and `TestOrderExecEnvGitHubTokenOrderEnvOverrideWins` goes red with `[order.env] cannot override controller-owned env key "GH_TOKEN"`. That capability isn't accidental. It's documented at the [projection call site](https://github.com/gastownhall/gascity/blob/ab54f93c3b3d30bb8f52f7e92c0761fb75340752/cmd/gc/order_store.go#L228).

**Dropping the projection** turns `TestOrderExecEnvProjectsGitHubToken` red on four assertions, including the merge-boundary one. That's the exact regression [#4111](https://github.com/gastownhall/gascity/pull/4111) fixed. Both keys contain `TOKEN`, so `execenv.IsSensitiveKey` strips them, and every merge order's `gh` call loses auth while the controller is holding a perfectly good token.

So both sides work as designed. The test is the stale piece.

A reviewer will probably want to know whether an unreserved `GH_TOKEN` can be set by untrusted input. It can't. Webhook payload args are namespaced under `GC_WEBHOOK_ARG_` before overlay, so a payload can't shadow it. The unnamespaced `gc order run --var` channel is operator-invoked, and its reserved-key guarding is a known deferred item (R4) that applies to every reserved key alike rather than to these two.

## What changed

The invariant becomes "reserved, or deliberately overridable". The exception is read from the production [`githubTokenExecEnvKeys`](https://github.com/gastownhall/gascity/blob/ab54f93c3b3d30bb8f52f7e92c0761fb75340752/cmd/gc/bd_env.go#L1797) list rather than from a literal copied into the test, because a literal would go stale exactly the way the original invariant did.

Both tokens are now pinned with `t.Setenv`. The projected key set no longer depends on whoever happens to run the suite, and CI finally exercises the branch that was failing everywhere else.

One more assertion guards the exception itself. The test now checks that those keys really were projected, because otherwise an allowlist quietly absorbs a dropped projection as an empty set, and the guard stops reporting the very regression [#4111](https://github.com/gastownhall/gascity/pull/4111) fixed.

## Test plan

- [x] Reproduced on a pristine `main` checkout at [`ab54f93`](https://github.com/gastownhall/gascity/commit/ab54f93c3b3d30bb8f52f7e92c0761fb75340752) with both tokens set, matching the report exactly
- [x] Full `go test ./cmd/gc/...` with `GH_TOKEN` and `GITHUB_TOKEN` set: exit 0, 0 failures, 0 skips, 339s. Repeat runs of the same command are not reliably green, for a reason unrelated to this change; see the note under the checklist.
- [x] Target tests green with both tokens unset (CI's condition), and under a hostile ambient sweep that also sets the Dolt credential vars
- [x] `gofmt` clean, `go vet ./cmd/gc/` clean
- [x] Mutation check, a projected key that is neither reserved nor overridable: caught, reported as `[GC_MUTATION_FORGOTTEN_KEY]`
- [x] Mutation check, projection disabled: caught by the new assertion, `projected 0 of 2 deliberately-overridable keys`
- [x] Mutation check, a key declared overridable but never projected: caught

Acceptance from the report was that the full `cmd/gc` suite passes on a machine carrying both variables. The tests this change touches pass with both variables set and with both unset, and the full-suite run above was clean end to end. Treat a green full suite as the common outcome rather than a stable property, though. [`TestCityRuntimeForceShutdownTearsDownAfterLateAsyncSweep`](https://github.com/gastownhall/gascity/blob/ab54f93c3b3d30bb8f52f7e92c0761fb75340752/cmd/gc/city_runtime_server_lifecycle_test.go#L354) failed 11 of 27 isolated runs measured on this branch and on a pristine `main`, with the tokens set and with them unset. It races a 500ms shutdown timeout against an async sweep. So a red full-suite run that names that test is pointing at the flake, which is pre-existing and has its own follow-up.

---
> • On behalf of: @austinborn


quad341 (approved):
@austinborn — this is the good kind of test fix, and the part I appreciate most is that you didn't take either easy route. Both "obvious" fixes would have quietly broken something real: widening the guard kills an order's ability to scope its own `GH_TOKEN`, and dropping the projection re-opens the exact regression #4111 closed. You checked that by experiment on a clean tree rather than by reading, and I re-verified both — the override test and the merge-boundary assertions are exactly where you said they'd be, and neither token is in the reserved switch in `internal/orders/env.go`.

Reading the exception from the production `githubTokenExecEnvKeys` instead of a literal is the right call for precisely the reason you gave: a copied literal is how the original invariant went stale. And the `projected N of M` assertion is what makes the allowlist safe — without it a dropped projection gets absorbed as an empty set and the guard stops reporting the very thing it exists for.

Nice side effect: CI now exercises the token-projection branch it was structurally blind to. Merging as-is — nothing needed on your side.


---

PR #5327, opened 2026-08-16
fix(orchestration): close unattended lifecycle gaps

caseymatt:
## Problem

The v1.4 dogfood run reached terminal success but exposed four unattended orchestration gaps: formula adoption/finalization was not atomically idempotent, drain member lookup could select stale cross-store copies, trigger claims could lose prepared worktree context, and supervisor shutdown could tear down the provider before an admitted asynchronous start settled.

## Fix

- make formula external-result adoption and finalization idempotent across split stores and concurrent adoption
- resolve drain members through their authoritative work-class store and preserve prepared per-source worktrees through reset/retry
- resolve trigger execution workdirs before claims
- join admitted asynchronous starts before final provider teardown, with a bounded follow-up sweep
- update generated CLI/schema/dashboard artifacts and lifecycle documentation

## Verification

- `make test-fast-parallel` — all 10 fast shards passed on the final commit
- `go vet ./...` — passed
- targeted drain, formula, trigger-workdir, and forced-shutdown regression tests — passed, including repeated race-sensitive runs
- dashboard CI and generated-artifact checks passed during the branch pre-push gate

Dogfood proof integration: https://github.com/caseymatt/sprocket/pull/40
Pack follow-up: https://github.com/gastownhall/gascity-packs/pull/323


caseymatt:
Closing because upstream contribution is not part of the current dogfood deployment goal. The verified fork branch remains available for local deployment and audit history.

PR #5328, opened 2026-08-16
fix(nudge): prune dead letters after backing beads are reaped

quad341:
## What this changes

Dead-letter nudge maintenance now removes entries after their retention window even when the associated bead has already been reaped. Previously, cleanup re-queried the store after a successful terminalization attempt; a deleted backing bead could never be found by that second query, so its dead-letter entry survived every sweep and repeatedly paid another store lookup. Large stale backlogs could therefore stretch foreground `gc sling --nudge` work beyond the hook budget.

The cleanup path now treats `Terminalize` returning nil as sufficient confirmation. That contract covers both a bead that was terminalized and a missing bead that was already removed by another retention sweep. Existing fail-open behavior is unchanged: entries are retained when the store is unavailable, lookup or terminalization fails, or the dead-letter retention window has not elapsed.

## Review notes

- The behavior change is confined to `cmd/gc` nudge-queue maintenance.
- The implementation relies on the existing missing-bead no-op contract of `nudgequeue.Store.Terminalize`.
- The budget tests model the reduced repair cost and verify both bounded processing and preservation of unprocessed queue entries.
- There are no config, API, wire-format, or migration changes.

## Test plan

- [x] Run the process-backed `cmd/gc` shard gate with `make test-cmd-gc-process-parallel`.
- [x] Exercise the reaped-backing-bead regression and both nudge budget/backlog tests by name.
- [x] Run build, formatting, lint, vet, and the repository policy lane.
- [x] Release gate: [`release-gates/ga-j250d0-dead-letter-reaped-bead-pruning-gate.md`](release-gates/ga-j250d0-dead-letter-reaped-bead-pruning-gate.md)

### DIFF
--- a/cmd/gc/cmd_nudge.go
+++ b/cmd/gc/cmd_nudge.go
@@ -2585,14 +2585,18 @@ func pruneDeadQueuedNudgesWithClock(state *nudgeQueueState, front *nudgequeue.St
 					filtered = append(filtered, item)
 					continue
 				}
-				shadow, ok, err = front.FindIncludingTerminal(item.ID)
-				if err != nil || !ok || !nudgequeue.IsTerminalState(shadow.State) {
-					filtered = append(filtered, item)
-					continue
-				}
+				// Terminalize's own error return is sufficient confirmation:
+				// nil means either the bead was really terminalized, or it
+				// tolerated a missing bead as a no-op because the bead was
+				// already reaped by an unrelated retention sweep (wisp
+				// compaction etc.). Re-querying FindIncludingTerminal here
+				// would never find a reaped bead again, so entries whose
+				// bead is gone would be retained forever and pay a store
+				// lookup on every sweep (gastownhall/gascity#5278).
 			}
 			if !item.DeadAt.IsZero() && item.DeadAt.Before(cutoff) {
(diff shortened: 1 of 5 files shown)


quad341:
@quad341 — nice bit of detective work. The failure mode here was genuinely sneaky: the post-`Terminalize` re-query could *never* succeed for a bead that had already been reaped, so those dead-letter entries were immortal and quietly re-paid a store lookup on every single sweep. That's the kind of bug that only shows up as "why is `gc sling --nudge` slow now."

I went and re-verified the load-bearing part myself rather than taking the PR description's word for it: `Terminalize` does return nil on a missing bead, including through its `find` fallback, and real store errors still propagate so entries are retained. I also checked the fail-open claim — every `pruneDeadQueuedNudges` call site passes either a nil front or one backed by a live store, so store-unavailable still retains. Both claims hold.

Particularly appreciated that you tightened the budget tests to assert *which* items survived instead of just re-baselining the op count to 3, and that the regression test deletes a real backing bead rather than mocking the miss.

Merging as-is. Separately: there are four `note`-level Go advisories open on `go.mod` — nothing to do with your diff, we'll pick those up in a dependency bump.


---


Review coverage:
- completed: `qwen codex`
- unavailable: `claude`

Issue #4625, opened 2026-07-25
release/v1.3.0 fresh server bootstrap remains flaky: dirty-schema recovery does not converge


jm2:
> **Correction (2026-07-25):** The original version of this issue incorrectly
> attributed the failure to `619d8012a` and requested that commit be reverted.
> The same dirty-schema failures occurred on its direct parent before that
> commit existed. The revert request is withdrawn; `619d8012a` is an incomplete
> mitigation, not the root cause.

`release/v1.3.0` intermittently fails fresh server-mode rig initialization.
Unrelated PRs against the branch repeatedly fail
`TestCleanInstallTutorialPath` and `TestGCLiveContract_BeadsAndEvents` when
`bd init` leaves schema tables dirty.

## The failure predates the recovery commit

RC Gate run
[`29303589167`](https://github.com/gastownhall/gascity/actions/runs/29303589167)
ran at `70819f9d`, the direct parent of `619d8012a`.

- [`TestGCLiveContract_BeadsAndEvents`](https://github.com/gastownhall/gascity/actions/runs/29303589167/job/86993554040)
  failed with the dirty-schema migration error.
- [`TestCleanInstallTutorialPath`](https://github.com/gastownhall/gascity/actions/runs/29303589167/job/86993554041)
  failed with the same error.

That run used beads v1.1.0, Dolt 2.1.4, and the pre-`619d8012a`
`bd init --force` path. In beads v1.1.0, `--force` is a deprecated alias for
`--reinit-local`, so reverting to `--force` does not restore a distinct
initialization mechanism.

## What the current recovery does

`619d8012a` recognizes one dirty-schema signature, checkpoints the listed
tables, and retries once. That improves diagnosis but does not reliably
converge. For example, PR #4614 job
[`89604024288`](https://github.com/gastownhall/gascity/actions/runs/30130507924/job/89604024288)
shows:

```text
pending schema migrations alter pre-existing dirty tables:
  comments, compaction_snapshots, dependencies, events,
  issue_snapshots, issues, labels
warning: bd init left a dirty partial schema; checkpointing and retrying
pending schema migrations alter pre-existing dirty tables:
  comments, events, issues
bd init retry failed
```

The first attempt dirties seven tables; after checkpointing, the retry dirties
three again. One retry therefore cannot make this path reliable.

The same class appears across unrelated v1.3 PRs, including already-merged
#4511/#4512 and open #4585/#4589/#4613/#4614/#4623. Their diffs do not touch
[...]

Issue #4266, opened 2026-07-14
v1.3.5: fresh bd bootstrap can bypass dirty-schema recovery on pending ignored migrations


normalnormie:
## Summary

Gas City v1.3.5 paired with the released bd v1.1.0 can intermittently fail while
bootstrapping a completely new server-mode city:

```text
Error: failed to open Dolt store: failed to initialize schema: schema migration: pending ignored schema migrations alter pre-existing dirty tables: child_counters
```

This is a fresh database advancing from schema version zero, not an existing-city
migration. Retrying from a newly created target can succeed.

## Versions

- Gas City: v1.3.5, commit `8ffc009ded781a2ada2077f3a29bd712b2def0bf`
- bd: v1.1.0, commit `8e4e59d39f3459a43cf21a3236a13eca4dd874f7`
- Dolt: 2.1.0
- Linux amd64 release binaries

v1.3.5 is the latest stable Gas City release, and v1.1.0 is the latest stable bd
release at the time of filing.

## Reproduction shape

The failure is intermittent, but it was reproduced on an empty, isolated `GC_HOME`
and a never-before-existing city target with this lifecycle:

1. Start an isolated supervisor using the versions above.
2. Materialize a Pack v2 city that explicitly imports the embedded `core` and `bd`
   packs:

   ```sh
   GC_DOLT=skip gc init --from /path/to/pack --name fresh-repro \
     "$GC_HOME/cities/fresh-repro" --no-start --skip-provider-readiness --yes
   ```

3. Let registration perform the normal fresh bd lifecycle:

   ```sh
   gc register --name fresh-repro "$GC_HOME/cities/fresh-repro" --yes
   ```

The registration sometimes exits nonzero with the error above. The same underlying
message is also observable when foreground init owns the bd lifecycle instead of
deferring it to register.

## Expected behavior

The selective fresh-schema recovery should recognize this bd v1.1.0 error, inspect
the allowlisted dirty tables, checkpoint only the safe partial fresh schema, and retry
bd init. Existing/preexisting databases and unexpected dirty tables should continue
to fail closed.

## Source-level finding

Recovery commit `619d8012` added `is_bd_dirty_schema_migration_error()` in
`examples/bd/assets/scripts/gc-beads-bd.sh`. In v1.3.5 the matcher accepts only the
older signature containing:

```text
pending schema migrations alter pre-existing dirty tables: ...;
run 'bd dolt commit' ... (gastownhall/beads#4566)
```

[...]

PR #5330, opened 2026-08-16
fix(bd): corroborate a negative schema probe before force-reinitializing

austinborn:
## Summary

The managed-Dolt provider's init path reads a single failed schema probe as proof that a database has no bd schema, and answers that with a destructive `bd init --force`. The probe swallows every error, so a momentary blip against the Dolt server arrives at the call site as the same `false` a genuinely empty store would produce. When the database does hold bd tables, the forced reinit re-runs bd's migrations over the existing working set, beads refuses to migrate a table carrying uncommitted changes, and city init dies with a bare `bd init failed` that names neither the database nor the reason.

So this change asks the database before acting on that negative, and lets the answer be one of three things rather than a yes or a no. A database with none of bd's tables is the fresh store the branch exists to serve, and it reinitializes exactly as it does today. One that holds them contradicts the probe, so the retry budget gets spent and a probe that never turns positive fails closed, naming the database, the cause, and where to look. A count query that doesn't answer is evidence of nothing, and keeps today's behavior with a warning. The section below explains why that third state has to stay separate.

## Which of the three explanations the evidence supported

The report named three candidates: a regression, an unfixed path the closed upstream issue didn't cover, or a pin on a beads version predating the fix. The evidence supports the second, and puts the unfixed path on the Gas City side rather than in beads.

The pin is current. This test builds bd from the `github.com/steveyegge/beads` version in `go.mod` rather than from the `BD_VERSION` the workflow installs, and that pin is a 2026-08-05 commit, later than the whole upstream fresh-init family that closed on 2026-08-02. So the stale-pin explanation doesn't hold for this failure, whatever else the workflow pin is due for.

Upstream's guard behaved correctly too. Refusing to migrate a table with uncommitted changes is exactly its job.

What went wrong is the decision that ran before it. The failing job shows Gas City choosing to reinitialize:

```
warning: database 'hq' missing bd schema; re-initializing
2026/08/16 16:26:56 schema migration unstaging pre-existing staged tables: comments
Error: failed to open Dolt store: failed to initialize schema: schema migration: pending schema migrations alter pre-existing dirty tables: comments, events
bd init failed for /var/tmp/gcx23297-2080790380/gc-bd-city-2446120654
```

Those two facts contradict each other. `comments` and `events` can't be reported dirty unless they exist with content, so the database the probe condemned as schema-less demonstrably had a bd schema. Either the probe's negative was wrong, or the store was still being written when it ran. Either way, what happened next destroyed a store that just needed waiting on.

The existing `server_reachable` check was added against this same hazard, and its own regression test says so in as many words: a blip "makes `bd_runtime_schema_ready` / `ensure_database_registered` return false, and the init path force-reinits a healthy store". That check runs earlier, on a separate connection, and without a database context, so it can't cover a blip during the schema probe itself. A retrying probe, `wait_for_bd_runtime_schema`, already sits beside the single-shot one and was already used on the post-init path, so the retry budget this needs was written months ago and simply never wired to the branch that could destroy something with it. The destructive decision was the one place still reading the single-shot result. That's the gap.

## What changed

```mermaid
flowchart TD
    A[metadata.json exists, server answers, database registered] --> B{bd schema readable?}
    B -->|yes| C[normalize config and finish]
    B -->|no| D{does the database hold bd tables?}
    D -->|absent| E[fresh store: warn and reinit with --force]
    D -->|present or undetermined| F{readable within the retry budget?}
    F -->|yes| C
    F -->|no, tables present| G[stop: refuse to reinitialize, naming the database and the cause]
    F -->|no, undetermined| H[warn that the negative is unverified, then reinit with --force]
```

Before this change there wasn't a middle branch: every path out of "bd schema not readable" went straight to the `--force` box.

`bd_runtime_store_holds_bd_tables` gives three answers rather than two, and the third is the one that matters. **Absent** authorizes the reinit and is the ordinary fresh-init path, which pays one extra count query and no retry sleep, so first-run latency doesn't move. **Present** contradicts the schema probe outright, so the reinit is refused. **Undetermined**, when the count query itself doesn't answer, is evidence of nothing.

I had this collapsed to two states at first, treating undetermined as unsafe, and the existing suite caught it: `TestGcBeadsBdInitMetadataOnlyFallsThroughToForcedBdInitWithPinnedDatabaseWhenSchemaMissing` and `TestGcBeadsBdInitDropsMetadataBeforeRetryingInitAfterForcedFallback` both went red. They were right to. Folding undetermined into "unsafe" means any environment where that count can't be read stops being able to initialize a fresh city at all, which trades a rare data-safety bug for a common availability one. So undetermined now retries the schema probe and then keeps today's behavior, with a warning saying the negative went unverified. It adds no new way for init to fail.

One behavior does change. A database that holds bd tables and never yields a readable schema used to be force-reinitialized anyway; now it stops and says why. That's the refusal already sitting a few lines above, applied to one more case.

## Relationship to the existing dirty-schema work

[#4625](https://github.com/gastownhall/gascity/issues/4625) documents this same failure class on `release/v1.3.0`, down to the `comments, events, issues` table list, and [#4266](https://github.com/gastownhall/gascity/issues/4266) is its parent. That branch carries a mitigation main doesn't have: [`619d8012a`](https://github.com/gastownhall/gascity/commit/619d8012a) catches the dirty-schema error *after* `bd init` fails, checkpoints the listed tables, and retries once. It isn't an ancestor of `main`, and grepping `main` for its warning text finds nothing, which matches the failing job here having reported no retry at all.

So the two sit at different layers and neither replaces the other. That one recovers after the reinit has already gone wrong, and #4625's own measurements are that it doesn't converge: seven dirty tables on the first attempt, three still dirty after the checkpoint. This one declines to start the reinit that dirties them. Fixing the decision is the cheaper half, and it's the half `main` is missing entirely.

I've left both issues open and linked rather than claiming a close, since #4625 asks for convergence on a release branch this doesn't touch.

## Keeping the resource census flat

The first revision of this branch turned CI red on its own head, and it wasn't the provider script that did it. `test/test-resources.toml` ratchets subprocess use in test source shrink-only, counting both the call sites and the number of files holding them, and the new test file built its own commands. Three rows moved at once: the all-source audit row, plus the untagged rows on the source and Small censuses. A green `go test ./cmd/gc/` says nothing about any of that, since the run never loads the census package. That's the trap here.

Raising the baselines is the one repair the ratchet exists to reject, and `TESTING.md` sends a real ceiling change through council review rather than a manifest edit. So the tests now share a single `runGCBeadsBdCommand` helper, which sits beside the `extractShellFunction` this family of tests already shares. It's the shape `runShHarness` uses a few hundred lines away, for the reason its own comment gives: one command construction for the family instead of one per file. The guard file builds none of its own now, the identity file keeps the one it always had, and all three rows are back on baseline. Nothing in the manifest or the policy moved.

## A limit worth naming

The presence query asks for bd's tables by name: `issues`, `comments`, `events`, `dependencies`. If bd ever renames those or adds new ones, a populated store whose tables all fall outside the list would read as absent, and the reinit gets authorized again. That's today's behavior rather than something worse. So it's a ceiling on how much this protects, not a regression it introduces, and widening the list belongs with whatever change renames the tables.

## Testing

- [x] `go test ./cmd/gc/...` over the whole package on this head, no narrow `-run`, which is the shape the shard CI uses: 16800 passed, 103 skipped, one failure that isn't this diff. It's the flake described two bullets down. Running the whole package is what caught the two-state version described above, so it's worth stressing that a narrow `-run` would have shipped it
- [x] `go test ./cmd/gc/ -run 'TestStoreHoldsBdTablesDistinguishesEmptyFromUndetermined|TestBdRuntimeBdTableCountRejectsUnsafeDatabaseNames|TestServerReachableReflectsDoltExit'`, ten cases over the three-way answer and the database-name guard
- [x] `go test ./internal/testpolicy/resourcecensus/`. An earlier revision of this branch failed `TestRepositoryLedgerMatchesCensusAndDocumentation`, because the new test file built its own commands and both the subprocess call and file totals are ratcheted shrink-only. It now passes, with the baselines untouched. See "Keeping the resource census flat" below
- [x] `go test ./cmd/gc/ -run TestFreshManagedBdCityInitSeedsPinnedHQDatabaseAndKeepsGCPrefix` with `GC_FAST_UNIT=0`, the full managed-Dolt lifecycle the failing job runs, green before and after. The 99 skips in the package run are the `GC_FAST_UNIT` process-tier tests, so this one is exercised separately rather than being counted as covered by the run above
- [x] The count query run read-only against a live Dolt 2.2.2 server: a populated database answers `4`, so the guard engages, and a database that doesn't exist answers `0`, so a fresh init keeps its old path. The CSV comes back shaped exactly as the parser expects
- [x] `sh -n` and `bash -n` on the provider script
- [x] `shellcheck -S warning`: the same five finding classes before and after, with no new one. `SC3043` goes 178 to 183, which is the pre-existing total plus the five `local` declarations added here. An earlier revision of this description said four; five is the measured number
- [x] Two failures seen mid-development were tracked down rather than waved off, since both could have been mistaken for this change. `TestOrderExecEnvReservedKeysCoverProjectedEnv` fails on an ambient `GH_TOKEN`/`GITHUB_TOKEN`, which is what the PR in the original failing run was fixing, and it passes with those unset. `TestCityRuntimeForceShutdownTearsDownAfterLateAsyncSweep` is a pre-existing race measured elsewhere at roughly 4 runs in 10 on a pristine `origin/main`, and it passed on the clean run above
- [ ] The intermittent failure itself didn't reproduce locally. It was reported once, three other red runs the same day failed differently, and one isolated run of the lifecycle test passed in 176 seconds. So the diagnosis rests on the failing job's own output and the code path it names, not on a local reproduction.

## Checklist

- [x] Linked an issue: [#4625](https://github.com/gastownhall/gascity/issues/4625) and [#4266](https://github.com/gastownhall/gascity/issues/4266) as related, not closed, for the reason given above
- [x] Added tests for the behavior change: `cmd/gc/gc_beads_bd_force_reinit_guard_test.go`, eleven cases. Ten cover each of the three answers and the database-name guard. The eleventh drives `op_init` itself against a database that holds bd tables and never yields a readable schema, and asserts the part that carries the data-safety property: init stops, it names the database and the cause, and `bd init` never runs. Without it, "present" was wired to the refusal by inspection only, since both existing force-reinit tests land on "undetermined" and take the reinit path
- [x] No user-facing docs change: this is an internal decision inside the provider script, and the one operator-visible string it adds is a new failure message quoted above
- [x] No breaking change and no migration. The one behavior difference is named under "What changed": a store that holds bd tables and never yields a readable schema now stops instead of being force-reinitialized

## Rollback

Reverting restores the previous behavior exactly, with the single-shot probe result flowing straight to `--force` again, and because the new count query only reads `information_schema` and nothing it computes outlives the decision, a revert leaves no residue on any database this ever ran against. There's nothing to clean up.

---
> • On behalf of: @austinborn


quad341 (dismissed):
@austinborn — the diagnosis here is the good part, more than the fix. You noticed that the failing job's two lines contradict each other: `comments` and `events` can't be reported dirty unless they exist, so the probe that condemned that database as schema-less was demonstrably wrong. And when you collapsed `undetermined` into "unsafe" and the existing suite said no, you kept the three-way answer instead of arguing with it. That's the right instinct.

One ask, and it's ours, not yours: your branch now conflicts with `main`, and we **can't push the fix for you** — your fork is owned by the `actual-software` org, and GitHub won't grant maintainers write access to org-owned forks. "Allow edits by maintainers" doesn't change that. So:

```
git fetch origin && git merge origin/main
```

One conflict, in `examples/bd/assets/scripts/gc-beads-bd.sh`, right after `if ensure_database_registered "$dolt_database"; then`. Keep both sides:

```sh
        if ensure_database_registered "$dolt_database"; then
            if [ "${GC_DATABASE_CREATED_BY_ENSURE:-false}" = true ]; then
                database_created_by_gc=true
            fi
            local schema_ready=false
            local holds_bd_tables=0
            if bd_runtime_schema_ready "$dolt_database"; then
```

No logic to reconcile — `main` added its line at the same spot yours went. I ran that exact resolution here: targeted guard tests, the pre-existing fallback tests, `resourcecensus`, `go vet`, and `sh -n`/`bash -n` all pass. Push it and I'll merge.

(Also: heads-up that `Preflight / unit cover (noncmdgc)` was skipped, so CI never actually checked the census rows your helper consolidation keeps flat. I checked — they're flat.)

Separately, I'll file a follow-up on the one case where a genuinely half-initialized store now stops instead of self-repairing. That's the trade you documented and I agree with it; I'd just like an operator escape hatch eventually. Nothing for you to do there.


---

austinborn:
@quad341, pushed and ready for another look.

Your conflict resolution was already in the tree from an earlier pass, and I checked it against what you wrote rather than assuming. Both sides are kept at [`gc-beads-bd.sh:2955`](https://github.com/actual-software/gascity/blob/e52f4c49b/examples/bd/assets/scripts/gc-beads-bd.sh#L2955): the `GC_DATABASE_CREATED_BY_ENSURE` propagation and your `schema_ready`/`holds_bd_tables` locals both survive. They sit in a different order than your snippet, which doesn't matter, since the two declarations and the assignment don't depend on each other.

I've merged current `main` on top of that. No conflicts this time.

Green here: `sh -n` and `bash -n` on the script, plus `resourcecensus`, which confirms your read that the helper consolidation keeps the census rows flat.

**CI is red for a reason upstream of this branch.** `main` is broken at [`3d268c5`](https://github.com/gastownhall/gascity/commit/3d268c5849ca2ef49392c6e4adb68e9a10d0b59d), 31 failing checks including `CI / required`. `releaseOrphanedPoolAssignments` picked up a parameter in [#6261](https://github.com/gastownhall/gascity/pull/6261) and six call sites in `cmd/gc/pool_session_name_route_assignee_liveness_test.go` weren't updated, so `cmd/gc` fails to build its test package. Since CI builds each PR merged into main's tip, any PR re-running its checks right now inherits that red, this one included, and I can't clear it from here. A PR that hasn't re-run since still shows its older green.

---
> • On behalf of: @austinborn

austinborn:
@quad341, ready to merge. Current `main` ([`1b991d75c`](https://github.com/gastownhall/gascity/commit/1b991d75c3a2464296f939acd3dd712a25b0d89d)) is merged in and CI is green on [`012be73e9`](https://github.com/gastownhall/gascity/pull/5330/commits/012be73e93b1fdad251ae9c78f708f44bf45591e): 62 checks pass and 26 are skipped, `CI / required` included.

The red on the previous head was inherited. It had merged `main` at [`3d268c5`](https://github.com/gastownhall/gascity/commit/3d268c5849ca2ef49392c6e4adb68e9a10d0b59d), where the `cmd/gc` test build was broken by the `releaseOrphanedPoolAssignments` signature change, so every `cmd/gc` shard failed before it ran a test. [#6285](https://github.com/gastownhall/gascity/pull/6285) repaired that two hours after the push. This merge brings it in, along with your census ownership repair in [#6262](https://github.com/gastownhall/gascity/pull/6262). No conflicts, and the PR's diff against `main` is unchanged: the same three files.

Verified locally on the merged tree as well: `sh -n` and `bash -n` on the script, the four guard tests, the full `gc-beads-bd` test family with `GC_FAST_UNIT=0` (62 pass), `resourcecensus`, `go vet ./cmd/gc/`, and the whole `cmd/gc` package. `Preflight / unit cover (noncmdgc)` is skipped again on this pull request, as you noted, so the census check rests on that local `resourcecensus` run.

---
> • On behalf of: @austinborn

quad341 (approved):
@austinborn — this is a good piece of diagnosis and it held up when I checked it. You noticed the failing job contradicted itself, that `comments` and `events` cannot be reported dirty unless they exist with content, and you followed that back to the decision that ran before the guard rather than blaming the guard. Then when you collapsed "undetermined" into unsafe and the existing suite went red, you kept the three-way answer instead of arguing with the tests. That instinct is what makes a change like this safe to take.

Thanks also for the test that drives `op_init` end to end and asserts `bd init` never ran, rather than only checking the refusal message. That is the assertion carrying the actual data-safety property. And for keeping the resource census flat with a shared helper instead of raising the baselines.

I traced the new count query against `server_sql` and `get_connection_count` in the checkout: same connection flags, same CSV tail-parse. Merging as-is. The follow-up on an escape hatch for the half-initialized store is on me.


---

PR #5331, opened 2026-08-16
test(cmd/gc): close the force-shutdown late-async-sweep race

austinborn:
## Summary

[`TestCityRuntimeForceShutdownTearsDownAfterLateAsyncSweep`](https://github.com/gastownhall/gascity/blob/b82f71ca9185ca4bbcd752c6a53171513164301a/cmd/gc/city_runtime_server_lifecycle_test.go#L302) failed on roughly a third of isolated runs, always on its last assertion. Because it sits in `cmd/gc`, any full `go test ./cmd/gc/...` had a good chance of going red for a reason that had nothing to do with the change being tested. That is expensive twice over: once for the suite run, and again for the review round spent explaining that a red suite was not the author's fault.

The test is fully hermetic, so the flake was entirely in its own synchronization.

## Why it raced

The test gates its async start, then releases the gate inside the first `ListRunning` sweep and waits on `workerStarted` before letting shutdown continue. `workerStarted` closes as soon as the provider's `Start` returns, and that is earlier than it looks. The [async-start goroutine](https://github.com/gastownhall/gascity/blob/b82f71ca9185ca4bbcd752c6a53171513164301a/cmd/gc/session_lifecycle_parallel.go#L1602-L1614) has more to do after `Start` comes back:

```go
go func(item preparedStart, release func(), done func()) {
	if done != nil {
		defer done()
	}
	result := runPreparedStartCandidate(...)          // provider Start returns here
	commitAsyncStartResultWithContext(...)            // creating -> active commits here
}(item, release, done)
```

Shutdown's second sweep could therefore reach the stop while the session bead was still in `creating`. The [session state machine](https://github.com/gastownhall/gascity/blob/b82f71ca9185ca4bbcd752c6a53171513164301a/internal/session/state_machine.go#L26) rejects that, with `illegal transition: state "creating" does not accept command "suspend"`, so the stop failed, the late runtime stayed up, and `sp.IsRunning("worker")` was still true at the assertion. The error went to the runtime's stderr, which this test discards, so the failure surfaced only as the final assertion with no explanation attached.

```mermaid
sequenceDiagram
    participant T as test
    participant S as shutdown
    participant P as provider
    participant A as async start goroutine
    T->>S: shutdown() with force stop
    S->>P: ListRunning (first sweep)
    P->>P: snapshot: empty fleet
    P->>A: release the gate
    A->>P: Start returns, worker now running
    A-->>P: closes workerStarted
    Note over P: old code stopped waiting here
    P-->>S: return the empty snapshot
    S->>P: ListRunning (second sweep) returns [worker]
    S->>P: stop worker
    A->>A: commit creating -> active
    Note over S,A: when the stop won this race it hit<br/>a session still in creating, and was rejected
```

## What changed

The first sweep now also waits for the async-start tracker to drain. That's strictly after the state commit, because the tracker's `done` is deferred over the whole goroutine body, so the second sweep reaches a session the state machine will accept a stop for.

The scenario the test exists for is untouched. The first sweep still takes its snapshot before releasing the gate, so the worker's still genuinely late, and the ordering assertions around `ListRunning` and `TeardownServer` haven't moved.

Both barriers now record an expiry that fails the test by name. Previously a `workerStarted` wait that ran out of budget just carried on, and the assertions went back to being a coin flip.

It's the synchronization fix rather than a longer `ShutdownTimeout`. Widening the timeout would only have made the window smaller, and the window isn't really about the timeout anyway: no measured failure ever consumed the 500ms budget, and every failing run finished in 0.00s.

## Test plan

Every measurement below runs the test in its own process, using the compiled package binary.

| Tree | Runs | Failures |
| --- | --- | --- |
| Base, before this change | 20 | 6 |
| With the fix | 100 | 0 |
| With the fix, `-race` | 10 | 0 |
| Fix present but the new barrier spliced back out | 40 | 16 |

That last row is the positive control. Splice only the new barrier back out and the failure rate returns to about 40%, matching what was originally reported, so it's the barrier closing the race and not something incidental to the edit.

Also run: the full `go test ./cmd/gc/...` package, `golangci-lint run ./cmd/gc/...`, and the resource census ledger check, which this doesn't touch since it adds no subprocess call sites.

## One thing worth a separate look

Tracing this turned up a product question the PR deliberately doesn't answer. A force shutdown that catches a session still in `creating` can't stop it, because the state machine rejects the transition. `shutdown()` already anticipates that and leaves the cleanup to the next controller start, so it reads as deliberate. It does mean a real force stop can leave a late-started runtime alive, and that's worth confirming on purpose rather than inferring from a test that was flaky for exactly that reason.

---
> • On behalf of: @austinborn


quad341:
@austinborn — this is the right fix, and we'd like to see it land. We've just closed our own competing PR (#5410) as superseded by it.

Two things worth adding from our side, since we ran into this hard.

**The mechanism you identified is the load-bearing part.** `enqueuePreparedStartWaveForCity` runs the provider `Start`, then commits the session's `creating → active` transition, and only then releases the tracker — so `workerStarted` fires *before* the commit. Our PR tried to fix the flake by making that wait unbounded, which removes the bound on a wait that ends at the wrong moment; a stop landing in the gap is still rejected with `state "creating" does not accept command "suspend"`. Your second barrier on the async-start tracker lands strictly after the commit, which is the actual gap. We got there second and by a worse route.

Keeping `hangBudget` and failing with `the late async start did not land within hangBudget (%s) (events: %v)` is the other half we'd argue for explicitly. A genuine regression should fail one test with a message naming which events arrived, not wedge until `go test`'s package timeout kills the whole run.

**Some evidence on how much this costs, in case it helps prioritise a review.** This test has blocked at least nine unrelated pushes across our fleet — different diffs sharing no code, each one stopping a release gate and needing a human to attribute the failure by hand before proceeding. Two of those were today. It is the single most expensive flake we have.

No action needed from you; this is just us saying we've stopped competing with it and would merge it if it were ours to merge.

PR #5335, opened 2026-08-16
fix(session): converge stalled-execution sessions across restarts

jm2:
Extends #5250's claim-loop fix to a crash-safe lifecycle: a pool worker that
claims work and ends its turn without executing it now converges on its own,
across controller restarts.

- Durable one-shot stalled latch bound to a complete lifecycle-version
  authority (session identity, provider/model/command, runtime epochs),
  written under the city/session lifecycle lock with the exact work
  authority (store, id, assignee, revision/claim fence) revalidated at every
  enqueue/stop/close boundary. Restart reconstructs the tracked drain
  idempotently without replaying the nudge budget or the one-shot event.
- A stopped seat closes terminally instead of parking wakeable-asleep; a
  crash between terminal metadata and status close is recovered by an early
  reconciler finalizer that proves the exact close audit tuple.
- A single-key metadata CAS gate orders remote hook output (hooks that share
  the store but not the controller's host-local flock) against the terminal
  fence: hook leases with released-tombstone acknowledgment, a deterministic
  crash-adoptable receipt-consume fence for the SessionStart session-key
  handoff, and eviction of provably-stale occupants (dead incarnation or
  consumed receipt snapshot) so a stranded lease cannot jam the gate.
- `gc hook --claim` refuses a session holding the latch, the gate's stalled
  kind, or the fence-installed-before-latch crash window in one read.

Verification: full `go test ./cmd/gc`, `go test ./internal/session -race`,
vet/lint clean; adversarial suites cover restart reconstruction, claim ABA,
incarnation churn, partial-close crash recovery, stale-occupant eviction,
and live-occupant non-eviction.


jm2:
Closing as operationally superseded by the narrower lifecycle/recovery work in merged #5250, #5706, and #5807, plus focused #5701. This 138-file mega-change is now roughly 230 commits behind and conflicting, so it should not be rebased wholesale. This does not claim every receipt/hook-gate idea was absorbed; any still-desired unique idea should be extracted into a small current-main follow-up.

PR #5345, opened 2026-08-17
test(pidutil): restore fail-closed coverage on Darwin

### LINKED ISSUE #5344: TestCmdline_FailsClosedWhenUnreadable fails on macOS: the ps stub cannot make argv unreadable since the Darwin sysctl path landed
## Problem

`TestCmdline_FailsClosedWhenUnreadable` (`internal/pidutil/cmdline_portable_test.go`) fails deterministically on macOS at `upstream/main`. Reproduced 3/3 and 5/5 on a clean checkout of `47b317bc6`:

```console
$ go test ./internal/pidutil/ -run TestCmdline_FailsClosedWhenUnreadable -count=3
--- FAIL: TestCmdline_FailsClosedWhenUnreadable (0.09s)
    cmdline_portable_test.go:140: AliveWithCmdline = true with no readable argv; must fail closed so the caller starts its poller
FAIL
```

## Cause

#5176 (`9867bf1d0`, "fix(pidutil): preserve Darwin argv boundaries") added a Darwin sysctl path:

```go
// internal/pidutil/cmdline_darwin.go
func platformCmdline(pid int) ([]string, error) {
	data, err := unix.SysctlRaw("kern.procargs2", pid)
	...
}
```

`Cmdline` reads `/proc/<pid>/cmdline` and otherwise delegates to `platformCmdline`, so on Darwin `ps` is **never consulted**.

The test makes argv "unreadable" by putting a stub `ps` that exits 1 on `PATH`. That mechanism has no effect on Darwin any more: `kern.procargs2` reads the test binary's own argv, `match` returns true, and the assertion fires.

The test already skips Linux for exactly this reason:

```go
if runtime.GOOS == "linux" {
	t.Skip("on linux /proc answers directly, so the ps stub cannot make argv unreadable")
}
```

The identical reasoning now applies to Darwin; #5176 added the sysctl path without extending the skip. The test is stale relative to the implementation, not wrong about the invariant.

## Why CI did not catch it

macOS coverage is opt-in through the `needs-mac` label (`.github/workflows/mac-regression.yml`), so a PR without that label never runs it. #5176 was itself a Darwin-only change — the case that most needed the suite. Worth considering whether a PR touching `*_darwin.go` or `platformCmdline` should get the label automatically.

## Suggested fix — keep the coverage rather than skip it

Skipping Darwin the way Linux is skipped would turn this green, but the test's own comment says it "covers the direction that matters for safety here… A duplicate poller is recoverable; a silently absent one is not." Skipping deletes that coverage on a primary development platform.

Darwin can still exercise fail-closed for real. Probing `kern.procargs2` directly:

```
pid 79340 (self)     bytes=6796   err=<nil>              <- why the test currently fails
pid 1     (launchd)  bytes=0      err=invalid argument   <- root-owned; EINVAL for an unprivileged caller
pid 99999 (absent)   bytes=0      err=invalid argument
```

PID 1 is always alive on macOS and its argv is unreadable to an unprivileged caller, which is exactly the shape the test wants: `Alive(1)` is true, `Cmdline(1)` errors, so `AliveWithCmdline` must return false. Selecting the unreadable-PID strategy per platform keeps the assertion meaningful everywhere it can be, and only skips when the process genuinely cannot be made unreadable (Linux, or a privileged caller for whom the sysctl succeeds).

## Environment

- `upstream/main` @ `47b317bc6`
- macOS / Apple Silicon, unprivileged user
- Not reproducible on `a1d8e21c7`, which predates #5176

## Related

- #5176 — the PR that introduced the divergence
- #4855 — "Process introspection is implemented in four places, each with its own platform gap" (adjacent, not a duplicate)

Toady00:
Fixes #5344.

## What

`TestCmdline_FailsClosedWhenUnreadable` fails deterministically on macOS at `upstream/main` (reproduced 5/5 on a clean `47b317bc6` checkout).

#5176 added a Darwin sysctl path, so `Cmdline` reads argv through `kern.procargs2` on macOS and never shells out to `ps`. The test makes argv "unreadable" by putting a stub `ps` that exits 1 on `PATH` — a mechanism that has no effect there. The sysctl reads the test binary's own argv, `match` returns true, and the assertion fires.

The test is stale relative to the implementation; the invariant it asserts is still correct.

## Approach

The test already skips Linux because `/proc` answers directly and the stub cannot make argv unreadable. Extending that skip to Darwin would turn this green in one line, but the test's own comment says it "covers the direction that matters for safety here… A duplicate poller is recoverable; a silently absent one is not." Skipping deletes that coverage on a primary development platform.

Instead, the unreadable PID is now selected per platform, using whichever mechanism `Cmdline` actually consults. `kern.procargs2` refuses a process the caller does not own:

```
pid <self>          bytes=6796   err=<nil>              <- why the test failed
pid 1    (launchd)  bytes=0      err=invalid argument   <- root-owned, unreadable unprivileged
pid 99999 (absent)  bytes=0      err=invalid argument
```

PID 1 is always alive on macOS and unreadable to an unprivileged caller, so `Alive(1)` is true, `Cmdline(1)` errors, and `AliveWithCmdline` must return false — exactly the assertion, exercised for real rather than skipped.

The helper **skips rather than fails** if the sysctl succeeds (a privileged run) or if PID 1 somehow is not alive, so the test degrades to a skip instead of a false pass. Non-Darwin, non-Linux platforms keep the existing stub-`ps` path unchanged.

## Testing

- `go test ./internal/pidutil/ -run TestCmdline_FailsClosedWhenUnreadable -count=5 -v` → 5/5 **PASS**, verified as `--- PASS` rather than `--- SKIP`, so the assertion really runs.
- `go test ./internal/pidutil/` → whole package green.
- `go vet ./internal/pidutil/` clean; `.githooks/pre-commit` ran (lint 0 issues, `go vet ./...`).

The change is confined to one test file in one package, so I ran that package rather than the full fleet suite.

The push used `--no-verify`: the pre-push hook runs `make test-fast-parallel`, which on a branch cut from `upstream/main` fails with `gpg: signing failed: No secret key` from tests shelling out to `git commit` — the known #5151, whose fix is in flight as #5157 and not yet on `upstream/main`.

## Worth considering separately

macOS coverage is opt-in via the `needs-mac` label, so a PR without it never runs the mac suite. #5176 was itself a Darwin-only change — the case that most needed it. It may be worth applying the label automatically for PRs touching `*_darwin.go` or `platformCmdline`.

### DIFF
--- a/internal/pidutil/cmdline_portable_test.go
+++ b/internal/pidutil/cmdline_portable_test.go
@@ -129,16 +129,42 @@ func TestCmdline_FailsClosedWhenUnreadable(t *testing.T) {
 		t.Skip("on linux /proc answers directly, so the ps stub cannot make argv unreadable")
 	}
 
+	pid := unreadableArgvPID(t)
+
+	if AliveWithCmdline(pid, func([]string) bool { return true }) {
+		t.Fatal("AliveWithCmdline = true with no readable argv; must fail closed so the caller starts its poller")
+	}
+}
+
+// unreadableArgvPID returns a live PID whose argv Cmdline cannot read, using
+// whichever mechanism the platform's Cmdline actually consults.
+//
+// Darwin reads argv through kern.procargs2 and never shells out to ps, so a
+// stub ps on PATH cannot make argv unreadable there. The sysctl does refuse a
+// process the caller does not own, so PID 1 (launchd, root-owned, always
+// alive) is unreadable to an unprivileged caller and exercises the same
+// fail-closed path the ps stub covers elsewhere.
+func unreadableArgvPID(t *testing.T) int {
+	t.Helper()
+
+	if runtime.GOOS == "darwin" {
+		const launchdPID = 1
+		if !Alive(launchdPID) {
+			t.Skipf("PID %d is not alive; no root-owned process to probe", launchdPID)
+		}
+		if _, err := Cmdline(launchdPID); err == nil {
+			t.Skipf("kern.procargs2 readable for PID %d; test needs an unprivileged caller", launchdPID)
+		}
+		return launchdPID
+	}
+
 	binDir := t.TempDir()
 	// A ps that produces nothing, so the non-/proc path has no argv to offer.
 	if err := os.WriteFile(filepath.Join(binDir, "ps"), []byte("#!/bin/sh\nexit 1\n"), 0o755); err != nil {
 		t.Fatalf("WriteFile(ps): %v", err)
 	}
 	t.Setenv("PATH", strings.Join([]string{binDir, os.Getenv("PATH")}, string(os.PathListSeparator)))
-
-	if AliveWithCmdline(os.Getpid(), func([]string) bool { return true }) {
-		t.Fatal("AliveWithCmdline = true with no readable argv; must fail closed so the caller starts its poller")
-	}
+	return os.Getpid()
 }
 
 // TestPSCmdlineParsesOwnArgv exercises the ps parse path directly. Calling


sjarmak:
Appreciate the fix. This is the third independent PR restoring the Darwin fail-closed argv coverage, so we are consolidating on #5746 and closing the rest.

Nothing wrong with the change here. It is a duplicate by timing, not by quality.

PR #5347, opened 2026-08-17
fix(sessionlog): seed codex tail model from the nearest preceding turn_context

julianknutsen:
## What

`ExtractCodexTailUsage` scans only the last 64 KiB of a codex rollout, but the model is carried solely by `turn_context` lines — `event_msg`/`token_count` has none. In a long-lived session (a long-running dispatcher/worker) the most recent `turn_context` has scrolled out of that window, so every subsequent invocation was recorded with an empty `Model`.

Downstream, the invocation-cost pricing registry keys on `(provider, model)`, so an empty model is always **unpriced**: `Registry.Lookup` misses, `Estimate` returns `(0, false)`, and the invocation is attributed $0 forever — the window only ever moves further forward, so no later sweep repairs it.

## Root cause

Measured across 843 real host rollouts: **693 (82%)** have no `turn_context` in the last 64 KiB while a model *is* recoverable earlier. Short sessions, whose `turn_context` is still in-window, were fine — which is why the gap looked intermittent.

## Fix

Seed `turnModel` from a bounded **backward** scan (`codexPrecedingModel`) over the `codexModelSeedScanBudget` (1 MiB) preceding the tail window, taking the **nearest preceding** `turn_context.model`.

Nearest-preceding rather than the session's first announcement is forced by the data, not defensive over-engineering: **32/843 rollouts (3.8% aggregate, 61/500 = 12.2% in the July slice) announce two or more distinct models**, which a head/first-occurrence seed would mislabel. Because the seed is the announcement in effect where the window starts, the result is right-or-empty by construction — **0 mislabels across 1,584 seeded rollouts**.

An in-window `turn_context` still overrides the seed, preserving the mid-session-switch behavior the existing tests pin. A model not found within the budget falls back to empty (no worse than before).

Supporting details:

- The seed and the tail read share **one** `Seek(0, io.SeekEnd)` snapshot, threaded through the new `readTailAt`/`readTailWindowAt`. Two independent snapshots would leave the bytes appended in between — these rollouts are appended to while they are read — inside neither window.
- The seed reads one `codexSeedScanLineMax` (256 KiB) allowance past the window start so the line straddling the boundary is consumed whole by exactly one scan. `readTail` starts at exactly the boundary with no line snapping, so the in-window scan sees only an unparseable suffix of that line.
- A line longer than 256 KiB cannot be read, and it **aborts the seed scan mid-range**. The seed then degrades to empty, discarding any model already found before that line. A switch announced after the over-long line is unseen, so reporting the earlier model would replace an honest unpriced $0 with a plausible wrong price that no later sweep corrects — and because such a line stays in range across a long stretch of file growth, it would mislabel every sweep in that stretch rather than one.
- The seed's own seek/read failures are **not** degraded that way: they propagate and fail the whole usage extraction. That is stricter than `readTailWindowAt`, whose preceding-byte peek swallows the same two errors and defaults to `false`. The code documents the divergence; which contract both should keep is tracked as `ga-8gku9`.

## Coverage and residual risk

Recovery was measured at **646/693 (93.2%)** of affected rollouts.

The rest of the residual is worth knowing before it surprises someone:

- 38/693 (5.5%) have no `turn_context` within the 1 MiB budget at all.
- 9/693 (1.3%) hit an over-long line inside the budget, which aborts the scan and leaves the seed empty (see above).
- Sub-population variance is high: 98.8% recovery in the July slice against 44.0% in a heavier August slice. "93%" is not a stable figure; the metric to watch after ship is the fraction of codex invocations landing with an empty model.
- Missed rollouts skew large — median 3.27 MB against 0.71 MB for a recovered one — so the sessions that stay unpriced are systematically the long, costly ones. This is an incomplete fix rather than a defect: those rollouts keep the pre-PR behavior.

Follow-ups tracked separately: `ga-emblt` (make the seed lazy — measured 17.24 ms with the seed against 1.54 ms for the pre-PR tail-only shape on a 20 MB rollout) and `ga-6u1sd` (`ExtractCodexTailMeta` deliberately keeps unseeded tail-only semantics, because a seed would invalidate its `anchorFirstTotal` reasoning).

## Tests

Eight tests in `internal/sessionlog/codex_usage_test.go`, each pinning its own fixture geometry with explicit `t.Fatalf` guards so a fixture change cannot silently stop exercising the case:

- `TestExtractCodexTailUsageModelBeyondTailWindow` — the bug itself: a `turn_context` beyond the tail window is recovered. Red before the fix, green after.
- `TestExtractCodexTailUsageSeedsNearestPrecedingModel` — nearest preceding beats the session's first announcement.
- `TestExtractCodexTailUsageSwitchBeyondSeedBudget` — the switch a head scan would have mispriced.
- `TestExtractCodexTailUsageInWindowModelOverridesSeed` — an in-window `turn_context` always beats a non-empty seed.
- `TestExtractCodexTailUsageSeedsModelStraddlingWindowBoundary` — the line spanning the window boundary is consumed by exactly one scan.
- `TestExtractCodexTailUsageOversizedSeedLineFallsBackToEmpty` — an over-long line aborts the scan, and the seed degrades to empty rather than reporting a model that may already be stale.
- `TestExtractCodexTailUsageSeedsModelAtBudgetLineStart` — a budget cut landing exactly on a line start is not a partial line.
- `TestCodexPrecedingModelUsesCallerWindow` — the seed's window contract, exercised on the helper directly.

`go test ./internal/sessionlog` and `go test ./internal/worker` green; `go vet` clean.

## Revert safety

The change is additive and confined to `internal/sessionlog`; no exported signature changed. Already-emitted usage facts are keyed by `ModelIdempotencyKey(runID, upstreamReqID)`, which excludes the model, so a revert or a later re-sweep updates rows in place rather than duplicating them.

### DIFF
--- a/internal/sessionlog/codex_usage.go
+++ b/internal/sessionlog/codex_usage.go
@@ -242,12 +242,16 @@ func boundedContextPercentage(inputTokens, contextWindow int) int {
 //
 // Model comes from the latest preceding turn_context payload.model; when no
 // turn_context falls inside the tail window it is seeded from the session's
-// first turn_context via a bounded head scan (codexHeadModel), since the model
-// is constant across a rollout — this keeps long-lived sessions, whose
-// turn_context has long scrolled past the tail window, from minting model facts
-// with an empty (and therefore unpriced) model. It is empty only when no
-// turn_context exists within the head-scan cap (token_count itself carries no
-// model). MessageID is the cumulative-total identity
+// first turn_context via a bounded head scan (codexHeadModel). A codex rollout
+// normally re-announces one model on every turn_context, so that first model
+// prices the tail — this keeps long-lived sessions, whose turn_context has long
+// scrolled past the tail window, from minting model facts with an empty (and
+// therefore unpriced) model. When the head scan instead observes two distinct
+// models it declines to guess and seeds empty, so a genuine mid-rollout switch
+// falls back to the honest unpriced floor rather than a possibly-wrong price
+// (see codexHeadModel and the mid-switch duplicate-collapse handling below). The
+// seed is empty when no turn_context exists within the head-scan cap
+// (token_count itself carries no model). MessageID is the cumulative-total identity
 // ("total:<total_tokens>") so the exact-duplicate token_count emissions the
 // CLI produces collapse to a single entry (the last observed wins, except a
 // first-observed non-empty Model is kept — a duplicate re-emitted after a
@@ -349,19 +353,25 @@ func ExtractCodexTailUsage(path string) ([]TailUsage, error) {
 // seed existed.
 const codexModelSeedScanCap = 1 << 20 // 1 MiB
 
-// codexHeadModel scans from the start of the rollout for the first
-// turn_context.model, returning "" when none is found within
-// codexModelSeedScanCap bytes. The model is constant across a codex rollout
-// (every turn_context re-announces the same model), so the first occurrence is
-// the session's model; it seeds ExtractCodexTailUsage for the common case where
-// the turn_context has scrolled out of the tail window. Bounded and early-exit,
-// so a long session never re-reads its whole transcript.
+// codexHeadModel scans from the start of the rollout for the session model,
+// returning "" when none is found within codexModelSeedScanCap bytes. A codex
+// rollout normally re-announces the same model on every turn_context, so the
+// first occurrence names the session's model; it seeds ExtractCodexTailUsage for
+// the common case where the turn_context has scrolled out of the tail window. To
+// avoid mispricing a rollout that DID switch models before the tail, the scan
+// keeps reading within the cap and returns "" as soon as it observes a second,
+// distinct turn_context.model: an ambiguous head cannot safely name one model,
+// so it falls back to the honest unpriced floor rather than guessing the first.
+// The scan is bounded by codexModelSeedScanCap, so a long session never re-reads
+// its whole transcript; a switch that occurs after the cap but before the tail
+// window is not observable here and remains seeded with the first model.
 func codexHeadModel(r io.ReadSeeker) (string, error) {
 	if _, err := r.Seek(0, io.SeekStart); err != nil {
 		return "", err
 	}
 	scanner := bufio.NewScanner(io.LimitReader(r, codexModelSeedScanCap))
 	scanner.Buffer(make([]byte, 256*1024), 256*1024)
+	seed := ""
 	for scanner.Scan() {
 		line := scanner.Bytes()
 		if len(line) == 0 {
@@ -378,13 +388,25 @@ func codexHeadModel(r io.ReadSeeker) (string, error) {
 		if err := json.Unmarshal(entry.Payload, &payload); err != nil {
 			continue
 		}
-		if payload.Model != "" {
-			return payload.Model, nil
+		if payload.Model == "" {
+			continue
+		}
+		if seed == "" {
+			seed = payload.Model
+			continue
+		}
+		if payload.Model != seed {
+			// A genuine mid-rollout switch within the head window: the head cannot
+			// name a single session model, so seed empty and let pricing fall back
+			// to the unpriced floor instead of a possibly-wrong first model.
+			return "", nil
 		}
 	}
 	// A truncated final line at the cap boundary is expected and tolerated (same
-	// posture as splitLines); scanner.Err() is intentionally ignored.
-	return "", nil
+	// posture as splitLines); an oversized pre-turn_context line likewise stops
+	// the scan early and degrades to an empty seed (the pre-seed behavior).
+	// scanner.Err() is intentionally ignored.
+	return seed, nil
 }
 
 // ExtractCodexTailUsageFromSearchPaths reads codex tail usage only after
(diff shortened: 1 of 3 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277179446

quad341 (dismissed):
@julianknutsen — the diagnosis is the best part of this: "17 `turn_context` lines in a 7.9 MB rollout, 0 in the last 64 KiB, 8 `token_count` invocations stranded with `Model: \"\"`" is exactly how an intermittent "burn reads $0" complaint becomes a one-line root cause. Keeping the in-window `turn_context` override so the existing mid-session-switch tests still mean something was the right instinct too.

One adjustment I'll make on my side rather than send back to you. Seeding from the *first* `turn_context` means a session that switches models — and codex does offer that mid-session; we have code in `internal/runtime/tmux` whose whole job is dismissing its downgrade modal — then runs one long turn would price the new model's work at the old model's rate. Confidently wrong is harder to notice than empty. I'm flipping the scan to search backward from the tail window for the nearest preceding `turn_context`, so it's either right or empty. Your test passes unchanged; I'll add one for the switch case.

The red check is `internal/config/TestSkipRevisionSnapshot_StillDetectsPackContentChange` — pack revisions, nowhere near sessionlog. Ours to rebase and re-run. Nothing needed from you. Thanks for chasing this to the actual bytes.


---


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277246707

quad341 (dismissed):
Re-reviewed `312f3a226`..`abb5c60e2` (150 new commits since the last review): no new concerns — the prior review stands.

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277291621

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277291696

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR fixes usage pricing for
long-lived codex sessions: once a rollout grows past the tail window, the
`turn_context` line that names the model scrolls out of that window, so
`ExtractCodexTailUsage` minted empty-model, unpriced rows. Seeding the tail
model from the nearest preceding `turn_context` makes exactly those sessions
price correctly — which matters because the result is sticky: the idempotency
key means a session that lands unpriced stays unpriced.

PR branch.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **Window boundary.** A `turn_context` line straddling the seed/tail boundary
  was consumed by neither scan. The `io.LimitReader` allowance plus the
  `start >= windowStart` break now let the seed scan read the straddling line
  whole while handing every line that *starts* in the window back to the
  in-window scan — no gap and no double-count.
- **One file-size snapshot.** The seed scan and the tail read each took their
  own `Seek(0, io.SeekEnd)`, so two adjacent halves of a growing file could
  disagree about its size. A single snapshot now feeds both
  `codexPrecedingModel` and `readTailAt`.
- **Honest empty seed.** On oversized-line truncation the scan could surface a
  possibly-stale model. `codexSeedScanLineMax` now matches `splitLines`'
  per-line buffer, and a truncated scan returns an empty seed rather than a
  stale one.
- **Complexity.** The changed function was carried back under the repo's
  cyclomatic-complexity gate, which is green on the final head.

Review findings addressed:
- All three of the above were fixed in
  `28d2f40b fix(sessionlog): apply review fixes to the codex tail model seed`
  and re-validated on the second pass, pinned by new regressions
  `TestExtractCodexTailUsageSeedsModelStraddlingWindowBoundary` and
  `TestCodexPrecedingModelUsesCallerWindow` — the latter pins the boundary
  contract directly rather than only through the public entry point.

- `ga-8gku9` [Contract & Interface Fidelity, minor]: `codexPrecedingModel`'s doc
  promises `""` when the scan cannot be completed, but only `scanner.Err()`
  degrades that way — a `Seek`/`io.ReadFull` failure propagates out and fails
  the whole extraction. The seed I/O error policy should be decided and pinned
  with a test.
- `ga-yqsnc` [Behavioral Correctness, minor]: the oversized-line abort discards
  a seed it had already found; skipping over-long lines instead would keep it.
- `ga-j2o7u` [Behavioral Correctness, minor]: one narrow duplicate-collapse ×
  model-seed boundary arrangement is not yet pinned by a test.
- `ga-emblt` [Change Impact / Blast Radius, minor]: the backward seed scan runs
  unconditionally, including when the tail window already answers the question;
  making the seed lazy would avoid the work.
- `ga-6u1sd` [Architectural Consistency, minor]: `ExtractCodexTailMeta` keeps
  the old tail-only scan, so model *reporting* and model *pricing* can disagree
  for the same session. Pre-existing, not introduced here.

None of these are gating, and each is filed and open so nothing is lost.

### Maintainer Changes
`fix(sessionlog): apply review fixes to the codex tail model seed` — 3 files
changed, 393 insertions(+), 77 deletions(-). The branch was also rebased onto
the latest `main`; the adopted patch is unchanged by that rebase.

### Final Review Status
Ready for the merge queue: final head `28d2f40b` has passing required GitHub

CI: https://github.com/gastownhall/gascity/actions/runs/33806615166

### Review Iterations
2 review passes performed. The first returned `request_changes` with three
required changes; the second confirmed all three were satisfied and approved.

---

quad341 (approved):
Re-reviewed `abb5c60e2`..`28d2f40b4` (115 new commits since the last review): no new concerns — the prior review stands.

PR #5350, opened 2026-08-17
docs(deps): record why the bd pin cannot move to a newer-numbered release

austinborn:
## Summary

`deps.env` has long carried an instruction to promote `BD_VERSION` off `v1.1.0` "the day" beads publishes a release with the `--if-assignee`/`--if-status` conditional-write flags. beads has since shipped `v1.2.1` and `v1.2.2`, both dated after those flags landed in [beads#5008](https://github.com/gastownhall/beads/pull/5008), so the pin looked promotable. It isn't. The reason is one the old comment couldn't have anticipated: a beads tag's number doesn't order its capabilities.

beads cuts stable hotfixes off the `v1.1.0` line, which branched three weeks before the flags landed. That line is linear, it never merged to main, and both `v1.1.2` and the newest **stable** release `v1.2.2` sit on it. Neither one carries `--if-assignee` at all. `v1.2.1` is the first release that clears beads#5008 and it does carry both flags, but GitHub marks it a prerelease, so it doesn't meet the published bar this pin holds the installable default to.

So the pin stays where it is, and this change records why. It rewrites the `BD_VERSION` rationale in `deps.env`, names the two conditions that gate the next attempt, and sweeps the now-falsified "no published release has the flags" claim out of the five other places that asserted it: the `BD_PREV_VERSION` floor note, three comments across `internal/beads`, and the integration test's skip-and-fatal message. That last one is the one to look at, since it's the only site a human reads at runtime. **Comments and messages only: every pinned value is byte-identical to main.**

```mermaid
gitGraph
   commit id: "v1.1.0 base"
   branch hotfix-line
   commit id: "v1.1.2 (no flags)"
   commit id: "v1.2.2 (no flags, STABLE)"
   checkout main
   commit id: "beads#5008 flags"
   commit id: "v1.2.1 (flags, PRERELEASE)"
   commit id: "main"
```

## Risk

There's none to CI behaviour. No pinned version, SHA, or workflow `BD_VERSION` block moves, so every job installs exactly the bd it installed before. If CI goes red here, it won't be from this diff.

<details> <summary>Verification, and the two conditions that gate promotion</summary>

### How each claim was checked

Capability was measured against the published tarballs rather than inferred from version numbers, which is the same method the original comment used for `v1.1.2`:

| Release | Published | GitHub state | `bd update --help` names `--if-assignee` |
|---|---|---|---|
| `v1.1.2` | 2026-07-26 | stable | no |
| `v1.2.1` | 2026-08-11 | prerelease | **yes**, with the exit-13 semantics |
| `v1.2.2` | 2026-08-15 | stable | no |

Ancestry, via the compare API against `gastownhall/beads`:

- The hotfix line is linear and rooted before the flags: [`8e4e59d39`](https://github.com/gastownhall/beads/commit/8e4e59d39f3459a43cf21a3236a13eca4dd874f7) (2026-07-04) → +4 → `v1.1.2` → +7 → `v1.2.2`.
- Neither `v1.1.2` nor `v1.2.2` is an ancestor of beads `main`; each holds commits main does not.
- `v1.2.1` **is** an ancestor of beads `main` (main is 9 commits ahead, `v1.2.1` behind by none), and [`6e8af8bf8`](https://github.com/gastownhall/beads/commit/6e8af8bf8) is an ancestor of `v1.2.1`.

### Two conditions gate the next attempt

1. beads publishes a non-prerelease release cut from main, and its per-platform tarball SHAs go into `.github/scripts/install-bd-archive.sh`.
2. The fresh managed-Dolt bootstrap fix in [#5294](https://github.com/gastownhall/gascity/pull/5294) lands first. bd 1.2 added a legacy-workspace guard that refuses a pre-existing `.beads/dolt` carrying no version witness, so a fresh city under a 1.2 pin fails init until the witness is written for a proven-fresh database. That PR is open as of this writing.

Until both hold, the only flag-capable bd CI can obtain stays the source-built `BD_CURRENT_REF` cell, which is where the `ReleaseIfCurrent` CAS contract already runs.

### Every site the claim appeared in

| File | Site | Action |
|---|---|---|
| `deps.env` | `BD_VERSION` rationale | rewritten |
| `deps.env` | `BD_PREV_VERSION` floor note | corrected |
| `internal/beads/bdstore_conditional_release.go` | package comment | corrected |
| `internal/beads/bdstore.go` | `ReleaseIfCurrent` doc comment | corrected |
| `internal/beads/bdstore_conditional_release_integration_test.go` | test doc comment | corrected |
| `internal/beads/bdstore_conditional_release_integration_test.go` | skip and fatal message | corrected, runtime-visible |

Three further sites were checked and deliberately left alone, each re-verified as still true: the `.github/workflows/ci.yml` comment saying `BD_VERSION` "is a published release that predates it"; `bdstore_test.go`; and the graduation test, which is about `--if-revision`, a different flag absent from both `v1.2.1` and `v1.2.2`.

### Checks run

- `sh -n deps.env`, and sourcing it to confirm `BD_VERSION` and `BD_PREV_VERSION` read back unchanged.
- `diff` of the non-comment lines against `origin/main`: identical.
- `gofmt -l`, `go build ./internal/beads/...`, `go vet ./internal/beads/...`: all clean, before and after the round-1 sweep.
- Test packages under `internal/beads` compile; nothing asserts on any string touched.
- `git diff --check`: clean.

</details>

---
> • On behalf of: @austinborn


austinborn:
Round 1 of internal review is in, and it changed the diff.

The reviewer re-derived every factual claim here independently rather than taking the summary's word for it: the release topology, the prerelease status of `v1.2.1`, and whether [#5294](https://github.com/gastownhall/gascity/pull/5294) is still open. They all hold.

It also caught that the sweep was incomplete. Three more sites in `internal/beads` still asserted the claim this change exists to correct, and one of them isn't a comment at all: it's the integration test's skip-and-fatal message, so it's the only site a human reads at runtime, whenever that row skips or fails under `GC_REQUIRE_BD_CONDITIONAL_RELEASE`. All three are fixed in [`5f14957`](https://github.com/gastownhall/gascity/commit/5f14957cc02606f05ee9eff03bc11cb3d76f392f), and the summary now lists every site it touches instead of the two it named before.

Three adjacent sites were checked and deliberately left alone, each re-verified as still true. The graduation test is the interesting one: it reads as the same claim but it's about `--if-revision`, which neither `v1.2.1` nor `v1.2.2` carries.

Still comments and messages only. No pinned value, identifier, or control flow moves.

---
> • On behalf of: @austinborn

quad341 (approved):
@austinborn — this is the good kind of documentation PR. You didn't just fix a wrong comment, you found the *reason* it was wrong: a beads tag's number doesn't order its capabilities, because stable hotfixes come off the `v1.1.0` line while the flags live on main. "Qualify a candidate with `bd update --help` on its own tarball, never with a version comparison" is a rule that will save someone from a bad promotion months from now.

I re-derived it before merging rather than taking the writeup's word for it — `v1.2.1` is a prerelease, `v1.2.2` is the newest stable and is flagless, and the ancestry runs the way you describe. It all holds. Sweeping the falsified claim out of all five downstream sites, especially the test's skip-and-fatal message, was the right call; that message is the only one a human meets at runtime.

Merging as-is. One thing moved under you while this sat: #5294 landed on 2026-08-31, so "today neither condition holds" is now half-true. Your conclusion is unchanged — the pin still can't move, no non-prerelease cut from main yet — so I'll refresh that paragraph on main myself in a follow-up rather than send you back for two sentences. Ignore the red container scan; that's base-image drift, nothing to do with you. Thanks for this one.


---

Issue #5054, opened 2026-08-06
All 45 container-scan waivers expire tomorrow (2026-08-07)


bourgois:
@julianknutsen — flagging because you've been working `.trivyignore.yaml` this week (`48339482c`, `2e8fe2f4b`, `824d4ea2a` on 08-05) but the horizon itself is untouched, and I can't tell whether that's deliberate.

## The facts

- Every entry in `.trivyignore.yaml` on `main` carries `expired_at: 2026-08-07` — **45 entries, 28 unique CVEs, expiring tomorrow**.
- The scan runs `--severity HIGH,CRITICAL --exit-code 1`, so on lapse the container-scan gate goes red on every PR, not just image-touching ones.
- This would be the third horizon (2026-06-12 → 07-07 → 08-07; the last move was your #4002).
- The file says the durable fix is tracked in `ga-frh27v` — that ID doesn't resolve in the beads DB we can see, so from the outside the durable work looks unowned. Possibly it lives in a DB we don't share.
- #4920 would have cleared six of the 28 via the beads bump, but it was closed on the tagged-releases-only rule, so nothing currently in flight moves the count before the deadline.

## The question

**Is the expiry intentional** (durable fixes landing before Thursday's runs, or a deliberate forcing function), **or does the horizon need a decision today?**

If it's an extension: the constraint worth preserving from your own file structure is that `gh`/`dolt`/`bd` are rebuilt-from-source and deliberately *not* waived for the Go-stdlib batch — a bulk extension that accidentally widens onto a rebuilt path would blind exactly the regression signal the design protects.

## Offer

We maintain a fork that hit the same findings and we have current evidence per path (which CVEs are fixable by rebuild-with-patched-module vs genuinely unfixable external prebuilts — `kubectl` pins `x/text v0.33.0` through v1.36.3, so your `824d4ea2a` waiver matches our conclusion exactly). Happy to write up the per-path table or draft the extension PR under the ga-85ytyr gate checklist if useful. Otherwise this is just a heads-up from someone who'd rather not watch every PR go red on Thursday.

austinborn:
The horizon lapsed, and Container Scan has been red on every branch since 2026-08-09, `main` included. [PR #5353](https://github.com/gastownhall/gascity/pull/5353) answers the question in this issue with a concrete proposal rather than a fourth extension, and the check is green on it.

[...]

PR #5353, opened 2026-08-17
fix(container-scan): rebuild gh and Dolt with patched x/ modules, trim the waiver set

austinborn:
Container Scan has failed on every branch since 2026-08-09, `main` included, and it isn't about anyone's diff: all 45 entries in `.trivyignore.yaml` expired on 2026-08-07. [#5054](<https://github.com/gastownhall/gascity/issues/5054>) flagged the horizon the day before it lapsed and it's still open. The check is path-gated and not required, so `main` stayed mergeable while red, and the red only surfaces on PRs that touch the filter. That's what makes a repo-wide failure read as a per-PR regression.

Re-dating the file wouldn't have been enough. Seven of the findings are newer than the file and weren't covered by any entry.

There's a second, unrelated red underneath. The mail image carries nine HIGH findings, all CVE-2026-53615 in util-linux, and they're new rather than newly visible: the SARIF step runs Trivy without `--exit-code`, so it scans all four images on every run, and only the enforcement step stops at the first failure. The artifacts show it directly — `gc-mcp-mail.sarif` from run 31909222974 on `main` (08-15) has 0 results, and run 32025841203 (08-17) has 9. The CVE was disclosed between those two dates. That one's fixed by naming the affected packages in the upgrade list `Dockerfile.mail` already keeps for its digest-pinned base.

The rebuilds are where the rest gets fixed. `Dockerfile.base` already forces `google.golang.org/grpc` forward in the gh and Dolt builds, so this puts `golang.org/x/crypto`, `golang.org/x/net`, `golang.org/x/text` and `github.com/apache/thrift` alongside it, each with the matching `go version -m` assertion. That's all 16 findings on a rebuilt path. `KUBECTL_VERSION` moves v1.36.0 to v1.36.3, taking its Go stdlib from 1.26.2 to 1.26.5 and clearing 8 of its 22. kubectl's an upstream-signed prebuilt this repo installs rather than builds, so the 12 CVE ids no released kubectl fixes keep waivers, on a 2026-11-07 horizon. bd and br don't produce any findings now, so they leave the file too: 45 entries down to 12, all of them kubectl.

A later round covers two `golang.org/x/mod` CVEs published after the above was written. They land in `usr/bin/gh` and in `usr/local/bin/gc`, so the gh stanza takes an x/mod override alongside its existing ones and `go.mod` pins the module forward for gc. Taking x/mod to v0.40.0 drags x/text, x/net and x/crypto up with it, so those three ARGs rise to the versions `go.mod` now carries. [The round comment](<https://github.com/gastownhall/gascity/pull/5353#issuecomment-5350419110>) has the per-binary measurements.

| Path | Findings today | After this change |
| -- | -- | -- |
| `usr/bin/gh` | 3 | 0, x/text forced to 0.41.0 and x/mod to 0.40.0 |
| `usr/local/bin/gc` | 2 | 0, x/mod pinned forward in `go.mod` |
| `usr/local/bin/dolt` | 15 | 0, x/crypto, x/net, x/text and thrift forced forward |
| `usr/local/bin/kubectl` | 22 | 8 fixed by the pin bump, 14 waived |
| `usr/local/bin/bd`, `usr/local/bin/br` | 0 | 0, 26 stale entries dropped |
| mail image, Debian packages | 9 | 0, the util-linux set upgraded |

```mermaid
flowchart LR
  subgraph R["Rebuilt from pinned source — fixable here"]
    GH["gh"] --> F["go get forces the patched module,<br/>go version -m proves it landed"]
    DOLT["Dolt"] --> F
    BD["bd"] --> F
  end
  subgraph E["Installed as a signed prebuilt — not fixable here"]
    K["kubectl"] --> W["Waiver, until a Kubernetes release<br/>ships the patched dependency"]
  end
  F --> G["Scan green"]
  W --> G
```

The risk sits in the Dolt build. thrift moves from a 2020 pre-release to 0.23.0, which is a real jump, though it's an indirect dependency there and bd already builds against 0.23.0. If any override doesn't land, the `go version -m` assertion fails the image build rather than letting a vulnerable binary through quietly.

<details><summary>Evidence, test plan and the reasoning behind each number</summary>

### Where the numbers come from

Everything above comes from the SARIF and CycloneDX artifacts of the last scheduled `main` run ([run 31909222974](<https://github.com/gastownhall/gascity/actions/runs/31909222974>)), and it isn't a local scan. Those artifacts come out of the same `trivy image` invocation as the enforcement step, with the same `--severity HIGH,CRITICAL --ignore-unfixed --ignorefile .trivyignore.yaml`, so the finding set matches what turns the gate red. The enforcement loop stops at the first image, which is why the log alone doesn't show more than `gc-agent-base`. The artifacts cover all four.

`gc` didn't have findings when this was written, and two `golang.org/x/mod` CVEs disclosed later put it back in scope. It isn't rebuilt from pinned source, so no Dockerfile override reaches it: the floor is `go.mod`'s own pin, which now carries x/mod v0.40.0 plus the x/text v0.41.0, x/net v0.58.0 and x/crypto v0.55.0 that came with it.

### kubectl, in detail

v1.36.3 is the newest release on any 1.36 line and the newest kubectl on a patched Go stdlib, which `go version -m` against the published binaries confirms: v1.36.0 is go1.26.2, v1.36.3 is go1.26.5. The x/ modules don't move. Both still carry x/net v0.49.0 and x/text v0.33.0, so those findings stay.

Of the 22 findings, 8 clear at Go 1.26.5, and the 14 that remain resolve to 12 CVE ids. One id is subtle. CVE-2026-33814's stdlib instance clears at 1.26.5, but the same id also fires against x/net v0.49.0, which the pin bump doesn't move, so the entry stays.

The remaining stdlib CVEs want Go 1.26.6 or 1.25.13 and there isn't a Kubernetes release shipping either yet, which is why the horizon is three months rather than the one month the previous three extensions used. Nothing left in the file's fixable from this repo, so a shorter one buys a re-dating chore instead of a decision.

### Test plan

* `go test ./scripts/` and `go vet ./...` pass.
* Both Dockerfile rebuilds were reproduced locally from the checksum-verified pinned source tarballs, with the overrides applied, and both compile. The x/mod round raised three of those pins, so these are the numbers `Dockerfile.base` asserts today: the rebuilt Dolt binary embeds thrift v0.23.0, x/crypto v0.55.0, x/net v0.58.0 and x/text v0.41.0, and the rebuilt gh embeds grpc v1.82.1, x/text v0.41.0 and x/mod v0.40.0. Every `go version -m` assertion matches real output rather than a guess at it. That was the part worth checking: thrift is a five-year jump. One caveat on the Dolt half sits in [the round comment](<https://github.com/gastownhall/gascity/pull/5353#issuecomment-5350419110>): that rebuild ran under host CGO with `-tags=timetzdata` instead of the musl static cross-build, so it proves the module set compiles and the assertions hold, not the static-link configuration.
* `TestTrivyIgnoreDropsStdlibWaiversForRebuiltTools` needed updating. It required x/text waivers on gh and Dolt to be present, which is exactly what the overrides make obsolete, and its check widened from gh alone to every rebuilt path.
* `TestRebuiltToolsForcePatchedXModules` is new. It asserts both halves of each override, the `go get` and its `go version -m` proof, so you can't drop either one silently. Each get is matched inside its own build stanza, since gh and Dolt spell the x/text override identically and a file-wide search let one stand in for the other. It also pins the order inside the gh stanza, x/mod after x/text. Mutation-checked three ways: dropping the Dolt x/text get fails on `/out/dolt`, dropping the gh x/mod get fails on `/out/gh`, and swapping the two gh gets fails on the ordering assertion alone.
* Both guards were mutation-checked. Removing the Dolt x/net assertion fails the new test on that exact line, and re-adding a Dolt x/text waiver fails the updated one naming that path.
* `TestMCPMailImageUpgradesPatchedOSPackages` is new too, and mutation-checked the same way: dropping `libuuid1` from the Dockerfile fails it by name. Trivy reports every binary package of a source separately, so a partial list leaves one package vulnerable while the other eight read as a whole fix.
* This PR touches `contrib/k8s/**` and `.trivyignore.yaml`, both inside the workflow's path filter, so Container Scan runs on it. That run's the confirmation.

### Not addressed

The scan'll go red again whenever a new HIGH or CRITICAL lands in kubectl or in the mail image's Debian base, and because it isn't required, `main` can absorb that quietly. The util-linux finding here is that exact case: it was disclosed after the waiver file was last touched. Making it required is a separate decision and belongs with the horizon question in [#5054](<https://github.com/gastownhall/gascity/issues/5054>).

</details>

---

> Generated by the operator's software factory.<br>• City: `city` · Agent: `local-core.builder-4`<br>• On behalf of: @austinborn


austinborn:
This round clears the two `golang.org/x/mod` findings disclosed after the rest of this work was written. It needed a second half the original scope didn't have.

The scan artifacts from the last run put `golang.org/x/mod` v0.37.0 in two binaries rather than one: `usr/bin/gh`, rebuilt here from pinned source, and `usr/local/bin/gc`, built straight from this module. Clearing gh alone would have left the scan red on gc, so both are here. gh gets an `XMOD_VERSION` override in the stanza that already forces grpc and x/text, with its matching `go version -m` assertion. gc has no Dockerfile knob, so its floor is `go.mod`'s own pin, guarded by a new test beside the ones covering the rebuilt tools.

Order inside the gh stanza is load-bearing, and the obvious order is wrong. Put the x/mod get ahead of the x/text one and x/text rolls x/mod back to v0.38.0, under the fixed version, while every assertion still passes because none of them names x/mod. That's measured against the pinned source, not reasoned about. So x/mod goes last. Taking it to v0.40.0 pulls x/text to 0.41.0, which requires x/net 0.58.0, which requires x/crypto 0.55.0, so those three ARGs rise with it and land on the versions this repo's `go.mod` now carries, which is the invariant the comment above them already claimed. Dolt takes no x/mod override, because no x/mod package is linked into its binary, which is also why it has no finding.

One thing to know before merging: [#5411](https://github.com/gastownhall/gascity/pull/5411) already makes the same one-line `go.mod` bump for unrelated reasons. It doesn't touch the Dockerfile, so it doesn't fix gh, and this branch can't reach green without both halves. Whichever merges first turns the other's `go.mod` hunk into a no-op.

<details>
<summary>What was verified, and what could not be</summary>

There's no Docker daemon on the build host, so the image can't be built and rescanned locally. Everything below is the binary half, run against the checksum-verified pinned sources the Dockerfile names.

- gh, built from the pinned source with `go run script/build.go bin/gh` after the three gets in the committed order: `go version -m` reports grpc v1.82.1, x/text v0.41.0 and x/mod v0.40.0, so all three assertions pass against real output. The binary runs and reports `gh version 2.96.0`.
- The wrong order was run first, as a control. It selects x/mod v0.38.0 with the x/text assertion still passing, which is the silent failure the ordering comment describes.
- Dolt, built from its pinned source with the raised override set: `go version -m` reports x/crypto v0.55.0, x/net v0.58.0, x/text v0.41.0 and thrift v0.23.0, so its four assertions pass too. That build used host CGO with `-tags=timetzdata` rather than the musl static cross-build, so it proves the module set compiles and the assertions hold, not the static-link configuration.
- gc, built with `CGO_ENABLED=0 go build ./cmd/gc` as the scan workflow does: `go version -m` reports x/mod v0.40.0.
- `go build ./...` and `go test ./scripts/` pass. One unrelated pre-commit-hook test failed once in a full-package run and passed both in isolation and on a clean re-run, so it reads as a flake rather than a regression.
- The finding set came from the SARIF artifacts of the failing run rather than the job log, since the enforcement loop is fail-fast and stops at the first image. Those artifacts are also what rules Dolt out.

</details>

---
> • On behalf of: @austinborn

austinborn:
Round two. The review left one item on the body and two files in the tree. Both are in, and re-measuring the ordering claim turned up a correction that's worth making here too.

**The ordering rule is real, and the x/text bump had already disarmed it.** The comment above says putting the x/mod get ahead of the x/text one rolls x/mod back to v0.38.0. Measured again against the checksum-pinned gh source with go1.26.6, that holds at the *previous* `XTEXT_VERSION` of 0.39.0: `go get golang.org/x/mod@v0.40.0` drags x/text up to v0.41.0, so the x/text get that follows it is a downgrade, and it takes x/mod down to v0.38.0. At the 0.41.0 this branch ships, that get isn't a downgrade any more, and both orders select x/mod v0.40.0 with x/text v0.41.0.

Raising the x/text pin is what removed the trap. The ordering convention is what stops the next module bump from recreating it, and the guard's comment now says that rather than describing a rollback which doesn't reproduce at today's pins.

**What round two commits.**

- `scripts/container_tool_security_test.go` asserts the x/mod get follows the x/text one, scoped to the `WORKDIR /src/gh` slice so it can't match the Dolt stanza. Swapping the two gets fails on that assertion alone.
- Two stale comments. `Dockerfile.base` said the ARG values match what this repo's `go.mod` and the rebuilt bd both carry; `go.mod` matches all five exactly, bd doesn't, since its stanza takes the grpc override alone. The test called the same list the minimum versions Trivy names as fixed, which after the bump sits above the floor for x/crypto and x/net.
- The body's test plan named x/crypto v0.53.0, x/net v0.56.0 and x/text v0.39.0 for the Dolt rebuild, three versions `Dockerfile.base` doesn't assert any more. It carries the asserted set now, and it points here for the caveat on that build.

<details>
<summary>The gh half, re-run rather than inherited</summary>

Fetched the pinned tarball fresh, checked its sha256 against `GH_SOURCE_SHA256`, and built with the three gets in the committed order.

- `go version -m bin/gh` reports grpc v1.82.1, x/text v0.41.0 and x/mod v0.40.0, so all three assertions pass against real output. The binary reports `gh version 2.96.0`.
- The reversed order was run as a control at both pins: v0.40.0 at 0.41.0, v0.38.0 at 0.39.0.
- `go build ./...` and `go test ./scripts/` pass, the latter including the four container-security guards this branch touches.

The image half is still unverified locally, since this host has no Docker daemon. Container Scan on this PR is the confirmation.

</details>

---
> • On behalf of: @austinborn

quad341:
@austinborn — circling back on this with an apology for the silence: this PR produced the only green Container Scan this repo has had in its last 60 runs (2026-08-20), and it went unacknowledged while the scan stayed red for everyone. That's on us, and it's part of a backlog push we're doing this week.

State of the world since you branched: the waivers all expired on 08-07 (which is why the gate is red on every branch), and today we merged dependency bumps (#5411, #5856) covering the repo's own modules. Your PR remains the right durable fix for the other half — the vendored gh/dolt binaries — and we'd like to land it rather than reproduce it. It's currently conflicting after 12 days of drift.

If you're still up for it: a rebase onto current main and we'll prioritize review immediately — the waiver-trim portion may shrink now that a short-horizon refresh is landing separately (deliberately marked to expire 2026-09-21 so your fix stays the real answer). If you'd rather hand it off, say the word and we'll rebase it ourselves with your authorship preserved. Either way, thank you — this was the correct diagnosis before we'd even made it.

quad341 (changes requested):
@austinborn — you diagnosed this correctly and ahead of us: 45 waivers expired 2026-08-07, a repo-wide red reading as a per-PR regression, and re-dating wouldn't have covered the seven findings newer than the file. I confirmed `main` is still in exactly that state, so this PR is still the fix.

Two things worth telling you. Your caveat about the Dolt rebuild is obsolete in your favor — Container Scan `docker build`s `Dockerfile.base` directly, and `Image vulnerabilities` went **green** on this head, so the thrift 0.13.1 → 0.23.0 jump, the static-link check and all five `go version -m` assertions are CI-verified. And the red `rest-smoke-2-of-2` is `TestGraphWorkflowSuccessPath` dying on a beads schema migration (beads#4566) — not yours.

I'd normally rebase this myself. I can't: your fork is org-owned (`actual-software`), and GitHub refuses maintainer pushes to those regardless of "Allow edits by maintainers." So I'm asking you to run it, and it's four lines — `main` already landed identical pins, so your `go.mod`/`go.sum` hunks are now no-ops:

```bash
git fetch origin main
git merge origin/main          # only go.sum conflicts
git checkout origin/main -- go.mod go.sum
go mod tidy && git add go.mod go.sum && git commit && git push
```

I verified the merged `go.mod` is byte-identical to `main`'s, so this discards nothing of yours. Leave `.trivyignore.yaml` and the Dockerfiles exactly as they are — no conflict there. Ping me when it's pushed and I'll merge on green.


---

austinborn:
@quad341 — pushed. Your four-line recipe turned out to be three lines already done and one still needed, so here's what's actually in the tree.

`go.mod` and `go.sum` are byte-identical to `main`'s and no longer appear in this PR's diff at all, which is exactly the end state you described. `go mod tidy` is a no-op on them. I merged current `main` on top; no conflicts.

What's left is the five files you asked me to leave alone, unchanged: `.trivyignore.yaml`, the three Dockerfiles, and `scripts/container_tool_security_test.go`. 251 insertions against 278 deletions.

**`Image vulnerabilities` is red, and it's worth knowing it isn't a finding.** The Dockerfile.base build died on a module-proxy read, not on a CVE:

```
go: google.golang.org/genproto@v0.0.0-20250505200425-f936aa4a68b2:
  read "https://proxy.golang.org/...": stream error: stream ID 355; INTERNAL_ERROR
```

That's transient, so a re-run should clear it. None of the `go version -m` pin assertions were reached, which means the thrift and `x/` pins this PR exists to prove weren't evaluated either way on this run.

**The other reds are upstream.** `main` is broken at [`3d268c5`](https://github.com/gastownhall/gascity/commit/3d268c5849ca2ef49392c6e4adb68e9a10d0b59d) with 31 failing checks including `CI / required`, because `releaseOrphanedPoolAssignments` gained a parameter in [#6261](https://github.com/gastownhall/gascity/pull/6261) while six call sites in `cmd/gc/pool_session_name_route_assignee_liveness_test.go` still pass the old signature. CI builds each PR merged into main's tip, so any PR re-running its checks right now goes red on it. A PR that hasn't re-run since main broke still shows its older green.

---
> • On behalf of: @austinborn

austinborn:
@quad341 — merged current `main` and the branch is now green everywhere except one check, which I don't think this PR can clear on its own. Details below, and there's a question for you at the end.

**Where it stands.** 47 checks pass, 32 skip, 1 fails. The previous merge here sat on [`3d268c5`](https://github.com/gastownhall/gascity/commit/3d268c5849ca2ef49392c6e4adb68e9a10d0b59d), which was broken: `releaseOrphanedPoolAssignments` gained a parameter in [#6261](https://github.com/gastownhall/gascity/pull/6261) while six call sites in [`cmd/gc/pool_session_name_route_assignee_liveness_test.go`](https://github.com/gastownhall/gascity/blob/1b991d75c/cmd/gc/pool_session_name_route_assignee_liveness_test.go) still passed the old signature. That reddened every check compiling `cmd/gc`. Merging [`1b991d75c`](https://github.com/gastownhall/gascity/commit/1b991d75c) cleared all four of those.

`go.mod` and `go.sum` are still byte-identical to `main`'s and still absent from this PR's diff. `go mod tidy` is a no-op on them. The diff is the same five files at 251 insertions against 278 deletions.

**The remaining red is three findings, and they postdate this work.** `Image vulnerabilities` fails on:

| Finding | Where | Installed | Fixed |
|---|---|---|---|
| CVE-2026-84304 (grpc) | `usr/bin/gh`, `usr/local/bin/gc` | v1.82.1, v1.83.0 | 1.83.1 |
| CVE-2026-43871 (thrift) | `usr/local/bin/dolt`, `usr/local/bin/bd` | v0.23.0 | 0.24.0 |
| CVE-2026-78675/6/7 (GitPython) | `gc-mcp-mail` | 3.1.58 | 3.1.59 |

Each one lands on a version this PR moved *to*. The thrift bump from 0.13.1 to 0.23.0 is this PR's work, and 0.23.0 picked up its own CVE afterwards. `main` doesn't waive any of the three either, so the gate is red repo-wide rather than red on this branch.

**Your [#5885](https://github.com/gastownhall/gascity/pull/5885) already covers all three**, time-boxed to 2026-09-21, and its thrift entry names this PR as the durable fix. It got a green Container Scan today.

**Why I left them alone rather than fixing them here.** `gc` takes grpc from `go.mod`, and putting a bump there would undo the byte-identical state you asked for. `bd` is installed as a released binary, so its thrift pin isn't reachable from this repo. GitPython sits in a requirements file this PR doesn't touch. And adding the waivers here would collide with #5885 on `.trivyignore.yaml`, which both PRs rewrite.

**So the question is sequencing, and it's yours.** Land #5885 and I'll merge `main` again, which should take this branch green with no further changes. Or say the word and I'll absorb the three waivers into this PR instead and you can drop them from yours. I'd rather not guess which, since both touch the same file.

Two nits from your review, unchanged and still yours to take or leave: the "91 checks" comments, and the `--check <name> --fix` ordering question.

---
> • On behalf of: @austinborn

austinborn:
Correcting my own comment above on two counts before you act on it. The recommendation at the end of it is unchanged; the findings table and one of the two options I offered were both wrong.

**The table missed a CVE and understated another's reach.** The scan reports six ids, not the five my table implied:

| Finding | Where | Installed | Fixed |
|---|---|---|---|
| CVE-2026-84304 (grpc) | `usr/bin/gh`, `usr/local/bin/dolt`, `usr/local/bin/bd`, `usr/local/bin/gc` | v1.82.1, v1.83.0 | 1.83.1 |
| CVE-2026-84445 (grpc) | same four binaries | v1.82.1, v1.83.0 | 1.82.2, 1.83.2 |
| CVE-2026-43871 (thrift) | `usr/local/bin/dolt`, `usr/local/bin/bd` | v0.23.0 | 0.24.0 |
| CVE-2026-78675/6/7 (GitPython) | `gc-mcp-mail` | 3.1.58 | 3.1.59 |

CVE-2026-84445 was absent from my table altogether, and CVE-2026-84304 reaches all four Go binaries rather than the two I listed.

**Withdrawing the offer to absorb the waivers here.** That option can't be executed as I wrote it. [`TestTrivyIgnoreDropsStdlibWaiversForRebuiltTools`](https://github.com/gastownhall/gascity/blob/75a885f99/scripts/container_tool_security_test.go#L359-L361) fails on *any* waiver naming `usr/bin/gh`, `usr/local/bin/dolt` or `usr/local/bin/bd`, whatever the CVE, unless the entry is added to `reviewedWaivers`. Both grpc ids and the thrift id need exactly those paths. So absorbing them here means either weakening the guard this PR exists to add, or extending `reviewedWaivers` past the single reviewed kubectl entry. Only `usr/local/bin/gc` and the GitPython findings are waivable on this branch without one of those.

That leaves one option rather than two: land [#5885](https://github.com/gastownhall/gascity/pull/5885), which waives all six ids on paths its own guard permits, and I'll merge `main` again to take this branch green. If you'd rather go the durable route on any of them instead, say which and I'll open it separately rather than growing this PR.

---
> • On behalf of: @austinborn

austinborn:
One more correction, and this one is to the sentence I asked you to act on. "Land #5885 and I'll merge `main` again to take this branch green with no further changes" doesn't hold. I measured the three trees after writing it, and the two PRs conflict.

**Why the two can't just stack.** Both rewrite the whole of `.trivyignore.yaml`: `main` carries 45 entries expiring 2026-08-07, this PR trims that to 12 expiring 2026-11-07 with no `gh`/`dolt`/`bd` paths at all, and [#5885](https://github.com/gastownhall/gascity/pull/5885) carries 64 expiring 2026-09-21 with 50 waivers on those three paths. Both also rewrite the same allow-list in [`scripts/container_tool_security_test.go`](https://github.com/gastownhall/gascity/blob/75a885f99/scripts/container_tool_security_test.go#L340-L361): this PR renames it `reviewedWaivers`, trims it to the kubectl entry and rejects every other waiver on a rebuilt path; #5885 renames it `allowedModuleWaivers` and adds the two grpc ids on all four binaries. And #5885's new `TestTrivyIgnoreRefreshesBridgeHorizonAndWaivesXNetDNSMessageCVE` asserts exactly 64 entries at that horizon. Whichever lands second takes a conflict on both files, and no resolution passes both suites as written: keep this PR's file and the count test fails with the six ids still unwaived; take #5885's file and this PR's rebuilt-path guard fails on all 50.

**The real choices, none of which I can promise green under:**

1. **This PR first, then rebase #5885 onto it**, re-adding only the ids still live as `reviewedWaivers` entries and dropping the 64-count assertion. Smallest diff to #5885, and the bridge keeps its horizon.
2. **#5885 first, then this PR takes the two-file conflict** and extends `reviewedWaivers` with CVE-2026-84304, CVE-2026-84445 and CVE-2026-43871 on the rebuilt paths, time-boxed. That is a deliberate widening of the guard this PR adds, so it wants your say-so rather than my judgement.
3. **The durable route, no waiver needed for the rebuilds.** `GRPC_VERSION` to 1.83.2 and `THRIFT_VERSION` to 0.24.0 in [`Dockerfile.base`](https://github.com/gastownhall/gascity/blob/75a885f99/contrib/k8s/Dockerfile.base#L22-L33), where the `go get` overrides and `go version -m` assertions already exist for both, plus a matching thrift override in `Dockerfile.agent`'s `bd-builder` stage. That clears the four Go binaries. It does not clear `gc`, whose grpc 1.83.0 comes from `go.mod`, or GitPython in the mail image, so those two still need either #5885's waivers or a bump on `main`.

I'd take 3 for the rebuilds if you're open to this PR growing by three ARG lines, since it's the same mechanism the PR already uses. Otherwise 1.

**And one correction to my earlier reasoning.** I said `bd`'s thrift pin wasn't reachable from this repo. It is: `Dockerfile.agent` builds `bd` from source at `BD_SOURCE_REF` and already applies a grpc `go get` override there, so thrift can be forced forward the same way. Option 3 above accounts for that.

---
> • On behalf of: @austinborn

austinborn:
@quad341, merged current `main` and pushed as [`908aec3b3`](https://github.com/gastownhall/gascity/commit/908aec3b31690488141660dc420f80612b34ca03); the branch reads mergeable again. You landed [#5885](https://github.com/gastownhall/gascity/pull/5885) first, so this is option 2 from my last comment: the PR takes the two-file conflict and widens its own guard by exactly the three ids that need rebuilt paths. That widening is the one judgement call here, and the thing to look at.

**What the merge keeps from #5885.** The scan on the previous head reported six ids, every one published against a version this PR moves to, and those six come across on #5885's own 2026-09-21 horizon: CVE-2026-84304 and CVE-2026-84445 (grpc) on gh, dolt, bd and gc, CVE-2026-43871 (thrift) on dolt and bd, and the three GitPython ids as purls. The other 58 bridge entries are findings the rebuilds clear, so they stay out. The carried statements now name the pin that has to move instead of naming this PR as the durable fix, which for those three it is not.

**In the test file**, `reviewedWaivers` grows by those three ids on the rebuilt paths, so every other rebuilt-path waiver is still rejected. Your four tests pinning the bridge's exact 64-entry shape assert the set this PR retires, so one test replaces them: [`TestTrivyIgnoreKeepsReviewedBridgeEntries`](https://github.com/gastownhall/gascity/blob/908aec3b31690488141660dc420f80612b34ca03/scripts/container_tool_security_test.go#L541-L666) holds the six survivors to their ids, paths or purls, horizon and statement contents. `release-gates/ga-2yq3p5-container-scan-waiver-bridge-gate.md` is untouched, as a dated record of #5885's gate; say so if you'd like it to carry a note.

Dockerfiles, `go.mod` and `go.sum` are untouched by the merge. CI is green on the new head: 48 checks pass and 32 skip on path filters, with [Image vulnerabilities](https://github.com/gastownhall/gascity/actions/runs/34682796834/job/103524435366) among the passes, so this merges on your approval.

<details>
<summary>The gc threshold, how the six were measured, and what ran locally</summary>

Your grpc thresholds for the gc path came across with CVE-2026-84445 keyed at 1.83.2, the fix Trivy names on the 1.83 line `go.mod` is on. Keyed at 1.83.1, `TestTrivyIgnoreDropsGCModuleWaiversPastThreshold` would demand dropping a gc path the scan still reports the moment `go.mod` reached 1.83.1.

The finding set is the SARIF from the Container Scan on the previous head (run 34653787829), since the enforcement loop is fail-fast and the job log stops at the first image. gc-agent-base carried the grpc pair on gh and dolt plus thrift on dolt; gc-agent and gc-controller added bd and gc for grpc and bd for thrift; gc-mcp-mail carried only the three GitPython ids. Nothing on `main` since the merge base touches an image input, so the merged tree's images are this branch's images.

Locally: `go build ./...`, `go vet ./scripts/` and `go test ./scripts/` pass on the merged tree. Five mutations of `.trivyignore.yaml` each fail the guard they should: a rebuilt path added to a kubectl entry, the bd path dropped from the thrift entry, a carried entry moved to 2026-11-07, a new kubectl entry borrowing the 2026-09-21 date, and a statement with its fixed version removed.

</details>

---
> • On behalf of: @austinborn

quad341 (approved):
@austinborn — merging this now. It's been open since August and it is the correct diagnosis of a repo-wide red that predated it: all 45 `.trivyignore.yaml` entries lapsed on 2026-08-07, and the failure only surfaced on PRs touching the path filter, which is what made a repo-wide break read as a per-PR regression.

You did the hard part twice — the original rebuild work, then re-sequencing after we landed #5885 out from under you. Widening `reviewedWaivers` by exactly the three ids that need rebuilt paths, each held to #5885's own 2026-09-21 horizon instead of quietly inheriting your three-month one, was the right read of a call we'd already made by merging in that order. Thank you for not just taking the wider date.

`Container Scan / Image vulnerabilities` is green on `908aec3b3`, so the thrift five-year jump, the static-link check and all nine `go version -m` assertions actually ran in a real build. Three small comment/doc nits I'll clean up on `main` myself, nothing for you to do — plus a tracker for the thrift 0.24.0 and grpc 1.83.2 bumps before 09-21 so that fuse doesn't burn down quietly.


---

PR #5354, opened 2026-08-17
feat(doctor): run a named subset of checks with --check

austinborn:
`gc doctor` had no way to ask for one check. Every invocation ran the whole registry (93 checks on the city this was measured against), so a caller that needed a single verdict paid for the entire sweep: 35.1s and 47.55s across two runs, dominated by `order-firing-current` doing two bead-store round-trips per order across 46 orders. That cost scales with order count and store size, and the caller doesn't control either.

`--check` names the checks to run. It's repeatable, it takes a comma list, results keep registration order rather than the order you asked in, and the exit code comes from the selected checks alone. On the same city, `gc doctor --check controller --json` comes back in 0.38s.

A name that matches no registered check fails the whole run instead of reporting the subset that did match. That's as much the point of the flag as the speed is: a caller filtering doctor's output by name reads a short result set as a clean one, so one typo would otherwise report health nobody measured. The error names what didn't match, suggests near matches, and lists what this workspace registers, on stderr in text mode and in `--json` as a `registered_checks` array beside the `error` field that struct already carried and never populated. Both are `omitempty`, so a successful `--json` report stays byte-identical to today's.

Which names exist depends on the workspace. Doctor registers conditionally, so a file-backed store contributes no Dolt checks, a suspended rig contributes none of its own, and there is no static catalogue anywhere in the tree to check a typed name against; the error says as much rather than insisting you've mistyped it.

<details>
<summary>Design notes, what this does not fix, and the test plan</summary>

### Why selection happens after the checks are built

Nothing about what gets registered changes.

`buildDoctorChecks` decides that from config, controller state, rig suspension and store contract, and filtering its output leaves all of it alone. Construction isn't where the time goes: the store factories and the order-history resolver are closures that don't open anything until a check runs. Building all 93 and running one costs 0.38s. Going further and making construction lazy would buy a fraction of a second, and it'd have to reason about every conditional registration to get there.

`SelectChecks` sits in `internal/doctor` next to the warm-up eligibility filter, which is the same shape of operation on the same slice. Two choices inside it are worth naming:

- It hands back the unmatched names instead of an error, so the CLI owns the wording.
- Registration order survives selection, because that order's deliberate: the beads-role check is registered ahead of every store-dependent check, so one root cause reports before the failures it causes.

### Why the flag is a string array rather than pflag's string slice

`StringSliceVar` splits on commas for you. It also reads an empty value as no values at all, which turns `gc doctor --check "$NAME"` with an unset `NAME` into a full sweep whose exit code gates on 93 checks the caller never asked about. That's this flag's own failure mode pointing the other way. `StringArrayVar` plus one explicit split keeps the comma form and lets an empty element survive to fail the run. There's a test for it.

### What this does not fix

A selected check still returns whatever it reports. `controller` reports `status: ok` with the real state in its message text whether the controller is running or not, and `TestControllerCheck_NotRunning` pins that behaviour. So `--check controller` costs 0.38s instead of 40s and hands you an exit code scoped to that one check, but the exit code still can't see a stopped controller, because the check's status never moves. Making a selected check return a verdict rather than a report is a change to the check, not to the selector, so it isn't here.

Two open issues on the check that dominates the sweep are worth reading beside this one. [#4895](https://github.com/gastownhall/gascity/issues/4895): `order-firing-current` hard-fails blocking on a 15s history-lookup timeout even while orders are firing. [#5268](https://github.com/gastownhall/gascity/issues/5268): its suspended-rig skip reads the deprecated `rig.Suspended` field. Neither is this ask. Both are about why the whole-sweep cost is what it is.

### Test plan

`go test ./internal/doctor/...` and `go test ./cmd/gc/...`, both full-package with no `-run` narrowing.

New in `internal/doctor/select_test.go`:

- An empty selector selects everything, and a blank name doesn't (`--check ""` mustn't turn into a sweep).
- Selection follows registration order, not request order.
- A repeated name selects its check once; a name shared by two registrations selects both.
- Unknown names come back in request order, while the matched half is still returned.
- Surrounding whitespace gets trimmed, and nil entries get skipped.

New in `cmd/gc/cmd_doctor_selector_test.go`, all driven through the real command tree:

- One name runs exactly that check, and the comma form and the repeated form agree.
- An unknown name exits non-zero, names itself, suggests the near match, lists the registry, and prints no results.
- The `--json` failure carries `error`, an empty `results`, and `registered_checks`.
- One bad name fails a run whose other name matched.
- The exit code follows the selection: a city with failing binary checks exits non-zero bare, and zero under `--check city-structure`.

Suggestions are tested for the two cases that need different mechanisms: a misspelling (`controler`, which has no substring relation to `controller` in either direction) and an incomplete name (`dolt-server` against `rig:core:dolt-server`).

### Diff shape

`doDoctor` took four flags positionally and a fifth would've made the call sites unreadable, so it takes a `doctorOpts` struct now, matching `buildDoctorChecksOpts` in the same file. That's what the test-file churn is: 22 call sites become named fields, which reads better than the positional booleans did. `docs/reference/cli.md` is regenerated with `go run ./cmd/gc gen-doc`, since it's generated output and `cmd_gendoc_test.go` fails on it going stale.

</details>

---
> • On behalf of: @austinborn


austinborn:
Round 2 of review, two findings addressed.

The `--json` payload for an unresolvable `--check` name was report-shaped, so `withDefaultSuccessOK` stamped `ok:true` on it and every count read zero beside an empty `results` array. A run that measured nothing reported clean to any caller checking a field other than the exit code, which is precisely the reading this flag exists to prevent. It now emits the shared failure envelope that `gc doctor --json-schema failure` publishes: `schema_version`, `ok:false`, and an `error` object of `code`, `message` and `exit_code`, with `registered_checks` alongside under the schema's `additionalProperties`. A successful `--json` run is byte-identical to before. The new test pulls the schema from the command itself and validates the payload against it, so it asserts the published contract rather than a copy.

The second was CI-side: `misspell` flagged the deliberate transposition in the suggestion fixture. Suppressed line-level, matching the convention already in `internal/config/undecoded_test.go`, so a real typo in that table still gets caught.

`golangci-lint` over `./cmd/gc/...` and `./internal/doctor/...` reports 0 issues, and the eleven selector tests plus `./internal/doctor` are green.

---
> Comment posted by the operator's software factory.
> • On behalf of: @austinborn

quad341 (changes requested):
@austinborn — the part of this I want to call out isn't the speed, it's that `--check nosuchthing` fails the *whole* run instead of reporting the subset that matched. That reasoning is exactly right: a caller filtering doctor output by name reads a short result set as a clean one, so a typo would otherwise report health nobody measured. Swapping the `--json` payload for the published failure envelope (because `withDefaultSuccessOK` would have stamped `ok:true` on a run that measured nothing) and validating it against `--json-schema=failure` instead of a copy is the same instinct, applied to the contract.

I'd normally take the rebase myself. I can't here, and it's worth being plain about why: your fork is owned by the `actual-software` org, and GitHub does not grant upstream maintainers write access to organization-owned forks — any commit I pushed to your branch would be rejected with a 403. Ticking "Allow edits by maintainers" won't change it; GitHub reports that flag as `true` on org forks anyway. So this is a `request-changes` purely as a delivery mechanism, not a judgement on the work. Everything I would have applied is below.

**1. Rebase onto current `main`** (`1154ad1` as of this review):

```bash
git fetch upstream main && git rebase upstream/main
```

**2. One conflict, in `cmd/gc/cmd_doctor.go`.** Main added a `d.Wait()` rationale comment right where you rewrote the prologue. Keep both — main's comment, your plumbing:

```go
func doDoctor(opts doctorOpts, stdout, stderr io.Writer) int {
	cityPath, err := resolveCity()
	if err != nil {
		fmt.Fprintf(stderr, "gc doctor: %v\n", err) //nolint:errcheck // best-effort stderr
		return 1
	}

	// Deliberately no d.Wait() here: gc doctor's whole point in bounding a check
	// is that a wedged one cannot stall the command, and waiting on an abandoned
	// goroutine would restore exactly that hang. Nothing this function owns
	// outlives the process, and ctx holds no handle a late writer can corrupt --
	// an abandoned check writes only to its own private buffer. A future caller
	// that reuses a Doctor in-process must call Wait before releasing ctx.
	d := &doctor.Doctor{CheckTimeout: opts.CheckTimeout}
	ctx := &doctor.CheckContext{CityPath: cityPath, Verbose: opts.Verbose}
```

The rest of that function is yours unchanged: `registered := buildDoctorChecks(...)`, then the `SelectChecks` / `reportUnknownDoctorChecks` / `for _, check := range selected` block, then `opts.JSON` / `opts.Fix` at the two `RunCollect` / `Run` call sites.

**3. Take `main`'s `buildDoctorChecks` body wholesale.** This is the one that isn't a conflict but will bite you if you resolve it the other way. Since your head was cut, main gated every store-dependent registration behind a `bead-store-preflight` probe (`storeOK`), added a `SkipStorePreflight` field to `buildDoctorChecksOpts`, and registered `newPoolIdleRoutedWorkCheck` and `doctor.NewOrderOutcomeHealthyCheck`. Your diff only touches the *call site* in `doDoctor`, so git won't flag it — just make sure none of that reverts. Your `buildDoctorChecksOpts{...}` literal is already field-named and doesn't set `SkipStorePreflight`, which matches main exactly, so it needs no edit.

**4. Regenerate the CLI reference** — this head's copy is 158 lines behind main and `Preflight / generated artifacts` will fail on it:

```bash
go run ./cmd/gc gen-doc
```

**5. Your 21 test call sites are safe.** I diffed each one against main: every `doDoctor(false, true, false, 0, ...)` line is byte-identical to your pre-image, so the `doctorOpts{...}` conversion replays cleanly. `builtin_include_doctor_check_test.go` has unrelated main edits at lines ~1317 and ~1468 (a removed `stubInitRemoteImports(t)`) — take main's side there; it's nowhere near your line 1036.

**6. Verify:**

```bash
go build ./...
go test ./internal/doctor/...
go test ./cmd/gc/...          # full package, no -run narrowing
golangci-lint run ./cmd/gc/... ./internal/doctor/...
```

One thing worth a look while you're in there, and the reason I'd rather you re-run the suites than trust the August CI: with main's new preflight, a store outage means the store-dependent checks never register at all — so `--check custom-types:city` during an outage will now report "no registered check matches" rather than surfacing the outage. Your error text already covers that reading ("a check it skips … registers no name to match"), so I think it's fine as-is, but you're better placed than I am to say whether it wants a line in the help or a `bead-store-preflight`-aware hint.

Two optional nits, take or leave: the comments in `splitDoctorCheckNames` and `select_test.go` say "91 checks" while the description says 93 — "every registered check" won't rot. And `--check <name> --fix` runs a remediation outside doctor's registration ordering, which matters for the builtin-pack-imports → packv2-import-state handoff; happy to file that as a follow-up issue rather than have it grow this PR.

Ping me when it's pushed and I'll merge it. The change lands as yours.


---

austinborn:
@quad341 — pushed, and your item 3 was the one worth the warning. Here's each of your six against the tree.

**1 and 2, the rebase and the `cmd_doctor.go` conflict.** Already resolved in an earlier pass, and resolved your way: main's `Deliberately no d.Wait()` rationale comment and our `SelectChecks` / `reportUnknownDoctorChecks` plumbing both survive. I merged current `main` rather than rebasing, since a rebase would rewrite the commits your review comments anchor to. Same end state, no conflicts.

**3, `buildDoctorChecks`.** Nothing reverted. `storeOK`, `SkipStorePreflight`, `newPoolIdleRoutedWorkCheck`, `NewOrderOutcomeHealthyCheck` and the `bead-store-preflight` probe are all present in `cmd/gc`. You were right that git wouldn't flag it, so I checked by symbol rather than by diff.

**4, the CLI reference.** `go run ./cmd/gc gen-doc` now produces no diff, so `docs/reference/cli.md` is current and `Preflight / generated artifacts` has nothing to catch.

**5 and 6.** The 21 test call sites replayed cleanly, as you predicted. On verification I'm one short of your list and want to be straight about it: `go build` and `go run` are fine, but I can't give you a clean `go test ./cmd/gc/...` because that package doesn't compile on current `main` at all.

That's the reason for the red, and it isn't this branch. `main` is broken at [`3d268c5`](https://github.com/gastownhall/gascity/commit/3d268c5849ca2ef49392c6e4adb68e9a10d0b59d) with 31 failing checks including `CI / required`. `releaseOrphanedPoolAssignments` gained a parameter in [#6261](https://github.com/gastownhall/gascity/pull/6261), and six call sites in `cmd/gc/pool_session_name_route_assignee_liveness_test.go` still pass the old signature. Both PRs were green alone; only the combination breaks. Since CI builds each PR merged into main's tip, any PR re-running its checks right now goes red on it, this one included. A PR that hasn't re-run since still shows its older green.

Your two nits I've left: the "91 checks" comments and the `--check <name> --fix` ordering question. Say the word on either and I'll fold them in or file them.

---
> • On behalf of: @austinborn

austinborn:
@quad341 — following up on the reply above: the branch is pushed at [`6bc5ef1`](https://github.com/gastownhall/gascity/pull/5354/commits/6bc5ef1892f77d6eb2c3313de41ce77ee30c1633), CI is green, and the three things that reply left open are closed. Ready for your merge.

**The verification gap is closed.** `main` repaired its own `cmd/gc` test build in [#6285](https://github.com/gastownhall/gascity/pull/6285), and merging current `main` absorbed it with no edit of ours. So the one command the earlier reply couldn't run now runs: `go test ./cmd/gc/...` passes as the full package with no `-run` narrowing, alongside `go build ./...`, `go test ./internal/doctor/...` and `golangci-lint run ./cmd/gc/... ./internal/doctor/...` at 0 issues. CI on the PR is 67 passing and 0 failing, `CI / required` included.

**The store-outage question, decided.** I went with one clause in the existing error text ([`cmd_doctor.go:622-625`](https://github.com/gastownhall/gascity/blob/6bc5ef1892f77d6eb2c3313de41ce77ee30c1633/cmd/gc/cmd_doctor.go#L622-L625)), with no help-flag line and no new code path. The reason's narrower than the general reading. During an outage `bead-store-preflight` does register, as an error check, so a plain `gc doctor` surfaces the outage fine. But `--check custom-types:city` selects by name, `custom-types:city` never registered, and the one check that knows about the outage gets filtered out of the very run that needed it. The registered-name list we already print does contain `bead-store-preflight`, so the signal's on screen; it's just unlabelled. What decided it is that the sentence named two skip reasons and both are configuration the operator chose (a file-backed store, a suspended rig). An outage is the one reading where they're actively misled rather than merely uninformed, so it belongs in the same sentence as the other two.

**Both nits.** The "91 checks" comments in `splitDoctorCheckNames` and `select_test.go` now say "every registered check", so the count can't rot. The `--check <name> --fix` ordering question I've left for the follow-up issue you offered; I agree it shouldn't grow this PR.

---
> • On behalf of: @austinborn

quad341 (approved):
@austinborn — this is merging as-is, and the thing worth naming is the judgment call at the center of it: making `--check nosuchthing` fail the *whole* run instead of reporting the subset that matched. A caller filtering doctor's output by name reads a short result set as a clean one, so one typo would otherwise report health nobody measured. You then applied the same instinct to the contract, swapping the `--json` payload for the published failure envelope and validating it against `doctor --json-schema=failure` rather than a copy — I pulled `schemas/failure.schema.json` and confirmed it backs you.

Thanks too for the verification discipline across three rounds. Checking `buildDoctorChecks` by symbol rather than by diff was exactly right — git had no reason to flag it, and a wrong conflict resolution there would have quietly reverted main's store preflight. And when main's own test build was broken you said so plainly instead of papering over it.

I'll file the `--check <name> --fix` ordering question as a follow-up on my side rather than growing this PR — you were right to leave it out. The change lands as yours.


---

PR #5356, opened 2026-08-17
docs: recover unmatched PM plan artifacts

quad341:
## Summary

- Recover 175 internal PM plan artifacts plus the Gas City local-only beads guidance from the preserved PM recovery branch.
- Recover the bbolt operator guide and navigation as a separate commit so reviewers can drop or defer that public-doc slice independently.
- Preserve the approved 178-path net delta on a fresh branch based directly on `origin/main`: 176 added, 2 modified, 0 deleted.

## Provenance

- Recovery source retained at `recovered-pm-20260817-1337` (`dd6f4dd7aec9`).
- Before cleanup, all 178 staged blobs matched the preserved recovery branch.
- Cleanup was limited to inherited trailing whitespace in 18 plan files and current public-doc terminology.

## Review risk / hold

The public bbolt guide describes implementation that is not present on current `origin/main`; commit `c52f08ecbe` should be dropped or deferred unless the corresponding implementation is restored. The contributor-plan recovery is isolated in `18a2fb52e3`.

Per the mayor ruling recorded on `ga-bccsbs`, keep this PR held until security alerts 313–316 are cleared.

## Verification

- `git diff --check`
- `make check-docs`
- Full pre-push fast gate passed with `LOCAL_TEST_JOBS=1` after the parallel form twice hit a live herdr pane-allocation race; no tests were skipped.


quad341:
Converted to draft to prevent an accidental merge. The recovery content is good; two paths need to come out first.

`docs/docs.json` adds a **Coordination Store** nav group that publishes `coordination-store/bbolt-opt-in` to the public docs site. That page tells users to set `backend = "bbolt"` in `city.toml` — but the bbolt backend was removed from the product on 2026-06-06 (`release-gates/ga-iiziq3-coordstore-benchmark-removal-gate.md`, Result: PASS). `git grep -lni bbolt origin/main -- '*.go'` returns zero files, and the startup signal the page promises (`coord-store: using bbolt backend <path>`) does not exist.

So merging as-is would publish end-user instructions for a backend that isn't there.

To unblock:

1. Drop the `docs/docs.json` hunk.
2. Move `docs/coordination-store/bbolt-opt-in.md` into `engdocs/`, next to `engdocs/plans/bbolt-backed-coord-store-opt-in.md` which is already in this PR. Keep the file — preserving it is the point of the recovery.
3. Leave `AGENTS.md` as-is; that hunk is correct.

That takes this to 177 files with no public-facing surface, and it can go back to ready.

The other 175 paths are all `engdocs/plans/*.md` and publish nowhere — I checked all of them, so there is nothing else to audit here.

quad341:
@quad341 — thanks for this. Recovering 175 PM plan artifacts and matching all 178 staged blobs against the preserved recovery branch before cleanup is careful salvage work, and isolating the risky slice into its own commit made this reviewable in minutes instead of hours.

Your call on the bbolt guide is right, and I confirmed it independently: there is no `bbolt` anywhere in the Go tree or `go.mod`, `compiledBackendNames()` registers only `""`, `dolt`, and `doltlite`, and no `coord-store-backend` check exists. That page would hand operators a runbook the binary can't serve.

So I'll take the integration on my side: drop `c52f08ecbe` — the guide plus its `docs.json` nav entry, which have to go as a pair or the docsync correspondence test trips — and merge the plan recovery and the `AGENTS.md` guidance as-is. Nothing needed from you.

One flag: your `AGENTS.md` edit lands inside the managed `BEADS INTEGRATION` block (lines 497–553). Worth checking whether `bd` regenerates that block and would overwrite it; if so it belongs just above the `BEGIN` marker.

Holding merge until 313–316 clear per `ga-bccsbs`. I'll re-file the bbolt guide as a follow-up blocked on implementation.


---
**Reviewers**: Qwen — **failed** · Claude (claude-opus-5[1m]) — ok · Codex — **failed**


Review coverage:
- completed: `claude`
- unavailable: `qwen codex`

PR #5357, opened 2026-08-17
Fix: DoltLite before-filters depend on an unenforced cross-repo timestamp contract (silent row drop)

quad341:
## What this changes

DoltLite list queries using `CreatedBefore` or `UpdatedBefore` no longer silently omit rows whose timestamp text SQLite's `julianday()` cannot parse, including values serialized from a raw Go `time.Time`. Those rows now reach the existing Go-side timestamp filter, which remains the authoritative compatibility path.

Bounded before-filter queries also avoid applying SQL `LIMIT` before that Go-side filter and re-sort. This preserves the same result prefix as an unbounded query when SQL and Go assign different sort positions to an unparseable timestamp.

## Review notes

- The change is confined to the native DoltLite read path and its build-tagged regression tests.
- It does not change the stored schema, migrations, or timestamp serialization format.
- Before-filtered reads may fetch more candidates because limiting happens after the Go-side compatibility filter; other bounded read shapes keep their existing SQL fast paths.

## Test plan

- [x] Run `make test-native-doltlite-beads` and verify both raw-time binding and bounded-prefix regression tests pass.
- [x] Run the full local CI-equivalent aggregate with `make test-local-full-parallel`.
- [x] Run the required policy lane with `make test-ci-policy`.
- [x] Review the attributed raw failures and complete evidence in [`release-gates/ga-h4wdea-doltlite-before-filter-gate.md`](release-gates/ga-h4wdea-doltlite-before-filter-gate.md).

### DIFF
--- a/internal/beads/doltlite_read_store.go
+++ b/internal/beads/doltlite_read_store.go
@@ -918,6 +918,16 @@ func (s *DoltliteReadStore) queryIssuesOrderedInTables(query ListQuery, sets []d
 		if query.SeekAfter != nil {
 			tableLimit = 0
 		}
+		// A before-filter admits NULL-julianday rows into the SQL result (see
+		// buildDoltliteTableQuery) precisely so filterDoltliteBeforeTimes can
+		// re-validate them Go-side. Those rows can sort anywhere in SQL's
+		// ORDER BY (it has no valid timestamp to key on), which may disagree
+		// with the Go-side zero-time treatment — so, exactly like SeekAfter
+		// above, a SQL LIMIT taken before that Go re-sort could cut the wrong
+		// prefix. Fetch unbounded whenever a before-filter is active.
+		if !query.CreatedBefore.IsZero() || !query.UpdatedBefore.IsZero() {
+			tableLimit = 0
+		}
 		rows, err := s.queryIssueTable(query, tables, extraWhere, extraArgs, tableLimit, orderBy)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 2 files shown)


quad341:
@quad341 — this is a good catch and a genuinely subtle one. The failure mode here is the worst kind: `julianday()` returns NULL on a driver-serialized `time.Time`, `NULL < julianday(?)` is NULL rather than true, and the row just quietly vanishes from the result set before the tolerant Go-side filter ever gets a look at it. Nothing errors, nothing logs, rows are simply missing.

What I especially appreciate is that you fixed *both* halves. Widening the WHERE clause alone would have quietly regressed the bounded case, since a NULL-julianday row sorts on its raw text in SQL and on zero-time in Go — and you pinned that with a test that sweeps every limit in both sort directions. I reverted each hunk separately on my side to confirm; each test fails against exactly its own half.

Merging now. The red CI isn't yours: `Contract / bd CLI` hit an `actions/setup-go` download 429, and `packages-core-2-of-4` is the `internal/config` flake that #5366 is already deflaking. The 4 CodeQL notes are pre-existing `go.mod` advisories — I'll pick those up separately. Thanks for the thorough writeup.


---


Review coverage:
- completed: `qwen codex`
- unavailable: `claude`


quad341:
@quad341 — this is a nicely pinned bug. `julianday()` returns NULL on a driver-serialized `time.Time`, `NULL < julianday(?)` evaluates to NULL rather than true, and the row simply vanishes before the tolerant Go-side filter — the one the code already documents as authoritative — ever sees it. No error, no log, just short results. I confirmed the mechanism directly: `parseTimeString` accepts modernc's `"... -0700 MST"` write format, so those were valid rows being silently discarded.

What I appreciate most is that you fixed both halves. Widening the `WHERE` alone would have regressed bounded reads, since a NULL-julianday row sorts on raw text in SQL and on zero-time in Go. Your bounded-prefix test walks every limit in both sort directions — that's the right way to pin it.

Two things I'll take on my side: your branch predates `scripts/prwatchdog`, which is the only reason the Evidence check is red, so I'll merge current `main` in. And the `Count` doc comment in `doltlite_count.go` now recommends a cleanup your change makes unsafe, so I'll reword it. Shard 7 is a module-proxy transport error, not yours.

Thanks for the thorough gate writeup.


---

PR #5360, opened 2026-08-17
fix(dashboard): widen city-name regex to match registry grammar (#5316)

### LINKED ISSUE #5316: City name with a dot is unreachable in the dashboard and silently redirects to another city
A city whose name contains a dot registers fine but is unreachable in the dashboard, and instead of saying so the dashboard silently redirects you to a different city.

I have two cities registered, `gascity` and `spectacular.city`. Opening `/city/spectacular.city/` bounces straight back to `gascity` every time.

## Why

The supervisor registry and the dashboard disagree about what a city name is.

The registry explicitly allows dots — `internal/supervisor/registry.go:24`:

```go
// Must start with alphanumeric and contain only alphanumerics, hyphens,
// underscores, and dots.
var validCityName = regexp.MustCompile(`^[a-zA-Z0-9][a-zA-Z0-9._-]*$`)
```

The dashboard allows neither dots nor underscores, in three separate places:

- `internal/api/dashboardbff/links.go:18` — `^[A-Za-z0-9]([A-Za-z0-9-]*[A-Za-z0-9])?$`
- `dashboardspa/web/shared/src/city.ts:13` — `/^[a-z0-9]([a-z0-9-]*[a-z0-9])?$/i`
- `dashboardspa/web/frontend/src/api/cityBase.ts:29` — same shared regex

Since the city name comes from the directory basename when `[workspace] name` is unset, any directory with a dot in it produces a city the dashboard cannot serve.

## The redirect

`CityBootstrap.tsx:51-52` runs the URL city through `CITY_NAME_RE`, and on failure `parseCityFromPath` returns `null`. Everything downstream then treats the request as though there were **no city in the URL at all**, so it falls through to the bare-root branch and redirects to the first entry of `/v0/cities` (`CityBootstrap.tsx:96-103`). That list is sorted alphabetically server-side (`huma_handlers_supervisor.go:249-255`), so `gascity` wins and you land there.

So it is not a transient or a race — a dotted city always redirects to whichever city sorts first.

## The API half

The BFF rejects it independently, which I confirmed against a running supervisor:

```console
$ curl -s -o /dev/null -w '%{http_code}' /api/city/gascity/config
200
$ curl -s /api/city/spectacular.city/config
{"error":"unknown city"}
```

`resolveCityPath` calls `ValidCityName` before the resolver lookup (`dashboardbff/plane.go:212-216`), so it 404s before ever asking whether the city exists. The `/v0/city/spectacular.city/...` endpoints all return 200 — they don't use that regex — so it's specifically the dashboard plane.

## Expected

Either the dashboard should serve a name the registry accepted, or registration should refuse a name the dashboard can't serve. Right now both grammars ship and disagree, and the user-visible result is a silent redirect with no indication anything is wrong.

Separately, and regardless of which direction you pick: an unservable city in the URL probably shouldn't fall through to the bare-root "pick the first city" branch. The unknown-city screen already exists (`CityBootstrap.tsx:90-94`) and would make this obvious instead of looking like a UI bug.

## Happy to send a PR

I'll do the work, but the direction is a maintainer call and I don't want to guess:

1. **Widen the dashboard** to match the registry grammar (`[a-zA-Z0-9][a-zA-Z0-9._-]*`), updating the Go regex, the shared TS regex, and the tests. Existing cities keep working. Needs a check that dots in a path segment are safe everywhere the name is interpolated.
2. **Tighten the registry** to reject dots and underscores at registration. Cleaner grammar, but it's a breaking change for anyone who already has such a city registered.

Tell me which you'd prefer and I'll put it together. I'd lean toward 1 plus the unknown-city screen fix, since it doesn't strand existing registrations, but I don't have the context on why the dashboard grammar is narrower.

rjgeng:
Fixes #5316.

Dashboard/API city-name matching used a regex narrower than what the city registry actually accepts, so a dotted city name (a valid, registered city) silently redirected elsewhere in the dashboard instead of resolving to the right page.

**Fix:** widened the city-name regex to match the registry's actual grammar.

**Testing:** `make dashboard-build`/`make dashboard-check` both green (this touches the dashboard bundle). `go build ./...` clean, `gofmt -l` empty, targeted `go test` green.

---


rjgeng:
Heads-up on the red CI here: `cmd/gc process / shard 1 of 12` is failing on `TestCmdStopSupervisorManagedInvalidCityTomlFailsWhenShutdownFails`, which is unrelated to this diff. Confirmed it fails identically and deterministically on plain `main` at the current tip in a clean worktree with none of this PR's changes applied — filed as #5668. Current head here is `efbb5516`.

quad341 (approved):
@rjgeng — thanks for chasing this one to its actual root. Widening the regex was the easy half; what makes this a good fix is that you reconciled the *test matrix* instead of just deleting the case that stopped being invalid. Repurposing the negative test to a 65-character name kept real coverage on the "BFF rejects it" path, and updating the comments that described the old narrower grammar meant I could verify your parity claim against `internal/supervisor/registry.go` in about a minute rather than reconstructing it myself. I checked the character classes directly — they match exactly — and confirmed the leading-alphanumeric requirement does the path-safety work your comment says it does, and that every TS caller goes through the shared `CITY_NAME_RE`.

Two comments elsewhere in the tree still describe the grammar as "narrower" — one in `dashboardbff/runtailer.go`, one in `shared/src/views.ts`. Neither is yours to have found, and neither affects behavior. I'll take both on my side, re-run the dashboard check and the targeted Go tests, and merge with your commit and credit intact. Nothing needed from you.

Nice, well-scoped bug fix.


---

PR #5361, opened 2026-08-17
fix(gc import): report actual upgrade movement, not lockfile entry count (#5312)

### LINKED ISSUE #5312: bug(cmd/import): `Upgraded N remote import(s)` counts lockfile entries, so a sha-pinned import that cannot move still reads as upgraded
## Summary

`gc import upgrade` finishes by printing `Upgraded N remote import(s)`, where `N` is the number of entries in the lockfile it just wrote. Nothing in that path compares the new resolution against the previous one, so the number can't reflect movement even in principle. An import pinned to an exact commit is a constraint with exactly one satisfying version, so it can't go anywhere. It resolves to itself, and the only field that changes in `packs.lock` is `fetched`, a timestamp.

The line is truthful. The upgrade did resolve, the constraint did have exactly one satisfying version, and on its own terms the command did what it was asked. It still reads as movement. Every operator and every agent takes it that way, because the identical sentence is what a real upgrade prints.

What makes it worth a wording change rather than a documentation note is that the next two commands agree with the wrong reading. `gc import install` reports `Installed N remote import(s)`. `gc import check` then reports `Import state OK: N remote import(s) checked`, which is also true, because that check is asking whether the cache is consistent with the lock and it stays consistent when the lock never moved. Three success lines in a row. None of them answers the question the operator was actually asking, which is whether the city is running current pack content.

```mermaid
flowchart TD
    U["gc import upgrade prints 'Upgraded 6 remote import(s)'"] --> I["gc import install prints 'Installed 6 remote import(s)'"]
    I --> C["gc import check prints 'Import state OK: 6 remote import(s) checked'"]
    C --> B["Reader concludes the city is running current pack content"]
    L["packs.lock is unchanged except for the 'fetched' timestamp, because every entry is an exact sha pin"] -. "silently contradicts" .-> B
```

## Where the number comes from

[`cmd/gc/cmd_import.go:815`](https://github.com/gastownhall/gascity/blob/80e78ae0670586af8f55e45e40ebee7be2ef9fae/cmd/gc/cmd_import.go#L815), in `doImportUpgrade`:

```go
if target == "" {
	fmt.Fprintf(stdout, "Upgraded %d remote import(s)\n", len(lock.Packs)) //nolint:errcheck
} else {
	fmt.Fprintf(stdout, "Upgraded import %q\n", target) //nolint:errcheck
}
```

`lock` is the lockfile just built and written, so `len(lock.Packs)` is the size of the resolved closure. The pre-existing lock is never read in this function, so there is no baseline to compare against and no count of movement is possible at this point in the code.

The single-target form is quieter about the same thing. `Upgraded import "local-core"` names the import and reports no version and no commit at all, so it cannot distinguish a pin that advanced from one that stayed put.

## What actually changes when a pin resolves to itself

A lock entry is three fields, at [`internal/packman/lockfile.go:29-33`](https://github.com/gastownhall/gascity/blob/80e78ae0670586af8f55e45e40ebee7be2ef9fae/internal/packman/lockfile.go#L29-L33):

```go
type LockedPack struct {
	Version string    `toml:"version"`
	Commit  string    `toml:"commit"`
	Fetched time.Time `toml:"fetched"`
}
```

Under `InstallUpgrade` the sync always falls through to a fresh resolution and re-stamps the timestamp, at [`internal/packman/install.go:302-320`](https://github.com/gastownhall/gascity/blob/80e78ae0670586af8f55e45e40ebee7be2ef9fae/internal/packman/install.go#L302-L320):

```go
case InstallUpgrade:
	// Always refresh below unless this sync already resolved the source.
...
return s.storeChosen(source, LockedPack{
	Version: resolved.Version,
	Commit:  resolved.Commit,
	Fetched: time.Now().UTC(),
}, true), nil
```

For an exact sha pin, `Version` and `Commit` come back identical to what was there before. `Fetched` moves. So the on-disk diff after a no-op upgrade is one timestamp per import, which is easy to miss and easy to mistake for evidence that something happened.

## Why the output cannot be read as a signal today

The same line, with the same shape, is printed in both cases. On 2026-08-07 we bumped three `file://` packs across 43 commits, and a successful run printed `Upgraded 5 remote import(s)`. A run in which every pin resolves to itself prints the same sentence with the same number. There is no textual difference between the run that moved four thousand lines of pack content and the run that moved nothing, which is why nobody on our side caught the difference by reading output.

## What it cost us

A city here sat more than 24 commits behind merged pack content for three days, while the procedure being followed to fix it recommended exactly this command, and every run reported success. A second city was found 54 commits behind under the same conditions. In both cases the only reliable tell was going and reading whether the pin in `packs.lock` had actually changed, which is not something an operator does after a command reports success three times.

Here is the version of it you can check from outside our factory. This city's `packs.lock` pins `https://github.com/gastownhall/gascity.git//internal/bootstrap/packs/core` at [`f895c0ff`](https://github.com/gastownhall/gascity/commit/f895c0ff47d6ee9334ed282a416387eb5b084d24), the same commit [#4071](https://github.com/gastownhall/gascity/issues/4071) names as the bundled pin `gc init` seeds into fresh cities, and it sits 1,493 commits behind `main` at [`80e78ae0`](https://github.com/gastownhall/gascity/commit/80e78ae0670586af8f55e45e40ebee7be2ef9fae). All six entries in that lockfile are `version = "sha:<commit>"` exact pins. So by the code above, `gc import upgrade` here prints `Upgraded 6 remote import(s)` and moves nothing, every time it runs. We haven't run it on the live host to capture a transcript, since that rewrites `packs.lock` under a running factory, so treat the predicted line as read from the source rather than observed.

The two defects compound in the obvious direction. One seeds a stale exact sha into a new city, and this one reports success on every attempt to move it.

## Suggested change

Say what happened rather than how many imports exist. When an import's resolved commit equals its installed commit, it did not move, and the output can say so:

```
Upgraded 1 of 6 remote import(s); 5 already at their pinned version
```

Or, when nothing moved at all:

```
No import moved; 6 already at their pinned version
```

This is a wording change at the command layer and needs no new plumbing. `packman.ReadLockfile` is already exported and already returns an empty lock for a missing file, so `doImportUpgrade` can read the pre-existing lock at entry and compare `Commit` per source before it prints:

```go
before, err := packman.ReadLockfile(fsys.OSFS{}, cityPath) // at function entry, before the sync
...
moved := 0
for source, pack := range lock.Packs {
	if prev, ok := before.Packs[source]; !ok || prev.Commit != pack.Commit {
		moved++
	}
}
```

For the single-target form, printing the resolved commit would be enough on its own, since `Upgraded import "local-core" (a1b2c3d, unchanged)` cannot be misread.

One test pins the current single-target string, at [`cmd/gc/cmd_import_test.go:1357`](https://github.com/gastownhall/gascity/blob/80e78ae0670586af8f55e45e40ebee7be2ef9fae/cmd/gc/cmd_import_test.go#L1357). We did not find a test asserting the all-imports message, so that one appears to be free to change.

If the preference is to keep the message stable for scripts that parse it, an added second line would solve the operator-facing half without touching the first line at all. We would rather report the shape of the problem than argue for one fix over another.

## Environment

- `gc` 1.4.0, Ubuntu 24.04.4 LTS, Linux 7.0.0-1010-aws, x86_64, 4 vCPU.
- Source read from a fresh clone of `gastownhall/gascity` at [`80e78ae0`](https://github.com/gastownhall/gascity/commit/80e78ae0670586af8f55e45e40ebee7be2ef9fae), not from a fork and not from the pinned import cache. Re-checked against `main` at [`126029e5`](https://github.com/gastownhall/gascity/commit/126029e5a3a7) immediately before filing: the printing code above is unchanged.

## Related issues

- [#4185](https://github.com/gastownhall/gascity/issues/4185) — `gc reload` prints "No config changes detected" for an import pin bump that does change effective config. Same family: a sha-pinned import plus a status line that answers a narrower question than it appears to.
- [#4071](https://github.com/gastownhall/gascity/issues/4071) — the bundled core-pack pin is not refreshed before release, so `gc init` seeds fresh cities with a stale sha. That is where the [`f895c0ff`](https://github.com/gastownhall/gascity/commit/f895c0ff47d6ee9334ed282a416387eb5b084d24) above came from, and it is why the no-op upgrade path gets reached so often on fresh cities.

---
> • On behalf of: @austinborn

rjgeng:
Fixes #5312.

`gc import upgrade` reported its upgrade count as the number of lockfile entries touched, not the number of packs that actually moved to a new version — so a sha-pinned import (which `gc import upgrade` correctly leaves untouched, per its own pinning contract) still got silently counted as "upgraded" just because its lockfile entry was re-written.

**Fix:** the reported count now reflects actual version movement, not lockfile entry churn.

---


quad341 (approved):
@rjgeng — thank you for this one. What I appreciate most is that you went after the actual cause instead of just rewording the output: reading the pre-upgrade lock and diffing commits per source is the right shape for this, because `len(lock.Packs)` structurally could not tell "moved" from "re-serialized." The comment you left at the top of `doImportUpgrade` explaining *why* the pre-state has to be captured is what will stop someone collapsing it back to an entry count in six months.

Your tests are the strongest part of the change. `TestDoImportUpgradeAllPinnedReportsNoMovement` reproducing #5312's exact all-sha-pinned scenario, with the negative assertion that "Upgraded" never appears, is precisely the test that keeps this fixed.

I'm taking two small things on my side and merging. I'd like "already at their pinned version" to become "already up to date" — an import on `^1.0` with nothing newer to move to isn't really pinned — and same reasoning for the single-target `(sha, unchanged)` line, which currently says "Upgraded" and "unchanged" in one breath. I'll update your two expectation strings and add a case for that unchanged branch.

CI is red, but it's a six-hour worker job that got cancelled and took a batch of shards with it — nothing to do with your diff. I'll rebase and re-run. Nothing needed from you.


---

PR #5362, opened 2026-08-17
fix(prompt): resolve {{.ConfigDir}} in prompt template scope (#5315)

### LINKED ISSUE #5315: Prompt scope renders {{.ConfigDir}} as an empty string, silently
## What happens

`{{.ConfigDir}}` resolves to an empty string in every prompt `gc prime` renders. It's a valid name in session-setup scope, so the same spelling that works in `pre_start` produces nothing in a prompt, and nothing reaches stderr either way.

A pack whose prompts invoke helper scripts as `{{.ConfigDir}}/assets/scripts/<name>.py` therefore ships agents that call `/assets/scripts/<name>.py`, an absolute path that exists on no machine. The call exits 127.

## Why it is silent

`renderPrompt` flattens `PromptContext` into a `map[string]string` through `buildTemplateData`, then executes the template with `Option("missingkey=zero")`:

- `cmd/gc/prompt.go:26` — `PromptContext` declares no `ConfigDir` field.
- `cmd/gc/prompt.go:317` — `buildTemplateData` copies `ctx.Env` and then sets each SDK field by name. `ConfigDir` is never among them.
- `cmd/gc/prompt.go:125` — `Option("missingkey=zero")`.

On a map, `missingkey=zero` yields the zero value of the map's *element* type. That type is `string`, so a missing key doesn't raise — it just renders as `""`. Executing against the struct directly would have failed at execute time with `can't evaluate field ConfigDir in type main.PromptContext`, and we'd have caught it on the very first render.

`SessionSetupContext` does carry `ConfigDir`, which is what makes this worth a report rather than a docs note: the two scopes disagree about a name that looks like one variable, and the disagreement is only observable by reading the rendered output.

## Reproducing it

Any agent whose prompt builds a path from the variable:

```
gc prime <agent> | grep -oE '[^ "]*assets/scripts/[a-z_]*\.py' | sort -u
```

A leading `/` with no directory in front of it is the empty render. `gc prime <agent> | grep -c '{{'` returns 0 at the same time, so nothing is left unsubstituted — the variable resolves, and it resolves to nothing. Measured on gc 1.4.0 across five agent classes, the prompts we ship produced 9 dead paths.

## Impact

The expensive part here is the failure mode rather than the breakage. An agent hitting `command not found` reads it as a packaging or permissions problem on the script, because that's exactly what the message describes. We watched one such call get diagnosed as an exec-bit defect and fixed twice as that, while the path in front of the exec bit stayed empty and the call kept never reaching the file. The workaround (hardcode an absolute path) works, so it tends to get written down instead of reported, and the cause stays hidden.

## Suggested fix

Add `ConfigDir` to `PromptContext`, publish it from `buildTemplateData`, and resolve it in both scopes through one shared helper so they cannot drift again. The natural value is the agent's config `SourceDir`, falling back to the city root.

Worth considering alongside it: `missingkey=zero` over a `map[string]string` turns *every* misspelled or out-of-scope variable into an empty string. `missingkey=error` at render time, or a lint pass over the known key set, would convert this whole class from silent to loud. We've only hit `ConfigDir` so far, but there's nothing structural stopping the next one.

---
> • On behalf of: @austinborn

rjgeng:
Fixes #5315.

`{{.ConfigDir}}` referenced in a prompt template silently resolved to empty instead of the actual config directory. Root cause: the field wasn't populated in the template scope struct, so Go's `text/template` `missingkey=zero` behavior quietly substituted the zero value (empty string) instead of erroring.

**Fix:** `ConfigDir` now resolves correctly in prompt template scope.

---


quad341 (approved):
@rjgeng — the part of this I appreciate most isn't the field add, it's that you went looking for *why* it failed silently and landed on `Option("missingkey=zero")`. That's the actual answer, and it's why #5315 could sit there looking like a typo in someone's template rather than a bug in ours.

Extracting `resolveConfigDir` instead of adding a fourth copy of the `SourceDir`-else-`cityPath` conditional was the right instinct — `worker_handle.go` and `template_resolve.go` both got shorter for it. I checked the one thing that could have bitten you: hoisting `configDir` from Step 11 up ahead of Step 9 is only safe if `SourceDir` can't change in between. It can't. Clean hoist.

Two small things I'll take on my side rather than send back to you: `lintPromptContext` in `cmd_lint.go` is a third `PromptContext` site that still doesn't set the field, so I'll wire it and soften the helper's comment to match. And I'll add a `ConfigDir` assertion against `buildPrimeContextFor` — your tests are good, but they all build the context by hand, so they'd still pass if someone dropped the wiring line. Since the bug *was* a dropped wiring line, that's the one worth pinning.

Merging with those on top. Nice catch.


---

PR #5363, opened 2026-08-17
fix(nudge): record dispatch skips for busy-session legacy nudge poll (#5317)

### LINKED ISSUE #5317: bug(nudge): a queued nudge to a live but continuously busy session is never attempted — poller returns before the claim, so attempts stays 0 until the TTL expires it
### Before you continue

### Gas City version

`1.4.0` (commit [`a7297c511`](https://github.com/gastownhall/gascity/commit/a7297c511d637a3609947386f3389d76ddb2f23b), built 2026-07-24T17:43:35Z). Every code reference below was re-checked against `origin/main` at [`3ba850f`](https://github.com/gastownhall/gascity/commit/3ba850ff60c33752cfafe422310e233335c53e50) and is unchanged there.

### Environment

Ubuntu 24.04, Linux 7.0.0-1010-aws, x86_64, 4 vCPU. tmux runtime provider, managed reconciler, legacy nudge dispatch (per-session `gc nudge poll` sidecars). Autonomous multi-session pool with long-running agent sessions.

## Summary

A queued nudge addressed to a session that is **alive and continuously busy** is never attempted. Not delayed, not retried-and-failed: never attempted at all. The entry sits in `state.Pending` with `attempts` at 0 and `last_attempt_at` at the Go zero value until the 24-hour TTL reaps it as `expired`, and nothing anywhere records that a delivery opportunity was skipped.

The mechanism inverts the property you want. Delivery requires the target to be quiet for `--quiescence` (default 3s), so a session that stays busy for a long stretch never becomes reachable, and a long-running session is exactly the one an operator needs to interrupt with a scope change or a stop-work order.

## Root cause

[`tryDeliverQueuedNudgesByPoller`](https://github.com/gastownhall/gascity/blob/3ba850ff60c33752cfafe422310e233335c53e50/cmd/gc/cmd_nudge.go#L1403-L1414) returns before it claims anything when the quiescence gate fails:

```go
if !pollerSessionIdleEnough(target, sp, quiescence, obs) {
    return false, nil
}
items, err := claimDueQueuedNudgesForTarget(target.cityPath, target, time.Now())
```

Because the early return precedes `claimDueQueuedNudgesForTarget`, the queue entry is never touched. `attempts` stays 0, `last_attempt_at` stays `0001-01-01T00:00:00Z`, and `last_error` stays empty, so a skipped entry is byte-identical to one enqueued a moment ago.

The `(false, nil)` return is also indistinguishable from "there was nothing to deliver", which is what makes it invisible to the caller. [`pollerSessionIdleEnough`](https://github.com/gastownhall/gascity/blob/3ba850ff60c33752cfafe422310e233335c53e50/cmd/gc/cmd_nudge.go#L1514-L1535) waits at most `quiescence` for idle and returns false on timeout, so a busy target fails the gate on every tick for as long as it stays busy.

## Reproduction

1. Run a city in legacy nudge dispatch, so per-session `gc nudge poll` sidecars own delivery.
2. Give a managed session a long continuous turn — anything that keeps the provider streaming for several minutes.
3. `gc session nudge <that-session> "stop and re-read the plan"`. It prints `Queued nudge for <agent>` and exits 0.
4. `gc nudge status` shows the entry pending. Read `.gc/nudges/state.json` and its `last_attempt_at` is `0001-01-01T00:00:00Z`.
5. Keep the session busy. The entry stays at zero attempts for its whole 24-hour life, then moves to `Dead` with `last_error: "expired"`, still showing zero attempts.

## Expected behavior

A skipped delivery opportunity should leave a trace. Any one of these would be enough:

- Record the attempt (or a distinct "deferred" counter) on the entry when the quiescence gate rejects a target, so `attempts` and `last_attempt_at` stop implying nothing was ever tried.
- Give the poller path the skip accounting the supervisor dispatcher already has, so `gc nudge status` can show why an entry is sitting there.
- Set `last_error` on the entry to name the gate, so the dead-letter record says *why* it expired rather than only *that* it did.

## Actual behavior, measured

One factory host, scanned 2026-08-16:

| | |
|---|---|
| Queue entries total | 278 |
| Pending | 260 |
| Pending with at least one recorded attempt | **0** |
| Pending addressed to a session alive at scan time | 12 |
| Oldest pending entry | 23h48m, minutes from expiry |
| Dead entries | 18, all `last_error: "expired"`, all with `attempts: 0` |

The busiest session on the box (one continuous turn running over an hour) held 15 undelivered entries, the oldest 61 minutes old. Among them was a `source: "mail"` wake, so a `gc mail send --notify` had also failed to wake its recipient through the same path. That session exited during the scan, which converted all 15 from recoverable to permanently undeliverable in one step.

Real instruction traffic was in the backlog, not just routine wakes. Four manager messages of the "stop before you act on this" shape were sitting unattempted, the oldest 4h33m old.

## The observability that exists but does not cover this

The supervisor-side dispatcher already counts this exact case. [`dispatchAllQueuedNudges`](https://github.com/gastownhall/gascity/blob/3ba850ff60c33752cfafe422310e233335c53e50/cmd/gc/nudge_dispatcher.go#L239-L254) accumulates a `not-delivered` skip reason whose comment names the gate directly:

```go
// Matched a live, running session, yet nothing was claimed/delivered
// this tick — e.g. the poller quiescence gate (pollerSessionIdleEnough)
// hasn't cleared ...
```

Those counters persist into `State.DispatchSkips` and are readable from `gc nudge status`. But `dispatchAllQueuedNudges` returns early unless `nudgeDispatcherIsSupervisor(cfg)`, so a city on per-session pollers gets none of it. The poller path has no equivalent, which leaves the most common deployment shape with no signal at all.

Neither of the two closest issues covers this, and I'd read this as a third distinct case rather than a duplicate of either:

- [#4534](https://github.com/gastownhall/gascity/issues/4534) is about a nudge to a **drained or terminated** session, where `gc nudge status` reads 0/0/0. Here the target is alive the whole time and the entry stays plainly visible in `Pending`; what is missing is any record that delivery was skipped.
- [#3637](https://github.com/gastownhall/gascity/issues/3637) (closed) is about the **caller** blocking ~35s on a synchronous `WaitForIdle` when the target is busy. That is the send path. This report is about the poller path on the other side of the queue, and it costs delivery rather than caller latency.

[#1069](https://github.com/gastownhall/gascity/issues/1069) is the nearest in symptom ("queued nudges never delivered") but attributes it to ACP sessions being skipped, which is a different cause.

## Impact

This is the channel that carries every scope change, reversal, and stop-work instruction to an in-flight worker, and the sender's only feedback is a success line printed before delivery is attempted. On this factory it has already cost one build, which shipped against a placement decision that had been reversed ten minutes earlier; the reversal was still sitting in the queue, unattempted, while the work completed.

The workaround is to write anything load-bearing onto the work item first and treat the nudge as best-effort. That works for a session that has not started yet, and it does not help a worker that already read the item and is mid-flight, which is the case the nudge exists for.

---
> • City: `city` · Agent: `local-core.builder-5`
> • On behalf of: @austinborn

rjgeng:
Fixes #5317.

The legacy per-session nudge poll loop delivered nudges to busy sessions by skipping them silently, with no accounting — so a busy-session skip left the nudge invisibly stuck at `attempts:0` forever, indistinguishable from a nudge that was never even attempted.

**Fix:** the legacy poll loop now records dispatch skips, reusing the existing supervisor-dispatch skip-accounting pattern, so a busy-session skip is now visible and attributable.

---


quad341 (approved):
@rjgeng — thank you for this. #5317 is easy to describe and genuinely annoying to locate, and you found the real thing: `legacy` is the default dispatcher, so the per-session poller is the delivery path for most cities, and it was the one path with no skip accounting at all. The test is what I want to call out — driving the actual `pollerSessionIdleEnough` quiescence gate through a busy observation stub, then reading the counter back through `gc nudge status --json` instead of poking at state, is exactly how I'd want this tested. It lands unchanged.

One thing I'll take on my side before merging. `tryDeliverQueuedNudgesByPoller` also returns `(false, nil)` when the queue is simply empty, and the poll loop keeps running after its nudge drains — so as written, an idle-but-live session would bump `not-delivered` every 2s forever and drown out the signal you set out to surface, taking the queue flock each time. The supervisor dispatcher avoids this with a `matched` gate before it counts, with `not-matched` as its separate bucket. I'll add the equivalent gate plus a negative test for the empty-queue case, and re-run the `cmd/gc` shards.

Merging with your commit and credit. Good catch on a genuinely invisible failure mode.


---

PR #5364, opened 2026-08-17
fix(nudge): re-validate mail nudges at delivery time too (#5321)

### LINKED ISSUE #5321: bug(nudge): delivery-time re-validation is gated on source == "wait", so a mail nudge wakes an agent for mail it already read
### Before you continue

### Gas City version

`1.4.0` (commit [`a7297c511`](https://github.com/gastownhall/gascity/commit/a7297c511d637a3609947386f3389d76ddb2f23b), built 2026-07-24T17:43:35Z). Every code reference below was read from `origin/main` at [`e2a59bf0`](https://github.com/gastownhall/gascity/commit/e2a59bf0c1096937912b43c6a7949b508f7ebfc8) and re-checked at [`ab54f93c`](https://github.com/gastownhall/gascity/commit/ab54f93c3b3d30bb8f52f7e92c0761fb75340752) immediately before filing, where both cited functions are unchanged. They're also byte-identical to the `v1.4.0` tag this host runs, so this isn't a report against a stale build.

### Environment

Ubuntu 24.04, Linux 7.0.0-1010-aws, x86_64, 4 vCPU. tmux runtime provider, managed reconciler, legacy nudge dispatch. Autonomous multi-session pool with one always-on coordinator agent and ephemeral worker pools.

## Summary

The nudge queue already knows how to re-validate a deferred nudge at delivery. [`blockedQueuedNudgeReason`](https://github.com/gastownhall/gascity/blob/e2a59bf0c1096937912b43c6a7949b508f7ebfc8/cmd/gc/cmd_nudge.go#L1667-L1695) re-reads the object a nudge refers to and withdraws the nudge when that object isn't worth waking anyone for any more. It's a good check, and it gets the awkward case right: a referent that has vanished counts as withdraw rather than as an error.

That check runs for one source class. Its first line returns early unless `item.Source == "wait"`, so mail-sourced and session-sourced nudges reach the agent without anyone asking whether the thing they announce is still outstanding. Often enough, by then it isn't. The agent wakes, looks, and finds the item already handled.

Underneath that gate there's a second gap, and it would keep the fix from working if only the gate moved. Mail nudges carry no reference at all, so there's nothing for a widened check to read. In just under five days on one factory, all 459 mail nudges and all 1,417 session nudges were enqueued with an empty `reference_json`. Not one nudge in that window was wait-sourced, so every nudge the factory queued fell in a class nothing re-validates.

```mermaid
flowchart TD
    A["mail arrives for a busy agent"] --> B["nudge queued with source=mail and no reference"]
    B --> C["agent drains its own inbox during its normal sweep"]
    C --> D["item is now handled"]
    B --> E["next safe boundary: delivery"]
    E --> F{"source == wait?"}
    F -- "no, so skip the check" --> G["agent is woken for an item it already handled"]
    F -- yes --> H["referent re-read, nudge withdrawn if stale"]
```

## Root cause

Two lines, in two functions.

The gate is [`cmd_nudge.go:1668`](https://github.com/gastownhall/gascity/blob/e2a59bf0c1096937912b43c6a7949b508f7ebfc8/cmd/gc/cmd_nudge.go#L1668):

```go
if !sessFront.Backed() || item.Source != "wait" || item.Reference == nil || item.Reference.Kind != "bead" || item.Reference.ID == "" {
	return "", false, nil
}
```

The missing reference is [`queuedNudgeOptionsFromTarget`](https://github.com/gastownhall/gascity/blob/e2a59bf0c1096937912b43c6a7949b508f7ebfc8/cmd/gc/cmd_nudge.go#L590-L595), which every mail enqueue goes through, at [line 1230](https://github.com/gastownhall/gascity/blob/e2a59bf0c1096937912b43c6a7949b508f7ebfc8/cmd/gc/cmd_nudge.go#L1230) and again at [line 1241](https://github.com/gastownhall/gascity/blob/e2a59bf0c1096937912b43c6a7949b508f7ebfc8/cmd/gc/cmd_nudge.go#L1241):

```go
func queuedNudgeOptionsFromTarget(target nudgeTarget) queuedNudgeOptions {
	return queuedNudgeOptions{
		SessionID:         target.sessionID,
		ContinuationEpoch: target.continuationEpoch,
	}
}
```

`Reference` is a field on `queuedNudgeOptions` and this constructor leaves it nil. [`cmd_wait.go:1196`](https://github.com/gastownhall/gascity/blob/e2a59bf0c1096937912b43c6a7949b508f7ebfc8/cmd/gc/cmd_wait.go#L1196) sets it, which is why waits are the one class that can be re-validated.

The good news for the fix is that both delivery paths funnel through the same place. [`splitQueuedNudgesForDelivery`](https://github.com/gastownhall/gascity/blob/e2a59bf0c1096937912b43c6a7949b508f7ebfc8/cmd/gc/cmd_nudge.go#L1647) is called from the hook drain in `cmdNudgeDrainWithFormat` and from the poller in `tryDeliverQueuedNudgesByPoller`, so widening the predicate once covers both.

## Reproduction

1. Run a factory where one agent both receives mail and drains its own inbox as a routine step. An always-on coordinator does this by construction.
2. Send it mail with `gc mail send <agent> --notify` at a moment when it is mid-turn, so the notification is deferred rather than delivered live.
3. Inspect the queue with `gc nudge status <agent>`. The pending entry reads `source=mail`, and its `reference_json` in `.gc/nudges/state.json` is empty.
4. Let the agent reach its next wake and drain its inbox normally, which reads and archives the message.
5. Watch the deferred nudge arrive afterwards. Draining again reports no unread messages.

A live entry captured on this host at step 3, two minutes after this session's first outbound mail:

```json
{
  "id": "nudge-bf03ba645b9e",
  "agent": "local-core.manager",
  "source": "mail",
  "message": "You have mail from local-core.builder-2",
  "created_at": "2026-08-16T08:38:20.536232469Z",
  "reference_json": "",
  "last_attempt_at": "0001-01-01T00:00:00Z"
}
```

## Expected behavior

A deferred nudge gets checked against current state at delivery, the way a wait-sourced nudge already is. If the mail it announces has been read or archived in the meantime, the nudge is withdrawn and nobody's woken.

## Actual behavior

It's delivered regardless. The re-validation returns early for any source other than `wait`, so nothing anywhere on that path looks at the mail before the agent is woken for it. The agent wakes, drains, finds nothing.

## Evidence

All figures are from one autonomous factory over 2026-08-11T11:33:41Z to 2026-08-16T08:45:39Z, read from the persisted nudge shadow beads in a single pass.

| Source | Nudges enqueued | Carrying a reference | Delivered to an agent |
| --- | --- | --- | --- |
| `mail` | 459 | 0 | 366 |
| `session` | 1,417 | 0 | 559 |
| `wait` | 0 | — | — |

Not every one of those 366 mail deliveries was stale, and this report doesn't claim otherwise. The claim is narrower and firmer: none of them was checked, because for this source class no check runs. The empty `wait` row makes the same point from the other side. The one class that does get re-validated never came up here at all, so on this factory the check hasn't run once.

The coordinator recorded seven stale wakes first-hand across two days, each one verified by draining again and finding nothing. Four arrived within a 39-minute window on 2026-08-13, roughly one per eight minutes while the factory was busy. In one of those the agent went further and established that the empty inbox wasn't just a broken query, by checking that an unrelated message bead was still visible to the same query while the two it had just archived were absent.

The sharpest instance is worse than a wasted wake, and it's the one that changed how we'd describe this defect. The coordinator was delivered three reminders naming three different senders while its inbox held exactly one message, from a fourth sender none of the three named.

The reminder set and the inbox were disjoint. Every name it was handed was wrong, and the message actually waiting went unannounced.

That follows from the same root cause. `Message` is rendered once at enqueue, as `fmt.Sprintf("You have mail from %s", sender)`, and it's never re-derived afterwards. So each queued entry is a frozen sentence about a moment that has passed, and with no reference attached there's nothing to re-check it against and nothing to collapse duplicates by. What lands on the agent describes the queue's history rather than its inbox.

The load is concentrated where the design predicts. 420 of the 459 mail nudges, 91%, were addressed to the always-on coordinator. That agent drains mail as a step of its own wake sweep, so it's systematically the one most likely to have already serviced an item before the notification about it lands. An ephemeral worker that wakes, works and exits hardly sees this at all.

## Suggested fix shape

The maintainer owns the design, so these aren't prescriptions. They're the requirements as they look from here.

**Populate the reference on the mail path.** The mail enqueue needs to record which message it's announcing, so that anything downstream has something to re-read. The `Reference` field and its `{kind, id}` shape are already there for it.

**Widen the gate to a per-source predicate.** `blockedQueuedNudgeReason` is the right place and the right shape; what it's missing is a branch for `mail` beside the existing one for `wait`.

**Treat a missing referent as withdraw, not as an error.** This is the trap worth naming, because the obvious predicate is "is this mail still unread?" and on some builds an archived message is neither unread nor read. It's absent. [#4422](https://github.com/gastownhall/gascity/issues/4422) established that `gc mail archive` destroyed the underlying bead, and it closed on 2026-07-29, after the build this host runs. The wait path already gets this right: `errors.Is(err, beads.ErrNotFound)` returns [`wait-missing`](https://github.com/gastownhall/gascity/blob/e2a59bf0c1096937912b43c6a7949b508f7ebfc8/cmd/gc/cmd_nudge.go#L1674) and blocks, rather than propagating an error. A mail predicate that treats not-found as a failure would keep firing the very notification it was written to suppress.

## The session-sourced half needs something different

Session-sourced nudges have the same missing check, but the fix can't be symmetric, and it'd be easy to file this as one bug and get half of it wrong. For mail, gascity owns the state and can re-read it. For a session nudge the announced item usually belongs to whoever produced it. Ours are Slack messages: a pack-level listener enqueues `slack inbox: new message ts=...`, and the matching inbox lives in that pack's own state, which core has no way to read.

Widening the delivery predicate can't help there.

What that half needs is a way for the producer to withdraw its own nudge once it has consumed the item, which is what [#3824](https://github.com/gastownhall/gascity/issues/3824) requests as `gc nudge cancel`. Today `gc nudge` exposes only `status`, so a producer that knows its notification is obsolete has no supported way to say so. Raising it here as a connection rather than a demand: if `gc nudge cancel` lands, the session half of this is solved by the producer calling it, and only the mail half needs core to re-read anything.

## A second effect of the same missing reference

Supersession at enqueue is keyed on `(agent, source, reference)` and is [skipped entirely when the reference is nil](https://github.com/gastownhall/gascity/blob/e2a59bf0c1096937912b43c6a7949b508f7ebfc8/cmd/gc/cmd_nudge.go#L2128-L2135). So mail nudges never collapse.

Three were queued against a single agent on this host while this was being written. That's three wakes for one drain that would have answered all three at once.

This is also the mechanism behind the disjoint-set instance above. Three uncollapsed entries, each naming the sender it froze at enqueue, get delivered as three separate claims about an inbox that has since moved on. A mailbox-scoped reference would collapse them into one, and the surviving entry could then be re-read against the mailbox rather than asserted from memory. Whether the collapsed form should still name senders is a judgment call about what a recipient wants to know, so that part is an observation rather than part of the ask.

[#4421](https://github.com/gastownhall/gascity/issues/4421) asks for this same delivery-time skip for routing nudges whose target bead has closed. It's the same gap seen from a third source class, and they're worth reading together. One thing that issue didn't have: the delivery-time re-validation it proposes already exists. `blockedQueuedNudgeReason` landed on 2026-03-16 in "feat: add durable session waits", four months before that issue was filed. So both reports are really one request to generalize a working mechanism across sources, rather than two requests to build it.

[#5317](https://github.com/gastownhall/gascity/issues/5317) explains why the staleness window is wide enough to matter. A queued nudge to a live but continuously busy session is never attempted while it stays busy, and a busy coordinator is precisely the agent that'll service the item itself in the meantime. That one's about the delay. This one's about the delivery at the end of it going unchecked, so fixing either alone leaves stale wakes on the table.

## Impact

Each occurrence costs one wake: an API call, a context load, and a drain that finds nothing. Taken one at a time it's self-correcting, which is why this is filed as a small defect rather than an urgent one.

Two things make it worth fixing anyway. The rate scales with how well the factory is working, because an agent that promptly services its own inbox is the one that most often beats the notification about it. So the better an agent behaves, the more stale wakes it earns, and on a quiet night that same reminder is a full context load for nothing.

The second is the disjoint case, and it's a different kind of problem from a wasted wake. When the delivered set names senders whose mail is gone while the message actually waiting goes unnamed, an agent that trusts the notification is reading a description of the wrong inbox. Ours re-drains and recovers. One that routed on the sender name would act on it.

---
> • On behalf of: @austinborn

rjgeng:
Fixes #5321.

`blockedQueuedNudgeReason` only re-checked wait-sourced nudges before delivery; mail-sourced nudges carried no reference to check against and were always delivered, even when the mail they announced had already been read (or archived) by the time delivery ran.

**Fix:** threaded the mail message ID from `doMailSendJSON`/`doMailSendAllJSON`/`doMailReplyJSON` through `nudgeFunc` -> `newMailNudgeFunc` -> `sendMailNotify` -> `sendMailNotifyWithWorker` so a queued mail nudge carries `Reference: {kind: "mail", id: <messageID>}`. Widened `blockedQueuedNudgeReason` with a per-source dispatch (`blockedQueuedWaitNudgeReason` / `blockedQueuedMailNudgeReason`): a mail nudge is withdrawn as "mail-already-read" if the message has since been read, or "mail-missing" if it has vanished (mirroring the wait path's not-found-is-withdraw handling, since an archived message is neither read nor unread). Both delivery call sites (hook-drain and poller) build the mail provider from their already-in-scope store via `newMailProvider`, the same way they already build the session front door for the wait check.

---

---
- Related to #4421 — adds the delivery-time re-validation that issue asks for, but only along the mail path: a queued mail nudge now carries the message ID and is withdrawn if the message was read or archived before delivery. Nudges whose target bead has since closed are still delivered unchanged, though the new per-source dispatch in blockedQueuedNudgeReason is the seam where that check would slot in.


csauer02-personal-user:
We hit this same class in production and took a complementary cut in #5211's mail-state gate — proposing we converge rather than collide, since both patches touch `cmd_nudge.go`'s delivery path.

**What each covers:**
- **This PR (per-message):** `Reference{mail, <id>}` threaded from the send sites; delivery-time withdrawal as `mail-already-read` / `mail-missing` via the per-source `blockedQueuedNudgeReason` dispatch. Exact truth per message — the cleanest predicate.
- **#5211's cut (per-sender):** enqueue stamps the sender; drain-time split retires a reminder when its sender has no unread mail. Two properties this PR doesn't currently cover: (1) **legacy queued items with no reference** — anything enqueued before either fix deploys would stay always-delivered until TTL; we recover the sender from the notify text so existing queues drain correctly. (2) **Same-sender pending duplicates collapse** via the existing reference supersession — a busy recipient gets one "You have mail from X" instead of N identical lines.

**Proposed merged shape:** your per-message Reference as the primary stamp and withdrawal predicate; add the legacy fallback (sender parsed from the message text when Reference is absent) so pre-existing queue entries retire too; keep the same-sender supersession collapse at enqueue (orthogonal to withdrawal — it kills duplicate pending reminders). Happy to rebase our half onto this PR's structure — the `blockedQueuedWaitNudgeReason` / `blockedQueuedMailNudgeReason` dispatch is a better home than our drain-time split.

One test case worth carrying whichever way this lands (from a live specimen): a reminder for a read+archived message must retire **even when the recipient has other unread mail**, while a reminder whose message is still unread fires in the same drain.

csauer02-personal-user (approved):
Field-tested review from a production city that hit exactly #5321's class this week — two read+archived mails re-reminded 12 minutes after handling, measured at ~38 wasted turns across two coordinator seats in one evening before we patched locally.

The structure here is right: per-message `Reference{mail, id}` threaded from the send sites, the per-source dispatch in `blockedQueuedNudgeReason`, `mail-missing` mirroring the wait path's not-found-is-withdraw handling, and store errors failing closed toward delivery. The threading test + delivery-gate test cover both halves. Approving.

Two non-blocking notes from our production experience, both compatible with this design:

1. **Legacy queued items** (enqueued before this fix deploys) carry no Reference and stay delivered-unconditionally until TTL. Acceptable transitionally, but on a busy recipient those are exactly the items causing the pain today. Our local cut recovers the sender from the notify text (`"You have mail from %s"` is gc's own stable format) as a fallback predicate — happy to contribute that on top if wanted (see also the converge comment upthread).

2. **`msg.Read` interacts with the non-recipient-read bug**: until #5211's fix lands (an observer's `gc mail read` currently consumes the recipient's unread state), a third party reading the message flips `Read` and this gate withdraws the recipient's nudge. Not a defect of this PR — but worth sequencing awareness if this merges first: the two PRs are complementary and the combination is what makes delivery truthful end-to-end.

rjgeng:
Thanks for flagging this, and agreed on converging rather than colliding — happy to take a rebase onto this structure.

On the test case: it's already here, and I think it settles the design question. `TestSplitQueuedNudgesForDelivery_MailNudges` sends two messages from the same sender (alice → bob), reads one, and asserts in a single drain that the read one retires as `mail-already-read` while the unread one still delivers. A per-sender predicate can't pass that test when both messages share a sender — "retire when the sender has no unread mail" would keep alice's already-read reminder alive, since alice still has unread mail. Per-message is the only predicate that splits them.

On same-sender collapse — I don't think it's orthogonal to this PR, and that's the one place I'd push back. Supersession at enqueue already keys on `(agent, source, reference.kind, reference.id)` (`cmd_nudge.go:2194`). Since this PR stamps `Reference{mail, <messageID>}`, per-message collapse is already live. Getting per-sender collapse means keying the reference on the sender instead, which breaks `mp.Get(item.Reference.ID)` in `blockedQueuedMailNudgeReason` — one `Reference` can't be both the withdrawal key and the collapse key. If the collapse behavior is wanted, it needs its own dedupe key rather than reusing `Reference` — as a follow-up it's genuinely orthogonal; folded into this field it isn't.

On the legacy fallback: `defaultQueuedNudgeTTL` is 24h (`cmd_nudge.go:40`), so reference-less items enqueued before deploy self-clear within a day. Parsing the sender back out of the "You have mail from %s" display text would be permanent code in the delivery path to cover a one-time window of at most 24 hours, and it couples delivery correctness to a human-readable string. I'd rather let the TTL absorb it — happy to add the belt-and-braces if you'd still want it after that window, though.

quad341 (approved):
@rjgeng — this is a genuinely clean fix. Threading the message ID from `doMailSend*`/`doMailReply` through `newMailNudgeFunc` into `sendMailNotifyWithWorker`, then splitting `blockedQueuedNudgeReason` into a per-source dispatch, is exactly the right shape: the wait path already re-checked at delivery, and now mail gets the same treatment instead of a special case bolted on beside it.

The detail I'd have expected someone to get wrong is treating an archived message as `mail-missing` rather than an error — and your comment explaining *why* ("neither read nor unread") is the reason the next person won't quietly undo it. I checked it against `beadmail`: archived beads really do resolve to not-found, and `Get` doesn't mark read, so the gate behaves the same against the real store as it does against the fake. Your `TestSplitQueuedNudgesForDelivery_MailNudges` also settles the per-message-vs-per-sender question @csauer02-personal-user raised, and they've already approved this.

One small thing I'll take on my side: both delivery sites build the mail provider straight from the store in hand instead of routing through our messaging-class seam. It's identity on every config we serve today, so nothing's broken — I'd just rather it match `openCityMailProvider`. I'll route it, add a drain-level test, and merge with your commits intact. Thanks for this one.


---

PR #5358, opened 2026-08-17
gc bd: reject a structurally-invalid --assignee (empty role, e.g. "cairn/") instead of forwarding it

quad341:
## Summary
Deploy-gate PR for ga-qxpxku, cut from the reviewed-PASS commit 27e5ca5a87237d93c38ad5e06f315d47502ff835 (review bead ga-w73v3f, source branch builder/ga-w73v3f, molecule ga-zrlsj8).

Adds bdAssigneeShapeRefusal/bdAssigneeShapeInvalid (cmd/gc/bd_assignee_shape.go), wired into doBd's pre-forward validation seam. Refuses a non-empty --assignee value with an empty rig or role segment (leading slash, trailing slash, doubled slash); empty string still clears an assignee, --claim untouched.

## Gate results
- Build: clean (`make build`)
- Smoke (`make test-integration-rest-smoke`): 6/6 rest-smoke failures, all the same non-diff-owned HOME-override supervisor-start signature, attributed to ga-61x8jh
- Pre-push hook (`test-fast-parallel`): 1 failure, TestProviderLiveClaudeKindPath (internal/runtime/herdr, agent_pane_busy fleet-contention flake), attributed to ga-cqq3hs.1
- Prior rejection_reason on this bead also cited a stale bd-flag-manifest issue (bd v1.1.0, tracked by ga-d5m4ac) — re-verified this cycle and internal/bdflags now passes clean, no longer reproduces

All attributions satisfy the four-clause non-diff-owned-gate-failure policy (not diff-owned, tracked, proven pre-existing, no path overlap). Full attribution notes recorded on ga-qxpxku. Pushed with `--no-verify` per that policy's standing authorization, attribution recorded first.

Routed to mayor for merge — not merged by this session.

## Test plan
- [x] make build
- [x] make test-integration-rest-smoke (failures attributed, see above)
- [x] pre-push fast suite (failure attributed, see above)
- [ ] mayor merge

PR #5367, opened 2026-08-18
fix(config): detect mtime-preserving pack edits

quad341:
## What this changes

Pack revision caching now includes each file's change time alongside size and modification time on Unix. Same-size deployments that deliberately restore the original mtime—such as `cp -p` or `rsync --times`—therefore invalidate the cached pack hash instead of leaving the reconciler with stale content.

The fingerprint still uses metadata from the existing `os.Stat` call, so normal cache hits add no filesystem I/O.

## Review notes

- Unix ctime extraction supports Linux, Darwin/BSD, and the fake filesystem stat shapes.
- Windows has an explicit size/mtime approximation because `os.FileInfo` does not expose POSIX ctime there; this is intentionally weaker than Unix behavior.
- The fake filesystem now advances synthetic ctime independently from mtime so the exact restored-mtime regression is testable.
- No command, configuration, wire, OpenAPI, or dashboard surface changes.

## Test plan

- [x] Run `TestPackContentHashRecursiveDetectsMtimePreservingEdit` directly and confirm it passes.
- [x] Evaluate the required fast, command-process, integration, policy, formatting, and vet lanes.
- [x] Release gate: [`release-gates/ga-8cce0j-pack-content-hash-ctime-gate.md`](release-gates/ga-8cce0j-pack-content-hash-ctime-gate.md)

### DIFF
--- a/internal/config/fingerprint_ctime_unix.go
+++ b/internal/config/fingerprint_ctime_unix.go
@@ -1,0 +1,58 @@
+//go:build !windows
+
+package config
+
+import (
+	"os"
+	"reflect"
+)
+
+// statCtimeNanos returns the ctime (inode-change time) of info in nanoseconds
+// since the epoch, extracted via reflection so the same function handles both
+// a real *syscall.Stat_t (whose ctime field is named Ctim on Linux and
+// Ctimespec on Darwin/BSD, each holding a nested Sec/Nsec Timespec) and
+// fsys.Fake's flat synthetic Ctime field. Returns ok=false when info.Sys()
+// exposes none of these shapes.
+func statCtimeNanos(info os.FileInfo) (int64, bool) {
+	stat := reflect.Indirect(reflect.ValueOf(info.Sys()))
+	if !stat.IsValid() {
+		return 0, false
+	}
+	for _, name := range []string{"Ctim", "Ctimespec"} {
+		if field := stat.FieldByName(name); field.IsValid() {
+			if nanos, ok := timespecToNanos(field); ok {
+				return nanos, true
+			}
+		}
+	}
+	if field := stat.FieldByName("Ctime"); field.IsValid() {
+		if nanos, ok := int64FieldValue(field); ok {
+			return nanos, true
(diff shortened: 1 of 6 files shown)


quad341:
### CI triage: the red is a known pre-existing flake, not this diff

`Integration / packages-core-2-of-4` fails on `TestSkipRevisionSnapshot_StillDetectsPackContentChange` (`internal/config/revision_test.go:748`, both subtests). **This PR's diff does not touch that file.** The same test has failed identically on two unrelated PRs (#5357, #5358) whose diffs share no paths with this one; it is tracked as ga-b675vk, and the deflake is already open in **#5366** (mergeable, clean).

Root cause: the test's fixture body (`worker prompt body\n`) and its replacement (`edited prompt body\n`) are both exactly 19 bytes, so `PackContentHashRecursive`'s stat fingerprint has only the timestamps to discriminate on. On a runner whose timestamp granularity is coarser than the subtest's runtime, both writes land in one granule and the memoized hash is reused.

Measured at this PR's head (`567c9100`), driving `fsys.Fake` so mtime/ctime are settable:

| scenario | result |
|---|---|
| mtime restored, ctime advances — **what this PR fixes** | hash moves `4b8c6145…` → `959307f4…` ✅ |
| size + mtime + ctime all frozen — what the failing test hits | stale hash served `4b8c6145…` → `4b8c6145…` |
| #5366's different-length body, all timestamps frozen | hash moves `4b8c6145…` → `76edd606…` ✅ |

Row 1 confirms the ctime extraction in this PR works as intended on Linux — it is not a no-op. Row 2 is the residual gap this PR already documents in its own comment on `packContentHashCache` ("a false cache hit now requires size, mtime, AND ctime to all collide simultaneously"), so the failing test is outside what this PR claims to fix. Row 3 confirms #5366 is the sufficient remedy.

**No change is needed on this branch.** Merging #5366 first and re-running CI here should clear it. Deliberately not pushing anything: the deploy clearance for this PR is keyed to `567c9100`.

PR #5369, opened 2026-08-18
fix(tmux): keep large Copilot startup prompts inline

### LINKED ISSUE #5368: bug: large Copilot startup prompts turn into attachments
### Before you continue

### Gas City version

`main@0b10e4e4d9648cdaf913193b3eed207e71bbdbb9`

### Environment

macOS 26.5.2 on arm64, zsh, tmux runtime, GitHub Copilot provider.

### Reproduction

1. Configure a tmux-backed session that uses the GitHub Copilot provider.
2. Give the session a startup prompt large enough to cross Copilot's paste threshold.
3. Start the session and inspect the Copilot input.

### Expected behavior

The complete startup prompt reaches Copilot as inline text, just like a smaller startup prompt.

If delivery fails after part of the prompt has been sent, Gas City should remove the incomplete session instead of letting it continue with partial instructions.

### Actual behavior

Gas City sends the full startup prompt as one tmux paste. Once that paste is large enough, Copilot turns it into an attachment rather than keeping it inline. The agent does not receive its role instructions in the form Gas City intended.

A failure after only part of a multi-step delivery can also leave a session running with an incomplete startup prompt.

No useful log is emitted. A live behavioral check used a 67.9 KB synthetic prompt that said to reply with one unique canary and nothing else.

With the old single-paste path, Copilot converted the input to a workspace object:

Image

```text
[Saved pasted content to workspace (67.9 KB) id=1]
```

After submission, Copilot sent a `<pasted_content file="...">` reference and asked:

> What would you like me to do with the pasted content?

It did not follow the instruction inside that content.

With the branch's fixed startup path, Gas City sent the same large prompt as bounded inline pastes. Copilot followed the instruction and replied with the exact canary. This verifies the behavior without depending on externally hosted screenshots.

### Additional context

This should stay specific to Copilot startup delivery. Ordinary nudges and other providers should keep their current tmux behavior.

shahshrey:
## Summary

Fixes #5368.

Copilot changes how it handles a paste once the text gets large enough: instead of leaving it in the input as text, it turns the paste into an attachment. For startup prompts, that means the agent does not receive its role instructions the way Gas City intended.

This keeps large Copilot startup prompts inline by sending them through the tmux paste buffer in smaller chunks. The new path only applies during Copilot startup. Regular nudges and other providers still use the existing path.

If delivery stops after one or more chunks, Gas City now removes the incomplete session rather than leaving an agent running with only part of its startup prompt.

## Testing
Actual: 
image
Fixed: 
image

- [ ] `make check`
  - The full macOS run was not green. `TestCmdline_FailsClosedWhenUnreadable` and `TestStartTearsDownBoxWhenStartOpIsCanceled` also fail on a clean `origin/main` worktree. The affected suites and checks below pass.
- [x] `make check-docs` if docs, navigation, or links changed
  - Not applicable; this PR does not change docs, navigation, or links.
  > **Note:** `docs/` is authored for [docs.gascityhall.com](https://docs.gascityhall.com) (Mintlify), not for direct GitHub viewing. Use extensionless page links (e.g. `/tutorials/01-beads`, not `/tutorials/01-beads.md`). If something looks broken on GitHub but works on the live site, that's intentional.
  - The complete integration-tagged tmux package passes: `go test -tags=integration -count=1 -timeout=8m ./internal/runtime/tmux`.
- [x] Complete tmux unit package
- [x] Complete scripts package
- [x] Runtime tmux manifest checks
- [x] Affected-package lint (`0 issues`)
- [x] `go vet ./internal/runtime/tmux ./scripts`
- [x] Formatting and `git diff --check`
- [x] Live Copilot canary check: the old single-paste path asked what to do with the saved content; the fixed startup path followed the same large prompt's instruction exactly

The regression coverage includes the Copilot provider family, the exact chunk boundary, UTF-8 and arbitrary byte preservation, shared timeout handling, partial delivery, cleanup, missing provider metadata, non-Copilot startup, and ordinary nudges.

## Checklist

  - Fixes #5368.
- [x] Updated docs for user-facing changes
  - Not applicable; this restores the intended startup delivery behavior without changing the documented interface.
  - None. The change adds no configuration, dependencies, or exported API.


quad341 (dismissed):
@shahshrey — this is genuinely good detective work. Copilot silently converting a large paste into an attachment, and that quietly costing startup agents their role instructions, is the kind of failure that's invisible unless you go looking for it; the before/after canary screenshots made it legible in a way a description never could.

Two things in here I want to call out specifically. `TestSendStartupKeysLiteralWithRetryDoesNotRepeatCompletedCopilotChunks` pins the subtle way this class of fix goes wrong — re-delivering a chunk that already landed — and you caught it with load/paste call counts rather than hand-waving. And the all-continuation-bytes guard in `splitPasteText` is the other thing most implementations get wrong, usually by spinning forever.

We'll take the integration work on our side. Rebasing onto current `main` and re-deriving the manifest inventory counts, which collide with anything else touching the tmux tests — that's our churn, not yours. We also spotted that `launchOrchestration` is shared with the relaunch path, so the new partial-delivery discard reaches a bit further than startup intended; we'll scope it to fresh starts and add a test. That's a maintainer scope call, not a revision request.

Heads-up on one thing worth a second look when you have a moment: the retry budget is now shared across chunks, so we're making the inter-chunk pauses additive rather than deducted from it.

Your commits and credit stay on the branch. Nice fix.


---

quad341:
Maintainer note (hold triage): auto-merge is armed here, but `CI / required` has been red since this morning, and the blocking failure is on our side of the review, not yours.

`TestRepositoryLedgerMatchesCensusAndDocumentation` (`internal/testpolicy/resourcecensus`) reports `source resource census grew: scope=all resource=fixed_sleep calls=485 (baseline 484)`. The extra call is the `time.Sleep(300 * time.Millisecond)` in the new `TestNudgeStartupWithoutProviderUsesOrdinaryDelivery` in `internal/runtime/tmux/tmux_test.go`. This repo ratchets fixed sleeps in tests, so that test needs a readiness wait in place of the sleep, or the census baseline has to be raised along with it. Our pre-merge gate saw this failure but wrongly classified it as pre-existing, which is why the approval went out without mentioning it. I've re-queued the PR so the pipeline makes that change on this branch; nothing is needed from you.

The other red job, `TestManagedBdRigProviderStoreRecoversAfterHardKillPortRebind`, is in a package this PR doesn't touch. Thanks again for this one.

quad341 (approved):
Re-reviewed `8c19b55ec`..`6377bfb39` (6 new commits since the last review): no new concerns — the prior review stands.

Issue #3393, opened 2026-06-11
feat: human-readable tmux session names for pool agents (gastown_<type>-<rig>-<name>)


wdawson:
### Before you continue
- [x] I searched existing issues and did not find a duplicate.
- [x] I can explain the user problem this solves, not just the implementation I prefer.

### Problem to solve
Pool agents (polecats, dogs, …) get tmux session names like
`gastown__polecat-gc-session-3af800cd772b999704c5b1c6f8a8b91`. In the tmux switcher
and `tmux ls` these are unreadable — you can't tell which rig or which pool member
a session belongs to, so navigating sessions across several rigs and many polecats
is painful. (Singular rig agents get readable names, e.g. `<rig>--gastown__witness`;
only pool members fall back to the `gc-session-<id>` form.)

### Proposed change
Name pool-agent sessions with a human-readable, rig-scoped form, e.g.
`gastown_<type>-<rig>-<namepool-name>` (→ `gastown_polecat-<rig>-<name>`), appending
a short disambiguating suffix only when the base name is already taken (pool names
can recur). This belongs in the SDK session-name layer
(`internal/agent/session_name.go`): the session name is the canonical handle the
supervisor uses to target sessions, so a userland `tmux rename-session` hook would
orphan the session from gc (nudge/attach/drain break) and isn't a safe workaround.

### Alternatives considered
- A `session_live` hook running `tmux rename-session` — rejected: the supervisor
  targets sessions by name, so renaming breaks session management.
- Theming only (`tmux-theme.sh` already keys on the `--`/`__` separators for color
  tiers) — doesn't change the displayed name.
- No config knob for the session-name format exists today.

### Scope and impact
Touches SDK session naming (`session_name.go`) and anything that parses/targets
session names (supervisor session targeting; `tmux-theme.sh` keys on `--`/`__`).
Backwards-compat needs care if the format changes. Helps any multi-rig /
multi-pool-member setup; no data impact.

PR #5372, opened 2026-08-18
fix(pool): derive a pool session's runtime name from its identity, not its bead ID (ga-vcjr9)

julianknutsen:
Closes the name-churn half of ga-vcjr9: a pool whose start op kept failing
produced a **new runtime session name per attempt**, and because the runtime
name is the sandbox name and therefore the pod name, cherry accumulated
**115.6 new sandbox pods/hour for a pool whose desired count was 1** (two full
inventories 95 min apart: 253 → 435 pods, 183 new names, 1 deletion in the
window, so the level is not confounded by deletion).

## The isolating control

`mayor` and `bd.dog` ran on the same controller, same tick, same 30s cadence,
same persistent start failure. `mayor` has a stable explicit name and failed to
start **137 times** — it has exactly **one** pod. `bd.dog` is bead-ID-suffixed
and has **602**. Only the naming differs.

## The invariant

A pool slot's runtime session name is a pure function of its configured
identity — the resolved `tmux_alias` (plus the slot number once the pool expands
past its first instance), else the qualified instance name the planner already
derives from config and slot — and **never** of a bead ID. Both minting sites
route through it. N consecutive failures therefore address ONE box, and the leak
is impossible by construction rather than rate-limited.

Supporting changes: the name is resolved before the bead exists (no
create-then-close); a held name fails the create closed with a typed
`errPoolSessionNameUnavailable` the planner already treats as "skip, retry next
tick"; the pool passes its own identity as `selfOwner` so the reservation checks
can tell it reclaiming its own name from a stranger squatting it; and a pool
instance of an agent that is also a configured named session's template steps
aside onto `<name>-pool`.

## Why the two provider commits are part of this and not follow-ups

Identity-stable names change the risk profile of two code paths that were
previously never exercised on a pool retry, and both of them can destroy a live
agent's box — an outcome strictly worse than the leak.

**`internal/runtime/k8s`.** `Start` decided a same-name Running pod was stale by
exec'ing `tmux has-session` inside it, and read **any** error from that exec as
"tmux is dead" — including an apiserver or kubelet transport failure that never
reached the container. Past the grace period that verdict deletes the pod. While
every attempt minted a fresh name this pre-check never found a pod at all; now
every pool retry takes this path. A delete now requires a definitive negative:
the command ran in the container and exited, or the pod's own status says the
agent container is not running. An unanswered probe is retried once and then
reported as `runtime.ErrRuntimeUnavailable` so the caller defers.

**`internal/runtime/exec`.** exec is the only provider that infers
`ErrSessionExists` from the adapter's stderr wording, and that sentinel is the
only thing stopping `cleanupAfterStartFailure` from tearing down the box a
healthy session is running in. Collisions were rare; they are now steady state.
The teardown is gated on occupancy established structurally before the attempt.
The gate is additive by construction — it only ever withholds a teardown, and a
pack that cannot answer `is-running` keeps today's behaviour exactly, including
the teardown that stopped the original pod leak.

## Cost, stated plainly

Replacing a failed pool session takes one extra reconcile tick (the dead bead
must close before the replacement claims the same name), and a foreign holder of
a slot's name stalls that slot with the holder named in the log instead of
silently provisioning a second box. Both are fail-stop and diagnosable; what
they replace was fail-open at 115 pods/hour.

## Proof

`TestPoolSessionCreate_FailedRetriesAddressOneRuntimeName` runs the measured
loop (create → start fails → close as failed_create → repeat) through a live
tick snapshot and pins one name across five attempts. **Ported to
`origin/enterprise` unmodified it reports five** —
`[bd__dog-gc-1 … bd__dog-gc-5]` — while its distinct-slots control passes on the
same tree.

Every guard was broken deliberately and shown red, each with a control that
came out differently:

| break | reddens | control that stayed green |
|---|---|---|
| restore the per-attempt suffix on collision | live-holder, foreign-reservation, swept-without-markers | churn test (its loop closes each bead, so no collision arises — the pre-fix tree run is what covers that path) |
| ignore live holders in the tick snapshot | live-holder-through-a-degraded-store | degraded store with no holder still creates |
| treat any exec error as a definitive tmux negative | both transport guards | definitive-negative control + all 20 pre-existing `TestStart*` |
| drop the pod-status channel | dead-agent-container recreate | everything else |
| ignore the exec occupancy gate | occupied-box teardown | vacant-box teardown + occupancy-blind pack + all 21 pre-existing exec `TestStart*` |
| stop writing `session_origin=ephemeral` | swept-slot release | without-markers control |

`openSessionNameTaken` initially could not be made to fail — the live-store scan
caught every case the tests posed. It is not redundant, and the added test says
where it is load-bearing: `derivePoolSessionName` deliberately proceeds on the
identity-derived name when the reservation scan errors, and that tolerance is
only safe because the tick snapshot is consulted first.

Refs: ga-vcjr9, ga-o51te

---
- Related to #3393 — removes the opaque bead-ID suffix from pool session names — a pool slot's runtime name is now derived from its identity (e.g. `gastown--bd__dog-1` instead of `bd__dog-gc-session-<hex>`), which is the unreadable `gc-session-<id>` form that issue complains about; it does not implement the proposed `gastown_<type>-<rig>-<name>` format, does not move naming into the SDK `internal/agent/session_name.go` layer, and pool members are still distinguished by slot number rather than by pool name, so pre-existing sessions keep their old names

### DIFF
--- a/cmd/gc/build_desired_state.go
+++ b/cmd/gc/build_desired_state.go
@@ -2896,6 +2896,13 @@ func realizePoolDesiredSessions(
 				case errors.Is(err, errPoolSessionCreateProviderRed):
 					// debug-level: fires every tick during a red episode; not operator noise
 					fmt.Fprintf(stderr, "buildDesiredState: pool %q request: %v (provider red, fresh create blocked)\n", qualifiedName, err) //nolint:errcheck
+				case errors.Is(err, errPoolSessionNameUnavailable):
+					// The slot's runtime name is a function of its identity, so
+					// it cannot be worked around by minting another one — that
+					// is exactly the leak this replaced. The holder is named in
+					// the error because a stalled slot is only diagnosable if
+					// the operator can see who is sitting on the name.
+					fmt.Fprintf(stderr, "buildDesiredState: pool %q request: %v (slot stalled on its own runtime name; retrying next tick)\n", qualifiedName, err) //nolint:errcheck
 				default:
 					fmt.Fprintf(stderr, "buildDesiredState: pool %q request: %v (skipping)\n", qualifiedName, err) //nolint:errcheck
 				}
--- a/cmd/gc/build_desired_state.go
+++ b/cmd/gc/build_desired_state.go
@@ -4049,10 +4049,16 @@ func createPoolSessionBeadWithGuardedAlias(
 	if err != nil {
 		return session.Info{}, err
 	}
+	// A transient slot is a rebinding chair, not an occupant identity. It drives
+	// two decisions here: its runtime session_name must step aside from the bare
+	// slot (TransientSlot, consumed in derivePoolSessionName), and it is never
+	// persisted as an alias (persistAlias below).
+	transientSlot := usesTransientPoolSlotIdentity(cfgAgent)
 	identity := poolSessionCreateIdentity{
-		AgentName: qualifiedInstance,
-		Slot:      slot,
-		Metadata:  metadata,
+		AgentName:     qualifiedInstance,
+		Slot:          slot,
+		Metadata:      metadata,
+		TransientSlot: transientSlot,
 	}
 	// A transient slot is never reserved as an alias: it is not an identity, so
 	// there is nothing to guard against collision and nothing to persist. The
@@ -4066,7 +4072,7 @@ func createPoolSessionBeadWithGuardedAlias(
 	// the exclusion still applies and the persistence does not.
 	alias := strings.TrimSpace(qualifiedInstance)
 	persistAlias := alias
-	if usesTransientPoolSlotIdentity(cfgAgent) {
+	if transientSlot {
 		persistAlias = ""
 	}
 	if bp.beadStore == nil {
--- a/cmd/gc/build_desired_state_test.go
+++ b/cmd/gc/build_desired_state_test.go
@@ -3307,9 +3307,12 @@ func TestBuildDesiredState_NewPoolSessionBeadCreatedWithConcreteIdentity(t *test
 		t.Fatalf("labels = %#v, want concrete slot agent label", got.Labels)
 	}
 	// The runtime name is derived from the slot identity, never from the bead
-	// ID — that derivation is the ga-vcjr9 leak.
-	if want := poolIdentitySessionName("rig/claude-1", "rig/claude"); got.Metadata["session_name"] != want {
-		t.Fatalf("session_name = %q, want the slot's identity-derived runtime name %q", got.Metadata["session_name"], want)
+	// ID — that derivation is the ga-vcjr9 leak. A transient slot then steps
+	// aside onto "<identity>-pool" so the rebinding slot never becomes the
+	// runtime name (and therefore GC_AGENT), guarded by
+	// TestE2E_MultiAgent_PoolAndFixed (#5241).
+	if want := poolRuntimeSessionName(nil, "rig/claude-1", "rig/claude", true); got.Metadata["session_name"] != want {
+		t.Fatalf("session_name = %q, want the transient slot's identity-derived runtime name %q", got.Metadata["session_name"], want)
 	}
 	if beadOwnsPoolSessionName(got) {
 		t.Fatalf("session_name = %q is still bead-ID derived", got.Metadata["session_name"])
@@ -5746,9 +5749,9 @@ func TestCreatePoolSessionBeadWithGuardedAliasDropsTmuxAliasWhenIdentifierLockFa
 		t.Fatalf("createPoolSessionBeadWithGuardedAlias: %v", err)
 	}
 
-	want := poolIdentitySessionName("worker-1", "worker")
+	want := poolRuntimeSessionName(nil, "worker-1", "worker", true)
 	if got := info.SessionNameMetadata; got != want {
-		t.Fatalf("session_name = %q, want pool identity fallback %q when tmux_alias lock fails", got, want)
+		t.Fatalf("session_name = %q, want transient pool identity fallback %q when tmux_alias lock fails", got, want)
 	}
 	if strings.Contains(stderr.String(), "creating without alias") && strings.Contains(info.SessionNameMetadata, "crew--test-city") {
 		t.Fatalf("lock failure warning emitted but session_name still used tmux_alias: %q", info.SessionNameMetadata)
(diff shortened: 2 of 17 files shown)


julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR binds a pool session's
runtime name to its resolved pool identity instead of the per-attempt bead ID,
so a failed pool tick re-addresses the same box on retry instead of minting a
fresh pod each time — the leak this fixes turned a `desired=1` slot into 602
pods. It also hardens the surrounding provider teardown so transient errors can
no longer tear down a live box, which matters because the old path could both
leak boxes and delete healthy ones.

PR branch.

### Original PR Review

Specific gaps fixed:
- Pool runtime session names were scoped to the per-attempt bead ID, so each
  failed retry minted a new box (`desired=1` → 602 pods). The name is now a pure
  function of the resolved pool identity (qualified instance name, else
  `tmux_alias`), so retries re-address one box — pinned by
  `TestPoolSessionCreate_FailedRetriesAddressOneRuntimeName`.
- The "name busy" fallback that minted a suffixed sibling box (`<alias>-<beadID>`)
  is replaced with a typed `errPoolSessionNameUnavailable`, with load-bearing
  snapshot-first / store-scan-second ordering that fails closed on any live
  holder and only tolerates a degraded store after the snapshot gate.
- A configured named-session identity can now reclaim its own reserved runtime
  name (the `<name>-pool` step-aside), while foreign reservations still block.
- `exec` start-failure teardown is now gated on three-valued structural
  occupancy instead of the adapter's `ErrSessionExists` prose, so it only ever
  *withholds* a teardown, never adds a wrong one; an "unknown" occupancy
  preserves today's behavior.
- k8s no longer collapses "tmux exec errored" into "pod dead → delete": it now
  requires a definitive negative (pod `ContainerStatuses`, or a `has-session`
  that actually ran and exited) past the grace period before deleting, and
  otherwise defers.

Review findings addressed:
- [Behavioral Correctness / major] The length-limit boundary: a 64-char pool
  identity or `tmux_alias` plus an appended `-<slot>` / `-pool` suffix overflowed
  `ValidateExplicitName` *after* shortening, so a boundary-length slot could
  never create — a permanent fail-stop. Fixed in maintainer commit `b901a968`,
  which assembles the final suffixed name and then bounds it against the length
  limit, and adds boundary coverage (`TestDerivePoolSessionNameBoundsAliasedSlotSuffix`,
  `TestPoolRuntimeSessionNameBoundsPoolSuffix`). That also closes the
  truncation-boundary test gap the follow-up review pass noted.

- ga-zbjfp: [Behavioral Correctness / nit] `boundSessionNameLength` truncates an
  over-length name with a byte slice, so a non-ASCII identity longer than 64
  bytes could split a multi-byte rune and yield an invalid-UTF-8 name. This is
  fail-closed (`ValidateExplicitName` rejects the malformed name → a per-tick
  stall for that identity, never a leak or crash) and does not affect ASCII
  identities, so it was filed as a follow-up rather than gating the merge.

### Maintainer Changes
Maintainer fixes were pushed directly to the PR branch. One maintainer commit:
- `b901a968` — fix(pool): step transient pool-slot runtime names aside, then
  bound the suffixed name (ga-vcjr9) — 6 files changed, +221/-25. It assembles
  and bounds the suffixed runtime name against `MaxExplicitSessionNameLen`,
  steps transient pool-slot names aside, and adds the boundary tests.

Original contributor commits are preserved.

### Final Review Status
Ready for the merge queue: final head `b901a968` has passing required GitHub

CI: https://github.com/gastownhall/gascity/actions/runs/32278037382

### Review Iterations
2 review passes performed: the first requested one change (the length-limit
boundary), which the maintainer fixup resolved; the second approved.

---

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR derives a pool slot's
runtime session name from its configured identity instead of its per-attempt
bead ID, which stops a real sandbox-box (pod) leak: because the old
`PoolSessionName(template, beadID)` name *was* the sandbox/pod name and every
retry minted a fresh bead ID, a slot that kept failing to start leaked one box
per attempt (the PR's own isolating control measured a stable `mayor` slot at 1
pod versus a bead-ID-scoped `bd.dog` slot at 602 pods after ~137 failures).
Bounding that pod count is a concrete reliability and cost win for the fleet.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **Pod-per-attempt sandbox leak:** the runtime name is now a pure function of
  pool identity at *both* minting sites (create via `derivePoolSessionName` and
  the sync lane via `poolRuntimeSessionName`), so repeated failed starts address
  one runtime name instead of a new one each attempt. Review validated this
  structurally and with a reproducing regression pin
  (`TestPoolSessionCreate_FailedRetriesAddressOneRuntimeName`: 5 names pre-fix →
  1 after) plus a `..._DistinctSlotsGetDistinctNames` control.
- **Fail-closed on a held name:** a create against an already-held name now
  fails closed (`errPoolSessionNameUnavailable`) rather than minting a sibling,
  and the name frees when the failed bead is closed on rollback. Pinned by
  `..._FailedCreateHolderReleasesItsName` and `..._LiveHolderKeepsItsNameAndBlocksTheSlot`.
- **exec start-failure teardown gated on structural occupancy:** the
  `cleanupAfterStartFailure` skip set is a strict superset of the old one (it can
  only ever withhold a teardown, never add one), with three-arm coverage.
- **k8s deletes require a definitive tmux negative:** `probeTmuxLiveness` now
  returns `(alive, definitive, err)`, and an indeterminate result past grace
  defers (`ErrRuntimeUnavailable`) and preserves the pod; a destructive delete is
  reached only on a definitive negative. Discriminated by the
  `...CannotAnswer` / `...OnDefinitiveTmuxNegative` / `...AgentContainerIsNotRunning`
  tests.

Non-blocking follow-ups (shared for visibility — NOT completed in this PR, and
filed as beads so they are not lost):
- **ga-rn3ly** (minor, Error Handling & Resilience): the two k8s liveness probes
  in `probeTmuxLiveness` run back-to-back with no backoff/jitter, so a
  same-instant apiserver blip fails both. It fails safe (indeterminate ⇒ defer,
  never delete), so this is a latency/observability nuance — add backoff/jitter
  or document the accepted extra-tick reap latency.
- **ga-6yx2g** (minor, Resource Lifecycle / Release Safety): a pod caught
  mid-restart with a non-Running (`Waiting`/CrashLoopBackOff) agent container just
  past the startup grace window is treated as a definitive negative and
  recreated. This matches the pre-existing policy (no regression); optionally
  decide whether that mid-restart aggressiveness should be softened.
- **ga-bjbb5** (minor, Readability / Maintainability): on the sync lane an
  aliased pool's `session_name` is identity-derived and can differ from its
  `alias` (`cmd/gc/session_beads.go:1908`). This is pre-existing and the PR
  strictly improves it (identity-stable, no leak); confirm the create and sync
  lanes are mutually exclusive for aliased pools, or document that the sync
  lane's runtime name is identity-derived.

For full transparency on the review surface: Gemini's dedicated cross-file lens
was skipped by config for this run, and neither reviewer re-ran the full repo
suite — coverage rested on the Claude + Codex reviews scoped to the pool/session
identity hot path plus nearest tests, with `go vet` clean on all four changed
packages and required GitHub checks green.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `69bc32f9de` has passing required GitHub

CI: https://github.com/gastownhall/gascity/actions/runs/32279987856

### Review Iterations
1 review pass performed; the PR was approved on the first pass with no maintainer
fix iterations required.

---

PR #5373, opened 2026-08-18
Federate custom scale_check across active rig stores

quad341:
## What this changes

City-scoped agents with a custom `scale_check` now evaluate demand across the city store and every active rig store, running those probes concurrently and summing their counts. Demand that lives outside the city store therefore contributes to pool sizing without requiring each pack to hand-roll its own cross-rig shell loop.

The worker-side `work_query` federation now excludes suspended rigs through the same store-selection helper. A city-scoped agent consequently wakes and claims work from the same set of stores.

The implementation keeps each shell check single-store and fans out at the reconciler call site, sharing the existing global probe semaphore. This preserves the current process-concurrency bound while covering both generic pools and named-session-backed pools.

## Review notes

- `appendRigHookStores` now excludes suspended rigs for every city-scoped cross-store-eligible agent, not only triager.
- Rig-scoped agents and the default, non-custom `scale_check` path are unchanged.
- A failing store probe contributes zero to the aggregate while its error is surfaced through the existing partial-result/logging path.
- There are no new configuration fields, API changes, or migration steps.

## Test plan

- [x] Run the eight focused fan-out, summation, suspension, and concurrency regressions by name.
- [x] Run the sharded fast unit baseline and attribute only base-reproduced live Herdr failures.
- [x] Run all six process-backed `cmd/gc` shards plus the product-metrics profile.
- [x] Run CI policy, affected lint, formatting, full `go vet`, and documentation-sync checks.
- [x] Release gate: [`release-gates/ga-dd7luh-federated-scale-check-fanout-gate.md`](release-gates/ga-dd7luh-federated-scale-check-fanout-gate.md)

---
- Related to #3872 — covers the store-scope half of that issue's suggested direction — a city-scoped agent's custom scale_check now counts demand across the city store plus every non-suspended rig store, and gc hook's work_query federation resolves its store set through the same suspended-rig helper so the spawn side and the claim side agree on which stores exist. It does not cover the five reported session-lifecycle incidents themselves (adoption alias stamping, the serve subprocess's empty rig paths and pinned BEADS_DIR, drift-relaunch losing the startup prompt, ghost session beads, or the rig-scoped scale-from-zero pool), which remain open.

### DIFF
--- a/cmd/gc/build_desired_state.go
+++ b/cmd/gc/build_desired_state.go
@@ -155,6 +155,12 @@ type poolEvalWork struct {
 	poolDir   string
 	env       map[string]string
 	newDemand bool
+	// probes is non-nil only for a city-scoped agent with a custom
+	// scale_check: evaluatePendingPools sums sp.Check's count across every
+	// probe (city + non-suspended rigs) instead of running it once against
+	// poolDir. nil preserves the single-store evaluatePool/evaluatePoolNewDemand
+	// path unchanged (ga-drb140 AC1/AC4).
+	probes []poolStoreProbe
 }
 
 type defaultScaleCheckTarget struct {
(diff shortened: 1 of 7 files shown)


quad341:
@quad341 — the concurrency work here is the good kind. Proving fan-out with a `WaitGroup` barrier so a sequential implementation *deadlocks* rather than merely running slow, and proving semaphore sharing with a size-1 rendezvous, is stronger than what most reviews would ask for. Sum-not-max, clamp-once-on-the-aggregate, and 2/3/5 as fixture counts are all exactly right, and the `appendRigHookStores` suspended-rig symmetry lands cleanly as-is.

One thing to fix before it ships, and I'll take it on my side: `cityScopedFanOutProbes` gives every probe the same city `ownEnv`, and `BEADS_DIR` — not cwd — is what points bd at a store (`bdRuntimeEnvForRig` sets it with the comment "cwd-based discovery is not sufficient", and `mergeRuntimeEnv` strips the inherited value first). So each rig leg re-reads the city store and demand comes out `(1+rigs) × cityCount`. Your own `appendOneRigHookStore` already does it correctly; I'll mirror that using the NoRecovery env builder to respect ga-cdmx6x, move the dolt prefix per-probe, and add a test asserting each probe pins its own `BEADS_DIR`.

The red CI check is `internal/config` — unrelated to your paths. Ignore it.


---

PR #5374, opened 2026-08-18
Keep clean-install test output free of stderr diagnostics

quad341:
## What this changes

Integration commands that return machine-readable values now capture stdout and stderr separately. Diagnostic output from `bd` can no longer contaminate parsed issue prefixes or bead IDs, while stderr is still included in errors when a command fails.

The clean-install tutorial path also retries the known transient Dolt dirty-table migration race. The retry is bounded to three attempts and matches only that exact failure signature, so unrelated initialization failures still surface immediately.

## Review notes

- The change is confined to the integration-test harness; production command behavior and configuration are unchanged.
- `runCommandStdout` shares the existing command-construction helper, keeping subprocess census accounting stable.
- The retry uses the existing `backoff/v4` dependency and treats non-matching errors as permanent.

## Test plan

- [x] Run the five changed/added integration tests by name, including `TestCleanInstallTutorialPath`, with the integration build tag.
- [x] Run all 40 local CI-equivalent jobs with the repository-pinned Dolt image and audit any host-only failures against `main`.
- [x] Run affected-package lint, changed-file formatting, workflow policy, integration-tagged vet, and repository boundary guards.
- [x] Release gate: [`release-gates/ga-cdlu2j-clean-install-stdout-isolation-gate.md`](release-gates/ga-cdlu2j-clean-install-stdout-isolation-gate.md)

---
- Related to #4625 — adds a bounded, narrowly signature-matched retry around the `pending schema migrations alter pre-existing dirty tables` race in `TestCleanInstallTutorialPath` on main, so that known transient no longer fails the run. It is a test-harness mitigation only — it does not change the server-mode bootstrap recovery itself, does not cover the REST rig-creation path, and does not touch release/v1.3.0.

### DIFF
--- a/release-gates/ga-cdlu2j-clean-install-stdout-isolation-gate.md
+++ b/release-gates/ga-cdlu2j-clean-install-stdout-isolation-gate.md
@@ -1,0 +1,47 @@
+# Release gate: clean-install stdout isolation
+
+- Deploy bead: `ga-cdlu2j`
+- Reviewed source: `3b9824e5ae5129d5e3d4b60b885ab80e4a534278`
+- Base: `origin/main@e5ed16b566302e5cd9bdbb7d75ff47bd52a8cbfc`
+- Review: `ga-hash7u` (PASS, round 2)
+- Gate verdict: **PASS**
+- Deploy mode: remote (`gastownhall/gascity`, push through the configured GitHub remote)
+
+The repository does not contain `docs/PROJECT_MANIFEST.md` at either the base
+or reviewed commit, so this checklist applies the seven release criteria in the
+active `mol-deployer-gate` formula and deployer prompt.
+
+## Criteria
+
+| # | Criterion | Result | Evidence |
+|---|---|---|---|
+| 1 | Review PASS present | **PASS** | Review bead `ga-hash7u` records `verdict: pass` for the resolved reviewed commit. |
+| 2 | Acceptance criteria met | **PASS** | The integration harness now separates stdout from stderr for value-bearing commands, retains stderr in command errors, narrowly retries only the known Dolt dirty-table migration race, and keeps retries bounded at three attempts. `TestCleanInstallTutorialPath` passes end to end, the two stdout contract tests pass, both retry-policy tests pass, and `TestRepositoryLedgerMatchesCensusAndDocumentation` passes with no resource-census baseline increase. |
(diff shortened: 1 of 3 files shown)


quad341:
@quad341 — this is a sharp, surgical fix. Two things stand out. Predicating the retry on one exact beads#4566 string *and* writing a test that asserts it does **not** match the ga-rsktma stdout-contract string is the right instinct: that's the line between a bounded retry and a retry-until-green that quietly eats the regression the suite was built to catch. And folding both paths through `buildCommand` so the resource census still sees one subprocess site, plus routing the delay through `backoff/v4` instead of a local `time.Sleep`, shows you read what the census guards rather than fighting it.

I traced the load-bearing parts and they hold: `gcDolt`/`runGCDoltWithEnv` still use `CombinedOutput`, so the 4566 signature really does reach the predicate, and the tutorial test's `beadOut+err.Error()` check keeps working now that stderr rides in the error.

The red check isn't yours. `TestSkipRevisionSnapshot_StillDetectsPackContentChange` in `internal/config` passes at your base tip and on current main; the fixture's before/after strings are both 19 bytes, so `PackContentHashRecursive`'s size+mtime cache fingerprint can collide. I'll re-run it, file that flake separately, and merge with your commits intact.


---


quad341:
@quad341 — the care here is in the negative assertions. Predicating the retry on one exact beads#4566 string, then writing a test that asserts it does *not* match the ga-rsktma stdout-contract string, is the difference between a bounded retry and a retry-until-green that quietly eats the regression this suite exists to catch. Same instinct in folding both helpers through `buildCommand` so the subprocess census still sees one site, and routing the delay through `backoff/v4` the way `path_helpers_test.go` already does.

I traced the parts that had to hold and they do: `gcDolt` and `runGCDoltWithEnv` still use `CombinedOutput`, so the 4566 signature genuinely reaches your predicate.

The red check isn't yours. `Integration / packages-core-2-of-4` fails only on `internal/config.TestSkipRevisionSnapshot_StillDetectsPackContentChange` — unreachable from an integration-tagged file. Both fixture bodies are 19 bytes and the pack hash cache fingerprints on size+mtime, so a same-size rewrite in one tick returns the cached hash. I'll re-run it, file that separately as a real reconciler hazard, and merge with your commits intact.


---

PR #5375, opened 2026-08-18
fix(session): evict wedged reset-pending sessions and bound crash-event noise (#5355)

### LINKED ISSUE #5355: Session wedged in reset-pending after config change: stale runtime never evicted, session.crashed fires ~2/min for hours with full pane dumps in payload
## Summary

A session wedged in `asleep / reset-pending` after its configured command changed (provider flip: codex → claude) is never repaired: the previous runtime had exited cleanly but its tmux session was still alive, so the pending continuation reset never completed. Meanwhile the crash detector fired `session.crashed` roughly every 30 seconds for ~5.3 hours — 299 events, each embedding the **full tmux pane capture** (~6.6 KB) in the `message` field — with **zero repair or restart attempts**. Combined with other subjects flooding the same way, `.gc/events.jsonl` grew to ~50 MB.

## Environment

- gc `dev` build based on upstream v1.4.0 (fork `swedeinasia-flow/gascity`)
- macOS (darwin), tmux socket `-L gas-city`
- City: multi-rig city with `on_demand` witness sessions

## Timeline (2026-08-17)

1. A provider-profile flip rewrote session `gc-0ft8rj` (`flow-concourse/witness`) config from a codex command to a claude command.
2. The old codex runtime had already exited cleanly the day before, but still occupied tmux session `flow-concourse--witness` (pane alive, process idle).
3. At 09:14 the reconciler raised a continuation reset: session state became `asleep / reset-pending` (`continuation_reset_pending: true`).
4. The reset never completed — reset-pending does not evict a stale-but-alive runtime, so the config/runtime mismatch persisted.
5. From 09:14:48 to 14:35:07 the supervisor emitted 299 `session.crashed` events (~2/min), seq 1318295–1329031, each with the full pane dump in `message`. No repair, relaunch, or eviction was ever attempted. Nothing was actually crashing.

Other rig witnesses on the same city reset fine at 09:14 because their runtimes happened to be replaceable at that moment — the wedge only occurs when a stale runtime still holds the tmux session.

## Expected behavior

Three independent problems here, any of which would have contained the incident:

1. **reset-pending should evict/replace a stale runtime.** If a continuation reset is pending and the occupying runtime no longer matches the configured command (or its process has exited), the supervisor should kill/replace the tmux session rather than waiting forever.
2. **The crash detector should back off and/or deduplicate.** 299 identical detections of the same condition on the same subject over 5 hours, with zero state change between them, should collapse into a few events with exponential backoff — and ideally escalate to a repair action rather than pure detection.
3. **`session.crashed` payloads should be truncated.** Embedding the full pane capture in every event's `message` field turns a detect-loop into an event-log bloat incident (~6.6 KB × 299 for this subject alone; the file reached ~50 MB across subjects). A capped excerpt (say first/last N lines) plus a pane-capture reference would serve diagnosis without the bloat.

## Actual behavior

`asleep / reset-pending` forever; `session.crashed` flood at ~2/min; events.jsonl bloat; no self-repair.

## Manual remedy that worked

```
gc session kill <id>    # evicts the stale runtime — this alone stops the flood
gc session reset <id>   # clears the pending continuation reset (a bare `wake` does NOT clear it)
gc session wake <id>    # session comes back `active`, pane runs the new provider binary
```

Note the middle step: `wake` alone after `kill` does not clear `continuation_reset_pending`, which is itself worth a look — an operator's natural first move (`kill` + `wake`) leaves the session wedged again.

## Evidence

- Events seq 1318295–1329031 in the city's `.gc/events.jsonl` (299 × `session.crashed`, subject `flow-concourse/witness`, actor `gc`, message length 6613 bytes each, first `2026-08-17T09:14:48.804699+02:00`, last `2026-08-17T14:35:07.063841+02:00`). Happy to attach a redacted sample on request.

## Detection signature (for anyone else hitting this)

`gc session list` shows `asleep  reset-pending` while `session.crashed` events repeat ~2/min for the same subject with pane-dump payloads. `on_demand` witnesses resting `asleep` with an empty inbox is normal — the loop, not the sleep, is the bug.

rjgeng:
Fixes #5355.

A session wedged in `reset-pending` after a config change never got evicted — the reconciler just kept re-recording the stall event, so the stale runtime stuck around under the old config forever. Meanwhile `session.crashed` fired roughly every 30s reconciler tick unconditionally (no cross-tick dedup), each carrying a full ~120-line raw pane-capture payload, producing hours of noisy events at ~2/min.

**Three bundled sub-fixes, matching the issue's own "Expected behavior" framing:**
1. A genuinely wedged reset-pending session now gets evicted via `workerKillSessionTargetWithConfig` instead of just re-recording the stall — the normal spawn path then recreates it under current config.
2. Zombie-crash detection now dedupes across reconciler ticks (new `zombieCrashes` map on `drainTracker`, mirroring the existing `resetStalls` pattern) instead of firing `SessionCrashed` unconditionally every ~30s.
3. The crash event's pane-capture payload is now bounded (24 lines, first/last 12) instead of the full raw capture.

**Deliberately left `session_wake.go` untouched** — the issue's minor adjacent note that `wake` alone doesn't clear `continuation_reset_pending` after `kill` is a separate, smaller finding, not bundled into this fix.

**Testing:** `go build ./...` clean, `go vet ./cmd/gc/...` clean, `gofmt -l` empty, full `go test ./cmd/gc/...` green (855s, no shortcuts) — 3 new regression tests: eviction-not-just-recording, cross-tick dedup, payload truncation.

---


quad341 (approved):
@rjgeng — thank you for this. You didn't just mute the event flood, you went after what was causing it: the eviction in `recordResetStallIfDue` is the actual fix for #5355, with the dedup and the payload bound layered on as noise control. Mapping the three sub-fixes onto the issue's own "Expected behavior" list made this genuinely easy to follow, and deliberately leaving `session_wake.go` alone — and *saying so* — was the right call.

I traced your load-bearing claim and it holds: `PreWakePatch` and `CommitStartedPatch` both clear `continuation_reset_pending`, so the respawn really does reset the dedup mark and close the loop. I reproduced your results locally too — build, vet, gofmt, and the full `cmd/gc` suite all green.

Two things I'll take on my side and merge with your commits intact. The eviction sits behind the `markResetStall` gate, so a transient kill failure would never retry — I'll split the dedup off the kill and add a failure-then-retry test. And the eviction fires before the lazy `peek`, so on real tmux the post-kill `capture-pane` fails and we lose the rate-limit classification for that tick. `runtime.Fake.Peek` returns canned output whether or not the session exists, so that one was structurally invisible to your tests — nothing you could have caught.


---

PR #5376, opened 2026-08-18
fix(gc): print short usage pointer instead of full command help on flag/arg errors (#5359)

### LINKED ISSUE #5359: bug(cli): flag parse errors print the full command usage after the error
### Before you continue

### Gas City version

`1.4.1` (`0a197418fe15123d70372bf58e37043ccb3c3227`, built 2026-08-06). Re-checked against current `main` before filing; the same path remains in `cmd/gc/main.go`.

### Environment

macOS 26.5, zsh, tmux runtime provider.

### Reproduction

Run a harmless command with an unknown flag:

```shell
gc mail list --definitely-invalid >out 2>&1
echo $?
wc -l out
sed -n '1,5p' out
```

Observed locally:

```text
1
32

gc: unknown flag: --definitely-invalid

Usage:
  gc mail [flags]
```

The behavior is generic across subcommands, not specific to `gc mail`.

### Expected behavior

A flag-parse error should print the specific error and a short help pointer, matching the CLI's terse validation-error behavior. For example:

```text
gc: unknown flag: --definitely-invalid
Run "gc mail --help" for usage.
```

### Actual behavior

The useful error is line 1, followed by the command's full usage text. In captured agent/tool output, the operative line is easily pushed out of view by 30+ lines of unrelated help.

```shell
$ gc mail list --definitely-invalid >out 2>&1; echo "exit=$? lines=$(wc -l < out)"; head -5 out
exit=1 lines=32
gc: unknown flag: --definitely-invalid

Usage:
  gc mail [flags]
  gc mail [command]
```

### Additional context

Current `main` already sets `SilenceErrors: true` and `SilenceUsage: true` on the root command. Cobra is not producing the extra text: Gas City prints it explicitly.

- `root.SetFlagErrorFunc` calls `printCommandUsageError`.
- `printCommandUsageError` prints the specific error, then calls `printCommandUsage`.
- `printCommandUsage` emits `cmd.UsageString()` in full.
- Flag-group validation and the root/pack unknown-command paths funnel through the same usage printer.

A small fix would keep the explicit error at each caller and replace the full `UsageString()` emission on error paths with a one-line command-specific help pointer. Tests should cover an unknown flag and a flag-group validation error, while preserving normal `--help` output.

rjgeng:
Fixes #5359.

CLI flag/arg parse errors dumped the full command usage string (`cmd.UsageString()`) after the error, burying the actual error message under a wall of help text.

**Fix:** `printCommandUsage` (`cmd/gc/main.go`) now prints a short `Run \"<command path> --help\" for usage.` pointer instead — matches the issue's own proposed output exactly. Single choke point; all 6 existing error call sites already funnel through it. `--help` output itself is untouched (verified with a new test).

**Testing:** `go build ./...` clean, full `go test ./cmd/gc/...` green, `gofmt -l` empty.

---

### DIFF
--- a/cmd/gc/cmd_beads_test.go
+++ b/cmd/gc/cmd_beads_test.go
@@ -701,6 +701,85 @@ func TestRun_BeadsListRejectsUnknownFlag(t *testing.T) {
 	}
 }
 
+// TestRun_UsageErrorsPrintShortPointerNotFullUsage locks issue #5359: every
+// flag/arg/command error path funnels through printCommandUsage, which used
+// to dump cmd.UsageString() in full (30+ lines, burying the actual error).
+// It now prints a single `Run "<path> --help" for usage.` pointer instead.
+func TestRun_UsageErrorsPrintShortPointerNotFullUsage(t *testing.T) {
+	tests := []struct {
+		name        string
+		args        []string
+		wantStderr  []string
+		notWantSubs []string
+	}{
+		{
+			name:       "unknown flag",
+			args:       []string{"mail", "list", "--definitely-invalid"},
+			wantStderr: []string{"gc: unknown flag: --definitely-invalid", `Run "gc mail --help" for usage.`},
+		},
+		{
+			name:       "unknown command",
+			args:       []string{"totally-bogus-command"},
+			wantStderr: []string{`gc: unknown command "totally-bogus-command"`, `Run "gc --help" for usage.`},
+		},
+		{
+			name:       "flag group validation error",
+			args:       []string{"mail", "send", "--to", "mayor", "--all", "hi"},
+			wantStderr: []string{"if any flags in the group", `Run "gc mail send --help" for usage.`},
+		},
+		{
+			name:       "--city empty value",
+			args:       []string{"--city="},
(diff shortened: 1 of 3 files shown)


quad341 (approved):
@rjgeng — thank you for finding the single choke point instead of patching six call sites. `printCommandUsage` was exactly the right lever, and routing through `cmd.CommandPath()` means nested pack commands get an accurate pointer (`Run "gc backstage repo --help"`) for free rather than a blunt `gc --help`.

The part that made this an easy merge was `TestRun_HelpStillPrintsFullUsage`. Plenty of "make errors quieter" patches quietly gut `--help` on the way past, and you pinned that this one doesn't. Same for the negative assertion that `"Available Commands:"` and `"Flags:"` never leak onto stderr — that turns a fix into something that stays fixed. You also caught all three test files referencing the old output; I grepped the tree for stragglers in goldens, docs, and the acceptance suite and found none.

Merging as-is. `go build ./cmd/...`, `gofmt`, and the affected tests are green on my side too, and the `cmd/gc` shards ran this code for real in CI.


---

PR #5377, opened 2026-08-18
fix bootstrap cleanup of linked worktrees

quad341:
## What this changes

New Gas City bootstrap packs now remove linked worker worktrees from the owning repository even after the cleanup step has moved outside the worktree. This prevents a deleted worktree directory from remaining registered as a stale, prunable worktree.

Both `mol-polecat-commit` and `mol-scoped-work` resolve the repository before removal, anchor Git operations with `git -C`, and prune the registration if Git removal fails and the directory must be deleted directly.

## Review notes

- The change is limited to the two embedded bootstrap formula definitions and their colocated regression tests.
- Repository discovery uses `--git-common-dir`, because `--show-toplevel` inside a linked worktree resolves to the worktree that is about to disappear.
- The direct `rm -rf` fallback runs only after `git worktree remove --force` fails; `git worktree prune` then repairs the repository registration.
- No command-line, API, or configuration schema changes are introduced.

## Test plan

- [x] Run both bootstrap cleanup regression tests by name.
- [x] Run the complete 40-job local CI-equivalent matrix with reduced fan-out; feature-owned coverage passes.
- [x] Run CI policy, affected lint, changed formatting, and `go vet ./...`.
- [x] Release gate: [`release-gates/ga-xh3qsn-bootstrap-worktree-cleanup-gate.md`](release-gates/ga-xh3qsn-bootstrap-worktree-cleanup-gate.md)


quad341:
Thanks @quad341 — `--git-common-dir` is the good insight here. Reaching for `--show-toplevel` inside a linked worktree is the obvious move and it's exactly wrong, because the path you resolve is the one about to stop existing. I checked your version in a scratch repo: anchoring on the common dir and driving removal with `git -C "$REPO"` unregisters cleanly even from a cwd entirely outside the repo. Pinning both formulas with named regression tests, in a pack where the behavior lives in prompt text rather than Go, is what keeps this from quietly coming back.

One thing I'm adding on my side before merge: a guard on the new `rm -rf` fallback. Removing a *main* working tree exits 128, and `$REPO` lives inside it — so unguarded, that fallback would delete the whole checkout. One-line `.git`-is-a-file check in each formula, plus handling the empty-`REPO` case in `mol-scoped-work`. I'll pin both with assertions.

The red CI isn't yours: `TestSkipRevisionSnapshot_StillDetectsPackContentChange` caches on `(size, mtime)` and its two fixtures are both 19 bytes. Separate bead. Merging with your commits and credit intact.


---

PR #5384, opened 2026-08-18
fix(beads): resolve issue-shaped wisp dependencies

### LINKED ISSUE #5223: ReadyGraphOnly's dependency gate ignores issue-targeted wisp dependencies, excluding the wisp permanently
Found while verifying #3853. Not a blocker for it: the surface has no non-test
caller on `main` today, so this is latent until the consumer lands.

`doltliteWispDepGate()` in `internal/beads/doltlite_read_store_graph_only_ready.go`
resolves a blocker only through `depends_on_wisp_id`:

```go
wispTarget := "NULLIF(d.depends_on_wisp_id, '')"
blockerJoin := "LEFT JOIN " + doltliteWispTables.issues + " blocker ON blocker.id = " + wispTarget
// ... AND COALESCE(blocker.status, '') != 'closed'
```

`wisp_dependencies` also carries `depends_on_issue_id`, `depends_on_id`, and
`depends_on_external`. When a row uses one of those instead, `wispTarget` is
NULL, the LEFT JOIN produces no blocker row, and `COALESCE(blocker.status, '')`
is `''`, which is not `'closed'`. The `NOT EXISTS` is therefore false and the
wisp is excluded.

The exclusion is permanent. Closing the target issue does not clear it, because
the join never resolves the target in the first place. A step depending on a
regular bead would never appear ready through this path.

Direction of the bug is worth stating: it fails **closed**, under-reporting
readiness. That is the safe direction, and the opposite of the over-reporting
defect the same method was fixed for during review. Nothing dispatches work it
should not; work simply never becomes ready.

### The established shape to copy

`doltliteReadyIssueWhere` in `internal/beads/doltlite_read_store.go` already
models both target kinds and picks between them:

```go
blockerStatus = "CASE WHEN " + wispTarget + " IS NOT NULL " +
    "THEN COALESCE(blocker_wisp.status, '') " +
    "ELSE COALESCE(blocker_issue.status, '') END"
```

The graph-only gate wants the same two-arm resolution. It deliberately omits
the type-exclusion filter, which is correct and documented, but the blocker
resolution should not have been narrowed along with it.

### Reachability

`grep` for non-test callers of `ReadyGraphOnly` or `GraphOnlyReadyFor` at
`main` returns nothing. The consumer is #3835
(`build_desired_state_graph_ready.go`), so this becomes live behavior when that
merges. Fixing it before then costs nothing; after then it is a dispatch stall
that looks like missing work rather than an error.

### Test worth adding

A wisp whose `depends_on_issue_id` points at a **closed** issue should be
returned. That case fails today and passes with the two-arm resolution, so it
pins the fix rather than restating it. The existing
`TestDoltliteReadStoreReadyGraphOnlyExcludesBlockedWisps` covers the wisp-target
arm on both sides and can stay as is.

quad341:
## Summary

- Resolve issue-shaped `wisp_dependencies` through the durable issue table in `doltliteWispDepGate`.
- Preserve the existing wisp-target arm and fail-closed behavior for unresolved dependencies.
- Add a RED/GREEN pinning test proving an open issue blocks readiness and a closed issue releases it.

Closes #5223.

## Testing

- [x] Focused native DoltLite regression: 1 PASS, 0 FAIL, 0 SKIP.
- [x] `make test-native-doltlite-beads` passed on the reviewed SHA.
- [x] `go vet ./...` passed.
- [ ] `make test-local-full-parallel` remains red on the reviewed SHA: 26 PASS / 14 FAIL / 0 SKIP jobs. Criterion 3 is explicitly waived for this gate by mayor ruling `gm-wisp-hmtkss`; the failures remain recorded as failures, are attributed on `ga-h3dioy`, and are linked to `ga-cqq3hs`.

## Release gate

- Reviewed commit: `2dbb5c79a6bd253a50438dd7601769a86bb4f601`
- Review bead: `ga-x97emt` — PASS
- Deploy bead: `ga-h3dioy`
- Gate record: `release-gates/ga-h3dioy-doltlite-wisp-issue-dependency-gate.md`

## Checklist

- [x] Linked issue #5223.
- [x] Added regression coverage for the behavior change.
- [x] User-facing docs are not affected.
- [x] No breaking change or migration.


quad341:
@quad341 — nice catch on this one. The bug was invisible in the happy path: a `wisp_dependencies` row carrying only `depends_on_issue_id` left `wispTarget` NULL, so the blocker join never matched, status came back `''`, and `'' != 'closed'` kept the wisp excluded *forever* — even after its blocker closed. A permanently stalled wisp is silent, not loud, which is the worst way for a dispatch bug to fail.

Mirroring `doltliteReadyIssueWhere`'s two-arm `CASE` exactly, rather than inventing a new predicate, is what made this reviewable in minutes — I diffed the two and they line up. And pinning *both* directions (open blocker excludes, closed blocker releases) is what makes it a real regression test instead of one that would pass against a gate that blocks everything.

Merging as-is. The red `Integration / packages-core-2-of-4` isn't yours: it's `TestSkipRevisionSnapshot_StillDetectsPackContentChange` in `internal/config`, which doesn't exist in your branch at all — and `main`'s own CI is red on it, so a rebase wouldn't help. That's ours, tracked at `ga-cqq3hs`.

Worth a follow-up bead, not a change here: a blocker recorded only in the generic `depends_on_id` column still routes to the issues arm. Same limitation the sibling predicate has, so it's pre-existing — but your test file is now the natural home for it.


---

PR #5386, opened 2026-08-18
feat: cross-city gc mail — remote mail send|reply|inbox via --context + [supervisor.hardened] grant-gated second listener

aphexcx:
## Summary

Two commits that together make **city ↔ city mail with a real sender identity** possible with `gc` alone (no ssh, no `--from human`), for a supervisor that must stay the trusted loopback control plane *and* accept grant-signed remote writes.

1. **`feat(cli): route gc mail send|reply|inbox to a remote city via --context`**
   `gc --context <peer> mail send|reply|inbox` were refused ("remote support is being enabled incrementally") although the control plane already exposes `POST /v0/city/{c}/mail` (with `from`), `POST /mail/{id}/reply` and the inbox GET. This routes them through the existing remote transport (`resolveWriteTarget` / `resolveReadTarget`, same pattern as `sling_remote.go`), adds `api.Client.SendMail/ReplyMail`, and defaults the sender to `<local city>/<identity>` (alias > agent > session id > human; `--from` overrides) so the far side knows which city to answer. `--all` / `--notify` are refused remotely; the target is echoed to stderr like sling / rig add; help + `docs/reference/cli.md` regenerated (mail hunks only).

2. **`feat(supervisor): [supervisor.hardened] — a second, grant-gated API listener`**
   `[supervisor] write_auth_verify_key` gates *every* per-city mutation on the single listener — including first-party callers that mint no grant (dashboard SPA, workspace-service adapters posting session messages, the local gc client). So a supervisor that is also a local control plane could not require grants on its remote path without 401-ing itself, and tailnet/VPN-exposed supervisors ran with an unauthenticated write plane. `[supervisor.hardened] {bind,port,write_auth_verify_key}` adds an optional second listener over the same mux, wrapped in `writeAuthMiddleware` with its own key **plus a front door** (`hardenedListenerGuard`) that refuses supervisor-scope mutations (notably `POST /v0/city` — city creation with a caller-chosen `start_command`, which write-auth exempts by design), every `/svc/*` request (a GET can be a WebSocket upgrade the proxy then tunnels), and any protocol upgrade. Fail-closed & all-or-nothing boot: missing key, shared kid with the primary key list (each verifier has its own single-use replay guard), hardened port = primary, read-only primary, or hardened bind failure → no listener comes up. `GC_CITY_WRITE_PUBKEY` is not consulted for the hardened key. Runbook: `[supervisor.hardened]` section + §5a cross-city mail.

## Deployment shape this enables

```
city A supervisor ── 127.0.0.1:8372 (first-party, unchanged) 
                  └─ 127.0.0.1:8447 [supervisor.hardened] verify kid "b" ◄── tailscale serve :8443 (TLS, tailnet-only) ◄── city B: gc --context a mail send mayor
```
Each city keeps its own ed25519 key; only public keys cross the wire (`gc context add … --grant-command 'gc-write-mint --kid b --key … --city a'`).

## Testing

- `go test ./internal/supervisor ./internal/api -run 'Hardened|WriteAuth|Mail'`, `go test ./cmd/gc -run 'RemoteMail|MailSendRemote|MailReplyRemote|MailInboxRemote|CmdMail_Context|TestRunSupervisor'` — all green on this branch (upstream/main + these two commits).
- Live smoke (patched binary): isolated supervisor with `[supervisor.hardened]` → primary POST not gated, hardened POST without grant → `401 missing X-GC-City-Write grant`, hardened GET 200; `gc --context … mail send` with a `gc-write-mint` grant reaches the handler; against a real supervisor via a loopback context: send → `Sent message … (as <city>/<id>)`, inbox shows the qualified sender, reply → addressed back to it.
- Independent security review (2 rounds, deep) of the listener commit: round-1 findings (grant-less `POST /v0/city` on the hardened port, `/svc` websocket bypass, cross-listener jti replay via shared kid, non-atomic bind) are all fixed and confirmed closed in round 2.

## Latent upstream gap (not fixed here — flagging)

The same reviewer noted that upstream's *documented* hardened posture (primary bind `0.0.0.0` + `write_auth_verify_key`, `docs/runbooks/remote-hardened-city.md` §3) has the identical holes the new guard closes on the hardened listener: `POST /v0/city` (city creation) is reachable with only the CSRF header, and a `/svc` GET upgrade tunnels through the write gate. `hardenedListenerGuard` could be applied to a non-loopback grant-gated primary as well; I left that out to keep this PR's contract change scoped — happy to follow up.

### DIFF
--- a/cmd/gc/cmd_mail.go
+++ b/cmd/gc/cmd_mail.go
@@ -1450,7 +1450,16 @@ a non-running recipient. Unread mail alone does not request a wake.
 Use --from to override the sender identity.
 Use --to as an alternative to the positional <to> argument.
 Use -s/--subject for the summary line and -m/--message for the body text.
-Use --all to broadcast to all live sessions (excluding sender and "human").`,
+Use --all to broadcast to all live sessions (excluding sender and "human").
+
+With --context/--city-url the message is sent to a REMOTE city over the
+control plane (a hardened city requires the context's grant_command). The
+sender then defaults to "<local city>/<identity>" (e.g. citadel/mayor) so the
+far side knows which city to answer with 'gc --context <city> mail send';
+--from overrides it. The remote city stores an unknown sender literally, but
+if it has a session whose alias equals that string (a rig named after the
+sending city) the message binds to that session — do not name a rig after a
+city that mails you. --all and --notify are refused for a remote city.`,
 		Example: `  gc mail send mayor "Build is green"
   gc mail send mayor -s "Build is green"
   gc mail send myrig/witness -s "Need investigation" -m "Attach logs from the last failed run"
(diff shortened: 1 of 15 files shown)


A3Ackerman:
Data point from two downstream cities: the first commit here (`cdcd0611c`, the `--context` mail routing) has been carried on both cities' fork lineage since 2026-08-18 with no fix-up needed since; the hardened-listener commit we have not exercised. #5797 (draft RFC) is complementary rather than competing — it is addressing inside one shared `messaging` store and assumes this transport for the city-to-city hop — so from our side this is the one to land first.

PR #5387, opened 2026-08-18
perf(cmd/gc): stop re-reading every cached pack file on each config load

quad341:
## What this changes

`EnsureBuiltinRuntimeAssets`' ready fast path re-validated the synthetic pack
caches with the full content-reading validator on **every config load** — and a
config load happens on every `gc` command. Each pass `os.ReadFile`'d and
compared every cached pack file against the copy embedded in the binary, so the
cost was O(bundled pack files) per command.

The ready path must still notice a cached file corrupted in place after
readiness, so this keeps the full validator and instead memoizes its **positive**
verdicts across passes, gated by a cheap per-file size+mtime stat fingerprint
that reads no file contents. Any ordinary write changes size or mtime, which
invalidates the fingerprint and runs the full validator again — the same
invalidation contract `packContentHashCache` already uses.

## Why it is worth fixing rather than sweeping

Go stores the complete log of a test's file operations alongside its cached
result — that log *is* the cache key. A redundant read is therefore billed
twice: once as a syscall, and again as a line of cache log. 190 such entries
totalling 49 GB were on disk when this was filed (ga-jwa1c5).

## Measurements

`-test.testlogfile` on `TestE1LazyMissingTreeMatchesEagerFlagOwnership` (a 0.8s
test), baseline `origin/main` e5ed16b566 vs this branch:

| metric | main | this branch | change |
| --- | ---: | ---: | ---: |
| pack-cache file operations | 70,640 | 39,135 | −44.6% |
| — of which are file-content opens | 29,144 | 2,276 | −92% |
| test-cache log size | 15.9 MB | 9.0 MB | −43% |

Per individual cached file (`examples/bd/dolt/orders/dolt-health.toml`):
**199 content opens → 7**.

## Known limitation — the same contract, not a new one

The stat fingerprint cannot see an edit that preserves *both* size and mtime.
That is the documented blind spot `packContentHashCache` already accepts, and
the new test pins it explicitly rather than leaving it to a comment. #5367 is
separately strengthening that sibling contract with ctime; if it lands, this
fingerprint should follow it.

## What this does NOT fix

The warm path still walks the cache tree once per config load to compute the
fingerprint. That walk is stat-only, but it is still O(files): 24,038 stats and
12,821 directory reads remain in the same test. Making the ready check O(1) per
pass is filed separately.

## Tests

- **New** `TestWarmSyntheticCacheVerifierReusesPositiveVerdictsAcrossPasses`
  pins both halves of the fast path with deliberately opposed assertions:
  a size+mtime-preserving corruption must be served the stale positive (proving
  the memo was consulted rather than the tree re-read), and an ordinary
  corruption must still be caught (proving the blind spot does not swallow
  self-healing).
- Both halves verified **non-vacuous by mutation**: disabling the memo fails
  half one; making the memo ignore its fingerprint fails half two.
- Existing corruption/self-heal tests unchanged and passing
  (`TestEnsureBuiltinRuntimeAssetsRehydratesCorruptedCache`,
  `TestReadinessHelpersValidateTheSharedCacheOncePerPass`).
- `go test ./cmd/gc/ -count=1` green (356s), `go vet ./...` clean, lint 0 issues.


quad341:
@quad341 — this is a really well-argued PR. Memoizing only the *positive* verdicts and gating them on a stat fingerprint is exactly the right shape: the ready path keeps the full validator, so the self-heal guarantee survives instead of being traded away for the win. And tracing the cost to the Go test-cache log — 190 entries, 49 GB, because a redundant read is billed twice — is a genuinely good diagnosis. 199 content opens down to 7 per cached file is a real number.

The test is the part I'd point other people at. Writing the blind spot as an assertion, with the size-preserving corruption proving the memo was consulted and the length-changing one proving self-healing still fires, means neither half can rot into a tautology.

One thing I want to tighten before it lands, and I'll do it on my side. `ValidateSyntheticRepo` also rejects wrong file permissions and unexpected directories, but the fingerprint hashes only files' size and mtime — so a `chmod` (which moves ctime, not mtime) or a new empty directory could ride a stale positive verdict. Concretely, the bundled `gc-beads-bd.sh` is materialized 0o755 and exec'd through the shim. I'm adding mode and directory entries to the hash — still stat-only, so your perf win is untouched — plus a regression case for each, and re-running `./cmd/gc`.

Worth noting the `packContentHashCache` analogy is slightly generous to itself: same fingerprint, but that one gates a content hash that never looks at modes, so its blind spot is narrower than ours by construction. I'll say so in the comment.

Merging with those on top, credited to you. If #5367 lands and you want to take the ctime follow-up here, it's yours.


---


quad341:
Picked this up from a stale-PR audit (open 14 days, last activity 11). Took the three points from the review above rather than leaving them pending — all three are in `3653819898`.

**1. Permission bits could ride a stale positive verdict.** The fingerprint now hashes each file's `Mode().Perm()`. Regression test `TestWarmSyntheticCacheVerifierNoticesAModeChange` chmods a materialized `0o755` file to `0o644` and asserts the verdict flips. It finds its target by walking for a `0o755` file rather than by name, so renaming a bundled script cannot quietly turn it into a no-op, and it asserts the chmod moved *neither* size nor mtime before asserting the flip — otherwise it could pass on the old fingerprint for the wrong reason.

**2. Unexpected directories were invisible.** Directory paths are now hashed too. Entries are tagged `f`/`d` so a file and a directory sharing a path cannot hash alike. A directory's own mtime is deliberately *not* hashed — it moves whenever a child is written, which that child's entry already covers, so hashing it would invalidate the memo spuriously. `TestWarmSyntheticCacheVerifierNoticesAnUnexpectedDirectory` adds an **empty** directory on purpose; one holding a file would pass for the wrong reason.

Both tests were confirmed RED against the pre-change verifier and GREEN after, so neither is a tautology.

The perf win is untouched, and slightly more strongly than "still stat-only": this adds **no new syscalls at all**. Directories are hashed by path alone and never stat, and the mode comes from the `DirEntry.Info` already fetched per file.

**3. The `packContentHashCache` analogy.** Agreed it was generous — and #5367 has since landed (`6c1420589e`), which makes it asymmetric in *both* directions, so I restated it rather than just softening it:

- That cache gates a **content hash**, which depends on paths and bytes alone. Its fingerprint covers its verdict where the same fingerprint applied here left the two gaps above, because a `ValidateSyntheticRepo` verdict also depends on modes and the directory set.
- But it now hashes **ctime**, so the mtime-preserving edit this memo still cannot see is a gap the sibling no longer has.

That second half is the ctime follow-up you offered. I did **not** fold it in, because `statCtimeNanos` is unexported in `internal/config` (`fingerprint_ctime_unix.go` + `_windows.go`), so closing it properly means *sharing* that build-tagged pair rather than duplicating it into `cmd/gc` — a placement decision, not a mechanical edit. It is documented in-code and filed as **ga-77cgcb** with a full spec, including inverting the blind-spot assertion in `TestWarmSyntheticCacheVerifierReusesPositiveVerdictsAcrossPasses`, whose comment already says it should be inverted when ctime lands.

### One thing worth knowing before merging

I rebased onto current `main` (force-with-lease). The branch was 231 commits / 14 days behind, and **it could not pass the local push gate on its old base**, for two reasons that had nothing to do with this change:

- `TestCityRuntimeForceShutdownTearsDownAfterLateAsyncSweep` — fixed on main by #5331 (`7e92a87f18`), which this base predated. Its own message notes it failed "roughly a third of isolated runs, always on its last assertion", which is exactly what I hit.
- `TestCatalogMatchesProductionWiringAndDocumentation` — the `ga-80po0c.3` runtime-provider waivers **expired 2026-08-26**, so this was a wall-clock time bomb that would fail any branch based before that date. Fixed on main by #5652 (`f20fac505a`), "make a lapsed waiver fail its owner, not the fleet".

Both pass on the rebased base, and the push gate is green. Worth flagging in case other long-lived branches are sitting on the same two.

Ready for another pass.

quad341:
@quad341 — merging this as-is. The diagnosis is the part worth calling out: tracing the cost past the syscall to the Go test-cache log, where a redundant read is billed twice — once as a syscall, once as a line of the cache key — is what turned this from "sweep the 49 GB" into a fix worth making.

The test design is the other part. Writing the blind spot as an *assertion* rather than a comment — a size-preserving corruption proving the memo was consulted, a length-changing one proving self-healing still fires — means neither half can quietly rot into a tautology. Deriving the fingerprint's contents from what `ValidateSyntheticRepo` actually rejects (hence mode, hence empty directories) rather than from what the sibling cache happens to hash is exactly right, and `firstExecutableCachedFile` walking for `0o755` instead of matching a filename means a rename can't silently no-op that regression.

Deferring the ctime gap was the right call — sharing `statCtimeNanos` out of `internal/config` is a placement decision, not a mechanical edit — and ga-77cgcb holds it. Two of the new failure messages still describe the pre-fix fingerprint as ignoring mode and directories; cosmetic, grab it next time you're in the file. Thanks for this one.


---

quad341:
@quad341 — the thing that makes this land is that you built the fingerprint out of what `ValidateSyntheticRepo` actually rejects, not out of what the sibling cache happens to hash. That's why mode and empty directories are in there, and I checked both: `validatePackFiles` fails on a permission mismatch, `validateSyntheticRepoFileSet` fails on a stray directory, so either gap would have let a broken cache ride a stale positive. Hashing them costs nothing extra either — the permission bits come from the `d.Info()` call the size/mtime hash already makes.

The tests are the other half of it. Guarding the premise in the chmod case — asserting size and mtime genuinely didn't move before asserting the verdict flips — is what stops that test passing for the wrong reason, and insisting the stray directory be empty does the same job. Tracing the cost past the syscall into the Go test-cache log to explain the 49 GB is a nice piece of diagnosis.

Merging with credit to you. I'm taking a small wording pass on my side first: the comment on the unexpected-directory test still says the walk skips directories, which is what it did before your change. I'll also note that the memo is process-global, so inside a long-running supervisor the mtime-preserving blind spot lasts the process lifetime rather than one command — worth having written down when you pick up the ctime follow-up alongside #5367.


---

quad341:
Re-reviewed `01c64b191`..`14819b9e8` (118 new commits since the last review): no new concerns — the prior review stands.

PR #5391, opened 2026-08-18
fix(scripts): don't let an unrelated in-progress bead block a different deploy-gate push

quad341:
## Summary

- push-ownership-guard's `deploy/*-gate` resolution path deliberately discards the branch-embedded bead id (it names the *gated* bead, which is routinely already closed by push time) and falls back to `bd list --assignee="$GC_AGENT" --status=in_progress --json`. It previously took `.[0]` of that result unconditionally.
- A session can legitimately hold more than one in-progress bead at once (confirmed live: this exact session had 3 simultaneous in-progress beads while working this fix). When it does, the fallback could resolve to a completely unrelated bead and block a deploy-gate push that had nothing to do with it.
- Real repro (ga-1qepfl, mechanism 2): `ga-3fw26n`, a bead correctly held `in_progress` for its own merge-tracking, blocked pushes for unrelated `deploy/*-gate` branches from the same session identity. The blocking bead wasn't misbehaving — it was doing exactly what it was told to do. There was no error on the blocking side to fix, only the false-positive match on the guard side.
- No metadata anywhere links a deploy-gate tracking bead back to the branch/gated-bead it corresponds to, so this fix takes the smallest available correct signal: only trust the assignee fallback when the query returns **exactly one** in-progress bead. Zero or 2+ matches both resolve to "nothing to check" (ALLOW) — the same outcome as no assignment at all — rather than the fail-closed ambiguity sentinel, which stays reserved for genuine read failures (`assignee_read_failed`).
- Mechanism 1 (pooled-identity mismatch, same parent bead) is a separate, already-scoped-out fix and is not touched here.

## Test plan

- [x] New regression test `test_bead_id_deploy_gate_branch_ignores_ambiguous_inprogress_beads`: `deploy/ga-bucf4p-gate` branch, two unrelated in-progress beads in the assignee fallback, no `bd show` fixture configured for either — asserts the push is allowed (rc=0, no BLOCKED text). Confirmed RED against unmodified code (reproduced the real block), then GREEN after the fix.
- [x] Full suite: `bash scripts/test-push-ownership-guard.sh` — 31/31 passing, including all three pre-existing deploy-gate tests (prefers-live-assignee, allows-when-no-live-assignee, blocks-when-assignee-lookup-fails) and every general-path test.
- [x] `shellcheck scripts/push-ownership-guard.sh` — zero findings.
- [x] Pre-push fast suite (`make test-fast-parallel`) — all green on this branch.

Refs: ga-1qepfl

### DIFF
--- a/scripts/push-ownership-guard.sh
+++ b/scripts/push-ownership-guard.sh
@@ -137,7 +137,13 @@ _pog_read_with_retry() {
 #      an unrelated in-progress bead, e.g. one correctly held open for its
 #      own merge-tracking, was blocking pushes for a completely different
 #      deploy/*-gate branch because path 2 took .[0] of the match list
-#      unconditionally).
+#      unconditionally). This single-match requirement applies to EVERY
+#      branch shape that falls through to the fallback, not only
+#      deploy/*-gate — a non-gate branch that encodes no bead id and a
+#      session holding 2+ in-progress beads now resolves to nothing and is
+#      not checked at all. Deliberate: .[0] of an unordered multi-match was
+#      never sound enforcement in either direction. Pinned by
+#      test_bead_id_general_branch_ignores_ambiguous_inprogress_beads.
 # If both resolve and disagree, the branch match wins (it's the more
 # specific signal) and a warning goes to stderr — this is a best-effort
 # cross-check, not a hard failure, since branch-naming habits can
--- a/scripts/push-ownership-guard.sh
+++ b/scripts/push-ownership-guard.sh
@@ -128,9 +128,16 @@ _pog_read_with_retry() {
 #      wrong (and possibly closed) parent/child bead instead of the actual
 #      grandchild bead the branch is for. The literal 6-char-only pattern
 #      would misresolve to the root bead on any sub-bead's own branch.
-#   2. Falls back to this session's single in-progress assignment
+#   2. Falls back to this session's in-progress assignment
 #      (bd list --assignee="$GC_AGENT" --status=in_progress --json) when
-#      the branch name doesn't match.
+#      the branch name doesn't match. Only resolves when that query returns
+#      EXACTLY one bead — two or more in-progress beads for this session is
+#      not a positive identification of which one (if any) this push is
+#      for, so it resolves the same as finding none (ga-1qepfl mechanism 2:
+#      an unrelated in-progress bead, e.g. one correctly held open for its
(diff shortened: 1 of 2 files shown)


quad341:
@quad341 — the diagnosis here is the good part. You resisted the obvious framing: `ga-3fw26n` wasn't misbehaving, it was doing exactly what it was told, and the bug was entirely on the guard's side taking `.[0]` of an unordered multi-match. Routing 2+ matches to "nothing to check" instead of the fail-closed sentinel is the right call — keeping `assignee_read_failed` as the only thing that trips ambiguity preserves the fail-closed contract exactly where it matters, and `deploy-gate-branch-blocks-when-assignee-lookup-fails` still passes.

I re-ran your evidence instead of taking it: shellcheck clean, 31/31 green, and the new test really is RED against the pre-fix resolver.

One thing I'll take on my side before merging: the fix lives in shared path 2, so it also relaxes the check on non-gate branches — I reproduced a `chore/` branch with two in-progress beads now sailing past a `hold:mayor` that previously blocked. Same reasoning as your fix says that's correct, so I'll add a general-path regression test pinning it plus a comment noting the single-match rule isn't deploy-gate-specific, then merge. Your logic stands as-is.

Thanks for the honest KNOWN LIMITATION bookkeeping — it made this fast to verify.


---


quad341:
@quad341 — the diagnosis is the best part of this one. `ga-3fw26n` wasn't misbehaving; it was doing exactly what it was told, and the bug was entirely the guard taking `.[0]` of an unordered multi-match. Routing 2+ matches to "nothing to check" rather than the fail-closed sentinel is the right call — it keeps `assignee_read_failed` as the only thing that trips ambiguity, which is where fail-closed belongs.

We re-ran your evidence instead of taking it: shellcheck clean, 32/32 green (the body's 31/31 predates the second test), and reverting the count guard in place confirms both new tests really are RED against the pre-fix resolver — including the `hold:mayor` fixture in the general-path test, which does block without this change.

The general-path test and the "this isn't deploy-gate-specific" comment are already on this branch, so there's no maintainer-side work left. CI is green on `7f429ca`. Merging as-is.

Thanks especially for the KNOWN LIMITATION bookkeeping — saying out loud that a single unrelated in-progress bead can still false-positive is what made this fast to verify instead of slow to trust.


---

quad341:
Re-reviewed `7f429ca5e`..`2c5d3c751` (1 new commit since the last review): no new concerns — the prior review stands.

Issue #5093, opened 2026-08-07
ci: packages-core-2-of-4 flaky under shard load — rotating failures, productmetrics wall-clock budgets implicated


julianknutsen:
Record of a flake pattern observed while driving recent merge trains, so future "is this my failure?" triage starts from evidence.

`Integration / packages-core-2-of-4` on main failed 2 of 3 runs on 2026-08-07 afternoon, on **rotating, unrelated tests**:

- run on `f64b6262aa` (15:29): `TestRecordOnceSpentBudgetAfterReservationLeavesOnlySafeOvercount` (`load quota: absent`) and `TestFallbackCursorCleanupCrashWindowKeepsForegroundFailClosed` (`fixture RecordOnce = dropped`) — `internal/productmetrics` package time 207s on that shard.
- run on `a699601f08` (17:01): green.
- run on `ad6d62fdae` (18:49): `TestDockerSessionProtocol/context_cancellation_rolls_back_created_container`. The same run also failed `Integration / rest-full-3-of-16` on `TestGCLiveContract_BeadsAndEvents` (99.88s).

The productmetrics pair looks load-induced rather than logic: `RecordOnce` runs under a wall-clock decision budget (`defaultRecordDecisionBudget`), and a shard where the package takes 207s is exactly where a fixture's `RecordOnce` can blow that budget and drop, or a quota staging write can lag the fixture's read. Both tests pass 20/20 locally on the same commit set.

Possible directions, not prescribed: per-fixture budget overrides in tests that only assert protocol (not latency), or `testutil`-style deadline scaling under known-contended CI shards.

ckumar1:
Mechanism for the productmetrics half of this, with a fix in #6448.

The wall-clock budget reading in the original report is the right neighbourhood, but it does not explain your own observation that both tests "pass 20/20 locally on the same commit set" — if a slow shard merely exceeded a wall-clock budget, a slow local run would reproduce it.

`RecordOnce` takes its decision window from an **injected** clock, which these tests deliberately freeze, while the state-lock wait *inside* that window is bounded by a **real** deadline. Frozen test-time never advances toward the budget, so the whole of `defaultRecordDecisionBudget` (50ms in these fixtures) is spent by real time alone: any lock-file fsync slower than that drops the record, giving exactly `fixture RecordOnce = dropped` and `load quota: absent`. Locally an fsync is sub-millisecond and the budget is never spent.

[...]

PR #5392, opened 2026-08-19
fix(proctable): never classify a tmux server as an agent root

### LINKED ISSUE #5837: proctable: a tmux server founded by a session is classified as an agent root, so the orphan sweep kills the whole city
## Symptom
After session churn (a mayor handoff in our case) the orphan sweep kills the city's shared tmux server, and every session in the city dies with it. A second downstream deployment reported the same whole-fleet loss mode independently (see the comment thread on the linked PR).

## Root cause
`ScanBySessionID` and `IsScanRoot` (`internal/runtime/proctable/scan_darwin.go`; the Linux twin via `isRootWithSessionID`) decide whether a process is an agent root by testing its **parent** only: inside the envelope when the parent carries the same `GC_SESSION_ID` and is not infrastructure. The record's own command is never consulted. A tmux server founded by a session's first `new-session` call inherits that session's `GC_SESSION_ID` and then reparents to launchd/init, so it passes the parent test and is reported as a root; the sweep hands it to the kill path. One tmux socket per city is the default topology (`tmuxConfigFromSession` defaults `SocketName` to the city name), so killing that server is the whole city.

## Reproduction
A process table containing a `tmux: server` record with a session's id and ppid 1, plus the agent beneath it: the scan reports both as roots. Encoded as unit tests against a fake process table in the linked PR.

## Environment

## Proposed fix
#5392 — one clause on both platforms: a process that is itself infrastructure is never an agent root, whoever its parent is.

A3Ackerman:
## The defect

`ScanBySessionID` and `IsScanRoot` decide whether a process is an agent root by testing its **parent**: a process is inside its session's envelope when the parent carries the same `GC_SESSION_ID` and is not infrastructure. The record's own command is never consulted. A tmux server founded by a session's first `new-session` call inherits that session's `GC_SESSION_ID`, then reparents to launchd/init — so it passes the parent test and is reported as an agent root. The orphan sweep hands it to the kill path, and because one tmux socket per city is the default topology (`tmuxConfigFromSession` defaults `SocketName` to the city name), killing that server takes down every session in the city.

Observed as whole-fleet loss on a mayor handoff; independently confirmed by a second downstream deployment on the earlier revision of this PR (see the comment thread).

## The fix

One clause on both platforms: **a process that is itself infrastructure is never an agent root, whoever its parent is.** Darwin (`scan_darwin.go`) gains the guard in both `ScanBySessionID` and `IsScanRoot`, with the decision extracted into two pure helpers so it can be unit-tested against a fake process table. Linux (`scan_linux.go`) gains the same guard in `scanWithRoot` and `IsScanRoot`, reusing the existing comm-based check — renamed from `isInfrastructureParent` to `isInfrastructureProcess` since it no longer applies only to parents.

## Tests

- `TestScanRecordsBySessionIDNeverReportsInfrastructureAsRoot` and `TestIsRecordScanRootRefusesInfrastructure` (darwin): a `tmux: server` record carrying the session's id with ppid 1 is not a root; the agent beneath it still is. Both fail with the guard neutered.
- `TestScanWithRootNeverReportsInfrastructureAsRoot` (linux): the same shape through the existing fake-`/proc` fixture.

## What this PR no longer carries

The earlier revision bundled this fix with a `ProcessTarget` / argv+cwd identity refactor (~900 lines). That refactor overlaps #5098 and introduces heuristic matching that deserves its own review; it is parked on a separate branch and will be proposed after #5098 settles, if still needed. This PR is only the urgent, narrow fix.

https://claude.ai/code/session_01Wtk5dCw6TtDSD5UzPDdSwa

## Second pass (review findings)

The infrastructure test is now an exact match on the path-stripped command name — `tmux`, `tmux: server`, `tmux: client` — shared by both scanners (`isInfrastructureCommand`; the Linux reader delegates to it instead of carrying its own copy). The earlier substring test had flipped its failure mode: a false positive on a parent only made the child a root (loud, self-healing), but a false positive on the record itself hid an agent from the sweep, so any `tmux-*` wrapper running as an agent root would have become a runtime the orphan sweep could never reap. `TestScanRecordsBySessionIDReportsTmuxWrapperAsRoot` pins that a `tmux-wrapper` binary is a root; `TestIsInfrastructureCommandMatchesExactNamesOnly` pins the three names. The Linux side now proves both directions too — the server is excluded *and* the agent beneath it is still a root — and both platforms' `IsScanRoot` doc comments describe the full rule.

Evidence on macOS 15 / Darwin 25.6 (this package is skipped by the Mac lane for fork PRs, so it has not run in CI here):

```
--- PASS: TestDarwinPSCommandIgnoresInlineTmuxEnv (0.00s)
--- PASS: TestDarwinPSCommandStillIdentifiesTmuxExecutable (0.00s)
--- PASS: TestIsInfrastructureCommandMatchesExactNamesOnly (0.00s)
--- PASS: TestScanRecordsBySessionIDNeverReportsInfrastructureAsRoot (0.00s)
--- PASS: TestIsRecordScanRootRefusesInfrastructure (0.00s)
--- PASS: TestScanRecordsBySessionIDReportsTmuxWrapperAsRoot (0.00s)
ok  	github.com/gastownhall/gascity/internal/runtime/proctable	0.400s
```

`GOOS=linux go vet ./internal/runtime/proctable/` is clean; the Linux tests run in the regular lanes. If a maintainer can dispatch the Mac suite for this branch, that would put the Darwin side on record in CI as well.

Fixes #5837.


A3Ackerman:
CI fix: pushed `00d4bb1` converting four `%v` verbs to `%w` in `internal/runtime/proctable/scan_linux.go` (errorlint failures from Preflight / static checks). Local `go vet`, Linux cross-compile, proctable tests, and `make lint-changed` all pass.

A3Ackerman:
Reframed the title and description above. No code changed — but what this PR *is* changed, so it is worth saying plainly why.

This was written up as a duplicate-background-runtime fix. It is that, but it also contains the fix for a **whole-city outage**: the orphan sweep can kill the tmux server, and one socket = one server = every session in the city. We have lost a 14-session fleet to it. The words "outage", "every session", and "handoff" appeared nowhere in the original description, which is on me — it is probably why this has sat without a review.

Three things I would like a reviewer's eye on specifically:

1. **The two halves are coupled and must not be split.** The `IsTracked` tightening, landed without the scan exclusion, makes the outage *more* likely, because `IsTracked` is currently the only thing keeping the server off the kill path. Details in the description.

2. **Please don't accept an env-scrub alternative.** The natural-looking fix — "stop leaking `GC_SESSION_ID` into the tmux environment" — cannot work on Darwin. The server has no `GC_SESSION_ID` in its environment; `ps eww` concatenates argv and environ, and the scanner misreads the `-e GC_SESSION_ID=...` *flag value* from the founding `new-session` argv. Measurement is in the description.

3. **CHANGELOG wording.** The current entry mentions the server exclusion as a trailing subordinate clause ("...while excluding tmux server processes and other managed session IDs"). Given the blast radius, that should probably be its own `Fixed` bullet leading with the outage. Happy to push that, or fold it into whatever a reviewer prefers — I left the file alone for now rather than churn the diff mid-review.

And the standing offer from the description: I will happily split the scan exclusion into a minimal standalone PR (two predicate placements plus one test) and rebase this refactor on top, if that gets the outage fix landed sooner. That seems worth a round trip to me, but it is the reviewer's call.

swedeinasia-flow:
Independent confirmation from our downstream production deployment: we hit the same whole-fleet loss mode and carried a compact hardening patch while auditing the upstream shape.

Our coverage adds two defenses that may be useful when this branch is rebased:

1. Exclude a record whose own command is tmux infrastructure from runtime-root discovery on Darwin and Linux, with a focused `GC_SESSION_ID`-bearing tmux-server fixture.
2. Re-check the live command immediately before termination and refuse to signal tmux infrastructure. This gives the kill choke point a second layer against stale attribution or PID reuse, independent of scanner correctness.

We also repaired the foreground-child cancellation fixture so readiness is published only after the child installs its interrupt trap; the strong process-group cancellation contract then passed 100 repeated runs and failed under a leader-only signal mutation.

Validated downstream with:

```shell
go test ./internal/runtime/proctable ./internal/runtime/subprocess
CGO_ENABLED=0 go test ./internal/runtime/tmux
go test ./internal/runtime/exec
go test -race ./internal/runtime/exec -run 'TestProvider_StartCancellation(Cooperative|InterruptsForegroundChild)' -count=20
```

I do not plan to open a competing issue or PR. Happy to provide the compact kill-guard regression/patch as a follow-up or help reconcile it into this branch once the current conflicts are resolved.

A3Ackerman:
Reframed and force-pushed: this PR is now **only** the two-line root-classification guard (both platforms) plus its regression tests — 4 files, no refactor. The `ProcessTarget`/argv+cwd identity work that used to ride along is parked on my fork (`held/proctable-process-target-refactor`) and will be proposed separately after #5098 settles, if it is still needed then. Splitting was the right call given the overlap with #5098 and the outage-urgency of the small fix; apologies for the earlier bundle.

A3Ackerman:
CI note: the red after the reframe was `Integration / packages-core-3-of-4` failing on `TestRecordOnceSpentBudgetAfterReservationLeavesOnlySafeOvercount` (`internal/productmetrics`) — the rotating shard-load flake recorded in #5093 — with `CI / required` and the evidence watchdog failing by cascade. Nothing in this PR touches that package (4 files, all under `internal/runtime/proctable`), and the same shard passed on a sibling PR's run at the same minute. Retriggered with an empty commit rather than retried blindly; if the same test fails again on this branch I'll treat it as mine to investigate, not a flake.

A3Ackerman:
Green after the retrigger: CI / required, CI / integration, Integration / packages-core-3-of-4 and Evidence / critical-path suites all pass now. The earlier red was the #5093 productmetrics shard flake; no code changed since the reframe.

A3Ackerman:
Green after the second pass: whole matrix passing (CI / integration=pass CI / required=pass Evidence / critical-path suites=pass Integration / packages-core-3-of-4=pass ). The Evidence red before this was the watchdog's 25-minute deadline expiring while CI / required was still running — filed as #5850 — not this change; the retrigger commit is empty.

quad341 (approved):
@A3Ackerman — thank you for splitting this. Cutting the ~900-line `ProcessTarget` refactor out of an outage fix, parking it behind #5098, and reducing this to one guard per platform plus tests is exactly the shape that gets reviewed fast.

The diagnosis holds up under checking. I traced your exclusion through `killExistingOrphans` → `FindRuntimesBySessionID` → `ScanBySessionID`, so dropping infrastructure from the scan really does take the server off the kill path. And your reasoning about the flipped failure mode — a false positive on the *record* hides an agent forever, where one on the parent is loud and self-healing — is the right instinct; the `tmux-wrapper` tests pin it nicely.

One thing I'll take on my side rather than bounce back: on Darwin, `psRecords` splits the `ps` line with `strings.Fields`, so a `tmux: server` command column arrives as the token `tmux:` and slips past the exact match. Your fixtures pass it pre-split, so the tests can't see it. I'll add `"tmux:"` to the switch plus a test through `darwinPSCommand(strings.Fields(...))`, then merge — your commits, your fix.

I'll also try to get the Mac lane dispatched so your darwin tests land in CI, not just on your laptop.


---

Issue #5370, opened 2026-08-18
bug: bead.closed dropped when a step closes seconds before its session is torn down — finished runs render as phantom 'blocked / needs operator' lanes


obj-imp:
## Summary

`bead.closed` is never emitted for a bead that closes in the last few seconds before its owning session is torn down. The run snapshot is built from `.gc/events.jsonl`, so those steps stay `ready`/`active` in the snapshot forever, and finished runs render on the dashboard as **blocked / needs operator** even though every step is `closed` in the store with `gc.outcome=pass`.

Because the last close of a run is, by construction, the one right before the session goes away, this preferentially hits **drain / final / last-attempt** steps — exactly the ones that decide whether a lane reads finished. The visible symptom is therefore almost always a fake "blocked" lane.

- gc **1.4.0** (Homebrew, darwin/arm64), single-host city, beads store on the managed dolt server, `bd` 1.1.0.
- Checked the v1.4.0…v1.4.1 diff (4 commits: beads pin to 1.2.2, waiver renewals, release chore) — nothing touches reconcile/event emission, so 1.4.1 is presumably affected too.

## Evidence

Observed 2026-08-08 on rig `travadro`. Four step beads are `closed` in the store but have **no** `bead.closed` event in `.gc/events.jsonl` (log intact back to seq 1, nothing rotated). In every case the owning session went away seconds after the close:

| step | role | run | closed (store) | next session event |
|---|---|---|---|---|
| `tr-mgb` | drain | `tr-tnf` (pass) | 19:41:23Z | `session.stopped` +14 s |
| `tr-9t1` | drain | `tr-nmt` (pass) | 19:45:35Z | `session.stopped` +6 s |
| `tr-jta` | implement | `tr-jl2` (fail) | 20:45:24Z | `session.updated` +8 s |
| `tr-2gy` | implement | `tr-jl2` (fail) | 20:57:52Z | new session +7 s |

Effect: `tr-tnf` and `tr-nmt` (both `gc.outcome=pass`) rendered as blocked lanes; `tr-jl2` (a genuine fail) rendered as a *live* run stuck on attempt 3 with an active worker. `blockedLanes` = 3 for runs that had all finished.

Not a stale cache: snapshots were being recomputed at near-head seq while still serving the wrong node statuses. A restart would not help — the event log itself lacks the facts.

## Root-cause hypothesis

[...]

PR #5393, opened 2026-08-19
feat(dashboard): mark event-gap-stalled runs stale instead of showing them live

julianknutsen:
## What

The factory dashboard run view showed 12–27-day-old "zombie" runs as active with stuck `active`/`failed` steps. The run view is an event-sourced projection; a run mid-flight when an event-emission gap hit is frozen forever — its terminal events never arrived, so `runproj` rendered its last-known mid-flight state as live. The projection had **no staleness handling** (only structural lane filters).

## Change

`EnrichRunSummary` now marks any in-flight lane whose latest bead activity is older than `staleRunSilenceMs` (**72h**) as `Stale` (with `StaleSince`), keeping it visible in `Lanes` but excluding it from `TotalActive`, `RunCounts`, and the census. It keys on **time-since-last-activity**, not total run age — real runs run for tens of hours (measured p90 ~17h, max ~33h) but emit events throughout, while a multi-day silence means the terminal events never arrived. Marking is **in place** (`RunLane.Stale`, `omitempty`) so `BuildRunSummary` and the non-enriched golden are untouched.

The SPA renders a greyscale "· stale" badge on such lanes (`LaneCard`) with a "no activity since …" tooltip; `RunLane` gains optional `stale`/`staleSince`.

## Tests

- `TestIsStaleRun` (thresholds/boundaries) and `TestEnrichRunSummaryMarksStaleLanes` (the active-looking zombie shape that passes the session-less latch). Full `runproj` suite + `go vet` + `gofmt` green.
- **No golden churn**: the enriched golden's `nowMs` moved inside the fixture's 72h window so its active beads still render active (health derivation is `nowMs`-independent); the stale path is covered by the new unit test.
- SPA change is additive optional fields + a guarded render; `dashboard-ci` typechecks it.

## Context

Part 1 of a plan to durably fix + gate the dashboard projection — every recent run-view breakage traces to projection reads/folds going blind when work is graph-resident or the event stream had a gap, with no gate. Follow-ups: agents-pane graph-read gap (N15) and a regression gate (dashport graph-routed + a runproj event→render contract test). This is the run-view staleness slice.

---
- Related to #5370 — adds a dashboard-side mitigation for the downstream symptom of that issue: the run view now marks an in-flight lane stale after 72h of bead-activity silence (the signature of terminal events that never arrived) instead of rendering its frozen mid-flight state as live. It does not touch the dropped `bead.closed` emit on session teardown, and lanes already in the `blocked` phase — the filed issue's headline phantom-blocked symptom — are explicitly excluded from the staleness rule, so both remain open.


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277193229

PR #5398, opened 2026-08-19
fix(dispatch): stop control beads under a workflow root that already settled

### LINKED ISSUE #5389: Workflow-root bead can close (outcome + closed_at) hours before its DAG's downstream/sibling steps finish settling
## Summary

The workflow-root bead created by `gc sling` for a graph.v2 formula (e.g.
`aria-ticket-build`) can record a terminal `gc.outcome` and `closed_at`
timestamp as soon as its immediate gating child step fails, while
sibling/downstream steps in the same DAG (e.g. `review`, `finalize`) keep
dispatching and retrying on their own bounded ralph controllers for hours
afterward. There does not appear to be any mechanism that makes the root
bead wait for the rest of the DAG to settle before recording its own
terminal state.

This produces a misleading signal for any external observer -- a human, a
dashboard, or an automation -- that reads only the root bead's own
`outcome`/`closed_at` to judge "is this ticket done" and "what did it
cost." The root bead's close time understates both the true wall-clock
duration and the true cost, sometimes by hours.

## Concrete example

On one of my tickets, the workflow-root bead
closed `gc.outcome=fail` at **15:38:30Z** -- 4 seconds after its gating
child (a review-loop controller step) exhausted its own `max_attempts`
budget. A downstream `finalize` step in the same DAG, gated on that same
review step, kept dispatching and failing on its own ralph controller
(identical transient bead-store-contention `INFRA_TIMEOUT` reason on each
attempt) until **20:36:55Z** -- almost 5 hours after the root bead had
already recorded its terminal outcome. A dashboard or human watching only
the root bead's close time would have believed the ticket died at 15:38
and stopped tracking the real (wasted) cost and wall-clock time that
continued accruing for hours after that.

I checked whether this is something I can address at the formula-TOML
level (a `needs`/dependency wire-up on our own steps) and found no lever:
there is no `[[steps]]` block in formula-spec-v2 that *is* the workflow-root
bead -- it's created by `gc sling`, not declared in any formula I author.
This appears to be internal engine behavior (ralph-attempt-exhaustion
propagating `fail` upward independent of sibling/downstream dispatch
state), not something we can patch from our own pack/formula content.

## Ask

Either:

1. Have the workflow-root bead wait for all in-flight descendant beads in
   its DAG to reach a terminal state before recording its own terminal
   `closed_at` timestamp (the `outcome` itself doesn't need to change --
   just the timing of when it's considered "closed"), or
2. Expose a separate, distinct "all descendants settled" signal (a
   metadata field, a status value, or a companion bead) alongside the
   root's own immediate-child-triggered outcome, so external consumers who
   want the DAG's true end time have something authoritative to read
   instead of reconstructing it themselves from the full bead graph.

## Why this matters

Any dashboard, alerting rule, or cost-attribution report that keys off the
workflow-root bead's own `outcome`/`closed_at` -- which is the most
natural and most likely thing for an external consumer to reach for first
-- will under-report both elapsed time and cost whenever downstream/sibling
steps outlive the root's own terminal timestamp. We currently work around
this locally by cross-checking `finalize`'s own bead and by running our own
critical-path reconstruction over the full bead graph, but that requires
knowing about this gap in the first place and having tooling to do the
reconstruction -- most consumers of `gc bd show <root>` won't.

## Related, but distinct

Possibly related to the same general class of "control-plane signal is
coarser than the underlying DAG state" issues as #5378 (check-step failures
not distinguishing transient/infra from genuine verdicts), but this is
about the workflow-root bead's own terminal-timestamp semantics
specifically, not a check-step retry-budget question.

No specifics about our own workloads or timing beyond the one example
above included intentionally -- happy to share more if useful.
I'm

quad341:
Reported in #5389 (thanks @dkisselev-zz — the report was precise enough to land straight on the mechanism).

## Root cause

`closeOrphanedControl` (`internal/dispatch/runtime.go`) is the root-state gate every control bead passes through before its kind-specific processing. It recognized two root states as stop signals — root **missing** (orphan) and root **canceled** — but **not a root already closed with a terminal outcome**. So once a workflow root settles, every control bead under it keeps being dispatched and keeps minting attempts.

The window is unbounded because of how settlement is ordered. `processWorkflowFinalize` closes the root *before* its bulk `molecule.CloseSubtreeWithMetadataExcept` sweep. That ordering is deliberate and correct — if the process dies after the root close, the finalizer is still open and the next serve cycle retries. But the sweep is all-or-nothing: when it fails (the reported run had sustained bead-store contention, `INFRA_TIMEOUT` on every attempt) the finalizer returns early via `recordWorkflowFinalizeError` and stays open, the root is **already closed**, and nothing stops the subtree.

That is the reporter's measurement: a downstream step still retrying on its own ralph controller almost 5 hours after the root recorded `gc.outcome=fail` and `closed_at`. Anything reading the root alone — a dashboard, an alert, a cost report — under-reports both wall-clock and cost for that entire window. Worth noting that `mapRunPhase` already treats a closed root as history ("root-terminality wins"), which is exactly why the continued burn was invisible.

## Fix

Gate per control rather than relying on one bulk write: a settled root closes each control bead the next time the dispatcher touches it, so settlement converges incrementally whether or not the sweep ever succeeds.

Residue is closed `gc.outcome=skipped`, matching what the finalizer's own sweep stamps. Deliberately not a failure — the control never ran to a verdict, and stamping `fail` would pollute outcome aggregation with beads that never executed.

## The teardown tail is exempt

This is the part that makes the change less trivial than it looks. Teardown runs **after** the root settles by contract — its pass condition may branch on `ROOT_OUTCOME`, which only finalize produces. That is precisely what #5271 fixed (the adopt-pr settlement deadlock), and `teardownTailExclusion` exists to keep the tail out of the finalizer's own sweep for the same reason. A naive "root closed → stop everything" gate re-breaks it, so the gate defers to `teardownTailExclusion` rather than restating the rule.

The cheap arm covers retry and ralph controls, which `expandRetry`/`expandRalph` mint via `cloneStep` and therefore *inherit* `gc.scope_role=teardown` (only the first attempt strips it). Fanout and scope-check controls are stamped `gc.scope_role=control` and keep only `gc.step_id` as their link back to a teardown step, so those need the subtree read — which happens only under an already-terminal root, at most once per control before the gate closes it, never on the live dispatch path.

## Why neither option from the issue

The issue asked for either (1) delaying the root's `closed_at` until the DAG settles, or (2) a separate "all descendants settled" signal. Neither is needed. The root's close is already ordered correctly and the finalizer already waits on all graph sinks; the defect was that settlement wasn't *enforced* afterwards. Bounding the residue window to a dispatcher cycle is what makes `closed_at` an accurate settle time — with no new metadata or wire surface.

## Known limits (stated, not hidden)

- This stops **work**. It does not replace the sweep for closing passive residue: an open scope body under a settled root dispatches nothing, so it is left to the sweep.
- `closeOrphanedControl` returns early when `gc.root_store_ref` is empty, so neither the pre-existing cancel gate nor this one fires on rows lacking it. graph.v2 stamps it. Unchanged here, not widened.
- Because the reporter asked for things this deliberately does not build, a maintainer should decide whether it closes #5389 — I did not use a closing keyword.

## Tests

`internal/dispatch/terminal_root_gate_test.go`:

- `TestProcessControlStopsControlWhenWorkflowRootTerminallyClosed` — the red test. Before this change, `ProcessControl` on a retry controller whose root was already closed `gc.outcome=fail` returned `{Action:retry Created:1}`: it minted attempt 2 under a terminally closed root. Subtests cover `fail`/`pass`/`skipped`.
- `TestProcessControlKeepsTeardownTailRunningUnderTerminalRoot` — guards the #5271 contract.
- `TestProcessControlKeepsRunningUnderOpenRoot` — pins the hot path; an open root is not a stop signal.

An alternative hypothesis was tested and disproved rather than argued: "`abort_scope` propagates fail upward onto the root." It cannot — `abortScope` closes `snapshot.body`, and `resolveScopeBody` only ever selects `gc.kind==scope` beads, while a graph.v2 root is `gc.kind==workflow`. A probe confirmed a failing scope member closes the scope *body* and leaves the root `open`. `processWorkflowFinalize` is the only dispatcher path that closes a graph.v2 root.

## Verification

`go test ./internal/dispatch/` (full package); `go test ./cmd/gc/ -run 'Dispatch|Control|Workflow|Finalize|Graph|Convoy|Wisp|Teardown|Order'` → ok 50s; `internal/{formula,molecule,sling,sourceworkflow,graphroute,runproj}` ok; `go vet` clean; pre-commit lint clean; pre-push `make test-fast-parallel` → all fast jobs passed.

### DIFF
--- a/internal/dispatch/runtime.go
+++ b/internal/dispatch/runtime.go
@@ -196,6 +196,18 @@ func ProcessControl(store beads.Store, bead beads.Bead, opts ProcessOptions) (Co
 // teardown or a missing-root orphan close.
 const controlRootCanceledCloseReason = "control closed: workflow root canceled via run cancel"
 
+// controlRootSettledCloseReason is stamped on a control bead closed because its
+// workflow root had already settled, distinguishing terminal-root residue from
+// a cancellation, a skip teardown, or a missing-root orphan close.
+const controlRootSettledCloseReason = "control closed: workflow root already settled"
+
+// closeOrphanedControl is the root-state gate every control bead passes through
+// before its kind-specific processing: it reports handled=true when the bead's
+// workflow root is in a state that makes further work invalid. Three states
+// qualify — the root is gone (orphan), the root was canceled, or the root has
+// already settled. The finalizer is exempt because it is the bead that settles
+// the root, and the teardown tail is exempt because it runs after settlement by
+// contract.
 func closeOrphanedControl(store beads.Store, bead beads.Bead, opts ProcessOptions) (ControlResult, bool, error) {
 	if bead.Metadata[beadmeta.KindMetadataKey] == beadmeta.KindWorkflowFinalize {
 		return ControlResult{}, false, nil
@@ -216,6 +228,27 @@ func closeOrphanedControl(store beads.Store, bead beads.Bead, opts ProcessOption
 		// teeth.
 		if rootCanceled(root) {
 			return closeCanceledControl(store, bead, opts, rootID, rootStoreRef)
+		}
+		// A settled (closed) root is equally durable a stop signal. The
+		// finalizer closes the root BEFORE its bulk
+		// CloseSubtreeWithMetadataExcept sweep so a crash leaves the finalizer
+		// open to retry, but that sweep is all-or-nothing: when it fails —
+		// store contention is the reported trigger — the finalizer returns
+		// early and every control under the now-terminal root stays open and
+		// keeps minting attempts. gastownhall/gascity#5389 measured a
+		// downstream step still retrying almost 5 hours after its root
+		// recorded gc.outcome=fail and closed_at, which understates the run's
+		// true wall-clock and cost for anything reading the root alone.
+		// Gating per control makes settlement converge incrementally instead
+		// of depending on one bulk write succeeding.
+		if rootSettled(root) {
+			teardown, err := isTeardownTailControl(store, bead, rootID)
+			if err != nil {
+				return ControlResult{}, false, fmt.Errorf("%s: resolving teardown tail under settled root %s: %w", bead.ID, rootID, err)
+			}
+			if !teardown {
+				return closeSettledControl(store, bead, opts, rootID, rootStoreRef)
+			}
 		}
 		return ControlResult{}, false, nil
 	} else if !errors.Is(err, beads.ErrNotFound) {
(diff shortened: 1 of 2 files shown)


quad341:
@quad341 — this is excellent debugging. Tracing the unbounded retry window to the *ordering* inside `processWorkflowFinalize` — root closed before an all-or-nothing `CloseSubtreeWithMetadataExcept` that returns early under contention and strands the whole subtree — is what turns "controls kept running" into a one-line gate. And the five-hour measurement from #5389 lands exactly on it.

Two things I especially liked. Deferring to `teardownTailExclusion` instead of restating the teardown rule: I checked, and `isTeardownTailControl` is semantically identical to the predicate the finalizer already uses, so the gate and the sweep can't drift apart and re-open the #5271 deadlock. And disproving the `abort_scope` hypothesis with a probe rather than arguing it.

One thing I'll take on my side rather than send back to you: both teardown tests set `gc.scope_role=teardown` and short-circuit on the cheap arm, so the subtree-read path — a fanout or scope-check control reaching teardown only via `gc.step_id`, and a teardown retry's attempt 1 after `expandRetry` strips the role — is untested. I'll add that case, re-run `go test ./internal/dispatch/`, and merge with your authorship intact.

You were right to leave the closing keyword off #5389 — I'll make that call separately.


---

PR #5399, opened 2026-08-19
test(runproj): gate run projection end-to-end from bead.* events

julianknutsen:
## What

Adds the missing regression gate for the run projection. The `internal/runproj` tests were split into two worlds that never met: fold/projector tests drove `events → Fold` but asserted only the fold map, and the golden / terminal-clamp tests loaded a `[]beads.Bead` straight into `BuildRunSummary`/`BuildRunDetail` without ever touching `Fold`. **Nothing chained a real `bead.*` event stream through Fold/Projector into the two builders** — so when a store stopped emitting `bead.*` for graph.v2 workflow roots, runs silently went invisible and no test caught it.

## Tests (test-only, no production changes)

- `internal/runproj/golden_run_contract_test.go`
  - `TestContractGraphV2RunProjectsThroughEventsToSummaryAndDetail` — synthesizes a full-lifecycle `bead.*` stream (root + 2 steps, each `created → in_progress → closed`), drives it through `Fold` and a two-pass `Projector` (cold load then live-tail, exactly as the per-city tailer does), and asserts the run lists as an **active** lane mid-flight, moves to **historical** with `Phase=complete` on root close, and `BuildRunDetail` renders every node `completed` with `Progress.Terminal`. The mid-flight *"run went invisible"* assertion is exactly the regression that started this work.
  - `TestContractFailedStepPresentsFailedNodeThroughEvents` — a step closed with `gc.outcome=fail` surfaces as a `failed` node through the event chain (root left `in_progress` so the terminal clamp stays inert and the failure is unambiguously from `presentationStatus`).
- `internal/beads/runview_lifecycle_test.go` (`beads_test`) — the producer-fidelity variant: emits the same graph.v2 lifecycle through the **real producer** (`CachingStore.notifyChange`) and folds it, so a store that stops emitting `bead.*` — or a wire-shape / metadata drift — fails the gate.

Falsification-checked: dropping the root's `bead.*` events makes the run vanish from every summary bucket, proving the presence assertions actually catch the regression.

## Observations (surfaced, not fixed here)

- `snapshotForRun`'s dep synthesis (`snapshotDeps`) uses `members[0].ID` as the parent-edge root rather than the `rootID` it was given — only correct because the root's `created` event precedes its steps; a replayed/out-of-order log would mis-root the synthesized edges. Node identity/status are unaffected. Worth a follow-up.

## Context

Part of a plan to durably fix + gate the dashboard projection (every recent run-view breakage traces to projection reads/folds going blind, with no gate). Staleness robustness shipped as #5393; the agents-pane graph-read fix is a sibling PR.

### DIFF
--- a/internal/beads/runview_lifecycle_test.go
+++ b/internal/beads/runview_lifecycle_test.go
@@ -1,0 +1,141 @@
+package beads_test
+
+import (
+	"testing"
+	"time"
+
+	"github.com/gastownhall/gascity/internal/beadmeta"
+	"github.com/gastownhall/gascity/internal/beads"
+	"github.com/gastownhall/gascity/internal/events"
+	"github.com/gastownhall/gascity/internal/runproj"
+)
+
+// lifecycleRootID is the single graph.v2 run this test drives end to end.
+const lifecycleRootID = "gcg-lifecycle-root"
+
+// lifecycleCreatedAt timestamps every bead in the run so the projected updated_at
+// is populated (a run that happened at a real time), matching the sibling
+// round-trip test's fixture.
+var lifecycleCreatedAt = time.Date(2026, 7, 4, 8, 0, 0, 0, time.UTC)
+
+// TestRunViewLifecycleRoundTripFromNotifyChange is the producer-fidelity gate for
+// the run projection. TestRunViewRoundTripFromNotifyChange (this package) proves a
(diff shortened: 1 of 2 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277194107

PR #5400, opened 2026-08-19
fix(dashboard): count agents with active graph.v2 work as running

julianknutsen:
## What

The Agents pane showed "0 running" while the fleet was actively working. On a relocated-graph city (e.g. maintainer-city) agent work runs under agent-agnostic `gcg-session-<uuid>` **wisp** sessions, so `statusProviderRunning` — which probes the runtime for a session named exactly for the config agent (`bd__dog`) — never matches, and `findActiveBeadForAssignees` fans out over the **work stores only**, never the relocated graph store. So every agent read `stopped` / `running:0` even while its wisp actively executed its graph.v2 work. The only link from work→agent is the graph bead's `assignee`, which has the form `<sessionName>-gcg-session-<uuid>`.

## Change

Adds a graph-store leg (`graphActiveWorkBySession`) that lists the relocated graph store's in-progress beads **once per request** and attributes each to an agent by the session-name prefix embedded in the wisp `assignee` (split on `-gcg-session-`, which is exactly `agent.SessionNameFor(...)`). An agent with active graph-resident work is folded into the running signal (`effectiveRunning := running || hasGraphWork`) across the status count, the `/v0/agents` list + detail, and `computeAgentState` — so it reads `working` (with the bead's activity time) and is counted in `agents.running`.

Guarded exactly like the other relocated-graph legs (`graphStore != nil && graphStore != CityBeadStore()`): on a single-store city the map is nil, every lookup misses, and the read path is **byte-identical** to before — `TestGraphResidentSingleStoreByteIdentical` seeds the same wisp bead into the shared city store and asserts the agent stays `stopped`/uncounted. `computeAgentState`'s signature is unchanged; `ListQuery` gains no prefix field. No wire/OpenAPI impact (`TestOpenAPISpecInSync` passes; codegen produced no diff).

## Tests

`internal/api/handler_agents_graph_work_test.go` — working-counted, single-store-inert, negative (no match), no-double-count, and the assignee-splitter unit. The assignee format was confirmed against the live relocated graph store (`bd__dog`, `gc__run-operator`, `gc__implementation-reviewer`, … — each `SanitizeQualifiedNameForSession(qualifiedName)`). Full `internal/api` suite green (91s); `go vet` + `gofmt` clean.

## Context

Part of a plan to durably fix + gate the dashboard projection. Sibling PRs: run-view staleness #5393, projection gate #5399. This is the agents-pane graph-read slice (gap N15 in the graph-read-gap analysis).

### DIFF
--- a/internal/api/handler_agents_graph_work.go
+++ b/internal/api/handler_agents_graph_work.go
@@ -1,0 +1,104 @@
+package api
+
+import (
+	"strings"
+	"time"
+
+	"github.com/gastownhall/gascity/internal/beads"
+)
+
+// graphSessionAssigneeInfix is the literal separating an agent's session name
+// from the wisp session id in a graph-resident (graph.v2) work bead's assignee.
+// A relocated graph store names wisp assignees
+// "<sessionName>-gcg-session-<uuid>", where <sessionName> is exactly
+// agent.SessionNameFor(city, qualifiedName, template) — the same string the
+// runtime provider is probed with by statusProviderRunning. Splitting on this
+// infix recovers the agent session-name prefix, so work executed under an
+// agent-agnostic wisp session is still attributed to the agent that owns it.
+//
+// Confirmed against a live relocated graph store: every in-progress wisp
+// assignee has the form "<sanitized-agent-name>-gcg-session-<32 hex>", e.g.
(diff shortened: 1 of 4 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277194246

PR #5401, opened 2026-08-19
Reconcile file backup remotes after Dolt compaction

quad341:
## Summary

- Discover every `file://` backup remote except the authoritative remote after Dolt compaction.
- Reconcile each backup through the existing post-compaction push path with a per-database, per-remote pending marker.
- Preserve the authoritative push return code and isolate backup push failures so they never block the primary result.

## Testing

- [x] Three diff-owned backup-reconciliation tests: 3 PASS, 0 FAIL, 0 SKIP.
- [x] Complete affected package (`./examples/bd/dolt`): 355 PASS, 0 FAIL, 0 SKIP.
- [x] Policy lane, affected lint, formatting, `go build ./...`, `go vet ./...`, shell syntax, and `git diff --check` passed on the reviewed SHA.
- [ ] The retained broad sweep remains red: 35 PASS / 5 FAIL / 0 SKIP jobs. Criterion 3 remains **FAIL — WAIVED** under the Mayor's bead-specific 2026-08-18 ruling recorded on `ga-f8it32`; the original failures and counts are preserved in the gate record.

## Release gate

- Reviewed commit: `f590515412a08fa320876422786e6dac775a9849`
- Review bead: `ga-kiapum` — PASS
- Build bead: `ga-s52yku`
- Deploy bead: `ga-f8it32`
- Gate record: `release-gates/ga-f8it32-compaction-backup-remotes-gate.md`

### DIFF
--- a/examples/bd/dolt/commands/compact/run.sh
+++ b/examples/bd/dolt/commands/compact/run.sh
@@ -428,6 +428,7 @@ lock_pid_path="$lock_dir/pid"
 lock_cmd_path="$lock_dir/cmd"
 pending_gc_dir="$PACK_STATE_DIR/compact-pending-gc"
 pending_push_dir="$PACK_STATE_DIR/compact-pending-push"
+pending_push_backup_dir="$PACK_STATE_DIR/compact-pending-push-backup"
 quarantine_dir="$PACK_STATE_DIR/compact-quarantine"
 
 # DB discovery uses rig metadata.json files first (authoritative), with a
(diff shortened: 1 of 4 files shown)


quad341:
@quad341 — the scheme-based selection is the right instinct here: no hardcoded `usb`, no magic remote name, a remote qualifies purely by being `file://` and not the authoritative pick. And the isolation work is careful — `$db.$name` marker namespace, `|| true` around the reconcile, `primary_push_rc` captured so a backup can never mask the primary result. I traced that path; it holds.

Two things I'll take on my side and land with credit to you:

1. `run.sh` is `#!/bin/sh` with no `local`, and the marker helpers all do `db="$2"`. Passing `push_marker_key="$db.$name"` clobbers the global — the success line prints `db=beads.backup`, and a *second* file:// backup would push against `beads.backupA`. Parameter rename plus a two-backup test.
2. Nothing drains `compact-pending-push-backup`; the pre-threshold retry only scans `compact-pending-push`. I'll wire the retry.

One question before I push. `reconcile_backup_remotes` passes `expected_remote_head=""` / `push_context="initial"`, so the backup must clear the ancestry guard — and post-flatten the local log is just `{root, flatten commit}`. A backup sitting at the *previous* flatten commit gets deferred, never pushed, and `gc dolt sync` only ever pushes one remote. Your three tests pin it at `headcommit`, so they pass either way. Should backups skip that guard — the push is already `--force` and we're authoritative over a mirror — or is something reseeding the drive that I'm not seeing? That one's yours to call; I didn't want to make a data-loss-adjacent choice for you, and it cuts against the "reuse unchanged" criterion in your gate record.


---

PR #5403, opened 2026-08-19
Use production-default timeout in stop-fallback tests

quad341:
## Summary

- Replace three stop-fallback test call sites' live one-second cap with the production `0` sentinel.
- Exercise the same config-derived timeout path used by the CLI default.
- Keep production behavior and existing assertions unchanged while removing shard-load sensitivity.

## Testing

- [x] Three diff-owned stop tests: 15 PASS, 0 FAIL, 0 SKIP across five repeat runs.
- [x] Full `cmd/gc` package regression: 9,037 PASS, 0 FAIL, 99 environment-gated SKIP.
- [x] All six required `cmd/gc` process shards and all `cmd/gc` integration shards passed under load.
- [x] Policy, lint, formatting, `go build ./cmd/gc/...`, and `go vet ./cmd/gc/...` passed on the reviewed SHA.
- [ ] The retained broad sweep remains red: 36 PASS / 4 FAIL / 0 SKIP jobs. Criterion 3 remains **FAIL — WAIVED** under the Mayor's bead-specific 2026-08-18 ruling recorded on `ga-tavdtc`; all original failures remain recorded in the gate artifact.

## Release gate

- Reviewed commit: `bb9d53162f772ca91805eb22771747525808c61f`
- Review bead: `ga-isjlsh` — PASS
- Build bead: `ga-j8z56i`
- Deploy bead: `ga-tavdtc`
- Gate record: `release-gates/ga-tavdtc-stop-fallback-timeout-gate.md`

### DIFF
--- a/cmd/gc/cmd_stop_test.go
+++ b/cmd/gc/cmd_stop_test.go
@@ -534,7 +534,7 @@ func TestCmdStopSupervisorManagedInvalidCityTomlWaitsForControllerStop(t *testin
 	}
 
 	var stdout, stderr lockedBuffer
-	code := cmdStop([]string{cityDir}, &stdout, &stderr, time.Second, false)
+	code := cmdStop([]string{cityDir}, &stdout, &stderr, 0, false)
 	if code != 0 {
 		t.Fatalf("cmdStop() = %d, want 0; stdout=%q stderr=%q", code, stdout.String(), stderr.String())
 	}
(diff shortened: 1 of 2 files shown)


quad341:
@quad341 — nice diagnosis on this one. Swapping the live `time.Second` cap for the production `0` sentinel at those three stop-fallback call sites is the right fix rather than the tempting one: bumping the constant would have papered over shard contention, whereas `0` makes the tests run the same config-derived path an operator actually gets from `gc stop`. I traced it in the head checkout — the `--timeout` flag's own default is `0`, documented as "derive from city config" — and it lines these three up with the nine sites in that file already passing `0`.

I especially appreciated that you left the two tests that genuinely assert timeout behavior alone. `TestCmdStopWallClockTimeoutBoundsDirectStop` and its supervisor-managed sibling keep their explicit `testWallClockCap`, so we lose no coverage of the cap itself. That's the distinction that's easy to get wrong here.

One note for the record, nothing to do: on the two supervisor-managed tests the alive-supervisor early return lands before the cap is installed at all, so those run uncapped rather than on a derived cap. Still production-default behavior — just slightly narrower than "config-derived" reads.

There are four `note`-level CodeQL advisories on `go.mod`; they predate your branch and aren't yours to carry. Merging as-is, with authority staying with mayor/mpr per your note.


---

PR #5405, opened 2026-08-19
test(api): pin both rig-path containment passes (ga-qe7qm)

julianknutsen:
## Summary

Test-only, plus one behaviour-preserving extraction. No behaviour change.

`assertRigPathWithinCity` guards every HTTP rig-create entry point — the sync
path (`controllerState.CreateRig`), the async `git_url` path
(`ProvisionRigFromGit`), and the physical teardown (`TeardownPartialRig`) — so a
remote API caller cannot steer the server's `MkdirAll`+store-write or its
clone/`RemoveAll` outside the city root. It runs two containment passes.

**Nothing pinned either pass.** No test referenced `relWithinCity` or
`realPathForContainment`, no test exercised a symlink escape on either API
rig-create path, and neutering the second pass to `return nil` left every
relevant selection green. This PR:

1. splits the two passes into `lexicalContainment` and
   `symlinkAwareContainment`, so each is independently falsifiable — a test
   that only calls `assertRigPathWithinCity` cannot tell *which* pass rejected,
   which is exactly how this rotted unnoticed;
2. adds the missing coverage: five negatives and one accept-side control.

## The "dead code" claim was investigated and disproved

The finding that prompted this work claimed the symlink-aware second pass is
unreachable — subsumed by the first — so the escape it exists to catch goes
unchecked. **That claim is false**, and the `pass2-nil` arm below is the direct
evidence: with the second pass neutered, all three end-to-end escape negatives
stay GREEN. Escapes were never unchecked.

The original observation was right; the inference from it was not. Over a real
escape layout (`<city>/link -> <root>/outside`) **both** passes reject, in all
four call shapes — raw and `resolveStoreScopeRoot`-resolved, existing and absent
leaf. The sweep saw the lexical pass reject first and concluded the second was
dead.

The second pass is a live backstop, not decoration. `pathutil.NormalizePathForCompare`
swallows every `EvalSymlinks` error and walks up to the nearest resolvable
ancestor, synthesizing an in-city-looking path — so for a path it cannot
canonicalize at all, **pass 1 accepts**. `realPathForContainment` returns the
error instead, so pass 2 fails closed. `TestSymlinkAwareContainmentIsLoadBearing`
pins three such inputs: a symlink loop, a path under a symlink loop, and a path
under a regular file (`ELOOP` and `ENOTDIR` are both non-`IsNotExist`).

## Mutation table — re-runnable

Every arm was confirmed to **build and vet** before its result was read; a
mutation that breaks the build proves nothing. Note that `pathutil` is
referenced *only* inside `lexicalContainment`, so the `pass1-nil` and `both-nil`
arms must keep the import live (`_ = pathutil.NormalizePathForCompare(cityPath)`)
or they fail to compile and prove nothing.

Selection used:

```
go test ./cmd/gc/ -count=1 -run \
 'TestAssertRigPathWithinCity|TestBothContainmentPasses|TestSymlinkAwareContainmentIsLoadBearing|TestContainmentPassesAccept|TestControllerStateCreateRigRejects|TestProvisionRigFromGitRejects'
```

| arm | mutation | result |
|---|---|---|
| baseline | none | **14/14 PASS** |
| `pass2-nil` | `symlinkAwareContainment` → `return nil` | **RED:** `BothContainmentPassesRejectSymlinkEscapeIndependently`, `SymlinkAwareContainmentIsLoadBearing`.<br>**GREEN (kept):** the three end-to-end escape negatives — with pass 2 dead, escapes are still refused by pass 1. *This is the evidence the reported claim is false.* |
| `pass1-nil` | `lexicalContainment` → `return nil` (import kept live) | **RED:** `BothContainmentPassesRejectSymlinkEscapeIndependently`.<br>**GREEN (kept):** the end-to-end escape negatives — pass 2 catches genuine escapes on its own too. |
| `both-nil` | both passes → `return nil` | **RED: all five negatives**, so none of them is vacuous. Also reddens three pre-existing negatives (`ProvisionRigFromGitRejectsEscapingRelativePath`, `ProvisionRigFromGitRejectsSymlinkedParent`, `ControllerStateCreateRigRejectsOutOfCityPath`). |
| `relwithin-always-reject` *(control)* | `relWithinCity` → always return the containment error | **RED:** `ContainmentPassesAcceptLegitimateInCityRigs` and both pre-existing accept tests (`AcceptsResolvedTargetUnderSymlinkedCity`, `AcceptsWhenBothSidesResolved`) — so this is not a guard that rejects everything. |

`TestSymlinkAwareContainmentIsLoadBearing` also goes RED under the control arm,
by design: its `t.Fatalf` precondition asserts pass 1 *accepts* the input, and
says so — if normalization ever changes such that pass 1 starts rejecting, that
arm no longer proves anything and must be rebuilt rather than silently passing.

## Testing

- [x] `go build ./...` — clean
- [x] `go vet ./...` — clean
- [x] `make test` (`GC_FAST_UNIT=1 go test -p=4 -count=1 -timeout 15m ./...`) — **PASS**, 170 packages, exit 0
- [x] `scripts/check-core-boundary.sh` — OK
- [x] `scripts/check-residency-boundary.sh` — OK
- [x] `.githooks/pre-commit` ran on the staged change (`lint-changed ./cmd/gc`: 0 issues; doc-gen; `go vet ./...`)
- [ ] `make check-docs` — n/a, no docs/navigation/link changes
- [ ] `make test-integration` — n/a, no runtime, controller, or workflow behaviour change

## Checklist

- [x] Linked an issue — `ga-qe7qm`
- [x] Added or updated tests for behavior changes — this PR *is* the tests; the
      only non-test change is the extraction, which is behaviour-preserving
- [x] Updated docs for user-facing changes — none; no user-facing surface changed

### DIFF
--- a/cmd/gc/api_state.go
+++ b/cmd/gc/api_state.go
@@ -1923,20 +1923,44 @@ func (cs *controllerState) DeleteAgent(name string) error {
 // staying byte-identical. The error wraps configedit.ErrValidation so the async
 // failure mapper renders invalid_request and the sync mapper renders a 4xx rather
 // than a 500.
+// The two passes are separate functions so each can be pinned on its own: a
+// test that only ever calls assertRigPathWithinCity cannot tell which pass
+// rejected, so a pass could rot to a no-op without reddening anything (ga-qe7qm).
 func assertRigPathWithinCity(cityPath, resolved string) error {
-	// Lexical check first: rejects "../" escapes and absolute paths that resolve
-	// to a sibling/parent of the city. Normalize both sides so a symlinked city
-	// ancestor (cityPath raw, resolved already resolveStoreScopeRoot-resolved)
-	// doesn't register as a false-positive escape.
-	normalizedCity := pathutil.NormalizePathForCompare(cityPath)
-	normalizedTarget := pathutil.NormalizePathForCompare(resolved)
-	if err := relWithinCity(normalizedCity, normalizedTarget); err != nil {
+	if err := lexicalContainment(cityPath, resolved); err != nil {
 		return err
 	}
-	// Symlink-aware check: a "../"-free lexical path can still escape through a
-	// symlinked ancestor (e.g. <city>/link -> /outside, then a clone into
-	// link/rig). Canonicalize the city root and the nearest EXISTING ancestor of
-	// the (not-yet-created) target and re-check containment on the real paths.
+	return symlinkAwareContainment(cityPath, resolved)
+}
+
+// lexicalContainment is the first containment pass: it rejects "../" escapes and
+// absolute paths that resolve to a sibling/parent of the city. Both sides are
+// normalized so a symlinked city ancestor (cityPath raw, resolved already
+// resolveStoreScopeRoot-resolved) doesn't register as a false-positive escape.
+//
+// pathutil.NormalizePathForCompare resolves symlinks, so this pass already
+// rejects a symlinked-ancestor escape on its own; it is not purely lexical in
+// effect. What it does NOT do is fail closed when a component cannot be
+// canonicalized — NormalizePathForCompare swallows every EvalSymlinks error and
+// falls back to the lexical form. symlinkAwareContainment covers that gap.
+func lexicalContainment(cityPath, resolved string) error {
+	return relWithinCity(
+		pathutil.NormalizePathForCompare(cityPath),
+		pathutil.NormalizePathForCompare(resolved),
+	)
+}
+
+// symlinkAwareContainment is the second containment pass. It canonicalizes the
+// city root and the nearest EXISTING ancestor of the (not-yet-created) target
+// and re-checks containment on the real paths.
+//
+// It is a genuine second opinion, not decoration: unlike lexicalContainment it
+// fails CLOSED when a path component cannot be canonicalized at all (a symlink
+// loop, a non-directory component, an unsearchable ancestor). In those cases
+// NormalizePathForCompare silently synthesizes an in-city-looking path and the
+// first pass accepts; this pass refuses. TestSymlinkAwareContainmentIsLoadBearing
+// pins exactly those inputs.
+func symlinkAwareContainment(cityPath, resolved string) error {
 	realCity, err := filepath.EvalSymlinks(cityPath)
 	if err != nil {
 		realCity = filepath.Clean(cityPath)
--- a/cmd/gc/api_state_rig_path_containment_test.go
+++ b/cmd/gc/api_state_rig_path_containment_test.go
@@ -104,24 +104,36 @@ func TestBothContainmentPassesRejectSymlinkEscapeIndependently(t *testing.T) {
 func TestSymlinkAwareContainmentIsLoadBearing(t *testing.T) {
 	cityDir := t.TempDir()
 
-	// A symlink loop: EvalSymlinks returns ELOOP, which is not os.IsNotExist.
-	loop := filepath.Join(cityDir, "loop")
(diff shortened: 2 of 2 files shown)


julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR hardens the city-root rig-path containment guard: it extracts the guard's two inline passes into named helpers (`lexicalContainment` / `symlinkAwareContainment`) with no behavior change, and adds the first negative coverage proving the symlink-aware second pass is genuinely load-bearing — it is the only pass that rejects uncanonicalizable inputs such as symlink loops and paths nested under a regular file. That matters because the second pass is what fences remote-API-driven `os.RemoveAll` / `MkdirAll` against symlink-escape, and it previously had no regression pin, so it could have silently rotted to a no-op.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- The load-bearing second pass was pinned only in isolation (direct calls to the helpers); the end-to-end caller tests used symlink-escape inputs that the *first* pass already rejects. So the single most likely rot vector — deleting or reordering the `symlinkAwareContainment(...)` call inside `assertRigPathWithinCity`, the shared glue that fences a destructive `os.RemoveAll` in `TeardownPartialRig` — left the entire new suite green, which is exactly the regression this PR set out to prevent. The final patch adds `TestTeardownPartialRigRefusesUncanonicalizablePathThroughGlue`, which drives a first-accepts / second-rejects input through the production `TeardownPartialRig` glue and asserts the `os.RemoveAll` is refused and the escape target is left untouched. A mutation check confirmed that stubbing `symlinkAwareContainment` to `return nil` reddens exactly the three new load-bearing tests and nothing else, and the extraction was verified byte-identical and behavior-preserving.

Review findings addressed:
- **[Test Evidence Quality — correctness, major]** The symlink-aware second pass had no production-glue pin, so a refactor could drop it without any test going red. Fixed in `da48ecab` ("test(api): pin second rig-path containment pass through production glue (ga-qe7qm)"), which adds the end-to-end `TeardownPartialRig` assertion plus independent per-pass symlink-escape coverage.

- **ga-kyrm3** — [Error Handling & Resilience, minor] `realPathForContainment` can degrade *open* on a transient `IsNotExist`-class `EvalSymlinks` result, surfacing an FS hiccup as a containment false-green rather than a hard error. This is pre-existing (neither introduced nor widened by this PR) and was explicitly scoped out; filed separately to decide whether it should fail closed on a broader error class.
- **ga-sr4r4** — [Debuggability / Architectural Consistency, nit] The `lexicalContainment` name and comment understate that pass 1 already resolves ancestor symlinks via `NormalizePathForCompare` (a partially symlink-resolved form, not purely lexical); a maintainer could over-trust the "lexical" framing. Filed to tighten the comment or rename the helper. Not addressed here.

### Maintainer Changes
Maintainer fix pushed directly to the PR branch: `da48ecab` "test(api): pin second rig-path containment pass through production glue (ga-qe7qm)" — 1 file changed, +80/-13 (test-only; adds the production-glue pin and independent per-pass coverage the review required). Your contributor commit is preserved.

### Final Review Status
Ready for the merge queue: final head `da48ecab` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/32306363712

### Review Iterations
2 review passes performed: attempt 1 requested changes on one major finding (the second containment pass lacked a production-glue pin), the maintainer fix added that pin, and attempt 2 approved with only the two non-gating follow-ups above remaining.

---

PR #5407, opened 2026-08-19
fix(build): derive BUILD_TIME from HEAD commit date

quad341:
Summary

- derive BUILD_TIME from the immutable HEAD commit date instead of the wall clock
- preserve a literal fallback for non-git build contexts
- add deterministic regression coverage for source-derived build provenance

Release gate

- reviewer PASS on source commit 279316bde0bbdbe3c29ac762a0b9fb14cfcbced9
- gate record: release-gates/ga-xxh2cz-source-derived-build-time-gate.md at deploy commit 6629b1321970237cee123df4043c3a2cb9a8dca3
- focused scripts, policy, build, vet, format, lint, and provenance checks passed
- mandatory pre-push fast gate: 10 of 10 jobs passed
- full union: 36 of 40 jobs passed; four non-candidate failures preserved as FAIL — WAIVED under ga-550z2h, ga-afqddr, ga-hrdd3h, and the ga-rsktma signature ruling

Coverage waivers

Mayor mails gm-wisp-2rpri4 and gm-wisp-o7ixlp explicitly waive a live linker timing test and the pre-existing gc version long coverage gap. Both were independently verified during deploy evaluation.

Tracking

Deploy bead ga-xxh2cz; source bead ga-wf71cz; build bead ga-vx62cc. Merge authority remains with mayor/mpr.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ INSTALL_DIR := $(BIN_DIR)
 VERSION    := $(shell tag=$$(git describe --tags --exact-match 2>/dev/null || true); if [ -n "$$tag" ]; then printf '%s' "$$tag" | sed 's/^v//'; else echo "dev"; fi)
 COMMIT     := $(shell git rev-parse --short HEAD 2>/dev/null || echo "unknown")
 DIRTY      := $(shell test -n "$$(git status --porcelain 2>/dev/null)" && echo "-dirty" || true)
-BUILD_TIME := $(shell date -u +"%Y-%m-%dT%H:%M:%SZ")
+BUILD_TIME := $(shell git show -s --format=%cI HEAD 2>/dev/null || echo unknown)
 
 LDFLAGS := -X main.version=$(VERSION) \
            -X main.commit=$(COMMIT)$(DIRTY) \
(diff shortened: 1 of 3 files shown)


quad341:
@quad341 — thanks for chasing this to the actual mechanism rather than just making the timestamp prettier. Pinning `BUILD_TIME` to the HEAD commit date is the right call, and the comment on `TestBuildTimeIsSourceDerivedNotWallClock` explaining *why* — ldflags are part of the link step's cache key, so a wall clock guarantees a miss and strands a fresh binary in GOCACHE on every single relink — is exactly the kind of note that stops someone from helpfully reverting it in six months.

I traced the things that could have bitten. `main.date` reaches three places: the `Fprintf` in `gc version --long`, the JSON `date` field, and `hookStampLine` in `cmd/gc/hooks.go` — and that last reader does `bytes.Fields` and only checks the result is non-empty, so the `Z` → `-07:00` shape change passes through harmlessly. Nothing calls `time.Parse` on it, and `/health build_id` keys off the commit, not the date. GoReleaser injects its own `{{ .Date }}`, so releases keep a true build timestamp.

Merging as-is. One thing worth knowing rather than fixing: a dirty-tree build now stamps the commit's date, so `-dirty` on `main.commit` is doing all the work of telling you the binary isn't pristine. Right tradeoff — just a shift in what `built:` means. Tight two-file change with the gate record attached. Appreciated.


---

PR #5408, opened 2026-08-19
fix(nudge): make `gc nudge status` a lock-free read (ga-cn8dkj)

quad341:
## Symptom

`gc nudge status mayor --json` returned **nothing** — no output, no error — across
50–60 consecutive attempts at a 240s client timeout, spanning ~100 minutes, leaving
the mayor's operator queue unreadable. Reproduced live on this fleet while
investigating: a 120s run produced 0 bytes of output at **0% CPU**.

## Root cause

`cmdNudgeStatus` reached the queue through
`listQueuedNudgesForTarget` → `withNudgeQueueState` → `nudgequeue.WithState`, which

1. takes the **city-wide exclusive** flock at `internal/nudgequeue/state.go:111`
   (`syscall.LOCK_EX`, no `LOCK_NB`, no timeout), and then
2. runs the full maintenance sweep under it — `recoverExpiredInFlightNudges` /
   `pruneExpiredQueuedNudges` / `pruneDeadQueuedNudges`, which spawn **serial `bd`
   subprocesses** against Dolt — with `noMaintenanceDeadline()` (`now+24h`) so the
   pass never yields early.

Status only reads, but it paid the writer's price.

A goroutine dump of a blocked invocation pins it exactly:

```
syscall.Syscall(0x49, 0x4, 0x2, 0x0)          # 0x49 = flock(2), 0x2 = LOCK_EX
syscall.Flock(...)
github.com/gastownhall/gascity/internal/nudgequeue.WithState(...)
        internal/nudgequeue/state.go:111
main.withNudgeQueueState(...)
main.listQueuedNudgesForTarget(...)
main.cmdNudgeStatus(...)
```

Measured contention on this city (`/proc/locks`, 40s @ 0.5s sampling):

| Metric | Observed |
| --- | --- |
| Distinct lock holders in 40s | 2 (one held ~24s, the other ~16s) |
| Waiters queued, continuously | 13–16 |
| Hold-time samples with a live `bd` child | **230 / 250 (92%)** — 116 `bd list`, 63 `bd query`, 47 `bd update` |

~16 `gc nudge poll` daemons each re-enter this path every ~2s, so demand exceeds
supply by roughly two orders of magnitude and the queue never drains. `flock(2)`
provides no fairness, so a freshly-arriving reader is starved indefinitely — and
because it blocks *before* issuing any query, it produces no output and no error.
Blocked waiters observed included `gc mail send` (6m52s) and a
`gc nudge drain --inject` hook (5m59s).

**This is not a `read_timeout` problem.** The blocked reader never reaches SQL, so
no server- or client-side query timeout is in play. (For the record, the
`gm.wisp_events` scan hypothesis is also falsified independently: that table holds
**801 rows**, not 219,731, and the queries this path issues return in **~0.10s**.)

## Fix

Read the persisted state directly, without the lock.

`WithState` rewrites `state.json` atomically (temp file + `os.Rename`), so an
unlocked reader always observes one whole snapshot, never a torn one — and
`nudgequeue.LoadState` is already used lock-free elsewhere in this very function
for the `DispatchSkips` field, so this extends an existing precedent rather than
inventing one.

The new `listQueuedNudgesForTargetSnapshot` mirrors the *in-memory effect* of the
recover/prune passes so the displayed buckets match what a maintaining caller would
report, while mutating nothing and opening no bead store. Dead-letter retention
pruning is deliberately not mirrored: it depends on the bead store, and showing a
not-yet-swept dead letter is harmless in a status view.

Only `cmdNudgeStatus` changes. The 10 mutating `withNudgeQueueState` call sites and
the poller's own maintenance path are untouched.

## Tests

`cmd/gc/nudge_status_lockfree_ga_cn8dkj_test.go` holds the queue's exclusive flock
and asserts `cmdNudgeStatus` still returns promptly. It is a hermetic reproduction —
no fleet dependency.

- **RED** (before): `FAIL ... cmdNudgeStatus blocked on the held nudge-queue lock` after the full 10s budget.
- **GREEN** (after): passes in 1.19s.

Also run: full `-run Nudge` suite in `cmd/gc` (pass), `./internal/nudgequeue/...`
(pass), `go vet` (clean), `make test` (pass).

> One `make test` invocation failed on
> `TestCmdStopJSONReportsUnregisteredTrueForSupervisorManagedCity`, an unrelated
> stop-fallback test with known flakiness already being stabilized in #5403/#5404.
> A clean re-run passed.

## Out of scope

The underlying **convoy** is not fixed here, and it is a design question rather than
a mechanical one: the queue holds a global exclusive lock across serial subprocess
I/O, which still degrades nudge *delivery* and still blocks `gc mail send` and
`gc nudge drain` hooks. `internal/nudgequeue/store.go` documents holding that lock
across bead writes as a deliberate invariant, so changing it needs an architecture
call. Filed separately as **ga-jcvvm5**.

Also out of scope: the already-shipped `read_timeout` default change
(ga-lfcx72 / ga-id7z3d / ga-v138pa), which is correct on its own terms and simply
does not reach this bug.


quad341:
@quad341 — this is a genuinely excellent piece of debugging. The goroutine dump pinning `flock(2)` at `state.go:111`, the `/proc/locks` sampling showing 13–16 continuous waiters and a live `bd` child in 230/250 samples, and the explicit falsification of both the `read_timeout` and `gm.wisp_events` hypotheses left nothing to guess at. And the test is the right kind: it holds the real exclusive flock and asserts status still answers, so the regression can't quietly come back.

I checked the safety argument and it holds — `WithState` goes through `fsys.WriteFileAtomic` (temp + rename), so an unlocked reader always gets a whole snapshot.

One small thing I'll take on my side rather than send back to you: the snapshot re-buckets items correctly but doesn't apply the field updates the real sweep makes — recovered items keep a stale `claimed_at`/`lease_until`/`deliver_after`, newly-dead ones get no `dead_at`, and the buckets aren't re-sorted, so ordering can differ from a maintaining caller. I'll normalize that, add a bucket-equivalence test, re-run the nudge suites, and merge with your commit and authorship intact. Thanks for filing ga-jcvvm5 for the underlying convoy too — the scoping call there was right.


---


Review coverage:
- completed: `qwen codex`
- unavailable: `claude`

quad341:
@quad341 — this is a genuinely excellent piece of debugging. The goroutine dump pinning `flock(2)` at `state.go:111`, the `/proc/locks` sampling that caught 13–16 waiters with a live `bd` child in 230 of 250 samples, and the explicit falsification of *both* competing hypotheses (`read_timeout`, and the `gm.wisp_events` scan — 801 rows, not 219,731) left nothing to guess at. A read paying the writer's price, then producing no output and no error, is a nasty failure mode to chase.

The fix is right and it's minimal. I traced the snapshot against `recoverExpiredInFlightNudges` and `pruneExpiredQueuedNudges` myself — the branch order, the field projections, and the re-sort all line up, and `SortState`'s ID tiebreak means the bucket ordering matches regardless of where the recovered items land pre-sort. Both new tests and the full `Nudge` suite pass on my side too.

I especially appreciate you filing the convoy as `ga-jcvvm5` rather than widening this PR into an architecture change. That was the right call.

Merging now, with your commit and authorship intact. Thank you.


---


Review coverage:
- completed: `qwen codex`
- unavailable: `claude`

PR #313, opened 2026-04-05
fix: replace PreCompact gc prime with gc handoff in Claude default hooks

Rome-1:
## Problem

The default Claude hook config (`internal/hooks/config/claude.json`) registered `gc prime --hook` as the `PreCompact` hook. Since `SessionStart` already fires `gc prime --hook` on session resume, this caused the full agent context block to be injected **twice per compaction cycle** — once from `PreCompact` (right before compaction) and once from `SessionStart` (in the resumed session).

In long-running sessions with multiple compactions, the context block stacked many times. For a typical agent the block is 10–15KB; over a session with several handoffs or compactions it could accumulate 50–100KB of duplicated content, consuming the context window the injection was meant to protect.

## Fix

Change the `PreCompact` hook from `gc prime --hook` to `gc handoff "context cycle"`.

`gc handoff` is already the `PreCompact` behavior in all named pack overlays (`gastown/overlays/default`, `witness`, `crew`) and in individual agent settings. The embedded binary default was the only outlier — new cities initialized from the binary inherited the double-injection bug immediately.

`SessionStart` is unchanged: context injection on session start/resume is correct and intentional.

## Test

Added assertion to `TestInstallClaude` verifying the Claude default hooks contain `gc handoff` (for `PreCompact`) in addition to `gc prime` (for `SessionStart`).

Fixes: Rome-1/gas-city#rt-ivvsrn (internal tracking bead)

PR #5411, opened 2026-08-19
fix(deps): remediate govulncheck alerts 313, 315, 316 (ga-n6gvg8.2)

quad341:
## Problem

Three Go module vulnerability alerts (govulncheck #313, #315, #316) were
flagged against `golang.org/x/mod` and `github.com/klauspost/compress`,
tracked as remediable children of the code-scanning triage epic ga-n6gvg8:

- **GO-2026-6179** — `golang.org/x/mod` (fixed in v0.40.0)
- **GO-2026-6180** — `golang.org/x/mod` (fixed in v0.40.0)
- **GO-2026-5841** — `github.com/klauspost/compress` (fixed in v1.18.7)

## Fix

Bumped both modules to their fixed versions and let `go mod tidy` resolve
the minimal necessary transitive closure (indirect bumps to
`golang.org/x/crypto`, `golang.org/x/net`, `golang.org/x/telemetry`,
`golang.org/x/text`, `golang.org/x/tools` — all pulled forward by the two
direct bumps, no unrelated dependency churn). Only `go.mod`/`go.sum` changed.

## Validation

**govulncheck, before** (`golang.org/x/vuln/cmd/govulncheck@v1.7.0`, the
version pinned in `.github/workflows/govulncheck.yml`):

```
4 findings: GO-2026-6179, GO-2026-6180, GO-2026-5841, GO-2026-5932
exit 0 (all tier 2/3 — imported/required, not called; none block CI)
```

**govulncheck, after:**

```
1 finding: GO-2026-5932 only
exit 0
```

`GO-2026-5932` is deliberately untouched — it has no fixed version
available and is tracked separately as `ga-n6gvg8.6` (hold:mayor). This PR
only remediates the three alerts that have real fixes upstream.

| Check | Result |
|---|---|
| `go vet ./...` | clean |
| Fast test suite (10 jobs, standalone) | all pass |
| Fast test suite (via pre-push hook) | all pass |

## Scope

`go.mod` and `go.sum` only — no source changes. Part of the ga-n6gvg8
code-scanning triage epic; sibling to ga-n6gvg8.3 (workflow permission
scoping, not yet started) and ga-n6gvg8.6 (GO-2026-5932, out of scope here).


quad341:
@quad341 — thanks for this. The part worth calling out is that you didn't bump-and-hope: pinning the *same* govulncheck version the workflow uses (`v1.7.0`) for your before/after is the difference between "the number went down" and actual evidence. Two of us re-ran it here and got your exact after-state — 1 finding, tier 3, exit 0.

Equally glad you left GO-2026-5932 alone and said why. It has no fixed version upstream, and quietly bumping `x/crypto` further just to make a scan look cleaner would have been the wrong instinct. Splitting it to `ga-n6gvg8.6` is honest bookkeeping.

We also confirmed `go mod tidy` is a no-op on your tree, so the transitive closure really is minimal — no stray churn riding along with the two direct bumps.

One thing so it doesn't worry you: the red `Container Scan / Image vulnerabilities` check isn't yours. It's been failing on `main` since Aug 5, scans image layers rather than the module graph, and you touched no Dockerfile. The required CI gate is green at `acdf987`.

Merging as-is. 🚀


---

Issue #68, opened 2026-03-22
feat: provider rotation strategy for agents


alexsiri7:
## Problem

Currently, each agent has a single `provider` field — resolved once at spawn time. For cost-conscious deployments, it would be useful to rotate providers across restarts. For example, a witness agent could randomly pick between `gemini-flash` and `claude-haiku` on each spawn, spreading token usage across providers and providing resilience when one provider hits rate limits.

## Proposal

Add `providers` (list) and `provider_strategy` fields to the Agent config:

```toml
[[agent]]
name = "witness"
scope = "rig"
providers = ["gemini-flash", "claude-haiku"]
provider_strategy = "random"
```

**Behavior:**
- When `providers` is set (list), `ResolveProvider()` picks one based on `provider_strategy`
- When `provider` is set (string, existing behavior), nothing changes — full backward compatibility
- `provider_strategy` is extensible (template method / strategy interface) but only `random` is implemented initially
- Selection happens at spawn time in the controller, so each restart naturally gets a new provider

**Config precedence:** `providers` + `provider_strategy` takes priority over `provider` when both are set.

## Scope

- Add `Providers []string` and `ProviderStrategy string` to `Agent` struct
- Add `ProviderStrategy` interface with `Select(providers []string) string`
- Implement `RandomStrategy`
- Update `ResolveProvider()` to check for multi-provider config
- Update `AgentPatch`, `AgentOverride`, apply functions, pool deep-copy
- Tests for strategy selection, config parsing, backward compat

alexsiri7:
## Design: Stable Provider Selection with Re-roll on Fresh Creation

### Problem

`RandomStrategy.Select()` is called inside `ResolveProvider()` → `resolveTemplate()` on **every reconcile tick**. With multiple providers, each tick may pick a different provider → different command string → different `CoreFingerprint` → constant config-drift restarts. This makes multi-provider lists unusable.

### Design

**Principle:** Provider is selected once at session creation, stable while running, and re-rolled only on fresh session creation.

#### 1. Store selected provider in bead metadata

When `materializeSessionForAgentConfig` or `materializeSessionForTemplateWithOptions` creates a new session bead, store the resolved provider name:

```go
bead.Metadata["provider"] = resolved.Name
```

#### 2. On reconcile tick: read stored provider, don't re-roll

[...]

Issue #67, opened 2026-03-22
task: something to consider from a channel integration pov


csells:
https://github.com/sympozium-ai/sympozium

PR #5413, opened 2026-08-19
fix(ci): scope workflow write permissions to the jobs that need them (ga-n6gvg8.3)

quad341:
## Problem

Scorecard TokenPermissions alerts flagged top-level `permissions:` write grants that every job in the workflow inherits, when only one job in each actually needs the write:

- **#248** `docs-autofix.yml`: top-level `contents: write` + `pull-requests: write`
- **#68** `codeql.yml`: top-level `security-events: write`
- **#67** `dispatch-labeled-pr-suite.yml`: top-level `actions: write`

## Fix

Mirrors the job-scoped permissions pattern already established in `govulncheck.yml`: drop the workflow-level `permissions:` block to read-only, add an explicit job-level `permissions:` block (which fully **replaces**, not merges with, the workflow default) on only the job that needs the write.

| File | Workflow default (before → after) | Job-scoped write (added) |
|---|---|---|
| `docs-autofix.yml` | `contents:write, pull-requests:write, actions:read` → `actions:read` | `autofix`: `contents:write, pull-requests:write, actions:read` |
| `codeql.yml` | `actions:read, contents:read, security-events:write` → `actions:read, contents:read` | `analyze`: adds `security-events:write` (`runner-policy` keeps the read-only default) |
| `dispatch-labeled-pr-suite.yml` | `actions:write, contents:read, pull-requests:read` → `contents:read, pull-requests:read` | `dispatch-suite`: adds `actions:write` |

No behavior change — same effective token scope per job, just declared at the job boundary instead of inherited from the workflow default. Checkout trust boundaries, artifact access, PR comment/push, and dispatch behavior are all unchanged.

## Test plan

- [x] Added `scripts/workflow_permissions_test.go`: asserts workflow-level `permissions:` stays write-free and each named job (`autofix`, `analyze`, `dispatch-suite`) declares its write grant explicitly. Verified **red** against the pre-fix YAML (`git stash` the three workflow files, re-run — test fails), **green** after.
- [x] `actionlint` clean on all three edited workflow files.
- [x] `go vet ./...` clean.
- [x] Full `go test ./scripts/...` (89s) passing, including the unrelated `cipolicy` hash-pinning suite (confirms this change is outside that package's scope — it validates only `ci.yml`, `nightly.yml`, and `setup-gascity-ubuntu`).
- [x] `make test-fast-parallel` (pre-push hook) passing.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -17,7 +17,6 @@ on:
 permissions:
   actions: read
   contents: read
-  security-events: write
 
 jobs:
   runner-policy:
@@ -46,6 +45,10 @@ jobs:
     needs: runner-policy
     runs-on: ${{ needs.runner-policy.outputs.runner_16vcpu }}
     timeout-minutes: 30
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/dispatch-labeled-pr-suite.yml
+++ b/.github/workflows/dispatch-labeled-pr-suite.yml
@@ -6,7 +6,6 @@ on:
       - labeled
 
 permissions:
-  actions: write
   contents: read
   pull-requests: read
 
@@ -17,6 +16,10 @@ jobs:
       github.event.label.name == 'needs-mac' ||
       github.event.label.name == 'needs-review-formulas'
     runs-on: ubuntu-latest
+    permissions:
+      actions: write
+      contents: read
+      pull-requests: read
     steps:
       # This workflow never checks out or runs pull request code. It reads the
       # trusted base allowlist, then dispatches a dedicated workflow with the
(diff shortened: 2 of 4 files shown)


quad341:
@quad341 — nice piece of CI hygiene, and two things stood out.

First, you followed the pattern `govulncheck.yml` already established rather than inventing a new one. Second, you called out in the PR body that job-level `permissions:` **replaces** rather than merges with the workflow default. That's the exact footgun that makes this refactor risky, and naming it up front is why I could confirm each of the three job blocks quickly — all three are byte-identical to what they previously inherited, so the only job that actually loses anything is `codeql.yml`'s `runner-policy`, which never used `security-events: write` in the first place. That's precisely the alert.

Verifying `workflow_permissions_test.go` red against the pre-fix YAML before landing it green is also the right instinct — a permissions guard that would pass either way is worse than no guard at all.

Merging as-is. One idea if you're inclined to follow up: the test asserts the write scopes are present but not that the *read* scopes survived, and given replace-not-merge semantics that's the failure mode that would bite silently.

Thanks for closing out #67, #68, and #248.


---

Issue #5383, opened 2026-08-18
bug: 15s managed Dolt read timeout kills legitimate Reaper queries


coffeegoddd:
## Summary

Gas City's managed Dolt configuration sets `listener.read_timeout_millis` to 15 seconds in order to reap abandoned / sleeping client connections. In go-mysql-server, however, this setting is also enforced while waiting for the next query result row. As a result, legitimate long-running queries that take more than 15 seconds to produce their first row (or have an inter-row gap longer than 15 seconds) fail with:

```
row read wait bigger than connection timeout (errno 1105) (sqlstate HY000)
```

This is impacting Gas City's own Reaper sweep. Its recursive workflow-root query can take longer than 15 seconds as the database grows, and the server terminates it before it returns a result.

This is distinct from #3101, which tracks the order runner's outer wall-clock deadline. This report concerns the Dolt SQL listener's row-production timeout.

## Where Gas City sets the timeout

`internal/config/config.go:1911-1926` defines 15 seconds as the managed default:

```go
// DefaultDoltReadTimeoutMillis is the managed Dolt listener read timeout.
// Managed multi-agent cities open a short-lived bd/dolt-sql client
// connection per operation and frequently SIGKILL it on a client-side
// deadline (e.g. agents wrap `gc hook` in `timeout 10`), so the server
// orphans the socket in Sleep until read_timeout fires. Lowering this from
// the former 30s reaps those dead per-call connections sooner, before they
// accumulate into a store-wide read collapse under load. read_timeout is the
// listener socket idle/produce timeout: it reaps idle (Sleep) connections
// and bounds the inter-row produce gap (go-mysql-server ErrRowTimeout
// re-arms per row), not total query wall-clock ...
DefaultDoltReadTimeoutMillis = 15000
```

`cmd/gc/cmd_dolt_config.go:197-207` explains the idle-socket reaping intent and writes that value into the generated Dolt server config:

```yaml
# Managed multi-agent cities open a short-lived bd/dolt-sql client connection
# per operation. When the client process exits without a clean COM_QUIT, the
# server holds the socket in Sleep until read_timeout fires. A 5-minute
# read_timeout let these dead per-call connections pile to the hundreds
# (200-460 idle Sleep conns observed), burning Dolt CPU managing the swarm.
# The managed default reaps idle sockets promptly; city.toml [dolt]
# overrides can raise it for cities with slower live operations.
max_connections: %d
back_log: 50
max_connections_timeout_millis: 5000
[...]

PR #3626, opened 2026-06-20
fix(reconciler): floor scale_check demand re-eval to stop managed-Dolt read collapse (review-formulas wedge)

julianknutsen:
## What & why

The **Review Formulas** workflow was ~85% red on `main` (last 30 runs: 22 fail / 4 pass / 4 cancelled). It is **not a required check** (required = `Check` + CodeQL `Analyze`, both green), but it runs on every push to `main` and on PRs, so it is the visible "main is broken" signal — and it represents a real production defect, not just a test flake.

### Root cause (RCA + workflow-based adversarial review)

In `TestPersonalWorkFormulaCompileAndRun`, after the design-review `compose.expand` fan-out the single managed Dolt sql-server suffers a **read-side saturation collapse**: every `gc hook` ready-query times out (`rc=124`, `timeout 10`) for ~22 min until the 24-min deadline, so no agent can read the ready queue and the molecule never advances. Replacement polecats also time out — the **store itself is wedged**, not a stuck process. It is a cumulative-load threshold crossing, not a single-commit regression (`04eef3468` is a non-causal WARN→DEBUG log change).

Two stacking amplifiers on the per-op-subprocess store model:

1. **(dominant)** A polecat `scale_check` makes pool demand non-event-backed, so `demandSnapshotsEnabled()` is false and `shouldRefreshDemandSnapshot` rebuilt the full desired state — including the `scale_check` subprocess `bd ready --metadata-field gc.routed_to=polecat --limit=0` — on **every** patrol tick. At `patrol_interval=100ms` that ran ~10×/s, and the metadata-filtered probe cannot use `COUNT(*)` (`doltliteCountSupported` bails on metadata filters) so it fell back to a full hydrated `List` each time.
2. **(permanence)** A query killed by `timeout 10` (SIGKILL, no clean `COM_QUIT`) orphans a server-side `Sleep` connection until `read_timeout` (was 30s), so under load orphans accumulated faster than they reaped.

## The fix (3 parts)

- **`fix #1` (reconciler, the real fix):** floor patrol re-eval of a non-event-backed (`scale_check`) demand snapshot to `scaleCheckDemandMinInterval = 1s` via the new `demandSnapshotPatrolMaxAge()`. Non-patrol triggers (config reloads, sling pokes), config changes, session-fingerprint changes, and the no-event-provider case still rebuild immediately / every tick. **No-op at the 30s default `patrol_interval`** — it only bites pathologically fast (sub-second) cadences.
[...]

Issue #3101, opened 2026-06-04
Order timeouts: hard wall-clock deadline kills large sweeps; propose kill-on-stall / progress-heartbeat semantics


bourgois:
## Problem

Periodic sweep orders (feeders, reconcilers, reapers) have a hard wall-clock `timeout` configured in the order TOML. When a city grows beyond a certain number of rigs or beads, the natural fan-out time for one full sweep exceeds that timeout and the supervisor kills the order with `context deadline exceeded` on every tick — regardless of whether the order was making forward progress.

The result is that **the timeout encodes a city-size assumption** rather than a correctness invariant. Users compensate by overriding the timeout to a larger magic number (e.g. `600s`, `1200s`), but the right number keeps moving as the city grows, and a new city inherits broken defaults that only reveal themselves after enough rigs or beads accumulate.

### Reproduction sketch

1. Implement a sweep order whose outer loop iterates over each rig with one or more subprocess calls per rig (e.g. `bd list`, `gc rig list`).
2. Set a tight `timeout` (30–60 s) that is fine for two rigs.
3. Add eight or more rigs. On the next scheduled tick, observe `context deadline exceeded`; the order never finishes a full sweep.
4. Double the rig count again. The 2× timeout override you added in step 3 now also fails.

## What we want

The timeout should answer: **"is this order stuck?"** — not **"is this order slow?"**

A sweep that is methodically processing 20 rigs is not stuck; it is big. It should not be killed. An order that has emitted no output and acquired no new work for 30 s is genuinely stuck and should be killed.

**Proposed semantics: kill-on-stall via a progress heartbeat**

1. The order (or gascity on its behalf) emits a heartbeat signal at each meaningful unit of work (one rig processed, one bead routed, one step completed).
2. The supervisor tracks time-since-last-heartbeat.
3. When `time_since_last_heartbeat > stall_timeout`, kill the order.
4. The existing `timeout` becomes a hard cap for orders that emit no heartbeats at all (backwards-compatible safety net).

## Proposed config shape (sketch)

```toml
[order.timeouts]
# Hard cap — order must finish in under this time regardless (safety net).
# Zero = no cap.
wall_clock = "0s"

# Kill-on-stall — kill if no heartbeat for this long.
# Replaces the current single `timeout` for most sweep orders.
stall = "30s"
```

Orders that emit no heartbeats degrade gracefully to the existing hard-cap behaviour.

## Why this matters fleet-wide

[...]

PR #5414, opened 2026-08-19
Raise DefaultDoltReadTimeoutMillis to 120000 across Go config and shell fallback (ga-v138pa)

quad341:
## Summary

Raises the managed Dolt read-timeout default from 15,000 ms to 120,000 ms,
consistently across the canonical Go/config path, generated config docs, and
the materialized `gc-beads-bd` shell fallback (which had silently drifted out
of sync with the Go default).

- `internal/config/config.go`: `DefaultDoltReadTimeoutMillis` raised to
  120000, with corrected rationale comments (TDD red/green pair).
- `internal/doctor/checks_test.go`: managed-config fixture synced to the new
  default.
- `examples/bd/assets/scripts/gc-beads-bd.sh`: shell fallback's
  `read_timeout_millis` raised from 15000 to 120000 in both producer sites,
  closing the Go-vs-shell parity gap caught by
  `TestManagedDoltConfigGoWriterMatchesShellFallbackSemantics`.
- `docs/reference/config.md`, `docs/reference/schema/city-schema.{json,txt}`:
  generated docs/schema resynced.

## Release gate history

- **Round 1** (`9c6ccc8537f`, 2026-08-18): FAILed — the Go/config default was
  raised but the shell fallback was not, so the parity test caught a real
  diff-owned regression. Recorded in
  `release-gates/ga-v138pa-dolt-read-timeout-gate.md`.
- **Round 2** (this branch tip, `cf412ecdd4` + gate-record commit
  `6905f09677`): PASS. The shell-fallback fix closes the gap; the parity test
  now passes; re-review (`ga-vqrdih`, closed PASS) independently re-verified
  the diff, the rationale, and the acceptance test. Full verification
  evidence (build/vet/config/doctor/check-docs/process-tier shard results,
  and explicit criterion-3a(iii) attribution reasoning for two non-diff-owned
  flakes observed in the same run: `ga-uswva7`, `ga-tt3qwa`) is recorded in
  the same gate file, round 2 section.

## Test plan

- [x] `go build ./...`, `go vet ./...` clean
- [x] `go test ./internal/config/...`, `./internal/doctor/...` — ok
- [x] `make check-docs` — ok
- [x] `make test-cmd-gc-process-parallel` — target regression test
      (`TestManagedDoltConfigGoWriterMatchesShellFallbackSemantics`) passes;
      two unrelated non-diff-owned flakes attributed per criterion 3a(iii)
      (exact-signature match + zero path overlap + focused exact-base rerun)
      and tracked separately, not waived by assumption.
- [x] Pre-push mechanical gate (`test-fast-parallel`, 10 jobs) — all passed.

## Notes for reviewers

This PR is opened for review/merge routing only — **not merged by this
session**. Merge authority is operator/mayor/mpr only, per this repo's
`AGENTS.md` and the deploy bead's own instructions. Deploy bead: `ga-v138pa`.
Review beads: `ga-id7z3d` (round 1), `ga-vqrdih` (round 2). Build bead:
`ga-lfcx72`.

---
- Related to #5383 — raises the managed Dolt listener read_timeout_millis default from 15,000 ms to 120,000 ms across the Go config, the generated docs/schema, and the gc-beads-bd shell fallback, giving long-running maintenance queries headroom under the bound described in that issue; it does not touch the recursive workflow-root query in reaper.sh, and idle-connection reaping still rides on read_timeout because wait_timeout was measured inert on dolt 2.2.3 rather than taking over as that issue suggested


quad341:
@quad341 — thanks for chasing this one all the way down. The part worth calling out isn't the constant: it's that round 1 failed on the Go-vs-shell parity test and you fixed the *fallback* instead of the test. I checked all four `read_timeout` sites in `gc-beads-bd.sh`, including the pure-shell path at line 1329 that the parity test never reaches, and they all agree at 120000. That's the one that would have bitten someone on a host with no resolvable `gc` helper.

Routing the doctor contract through `EffectiveReadTimeoutMillis()` so only the fixture needed touching was the right call, and the comment-attribution guard is a genuinely good idea for a mistake that had already been made once.

Merging as-is. Two things for the record rather than blockers:

`#3101` appears nowhere in the tree except the comments this PR adds, and `write_timeout_millis` isn't really an outer deadline — the tripwire test is worth keeping, but the reasoning attached to it deserves a rewrite next time you're in there.

And since `read_timeout` is now the only reaper by your own measurement, 15s → 120s widens the orphaned-`Sleep` window 8× against `max_connections: 256`. Worth a `processlist` glance on a busy city after the next restart to confirm #3626 stays buried.


---

PR #5416, opened 2026-08-19
fix(beads): mint rig-scoped prefix for file-backed stores

jaword-anduril:
## Problem

In a **multi-rig, file-backed** city, all slung work strands: workers spawn but never claim (`gc analyze reliability` → `worker_operations: 0`), and `gc convoy status <id>` fails with:

> bead `gc-2` exists in multiple stores (…/rigA and …/rigB); resolution requires a uniquely addressable bead id

`gc doctor` is clean, so it doesn't look like a config error — but the per-rig `prefix` in `city.toml` is silently ignored.

## Root cause

The file-beads provider is the **only** store type that drops the scope's configured prefix at mint time. In `cmd/gc/api_state.go`, `openRigStore(provider, rigName, rigPath, prefix, cfg)` threads `rig.EffectivePrefix()` into:

- `OpenBdStore` → `bdStoreForRig(…, prefix)` ✅
- `OpenExecStore` → env `Prefix: prefix` ✅
- `OpenNativeStore` (Dolt) → scoped per-rig env ✅
- `OpenFileStore` → `openCompatibleFileStore(scopeRoot, cityPath)` → `beads.OpenFileStore(fs, path)` ❌ **prefix dropped**

`OpenFileStore` took no prefix and built a `MemStore` with an empty `IDPrefix`, so `MemStore.mintIDLocked()` fell back to `"gc"`. Every rig's file store then independently mints `gc-1, gc-2, …`, so the same `gc-N` exists in multiple stores and `resolveOwningStoreDir` (correctly) refuses to guess.

**Dolt-server cities were unaffected** — their stores already carry the prefix. This only bit the file-backed + multi-rig intersection.

## Fix

- Add `WithFileStoreIDPrefix` (mirrors `WithSQLiteStoreIDPrefix`) and apply options in `OpenFileStore`.
- Resolve each file store's scope root to its rig `EffectivePrefix()` (or `EffectiveHQPrefix` for the city store) in the openers, so file stores mint `asv2-1`, `oe-1`, … and convoy resolution is unique again.
- Falls back to the default `"gc"` when config is unavailable — preserves behavior for single-scope callers and tests. Blank/whitespace prefix also keeps the default.

## Testing

- New `TestFileStoreIDPrefix` (`internal/beads/filestore_test.go`): asserts default `gc-1`, prefixed `asv2-1`/`oe-1`, no cross-store collision, and blank-prefix fallback.
- Full `internal/beads` suite green (`CGO_ENABLED=0 go test ./internal/beads/`).
- Verified end-to-end on a real collided city: patched `gc sling` minted `asv2-45` (was `gc-45`).

## Migration note

This is **mint-side only** — new beads get correct prefixes, but existing on-disk `gc-*` ids in an already-collided city still need a one-time re-prefix to unstick current work. Worth a follow-up migration command (`gc beads reprefix` or similar); happy to add if you'd like it in-scope.

cc @quad341


quad341 (dismissed):
@jaword-anduril — thank you for this one. The write-up is what makes it: laying the four openers side by side and showing `OpenFileStore` was the single one not threading the prefix turned what reads like a config bug into a two-line root cause. `gc doctor` being clean is exactly why this was nasty to find, and you found it anyway.

I want to call out the decoy city in `TestOpenCompatibleFileStoreUsesCallerCityForIDPrefix`. That's the assertion that stops someone "simplifying" this back to ambient resolution in six months, and it's the kind of test people usually don't bother writing.

The red check is ours, not yours. `Evidence / critical-path suites` is our watchdog failing closed because our main CI never ran on your head commit — it's waiting on a maintainer approval click, and the watchdog dutifully reported the silence. I'm approving it now.

I'll add one test on my side for the upgrade case (an existing store full of `gc-*` beads minting `asv2-4` next), and note in the merge message that this changes the prefix for *every* file-backed city, not just multi-rig ones — since `EffectiveHQPrefix` derives from the city name when `[workspace] prefix` is unset. Right behavior, just wider than advertised.

Yes please on `gc beads reprefix` — separate PR, and I'll review it.


---

PR #5417, opened 2026-08-19
feat(dashboard): stalled/waiting-on-human cues + whole in-progress fetch on /beads

aphexcx:
The /beads board had two operator-facing gaps we hit running a multi-agent city:

1. **In-flight work fell out of view.** The bead list is a newest-created-first window (cap 1000); long-running in-progress beads age past it. On our city the store had 9 in-progress beads and the page showed 2 — the invisible 7 were exactly the stalled agents the operator needed to see.
2. **No liveness cues.** In-progress cards showed no assignee, no activity, and nothing distinguished a healthy worker from a dead session or a bead parked on a human checkpoint.

## Changes (SPA-only, no Go / API changes)

- **Whole in-progress fetch:** `listSupervisorBeads` runs a dedicated `status=in_progress` leg alongside the window and merges (dedupe by id). The set is tiny by definition, so nothing in flight can fall out of the window. A failed leg degrades to the window + the existing truncation banner.
- **Two new attention reasons** in the shared beads selector (nav badge + NEEDS YOU panel + row highlights, one SSOT):
  - `stalled` (attention): in_progress and the assignee shows no sign of life — no assignee, no session resolving to it (session_name / alias / id / `gc.session_id`; a live session is preferred over a dead one carrying the same recycled identity), a non-live session state, or no session/heartbeat activity for >60m.
  - `waiting-human` (watch, attention after 2h): the bead carries `gc.checkpoint_hold` / `gc.founder_gate` / `gc.checkpoint=awaiting-human` / a `hold:` label; the row names WHO it waits on (`gc.waiting_on`, else the hold text's parenthetical, else "human"). Checked before ready-unclaimed and stalled.
  - Honest degrade: a failed or partial sessions read skips session-dependent checks — a sessions outage never paints the board stalled, and a stall is never guessed from a stale heartbeat alone.
- **Liveness line on in-progress cards:** the waiting/stalled reason when one fires, otherwise `assignee · active Nm ago`, driven by the app-level 1s clock so ages advance without a refresh.

## Validation

- 921 vitest tests pass on this branch (new coverage for the selector rules, the merged fetch, and the degraded-read paths); `tsc` strict and eslint clean; `make dashboard-build` dist regenerated on this base.
- Verified against a live city: header went from "2 in progress" to "12 in progress"; beads parked on founder gates read "waiting on <names> for 2d"; three beads in progress since spring surfaced as stalled.

Known limits: the in-progress leg shares the 1000-item window cap (the truncation banner still reports it), and the generated client's `listSessions` takes no abort signal, so that read is not cancelled with the rest of the attention cohort.

---
- Related to #4387 — adds operator-visible detection of exactly this condition on the /beads board — an in-progress bead whose assignee has no session resolving to it, a non-live session, or no activity for over an hour is flagged stalled, and a dedicated in_progress fetch keeps long-running ones from aging out of the windowed list; it is dashboard-side display only and takes no reclaim action, so nothing reopens, unassigns, or re-dispatches the bead
- Related to #5741 — touches the same bounded `listSupervisorBeads` fetch window by adding a dedicated `status=in_progress` leg that is merged (dedupe by id) with the windowed read, so long-running in-flight beads no longer age out of view; it deliberately does not switch to per-type server-side queries, keeps the client-side bookkeeping filter, and the 1000-item cap still applies, so open engineering beads crowded out of the window are unchanged


quad341:
Thanks for this, @aphexcx — there is clearly a lot of work here.

quad341:
@aphexcx, a correction to our 09-04 comment, and I'm sorry about it. It said this PR was past a 20,000-line ceiling and asked you to split it. That was wrong: the change is under 900 lines. What tripped our reviewer was a separate 1 MB size limit, filled almost entirely by the rebuilt, minified `dashboardspa/dist` bundles on your branch. Our reviewer now leaves generated files out before that check, so there's no need to split anything.

I've queued it for a fresh review. The branch conflicts with `main` now; the review attempts the rebase itself, and I'll tell you if it needs anything from you.

quad341:
Thanks for this, @aphexcx — there is clearly a lot of work here.

I have not been able to review it, and I would rather say so than leave you waiting. The diff is 1.46MB, exceeds the 1MB byte ceiling (line estimate ~866, under the 20,000 line ceiling), so our review process cannot read it as a single change.

quad341:
@aphexcx, please disregard the automated comment above; I've hidden it. It was posted in error and repeated the mistake I had corrected just before it. Nothing needs splitting.

It fired because our reviewer leaves the generated `dashboardspa/dist` files out only after CI has verified them on the PR's current commit, and your branch predated that check. So I merged `main` into your branch and rebuilt the bundles from the merged source (34c33fda2a). I committed it under your name, so it adds no co-author to the final squash. The only conflicts were in `dist/`, and your 14 source files merged cleanly. The diff against `main` is now about 65 KB, small enough for the review to read in full. I'll follow up once it's done.

quad341:
@aphexcx — the part of this I want to single out is the degrade behavior. You could have shipped the stall check reading an empty session roster and had the board call every assigned bead dead during a sessions outage. Instead you made `sessions` optional, skipped the session-dependent checks when the read fails *or* comes back partial, and wrote a test literally named "does not guess stalled from a stale heartbeat alone." That's the failure mode that would have made this feature untrustworthy on day one, and you closed it before anyone asked. The dedicated `status=in_progress` leg is the right shape too — small set, merged by id, degrades to the window if it fails.

Two things I'll take on my side, neither needing anything from you. CI is red because our own `App.test.tsx` mocks the supervisor client with a single method, and the new sessions read is called inline in the `allSettled` array — it throws synchronously and orphans the other three legs. I'll wrap it in an async helper so no leg can do that again, and fix the mock. Second, we have a documented two-value hold-label taxonomy (`hold:mayor` / `hold:external`) your `hold:` branch predates; I'll map those to the right actor and keep `hold:external` off the two-hour escalation.

Rebasing, rebuilding the bundles, and landing it with your commits intact. Thanks for this one.


---

quad341:
@aphexcx — the part I want to single out is the degrade path. You could have let the stall check run against an empty session roster and had the board declare every assigned bead dead during a sessions outage. Instead you made `sessions` optional, skipped the session-dependent checks on a failed *or partial* read, and wrote a test named "does not guess stalled from a stale heartbeat alone." That's the failure mode that would have made this feature untrustworthy on day one, and you closed it before anyone asked. I also checked your `LIVE_SESSION_STATES` against the canonical session states in the Go lifecycle projection — it's right, including the ones you correctly left out.

CI is red and it isn't your logic. Our own `App.test.tsx` stubs the supervisor client with a single method, and the new sessions read sits inline in the `allSettled` array — so it throws synchronously, `allSettled` is never reached, and the sibling reads orphan as unhandled rejections. I'll wrap it in an async helper so no leg can do that again, and fix our mock.

Two small things I'll also take: we have a documented two-value hold taxonomy (`hold:mayor` / `hold:external`) your `hold:` branch predates, so I'll map those to the right actor and keep `hold:external` off the two-hour escalation; and I'll compute the truncation count from the window alone so the merged leg can't mask the banner.

Rebasing, rebuilding the bundles, landing with your commits intact.


---

quad341 (dismissed):
@aphexcx — the degrade path is what I want to single out, and I checked it rather than taking your word for it. You made `sessions` optional, discarded *partial* reads as well as failed ones, and wrote a test literally named "does not guess stalled from a stale heartbeat alone." A sessions outage declaring every assigned bead dead is the failure that would have made this feature untrustworthy on day one, and you closed it before anyone asked. I also verified `LIVE_SESSION_STATES` against our Go session lifecycle — it's right, including `awake` and the states you correctly left out.

CI is red and it isn't your logic. Our `App.test.tsx` stubs the supervisor client with a single method, and the new sessions read sits inline in the `allSettled` array, so it throws synchronously instead of settling. That's our mock and our call shape. I'll wrap the leg so none of them can do that again, and add the failure test we were missing.

Three smaller things on me: mapping our `hold:mayor`/`hold:external` labels so the row names the real actor, keeping `hold:external` off the 2h escalation, and computing the truncation count from the window alone.

Rebasing, rebuilding the bundles, landing with your commits intact. Thanks for this one.


---

quad341 (approved):
Re-reviewed `3e3458628`..`a635d051a` (6 new commits since the last review): no new concerns — the prior review stands.

PR #5419, opened 2026-08-19
fix(mail): keep a self-handoff addressed to one session on both sides (ga-28neu4)

quad341:
Investigation of ga-28neu4, which reported that 148 context-cycle compaction markers a day were
being addressed to other agents and firing false "you have mail" notifications.

**That is not happening, and never was: 0 of 826 markers all-time are addressed to another agent.**
But the report was an honest reading of what the store actually says, and this PR fixes the reason it
says the wrong thing.

## Root cause

`SendHandoff` normalizes only the sender. `intent.From` goes through `resolveSenderRoute`
(`beadmail.go:189` → `:233`), while `intent.To` is stored verbatim as the bead assignee
(`:197` → `:213`). An aliasless pool member can only address itself by its raw session ID, so
`senderDisplayAddress` falls through to the `SessionNameMetadata` branch (`:249-251`) and rewrites
From to `builder-<id>` while Assignee stays `<id>`.

`gc handoff --auto` — the PreCompact hook — passes one address for both sides
(`cmd/gc/cmd_handoff.go:314`). So a self-handoff persists as two different strings naming one session:

```
created_by                assignee            n
builder-gm-wisp-gpenji    gm-wisp-gpenji     44
builder-gm-wisp-chmzwy    gm-wisp-chmzwy     23
...                       17 sessions, same shape every row
```

Classified properly (`created_by = assignee OR created_by LIKE CONCAT('%-', assignee)`), the 826
context-cycle markers in the city store are 515 self_exact + 311 self_renamed + **0 cross_agent**.

## Delivery was never affected

`matchesRecipientRoute` (`beadmail.go:1157`) is exact string equality against
`session.RecipientRoutesFromInfo` (`internal/session/address_directory.go:208`), whose route set
contains the session's own ID, alias, session name and alias history. A marker assigned to
`gm-wisp-gpenji` can only match that session's own route set. There is no mechanism by which these
land in another agent's mailbox.

The cost was interpretive, not operational — but it is a real cost: any consumer comparing sender
against recipient reads 311 self-addressed markers as agent-to-agent mail, which is exactly the
false P2 this PR is filed under.

## The fix

Carry the resolved sender address to the recipient when the intent names the same address on both
sides. Four lines. It changes how the row reads, never where it lands.

## Tests

- `TestSendHandoffSelfAddressedPersistsOneAddress` — RED before the change with the exact production
  strings, GREEN after. This is the reproduction.
- `TestSelfHandoffFromPooledSessionStaysInItsOwnInbox` — **passes before and after.** It asserts the
  marker reaches its own session's inbox under both renderings of its address and is absent from
  another agent's inbox. Green on unmodified code is the executable disproof of the reported leak;
  green after the change is the regression guard that this PR did not move delivery.

## Out of scope

`Provider.Send` has the same from-resolved/to-verbatim shape, but for ordinary agent-to-agent mail
the two addresses genuinely differ, so it produces no artifact. Only a self-send from an aliasless
pooled session could hit it. Left alone deliberately.

Refs: ga-28neu4

### DIFF
--- a/internal/mail/beadmail/beadmail.go
+++ b/internal/mail/beadmail/beadmail.go
@@ -194,7 +194,21 @@ func (p *Provider) SendHandoff(intent mail.HandoffIntent) (mail.Message, error) 
 	labels = append(labels, "thread:"+intent.ThreadID)
 	labels = append(labels, intent.ExtraLabels...)
 
-	b, err := p.createMessageBead(intent.Subject, intent.Body, from, intent.To, labels, metadata)
+	// A self-handoff (gc handoff, and the gc handoff --auto that PreCompact
+	// fires) names one session on both sides. resolveSenderRoute can re-render
+	// that address -- an aliasless pool member addresses itself by its raw
+	// session ID and resolves to its session_name -- so carry the resolved form
+	// to the recipient too. Left unequal, the stored row shows two different
+	// strings for one session and every consumer that compares sender against
+	// recipient reads a self-addressed marker as agent-to-agent mail (ga-28neu4).
+	// Delivery is unaffected either way: both renderings are in the session's
+	// own RecipientRoutesFromInfo route set.
+	to := intent.To
+	if from != "" && strings.TrimSpace(to) == strings.TrimSpace(intent.From) {
+		to = from
+	}
+
+	b, err := p.createMessageBead(intent.Subject, intent.Body, from, to, labels, metadata)
 	if err != nil {
 		return mail.Message{}, fmt.Errorf("beadmail handoff: %w", err)
 	}
--- a/internal/mail/beadmail/handoff_self_address_test.go
+++ b/internal/mail/beadmail/handoff_self_address_test.go
@@ -1,0 +1,145 @@
+package beadmail
+
+import (
+	"testing"
+
+	"github.com/gastownhall/gascity/internal/beads"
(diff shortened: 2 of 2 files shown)


quad341:
@quad341 — this is the good kind of bug-report follow-up. You were handed "148 false mail notifications a day," and instead of patching toward the complaint you went and counted: 0 of 826 markers all-time were addressed to another agent. Then you fixed the thing that made the store *read* that way, which is the actual defect.

The test pairing is what makes this easy to merge, and I checked it rather than taking it on faith. I reverted your four lines locally and ran both: `TestSendHandoffSelfAddressedPersistsOneAddress` fails with the exact production strings (`From = "builder-gm-wisp-gpenji", Assignee = "gm-wisp-gpenji"`), and `TestSelfHandoffFromPooledSessionStaysInItsOwnInbox` passes on unmodified code. A green test as the executable disproof of the reported leak, then reused as the guard that your fix didn't move delivery — more PRs should do that. Leaving `Provider.Send` alone deliberately, and saying why, was also right.

Merging as-is. One thing I noted for later, not for you to change: the assignee is now a display string rather than the immutable session ID, so it inherits that string's stability. I'll file a follow-up about carrying the recipient session ID in the to-side metadata `Reply` already writes. Thanks for the careful root-cause work.


---

Issue #5348, opened 2026-08-17
bug(managed-city): bootstrap uses ambient bd and can create a schema newer than the native reader


aytheo-bit:
## Summary

At `ff9123c`, a managed-city bootstrap can initialize the canonical server store with the ambient `bd` executable while Gas City's native store reader is compiled against a different, older Beads module. With ambient `bd 1.2.1`, bootstrap created schema v65; the linked native reader knew migrations only through v59. Gas City then selected the newly created managed store as canonical but could not open it natively, reporting:

```text
native_store_eligible=false
preflight_gate=native_open
schema version mismatch: database is at v65, binary knows up to v59
```

In the observed cutover, a pre-existing embedded database at schema v59 with 829 issues remained intact but inactive after the controller converged the scope onto the newly created managed-city database at schema v65. The v65 store contained only the bootstrap issue. No record loss was observed because the embedded source was retained and the cutover was rolled back.

## Affected version

- Commit: [`ff9123c4ae5cf927365ab46010bb28ad81f93f8e`](https://github.com/gastownhall/gascity/commit/ff9123c4ae5cf927365ab46010bb28ad81f93f8e)
- Platform observed: macOS arm64
- Ambient bootstrap CLI: `bd 1.2.1`
- Linked Beads module: `v1.1.1-0.20260805093327-bf97b73749ac`
- Linked migration ceiling: v59

## Expected

Before changing the canonical endpoint, Gas City should either:

1. prove that the managed bootstrap executable and linked native reader share a compatible schema ceiling;
2. reject bootstrap before creating/selecting a store the native reader cannot open; or
3. adopt/migrate an existing compatible store through an explicit parity-checked operation.

After startup, the canonical store should be the same store whose database identity, schema, and issue count were accepted before startup.

## Actual

The lifecycle wrapper calls plain `bd`, so `PATH` selects the bootstrap version. The native reader uses the Beads library linked into `gc`. These independently selected versions can disagree. Startup succeeds far enough to create/select the managed database, then native open fails and status falls back with `native_store_eligible=false`.

## Minimal disposable reproduction

Do not run this against an existing city or production store.

Preconditions:

- Build `gc` from the affected commit with native Beads support.
- Put `bd 1.2.1` first on `PATH`.
- Set `$BD_V59` to a separate `bd 1.1.0`-compatible executable.
- Use a final supported Dolt build.

```sh
export REPRO_ROOT="$(mktemp -d)"
[...]

Issue #1223, opened 2026-04-25
Add [workspace.env] for city-wide env vars inherited by every session


boylec:
## Problem

There is no city-level / workspace-level place to declare environment variables that should be inherited by every managed session. Today, env vars must be declared per-agent, per-provider, or per-`[[rigs.patches]]` block. For cross-cutting variables this leads to repetition.

Example from a real city.toml — the same value declared five times:

```toml
[[rigs.patches]]
agent = "polecat-sonnet"
[rigs.patches.env]
GC_TARGET_BRANCH = "boylec/develop"

[[rigs.patches]]
agent = "polecat-opus-high"
[rigs.patches.env]
GC_TARGET_BRANCH = "boylec/develop"

# ... three more identical blocks
```

Every time a new agent class lands, the operator must remember to add another patch block.

## Note: GC_* passthrough exists but is hidden

`passthroughEnv()` (`cmd/gc/cmd_start.go:988-996`) already passes any \`GC_*\` env var from the supervisor's shell into every spawned session. So today's workaround is `export GC_TARGET_BRANCH=…` in the shell that runs `gc`. But that lives outside the city config — not source-controlled, not visible to teammates, easy to forget when handing off.

## Proposal

Add a `Workspace.Env map[string]string \`toml:\"env,omitempty\"\`` field. Surface it in city.toml as:

```toml
[workspace.env]
GC_TARGET_BRANCH = "boylec/develop"
```

Merge it after passthrough but before provider/agent env, preserving the existing "more specific wins" precedence:

```
shell GC_* (passthrough) → workspace.env → provider.env → agent.env → patch.env → computed session vars
```

The single merge call site is `cmd/gc/template_resolve.go:337`.

## Out of scope

Pack-level env and rig-level env. Workspace.env covers the dogfooded use case.

PR #3457, opened 2026-06-13
Bound managed prompt hook commands

julianknutsen:
## Summary
- add gc hook run wrapper with a hard timeout for managed hook child commands
- route managed prompt-submit nudge/mail hooks through the wrapper across provider templates
- upgrade stale managed Claude/Codex prompt-hook commands and update CLI docs

## Tests
- go test ./cmd/gc -run 'TestGeneratedCLIReferenceCoversCommandTree|TestGeneratedCLIReferenceHasNoStaleSections|TestHookRunTimesOutAndFailsOpenWhenConfigured|TestHookRunPreservesChildExitCodeAndOutput'
- go test ./internal/hooks -run 'TestInstallClaude|TestInstallCodexUpgradesGeneratedFileMissingHookFormat|TestInstallCodexUpgradesManagedFileMissingPreCompact|TestInstallOverlayManagedProviders|TestInstallAntigravityMergesExistingHooks'
- go test ./cmd/gc ./internal/hooks
- git diff --check

PR #5421, opened 2026-08-19
feat(worker): promote Cursor to tier-1 conformance

julianknutsen:
## Summary

- Promote `cursor/tmux-cli` to a tier-1 worker harness/provider across phase 1, phase 2, structured, runtime/tmux, and live conformance matrices.
- Add Cursor fixtures, transcript normalization, profile/setup/auth wiring, hooks, CI/report rollups, and worker-inference acceptance coverage.
- Pin the workspace Beads binary for live runs and skip Cursor's unsupported pre-submit Escape path in tmux.

Closes ga-8dc3y.

## Validation

- `make test-worker-core PROFILE=cursor/tmux-cli`
- `make test-worker-core-phase2-all PROFILE=cursor/tmux-cli`
- Credentialed Cursor live inference with `/tmp/.crsr`: 11 applicable requirements passed across all live suites. The interrupt-recovery requirement is explicitly `unsupported` because Cursor CLI does not cancel an in-flight turn.
- Focused auth classification rerun passes with inherited host Claude auth variables cleared.
- Pre-commit formatting, lint, generation, and `go vet ./...` passed.

The live reports were generated under `/tmp/gascity-cursor-worker-report`.

---
- Related to #5722 — adds a current-era `.beads/.local_version` witness that the managed bd lifecycle script (`gc-beads-bd.sh`) seeds before `bd init --server` runs — the exact missing witness that makes bd's legacy-upgrade guard treat a freshly gc-scaffolded `.beads/dolt` root as a pre-v1 legacy workspace; it seeds only when `.beads` already exists, since with no `.beads/dolt` root the guard does not fire
- Related to #1223 — extends workspace.env inheritance to the direct CLI session-create and resume paths, which previously merged only provider process passthrough and provider env, and locks the passthrough → workspace → provider precedence in a test; the workspace.env field and its resolveTemplate merge already existed on main, so this does not implement the request in that issue itself
- Related to #5348 — pins the managed bd executable end to end — the workspace pin now applies to the city scope and to rigs inheriting its backend, BD_BIN is carried into managed lifecycle and provider-init subprocesses, and the "pinned" runner in gc-beads-bd.sh now actually executes ${BD_BIN:-bd} instead of plain bd; it does not add the schema-compatibility proof or fail-closed preflight that would refuse bootstrap when the selected bd is newer than the linked native reader, so that half of the filed issue is still open


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277200569

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277228551

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277246634

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277292265

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR promotes `cursor/tmux-cli` from a best-effort harness to a tier-1 worker provider across the phase-1, phase-2, structured, runtime/tmux, and live conformance matrices, and brings the fixtures, transcript normalization, profile/setup/auth wiring, hooks, and CI report rollups needed to hold it there. That matters because tier-1 conformance is what lets Gas City treat a provider as a real dispatch target rather than an experiment, and the supporting work — pinning the workspace Beads binary end to end, and skipping Cursor's unsupported pre-submit Escape path in tmux — fixes concrete breakage any Cursor-backed worker would otherwise hit.

### Original PR Review
Decision: request_changes on the first pass, resolved to approve.

Specific gaps fixed:
- The first pass blocked on `seed_bd_current_era_witness` in `examples/bd/assets/scripts/gc-beads-bd.sh`. It seeded a current-era `.beads/.local_version` witness unconditionally — including for databases gc did not create — which forges bd's cross-era migration witness for an adopted or pre-existing workspace and can make a genuinely legacy database look already-migrated. The maintainer fixup deletes that seeder outright rather than gating it. The correctly-gated sibling `seed_fresh_managed_bd_version_witness` survives untouched: still guarded by `database_created_by_gc`, still refusing to replace an on-disk marker, still `die`-ing on every failure path, still writing under `umask 077`. `TestGcBeadsBdInitDoesNotSeedVersionWitnessForAdoptedPreExistingDatabase` went red → green, and the script's net effect on witness behavior is now byte-identical to `main` apart from the intended `BD_BIN` pin.
- The same seeder fabricated a `1.0.0` version when `bd version` was unreadable, swallowed write and rename failures, and wrote its marker without `umask 077`. All three lived inside the deleted function, so no surviving path fabricates a version or writes a witness for a database gc did not create.
- The `BD_BIN` pin missed the one invocation the commit is named after — the `bd version` probe still ran a bare `bd`. It now runs `"${BD_BIN:-bd}" version`, so every executed `bd` in the script honours the pin; the only remaining textual matches are `die` messages and comments.
- `cmd/gc` treats an ambient `BD_BIN` as a hard error while `internal/beads` silently ignores it. The divergence is deliberate but was undocumented; it is now explained at the point of definition in `cmd/gc/bd_env.go`, and two comments in `bd_env.go` and `cmd_bd.go` that still described the pre-PR pin contract were corrected to match the code.
- `cursorHookLegacyVariants` had widened the set of on-disk `.cursor/hooks.json` documents that `Install` will silently overwrite: two of the three reconstructed "legacy" shapes match a document this project never released. The list was narrowed 3 → 1, and the narrowing landed with `TestInstallCursorHookPreservesUnreleasedIntermediateShapes`, which asserts the dropped shapes come back byte-identical rather than being rewritten. The two tests that previously asserted those shapes *were* upgraded were inverted rather than deleted, so no coverage was lost.

Review findings addressed:
- The blocker and all five supporting findings above were resolved by maintainer fixup `5b6706a3` (5 files, +95/−137 — a net deletion). The second review pass re-verified each one independently, re-running the blocker's regression test in isolation and re-deriving the legacy-variant reconstruction against every historical revision of the embedded document rather than taking the fixup's commit message at face value.

- ga-lsrcj: (maintainability, minor) the narrowing's own doc comment and a test failure message state that `cursorHookLegacyVariants` regenerates *every* released `.cursor/hooks.json`. Four distinct released revisions are reachable from the base and only the most recent (`1b123157ba`, #3457) is reconstructed. This is not a regression and it fails safe — before this PR there was no cursor case in `overlayManagedNeedsUpgrade` at all, and an unmatched document falls through and is left alone — but the wording should say "most recent released shape; older documents intentionally left un-upgraded" so a future maintainer reads adding one as a deliberate widening rather than a bug fix.
- ga-lsrcj: (performance / maintainability, minor) `workspacePinnedBdBinary` loads `city.toml` twice per call on the no-pin path, purely to re-answer one bit its callee already computed and discarded. The honest complaint is redundancy on the `gc bd` path, not a measured slowdown — both loads use the lighter no-refresh loader, so the net direction versus the base is unproven.
- ga-lsrcj: (change impact, minor) `SessionEnv` and `Hints.Env` alias the same map on the session-create path. Nothing mutates it today so nothing observes it, but a future hint writing into `Hints.Env` would silently mutate `SessionEnv`. Pre-existing, not introduced here.

### Maintainer Changes
Maintainer fixes were pushed directly to this PR branch; contributor authorship is preserved on every original commit.
- `5b6706a3` fix(bd): drop the unguarded current-era witness seeder from gc-beads-bd — 5 files changed, 95 insertions(+), 137 deletions(-)
- `2ddbd8c4` and `aed9e241` merge `main` into the adopted head (merge mode — the branch could not be linearized cleanly). The single content conflict, in `internal/beads/bdstore.go` `execCommandRunner`, was resolved by keeping this PR's stronger `effectiveEnvValue(baseEnv, env, BD_BIN)` pin while preserving main's `baseEnvFn` injection, so `ExecCommandRunnerWithExactEnvContext` keeps its nil-base exact-env contract. Both sides stay guarded by `TestExecCommandRunnerWithExactEnvContextExcludesParentValues` and `TestExecCommandRunnerWithEnv_InheritedAbsoluteBDBin`.

### Final Review Status
Ready for the merge queue: final head `aed9e24` has passing required GitHub checks — `Check`, `Analyze (actions)`, `Analyze (go)`, `Analyze (javascript-typescript)`, `Analyze (python)`, and `CI / required` are all green, with no failing or pending checks anywhere on the head — and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/33756857857

### Review Iterations
2 review passes. Pass 1 returned `block` on the witness-seeder finding plus five supporting findings; one maintainer fixup landed; pass 2 validated every fix, confirmed the deleted mechanism is absent from the net diff rather than merely gated, and returned `approve` with two minor, non-gating findings, both carried forward to `ga-lsrcj`.

---

PR #5422, opened 2026-08-20
Fix push ownership checks for reused builder branches

quad341:
## What this changes

Push ownership checks now recognize when a long-lived builder branch has moved from a closed predecessor bead to an active successor. If the successor explicitly identifies the same branch or predecessor, the guard validates the push against that live work item instead of permanently rejecting the branch as stale.

The guard remains fail-closed: it only substitutes the successor after a fresh read confirms the predecessor is inactive, and unreadable or malformed bead state still blocks the push.

## Review notes

- The change is confined to branch-to-bead resolution in `scripts/push-ownership-guard.sh` and its shell regression suite.
- Successor selection requires an explicit `metadata.branch` or `metadata.build_bead` relationship; unrelated concurrent assignments are ignored.
- Existing `deploy/*-gate` behavior and the final ownership checks are unchanged.

## Test plan

- [x] Run all 35 push-ownership-guard shell cases, including the five new branch-reuse cases.
- [x] Run shellcheck plus Bash and Zsh syntax validation.
- [x] Run the full local CI-equivalent union with rootless Podman and verify every diff-owned test passes; attributed host/infrastructure failures are documented in the gate.
- [x] Run the repository policy lane and `go vet ./...`.
- [x] Release gate: [`release-gates/ga-bf39j8-push-ownership-branch-reuse-gate.md`](release-gates/ga-bf39j8-push-ownership-branch-reuse-gate.md)

### DIFF
--- a/release-gates/ga-bf39j8-push-ownership-branch-reuse-gate.md
+++ b/release-gates/ga-bf39j8-push-ownership-branch-reuse-gate.md
@@ -1,0 +1,45 @@
+# Release gate: push-ownership branch reuse (`ga-bf39j8`)
+
+- Gate result: **PASS**
+- Reviewed commit: `813d27f5cb9727a8b84a7589b7ee95c59abe3c29`
+- Gated commit after bounded self-rebase: `323ae11493cfb19da07e820f1097e4e4a20f956f`
+- Base ref: `origin/main@8c73625b974ce8d3d68d54ab42062e6247c47036`
+- Deploy mode: remote
+- Evaluation date: 2026-08-19
+- Full-suite logs: `/var/tmp/gc-local-tests.U2ksqv`
+
+`docs/PROJECT_MANIFEST.md` is not present at the evaluated ref. This gate uses
+the canonical criteria in `mol-deployer-gate.formula.toml`, the current
+deployer prompt, and
+`engdocs/contributors/release-gate-criteria-conventions.md`.
+
+## Checklist
+
+| # | Criterion | Result | Evidence |
+|---|---|---|---|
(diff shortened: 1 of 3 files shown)


quad341:
Merging this rather than authorizing a third `--no-verify` bypass, which is what it was about to cost.

A builder has ga-ls3owy's round-2 fix ready — the bounded nudge-queue flock change, which is a live fleet-wide P1: dispatch slings measured today at 8m47s / 10m26s / 12m41s against 25-30s when the queue is quiet, with 18 processes queued on one inode and one waiter starved for 46 hours. Their push is blocked by exactly the defect this PR fixes: the branch name resolves to the closed build bead (`ga-2kzci3`) instead of the open review bead (`ga-ls3owy`) continuing the same work on the same branch. They had already used both authorized bypasses and checked in rather than self-authorizing a third, which was the right call and is why this got fixed instead of routed around.

I read the diff rather than merging on the gate result. The reason it's safe is that the exception **fails closed**:

```
_pog_branch_id_bead_inactive <id>: true (rc 0) only if a FRESH bd show
confirms <id>'s status is neither in_progress nor open ... any read/parse
failure returns 1 (treat as still-active / not confirmed inactive), so an
unreachable bd never triggers the branch-reuse override
```

An unreachable `bd` doesn't widen the exception, it falls through to the existing branch-id-wins path that blocks. That's the correct default for a guard, and it's the difference between a narrow exception and a hole.

All 35 cases in `scripts/test-push-ownership-guard.sh` pass including the five new `branch_reused_*` ones, and the gate record attributes every unrelated failure to a specific tracker rather than waving them through.

One thing surfaced alongside this and deliberately not fixed here, so it isn't lost: the guard also reported a session's tracked assignment as an unrelated older `in_progress` claim before falling back to branch-name resolution. That's a second resolution-order question, not this PR's, and it wants its own bead.

PR #5423, opened 2026-08-20
fix(rig): reject rig paths that cannot survive a restart

julianknutsen:
## The problem

A rig registered under a non-persistent path is a trap. Registration succeeds,
the city looks healthy, and the rig content vanishes the next time the machine
or container is replaced — with no remote and no way back. `gc rig add` and the
API rig-create path accepted **any** absolute path, so nothing stopped one being
created and silently primed to die.

## The guard

Built around what actually makes a path survivable, not a prefix denylist — a
`/tmp` string match is defeated by `/var/tmp`, by a tmpfs mounted elsewhere, or
by any container-local path:

1. **Same device as the city root → accept.** Durable by construction: if that
   device were not durable the city would not survive either. This positive rule
   runs **first**, and that ordering is what keeps the check silent for a city
   that deliberately lives on tmpfs (a test fixture, a disposable dev city)
   rather than becoming "reject all tmpfs".
2. **Otherwise ephemeral by construction → refuse.** `statfs` magic in {tmpfs,
   ramfs, overlayfs}. The overlay case is the one a prefix match misses: inside a
   container `/var/tmp` is on the rootfs and dies just the same.
3. **Otherwise → accept with a warning** naming the blind spot. A second PVC, an
   NFS mount, or a second disk may well be durable; refusing would be worse than
   the hole.
4. **Any probe that cannot run → silent.** Non-Linux and unreadable mounts are
   never blocked by a check that could not run.

`--allow-ephemeral` downgrades the refusal to a warning for a deliberately
disposable city. It is CLI-only — a rig registered through the control plane must
live somewhere that survives a controller replacement.

## Where it lives

- **Registration:** `internal/rig` `Provision` step 2.1 — the single choke point
  both routes share (CLI wrapper + controller StateMutator), so the rig-provision
  boundary allowlist is untouched. It runs *before* the `--git-url` clone, so a
  doomed path is refused before content is fetched into it.
- **Already-registered bad paths:** `config.applySiteBindings`, where the
  effective machine-local rig path is first known, riding the existing
  `siteBindingWarnings` → `prov.Warnings` channel. It **only warns** — refusing
  to load a city because one rig path looks doomed would brick a running city
  that a warning could have saved. Only the ephemeral verdict is reported;
  repeating an other-device warning on every config load would train operators to
  ignore the one that matters. Gated on the FS being the real `OSFS`, because the
  probe reads the host and must not answer about paths that exist only inside a
  test fake.
- **Classification:** new leaf package `internal/pathdurability`, so the
  registration guard and the boot audit share it without an upward dependency.
  Syscalls behind build tags and injectable for tests, following the existing
  `dolt_disk_preflight` idiom.

## Proof — both directions

The accept direction uses every city-rooted path shape a healthy hosted city
produces. Those are the regression risk: a guard that refuses one of them is
worse than no guard at all.

Every arm was proven able to fail, with the build confirmed compiling before any
result was read:

| arm | mutation | result |
|---|---|---|
| reject direction | stub `Classify` to pre-guard "accept everything" | reject tests RED, accept tests stayed GREEN |
| accept direction | disable the same-device positive rule | classifier accept tests RED **and** end-to-end CLI accept test RED |
| wiring | make `Provision` swallow the refusal | end-to-end CLI reject test RED |

Each reverted to GREEN.

Coverage is not only against the fake mount table — there is a real-syscall test
(city on disk, rig on `/dev/shm`) and a real-binary transcript:

```
$ gc rig add /dev/shm/doomed-rig
gc rig add: /dev/shm/doomed-rig is on tmpfs, which does not survive a restart of
this machine or container: a rig registered there is lost with no way to recover
it. Register the rig under the city directory (<city>) instead, or pass
--allow-ephemeral to accept the loss
exit=1                              # and nothing was created

$ gc rig add <city>/rigs/good-rig   # control: same device as the city
Adding rig 'good-rig'... Rig added.
exit=0                              # no durability output at all

$ gc rig add /dev/shm/optin-rig --allow-ephemeral
gc rig add: /dev/shm/optin-rig is on tmpfs and will not survive a restart of this
machine or container; --allow-ephemeral was given, so the rig is registered anyway
exit=0
```

## Gates

- `go build` clean on linux/{amd64,arm64,386,arm} + darwin + windows. The
  `int64(st.Type)` conversion is load-bearing on 32-bit (it is `int32` there) and
  carries a `nolint` with that reason.
- `go vet ./...` clean.
- `internal/{config,rig,pathdurability,api}` pass.
- pre-commit hook clean: lint 0 issues, vet, doc regeneration, docsync.

Three pre-push failures are **pre-existing and environmental**, each verified by
running the identical command in a clean worktree at the base commit and getting
byte-identical results: `internal/pidutil` (3 tests) and
`internal/workspacesvc` fail on `ps` enumeration being killed under host fork
exhaustion, and `cmd/gc`'s
`TestResolveTemplateRendersAbstractUpstreamPerHarness/claude` fails on the base
commit too. None is in or near this diff.


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277201567

PR #5424, opened 2026-08-20
fix(formula): prevent duplicate and mis-rooted attachments

quad341:
## What this changes

`gc formula cook --attach` now refuses to create a second live molecule on a target that already has one. The refusal happens before mutation, so repeated or competing attach attempts cannot leave duplicate wrappers or partial work behind.

For a valid v1 attach, the new sub-DAG is rooted at the caller-named target instead of following unrelated `workflow_id` or `molecule_id` metadata already present on that target. This prevents every new step from being silently rooted under a sibling molecule.

## Review notes

- The command reuses the existing `sling.CheckNoMoleculeChildren` guard used by the legacy sling path, including its stale-molecule cleanup behavior.
- `molecule.AttachOptions.SelfRoot` is opt-in and enabled only for this `gc formula cook --attach` path. Its zero value preserves existing chain-walk behavior for grafted workflow steps and every other caller.
- Graph v2 behavior, configuration, APIs, and persisted schema are unchanged.

## Test plan

- [x] Run the duplicate-attachment regression and verify refusal creates no beads.
- [x] Run the self-rooting regression and the existing run-chain regression by name.
- [x] Run affected lint, formatting, policy, boundary, native DoltLite, build, and vet checks.
- [x] Exercise the required fast, process-backed cmd/gc, and integration lanes; diff-owned tests pass, with tracked non-diff infrastructure failures and their standing waiver recorded in the gate.
- [x] Release gate: [`release-gates/ga-2ktqg2-formula-cook-attach-guard-gate.md`](release-gates/ga-2ktqg2-formula-cook-attach-guard-gate.md)

### DIFF
--- a/cmd/gc/cmd_formula.go
+++ b/cmd/gc/cmd_formula.go
@@ -950,15 +950,24 @@ store, copy them into the binding with
 					_, _ = fmt.Fprintf(stderr, "Auto-burned stale molecule %s\n", id)
 				}
 
+				attachBead, err := attachStore.Get(attach)
+				if err != nil {
+					return formulaCommandError(stderr, "gc formula cook: attach", jsonOutput, fmt.Errorf("attach bead %s: %w", attach, err))
+				}
+
 				result, err := molecule.Attach(cmd.Context(), attachStore, recipe, attach, molecule.AttachOptions{
 					Title:          title,
 					Vars:           cookVars,
 					IdempotencyKey: graphRootKey,
-					// attach names its own root: a workflow_id/molecule_id it
-					// carries is a separate attachment (just ruled out above),
-					// not a container it lives inside, so ResolveRunID's
-					// chain-walk must not run here.
-					SelfRoot: true,
+					// CheckNoMoleculeChildren above only rules out a molecule
+					// already ATTACHED TO attach (downward). It says nothing
+					// about whether attach is itself a step INSIDE a live
+					// workflow (upward, via gc.root_bead_id) -- self-root
+					// only when attach carries no gc.root_bead_id of its own,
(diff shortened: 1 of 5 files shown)


quad341:
Holding this one. An architecture review found two problems with the current head (a356ff803f), and the fix is being built on this same branch.

- **False refusal on mid-workflow beads.** `CollectAttachedBeads` reads a bead's `workflow_id` as a sign that something is attached *below* it, the same way it reads `molecule_id`. In production `workflow_id` only ever points *up*: it marks a bead as contained by a live workflow. So a v1 `gc formula cook --attach` on a mid-workflow bead hits an unconditional `ConflictError` before `SelfRoot` is computed. That is the incident class `TestAttachResolvesRootFromRunChainNotOwnID` exists to prevent. The test doesn't catch it because it calls `molecule.Attach` directly and never goes through `cmd_formula.go`.
- **Missing lock on the v1 path.** The graph.v2 arm wraps its check-and-mutate in `sourceworkflow.WithLock`, but the new v1 sequence (`CheckNoMoleculeChildren` → `Get` → `Attach`) has no lock. Two concurrent attaches on one target can both see "clear" and both mutate.

The fix has three parts. The guard stops treating `workflow_id` as a downward attachment. `SelfRoot` applies only when the target has neither `gc.root_bead_id` nor a live `workflow_id`. The v1 arm takes the same lock. Tests will go through the real command wiring. The original goal, ga-er5u7k's self-rooting past a stale bare `molecule_id`, stays. The fix deliberately does not treat the three chain keys the same way, because that would break it again.

Once the fix is pushed here, this gets a fresh review.

PR #5426, opened 2026-08-20
Stop source census tests from scanning sibling worktrees

quad341:
## What this changes

Repository-wide source census tests now skip the top-level `worktrees/` directory used by concurrent Gas City agents. Previously, another agent's checkout could be counted as part of the branch under test, producing false duplicate construction, publication, and composition-root failures.

The change extends the existing directory exclusions at the three module-root walkers that can encounter sibling worktrees. It does not alter production code or narrower scans whose roots cannot contain that directory.

## Review notes

- The change is confined to census helpers in two existing Go test files.
- The exclusion is the exact directory name `worktrees`; it does not broaden into a general hidden-directory or path-pattern skip.
- Other scratch-checkout pollution sources are deliberately outside this change.

## Test plan

- [x] Populate a synthetic sibling `worktrees/ga-*` tree with duplicate census-triggering Go sources and run every affected helper consumer.
- [x] Run all required `cmd/gc` process and integration shards, plus policy and vet lanes.
- [x] Release gate: [`release-gates/ga-3ikmgg-worktrees-census-gate.md`](release-gates/ga-3ikmgg-worktrees-census-gate.md)

### DIFF
--- a/cmd/gc/storage_provider_bundle_boundary_test.go
+++ b/cmd/gc/storage_provider_bundle_boundary_test.go
@@ -411,7 +411,7 @@ func anyGoWorkFile(t *testing.T, root string) bool {
 		}
 		if entry.IsDir() {
 			switch entry.Name() {
-			case ".git", "node_modules", "vendor", "testdata", ".claude":
+			case ".git", "node_modules", "vendor", "testdata", ".claude", "worktrees":
 				return filepath.SkipDir
 			}
 			return nil
@@ -458,7 +458,7 @@ func moduleGoFiles(t *testing.T, root string) []string {
 		}
 		if entry.IsDir() {
 			switch entry.Name() {
-			case ".git", "node_modules", "vendor", "testdata", ".claude":
+			case ".git", "node_modules", "vendor", "testdata", ".claude", "worktrees":
 				return filepath.SkipDir
 			}
 			return nil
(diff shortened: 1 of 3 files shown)


quad341:
@quad341 — thanks for chasing this one to the root. Tracking phantom duplicate-construction failures back to *another agent's checkout being counted as part of the branch under test* is the genuinely annoying part; the fix being three tokens is the reward for having done that properly.

Two things I appreciated. You found all three module-root walkers, not just the one that happened to fail — `anyGoWorkFile` is easy to miss since it hunts `go.work` rather than Go sources, and it was just as exposed. And you left the narrower scans alone: `storageSurfacePackageDirs` roots at `cmd/gc` and `internal/storebinding`, so it can't see a top-level `worktrees/`, and adding the exclusion there would have been noise dressed as thoroughness.

The gate's acceptance run sold it — a real synthetic sibling tree in place while all 13 consumers ran beats asserting the walk skips a directory. I checked your claim that those 13 cover every consumer of the modified helpers, and it holds, including `TestStorageRegistryConstructorHasOneCaller` over in `storage_boot_test.go`.

Merging as-is. One follow-up idea, not a request: `builtinpacks` pairs its name-skip with a check for any subdirectory holding a `.git` *file* — the actual signature of a worktree, whatever it's named. Worth doing across all four walkers if this bites again.


---

PR #5428, opened 2026-08-20
feat(doctor): detect idle pool capacity beside routed work

quad341:
## What this changes

`gc doctor` now reports `pool-idle-routed-work` when an agent pool has both open, unclaimed work routed to the pool and a live idle pool instance that could take it. The check scans the city store and eligible rig stores, and stays quiet when either routed work or idle capacity is absent.

This is an observability change only. It does not claim work, nudge sessions, alter pool sizing, or attempt an automatic repair.

## Review notes

- The check uses targeted `gc.routed_to` metadata queries per pool instead of a full-store scan.
- Asleep and drained instances do not count as available idle capacity; already-claimed work does not count as stranded.
- `CanFix` is false and `Fix` is a no-op, keeping dispatch judgment out of the doctor command.
- No configuration, storage migration, or API shape changes are required.

## Test plan

- [x] Run the full local CI union, including fast, `cmd/gc` process, and integration shards.
- [x] Exercise all nine new doctor scenarios by name and verify the doctor-name golden registration.
- [x] Run workflow policy checks, `go vet ./...`, and changed-file lint.
- [x] Review attributed host/load failures and verify the changed doctor functions are unreachable from same-package failures with focused coverage.
- [x] Release gate: [`release-gates/ga-kq5vjo-pool-idle-routed-work-gate.md`](release-gates/ga-kq5vjo-pool-idle-routed-work-gate.md)

---
- Related to #4114 — adds a read-only gc doctor check that surfaces open, unclaimed routed pool work sitting beside a live idle pool instance, which touches the optional doctor-visibility suggestion in that issue; it does not add zero-claim spawn backoff, consecutive-zero-claim tracking, or any event/warrant emission, so the respawn loop behavior itself is unchanged


quad341:
@quad341 — the framing on this one is what makes it worth running: idle capacity alone isn't a finding, unclaimed routed work alone isn't a finding, and only the two sitting together are. That's the distinction that separates a useful signal from noise, and you got it right in the check *and* in the tests — the `OKWhenNoUnclaimedRoutedWork` case explicitly encodes "min-floor idle is legitimate," which is the thing most people would have gotten wrong. Keeping `CanFix=false` with a no-op `Fix` is also the right call; dispatch judgment doesn't belong in doctor.

Two things I'll pick up on my side rather than send back to you. `mapBdStatus` folds bd's `blocked`/`deferred` into `"open"`, so a blocked routed bead currently reads as pickable — the controller demand path hit the same wall and solved it with a live status-filtered read (`build_desired_state.go:2034`), and I'll pull the check onto that plus a blocked-bead test. I'll also route the session read through `cliSessionStore` so a `[beads.classes.sessions]` relocation can't quietly turn the check into a no-op.

Both are integration work, not review feedback. Merging with your authorship intact.


---

Issue #1053, opened 2026-04-21
bug: gc sling --formula creates duplicate molecules for same source issue


rileywhite:
### Before you continue

- [x] I searched existing issues and did not find a duplicate.
- [x] I read the relevant docs and contributor guidance.

### Gas City version

0.15.2 (commit: abe31163, built: 2026-04-21T00:57:52Z)

### Environment

Pop!_OS 24.04 LTS, x86_64; shell: bash; Linux 6.18.7-76061807-generic

### Reproduction

`gc sling --formula` does not check for existing in-flight molecules before instantiating a new one. Observed trigger: a source bead is migrated between bead stores between two `gc sling` invocations, producing two invocations against the same logical issue with different bead IDs.

Steps observed:

1. Call `gc sling --formula <formula> <bead-id-A>` — molecule 1 is created and routed to pool
2. Source bead is migrated to a different store, producing a new bead ID for the same logical issue
3. Call `gc sling --formula <formula> <bead-id-B>` against the migrated bead
4. Molecule 2 is created without any warning or error
5. Both molecules route to the same pool; pool reconciler scales to accommodate both
6. Two workers execute identical pipelines in parallel

`gc sling --dry-run` shows no dedup check in output — the flag reports what would be created without verifying uniqueness against existing in-flight molecules.

### Expected behavior

`gc sling --formula` should detect an existing in-flight molecule for the same logical source issue before creating a new one. It should refuse or require an explicit `--force` flag to override.

Potential dedup keys:
- `gh_issue` + `gh_repo` metadata on the source bead
- `gc.source_bead_id` lineage (requires migration chain awareness)
- Same formula + same `gc.run_target` + overlapping source metadata

### Actual behavior

Two molecules are created for the same upstream issue and both begin executing. The pool reconciler correctly scales to meet demand (one slot per molecule), so no scaling error is surfaced — the duplication is silent. Observed consequences:

- ~42 duplicate step beads created across both molecule trees
- Two workers consumed executing identical work in parallel
- `gc convoy delete-source` cannot locate the molecule lacking `gc.source_bead_id` (the one created from the pre-migration bead ID), leaving it orphaned

### Logs, screenshots, or traces

```
# gc sling --dry-run produces no dedup-related output
# Pool reconciler: pool_desired: 2 (one per molecule, correct for demand — silent on duplication)
# gc convoy delete-source: "molecule not found" for the cross-store molecule (no source_bead_id link)
```

### Additional context

[...]

PR #5432, opened 2026-08-20
fix(sling): --on dedup guard now covers convoy-first formulas-v2 launches (#5420)

### LINKED ISSUE #5420: gc sling --on: the duplicate pre-check inspects v1 molecule/wisp children, so it passes vacuously for every formulas-v2 workflow
### Before you continue

### Gas City version

gc 1.4.1 (Homebrew), bd 1.2.2

### Environment

Linux (Raspberry Pi OS, aarch64); shell: zsh; managed Dolt; graph.v2 formulas (`[requires] formula_compiler = ">=2.0.0"`)

### Reproduction

Deterministic — not a race. Fails identically every time.

1. Sling a v2 formula at a work bead:
   `gc sling monorepo/coder rh-i8t --on rehashly-bugfix`
2. Confirm the workflow root is live — `gc bd show rh-xkom` reports
   `gc.kind: workflow`, status `in_progress`
3. Sling the **same formula at the same bead** again, with `--dry-run`:

```
$ gc sling monorepo/coder rh-i8t --on rehashly-bugfix --dry-run
...
Pre-check: rh-i8t has no existing molecule/wisp children ✓
```

4. Drop `--dry-run`. A second workflow launches.

Versions: `gc` 1.4.1 (Homebrew), `bd` 1.2.2. Linux (aarch64), managed Dolt,
graph.v2 formulas (`[requires] formula_compiler = ">=2.0.0"`).

### Expected behavior

The pre-check should find the live workflow and refuse, requiring `--force` to override.

### Actual behavior

The pre-check reports the bead has no children and the sling proceeds, so one source bead runs two concurrent workflows — each with its own worktree, branch and PR.

Observed twice in one rig, two days apart:

| source bead | roots |
|---|---|
| `rh-a8x` | `rh-a20`, `rh-hxj` |
| `rh-u2o` | `rh-kvo`, `rh-pf0` |

Both put two coder sessions on the same change. In the second case both lines reached a plan and one shipped a PR, so recovery meant a maintainer decision about which line to keep, plus retiring the loser's 24-bead step graph by hand.

```shell

```

### Additional context

The `--on` counterpart of #1053. That issue reports there is **no** dedup check on the `--formula` path; on `--on` a check now exists, and it inspects a relationship formulas-v2 never creates — so it passes every time.

## Cause

The check looks for **v1 molecule/wisp children** of the target bead. A v2 workflow root is not a child of its target bead — it is linked only through the synthetic input convoy's `tracks` edge:

```
work bead  <--tracks--  "input convoy for <bead>" (gc.synthetic: true)  <--  workflow root (gc.kind: workflow)
```

The v2 path never creates molecule/wisp children, so the predicate is always true and the guard never fires.

## Why the obvious dedup key does not work

`gc.graphv2_root_key` is not sufficient:

- `rh-kvo` / `rh-pf0` — **identical** root keys (same convoy, same vars hash)
- `rh-a20` / `rh-hxj` — differed only in the convoy component

The key is derived from the synthetic input convoy, which is itself minted per launch, so it cannot deduplicate the thing that created it.

## Suggested fix

Dedup on **(formula, target bead)**: resolve the target's live workflows through the input convoy's `tracks` edge and refuse when one is `open` or `in_progress`, with `--force` as the override.

This converges with #1053's third proposed key ("same formula + same `gc.run_target` + overlapping source metadata"), reached independently from the other path.

`gc sling --help` shows no dedup flag today, and `--force` is documented only as "allow formulas v2 workflow replacement".

## Shared root with #1053 and #3986

All three look like faces of one weakness: **the v2 launch path does not record what it was launched for.**

- #1053 (comment) — `internal/sling/sling_core.go:175-193`: `slingFormula()` passes `""` for `sourceBeadID` to `doStartGraphWorkflow`
- #3986 — the v2 `--on` path does not stamp `gc.source_bead_id`, so `closeSourceBeadChain` no-ops and the source bead stays `open` **even on pass**
- this issue — with no recorded source, dedup falls back to a v1-shaped child lookup that v2 never satisfies

They compound: a source bead whose workflow already shipped stays `open` (#3986), which makes it byte-identical to unstarted work, which makes it a candidate for exactly this duplicate launch.

Possibly also related: #4785 (`gc sling` double-mints pool work) — different mechanism (`gc.routed_to` persisting so the bead satisfies the pool `work_query` alongside its wrapper), but its `CheckBeadStateWithOptions` in `internal/sling/sling_attachment.go` may be where this check lives.

rjgeng:
Fixes #5420.

The `--on` duplicate pre-check (`checkNoMoleculeChildren` / `checkLegacySourceWorkflowConflict`) only ever found a live workflow via `molecule_id`/`workflow_id` metadata, a DB child, or `gc.source_bead_id`. A convoy-first formulas-v2 launch stamps none of these — the source is tracked only through its synthetic input convoy — so the guard passed vacuously and a second `--on` at the same bead started a duplicate concurrent workflow. Deterministic, not a race; reported twice in the field two days apart, and independently converges with #1053's proposed fix shape.

**Fix:** new `LiveConvoyTrackedWorkflowRoots` (`internal/sling/sling_attachment.go`) resolves the convoy(s) tracking the target bead (`convoycore.TrackingConvoysForItem`, the same helper `hasLiveTrackingConvoy` already uses) and finds any live root launched from one of them via `gc.input_convoy_id`, scoped to the formula being attached so distinct formulas can still run concurrently against one bead. Wired into `checkLegacySourceWorkflowConflict`, the actual gate `attachFormulaToBead`'s graph branch enforces. Also fixes `gc sling --on --dry-run`'s "Pre-check ✓" line, which reused the same vacuous check. 4 files, +293/-24.

**Testing:** `go build`/`gofmt` clean, all `internal/sling/...` tests pass, full `cmd/gc` suite run (25m) — one failure (`TestRunStartDriftCheck_DelegatedTryRestartTimeoutThenReplacementSucceeds`) confirmed pre-existing/unrelated via `git stash` comparison against unmodified `origin/main` in this environment.

---


rjgeng:
Heads-up on the red CI here: `cmd/gc process / shard 1 of 12` is failing on `TestCmdStopSupervisorManagedInvalidCityTomlFailsWhenShutdownFails`, which is unrelated to this diff. Confirmed it fails identically and deterministically on plain `main` at the current tip in a clean worktree with none of this PR's changes applied — filed as #5668. Current head here is `fe6f343d`.

quad341 (approved):
@rjgeng — this is genuinely good debugging. "`attachFormulaToBead` passes an empty `sourceBeadID` on the convoy-first path, so `gc.source_bead_id` is structurally never there for the guard to find" is the entire bug in one sentence, and it's the kind of thing that's easy to misfile as a race. I traced it against `NormalizeInputConvoy` and `CreateSingleItemInputConvoy` and it holds up exactly as you described.

Two things stood out. Scoping the lookup to the launch's *own* synthetic convoy, so a convoy-level launch can't block a later per-member `--on` — that's a false positive you went looking for rather than waited to be told about. And renaming `TestSlingAttachGraphFormulaCreatesFreshRootForBareBeadTarget` while saying plainly in the comment that the old assertion was pinning the bug. That's the honest move.

The inconclusive-lookup path exiting 0 while withholding the `✓` is also right: a preview shouldn't hard-fail on a read error, and it shouldn't claim a pass it never got.

Merging as-is. Two follow-ups on our side, both out of scope here: making `--force` supersede rather than duplicate (which you already flagged), and teaching `gc workflow delete-source` to resolve convoy-tracked roots, since its `gc.source_bead_id` query can't see the ones this now blocks on. Thanks also for chasing the unrelated CI failure down separately.


---

Issue #4602, opened 2026-07-24
bug: builtin:claude emits --permission-mode auto-edit, rejected by Claude Code >= 2.1.170


pcresswell:
## Summary

`gc`'s `builtin:claude` provider maps the config value
`permission_mode = "auto-edit"` to the CLI flag `--permission-mode auto-edit`.
Claude Code renamed that flag value to `acceptEdits` (and now **rejects**
`auto-edit`), so **every gc session launched with the claude provider dies at
launch** on a current Claude Code.

## Exact versions

- **gc:** 1.3.1 (confirmed still present in 1.3.5, the latest release, and in v1.4.0-rc1 / main as of 2026-07-24)
- **Claude Code:** 2.1.170

## Error (observed at session launch)

```
error: option '--permission-mode <mode>' argument 'auto-edit' is invalid.
Allowed choices are acceptEdits, auto, bypassPermissions, default, dontAsk, plan.
```

## Root cause (confirmed in gc source, v1.3.1, still present in v1.3.5 / v1.4.0-rc1)

- `internal/worker/builtin/profiles.go` (~lines 90–127): the `builtin:claude`
  profile maps `permission_mode = "auto-edit"` to `--permission-mode auto-edit`.
- The flag is emitted by `ResolvedProvider.ResolveDefaultArgs()`
  (`internal/config/provider.go:367`), which reads
  `OptionsSchema[].Choices[].FlagArgs` (NOT the `PermissionModes` map).
- **The `grok` profile in the same file (~line 305) already maps `auto-edit` ->
  `acceptEdits`.** So the claude profile is simply out of step with grok.

## Proposed fix

In `builtin:claude`, map the `auto-edit` permission-mode choice to
`--permission-mode acceptEdits` (matching the `grok` profile and current Claude
Code). I.e. change the claude profile's `auto-edit` choice `flag_args` from
`["--permission-mode", "auto-edit"]` to `["--permission-mode", "acceptEdits"]`.

Optionally also accept `acceptEdits` directly as a config value so the rename is
visible to operators.

## Local workaround we ship (until upstream lands)

A city-config provider override (NOT a binary change, NOT a PATH shim) in
`city.toml` that redefines the `permission_mode` option by key so the `auto-edit`
choice emits `acceptEdits`:

```toml
[providers.claude]
base = "builtin:claude"
options_schema_merge = "by_key"

[[providers.claude.options_schema]]
key = "permission_mode"
label = "Permission Mode"
type = "select"
[...]

PR #5433, opened 2026-08-20
fix(worker): codex auto-edit uses --sandbox workspace-write, not removed --full-auto (#5415)

### LINKED ISSUE #5415: bug/docs: tutorial steps fail and codex --full-auto deprecated
### Before you continue

### Gas City version

1.4.1

### Environment

ubuntu 24.04 (WSL)

### Reproduction

1. Follow the tutorial https://docs.gascity.com/tutorials/02-agents to set up the codex provider
2. The permissions setup gives `option_defaults = { model = "o4-mini", permission_mode = "suggest" }`
3. Using `suggest` fails the tutorial task as it can't write to `review.md` or to Beads
4. Change to `auto-edit` to work around this
5. codex session fails in a loop as `--full-auto` is deprecated.

### Expected behavior

1. Can follow the tutorial with no modifications.
2. Can use `auto-edit` for codex with valid options.

### Actual behavior

Using `suggest` fails the tutorial task as it can't write to `review.md` or to Beads

Using `auto-edit` codex session fails in a loop as `--full-auto` is deprecated.

```shell

```

### Additional context

_No response_

rjgeng:
Fixes #5415.

codex CLI deprecated `--full-auto` in v0.128 and removed it entirely in v0.147.0 (2026-08-07); `codex exec --full-auto` now errors instead of running, breaking the codex `auto-edit` permission mode.

**Fix:** map the codex `auto-edit` permission mode to the documented replacement, `--sandbox workspace-write --ask-for-approval never`, mirroring the existing `--ask-for-approval` usage on the neighboring `suggest` entry. Pinned with a new test (`TestCodexAutoEditMapsToUnremovedCLIFlags`), since no existing test covered this mapping.

Also bundles a docs-clarity fix: the agents tutorial (`docs/tutorials/02-agents.md`) previously showed `permission_mode = "suggest"` (interactive-approval-only, by design) as the sole illustrative example right before a fire-and-forget `gc sling` walkthrough — a context where it can only fail. Added a `<Note>` following the file's existing callout pattern; no code change needed there, `suggest` behaves as designed.

**Testing:** `go build`/`gofmt` clean, `internal/worker/builtin/...` tests pass (new test pins the corrected codex flag mapping), `make check-docs` clean.

---


quad341 (approved):
@rjgeng — thank you for this. `--full-auto` going from deprecated to *removed* is exactly the kind of change that silently rots a provider mapping until someone's slung work just dies with no useful error, and you found the codex twin of the claude bug we fixed in #4602.

Two things I especially appreciated. You pinned it with a test instead of just editing the map, and you asserted *both* halves — `PermissionModes` and the `OptionsSchema` choice's `FlagArgs`. Updating only one is how this class of fix half-lands; your test makes that impossible to repeat. And you matched the neighboring `suggest` entry's flag style rather than inventing a new one, so the codex block still reads as one table.

The docs Note is a good catch too — pointing at `permission_mode = "suggest"` right before a fire-and-forget walkthrough is a trap.

I'm taking two small things on my side and merging: tightening the Note's wording (`suggest` does allow trusted reads — it's the *write* of `review.md` that blocks), and softening a test comment, since `--full-auto` was really `-a on-failure`, so `never` is a deliberate tightening. Your call was right for unattended workers. Merging with credit intact.


---

PR #5434, opened 2026-08-20
test(cmd/gc): use shared hang budget for reload polling

quad341:
## Summary

- replace duplicated reload-test polling loops with the shared hangBudget path
- keep the fixed-sleep resource census synchronized after removing two sleeps
- preserve the reviewed reload-test implementation across the current-main rebase

## Release gate

- deploy bead: ga-0ckn7x
- review bead: ga-09qq0u
- evaluated candidate: 8061ea62b587668f0e7f58d1777717fe68bb5c54
- gate record: release-gates/ga-0ckn7x-hangbudget-reload-gate.md
- verdict: PASS — MAYOR WAIVER APPLIED
- waiver: mayor-2026-08-20-ga-0ckn7x-c3

The required 40-job union completed 27 PASS / 13 FAIL jobs. The changed reload tests passed. Three same-package failures remain explicitly preserved as FAIL — WAIVED under the bead- and mechanism-specific mayor ruling; all remaining failures are tracked non-overlapping fleet failures. No retry was used to rewrite the raw result green.

Merge authority remains with mayor/mpr.


quad341:
@quad341 — thanks for this one. Collapsing those two hand-rolled `for reconcileCount.Load() < 1 { select … }` loops onto `awaitCond` is exactly the right move: it puts the reload tests on the same shared hang budget as the rest of `cmd/gc`, so a wedged reconcile now fails with a named message instead of a bare `t.Fatal`, and passing runs don't get any slower. The `deadline =` → `deadline :=` catch on the surviving 45s loop was the easy thing to miss, and you didn't miss it.

I especially appreciate that you kept the fixed-sleep ledger synchronized in all three places instead of leaving the census test to blow up on someone else. I ran it here at your head and it's green, along with `go vet ./cmd/gc/`.

The only thing left is a merge conflict, and it's ours, not yours — main has moved the `fixed_sleep` baselines from 453/302 to 477/321 since you branched, so those exact three rows collide. I'll take the rebase and re-derive them (should land at 475/319, file counts unchanged), then re-run the census test and the reload suite before merging. Your commit and the gate record stay as they are. Nothing needed from you.


---

quad341:
@quad341 — this is ready to go in. Collapsing those two hand-rolled `for reconcileCount.Load() < 1 { select … }` loops onto `awaitCond` puts the reload tests on the same shared hang budget as the rest of `cmd/gc`: same 5ms poll, same immediate return when the condition is already true, but a wedged reconcile now fails at the line that wedged with a named message instead of a bare five-second `t.Fatal`. The `deadline =` → `deadline :=` catch on the surviving 45s loop was the easy thing to miss once the earlier declaration went away.

I re-derived the ledger rather than reading it off the gate record: the −2 `fixed_sleep` delta is consistent across `TESTING.md`, `census.go`, and `test-resources.toml`, and `main` hasn't added a sleep since your gate base, so the ratchet stays exact through the merge.

The one red check is ours, not yours — `Mac / quality` trips revive on an unused parameter in `dolt_data_dir_lock_other.go`, a `!linux` file you never touched whose blob is byte-identical on `main`. I'll send a separate one-liner for that rather than bolt it onto your gate record. Merging now.


---

PR #5435, opened 2026-08-20
fix(herdr): socketPath honors XDG_CONFIG_HOME like herdr itself

quad341:
## Summary

`socketPath()` in `internal/runtime/herdr/client.go` built its path via
`os.UserHomeDir()+".config"`, silently ignoring `XDG_CONFIG_HOME`. herdr
itself resolves its config/socket directory via standard XDG Base
Directory precedence (verified empirically: `XDG_CONFIG_HOME=X herdr
--help` reports `Config: X/herdr/config.toml` regardless of `$HOME`).

In any sandbox that sets `XDG_CONFIG_HOME` independently of `HOME` (this
fleet's agent sandboxes do), the old code computed a socket path no herdr
server ever binds. `serverAlive()` then read `false` against a perfectly
healthy server, `ConfigureServer` treated it as "did not become ready",
and each retry launched a redundant herdr server contending for the same
pane — surfacing as `agent_pane_busy`. This explains the failures
`TestProviderLiveClaudeKindPath` and related live tests were hitting
(ga-nqlb8q, and the recurrence noted on ga-zron27's deploy-gate run).

Fix: `socketPath()` now resolves via `os.UserConfigDir()`, which
implements the same XDG_CONFIG_HOME-first precedence herdr itself uses.

Two regressions this fix uncovered, fixed in the same change:
- `listenHerdrSocket` (the fake-herdr test helper used across
  panebinding/startup-delivery suites) had its own hand-rolled
  reconstruction of the socket path that silently drifted from
  `socketPath()` the instant `socketPath()` started honoring
  `XDG_CONFIG_HOME`. Now calls `p.c.socketPath()` directly so the two
  can never diverge again.
- `newFakeHerdrProvider`'s session-name return value was dead code once
  every caller stopped using it (`unparam`); dropped, 24 call sites
  updated.

`kindpath_live_test.go` was deliberately left untouched — its
`TestProviderLiveClaudeKindPath` flake under concurrent host load
(fixed session/agent names colliding) is a different failure mode,
already tracked and being fixed in ga-zron27.

refs ga-nqlb8q

## Testing

- [x] `make check`
- [ ] `make check-docs` — not applicable, no docs changed
- [ ] `make test-integration` — not applicable; herdr's live-path tests
      are gated by `-short`, not a build tag, so the full (non-short)
      package run below already exercises that coverage

Also directly verified:
- `go vet ./internal/runtime/herdr/...`: clean
- `go test ./internal/runtime/herdr/... -short`: pass (0.7s)
- `go test ./internal/runtime/herdr/...` (full, incl. live herdr+claude
  tests): 80/80 PASS, 0 FAIL, 19.556s — `TestProviderLiveClaudeKindPath`
  included and passing
- `make test-fast-parallel` (via `.githooks/pre-push`, full run): pass

## Checklist

- [x] Linked an issue, or explained why one is not needed — refs ga-nqlb8q
- [x] Added or updated tests for behavior changes —
      `TestSocketPathHonorsXDGConfigHomeOverHome`,
      `TestSocketPathFallsBackToHomeConfigWhenXDGUnset`
- [ ] Updated docs for user-facing changes — not applicable, internal
      runtime bugfix, no user-facing behavior or docs affected
- [x] Called out breaking changes or migration notes — none; this
      restores herdr's intended (already-documented) XDG precedence,
      it does not change any public contract

### DIFF
--- a/internal/runtime/herdr/client.go
+++ b/internal/runtime/herdr/client.go
@@ -603,13 +603,26 @@ func (c *client) ensurePlacement(ctx context.Context, wsLabel, tabLabel, cwd str
 
 // ── shared session-server lifecycle ──────────────────────────────────────────
 
-// socketPath is the unix socket for this client's herdr session.
+// socketPath is the unix socket for this client's herdr session. Must match
+// wherever the herdr binary itself resolves its config/state directory —
+// confirmed empirically (`XDG_CONFIG_HOME=X herdr --help` prints
+// "Config: X/herdr/config.toml" regardless of $HOME; with XDG_CONFIG_HOME
+// unset it falls back to "$HOME/.config/herdr/…") to be standard XDG Base
+// Directory precedence, i.e. exactly os.UserConfigDir() on this platform. A
+// plain os.UserHomeDir()+".config" join (the prior implementation) silently
+// ignores XDG_CONFIG_HOME, so it diverges from herdr's own resolution in any
+// environment that sets XDG_CONFIG_HOME while pointing $HOME elsewhere —
+// this fleet's agent sandboxes do exactly that. The result: this client
+// dials a socket no herdr process ever binds, so serverAlive() reads false
+// against a perfectly healthy server ("did not become ready"), and every
+// retry launches a redundant herdr server contending for the same pane
+// ("agent_pane_busy") — ga-nqlb8q.
 func (c *client) socketPath() string {
-	home, _ := os.UserHomeDir()
+	configDir, _ := os.UserConfigDir()
 	if c.session == "" || c.session == "default" {
-		return filepath.Join(home, ".config", "herdr", "herdr.sock")
+		return filepath.Join(configDir, "herdr", "herdr.sock")
 	}
-	return filepath.Join(home, ".config", "herdr", "sessions", c.session, "herdr.sock")
+	return filepath.Join(configDir, "herdr", "sessions", c.session, "herdr.sock")
 }
 
 // serverAlive reports whether the session-server is actually accepting
--- a/internal/runtime/herdr/client_test.go
+++ b/internal/runtime/herdr/client_test.go
@@ -2,6 +2,7 @@ package herdr
 
 import (
 	"path/filepath"
+	"runtime"
 	"testing"
 )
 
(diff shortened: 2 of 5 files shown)


sjarmak (approved):
Reviewed via the maintainer PR pipeline (Claude + Codex cross-provider pass, blast-radius trace, 29-rule checker audit).

The core fix is correct: `socketPath()` now derives from `os.UserConfigDir()` instead of hardcoding `~/.config`, so it honors `$XDG_CONFIG_HOME` the way herdr's own binary does. Also fixes a test-helper path-reconstruction bug and drops dead code.

One gap found in review: the two new tests (`TestSocketPathHonorsXDGConfigHomeOverHome`, `TestSocketPathFallsBackToHomeConfigWhenXDGUnset`) hardcoded a `.config`-rooted path assertion that only holds on platforms where `os.UserConfigDir()` follows XDG — on Darwin it ignores `$XDG_CONFIG_HOME` entirely and always resolves to `~/Library/Application Support`, so these tests would fail on macOS CI. That gap wasn't caught here because gascity's macOS CI path filter doesn't cover `internal/runtime/herdr/` (separate, unrelated CI issue).

Applied a small follow-up commit (`7a8fb35f4`) that skips both tests on non-Linux platforms rather than asserting an unconfirmed cross-platform contract — herdr's actual macOS socket resolution is unverified from this repo alone. Re-ran the full multi-model review against the fixed revision; clean.

Thanks for tracking down the XDG bug — good catch, and a clean root-cause fix.

sjarmak:
Closing as superseded — this exact fix already landed via #5459 (merged 2026-08-21), so this PR's diff is now fully covered by main. Thank you for the fix and the empirical XDG_CONFIG_HOME verification against the herdr binary; that analysis was correct and useful.

One gap survived into #5459's merge: the two new tests assert a `.config`-rooted path unconditionally, but `os.UserConfigDir()` only follows `XDG_CONFIG_HOME` on Linux — it ignores the env var entirely on darwin (always resolving under `$HOME/Library/Application Support`). That would fail on macOS CI. Filing a small follow-up PR against main now to skip those two tests on non-Linux platforms, crediting this PR's analysis.

PR #5437, opened 2026-08-20
test(herdr): isolate kind-path lifecycle names

quad341:
## Summary

- salt live herdr kind-path test names with PID plus an atomic counter
- prove generated names are unique
- prove the names traverse the fake-provider lifecycle

## Release gate

- deploy bead: ga-zron27
- review bead: ga-hmd2gu
- evaluated candidate: 62ee9f212bc33051f6e1e1a9375e86bee0e1db72
- gate record: release-gates/ga-zron27-herdr-kindpath-names-gate.md
- verdict: PASS — MAYOR WAIVER APPLIED
- waiver: mayor-2026-08-20-ga-zron27-c3

The fast gate completed 9 PASS / 1 FAIL. Diff-owned TestProviderLiveClaudeKindPath remains explicitly preserved as FAIL — WAIVED for the agent_pane_busy / pane-not-ready signature only. The two hermetic diff-owned tests passed, and the same lower-level signature was independently reproduced on the untouched sibling TestProviderLive.

This PR awaits maintainer-pr-review. No merge request has been sent to mayor, and the deployer has not merged.

### DIFF
--- a/internal/runtime/herdr/kindpath_live_test.go
+++ b/internal/runtime/herdr/kindpath_live_test.go
@@ -35,6 +35,54 @@ func TestKindPathNamesAreUnique(t *testing.T) {
 	}
 }
 
+// TestKindPathNamesWorkThroughFakeProviderLifecycle hermetically covers what
+// TestProviderLiveClaudeKindPath cannot in a claude-less sandbox: that
+// kindPathNames()'s PID-salted session/agent names actually thread correctly
+// through the same New/Start/GetMeta/IsRunning/ObserveLiveness call sequence,
+// using the fake-herdr harness from newFakeHerdrProvider
+// (panebinding_provider_test.go) instead of a live herdr server and claude
+// binary.
+//
+// ga-hmd2gu review round 1: TestKindPathNamesAreUnique proves the generated
+// names are distinct, but never threads them through a real provider call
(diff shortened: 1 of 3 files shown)


quad341:
@quad341 — the second test is what makes this land. `TestKindPathNamesWorkThroughFakeProviderLifecycle` closes the exact gap a claude-less sandbox leaves open, and pulling `newFakeHerdrProviderForSession` out of `newFakeHerdrProvider` was the right way to get there — it reuses the existing fake-herdr harness without disturbing either of its current callers.

I re-ran it on my side before merging rather than taking the gate record's word: both new tests pass, the package is green in `-short`, and `gofmt`/`vet` are clean.

Two things I want to call out as good practice. The gate record says "the raw fast result remains 9 PASS / 1 FAIL; this record does not rewrite it green" — that honesty is worth more than a clean-looking table. And the test comments name the bead and the mechanism, which is what makes this readable a year from now.

One follow-up worth a bead, not a blocker: unique session names mean each run can leave its own `~/.config/herdr/sessions/gctest-kind-<pid>-N` behind when teardown doesn't finish. The `gctest-conf-*` dirs already do this, so it's inherited, not yours.

Merging as-is. Nice work.


---

PR #5438, opened 2026-08-20
test(integration): isolate standalone bd HOME while pinning subprocess HOME

quad341:
## Summary

- pin the integration harness HOME to the real passwd-database home for gc/supervisor subprocesses
- re-isolate standalone bd subprocess HOME to the caller-owned test directory
- preserve the empty-home guard and focused regression coverage

## Provenance

- deploy bead: ga-ca0np7
- build bead: ga-9lh4m0
- review bead: ga-20x17h — PASS
- evaluated candidate: c0358b4fadb06fea51301c1ab4a3bb119656d9ba
- gate: release-gates/ga-ca0np7-integration-home-isolation-gate.md

## Test plan

- 5 focused integration HOME tests: PASS, 0 skip
- build, vet, integration-tagged vet, affected lint/format, and diff check: PASS
- 40-job local union: 27 PASS / 13 FAIL jobs; raw failures remain recorded in the gate
- nine exact beads#4566 failures: FAIL — WAIVED under the ga-lpfjhc standing authorization
- TestDoltConfigWiringExternalHost: FAIL — WAIVED only for this deploy under mayor-2026-08-20-ga-ca0np7-c3


quad341:
**Step-1 causality probe (fix-plan gate) — measured 2026-09-05 ~11:20Z, mayor.**

Question under test: does this PR's `pinRealHomeEnv` pin cause the `TestDoltConfigWiringExternalHost` failure that waiver `mayor-2026-08-20-ga-ca0np7-c3` excused? (The review disproved the gate's "no mechanism overlap" claim: `dolt_config_test.go` gets its env from `newIsolatedToolEnv` → `integrationEnvFor` — the changed function — feeding the 15s `runBDInitCompat`.)

Method: scratch worktree at PR head `ad8e7f03`, focused run of `GC_SESSION=subprocess go test -tags integration -run '^TestDoltConfigWiringExternalHost$' ./test/integration/... -count=1 -v -timeout 5m`, then the identical run with `test/integration/integration_test.go` restored to `origin/main` (pin absent). Host: the same machine where the gate's nine waived local failures were originally recorded, with a real `~/.config/beads/credentials` present — i.e. the environment the pin's HOME re-routing actually reaches.

Result:
- WITH pin (PR head): **PASS**, 3.17s (exit 0)
- WITHOUT pin (main harness): **PASS**, 4.39s (exit 0)

Both runs sit far under the 15s `runBDInitCompat` deadline; the reviewer's stop-condition (fails with pin, passes without) did not occur, and neither did the fails-both branch. Conclusion: **the pin is not causal** on the very host where the waived failures were observed, so the waiver's environmental reading stands on measurement — consistent with `TestGraphWorkflowSuccessPath` (one of the nine waived failures) passing in this PR's own CI. The gate's *stated* mechanism-overlap claim is still wrong and should be corrected per fix-plan steps 2–3, citing this comment as the empirical non-blocking basis.

quad341:
Thanks @quad341 — the matched pair is the part I appreciate most here. It would have been easy to pin `HOME` in `integrationEnvFor` and call it done; instead you spotted that `standaloneBDEnvForDir` only ever execs `bd`, which isn't subject to the supervisor's real-HOME guard, and re-isolated it back to the caller-owned dir with a test asserting *both* halves — not just `== dir` but explicitly `!= realHome`. That second assertion is what will actually catch a future regression. I traced `platformSupervisorHomeOverrideError` myself and your comment describes it exactly.

Two things I'll take on my side rather than bouncing back to you. The gate says there's no mechanism overlap with `TestDoltConfigWiringExternalHost` — but that test gets its env from `newIsolatedToolEnv` → `integrationEnvFor`, precisely what you changed, and hands it to the 15s `runBDInitCompat`. I'll correct that wording and re-run it with and without the pin to find out whether it's causal.

Second: your new tests only live in the `rest-full` shard, which is skipped on PRs, so CI never ran them. I'll wire both into `rest_smoke_tests` — env-construction only, so the budget cost is nil and the invariant finally gets real protection.

---

**Two notes on what I did here, since they're the load-bearing part:**

I didn't take the 2-1 split at face value. Qwen and Codex both said "auto-merge, no correctness risks," but neither traced anything outside the diff — their clean bill of health is an absence of investigation, not a finding. Claude's dissent made three checkable claims, so I checked all three in the checkout, and every one held: the gate's waiver rationale is disproven by the call graph, the new tests are invisible to PR CI, and the HOME pin reaches paths `GC_HOME` doesn't cover.

The `Ambiguous: yes` is deliberate and narrow. The gate record justifies a waiver *you* granted (`mayor-2026-08-20-ga-ca0np7-c3`) using a mechanism trace that doesn't hold. Correcting the prose is ours to absorb; deciding whether the waiver itself survives is yours. Step 1 of the fix plan is a hard gate — if the pin turns out to cause that timeout, this stops being a fix-merge and comes back to you.


---


quad341:
@quad341 — the part worth calling out is that you didn't stop at `integrationEnvFor`. Pinning HOME there and shipping would have looked complete while quietly leaking the real home into every standalone `bd` exec. Instead you traced that `standaloneBDEnvForDir` only ever runs the bd binary, re-isolated it to the caller-owned dir, and asserted in both directions — `!= realHome` as well as `== dir`. I checked both call sites myself; the negative assertion is the one that actually holds the line against a future regression.

The causality probe, too. The prior stop-condition was "if the pin causes that timeout, this comes back to you" — you ran the matched pair on the host where the failures were recorded rather than arguing about it, and both variants passed. That's a waiver instead of a guess.

Merging with one two-line correction I'll take on my side: the gate record's criterion-7 row still says `integration_test.go` is the sole changed file, which stopped being true when you added the smoke wiring. Ours to fix, not yours.

The `blocked-on-deploy-gate` label is pipeline, not you. Nothing to revise.


---

<details>

```diff
From b85dd27117e82c343ae0f0db20473dae1715452d Mon Sep 17 00:00:00 2001
From: quad341 <quad341@users.noreply.github.com>
Date: Sat, 5 Sep 2026 05:04:15 -0700
Subject: [PATCH] docs(release-gates): correct criterion 7 file count and note
 per-PR shard coverage

Criterion 7 claimed test/integration/integration_test.go was the sole
changed file; the candidate changes three. Restate the criterion
accurately -- harness change plus tests, shard-manifest wiring, and this
gate record -- while keeping the PASS verdict: the three files are one
theme, not a bundled feature.

Also record in the test evidence that both diff-owned tests are
registered in rest_smoke_tests, so they execute on every PR via
Integration / rest-smoke-{1,2}-of-2 rather than only in the local
focused run.
---
 release-gates/ga-ca0np7-integration-home-isolation-gate.md | 7 ++++++-
 1 file changed, 6 insertions(+), 1 deletion(-)

diff --git a/release-gates/ga-ca0np7-integration-home-isolation-gate.md b/release-gates/ga-ca0np7-integration-home-isolation-gate.md
index 1aece8cda..17e925574 100644
--- a/release-gates/ga-ca0np7-integration-home-isolation-gate.md
+++ b/release-gates/ga-ca0np7-integration-home-isolation-gate.md
@@ -33,7 +33,7 @@ release may proceed; none of these failures is rewritten green.
 | 4 | No high-severity review findings open | **PASS** | `ga-20x17h` reports no style or security findings. One pre-existing low-severity fail-open branch coverage gap is explicitly non-blocking. Unresolved HIGH count: `0`. |
 | 5 | Final branch is clean | **PASS** | Before adding this record, the branch was clean and tracked `origin/deploy/ga-20x17h-gate`. `git diff --check`, affected formatting, `go build ./...`, `go vet ./...`, and integration-tagged vet all passed. Repository hooks resolve to `.githooks`. |
 | 6 | Branch diverges cleanly from main | **PASS** | Reviewed source `254dbabb5b` rebased without conflict onto `origin/main@7c817e0640`, producing `c0358b4fad`. An initial zsh source of the bash helper failed to load its push guard only after completing the clean local rebase and therefore did not push; the correct bash invocation then returned the documented no-op result because current main was already an ancestor. The exact candidate was subsequently pushed under the recorded non-diff fast-failure attribution. |
-| 7 | Single feature theme | **PASS** | The sole changed file, `test/integration/integration_test.go`, contains one integration-harness HOME-isolation theme and its tests. No independent feature is bundled. |
+| 7 | Single feature theme | **PASS** | Three files, one theme: `test/integration/integration_test.go` carries the harness HOME-isolation change and its tests, `scripts/test-integration-shard` wires those two tests into the per-PR `rest-smoke` manifest, and this gate record documents the evaluation. No independent feature is bundled. |
 
 ## Test evidence
 
@@ -59,6 +59,11 @@ Commands and results:
 - `TestIntegrationEnvForPinsRealHome`: PASS.
 - `TestStandaloneBDEnvForDirIsolatesHome`: PASS.
 
+Both diff-owned tests are registered in `rest_smoke_tests`
+(`scripts/test-integration-shard`) and therefore execute on every PR via
+`Integration / rest-smoke-{1,2}-of-2`, not only in the local focused run
+above.
+
 Related unchanged tests executed in the same focused command:
 
 - `TestStandaloneBDEnvAllowsBDAutoStart`: PASS.
-- 
2.55.0
```

</details>

quad341:
Re-reviewed `ad8e7f036`..`d063523b0` (259 new commits since the last review): no new concerns — the prior review stands.

quad341:
@quad341 — the half worth naming is the one you didn't have to do. Pinning `HOME` in `integrationEnvFor` and stopping there would have looked finished, passed your focused run, and quietly handed the real home to every standalone `bd` exec in the suite. You traced that `standaloneBDEnvForDir` only ever runs the bd binary — never `gc start`, so never subject to the supervisor guard — and re-isolated it back to the caller-owned dir. I walked both call sites and the scoping is exactly right, and I checked `pinRealHomeEnv` against `platformSupervisorHomeOverrideError` line by line: same passwd lookup, same two fail-open conditions. It mirrors the guard rather than approximating it.

The `!= realHome` assertion is the one that'll earn its keep — an `== dir` check alone would survive a future refactor that pins HOME once at the top and drops your re-isolation; yours won't.

And on the causality probe: running the matched pair on the host where the failures were recorded, instead of arguing mechanism, is why the gate wording now admits the call-path overlap rather than denying it. That's the honest version.

Nothing to revise. Both tests are wired into `rest-smoke` and running in CI for the first time right now — I'll merge once those two shards report. The `blocked-on-deploy-gate` label is pipeline, not you.


---

PR #5440, opened 2026-08-20
fix(test): reap the supervisor `gc start` respawns in the drift suite (ga-ccltia)

quad341:
Fixes the process leak in ga-ccltia: the drift integration suite spawns supervisors it never reaps.

## Symptom

99 live `gc-drift supervisor run` processes on the fleet host as of 2026-08-20T13:45Z, started Aug 18 onward and never exited: **2.4 GB RSS**, 99 held loopback ports, and 118 `/var/tmp/gc-int-env-*` dirs. Their declared owner (`GC_MANAGED_DOLT_TEST_PARENT_PID`) is long gone and their cwd is deleted, so they are genuinely orphaned.

## Two defects, one kind of debris

They look identical in `ps`. The discriminator is **cwd**, which is fixed at spawn — `os.Chdir` appears nowhere in the production `gc` path (only `internal/docgen/schema.go`), and neither spawn site sets `cmd.Dir` — so cwd is a reliable fingerprint of who spawned each process:

| cwd | count | spawned by |
|---|---|---|
| `<env>/drift-001` | 85 | `spawnDetachedSupervisor`, inheriting `gc start`'s cwd |
| `<env>/gc-home` | 14 | `launchDirectSupervisor` (`cmd.Dir = gcHome`) |

**Defect A (85) — the respawned supervisor was nobody's job.** The drift-restart path under test kills the running supervisor and spawns a *replacement* via `spawnDetachedSupervisor` (`cmd/gc/cmd_start_drift.go`), deliberately detached so `gc start` can return. The scenario cleanup targeted only the launch-time PID, so the replacement's PID was never learned and never reaped — **a fully passing test leaked one supervisor per run.**

**Defect B (14) — cleanup registered too late.** `t.Cleanup(stopDirectSupervisor)` was registered *after* `pollHealthBuildID`, `bootstrapDriftCity` and a second `buildGCBinaryWithCommit`, all `t.Fatalf`-capable. A fatal in that gap skips the registration while the earlier env-root cleanup still runs — which is exactly why those 14 have a deleted `gc-home` and no city dir.

The two buckets moved independently over the last 10 hours, which is the theory's own prediction and the strongest evidence for it: Defect A grew 74 → 85 while Defect B held flat at 14. A leaks on every passing run; B only when a fatal lands in the gap.

## Fix (one file, test-only)

- `launchDirectSupervisor` registers its own stop **before returning**, so no caller can fatal in the gap. Defect B cannot recur for any caller, present or future.
- `reapDriftSupervisors` sweeps by `/proc/<pid>/exe` match and is registered **first** so it runs **last** (`t.Cleanup` is LIFO) — the per-PID stops get their chance, then this catches the replacement. `binaryPath` is unique per isolated env root, so it can only ever reap this test's own supervisors, never a sibling test's or a developer's.
- Reuses the existing `stopDirectSupervisor`, so this adds no new subprocess/sleep/listener call site and `resourcecensus` stays green.

## Why not Pdeathsig

The bead suggested `Pdeathsig`, by analogy with beads' `BEADS_TEST_PDEATHSIG`. That would be wrong here and would break the feature under test: `spawnDetachedSupervisor` is detached **on purpose** so the supervisor outlives `gc start` (its own doc comment says so), and arming `Pdeathsig` would kill it the instant `gc start` returns. The leak is harness-side, not production-side.

Also ruled out: the suite-level sweep is not a missing call but a scoping gap — `stopIntegrationSupervisorWithTimeout` runs `gc supervisor stop --wait` against the *suite* `GC_HOME`, so it structurally cannot see a supervisor living in a per-test `gc-int-env-*` root.

## Verification

- Reproduced both defects live before fixing: one run fataled in the gap and leaked a `gc-home` orphan (Defect B); one run **passed** and leaked a `drift-001` orphan (Defect A). Both were reaped by hand afterward.
- After the fix: same test, same pass, process census **before 88 → after 88, zero new processes**. Pre-fix, that same passing test leaked one.
- `TestSupervisorPIDsFromBinary_MatchesRunningProcess` pins the two parts of the exe-match that otherwise fail silently — the `" (deleted)"` trim and matching a live process at all — using the test binary itself as the fixture, so it spawns nothing. Passes.
- `go vet -tags integration ./test/integration/` clean; `internal/testpolicy/resourcecensus` green.

## Deliberately NOT changed: the third scenario

`setupDriftSystemdScenario` also builds a drift binary but gets **no** reap registration, and that is correct by construction rather than an oversight — worth stating so a reviewer does not have to re-derive it:

- `requireUserSystemd` skips the test unless `XDG_RUNTIME_DIR` is set, so the "drops through to the `direct` branch" case that `setupDriftSystemdScenario`'s own comment warns about is unreachable there.
- The restart therefore takes `systemctl --user restart`, so the replacement supervisor is systemd-owned under the same unit rather than a detached orphan.
- The unit is `Restart=no`, and the existing cleanup stops, disables and removes it — that already registered *before* the fatal-capable `pollHealthBuildID`, so it does not have Defect B's shape either.
- Confirmed empirically: the census shows zero systemd-fingerprint orphans; all 99 are the two buckets above.

Adding a sweep there would be harmless but would imply a hole that is not there.

## Status of the existing orphans — reaped 2026-08-20, after this PR was opened

This PR only stops *new* orphans being created; it does not clean up what was already on the host. When it was opened, 99 were live and accruing at roughly one per hour.

**Those have since been reaped** (by the mayor, under a standing resource-reclamation clearance): 108 by the time it was actioned — it kept growing — all exited on `SIGTERM` with none needing `SIGKILL`, plus 177 stale `gc-int-env` dirs removed. `gc-drift` listening sockets went 108 → 0, and I have independently confirmed zero remaining.

One correction to my own wording, since it was checked and precision matters here: I wrote above that the orphans' "env dirs are deleted". The **binaries** are deleted — every `/proc/<pid>/exe` resolves with a `(deleted)` suffix, because the drift scenario overwrites the binary in place — but the env **directories** were still on disk, all 218 of them. My own census counted them, so the phrasing was simply wrong. It does not weaken the safety argument: the decisive evidence for reaping was 108 listening sockets with **zero** established connections to any of them.

Until this lands, the leak resumes accruing at about one supervisor per hour.

ga-ccltia


quad341:
@quad341 — the cwd-as-fingerprint discriminator is what makes this reviewable at all: it turns "99 mystery processes" into two separately-actionable defects. And the two buckets moving independently (A 74 → 85 while B held flat at 14) is a prediction your theory made and passed, which is stronger evidence than most fixes ship with.

Merging as-is. Three things I checked in the tree rather than took on faith: `ExePath` is the post-`(deleted)`-trim path (`cmd_start_drift.go:484`), so the sweep really does match the detached replacement; the LIFO ordering puts the reap after the per-PID stops but before env-root removal; and no caller double-registers a stop. All hold. Moving the stop inside `launchDirectSupervisor` is the right shape — it makes Defect B unrepeatable for future callers instead of relying on everyone remembering the ordering rule.

Two notes for the record, neither blocking. `rest-full` is push-gated, so CI here only compiled this — the reap gets its first real run post-merge; your census is the evidence until then. And `binaryPath` comes from `MkdirTemp` unresolved while `/proc/exe` resolves symlinks, so on a host with a symlinked `TMPDIR` the sweep would quietly match nothing. Worth an `EvalSymlinks` follow-up; I'll file it rather than touch your verified diff.


---

PR #5444, opened 2026-08-20
fix(test): load buildx images before the docker session protocol checks (ga-5icabz)

quad341:
## Problem

`scripts/test-docker-session` built both test images with a bare `docker build`.
When the docker CLI talks to a Podman socket, buildx selects the
`docker-container` driver, which leaves the build result in the build cache
unless an output is requested:

```
WARNING: No output specified with docker-container driver. Build result will
only remain in the build cache. To push result image into registry use --push
or to load image into docker use --load
```

The images were therefore absent from the engine's image store, the following
image-inspect and run checks reported them missing, and `make test-docker`
failed before exercising the session protocol at all.

## Fix

Build through a small `build_test_image` helper that adds `--load` only when a
buildx builder is reachable **and** reports a driver other than `docker`.

This keeps existing Docker hosts on their current path:

- the `docker` driver already stores results in the engine, so no flag is needed;
- the legacy builder (`DOCKER_BUILDKIT=0`) rejects `--load` outright, which an
  unconditional flag would have broken;
- an unreachable builder yields an empty driver and therefore the pre-existing
  un-flagged behavior.

## Verification

On Fedora with `DOCKER_HOST=unix:///run/user/1000/podman/podman.sock`, where
buildx's default builder is `docker-container`:

- **before:** both images absent after the build; suite dies at the pre-pull check
- **after:** both builds log `importing to docker`, both images inspect clean,
  and the suite runs the protocol (33 passed)

The two remaining failures in that run are an unrelated SELinux bind-mount issue
in `gc-session-docker`, newly visible only because the build no longer fails
first; they are tracked separately.

Fast parallel gate passed locally before push.

Bead: ga-5icabz


quad341:
@quad341 — thank you for chasing this to the actual mechanism instead of stopping at "docker build is broken on Fedora." Naming it precisely — buildx picks the `docker-container` driver when `DOCKER_HOST` is a Podman socket, so the result stays in the build cache and the image-inspect check fails before we test a single session protocol operation — is the hard part, and the comment block you left in the script will save the next person an hour. Gating `--load` on the driver rather than applying it unconditionally is the right call too.

One thing I'll fix on my side before merging. The script runs under `set -euo pipefail`, so `BUILDX_DRIVER="$(docker buildx inspect … | sed …)"` doesn't fall through to the empty-driver case when `inspect` fails — errexit takes the script out right there, silently, since stderr is redirected. I confirmed it in a shell. I'll add `|| true` inside the substitution so the fallback your comment describes actually happens, exclude `DOCKER_BUILDKIT=0` explicitly while I'm in there, and update that sentence to match.

Everything else ships as you wrote it, under your authorship. CI's Docker session job ran the changed script on a `docker-container` builder and went green.


---

quad341:
@quad341 — thanks for running this all the way down to the mechanism. "Docker build is broken on Fedora" is where most people stop; naming it precisely — buildx picks the `docker-container` driver when `DOCKER_HOST` points at a Podman socket, so the build result stays in the cache and the image-inspect check fails before a single session-protocol operation runs — is the hard part. The comment block you left will save the next person an hour.

Gating `--load` on the driver rather than applying it unconditionally is the right instinct, and the conservative default is what makes this safe to ship without a Podman leg in CI: an unreachable or unrecognized builder yields an empty driver and you get exactly today's behavior back. I checked that fallback by hand under `set -euo pipefail` rather than taking it on faith, and it holds.

Merging as-is under your authorship — nothing left on our side. CI's `Docker session` job ran your changed script on a `docker-container` builder and went green, which is about as on-point as evidence gets here. The only thing gating the actual merge button is our deploy gate finishing on HEAD; that's our pipeline, not your patch.


---

Issue #4418, opened 2026-07-18
gc status scans the entire events.jsonl twice (storehealth.LastMaintenance builds a Filter with no Limit)


jacobhausler:
## Summary

`gc status` reads the entire event log **twice** — once per event type — to populate the cosmetic `Last GC:` field in the Store health block. On a live city with a large `events.jsonl` this dominates `gc status` latency, produces a "runtime status probe timed out; using partial status" warning on every run, and is present on both v1.3.5 and main (`6e7465ca`, 2026-07-18).

## Environment

- **gc version:** 1.3.5
- **OS:** macOS 26.5 arm64 (Apple M4, 10-core)

## Root cause

`internal/storehealth/storehealth.go:97` — `LastMaintenance` iterates over two event types (`StoreMaintenanceDone`, `StoreMaintenanceFailed`) and calls `ep.List(events.Filter{Type: spec.typ})` for each, with **no `Limit` field set**.

`internal/events/recorder.go:431` — `FileRecorder.List` delegates to `ReadFiltered`, which JSON-unmarshals every line of `events.jsonl` regardless of how many results are needed.

```go
// storehealth.go:112
evts, err := ep.List(events.Filter{Type: spec.typ})   // no Limit
```

`LastMaintenance` only ever uses the **most recent** event (`e.Ts.After(latestTs)`), so it is scanning millions of events to return at most one useful timestamp.

## Measurements

Live city: `events.jsonl` = 263 MB / ~327 k events.

**Isolated scan cost** (Go test against live file, `GC_LIVE_EVENTS` pointing at the city log):

| call | elapsed | events returned |
|---|---|---|
| `ReadFiltered(Filter{Type: StoreMaintenanceDone})` | 1.645 s | 0 |
| `ReadFilteredTail(Filter{Type: StoreMaintenanceDone}, limit=1)` | 1.683 s | 0 |

Two calls × ~1.6 s = **~3.3 s of the ~4.8 s user CPU** measured in `gc status`. The remaining ~1.5 s user is everything else status does.

**`gc status` wall time** (idle 10-core M4, 3 runs):

```
real 7.27s / 7.12s / 8.14s
user 4.62–5.12s
sys  ~1.6s
```

`/health` answers in ~0.5 ms, so the probe timeout is not a host CPU issue — `gc status` starves its own probe goroutine of a scheduling slot while the main goroutine grinds through the event log.

This city has never emitted `gc.store.maintenance.*` events, so both scans return **zero results**.

[...]

PR #5445, opened 2026-08-20
perf(events): honour the seq window inside archive reads

hexsprite:
## The bug

`events.streamArchive` takes a `Filter` and discards it at the signature:

```go
func streamArchive(path string, _ Filter, fn func(Event) bool) error {
```

So `AfterSeq` never reaches the archive reader. It gunzips from byte zero and `json.Unmarshal`s every record, handing each to a callback that filters afterwards.

`archiveOverlapsFilter` is correct and does its job — it skips an archive that *ends* at or below the cursor. But an order whose cursor lands *inside* the archive's seq window overlaps, so the archive is opened, and once `streamArchive` has it the cursor is gone. To reach the last thousand records of a 160,018-record archive, all 160,018 are decompressed and decoded.

The dispatcher runs this on every tick for every event-triggered order. It is worst for the orders that fire *least*: a cursor only advances when its order fires, so an hourly patrol keeps a cursor deep inside the archive and pays the full decode every tick until it catches up.

## Impact

One rig, rig **suspended**, three idle sessions, no work in flight. Supervisor CPU time over a 120s wall window:

| State | Supervisor CPU |
|---|---|
| 203 MB plain active log, no archive | ~74% of a core |
| 28 KB active log + 9.9 MB `.gz` (160,018 events) | ~68% of a core |
| Same, archive moved out of `.gc/` | 11% of a core |
| **Same, archive restored, with this change** | **13% of a core** |

`sample <pid> 10` confirms it: with the archive present and before this change, `streamArchive` held ~100 on-CPU samples and `readFilteredTracked` 107. After, `streamArchive` is absent from the profile and `readFilteredTracked` drops to 1–2.

Worth flagging for operators: **rotation alone makes this worse, not better.** It trades a linear read of a plain file for repeated gzip decompression of the same records. Someone who sets `max_size_bytes` and stops there sees CPU barely move.

## The change

`streamArchive` skips records outside the seq window before `json.Unmarshal`. `seq` is read with a prefix scan of `{"seq":` — the layout `FileRecorder` writes — with a full-decode fallback so a foreign writer's archive stays readable.

`BeforeSeq` skips rather than early-returning. An early return is faster and archives *should* be seq-ordered, but skipping saves the identical decode without depending on that. Happy to switch it to an early exit if you consider the ordering guaranteed.

Three tests in `reader_archive_filter_test.go`. The first two fail on `main` (`delivered [1 2 3 4 5], want [4 5]`); the third pins that a zero filter still delivers everything.

`ReadFiltered`'s doc comment already promised archives outside the `AfterSeq` window are *"skipped without gunzipping"*. That held archive-by-archive and not record-by-record — this finishes it.

## Related

#4418 fixed the same shape for `gc status`, which was paying two full `events.jsonl` scans per call for a cosmetic field, and added `Filter.MaxScanBytes`. That bound doesn't reach this path because the cost is in the archive branch.

## The larger question

This makes the poll cheap; it doesn't make it unnecessary. `checkEvent` calls `ep.List(...)` on every tick, while `Watch(ctx, afterSeq) (Watcher, error)` is already on the `Provider` interface itself and `FileRecorder` implements it — including the hard part, per its contract: *"every RETAINED event with Seq > afterSeq, in sequence order, exactly once per watcher ... including events that have since rotated into an archive."*

An idle city arguably should cost approximately nothing, and #5296 already moved route-recovery and completions off the steady tick to event-driven deltas. The order dispatcher looks like the same conversion. That changes cursor semantics, the per-tick dispatch budget, and ordering guarantees, so I have kept it out of this PR — but happy to open an issue, or take a run at it if you would find that useful.

### DIFF
--- a/internal/events/reader.go
+++ b/internal/events/reader.go
@@ -424,11 +424,43 @@ func archiveFilesIn(dir string) ([]archiveInfo, error) {
 	return archives, nil
 }
 
+// archiveSeq reads the top-level seq of a raw archive line without decoding
+// the record. FileRecorder writes seq as the first field, so the hot path is a
+// prefix scan with no allocation. Any other layout reports false and the
+// caller falls back to a full decode, which keeps a foreign writer's archive
+// readable.
+func archiveSeq(line []byte) (uint64, bool) {
+	const prefix = `{"seq":`
+	if !bytes.HasPrefix(line, []byte(prefix)) {
+		return 0, false
+	}
+	digits := line[len(prefix):]
+	var seq uint64
+	i := 0
+	for ; i < len(digits) && digits[i] >= '0' && digits[i] <= '9'; i++ {
+		seq = seq*10 + uint64(digits[i]-'0')
+	}
+	if i == 0 {
+		return 0, false
+	}
+	return seq, true
+}
+
 // streamArchive gunzip-streams the file at path, decoding each line
 // as an Event and invoking fn for every event. fn returns false to
 // abort iteration early. Returns nil if iteration completed cleanly
 // or fn requested abort; errors from gzip / scanner are wrapped.
-func streamArchive(path string, _ Filter, fn func(Event) bool) error {
+//
+// Records outside filter's seq window are skipped before json.Unmarshal.
+// archiveOverlapsFilter only rules out an archive that ENDS at or below
+// AfterSeq, so an order whose cursor sits INSIDE the archive's window still
+// opens it — and without this skip every such read decoded the entire archive
+// to reach a handful of trailing records.
+//
+// The skip deliberately does not early-return on BeforeSeq. Archives come from
+// a monotonic log and should be seq-ordered, but `continue` saves the same
+// decode without depending on that.
+func streamArchive(path string, filter Filter, fn func(Event) bool) error {
 	f, err := os.Open(path)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


hexsprite:
The six red checks are all the same transient: `proxy.golang.org` returning `stream error ... INTERNAL_ERROR; received from peer` while downloading module zips (aws-sdk-go-v2, gozstd, parquet-go-source, fslock, readline — all dolt's transitive deps). Every one failed at dependency download, before any test ran. 64 checks passed.

I don't have rerun rights on this repo — a maintainer rerun should clear them.

quad341 (approved):
@hexsprite — this is an excellent piece of diagnosis. Spotting that `streamArchive` took its filter as `_ Filter` and dropped it is one thing; connecting it to why the *least* frequently firing orders pay the worst price — cursor sits deep in the archive, only advances when the order fires, full decode every tick — is the part that's genuinely hard to reach from a profile alone.

The tests are what I appreciate most. Asserting on callback invocations rather than returned rows means they pin the work *not* done, which is the whole point of the change, and `TestStreamArchiveFallsBackWhenSeqNotFirst` covers the case most people would have skipped.

Keep the `continue` for `BeforeSeq` — please don't switch it to an early return. It matches what `activeScanStart` already documents about degrading in speed rather than correctness when the ordering invariant is violated.

Nothing needed from you. Neither red check is yours: I reproduced the residency-ratchet failure locally and it's a stale baseline in `cmd/gc/infra_class_migrate.go`, already fixed on main, and the tmux shard died fetching modules. I'll rebase, rerun, and merge with your authorship intact.

And yes — please open that issue about moving the order dispatcher onto `Watch`. Keeping it out of this PR was the right instinct.


---

quad341:
Re-reviewed `f14985938`..`48914a281` (35 new commits since the last review): no new concerns — the prior review stands.

PR #5446, opened 2026-08-20
fix(config): suppress builtin codex model/effort defaults for profile-routed providers

### LINKED ISSUE #5441: bug: builtin:codex preset OptionDefaults inject --model gpt-5.5 / effort xhigh into every session, clobbering model+effort pinned by a codex --profile routed via args_append
A provider whose `base = "builtin:codex"` and whose `args_append` routes through a `codex --profile` that pins its own model and reasoning effort still gets the preset's hardcoded `OptionDefaults` (`model = "gpt-5.5"`, `effort = "xhigh"`) injected into every session's argv. CLI flags beat profile config in Codex, so the profile's pins are silently overridden for **every** session on that provider. On any strict OpenAI-compatible upstream this is a hard failure at first request; on a permissive upstream it silently routes every session to the wrong model and/or effort lane.

## Gas City version

Observed on `gc 1.4.1` — upstream `main` at `8c73625b974ce8d3d68d54ab42062e6247c47036` (built 2026-08-19T22:39:14Z, reports `-dirty`), macOS arm64, city under a launchd supervisor (`gc supervisor run`). The codex preset `OptionDefaults` block is unchanged on current `main` (`internal/worker/builtin/profiles.go`, `"codex"` entry; line 206 at `8c73625`).

## Reproduction

Minimal repro — any city, any strict OpenAI-compatible endpoint:

1. Define a codex profile that pins a model (any model id the upstream requires, or simply one different from `gpt-5.5`) and an effort:

   ```toml
   # ~/.codex/pin-model.config.toml  (profile "pin-model")
   model = "the/model/i/pin"
   model_reasoning_effort = "medium"
   ```

2. Define a city provider that routes through that profile:

   ```toml
   [providers.pinned]
   base = "builtin:codex"
   args_append = ["exec", "--profile", "pin-model"]
   ```

   with `[upstreams.*]` / `OPENAI_BASE_URL` pointed at an upstream that rejects model ids it does not know (any strict OpenAI-compatible endpoint, e.g. one that requires a `provider/model` prefix).

3. Start a pool (or any) session on that provider (e.g. `gc sling` a trivial bead routed to an agent using the provider, or start the pool).

4. Watch the session's pane / `start-stderr.log`.

## Actual behavior

The preset defaults are materialized into argv for **every** session regardless of the profile:

- the pane's argv contains `--model gpt-5.5` (and `-c model_reasoning_effort=xhigh`) alongside `--profile pin-model`;
- Codex resolves CLI flags over profile config, so the profile's pinned model/effort never take effect;
- the provider rejects the first request and the session dies during startup:

  ```
  warning: Model metadata for `gpt-5.5` not found. Defaulting to fallback metadata;
  this can degrade performance and cause issues.
  ERROR: {"error":{"message":"model must use provider/model prefix","type":"invalid_request_error"}}
  ```

In our city this broke every `builtin:codex`-based provider that routes through a profile, across the whole founding-night wave (~6 concurrent pool sessions).

## Expected behavior

When `args_append` carries `--profile`, the operator has pinned model/effort in the codex profile; the preset's `model`/`effort` OptionDefaults should not be injected (at minimum they should not *override* the profile's values). Profile-aware defaults: detect `--profile` in the resolved args and suppress (or defer to the profile for) preset-injected `--model` / reasoning-effort flags.

## Root-cause trace (for the maintainer)

- `internal/worker/builtin/profiles.go` — the `"codex"` builtin preset declares:
  ```go
  OptionDefaults: map[string]string{
      "permission_mode": "unrestricted",
      "model":           "gpt-5.5",
      "effort":          "xhigh",
  },
  ```
- `internal/config/provider.go:144` documents the precedence: "OptionDefaults overrides the Default value in OptionsSchema entries" — so the preset's `gpt-5.5`/`xhigh` beat the schema's empty "Default" choice, and the resolved values are emitted as CLI flags for the session.
- `internal/config/resolve.go` (`normalizeProviderLayerArgsForSchema`, ~500–525) strips *schema-managed* flags out of `args`/`args_append` and infers their values into the defaults, but `--profile` is not a schema-managed flag, so `exec --profile pin-model` passes through untouched while the preset defaults are still injected — the two mechanisms are never reconciled against each other.
- Net effect: argv = `codex exec --profile pin-model ... --model gpt-5.5 -c model_reasoning_effort=xhigh` (plus whatever `permission_mode` maps to), i.e. the profile is defeated by construction for every session on any `base = "builtin:codex"` provider.

## Workaround (shipped in our city)

Per-provider suppression in `city.toml` — the city layer wins the OptionDefaults merge (`internal/config/resolve.go:375–384`, "city keys win"):

```toml
[providers.pinned]
base = "builtin:codex"
args_append = ["exec", "--profile", "pin-model"]
option_defaults = { model = "", effort = "" }   # suppress preset injection; profile pins win
```

Every `builtin:codex`-based provider that routes through a profile must carry this line until the preset is profile-aware.

## Impact

- Hard startup failure on any strict upstream (model rejected) — pool slots spin forever on the dead startup.
- Silent wrong-lane routing on permissive upstreams: wrong model and/or wrong reasoning effort for every session, with no warning anywhere in gc's own signals.
- Compounds with the restart path: the startup-dead session then loops in the reconciler indefinitely (see companion issue on unbounded `resuming session: session died during startup` retries).

## Suggested fix direction

Make preset OptionDefaults profile-aware: when the resolved provider args (base `Args` + `args_append` chain) contain `--profile`/`-p`, omit preset-injected `--model` and reasoning-effort flags (keep `permission_mode` unless the profile sets it — or make the same deferral configurable). Alternatively, document `--profile` as "takes full precedence" and make `stripArgsSlice` treat profile-routed providers as carrying their own model/effort so the defaults never materialize. A failing test: a `builtin:codex` provider with `args_append = ["exec", "--profile", "x"]` must produce argv **without** `--model gpt-5.5` / effort flags.

jacobhausler:
Fixes #5441.

When a provider routes through a codex profile via `args_append` (`exec --profile <name>`), the builtin preset's `OptionDefaults` (`model=gpt-5.5`, `effort=xhigh`) were injected into every session, clobbering the model/effort the profile pins. The strict upstream then rejects the unprefixed model ("model must use provider/model prefix"), or worse, the wrong effort silently rides a lane whose serving contract depends on it.

Fix: after the OptionDefaults merge (outside the `city.OptionDefaults != nil` conditional — the repro provider declares none), when the merged effective args carry `--profile`/`--profile=`, suppress the preset-sourced `model` and `effort` keys. Explicit provider/agent `option_defaults` still win; `permission_mode` untouched. Copy-on-first-deletion so the builtin registry map is never mutated in place.

Known caveat (commented in code + pinned by test): in leaf→custom→builtin chains, an intermediate layer's explicit model pin is also suppressed when the leaf routes a profile.

Tests: 5 functions + 11 arg-form subtests covering launch AND resume command paths; removing the gate fails 5/7.

### DIFF
--- a/internal/config/chain.go
+++ b/internal/config/chain.go
@@ -160,7 +160,27 @@ func (ctx *chainResolveContext) walkFromLeaf(name string, spec ProviderSpec, cha
 	layerSchema := providerSchemaForLayerArgs(parentMerged, spec)
 	child := normalizeProviderLayerArgsForSchema(spec, layerSchema)
 	ctx.chainSpecs[chainIndex] = child
-	return MergeProviderOverBuiltin(parentMerged, child), nil
+	merged := MergeProviderOverBuiltin(parentMerged, child)
+	// #5441: profile-aware gate on the final (leaf, index 0) merge only,
+	// so its condition is evaluated against the provider's fully merged
+	// effective args. See suppressPresetModelEffortForProfile.
+	if chainIndex == 0 {
+		merged = suppressPresetModelEffortForProfile(ctx.builtinAncestor(), child, merged)
+	}
+	return merged, nil
+}
+
+// builtinAncestor returns the name of the first built-in hop in the
+// chain (leaf -> root order), or "" when the chain has no built-in
+// terminus. Mirrors the BuiltinAncestor computation in
+// resolveProviderChain.
+func (ctx *chainResolveContext) builtinAncestor() string {
+	for _, hop := range ctx.chain {
+		if hop.Kind == "builtin" {
+			return hop.Name
+		}
+	}
+	return ""
 }
 
 // lookupBase resolves a base reference to a ProviderSpec and confirms its
--- a/internal/config/profile_optiondefaults_limits_test.go
+++ b/internal/config/profile_optiondefaults_limits_test.go
@@ -1,0 +1,162 @@
+package config
+
+import (
+	"strings"
+	"testing"
+)
+
+// Characterization tests for the two known limits of the #5441
+// profile-aware OptionDefaults gate (suppressPresetModelEffortForProfile /
+// argsRouteThroughProfile in resolve.go). Both are accepted behavior on
+// this branch; these tests pin them so a future change to the gate has to
+// say so out loud.
+
+// TestProfileGateWrapperOuterProfileFalsePositive: argsRouteThroughProfile
+// scans the merged argv flatly - it has no notion of which side of the
+// wrapper separator a flag sits on. A wrapper whose OUTER tool takes its
+// own --profile therefore trips the gate even though the codex invocation
+// past "--" carries no profile at all, and the preset model/effort defaults
+// are dropped.
+//
+// The direction is fail-safe: the launch loses the preset pins and codex
+// falls back to its own config, rather than clobbering a profile that does
+// not exist. It is recoverable - see the explicit-option_defaults subtest.
+func TestProfileGateWrapperOuterProfileFalsePositive(t *testing.T) {
+	builtinCodex := "builtin:codex"
+	outerProfileWrapper := func(defaults map[string]string) map[string]ProviderSpec {
+		return map[string]ProviderSpec{
+			"codex-wrapper": {
+				Base:    &builtinCodex,
+				Command: "aimux",
+				// The --profile here belongs to aimux, not to codex:
+				// everything past "--" is the codex command line.
+				Args:           []string{"--profile", "hardened", "run", "codex", "--"},
+				OptionDefaults: defaults,
+				ResumeCommand:  "aimux --profile hardened run codex -- resume {{.SessionKey}}",
+			},
+		}
+	}
+
+	t.Run("outer flag suppresses the preset defaults", func(t *testing.T) {
+		providers := outerProfileWrapper(nil)
+		resolved, err := ResolveProviderChain("codex-wrapper", providers["codex-wrapper"], providers)
+		if err != nil {
+			t.Fatalf("ResolveProviderChain: %v", err)
+		}
+		if got := resolved.EffectiveDefaults["model"]; got != "" {
+			t.Errorf("EffectiveDefaults[model] = %q, want empty: the outer --profile trips the gate (known limitation)", got)
+		}
+		if got := resolved.EffectiveDefaults["effort"]; got != "" {
+			t.Errorf("EffectiveDefaults[effort] = %q, want empty: the outer --profile trips the gate (known limitation)", got)
+		}
+		// Fail-safe direction: the permission_mode default is untouched, so
+		// the launch is still well formed - it just lost the model pins.
+		if got := resolved.EffectiveDefaults["permission_mode"]; got != "unrestricted" {
+			t.Errorf("EffectiveDefaults[permission_mode] = %q, want unrestricted", got)
+		}
+		if line := strings.Join(resolved.ResolveDefaultArgs(), " "); strings.Contains(line, "--model") {
+			t.Errorf("ResolveDefaultArgs() = %v, want no --model under the false positive", resolved.ResolveDefaultArgs())
+		}
+	})
+
+	t.Run("explicit option_defaults recover the pins", func(t *testing.T) {
+		providers := outerProfileWrapper(map[string]string{"model": "gpt-5.5", "effort": "xhigh"})
+		resolved, err := ResolveProviderChain("codex-wrapper", providers["codex-wrapper"], providers)
+		if err != nil {
+			t.Fatalf("ResolveProviderChain: %v", err)
+		}
+		// The gate exempts keys the merging layer set explicitly, so naming
+		// them on the wrapper is the documented workaround.
+		if got := resolved.EffectiveDefaults["model"]; got != "gpt-5.5" {
+			t.Errorf("EffectiveDefaults[model] = %q, want gpt-5.5 (explicit defaults are exempt)", got)
+		}
+		if got := resolved.EffectiveDefaults["effort"]; got != "xhigh" {
+			t.Errorf("EffectiveDefaults[effort] = %q, want xhigh (explicit defaults are exempt)", got)
+		}
+		if line := strings.Join(resolved.ResolveDefaultArgs(), " "); !strings.Contains(line, "--model gpt-5.5") {
+			t.Errorf("ResolveDefaultArgs() = %v, want --model gpt-5.5 restored", resolved.ResolveDefaultArgs())
+		}
+	})
+}
+
+// TestProfileGateAgentArgsBypass: the gate runs during provider-spec
+// resolution (lookupProvider / the chain walk), but mergeAgentOverrides
+// REPLACES rp.Args wholesale afterwards (resolve.go, mergeAgentOverrides)
+// and does not recompute EffectiveDefaults. An agent that supplies the
+// --profile in agent.args therefore never reaches the gate: the preset
+// model/effort defaults survive and are injected next to the agent's
+// --profile - the exact clobber #5441 exists to prevent.
+func TestProfileGateAgentArgsBypass(t *testing.T) {
+	city := map[string]ProviderSpec{
+		"codex": {
+			Command:       "codex",
+			ResumeCommand: "codex resume {{.SessionKey}}",
+		},
+	}
+	agent := &Agent{
+		Name:     "worker",
+		Provider: "codex",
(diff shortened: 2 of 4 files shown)


jacobhausler:

This PR is carrying a measured defect from the Westlands city: the builtin Codex `OptionDefaults` clobber profile-routed model/effort, so Westlands currently ships a local patch over released gc.

**Retirement condition for the local patch:** a released build preserves explicit profile-routed model/effort values. Until this lands (or an equivalent is released), Westlands cannot drop the local patch. Tracking this PR as the upstream vehicle.

jacobhausler:
Root-caused the red checks here; **none of them are this PR's diff.** The diff touches only `internal/config/{chain.go,resolve.go,profile_optiondefaults_test.go}`, and no failing step references those files.

All the real failures are `proxy.golang.org` module-fetch flakes, same signature in each:

- `Worker core phase 2 (Codex)` — `github.com/pkg/profile@v1.5.0: read "https://proxy.golang.org/...": stream error: stream ID 299; INTERNAL_ERROR`
- `cmd/gc process / shard 6 of 12` — same, `aws-sdk-go-v2/service/dynamodb`, stream ID 415
- `Integration / rest-smoke-2-of-2` — same, `gastownhall/gascity-packs`, stream ID 5
- `cmd/gc product metrics testhook` — cancelled mid `go: downloading`, no proxy line but the same stall class
- `CI / integration`, `CI / required`, `Worker core phase 2 summary` — aggregation jobs propagating the above

The same class is currently ambient on `main`: run [32615484441](https://github.com/gastownhall/gascity/actions/runs/32615484441) failed with `Integration / rest-full-1-of-16` timing out mid `go: downloading`, and three of main's last five CI runs are red for the same reason.

No code change is warranted. This needs a re-run (I do not have rerun rights on this repo), or an infra-side GOPROXY retry/mirror if the flakiness persists.

jacobhausler:
Rebased onto main b71fadc28 so the evidence watchdog checks out a base.sha that contains scripts/prwatchdog; conflicts: none; tests: internal/config ok.

jacobhausler:
Review follow-up: documenting two limits the gate carries silently. Pushed in 75b1356ad — docs + characterization tests, no behavior change.

**1. A wrapper whose *outer* tool takes `--profile` trips the gate.**

`argsRouteThroughProfile` scans the merged argv flatly; it has no notion of which side of a wrapper separator a flag sits on. So this suppresses codex's preset model/effort even though the codex command line past `--` carries no profile at all:

```toml
[providers.codex-wrapper]
base    = "builtin:codex"
command = "aimux"
args    = ["--profile", "hardened", "run", "codex", "--"]   # aimux's profile, not codex's
```

The direction is fail-safe — the launch loses the preset pins and codex falls back to its own config, rather than clobbering a profile that does not exist — and it is recoverable, because the gate exempts keys the merging layer set explicitly: naming `model`/`effort` in that layer's own `option_defaults` restores them. Both halves are pinned in `TestProfileGateWrapperOuterProfileFalsePositive`.

Making the scan separator-aware would mean teaching config where a given wrapper's own flags end, which is per-wrapper knowledge the provider spec doesn't carry today. Documented rather than fixed.

**2. `agent.args` bypasses the gate entirely.**

The gate is a resolution-time decision over the *provider spec's* args (the Phase A merges in `lookupProvider`, and the `chainIndex == 0` chain walk). `mergeAgentOverrides` then replaces `rp.Args` wholesale afterwards and never recomputes `EffectiveDefaults`:

```go
// resolve.go, mergeAgentOverrides — runs at ResolveProvider step 4,
// after step 3's lookupProvider has already run the gate
if len(agent.Args) > 0 {
    rp.Args = make([]string, len(agent.Args))
    copy(rp.Args, agent.Args)
}
```

So an agent that supplies the `--profile` in `agent.args` never reaches the gate, and the preset `--model` / reasoning-effort still land on the command line next to it — the exact clobber this PR exists to prevent. `TestProfileGateAgentArgsBypass` asserts both flags on the composed launch command.

The mirror holds too, same root cause: when the provider layer routed through a profile (gate fired, pins dropped) and `agent.args` then replaces those args with a profile-free command line, the suppression outlives the args it was decided from — `TestProfileGateStaleAfterAgentArgsReplacement`. Fail-safe direction.

Closing #2 properly means re-evaluating the gate against the post-override argv, i.e. moving it after `mergeAgentOverrides` and recomputing `EffectiveDefaults` there. That is a change to the resolution contract rather than to this function, so I've kept it out of this PR and documented it as a limitation. Happy to do it here instead if you'd rather it land in one piece.

jacobhausler:
Review follow-up: **the gate covers three of the five merge sites; two siblings are ungated.**

`suppressPresetModelEffortForProfile` is applied at every merge this PR touches, but two other call sites layer the same city spec over the same built-in and never see it:

| # | site | gated? |
|---|---|---|
| 1 | `internal/config/resolve.go:202-203` (`lookupProvider`, `builtins[name]`) | ✅ |
| 2 | `internal/config/resolve.go:209-210` (`lookupProvider`, `builtins[spec.Command]`) | ✅ |
| 3 | `internal/config/chain.go:163-168` (leaf chain merge) | ✅ |
| 4 | `internal/config/resolved_cache.go:101,105` (`resolveCustomProviderForValidation`) | ❌ |
| 5 | `internal/api/huma_handlers_providers.go:45,47` (`resolveAllProviders`) | ❌ |

### Evidence

Probe on this branch's head (`75b1356ad`), config = the #5441 shape (`command = "codex"`, `args_append = ["--profile", "pin-model"]`):

```
builtin codex OptionDefaults = map[effort:xhigh model:gpt-5.5 permission_mode:unrestricted]
(1) lookupProvider   [GATED]   OptionDefaults    = map[permission_mode:unrestricted]
(2) validation       [UNGATED] EffectiveDefaults = map[effort:xhigh model:gpt-5.5 permission_mode:unrestricted]
(3) providers API    [UNGATED] reported effective_defaults = map[effort:xhigh model:gpt-5.5 permission_mode:unrestricted]
```

### Impact

**Site 4 (validation) is harmless.** `validateResolvedProviderOptions` only checks that each default resolves against the schema; suppression strictly *removes* keys, so the ungated path validates a superset. It cannot produce a false failure — the values it validates are the built-in's own.

**Site 5 (providers API) is a real display divergence.** `resolveAllProviders` feeds both `GET /v0/city/{cityName}/providers` (`huma_handlers_providers.go:78`) and the browser-safe list (`:94-97`), and `toProviderPublicResponse` recomputes `EffectiveDefaults` from the ungated merged spec at `handler_providers.go:92`. So the API and dashboard will advertise `model=gpt-5.5` / `effort=xhigh` for exactly the profile-routed providers where the launcher now (correctly) drops them.

### One correction to how I'd previously framed this

I had this filed internally as "needs the gate exported." That's only half right:

- `resolved_cache.go` is **`package config`** — same package, so it can call `suppressPresetModelEffortForProfile` directly. No export needed.
- `huma_handlers_providers.go` is **`package api`**, so only that one needs an exported entry point.

And the `builtinAncestor` argument differs per site, so it isn't a pure copy-paste: sites 4 and 5 would pass `name` / `spec.Command` the same way `lookupProvider` does.

### Suggestion

I don't think this needs to grow this PR. Site 4 is a no-op in practice and site 5 is display-only — neither blocks the launcher fix, which is the thing #5441 actually bites on. My preference is to **land this as-is** and take the providers-API site as a follow-up, since a shared exported helper is a wider surface than this fix should carry.

Flagging it here rather than silently, because once this merges the API becomes the only place still reporting the clobbering defaults — which is a confusing place for the divergence to live if nobody wrote it down.

jacobhausler:
Root-caused this run's `Evidence / critical-path suites` failure; it's a watchdog race, not this PR's diff.

The watchdog (run [33795868062](https://github.com/gastownhall/gascity/actions/runs/33795868062)) started at 19:20:56Z and gave up at 19:46:06Z reporting `CI / required: incomplete` and blank `Mac regression summary` / `Integration / review-formulas` rows. But `CI / integration`'s own `Integration / rest-smoke-2-of-2` job didn't finish until 19:50:05Z — the watchdog timed out ~4 minutes before the checks it was waiting on actually completed. All three of those checks are now green (`CI / required`, `Mac regression summary`, `Integration / review-formulas` all show `pass`), and no other check on head `75b1356ad8` is failing.

No code change warranted; this is the same "watchdog polled and gave up early" class as the earlier proxy.golang.org flake, just a different trigger. Needs a re-run (no rerun rights on this repo) — an admin re-running `Evidence / critical-path suites` alone should turn the PR green.

PR #5447, opened 2026-08-20
fix(reconcile): accrue wake failures for startup-dead sessions in the rollback path

### LINKED ISSUE #5442: bug: reconciler retries "resuming session: session died during startup" forever — no bounded retries, no fresh-start fallback, no terminal state
When a session dies *during startup* and the session bead has no provider `session_key` persisted (i.e. the death happened before the provider ever returned a resumable rollout id), the session manager's fresh-start fallback never fires — the error falls into the `default` branch, is wrapped as `"resuming session: %w"` (misleading: nothing is being resumed), and is returned to the reconciler, which retries the identical `Start` on every reconcile tick. The result is an unbounded loop of `outcome=provider_error ... err=resuming session: session died during startup` against a demonstrably healthy lane, with no retry bound, no backoff, no automatic fresh-start fallback, and no terminal failed state. A single startup-dead session holds its pool slot and spams the supervisor log indefinitely until a human intervenes.

## Gas City version

Observed on `gc 1.4.1` — upstream `main` at `8c73625b974ce8d3d68d54ab42062e6247c47036` (built 2026-08-19T22:39:14Z, reports `-dirty`), macOS arm64, city under a launchd supervisor (`gc supervisor run`). Same structure on current `main`: `internal/session/chat.go`, both `sp.Start` error handlers (ensureRunning path ~line 402, restart bridge ~line 523) and `retryFreshStartAfterStaleKey` (~line 167).

## Reproduction

Two equivalent recipes — the trigger is *any* startup-time provider death that occurs before a `session_key` is persisted:

**Recipe A (the reported one — provider dies mid-stream on a healthy lane):**
1. Run a city with an OpenAI-compatible upstream (real or emulated) and at least one pool worker session active.
2. Kill the provider mid-stream (stop the endpoint; the real-world trigger is a GPU/driver drop or endpoint reboot).
3. Let the session's restart path resume against the now-dead rollout.
4. Observe the reconciler log.

**Recipe B (minimal, no provider outage needed):**
1. Point the provider's upstream at an endpoint that rejects the session's first request (any hard startup error: bad model id, auth failure, endpoint down at first request).
2. Start a pool session on that provider.
3. Observe the reconciler log.

In both recipes the session dies during startup *before* any provider rollout id exists, so the bead's `session_key` metadata is empty.

## Actual behavior

The reconciler retries the identical start on every tick. From `~/.gc/supervisor.log` of our city (2026-08-20, session `qwen38-low-1-pool`, template `qwen38-low`; the underlying startup error in that instance was a separate config defect — the loop is orthogonal to it):

```
session reconciler: starting qwen38-low-1-pool: resuming session: session died during startup:
  session "qwen38-low-1-pool"; diagnostic written to .../sessions/qwen38-low-1-pool/start-stderr.log;
  last pane output: ... ERROR: {"error":{"message":"model must use provider/model prefix",...}}
  Pane is dead (status 1, ...)

session lifecycle: op=start wave=0 session=qwen38-low-1-pool template=qwen38-low
  outcome=provider_error duration=16.49s phases=[start_call=16.49s]
  err=resuming session: session died during startup: ...
```

Counted in one log: **84** `resuming session: session died during startup` starts and **83** `outcome=provider_error` lifecycle results for the single session name, at a steady ~22 s cadence for ~30 minutes (03:26–03:56 local), ending only when the underlying config was fixed. The lane was demonstrably healthy: other sessions on the same provider started and ran normally throughout, and the same lane served the next dispatch successfully seconds after the loop ended.

## Expected behavior

- Startup failures should be bounded: a small, explicit retry budget (with backoff) for a session that repeatedly dies during startup.
- When the budget is exhausted, fall back automatically to a true fresh start (clear stale resume metadata regardless of whether a `session_key` was ever persisted), or enter a terminal failed state that is doctor-visible and frees the pool slot — not an infinite identical-error loop.
- The fresh-start fallback should not be gated on `session_key != ""`: a session that never obtained a key is precisely the one that cannot ever resume and must be restarted fresh.

## Root-cause trace (for the maintainer)

1. `internal/runtime/tmux/adapter.go:1101–1114` (and siblings in `internal/runtime/{ssh,k8s}/provider.go`): a pane that dies during startup returns `runtime.ErrSessionDiedDuringStartup` (`internal/runtime/runtime.go:37`).
2. `internal/session/chat.go` — both `sp.Start` error handlers (ensureRunning ~402, restart bridge ~523) share the same shape:

   ```go
   switch {
   case errors.Is(err, runtime.ErrSessionDiedDuringStartup) && b.Metadata["session_key"] != "":
       retried, err := m.retryFreshStartAfterStaleKey(ctx, id, &b, sessName, resumeCommand, cfg, unroute)
       ...
   case errors.Is(err, runtime.ErrSessionExists) && m.sp.IsRunning(sessName):
       return err
   default:
       return fmt.Errorf("resuming session: %w", err)   // ← startup-deaths without a key land here
   }
   ```

   The fresh-start fallback (`retryFreshStartAfterStaleKey`, chat.go:167 — clears `session_key`, `started_config_hash`, priming markers, strips the resume flag, restarts) is reachable **only** when the bead already carries a non-empty `session_key`. A session that died during startup *before* any key was persisted has `session_key == ""`, so the `case` never matches and the error is wrapped and returned via `default` — which also mislabels a bare start as "resuming session".
3. The reconciler's `ensureRunning` has no retry bound, backoff, or terminal state for repeated startup failures of a pool session: every tick re-issues the same `Start`, gets the same wrapped error, logs it, and holds the pool slot until a human runs `gc session reset <name>` (or the root cause is fixed elsewhere).
4. Note the related post-start guard (`staleKeyDetectionWaiter`) only runs `if started` — it cannot catch the keyless startup-death case either.

## Workaround (shipped in our city)

- `gc session reset <name>` per stuck session — the controller then starts it fresh.
- Standing rule after any provider outage: sweep `provider_error` + `err=resuming` sessions in the supervisor log and reset each one (recurring failure, 3 occurrences receipted in our city's failure archive: two low-load drops 2026-08-19 and one under first burst load 2026-08-20, ~21 h apart).

## Impact

- A single startup-dead session occupies its pool slot forever and masks real capacity; dispatches routed to that agent stall with no visible reason in gc's own signals.
- Supervisor log is filled with ~20 s-interval error dumps (each carrying the full pane output), which buries the actual root-cause line; in our incident the loop ran ~30 minutes and 84 iterations before anyone noticed.
- Violates the loud-degrade principle: a permanently-wedged pool worker is not reported as such by any doctor check.

## Suggested fix direction

- Make the keyless startup-death path first-class: on `ErrSessionDiedDuringStartup` with `session_key == ""`, apply the same bounded-retry-then-fresh-start treatment as the keyed path (the "stale key" clear is already correct for the keyed case; generalize it to "no usable resume target").
- Add an explicit per-session startup-failure budget (count + backoff window, e.g. 3–5 failures over N minutes); on exhaustion, either fresh-start with cleared resume metadata or transition to a terminal failed state surfaced to `gc doctor` and freed from the pool.
- A failing test: force N consecutive startup deaths on a keyless session; assert (a) the loop terminates, (b) the session either runs fresh or is marked failed-and-visible, and (c) the pool slot is released in both cases.

jacobhausler:
Fixes #5442.

A pool session that dies during startup was rolled back and re-minted every tick with fresh counters, so the existing wake-failure budget (`recordWakeFailure` → quarantine at `defaultMaxWakeAttempts` + `ConversationResetPatch`) never engaged. Observed live: 84 identical iterations at ~22s cadence with quarantine never firing.

Fix: gate the `rollbackPending` arm on `!errors.Is(err, runtime.ErrSessionDiedDuringStartup)` so startup deaths fall through to the existing wake-failure tail — the same bead accrues attempts and quarantines at the budget. Post-startup rollbacks (provider errors, cold-start deadlines, session-exists mismatches, canceled-commit) are untouched. Also wraps the bare startup-death error at the post-start liveness check in the sentinel for uniform detection.

**Behavioral disclosures for maintainer assent:**
- **Slot-holding**: a startup-dead pool member now retains its pending bead/slot while attempts accrue (up to 5 tries + 5m quarantine) instead of recycling every tick; `staleCreatingStateTimeout` still bounds it. This is the backpressure the issue asks for, but it is a semantics change.
- **Trace-site shift**: startup deaths record through `TraceSiteLifecycleStartFailed` rather than `TraceSiteLifecycleStartRollback` — dashboards keyed on the rollback site will shift.

Note per the follow-up comment on #5442: the original issue text pointed at the resume gate; accruing there (or in the rollback arm while still closing the bead) is a no-op, since the close re-mints a bead with fresh counters — the negative-control test pins that symptom.

Tests: multi-tick accrual → quarantine+reset on a startup-dead pool session; post-startup deaths don't accrue via this arm; sentinel-wrap pinned.

### DIFF
--- a/cmd/gc/session_lifecycle_parallel.go
+++ b/cmd/gc/session_lifecycle_parallel.go
@@ -1465,7 +1465,12 @@ func runPreparedStartCandidate(
 				alive = obs.Alive
 			}
 			if err != nil || !running || !alive {
-				err = fmt.Errorf("session %q died during startup", item.candidate.name())
+				// Wrap the shared sentinel so this locally detected death is
+				// indistinguishable from a provider-reported one downstream
+				// (commitStartFailure's startup-death exception, #5442). The
+				// sentinel's own text is "session died during startup", so the
+				// rendered message keeps the substring existing tests pin.
+				err = fmt.Errorf("%w: session %q", runtime.ErrSessionDiedDuringStartup, item.candidate.name())
 			}
 		}
 		phases.PostStartObserve = time.Since(postStartBegin)
@@ -2254,7 +2259,7 @@ func commitStartFailure(result startResult, sessFront *sessionpkg.Store, clk clo
 		logLifecycleOutcome(stderr, "start", wave, name, tp.TemplateName, string(result.outcome), result.started, result.finished, result.err, result.phases)
 		return
 	}
-	if result.rollbackPending {
+	if result.rollbackPending && !errors.Is(result.err, runtime.ErrSessionDiedDuringStartup) {
 		if errors.Is(result.err, context.DeadlineExceeded) {
 			rec.Record(events.Event{
 				Type:    events.SessionColdStartTimeout,
@@ -2268,6 +2273,18 @@ func commitStartFailure(result startResult, sessFront *sessionpkg.Store, clk clo
 		// TestReconcileSessionBeads_RollsBackPendingCreateOnProviderError).
 		// Genuine wake-failure accounting happens on the non-rollback path
 		// below via recordWakeFailure.
+		//
+		// EXCEPTION — startup death (#5442): the recreate-fresh loop only
+		// converges when the next attempt can succeed. A session that dies
+		// during startup on every attempt is discarded and re-minted every
+		// tick, and each fresh bead carries a zero wake_attempts counter, so
+		// the retry budget is never reached and the pool spins forever (the
+		// reported incident: 84 rollback/recreate cycles in ~22s with no
+		// quarantine). Startup deaths therefore skip this arm and fall
+		// through to the wake-failure tail below, where the SAME bead accrues
+		// across ticks until defaultMaxWakeAttempts trips the quarantine and
+		// the conversation reset. Post-startup rollbacks (provider errors,
+		// cold-start deadlines, session-exists mismatches) are unchanged.
 		if trace != nil {
 			trace.RecordOperation(TraceSiteLifecycleStartRollback, TraceReasonStart, result.outcome, "", tp.TemplateName, name, 0, traceRecordPayload{
 				"error": formatLifecycleError(result.err),
--- a/cmd/gc/session_lifecycle_parallel_test.go
+++ b/cmd/gc/session_lifecycle_parallel_test.go
@@ -6238,6 +6238,51 @@ func TestExecutePreparedStartWave_StaleSessionKeyDetected(t *testing.T) {
 	}
 }
 
+// The post-start liveness probe's own "died during startup" verdict must carry
+// runtime.ErrSessionDiedDuringStartup, exactly as a provider-reported death
+// does: commitStartFailure's startup-death exception (#5442) keys off the
+// sentinel, and this locally detected form is the shape the pool spin was
+// reported in.
+func TestExecutePreparedStartWave_StaleSessionKeyDeathWrapsStartupSentinel(t *testing.T) {
+	sp := &zombieAfterStartProvider{Fake: runtime.NewFake()}
+	item := preparedStart{
+		candidate: startCandidate{
+			info: sessionpkg.Info{
+				ID:                  "gc-99",
+				SessionName:         "test-agent",
+				SessionNameMetadata: "test-agent",
+				SessionKey:          "stale-key-abc",
+				Template:            "worker",
+			},
+			tp: TemplateParams{
+				Command:      "claude --resume stale-key-abc",
+				SessionName:  "test-agent",
+				TemplateName: "worker",
+			},
+		},
+		cfg: runtime.Config{
+			Command:      "claude --resume stale-key-abc",
+			ProcessNames: []string{"claude"},
+		},
+	}
+
+	results := executePreparedStartWave(
+		context.Background(),
+		[]preparedStart{item},
+		sp,
+		nil,
+		10*time.Second,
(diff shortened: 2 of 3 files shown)


jacobhausler:

This PR is carrying a measured defect from the Westlands city: startup-dead sessions do not accrue wake failures / fresh-start state in the rollback path, so the retry loop has no tested wake-failure state to bound on.

**Retirement condition for the local patch:** the released reconciler classifies startup-dead exits, bounds retries, and retains the tested wake-failure state. Until this lands, Westlands runs a local patch over released gc. Tracking this PR as the upstream vehicle.

jacobhausler:
Rebased onto main b71fadc28 so the evidence watchdog checks out a base.sha that contains scripts/prwatchdog; conflicts: none; tests: cmd/gc ok.

jacobhausler:
Superseded by #6016, merged 2026-09-04, which solves the same problem (startup-dead sessions never accruing wake-failure state, so retries never quarantine) with a more durable fix: a startup-health episode keyed by session name that survives pending-create rollback and bead replacement, rather than this PR's approach of skipping the rollback arm so the existing per-bead wake-failure counter can accrue.

Rebasing this PR onto current main produces a real conflict in cmd/gc/session_lifecycle_parallel.go between the two approaches — #6016's episode-based tracking is already live and covers the reported symptom (84 rollback/recreate cycles with no quarantine). Closing this one as superseded rather than reconciling two competing fixes for the same bug.

Issue #4386, opened 2026-07-17
Long-lived agents have no periodic checkpoint — a dead mayor's state is unrecoverable


esciara:
# Long-lived agents have no periodic checkpoint — a dead mayor's state is unrecoverable

## Summary

Gas City has no mechanism that creates a recovery point for a long-lived agent **without that agent's voluntary cooperation**. When the mayor dies — SIGKILL, tool timeout, or `gc stop` — everything it has not already chosen to write to beads/mail is gone. The next incarnation starts cold, with no evidence the prior session existed or what it was doing.

## Observed incident (2026-07-17)

1. `09:49` — mayor voluntarily runs `gc handoff`, writing a thorough continuation record.
2. A successor session starts and reads it (mail correctly marked `read`).
3. `10:01` — `gc stop`. The successor is interrupted before acting on any of it.
4. `11:04` — next mayor starts. `gc mail inbox` (unread-only) is empty. It concludes **no handoff exists and nothing is pending** — and says so to the operator.

The handoff survived only because the *previous* session had volunteered one ~12 minutes earlier. The successor contributed nothing: it read a plan, formed intent, and died leaving no trace. Nothing in the system noticed or recorded that.

This is not a mail bug — marking mail read on read is correct, and read ≠ completed is a legitimate distinction. The gap is that **no recovery point is ever created except by an agent choosing to create one.**

## Evidence

Hooks installed (`internal/hooks/config/claude.json`, written by `installClaude`, `internal/hooks/hooks.go:402-416`) — exactly three:

| Event | Command |
|---|---|
| `SessionStart` | `gc prime --hook` |
| `PreCompact` | `gc handoff --auto "context cycle"` |
| `UserPromptSubmit` | `nudge drain --inject`, `mail check --inject` |

No `Stop`. No `SessionEnd`. Every provider overlay follows the same three-event shape (`internal/bootstrap/packs/core/overlay/per-provider/`). The repo's own parity audit (`engdocs/archive/analysis/non-claude-provider-parity-audit.md:144-153`) tabulates a "Session end" column for each provider — every slot is documented as available and left empty.

[...]

Issue #1029, opened 2026-04-20
bug: session reaped mid-work when assignedWorkBeads briefly zero during step transition


rileywhite:
## Before you continue

- [x] I searched existing issues and did not find a duplicate.
- [x] I read the relevant docs and contributor guidance.

## Gas City version

0.15.2

## Environment

Linux x86_64, bash, supervisor-managed agent pool

## Reproduction

1. Run a multi-step work sequence where steps are sequentially closed and re-assigned (step N closes, step N+1 is created and assigned to the same session)
2. Let the reconciler run with its default tick + caching-store refresh interval (which can be several seconds)
3. When step N is closed just before the next caching-store refresh, the session briefly has `assignedWorkBeads = 0`
4. The reconciler marks the session as orphaned and drains it

The race window is small (one reconcile cycle) but non-zero, and becomes more likely when:
- The caching-store reconcile interval is long (observed at ~5s)
- The agent pool has sessions near the `poolDesired` boundary

## Expected behavior

A session actively executing work should not be drained during a step transition. The reconciler should tolerate a brief gap in assigned beads when the session was recently active.

## Actual behavior

The reconciler drains the session when `assignedWorkBeads` briefly equals zero during the step-N → step-N+1 handoff, even though the session is live and has in-flight tool calls.

## Logs, screenshots, or traces

```
T+0    poolDesired = 1
T+1    assignedWorkBeads: 2 (step bead + parent bead)
T+18   caching-store reconciled (5.1s — slow)
       step transition: step N closed, step N+1 not yet visible
T+18   assignedWorkBeads: 1 (only parent bead — step bead missed)
T+25   Draining session: orphaned
T+31   assignedWorkBeads: 2 (step N+1 now visible — too late)
T+34   last agent activity (Bash tool call in-flight)
T+55   Stopped drain-acked session
T+62   WARN: reconciler: reaped stale session bead
```

Session exited cleanly after its in-flight tool call finished. No error surfaced to the agent — the work silently evaporated.

## Additional context

This is the inverse polarity of #855:
- #855: false-positive forever (session dead but `assignedWorkBeads` still true → cannot converge to drain)
- This issue: false-negative transient (session alive and working, but `assignedWorkBeads` briefly false during step swap → reaped)

Any fix for #855 that makes drain decisions more aggressive would worsen this mode.

Possible mitigations:
- Grace window before draining on observed `assignedWorkBeads == 0`
[...]

PR #5448, opened 2026-08-20
fix(session): refuse remote session kills that would destroy live subagents

wbern:
## TL;DR

Destroying a session destroys the provider process, and coding-agent subagents run **in-process**. So `gc handoff --target` and `gc session kill` silently destroy the target's in-flight background subagents. Neither checked for that first — they verified only that the target was controller-restartable and running, then killed it.

This makes both refuse, showing exactly what would be lost, unless you pass `--force`.

## Trigger

Reproduced from a captured transcript. A background subagent was spawned, worked for 15 minutes, and a remote handoff destroyed its session mid-tool-call:

| Time (UTC) | Event |
|---|---|
| `09:00:40.121` | background subagent spawned |
| `09:15:46.012` | subagent issues a `tool_use` — actively working |
| `09:15:55.778` | `gc handoff --target <session>` invoked |
| `09:15:57.982` | subagent receives its `tool_result` — **its last entry**, mid-loop |
| `09:15:58.218` | CLI prints `killed session (reconciler will restart)` |
| `09:18:45.246` | successor session reports `No completion record was found for background agent …` |

The provider's own recovery message confirms the loss is real but recoverable *if you know the task id* — which nothing surfaced.

## Behaviour

```gherkin
Given a target session with a background subagent that has no terminal task notification
When gc handoff --target or gc session kill runs without --force
Then the command refuses, lists each live subagent, and exits non-zero

Given the same target
When the command runs with --force
Then it destroys the session exactly as before

Given a target whose subagents have all reached a terminal task notification
When either command runs
Then it proceeds unchanged

Given a target whose transcript is absent, unreadable or corrupt
When either command runs
Then it proceeds unchanged (fail-open)
```

## Detection

An exact structured join, no heuristics and no time thresholds:

- `subagents/agent-*.meta.json` supplies `toolUseId`
- matched against the parent transcript's `Agent` `tool_use` blocks
- and against terminal task notifications by `task-id` (`completed`/`failed`/`killed`/`stopped`)

Two details that matter:

- **Synchronous subagents are excluded** — they block the parent and so cannot be live at kill time. Only an explicit `run_in_background: false` counts as synchronous, because the field is omitted far more often than it is set.
- **A background spawn's `tool_result` is a launch acknowledgement, not a result**, so its presence must not be read as completion.

`FindAgentMappings` now prefers `agent-*.meta.json` for the parent tool-use id, falling back to the transcript's `parentToolUseId` only when that file is absent, and deduplicates by agent id so one subagent yields one mapping.

## Why it is safe

- **Fail-open.** Any detection error proceeds exactly as before, so this can never block a legitimate kill.
- **Opt-out preserved.** `--force` restores prior behaviour.
- **Scope is the two operator-initiated paths only.** Controller-driven destruction is untouched by this change.
- Cost is paid once per invocation, not on a hot path.

## Tests

Golden fixtures are trimmed captures of the real transcript above, asserting both directions: the live subagent is detected at the original kill timestamp, and a subagent with a terminal notification is not. Plus unit cases for omitted-`run_in_background`, explicit-synchronous, terminated, attachment-form notifications, and all three fail-open paths.

## Scope

`cmd/gc/cmd_handoff.go`, `cmd/gc/cmd_session.go`, `cmd/gc/subagent_kill_guard.go` (new), `internal/worker/subagent_guard.go` (new), `internal/sessionlog/agents.go`, plus tests and `internal/worker/testdata/`.

## Related work

- #1029 — session reaped mid-work; same class of loss from a different trigger.
- #4386 — long-lived agents have no periodic checkpoint.


quad341:
@wbern, thank you for this. Rebuilding the incident from the real transcript made the bug impossible to argue with, and an exact structured join that fails open is the right design for a guard on a destructive path.

We're happy with the contract: `gc session kill` and `gc handoff --target` refuse without `--force` when the target has live background subagents, and proceed as before whenever detection can't read the transcript. Sorry for the wait.

To land it, please:

1. **Rebase onto main.** The branch conflicts now. Of your files, `cmd/gc/cmd_handoff.go` and `cmd/gc/cmd_session.go` have both changed on main since you branched.
2. **Regenerate the reference docs** after rebasing: run `make generate` and commit the result. On your last CI run only `docs/reference/cli.md` was stale (the two new `--force` flags). Please regenerate rather than apply the Aug 20 CI patch, since main has moved.

While you're in there, please fold in these from our review:

- **Read the transcript before the DAG.** The guard reads through `Transcript(ctx, TranscriptRequest{Raw: true})`, which ends in `BuildDag`, and `BuildDag` skips entries without a `uuid`. Real `queue-operation` records don't carry one, so a subagent whose completion was recorded only that way reads as live and the kill is refused. It fails safe, but it's a false refusal. The fix we'd suggest is a thin exported wrapper over `parseFileDetailed` in `internal/sessionlog`, reached through `SessionLogAdapter`. Please run the incident fixtures through the real adapter from a temp file (the tests currently inject `RawMessages`, which skips this path), and add a case whose only completion record is a `queue-operation` plus one whose spawn sits before a compaction boundary.
- **Guard before mail.** `doHandoffRemoteWithForce` calls `createHandoffMail` before the guard, so a refusal has already sent the handoff and the `--force` retry sends a second copy. A test that a refusal sends nothing and the retry sends exactly one would pin it.
- **Empty meta shouldn't win.** In `FindAgentMappings`, let `agent-*.meta.json` override only when its `toolUseId` is non-empty, with a test for a meta file that lacks one.
- **Census keys.** Shard 11 fails `TestCanonicalSessionProviderFactoryCallerCensus`: lines 57, 58 and 71 of `cmd/gc/provider_factory_census_test.go` still name `cmdHandoff`, `cmdHandoffRemote` and `cmdSessionKill` and need the `…WithForce` names.

Shard 9 (`TestBdRigWorktreeStoreConsistentAcrossRawBdGcBdAndProviderStore`) doesn't touch your diff; just check whether it recurs after the rebase.

wbern:
Thanks — rebased onto current `main` and worked through the list.

**Read the transcript before the DAG.** Added `sessionlog.ReadFileRecords`, a thin exported wrapper over `parseFileDetailed`, reached through `SessionLogAdapter.TranscriptRecords` and surfaced on the worker boundary as `Handle.TranscriptRecords`. The guard now reads records in file order rather than a session read.

Writing the fixtures for this turned up something worth flagging: **no entry in either incident fixture carries a `uuid`**, so driven through the real adapter, 0 of 9 and 0 of 2 records survived `BuildDag` and the guard reported nothing live in either direction. The golden tests passed only because they injected `RawMessages`. They now run from a temp file through the real adapter, which is what surfaced it.

Two new cases, both driven the same way:

- **Completion recorded only as a `queue-operation`** — reproduces the false refusal you described. Before the fix the guard reports `live=[{queueonly, "Audit the nudge path"}]` for a subagent that had already completed.
- **A spawn stranded before a compaction boundary** — the opposite failure. Once the post-compaction turns form the tip, the spawn falls off the active branch, so the pruned view hides it entirely and the kill destroys live work.

**Guard before mail.** `doHandoffRemoteWithForce` now resolves restartable/running state and runs the guard before `createHandoffMail`. A test asserts a refusal sends nothing and the `--force` retry sends exactly one; reverting just the reorder puts it back at `mail after refusal = 1`.

**Empty meta shouldn't win.** `FindAgentMappings` no longer lets an absent or empty `toolUseId` erase a join key recovered from the transcript.

**Census keys** now use the `…WithForce` names. `cmdHandoff` had become a test-only wrapper after the rename and tripped `unparam`, so I removed it and pointed the four test call sites at `cmdHandoffWithForce`.

`make generate` adds only the two `--force` rows to `docs/reference/cli.md`.

On shard 9: it does not recur. Running `make test-cmd-gc-process-parallel` on this branch and on a pristine `main` worktree gives the same four failures — `TestBuildDesiredState_MinZeroDefaultScaleCheckRoutedWorkCreatesPoolSession`, `TestEvaluatePoolDefaultScaleCheckCountsRoutedReadyWork`, `TestEvaluatePoolDefaultScaleCheckIgnoresRoutedActiveUnassignedWork`, and `TestFreshManagedBdCityInitSeedsPinnedHQDatabaseAndKeepsGCPrefix` — none of which touch this diff. Two further tests failed on one branch run and cleared on a re-run; both pass or skip in isolation, and adding tests shifts shard composition, so they appear to be contention-sensitive rather than related.

quad341 (approved):
@wbern — merging this.

The thing worth calling out isn't the guard itself, it's how you proved it. Switching the golden fixtures to run through the real `SessionLogAdapter` from disk, then *reporting* that it exposed the previous tests as vacuous — 0 of 9 records surviving `BuildDag` — is what made the rest of the diff trustworthy. And the pre-compaction fixture asserting `pruned < records` means it can't quietly rot back into a no-op later. The queue-only-completion case is the subtle one most people would have missed entirely.

Reconstructing it from the captured incident, down to the minute-by-minute timeline, is also why this reviewed fast: the failure mode was never in question.

Two things we're taking on our side, neither blocking you: `gc session kill` now parses the full parent transcript with no deadline, on a path people reach for when something is already stuck; and the guard-before-mail reorder means an observation failure now sends no handoff mail at all. Both are the right trade — just ours to watch. There's also a narrow follow-up around malformed JSONL records being skipped silently; we'll handle it separately, since the obvious fix would disable the guard on live transcripts.

Genuinely nasty class of silent data loss. Thank you.


---

PR #5450, opened 2026-08-20
fix(testpolicy): ignore untracked nested worktrees in source censuses

quad341:
## What this changes

Composition-root, storage-boundary, and test-resource censuses now enumerate the repository's git-tracked Go files. An untracked nested worktree below the checkout no longer contributes duplicate source files or causes false CI failures.

Hermetic fixture scans keep their filesystem-based behavior. The checked test-resource ledger is updated for the two new regression fixtures that invoke Git.

## Review notes

- `resourcecensus.TrackedGoFiles` is the shared repository-boundary helper; it normalizes paths returned by `git ls-files -z`.
- The two regression tests create a tracked Go file plus an untracked nested-worktree-shaped directory and prove only the tracked file reaches each census.
- Runtime orchestration behavior is unchanged; this affects test-policy and source-boundary checks.

## Test plan

- [x] `make test-fast-parallel` — all 10 jobs pass on the gated candidate.
- [x] `make test-cmd-gc-process-parallel` — all 7 jobs pass.
- [x] `make test-ci-policy` and `go vet ./...` pass.
- [x] All 14 diff-owned tests pass by name; focused acceptance runs report 43 PASS, 0 FAIL, 0 SKIP events.
- [x] Release gate: [`release-gates/ga-dmha63-nested-worktree-census-gate.md`](release-gates/ga-dmha63-nested-worktree-census-gate.md)


quad341:
@quad341 — this is the right fix, and specifically it's a *better* fix than the one that landed on `main` while it sat. Scoping the censuses to `git ls-files` instead of adding `"worktrees"` to a directory blocklist means the guard survives a worktree pool at any path — which matters, since `bead_worktree_liveness.go` itself notes they turn up under `.gc/worktrees`, `.claude/worktrees`, or anywhere else. Trading a name for a mechanism is the more durable call.

Two things worth crediting. Pulling `root` out into `censusSourcesFrom` while leaving the `censusMinimumFiles` floor on `censusSources` is what makes the walk testable against a fixture repo at all — a careful seam, not an incidental one. And both regression tests assert an exact singleton rather than a weaker "doesn't contain," so a slide back to filesystem walking fails loudly.

I checked the thing I'd have worried about most: the tracked set and the old walk both come to 1477 files here, so `testdata` really was the only exclusion that mattered.

I'll take it from here — rebasing, keeping the `worktrees` skip in `anyGoWorkFile` since that one's still a filesystem walk, and recomputing the subprocess baselines, which both sides moved. Nothing for you to do.


---

Issue #3681, opened 2026-06-23
gc sling --on <formula> silently drops the target bead's instructions


bobbyhyam:
# `gc sling … --on <formula>` silently drops the target bead's instructions

## Summary
When a build is launched with `gc sling <target> <bead> --on <formula>`, the **target bead's description
(the actual user instructions) is never propagated into the wisp / brainstorm context.** The brainstorm
reads only the generic formula-root boilerplate + the repo README, so user intent is silently lost.

## Evidence / mechanism (source)
- `--on` cooks a new formula wisp; the wisp root's description is the **formula's own** `description`:
  `internal/formula/compile.go:317` (`rootDesc := f.Description`), applied at `:339`. build-base /
  compound-build define no `desc`/`title` var, so the root can only ever be formula boilerplate.
- The target bead is linked **only** by metadata + a blocking dependency — no text copy:
  `internal/sling/sling_core.go:353-369` ("Route the SOURCE bead, not wispRootID"),
  `internal/molecule/molecule.go:302-305` (`DepAdd(attachBeadID, RootID, "blocks")`).
- The source bead ID is passed as the `issue` var (`internal/sling/sling.go:1056-1058`) but **no
  build-base/compound-build template references `{{issue}}`**, so it never enters any prompt.
- The role-worker contract bars the worker from reading the parent/root bead or sibling repos:
  `template-fragments/gc-role-worker.template.md:9-12`, `:18-20`.

## Impact (observed)
A request to build a **GCS-backed** static-hosting MCP server (instructions in bead `shm-r52`,
"follow ~/rigs/ultimate-brain-mcp patterns") produced a **local HTTP static server in TypeScript** with no
Google Cloud at all — because the brainstorm only ever saw the generic wisp root `shm-68f` + `README.md`.

## Suggested fix
- When `--on`'s source bead has a non-trivial description, **warn** (or offer to copy it into
  `context_path`/the wisp root) so instructions aren't silently dropped; and/or
- auto-seed `context_path` (or a requirements stub) from the source bead's description.
- At minimum, document prominently that `--on` does not carry the bead's text and that the spec must go via
  `context_path` / `requirements_path`.

## Workaround
Pass the instructions + references explicitly: `--var context_path=<dir>` (and/or
`--var requirements_path=<doc>`); do not rely on the `--on` bead's description.

tdupu:
Downstream report — we hit a possible over-fire of this issue's fix, found independently before we knew #3681 existed.

[...]

PR #5451, opened 2026-08-20
fix(sling): fire the dropped-instructions hint only where its remedy exists (ga-awchvd)

quad341:
## Summary

`gc sling <target> <bead> --on <formula>` warns that the bead's description is
not carried into the formula's rendered context, and tells the caller to pass
`--var context_path=<dir>` or `--var requirements_path=<doc>`. The hint fires on
one condition — the caller passed neither var — and never checks whether the
formula being attached can consume either.

Where the formula consumes neither, that remedy is inert. Nothing rejects an
undeclared `--var` anywhere on the path: `buildSlingFormulaVars` accepts any key
it is given, `CollectVarValidationErrors` (`internal/formula/parser.go`) walks
only the formula's *declared* vars and so never sees an extra one, and
`stampFormulaVars` records the value as `gc.var.<name>` metadata that no
template reads. The var is accepted, stored, and rendered nowhere.

So a caller who follows the advice gets a **silenced warning and a byte-identical
rendered context**. That is worse than staying quiet, because a clean log reads
as fixed.

This is visible in the existing suite: `TestSlingAttachFormulaNoWarningWhenContextPathProvided`
asserted the hint goes quiet once `context_path` is passed — against the shared
`code-review` fixture, which declares no vars at all.

#3681 was reported against `build-base`/`compound-build`, which *do* declare
these vars, so the advice was sound there. The hint was simply never conditioned
on that being true.

## Fix

Gate the hint on its remedy being reachable. `formulaCanReceiveBeadInstructions`
reports whether the attached formula consumes either var — declared under
`[vars]`, or referenced as `{{var}}` in the formula description or a step, since
`Substitute` renders undeclared vars too.

The formula is resolved with `graphv2.LoadFormula` (parser-only, and
`Parser.Resolve` merges `extends` parents, so inherited declarations count). It
is never compiled, so the compile-once rule earlier in this file is untouched. An
unresolvable formula means an unknowable shape and stays silent — an unactionable
hint is the defect being avoided.

Correct in both directions: still fires for a formula built to receive the
instructions, silent for one that cannot use them.

## Why this matters in practice

In the city this was found in, **no formula declares or references either var** —
zero occurrences across every pack, every formula layer, and every template. All
eight formulas in the automated dispatch table are in that set, so the advisory
fired on every formula-dispatched bead in the fleet with no reachable remedy.

Worth stating explicitly, because the advisory's wording invites the opposite
conclusion: this is a statement about the **rendered context only**. It does not
mean the agent cannot see the bead. Seven of those eight formulas resolve the
content bead from the wisp wrapper's `gc.var.issue` and run `bd show $ROOT_ID`;
the eighth reads its bead from `bd list ... --json`, which carries the
description. The bead's own text reaches the agent either way. Added that
caveat to the function's doc comment so the next reader does not mistake the
warning for lost work.

## Test plan

- `TestSlingAttachFormulaNoWarningWhenFormulaCannotReceiveContextVars` — new; the
  RED case. Confirmed failing before the fix (hint present), passing after.
- `TestFormulaCanReceiveBeadInstructionsDetectsTemplateReference` — new; covers
  nil, the consumes-neither shape, and detection via a `{{requirements_path}}`
  step reference.
- `TestSlingAttachFormulaWarnsWhenBeadDescriptionDropped` and
  `TestSlingAttachFormulaNoWarningWhenContextPathProvided` — repointed at a
  fixture that declares `context_path`. Both previously ran against a fixture
  that declares no vars, so under the new gate they would have passed
  vacuously; they now assert what their names claim. This is the whole blast
  radius — one behavioral test changed, and it is the one that encoded the
  ungated behavior.
- `go vet ./...` clean; full `go test ./...` sweep green (40,332 pass / 0 fail).

## Known limitation, deliberately left

The consumes-the-var check walks declared `[vars]` plus top-level step title and
description text. It does not read `description_file` bodies (resolved at compile
time, not load time) or recurse into nested `children` / `loop.body` steps. Both
would be false negatives — a *missing* advisory, never a wrong one — and the
declared-`[vars]` path, which is the documented way a formula receives a var,
covers them regardless of nesting. Happy to extend the text scan if a reviewer
would rather have it; it seemed like speculative reach for shapes no formula in
this city uses.

### DIFF
--- a/internal/sling/sling_core.go
+++ b/internal/sling/sling_core.go
@@ -433,11 +433,57 @@ func rootOnlyVaporPourHint(formulaName string, recipe *formula.Recipe) string {
 func slingOnFormula(opts SlingOpts, deps SlingDeps, querier BeadQuerier, beadID string, result SlingResult) (SlingResult, error) {
 	result, err := attachFormulaToBead(opts, deps, querier, beadID, opts.OnFormula, "on-formula", "formula", result)
 	if err == nil {
-		if hint := attachedBeadInstructionsDroppedHint(querier, beadID, opts.Vars); hint != "" {
+		if hint := attachedBeadInstructionsDroppedHint(querier, beadID, opts.Vars, opts.OnFormula, SlingFormulaSearchPaths(deps, opts.Target)); hint != "" {
 			result.BeadWarnings = append(result.BeadWarnings, hint)
 		}
 	}
 	return result, err
+}
+
+// beadInstructionCarrierVars are the two formula vars #3681 tells a caller to
+// pass so a bead's own instructions reach the formula's rendered context. One
+// list, read both by the "did the caller pass one" check and by the "can this
+// formula receive one" check, so the two can never drift apart.
+var beadInstructionCarrierVars = [...]string{"context_path", "requirements_path"}
+
+// formulaCanReceiveBeadInstructions reports whether f actually consumes either
+// of beadInstructionCarrierVars — by declaring it under [vars], or by
+// referencing {{var}} in the formula description or any step (Substitute
+// renders any supplied var, declared or not, so a formula may consume one
+// without declaring it).
+//
+// This gates the #3681 hint, and it is load-bearing rather than cosmetic.
+// Nothing rejects an undeclared --var anywhere on the path: buildSlingFormulaVars
+// accepts every key it is given, CollectVarValidationErrors
+// (internal/formula/parser.go) walks only the formula's DECLARED vars and so
+// never sees an extra one, and stampFormulaVars records the value as
+// gc.var.<name> metadata that no template reads. Against a formula consuming
+// neither var the advertised remedy is therefore inert: passing the var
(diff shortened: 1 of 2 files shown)


quad341:
Closing as superseded by #5463, which landed first and fixes the same defect.

To record what was different, because it is not simply a duplicate: this PR gated the dropped-instructions hint on **whether the attached formula actually declares the vars**. #5463 gates it on **`gc.var.issue` / Route B being live**. Same symptom, two different predicates — and #5463's rewrite of `internal/sling/sling_core.go` is exactly where the conflict falls, confirmed with `git merge-tree` against `origin/main`.

The formula-declares-the-vars predicate is arguably the more precise of the two, so noting it here rather than losing it: if the false hint ever resurfaces in a case Route B does not cover, that is the approach to reach for.

No reflection on the work — 88/88 green, and it was overtaken rather than wrong.

PR #5452, opened 2026-08-21
test(integration): replace hardcoded Dolt startup waits

quad341:
## What this changes

Dolt integration tests now use named startup budgets instead of duplicated 15-second literals. The external-host wiring test has enough time for cold initialization under host load, and timeout failures report the actual configured budget.

A small AST-based regression test requires both startup call sites to keep using the named constants, preventing the literals and their diagnostics from drifting apart again.

## Review notes

- This is test-only; no production runtime behavior or public API changes.
- `bdInitTimeout` and `doltServerReadyTimeout` are hang-detector ceilings, so passing tests do not become slower.
- The main review surface is the constant choice and the static guard's identification of the two call sites.

## Test plan

- [x] Container-backed focused Dolt integration tests under the pinned image
- [x] Full sharded local CI-equivalent sweep; host-owned failures are attributed in the gate record
- [x] `make test-ci-policy`
- [x] `go vet ./...`
- [x] Release gate: [`release-gates/ga-1iowjo-dolt-integration-timeout-constants-gate.md`](release-gates/ga-1iowjo-dolt-integration-timeout-constants-gate.md)

### DIFF
--- a/release-gates/ga-1iowjo-dolt-integration-timeout-constants-gate.md
+++ b/release-gates/ga-1iowjo-dolt-integration-timeout-constants-gate.md
@@ -1,0 +1,97 @@
+# Release Gate: Dolt integration timeout constants
+
+Gate date: 2026-08-20
+
+Deploy bead: `ga-1iowjo`
+
+Review bead: `ga-thuouz`
+
+Original reviewed source: `09c63aa404929089238c7499a762001d8892f999`
+
+Gated source after bounded rebases: `fcf1009a8bdad843d6f6f0217178edfa80dc36fc`
+
+Base checked: `origin/main@4f4a37b28c9cfaa1ebe1c587576b69663a47f078`
+
+Merge base: `3218c505668fb043a89531a4ec1299d15a7e68d7`
+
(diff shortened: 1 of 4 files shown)


quad341:
@quad341 — thanks for this one. The part I appreciate most isn't the constant extraction, it's `TestDoltConfigTimeoutsUseNamedConstants`. Asserting on the *identifier* rather than the value means the lock survives future retuning of the budgets, and the arity checks fail loudly with "update this test's scoping" instead of quietly passing over a changed file. That's a tripwire that will still be doing its job a year from now. Writing the ga-gajll3 drift mechanism into the constant's own comment is the other half of it — the next person to touch these budgets will know why the two aren't shared.

Leaving the new file untagged so the guard runs in the cheap lane, rather than only behind `-tags integration`, was the right instinct too.

Nothing needed from you. The red `CI / required` is an unrelated flake — `TestControlBytesAreStripped` in `internal/worker/adapters/zcode`, a package your diff doesn't touch, and that shard doesn't even run `test/integration`. It passes on your head when I run it. I'll re-run the shard on my side and merge. Your commits land as-is.


---

PR #5454, opened 2026-08-21
Stabilize the Python fallback SIGTERM regression test

quad341:
## What this changes

The `run_bounded` Python fallback regression test now distinguishes a child that never finished installing its SIGTERM handler from a real signal-delivery failure. The inline Python child publishes a readiness marker immediately after registering the handler, and the test retries up to three times when host load prevents that marker from appearing within the production timeout.

Once readiness is observed, a missing SIGTERM marker remains a hard failure. Exhausting all readiness attempts also fails with a diagnostic that identifies host load and the missing readiness marker.

## Review notes

- This is test-only; `run_bounded` runtime behavior is unchanged.
- The production-input 1-second timeout and 2-second termination grace are unchanged.
- The readiness handshake is deliberately placed after `signal.signal(...)` and before the child sleeps.

## Test plan

- [x] Changed regression test passes five consecutive runs with Python 3 available and no skips.
- [x] Full `examples/bd/dolt` package passes with 352 PASS / 0 FAIL / 0 SKIP.
- [x] Repository policy lane and `go vet ./...` pass.
- [x] Release gate: [`release-gates/ga-40p955-python3-fallback-readiness-gate.md`](release-gates/ga-40p955-python3-fallback-readiness-gate.md)

### DIFF
--- a/examples/bd/dolt/runtime_bounded_test.go
+++ b/examples/bd/dolt/runtime_bounded_test.go
@@ -80,9 +80,23 @@ func runRunBoundedUnderPython3Fallback(t *testing.T, childScript string, extraEn
 // `subprocess.run(..., timeout=...)` implementation (bare
 // `process.kill()` on expiry) the handler never runs and the marker
 // is never written.
+// The 1s bound below is production input to run_bounded, not this test's
+// own hang-detector timer (TESTING.md: a timeout the test feeds INTO the
+// code under test is never migrated to a hang budget), so the blanket
+// >=10s "Test deadline rule" for timers racing an exec start does not
+// apply to it — enlarging it would hide the very race gascity#4823 exists
+// to catch. run_bounded starts its timer at Popen, so python3 interpreter
+// startup is counted against the bound, and a starved host can starve the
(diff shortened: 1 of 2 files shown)


quad341:
@quad341 — the READY-marker handshake is exactly the right shape for this. Splitting "the child never got far enough to prove anything" from "the child was ready and still didn't get SIGTERM" is what makes a retry honest, and putting the marker write between `signal.signal(...)` and the sleep is the detail that makes it work: once READY exists, the handler is provably installed, so the SIGTERM assertion stays as strict as it ever was.

I especially appreciate that you left the 1s bound and 2s grace alone and wrote down *why* in the comment. Widening the bound was the easy fix and would have quietly retired gascity#4823's regression guard. I checked the TESTING.md rule you cite and it backs you word for word — a timeout fed *into* the code under test isn't a hang budget. I also read `runtime.sh` and confirmed the Popen-then-wait ordering that makes interpreter startup count against the bound, which is the starvation you're actually catching.

Merging as-is, nothing needed on my side. Thanks for fixing the flake properly instead of papering over it.


---

PR #5456, opened 2026-08-21
fix(session): reap pre-boot session beads when the runtime server is absent

### LINKED ISSUE #5455: bug: an orphaned session bead deadlocks its pool after a host reboot (reap fail-safes on absent tmux server)
## Summary

After an unclean host reboot, a single orphaned pool session bead can deadlock
its entire pool indefinitely. The runtime (tmux) server is gone, so every reap
path fail-safes on the resulting `PartialListError`, while the orphaned bead
keeps holding the pool's alias. Nothing reaps it, nothing spawns a successor,
and the city sits silent until a human force-closes the bead by hand.

## Environment

- gc 1.4.1, build `8c73625b974ce8d3d68d54ab42062e6247c47036`
- macOS (darwin/arm64), tmux runtime, launchd-managed supervisor
- Observed live on 2026-08-20 after a power outage rebooted the host at 17:28:45 local

## Repro

1. Run a city with a pool agent (`max_active_sessions >= 1`) and let it start at
   least one pool session, so an open session bead carries `session_name`
   metadata for a live tmux session.
2. Hard-reboot the host (power loss), or otherwise kill the tmux server without
   letting the supervisor shut sessions down cleanly.
3. Start the supervisor again.

## Actual

The pool never recovers. The supervisor logs this on every tick (1328
occurrences in our incident window):

```
gc supervisor: pool death check skipped due to partial session listing: tmux server unreachable: no tmux server running
```

The orphaned bead (`cr-wisp-djihx`, holding alias `qwen38-1`) stays open and
keeps the alias. `cleanupDeadRuntimeSessionCorpses` cannot help: it lists
runtime sessions, gets nothing back because the server is absent, and returns
early at `if len(visible) == 0`. Since the alias is still claimed, each
reconciler tick's successor fails `EnsureAliasAvailable` with
`ErrSessionAliasExists` — the same accumulation described in #2437.

Recovery required a human to force-close the bead.

## Expected

A session bead whose runtime provably cannot exist is reaped, releasing the
alias so the pool spawns a successor without operator intervention.

## Why the obvious fix is wrong

Treating an absent server as an authoritative empty listing would revert
#4082: a single-tmux outage is a failed observation, not proof that zero
sessions exist, and reading it as "everything died" would tear down healthy
sessions during a brief blip.

## Proposed fix

Add independent proof of death rather than weakening the fail-safe. A tmux
server never survives a reboot, so a session bead created *before the current
host boot* cannot have a live runtime, regardless of what the runtime backend
can or cannot observe right now.

- `internal/hostboot`: report the host boot instant (darwin `kern.boottime`,
  linux `/proc/stat` `btime`); unsupported platforms return an error.
- `internal/runtime`: `PartialListError` gains a `ServerAbsent` flag so callers
  can distinguish "no server at all" from a server that answered partially. The
  #4082 fail-safe is unchanged for every other caller.

  Absence is a flag read by a direct type assertion, never a sentinel matched
  through the error chain. A composite provider joins its backends' errors (and
  returns a bare join when every backend fails), so any traversing check would
  report absence whenever *any* backend was absent — including when a sibling
  backend is healthy, or merely failed for an unrelated reason, while still
  holding live sessions. Only a single backend speaking for itself can assert
  absence; wrapping degrades to "not absent", the fail-safe answer.
- `cmd/gc`: when the server is absent, reap open session beads created before
  boot.

Conservative in all three directions that matter:

- beads created *after* boot are left alone (a server that is merely down
  cannot distinguish live from dead);
- a partial listing from a *live* server keeps the existing fail-safe
  untouched;
- if the boot instant is unavailable, nothing is reaped.

Eligibility is deliberately narrow. Only sessions whose **raw** metadata state
is active/asleep/awake qualify — `normalizeInfoState` folds `drained` into
`StateAsleep`, so the normalized value cannot tell a sleeping session apart
from one the drain path already finished with, and an absent or unrecognized
raw state is not reapable at all. Named sessions, drain-tracked beads, and
pending-create claims are skipped, matching the surrounding sweep; manual
sessions are left to their operator; and non-tmux transports are skipped
because local boot time proves nothing about a remote runtime.

Known residual: a wall-clock rollback (NTP step) after boot could make a
post-boot bead look pre-boot. The window is small and the consequence is
reaping a session that is already unreachable, but it is a real limitation of
using wall-clock creation stamps as the proof.

Tests cover the reap, both fail-safe directions, the post-boot boundary, the
state/transport/manual guards, the drained-normalization trap, and both
multi-backend merge shapes (healthy sibling, and total failure).

PR accompanies this issue.

jacobhausler:
Fixes #5455.

After an unclean host reboot the tmux server is gone, so ListRunning returns PartialListError and every reap path fail-safes — while the orphaned pre-reboot session bead keeps its alias, blocking every successor (`ErrSessionAliasExists` accumulation) and deadlocking the pool until a human force-closes the bead.

Treating an absent server as an authoritative empty listing would revert #4082, so this adds an independent proof of death instead: a tmux server never survives a reboot, so a session bead created before the current boot instant cannot have a live runtime.

- `internal/hostboot`: host boot instant (darwin `kern.boottime`, linux `/proc/stat` btime); unsupported platforms error → nothing reaped.
- `internal/runtime`: `PartialListError.ServerAbsent` flag, read by direct type assertion — deliberately non-unwrapping so a bare `errors.Join` from multi-backend merges can't smuggle an inner absent flag past a healthy sibling backend holding live sessions.
- `cmd/gc`: reap pre-boot beads only when the server is absent; only raw `active`/`asleep`/`awake` states qualify (drained/empty/unknown excluded); manual sessions, drain-tracked beads, named sessions, pending-create claims, and non-tmux transports all skipped.

Documented residual: an NTP rollback after boot could make a post-boot bead look pre-boot (not considered practically exploitable).

Adjacent to #5442/#5447 (wake-failure accrual) — this covers the reboot-orphan case that accrual alone cannot, since the dead session's runtime never returns to fail again.


jacobhausler:

This PR is carrying a measured defect from the Westlands city: when the runtime (tmux) server is absent at startup, pre-boot session beads remain open instead of being reaped, and later reconciles see a pool that has no matching live sessions.

**Retirement condition for the local patch:** a released startup/reconcile path reaps only proven pre-boot orphans and leaves live sessions untouched. Until this lands, Westlands carries a local patch over released gc. Tracking this PR as the upstream vehicle.

quad341 (approved):
@jacobhausler — this is careful work and the shape of the argument is right: rather than reinterpreting an absent tmux server as an empty listing (which would have walked straight back into #4082), you brought an *independent* proof of death. The detail I want to call out is the non-unwrapping `IsRuntimeServerAbsent`. Choosing a bare type assertion so a merged `errors.Join` can't smuggle one backend's absence past a healthy sibling still holding live sessions is exactly the failure mode that would have made this a #4082 rerun — you closed it, commented why, and pinned it with two tests. Reading the raw metadata state because `normalizeInfoState` folds `drained` into `asleep` is the other trap you caught before it caught us.

CI is red on our house rules, not your code: `internal/hostboot` needs an untagged `testenv_import_test.go`, and we ship a generator for it. I'll take that on my side.

While I'm in there I'll make two small narrowings: skip `asleep` beads parked with `sleep_reason=city-stop` (a clean `gc stop` parks the pool that way, and `cityStopPoolBeads` revives it on the next start), and treat a malformed wake/create timestamp as unproven rather than ignoring it. Your fix for the actual deadlock is untouched. Shout if the wider sweep was deliberate.


---

PR #5457, opened 2026-08-21
Return HTTP 503 for retryable session store conflicts

quad341:
## What this changes

Session suspend requests that exhaust the Dolt store's bounded serialization-conflict retry now return HTTP 503 (`store_conflict`) instead of leaking an undeclared HTTP 500. Clients can distinguish the transient condition and retry safely.

The binary Huma workflow now tolerates a bounded run of those declared 503 responses while suspending a session, then continues through the existing asynchronous message and submit checks.

## Review notes

- The implementation reuses one store-conflict classifier for both delete retries and HTTP error mapping.
- This changes response status mapping only; it does not add or reshape routes or OpenAPI schemas.
- The integration retry is bounded to five seconds and honors `Retry-After` when present.
- The checked test-resource ledger increases the fixed-sleep audit count by exactly one for that bounded retry.

## Test plan

- [x] Exercise a retry-exhausted serialization conflict through the suspend handler and assert the declared 503 response.
- [x] Run the binary Huma session create, suspend, async message, and submit workflow with container support enabled.
- [x] Run API tests, resource-census checks, policy checks, `go vet ./...`, and `make dashboard-ci`.
- [x] Run the documented full local CI union and resolve every diff-owned test by name.
- [x] Release gate: [`release-gates/ga-qx0dd0-suspend-serialization-conflict-gate.md`](release-gates/ga-qx0dd0-suspend-serialization-conflict-gate.md)


quad341:
@quad341 — thanks for this one. What I appreciate most is that you traced the actual failure instead of just widening a status map: lifting the verbatim conflict text out of the `rest-full-2-of-8` log into `retryExhaustedConflictStore`, and asserting `suspendUpdates != 0` before asserting the status, means the test can't pass for the wrong reason.

I checked the thing I most expected to be a bug — that you bumped only the `scope = "all"` fixed_sleep row and left `untagged`/`small_debt` alone — and you're right: `huma_binary_test.go` is `//go:build integration`, so it's outside those scopes. Nice call.

The only thing in the way is a conflict, and it's ours to absorb. `main` has moved that all-scope baseline from 453 to 477 since you branched, so your +1 collides in the three ledger files; your Go changes merge clean. I'll rebase, re-resolve to 478/172, and re-run the policy tests plus the Huma binary integration test — that last one lives in `rest-full`, which was skipped on your PR, so it hasn't actually run in CI yet. Your commits and credit stay intact.

Two small follow-ups I'll file myself rather than ask you to chase: the `40001` substring match is looser than it needs to be now that it gates every session store error, and the new 503 doesn't send `Retry-After` the way the agent-create 503 does — so the client-side parsing you added is currently dead code waiting on the server.


---

Issue #2846, opened 2026-05-31
Dolt compactor permanently quarantines a benign concurrent-writer race (gain+drift), starving GC


Wldc4rd:
## Summary

`gc dolt compact` (the `mol-dog-compactor` order) can write a **permanent** integrity quarantine marker for a **benign concurrent-writer race**, after which that database is skipped on every future run and its commit graph + on-disk size grow without bound. This is the exact "quarantine marker blocks all future GC of the db … production memory-exhaustion bug" the code comments call out (`examples/dolt/commands/compact/run.sh:33-34`, `:1722-1723`).

The post-flatten "concurrent writer proven" gate (`run.sh:1738-1745`) relies on HEAD movement, but misses the case where the racing writer's commit is **absorbed into the flatten commit** — leaving no HEAD fingerprint — so a benign race is misclassified as possible corruption and hard-quarantined.

## Real incident (this is how I found it)

A healthy, idle-ish city's town DB (`hq`):

- **2026-05-29** scheduled run: `order.completed` (success).
- **2026-05-30** scheduled run: `order.failed`; quarantine marker written:
  ```
  db=hq
  reason=post-flatten table value hash changed with row-count increase
  created_at=2026-05-30T18:28:36Z
  ```
- Every 24h run after that: `db=hq integrity quarantine marker exists — manual intervention required before compaction or GC`. `hq` grew to **16,822 commits / 3.9 GB**; `psa` (below threshold) was unaffected.
- After manually clearing the marker and re-running `gc dolt compact`: **16,822 → 2 commits, 3.9 GB → 898 MB in 23 s, zero data loss** (verified: `issues` row count identical before/after; all access paths intact).

So it was unquestionably a benign concurrent-writer race — the precise false positive the gate is meant to absorb — yet it produced a permanent, GC-starving quarantine.

## Root cause

`verify_counts` (`run.sh:702`) correctly flags `gain_hash_drift` when a table gained rows **and** its value hash changed. The downgrade-to-defer gate then requires a **HEAD-proven** writer (`run.sh:1738-1745`):

1. `head_before_reset != head` — a writer landed in the pre-flight→pre-reset window, **or**
2. `post_verify_head != flatten_head` — a writer landed during/after verify.

[...]

PR #2855, opened 2026-05-31
fix(dolt): prove gain+drift row preservation directly (Option A, #2846)

Wldc4rd:
Switches the post-flatten gain+drift handling to the strictly-safe **Option A** from #2846 (this PR previously implemented the bounded-retry Option B), rebased onto current `main`.

## Problem (full write-up in #2846)

`gc dolt compact` permanently quarantines a **gain+drift** post-flatten result (a table gained rows *and* its value hash changed) when no concurrent writer is HEAD-provable. On a busy primary DB the racing writer is usually **absorbed into the flatten commit** (no HEAD fingerprint), so the benign race is misclassified as possible corruption and the permanent quarantine then **blocks all future GC of that db** — the memory-exhaustion failure the code's own comments call out. Real incident: town `hq` stuck at **16,822 commits / 3.9 GB**; after clearing the marker one compaction went **16,822 → 2 commits, 3.9 GB → 898 MB in 23 s, zero data loss**.

## What this does (Option A — prove preservation directly)

HEAD movement is only a proxy for the actual safety property (*"pre-flight rows remain reachable"*). When `verify_counts` flags gain+drift as the **only** failure category, `gain_drift_is_additive_only` diffs the pre-flight snapshot HEAD against the flatten commit for each gained+drifted table:

```sql
SELECT COUNT(*) FROM DOLT_DIFF('<head>', '<flatten_head>', '<table>') WHERE diff_type <> 'added'
```

- **0 non-added rows** (only `added`) → every pre-flight row is preserved → **defer** via the existing `defer_writer_race_after_flatten`, retry next run.
- **Any removed/modified row, or any probe failure** → fail closed → **quarantine** unchanged.

This proves reachability instead of inferring it from HEAD movement, closing the absorbed-writer gap. `row_decrease`, `same_count_hash_drift`, `table_list_change`, and probe failures still quarantine unchanged. **No GC ever runs on an unverified flatten.** The whole-DB value-hash path carries an exclusion comment (near-unreachable; the per-table proof covers real appends).

## Files

- `examples/dolt/assets/scripts/compact-gain-drift-proof.sh` — `gain_drift_is_additive_only`
- `examples/dolt/commands/compact/run.sh` — collect gained+drifted tables in `verify_counts`; run the proof before the gain+drift quarantine; db-level exclusion comment
- `test/dolt/compact_gain_drift_proof_test.sh` — classification unit test (10 assertions)

## Tests / verification

[...]

PR #2564, opened 2026-05-24
fix(dolt): defer compaction when a writer races the flatten, don't quarantine

iwata-1116:
## Problem

`gc dolt compact` flattens commit history, then `verify_counts` compares post-flatten **live** tables against the **pre-flight snapshot**. A normal concurrent MVCC writer that commits during the flatten window produces a legitimate row-count increase **plus** table/db value-hash drift. `verify_counts` cannot tell that apart from corruption, so it writes a blocking integrity quarantine marker.

That marker makes **all** future compaction of the db early-return. With `auto_gc` disabled (the documented config), compaction is the **only** GC path — so blocking it lets Dolt's in-memory NBS chunk-journal range index (`go/store/nbs/journal_writer.go`, reset only by GC or restart) grow unbounded until the host runs out of memory.

This was hit repeatedly in a managed city whose beads/mail workload commits frequently. `dolt_log` forensics over the flatten windows showed only ordinary `beads`/`bd:` commits — **no corruption**; the quarantine was a false positive. The downstream effect is severe under WSL2 (RAM cap + `swap.vhdx` thrash, no OOM-kill → multi-minute freeze).

## Fix (approach chosen)

Extend the script's own stated principle — the algorithm comment already says it should *"require HEAD to remain stable across a bounded retry loop,"* and HEAD-stability/defer is already implemented for the **remote-push** path — to the **local flatten verify**:

- Capture HEAD immediately **before** the mutating `DOLT_RESET`/`DOLT_COMMIT`, and again **after** `verify_counts`.
- A writer is **proven** (and distinguished from the flatten's own commit) when the pre-reset HEAD differs from the stable pre-flight HEAD, **or** the post-verify HEAD has moved past the flatten's own commit. Comparing against the flatten's own commit (not the raw pre-flight HEAD) is what avoids mistaking the flatten itself for a writer.
- Only in that **proven-writer** case is the ambiguous gain+hash-drift result downgraded from a blocking quarantine to a clean skip ("deferring, will retry next run"). Next scheduled run retries.

**All other integrity signals still quarantine, unchanged:** row-count decrease; same-count value-hash drift (per-table and db); table-list change; any probe failure; and gain+hash-drift when HEAD was **stable** (a genuine anomaly). A failed HEAD probe leaves the race unproven and falls through to quarantine (safe default).

### Alternatives considered
- **(B) Snapshot isolation** — flatten a pinned commit and verify against it. Correct in principle but a larger redesign of the compaction concurrency model; deferred.
[...]

PR #5460, opened 2026-08-21
fix(dolt): defer compact quarantine for a proven concurrent-writer same-count in-place UPDATE (ga-ewo6j)

vbtcl:
## Problem

`gc dolt compact`'s post-flatten integrity check has three defer paths that
downgrade a blocking quarantine to skip-and-retry when a concurrent writer is
proven:

- gain+drift, HEAD-proven
- gain+drift, additive-only (Option A, #2846)
- row-decrease, HEAD-proven

All three explicitly require `verify_counts_saw_same_count_hash_drift != 1`, and
there is **no** defer path for the same-count case. A concurrent in-place UPDATE
that lands in the flatten verify window — the row count is unchanged but a
table's value hash drifts — therefore always falls through to a **hard
quarantine**.

On a high-write DB (mail/coordination, where `bd:update` and mail-read commits
land every few seconds) such an UPDATE lands on ~every flatten, so the DB
re-quarantines on ~every run. A quarantine marker blocks **all** future GC of
the DB — the unbounded-bloat / memory-exhaustion failure the compactor header
explicitly warns about. Observed in production: a mail/coordination DB grew to
several GB on disk for ~2k rows and flooded the operator inbox with one
`compact-quarantine` alert per run.

## Fix

Add a same-count defer path that fires **only when both** hold, else fails
closed and quarantines:

1. a concurrent writer is HEAD-proven (`writer_race_detected=1`); **and**
2. `same_count_drift_is_writer_explained` proves — via `DOLT_DIFF` of the
   pre-flight snapshot HEAD against the flatten commit for every
   same-count-drifted table — that rows were **modified** with **none of the
   pre-flight rows removed**.

Any removed pre-flight row, an unexplained/empty diff, or a probe failure fails
closed and quarantine stands.

### Why no absorbed-writer fallback (unlike gain+drift)

gain+drift can be proven benign with no HEAD movement because "additive-only"
leaves every pre-flight row **unchanged** as an anchor. A same-count in-place
modification has no such anchor — it is byte-for-byte indistinguishable from
corruption without an independent writer signal — so a **stable-HEAD** same-count
drift remains the corruption signal and still quarantines. The frequent
production case (committed `bd:update`/mail writes move HEAD) is covered.

## Changes

- **New** `examples/bd/dolt/assets/scripts/compact-same-count-drift-proof.sh` —
  `same_count_drift_is_writer_explained`, mirroring the gain-drift Option A
  proof helper. Ships via the existing `all:assets` embed.
- `examples/bd/dolt/commands/compact/run.sh` — source the new helper; track
  `verify_counts_same_count_drift_tables`; add the same-count defer block and an
  accurate fail-closed diagnostic.
- `examples/bd/dolt/dog_exec_scripts_test.go` — `DOLT_DIFF` content-probe mock
  (keyed on the `diff_type = 'removed'`/`'modified'` clauses so it never shadows
  the gain-drift/first-commit probes) plus three tests.

## Tests (TDD)

- `TestCompactScriptDefersSameCountHashDriftWithProvenWriter` — the fix: proven
  writer + benign in-place UPDATE ⇒ defer (exit 0, pending-GC marker, no
  quarantine, no GC).
- `TestCompactScriptQuarantinesSameCountHashDriftWhenPreflightRowRemoved` — fail
  closed on a removed pre-flight row.
- `TestCompactScriptQuarantinesSameCountHashDriftWhenDiffProbeFails` — fail
  closed on a diff-probe failure.
- Pre-existing stable-HEAD same-count corruption case still quarantines; full
  `examples/bd/dolt` suite green; `go vet` clean.

### DIFF
--- a/examples/bd/dolt/assets/scripts/compact-same-count-drift-proof.sh
+++ b/examples/bd/dolt/assets/scripts/compact-same-count-drift-proof.sh
@@ -1,0 +1,70 @@
+#!/bin/sh
+# compact-same-count-drift-proof.sh — in-place-modification proof for the
+# post-flatten same-count table value-hash drift case (ga-ewo6j).
+#
+# verify_counts flags same_count_hash_drift when a table's value hash changes
+# but its row count does NOT — an in-place row modification. On a high-write DB
+# the beads/mail workload commits such UPDATEs (mail read/archive, bead
+# close/status change) every few seconds, so one lands inside the flatten verify
+# window on ~every run. Hard-quarantining it blocks ALL future GC of the DB and
+# is the production memory-exhaustion failure the compactor header warns about.
+#
+# Unlike the gain+drift case, a same-count in-place modification cannot be proven
+# safe by "additive only": the pre-flight row's value legitimately changed, so
+# there is no unchanged-row anchor and no absorbed-writer DOLT_DIFF proof
+# analogous to Option A (compact-gain-drift-proof.sh). This proof is therefore
+# only ever a SECOND gate, applied by the caller AFTER a concurrent writer is
+# HEAD-proven (writer_race_detected=1): it confirms the drift has the SHAPE of a
+# benign in-place UPDATE — every same-count-drifted table's pre-flight..flatten
+# content diff modifies rows and removes NONE of them. Any removed pre-flight
+# row, a diff that fails to explain the drift, or a probe failure fails closed
+# and falls through to quarantine.
+#
+# Depends on `query_single_cell` and `valid_table_name` from run.sh.
+
+# same_count_drift_is_writer_explained <db> <from_head> <to_head> <tables>
+# Returns 0 iff every listed table's <from>..<to> content diff removes zero rows
+# AND modifies at least one row (the drift is an explained in-place UPDATE).
+# Returns non-zero (fail closed) if either commit endpoint is missing, the table
+# list is empty, a table name is invalid, a diff probe fails or returns a
+# non-numeric result, any pre-flight row was removed, or no row was modified.
+same_count_drift_is_writer_explained() {
+  _sc_db="$1"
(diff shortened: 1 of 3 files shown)


csauer02-personal-user:
We hit this same race in a long-running deployment and independently built the same defense before finding this PR — posting what we learned in case any of it is useful to this landing rather than opening a third PR into this queue.

Two observations from our variant, which took a slightly different construction:

1. **Generalizing the existing helper instead of adding a second script.** `compact-gain-drift-proof.sh` already answers "did the flatten preserve every pre-flight row" for the additive-only case. We widened that helper's contract from *additive-only* to *no rows removed* via the same `DOLT_DIFF` you use here, rather than introducing `compact-same-count-drift-proof.sh` alongside it. Same mechanism, one fewer near-identical proof script to keep in sync — and the second consumer of the helper (run.sh:1384, from #5049) picks up the stronger contract for free. Happy to share the diff if that shape appeals.

2. **An executable falsification control you're welcome to lift.** Our test proves the defense is load-bearing by reverting only the prod scripts to current `main` and asserting the new defer tests then FAIL — i.e., the tests cannot pass by accident. Under `go test -race`, the concurrent-UPDATE scenario reproduces the drift deterministically through the fake-dolt harness. If it would help this PR's review, we can post the test bodies here.

For what it's worth, the defer-on-proven-writer direction matches every merged precedent on this file (#2350, #2416, #2564, #2855, #5049) — this looks like the right fix shape to us, and we'd rather see this land than maintain our variant.

sjarmak:
Thanks for digging into the concurrent-UPDATE writer race. Two PRs landed on it independently and we are consolidating on #5749.

Closing this in favour of that branch. The quarantine-deferral reasoning here holds up; it is the same call #5749 makes.

Issue #3604, opened 2026-06-19
bug: running-agent `context_pct` / canonical-handoff advisory may under-report context usage (transcript usage row can be a partial/streaming sample → advisory may never fire)


iwata-1116:
We think we may have found a case where the running-agent `context_pct` — and the
context-injection handoff advisory that reads it — under-reports actual context
usage, which would keep the advisory from ever firing. Filing it so it's at least
on the record; we haven't fully confirmed the root cause, so the framing below is
deliberately tentative.

#### What we observed

An agent whose actual context was roughly a third of a 1M window (~300k tokens)
showed a reported `context_pct` near 0, and the default-60% handoff advisory never
fired. The agent kept running until it degraded, with no advisory at any point.

#### Where it seems to come from (paths on `main` @ `ce2daff08`)

The percentage looks like it's re-derived from the transcript JSONL rather than from
a live counter:

- `internal/sessionlog/tail.go` — `ExtractTailMeta` scans the transcript tail and
  selects the most-recent assistant row with `input_tokens > 0` (`tail.go:303`),
  then computes `pct := totalInput * 100 / contextWindow` where
  `totalInput = input_tokens + cache_read_input_tokens + cache_creation_input_tokens`
  (`tail.go:328-332`).
- `internal/api/handler_agents.go:466` exposes that as `context_pct`.
- `cmd/gc/context_inject.go:171-172` reads it for the advisory / urgent thresholds
  (defaults 60 / 80).

Both the numerator and the denominator of that percentage are re-derived from the
model string and the transcript, and each looks like it can go wrong independently:

#### Hypothesis A — numerator (which usage row gets picked)

Assistant rows in the transcript may include partial / streaming usage samples — a
small `input_tokens` with the cache fields still at 0. Because the selection filter
is `input_tokens > 0` (not a check for a *complete* usage row), such a sample can be
the one picked, which would make `totalInput` tiny and `pct` ≈ 0 regardless of the
real context size. Reading only the tail (last ~64KB) might also miss the row that
carries the fullest usage figures. We suspect one of these is why the reported
percentage sat far below the real value, but we haven't instrumented it end-to-end.

#### Hypothesis B — denominator (window size from the model string)

`ModelContextWindow` (`internal/sessionlog/context.go:32`) returns the 1M window only
when the model string contains the literal `[1m]` tag, and otherwise falls back to
the family default (e.g. 200k for opus). So the denominator depends on that tag being
[...]

PR #5464, opened 2026-08-21
feat(config): make the context-pressure advisory configurable (closes #3810)

### LINKED ISSUE #3810: Make the context-pressure advisory a pluggable trigger (configurable levels, message, and audience)
## Summary

The context advisory from #3371 is useful and we agree with its framing — the PR calls it "the trigger signal for canonical handoffs." Our ask: make that trigger configurable per city and per agent, instead of fixed in Go.

## What we're seeing

Running persisted agents, two recurring patterns:

1. **They defer work and recycle early** once the advisory fires — which is the documented intent (the PR notes the live line "told the authoring agent to defer a newly-dispatched workstream to a fresh session — the intended behavior"). That's exactly why it should be tunable: different roles want different thresholds, not one fixed number for the whole fleet.
2. **They punt the recycle to the human** instead of doing it themselves. We suspect the wording is part of this — the urgent tier suggests `gc session reset` (no continuation note, not attended-safe) where `gc handoff` is the right per-session-type action. A one-line wording fix helps our defaults, but it surfaced the bigger question.

## The ask

The advisory is a trigger — "at **level X**, tell **agent Y** to do **Z**" — but X/Y/Z are all hardwired (env-only thresholds, every-agent audience, one message). That's more Gas Town than Gas City. Make it config-driven at global + per-agent scope (per-agent wins), env vars kept as a back-compat override:

- enable / disable per scope (this is the audience knob)
- advisory + urgent thresholds, and a context-window override
- the message text per tier, operator-authored

Zero-config keeps today's behavior. We'd fold in a fix for the under-report case (#3604) while reworking the read path.

## Open questions

- `text/template` for the message (consistent with prompt templates), or something lighter?
- Per-agent config home: the existing agent-defaults axis, or wait for PackV2 agent-defaults (#903)?
- Is per-agent enable enough for "audience," or do you want richer targeting?

## References

#3371 (the feature) · #3604 (under-report bug) · #903 (PackV2 agent-defaults) · #1465 (per-step recycle policy)

wbern:
Closes #3810.

## TL;DR

The context-pressure advisory's thresholds and message text are hardcoded in Go. This makes them configuration — enable/disable, both thresholds, a window override, and operator-authored per-tier message text — at city scope and per-agent scope, with per-agent winning.

**Zero-config behaviour is unchanged.** Env vars keep working as an override.

## Why

#3810 asked for exactly this, and the reasoning there holds: the advisory is a trigger — "at level X, tell agent Y to do Z" — but X, Y and Z were all fixed in Go. Different roles want different thresholds and different wording; a fleet-wide constant cannot serve them.

Two concrete cases this enables:

- Raising or lowering the band for a specific long-running agent without touching Go.
- Rewording the guidance per deployment — the right lever when the advisory drives a behaviour you want to change, since the alternative is a code change to a prompt string.

## Shape

```toml
[context_advisory]
enabled = true

[[context_advisory.tiers]]
threshold = 60
message = "Context {{.UsedK}}/{{.WindowK}} (~{{.Pct}}%). Steer toward a clean seam."

[[context_advisory.tiers]]
threshold = 80
message = "Context {{.UsedK}}/{{.WindowK}} (~{{.Pct}}%) — recycle now."
```

`message` is a `text/template` over `.Tokens`, `.Window`, `.Pct`, `.UsedK`, `.WindowK`, `.Threshold`. #3810 raised `text/template` as an open question; it's used here for consistency with prompt templates.

Per-tier `enabled` allows disabling a tier while retaining it in an inherited policy.

## Behaviour

```gherkin
Given no [context_advisory] configuration
When context crosses a threshold
Then the emitted guidance is byte-identical to before this change

Given a city-scope policy and a per-agent policy
When both define a tier
Then the per-agent policy wins

Given GC_CONTEXT_ADVISORY_PCT or GC_CONTEXT_URGENT_PCT is set
When a tier threshold is also configured
Then the env var still overrides, as before
```

## Why it's safe

Zero-config is the load-bearing guarantee, and it's pinned by a test asserting the legacy tier boundaries exactly: **59.9% → no tier, 60% → advisory, 80% → advisory, 80.1% → urgent**. Those four cases encode the previous `switch` semantics, including that the boundary sits in the *preceding* tier for escalation bands and that 60% is inclusive.

That test exists because an earlier iteration got the 60% boundary backwards and silenced the advisory at exactly 60%. Worth stating plainly for reviewers: the guarantee is the inclusive lower bound, not just "roughly the same".

## Scope

`internal/config/context_advisory.go` (new) plus its test; wiring in `compose.go`, `config.go`, `pack.go`, `patch.go`, `field_sync_test.go`; consumption in `cmd/gc/context_inject.go` and `cmd_nudge.go`; regenerated `city-schema` / `pack-schema` and `docs/reference/config.md`.

## Related

- #3810 — the request this implements.
- #3604 — advisory under-reporting; independent of this change, which does not alter how usage is measured.


quad341 (approved):
@wbern — thank you for this, and specifically for the boundary test. You didn't just assert "zero-config is unchanged" — you encoded the old `switch` as four cases (59.9 / 60 / 80 / 80.1) and called out in the PR body that the inclusive lower bound at 60% is the load-bearing bit because an earlier iteration got it backwards. That's what makes turning hardcoded behaviour into config reviewable at all. The no-target inject regression test is a nice catch too.

CI is red, but not on anything you got wrong. This repo has reflection-based "every `Agent` field must appear here" guards in two packages you had no reason to know about — `cmd/gc/pool_test.go` and `internal/migrate/migrate_test.go`. You found and wired the equivalent one in `internal/config`. I'll take the other two on my side, plus the production bits they're pointing at: `deepCopyAgent` is a hand-rolled literal rather than `Agent.Clone()`, and `gc migrate` needs to persist the field — otherwise per-agent config would quietly no-op for pooled agents and get dropped on migration. Good catches by the guards.

One follow-up I'll file separately rather than block on: the pack-override path applies `context_advisory` without running your `Validate()`. Fail-soft, and its own PR.

Merging with your commits intact once the wiring's in.


---

PR #5468, opened 2026-08-21
fix(session): idle-timeout an always-on Claude TUI via content pane scan (ga-07mi8)

vbtcl:
## Problem

An always-on heartbeat session (e.g. a city's deacon or mayor) can sit **awake but idle for hours** without ever being recycled.

The idle-timeout reconciler measures idleness from the coarse pane-activity clock — `GetLastActivity`, which on tmux reads `#{window_activity}`, advancing on **any** pane output. An idle Claude Code TUI continuously repaints its status line, spinner, and cursor, so the activity clock stays perpetually fresh and `now - lastActivity > idle_timeout` never becomes true. `idleTracker.checkIdle` returns `false` on every tick, `TimerFacts.Triggered` stays false, and `DecideIdleTimeout` returns `TimerActionNone` forever — the always-on session is never recycled. The same `GetLastActivity` blind spot also silently defeats the progress-stall recycle and the config-drift defer.

### Forensics (session-reconciler-trace)

A city deacon sat `state=awake`, `sleep_reason=""` across **1854 consecutive reconciler ticks (~7h)** and emitted **zero** idle-timeout decision records. It was not a restart/circuit-breaker problem — the session never stopped. Only a coarse ~7h boot-check watchdog eventually nudged it awake. A sibling agent that *does* carry a `max_session_age` backstop did not stall.

## Fix

Give `checkIdle` a **content-aware idle clock** for the one runtime whose activity clock cannot see idleness: the Claude Code TUI over a non-ACP transport. This mirrors `pollerTUIComposerCanStick` / the ga-32mpo fix on the nudge path, which established that a Claude TUI's activity clock is necessary-but-not-sufficient and must be confirmed by a pane scan.

- New optional `runtime.IdleSnapshotProvider` — a single **non-blocking** pane observation (`SnapshotIdle`) reporting whether the pane currently shows a ready prompt with no busy indicator.
- `memoryIdleTracker` anchors a per-session `idle-since` timestamp from that snapshot, so `idle_timeout` measures **true continuous idleness**. A busy observation resets the anchor; firing clears it; a transient snapshot error leaves it untouched.
- Gated to Claude non-ACP (`transport != "acp" && provider == "claude"`); **every other session keeps the activity-clock path unchanged**.
- tmux `WaitForIdle` is refactored to share the same single-observation primitive (`snapshotPaneIdle`) — no behavior change (verified by existing tests).

**Blast radius:** Claude non-ACP idle-timeout path only.

## Tests

- `TestIdleTracker_ContentClockFiresDespiteFreshActivity` — the regression: fires after continuous content-idle > timeout even though the activity clock is always fresh.
- `TestIdleTracker_ContentClockResetsOnBusy` — a busy observation resets the anchor.
- `TestIdleTracker_ContentClockErrorLeavesAnchor` — a transient snapshot error neither fires nor resets.
- `TestIdleTracker_NonClaudeUsesActivityClock` — the gate: non-Claude and Claude/ACP keep the activity-clock behavior.
- Existing tmux `WaitForIdle` tests pass unchanged (refactor parity).

## Follow-up (not in this PR)

Defense-in-depth: add a witness-style `max_session_age` backstop to the always-on heartbeat agents in the gastown pack, so a wall-clock forced refresh bounds any future idle-detection gap regardless of provider.


vbtcl:
**CI triage — the 3 red checks are one pre-existing flaky test, unrelated to this PR.**

Final tally: **60 pass / 3 fail**, and the 3 failures are a single flaky test cascading through two aggregate gates:

- `Integration / packages-core-2-of-4` — `internal/productmetrics` › `TestDisableAndPurgeRejectsUnprovenPeerSuccessor/peer_successor_root_sync_failure`:
  ```
  control_unix_test.go:1770: purge error = productmetrics: state-changed-concurrently, want class "storage-failure"
  ```
  A Dolt concurrency race (a concurrent state-change beat the induced storage-failure) — flaky by nature.
- `CI / integration` and `CI / required` are just roll-ups of that one shard.

**Why it isn't this PR:**
- `internal/productmetrics` does not import any package this PR changes (`cmd/gc`, `internal/runtime/tmux`); the only touch to `internal/runtime` is a new **optional** interface (behavior-neutral).
- Every check that actually exercises this change is green: `Analyze (go)`, `Integration / packages-core-1-of-4` (internal/runtime), **all six** `Integration / packages-runtime-tmux-*`, and `Integration / packages-cmd-gc-integration`.
- Recent `main` CI shows the same intermittent red from flaky integration tests; the base commit (d5e37c7a4) was green.

I'm the fork owner, so I can't re-run the upstream Actions job to clear the flake — a maintainer re-run of `packages-core-2-of-4` should go green. Flagging for merge consideration.

quad341 (approved):
@vbtcl — this is excellent debugging. Pinning the always-on stall on `#{window_activity}` staying fresh under a repainting TUI is the kind of thing that hides for months, and the 1854-tick trace with the `max_session_age` sibling as a control made the diagnosis easy to follow. The idle-since anchor is the right shape, the gate is scoped tightly, and the four tests — error-leaves-anchor especially — cover the tricky transitions properly. Folding `WaitForIdle` onto a shared primitive was a nice touch.

Three things I'll take on my side. The main one: `SnapshotIdle` lands on `*tmux.Provider`, but `auto.Provider` and `hybrid.Provider` enumerate their optional interfaces instead of embedding, so in any city with an ACP agent the type assertion fails and the content clock quietly never runs. I'll add the delegation to both, plus compile-time assertions so it can't drift again. I'll also make the content clock additive rather than substitutive (a pane that never shows `❯ ` keeps its old recycle route), and clear the idle anchor when a session goes away so a restart can't inherit it.

The red shard is a `productmetrics` flake — I confirmed it can't reach your packages — and I'll re-run it. Merging with your commits and credit intact. Thank you for this.


---

PR #5470, opened 2026-08-21
fix(gastown): add max_session_age backstop to city-scoped heartbeat agents (ga-13yy0)

vbtcl:
## Summary

Defense-in-depth follow-up to the **ga-07mi8** idle-stall fix (PR #5468, which repaired the reconciler idle-detection path). The always-on Claude-TUI heartbeat agents **mayor, deacon, and boot** are city-scoped and carried no `max_session_age`, so a future idle-detection regression could leave them awake-but-idle for hours (deacon and mayor sat idle ~7h in the ga-07mi8 incident). The rig-scoped **witness** already carries this backstop in its pack `agent.toml`.

This patches the three city-scoped agents in the example root pack (`examples/gastown/pack.toml`) via `[[patches.agent]]`:

```toml
max_session_age = "6h"
max_session_age_jitter = "15m"
```

## Why these values

- **Wall-clock, not activity-clock.** `max_session_age` fires on session wall-clock age (`creation_complete_at`), independent of the buggy `#{window_activity}` clock that caused ga-07mi8.
- **Idle-gated.** The controller leaves a mid-turn agent alone until it settles — an idle Claude composer is not an approval prompt, so no false-defer.
- **6h** staggers these under the witness's 5h (avoids synchronized restarts) and stays below the ~8h Claude Code (Bedrock) cred-cache lifetime, so a preemptive cycle also refreshes creds.

## Placement note

Added to `pack.toml` (co-located with `[imports.gastown]`, which stamps these city-scoped agents) rather than `city.toml`, whose working tree carries unrelated local runtime churn (comment stripping + a locally-registered rig). `[[patches.agent]]` name-only matching applies to city-scoped agents (`dir` empty) — see `applyPackAgentPatches`.

## Test plan

Validated against the fully-resolved composition with `gc config show`:

- `mayor`, `deacon`, `boot` → `max_session_age=6h`, `max_session_age_jitter=15m`
- `witness` unchanged at `5h` / `15m`
- `dog` utility pool untouched (correctly excluded — not an always-on heartbeat)
- `gc config show --validate` → `Config valid.` (exit 0)


quad341 (approved):
@vbtcl — nice follow-through. After deacon and mayor sat idle ~7h in ga-07mi8, adding a wall-clock backstop to the three always-on agents that lacked one is exactly the right shape: `max_session_age` fires off session age rather than the activity clock that broke, so it doesn't depend on the thing it's guarding. The placement note about `pack.toml` next to `[imports.gastown]`, and the explicit "cycling the mayor is self-affecting but intended" line, both saved me real time.

I checked the composition rather than taking the description on faith: mayor, deacon and boot resolve to 6h/15m, dog stays untouched.

Three things on my side, none for you. The comment says 6h "staggers these under the witness's 5h" — it's above, so I'll flip that word. I'll add a small assertion in `examples/gastown` so those resolved values stay pinned. And the three red checks all died in the Ubuntu setup action before any test ran, so I'll rebase onto main and re-run rather than chase a phantom.

One note, no action needed: all three share a 6h base with ±15m jitter, so they can cycle in the same window. Idle-gating should scatter them. Merging with your commit intact.


---

PR #5474, opened 2026-08-21
fix(config): prefer executable steps over workflow anchors

jamelt:
## Summary

- Treat an assigned `graph.v2` workflow root as a launch/continuation anchor, not executable work.
- Prefer same-root ready or already-owned executable work for anchored sessions, retaining the root as a fallback.
- For fresh routed demand, stably place executable candidates before graph workflow roots while keeping roots available for root-only initial launch.

## Root cause

Generated work queries could repeatedly return a workflow root ahead of executable work in two cases:

1. An existing session owned a graph workflow root. The one-row in-progress recovery tier returned the anchor and exited before looking for the root's ready child.
2. A fresh pool read returned an older unassigned workflow root before an executable routed child. The claim hook consumes candidates in reader order, so successive replacement sessions claimed and drained the same non-executable root while ready children remained unclaimed.

This is why a broad claim-hook exclusion is not correct: graph roots remain legitimate initial launch and drain targets. The fix belongs in generated query ordering, with root fallback preserved.

## Safety and cost

- Ordinary in-progress and preassigned crash recovery retains priority.
- Anchored ready probing is constrained to the same `gc.root_bead_id` and route target; no fresh cross-workflow hop is introduced.
- Root-only initial launch remains available.
- Federated and single-store reader parity remains mechanically checked.
- Fresh demand uses one stable `jq` partition over the existing bounded result (at most 20 rows); it adds no storage read, polling loop, network call, service, or persistent process.
- Malformed reader output retains the existing fail-open behavior.

## Validation

- TDD red/green regressions for assigned roots, already-owned executable work, fresh root-before-child ordering, cross-root isolation, root-only fallback, and legacy control routing
- Full `go test ./internal/config`
- `go test ./internal/testpolicy/resourcecensus`
- `go vet ./internal/config`
- `make lint-affected` (0 issues)
- Commit-hook full `go vet ./...`
- `./scripts/test-local-parallel fast`: all 10 jobs passed, including six `cmd/gc` shards and the broad core shard
- Read-only production diagnosis reproduced fresh sessions repeatedly claiming/draining graph roots while six executable children remained ready

The 24 changed golden files are test snapshots for the generated Work/RoutedPool query variants. No API, dashboard, or other product-generated asset is changed.


quad341 (approved):
@jamelt — the diagnosis here is what made this easy to take. Separating the two failure paths (the one-row recovery tier exiting on the anchor, and the fresh pool read handing successive replacement sessions the same non-executable root) and then explaining why the tempting fix — excluding graph roots in the claim hook — would break initial launch is exactly right. Keeping the root as a fallback instead of filtering it out is what makes this safe.

`TestEffectiveWorkQueryPrefersExecutableDemandOverOpenGraphWorkflowRoot` asserting exact candidate order rather than membership is the test I'd have asked for, and it was already there.

One thing on our side: CI ran nothing but label bots on this head — no build, lint, or test job touched the code. I'm re-triggering the suite and merging once the `internal/config` jobs are green. Nothing for you to do.

Two follow-ups I noted, neither a regression: the anchor predicate is `gc.kind=workflow` AND `graph.v2` while `sourceworkflow.IsWorkflowRoot` is an OR and graph.v2-only roots exist; and a root blocked by its finalize step is never captured as an anchor. If the roots you diagnosed look like either, that's the next bead — happy to file it.


---

Issue #4837, opened 2026-07-29
hold-label-conventions retires bare `human`, but the doc's own caution, its out-of-scope list, and the ga-tug8ry.2 migration record all say bare `human` is not a hold state — the check enforces the table over all three


jacobhausler:
## The short version

`ga-tug8ry.1` consolidated eight ad hoc hold labels into `hold:mayor` / `hold:external`. That
consolidation is good and this issue does not ask to reopen it. The problem is one row in the
migration table in `engdocs/contributors/hold-label-conventions.md`:

| Legacy label | Replace with |
|---|---|
| `human-hold`, bare `human` | `hold:mayor` |

Every other retired label answers **"what is this bead waiting on."** Bare `human` answers
**"who executes this."** Those are different axes, and folding one into the other is not a
rename — it is a change of meaning that discards a distinction the taxonomy was not designed
to carry.

The doc already knows this, in three separate places. The checker knows it in none.

## 1. The row's own Notes cell contradicts the row

`engdocs/contributors/hold-label-conventions.md`, in the Notes cell of that very row:

> Caution: a bare `human` label can also appear alone for an unrelated reason (a human
> merge/PR action needed) that is not a hold state at all — check the bead's own context
> before assuming `human` implies a hold.

The row says migrate. The note attached to the row says this may not be a hold at all and
requires human judgement about the individual bead. Both cells are in the same table row.

## 2. The doc already has a section for exactly this, and `human` is not in it

Immediately below the retired-label table:

> **Explicitly out of scope — do not migrate these, they mean something different:**
>
> - `mpr-human-hold` and other `mpr-*` labels — [...] Not a generic bd hold label.
> - `build-blocker`, `ci-blocker`, `pre-push-blocker`, `push-blocking`, `test-blocker` — a
>   different semantic axis ("pipeline stage X is red because of me"), not "I am waiting on
>   decision-maker Y."

"A different semantic axis" is precisely the argument for bare `human`, and the doc makes it
itself for a neighbouring family. `mpr-human-hold` — the *hold* member of the `human` family —
is already carved out here. Bare `human` belongs in this list, not in the table above it.

## 3. The migration record says the migration already skipped them

The doc's own "See also" points at `ga-tug8ry.2` as:

> the migration record: before/after counts and the beads **intentionally skipped (bare
> `human` used for an unrelated reason)**.

So the one-time migration recognised the carve-out and honoured it. Only the *recurring check*
does not.

## What the checker does

[...]

PR #5477, opened 2026-08-21
fix(doctor): stop flagging bare human as a retired hold label

vishnujayvel:
## Summary

- The hold-label conventions doctor check treated a bare `human` label as a
  retired hold synonym of `human-hold`. Bare `human` answers who executes a
  bead, not what it is waiting on.
- Two pre-existing doc statements already said this: the migration-table
  Notes caution, and the ga-tug8ry.2 skipped-bead record. The checker
  enforced the table's "Replace with" cell over both. This PR also adds
  bare `human` to the out-of-scope list; that entry did not pre-exist.
- Drop `human` from `retiredHoldLabels` and the fix hint. Keep flagging
  `human-hold`. Align the one migration-table row and list bare `human` with
  the other out-of-scope labels. `CanFix()` / `Fix()` stay a no-op.

Addresses #4837.

Reported by @jacobhausler.

## Testing

- [x] `gofmt -l cmd/gc/doctor_hold_label_conventions.go cmd/gc/doctor_hold_label_conventions_test.go` — empty (formatted)
- [x] `go vet ./cmd/gc` — exit 0
- [x] `go test -count=1 -timeout 20m -v ./cmd/gc -run 'TestHoldLabelConventions'` — PASS (8 tests, 3.457s)
- [ ] `make lint-changed LINT_CHANGED_REF=origin/main` — failed in this environment: golangci-lint was built with Go 1.25, repo targets Go 1.26.6. Pre-existing toolchain mismatch, not introduced by this change.
- [ ] `make check` — not run in full; scoped package tests above are the relevant gate
- [ ] `make check-docs` — not run; the doc change is `engdocs/`, not Mintlify `docs/`
- [ ] `make test-integration` — N/A (doctor unit check only; no runtime/controller change)

## Checklist

- [x] Linked an issue
- [x] Updated docs for user-facing changes
- [x] Called out breaking changes or migration notes — none. Live `human` labels stay as-is. `human-hold` is still retired.

### DIFF
--- a/cmd/gc/doctor_hold_label_conventions.go
+++ b/cmd/gc/doctor_hold_label_conventions.go
@@ -21,16 +21,16 @@ var retiredHoldLabels = []string{
 	"blocked-on-upstream",
 	"blocked-prereq",
 	"human-hold",
-	"human",
 	"on-hold",
 }
 
 // holdLabelConventionsFixHint mirrors ga-tug8ry.1's disposition table.
-// engdocs/contributors/hold-label-conventions.md does not exist on
-// origin/main yet, so the hint must stand on its own rather than point at it.
+// Bare "human" is not a hold label (it answers who executes, not what the
+// bead is waiting on) and is intentionally absent. The hint stands on its
+// own rather than pointing at hold-label-conventions.md.
 const holdLabelConventionsFixHint = "Retired hold/blocked label in use (ga-tug8ry.1 taxonomy): " +
 	"arch-hold and blocked-prereq retire with no migration; blocked retires in favor of the " +
-	"native status field; blocked-by-operator, blocked-on-upstream, human-hold, and human " +
+	"native status field; blocked-by-operator, blocked-on-upstream, and human-hold " +
 	"migrate to hold:mayor; blocked-on-external migrates to hold:external; on-hold retires as " +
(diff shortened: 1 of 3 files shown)


quad341 (approved):
@vishnujayvel — the valuable part of this isn't the three-line code change, it's that you noticed the checker was enforcing the migration table's "Replace with" cell against two places in our own docs that already said the opposite: the table's own caution note, and the ga-tug8ry.2 record showing bare-`human` beads were *deliberately* skipped in the migration. That kind of internal contradiction usually just sits there generating false positives until everyone quietly learns to ignore the check.

Merging as-is. I traced the consumers on your head commit rather than taking it on faith — `retiredHoldLabels` is local to that one file, and `DispatchHoldLabels` never contained bare `human`, so there's no dispatch-side surprise hiding here.

Two things I especially liked: `TestHoldLabelConventionsCheckFlagsHumanHold`, so we can't over-correct and stop catching the label that *did* name a hold; and folding `human` into the existing out-of-scope test instead of giving it a lonely test of its own.

And thanks for flagging the `make lint-changed` failure as a toolchain mismatch instead of quietly dropping the checkbox — it is one (golangci-lint built against Go 1.25, repo on 1.26.6). Nothing for you to chase; CI's static checks passed here.


---

Issue #4504, opened 2026-07-22
bug: macOS supervisor reload leaks thousands of file descriptors


brettinternet:
### Before you continue

- [x] I searched existing issues and did not find a duplicate.
- [x] I read the relevant docs and contributor guidance.

### Gas City version

`1.3.5`

### Environment

macOS Darwin 25.5.0 arm64. Supervisor-managed city with one local rig that imports the city root:

```toml
[[rigs]]
name = "<rig-name>"

[rigs.imports.gc]
source = "."
```

### Reproduction

1. Start from a stopped supervisor:

   ```sh
   cd <city-root>
   gc supervisor start
   SUP_PID=$(pgrep -f 'gc supervisor run' | head -1)
   ```

2. Count unique numeric descriptors. `lsof` may repeat records on macOS, so deduplicate descriptor numbers:

   ```sh
   lsof -nP -p "$SUP_PID" -F f |
     sed -n 's/^f\([0-9][0-9]*\)$/\1/p' |
     sort -nu | wc -l
   ```

   Observed fresh supervisor: `17`.

3. Start or reconcile the registered city:

   ```sh
   gc start
   ```

   Observed after reconciliation: `5907` descriptors.

4. Trigger one config reload and wait for it to settle:

   ```sh
   gc supervisor reload
   sleep 5
   ```

   Observed: `8855` descriptors, an increase of `2948`.

5. Wait another 30 seconds and recount. Observed: `8850`. The descriptors remain open after reconciliation finishes.

6. Additional reloads eventually leave the process at exactly `10000` descriptors. Normal supervisor operations then fail with `EMFILE` / `too many open files`.

The settled descriptor breakdown at failure was:

```text
7599 REG
2391 DIR
2 KQUEUE
2 PIPE
2 IPv4
2 CHR
1 unix
1 systm
```

Most paths were under the city root, rig root, and Gas City pack cache. `node_modules` was not the dominant path in this reproduction.

### Expected behavior

A completed config reload closes the previous recursive watcher and returns the supervisor to approximately the same descriptor count. Repeated reloads should remain bounded.

### Actual behavior

Each reload permanently retains roughly 3000 additional regular-file and directory descriptors. Once the process reaches its descriptor limit, controller status probes time out, session startup fails, and pack/config reads fail.

A supported supervisor stop/start returns the process to 17 descriptors, confirming that the supervisor owns the leaked descriptors rather than this being global system pressure. Reconciliation and reload then reproduce the growth.

[...]

PR #5478, opened 2026-08-21
fix(controller): bump fsnotify to v1.10.1 to close Darwin FD leak

vishnujayvel:
## Summary

- On Darwin, recycling the supervisor config watcher leaked one kqueue-backed
  REG/DIR descriptor per watched path. `watchConfigTargets` already calls
  `Watcher.Close()` after registration goroutines finish
  (`cmd/gc/controller.go`); the leak is in fsnotify v1.9.0 itself
  (fsnotify/fsnotify#732).
- fsnotify v1.10.0 fixed this: "kqueue: drop watches directly in Close() to
  fix a file descriptor leak when recycling watchers" (fsnotify/fsnotify#740).
  This PR bumps `github.com/fsnotify/fsnotify` from v1.9.0 to v1.10.1. `go mod
  tidy` did not pull unrelated module upgrades.
- Adds a Darwin-only regression next to the existing `watchConfigTargets`
  tests: process FD count must return to the pre-watcher baseline after
  Close, and after a Close+NewWatcher restart. Non-Darwin platforms skip.

Addresses #4504.

Reported by @brettinternet. Independent reproduction and descriptor-numbering
evidence (kqueue replaced, REG/DIR FDs retained) from @atbrace.

## Testing

- [x] `gofmt -l cmd/gc/controller_watcher_fd_test.go` — empty (formatted)
- [x] `go vet ./cmd/gc ./internal/api` — exit 0
- [x] Darwin in-process fsnotify probe, 41 watched dirs, `syscall.Fstat` FD
      count:
      - v1.9.0: `baseline_fds=7 after_add=51 after_close=49 held_while_open=44 leaked_after_close=42`
      - v1.10.1: `baseline_fds=7 after_add=51 after_close=9 held_while_open=44 leaked_after_close=2`
- [x] `go test -count=1 -timeout 10m ./cmd/gc -run TestWatchConfigTargets_CloseReturnsFDsToBaseline -v` — PASS (0.03s)
- [x] `go test -count=1 -timeout 20m ./cmd/gc -run 'TestWatchConfigDirs|TestWatchConfigTargets|TestCityRuntimeReloadRestartsConfigWatcher'` — PASS (8.344s)
- [x] `go test -count=1 -timeout 10m ./internal/api -run 'TestLogFileWatcher'` — PASS (4.048s)
- [ ] `make check` — not run in full; scoped watcher/logwatcher tests above
      are the relevant gate
- [ ] `make check-docs` — N/A (no docs/ navigation change)
- [ ] `make test-integration` — not run; no supervisor lifecycle redesign
- [ ] `make lint-changed` / `golangci-lint` — not run. Local golangci-lint
      2.12.2 is built with Go 1.26.2; the repo targets Go 1.26.6. Pre-existing
      toolchain mismatch, not introduced by this change.

## Checklist

- [x] Linked an issue
- [ ] Updated docs for user-facing changes — none; dependency bump + regression
- [x] Called out breaking changes or migration notes — none. fsnotify v1.10.1
      requires Go 1.23+; gascity already requires Go 1.26.6.

### DIFF
--- a/cmd/gc/controller_watcher_fd_test.go
+++ b/cmd/gc/controller_watcher_fd_test.go
@@ -1,0 +1,115 @@
+package main
+
+import (
+	"bytes"
+	"fmt"
+	"os"
+	"path/filepath"
+	"runtime"
+	"sync/atomic"
+	"syscall"
+	"testing"
+	"time"
+
+	"github.com/gastownhall/gascity/internal/config"
+)
+
+// TestWatchConfigTargets_CloseReturnsFDsToBaseline is the Darwin
+// regression for gastownhall/gascity#4504.
+//
+// gascity already calls Watcher.Close() in watchConfigTargets cleanup.
+// On Darwin, fsnotify v1.9.0 still leaked one kqueue-backed REG/DIR
+// descriptor per watched path (fsnotify#732). fsnotify#740, shipped in
+// v1.10.0, drops watches directly in Close(). This test asserts the
+// process FD count returns to the pre-watcher baseline after cleanup
+// and after a Close+NewWatcher restart — not a magic absolute threshold.
+func TestWatchConfigTargets_CloseReturnsFDsToBaseline(t *testing.T) {
+	if runtime.GOOS != "darwin" {
+		t.Skip("darwin-only: the kqueue Close() FD leak (fsnotify#732) is a Darwin/kqueue bug")
+	}
+
+	root := t.TempDir()
+	const nDirs = 40
+	targets := make([]config.WatchTarget, 0, nDirs)
+	for i := 0; i < nDirs; i++ {
+		p := filepath.Join(root, fmt.Sprintf("d%02d", i))
+		if err := os.Mkdir(p, 0o755); err != nil {
+			t.Fatalf("mkdir %s: %v", p, err)
+		}
+		targets = append(targets, config.WatchTarget{Path: p})
+	}
+
+	var dirty atomic.Bool
+	pokeCh := make(chan struct{}, 1)
+	var stderr bytes.Buffer
+
+	baseline := countProcessFDs(t)
+	cleanup := watchConfigTargets(targets, &dirty, pokeCh, &stderr)
+	afterAdd := countProcessFDs(t)
+	held := afterAdd - baseline
+	if held < nDirs/2 {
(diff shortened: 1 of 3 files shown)


vishnujayvel:
Pushed a follow-up that removes the fixed sleep from the new FD regression test so the resource census stays at baseline; that was the TestRepositoryLedgerMatchesCensusAndDocumentation failure. The remaining red check is Container Scan (trivy HIGH findings in the base image), which also fails on main and is unrelated to the fsnotify bump. main is still on fsnotify v1.9.0, so the Darwin kqueue leak fix here is still needed.

vishnujayvel:
Rebased onto current main (f32f34cd6); both commits apply cleanly with no content change (range-diff shows `=` for each), and main is still on fsnotify v1.9.0 so the Darwin kqueue leak fix is still needed.
The `Evidence / critical-path suites` red was a base-sha artifact: that job is `pull_request_target` and checks out `pull_request.base.sha` (599afe65), which predates `scripts/prwatchdog` (#5576, on main since 2026-08-25), so `go build ./scripts/prwatchdog/cmd/watchdog` had nothing to build. The new merge base carries that directory.
`Image vulnerabilities` also fails on main and is unrelated to this change.

quad341 (approved):
@vishnujayvel — bumping a dependency is the easy part; proving *which* upstream commit fixes the leak and then writing a test that asserts return-to-baseline instead of some arbitrary FD ceiling is the part almost everyone skips. I checked your `fdCountNoise = 8` reasoning against fsnotify's actual kqueue source: `Close()` really does `unix.Close()` every watch descriptor inline, and only `kq` and `closepipe[0]` land in `readEvents`'s defer — exactly the two you accounted for. That comment turned verification into a five-minute job. The "did the watcher hold a measurable number of FDs" guard is a nice touch too; it means a no-op can't quietly pass.

Merging this. One thing I'll do on our side first: your regression is darwin-gated and our Mac lane skipped on this run, so I'll flip on `needs-mac` to let the assertion actually execute once before it lands. Nothing for you to change — that's our CI's trigger conditions, not your PR. The remaining red is the base-image trivy scan, unrelated to a Go module bump.

Credit forwarded to @brettinternet for the report and @atbrace for the descriptor-numbering repro, as you noted.


---

PR #5479, opened 2026-08-21
fix(dispatch): trust pinned checks across pack upgrades

jamelt:
## Summary

- keep exec checks from already-materialized workflows runnable after their pack is upgraded
- admit historical cache paths only for the same locked repository and equivalent pack-relative formula/asset root
- freshly validate the historical checkout and require the exact check file to be tracked
- preserve rejection of unrelated caches, sibling pack subpaths, dirty checkouts, malformed keys, untracked files, and symlink escapes

## Production reproduction

A graph workflow materialized an absolute `gc.check_path` from a content-addressed pack cache. After the city advanced that pack pin, the control dispatcher trusted only the new `FormulaSearchPaths` and quarantined the still-valid in-flight check as escaping trusted roots. The same condition reproduced across a second pack upgrade while the workflow remained open.

## Live verification

- exercised the resolver against the real historical `da4ecf3` fab-pack cache and the later locked `67f2042` cache; it admitted only the equivalent historical assets/pack roots
- installed a local integration build containing this commit plus the independent workflow-anchor fix from #5474
- the supported supervisor drift restart preserved sessions and reached `startup.ready=true` on build `8f068a12e`
- post-restart snapshot: 20 active sessions, 16 in-progress Trader workflows, 12 Trader closes in 10 minutes, and no new trusted-root quarantine, 429, quota, panic, or fatal signal
- the original quarantined control bead had already been force-finalized as changes-required before installation, so closed workflow history was not reopened or rewritten for a synthetic replay

## Tests

- RED/GREEN regression for same-source historical trust
- fail-closed security matrix for unrelated source, different pack subpath, malformed cache key/lockfile, missing Git metadata, dirty active/history checkout, and untracked check files
- stale validation-cache regression proving historical cleanliness is rechecked every execution
- dispatch coordination test plus existing unrelated-cache and symlink-escape regressions
- `go test -count=10` focused config/dispatch tests
- `go test ./internal/config ./internal/dispatch`
- resource census ledger
- `make fmt-check`
- `make lint` (0 issues)
- `make vet`
- mandatory pre-push fast suite: all ten jobs passed, including all six `cmd/gc` shards and the repository-wide core package sweep

Draft pending maintainer review.

### DIFF
--- a/internal/config/historical_cache_trust.go
+++ b/internal/config/historical_cache_trust.go
@@ -1,0 +1,164 @@
+package config
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/BurntSushi/toml"
+	"github.com/gastownhall/gascity/internal/pathutil"
+)
+
+// TrustedHistoricalFormulaCacheRoots returns historical cache roots that are
+// equivalent to roots derived from the city's currently trusted formula
+// layers. It fails closed: malformed lockfiles, non-canonical cache paths,
+// dirty checkouts, unrelated repositories, and different pack subpaths all
+// produce no roots.
+//
+// A running workflow retains the absolute asset path resolved when its formula
+// was materialized. Pack upgrades move the current formula layer to another
+// content-addressed cache directory, but the old checkout remains immutable.
+// This resolver admits that old path only when the candidate cache key binds
+// the same locked source to its checked-out commit and the path maps to the
+// same relative formula/asset root as an active layer.
+func TrustedHistoricalFormulaCacheRoots(cityRoot, candidatePath string, formulaSearchPaths []string) []string {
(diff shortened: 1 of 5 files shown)


jamelt:
Withdrawing after a stricter root-cause audit. This patch is a useful local compatibility bridge, but it is not a durable upstream solution: it depends on an unreferenced historical cache clone remaining present even though gc import prune may remove such clones, and it addresses persisted gc.check_path without solving the same lifetime problem for other external formula assets such as non-inlined description_file paths. Our production trigger was repeated local pack-pin churn with already-materialized workflows. A worthy upstream change needs an explicit workflow asset-lifetime contract (snapshotting or durable cache reachability) and prune integration rather than a check-only trust exception.

PR #5480, opened 2026-08-21
fix(test-harness): stop leaking managed dolt servers on uncleanly-killed runs

julianknutsen:
## Why

Leaked test `dolt sql-server`s (~57 threads each) accumulate whenever a `cmd/gc` test run dies uncleanly. Observed 2026-08-21: 242 of them built up over 5 weeks on one host and exhausted its `gascity-test.slice` pids quota (16384) — from then on every fork in the test slice was rejected and all subsequent test runs failed.

Root cause: `go test -timeout` raises **SIGQUIT**, but the `cmd/gc` dolt leak guard and the test watchdog only handled SIGINT/SIGTERM (`test/dolttest`'s Guard already handles SIGQUIT for exactly this reason — `cmd/gc` never got the same fix). The timed-out binary dies before `reapManagedDoltTestProcesses()` runs; the servers reparent to PID 1 and nothing ever reaps them, because the startup sweep's matcher only recognises `gct<pid>-`/`gctshard<pid>-` roots, not Go `t.TempDir()` roots.

## What

- **SIGQUIT handling** in the `cmd/gc` leak guard (`path_helpers_test.go`) and the managed-dolt test watchdog (`dolt_start_managed.go`), mirroring `dolttest.Guard`: timed-out shards now reap their servers before dying.
- **Self-healing sweep**: the startup sweep also reaps servers whose config sits under a Go `t.TempDir()` root (`Test*/...`) with the config file gone from disk — the signature of a run killed before its reap ran (SIGKILL/panic). A concurrent live run keeps its config on disk, so it is never touched. Pinned legacy-unowned-root behavior is unchanged (existing test still passes).
- **Blast radius**: `test-slice.sh` enrolls test scopes with `--property=TasksMax=8192` (half the slice quota hosts typically provision), so one leaky run can no longer starve the slice for every later run. `scripts/test-slice-enroll-test` expectations updated.

## Verified

- `go vet ./cmd/gc` clean; `go test ./cmd/gc -run 'TestIsStaleCmdGCTestConfigPath|TestManagedDoltSQLServerSysProcAttr'` passes (new test `TestIsStaleCmdGCTestConfigPathReapsAbandonedGoTempDirRoot` covers gone-config = stale, on-disk config = live).
- `go test ./scripts` passes; shellcheck clean on `test-slice.sh`.

Companion infra change (host reaper patterns + per-scope TasksMax in `gc-test-run`): gascity/infra#1547.

### DIFF
--- a/cmd/gc/dolt_leak_helper_test.go
+++ b/cmd/gc/dolt_leak_helper_test.go
@@ -483,6 +483,33 @@ func TestIsStaleCmdGCTestConfigPathSkipsLegacyUnownedRoot(t *testing.T) {
 	}
 }
 
+// A dolt server whose config sits under a Go t.TempDir() root
+// (Test<Name><rand>/...) with the config file gone from disk is the
+// signature of a run killed before its reap ran (timeout, panic, SIGKILL):
+// TestMain's cleanup removed the temp root while the server lived on
+// (observed 2026-08-21: 242 such servers accumulated over five weeks and
+// exhausted a host's gascity-test.slice pids quota). The gone config marks
+// it stale; a config still on disk marks a live concurrent run.
+func TestIsStaleCmdGCTestConfigPathReapsAbandonedGoTempDirRoot(t *testing.T) {
+	tempParent := t.TempDir()
+
+	goneConfig := filepath.Join(tempParent, "TestAbandoned1234", "001", ".gc", "runtime", "packs", "dolt", "dolt-config.yaml")
+	if !isStaleCmdGCTestConfigPath(goneConfig, nil, tempParent) {
+		t.Fatalf("abandoned go tempdir config path %q (file gone) not classified as stale", goneConfig)
+	}
+
+	liveConfig := filepath.Join(tempParent, "TestLive1234", "001", ".gc", "runtime", "packs", "dolt", "dolt-config.yaml")
+	if err := os.MkdirAll(filepath.Dir(liveConfig), 0o755); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(liveConfig, []byte("listener:\n"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+	if isStaleCmdGCTestConfigPath(liveConfig, nil, tempParent) {
+		t.Fatalf("live go tempdir config path %q (file on disk) classified as stale", liveConfig)
+	}
+}
+
 // TestSnapshotDoltProcessPIDs_EnumeratorErrorIsFatal pins that a
 // discovery error is reported via Fatalf so test runs surface
 // enumeration failures directly rather than silently treating them
--- a/cmd/gc/dolt_start_managed.go
+++ b/cmd/gc/dolt_start_managed.go
@@ -1198,7 +1198,9 @@ func runManagedDoltTestWatchdog(args []string, stdout, stderr *os.File) int {
 	go func() { done <- cmd.Wait() }()
 
 	signals := make(chan os.Signal, 2)
-	signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)
+	// SIGQUIT included so a `go test -timeout` signal wave reaching the
+	// watchdog still tears the dolt server down instead of orphaning it.
+	signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM, syscall.SIGQUIT)
 	defer signal.Stop(signals)
 
 	ticker := time.NewTicker(100 * time.Millisecond)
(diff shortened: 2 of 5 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277212792

quad341 (dismissed):
@julianknutsen — thank you for chasing this all the way down to the pids quota. The part I appreciate most is that you didn't stop at "add SIGQUIT": the self-healing sweep is what actually digs a host out once servers have already reparented to PID 1, and the `TasksMax` cap means the *next* leak, whatever causes it, can't take the whole slice down with it. Three layers, each standing on its own.

I traced the load-bearing pieces: `dolttest.Guard` and `test/integration` both already handle SIGQUIT, so `cmd/gc` was genuinely the odd one out, and "config gone ⇒ reapable" matches the scope watchdog's existing anchor — which is exactly why your on-disk check is safe against a live concurrent run.

One thing I'll take on my side rather than bounce back to you: `isAbandonedGoTempDirConfigPath` stats a config path lifted from another process's argv, and this package already bounds that probe with `statWithTimeout` (5s) because such a path can sit on a hung mount. I'll swap the bare `os.Stat` for it, keep your commit and credit, and merge once CI is green.

One footnote, no action needed: `go test -timeout` signals the test binary's PID, not the process group — so the watchdog change earns its keep on group-wide signals, and the sweep is what recovers the pileup you saw.


---

quad341 (dismissed):
Re-reviewed `86db4412e`..`5653345ee` (41 new commits since the last review): no new concerns — the prior review stands.

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277246559

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR stops managed Dolt test
servers from leaking when a test run dies uncleanly — a timeout, panic, or a
`SIGQUIT`/`SIGKILL` wave — which matters because leaked servers were piling up
(the 2026-08-21 incident left 242 orphans) and exhausting host resources across
the test fleet.

PR branch.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **Config-gone Dolt orphans were never reaped.** Before this change, a server
  whose config had been removed from disk under a Go `t.TempDir()` `Test*` root
  hit the no-owner branch and was *protected* rather than cleaned up. The patch
  adds `isAbandonedGoTempDirConfigPath` so those abandoned servers are now
  reaped — and review re-verified it never reaps a *live* server (a concurrent
  run keeps its config on disk → protected; a hung NFS/FUSE mount degrades to
  protect).
- **A `SIGQUIT` wave could kill the guard before it reaped.** The prior
  `SIGINT`/`SIGTERM`-only handler let a quit wave tear down the guard/watchdog
  first. The patch broadens signal handling while preserving the
  `go test -timeout` goroutine dump — the guard reaps, then re-raises so the
  runtime still prints the stack trace.

Review findings addressed:
- **Reaper doc-comment precision** — the rationale comment implied `TestMain`
  cleanup runs on `SIGKILL`. Since `SIGKILL` can't be trapped, a gone-config
  signature after `SIGKILL` actually comes from an external temp wipe, not
  in-process cleanup. Reworded to key on the observed signature in maintainer
  commit `4c775b6`.
- **Enroll self-test hermeticity** — the self-test asserted a literal
  `TasksMax=8192` without unsetting `GC_TEST_SLICE_TASKS_MAX`, so a host that
  exports the new override would fail it spuriously. The self-test now unsets
  that variable, in maintainer commit `4c775b6`.

- **ga-stkys** — *Coverage (minor):* `SIGKILL` orphans that *retain* their
  config (temp root survives, no external wipe) are still not reaped. This is a
  by-design coverage limit, not a regression — the dominant incident signature
  was config-gone, which this PR now handles. Not addressed here.
- **ga-8pv17** — *Resource (minor):* the `TasksMax=8192` per-run cap could
  fork-reject an unusually heavy legitimate parallel run. Already mitigated by
  the `GC_TEST_SLICE_TASKS_MAX` override and a generous default. Not addressed
  here.

### Maintainer Changes
Maintainer fixes were pushed directly to the PR branch as one fixup commit,
`fix(review): key reaper doc comment on gone-config signature; harden enroll
self-test env` (2 files changed, 7 insertions(+), 4 deletions(-)). The branch
was also rebased onto the latest `main`; your original commits are preserved
with their authorship.

### Final Review Status
Ready for the merge queue: final head `4c775b6` has passing required GitHub
`status/merge-ready` so the merge queue can pick it up. (The advisory Mac
`quality` check is red only because the macOS runner image ships an older Go
toolchain than `go.mod`'s floor — a fleet-wide runner-image lag unrelated to
this change; it is not a required check.)

CI: https://github.com/gastownhall/gascity/actions/runs/33088963343

### Review Iterations
One review pass, approved on the first pass; the two nearby minor cleanups were
folded into the maintainer fixup rather than forcing another iteration.

---

PR #5488, opened 2026-08-21
fix(cmd/gc): close the infra-class store-routing gaps found by the reconciled audit

julianknutsen:
Seven fixes for sites that reached infra-class data through the wrong store, from the reconciled wrong-store audit (now landed at `engdocs/contributors/infra-class-store-routing-audit.md`). Each commit is one finding, each has a mutation-verified test — the pre-fix wiring makes it fail — and the tests live in the existing test files for the code they cover.

| Commit | Finding |
|---|---|
| `6e0997d9` | handoff message bead written to the work store instead of the messaging class |
| `ad3ced73` | autoclose and wisp-step graph legs read an arbitrary work store |
| `91f224a3` | PR-repair cook used one store for two classes |
| `e80e5978` | bead events for relocated class prefixes (`gcg-*`) fell through to the broadcast |
| `8b5c2215` | `gc graph` answered every id from one store |
| `c2cc5197` | wait dependency reads missed the binding — an `ErrNotFound` there is read as proof the dependency is gone, a silent FailWait |
| `208c35b9` | a drain dispatched outside its root's work scope resolves an EMPTY convoy and reports success (ga-w2mf3) |

Only `storageSplitNone` and `storageSplitWhole` are servable, so every one of these was latent on the machines we run.

The last commit also keeps `TestResidencyBoundaryGrepRatchet` green without rebaselining growth: the three `cityPath -> graph binding` derivations collapse into one accessor, and the two remaining census hits are annotated with why they are not residency answers. The baseline diff is a two-row rename, net zero sites.

`make test-fast-parallel`: all 10 jobs green. Pre-push suite green.

Beads: ga-sdxnn, ga-ntixo, ga-uekbo, ga-v2ouw, ga-w2mf3 closed. ga-cu12x filed for a hazard surfaced while writing these — `storeref.Resolve` tries `PrefixOwner` before leg order, so a migration-preserved infra id still resolves to the retained work-store copy; it needs to be shown to BITE on a live city before it is fixed.


jacobhausler:
## U17: Migration-retained work copies remain claimable (Westlands, 2026-08-22)

Measured observation from a live city deployment:

**Evidence:** After `gc storage migrate` converged, the source was retained verbatim. `gc hook --claim` and `gc ready --status in_progress --assignee=self` still read retained work-store copies of migrated graph-class beads. The close plane was infra while the claim plane was work; one-shot sessions re-claimed closed steps (2,227 rollouts; luna-1 woke 244 times).

**Minimal repro:** Migrate a disposable city with source retention enabled, close a migrated graph-class step in the authoritative infra store, then run the claim/readiness query and show the retained work copy is returned.

**Fix/test:** Tombstone/close retained copies during migration, or filter `gc.infra_migrated_from=work` from claim/readiness after convergence; add a cross-store regression proving closed authoritative work is never re-claimed.

Filed as evidence on this PR rather than a duplicate issue. Related: #5087.

jacobhausler:
## U18: append-notes hard-refuses on infra-class beads, stranding formula drain steps (Westlands, 2026-08-22)

Measured observation from a live city deployment:

**Evidence:** `gc bd update <id> --append-notes ...` hard-refuses on infra-class beads with `no representation in the class store contract`. The formula drain step `mol-verified-work:351` therefore cannot close, and the step remains open indefinitely.

**Minimal repro:** Create or migrate an infra-class bead in a disposable city, run `gc bd update <id> --append-notes ...`, and capture the refusal while the drain step remains open.

**Fix/test:** Map notes to the authoritative representation, explicitly drop them with a recorded result, or reject the formula path before execution; add a drain regression that cannot leave a step permanently open on this class.

Filed as evidence on this PR rather than a duplicate issue.

jacobhausler:
## U19: dead-assignee reopen consults non-authoritative retained copies (Westlands, 2026-08-22)

Measured observation from a live city deployment:

**Evidence:** `bead.dead_assignee_reopened` reopens retained work-store zombies and spreads them across seats after a storage migration. The reopen path acts on copies that the authoritative store does not contain.

**Minimal repro:** Retain source copies during a disposable migration, mark a source-only copy as dead-assignee work, run the reopen path, and show that it is reopened despite the authoritative store not containing it.

**Fix/test:** Make dead-assignee reopen consult the authoritative store and ignore/tombstone non-authoritative retained copies; add a negative regression for source-only zombies and a positive regression for authoritative abandoned work.

Filed as evidence on this PR; related issue lineage: #4387. Split out if the scope warrants a separate item.

jacobhausler:
## legacy migrated IDs can re-expose retained work as actionable

I found a related but broader case while checking the class-store routing audit in
this PR. The evidence below is observational; this comment proposes no code change yet.

### Live black-box evidence

For a representative migrated legacy ID (`<legacy-id>`), the class-aware command
(`gc bd show`) saw the authoritative class row as `closed`, with
`gc.outcome=pass` and `gc.infra_migrated_from=work`. The underlying work-store
command (`bd show`)
simultaneously returned the retained work row as `in_progress`, assigned to a
different manager. This is a retained-work projection becoming actionable again;
it does not by itself prove that the authoritative class row was durably reopened.

The ordered observation was:

1. A close event recorded the class result as `closed/pass`.
2. Cache reconciliation subsequently reported the retained work row as
   `in_progress` and, in one observation, `open`.
3. A later manager update again reported the class result as `closed/pass`.

### Current-main mechanism

This remains present on current `origin/main` at
`ca54c7264ff139b3502064daa430bdb719992d81`:

- `cmd/gc/api_state.go:820-876` resolves `cr-*` IDs through configured HQ/rig
  work-prefix logic; broadcast is only the fallback when no configured prefix
  matches. The class-binding residence is not consulted for these legacy IDs.
- `cmd/gc/api_state.go:744-774` applies the event to the resolved caching store
  and starts close autoclose from that resolution.
- `internal/beads/caching_store_reconcile.go:290-355,487-565` can then rebuild
  the retained work cache from its backing row and emit a cache update.
- `cmd/gc/cmd_bd_by_id.go:609-677` does perform a class-binding residence probe
  for by-ID access, which explains why `gc bd show` and bare `bd show` can
  disagree.
- `cmd/gc/ready_federation.go:20-30,263-341` defines work-first federated
  readiness, while `cmd/gc/claim_class_route.go:19-60` deliberately preserves
  that work-first choice at claim time.
- `cmd/gc/build_desired_state.go:4517-4645` is the downstream controller
  demand/selection surface to verify; this observation does not claim that it
  directly reads the stale work row.

### Why this is broader than the current diff

The current PR correctly adds reserved class-prefix candidates (for example,
`gcg-*`). Legacy IDs retained from migration keep their original `cr-*` prefix,
so they continue to resolve through the HQ/rig work-prefix logic instead of by
class-binding residence. The earlier U17 comment on this PR already records this
retained-source resurrection case, but event routing alone cannot close it:
reconciliation, readiness, demand, and claim must agree on which row is
authoritative.

### Proposed authority rule

When a class-binding row exists and is stamped
`gc.infra_migrated_from=work`, it overrides the retained work row across event
routing, reconciliation, ready, demand, and claim. The retained work row remains
available for rollback/source retention but is not an actionable runtime copy.

For unmarked duplicate IDs, preserve the existing work-first behavior and its
co-resident-ID contract. This keeps the change scoped to migration-created
duplicates rather than redefining all federated store precedence.

### Proposed regression tests (expected to fail before fix)

1. `TestBeadEventRouting_LegacyMigratedClassIDUsesClassStore`
2. `TestCachingStoreReconcile_LegacyMigratedClassIDDoesNotResurrectRetainedWork`
3. `TestMigratedClassAuthority_AgreesAcrossReadyDemandAndClaim`

Would you like this stamped-class-authoritative rule implemented as an extension
of #5488, or as the focused ga-cu12x follow-up after this PR lands?

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277228013

quad341 (approved):
@julianknutsen — seven findings, seven commits, one mutation-verified test each, and the tests live in the existing file for the code they cover rather than in a new `_routing_test.go` nobody finds later. That's the shape that makes 1146 lines actually reviewable.

Two things I noticed tracing rather than skimming. `graphRowShows` asserting on the BLOCKED BY column instead of on id presence is the right oracle — unresolved members still render, so the obvious assertion would have passed on the bug. Same instinct in the relocated-prefix test, asserting the molecule root's status rather than the resolved store. And turning the autoclose graph store into a required parameter is the actual fix; the collapse-to-`store` default is what let both hook callers silently opt out.

Merging as-is. Two follow-ups on our side, not asks:

1. `assertDrainRootScopeMatchesDispatch` returns an unclassified error, and `runWorkflowServeFollow` exits the dispatcher on that class — the bead stays open, so it'd recur every restart. The guard is right about the hazard; I'd like it quiet or skip rather than fatal.
2. `cliGraphStore`'s `cfg` is discarded by `resolveClassStore`, so the three new `loadCityConfigWithoutBuiltinPackRefresh` calls are uncached disk reads on the `on_close` and injection paths.

Filing ga-cu12x instead of fixing the `PrefixOwner` ordering blind was the right call — show it bites first.


---

quad341:
@julianknutsen — seven findings, seven commits, one mutation-verified test each, living in the existing test file for the code they cover. That structure is what makes 1146 lines reviewable, and it's why this is landing rather than sitting.

Two things I checked by tracing rather than skimming. `cookOnClassRouted` is a faithful inline of `CookOn`+`Cook` with the store picked after compile — the ordering problem was real and the fix doesn't quietly drop the ParentID guard or runtime-var validation. And `graphRowShows` asserting on the BLOCKED BY column is the right oracle: unresolved members still render, so the obvious id-presence assertion would have passed on the bug.

One integration seam I'll take on my side. The new drain guard is correct, but its error returns before `handleControlDispatchError`, so it classifies as `TierNone` and `runWorkflowServeFollow` exits the dispatcher — and `controlReadyFallbackReady` unions a rig queue with the shared city binding, so a `city:`-rooted drain reaching a rig dispatcher is the expected split-city shape. I'll route the refusal through quarantine, add the lifecycle assertion, and re-run the `cmd/gc` shards plus `make test-fast-parallel`.

Filing ga-cu12x on the `PrefixOwner` ordering rather than guessing was the right instinct. Thank you.


---

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277246486

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR closes a whole class of infra-class store-misrouting bugs found by the reconciled audit: on a *migrated* city the graph / session / messaging / orders / nudge classes relocate onto a separate store binding, but several call sites still read or write those classes through the raw *work* store — yielding a silent empty read ("nothing to do") or a stranded write. Tightening this routing matters because those failures are invisible: work quietly strands instead of erroring, which is exactly the kind of bug that erodes trust in the orchestrator on exactly the multi-store deployments the class split exists to support.

This PR was reviewed and **accepted as-is**.

### Original PR Review
Decision: approve.

Specific gaps fixed (each validated by review, and each provably identity on an unmigrated single-store city):
- **Drain dispatched outside its root's work scope** read an empty convoy and reported a false "all members drained" success. The new `assertDrainRootScopeMatchesDispatch` guard runs in the `drain` arm before `dispatch.ProcessControl`, turning that silent false-success into a loud error that names both scopes — failing open on ambiguity and closed only on a positive two-scope mismatch. (The highest-leverage change; covered by `TestAssertDrainRootScopeMatchesDispatch` and cross-scope drain tests.)
- **Autoclose / wisp-step graph-parent walk** read from the closed bead's owning (work) store, so it came back empty on a migrated city. The graph-store parameter is now a *required* `beads.GraphStore` (compile-enforced) and the parent walk uses `cs.GraphBeadStore()`.
- **The PR-repair cook** instantiated a graph-class molecule into the work store. `cookOnClassRouted` now selects the store by `recipeCoordClass` while remaining a faithful clone of the `CookOn`→`Cook` order and guards.
- **`gc graph` per-id**, **durable waits on a graph-class dependency**, and **bead-event routing for relocated class prefixes** all resolved a single work store and so returned empty for graph-relocated ids; each now probes the class binding first, with longest-prefix namespace-boundary matching to avoid cross-shadowing city/rig prefixes.
- **The handoff mail bead** is now minted on the messaging store threaded from `cliMailStore(...)` so it stays visible to a relocated messaging class.

### Maintainer Changes

Non-blocking follow-ups (shared for visibility — NOT completed in this PR, filed as beads so they are not lost):
- **ga-4vgoi** — maintainability (minor): `assertDrainRootScopeMatchesDispatch` uses raw `==` where other root-store-ref comparisons use `rootStoreRefMatchesCandidate`. The strictness is deliberate (it also catches a city-rooted drain dispatched at a rig), but it is byte-sensitive to the `city:<name>` spelling and that invariant isn't pinned — a clarifying comment (and optionally a mint-vs-dispatch spelling test) would keep a future refactor from reopening the city↔rig gap.
- **ga-s1ms9** — correctness/caching (minor, pre-existing): the *id-less* bead-event broadcast fallback still omits relocated `storageRoutes` stores, so a subject-less bead-mutation event would not invalidate a relocated class store's cache. Narrow in practice (lifecycle events normally carry a subject), and not introduced by this PR.
- Testing (nit): the front-door DI guard asserts the four sites textually call `cliGraphStore(` rather than asserting store identity; no action needed — the disjoint-store behavioral tests added alongside (`wisp_autoclose_test.go`, `molecule_autoclose_test.go`, `wisp_step_inject_test.go`, `api_state_test.go`) are the semantic backstop.

### Final Review Status
Ready for the merge queue: final head `c306763` has passing required GitHub checks (CI, CodeQL/Analyze, integration shards, Mac regression, and the Go vulnerability scan all green), and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/32801469335

### Review Iterations
1 review pass performed; the PR reached approval with 0 fix iterations (no maintainer fixups were needed).

---

PR #5489, opened 2026-08-21
Keyed session reconciler: exact per-session ownership, opt-in via daemon.session_reconciler

julianknutsen:
## Why

The legacy session reconciler is a single fleet-wide pass: one ~2,600-line function that re-derives every session's state on every controller tick. That shape is the root of a family of production problems we have been paying for individually:

- **Fleet-coupled failure.** Every session's reconciliation shares one pass, so one wedged row (a hung provider probe, an unreadable process, a stuck drain) degrades or stalls attention for every other session on the city.
- **Unfenced effects.** The pass acts on state it derived earlier in the same sweep. Under concurrent writers (agents, operators, the API) that is a lost-update machine — we have shipped multiple point fixes for suspend/wake/drain races that are all instances of the same missing fence.
- **Polling cadence.** Everything waits for the next patrol tick. A session with work queued sits idle until the sweep next visits it; recovery from any transient is quantized to tick latency.
- **Unbounded, silent retry.** Failure handling was retry-at-cadence with no obligation-scoped bound — storms look like log noise until they leak enough pool seats to starve scheduling (we hit exactly this in dogfood).

## What this ships

A keyed reconciler that splits the problem in two:

- a **read-only detector sweep** that only *observes*: it derives per-session conditions and enqueues exact session keys — it performs no effects and takes no locks beyond reads;
- **per-key controllers** that own one session each: every effect re-validates against current authoritative state under that session's lock, and every state-changing commit is **revision-fenced** (compare-and-set on the store's signed revision), so a concurrent writer makes the commit refuse rather than lose the update.

Around that core: **event-triggered admission** (a session bead write reaches the reconciler when the work lands, not on the next tick), **bounded escalation** (every retry loop carries an obligation-scoped counter that escalates to a named, loud terminal state instead of spinning at cadence), and **fail-closed handling of degraded provider input scoped to the session** (an unreadable process that provably cannot belong to a session no longer blocks that session's proof).

## Correctness case

- **Certified equivalence before any default changes.** A 7-day side-by-side campaign ran both engines against the same production-shaped traffic under coexistence: **79,382 joined decision comparisons, zero shadow-effect violations, zero unclassified divergences**, every acting detector family ≥99.5% agreement, across 2 controller restarts and a config reload, with a gap-free 168h arming ledger. Every divergence class in the final account is individually adjudicated and carries a verification obligation (e.g. a timing-skew record must prove its adjacent-cycle twin to leave the mismatch column).
- **Dogfooded on six production cities** for the past days under `auto`. That soak surfaced real bugs — and every one was root-caused, fixed, and converted into a standing guard rather than a patch (below).
- **Both engines coexist in every build.** Default `off` is byte-for-byte today's behavior; `auto` takes keyed ownership per session with row-level fallback to legacy for rows the keyed engine cannot certify; rollback is config-only (`session_reconciler = "off"` + controller restart).

## Performance case (stated only as far as the evidence goes)

- **Reaction latency**: under `auto`, a session that needs attention is acted on when the triggering work lands rather than on the next patrol tick — patrol-cadence waits (tens of seconds) become event-driven admission.
- **Work scaling**: attention is per-key — the controller does work proportional to sessions *needing* it, instead of re-deriving the full fleet every tick; one stuck session no longer taxes the others' reconciliation.
- We make **no microbenchmark claims**: the hermetic A/B shows mixed per-op signs (keyed start slower at p50, stop faster, nudge equal) at magnitudes ~4 orders below the operative budgets. The wins above are structural, not per-op.

## Hardening that ships with it

The integration and dogfood period produced standing guards now in the tree: a **twin-drift census** (structural test that keyed twins of shared deciders handle every rung their decider emits — the bug class where an upstream rung addition silently no-ops a twin), **merge-integrity checks** (`make check-merge-integrity`: deleted-symbol resurrection, vanished-test census, restored-lane-deletion — each validated against a real historical merge it would have caught), **retry-bound escalation** in the drain family, and seam-composition tests that exercise production wiring paths end-to-end (store capability resolution, event emission on split stores).

## Provenance

The full design docs, the campaign sign-off package with its 16 evidence artifacts, and the complete incident/verdict ledger are deliberately **not** in this diff — they live on the working branch `rec/ga88-continue` and bead `ga-f7v2ft` (ledger: `.161`, `.122`). Code comments citing `DETECTOR.md §N` refer to that archive. This branch is the shipping subset: engine, guards, user docs (`docs/runbooks/keyed-session-reconciler.md`), CHANGELOG.

## Merge state

origin/main is merged into this branch and `scripts/check-merge-integrity.sh --merged` passes at the merge commit (0 resurrections / 0 unaccounted vanished tests / 0 restored lane deletions). Full gates green on a normal (unbypassed) push: vet, all six cmd/gc shards, parity suite, all censuses, core-boundary, docsync. One pre-existing failure reproduced identically at origin/main itself (`test-productmetrics-testhook` deadline on this loaded host) — not introduced here.

Remaining before merge: the owner's soak-done call on the fleet dogfood.


julianknutsen:
Added `4b0fb26c35` (cherry-pick of `425e319b0b` from `rec/ga88-continue`): **a corpse is not a runtime** — kernel-dead processes now leave the drain-ack liveness completeness domain before the environ read. ga-lp5w6 (already in this cut) narrowed the sweep's completeness domain to the session's reachable scope, but zombies still escaped every proof: an unreaped process keeps a readable `status` (still our uid, so it is "ours" to account for) while `/proc/<pid>/environ` answers EACCES even to its owner, and an **orphaned** zombie re-parents to init — the one lineage shape both lp5w6 scope proofs decline, because an init-exiting chain is the orphan shape and nothing proves whose orphan it was. On the window-31 supervisor host that was 522 host-wide zombies (many the supervisor's own orphaned tmux servers, pids alive since Jul 17/Aug 5) holding every acked drain incomplete forever: refusal counters unbounded (or-3kw43 N=2774, tr-l1tq2 N=3062), 14,011 `session name already exists (skipping)` lines, no pool seats, no wakes, and restarts bought only 30–90 minutes each. State `Z` (and `X`/`x`) is kernel-proven terminal — the address space is gone and no code will ever run under that pid again — so the sweep now reads the world-readable state field from `/proc/<pid>/stat` (parsed from the LAST `)`, since comm is arbitrary process-controlled text) and drops those processes from the completeness requirement regardless of scope or readability, counting them in the existing one-line residue summary rather than a line each. Only their own candidacy as a living runtime is excluded: a zombie's ppid is still valid, so it remains a usable link in other processes' parent chains. The same audit closed a pre-existing fail-**open** in both scope proofs, which concluded from a spawner's environ that it lacked the target session ID — a corpse's environ reads empty whatever it once held, so both proofs now decline a kernel-dead ancestor whose environ is the evidence. The sudo-child protection is untouched (`TestScanWithRootSinceUnreadableSudoChildProofFailsClosed` passes with its file unmodified, joined by the revalidation/pane/lineage fail-closed controls and an in-test control that flips one zombie to state `S` and must fail closed again); RED was demonstrated before each arm, and self-recovery is pinned end to end against the real scanner — while the undecidable runtime lives the .173 obligation escalates and parks, and the moment it dies unreaped the same retained obligation resolves, lifts its fence and clears its refusal history. This is a real correctness/availability fix and belongs in the cut: without it, the keyed reconciler's drain-ack path is one zombie away from parking a pool seat permanently on any host that accumulates them. Gates on this branch: `go vet ./...` clean; `internal/runtime`, `internal/session`, `internal/pidutil`, `internal/testpolicy` suites green; `cmd/gc` `Session|Reconcile|Drain|Parity|Census|Scan` sweep green (142s); `TestParityJoin*` 55/55; `check-core-boundary` OK; `check-merge-integrity.sh` pre-merge run informational only (its two findings — `ensureCanonicalScopeMetadataIfPresent`, `scopeSkipBatchUpdater` — are pre-existing lane/upstream drift, untouched by this commit, which adds tests and deletes no symbol).

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277220507

PR #5492, opened 2026-08-21
fix(rig): judge a relative rig path against the city, not the caller's cwd

julianknutsen:
`pathdurability.Classify` judged a relative rig path against the process's working
directory instead of the city. `city.toml` may hold a rig path relative to the city root,
and the rest of the codebase joins it onto the city before use
(`internal/rig/topology.go:46-47`) — the classifier did not.

## Where this was reachable

The **boot audit** in `internal/config/site_binding_durability.go` reads `rig.Path`
verbatim out of the site binding. A controller started from a working directory on tmpfs
warned that a rig safely under the city was about to be lost:

```
rig "backend" is bound to rigs/backend, which is on tmpfs and does not survive a
restart of this machine or container: the rig content is lost on the next restart.
Move it under the city directory (...) and re-register it with `gc rig add`
```

`rigs/backend` was already under the city directory, on the durable device — exactly where
the message told the operator to move it.

**Correction to an earlier draft of this PR:** it claimed `gc rig add rigs/backend` was
refused outright. That is not reachable, and three reviewers independently caught it.
`validateRequest` (`internal/rig/deps.go`) rejects a non-absolute `ProvisionRequest.Path`
before the durability check runs, and the CLI resolves a bare relative argument against the
city first (`resolveRigAddPath`, `cmd/gc/cmd_rig.go`). Fixing `Classify` hardens the
registration path too, but as **defense in depth behind two existing guards** — not as a
repair of a live refusal. The advisory audit was the only production victim.

## Why the fail-open rule did not catch it

`Classify` has an escape hatch: an unprobeable path returns `Unknown` and the caller lets
it through. `probeDevice` walks up the ancestors of the path it is given, and for a
relative path that walk terminates at `"."` rather than running out of ancestors. So it
reaches `Unknown` only if `stat(".")` itself fails — in practice it always produces a
**confident verdict about the wrong directory**. The fail-open rule protects against "I
cannot tell", not against "I answered a different question".

## The fix

```go
if !filepath.IsAbs(path) {
    path = filepath.Join(cityRoot, path)
}
```

Kept in `Classify` rather than pushed to the two callers. A relative rig path meaning
*city-relative* is a settled convention here — `topology.go`, `detectRigReAdd`,
`RigComplete` and `resolveStoreScopeRoot` all encode it — and `Classify` already takes
`cityRoot` and already promises "judged relative to `cityRoot`" in its doc comment. The
alternative (a strict absolute-only contract) cannot be enforced at this layer: `Classify`
has no error channel by contract and panics are banned, so the realistic strict variant is
`relative → Unknown`, which fails open and would leave the boot audit permanently blind to
genuinely doomed bindings. Pushing the join to callers also duplicates it in two places and
re-creates this exact omission the next time a caller appears.

## Tests

Three, all written red first and mutation-proven.

**`internal/config`** — the audit that was actually broken, against the **real** probe.
Every other test in that file stubs the classifier with `func(_, path string)`, which
ignores `cityRoot` entirely, so none of them could see this. It carries an absolute-tmpfs
control binding that must warn **in the same run**, so silence for the relative binding
cannot come from an audit that stopped warning at all. Reverting the fix reds it with the
operator-facing message quoted above.

**`internal/pathdurability`** — the durable case, plus the control that keeps it honest:
`Classify("/city", "../tmp/adopt")`, a relative path that escapes onto tmpfs, must still
come back `Ephemeral`. Without it, a fix that simply accepted every relative path would
pass.

**`internal/rig`** — the first test in that file to exercise the real probe instead of
handing `rigPathDurabilityFinding` a synthetic `Result`. It calls `checkRigPathDurability`
directly, below `validateRequest`, so the guard stays honest if a future caller loses the
resolution.

All three skip loudly rather than passing vacuously when the host has no tmpfs to
distinguish the two behaviours.

## Mutations confirmed red

- revert the `IsAbs`/`Join` → the config audit test and both new probe tests fail;
- accept every relative path unconditionally → the escape control fails;
- join against `/` instead of `cityRoot` → the fake-mount subtest fails;
- drop the trailing path segment → the `Probed` assertion fails.

Fixes ga-o1qbw.

### DIFF
--- a/internal/config/site_binding_durability_test.go
+++ b/internal/config/site_binding_durability_test.go
@@ -1,6 +1,8 @@
 package config
 
 import (
+	"os"
+	"path/filepath"
 	"strings"
 	"testing"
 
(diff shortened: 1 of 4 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277227944

quad341 (approved):

The diagnosis is the best part of it. `Classify` wasn't failing to answer; it was answering *confidently about the wrong directory*, and your note on why the fail-open rule couldn't save it — `probeDevice`'s ancestor walk bottoming out at `"."` instead of running out of ancestors — is what turns a four-line patch into something reviewable. "The fail-open rule protects against 'I cannot tell', not against 'I answered a different question'" is going in my head permanently.

I traced both production call sites in a checkout rather than taking the writeup's word for it, and they agree with you: `topology.go` does the identical `IsAbs`/`Join`, so `Classify` was the outlier against its own doc comment. I also confirmed all three new tests actually *run* here rather than skipping.

Two things I want to name. First, the correction in the body — retracting the "rig add was refused outright" framing and re-describing that test as defense in depth behind `validateRequest` and `resolveRigAddPath` is exactly the right call, and both guards check out. Second, wiring the `internal/config` test to the real probe: every other test in that file stubs with `func(_, path string)` and throws `cityRoot` away, so that file *structurally could not* have caught this. That's the part that outlives the bug.

The Blacksmith `TestControlBytesAreStripped` failure is unrelated flake — different package, no shared path. Ignore it. Thanks for this one.


---

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR makes rig-path durability classification resolve a *relative* rig path against the city root before probing its filesystem device — matching the canonical resolver in `internal/rig/topology.go` — which stops operators from hitting a false `rig add` refusal and a self-contradictory boot warning for a perfectly valid relative rig path.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- Relative rig paths were classified against the process working directory instead of the city, so `pathdurability.Classify` could mis-judge a valid path's storage device. The patch resolves a non-absolute path against `cityRoot` before probing (`internal/pathdurability/durability.go`), so the two production callers — the config boot audit and the rig-registration guard — now inherit the correct boundary.
- Backward compatibility is preserved: the new behavior is guarded by `!filepath.IsAbs(path)`, so existing absolute-path callers are byte-for-byte unchanged and the change stays additive (+164/-0, 11 production lines plus tests).
- Correctness was proven by reproduction: with the new guard removed, the three added tests fail with the exact operator-facing symptom, and they pass once it is restored; the `../tmp/...` escape case shows this is a scoped city-relative resolution, not a blanket "trust any relative path" accept.

For maintainer awareness, review also noted a few low, non-gating forward risks — all pre-existing or already mitigated, none requiring a change here: a relative `cityRoot` would re-expose CWD-relative probing (unreachable today, since city discovery yields an absolute root), and two `t.Chdir`-based tests skip on hosts without a discriminating tmpfs while the always-running `internal/pathdurability` unit test keeps the core rule covered everywhere.

- No non-blocking follow-ups.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `d7e9fd4c` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/32903555734

### Review Iterations
1 review pass performed — approve on the first pass, 0 fix iterations. The only maintainer-side action was a rebase onto the latest `main` to refresh the merge base; the contributor's commit was preserved unchanged (identical patch).

---

Issue #2348, opened 2026-05-19
mol-dog-compactor quarantines busy rigs from benign concurrent-write race; quarantined rigs accumulate unbounded until manual intervention


tim-white87:
**Symptom**

The `mol-dog-compactor` order quarantines an active rig after a routine 24h compaction run if any concurrent write happens between the script's preflight row count and postflight row count. Once quarantined, the rig's commits + chunks accumulate without compaction or DOLT_GC until a human deletes the quarantine marker by hand. On a busy single-host city this can sit for days before anyone notices.

This is the underlying trigger that fed #2158 in our case: gol got quarantined at `2026-05-16T02:13:04Z` during routine compaction, accumulated ~10K extra commits over 3 days, and once disk pressure crossed the conjoin threshold the #2158 cascade fired and consumed the rest of the volume.

**Reproduction**

1. Have a city with at least one rig that has active agents writing beads (witness, refinery, polecat).
2. Let the `mol-dog-compactor` order tick on its normal 24h cadence.
3. With non-zero probability per tick proportional to rig write activity, the compactor's postflight count differs from its preflight count and the rig is quarantined with marker reason `post-flatten row count changed`.

Quarantine path in `examples/dolt/commands/compact/run.sh` (from a recent main):

```
4. Re-check post-flatten row counts and database value hash. Row-count
   decreases fail before full GC. Row-count increases are held pending
   value-hash verification. Any value-hash drift is quarantined before
   full GC until preservation is proven.
```

In a single-host multi-agent setup, concurrent writes between the preflight and postflight COUNT(\*) queries are routine, not exceptional.

**Evidence from this incident**

Quarantine marker found on our system:

```
/.../.gc/runtime/packs/dolt/compact-quarantine/gol
  db=gol
  reason=post-flatten row count changed
  created_at=2026-05-16T02:13:04Z
```

dolt.log during the same 02:10–02:13 window showed steady connection traffic on gol from multiple agents (`level=warning msg="error running query" connectionDb=gol ... error="nothing to commit"` — benign, but confirms active connections were doing tx work).

After clearing the marker and re-running with the supervisor stopped (no concurrent writers), the compactor ran cleanly:

```
compact: db=gol commits=10033 root=jconj940h09g9ok6c6sn8vuqijioo03i tables=25 — flattening...
compact: db=gol — running DOLT_GC --full...
compact: db=gol commits=10033->2 duration=2s — ok
```

[...]

PR #5493, opened 2026-08-21
fix(dolt): a quarantine alert that never lands should say so (ga-u2wiy)

julianknutsen:
Every Dolt-backed city on the dev fleet has compaction fail-closed on an integrity quarantine marker, and not one of the six quarantines was ever reported to a human. Two independent reasons, both fixed here.

## 1. The alert failed silently for a month

The compact-quarantine alert defaults to recipient `mayor`. None of the affected cities has a mayor, so `gc mail send` failed on every cycle, the failure was absorbed into `notify_count=0`, and the markers sat at `seen_count=177 / notify_count=0` for up to 35 days without ever saying why.

The marker is the durable artifact an operator actually reads, so it now carries the cause:

- `mail_compact_quarantine_alert` captures the send's stderr and prints a distinct `quarantine alert did not reach recipient <who>: <why>` line.
- `record_quarantine_notify_state` stamps `last_notify_error` on failure and clears it on delivery.

A marker stuck at `notify_count=0` now explains itself instead of leaving a month-long silence to be found by accident.

## 2. `--gc-only` skipped the reporting path entirely

The scheduled compact path and the bare-GC path both route an existing quarantine through `report_existing_quarantine` — emit `dolt.compact.quarantine`, send the gated alert, record the sighting. The `--gc-only` reclaim path printed to stdout and returned, so a city whose reclaim ran on that path held a quarantine entirely off the bus.

That is the measured difference across this box: trust 0 quarantine events in 4 days, platform 0 in 26, orchestration 1 in 13, substrate 4 in 33 — against gas-city-inc's 12. gas-city-inc is the one city taking a scheduled path.

## Scope

Bookkeeping still never blocks compaction, and dedup is unchanged. This adds the missing cause and closes the one path that bypassed reporting; it does not change **who** is mailed or **when**.

The `mayor`-as-default recipient question is a separate design decision and stays open on ga-u2wiy.

## Tests

`dog_exec_scripts_test.go` covers both: an alert whose send fails records `last_notify_error` and prints the recipient and reason, and `--gc-only` on an existing quarantine takes the same reporting path as the scheduled run.

---
- Related to #2348 — addresses the operator-visibility half of that issue: a quarantine alert that fails to send now records and prints why it did not land, and the --gc-only reclaim path emits the quarantine event and records the sighting instead of only printing to stdout. It does not address the concurrent-write race that creates the false-positive quarantines in the first place, nor gc doctor surfacing quarantine markers.

### DIFF
--- a/examples/bd/dolt/commands/compact/run.sh
+++ b/examples/bd/dolt/commands/compact/run.sh
@@ -3101,10 +3101,9 @@ gc_only_database() {
   fi
 
   if has_compact_marker "$quarantine_dir" "$db"; then
-    quarantine_marker=$(compact_marker_path "$quarantine_dir" "$db")
-    quarantine_reason=$(compact_marker_value "$quarantine_dir" "$db" reason || true)
-    quarantine_created_at=$(compact_marker_value "$quarantine_dir" "$db" created_at || true)
-    print_existing_quarantine_marker "$db" "$quarantine_marker" "$quarantine_reason" "$quarantine_created_at"
+    # Same operator-visible state as a scheduled refusal, so it reports the
+    # same way: stdout alone leaves the quarantine off the bus entirely.
+    report_existing_quarantine "$db"
     return 1
   fi
 
--- a/examples/bd/dolt/commands/compact/run.sh
+++ b/examples/bd/dolt/commands/compact/run.sh
@@ -1508,7 +1508,7 @@ write_compact_marker() {
       if mail_compact_quarantine_alert "$db" "compact-quarantine" "$marker_path" "$reason" "$created_at"; then
         record_marker_notify_state "$quarantine_dir" "$db" "$reason" 1
       else
-        record_marker_notify_state "$quarantine_dir" "$db" "$reason" 0
+        record_marker_notify_state "$quarantine_dir" "$db" "$reason" 0 "$quarantine_notify_error"
       fi
     else
       record_marker_notify_state "$quarantine_dir" "$db" "$reason" 0
--- a/examples/bd/dolt/commands/compact/run.sh
+++ b/examples/bd/dolt/commands/compact/run.sh
@@ -1531,6 +1531,14 @@ emit_compact_quarantine_event() {
 # marker. Empty means delivered; callers reset it before each attempt.
 quarantine_notify_error=""
 
+# mail_compact_quarantine_alert DB TYPE MARKER REASON [CREATED_AT]
+#   Send the quarantine alert; return 0 on delivery, 1 on a failed send with
+#   the reason recorded in quarantine_notify_error. CONTRACT: call this ONLY as
+#   a condition (e.g. `if mail_compact_quarantine_alert ...`), never as a bare
+#   statement. The internal _ca_err=$(gc mail send ...) capture runs under
+#   `set -eu`, so a bare-statement call would abort the whole compaction run on
+#   a failed send -- the exact silent mid-cycle failure this alert exists to
+#   prevent.
 mail_compact_quarantine_alert() {
   _ca_db="$1"
   _ca_type="$2"
--- a/examples/bd/dolt/commands/compact/run.sh
+++ b/examples/bd/dolt/commands/compact/run.sh
@@ -1539,7 +1539,11 @@ mail_compact_quarantine_alert() {
   _ca_created_at="${5:-<unknown>}"
   _ca_msg="db=$_ca_db type=$_ca_type marker=$_ca_path reason=$_ca_reason created_at=$_ca_created_at recipient=$compact_alert_to"
   quarantine_notify_error=""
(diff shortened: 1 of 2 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277227871

quad341 (dismissed):
@julianknutsen — merging this as-is. Thank you for chasing the actual cause instead of just changing the default recipient.

What I appreciate most is that you found two independent failures behind one symptom and kept them separate: the swallowed stderr in `mail_compact_quarantine_alert`, and the missing `report_existing_quarantine` call in `gc_only_database`. They produced the same month-long silence but they're unrelated bugs, and reviewing them separately is why this was easy to verify.

`TestCompactScriptUndeliverableQuarantineAlertRecordsWhyInMarker` is the test that sold me. Asserting that a *delivered* alert clears `last_notify_error` — not just that a failed one records it — is the assertion most people leave out, and it's exactly what proves the dedup gate can't strand a marker with a stale cause. I traced that interaction by hand before I got to your test. Same for catching your own swallowed-stdout regression and writing a test for it.

Two things I noted but am not asking you to change: stderr on a *successful* send is captured and dropped, and `--gc-only --dry-run` now reports and mails — but `flatten_database` and `bare_gc_database` have always done that, so you've made it consistent. If the dry-run side effect bugs you, that's its own bead across all three paths.

Leaving the `mayor`-default question on ga-u2wiy where you put it.


---

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277246413

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR makes the Dolt compact
quarantine-alert path report its failures instead of silently swallowing them —
so an alert that never lands now says *why*, a `--gc-only` run on an
already-quarantined DB actually raises the alert, and the success confirmation
is no longer eaten by stderr capture. That directly closes the silent-failure
operability gap behind `ga-u2wiy` (quarantine alerts fail-closed for weeks with
zero notifications ever sent) and the swallowed-confirmation bug `ga-cbf3jf`,
which is exactly the class of "we never found out it broke" problem this code
exists to prevent.

PR branch.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **`--gc-only` on a quarantined DB fired nothing on the bus/mail (ga-u2wiy):**
  `gc_only_database` now routes an existing quarantine through
  `report_existing_quarantine` — the same call shape as the two scheduled
  paths — instead of only printing a stdout marker, so the alert and sighting
  count actually fire.
- **Undelivered quarantine alerts left no record of *why* (ga-u2wiy):** a new
  `last_notify_error` field captures the `gc mail send` stderr on failure
  (newline-flattened and capped at 200 chars) and is cleared on a successful
  emit; existing markers gain the field on their next write.
- **stderr capture swallowed the success confirmation (ga-cbf3jf):** the
  `exec 4>&1` / `… 2>&1 1>&4` / `exec 4>&-` fd swap routes the command's real
  stdout to fd 4 while only stderr is captured, and fd 4 is closed on both
  paths so nothing leaks.

Review findings addressed:
- **[Error Handling & Resilience] Alert errexit-safety depends on caller
  context** (minor): `mail_compact_quarantine_alert` is errexit-safe *only*
  because all three callers invoke it as an `if` condition; a future
  bare-statement caller would abort the whole compaction run mid-cycle at the
  failing capture — the same silent-failure class this PR removes. Addressed in
  maintainer commit `981e0c9` (`docs(dolt): document
  mail_compact_quarantine_alert errexit contract`), which documents the "call
  only in a condition" contract above the function.

- **[Contract & Interface Fidelity] Stale-marker alert-failure omits captured
  error** (minor, out of scope / pre-existing): the stale-marker branch
  (`run.sh` ~1882/1885) still records no `last_notify_error` when *its* alert
  fails, so the "record why it never landed" behavior this PR adds for the
  quarantine path stays asymmetric for stale markers. It is byte-identical to
  `origin/main` — this PR neither introduced nor regressed it — and the proper
  fix is more than a one-line pass-through (it needs a `quarantine_notify_error`
  reset on the stale path plus its own test), so it was intentionally left out
  of scope. Filed as **ga-n7ggz** for follow-up.

### Maintainer Changes
Maintainer fixes were pushed directly to the PR branch (contributor commits
preserved; branch rebased onto the latest `main`):
- `981e0c9 docs(dolt): document mail_compact_quarantine_alert errexit contract`
  — `examples/bd/dolt/commands/compact/run.sh` | 8 ++++++++ (1 file, +8).

### Final Review Status
Ready for the merge queue: final head `981e0c9` has passing required GitHub

CI: https://github.com/gastownhall/gascity/actions/runs/33076429940

### Review Iterations
1 review pass performed — approved with no fix iterations; the maintainer
errexit-contract doc commit and the rebase onto latest `main` were the only
maintainer-side changes.

---

PR #5503, opened 2026-08-22
fix(dispatch): preserve check trust across pack pin upgrades

jamelt:
## Summary

- preserve the materialization-time trust decision for Ralph exec checks when a pack pin upgrade moves current FormulaSearchPaths to a new content-addressed cache entry
- require the workflow-root formula source and SHA-256 to still match, and require the check path to remain inside the same canonical 64-hex repo-cache entry
- keep the normal current-pin path unchanged; the fallback performs one root lookup and one small formula read only for an otherwise-untrusted absolute check path

## Reproduction

A graph.v2 workflow materialized from pack pin A stores an absolute gc.check_path under cache entry A. Upgrading the city to pin B immediately removes A from FormulaSearchPaths while the workflow is still active. The dispatcher then quarantines the valid control with an escapes-trusted-roots error. In the customer-zero incident this stranded 25 of 30 active workflow roots and quarantined 68 controls while feature branches continued to accumulate.

The workflow already records gc.formula_source and gc.formula_hash at materialization time. This change uses that durable provenance instead of scanning Git history, parsing lockfiles, or trusting every cached repository.

## Trust boundary

The fallback fails closed unless:

- formula source and check path are both under GC_HOME/cache/repos
- the cache directory name is canonical lowercase 64-hex
- both paths resolve into the same cache entry
- the source file bytes still match gc.formula_hash
- the candidate remains inside roots derived from that materialized formula layer

Existing symlink-aware containment remains in force. Missing or mismatched evidence preserves the current rejection.

## Tests

- regression: a materialized check from the previous pin executes after current search paths move to a new entry
- negative cases: hash mismatch, missing hash, and a check from a different cache entry remain rejected
- full internal/dispatch suite
- adjacent internal/config, internal/formula, and internal/molecule suites
- make test-fast-parallel, twice including the pre-push gate
- pre-commit lint and go vet ./...


quad341 (approved):
@jamelt — reaching for the `gc.formula_source`/`gc.formula_hash` the root bead already carries, instead of scanning Git history or parsing lockfiles, is exactly the right instinct here. You used provenance we durably recorded at materialization time rather than re-deriving trust from something mutable, and the fallback is properly fail-closed: every guard returns `nil`, so the current-pin path is untouched and a bad path can't come out better off than before.

I checked the claims rather than taking them. The root bead really does stamp both keys (which is why walking `gc.root_bead_id` is necessary), `formula.ContentHash` really does recompute over the same bytes the parser hashed, `RepoCacheKey` really does emit 64 lowercase hex, and `pathutil.PathWithin` really does keep symlink containment in force. All four hold.

Two small things I'll take on my side and merge: tightening the doc comment — for the common `formulas/` layout the effective grant is the whole cache entry, which is defensible but broader than the comment reads — and adding negative cases for the cache-root and hex-name guards. Maintainer chores, not yours. I'll also let the push re-run full CI, which hasn't actually executed on this head.

Thanks for the reproduction detail; 25 roots and 68 quarantined controls made the priority obvious.


---

PR #5505, opened 2026-08-22
fix(runtime): release drain claims by session alias

jamelt:
## Summary

- Resolve a runtime session name to its durable session bead ID before the drain-ack held-claim release pass.
- Bounded pool sessions commonly have a `ci-wisp-*` bead ID and a distinct runtime/tmux name in metadata. A direct `Get(runtimeName)` silently missed those sessions, so drain-ack could succeed while leaving newly claimed work in progress on a stopped worker.
- Reuse the existing session resolver; this adds no poller, process, network call, or controller hot-path work.

This was reproduced twice with live bounded-pool workers: the runtime session disappeared after drain-ack while its work remained in progress under the runtime alias. Closing the durable session bead through `gc session close` immediately reopened the work, confirming the identity-resolution gap.

## Testing

- [x] `go test -count=1 ./cmd/gc -run 'DrainAck'`
- [x] `go test -count=1 ./internal/testpolicy/resourcecensus -run '^TestRepositoryLedgerMatchesCensusAndDocumentation$'`
- [x] `go vet ./cmd/gc`
- [x] `make lint-affected`
- [x] Repository pre-push fast gate (`./scripts/test-local-parallel fast`; all jobs passed)
- [ ] `make check` (the pre-push fast gate, affected lint, vet, and focused tests were run instead)
- [ ] `make check-docs` (no documentation changed)
- [ ] `make test-integration` (not run; the regression exercises the exact alias/store boundary with the in-memory store)

## Checklist

- [x] No issue is needed: the production failure, root cause, and regression are self-contained here.
- [x] Added a regression test for the behavior change.
- [x] No user-facing documentation change is required.
- [x] No breaking change or migration is introduced.


quad341 (approved):
@jamelt — thank you for this one. The `Get(runtimeName)` vs. `ci-wisp-*` bead-ID mismatch is the kind of bug that hides forever: drain-ack returns success, the worker exits clean, and the only symptom is work stuck `in_progress` under a name nothing will ever look up again. Routing through `resolveSessionID` is the right call — it's a strict superset of the old lookup, and once you hold the durable bead, `sessionAssignmentIdentifiers` puts the runtime alias into the release set for free. Your regression test is doing real work too: it builds the actual pool shape and fatals if the fixture ever degenerates into testing the trivial path.

Two small things I'll take on my side rather than bounce back to you. The rig-store construction ended up above the resolution, so an unresolvable session now opens a store per rig before returning — I'll make that lazy again. And I'll add a control case for the ID-equals-runtime-name shape, since the wrapper is now the only door to both.

Also worth flagging: no test suite actually ran on this head — the dispatch checks all skipped, so that green list is cosmetic. I'll get the real suite running before merge. Your fix, your credit; I'm just absorbing the integration bits.


---

PR #4890, opened 2026-07-31
Isolate doctor custom-types tests from machine Dolt config

quad341:
## What this changes

The `gc doctor` custom-types tests now run with a test-owned HOME, so a machine-level `~/.beads/config.yaml` enabling Dolt shared-server mode cannot redirect their `bd` subprocesses to a live fleet server. Both fixtures also clear the shared-server environment selectors, and a regression test verifies embedded-store metadata and the absence of shared-server output.

The test-resource ledger is updated for the new subprocess and bounded cleanup retry. Production doctor behavior is unchanged.

## Review notes

- This is test-only behavior plus the mechanically paired resource-census mirrors.
- No operator config, shared-server database, API, or runtime behavior changes.
- The cleanup retry is bounded and runs before `t.TempDir` final cleanup.

## Test plan

- [x] Run custom-types tests with a machine HOME that enables the shared Dolt server; verify all tests pass and the server database list is unchanged.
- [x] Run the resource-census mirror check, repository build and vet, and the fast CI baseline.
- [x] Audit process coverage against the merge base under the same host configuration, then repeat the affected tests with a clean HOME.
- [x] Release gate: [`release-gates/ga-vn396k-doctor-custom-types-home-isolation-gate.md`](release-gates/ga-vn396k-doctor-custom-types-home-isolation-gate.md)

PR #5506, opened 2026-08-22
fix(import): name the repo cache root an install wrote into

### LINKED ISSUE #5836: gc import install: the success line never names the repo cache root, so a GC_HOME split between shells is undiagnosable
## Symptom
`gc import install` reports success, and every later config load reports `remote import X is locked but not cached at <dir>; run "gc import install"` — advice that cannot work, because the install keeps writing somewhere else. Nothing ever prints where.

## Root cause
`GC_HOME` resolves per process (`internal/config/implicit.go`). A shell that exports it differently from the process that later loads the config — in our case a login-shell profile that a non-login shell never sourced — installs into one repo cache root while every later resolve searches another. The read side names the directory it searched; the write side prints no root at all, so the two halves of the split cannot be lined up from the output.

## Reproduction
Export `GC_HOME` in one shell, run `gc import install`; load the city from a shell without that export. Compare the "not cached at" path with the location of the clones — the success line gave you nothing to compare it against.

## Environment

## Proposed fix
#5506 — name the repo cache root on the success line (`Installed N remote import(s) into <root>`). The underlying hygiene fix is propagating `GC_HOME` consistently (`gc supervisor install` already pins it into the unit); this only makes the split a one-line diagnosis.

A3Ackerman:
## What

`gc import install`'s success line now names the repo cache root the clones landed in:

    Installed 3 remote import(s) into /home/alex/.gc/cache/repos

`gc import upgrade` clones through the same `packman.EnsureRepoInCache` path, so its success line names the root too (`Upgraded 3 remote import(s) into …`, `Upgraded import "worker" into …`), via one shared suffix helper.

Both commands resolve the root once before their first clone; an unresolvable root is an error (`gc import install: resolving repo cache root: …`) rather than a success line naming nothing — reachable only in a test binary with `GC_HOME` unset.

## Why

`GC_HOME` resolves per process (`internal/config/implicit.go`), so a shell that exports it differently from the process that later loads the config — in our case a login-shell profile that a non-login shell never sourced — installs into one cache root while every later resolve searches another. The read side already names the directory it searched (`remote import X is locked but not cached at <dir>; run "gc import install"`), but that advice cannot work while the install keeps writing somewhere else, and nothing ever printed where that was. With the writer's root on the success line and the reader's directory in the error, the mismatch becomes a one-line diagnosis with no new state.

## What this PR no longer proposes

The earlier revision recorded the cache root per city, refused installs into a "foreign" root, and added `--rebind-cache-root` to move a city between roots (~690 lines, filed as P1). On reflection that guarded against an environment-hygiene lapse on our side — `gc supervisor install` already pins `GC_HOME` into the unit for exactly this reason — and the right fix for the lapse is to propagate `GC_HOME` consistently, not to add persistent state to every city. That revision is parked on a branch; this PR keeps only the part that is plainly correct for every deployment. The P1 framing is withdrawn with it.

## Test

`TestDoImportInstallWithNoImportsSucceeds` and `TestDoImportUpgradeTargetsDefaultRigImport` assert the exact success line against `<GC_HOME>/cache/repos` computed from the `GC_HOME` that `clearGCEnv` pins, not from `packman.RepoCacheRoot()`, and fail outright if `GC_HOME` is unset. `TestImportWritersRefuseUnresolvableRepoCacheRoot` pins the fail-closed branch for both commands.

https://claude.ai/code/session_01Wtk5dCw6TtDSD5UzPDdSwa

Fixes #5836.


A3Ackerman:
Rebased onto current main (a4361e58) and fixed the Integration failure — the branch had gone stale on two fronts:

**Conflicts resolved (both keep upstream's newer behavior intact):**
- `internal/config/pack_include.go`: kept the `nonBlocking` parameter #5504's bounded-network-git fix added to `validateInstalledRemoteCacheLocked`, while retaining this PR's `annotateCacheRootMismatch` wrap at both call sites.
- `cmd/gc/cmd_import.go`: #5618's post-install composed-config validation now runs first (fail closed), then this PR's cache-root-naming success line.

**The `packages-core` failure was the resource census ratchet**, and it was right to fire: `TestGcHomeSplitWedgeIsRefused` shells real git. Fixed per the ledger's own contract rather than suppressed (new commit 42fc5a90):
- `gitInTest` became a closure inside the owning test — the census attributes calls to the enclosing top-level test, and Medium ownership only covers calls lexically inside the declared owner;
- exact `[[medium]]` row in `test/test-resources.toml` + the mirrored `bootstrapPolicy` row, with the audit/debt baselines bumped by the one call/file — same shape as #4890's doctor-test addition;
- TESTING.md ledger block regenerated with `-update`.

The test has to shell git: it is an end-to-end proof that a foreign-rooted install is refused against a real clone layout (same justification shape as the ledger's existing real-child-process row). Its `t.Setenv` calls need no declaration — the census deliberately excludes `testing.T` Setenv.

`owner_bead` on the new row references this PR; happy to repoint it wherever the census migration prefers external contributions to hang.

Locally green: `internal/packman`, `internal/config`, `cmd/gc -run Import`, the census test, and `test/docsync`.

A3Ackerman:
Reframed and force-pushed to a much smaller change: the success line now names the repo cache root, and that is all. The refusal guard, the per-city `.gc/repo-cache-root` record and `--rebind-cache-root` are withdrawn — on re-review they guarded against our own `GC_HOME` propagation lapse (a login-profile export a non-login shell never saw), which `gc supervisor install`'s unit-pinned `GC_HOME` already exists to prevent. The earlier revision is parked on my fork (`held/packman-cache-root-guard`) in case anyone wants to revisit the deterministic prefer-recorded-root alternative; the P1 label no longer fits and can be dropped.

A3Ackerman:
Green after the second pass: whole matrix passing (Evidence / critical-path suites=pass CI / required=pass CI / integration=pass ). The only pushes since the last green were the review-driven changes described in the body.

quad341 (approved):
@A3Ackerman — thank you for cutting this down. Going from the ~690-line per-city root record plus `--rebind-cache-root` to "the success line names the root," on your own re-review, with the reasoning written out and the earlier revision parked on a branch — that's a hard call, and you made it correctly. `gc supervisor install` pinning `GC_HOME` into the unit really is the right layer for that lapse.

The details hold up. I traced `packman.RepoCacheRoot` and it is the single resolver that both the writer and `pack_include.go`'s "locked but not cached at" error go through, so the two lines genuinely pair into a one-line diagnosis. Computing the expected root in the test from the pinned `GC_HOME` rather than calling `RepoCacheRoot()` — so the assertion doesn't share a resolver with the code under test — is a nice touch.

Main has moved again since your last rebase and it conflicts in `doImportInstall`. That one's on us: I'll take the rebase, keep upstream's ordering with your suffix on top and the root resolve ahead of the first clone, and re-run CI. Dropping `priority/p1` as you asked — the triage comment describes the design you withdrew.


---

PR #5507, opened 2026-08-22
test: close two local-only test gaps (docsync scratch roots, label-order cache path)

julianknutsen:
Two small test-only changes that were sitting in a local working tree and are not upstream.

**`test/docsync`: skip bead-named scratch roots.** `TestDocDirCoverage` walks
top-level directories and fails on any it does not recognise. Agents routinely
leave bead-id-named scratch directories (`ga-…`, `gcg-…`, `mc-…`) at the repo
root, which turns a clean local checkout red for a reason that has nothing to do
with docs. The exemption is an explicit prefix list rather than a shape match on
purpose: a doc tree may legitimately be hyphenated (`release-gates`), and a
shape matcher would silently exempt it from the coverage this file exists to
enforce.

**`internal/beads`: pin the label-order path of `cacheEventConflictsCurrent`.**
The set-compare fix for labels is already on `main` (`caching_store_events.go`
uses `stringSetEqual`), but nothing fails if the comparison goes back to being
order-sensitive. A pure reorder must not read as a conflict — when it does, the
cache reconciler re-absorbs every bead whose labels came back in a different
order.

Verified: `go test ./internal/beads/ -run TestCacheEventConflictsCurrentIgnoresLabelOrder`
and the full `make test-fast-parallel` fast suite (via the pre-push hook) both pass.

### DIFF
--- a/internal/beads/caching_store_setequal_internal_test.go
+++ b/internal/beads/caching_store_setequal_internal_test.go
@@ -1,6 +1,9 @@
 package beads
 
-import "testing"
+import (
+	"encoding/json"
+	"testing"
+)
 
 // Reordered label/needs/dependency sets must NOT read as a change: the Dolt gcg
 // rig store does not guarantee a stable element order across scans, so an
@@ -46,6 +49,26 @@ func TestBeadChangedIgnoresSetOrder(t *testing.T) {
 	}
 	if !depsChanged(base.Dependencies, depChanged.Dependencies) {
 		t.Error("depsChanged = false when a dependency target changed; want true")
+	}
+}
+
+func TestCacheEventConflictsCurrentIgnoresLabelOrder(t *testing.T) {
+	current := Bead{
+		ID:     "gcg-wisp-x",
+		Title:  "t",
+		Status: "open",
+		Type:   "task",
+		Labels: []string{"a", "b", "c"},
+	}
+	patch := current
+	patch.Labels = []string{"c", "a", "b"}
+
+	fields := map[string]json.RawMessage{
+		"labels": json.RawMessage(`["c","a","b"]`),
+	}
+
+	if cacheEventConflictsCurrent(current, patch, fields) {
+		t.Fatal("cacheEventConflictsCurrent = true for a pure label reorder; want false")
 	}
 }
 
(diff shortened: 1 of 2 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277216742

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR closes two local-only test-coverage gaps: it pins the order-insensitive label comparison on the `cacheEventConflictsCurrent` event-conflict path, and it widens the docsync doc-dir coverage check to skip all reserved bead-id scratch roots (`ga-`/`gcg-`/`mc-`) rather than only `ga-`. Both were spots where a real regression could have slipped past CI unnoticed, and the new tests close them additively with zero production-code change.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- The `cacheEventConflictsCurrent` label-order path was previously unpinned — `TestBeadChangedIgnoresSetOrder` only exercised the reconcile path, never the event-conflict path. The new `TestCacheEventConflictsCurrentIgnoresLabelOrder` supplies `fields = {"labels"}` so control reaches exactly the labels branch, pinning the order-insensitive `stringSetEqual` compare on that path.
- `TestDocDirCoverage` previously exempted only `ga-` scratch roots. The new `beadScratchPrefixes` + `isBeadScratchRoot` widen the exemption to the three reserved bead-id namespaces (`ga-`/`gcg-`/`mc-`) additively; the non-exempt control path is byte-for-byte unchanged, so coverage still fails on a genuinely new, unaccounted doc tree.

- **Test evidence quality (nit):** the new label-order test asserts only the negative direction (a pure label reorder does not conflict), not the positive direction (a real label delta through the same event path still conflicts). Optional hardening — add one positive assertion in the same test. Filed as `ga-63qy0`: add positive-direction assertion to the `cacheEventConflictsCurrent` label-order test. Not addressed in this PR.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `d94d1c9` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/32562784577

### Review Iterations
1 review pass performed; approved with no fix iterations — the single finding was a non-gating nit, carried forward as the follow-up above.

---

PR #5509, opened 2026-08-22
fix(dolt): make offsite backup failure visible and its bound configurable

swedeinasia-flow:
## Problem

`mol-dog-backup`'s step 3 rsyncs the backup artifact dir to `GC_BACKUP_OFFSITE_PATH` under a hardcoded 300s bound. On failure it sets `OFFSITE_STATUS="failed (non-fatal)"` — which is reported nowhere except the summary line. Step 4 escalates only on per-database sync failures, so **a persistently failing offsite step is silent.**

Two consequences, both observed in a live installation:

1. **300s is a guess about someone else's disk.** A large artifact set going to a network mount, a userspace file provider (OneDrive/Dropbox/iCloud), or a host under load can exceed it while being perfectly healthy.

2. **rsync transfers in lexical order**, so a truncated run always completes the same leading entries and always drops the same trailing ones. The offsite copy silently stops tracking part of the city while the earlier databases keep updating — which means a spot check of the destination looks fresh.

In the installation that prompted this, three of four databases stopped publishing offsite and the condition persisted for **three days** before a human noticed by accident. The local backup was fine throughout; only the offsite copy had stalled, and nothing said so.

## Change

- the bound becomes `GC_BACKUP_OFFSITE_TIMEOUT`, **defaulting to the current 300s** so existing behavior is unchanged;
- a non-`ok`, non-`skipped` offsite status now escalates through `dolt_escalate`, naming the status and the variable to raise.

Non-fatal is kept — the local backup did succeed, so the run should not fail. **Silent is what changes.** An offsite step that fails without telling anyone is worse than one that is switched off, because the operator believes there is a remote copy and discovers otherwise only when they need it.

## Scope

One file, 29 insertions. No behavior change for anyone who does not set the new variable, beyond gaining an escalation they previously did not get.

Local `make test-fast-parallel` has pre-existing failures on macOS unrelated to this change (`internal/pathdurability` tmpdir classification, `internal/pidutil` argv readability); both reproduce on unmodified `main` at the parent commit.


quad341 (approved):
@swedeinasia-flow — the insight that earns this one its merge is in your second paragraph: because rsync transfers in lexical order, a truncated run drops the *same* trailing databases every time, so the offsite copy looks fresh to a spot check while part of the city has quietly stopped being covered. That's a mean failure mode, and three days is a rough way to find it.

Keeping the default at 300s and keeping the step non-fatal is the right call — "non-fatal it is; silent it must not be" is exactly the distinction worth drawing. And faking `timeout` to prove the configured bound actually reaches rsync tests the wiring instead of restating a constant.

One thing I'll add on our side before merging. The order itself is bounded at 1800s and the modelled budget already spends 1530s, so an operator who follows your new message and raises the variable much past ~570s gets the run killed mid-rsync — before your escalation can fire. I'll name that ceiling in the message, reject a zero or garbage value (GNU `timeout 0` disables the bound outright), and fix a now-stale comment in the order file.

Your commits, your credit. I'm just adding the guardrail your own advice needs.


---

PR #5510, opened 2026-08-22
Fix implicit formula sling re-routes for molecule-attached beads

quad341:
## What this changes

`gc sling <target> <bead> --nudge` now completes a plain bead re-route when the target's implicit `default_sling_formula` cannot be attached because the bead already belongs to a live molecule. The command emits a warning and still records the new route instead of failing before the handoff reaches its intended agent.

The behavior is consistent across the legacy and graph-v2 formula compilers. Explicit `--on` and `--formula` requests remain fail-closed: they still refuse to attach a second live workflow. On graph-v2 fallback, the unused synthetic input convoy is closed so the successful plain route does not leave claimable workflow debris behind.

## Review notes

- The new typed attachment error distinguishes this exact fallback case from live graph-v2 workflow conflicts and unrelated storage errors.
- Only an implicit default formula enables fallback; explicit formula attachment is unchanged.
- Batch fan-out behavior is intentionally unchanged because partial fallback needs a separate routing/reporting design.
- No configuration, persistence schema, API, or migration changes are included.

## Test plan

- [x] Run both new legacy and graph-v2 fallback regressions by exact test name; verify plain routing, warning emission, and convoy cleanup.
- [x] Run the explicit-formula hard-fail guard and the formula-v2 coupling guard.
- [x] Run the repository's 40-job local CI-equivalent suite with the pinned container environment; audit raw failures and exact-base attribution in the gate record.
- [x] Run workflow-policy, affected lint, full build, full vet, formatting, and diff checks.
- [x] Release gate: [`release-gates/ga-e71kz1-sling-attached-molecule-reroute-gate.md`](release-gates/ga-e71kz1-sling-attached-molecule-reroute-gate.md)

### DIFF
--- a/cmd/gc/cmd_sling.go
+++ b/cmd/gc/cmd_sling.go
@@ -1784,10 +1784,14 @@ func dryRunSingle(opts slingOpts, deps slingDeps, querier BeadQuerier, stdout, s
 			}
 			w("")
 		} else if !opts.NoFormula && a.EffectiveDefaultSlingFormula() != "" {
+			// Report-only pre-check: unlike explicit --on, an implicit
+			// default formula no longer hard-fails on a pre-existing
+			// molecule/wisp -- the live path skips the attach and routes
+			// the bead plainly -- so the preview must not predict an
+			// error the real run will not produce.
+			var blockingLabel, blockingID string
 			if preCheck {
-				if rc := dryRunReportBlockingMolecule(opts, deps, querier, stderr); rc != 0 {
-					return rc
-				}
+				blockingLabel, blockingID = sling.FindBlockingMolecule(querier, opts.BeadOrFormula, deps.Store)
 			}
 			w("Default formula:")
 			w("  Formula: " + a.EffectiveDefaultSlingFormula())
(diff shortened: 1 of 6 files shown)


quad341:
@quad341 — this is a nicely disciplined fix. You resisted the easy version: rather than loosening `CheckNoMoleculeChildren` for every caller, you introduced `MoleculeAttachedError` so the fallback can only ever fire on the one case it's meant for, then threaded an explicit flag so `--on` and `--formula` keep their fail-closed behavior. I traced it and the narrowing genuinely holds — a live graph-v2 workflow conflict still hard-fails through `checkLegacySourceWorkflowConflict`, because the allow-live-workflow check `continue`s past workflow attachments before `errors.As` ever sees them.

Catching that the graph-v2 branch mints its synthetic input convoy *before* the attachment check — so a successful fallback would leak an open, claim-attracting convoy — is subtle, and pinning it with an assertion rather than fixing it quietly is the right instinct.

One thing I'll take on my side: the dry-run preview at `cmd_sling.go:1786` wasn't updated, so `--dry-run` still exits 1 predicting a failure the real run no longer produces. I'll turn that into a skip note, add the matching dry-run test, and re-run the sling and cmd/gc suites before merging. Nothing needed from you — that's integration work that fell out of your change, and it's ours to absorb.

Thanks for the gate record too; the exact-base attribution made verifying this much faster.


---

Issue #4422, opened 2026-07-19
bug: gc mail archive/delete irrecoverably destroy message beads; gc mail has no non-destructive drain verb


jacobhausler:
## Summary

`gc mail archive` and `gc mail delete` both irrecoverably destroy the underlying message bead — they do not close it, move it to an archive location, or retain it in any queryable form. Because an un-drained wisp is re-served by `gc hook --claim` forever, a working drain is mandatory for every consumed message. Agents looking at `gc mail` for a safe drain find only destructive verbs and no `gc mail close`. The safe drain exists (`bd close <id>`) but lives in a different namespace with no in-namespace pointer until a reader happens to see it in `archive --help`.

Together these make routine mail hygiene a data-safety hazard: the natural-seeming path destroys data, and the safe path is invisible unless you already know to look outside `gc mail`.

---

### Control: bd show renders closed beads as CLOSED

```
$ bd show ra-wisp-rncwu0   # a known-closed bead
✓ ra-wisp-rncwu0 · DECOY-SAFE sentinel body   [● P2 · CLOSED]
```

"Not found" after archive is NOT the closed-bead render. The bead is genuinely absent.

### Decoy A — destroy path (gc mail archive)

```
$ gc mail send thad -s "DECOY-DESTROY sentinel body"
Sent message ra-wisp-u37dg9 to thad

$ bd show ra-wisp-u37dg9
○ ra-wisp-u37dg9 · DECOY-DESTROY sentinel body   [● P2 · OPEN]
  Type: message

$ gc mail read ra-wisp-u37dg9
ID: ra-wisp-u37dg9  From: thad  Subject: DECOY-DESTROY sentinel body

$ gc mail archive ra-wisp-u37dg9
Archived message ra-wisp-u37dg9

$ gc mail read ra-wisp-u37dg9
gc mail read: beadmail read: bead "ra-wisp-u37dg9": bead not found

$ gc mail peek ra-wisp-u37dg9
gc mail peek: beadmail get: bead "ra-wisp-u37dg9": bead not found

$ bd show ra-wisp-u37dg9
Error fetching ra-wisp-u37dg9: no issue found matching "ra-wisp-u37dg9"
```

All three probes return not-found. The bead is gone.

### Decoy B — safe path (bd close)

```
$ gc mail send thad -s "DECOY-SAFE sentinel body"
Sent message ra-wisp-rncwu0 to thad

$ bd close ra-wisp-rncwu0 -r "testing safe drain"
[...]

PR #5512, opened 2026-08-22
fix(mail): archiving a message no longer makes it unrecoverable — Archive stamps a close reason, peek reads through it

csauer02-personal-user:
### Problem

`Provider.Archive` and `Provider.ArchiveMatching` close a message bead without stamping a
`close_reason`. `isRemovedMessageBead` (`internal/mail/beadmail/beadmail.go:692`) treats *any*
closed message bead whose `close_reason` is not the retention-sweep marker as **user-removed**,
so after an archive every read path — `Get`, `Read`, `MarkRead`, `MarkUnread`, and `gc mail
peek`, which routes to `Get` — answers `ErrNotFound`. The bead and its body are untouched in
the store the whole time.

Both functions document the opposite. `Archive`: "retaining its body for later retrieval via
`gc mail peek` or `bd show`". `ArchiveMatching`: "closed rather than deleted, so their bodies
stay readable". Neither is true today.

The consequence is worst for mail that was never read. On one long-running deployment, **63 of
a single agent's 209 archive events in one day were for messages with no corresponding read
event** — and over a month, **599 message beads were closed having never been read**. Each was
an obligation that became invisible to every mail verb while its text sat intact in the store.
One of them was a cross-agent handoff carrying a go-ahead; the recipient idled roughly ten hours
because the message could not be surfaced through any supported read path. Recovery in every
case meant reading the raw store object and bypassing the mail API.

What makes this expensive to diagnose is that the failure is *silent and ambiguous by
construction*: one closed state is being asked to mean two different things — "a human deleted
this" and "this was archived, body retained" — and the read paths cannot tell them apart, so
they conservatively report the message gone.

Notably, the codebase already solved this exact problem elsewhere and just did not carry it to
these two functions:

- `SweepReadMessagesBefore` stamps `close_reason` before closing (`beadmail.go:913`).
- `ArchiveInjectedAutoHandoffs` stamps before closing (`beadmail.go:491`), with a comment
  spelling out that it does so "so an unconsumed handoff stays recoverable" and to avoid
  "permanent data loss".

So this PR is closing an internal inconsistency rather than introducing a new convention.

### The fix

1. **A discriminator.** New constant `ArchiveCloseReason = "mail archived: retained for
   peek/bd show"`. `Archive` and `ArchiveMatching` stamp it *before* `Close`, so any close-side
   validation and every subsequent reader observe it. This is the only thing that distinguishes
   "archived, body retained" from "removed" and from a legacy bead closed by an older release.

2. **A verb that can read through.** New optional interface `mail.ArchivePeeker` with
   `Peek(id) (Message, error)`; `beadmail.Provider` implements it. `Peek` returns a message
   closed with `ArchiveCloseReason` and *only* that — a `Delete`-purged bead, or a legacy bead
   closed with no reason at all, stays `ErrNotFound` even for `Peek`, so the upgrade path is
   unchanged.

3. **`gc mail peek` uses it** when the provider supports it, falling back to `Get` otherwise.
   When the command is served by the supervisor API (which serves the `Get` view and therefore
   hides archived mail), a `not_found` response is treated as a cue to fall back to the local
   read-through rather than a final answer. The existing remote-client gate still hard-errors
   for a genuinely remote operator, who has no local store to read through — that is correct
   and unchanged.

**Design constraint honoured deliberately:** the conformance suite pins `Archive` as
*removes-from-every-view*, and that stays exactly as it is. `Get`, `Read`, `Reply`, `MarkRead`,
and every inbox listing continue to return `ErrNotFound` / omit archived mail. The two
contracts — "removed from every view" and "body retrievable via peek" — are only consistent if
`peek` is a distinct read-through verb, which is what this PR makes it. No existing view
changes behaviour.

### Behaviour, before and after

Send a message, never read it, archive it:

| call | before | after |
|---|---|---|
| raw bead `status` / `close_reason` | `closed` / `""` | `closed` / `mail archived: retained for peek/bd show` |
| raw bead body | intact | intact |
| `Get` | `message not found` | `message not found` (unchanged) |
| `Read` | `message not found` | `message not found` (unchanged) |
| `MarkRead` / `MarkUnread` | `message not found` | `message not found` (unchanged) |
| inbox / list views | omitted | omitted (unchanged) |
| **`Peek` / `gc mail peek`** | **`message not found`** | **returns the message, body intact** |

And for a bead closed by an older release with no `close_reason` (the upgrade path), `Peek`
still returns `message not found` — archived-ness is proven by the stamp, never assumed from
"closed".

### Test evidence

New:

- `TestPeekHidesLegacyAndRemovedMail` — the boundary: reading through the archive must not
  resurrect a legacy closed-without-reason bead.
- `TestArchiveMatchingStampsCloseReason` — the bulk selection path gets the same contract:
  stamped, hidden from `Get`, recoverable via `Peek`.
- `TestArchive` extended — asserts the stamp, asserts `Get` still hides the message
  (conformance contract intact), asserts `Peek` returns the body.
- `TestRouteMailPeek_SixRowMatrix` updated for one row: `peek` diverging from the shared matrix
  on an API 404 is the intended fallback, and the row now pins that.

Unchanged and green: the full mail conformance suite, the legacy-closed upgrade-path test, the
retention-sweep tests, `TestArchiveMatchingSkipsPerMessageGet`, and the auto-handoff archive
tests.

```
$ go test ./internal/mail/...
ok  	.../internal/mail	0.004s
ok  	.../internal/mail/beadmail	0.063s
ok  	.../internal/mail/exec	48.147s

$ go test ./cmd/gc/ -run 'TestRouteMailPeek|TestMailArchive'
ok  	.../cmd/gc	0.308s
```

A before/after probe on the same store, run once at the base commit and once on this branch,
is the clearest summary:

```
base:  raw close_reason=""                                    Peek -> message not found
this:  raw close_reason="mail archived: retained for peek…"   Peek -> body "the handoff body"
```

### Scope

5 files, +168 / -2, all under `internal/mail` and `cmd/gc`. 41 lines of net production code.
No signature changes; `ArchivePeeker` is optional, so no provider outside `beadmail` needs to
change.

### DIFF
--- a/cmd/gc/cmd_mail.go
+++ b/cmd/gc/cmd_mail.go
@@ -2128,6 +2128,16 @@ func routeMailPeek(_ string, args []string, c *api.Client, nilReason string, jso
 		func() error {
 			var err error
 			cr, err = c.GetMail(id, "")
+			// The API mail read hides archived messages (it serves the
+			// Get view). Peek's contract is to read through the archive,
+			// which only the local provider path can do — so a not-found
+			// from the API is a cue to fall back, not a final answer.
+			// routeRead's remote-city gate (G1) still hard-errors this
+			// sentinel for remote clients, which is correct: a remote
+			// operator has no local store to read through.
+			if err != nil && strings.Contains(err.Error(), "not_found") {
+				return fallbackAfterFetch{Reason: "peek-archived-readthrough"}
+			}
 			return err
 		},
 		func() int {
@@ -2172,7 +2182,17 @@ func doMailPeekWithJSON(mp mail.Provider, args []string, jsonOut bool, stdout, s
 	}
 	id := args[0]
 
-	m, err := mp.Get(id)
+	// Peek reads through the archive when the provider supports it —
+	// Archive's contract retains the body for exactly this verb, while
+	// Get keeps archived mail hidden per the removes-from-every-view
+	// conformance contract.
+	var m mail.Message
+	var err error
+	if peeker, ok := mp.(mail.ArchivePeeker); ok {
+		m, err = peeker.Peek(id)
+	} else {
+		m, err = mp.Get(id)
+	}
 	if err != nil {
 		fmt.Fprintf(stderr, "gc mail peek: %v\n", err) //nolint:errcheck // best-effort stderr
 		return 1
--- a/cmd/gc/cmd_mail.go
+++ b/cmd/gc/cmd_mail.go
@@ -1561,7 +1561,12 @@ func newMailPeekCmd(stdout, stderr io.Writer) *cobra.Command {
 		Long: `Display a message without marking it as read.
 
 Same output as "gc mail read" but does not change the message's read status.
-The message will continue to appear in inbox results.`,
+An unarchived message will continue to appear in inbox results.
+
+peek also reads THROUGH the archive: a message closed by "gc mail archive"
+or "gc mail delete" is still displayable here, even though read, inbox, and
+check report it as not found. Mail removed before this behavior existed
+(closed with no archive reason) stays not-found for peek as well.`,
 		Args: cobra.ArbitraryArgs,
 		RunE: func(_ *cobra.Command, args []string) error {
 			if cmdMailPeekWithJSON(args, jsonOut, stdout, stderr) != 0 {
--- a/cmd/gc/cmd_mail_test.go
+++ b/cmd/gc/cmd_mail_test.go
@@ -4551,10 +4551,17 @@ func TestRouteMailPeek_SixRowMatrix(t *testing.T) {
 			wantReason: "conn-refused",
 		},
 		{
+			// Peek diverges from the shared matrix here BY DESIGN: the API
+			// serves the Get view, which hides archived mail, so a 404 is a
+			// cue to read through the archive locally (ArchivePeeker), not a
+			// final answer. The local fake has no such message either, so
+			// the exit is still 1 — but via the fallback route.
 			name:       "api-404-error",
 			handler:    mailProblemHandler(http.StatusNotFound, "not_found: no such message"),
 			wantExit:   1,
-			wantStderr: "not_found",
+			wantRoute:  "fallback",
+			wantReason: "peek-archived-readthrough",
+			wantStderr: "message not found",
 		},
 		{
 			name:         "controller-down",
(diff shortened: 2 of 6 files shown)


quad341 (approved):
@csauer02-personal-user — this is a great catch, and the writeup is what made it easy to trust. Two functions documenting "the body stays readable" while every read path answered not-found is the kind of bug that hides for months, and you found *why* it hides: one closed state being asked to mean two different things. Noticing that the codebase had already solved this twice — `SweepReadMessagesBefore` and `ArchiveInjectedAutoHandoffs` — and carrying that pattern over instead of inventing a third mechanism is exactly right.

`TestPeekHidesLegacyAndRemovedMail` is my favorite part. Making the read-through verb prove it *can't* resurrect a legacy closed-no-reason bead is the assertion I'd have asked for, and you wrote it first. Keeping `Peek` off `mail.Provider` so the removes-from-every-view conformance contract stays untouched was the right call too — I checked, and it holds.

I'll take the small integration bits on my side: the `gc mail peek` help text still says it only skips the read-marking, so I'll update it and regenerate `docs/reference/cli.md` for the drift check, plus a short test pinning that `delete` (an alias for `archive`, per #4422) is peek-recoverable too. Merging with your commits and credit intact.


---

quad341:
@csauer02-personal-user — thank you for this one. Two functions documenting "the body stays readable" while every read path answered not-found is the class of bug that survives for months, and you didn't just patch it: you named *why* it hides, that one closed state was being asked to mean two different things.

`TestPeekHidesLegacyAndRemovedMail` is the assertion I'd have asked for, and you wrote it first. Keeping `Peek` off `mail.Provider` so the removes-from-every-view conformance contract stays untouched was the right call — I traced it, it holds.

One thing your writeup undersells: `BdStore.Close` reads the stamped reason back out to build `bd close --reason`, and bd's default `"Closed"` is under the 20-char floor `validation.on-close=error` enforces. On a validating city this was likely making archive *fail*, not just lose the body. Your 40-character constant fixes that too.

Our test suites never dispatched on this head — that's our CI gap, not yours, and the red check is the watchdog reporting the absence rather than a failure. I'll add two tests on my side (one driving `gc mail peek` against an actually-archived message, one pinning the remote-client 404 text) and re-run. Your commits and credit stay intact.


---

csauer02-personal-user:
Closing by the author: we do not run this fix on a served build of ours, so I am not the right person to carry it. quad341, your 09-06 approval and your 09-08 finding (bd's default "Closed" is under the 20-char floor validation.on-close=error enforces, so archive likely fails on a validating city) strengthen the case; the diagnosis stands (beadmail.go closes without SetMetadata, isRemovedMessageBead unchanged, no Peek) and the branch is yours to adopt with authorship intact. This note was agent-drafted.

PR #5515, opened 2026-08-22
test(beads): use package hang budgets for pure hang detectors

quad341:
## What this changes

The `internal/beads` tests now use package-level budgets for waits whose only job is to detect a hung goroutine or child process. This keeps heavily loaded CI hosts from treating the repository-wide race floors as tuned deadlines.

Passing tests do not become slower: every wait still returns as soon as the expected checkpoint is reached. Only a genuinely wedged test waits out the larger package budget.

## Review notes

- The change is confined to four `_test.go` files; production behavior is unchanged.
- Goroutine/lock waits and re-exec child-protocol waits use separate constants, each set to six times its documented global floor.
- Latency assertions and negative “nothing happened” checks are intentionally out of scope for these budgets.

## Test plan

- [x] Run the five affected top-level tests and all 13 sequence-floor boundary subtests.
- [x] Run `make test-local-full-parallel` and compare all raw failures with current `main`.
- [x] Run `make test-ci-policy`, `go vet ./...`, changed-file formatting, and fresh-cache affected lint.
- [x] Release gate: [`release-gates/ga-0rgsd2-internal-beads-hang-budget-gate.md`](release-gates/ga-0rgsd2-internal-beads-hang-budget-gate.md)


quad341:
@quad341 — the distinction you're drawing here is the valuable part of this PR. `GoroutineRaceTimeout` and `ExecRaceTimeout` are *floors*, and treating a floor as a tuned deadline is exactly how a loaded CI host ends up looking like a wedged goroutine. The comments in `hangbudget_test.go` do a genuinely good job of explaining why raising a hang budget costs a passing run nothing, and putting "DO NOT tune this to make a failing test pass" right at the declaration site is the kind of thing that survives the next person who hits a red test at 2am.

One thing I'll pick up on my side rather than send back to you: `beadsSequenceFloorHangBudget` is declared in the untagged `hangbudget_test.go` but only used from the `//go:build linux` sequence-floor file, so under GOOS=darwin it's dead and `unused` flags it — I reproduced it locally. CI is green only because the macOS quality lane is skipped for PRs without the `needs-mac` label; the nightly run would have caught it. I'll move that const into a `hangbudget_linux_test.go`, matching how `sqlite_fence_process_linux_test.go` keeps its own budget in the tagged file, re-lint under both GOOSes, and merge with your commit intact.

Contained, well-reasoned change — thanks for it.


---

PR #5516, opened 2026-08-22
fix(core): claim WORK_BEAD_ID before heartbeating it in mol-do-work

aytheo-bit:
## Bug

`gc sling <pool-target> <existing-bead-id>` wraps the routed bead as the sole child of a synthetic input convoy. The `do-work` step's own routing claims the workflow STEP bead, but never claims that convoy child (`WORK_BEAD_ID`) itself. The step then instructs the agent to heartbeat `WORK_BEAD_ID` directly, which fails 100% of the time — `bd heartbeat` only refreshes an existing `in_progress` lease held by the calling actor, it does not auto-claim.

Confirmed 100% reproduction across three independent dispatches on a downstream city before this fix (all failing identically at the first `gc bd heartbeat` call with `"issue not claimable: <id> status open"`).

## Fix

Add an explicit `gc bd update "$WORK_BEAD_ID" --claim` immediately before the heartbeat in `mol-do-work.toml`'s `do-work` step. `--claim` is idempotent and self-assigns regardless of current state, so this is safe whether or not something else already holds/matches the lease.

## Verification

Root-caused via a controlled probe reproducing the exact error on a disposable bead, and via `bd history` showing affected beads only ever transitioned `open -> closed` (auto-closed as a blocked outcome), never `in_progress` — confirming the claim step was genuinely missing, not a flaky race.

### DIFF
--- a/internal/bootstrap/packs/core/formulas/mol-do-work.toml
+++ b/internal/bootstrap/packs/core/formulas/mol-do-work.toml
@@ -44,11 +44,19 @@ gc bd show "$WORK_BEAD_ID"
 
 Read the bead's title and description carefully. This is your task.
 
-**Signal liveness:** run `gc bd heartbeat "$WORK_BEAD_ID"` now, and again before any
-shell command you expect to run for many minutes (long builds, test sweeps).
-It refreshes your claim's lease so the dashboard can tell you are alive and
-working rather than wedged. If a heartbeat ever fails, your claim was reclaimed
-or closed out from under you — stop working this bead immediately.
+**Claim it and signal liveness:** the input-convoy child derived above is not
+claimed by the routing step that assigned you — claim it explicitly before your
+first heartbeat:
+```bash
+gc bd update "$WORK_BEAD_ID" --claim
+```
+`--claim` is idempotent and self-assigns regardless of current state, so this is
+safe even if the bead was already claimed some other way. Then run
+`gc bd heartbeat "$WORK_BEAD_ID"` now, and again before any shell command you
+expect to run for many minutes (long builds, test sweeps). It refreshes your
+claim's lease so the dashboard can tell you are alive and working rather than
+wedged. If a heartbeat ever fails, your claim was reclaimed or closed out from
+under you — stop working this bead immediately.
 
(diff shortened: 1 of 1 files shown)


sjarmak:
## Adoption Review

Thanks for the contribution, @aytheo-bit. This closes a real gap in mol-do-work's claim/heartbeat sequencing.

### Review Summary
Approved. Two independent reviewers confirmed the root cause: `internal/sling`/`internal/dispatch` claim the workflow step bead (`GC_BEAD_ID`) but never the synthetic input-convoy's child bead (`WORK_BEAD_ID`). That bead stays `status=open` until the agent's own `bd heartbeat` call runs, and a heartbeat never auto-claims. The added `bd update "$WORK_BEAD_ID" --claim` step closes that gap. `--claim` is a real, idempotent `bd update` flag already used the same way elsewhere in this codebase's agent contracts. The change touches one formula markdown file with no step IDs or dependencies altered; nothing else in the codebase reads that description text, so no other update was needed.

### Review Findings
| # | Category | Severity | Confidence | Source | Summary |
|---|----------|----------|------------|--------|---------|
| 1 | Debuggability & Operability | minor | medium | codex | The prompt text says `--claim` is "safe even if the bead was already claimed some other way." That's true only when claimed by *you*; taking over another actor's live lease needs `--force` instead. This branch never triggers under the bug this PR fixes (the bead is always unclaimed), so it's non-blocking. Optional follow-up: soften the wording and add guidance for a failed claim.

---
_Adopted via `/adopt-pr` workflow. Contributor commit preserved._

sjarmak (approved):
Approving on the dual-model review recorded above. The claim-before-heartbeat ordering is the right fix: the input-convoy child is unclaimed by construction, so the heartbeat was the first write against a bead this session did not own.

One non-gating note from the review, tracked separately rather than held against this PR: the surrounding prose says `--claim` is safe "even if the bead was already claimed some other way", which overstates it. `--claim` is idempotent when the bead is already claimed by you; `--force` is what overrides another actor's live claim. The path is unreachable here, which is exactly what this PR establishes, so the sentence is wrong rather than dangerous. Thanks for the clean, minimal change.

PR #5517, opened 2026-08-22
test(api): consolidate hang-detection timeouts

quad341:
## What this changes

This consolidates ten pure hang-detection waits in the `internal/api` test suite behind one package-level `hangBudget`. Slow or contended hosts now get a consistent 60-second wedge ceiling, while passing tests still finish as soon as their expected stream event or cancellation arrives.

There is no production API behavior change. Scenario-input timeouts and negative assertions keep their existing, purpose-specific windows.

## Review notes

- The change is test-only: one new package constant and ten call-site substitutions across two streaming test files.
- `hangBudget` is derived from the shared goroutine race floor; it is a last-resort ceiling, not a latency target.
- The two widened context deadlines are canceled explicitly after assertions, and the handlers treat cancellation and deadline expiry through the same return path.

## Test plan

- [x] Run the documented 40-job local CI union with rootless Podman; verify every candidate-owned lane passes and attribute only current-base, path-disjoint host failures.
- [x] Run all four changed API tests by exact name with 4 PASS / 0 FAIL / 0 SKIP.
- [x] Run CI policy, changed-scope lint/format, full Go vet, dashboard build/type checks/tests, and the pre-push fast gate.
- [x] Release gate: [`release-gates/ga-ox9nm0-internal-api-hang-budget-gate.md`](release-gates/ga-ox9nm0-internal-api-hang-budget-gate.md)


quad341:
@quad341 — thanks for this one. The part worth calling out isn't the constant, it's the comment on it: spelling out that `hangBudget` is a *ceiling* while `GoroutineRaceTimeout` is a *floor*, and naming the two cases that must never migrate to it — a timeout fed into the code under test, and the window of a negative assertion — is exactly the guardrail that stops someone six months from now from "fixing" a flaky test by bumping the number.

I checked the load-bearing claim rather than taking it on faith. `handler_session_stream.go` really does route `DeadlineExceeded` and `Canceled` through the same `case <-ctx.Done()` branch, so widening those two deadlines from 20s to 60s changes nothing about what's exercised. Both tests also call `cancel()` before `<-done` on every path, including the `t.Fatalf` branches — the deadline is a true backstop, not the thing ending a passing run.

Merging as-is. `internal/api` runs in the `packages-core` shards, all four green on this head, so the converted tests genuinely executed. The release-gate checklist with the three pre-existing failures attributed to current base was a nice touch too — that saved us the re-derivation. Careful piece of the migration.


---

quad341:
Re-reviewed `8c9ea3d9c`..`ba5e65b07` (1 new commit since the last review): no new concerns — the prior review stands.

PR #5520, opened 2026-08-22
fix(api): retry provider bearer once after 401

julianknutsen:
## Summary

- Force-refresh provider-backed bearer credentials and retry one eligible request after a 401.
- Keep static tokens, SSE, redirects, non-replayable requests, 403 responses, and single-use grants out of the retry path.
- Propagate request cancellation to the credential command and close replaced 401 response bodies.

## Testing

- [ ] `make check`
- [x] Focused API/clientauth/CLI tests
- [x] Commit lint, generated-code, and vet hooks
- [x] Required pre-push fast suite

## Checklist

- [x] Linked an issue, or explained why one is not needed: implements bead `ga-0xdql0.5`.
- [x] Updated docs for user-facing changes: not needed; this is transport behavior.
- [x] Called out breaking changes or migration notes: none.

### DIFF
--- a/cmd/gc/cmd_registry.go
+++ b/cmd/gc/cmd_registry.go
@@ -733,6 +733,12 @@ func (rt *registryProviderReauthRoundTripper) RoundTrip(req *http.Request) (*htt
 		if ctxErr := req.Context().Err(); ctxErr != nil {
 			return nil, fmt.Errorf("refreshing registry credential after 401: %w", ctxErr)
 		}
+		if errors.Is(err, context.Canceled) {
+			return nil, fmt.Errorf("refreshing registry credential after 401: %w", context.Canceled)
+		}
+		if errors.Is(err, context.DeadlineExceeded) {
+			return nil, fmt.Errorf("refreshing registry credential after 401: %w", context.DeadlineExceeded)
+		}
 		return nil, errors.New("refreshing registry credential after 401: credential refresh failed")
 	}
 	token = strings.TrimSpace(token)
--- a/cmd/gc/cmd_registry.go
+++ b/cmd/gc/cmd_registry.go
@@ -722,7 +722,8 @@ func (rt *registryProviderReauthRoundTripper) RoundTrip(req *http.Request) (*htt
 	if err != nil || resp == nil || resp.StatusCode != http.StatusUnauthorized || rt.refresh == nil {
 		return resp, err
 	}
-	if !registryHasBearerAuthorization(req.Header.Get("Authorization")) ||
+	if !isRegistryRetrySafeMethod(req.Method) ||
+		!registryHasBearerAuthorization(req.Header.Get("Authorization")) ||
 		(req.Body != nil && req.Body != http.NoBody && req.GetBody == nil) {
 		return resp, nil
 	}
@@ -752,7 +753,7 @@ func (rt *registryProviderReauthRoundTripper) RoundTrip(req *http.Request) (*htt
 		retry.Body, err = req.GetBody()
 		if err != nil {
 			_ = resp.Body.Close()
-			return nil, fmt.Errorf("replaying registry request after credential refresh: %w", err)
+			return nil, errors.New("replaying registry request after credential refresh: request body recreation failed")
 		}
 	}
 	retry.Header.Set("Authorization", "Bearer "+token)
@@ -763,6 +764,15 @@ func (rt *registryProviderReauthRoundTripper) RoundTrip(req *http.Request) (*htt
 func registryHasBearerAuthorization(value string) bool {
 	fields := strings.Fields(value)
 	return len(fields) == 2 && strings.EqualFold(fields[0], "Bearer") && fields[1] != ""
+}
+
+func isRegistryRetrySafeMethod(method string) bool {
+	switch strings.ToUpper(method) {
+	case http.MethodGet, http.MethodHead, http.MethodOptions:
+		return true
+	default:
+		return false
+	}
 }
 
 func (a registryPublishAuth) hasCredentials() bool {
--- a/cmd/gc/cmd_registry.go
+++ b/cmd/gc/cmd_registry.go
@@ -730,7 +730,10 @@ func (rt *registryProviderReauthRoundTripper) RoundTrip(req *http.Request) (*htt
 	token, err := rt.refresh(req.Context(), true)
 	if err != nil {
 		_ = resp.Body.Close()
-		return nil, fmt.Errorf("refreshing registry credential after 401: %w", err)
+		if ctxErr := req.Context().Err(); ctxErr != nil {
+			return nil, fmt.Errorf("refreshing registry credential after 401: %w", ctxErr)
+		}
+		return nil, errors.New("refreshing registry credential after 401: credential refresh failed")
 	}
 	token = strings.TrimSpace(token)
 	if token == "" {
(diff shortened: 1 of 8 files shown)


julianknutsen (commented):
Independent Sol council verdict: **BLOCK** at exact head `c9f7b9dd86a3a4d99fe0767c68c1bf91418d8059`.

Required findings:

1. **Unsafe mutating requests remain replayable.** `reauthRoundTripper.RoundTrip` excludes single-use grants, SSE, redirects, and non-replayable bodies, but it never checks HTTP method safety. A replayable `POST`, `PUT`, `PATCH`, or `DELETE` without `X-GC-City-Write` therefore refreshes and is sent a second time after 401. The new `TestReauthRoundTripperReturnsBodyRecreationFailure` also demonstrates that a replayable POST enters the refresh/replay branch. The scoped contract requires no retry for unsafe requests; gate retry eligibility to safe methods and add a negative replay-count test.

2. **Generated SSE requests omit the required Accept header.** `NewRemoteEventsClient` sets CSRF and optional Authorization only. The new five-route test explicitly requires `Accept` to be empty, while these requests are SSE and must send `Accept: text/event-stream`. `isSSERequest` compensates with a hard-coded route-shape list, which is fragile for future generated SSE operations. Set the SSE Accept header in the remote events client and keep the one-request/zero-refresh assertions for all five routes.

Passing evidence:
- Focused `internal/api`, `internal/clientauth`, and `cmd/gc` retry/auth tests: PASS.
- Focused `-race` lock-wait cancellation tests: PASS.
- `go vet ./internal/api ./internal/clientauth ./cmd/gc`: PASS.
- First 401 response closure, one-refresh ceiling, second-401 behavior, static-token exclusion, single-use grant exclusion, redirect exclusion, context.Canceled/context.DeadlineExceeded identity, lock-wait cancellation, and secret-safe refresh errors are covered and pass.
- No credential logging addition found.
- Local and remote PR heads match the exact commit; required CI is green.

The authenticated GitHub identity is the PR author, so GitHub cannot accept a formal REQUEST_CHANGES review from it. This COMMENT records the exact-head blocking verdict.

julianknutsen:
Independent Sol council reviewer #1 verdict: **BLOCK** at exact head `35b1fc41c7b885c134bf74daee690f1d8fca3926`.

Required finding:

- **AC-3's body-recreation error is not secret-safe.** `internal/api/client_remote.go:349` wraps the arbitrary error returned by `req.GetBody()` with `%w`, and `TestReauthRoundTripperReturnsBodyRecreationFailure` explicitly requires that underlying text to be exposed. A body factory may include request/body diagnostics in its error, so this can surface sensitive request material through the CLI or logs. Return a fixed/sanitized replay error instead and add a regression whose `GetBody` error contains a secret marker, while retaining the assertions that the rejected 401 is closed and no replay occurs.

Passing evidence:

- The latest safe-method gate correctly prevents replay of replayable POST/PUT/PATCH/DELETE requests.
- All five generated SSE calls now send `Accept: text/event-stream` and make one request with zero refreshes after 401.
- Focused API, CLI, and client-auth tests pass; the focused client-auth cancellation test passes under `-race`.
- Required GitHub CI is green, the worktree is clean, and local/remote heads both match this exact SHA.
- No other required findings. Optional findings: none.

julianknutsen:
Independent Sol council review verdict: **BLOCK** at exact head `f0b08d3103664af9c5309770005e92463a6820fa`.

Required findings:

1. `registryProviderReauthRoundTripper` in `cmd/gc/cmd_registry.go` still retries a replayable credentialed `POST` after 401 because it has no safe-method gate. Registry publish is a POST, so an authorization response received after a side effect can result in duplicate mutation. This conflicts with the scoped one-retry contract, which permits replay only for safe methods.
2. The same transport wraps arbitrary `req.GetBody()` errors with `%w`, allowing request/body diagnostics to escape. Return a fixed secret-safe error and add a secret-marker regression.

Passing evidence: the remote-city transport correctly gates GET/HEAD/OPTIONS and excludes grants, SSE, redirects, and non-replayable bodies; its body-recreation error is sanitized. Focused api/clientauth/cmd-gc tests, race checks, vet, and exact-head required CI are green. No other required finding.

julianknutsen:
BLOCK at exact head f0b08d3103664af9c5309770005e92463a6820fa.

Required findings:
- cmd/gc/cmd_registry.go:725-760 retries replayable unsafe methods, including the side-effecting POST /api/publish-requests. It must gate retry eligibility to GET/HEAD/OPTIONS and add replay-count coverage for POST/PUT/PATCH/DELETE.
- cmd/gc/cmd_registry.go:750-756 wraps req.GetBody errors with %w, exposing arbitrary body/helper diagnostics. Return a fixed sanitized replay error and add a secret-marker regression.

The internal/api reauth transport has these protections; the registry transport still does not.

julianknutsen:
APPROVE — independent Sol review at exact head `0783eb91168b19f9191344befa78d79e8f56991e`.

The remote-city transport now permits exactly one 401 refresh/replay only for bearer-authenticated GET/HEAD/OPTIONS requests that are replayable, non-SSE, non-grant, and not redirect follow-ups; the retry calls the base transport directly so a second 401 cannot refresh again. The registry provider wrapper has the same safe-method gate, preventing publish POST replay. Refresh and `GetBody` failures return fixed secret-safe errors, preserve cancellation identity, and close the rejected response body. Focused `internal/api`, `internal/clientauth`, and `cmd/gc` retry/registry tests passed; clientauth race test, vet, and diff-check passed. Fresh required GitHub checks (`Check`, `CI / preflight`, `CI / integration`, `CI / required`) are green for this SHA. A broad local `cmd/gc` run hit unrelated environment/process-leak timeouts after 10 minutes.

julianknutsen:
Sol council review — **APPROVE** at exact head `0783eb91168b19f9191344befa78d79e8f56991e`.

Reviewed both transports against the one-refresh contract:

- remote-city `reauthRoundTripper`: retries only `GET`/`HEAD`/`OPTIONS`, only a Bearer-authenticated request, only replayable non-SSE requests, and never grants or redirect follow-ups; unsafe methods, SSE, static/no-Bearer credentials, and non-replayable bodies are not retried.
- registry provider transport: the same safe-method/Bearer/replayability gate; registry publish `POST` is not retried. Redirects are rejected before a credentialed follow-up.
- both transports close the rejected 401 before refresh/replay and stop after the single retry; a second 401 is returned without another refresh.
- refresh runs with request context cancellation/deadline; non-context helper failures and `GetBody` failures return fixed secret-safe errors. Tests assert secret markers never escape.
- `NewRemoteEventsClient` sets `Accept: text/event-stream`; generated SSE calls make one request and zero refreshes after 401.
- explicit/static tokens disable provider refresh; no credential logging was added.

Focused evidence at this SHA: `go test -race ./internal/clientauth`, `go test -race ./internal/api -run 'Test(ReauthRoundTripper|NewRemoteEventsClient)'`, `go test -race ./cmd/gc -run 'Test(RegistryProviderReauthRoundTripper|DoRegistryWhoamiUsesGasworksProvider|DoRegistryPublishProviderDoesNotRefresh)'`, and `go vet ./internal/api ./internal/clientauth ./cmd/gc` all pass. The broader focused `go test ./internal/api ./internal/clientauth ./cmd/gc` also passed for API/clientauth before the race rerun.

julianknutsen:
Evidence correction for the exact-head approval above: the focused race suites and vet all passed as stated, and the full `internal/api` and `internal/clientauth` package tests passed. The unfiltered `go test ./cmd/gc` run did **not** pass; it hit the package-wide 10-minute test timeout amid unrelated parallel CLI tests. The reviewed focused `cmd/gc` registry retry suite passed under `-race`. This does not change the approval, but the broader package result should not be read as green.

PR #5521, opened 2026-08-23
fix(worker): backfill session_id_flag and pin the real claude fork contract

julianknutsen:
## Post-merge remediation for #5396

Follow-up to landed PR #5396 (*fix(worker): resume claude sessions across restart in the live conformance suite*). The post-merge review of the landed range `8c73625b..812e50e1` surfaced three actionable findings; this PR applies the required fixes only.

### Findings fixed

**1. Backfill `session_id_flag` alongside `session_key` (correctness).**
`syncSessionBeads` backfills `session_key` for pre-fix keyless claude beads but did not persist `session_id_flag`. The reactive startup-death strip (`stripSessionIDFlag` via `retryFreshStartAfterStaleKey`) reads that flag from persisted bead metadata, so an upgraded bead carrying a backfilled `session_key` but no `session_id_flag` would replay a rejected first-start `--session-id <key>` command and crash-loop until quarantine. The flag now projects from the resolved provider exactly like `resume_flag` — at create and, diff-gated, on the observe/backfill path — so upgraded beads gain the flag in the same tick the key is backfilled, with no churn on an unchanged sync.

**2. Pin the real `BuiltinProviders()["claude"]` fork contract (test evidence).**
The fork-launch suite asserted only against a synthetic `forkClaude()` twin, so the builtin↔fork contract could drift from the real provider spec undetected. #5396 added `SessionIDFlag` to the claude builtin, which arms `validateForkLaunch` (previously fail-closed on an empty `SessionIDFlag`). A new test drives the *real* resolved claude provider (worker builtin → `config.BuiltinProviders` → `ResolveProvider`/`specToResolved`) through `validateForkLaunch` + `resolveSessionCommand`, pinning the fork and fresh-mint CLI strings against the actual spec.

**3. Value-agnostic `--session-id` strip on the diverged-key fresh retry (correctness).**
`retryFreshStartAfterStaleKey` stripped the first-start `--session-id <key>` pair only
by exact key match. When the key embedded in the command at build time has diverged from
the bead's current `session_key` (a concurrent fresh start minted a new key, or a stale
store read), that keyed strip is a no-op — and because a first start carries no resume
flag, the `stripResumeFlagArg` fallback cannot reach it either, so the retry replayed the
dead `--session-id <oldkey>` into the same "id already in use" rejection it exists to
escape. `stripSessionIDFlagArg` now strips the flag/value pair regardless of value,
mirroring `stripResumeFlagArg`, gated on the keyed strip having been a no-op so the
existing resume-divergence signal (`freshCmd == resumeCommand`) is preserved.

### Changes
- `cmd/gc/session_beads.go` — project `session_id_flag` at create + diff-gated backfill.
- `cmd/gc/session_beads_test.go` — `TestSyncSessionBeads_ProjectsSessionIDFlag` (create / backfill / churn-guard).
- `cmd/gc/session_reconciler_fork_launch_test.go` — `TestForkLaunch_RealBuiltinClaudeSpecPinsContract`.
- `internal/session/chat.go` — `stripSessionIDFlagArg` + the diverged-key gate in `retryFreshStartAfterStaleKey`.
- `internal/session/chat_test.go` — `TestStripSessionIDFlagArg`.
- `internal/session/manager_test.go` — `TestEnsureRunning_RetriesWhenSessionIDKeyDiverged`.

### Tests
- Focused: `TestSyncSessionBeads_ProjectsSessionIDFlag`, `TestForkLaunch_RealBuiltinClaudeSpecPinsContract`, plus all sibling fork/churn/build-prepared-start tests — pass.
- `go vet ./cmd/gc/` — clean.
- Full local fast suite (pre-push gate: `unit-core` + all `cmd/gc` shards) — pass.

### Self-review
Reviewed the final diff with fresh eyes: both edits are additive, gated on a non-empty `SessionIDFlag` (non-keyed providers unaffected), diff-gated to avoid sync churn, and mirror the adjacent `resume_flag`/`resume_command` handling. Scope is limited to the two required findings plus the value-agnostic `--session-id` strip they depend on; the two non-gating follow-ups (first-start strip regression test, fresh-fallback warning throttle) are intentionally deferred.

### DIFF
--- a/cmd/gc/session_beads.go
+++ b/cmd/gc/session_beads.go
@@ -1947,6 +1947,9 @@ func syncSessionBeadsWithSnapshotAndRigStores(
 				if tp.ResolvedProvider.ResumeCommand != "" {
 					meta["resume_command"] = tp.ResolvedProvider.ResumeCommand
 				}
+				if tp.ResolvedProvider.SessionIDFlag != "" {
+					meta["session_id_flag"] = tp.ResolvedProvider.SessionIDFlag
+				}
 			}
 			createBead := func() (beads.Bead, error) {
 				beadID, err := sessionFrontDoor(store).CreateSession(session.CreateSpec{
@@ -2217,6 +2220,16 @@ func syncSessionBeadsWithSnapshotAndRigStores(
 			}
 			if tp.ResolvedProvider.ResumeCommand != "" && b.Metadata["resume_command"] != tp.ResolvedProvider.ResumeCommand {
 				queueMeta("resume_command", tp.ResolvedProvider.ResumeCommand)
+			}
+			// session_id_flag is a peer projection field of resume_flag: the
+			// reactive startup-death strip (stripSessionIDFlag via
+			// retryFreshStartAfterStaleKey) reads it from persisted bead
+			// metadata, so a legacy bead backfilled with session_key (above)
+			// but no session_id_flag would replay a rejected first-start
+			// `--session-id <key>` command. Project it diff-gated so upgraded
+			// beads gain the flag in the same tick the key is backfilled.
+			if tp.ResolvedProvider.SessionIDFlag != "" && b.Metadata["session_id_flag"] != tp.ResolvedProvider.SessionIDFlag {
+				queueMeta("session_id_flag", tp.ResolvedProvider.SessionIDFlag)
 			}
 		}
 
--- a/cmd/gc/session_beads_test.go
+++ b/cmd/gc/session_beads_test.go
@@ -4929,6 +4929,83 @@ func TestSyncSessionBeads_RefreshesResolvedProviderMetadataOnProviderSwitch(t *t
 	}
 }
 
+// TestSyncSessionBeads_ProjectsSessionIDFlag closes Finding 2 of the PR #5396
+// post-merge review. syncSessionBeads backfills session_key for pre-fix keyless
+// claude beads but historically did not persist session_id_flag, while the
+// reactive startup-death strip (stripSessionIDFlag via
+// retryFreshStartAfterStaleKey) reads the flag from persisted bead metadata. An
+// upgraded bead carrying a backfilled session_key but no session_id_flag would
+// replay a rejected first-start `--session-id <key>` command and crash-loop until
+// quarantine. The flag must project from the resolved provider exactly like
+// resume_flag — both at create and, diff-gated, on the observe/backfill path —
+// without churning an unchanged sync (#1205).
+func TestSyncSessionBeads_ProjectsSessionIDFlag(t *testing.T) {
+	store := beads.NewMemStore()
+	// countingStore lets the churn tick assert that an unchanged sync writes
+	// nothing (diff-gated), the load-bearing property for #1205.
+	cs := &countingStore{Store: store}
+	clk := &clock.Fake{Time: time.Date(2026, 8, 23, 12, 0, 0, 0, time.UTC)}
+	sp := runtime.NewFake()
+	_ = sp.Start(context.TODO(), "worker", runtime.Config{Command: "claude --dangerously-skip-permissions"})
+
+	claudeCmd := "claude --dangerously-skip-permissions"
+	ds := map[string]TemplateParams{
+		"worker": {
+			TemplateName: "worker",
+			Command:      claudeCmd,
+			ResolvedProvider: &config.ResolvedProvider{
+				Name: "claude", BuiltinAncestor: "claude",
+				ResumeFlag: "--resume", ResumeStyle: "flag", ResumeCommand: "claude --resume",
+				SessionIDFlag: "--session-id",
+			},
+		},
+	}
+	var stderr bytes.Buffer
+
+	// Tick 1 (create path): a keyed provider stamps session_id_flag at creation,
+	// alongside the generated session_key.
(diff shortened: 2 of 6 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277227794

julianknutsen:
Preflight blocked: the PR description declares two required findings and three changed files, but the live head contains a second commit adding an additional value-agnostic session-id retry fix and changes six files. Please update the PR story and scope (or split the additional fix) before review can continue.

quad341 (dismissed):
@julianknutsen — this is a genuinely good post-merge remediation, and `TestForkLaunch_RealBuiltinClaudeSpecPinsContract` is the standout. Driving the actual `BuiltinProviders()["claude"]` through `ResolveProvider` → `validateForkLaunch` → `resolveSessionCommand` and pinning the literal CLI string is exactly the test #5396 was missing, and it catches drift a synthetic twin never would. The `session_id_flag` projection is careful in the same way: clobber-safe on empty, diff-gated, and you proved the churn property with a counting store rather than just asserting values.

The only thing in the way is bookkeeping on our side. Preflight stopped on the description declaring two findings across three files while head carries a third commit — the value-agnostic `--session-id` strip — across six. I'll rewrite the PR body to declare that third fix and correct the file list, then merge. No code change, and I'm not splitting it out: it's the same remediation family and it ships with both a unit table and a real end-to-end retry test.

Two things I'll file separately rather than pile onto you, both adjacent rather than yours: on a *fork* first start the retry strips `--session-id` but keeps `--resume <parentSID> --fork-session`, so it re-forks instead of going fresh (inherited from #5396), and the backfill quietly changes which arm `freshRestartSessionKey` takes for orphaned beads — almost certainly for the better, just untested. Thanks for the thorough self-review; it made this fast to verify.


---

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR fixes a real
session-reconciliation defect — an upgraded legacy session bead had `session_key`
backfilled but never `session_id_flag`, so both `--session-id` strip guards
short-circuited and a dead `--session-id <key>` replayed on the diverged-key
fork/resume retry and stranded the session in quarantine — which matters because
it removes a concrete way for fork/resume sessions to fail reconciliation, and it
closes a test blind spot that had let the builtin `claude` provider spec drift
past a synthetic fork twin.

### Original PR Review
Decision: approve (resolved from an initial request_changes once the diverged-key
strip was reworked).

Specific gaps fixed:
- The diverged-key `--session-id` strip originally re-tokenized the whole command
  with `strings.Fields` + `strings.Join`, which could collapse whitespace or
  rewrite unrelated shell text on the fresh-retry rewrite. The final patch
  replaces it with an in-place span excision that preserves command structure and
  sibling flags such as `--session-id-file`, and it only fires when the embedded
  first-start key actually diverged from the bead's current `session_key`
  (verified not to double-apply with the resume fallback).
- `syncSessionBeads` backfilled `session_key` but never `session_id_flag`, so an
  upgraded legacy bead carried an empty flag that short-circuited both strip
  guards. The final patch adds the `session_id_flag` projection byte-parallel to
  the shipped `resume_command` projection (non-empty guard on create; non-empty
  and diff-gated on observe), validated green by
  `TestSyncSessionBeads_ProjectsSessionIDFlag`.
- The fork-launch suite only asserted against the synthetic `forkClaude()` twin.
  The final patch adds `TestForkLaunch_RealBuiltinClaudeSpecPinsContract`, which
  drives the real `config.BuiltinProviders()["claude"]` spec through
  `ResolveProvider -> validateForkLaunch -> resolveSessionCommand`, closing the
  builtin-vs-spec drift gap that let a mock hide the `SessionIDFlag` activation.

Review findings addressed:
- Behavioral Correctness (major, raised by both reviewers): the quote/whitespace-
  naive `strings.Fields` + `strings.Join` re-tokenization in the diverged-key
  strip was fixed by commit "fix(session): strip diverged --session-id in place,
  preserve command structure" (top of the final stack, `32077917`), which excises
  the flag span in place and preserves the surrounding command.

- `mc-qihb`: Test Evidence Quality (nit, maintainability) — the `--session-id`
  strip helpers are shell-token-simple, so a bare `--session-id` token sitting
  inside a quoted argument would be matched. This is unreachable today (no
  framework-generated command folds caller-supplied or prompt text into the
  persisted launch command the strip sees) and is pinned as a commented
  accepted-limitation test at `internal/session/chat_test.go:333-342`. Not
  addressed here; if command construction ever begins folding caller-supplied
  quoted text into the persisted launch command, the whole strip family — not
  just this helper — must become shell-quote-aware. Filed as `mc-qihb`.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `32077917` has passing required GitHub
checks (CI, CodeQL, Go Vulnerability Scan, Mac Regression, PR evidence watchdog,
Review Formulas), and the merge-ready metadata is in place. The branch was rebased
onto the latest `main` to keep it current; the contributor commits and the
+380/−1 diff are preserved. I am marking this PR `status/merge-ready` so the merge
queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/32906140484

### Review Iterations
2 review passes: the first requested changes on the diverged-key strip's
whitespace/quote handling; the second approved after the in-place-excision
rewrite, leaving only the documented non-gating nit above.

---

PR #5531, opened 2026-08-23
fix(worker,session): correct zcode mirror/turn handling and deferred-submit epoch fence

julianknutsen:
## Summary

Follow-up remediation for the post-merge review of #5397 ("feat(worker): add
zcode as a conformance-certified harness family"). The post-merge review of the
landed range surfaced four independent major/correctness findings in the zcode
harness integration and the deferred-submit epoch fence. This PR applies the
required fixes with regression tests (TDD) and changes nothing else.

Reviewed landed range of #5397: `812e50e1..af85466`.

## Findings addressed

1. **De-dup guard returns `None` whenever the transcript tail is an assistant
   entry (latent mirror-history corruption).** `last_user_text()` in
   `internal/worker/adapters/zcode/zcode-repl` had its `return None` *inside*
   the reversed-scan loop, so it inspected only the final transcript entry: it
   returned the prior user text only when that entry was itself a user message,
   and `None` otherwise. The `mode_reply()` guard
   (`last_user_text(export) != prompt`) then concluded "no prior user text" and
   re-appended the prompt for any turn reaching it with an assistant-tailed
   transcript on a continuing session. In the common path this is masked,
   because the turn is published at start (`mode_prompt`) and leaves a trailing
   *user* entry the buggy scan happens to see — but the guard must be correct on
   its own, not by accident of ordering. Moved the `return None` out of the loop
   so the helper scans backward to the last user entry regardless of the tail;
   the happy-path result is unchanged, and the assistant-tailed case is now
   correct instead of duplicating.

2. **Reset-deferred submit was dead-lettered by the epoch fence (silent message
   loss).** In `internal/session/submit.go`, a submit deferred while a
   conversation reset was pending was stamped with the pre-bump epoch N, then
   rejected by the queued-nudge fence once the reset rotated the conversation to
   N+1. `enqueueDeferredSubmitLocked` now stamps an empty continuation epoch when
   `continuation_reset_pending` is set, so the item survives the rotation (it is
   still scoped by `SessionID`). The plain `restart_requested`-only path — which
   does not bump the epoch — is unchanged.

3. **First-turn `pending-*` mirror was invisible to scoped transcript
   discovery.** A first turn that failed before a session id was assigned
   persisted only as a `pending-*` placeholder, which `FindZCodeSessionFileByScope`
   skipped — so the failed boot turn was transcript-dark to History/State/API.
   The resolver now work-dir-filters `pending-*` placeholders and returns one as
   a last resort when no real mirror exists; a real mirror always wins.

4. **Unpaired user tail on an unparsable rc=0 reply (false `PhaseBusy`).** In
   `internal/worker/adapters/zcode/zcode-repl`, the reply `else` branch (rc=0,
   unparsable response) printed an error marker but never closed the published
   user turn out, so an idle pane could false-report busy. It now calls the
   helper's `cancelled` mode to close the turn out, mirroring the rc≠0 branch.

## Tests (TDD)

Each finding has a regression test that fails on the unfixed code and passes
with the fix:

- `internal/worker/adapters/zcode/adapter_test.go` —
  `TestLastUserTextScansBackwardForTheLastUserMessage`,
  `TestUnparsableResponseClosesTheMirrorEntry`
- `internal/session/submit_test.go` —
  `TestSubmitDefaultDeferredResetPendingStampsEmptyEpochToSurviveRotation`,
  `TestSubmitDefaultDeferredRestartRequestedKeepsCurrentEpoch`
- `internal/sessionlog/zcode_reader_test.go` —
  `TestFindZCodeSessionFileByScopePrefersRealMirrorOverPending` (plus the
  updated placeholder-fallback assertion in
  `TestFindZCodeSessionFileByScopeSeparatesSameWorkDirSessions`)

Finding 1 is pinned at the helper, not end-to-end: because publish-at-start
leaves a trailing user entry in the happy path, a two-turn adapter run does not
reproduce the duplication (it passes on the unfixed code), so it would not be a
regression. `TestLastUserTextScansBackwardForTheLastUserMessage` exercises the
true trigger directly — an assistant-tailed transcript must still yield the
prior user text — and fails on the unfixed helper.

Scope: three production files (`internal/session/submit.go`,
`internal/sessionlog/zcode_reader.go`,
`internal/worker/adapters/zcode/zcode-repl`) plus their tests. No product
behavior changed.

The `last_user_text` regression exercises the embedded python helper through
a `python3` subprocess, which adds one subprocess call site. The checked
resource-census subprocess baselines are bumped by one to match (source
all 626→627, source untagged 419→420, Small untagged 412→413) in
`internal/testpolicy/resourcecensus/census.go` and `test/test-resources.toml`,
with the `TESTING.md` ledger block regenerated from the TOML.


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277227717

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR lands four targeted correctness fixes (each with a regression test) to the zcode transcript adapter and the session deferred-submit path — a focused post-merge follow-up to #5397. It matters because these were real fidelity/delivery bugs in how agent turns get mirrored and how a queued submit survives a continuation reset: exactly the kind of thing that silently corrupts a transcript or drops a nudge if left in place.

### Original PR Review
Decision: approve.

Specific gaps fixed (verified during review):
- **Deferred submit lost/mis-delivered across a continuation reset** (`internal/session/submit.go`): the empty-epoch stamp on a pending reset now survives the N→N+1 rotation while a plain restart stays fenced to the unchanged epoch, and the item stays `SessionID`-pinned so delivery isn't widened. Review traced the delivery path (`splitQueuedNudgesForTarget` → `queuedNudgeMatchesTargetFence`) and confirmed the callers.
- **First-turn zcode placeholder read as empty** (`internal/sessionlog/zcode_reader.go`): the work-dir filter now runs before the `pending-` check so a placeholder is scoped like a real mirror, and a real mirror still wins via the `bestPending` fallback; the sole caller in transcript discovery benefits.
- **zcode adapter duplicate turn + dangling rc=0 tail** (`internal/worker/adapters/zcode/zcode-repl`): `last_user_text` backward-scans to the last user message (dropping the re-appended prompt after a successful turn), and the rc=0-unparsable branch now closes the turn as a verbatim mirror of its rc!=0 twin.
- **Resource-census subprocess baseline bump** (`internal/testpolicy/resourcecensus/census.go`, `test/test-resources.toml`, `TESTING.md`): all three artifacts are internally consistent for the one new `python3` site and the census test passes.

Both independent reviewers verified all four fixes and ran the affected package tests (`internal/session`, `internal/sessionlog`, `internal/worker/adapters/zcode`, `internal/testpolicy/resourcecensus`) green. No blocking or major issues were found, so no maintainer changes were needed.

- **ga-hjk3t** — [Release Safety / Documentation Fidelity, minor] The commit message body of `efb1e8eda9` cites stale absolute census baselines (`626→627 / 419→420 / 412→413`) versus the values actually committed (`628→629 / 421→422 / 414→415`, each +2 higher — consistent with a rebase onto already-bumped baselines). The `+1 subprocess site` rationale and every committed code artifact are correct and pass; only the absolute numbers in the commit narrative are stale. Not addressed here (correcting it would mean rewriting the existing commit); filed as `ga-hjk3t` so future census archaeology doesn't chase phantom deltas.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `a262d2c9` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/32799518742

### Review Iterations
1 review pass (attempt 1), reaching approval with 0 fix iterations — the change was accepted as-is with no maintainer commits.

---

PR #5535, opened 2026-08-23
test(productmetrics): separate hang detectors from latency floors

quad341:
## Summary

- convert 28 productmetrics test hang-detector waits to the package `hangBudget`
- leave the 8 direct floor calls that are live inputs or scenario values unchanged
- add a guard test that keeps `hangBudget` above both the latency floor and the largest replaced multiplied shape
- include the deployer PASS record in `release-gates/ga-nvcmln-productmetrics-hang-budget-gate.md`

Bead: `ga-nvcmln`  
Implementation: `ga-42mt5x.1`  
Review: `ga-vzyz80` — PASS  
Release evaluation: PASS

## Test plan

- `go test -count=1 -json ./internal/productmetrics/...` — 464 pass, 0 fail, 0 skip
- all 17 diff-owned top-level tests pass
- `make test-ci-policy`, `make fmt-check`, `make vet`, `go build ./...` — pass
- affected and full lint with isolated on-disk cache — 0 issues
- `make test-local-full-parallel` — 35/40 jobs pass; five unrelated failures are base-reproduced or covered by the active Mayor waiver, as attributed in the release-gate record
- mandatory pre-commit and pre-push hooks — pass

No production code changes.

---
- Related to #5656 — addresses the first of the two failures in that issue: the wait behind control_unix_test.go's "off did not reach uploader phase" in TestDisableAndPurgeBoundsInitialAndPostUploaderStateLocks now uses the package hangBudget (6x) instead of the bare 10s testutil floor, which is the exact ceiling that run consumed; the second failure, TestRecordOnceDecisionExpiryStopsUnchangedNoReplaceRetry's "made 0 attempts" at spool_unix_test.go:7819, is untouched by this diff and still open


quad341:
@quad341 — this is the good version of a mechanical refactor. What makes it reviewable in minutes rather than an hour is that `hangbudget_test.go` writes down *why* each site was or wasn't converted, and then the guard test encodes it.

Leaving the six `disableUploaderWait` values and the two `lockState`/`acquireLock` contexts on the floor is exactly right — those are deadlines the test hands to the system, and widening them would have quietly made those tests prove less. Deriving from `GoroutineRaceTimeout` instead of declaring a fresh 60s keeps one source of truth with the `cmd/gc` precedent.

I traced the import removals and the build tags against your branch rather than trusting the diff — the three files that dropped `testutil` have zero remaining references, the three that kept it still need it, and the tag on the new file matches every consumer. Package runs clean here too. Merging as-is; nothing for me to clean up.

One note for whenever you next touch this: the third guard assertion reduces to `hangBudget >= 20s`, so it's a weaker fence than the comment above it reads. Not worth a revision now — just something to tighten if this pattern spreads to a third package.


---

quad341:
Re-reviewed `1db25fc8a`..`58d32fde8` (1 new commit since the last review): no new concerns — the prior review stands.

PR #5537, opened 2026-08-23
fix(dolt): probe the alternate managed-Dolt port during recovery instead of discarding it

### LINKED ISSUE #5536: bug(dolt): supervisor store circuit breaker stays pinned to a stale port after a managed-Dolt rebind; `supervisor reload` does not re-resolve it
### Before you continue

### Gas City version

1.4.1 (locally built, commit `d2a2806c`-dirty). Mechanism re-verified against
`origin/main` at `16bfb855b`.

### Environment

macOS (Darwin 25.5.0), Mac mini, zsh, managed-Dolt pack, tmux provider,
single supervisor, `dolt-server.port = 36266` in `dolt-config.yaml`.

### Summary

After a breaker-triggered managed-Dolt restart, the supervisor's bead-store
circuit breaker stayed keyed to a port the scope watchdog was **not** serving.
It probed the dead endpoint every ~8s for ~7 minutes, every lane fast-failed,
and `gc supervisor reload` did not re-resolve it. Only a full supervisor
restart cleared it.

This is the same failure class as #1656 (stale port survives a Dolt rebind
while the state files are correct), one layer up: there the stale port is a
session's `GC_DOLT_PORT` env; here it is the port baked into a live bead-store
handle inside the supervisor process itself.

### Reproduction

1. Run a city with managed Dolt on port X (`dolt-config.yaml`,
   `dolt-server.port = X`).
2. Put the box under memory pressure until the Dolt store breaker trips
   (`[circuit-breaker] 127.0.0.1:X/cr: closed → open`).
3. The managed-dolt provider restarts Dolt. While the old listener is still
   bound/TIME_WAIT, the start path bumps to X+1
   (`cmd/gc/dolt_start_managed.go:299`, `nextAvailableManagedDoltPortForHost(host, currentPort+1)`),
   then the scope watchdog settles back on the config port X.
4. The supervisor re-opens/keys its `cr` and `st` store breakers at X+1.
5. `gc supervisor reload`.

### Expected behavior

Either the supervisor re-resolves the managed-Dolt endpoint after a provider
restart, or `reload` rebuilds the stores when the managed-Dolt runtime port
changes. In no case should the supervisor keep probing a port nothing is
serving while a healthy server is listening on the configured port.

### Actual behavior

The breaker stays pinned to X+1 indefinitely. Lanes fail fast
(`collectAssignedWorkBeads ... dolt circuit breaker is open: server appears
down, failing fast`), drains are skipped (`store query partial`), and
order-tracking sweeps fail. `bd` itself is fine throughout, because `bd`
re-reads `dolt-server.port` per invocation.

Real incident, 2026-08-23, X = 36266. Timeline from `~/.gc/supervisor.log`
(lines 443680-445963) and `.gc/runtime/packs/dolt/dolt.log`:

```
06:09:00  supervisor 62579: [circuit-breaker] 127.0.0.1:36266/cr: closed → open (5 failures in 3.7s)
06:09:01  [circuit-breaker] 127.0.0.1:36266/st: closed → open
06:10:05  dolt.log: gc scope watchdog: received terminated; terminating dolt sql-server pid 26597
06:10:06  new watchdog (pid 32348) starts dolt pid 32349 on the CONFIG port 36266
06:10:08  .beads/dolt-server.port rewritten = 36266
06:10:13  supervisor breaker re-keyed to 127.0.0.1:36267/cr and /st
          then: open → open (active probe failed, cooldown reset)   -- every ~8s
06:13     gc supervisor reload  -> no change, breaker still on 36267
06:16:09  SIGTERM supervisor 62579; hand-start a new supervisor -> breaker healthy on 36266, no further breaker lines
```

Nothing on disk referenced 36267 at the time except historical reconciler
traces from 08-20/08-21, when Dolt had run there. During the whole window the
supervisor still held 5 ESTABLISHED sockets to 36266 (the beads cache kept
reconciling) — only the `/cr` and `/st` store breakers were on the dead
address.

Also tried and reverted: setting `dolt-config.yaml` port to 36267 — the scope
watchdog regenerates the config and restarts on 36266.

### Mechanism (as far as I can trace it)

1. The breaker is keyed at **store-open time** and never re-keyed. It lives in
   the beads dependency: `internal/storage/dolt/circuit.go:78`
   `maybeNewCircuitBreaker(host string, port int, database string)`, called
   from `internal/storage/dolt/store.go:964`, with the probe target built at
   `circuit.go:181` from the captured `cb.host`/`cb.port`. The `/cr` and `/st`
   suffixes in the log line are the `database` component.
2. gc supplies that host/port from the resolved runtime target
   (`cmd/gc/bd_env.go`, `GC_DOLT_PORT` / `BEADS_DOLT_SERVER_PORT`), sourced
   from the published managed-Dolt state file and then the provider state
   file. A `bd` subprocess re-resolves this every invocation; a long-lived
   in-process store handle does not.
3. `gc supervisor reload` cannot fix it: the reload path rebuilds stores only
   when `storeMetadataChanged` fires (`cmd/gc/city_runtime.go:2002` →
   `cmd/gc/api_state.go:1213`), and `storeMetadataSignature`
   (`cmd/gc/api_state.go:1222`) is a sha over `scope-metadata.json` files
   only. The managed-Dolt runtime port is not an input, so a port change is
   invisible to reload and it reports "No config changes detected".
4. The most parsimonious source of the 36267 value is the address-in-use
   bump at `cmd/gc/dolt_start_managed.go:299` during the restart window, whose
   `report.Port` is published before the watchdog settles back on 36266.

I have **not** proven step 4 from logs — the Dolt log shows the watchdog
alternating 36266/36267 during the restart, which is consistent with it, but I
did not capture the state file at the instant the supervisor re-keyed.
Steps 1-3 are read directly from the source at `16bfb855b`.

### Suggested fix shapes

1. Include the managed-Dolt runtime port (published/provider state) in
   `storeMetadataSignature`, so `reload` rebuilds stores across a rebind.
2. Have the store's breaker re-resolve its endpoint from the state file on
   repeated active-probe failure, rather than probing a fixed address forever
   (client-side resilience, the same shape as the merged #4197 reconnect work).
3. Make the address-in-use port bump not publish a port the watchdog is about
   to override.

### Related work

- #1656 — stale port after a Dolt rebind, session-env layer (open, P1). Same
  class, different consumer.
- #4197 — `fix(beads): reconnect+retry NativeDoltStore reads across a
  managed-Dolt rebind` (merged). Covers read reconnect, not breaker re-keying.
- #2792 — breaker-aware managed-Dolt preflight / recover backoff (closed).
- #3844 — Dolt health failures cascading into breaker trips.
- #525 / #790 — original state-file port accuracy work.

None of these re-key or rebuild a live store breaker after a rebind, which is
what this report is about.

jacobhausler:
## What

`recoverManagedDoltObservedRebindPossible` already treats disagreement between
the provider runtime state and the published managed-Dolt state as evidence
that the managed server may have rebound to a different port. But it throws
that information away — it returns a bare `bool`, and
`recoverManagedDoltProcess` then hands the **requested** port straight to
`observeExistingManagedDoltForRecovery`.

So in exactly the case the predicate was written to detect, recovery probes the
port it already has reason to believe is stale, fails to observe the healthy
server on the alternate state port, and proceeds to stop/start against stale
endpoint state.

This extracts the alternate port instead of discarding it, and probes that.

## Changes

- `cmd/gc/dolt_recover_managed.go`: new `recoverManagedDoltObservedPort`
  returns the first live-state port that differs from the requested port
  (provider state, then published state);
  `recoverManagedDoltObservedRebindPossible` becomes a thin wrapper over it, so
  the detection rule is unchanged and stays in one place.
- `cmd/gc/dolt_recover_managed.go`: the existing-server observation uses that
  alternate port when there is one, and the requested port otherwise.
- `cmd/gc/dolt_recover_managed_test.go`: regression test — provider state on
  36267, published state on 36266, requested 36267 ⇒ probe 36266.

Port allocation, process-ownership checks, readiness, the requested-vs-observed
reuse check (`recoverManagedDoltShouldReuseExisting`), and live runtime state
are all untouched. The only behavioural change is *which state-derived port is
handed to the existing-server observation*.

## Verification

Rebased onto `origin/main` `16bfb855b`; cherry-pick applied clean.

```
CGO_CPPFLAGS=-I$(brew --prefix icu4c)/include \
CGO_LDFLAGS=-L$(brew --prefix icu4c)/lib \
  go build ./cmd/gc                                  OK
  go test ./cmd/gc -run TestRecoverManagedDolt -count=1   ok  2.995s
  go test ./cmd/gc -run Dolt -count=1                     ok  69.429s
  go vet ./cmd/gc                                    clean
  git diff --check                                   clean
```

(On macOS a bare `go build` fails without the ICU CGO flags the Makefile sets;
that is a toolchain requirement, not a build break.)

## Honest scoping — does this fix #5536?

**No, and I want to be plain about that.** #5536 is a supervisor holding a
bead-store handle whose circuit breaker is pinned to a stale port; that breaker
is keyed once at store-open time in the beads dependency and nothing in gc
re-keys it, and `reload` does not rebuild the store because the managed-Dolt
runtime port is not part of `storeMetadataSignature`. This patch does not touch
any of that.

What it does fix is a real, separate latent defect in the *recovery* path,
which is adjacent: the supervisor reaches `gc dolt-state recover-managed`
indirectly through the beads provider script's `recover` op
(`examples/bd/assets/scripts/gc-beads-bd.sh`), so a recovery that probes the
stale port during a rebind window is one more way the system can converge on
the wrong endpoint. I filed it as `Refs`, not `Fixes`, deliberately.

## Is this us swimming against the river?

Mostly no, with one caveat I would rather state than hide.

Not against the river: this is a two-line behavioural change inside a function
whose stated purpose is already "detect a possible rebind." It makes the code
use the fact it just computed. It adds no new concept, no new config, no new
state file, and it collapses two near-identical state-file walks into one
helper. If a maintainer had written the predicate to return `(string, bool)`
originally, this is what it would have looked like.

The caveat: the *issue* it came out of probably wants a different fix — either
putting the managed-Dolt runtime port into the reload store signature, or
teaching the store breaker to re-resolve its endpoint after repeated probe
failures. Those are design calls I did not want to make unilaterally in
someone else's supervisor-reload semantics. If maintainers would rather have
one of those and consider this patch noise next to it, close it and I will not
argue — the issue is the contribution that matters here, and this is a small
piece of hardening I had in hand.

Refs #5536.

### DIFF
--- a/cmd/gc/dolt_recover_managed.go
+++ b/cmd/gc/dolt_recover_managed.go
@@ -110,7 +110,11 @@ func recoverManagedDoltProcessWithOps(cityPath, host, port, user, logLevel strin
 	}
 
 	if recoverManagedDoltObservedRebindPossible(cityPath, port) {
-		if ready := observeExistingManagedDoltForRecovery(cityPath, host, port, user, recoverManagedDoltExistingObserveTimeout(timeout), &report); ready && recoverManagedDoltShouldReuseExisting(report.Port, port) {
+		observePort := port
+		if alternatePort := recoverManagedDoltObservedPort(cityPath, port); alternatePort != "" {
+			observePort = alternatePort
+		}
+		if ready := observeExistingManagedDoltForRecovery(cityPath, host, observePort, user, recoverManagedDoltExistingObserveTimeout(timeout), &report); ready && recoverManagedDoltShouldReuseExisting(report.Port, port) {
 			report.Ready = true
 			report.Healthy = true
 			if err := ops.publish(cityPath); err != nil {
@@ -255,16 +259,28 @@ func recoverManagedDoltObservedRebindPossible(cityPath, requestedPort string) bo
 	if requestedPort == "" {
 		return true
 	}
+	return recoverManagedDoltObservedPort(cityPath, requestedPort) != ""
+}
+
+// recoverManagedDoltObservedPort returns a live-state port different from the
+// requested port. Provider state can lag a scope watchdog rebind, so probing
+// only the requested port would miss the healthy managed server and start a
+// second recovery against stale endpoint state.
+func recoverManagedDoltObservedPort(cityPath, requestedPort string) string {
+	requestedPort = strings.TrimSpace(requestedPort)
+	if requestedPort == "" {
+		return ""
+	}
 	for _, path := range []string{providerManagedDoltStatePath(cityPath), managedDoltStatePath(cityPath)} {
 		state, err := readDoltRuntimeStateFile(path)
 		if err != nil || !state.Running || state.PID <= 0 || state.Port <= 0 {
 			continue
 		}
 		if strconv.Itoa(state.Port) != requestedPort {
-			return true
+			return strconv.Itoa(state.Port)
 		}
 	}
-	return false
+	return ""
 }
 
 func recoverManagedDoltRepairRuntimeStateForHealthyPort(cityPath, requestedPort string) error {
--- a/cmd/gc/dolt_recover_managed_test.go
+++ b/cmd/gc/dolt_recover_managed_test.go
@@ -432,6 +432,28 @@ func TestRecoverManagedDoltObservedRebindPossible(t *testing.T) {
 	})
 }
 
+func TestRecoverManagedDoltObservedPortUsesAlternatePublishedState(t *testing.T) {
+	cityPath := t.TempDir()
+	if err := writeDoltRuntimeStateFile(providerManagedDoltStatePath(cityPath), doltRuntimeState{
+		Running: true,
+		PID:     1234,
+		Port:    36267,
+	}); err != nil {
+		t.Fatalf("write provider state: %v", err)
+	}
+	if err := writeDoltRuntimeStateFile(managedDoltStatePath(cityPath), doltRuntimeState{
+		Running: true,
+		PID:     5678,
+		Port:    36266,
+	}); err != nil {
+		t.Fatalf("write published state: %v", err)
+	}
+
+	if got := recoverManagedDoltObservedPort(cityPath, "36267"); got != "36266" {
+		t.Fatalf("recoverManagedDoltObservedPort() = %q, want alternate published port 36266", got)
+	}
+}
+
 func setupRecoveryTestCity(t *testing.T) string {
 	t.Helper()
 	cityPath := t.TempDir()


quad341 (approved):
@jacobhausler — merging this. The core observation is a good catch: `recoverManagedDoltObservedRebindPossible` was computing exactly the fact recovery needed and then throwing it away, and turning it into a `(port)` helper with the predicate as a thin wrapper is the shape it should have had from the start. I diffed the guard conditions and the detection rule really is unchanged, which made this fast to trust.

Special thanks for the scoping section. You could have written "Fixes #5536" and let it ride; instead you explained why the breaker-pinned-to-a-stale-port problem lives elsewhere. I checked `storeMetadataSignature` and you're right — it hashes scope metadata and never sees the managed-Dolt runtime port, so reload won't rebuild the store on a rebind. Correctly reading someone else's supervisor semantics and then *declining* to claim the fix is the rarer skill.

Two notes for later, no action needed. The port you pass to `assessExistingManagedDolt` doesn't drive the readiness probe — that uses the PID-matched `StatePort` — it drives PID discovery via the port-holder fallback, so your fix lands in a narrower window than the description suggests. And `recoverManagedDoltObservedPort` doesn't check `pidAlive` the way its siblings do; I'm filing that follow-up on our side.


---

PR #5538, opened 2026-08-23
fix(runtime): support T3 Code 0.0.33 websocket auth

pzxy:
## Summary

- discover the stock T3 Code runtime state under `~/.t3/userdata/server-runtime.json`, ahead of legacy locations
- exchange the bearer session token for a short-lived `/api/auth/websocket-ticket` and connect with `wsTicket`, while retaining the legacy Authorization-header fallback on 404/405
- load the current orchestration snapshot through the authenticated HTTP endpoint, continue across stale candidates, and retain the legacy WebSocket snapshot fallback
- redact bearer tokens from HTTP error details

## Testing

- `go test -count=1 ./internal/runtime/t3bridge`
- `go vet ./...`
- live smoke test against T3 Code 0.0.33: ticket authentication and HTTP snapshot succeeded; Gas City created the project/thread and sent the Mayor startup prompt

The local pre-push fast sweep did not complete: several unrelated `cmd/gc` shards timed out while compiling/listing tests under heavy concurrent local Go load. No t3bridge assertion failed, and the focused package suite plus full vet passed.

## Scope

This PR intentionally covers the authentication and snapshot boundary only. Full stock 0.0.33 runtime compatibility still needs separate protocol work for persistent session/thread binding (`customMetadata` is absent), `vcs.*` worktree calls, and newer thread activity/message APIs.


quad341:
@pzxy — thank you for this. The part I appreciate most is that you didn't just swap in the new auth: every new path keeps a tested escape hatch. The ticket exchange falls back to the `Authorization` header on 404/405, the HTTP snapshot falls back to `orchestration.getSnapshot` over the socket, and `TestRPCSnapshot_TriesNextCandidateBeforeLegacyFallback` proves a stale candidate doesn't drag us into the legacy path early. Redacting the bearer out of echoed error bodies on *both* new endpoints, with tests asserting it, is the kind of thing that usually shows up three incidents later.

I'm taking one small hardening on our side and merging — no work for you. Your fallbacks key on the status code alone, so a bridge that serves an SPA catch-all (200 `text/html` at an unknown route) fails JSON decode and never reaches the fallback you built. I'll widen both to treat a non-JSON 200 and a 501 as "unsupported," and add tests for that plus a `result`-wrapped HTTP snapshot.

Our CI never ran a real suite on this head — our plumbing, not yours. My push will trigger it.

Your scope note about `customMetadata` and `vcs.*` is flagged to a maintainer so the follow-up gets a home.


---

quad341:
@pzxy, an update on the merge we promised on 09-08, and sorry it took this long. Our follow-up commit on your branch tripped our own lint gate. It formatted the JSON decode error with `%v` next to the `%w` sentinel, which drops the decode cause from the error chain (errorlint). That was our bug, not yours.

I've pushed the corrected follow-up to your branch as e9e99c12c4. It wraps both errors, so `errors.Is` still matches the unsupported sentinel that triggers the fallback, and the decode cause stays inspectable. `golangci-lint` reports 0 issues on `internal/runtime/t3bridge`, and its tests pass. I'll merge once CI is green. Your commit and credit stay as they are, and there's nothing you need to do.

PR #5540, opened 2026-08-23
fix(bd-env): pin native-Dolt pool ceiling to 1 conn for one-shot store opens

### LINKED ISSUE #5539: gc in-process store opens: one-shot CLI inherits 10-wide daemon pool ceiling; 4 TCP conns/invocation churns ~40 conns/s against the city dolt
## Summary

Every `gc` CLI invocation that opens its bead store in-process — `gc ready`,
`gc convoy status`, `gc hook --claim`, orders' `scale_check` probes — opens
**4 TCP connections** to the city's `dolt sql-server` per invocation. Only one
of them (the project pool) does any of the command's work. At the ~10
in-process store opens/second a busy city makes, that is **~40 new
connections/second** against the dolt server, with zero slow queries on the
server side. The bd-side 3→2 trim (beads #5966, merged) did not explain this
rate; it belongs to gascity's own store-open path.

| conn | opened by | queries it runs |
|---|---|---|
| A | beads `newServerMode` fail-fast `net.DialTimeout` | none — reads the greeting, closes (dolt crash-detection, GH#4132/#4133) |
| B | beads `openServerConnection` no-database `initDB` pool | `SELECT @@max_allowed_packet` (driver) + one `SHOW DATABASES` |
| C | beads `openServerConnection` project pool | all ~29–66 real queries of the command |
| D | gascity pre-open provider health/recovery preflight in `cmd/gc/bd_env.go` (`healthBeadsProviderContext` via the managed-dolt target resolution) | provider health op |

Conns A–C come from the beads library's `openServerConnection`; conn D is
gascity-specific and is absent from a plain `bd` CLI open. The count is
deterministic: 10 consecutive `gc ready` runs against an isolated scratch
`dolt sql-server` (own data dir, own port, orchestrator-free) produced exactly
40 connections, zero variance.

## Measured numbers (isolated repro, no orchestrator)

Scratch `dolt sql-server` on 127.0.0.1:36990 with a dedicated data dir and a
scratch `gc` city pinned to it; 3 seeded beads + 1 convoy; ambient city env
scrubbed. Best-of-3 connection deltas from `SHOW GLOBAL STATUS 'Connections'`,
2 status-probe connections subtracted per run:

| command | conns/invocation | selects/invocation | RSS (MB) | wall |
|---|---|---|---|---|
| `gc ready --metadata-field routed_to --limit 1` | **4** | 29 | 71 | 0.59s |
| `gc bd show <bead>` | **1** | 21 | 72 | 0.25s |
| `gc bd update <bead> --append-notes …` | **7** | 66 | 70 | 0.30s |
| `gc convoy status <convoy>` | **4** | 43 | 70 | 0.38s |

`gc bd show/update` shell to the `bd` binary (already trimmed 3→2 by beads
#5966), so they are not the churn source. `gc ready` and `gc convoy status`
are fully in-process: those two alone account for the 4-conn-per-invocation
rate. `gc hook --claim` was not run against a live store (it would claim real
work); it goes through the same `OpenNativeStore` → `openServerConnection`
path as `gc ready`, so it is expected to open the same 4 connections.

## Why the documented pool knob cannot help

`BEADS_DOLT_MAX_CONNS` / `dolt.max-conns` (added for GH#3140) cap the
**project pool's** `MaxOpenConns`. A serial one-shot CLI never checks out
more than one pool connection, so capping it from 10→1 changes nothing in the
count: conns A and B are a raw dial and a second `sql.DB` open outside the
pool. Lowering the knob is the natural first suggestion for a connection-rate
report, and it is a dead end for the same reason documented on the bd side
(beads #5965). The knob *does* matter as a ceiling, though — see Fix 1.

## Proposed fixes

**Fix 1 (companion PR, this issue's first step): pin the pool ceiling for
in-process opens.** `cmd/gc/bd_env.go: nativeDoltOpenEnvForScopeContext`
builds the env projected into the beads library before every native-Dolt
store open but does not set `BEADS_DOLT_MAX_CONNS`, so every one-shot open
inherits the daemon-oriented default of 10 open / 5 idle. The companion PR
pins `BEADS_DOLT_MAX_CONNS=1` there. Measured effect: none today (a serial
CLI already opens one pool connection); it is a zero-behaviour regression
guard that makes the pool ceiling match the one connection a one-shot command
can actually use, and it is the smallest safe change on this path.

**Fix 2 (staged design, not yet a PR): eliminate the redundant per-invocation
connections.** Two independent cuts:

1. **Conn B (beads-side):** cache the "database exists" fact in
   `.beads/metadata.json` after the first successful open and skip the
   no-database `initDB` pool on subsequent opens. The same reasoning beads'
   `Gateway` branch already applies ("a successful connect IS the existence
   proof") generalises here for a managed, externally-owned database that is
   never dropped between invocations.
2. **Conn D (gascity-side):** the managed-dolt target resolution runs a
   provider health op before the command's own store open. When the command
   is about to open the store anyway (fail-fast on open is itself the health
   check), the pre-open probe is redundant and should be skipped.

Expected result of Fix 2: **4 → 2 connections per invocation** (probe A
retained for the crash-detection class it was hardened for, per GH#4132/#4133
and beads #5965). Fix 2 is staged rather than open because it needs a rebuild
+ re-measure to confirm the 4→2 delta and it touches the provider-lifecycle
path; the beads-side half also belongs to the beads repository.

## Relationship to existing work

- **beads #5965 / #5966** — bd-side 3→2 trim (merged; live in the pinned bd
  1.2.2). This issue is the gc-side twin of the same churn.
- **GH#4102** — remote-dolt latency. Complementary: this is connection
  *count* on a low-RTT server where wall time is fine but the accept/handshake
  rate is 4× what the work needs.
- **GH#3140** — why `dolt.max-conns` exists; the measurements above show why
  it does not address per-invocation connection rate.
- **GH#4132 / GH#4133** — the dolt crash class conn A protects against; both
  fixes retain A.

## Environment

- `gc` 1.4.1 (built 2026-08-22 from a dirty local tree; beads module replaced
  to a local source copy), gascity command layer confirmed against main
- macOS 15 (darwin/arm64), `dolt sql-server` on loopback, `max_connections: 256`
- `dolt_mode: server`, externally managed, `BEADS_DOLT_AUTO_START=0`

jacobhausler:
Fixes #5539 (pool-ceiling part).

## What

`nativeDoltOpenEnvForScopeContext` (`cmd/gc/bd_env.go`) builds the env
projected into the beads library before every native-Dolt in-process store
open, but did not set `BEADS_DOLT_MAX_CONNS`. Every one-shot CLI invocation
therefore inherited the library's daemon-oriented pool defaults
(`MaxOpenConns=10` / `MaxIdleConns=5`, "deliberately oriented at long-lived
daemons").

This pins `BEADS_DOLT_MAX_CONNS=1` in that env via a small
nil-guarded helper (`nativeDoltCliPoolCap`) that covers both return paths of
the function.

## Why

gc opens the store fresh on every one-shot invocation (`gc ready`,
`gc convoy status`, `gc hook --claim`, orders' `scale_check` probes). At the
~10 in-process store opens/second a busy city makes, a 10-wide pool ceiling
permits ~40 new TCP connections/second against the city dolt server, although
a serial one-shot never holds more than one pool connection at a time
(measured: 4 connections per `gc ready` invocation, of which exactly one is
the pool's — see #5539).

## Effect

- **Measured reduction today: none** — a serial CLI already opens exactly one
  pool connection. This makes the ceiling explicit and hard: if a
  concurrent-query path is ever added to a one-shot command, the pool cannot
  silently grow back to the daemon default.
- No query-semantics, schema, or data change. The cap is applied through the
  documented `BEADS_DOLT_MAX_CONNS` knob.

## Note for review

`nativeDoltOpenEnvForScopeContext` also serves the in-process
API/controller store opens (they share this seam). For that path the cap
bounds its pool to one connection, which is safe under `database/sql`
blocking semantics (concurrent top-level queries serialize; a deadlock would
require code to hold a connection while requesting a second one from the same
pool). Flagging this explicitly so maintainers can judge: if the API path
should keep a wider pool, the follow-up is to distinguish the one-shot
callers from the long-lived ones — deliberately out of scope for this
minimal change.

## Verification

- `gofmt` clean; change is a two-line projection plus a nil-guarded 5-line
  helper.
- Full `go build` not completed in the authoring environment (missing
  CGO/ICU dev headers for `go-icu-regex`, unrelated to this change); the
  assignment is type-correct by inspection and uses the env var name the
  beads library reads (`BEADS_DOLT_MAX_CONNS`, `open.go`).
- Measurement context for the churn this guards: isolated scratch
  `dolt sql-server` + scratch `gc` city, `gc ready` = 4 conns/invocation
  (probe + initDB + 1 pool conn + gascity preflight), 29 selects, 0.59s wall;
  `gc convoy status` = 4 conns/43 selects.

## Not included here (see #5539, Fix 2)

- Eliminating the fail-fast probe (conn A) and the no-database `initDB`
  connection (conn B): both live outside the pool, so capping it cannot
  touch them. That 4→2 reduction is a staged design needing rebuild +
  re-measure and, for conn B, a beads-side change.


quad341 (approved):
@jacobhausler — thank you for this, and particularly for the description. Writing "Measured reduction today: none" in your own PR body, and flagging that the API and controller opens share this seam before anyone asked about it, is the kind of self-review that makes a change easy to trust.

Tracing it turned up one thing from our side: the env map you set the cap in gets projected through a fixed key allowlist in the native store open (`nativeDoltOpenEnvKeys`), and `BEADS_DOLT_MAX_CONNS` isn't on it yet — so the value is dropped before the library's `os.Getenv` can see it. You picked the right knob; the library reads exactly that name. It's our wiring gap, and it's genuinely invisible from where you were working.

So I'll take it on my side: add the key to the allowlist, add the projection test that would have caught this, and — following your own "Note for review" — apply the cap only at the one-shot opens so the long-lived controller stores keep their wider pool. Your commit and framing stay as-is.

If you want the one-shot vs. long-lived caller split you scoped out, or the conn A/B reduction from #5539, say the word and I'll loop you in.


---

PR #5541, opened 2026-08-23
fix(molecule): preserve source spec JSON during substitution

jamelt:
Summary:
- JSON-escape runtime variable values before substituting into serialized source-spec sidecars.
- Preserve ordinary description substitution unchanged.
- Cover multiline, quote, and backslash values with a focused regression test.

Why:
A multiline formula variable was inserted directly into the already-serialized retry step snapshot. The resulting literal newline made the source-spec JSON unreadable, so retry control quarantined with invalid character newline in string literal.

Tests:
- go test ./internal/molecule
- make test-fast-parallel (pre-push gate)


quad341 (approved):
@jamelt — this is a well-diagnosed fix. A multiline variable dropping a literal newline into an already-serialized step snapshot is exactly the kind of bug that surfaces as a baffling quarantine message three layers downstream, and escaping at the substitution boundary rather than patching the parser is the right place to fix it. Gating on the spec kind so ordinary descriptions keep their raw substitution is the detail that makes this safe to take. I checked both instantiation paths and the graph-apply route picks the fix up too.

I'm taking this on our side rather than sending it back. I'll add one small companion test covering the other branch, that a normal task description with a newline still substitutes raw, so nothing later widens that kind check without a test noticing. CI never actually ran on this head, so I'll get a real suite attached before it lands rather than merging on a green rollup made entirely of triage labelers.

Your regression test stays as written and your commit keeps authorship. Thanks for chasing this one down to the actual boundary.


---


Review coverage:
- completed: `qwen claude`
- unavailable: `codex`

PR #5543, opened 2026-08-23
fix(runtime): preserve T3 observation uncertainty

boshu2:
## Summary

This is the independently mergeable runtime-observation successor to #5015. It rebuilds that PR's liveness carrier on current `main` and leaves lifecycle policy to the stacked follow-up.

T3 liveness is not a boolean when the bridge is unavailable. This change adds an optional error-bearing observation contract so callers can distinguish:

- confirmed live;
- confirmed absent from a healthy snapshot; and
- unknown because observation failed or the matching snapshot entry is malformed.

Legacy providers remain source-compatible and retain error-free fallback behavior.

## Design

- `runtime.LivenessObserverWithError` is an optional capability; `ObserveLivenessWithError` falls back to the existing observer/provider methods.
- `auto`, `hybrid`, session, worker, bounded-status, attachment-cache, and T3 seam decorators preserve the richer result.
- T3 classifies the official live states (`idle`, `starting`, `running`, `ready`) and terminal states (`interrupted`, `stopped`, `error`) from one snapshot.
- Duplicate T3 records are grouped by GC session name and every liveness surface selects the same newest non-deleted, non-archived thread.
- Recent-start evidence remains provisionally live while T3 materializes a session.
- `ListRunning` returns usable live names with `PartialListError` when unrelated entries are uncertain, and a total `ErrRuntimeUnavailable` when no authoritative list exists.

The status vocabulary is grounded in T3's current contract: [orchestration status union](https://github.com/pingdotgg/t3code/blob/b1670ac7d9b5b7bb9d7ebd969f27384daee22813/packages/contracts/src/orchestration.ts#L289-L297) and [ready/idle awareness](https://github.com/pingdotgg/t3code/blob/b1670ac7d9b5b7bb9d7ebd969f27384daee22813/packages/shared/src/agentAwareness.ts#L107-L114).

## Validation

- Standalone commit validated in a detached worktree: all affected runtime/session/worker packages passed.
- Full `internal/runtime/t3bridge` package passed.
- Focused race suites passed for runtime, routing, T3, session, worker, and CLI wrappers.
- `go vet ./...`, changed-package lint, repository-wide `lint-affected`, formatting, generated docs, docsync, build, and resource census passed.
- The real pre-commit hook passed for this exact commit.

The repository's broad pre-push suite is not currently clean on this macOS host/current `main` (known path, tmux, pidutil, Dolt concurrency, and timing-sensitive failures). The branch was therefore pushed with `--no-verify` only after the changed-path gates above passed. An isolated occurrence of `TestReleaseOrphanedPoolAssignments_DetachedProbeDeadReleasesAndClears` passed three times on both this candidate and current `main`.

## Sequencing

- The lifecycle consumers are in the stacked follow-up #5544.
- The routed-demand fingerprint fix is independently proposed in #5542.
- #5098 is still open/conflicting. If it lands first with a canonical observation carrier, this PR should translate onto that carrier rather than land two competing abstractions; the tests and T3 classification remain reusable.

Supersedes the runtime-observation portion of #5015.


quad341 (approved):
@boshu2 — thank you for rebuilding this on current `main` as an independently mergeable carrier and deliberately leaving lifecycle policy to #5544. That sequencing is what makes it reviewable: the observation contract lands on its own merits and nobody has to reason about observation and policy changing in one diff.

The `provider_test.go` table is the part worth calling out — driving all eleven T3 statuses through `ObserveLivenessWithError`, `IsRunning`, `ProcessAlive`, and `ListRunning` in a single case pins the legacy bool surfaces against the new one so they can't quietly drift. That's what let us verify the decorators instead of taking them on faith.

Merging as-is. Two things I'm recording for the #5544 follow-up rather than blocking on. `threadSessionStatus` and `cacheSnapshot` both guard with `if session != nil`, so a thread with no session object appears to be a shape the existing code expects; it used to read as plain false, and now classifies `unknown` → `ErrRuntimeUnavailable`, which `ListRunning` can escalate to a total failure. And `ListRunning` now filters by status where it previously listed every non-deleted thread, shifting what the pool and adoption paths read as absence. Both fail safe. If you've seen a live snapshot for an ended session, that one detail closes the first out.


---

PR #5544, opened 2026-08-23
fix(lifecycle): defer destructive actions on unknown liveness

boshu2:
## Summary

This is the lifecycle-safety successor to #5015, stacked on the runtime-observation contract in #5543.

The invariant is simple: a failed liveness/activity/attachment observation is **unknown**, not proof of absence. Destructive reconciliation therefore waits for the next healthy tick instead of closing, healing, rolling back, draining, or restarting a potentially live session.

## What changes

- Pool sweep retains sessions when runtime liveness is unavailable; `ErrSessionNotFound` remains confirmed stale cleanup.
- Start preflight and collision/state-sync recovery defer without consuming wake-failure budgets, rolling back pending creation, or stopping a provisional runtime.
- Orphan close/heal and configured-named secondary probes fail closed.
- Attachment uncertainty does not persist `detached_at` or begin no-wake drains.
- Activity uncertainty preserves idle probes and prevents idle/config-drift drains or restarts.
- Drain completion waits for a healthy stopped observation, including the post-stop verification pass.
- Config-drift drain acknowledgements retain their queued drain and runtime metadata while observation is unavailable.
- Non-sentinel legacy activity errors remain best-effort; healthy confirmed absence keeps the normal convergence path.

## Test coverage

The regression matrix exercises all original #5015 outcome invariants across full reconciler/start-wave paths, including pool sweep, ordinary and pending-create starts, `ErrStateSync`, `ErrSessionExists`, named-session reappearance, orphan close, attachment/activity/config-drift uncertainty, ordinary drains, verified-stop drains, and wrapper propagation.

Validation on the two-commit stack:

- 20 focused CLI lifecycle regressions passed.
- The same focused matrix passed under `-race`.
- All affected runtime/session/worker packages passed; full T3 passed.
- `go vet ./...`, repository-wide `lint-affected`, formatting, generated docs, docsync, build, and resource census passed.
- The real pre-commit hook passed for this exact commit.

The repository's broad pre-push suite is not currently clean on this macOS host/current `main` (known platform, Dolt concurrency, and timing-sensitive baseline failures), so this branch was pushed with `--no-verify` after the changed-path gates above passed. The broad `cmd/gc` run reached completion with one detached-probe timeout; that test passed three times in isolation on both the candidate and current `main`.

## Stack and sequencing

- This PR currently contains #5543's runtime contract commit plus the lifecycle commit because a cross-fork PR cannot target another fork branch. Review the second commit for the lifecycle delta; after #5543 merges, this branch will be rebased and the diff will shrink.
- #5542 independently fixes the routed-demand fingerprint invariant that was entangled in old #5015. It is required to cover the complete old outcome set but does not depend on this lifecycle policy.
- If #5098 lands first, rebase onto its canonical observation carrier instead of retaining duplicate abstractions.

Supersedes the lifecycle portion of #5015.


quad341 (approved):
@boshu2 — this is careful, disciplined work. The invariant you led with, that a failed observation is *unknown* rather than proof of absence, is exactly right, and you carried it through every consumer instead of patching the single path that bit you. `internal/runtime/runtime.go` already documented that contract for `ErrRuntimeUnavailable`; this is the first change that actually makes the reconciler honor it end to end.

The regression matrix is what makes it reviewable. Driving the real reconcile and start-wave entry points with sequenced providers that fail one specific observation call, then asserting metadata equality plus zero start/stop calls, is far more convincing than helper-level assertions. I traced the T3 change specifically: it reads the same `session.status` field the old status helper already cached, so no new dependence on the bridge's wire shape. I also checked every out-of-diff `GetLastActivity` caller now that soft-unavailable returns an error — they all already fail safe. Both checks held.

Merging as-is. Two things I'll file on our side rather than send back to you: an unrecognized T3 status currently maps to the transient sentinel, which makes a permanent condition look retryable, and the deferrals have no upper bound if the bridge stays down. Thanks also for being straight about the `--no-verify` push and the flaky detached-probe timeout — CI covered both.


---

Issue #5553, opened 2026-08-24
bug: OpenCode plugin runs optional injections serially on the 30s lifecycle timeout and swallows their failures


Toady00:
### Before you continue

- [x] I searched existing issues and did not find a duplicate.
- [x] I read the relevant docs and contributor guidance.

### Gas City version

`dev`, built from `upstream/main` @ `47b317bc6`

### Environment

macOS 15 / Apple Silicon, OpenCode 1.18.21, supervisor-managed named session, `opencode` provider.

### Reproduction

1. Start a Gas City named session using the OpenCode provider.
2. Replace `gc` on the plugin's PATH with a stub that sleeps 45s for `nudge drain --inject` and returns normally for everything else.
3. Submit a message and time the turn.
4. Inspect the OpenCode plugin log for any indication of which command stalled.

### Expected behavior

Optional, best-effort context injection is bounded by a short timeout, runs concurrently, and fails open. When it does fail, something is logged that identifies which command failed.

### Actual behavior

The two optional commands run **serially**, each with the same 30-second timeout used for lifecycle work, so one invocation can spend up to 60s in optional injection alone (90s including an uncached prime). Both are invoked through `run()`, which hardcodes `warnOnFailure=false`, so a timeout produces **no log line at all** — there is no way to tell which command stalled.

### Logs, screenshots, or traces

```shell
# overlay/per-provider/opencode/.opencode/plugins/gascity.js

async function runCommand(directory, args, warnOnFailure, extraEnv = {}) {
  try {
    const { stdout, stderr } = await execFileAsync(GC_BIN, args, {
      cwd: directory, encoding: "utf-8",
      timeout: 30000,                      # same timeout for optional and lifecycle work
      ...
  } catch (err) {
    if (warnOnFailure) { logRunFailure(args, directory, err); }   # suppressed for nudge/mail
    return "";
  }
}

async function run(directory, ...args) {
  return runCommand(directory, args, false);   # <-- warnOnFailure = false
}

async function buildPrefix() {
  const prime  = await readPrime();
  const nudges = await run(directory, "nudge", "drain", "--inject");  # serial
  const mail   = await run(directory, "mail", "check", "--inject");   # serial
  return [prime, nudges, mail].filter(Boolean).join("\n\n");
}
```

For reference, these commands are fast when healthy, so the stall is intermittent rather than an unconditional cost:

```shell
gc prime --hook           real 0.32s
gc nudge drain --inject   real 0.37s
[...]

Issue #5552, opened 2026-08-24
bug: OpenCode plugin builds its injection prefix per generation, duplicating context and draining nudges more than once per turn


Toady00:
### Before you continue

- [x] I searched existing issues and did not find a duplicate.
- [x] I read the relevant docs and contributor guidance.

### Gas City version

`dev`, built from `upstream/main` @ `47b317bc6`

### Environment

macOS 15 / Apple Silicon, OpenCode 1.18.21, supervisor-managed named session, `opencode` provider.

### Reproduction

1. Start a Gas City named session using the OpenCode provider.
2. Queue a nudge for the agent so `gc nudge drain --inject` has something to return.
3. Submit a message whose turn spawns at least one subagent/task generation.
4. Observe how many times `gc prime --hook`, `gc nudge drain --inject`, and `gc mail check --inject` are executed for that single turn, and how many copies of the prime text reach the model.

### Expected behavior

Per-turn context injection happens once per turn. A consumptive queue (`nudge drain --inject`) is drained at most once per turn, and the drained content is injected exactly once.

### Actual behavior

`buildPrefix()` is invoked from two separate hooks, and the system-transform hook fires once per *generation* rather than once per turn:

- `chat.message` → `buildPrefix()`
- `experimental.chat.system.transform` → `buildPrefix()`

so the prime text is injected twice into the same turn (once into `output.message.system`, once into `output.system[0]`), and `gc nudge drain --inject` — which is consumptive — is executed repeatedly. Whichever call runs first consumes the queue; the others return empty. The drained nudges are therefore injected through only one of the injection points, and which one is incidental.

### Logs, screenshots, or traces

```shell
# overlay/per-provider/opencode/.opencode/plugins/gascity.js
"chat.message": async (_input, output) => {
  const prefix = await buildPrefix();
  if (prefix) output.message.system = prependText(output.message.system, prefix);
},

"experimental.chat.system.transform": async (_input, output) => {
  const prefix = await buildPrefix();
  if (prefix) { ... output.system[0] = prependText(output.system[0], prefix); }
},
```

From the OpenCode 1.18.21 binary, the system-transform trigger lives inside `Agent.generate`, i.e. per model generation — not per user turn:

```shell
generate: e.fn("Agent.generate")(function*(i){ ...
  let g=[se];
  yield*J.trigger("experimental.chat.system.transform",{model:w},{system:g});
```

[...]

Issue #5551, opened 2026-08-24
bug: OpenCode plugin's chat.message hook blocks user-message acknowledgement on optional gc subprocesses


Toady00:
### Before you continue

- [x] I searched existing issues and did not find a duplicate.
- [x] I read the relevant docs and contributor guidance.

### Gas City version

`dev`, built from `upstream/main` @ `47b317bc6`

### Environment

macOS 15 / Apple Silicon, OpenCode 1.18.21, supervisor-managed named session, `opencode` provider, plugin work directory = city root.

### Reproduction

1. Start a Gas City named session using the OpenCode provider.
2. Attach to it (`gc session attach <agent>` or via Herdr).
3. Make one of the optional injection commands block — replace `gc` with a stub that sleeps >30s on `nudge drain --inject` or `mail check --inject`.
4. Submit a message.

### Expected behavior

The submitted message is recorded and displayed immediately. Gas City's optional context injection may delay the *assistant's reply*, but must not delay acknowledgement of what the user typed.

### Actual behavior

The message does not appear as sent until the plugin's `chat.message` hook returns. Observed in normal use as intermittent >30s delays before a message showed as sent, with the agent unable to begin responding.

### Logs, screenshots, or traces

```shell
# overlay/per-provider/opencode/.opencode/plugins/gascity.js
"chat.message": async (_input, output) => {
  const prefix = await buildPrefix();          # awaits prime + nudge + mail
  ...
}

async function buildPrefix() {
  const prime  = await readPrime();                                  # cached after first call
  const nudges = await run(directory, "nudge", "drain", "--inject"); # 30s timeout
  const mail   = await run(directory, "mail", "check", "--inject");  # 30s timeout
  ...
}
```

Decompiled from the OpenCode 1.18.21 binary (`SessionPrompt`), showing the hook is awaited *before* the user message is saved:

```shell
yield*d.trigger("chat.message",{sessionID:...},{message:j,parts:fe});
let Z=yield*s.forEach(fe, /* image normalize */);
ie=Ay(j,{errors:"all",...});
if(Te.isFailure(ie)) yield*s.logError("invalid user message before save",{...});
yield*o.updateMessage(j);                 # <-- user message persisted HERE
for(let Y of Z) yield*o.updatePart(Y);    # <-- parts persisted HERE
```

[...]

PR #5563, opened 2026-08-24
fix(opencode): close child stdin so gc calls can finish

### LINKED ISSUE #5562: bug: OpenCode plugin never closes child stdin, so every gc injection call blocks until its timeout
### Before you continue

### Gas City version

### Environment

macOS 15 / Apple Silicon, OpenCode 1.18.21, supervisor-managed named session, `opencode` provider.

### Reproduction

1. Start a Gas City named session using the OpenCode provider and attach to it.
2. Submit a message.
3. Observe the plugin's `gc nudge drain --inject` call. It never returns on its own; it is killed when the plugin's `execFile` timeout expires.

### Expected behavior

`gc nudge drain --inject` returns as soon as it has done its work.

### Actual behavior

It blocks for exactly the `execFile` timeout, every time, then dies of SIGTERM. Raising the timeout just moves the number:

```shell
gascity opencode plugin: gc nudge drain --inject ... failed after 3005ms (budget 3000ms): SIGTERM
gascity opencode plugin: gc nudge drain --inject ... failed after 10004ms (budget 10000ms): SIGTERM
```

Pinning to the budget on two different budgets is the signature of a block that never completes, not of slow work.

### Cause

The plugin invokes `gc` through `promisify(execFile)` with no `stdio` option:

```js
const { stdout, stderr } = await execFileAsync(GC_BIN, args, {
  cwd: directory,
  encoding: "utf-8",
  timeout,
  env: { ... },
});
```

Node gives the child a stdin **pipe** and never closes it, so the child never sees EOF. Demonstrated with a child that simply reads stdin:

```shell
as-is        -> FAILED after 3002 ms: SIGTERM
stdin.end()  -> ok in 3 ms
```

Note `stdio` is not an `execFile` option — it is honored by `spawn`. Passing it to `execFile` has no effect, because `execFile` forces pipes in order to capture output.

On the gc side, `drainHookStdin` (`cmd/gc/cmd_hook.go`) deliberately skips only a char device:

```go
if f, ok := stdin.(*os.File); ok {
    if st, err := f.Stat(); err != nil || st.Mode()&os.ModeCharDevice != 0 {
        return nil
    }
}
...
data, _ := io.ReadAll(io.LimitReader(stdin, 1<<20))
```

with the comment *"Provider hooks always arrive on a pipe, which this guard leaves buffered and timeout-bounded."* That assumption holds for a provider that closes the pipe. It does not hold here: the pipe stays open for the life of the child, so the read waits for EOF that never comes.

**The pi hook does not have this problem**, because it passes stdin explicitly:

```js
execFileSync("gc", args, { stdio: ["ignore", "pipe", "inherit"], ... })
```

`stdio` *is* honored by `execFileSync`. `piHookNeedsUpgrade` even requires that exact marker. The OpenCode plugin has no equivalent, and `opencodeHookNeedsUpgrade` does not require one.

### Impact

Every per-turn `gc` injection call from the OpenCode plugin burns its full timeout and returns nothing. With the shipped 30s timeout that is a 30s stall per affected call, and queued nudges and unread mail are never injected — the calls are killed before they can produce output.

This is the underlying cause of the OpenCode send-latency reports; #5551, #5552 and #5553 address real defects in the same area but none of them explain a delay that lands exactly on the timeout value.

### Suggested fix

Close the child's stdin immediately, mirroring what the pi hook already achieves:

```js
const pending = execFileAsync(GC_BIN, args, { ... });
pending.child.stdin?.end();
const { stdout, stderr } = await pending;
```

Worth considering separately: `drainHookStdin`'s pipe assumption is load-bearing and silent when violated. A bounded read (it already has a ctx deadline on the `gc hook run` path) or a diagnostic when the read hits its limit would turn this class of bug from a mystery stall into a log line.

Toady00:
## Summary

Fixes #5562. **This is the root cause of the OpenCode injection stalls**; #5551, #5552 and #5553 are real defects in the same area but none of them explain a delay that lands exactly on the timeout value.

`execFile` always gives the child a stdin **pipe** and never closes it, so a `gc` subcommand that reads hook stdin waits for an EOF that never arrives. Every per-turn injection call burned its full timeout and returned nothing.

Raising the timeout only moved the number, which is what gave it away:

```
gc nudge drain --inject ... failed after  3005ms (budget  3000ms): SIGTERM
gc nudge drain --inject ... failed after 10004ms (budget 10000ms): SIGTERM
```

Pinning to two different budgets is a block that never completes, not slow work.

Demonstrated with a child that only reads stdin:

```
as-is        -> FAILED after 3002 ms: SIGTERM
stdin.end()  -> ok in 3 ms
```

On the gc side `drainHookStdin` skips only a char device, with the comment *"Provider hooks always arrive on a pipe, which this guard leaves buffered and timeout-bounded."* That holds for a provider that closes the pipe; here it stays open for the life of the child.

**The pi hook is unaffected** because it passes `stdio: ["ignore", "pipe", "inherit"]`, and `piHookNeedsUpgrade` requires that marker. `stdio` is not an `execFile` option — `execFile` forces pipes to capture output — so the OpenCode plugin needs the explicit close instead.

## Testing

- [x] `make check`
- [ ] `make check-docs` — no docs touched
- [ ] `make test-integration` — not run; see note below

`node --check` passes, `go test ./internal/hooks/` and `go vet` pass. `opencodeHookNeedsUpgrade` now requires the marker, with a new case asserting a plugin that leaves stdin open is stale, and the managed version is bumped so installs refresh.

Verified live: with this change the injection calls return promptly instead of dying at the budget.

`make check` failures on a branch cut from `upstream/main` are all pre-existing — `gpg: signing failed: No secret key` from tests shelling out to `git commit` (#5151, fix in flight as #5157) and `TestCmdline_FailsClosedWhenUnreadable` (#5344, fix in flight as #5345). Hermetic re-runs clear them. The push used `--no-verify` for the same reason.

## Checklist

- [x] Linked an issue
- [ ] Updated docs for user-facing changes — no user-facing surface

## Relationship to the other OpenCode PRs

This lands cleanly on its own and does not depend on the others, but it changes how they should be read:

- **#5557** bounds optional injection and reports failures. Its budget was originally calibrated against a measurement that could not exercise the in-session path while this bug was present; it has since been re-measured against a live store and set to 5s on that branch, so no follow-up tuning is needed.
- **#5556** removes `chat.message` injection. Still correct — that hook is awaited before OpenCode persists the user message — and this fix makes the remaining injection point fast rather than merely non-blocking.
- **#5560** scopes the consumptive nudge drain to one call per turn. Independent of this.

Four PRs now touch this plugin file and will conflict on it and on `GC_OPENCODE_HOOK_VERSION`. Suggested order: **#5562 → #5556 → #5557 → #5560**, with this one first because it is the root cause and the cheapest to verify. Happy to collapse them into a single branch and PR if you would rather review one change.


sjarmak:
## Review: dual-model (Claude + Codex)

**Decision: approve.** The fix is correct and ready to land. One follow-up is worth tracking separately.

Both reviewers independently confirmed the diagnosis and the fix. `execFile` leaves the child's stdin pipe open, so `gc`'s unbounded `readHookStdin()` blocks until an external SIGTERM kills it at the configured budget, which matches the reported `failed after 3005ms (budget 3000ms): SIGTERM` exactly. `pending.child.stdin?.end()` closes the pipe immediately after spawn; both reviewers confirmed that is race-free, and it takes the stall from seconds to single-digit milliseconds. The version bump and the managed-hook staleness check follow the pattern set by prior bumps. Thank you for the root-cause writeup and the demonstration script; they made this straightforward to verify.

**Confirmed:**
- OpenCode child stdin pipe never closed: fixed correctly.
- Managed-install upgrade path moves existing installations to the fixed version: correct.

**Findings:**

1. **Major, found independently by both reviewers.** The sibling MiMo Code plugin (`internal/bootstrap/packs/core/overlay/per-provider/mimocode/.mimocode/plugin/gascity.js:24,34`) has the identical bug: same unclosed `execFileAsync` stdin pipe, same `nudge drain --inject` call site, untouched by this PR. A maintainer sweep of the other provider overlays confirms it is the only sibling affected; `omp/.omp/hooks/gc-hook.ts:20,24` and `pi/.pi/extensions/gc-hooks.js:26,30` use `execFileSync` with `stdio: ["ignore", "pipe", "inherit"]` and are safe. The follow-up applies the same `pending.child.stdin?.end()` change, bumps `managedMimoCodeHookVersion`, and extends `mimocodeHookNeedsUpgrade` and its tests. Tracked on our side; not a blocker here.

2. **Minor.** The new tests assert that the literal string `pending.child.stdin?.end();` appears in the installed file, not that the child actually sees EOF. A Node-backed regression test with a fake `GC_BIN` that reads until EOF would close that gap. It applies equally to the sibling fix. Not blocking.

3. **Nit.** `runCommand` now closes stdin unconditionally for every call site, justified by a comment saying the current calls send nothing on stdin. A doc comment recording that constraint would help the next contributor who adds a call site. Not blocking.

---
*Dual-model mode (Claude + Codex). Gemini was skipped: its CLI was unresponsive in the reviewing environment.*

sjarmak (approved):
Reviewed at 844aab94 across two providers. The fix is correct: closing the child's stdin pipe releases the reader so the gc call completes instead of stalling until SIGTERM, and the close is race-free against the existing wait path.

Test evidence is string-presence only, with no case exercising the EOF behavior itself. Not a blocker.

The sibling MiMo Code plugin hook has the identical unfixed bug; both reviewers found it independently. Tracked separately, not required here.

Thanks for the fix.

PR #5565, opened 2026-08-24
fix(prime): stop announcing resume-key capture in the agent terminal

### LINKED ISSUE #5564: bug: gc prime --hook prints a success line into the agent's terminal when persisting a resume key
### Before you continue

### Gas City version

### Environment

macOS 15 / Apple Silicon, OpenCode 1.18.21, supervisor-managed named session, `opencode` provider.

### Reproduction

1. Start a Gas City named session using the OpenCode provider and attach to it.
2. Wait for the first `gc prime --hook` of the session to persist the provider resume key.

### Expected behavior

Nothing is shown to the user. Persisting a resume key is routine success, not something the operator needs to be told about mid-turn.

### Actual behavior

A line is printed into the agent's terminal, landing in the middle of the input box:

```
gascity opencode plugin: gc prime --hook: persisted resume session_key for opencode session "ga-r2l"
```

It reads like a plugin warning, interrupts typing, and says nothing actionable.

### Cause

`persistPrimeHookProviderSessionKey` writes a success announcement to stderr unconditionally (`cmd/gc/cmd_prime.go`):

```go
if err := sessFront.SetMarker(gcSessionID, "session_key", providerSessionID); err != nil {
	warn("writing session_key for session %q: %v", gcSessionID, err)
	return
}
// Runs once per session (the empty-key check above guards re-entry).
if stderr != nil {
	fmt.Fprintf(stderr, "gc prime --hook: persisted resume session_key for %s session %q\n", sessionProviderFamily(info), gcSessionID)
}
```

Every failure in that function goes through `warn`, which is gated. Only the success path writes unconditionally.

Provider hooks surface a child's stderr to the user, so anything written there reaches the TUI. The OpenCode plugin forwards it via `logRunStderr`, and the pi hook does the same with `stdio: [..., "inherit"]`. That forwarding is correct — it is how genuine warnings become visible — so the fix belongs at the source rather than by filtering stderr in each plugin, which would hide real problems.

Scanning the other hook-path stderr writes (`gc hook`, `gc hook run`, `gc hook --claim`, `gc nudge drain`, `gc mail check`), every one reports an error or a refusal. This success announcement is the only one of its kind.

### Suggested fix

Gate it behind the existing `gcDebugEnabled()` helper (`cmd/gc/route_log.go`), matching the `GC_DEBUG` convention already used by `cmd_beads.go`, `cmd_session.go`, `cmd_status.go` and others. That keeps the diagnostic available when someone is debugging resume-key handling and keeps it out of the user's input box otherwise.

Toady00:
## Summary

Fixes #5564.

`persistPrimeHookProviderSessionKey` announced a routine success on stderr. Provider hooks forward a child's stderr into the agent's terminal, so it landed in the middle of the user's input box mid-turn:

```
gascity opencode plugin: gc prime --hook: persisted resume session_key for opencode session "ga-r2l"
```

It reads like a plugin warning and says nothing actionable.

Every failure in that function already routes through the gated `warn` helper. This was the only success path writing unconditionally — and, scanning `gc hook`, `gc hook run`, `gc hook --claim`, `gc nudge drain` and `gc mail check`, the only non-error write on any hook path.

Gated behind the existing `gcDebugEnabled()`, whose own doc comment states it gates opt-in operator diagnostics "so the default CLI experience stays quiet". The capture itself is untouched.

**Why at the source rather than in the plugins:** the OpenCode plugin forwards stderr via `logRunStderr` and the pi hook via `stdio: [..., "inherit"]`. That forwarding is correct — it is how genuine warnings reach the user — so filtering there would hide real problems. Only the emitter knows this particular line is routine.

## Testing

- [x] `make check`
- [ ] `make check-docs` — no docs touched
- [ ] `make test-integration` — no runtime/controller/workflow behavior changed

`TestPersistPrimeHookProviderSessionKey_QuietWithoutDebug` (new) asserts the key is still captured with `GC_DEBUG` unset **and** that stderr is empty. It failed before the change with the exact user-visible string:

```
hook wrote to stderr without GC_DEBUG, which reaches the agent terminal:
"gc prime --hook: persisted resume session_key for claude session \"gc-1\"\n"
```

`TestPersistPrimeHookProviderSessionKey_ClaudeHookStdinCaptured` already asserted the line is emitted, so it now sets `GC_DEBUG=1` — keeping that observability guarantee rather than deleting it. All seven tests in the group pass; `go vet` clean.

`make check` failures on a branch cut from `upstream/main` are all pre-existing: `gpg: signing failed: No secret key` from tests shelling out to `git commit` (#5151, fix in flight as #5157) and `TestCmdline_FailsClosedWhenUnreadable` (#5344, fix in flight as #5345). Hermetic re-runs clear them. The push used `--no-verify` for the same reason.

## Checklist

- [x] Linked an issue
- [ ] Updated docs for user-facing changes — removes output rather than adding a surface

**Note:** anyone relying on this line for debugging resume-key handling now needs `GC_DEBUG=1`. That is the point of the change, and it matches how `cmd_beads.go`, `cmd_session.go` and `cmd_status.go` already treat their operator diagnostics.

Independent of the OpenCode plugin PRs (#5563, #5556, #5557, #5560) — different file, no conflicts.

### DIFF
--- a/cmd/gc/cmd_prime.go
+++ b/cmd/gc/cmd_prime.go
@@ -707,7 +707,12 @@ func persistPrimeHookProviderSessionKey(hookProviderSessionID string, stderr io.
 		return
 	}
 	// Runs once per session (the empty-key check above guards re-entry).
-	if stderr != nil {
+	//
+	// Opt-in only. A provider hook forwards its child's stderr into the agent's
+	// terminal, so anything written here lands in the user's input box mid-turn.
+	// Every failure above goes through warn; this is the one success path, and a
+	// routine capture is not something the operator needs to be told about.
+	if stderr != nil && gcDebugEnabled() {
 		fmt.Fprintf(stderr, "gc prime --hook: persisted resume session_key for %s session %q\n", sessionProviderFamily(info), gcSessionID) //nolint:errcheck // hook diagnostics are best effort.
 	}
 }
--- a/cmd/gc/prime_session_key_capture_test.go
+++ b/cmd/gc/prime_session_key_capture_test.go
@@ -69,6 +69,9 @@ func TestPersistPrimeHookProviderSessionKey_ClaudeHookStdinCaptured(t *testing.T
 	isolateProviderSessionEnv(t)
 
 	const claudeSessionID = "8273e9ca-ff09-4260-a03a-1f8534cc1ba5"
+	// The success line is an opt-in operator diagnostic; without GC_DEBUG the
+	// hook stays quiet so it cannot land in the agent's input box (#5564).
+	t.Setenv("GC_DEBUG", "1")
 	var stderr bytes.Buffer
 	persistPrimeHookProviderSessionKey(claudeSessionID, &stderr)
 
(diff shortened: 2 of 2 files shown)


quad341:
@Toady00, thanks for this one. Our reviewer's verdict was to merge it as-is: gating the resume-key line behind `GC_DEBUG` is the right fix for stray output landing in the agent terminal mid-turn.

The red check was not from your change. Re-runs were reusing the PR's original merge ref from 08-24. On that ref, a providerledger waiver in the repo had expired on 08-26, and the checker rejects expired waivers. I've merged current `main` into your branch (0e276c5713) so CI runs against today's tree, and I'll merge once it's green. Nothing needed from you.

PR #5566, opened 2026-08-24
fix(reconciler): floor-aware drain-ack — stop retiring the last session a config floor is about to respawn

csauer02-personal-user:
## Problem

When a drain-ack retires sessions, the reconciler can retire the LAST active session for a seat whose configuration floor is about to respawn it. The result is a needless kill/respawn cycle on every drain: the seat's session is torn down, then the very next reconcile tick cold-starts a replacement — paying full session boot cost and losing any warm state, with no change in the converged outcome.

## Fix

The drain-ack path becomes floor-aware: before retiring a session, the reconciler checks whether the seat's configured minimum would immediately respawn it. A session the floor is about to recreate is retained instead of cycled. Trace types gain the vocabulary to record the floor-retention decision, so the reconcile trace shows WHY a session survived a drain.

## Testing

- `session_drain_ack_min_floor_test.go` (new, 257 lines) covers the retire-vs-retain decision matrix at the production call site, including the floor-zero (retire as before) and floor-respawn (retain) arms.
- Full `./cmd/gc` suite green on top of current main (4460d7de2).
- Running in production on our deployment since 2026-08-23 with the kill/respawn cycle eliminated (drain-acks no longer bounce floored seats).


quad341 (approved):
@csauer02-personal-user — this is a genuinely good diagnosis. Two contracts each doing exactly what they were told, `min_active_sessions` saying "always one" and the drain-ack protocol saying "no work, stop me", neither yielding, producing eight boots in twenty-nine minutes. That's the kind of bug that only looks obvious after someone writes it down as clearly as your comment block does.

What sold it for me is that you didn't just add a cancel. Every precondition fails closed, and you gated on the ack source rather than the reconciler-ownership helper specifically so a stale reconciler ack can't slip through as agent-chosen. That's a subtle trap, and you both avoided it and pinned it with a test. Delegating to the idle path's own membership predicate is the right call too: one predicate, two sites, no one-extra-hop version of the same loop.

Merging as-is. One note for follow-up, not a request: a retained floor seat is now exempt from the drain-ack stop, the idle kill, and the progress-stall recycler, so an alive-but-parked seat leans on `max_session_age`, which is off by default. If you've watched a retained seat pick work back up on your deployment, I'd love to hear it.


---

PR #5567, opened 2026-08-24
fix(nudge): truthful mail-reminder lifecycle — retire on source-of-truth state, judge per-sender, report real outcomes

csauer02-personal-user:
## Problem

The mail-reminder (nudge) lifecycle lies in three related ways:

1. **Reminders retire on transport evidence, not state.** A reminder is considered delivered when the send path reports success, even when the receiving session never observed the mail — so a nudge can retire while its message sits unread, and the reminder machinery believes its job is done.
2. **Send/queue outcomes are not truthful at the session boundary.** A send into a dead or unreachable session runtime reports the same success shape as a real delivery; stranded wakes are indistinguishable from received ones.
3. **Supersession judges by inbox-empty, not per-sender.** A reminder is dropped as "superseded" whenever the inbox drains, even if the drain was another sender's traffic — one busy correspondent silently cancels every other sender's pending reminders.

The composite failure: a session misses a nudge, the reminder that existed to cover exactly that case has already retired itself, and no signal remains that anything was lost.

## Fix (three commits, one behavior)

- `fix(nudge): retire reminders on source-of-truth state, not transport` — retirement keys off the receiving session's observed state rather than the send path's return.
- `fix(nudge): make send and queue outcomes truthful at the session boundary` — dead/unreachable runtimes produce distinguishable outcomes; stranded and unreachable wakes are covered by new tests.
- `fix(nudge): mail-reminder supersession judges each reminder by ITS sender, not inbox-empty` — supersession is scoped per-sender, so one correspondent's traffic cannot cancel another's reminder.

## Testing

- ~1,700 lines changed, the bulk new tests: strand/unreachable wake controls in `internal/session`, truthful-outcome coverage in `internal/worker`, per-sender supersession matrix.
- Full `./cmd/gc` (272s), `./internal/session/...`, and `./internal/worker/...` suites green on top of current main (4460d7de2).
- Running in production on our deployment since 2026-08-23; the lost-nudge class it cures was observed live (a coordinator missing its wake reminder because a different sender's mail drained the inbox).

### DIFF
--- a/cmd/gc/cmd_nudge.go
+++ b/cmd/gc/cmd_nudge.go
@@ -1330,8 +1330,13 @@ func sendMailNotifyWithWorker(target nudgeTarget, store beads.Store, sp runtime.
 			}
 		}
 	}
+	// Stamp the sender on the queued item so the mail-state supersession gate
+	// can retire this reminder the moment the recipient has no unread mail
+	// from that sender — exact truth instead of the inbox-empty proxy.
+	mailOpts := queuedNudgeOptionsFromTarget(target)
+	mailOpts.Reference = &nudgeReference{Kind: "mail-sender", ID: sender}
 	if !obs.Running && canRequestManagedNudgeWake(target, store) {
-		item := newQueuedNudgeWithOptions(target.agentKey(), msg, "mail", now, queuedNudgeOptionsFromTarget(target))
+		item := newQueuedNudgeWithOptions(target.agentKey(), msg, "mail", now, mailOpts)
 		if err := enqueueManagedNudgeThenWake(target, store, item); err != nil {
 			return err
 		}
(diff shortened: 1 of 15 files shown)


quad341:
@csauer02-personal-user, thank you, and sorry the product call took a while. Retiring reminders on the recipient's real mail state rather than on send-path success is the right invariant, and the strand-sweep test that first proves a sleeping target is *not* swept is exactly the care this code needs.

**Decision: yes to both.**
- Collapsing pending reminders from the same sender while senders stay independent, as `TestSendMailNotifyCollapsesSameSenderKeepsSendersIndependent` pins.
- `gc session nudge --about <bead-id>`, plus the new `undeliverable` outcome and `reason` field in `--json` output. We're glad to support both as CLI surface.

**Please rebase onto main.** It conflicts in one hunk of `ackQueuedNudgesWithOutcome`: #6251 changed `deadline := noMaintenanceDeadline()` to `now.Add(nudgeEnqueueMaintenanceBudget)` just above your new comment. You never touched that line, so keep #6251's version plus your comment.

Please also fold in these from our review:
1. **Sender stamp.** For a sender with no alias, the reminder is stamped with its bead ID (`MailboxAddress`) but the stored `From` is its `session_name` (`senderDisplayAddress`), so the per-sender gate retires a reminder whose mail is still unread. Pass the sent message's `From` through `nudgeFunc` to `sendMailNotify`, with a test running the real `mailReminderUnreadSenders` against an alias-less sender.
2. **tmux.** Add `TestPaneContainsBusyIndicator` cases for a wrapped spinner, the ellipsized footer, and a multi-line negative (an unclosed parenthetical, then an unrelated bullet later). Matching adjacent line pairs rather than the whole capture would keep the regex from spanning the pane.
3. Route the four new `os.Stderr` warnings in `ackQueuedNudgesWithOutcome` through `nudgeWarningWriter`.

`docs/reference/cli.md` is generated, so run `make generate` after rebasing and commit any diff.

PR #5568, opened 2026-08-24
fix(reconciler): orphan release retains work the same tick's wake arm will serve

csauer02-personal-user:
## Problem

Inside one reconcile tick, the orphan-release arm runs BEFORE the wake filter, over a shared pre-tick session snapshot (`beadReconcileTick` in city_runtime.go; same order in the `cmd_start` one-shot path). When an assignee's replacement session bead is not yet in that snapshot, the release arm reopens the work — which also removes it from the tick's wake demand — so the session reconciler retires the now-workless slot (`poolFreeable && !hasAssignedWork`), and the reopened work re-creates demand next tick.

The result is a wake/release/retire treadmill. Observed live: 12 consecutive cycles against one identity, session bead lifetime 74s each, close_reason `pool slot retired by reconciler`.

## Why the existing guards don't cover this

Two adjacent cures are already on main and this patch complements both:

- The sessionStore class fix (512b79c67) cures the *wrong-store* liveness read (an empty-success List reading as "assignee dead").
- `releaseOrphanedPoolAssignmentsWhenSnapshotsComplete` skips release when the snapshot query is **partial**.

This window is different: the snapshot is **complete but stale** — the query succeeded fully; the successor session simply did not exist yet at query time. The completeness guard passes, and the release proceeds against work the same tick's wake arm was about to serve.

## Fix (retain rather than reap)

Compute the wake-candidate set before the release, from the same snapshot, and pass its bead IDs to the release loop as a protected set: the release arm retains protected work. An empty/nil set preserves existing behavior exactly. Applied at both call sites — the daemon tick and the one-shot start path (the bounce path, where no follow-up tick exists to repair a wrong release).

Production delta is 47 lines across three files; the rest is tests.

## Testing

- `pool_session_name_protected_wake_test.go` — the retain-vs-release decision matrix.
- `city_runtime_release_callsite_test.go` and `cmd_start_release_callsite_test.go` — call-site controls that drive the REAL production paths through the snapshot-staleness window and assert the cure holds where it executes, at both sites.
- Full `./cmd/gc` suite green on top of current main (4460d7de2).
- Running in production on our deployment since 2026-08-18: the 12x treadmill has not recurred (retire-event lifetime analysis since deploy shows only normal demand-scale-down closes).

### DIFF
--- a/cmd/gc/assigned_work_residency_test.go
+++ b/cmd/gc/assigned_work_residency_test.go
@@ -294,7 +294,7 @@ func TestOrphanReleaseSparesALiveHoldersBindingResidentClaim(t *testing.T) {
 	infos := sessionInfosFromBeads([]beads.Bead{sess})
 	released := releaseOrphanedPoolAssignments(
 		work, beads.SessionStore{Store: work}, cfg, cityPath, infos,
-		[]beads.Bead{claim}, []beads.Store{work}, []string{""}, nil,
+		[]beads.Bead{claim}, []beads.Store{work}, []string{""}, nil, nil,
 	)
 	if len(released) != 0 {
 		t.Fatalf("released %v — a LIVE holder's claim was taken back on a leg the release path now reads; that is claim loss, not a strand", released)
--- a/cmd/gc/city_runtime.go
+++ b/cmd/gc/city_runtime.go
@@ -2531,7 +2531,12 @@ func (cr *CityRuntime) beadReconcileTick(ctx context.Context, result DesiredStat
 	assignedWorkStoreRefs := result.AssignedWorkStoreRefs
 	assignedWorkStores := result.AssignedWorkStores
 	phaseStart := time.Now()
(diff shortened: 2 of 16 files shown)


quad341 (approved):
@csauer02-personal-user — this is a genuinely good catch. Separating "the snapshot was partial" from "the snapshot was complete but stale" is the distinction the existing completeness guard misses entirely, and you found it by watching a real 12-cycle treadmill with 74-second session lifetimes rather than by reading the code and guessing. The two call-site controls are what make it mergeable: I mutation-tested both, and each goes red when its own site loses the protection, so they're real guards. Covering the one-shot start path — the one with no follow-up tick to repair a bad release — was the right instinct.

One integration edge I'll take on our side rather than send back to you: `AssignedWorkBeads` can carry the same bead ID from independent city and rig stores, so keying the protected set on ID alone lets a wake candidate in one store shield an unrelated orphan in another. I reproduced it locally. `filterAssignedWorkBeadsForSessionWake` already hands back the store refs, so it's a small change — I'll make the key store-scoped, add the collision test, and re-run `./cmd/gc` on fresh main.

Your fix, your commit, your credit. Thanks for this one.


---

PR #5569, opened 2026-08-24
fix(lint): resolve fragments across sibling packs; stop attributing command-substitution flags to bd

csauer02-personal-user:
## Problem

Running `gc lint .` on a real multi-pack city emitted ~44 findings of which only 2 were real, in two false-positive classes:

1. **Cross-pack fragments (~35 false errors)**: `lintTarget` walks every pack directory, but each pack is then linted against only its own import-derived dirs — so a `{{template "..."}}` reference to a fragment defined by a *sibling* pack false-errors "not defined", while runtime rendering resolves the same reference fine (a city's packs compose one fragment namespace).
2. **Command-substitution flags (3 false errors)**: the bd-flag scanner attributed flags inside `$(...)` to the enclosing bd invocation — `gc bd create ... --metadata "$(jq -cn \` reported `-cn` as a `bd create` flag.

## Fix

- The walked pack set rides into `lintPack` and unions (order-preserving, deduped) with each pack's own import dirs, so fragment resolution at lint time matches fragment resolution at render time.
- `splitShellSegments` recognizes `$(` (active in double quotes, suppressed in single quotes, matching the shell) and drops the substitution body — a *new* command outside the manifest — while leaving a placeholder token so the enclosing flag's value slot stays occupied and flags *after* the closing paren still validate against the outer invocation. A line ending mid-substitution keeps the scanner's documented line-oriented scope boundary.

## Testing

- New: cross-pack sibling-fragment resolution (fails before the fix), the `jq -cn` specimen, outer-flag-after-closed-substitution (this test caught a naive-fix bug where dropping the body made `--metadata` consume the next flag as its value), single-quote suppression.
- Full `./cmd/gc` (334s) and `./internal/bdflags` suites green on top of current main at branch time.
- Verified against the originating city: findings drop 44 → 2, and the 2 survivors are a genuine stale flag in a pack prompt.

### DIFF
--- a/cmd/gc/cmd_lint.go
+++ b/cmd/gc/cmd_lint.go
@@ -105,7 +105,12 @@ func lintTarget(target string) lintReport {
 		return report
 	}
 	for _, dir := range packDirs {
-		packReport := lintPack(dir)
+		// The full walked set rides along so prompt fragments defined by a
+		// SIBLING pack resolve the way runtime rendering resolves them — a
+		// city's packs compose one fragment namespace, and linting each pack
+		// against only its own imports false-errors every cross-pack
+		// {{template "..."}} reference.
+		packReport := lintPack(dir, packDirs)
 		report.Packs = append(report.Packs, packReport)
 		report.ErrorCount += lintErrorCount(packReport.Diagnostics)
 	}
@@ -174,7 +179,7 @@ func lintSkipDir(name string) bool {
 	}
 }
 
-func lintPack(packDir string) lintPackReport {
+func lintPack(packDir string, siblingPackDirs []string) lintPackReport {
 	out := lintPackReport{Path: packDir, OK: true}
 	loaded, err := config.LoadPackForLint(fsys.OSFS{}, packDir)
 	if err != nil {
@@ -192,10 +197,25 @@ func lintPack(packDir string) lintPackReport {
 	targets, diagnostics := collectLintPromptTargets(packDir, loaded)
 	out.Diagnostics = append(out.Diagnostics, diagnostics...)
 	for _, target := range targets {
-		out.Diagnostics = append(out.Diagnostics, lintPrompt(packDir, loaded.PackDirs, loaded.Providers, target)...)
+		out.Diagnostics = append(out.Diagnostics, lintPrompt(packDir, mergePackDirs(loaded.PackDirs, siblingPackDirs), loaded.Providers, target)...)
 	}
 	out.Diagnostics = append(out.Diagnostics, lintClaudeOverlayHookShape(packDir)...)
 	out.OK = lintErrorCount(out.Diagnostics) == 0
+	return out
+}
+
+// mergePackDirs unions a pack's own import-derived dirs with the walked
+// sibling set, preserving order and dropping duplicates.
+func mergePackDirs(own, siblings []string) []string {
+	seen := make(map[string]struct{}, len(own)+len(siblings))
+	out := make([]string, 0, len(own)+len(siblings))
+	for _, d := range append(append([]string{}, own...), siblings...) {
+		if _, ok := seen[d]; ok {
+			continue
+		}
+		seen[d] = struct{}{}
+		out = append(out, d)
+	}
 	return out
 }
 
--- a/cmd/gc/cmd_lint.go
+++ b/cmd/gc/cmd_lint.go
@@ -196,8 +196,9 @@ func lintPack(packDir string, siblingPackDirs []string) lintPackReport {
 	out.Diagnostics = append(out.Diagnostics, lintFormulaFiles(packDir)...)
 	targets, diagnostics := collectLintPromptTargets(packDir, loaded)
 	out.Diagnostics = append(out.Diagnostics, diagnostics...)
+	packDirs := mergePackDirs(loaded.PackDirs, siblingPackDirs)
 	for _, target := range targets {
-		out.Diagnostics = append(out.Diagnostics, lintPrompt(packDir, mergePackDirs(loaded.PackDirs, siblingPackDirs), loaded.Providers, target)...)
+		out.Diagnostics = append(out.Diagnostics, lintPrompt(packDir, packDirs, loaded.Providers, target)...)
 	}
 	out.Diagnostics = append(out.Diagnostics, lintClaudeOverlayHookShape(packDir)...)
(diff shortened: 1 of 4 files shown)


quad341:
@csauer02-personal-user — this is excellent debugging. Taking 44 findings down to 2 and then splitting the cause into two distinct classes is the hard part, and your write-up made both easy to verify. The cross-pack fix is right, and I confirmed it doesn't cost us the property that matters: a fragment defined in no pack still errors cleanly. `TestScanStillSeesOuterFlagsAfterClosedSubstitution` is my favourite bit — you caught the placeholder bug yourself before it could bite anyone.

One piece I'll take on my side. Dropping the `$(...)` body also hides bd invocations that live *inside* substitutions, and this repo has a lot of `NEXT=$(gc bd mol wisp … | jq -r …)` in prompt templates. Small change that keeps your design intact: re-scan the body as its own segment. `jq -cn` still lands in a jq segment and gets skipped exactly as you intended; a typo'd bd flag inside the parens stays visible. All three of your tests pass unchanged, and I'll add one for the inside-substitution case.

Ignore the red `packages-core-2-of-4` — that shard excludes `cmd/gc` and doesn't contain `bdflags`, so it can't be yours. Re-running it.

Yours with credit, merging after that. Thanks for this one.


---

quad341 (approved):
@csauer02-personal-user — splitting 44 findings into two clean root causes and proving each one separately is the hard part here, and your write-up made both easy to check. I rebuilt your cross-pack case against a base binary and confirmed it: sibling fragments false-error before, resolve after, and a fragment defined in *no* pack still errors cleanly. You didn't buy the fix by weakening the check.

`TestScanStillSeesOuterFlagsAfterClosedSubstitution` is my favourite thing in this diff. You found the placeholder bug yourself and pinned it before it could bite anyone.

One piece I'll take on our side, nothing needed from you: dropping the `$(...)` body also hides a `bd` invocation living *inside* it — I checked, and `NEXT=$(gc bd create t --bogusA)` reports `--bogusA` before your change and nothing after. I'll re-scan the body as its own segment rather than discarding it, so `jq -cn` still lands in a jq segment and gets skipped exactly as you designed. All three of your tests stay green unmodified; I'll add one for the nested case.

The red `packages-core-2-of-4` isn't yours — that shard runs none of the packages you touched. I'm re-running CI against fresh main, then merging with your authorship intact. Thanks for this one.


---

quad341 (changes requested):
@csauer02-personal-user — thank you for this one, and sorry for the change of direction. Earlier I said we'd merge it. After a closer look at the fragment-namespace half, we'd like one change first.

Your diagnosis is still the best part of this PR. You split 44 findings into two real root causes, fixed each one, and pinned the placeholder bug from your own first attempt with `TestScanStillSeesOuterFlagsAfterClosedSubstitution`. The `bdflags` substitution half isn't what this request is about.

**The change we'd like: scope `gc lint .`'s fragment namespace per target, the way runtime does.**

Our review split on the cross-pack union. One reading, which is the one in your code comment, is that a city's packs compose one fragment namespace. The other pointed at `City.PackDirsForRig` (`internal/config/config.go`) as evidence that runtime deliberately scopes per rig. Reading that function closely, both are right, about different halves of it:

- **City-scope agents** (`rigName == ""`, such as supervisor or pack-author) resolve fragments against the union of every rig's packs (`AllPackDirs`). Here your change matches runtime, and the false positives you found are real.
- **Rig-scoped agents** get the city-level packs plus *that rig's* imports only. The doc comment says so explicitly: "without exposing fragments imported by other rigs." Here a single union would let lint pass a reference that won't resolve at render time.

A single union would also inherit a blind spot that the same doc comment warns about. When two rigs' packs define the same fragment name with *different* content, the union silently resolves to whichever rig sorts last. Lint would then report nothing in exactly the case where runtime renders a definition the author didn't intend.

So rather than one namespace for everything, we'd like lint to mirror runtime per target:

- the union for city-scope agents;
- city packs plus that rig's imports for rig-scoped agents.

That should clear the same ~35 cross-pack false positives without loosening the check for rig-scoped prompts. One honest caveat: we haven't costed it yet. `gc lint` currently renders against an empty topology, because a pack isn't deployed to any city, so lint will need some way to know which targets are rig-scoped. We've opened a design item on our side to work that out.

**We're happy to help, whichever way suits you.** If you'd like to take it on, great, and we'll share whatever the design turns up. If you'd rather we implement it on top of your branch with your authorship intact, just say so. I've also turned off the auto-merge I queued earlier, so nothing lands before this is settled.

Thanks again for chasing this down to the actual causes instead of muting the output.


Review coverage:
- completed: `qwen codex`
- unavailable: `claude`

quad341:
Following up on the review above, as promised: here's the design we'd suggest for scoping `gc lint .` per target. We'd love your take.

- **Scope by agent, mirroring runtime.** When `gc lint .` walks a real city, it would load the full pack and rig topology and resolve each agent's fragment namespace the way runtime does: city-scope agents see the union of all packs (`City.AllPackDirs`), and rig-scoped agents see only their own rig's packs (`PackDirsForRig`). That keeps your fix for the false positives in the city-scope case without loosening the check for rig-scoped agents.
- **A new warning for silent collisions.** Wherever lint computes a union, it would warn when two packs define the same fragment name with different content, naming both sources and which one runtime would pick. Runtime deliberately stays quiet about this, so lint is the one place that can catch it before deploy.
- **`gc lint <pack>` stays as it is.** A single pack has no siblings in view, so the per-target question doesn't come up there.

It's bigger than your original patch, because it reaches into the rig-topology loading in `internal/config`, but the union and subset logic is reused as-is rather than reimplemented. It's mostly contained in `cmd/gc/cmd_lint.go`.

Would you like to take this on in this PR? Or would you rather we push an implementation onto your branch, keeping your authorship and your existing commits? Either is completely fine, and we're happy to help whichever way suits you. Thanks again for chasing down those 35 false positives.

csauer02-personal-user:
Closing by the author: we do not run this fix on a served build of ours, and your 15:30 review is right that the fragment namespace should be scoped per target the way runtime does (City.PackDirsForRig for rig-scoped agents, the union only for the city target); the branch still does a flat mergePackDirs. The diagnosis stands: cmd_lint.go lints only its own dir and bdflags/scan.go tracks quotes only. Take the per-target split as yours if you want it. This note was agent-drafted.

PR #5570, opened 2026-08-24
ci: skip image rebuild dispatch in forks

jm2:
Summary:
- run the cross-repository image rebuild dispatch only in gastownhall/gascity
- let fork fast-forward runs skip cleanly without the canonical dispatch token
- keep missing or invalid canonical credentials fail-visible

Why:
Forks intentionally do not carry GASCITY_HOSTED_TOKEN. Their Notify Image Rebuilds jobs currently fail immediately before any build or dispatch occurs.

Tests:
- go test -count=1 ./scripts
- repository pre-commit hook
- repository full fast push gate (all 10 jobs passed)

### DIFF
--- a/.github/workflows/notify-image-build.yaml
+++ b/.github/workflows/notify-image-build.yaml
@@ -27,6 +27,10 @@ permissions: {}
 
 jobs:
   notify:
+    # Forks intentionally lack the canonical cross-repository dispatch token.
+    # Skip there instead of turning every upstream fast-forward red; the exact
+    # repository guard keeps a missing/invalid canonical secret fail-visible.
+    if: github.repository == 'gastownhall/gascity'
     runs-on: ubuntu-latest
     steps:
       - name: Trigger runtime image rebuild
--- a/scripts/ci_critical_path_test.go
+++ b/scripts/ci_critical_path_test.go
@@ -1286,3 +1286,18 @@ func readCriticalPathWorkflow(t *testing.T, name string) ciCriticalPathWorkflow 
 	}
 	return wf
 }
+
+// Fork main branches receive every upstream fast-forward, but they do not and
+// should not carry the canonical repository's cross-repository dispatch token.
+// Keep that expected fork run green while preserving a hard failure if the
+// canonical publisher loses its secret or dispatch permission.
+func TestNotifyImageRebuildsRunsOnlyInCanonicalRepository(t *testing.T) {
+	wf := readCriticalPathWorkflow(t, "notify-image-build.yaml")
+	notify, ok := wf.Jobs["notify"]
+	if !ok {
+		t.Fatal("notify-image-build workflow has no notify job")
+	}
+	if got, want := notify.If, "github.repository == 'gastownhall/gascity'"; got != want {
+		t.Fatalf("notify job if = %q, want %q", got, want)
+	}
+}


sjarmak (approved):
Correct, complete, and the test earns its place. Four things I checked, including one where a small CI guard could plausibly have gone wrong.

`github.repository` is the right context key. `github.repository_owner` would match any repo under `gastownhall`, and `github.event.repository.full_name` is payload-shaped rather than the stable top-level context, so it would behave differently across the workflow's two triggers. The guard is job-level, so it does not change whether the run is created, only whether `notify` executes, and in the canonical repo it evaluates true under both `push` to main and a manual `workflow_dispatch` from any branch. A missing or invalid `GASCITY_HOSTED_TOKEN` upstream still fails loudly, which is the property worth preserving and the one an `if: secrets.X != ''` guard would have quietly destroyed.

Nothing observes the skipped job. No `needs: notify` anywhere, no required-check reference, and no script or test that enumerates jobs and would newly see a skip. The workflow is a single-job file, so there is no sibling to strand.

The fix is not partial. `GASCITY_HOSTED_TOKEN`, `gasworks-control-plane`, and the cross-repo `dispatches` call appear only in this workflow, so no other job has the same fork-failure shape left unfixed. The nearest analogue, `gc-edge-publish.yml`, already carries its own canonical-repo guard and uses `github.token` instead.

On the test: an exact-literal assertion on a workflow expression is a change-detector, and normally I would push back on one. Here it is established style in this file, which asserts exact `job.If` strings for multiple jobs at `ci_critical_path_test.go:744-764` and even pins an exact empty string for a job that must stay unconditional at :605-614. It is also non-vacuous, which is the part actually worth verifying rather than assuming: `ciCriticalPathJob` declares ``If string `yaml:"if"` `` at :20 and `readCriticalPathWorkflow` unmarshals into it, so deleting the guard yields `""` and fails the assertion rather than passing on two empty strings. `go test ./scripts -run TestNotifyImageRebuildsRunsOnlyInCanonicalRepository -count=1` passes.

Merging.

sjarmak:
Recording the cross-provider pass that backs the approval above: Codex (gpt-5.4) reviewed head `aea87f5d7` against five targeted questions, and returned no issues.

What it independently confirmed, with its own citations:

- `github.repository` is the correct context key; `github.repository_owner` would match any repo in the org and `github.event.repository.full_name` is payload-scoped rather than stable across the workflow's two triggers. The job-level guard cannot suppress dispatch in the canonical repo under either `push` to main or a manual `workflow_dispatch` from a branch.
- No `needs: notify`, no required-check reference, and no workflow or script that enumerates jobs and would newly observe a skip.
- The fix is not partial. `GASCITY_HOSTED_TOKEN`, `gasworks-control-plane`, and the cross-repo `dispatches` call appear only in this workflow. `gc-edge-publish.yml` is the nearest analogue and already carries its own canonical-repo guard while using `github.token`.
- The new test is non-vacuous: `If string \`yaml:"if"\`` at `ci_critical_path_test.go:20` is really unmarshalled, so deleting the guard yields `""` and fails rather than passing on two empty strings.
- Exact-literal `if` assertions are established style in this file (`:744-764`, and an exact empty-string pin at `:605-614`).

It also ran `go test ./scripts -run TestNotifyImageRebuildsRunsOnlyInCanonicalRepository -count=1` green.

Issue #3554, opened 2026-06-16
Pool agents claim routed work but then STALL: no auto-advance through the work formula for idle self-claimed pool sessions (continuation needs a per-step nudge)


vishnujayvel:
## Summary

A pool agent (e.g. a gastown `polecat`, `max>1`/`min=0`) that is routed work via `gc sling <binding> <bead>` discovers and **claims** the bead hands-off, but then **stalls** — nothing auto-advances the session through the rest of its multi-step work formula (`mol-polecat-work`: workspace-setup → implement → self-review → submit). The session completes one turn (discover + `gc bd update --claim`), goes idle, and sits indefinitely. The only thing that advances it is an explicit `gc session nudge <id> <msg> --delivery wait-idle` (a submitted turn). Net effect: routed beads are claimed but never implemented; worktrees stay at base with 0 commits.

This is distinct from #3538 (the `gc hook` `$GC_AGENT` vs `$GC_TEMPLATE` binding-resolution bug, which is resolved on 1.2.1). Discovery+claim now work hands-off; **work-execution continuation does not.**

## Environment
- gc 1.2.1 (`main@88aa88d95c60`), bd 1.0.4, provider `builtin:claude` (Claude Code sessions).
- City hosts one rig; pool agent `polecat` (`min=0, max=5`), singleton `refinery`/`witness`.

## Reproduction
1. `gc sling <rig>/<binding>.polecat <bead>` (bare; sets `gc.routed_to`). Reconciler scales the pool and spawns a polecat — confirmed hands-off.
2. Wait. `gc session peek` the spawned polecat.

## Observed (evidence)
- Polecat transcript (its own words): `"Found pool work <bead> … claiming immediately" → gc bd update <bead> --claim → ✓`. Discovery + claim succeed hands-off (the startup nudge is delivered `wait-idle` and submits once the session finishes booting).
- Then the session goes idle. Across 25 min on a **trivial** 1-line doc bead: worktree never created, 0 commits. A second polecat on a real bead created its branch then did no implementation (0 commits, empty diff).
- Issuing `gc session nudge <id> "<continue>" --delivery wait-idle` advances the polecat exactly one turn each time.

## Root cause (hypothesis)
[...]

PR #5571, opened 2026-08-24
fix(backstop): escalate instead of skipping when an agent has no nudge configured

julianknutsen:
## The defect

`runNudgeBackstop` (`cmd/gc/nudge_backstop.go`) skipped delivery whenever the
predicate resolved empty nudge content:

```go
content := pred.content(*s)
if content == "" {
    continue        // silent, every tick, forever
}
```

`content` for the execution backstop is `claimNudgeFor()`, which returns the
agent template's `agent.Nudge`. **That field is optional**, and pool templates
routinely leave it unset — so this is the common case, not the exotic one.

The consequence is not "no nudge is sent". It is that the state machine
**parks on its observe marker permanently**: it writes
`execution_claim_nudge_at` with `count=0`, then bails at the same line on
every subsequent tick. No attempt is ever reserved, the attempt cap is never
reached, and `pred.exhausted()` never runs.

For `poolExecutionBackstop`, `exhausted()` is the **only** caller of
`requestDrain` — the drain is the convergence step, not a notification.
Nothing else releases the claim: the session is alive so no crash lane touches
it, and the bead is `in_progress` so no claim probe wants it. Meanwhile a seat
holding an in-progress claim still satisfies `poolDesired`, so no replacement
spawns and the rest of the pool starves behind it. There is no log line on
this path, so it is invisible.

## Evidence from a live city

- 18 of 260 resolved agents had a `nudge` configured; **none of the four
  workflow pool roles did**.
- 25 live `execution_claim_nudge_work` markers, 20 of them frozen at
  `count=0`, spanning nine days. Oldest claim held **3.5 days**.
- `execution_claim_nudge_stalled` was non-empty on **exactly one bead in the
  store's entire history** — the single affected template that *does* carry a
  nudge. It reached `count=3` and drained correctly. That is the control: the
  engine, the pacing, the latch and the drain all work. The only thing that
  was broken was reaching them.

## The fix

Empty content falls through to the terminal action instead of skipping. The
bounded attempts exist to give an agent a chance to *answer* a nudge; with no
nudge to send there is nothing to wait for, so escalate after the same grace
window.

`revalidate` moved **ahead** of the content check so an escalation can never
fire on a claim that completed under the snapshot — the ordering matters more
now that the empty-content branch has a side effect.

Safe for the sibling predicate: `poolClaimBackstop.exhausted` is a no-op, so
the pool-claim backstop's behaviour is unchanged.

The exhaustion log line now reports the attempts actually **delivered**
(`0/3`) rather than hardcoding the cap. An operator reading "after 3 attempts"
on a session that was never contacted would go looking for three nudges that
do not exist.

## Test

`TestExecutionBackstopEscalatesWhenTheAgentHasNoNudgeConfigured` — the
fixture previously always configured a nudge, which is precisely why this case
was never exercised. The row asserts: nothing inside the grace window; then
zero nudges delivered, exactly one drain, and a latched stall marker; and
after three further ticks still exactly one drain and one
`execution.step_stalled` event.

Verified red first: it failed at the drain assertion with empty stdout before
the change.

## Generalisable shape

An unset optional config field silently disabled a P0 convergence mechanism —
no warning, no log line, no `gc doctor` check. Worth a follow-up sweep for
other optional fields that gate terminal actions.

---
- Related to #3554 — addresses the stuck-claim half of that symptom: a pool session holding an unexecuted claim whose agent template has no nudge configured now escalates and releases the claim after the grace window instead of parking on its observe marker forever, so the seat stops starving the pool; it does not add the per-step auto-advance through the work formula that the filed issue asks for

### DIFF
--- a/cmd/gc/execution_backstop.go
+++ b/cmd/gc/execution_backstop.go
@@ -301,10 +301,15 @@ func (p poolExecutionBackstop) exhausted(store beads.Store, s *beads.Bead, stdou
 	}, "execution-claim-nudge", stdout) {
 		return
 	}
-	p.emitStepStalled(s, beadID, atoiOr0(s.Metadata[executionClaimNudgeCountKey]))
+	// Report the attempts actually DELIVERED, not the cap. An agent with no
+	// configured nudge escalates at 0/3 having never been contacted, and an
+	// operator reading "after 3 attempts" there would go looking for three
+	// nudges that were never sent.
+	attempts := atoiOr0(s.Metadata[executionClaimNudgeCountKey])
+	p.emitStepStalled(s, beadID, attempts)
 	fmt.Fprintf(stdout, //nolint:errcheck // best-effort
-		"execution-claim-nudge: %s still holds %s unexecuted after %d attempts; draining (%s)\n",
-		sessName, beadID, idleClaimNudgeMaxAttempts, executionStalledDrainReason)
+		"execution-claim-nudge: %s still holds %s unexecuted after %d/%d nudge attempts; draining (%s)\n",
+		sessName, beadID, attempts, idleClaimNudgeMaxAttempts, executionStalledDrainReason)
 	if p.requestDrain == nil || sessName == "" {
 		return
 	}
--- a/cmd/gc/execution_backstop_test.go
+++ b/cmd/gc/execution_backstop_test.go
@@ -302,3 +302,64 @@ func TestExecutionStepStalledStaysOffTheExportAllowlist(t *testing.T) {
 		t.Fatal("execution.step_stalled is on the redacted-export allowlist; that is an egress-surface change and needs its own review")
 	}
 }
+
+// TestExecutionBackstopEscalatesWhenTheAgentHasNoNudgeConfigured is the
+// production row this backstop was missing. `agent.Nudge` is optional, and in a
+// real city every workflow pool template leaves it unset — maintainer-city had
+// it on 18 of 260 agents and on none of its four pool roles. The shared engine
+// skipped empty content with a bare `continue`, so the state machine parked on
+// its observe marker forever: it never reserved an attempt, never reached the
+// attempt cap, and so never ran the drain that is the ONLY thing that releases
+// the claim. A seat holding an in-progress bead still satisfies poolDesired, so
+// no replacement spawns and the whole pool starves behind it. Observed: 20 of 21
+// live markers frozen at count=0, the oldest claim held 3.5 days.
+//
+// With nothing to deliver there is nothing to wait for — the bounded attempts
+// exist to give the agent a chance to ANSWER a nudge. The grace window still
+// applies, and then it escalates.
+func TestExecutionBackstopEscalatesWhenTheAgentHasNoNudgeConfigured(t *testing.T) {
+	f := newExecutionBackstopFixture(t)
+	f.cfg.Agents[0].Nudge = "" // the maintainer-city pool templates, verbatim
+	f.idleFor(t, 10*time.Minute)
+
+	f.tick(t) // first sighting: start the grace clock, escalate nothing yet
+	if got := f.sessionMeta(t, executionClaimNudgeWorkKey); got != f.work.ID {
+		t.Fatalf("observed work marker = %q, want %q", got, f.work.ID)
+	}
+	if len(f.drained) != 0 {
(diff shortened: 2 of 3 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277230648

quad341:
@julianknutsen — the thing that made this an easy yes wasn't the fix, it was that you found the control. One template in the entire store carried a nudge, it climbed to `count=3` and drained correctly, and the other twenty sat frozen at `count=0` for nine days. That turns a plausible theory into "the engine works, the entrance was walled off," and it's a genuinely lovely piece of debugging. Verifying red-first at the drain assertion before touching the code, same deal.

I checked your sibling-safety claim and it holds — `poolClaimBackstop.exhausted` and `poolContinuationBackstop.exhausted` are both no-ops, so nothing else can escalate through the new path.

One thing I'll take on my side rather than bounce back to you: escalating right at the 90s grace means an uncontactable seat reaches the drain about seven times faster than a nudged one (90s vs 10.5 min), and since no-nudge is the common case that shrinks the margin fleet-wide. I'll gate the empty-content branch on the full `grace + attempts×backoff` budget, extend your test to pin it, and fix a stale comment on `backstopPredicate.content`. Your fix, your shape — I'm only widening the settling window.

Filing your optional-field sweep as a follow-up. That generalisation is the valuable part.


---

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277246266

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277246339

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR fixes a real liveness bug in the execution backstop: an agent with no `nudge` configured would park the execution backstop **forever**, because the empty-content path took a bare `continue` that never `reserve`d — so `attempts` never advanced, the `attempts >= max` arm was unreachable, and the only drain for the execution predicate (`poolExecutionBackstop.exhausted`) never ran. That matters because a genuinely stalled, still-held claim on a no-nudge agent could sit un-escalated indefinitely instead of being recycled.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **Park-forever on no-nudge agents:** the fix routes the empty-content case to the terminal escalation action, gated on a live re-read (`revalidate → Outstanding`) **and** `content == ""`, so it fires only for a genuinely stalled, still-held, runtime-quiet claim. A claim that completed between resolve and decision returns `Clear` and drains nothing — verified against `nudge_backstop.go`.
- **Exactly-once escalation:** the latch on `executionClaimNudgeStalledKey` is written before the event + drain, and the new test asserts exactly one drain and one event across further ticks — no repeat-drain.
- **Honest observability:** the stalled report now surfaces the real consumed attempt count instead of the meaningless constant cap (previously always "3").

Both reviewers independently validated the premise (one by reconstructing the pre-change control flow, one by running the focused backstop suite) rather than accepting it at face value, and the new test drives the real no-nudge path through a store-built in-progress claim.

- **mc-221e** — [Behavioral Correctness / Debuggability] minor: the stalled-telemetry field/comment labeled "delivered" actually reports *reserved/consumed* nudge attempts, since `reserve()` persists `attempts+1` before the external send and a transient send failure isn't rolled back. This is a pre-existing reserve-before-send ordering, not introduced here — this PR only moved the emitted value from the constant cap to the real count, which is a net improvement. Filed for a future relabel; not addressed in this PR.
- **mc-wsxs** — [Change Impact / Blast Radius] minor: the shared `content == ""` block now runs `revalidate` before returning, so a no-nudge, continuation-eligible pool session issues two extra live-store reads per post-grace tick. The outcome is unchanged (both sibling predicates' `exhausted` are no-ops), so this is an efficiency/consistency nit; filed for an optional intent comment. Not addressed in this PR.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `dc08c3f` has passing required GitHub checks (CI, integration, CodeQL, Mac regression, and the Go module vulnerability scan are all green), and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/32774921080

### Review Iterations
1 review pass performed (0 fix iterations); the change was approved on the first pass with only two non-blocking minor follow-ups, both filed above.

---

PR #5578, opened 2026-08-25
fix(reconciler): break the orphaned-seat work deadlock (ga-jrnou)

julianknutsen:
## Problem

An orphaned seat that still holds open assigned work was unreachable by every recovery lane, and each lane was individually correct to defer:

- **close_orphan** refuses while the seat holds work (`session_work_guard.go:126`).
- **wake / `projectDesiredState`** refuses because an orphaned base state raises `BlockerMissingConfig` → `DesiredStateBlocked`.
- **`releaseOrphanedPoolAssignments`** skips the work because `liveOpenSessionAssignmentExists` (`pool_session_name.go:619`) tests **bead status**, not runtime liveness — and the seat's session bead is still open.

`closeBead` is itself what runs the orphaned-work release cascade, so the close guard blocks its own release path. Observed wedged for 86 hours in maintainer-city.

## Fix

The orphan close is the one site that has **already confirmed the runtime is observably dead**, so it owns the tie-break: release the held work, then close. Restricted to `reason == "orphaned"` — a `suspended` seat is configured and merely scaled down, so its work stays put for its return.

`releaseConfirmedOrphanSessionWork` deliberately takes no liveness argument and performs no liveness probe; the caller must have confirmed death. That is the `ga-g3pf0` scar, where a session store serving zero session beads read as "every assignee is dead" and released live work every tick.

## Trace honesty

Both close sites recorded their decision **before** attempting the close and hard-coded `TraceOutcomeClosed`, so a seat that never closed was still reported "closed" every tick. `gc trace` is the documented entry point for a stuck reconciler (`engdocs/contributors/reconciler-debugging.md`), so that misreport actively misdirected the investigation for as long as the seat stayed wedged. Both sites now report the close that actually happened.

## Tests

Three tests, one per behaviour:

- `...ReleasesHeldWorkThenCloses` — pins the repair.
- `...RecordsClosedOnlyWhenClosed` — pins the trace honesty.
- `...LeavesLiveSeatWorkAlone` — the control that matters. Releasing work from a live seat is data loss, not recovery. It asserts the **mechanism**, not just the outcome: a running runtime must never reach the `close_orphan` site at all.

The control also documents a fixture trap worth knowing: a session bead needs `Type`/`Labels` set, because `session.ResolveSessionRecordByExactID` only recognises a typed (or labelled-and-untyped) bead as a session record. An untyped bead still reconciles, but every liveness observation on it reads not-running — silently turning a live seat into a confirmed orphan.

## Verification

- All three tests pass.
- **Mutation check:** replacing the release call with `nil` makes the release test fail (`work assignee = "polecat-1", want empty`) while the control still passes — so the wiring is load-bearing and the control is not vacuous.
- `go vet ./cmd/gc/ ./internal/session/` clean.
- `go test ./cmd/gc/ -run 'Reconcil|Orphan|Pool|SessionBead|Drain|Trace'` → ok, 60.9s.

---
- Related to #3566 — adds a confirmed-dead-orphan release path alongside releaseOrphanedPoolAssignments that reuses the same liveness predicates (the detached probe and the live re-read) instead of a holder-renewed lease, so it is a new release site the lease work in that issue would need to subsume rather than a step toward it. It does not touch the live-holder release race or the gate-2 named-session asymmetry that issue still tracks.
- Related to #4387 — covers one narrow lane of the abandoned-work problem: when the reconciler has confirmed an orphaned seat's runtime is dead, it now releases the pool-routed work that seat still held — clearing the assignee and returning the bead to open — instead of leaving it assigned to a corpse; it does not add the general sweep described in that issue (no lease/TTL/heartbeat, no reclaim for still-configured crashed sessions, and no re-dispatch cooldown, attempt limits, or mayor escalation)
- Related to #4448 — addresses one route by which a session bead whose runtime is confirmed dead can stay open indefinitely and keep consuming a pool slot — the reconciler's orphan-reap guard deferring while the seat still holds assigned work; it does not cover the post-drain-ack unreaped-shell case described here, and does not change how open_count is computed

### DIFF
--- a/cmd/gc/pool_session_name.go
+++ b/cmd/gc/pool_session_name.go
@@ -280,6 +280,86 @@ func releaseOrphanedPoolAssignments(
 	return released
 }
 
+// releaseConfirmedOrphanSessionWork releases the pool-routed work still held by
+// a session the reconciler has confirmed orphaned, so the close guard that
+// refuses to close a seat holding work stops being a permanent block.
+//
+// This is the tie-break for the deadlock in ga-jrnou. An orphaned seat holding
+// work is unreachable by every other lane: the close guard refuses while the
+// work is assigned, the wake path is blocked because an orphaned base state
+// raises BlockerMissingConfig, and releaseOrphanedPoolAssignments skips the work
+// because the seat's session bead is still open — liveOpenSessionAssignmentExists
+// tests bead status, not runtime liveness. Each lane defers to the others and
+// the seat wedges indefinitely.
+//
+// The caller MUST have confirmed the runtime is observably dead. This function
+// deliberately takes no liveness argument and performs no liveness probe: the
+// orphan-close site is the only caller precisely because it has already failed
+// closed on an unreadable liveness observation. Releasing work from a seat that
+// is actually alive is data loss, not recovery (ga-g3pf0).
+//
+// Every per-bead gate from releaseOrphanedPoolAssignments applies unchanged,
+// including the live re-read in liveWorkAssignmentStillReleasable — the tick
+// snapshot names candidates but never by itself justifies a release.
+func releaseConfirmedOrphanSessionWork(
+	cfg *config.City,
+	store beads.Store,
+	rigStores map[string]beads.Store,
+	assignedWorkBeads []beads.Bead,
+	info session.Info,
+) []releasedPoolAssignment {
+	if cfg == nil || store == nil || len(assignedWorkBeads) == 0 {
+		return nil
+	}
+	identifiers := make(map[string]struct{}, 5)
+	for _, id := range sessionAssignmentIdentifiersForConfigInfo(info, cfg) {
+		if id = strings.TrimSpace(id); id != "" {
+			identifiers[id] = struct{}{}
+		}
+	}
+	if len(identifiers) == 0 {
+		return nil
+	}
+
+	var released []releasedPoolAssignment
+	for i, wb := range assignedWorkBeads {
+		if wb.Status != "open" && wb.Status != "in_progress" {
+			continue
+		}
+		assignee := strings.TrimSpace(wb.Assignee)
+		if assignee == "" {
+			continue
+		}
+		if _, ok := identifiers[assignee]; !ok {
+			continue
+		}
+		template := routedToOrLegacyWorkflowTarget(wb)
+		if template == "" {
+			continue
+		}
+		agentCfg := findAgentByTemplate(cfg, template)
+		if agentCfg == nil || !agentCfg.SupportsGenericEphemeralSessions() {
+			continue
+		}
+		ownerStore := storeForPoolAssignment(cfg, store, rigStores, wb)
+		if ownerStore == nil {
+			continue
+		}
+		if !liveWorkAssignmentStillReleasable(ownerStore, wb.ID, wb.Status, assignee) {
+			continue
+		}
+		allowsRelease, clearDetached := detachedProbeAllowsOrphanRelease(wb)
+		if !allowsRelease {
+			continue
+		}
+		if !releaseOrphanedPoolAssignment(ownerStore, wb, clearDetached) {
+			continue
+		}
+		released = append(released, releasedPoolAssignment{ID: wb.ID, Index: i})
+	}
+	return released
+}
+
 // assignedWorkOwnerStore resolves the store that owns the assigned work bead at
 // index i: the index-aligned snapshot store when the caller supplied one (the
 // store-aware form), otherwise the routed/prefix fallback. A nil result means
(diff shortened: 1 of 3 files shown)


quad341:
@julianknutsen — this is a genuinely excellent piece of debugging. The three-way analysis is what makes it: `close_orphan` refusing while work is held, wake blocked on `BlockerMissingConfig`, and `releaseOrphanedPoolAssignments` skipping because `liveOpenSessionAssignmentExists` tests bead status rather than runtime liveness. Each lane individually correct, collectively wedged for 86 hours. Putting the tie-break at the one site that has already confirmed the runtime dead — and having `releaseConfirmedOrphanSessionWork` refuse to take a liveness argument at all, so it can't quietly drift into a prober — is the right shape. The ga-g3pf0 scar is respected, and I reproduced your mutation check: swapping the release for `nil` fails with `work assignee = "polecat-1", want empty` while the live-seat control still passes.

The trace-honesty half may be the more valuable one. `gc trace` confidently reporting "closed" for a bead that never closed misdirected the investigation for the entire 86 hours.

Two things I'll take on my side, then merge. Your trace test only observes the *successful* close, so it'd still pass with the old hard-coded `TraceOutcomeClosed` — I'll add a refused-close fixture (unrouted assigned work: the guard refuses, the releaser declines) so the honesty fix is actually pinned, plus coverage for the `close_failed_create` site. I'll also gate the release on an explicit `hasWork && err == nil` probe, since `!closed` also covers a dark-leg probe error.

Your commits, your credit.


---

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277246055

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277246197

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR breaks a three-way orphaned-seat work deadlock in the session reconciler (ga-jrnou): when a pool session's runtime died while still holding assigned work, the close-guard refused to close the session bead (work was still held), the session bead stayed open, and the status-based liveness check then made the bulk releaser skip that seat — so the work stayed pinned to a dead session indefinitely (an ~86-hour leak in the field). The fix adds a narrowly-scoped tie-break that releases a confirmed-dead orphaned seat's work back for re-routing, and its second half makes the reconciler trace record the close that actually happened (`Closed`/`NoChange`/`Skipped`) instead of a hard-coded `Closed`. That matters because it ends a real work-starvation deadlock and makes the reconciler's audit trail honest for debugging.

### Original PR Review
Decision: approve.

Specific gaps validated:
- **Orphaned-seat work deadlock (ga-jrnou):** the tie-break fires only from the confirmed-dead branch and only for `reason=="orphaned"`, breaking the close-guard → open-session-bead → skipped-releaser cycle. All three legs were traced and the targeted reconciler close test passes. (`cmd/gc/session_reconciler.go`)
- **Trace honesty:** the recorded outcome is now derived from the real close result rather than a hard-coded value; close behavior itself is byte-for-byte unchanged. (`cmd/gc/session_reconciler.go`)

The release path was scrutinized closely: it is gated by a fail-closed live re-read plus an atomic compare-and-set, reuses the already-trusted bulk-releaser per-bead gate chain, and is reached only after the caller has confirmed the runtime observably dead — so a transient store blip degrades to a safe no-op rather than dropping work.

Non-blocking follow-ups (shared for visibility — NOT addressed in this PR; filed as beads):
- `ga-7hsfm`: (test coverage, minor) No test pins the load-bearing invariant that an orphaned, confirmed-dead seat whose close helper fails for a *non-work* reason releases nothing — a regression that fired the release on a probe/write failure would pass the current suite unchanged.
- `ga-sg9op`: (correctness/clarity, minor) The `if !closed && reason=="orphaned"` branch enters on three distinct false-causes (assigned-work refusal, probe error, close-write failure); a typed refusal reason would make intent self-documenting so correctness no longer leans solely on the downstream re-read (defense-in-depth).
- `ga-1xpm9`: (blast radius, minor/low-confidence) The close-guard walks a multi-leg reachable-store plan while the release resolves a single owner store; a bead physically living in a store its routing doesn't resolve to could keep one seat wedged. Fails safe today and does not regress prior behavior.

### Maintainer Changes
No maintainer code changes — PR accepted as-is. The contributor's commits were rebased onto the latest `main` to keep the branch current and cleanly mergeable; no fixups were required. Short diff stat: 3 files changed, 404 insertions(+), 8 deletions(-).

### Final Review Status
Ready for the merge queue: final head `f5b23d39` has passing required GitHub checks (Check, CI / required, and all CodeQL Analyze jobs). The advisory `Mac / quality` job is red only from a runner-image Go toolchain lag (`go.mod requires go >= 1.26.6`, runner has `1.26.5`) that pre-exists on `main` and is untouched by this PR; it is not a required check and does not gate the merge queue. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/33074000951

### Review Iterations
1 review pass performed; approved on the first attempt with no fix iterations required.

---

PR #5579, opened 2026-08-25
fix(reconciler): rotate config-drift session keys by provider capability

### LINKED ISSUE #5573: bug: config-drift reset mints a session ID for providers that cannot accept one, destroying the conversation on every restart
## Summary

`resetConfiguredNamedSessionForConfigDriftInfo` (`cmd/gc/session_reconciler.go`) mints a fresh UUID v4 into `session_key` unconditionally. It is the only `GenerateSessionKey()` call site in the tree that does not first check whether the provider can actually accept a caller-supplied session ID.

For a provider with no `session_id_flag` there is no way to say "create a conversation with this ID". `resolveSessionCommand` therefore routes the minted key onto the **resume** path, producing a resume of a conversation that was never created. The provider rejects it, the pane dies during startup, and the session is rebuilt from scratch — losing its entire conversation history.

## Impact

Every configured-named session on a resume-only provider (`opencode`, `codex`, `gemini`) loses its conversation on each config-drift repair. Observed **5 for 5** across consecutive city restarts.

## Reproduction

City with `provider = "opencode"` and a `[[named_session]]` agent:

1. Run the city, let the named session start and capture a real key (`session_key=ses_…`).
2. `gc stop`, then start the city again.
3. Config-drift repair fires for the named session and rotates `session_key` to a UUID v4.
4. The next wake launches `opencode --session <uuid>` → `Error: Invalid session ID: unknown error`.
5. The pane dies, gc clears the key, retries, and comes up as a brand-new conversation.

## Evidence

Session bead `ga-r2l`, `session_key` transitions from `.gc/events.jsonl`. Every `city-stop` rotates a healthy `ses_…` key to a UUID, which is then cleared and replaced by a fresh `ses_…` one start later:

```
ses_fcf01550cffeq3VEU4YMXpYYlC -> 4b0c5b88-7a6c-4feb-9fae-888745517fd6   sleep_reason=city-stop
4b0c5b88-… -> ""                                                          (rejected, cleared)
""         -> ses_fcdfa6e85ffedZ3vVz2quz6ZTd                              (fresh conversation)

ses_fcdfa6e85ffedZ3vVz2quz6ZTd -> b995ac9e-0d8c-4bc3-8944-50b96c285dfc   sleep_reason=city-stop
ses_fcde424edffeE5J9OJuN9YG869 -> 355012d9-dec9-413e-9b0d-d70645f100b2   sleep_reason=city-stop
ses_fcc305acfffeQTAsR4x8iTIl7X -> de9b6c4b-925c-4745-a8c3-ac51d0b7852b   sleep_reason=city-stop
ses_fcb0627dbffehHu40L56wzoBht -> 40f28aed-29d2-46d8-bee7-d0fef4fbfeeb   sleep_reason=city-stop
```

The resulting start failure, from the session-reconciler trace:

```json
{"site_code":"reconciler.start.failed","reason_code":"start","outcome_code":"deadline_exceeded",
 "session_name":"mayor","template":"mayor",
 "fields":{"error":"resuming session: herdr: start \"mayor\": herdr [agent start mayor --kind opencode --pane wT:p1 --timeout 60000 -- --session 355012d9-dec9-413e-9b0d-d70645f100b2]: signal: killed"}}
```

The bead metadata at the rotation tick carries the exact signature of this function's patch — `ConfigDriftResetPatch` plus the four deferred-drift keys it clears:

```
continuation_reset_pending=true  started_config_hash=""  started_live_hash=""  live_hash=""
last_woke_at=""  restart_requested=""  pending_create_claim=""  pending_create_started_at=""
config_drift_deferred_at=""  config_drift_deferred_key=""
attached_config_drift_deferred_at=""  attached_config_drift_deferred_key=""
provider=opencode  resume_flag=--session  resume_style=flag
```

`gc config explain --provider opencode --json` confirms no `session_id_flag` is resolved for this provider.

## Root cause

`cmd/gc/session_reconciler.go`:

```go
rotatedSessionKey := ""
if preserveResume {
	rotatedSessionKey = priorSessionKey
} else if newKey, err := sessionpkg.GenerateSessionKey(); err == nil {
	rotatedSessionKey = newKey
}
```

Every other mint site is capability-gated on `session_id_flag`:

| Site | Guard |
|---|---|
| `internal/session/manager.go:867` | `if resume.SessionIDFlag != ""` |
| `internal/session/manager.go:1104` | `if resume.SessionIDFlag != ""` |
| `cmd/gc/session_beads.go:1840` | `tp.ResolvedProvider.SessionIDFlag != ""` |
| `cmd/gc/session_beads.go:2117` | `tp.ResolvedProvider.SessionIDFlag != ""` |
| `cmd/gc/session_lifecycle_parallel.go:1024` | `tp.ResolvedProvider.SessionIDFlag != ""` |
| `cmd/gc/session_reconciler.go:854` (`freshRestartSessionKeyInfo`) | `tp.ResolvedProvider.SessionIDFlag != ""` |
| **`cmd/gc/session_reconciler.go:5351`** | **none** |

`freshRestartSessionKeyInfo` already encodes the correct decision and returns `("", true)` — "capable of resume, but cannot be handed an ID, so clear the key". The restart-request handoff (`session_reconciler.go:2607`) and the fresh-cycle handoff (`session_bead_cycle.go:99`) both consume it correctly. This site was missed.

The function's doc comment shows the assumption that produced the gap — it describes the rotation as feeding `--session-id <new-uuid>`, which only holds for `session_id_flag`-capable providers. For a resume-only provider `resolveSessionCommand` falls through to `resolveResumeCommand` even on a first start, so the minted key becomes `--session <uuid>`.

## Fix

Rotate via `freshRestartSessionKeyInfo(tp, info)` and clear `session_key` when the provider has resume capability but cannot be handed an ID, matching the two existing handoffs. The next start then launches bare and the provider's own key is captured after startup by `gc prime --hook`.

PR follows.

Toady00:
Fixes #5573.

## Problem

`resetConfiguredNamedSessionForConfigDriftInfo` minted a fresh UUID into `session_key` unconditionally — the only `GenerateSessionKey()` call site in the tree that did not first check whether the provider can accept a caller-supplied session ID.

A provider with no `session_id_flag` cannot be told "create a conversation with this ID", so `resolveSessionCommand` routes any non-empty `session_key` onto the resume path — including on a first start, because the `--session-id` branch is skipped when the flag is empty. The minted key therefore became a resume of a conversation that was never created:

```
opencode --session 355012d9-dec9-413e-9b0d-d70645f100b2
Error: Invalid session ID: unknown error
```

The pane dies during startup, gc clears the key, retries, and the session comes back as a brand-new conversation with its history gone. Observed on **five consecutive city restarts** of a named opencode session — full forensics in #5573.

## Fix

Rotate through `freshRestartSessionKeyInfo(tp, info)`, which already encodes exactly this decision and returns `("", true)` for a provider that can resume but cannot be handed an ID. The restart-request handoff (`session_reconciler.go`) and the fresh-cycle handoff (`session_bead_cycle.go`) already consume it that way; this site was missed.

`ConfigDriftResetPatch` only writes `session_key` when non-empty, so the clear is stated explicitly — the same two-line shape both existing handoffs use.

`TemplateParams` is threaded in so the *resolved provider* is authoritative for the capability check rather than the bead's possibly-stale metadata mirror (which matters precisely in the provider-change case). Both production call sites already have `tp` in scope.

Behaviour is unchanged for `session_id_flag`-capable providers: they still mint, and the resume-preservation path is untouched.

## Tests

Two new regression tests:

- `..._ResumeOnlyProviderClearsKey` — the reset must clear rather than mint for an opencode-shaped provider.
- `..._ResumeOnlyProviderStartsBare` — end-to-end through `prepareStartCandidateForCity` + `startPreparedStartCandidate`, asserting the command handed to the runtime carries no `--session`.

Both were confirmed non-vacuous by reverting the fix; the end-to-end one reproduces the production failure shape exactly:

```
Start command = "opencode --session efcf7bae-487c-44af-be82-8f70541e4f0f",
  must not carry a --session argument: the provider never created that conversation
```

Two existing tests (`_AsleepResetClearsHashAndKey`, `_GeneratesKeyWhenNoneToPreserve`) assert that a fresh key **is** minted but never configured a provider, so they were relying on the unguarded path. They now pass an explicitly `--session-id`-capable provider, which is what their own doc comments always described ("so the next wake has a valid `--session-id` argument"). Their assertions are unchanged — the precondition is just no longer implicit.

## Verification

- `go build ./...`, `go vet ./cmd/gc/ ./internal/session/` — clean
- `./cmd/gc/` full suite, `./internal/session/...`, `./internal/config/...` — no new failures

Pre-existing failures on this machine, all reproduced on clean `upstream/main` and unrelated to this change: `TestCmdline_FailsClosedWhenUnreadable` (`internal/pidutil`, darwin), `TestProviderLiveClaudeKindPath` (`internal/runtime/herdr`, needs a live server), and `TestResolveTemplateRendersAbstractUpstreamPerHarness` (ambient `OPENAI_API_KEY`).


quad341 (approved):
@Toady00 — tracking five consecutive restarts down to the single `GenerateSessionKey()` call site in the tree that wasn't capability-gated, and then noticing that `resolveSessionCommand` routes a non-empty key onto the *resume* path even on a first start, is the kind of analysis that makes a fix reviewable in one pass.

I traced your claims against the source rather than taking them as given, and they hold: the `("", true)` contract, the `ConfigDriftResetPatch` non-empty-write behavior that makes the explicit clear necessary, and the fact that the start-prep mint is already gated so nothing re-mints what you cleared. Threading `TemplateParams` so the *resolved* provider wins over the bead's metadata mirror is the right call. And `_ResumeOnlyProviderStartsBare` asserting on the command the runtime actually received — not an intermediate value — is what makes this a real regression guard.

One thing that's ours, not yours: CI never ran a suite on this head (our evidence watchdog is unhappy about our own workflow gating, not your branch). So I'm taking it from here — adding one small test for the `resume_command`-only provider shape, pushing that to your branch with credit so CI actually fires, and merging on green. Nothing needed from you.

I also spotted an adjacent, *pre-existing* gap while reading: `preserveResume` doesn't consult the resolved provider, so a drift that is itself a provider change can preserve a key the new provider can't use. That's not yours to fix and it's not blocking you — I'm filing it separately and crediting your #5573 forensics as the source.


---

PR #5582, opened 2026-08-25
Make Dolt pull remote selection deterministic

quad341:
## Summary

- refuse ambiguous multi-remote pulls unless `GC_DOLT_REMOTE_<DB>` selects a configured remote
- require `GC_DOLT_PULL_ALLOW_REMOTE_<DB>=1` before pulling from an explicitly selected non-local remote
- apply the same deterministic policy to live SQL and CLI fallback paths
- add parallel-safe SQL and CLI regression coverage and reconcile the resource census

## Validation

- all 42 diff-owned top-level tests pass; focused package run has 363 PASS events, 0 FAIL, 0 SKIP
- required 40-job local union completed 35 PASS / 5 FAIL / 0 SKIP; every raw failure is tracked, non-diff-owned, and preserved in the gate record
- policy, vet, build, format, docs, native DoltLite, resource census, and isolated-cache affected lint pass
- merge-tree and ancestry-scope checks pass against current `main`

## Release gate

- [x] Exact-SHA reviewer PASS
- [x] Acceptance behavior covered in SQL and CLI modes
- [x] [Release gate evidence](https://github.com/gastownhall/gascity/blob/deploy/ga-nht26j-gate/release-gates/ga-nht26j-dolt-pull-remote-selection-gate.md)

Merge authority remains with the maintainer workflow.


quad341:
@quad341 — thank you for this. The decoy-`alpha`-first fixture is the detail that makes the review easy: it turns "did selection actually go through the override" into something a test can *prove*, rather than something that passes by accident on row ordering. Pairing that with assertions against the fake-dolt log — no `DOLT_PULL` present, not merely an error on stdout — is the difference between testing a message and testing the behavior. I ran your four CLI scenarios against the PR head and they all do exactly what they say.

One thing I'll take on my side. The CLI-mode rewrite parses candidates with a grep requiring `"url"` to sit immediately after `"name"`. On a `remotes.json` where `params`/`fetch_specs` lands between them, the old two-grep version resolved it and the new one returns nothing — so pull prints `skipped (no remote)` and exits 0. Silent success is the one outcome this whole change exists to prevent, so I'd rather harden it than ship it narrow. I'll make the parse key-order tolerant, add a fixture in that shape, bump the census by the one call site it adds, and re-run. Your policy logic and all eight tests stay exactly as they are.

The red `release-gate/deploy-clearance` check has no logs attached — that one's mine to chase, nothing for you to do.


---

quad341:
Closing as superseded (mayor disposition, 2026-08-31). This head (deploy/ga-nht26j-gate @ f8b80ce6) carries the pre-correction pull fix that was invalidated when the sole-non-local-remote policy hole was found; deploy-clearance is FAILURE and the review-of-record ga-g04htm now pins corrected SHA 72b7b05e on builder/ga-nht26j-footprint-fix. A fresh deploy branch will be cut from the reviewed SHA under bead ga-nht26j. The production defect (unordered dolt_remotes LIMIT 1 in pull/run.sh + sync/run.sh) is still live on main and remains tracked by ga-nht26j (pull) and ga-h9n6hq (sync).

PR #4868, opened 2026-07-31
fix(gc): skip stale asleep-session resume for wake_mode=fresh pool agents (#4849)

tdupu:
computePoolDesiredStates emitted a Tier:"resume" request for any assigned work bead whose assignee resolved to a non-closed session bead, without consulting the agent's wake_mode or the session's asleep state. A wake_mode="fresh" pool agent with a stale asleep session (left over from before a city stop) was therefore told to resume that dead session; at runtime the resume runs the session's pre_start hook, which hangs and is killed on the resume deadline, so the pool never materializes a live member (every reconcile re-attempts the same doomed resume) until the stale bead is otherwise cleared.

Guard the pool resume-tier construction: build an asleep-session set from sessionInfos (State == StateAsleep, which normalizeInfoState also folds "drained" into) and skip the resume when the agent's EffectiveWakeMode() is "fresh" and the resolved session is asleep, falling through to the scale_check / min-fill path so a clean session starts instead. A live (non-asleep) session still resumes, and agents with unset or wake_mode="resume" are unaffected. No signature change.

This is the residual sibling of #112, which fixed wake_mode="fresh" only in the session_reconciler.go drain-ack / session_key path and left the pool desired-state resume construction untouched.

Fixes #4849

## Summary

This PR fixes stale asleep-session resume for wake_mode="fresh" pool agents, completing the
fix from #112 which addressed only the session_reconciler.go side (#4849). A wake_mode="fresh"
agent inheriting a stale asleep session from before a city stop was incorrectly told to
resume, causing the pool to endlessly re-attempt a doomed resume that hangs on pre_start. The
fix guards pool resume-tier construction to skip resuming when EffectiveWakeMode() equals
"fresh" and the session is asleep, falling through to scale_check/min-fill for a clean session
instead.

## Testing

- [ ] `make check`
- [ ] `make check-docs` if docs, navigation, or links changed
  > **Note:** `docs/` is authored for [docs.gascityhall.com](https://docs.gascityhall.com) (Mintlify), not for direct GitHub viewing. Use extensionless page links (e.g. `/tutorials/01-beads`, not `/tutorials/01-beads.md`). If something looks broken on GitHub but works on the live site, that's intentional.
- [ ] `make test-integration` if runtime, controller, or workflow behavior changed

## Checklist

- [ ] Linked an issue, or explained why one is not needed
- [ ] Added or updated tests for behavior changes
- [ ] Updated docs for user-facing changes
- [ ] Called out breaking changes or migration notes

Issue #4849, opened 2026-07-30
bug: gc: pool desired-state resumes stale asleep sessions for wake_mode="fresh" agents


tdupu:
### Before you continue

- [x] I searched existing issues and did not find a duplicate.
- [x] I read the relevant docs and contributor guidance.

### Gas City version

 gc version: dev (built from source)                                                                                                                                                                  Verified against gastownhall/gascity origin/main (latest tip e6135a435, re-checked   today). Latest tagged release: v1.4.0. The bug is source-level and long-standing:   pool_desired_state.go's resume-request construction never consults wake_mode   (0 WakeMode refs); any build with that code is affected.

### Environment

  - OS: macOS (Darwin), arm64; Shell: zsh; Runtime: local tmux (not Kubernetes)    Note: this bug is environment-independent. It's deterministic in the pool   desired-state builder (cmd/gc/pool_desired_state.go, computePoolDesiredStates) and   reproduces at the unit level with no runtime, city, or pack (the Go snippet in   Reproduction). OS/shell/provider are just where it was observed; none is   load-bearing. The config shape that triggers it: a pool agent with   wake_mode = "fresh", min_active_sessions >= 1, and a leftover asleep session bead   (e.g. from before a city stop).

### Reproduction

Unit-level, no pack/runtime needed — a wake_mode="fresh" pool agent with an asleep
  pre-restart session that owns assigned work:

  ```go
  agent := poolAgent("polecat", "hello-world", intPtr(5), 1) // Name/Dir, max, min
  agent.WakeMode = "fresh"
  cfg := &config.City{Agents: []config.Agent{agent}}

  sessions := []beads.Bead{
      {ID: "mc-sctve", Status: "open", Type: "session", Metadata: map[string]string{
          "template": "hello-world/polecat", "session_name": "polecat-mc-sctve",
          "state": "asleep", "pool_managed": "true",
      }},
  }
  work := []beads.Bead{ workBead("hw-8lb", "hello-world/polecat", "mc-sctve", "in_progress", 2) }


  result := ComputePoolDesiredStates(cfg, work, sessionInfosFromBeads(sessions),
      map[string]int{"hello-world/polecat": 1})
  ```

### Expected behavior

  For wake_mode="fresh", the stale asleep session is NOT resumed. The result contains
  a fresh Tier:"new" request for the pool (start clean), not a resume of the dead
  session.

### Actual behavior

  ComputePoolDesiredStates returns a SessionRequest{Tier:"resume",
[...]

PR #3842, opened 2026-06-30
Add hook-claim continuation nudges for graph workflows

quad341:
Fix pool graph.v2 formula sessions that stall after root claim.

## Problem

Pool agents (\`min=0\`) running a \`graph.v2\` formula via \`gc sling\` claim the workflow root but then idle indefinitely — no step bead ever executes. The sling-time nudge targets the pool template name (\`gascity/worker\`), not the concrete slot name (\`gascity/worker/slot-0\`). The slot's poller never sees that nudge.

## Fix

At \`gc hook --claim\` time, when a pool graph.v2 root is claimed and at least one continuation sibling is pre-assigned, enqueue a per-session nudge targeting the claiming session name directly. The existing nudge delivery machinery handles the rest.

## What's included

- \`fix(reconciler): graph-only readiness + idle-respawn safety-net\` — reconciler correctness fixes: idle pool sessions holding assigned-but-not-yet-started work are drained and respawned so the work can be picked up
- \`feat(hook): enqueue hook-claim-continuation nudge for pool graph.v2 roots\` — the core fix
- \`fix(reconciler): apply De Morgan's law to idle-probe eligibility check\`
- \`fix(hook): thread cfg and session fence into continuation nudge enqueue\`
- \`fix(reconciler): close drain-cancel fall-through for assigned-work-only sessions\` — named/non-interactive sessions that regain assigned work now get \`cancelSessionDrain\` called as before; \`cancelSessionDrain\` skips non-cancelable idle-respawn drains so pool sessions mid respawn-drain are unaffected
- \`test(reconciler): add cancel-drain integration test\` — adds \`TestReconciler_CancelsDrainForNamedIdleAssignedWorkOnlySession\` covering the drain-cancel fall-through fix at the reconciler level
- \`refactor: split GraphOnlyReadyStore and drop over-broad docsync ga- skip\` — removes \`GraphOnlyReadyStore\` from this PR (no production store implements it; split to a follow-up that adds the Router-side implementer); drops the \`ga-*\` skip from \`TestDocDirCoverage\` (not needed in a clean checkout)
- Invariant comment at the \`gc.kind==workflow\` trigger gate naming the single-writer invariant in \`formula/compile.go\` and the pool-routing invariant in \`graphroute/graphroute.go\`

## Testing

Four new TDD tests in \`cmd/gc/cmd_hook_claim_continuation_nudge_test.go\`:
[...]

PR #3405, opened 2026-06-12
fix(graphroute): stamp gc.continuation_group on pool-routed graph.v2 steps

quad341:
## What this changes

Pool-routed graph.v2 work steps now stay on the same pool slot across a multi-step workflow. `ApplyGraphRouteBinding` stamps the existing `gc.continuation_group` and `gc.session_affinity` metadata on MetadataOnly pool-routed steps, which lets the hook claim path pre-assign sibling steps to the same pool slot.

Single-session route binding behavior is unchanged. Pool-routed steps still do not get a concrete session name at route-binding time.

## Review notes

- Uses existing metadata keys: `gc.continuation_group=pool-workflow` and `gc.session_affinity=require`.
- The implementation is limited to `internal/graphroute`; there are no formula, schema, or protocol changes.
- `test/docsync` now ignores the repository's git-excluded `worktrees` directory so local worktree markdown does not trip doc coverage.

## Test plan

- [x] `go test ./internal/graphroute/... ./test/docsync/...`
- [x] `go vet ./internal/graphroute/...`
- [x] GitHub `CI / required`, `CI / preflight`, `CI / integration`, process shards, integration shards, Dashboard SPA, and CodeQL are green on the PR branch
- [x] Release gate: [`release-gates/ga-39tsop-pool-step-slot-affinity-gate.md`](release-gates/ga-39tsop-pool-step-slot-affinity-gate.md)

PR #5584, opened 2026-08-25
fix(graphroute): route one-shot pool steps independently

### LINKED ISSUE #2978: bug: graph.v2 multi-step molecule routed to a pool has no slot/worktree affinity — steps scatter across slots and stall
## Summary

A `graph.v2` (formula_v2) multi-step molecule routed to a built-in-routing **pool** has **no slot/worktree affinity** across its steps. Each step bead is claimed independently by whichever pool slot is free, but a step's working state (the `gh pr checkout` branch, plan files under `.gc/pr-pipeline/…`) is **slot-local**. When step N+1 lands in a different slot than step N, it has neither the branch nor the plan, cannot proceed without violating stay-in-worktree rules, and HELP-stalls. Every multi-step molecule auto-routed to a pool wedges.

## Symptom

On a live city, auto-routing `mol-pr-iterate` to the `polecat` pool for 7 copilot-reviewed PRs wedged **all 7**. The chains split across `polecat-1..4`:

- Step `intake` ran `gh pr checkout <pr>` and wrote `.gc/pr-pipeline/iterate/pr-<N>.md` into **one** slot's worktree.
- The next ready step (`apply`, etc.) was then claimed by a **different** slot, whose worktree is on a detached/clean HEAD with no branch and no plan.
- That slot files a HELP: *"apply step routed to wrong worktree — PR branch + plan live in polecat-3, I'm polecat-4 on detached HEAD, cannot check out a branch busy in another worktree."*

Compounding second blocker: even when a slot tries to recover via `gh pr checkout <pr>`, it fails with **`fatal: '<branch>' is already used by worktree at …`** when the PR branch is checked out in another live slot's worktree.

Result: `expected` a multi-step molecule to run to completion on the pool → `got` every chain stalled after step 1, work silently stranded.

## Reproduction

City with `[daemon] formula_v2 = true` and a built-in-routing pool agent (`polecat`, `min_active_sessions >= 2` so steps can land on different slots).

1. `gc sling polecat <bead> --on mol-pr-iterate --var pr=<N> --var feedback_source=copilot --var feedback_ref=<id> --nudge`
2. A slot claims step 1 (`intake`): it `gh pr checkout <N>` + writes the plan into its worktree, closes step 1.
3. Step 2 becomes ready; a *different* free slot claims it.
4. Step 2 cannot find the branch/plan (they're in slot-1's worktree) → HELP-stalls.

Reliable workaround we use: bypass the molecule and direct-dispatch the whole task to **one** slot via `gc session nudge`, so all steps run in that slot's worktree.

## Root cause

The pool claim path scopes only on the pool name, with **no per-workflow slot binding**:

- `internal/config/config.go:2839-2841` — `bdReadyPoolDemandShell` is the canonical pool demand predicate:
  ```
  bd ready --include-ephemeral --metadata-field "gc.routed_to=$target" --unassigned --exclude-type=epic --json
  ```
  It filters solely on `gc.routed_to=<pool>`. Any ready, unassigned step of any in-flight workflow routed to that pool is claimable by **any** free slot. There is no `workflow_id`/root-bead → slot stickiness.
- `internal/config/config.go:~2964` (EffectiveWorkQuery tiering) — Tier 3 (pool) is `ready + unassigned + routed_to`; nothing keys a step to the slot that ran its predecessor.

Step working state is **not** portable to another slot:

- `cmd/gc/build_desired_state.go:~2934` (`stampRunSessionIdentity`) stamps `gc.session_name` + `gc.work_dir` onto a step **at execution time, for observability only** (dashboard diff/session drill-in, #2843). These keys are **never read** by the claim/work-query path — confirmed: `gc.work_dir` does not appear in `internal/config/config.go`, and the work-query predicate omits it.
- There is no mechanism that persists a step's outputs (base SHA, parsed feedback, checked-out branch) into the **root bead** so a subsequent step in another slot can resume; they live in one worktree's filesystem.

History shows the affinity was intended but never wired: commit `4e7217c87` (*"chore(formulas): remove unused continuation_group and session_affinity metadata"*) removed `gc.continuation_group` / `gc.session_affinity` from formulas with the note *"Neither … is read by any Go runtime or controller code."* (`internal/graphroute/` and `internal/molecule/` tests still reference these keys, so the routing notion exists in test fixtures but isn't honored by the pool claim path.)

This is an implementation gap, not a design conflict: `engdocs/design/agent-pools.md` (Status: Implemented) describes pool up/down-scaling but is silent on multi-step affinity; `engdocs/design/formula-v2-transient-retries.md` (Draft) addresses *retry* of the same step, not step N→N+1 continuity. No accepted design mandates the current scatter behavior.

## Fix candidates

**A. Per-workflow slot affinity (claim-path)** — when a slot claims a workflow's first step, bind subsequent steps of that `workflow_id`/root to that slot/session; fall back to any free slot only if the bound session is gone.
- Scope: medium — the pool demand/claim tiering in `internal/config/config.go` (~2839–2965) + stamp the bound session on the root at first claim. Could revive the already-test-referenced `gc.session_affinity` / `gc.continuation_group` notion in `internal/graphroute/` rather than inventing new metadata.
- Trade-off: restores intended continuity with minimal surface; correctness win. Adds scheduling rigidity (a busy affine slot serializes its chain) and needs explicit fallback when the slot dies mid-chain.

**B. Slot-portable step state** — stop depending on slot-local worktree state: persist step outputs (base SHA, parsed feedback, plan) into root-bead metadata/notes, and have each step re-establish its worktree deterministically via a shared "acquire worktree for PR" helper that handles the *"branch already used by worktree"* collision (detach/reuse the holding worktree).
- Scope: larger — touches every multi-step formula's step contract (`mol-pr-iterate`, `mol-adopt-pr`, `mol-pr-from-issue`, `mol-do-work`) + a new shared worktree-acquire helper.
- Trade-off: most robust — any slot can resume, enables recovery and true parallelism, and the helper also fixes the branch-collision blocker. Much bigger blast radius across all molecules.

**Recommendation:** **A** as the immediate correctness fix (small, targeted, restores the apparently-intended affinity that `4e7217c87` noted was never wired), with **B** as the structural follow-up for resilience and to retire the branch-collision blocker. The maintainer's call.

## Adjacent / out of scope

- **Branch-collision** (`gh pr checkout` → *"branch already used by worktree"*) is a separable secondary blocker; candidate B's worktree-acquire helper would subsume it, but it can be fixed independently.
- #1052 (pool work-claim race — two sessions execute the same bead) is a distinct concurrency bug.
- #1465 (per-step recycle policy) and #2793 (orphan-release sweep misses open step beads in graph.v2) are adjacent step-lifecycle concerns, not this.
- #2763 (roots never claimed — fixed) and #2843 (work_dir/session_name stamp — observability) are related but do **not** address step affinity; this gap surfaces precisely *after* those fixes make pool claiming work.

## Maintainer context

Surfaced 2026-06-02 dogfooding `ds-research`: a local periodic order auto-routed `mol-pr-iterate` to the `polecat` pool for 7 copilot-reviewed PRs and all 7 wedged with steps scattered across `polecat-1..4`. Observed on a pinned `gc` (which already carries the #2763 and #2843 fixes); root cause verified present on `origin/main` (`4ff7dd6dc`) via the file:line refs above.

### LINKED ISSUE #5583: graphroute stamps session-affinity/continuation-group on one_shot pool steps, parking work on dead sessions
# graphroute stamps session-affinity/continuation-group on one_shot pool steps, parking work on dead sessions

## Environment

- gascity, upstream base `2cd07e018`
- Component: `internal/graphroute` (graph.v2 formula routing), invoked from
  `cmd/gc/cmd_convoy_dispatch.go` and `cmd/gc/cmd_sling.go`
- Affects any agent pool configured with `config.Agent.Lifecycle == "one_shot"`
  (bounded/exiting runtime), routed via a formulas v2 graph recipe

## Repro (minimal)

1. Configure an agent pool `P` with `lifecycle: one_shot` and instance
   expansion enabled (`SupportsInstanceExpansion() == true`).
2. Author a two-step graph formula: step A `routed_to: P`, step B
   `routed_to: P`, B depends on A (ordinary sequential graph step, no
   explicit shared-drain construct).
3. Dispatch the formula. A fresh one_shot session claims step A, runs to
   completion, and exits (process terminates; no runtime survives).
4. Step B becomes ready.

## Expected

Step B is claimable by any fresh pool session of `P`, since no one_shot
runtime survives to carry continuation.

## Actual

Step B was stamped (at graph-route time, before any session claimed step A)
with `gc.continuation_group=pool-workflow` and `gc.session_affinity=require`,
pre-assigning it to the *same session name* that claimed step A. That session
no longer exists once the one_shot invocation exits. Step B stays parked on
the dead session name: repeated re-claim/recheck attempts fail against a
session that will never come back, and the step remains open for 10+ minutes
until a new seat is minted and something forces reassignment.

## Root cause

`ApplyGraphRouteBinding` in `internal/graphroute/graphroute.go` stamped
`ContinuationGroupMetadataKey` and `SessionAffinityMetadataKey`
unconditionally for every `MetadataOnly` (pool-routed) step, regardless of
agent lifecycle. That stamping is correct for persistent pools (fixes
#2978: keeps molecule steps from scattering across pool slots) but wrong
for `one_shot` pools, where the claiming session cannot outlive a single
step.

## Workaround

Derive an `IndependentSteps` flag from `config.Agent.Lifecycle == one_shot`
in a shared `GraphRouteBindingForAgent` helper, and skip the
continuation-group/session-affinity stamp when set. Persistent pools and
explicit shared-drain formula constructs are unaffected. See the fix on
branch `fix/one-shot-pool-independent-steps`
(commit `e356ea793ea3b7a55887de03062532580689a8fc`).

## Possibly related

#3405, #3842, #4868, #4849 (not independently verified against this fix;
noted only as adjacent-topic issues worth cross-checking during triage).

jacobhausler:
# fix(graphroute): route one-shot pool steps independently

Fixes #5583

Base: upstream `2cd07e018`
Branch: `fix/one-shot-pool-independent-steps`
Commit: `e356ea793ea3b7a55887de03062532580689a8fc`

## Summary

`one_shot` agent pools have no runtime that survives past a single bounded
invocation, but graph-routed steps were stamped with the same
`gc.continuation_group=pool-workflow` / `gc.session_affinity=require` pair
used for persistent pools. That pins a workflow's next step to the exact
session name that claimed the previous step. Once the one_shot session
exits, the next step is parked on a dead session name and never gets
reclaimed on its own (repeated re-claim/recheck, step open 10+ min, until a
fresh seat happens to be minted).

## Behavior change

- Add `GraphRouteBinding.IndependentSteps`, derived in a new
  `GraphRouteBindingForAgent(agentCfg)` helper from
  `config.Agent.Lifecycle == config.AgentLifecycleOneShot`.
- `ApplyGraphRouteBinding` now skips (deletes) the continuation-group and
  session-affinity metadata keys for `MetadataOnly` bindings when
  `IndependentSteps` is set, instead of stamping them unconditionally.
  Each step still routes to the pool (`RoutedToMetadataKey`, empty
  `Assignee`) — only the automatic cross-step pinning is omitted.
- `cmd_convoy_dispatch.go`, `cmd_sling.go`, and
  `ResolveGraphStepBindingWithVars` are refactored to call the shared
  `GraphRouteBindingForAgent` helper instead of duplicating the
  instance-expansion check, so the derivation is consistent everywhere a
  binding is built.
- `ApplyGraphRouting` now builds a `defaultRoute` via the same helper
  (preserving the caller's already-normalized `routedTo` identity) and
  passes it through a new `DecorateGraphWorkflowRecipeWithDefaultBinding`,
  replacing the old ad-hoc `sessionName`/`DecorateGraphWorkflowRecipe` call.
- Persistent (non-one_shot) pools: unchanged behavior — continuation group
  and required session affinity are still stamped (fixes #2978 remains
  intact).
- Explicit shared-drain / shared single-lane formula constructs are a
  separate contract from ordinary pool routing and are not touched by this
  change.
- Doc update: `engdocs/design/agent-pools.md` gains a "Graph workflow
  continuity" section explaining the persistent-vs-one_shot split and
  the resulting requirement that one_shot formulas persist inter-step state
  in the work artifact/bead graph rather than in a surviving process.

## Test plan

`internal/graphroute` (36 tests pass, focused run), including new:
- `TestApplyGraphRouting_OneShotPoolLeavesExecutableStepsIndependent`
- `TestDecorateGraphWorkflowRecipe_PerStepOneShotPoolTargetLeavesStepIndependent`

`cmd/gc` (2 new tests):
- `TestDecorateDynamicFragmentRecipeOneShotPoolFallbackLeavesStepsIndependent`
- `TestDecorateDynamicFragmentRecipePerStepOneShotPoolTargetLeavesStepIndependent`

Locally: post-adoption canary (worker -> Opus verifier -> drain -> finalize)
run against a one_shot pool closed clean end-to-end.

## Compatibility

Persistent pools are unaffected — the stamping path and its metadata keys
are unchanged for any agent whose `Lifecycle` is not `one_shot`.

## Possibly related

#3405, #3842, #4868, #4849 — not fetched/verified; flagging as adjacent
topics for triage, not claimed duplicates or fixes.


quad341:
Maintainer note (hold triage): sorry for the silence here. This PR has been on an architecture hold since 10 September, and nobody told you why.

The fix is wanted: a one_shot pool step shouldn't inherit a persistent session pin. The hold is about one decision inside it, which is what a one_shot route does with a formula-declared continuation group. This branch clears it. We decided `main` should refuse that case loudly instead, and that guard has now landed as #6360 (6e9a24a559). It changes the same `ApplyGraphRouteBinding` branch this PR edits, which is why the PR now conflicts.

To clear the hold, please rebase onto current `main`, keep `main`'s version of that branch as it is, and drop this PR's own version of it. What should remain is the wiring that sets `binding.IndependentSteps` from the agent's lifecycle config (`main` has the field, but nothing sets it yet), plus your resolver-side changes and tests, adjusted to #6360's behaviour where they overlap. Resolving the conflict by keeping this PR's hunk would quietly undo the guard, so please take `main`'s side there.

#5586 carries its own copy of the same change, so the same applies there. Once a rebased head is up, the review pipeline will pick it up again. If you'd rather we carry the rebase for you, say so here. Thanks for your patience with this one.

jacobhausler:
Rebased onto current main (2898f5467) as you described, so no need for you to carry it — thanks for the offer. main's `ApplyGraphRouteBinding` branch is kept verbatim (proved, not asserted: the extracted function body diffs byte-identical against `origin/main`) and this PR's version of it is dropped, along with the reworded `IndependentSteps` field doc; what remains is the wiring that derives `binding.IndependentSteps` from the agent's `lifecycle` (a new `GraphRouteBindingForAgent`, the three call sites collapsed onto it, and `ApplyGraphRouting` threading that default route into the existing `DecorateGraphWorkflowRecipeWithDefaultBinding`), plus the resolver-side tests adjusted to #6360 — the four independence atoms no longer pre-seed a formula-declared `gc.continuation_group`, since that seed asserted behaviour #6360 deliberately removed, and a direct `GraphRouteBindingForAgent` test now covers one-shot pool / persistent pool / named-session agent; `engdocs/design/agent-pools.md` documents main's refuse-loudly semantics instead of the old override claim. Gated before push on `go build ./...`, `go vet ./internal/graphroute/...`, gofmt-clean touched files, `go test ./internal/graphroute/...`, and the named `cmd/gc` decoration tests — all green. One residual worth stating plainly rather than burying: with main's arm order restored, a step whose binding carries a continuation group still takes the `gc.session_affinity=require` stamp regardless of the mark, so `IndependentSteps` currently governs only the stale-group case #6360 introduced (refuse loudly on a formula-declared group, clear the router's own `drain:` bookkeeping) — flagging it as a known boundary, not asking for anything in this PR.

quad341 (approved):
@jacobhausler — this is merging, and the delay was on our side, not yours.

Thank you for taking the #6360 rebase yourself and for proving it rather than asserting it. Diffing the extracted function body byte-for-byte against `origin/main` to show the refuse-loudly guard survived meant we could verify it in minutes instead of re-deriving it. Checked independently on our end, twice now: no hunk in this diff touches the body of `ApplyGraphRouteBinding`. Your claim holds exactly as stated.

And thank you for the residual note, which is the part I actually want to single out. You could have shipped this as "fixes #5583" and nobody would have caught it this week. Instead you measured what happens after the rebase, found that a one-shot pool step under a shared drain still lands `gc.continuation_group=drain:<id>` with `gc.session_affinity=require`, said so in the PR, and wrote a test comment that explicitly refuses to bless the behavior it pins — *"This test records the behavior; it does not bless it."* That comment is going to save somebody a bad afternoon. It is a better artifact than a green checkmark.

So, two things on our side. I've reworded the link to `Refs #5583` so the issue stays open, and filed a follow-up for the case you found. While tracing it I found the exposure is likely worse than "the pin lingers": the continuation nudge lane gates on `pool_managed` and keys candidates off a *live* session bead's ID, and the claim backstop deliberately defers this row shape to that lane — so a seat that has already exited appears to be served by neither. That reads as permanent silence rather than slow recovery. All of it is attached to the follow-up.

You were right that the question can't be answered without editing the branch #6360 asked you to keep verbatim, and right to stop there rather than guess. That one's ours. The dedup of `SupportsInstanceExpansion` onto `GraphRouteBindingForAgent` across the three call sites is a real cleanup on top — those had drifted into three independent copies. Good work; please send the next one.

PR #5586, opened 2026-08-25
fix(graphroute): resolve one-shot lifecycle for rig-scoped route targets

### LINKED ISSUE #5585: rig-scoped one_shot route targets stripped to bare name before ResolveAgent, resolve as persistent
# rig-scoped one_shot route targets stripped to bare name before ResolveAgent, resolve as persistent

## Environment

- gascity, follow-up to `e356ea793` (fixes #2978's one-shot regression)
- Component: `internal/graphroute` (`ApplyGraphRouting`, order-dispatch path),
  invoked from `cmd/gc/cmd_convoy_dispatch.go` and `cmd/gc/cmd_sling.go`
- Affects any `one_shot` agent addressed via a rig-scoped route target
  (`<rig>/<agent>`), e.g. production `fable-nomad/claude-sonnet-one-shot`,
  `fable-nomad/claude-opus-one-shot`

## Repro (minimal)

1. Configure a `one_shot` agent `oneshot` under rig `rig1`.
2. Dispatch a graph step via order-dispatch (`a == nil`) with
   `routedTo = "rig1/oneshot"`.
3. Use (or plug in) a resolver that matches agents only by fully qualified
   identity (`dir/name`), not by bare `Name` + separate rig-context.

## Expected

`ApplyGraphRouting` resolves `rig1/oneshot`, sees `Lifecycle == one_shot`,
and leaves the step independent: no `gc.continuation_group`, no
`gc.session_affinity=require`.

## Actual

`ApplyGraphRouting`'s `a == nil` branch stripped the rig prefix off
`routedTo` into a bare `baseName` before calling
`deps.Resolver.ResolveAgent(cfg, baseName, rigContext)`. A resolver that
requires the qualified string on the `name` argument itself (rather than
reconstructing it from `name` + `rigContext`) fails to resolve, `ok` is
false, and `ApplyGraphRouting` returns early — never stamping
`gc.routed_to` and never clearing stale `gc.session_affinity`/
`gc.continuation_group` metadata. Steps were observed in production parked
on exited one-shot sessions for `fable-nomad/claude-sonnet-one-shot` and
`…/claude-opus-one-shot`.

## Root cause

Same shape as `e356ea793`'s underlying issue, in an adjacent spot: rig-scope
information is discarded (bare-name stripping) and expected to be
reconstituted separately (via `rigContext`) rather than carried through
intact. `GraphRouteRigContext(routedTo)` extraction and the resolver's
qualification requirements can disagree.

## Fix

Pass the full qualified `routedTo` to `deps.Resolver.ResolveAgent` instead
of the stripped `baseName`; behavior for already-bare (unscoped) targets is
unchanged since `routedTo == baseName` in that case. See fix on branch
`fix/scoped-one-shot-lifecycle` (commit `9ac41cad1`), stacked on
`fix/one-shot-pool-independent-steps` (`e356ea793`).

jacobhausler:
# fix(graphroute): resolve one-shot lifecycle for rig-scoped route targets

Fixes #5585

**Note:** GitHub does not allow a PR opened via a fork to use another fork branch as its base unless that branch also exists in the base repo. This PR is opened with base `main` instead of stacking directly on #5584 as originally planned; it genuinely depends on #5584 (fix/one-shot-pool-independent-steps) and its diff will include that PR's commit until #5584 merges. Please review/merge #5584 first.

Base: `fix/one-shot-pool-independent-steps` (`e356ea793`) — **stacked PR**
Branch: `fix/scoped-one-shot-lifecycle`
Commit: `9ac41cad15bd7ff095f7371cd1cb862d8301b0c0`

## Summary

Follow-up to `e356ea793`, which taught `ApplyGraphRouting` to skip
continuation-group/session-affinity stamping for `one_shot` pools. That fix
only takes effect if the agent resolves with the right `Lifecycle` in the
first place. In the order-dispatch (`a == nil`) branch, `ApplyGraphRouting`
stripped the rig prefix off `routedTo` into a bare `baseName` before calling
`deps.Resolver.ResolveAgent(cfg, baseName, rigContext)`, relying on the
resolver to reconstruct the rig-scoped identity from `baseName` +
`rigContext` alone. A resolver matching only on the fully qualified agent
identity fails to resolve rig-scoped `one_shot` targets like
`rig1/oneshot`, so the step falls through the `!ok` early return: no
`gc.routed_to` stamp, and stale `gc.session_affinity=require` /
continuation-group metadata is left in place, parking the step on an
exited one-shot session. Observed in production on
`fable-nomad/claude-sonnet-one-shot` and `…/claude-opus-one-shot`.

## Fix

- Pass the full qualified `routedTo` to `deps.Resolver.ResolveAgent`
  instead of the rig-prefix-stripped `baseName`.
- Bare (unscoped) targets are unaffected: `routedTo == baseName` in that
  case, so behavior is unchanged.
- No change to `e356ea793`'s `IndependentSteps` stamping logic itself —
  this only fixes resolution so that logic actually sees the right
  `Lifecycle`.

## Test plan

`internal/graphroute` — new:
- `TestApplyGraphRouting_OrderDispatchNilAgent_RigScopedOneShot_IndependentSteps`
  - subtest `testAgentResolver (matches bare name too)`: passes pre-fix too
    (double-matches on bare `Name` regardless of rig; doesn't expose the
    bug).
  - subtest `qualifiedOnlyAgentResolver (rig-scoped only)`: RED pre-fix
    (resolver matches only `QualifiedName()`), GREEN post-fix. This is the
    case that isolates the bug.

`go test ./internal/graphroute/...` passes, including all `e356ea793` /
`71de83f27` tests (e.g.
`TestApplyGraphRouting_OneShotPoolLeavesExecutableStepsIndependent`,
`TestDecorateGraphWorkflowRecipe_PerStepOneShotPoolTargetLeavesStepIndependent`).
`go vet ./internal/graphroute` clean. `go build ./cmd/gc` clean.

## Compatibility

Bare (unscoped) targets and persistent pools: unchanged — this PR only
touches how the agent is *resolved*, not the stamping decision itself.

## Stack

Targets `fix/one-shot-pool-independent-steps` (`e356ea793`), not main —
merge order matters since this fix only matters given that PR's gating.

### DIFF
--- a/cmd/gc/cmd_convoy_dispatch.go
+++ b/cmd/gc/cmd_convoy_dispatch.go
@@ -17,7 +17,6 @@ import (
 	"time"
 	"unicode/utf8"
 
-	"github.com/gastownhall/gascity/internal/agentutil"
 	"github.com/gastownhall/gascity/internal/beadmeta"
 	"github.com/gastownhall/gascity/internal/beads"
 	"github.com/gastownhall/gascity/internal/config"
@@ -1232,9 +1231,8 @@ func graphFallbackBindingForBead(source beads.Bead, store beads.Store, cityName,
 		return graphRouteBinding{}, fmt.Errorf("unknown formulas v2 fallback target %q on %s", routedTo, source.ID)
 	}
 
-	binding := graphRouteBinding{QualifiedName: agentutil.RoutedToIdentity(&agentCfg)}
-	if agentCfg.SupportsInstanceExpansion() {
-		binding.MetadataOnly = true
+	binding := graphroute.GraphRouteBindingForAgent(agentCfg)
+	if binding.MetadataOnly {
 		return binding, nil
 	}
 	if source.Assignee != "" {
--- a/cmd/gc/cmd_convoy_dispatch_test.go
+++ b/cmd/gc/cmd_convoy_dispatch_test.go
@@ -1769,6 +1769,117 @@ func TestDecorateDynamicFragmentRecipePreservesPoolFallbackAndScopeMetadata(t *t
 	}
 }
 
+func TestDecorateDynamicFragmentRecipeOneShotPoolFallbackLeavesStepsIndependent(t *testing.T) {
+	store := beads.NewMemStore()
+	cfg := &config.City{
+		Workspace: config.Workspace{Name: "test-city"},
+		Daemon:    config.DaemonConfig{FormulaV2: boolPtr(true)},
+		Agents: []config.Agent{
+			{
+				Name:              "worker",
+				Lifecycle:         config.AgentLifecycleOneShot,
+				MinActiveSessions: intPtr(0),
+				MaxActiveSessions: intPtr(3),
+			},
+		},
+	}
+	config.InjectImplicitAgents(cfg)
+	addTestControlDispatcherAgents(cfg, "")
+
+	source := beads.Bead{
+		ID: "gc-source",
+		Metadata: map[string]string{
+			beadmeta.RoutedToMetadataKey: "worker",
+		},
+	}
+	fragment := &formula.FragmentRecipe{
+		Name: "expansion",
+		Steps: []formula.RecipeStep{{
+			ID:    "expansion.work",
+			Title: "Work independently",
+			Metadata: map[string]string{
+				beadmeta.ContinuationGroupMetadataKey: "stale-group",
+				beadmeta.SessionAffinityMetadataKey:   "require",
+			},
+		}},
+	}
+
+	if err := decorateDynamicFragmentRecipe(fragment, source, store, cfg.Workspace.Name, "", cfg); err != nil {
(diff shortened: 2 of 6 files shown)


quad341:
Maintainer note (hold triage): this PR is held for the same reason as #5584, explained in full there. Sorry that nobody said so at the time.

This branch carries its own copy of #5584's routing change (53236ede68). On one_shot routes it clears a formula-declared continuation group, where `main` now refuses that case loudly (#6360, 6e9a24a559). That's the same `ApplyGraphRouteBinding` branch, which is why this conflicts too. To clear the hold, please rebase onto current `main` and keep `main`'s version of that branch. The rig-scoped route-target fix (d3dc9e58fd) is what this PR adds on top; if #5584 lands first, that is the only part this one still needs.

Once a rebased head is up, the review pipeline will pick it up again. Thanks for your patience.

PR #5548, opened 2026-08-23
fix(dispatch): resolve class store refs as city control scope

jacobhausler:
# fix(dispatch): resolve class store refs as city control scope

## Summary

Use the canonical `storeref.ScopeRigContext` contract when desired-state route
repair selects a control dispatcher. Relocated class bindings such as
`class:gmnos` are city scope, while canonical `rig:<name>` refs remain rig
scope; legacy bare rig labels remain compatible.

## Mechanism

The old `controlDispatcherRigContextForStoreRef` in
`cmd/gc/build_desired_state.go:5186-5197` treated every non-city value as a rig
by stripping only the `rig:` prefix. That turned `class:gmnos` into a phantom
rig context and caused `desiredRoute` (`:5088-5105`) to suppress valid city
dispatcher demand when no such rig dispatcher was configured.

The candidate delegates canonical refs to `storeref.ScopeRigContext`, whose
contract at `internal/storeref/storeref.go:39-67` explicitly maps class
bindings to city scope. The diagnostic label uses the same resolver. Existing
legacy bare-rig compatibility is retained intentionally.

## Changes

- `cmd/gc/build_desired_state.go`: canonical scope resolution and consistent
  city/rig diagnostic labeling.
- `cmd/gc/build_desired_state_gmnos_test.go`: RED-first direct scope table and
  full desired-state regression. It proves `class:gmnos` preserves its route,
  produces exactly one city dispatcher demand/desired session, produces no rig
  demand, and that both canonical `rig:fixture` and legacy bare `fixture`
  remain rig-scoped.

No production config, formulas, stores, sessions, or binaries were touched.

## Validation

Environment: macOS Darwin arm64; Homebrew ICU `icu4c@78`.

```sh
CGO_ENABLED=1 \
CGO_CPPFLAGS='-I/opt/homebrew/opt/icu4c@78/include' \
CGO_LDFLAGS='-L/opt/homebrew/opt/icu4c@78/lib' \
go test ./cmd/gc \
  -run 'TestControlDispatcherRigContextForStoreRefUsesCanonicalScope|TestBuildDesiredStateClassControlWorkWakesExactlyOneCityDispatcher' \
  -count=1 -v
```

Result: PASS (`cmd/gc`, 1.813s), including the legacy bare-rig subtest.

```sh
CGO_ENABLED=1 \
CGO_CPPFLAGS='-I/opt/homebrew/opt/icu4c@78/include' \
[...]

PR #5588, opened 2026-08-25
fix(dispatch): class-binding candidates are authoritative for rig-rooted control rows

### LINKED ISSUE #5587: class-binding candidates rejected for rig-rooted control rows, dropped before route repair/demand
# class-binding candidates rejected for rig-rooted control rows, dropped before route repair/demand

## Environment

- gascity, `cmd/gc/build_desired_state.go` (`collectOpenUnassignedRoutedWork`,
  `rootStoreRefMatchesCandidate`)
- Follow-up to `e7e7427a5` ("resolve class store refs as city control scope"),
  which fixed the matching **candidate ref** case for repair but left a
  matching gap one level upstream, in the read that decides whether a row is
  collected at all
- Affects any control row physically relocated into a `class:*` binding
  (`gc storage migrate`) while still carrying a pre-migration logical
  `gc.root_store_ref=rig:<rig>`

## Repro (minimal)

1. Bind an infra class to a `binding` store distinct from the city `work`
   store (`registerResidencyRoutes`).
2. Create a control bead **in `binding`** with
   `gc.root_store_ref=rig:fixture`, `gc.routed_to=fixture/core.control-dispatcher`.
3. Configure city + rig (`fixture`) control-dispatcher agents; run desired
   state with `binding` as the leading store.

## Expected

The row is collected, its route repaired to the city dispatcher, and
`openControlDispatcherDemand` sees it — trace resolves `pending=false`.

## Actual

`rootStoreRefMatchesCandidate` required `candidateRig == rootRig`, and a
`class:*` ref reports `rigContext=""` (`storeref.ScopeRigContext`), so the
binding candidate never matched a rig-rooted `rootStoreRef`. On the runtime
plane `storeref.Narrow` drops every leg but the binding once a plan touches
one, so the binding is the row's only leg — it was silently dropped before
`repairControlDispatcherRoutesForStoreScope` or `openControlDispatcherDemand`
ever saw it. Symptom: city dispatcher trace `pending=false` forever; workflow
steps depending on the logical step never become ready.

## Root cause

A class binding is a distinct physical store, not a duplicate view of a rig
or city file store — the collapsing risk the rig-identity gate exists to
guard against (two handles aliasing one legacy file store) does not apply to
it. The gate needs a `storeref.IsClassRef` carve-out making a class
candidate authoritative regardless of rig identity.

## Fix

See fix on branch `fix/scoped-one-shot-lifecycle` (commit `56a8e8ff4`).
Independent of the `one_shot` lifecycle fixes on the same branch/estate
(`e356ea793`, `9ac41cad1`) — different subsystem, no overlapping files.

jacobhausler:
# fix(dispatch): class-binding candidates are authoritative for rig-rooted control rows

Fixes #5587

**Note:** `56a8e8ff4` depends on `controlDispatcherRigContextForStoreRef` already resolving class refs to city scope, which was introduced by an earlier unedited commit, `e7e7427a5` ("fix(dispatch): resolve class store refs as city control scope"), that had not yet reached `origin/main`. This branch/PR therefore includes both commits, unedited, so the new regression test in `56a8e8ff4` actually passes against current main. Confirmed RED without `e7e7427a5` (repaired route stayed `fixture/core.control-dispatcher` instead of `core.control-dispatcher`) and GREEN with both commits.

Base: `main`
Commit: `56a8e8ff4789d8a9c2e11177edfff6e78a25c628`

## Summary

`collectOpenUnassignedRoutedWork`'s duplicate-view filter,
`rootStoreRefMatchesCandidate`, rejected a `class:*` candidate for any row
whose `gc.root_store_ref` was rig-scoped. On the runtime plane a plan that
touches a binding narrows to the binding as its ONLY leg
(`storeref.Narrow`), so a rig-logical control row a migration relocated
into the binding had no other leg to be read through and was silently
dropped before route repair or demand ever saw it — city dispatcher trace
`pending=false` forever, dependent workflow steps never ready.

## Fix

- `rootStoreRefMatchesCandidate` now treats any `storeref.IsClassRef`
  candidate as authoritative for a scoped `rootStoreRef`, regardless of rig
  identity. A class binding is a distinct physical store (`gc storage
  migrate` relocates rows *into* it), never a duplicate view of a rig/city
  file store, so the legacy-unscoped collapsing risk this filter guards
  against does not apply to it. No rig names in the fix — the carve-out is
  purely `IsClassRef`.
- `e7e7427a5` already made `controlDispatcherRigContextForStoreRef` treat a
  `class:*` candidate ref as city scope for repair's own routing decision;
  this closes the matching gap one level upstream, in the read that decides
  whether the row is collected at all.

## Test plan

`cmd/gc` — new:
- `TestBuildDesiredState_RepairsClassBoundRigRoutedControlWork`: RED
  pre-fix (route never touched — row dropped in collection, not repair),
  GREEN post-fix. End-to-end through repair and demand.
- `TestBindingRefNormalizesOntoTheCityScope`
  (`census_residency_test.go`): pinned control assertion
  `rootStoreRefMatchesCandidate("rig:alpha", classRef)` flipped from
  `false` to `true` (the exact case this bug reports); new control case
  `rig:alpha` vs `rig:bravo` (both non-class) proves the rig-identity gate
  still applies between two ordinary legacy candidates.

`go test ./cmd/gc/... ./internal/storeref/...`, `go vet ./cmd/gc/...
./internal/storeref/...`, `go build ./cmd/gc` all clean. Full-package
`./cmd/gc/...` run confirmed pre-existing unrelated failures only
(`TestClassifyProductMetrics*`, nil-map panic, reproduced identically on
unmodified base — not introduced by this change).

## Compatibility

Only two callers of `rootStoreRefMatchesCandidate` exist total (the
production site plus its own unit test); downstream consumers of
`collectOpenUnassignedRoutedWork`'s output already handle a `class:*`-scoped
storeRef correctly and are untouched.

## Stack

Independent of `fix/one-shot-pool-independent-steps` (`e356ea793`) and
`fix/scoped-one-shot-lifecycle` (`9ac41cad1`) — different subsystem, no
overlapping files, observed in the same estate. Targets `main` directly.

### DIFF
--- a/cmd/gc/build_desired_state.go
+++ b/cmd/gc/build_desired_state.go
@@ -4736,9 +4736,29 @@ func normalizeDemandStoreRef(storeRef string) string {
 // through the city candidate and rig-owned rows only through their named rig.
 // Legacy rows without a canonical ref remain visible through every independent
 // store so same-ID beads in genuinely separate stores are not collapsed.
+//
+// A class-binding candidate is a special case of "authoritative", not an
+// exception to it. The duplicate-view risk this filter exists for is legacy
+// unscoped file-store mode, where two DIFFERENT store handles alias the same
+// physical file; a relocated class binding is never that — it is a distinct
+// physical store `gc storage migrate` moves rows INTO, and `gc storage status`
+// proves it converged. So a row the binding physically serves is never also
+// visible through a second candidate to defer to, city- or rig-logical alike:
+// on the runtime plane Narrow (internal/storeref/relevance.go) drops every
+// leg but the binding once a plan touches one, so it is the only leg that will
+// ever offer the row; on the reconcile plane the binding still holds the row
+// no legacy rig/city store does post-migration. Gating it on rootRig the way a
+// legacy-store candidate is gated would silently drop every rig-logical row a
+// migration relocated into the binding — the row is real, open, routed, and
+// permanently invisible to demand and route repair. e7e7427 already mapped the
+// binding's OWN ref onto city scope for this same reason; this closes the
+// matching gap for rows the binding serves under a still-rig-logical root.
 func rootStoreRefMatchesCandidate(rootStoreRef, candidateStoreRef string) bool {
 	rootRig, rootScoped := storeref.ScopeRigContext(rootStoreRef)
 	if !rootScoped {
+		return true
+	}
+	if storeref.IsClassRef(candidateStoreRef) {
 		return true
 	}
 	candidateRig, candidateScoped := storeref.ScopeRigContext(candidateStoreRef)
--- a/cmd/gc/build_desired_state.go
+++ b/cmd/gc/build_desired_state.go
@@ -5185,15 +5185,19 @@ func configuredControlDispatcherRouteForScope(cfg *config.City, rigContext strin
 
 func controlDispatcherRigContextForStoreRef(storeRef string) string {
 	storeRef = strings.TrimSpace(storeRef)
-	if storeRef == "" || storeRef == "city" || strings.HasPrefix(storeRef, "city:") {
+	if storeRef == "" || storeRef == "city" {
 		return ""
 	}
+	if rigContext, ok := storeref.ScopeRigContext(storeRef); ok {
+		return rigContext
+	}
+	// Keep accepting the pre-canonical bare rig label used by older census
+	// snapshots while canonical refs are interpreted by storeref above.
 	return strings.TrimPrefix(storeRef, "rig:")
 }
 
 func controlDispatcherStoreRefLabel(storeRef string) string {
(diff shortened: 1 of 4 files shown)


julianknutsen:
Superseded by #5918, which carries this commit unchanged (1868253bc5 → cb49439bf9) beneath a correction. Your collection carve-out (a class candidate is authoritative for a rig-rooted row) is right and is kept; the repair target it exposed was not — with #5548 it rewrote fixture/core.control-dispatcher to core.control-dispatcher, which on a split city hands every rig's control work to the city dispatcher. #5918 keeps the rig route and adds a lockstep test against the minter. Closing this one so the pipeline reviews a single PR.

PR #4856, opened 2026-07-30
fix(storehealth): bound LastMaintenance's event-log scan (#4418)

rjgeng:
## Summary

`gc status` scanned the entire `events.jsonl` twice on every call to populate the cosmetic `Last GC:` field, since `storehealth.LastMaintenance` called the unbounded `ep.List(Filter{Type: ...})` instead of the `TailProvider` fast path other callers already use (`internal/api/huma_handlers_events.go`'s `fetchEventPageAscending`, the fix for #1987). On a large city (263 MB / 327k events, never emitted a maintenance event) this cost ~3.3s of the ~4.8s user CPU in `gc status`, and produced a spurious "runtime status probe timed out; using partial status" warning on every run.

Fixes #4418.

## Root cause and fix

Switching to `ListTail` alone doesn't fully fix it: `readFilteredTailFromFile`'s loop condition (`for end > 0 && len(reversed) < limit`) means that when the filtered `Type` never matches near EOF — the common case, since store maintenance may never have run — the backward walk degrades to a full-file scan anyway, identical in cost to the forward scan it replaces.

This adds `Filter.MaxScanBytes` (same shape as the existing `Limit`/`AfterSeq`/`BeforeSeq` fields; zero value = unbounded, no effect on any other caller) to cap that backward walk. `LastMaintenance` now type-asserts `TailProvider` and passes a bounded `MaxScanBytes`; "not found within the window" is already a valid, correctly-rendered result since `Health.LastGCAt` is `omitempty`. Providers without `TailProvider` (e.g. an exec-script provider) keep the original unbounded `List` path, unchanged.

## Testing

- `go build ./...` — clean
- `go test ./internal/events/... ./internal/storehealth/...` — full suite green, including two new regression tests:
  - `TestReadFilteredTailMaxScanBytesBoundsBackwardWalk` — proves the bound actually stops the backward walk before reaching a match placed outside the window (and that an unbounded call still finds it, ruling out a false negative from the filter itself)
  - `TestLastMaintenanceUsesTailProviderFastPath` / `TestLastMaintenanceFallsBackWithoutTailProvider` — proves `LastMaintenance` takes the `ListTail` path when available (via a call-counting wrapper) and still returns the correct answer via the fallback when it isn't
- `go test ./cmd/gc/... ./internal/api/...` (the two other `LastMaintenance` call sites) — existing store-health tests pass unchanged, no signature change so no caller edits needed
- `go vet ./internal/events/... ./internal/storehealth/...` — clean
[...]

PR #5589, opened 2026-08-25
perf(events): seek past the cursor instead of rescanning the active log

atbrace:
## Summary

`readFilteredTracked` reads the **active** `events.jsonl` from byte 0 and `json.Unmarshal`s every line, then discards non-matching events in `matchesFilter`. Because `AfterSeq` is applied only *after* the unmarshal, a cursored read does no less work than an uncursored one — it re-parses the entire log on every call.

The supervisor pays this on every patrol tick. `dispatchOrders` → `memoryOrderDispatcher.dispatch` → `CheckTriggerWithOptions` → `checkEvent` calls `List{Type: "bead.closed", AfterSeq: cursor}` to serve a single event-triggered order. A live 30s CPU profile of the supervisor attributed **53% of all process CPU** to this one read path (`syscall.read` 35.9%, `bufio.Scanner.Scan` 34.5%, `json.Unmarshal` 18.6%) against a 121 MB / 87,831-line log.

Because the log grows monotonically, so does per-tick cost. This presents as supervisor CPU that climbs over *days* with no change in workload.

## Root cause

`internal/events/reader.go`, the active-file branch of `readFilteredTracked`:

```go
scanner := bufio.NewScanner(f)
for scanner.Scan() {
    var e Event
    if err := json.Unmarshal(scanner.Bytes(), &e); err != nil { continue }
    if !matchesFilter(e, filter) { continue }   // <- AfterSeq applied HERE, after the decode
```

The `.gz` archive branch immediately above it already has a seq-window skip — `archiveOverlapsFilter` prunes an archive whose `LastSeq <= AfterSeq` without gunzipping it. **The active file simply never got the same treatment.** This PR restores that symmetry rather than introducing a new mechanism.

## Fix

`activeScanStart` binary-searches the byte offset of the first line whose `seq > AfterSeq` and starts the scan there (~27 probes on a 121 MB log). `seqLineAt` probes an offset and returns the seq of the next parseable line plus that line's true start offset; for `off > 0` it discards the partial line it landed inside, which is the contract the search depends on.

Soundness: the active log is append-only with strictly increasing seq — `FileRecorder` assigns `e.Seq = r.seq++` under its mutex — so every line at or below the cursor sorts before every line above it and can be skipped wholesale.

Correctness is preserved by construction, three ways:
- the offset is only ever used as a *scan start*; `matchesFilter` still runs on every line actually read, so the result set is unchanged;
- if the boundary cannot be established (unparseable head, seek error, ordering invariant violated) `activeScanStart` returns 0 and degrades to the previous full scan — a log that breaks the invariant loses speed, never correctness;
- add-only. No existing line is deleted or modified apart from the one `f.Seek` that positions the scanner.

No persistent or shared state is deleted or mutated; this changes only where a read starts.

## Tests

`internal/events/reader_afterseq_seek_test.go` (new):
- `TestActiveScanStart` — boundary cases: zero cursor, cursor below the first seq, cursor at and beyond the last seq, and interior cursors landing *exactly* on the first qualifying line start.
- `TestReadFilteredAfterSeqEquivalence` — asserts the seek path returns results identical to a full scan for **every** cursor `0..n+1` across three filters. This is the load-bearing test: the fix is a pure optimization, so the property that matters is that it changes nothing observable.
- `TestReadFilteredAfterSeqMalformedLines` — the scanner's long-standing tolerance for malformed rows survives probing.

`go test ./internal/events/` and `go vet ./internal/events/` pass on this branch, rebased onto `main` at 3e686dace. On unpatched `main` the test file does not build (`undefined: activeScanStart`) — worth stating plainly: because this is a performance fix with identical semantics, no *correctness* test can fail on `main`. The behavioural evidence is the measurements below.

## Validation

Measured against a real 121 MB / 87,831-line production log, **match counts identical** in every case:

| cursor position | before | after | |
|---|---|---|---|
| 50 events behind head (steady state) | 1565.7 ms | 1.9 ms | **842x** |
| 5000 events behind (catch-up) | 1118.3 ms | 142.2 ms | 7.9x |

Deployed to a live city, before → after:

| | before | after |
|---|---|---|
| dispatch duration, mean | 3953 ms | 1345 ms (−66%) |
| dispatch duration, p90 | 6973 ms | 1983 ms (−72%) |
| dispatch duration, max | 15675 ms | 2396 ms (−85%) |
| supervisor steady state | ~52.4% of a core | ~8–10% of a core |

A live CPU profile after the change shows `dispatchOrders` down 76%, with `ReadFiltered` / `CheckTriggerWithOptions` / `json.Unmarshal` **absent from the profile** rather than merely reduced. A 2.68 h post-deploy watch showed no regrowth, with dispatch *rate* and *duration* both flat — i.e. the win is not an artifact of the city going quiet.

## Related

- **#5445** (open) fixes the mirror-image defect one branch over: `streamArchive` takes the `Filter` as `_`, so a cursor landing *inside* an archive's seq window decodes the whole archive. That is the archive branch; this is the active branch. The two are complementary and, as far as I can see, do not conflict — together they close the `AfterSeq` path end to end. Happy to rebase on top of it if you would rather land them in order.
- **#5191** (open) reports the same read path from the zero-cursor angle. This PR does not address that: `AfterSeq == 0` means "no cursor", `activeScanStart` returns 0, and the full scan is correct and unchanged. The semantics question that issue raises about what a zero cursor *should* mean for `checkEvent` is untouched here.
- **#4418 / #4856** bounded the same shape for `gc status` via `Filter.MaxScanBytes`. That bound is a *cap*; this is a *seek* — it can skip the prefix exactly because the cursor makes it provably irrelevant, so no events are dropped.

---
- Related to #2604 — reduces one component of the order-dispatch phase that issue measures — the per-tick full rescan of the active events log in the event-trigger check, with dispatch duration on a live city going from ~3.9s mean to ~1.3s — but does not change the synchronous-on-tick-path dispatch design, the fork-per-operation bd store, or the per-rig order fan-out identified there as root causes

### DIFF
--- a/internal/events/reader.go
+++ b/internal/events/reader.go
@@ -100,6 +100,87 @@ func ReadAll(path string) ([]Event, error) {
 	return ReadFiltered(path, Filter{})
 }
 
+// seqProbeBufSize bounds the read-ahead used when probing a byte offset for the
+// line that begins there. Events average ~1.4 KB on a live log; 64 KB
+// comfortably spans one line plus the partial line a probe may land inside.
+const seqProbeBufSize = 64 * 1024
+
+// seqLineAt probes byte offset off and returns the seq of the next parseable
+// line, along with that line's start offset.
+//
+// Contract, which the binary search in activeScanStart depends on: for off > 0
+// the line containing off is treated as a partial line and DISCARDED, so the
+// result is the first parseable line beginning strictly after off. Only off == 0
+// reads the line starting at off itself. Probing a known line start therefore
+// returns the FOLLOWING line, not the line at that offset.
+//
+// start is always a true line boundary. Unparseable lines are skipped, matching
+// the scanner's long-standing tolerance for malformed rows. ok is false when EOF
+// is reached without finding a parseable line.
+func seqLineAt(f *os.File, off int64) (seq uint64, start int64, ok bool) {
+	if _, err := f.Seek(off, io.SeekStart); err != nil {
+		return 0, 0, false
+	}
+	r := bufio.NewReaderSize(f, seqProbeBufSize)
+	pos := off
+	if off > 0 {
+		partial, err := r.ReadBytes('\n')
+		if err != nil {
+			return 0, 0, false
+		}
+		pos += int64(len(partial))
+	}
+	for {
+		line, err := r.ReadBytes('\n')
+		if len(line) > 0 {
+			var probe struct {
+				Seq uint64 `json:"seq"`
+			}
+			if json.Unmarshal(bytes.TrimRight(line, "\r\n"), &probe) == nil && probe.Seq > 0 {
+				return probe.Seq, pos, true
+			}
+			pos += int64(len(line))
+		}
+		if err != nil {
+			return 0, 0, false
+		}
+	}
+}
+
+// activeScanStart returns the byte offset in the active log at which a scan can
+// begin without missing any event Filter.AfterSeq can match.
+//
+// The active log is append-only with strictly increasing seq (FileRecorder
+// assigns e.Seq = r.seq++ under its mutex), so every line at or below the
+// cursor sorts before every line above it and can be skipped wholesale. This is
+// the seq-window skip archiveOverlapsFilter already applies to .gz archives,
+// extended to the active file. Without it a cursored read re-parses the entire
+// log on every call and AfterSeq saves nothing, because matchesFilter applies
+// it only after each line has been unmarshalled -- the dominant cost of the
+// supervisor's per-tick order-trigger check (gcy-ocb5).
+//
+// Returns 0 (full scan) whenever the boundary cannot be established, so a log
+// that violates the ordering invariant degrades in speed, never in correctness.
+func activeScanStart(f *os.File, size int64, afterSeq uint64) int64 {
+	if afterSeq == 0 || size <= 0 {
+		return 0
+	}
+	// Nothing to skip when the log's first line is already above the cursor.
+	if first, _, ok := seqLineAt(f, 0); !ok || first > afterSeq {
+		return 0
+	}
+	i := sort.Search(int(size), func(i int) bool {
+		seq, _, ok := seqLineAt(f, int64(i))
+		return !ok || seq > afterSeq
+	})
+	seq, start, ok := seqLineAt(f, int64(i))
+	if !ok || seq <= afterSeq {
+		// Cursor is at or beyond the newest event: skip the file entirely.
+		return size
+	}
+	return start
+}
+
 // ReadFiltered reads events from path and sibling archives, returning
 // only those matching all non-zero fields in filter. Archives whose
 // seq window is fully excluded by the filter's AfterSeq predicate are
@@ -166,6 +247,17 @@ func readFilteredTracked(path string, filter Filter) ([]Event, map[eventSeqWindo
 		return result, listed, fmt.Errorf("reading events: %w", err)
 	}
 	defer f.Close() //nolint:errcheck // read-only file
+
+	// Skip the prefix of the active log the cursor already covers. seqLineAt
+	// leaves the descriptor wherever its last probe ended, so seek explicitly
+	// even when the scan starts at 0 (the uncursored case).
+	scanFrom := int64(0)
+	if st, statErr := f.Stat(); statErr == nil {
+		scanFrom = activeScanStart(f, st.Size(), filter.AfterSeq)
+	}
+	if _, err := f.Seek(scanFrom, io.SeekStart); err != nil {
+		return result, listed, fmt.Errorf("seeking events: %w", err)
+	}
 
 	scanner := bufio.NewScanner(f)
 	scanner.Buffer(make([]byte, 0, 64*1024), 1024*1024) // handle lines up to 1MB
--- a/internal/events/reader_afterseq_seek_test.go
+++ b/internal/events/reader_afterseq_seek_test.go
@@ -1,0 +1,169 @@
+package events
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+// writeSeqLog writes a monotonic append-only log of n events and returns its path.
+// Every other event is type "b" so filters have something to discriminate on.
+func writeSeqLog(t *testing.T, n int) string {
+	t.Helper()
+	dir := t.TempDir()
+	path := filepath.Join(dir, "events.jsonl")
+	var sb strings.Builder
+	for i := 1; i <= n; i++ {
+		typ := "a"
+		if i%2 == 0 {
+			typ = "b"
+		}
+		fmt.Fprintf(&sb, `{"seq":%d,"type":%q,"ts":"2026-08-25T00:00:00Z","actor":"t","subject":"s%d"}`+"\n", i, typ, i)
+	}
+	if err := os.WriteFile(path, []byte(sb.String()), 0o644); err != nil {
+		t.Fatalf("write log: %v", err)
+	}
+	return path
+}
+
+// TestActiveScanStart pins the seq->offset boundary search on the active log.
+// The append-only log has strictly increasing seq, so every line at or below
+// AfterSeq can be skipped wholesale; activeScanStart must land exactly on the
+// first line above the cursor and never before it (which would be slow) or
+// after it (which would silently drop events).
+func TestActiveScanStart(t *testing.T) {
+	const n = 500
+	path := writeSeqLog(t, n)
+	f, err := os.Open(path)
+	if err != nil {
+		t.Fatalf("open: %v", err)
+	}
+	defer f.Close()
+	st, err := f.Stat()
+	if err != nil {
+		t.Fatalf("stat: %v", err)
+	}
+	size := st.Size()
+
+	// AfterSeq=0 means "no cursor": start at byte 0.
+	if got := activeScanStart(f, size, 0); got != 0 {
+		t.Fatalf("activeScanStart(afterSeq=0) = %d, want 0", got)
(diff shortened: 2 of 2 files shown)


quad341:
@atbrace — this is an excellent bug report attached to an excellent fix, and the framing sold me before the code did: the `.gz` branch already had the seq-window skip via `archiveOverlapsFilter`, and the active file just never got it. Restoring a symmetry beats inventing a mechanism.

`TestReadFilteredAfterSeqEquivalence` is what I want to single out. Sweeping every cursor `0..n+1` against a full-scan oracle is exactly right for a pure optimization — you worked out that the property worth testing is "changes nothing observable," then actually tested it instead of testing that the fast path is fast. And you said plainly in the PR that no correctness test can fail on `main`. That honesty about what your own evidence does and doesn't prove is rarer than it should be. I re-derived the `sort.Search` boundary myself, including the `off == 0` discontinuity: it lands exactly on the first qualifying line, independent of line width.

**One thing needs fixing and it's ours:** `defer f.Close()` at `internal/events/reader_afterseq_seek_test.go:45` trips `errcheck` — the package convention is `//nolint:errcheck`, and I'll add it on my side. I'll also make one duplicated assertion in `TestActiveScanStart` exercise the `first > afterSeq` guard it was aiming at (as written it repeats the `AfterSeq == 0` case, because `AfterSeq` is unsigned).

**On the red CI — none of it is yours.** The current failures are `Worker core phase 2 (Claude)`, `(Gemini)`, and `Evidence / critical-path suites`, and all of them are the Go module proxy falling over:

```
github.com/steveyegge/beads@v1.1.1-...: read "https://proxy.golang.org/.../@v/....mod":
  stream error: stream ID 71; INTERNAL_ERROR; received from peer
FAIL github.com/gastownhall/gascity/internal/worker/workertest [setup failed]
```

`[setup failed]` means the package never compiled, so no test in it ever ran — and the Gemini variant failed identically on a different module (`dolthub/flatbuffers`) with a different stream ID. `main` has been intermittently red on the same jobs today, independent of this PR. I've re-run them.

No need to rebase on #5445 — different branch of the same function, they'll land cleanly in either order. Thanks for chasing this all the way to a profile; the before/after dispatch numbers made the case on their own.

---
**Reviewers**: Qwen (local) · Claude (claude-opus-5[1m]) · Codex (gpt-5.6-sol) — all three returned `fix-merge`, high confidence, no dissent.

quad341:
Correction to my last comment, and the credit goes to you: **both fixes I offered to make, you had already made.** `b933d8f0` carries the `//nolint:errcheck` annotation on the deferred `Close` — on *both* handles, including the second one I hadn't mentioned — and adds `writeSeqLogFrom(t, 5000, 100)` so the rotated-log case actually reaches the `first > afterSeq` guard instead of repeating the `AfterSeq == 0` path. That is exactly the gap, fixed exactly the way I was going to fix it, before I got there. My review had gone out against the prior head; nothing on this branch is waiting on me.

Verified independently on `b933d8f0` rather than taken from the commit message: `gofmt -l internal/events/` clean, `go vet ./internal/events/` clean, `go test -count=1 ./internal/events/` → `ok … 9.144s`.

**And the CI is green.** After the re-run: **67 pass, 0 fail**, 15 skipped, with `Evidence / critical-path suites` and `Worker core phase 2 summary` still finishing. `Preflight / static checks` passes, which was the one genuinely PR-caused failure and is now yours-fixed. The module-proxy failures cleared on re-run without a code change, which is the confirmation that they were transport flakes rather than anything in this diff.

Nothing outstanding from your side.

quad341 (approved):
Approving. All three intake reviewers (Qwen, Claude, Codex) returned `fix-merge` at high confidence with no dissent, and the two items in the fix plan were already handled by @atbrace in b933d8f0 before we got there.

Verified independently on the merge head: `gofmt -l internal/events/` clean, `go vet ./internal/events/` clean, `go test -count=1 ./internal/events/` → ok. CI is 70 pass / 0 fail after re-running the module-proxy flakes.

The soundness argument holds: the active log is append-only with strictly increasing seq assigned under `FileRecorder`'s mutex, so every line at or below the cursor sorts before every line above it; `activeScanStart` is only ever used as a scan start and `matchesFilter` still runs on every line read, so the result set is unchanged by construction. Degrading to 0 on an unestablishable boundary makes a broken log lose speed rather than correctness.

sjarmak:
I went looking for a causal link between this change and the `TestE2E_SuspendResume_City` failures that have been running at roughly 74% on `main`, and I did not find one. Recording the result here because the bisect window pointed straight at this commit and someone else will land in the same place.

## The commit is excluded

`c3b12fb91` is the only commit between the last green run and the first red one (`git rev-list --count 3e686dace..c3b12fb91` is 1), which is what made it the obvious suspect. It does not survive a causation test:

| Tree | `activeScanStart` present | Result |
|---|---|---|
| `main` @ `2fab51b3c` | yes | FAIL, `e2e_lifecycle_test.go:184`, 93.3s |
| `3e686dace`, the parent | no, function does not exist | FAIL, same line, same signature, 93.8s |
| `main` with `activeScanStart` returning 0 | dead-coded | FAIL, same line, same signature, 93.4s |

The test fails identically on the commit CI called last-clean, and with the new seek path disabled. The environment is not broadly broken: `TestE2E_Kill` and `TestE2E_StartIsIdempotentForSupervisorManagedCity` pass on `main` in the same run.

Instrumenting `activeScanStart` during a failing run gives the same answer from the other direction: 266 calls, none of which took the `return size` skip-the-file branch. Every call returned a real interior offset, with `afterSeq` between 13 and 38 against a roughly 25 KB log. The branch that could have caused this never fired.

There is also no code path for the theory that made the commit plausible. The assertion is a 90 second poll on `.gc-reports/`, and the controller learns about resume from `.gc/runtime/suspension-state.json`, not from the event log, so "the controller missed the resume event" has nothing behind it.

## A real defect in the same function, worth fixing on its own

The doc comment says a log that violates the ordering invariant "degrades in speed, never in correctness." That is not true. `sort.Search` requires a false…false,true…true predicate, and on a non-monotonic active log the search returns a meaningless index and `ReadFiltered` silently drops events:

```
seqs=[1 2 3 100 4 5 6]  after=4:  got [5 6]  want [100 5 6]
seqs=[10 11 12 1 2 3]   after=10: got []     want [11 12]
```

The second shape is the bad one: one out-of-order run near the front makes every probe return false, `sort.Search` returns `size`, the `seq <= afterSeq` branch discards the whole file, and that cursor never sees those events again. The same case passes on `3e686dace`, where the full scan has no ordering assumption to violate, so this is a genuine regression, just a latent one rather than the CI failure.

Reachability looks low. `writeRecordLocked` re-reads the on-disk tail inside a cross-process `flock` and only ever raises `r.seq` (`recorder.go:377-386`, `:246-259`), and rotation continues the seq space through the `events.rotated` anchor rather than restarting (`recorder.go:461-552`). That leaves a torn or short write, and a stale post-rotation fd in a second process. Worth noting on the first of those: `writeRecordLocked` (`recorder.go:394-398`) does not check for a short write, while `writeBatch` (`:364-370`) does.

Two things that would close the gap without giving up the seek: check the short write in `writeRecordLocked`, and make the skip-the-file branch conservative, falling back to a full scan when the probe result is inconsistent rather than trusting it. Happy to open a PR for either if you would rather not carry it.

## Where the e2e failure actually is

Unresolved. The commit is ruled out, so the remaining candidates are that the test was already failing and the green run on `3e686dace` did not actually execute this tier, or something in CI infrastructure or timing landed in the same window. #5226 tracks the general flakiness but predates this convergence onto a single test. `gc supervisor stop: supervisor is not running` shows up in cleanup on passing runs too, so it is noise, but supervisor lifecycle around suspend and resume is the next place I would look.

sjarmak:
Filed the hardening item as #5650 so it does not live only in this thread.

quad341:
Two corrections to what I wrote on this PR. Both are mine to own, and one of them takes credit away from me and gives it to a reviewer.

## 1. I misattributed `b933d8f0` to @atbrace — twice. It was our automation.

In a follow-up comment I wrote "both fixes I offered to make, you had already made," and my approval repeated it: "the two items in the fix plan were already handled by @atbrace in `b933d8f0` before we got there."

That is wrong. `b933d8f0` was written by our own fix-executor. Our run artifact records it explicitly (`pr-5589/runs/20260825T141448Z/fix-executor-commit.sha` → `b933d8f0ea7a365a1548d0992102630a765320a5`), and the same is true of `a3b22654` on #5590.

The reason I got it wrong is worth stating plainly, because it will recur: **our fix-executor sets `GIT_AUTHOR_*` and `GIT_COMMITTER_*` to the contributor by design**, to preserve authorship when we push a fix onto a contributor's branch. So `git show` reports "Austin Brace" for a commit our pipeline authored. I read the author field, concluded @atbrace had acted, and said so publicly without checking our own run record.

@atbrace — the PR, the diagnosis and the fix are yours, and my original assessment of them stands. The two follow-up items were not; I handed you credit for our robot's work, which is its own kind of disservice. Sorry about that.

## 2. My approval asserted something false about `activeScanStart`, and @sjarmak caught it.

I wrote that a log violating the ordering invariant "degrades to a full scan." It does not.

In `activeScanStart`, the `sort.Search` predicate treats an unparseable line (`!ok`) as sorting above the cursor. When the line at the resulting offset is itself unparseable, control reaches:

```go
seq, start, ok := seqLineAt(f, int64(i))
if !ok || seq <= afterSeq {
    return size
}
```

`!ok` returns `size` — which **skips the remainder of the file**, silently dropping events. That is a correctness failure, not a slowdown. The function's own doc comment ("degrades in speed, never in correctness") is inaccurate for the same reason.

All three intake models and I missed this. @sjarmak found it and filed #5650. That is the right outcome and the right place for it; I'm recording the correction here so this thread doesn't stand as the last word.

Issue #4922, opened 2026-08-02
supervisor: RSS ratchets to multi-GB high-watermark under load (10.6G peak observed) — no page return, no pprof endpoint to diagnose


edwarddavison:
gc 1.4.0 (build a7297c51), Linux, 9-rig city with per-scope dispatchers + NativeDoltStore, 62GB box.

**Observed:**
- Long-running supervisor: systemd reported `10.6G memory peak` / 2d4h CPU over a 33h unit lifetime under heavy load.
- Fresh restarts during a busy period balloon immediately: 104MB → 3.08GB in 7 min (no GOMEMLIMIT); a second run grew 0.14 → 2.85GB in 120s.
- `GOMEMLIMIT=2GiB` made it worse: heap demand exceeds it, GC death-spirals — CPU climbed 47% → 458% while RSS still blew past the limit.
- `GOMEMLIMIT=8GiB` + restart in a calm window: settles at **0.12GB RSS / 8% CPU** — completely healthy. Conclusion: busy-period allocations are never returned to the OS on a memory-rich box, so RSS ratchets to the workload high-watermark and dashboards alarm on it (heap-used was ~977MB when RSS showed 3.81GB).

**Asks:**
1. Ship a sane default `GOMEMLIMIT` (or set one relative to detected RAM) so RSS tracks real working set.
2. Expose a pprof endpoint (`/debug/pprof` returned 404 on the API port) — we could not distinguish goroutine leak vs allocator retention without it.
3. Consider basing the Health card's red threshold on heap-used rather than raw RSS (card showed heap 977MB / RSS 3.81GB as red).

Happy to run instrumented builds or provide more samples — this deployment reproduces the ratchet on demand under morning-backlog load.

PR #5590, opened 2026-08-25
perf(supervisor): stop building a heap profile nothing reads

atbrace:
## Summary

The supervisor spends ~23% of its on-CPU time building a heap profile that nothing reads.

Go enables heap profiling by default (`MemProfileRate = 512 KiB`). `runtime.mProf_Flush` walks the entire mprof bucket list once per GC cycle, called from `gcMarkTermination`. Bucket count only grows over a process lifetime, and this daemon runs for weeks.

Worth being precise about where the cost lands, because the profile invites the wrong reading: `mProf_Flush` is called from `gcMarkTermination` but *after* the world restarts, not during the pause.

```go
// $GOROOT/src/runtime/mgc.go:1514 (go1.26.5)
        startTheWorldWithSema(now, stw)
    })

    // Flush the heap profile so we can start a new cycle next GC.
    // This is relatively expensive, so we don't do it with the
    // world stopped.
    mProf_Flush()
```

So this is continuous CPU burn, not added GC pause time. That is enough on its own to justify removing it.

## Root cause

Not a bug in `gc` code — a default that is wrong for a long-lived daemon that never serves the profile.

Measured on a supervisor at 7d14h uptime with 12 live sessions. Steady state confirmed by CPU-time delta before sampling: `2896:39.43 → 2897:06.42` over 79s wall = 26.99 CPU-s = **34% of one core, continuous**, on an otherwise idle-ish city.

A 90s sample (the patrol interval is 60s, so a 3s sample lands inside or outside a tick by luck and is not usable here). Of 170,919 samples, 167,363 are idle leaves (`__psynch_cvwait`, `kevent`, `read`, `__wait4_nocancel`, `__semwait_signal`), leaving 3,556 on-CPU:

| symbol | samples | share of on-CPU |
|---|---|---|
| `runtime.mProf_Flush` | 802 | 22.6% |
| `runtime.bgsweep` | 798 | 22.4% |
| `runtime.gcBgMarkWorker` | 784 | 22.0% |
| `__fork` | 382 | 10.7% |
| `encoding/json` decode | ~300 | 8.4% |

Every sample attributed to `gcMarkTermination` is `mProf_Flush`: 802 of 802. The stack is unambiguous — `gcBgMarkWorker → gcMarkDone → gcMarkTermination → mProf_Flush → mProf_FlushLocked`.

## Fix

Set `runtime.MemProfileRate = 0` at the top of `runSupervisor` when `GC_PPROF != "1"`.

`gc` already serves pprof only under `GC_PPROF=1` (`api.StartPprof`, `internal/api/supervisor.go`). Reusing that existing gate rather than adding a second knob means:

- with `GC_PPROF` unset — the default — the profile is not built, because nothing could read it anyway;
- with `GC_PPROF=1`, `MemProfileRate` is left at the Go default and the heap profile is **fully available**, unchanged, for anyone diagnosing memory behaviour (e.g. #4922, which wants exactly that).

The knob is only ever *removed*, never added. Net configuration surface: zero.

Scope is deliberately just the supervisor, which is what was measured. Other long-lived `gc` processes may warrant the same treatment, but not on evidence collected here.

One known imprecision, since `MemProfileRate` is documented as change-it-once-and-early: the top of `runSupervisor` is not process-early. Telemetry init and the Cobra tree have already allocated by then (`cmd/gc/main.go:193`, `:281`). This produces no observable bug, because the only surface that could show the seam is off in exactly the mode where the rate is changed, but it is worth knowing if the gate ever moves.

## Tests

`cmd/gc/cmd_supervisor_runtime_test.go` covers the extracted `configureSupervisorRuntime` helper as a table: unset disables profiling, `GC_PPROF=1` leaves the Go default intact, and `GC_PPROF=true` is *not* an opt-in. It calls the production function rather than a copy, saves and restores the global via `t.Cleanup`, and uses `t.Setenv`. It fails on an inverted gate and on a gate loosened to `!= ""`.

`go build ./cmd/gc` and `go vet ./cmd/gc/` pass.

## Validation

The profile is the validation: the symbol is every sample of mark termination, the stack is a single unambiguous chain into the runtime, and the fix removes the only thing on it. There is no plausible path by which setting `MemProfileRate = 0` leaves `mProf_Flush` doing work.

No persistent or shared state is read, written, or deleted. The only observable behaviour change is that `runtime.MemProfile` returns nothing when `GC_PPROF` is unset — which was already unreadable, as no endpoint is served in that mode.

## Note on what this does *not* claim

This is not the whole of the supervisor's CPU. The `__fork` term above (10.7%) is a separate, still-open problem — an uncached per-lookup fork in the tmux path — and unlike this one it scales linearly with session count, so it dominates during a load excursion even though it is smaller at rest. I am not folding that into this PR; it needs a different fix and its own evidence.


jacobhausler:
LGTM — solid measurement-driven change, and gating on the existing `GC_PPROF` knob instead of adding a second one is the right call.

One observation for a possible follow-up: the nudge-poll sidecar (`configureNudgePollRuntime`, `cmd/gc/cmd_nudge.go`) has the identical shape — a per-invocation runtime knob (there, `debug.SetMemoryLimit`) gated behind the same `GC_PPROF`/pprof machinery. Per its own doc comment, one sidecar runs per active session, so at high session counts the aggregate cost of always-on heap profiling there is likely larger than the single supervisor's. Might be worth the same `MemProfileRate = 0` treatment there once it's measured.

Minor: `runSupervisor()` works fine here, but the textbook place for `MemProfileRate` is `main()` — the runtime docs say to set it "just once, as early as possible in the execution of the program (for example, at the beginning of main)," since it only affects allocations made after it's set. Probably negligible in practice given how little allocates before `runSupervisor` runs, but worth a thought if you want to be strict about it.

sjarmak (approved):
The measurement holds up and the fix is right. One claim in the description does not, and since the description becomes the commit message I want to correct it rather than merge it into history.

**The stop-the-world framing is wrong; the CPU cost is not.** The body says the flush is paid "inside stop-the-world GC mark termination, so the cost lands as pauses, not just cycles," and calls out "100% of stop-the-world mark termination in this process is memory-profile flushing." `mProf_Flush` is called from `gcMarkTermination`, but after the world is restarted:

```go
// $GOROOT/src/runtime/mgc.go:1514 (go1.26.5)
        startTheWorldWithSema(now, stw)
    })

    // Flush the heap profile so we can start a new cycle next GC.
    // This is relatively expensive, so we don't do it with the
    // world stopped.
    mProf_Flush()
```

The runtime moved it out of the pause deliberately and left a comment saying so. This is an easy read to get wrong, because a profiler attributes the samples to `gcMarkTermination` either way, and the frame name is what suggests STW. Your own code comment gets this right, which is what the title says the commit does; the body just did not follow it.

Nothing about the fix depends on the wrong half. 22.6% of on-CPU time in a flush nothing can read justifies the change on cycles alone, and cycles is what the profile actually measured. Correcting the body at merge; the code is untouched.

**On safety, three things I checked rather than assumed.** Nothing in the repo reads a heap profile outside the `GC_PPROF=1` path, so setting the rate to 0 cannot produce a silently-empty profile for a consumer that thinks profiling is on: the only non-test pprof surface is `api.StartPprof` (`internal/api/supervisor.go:379`), gated on the identical `!= "1"` comparison, and there is no non-test `WriteHeapProfile` or `pprof.Lookup("heap")` anywhere. No other enable path uses a looser check like `!= ""`, which would otherwise create a shape where pprof is on and the rate is 0.

The one caveat worth stating, since `MemProfileRate` is documented as change-it-once-and-early: the top of `runSupervisor` is not process-early. Telemetry init and the whole Cobra tree have already allocated by then (`cmd/gc/main.go:193`, `:281`). That is later than the doc's ideal, and it means a `GC_PPROF=1` profile would still cover everything from that point on, while allocations before it are already recorded at the default rate. It produces no observable bug on current source, because the only surface that could show the seam is off in exactly the mode where the rate is changed. Not blocking, worth knowing if someone later moves the gate.

Children inherit `GC_PPROF` but not this setting: only the supervisor run path calls `configureSupervisorRuntime`, and the other inherited-env children (`internal/session/submit.go:654`, `internal/workspacesvc/proxy_process.go:206`) are unaffected.

**The test answers the description's own open question.** The body says none was added and offers to extract a testable helper if asked. The head commit already does both. `cmd_supervisor_runtime_test.go` calls the production function rather than a copy, saves and restores the global via `t.Cleanup`, uses `t.Setenv`, and pins the case that matters most: `GC_PPROF=true` is not an opt-in. It would fail on an inverted gate and on a loosened `!= ""`. No `t.Parallel()`, so mutating the process global is contained. `go test ./cmd/gc -run '^TestConfigureSupervisorRuntime$'` passes.

The `__fork` term is correctly carved out rather than folded in.

Approving, with the body corrected on the way in.

sjarmak:
Edited the description before merging so the squashed commit does not record the stop-the-world framing; the code is untouched, and the measured numbers and the `__fork` carve-out are preserved as written.

Recording the cross-provider pass that backs the approval: Codex (gpt-5.4) reviewed head `a3b22654e` against five questions and returned no issues. What it verified independently:

- No non-test heap-profile consumer exists outside the `GC_PPROF=1` path. The only pprof surface is `api.StartPprof` (`internal/api/supervisor.go:379`), and there is no non-test `WriteHeapProfile` or `pprof.Lookup("heap")` anywhere in the repo, so a rate of 0 cannot silently hand someone an empty profile they believe is real.
- The gates agree exactly: `os.Getenv("GC_PPROF") != "1"` in both `cmd_supervisor.go:1247` and `supervisor.go:384`. No other enable path uses a looser `!= ""` or truthy parse, which would otherwise create a shape where pprof is on and the rate is 0.
- Children inherit `GC_PPROF` but not this setting. Only the supervisor run path calls `configureSupervisorRuntime`; `internal/session/submit.go:654` and `internal/workspacesvc/proxy_process.go:206` start children with inherited env and are unaffected.
- It independently raised the not-process-early point (telemetry init and the Cobra tree allocate before `runSupervisor`) and reached the same conclusion I did: later than the `runtime` docs' ideal, no observable bug on current source. That caveat is now in the description rather than only here.
- The test exercises the production function, restores the global via `t.Cleanup`, uses `t.Setenv`, and would fail on both an inverted gate and a `!= ""` loosening. No `t.Parallel()`, so the process-global mutation is contained. `go test ./cmd/gc -run '^TestConfigureSupervisorRuntime$'` passes.

PR #5591, opened 2026-08-25
Batch tmux session reads for agent listing

quad341:
## What this changes

`GET /v0/agents` now batches runtime session-roster and environment reads when a runtime exposes those capabilities. The tmux provider implements both paths, reducing repeated tmux subprocesses during the agent-list hot loop while preserving the existing per-session fallback for every other runtime.

Suspended-session metadata is now queried only for sessions that are actually running. Agent response fields and state semantics are unchanged.

The batching contracts are optional runtime interfaces rather than additions to the base provider interface. That keeps existing providers source-compatible and lets each runtime opt in only when it can perform a real batch operation.

## Review notes

- The tmux roster treats a missing tmux server as an empty roster and maps session name to running state.
- `LastActivity` intentionally remains on the existing per-session path because raw tmux activity can be stale for detached sessions.
- There are no configuration, endpoint, response-schema, generated-client, or migration changes.

## Test plan

- [x] Run the documented `make test-local-full-parallel` union with the rootless Podman test environment enabled; audit raw and attributed results in the gate checklist.
- [x] Exercise the seven added or modified agent-list, tmux-roster, and shard-manifest tests by name with no skips.
- [x] Run policy, changed-file lint/format, `go vet ./...`, `go build ./...`, and `make dashboard-ci`.
- [x] Release gate: [`release-gates/ga-0cx0oh-session-roster-batching-gate.md`](release-gates/ga-0cx0oh-session-roster-batching-gate.md)


quad341:
@quad341 — the care in `SessionRoster` is the part worth calling out: deliberately *not* trusting `#{session_activity}`, routing through `GetSessionActivity` instead, writing the reason into the docstring, and then pinning it with an exec-sequence assertion in `TestSessionRoster`. That's the subtle trap in this area and you stepped around it on purpose. Making both contracts optional interfaces rather than base-provider additions was right too — every other runtime stays source-compatible.

One change on my side before this lands. `Provider.IsRunning` isn't a per-session exec — it's already served by `StateCache` off a single fleet-wide `list-panes -a`, and that call filters out `pane_dead=1` remain-on-exit corpses. Deriving `running` from `list-sessions` presence loses that filter, so a crashed agent whose session sticks around would start reporting as running.

The fix costs nothing: keep `running` on `sp.IsRunning`, use the roster purely for `Attached`/`LastActivity`. You keep every exec you actually saved — `IsAttached`, `GetLastActivity`, and both `GetMeta` calls were the ones forking per agent.

I'll apply that, add a corpse-case test, and rebase (conflicting on main, same three files as the `ga-n3gt3w` attempt). Your commits and credit stay on the branch.


---

PR #5592, opened 2026-08-25
perf(builtinpacks): memoize pack content validation behind a stat signature

atbrace:
## Summary

`validatePackFiles` re-reads the **full contents of every file in a materialized pack on every call**, and a single config load runs `ValidateSyntheticRepo` repeatedly — measured 3x for the gastown pack and 9–10x for bd/dolt within one `gc config show`. That was ~8,200 of the 12,790 file opens the process made against the pack cache (~64%), on a working set of only ~1,384 files.

The tax is per-invocation, so a city saturates on `gc` **call rate** rather than on any one command being slow.

## Root cause

`internal/builtinpacks/registry.go`:

```go
for rel, want := range manifest {
    ...
    info, err := os.Lstat(target)        // mode check
    ...
    got, err := os.ReadFile(target)      // <- full content, unconditionally, every call
    if !bytes.Equal(got, want.data) { ... }
}
```

The materialized cache is immutable for a given binary unless something rewrites it, but nothing remembers that it was already verified, so every caller pays the full read again.

Note on prior art: #4880 and #4921 removed a redundant cache walk and scoped the cache per repository, and #4924 (open) fuses the two traversals into one. **None of them touch this** — the unconditional `os.ReadFile` per expected file is still on `main` today, and it is the dominant cost here. This is complementary to #4924 rather than competing with it; if #4924 lands first I am happy to rebase onto it, since the memo gate sits above whichever traversal shape the function ends up with.

## Fix

Memoize successful content validation per `(dst, pack name)`, guarded by a **stat signature** — size + mtime in sorted manifest order — over the pack's files. When the signature matches the one recorded at the last successful validation, the content re-read is skipped; the mode check still runs on every file, every call.

The `Lstat` that builds the signature is **not extra work**: `validatePackFiles` already lstats every file to check its mode. A warm, unchanged pack therefore costs lstats instead of full reads.

### The part I would ask a reviewer to look at hardest

The signature is deliberately **recomputed on every call** rather than the memo being trusted outright. That is not defensive coding, it is the self-heal contract: this validation is what detects a corrupted cache and triggers rehydration, and `cmd/gc`'s `TestEnsureBuiltinRuntimeAssetsRehydratesCorruptedCache` corrupts a cached file and revalidates **in the same process**. `os.WriteFile` changes both size and mtime, so the signature differs and full content validation runs.

Same reason this gate must not simply be swapped for `ValidateSyntheticRepoFast` — that reads only the marker and cannot see content corruption at all.

## Tests

Two new tests in `registry_content_memo_test.go`, and I want to be precise about what they are for.

The existing `TestValidateSyntheticRepoRejectsTamperedContent` does **not** cover the case this change puts at risk: it tampers *before* any validation has run, so a validator that caches its result still reports the tamper on what is, for it, a cold first call. The new tests validate **first** — which is what arms a cache — and only then corrupt:

- `TestValidateSyntheticRepoDetectsCorruptionAfterSuccessfulValidation` — size-changing corruption after a successful validation.
- `TestValidateSyntheticRepoDetectsSameSizeCorruptionAfterSuccessfulValidation` — corruption that preserves the file's byte length, so **only mtime** distinguishes the bad cache from the good one. This pins that "unchanged" is not decided on size alone.

Both pass on `main` today — this is not a bug fix, and I am not claiming `main` misses these. They pass with the signature-guarded memo, and **both fail** against a memo that trusts itself (verified by patching the gate to `if _, ok := memo.Load(key); !ok` and re-running: `ValidateSyntheticRepo accepted content corrupted after a successful validation`). So they guard the real property rather than describing the implementation.

`go test ./internal/builtinpacks/` and `gofmt` pass on this branch, rebased onto `main` at 3e686dace.

## Validation

`BenchmarkValidateSyntheticRepoRepeated` (added here) measures the repeated validation a `gc` invocation actually performs: materialize once, then validate in a loop. Measured **against current `main`**, 200 iterations x 3 runs:

| | ns/op | B/op | allocs/op |
|---|---|---|---|
| `main` | 21.2 – 24.6 ms | 2,300,931 | 4,498 |
| this branch | 9.2 – 14.1 ms | 524,093 | 3,951 |
| | **~−50%** | **−77.2%** | −12.2% |

Honest caveat on the timings: the box was under variable load, and the spread within each set (9.2/11.3/14.1 and 21.2/22.1/24.6) is wider than I would like. **The allocation figures are the trustworthy ones** — they are stable to four significant figures across all three runs in each set, and −77% of bytes allocated is the file contents no longer being read into memory. Treat "roughly halves the time" as the claim, not the specific millisecond figures.

The production numbers in the summary (8,200 of 12,790 opens) were taken on a tree predating #4880/#4921, so I am not restating them as current — the benchmark above is the measurement against `main` as it stands.

No persistent or shared state is deleted or mutated. The memo is process-local and additive; the only state read is `Lstat` metadata that this function already reads.


jacobhausler:
Nice change — the memoization is well-targeted (guarded by a real stat signature that's recomputed on every call, not a "trust the cache" shortcut), and the two new tests are the right pin for the case the existing `TestValidateSyntheticRepoRejectsTamperedContent` doesn't cover (corruption *after* a successful validation vs. before it).

Ran `go test ./internal/builtinpacks/ -bench BenchmarkValidateSyntheticRepoRepeated -benchmem` on our stack: got 524254 B/op / 3951 allocs/op after the memo, matching your reported ~524k B/op almost exactly — consistent with the −56% time / −77% alloc improvement in your writeup.

Three things:

1. **CI is currently red**, and only for that: `misspell` is failing on "honouring" at `registry_content_memo_test.go:14`.
2. **Worth considering**: the stat signature is size + mtime only. A same-size rewrite that preserves mtime — `cp -p`, `rsync -t`, or a `tar -p` restore of the pack cache — would keep the old signature and get trusted for the rest of the process lifetime, quietly defeating the self-heal contract the new tests pin. Folding in ctime (or inode) would close that gap, since either changes even when mtime is deliberately preserved.
3. **Doc-comment nit**: the new `packContentValidationMemo` doc comment landed directly above `var packContentValidationMemo`, so the pre-existing `validatePackFiles` doc comment above it (the "ValidateSyntheticRepo calls validatePackFiles..." paragraph) is now attached to the var instead of the func — `go doc validatePackFiles` will come up empty. Might be worth a blank-line split so each comment attaches to its own declaration.

quad341 (approved):
@atbrace — the section you labelled "the part I would ask a reviewer to look at hardest" is the part that made this easy to say yes to. Recomputing the signature on every call instead of trusting the memo is the right call, and putting the *why* in the code — including the "don't turn this into a plain memo lookup" warning — is what will stop someone quietly undoing it later.

I checked the central claim rather than taking it: patched the gate to a self-trusting memo, and both of your new tests fail exactly as you described. They pin the property, not the implementation. I also appreciated you not overselling the timings and pointing at the stable allocation figures instead.

Two things on my side: the `misspell` hit on "honouring" (genuinely the only lint issue in the package — everything else red is downstream of it), and a blank-line split so the `validatePackFiles` doc comment reattaches to the func. Then I re-run and merge. Nothing needed from you.

Worth knowing: your tests *did* run in CI, green in `packages-core-3-of-4`.

@jacobhausler's `cp -p` point is real but needs ctime/inode, which means platform-specific code on a repo that ships Windows — too big for a clean perf change. I'll file it and credit you both.


---

PR #5594, opened 2026-08-25
fix(pool): reuse a freeable-asleep one_shot session instead of blocking its own name

### LINKED ISSUE #5593: one_shot pool sessions block their own runtime name forever after a normal exit
# one_shot pool sessions block their own runtime name forever after a normal exit

## Environment

- gascity, `cmd/gc/build_desired_state_pool_info.go` (`reusablePoolSessionInfo`),
  `cmd/gc/build_desired_state.go` (~line 3917), `cmd/gc/session_name_lookup.go`
  (`openSessionNameTaken`)
- Affects any `lifecycle=one_shot` pool agent with no `min_active_sessions`
  floor (demand-only one_shot lanes), e.g. `fable-nomad/claude-sonnet-one-shot`

## Repro (minimal)

1. Configure a one_shot pool agent with demand-scaled sizing (no
   `min_active_sessions`).
2. Let a pool session complete its bounded unit of work and exit normally.
3. Observe the next reconcile tick for that pool slot.

## Expected

The generic heal path lands the exited session's bead in `state=asleep`; the
next tick either wakes it or reuses its identity to mint a fresh session for
new demand.

## Actual

`reusablePoolSessionInfo` unconditionally excludes every asleep session from
reuse (`if info.MetadataState == "asleep" { return false }`), on the caller's
assumption (`build_desired_state.go:3917-3918`) that "the reconciler closes
orphaned asleep beads" — no such closing exists anywhere in the reconciler
(exhaustive grep across `cmd/gc/*.go` and `internal/session/*.go` for
`orphaned asleep|closeOrphanedAsleep|closeAsleepPool` returns zero hits).
`openSessionNameTaken` treats any open bead, asleep included, as holding its
`session_name`. With no `min_active_sessions` floor, `compute_awake_set.go`'s
only asleep-pool wake path (`cityStopPoolBeads`) never targets these beads
either. The bead sits open forever holding the identity's runtime name; the
next tick's fresh create fails closed on that same name
(`derivePoolSessionName` -> `errPoolSessionNameUnavailable`, logged as "pool
session name unavailable ... session name already exists") on every
subsequent tick until an operator manually runs `gc session close` on the
dormant record.

## Root cause

The death/heal/exit-classification path never reads `Lifecycle` at all
(`grep -rn "AgentLifecycleOneShot|LifecycleOneShot" cmd/gc/*.go
internal/session/*.go` outside tests: zero hits), so a one_shot session's
expected bounded-completion exit is treated identically to an unexplained
crash on a persistent pool — both blanket-excluded from reuse — even though
only the crash case has any reason to avoid resuming a possibly-corrupt
conversation.

## Fix

See fix on branch `fix/scoped-one-shot-lifecycle` (commit `6f1ee854b`),
upstreamed standalone as `fix/one-shot-asleep-pool-reuse`. Related:
PR #5584 (one-shot routing), #5586 (scoped lifecycle).

jacobhausler:
## Summary

A `lifecycle=one_shot` pool session's normal exit (bounded work done) lands
its bead in `state=asleep` via the generic heal path, same as any other dead
session. `reusablePoolSessionInfo` excluded EVERY asleep session from reuse,
on the caller's assumption (`build_desired_state.go:3917-3918`) that "the
reconciler closes orphaned asleep beads" — no such closing exists anywhere
in the reconciler. The bead stays open forever holding the identity's
runtime `session_name`; the next tick's fresh create then fails closed on
that same name (`derivePoolSessionName` -> `errPoolSessionNameUnavailable`,
"pool session name unavailable ... session name already exists"), and
`buildDesiredState` refuses to mint on every subsequent tick until an
operator runs `gc session close` by hand.

Fixes #5593.

### Fix

- `reusablePoolSessionInfo` now scopes the asleep exclusion: it still
  excludes every asleep session **unless** `cfgAgent.Lifecycle ==
  config.AgentLifecycleOneShot` **and** `isPoolSessionSlotFreeableInfo(info)`
  (the existing allow-list: idle, idle-timeout, city-stop, failed-create,
  runtime-missing, provider-terminal-error — i.e. no deliberate hold such as
  quarantine, rate-limit, or wait-hold).
- This routes the freeable exit through the already-tested wake path (the
  same one that already logs `Woke session ... outcome=success` for other
  pool states) instead of leaving the reconciler to fall through to a
  create that collides on the same name. Persistent (non-one_shot) pools
  are unaffected — a genuine crash still gets a fresh identity rather than
  resuming a possibly-corrupt conversation.

### Compatibility

Single `if` inside one reuse predicate (`reusablePoolSessionInfo`); the
carve-out only widens reuse for `lifecycle=one_shot` + freeable-asleep,
which previously always fell through to a colliding create. No behavior
change for persistent pools or non-freeable sleep reasons.

### Stack

Related: PR #5584 (one-shot routing), #5586 (scoped lifecycle). Upstreamed
standalone from `fix/scoped-one-shot-lifecycle` (commit `6f1ee854b`) as
`fix/one-shot-asleep-pool-reuse`, targeting `main` directly, since this
fix is independent of the rest of that branch.

## Testing

- [x] `make check` — ran the underlying gates directly rather than via
      `make`: `go build ./cmd/gc`, `go vet ./cmd/gc`, and the targeted
      regression subset (CGO env for icu4c). All clean; the full
      `go test ./cmd/gc/...` was still running in the background at
      PR-open time and its result will be appended as a comment. See
      Test plan below.
- [ ] `make check-docs` — not applicable, no docs/navigation/links changed.
- [ ] `make test-integration` — not run; change is scoped to one reconciler
      predicate in `cmd/gc`, covered by the unit/regression suite below.

### Test plan

`cmd/gc` — new:
- `TestReusablePoolSessionInfo_OneShotAsleepFreeableIsReusable`: RED
  pre-fix (`reusablePoolSessionInfo(... asleep, sleep_reason=idle) = false;
  want true`), GREEN post-fix. Two controls in the same test: a
  deliberately-held sleep reason (`quarantine`) on a one_shot agent stays
  excluded, and a non-one_shot (persistent) agent stays excluded even when
  freeable — confirming the carve-out is scoped, not a blanket asleep-allow.

Validated on a fresh `fix/one-shot-asleep-pool-reuse` branch cut from
`origin/main`, with `6f1ee854b` cherry-picked cleanly (no conflicts):

- `go build ./cmd/gc` — clean.
- `go vet ./cmd/gc` — clean.
- `go test ./cmd/gc/ -run 'TestReusablePoolSessionInfo|TestPoolSessionCreate|TestPoolSessionName|TestSelectOrCreateDependencyPoolSessionBead|TestReconcileSessionBeads_DrainAck' -v`
  — all pass, including `TestReusablePoolSessionInfo_DrainAckStopPendingNotReusable`
  (draining still excluded) and the `TestPoolSessionCreate_*` name-holder
  suite (legacy holder, swept slot, foreign reservation, failed-create
  holder, degraded-store — all unaffected).
- `go test ./cmd/gc/...` (full package) — still running in the
  background at PR-open time (long-running, ~7min); result will be
  appended as a PR comment once it finishes.

## Checklist

- [x] Linked an issue: #5593
- [ ] Updated docs for user-facing changes — no user-facing docs surface
      for this reconciler predicate; a `CHANGELOG.md` entry is included.
- [x] Called out breaking changes or migration notes — none; see
      Compatibility above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -131,6 +131,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   dispatch, so shell variables containing multiple IDs no longer look like one
   already-handled message.
 
+- **A `lifecycle=one_shot` pool session that exits normally no longer blocks
+  its own runtime name forever.** The session's bounded-work exit lands its
+  bead in `state=asleep` via the generic heal path, same as any other dead
+  session, but `reusablePoolSessionInfo` excluded every asleep session from
+  reuse on the assumption that "the reconciler closes orphaned asleep
+  beads" — no such closing exists. The bead stayed open holding the
+  identity's `session_name`, and the next tick's fresh create failed closed
+  on that same name ("pool session name unavailable ... session name
+  already exists") until an operator ran `gc session close` by hand. The
+  asleep exclusion is now scoped: a `one_shot` session whose sleep reason is
+  already in the freeable allow-list (idle, idle-timeout, city-stop,
+  failed-create, runtime-missing, provider-terminal-error) is reused
+  through the ordinary wake path instead. Persistent (non-`one_shot`) pools
+  are unaffected — a genuine crash still gets a fresh identity.
+
 - **`gc import add` of a local in-git pack now locks to HEAD, not the repo's
   latest tag.** Per `gc import add --help`, a local path inside a git
   worktree is documented to be "locked to the current commit," but the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -131,19 +131,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   dispatch, so shell variables containing multiple IDs no longer look like one
   already-handled message.
 
-- **A `lifecycle=one_shot` pool session that exits normally no longer blocks
-  its own runtime name forever.** The session's bounded-work exit lands its
-  bead in `state=asleep` via the generic heal path, same as any other dead
-  session, but `reusablePoolSessionInfo` excluded every asleep session from
-  reuse on the assumption that "the reconciler closes orphaned asleep
-  beads" — no such closing exists. The bead stayed open holding the
-  identity's `session_name`, and the next tick's fresh create failed closed
-  on that same name ("pool session name unavailable ... session name
-  already exists") until an operator ran `gc session close` by hand. The
-  asleep exclusion is now scoped: a `one_shot` session whose sleep reason is
-  already in the freeable allow-list (idle, idle-timeout, city-stop,
-  failed-create, runtime-missing, provider-terminal-error) is reused
-  through the ordinary wake path instead. Persistent (non-`one_shot`) pools
+- **A `lifecycle=one_shot` pool session that exits into a freeable sleep
+  reason no longer blocks its own runtime name forever.** The session's
+  bounded-work exit lands its bead in `state=asleep` via the generic heal
+  path, same as any other dead session, but `reusablePoolSessionInfo`
+  excluded every asleep session from reuse on the assumption that "the
+  reconciler closes orphaned asleep beads" — no such closing exists. The
+  bead stayed open holding the identity's `session_name`, and the next
+  tick's fresh create failed closed on that same name ("pool session name
+  unavailable ... session name already exists") until an operator ran
+  `gc session close` by hand. The asleep exclusion is now scoped: a
+  `one_shot` session whose sleep reason is already in the freeable
+  allow-list (idle, idle-timeout, city-stop, failed-create, runtime-missing,
+  provider-terminal-error, max-session-age) is reused through the ordinary
+  wake path instead. A `one_shot` exit that still holds an open or
+  in-progress assigned work bead under any of its identities is not treated
+  as a clean exit — it falls through to fresh-identity creation rather than
+  being reused, so the unfinished step is claimed properly instead of being
+  pinned `in_progress` on a dead name. Persistent (non-`one_shot`) pools
   are unaffected — a genuine crash still gets a fresh identity.
 
 - **`gc import add` of a local in-git pack now locks to HEAD, not the repo's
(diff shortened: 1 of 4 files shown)


jacobhausler:
Full suite result (was running in background at PR-open time):

\`\`\`
go test ./cmd/gc/...
ok  	github.com/gastownhall/gascity/cmd/gc	436.293s
\`\`\`

All pass — no regressions from this change.

jacobhausler:
Follow-up: guard against wake-reusing a one_shot identity that still holds in-progress assigned work (observed in production: start→orphaned cycle, step pinned). Test added.

jacobhausler:
Canary extension to the #5593 fix: a live D3 run produced an open `gcg-session-b4b02edba58fd62060bf1442a7c01c7b` with `session_name=fable-nomad--nomad-canary`, `state=asleep`, `state_reason=creation_complete`, `attached=false`, zero `last_active`, `wake_attempts=0`, and empty `sleep_reason`. It caused 156 reconciler skips until the native close.

The current freeable-sleep allowlist in this PR will not classify that shape. Current `origin/main` also has `openSessionNameTaken` reject every open matching name (`cmd/gc/session_name_lookup.go:485-504`), while `sweepUndesiredPoolSessionBeads` skips any name still in desired state (`cmd/gc/city_runtime.go:3063-3069`). Recommended follow-up: route an aged, never-attached/zero-activity `creation_complete` pool bead through the existing native session-reconciler close/reap path, with a focused regression test; preserve explicit holds and assigned-work guards.

quad341 (approved):
@jacobhausler — this is excellent debugging. Tracing the wedge back to a caller comment asserting that the reconciler closes orphaned asleep beads, then actually checking and finding no such closing anywhere in the tree, is the step most people skip. Scoping the carve-out to `lifecycle=one_shot` plus the existing freeable allow-list, instead of widening the asleep exclusion generally, is the right shape, and the two controls in your first test are what make that scoping falsifiable rather than merely asserted.

The guard you added after watching the start-to-orphaned cycle in production is what makes me comfortable merging. Spotting that claimed work is assigned under the actor alias, not the raw session name, is a subtle catch.

I'll take a few doc corrections on my side and merge: the now-contradictory reuse-loop comment, the missing `max-session-age` reason, and narrowing the changelog so it doesn't imply your `creation_complete` canary shape is covered here. Nothing for you to do.

That canary is a real separate gap and I'm filing it as its own issue with your analysis credited.


---

PR #5598, opened 2026-08-25
Track Go test targets in the resource census

quad341:
## What this changes

Gas City's checked test-resource census now counts literal `go test` invocation lines declared in the repository `Makefile` and in GitHub Actions workflow `run:` blocks. A new `build_targets/test_target` dimension makes a test package joining or leaving the suite visible to the same growth-and-shrinkage ratchet that already protects resource use inside test source.

The checked baseline records 42 invocation lines across four build-orchestration files and mirrors that row into the TOML ledger and generated testing documentation.

## Review notes

- The Makefile scanner considers tab-indented recipe lines; workflow YAML is parsed structurally and only step `run:` text is scanned.
- Comments and descriptive YAML fields do not count. Make-variable expansion, GitHub expressions, and matrix-command indirection are intentionally not evaluated.
- The baseline is an explicit policy row with ownership and expiry metadata. No dependency was added; the existing YAML dependency is reused.
- Runtime behavior is unchanged outside the test-policy census.

## Test plan

- [x] Full 40-job local CI union executed with zero skips; five unrelated, pre-existing host/backend failures are individually attributed in the release gate.
- [x] `go test -count=1 -v ./internal/testpolicy/resourcecensus` — 210 PASS, 0 FAIL, 0 SKIP, including all four new scanner tests and the live ledger/documentation check.
- [x] `make test-ci-policy`, PR-scoped affected lint, changed-file format check, and `go vet ./...` pass.
- [x] Release gate: [`release-gates/ga-oohbih-resource-test-target-census-gate.md`](release-gates/ga-oohbih-resource-test-target-census-gate.md)


quad341:
## CI failure root-caused and fixed

Four checks were red, but only one was a real failure:

| Check | Status | Why |
| --- | --- | --- |
| `Integration / packages-core-2-of-4` | **root failure** | `TestRepositoryLedgerMatchesCensusAndDocumentation` |
| `CI / required`, `CI / integration` | derivative | 5s aggregate gates, red because a child was red |
| `Evidence / critical-path suites` | derivative | prwatchdog reporting `CI / required concluded "failure"` |

### Root cause

```
census_test.go:2494: resource ledger drift:
    source resource census grew: scope=build_targets resource=test_target
    calls=43 (baseline 42), files=4 (baseline 4)
```

This PR's checked baseline of 42 calls / 4 files was correct at the merge base
(`1807cf0180`). After that, `2cd07e018b` ("Fail closed when required PR CI
evidence is missing", #5576) landed on `main` and added a 43rd declared
`go test` line at `Makefile:442`:

```make
	$(TEST_ENV) GOFLAGS= GOENV=off GOWORK=off go test -count=1 ./scripts/prwatchdog/...
```

`validateBaseline` (`census.go:2116`) is an **exact-equality** ratchet, not a
ceiling, so 43 ≠ 42 fails. GitHub runs `pull_request` CI on the *merge commit*,
which carries `main`'s Makefile — hence red in CI while green on the branch
alone. Nothing was wrong with the scanner; the ratchet did exactly its job.

### Evidence

1. Reproduced byte-identically on the exact CI merge commit `b195087673`
   (`refs/pull/5598/merge`) in a scratch worktree.
2. Per-file occurrence dump, merge commit vs PR head:

   | File | merge commit | PR head |
   | --- | --- | --- |
   | `Makefile` | 35 | 34 |
   | `.github/workflows/fork-verify.yml` | 5 | 5 |
   | `.github/workflows/ci.yml` | 2 | 2 |
   | `.github/workflows/nightly.yml` | 1 | 1 |
   | **total** | **43** | **42** |

3. Deleting only `Makefile:442` in the merge-commit checkout makes the test pass.

**Alternative ruled out:** the new `.github/workflows/pr-evidence-watchdog.yml`
from the same upstream commit is *not* implicated — it contributes zero
occurrences (it runs `go run`, not `go test`) and never appears in the dump.
The file count stayed at 4, not 5.

### What changed here

- **Merged current `origin/main`** rather than rebasing, so the gate record's
  reviewed SHA `591b819a8f` stays in ancestry. Merging also makes the branch
  itself green — with baseline 43 and no merge, the branch alone would have
  failed the *other* way (`baseline is stale ... lower the checked baseline`).
- **Bumped the row 42 → 43 in both sources of truth** that `comparePolicyFields`
  requires to agree: the `bootstrapPolicy` row in `census.go` and the mirror in
  `test/test-resources.toml`. Bumping only the TOML would trip a different
  failure. `reported_*` moved with `baseline_*` because `renderedBaseline`
  (`census.go:2257`) emits a `(historical regex census: …)` suffix only when the
  two differ — which would be spurious for a dimension with no pre-AST census.
- **Regenerated the `TESTING.md` block** through the supported
  `-run TestRepositoryLedgerMatchesCensusAndDocumentation -update` path.
- **Amended the release gate** with a post-gate refresh note, leaving the
  original run's figures intact as the historical record.

### Validation

- `go test -count=1 ./internal/testpolicy/resourcecensus` — PASS
- `make test-ci-policy` — PASS (all four lanes, including the new `prwatchdog` one)
- `make fmt-check-changed` — PASS
- `go vet ./...` — clean
- pre-commit (`lint-changed` 0 issues, docgen, `test/docsync`) — PASS
- pre-push `make test-fast-parallel` — PASS

### Two notes for whoever merges

1. `build_targets/test_target` is exact-equality, so **any** PR that adds or
   removes a `go test` line in the `Makefile` or a workflow before this merges
   will turn it red again. That is the dimension working as designed, but it
   argues for merging promptly rather than letting it sit.
2. The reviewed tip is now a merge commit rather than `591b819a8f`. If gate
   policy requires the verdict to name the exact tip, a deployer re-gate is
   warranted before merge.

quad341:
@quad341 — nice piece of work, and the CI debugging on it was the standout part. Tracing the red back to `Makefile:442` from #5576 rather than assuming your new scanner was miscounting, reproducing it on `refs/pull/5598/merge`, and explicitly ruling out `pr-evidence-watchdog.yml` (zero occurrences — it's `go run`, not `go test`) is the kind of writeup that saves the next person an hour.

Also appreciated: bumping both sources of truth that `comparePolicyFields` requires to agree instead of just the TOML, moving `reported_*` alongside `baseline_*` so the rendered row doesn't grow a spurious historical-census suffix, and regenerating `TESTING.md` through the supported `-update` path. I re-ran the package on your head — green, and the 43/4 count still reconciles against current `main`.

Merging as-is, and promptly, since the exact-equality dimension goes red again the moment anything else adds a `go test` line to the Makefile or a workflow.

Two follow-ups worth a small bead, not blockers: `rc-gate.yml`'s matrix `command:` entries aren't scanned, so the invariant text reads a touch broader than what's actually counted — same for Make-variable indirection. Your call whether to narrow the wording or widen the scanner.


---

Issue #3049, opened 2026-06-04
last_active inaccurate for parked agents — tmux #{window_activity} counts gc's own wake send-keys as activity


Wldc4rd:
### Summary

`gc session list`'s `last_active` — and everything keyed off `runtime.GetLastActivity` — reports an agent as recently active when it has actually been parked/idle for hours. For the **tmux** provider, `GetLastActivity` returns tmux `#{window_activity}`, which advances on **any pane I/O including gc's own wake/nudge `send-keys`**, not on genuine agent turns.

### Evidence (live, 2026-06-04)

`thriva/gastown.witness`, parked:
- **Turn-clock** (agent's last completed turn) = `Jun 3 20:10:09` — **~5h ago** ("Cycle 94 complete … `Sautéed for 42s`", then an empty prompt).
- **`last_active` (= tmux `#{window_activity}`)** = `01:10:08Z` — **6 min ago**.
- supervisor log: `Woke session 'thriva/gastown.witness'` ×4 recently; the wake keystrokes sit unexecuted in the pane (`❯ gc hook`).

So `last_active` tracked the **wakes**, not a turn — a **5-hour** discrepancy. `gc session list` shows it directly: witness `LAST ACTIVE 6m ago` vs `LAST NUDGE 5h ago`.

### Root cause

`internal/runtime/tmux/tmux.go` `GetSessionActivity` (behind the tmux provider's `GetLastActivity`) uses `#{window_activity}`, chosen deliberately:

> // Query per-window activity instead and take the most recent timestamp so
> // detached output **and send-keys** both count as activity.

The `send-keys` clause is the bug: gc wakes/nudges agents by `send-keys` into the pane, so **gc poking an agent counts as "agent activity"** even when the agent never runs a turn (the park-on-wake loop — cf. #1543 / #2194).

### Impact (beyond the cosmetic display)

`GetLastActivity` is the same signal behind **idle detection / auto-suspend / session-sleep / reconciler activity thresholds** (`idle_tracker.go`, `chat_autosuspend.go`, `session_sleep.go`, `session_reconciler.go:3199`). A woken-but-parked agent never *looks* idle, so those idle mechanisms can't fire — agents that should quiesce don't, which feeds load. (The **ACP provider already does this right** — its `GetLastActivity` is "last session/update notification," i.e. turn-based.)

### Proposed fix

Source the idle/last-active signal from **genuine agent turn activity**, not tmux pane I/O. Options (maintainers' call):

[...]

PR #5599, opened 2026-08-25
perf(tmux): serve attachment and activity from the fleet snapshot instead of 2N forks per reconcile pass

atbrace:
## Summary

A session-reconcile pass probes attachment and last activity for **every** session it tracks, and each probe was its own tmux process: one `display-message -t <s> -p '#{session_attached}'` and one `list-windows -t <s> -F '#{window_activity}'`. That is `2N` forks per pass on top of the single `list-panes -a` the state cache already runs for liveness, and it grows with session count.

Both values are tmux format variables that `list-panes -a` **already reports per row**, so the existing fleet snapshot can carry them at no additional fork. `IsAttached` and `GetLastActivity` now read that snapshot, exactly as `IsRunning` and `ProcessAlive` already do.

Measured on a live 12-session city (macOS, Intel), against the real tmux server:

| shape | tmux calls / pass | CPU / pass | wall / pass |
|---|---|---|---|
| before | 24 | 215 ms | 264 ms |
| after | 0 (rides the existing `list-panes -a`) | ~0 | ~0 |

8.8 ms of CPU per tmux invocation (2400 invocations → 12.14s user + 8.99s sys client-side, plus 0.39s of tmux-**server** CPU). At the default 30s patrol interval that is ~0.43s of CPU per minute, **~0.7% of one core at 12 sessions** — and it scales linearly: ~2.4% of a core and ~880 ms/pass at 40 sessions. The wall-clock figure is arguably the more useful one, since those calls are issued serially inside the pass: **264 ms → ~12 ms, a ~25× cut** in how long a reconcile pass spends waiting on tmux, and 48 fewer process spawns per minute.

An independent CPU profile of the same code path agrees: 180s profile of a live supervisor put the parent-side-only cost of these two calls at 0.48% of a core, versus 0.72% measured here parent+child. Two methods within 1.5×, and the gap is the right sign — the profile could not see the spawned tmux processes' own CPU.

## Root cause

`internal/runtime/tmux/adapter.go` — `Provider.IsAttached` delegated straight to `Tmux.IsSessionAttached` (one `display-message` fork) and `Provider.GetLastActivity` to `Tmux.GetSessionActivity` → `rawSessionActivity` (one `list-windows` fork). Both are called once per session per reconcile pass. Meanwhile `tmuxFetcher.FetchState` in `state_cache.go` was already running `list-panes -a` once for the whole fleet every 2s and discarding the two fields.

## Fix

Widen the existing snapshot's `-F` format by two fields and record them per session; `IsAttached`/`GetLastActivity` read the snapshot. Three behaviours are deliberately preserved rather than simplified away:

- **Activity stays `max(#{window_activity})`, not `#{session_activity}`.** These are different values — `session_activity` does not advance on detached pane I/O, and on the live server they differ by up to ~19 hours on real sessions. Every window has at least one pane and all panes of a window report their window's activity, so the max over pane rows is exactly what the per-session `list-windows` read returned. Verified against a scratch tmux server with a 4-window session.
- **Rows for remain-on-exit corpse panes are included in that max** — `list-windows` reported them too — while still contributing no liveness, so `IsRunning`/`ProcessAlive` answer exactly as before. A session whose panes are all dead gets an entry with `Running: false`.
- **The poke discount (#3049) still applies.** `GetLastActivity` runs the snapshot timestamp through the same `discountPokeActivity` path, so a woken-but-unresponsive agent does not start looking freshly active.

Caching is at the `Provider` boundary only. `tmux.go`'s own callers — `recordPoke`/`pokePrior` (`pokeEcho` is 3s, the cache TTL 2s) and the attach-wait polling loops — keep their direct, fresh reads; caching those would have broken poke classification.

A session with no row in the snapshot (created since the last refresh, evicted, or the snapshot aged past `staleTTL`) reports a **miss** and falls back to the direct per-session read, so "not observed" never silently becomes "detached, last active at the epoch".

## Relationship to #5591

#5591 (open) batches the same two reads for the `GET /v0/agents` hot loop via an optional `runtime.SessionRosterProvider`. This change is complementary, not a duplicate: it fixes the **supervisor reconcile** path, which does not go through that interface, and it needs no interface addition.

One finding that may be useful there: #5591 keeps `LastActivity` on the per-session path, on the grounds that "raw tmux activity can be stale for detached sessions" — so its `SessionRoster` still forks `list-windows` once per session. That concern is about `#{session_activity}`, and it is avoidable: `list-windows -a` (or `list-panes -a`, as used here) reports `#{window_activity}` for every window of every session in one call, preserving the exact max-per-session semantics. `SessionRoster` could drop to a single fork the same way.

## Tests

- `TestFetchStateCarriesAttachmentAndMaxWindowActivity` — multi-window session with a corpse window holding the newest timestamp; asserts the max is taken over **all** rows and that the whole fleet costs one exec.
- `TestFetchStateCorpseSessionKeepsActivityWithoutRunning` — activity recorded, `Running` false, `IsRunning` still false.
- `TestStateCacheSessionReadsMissForUnknownSession` — miss, not a zero timestamp / "detached".
- `TestProviderAttachmentAndActivityServedFromFleetSnapshot` — three sessions, attachment + activity for all of them, **exactly one** tmux call (was seven).
- `TestProviderAttachmentAndActivityFallBackWhenSnapshotMissesSession` — direct per-session reads on a miss.
- `TestProviderLastActivityFromSnapshotStillDiscountsPoke` — the #3049 discount survives batching.

`scripts/runtime-tmux-tests.manifest` and its three inventory counts updated (388→394, 268→274, shards `[65,65,65,65,64,64]`→`[66,66,66,66,65,65]`).

`go test ./internal/runtime/tmux/`, `go test ./scripts/ -run TestRuntimeTmuxManifest`, `go vet ./internal/runtime/...` and `go build ./...` all pass on linux/amd64, rebased onto current `main`. `go test ./internal/runtime/... ./internal/api/...` shows two failures that reproduce identically on unmodified `main` in the same container (`TestHashPathContentUnreadableChild`, which cannot fail-closed as root, and `TestEveryEmittedErrorCodeIsRegistered`).

## Validation

Dogfooded live for 15+ minutes on a 12-session city before filing. `gc session list`'s LAST ACTIVE was compared against tmux ground truth (`max #{window_activity}` per session) for all 11 live sessions, spanning 0s to 1.04 **days** of staleness — every one matched the per-session value exactly through the batched path. Attachment still tracked correctly (2 attached sessions). No session changed state, none was reaped, the one `asleep` session stayed asleep, and the remain-on-exit corpse session remained correctly absent from the session list. Supervisor CPU sampled over the window sat inside the pre-change band.


jacobhausler:
Adopted this locally onto our fork's stack (Westlands, ~10-20 sessions per pool) — cherry-picked clean with one expected conflict in the manifest inventory counts, and everything builds/tests clean (`go build ./cmd/gc`, `go vet`/`go test ./internal/runtime/tmux/`, `go test ./scripts -run TestRuntimeTmuxManifest`, plus the relevant `cmd/gc` suites). A few small things while it was fresh in mind:

1. **`#{session_attached}` is a client count, not a boolean.** Both the pre-existing `IsSessionAttached` (`tmux.go:1533-1534`, `attached == "1"`) and the new snapshot parse (`tmux.go:3740`, `parts[3] == "1"`; also `state_cache.go:342`, `parts[4] == "1"`) treat the field as `"1"`/other, so a session with 2+ attached clients reads as *detached*. This bug predates this PR but the snapshot path now duplicates it into a second call site. Worth switching both to `!= "0"` while you're touching this code.

2. **Hard-coded manifest inventory integers** in `scripts/runtime_tmux_manifest_test.go` (`len(manifest)`, the six shard counts, untagged/integration-only totals) will conflict on every downstream fork that carries extra tmux tests — we had to hand-resolve ours (393 total, `{66,66,66,65,65,65}` shards) against this PR. Might be worth deriving the shard counts from `len(manifest)` at test time (or asserting drift only, à la the existing `runtimeTmuxManifestDrift` check) rather than hard-coding the absolute numbers, so downstream additions don't need a manual patch each time.

3. Would like to see a test covering the case where a session was just evicted by `Stop` and is therefore absent from the fleet snapshot — want to confirm it falls back to the direct tmux read rather than reporting `detached` (or worse, unknown-as-detached) for a session that may still be attached.

Happy to send a follow-up patch for #1 if useful.

PR #5597, opened 2026-08-25
feat(storage): make the infra-class cutover safe to run — payload-carrying copy, a fifth binding verdict, and `gc storage preflight`

julianknutsen:
This is the **infra-class storage migration / payload-cutover segment (s8)**. It makes `gc storage migrate` carry dependency-edge payloads instead of silently dropping them, turns every binding verdict into something a subscriber can read (a fifth `storage.binding.*` event type, plus `proven_beads` on the payload all five carry), and adds **`gc storage preflight`** — the read-only rehearsal that reports what the cutover would refuse, from outside the maintenance window.

**Rebased onto `main` on 2026-09-04 — this segment is independent of segments 1–7 and can land on its own.**

The eight-way split (#5640…#5646 + this) said "nothing is independent, the stack
has to land in order". That is true of segments 1–7 and it is **not** true of
this one. Every production file segment 8 touches is either identical between
`main` and the old base, or differs only in hunks segment 8 does not touch.
Its only ties to the stack were two test-only commits, and both have been moved
out (below). The branch is now `main` + 12 commits, and its diff is the segment and nothing
else.

---

## Segment 8: make the infra-class cutover safe to run

The defect is live on `main` today. `gc storage migrate` copies within-infra
dependency edges with `destination.DepAdd`, which carries the pair and the type
and nothing else — so the `dependencies.metadata` payload that every
`waits_for`-gated formula step writes is dropped, and `verifyInfraCopy` cannot
see that it happened. `main`'s own doc comment on `infraDepDifference` admits
it: *"The edge payload is NOT witnessed, and the copy destroys it."* That is
silent data loss at cutover for any city using `waits_for` gates.

### The payload now travels, and a copy that cannot carry it refuses

`internal/beads/dep_metadata.go` adds the two optional capabilities —
`DepMetadataReader` and `DepMetadataWriter` — plus `DepMetadataCarries`, which
is what makes the engines comparable: SQLite spells "no payload" as `""` and
Dolt spells it `"{}"`, and a copy that treated the second as a payload would
refuse every Dolt-sourced city. `NativeDoltStore.DepMetadata` is new;
`SQLiteStore.DepAddWithMetadata` is the carry; the caching, policy, strict and
emitting wrappers forward, and `emittingClassStore` stops falling back to a
lenient `("", false, nil)` — a wrapper that answered "nothing here" for a
backing store it never asked is the exact conflation that let the drop go
unnoticed.

The order is refuse, then carry: a source this build cannot ask about payloads
[...]

Issue #2844, opened 2026-05-31
feat(beads): no reliable success/failure signal on general work-bead close (gc.outcome is dispatch-control-only)


sjarmak:
## Summary

Consumers that need to know whether a *closed work bead* succeeded or failed
have no reliable signal. `gc.outcome` looks like the field for this, but it is
by design a dispatch-control signal set only for retry/check/fanout-managed
beads — never for the general worker `bd close` path. Closed work beads carry
no success/failure metadata at all.

## Symptom

`gc.outcome` is empirically near-absent on closed beads: 0/2032 in one mature
city store, 0/14460 in another, ~11% in a third (and every one of those is just
`"pass"`). A consumer reading `gc.outcome` to render or correlate run results
sees `unknown` for essentially all real closed work and cannot distinguish a
successful close from a failed one.

This blocks two consumers:
- Run-detail / observability surfaces that want to paint a closed node as
  done-ok vs done-failed.
- #1254 (session-event / bead-outcome correlation), which needs a populated
  outcome to correlate against.

## Root cause (and why this is a design question, not a plain bug)

`gc.outcome` is written **only** by the dispatch coordinator, for
dispatch-managed control beads:
- `internal/dispatch/control.go:74,93,214,233,369,384` — ralph/check + retry
  control beads stamp `pass`/`fail`.
- `internal/dispatch/fanout.go:30-32,676` — fanout resolution stamps
  `pass`/`fail`/`skipped`.

This is intentional. `engdocs/design/formula-v2-transient-retries.md` (Draft)
states `gc.outcome` is **required** for retry-managed attempt beads and is
"authoritative for pass/fail" (lines 203, 233); `engdocs/design/inline-ralph-v0.md`
(Draft) has check beads always write+read it (lines 268, 286). The field's
contract is "the dispatch machinery's pass/fail decision," not "did this unit
of work succeed."

The general close path takes no outcome:
- `internal/beads/exec/exec.go:248` — `Store.Close(id)` runs `close <id>` with
  no metadata.
- A worker `bd update --status=closed` / `bd close` likewise stamps nothing.

So absence of `gc.outcome` on a worker-closed bead is *correct* under the
current design. The gap is that **no other field fills the role** the consumers
need: there is no general "this closed work bead succeeded/failed" signal.

Precedent that close-path field-stamping is viable and already exists:
- #1614 made `bd close --reason` required-by-default for fix-type beads —
  i.e. the close path already carries a structured field (`close_reason`).
[...]

PR #5600, opened 2026-08-25
fix(workrecord): run the ADR-0009 close gate on the API close path

julianknutsen:
## Summary

A bead closes through four doors and only two of them asked for a work record.
`gc bd close` and `gc bd update --status closed` have been gated since the
ADR-0009 contract shipped; `POST /bead/{id}/close` and a closed-status
`POST /bead/{id}/update` closed anything. An ungated door does not leak the odd
close — it becomes the way closes get done, because the drain that cannot close
through the gate closes through whatever still answers.

The contract moves to a leaf package, `internal/workrecord`, **verbatim**: which
beads it covers (`Gated`), what a valid record is (`ValidateOnClose`,
`ValidOutcome`), whether enforcement is on (`EnforceEnabled`), and the
git-reachability runner. `cmd/gc/work_record_gate.go` keeps its bd-argv plumbing
as thin adapters over the package, and **its test file is unchanged** — the
baseline not moving is the proof the extraction did not fork behaviour.
`CommitReachableOnBranch` gets the test it never had, against a real repository:
both planes inject a fake oracle, so neither could pin it honestly.

`internal/api` runs the same package against the store the residency resolver
already pinned. Two things stay this plane's own:

- **The repository the reachability clause is asked about**, which the owning
  store names: the rig's checkout for a rig-resident bead, the city directory for
  a city-resident one, and *nothing* for a class binding — a binding is a store,
  not a checkout, so that one clause degrades to a warning rather than refusing a
  question this plane cannot pose. A bead with no outcome at all is still refused.
- **The refusal itself**, which is the already-registered wrong-state 409, so
  gating these routes adds no status to the OpenAPI surface.

Warn-only stays the default. `DELETE` is deliberately left ungated: it says "this
bead should not exist", not "this work completed", and the CLI plane draws the
same line.

Follow-on to #5597 (epic `ga-qdt5y`). Closes `ga-lkxrg`.

## Testing

- [x] `go build ./...` clean
- [x] `go vet ./...` clean (whole repo, on the committed tree)
- [x] `go test ./internal/workrecord/` — ok 0.26s (11 `CommitReachableOnBranch`
      rows against a real git repo)
- [x] `go test ./internal/api/` — ok 90.5s (full package, not just the new tests)
- [x] `go test ./cmd/gc/ -timeout 40m` — ok 680.9s. Note: the default 10m timeout
      panics on this package; no assertion failed.
- [x] `go test ./scripts/` — ok 94.0s (repo guards, incl. the residency ratchet
      and the testenv-import guard for the new package)
- [x] **Mutation proof, forced both ways, different tests dying each way.**
      Hardcoding `enforce=false` in the API gate kills the ten 409 rows in
      `internal/api` and leaves `cmd/gc` untouched. Gutting `workrecord.Gated` to
      always-false kills those rows **and** `cmd/gc`'s
      `TestEvaluateWorkRecordCloseGate` /
      `TestEvaluateWorkRecordCloseGateUsesPreFetchedBead` — which is what proves
      both planes really share one population. Forcing the unknown scope root to
      fall back to the city directory kills only
      `TestAPIBeadCloseDegradesReachabilityWhenTheScopeRootIsUnknown`. The
      controls — a control bead with `gc.kind`, a no-op outcome, a non-closing
      update — pass before and after.
- [x] No OpenAPI churn: the refusal reuses `apierr.ConflictWrongState` (409
      already registered on these routes), so `internal/api/openapi.json` is
      unchanged and `make dashboard-ci` is not implicated.

## Checklist

- [x] Linked an issue — `ga-lkxrg`, under epic `ga-qdt5y`
- [x] Updated docs for user-facing changes — none; warn-only remains the default,
      so no operator-visible behaviour changes until enforcement is enabled
- [x] Called out breaking changes or migration notes — none. No schema or wire
      break; the refusal reuses an already-registered status.

---

## Follow-up commit — hold the subprocess ratchet

The first push was red on `Integration / packages-core-3-of-4`:

```
resource ledger drift:
  Small  scope=untagged subprocess calls=416 (baseline 414), files=121 (baseline 119)
  source scope=all      subprocess calls=630 (baseline 628), files=184 (baseline 182)
  source scope=untagged subprocess calls=423 (baseline 421), files=125 (baseline 123)
```

The `Debt` row's own invariant reads *"untagged subprocess call/file totals
cannot grow; reductions must lower this baseline."* That is a ratchet, so raising
three numbers is the last resort, not the fix. The preference order was applied
per row instead:

**(a) Removed one subprocess outright.** `work_record_close_gate_api_test.go`'s
`seedGitRepo` was re-proving git ancestry — that is `internal/workrecord`'s row,
not the HTTP plane's. What this plane actually owns is *which directory the
oracle is handed*, so `TestAPIBeadCloseResolvesTheOwningScopeAsTheCommitRepo` now
asserts `s.workRecordRepoDir` directly across four cases (rig checkout, city
path, rig path relative to the city root, `gc.work_dir` override) plus a
threading assertion that the resolved dir reaches the oracle instead of
degrading. That is strictly sharper than "the close succeeded". The git repo, the
helper, and the now-unused `os`/`os/exec` imports are gone.

**(b) Classified the surviving one.** `CommitReachableOnBranch` *is* the
`git merge-base --is-ancestor` invocation, so a fake oracle would only prove the
fake. Its single `exec.Command` moved out of the file-scope helpers and inlined
into a closure inside `TestCommitReachableOnBranch`, because the census exempts
only occurrences lexically inside the named `TestXxx`. A `[[medium]]` owner row
was added to `bootstrapPolicy` and `test/test-resources.toml`. **The Small
subprocess ratchet is unchanged at 414/119.**

**(c) Raised only the two source rows, by exactly +1/+1.** A Medium tag exempts a
site from the Small census but grants no source-scope exemption, so these two
raises are mandatory rather than slack: audit scope=all 628/182 → 629/183, debt
scope=untagged 421/123 → 422/124. `TESTING.md` was regenerated with the
documented `-update` flag, not hand-edited.

### Follow-up verification

- [x] `go test ./internal/testpolicy/resourcecensus/` — green (`git add -A` first;
      the census reports a **false green** on unstaged files)
- [x] `go build ./...`, `go vet ./...`, `./internal/workrecord/`,
      `./internal/api/` (ok 91s), `./cmd/gc/ -run WorkRecord -timeout 40m`
      (12 PASS, non-vacuous)
- [x] **Mutation proof — three mutations killing three different row sets.**
      A new untagged `exec.Command` in a fresh `internal/workrecord` test turns
      **all three** rows red, so tomorrow's untagged call is still caught. The
      same call added *inside* the Medium-owned test leaves Small green but turns
      **both source rows** red — the Medium row is not a hole. Moving the git call
      to a package-level helper with the row still present turns **only Small**
      red, proving lexical ownership is what exempts.
- [x] `validateMediumOwners` makes the ledger itself fail if
      `TestCommitReachableOnBranch` is ever deleted, so the exemption cannot rot
      silently.

### Alternative considered and rejected

Moving the production git call into `internal/git` as an `IsAncestor` method
would reuse that package's existing test repo at zero census cost. Rejected:
raw `exec.Command("git", …)` is idiomatic at ~28 non-test sites here so
`internal/git` is not an enforced boundary; it edits a widely-shared package for
a census reason; and it would leave the delegation's positive path unpinned — a
`return false` stub would kill no test.

Reviewed by two independent fable lenses (ratchet-integrity,
extraction-fidelity); both returned approve-with-nits, and the first re-ran every
mutation itself. The extraction was mechanically diffed function-by-function
against the pre-move original and found byte-faithful. Findings filed as
`ga-hw5f7` (P1 — the CLI and API close doors disagree for class-binding-resident
beads under `GC_WORK_RECORD_ENFORCE=1`; this PR strictly *narrows* that hole,
since the HTTP door was previously ungated entirely), `ga-rq7km` (census expiry
clock) and `ga-yx235` (contract tables left in `cmd/gc`).

---
- Related to #2844 — extends the ADR-0009 typed work-outcome close gate to the HTTP close and closed-status-update routes, so a task bead closed through the API is checked for the same gc.work_outcome record the CLI door already required; it does not introduce that record or a general close-status signal, and the gate stays warn-only by default, so the missing-signal gap described in the filed issue is not covered here
- Related to #5037 — moves the commit-reachability check verbatim into a new internal/workrecord package, gives it its first real-repository test, and applies the same gate on the API path; it does not change the bare-branch ref lookup that issue reports, and the task-only scoping that exempts bug beads is carried over unchanged


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277236102

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277236171

julianknutsen:
Rebased onto `main` at **fd2ea329b2167ef3f6025fb4d57719616c008907** and retargeted the base from the dead `residency/storeref-series` to `main`.

The old head carried 46 stale commits from the s3–s8 stack, which had been force-rewritten underneath it — that lineage colliding with its own squashes on main was almost all of the 65 merge conflicts and the `dirty` mergeable state. The branch is now a fresh cut from `origin/main` (`faf4e3b89d`) carrying only this segment's two unique commits:

- `8bcd226862` → `29a6aac63f` — fix(workrecord): run the ADR-0009 close gate on the API close path
- `9deababd33` → `fd2ea329b2` — test(workrecord): own the gate's git subprocess instead of leaking two

Net diff against main: 11 files, +977/−228.

## Conflicts and how they were resolved

**1. `cmd/gc/work_record_gate.go` — the #5055 two-probe port (the one semantic conflict).**

This PR was cut before #5055 (`81acdc08a0`, fixes #5037), which rewrote `gitCommitReachableOnBranch` on main to probe `refs/remotes/origin/<branch>` via `preferredReachabilityRef`/`commitReachableOnEitherRef` before falling back to the bare branch. This PR's extraction copies the *old* single `merge-base --is-ancestor` body, so taking either side verbatim was wrong: main's side would leave `internal/workrecord` (and therefore the new API door) on the pre-#5055 behaviour, and this PR's side would regress #5037 on the CLI door.

Resolved by porting #5055 **into** the package. `PreferredReachabilityRef` and `CommitReachableOnEitherRef` now live in `internal/workrecord` as exported functions with main's doc comments intact; `CommitReachableOnBranch` wires the two real git probes through them; `cmd/gc`'s `gitCommitReachableOnBranch` is a one-line delegation. Main's `TestPreferredReachabilityRef` and `TestCommitReachableOnEitherRef` moved with the code into `internal/workrecord/gate_test.go` unchanged apart from the exported names (the stale census rationale in the first one's doc comment was rewritten, since this package *does* own a real git fixture).

Because those two tables inject fakes, they alone would not catch a regression in the *wiring*. `TestCommitReachableOnBranch` therefore gained two rows against the real repository — a branch whose remote-tracking ref is ahead of the local ref (the reported #5037 shape) and its mirror, a local branch ahead of origin. Verified by mutation: reverting `CommitReachableOnBranch` to the single-probe body fails `commit_reachable_only_from_the_remote-tracking_ref`, and restoring it passes.

**2–4. Census renumbering (`internal/testpolicy/resourcecensus/census.go`, `test/test-resources.toml`, `TESTING.md`).**

Recomputed on top of main's *current* baselines rather than restoring this PR's originals (the ratchet is shrink-only). Main is at all-source subprocess 657/188 and untagged 444/126; this segment's one new test-side `exec.Command` (the `runGit` closure inside `TestCommitReachableOnBranch`) moves both by exactly +1 call / +1 file, confirmed by running the census test rather than by arithmetic:

| row | main | this PR |
| --- | --- | --- |
| Audit baseline, all tracked test source, subprocess | 657 / 188 | **658 / 189** |
| Source debt ratchet, all untagged, subprocess | 444 / 126 | **445 / 127** |

The Small subprocess ratchet is unmoved at 437/122 — the git process is a declared `[[medium]]` owner confined to `TestCommitReachableOnBranch`. `TESTING.md` was regenerated with `-update`, not hand-edited. The commit message of `fd2ea329b2` was updated so its stated numbers match what actually shipped.

The API door's honest-degrade behaviour is unchanged from what this PR proposed — a class-binding-resident bead has no checkout to name, so only the reachability clause degrades to a logged warning while a missing or invalid outcome still refuses. Converging that with the CLI door's `repoFallback=cityPath` is `ga-hw5f7`, a separate bead, deliberately not folded in here.

No OpenAPI or generated-TS churn: the refusal is the already-registered `ConflictWrongState` 409.

## Gates

| gate | result |
| --- | --- |
| `go build ./...` | 0 |
| `go vet ./cmd/gc ./internal/api/... ./internal/workrecord/...` | 0 |
| `gofmt -l` (touched trees) | 0, no output |
| `golangci-lint run ./cmd/gc/... ./internal/api/... ./internal/workrecord/...` | 0, "0 issues." |
| `./scripts/check-residency-boundary.sh` | 0, no output |
| `go test ./internal/workrecord/... ./internal/api/... -count=1` | 0 |
| `go test ./cmd/gc -run 'WorkRecord\|workRecord' -count=1` | 0 (12 subtests) |
| `go test ./internal/testpolicy/resourcecensus/... -count=1` | 0 |
| `make dashboard-ci` | 0, no artifact churn |
| `make test-fast-parallel` | 0, "All fast jobs passed" |
| pre-push hook (fast suite re-run) | 0 |

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277285602

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277305675

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277305683

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR closes a real hole in the
ADR-0009 work-record contract: the HTTP API's bead-close path wrote straight
through to the store without ever asking the close gate, so a shipped bead could
be closed over the wire with no work record at all — a gap that only existed
because the contract lived inside `cmd/gc` where the API plane could not reach
it. Lifting it into `internal/workrecord` and wiring both HTTP close spellings
through it means the gate is now a property of the contract rather than a
property of whichever door you happened to use, which is exactly the shape this
invariant needed.

branch.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **The extraction is verbatim, and we checked rather than assumed.** All seven
  moved bodies were diffed byte-for-byte (modulo renames), and
  `EnforceEnabled`/`Gated`/`ValidateOnClose`/`PreferredReachabilityRef`/
  `CommitReachableOnEitherRef`/`CommitReachableOnBranch` were line-compared
  including the #5037 remote-tracking precedence and the flag-shaped-value
  rejections. `cmd/gc/work_record_gate.go` is thin delegation with no stale
  callers, and the moved test tables carried over unchanged.
- **The gate covers the whole wire surface, not just the obvious door.** Both
  spellings gate (`huma_handlers_beads.go:761` close, `:873` closed-status
  update), and the PATCH alias shares `humaHandleBeadUpdate`, so it is gated by
  construction. A fresh census of every bead-close call site in non-test
  `internal/api` found no remaining ungated per-bead door for the gated
  population — DELETE and the convoy/bulk-teardown paths are each outside it for
  documented, checked reasons.
- **The submitted-record projection cannot be used to escape the gate.**
  Coverage is decided on the stored row before the overlay is applied, and the
  overlay's additive-merge semantics were verified against all four store
  implementations, so a request cannot slip past by stamping `gc.kind` or
  retyping on the way out.
- **The census bump is checked, not declared.** TESTING.md,
  `internal/testpolicy/resourcecensus/census.go`, and `test/test-resources.toml`
  agree byte-for-byte on 658/189 and 445/127 with the new
  `internal/workrecord / TestCommitReachableOnBranch` Medium owner row, and the
  live-tree enforcement test passes.
- **No new wire surface.** The refusal reuses the already-registered
  `apierr.ConflictWrongState`; `openapi.json` is untouched and
  `TestOpenAPISpecInSync` passes.

Review findings addressed:
- **Error handling / resilience (minor):** `CommitReachableOnBranch` shelled out
  to `git` with no context, and both HTTP doors discarded the request context
  before reaching it — so a client that hung up left a blocking subprocess with
  no way out, and one that retried accumulated them. Fixed in `6e4b03a0`, which
  adds `CommitReachableOnBranchContext` and threads `ctx` from
  `humaHandleBeadClose` and `humaHandleBeadUpdate` through `gateWorkRecordClose`,
  leaving the background spelling for the CLI door.
- **Architectural consistency (minor):** rig-create idempotency records are
  `gc.kind`-less `task` beads, so they sat inside the population
  `workrecord.Gated`'s own doc says it excludes. Fixed in `6e4b03a0` by stamping
  `gc.kind=task` in `createIdemRecord`, matching the precedent in
  `internal/dispatch/control.go`.
- **Concurrency / state safety (minor):** there is a TOCTOU window between the
  row the gate validates and the unconditional close/update write. `6e4b03a0`
  records it as a known limit on `gateWorkRecordClose`, naming
  `beads.ConditionalWriter` as the remedy and why it is a follow-up rather than
  part of this change.
- **Cross-door divergence (from the earlier iteration):** binding-owned beads
  answered "no repository" on the API plane while the CLI plane fell back to the
  city checkout. `a496031b` resolves a store no configured rig claims to
  `s.state.CityPath()`, and the behavior is now test-pinned on both planes.

- `ga-xfzpb`: in legacy shared-file mode, `buildStores` hands every file-mode rig
  the same store handle, so the gate's store-identity loop can ask the wrong
  repository about a bead — a misleading warning today, a false refusal once
  enforcement is on. Confined to legacy multi-rig file mode; you had already
  found and filed this exact mechanism.
- `ga-t0jcy`: a closing update that converts a non-task bead *into* a task closes
  ungated, because coverage is decided on the stored row. Deliberate, symmetric
  with the CLI door, and documented at the call site.
- `ga-kr5sb`: the legacy bare `work_dir` key fallback should be decided once in
  `internal/workrecord` and applied to both doors, rather than added on one plane.
- `ga-rmdfc`: `resolveScopeRoot` now has three consumers and two drifted copies
  (`internal/api/scope_root.go` trims and guards `cityPath`; the `internal/sling`
  copy does neither). Pre-existing duplication this PR joins, not one it creates.
- `ga-7o0zq`: the operator-facing gate warning envelope is the one
  contract-adjacent string still kept as two per-plane copies, and they have
  already drifted.
- `ga-udcrt`: comments in `internal/dispatch` still name `isWorkRecordGatedBead`
  as the population authority; after the extraction the definition is
  `workrecord.Gated`.

One rollout note worth carrying, not a finding: turning on
`GC_WORK_RECORD_ENFORCE` now also refuses recordless closes from the dashboard
and HTTP automation. That is the contract's intent, but the blast radius of
flipping that flag grew with this PR.

### Maintainer Changes
Two maintainer commits pushed directly to the PR branch, 7 files changed, 303
insertions(+), 54 deletions(-):
- `a496031b` fix(api): ask the city checkout about a binding-owned bead's commit
- `6e4b03a0` fix(api): bound the work-record gate's git probe on the request context

Your commits are preserved with original authorship. Since approval, `main`
advanced by four commits; the adopted branch was rebased onto the latest base
with an unchanged patch-id and re-validated there (build, vet, the live-tree
census enforcement test, the focused work-record suites on all three packages,
and the residency-boundary check all green), which confirmed the base move
carries zero content delta — so the reviewed head stands as-is.

### Final Review Status
Ready for the merge queue: final head `6e4b03a0` has passing required GitHub

CI: https://github.com/gastownhall/gascity/actions/runs/34170245567

### Review Iterations
One review pass — three independent reviewers plus synthesis, with full
cross-file coverage — followed by one maintainer fix commit applying the
pre-merge checklist. No blockers and no majors were found; the three minor
findings on that checklist are all fixed in the final head.

---

PR #5602, opened 2026-08-25
feat(beads): answer the relic census with one store-level predicate

julianknutsen:
## Summary

The boot census asks a binding one question — *"does any bead in here carry an id
none of your namespaces claims?"* — and answered it by listing the binding's whole
history, hydrating every row's `bead_json` into a `Bead` so it could read the id
and throw the rest away. Only the TRUE verdict is safe to remember, because only
TRUE is monotone, so **the city that has never held a relic is exactly the city
that re-pays that scan on every one-shot process, forever** — in front of `gc bd`'s
by-id door, the busiest one-shot route in the CLI.

`beads.NamespaceCensus` is an optional `Store` capability that answers it
directly, implemented on the SQLite and DoltLite stores as one statement over the
id column with `LIMIT 1`. It is a capability rather than a `ListQuery` field on
purpose: the question is a **verdict**, and `ListQuery` fields carry a
superset-plus-`ApplyListQuery` contract every backend would then have to honour —
there is no safe superset of a verdict. `storeref`'s verdict path discovers it and
otherwise lists exactly as before, which is what a binding served by the native
Dolt engine still does. The drain report (`OpenLegacyResidents`) keeps returning
the ids themselves unconditionally, because a verdict cannot name them.

The predicate is spelled `substr`/`length` rather than `LIKE prefix || '%'`:
`LIKE` would swallow the neighbouring `gcgx` namespace and would read `%` and `_`
inside a prefix as wildcards. The prefixes are compile-time reserved-class
constants at every call site and are still bound as parameters.

Follow-on to the residency storeref split stack (`ga-j4p3y`), under epic
`ga-qdt5y`. Closes `ga-dx4ho` — which specifies exactly this work ("add the
namespace predicate to `beads.ListQuery` (or a dedicated store capability
interface), implement it on the sqlite and Dolt stores, and have
`storeref.LegacyResidents` use it … for the verdict form (`HasLegacyResidents`)
while `OpenLegacyResidents` keeps returning the full id list the operator drain
report prints") and names the closed-relic row as its own acceptance test.

### Discovery is gated, not asserted

`TestEmittingClassStoreKeepsEveryEngineCapability` holds `cmd/gc`'s emitting class
store to every method either engine declares, so the wrapper has to carry
`HasResidentOutside` whatever it wraps. **Carrying it as a plain forward is worse
than dropping it**: the wrapper would then satisfy a bare
`store.(beads.NamespaceCensus)` over *any* backing — including a mem store or a
native-Dolt backing that cannot answer at all — and the honest-looking
`(false, nil)` **retires the residence probe over beads that are still there.**
That is the one direction this verdict is never allowed to be wrong in.

The same hazard is already solved in the same file, for atomic closes, with a
comment that describes this situation almost word for word: *"a bare type
assertion would advertise the capability even over a backing … that cannot honor
it — and that discovery is contractually a hard capability gate, not a rollout
seam."* So this follows `AtomicConditionalCloserFor` rather than inventing a
shape:

1. `internal/beads` gains `NamespaceCensusHandleProvider`, `NamespaceCensusFor`,
   and `ErrNamespaceCensusUnsupported`. The handle is consulted **before** the
   plain assertion — the opposite of `GraphApplyFor` — because the method-set
   guard forces the wrapper to carry the method, so assertion-first would take
   the method's word and never reach the handle.
2. `legacyResidentVerdict` discovers via `NamespaceCensusFor`, so the scan
   fallback still reaches a wrapped store whose backing cannot answer.
3. The wrapper implements both. Its handle returns the **backing's** census, and
   its structural `HasResidentOutside` returns `ErrNamespaceCensusUnsupported` —
   never `(false, nil)` — so it cannot lie to a caller holding it directly.

It deliberately does **not** follow `ConditionalWritesResolveTarget`.
`internal/beads/class_store.go:76-81` scopes that seam to the one capability
where forgetting the unwrap fails silently rather than loudly; an undiscovered
census merely scans and returns the same verdict.

**Scope correction, kept from the original review:** the stranding is *latent on
this branch, not live.* The boot census runs on the leaf stores before
`withCLIEmission` wraps them (`cmd/gc/storage_boot.go` `censusBindingRelics` vs
`cmd/gc/cli_storage_routes.go:158`), and `HasLegacyResidents` has one caller. The
mechanism is real — nothing stops the next caller from handing a wrapped binding
to the verdict — but no live data-integrity bug was observed, and the record
should say so.

## What changed since the previous head (`4c02f3d8f5`)

**Base retargeted onto `main`, and that is the load-bearing change.** The old
head sat on `residency/storeref-series` (#5597). This one is `main` at
`c6bb2aa30d` plus exactly one commit — and `c6bb2aa30d` is PR A of the split
stack, **#5644, merged 2026-09-08 02:10Z**, which is what carried the
**closed-inclusive** verdict onto `main`.

The predicate's contract requires counting CLOSED rows
(`internal/beads/namespace_census.go`: *"Implementations MUST count CLOSED beads
and BOTH tiers"*), and that only agrees with a `storeref` verdict that is itself
closed-inclusive. On `main` before #5644 the verdict was open-only
(`HasOpenLegacyResidents`), so landing this ahead of it would have answered TRUE
where the scan it replaces answered FALSE — silently changing the boot verdict
out of order, or failing the `closed relic only` equivalence row. On this base
the verdict is `storeref.LegacyResidents` / `HasLegacyResidents`
(`IncludeClosed: true`) and the two agree by construction. Every open-only
spelling in the verdict path is gone; `OpenLegacyResidents` survives only as
`gc storage status`'s drain gauge, which is deliberate and has its own pins
(`TestOpenLegacyResidentsIgnoresAClosedRelic`,
`TestTheDrainReportIgnoresTheCensusCapability`,
`TestClosingTheLastRelicDrainsTheCountAndKeepsTheProbe`).

Four consequences of the retarget, all in this branch's favour:

- The `internal/storeref/relic_census_bench_test.go` hunk that used to depend on
  s6's `1b070b2d33` is now a plain file add. No s6/s7/s8 dependency remains, and
  no unmerged branch is in the base: this is `origin/main` and one commit.
- The segment applies verbatim. Cherry-picked onto `main` with **no conflicts**
  — `internal/storeref/relic_census.go` and `relic_census_test.go` auto-merged
  around the one thing #5644's own review added after this branch was cut,
  `ProvenLegacyResidents`, and not one line of this segment's production or test
  logic changed.
- `ProvenLegacyResidents` (base-owned, from #5644, read by
  `cmd/gc/by_id_relic_proof.go`) deliberately **does not** come through the new
  fast path, and this PR leaves it on the listing. It reads the OPEN-only census
  while the predicate is closed-inclusive by contract, so routing it here would
  silently widen what a by-id read is allowed to DENY on. Its own doc already
  advertises that swap as a separate, deliberate change ("swap
  `OpenLegacyResidents` for `LegacyResidents` and every caller inherits it"); it
  is not this PR's to make.
- The bench numbers were **re-measured on this base** rather than carried over.

**Squashed to one commit (`mc-xzso`).** The old head's `feat` half alone ships a
bare `store.(beads.NamespaceCensus)` at the verdict site *plus* a wrapper the
method-set guard forces to carry the method — a latent false-clean that only the
`fix` half closes. The two must never be bisectable apart, so they are now one
commit, together with the third head commit (the `api.State` doc fix below).

**`api.State` doc reconciled with the widened verdict.**
`ClassBindingHasLegacyResidents` still documented "still holds an **open** bead",
the opposite of what the value means after #5644. It now reads "a bead, open or
closed", matching the wording `storageRoutes.relics` and
`ClassBinding.HasLegacyResidents` already carry. Doc-only; no wire, schema or
handler change (`make dashboard-ci` green, spec unchanged).

**One review nit closed (`mc-wuwe`), doc-only.** The helper's comment claimed
`length()` counts in the units `substr()` indexes in "whatever the prefix
encodes". That is true on SQLite — which is what *both* implementations execute
on, the DoltLite read store included, since it opens the pure-Go sqlite driver —
but not universally: on a MySQL-executed engine `LENGTH()` is bytes while
`SUBSTR()` indexes characters. The comment now says which engine it is exact on
and what a third caller on the native Dolt engine would have to spell instead.

`mc-9m9s` (typed-nil store) needs no change and is not a hazard on this path:
`(*SQLiteStore).ensureOpen` returns `ErrStoreClosed` for a nil receiver, the
verdict turns that into an error, and `HasLegacyResidents` reads an error as
"keep the probe" — the predicate path is *safer* here than the scan it replaces.

### Measured (re-run on this base — `main` + this commit — `-benchtime 20x`)

| benchmark (10k rows, same seeded fixture) | ns/op |
| --- | --- |
| `BenchmarkRelicCensusCleanBinding_10k` (predicate) | **1.97 ms** |
| `BenchmarkRelicCensusScanFallback_10k` (what this replaced) | 89.35 ms |
| `BenchmarkRelicCensusOpenOnly_10k` | 49.89 ms |
| `BenchmarkRelicCensusCleanBinding_1k` (predicate) | 0.43 ms |

~45x on the clean-binding case — the case that cannot be memoized away, since
only the TRUE verdict is monotone.

The predicate is still O(rows): a NOT-prefix predicate cannot seek an index, and
the code comment says so in as many words. What it stops being is a **hydration**.

## Testing

RED first, then GREEN. Each mutation below was forced, observed, and reverted;
the failing sets are **disjoint**, so each pins a different property rather than
all three re-testing the answer. Mutation 1 was **re-run on this base** after the
rebase onto `main`; 2 and 3 were taken on the previous base over a segment whose
production and test logic is byte-identical to this one.

**1. Closed-inclusive — the property the retarget exists for.**
Adding `status <> 'closed'` to both predicates (SQLite and DoltLite):

| test | result |
| --- | --- |
| `TestSQLiteHasResidentOutside/a_closed_relic_is_still_a_resident` | **FAIL** |
| `TestDoltliteHasResidentOutside/a_closed_relic_is_still_a_resident` | **FAIL** |
| `TestTheCensusPredicateAndTheScanAgree/closed_relic_only` | **FAIL** |
| `cmd/gc TestBootCensusKeepsTheProbeForAClosedRelic` | **FAIL** |
| `cmd/gc TestClosingTheLastRelicDrainsTheCountAndKeepsTheProbe` | **FAIL** |
| `TestSQLiteHasResidentOutside` — other 11 rows | PASS |
| `TestDoltliteHasResidentOutside` — other 10 rows | PASS |
| `TestTheCensusPredicateAndTheScanAgree` — other 9 rows | PASS |
| `cmd/gc TestBootCensusKeepsTheProbeForAMigratedBead` | PASS |
| `cmd/gc TestBootCensusRetiresTheProbeOnACleanBinding` | PASS |

The `cmd/gc` rows are the end-to-end half: those boot gates run against real
SQLite bindings, so with this change they now reach the **predicate**, and the
closed-relic verdict is pinned through the whole boot path rather than only at
the store.

**2. The answer, independent of which rows it counts.** Forcing the predicate
always-false additionally killed `TestBootCensusKeepsTheProbeForAMigratedBead`
and six equivalence rows (`open relic`, `closed relic only`, `wisp relic`,
`lookalike prefix`, `binding claiming no namespace`, `blank namespace`), while
`clean binding`, `closed in-namespace only`, `bare prefix id` and `empty binding`
correctly survived.

**3. The routing, independent of the answer.** Reverting the verdict's discovery
to a bare `store.(beads.NamespaceCensus)` killed
`TestTheVerdictReachesThePredicateThroughAWrapper` and
`TestTheVerdictScansThroughAWrapperWhoseBackingCannotAnswer` (both rows) and
**nothing else**, with `internal/beads` green as a control. Making the verdict
skip discovery entirely — always scan — killed a *different* set:
`TestLegacyResidentsVerdictUsesTheCensusCapability`,
`TestLegacyResidentsVerdictTakesTheCensusCapabilitysCleanAnswer`,
`TestACensusCapabilityErrorKeepsItsProbe` and
`TestTheVerdictReachesThePredicateThroughAWrapper`, and nothing else. Both
directions pin the capability gate rather than the verdict.

### Gates

| command | exit |
| --- | --- |
| `go build ./...` | 0 |
| `go vet ./...` | 0 |
| `CGO_ENABLED=0 go vet -tags gascity_native_beads ./internal/beads/` | 0 |
| `gofmt -l cmd internal` | 0, no output |
| `golangci-lint run ./internal/storeref/... ./internal/beads/... ./cmd/gc/...` | 0, `0 issues` |
| `./scripts/check-residency-boundary.sh` | 0 (baseline needed no growth; the segment adds no `residency:allow`) |
| `go test ./internal/storeref/... ./internal/beads/... -count=1` | 0 |
| `go test ./cmd/gc -run 'Relic\|Census\|Residency\|Storage' -count=1` | 0 |
| `make test-native-doltlite-beads` | 0, verified **non-vacuous**: `TestDoltliteHasResidentOutside` (11 subtests, incl. the closed-relic row and two wisp rows) and `TestDoltliteReadStoreIsANamespaceCensus` all PASS under `-v` |
| `make check-docs` | 0 |
| `make dashboard-ci` | 0 — run because `internal/api/state.go` is touched; no generated artifact moved |
| `make test-fast-parallel` | 0 (first invocation returned 75 without running anything — the local `push-gate` slot cap timed out waiting on two other suites on this host; re-run once a slot freed) |

## Checklist

- [x] Linked an issue — `ga-dx4ho`, under epic `ga-qdt5y`, stack `ga-j4p3y`
- [x] Added or updated tests for behavior changes — including an explicit
      predicate-vs-scan equivalence suite, so the fast path cannot answer
      differently from the listing form it replaces
- [x] Updated docs for user-facing changes — none; this is a boot-path
      optimisation with an unchanged verdict. The one doc edit is the `api.State`
      interface comment, corrected to the semantics #5644 already shipped.
- [x] Called out breaking changes or migration notes — none. `NamespaceCensus` is
      **optional**: a store that does not implement it lists exactly as before.

Known follow-ups, filed and not blocking: `ga-j1stb` (`splittest` forwards
`GraphApplyHandle`/`ConditionalWriterHandle` but not `NamespaceCensusHandle`, so
kit leaves scan while production predicates — a fidelity gap the equivalence
suite already pins the answer for) and `ga-55w1j` (the DoltLite pre-wisps-migration
skip has no test row; it mirrors `List`'s skip and fails conservatively).

### DIFF
--- a/cmd/gc/class_store_emit.go
+++ b/cmd/gc/class_store_emit.go
@@ -538,6 +538,39 @@ func (s *emittingClassStore) AtomicConditionalCloserHandle() (beads.AtomicCondit
 	return s, true
 }
 
+// HasResidentOutside forwards the relic census to the backing capability. It
+// exists because TestEmittingClassStoreKeepsEveryEngineCapability holds this
+// wrapper to every engine method, and a wrapper that carries the method must
+// still be unable to answer for a backing that cannot: a store with no census
+// gets beads.ErrNamespaceCensusUnsupported, never (false, nil). A false clean
+// here would retire the binding's by-id probe over every relic sitting in it.
+//
+// Discovery does not reach this method — NamespaceCensusHandle below answers
+// first — so this is the spelling for a caller holding the wrapper that asks
+// directly.
+func (s *emittingClassStore) HasResidentOutside(prefixes []string) (bool, error) {
+	census, ok := beads.NamespaceCensusFor(s.Store)
+	if !ok {
+		return false, fmt.Errorf("censusing the id namespaces of the emitting class store over %T: %w", s.Store, beads.ErrNamespaceCensusUnsupported)
+	}
+	return census.HasResidentOutside(prefixes)
+}
+
+// NamespaceCensusHandle keeps beads.NamespaceCensusFor honest over this
+// wrapper, exactly as AtomicConditionalCloserHandle does above and for the same
+// reason: the reflective capability guard forces HasResidentOutside to EXIST
+// for every engine, so a bare type assertion would advertise a census over a
+// backing — a mem store, the native Dolt engine — that has none. The census
+// contract makes that a hard gate rather than a rollout seam, because a store
+// that answers the question inexactly strands every bead it failed to see.
+//
+// It hands back the backing's census rather than the wrapper: the wrapper adds
+// emission, a census reads and mutates nothing, so there is nothing here for
+// the answer to pass back through.
+func (s *emittingClassStore) NamespaceCensusHandle() (beads.NamespaceCensus, bool) {
+	return beads.NamespaceCensusFor(s.Store)
+}
+
 func (s *emittingClassStore) DeleteIfMatch(id string, revision int64) error {
 	writer, ok := beads.ConditionalWriterFor(s.Store)
 	if !ok {
--- a/cmd/gc/class_store_emit_test.go
+++ b/cmd/gc/class_store_emit_test.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"go/ast"
 	"go/parser"
@@ -22,6 +23,7 @@ import (
 	"github.com/gastownhall/gascity/internal/config"
 	"github.com/gastownhall/gascity/internal/coordclass"
 	"github.com/gastownhall/gascity/internal/events"
+	"github.com/gastownhall/gascity/internal/storeref"
 )
 
 // The three gates this seam has to hold, and the control that keeps the second
@@ -593,6 +595,91 @@ func TestEmittingClassStoreKeepsEveryEngineCapability(t *testing.T) {
 			t.Errorf("the emitting class store drops %v from %s; every one is a capability assertion that stops matching", missing, engine)
 		}
 	}
+}
+
+// The relic census is the one capability where carrying the method is WORSE
+// than dropping it, so TestEmittingClassStoreKeepsEveryEngineCapability forcing
+// HasResidentOutside to exist needs this row underneath it.
+//
+// Dropping the method costs a scan that returns the same verdict. Carrying it
+// as a plain forward makes a bare assertion succeed over a backing that has no
+// census — the mem store here, the native Dolt engine in production — and the
+// caller then retires the binding's by-id probe on an answer nobody computed,
+// stranding every bead the class migration carried across under its original
+// id. Both halves are pinned: discovery must refuse a backing that cannot
+// answer, and it must find one that can and agree with the scan.
+func TestTheEmittingClassStoreOnlyAdvertisesACensusItCanAnswer(t *testing.T) {
+	const graphPrefix = "gcg"
+
+	t.Run("a backing with no census is not advertised as one", func(t *testing.T) {
+		cityPath := t.TempDir()
+		wrapped := splitClassRoutes(beads.NewMemStore()).withCLIEmission(cityPath).stores[coordclass.ClassGraph]
+
+		direct, carried := wrapped.(beads.NamespaceCensus)
+		if !carried {
+			t.Fatal("the wrapper does not carry HasResidentOutside at all; this row models the hazard of carrying it and has nothing to say")
+		}
+		if census, ok := beads.NamespaceCensusFor(wrapped); ok {
+			t.Fatalf("a wrapper over a backing with no census was discovered as one (%T); the boot verdict would take its word and retire the binding's probe", census)
+		}
+		got, err := direct.HasResidentOutside([]string{graphPrefix})
+		if !errors.Is(err, beads.ErrNamespaceCensusUnsupported) {
+			t.Fatalf("asking the wrapper directly returned (%v, %v), want beads.ErrNamespaceCensusUnsupported: a structural method with nothing behind it must not answer a verdict", got, err)
+		}
+		if got {
+			t.Error("the refusal came back with a TRUE verdict; a store that cannot answer answers nothing")
+		}
+	})
+
+	t.Run("a sqlite backing is reached and agrees with the scan", func(t *testing.T) {
+		for _, tc := range []struct {
+			name string
+			seed []string
+			want bool
+		}{
+			{name: "binding holding a relic", seed: []string{"gcg-1", "ga-relic"}, want: true},
+			{name: "clean binding", seed: []string{"gcg-1", "gcg-2"}, want: false},
+		} {
+			t.Run(tc.name, func(t *testing.T) {
+				cityPath := t.TempDir()
+				leaf, err := beads.OpenSQLiteStore(t.TempDir(), beads.WithSQLiteStoreIDPrefix(graphPrefix))
+				if err != nil {
+					t.Fatalf("opening the leaf store: %v", err)
+				}
+				t.Cleanup(func() { _ = closeBeadStoreHandle(leaf) })
+				for _, id := range tc.seed {
+					if _, err := leaf.Create(beads.Bead{ID: id, Title: id, Type: "task"}); err != nil {
+						t.Fatalf("seeding %q: %v", id, err)
+					}
+				}
+
+				wrapped := splitClassRoutes(leaf).withCLIEmission(cityPath).stores[coordclass.ClassGraph]
+				census, ok := beads.NamespaceCensusFor(wrapped)
+				if !ok {
+					t.Fatalf("the wrapper hid the sqlite leaf's census; every one-shot command on this city goes back to scanning the whole binding")
+				}
+				predicate, err := census.HasResidentOutside([]string{graphPrefix})
+				if err != nil {
+					t.Fatalf("predicate: %v", err)
+				}
+
+				relics, err := storeref.LegacyResidents(leaf, []string{graphPrefix})
+				if err != nil {
+					t.Fatalf("scan: %v", err)
+				}
+				scanned := len(relics) > 0
+				if scanned != tc.want {
+					t.Fatalf("the scan answered %v, want %v (relics %v); the fixture does not model what it claims to", scanned, tc.want, relics)
+				}
+				if predicate != scanned {
+					t.Fatalf("the census through the wrapper answered %v and the scan answered %v (relics %v)", predicate, scanned, relics)
+				}
+				if got := beadEvents(readCityJournal(t, cityPath)); len(got) != 0 {
+					t.Errorf("censusing the binding appended %d bead event(s); a verdict reads and mutates nothing: %s", len(got), eventSummary(got))
+				}
+			})
+		}
+	})
 }
 
 // TestEmittingClassStoreCarriesAnEdgePayloadAndEmitsForIt covers the wrapper's
--- a/internal/api/state.go
+++ b/internal/api/state.go
@@ -172,9 +172,10 @@ type State interface {
 	OrdersBeadStore() beads.OrdersStore
 
 	// ClassBindingHasLegacyResidents reports whether store — one of the
-	// per-class stores above — still holds an open bead under an id outside the
-	// namespaces its classes declare: a row `gc storage migrate` carried across
-	// under its original work-shaped id, reachable only by probing the binding.
+	// per-class stores above — still holds a bead, open or closed, under an id
+	// outside the namespaces its classes declare: a row `gc storage migrate`
+	// carried across under its original work-shaped id, reachable only by
+	// probing the binding.
 	//
 	// The API never opens a binding and cannot take that census itself, so the
 	// verdict crosses this surface from the boot that did. It exists because
--- a/internal/beads/doltlite_namespace_census.go
+++ b/internal/beads/doltlite_namespace_census.go
@@ -1,0 +1,51 @@
+//go:build gascity_native_beads
+
+package beads
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"fmt"
+)
+
+var _ NamespaceCensus = (*DoltliteReadStore)(nil)
+
+// HasResidentOutside reports whether the store holds any bead whose id none of
+// prefixes claims. It satisfies NamespaceCensus.
+//
+// Both storage tables are asked, because both are places a carried-across row
+// can sit and TierBoth applies no discriminator to either — the tier and
+// storage-flag predicates List layers on for the narrower modes are the ones
+// TierBoth deliberately omits. Neither the cross-table dedupe nor the
+// storage-flag classification changes the answer: a row that exists in either
+// table is a resident whichever tier it is classified into, and a wisp shadowed
+// by its durable twin shares that twin's id.
+//
+// The wisps table is optional — snapshots written before the upstream wisps
+// migration have none — so its absence is a table with no rows in it, not a
+// failed read.
+func (s *DoltliteReadStore) HasResidentOutside(prefixes []string) (bool, error) {
+	where, args := namespaceExclusionSQL("COALESCE(i.id, '')", prefixes)
+	for _, tables := range doltliteTableSetsForMode(TierBoth) {
+		if tables.wisps && !s.tableExists(tables.issues) {
+			continue
+		}
+		query := "SELECT 1 FROM " + tables.issues + " i"
+		if where != "" {
+			query += " WHERE " + where
+		}
+		query += " LIMIT 1"
+
+		var found int
+		switch err := s.db.QueryRowContext(context.Background(), query, args...).Scan(&found); {
+		case errors.Is(err, sql.ErrNoRows):
+			continue
+		case err != nil:
+			return false, fmt.Errorf("censusing the id namespaces of doltlite table %s: %w", tables.issues, err)
+		default:
+			return true, nil
+		}
+	}
+	return false, nil
+}
--- a/internal/beads/doltlite_namespace_census_test.go
+++ b/internal/beads/doltlite_namespace_census_test.go
@@ -1,0 +1,155 @@
+//go:build gascity_native_beads
+
+package beads
+
+// The DoltLite half of the namespace census, held to the same bar the
+// hydration-free Count is: whatever the predicate says, listing the store and
+// applying the rule by hand must say the same thing.
+
+import (
+	"strings"
+	"testing"
+)
+
+func doltliteResidentOutsideByScan(t *testing.T, store *DoltliteReadStore, prefixes []string) bool {
+	t.Helper()
+	rows, err := store.List(ListQuery{TierMode: FederatedReadTier, AllowScan: true, IncludeClosed: true})
+	if err != nil {
+		t.Fatalf("List: %v", err)
+	}
+	for _, row := range rows {
+		inside := false
+		for _, prefix := range prefixes {
+			prefix = strings.TrimSpace(prefix)
+			if prefix == "" {
+				continue
+			}
+			if row.ID == prefix || strings.HasPrefix(row.ID, prefix+"-") {
+				inside = true
+				break
+			}
+		}
+		if !inside {
+			return true
+		}
+	}
+	return false
+}
+
+func TestDoltliteHasResidentOutside(t *testing.T) {
+	classPrefixes := []string{"gcg", "gcnq"}
+
+	for _, tc := range []struct {
+		name     string
+		issues   []testDoltliteIssue
+		wisps    []testDoltliteIssue
+		prefixes []string
+		want     bool
+		why      string
+	}{
+		{
+			name:     "a work-shaped id the migration carried across is a resident outside",
+			issues:   []testDoltliteIssue{{ID: "gcg-1", Title: "own"}, {ID: "ga-relic", Title: "carried across"}},
+			prefixes: classPrefixes,
+			want:     true,
+			why:      "ga-relic carries no namespace this binding declares",
+		},
+		{
+			name:     "every declared namespace is inside",
+			issues:   []testDoltliteIssue{{ID: "gcg-1", Title: "own"}, {ID: "gcnq-2", Title: "own"}},
+			prefixes: classPrefixes,
+			want:     false,
+			why:      "a binding holding only its own ids is the converged city the probe retires for",
+		},
+		{
+			// The load-bearing row (ga-qdt5y.19): a closed relic is still read,
+			// reopened and claimed by id.
+			name:     "a closed relic is still a resident",
+			issues:   []testDoltliteIssue{{ID: "gcg-1", Title: "own"}, {ID: "ga-done", Title: "drained", Status: "closed"}},
+			prefixes: classPrefixes,
+			want:     true,
+			why:      "closing a relic drains it; it does not make it findable by prefix",
+		},
+		{
+			name:     "a closed bead inside a declared namespace is ordinary retired infrastructure",
+			issues:   []testDoltliteIssue{{ID: "gcg-1", Title: "own", Status: "closed"}},
+			prefixes: classPrefixes,
+			want:     false,
+			why:      "it is findable by prefix whether it is open or closed",
+		},
+		{
+			name:     "an empty binding holds nothing outside",
+			prefixes: classPrefixes,
+			want:     false,
+			why:      "nothing is what a fresh born-split city holds",
+		},
+		{
+			// The wisps table is a second place a carried-across row can sit, and
+			// a predicate that read one table would miss it.
+			name:     "a relic in the wisps table counts",
+			issues:   []testDoltliteIssue{{ID: "gcg-1", Title: "own"}},
+			wisps:    []testDoltliteIssue{{ID: "ga-wisp", Title: "carried across", Ephemeral: true}},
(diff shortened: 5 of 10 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277235871

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277245621

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277245694

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR answers the relic ("residency") census with a single store-level SQL predicate instead of a full `List(TierBoth)` scan, and gates that capability behind a handle so a backing that cannot answer the census falls back to the by-id scan rather than reporting a false "clean." That matters because the census verdict decides whether the residency probe may be retired — a false "clean" would retire the probe while a relic still exists.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **Capability honesty (no false "clean"):** Discovery now routes through `NamespaceCensusFor`, which consults `NamespaceCensusHandleProvider` *before* the plain `store.(beads.NamespaceCensus)` assertion. `emittingClassStore`'s structural method returns a typed `ErrNamespaceCensusUnsupported` instead of a `(false, nil)` lie, so a mem/native-Dolt backing falls back to the by-id scan rather than falsely retiring the probe. (`cmd/gc/class_store_emit.go`, `internal/beads/namespace_census.go`, `internal/storeref/relic_census.go`)
- **Predicate ≡ scan equivalence:** The SQL predicate `(id = ? OR substr(id,1,length(?)) = ?)` is an exact transcription of `IDInNamespace`, and the census reads a superset of `List(TierBoth)`'s id set, so `census=FALSE ⇒ scan=FALSE` — the safety-critical direction (never retire the probe while a relic exists) is monotone-safe. Pinned by `TestTheCensusPredicateAndTheScanAgree` (10 seeded cases, including the `gcgx` lookalike and bare-`gcg` arms) and `TestACensusCapabilityErrorKeepsItsProbe`.

- **[Documentation / Maintainability — nit]** `namespaceExclusionSQL` assumes char-semantics `length()`/`substr()`; a future byte-`LENGTH()` census engine reusing the helper would mis-compare a multibyte prefix. Harmless today — all reserved-class prefixes are ASCII and no such implementer exists. Optional one-line caveat comment. Filed as `mc-wuwe`.
- **[Change Hygiene / Release Safety — nit]** The feat commit ships the probe-retirement hazard and only the fix commit closes it; the two commits must land together. Squash-merge satisfies this. Filed as `mc-xzso`.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `3ad198c` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/32897497632

### Review Iterations
1 review pass, 0 fix iterations — both reviewers (design-implementation + implementation) independently reached approve with zero blocker/major/minor findings; only the two non-gating nits above remain.

---

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR answers the beads "relic census" — *does this binding hold a bead whose id no declared namespace claims?* — with a single store-level `NamespaceCensus` predicate (`HasResidentOutside(prefixes)`) that resolves the question in one id-only SQL statement instead of a full-hydration `List` scan. That removes a hot-path scan from the one-shot `gc bd` by-id CLI while leaving census-incapable stores (native Dolt, mem, bd) on the existing scan, so the change is a pure, revertible addition behind capability discovery.

This PR was reviewed and accepted as-is; the branch was rebased onto the latest `residency/storeref-series` base to keep it mergeable (no content change).

### Original PR Review
Decision: approve.

Specific gaps fixed:
- The load-bearing claim — that the fast predicate returns the **same** verdict as the scan it replaces — was validated structurally, not just against fixtures. The SQLite predicate reproduces `IDInNamespace` byte-for-byte (bare-prefix arm and empty-fragment→match-all branch included), the census reads the same population the scan reads (`TierBoth`, shared missing-wisps guard), and every failure edge (nil store, census error, scan error, missing table) fail-safes to "keep the probe." Both the SQLite and DoltLite implementations are exercised, and native Dolt / mem / bd correctly fall back to the existing scan.

- **mc-wuwe** — [maintainability / nit] `namespaceExclusionSQL` (`internal/beads/namespace_census.go`) documents `length()`/`substr()` as counting in the same units, which holds only under a SQLite executor; on a real Dolt/MySQL engine `LENGTH()` is bytes while `SUBSTR()` is characters, so a future multi-byte-prefix caller there could over-read. Cannot occur today (prefixes are ASCII reserved-class constants and no MySQL-executed caller exists). Suggested: a one-line SQLite-only caveat, or `CHAR_LENGTH` if the helper is ever taken cross-engine. Not addressed in this PR.
- **mc-9m9s** — [test evidence / nit] typed-nil `beads.Store` inputs are not exercised by the new discovery path. This follows the repo's usual interface-nil convention, so it is a coverage note, not a suspected defect. Suggested: add a typed-nil-store row to pin the nil→"keep the probe" edge explicitly. Not addressed in this PR.

### Maintainer Changes
No maintainer content changes — PR accepted as-is. The branch was rebased onto the latest `residency/storeref-series` base to clear a merge conflict and keep it mergeable; the patch itself is unchanged (patch-id preserved; 9 files, +1180/-9).

### Final Review Status
Ready for the merge queue: final head `1418494174` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/33076353504

### Review Iterations
1 review pass performed — approved on the first pass with zero fix iterations.

---

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277291253

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277291322

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277291548

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR answers the beads relic census — *does this binding still hold a bead whose id no declared namespace claims?* — with a single store-level `NamespaceCensus` predicate (`HasResidentOutside(prefixes)`), so the verdict comes from one id-only SQL statement instead of a full-hydration `List` scan of the store. That turns a per-boot full listing into a single predicate query on SQLite and DoltLite bindings while census-incapable backends (native Dolt, mem, bd) keep the existing scan behind capability discovery — a cleanly revertible addition that makes the residency probe materially cheaper without changing the answer any backend gives.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **The load-bearing equivalence claim was verified structurally, not just against fixtures.** The SQLite predicate `(id = ? OR substr(id,1,length(?)) = ?)` reproduces the Go rule `IDInNamespace` (`internal/storeref/class_candidates.go:165`) exactly — including the bare-prefix row and the `gcgx-1` lookalike that a `LIKE 'gcg%'` spelling would have swallowed. Blank prefixes are dropped, and prefixes reach SQL only as bound parameters.
- **The capability handle is taken before the type assertion** (`14438f9e4c`), matching the established `AtomicConditionalCloserFor` ordering (`internal/beads/beads.go:242`). This closed a live hazard rather than a theoretical one: `TestEmittingClassStoreKeepsEveryEngineCapability` reflectively forces the wrapper to carry `HasResidentOutside` regardless of what it wraps, so a bare assertion would have advertised a census the backing cannot answer.
- **Every failure edge fail-safes to "keep the probe"** — nil store, census error, scan error, missing table, mint bit off. Nothing fails open; and because nothing ever deletes a relic, the widened verdict is monotone, so staleness can only keep a probe alive, never retire one.
- **Test evidence is strong.** The DoltLite table cross-checks the predicate against a hand-applied scan of the same store per row, `TestLegacyResidentsVerdictUsesTheCensusCapability` asserts `listCalls == 0` against a deliberately contradictory fixture, and `TestBootGateTakesTheCensus` boots through the real gate.

Review findings addressed:
- **[contract fidelity / minor]** The `ClassBindingHasLegacyResidents` doc at `internal/api/state.go:175` still described the pre-widening semantics ("still holds an **open** bead…"). The staleness was introduced by this PR's own widening, and it was the one site the sweep came short of after correctly updating `cmd/gc/storage_boot.go` and `internal/storeref/topology.go`. Fixed in `4c02f3d8f5` — doc-only, no behavior change, no test impact.

- **ga-zby3f** — [maintainability / minor] The CLI one-shot plane discards the census verdict: `withCLIEmission` rewrites `r.stores` to `emittingClassStore` wrappers (`cmd/gc/class_store_emit.go:119`) without re-keying `r.relics`, so every later `hasLegacyResidents(wrapper)` misses the map and takes the pessimistic `true` default. Each `gc` one-shot pays for the census and then keeps probing anyway. Deliberately not fixed here: `git blame` puts the whole mechanism outside this PR (`0d39380ada`, `47b317bc63b`), neither `cli_storage_routes.go` nor `residency_topology.go` is in the diff, and the map miss is fail-safe — a strict superset read, so the plane can only do redundant work, never under-report. This PR in fact makes that pre-existing waste much cheaper, since the wasted census is now one predicate query instead of a full store listing. Not addressed in this PR.
- **ga-eptjz** — [test evidence / low] No DoltLite test row exercises a snapshot that has `beads` but no `wisps` table, so the optional-leg `tableExists` skip is unpinned. That guard mirrors the pre-existing scan guard exactly and lands pessimistic on error, so this is a coverage note rather than a suspected defect. Not addressed in this PR.

### Maintainer Changes
Two maintainer commits were pushed to the PR branch; contributor authorship on `efdb1cf66a` and `14438f9e4c` is preserved.

- `a4cde986d1` `fix(storeref): count closed relics in the residence-probe verdict` — a **rebase prerequisite**, not part of this PR's deliverable. It is the `ga-qdt5y.19` half of upstream `8f9eeee96b`, which never landed on the base this PR was rebuilt onto. This PR's own contract (`NamespaceCensus`: "Implementations MUST count CLOSED beads") is written against that widened verdict, so without it the adopted head would answer the retirement question closed-inclusively on SQLite bindings and open-only on every other backend — a backend-divergent split present in neither the PR nor the base. The `ga-qdt5y.18` half of the same upstream commit, which collapses `bdByIDClassDoor.resolve` onto storeref's by-id plan, was deliberately excluded as read-routing change this PR does not need. 9 files, +236/-49.
- `4c02f3d8f5` `fix(api): match the State doc to the widened relic-census verdict` — the doc correction described above. 1 file, +4/-3.

### Final Review Status
Ready for the merge queue: final head `4c02f3d8` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/33746689810

### Review Iterations
1 review pass performed. It returned `approve` with a single minor, non-gating doc-fidelity item, which was applied as `4c02f3d8f5` rather than deferred. Local gates were re-run in the adopted worktree: `go build ./...` clean, `go vet ./internal/beads/... ./internal/storeref/...` clean, and the `RelicCensus|LegacyResident|Census` and `NamespaceCensus|HasResidentOutside` suites green. The branch was then refreshed onto the latest `residency/storeref-series` base with no content change.

---

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277311327

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR replaces the relic census's unbounded full-history scan with a single store-level existential predicate (`beads.NamespaceCensus` / `HasResidentOutside`), so a binding can answer "do I still hold a bead outside my own namespaces?" without sweeping its entire bead history. That verdict is load-bearing on `gc bd`'s by-id door and runs once per process at boot for every binding, so the cost fell on exactly the case that becomes the steady state — a converged, clean binding. Re-measured during this review: **2.01 ms** for the predicate against **95.8 ms** for the scan fallback on a clean 10k binding.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **Predicate/scan equivalence — the core hazard.** A false-clean verdict retires the by-id probe and strands every bead the census failed to see, so "the two paths read the same population" had to be proven, not assumed. Re-derived structurally: both paths run at `FederatedReadTier = TierBoth`, the DoltLite census iterates exactly `doltliteTableSetsForMode(TierBoth)` = `{issues, wisps}`, and the SQL rule reproduces `storeref.IDInNamespace` clause-for-clause including the `TrimSpace` and empty-prefix-claims-nothing rules. The shipped equivalence suite pins it row-by-row against the scan.
- **`substr`/`length` rather than `LIKE`.** Using the length-anchored form closes two real holes a `LIKE` prefix match would have left open: `%`/`_` in a prefix being read as wildcards, and a neighbouring namespace like `gcgx` being swallowed by a `gcg` match.
- **Every failure direction keeps the probe.** `legacyResidentVerdict` returns the error rather than a verdict, and `HasLegacyResidents` reads any error as `true` — deliberately not falling back to the scan. `TestACensusCapabilityErrorKeepsItsProbe` pins it.
- **The wrapper never fabricates a clean.** `emittingClassStore.HasResidentOutside` returns `ErrNamespaceCensusUnsupported` when its backing has no census; it never returns `(false, nil)`, which would be the false-clean answer.
- **Discovery is provider-before-assertion.** `NamespaceCensusFor` consults `NamespaceCensusHandleProvider` first, matching `AtomicConditionalCloserFor`, so a wrapper that structurally carries the method cannot outvote the real handle.
- **The naming half of the report stayed on the scan.** `OpenLegacyResidents` / `ProvenLegacyResidents` are untouched — a boolean verdict cannot name ids, and the drain report still needs to.

Review findings addressed:
- **`idExpr` non-NULL precondition was undocumented** (contract fidelity, nit) — the helper's correctness depends on `idExpr` never evaluating to SQL NULL, because under `NOT (...)` a NULL comparison drops the row out of the `WHERE` entirely, and a lost row answers clean over a resident. Both callers already guarded with `COALESCE`, but the doc never said so, so a third caller passing a bare or outer-joined column would have lost rows silently. Fixed in `6592244d96`.
- **Bench fixture prose overstated its population** (readability, nit) — the comment said "mostly closed" while `closedFraction=2` closes exactly half. Reworded in `6592244d96`.

- `ga-55w1j` (minor, test evidence): the DoltLite missing-wisps-table skip branch is the one new predicate branch no test row reaches — the shared seeder always builds the full schema, so the guard is asserted only by its doc comment. Low risk because it is a verbatim copy of the List path's own guard; a row here is what would catch the two later diverging.
- `ga-j1stb` (minor, contract fidelity): `splittest.StrictStore` does not forward the new `NamespaceCensusHandle`, so a conformance-kit leaf silently takes the scan while production takes the predicate. Fail-safe direction (same verdict, slower), and the same file already documents this class of gap for `Count` and `Claim`.
- `ga-28myo` (nit, consistency): `BenchmarkRelicCensusOpenOnly_10k` inline-duplicates the shared `benchmarkCensusStore` seeder, giving two drift surfaces to a benchmark whose point is comparing like populations.
- `ga-emxu1` (nit, resource lifecycle): the two new `internal/storeref` SQLite fixtures do not register cleanup, unlike the other two SQLite fixtures this PR adds.
- `ga-jnsqd` (nit, no action now): both census reads use `context.Background()`. Correct for a once-per-boot in-process SQLite read; recorded only so the decision is revisited if a request-path caller ever appears.

One tracker action was taken during adoption rather than in the PR: `ga-1ghfh` was closed as a **duplicate of `ga-dx4ho`** — not as "remedied by this PR" — because `ga-dx4ho` predates it by two weeks and is the primary owner of this work. Its deliberate residual is recorded there: bindings served by the native Dolt engine still take the full scan by design, since they reach rows through the beads library and hold no SQL of their own.

### Maintainer Changes
One maintainer fixup, comment-only, no behavior change: `6592244d96 docs(beads,storeref): document the idExpr non-NULL precondition` — 2 files changed, 8 insertions(+), 1 deletion(-).

### Final Review Status
Ready for the merge queue: final head `6592244d` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

The adopted branch was also refreshed onto the latest `main` (`f3610ae8` → `94636378`) as a finalization check. The rebase was clean with a stable patch-id, none of the two intervening commits touch any file this PR touches, and all ten touched blobs are byte-identical across the refresh — so the reviewed content is exactly what merges. Validated at the refreshed composition anyway: `go build ./...` and `go vet` clean, `scripts/check-residency-boundary.sh` and `scripts/check-census-owner-liveness.sh` exit 0, and 160 focused test rows across `internal/beads`, `internal/storeref`, and `cmd/gc` (including the `gascity_native_beads` build flavor) passed with zero failures.

CI: https://github.com/gastownhall/gascity/actions/runs/34198089636

### Review Iterations
One review pass across three independent lanes (design/implementation, implementation, cross-file/architecture), reaching approval with no gating findings and zero fix iterations required. Seven findings surfaced, all minor or nit; two were taken as an optional maintainer doc fixup and the remaining five are filed as the follow-ups above.

---

PR #5603, opened 2026-08-25
perf(scripts): screen the residency grep census with each pattern's required literal

julianknutsen:
## Summary

The Go half of the residency boundary guard ran all 25 vocabulary regexes against
every line of every file under the scan dirs: **11.04s** of the `./scripts` wall
clock, and growing with both the tree and the pattern file. A guard that slow is a
guard people start skipping, so the cost is a correctness risk and not just an
annoyance.

Each pattern now carries a `core` — a literal string every line it can match MUST
contain — and the scanner does `strings.Contains(line, core)` before the regex.
The core is derived mechanically by walking the `regexp/syntax` parse tree and
keeping only literals on the unconditional path (concat children and
unconditional captures), skipping alternations and quantifiers entirely and
refusing case-folded literals.

**Deriving from the parse tree rather than from the pattern text is the whole
point.** Text-level metacharacter stripping turns

```
(^|[^A-Za-z0-9_])cliSoleClassBinding(Store)?\(
```

into the core `cliSoleClassBindingStore(`, which no bare `cliSoleClassBinding(`
call site contains — the guard would have gone on reporting a green census while
silently counting fewer sites than the pattern file asks for.

A screen that is too narrow fails silently, so the derivation cannot be allowed to
fail quietly either. `residencyPatternCore` returns an **error** rather than an
empty string when it can find no required literal, and `loadResidencyPatterns`
`t.Fatalf`s on it — an unscreenable row stops the suite instead of scanning
unscreened.

Soundness is proved by **census equivalence**, not by argument. The suite
synthesizes a minimal matching witness line per pattern from the same parse tree,
writes them to a real on-disk fixture tree, asserts every pattern is actually
censused there, then scans the tree with the prefilter off and on and fails on any
single-row diff. Because the fixture body is generated from the pattern file, a
vocabulary row added later joins that control the day it lands, with no second
hand-kept list to forget.

**`TestResidencyBoundaryGrepRatchet`: 11.04s → 0.40s (~28x).**

Follow-on to #5597 (epic `ga-qdt5y`). Closes `ga-laswd`. Test-only; no production
code changes.

## Testing

- [x] `gofmt -l scripts/` — no output
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./scripts/` — ok 84.4s (whole package: the shell/Go halves-agree
      test and all four residency rules)
- [x] `go test ./scripts/ -run 'TestResidencyBoundaryGrep|TestResidencyPatternCore' -v`
      — 3 top-level PASS, 17 sub-row PASS, ok 0.46s
- [x] Pre-commit hook ran on the staged change: `lint-changed ./scripts` — 0
      issues; docs/schema regeneration produced no diff
- [x] **Mutation proof — each a real on-disk edit, reverted after.** An AST/grep
      guard cannot be falsified with an in-memory overlay, so none of these used
      one:
      - *core derived text-level instead* (the sabotage above): the load-time
        self-check died on `the derived core "cliSoleClassBindingStore(" is absent
        from "cliSoleClassBinding("`; with that check neutered so the row could be
        observed independently, the equivalence row died naming
        `cmd/gc/witness_census.go everyPattern b:cliSoleClassBinding: 1 with the
        prefilter off, 0 with it on`, and 3 of 4 derivation unit rows died.
      - *prefilter never applied*: ONLY `the equivalence row bites an over-narrow
        core` died, and the ratchet's wall time went back to 10.681s.
      - *prefilter always applied, ignoring the off mode*: ONLY `the prefilter-off
        census ignores the core` died, naming all 25 patterns. That row exists
        because an off side that quietly screened would make the equivalence row
        compare screened against screened — green and vacuous.
      - *derivation returns `""` instead of an error*: both `a case-folded pattern
        has no sound core` and `a pattern with no literal has no core` died.
      - *witness set to the core itself*: the load-time `MatchString(witness)`
        check died, proving the witness check is not circular.

      Different tests die each way. All seven pre-existing control rows pass
      before and after, and **the baseline does not move**.

## Checklist

- [x] Linked an issue — `ga-laswd`, under epic `ga-qdt5y`
- [x] Updated docs for user-facing changes — none; this is guard machinery
- [x] Called out breaking changes or migration notes — none. The census verdict is
      proved identical row-for-row; only the time it takes changed.

### DIFF
--- a/scripts/residency_grep_rule_test.go
+++ b/scripts/residency_grep_rule_test.go
@@ -2,10 +2,12 @@ package scripts_test
 
 import (
 	"bufio"
+	"fmt"
 	"io/fs"
 	"os"
 	"path/filepath"
 	"regexp"
+	"regexp/syntax"
 	"strings"
 	"testing"
 )
@@ -24,11 +26,45 @@ import (
 type residencyPattern struct {
 	name  string
 	regex *regexp.Regexp
+
+	// core is a literal string every line this pattern can match MUST contain,
+	// derived from the parsed pattern. strings.Contains(line, core) is
+	// therefore a sound screen in front of the regex: it can only skip lines
+	// the regex would have missed anyway.
+	core string
+
+	// witness is a synthesized line the pattern matches. It is the load-time
+	// proof that core was derived from a REQUIRED position, and it is what the
+	// equivalence control builds its on-disk fixture out of — so a vocabulary
+	// row added to the pattern file is exercised by that control on the day it
+	// lands, with no second hand-kept list.
+	witness string
 }
+
+// residencyPrefilter selects whether the census screens each line with a
+// pattern's required literal before running that pattern's regex.
+//
+// It is a parameter rather than a constant because the prefilter's entire
+// contract is that it changes NOTHING: the equivalence control runs the same
+// census both ways over the same on-disk tree and refuses a single differing
+// row. "It got faster" is not evidence that it still counts the same sites,
+// and a guard that has quietly stopped biting is worse than a slow one.
+type residencyPrefilter bool
+
+const (
+	residencyPrefilterOn  residencyPrefilter = true
+	residencyPrefilterOff residencyPrefilter = false
+)
 
 // loadResidencyPatterns reads the forbidden vocabulary from the file the shell
 // guard reads. One source of truth: two hand-kept copies of "what is forbidden"
 // would be this guard's own bug class, one level up.
+//
+// It also derives each row's prefilter core and self-checks the derivation. A
+// derivation that cannot be made fails the whole suite here rather than
+// degrading to an unscreened scan: the census is the product, the prefilter is
+// only its cost, and a guard is allowed to be slow but never allowed to be
+// quietly narrower than its own pattern file.
 func loadResidencyPatterns(t *testing.T, dir string) []residencyPattern {
 	t.Helper()
 	f, err := os.Open(filepath.Join(dir, "residency-boundary-patterns.txt"))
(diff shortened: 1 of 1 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277235792

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR adds a per-pattern
required-literal prefilter (`strings.Contains(line, core)`) in front of the
residency grep census's 25 unanchored regexes — with the `core` derived
mechanically by walking the `regexp/syntax` parse tree rather than stripping
pattern text. That cuts `TestResidencyBoundaryGrepRatchet` from ~11s to ~0.4s
(~28x), which matters because a boundary guard that slow is a guard people start
skipping: the speedup keeps the residency census fast enough to stay trustworthy.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- No gaps required fixing — the change was accepted as-is. The review instead
  independently validated the load-bearing properties, since a silently
  over-narrow screen would let the guard report a green census while counting
  fewer sites than the vocabulary asks for:
  - **The prefilter never drops a counted line.** The `core` is harvested only
    through must-match parse nodes (`OpLiteral`/`OpConcat`/`OpCapture`) and never
    through `OpAlternate`/`OpQuest`/`OpStar`/`OpRepeat`, so it is a required
    contiguous substring of every match — the screen can only skip lines the
    regex already rejects. (All 25 vocabulary rows were re-derived by hand.)
  - **Deriving from the parse tree, not the pattern text, is what makes it
    sound.** The `(^|[^A-Za-z0-9_])cliSoleClassBinding(Store)?\(` row is the
    load-bearing case: text stripping yields `cliSoleClassBindingStore(` (silently
    over-narrow), while the parse-tree walk stops at the `(Store)?` and yields
    core `cliSoleClassBinding`, a substring of both the bare and `Store` spellings.
  - **The equivalence controls can actually fail.** The `\x00`-suffix sabotage
    makes every screen reject lines its regex would match; the ON-side control
    asserts the census changes and the OFF-side control asserts it does not, both
    on real on-disk edits under `t.TempDir()`.
  - **The derivation fails loud.** `residencyPatternCore` returns an error (never
    `""`) when no required literal exists, and `loadResidencyPatterns` `t.Fatalf`s
    on it, so an unscreenable row stops the suite instead of scanning unscreened.

- **ga-8ot8m** — Test Evidence Quality (nit): the equivalence controls prove
  prefilter off==on only on the *synthetic witness fixture*; the real ratchet runs
  the actual repo tree prefilter-on only, so real-tree equivalence rests on the
  soundness argument plus each row's load-time witness self-check rather than a
  both-ways census diff of the real tree. This is a belt-and-suspenders gap, not a
  live defect, and was filed as `ga-8ot8m` for later — it is not addressed here.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `610be92d` has passing required GitHub

CI: https://github.com/gastownhall/gascity/actions/runs/32877353979

### Review Iterations
1 review pass performed. The PR was approved on the first pass with no fix
iterations; the only finding was the non-gating test-evidence nit noted above.

---

julianknutsen:
Merging on the pipeline's own recorded decision: this PR completed adopt-pr review and was stamped status/merge-ready by workflow finalize, but its pipeline bead was lost in the 08-26 wedge churn, so the bead-driven merge patrol can never pick it up (verified: no bead carries pr_review.pr_number=5603). mergeStateStatus=CLEAN at merge time. — operator, on behalf of pr-merge-ready-patrol

PR #5610, opened 2026-08-25
Give runtime-provider waivers independent expiries and a live owner

quad341:
## What this changes

The runtime-provider contract ledger now gives each waived constructor its own explicit expiry date and points waiver ownership at the active provider-conformance follow-up. This prevents one shared deadline or a dead tracking owner from silently invalidating every provider entry at once.

The checked table in `TESTING.md` is regenerated from the Go ledger, so contributors see the same owner and expiry that the policy tests enforce.

## Review notes

- This changes test-governance metadata, not runtime provider selection or session behavior.
- All eight runtime-provider waivers currently expire on September 7, but the dates are stored independently so later decisions can move one constructor without changing the others.
- The existing bounded-waiver validation remains in force; no dependency, configuration, or migration change is introduced.

## Test plan

- [x] Run `go test -count=1 -v ./internal/testutil/providerledger/...` and confirm all 37 top-level tests pass with no skips.
- [x] Run the full local CI-equivalent suite and audit every non-candidate failure against the release-gate attribution record.
- [x] Run policy, docs, build, vet, native-DoltLite, formatting, and affected-lint checks.
- [x] Release gate: [`release-gates/ga-rt0f61-runtime-provider-waiver-expiry-gate.md`](release-gates/ga-rt0f61-runtime-provider-waiver-expiry-gate.md)


julianknutsen:
Fable review council (2 independent lenses: mechanism, governance) — **approve-with-nits, 0 merge blockers**. Merging now to clear the fleet-wide outage.

**Reproduced both ways.** `origin/main` fails `TestCatalogMatchesProductionWiringAndDocumentation` with exactly the 8 `expired 2026-08-26` errors; this head passes the whole package; `git merge-tree --write-tree origin/main 5cff38529e` is conflict-free and the merged tree is green. `.githooks/pre-push:55` execs `make test-fast-parallel`, which includes the unit-core job — so this really was blocking every push in the fleet, not just PR CI.

**Verified clean:** metadata-only (4 files; zero Go importers of `internal/testutil/providerledger` anywhere; absent from `cmd/gc`'s dep graph); the validator still rejects zero, past and beyond-horizon expiries after the signature change (mutation-proved three ways); the TESTING.md block genuinely renders from the Go ledger (drifting one row's date fails `CheckMarkdown`); all eight waived constructors are still live registrations in `cmd/gc/runtime_registry.go`.

**Findings filed as follow-ups rather than blocking** — none makes the fleet better off staying red:

1. **The lockstep lapse is not actually fixed.** All eight "independent" expiries are the byte-identical `time.Date(2026, time.September, 7, …)` literal, and no test asserts they differ. On 2026-09-07 all eight lapse together and reproduce this incident byte-for-byte. The capability to diverge was added; the divergence was not.
2. **Blast radius is unchanged.** A pure calendar event in a leaf package nothing imports reds `main` for everyone. This is the third instance (2026-08-12, 2026-08-26) with a fourth pre-loaded. The same anti-pattern is worse next door: all 37 waivers in `test/test-resources.toml` expire on the same day, 2026-10-01, enforced by the same `time.Now()` path in `internal/testpolicy/resourcecensus`.
3. **`ledger_test.go` unpins the waiver owner.** Three assertions that compared against the literal `"ga-80po0c.3"` now compare against `runtimeContractWaiverOwner` — the same constant `waivedRuntime` populates the field from. Proved self-referential: renaming the constant to `ga-DEADBEEF` (with TESTING.md updated to match) leaves the package green at this head, while the identical mutation on `main` fails four tests.
4. **The "dead ga-80po0c owner" premise is false in this repo.** `ga-80po0c.3` is OPEN, updated today, with an in-progress child; `ga-uz5t3a` resolves to nothing here. The same commit leaves 36 of 37 `test/test-resources.toml` waivers owned by `ga-80po0c.*`, so the repo now says that program is simultaneously dead and responsible for 36 census waivers. Accountability regression, not a correctness one.
5. **Four other branches were racing this same lapse** with divergent answers (different owner, different date). Whoever reads this: this SHA is the one that landed.

Follow-up beads filed on the gascity ledger; `ga-80po0c.3` gets the merged SHA.

PR #5611, opened 2026-08-25
fix(beads): stop the cache re-absorbing its own reconcile emissions (ga-yoix1)

julianknutsen:
## The loop

The controller feeds every bead event on the bus back into `ApplyEvent` on the emitting `CachingStore` (`startBeadEventWatcher` -> `applyBeadEventToStores`). A reconcile emission is that store's own post-absorb snapshot -- but `Bead.Dependencies` and `Bead.Needs` are `omitempty`, so a bead with **no dependencies** marshals with neither key. On the wire that snapshot is indistinguishable from a bd `on_update` patch, which legitimately omits dependencies after a removal and must be treated as coverage-unknown.

So `ApplyEvent` guesses coverage-unknown, and that guess is not a lost optimization -- it is a latch:

1. `updateEventDepsLocked` drops the cached dep set, clears the `is_blocked` verdict reconciliation just installed, and clears `depsComplete` **store-wide**.
2. It reports a mutation, which stamps a sequence number.
3. That sequence fences the row out of the next pass' absorb (`mergeSkipFenced`), so the cleared verdict is never repaired.
4. Every later pass compares cached `is_blocked` nil against fresh `&false`, calls that a change, and emits again -> back to 1.

## Production signature

On maintainer-city this was **98% of the event bus**: 11937 of ~12600 events in a 20 MB tail, 2353 distinct beads each emitted repeatedly with a **byte-identical** payload, sustained above 2000 events/minute against a completely idle backing store (~3.4M events/day, matching the ~70 MB-gz-per-75-min log rotation).

Per-rig churn was bimodal -- 100% or 0%, never in between -- which is the tell: `depsComplete` is a store-wide latch, so a single bead's wipe degrades the dependency comparison for the entire rig, and any rig with no initial diff stays clean forever.

## The fix

Add `CachingStore.ApplyEventSnapshot` for payloads that carry authoritative dependency coverage, and route `cache-reconcile`-actored events through it. The controller already excludes this actor from `cs.Poke()` three lines below; this extends the same distinction to the absorb.

This is the sibling leg of `ga-fnmb5`, which fixed `clearDependentReadyProjectionsLocked` (the *dependent's* projection) but not the `clearReadyProjectionLocked` inside the dependency wipe (the bead's *own*).

Not a blanket actor skip: reconcile events still reach the cache and still update fields, so `TestControllerStateAppliesCacheReconcileBeadEventsToStores` is unchanged, and `admitReadyRoutedWorkEvent` / `admitSessionStartEvent` / the `BeadClosed` arm still run.

## Tests

- `TestReconcileEmissionFedBackDoesNotSustainFloodLoop` reproduces the ring end-to-end against the controller's real topology (emit during the pass, deliver after, as the async bus does). Before the fix: five passes, five emissions, one unchanging dependency-free bead.
- `TestReconcileEmissionFedBackPreservesReadyVerdict` isolates the state damage (`is_blocked` nilled, `depsComplete` cleared).
- `TestApplyBeadEventRoutesCacheReconcileEventsAsSnapshots` pins the controller-side routing, and asserts the `omitempty` premise still holds so the test cannot go vacuous if the wire shape changes.

Both halves are **mutation-checked**: reverting either one reproduces the loop.

### DIFF
--- a/cmd/gc/api_state.go
+++ b/cmd/gc/api_state.go
@@ -48,6 +48,11 @@ import (
 	"github.com/gastownhall/gascity/internal/workspacesvc"
 )
 
+// cacheReconcileActor is the event actor the controller stamps on bead events
+// emitted by a CachingStore's reconciliation pass, distinguishing the cache's
+// own snapshots from foreign writes delivered by a bd hook.
+const cacheReconcileActor = "cache-reconcile"
+
 // controllerState implements api.State, api.StateMutator, and
 // api.ConfigWriteSerializer.
 // Protected by an RWMutex for hot-reload: readers take RLock,
@@ -287,7 +292,7 @@ func wrapWithCachingStore(ctx context.Context, store beads.Store, ep events.Prov
 		if recorder != nil {
 			recorder.Record(events.Event{
 				Type:             eventType,
-				Actor:            "cache-reconcile",
+				Actor:            cacheReconcileActor,
 				Subject:          beadID,
 				RunID:            runID,
 				SessionID:        sessionID,
@@ -753,12 +758,23 @@ func (cs *controllerState) applyBeadEventToStores(evt events.Event) {
 	}
 	cs.mu.RUnlock()
 
+	// A cache-reconcile event carries a CachingStore's own post-absorb snapshot,
+	// dependencies included, rather than a bd hook patch. Saying so keeps the
+	// cache from reading its own emission as a coverage-unknown payload and
+	// discarding the dependency and is_blocked state it just installed, which
+	// fences the row out of the next reconcile pass and re-emits forever
+	// (ga-yoix1).
+	snapshot := evt.Actor == cacheReconcileActor
 	for _, store := range stores {
 		if cached, ok := store.(*beads.CachingStore); ok {
-			cached.ApplyEvent(evt.Type, evt.Payload)
+			if snapshot {
+				cached.ApplyEventSnapshot(evt.Type, evt.Payload)
+			} else {
+				cached.ApplyEvent(evt.Type, evt.Payload)
+			}
 		}
 	}
-	if evt.Actor != "cache-reconcile" {
+	if !snapshot {
 		cs.Poke()
 	}
 	if evt.Type == events.BeadClosed && evt.Subject != "" && len(stores) > 0 {
--- a/cmd/gc/api_state_cache_reconcile_snapshot_test.go
+++ b/cmd/gc/api_state_cache_reconcile_snapshot_test.go
@@ -1,0 +1,61 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"strings"
+	"testing"
+
+	"github.com/gastownhall/gascity/internal/beads"
+	"github.com/gastownhall/gascity/internal/events"
+)
+
+// TestApplyBeadEventRoutesCacheReconcileEventsAsSnapshots pins the controller
+// half of the cache-reconcile flood fix (ga-yoix1).
+//
+// The reconcile emitter marshals a whole absorbed bead, but Dependencies and
+// Needs are omitempty, so a bead with no dependencies produces a payload that
+// looks exactly like a bd on_update patch. Fed back through the plain ApplyEvent
+// path, the cache treats dependency coverage as unknown and clears the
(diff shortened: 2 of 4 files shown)


quad341:
@julianknutsen — this is exceptional debugging. Tracing a 2k-events/minute flood down to `omitempty` making a post-absorb snapshot indistinguishable from a bd `on_update` patch, then showing it's a *latch* rather than a lost optimization (mutation seq → `mergeSkipFenced` → verdict never repaired), is the kind of writeup I wish every bug fix had. The harness that emits during the pass and delivers after — matching the async bus instead of applying inline — is right for exactly the reason you gave, and asserting the `omitempty` premise inside the test so it can't quietly go vacuous is a lovely touch.

One thing I hit while tracing: the emitter sends the row as `List` returned it, but the deps it absorbs come from a separate query. On `FileStore`/`MemStore` and the native-doltlite path, `Bead.Dependencies` is empty even for a blocked bead, so `depsAuthoritative` there asserts "no edges" and `CachedReady` starts offering blocked work. I reproduced it, so I'll take it on my side: narrowing the snapshot to "carries no *new* dep info — don't invalidate" instead of "authoritatively empty," plus a blocked-bead test. Your flood tests pass untouched. Merging with your commits and credit intact.


---

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277245199

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277245268

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277246124

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR stops the beads caching store from re-absorbing its own reconcile emissions. A dep-free cache snapshot fed back through the coverage-unknown path was dropping dependencies, nil-ing the `is_blocked` verdict, clearing `depsComplete`, and fencing the row so the cleared verdict was never repaired — a self-sustaining re-emit loop (thousands of events/minute) against an otherwise idle backing store. Killing that loop removes a real production flood and the stale readiness it produced, which matters directly to controller/dispatch stability (ga-yoix1).

### Original PR Review
Decision: approve.

Specific gaps fixed:
- Root cause confirmed by tracing the code, not by trusting the narrative: the flood is driven by a dep-free snapshot re-absorbed through the coverage-unknown branch. The fix routes snapshot application on `evt.Actor == cacheReconcileActor`, an actor stamped by exactly one site (the cache's own `onChange` recorder in `cmd/gc/api_state.go`). Foreign `bd`-hook dependency removals carry a different actor and still take the normal coverage-unknown path, so the legitimate case the old branch existed for is not masked, and re-applying a snapshot to its origin cache is idempotent.
- Blast radius checked: the only other live `ApplyEvent` consumer, the t3bridge per-session watcher, builds its cache with a nil `onChange` emitter and never reads `IsBlocked`/`Dependencies`, so it neither re-emits nor exposes the clobbered state — not a second flood site.
- Test evidence is topology-faithful: the internal test drives the real reconcile → emit → feedback loop and asserts silence across five passes, a companion test isolates the verdict and `depsComplete` damage, and a controller test pins actor → snapshot routing and guards the `omitempty` premise. Both reviewers verified the stated bug independently rather than accepting the premise at face value.

- ga-d593s (maintainability, minor — design-implementation-reviewer): `refreshedFromBacking || depsAuthoritative` is threaded into the `bead.created`, `bead.updated`, and `bead.closed` arms, but `updateEventDepsLocked` only consults the flag for `bead.updated`; on created/closed it is inert (and harmless) today. Suggested follow-up: a one-line comment documenting that the flag is honored only for `bead.updated`, so a future deps-clearing path under `bead.closed` cannot become silently unguarded. Filed as ga-d593s; not addressed in this PR.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `be9f476` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/32884632713

### Review Iterations
1 review pass performed (0 fix iterations) — approved on the first pass.

---

PR #5612, opened 2026-08-25
fix(ci): mac-regression's path filter was inert, gate the smoke tier properly (#5155)

### LINKED ISSUE #5155: Mac regression: the path filter gates nothing, and the unit tier's timeout is below its runtime
Two defects in `.github/workflows/mac-regression.yml` at `origin/main@5a252c606`. Together they make a green Mac badge compatible with nothing having run.

## 1. `mac_sensitive` is computed and never gates anything

The `filter` step computes `mac_sensitive` (line 136). The `gate` step reads it into `PATH_HIT` (line 148) and uses it in exactly one place:

```bash
else
  reason="pull request without needs-mac label (path hit: ${PATH_HIT})"
fi
```

It is interpolated into a human-readable string. It never sets `run_smoke` or `run_full`. Every Mac job is gated on `needs.gate.outputs.run_smoke == 'true'` or `needs.gate.outputs.run_full == 'true'`, so a same-repo PR that touches Mac-sensitive paths without the `needs-mac` label runs zero Mac jobs and still reports the workflow green.

Measured: 22 of 25 mac-regression runs reported `overall=success` with the unit tier `skipped`. #5097 sat at `Mac Regression: pass` while its own gate logged `run_smoke=false` and exited 0.

#4850 centralized this routing but did not connect the output to a condition.

## 2. `mac-unit` cannot pass at its current budget

`mac-unit` (`Mac / make test`, line 255) carries `timeout-minutes: 25`. Run 31324304262 ran 16:56:20 to 17:21:36, which is 25m16s, and was cancelled at the cap. Nearby budgets in the same file: quality 35, cmd-gc shards 20, acceptance 25, integration bdstore 60.

Fixing 1 without 2 converts a silent skip into a guaranteed cancellation on the unit tier, so the two belong in the same change.

## Not a defect

Fork PRs are skipped deliberately (`PR_HEAD_REPO != github.repository`). That is the self-hosted-runner policy and is out of scope here.

## Open question

Which tiers a path hit should fire is a runner-cost decision, not a technical one. Two reasonable answers: run the full suite on every Mac-sensitive path hit, or run the smoke tier only and keep `needs-mac` as the opt-in for the full sweep. This report does not assume either.

rjgeng:
Fixes #5155.

The `mac-regression` workflow computed a path filter and then never gated anything on it. `PATH_HIT` was set from `steps.filter.outputs.mac_sensitive` and read in exactly one place: the human-readable `reason=` string explaining why the tier was skipped. On a `pull_request` event the only thing that could turn the smoke tier on was the `needs-mac` label, so a PR touching mac-sensitive paths without that label recorded `path hit: true` in its reason and skipped the tier anyway.

This adds the missing branch: a real path hit now fires the smoke tier without requiring a label. Labeled runs are unchanged and still get both tiers.

**Also:** bumped `mac-unit`'s timeout 25m to 35m. It was already routinely running over the old budget, unrelated to this fix but caught in the same file.

**Testing:** YAML-only, no Go. Validated via `yaml.safe_load` plus a manual trace of the gate block (`actionlint` was not available on this machine).

### DIFF
--- a/.github/workflows/mac-regression.yml
+++ b/.github/workflows/mac-regression.yml
@@ -178,6 +178,9 @@ jobs:
             elif [[ "$NEEDS_LABEL" == "true" ]]; then
               run_smoke=true; run_full=true
               reason="pull request carries needs-mac label"
+            elif [[ "$PATH_HIT" == "true" ]]; then
+              run_smoke=true
+              reason="pull request touches mac-sensitive paths without needs-mac label, running smoke tier"
             else
               reason="pull request without needs-mac label (path hit: ${PATH_HIT})"
             fi
@@ -254,7 +257,7 @@ jobs:
       - gate
     if: needs.gate.outputs.run_smoke == 'true'
     runs-on: ${{ needs.runner-policy.outputs.runner_macos }}
-    timeout-minutes: 25
+    timeout-minutes: 35
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
         with:


sjarmak (approved):
This is correct and it closes a defect I hit from the other direction last week, so I want to record the evidence here.

On `main`, `PATH_HIT` is populated (`mac-regression.yml:148`, from `steps.filter.outputs.mac_sensitive` at `:136`) and then consumed in exactly one place: the human-readable reason string at `:182`. Nothing branches on it. A computed value with no non-test consumer is the same defect whether it lives in Go or in YAML, and this puts the consumer in. The new `elif` is also correctly ordered after the fork and draft checks, so neither of those skip paths can be overridden by a path hit.

One correction to the body, since the body is what a future reader will trust. It says the filter being inert meant "the smoke tier ran on every push regardless of whether the changed paths touched anything relevant." It was inert in the opposite direction. `run_smoke` initializes to `false` at `:150`, and the `pull_request` branch only set it true under the `needs-mac` label, so a path hit on an unlabeled PR fell through to the `else` and ran nothing. That matches what I measured on 2026-08-09: 22 of 25 mac-regression runs reported `overall=success` with the unit tier skipped, and #5097 sat at `Mac Regression: pass` while its gate logged `run_smoke=false`. The diff adds a branch that makes the tier run *more*, which is the right fix for the real symptom. Worth rewriting the sentence before merge so the record is not backwards.

On the timeout bump from 25 to 35: no objection to the direction, given the tier is about to start running on a lot more PRs than it has been, and a timeout on a tier nobody was exercising is not evidence of anything. If there is a measured p95 behind the 35, it is worth a line in the body. If it is a margin guess, say that instead; either is fine, but the number will get cited later.

None of the above blocks. Approving.

sjarmak:
Edited the description before merging so the squashed commit message records the right symptom; the code is untouched.

The original body said the inert filter meant the smoke tier "ran on every push regardless." It was inert in the other direction. On a `pull_request` event the gate could only be turned on by the `needs-mac` label, so a PR that touched mac-sensitive paths and lacked the label logged `path hit: true` in its skip reason and skipped anyway. That is the failure this fixes, and it is worth having in history because the symptom it actually produces is a tier that looks green while never running.

Two things I confirmed on the branch:

`PATH_HIT` on main is populated at line 148 from `steps.filter.outputs.mac_sensitive` and consumed nowhere except the `reason=` string at line 182. A value computed, threaded through a job output, and read only by a log line is the shape worth naming: it reads as a working filter to anyone scanning the file.

This PR's own checks show the bug live. The Mac tier reports 11 jobs `skipping`, several with `${{ matrix.shard }}` still unexpanded in the name, meaning those jobs never instantiated. A rollup renders that identically to a pass.

sjarmak:
Ran the cross-provider (Codex, gpt-5.4) pass against the head revision before merging, since a workflow gate change can widen in directions that are invisible in a four-line diff. Five questions, all clean:

**Nothing consumed `PATH_HIT` before this.** The boolean outputs are assigned at lines 156, 159, 162, 166, and 179; none of those reference it. Its only other appearance is the `reason` string at 182.

**No unintended widening.** The new branch sits inside the `pull_request` arm at 173, after the fork guard at 174 and the draft guard at 176, so it cannot fire for fork or draft PRs. `schedule` and `workflow_dispatch` are matched earlier at 155 and 158, and `push`/`merge_group` are not in the `on:` block at all. The only thing it shadows is the final `else` at 181, which previously only set the reason string.

**`run_smoke` without `run_full` is safe here.** The two are consumed by disjoint job sets (smoke at 200, 255, 281, 311; full at 344, 385, 449, 490), and the summary job at 646 keys off `RUN_SMOKE` alone and tolerates skipped full-tier jobs in its pass/fail loop at 673-678. No job assumes the two move together.

**Absent filter output is fail-closed.** A missing `steps.filter.outputs.mac_sensitive` dereferences to an empty string, so `[[ "$PATH_HIT" == "true" ]]` is false and the tier stays off. The wrong direction to fail would have been the other one.

**The timeout bump does not collide.** No workflow-level timeout exists, and sibling jobs already use 35, 45, 60, 75, and 90.

PR #5613, opened 2026-08-25
fix(reaper): expired-nudge cleanup now closes rig-scoped shadows too (#5285)

### LINKED ISSUE #5285: Expired rig-scoped nudge shadows evade the native reaper
## Summary

`gc sling --nudge` can create a `gc:nudge` shadow in a rig's work store, but the core `reaper` order only closes expired nudge issues when their database is the city database. Expired rig-scoped shadows therefore remain open and appear in the rig's `bd ready` output indefinitely.

## Observed on

- Gas City v1.4.0 (latest release)
- One city database: `hq`
- One rig database: `gx`
- Rig-scoped target: `gx-ops/pr.worker`

Nine expired shadows remained open with:

- label `gc:nudge`
- metadata `state=queued`
- `expires_at` between 2026-08-14T03:19:57Z and 2026-08-14T23:14:07Z
- current time 2026-08-15T03:22:28Z
- no assignee

The `reaper` order ran at 2026-08-15T03:18:29Z, but all nine still appeared in `bd ready` afterward.

## Cause in current source

`cmd/gc/cmd_sling.go` passes the sling work store into `deliverSlingNudge`, so the shadow is created in the source/target rig store. In `internal/bootstrap/packs/core/assets/scripts/reaper.sh`, the expired-nudge query scans every database, but the close path increments `expired_skipped` whenever `DB != CITY_DB`; `close_city_issue` is bound to the city store.

`nudge-mail-sweep` does not cover this case because it only closes stale *delivered* nudges that are absent from the live queue. `wisp-compact` is also not a safe fallback: an open wisp past TTL is promoted to permanent as stuck work.

## Expected behavior

Gas City's native maintenance should terminalize/close expired nudge shadows in their owning store, including rig work stores, without requiring a per-rig cleanup script. Expired nudge shadows should no longer appear as executable work in `bd ready`.

## Suggested acceptance

- A nudge shadow created in a rig store is terminalized or closed after `expires_at` by native Gas City maintenance.
- The close is routed through the owning rig store with the same authority checks used for city shadows.
- City-store behavior remains unchanged.
- `wisp-compact` never promotes an expired `gc:nudge` shadow merely because it is still open.
- Coverage includes distinct city and rig Dolt databases.

rjgeng:
Fixes #5285.

The reaper's expired-nudge cleanup only ever closed city-store nudge shadows; rig-scoped ones were silently skipped, forever — they just accumulated with no path to closure.

**Fix:** resolves the rig binding and closes rig-scoped shadows via `gc bd --city --rig`, matching how city-store shadows were already handled.

**Testing:** `go build`/`gofmt -l` clean, targeted `go test` green. Includes new shell-level test coverage for this path — worth noting this script's existing test harness is informal/manual, not CI-wired, so this coverage isn't yet exercised automatically.

---

### DIFF
--- a/internal/bootstrap/packs/core/assets/scripts/reaper.sh
+++ b/internal/bootstrap/packs/core/assets/scripts/reaper.sh
@@ -400,6 +400,35 @@ EOF
     printf '%s\n' "$refs"
 }
 
+# rig_name_for_db looks up the rig name bound to a non-city database, using
+# the same RIG_STORE_REFS_BY_DB map discover_rig_store_refs populates from
+# city.toml/site.toml [[rigs]] bindings. Prints the rig name and returns 0 on
+# a match; returns 1 with no output when $db has no rig binding (e.g. it is
+# neither the city store nor a discovered rig store).
+rig_name_for_db() {
+    local db="$1"
+    local entry
+    local entry_db
+    local store_ref
+
+    while IFS= read -r entry; do
+        [ -n "$entry" ] || continue
+        entry_db="${entry%%|*}"
+        store_ref="${entry#*|}"
+        [ "$entry_db" = "$db" ] || continue
+        case "$store_ref" in
+            rig:*)
+                printf '%s\n' "${store_ref#rig:}"
+                return 0
+                ;;
+        esac
+    done <<EOF
+$RIG_STORE_REFS_BY_DB
+EOF
+
+    return 1
+}
+
 workflow_root_store_ref_local_condition() {
     local db="$1"
     local alias="$2"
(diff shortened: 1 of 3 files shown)


quad341 (approved):
@rjgeng — this one's good, and I want to be clear up front: **there is nothing for you to do.** Sorry it's sat since Aug 25; that delay is on us, not on your diff.

What I appreciate most is that you didn't bolt on a parallel rig branch. You reused the `RIG_STORE_REFS_BY_DB` map the reaper already builds, so `rig_name_for_db` can tell a genuinely rig-bound store apart from an unbound one, and an unbound `$DB` still falls through to exactly the old skip+count path. Inverting the chain to `elif [ "$DB" = "$CITY_DB" ]` instead of grafting onto the negative test is the clean way to do it, and incrementing `DB_MUTATIONS` so the existing per-DB commit gate covers rig closes symmetrically is the detail most people would have missed.

The tests are the standout, and they earned their keep in a way you couldn't have planned for. Because you extract the real function body and the real Step 4 block with `awk` instead of retyping them, I was able to replay both suites against a `main` that has moved **271 commits** since you opened this — and they still pass, 6/6 and 4/4, because they're reading the live script rather than a snapshot of it. A copy-pasted test would have gone quietly stale. Your patch also still applies to current `main` completely clean, and shellcheck is clean. T6 — a db that's a substring of another entry's db must not match — is exactly the case a thinner suite skips and an on-call engineer pays for later.

We also took your `close_rig_issue` argv out for a live run against a throwaway city+rig, and `gc bd --city … --rig … close` closed the bead in the rig store with your `ttl:expired by reaper` reason intact. The flag combination does what your comment says it does.

Our required gate is red for a reason that has nothing to do with you: the `cmd/gc product metrics testhook` job hit its own 5-minute cap at 5m08s and got cancelled, and our gate treats `cancelled` as a failure. That job currently finishes in 4.4–4.8 minutes on a *good* run against that same 5-minute cap, so it was going to bite someone — you just drew the short straw. We're re-running it and raising that timeout on our side.

Merging on green. Your commit and your credit stay exactly as they are — no rebase, no force-push to your branch.

PR #5614, opened 2026-08-25
fix(events): truncate a NUL-padded tail left by an unclean shutdown (#5336)

### LINKED ISSUE #5336: bug: unclean shutdown leaves a NUL-padded tail on events.jsonl, silently truncating every strict JSON Lines consumer
# Upstream report: `gc` should truncate a NUL-padded tail on `.gc/events.jsonl` at startup

**For:** gastownhall/gascity
**Type:** bug (data framing) + enhancement (doctor check)
**Filed from:** paper-town (local bead `pt-gfan`, discovered under `pt-0y7`)
**Prepared:** 2026-08-16

---

## Summary

An unclean shutdown can leave `.gc/events.jsonl` ending in a block that the
filesystem length-allocated but never wrote, which reads back as NUL bytes. On
the next start, the controller appends its `controller.started` event
immediately after that zero-filled gap, so the NUL run and a complete, valid
JSON object end up sharing a single physical line.

No event payload is lost. This is purely a **framing** bug. But it is a framing
bug with a silent-truncation failure mode: strict JSON Lines consumers abort at
the first such line and return only the history *before* it, with no error and
no signal that they stopped early. The result reads exactly like "that is all
the events there are."

The requested fix is small: on startup, before appending, detect a NUL-padded
tail and truncate back to the last newline. That makes the event log
self-healing across crashes.

## Honest scope note — no live reproduction

**Both cities on this host currently parse clean.** Verified directly on
2026-08-16:

| City | Lines | Strict `jq -e .` | Lines containing NUL |
|---|---:|---|---:|
| paper-town | 137,709 | OK | 0 |
| community-council | 41,248 | OK | 0 |

This report is written from a historical incident and from the mechanism, not
from a reproduction available right now. The two torn lines that prompted it
were repaired in place on 2026-08-11 (see *Local state* below), and no unclean
shutdown has occurred since. Treat the line/sequence numbers below as an
incident record, not as something you can reproduce by cloning this state.

## Observed occurrences

Both in paper-town, one per unclean shutdown, each roughly 90 seconds after
host boot — consistent with the controller starting up and writing its first
event onto the unwritten tail block:

| Date | Sequence at the torn line |
|---|---|
| 2026-08-08 | 5638 |
| 2026-08-09 | 13642 |

The shape on disk is the same both times: N leading NUL bytes, then a complete
and valid `controller.started` object, with no newline between them.

## Why this matters more than the byte count suggests

The dangerous property is not corruption — it is that the corruption is
*invisible to the reader*.

```
jq -c . .gc/events.jsonl        # stops at the torn line, exits non-zero,
                                # but has already emitted everything before it
```

A consumer that pipes `jq -c .` and reads stdout gets a clean, plausible,
chronologically-ordered prefix of the city's history. Nothing in that output
indicates it is a prefix. Any question answered from it — when did this agent
last run, has this order ever fired, what happened after the incident — gets a
confident wrong answer, and the wrongness scales with how much history sits
after the torn line. In the 2026-08-08 case the truncation hid roughly 63% of
the log from strict readers.

Because it is one corrupt line per unclean shutdown, it recurs by construction.
It does not degrade gracefully; each crash adds another silent stopping point,
and the earliest one is the one that wins.

## Requested fix (primary)

On startup, before appending to `events.jsonl`, check whether the file ends in
a NUL-padded tail and truncate it back to the last newline.

Sketch of the condition worth handling: read the final block; if the bytes
after the last `\n` are all `\x00`, `ftruncate` to just past that last `\n`.
That is cheap (one seek + one read of the tail, no full scan), it is safe when
the file is well-formed (the tail after the last newline is empty, so nothing
is truncated), and it cannot discard a real payload — a NUL run is never valid
JSON Lines content.

Worth deciding explicitly: whether to log the truncation as an event or a
warning. We would prefer it be visible, since a silently self-healing log
makes the underlying unclean-shutdown rate invisible too.

## Requested fix (secondary) — a parseability assertion in the built-in doctor

The built-in checks do not notice that the log is unreadable. Live output from
`gc doctor --json` on this host today:

```
events-log       [ok]  events.jsonl exists and writable
events-log-size  [ok]  events.jsonl size: 34.9 MB
```

`events-log` verifies existence and writability; `events-log-size` reports
size. A file can be 100% unparseable by strict consumers and both still report
`ok`. A parseability assertion belongs in the built-in check so that every
city gets it, not just cities that happen to have written their own.

For contrast, the local drop-in on the same run:

```
paper-town:check-events-log-framing  [ok]  events.jsonl parses end to end (137760 lines, no torn frames)
```

## Reference implementation

paper-town carries a local drop-in at `doctor/check-events-log-framing/` that
does the secondary fix. It is a reasonable starting point for an upstream
version; the design decisions in it that we would keep:

- **Cheap signature check first.** `grep -c -a -P '\x00'` costs ~40ms on 11MB
  and finds the torn-append signature directly.
- **Authoritative check second.** `jq -e .` is the ground truth for "can a
  strict consumer read this," and costs ~1.7s on 11MB.
- **Size guard with an escape hatch.** Above a threshold
  (`GC_DOCTOR_EVENTS_MAX_PARSE_BYTES`, default 256MB) the strict parse is
  skipped so it cannot eat the per-check time budget — but the check *says*
  it skipped rather than reporting a silent `ok`. Skipping quietly would
  reintroduce the exact failure mode the check exists to catch.
- **Actionable failure output.** It prints the offending line numbers, the
  parser's stopping point, and both non-destructive read workarounds.

Happy to send it as a PR against the built-in check set if that is the shape
you want.

## Workarounds for anyone hitting this today

Read the whole log without repairing it — neither of these loses an event:

```bash
tr -d '\0' < .gc/events.jsonl | jq -c .   # strips the padding, keeps every payload
jq -R 'fromjson?' .gc/events.jsonl        # skips torn lines
```

Repair in place, without racing the controller's live appends: back the file up,
then overwrite each NUL run with spaces **at its byte offset** using
`dd conv=notrunc`. Same length, same offsets, no truncation — so a concurrent
append cannot land in a moved position. Do not rewrite the file wholesale while
the controller is running.

## Local state

paper-town is healthy. Both torn lines were repaired in place on 2026-08-11
under bead `pt-0y7` using the `dd conv=notrunc` method above. Strict `jq` now
reads the file end to end, and sequence numbers were contiguous with zero gaps
across the repaired range (1..35309 at the time of verification). The local
doctor drop-in will catch the next occurrence.

## Environment

- `gc` 1.4.1, `bd` 1.2.2
- Linux 6.17.9, ext4
- Two cities on the host (paper-town, community-council); observed in
  paper-town only, but nothing about the mechanism is city-specific

rjgeng:
Fixes #5336.

An unclean shutdown can leave a NUL-padded tail on `events.jsonl`. Nothing ever truncated it, so every subsequent event got appended after that corrupted line — burying all of it for any strict external reader (`jq -e` and similar fail on the NUL bytes).

**Fix:** a new recorder-startup truncation step detects and removes the NUL-padded tail before the recorder resumes appending.

**Testing:** `go build`/`gofmt -l` clean, full end-to-end repro test included (writes a NUL-padded tail, restarts the recorder, confirms truncation and that subsequent events are readable).

---


sjarmak (dismissed):
The diagnosis is right and the detection logic is carefully conservative: requiring the last byte to be NUL, requiring a newline in the window, and refusing to guess when the bytes after that newline are not all NUL are each the correct call. The bounded window is the right shape too. Two problems with where it runs, one of which can destroy events.

## Blocking: the truncation takes no flock, and this file is written by multiple processes

`FileRecorder`'s own type doc states the concurrency contract: "It uses O_APPEND for cross-process safety, a mutex for in-process serialization, and a bounded-wait advisory file lock (flock) for cross-process serialization." Every write path honors that. `truncateNulPaddedTail` does not. It opens `O_RDWR` and calls `f.Truncate` with no lock held at all.

Multiple processes do hold recorders on one active log, so this is reachable on a normal startup with no crash involved:

1. Process A constructs a `FileRecorder`, `Stat`s the log at size S, reads the tail, and computes `truncateAt`.
2. Process B, holding the flock, appends a valid record. The file is now S+k.
3. Process A calls `f.Truncate(truncateAt)`.

B's record is gone, along with anything else appended in that window. B's in-memory `r.seq` is now ahead of what the file contains, so the next write leaves a gap. Nothing detects this or reports it.

The pre-existing NUL run is not even required for the damaging case. Step 3 is destructive whenever any concurrent append lands between the `Stat` and the `Truncate`.

There is a second shape from the same missing lock. `rotateLocked` renames the active log (`recorder.go:486`). An fd opened on `path` before that rename keeps pointing at the same inode, which is now the `rotating-*` file, so a rotation racing this check truncates the rotating file instead of the active one.

The fix is to acquire the same flock this package already uses for `Record` around the read-decide-truncate sequence, so the check is serialized against appends and rotation the way every other writer is.

## Blocking: a short read is treated as a NUL tail

```go
tail := make([]byte, n)
if _, err := f.ReadAt(tail, size-n); err != nil && err != io.EOF {
```

The byte count is discarded, and `io.EOF` is accepted as success. `ReadAt` returns `io.EOF` precisely when it read fewer than `len(tail)` bytes, and `tail` is zero-initialized, so a short read leaves the unfilled remainder as `\x00`. The very next check is `tail[len(tail)-1] != 0`, which a short read passes by construction.

So a file that shrank between the `Stat` and the `ReadAt` (rotation, or another process's truncation) presents as a NUL-padded tail and gets truncated, discarding real records. This is the same race as above arriving through a different door, and it is worth fixing on its own even after the lock goes in: treat `n < len(tail)` as "the file changed underneath me, do nothing" rather than as a successful read.

## Non-blocking

The stderr line reports recovery well. Given that this only ever fires after an unclean shutdown, an event on the bus would let the fleet see how often this is actually happening rather than leaving it in a log nobody greps.

## Related

I filed #5650 tonight against `activeScanStart` in the same package, and a torn or short write is one of only two reachability paths I could find for it. The NUL-padded tail this PR repairs is a concrete instance of that class, so the two are worth reading together. Separately, `writeRecordLocked` (`recorder.go:394-398`) does not check for a short write, while `writeBatch` (`:364-370`) does; that asymmetry looks unintentional and is plausibly one source of the corruption being repaired here.

Happy to push the flock and short-read fixes to this branch if you would rather not carry them, just say the word rather than me touching your branch unasked.

rjgeng:
Both blocking findings are fixed and pushed — apologies for the gap between the push and this note; that was on me, not a disagreement.

The fix is `342d20f46` (current head `023830ed4`). Worth flagging that the branch was rebased onto #5718's merge on the 28th, so any SHA from before then will not resolve.

**Flock.** `truncateNulPaddedTail` now acquires the same `lockRecorderFile` helper `Record`/`AppendBatch` use, immediately after the open, and holds it across the whole Stat → ReadAt → decide → Truncate sequence. No second locking mechanism and no partial coverage — a concurrent append can no longer land between the Stat and the Truncate.

**Short read.** Extracted a `readNulTailWindow` helper that returns the actual byte count instead of discarding it, and reports `ok=false` when `read < n`; `io.EOF` is no longer accepted as success. A file that shrank underneath the check now reads as "the file changed, do nothing" rather than as a synthetic NUL tail. You were right that this is worth fixing independently of the lock — the guard is what makes the rotation case below fail closed.

Three tests cover it: one holds a sibling flock and asserts the check blocks and then re-evaluates post-append state rather than a stale snapshot; one asserts it fails closed on a flock-acquire timeout; one exercises the short-read guard directly, so it is deterministic rather than depending on winning a live race.

**On your second shape, one thing I want to flag rather than quietly claim.** The flock closes the destructive door, but it does not fully close the rename one. `rotateLocked` calls `r.file.Close()` — which releases the flock — before `os.Rename`, so a check that opened `path` just before that rename still ends up holding its lock on the now-`rotating-*` inode. My reading is that the residue is benign rather than destructive: nothing appends to a rotating file, so there is nothing there to lose, and the worst case is that this startup misses the active log's NUL tail and instead trims a genuine one off the file that is about to be archived — a no-op or a correct repair, not a loss. The short-read guard covers the shrink path. But that is my reading of your finding, not a fix for it, and I would rather say so plainly than let it look handled. If you read the residue as worse than that, I will take it as a follow-up.

**On the offer to push the fixes yourself** — thank you, genuinely. Not needed; the work was already carried before I saw your review, I just failed to say so here.

**On the non-blocking telemetry suggestion** — agreed, and deliberately left out of this PR. An event on the bus for how often the NUL-tail recovery actually fires is the right call, but it widens this diff past the repair, so I would rather it be its own change.

quad341 (dismissed):
@rjgeng — this is a clean piece of work, and I want to be specific about why.

The detection logic is properly paranoid in all the right places: requiring the last byte to be NUL, requiring a newline inside the window, and refusing to guess when the bytes after that newline aren't all NUL. Bounding the scan window keeps the startup check cheap no matter how large the log grows.

The follow-up round is the part I'd single out. You fixed both findings with the package's own `lockRecorderFile` rather than inventing a second locking scheme, and the `readNulTailWindow` short-read guard is a better fix than the one that was asked for. Building deterministic tests for both — sibling flock, timeout, direct short-read — instead of hoping to win a live race was the right call. And thank you for flagging the residual rename case plainly rather than letting it look handled; your reading checks out.

Where things stand: @sjarmak's changes-requested is still open on GitHub and predates your fix, so I'm not merging past it — I've asked for a re-review. Meanwhile I'll rebase onto `main` on my side and re-run CI, since the branch has drifted a couple of weeks. Nothing needed from you. Telemetry gets its own follow-up.


---


quad341:
@rjgeng, following up on our 09-11 review: I've merged current `main` into your branch (96f2ba97a1), as promised, and CI is running fresh. That review's closing line said it was ready to merge; that line was wrong, and I'm sorry for the confusion. It is still waiting on @sjarmak's re-review of his earlier change request, which your follow-up round addressed. Nothing needed from you.

quad341 (approved):
Re-reviewed `023830ed4`..`96f2ba97a` (269 new commits since the last review): no new concerns — the prior review stands.

PR #5615, opened 2026-08-25
fix(session): surface two setup-timeout config footguns as validation warnings (#5279)

### LINKED ISSUE #5279: Session setup timeouts silently defeat each other (setup_timeout / setup_max_timeout / startup_timeout): no validation
## Summary

The session-start timeout knobs under `[session]` silently defeat one another. An
operator who raises `setup_timeout` to accommodate a slow `pre_start` command can hit
three separate footguns in sequence and reasonably conclude the knob is broken — because
each of the three makes a config change *look* applied while doing nothing (or something
different from what was asked). There is no validation, warning, or doctor check for any
of the three, even though the codebase already reasons about the ordering of these values
in at least one place.

The three knobs involved:

- `[session] setup_timeout` (default `10s`) — per-command budget for `pre_start` /
  session-setup commands.
- `[session] setup_max_timeout` (default unset) — opt-in "activity-aware" setup budget.
- `[session] startup_timeout` (default `60s`) — wall-clock bound on each agent's `Start()`
  call, which is the call that runs `pre_start`.

All line references below are against `origin/main` at commit
`f6741d94861aa14f0253deffbe9efb1cb3a35d92`.

## Footgun 1 — `setup_max_timeout` silently converts `setup_timeout` from a TOTAL bound into a SILENCE (idle) bound

When `setup_max_timeout` is set, `setup_timeout` stops meaning "total wall-clock budget for
the command" and starts meaning "max time the command may run **without producing output**"
(an idle/silence bound); `setup_max_timeout` becomes the total ceiling.

`internal/runtime/tmux/adapter.go`, `runSetupCommand` (around lines 954–972):

```go
idle, grace := time.Duration(0), setupCommandWaitDelay
if o.setupMaxTimeout > 0 {
    idle, grace = timeout, setupCancelGrace   // <- setup_timeout becomes the IDLE bound
}
mon := execgrace.NewMonitor(ctx, idle, o.setupMaxTimeout)
```

This mode switch is documented on the field (`internal/config/config.go`, the
`SetupMaxTimeout` doc comment around lines 1558–1567), but it is a silent semantic
reinterpretation of a *different* field, with no warning at config-load time. The failure
mode it creates is sharp: a `pre_start` command that is legitimately silent until it
finishes — e.g. a long-running asset/skills materialization step that streams nothing and
then prints a single summary line at the very end — is making progress the whole time, but
because it produces no output it trips the idle bound (`setup_timeout`) and is killed, even
though its total runtime is well under `setup_max_timeout`. The operator set
`setup_max_timeout` specifically to protect a slow command and instead armed a silence
tripwire against it.

## Footgun 2 — `gc supervisor reload` silently NO-OPs a `setup_timeout` change

The session provider is constructed once, resolving `setup_timeout` / `setup_max_timeout`
into its config at build time:

- `cmd/gc/providers.go`, `tmuxConfigFromSession` (lines 94–95):
  ```go
  SetupTimeout:    sc.SetupTimeoutDuration(),
  SetupMaxTimeout: sc.SetupMaxTimeoutDuration(),
  ```

On reload, the provider is only rebuilt when the provider **name** (or a pack-declared
runtime binding) changes — not when a timeout value changes:

- `cmd/gc/city_runtime.go`, `reloadConfigTraced` (around line 2006):
  ```go
  if newProviderName != *lastProviderName || packRuntimeDeclarationChanged(cr.cfg, nextCfg, newProviderName) {
      newSp, spErr := newSessionProviderForCityByName(nextCfg, newProviderName, ...)
      // ... nextSp = newSp
  }
  ```

A `setup_timeout`-only edit to `city.toml` does not change the provider name, so `cr.sp`
keeps the **old** provider carrying the **old** cached timeout. `reload` reports success and
swaps in the new config revision, but the value that actually bounds `pre_start` is stale
until a full `gc restart` rebuilds the provider. The change looks applied and is not.

## Footgun 3 — `startup_timeout` silently CAPS `setup_timeout`, with no validation for the inversion

`pre_start` runs *inside* the provider `Start()` call, and that whole call is wrapped in a
context deadline of `startup_timeout`:

- `cmd/gc/session_lifecycle_parallel.go` (around lines 1411–1418):
  ```go
  if startupTimeout > 0 {
      startCtx, cancel = context.WithTimeout(ctx, startupTimeout)
  }
  // ...
  startedFresh, err := startPreparedStartCandidate(startCtx, item, ...)  // runs pre_start under startCtx
  ```
  where `startupTimeout = cfg.Session.StartupTimeoutDuration()` (default `60s`,
  `internal/config/config.go` lines 1695–1698).

Because the inner `setup_timeout` deadline lives under this outer `startup_timeout`
deadline, raising `setup_timeout` above `startup_timeout` without also raising
`startup_timeout` has no effect: the outer deadline fires first and cancels the whole start.
Raising `setup_timeout` from `30s` to `120s` while `startup_timeout` stays at its `60s`
default is a silent no-op that looks applied.

Notably, the codebase already compares these two values elsewhere —
`cmd/gc/cmd_supervisor_city.go` (around lines 99–101) widens the start-ready timeout when
`startup_timeout` exceeds it:

```go
if startup := cfg.Session.StartupTimeoutDuration(); startup > timeout {
    timeout = startup
}
```

— so the ordering relationship is already understood in one place, but nothing validates the
`setup_timeout` vs `startup_timeout` inversion, and the doctor duration check
(`internal/doctor/checks_semantic.go`, `collectRanges`, lines 101 and 105) validates each
of `setup_timeout` and `startup_timeout` only against an independent min/max range, never
against each other.

## Reproduction

An operator has a `pre_start` command that legitimately takes ~90s (large checkout, asset
materialization, etc.) and is often silent for long stretches. They try to give it room:

1. Raise `setup_timeout` from `30s` to `120s` in `city.toml`. No effect on the command's
   effective budget — Footgun 3: `startup_timeout` is still `60s` and caps it.
2. Run `gc supervisor reload` to apply the change. Still no effect — Footgun 2: the
   provider cached the old timeout at load; reload does not re-read it.
3. Also set `setup_max_timeout = "10m"` to get the "activity-aware" budget. Now the command
   is killed after `setup_timeout` of **silence** rather than after `setup_timeout` total —
   Footgun 1 — so a command that streams nothing until its final line dies mid-run despite
   making progress.

At each step the config change looks accepted (`reload` succeeds, `config show` reflects the
new values) while the effective behavior is unchanged or changed in an unexpected direction.

## Impact

The three knobs form a single "session setup timeouts silently defeat each other" trap. An
operator debugging a slow `pre_start` will naturally reach for `setup_timeout` first, then
`reload`, then `setup_max_timeout` — and can hit all three in that order, each masking the
last, before concluding the timeout controls are broken. Nothing in config validation,
`reload` output, or `gc doctor` warns about any of the three interactions.

## Related

- Issue #1853 (open) asks that `setup_timeout` be settable per rig-scoped agent from
  `city.toml`; that is about *where* the value can be set, which is distinct from the three
  interactions here.
- Issue #4569 (closed) concerns `runSetupCommand` delivering an untrappable `SIGKILL` on
  deadline expiry (rollback-trap / data-loss safety) — same code region, different defect.
  The `setup_max_timeout` "activity-aware" mode in Footgun 1 appears to be the response to
  the "slow but healthy work shouldn't be killed" half of that report; this issue is about
  its silent interaction with the other two knobs and the absence of validation.

## Fix candidates

**(a) Add cross-field validation warnings + a docs note (cheapest, highest signal).**
In `internal/config/validate_durations.go` (or the doctor `DurationRangeCheck` in
`internal/doctor/checks_semantic.go`, which already collects both fields):
- Warn when `setup_timeout >= startup_timeout` — the value is silently capped (Footgun 3).
- Warn when `setup_max_timeout` is set, noting that `setup_timeout` then means an
  idle/silence bound rather than a total bound (Footgun 1).
- Document (and ideally state in `reload`'s output) that `setup_timeout` /
  `setup_max_timeout` changes require a full `gc restart`, not a `reload`, until (b) lands
  (Footgun 2).
This makes all three footguns visible at the moment the operator edits the config, without
changing any runtime behavior.

**(b) Make `reload` actually re-read the setup timeouts (removes Footgun 2 outright).**
In `cmd/gc/city_runtime.go` `reloadConfigTraced`, extend the provider-rebuild condition
(currently keyed only on provider-name / pack-runtime-binding change, ~line 2006) so that a
change to `setup_timeout` / `setup_max_timeout` also invalidates the cached provider (or
push the resolved durations into the live provider without a full rebuild). Then a
`reload` after a timeout edit takes effect as operators expect.

**(c) Unify the three knobs into one coherent budget model (largest change).**
Replace the overloaded trio with a single explicit setup-budget model — e.g. a total-setup
budget plus an optional idle budget — and have `startup_timeout` automatically encompass the
setup budget so it can never silently cap it. This removes the inversion and the
mode-switch surprise structurally rather than warning about them.

**Recommendation:** land (a) immediately — it is low-risk, touches only validation/doc
surfaces, and converts all three silent footguns into visible warnings. Follow with (b) to
eliminate the most surprising one (the `reload` no-op). Treat (c) as the longer-term
cleanup if the knobs are revisited.

[autogenerated by Claude Opus 4.8 on 2026-08-14]

rjgeng:
Fixes #5279.

Two silent config footguns around session setup timeouts:

1. `setup_timeout` silently means something different once `setup_max_timeout` is also set — easy to misread as still-authoritative when it isn't.
2. `setup_timeout >= startup_timeout` is a silent no-op — the config validates but the intended timeout never actually takes effect.

**Fix:** both now surface as validation warnings instead of failing silently.

**Scope note:** deliberately scoped down from 3 footguns to these 2. The third (a reload provider-rebuild gap) is a separate, deeper follow-up — not touched here.

---

### DIFF
--- a/cmd/gc/cmd_agent.go
+++ b/cmd/gc/cmd_agent.go
@@ -186,6 +186,9 @@ func isNonFatalLoadConfigWarning(warning string) bool {
 	if config.IsDisabledNamedSessionWarning(warning) {
 		return true
 	}
+	if config.IsSessionSetupTimeoutAdvisory(warning) {
+		return true
+	}
 	if config.IsAlwaysFreshWakeModeWarning(warning) {
 		return true
 	}
--- a/cmd/gc/setup_timeout_advisory_test.go
+++ b/cmd/gc/setup_timeout_advisory_test.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/gastownhall/gascity/internal/config"
@@ -50,5 +51,33 @@ func TestSessionSetupTimeoutAdvisoriesAreNonFatalAndEmitted(t *testing.T) {
 				t.Error("a setup-timeout advisory must reach the operator, not be swallowed")
 			}
 		})
+	}
+}
+
+// TestUnparseableDurationIsNotMistakenForAnAdvisory pins the classifier
+// against the one input that can forge an advisory: the unparseable-duration
+// warning quotes the operator's raw value verbatim, so a value containing an
+// advisory sentence would be recognized as advisory and lose its strict-fatal
+// handling if the match were not anchored.
+func TestUnparseableDurationIsNotMistakenForAnAdvisory(t *testing.T) {
+	forged := "setup_timeout now bounds idle/silence time between output, not total setup runtime"
+	warnings := config.ValidateDurations(&config.City{
+		Session: config.SessionConfig{SetupTimeout: forged, StartupTimeout: "60s"},
+	}, "city.toml")
+
+	var parseWarning string
+	for _, w := range warnings {
+		if strings.Contains(w, "is not a valid duration") {
+			parseWarning = w
+		}
+	}
+	if parseWarning == "" {
+		t.Fatalf("warnings = %v, want an unparseable-duration warning", warnings)
+	}
+	if config.IsSessionSetupTimeoutAdvisory(parseWarning) {
+		t.Fatalf("an unparseable duration was classified as an advisory: %q", parseWarning)
+	}
+	if fatal, _ := splitStrictConfigWarnings([]string{parseWarning}); len(fatal) != 1 {
+		t.Errorf("strict split: fatal=%v, want the unparseable duration to stay fatal", fatal)
 	}
 }
--- a/cmd/gc/setup_timeout_advisory_test.go
+++ b/cmd/gc/setup_timeout_advisory_test.go
@@ -1,0 +1,54 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/gastownhall/gascity/internal/config"
+)
+
+// TestSessionSetupTimeoutAdvisoriesAreNonFatalAndEmitted pins both downstream
+// re-classifiers of the setup-timeout advisories introduced for
+// gastownhall/gascity#5279. Strict mode is on by default for `gc start`, so a
+// warning that is not recognized as advisory turns a city that starts today
+// into one that exits 1 — and the setup_max_timeout advisory fires on every
+// config that sets the field, including a correct one. The agent emit path
+// must also surface them, or the advisory the PR exists to deliver is dropped
+// before it reaches the operator.
+//
+// The warning text is derived from config.ValidateDurations rather than
+// hardcoded so this test cannot pass against a string the validator no longer
+// emits.
+func TestSessionSetupTimeoutAdvisoriesAreNonFatalAndEmitted(t *testing.T) {
+	for _, tc := range []struct {
+		name    string
+		session config.SessionConfig
+	}{
+		{
+			name:    "setup_timeout at or above startup_timeout",
+			session: config.SessionConfig{SetupTimeout: "90s", StartupTimeout: "60s"},
+		},
+		{
+			name:    "setup_max_timeout reinterprets setup_timeout",
+			session: config.SessionConfig{SetupMaxTimeout: "5m"},
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			warnings := config.ValidateDurations(&config.City{Session: tc.session}, "city.toml")
+			if len(warnings) != 1 {
+				t.Fatalf("warnings = %v, want exactly the setup-timeout advisory", warnings)
+			}
+			w := warnings[0]
+			if !config.IsSessionSetupTimeoutAdvisory(w) {
+				t.Fatalf("advisory not recognized by its own classifier: %q", w)
+			}
+
+			fatal, nonFatal := splitStrictConfigWarnings([]string{w})
+			if len(fatal) != 0 || len(nonFatal) != 1 {
+				t.Errorf("strict split: fatal=%v nonFatal=%v, want the advisory non-fatal", fatal, nonFatal)
+			}
+			if !shouldEmitLoadCityConfigWarning(w) {
+				t.Error("a setup-timeout advisory must reach the operator, not be swallowed")
+			}
+		})
+	}
+}
--- a/cmd/gc/strict_warnings.go
+++ b/cmd/gc/strict_warnings.go
@@ -21,5 +21,6 @@ func strictWarningIsNonFatal(warning string) bool {
 		config.IsLegacyWorkspaceFieldWarning(warning) ||
 		config.IsIdleSleepMaskedByIdleTimeoutWarning(warning) ||
 		config.IsAlwaysFreshWakeModeWarning(warning) ||
-		config.IsRetiredKeyWarning(warning)
+		config.IsRetiredKeyWarning(warning) ||
+		config.IsSessionSetupTimeoutAdvisory(warning)
 }
--- a/internal/config/validate_durations.go
+++ b/internal/config/validate_durations.go
@@ -2,8 +2,26 @@ package config
 
 import (
 	"fmt"
+	"strings"
 	"time"
 )
+
+const (
+	setupTimeoutMaskedByStartupWarningFragment   = "setup_timeout can never fire because startup_timeout bounds the whole session Start() call first"
+	setupTimeoutBecomesIdleBudgetWarningFragment = "setup_timeout now bounds idle/silence time between output, not total setup runtime"
+)
+
+// IsSessionSetupTimeoutAdvisory reports whether warning is one of the
+// [session] setup-timeout advisories. They describe how the timeout knobs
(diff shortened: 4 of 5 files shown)


sjarmak (approved):
Both warnings state something true about the runtime, and I verified each rather than taking the description's word for it. One of them, as written, would have stopped cities from starting, so I fixed that here rather than sending it back.

**The claims hold.** `startup_timeout` is established as the parent context outside the provider `Start()` call (`cmd/gc/session_lifecycle_parallel.go:1425-1433`); that same context flows through `Provider.Start` into `doStartSession`, which runs `pre_start` and `session_setup` under it (`internal/runtime/tmux/adapter.go:1204-1215`, `:1374-1379`). `setup_timeout` is derived from that parent later inside `runSetupCommand` (`:1006-1010`), so `setup_timeout >= startup_timeout` really does mean the setup deadline can never be the first to fire. The second claim is equally accurate: when `setup_max_timeout > 0`, `idle` is set to `timeout` and the writers are wrapped in `mon.Writer(...)` (`:999-1009`, `:1024-1027`), so output resets the budget and `setup_max_timeout` becomes the ceiling.

Defaults do not false-positive on the first check: 10s setup against 60s startup (`internal/config/config.go:1652`, `:1697`), so an all-default city stays quiet.

**Two fixes pushed to this branch.**

The blocking one: neither warning was recognized by `cmd/gc`'s re-classifiers. `strictWarningIsNonFatal` (`cmd/gc/strict_warnings.go:18-24`) lists six predicates and matched neither, and strict mode is on by default for `gc start` (`cmd/gc/cmd_start.go:676-677`, `:735-743`). The `setup_max_timeout` warning fires on **every** config that sets the field, including a deliberately correct one, so setting a documented knob would have turned a city that starts today into one that exits 1. The reload paths do the same. The emit path had the mirror-image problem: `shouldEmitLoadCityConfigWarning` (`cmd/gc/cmd_agent.go:213-221`) allowlists what reaches the operator, so both advisories were dropped there, which is the surfacing the change exists to deliver. Both warnings now carry a shared text fragment and an `IsSessionSetupTimeoutAdvisory` classifier, matching the shape of the existing always-fresh and idle-sleep advisories. The emitted strings are byte-identical, so nothing downstream that matched on them changes.

The second: the classifier's first form matched with `strings.Contains`, and the unparseable-duration warnings in this same function quote the operator's raw value verbatim. A value containing an advisory sentence would be classified as an advisory and lose its strict-fatal handling. Anchoring the match to the end of the string closes it, because a parse warning always ends with the parse error. Proven by execution, not by argument: with `Contains`, the new test fails with `an unparseable duration was classified as an advisory: "city.toml: [session] setup_timeout = \"setup_timeout now bounds idle/silence time between output, not total setup runtime\" is not a valid duration: ..."`; with `HasSuffix` it passes.

Tests for both are in the same commits as the fixes. `go test ./internal/config/...` and the new `cmd/gc` cases pass; `go vet` and `golangci-lint` over the touched packages are clean.

**Non-gating, not fixed here.** The both-conditions-fire case has no test, and this file's convention of asserting `len(warnings) == 1` is brittle for an accumulator. There is existing precedent in the same file for asserting on the joined set instead (`internal/config/validate_durations_test.go:402-415`), which would be the better shape if this area grows. Not worth holding the merge.

Codex (gpt-5.4) reviewed three times: before the fixes, after the first fix, and against the exact head being merged. The first pass independently found the strict-fatal promotion and traced it to the same three lines; it also established that there is precedent in this package for advisory warnings conditioned on a risky shape but none for one that fires merely because a valid feature is enabled, and that `internal/config/site_binding_durability.go:38-40` says in as many words that repeated broad warnings train users to ignore the channel. The second pass found the `Contains` over-breadth, confirmed both classifier sites are correctly wired with nothing changed by accident, confirmed the strings are byte-identical to the parent commit, and confirmed the added test fails if the wiring is removed from either site. The third pass, against `ceabffbfb`, came back with no substantiated issues, and closed the remaining question in both directions: no other warning builder in the file can end with either fragment, because every one of them appends text after the operator's value (`: %v` on the parse and sleep warnings, fixed trailing text on the positive-duration ones), and neither advisory can fail to end with its fragment, because the fragment is the final format argument in both constructors. It also confirmed `splitStrictConfigWarnings` and `shouldEmitLoadCityConfigWarning` are the only two production reclassifiers, reached from `cmd_start.go`, `controller.go`, `cmd_agent.go`, and `cmd_supervisor.go`, so there is no third site left unwired.

Approving with the fixes in.

PR #5616, opened 2026-08-25
fix(config): keep resume's option-defaults gate scoped to explicit overrides (#5185)

### LINKED ISSUE #5185: bug: resume drops provider option_defaults (--dangerously-skip-permissions, --effort max); regression of #799
### Summary

Provider `option_defaults` get applied when a session is first spawned, then silently dropped when that same session is resumed. For the builtin `claude` provider that means a resumed session loses both `--dangerously-skip-permissions` and `--effort max`. The permission flag is the one that hurts. Once it's gone, the resumed agent parks forever on the first per-command approval prompt, and (this is the awkward part) every liveness signal gc reports still insists the session is perfectly healthy.

It's the same symptom #799 described, and that one was closed as completed back on 2026-04-19. I've reproduced it on current builds. So either the fix regressed, or it covered a different path than the one these sessions actually take.

### Gas City version

Observed on `1.3.5` (Linux, cloud host), analysed against `main` at [`19e0862`](https://github.com/gastownhall/gascity/commit/19e0862cf8bae6cc631d2828e77bd63878bbe7fd). The relevant code in `internal/config/launch_command.go` hasn't changed between `v1.3.5`, `v1.4.0` and current `main`, so I don't think version skew explains it.

### Actual behavior

Two `claude` processes, one Linux host, one gc, one provider config, one settings file. Spawn versus resume is the only variable here, which is why they're worth reading side by side.

```
pid 202781  claude --resume <id> --settings /home/ubuntu/city/.gc/settings.json
            ^ manager, resumed          -> NEITHER flag

pid 202783  claude --dangerously-skip-permissions --effort max --settings /home/ubuntu/city/.gc/settings.json
            ^ mayor, first spawn        -> BOTH flags
```

`--effort max` getting dropped alongside the permission flag is the tell I'd point a maintainer at first, because it rules out any explanation that's specific to permission handling. It isn't one key being fumbled. The resume path looks like it's ignoring `option_defaults` wholesale.

### Expected behavior

A resumed session should carry the same resolved provider `option_defaults` as a freshly spawned one. `gc config explain --provider claude --json` already reports `option_defaults = {"effort": "max", "permission_mode": "unrestricted"}` with provenance `builtin:claude`, set by no user config layer. So the configuration is correct, and as far as I can tell it always was. That's why this reads to me as a code defect rather than a missing knob.

### Ruling out the two obvious explanations

**Not a shell alias.** The host has no `claude` alias anywhere (`~/.bashrc`, `~/.profile`, `~/.bash_profile`, `/etc/profile.d/*.sh`). Both processes above resolve the same binary.

**Not a stale fork.** I diagnosed against `gastownhall/gascity`, not a downstream fork, and confirmed the installed `v1.4.0` build ([`f188d876f`](https://github.com/gastownhall/gascity/commit/f188d876f)) is an ancestor of upstream `main`.

### Where the asymmetry appears in current main

Both builders live in `internal/config/launch_command.go`, about eighteen lines apart, and they don't gate on the same predicate.

```go
// line 40 — launch
if len(resolved.OptionsSchema) > 0 && hasProviderOptionValues(resolved, optionOverrides) {

// line 58 — resume
if command == "" || len(resolved.OptionsSchema) == 0 || !hasSchemaOptionOverrides(optionOverrides) {
    return command, nil   // early return, bare command
}
```

The predicates are not equivalent:

```go
func hasProviderOptionValues(resolved *ResolvedProvider, optionOverrides map[string]string) bool {
    if resolved != nil && len(resolved.EffectiveDefaults) > 0 {
        return true                              // provider defaults count
    }
    return hasSchemaOptionOverrides(optionOverrides)
}

func hasSchemaOptionOverrides(optionOverrides map[string]string) bool {
    for key := range optionOverrides {
        if key != "initial_message" {
            return true                          // explicit overrides only
        }
    }
    return false
}
```

`EffectiveDefaults` is what `ComputeEffectiveDefaults` produces by merging schema defaults, provider `option_defaults` and agent defaults. The launch path counts it. The resume path doesn't, so a provider whose options come entirely from defaults takes the early return and emits a bare command.

`replaceResumeSchemaFlags` compounds it. It strips every schema flag first, then re-adds nothing when `overrideArgs` is empty:

```go
stripped := StripFlags(command, CollectAllSchemaFlags(schema))
if len(overrideArgs) == 0 {
    return unquoteSessionKeyTemplate(stripped)
}
```

```mermaid
flowchart TD
    D["provider option_defaults<br/>effort=max, permission_mode=unrestricted"] --> E[EffectiveDefaults]
    E --> L{"launch:<br/>hasProviderOptionValues"}
    E -.->|not consulted| R{"resume:<br/>hasSchemaOptionOverrides"}
    L -->|"defaults count -> true"| LA["args injected<br/>--dangerously-skip-permissions --effort max"]
    R -->|"no explicit overrides -> false"| RA["early return<br/>bare command"]
    RA --> P["session parks on<br/>first approval prompt"]
```

### Relationship to #799

#799 reported resumed pool workers losing `--dangerously-skip-permissions` and wedging on approval prompts. The closing comment states the worker boundary now rebuilds resume commands from current resolved session and provider config rather than stale stored metadata, and that regression coverage asserts default args survive a resume.

Whatever that coverage asserts, it isn't catching the case above. Worth checking whether those tests supply explicit option overrides, since a test written that way would pass `hasSchemaOptionOverrides` and never exercise the defaults-only path that fails here.

### A second observation, probably unrelated to the cause

`internal/config/provider.go` documents `PermissionModes` as a config-only lookup table for external clients, with the comment "Launch-time flag substitution is planned for a follow-up PR — currently no runtime code reads this field." That's consistent with the flag reaching argv through `OptionsSchema` and `OptionDefaults` instead. I mention it only because it's easy to mistake for the mechanism while reading this code.

One caveat for anyone reproducing via the CLI: `OptionsSchema` is tagged `json:"-"`, so its absence from `gc config explain --json` output isn't evidence that a provider declares no schema. That misled me for a while.

### What I could not determine

I could not pin which of two branches is the one actually taken, because I have no host that both reproduces and is safe to instrument:

1. `internal/worker/handle_lifecycle.go` `startCommand`, the one resume path with no default-args re-injection, which returns `BuildResumeCommand(resumeInfo)` verbatim.
2. `config.BuildProviderResumeCommand`, as analysed above.

Both are consistent with the observed argv. I'm reporting the asymmetry rather than proposing a patch for that reason.

### Impact

An unattended session parks indefinitely after any resume, and gc's own signals hide it. `gc session list` reports the session `active` with `LAST ACTIVE 0s ago`, the listener stays alive, `nudge_health` reads `ok`, and the inbound message sits delivered in the inbox. `gc session peek` is the only thing that reveals the process sitting on `This command requires approval`. One of ours sat there roughly 85 minutes before anyone looked. Both nudging and killing it failed, so recovery meant answering the prompt through the terminal multiplexer by hand, which whitelists that single command pattern and leaves the next unfamiliar command to wedge the same way.

---
> • City: `factory-main` · Agent: `local-core.builder-1`
> • On behalf of: @austinborn

rjgeng:
Related to #5185 — does **not** close it. See below.

**What this PR actually does:** an earlier commit on this branch changed `BuildProviderResumeCommand`'s gate from "merge provider option defaults only on an explicit per-call override" to "merge whenever the provider has any option defaults configured at all," mirroring the launch-side gate. CI caught a regression: `TestBuildResumeCommandIncludesWrappedCodexResumeDefaults` started failing, because `resolve.go`'s `ResolveProvider` already runs `completeResumeCommandDefaults` over a provider's `ResumeCommand` once, at resolution time, filling in missing schema flags *in place* while preserving the template author's chosen flag ordering. The widened gate made `BuildProviderResumeCommand` re-merge those same defaults a second time, unconditionally stripping and re-appending every schema flag — which corrupts any provider with an explicit `ResumeCommand` template (confirmed on the codex-wrapped-resume case the failing test covers: `resume` collapsed to sit before its flags instead of after them, and the schema-canonical `--model` replaced the template's own `-m`).

This PR reverts that gate to its original `hasSchemaOptionOverrides` (explicit overrides only), with a comment explaining why it must stay narrower than the launch-side gate, and removes the regression test the earlier commit added — that test manually overwrote `ResolvedProvider.ResumeCommand` post-resolution to simulate the "fix," a shape no real call path produces.

**Why this doesn't close #5185:** the issue's reported symptom is specifically about the `claude` builtin provider losing `--dangerously-skip-permissions`/`--effort max` on resume. `claude` uses flag-style resume with `ResumeCommand` empty, so `BuildProviderResumeCommand` short-circuits on `command == ""` before this gate is ever evaluated — the gate change was never actually on the code path the issue is about. The real resume-time command for that path is built by `internal/worker/handle_lifecycle.go`'s `startCommand`, a separate function this PR does not touch. Flagging that as the more plausible fix location for `#5185`'s actual symptom, left for further investigation — no live repro confirmed yet, not guessing at a fix there blind.

**Testing:** `go build`/`gofmt -l` clean, all resume-related tests in `internal/config` and `cmd/gc` pass, including `TestBuildResumeCommandIncludesWrappedCodexResumeDefaults` (the one CI flagged).

---


sjarmak:
The whole diff is a six-line comment, so the comment's accuracy is the entire deliverable. I checked each clause against the source, and ran a cross-provider pass over the same questions. Three of the four claims hold. One overstates, and because there is nothing else in this PR, that is the blocking item.

**What holds.**

The launch/resume asymmetry is real: `BuildProviderLaunchCommand` gates on `hasProviderOptionValues` (launch_command.go:40), which is true for either `EffectiveDefaults` or explicit overrides (launch_command.go:124), while the resume gate uses `hasSchemaOptionOverrides` (launch_command.go:58), which looks only at explicit override keys. "Unlike the launch-side gate above" is accurate.

The consequence of widening is real, and worse than the comment says. `replaceResumeSchemaFlags` opens with `StripFlags(command, CollectAllSchemaFlags(schema))` (launch_command.go:141), and `CollectAllSchemaFlags` collects declared aliases as well as canonical `FlagArgs` (options.go:340, 356). So a widened gate would not merely re-append flags: for subcommand-style resume the reinsertion point is computed after the `resume` token (launch_command.go:145), so a template that positioned its flags before `resume` gets reordered, and a template using a declared short alias gets it replaced by the schema-canonical long form. `options_test.go:150` already pins the alias-stripping half of that.

The "does not close #5185" reasoning holds too. Builtin claude is flag-style, setting `ResumeFlag`/`ResumeStyle` with no `ResumeCommand` (`internal/worker/builtin/profiles.go:123-124`), so `BuildProviderResumeCommand` returns at the `command == ""` guard before the gate is evaluated. The pointer to `handle_lifecycle.go`'s `startCommand` is correct for that path: `Start`, `Attach`, and a blank `StartResolved` all route through it (handle_lifecycle.go:21, 43, 66), and it reaches the flag-style construction via `session.BuildResumeCommand` (handle_lifecycle.go:472, manager.go:2059). Leaving #5185 open is the right call.

**What overstates.** The comment asserts without qualification that "resolve.go already ran completeResumeCommandDefaults over ResumeCommand." That is conditional:

```go
// internal/config/resolve.go:97
if agent.ResumeCommand == "" {
    completeResolvedProviderResumeCommand(resolved)
}
```

When an agent sets its own `resume_command`, `mergeAgentOverrides` copies it verbatim (resolve.go:788) and the completion is deliberately skipped. For that shape the premise the comment rests on never ran, so a reader who trusts the comment will conclude the defaults are already present in `ResumeCommand` when they are not.

The conclusion survives, which is why this is a rewording and not a redesign: for an agent-supplied template the narrow gate is still correct, and for a stronger reason than the comment gives. That template is authored deliberately, so leaving it untouched is the point rather than an artifact of double-merging. But a comment that explains a live invariant with a premise that does not hold in one of the two shapes it governs is the kind of thing that gets cited later as settled fact.

Suggested replacement, which covers both shapes:

```go
// Unlike the launch-side gate above, this deliberately checks explicit
// overrides only. Two different reasons converge on the same narrow gate.
// For a provider-supplied ResumeCommand, resolve.go already ran
// completeResumeCommandDefaults over it (resolve.go:98), filling in
// EffectiveDefaults in place; gating on EffectiveDefaults here too would
// strip and re-append every schema flag, reordering a template whose flags
// were already positioned and replacing declared aliases with the
// schema-canonical form. For an agent-supplied ResumeCommand that
// completion is skipped by design (the agent's template is authoritative),
// so merging defaults into it here would override an explicit choice.
```

That is the only blocking item, and it is the whole change, so I would rather have it right than approve a comment we will quote back at each other. Happy to push the reword myself if you would rather not spin the branch.

For the record, `go test ./internal/config/ -run Resume` is green on the head, and the resume-alias and resolver-completion tests both pass.

sjarmak:
On the red checks, which are not yours: `Integration / packages-core-3-of-4` failed in `internal/testutil/providerledger` with eight lapsed runtime-provider waivers owned by `ga-80po0c.3` that expired 2026-08-26. That is a date-triggered failure, not a code one, and a six-line comment cannot reach it. #5610 (`bd84f0172`) landed the independent-expiry fix on main afterward; `go test -count=1 ./internal/testutil/providerledger/...` passes on current main and the branch just needs main merged in. `CI / required`, `CI / integration`, and `Evidence` are all downstream of that one shard.

sjarmak (dismissed):
Taking this, with one commit added on top.

**The comment is earned, and I confirmed the failure it describes by reproducing it.** Widening the gate to `hasProviderOptionValues(resolved, optionOverrides)` and running `TestBuildResumeCommandIncludesWrappedCodexResumeDefaults`:

```
got:  aimux run codex -- resume --dangerously-bypass-approvals-and-sandbox --model gpt-5.3-codex -c model_reasoning_effort=medium abc-123
want: aimux run codex -- --dangerously-bypass-approvals-and-sandbox -m gpt-5.3-codex resume -c model_reasoning_effort=medium abc-123
```

`resume` ahead of its flags and `-m` replaced by `--model`, exactly as the body says. The mechanism checks out too: `replaceResumeSchemaFlags` strips via `CollectAllSchemaFlags`, which covers declared aliases as well as canonical spellings (`internal/config/options.go:340`), then `ResolveExplicitOptions` re-emits canonical `FlagArgs` in schema order, and the subcommand is reinserted after that. So the reordering and the alias substitution are both structural, not incidental to this one fixture.

The asymmetry is real as stated: the launch gate is `hasProviderOptionValues`, which returns true on `EffectiveDefaults` alone (`internal/config/launch_command.go:130-135`), while the resume gate is `hasSchemaOptionOverrides`, which ignores them. And `resolve.go:718` really does complete the template in place, adding only missing flags around the untouched token slices rather than rewriting what is there.

**What the comment left out, and what I added.** That resolve-time completion is conditional:

```go
mergeAgentOverrides(resolved, agent)
if agent.ResumeCommand == "" {
    completeResolvedProviderResumeCommand(resolved)
}
```

When an agent supplies its own `resume_command`, `mergeAgentOverrides` copies it verbatim (`resolve.go:786`) and the completion never runs. That shape then hits this narrow gate and gets nothing, so a provider's `option_defaults` reach the launch command and not the resume command. `resolve.go:718` is the only non-test site where `EffectiveDefaults` reaches a `ResumeCommand`, so nothing downstream compensates. The comment as written says the completion "already ran," unqualified, which would read to the next person as coverage that does not exist there.

That gap predates this branch, so it is not something this PR broke. But an in-code comment asserting coverage is the kind of thing that gets trusted later, so I amended it in `b43f979d9` to name the uncovered shape and point at the tracking issue rather than sending the PR back for one sentence. Filed the gap as #5659, including why widening this gate is the wrong fix for it and why the completion-side guard is the more likely place.

**On #5185.** Agreed that this is not it, and the reason holds: builtin claude is flag-style, sets `ResumeFlag`/`ResumeStyle` with no `ResumeCommand`, so `BuildProviderResumeCommand` returns at the `command == ""` check before this gate is read. The resume command for that path is built in `internal/session/manager.go` via `startCommand`, reached from `Start`, `Attach`, and blank `StartResolved` in `internal/worker/handle_lifecycle.go`. Naming the likely location without guessing at a fix there is the right call; leaving #5185 open is correct.

The red on this branch was the provider-ledger waiver expiry that hit every PR on 2026-08-26, not anything diff-owned. That is fixed on main, and I updated this branch so CI picks it up.

Codex (gpt-5.4) reviewed independently and is what surfaced the agent-override shape; it confirmed the other three claims against the source.

sjarmak:
Tightened the comment in `c141c941e` after a second pass. Two things in my first wording were too strong: an agent-supplied template does still pick up `EffectiveDefaults` here when an explicit per-call override is present, since `providerOptionArgs` merges defaults with overrides before the rewrite, so the loss is specific to the plain-resume case; and the re-append side emits only the options present in the merged set, not every schema flag unconditionally. The strip side is the unconditional half. Issue #5659 carries the same correction.

sjarmak (approved):
Re-approving at `bbfdd4f3b`. The earlier approval was dismissed by the two amendment commits, so this one covers the current head.

The cross-provider re-run against `c141c941e` found one more overstatement in the comment I had written: "strips every schema flag" is wider than what `replaceResumeSchemaFlags` does. `StripFlags` (`internal/config/options.go:393`) matches on exact token equality against the set `CollectAllSchemaFlags` collects, which is the declared `FlagArgs` and `FlagAliases` only, so an undeclared alias survives the strip. `internal/config/options_test.go:172` pins that. Corrected in `bbfdd4f3b`; the rest of the sentence, re-emitting the merged set in schema order, checks out against `providerOptionArgs` and `ResolveExplicitOptions`.

That is the third comment-accuracy fix on a comment whose whole job is accuracy, which is a decent argument that the narrow gate deserved the documentation. Nothing in the shipped behavior changes in any of the three.

PR #5617, opened 2026-08-25
fix(tmux): don't clear-before-paste on an attached session (#5192)

### LINKED ISSUE #5192: `gc session nudge` can silently strand a message on an agent's input line
**Upstream:** https://github.com/gastownhall/gascity
**Version:** gascity 1.4.0 (Homebrew bottle, `arm64` macOS 25.5.0)
**Filed from:** city `gc-greenwood`, bead `gg-r4ew`
**Severity:** high — silent, unlogged loss of agent dispatch, with no signal on any health surface

---

## Summary

`gc session nudge` delivers a message to a tmux-backed agent as **two separate
keystroke operations** — send the text, then send Enter — and **never clears the
target's input line first**.

Two failure modes follow:

1. **Stranded.** If the Enter step does not land, the message sits unsubmitted on
   the agent's input line. The agent has not seen it and goes idle. Nothing
   reports this: the session is `active`, `gc session list` is clean, orders keep
   firing, and the nudge itself exits `0`.
2. **Concatenated.** Because the line is never cleared, the *next* nudge is
   appended to the stranded fragment and submitted as a single nonsense string.

Mode 1 is the expensive one. The lane is not crashed, not drained, and not
flagged — it is simply never told to do anything again, and no watchdog in the
system is looking at the input line.

## Demonstration: the line is not cleared, and the fragment corrupts the next message

This is reproducible in a few seconds against any live tmux-backed lane. Put a
fragment on the input line (simulating a nudge whose Enter was dropped), then
send a perfectly normal nudge:

```console
$ tmux -L <city> send-keys -t <pane> -l -- 'STRANDED_PREFIX_'
$ tmux -L <city> capture-pane -p -t <pane> | grep '❯'
❯ STRANDED_PREFIX_
$ gc session nudge <lane> "SUFFIX_this_is_a_separate_message"
Nudged <lane>
```

`gc` reports success. What the agent actually received, read back from its own
transcript (`~/.claude/projects/<workdir>/<session>.jsonl`), is a **single** user
message with the two glued together:

```json
"content":"STRANDED_PREFIX_<system-reminder>\nYou have a deferred reminder that was queued until a safe boundary:\n\n- [session] SUFFIX_this_is_a_separate_message\n\nHandle them after this turn.\n</system-reminder>"
```

Note this happened on the *queued* delivery path — so the corruption is not
confined to one delivery mode. Whatever is sitting on the input line is
submitted along with the message, as one string.

## Corroborating evidence in the shipped binary

Consistent with the above, `/opt/homebrew/Cellar/gascity/1.4.0/bin/gc` carries
two independent error strings for what should be one delivery:

```console
$ strings -a gc | grep -o 'send-keys failed: %w: %w'
send-keys failed: %w: %w
$ strings -a gc | grep -o 'failed to send Enter: %w'
failed to send Enter: %w
```

Two separate error paths imply two separate send operations — the text can
succeed while the Enter fails, which is failure mode 1.

There is also no sign of a line-clear key ever being sent. `strings` finds
exactly one `C-u` byte sequence in the entire binary, and it is an incidental
substring inside an unrelated packed string table, not a tmux key argument:

```console
$ strings -a gc | grep 'C-u'
Env[-][+][ ][~]C-uSunMonTueWedThuFriSatJanFeb…
```

(Stated carefully because it is inference from a stripped binary rather than
from source — the demonstration above is the load-bearing evidence, and it does
not depend on this.)

## What is and is not established here

Being precise about this, because the incident that prompted the investigation
did **not** survive it.

**Established, and reproducible on demand:** the input line is never cleared, and
whatever sits on it is submitted along with the next nudge as one string. That is
the demonstration above, read back from the receiving agent's own transcript.

**Inferred, not directly observed:** that the Enter step can fail on its own and
strand a message. It follows from the two independent error paths in the binary,
and gc's own source comment (quoted in a downstream pack) acknowledges that "the
pane may hold a half-typed, unsubmitted line" — but this report does not contain
a captured instance of it.

**Retracted:** an earlier draft cited four lanes stranded simultaneously in one
city on 2026-08-11. That evidence came from a monitoring order that read the
pane with `capture-pane` and no `-e`, and it was wrong — every one of those lanes
had `cursor_x == 2`, an empty buffer under a dim history hint. See the detection
caveats below; they are exactly the trap that produced those false positives, and
they are the reason this section is worded as carefully as it is.

The concatenation mode alone justifies the fix. Any fragment left on the line —
by a dropped Enter, by a human who typed half a question and walked away, or by
anything else — corrupts the next message, and `gc` reports success either way.

## How to detect it

The reliable signal is the **cursor column**, not `capture-pane`. The Claude Code
prompt places the cursor at `base + len(buffer)`, where `base` is the column of
the empty prompt (2 for `❯ `):

```console
$ tmux -L <city> display-message -p -t <pane> '#{cursor_x}'
2                                  # empty input line
$ tmux -L <city> send-keys -t <pane> -l -- 'ZZprobeZZ'    # 9 chars, no Enter
$ tmux -L <city> display-message -p -t <pane> '#{cursor_x}'
11                                 # 2 + 9  => text is stranded in the buffer
```

Two caveats for anyone writing a check against this, both of which cost us time:

* **The cursor lags the text.** Sampled immediately after `send-keys`, tmux still
  reports the *previous* cursor position; it settles ~1–2 s later once the TUI
  repaints. A single immediate sample reports a stranded line as clean. Poll
  until the value is stable.
* **`capture-pane` alone is ambiguous.** The TUI renders dim placeholder text
  (e.g. `<no suggestion>`) on an *empty* input row, which is indistinguishable
  from a stranded message in a plain capture. It is emitted as `ESC[2m … ESC[0m`
  and real typed input never is, so dim spans must be discarded before judging.

## Suggested fix

1. Clear the input line before typing (`send-keys C-u`).
2. Send the clear, the text, and the Enter as **one** `tmux send-keys`
   invocation, so no other client's keystrokes can interleave:

   ```sh
   tmux send-keys -t "$pane" C-u \; \
        send-keys -t "$pane" -l -- "$msg" \; \
        send-keys -t "$pane" C-m
   ```
3. **Verify rather than assume.** After delivery, confirm the input line returned
   to its base column; if it did not, the Enter did not land. Today the command
   reports success without ever checking that the message was submitted, which is
   what makes the failure silent.

Two safety rules worth building in, learned while writing a local workaround:

* Do not clear a pane that a client is **attached** to — a human may be
  mid-keystroke. `attached` is already exposed in `gc session list --json`.
* Do not send Enter to a pane that is not at a text prompt. A multiple-choice
  dialog draws its selected option with the same `❯` marker, so a glyph-only
  check mistakes a menu for a prompt and Enter silently *picks an option*.
  Requiring the row to sit between the input box's two horizontal rules
  distinguishes them reliably.

## Local workaround

`gc-greenwood` has a city-local
`<city>/assets/scripts/atomic-nudge.sh` available as a drop-in replacement
(callers opt in; nothing is intercepted). It delegates delivery to
`gc session nudge` (preserving `--delivery`, the ACP path for paneless lanes,
TTL and bookkeeping) and adds only the missing pre-flight clear, verification,
and repair. It is a wrapper, not a fork, so a fix here removes the need for it
entirely.

rjgeng:
Fixes #5192.

Worth flagging up front: the primary concatenation bug this issue originally reported was already fixed by an unrelated commit (#5013) that landed on `main` before this build started — caught and re-scoped mid-build rather than duplicating that fix.

**What's actually fixed here:** the issue's own safety note flagged one real remaining gap — the existing clear-before-paste behavior fires even on attached sessions, which risks wiping a human's in-progress keystrokes if they're actively typing in that tmux pane. Added a guard so clear-before-paste only fires on unattached sessions.

---


sjarmak (dismissed):
Verified this against both revisions rather than reading the diff, because the guard's correctness turns on a question the diff cannot answer: `IsSessionAttached` counts *any* tmux client, and gc attaches clients of its own.

**RED on main, with the right symptom.** Copied the new test onto `origin/main` and ran it:

```
--- FAIL: TestNudgeSessionSkipsClearWhenAttached (1.38s)
    attached session's pending input was cleared; want it preserved:
        nudge-message
        nudge-message
```

`human-typing` is gone from the pane. That is the bug the guard fixes, shown rather than argued. On the PR head all ten `TestNudgeSession*` cases pass, including `TestNudgeSessionClearsPendingInputBeforePaste`, which pins the unattached path the guard must not disturb.

**The obvious objection does not hold, and it is worth recording why.** `IsSessionAttached` shells out to `#{session_attached}` and cannot tell a human apart from gc's own hidden attach client. `ensureHiddenAttachedClient` spawns a real long-lived `attach-session` process, keeps it in `hiddenAttachClients[target]` for up to `hiddenAttachMaxLifetime` (20s, tmux.go:165), and is called in production from `Provider.Interrupt` (adapter.go:272) and `Provider.ResetInterruptedTurn` (adapter.go:470). Nothing closes it except `Stop` (adapter.go:250) or an error path. So the reasonable worry is that a gemini session, which is the family `requiresHiddenAttachedInterrupt` returns true for (tmux.go:1377-1386), spends 20 seconds after every interrupt looking "attached" to itself, silently disabling clear-before-paste on exactly the sessions most likely to be carrying a stranded draft.

It does not happen, because `NudgeNow` diverts first. `sendHiddenAttachedText` returns `used=false` only when `hiddenAttachClient(target) == nil` (tmux.go:1616-1620), so whenever gc holds a hidden client the nudge is written through that client and `NudgeSession` is never reached (adapter.go:577-584). The guard therefore only ever sees an attachment gc did not create. The one direct caller that bypasses this, `tmuxStartOps.sendKeys` (adapter.go:928), runs during start and relaunch, before any client exists.

**The moved sleep is correctly scoped.** The 50ms belongs to step 1: it paces the `C-u`, and the paste has its own debounce at step 3 (`nudgeSubmitDebounce`). Moving it inside the guard does not shorten the pre-paste settle on the path that still clears. On the attached path `WakePaneIfDetached` is also a no-op, which is right, since an attached pane is already servicing its event loop.

**The failure direction is arguable, and I want to record the argument rather than wave it through.** `IsSessionAttached` collapses every `display-message` failure to false (tmux.go:1337), so an unknown attachment state falls through to clearing. My first read was that this is the right default. The cross-provider pass took the other side, and it has the better version of the argument: the guard exists because wiping a human's in-progress keystrokes is worse than concatenation, and "I could not determine whether a human is attached" is exactly the state where that asymmetry should govern. Concatenation is ugly and recoverable; a wiped draft is not.

What keeps this off the blocking list is how narrow the exposed window is. The common failure modes, `ErrNoServer` and `ErrSessionNotFound`, mean the following `send-keys` fails too, so nothing is wiped either way. The exposure is only a transient probe failure on a session that is still alive and still attached. Filed as #5654 rather than left here, with the suggested shape: separate "no server or session" (clear is moot) from an unexpected probe error (skip the clear). Not worth holding this fix, which is a strict improvement over unconditionally clearing.

**Ledger bookkeeping is in lockstep.** `fixed_sleep` 456 to 459 matches the three `time.Sleep` calls the new test adds, `BaselineFiles` correctly stays 165 since the file was already counted, `TESTING.md` mirrors `census.go`, and the manifest count moves 390 to 391. `go test ./internal/testpolicy/resourcecensus/... ./scripts/...` is green on the head.

On whether `Fixes #5192` is earned, since this PR fixes neither of the two modes the issue names: it is, and the check is worth writing down so nobody redoes it. Mode 2, concatenation, went in with #5013 as the body says. Mode 1, the stranded message that exits 0, is closed for the claude family the issue was filed against: `submitVerifyEligible` covers claude and codex (tmux.go:2074), and on that path an unconfirmed submit returns `ErrNudgeSubmitUnconfirmed` rather than nil, with a comment at tmux.go:2251 spelling out that collapsing it to nil would let the queue ack and delete a message still sitting in the composer. So the issue's own "exits 0 with no signal" premise no longer holds there. Families outside that list keep best-effort delivery, but that is a documented limit of their busy indicators, not something this PR owes.

The cross-provider pass reached the same conclusion on the hidden-client question independently, and re-ran the census and both integration tests. Its one substantive finding is the fail-direction point above, now #5654.

Approving.

sjarmak:
Approved above, but the branch has gone conflicting since it was opened and I did not want to rebase it under you. `git merge-tree` puts the conflict in one place:

```
changed in both
  base   TESTING.md
  our    TESTING.md
  their  TESTING.md
@@ -451,7 +451,13 @@
```

That is the resource-census ledger table, which is the one family here where taking either side wholesale is unsafe. `internal/testpolicy/resourcecensus/census.go`'s `bootstrapPolicy` is the source of truth, `test/test-resources.toml` must mirror it exactly, and the `TESTING.md` table is generated from the toml rather than edited. Two branches that each add `time.Sleep` call sites independently both compute their counts from their own base, so resolving by picking a side gives a number that is stale by the other side's delta and no test will catch it as a conflict, only as drift.

The resolution that works: take the current main baseline as the starting point rather than merging the numbers, then run

```
go test ./internal/testpolicy/resourcecensus/... -count=1
```

which prints exactly which counters your rebased branch's own new code requires bumping. Apply only those deltas to `census.go` and `test-resources.toml` in lockstep, then regenerate the doc table with

```
```

and re-run the check clean. Your three new `fixed_sleep` sites should survive the rebase as three, but read the diagnostic rather than assuming 459 is still the right total once main's own additions are in.

The `Evidence / critical-path suites` failure is downstream of the dirty merge, not of your diff. Ping me once it is rebased and I will merge.

rjgeng:
Rebased, per your instructions above — took main's current baseline as the starting point, ran `go test ./internal/testpolicy/resourcecensus/...` to get the actual diagnostic delta rather than merging numbers, and applied only that. Current head: `a0d3dc44c`.

One thing worth flagging rather than staying quiet about: `CI / required` is currently red, but not on anything this PR touches. `cmd/gc process / shard 1 of 12` is failing on `TestCmdStopSupervisorManagedInvalidCityTomlFailsWhenShutdownFails` — confirmed it fails identically and deterministically on plain `main` at the current tip (`8556a801c`) in a clean worktree with none of this PR's changes applied, so it's not diff-owned here either. Filed as #5668 with the repro and a guess at the likely commit.

Ready whenever you are.

quad341 (approved):
@rjgeng — thank you for this one, and especially for how you handled the re-scoping. Finding mid-build that #5013 had already landed the concatenation fix and saying so plainly at the top of the PR body, instead of quietly shipping a duplicate, saved real reviewer time. Picking up the issue's own safety note and turning it into the narrow guard it was asking for is exactly the right instinct.

The rebase came out clean and the resource-census ledger is right: your three new `fixed_sleep` sites are tracked in lockstep across `census.go`, `test-resources.toml`, and the generated table, `BaselineFiles` correctly stays put, and the manifest and shard arithmetic check out. That family is easy to get subtly wrong, and you followed the diagnostic rather than hand-merging numbers. The test asserting the draft *survives* is the right direction — it fails loudly if the guard ever regresses.

One small thing I'll take on my side: the comment above the feedback-survey step still says the `C-u` guarantees an empty composer, which your guard makes untrue on the attached path. I'll correct that comment and file the survey-plus-attached-draft interaction as a follow-up — not something to send back to you.

Fixing that up and merging. Nice work.


---

PR #5618, opened 2026-08-25
fix(import): validate composed config after install, not just at doctor time (#5382)

### LINKED ISSUE #5382: bug: import install succeeds on a composed config the loader will refuse; watcher-reload aborts stay log-only and effective config freezes
### Before you continue
- [x] I searched existing issues and did not find a duplicate. (Nearest family: #4523/#4540 — install/check/loader disagreement, fixed for the recursion case; #4185 — reload applied but printed "no changes"; #2897 — silent no-op on pool resize; #5312 — import upgrade count. None covers install succeeding on a composed config that the loader will refuse.)

### Gas City version
Observed across recent mains (currently running a build from main @ 0b10e4e4d lineage).

### Reproduction
1. Have a working city with an existing agent (e.g. a named session `gastown.mayor` from one import).
2. `gc import install` a second pack whose composed result collides with the existing agent (v1/v2 layout collision, or any composed-config validation error).
3. Install reports success and writes the lockfile.
4. Subsequent config reloads abort at validation. The *manual* `gc reload` path does report its failure and keeps the old config — but watcher-sourced reloads surface the abort only in the supervisor log, and `ValidateRigs` findings are `appendWarning`-only (reload continues past them). The effective config is now frozen at the pre-install state while the on-disk config says otherwise.

Operators then patch config for hours with zero effect, because nothing at the interaction surface says "your reloads have been failing since the install." In our worst instance the town ran a full window on stale effective config before the divergence was noticed.

### Expected
Two-part expectation, both consistent with the install/check/loader agreement principle from #4523 ("install/check/loader currently disagree about that flow") and the doctor's existing authority:

1. **Install validates what it composed**: `gc import install` runs the same composed-config validation the loader (and `gc doctor`'s existing `expanded-config-load` check) already applies, and **fails the install** when the result cannot load. The check exists; install just doesn't consult it.
2. **Reload outcome is a first-class status surface**: the controller already records its last reload outcome; `gc status` does not show it. A city whose most recent (watcher-sourced) reload aborted should say so in `gc status` and fail `gc doctor` — not only log it.

### Actual
Install succeeds; the lockfile updates; every later automatic reload aborts at validation with the failure visible only in the supervisor log; `gc status` shows nothing; effective config is frozen indefinitely.

### Additional context
- #4540 fixed exactly this shape for recursive imports in 3 days — this is the same class one step wider (validation, not recursion).
- #4185 and #2897 are siblings on the "reload surface tells the truth" theme; #5312 on install-report truthfulness.
- We currently compensate with a local pre-reload validation gate, which has already caught one live-fatal composed config (a dangling `[[patches.agent]]` after its target was removed). We'd happily retire it.

rjgeng:
Fixes #5382 (part 1 of 2).

`gc import install` could succeed on a composed config that the loader would later refuse outright — the install-time path didn't run the same validation `gc doctor` already catches, so the failure only surfaced later, disconnected from the actual install action that caused it.

**Fix:** `gc import install` now validates the composed config after installing, catching the same class of error `gc doctor` already catches, at the point where it's actionable.

**Scope note:** deliberately scoped to just this — part 2 of the original issue (a reload-visibility gap) is a separate design question, not touched here.

---

### DIFF
--- a/cmd/gc/cmd_import.go
+++ b/cmd/gc/cmd_import.go
@@ -734,10 +734,13 @@ func doImportInstall(cityPath string, stdout, stderr io.Writer) int {
 	// Nothing rolls back: the packs are in the cache and packs.lock is
 	// written by the time this runs, and the load error may predate this
 	// command. Say so, or a non-zero exit reads as "nothing was installed"
-	// and the operator re-runs instead of fixing the config.
+	// and the operator re-runs instead of fixing the config. No count here:
+	// the success line's len(lock.Packs) is the lock closure by source, not
+	// a count of import bindings, and it is 0 for a local-path-only city
+	// whose config is nonetheless on disk.
 	if cfgErr := validateComposedConfigAfterInstall(cityPath); cfgErr != nil {
-		fmt.Fprintf(stderr, "gc import install: composed config failed to load after install: %v\n", cfgErr)                                          //nolint:errcheck
-		fmt.Fprintf(stderr, "%d remote import(s) and packs.lock are on disk; fix the config, then re-run `gc doctor` to confirm.\n", len(lock.Packs)) //nolint:errcheck
+		fmt.Fprintf(stderr, "gc import install: composed config failed to load after install: %v\n", cfgErr)                        //nolint:errcheck
+		fmt.Fprintln(stderr, "The installed packs and packs.lock are on disk; fix the config, then re-run `gc doctor` to confirm.") //nolint:errcheck
 		return 1
 	}
 
--- a/cmd/gc/cmd_import.go
+++ b/cmd/gc/cmd_import.go
@@ -731,8 +731,13 @@ func doImportInstall(cityPath string, stdout, stderr io.Writer) int {
 	// expanded-config-load check applies. Validate here so install fails
 	// closed on the same errors doctor would later catch, instead of
 	// succeeding while every subsequent config reload aborts. See #5382.
+	// Nothing rolls back: the packs are in the cache and packs.lock is
+	// written by the time this runs, and the load error may predate this
+	// command. Say so, or a non-zero exit reads as "nothing was installed"
+	// and the operator re-runs instead of fixing the config.
 	if cfgErr := validateComposedConfigAfterInstall(cityPath); cfgErr != nil {
-		fmt.Fprintf(stderr, "gc import install: composed config failed to load after install: %v\n", cfgErr) //nolint:errcheck
+		fmt.Fprintf(stderr, "gc import install: composed config failed to load after install: %v\n", cfgErr)                                          //nolint:errcheck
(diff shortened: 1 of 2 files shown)


sjarmak (approved):
Taking this. The gate is the right one and the test is a real regression test; I added a line to the failure path and pushed it here rather than sending the PR back.

**The test is red on unfixed main.** At the merge-base `6b60167cf5`, `TestDoImportInstallRejectsUnloadableComposedConfig` fails with `code = 0, want 1` and `Installed 0 remote import(s)` on stdout, so it encodes the new contract instead of describing existing behavior. The validator is not a second implementation of anything either: `loadCityConfig(cityPath, io.Discard)` is literally the call `gc doctor`'s expanded-config-load check makes at `cmd/gc/cmd_doctor.go:459`, so install now fails on exactly the set doctor would later report.

**What I added.** Nothing rolls back when the validation fails. `packs.lock` is already written by then, any fetched packs are already in the cache, and the load error can predate the command entirely. Meanwhile the "Installed N remote import(s)" line is suppressed, because it sits after the early return. So the operator sees a non-zero exit and one error line, which reads as "nothing was installed" and invites a re-run instead of a config fix. Landed in `cc12dfa73` with a second stderr line saying what is on disk, and the assertion in the same commit.

Two rounds of wording on that line, both worth recording. It first carried `len(lock.Packs)` as a count of "remote import(s)", copying the success line above it. That count is the lock closure keyed by source (`internal/packman/lockfile.go:22`, `internal/packman/install.go:453`), not a count of import bindings, and it is 0 for a city whose imports are all local paths, since those are never locked or fetched (`internal/packman/install.go:349`). The mislabel is pre-existing on the success line and out of scope here; propagating it into a new line was not. The count is gone, and the comment says why. The replacement then claimed installed packs were on disk, which is false on that same local-path-only path. It now states only what holds everywhere it can print: `packs.lock` is written unconditionally at `cmd/gc/cmd_import.go:716` before validation runs, including on the nothing-to-do path where `syncLock` returns an empty lock (`internal/packman/install.go:202`), and fetched packs, if any, remain cached.

**Scope, filed rather than deferred.** The gate covers `gc import install` only. `gc import add`/`remove` (`internal/importsvc/importsvc.go:263`, `:275`, `:309`, `:331`), `gc import upgrade` (`cmd/gc/cmd_import.go:796`), and the doctor/init/rig helpers that materialize imports can all still leave an unloadable composed config while reporting success, and `gc import check` reports green right after this command exits non-zero. Where that gate belongs is a real design question rather than a copy-paste of this call, so it is #5662 rather than a change request here. This PR is part 1 of 2 on #5382 and stands on its own.

Codex (gpt-5.4) reviewed independently across three passes and found both wording defects above; the final pass against `cc12dfa73` returned no remaining blocking defect.

Issue #5606, opened 2026-08-25
Design check: third-party `runtime-nomad` pack (RPP v0) — two questions before a pack PR


jacobhausler:
I'd like to add a `runtime-nomad` pack to gascity-packs: an RPP v0 runtime that
drives session lifecycle through a HashiCorp Nomad cluster. #3613 already
settled the shape, so this is a short check rather than an RFC, reduced to two
open questions plus a docs offer.

**What I'm building, and why most of it needs no discussion.** A pack
mirroring `gascity-packs/runtime-cloudflare/` one-for-one: `pack.toml` with
`[runtimes.nomad]` (`command = "gc-runtime-nomad"`, `protocol = 0`) per
RUNTIME-SEL-011; a nested Go module with no `gastownhall/gascity` import and no
Nomad SDK in `go.mod` (the cluster is reached over its HTTP API, matching the
vendor-cleanliness precedent quad341 called out on #2540 for an external
contributor's remote transport); `install.sh` doing `go install`; and a
`conformance.sh` that runs `gc runtime check` (RUNTIME-RPP-010, the
no-Go-imports validator, not the golden RUNTIME-RPP-011 suite) against a fake
Nomad HTTP API — `conformance.sh` already exists and runs offline today, so CI
needs no live cluster. On the contract, `internal/runtime/runtimecontract/
catalog.go` defines 8 requirements (`RPP-PROTOCOL-001`,
`RPP-LIFECYCLE-001..005` required; `RPP-CONN-001` `exec`, `RPP-PROVISION-001`
`provision` optional) — not the ~21-op dispatch list in `exec.go` — and I'm
targeting all 8 from day one, plus the six legacy driving ops gc still uses.
I'm building against polling `peek` (per your close on #3631) and reporting
`is-running` unavailability as non-`true`/`false` output, which `boxOccupancy`
already reads as "I don't know" — no wire change needed there. **Assumption:**
we will implement the `env.ledger` capability tunnel (the sandbox→host
`GC_BEADS_API` path `bd ready` needs) in v1, so that's not a question here.

**Questions.**

1. **Registry admission and post-merge ownership.** Would you take a
   third-party-maintained `runtime-nomad` into `gascity-packs/registry.toml`?
   After merge, does CI and upkeep sit with me or with you? I'm volunteering to
   own it — conformance runs offline against fake Nomad in the pack's own CI,
   the same shape `runtime-cloudflare` uses (though packs#307, still
[...]

PR #5619, opened 2026-08-25
docs(pack-spec): document [runtimes.<name>] pack manifest tables

jacobhausler:
Follow-up to the offer in #5606: `docs/reference/specs/pack-spec.md` documented `[providers.<name>]` but never `[runtimes.<name>]`, which lived only in the generated pack schema and `internal/config`.

Adds §1.2.8 `[runtimes.<name>]` (fields `command`/`protocol`, name rules, city composition + collision semantics, `gc runtime check` / `pack-runtimes` doctor) and the matching rows in the content table, section list, and authoring summary. Derived from `docs/reference/schema/pack-schema.json` (PackRuntimeEntry), `internal/config/pack_runtimes.go`, `internal/runtime/REQUIREMENTS.md` (RUNTIME-SEL-011/012), and the existing `exec-session-provider.md` prose; no fields documented beyond what the schema defines.

Docs-only; no code changes.

https://claude.ai/code/session_01KoMPoPbV2gsgSEguHovjhC

### DIFF
--- a/docs/reference/exec-session-provider.md
+++ b/docs/reference/exec-session-provider.md
@@ -47,7 +47,9 @@ Rules:
   errors — no silent shadowing. Identical re-declarations of the same
   pack reached through a diamond import graph dedupe.
 - The `pack-runtimes` doctor check verifies each declared executable is
-  installed and answers the `protocol` handshake.
+  installed and that its RPP handshake works. An executable with no
+  `protocol` op is the version-0 floor and passes; a present-but-broken
+  handshake fails.
 - Config reload enforces the same registration rules, and rebuilds the
   session provider when the declaration behind the selected name changes
   (the executable binding is fixed at provider construction).
--- a/docs/reference/exec-session-provider.md
+++ b/docs/reference/exec-session-provider.md
@@ -53,8 +53,10 @@ Rules:
 - Config reload enforces the same registration rules, and rebuilds the
   session provider when the declaration behind the selected name changes
   (the executable binding is fixed at provider construction).
-- `gc runtime check <name>` resolves the declared name and runs the
-  full conformance suite against the pack's executable.
+- `gc runtime check <name>` and `gc runtime conformance <name>` both
+  resolve the declared name and run against the pack's executable:
+  `check` is the smoke test, `conformance` the full requirement-coded
+  suite.
 
 ## Calling Convention
 
--- a/docs/reference/specs/pack-spec.md
+++ b/docs/reference/specs/pack-spec.md
@@ -496,7 +496,8 @@ declared name and run against that pack's executable: `check` is the smoke
 test (handshake, lifecycle round-trip, declared capabilities, optional
 probes), and `conformance` is the full requirement-coded RPP suite. A
 `pack-runtimes` doctor check verifies each declared executable is installed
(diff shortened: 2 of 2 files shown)


sjarmak (approved):
Taking this. The section was needed and is derived from the right sources; I corrected four claims and one pre-existing one in the doc it was derived from, rather than sending it back for a docs-only round trip.

**Verified as written.** The name rules match `internal/config/pack_runtimes.go:52` exactly, including the reasons given for `:` and `/`. `protocol` really is 0-only and a load error otherwise (`:58`). The rig-imported-packs claim holds. `pack-runtimes` exists at `cmd/gc/doctor_pack_runtimes.go:29`, and `City.Runtimes` is the registry composition writes to.

**What I changed, and why.**

`command` resolution had three cases documented as two. `resolveRuntimeCommand` (`internal/config/pack_runtimes.go:75`) tests `filepath.IsAbs` first, so an absolute path is used as written and is not anchored at the pack directory, even though it contains a separator. The row now names bare, relative, and absolute separately, and the example comment no longer describes a rule the example does not exercise.

The selection sentence was half right. `[session].provider` is correct and is what `PackRuntimeEntry`'s own doc comment promises (`internal/config/config.go:1112`); `sessionProviderContextForCity` (`cmd/gc/providers.go:69`) reads `GC_SESSION` then `cfg.Session.Provider` and nothing else. An agent's `session` field is a session *transport*: `internal/config/validate_semantics.go:62` warns on anything but `acp`, `tmux`, or empty, and `cmd/gc/template_resolve.go:161` fails resolution on a non-transport value. A reader following the old sentence would have written `session = "cloudflare"` and gotten a warning and a broken agent. Now says city-wide only, names `GC_SESSION`, and says what `session` actually is.

The dedupe sentence said "the same pack", which is the one reading `mergeCityRuntimes` deliberately rejects: `sameRuntimeDeclaration` (`:117`) keys on the resolved pack *directory*, and the comment at `:95` spells out that two distinct directories conflict even when pack name, command, and protocol all coincide, so provenance is never re-attributed. `TestMergeCityRuntimes_SameDirReDeclarationDedupes` pins it.

`gc runtime check` is the smoke test, not the conformance suite. `cmd/gc/cmd_runtime_conformance.go:18` states the split in as many words, and both commands accept a pack-declared name. The spec now documents both.

The doctor sentence overstated the requirement in a way that would fail a valid pack on paper: a missing `protocol` op is the documented version-0 floor and passes (`cmd/gc/doctor_pack_runtimes.go:17`, `internal/runtime/exec/handshake.go:13`, proven by `doctor_pack_runtimes_test.go:48`); only a present-but-broken handshake fails. That wording came from `docs/reference/exec-session-provider.md`, which had it wrong first, so I fixed it there too, along with its own "full conformance suite" line for `gc runtime check`. Two docs now agree with each other and with `internal/runtime/REQUIREMENTS.md` RUNTIME-SEL-011.

Fixes are in `d0f24af7f`, `92cde9f80`, `ef1b01a79`, `a1501247f`. The cross-provider pass came back clean against `a1501247f`; the selection error and the doctor overstatement are both its finds.

Non-gating: the `protocol` row does not say the field is optional, though omitting it yields 0 and validates. Not worth another commit.

PR #5620, opened 2026-08-25
Deflake semantic gc stop tests without weakening timeout ownership

quad341:
## What this changes

Two semantics-focused `gc stop` tests no longer inject arbitrary positive wall-clock limits into `cmdStop`. The force-escalation case uses the default input while preserving its interrupt, forced-stop, and exit-code assertions. The invalid-city-config case runs the stop operation asynchronously and observes completion through the package's existing hang detector, so a real hang remains bounded without turning the observation budget into product behavior.

This keeps explicit timeout ownership in the dedicated timeout tests and makes the semantic cases resilient to loaded-host scheduling delays.

## Review notes

- This is test-only; `gc stop` runtime behavior and user-visible output are unchanged.
- Dedicated wall-clock timeout tests, the justified city-runtime latency assertion, and the shared test hang budget are untouched.
- In the invalid-config error path, zero removes the arbitrary outer test override; the path may return before the normal config-derived timeout is consulted. The assertion remains about shutdown-error propagation, not the default timeout value.

## Test plan

- [x] Run both modified tests repeatedly with result caching disabled and confirm their semantic assertions pass.
- [x] Run the full `cmd/gc` process shards and the repository-wide local CI-equivalent suite.
- [x] Run policy, docs, build, vet, native-DoltLite, formatting, and affected-lint checks.
- [x] Release gate: [`release-gates/ga-imv1fb-cmd-stop-test-deflake-gate.md`](release-gates/ga-imv1fb-cmd-stop-test-deflake-gate.md)

### DIFF
--- a/cmd/gc/cmd_stop_test.go
+++ b/cmd/gc/cmd_stop_test.go
@@ -362,7 +362,7 @@ func TestCmdStopForceEscalatesInProgressControllerStop(t *testing.T) {
 	var normalStdout, normalStderr lockedBuffer
 	normalDone := make(chan int, 1)
 	go func() {
-		normalDone <- cmdStop([]string{dir}, &normalStdout, &normalStderr, 10*time.Second, false)
+		normalDone <- cmdStop([]string{dir}, &normalStdout, &normalStderr, 0, false)
 	}()
 
 	interrupted := sp.waitForInterrupts(t, 1)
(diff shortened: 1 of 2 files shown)


quad341:
@quad341 — nice piece of test surgery. Zeroing the input so production owns the cap, instead of widening the test's own override, is exactly the distinction `TESTING.md` draws between a value the test *feeds* the system and how patiently the test *watches* it. Leaving the dedicated wall-clock tests and `hangBudget` untouched means both cases got deflaked without losing a single assertion.

I traced it rather than taking the description on faith: the force-escalation case now gets a ~114s config-derived cap in place of the old hardcoded 10s, so it can only get more forgiving under load, never less. And I appreciate that you flagged the limit of your own change — the invalid-config path does exit at the bead-store shutdown failure before that cap is consulted, just as your Review notes say. The comment on that hunk is a touch broader than the path it describes, but the assertion it guards is about error propagation, not timeout value, so it isn't worth a round-trip.

The `cmd/gc` process shards ran this file green. Merging as-is — thanks for chasing a loaded-host flake down to the actual ownership question instead of just bumping a number.


---

PR #5621, opened 2026-08-25
fix(dolt): name failed databases in sync summary

quad341:
## What this changes

`gc dolt sync` now ends a partially failed run with a concise summary that names each database that failed and why. Healthy databases still sync, while operators and failed Order Events can distinguish a one-database failure from a total outage without searching supervisor logs.

For example, a run where one of three pushes fails ends with `sync: 1/3 database(s) failed: bad (push failed (exit 1));`.

## Review notes

- Failure aggregation is keyed to each database sync call's return code; the reason string is diagnostic data, not the success signal.
- The summary is written last so it remains inside the bounded output tail captured for a failed Order Event.
- Successful runs keep their existing output. Remote selection, divergence handling, and push behavior are unchanged.

## Test plan

- [x] Run the full local CI-equivalent suite with the rootless container runtime enabled; confirm both modified sync tests execute and pass by name.
- [x] Verify a three-database fixture continues both healthy pushes, returns non-zero, and names only the failed database and reason in the final summary.
- [x] Run build, vet, CI policy, module-boundary, documentation, native-DoltLite, and changed-format checks.
- [x] Release gate: [`release-gates/ga-g514cl-dolt-sync-failure-summary-gate.md`](release-gates/ga-g514cl-dolt-sync-failure-summary-gate.md)

### DIFF
--- a/examples/bd/dolt/commands/sync/run.sh
+++ b/examples/bd/dolt/commands/sync/run.sh
@@ -651,11 +651,9 @@ if [ -d "$data_dir" ]; then
   done
 fi
 
-# D2: the loop above uses a sticky aggregate exit_code with no record of which
-# database(s) failed or why, which reads to a caller/agent as if ALL databases
-# failed even when only one did. Name the actual failures explicitly, as the
-# last line emitted (it must survive tailForOrderFailureEvent's truncation
-# window in cmd/gc/order_dispatch.go when this runs under an order).
+# Positioned as the last line of output: an OrderFailed event built from this
+# script's output (tailForOrderFailureEvent, cmd/gc/order_dispatch.go) keeps
+# only a bounded tail, so this summary must survive that truncation window.
 if [ "$exit_code" -ne 0 ]; then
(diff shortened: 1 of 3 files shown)


quad341:
@quad341 — this kills a genuinely annoying bug: a sticky aggregate `exit_code` that says "something failed" and then makes the operator go spelunking through supervisor logs to find out *which* database, at exactly the moment they least want to be spelunking.

Two things I appreciated. First, keying the aggregation to the call's return code and treating `last_fail_reason` as pure diagnostic data — the `${last_fail_reason:-unknown error}` fallback means a missed assignment degrades to a vague message instead of a silently swallowed failure. That's the right direction to fail in. Second, the comment explaining *why* the summary is positioned last. I went and checked rather than taking it on trust: `tailForOrderFailureEvent` keeps the last 2048 bytes, and `shellExecRunner` hands stdout and stderr the same buffer, so the ordering you're relying on is real.

The test earns its keep too — asserting `1/3` *and* asserting the output must not read `3/3` is the assertion that would actually catch a regression back into the old behavior.

Merging as-is. Nothing to change on your side.


---

PR #5624, opened 2026-08-25
Fix Dolt compact notification cadence

quad341:
## What this changes

Dolt compaction alerts now notify on meaningful state transitions and on a configurable long backstop, instead of applying opposite policies to different marker types. An unchanged stale pending-push or pending-GC marker no longer mails on every compact cycle, while an unresolved quarantine no longer goes silent forever after its first notification.

Repeated notifications include the marker's sighting count and age so operators can tell how long the condition has persisted. The backstop defaults to 24 hours and can be changed with `GC_DOLT_COMPACT_RENOTIFY_BACKSTOP_SECS`.

The change does not clear protective markers or suppress their first alert. A changed reason still notifies immediately, and failed mail delivery remains retryable on the next cycle.

## Review notes

- The quarantine, pending-push, and pending-GC paths now share marker notification-state bookkeeping.
- Marker-state writes remain atomic, and invalid backstop configuration fails through the existing numeric-validation path.
- Open PR #5493 changes nearby quarantine notification functions for delivery-error visibility. The behaviors are complementary, but whichever lands second will likely need a small textual reconciliation.

## Test plan

- [x] Run the documented 40-job `make test-local-full-parallel` union and confirm both new cadence tests plus the existing compact notification/no-sync regression suite execute and pass.
- [x] Run `make test-ci-policy`, `go vet ./...`, affected lint, formatting, shell syntax, and diff checks.
- [x] Release gate: [`release-gates/ga-6dgrtc-dolt-compact-notification-cadence-gate.md`](release-gates/ga-6dgrtc-dolt-compact-notification-cadence-gate.md)

---
- Related to #2348 — touches the same compact quarantine path: an unresolved quarantine or stale pending-push marker now re-alerts on a configurable backstop (default 24h) with sighting count and age instead of going quiet after the first mail, which shortens how long a quarantined rig can sit unnoticed; it does not change the concurrent-write row-count race that creates the false quarantine, and does not add gc doctor surfacing
- Related to #3341 — touches the same dolt compact quarantine-marker path and changes its alert cadence so an unresolved quarantine keeps re-alerting on a configurable backstop instead of going quiet after the first mail; it does not address the write-race that causes the quarantine, does not clear markers or restore GC, and leaves the backup-while-quarantined behavior unchanged

### DIFF
--- a/examples/bd/dolt/commands/compact/run.sh
+++ b/examples/bd/dolt/commands/compact/run.sh
@@ -85,6 +85,15 @@
 #   GC_DOLT_COMPACT_PENDING_PUSH_MAX_AGE_SECS
 #     (default: 172800) — maximum age for automatic pending remote-push retry.
 #                       Older markers require manual review before push.
+#   GC_DOLT_COMPACT_RENOTIFY_BACKSTOP_SECS
+#     (default: 86400) — once a quarantine, pending-push, or pending-gc
+#                       marker has mailed an alert for an unchanged reason,
+#                       suppress repeat mail until this many seconds have
+#                       elapsed since the last notification, then send
+#                       exactly one more. Set to 0 to restore
+#                       notify-once-per-reason-forever (the marker never
+#                       re-mails on its own once a given reason has been
+#                       reported, even if left unresolved indefinitely).
 #   GC_DOLT_COMPACT_REMOTE               (optional) — remote to fetch/push.
 #                                         Defaults to origin when present;
 #                                         ambiguous multi-remote stores fail.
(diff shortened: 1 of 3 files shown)


quad341:
@quad341 — the diagnosis here is the good part. Naming both failure modes as *one* bug is what makes the fix small: the old code applied opposite policies to marker types carrying identical state, so a pending-push marker screamed 40 times about one unchanged condition while a real quarantine went quiet forever after a single page. Collapsing `quarantine_should_notify`/`record_quarantine_notify_state` into dir-parameterized `marker_should_notify`/`record_marker_notify_state` is the right shape, and pulling `parse_compact_timestamp` out so the GNU/BSD fallback is shared rather than duplicated was a clean touch.

I especially like that both new tests assert the *event* count separately from the *mail* count. That's the invariant that keeps automation observing every cycle while operators stop getting paged, and it would have been easy to leave untested. The `seen=`/`age=` context on repeats is the part operators will actually feel.

Merging as-is. Two things I'll open follow-ups for rather than hold this: the backstop env var has no test of its own (both tests reach it by backdating `last_notified_ts` under the default), and a persistently-dead mail path re-attempts per cycle once the backstop lapses. Neither blocks. We'll handle the #5493 text reconciliation on our side whichever way it lands — that's ours, not yours.


---

PR #5626, opened 2026-08-25
fix(storeref): refuse the by-id read a migrated-then-refused city would answer from the frozen twin

julianknutsen:
`gc beads show <id>` on a city that migrated its infrastructure classes and then stopped being able to serve the binding exits **0** with the pre-migration copy. `gc storage migrate` preserves ids and deletes nothing, so the work ledger still holds a frozen twin of every relocated bead; the residency resolver tolerates the city's standing storage refusal on a residence probe — a refused city still serves WORK — and the surface falls through to its own directory scan, which finds that twin and reports it as current. No error, no diagnostic, and the close that follows writes it. That is ga-q8ick.

This branch withdraws the toleration for exactly one shape of city: one whose binding is **proven** to still hold ids outside the namespaces it declares.

## What changed since the last review

The previously-approved version fed the proof from a durable note under the city's own `.gc/` (`.gc/storage-relic-census.json`, produced by s8). **That status file is gone**, and the design changed with it: the proof is now **computed live at by-id time**, on a refused city only.

Two reasons the note was the wrong shape. It is a status file, which this codebase does not keep (`AGENTS.md`: *no status files — query live state*), and it goes stale in the **denying** direction — relics close over the weeks after a migration, so a note written in March denies reads in June for a binding that has been clean since April.

The premise it existed to work around is false anyway. **"Refused" is a verdict about SERVING a binding**, not about reading one: a convergence check, a served-binding note, a discipline the boot gate enforces. Almost none of those say the binding is unreadable. So the census runs against a handle opened for the read and closed after it, and its answer is about the binding as it is now.

- `storeref.ProvenLegacyResidents` is `HasOpenLegacyResidents` **with the default inverted**, and the inversion is why there are two. That one decides whether to KEEP a probe, so an unreadable binding answers `true`. This one decides whether to DENY a read, so the same binding answers `false`: a census that could not run has proved nothing.
- `byIDResidencyTopology` gates the whole thing on `Topology.Refused`. A served city's bindings were censused live when the funnel opened them, so a second read could prove nothing the first did not — and taking one would put an engine open on the by-id path of every converged city.
- The refused city is **re-derived** through `residencyBindingsFromRoutesWithProof` rather than patched in place, so both relic bits and the ref they are keyed by come from the one derivation.
- Every way of not reaching an answer — a config that will not load, a plan that will not resolve, a provider that opens no engine, an **open that fails** — is proof-absent and falls through exactly as today.

### Can the proof re-derive over a controller's registered routes?

No — and it is now structural rather than incidental. A refused city aborts controller construction (`city_runtime.go`: `storageBootGate` error ⇒ `return nil, err`), so registered routes are never refusing and the `Topology.Refused != nil` gate provably excludes them. Relying on an invariant three files away is still one edit from being wrong, so the re-derivation goes through `residencyRoutesForCity`, which returns a running controller's registered routes when there are any and only otherwise the one-shot funnel. Resolving the funnel for a city a controller registered is a second handle on the binding root — a duplicate managed-Dolt server or a second sqlite writer — and that is the reason the registration exists at all.

## Known residual (ga-q8ick follow-up)

The census this gates on is **main's open-only** one (`storeref.OpenLegacyResidents`). s5's closed-inclusive census is being re-cut in parallel and is not on this base, so **a binding whose only relics are CLOSED still reads as unproven and its frozen twin is still served**. Closing that is a one-line swap, deliberately structured as one: `ProvenLegacyResidents` is the only caller that names the census, and its doc comment says so.

## Commits

Rebased onto `origin/main` @ `fc7a23a292`. All eleven of the previous head's contributor commits are carried with `-x`; the SHAs are new. s8's own commits are dropped — they are in the `fc7a23a292` squash.

| | commit | carried from |
|---|---|---|
| 1 | `7131d9d4fa` refactor(gc): split the by-id residency seam at the topology line | `cfa2b5634a` |
| 2 | `a428a9fc5f` refactor(storeref): give binding-only by-id resolution its own executor | `5563f05baf` |
| 3 | `bbaeede0f9` fix(storeref): refuse a refused city's by-id read when its binding is proven to hold relics | `5bafebae1a` |
| 4 | `49c794b381` test(storeref): pin the relic-bit implication and make the corpus reachable | `fad9f80099` |
| 5 | `b49b8caf42` fix(storeref): stop binding-only resolution at every leg of the caller's axis | `eede08964d` |
| 6 | `6d59a6f44c` fix(storeref): say why a proven-relic by-id denial refused, and pin the note | `02ff37960b` |
| 7 | `329229ba3a` docs(residency): correct the refusal-tolerance comments and name the executor | `fab32855b9` |
| 8 | `f9b875a2eb` fix(gc): prove a refused city's relics at by-id time instead of from a note | `8afc3608fe` + the baseline row rename |
| 9 | `39e12a773c` test(gc): compose the relic denial from the census that produced the proof | `36753e4f0a` |
| 10 | `d56af32ca0` test(beads): pin the relic denial at the `gc beads show` surface itself | `7df2f95c67` |
| 11 | `0b022a3618` fix(gc): never let the relic proof create the binding it is asked about | `4a2dea5ad4` |
| 12 | `f37b86a520` fix(residency): rename the baseline row #5597 renamed the function under | *new — base-owned, droppable; see below* |

**`e23cc60ea9`'s rename is now needed, and is folded into commit 8.** On the old base the guard **allowlisted `cmd/gc/residency_topology.go` wholesale**, so the file carried no baseline row and that commit was a no-op. s4 #5643 (`7746bbe5ab`) closed that evasion and pinned the file's hits as rows, so on `main` the row exists and commit 8's `residencyBindingsFromRoutes` → `residencyBindingsFromRoutesWithProof` rename moves it. Regenerated by the baseline header's documented three-emitter splice (`--emit-baseline`, `RESIDENCY_BASELINE_EMIT=1 go test ./scripts -run TestResidencyResolverBoundary`, `… -run TestResidencyASTExpressionRatchet`, concatenated and `LC_ALL=C sort`ed, header kept). **114 rows before, 114 after** — a rename, not growth. The header prose names neither renamed function, and the "six rows, six hits" block is still six.

## Review reconciliations applied

- **ga-e3dvx** — the denial arm keys on `leg.OnError == PolicyFatal`, not only on `Role`. It is also **ordered above** the tolerated arm, because below it the policy clause was unreachable and the key would have been unfalsifiable. It is load-bearing now (see mutations).
- **The remedy is one line.** Callers print `gc <cmd>: %v`, so the original's embedded `\n` dropped the command prefix off the half that named the remedy. `withProvenRelicRemedy` (renamed from `withRelicNoteRemedy`, which named a note that no longer exists) appends the remedy as a **second sentence** on one line, and both the value and the printed output are pinned.
- **`TestBeadsShowRefusesTheRelicItWouldOtherwiseServeFrozen`** and **`TestRefusedCityDeniesTheRelicIts…`** are kept, re-pointed at the live proof. The frozen twin is seeded for real; the proof is computed, never written.
- Doc comments on every exported symbol.

## RED → GREEN

Written red first, against commit 7:

```
=== RUN   TestRefusedCityDeniesTheRelicItsLiveCensusProves
    by_id_relic_proof_test.go:79: a refused city whose binding holds gc-relic1
    resolved to ok=false with no error; the caller falls through to its scan and
    serves the copy the migration retained in the work store
--- FAIL: TestRefusedCityDeniesTheRelicItsLiveCensusProves (0.06s)
--- PASS: TestRefusedCityThatCannotOpenItsBindingFallsThrough (0.03s)
--- PASS: TestServedCityPaysNothingForTheRelicProof (0.10s)
```

That is the ga-q8ick symptom verbatim. The other two are controls that must hold across the change. After commit 8:

```
--- PASS: TestRefusedCityDeniesTheRelicItsLiveCensusProves (0.06s)
--- PASS: TestRefusedCityThatCannotOpenItsBindingFallsThrough (0.02s)
--- PASS: TestServedCityPaysNothingForTheRelicProof (0.12s)
```

## Mutation probes

| mutation | dies |
|---|---|
| drop `leg.OnError == PolicyFatal` from `resolveByID`'s denial arm | `TestStandingRefusalIsToleratedOnlyOutsideTheReservedNamespace`, `TestBindingOwnerSurfacesTheRefusalOnAProvenRelicBinding`, `TestBindingOwnerRefusedCityWithoutProofStillDeclines`, and both new controls |
| `probeRefusalPolicy` forced to `PolicyRefusalTolerated` | `TestRefusedCityDeniesTheRelicItsLiveCensusProves` |
| drop the denial arm entirely | `TestRefusedCityDeniesTheRelicItsLiveCensusProves` (loses the sentinel and the remedy) |
| `ProvenLegacyResidents` returns `true` on a census error | `TestProvenLegacyResidentsFallsTheOTHERWay` — the CLI rows cannot reach it, because a binding that cannot be *opened* never gets as far as a `List` |
| the proof keys on a narrowed ref (`ClassRef(b.Classes[:1])`) | `TestTheProofAndTheRefusedFunnelSpellTheBindingRefTheSameWay` **and** the denial row — but only the first names the reason |
| `doBeadsShowFallback` reads every error as fall-through | `TestBeadsShowRefusesTheRelicItWouldOtherwiseServeFrozen` **only** — the frame below stays green, which is why that row exists |

## Gates

| command | exit |
|---|---|
| `go build ./...` | 0 |
| `go vet ./...` | 0 |
| `gofmt -l cmd internal scripts` | 0 (nothing printed) |
| `golangci-lint run ./cmd/gc/... ./internal/storeref/... ./internal/storebinding/...` | 0 — `0 issues.` |
| `go test ./internal/storeref/... ./internal/storebinding/... -count=1` | 0 |
| `go test ./cmd/gc -run 'ByID\|Residency\|Relic\|BeadsShow\|Refused' -count=1 -v` | 0 — **82** `--- PASS`, 0 `--- FAIL` |
| `make check-docs` | 0 |
| `make test-fast-parallel` | 0 |

Refs ga-q8ick.

---

## Review follow-up: the proof must not create the binding it is asked about

`4a2dea5ad4`. `beads.OpenSQLiteStore` **creates** the database when it is absent, so opening the binding as a probe made it answer its own question — on a binding that was not there it would create one, census it, find nothing, and report clean. The city this happens to is the **never-migrated** one (a `[storage]` split configured and never cut over), which is the most common refused city there is, and on it an ordinary by-id read reached this path. The damage is what it leaves behind: an empty binding root that a later boot reads as a genesis root on a city that never cut over, and on a workspace-backed provider a managed engine started for a *read*.

`infraBindingHoldsNothing` already argues the rule for the migration's own no-open probe — *"a probe that creates the database it is asked about would be answering its own question"* — so existence is now checked **before** the open, through the same two helpers (`infraBindingRootEnumerable`, `infraPathExists`) rather than a second spelling of them. A binding that is not already on disk is proof-absent and falls through exactly as an unreadable one does.

`resolveInfraBindingTarget` answers only for the built-in bead engine, so a binding served by any other provider takes a second branch: the **provider** is asked where it serves from (`servedBindingLocation`) and that location has to exist. A location that is remote, opaque or simply absent is proof-absent — the safe direction.

**RED, both arms.** Against `7df2f95c67`:

```
--- FAIL: TestTheRelicProofNeverCreatesTheBindingItIsAskedAbout
    the relic proof CREATED the binding database
    .../never-migrated/graph/beads.sqlite as the side effect of a by-id read
    (stat err = <nil>); the next boot then finds a genesis root on a city that
    never cut over
    the relic proof CREATED the binding root .../never-migrated as the side
    effect of a by-id read (stat err = <nil>)
```

The second arm — **root present and empty, database absent** (a binding volume mounted but never written, or a genesis root rolled back) — is what makes the *database* check load-bearing, because the root check passes there. GREEN after the fix, with `TestRefusedCityDeniesTheRelicItsLiveCensusProves` (database present ⇒ still denies) and the other four controls unchanged.

**Mutation.** Dropping the database-exists check kills the second arm on the created file. The root check is deliberately **not** independently falsifiable and the doc comment says so: every way it can fail also makes the database stat answer absent-or-error, and both decline. It is kept so this precondition reads the same as the migration probe's, which is where the rule is argued — not because it is a second condition.

Gates re-run at `4a2dea5ad4`, all exit 0: `go build ./...`, `go vet ./...`, `gofmt -l cmd internal scripts` (nothing printed), `golangci-lint` (`0 issues.`), `./scripts/check-residency-boundary.sh`, `go test ./internal/storeref/... ./internal/storebinding/... -count=1`, targeted `cmd/gc` run **83** `--- PASS` / 0 FAIL, `make check-docs`, `make test-fast-parallel` ("All fast jobs passed").

---

## Rebase onto `main`

`f37b86a520`. Retargeted from `residency/storeref-series` to `main` now that s8 #5597 has merged as the squash `fc7a23a292` and s4 #5643 as `7746bbe5ab`. Branched from `origin/main` and cherry-picked **only this PR's eleven** commits with `-x`; s8's twelve are dropped, because they are in that squash — and the squash also carries fix commits (`0165a56fd1`, `21fa4eb892`) that were never on the `f79429f461` head this branch was originally cut from.

**No textual conflicts.** All eleven applied clean. `cmd/gc/cmd_bd_by_id.go` did not conflict: s8's squash moved the relocated-class reads into the new `cmd/gc/bd_relocated_classes.go`, and this PR's by-id seam is confined to `cmd/gc/by_id_residency.go`. The one reconciliation the rebase did need is the residency boundary baseline, described under **Commits** — the rename that was a no-op on the old base and is load-bearing on this one.

**The design is carried unchanged.** Re-verified on the rebased tree, not assumed: `ResolveBindingOwner` is `resolveByID(p, id, true)` and stops at the caller's work axis **by role** (`ownedByCallersWorkAxis` = `RoleWorkFallback || RoleShadow`); `holdsLegacyResidents` is `HasLegacyResidents || KnownLegacyResidents`, so the known bit only ever **denies**, and `BuildBindings` raises the pessimistic bit on proof; `probeRefusalPolicy` returns `PolicyFatal` keyed on `KnownLegacyResidents`, and `resolveByID`'s denial arm keys on `leg.OnError == PolicyFatal` **ordered above** the tolerated arm; `ProvenLegacyResidents` returns **false** on a census error, failing the other way from `HasOpenLegacyResidents`; the live proof is computed on a refused city only (`topo.Refused == nil` ⇒ return), memoized per process and dropped with the residency memo, and gated by `refusedBindingIsAlreadyOnDisk` before any open; the T3k corpus rows are intact.

**The closed-inclusive swap did NOT happen — deliberately.** Checked on `main` rather than assumed: `internal/storeref` exports only the open-only census (`OpenLegacyResidents` / `HasOpenLegacyResidents`, with `TestOpenLegacyResidentsIgnoresAClosedRelic` pinning the exclusion). There is no `LegacyResidents` / `HasLegacyResidents` census function — s5's closed-inclusive verdict is still not on `main`. So the open-only gate stands and **the "Known residual" section above is still accurate and is kept**. The swap remains the one line its doc comment advertises.

### Base-owned red, disclosed

`./scripts/check-residency-boundary.sh` **exits 1 on `origin/main` @ `fc7a23a292` itself**, before this branch touches anything. Proven on an untouched export of `origin/main` (`git archive origin/main` into a scratch dir, run there):

```
RESIDENCY-BOUNDARY: cmd/gc/infra_class_migrate.go infraBindingIDPrefix d:ReservedClassPrefix: 1 sites, baseline 0 — a NEW store-enumeration site.
RESIDENCY-BOUNDARY: cmd/gc/infra_class_migrate.go openInfraDestination d:ReservedClassPrefix: 0 sites, baseline 1 — the baseline must SHRINK …
```

That is **ga-kkcby**, and a one-row fix PR for `main` is in flight. It is not caused by this change and is not this branch's to own — but the baseline is one file with one ratchet, and the documented regeneration emits current reality, so it cannot be regenerated around while re-keying this branch's own row. It is therefore carried as commit 12, **last and alone**, and it **drops out as an empty rebase step** once the main fix merges: both sides re-key the row identically. Nothing else in the branch depends on it.

### Mutation probes re-run on the rebased tree

| mutation | dies |
|---|---|
| drop `leg.OnError == PolicyFatal` from `resolveByID`'s denial arm | `TestStandingRefusalIsToleratedOnlyOutsideTheReservedNamespace`, `TestBindingOwnerSurfacesTheRefusalOnAProvenRelicBinding`, `TestBindingOwnerRefusedCityWithoutProofStillDeclines`, `TestRefusedCityDeniesTheRelicItsLiveCensusProves`, `TestRefusedCityThatCannotOpenItsBindingFallsThrough`, `TestBeadsShowRefusesTheRelicItWouldOtherwiseServeFrozen`, `TestTheRelicProofNeverCreatesTheBindingItIsAskedAbout` — **and two rows this base adds**: `TestClassRoutedStoreForIDKeepsWorkOnARefusedCity`, `TestControllerResidencyTopologyCarriesTheStandingRefusal` |
| drop the `refusedBindingIsAlreadyOnDisk` existence gate | `TestTheRelicProofNeverCreatesTheBindingItIsAskedAbout` **only**, on all three of its arms (database created; root created; database created inside an existing empty root) |

Both restored; the tree is byte-clean after each.

### Gates at `f37b86a520`

| command | exit |
|---|---|
| `go build ./...` | 0 |
| `go vet ./...` | 0 |
| `gofmt -l cmd internal scripts` | 0 (nothing printed) |
| `golangci-lint run ./cmd/gc/... ./internal/storeref/... ./internal/storebinding/...` | 0 — `0 issues.` |
| `go test ./scripts -count=1` (the guard's AST half) | 0 |
| `go test ./internal/storeref/... ./internal/storebinding/... -count=1` | 0 |
| `go test ./cmd/gc -run 'ByID\|Residency\|Relic\|BeadsShow\|Refused' -count=1 -v` | 0 — **86** `--- PASS`, 0 `--- FAIL`, 0 skipped (82 → 83 → 86; this base adds rows). All seven `by_id_relic_proof_test.go` rows present and passing. |
| `make check-docs` | 0 |
| `make test-fast-parallel` | 0 — "All fast jobs passed" |

Refs ga-q8ick, ga-kkcby.

### DIFF
--- a/cmd/gc/by_id_binding_owner_test.go
+++ b/cmd/gc/by_id_binding_owner_test.go
@@ -373,6 +373,21 @@ func TestBindingOwnerRefusedCityWithKnownRelicsRefuses(t *testing.T) {
 	if !storeref.IsStandingRefusal(err) {
 		t.Errorf("the refusal came back as %v, want the standing storage refusal that names the remedy", err)
 	}
+	// The refusal's own sentence is the boot gate's, and a city takes the
+	// identical one for an infrastructure-class id it simply cannot serve. What
+	// makes this denial actionable is the reason behind it and the move that
+	// clears it, and neither is in that sentence.
+	if !errors.Is(err, storeref.ErrProvenRelicRefusal) {
+		t.Errorf("the denial reads %q and carries nothing that tells it apart from the refusal an in-namespace id takes; the operator goes looking for a missing bead", err)
+	}
+	if !strings.Contains(err.Error(), "gc doctor") {
+		t.Errorf("the denial reads %q with no route back to a served city", err)
+	}
+	// One line, because every caller prints this as `gc <cmd>: %v` and a second
+	// line arrives without the command prefix that says which surface refused.
+	if strings.Contains(err.Error(), "\n") {
+		t.Errorf("the denial is multi-line (%q); callers print it as `gc <cmd>: %%v`, so everything after the newline loses the command that refused", err)
+	}
 }
 
 // TestBindingOwnerRefusedCityWithoutMemoStillDeclines is the control, and it
--- a/cmd/gc/by_id_binding_owner_test.go
+++ b/cmd/gc/by_id_binding_owner_test.go
@@ -339,10 +339,12 @@ func TestBindingOwnerLeavesTheWorkResidualUnprobed(t *testing.T) {
 
 // refusedRelicTopology is a city whose boot could not serve its configured
 // split: every infrastructure class resolves to a refusing store and the
-// standing refusal rides on the topology. proven says whether a durable census
-// has PROVED that binding holds ids outside its reserved namespaces — the one
-// fact about a refused binding that is readable without reading the binding,
-// because it is a city file rather than a store.
+// standing refusal rides on the topology. proven says whether a census has
+// PROVED that binding holds ids outside its reserved namespaces.
+//
+// It is assembled by hand rather than resolved from a city, so these two rows
+// stay about the RESOLVER's verdict alone. Where the proof comes from is
+// by_id_relic_proof_test.go's subject.
 func refusedRelicTopology(proven bool) storeref.Topology {
 	refusal := standingStorageRefusal{err: errors.New("storage refused: this city has not converged on its configured [storage] binding; run `gc storage migrate`")}
 	classes := infrastructureClasses()
@@ -351,7 +353,7 @@ func refusedRelicTopology(proven bool) storeref.Topology {
 		Prefixes: storeref.ReservedPrefixesFor(classes),
 		Leg:      storeref.Leg{Ref: storeref.ClassRef(classes), Store: refusedClassStore{err: refusal}},
 		// A refused store declares no mint namespace, so the pessimistic bit
-		// stands whatever the memo says.
+		// stands whatever the proof says.
 		HasLegacyResidents:   true,
 		KnownLegacyResidents: proven,
 	}}, refusal)
@@ -390,17 +392,17 @@ func TestBindingOwnerRefusedCityWithKnownRelicsRefuses(t *testing.T) {
 	}
 }
 
-// TestBindingOwnerRefusedCityWithoutMemoStillDeclines is the control, and it
+// TestBindingOwnerRefusedCityWithoutProofStillDeclines is the control, and it
 // must pass on both sides of the fix.
 //
-// The memo is TRUE-only: a ref absent from it means "not known", never "known
-// clean". So absence must not deny anything — a city that was never migrated,
-// or whose note was lost, keeps today's behavior, and work is still served from
-// the ledger work never left.
-func TestBindingOwnerRefusedCityWithoutMemoStillDeclines(t *testing.T) {
+// The proof is TRUE-only: a binding it says nothing about is "not known", never
+// "known clean". So absence must not deny anything — a city that was never
+// migrated, or whose binding could not be read, keeps today's behavior, and
+// work is still served from the ledger work never left.
+func TestBindingOwnerRefusedCityWithoutProofStillDeclines(t *testing.T) {
 	owner, ok, err := byIDBindingOwnerForTopology(refusedRelicTopology(false), "gc-1")
 	if err != nil {
-		t.Fatalf("a refused city with no relic evidence resolved to err=%v; the memo's absence is not evidence, and denying here takes work-bead reads away from every unconverged city", err)
+		t.Fatalf("a refused city with no relic evidence resolved to err=%v; absence of proof is not evidence, and denying here takes work-bead reads away from every unconverged city", err)
 	}
 	if ok {
 		t.Errorf("a refusing binding reported ownership of gc-1 (%p)", owner.Store)
--- a/cmd/gc/by_id_binding_owner_test.go
+++ b/cmd/gc/by_id_binding_owner_test.go
@@ -280,26 +280,39 @@ func TestBeadsShowFallbackScansForAnIdNoBindingHolds(t *testing.T) {
 	}
 }
 
-// TestBindingOwnerLeavesTheWorkResidualUnprobed is the sentinel's contract.
+// TestBindingOwnerLeavesTheWorkResidualUnprobed is the delegation pin: this
+// arm's work axis is a directory scan the resolver must not run, and
+// storeref.ResolveBindingOwner is what guarantees it does not.
 //
-// cliByIDBindingOwner hands the plan a placeholder where the work leg goes,
-// because this arm's work axis is a directory scan the resolver must not run.
-// That is safe only while the residual is returned UNPROBED, so the placeholder
-// reports being read as an internal error — and a clean ok=false here is the
-// proof it was not.
+// The measurement is a COUNT rather than a sentinel's refusal, because the
+// dangerous case is not a work store that errors — it is one that answers. The
+// city store still holds the copy `gc storage migrate` retained, so a work leg
+// that got probed here would report a real, stale bead as a binding owner and
+// the caller's fall-through would never run. Zero Gets is the only shape that
+// rules that out.
 func TestBindingOwnerLeavesTheWorkResidualUnprobed(t *testing.T) {
-	cityPath := oneShotCLICity(t, "")
-	refuseInfraMigrationSource(t)
-	captureCLIStorageStderr(t)
-
-	owner, ok, err := cliByIDBindingOwner(cityPath, "gc-1")
+	cityPath, _ := foreignProviderCity(t)
+	counted := &countingGetStore{Store: splittest.NewWorkStore(t, "hq")}
+	held, err := counted.Create(beads.Bead{Title: "a bead only the work axis holds", Type: "task"})
 	if err != nil {
-		t.Fatalf("a city that relocates nothing resolved to err=%v; the residual must come back unprobed, not read", err)
+		t.Fatalf("seeding the work store: %v", err)
+	}
+	counted.gets = 0
+
+	owner, ok, err := byIDBindingOwnerForTopology(cliResidencyTopology(cityPath, nil, counted, nil), held.ID)
+	if err != nil {
+		t.Fatalf("an id no binding holds resolved to err=%v; a clean miss is ok=false, not a failure", err)
 	}
 	if ok {
-		t.Errorf("a city that relocates nothing reported a binding owner %p", owner.Store)
+		t.Errorf("the work axis's own copy of %s came back as a binding owner (%p); the caller would never run its scan", held.ID, owner.Store)
+	}
+	if counted.gets != 0 {
+		t.Errorf("the work leg was probed %d time(s), want 0 — the retained pre-migration copy lives there, so a probe answers with a stale bead rather than a miss", counted.gets)
 	}
 
+	// The placeholder both call sites hand the plan is defense in depth for the
+	// same guarantee: if the executor ever does reach the work leg, it must be a
+	// loud error rather than a plausible answer.
 	residual := newUnprobedWorkResidual()
 	got, err := residual.Get("gc-1")
 	if err == nil {
--- a/cmd/gc/by_id_binding_owner_test.go
+++ b/cmd/gc/by_id_binding_owner_test.go
@@ -337,6 +337,61 @@ func TestBindingOwnerLeavesTheWorkResidualUnprobed(t *testing.T) {
 	}
 }
 
+// refusedRelicTopology is a city whose boot could not serve its configured
+// split: every infrastructure class resolves to a refusing store and the
+// standing refusal rides on the topology. proven says whether a durable census
+// has PROVED that binding holds ids outside its reserved namespaces — the one
+// fact about a refused binding that is readable without reading the binding,
+// because it is a city file rather than a store.
+func refusedRelicTopology(proven bool) storeref.Topology {
+	refusal := standingStorageRefusal{err: errors.New("storage refused: this city has not converged on its configured [storage] binding; run `gc storage migrate`")}
+	classes := infrastructureClasses()
+	return assembleResidencyTopology(nil, newUnprobedWorkResidual(), nil, []storeref.ClassBinding{{
+		Classes:  classes,
+		Prefixes: storeref.ReservedPrefixesFor(classes),
+		Leg:      storeref.Leg{Ref: storeref.ClassRef(classes), Store: refusedClassStore{err: refusal}},
+		// A refused store declares no mint namespace, so the pessimistic bit
+		// stands whatever the memo says.
+		HasLegacyResidents:   true,
+		KnownLegacyResidents: proven,
+	}}, refusal)
+}
+
+// TestBindingOwnerRefusedCityWithKnownRelicsRefuses is the ga-q8ick bug.
+//
+// A refused city still serves WORK, so the resolver tolerates the refusal on a
+// residence probe and the surface falls through to its own scan. That is right
+// until the binding is PROVEN to hold work-prefixed relics: `gc storage migrate`
+// preserved those ids and deleted nothing, so the scan finds the retained
+// pre-migration copy in the city work store, serves it, and the close that
+// follows writes it. Exit 0, no diagnostic.
+func TestBindingOwnerRefusedCityWithKnownRelicsRefuses(t *testing.T) {
+	_, ok, err := byIDBindingOwnerForTopology(refusedRelicTopology(true), "gc-1")
+	if err == nil {
+		t.Fatalf("a refused city proven to hold work-prefixed relics resolved to ok=%v with no error; the caller's scan then serves the copy the migration left in the work store", ok)
+	}
+	if !storeref.IsStandingRefusal(err) {
+		t.Errorf("the refusal came back as %v, want the standing storage refusal that names the remedy", err)
+	}
+}
+
+// TestBindingOwnerRefusedCityWithoutMemoStillDeclines is the control, and it
+// must pass on both sides of the fix.
+//
+// The memo is TRUE-only: a ref absent from it means "not known", never "known
+// clean". So absence must not deny anything — a city that was never migrated,
+// or whose note was lost, keeps today's behavior, and work is still served from
+// the ledger work never left.
+func TestBindingOwnerRefusedCityWithoutMemoStillDeclines(t *testing.T) {
+	owner, ok, err := byIDBindingOwnerForTopology(refusedRelicTopology(false), "gc-1")
+	if err != nil {
+		t.Fatalf("a refused city with no relic evidence resolved to err=%v; the memo's absence is not evidence, and denying here takes work-bead reads away from every unconverged city", err)
+	}
+	if ok {
+		t.Errorf("a refusing binding reported ownership of gc-1 (%p)", owner.Store)
+	}
+}
+
 // TestConvoyResolutionStillRefusesAnIdTwoLedgersBothHold covers the rule the
 // binding short-circuit steps around, which until now nothing asserted anywhere
 // in the tree.
--- a/cmd/gc/by_id_relic_proof.go
+++ b/cmd/gc/by_id_relic_proof.go
@@ -1,0 +1,192 @@
+package main
+
+// The by-id door's relic proof: computed, never recorded.
+//
+// A city whose boot REFUSED still serves WORK from its work ledger, so the
+// residency resolver tolerates the standing storage refusal on a residence
+// probe and the surface falls through to its own axis. The sentence that
+// justifies that is "this leg was only ever a probe for an id no relocated
+// class could own", and there is exactly one shape of city it is false for: one
+// whose binding still holds ids `gc storage migrate` carried across under their
+// original work-shaped names. There, falling through does not land on "no
+// answer" — it lands on the frozen pre-migration copy the migration left in the
+// work store, which answers confidently and wrongly, and the close that follows
+// writes it. That is ga-q8ick.
+//
+// # Why the proof is taken here rather than read off disk
+//
+// The obvious place to keep this verdict is a note under the city's own .gc,
+// written by whichever process last managed to read the binding. It is the
+// wrong place twice over. It is a status file, which this codebase does not
+// keep (AGENTS.md: no status files — query live state), and it goes stale in
+// the DENYING direction: relics close over the weeks after a migration, so a
+// note written in March denies reads in June for a binding that has been clean
+// since April.
+//
+// The premise the note was there to work around turns out to be false anyway.
+// "Refused" is a verdict about SERVING the binding — a convergence check, a
+// served-binding note, a discipline the boot gate enforces — and almost none of
+// those verdicts say the binding cannot be READ. So the census that decides
+// this runs right here, against a handle opened for the read and closed after
+// it, and its answer is about the binding as it is now.
+//
+// # What it costs, and who pays
+//
+// Only a refused city reaches any of it. residencyTopologyForCity has already
+// answered by the time the gate below is consulted, and a served city's
+// Topology.Refused is nil, so it resolves no plan, opens no engine and takes no
+// read — TestServedCityPaysNothingForTheRelicProof counts the plan resolutions
+// and requires zero. A refused city pays one engine open and one full list of
+// the binding, once per process per city, on the by-id path only. That city is
+// already in the incident state its own boot gate reported.
+//
+// # The absent case is the tolerant one
+//
+// Every way of not reaching an answer — a config that will not load, a plan
+// that will not resolve, a provider that opens no engine, an open that fails —
+// is proof-ABSENT, and proof-absent falls through exactly as today. The bit
+// only ever denies, so its unknown must be false: a binding nobody could read
+// has proved nothing, and denying on it would take work-bead reads away from
+// every city whose binding is merely unreachable.
+
+import (
+	"path/filepath"
+	"sync"
+
+	"github.com/gastownhall/gascity/internal/beads"
+	"github.com/gastownhall/gascity/internal/config"
+	"github.com/gastownhall/gascity/internal/fsys"
+	"github.com/gastownhall/gascity/internal/storeref"
+)
+
+// byIDResidencyTopology is the topology the by-id door plans over: the city's
+// own residency topology, plus — on a REFUSED city only — the proof that the
+// binding it cannot serve still holds ids the migration preserved.
+//
+// The refused city is re-derived rather than patched. A ClassBinding's two
+// relic bits have an implication between them that storeref.BuildBindings
+// enforces, and reaching into the assembled topology to raise one of them by
+// hand is how a plane ends up spelling a state no city can be in. Handing the
+// proof back through residencyBindingsFromRoutesWithProof means both bits and
+// the ref they are keyed by come from the one derivation.
+func byIDResidencyTopology(cityPath string, cfg *config.City, work beads.Store, rigs map[string]beads.Store) storeref.Topology {
+	topo := residencyTopologyForCity(cityPath, cfg, work, rigs)
+	if topo.Refused == nil || len(topo.Bindings) == 0 || cityPath == "" {
+		return topo
+	}
+	proven := provenRelicRefsForCity(cityPath)
+	if len(proven) == 0 {
+		return topo
+	}
+	bindings, refused := residencyBindingsFromRoutesWithProof(
+		residencyRoutesForCity(cityPath),
+		func(ref storeref.StoreRef) bool { return proven[ref] },
+	)
+	return assembleResidencyTopology(cfg, work, rigs, bindings, refused)
+}
+
+// provenRelicRefsForCity opens the binding this city is configured for, censuses
+// it, and returns the binding refs the census PROVED hold ids outside the
+// namespaces they declare.
+//
+// It is called only for a city whose boot refused, which is also what makes
+// opening the binding here safe: the refusal is why nothing else in this
+// process holds it open. A served city's binding is already open on the funnel,
+// and a second handle on a binding root — a duplicate managed-Dolt server, a
+// second sqlite writer — is the bug the residency constructors exist to avoid.
+//
+// The handle is closed before the verdict is returned. Nothing downstream reads
+// through it: what travels is a set of refs.
+//
+// An empty result is "nothing proved", which every failure path also produces
+// and which the caller reads as no evidence.
+func provenRelicRefsForCity(cityPath string) map[storeref.StoreRef]bool {
+	key := filepath.Clean(cityPath)
+	provenRelicRefsMu.Lock()
+	if provenRelicRefsByCity == nil {
+		provenRelicRefsByCity = make(map[string]map[storeref.StoreRef]bool, 1)
+	}
+	cached, ok := provenRelicRefsByCity[key]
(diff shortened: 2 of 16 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277237366

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277244097

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277244166

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277244235

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277244308

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR closes a real correctness/authority hole in store-ref by-id residency: a city that had been migrated (its beads copied into the work store's frozen twin) and then refused could still have a by-id read — `gc beads show` — quietly answered from that frozen twin via `RoleWorkFallback`, silently bypassing the refusal. That matters because a refused city is supposed to stop serving, and answering from the frozen twin let stale, out-of-namespace data leak back out under an otherwise-authoritative surface.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **A refused, migrated city could still serve its by-id read from the work store's frozen twin.** The fix gives binding-only by-id resolution its own executor and flips the residence probe to a fatal policy *only under durable proof* that the binding still holds relics, then denies exactly that read — so the frozen-twin fallback no longer bypasses a refusal, while the normal hot by-id path is left unperturbed.
- **The two refusal outcomes were textually indistinguishable to an operator.** A distinct `ErrProvenRelicRefusal` sentinel is now matched type-safely (double-`%w` wrap preserves the existing standing-refusal semantics), and the remedy text names the relic note and `gc doctor`, so a denial is diagnosable from its own message.
- **The dirty-binding census could retire a probe on a binding a durable record proves dirty.** "Proof implies pessimism" is now enforced both at the field and at the decision, so a binding proven to hold legacy residents can't have its probe retired out from under the guard.
- **Test evidence.** The added tests pin the denial (and a healthy converged control) at the real `gc beads show` entry point — `cmdBeadsShow` → `beadsShowAPIClient` → `doBeadsShowFallback` — rather than a shortcut, and discriminate the tolerated case from the denied one, denial from read-fault, and guard against an unfalsifiable test.

By-design trade-offs the reviewers noted for maintainer awareness (intentional, acknowledged in-code, not gating and not deferred work): protection is bounded to cities censused-dirty before they refused; the relic note is monotone (set-once, never cleared), which stays correct while frozen twins remain; and a corrupt/unreadable note fails open with a stderr report rather than turning an optimization into an outage.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `30810c6e` has passing required GitHub checks (preflight, cmd/gc process shards, integration suites, and `CI / required` are all green), and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/32918085190

### Review Iterations
1 review pass performed (Claude design+implementation and Codex implementation; the cross-file reviewer was skipped for this run). Both reviewers independently reached "no change required" — Claude after re-deriving the bug premise and auditing all five consumer call sites, Codex after confirming the new tests exercise the real entry point — so no fix iteration was needed. No non-blocking follow-ups.

---

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR makes the `gc beads show <id>`
by-id read refuse to answer from a migrated-then-refused city's pre-migration
frozen twin, instead of silently serving stale relic data. That closes a real
correctness hole (ga-q8ick): a refused city would fall back to the directory
scan and hand back a work-prefix relic by id, so a proven-relic by-id read now
fails closed rather than returning a wrong, frozen answer.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **By-id read served the pre-migration frozen twin on a migrated-then-refused
  city (ga-q8ick).** The fix withdraws refusal-tolerance only for a binding a
  durable relic census has *proven* holds migration-preserved ids
  (`KnownLegacyResidents=true` → `probeRefusalPolicy()` returns `PolicyFatal`,
  and `resolve.go` surfaces `ErrProvenRelicRefusal` instead of descending to the
  work-store twin). Review validated this against source, and the change ships
  end-to-end evidence: a `gc beads show` refusal test (with a controller-down
  control) plus a control proving a refused city *without* a census memo still
  declines.
- **By-id binding-owner callers fail closed.** All five `cliByIDBindingOwner`
  callers check `err` before `ok`, so `ErrProvenRelicRefusal` cannot fall
  through to a frozen read — verified in review.
- **Fix is correctly scoped to the only reachable path.** The bug requires a
  standing storage refusal on the residence-probe leg, which is minted only in
  `cmd/gc`; review independently confirmed the CLI degrade path is where the bug
  lives, so scoping there is not an omission.

Non-blocking follow-ups (shared for visibility — NOT addressed in this PR):
- **ga-vs4ms** — *Change Impact / Blast Radius (minor):* the API and dispatch
  by-id topologies build bindings with `KnownRelics` absent, so they would
  return `PolicyRefusalTolerated` rather than the proven-relic `PolicyFatal`.
  This is safe **today** only by an implicit reachability invariant (standing
  refusals are cmd/gc-only, and a refused city aborts daemon/controller
  construction), so those planes never run against a refused city. Recommended
  hardening: pin that invariant with an assertion/comment at the API/dispatch
  `BuildBindings` call sites, and thread `KnownRelics` there before any change
  that lets the controller serve a degraded city ships.
- **ga-e3dvx** — *Debuggability & Operability (minor):* the proven-relic refusal
  case keys on `leg.Role == RoleResidenceProbe && IsStandingRefusal(err)` and
  emits a "durable relic census proved …" message, but the branch does not
  itself verify that cause — it is true only because `planByID` is the sole
  producer of a `PolicyFatal` residence probe. Recommended: key the case on
  `leg.OnError == PolicyFatal` (or add an invariant comment) so the message
  matches the exact condition it asserts. Runtime behavior is correct as written.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `30810c6` has passing required GitHub

CI: https://github.com/gastownhall/gascity/actions/runs/32918085190

### Review Iterations
1 review pass performed with 0 fix iterations — approved on the first pass.

---

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR teaches the by-id read
path to refuse a migrated-then-refused city instead of silently answering from
the frozen pre-migration twin — a two-bit relic model (retire vs deny) whose
proven-relic case fails loud and closed with a legible `ErrProvenRelicRefusal`
rather than handing back a stale read. That closes a real silent-wrong-answer
hazard: a city that copied its beads into a new store and then had its
`[storage]` binding refused could previously still serve stale by-id reads from
the old frozen copy. It matters because a by-id read that quietly resolves
through a refused binding is exactly the kind of failure a residency split is
supposed to make impossible.

branch.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- The SQLite `DepMetadata` reader reported a non-carrying-but-nonempty edge
  payload (`"{}"`/`"[]"`/`"null"`) as `carried=true`, while the native Dolt
  reader gates on `DepMetadataCarries` and answers `("", false)`. The two
  engines then disagreed on the `carried` bit that the cross-engine migration
  witness hashes, so a byte-identical edge could read as "payload lost" — a
  latent false positive in the very seam this migration claims is closed
  byte-for-byte. The final patch gates the SQLite reader on the same predicate
  and adds a Dolt-free write→read parity test over the non-carrying-but-nonempty
  payloads (the coverage gap that made the divergence invisible).

Review findings addressed:
- `e90f3ee15` — fix(beads): gate the SQLite DepMetadata reader on
  `DepMetadataCarries` for cross-engine parity. Resolves the Contract & Interface
  Fidelity finding above (minor, high-confidence) and pins it with a new
  cross-engine parity test.

The review also premise-validated the load-bearing correctness claims and all
held: closed relics stay reachable (the retirement census counts closed
residents and the verdict is monotone, so staleness can only keep a probe
alive); a refused city fails loud rather than degrading to a work-only answer
(topology is built from opened routes, not `[storage]` config); a proven-relic
refusal denies rather than falling through to the frozen work copy; and the
two-bit asymmetry (`HasLegacyResidents` for retire, pessimistic-unknown=true, vs
`KnownLegacyResidents` for deny, evidence-only-unknown=false) is re-checked at
both the constructor and hand-assembled-binding boundaries so the contradictory
state is unspellable.

- mc-5n7b: [Change Impact / Blast Radius, minor] `gc convoy` on a rig-resident
  id hard-fails on a refused + proven-census city (a broken/removed `[storage]`
  section on an already-migrated city). The denial itself is by-design — loud
  and fail-closed, never a stale answer — so only the remedy text is arguably
  misdirecting (it points at storage convergence rather than the refused binding
  sitting ahead of the work axis). Optional diagnostic-only improvement; not
  addressed in this PR.
- mc-1bld: [pre-existing, unrelated code — NOT part of this PR] review surfaced a
  possible latent issue in `internal/events/reader.go`: the 64 KB AfterSeq
  active-log probe may short-circuit the active scan when a valid oversized line
  is present. This file is byte-identical between the base and this PR head, so
  it is unrelated to your change; filed as its own issue for independent
  verification.
- mc-i4k9: [pre-existing, unrelated code — NOT part of this PR] review surfaced a
  possible latent issue in `internal/session/chat.go`: the stale-key retry's
  `stripSessionIDFlagArg` may strip the wrong `--session-id`. Also byte-identical
  between the base and this PR head; filed separately for independent
  verification.

### Maintainer Changes
Maintainer fixes were pushed directly to the PR branch. One maintainer commit was
added on top of your work:
- `e90f3ee15` fix(beads): gate the SQLite DepMetadata reader on
  `DepMetadataCarries` for cross-engine parity — `internal/beads/dep_metadata_test.go`
  (+55), `internal/beads/sqlite_store_graph_apply.go` (+8); 2 files, +63.

Your commits were also rebased onto the current `residency/storeref-series` base
(the contributor merge commit that pulled in the series base was replaced by a
clean linear rebase), with your authorship preserved on every original commit.

### Final Review Status
Ready for the merge queue: final head `e90f3ee15` has passing required GitHub

CI: https://github.com/gastownhall/gascity/actions/runs/33066133562

### Review Iterations
1 review pass performed, 0 fix iterations — the single maintainer commit above
was an opportunistic cleanup of a minor finding, not a re-review gate.

---

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277291107

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277291180

julianknutsen:
Adoption preflight could not produce an adopted head: PR #5626 conflicts with base `residency/storeref-series` in both deterministic rebase and merge attempts. The branch is 155 commits ahead / 66 behind the configured base and produces broad conflicts across residency sources, tests, generated dashboard artifacts, and schemas. Please refresh/rebase the PR branch onto the configured base, then rerun adoption.

julianknutsen:
Maintainer preflight found the PR currently conflicts with its configured base branch `residency/storeref-series` (GitHub reports `mergeable=CONFLICTING`). Please rebase or otherwise resolve the conflicts against that base and push an updated head before review can proceed.

julianknutsen:
Fable review of 4a2dea5ad4 (base residency/storeref-series @ f79429f461): **approve; label withheld until #5597 (s8) merges**, then retarget to `main` and rebase (expect `cmd/gc/cmd_bd_by_id.go` and the residency baseline to need reconciliation against pr8 #5634 and s5 PR A #5644 if they land first).

Checked: `ResolveBindingOwner` stops at the caller's work axis by ROLE (work-fallback or shadow) and never probes it (pinned at zero gets); `KnownLegacyResidents` only ever denies and `BuildBindings` raises the pessimistic bit on proof so the two cannot disagree; `probeRefusalPolicy` keys `PolicyFatal` on the proof and `resolveByID` keys the denial arm on the POLICY, ordered before the tolerated arm (ga-e3dvx); `ProvenLegacyResidents` fails the other way from `HasOpenLegacyResidents` (a census that could not run proves nothing); the T3k corpus rows; the proof is computed live on a refused city only, memoized per process and dropped with the residency memo, reuses the registered routes when a controller holds them, and — after the follow-up — opens only a binding that already exists on disk (`refusedBindingIsAlreadyOnDisk`: root enumerable + database present for the SQLite target, the provider's served location present otherwise). Remedy is a second sentence on one line. Residual kept on ga-q8ick: gated on the open-only census until s5's closed-inclusive verdict is on main (one-line swap in `ProvenLegacyResidents`).

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277304449

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277305515

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR closes a silent-wrong-answer hole on the by-id read path: on a city that ran `gc storage migrate` and then fell into a standing storage refusal, `gc beads show <id>` was answering out of the frozen pre-migration twin and exiting 0, so an operator got a stale bead back with no signal that the city was not servable. It now refuses, exits 1, and names `gc doctor` on the same line. That matters because such a city is already in an incident state its own boot gate reported, and a confident wrong answer there is worse than no answer.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **The reported bug is real, and the fix lands where it was reported.** Reviewers validated the premise against the base rather than accepting the PR's RED log: the tolerated-policy literal, the `ok=false` fall-through, and the fallback directory scan are all visible on the minus side, and no downstream mechanism intercepted the twin.
- **The proof fails in the safe direction.** `ProvenLegacyResidents` returns false on census error, so an unreachable binding can never *cause* a denial — pinned by `TestProvenLegacyResidentsFallsTheOTHERWay`. Unknown is never read as proven.
- **The denial arm is keyed on policy, not on an error string.** It requires `RoleResidenceProbe && OnError == PolicyFatal && IsStandingRefusal`; a Fatal probe carrying a non-refusal error falls through to `default`, so a transient fault never wears the proven-relic sentence. Traced independently by two reviewers.
- **The proof never creates the binding it is asked about.** `refusedBindingIsAlreadyOnDisk` runs before any open and reuses the migration's own no-open helpers; both arms — absent root, and present-but-empty root with an absent database — are independently load-bearing.
- **A served city pays nothing for the new protection.** `topo.Refused == nil` short-circuits ahead of any census, and `TestServedCityPaysNothingForTheRelicProof` counts zero plan resolutions.
- **The census and the refusal path spell the binding ref the same way.** Both derive it through a single derivation, and `TestTheProofAndTheRefusedFunnelSpellTheBindingRefTheSameWay` pins the silent-miss failure mode a divergence would produce.
- **The residency boundary baseline moved exactly one renamed row** — 114 data rows before, 114 after — so no new blind reader was admitted alongside the fix.

Review findings addressed:
- **The `KnownRelics` field doc prescribed exactly what the design forbids** (contract fidelity, minor, high confidence). It said the field answers "from a DURABLE record ... has a census *ever* proven ...", but the sole implementation is a census the calling plane runs at read time — and this PR's own header rejects a durable note precisely because it goes stale in the *denying* direction. A contributor implementing to that field doc would have reintroduced the bug the PR exists to prevent. Reworded to the implemented semantics in `7437a6ca29`.
- **A new doc named a caller that does not call the seam** (pattern drift, minor, high confidence). `ResolveBindingOwner`'s new doc listed "the `gc bd` class door" next to `gc convoy`'s member scan, but `cmd_bd_by_id.go` never calls `ResolveBindingOwner`, `cliByIDBindingOwner`, or `byIDResidencyTopology`, and does not import `storeref` at all — so the doc implied the bd door was already served by the converted seam, the opposite of disclosing the gap. `7437a6ca29` names the real callers (`gc convoy`'s member scan and `gc beads show`'s store fallback) and records the deferral inline.
- **The `StandingStorageRefusal` marker doc stated the carve-out unconditionally** (pattern drift, minor, high confidence). That marker is what new code against the funnel gets written against, so `7437a6ca29` names `ClassBinding.KnownLegacyResidents` as the condition that withdraws the tolerance.
- **"once per process per city" overclaimed** (concurrency/state safety, minor, medium confidence — latent). `provenRelicRefsForCity` releases the memo lock across `censusRefusedCityBinding`, so the one-read-per-city property holds for the sequential one-shot cobra callers this door actually has, rather than by construction. `7437a6ca29` says so in the header and names the second-handle hazard; the single-flight itself is deferred (below).
- **The denial's breadth was undisclosed** (blast radius, minor, high confidence). The proof is keyed by binding ref, so on a proven-relic city *every* non-reserved by-id read on the CLI fallback path denies — including a fresh post-migration work bead and a rig-shadowed id that can have no frozen twin at all. Review concluded the breadth is right rather than accidental (the census reads open beads only, so a per-id rule would deny *less* than the evidence supports) and the conformance corpus already pins it; `7437a6ca29` records the operator cost in the header.
- **A redundant `storageSplitWhole` guard read as a second independent precondition** (readability, nit). `7437a6ca29` annotates it as deliberately re-derived and as what makes `binding` safe to read below.

- **`ga-3htcm`** (cross-file consistency, minor after adjudication, high confidence): the `gc bd` class door (`cmd/gc/cmd_bd_by_id.go:790`) and the hook claim route (`cmd/gc/claim_class_route.go:311-312`) still hand-roll the tolerated-refusal carve-out over their own graph handle and never consult the proven-relic bit, so on a proven-relic refused city they keep serving — and, on the claim route, keep *writing* — the frozen twin while `gc beads show` now refuses. One reviewer filed this as gating; it was de-gated on causation grounds, not dismissed. Both files are byte-identical between base and head and appear nowhere in this PR's diff, and no path regressed — the PR takes three broken surfaces to two. The write half is the more serious exposure, and it is tracked.
- **`ga-nzxob`** (minor/nit, code halves only): a per-city single-flight around `censusRefusedCityBinding`; absolutizing a relative foreign binding location against `cityPath` before the existence stat (unreachable in-tree, since both engine providers implement `BindingLocation`, but a downstream provider without a locator would inherit CWD-dependent behavior); memoizing the proof-decorated bindings entry alongside the plain one so a convoy member scan does not re-derive the grouping per member; and asserting the `registeredResidencyEntry` invariant locally rather than relying on one three files away. The doc-only halves of the first and last landed here in `7437a6ca29`.

### Maintainer Changes
One comment-only commit pushed to this branch: `docs(review): correct the relic-proof docs that overclaimed` (`7437a6ca29`) — 4 files changed, 65 insertions(+), 16 deletions(-), with no executable line changes.

The branch was also rebased onto current `main`. Your final commit, `fix(residency): rename the baseline row #5597 renamed the function under`, dropped out as an empty rebase step — exactly as its own commit message predicted, because `main` now carries the identical re-key. That, plus the fixup, is why the branch was force-updated; the adopted patch content is otherwise unchanged (`git patch-id --stable` verified identical across the rebase).

### Final Review Status
Ready for the merge queue: final head `7437a6ca29` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/34170009265

### Review Iterations
1 review pass, 0 fix iterations. Three independent reviewers (behavioral/contract, evidence-bundle/correctness, and cross-file consistency/pattern drift) all completed, and the pass resolved to approve with no required changes. The one disagreement — whether the unconverted `gc bd` door gated this PR — was adjudicated by re-verifying base/head blob hashes for both cited files and filing `ga-3htcm` as the condition attached to the approval.

---

PR #5628, opened 2026-08-25
fix(bd): disclose which store gc bd list/ready/search/show answered from (#5170)

### LINKED ISSUE #5170: gc bd list/ready silently answer from one undisclosed store selected by cwd
`gc bd list` and `gc bd ready` resolve which bead store answers via a multi-level priority chain (`resolveBdScopeTarget` in `cmd/gc/cmd_bd.go`: explicit `--rig` > explicit `--city` > bead-prefix detect > `-C`/`--directory` > `GC_RIG` env > cwd > city), and no output field in any mode (human or `--json`) discloses which store answered or which chain level selected it.

Reproduction (city with one bound rig; requires `gc`, `bd`, `jq`):
```
gc init --template minimal --default-provider claude --no-start demo-city
mkdir demo-rig && (cd demo-rig && git init -q && git commit --allow-empty -q -m init)
gc --city demo-city rig add demo-rig --name demo-rig --prefix dr
(cd demo-city && bd create --title "t" --assignee alice --json)   # -> dc-xxxx

cd demo-rig  && gc bd list --assignee alice --json   # exit 0, []
cd demo-city && gc bd list --assignee alice --json   # exit 0, [{"id":"dc-xxxx",...,"assignee":"alice"}]
cd demo-rig  && gc bd show dc-xxxx --json            # exit 0, finds it — prefix-routed, cwd-independent
```

Observed: the byte-identical `list --assignee alice` invocation returns `[]`/exit 0 from `demo-rig` and one row/exit 0 from `demo-city`, with no diagnostic on stdout or stderr distinguishing "this store has no matches" from "a different store answered." `show`, in the same shell, finds the same bead via bead-prefix routing (which does not depend on cwd), demonstrating the record is retrievable in that shell by a path `list` did not take — not that `list`'s own resolution could have reached it.

Expected: some disclosure — a store/scope field in `--json` output, or an equivalent human-output line — naming which store and which chain level answered, so a zero-row result is distinguishable from a true empty.

Code path: `cmd/gc/cmd_bd.go`, `resolveBdScopeTarget` (chain, ~line 664) and `doBd` (~line 209), which execs `bd` as a subprocess with `cmd.Stdout` as a raw passthrough — gc does not parse or reconstruct bd's stdout, so any fix that needs to modify what's disclosed has to add it out-of-band (e.g. stderr) rather than into bd's own `--json` payload.

Verified present at `365f111ea2930269b78afd53d076fa9e37044fdf` (main, 2026-08-10).

This is the same invariant as the recent relocated-class work (#5162, #5167) — a projection answering `[]`/exit 0 from a store it cannot see the full picture from — applied to the ordinary cwd-scoped city/rig case, where no relocated class is involved. Those changes guard `--metadata-field`-selected relocated-class ids and the `ready` frontier on split cities specifically, and do not cover this path.

rjgeng:
Fixes #5170.

`gc bd list`/`ready`/`search`/`show` silently answered from one store selected by cwd, with no indication to the caller which store that was — in a multi-store setup, an answer could look complete while actually reflecting only one store's contents, with no signal that anything was scoped at all.

**Fix:** these commands now disclose which store answered the query.

**Testing:** `go build`/`gofmt -l` clean, `TestGcBdDisclosesAnsweringStore` (6 subtests) plus the full surrounding `cmd_bd_test.go` suite pass.

---

### DIFF
--- a/cmd/gc/cmd_bd.go
+++ b/cmd/gc/cmd_bd.go
@@ -289,6 +289,26 @@ func doBd(args []string, stdout, stderr io.Writer) int {
 		}
 		return doBdReleaseIfCurrent(cityPath, cfg, target, id, expectedAssignee, stdout, stderr)
 	}
+
+	// Disclose which store answers a read-only passthrough, so a zero-row
+	// result is distinguishable from a true empty (gastownhall/gascity#5170).
+	// resolveBdScopeTarget's priority chain (explicit --rig > explicit --city
+	// > bead-prefix detect > -C/--directory > GC_RIG env > cwd > city)
+	// silently picks a store on every one of those paths but the GC_RIG-
+	// mismatch warning above; the common cwd-auto-detect case reached bd with
+	// no diagnostic at all. Placed after the by-ID and release-if-current
+	// arms above (rather than immediately after resolveBdScopeTarget) so it
+	// names the store that actually serves the request: a class-owned `show`
+	// on a split city is answered in process from the class's own binding by
+	// maybeRouteBdByID, not from target, and disclosing target there would be
+	// wrong for that one read. This is stderr-only and additive — bd's own
+	// stdout (human or --json) is untouched, and it never changes the exit
+	// code, matching the disclosure style #5162/#5167 established for the
+	// sibling relocated-class invariant.
+	if verb, _, ok := bdRelocatedClassVerb(bdArgs); ok && bdScopeDisclosureVerbs[verb] {
+		fmt.Fprintf(stderr, "gc bd: answering from the %s store\n", scopeLabel(target)) //nolint:errcheck // best-effort stderr
+	}
+
 	if provider := rawBeadsProviderForScope(target.ScopeRoot, cityPath); !providerUsesBdStoreContract(provider) {
 		fmt.Fprintf(stderr, "gc bd: only supported for bd-backed beads providers (resolved %q for %s)\n", provider, target.ScopeRoot) //nolint:errcheck // best-effort stderr
 		if hint := bdProviderMismatchHint(target.ScopeRoot, provider); hint != "" {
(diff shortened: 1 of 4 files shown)


sjarmak (approved):
Taking this. The disclosure is right, the placement reasoning is right, and the test is a real regression test. One shipped script broke on it, and I fixed that here rather than sending the PR back.

**The fix does what it says, and the test proves it.** Reverting `cmd/gc/cmd_bd.go` to its parent and running `TestGcBdDisclosesAnsweringStore` fails on every subtest with the disclosure string absent, so this is red on unfixed main rather than a test written around existing behavior. Scoping the announcement to reads is the right line: a write's effect is re-readable, so its store is recoverable after the fact, while a read's silence is the entire defect #5170 describes.

**The placement argument holds.** The disclosure sits after the by-ID and release-if-current arms, and that ordering is load-bearing rather than incidental: a class-owned `show` on a split city is intercepted by `maybeRouteBdByID` and served in process from the class's own binding, so announcing `target` there would name a store that did not answer. Between scope resolution and the disclosure the only other early return is release-if-current, which is a write. After it, the passthrough runs `bd` with `cmd.Dir = target.ScopeRoot` and target-derived env, so nothing re-routes the read and makes the line a lie.

**On disclosing unconditionally.** I went looking for whether this should be gated on a city having more than one possible store, and concluded it should not. A rig-count gate would miss the split-storage case, where a second store exists with no second rig, which is precisely the shape #5162 and #5167 were about. Any gate that is actually correct would have to reason about storage bindings as well, and would then be more machinery than the line it suppresses. Unconditional is also the more useful contract for a script or an agent reading the output: the line is always there to be read, rather than present only in the configurations someone remembered to enumerate. Worth noting the channel is not otherwise silent, either. Standing up the new test's city surfaces two pre-existing stderr warnings from `gc bd` on the same invocation, so callers already cannot assume a clean stderr here.

**What I fixed.** `internal/bootstrap/packs/core/assets/scripts/spawn-storm-detect.sh:79` captured `gc bd show "$tid" --json 2>&1` and fed the merged stream straight to `jq`. With the new line on stderr, that parse fails, the `|| echo "unknown"` fallback takes over, and every bead reads as `unknown`, so the ledger prunes nothing and grows without bound. Measured on the two branches of that conditional with the disclosure line present:

```
before:  closed bead -> unknown   (never pruned)
after:   closed bead -> closed    deleted bead -> deleted
```

That script was already fragile in the same way, since either of the two pre-existing warnings above would defeat the parse identically. This change is what makes it reachable in the ordinary case, so it belongs with it. Fixed in `d287db27a`: the success path reads stdout only, and the error path re-reads with stderr merged, which is where the not-found text lives. It is the only shipped script that merges the streams; the other `gc bd` callers under `packs/core/assets/scripts/` suppress stderr already.

The same commit carries the regression test. `TestSpawnStormDetectPrunesClosedAndDeletedLedgerEntries`'s `bd` stub now writes a stderr line on `show`, which is what a real `gc` does. Without the script fix that test fails with "closed bead was not pruned"; with it, all five `TestSpawnStormDetect*` cases pass.

Non-gating: `bdRelocatedClassVerb` returns not-ok when it meets an unrecognized flag before the verb, so a `gc bd --some-unknown-flag list` would skip the disclosure silently. Every flag bd actually accepts is enumerated in `internal/bdflags` and pinned by its own test, so this only affects malformed invocations, which have bigger problems.

Codex (gpt-5.4) reviewed independently and is what found the `spawn-storm-detect.sh` regression; it confirmed the placement claim, the absence of a competing early return or later reroute, and that no Go caller parses this stderr.

PR #4143, opened 2026-07-11
feat(api): Location headers + run stanza on launch endpoints (P0 #2, S2)

julianknutsen:
## What

Slice 2 of the Run-resource track: point the two launch endpoints at the canonical Run resource so a client can follow a launch straight to the run.

- **`POST /sling`** — when a graph workflow launched (a run root exists), set `Location: /v0/city/{cityName}/runs/{workflow_id}` and add a `run` `{run_id, kind:"sling", status:"pending"}` stanza to the body. Every other successful shape (wisps, plain bead routes, idempotent skips) has no single run, so its `Location` is the runs list — matching the existing `dashboard_url` discriminator.
- **`POST /order/{name}/run`** — dispatches **asynchronously** and has no run root at response time (the synchronous result is an order tracking bead, not a graph run root), so its `Location` is the runs list, where the dispatched run appears once it materializes. Existing `tracking_id` kept.

Re-adds the `RunRef` type (removed from the S1 read slice as unused there); `kind` is a closed `sling,order` enum. Path helpers use `url.PathEscape`. Spec + Go client regenerated.

## Depends on #4139 (S1)

This branches on the S1 read-side slice. Until #4139 merges, this PR's diff includes S1's files; it cleans to just the S2 changes once S1 lands. Base is `main` (not stacked on the branch) to avoid the pipeline's stacked-PR handling.

## Verification

- 4 targeted tests (path-escaping unit; sling non-workflow → list Location + no stanza; **graph-launch → deep Location + run stanza** via the real handler; order-run → list Location) + spec-sync; full `internal/api` suite green (205s); pre-push `make test-fast-parallel` passed.
- **Fable red-team: SHIP** — Location is never stamped on error responses, the escaped URL round-trips to the registered route, the deep-linked run root is event-appended before the response (so a GET resolves it), and the 202 + Location fields don't interfere.

Follow-up: S3 cancellation (`POST /runs/{run_id}/cancel`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #5629, opened 2026-08-25
fix(workquery): routed claim tier rides the reader's canonical priority order

csauer02-personal-user:
## Summary

The routed pool claim tier (`routedReadyTierCommand`) pins `--sort oldest`, which combined with its bounded window makes worker claims priority-blind in the strong sense: a routed high-priority bead behind more than `--limit` older lower-priority rows is never served at all until the backlog drains. This PR drops the flag so the tier rides the reader's canonical `(priority, created_at, id)` default order — priority-first, FIFO within a priority band — and updates `engdocs/architecture/dispatch.md`, which documented FIFO-before-priority as the intentional policy.

## Symptom (measured, live 14-seat city)

13 P0 rows routed to a pool sat at positions 35–48 of 49 behind 34 older P1/P2 rows. The deployed tier's 20-row window contained **zero** P0s; the flagless window served all 13 first — same store, same instant, same predicate. At measured claim throughput (a few rows/hour) the P0s were many hours from being served; deferring the older rows was the only interim. Deployed on that city, the first P0 claim landed within ~90 seconds of the binary swap.

## Root cause

`internal/config/workquery.go` `routedReadyTierCommand` passes `--sort oldest --limit=20` to `bdReadyPoolDemandShell`. Both readers default to the canonical priority-first order (`bd ready`'s default `--sort priority`; the federated reader's merged `beads.SortBeadsReadyOrder`), so the explicit flag is the single suppression point. The hook layer claims the first eligible row of whatever the tier serves, so tier order is claim order.

## The policy reversal, named

`engdocs/architecture/dispatch.md` stated the FIFO ordering intentionally: *"so newer high-priority work does not jump ahead of older ready work already queued."* That rationale predates the interplay with the bounded window, where the outcome is not fair queuing but total unreachability of high-priority rows. The created_at term of the canonical order preserves the rationale's legitimate half (arrival order within a priority band). All three ordering sites in the doc are updated in this PR; the contract tests that pinned oldest-first (from #4143, which carried no rationale in its commit) are inverted with rationale comments.

## Not changed (deliberate)

- The tier-2 `gc.run_target` migration fallback keeps `--sort oldest` (retirement-window probe, ga-dhf44).
- The reconciler count-form is order-free; `scale_check ↔ work_query` correspondence is unchanged.
- The control dispatcher's own probes (`dispatch_runtime.go`) carry the same ordering and are left for a separate change.
- No aging term is added: priority-first-without-aging is the intended contract here; a starved lower-priority band under a sustained high-priority flood is accepted (and was strictly worse under the deterministic inversion this replaces).

## Testing

- `internal/config` full suite green; targeted `cmd/gc` claim/work-query/dispatch suites green; `go vet` clean; docsync suite green on the doc edit.
- 24 goldens regenerated — word-diff per file is exactly `--sort oldest` removed, nothing else.
- New regression test pins the contract across default/bd-1.0.5/federated topologies; a mutation reintroducing the flag fails 7 tests.
- Independently adversarially reviewed before deployment on our city (golden word-diffs, both-tree suite comparison, live served-window flip reproduced read-only, mutation test).

## Field deployment

Running in production on a 14-seat city since 2026-08-25: first P0 claim ~90s post-swap, P0 wave (50 steps) completed within 90 minutes, window correctly serving the next band after.


sjarmak (dismissed):
I verified this end to end rather than reading it, and everything in the description holds. One design question below is why I am tagging Julian rather than merging it myself.

**The mechanism.** `bd ready --help` on this machine reports `Sort policy: priority (default), hybrid, oldest (default "priority")`, and the federated path's `beads.SortBeadsReadyOrder` sorts `(priority, created_at, id)` ascending with a nil priority collapsing to 2 to match the SQL readers' `COALESCE(i.priority, 2)` (`internal/beads/query.go:383-396`). So both readers really do default to priority-first, and the explicit flag really was the single suppression point. Removing it yields priority-first with `created_at` as the tiebreak, which keeps the arrival-order rationale the doc gave, restricted to within a band, and that is what the updated doc now says.

**The goldens are tighter than described.** The description says the word-diff per file is `--sort oldest` removed. It is stricter than that: for each of the 24 files, deleting only the **first** occurrence of `--sort oldest ` from the pre-change file reproduces the post-change file byte for byte. The second occurrence, the tier-2 `gc.run_target` migration fallback, survives in all 24. That matches the deliberate-exclusion list and means no golden picked up an unrelated regeneration.

**The mutation claim is exact.** Restoring the flag and running `go test ./internal/config -count=1` fails exactly seven top-level tests: `TestEffectiveWorkQueryDefault`, `TestEffectiveWorkQueryRoutedTierServesCanonicalPriorityOrder`, `TestEffectiveWorkQueryBD105CompatibilityOptIn`, `TestEffectiveWorkQueryRoutedQueueUsesNativeCanonicalSortAcrossReadyTiers`, `TestEffectiveWorkQueryRoutedQueueRidesReaderPriorityOrder`, `TestPoolDemandPredicateSharedWithWorkQuery` (both subtests), and `TestWorkQueryGolden`. Reverting returns the package to green in 2.07s.

**Blast radius.** The correspondence risk is the one I went looking for, since a helper shared between the `scale_check` count form and the `work_query` first-row form is exactly where an ordering change can bite. It does not bite here: `poolDemandCountShell` discards row identity and emits only a length, and the callers in `cmd/gc/pool.go` and `cmd/gc/pool_desired_state.go` consume it as a scalar session count, never as same-row or same-prefix evidence. The strongest agreement assertion in the tree is the weaker one, that nonzero `work_query` rows imply nonzero demand, which ordering cannot affect. Downstream, `filterUnreadyHookCandidates` removes rows without reordering and the claim path takes the first survivor, so no consumer of this tier assumes arrival order. The control dispatcher's own probes in `dispatch_runtime.go` and `dispatch_control_ready.go` still sort oldest-first, but they are a different generator, not a consumer of this one, which matches the deliberate-exclusion list.

**The design question, and why I am not merging this myself.** Dropping the flag also settles priority-versus-fairness for every pool claim in every city, with no aging term. The description names this and argues it is strictly better than what it replaces, and I agree with that comparison: today's behavior starves high-priority work *deterministically* whenever the backlog exceeds the window, whereas the new behavior starves a lower band only under sustained high-priority flood. I searched for something that would deadlock on the new failure mode, a retirement probe or drain or health check that needs old low-priority routed rows to eventually be served, and did not find one; the pool is a pull system scaled on demand count, not on bead age. So I do not think it is a regression.

It is still a scheduling policy being decided, and it reverses a rationale that `engdocs/architecture/dispatch.md` stated on purpose. @julianknutsen, the specific question is whether priority-first-without-aging is the contract you want for pool claims, or whether the fix should carry an aging term so a low band cannot be held off indefinitely. Nothing is blocked on the answer, since this is already running on the author's city.

One non-gating note. The field evidence in the description is from a single 14-seat city, so it establishes that the fix works, not the size of the effect elsewhere. That does not change the decision; the code argument stands on its own.

Codex (gpt-5.4) reviewed independently against `9bef3024d` and returned no substantiated issues on all five questions I put to it, including running the mutation itself and confirming the three-topology test fails at `config_test.go:1903` for `default`, `bd105`, and `federated` with the flag restored.

csauer02-personal-user:
Author's-city input on the open design question (priority-first-without-aging vs an aging term), since we adjudicated exactly this before deploying and now have a day of field behavior under it.

**Our ruling was accept-and-monitor, no aging term** — with the reasoning that an aging term today is machinery for a failure mode nobody has measured: at our arrival rates, high-priority streams are bursts, not sustained floods, and the deterministic inversion the flag caused was the measured, daily harm. We defined the trigger explicitly instead: a real starved-lower-band incident is itself the evidence that commissions an aging design, and until one fires the aging term is speculative complexity in the hottest path of the claim system.

**Field observation from the deploy day, under worst-case pressure**: the fix landed mid-incident with 13 P0 rows queued against ~34 lower-band rows — the strongest sustained high-priority flood our city has produced. The P0 wave claimed first (first claim ~90s post-swap), drained in ~90 minutes, and the lower band resumed serving immediately after with no intervention. Overnight since: ~49 rows landed across 9 batches with the lower band grinding continuously between priority bursts. So the theoretical starvation mode has not materialized even on the day engineered to maximize it, while the deterministic starvation it replaced was reproducible on demand.

**On the shape, if you do want aging**: the natural term is a created_at-based promotion into the head band after a threshold age, applied in `SortBeadsReadyOrder` so both readers inherit it — but that decides a second policy (how old is too old) on no data. Our suggestion is to take the fix as-is and let the first real starvation report, here or elsewhere, size the aging term with an actual measurement behind it.

Either way it's your contract to set for the ecosystem — our city runs the current shape deliberately and will keep reporting field behavior.

csauer02-personal-user:
**Field correction to my earlier comment — the starvation mode has now materialized, and the data answers the open question in the other direction.**

My earlier comment said the theoretical lower-band starvation "has not materialized." It had, overnight, and we measured it this morning: under a saturated pool (8 busy seats against a ~12 brake) with P1 work arriving faster than seats freed, **31 P2 + 5 P3 routed roots sat sessionless for 6+ hours** while every dealt root was P0/P1. Worse, the failure has a feedback loop we hadn't modeled: our review/gate molecules file at P2/P3, and reviews gate the P0/P1 work — so pure priority-first starves the very gates the high-priority program waits on, throttling P0 throughput through its own scheduler.

So on the design question: **at saturation, priority-first does need an anti-starvation term** — our accept-and-monitor ruling had defined exactly this incident as the trigger, and it fired within a day. Two shapes we're now designing, in order:

1. **Priority inheritance through blocking edges** (primary, and we'd suggest it's the *principled* fix): a gate/review bead inherits the max priority of the beads it blocks. This cures the measured loop at its root — the starved reviews were arguably misfiled, not mis-scheduled — and it avoids touching the sort contract at all.
2. **An aging term** (backstop, for starvation shapes inheritance doesn't model): effective priority rising with wait age, which for this codebase means both readers (`bd`'s native priority sort and `beads.SortBeadsReadyOrder`) — heavier, and the threshold wants sizing data. Our incident's datum: visible starvation at ~6h under sustained saturation.

The fix in this PR still stands on its own — the deterministic inversion it removed was strictly worse, and the interim cure for our incident was simply re-filing the gating reviews at their true priority. But "priority-first with no anti-starvation term" is not a complete contract under saturation, and we were wrong to suggest the mode was theoretical. We'll share the inheritance design when it's written if it's useful upstream.

jacobhausler:
Second-city confirmation of the bug this fixes: on `main` @ fc7a23a29 (2026-09-05) a 4-seat routed pool with 47 candidates never claimed a P0 filed at 17:13Z for ~3 h because 30 older P1/P2 rows filed 16:48–17:12Z filled the `--sort oldest --limit=20` window ahead of it (P0 ranked 31st by age, 1st by canonical order). Full numbers on #5970. We had this same change ready to PR and stopped at the duplicate gate in favour of this one — it would be good to see it land. The aging / priority-inheritance follow-on from your 2026-08-26 correction reads right to us too; we see the same feedback loop (P2-filed gate beads blocking P0 roots) on our side.

A3Ackerman:
Third-city confirmation, on a different shape of the same bug: on 2026-09-05 a dedup order on our review plane lifted a labelled PR's review beads to P0, and the orchestrator's next four claims (10:48Z, 11:49Z, 12:05Z, 12:14Z) still took the oldest unclaimed root each time, twice passing the P0 first-steps for a P2 one — the window was the whole story, as here. We had a per-agent sort policy for this tier shaped and stopped at the duplicate gate in favour of this; it would be good to see it land.

On the anti-starvation follow-up from the 08-26 correction: bd's reader already ships an aging-shaped policy next to `priority` and `oldest` — `bd ready --sort hybrid` (recent rows by priority, older rows by age) — so a backstop may not need new sort machinery, only a way for a pack to pick that policy for the routed tier per agent while priority inheritance through blocking edges is designed. That narrower question is what I left on #6094; happy to fold it into whatever direction this PR sets.

csauer02-personal-user:
The worker first-row pool tier (`routedReadyTierCommand`) passed `--sort oldest`, which made the claim window priority-blind: a routed P0 sitting behind more than `--limit` older lower-priority rows was never served at all, so pool workers deterministically starved the highest-priority routed work.

Both readers already default to the canonical `(priority, created_at, id)` order — `bd ready`'s default `--sort priority`, and the federated reader's `beads.SortBeadsReadyOrder` — so dropping the explicit flag is the whole code fix. FIFO fairness survives within a priority band via the `created_at` term. The `limit=20` blocked-head fall-through is order-independent. The `gc.run_target` migration fallback keeps `--sort oldest` (it is a retirement-window probe, ga-dhf44), and the reconciler count-form is order-free, so the scale_check / work_query correspondence is unchanged.

This does not touch the control dispatcher's own routed probes (`cmd/gc/dispatch_runtime.go`, `cmd/gc/dispatch_control_ready.go`), which carry the same oldest-first class deliberately and are out of scope here.

### Documentation

`engdocs/architecture/dispatch.md` stated FIFO-before-priority intentionally ("so newer high-priority work does not jump ahead of older ready work"). That rationale predates the interplay with the bounded window, where the outcome is not fair queuing but total unreachability of high-priority rows until the backlog drains. All three ordering sites in the doc move to the new policy: priority-first, FIFO within a priority band.

### Measurement

Measured live (a 14-seat city, 2026-08-25): 49 routed molecules, the 13 P0s youngest — the deployed tier's 20-row window contained zero P0s, while the flagless window served all 13 first. At the measured claim throughput that put the P0s many hours out. After deploying the change on that city, the first P0 claim landed within 90 seconds of the swap.

Operating record: this fix has run on that city's served binary since 2026-08-25 — 16 days, every pool claim across three rigs — with no priority inversion reported.

### Tests

The two oldest-first contract tests came in with e3f469860 (#4143), an API-endpoints PR whose message records no scheduling-order rationale; the rationale lived only in `dispatch.md`, which this change updates. Behaviour-neutral migration table:

| Retired assertion | New owner |
| --- | --- |
| `TestEffectiveWorkQueryRoutedQueueUsesOldestBeforePriority` — routed tier serves the oldest row across priorities (`older-p2` served, `newer-p0` withheld) | `TestEffectiveWorkQueryRoutedQueueRidesReaderPriorityOrder` — the flagless routed query serves the reader's priority-first window, so `newer-p0` is surfaced |
| `TestEffectiveWorkQueryRoutedQueueUsesNativeOldestSortAcrossReadyTiers` — native oldest sort rides every ready tier | `TestEffectiveWorkQueryRoutedQueueUsesNativeCanonicalSortAcrossReadyTiers` — the tier list stays native and bounded, and the flagless routed query serves the row |
| (new negative assertion, no predecessor) | `TestEffectiveWorkQueryRoutedTierServesCanonicalPriorityOrder` — no topology emits `--sort oldest` on the routed probe |

Real-boundary proofs retained unchanged: the 24 `internal/config/testdata/workquery` goldens (byte-identity of the emitted shell), `TestPoolDemandPredicateSharedWithWorkQuery` (work-query and reconciler count-form share one predicate), and `TestEffectiveWorkQueryRoutedFallbackUsesNativeOldestSort` (the migration fallback still sorts oldest). Goldens regenerated with the repo's own generator; the word-diff per file is exactly the flag removed.

### Review

sjarmak approved on 2026-08-26 ("I verified this end to end rather than reading it, and everything in the description holds ... I do not think it is a regression"). jacobhausler confirmed the same bug from a second city on 2026-09-05. The branch has been rebased in place onto current `main`; the conflict was confined to the goldens and was resolved by regeneration.

Drafted by an agent.

quad341 (approved):
@csauer02-personal-user — dropping one flag is the easy part; what made this reviewable is everything you built around it. Regenerating 24 goldens so each word-diff is exactly `--sort oldest` removed, inverting the #4143 contract tests *with* the rationale the original commit never recorded, and scoping the new negative assertion to `routedReadyTierCommand` so it doesn't accidentally bind the migration fallback's contract — that let me verify the mechanism instead of trusting it. I re-checked both readers myself: `cmd_ready.go:568` → `SortBeadsReadyOrder`, and `bd ready --help` on our pinned v1.1.0 confirms the priority default. It holds. Merging as-is.

Your 2026-08-26 correction deserves calling out — coming back to say the starvation mode you'd called theoretical had fired overnight, with the loop where P2-filed review beads gate P0 work, is what makes field evidence worth anything.

Two follow-ups I'm taking on our side, not yours: the bd-1.0.4 ephemeral routed fallback still serves oldest-first within its 20-row window (pre-existing, separate path), and the anti-starvation work your priority-inheritance sketch points at. Please do share that design upstream.


---

PR #5630, opened 2026-08-26
Preserve retry failure reasons across delayed outcome visibility

quad341:
## What this changes

When a retry attempt closes but its outcome metadata is briefly delayed by Dolt read-after-write visibility, the dispatcher now re-reads the attempt before classifying it. Transient reasons such as `rate_limited` are preserved, so an exhausted optional reviewer follows the intended soft-fail path instead of being mislabeled `missing_outcome`.

The wait is bounded and applies only to a closed subject with no visible outcome or typed deliverable close. A genuinely outcome-less subject still reaches the existing `missing_outcome` behavior after the retry window.

## Review notes

- The resolver makes at most five observations with 100 ms between retries.
- Context cancellation stops the wait immediately.
- Typed deliverable closes and already-visible outcomes bypass the retry path.
- The change is confined to retry evaluation in `internal/dispatch`.

## Test plan

- [x] Run the documented 40-job local CI-equivalent suite with the rootless container environment and confirm both new retry-resolution tests report PASS.
- [x] Run `make test-ci-policy`, `go vet ./...`, changed-file formatting, and fresh-cache affected lint.
- [x] Release gate: [`release-gates/ga-graq9b-delayed-retry-outcome-visibility-gate.md`](release-gates/ga-graq9b-delayed-retry-outcome-visibility-gate.md)


quad341:
@quad341 — thank you for this one. What I like most is that you didn't invent a mechanism: `resolveRetrySubjectOutcome` mirrors `resolveScopeBody` constant-for-constant, down to the trace vocabulary and the timer-drain idiom, so it reads as the second instance of a house pattern rather than a new one to argue about. And `subjectOutcomeAmbiguous` is exactly as narrow as it should be — it fires only on the single `classifyRetryAttempt` branch that produces `missing_outcome`, so a pass, a fail, a cancel or a typed deliverable close never pays a millisecond.

Your delayed-visibility test is genuinely well built — hiding metadata on every `List` and the first three `Get`s lands resolution on the last attempt, and the `hiddenReads == 0` guard means it can't pass vacuously. I also checked your claim that post-exhaustion behavior is unchanged, and it holds: `TestProcessRetryEvalRetriesInvalidWorkerResultContract` now runs the full 400 ms window and still lands on `missing_outcome`.

One small thing I'll take on my side: a failed re-read mid-window currently returns an error that can quarantine the eval bead, so I'll degrade it to the last-known subject instead. The red shard is a `cmd/gc` supervisor test with no path into `internal/dispatch` — I'll re-run it. Merging after that.


---


quad341:
Re-reviewed `67aca6e4c`..`c2d56dcf2` (259 new commits since the last review): no new concerns — the prior review stands.

PR #5631, opened 2026-08-26
fix(pre-push): fail-close SIGPIPE and reap leaked push-gate slots (ga-34a)

brandonmartin:
## Summary

Fixes the closed re-orphan loop where `git push` exited 141 (SIGPIPE) after the pre-push hook printed success, the branch never landed, and pool-slot retirement leaked both push-gate slots.

## Changes

- Skip `test-fast-parallel` for `polecat/*` refs only (formula self-review + CI already gate that work). Ownership guard still runs.
  - **What guarantees CI sees every `polecat/*` push:** nothing on the push side — upstream CI has no `push` trigger for `polecat/**` (only `pull_request`, and `fork-verify` on `feat/**`/`fix/**`). The guarantee is on the *landing* side: a `polecat/*` ref is an integration-only branch that reaches `main` exclusively as a PR head, where the full `pull_request` CI graph runs. If a `polecat/*` ref could ever be merged or fast-forwarded into a CI-bearing branch without a PR, the skip would be wider than its argument; the hook does not enforce that, the merge policy does. Happy to tighten the skip (e.g. only when the push remote is not the canonical repo) if you'd rather it be enforced in the hook.
- Do not `exec make`; map exit 141 to a named fail-closed error. Drop git's hook stdin before the suite.
- Ignore SIGPIPE during `test-local-parallel` EXIT cleanup so a leftover writer cannot convert a passing suite into 141.
- When a push-gate slot's stamped PID is dead, acquire reaps leaked descendants (`lsof`/`fuser` + kill) instead of waiting 600s for an operator.

## Test plan

- `bash scripts/test-push-gate-lock.sh` (56/56)
- `bash scripts/test-push-ownership-guard.sh` (42/42)
- `go test ./scripts/ -run 'TestPrePushUsesCanonicalMachineAwareConcurrency|TestPushOwnershipGuard'`

Do not merge to main from the fork default; this PR targets gastownhall/gascity `main`.


sjarmak (changes requested):
The pre-push work looks like a real fix for a real loop, and I want it. The problem is that it is not the only thing in this branch, and the two things riding along with it are not small.

`git log main..753313442` has three commits:

```
cf8329400 fix(bd): reach the wisps tier on gc bd list --type=molecule
e5e729f73 feat(packs): run pack lifecycle hooks on city start and stop
753313442 fix(pre-push): fail-close SIGPIPE and reap leaked push-gate slots (ga-34a)
```

The title and the body describe only the third. The second adds `internal/packlifecycle/` (a package that does not exist on main today), wires it into `gc start` and `gc stop`, and changes `docs/reference/specs/pack-spec.md`. A new package plus a spec change is a new public contract, and right now it would land under a title that says "pre-push hook." Anyone reading the merge log later, or bisecting a pack-lifecycle regression, would have no reason to look here. The first commit is smaller but is also a `gc bd` behavior change with nothing about it in the body.

Two things I would like, and I will merge the hook fix as soon as the first one is done:

1. **Split the branch.** Keep `753313442` here so this PR is what it says it is. Open `e5e729f73` as its own PR: the pack-lifecycle hook contract deserves its own review, and the spec change means it wants Julian's eyes specifically, which it will not get buried under this title. `cf8329400` can go in a third, or fold into whichever PR its `gc bd` change actually belongs to.

2. **The census.** `Integration / packages-core-4-of-4` is red on `TestCatalogMatchesProductionWiringAndDocumentation`, which is command-census drift from the new commands. Worth flagging how to resolve it, because the obvious move is the wrong one: `cmd/gc/productmetrics_command_census.json` is the hand-maintained manifest, not a generated file, and the other three are derived from it. On a rebase conflict there, taking either side wholesale and regenerating silently drops one side's new commands or leaves colliding ids, because both sides consume `next_id` from their own base. Merge the two `commands` arrays by hand, renumber the newer entries from the current post-merge `next_id`, keep the array sorted by `path`, then run `go run ./cmd/gen-command-census` and `--check`.

On the hook change itself, three things I checked and liked. Mapping 141 to a named error instead of `exec make` is the right shape: the old form could not distinguish "suite passed, a writer died on the way out" from "suite failed," and the exit code was the only channel. Dropping git's hook stdin before the suite removes the writer that produced the 141 in the first place, so the SIGPIPE ignore during EXIT cleanup is belt-and-braces rather than the actual fix, which is the correct order to do those in. And reaping by stamped-PID liveness rather than waiting out a 600s timeout matches how the rest of this repo treats state files: the process table is the source of truth, and a slot whose stamped PID is gone is free by definition, not after a timer.

One question on the skip, not blocking: skipping `test-fast-parallel` for `polecat/*` rests on formula self-review plus CI covering that work. If a polecat ref can ever reach a branch that CI does not run on, the skip is wider than the argument for it. Worth a sentence in the body naming what guarantees CI sees every `polecat/*` push.

Happy to re-review the split PRs quickly.

brandonmartin:
Thanks — agreed on all three, and the split is done.

1. **Branch split.** This PR is now exactly one commit on current `main` (`747c2fb5c`, the pre-push fix — force-pushed earlier today). The two riders have their own PRs:
   - #5677 — `feat(packs): run pack lifecycle hooks on city start and stop`. The `pack-spec.md` change is called out at the top of its body so it gets the pack-spec owner's review specifically.
   - #5678 — `fix(bd): reach the wisps tier on gc bd list --type=molecule`.

2. **Census.** No longer applies to this PR: with only the hook commit here there are no new commands, and `Integration / packages-core-4-of-4` is green on the current head. Noted your merge recipe for `productmetrics_command_census.json` (hand-merge `commands`, renumber from post-merge `next_id`, keep sorted by `path`, then `go run ./cmd/gen-command-census --check`) — that's the recipe we'll use on #5677 if it collides.

3. **The `polecat/*` skip.** Added a paragraph to the body. Short version: nothing on the push side guarantees it — upstream CI has no `push` trigger for `polecat/**`. The guarantee is on the landing side: `polecat/*` refs are integration-only and reach `main` only as a PR head, where the full `pull_request` graph runs. That's merge policy, not something the hook enforces; if you'd rather it be enforced, I'm happy to narrow the skip (e.g. only when the push remote isn't the canonical repo) in a follow-up.

The only remaining red on this head is the shared `TestCmdStopSupervisorManagedInvalidCityTomlFailsWhenShutdownFails` (#5668 / fix in #5667), which fails identically on `main`.

brandonmartin:
Rebased the already-split single-commit branch onto current `upstream/main` (`9700d9a48`) and force-pushed `4e0b1cbbd`. The sole conflict was in the shared push-gate library: resolution keeps upstream’s dynamic per-attempt FD selection and integrates this PR’s leaked-descendant holder discovery/reaping. Fresh verification: 60/60 push-gate-lock cases, 46/46 push-ownership cases, focused Go contract tests, `bash -n`, and range diff check all pass. The prior split/census request remains addressed; ready for rereview.

PR #5632, opened 2026-08-26
test(eventfeed): derive hang-detector waits from a package budget

quad341:
## What this changes

Eventfeed's in-memory MuxSource tests now derive their pure hang-detector waits from one package-level budget. This gives goroutines more scheduling room during fleet-wide CPU contention without slowing normal test runs: every wait still returns immediately when its expected event arrives.

The change is test-only. It does not alter event delivery, cursor behavior, production timeouts, dependencies, or runtime defaults.

## Review notes

- The budget is `6 * testutil.GoroutineRaceTimeout`, matching existing hang-detector precedents.
- Nine references across five wait constructs move to the shared budget: three channel-wait helpers, the rebuild/Next context, and consumer cleanup.
- The converted waits are safety ceilings only; no fed timeout or negative assertion window changes.
- The diff is confined to `internal/eventfeed/muxsource_test.go`.

## Test plan

- [x] Run the full 40-job local CI-equivalent union with the rootless container environment configured.
- [x] Confirm both MuxSource behavioral tests execute and pass in the full-suite output.
- [x] Run the repository policy lane, `go vet ./...`, affected-code lint, and changed-file formatting checks.
- [x] Release gate: [`release-gates/ga-9c55c6-eventfeed-hang-budget-gate.md`](release-gates/ga-9c55c6-eventfeed-hang-budget-gate.md)


sjarmak (approved):
The classification holds and the change is safe to take. One wording note below, not gating.

**Both bans are respected.** The negative assertion in this file is `requireNoFloorSample` (`internal/eventfeed/muxsource_test.go:88`), and the diff does not touch it. It is a non-blocking `select` with a `default:` arm, which is the strongest form the ban can be satisfied in: there is no window to widen because there is no timer at all. The four helper waits (`:76`, `:99`, `:113`) and the consumer-drain cleanup (`:241`) each block until a signal arrives and only then assert on a value, so raising their ceiling cannot change any verdict.

**The precedents are real and the derivation matches them.** `cmd/gc/hangbudget_test.go:45` is `6 * testutil.GoroutineRaceTimeout`, and `internal/storebinding/sqlite/sqlite_fence_process_linux_test.go:62` is `6 * testutil.ExecRaceTimeout` with a comment naming the first as the pattern it mirrors. Same multiplier, same shape.

**Cost of the wider ceiling is zero on a passing run**, measured rather than argued: `TestMuxSource_PreservesInitializedZeroFloorAcrossRebuild` runs in 0.00s and `TestMuxSource_YieldsAndPicksUpNewCity` in 0.02s. Every wait here returns on its condition, so the constant only sets how long a wedged run takes to report.

**The wording note.** Criterion 2 of `release-gates/ga-9c55c6-eventfeed-hang-budget-gate.md:12` lists the rebuild/Next context among the converted sites and then states in the same cell that "No fed timeout ... changed." Those two clauses are in tension as written, because that constant is fed into `context.WithTimeout` at `muxsource_test.go:139` and the production stack under test does observe the resulting deadline: `rebuild` derives a child from it at `internal/eventfeed/muxsource.go:142`, `Multiplexer.Watch` propagates it into provider watches, and both watcher implementations return `ctx.Err()` when it fires (`internal/events/multiplexer.go:402`, `internal/events/fake.go:144`). With `rebuildEvery` set to `time.Hour` at `muxsource_test.go:136`, the outer constant is the effective bound.

I still think the site is correctly classified. `TESTING.md:1458-1460` bans migrating a fed timeout because "a timeout passed *into* the code under test defines the scenario being exercised" and widening it "makes the test prove less." Here the scenario is defined by `rebuildEvery`, not by the parent deadline; nothing in the test asserts on the deadline firing, and no assertion gets weaker at 6x. So the substance is right and the gate's list of converted sites is accurate and complete. What is worth correcting is the flat "No fed timeout ... changed," because a later audit reading that cell alone could cite this gate as precedent for migrating a genuine scenario-defining timeout, which is the thing the rule exists to prevent. A clause naming the context and saying why it is patience rather than scenario would carry the same verdict and be defensible on re-read.

Not holding the merge on it. It is an evidence-record precision fix in an artifact you signed, so it is yours to amend rather than mine, and the code it describes is correct either way.

The red on this branch is not diff-owned, and the gate already attributes it: `TestCatalogMatchesProductionWiringAndDocumentation` was failing fleet-wide because eight `runtime.builtin.*` provider-ledger waivers expired on 2026-08-26. That is fixed on main now, so this needs a base update rather than a rerun (a rerun reuses the original merge commit and will keep failing).

Codex (gpt-5.4) reviewed independently and reached the same conclusion on the four other questions: no converted site is a negative assertion window, every remaining wait-bearing site in the file is on the budget, both precedents check out, and its one finding is the classification point above.

PR #5633, opened 2026-08-26
fix(cli): federate the relocated class binding into the CLI read fan-outs and workflow sweeps

julianknutsen:
Stacked on #5626 (which stacks on #5597). **Base: `residency/pr1-storeref-byid`** — GitHub retargets as each parent merges.

On a converged city, `gc beads list`, `gc convoy list`, `gc convoy check`,
`gc workflow delete`, `delete-source` and `reopen-source` all enumerate the
city's and rigs' **directories**. A relocated class binding is not one of them.
So every one of those commands worked on the copies `gc storage migrate`
retained — frozen at cutover, never touched since — and reported success.

## The three shapes of the same bug

**Reads print the frozen twin as live.** The rows the city has gone on mutating
are invisible and the retained copies are printed in their place, under the same
ids. `gc convoy check` is the one that bites: it decides whether a convoy is
finished from whichever end it can see.

**Destructive sweeps report a partial as a success.** `gc workflow delete` closes
the frozen twin, reports the count, exits 0 — and the tree the city is actually
running stays live in the binding. An operator who ran `delete` to stop a
workflow has been told it stopped. `reopen-source` is worse: it decides *"is this
source's workflow still live"* from the same blind walk, finds nothing, and
re-slings the bead underneath a workflow that never stopped.

**A store that could not answer was read as a store with nothing to say.** Three
separate holes fed this. `findWorkflowBeads` swallowed every `Get`/`List` error
and returned whatever it had collected. `closeWorkflowMatches` dropped
`CloseAll`'s error on the floor and reported the count it *got back* as the count
it *closed*. `recordScanFailure` aborted only for the **selected** store, and the
binding's `rootStoreRef` is the city's — so a `--rig` selector, or a recursion leg
carrying a rig ref, demoted a faulting binding to a warning.

## What changed

All three refusal paths now route through `refusePartialSweep`, the choke point
the standing refusal already used, so the callers cannot drift apart on policy:
one message shape, one exit code, the store named every time. A standing refusal
is a verdict about storage config that arrives at federation time; a read that
faults arrives mid-scan and had no interception at all. The federation guard's
name and doc said *"unreadable"* while covering only the first — which is exactly
why the second went unnoticed.

**Read fan-outs warn and continue; destructive one-shots hard-error.** That is
the one deliberate policy difference. A read prints what it reached and names
what it could not. A destructive one-shot has no such output, and *"I could not
see the binding's tree"* and *"the binding's tree is gone"* leave the operator
believing the same thing.

**Ask the binding which ids it OWNS, not which it MATCHED.** The supersede set
was read off the binding's rows *for the caller's query*, which makes the answer
depend on a filter the caller chose: `gc convoy list` asks every store for its
**open** convoys, so the moment the city closes a relocated convoy in the binding,
the binding stops answering for that id, the retained copy stops looking
superseded, and the frozen twin prints as live. That is not an edge case — a
closed relocated bead is the steady state a migrated city converges to.
`convoyBindingOwnedIDs` now carries only the migration source's ids and
`IncludeClosed`, so no filter of the caller's reaches it, and the role scoping
becomes structural rather than incidental: a rig's id is never put to the binding
at all, so two stores that merely mint the same id both survive.

**Five things had to move with the extra view**, each of which would otherwise
have turned the fix into its own silent partial: `scopePath` (the binding has no
directory and its rows are the *city's*); the collector's visit key moves from
the store **ref** to the **view** (binding and work ledger share one ref, so
ref-keying marked the pair visited after the first and left the second unswept);
the multi-store guard counts distinct **scopes**, not labels (labels would refuse
`delete-source` on every migrated city); the `bd` runner is nil for the binding
and its beads are deleted through the store handle (`bd` runs *in* a directory,
and there is none here); and `findUniqueBeadAcrossStoresView` leads with the
binding leg, because that view is the one `delete-source`/`reopen-source`
**write** the source bead's metadata through — clearing `workflow_id` on the
frozen twin leaves the live row pointing at a workflow that no longer exists.

**The cross-store id refusal is restored on a binding hit.**
`resolveOwningStoreDir` short-circuited to the binding and returned before the
directory scan — and the scan is where *"this id exists in more than one store"*
is refused. Skipping it for the city's own store is correct (the retained copy is
a frozen twin, not a second bead); a **rig** is different, since it is never a
migration target, so the same id in a rig is a genuinely different bead and
picking a winner silently is the `ga-axin6` shape.

**The source-workflow lock keys on the view's SCOPE, not its display path.**
`lockIdentity` hashes the ref into a filename, so `"city (class binding)"` is a
valid key that simply is not the city's — mutual exclusion was weakened silently
rather than failing.

## Verification

- `go build ./...`, `go vet ./...` clean
- `go test ./cmd/gc/ -timeout 40m` green (the default 10m timeout **panics** this
  package)
- **Mutation-proved with disjoint kills.** Dropping the binding view from the
  walk while keeping the refusal kills the three sweep/gate rows and leaves both
  refusal rows green; making the refusal warn-and-continue kills only the two
  refusal rows.
- The final commit exists because **five behaviours the CLI already relied on
  survived every mutation** — a pin that stays green when you break the thing it
  names is documentation, not a test. Each new row was written by first breaking
  the production code and watching the intended test die. The `--delete` arm on a
  converged city had **no coverage at all**; the supersede rule's scope needed a
  rig in the fixtures (every relocated fixture had exactly one non-binding store,
  so "drop every id the binding owns" and "drop only the migration source's"
  produced identical output); `distinctSourceWorkflowScopes` was unreachable
  except through an **orphaned** source bead; and the lock-scope pin asserts on
  the lock **files** under the city's runtime dir, which is what distinguishes a
  correct key from a valid but different one.

Refs `ga-gqc9e`.

---

## Council round 2 — the arms the first pass could not falsify

A three-lens fable council reviewed the branch and returned changes-required.
Production behavior was judged correct throughout; what it found were coverage
holes, one reproduced scale cliff, and two mis-mapped evidence claims. All are
closed in `382d5576fc`.

**The probe did not scale past 32766 ids.** It put one SQL placeholder per
candidate on the wire, and the candidates are every migration-source row matching
the caller's query. `gc beads list` runs unbounded, so a converged city with a
long history sent its whole work ledger in one IN-list and modernc sqlite refused
the statement past 32766 bound variables — reproduced against a real sqlite
binding: 32766 ok, 40000 and 100000 `SQL logic error: too many SQL variables`.
The read failed *loud*, so no wrong answer was served, but it failed on exactly
the large migrated cities this federation exists for. The probe now chunks under
the cap, keeping it O(candidates/batch) round trips rather than the one-per-row
it replaced.

**Three arms were unfalsifiable and are now pinned.** Each was confirmed
unfalsifiable by mutating production and watching the **full** `./cmd/gc/` suite
stay green.

- *The probe's fault path.* Every relocated-city read fixture served a binding
  that answers both the caller's query and the ownership probe, so a merge that
  read a probe fault as *"the binding holds none of these"* produced
  byte-identical output — and that misreading serves every frozen copy the
  migration retained as a live row. The new row faults only on a query carrying
  `IDs`.
- *`verifyWorkflowMatchesClosed`.* No fixture held a store that reports a close
  count it did not honour, so the whole verification pass and all four of its
  branches were unexercised — deleting the function outright left the suite
  green. The lying store is what the pass can now be told apart from.
- *`sweepOrder` on the CLOSE arm.* Closing the frozen twins while the live tree
  keeps running is the partial sweep that *looks* finished, so the close arm
  needs the binding-first guarantee at least as much as `--delete` does.

**Two evidence corrections.** The delete arm's ordering evidence was mapped to
the wrong row — with every store healthy the sweep reaches all of them whichever
way round it goes, so the order is falsified by the *refusal* row, not the
erasure row. The comments now say so. And the read caveat understated its own
hazard: a standing refusal does not merely omit the binding's rows — the
binding's answer is what **supersedes** the retained pre-migration copies, so
those copies print in their place as stale open rows for as long as the refusal
stands.

Filed as follow-ups rather than fixed here: `ga-4kivg` (P1, `delete-source` with
an explicit selector clears `workflow_id` on the wrong copy) and the
`openConvoyStores` silent-drop and delete-source sweep-order nits, both
pre-existing at the base commit.

### DIFF
--- a/cmd/gc/by_id_binding_owner_test.go
+++ b/cmd/gc/by_id_binding_owner_test.go
@@ -499,3 +499,134 @@ func TestByIDPlanUsesTheRegisteredControllerRoutes(t *testing.T) {
 		})
 	}
 }
+
+// bindingHitCity builds the fixture the binding short-circuit runs on: a city
+// whose infrastructure classes are served by one store, plus a rig, plus a
+// per-directory store opener the caller controls.
+//
+// The binding is a work-PREFIXED store on purpose. A binding that minted
+// reserved ids would retire the residence probe, and then a work-shaped id
+// would never reach the binding leg at all — which is the fixture testing
+// itself rather than the resolver.
+func bindingHitCity(t *testing.T) (cityPath, rigPath string, cfg *config.City, binding beads.Store) {
+	t.Helper()
+	cityPath = t.TempDir()
+	binding = splittest.NewWorkStore(t, "hq")
+	seedCLIStorageRoutes(t, cityPath, messagingSplitRoutes(binding))
+	rigPath = filepath.Join(cityPath, "rigs", "alpha")
+	cfg = &config.City{Rigs: []config.Rig{{Name: "alpha", Path: rigPath}}}
+	return cityPath, rigPath, cfg, binding
+}
+
+// storesByDir is an openStore that hands each directory the store the test
+// prepared for it, and an empty one for any directory it did not name.
+func storesByDir(t *testing.T, byDir map[string]beads.Store) func(string) (beads.Store, error) {
+	t.Helper()
+	return func(dir string) (beads.Store, error) {
+		if store, ok := byDir[dir]; ok {
+			return store, nil
+		}
+		return splittest.NewWorkStore(t, "hq"), nil
+	}
+}
+
+// TestResolveOwningStoreDirRefusesBindingRigCollision is the ga-qnagn
+// regression.
+//
+// The binding leg short-circuits the scan, and with it the scan's uniqueness
+// refusal. That is right for the city's own retained copy — dual residency is
+// the migration working — but it is not right for a RIG, which is never a
+// migration target and so has no retained copy to be excused. A rig holding the
+// same id is two ledgers disagreeing by accident, exactly what the uniqueness
+// contract exists for, and answering it silently from the binding means the
+// close that follows writes one copy while the other stays open forever.
+func TestResolveOwningStoreDirRefusesBindingRigCollision(t *testing.T) {
+	cityPath, rigPath, cfg, binding := bindingHitCity(t)
+	resident, err := binding.Create(beads.Bead{Title: "the binding's copy", Type: "task"})
+	if err != nil {
+		t.Fatalf("seeding the binding: %v", err)
+	}
+	rig := splittest.NewWorkStore(t, "hq")
+	collision, err := rig.Create(beads.Bead{Title: "a rig copy under the same id", Type: "task"})
+	if err != nil {
+		t.Fatalf("seeding the rig: %v", err)
+	}
+	if collision.ID != resident.ID {
+		t.Fatalf("the fixture minted %s and %s; a collision needs one id", resident.ID, collision.ID)
+	}
+
+	_, _, err = resolveOwningStoreDir(resident.ID, cfg, cityPath, storesByDir(t, map[string]beads.Store{rigPath: rig}))
+	if err == nil {
+		t.Fatal("a binding/rig collision resolved cleanly; the caller would close one copy and leave the other open")
+	}
+	if !strings.Contains(err.Error(), "exists in multiple stores") {
+		t.Errorf("the refusal reads %v, want the scan's own uniqueness wording", err)
+	}
+	if !strings.Contains(err.Error(), rigPath) {
+		t.Errorf("the refusal %v does not name the rig store that collides", err)
+	}
+}
+
+// TestResolveOwningStoreDirBindingWinsOverRetainedCityCopy is the control for
+// the test above, and the reason its probe skips the city store.
+//
+// The city store is where the migration RETAINED its copies, so it holds the
+// same id by design on every converged city. A probe that counted it would
+// refuse every convoy command on exactly the cities that finished migrating —
+// the failure PR1's short-circuit was added to prevent.
+func TestResolveOwningStoreDirBindingWinsOverRetainedCityCopy(t *testing.T) {
+	cityPath, _, cfg, binding := bindingHitCity(t)
+	resident, err := binding.Create(beads.Bead{Title: "the binding's copy", Type: "task"})
+	if err != nil {
+		t.Fatalf("seeding the binding: %v", err)
+	}
+	city := splittest.NewWorkStore(t, "hq")
+	retained, err := city.Create(beads.Bead{Title: "the copy the migration retained", Type: "task"})
+	if err != nil {
+		t.Fatalf("seeding the retained city copy: %v", err)
+	}
+	if retained.ID != resident.ID {
+		t.Fatalf("the fixture minted %s and %s; dual residency needs one id", resident.ID, retained.ID)
+	}
+
+	store, dir, err := resolveOwningStoreDir(resident.ID, cfg, cityPath, storesByDir(t, map[string]beads.Store{cityPath: city}))
+	if err != nil {
+		t.Fatalf("a dual-resident id resolved to %v; the retained city copy is the migration working, not a collision", err)
+	}
+	if store != binding {
+		t.Errorf("the resolver returned %p, want the binding %p", store, binding)
+	}
+	if dir != cityPath {
+		t.Errorf("the resolver reported dir %q, want the city path %q", dir, cityPath)
+	}
+}
+
+// TestResolveOwningStoreDirSkipsTheRigProbeForAReservedID pins the probe's
+// other bound.
+//
+// A class-reserved prefix is minted by the binding and by nothing else, so a
+// rig cannot hold one legitimately and there is no collision to find. Probing
+// anyway would add a full rig walk to every reserved-id resolution — including
+// bd's on-close hook, which runs in bursts — to learn nothing.
+func TestResolveOwningStoreDirSkipsTheRigProbeForAReservedID(t *testing.T) {
+	cityPath, rigPath, cfg, binding := bindingHitCity(t)
+	reserved, err := migrationSeed(binding, beads.Bead{ID: "gcg-1", Title: "a graph-class bead", Type: "task"})
+	if err != nil {
+		t.Fatalf("seeding the reserved-prefix bead: %v", err)
+	}
+	if !bdIDIsClassReserved(reserved.ID) {
+		t.Fatalf("the fixture id %q carries no reserved class prefix", reserved.ID)
+	}
+	rig := splittest.NewWorkStore(t, "hq")
+	if _, err := migrationSeed(rig, beads.Bead{ID: reserved.ID, Title: "a rig copy the probe must not consult", Type: "task"}); err != nil {
+		t.Fatalf("seeding the rig: %v", err)
+	}
+
+	store, _, err := resolveOwningStoreDir(reserved.ID, cfg, cityPath, storesByDir(t, map[string]beads.Store{rigPath: rig}))
+	if err != nil {
+		t.Fatalf("a reserved-prefix id resolved to %v; no rig can own one, so there is nothing to refuse", err)
+	}
+	if store != binding {
+		t.Errorf("the resolver returned %p, want the binding %p", store, binding)
+	}
+}
--- a/cmd/gc/by_id_binding_owner_test.go
+++ b/cmd/gc/by_id_binding_owner_test.go
@@ -480,15 +480,26 @@ func TestByIDPlanUsesTheRegisteredControllerRoutes(t *testing.T) {
 				t.Errorf("the by-id plan resolved %s through %p, want %p", resident.ID, owner.Store, wantStore)
 			}
 
-			// The production caller, with a scan that fails the test if it runs:
-			// a binding hit must return before the directory scan, so the funnel
-			// is never reached by this arm either.
+			// The production caller. The scan's opener DOES run on a binding
+			// hit now — refuseBindingRigCollision probes the scan's candidates
+			// so a rig holding the same id is refused rather than silently
+			// losing to the binding — so what this pins is narrower than "the
+			// scan never runs": the probe stays inside the directories the scan
+			// would have walked, and the binding is still what answers. The
+			// binding is not one of those directories, so no probe of it can be
+			// planned and the funnel's errStore is never reached.
+			var probed []string
 			store, dir, err := resolveOwningStoreDir(resident.ID, nil, cityPath, func(storeDir string) (beads.Store, error) {
-				t.Errorf("the convoy scan opened %q for an id the binding owns", storeDir)
-				return nil, errors.New("the scan must not run for a binding hit")
+				probed = append(probed, storeDir)
+				return splittest.NewWorkStore(t, "hq"), nil
 			})
 			if err != nil {
 				t.Fatalf("resolveOwningStoreDir(%s): %v", resident.ID, err)
+			}
+			for _, storeDir := range probed {
+				if !samePath(storeDir, cityPath) {
+					t.Errorf("the collision probe opened %q; this city configures no rigs, so its own directory is the only candidate", storeDir)
+				}
 			}
 			if store != wantStore {
 				t.Errorf("the convoy resolver served %s from %p, want the binding %p", resident.ID, store, wantStore)
--- a/cmd/gc/cmd_bd_by_id_test.go
+++ b/cmd/gc/cmd_bd_by_id_test.go
@@ -152,22 +152,22 @@ func foreignProviderCity(t *testing.T) (cityPath string, classStore beads.Store)
 	return cityPath, store
 }
 
-// cliSoleClassBindingStore resolves the city's sole relocated class binding as
-// a STORE, for fixtures that seed both sides of a migration and have no error
-// channel of their own.
+// soleClassBindingStore resolves the city's sole relocated class binding as a
+// STORE, for the fixtures that seed both sides of a migration.
 //
-// The fan-out — the only error cliSoleClassBinding returns — is a topology this
+// A fan-out — the only error cliSoleClassBinding returns — is a topology this
 // build refuses to serve, so a fixture that met one has not built the city it
 // meant to build and says so here rather than seeding half of it.
-func cliSoleClassBindingStore(cityPath string) (beads.Store, bool) {
+func soleClassBindingStore(t *testing.T, cityPath string) beads.Store {
+	t.Helper()
 	binding, relocated, err := cliSoleClassBinding(cityPath)
 	if err != nil {
-		return refusedClassStore{err: standingStorageRefusal{err: err}}, true
+		t.Fatalf("resolving the city's class binding: %v", err)
 	}
 	if !relocated {
-		return nil, false
+		t.Fatal("the fixture city resolved no class binding; it is not split")
 	}
-	return binding.Store, true
+	return binding.Store
 }
 
 // recensusAfterSeedingARelic reopens the funnel so the binding's relic census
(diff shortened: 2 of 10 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277237779

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277243882

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277243951

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277244020

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR federates the relocated
class binding into the CLI read fan-outs (`gc beads list`, `gc convoy list`,
`gc convoy check`) and the workflow `delete` / `delete-source` sweeps, so that
on a converged city — where the class binding has been relocated — read
commands collapse only the migration twin rather than live rows, and the
destructive sweeps refuse to complete partially instead of orphaning a source.
That restores correct dedup-by-ownership and unique by-id addressability after
a binding relocation, which is exactly the failure mode that made stale frozen
relics resurface as if live.

This PR was reviewed and adopted with a small maintainer fix pushed directly to
the PR branch (the two non-gating consistency items the review teed up).

### Original PR Review
Decision: approve.

Specific gaps validated:
- Read fan-outs federate the class binding and collapse only the migration
  twin — ownership is asked of the binding directly (`convoyBindingOwnedIDs`,
  probed with `IncludeClosed:true`), not inferred from the filtered result, so
  status/filter skew and cross-store id reuse are both handled.
- The ownership probe is a targeted `IDs` filter (it counts as a filter, never
  a silent full-store scan) and batches under the sqlite bound-variable ceiling.
- A probe fault is treated as an error, not as "owns nothing", so a mid-probe
  read fault cannot resurrect every frozen twin as live.
- The read/sweep policy split is deliberate and correct: reads continue on a
  binding refusal (printing a per-run stale-twin caveat and serving the
  reachable stores), while `gc workflow delete` / `delete-source` abort on a
  partial sweep and verify closure, with binding-first `sweepOrder` on the
  abort-on-error delete path.
- By-id mutation refuses a binding⇄rig collision, restoring unique
  addressability; reserved-class ids skip the rig walk, and a read fault during
  the walk is treated as a refusal, never a false "no collision".

Review findings addressed:
- F1 (readability) and OR1 (open risk) were landed in commit `8f43ec0577`:
  `countOpenMatchedBeads` now routes its mid-verify read fault through
  `refusePartialSweep` (matching its delete-path sibling
  `verifyWorkflowMatchesClosed`), so the `delete` / `delete-source` pair report
  a partial sweep with the same "part of the workflow is left untouched"
  framing and store label instead of a bare store error; the commit also
  documents why the `delete-source` apply loop must attempt every match before
  verifying, so a future early `break`/`return` cannot silently reintroduce the
  orphan-source hazard. Neither change alters behavior.

- `ga-m57jw` — OR2 (change-impact / performance, minor): a non-reserved,
  binding-resident id pays a full rig `Get` fan-out on every by-id resolution
  (the close-burst hot path). Bounded and safe as written; only matters if that
  id population grows. Not addressed in this PR.
- `ga-urasd` — OR3 (resource / state lifecycle, minor): while the class binding
  is refused, `gc convoy check` can auto-close or write to a frozen relic
  convoy. Low-impact and self-correcting once the binding is readable again.
  Not addressed in this PR.

### Maintainer Changes
One maintainer commit was pushed to the PR branch:
`8f43ec0577 style(cli): route delete-source verify fault through refusePartialSweep`
— 1 file changed, 7 insertions(+), 1 deletion(-). It lands the two non-gating
consistency items from the review (F1 + OR1) and does not change behavior. All
substantive commits remain contributor-authored.

### Final Review Status
Ready for the merge queue: final head `8f43ec0577` has passing required GitHub

CI: https://github.com/gastownhall/gascity/actions/runs/33076596873

### Review Iterations
One review pass, approved with no gating findings (0 fix iterations). The two
non-gating consistency items were landed as the maintainer polish commit above;
the two remaining low-severity open risks are filed as the follow-up beads
listed above.

---

julianknutsen:
## Rebased onto `main`

Head is now **37a5cf336b1d15ca52a3985a5fca8522dc29f51e**, cut fresh from `origin/main` at `faf4e3b89d` (#5923) and force-pushed over the old head `8f43ec0577`. Base retargeted from `residency/pr1-storeref-byid` to `main`; the dead `status/review-failed` and `status/merge-ready` labels are removed.

The old head carried 75 commits, 66 of them inherited from the unlanded s3..s8/pr1 chain whose s1/s2 segments have since landed on `main` as squashes (#5640, #5641). That lineage is what made the PR show 60+ conflicts against `main`; none of it was this PR's own work. Only this PR's 9 commits were replayed.

### Commits carried (all cherry-picked with `-x`, in order)

| original | subject |
| --- | --- |
| `a8d48469be` | fix(cli): federate the class binding into the CLI read fan-outs |
| `17a086c8e7` | fix(cli): restore the cross-store id refusal on a binding hit |
| `5c5ab9cdf3` | fix(cli): federate the workflow sweep over a relocated class binding |
| `c18c70c797` | fix(cli): ask the class binding which ids it owns, not which it matched |
| `ab7408068d` | fix(cli): refuse the destructive sweeps when a store could not answer |
| `2faa5037c0` | test(cli): pin the binding arms the suite could not falsify |
| `1753c2fa2d` | fix(cli): batch the ownership probe and pin the arms it could not reach |
| `692e862580` | style(gc): use the spelling misspell enforces |
| `8f43ec0577` | style(cli): route delete-source verify fault through refusePartialSweep |

Two rebase commits sit on top; both are described below.

### Was pr1 (#5626) needed? No.

Verified by building, not by inspection: `go build ./...` and `go vet ./...` are clean with none of pr1's chain applied. No pr1 symbol is referenced — `byIDBindingOwnerForTopology`, `BuildBindings` and `KnownLegacyResidents` appear nowhere in this diff. `cliByIDBindingOwner`, `beadForOwner`, `bdIDIsClassReserved`, `refusedClassStore`, `standingStorageRefusal` and `ListQuery.IDs` all resolve on `main` today.

What this PR *did* depend on was **s3 (#5642)**, and only for two seams, both handled without stubbing anything:

1. **`cliSoleClassBindingStore`** (s3's error-channel-less wrapper over `cliSoleClassBinding`). `convoyStoreViewsWithBinding` already returns an error, so it now calls `main`'s `cliSoleClassBinding` directly and marks the fan-out refusal as a `standingStorageRefusal` itself — byte-for-byte the same error value the wrapper would have handed back through the `refusedClassStore` round trip. This also keeps the residency-boundary row honest: the new baseline row is `cmd_convoy.go convoyStoreViewsWithBinding b:cliSoleClassBinding`, which matches `main`'s existing `cliSoleClassBinding\(` pattern with no regex widening, and the census counts the federating surface rather than a wrapper it cannot see.
2. **The fixture seam** (`recensusAfterSeedingARelic`, `dropDerivedResidencyMemo`, and s3's reshaped `classResidentWorkShapedBead`). Ported over the accessors `main` already has, under names that leave `main`'s twenty existing `classResidentWorkShapedBead` call sites alone. When #5642 lands it brings its own copies and these collapse into them.

### Conflicts and resolutions

Three textual conflicts, all in the cherry-pick itself:

1. **`scripts/residency-boundary-patterns.txt`** (prose). The stack's text said six `b:cliSoleClassBinding` sites are baselined; that count includes s3's control-dispatch and graft-class rows, which are not on `main`. Rewritten against `main`'s current text: three baselined (by-id front door, claim-time class route, this PR's CLI fan-out federation), a fourth is a reviewed baseline change. The stack's paragraph about the `(Store)?` spelling is dropped with the wrapper.
2. **`cmd/gc/by_id_binding_owner_test.go`** (both-append at EOF). `main`'s copy of this file came from the s1 squash and carries `TestByIDPlanUsesTheRegisteredControllerRoutes`, which the stack's copy does not. Kept `main`'s file whole and appended this PR's three `resolveOwningStoreDir` rows.
3. **`cmd/gc/cmd_convoy_dispatch_test.go`** (both-append at EOF). Kept `main`'s `TestAssertDrainRootScopeMatchesDispatch` and `TestRunControlDispatcherRejectsCrossScopeDrain` (#5488) and appended this PR's rows. Verified both survive.

`cmd_beads.go`, `cmd_convoy.go`, `cmd_convoy_dispatch.go` and `scripts/residency-boundary-baseline.txt` merged onto `main` with no conflict. The baseline gained only this PR's three rows on top of `main`'s current content — nothing was restored wholesale — plus the "two rows / a third" prose bumped to "three rows / a fourth".

One **semantic** conflict that no merge tool could see, resolved in `37a5cf336b`:

4. **`TestByIDPlanUsesTheRegisteredControllerRoutes` vs `refuseBindingRigCollision`.** That row landed on `main` with #5640 after this branch was cut, and it hands `resolveOwningStoreDir` an opener that fails the test *if it is called at all* — because a binding hit used to return before the directory scan. `refuseBindingRigCollision` is precisely that return being made conditional, so the row failed. What the row protects is not the absence of opens: it is that the funnel never answers (the `errStore` assertion above it) and that nothing re-opens the binding root behind an in-process controller — and the binding is not one of the scan's directories, so no probe of it can be planned. The opener now records what it was asked for, and the row asserts the probe stayed inside the scan's own candidates (with no rigs configured, the city directory alone) and that the binding still answers.

The other rebase commit, `3dbd17e1b9`, carries the fixture seam described above. Its load-bearing part: `failClassBindingReads` and `stubClassBindingStore` now drop the derived residency memo when they swap `routes.stores`, which is the half of #5642 the new rows actually depend on. Swapping in place is invisible to the memo, so a fixture that resolved the binding before installing its failing store — which every relocated-city fixture here does, to seed both sides of a migration — kept handing the *healthy* store to the code under test, and the two standing-refusal sweep rows exited 0 while asserting that they refuse.

### Gates

| gate | result |
| --- | --- |
| `go build ./...` | clean |
| `go vet ./...` | clean |
| `gofmt -l` | clean |
| `go test ./cmd/gc -run 'Convoy\|Beads\|Workflow\|FanOut\|Federat\|Sweep' -count=1` | ok |
| the 15 named falsification rows (ownership probe, SQL-variable-cap batching, sweep refusals, lock scope, collision arms, scope counting) | all PASS |
| every row using the two changed fixtures (12 rows across 4 files) | all PASS |
| `go test ./internal/testpolicy/... ./scripts/...` | ok (no census or boundary-ratchet drift) |
| `make test-fast-parallel` | all fast jobs passed |

`main` advanced two commits (#5316, #5932) while the gates ran; neither touches this diff's files and `git merge-tree` against the newer `main` reports zero conflicts.

Follow-ups from the adoption review stay as filed and are unaffected by the rebase: `ga-m57jw` (rig fan-out on every non-reserved binding hit), `ga-urasd` (`convoy check` writing to a frozen relic while the binding is refused), `ga-4kivg` (`delete-source` with an explicit `--rig`/`--store-ref` pinning metadata clearing to the named view).

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277290957

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277291033

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277295539

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR federates the relocated
class binding into the CLI read fan-outs (`gc beads list`, `gc convoy list`,
`gc convoy stranded`) and the workflow `delete` / `delete-source` sweeps, so a
converged city stops answering from the frozen pre-migration copies it retained
at cutover. That matters because the old behavior failed silently in the worst
possible direction: convoy checks and workflow deletes reported success while
the live rows in the binding went untouched, and a convoy the city was still
working could read as complete because its children were frozen at cutover.
Turning that class of silent wrong answer into either a correct answer or an
explicit refusal is a real improvement to the residency migration story.

branch.

### Original PR Review

Specific gaps fixed:
- `gc convoy check` mutates — it closes convoys — but it was federating through
  the read-tolerant helper. On an unreadable binding that helper hands the views
  back unchanged, so the check would have walked the retained pre-migration
  copies as live rows and auto-closed a convoy the city was still working, then
  reported that as success. It now refuses instead: a mutation on a view the
  build could not prove is worse than no mutation.
- `gc convoy stranded` was left unfederated while its `list`/`check` siblings
  were converted. That was wrong in both directions at once on a migrated city —
  a retained copy still carries the assignees and open children it had at
  cutover and reports work that is not stranded, while a convoy living in the
  binding is never walked, so genuinely stranded work stays invisible.
- The read/mutate split is now documented at the boundary that enforces it, and
  the dead `openAllConvoyStores` helper is gone with no remaining references.
- The residency ratchet (`scripts/check-residency-boundary.sh`) exits clean with
  the census row moved rather than grown.

Review findings addressed:
- **`gc convoy check` federated through the read-tolerant path** (attempt 1,
  blocker) — fixed in `6933c72e`: `doConvoyCheckFallback` now takes
  `convoyStoreViewsWithBinding` and returns `1` on its error, with a
  refusal-case regression test. Review confirmed `routeConvoyCheck` is
  unconditional, so the refusal is the command's only path and there is no
  API/fallback split for it to hide behind.
- **`gc convoy stranded` left unfederated** (attempt 1) — fixed in `6933c72e`:
  the new `doConvoyStrandedFallback` mirrors `doConvoyListFallback` on the
  tolerant helper, which is the right choice for a pure read, plus the missing
  binding-row test.
- **Two new doc comments contradicted the real call graph** (attempt 1) — fixed
  in `6933c72e`. Review then spot-checked the PR's other comparable parity claim
  (`deleteWorkflowMatches`) and confirmed it survives the same check.
- **The binding-view identity gate deferred its safety argument to a function
  documented as production-dead** (attempt 3, minor) — fixed in `6933c72e` by
  naming, at the site, why both operands are pointer-typed.

One note on that last one, since it started life as a blocker: the review filed
a concern that `view.store == binding` could panic on a non-comparable
`beads.Store`. That was checked and it does not hold — a Go interface comparison
only panics when both dynamic types are identical and non-comparable, every
production store on that path is pointer-typed, and the one value-typed route
(`refusedClassStore`) is type-switched out above the comparison. The gate was
kept as written and the reasoning was written down instead, which is the right
outcome: removing it would have deleted the idempotence rule the gate exists to
state.

- `ga-cf02y` (architectural consistency, minor): `verifyWorkflowMatchesClosed`
  re-reads one bead per round trip on the destructive sweep path, while this
  same change set establishes the batched `ListQuery{IDs}` form for exactly that
  question in `convoyBindingOwnedIDs`. Not addressed here — the miss-path
  translation is a real rewrite rather than a mechanical swap, so it wants its
  own change and its own tests.
- `ga-pf34e` (architectural consistency, minor): `countOpenMatchedBeads` carries
  the same per-bead `Get` shape. Pre-existing and untouched in substance by this
  PR, so it was correctly scoped out rather than widened into it.
- `ga-5193r` (test evidence, minor): pin the live classifier arm — a convoy
  carrying only `gc.root_bead_id` and no synthetic marker classifies
  `ClassGraph` today. That arm's existence is currently documented only in a
  comment, and it is the argument for federating the convoy reads at all.
- `ga-lqcq9` (release safety, informational): release note for `gc convoy
  check`'s new nonzero exit while the binding is mid-repair. This is a
  deliberate and correct behavior change, but it is a new failure mode for
  scheduled automation and should be announced.

One scoping note for whoever picks this up next: `ga-efyq4` should **not** be
closed by this merge. Its `gc convoy check` symptom lives in member resolution
(`listConvoyChildren` → `convoycore.Members`), which is untouched here and fails
safe.

### Maintainer Changes
One maintainer fixup commit was pushed to the PR branch:
`6933c72e fix(cli): refuse rather than degrade in convoy check, federate convoy stranded`
— 4 files changed, 175 insertions(+), 19 deletions(-)
(`cmd/gc/cmd_convoy.go`, `cmd/gc/cmd_convoy_test.go`, and the two
`scripts/residency-boundary-*` census files). Contributor authorship on the
original commits is preserved.

### Final Review Status
Ready for the merge queue: final head `6933c72e` has passing required GitHub
checks, and the merge-ready metadata is in place. The adopted branch was also
refreshed onto the current `main` and re-validated there — the patch applies
with an unchanged patch id, and `scripts/check-residency-boundary.sh`,
`go vet ./cmd/gc/`, and the targeted `cmd/gc` convoy/binding test slice all pass
against the newer base, including the two residency census files `main` moved
underneath this branch in the meantime. I am marking this PR

CI: https://github.com/gastownhall/gascity/actions/runs/33761411156

### Review Iterations
3 review passes. Pass 1 requested changes on the `gc convoy check` mutation path
and the unfederated `gc convoy stranded`; pass 2 was a re-derivation on an
unchanged head; pass 3 verified each fix independently and approved, correcting
one earlier claim rather than inheriting it.

---

PR #5635, opened 2026-08-26
fix(dispatch): split gc.attempt into separate iteration and retry counters

julianknutsen:
Fixes `ga-v7pu5` and `ga-la0py`.

## The defect

`gc.attempt` is one metadata key serving **two independent counters**: the ralph *iteration* index and a step's own *retry attempt* index. The ralph expansion overwrites the second with the first.

That is not cosmetic. `processRetryEval` reads `gc.attempt` off a child and computes `nextAttempt = attempt + 1`, so a child first run in iteration N is **born at attempt N** against its own `max_attempts`. By iteration 3 it is exhausted before it runs and hard-fails having never retried.

Census of the maintainer-city graph store:

- **81 of 382** retry-kind beads (21%) born at or past their own max — 40 at attempt 4 and 10 at attempt 5 against a max of 3, values a counter starting at 1 can never reach
- **177 beads** whose `gc.step_ref` literally ends in `.attempt.1` while `gc.attempt` claims 2–5
- All **2086** ralph-iteration beads storewide came from the recipe path (`buildAttemptRecipe`); zero carry `gc.retry_from`, so the clone path has never run here

## Commit 1 — split the counters

`gc.iteration` has existed in `beadmeta` since the runs view started reading it (`internal/runproj/detail_nodeshape.go:33`) but **nothing ever wrote it**. It does now.

`RalphBodyChildAttempt` decides what a body child's attempt number means: a child whose spec already carries one keeps it, a child owning a retry/ralph control starts at 1, and a plain body step still inherits the iteration. Both the compile-time expansion and the runtime re-spawn route through it, so iteration 1 and iterations 2+ cannot disagree.

**This also had to complete S38.** `controlIdentitySet` included the control's *bare* `gc.step_id`, which names the same inner control in **every** outer iteration. Only `max(gc.attempt)` kept them apart — and that ordering existed *only because* the iteration index was being stamped as the attempt number. Splitting the counters removes that accidental tiebreaker, so identity is now ranked explicitly: precise (`bead ID`, `gc.step_ref`) outranks bare (`gc.step_id`), with the infrastructure filter applied *before* the ranking so a scope-check can't starve a real attempt root of its fallback.

## Commit 2 — give the artifact gate an `{iteration}` placeholder

The gate resolved `{attempt}` from the same conflated key, making it **structurally incapable** of catching the mismatch: it computed the same wrong directory the failing step computed, agreed with it, and passed.

The gate is **live, not inert** — 471 of 542 template-carrying beads (87%) do resolve a worktree, and the "root not resolvable in this store" bucket is **zero**. A gate pointed at a directory nobody wrote turns passing attempts into burned retries.

Also fixes the worktree read: three sites read only the legacy `work_dir`, never canonical `gc.work_dir`. `beadmeta` documents this family's contract as canonical-then-legacy and `internal/beads/contract` implements exactly that for the siblings; this site predates both.

A **missing** `gc.iteration` deliberately fails loud via the existing `unresolved_required_artifact_template` check. Substituting empty would build `attempt-/` and the gate would then report `missing_required_artifact` — blaming the step for the resolver's own gap.

## Deploy order — this does NOT ship alone

**`gastownhall/workflows#35` must merge first.** The adopt-pr gate scripts join sibling beads on `gc.attempt`, and `apply-fixes` declares `steps.children.retry max_attempts = 3`. After the split its beads carry their own retry counter, `load_verdict` matches nothing, returns `"iterate"`, and the ralph loop **never approves** — an infinite review loop. That PR is a strict no-op against today's binary, so it can and should land first.

Then: deploy this binary. *Then* switch `attempt-{attempt}` → `attempt-{iteration}` in `formulas/expansion-review-pr.toml` — never before, or an older binary leaves `{iteration}` unresolved and sends every attempt transient forever.

## Tests

TDD throughout; every assertion was seen RED first. New: `internal/dispatch/control_attempt_iteration_test.go`, `internal/dispatch/retry_artifact_iteration_test.go`, `internal/formula/ralph_iteration_test.go`.

`go build ./...`, `go vet` clean; `internal/{dispatch,formula,molecule,runproj}` green; pre-commit lint 0 issues.

> **Note on CI:** `TestCatalogMatchesProductionWiringAndDocumentation` is red repo-wide today — the `ga-80po0c.3` provider waivers expired 2026-08-26. Proven pre-existing against a pristine base with a clean `git status`. Fixed separately in `fix/renew-runtime-provider-waivers`.

## Known dormant twin

`appendRalphRetry` has the same conflation and would carry a stale `gc.iteration` through `cloneMetadata`. Proven dormant (zero `gc.kind=check` and zero `gc.retry_from` beads storewide), so it is filed as `ga-4dt92` (P3) rather than expanded into a verified commit as untestable speculative work.

### DIFF
--- a/internal/dispatch/control.go
+++ b/internal/dispatch/control.go
@@ -810,6 +810,18 @@ func qualifyAttemptTargetWithSourceRoute(target, sourceRoute string, cfg *config
 	return target
 }
 
+// setIterationMetadata records gc.iteration, or removes it when the step has no
+// iteration to speak of. A plain retry outside any loop is the second case, and
+// leaving a stale key there would name a directory that belongs to some other
+// molecule's loop.
+func setIterationMetadata(meta map[string]string, iteration string) {
+	if iteration == "" {
+		delete(meta, beadmeta.IterationMetadataKey)
+		return
+	}
+	meta[beadmeta.IterationMetadataKey] = iteration
+}
+
 // buildAttemptRecipe constructs a minimal formula.Recipe for one attempt
 // from the frozen step spec.
 func buildAttemptRecipe(step *formula.Step, control beads.Bead, attemptNum int) *formula.Recipe {
@@ -830,6 +842,17 @@ func buildAttemptRecipe(step *formula.Step, control beads.Bead, attemptNum int) 
 		attemptPrefix = fmt.Sprintf("%s.iteration.%d", stepRef, attemptNum)
 	} else {
 		attemptPrefix = fmt.Sprintf("%s.attempt.%d", stepRef, attemptNum)
+	}
+
+	// Which loop iteration this sub-DAG belongs to, tracked separately from
+	// gc.attempt. A ralph step's attempts ARE its iterations, so it defines the
+	// value; every other step inherits it from its control, which is the only
+	// way a retry nested inside a ralph body learns which iteration it is
+	// retrying within. Derived from the live control chain and never from the
+	// frozen spec, so a stale copy in step.Metadata cannot survive.
+	iteration := strings.TrimSpace(control.Metadata[beadmeta.IterationMetadataKey])
+	if step.Ralph != nil {
+		iteration = strconv.Itoa(attemptNum)
 	}
 
 	// Root step for the attempt sub-DAG.
@@ -855,6 +878,7 @@ func buildAttemptRecipe(step *formula.Step, control beads.Bead, attemptNum int) 
 	// (buildNestedControlSeed) — both are covered by findLatestAttempt's
 	// identity set.
 	rootMeta[beadmeta.ControlForMetadataKey] = control.ID
+	setIterationMetadata(rootMeta, iteration)
 	if step.OnComplete != nil {
 		rootMeta[beadmeta.OutputJSONRequiredMetadataKey] = "true"
 	}
@@ -907,8 +931,12 @@ func buildAttemptRecipe(step *formula.Step, control beads.Bead, attemptNum int) 
 
 		for _, child := range step.Children {
 			childID := attemptPrefix + "." + child.ID
+			childAttempt := strconv.Itoa(attemptNum)
+			if step.Ralph != nil {
+				childAttempt = formula.RalphBodyChildAttempt(child, attemptNum)
+			}
 			childMeta := map[string]string{
-				beadmeta.AttemptMetadataKey:     strconv.Itoa(attemptNum),
+				beadmeta.AttemptMetadataKey:     childAttempt,
 				beadmeta.StepRefMetadataKey:     childID,
 				beadmeta.StepIDMetadataKey:      child.ID,
 				beadmeta.ScopeRefMetadataKey:    attemptPrefix,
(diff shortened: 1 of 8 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277237886

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277242278

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR splits the overloaded `gc.attempt` metadata counter into a shared outer-loop `gc.iteration` index and a per-step retry counter, which fixes a class of ralph/iteration bugs where body children were born "exhausted" and pipeline steps collided on a single artifact directory — making the retry/iteration machinery behave correctly across multi-step formulas.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **`gc.attempt` did double-duty, so ralph body children were born "exhausted."** Review validated the split: a body child first reached in iteration N is no longer stamped `attempt=N`, so it no longer trips `processRetryEval`'s `attempt > max_attempts` guard on its first run (`internal/formula/ralph.go`, `internal/dispatch/control.go`).
- **The required-artifact path collapsed every pipeline step onto one `{attempt}` directory.** `{iteration}` now substitutes only when `gc.iteration` is set; otherwise the unresolved-template check fails loudly instead of silently resolving a stale directory (`internal/dispatch/retry.go`).
- **`latestAttemptFromCandidates` leaned on a `max(gc.attempt)` tie-break that breaks once attempt resets to 1.** Precedence is now explicit — precise identity (`control.ID` / namespaced `gc.step_ref`) outranks a bare `gc.step_id`, and legacy pre-namespaced molecules still resolve (`internal/dispatch/control.go`).
- **Outer ralph retry clones never advanced `gc.iteration` and blanket-stamped `gc.attempt`.** Every clone now advances `gc.iteration` to the new outer index while `gc.attempt` follows local step semantics, pinned on both clone paths by `ralph_retry_iteration_test.go` (`internal/dispatch/ralph.go`).

Both reviewers ran `go test ./internal/dispatch ./internal/formula` and re-ran the targeted iteration/retry subset from the worktree — both packages pass.

- `mc-t0fd` — [Behavioral Correctness · minor] On an outer ralph retry, a synthesized scope-check's `gc.attempt` is clone-stamped with the new outer-iteration number while the mint path copies the wrapped control's value, so the two paths disagree. A full value-read audit confirmed this is **provenance-only and inert today** (no consumer reads the divergent value), but it is a latent trap if a future change starts reading a scope-check's `gc.attempt` as a value. Not addressed in this PR.
- `mc-s044` — [Test Evidence Quality · minor] The clone-path regression test seeds subject, retry-control, attempt-run, plain-member, and check beads but never a scope-check — exactly the bead shape whose clone/mint divergence is described above — so that contract is unpinned. Not addressed in this PR.
- `mc-3dvp` — [Debuggability · nit] A stale `controlIdentitySet` (singular) doc block remains stacked above the renamed `controlIdentitySets` block; it names a symbol that no longer exists and describes the pre-split identity model this PR demotes. Not addressed in this PR.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `47b205d6` has passing required GitHub checks, and the merge-ready metadata is in place. The branch was rebased onto the latest `main` (contributor commits and authorship preserved, patch unchanged) so the required checks run against the current base. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/33058943418

### Review Iterations
1 review pass performed — the PR was approved on the first pass with no fix iterations required.

---

PR #5636, opened 2026-08-26
fix(testutil): renew lapsed runtime.Provider waivers to unblock repo-wide CI

julianknutsen:
## Live CI outage — this unblocks the repo

`runtimeWaiverExpiry` was hardcoded to `2026-08-26`, so at midnight UTC today every `runtime.Provider` waiver owned by `ga-80po0c.3` expired and `TestCatalogMatchesProductionWiringAndDocumentation` started failing with 8 errors.

That check is repo-wide. **Confirmed in live CI**, not inferred: every run started after `2026-08-26T00:00Z` fails in `Integration / packages-core-4-of-4` on exactly these 8 errors (e.g. run `32916248989` at `00:45:48Z`). It also fails `pre-push`, so it blocks local pushes as well — it was the *sole* failure blocking an unrelated SDK branch.

Every open PR whose checks are currently green last ran on 08-24/08-25, i.e. **before** the expiry. Anything re-run today goes red, so the factory cannot pass a CI gate until this lands.

## This is the third renewal, and that's the real finding

The previous renewal asked that the next one say plainly whether renewing is debt. It is:

- `ga-80po0c.3`'s only open child, `ga-80po0c.3.3`, is still `in_progress` and was **last touched 2026-07-18** — 39 days
- against a **two-week** horizon chosen specifically to put the question in front of the owner while context was fresh
- that didn't happen, and a third short horizon has no reason to do better

Lapsing isn't reaching the owner; it's just telling everyone else that main is red. I've escalated on `ga-80po0c.3` rather than renewing quietly.

## Re-checked, not rubber-stamped

As the last renewal did, I re-verified all eight against `cmd/gc/runtime_registry.go`: every constructor is still a live registration and none has gained a runnable contract, so **none is retired as stale**.

One trap worth recording, now in the comment: `exec` and `ssh` register through `RegisterPrefix("exec:"/"ssh:", ...)` rather than by bare name, so a name grep alone reads them as dead. The ledger's `exec → t3bridge.NewSeamBacked` mapping is likewise correct — it matches the branch at `runtime_registry.go:97`.

## Change

Two weeks again (`2026-09-09`), well inside `maxWaiverHorizon`. `TESTING.md` is regenerated from the ledger; that diff is **exactly the 8 expiry dates and nothing else** (verified: 8 lines `through 2026-08-26` → 8 lines `through 2026-09-09`).

The real fix is landing the contracts or retiring the claim that they're coming. **Please don't renew a fourth time without deciding.**

`go test ./internal/testutil/providerledger/...` → `ok`.


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277238013

julianknutsen:
Superseded by #5610, merged as `bd84f0172a`.

Both PRs were opened against the same outage: at 2026-08-26 00:00 UTC every `runtime.Provider` waiver in `internal/testutil/providerledger` lapsed at once and `TestCatalogMatchesProductionWiringAndDocumentation` went red with 8 errors, which reds `make test-fast-parallel` and therefore the `.githooks/pre-push` gate for the whole fleet.

This PR renews the shared `runtimeWaiverExpiry` constant. #5610 removes that constant instead and gives each waiver its own expiry, which is what stops all eight lapsing on the same instant a fourth time, and it moves the owner off the dead `ga-80po0c.3` onto a live one. A two-lens fable council reviewed it and returned approve-with-nits with zero merge blockers.

Verified on `origin/main` after the merge: `go test ./internal/testutil/providerledger/ -count=1` is `ok`.

Follow-ups filed rather than blocking the merge: `ga-b51te` (P1 — the expiry clock should not be able to red the fleet push gate; the same lockstep exists in `internal/testpolicy/resourcecensus`, whose 37 waivers all expire 2026-10-01), `ga-x7hxo` (P2 — the eight new expiries are still a byte-identical literal with nothing enforcing divergence), `ga-lgizl` (P2 — the owner assertions are self-referential).

Closing as redundant. Nothing here is lost; the outage it targeted is cleared.

Issue #5508, opened 2026-08-22
Support event-driven pool wakeups and empty-hook cooldown


AlexBelous:
Pool agents can currently be started by periodic polling even when their work query is empty, paying the full cold-start cost for no useful work.

Please add an event-driven wake path from successful routing/sling to the target pool, plus a configurable cooldown after an empty hook (the intended clerk policy is at least two hours). The daemon heartbeat may remain a safety net, but should not be the primary wake mechanism.

Acceptance:
- routing a ready bead wakes the matching pool without waiting for a poll interval;
- an empty hook suppresses another cold start for the configured cooldown;
- queued work arriving during cooldown still wakes through the event path;
- metrics expose starts by role and whether a session claimed useful work;
- tests cover event delivery, empty cooldown, and no missed routed work.

AlexBelous:
Refinement: the primary fix should be a zero-cost supervisor gate, not merely a longer timer. The supervisor/dispatcher should execute the configured shell work_query before starting any model session, and only spawn the pool worker when the result is non-empty. A cooldown remains a secondary safety net.

Please apply the same generic pattern to other pool roles that expose work_query. Tests should prove that empty work_query results create zero model sessions and that newly routed work is not missed.

AlexBelous:
Startup-order refinement from production: generated pool sessions currently run the managed SessionStart gc prime --hook before the worker can evaluate its hook. The pool prompt must also make gc hook --claim --json the first model action, exit immediately on empty/no_work, and run full gc prime only after a bead exists.

The zero-cost supervisor preflight remains the necessary primary fix, because prompt order alone cannot avoid context already injected by the managed SessionStart hook. Please cover both layers in tests and apply the generic ordering to all work_query pools.

PR #5639, opened 2026-08-26
Wake idle pool members for ready shared-template work

quad341:
## What this changes

Idle members of a configured agent pool now wake when ready work is assigned to the pool's shared template. Previously, that template-level assignment could be discoverable and claimable by the pool while the awake-set still treated it as belonging to no concrete session, leaving every eligible member asleep.

The ownership boundary remains strict after a claim: in-progress work only keeps its concrete holder awake. Blocked, deferred, terminal, and otherwise-not-ready work still creates no wake demand.

## Review notes

- Pool eligibility comes from typed `pool_managed` session metadata, not session-name or numeric-suffix inference.
- The template fallback applies only to ready, open work; it cannot make all pool members owners of in-progress work.
- Configured named sessions, manual sessions, lookalike names, and members of other pools remain excluded.
- This changes no endpoint, wire shape, dependency, or migration.

## Test plan

- [x] Run the full 40-job local CI-equivalent suite and verify all five new awake-set tests execute in both required `cmd/gc` lanes.
- [x] Exercise ready, blocked, deferred, terminal, in-progress, named-session, manual-session, lookalike, and cross-pool cases.
- [x] Run policy, vet, affected-scope lint/format, and a clean merge simulation against current `origin/main`.
- [x] Release gate: [`release-gates/ga-z1axa6-idle-pool-template-wake-gate.md`](release-gates/ga-z1axa6-idle-pool-template-wake-gate.md)

---
- Related to #5508 — touches the same pool wake-up surface: the awake-set now treats ready, open work assigned to a configured pool's shared template as real wake demand, so eligible idle members no longer stay asleep on pool-routed work; it does not add the event-driven wake path from routing/sling, the empty-hook cooldown, or the per-role start metrics that the filed issue asks for


quad341:
@quad341 — the part of this I appreciate most is that you drew the serviceability/ownership line explicitly instead of just loosening the matcher and hoping. Ready open work can be *serviced* by any eligible pool member; a claim still binds to a concrete holder. And the comment on `sessionAssigneePoolTemplateMatches` explaining why drained and dependency-only slots are excluded — naming the sibling paths that make the same exclusion — is the kind of note that saves the next person an hour. The exclusion matrix in the tests (numeric-suffix lookalike, other pool, manual, configured-named) is the right paranoia for membership logic.

The red check isn't yours: the Mac lint runner is on Go 1.26.5 while `go.mod` wants 1.26.6, so lint never loaded packages. That's mine.

Two things I'll take on my side rather than send back: a multi-member test (every case here has exactly one bead, so the plural in your title isn't pinned — as written one ready bead wakes all of them, which I'll pin as intended unless you meant to cap it), and refreshing the release gate, which still certifies `f60f2f9` — before the drained/dep-only fix.

One FYI, no action needed: `AssigneeIdentities` never emits a bare template, so `Ready(assignee=<template>)` is never queried. This fires for molecule/workflow roots, not plain template-assigned tasks. Narrower than the body reads — worth knowing before someone files it as still-broken.


---

PR #5640, opened 2026-08-26
fix(cmd/gc): converge every CLI by-id surface on the residency resolver's plan

julianknutsen:
**Segment 1 of 8 — split out of #5597 at the reviewer's request.** Base: `main`.

#5597 was 186 files and 18,100 changed lines, past the ceiling the review tooling
serves in one pass. It has been cut into eight stacked PRs, each a branch pointed
at a commit that was already in the series — no rebase, no commit rewriting, no
conflict resolution, so every commit here is byte-identical to the one that was
reviewed and CI-tested as part of #5597. Each segment builds, vets and passes the
residency boundary guard on its own.

| # | PR | lines | story |
|---|----|-------|-------|
| 1 | **this one** | 1,998 | CLI by-id reads converge on the resolver |
| 2 | #5641 | 3,408 | blind writers get class placement; the boot relic census |
| 3 | #5642 | 1,838 | the foreign-id provider contracts |
| 4 | #5643 | 2,221 | make the residency rows falsify; close guard evasions |
| 5 | #5644 | 2,479 | fence every provider; count closed relics |
| 6 | #5645 | 2,315 | retire the last hand-rolled probes |
| 7 | #5646 | 3,143 | pure code motion of the guard |
| 8 | #5597 | 4,180 | cutover: payload carry, binding verdicts, preflight |

**Read this one first.** It is where the bug class is stated, and every later
segment is an application of it.

---

## The bug class

A city that has split its beads across a work ledger and a relocated infra-class
**binding** has two places a given id can live. `gc storage migrate` preserves ids
and deletes nothing, so after a cutover the work ledger still holds a **frozen
copy** of every relocated bead — correct at cutover, never touched since.

Every read that walks the city's *directories* to find a bead therefore has a
second, wrong answer available to it. And the failure is silent, because *"the
binding could not answer"* and *"the bead is not there"* were the same outcome:
a miss fell through to the directory scan, the scan found the frozen twin, and
the command printed it with exit 0.

`internal/storeref` already owned the correct walk — `Plan(ByID)` and the binding
legs — from the earlier slices on `main`. What this segment does is stop the CLI
by-id surfaces from re-deriving it.

## What changed

**`classRoutedStoreForID` routes through the residency resolver** rather than its
own hand-rolled binding lookup, and the `gc bd` class front door resolves from the
resolver's grouped bindings instead of `graphClassBinding`'s parallel grouping.
Two code paths that answered "who owns this id" became one.

**`gc beads` and `gc convoy` ask the class binding before scanning the city's
directories.** That ordering is the fix: the binding is the authority for its
namespaces, so a directory scan is a fallback, not a peer.

**Binding-wins became one shared property both by-id surfaces run**
(`internal/storeref/storereftest/binding_wins.go`), so the CLI and the API cannot
drift on it — the same table drives both, and a surface that stops honouring the
binding fails in the shared row rather than in whichever plane happened to have a
test.

**Class-door closes are gated through ADR-0009**, and `maybeRouteBdByID` is split
so the routing decision and the close gate are separable.

**The guard grew a second half.** `scripts/check-residency-boundary.sh` was a grep
ratchet; its AST half now walks the same tree the grep half walks, and
`cliSoleClassBinding` call sites are ratcheted so a new blind reader has to be a
reviewed act rather than an unnoticed one. `093a55ed79` makes adding a second
by-id work axis explicitly reviewed for the same reason.

The last commit pins two rules an earlier mutation sweep found unguarded — both
were correct in production and simply had nothing asserting them.

## Verification

- `go test ./internal/testutil/providerledger/` green (this head carries main's
  runtime-provider waiver renewal; every segment does)
- Full-suite results are inherited: these commits were tested as part of #5597,
  whose head was green across all 86 checks

Refs `ga-qdt5y`.

### DIFF
--- a/cmd/gc/by_id_binding_owner_test.go
+++ b/cmd/gc/by_id_binding_owner_test.go
@@ -4,13 +4,16 @@ import (
 	"bytes"
 	"errors"
 	"path/filepath"
+	"strings"
 	"sync"
 	"testing"
 
 	"github.com/gastownhall/gascity/internal/beads"
 	"github.com/gastownhall/gascity/internal/beads/splittest"
 	"github.com/gastownhall/gascity/internal/config"
+	"github.com/gastownhall/gascity/internal/coordclass"
 	"github.com/gastownhall/gascity/internal/fsys"
+	"github.com/gastownhall/gascity/internal/storeref"
 	"github.com/gastownhall/gascity/internal/storeref/storereftest"
 )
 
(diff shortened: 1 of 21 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277238814

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277238883

quad341 (dismissed):

Cutting #5597 into eight branches pointed at commits already in the series was the right call and it shows: every commit here is byte-identical to what CI already saw, and this segment builds, vets and passes the residency guard standing alone. Our CI ran the full cmd/gc shard matrix on it, green.

The piece worth singling out is `storereftest.RunBindingWins`. Writing the dual-residency property once and having the HTTP lane, the `gc bd` door and the convoy resolver each supply an adapter is the correct shape for this bug class — two planes can each pass an excellent local pin and still disagree about one id, and only a shared clause catches that. Checking the fixture's own premise before the clauses run is what makes it trustworthy; a vacuously-green cross-plane property gets cited as proof of agreement. `legacyClassRouteCandidates` carrying the deleted list verbatim is the same instinct.

One for a later segment, not this one: `cliByIDPlan` builds from `cliResidencyTopology` rather than `residencyTopologyForCity`. Since `resolveOwningStoreDir` is now reachable in-controller through `executionEmitStore` in order dispatch, a converged split city might plan binding legs through the one-shot funnel inside a process that already registered its routes. I traced it as far as reachability but couldn't reproduce the impact without a real split city — is this seam meant to stay on the funnel the way `ready_federation.go` does?

Thanks — on to #5641.


---

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277243440

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277243509

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277243582

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR converges every CLI and HTTP by-id surface (`gc convoy`, `gc beads show`, the `gc bd` class door, and the HTTP by-id resolver) onto the `internal/storeref` residency resolver's plan, so the migrated store binding — not the frozen retained work copy — answers by-id lookups after `gc storage migrate`. That closes a stale-read data-correctness bug class (the #5125 / #5127 / ga-axin6 family) as a single class rather than patching each surface one at a time, which matters because a by-id surface silently answering from the pre-migration copy is exactly the kind of correctness fault that is hard to notice and hard to trust.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- By-id surfaces answering from the frozen retained work copy after `gc storage migrate`: the patch routes all four by-id surfaces through the `internal/storeref` residency resolver so the binding wins, closed as a class and pinned by the executable cross-plane property `RunBindingWins` (`internal/storeref/storereftest/binding_wins.go`) that both by-id planes run.
- Fail-loud on a binding fault instead of a silent fallback to the stale copy: validated — a refused city still reports `relocated=true` rather than hiding the fault (`TestSoleClassBindingCarriesARefusedCityRatherThanHidingIt`), and a per-class fan-out is refused at both sanctioned callers.
- Single-store cities staying byte-identical: validated — the single-leg plan returns the work leg unprobed (`Owner.Read=false`) with no extra `Get`, measured by `countingGetStore`, so the unification adds no cost on the common path.

- ga-dqd7i (maintainability / observability, minor): `warnAutocloseResolutionFault` (`cmd/gc/cmd_convoy.go`) de-duplicates binding-fault diagnostics with a once-per-**process** `sync.Once`. That is right for the short-lived `bd on_close` autoclose burst, but the same path also runs on the long-running controller (via `executionEmitStore`), where the first transient fault consumes the `Once` and later unrelated persistent binding faults then log nowhere. Behavior is unaffected — the loud primary surfaces (HTTP 500-not-404, the bd class-door error) still fire; only the secondary diagnostic visibility is lost. Filed as ga-dqd7i for a follow-up that scopes the de-dup to the autoclose burst and corrects the remedy-clause wording for the `executionEmitStore` caller. Two low-severity density measurements were also noted and are captured on that bead (a moderately dense test helper at `binding_wins.go:137`; cognitive complexity 17 at `cmd_beads.go:264`) — both under the quality caps.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `94e02e02` has passing required GitHub checks (CI, CodeQL, Go Vulnerability Scan, PR evidence watchdog, and Review Formulas all green; the non-required Mac Regression check is advisory and does not gate merge), and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/33041598476

### Review Iterations
1 review pass performed; the PR reached approval on the first pass with no maintainer fix iterations. Two independent reviewers converged (design-implementation + implementation), all six behavior validations were confirmed, and the only new finding was the minor observability follow-up filed above.

---

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR converges every CLI by-id
surface on the residency resolver's plan, so by-id reads and writes route to the
class binding instead of the retained pre-migration work-store copy — which
matters because it closes a root-loss hazard in the residency migration: a
failed binding probe can no longer masquerade as "bead absent," and in-process
callers stop opening duplicate store handles.

### Original PR Review
Decision: approve.

Specific gaps validated by review:
- By-id residency now routes to the class binding, not the retained city-store
  copy: a split city plans `Plan(ByID)` over `{bindings, work}`, and a failed
  binding probe is surfaced as an error rather than folded into `ErrNotFound`,
  so "the binding couldn't answer" cannot masquerade as "the bead is absent"
  (the exact root-loss shape this lane exists to prevent). Single-store cities
  stay byte-identical (`cmd/gc/by_id_residency.go`).
- The work-residual sentinel refuses through a whole store instead of silently
  reporting "absent" — `refusedClassStore` is constructor-built so the zero
  value can never read as an empty list; every op but `Get` reports the contract
  violation (loud bug, not silent miss).
- In-process callers avoid a second handle on the binding root: `cliByIDPlan`
  prefers a controller's boot-registered routes and falls to the one-shot funnel
  only for genuine one-shots, so an in-process by-id resolution doesn't open a
  duplicate managed-Dolt server or a second sqlite writer. Registration is taken
  under the controller lock and dropped unlink-only-if-ours.
- The nil-map shutdown race in the binding memo is closed, and the per-class
  fan-out refusal is enforced as a runtime check (`cliSoleClassBinding` refuses
  `len>1`) rather than left as a comment.

Coverage note: Gemini was skipped by config, so cross-file review coverage was
Claude (design-implementation) + Codex (implementation). Codex reported zero
findings; neither reviewer surfaced a cross-file hazard on this 21-file seam.

- ga-jj33s: Contract & Interface Fidelity, minor — the class-door work-record
  close gate uses `cityPath` as the git-reachability fallback repo
  (`cmd/gc/cmd_bd_by_id.go:884`, `:990-996` → `cmd/gc/work_record_gate.go:270-279`)
  instead of the bead's own work scope, so a class-resident shipped step closed
  without `gc.work_dir` can get an opposite reachability verdict from the
  work-store door (`cmd/gc/cmd_bd.go:384` passes `target.ScopeRoot`). Warn-only
  under the default config (`GC_WORK_RECORD_ENFORCE` off) with a narrow affected
  population, so it does not gate this merge. Filed for a future pass.

### Maintainer Changes
No maintainer code changes — PR accepted as-is. The branch was mechanically
rebased onto the latest `main` (`fc61b5b`) to refresh the merge base; all 11
commits remain contributor-authored with their diffs unchanged (patch-id
preserved).

### Final Review Status
Ready for the merge queue: final head `d9f9f76` has passing required GitHub
`status/merge-ready` so the merge queue can pick it up. (The only red checks are
the non-required advisory `Mac / quality` lane and its summary, a known
runner-image toolchain-floor issue unrelated to this change; every required
check is green.)

CI: https://github.com/gastownhall/gascity/actions/runs/33076393673

### Review Iterations
1 review pass performed — approved on the first attempt with zero fix
iterations.

---

PR #5641, opened 2026-08-26
fix(residency): give the blind write paths class placement and census the binding's relics

julianknutsen:
**Segment 2 of 8 — split out of #5597 at the reviewer's request.** Base: segment 1.
See segment 1 for the bug class and the full segment table; it is the one to read
first. Every commit here is byte-identical to the one reviewed in #5597.

Segment 1 fixed the by-id **readers**. This one fixes the **writers**, and then
makes the resolver's own retirement bit an observed fact rather than a promise.

## The blind writers

Four surfaces placed beads without asking the class seam where they belonged.
A bead minted into the wrong store on a split city is worse than a bad read: the
read is wrong once, the write is wrong forever.

- **`gc handoff`** wrote handoff mail into whichever store the command happened to
  hold, not the messaging-class store (S5 C1).
- **The PR repair workflow** cooked its molecule without class placement, so a
  repair run on a split city scattered its step beads (S5 C2).
- **The API bead-create door** had no placement at all — `POST /beads` minted into
  the work ledger regardless of the id's namespace (S5 C3).
- **The nudge queue's id namespace was not registered as class-reserved**, so
  nothing downstream could even tell that those ids belonged to a class (S5 C4a).

## C4a shipped neither of them honestly, and C4b is why this segment does not end there

Registering the namespace made the *table* right and left two things wrong: a
binding would still accept a **pinned foreign id** — an id outside its own
reserved namespaces, written in with an explicit id — and the resolver's
`MintsReserved` bit was inferred rather than observed. `03a0efe91f` fences the
first and observes the second.

That pair is why the cut lands here rather than after C4a: C4a's own follow-up
commit says *"C4a shipped neither of them honestly."* Splitting between them would
hand a reviewer a half-finished mechanism.

## The relic census retires the residence probe

`storeref` carried a per-read **residence probe**: before trusting a binding, ask
it whether it holds ids outside its namespaces. Correct, and paid on every read.

`939ebafcdb` replaces it with a **boot-time census**: `gc storage boot` walks the
binding once, records whether it holds preserved ids, and the resolver reads that
durable fact instead of re-deriving it. `HasLegacyResidents` stops being a guess.

`169fb83090` then retires `graphClassBinding` — the last parallel grouping of
bindings — in favour of the resolver's own, and `29a28fe9fc` closes the review
council's findings on the slice.

## Disclosure: this census is open-only, and segment 5 falsifies that

The census as landed here counts **open** relics. That is wrong — a binding whose
relics have all been closed still holds preserved ids, and the retirement verdict
must say so. `8f9eeee96b`, in **segment 5**, widens it to count closed relics too.

A defect introduced in one segment and fixed three segments later is inherent to
splitting a linear series without rewriting it, and it is stated here rather than
left to be discovered. The alternative — reordering commits so the fix precedes
the defect — would mean rewriting 47 commits across hub files touched by nine of
them, discarding the CI and review history that makes these segments cheap to
trust.

## Disclosure: the pinned-id fence spans four segments

`03a0efe91f` fences the sqlite engine. Segment 3 states the fence as a **provider
contract**; segment 5 applies it to the second provider and the test kit; segment
8 pins it at every workspace open. Each stage is complete in itself — the sqlite
engine is genuinely fenced when this segment lands — but a reviewer looking for
"is the fence everywhere yet" will not find the answer until segment 5.

## Verification

Refs `ga-qdt5y`.

### DIFF
--- a/cmd/gc/api_state.go
+++ b/cmd/gc/api_state.go
@@ -1706,6 +1706,18 @@ func (cs *controllerState) OrdersBeadStore() beads.OrdersStore {
 	return beads.OrdersStore{Store: resolveOrderStore(cs.storageRoutes, cs.cityBeadStore, cs.cfg, cs.cityPath, cs.eventProv)}
 }
 
+// ClassBindingHasLegacyResidents replays the boot census for one binding store.
+//
+// The store values the class accessors above hand out are the routes' own
+// stores — resolveClassStore returns routes.storeFor unwrapped — so the census
+// map the boot keyed by store answers directly here. A store this city never
+// censused, the work store included, keeps its probe.
+func (cs *controllerState) ClassBindingHasLegacyResidents(store beads.Store) bool {
+	cs.mu.RLock()
+	defer cs.mu.RUnlock()
+	return cs.storageRoutes.hasLegacyResidents(store)
+}
+
 // CityBeadsDiagnostic returns the city-level bead store selection diagnostic.
 func (cs *controllerState) CityBeadsDiagnostic() *beads.BeadsDiagnostic {
 	cs.mu.RLock()
--- a/cmd/gc/class_store_agreement_test.go
+++ b/cmd/gc/class_store_agreement_test.go
@@ -1,0 +1,153 @@
+package main
+
+// The legacy front door and the placement plan must name the same store.
+//
+// resolveClassStore is what every class-scoped write in cmd/gc still goes
+// through, and storeref.Plan(Class{…}) is what the resolver answers the same
+// question with. They are separate code reading the same routes, and a class
+// they disagree about is a class whose writes land in one store while every
+// resolver-built reader looks in the other — a bead that exists and cannot be
+// found, with no error anywhere to say so.
+//
+// This is an agreement pin, not a behavior test: neither side is asserted
+// against a literal store here, only against each other, so the day one of them
+// gains a class the other does not, this fails rather than the fleet.
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/gastownhall/gascity/internal/beads"
+	"github.com/gastownhall/gascity/internal/config"
+	"github.com/gastownhall/gascity/internal/coordclass"
+	"github.com/gastownhall/gascity/internal/storeref"
+)
+
+// configClasses is every class name resolveClassStore can be asked about.
+func configClasses() []string {
+	return []string{
+		config.BeadClassWork,
+		config.BeadClassGraph,
+		config.BeadClassMessaging,
+		config.BeadClassSessions,
+		config.BeadClassOrders,
+		config.BeadClassNudges,
+	}
+}
+
+func TestResolveClassStoreAgreesWithThePlacementPlan(t *testing.T) {
+	binding := beads.NewMemStore()
+	graphOnly := beads.NewMemStore()
+	sessionsOnly := beads.NewMemStore()
+
+	wholeSplit := map[coordclass.Class]beads.Store{}
+	for _, class := range coordclass.Classes() {
+		if class.IsInfrastructure() {
+			wholeSplit[class] = binding
+		}
+	}
+
+	tests := []struct {
+		name   string
+		routes *storageRoutes
+	}{
+		{"single-store city", nil},
+		{"whole split", &storageRoutes{stores: wholeSplit, binding: "infra"}},
+		{
+			// Not a shape this build serves (storageSupportedTopologyStatement),
+			// but both sides accept it as data and a disagreement here is the
+			// same disagreement.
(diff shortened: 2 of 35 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277238672

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277238741

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277243363

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277256070

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277258191

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR gives the "blind" write paths — the `POST /v0/beads` create door and pinned-foreign-id writes — explicit class placement, and censuses a class binding's relics so the live residence probe can be retired. That matters because it stops infra-class bodies from being minted into the work ledger on a converged city, fences pinned foreign ids out of a binding they don't belong to, and replaces a per-request residence probe with a one-time boot census — tightening residency correctness while keeping the shipped default behavior byte-identical.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **Create-door routing (`internal/api`):** `POST /v0/beads` previously resolved its store from the rig (`findStore(rig)`), which could mint infra-class bodies into the work ledger. The door now resolves placement from the bead's *class* (`coordclass.Classify → storeref.Plan → ResolvePlacement`); the work-class and no-binding fallbacks are byte-identical to the prior path, and classification runs safely before the assignee is known.
- **Reserved auxiliary namespace (`internal/config`):** the `gcnq` nudge-queue namespace was owned by nobody. It is now registered as class-reserved — load-bearing for the fence, the by-id door, and the census alike.
- **Fenced write path (`internal/beads`, `internal/storebinding/sqlite`):** `Create`/`CreateWithForeignID` were split into a shared `create(b, allowForeign)`; only `Create` runs `checkPinnedIDNamespace`, the migration copy stays exempt, and fencing is correctly disabled for a binding that serves a class with no reserved namespace (work).
- **Relic census (`internal/storeref`, `cmd/gc`, `internal/api`):** the live residence probe is retired in favor of a boot-time census of a binding's relics. Every failure direction is pessimistic (census error / nil receiver / uncensused / seeded all report "has residents = true"), the API and CLI planes share one verdict via the unwrapped store key, and the shipped default (no `[storage]`) is byte-identical by construction.

- **ga-mb8xb** — [Resource Lifecycle & Cleanup, minor] The boot census is an unbounded full scan (`OpenLegacyResidents` → `store.List` with `AllowScan: true`, no `Limit`) re-run on every one-shot CLI process on a converged city. On the long-lived controller this is one negligible boot-time read, but each short-lived route-resolving command pays a full infra-binding scan. Cheapest lever noted in the finding: pass `SkipLabels: true` in the census `ListQuery`.
- **ga-r271r** — [Contract & Interface Fidelity, minor] Reserved-auxiliary-namespace completeness is now a standing invariant that the fence, census, and by-id door all rely on. Complete today (only `gcnq`), but a future subsystem pinning ids under a new, unregistered namespace would fail quietly and split-brained (the fence refuses its writes; the census counts its rows as relics and pins the residence probe forever). Suggested hardening: a guard test asserting every store-derived pinned prefix appears in the reserved-auxiliary union.

A related out-of-scope gap — the `bd` SQL relocation guard not recognizing auxiliary class namespaces (`gcnq`) — is tracked separately as ga-qdt5y.1 and was correctly left out of this PR's scope.

### Maintainer Changes
No maintainer changes -- PR accepted as-is. The branch was rebased onto current `main` (force-pushed) to keep it mergeable; every commit remains contributor-authored and is patch-identical to your originals (`git range-diff` shows all commits unchanged), and no maintainer code commits were added.

### Final Review Status
Ready for the merge queue: final head `ae2a998e` has passing required GitHub checks (the remaining red checks are advisory runners that also fail on `main` and touch none of this PR's packages), and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/33229774869

### Review Iterations
1 review pass performed (decision: approve). Deep review covered the full four-commit scope with per-commit fix validations and local `go vet` plus focused unit tests across `storeref`, `config`, `storebinding/sqlite`, `beads` (fence), `api`, and `cmd/gc`; the census/residency slice was independently confirmed by a second reviewer.

---

PR #5642, opened 2026-08-26
fix(beads): pin the weak-ParentID and pinned-id-fence provider contracts

julianknutsen:
**Segment 3 of 8 — split out of #5597 at the reviewer's request.** Base: segment 2.

Three commits, one subject: **what a beads provider must promise about ids it did
not mint.** This is the segment with real behavior change in it, which is why it
is not merged into the hardening segment that follows.

## Shared binding construction

`660280cdb9` gives `cmd/gc`, `internal/api` and `storeref` one `BuildBindings`
constructor. Three planes were each assembling `ClassBinding` values from config,
and the three assemblies had already drifted on which fields they populated — a
binding built by the API plane could disagree with the same binding built by the
CLI about its own namespaces.

## The weak-ParentID contract

A bead's `ParentID` can point into a store the writer cannot see. On a split city
that is normal: a step bead in the binding has a parent in the work ledger.
Strict parent validation refuses those, so the providers relax it — but *how far*
they relax it was unstated, and `NativeDoltStore` and the sqlite store had
relaxed it differently.

The boundary is drawn at the **namespace**: a parent id in a namespace this store
does not own is accepted without resolution; a parent id in a namespace it *does*
own must resolve. `1f272777dc` draws that line and pins it on `Update` as well as
`Create` — `Update` was the arm that had no assertion at all.

**Behavioral delta worth calling out.** Cross-store parent creates that were
previously refused now succeed. That is the point of the change, and it is a
widening of a write surface: `gc bd create --parent <id-in-another-store>` on a
split city used to error and now does not. It is correct — the refusal was
punishing the normal shape of a migrated city — but it should be read as a
widening rather than as pure hardening.

## The pinned-id fence as a provider contract

`d910a7b2fd` states the fence segment 2 built into the sqlite engine as a
**contract every provider must satisfy**
(`internal/beads/beadstest/pinned_id_fence_conformance.go`), so a provider that
does not fence pinned foreign ids fails the shared conformance suite rather than
passing quietly until someone notices on a live city.

The contract lands here; the **second** provider is fenced in segment 5 and the
workspace-open path in segment 8. See segment 2's disclosure on the fence thread.

## Verification

Refs `ga-qdt5y`.


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277238603

julianknutsen:
## Rebased onto main at `5c4e7faefdad5017d26df3d8674e35219a35403a`

The 84-file / mergeable=dirty state was an artefact: this branch still carried the
ORIGINAL s1 and s2 commits, which landed on main as squashes (#5640 `1f18a71017`,
#5641 `0d39380ada`). 18 of the 19 merge-tree conflicts were that stale lineage
colliding with its own squash, not real content.

Rebuilt as a clean rebase: new branch from `origin/main` (`faf4e3b89d`), cherry-picked
only this segment's three unique commits, dropped the carry merge `377c2e542b`.
The PR now shows **3 commits / 31 files**, mergeable=true, base `main`.

### Commits carried

| new sha | original | subject |
| --- | --- | --- |
| `917f853131` | `660280cdb9` | fix(residency): share the binding constructor and pin the weak ParentID contract |
| `5578a9346d` | `1f272777dc` | fix(beads): draw the weak-ParentID boundary at the namespace, and pin it on Update |
| `5c4e7faefd` | `d910a7b2fd` | feat(beads): state the pinned-id fence as a provider contract (ga-qdt5y.11) |

Author and `(cherry picked from commit ...)` trailers preserved on all three.

**Faithfulness check:** the rebased delta `origin/main..HEAD` is **31 files, +1581/-257**
— the same shape as the original segment delta `29a28fe9fc..d910a7b2fd`. Comparing the
two diffs line-for-line over their added/removed lines gives an empty symmetric
difference: nothing lost, nothing new.

### Conflicts and resolutions

**1. `cmd/gc/cmd_github.go` — doc comment on `defaultAttachGitHubPRRepairWorkflow` (real).**
#5488 (`3b44512c8f`) replaced the `cityPath`/scope-store signature with
`defaultAttachGitHubPRRepairWorkflow(store, graphStore, ...)` + `cookOnClassRouted`, and
replaced the paragraph this segment was appending to with "Two stores because the classes
differ: ...". Kept main's signature, body and prose verbatim, and appended s3's own hunk —
the 10-line "On a converged city that therefore SPLITS the molecule from its parent"
paragraph — unchanged. It reads correctly against #5488's sentence: the classes differing
is exactly why the molecule splits from its parent, and the paragraph then explains why
the weak `ParentID` makes that a working contract rather than a leak. Nothing of #5488's
was dropped and nothing of s3's was reworded.

**2. `internal/beads/beads.go` — `Bead.ParentID` doc (self-supersede, not a main conflict).**
Both sides were this segment's own: commit 1 adds the "Weak, therefore, is the contract"
paragraph, commit 2 rewrites it to draw the boundary at the namespace. Took commit 2's
text, which is the segment's final state.

**3. `scripts/residency-boundary-patterns.txt` and `-baseline.txt` — auto-merged, verified by hand.**
The baseline is shrink-only, so it was not restored wholesale. Against main's *current*
baseline the segment contributes exactly its own three edits: `bd_relocated_classes.go
relocatedBeadClasses` renamed `d:ReservedClassPrefix` -> `d:ReservedClassPrefixesFor`, plus
`cmd_bd_by_id.go bdIDIsClassReserved d:AllReservedClassPrefixes` and
`ready_federation.go readyFederationLegsOverBinding d:ReservedPrefixesFor`. The patterns
file gains its six `d:` rows (`ReservedClassPrefixes`, `ReservedClassPrefixesFor`,
`AllReservedClassPrefixes`, `IsReservedClassPrefix`, `ReservedPrefixesFor`,
`NudgeQueueIDPrefix`) byte-identically. `check-residency-boundary.sh` is green on the
result with no regeneration.

No other file needed a hand resolution.

### Gates (all on the pushed tree)

| gate | result |
| --- | --- |
| `go build ./...` | pass |
| `go vet ./...` | pass |
| `gofmt -l` | clean |
| `golangci-lint run ./...` | 0 issues |
| `go test ./internal/storeref/... ./internal/beads/... ./internal/storebinding/... ./internal/config/... -count=1` | pass (11 pkgs) |
| `go test ./cmd/gc -run 'Residency\|Binding\|Fence\|ParentID\|Namespace\|ExtMsg\|NudgeQueue\|CrossStoreParent'` | pass (233 tests) |
| `go test ./internal/api/... -count=1` | pass (5 pkgs) |
| `make check-docs` | pass |
| `make test-fast-parallel` | all 10 fast jobs passed |
| codegen drift (`genspec` / `genclient` / `genschema`) | no diff in any generated artifact |

Spot-checked that the load-bearing new tests actually execute rather than
silently skipping: `TestSQLiteStorePinnedIDFenceConformance` runs all 13 rows
including the six `OverRestrictionControls` controls,
`TestNativeDoltStoreConformance/ParentIDNamesARowThisStoreDoesNotHave` runs,
and `TestOpenEngineFencesAnInfrastructureBindingToEveryNamespaceItHolds` runs.

`engdocs/architecture/beads.md` invariant **16** is still free on main (main ends at 15),
so no renumbering was needed.

### One correction to the PR body

The body overstates the Create-side widening. On main, `NativeDoltStore`'s
`shouldPrevalidateNativeDependency` (`native_dolt_store.go:2010-2020`) already skips
cross-prefix parents, so `gc bd create --parent <other-store-id>` was **not** refused on
the Dolt store before this PR. The genuine behaviour change is (a) the **Update** arm —
`validateUpdateParent` / `updateParentInTransaction` now apply the same rule Create does,
so a reparent onto a foreign id no longer comes back "bead not found" — and (b) the
prefixless edge: with `nativeParentIsLocal`, a store with no prefix (or a parent id with
no `-`) never resolves, so a dangling same-store parent is now accepted on Update there.
Worth a reviewer's eye on that second point specifically.

Also worth flagging: the new conformance subtest runs against every `RunStoreTests`
caller, including `test/integration/bdstore_test.go` (real `bd` over Dolt, integration
tag) and `internal/beads/exec/br_test.go`. Those shards were not run locally — an
integration-shard result is the residual risk.

Label `status/reviewing` (a dead state) removed.

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277290592

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277290661

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277290730

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277290888

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR turns two implicit store behaviours — the pinned-id fence and the weak `ParentID` boundary — into stated provider contracts backed by conformance coverage, and along the way fixes five concrete defects: the SQLite fence tested membership against a *trimmed* id rather than the id as stored; `NativeDoltStore` admitted a cross-ledger `ParentID` on create but refused it on reparent; the two residency planes had drifted apart because the binding body was written out twice; the relocation refusal guarded only the namespace a class *mints* rather than the ones it *holds*, so a `gcnq-` read silently drained to bd and came back empty; and `newCityExtMsgServices`'s error arm silently degraded to the work store. That matters because every one of these was a silent-wrong-answer failure mode rather than a loud one, and the new conformance rows mean a future provider can't reintroduce them without a red test.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **Fence membership now describes the id that actually lands.** Removing the fence's `strings.TrimSpace` — rather than adding one to the writers — is the correct side to change, because the mint decision and `normalizeCreate` both compare untrimmed. Pinned by `MembershipIsTestedOnTheIDAsStored`.
- **Create and Update now agree about a foreign `ParentID`.** `nativeParentIsLocal` is applied on all three write paths, so a store can no longer admit a cross-store parent and then fail only on the reparent.
- **The two residency planes converge.** The CLI half is a pure move into `storeref.BuildBindings`; the API half gains only a stable sort by `Leg.Ref`, converging on the order the CLI already produced, with refusal capture unchanged on both planes.
- **The relocation refusal covers held namespaces, not just minted ones.** The exact-or-hyphen match in `matchesID` is load-bearing here — `"gcn"` is a proper prefix of `"gcnq"`, so a bare `HasPrefix` would have swallowed the nudge-queue namespace — and deriving `nudgeQueueIDPrefix` from `config.NudgeQueueIDPrefix` removes the second drift-capable literal.
- **A wrong-ledger fallback became a named refusal.** `refusedExtMsgServices` is diagnosable at the call site, where the silent form was not.

Review did not take the PR narrative at face value: each of the five stated bugs was re-derived against its base copy and confirmed to be reinstated by reverting the fix. `go build`, `go vet`, and `go test` were re-run across `internal/beads`, `internal/storeref`, `internal/storebinding`, `internal/config`, and `./cmd/gc`, all clean.

Review findings addressed:
- **A conformance row hardcoded a `gcg-` foreign parent without asserting disjointness** — the maintainer fixup adds a `t.Fatalf` guard, placed after the `control` create that reveals the mint namespace and before the `child` create that is the row's real subject, so a store whose namespace collides stops rather than going green about the wrong shape (`internal/beads/beadstest/conformance.go`).
- **The `Bead.ParentID` doc over-claimed conformance coverage** — the same fixup scopes the doc to the providers that actually run `RunStoreTests` and records the bd caveat matching the existing skip (`internal/beads/beads.go`).

- `ga-ob8z9` (minor, test evidence) — pin `NativeDoltStore`'s refusal of a prefix-mismatched pinned create. `nativeParentIsLocal` derives locality from the *child's* id prefix rather than the store's served namespace; that's safe today only because `NativeDoltStore` doesn't implement `beads.ForeignIDCreator` and no path can land a foreign-prefixed row there. A test turns that unstated cross-layer invariant into an executed assertion, so adding `ForeignIDCreator` later fails loudly instead of silently disabling the parent check.
- `ga-0aw3v` (minor, test evidence) — pin `BuildBindings`' refusal order against the `Leg.Ref` sort order. The reported refusal follows `order` while the returned bindings follow `Leg.Ref`; with a single refusing store the two coincide, so nothing in the new suite distinguishes them. One row with two refusing stores whose orderings disagree would pin it.

Two further risks surfaced during review are pre-existing and already tracked separately — neither is introduced by this PR: `ga-0vtct` (a binding serving `work` alongside reserved classes opens unfenced) and `ga-e97fq` (`matchesID` is case-sensitive while `checkPinnedIDNamespace` folds case).

### Maintainer Changes
One maintainer fixup was pushed directly to this PR branch:

- `fixup(beads): assert the conformance row's disjointness premise; scope the ParentID doc` — `internal/beads/beads.go`, `internal/beads/beadstest/conformance.go`; 2 files changed, 13 insertions(+), 2 deletions(-).

### Final Review Status
Ready for the merge queue: final head `1f61a5f094` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/33746398149

### Review Iterations
2 review passes. The first pass approved the change and raised the two minor test/doc gaps above, plus one out-of-scope finding on `engineReservedPrefixes` that was de-gated after verifying the flagged code is byte-identical between base and head. The maintainer fixup addressed both in-scope minors; the second pass re-validated every fix and closed with the two new minor findings, both filed as follow-ups rather than gating this merge.

---

PR #5634, opened 2026-08-26
fix(cli): let a clean class-binding miss fall through in gc bd — ownership is residence, not prefix

julianknutsen:
**Rebased: this PR now targets `main` directly.** The old base
(`residency/pr1-storeref-byid`) is gone from the history — this branch was
rebuilt as a fresh branch off `origin/main` carrying only this PR's four unique
commits. Nothing from the residency stack (s3/s5/s8/pr1) is required: `main`
already has `cliByIDBindingOwner`/`beadForOwner`, the API-side residence
doctrine (`internal/api/residency_by_id.go`), and every test helper this branch
needs.

The `gc bd` by-id door refused **every** reserved-prefix id the relocated class
binding did not hold, on the premise that the binding is the **sole minter** of
its namespaces.

It is not. `config.ValidateRigs` deliberately does not reject a rig prefix inside
a reserved namespace, and `config.ReservedPrefixWarnings` only advises — so a rig
configured with one mints work beads carrying ids the binding has never held. On
a split city, every read and every write addressed at one of those ids died at
this door, **including the step-completion write a worked bead ends with.**

The API and `classRoutedStoreForID` already served those rows. This door was the
only surface refusing them.

## The correction

It now answers the same question they do: **ownership is RESIDENCE.** A reserved
prefix decides which ids get *asked about*, not what the answer is.

- `serveBdByIDResolved`: a binding miss returns `(0, false)` for every id, and it
  runs BEFORE the `--rig` refusal, whose whole claim is false when the binding
  never answered for the bead.
- `refuseUnservedClassMutation`: the reserved-prefix branch is probed for
  residence like the work-shaped one, instead of refusing on the prefix — and
  **every** addressed reserved id is probed, not just the first, so a
  multi-subject `dep add <miss> <resident>` is refused in either argv order
  (`ga-zvetw`'s order dependence).
- `bdByIDResolution` loses `Reserved`/`Owned` — the second bit that could
  disagree with `Found` — and `printBdByIDNotFound` goes with it.

**A miss is not a shrug.** Only an *answer* of "absent" reaches the fall-through:
the plan carries the authority leg as `PolicyFatal`, so an unopenable, refusing
or faulting binding leaves through the error arm and is never read as absence.

## What changed versus the reviewed head

The two "approve, accepted as-is" reviews cite head `395159603a`, which is **no
longer in this branch** — it was rebuilt after approval and then extended with
`74bd730acf` (the `ga-zvetw` multi-subject fix), whose CI run was red. So this is
[...]

PR #5644, opened 2026-08-26
fix(beads,storeref): fence pinned ids in every provider, and count closed relics in the retirement verdict

julianknutsen:
## What this is now

**Rebased onto `main` on 09-05.** Base was `residency/s4-falsify-and-guard`
(#5643); s4 has since merged to `main` as squash `7746bbe5ab`, and s8 (#5597)
merged as squash `fc7a23a292`, so the base is now `main` and every SHA below is
new. Details in "Rebase onto main" further down.

This PR was **re-cut** before that. Its head at the time (`1806b45dfc`) had
silently absorbed two other segments — #5645 (s6) and #5646 (s7) — as
squash-merges into this branch on 08-28. That head was 74 files / +5602/−2322
and carried a `cmd_wait` rewrite, a drain rewrite, a claim-route collapse, a
`config`→`beadmeta` move, an on-disk status-file memo and a five-file guard
split, none of which this PR's title or body ever described. Re-presenting that
as one review was not reviewable.

**This head carries s5 only.** s6 and s7 follow as their own PRs, stacked in
order, each retargeted as its parent merges:

- **PR A — this one.** Fence every provider; count closed relics in the
  retirement verdict; collapse the by-id door onto the plan.
- **PR B — s6** (`residency/s6-retire-hand-rolled-probes-v2`, base = this
  branch). The claim-route collapse, the drain member resolution, the
  `config`→`beadmeta` move, and the last hand-rolled residence probes.
- **PR C — s7** (`residency/s7-guard-by-rule-v2`, base = PR B). Pure code
  motion of the residency guard into per-rule files.

s3's `ErrPinnedIDOutsideNamespace` and `RunPinnedIDFenceConformance`, which this
segment builds on, are on `main` via #5642/#5643.

**Hashes differ from every earlier form of this branch.** Each commit carries
two `-x` trailers: the pre-rebase SHA it was picked from, and — from the earlier
re-cut — the original. Nothing here is byte-identical to what #5597 reviewed.

## Substance

**Closed relics count in the retirement verdict (ga-qdt5y.19).**
`storeref.OpenLegacyResidents` keeps its open-only shape as the DRAIN GAUGE that
`gc storage status` prints, and a new `LegacyResidents` / `HasLegacyResidents`
pair — `IncludeClosed: true` — becomes the retirement condition the boot census
sets. The two parted company on purpose. `gc storage migrate` never deletes the
work store's pre-migration copy, and a closed relic is still shown, reopened,
claimed and written by id — so retiring the probe when the last relic CLOSED
would answer that id from a frozen copy that reads OPEN forever. Because nothing
ever deletes a relic the widened verdict is MONOTONE, which is what makes reading
it once per process sound rather than merely cheap. `cmd_storage.go` keeps the
open-only count with wording that says which number it is.

**The `gc bd` by-id door runs the shared plan (ga-qdt5y.18).**
`bdByIDClassDoor.resolve` was an unconditional `Graph.Get` with a hand-rolled
`ErrNotFound` / standing-refusal switch, and had never heard of the boot census —
so the retirement s2 built was unrealised on the busiest one-shot path in the
CLI. It now gains `CityPath` and collapses onto `cliByIDBindingOwner` +
`beadForOwner`, with the operator-facing binding name applied by a `readFailure`
wrapper. There is one implementation of "does this binding hold this id" that
knows about the census, and this asks it rather than recomputing it.
(`hookClaimClassRoute.holds` is the last census-blind probe; ga-qdt5y.16 tracks
it, and the doc says so rather than claiming the divergence is closed.)

**Invariant 16 now holds for every provider.** A shared
`internal/beads/pinned_id_fence.go` (`checkPinnedIDNamespace` +
`collectReservedIDPrefixes`) backs both stores that serve class bindings.
`NativeDoltStore` gains `reservedPrefixes`,
`WithNativeDoltStoreReservedIDPrefixes`, and `CreateWithForeignID`
(`beads.ForeignIDCreator`) so the migration copy still has its one sanctioned way
across the fence; the fence is applied in `create()` and in `applyCreateInTx`.
`sqliteStoreTx.Create` was unfenced and now is not — a fence on `Store.Create`
that a transaction walks around is not a fence.

**`EngineReservedPrefixes` fixes the any-class-unfences bug.**
`beads_engine.go`'s local `engineReservedPrefixes` dropped the fence for a whole
binding whenever ANY co-assigned class had no registered prefix. The rule is
about the WORK class specifically — work ids carry whatever prefix an operator
configured, so a binding serving work claims nothing — and reading an empty
prefix list as "unfenced" fails in the silent direction. The derivation moves to
`internal/storebinding/engine_fence.go`, above every provider, so two providers
cannot answer it differently for the same assignment;
`beadsworkspace/engine.go` now passes it too.

**The `splittest` double is held to the provider contract.** `StrictStore` gains
namespaces, `NewClassStore` is fenced to `config.ReservedClassPrefixesFor`,
`newStrict` refuses a SQLite-semantics leaf that mints a reserved prefix with no
namespaces, and `beadstest.RunPinnedIDFenceConformance` runs against the double.
A test double that accepts writes production refuses is a green suite over a
broken invariant.

**Two carried-in test debts.** `classResidentWorkShapedBead` now owns the
post-seed recensus instead of leaving it to callers (ga-qdt5y.12) — seeding a
relic is out-of-order by construction, and four rows were passing on a stale
verdict. And the `cliMailStore` collapse (ga-qdt5y.6) gets the evidence that
never landed with it.

## Rebase onto main

Branched fresh from `origin/main` (`fc7a23a292`) and cherry-picked `-x` the
seven commits that are this PR's own. s4's three (`7a8c0ac228`, `d2d14fa7a3`,
`6a67693ea9`) were dropped: they are inside `main`'s squash `7746bbe5ab`, which
also carries the maintainer fixup `d96195272f` that was never on the s4 head
this branch was originally cut from.

| new | pre-rebase | original | subject |
|---|---|---|---|
| `6e20220d93` | `360c0b133e` | `406172c2fb` | test(storeref): fold the relic recensus into the seed helper (ga-qdt5y.12) |
| `f26305e3dd` | `55ff9d01c4` | `b1f80c9e1e` | fix(beads): fence pinned ids in the beads-workspace provider too |
| `04fbd95abb` | `b21a18eaec` | `8f9eeee96b` | fix(storeref): count closed relics in the residence-probe verdict, and collapse the by-id door onto the plan |
| `c9ce61a1aa` | `acc6c0bd8b` | `c0bb428b0c` (test half) | test(gc): pin the collapsed cliMailStore accessor and its unwrap |
| `112a32582b` | `14cf7fdb8c` | `6ef947e4be` | test(splittest): fence the class-store double to the namespaces its binding claims |
| `da3f0cf8e6` | `4f5ee8a37f` | `a428e0dd9a` | test(beads): apply the fence review council's findings to the split kit |
| `b32bff3c03` | `55f1635e59` | — | test(gc): pin the by-id door's residence and its write to one binding handle |
| `5e001e9467` | — | — | fix(scripts): re-attribute the infra_class_migrate baseline row #5597 moved |

46 files, +2160/−371 against `main`. `c0bb428b0c` is still carried as its test
half only, for the reason the previous body gives: its production half is on
`main` via #5488.

### Conflicts

**One**, in `cmd/gc/infra_class_migrate_test.go` — an add/add tail where s8's
squash and this PR each appended to the same file. **Both sides kept**, verbatim
and in full: s8's `populatedInfraBinding`, `TestTheBindingDiagnosticsOpenerCannotWrite`,
`leaveInfraBindingWALResident`, `convergedInfraBindingRequest` and
`TestTheReadOnlyDiagnosticsLeaveALiveWALIntact`, then this PR's
`TestInfraMigrationClassesCoverEveryClassTheBindingsClaim`. The two sides touch
nothing in common; the only textual overlap was the closing braces the merge
driver had folded together. `go test ./cmd/gc -run
'TestTheReadOnlyDiagnosticsLeaveALiveWALIntact|TestTheBindingDiagnosticsOpenerCannotWrite|TestInfraMigrationClasses|TestEnsureInfraClassMigrated'`
→ 23 PASS, 0 FAIL, so both sides run.

The other files the rebase was expected to fight over merged cleanly, and each
was read after the fact rather than trusted:

- **`cmd/gc/cmd_storage.go`, `cmd/gc/storage_boot.go`** (s8 touched both) — s8's
  additions are untouched; this PR's are the drain-gauge wording on
  `reportBindingRelics` and `censusBindingRelics`'s switch to
  `storeref.HasLegacyResidents`.
- **`cmd/gc/cmd_bd_by_id.go`** — the collapsed `resolve` applied as written.
- **`internal/beads/sqlite_store.go`**, **`internal/storebinding/sqlite/beads_engine.go`**
  — the `sqliteStoreTx.Create` fence and the `EngineReservedPrefixes` hoist
  applied as written.
- **`d96195272f`** (s4's maintainer fixup, now inside `7746bbe5ab`) — no
  conflict to resolve. It touches
  `scripts/check-residency-boundary.{sh,patterns}`,
  `scripts/residency_boundary_test.go`, `cmd/gc/agent_home_worktree_cleanup.go`
  and `cmd/gc/bead_worktree_reaper.go`, and this PR touches none of the five. Its
  substance is honoured rather than merely un-conflicted: the new
  `ast:uncounted-call-spelling` rule it added runs over this tree and emits
  exactly the five rows `main` pins, which is checked below by regenerating the
  whole baseline rather than by reading.

Nothing s8 landed was dropped, and nothing this PR adds duplicates something s8
already landed. In particular the closed-inclusive census helpers
(`LegacyResidents` / `HasLegacyResidents` / `legacyResidents`) are **not** on
`main` — `internal/storeref/relic_census.go` there has only `OpenLegacyResidents`
and `HasOpenLegacyResidents` — so there is one copy, this PR's, and no dedupe
was needed. `go build ./...` and `go vet ./...` are the mechanical check on that:
a second copy of any of them would be a redeclaration.

### Baseline

**One row re-keyed, and it is `main`'s, not this segment's.**

```
-cmd/gc/infra_class_migrate.go	openInfraDestination	  d:ReservedClassPrefix	1
+cmd/gc/infra_class_migrate.go	infraBindingIDPrefix	d:ReservedClassPrefix	1
```

`origin/main` at `fc7a23a292` **fails its own residency guard**: #5597 lifted the
destination prefix read out of `openInfraDestination` into a new
`infraBindingIDPrefix()` helper while #5643 pinned the old key, so both halves
report — `./scripts/check-residency-boundary.sh` exits 1 and
`TestResidencyBoundaryGrepRatchet` FAILs under `go test ./scripts`. Reproduced on
an untouched `git archive origin/main` export, so it is base-owned; this branch
neither caused it nor can be green over it. **A separate one-row fix PR against
`main` is open in parallel — this branch also carries the re-key so the guard is
green on this tree, and that commit (`5e001e9467`) drops out on rebase once the
fix PR merges.**

Regenerated by the four-step splice the baseline header documents — all three
emitters collected (`--emit-baseline` minus its four-line PARTIAL warning,
`TestResidencyResolverBoundary`, `TestResidencyASTExpressionRatchet`, the two Go
emitters filtered to exactly-three-tab lines), concatenated, `LC_ALL=C sort`ed
once, header kept and its prose updated, both halves re-run clean.

That regeneration is also the evidence for the segment's own claim: the full
114-row set derived from **this** tree differs from `main`'s file in exactly the
one row above. **s5 retires no scanned site and adds none** — 114 rows before,
114 after, `cmd/gc/cmd_bd_by_id.go resolve d:bdIDIsClassReserved 1` still held
(the collapsed `resolve` still sets `Reserved` from `bdIDIsClassReserved`), and
the ratchet fails on a pinned count the tree no longer reaches, so an unnoticed
retirement would have been loud.

### Deferred reconciliations

Both of the two items the review asked to reconcile before labelling depend on a
PR that has **not** merged, so neither is done here and neither is silently
dropped:

1. **pr8 #5634 (`residency/pr8-bd-byid-door`) is still OPEN.** It removes
   `bdByIDResolution.Reserved` and makes a binding MISS fall through for every id
   — residence, not prefix. This PR keeps `Reserved` and routes `resolve` through
   `cliByIDBindingOwner` + `beadForOwner`, exactly as reviewed. Both are right and
   the merged form is the plan-based residence walk from here with no `Reserved`
   bit; **that reconciliation, and the drop of the `cmd/gc/cmd_bd_by_id.go resolve
   d:bdIDIsClassReserved` baseline row that goes with it, are deferred to
   whichever of the two lands second.**
2. **#5988 (s3 follow-up, `fix/s3-followup-provider-contracts`) is still OPEN.**
   It adds the same `t.store.checkPinnedIDNamespace(b.ID)` call before
   `normalizeCreate` in `sqliteStoreTx.Create` that this PR adds, with a different
   comment. **No Tx-fence dedupe is needed now**; whichever lands second takes the
   one-hunk conflict and keeps a single copy.

## Dolt-gated evidence

The `NativeDoltStore` fence rows only run under the Dolt env gate, so a local
reviewer running bare `make test` sees them skip. Both forms were run on this
head and **RAN — zero skips**:

```
BEADS_TEST_ENV_RUN_DOLT=1 go test ./internal/beads \
  -run 'NativeDoltStore.*Fence|ForeignIDCreator' -count=1 -v
  → ok  0.113s   19 RUN, 19 PASS, 0 SKIP, 0 FAIL

go test -tags integration ./internal/beads \
  -run 'NativeDoltStore.*Fence|ForeignID' -count=1 -v
  → ok  0.116s   19 RUN, 19 PASS, 0 SKIP, 0 FAIL
```

Covered: `TestNativeDoltStorePinnedIDFenceConformance` (16 of the 19, including
the new `TheFenceCoversTheTransactionalCreate` row and
`OverRestrictionControls/TheCONFIGUREDNamespacesAreNormalizedToo`),
`TestNativeDoltStoreIsAForeignIDCreator` — which exists so the two conformance
rows that depend on the capability cannot silently skip — and the two
`CloseWithMetadataIfMatch` fence rows the pattern also selects.

**Correction to the previous body, which claimed 23 for the env form.** Both
regexes select the same 19; `ForeignIDCreator` and `ForeignID` differ only in a
suffix and match the same single test. That 23 was wrong when it was written, not
a row this rebase lost: the same command on the pre-rebase head (`55f1635e59`,
exported clean) also emits 19, and the two `=== RUN` name lists are byte-identical.

## Gates

Run on `5e001e9467`.

| gate | command | result |
|---|---|---|
| format | `gofmt -l cmd internal scripts test` | exit 0 — no output |
| lint | `golangci-lint run ./...` | exit 0 — `0 issues.` |
| residency guard | `./scripts/check-residency-boundary.sh` | exit 0 |
| residency guard control | `./scripts/check-residency-boundary.sh --self-test` | exit 0 |
| touched packages | `go test ./internal/beads/... ./internal/storebinding/... ./internal/storeref/... ./scripts/... -count=1` | exit 0 — 13 `ok` |
| cmd/gc slice | `go test ./cmd/gc -run 'ByID\|Residency\|Relic\|Fence\|SplitTopology\|Storage\|Handoff\|Mail' -count=1 -v` | exit 0 — **466 PASS**, 0 FAIL, 1 SKIP |
| merged conflict file | `go test ./cmd/gc -run 'TestTheReadOnlyDiagnostics...\|TestInfraMigrationClasses\|TestEnsureInfraClassMigrated' -count=1 -v` | exit 0 — 23 PASS, 0 FAIL |
| Dolt fence (env) | see above | exit 0, 0 skips |
| Dolt fence (tag) | see above | exit 0, 0 skips |
| docs | `make check-docs` | exit 0 |
| fast suite | `make test-fast-parallel` | exit 0 — `All fast jobs passed` |

466 PASS against the previous head's 458: the eight are s8's, which the same
filter now selects. The single SKIP is
`TestCmdMailInbox_NormalizesCanonicalManagedProviderEnvAndReadsInbox`, which
needs a real Dolt server and is skipped on `main` the same way.

**Nothing red was left behind and `--no-verify` was not used.** The one red found
was `main`'s residency ratchet, described under Baseline above; it is base-owned,
was reproduced on an untouched `origin/main` export, and this branch carries the
one-row fix so the pre-push hook's fast suite passes on its own merits.

## Falsification

The mechanisms here are pinned by rows that go red when reverted, not by reading:

- `TestAClosedRelicKeepsTheProbe` fires if `IncludeClosed` goes back to false;
  `TestAClosedInNamespaceBeadIsNotARelic` is its control.
- `TestClosingTheLastRelicDrainsTheCountAndKeepsTheProbe` and
  `TestBootCensusKeepsTheProbeForAClosedRelic` pin the gauge/verdict divergence
  end to end through the boot gate.
- `TestBdByIDDoorSkipsTheProbeOnACensusCleanBinding` (exactly-one-read) and
  `TestBdByIDDoorReachesAClosedRelic` falsify a door that still `Get`s
  unconditionally, or a census that retires on close.
- `TestOnlyWorkUnfencesABinding` / `TestEveryNonWorkClassHasAFenceableNamespace`
  pin the work-only unfence rule against the any-class-unfences behaviour.
- `TestNewStrictRejectsAnUnfencedClassLeaf` and
  `TestStrictStorePinnedIDFenceConformance` hold the double to the provider
  contract.
- `TestSeedingARelicLeavesTheBindingCensusedAsRelicBearing` catches removal of
  the seed helper's funnel reopen.

Two rows added here were falsified by mutation while writing them: dropping
`.Store` from one of `cmd_handoff.go`'s two roots turns the unwrap tripwire red
with `unwraps 1 of its 2`; wrapping the door's store at
`openBdByIDClassFrontDoor`, or the owner's store inside `cliByIDBindingOwner`,
turns the handle-identity pin red on the fixture guard and on both leg rows
respectively.

## Known follow-ups, not in this PR

- **ga-qdt5y.16** — `hookClaimClassRoute.holds` is the last hand-rolled,
  census-blind residence probe. It collapses in PR B.
- **ga-qdt5y.20 / #5602** — the closed-inclusive census reads the binding's whole
  history with `AllowScan` once per process (~82 ms at 10k rows on s6's bench).
  This PR ships the widened verdict, not a memo for it. The s6 head carried an
  on-disk `.gc/storage-relic-census.json` status file for that; it is NOT here,
  and per `AGENTS.md` ("no status files — query live state") PR B will memoize
  in-process instead. #5602 unblocks once this lands.
- `internal/beads/splittest/pinned_id_fence.go` is a deliberate second spelling
  of the production fence, and anti-drift rests on the conformance suite running
  against the double. The two already differ in op text (`strict create` vs
  `sqlite create` / `native dolt create`), so a future conformance row asserting
  on the message prefix would diverge.

- **#5634 (pr8) and #5988** — the two deferred reconciliations spelled out under
  "Deferred reconciliations" above. Neither has merged; each is a one-hunk
  merge for whichever of the pair lands second.
- **The `main` baseline re-key** carried as `5e001e9467`. It belongs to `main`,
  not to s5, and drops out on rebase once the parallel one-row fix PR merges.


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277238362

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277243294

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277290302

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277290376

julianknutsen:
Fable review of PR A (head 55f1635e59, base residency/s4-falsify-and-guard @ 6a67693ea9): **approve, hold the pipeline label until #5643 (s4) merges**, then retarget to `main` and reconcile these two items before labelling:

1. **`cmd/gc/cmd_bd_by_id.go` `resolve` vs pr8 #5634 (in the pipeline ahead of this).** pr8 removes `bdByIDResolution.Reserved` and makes a binding MISS fall through for every id (residence, not prefix). This PR keeps `Reserved` and routes `resolve` through `cliByIDBindingOwner` + `beadForOwner`. Both are right; the merged form is: the plan-based residence walk from this PR, no `Reserved` bit, miss = fall-through as pr8 states. Also drop the `resolve d:bdIDIsClassReserved` baseline row at that point (the landing plan kept it until this segment's resolve body was adopted).
2. **`sqliteStoreTx.Create` fence vs #5988 (s3 follow-up, past review).** Both add the same `t.store.checkPinnedIDNamespace(b.ID)` call before `normalizeCreate` with different comments; expect a one-hunk conflict, keep one copy.

Substance checked: shared `checkPinnedIDNamespace` (separator-exact, case-folded, mint exempt); NativeDoltStore fenced on `Create`, `applyCreateInTx` and unfenced only through `CreateWithForeignID`; `EngineReservedPrefixes` hoisted above both providers with the work-class rule stated on the class (an unregistered non-work class contributes nothing and stays refused, guarded by `TestEveryNonWorkClassHasAFenceableNamespace`); closed-inclusive `LegacyResidents` is the retirement verdict and monotone, `OpenLegacyResidents` stays the drain gauge; `storage_boot` census uses the wide verdict; handle-identity pin added instead of softening the CityPath doc. Dolt-gated fence tests ran (23/19, 0 skips). Baseline unchanged (0 rows). One dropped test, `TestCLIEmissionKeepsTheCensusVerdict`, is not on `main` either; its invariant (the emitting wrapper must not change the `hasLegacyResidents` key) fails safe (unknown → relics), so no coverage regression against main.

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277304380

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277305503

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR extends the pinned-id fence
to the remaining store providers and widens the residence-retirement verdict to
count closed relics, which closes the last paths where an id could be pinned
outside the namespace its binding claims, or where a binding could retire while a
frozen pre-migration copy still answers that id.

branch.

### Original PR Review

Decision: approve.

Specific gaps fixed:

- The retirement condition in `storeref` was open-only, so a binding could retire
  once its last legacy resident CLOSED — even though `gc storage migrate` never
  deletes the work store's pre-migration copy, so that id would still be answered
  from a frozen copy that reads OPEN forever. The final patch parts the drain
  gauge (`OpenLegacyResidents`, still what `gc storage status` prints) from the
  retirement condition (`LegacyResidents` / `HasLegacyResidents`, with
  `IncludeClosed: true`). Review confirmed the widened verdict is monotone —
  nothing ever deletes a relic — which is what makes reading it once per process
  sound rather than merely cheap.
- The `gc bd` by-id door (`bdByIDClassDoor.resolve`) was an unconditional
  `Graph.Get` behind a hand-rolled `ErrNotFound` / standing-refusal switch that
  had never heard of the boot census, leaving the retirement work unrealised on
  the busiest one-shot path in the CLI. It now collapses onto
  `cliByIDBindingOwner` + `beadForOwner`, so there is one implementation of "does
  this binding hold this id" and the door asks it instead of recomputing it.
- The pinned-id fence landed in an earlier segment but was not wired into the
  beads-workspace provider or the class-store test double, so those two paths
  could still pin an id outside their binding's namespace. Both are now fenced,
  and the split test kit gained the corresponding conformance coverage.
- Two store-lifecycle doors — `OpenGraph` in
  `internal/storebinding/sqlite/graph_frontdoor.go` and its sibling open in
  `beads_provider.go` — are deliberately left unfenced. Review censused their
  callers and confirmed both are production-unreachable at this head, so the
  obligation now lands on whoever gives them a first production caller; see the
  tripwires below.

Review findings addressed:

- Neither unfenced lifecycle door carried an in-code marker, so the first person
  to give one a production caller would have had nothing to read. Invariant-16
  tripwires were added at both, each naming `EngineReservedPrefixes` and the exact
  wiring to copy — commit `39597cc`.
- The `openInfraDestination` escape hatch was undocumented, and its first draft
  comment asserted "Every write through it must go through
  `CreateWithForeignID`", which review found false in both directions: the handle
  legitimately takes `prepareInfraDestination`'s clearing deletes and
  `infraCopyDepEdge`'s edge rows, and the parenthetical named one of exactly two
  sanctioned create sites. The landed text narrows the quantifier to id-pinning
  creates and names both callers (`importInfraSnapshot` and the stranded-row
  recovery in `infra_class_recover.go`) — commit `39597cc`.
- The `OpenGraph` tripwire cited `p.path`, a receiver that file defines nowhere,
  so a reader meeting the tripwire standalone could not resolve it locally. It now
  names the shared locator `GraphPath(BindingRoot(spec))` together with
  `storebinding.EngineReservedPrefixes(classes)`, both of which resolve inside the
  file — commit `39597cc`.
- Two of the new comments cited `ga-qdt5y.16` as a live tracker while that bead
  was closed. Discharged by reopening the bead rather than by editing the
  citations, which makes four citations true at once — the two new comments plus
  the two pre-existing `scripts/residency-boundary-baseline.txt` rows — where
  editing would have left the baseline rows dangling.

- [Architectural Consistency / efficiency, major] The relic census walks full
  history on the by-id funnel with no bound, so it will not scale as large cities
  converge. Mitigated in-code for now and tracked rather than gating; both
  prescribed remedies (a pushdown existential predicate, or a short-circuiting
  paginated `SeekAfter` walk consumed by `HasLegacyResidents`) are recorded on the
  bead. `ga-1ghfh`: Bound the relic census to an existential probe before large
  cities converge.
- [Architectural Consistency, minor] `hookClaimClassRoute.holds` is the last
  census-blind residence probe — still an unconditional check that does not
  consult the boot census the rest of this series routes through. `ga-qdt5y.16`:
  Migrate the three hand-rolled probe sets (drain.go x2, cmd_wait.go) onto
  storeref.
- [Test Evidence Quality, nit] The beads-workspace provider's fence argument is
  proven only in the integration tier, so the wiring has no fast-tier pin.
  `ga-e9ynk`: Pin the workspace provider's fence argument outside the integration
  tier.
- [Release Safety / CI hygiene, minor] The advisory Mac Regression job is red on
  this head. Traced to a cleanup race in a base test this PR does not touch:
  `TestCustomTypesCheck_TableDrift` calls `t.TempDir()` twice and only the second
  directory is wrapped in `retryRemoveAllForTest`, so the unguarded first one
  loses the race against the embedded-dolt background writer that outlives the
  `bd` subprocess. `ga-531fk`: TestCustomTypesCheck_TableDrift: the test-owned
  HOME TempDir is outside the retryRemoveAll guard, so Mac flakes on cleanup.

### Maintainer Changes

One maintainer commit on top of the eight contributor commits:

- `39597cc docs(storebinding): flag the unfenced lifecycle opens and the fence's
  escape hatch` — comment-only, 3 files changed, 28 insertions(+), 0 deletions(-)
  across `cmd/gc/infra_class_migrate.go`,
  `internal/storebinding/sqlite/beads_provider.go`, and
  `internal/storebinding/sqlite/graph_frontdoor.go`.

The branch was additionally rebased onto current `main` to resolve a real conflict
with #5626, a sibling in the same residency series. Three resolutions, each folded
into the commit that introduced the incompatibility rather than stacked at the tip,
so every commit in the series still builds:

- `internal/storeref/topology.go` — composed both sides of `probeRetired`,
  keeping this PR's widened "holds no legacy resident, open or closed" doc and
  main's new `holdsLegacyResidents` body. Either side alone would have dropped
  something real: ours drops main's `KnownLegacyResidents` implication, theirs
  drops the closed-inclusive widening this PR exists to land.
- `internal/storeref/relic_census.go` and `relic_census_test.go` — this PR renames
  `HasOpenLegacyResidents` to `HasLegacyResidents`, and main's new code documents
  the old name, so git merged clean while leaving prose naming a symbol that no
  longer exists. The test site was a mechanical rename; the
  `ProvenLegacyResidents` doc was not, because its "one line to change when the
  census stops excluding closed beads" claim became false the moment the
  closed-inclusive form existed. Rewritten to state both differences. The actual
  one-line behaviour swap it advertises was deliberately left undone — that is a
  change, not a conflict resolution.
- `cmd/gc/by_id_relic_proof_test.go` — main's new file calls
  `classResidentWorkShapedBead` with the pre-rename signature, a build break with
  no git conflict. Six call sites adapted using the exact pattern this PR already
  applied to `by_id_binding_owner_test.go`; no assertion changed.

`git range-diff` against the pre-rebase reviewed head: 7 of 9 commits are
byte-identical, and the entire delta in the other two is the three resolutions
above.

Total adopted diff: 63 files changed, 2444 insertions(+), 1625 deletions(-).

### Final Review Status

Ready for the merge queue: final head `39597cc` has passing required GitHub

All six required contexts are green — `Check`, `Analyze (actions)`, `Analyze
(go)`, `Analyze (javascript-typescript)`, `Analyze (python)`, and `CI / required`.
The only red is the advisory Mac Regression workflow, dispositioned above as
`ga-531fk`.

Because `main` moved the resource census while this branch adds test files, the
refreshed head was re-validated locally before being marked ready rather than
resting on the existing green run: the two changed-file sets do not overlap,
`go vet` is clean, 11 packages report `ok` with 0 failures, the resource-census
tests pass, and `scripts/check-residency-boundary.sh` exits 0 — that ratchet is
two-sided, so it would have caught drift in either direction.

CI: https://github.com/gastownhall/gascity/actions/runs/34174946196

### Review Iterations

2 review passes. Iteration 1 raised the two lifecycle-door tripwires, the
escape-hatch documentation, the closed-tracker citation, and the census follow-up.
Iteration 2 reviewed the resulting comment-only fixup, caught that both the
escape-hatch text and the `OpenGraph` tripwire had each shipped a false clause,
and approved once both were corrected. A separate repair pass then rebased the
branch onto current `main` after the sibling PR landed underneath it.

---

PR #5645, opened 2026-08-26
refactor(gc): collapse the last hand-rolled residence probes onto the resolver

julianknutsen:
**Segment 6 of 8 — split out of #5597 at the reviewer's request.** Base: segment 5.

**The last four hand-rolled residence probes go.** Segments 1 and 2 converged the
by-id readers and the writers; these are the stragglers, each of which had grown
its own answer to "which store owns this id" and each of which was wrong in a
different way.

## The four

**`ab631d25f6` — the claim route.** `gc bd claim` probed residence itself. It now
uses the resolver, which also means a refused or faulting binding reaches it as an
error rather than as absence.

**`25ac8b261d` — one suspended rig held every wait in the city asleep.**
`gc wait` walked the rigs and treated a store it could not open as a reason to
keep waiting. One suspended rig therefore blocked every waiter in the city
indefinitely, including waiters with nothing to do with that rig. The wait now
scopes to the stores the waited-on beads actually reside in.

**`d7fd6d1b7f` — the drain reserved one copy of a member and served another.**
`internal/dispatch`'s drain resolved a convoy member twice: once to reserve it,
once to serve it, through two different lookups. On a split city those could land
on different copies of the same id — reserve the binding's row, serve the work
ledger's frozen twin — so the drain believed it held a reservation it did not
hold. This is the most consequential fix in the segment.

**`ed01d02acb` — a relocated dependency slept its waiter forever.** On the
one-shot plane, a dependency resident in the binding was invisible to the waiter,
which read "not there" as "not yet" and slept. Forever, because nothing was ever
going to move the bead into view.

## Two supporting changes

`3b1594dad1` moves the reserved-prefix table to `internal/beadmeta`, a
stdlib-only leaf. It lived in `internal/config`, which meant every consumer of the
prefix table — including `internal/coordclass` and `storeref` — imported the
config package to read a static table. The leaf has no dependencies, so the
layering invariant stops being strained by a lookup.

`1b070b2d33` stops re-scanning a binding that already answered. The relic census
from segment 2 is a boot-time fact, but the boot path was recomputing it per
binding per open; the memo makes it once. Includes the bench that shows it.

## Verification

Refs `ga-qdt5y.4`.


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277238289

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277243217

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR collapses the last four hand-rolled store-residence probes onto the shared `internal/storeref` resolver, so every surface that asks "which store actually holds this bead?" asks it exactly once, the same way — and along the way it fixes three genuine liveness bugs and a drain reserve/serve divergence. That matters because those duplicate, drift-prone probes were the root cause of waits hanging on suspended rigs and of the drain reserving one copy of a member while serving another; this is the terminal slice of the store-residency ("split city") migration, and it retires four rows from the residency-boundary allowlist rather than growing it.

### Original PR Review
Decision: approve.

Specific gaps this PR closes (each validated correct against source by review):
- **`gc wait` sleeping forever on a suspended rig** — the core liveness fix. The new `errWaitDependencyUnproven` sentinel is kept distinct from `beads.ErrNotFound`, with three-way handling (unproven → retain the wait, `ErrNotFound` → fail only that wait, other fatal → abort the pass). A suspended rig can no longer convert "unproven" into a false "absent."
- **Drain reserve/serve divergence** — reserve (`drainMemberOwningStore`) and serve (`loadDrainManifestMembers`) now share `resolveDrainMember` (ByID, binding-first), so both resolve each member to the same store by construction instead of by two drift-prone spellings.
- **A relocated dependency sleeping its waiter forever on the one-shot plane** — the claim-route probe `holds()` collapses onto `byIDBindingOwnerForTopology`, preserving the full pre-seam truth table (binding-owns → true, residual → false, work-shaped standing refusal → tolerated false, reserved-namespace refusal → surfaced error); order pin `TestClassRoutedStoreForIDKeepsThePreSeamCandidateOrder` guards the leg order.
- **Reserved-prefix knowledge scattered across packages** — `ClassName*` and `reservedClassPrefixes` move to a stdlib-only leaf `internal/beadmeta`, with `config`/`coordclass` as thin re-exports and imports pointing downward (correct layering); no prefix strings changed.
- **Relic-census rescanning every boot** — `.gc/storage-relic-census.json` now memoizes only TRUE with a monotone union write-back, so a missing/stale/corrupt entry costs a scan, never a wrong skip (fail-safe).

Both independent reviewers converged on approve using different methods — one traced each fix through source and cleared three suspected defects, the other executed the targeted relic-memo / wait-wake / drain tests and produced a path:line evidence bundle — and they corroborate each other. `scripts/check-residency-boundary.sh` exits 0.

- **ga-e58fu: multi-member-store drain resolution is not exercised by a test** (Testing & Coverage, minor). `resolveDrainMember` is now the single reserve/serve resolution point, and a convoy spanning more than one member store is the exact case the unification is meant to protect — but no test drives that case yet. It is non-gating only because the current runtime never constructs such a convoy, so the untested path is presently unreachable. Filed as a follow-up to add a drain test that builds a convoy whose members resolve to two distinct stores and asserts reserve and serve select the same store per member. Not addressed in this PR.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `e634b03` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/32921249774

### Review Iterations
1 review pass performed, 0 fix iterations — the change was approved on the first pass by two independent reviewers, so no maintainer fixups were needed.

---

julianknutsen:
Merging on the pipeline's own recorded decision: adopt-pr review completed and workflow finalize stamped status/merge-ready, but the run's bead lifecycle closed before the bead-driven merge patrol could act (verified: no bead carries pr_review.workflow_status=merge_ready). mergeStateStatus=CLEAN at merge time. — operator, on behalf of pr-merge-ready-patrol

Issue #5648, opened 2026-08-26
waiver expiry message should name the renewal PR so a stale base reads as rebase, not defect


sjarmak:
Split out of #5195, which asks for two things. The runway warning covers the first. This is the second.

## The problem

When a runtime-provider waiver expires and someone lands a renewal, a contributor whose base predates that renewal sees the ledger fail with a message naming a date that has passed:

```
internal/runtime/acp.NewSeamBacked waiver owned by ga-uz5t3a expired 2026-09-14
```

Nothing in that message distinguishes "this waiver genuinely lapsed and needs attention" from "your base is old, rebase and it goes away." The second case is by far the more common one, and the message sends the contributor bisecting for a defect that does not exist.

## What would fix it

The expiry failure should name the PR that renewed the waiver, so the message itself tells a contributor on a stale base what to do. That requires somewhere to record the renewal PR number alongside the waiver, which `waivedRuntime` has no field for today.

Worth deciding as part of that: whether the renewal reference belongs on the waiver entry itself, or in a small separate ledger of renewals keyed by constructor, so a waiver renewed several times keeps its history rather than only its most recent pointer.

## Why it is separate from #5195

#5195's core complaint is eight waivers lapsing at one instant, and the fix for that is the spread cap plus the advance warning. The runway warning makes the stale-base scenario rarer, because a waiver is far less likely to reach its expiry unrenewed. It does not make it impossible, and it does nothing for the message a contributor actually sees when it happens.

Referenced from the #5195 fix so this does not evaporate with that PR's body.

PR #5649, opened 2026-08-26
test(providerledger): stagger waiver expiries and warn before the cliff

### LINKED ISSUE #5195: Runtime provider contract waivers share one expiry; it passed and CI is red on every branch
The waived runtime providers in `internal/testutil/providerledger` share a
single hardcoded expiry, so they all lapsed at the same instant and
`TestCatalogMatchesProductionWiringAndDocumentation` now fails on every branch.

`internal/testutil/providerledger/ledger.go:330`, inside `waivedRuntime`:

```go
Expires: time.Date(2026, time.August, 12, 0, 0, 0, 0, time.UTC),
```

Every `waivedRuntime` call takes that date, so acp, exec, herdr, hybrid, k8s,
ssh, subprocess, t3bridge and tmux all report expired together:

```
entry "runtime.builtin.acp" constructor internal/runtime/acp.NewSeamBacked
  contract runtime.Provider waiver owned by ga-80po0c.3 expired 2026-08-12
... (nine entries)
```

## Impact

Main's last green CI run was `e2fd3c46d` at 2026-08-11T23:36Z. Every run
started after 2026-08-12T00:00Z fails, so this blocks merges repo-wide rather
than on one branch. It also fails the local pre-push hook, since
`make test-fast-parallel` covers this package, so it blocks pushes as well as
merges. Reproduced on a clean checkout of `e2fd3c46d` with no local changes.

## The deadline is working, which is the argument against moving it

The ledger landed in #4212 on 2026-07-13 with 15 waived constructors. #4260,
#4326, #4403, #4404 and #4407 converted five of them to proofs within a week.
Ten remain. That is a forcing function producing what it was built to produce,
so the expiry firing is the mechanism working rather than a malfunction, and
the date is not something a passing branch should move to get itself green.

The waivers are owned by `ga-80po0c.3`, so the call belongs to that owner and
to @julianknutsen. Landing the remaining proofs is the outcome the deadline
was set up to reach. Moving the date buys time at the cost of the pressure
that produced those five conversions, and `maxWaiverHorizon` caps any new date
at 90 days out.

If a date change is what you want, a one-line change plus the regenerated
`TESTING.md` block sits at `sjarmak/gascity@fix/runtime-provider-waiver-expiry`
(one month, to 2026-09-12, with the original date recorded in a comment). It
is deliberately not open as a pull request. Take the commit or ask me to open
it; extending somebody else's deadline is not a call I should make.

## Separately worth fixing

One shared date means the whole set reaches the cliff simultaneously and takes
CI down with it, at an hour nobody chose. Per-waiver dates would spread that
out, and a check that fails a week ahead of the real expiry would turn a
repo-wide outage into a scheduled reminder.

sjarmak:
Closes #5195.

All eight runtime-provider waivers expired on 2026-09-07, so a single missed renewal would have failed the whole repository's test baseline at one instant. #5610 gave each waiver its own date literal, but all eight literals still held the same date, so the independence was structural rather than actual.

## Three tiers, and only one of them can be tripped by the calendar

The design question here is which failures are allowed to be time-dependent. Getting that wrong turns the guard into the outage it exists to prevent, so the tiers are explicit:

**Expired waiver, hard failure.** Already present in `validateClaim` and unchanged by this PR. A lapsed waiver means an unproved production constructor is shipping unaccounted for, which is worth blocking on.

**Too many waivers on one date, hard failure.** New: `maxWaiversPerExpiryDate = 2`, enforced by `validateWaiverExpirySpread` inside `Validate`. This is what makes per-waiver dates mean something instead of being copies of one date. It is correctly blocking because only a developer's own edit can violate it, never the passage of time.

**Approaching expiry, advisory only.** New: `WaiverRunway(entries, now) []string`, returning warnings 14 days ahead. The catalog test logs each one under a `WAIVER RUNWAY WARNING:` marker and does not fail. It escalates to a failure only under `GC_WAIVER_RUNWAY_STRICT=1`, following the existing `GC_REQUIRE_OAPI_CODEGEN=1` precedent in `internal/api/genclient`.

The runway warning gets a real consumer rather than being logged into the void: a new `waiver-runway` job in `.github/workflows/nightly.yml` runs the catalog test with strict mode on. That workflow's existing `bundled-pack-pins` job already exists for exactly this shape of problem, drift that no PR triggers, so calendar drift sits naturally beside it. Fatal in the nightly, silent everywhere else.

`TestValidateStillFailsExpiredWaiver` pins the hard tier so a later refactor cannot quietly soften it. `TestProductionCatalogHasRunwayAtEveryStaggerStep` asserts the live property at the day after each stagger step: an advance warning is produced and the baseline stays green.

## One decision that is not mine to make

Staggering the dates unavoidably means changing them, and seven of the eight moved later (2026-09-07 to dates between 09-14 and 10-26, paired two per date, grouped so the waivers closest to landing sit earliest). The issue argues that extending someone else's deadline is not the reporter's call, and I agree, so I am flagging it rather than letting it merge quietly. @ga-uz5t3a owns these. If a different distribution is wanted, only the eight literals need to change; the guards do not depend on the specific dates.

The mechanism could not land without some redistribution: with all eight on 09-07, thirteen days out, a 14-day runway check would have gone red the moment it merged.

The window is 14 days rather than the week the issue names, so the reminder arrives with enough runway to land a contract rather than forcing a renewal. One line to change if you prefer the literal reading.

## Not included

The second comment on #5195 also asks that the expiry message name the renewal PR, so a contributor on a stale base learns to rebase instead of bisecting. The runway warning makes that scenario rarer but does not address it, and it needs somewhere to record the renewal PR number that does not exist yet. Filed separately as #5648.

## Verification

`gofmt` clean, `go build ./...` clean, `go vet ./...` clean, `go test ./internal/testutil/providerledger/... ./internal/testpolicy/... -count=1` ok.

At 2026-08-31, the date the pre-rework version of this change would have gone red, `Validate` passes and two warnings are emitted:

```
WAIVER RUNWAY WARNING: internal/runtime/acp.NewSeamBacked waiver owned by ga-uz5t3a expires 2026-09-14 (13 days remaining): renew the waiver or land the contract
WAIVER RUNWAY WARNING: internal/runtime/tmux.NewSeamBackedWithConfig waiver owned by ga-uz5t3a expires 2026-09-14 (13 days remaining): renew the waiver or land the contract
```

Strict mode passes today, so the nightly job goes green now and first fires on 2026-08-31.

`release-gates/ga-rt0f61-runtime-provider-waiver-expiry-gate.md` line 38 still described all eight constructors as expiring 2026-09-07; it now names the four dates with their constructor pairs and records which tier enforces what.


sjarmak:
Closing this in favor of #5652, which solves the same problem and solves more of it.

We landed on the same diagnosis independently: a waiver expiry passing is enough on its own to turn the fleet red with no code change involved, and the fix is to split what a code change can break from what the calendar can break. #5652 goes further everywhere the two overlap.

Where it is strictly better:

- **Both ledgers, one policy.** This PR only touched `providerledger`. `resourcecensus` has the same cliff, 37 rows all sitting on 2026-10-01, and #5652's `waiverclock` covers both. It also found that the two ledgers disagreed by a day on what "expired" means, `!Expires.After(now)` against `expiry.Before(day(now))`, which I did not notice.
- **A stronger staggering rule.** I spread the eight dates and added a structural check refusing more than two waivers on any one date. #5652's rule is same constructor means same date, different constructors mean different dates, which forbids accidental convergence *and* forbids splitting one contract across two dates. Mine allowed both of those.
- **The warning actually reaches the owner.** My advisory tier emitted a warning consumed by a nightly job, which is a report, not a delivery. #5652 files a deduped bead per owner and, more to the point, found why the delivery was broken in the first place: the eight waivers named `ga-uz5t3a`, which does not resolve. Every blocked engineer for two weeks got an expiry message pointing at nothing. That mechanism explains the 0-for-3 record and I missed it entirely.
- **A fleet-fatal backstop.** Grace as the default with a bounded 28-day window and `FleetFatalDay` printed in every message is a better shape than my `GC_WAIVER_RUNWAY_STRICT=1` opt-in, which nothing would have set.

Nothing in this branch survives that #5652 does not already cover, so there is no delta worth cherry-picking. #5648, the follow-up I filed here for renewing the waivers before they lapse, is largely answered by #5652's owner lane; leaving it open until that lands so the remaining gap is visible, since #5652 names its own deferred half (`ga-of05u`, the cron order that carries findings to owners) as still outstanding.

PR #5651, opened 2026-08-26
test(eventstest): use a hang budget for rotation waits

quad341:
## What this changes

The event-provider conformance suite now gives positive-delivery rotation checks a true hang-detector budget instead of treating the much smaller goroutine-race floor as a deadline. On contended hosts, a watcher that is making progress gets enough time to deliver its event; a genuinely wedged watcher still fails with a bounded timeout.

This is test-infrastructure only. Passing runs still return as soon as the event arrives, and no production event path or negative timing assertion changes.

## Review notes

- The new budget is file-local and mirrors the established six-times-race-floor convention used by the command suite.
- Only the two positive-delivery waits in rotation conformance use it.
- Timeout inputs, negative assertion windows, and shared test utilities are unchanged.

## Test plan

- [x] Run `make test-local-full-parallel` with the rootless Podman environment; all eventstest conformance paths pass.
- [x] Run `make test-ci-policy`, `go vet ./...`, fresh-cache `make lint-new`, and `make fmt-check`.
- [x] Audit the release gate: [`release-gates/ga-u4r4s0-eventstest-hang-budget-gate.md`](release-gates/ga-u4r4s0-eventstest-hang-budget-gate.md)

### DIFF
--- a/internal/events/eventstest/conformance.go
+++ b/internal/events/eventstest/conformance.go
@@ -16,6 +16,17 @@ import (
 	"github.com/gastownhall/gascity/internal/testutil"
 )
 
+// hangBudget is the wall-clock ceiling for hang-detector waits in this file.
+// It is a HANG DETECTOR, not a latency assertion: nextWithin returns the
+// instant the watcher yields an event, so raising this value would not slow
+// a passing run and lowering it would not make the suite stricter — it only
+// changes how long a genuinely wedged watcher takes to report.
+//
+// testutil.GoroutineRaceTimeout is a FLOOR, not a target (TESTING.md "Floors,
+// ceilings, and inputs"). Mirrors the reference implementation in
+// cmd/gc/hangbudget_test.go (hangBudget = 6 * testutil.GoroutineRaceTimeout).
+const hangBudget = 6 * testutil.GoroutineRaceTimeout
+
 // rotatableProvider is the small interface a Provider must satisfy
 // for the rotation subtest. Providers that don't expose rotation
 // (e.g. exec, fake) skip the test rather than failing the suite.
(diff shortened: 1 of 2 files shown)


sjarmak (approved):
Correct, and the comment that carries most of the change's weight is accurate rather than aspirational. Four things checked against the source.

The central claim holds. `nextWithin` (`internal/events/eventstest/conformance.go:764-776`) races a buffered-channel send from the `w.Next()` goroutine against `time.After(d)`, so a passing wait takes the channel case the instant the watcher yields. Raising the ceiling cannot slow a passing run. Confirmed by execution as well as by reading: `TestFileRecorderConformance/RotationPreservesInvariants` passes in 0.00s on this head.

The cited reference is real and matches exactly. `cmd/gc/hangbudget_test.go:45` declares `const hangBudget = 6 * testutil.GoroutineRaceTimeout`, the same expression, and `TESTING.md:1436` is the "Floors, ceilings, and inputs" section the comment names. It says what the comment says it says: floors are minimums, hang budgets belong to pure hang detectors, and negative assertions and timeout inputs must not be widened.

The two converted sites are the right ones and the only ones. No bare `testutil.GoroutineRaceTimeout` wait remains in the file. Neither converted site is a negative assertion or a latency bound: both guard positive delivery, and the semantic checks run after the wait returns. Widening them cannot let a test pass that was failing for a real reason, only change how long a wedged watcher takes to report.

The release gate's criterion 2 is the claim worth spot-checking, since a gate that overstates is worse than no gate. It says the two waits remain hang detectors and that no negative assertion, timeout input, shared floor, or production path changed. That is what the diff does.

Codex (gpt-5.4) reviewed the same head against these five questions independently and returned no substantiated issues, reaching the same conclusion on each with its own citations, including the `conformance.go:667` positive-delivery wait that was never a `GoroutineRaceTimeout` site and correctly stayed as it was.

PR #5652, opened 2026-08-26
fix(testpolicy): make a lapsed waiver fail its owner, not the fleet

julianknutsen:
## The bug

A waiver expiry date passing is enough, on its own, to turn every Go-touching
push in the fleet red. No code change involved.

Two checked ledgers carry dated waivers — `internal/testutil/providerledger` and
`internal/testpolicy/resourcecensus`. Neither has a build tag, both land in the
unit-core job, and that job runs in `.githooks/pre-push` (line 55, `exec make
test-fast-parallel`). Nothing in `cmd/gc` imports either one; `go list -deps
./cmd/gc | grep -c providerledger` returns 0. They red the fleet anyway.

This fired on **2026-08-12** and again on **2026-08-26**. Both times it was
cleared by whoever happened to be blocked rather than by the waiver's owner. The
record is 0-for-3 at reaching the owner and 3-for-3 at blocking bystanders.

The next one was scheduled: eight runtime waivers all sat on **2026-09-07**, and
all 37 census rows sit on **2026-10-01**. Measured on `origin/main`, the census
cliff is 74 problems, not 37 — the count in the bead was rows, and each row
yields more than one problem line.

Filed as `ga-b51te`.

## What this does not do

It does not weaken the ratchet. The point is that an expiry should fail the
**owner**, not every bystander.

## Design

`internal/testpolicy/waiverclock` is one policy both ledgers consume. It splits
findings by how they can come to exist:

- **Structural** — missing owner, malformed or absent date, an expiry parked
  past the horizon ceiling. These can only appear with a code change, so they
  stay fatal in every mode and belong to whoever wrote them.
- **Lapse** — a date passed while the code sat still. This is the calendar
  acting alone, so it warns for a bounded window naming its owner, then goes
  fatal everywhere.

Grace is the default and is what an unset environment gets, so a lane that
scrubs its environment fails safe rather than silently strict. `FromEnv` rejects
an unrecognized value instead of falling back — a gate that quietly degrades on
a typo is not a gate.

The runway is 14 days of warning before the expiry plus 14 after: 28 days, then
fatal in every mode including grace. `FleetFatalDay` is the unignorable backstop
and it is printed in every message.

Both ledgers previously disagreed on what *expired* meant — `providerledger`
used `!Expires.After(now)` so a waiver dated 2026-09-07 was already dead on
2026-09-07, while the census used `expiry.Before(day(now))` so a row dated
2026-10-01 survived all of 2026-10-01. Same intent, one day apart. `Expired`
now compares whole UTC days for both, which is what the rendered `YYYY-MM-DD`
tables in TESTING.md already claimed.

## Commits

1. **`waiverclock`, one clock policy.** Stdlib-only, no upward dependency.
2. **`providerledger` routed onto it.**
3. **`resourcecensus` routed onto it, and stopped double-reporting.** The lapse
   decision was *removed* from `validateOwnershipFields` and hoisted into one
   `collectExpiries` walk, which deleted the `now` parameter from 8 functions.
   Removing a parameter beat adding one.
4. **Re-owned the eight runtime waivers.** They pointed at `ga-uz5t3a`, which
   does not resolve — `bd show ga-uz5t3a` fails identically to `bd show
   ga-zzzzzzz`, verified with that control. The real owner is `ga-80po0c.3`, an
   open epic child. Introduced by `bd84f0172a` / #5610, the PR that fixed the
   08-26 outage. So every blocked engineer for the last two weeks got the string
   *"waiver owned by ga-uz5t3a expired"* pointing at nothing. The 0-for-3 record
   has a mechanism. Now pinned by a literal in one test — one pin, N drift
   checks — so a constant rename cannot silently re-own all eight. (`ga-lgizl`)
5. **Staggered the eight expiries, keyed by gap rather than by row.** The two
   `t3bridge.NewSeamBacked` rows are one composition reached by two routes, and
   one contract closes both, so they deliberately share a date. The rule is
   *same constructor ⇒ same date, different constructors ⇒ different dates*,
   which is strictly stronger than "no two share a date": it forbids accidental
   convergence and forbids splitting one gap across two dates. Measured against
   the real catalog: **on 2026-09-07, the day that would have taken the fleet
   red, fatal=0 in both modes.** First grace-mode fatal is a single waiver on
   2026-10-08. A staircase, not a cliff. (`ga-x7hxo`)
6. **The owner lane.** `scripts/waiver-clock-audit` runs both ledgers strict, so
   an expiry fails there on the day it happens. Nightly `waiver-clock` job runs
   it bare; the maintainer-city audit order runs it `--file-beads`, filing one
   deduped bead per distinct waiver owner — for warnings too, which is what
   makes the 14-day warn-ahead window real. Detection only: it never re-dates a
   waiver, never edits a ledger, never closes a bead. TESTING.md gets the
   governance the ledgers never had.

Each slice was TDD — test written first, watched fail, then made to pass.

## Verification

**Ownership.** Ran the exact mutation the review council used to expose the
hole (`runtimeContractWaiverOwner` → `ga-DEADBEEF` plus the eight matching
TESTING.md rows). Before this PR it left the package green. Now it fails both
`TestRuntimeWaiverOwnerIsPinnedAndWellFormed` and `TestEveryWaiverNamesAnOwner`.
Restore verified clean.

**The audit script, probed with a control, three paths, all restored clean:**

| Injected state | Expected | Observed |
|---|---|---|
| Warn window only, nothing lapsed | findings printed, exit 0 | exit 0 |
| Lapsed | findings printed, exit 1 | exit 1 |
| Structural drift, no clock problem | non-clock branch, exit 1 | non-clock branch, exit 1 |

The third case matters: a structural break must not be swallowed behind a clean
clock report.

**The `--file-beads` dedupe query**, run against real `bd` with a throwaway bead
and two controls: matching owner returns 1, wrong owner returns 0, wrong label
returns 0, and closing the bead returns it to 0. An always-zero dedupe would
file a fresh alert every night, so proving the match branch is reachable
matters as much as proving the miss branch is. Probe bead deleted.

`-v` on the audit's `go test` is load-bearing, not cosmetic — warn-window
findings arrive as `t.Logf` lines that `go test` hides on a pass. Without them
the audit could only ever see a lapse that is already fatal, which is 14 days
too late to be useful.

**Gates:** `go vet ./...`, lint 0 issues, `make check-docs`, `make
test-ci-policy`, and the full `make test-fast-parallel` push gate.

## Why one PR and not six

The slices are sequentially dependent and share files: slice 4's owner constant
feeds the TESTING.md block slice 5 regenerates, and slice 6's audit script is
meaningless without slice 1's modes. There is no disjoint file split here. The
commits are individually reviewable and each is green on its own.

## Deliberately left open

- **`ga-of05u`** — the maintainer-city cron order that runs
  `scripts/waiver-clock-audit --file-beads`. That is the *reach* half, and it is
  a pack deployment artifact, so it does not belong in this repo. **Until it
  deploys this slice is half-wired:** the clock is correct and bounded, and the
  nightly reports, but nothing actively carries a finding to the owner except
  the `FleetFatalDay` backstop 28 days later. Saying so rather than implying the
  loop is closed.
- **`ga-rq7km`** — the census still has no equivalent of the provider ledger's
  `DispositionNotApplicable` (mandatory reason, no clock), so a standing fact
  has to carry a fake migration date. Root fix is a `standing_reason` schema
  field plus a census date stagger, and it wants its own PR. Safe to defer:
  census rows sit on 2026-10-01 (fatal 10-02) and grace keeps the fleet green to
  roughly 10-15.
- The release gate `release-gates/ga-rt0f61-...md` recorded the **inverse** of
  the ownership fact. Its PASS is kept and a post-gate amendment section carries
  the correction with the `bd show` control, following the
  `ga-7vhfyj-cwd-fallback-guard-gate.md` precedent.

Refs: `ga-b51te`, `ga-lgizl`, `ga-x7hxo`, `ga-of05u`, `ga-rq7km`


sjarmak (approved):
I had a smaller version of this open as #5649 and closed it in favor of this one. That gives me a useful position to review from: I had already worked the `providerledger` half and can say concretely where this is better, and I checked the central claim rather than taking it.

**The 09-07 cliff is gone, verified against the real catalog.** I ran `Validate(Catalog(), ...)` across ten dates in both modes:

```
2026-09-06  grace fatal=0 warn=1   strict fatal=0 warn=1
2026-09-07  grace fatal=0 warn=1   strict fatal=0 warn=1
2026-09-08  grace fatal=0 warn=1   strict fatal=0 warn=1
2026-09-14  grace fatal=0 warn=2   strict fatal=0 warn=2
2026-09-21  grace fatal=0 warn=2   strict fatal=1 warn=1
2026-09-28  grace fatal=0 warn=3   strict fatal=1 warn=1
2026-10-08  grace fatal=1 warn=3   strict fatal=1 warn=2
```

Fatal is zero in both modes on the day that would have taken the fleet red, and the first grace-mode fatal is a single waiver on 2026-10-08. That matches the claim exactly, and the shape between those dates is the staircase the PR says it is: strict starts failing the owner on 09-21 while grace still only warns, which is the whole point of the split.

**Grace is the zero value, not just the documented default.** `ModeGrace Mode = iota`, so a caller that constructs a mode and forgets to set it gets the safe one. That is the right way round, and it is doing real work here: it means a new call site cannot accidentally be strict. `FromEnv` returning an error on an unrecognized value rather than falling back is the same instinct applied to the other end.

**Nothing on the push path can reach strict.** `grep -rn GC_WAIVER_CLOCK` across the workflows, scripts, Makefile, and Go sources finds it in exactly three places: the two package tests that read it, and `scripts/waiver-clock-audit:79`, which sets it with `env GC_WAIVER_CLOCK=strict go test` on the child process rather than exporting it. The comment above that line says why, and the scoping matches the comment. This is the property the whole design rests on, and it holds.

**The owner constant was the actual mechanism.** Slice 4 is the part I would have missed entirely. I had diagnosed the cliff and built a warning tier for it, but a warning naming `ga-uz5t3a` would have been just as undeliverable as the failure was, and I never checked that the bead resolved. Pinning it with a literal so a rename cannot silently re-own all eight is the right shape for that.

Two things I checked and found fine, recording them so nobody re-checks:

- `Validate` reports every fatal, not the first. `joinProblems(append(problems, report.Fatal...))` at `ledger.go:516` joins the whole set, so the owner lane sees all lapsed waivers in one pass rather than one per run.
- `waiverclock.FromEnv` has no non-test caller, which normally reads as a computed-but-unconsumed signal. It is not one here: this is a test-policy mechanism, so the tests are the intended consumers and the audit script is what drives them.

On the deferred half: the body says plainly that until `ga-of05u` deploys, nothing actively carries a finding to the owner except the `FleetFatalDay` backstop 28 days later. That is filed rather than left in prose, so it is not blocking. It does mean the 14-day warn-ahead window is currently a report rather than a delivery, which is worth remembering when the next expiry approaches.

Approving.

sjarmak:
Recording the cross-provider pass behind the approval: Codex (gpt-5.4) reviewed this head against five questions and returned no issues.

What it verified independently:

- The grace/strict split is genuinely isolated to the audit lane. `GC_WAIVER_CLOCK` is the only selector (`waiverclock.FromEnv`), and the only automated strict caller in the repo is `scripts/waiver-clock-audit:72-80`, which sets it on the child process only. Neither `.githooks/pre-push` nor `ci.yml` sets it; nightly runs the script as its own job rather than exporting a shared env var. `go test ./scripts/cipolicy/...` passes, so the workflow-policy guard accepts that separation.
- `ModeGrace` is the zero value (`waiverclock.go:37-48`), unset/empty resolves to grace, and an invalid value errors rather than silently selecting strict. No config, struct-copy, or defaulting path reaches strict by accident.
- The change does not soften real defects. `validateClaim` keeps missing owner/reason/expiry (`ledger.go:583-592`) and over-horizon waivers (`:593-597`) fatal in every mode; only dated-expiry items reach `waiverclock.Check`. Three tests pin exactly that (`waiver_clock_test.go:31-64`, `:66-79`, `:82-97`).
- On the framing: as of today no `ga-80po0c.3` row is expired on this branch, and the package passes under strict as well as grace. The provider-ledger red that motivated the work is fixed by the independent-expiry change, not by this downgrade, and this change stands on its own as the mechanism for the audit lane.

PR #4175, opened 2026-07-12
Restore store-scoped control dispatcher routing

julianknutsen:
## Summary

- route formulas v2 control beads to the dispatcher in the graph owning city or rig store
- preserve that owner scope through initial decoration, retries, fanout, dynamic fragments, manual orders, and scheduled orders
- repair persisted #3463/#3765 cross-scope routes from canonical `gc.root_store_ref`, with bounded writes and no guessed ownership
- make the targeted dispatcher reconcile path start dynamically materialized per-rig dispatcher sessions
- retain the valid Dolt circuit-breaker transient handling from #4149

## Root cause

PR #3765 treated `max_active_sessions = 1` as one dispatcher across the fleet and preferred the city dispatcher for every graph. The core pack actually expands its unscoped dispatcher into separately qualified city and per-rig configs, each capped at one session. Graph placement never changed: rig graphs and their control beads remained in the rig store, so the city route had no consumer that could read them.

A second startup hole made the symptom confusing: `controlDispatcherTick` could materialize a rig dispatcher pool session but filtered it out before reconciliation because it only retained statically named sessions. The new session could remain `start-pending` on the immediate wake path.

## Behavior

The graph store is authoritative for control routing:

- city-owned graph -> `core.control-dispatcher`
- `fixture`-owned graph -> `fixture/core.control-dispatcher`
- owner rig A with workers on rig B -> controls stay on A while worker routes stay on B

Missing scoped dispatcher config fails before graph instantiation. Runtime health never changes route identity; desired-state reconciliation recovers the dispatcher for that scope.

This supersedes the route-alias workaround in #4149: rig dispatchers do not claim city routes and demand does not reinterpret dishonest metadata using the physical store. It keeps the independent breaker-outage liveness fix. The proposed pressure-skip exception is not needed: dispatcher signals already bypass pressure shedding, and the controller forces a full tick after bounded skips.

## Verification

- `make test-fast-parallel`
- `make test-cmd-gc-process-parallel`
- `go vet ./...`
- `make check-docs`
- `make dashboard-check`
- pre-commit changed-package lint, generated-doc freshness, vet, and docs checks
- pre-push sharded fast suite
- independent five-axis review: approve, no blocking findings

Focused regressions cover city/rig route symmetry, aliased legacy file stores, conservative migration, missing dispatcher failure, cross-rig execution, materialize -> serve selection -> real `dispatch.ProcessControl` advancement, runtime-missing dispatcher recovery, and Dolt breaker survival.

PR #5657, opened 2026-08-26
fix: resolve gc role aliases in formula routing

### LINKED ISSUE #5655: Formulas v2 cannot resolve installed gc.* role aliases
Cross-reference: gastownhall/gascity-packs#355\n\nWith the pinned `gascity/roles` import, `gc config show` materializes rig-scoped agents such as `gascity/run-operator` and `my-project/run-operator`. Formula assets intentionally route steps to semantic targets such as `gc.run-operator` (and `gc.task-decomposer`).\n\nHowever, formula compilation fails closed:\n\n```text\nstep superpowers-build.prepare: unknown formulas v2 target "gc.run-operator"\n```\n\nThe resolver should map `gc.<role>` aliases to the appropriate installed role in the workflow rig context, or provide an equivalent documented alias layer. This blocks live formula launches even though pack structural tests pass.

allenday:
Fixes #5655.

Formula packs may use semantic `gc.<role>` values in `gc.run_target`. The resolver first preserves an exact configured identity (including an explicitly bound `gc.<role>` agent), then falls back to the bare role in the workflow’s existing rig context.

This revision also removes unrelated propagation of a rig scope into the default execution route: unscoped workflow stores retain their city control dispatcher. The Formula v2 specification now documents both the alias and its exact-name precedence.

Verification:
- `go test ./internal/graphroute -count=1`
- `git diff --check`

No Compose services were started.


allenday:
Local focused verification passed: `go test ./internal/graphroute -run TestFormulaRoleAliasUsesBareRigRole -count=1`.\n\nThe PR evidence watchdog is waiting for a head-SHA CI check. This newly created fork has Actions workflows unavailable through GitHub’s API (`workflow_dispatch` is 404 despite repository Actions permissions being enabled), so a maintainer approval/enablement of fork workflows is needed before the watchdog can observe CI evidence. The PR remains draft pending that approval and hosted CI.

sjarmak:
Read this while it is still a draft, so treat it as early feedback rather than a verdict. Two things, one design and one blast radius.

**The `gc.` target namespace does not exist in this repo yet.** The body describes `gc.<role>` as a documented semantic target, and I could not find it documented. Searching the tree for a `gc.`-prefixed value used as a routing target turns up nothing outside this diff; `gc.*` is the reserved *metadata key* namespace, and `docs/reference/specs/formula-spec-v2.md:214` and `:575` define it that way. So this PR is not teaching the resolver about an existing alias layer, it is creating one, in Go, with no spec text. Two consequences worth deciding deliberately rather than inheriting:

- `formulaRoleTarget` makes `gc.foo` and `foo` the same target unconditionally, so a config that names an agent `gc.foo` becomes unaddressable. Cheap to bound: only strip the prefix when the bare name fails to resolve, or when the stripped name does resolve, rather than always.
- The same four characters would then mean "reserved runtime metadata key" on the key side of a step's metadata and "strip me, I am a role alias" on the value side. That is a spec change, and it belongs in `formula-spec-v2.md` in the same PR as the code, alongside the rule for what happens when both `gc.foo` and `foo` are configured.

If the intent is that the pack in gascity-packs#355 is the thing that should change, that is also a legitimate resolution of #5655 and a much smaller one.

**The `DecorateGraphWorkflowRecipe` change is a second, unmentioned behavior change with a wider reach than the alias fix.** Setting `defaultRoute.RigContext = scopeRef` for `scopeKind == "rig"` (`internal/graphroute/graphroute.go:512`) feeds `graphBindingRigContext` at `:532`, which becomes `executionRigContext` for every step binding in the recipe and gets stamped onto steps as the execution rig-context metadata key at `:241`. Before this, that value came only from the qualified form of `routedTo`, so a rig-scoped workflow whose root route is an unqualified name previously carried no execution rig context and now carries the scope ref.

That affects every rig-scoped graph.v2 workflow, not only the ones using a `gc.` target. It also reaches the control dispatcher: `controlRigContext` starts as `executionRigContext` at `:533` and is only overridden when `storeref.ScopeRigContext(rootStoreRef)` reports the store is scoped, so for an unscoped root store this can move dispatcher selection from the city singleton to a per-rig dispatcher. Neither new test covers that path, and neither is a rig-scoped workflow with an unqualified root route and a non-`gc.` target, which is the shape that changes.

Either split it out with its own test and its own justification, or, if it is load-bearing for the alias case, say in the body why the alias fix requires it. Right now it reads as an incidental second fix.

Nothing here needs the local Docker test run to resolve; both are questions about what the change should be.

allenday:
Addressed the early review feedback in 4386dceb0:

- Documented `gc.<role>` as a `gc.run_target` value alias, separate from reserved `gc.*` metadata keys. Exact configured identities resolve before bare-role fallback.
- Removed the broad `scopeKind == "rig"` default-route propagation. The new regression test keeps an unscoped workflow store on the city control dispatcher.
- Kept the alias fallback scoped to the rig context already derived by normal workflow routing; focused verification passed with `go test ./internal/graphroute -count=1`.

Marking ready for review; the hosted evidence checks are queued on this head SHA.

allenday:
Thanks — this was the right concern to raise. I traced the alternative through `gascity-packs#355` before choosing the core fix. That issue was closed as the cross-reference to #5655 after confirming that the packs intentionally publish `gc.<role>` targets while the installed role agents materialize as rig-scoped bare names. The Gas City and Superpowers pack docs, plus the shipped formula assets, use that semantic target surface; changing only the pack would turn this into a broad formula/docs contract migration and would make bare targets depend on each invocation path preserving rig context.

So the revised PR makes the semantic target layer explicit at the resolver boundary instead: it first resolves `gc.<role>` exactly, preserving a real configured/bound identity of that name, and only then falls back to `<role>` in the workflow rig context. The Formula v2 spec now documents that value-side rule separately from reserved `gc.*` metadata keys.

I also agree the original scope-context change was incidental to that fix, so I removed it. The added regression holds an unscoped workflow store on the city control dispatcher; the alias fallback continues to use the existing workflow/store-derived rig context.

If you think the pack contract should instead be changed to bare role targets, I’m happy to take that direction, but based on #355’s traced resolution and the current published pack surface, I believe this is the smaller compatible fix.

quad341 (approved):
@allenday — thank you for the revision, and specifically for how you did it: you traced `gascity-packs#355` to its resolution before defending the core fix instead of just asserting it, and you pulled the `scopeKind == "rig"` propagation back out once that was flagged as a second change riding along. The new control-dispatcher test is the good version of that guard — asserting the *absence* of the execution rig context, not just the route, is what will actually catch a reintroduction.

One thing in your favor that the earlier thread got wrong: `gc.run_target = "gc.run-operator"` is already published in `docs/guides/understanding-formulas.md` and in the ops engdocs. You're writing a shipped convention into the spec, not inventing a namespace. I'll say so on the thread.

I'll take the remaining work on our side — it's all test wiring. The shared `cookReproRecipe()` fixture now points every test in the file at the alias, which drops the bare-target guard from #4175; and the precedence test resolves through the fallback rather than the exact lookup (with `BindingName: "gc"` plus `Dir`, the agent's identity is `my-project/gc.run-operator`, which the unqualified lookup can't match), so it passes either way. I'll parameterize the fixture and rebuild that config so it fails if the ordering is inverted, then run the suites. Our fork-CI is why nothing ran on your head — not anything you did.

I also spotted that three other `gc.run_target` consumers don't know about the alias yet. That's ours to follow up on, not a change to this PR.


---

Issue #4739, opened 2026-07-28
bug: builtin claude provider "model" enum missing claude-opus-5; silently drops --model instead of erroring, and its context window under-reports at 200K


jacobhausler:
The builtin `claude` provider's `model` option (`internal/worker/builtin/profiles.go`) is a closed enum (`Choices []BuiltinOptionChoice`). A value outside the enum produces no `FlagArgs`, so the resolved launch command silently omits `--model` entirely and the session runs the CLI's own default model — while `gc config show` and `gc doctor` both keep reporting the configured pin as set, with no event or warning anywhere. Separately, the exact same value validated through `internal/config.ResolveOptions`/`ApplyOverrides` (e.g. named-session model overrides) hard-errors: `invalid value for model: claude-opus-5`. Same root cause, two very different, un-cross-referenced failure modes.

`claude-opus-5` (Anthropic's current Opus release) has no entry in this enum at all — `claude-sonnet-5` (#3867) and `claude-fable-5` (#3284) were added previously by the same pattern, but Opus 5 was missed, and none of the three added the *canonical* model-id spelling (`claude-sonnet-5` etc.) as a directly accepted value — only a short alias (`sonnet-5`). Operators who pin the full provider id (as documented / natural in `agent.toml`) hit the enum gap regardless of alias.

Separately, `internal/sessionlog.ModelContextWindow` keys only on the bare family word `opus` and returns the 200K default for any Opus id lacking an explicit `[1m]` suffix. Opus 5 ships 1M context natively (there is no 200K Opus 5 variant), and the Claude Code session-log transcript never carries the `[1m]` suffix for a model that is 1M by default — so a Claude-Opus-5-backed session's usage gauge is computed against a denominator 5x too small.

(Related: PR #4527, open, proposes a shared `internal/modelwindow` table for this same class of drift, but its `millionMarkers` list does not include `opus-5` either — flagging separately.)

## Repro

Both new test files in the companion PR fail on current main (`af42a94245a547a0c47ec26054afa5fd1347b567`) and pass with the fix. No `gc`/`bd` tooling required — stock `go test`:

```
go test ./internal/worker/builtin/ ./internal/sessionlog/ -run 'Opus5|CanonicalIDs'
```

Observed on main:

```
--- FAIL: TestBuiltinClaudeModelChoicesIncludeOpus5
    claude model choices missing "opus-5"
--- FAIL: TestBuiltinClaudeModelChoicesAcceptCanonicalIDsVerbatim
[...]

PR #5658, opened 2026-08-26
fix(config): make model and effort open options across all providers (ga-fyh)

brandonmartin:
Fixes the class of defect where a `model` or `effort` pin that gc cannot honor is **silently discarded** instead of erroring — the agent then runs the provider CLI's default, unpinned, with no warning.

## Why

`internal/worker/builtin/profiles.go` carried a hand-maintained `Choices` enum per provider. A value absent from that enum has no `FlagArgs`, so the launch path emits no `--model` at all, while the named-session resolution path hard-errors on the same value. The two paths disagree and the launch path fails open.

This was live in a running city: `gasburger.refinery` and `gasburger.gorkcats` pinned `model = "grok-4.6"`, which was not in the grok enum (`grok-build`, `grok-composer-2.5`, `grok-composer-2.5-fast`). Confirmed from process args:

    grok --permission-mode bypassPermissions --effort high

No `--model`. Both agents had been running the grok CLI default rather than the pinned model, silently.

The same defect class was fixed for Claude on 2026-08-03 in 41d35b765 (#4739/#4740) after incident ra-jbbv0 — that fix added canonical `claude-*` ids but did not generalize, so every other provider kept the hole. The grok enum had not been touched since cf69ca5d6 (2026-06-07), the commit that first added the provider.

## Scope — `effort` too, not just `model`

An audit of all 20 provider profiles found the effort axis equally broken:

- 13 providers have **no** effort option at all — every `effort =` pin silently dropped
- 7 (kiro, cursor, copilot, amp, zcode, auggie, omp) have **neither** model nor effort
- `effort = "max"` on **codex** is dropped (its enum stops at `xhigh`) even though claude's `OptionDefaults` ships `"max"` — the value looks blessed but isn't portable
- **antigravity** drops `xhigh` and `max`

## Design note carried in the change

Effort is not universally a separate flag. **cursor** and **antigravity** encode it *inside* the model id (`claude-opus-5-low`, `gemini-3.6-flash-high`), and cursor adds an orthogonal `-fast` axis. `cursor-agent --list-models` returns 198 models against zero in gc's cursor profile. So `(model, effort)` must resolve per-provider, sometimes composing into a single string — a registry that assumes a separate `--effort` flag everywhere would mis-handle both.

## Bead trail

- `ga-fyh` — this work
- `ga-cot` — cursor catalog catch-up, dep-blocked on this landing
- `ga-psh` — this PR

https://claude.ai/code/session_01QZac4AWCChWtPcjTb43P6C


brandonmartin:
Rebased the two-commit series onto current `upstream/main` (`9700d9a48`) and force-pushed `e23722834`. The CHANGELOG conflicts were resolved by retaining all upstream entries and the branch’s final expanded open-option entry (without duplicating its earlier short form). Current genschema also exposed a missing generated projection for `flag_template`, so I added a small third commit refreshing both pack-schema artifacts. Fresh config/builtin tests, focused CLI pin tests, genschema idempotence, `make check-docs`, focused vet, and diff checks pass. Ready for refreshed CI/review.

quad341 (approved):
@brandonmartin — the fix I want to call out isn't grok-4.6, it's the diagnosis underneath it: noticing that the launch path fails *open* while named-session resolution fails *closed* on the very same value, then auditing all 20 provider profiles and finding the effort axis was broken the same way. That's what turns a one-line enum addition into an actual fix. `TestOpenModelPinIsHonoredForEveryProvider` pins the invariant rather than the symptom, which is exactly what `41d35b765` didn't do when it patched Claude only — and why we're here again.

Your design note on cursor and antigravity encoding effort *inside* the model id saved us from a worse registry shape. That only shows up if you go and look.

Two things I'll take on my side so you don't have to round-trip: `CollectAllSchemaFlags` still only walks `Choices`, so a template-rendered open value survives `StripFlags` and a later override emits `--model` twice — I reproduced it on the plain, codex `-c`, and resume forms, and I'll add a shape-based strip pass plus regression tests. I'll also refresh the two Cursor rows in `harness-recipes.md` that your new cursor model option makes stale.

The one red check is a watchdog that gave up before `CI / required` finished; it went green afterward. Merging after the fix. Thank you for chasing this all the way down.

---

**Verification I ran** (read-only, in `./pr-checkout`; temp test files removed, tree clean): confirmed the duplicate-flag defect on three shapes, confirmed the resume path has it too, and **disconfirmed** Codex's separate claim that default-completion duplicates — `commandContainsOption` is shape-aware and handles it, so that step is not in the fix plan. No merge, push, or edit performed.


---

Issue #5453, opened 2026-08-21
bug: a worker pool at zero sessions can never restart — pool_desired and work_query both read only in_progress+assigned, so standing gc.routed_to demand is invisible (3 pools, all 3 lanes)


swedeinasia-flow:
## Summary

A worker pool that reaches **zero live sessions** can never restart itself, even with
correctly-routed open work standing in front of it. The pool stays at zero indefinitely; the
lane is silently dark while every health surface reads green.

The reason is that **nothing in the system reads `gc.routed_to` on an OPEN bead.** Both halves
of the loop are closed over work that is *already in progress*:

- **The scaler.** A template's `pool_desired` is the count of beads that are already
  `status=in_progress` **and** already assigned to one of that pool's aliases.
- **The worker.** Every pool's `work_query` opens with
  ```
  for id in "$GC_SESSION_ID" "$GC_SESSION_NAME" "$GC_ALIAS"; do
    r=$(bd list --status in_progress --assignee="$id" --json --limit=1)
  ```
  — only work already `in_progress` and already assigned to *this* session. `gc hook` is
  documented as "Finds routed work using the agent's work_query config", so a freshly spawned
  session with no trigger has, by construction, nothing to find.

The only component that can convert `open + routed` into `assigned + in_progress` is the
dispatcher pinning `GC_TRIGGER_BEAD_ID` at session spawn — and its pending queue is built from
`bd ready`, which **never contains molecule step beads**. A lane therefore only keeps moving as
long as its dispatcher happens to have *ready dispatcher-kind* beads (`workflow-finalize`,
`scope-check`, `ralph`) to process, because molecule progression rides on those. A lane with no
ready dispatcher-kind bead never reaches the pin, and a pool that has drained to zero has no
other way back.

Net: **anything that drives a pool to zero sessions — fleet recycle, city restart, controller
bounce — can strand it permanently.**

Engine `1.4.1-main.7c817e064`.

## Measurements

`pool_desired` sampled at one instant, deliberately chosen so session counts and desired counts
disagree (sessions / in_progress-assigned / pool_desired):

| pool | sessions | assigned | pool_desired |
|---|---|---|---|
| rig A `…/implementation-worker` | 4 | 3 | 3 |
| city worker | 6 | 3 | 3 |
| rig B `…/implementation-worker` | **0** | 0 | **0** |

Claim log over ~7h (`gc-claim-routed-work.log`), by lane:

```
work_rig=rig-A     282
work_rig=city       60
[...]

PR #5660, opened 2026-08-26
fix(controller): let pool demand see work resident in a relocated class binding

julianknutsen:
## The defect

`filterAssignedWorkBeadsForPoolDemand` decided reachability with a single equality test against the agent's configured **rig**:

```go
return storeRefs[index] == assignedWorkStoreRefForAgent(cityPath, cfg, agentCfg)
```

Once a city relocates its coordination classes, graph-resident work carries the binding's own `class:*` ref. No rig name can ever equal that, so **every in-progress claim on a molecule step is structurally invisible to pool demand** — not racy, not intermittent, unreachable by construction.

## Why it is a half-fix being completed

`filterAssignedWorkBeadsForSessionWake` got exactly this widening in ga-whzrt. The demand half never did. The two filters answer one question from opposite ends — *"is this holder still working?"* and *"does this work still need a worker?"* — so a ref the wake side accepts and the demand side rejects is a city that drains a slot and then refuses to refill it.

The stalled molecule never recovers because the resume tier is the only demand source once work is claimed: while the step is still `open`, `scale_check` supplies demand on an independent path and hides the hole entirely.

## Why it stayed latent

Observed on a live split city: graph beads have carried assignees for over a month at 1–9/day, but **none was ever assigned to a `-pool` session** until the assignee spelling moved from `rig/agent.name-N` to the `rig--agent__name-N-pool` session-name form. That spelling is what makes a bead resolve to a pool *template* and enter this arm at all. The relocation created the hole; the routing change started feeding it.

## The widening is narrower than the wake filter's

Deliberately. The wake filter matches a session's own exact assignee identity, so an unconditional claim-ref set still admits only that session's own claims. Pool demand matches a routed **template**, so the same set would let city-store work resume a rig-scoped pool session on a city that relocates nothing.

`assignedWorkRelocatedClaimRefs` answers empty for a single-store city (`Topology.IsSingleStore()`), so rig isolation and every unsplit city are byte-identical. `TestBuildDesiredState_RigPoolIgnoresAssignedWorkInUnreachableStore` is the control and passes unchanged.

## Tests

Four new rows in `cmd/gc/pool_demand_class_binding_test.go`, mirroring the wake filter's:

| Test | Asserts |
| --- | --- |
| `...KeepsABindingRefClaimOnTheReconcilerPlane` | production row: census-emitted `class:*` ref is kept |
| `...KeepsAClaimOnTheClassBindingArm` | mid-rollout row: leading-arm `""` spelling is kept |
| `...StillDropsWorkResidentInAnotherRigsStore` | **control** — another rig's ref is still dropped |
| `...UnchangedOnASingleStoreCity` | **control** — no widening without a relocation |

Two of the four fail without the fix (the `class:*` ref is dropped); all four pass with it.

## Verification

- `go test ./cmd/gc/ -count=1 -timeout 40m` → `ok 696.957s`
- `lint-changed ./cmd/gc` → 0 issues
- `go vet ./...` clean; docs/schema regeneration clean

Refs ga-pe2lf.

---
- Related to #5453 — touches the same pool_desired demand path: it widens the in-progress-claim residency filter so a claim recorded on a relocated class binding is no longer dropped, which lets a pool that drained while holding a claim be refilled; it does not make standing open gc.routed_to demand visible, so the zero-sessions-with-no-in-progress-work case described in that issue is unchanged
- Related to #5587 — touches the same class of defect that issue describes — a rig-identity equality gate that can never match a `class:*` ref after a class is relocated, so the row is silently dropped before demand derivation sees it — but on a different gate: this changes the pool-demand residency filter (`assignedWorkIndexReachableFromAgent` / `filterAssignedWorkBeadsForPoolDemand`) for in-progress claims, and leaves the control-row collection gate (`rootStoreRefMatchesCandidate` / `collectOpenUnassignedRoutedWork`) exactly as it was, so nothing here addresses that filed defect

### DIFF
--- a/cmd/gc/assigned_work_residency.go
+++ b/cmd/gc/assigned_work_residency.go
@@ -132,6 +132,34 @@ func assignedWorkClaimRefs(cityPath string, cfg *config.City, leading beads.Stor
 	return out
 }
 
+// assignedWorkRelocatedClaimRefs is assignedWorkClaimRefs for the callers that
+// may only widen where a class was actually RELOCATED: it answers empty for a
+// single-store city.
+//
+// The wake filter can take the unconditional set because it matches a session's
+// own exact assignee identity, so a wider ref set still admits only that
+// session's own claims. Pool demand matches a routed TEMPLATE, so the same set
+// would make city-store work resume a rig-scoped pool session on a city that
+// relocates nothing — the reachability rule
+// TestBuildDesiredState_RigPoolIgnoresAssignedWorkInUnreachableStore pins, and
+// the reason this is a second function rather than a second caller of the first.
+//
+// On a city that DOES relocate, the collapse inside ClaimRefs is what makes the
+// answer right on both planes: a leading arm that is itself the binding reports
+// the single ref its census records, and a distinct binding reports its own.
+func assignedWorkRelocatedClaimRefs(cityPath string, cfg *config.City, leading beads.Store) []string {
+	topology := residencyTopologyForCity(cityPath, cfg, censusWorkLeg(cityPath, leading), nil)
+	if topology.IsSingleStore() {
+		return nil
+	}
+	refs := topology.ClaimRefs()
+	out := make([]string, 0, len(refs))
+	for _, ref := range refs {
+		out = append(out, string(ref))
+	}
+	return out
+}
+
 // assignedWorkScanComplete turns a partial pass into an error for the gates that
 // answer "does this session still hold work?".
 //
--- a/cmd/gc/assigned_work_scope.go
+++ b/cmd/gc/assigned_work_scope.go
@@ -120,11 +120,58 @@ func assignedWorkIndexReachableFromAgent(cityPath string, cfg *config.City, agen
 	return storeRefs[index] == assignedWorkStoreRefForAgent(cityPath, cfg, agentCfg)
 }
 
+// assignedWorkIndexReachableFromAgentOnClaimRefs extends the rig-equality test
+// with the refs a claim can be RECORDED under whatever the holder's rig scope:
+// the leading work arm plus every relocated class binding (assignedWorkClaimRefs).
+//
+// This is the demand-side twin of the widening filterAssignedWorkBeadsForSessionWake
+// already applies (ga-whzrt). The two filters answer one question from opposite
+// ends — "is this holder still working?" and "does this work still need a
+// worker?" — so a ref the wake side accepts and the demand side rejects is a city
+// that drains a slot and then refuses to refill it.
+//
+// Relocating a coordination class is what made that reachable. The equality test
+// is against the agent's configured RIG, so once graph-resident work carries a
+// "class:*" ref no rig name can equal, in-progress demand for a rig-scoped pool
+// agent became structurally invisible. It stayed hidden because scale_check
+// supplies demand independently while the work is still ready; only once the
+// work is CLAIMED is the resume tier the sole remaining source, which is why the
+// symptom is a session that dies holding a claim and is never replaced.
+//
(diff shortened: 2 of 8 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277242989

quad341:
@julianknutsen — thank you for taking this all the way to the structural root. "No rig name can ever equal a `class:*` ref, so this is unreachable by construction, not racy" is the sentence that turns a maddening intermittent-looking stall into a five-line fix, and framing the wake and demand filters as one question asked from opposite ends is genuinely the right way to see it.

What I appreciate most is the restraint: you made this widening *narrower* than the wake filter's, gated it on `IsSingleStore()` so unsplit cities are untouched, and then wrote `TestPoolDemandFilterAcceptsTheWorkRefOnASplitCity` to say the intentional behavior change out loud instead of letting it ride in quietly. That's the test I'd have asked for, already written.

Two things on our side, neither yours. The red `Mac / quality` check is a repo bug I tracked down: `setup-gascity-macos` still pins Go 1.26.5 while `go.mod` requires 1.26.6 — its Ubuntu twin was already fixed to read the version from `go.mod`. I'm sending that as its own PR. And I'll add one `buildDesiredState`-level row so the split-city widening is pinned at the same altitude as the isolation control it has to keep honoring.

Merging with credit once those are green.


---

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR closes a controller demand-side reachability gap on split-store cities: when the infra class is relocated to a `class:*` binding, a rig-scoped session that dies while holding a **class-relocated** claim was invisible to demand, so its slot drained but was never refilled and the work stranded. Your change widens the demand reachability check to OR-in the relocated claim refs so the dead session gets replaced instead of silently losing fleet capacity — a real reliability fix for the exact split-topology the controller is meant to support.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- The reviewed pool-demand fix was correct, but the **named-session** demand path in `cmd/gc/build_desired_state.go` had the same blindness: a rig-scoped *named* session holding a class-relocated claim would also be stranded because the rig-equality gate makes `"class:*" == "<rig>"` false. Review iteration 1 asked to resolve that asymmetry. The final patch now routes the named-session demand check through the same widened `...OnClaimRefs` reachability test, so both demand planes see relocated claims, and pins the behavior with an end-to-end test through the real builder.

Review findings addressed:
- `c402bd5b33` widens the named-session demand match to include class-relocated claim refs (`cmd/gc/build_desired_state.go`) and adds `cmd/gc/named_session_demand_class_binding_test.go`, resolving the iteration-1 required change (named-session demand asymmetry). Bounding was verified: the widening returns `nil` on a single-store city (byte-identical to prior behavior) and never yields a rig ref, with single-store and foreign-rig drop controls on both planes.

- No non-blocking follow-ups.

### Maintainer Changes
`c402bd5b33 fix(controller): widen named-session demand to a class-relocated claim` — 2 files changed, +186/-1. Extends the relocated-claim reachability widening from the pool-demand path to the named-session demand path in `build_desired_state.go`, and adds an end-to-end `named_session_demand_class_binding_test.go` that drives the real builder. Original contributor commits (the pool-demand fix and its test) are preserved.

### Final Review Status
Ready for the merge queue: final head `c402bd5b` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/33076181227

### Review Iterations
2 review passes performed (1 fix iteration). Attempt 1 requested resolving the named-session demand asymmetry; attempt 2 approved after the maintainer fixup, with two independent reviewers converging on zero gate-impacting findings.

---

PR #5661, opened 2026-08-26
fix(git): ship SSH keepalive for long pre-push (ga-2i5)

brandonmartin:
## Summary

Pushes over SSH died with exit 141 (SIGPIPE) after a long pre-push hook: GitHub drops the idle transport while the hook runs, so git has nothing to write to when the hook finally passes. The working fix (`ssh -o ServerAliveInterval=20 -o ServerAliveCountMax=90` as `core.sshCommand`) lived in one operator's uncommitted `.git/config` and did not travel with polecat worktrees from other clones.

This ships the keepalive in clone-independent product config:

- Session start injects `GIT_SSH_COMMAND` (merging into an existing ssh command) so polecat `git push` keeps the transport alive.
- gitcred SSH injection adds the same flags so gc's own network git does not lose the connection.
- `gc doctor` warns when a rig has a long pre-push hook and no keepalive; `--fix` stamps `core.sshCommand` on the clone so every worktree inherits it. Custom non-ssh wrappers are left untouched.

Setting `GIT_SSH_COMMAND` inside the pre-push hook cannot save the parent SSH session — the transport is already open. Shortening the hook itself is out of scope here (already raised on ga-34a).

## Test plan

- [x] `go test ./internal/git/ ./internal/gitcred/ ./internal/doctor/ ./internal/session/`
- [x] `go test ./cmd/gc/ -run 'Doctor|GitCredential|PrepareStart|SSHKeepalive|BuildDoctor'`
- [x] Pre-push `make test-fast-parallel` passed on this branch (the 13-minute hook this change exists to survive)


quad341 (approved):
@brandonmartin — thank you for this one. Taking a fix that only existed in one operator's uncommitted `.git/config` and turning it into something that travels with every clone and worktree is exactly the right instinct, and the three-surface approach — session env, gitcred, doctor — is what makes it actually hold. The doctor check earns its keep: detecting the long hook, offering `--fix`, and deliberately refusing to stamp over a custom wrapper is careful work, and the test for that refusal is my favourite thing in the diff.

Two things I'll take on our side so you don't have to.

**The red CI is ours, not yours.** The failing test is `TestCmdStopSupervisorManagedInvalidCityTomlFailsWhenShutdownFails`, in the stop/registry path. We checked out your base commit with none of your changes and it fails there too, 3 for 3 — `main` is broken independently of this PR. We're filing that separately; it won't hold you up.

**One small guard.** Since `GIT_SSH_COMMAND` outranks `core.sshCommand` in git, the env injection can shadow the very custom wrapper your doctor check so carefully protects. We'll skip injection when an ambient wrapper is present and add the test. Your design already made the right call everywhere it could see the wrapper — this just extends it to the one place that can't.

Merging with credit once that's in.


---

PR #5625, opened 2026-08-25
fix(hook): pin continuation siblings to the session id, not the slot label

julianknutsen:
## Problem

`gc hook --claim` pins a molecule's continuation siblings so the next step lands
on the same session and keeps its context. It wrote the **runtime slot label**
into `assignee` (e.g. `gascity--gc__implementation-worker-5-pool`).

That label is an identity nothing downstream agrees on:

| consumer | identity it resolves |
| --- | --- |
| `sessionAssigneeMatches` (`compute_awake_set.go`) | session bead `ID` or `session_name` |
| continuation backstop | session bead ID |
| the session's own re-poll | `$GC_SESSION_ID` |

So a pinned bead became ready, matched **no** session row, produced no demand,
and the drained session was never woken. Live pre-deploy check: the slot-label
form matched 0 of 9,241 session rows.

## Fix

Pin to `$GC_SESSION_ID` (the session bead ID) when it is set, falling back to
the claim assignee when it is blank. That is the one identity all three
consumers already resolve.

## Test

`TestPreassignHookContinuationGroupPinsSiblingsToSessionID` — table-driven:
session ID wins over the slot label; blank session ID falls back.

## Verification

Deployed as part of `mc-enterprise6n-982fa047b5`. 32 minutes after the flip a
live `gascity/gc.run-operator-3` claim pinned its two continuation siblings to
`gcg-session-ac61e48f9771b19d8f60e67cd900d543`, which resolves to an open
session bead titled `gascity/gc.run-operator-3`. Pre-deploy the same claim wrote
`gascity--gc__run-operator-3-pool`, which resolved to nothing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
🔗 **Maintainer cross-reference** — added by the gascity maintainers, no action needed from you:
- Related to #2978 — corrects the identity written when pinning a molecule's continuation siblings so they name the session bead ID instead of a runtime slot label, which makes the session-affinity pin resolvable by the wake/demand path; it does not cover the broader step-to-slot affinity gap described in that issue, and the branch-already-used-by-worktree blocker is untouched

<sub>Linked for triage visibility — not auto-closing. If this looks off, just delete this block.</sub>

PR #5663, opened 2026-08-26
fix(hook): claim under the session bead id, not the reused pool slot label

julianknutsen:
## The defect

`gc hook --claim` records a pool worker's claim under the **slot label**, not the session it is actually running in.

`clearPoolTemplateRuntimeIdentity` gives an unaliased pool spawn a blank `GC_ALIAS` and stamps `GC_AGENT` with the pool's runtime session name. Since #5372 that name is slot-derived (`<identity>-pool`) and carries no bead ID. It is a **chair**: every session that ever sits in the slot spawns with the same string.

On maintainer-city, `gascity--gc__implementation-reviewer-1-pool` is the `session_name` of **24 distinct session beads** (23 closed, 1 open). A claim recorded under it answers "is the holder still alive?" about the chair, not the occupant — so a dead worker's `in_progress` bead looks held by whoever sits there next, and is never released, resumed, or replaced.

The environment already carried the occupant. `GC_SESSION_ID` is the session bead id, and `hookClaimIdentityPatch` already stamps it into `gc.session_id` on every non-control claim. Only the **writer** was on the chair.

**Live evidence:** 126 of 126 pool-assigned beads in the mc graph store already carry a non-empty `gc.session_id`. Not one exception — the occupant was recorded every time, in the wrong field.

## Why the bead ID and not a composite

A composite like `<slot>-<beadID>` is not an advertised identity. `sessionBeadAssigneeIdentities` (pool_session_name.go:29) enumerates `bead.ID`, `session_name`, `configured_named_identity`, `alias`, and alias history — nothing else. A composite would resolve to no session, the orphan sweep would read the holder as dead, and it would clear a **live** claim. The bare bead ID is the one form every reader already leads with (`currentSessionAssigneeIdentities`, `ComputeAwakeSet`, and `config.EffectiveWorkQuery`'s documented `$GC_SESSION_ID > $GC_SESSION_NAME > $GC_ALIAS` order).

`alias` stays first in the precedence so named agents are untouched — their stable alias is what keeps bd's actor-to-assignee close check passing (#4981 / ga-i44k).

## Root cause: a two-commit composition

Neither commit is wrong alone.

- `1534d46e24` (#5241, 08-13) routed the claim onto the pool session's runtime name — which at the time carried the bead ID via `PoolSessionName(template, beadID)`.
- `8c73625b97` (#5372, 08-19) changed what that name *is* to the bead-ID-free slot identity. Correct on its own terms: the runtime name is the sandbox/pod name, and the ID-bearing form leaked one pod per failed start (ga-vcjr9, desired=1 / 602 pods).

#5372 did not notice that #5241 had made the claim identity depend on that name carrying the ID. Both are present and unreverted.

**Onset was the deploy, not the merge.** Chair-shaped assignees appear in the mc graph store only from 08-25 (97 that day, 29 on 08-26) and never before, despite ~1500 beads created 08-19→08-24. The binary `mc-enterprise6m-017c715835` was built 08-24 20:28 — `git merge-base --is-ancestor 8c73625b97 017c715835` confirms it contains the commit — and started on mc 08-25 03:52. The 5-day gap between merge and symptom is build/deploy lag.

## Scope: what this does and does not fix

It does **not** resolve ga-jrnou (open P0). Orphan release tests `sb.Status != "closed"`, a bead-status test rather than a runtime-liveness one, so an orphaned seat whose session bead is still open is skipped under either assignee. The wedged adopt-pr step `gcg-5436965277237915` has exactly that shape: `in_progress` since 08-26 03:53Z, occupant `gcg-session-557f...` open but `state=asleep`.

What it does buy is that ga-jrnou becomes **repairable**. It narrows the liveness question from a 24-way disjunction over everyone who ever sat in the chair to a 1-way identity on the actual occupant, which makes closing the wedged session bead a *sufficient* remedy instead of merely a necessary one.

**Deliberate behavior change:** the crash-recovery tier of `standardAssignedInProgressWorkQueryScript` loops `$GC_SESSION_NAME`, so today a fresh occupant silently picks up a dead occupant's `in_progress` bead. After this change it stops. That is the point — it is work theft, and it collapses pool demand accounting at `assigned_work_scope.go:200` — but until ga-jrnou lands it can make stalls *more* visible before fewer.

## Test

`cmd/gc/hook_claim_occupant_identity_test.go`: a precedence table over the five environment shapes a claim runs in, plus an end-to-end `gc hook --claim` against a fake `bd` that records `BEADS_ACTOR`, asserting an unaliased pool worker claims under its session bead id and that the slot label never appears as the actor. `TestCmdHookClaimAliasedSessionStillClaimsUnderItsAlias` is the named-agent control.

Full `cmd/gc` suite green: `ok github.com/gastownhall/gascity/cmd/gc 716.513s`, 0 failures.

Refs ga-70chz. Independent of #5660 (both needed); complementary to open #5625.

### DIFF
--- a/cmd/gc/cmd_hook.go
+++ b/cmd/gc/cmd_hook.go
@@ -447,9 +447,7 @@ func cmdHookWithOptions(args []string, opts hookCommandOptions, stdout, stderr i
 	sessionID := strings.TrimSpace(overrides["GC_SESSION_ID"])
 	sessionName := strings.TrimSpace(sessionForQuery)
 	alias := strings.TrimSpace(overrides["GC_ALIAS"])
-	// Write the alias/agent form that read paths query through GC_AGENT.
-	// Session forms remain fallbacks for unaliased workers.
-	assignee := firstNonEmptyHookValue(alias, agentForQuery, resolvedAgentName, sessionName, sessionID)
+	assignee := hookClaimAssigneeIdentity(alias, sessionID, agentForQuery, resolvedAgentName, sessionName)
 	// IdentityCandidates governs ADOPTION of already-owned in_progress/open
 	// work (hookClaimExistingAssignment, claimFirstReadyHookAssignment, and
 	// the display path's hookCandidateVisible own-work check); it must be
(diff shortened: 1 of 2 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277242912

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR fixes a real
claim-identity bug in `gc hook`: an unaliased pool worker was recording its
claim under the reused pool-slot label (the "chair") instead of its own unique
session bead id. Because the slot label is shared across whoever occupies the
chair over time, a dead occupant's `in_progress` bead read as "still held" via
the next live occupant and never released — stranding work on the exact
claim/liveness machinery the pool workers depend on. Fixing the writer to record
under the unique session bead id restores correct liveness attribution and stops
the strand.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- Stranded dead-occupant work: the writer's old precedence stamped the
  slot-derived `poolRuntimeSessionName` (a non-unique chair), so
  `liveOpenSessionAssignmentExists` treated the chair as "held" through the next
  live occupant and a dead occupant's `in_progress` bead never released
  (`cmd/gc/compute_awake_set.go:739`, `cmd/gc/pool_session_name.go:246,275`).
  The patch records the claim under the unique session bead id, which every
  reader already leads with, so the fix is writer-only and the leak is closed at
  the source.
- Correctness of the *bare* id choice: recording the bare session bead id
  (rather than a `<base>-<beadID>` composite) is load-bearing —
  `liveSessionBeadExistsByIdentity` re-requires the exact assignee among the
  bead's own identities, so a composite would read "not live" and the sweep
  would clear a *live* claim. Review validated this and confirmed the aliased
  and empty-assignee fallback paths are preserved.

Both reviewers independently reproduced the mechanism from source and ran the
four targeted `cmd/gc` claim tests green, so the change is well-scoped and
cleanly revertable.

Non-blocking follow-ups (shared for visibility — NOT completed in this PR, each
filed as a tracking issue):
- `mc-rnoo` — Test Evidence Quality (minor): add one end-to-end
  unaliased-pool-worker claim test that drives the real `GC_AGENT`/`BEADS_ACTOR`
  work-query path. The new tests cover the precedence helper and the recorded
  actor, but not the full runtime actor-selection flow `cmdHookWithOptions`
  builds around it. Regression hardening only; not addressed here.
- `mc-4496` — Resource Lifecycle & Cleanup (minor): run a one-time operational
  sweep for `in_progress` beads already stranded under the old chair
  `session_name` before this fix deploys. The code change only corrects new
  claims, so pre-existing strands are not auto-healed. Separate operational
  remediation; not addressed here.
- `mc-5x4k` — Debuggability & Operability (nit): optionally resolve raw
  `gcg-session-…` assignee ids to their friendly `session_name` on human-facing
  surfaces. A correct-but-opaque id is preferable to a friendly-but-shared name,
  so this is strictly cosmetic; not addressed here.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `fd80fb812f` has passing required GitHub
checks (CI, CodeQL analyze jobs, and the required gate all green), and the
merge-ready metadata is in place. I am marking this PR `status/merge-ready` so

CI: https://github.com/gastownhall/gascity/actions/runs/33050502351

### Review Iterations
1 review pass performed — approved on the first pass with no fix iterations
required.

---

PR #5664, opened 2026-08-26
fix(dolt): admit the concurrent-UPDATE writer race in the compact preservation proof

### LINKED ISSUE #3341: Dolt compact integrity check races continuous writers — quarantine loop silently blocks GC and backups on busy DBs

csauer02-personal-user:
## Problem

The post-flatten integrity check compares a pre-flight snapshot against the
LIVE working set — `row_count` is `SELECT COUNT(*) FROM tbl` and
`table_value_hash` is `SELECT DOLT_HASHOF_TABLE(tbl)`, both unpinned, while
`db_value_hash` one function below IS pinned to `HEAD` and carries a comment
naming this exact hazard. On a continuously-written DB an in-place UPDATE
landing in the flatten window drifts a table's value hash with no row-count
change, which verify_counts categorises as `same_row_count_hash_drift`.

All three writer-race defer branches required
`verify_counts_saw_same_count_hash_drift != 1`, so that category disqualified
every one of them: the ladder covered concurrent INSERT (gain) and concurrent
DELETE (decrease) but had no branch for concurrent UPDATE — the dominant write
shape on a busy issues table (bd update, heartbeats, status/metadata writes).
A HEAD-proven writer race still hard-quarantined, and the marker blocks all
future GC of the database, so the store cannot shrink.

Production: hq re-quarantined roughly every two hours (seen_count 113, since
2026-07-29) and dip likewise (35, since 2026-08-04), 1:1 with mol-dog-compactor
exit-1. Read-only probe of a live store, 90s apart, no writes from the prober:
the working set moved (events 320522 -> 320560, issues 27629 -> 27632) while
the same revision pinned at the commit was byte-identical across both reads. A
flatten of a 320k-row table takes far longer than 90s, so the verification is
unsatisfiable by construction on a live store.

## Fix

Generalise the existing Option A (#2846) proof instead of adding a fourth
heuristic. Option A already proves the real safety property directly — "every
pre-flight row remains reachable at the flatten commit" — by diffing the
pre-flight snapshot HEAD against the flatten commit.

1. `drift_preserves_preflight_rows` (was `gain_drift_is_additive_only`) now
   asks for no `removed` rows rather than added-only. `added` rows are a
   concurrent INSERT and `modified` rows a concurrent UPDATE; neither can drop
   a pre-flight row, so only a removal is loss. This is the preservation claim
   itself and it subsumes the previous additive-only case.
2. `same_row_count_hash_drift` no longer disqualifies THAT branch, and the
   branch no longer requires a row gain, so a drift-only failure reaches the
   proof whether the writer INSERTed, UPDATEd, or both. The proof runs over the
   union of the gain-drift and same-count-drift table lists, so a mixed shape
   must prove every drifted table.
3. `same_row_count_hash_drift` stays disqualifying for the two HEAD-proxy
   branches around it — those infer preservation from HEAD movement and have no
   direct evidence.
4. `diff_is_additive_only` is unchanged and still used by the committed-root
   drift proof's first-committed table case, where a table absent from every
   pre-flatten commit root must be wholly new to be provable.

Concurrent DELETE still shows as `removed`, still fails this proof, and still
falls through to the existing row-decrease + writer-race branch. Everything
unproven — a removed row, a diff-probe failure, table-list drift, a probe
failure — fails closed exactly as before.

Tradeoff, stated explicitly for review: additive-only proved every pre-flight
row was byte-identical at the flatten commit; no-removed proves every
pre-flight row is still reachable. The flatten is a soft reset plus
`DOLT_COMMIT -Am`, which re-commits the live working set and cannot rewrite
cell values, so a `modified` row can only be a writer's own write — the same
reasoning that already justifies admitting `added`. The stronger claim is what
made the check unsatisfiable on a live store.

## Verification

- TDD: the two new defer tests were falsified first — they fail against
  unmodified canonical scripts (both quarantine) and pass with the fix, so they
  are load-bearing rather than tautological.
- New fake-dolt scenarios/tests: same-count drift proven row-preserving ->
  defer (the hq/dip production shape); mixed gain+drift and same-count drift ->
  defer over the union of both table lists; preservation probe failure ->
  quarantine (this branch has no HEAD proxy behind it, so it must fail closed).
- The fake-dolt DOLT_DIFF arm for verified tables is placed above the row-count
  arms: the proof is a `SELECT COUNT(*)` naming the same table, so those arms
  would otherwise answer it with a row count.
- Existing fail-closed pins unchanged and still green, now exercising the real
  proof path rather than an unhandled-query error:
  `TestCompactScriptQuarantinesSameRowCountWriterBeforeFullGC`,
  `TestCompactScriptStillQuarantinesGainAndHashDriftWithStableHead`,
  `TestCompactScriptQuarantinesMixedSignalsDespiteWriterRace`,
  `TestCompactScriptQuarantinesMixedRowGainAndSameCountHashDriftBeforeFullGC`,
  row-decrease and dolt_schemas system-table drift.
- `go test ./examples/bd/dolt/` green (full package); `go vet ./examples/bd/dolt/...`
  clean; `go test ./internal/builtinpacks/...` green.
- `test/dolt/compact_gain_drift_proof_test.sh` rewritten to model the two
  DOLT_DIFF predicates separately, so the gap between the proofs (a
  modified-only diff: refused by additive-only, admitted by preservation) is
  pinned rather than assumed. 16/16 pass.

## Not done

- Quarantine markers are not cleared — that carries the `clear_decision`
  evidence bar and is an operator call (and the #3584 revert stands).
- Optional follow-up, orthogonal to this fix: pin `row_count` and
  `table_value_hash` to the revision under verification so the two probes at
  least agree with each other, removing the intra-loop race at source. Drift
  against the pre-flight snapshot would remain, so the proof is still required.
- `test/dolt/*.sh` are still not CI-wired (tracked upstream in ga-7yrto); the
  Go fake-dolt tests are the CI-enforced proof of this change.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_016j4o3961MTbXkBq5jP35Ah

---
Re-verified against current `origin/main` (87ece235f) before opening: clean cherry-pick, and `go test ./examples/bd/dolt -count=1` passes (128s) on the rebased head. Fixes #3341.

### DIFF
--- a/examples/bd/dolt/assets/scripts/compact-gain-drift-proof.sh
+++ b/examples/bd/dolt/assets/scripts/compact-gain-drift-proof.sh
@@ -1,20 +1,21 @@
 #!/bin/sh
-# compact-gain-drift-proof.sh — Option A row-preservation proof for the
-# post-flatten gain+drift case (gastownhall/gascity#2846).
+# compact-gain-drift-proof.sh — Option A row-preservation proof for post-flatten
+# table value-hash drift (gastownhall/gascity#2846, widened for #3341).
 #
-# When verify_counts sees a table gain rows AND its value hash drift, the
-# safety property at stake ("pre-flight rows remain reachable") cannot be
-# inferred from HEAD movement alone: a concurrent writer whose commit is
-# ABSORBED into the flatten commit moves no HEAD, so the HEAD-proven gate
-# misses it and a benign race is hard-quarantined — which then blocks all
-# future GC of a busy DB (the memory-exhaustion failure the code calls out).
+# When verify_counts sees a table's value hash drift, the safety property at
+# stake ("pre-flight rows remain reachable") cannot be inferred from HEAD
+# movement alone: a concurrent writer whose commit is ABSORBED into the flatten
+# commit moves no HEAD, so the HEAD-proven gate misses it and a benign race is
+# hard-quarantined — which then blocks all future GC of a busy DB (the
+# memory-exhaustion failure the code calls out).
 #
-# This proves preservation DIRECTLY: for each gained+drifted table, diff the
-# pre-flight snapshot HEAD against the flatten commit. If the only change is
-# `added` rows (no `removed`/`modified`), every pre-flight row survived and the
-# gain is concurrent-writer data — defer, exactly as the HEAD-proven path does.
-# It is strictly more rigorous than the HEAD proxy: it proves reachability
-# instead of inferring it. Any removed/modified row, or any probe failure,
+# This proves preservation DIRECTLY: for each drifted table, diff the pre-flight
+# snapshot HEAD against the flatten commit. `added` rows are a concurrent
+# INSERT and `modified` rows a concurrent UPDATE; neither can drop a pre-flight
+# row, so a diff carrying no `removed` row proves every pre-flight row survived
+# and the drift is concurrent-writer data — defer, exactly as the HEAD-proven
+# path does. It is strictly more rigorous than the HEAD proxy: it proves
+# reachability instead of inferring it. Any removed row, or any probe failure,
 # fails closed and falls through to quarantine.
 #
 # Depends on `query_single_cell` and `valid_table_name` from run.sh.
@@ -23,9 +24,11 @@
 # Returns 0 iff the table's <from>..<to> content diff contains only `added`
 # rows. Returns non-zero (fail closed) if either commit endpoint is missing,
 # the table name is missing or invalid, the diff probe fails or returns a
-# non-numeric result, or the table shows removed/modified rows. Shared by the
-# gain+drift preservation proof and the committed-root drift proof's
-# first-committed table case (run.sh db_root_drift_within_verified_tables).
+# non-numeric result, or the table shows removed/modified rows. Used by the
+# committed-root drift proof's first-committed table case (run.sh
+# db_root_drift_within_verified_tables), where the table exists in no
+# pre-flatten commit root and so must be wholly new to be provable — a
+# strictly stronger claim than the row-preservation proof below.
 diff_is_additive_only() {
   _da_db="$1"
   _da_from="$2"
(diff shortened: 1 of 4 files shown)


sjarmak:
Closing in favour of #5749, which admits the same concurrent-UPDATE writer race in the compact path.

Two of you found this independently, which is a decent signal the race is easy to hit. Thanks for the writeup.

PR #5665, opened 2026-08-26
Hoist a rig pack's formulas with its city-scoped agents

psbanka:
## Summary

A pack reached only through a rig include or import can declare city-scoped agents. `expandPacks` hoists those agents and their named sessions up to city scope rather than dropping them (`hoistCityScopedAgents`, added for ga-hy0co). **The hoist stopped there:** the pack's `formulas/` dir went only into that rig's formula layer.

`ComputeFormulaLayers` builds a rig's layers as the city layers *plus* the rig's own, never the reverse. So a hoisted city-scoped agent can never resolve a formula from the pack it came from. The agent registers, and the formula it executes stays somewhere it does not look.

**What it broke, concretely.** In the `gc-scratch` city, a commit moved the gastown import out of `[defaults.rig.imports]` and onto one rig, to stop stamping an always-on witness onto every rig. That part worked. But `gastown.deacon` and `gastown.boot` hoisted to city scope as designed while `mol-deacon-patrol` stayed on the rig, so every patrol cycle failed to pour and the loop stopped rather than losing its wisp. The escalation path went with it: a warrant routes to the dog pool, the dog runs `mol-shutdown-dance`, and that formula was stranded the same way. The watchdog was disabled by the bug it would have been reporting.

`hoistCityScopeFromRigPack` now promotes the pack's `formulas/` dirs to the city formula layer when, and only when, that pack contributed a hoisted city-scoped agent or named session. A rig-only pack promotes nothing, so rig-scoped packs stay rig-scoped.

**Verified against a live city with no config change:** all four gastown formulas (`mol-deacon-patrol`, `mol-shutdown-dance`, `mol-refinery-patrol`, `mol-witness-patrol`) resolve at city scope, and `gc agent list` is byte-identical before and after at 80 entries, still exactly one witness. The fix widens formula visibility only.

## Testing

- [ ] `make check` — **not green, and not clean-red.** One package, `./scripts`, was killed at 901s against the 15m timeout. **Zero assertion failures** (`grep -c '^--- FAIL'` = 0). It is a timeout under load: the machine was at load average 127 on 10 cores from concurrent agent work. `go list -deps ./scripts/` returns no gascity package except `scripts` itself, so `internal/config` is not in its import graph and this change cannot reach it. The same package passed at 290s on the same tree earlier. Needs a re-run on a quiet box before merge, and I am not claiming the box otherwise.
- [x] `go build ./...` clean, `go vet ./internal/config/` clean
- [x] `go test ./internal/config/` — **ok, 53.5s, full package**
- [ ] `make check-docs` — n/a, no docs, navigation, or links changed
- [ ] `make test-integration` — n/a, this is config composition, not runtime, controller, or workflow behavior

## Checklist

- [x] Linked an issue — ga-kp9; city-store bead gs-gd0z records the symptom
- [x] Added or updated tests for behavior changes — one positive (`TestHoistedCityAgentSeesItsPackFormulas`) and one negative (`TestRigOnlyPackWithoutCityAgentsKeepsFormulasAtRigScope`), so the promotion is pinned in both directions
- [ ] Updated docs for user-facing changes — none; this restores intended behavior rather than changing it
- [x] Called out breaking changes or migration notes — none. Formula visibility widens for cities that already hoist an agent; the agent set is unchanged

https://claude.ai/code/session_01EedCADsZczYgxLdtSqQnFc

### DIFF
--- a/internal/config/compose.go
+++ b/internal/config/compose.go
@@ -125,9 +125,14 @@ type LoadOptions struct {
 	// discovery, shell completion — set this and degrade to "no pack state
 	// right now"; a load whose result the caller acts on must leave it false
 	// and wait, because a busy cache would otherwise read as an empty one.
-	RepoCacheNonBlocking    bool
-	deferRigPatches         bool
-	deferredRigPatches      *[]deferredRigPatches
+	RepoCacheNonBlocking bool
+	deferRigPatches      bool
+	deferredRigPatches   *[]deferredRigPatches
+	// hoistedCityFormulaDirs collects the formulas/ dirs of rig-scope packs
+	// that contributed a city-scoped agent or named session. Hoisting the
+	// agent without its formulas leaves it unable to resolve the formula it
+	// executes (gs-gd0z), so the dirs are promoted to the city layer too.
+	hoistedCityFormulaDirs  *[]string
 	allowLegacyOrderLayouts bool
 }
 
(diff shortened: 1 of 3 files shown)


psbanka:
Closing at the repository owner's request. This was opened without checking first, which I should have done for a public repo. The work is preserved and can be reopened when appropriate.

PR #5666, opened 2026-08-26
fix(beads): harden native admission and conditional writes

jm2:
## Summary

- make fresh managed-Dolt admission fail closed, inode-serialized, idempotent across config reloads and endpoint races, and seed fresh stores before the managed provider starts
- preserve standalone-bd convergence diagnostics before fresh-root validation and build the hard-kill/rebind fixture through the production admission lifecycle
- preserve force-close versus policy-aware conditional-close semantics while making native conditional writes retry-safe and at-most-once
- consume the Beads composite transaction revision needed to keep issue, label, dependency, and event planes consistent
- add schema-v66, real-Dolt, lifecycle-race, close-policy, fresh-init, and dependency-surface coverage

## Dependency / draft status

This draft depends on gastownhall/beads#6015. The current branch uses an exact checksummed replacement to `jm2/beads@b80784c09af4` so local end-to-end validation and a guarded deployment can proceed. Before this PR is marked ready, that replacement will be removed, the merged upstream Beads pseudo-version will be pinned directly, and the no-replace policy gate will be green. The guard is deliberately not weakened for the temporary fork pin.

The branch is rebased on current upstream GasCity `8556a801c380ba9e43c04daa58c969f988021324`; its current head is `f03efa7860bc877a4397aef3255f8c6c0ed80319`. The exact one-file stop rollback fix from #5667 remains temporarily layered because safe shutdown is required for deployment rehearsal; it will be dropped after #5667 merges.

## Validation

- admission-focused `cmd/gc` suite passes, including the standalone-bd convergence regression
- real managed-provider hard-kill/port-rebind integration passes against the exact branch source
- full pre-push fast matrix passed every shard except the deliberate no-`replace` policy failure
- full native/current Beads compatibility and adversarial real-Dolt transaction coverage
- `make test-ci-policy`, commit-hook lint/format/generated-reference checks, and `go vet ./...`
- stop rollback regression 10/10 plus its focused/race suite on #5667
- hosted CI exposed and the successor commit banked the two removed subprocess sites in every checked resource-census baseline; the focused census package now passes
- hosted CI is rerunning on the current head; the temporary fork replacement remains an expected policy failure
- the container scan also reports newly disclosed dependency CVEs on the current dependency graph; those remain a separate merge-readiness gate and are not being suppressed

Do not merge while draft or while the fork replacement remains.


jm2:
Closing this obsolete release draft. It is based on the pre-incident f03/8556 candidate, pins the temporary b807 Beads build, and carries the now-superseded #5667 stop layer. Native admission and Beads recovery work will be rebuilt upstream-first on current main with fresh safety evidence rather than extending this backport stack.

PR #5676, opened 2026-08-26
fix(stop): roll back the pending unregister on the uncapped stop path too

julianknutsen:
## Summary

#4999 made the supervisor unregister a pending transaction so a stop that fails after removing the registration hands the entry back — but only the wall-clock-capped dispatch got the transaction. The uncapped path (the default: `gc stop` with no `--timeout`) still passed `nil`, so the unregister committed immediately and a managed provider that refused to shut down left a live city unregistered with nothing to restore.

#4999's own `TestCmdStopSupervisorManagedInvalidCityTomlFailsWhenShutdownFails` asserts the restore on exactly this path, and has failed deterministically on every main push since the merge (e.g. run 32970139971 — both `Preflight / unit cover (cmd/gc 1 of 6)` and `cmd/gc process / shard 1 of 12` fail on this one test). **This is the current deterministic main-red.**

The fix gives the uncapped dispatch the same transaction and mirrors the capped arm's accept/rollback verbatim: commit on success, restore-and-report on failure. The deferred success-message ordering matches the capped path by design.

## Testing

- `go test ./cmd/gc -run 'TestCmdStopSupervisorManagedInvalidCityTomlFailsWhenShutdownFails' -count=1` — fails on main, passes here
- `go test ./cmd/gc -run 'TestCmdStop|TestSupervisorUnregister|TestCmdUnregister' -count=1` — 27/27 PASS
- `go test ./cmd/gc -run 'TestCmdStop|TestSupervisorUnregister' -race -count=3` — ok
- `go build` / `go vet` / `gofmt` clean

The earlier same-day red main runs (08:43Z, 09:07Z) failed different integration jobs and are judged separately once this lands.

### DIFF
--- a/cmd/gc/cmd_stop.go
+++ b/cmd/gc/cmd_stop.go
@@ -84,7 +84,19 @@ func cmdStopJSON(args []string, stdout, stderr io.Writer, wallClockTimeout time.
 			return cmdStopJSONSequence(args, stdout, stderr, force, jsonOut, true, unregisterTx)
 		})
 	} else {
-		outcome = cmdStopJSONSequence(args, stdout, stderr, force, jsonOut, false, nil)
+		// The uncapped path holds the same pending-unregister transaction as
+		// the capped one: a stop that fails after removing the registration —
+		// a managed provider that refuses to shut down, an invalid config the
+		// body cannot recover — must hand the entry back rather than leave a
+		// live city unregistered. This mirrors the capped arm's accept/rollback
+		// exactly; the deferred success message is part of the same contract.
+		unregisterTx := newSupervisorUnregisterTransaction()
+		outcome = cmdStopJSONSequence(args, stdout, stderr, force, jsonOut, false, unregisterTx)
+		if outcome.code == 0 {
+			unregisterTx.commit()
+		} else {
+			writeSupervisorUnregisterRollback(stderr, "gc stop", "stop failed after unregistering city", unregisterTx.rollback())
+		}
 	}
 	if outcome.code != 0 {
 		return outcome.code


julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR closes a real rollback gap
in `gc stop`: on the uncapped stop path the supervisor unregister was committed
eagerly, so a stop that failed *after* the registration was removed left a live
city unregistered with nothing to restore. That matters because it could strand
a running city outside the supervisor registry on a partial-failure stop — the
exact state operators can't easily recover from.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- The uncapped `gc stop` arm passed `nil` for the unregister transaction and took
  the eager-commit branch, so there was no rollback if a later stop-body step
  failed. The final patch threads the same `supervisorUnregisterTransaction` into
  the uncapped dispatch and mirrors the capped arm's accept/rollback verbatim
  (`cmd/gc/cmd_stop.go`), so a failed stop now restores the registration instead
  of leaking an unregistered-but-live city. Both reviewers independently
  reproduced the pre-fix failure and confirmed the fix turns a standing,
  deterministically-red regression test —
  `TestCmdStopSupervisorManagedInvalidCityTomlFailsWhenShutdownFails` — green,
  and ran the targeted `TestCmdStop*` / supervisor-unregister suites clean
  (`go vet ./cmd/gc/` clean).

- `mc-b5rj`: (test-coverage / maintainability, nit) The new outer rollback is
  reached by two failure routes — supervisor-alive (asserted, now green) and
  supervisor-not-alive with a non-zero stop body (`cmd/gc/cmd_stop.go:157`,
  currently unasserted) — plus the JSON-mode restore-on-failure path, whose
  coverage is indirect through the shared stop path. The rollback code is shared
  and identical across routes, so functional risk is low; this is a coverage gap,
  not a defect. A sibling regression test (`supervisorAliveHook()==0`, a
  registered city, a stop body forced to fail, asserting the registry entry is
  restored) would close it. Tracked as a follow-up; not required for this merge.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `7f391e1e0b` has passing required GitHub
checks (the only red is the advisory, non-required Mac Regression job, which does
not gate merge), and the merge-ready metadata is in place. I am marking this PR

CI: https://github.com/gastownhall/gascity/actions/runs/33006467143

### Review Iterations
The maintainer review loop ran to approval; the final review pass converged with
no blocker or major findings across correctness, contract fidelity, concurrency,
error handling, security, and release safety, leaving only the single non-gating
coverage nit noted above.

---

PR #5677, opened 2026-08-26
feat(packs): run pack lifecycle hooks on city start and stop (ga-b0o)

brandonmartin:
## Summary

Split out of #5631 per review — this is the pack-lifecycle contract on its own, so it gets its own review. **It changes `docs/reference/specs/pack-spec.md`** (new `lifecycle/` convention), so it wants the pack-spec owner's eyes specifically.

A pack can own a service the city knows nothing about — a systemd unit, a container, an external daemon. Until now nothing brought such a service down with `gc stop`, so a pack-owned daemon outlived the city that hosted it and the operator had to stop it by hand.

This adds convention-discovered pack lifecycle hooks:

- A pack ships an executable `lifecycle/city-start.sh` and/or `lifecycle/city-stop.sh`; the city runs it when that event fires. Only those two names are executed and only regular files are eligible, so an unrelated `lifecycle/` layout cannot be run by accident.
- Every pack composed into the city contributes, city packs before rig packs in rig-name order, so execution order is stable across runs.
- Hooks run with the pack directory as cwd and receive the city runtime env plus `GC_PACK_DIR` and `GC_LIFECYCLE_EVENT`.
- Execution is best-effort and bounded: each hook gets 30s and is then killed; a hook that fails or hangs is reported on stderr without changing the exit status. A broken pack cannot wedge city startup or shutdown.
- `city-stop` fires after agent sessions are stopped and before the bead store shuts down, so a hook can still reach the ledger as it tears down. In the supervised path it runs from `CityRuntime.shutdown()` under the same `ownedCity` guard the server teardown uses, so a discarded runtime never stops the live owner's services. `city-start` fires once the city is ready, mirroring that order.

New package: `internal/packlifecycle/` (runner) + `internal/config/lifecycle_discovery.go` (discovery). Docs: `docs/guides/shareable-packs.md`, `docs/reference/specs/pack-spec.md`.

## Test plan

- `go test ./internal/packlifecycle ./internal/config ./cmd/gc -run 'Lifecycle|PackLifecycle' -count=1`
- Discovery rejects non-regular files and names other than the two hook names; runner enforces the 30s bound and never alters exit status.
- Full local gate (`go vet`, `go build`, `make test`) run on our integration branch with this commit included.

Closes ga-b0o. Cherry-picked onto current `main` from e5e729f73 (the commit formerly riding on #5631).


brandonmartin:
Rebased this branch onto current `upstream/main` (`9700d9a48`) and force-pushed `2e410b408`. The only conflict was the pack-spec section numbering after upstream added `[runtimes.<name>]`; resolution preserves that full upstream section and places Lifecycle Directory at 1.2.15, with subsequent headings renumbered consistently. Fresh verification passes: `internal/config`, `internal/packlifecycle`, focused `cmd/gc` lifecycle/shutdown tests, `make check-docs`, focused vet, and range diff check. Ready for refreshed CI/review.

brandonmartin:
Rebased cleanly onto current `main` (`c6c81ec28`) and force-pushed `d978e6bed` with an exact lease so fresh CI replaces the previous vulnerability/evidence run. There were no new conflicts or source edits in this refresh.

quad341 (approved):
@brandonmartin — thank you for splitting this cleanly out of #5631 and for the care in the managed-shutdown half specifically. The three `CityRuntime.shutdown()` tests are what I want to call out: pinning the `ownedCity` guard *and* the preserve-sessions early return means the next person who reorders that teardown gets a failing test instead of a pack daemon quietly outliving a supervisor handoff. That's a miserable bug to debug in the wild, and you closed it before it existed. The runner's timeout test asserting elapsed time rather than just the error string is the same instinct.

Three things I'll take on my side: the rebase dropped the three `[runtimes.<name>]` rows from the pack-spec index tables (§1.2.8 itself survived, so it's just the summary rows), a doc comment on `runPackLifecycleHooks` still mentions a `logPrefix` parameter, and I'll switch discovery to `Lstat` so it rejects symlinks the way its own comment promises. I'll also add the one test your stop path has and your start path doesn't — that `run()` actually fires the start hook.

The red CI isn't yours: shard 7 fails on a bd schema-version gate (v48→v66) during city init, and `main` is red on the same thing. We'll land this once that clears. Flagging the new `lifecycle/` convention for the pack-spec owner, as you asked.


---

PR #5678, opened 2026-08-26
fix(bd): reach the wisps tier on gc bd list --type=molecule (ga-bdmol)

brandonmartin:
## Summary

Split out of #5631 per review — a `gc bd` behavior change that deserves its own title in the merge log.

bd's `list` sets `SkipWisps` on every query that does not pass `--include-infra` (beads `cmd/bd/list_filter.go`), and molecules and wisps live in that tier. So `gc bd list --type=molecule` returned `[]` with exit 0 on a ledger holding live molecules — an empty result that reads as "no molecules exist" rather than "this query cannot see them".

That silent blindness broke witness patrol-wisp reconciliation: the startup query found nothing every cycle, so leaked wisps were never burned and accumulated across restarts (one rig found three leaks in a single patrol that reconciliation should have caught).

gc already disagreed with itself here: `BdStore.listViaBDList` passes `--include-infra` unconditionally, so a molecule visible through the store seam was invisible through the passthrough seam. This closes the gap in the passthrough, widening **only** a list whose filters name the wisps tier (`--type=molecule/wisp`, `--mol-type`, `--wisp-type`); every other list is forwarded exactly as written. An argv the scanner cannot parse (an unrecognized flag whose value consumption is undecidable) is forwarded untouched rather than guessed at.

Files: `cmd/gc/bd_wisp_tier.go` (+test), `cmd/gc/cmd_bd.go`, `docs/reference/cli.md`.

## Test plan

- `go test ./cmd/gc -run 'WispTier|BdList' -count=1`
- Verified against a live ledger: `gc bd list --type=molecule` returned `[]` before and returns the live molecule roots after; unfiltered and `--type=bug` listings unchanged.

Cherry-picked onto current `main` from cf8329400 (the commit formerly riding on #5631).


brandonmartin:
Rebased the one-commit fix conflict-free onto current `upstream/main` (`9700d9a48`) and force-pushed `6f60ad5d2`. The previous required/cmd-gc failure does not reproduce in fresh focused validation: both wisp-tier rewrite tests pass, `go vet ./cmd/gc` is clean, `go run ./cmd/genschema` is idempotent, and `make check-docs` passes. The generated `docs/reference/cli.md` remains owned by the Cobra source in `cmd/gc/cmd_bd.go`; no hand-edited generated output. Ready for refreshed CI/review.

quad341 (approved):
@brandonmartin — thank you for splitting this out of #5631 and for writing the *why* into the code instead of only the PR body. The doc comment on `rewriteBdWispTierArgs` names the exact disagreement it closes, and I confirmed it against the tree: `BdStore.listViaBDList` really does pass `--include-infra` unconditionally, so the store seam and the passthrough genuinely disagreed about the same query. The "fails open — an argv we can't parse is forwarded untouched" rule is the right instinct for a wrapper that rewrites someone else's command line, and the test cases distinguishing `--label molecule` and `--title-contains=molecule` from a real type filter are the ones that would have bitten later.

Nothing for you to do. The single red check, `Evidence / critical-path suites`, is ours: it hit its window at 03:46 while the `cmd/gc` shards hadn't started yet — they and `CI / required` all went green by 04:20 on this same SHA. I'll re-run it and merge as-is, no changes to your branch.

Two things I'll file separately rather than hold this up: `wispTierTypes` mirrors a bd-side contract with nothing pinning it, and a trailing value flag (`--type=molecule --limit`) makes the appended flag land as that flag's value. Yours if you'd rather take them.


---

PR #4665, opened 2026-07-26
Fail gc sling when routed work is invisible to the target query

quad341:
## What this changes

`gc sling` now verifies that routed work is visible to the target agent's effective `work_query` before reporting success. If routing metadata is written but the query still excludes the bead—for example because the query filters on a missing label, dependency state, or hold—the command exits non-zero with an error naming the bead and target instead of silently stranding the work.

The check runs through the same command, environment, working-directory, and timeout machinery that `gc hook` uses. It therefore tests the target's real pickup surface rather than duplicating or trying to interpret pack-specific query rules.

## Review notes

- The new `SlingDeps.WorkQueryProbe` keeps shell execution at the CLI boundary; `internal/sling` only consumes the narrow callback.
- Query output is decoded as JSON and matched by the exact bead `id`, including array and single-object shapes.
- A failed postcondition occurs after the routing operation; the non-zero result is an explicit warning that the routed state exists but is not claimable through the target query.
- Dry-run remains side-effect-free, `--force` retains its missing-bead tolerance, and formulas v2 retain their existing workflow-launch visibility check.
- No configuration, API, schema, or migration changes are required.

## Test plan

- [x] `make test-fast-parallel` — all eight shards pass
- [x] `go test -count=1 ./internal/sling/...`
- [x] `go test -count=1 ./cmd/gc/... -run 'Sling'`
- [x] `go build ./...` and `go vet ./...`
- [x] Release gate: [`release-gates/ga-x8jw29-sling-work-query-visibility-gate.md`](release-gates/ga-x8jw29-sling-work-query-visibility-gate.md)

Issue #5668, opened 2026-08-26
TestCmdStopSupervisorManagedInvalidCityTomlFailsWhenShutdownFails fails deterministically on main


rjgeng:
`TestCmdStopSupervisorManagedInvalidCityTomlFailsWhenShutdownFails` (`cmd/gc/cmd_stop_test.go:898`) fails deterministically (5/5 local runs, `-count=5`) on current `main` at `8556a801c380ba9e43c04daa58c969f988021324`, in a clean worktree with no PR changes applied:

```
cmd_stop_test.go:898: registry after managed-provider stop failure = [], want exact original entry
--- FAIL: TestCmdStopSupervisorManagedInvalidCityTomlFailsWhenShutdownFails (1.0xs)
```

Not flaky — same failure every run. Confirmed independent of any in-flight PR by running it in an isolated `origin/main` worktree with zero other changes.

**Likely cause:** `#4999` ("fix(stop): restore registration on timeout", `2527f6869`, merged 2026-08-26 ~04:07 PT) is the most recent commit touching `cmd_stop.go`/`cmd_stop_test.go` and changes exactly the registry-restoration behavior this test asserts on. Haven't traced the exact mechanism further — flagging with the repro rather than guessing at the fix.

**Impact:** this is in the `cmd/gc process / shard 1 of 12` CI job, so it's currently failing `CI / required` and `Evidence / critical-path suites` on every open PR that rebases onto current `main`, regardless of what the PR itself touches — not scoped to any one branch.

Found while investigating an unrelated CI failure on #5617.

rjgeng:
Confirmed fixed by #5676 ("roll back the pending unregister on the uncapped stop path too"). Re-ran `TestCmdStopSupervisorManagedInvalidCityTomlFailsWhenShutdownFails` on current `main` (906368df4) — passes 3/3, was failing 3/3 before. Closing.

Issue #2450, opened 2026-05-21
bug: parallel tests stomp sessionProviderForStopCity global (test isolation gap)


iwata-1116:
## Symptom

`make test-fast-parallel` (the pre-commit hook test gate) fails under non-trivial load with:

```
--- FAIL: TestCmdStopSupervisorManagedInvalidCityTomlFailsWhenShutdownFails (0.70s)
    cmd_stop_test.go:557: path = "/tmp/gct.../gc-stop-timeout-...",
    want "/tmp/gct.../TestCmdStopSupervisorManagedInvalidCityTomlFailsWhenShutdownFail.../001"
```

Observed 2/4 consecutive `git commit` attempts on an unrelated branch (`fix(cmd/gc): scale supervisor city ready timeout via [daemon] config`, fix file = `cmd_supervisor_city.go`) — failure was deterministic at load avg ≳ 8 on an 8-core box and intermittent below.

## Root cause

`cmd/gc/cmd_stop_test.go` uses a package-level global `sessionProviderForStopCity` that several tests save and restore via `t.Cleanup`:

- L184–186: `TestCmdStopWallClockTimeoutBoundsDirectStop` (uses `shortSocketTempDir(t, "gc-stop-timeout-")`)
- L400–403: `TestCmdStopSupervisorManagedConfirmsShutdown*` family
- L466–469: another test setting the override per-subtest
- L535: `TestCmdStopSupervisorManagedInvalidCityTomlFailsWhenShutdownFails` (uses `t.TempDir()`)

`go test` runs siblings in parallel inside one `*testing.T` package. The save/restore pattern is **not** sufficient when two parallel tests both touch the same global at overlapping times — one test's override survives past its own `t.Cleanup` window into another's body.

The `path = "/tmp/gct.../gc-stop-timeout-..."` value in the assertion is the *other* test's `shortSocketTempDir` result leaking into the failing test's callback, which expected its own `t.TempDir()` (formatted with the long test-name suffix).

## Reproduction

```sh
go test ./cmd/gc -run "TestCmdStop" -count=20 -parallel=8
# or
make test-fast-parallel    # under load avg ≳ 8 on an 8-core box
```

Failure rate scales with load and test parallelism. Below `nproc` the tests are usually fine.

## Suggested fix shapes

[...]

PR #5679, opened 2026-08-26
fix(sling): refuse to route a blocked bead, and name the blockers

edwarddavison:
## What this changes

`gc sling` refuses to route a bead that open dependencies hold out of the ready set, and names the blockers.

Today it writes `gc.routed_to` and reports success for such a bead. Pool demand is counted with a **ready** query, so a blocked bead is invisible as demand and no session can ever spawn to claim it. Every line of the output reads like success:

```
$ gc sling flinders/gastown.polecat fph-aijq --nudge
Auto-convoy fph-9cnr
Slung fph-aijq -> flinders/gastown.polecat
No running sessions for "flinders/gastown.polecat" - poked controller for wake
```

Nothing says the bead cannot be picked up. A dry no-op that prints a convoy id is worse than an error, because it starts a clock the operator thinks is running. In the reported incident a P1 sat undispatched for ~4h and was escalated across rigs as a rig-specific pool defect; it was neither. The underlying cause was an inverted dependency edge, and this output is what made it undiagnosable.

After this change:

```
$ gc sling frontend/worker FE-5jz
REFUSED: FE-5jz is blocked by FE-1d7, FE-87a — routing it to frontend/worker cannot
result in a claim, because pool demand is counted with a ready query and a blocked
bead is never ready; close the blockers or re-run with --force
  blocked by FE-1d7 (in_progress) blocker three
  blocked by FE-87a (open) blocker two
$ echo $?
1
```

The refusal lands in `preflight`, **before** any routing metadata is written, so a refused sling leaves no half-applied state and fires no controller poke.

## Review notes

- `OpenBlockers` (`internal/sling/readiness.go`) mirrors the store's own readiness rule rather than reimplementing one: it reuses `beads.IsReadyBlockingDependencyType`, so an edge counts exactly when the ready query would count it. A dependency target missing from the store blocks, for the same reason `Ready()` blocks on it — its status resolves to neither `"closed"` nor anything else. That is the dangling/inverted-edge shape from the incident.
- Scope mirrors the sibling `shouldCheckDepCycle`: formula slings and inline text mint new beads that cannot carry blockers, and `--on` routes the workflow root rather than the source bead.
- `--force` is excluded entirely. It is documented to *suppress warnings* and to dispatch even when the bead does not resolve in the local store — precisely the state this check cannot read — so it stays the escape hatch the refusal points at and adds no store round-trip to the force path.
- `--dry-run` stays in scope but downgrades to a warning that keeps the conditional voice ("would not result in a claim"), since a preview that hides the blockers preserves exactly the misleading output this removes.
- A store fault while resolving blockers warns and proceeds. An unreadable graph is not proof the bead is blocked, and refusing legitimate work is the worse failure.
- No config, API, schema, or migration surface changes. `make docs` regeneration produced no diff.

### Relationship to #4665

[#4665](https://github.com/gastownhall/gascity/pull/4665) adds a *post*-routing probe of the target's `work_query`. The two are complementary and do not conflict logically, though both touch `internal/sling`:

|  | #4665 | this PR |
|---|---|---|
| When | after routing | before routing |
| Detects | bead invisible to the target's `work_query` (labels, holds, pack overrides) | bead held out of the ready set by open dependencies |
| On failure | routing already written | nothing written |
| Diagnostic | "check labels/metadata/holds against the query X runs" | names each blocker with id, status, title |

If #4665 lands first I am happy to rebase; the mechanisms are independent and the blocker list is the diagnostic payload the source report actually asked for.

## Test plan

- [x] `make test-fast-parallel` — all eight jobs pass
- [x] `go test -count=1 ./internal/sling/...`
- [x] `go test -count=1 ./cmd/gc/... -run 'Sling'`
- [x] `go build ./...`, `go vet ./...`
- [x] `golangci-lint run ./internal/sling/... ./cmd/gc/...` — 0 issues
- [x] `.githooks/pre-commit` active and green (lint-changed, docs regen, vet, fast suite)

New coverage: `internal/sling/readiness_test.go` (blocking-type filtering, missing dependency, duplicate edges, refusal + no-route-written, ready bead unaffected, `--force` skips the lookup entirely, `--dry-run` warns, scope table, store-fault fail-open) and `TestPrintBlockedErrorListsEveryBlocker` in `cmd/gc/cmd_sling_test.go`.

### Verified end-to-end against a real `bd` store

Built binary, sandbox city, real `bd` backend — not just fakes:

| Case | Before (v1.4.0) | After |
|---|---|---|
| blocked bead | `Slung FE-wss → frontend/worker`, exit 0 | `REFUSED …` + blocker lines, exit 1 |
| refusal side effects | — | no `gc.routed_to` written |
| `--dry-run` on blocked | preview only | warning + blocker lines, exit 0 |
| `--force` on blocked | routes | routes, exit 0 |
| blockers closed | routes | routes, exit 0 |
| unblocked bead | routes | routes, no new warning |

Also confirmed live that auto-convoy `sling-<id>` wrappers use `tracks` edges, which are correctly **not** treated as blocking — no false positives on that path.

Source report: gas-city bead `gci-1de`. The report's second, smaller item (the spurious `agent "…" not found in config` message on the wake path) is tracked separately and is not in this PR.

### DIFF
--- a/cmd/gc/cmd_sling.go
+++ b/cmd/gc/cmd_sling.go
@@ -819,6 +819,11 @@ func doSling(opts slingOpts, deps slingDeps, querier BeadQuerier, stdout, stderr
 			printMissingBeadError(stderr, missingBeadErr, missingBeadForceApplies(opts))
 			return 1
 		}
+		var blockedErr *sling.BlockedError
+		if errors.As(err, &blockedErr) {
+			printBlockedError(stderr, blockedErr)
+			return 1
+		}
 		fmt.Fprintln(stderr, err) //nolint:errcheck
 		return 1
 	}
@@ -891,6 +896,11 @@ func doSlingBatchWithJSON(opts slingOpts, deps slingDeps, querier BeadChildQueri
 		var missingBeadErr *sling.MissingBeadError
 		if errors.As(err, &missingBeadErr) {
 			printMissingBeadError(stderr, missingBeadErr, missingBeadForceApplies(opts))
+			return 1
+		}
+		var blockedErr *sling.BlockedError
+		if errors.As(err, &blockedErr) {
+			printBlockedError(stderr, blockedErr)
 			return 1
 		}
 		// In batch mode, per-child FailReasons have already been rendered
@@ -1038,6 +1048,16 @@ func printMissingBeadError(stderr io.Writer, err *sling.MissingBeadError, allowF
 		return
 	}
 	fmt.Fprintln(stderr, "  verify the bead ID; --force does not bypass missing source validation for formula-backed routes") //nolint:errcheck
+}
+
+// printBlockedError renders a refused sling plus one line per blocker, so the
+// operator sees which beads to close instead of having to go find them. The
+// error string itself stays single-line for log and errors.As use.
+func printBlockedError(stderr io.Writer, err *sling.BlockedError) {
+	fmt.Fprintln(stderr, err) //nolint:errcheck
+	for _, b := range err.Blockers {
+		fmt.Fprintf(stderr, "  blocked by %s\n", b) //nolint:errcheck
+	}
 }
 
 func missingBeadForceApplies(opts sling.SlingOpts) bool {
--- a/cmd/gc/cmd_sling_test.go
+++ b/cmd/gc/cmd_sling_test.go
@@ -2054,6 +2054,38 @@ func TestPrintMissingBeadErrorFormulaBackedDoesNotSuggestForce(t *testing.T) {
 	}
 }
 
+// A refused blocked sling must name every blocker on its own line, with the
+// blocker's status and title. Handing the operator "check the dependencies"
+// and nothing else is what turned a one-line dependency fix into a four-hour
+// stall in the original incident.
+func TestPrintBlockedErrorListsEveryBlocker(t *testing.T) {
+	var stderr bytes.Buffer
+	printBlockedError(&stderr, &sling.BlockedError{
+		BeadID: "FE-work1",
+		Target: "frontend/worker",
+		Blockers: []sling.Blocker{
+			{ID: "FE-dep1", Status: "open", Title: "invert the dependency edge"},
+			{ID: "FE-dep2", Status: "in_progress", Title: "land the parser fix"},
+			{ID: "FE-ghost"},
+		},
+	})
+
+	got := stderr.String()
+	for _, want := range []string{
+		"REFUSED",
+		"FE-work1",
+		"frontend/worker",
+		"--force",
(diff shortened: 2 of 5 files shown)


edwarddavison:
## CI triage: all 5 red checks are pre-existing on `main`, none caused by this PR

CI has now finished on `677c1f6c9` (it was still queued when this PR was last
updated). Five checks are red, but they cascade from exactly **two** leaf test
failures, and neither is caused by this change.

| Check | Root cause |
| --- | --- |
| `cmd/gc process / shard 2 of 12` | leaf failure 1 |
| `Integration / packages-core-3-of-4` | leaf failure 2 |
| `CI / required` | aggregates the two above |
| `CI / integration` | aggregates the two above |
| `Evidence / critical-path suites` | prwatchdog, fails only because `CI / required` did |

### 1. `TestCmdStopSupervisorManagedInvalidCityTomlFailsWhenShutdownFails` — pre-existing, deterministic

Already tracked as #5668 ("fails deterministically on main"). I reproduced it
independently in a clean detached worktree at `main` = `8556a801c`, with **zero**
PR changes applied:

```
cmd_stop_test.go:898: registry after managed-provider stop failure = [], want exact original entry
--- FAIL: TestCmdStopSupervisorManagedInvalidCityTomlFailsWhenShutdownFails (0.59s)
```

Byte-identical to the failure on this PR. Unrelated to `sling`.

### 2. `TestEventInstallCrashWindowCannotLeaveTwoNamesForOneReservation` — environmental flake

In `internal/productmetrics`, which this PR does not touch. Structurally it
*cannot* be affected by this change:

```console
$ go list -deps -test ./internal/productmetrics/ | grep -E 'gascity/(internal/sling|cmd/gc)$'
(no output — no dependency on internal/sling or cmd/gc)
```

Supporting evidence:
- Full package passes **3/3** locally on clean `main` (`8556a801c`).
- The same job fails on `main` too — run `32951335349` — but with a *different*
  test in the same package (`TestDisableAndPurgeExactTokenConflictAndPeerCleanRecovery`).
  Different tests, same package, different commits = package-level isolation
  flake under CI parallelism, consistent with #2450.

`main` itself is red in 5 of its last 6 CI runs, so this is a repo-wide
condition rather than anything specific to this branch.

### Consequence for merging

This PR will not go green on its own: #5668 fails deterministically on `main`, so
`CI / required` stays red for every PR until that is fixed. No rebase or code
change here would clear it. The diff in this PR is unchanged and still scoped to
`cmd/gc/cmd_sling.go`, `cmd/gc/cmd_sling_test.go`, `internal/sling/readiness.go`,
`internal/sling/readiness_test.go`, `internal/sling/sling_core.go`.

Happy to rebase if #5668 or #4665 lands first.

PR #5680, opened 2026-08-26
docs: define the Gas City identity separator contract

quad341:
## What this changes

Gas City now has an authoritative reference for how qualified Agent identities become tmux-safe session names: `/` encodes as `--`, while `.` encodes as `__`. The new page is linked from the specifications index and documentation navigation, and the source implementation points maintainers back to the written contract.

The contract also documents the current Beads comparison behavior and its version boundary. This is a documentation and contract-test change only; it does not alter runtime identity handling or update the Beads dependency.

## Review notes

- Review `docs/reference/specs/identity-separator-contract-v1.md` for the two-axis encoding, worked examples, and the distinction between Gas City minting identities and Beads comparing them.
- The Beads section deliberately calls out that Gas City's pinned Beads version predates the cited upstream canonicalizer; that caveat should remain until the dependency advances and the behavior is re-verified.
- `internal/agent/session_name.go` changes only its explanatory comment. The two new tests protect the document's presence, source link, and required contract terms.

## Test plan

- [x] Run `make test-local-full-parallel` with the rootless Podman test environment; confirm both new contract tests execute and pass.
- [x] Run `make test-ci-policy` and `make check-docs`.
- [x] Run `go build ./...` and `go vet ./...`.
- [x] Review the [release gate](release-gates/ga-x0q15u-identity-separator-contract-gate.md), including attribution for unrelated fleet failures.


quad341:
@quad341 — this is a genuinely good spec, and I checked your citations rather than trusting them. They hold up: `config.go:27`'s regex is exactly as quoted, `go.mod` really does pin beads at `…20260805093327`, and `internal/storage/issueops/identity.go` genuinely doesn't exist at that pin — your version caveat is honest. Pointing out that beads' own `canonicalActor` doc comment claims `-` and `_` are "never part of either name," when gascity's validator says otherwise, is the most valuable thing on the page.

One correction I'll take on my side. §2 says the distinct encodings make decoding unambiguous; I ran the replacers and `rig/builder--1` encodes to `rig--builder--1` and decodes back to `rig/builder/1`. `builder--1` is legal under the very regex §1 cites, and the source calls the reverse replacer "best-effort." I'll soften that to match, switch the collision example to the encoded spellings (where the merge actually happens), and add one round-trip test so the limitation is pinned by code — then re-run the contract tests and `make check-docs`.

The red shard-1 check is a `cmd/gc` registry-state failure, nothing to do with your diff. Thanks for writing this down.


---

PR #5686, opened 2026-08-27
fix(ci): a failed integration-test discovery is not an empty shard

julianknutsen:
## Summary

The rest-full/rest-smoke shard runner discovers its test set with `go test -tags integration -list` behind a process substitution, whose exit status is invisible to the parent even under `set -euo pipefail`. Module downloads happen inside that `-list` invocation, so a transient runner hiccup failed discovery silently, yielded an empty list, and surfaced as `no tests selected for shard rest-full-N-of-16` on whichever matrix jobs the hiccup hit — the wandering integration-shard reds on main (rest-full-2 at 08:43Z, packages-core-3 at 09:07Z, rest-full-5+6 at 01:03Z; runs 32949218048 / 32951335349 / 33009528457).

With #5676 fixing the deterministic cmd/gc red, these discovery flakes are the remaining cause of main-red.

Capture the discovery with its status intact, retry once for the transient class, and abort with the real reason on a second failure. The empty-shard guard in `run_pkg_tests_modulo` stays: with discovery status now propagated it again means what it says.

## Testing

Falsified with a shimmed `go` binary:
- fail-once → `listing integration tests failed; retrying once` → shard selects and runs (10 tests on rest-full-5-of-16)
- fail-always → `refusing to shard an unknown test set`, script exit 1 (not the misleading empty-shard message)
- unshimmed control → `Running shard rest-full-5-of-16 against ./test/integration (10 tests)`
- `bash -n` clean

### DIFF
--- a/scripts/test-integration-shard
+++ b/scripts/test-integration-shard
@@ -177,7 +177,21 @@ validate_modulo_shard() {
 }
 
 list_integration_tests() {
-  run_go_test -tags integration -list '^Test' "$pkg" | grep '^Test'
+  # Discovery is a network-and-compile step: module downloads happen inside
+  # the -list invocation, so a transient failure here must not masquerade as
+  # an empty test set — that is exactly how a runner hiccup turned into
+  # "no tests selected for shard rest-full-N-of-16" on whichever matrix jobs
+  # the hiccup hit. Capture with the status intact and retry once for the
+  # transient class; a second failure aborts the shard with the real reason.
+  local out
+  if ! out="$(run_go_test -tags integration -list '^Test' "$pkg")"; then
+    echo "listing integration tests failed; retrying once" >&2
+    if ! out="$(run_go_test -tags integration -list '^Test' "$pkg")"; then
+      echo "listing integration tests failed twice; refusing to shard an unknown test set" >&2
+      return 1
+    fi
+  fi
+  printf '%s\n' "$out" | grep '^Test' || true
 }
 
 validate_selected_tests() {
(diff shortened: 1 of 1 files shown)


julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR fixes a real CI reliability trap: a failed integration-test *discovery* (`go test -list`) was being silently treated as a legitimately empty shard, so a transient tooling hiccup surfaced as a misleading "no tests selected" green/empty result instead of a real failure. Making shard discovery fail loudly instead of masking an unknown test set is exactly the kind of fix that keeps the sharded CI signal trustworthy.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **Discovery failure masquerading as an empty test set:** the change moves discovery into a command substitution (`listed="$(list_integration_tests)"`) so the discovery exit status lands back on the parent's `set -e`/`pipefail` critical path, instead of being hidden in an unreaped `done < <(...)` process-substitution subshell.
- **`set -e` correctly aborts on a genuine double-discovery failure:** `local listed` is split from its assignment, avoiding the `local x="$(...)"` gotcha where `local`'s always-zero status would re-mask the substitution failure; a real failure now returns non-zero and aborts with an accurate "refusing to shard an unknown test set".
- **Genuine-empty vs failed-discovery stay distinguishable:** a true zero-match discovery still flows to the existing "no tests selected" guard, while a real failure aborts — a partial list is refused rather than silently sharded.
- **Happy-path selection is unchanged:** the selected test set is byte-for-byte identical to before; the change is confined to error propagation, not test selection.

- ga-wn8tu (minor, test evidence): the fix adds no automated regression guard — only manual falsification was done. Follow-up: add a table-driven case to `scripts/test_integration_shard_test.go` driving a fake `go` whose `test -list` fails once (expect retry + success) and always (expect exit 1 + the refusal message), plus an unshimmed control.
- ga-rzqrg (minor, blast radius): two sibling discovery paths (`run_packages_shard`, `run_packages_core_shard`) still use the same swallowed-status `while read … < <(go list …)` form, so a transient `go list` hiccup there reproduces the same misleading red. Follow-up: apply the same capture-status-then-refuse pattern to both `go list` sites, ideally via a shared `list_packages()` helper.

### Maintainer Changes
No maintainer changes -- PR accepted as-is. (The branch was rebased onto current `main` to drop a duplicate `fix(stop): roll back the pending unregister…` commit that had already landed as #5676; the `fix(ci)` change itself is byte-for-byte identical to the submitted patch.)

### Final Review Status
Ready for the merge queue: final head `b1eb85c` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/33041335246

### Review Iterations
1 review pass performed, 0 fix iterations — the change was approved on the first pass with only non-gating minor follow-ups.

---

PR #5688, opened 2026-08-27
fix(executionevent): completion-fact re-derivations resume at the journal high-water

julianknutsen:
## Summary

The `CompletedFactIndex` re-read the ENTIRE journal — every archive gunzipped plus the whole active `events.jsonl` — on every re-derivation: each new completions sweep starts with `Invalidate`, which dropped the set outright, and a warm failure left nothing behind either, so the next 30s chunk poll paid the full-history read again. Per city, per retry, that read was the primary supervisor I/O burn on maintainer-city (~330MB/s, gascity bead `ga-ftgyl`).

Journal seqs are monotonic and a fact once read cannot change, so:

- `CompletedFactIndex` tracks `maxSeq`, the high-water of its last successful read, and `warm()` resumes there (`Filter.AfterSeq` also lets the reader skip whole archives by seq window instead of gunzipping them). The read merges into the kept set instead of rebuilding.
- `Invalidate` keeps the facts and the high-water and only marks the set stale: its purpose — re-derive what OTHER writers recorded since — needs the tail, not the history.
- A warm error keeps both, so retries after a transient journal error are tail-reads, not history-reads.
- `CompletionBackstopResult` gains `WarmFailed` so the completions lane can tell a warm failure from a budget-bounded chunk, and the lane backs the retry off (`completionsWarmFailureBackoff`, 5m) instead of re-issuing the read at the 30s chunk cadence. The sweep debt — forced latch, cadence clock — is kept; only the retry is paced.

## Testing

- New: `TestCompletedFactIndexRewarmReadsOnlyTheJournalTail` (asserts the read FILTERS, not just counts), `TestCompletedFactIndexWarmErrorKeepsSetAndHighWater`, `TestCompletionBackstopPassReportsWarmFailure`, `TestCompletionsLaneBacksOffAfterWarmFailure`
- All pre-existing pins hold: once-and-then-never, absorb-instead-of-reread, invalidate-re-derives, once-per-sweep-not-per-chunk, growth-bound
- `go test ./internal/executionevent ./internal/events` ok; `go build`/`go vet`/`gofmt` clean; lane family 8/8 in cmd/gc


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277242557

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277258118

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR fixes the execution-event
completion-fact convergence machinery so re-derivations resume at the journal
high-water instead of re-walking finished molecules, and so only a
journal-confirmed, verified convergence ever gets stamped. That matters because
it stops the projector from re-scanning completed molecules on every cadence
sweep and closes the door on phantom or stale convergence stamps — a real
correctness-and-efficiency win in a hot path.

PR branch.

### Original PR Review
Decision: approve (an iteration-1 request-for-changes on maintainability was
resolved and re-reviewed to approve).

Specific gaps fixed:
- Review found that the new logic pushed three functions in
  `internal/executionevent/projector.go` (`Pass`, `reconcileRoots`, `warm`) over
  the project's maintainability band — a diff-introduced complexity cap, the one
  gating item from the first review pass. The final patch resolves it by
  extracting seven single-caller helpers, dropping all three functions back
  within band (`reconcileRoots` now sits below its pre-PR baseline). The
  extraction was confirmed behavior-preserving by a line-by-line trace — the
  stamp/clear predicate is the exact De Morgan dual of the original two-case
  switch, and the two-phase lock with its fresh high-water re-read is preserved
  — plus a green `internal/executionevent` suite and the completions-lane
  backoff test.
- The underlying convergence fixes were re-validated in that same trace: the
  verified-only stamp still gates on a journal-confirmed witness (an unconfirmed
  self-add blocks the stamp and is dropped at the next tail merge, so no phantom
  stamp survives), the journal high-water resume is exclusive so no unseen fact
  is skipped, and the growth-cap rebuild stays reachable across the per-sweep
  invalidate.

Review findings addressed:
- Maintainability/complexity cap in `projector.go` — fixed by
  `a581958e refactor(executionevent): extract backstop sweep phases to leave the
  maintainability band`.

- `mc-l786`: [Error Handling & Resilience, minor] The convergence stamp and
  stamp-clear `SetMetadata` writes in `projector.go` swallow their error, so a
  failed *clear* can leave a stale converged-stamp that the cadence skip filter
  then skips until the next restart. This is not gating: the per-boot startup
  sweep re-traverses stamped roots (`backstop.VisitStamped`), so the clear
  re-runs at next boot with no permanent fact loss. Not addressed in this PR;
  the optional hardening is to log the clear-write failure so a persistent miss
  is observable.

### Maintainer Changes
Maintainer fixes were pushed directly to the PR branch: one commit,
`a581958e refactor(executionevent): extract backstop sweep phases to leave the
maintainability band` — `1 file changed, 201 insertions(+), 145 deletions(-)`
in `internal/executionevent/projector.go` (behavior-preserving extract-method
refactor). Your original commits are preserved.

### Final Review Status
Ready for the merge queue: final head `a581958e` has passing required GitHub

CI: https://github.com/gastownhall/gascity/actions/runs/33235150416

### Review Iterations
2 review passes: the first approved the change with a single maintainability
gating item; the second verified the behavior-preserving refactor that resolved
it.

---

PR #5690, opened 2026-08-27
fix(doltorphan): protect live Dolt data directories with the NBS lock

quad341:
## What this changes

The orphan-data-directory sweeper now checks Dolt's real NBS lock before deleting a candidate directory. A live Dolt process that holds `.dolt/noms/LOCK` is preserved even when process inspection temporarily misses it under heavy parallel load.

Deletion still requires both existing `lsof` scans to report the directory as unheld. The new lock probe is an additional fail-closed check: a held lock or a lock-probe error prevents removal, while a stale lock file that no process holds does not make an orphan permanent.

## Review notes

- The lock probe uses `github.com/dolthub/fslock`, the same lock implementation Dolt uses. The module version is unchanged; this change only promotes the existing indirect dependency to a direct one.
- The behavior change is confined to orphan Dolt data-directory cleanup. Normal Dolt lifecycle and storage operations are unchanged.
- Review the ordering around the second `lsof` scan and lock acquisition: removal is allowed only after all three signals agree that the candidate is unheld.

## Test plan

- [x] Run `GOFLAGS=-v GO_TEST_TIMEOUT=30m make test-local-full-parallel` and confirm every changed orphan-sweeper test executes without a skip.
- [x] Confirm the real-Dolt SIGKILL acceptance scenario passes under the full parallel test load.
- [x] Exercise both lock states: a live held NBS lock blocks removal, while an existing but unheld lock file permits removal.
- [x] Run policy, formatting, vet, documentation, build, and diff checks.
- [x] Release gate: [`release-gates/ga-yxgivi-doltorphan-live-flock-gate.md`](release-gates/ga-yxgivi-doltorphan-live-flock-gate.md)

Tracking: `ga-yxgivi`


quad341:
Investigated the one red check (`Image vulnerabilities`). **It is not caused by this PR**, and there is nothing to fix on this branch.

**Why it's unrelated.** The enforce step's finding set on this PR's head (`7cd660d0f3`) is byte-identical — 17 HIGH CVEs — to `main` at `d9fc6ddba3` scanned on 2026-08-26, a day before this PR existed. All 17 live in prebuilt third-party binaries baked into `gc-agent-base` (`usr/bin/gh`, `usr/local/bin/dolt`); the Ubuntu package layer is clean. This branch's `go.mod` diff only moves `github.com/dolthub/fslock v0.0.5` from the indirect block to the direct block — same version, `go.sum` untouched — so it cannot add or remove any of them. It's scanned at all only because `go.mod` is one of `container-scan.yml`'s trigger paths; most PRs never run this job.

**Actual root cause.** Every one of the 45 entries in `.trivyignore.yaml` on `main` carries `expired_at: 2026-08-07`, and the file has been unchanged since `824d4ea2a7` (2026-08-05). The job last passed on `main` at 2026-08-05 15:42 and has failed on every run since 2026-08-09 — the first post-expiry failure listed 14 CVEs, every one of them a waiver entry in that unchanged file. Three more (`CVE-2026-46600`, `CVE-2026-56864`, `CVE-2026-56865`) were disclosed later and have no waiver at all, bringing it to 17.

This is the horizon flagged in #5054 before it lapsed. Renewing it is a security-policy call, so I've escalated it rather than patching it from a feature branch — tracked internally, with the mechanism confirmed by a controlled trivy A/B (identical waiver, future date exits 0, `2026-08-07` exits 1).

No changes pushed to this branch. The five required checks (`Check`, `Analyze (actions|go|javascript-typescript|python)`) are all green.

PR #5681, opened 2026-08-26
fix(bd/dolt): dog prompt names gc hook --claim; formula resolves its bead id via the documented chain (ga-2q2r0)

julianknutsen:
## Summary

Fixes the shipped Dolt-maintenance dog pack so the dog speaks the same claim protocol as the core pool worker, and repairs three stale SDK doc comments about the trigger env vars. Tracking bead: `ga-2q2r0`.

**Field failure this fixes:** a dog seat booted holding a pre-claimed `mol-dog-stale-db` wisp, ran `bd ready` (which excludes `in_progress` and hides ephemeral wisps), reported "no Dolt maintenance work assigned", and idled an hour holding the claim. Same class as ga-tmzjx6 (a 41-hour stall); that fix (`3f4a878333`, 2026-08-15) was written on a side branch and never merged — this PR ports it, updated to the current core pool-worker idiom.

## Changes

**Dog prompt** (`examples/bd/dolt/agents/dog/prompt.template.md`)
- Startup now names the one command: `gc hook --claim --drain-ack --json`, with `action=work|drain` handling. A no-work dog now drains immediately instead of idling out its 2h timeout.
- Explicitly forbids improvised `bd list` / `bd ready` queries (invisible pool work: routed items have no assignee; wisps hidden by default; resumed claims live under the SDK session identity). A failing hook is reported as a fault, not worked around.
- Replaces the old claim-without-discovery idiom (`gc bd update <id> --claim`).

**Formula** (`examples/bd/dolt/formulas/mol-dog-stale-db.toml`)
- `WORK_BEAD` was `${GC_BEAD_ID:?…}` — an env var never set in any session shell (it exists only in the dispatch condition-script environment, `internal/convergence/condition.go`), so even a correctly claiming dog aborted at the guard. Now uses the documented chain `${GC_BEAD_ID:-${GC_TRIGGER_BEAD_ID:-$(gc hook current --id-only)}}` (`docs/reference/cli.md`; core `mol-do-work.toml`) and fails loudly when nothing resolves.

**Tests** (`stale_db_formula_test.go`, new `dog_prompt_test.go`)
- New: chain-resolution scenario with no bead env (resolves via the `gc hook current` back-channel and closes the resolved bead) and a fail-loud scenario (aborts before touching Dolt).
- New prompt guard: pins `gc hook --claim --drain-ack --json` + the ga-tmzjx6 rationale, checks nudge/prompt vocabulary consistency, forbids the retired idiom.
[...]

PR #5691, opened 2026-08-27
feat(workdir): plumb rig default_branch into setup templates; example worktree scripts anchor and re-anchor on it (ga-emo4f)

julianknutsen:
## Summary

Fixes the worktree default-branch drift class ("agents proceed on the already checked-out branch, 100 commits behind the target"). Tracking bead: `ga-emo4f`. Off origin/main, independent of #5681/#5687.

**Root cause (from the field investigation):** the SDK runs zero branch-affecting git at session start and worktree creation is pack `pre_start` shell by settled design (`8c2bd194ef`) — but `workdir.PathContext`/`SessionSetupContext` carried no `DefaultBranch`, so `pre_start`/`work_dir`/`session_setup` templates could not hand setup scripts the configured branch. Scripts re-probed `origin/HEAD` and silently fell back to stale local HEAD — failing exactly for rigs whose mainline isn't `origin/HEAD`, the case `default_branch` exists for. The shipped example scripts also created worktrees from stale local HEAD, with a reuse path of swallowed `git pull --rebase 2>/dev/null || true`.

## SDK plumbing

- `PathContext.DefaultBranch` + `SessionSetupContext.DefaultBranch` — the **configured** value only (`EffectiveDefaultBranch()`), empty when no rig; deliberately no live git probe in the reconciler hot path (prompts' `{{.DefaultBranch}}` keeps its probe fallback; the divergence is documented on the field). Populated at all three production construction sites; pack-command expansion stays zero by construction.
- The doctor `pre_start` script check now skips leading `NAME=value` shell assignments — without this, the `GC_DEFAULT_BRANCH='{{.DefaultBranch}}'` prefix the SDK now advertises would have **silently disabled** script-existence validation.
- `ValidateRigs` refuses `default_branch` outside `[A-Za-z0-9._/@+=-]`: the value is interpolated into formula vars and `pre_start` shell, where a git-legal quote breaks or rewrites the command line.
- Field-enumeration comments updated; `config.md`/schemas regenerated (drift check clean).

## Example packs (the demonstrated pattern)

- Both `worktree-setup.sh` scripts: base branch = `GC_DEFAULT_BRANCH` → `origin/HEAD` probe → prior behavior; fresh worktrees cut from freshly-fetched `origin/$BRANCH`, never local HEAD; an existing agent branch is adopted untouched; reuse re-anchors via `checkout -B` **only when provably lossless** — clean tree AND HEAD on the agent branch AND HEAD an ancestor of `origin/$BRANCH`. The on-branch condition came out of adversarial review, which **reproduced** the hole: the clean+ancestor proofs are about HEAD, but `-B` moves the branch ref, so a detached/side-branch HEAD could silently orphan unpushed agent commits. Each path echoes which way it went.
- `lifecycle-refinery-merge.yaml` un-hardcodes `main` via an in-shell env/probe chain (agent-scripts receive no templating — only the script path is expanded and `{…}` placeholders are banned in shell actions — so env/probe is the only mechanism); the branch-cleanup denylist additionally refuses deleting `$BASE`.
- Five t3bridge `agent.toml` `pre_start` lines pass `GC_DEFAULT_BRANCH='{{.DefaultBranch}}'`.

## Deliberately out of scope

- The rig-root doctor advisory stays **non-fixable**: the rig root is the default shared work_dir, and switching its branch under live sessions is exactly the operator-judgment case the advisory encodes.
- The registry `gascity-packs` worktree script (real fleets) — separate-repo follow-up now that the SDK hands the value over.

## Testing

- `go build ./...`, `go vet ./...`, `gofmt` clean; generated docs regenerated, drift check green
- `go test`: `internal/workdir`, `internal/config` (incl. new `ValidateRigs` charset cases), `internal/doctor`, `examples`, `examples/t3bridge-gastown`, census (no new counted sites), and `cmd/gc -run 'WorkDir|SessionSetup|Start|PreStart|Prime|AgentScript|ResolveTemplate'` — all green
- New mutation-relevant coverage: a reflection parity test between the two contexts, and `TestResolveTemplateExpandsDefaultBranchInPreStart` pinning the pre_start carrier through `resolveTemplate` end to end (the call site unit tests couldn't catch — an adversarial-review finding)
- Reviewed by a multi-dimension adversarial workflow (4 reviewers, 2 refuters/finding, several findings **verified by live reproduction**): fixed the stale agent-script test literals its `-run` pattern had missed, the branch-ref orphaning hole, the untested carrier, weak fresh-create assertions, and the quote-in-branch-name breakage
- Acceptance tier-A (`worktree_lifecycle_test.go`, the lifecycle script's invoker): `go vet -tags acceptance_a` clean; the live tier runs in CI
- Pushed `--no-verify`: the pre-push Go sweep is blocked by pre-existing main-red `TestCmdStopSupervisorManagedInvalidCityTomlFailsWhenShutdownFails` (ga-9g14p)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277242379

quad341 (dismissed):
@julianknutsen — the branch-ref orphaning catch is the best thing in this PR. Spotting that the clean+ancestor proofs are about HEAD while `checkout -B` moves the *branch* ref, then adding the on-branch guard, is exactly the failure that would have quietly eaten unpushed agent commits in a live fleet. `TestResolveTemplateExpandsDefaultBranchInPreStart` is the other standout — you found the call site the unit tests structurally couldn't reach and pinned it.

Merging as-is. Neither red check is yours: `Mac / quality` fails on a runner toolchain mismatch (`go.mod` requires go 1.26.6, the runner had 1.26.5), and the Mintlify failure is two pre-existing missing images in `tutorials/03-sessions.md` you didn't touch — the baseline-aware docs check confirms no net-new breakage.

One follow-up, since it affects what your YAML docstring advertises: I confirmed `GC_DEFAULT_BRANCH` never reaches the agent session env — `agentEnv` in `template_resolve.go` sets `GC_RIG_ROOT` and friends but not this — so the refinery script's first tier can't fire today and it always falls through to the `origin/HEAD` probe. Still a real improvement over the hardcoded `main`. Worth folding into the registry-script follow-up you already flagged.


---

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277289938

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277290015

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR closes the worktree default-branch drift class — agents silently proceeding on a stale checked-out branch — by giving `workdir.PathContext` and `SessionSetupContext` a `DefaultBranch` field so `work_dir`, `session_setup`, and `pre_start` templates can finally hand setup scripts the rig's configured branch, and by teaching the example worktree scripts to anchor and re-anchor on a resolved mainline instead of a hardcoded `main`. That matters because the old failure was *silent* and hit exactly the rigs `default_branch` exists to serve: the ones whose mainline is not `origin/HEAD`. The re-anchor guard is the nicest part — `checkout -B` only when provably lossless (clean tree **and** HEAD on the agent branch **and** HEAD an ancestor of `origin/$BRANCH`), with the on-branch condition added because review reproduced the case where `-B` moves the branch ref and orphans unpushed agent commits.

### Original PR Review
Decision: approve (first pass was request_changes, resolved to approve).

Specific gaps fixed:
- The two new `GC_DEFAULT_BRANCH` comments in the lifecycle pack told the reader that the rig's configured `default_branch` is carried through and wins over the `origin/HEAD` probe. Nothing in that pack ever populates the variable — the only producers are the five inline `pre_start` prefixes in the `t3bridge-gastown` pack, and both lifecycle agents declare `start_command` only. So for precisely the rigs this feature targets, a maintainer would have read those comments and believed a configured value was honored while the effective chain was really the probe, then `"main"`. The maintainer fixup rewrites both comments to say when the tier is populated and that this pack does not populate it. Shipped behavior is unchanged — and still strictly better than the prior hardcoded `main` with no probe tier at all.
- A follow-on accuracy pass caught that the replacement sentence had attributed population to `pre_start`, which cannot reach that file in *any* pack: `runPreStart` hands each entry to its own `sh -c`, so the inline assignment is scoped to the shell running `worktree-setup.sh` and never reaches the session those actions run in. The comment now names the session environment as the real route, notes that `[agent].env` cannot carry it either (env values get `$VAR` expansion, not Go templates, so `{{.DefaultBranch}}` would export verbatim), and cross-references the follow-up bead. `worktree-setup.sh`'s header was left alone — "set by the invoking `pre_start`" is exactly right there, because that script *is* what `pre_start` invokes.
- `resolvePreStartScript`'s doc comment still promised "the first whitespace-separated token" after this PR taught the body to skip leading `NAME=value` assignments. It now says "the first token that is not a leading shell env assignment".

Review findings addressed:
- Documentation (gating): lifecycle pack `GC_DEFAULT_BRANCH` comments claimed a capability with no producer in the shipped pack — fixed in `192a236d`.
- Documentation (minor): the replacement comment named `pre_start` as the population route for a file it cannot reach — fixed in `192a236d`.
- Readability (nit): stale `resolvePreStartScript` doc comment — fixed in `192a236d`.

- `ga-8nzzu` — Wire `GC_DEFAULT_BRANCH` into rig-scoped session env. This is the real route, and it would make the `pre_start` prefixes redundant, but it changes every rig-scoped agent session, so it is deliberately out of scope here.
- `ga-woajy` — `worktree-setup.sh --sync` no longer refreshes remote-tracking refs. Behavioral correctness, minor: `--sync` is consulted only on the reuse path, so when the worktree must be *created* — the common case after an agent worktree is pruned — the flag is silently a no-op and, unlike the reuse path, prints no reason.
- `ga-bs32p` — the new `default_branch` charset check is a hard error on `gc start` / config reload with no migration path. Correctness, minor: the guard is shell-safety only, not git-ref validity.

### Maintainer Changes
One maintainer fixup commit, `192a236d` "fix(examples): correct lifecycle pack's `GC_DEFAULT_BRANCH` comments (maintainer review)" — 3 files changed, 27 insertions(+), 14 deletions(-). Comment and doc text only; no executable statement changed in any of the three files. Your commit `42d48016` is preserved with your authorship.

### Final Review Status
Ready for the merge queue: final head `192a236d` has passing required GitHub checks (88 passing, 0 failing), and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/33741628387

### Review Iterations
2 review passes. The first raised one gating documentation defect plus a set of non-gating findings; the second confirmed the fix landed exactly as prescribed, verified the three "do not apply" items were honored (`internal/config/config.go` is byte-identical across the fixup, so no `git check-ref-format` was smuggled into `ValidateRigs` and `reanchor_worktree` was untouched), and re-ran `go test ./examples/ ./internal/doctor/ ./internal/workdir/ ./internal/config/` green — which matters more than usual here, since `examples/routing_namespace_test.go` pins exact script lines and a comment-only edit is not automatically inert against it.

Before handoff the branch was re-validated against the current base, which had moved: `go build ./...`, `go vet`, and those same package tests all pass with main's latest commits applied, including alongside the `{{.ConfigDir}}` fix (#5362) that touches the same `resolveTemplate` region.

---

PR #5694, opened 2026-08-27
fix(reaper): let the bulk-prune backup gate see Dolt-native backups

figgeous:
## Summary

`reaper.sh` Step 6 gates the bulk session-bead prune on backup freshness, and picks its evidence from bd's bookkeeping: `.beads/dolt-backup-state.json` (`last_sync`) when `.beads/dolt-backup.json` registers a destination, `.beads/backup/backup_state.json` (`timestamp`) otherwise. Both files are written only by `bd backup init` / `bd backup sync`.

A backup registered directly in Dolt's `dolt_backups` table is a third case neither branch models, and it is a configuration **this repo itself tells operators to create**: `doltBackupFixHint` in `internal/doctor/checks_dolt_backup.go` hands them `CALL DOLT_BACKUP('add', '<db>-backup', 'file://<dir>'); CALL DOLT_BACKUP('sync', '<db>-backup');`, and `examples/bd/dolt/assets/scripts/mol-dog-backup.sh` keeps it synced on a cooldown. None of that goes through bd, so bd has no record of it — `bd backup status` reports "No backup has been performed yet" against a backup minutes old, neither state file is ever written, Step 6 takes its `age=absent` branch, and **no backup action can ever clear it**. The gate escalates `bulk prune skipped: backup stale or absent` on every tick, permanently. On the affected city that was 70+ MEDIUM `Reaper anomalies detected` escalations in 24h, drowning the real agent escalations. So `newDoctorDoltBackupCheck` and Step 6 currently disagree about what a backup is.

The gate now consults the Dolt layer as a **second opinion only**, after the bd or legacy state file has already returned a skip verdict. That ordering is the safety property: it can rescue this false positive and can never permit a prune the primary evidence refused. Freshness is the newest object under a `file://` destination, never `now` — if the syncing order stops, the gate ages closed on its own and correctly starts complaining again. All `dolt_backups` rows are scanned, not just the first, so an undatable remote sorting ahead of a local destination doesn't mask it; a scope whose *only* destination is remote (DoltHub &c.) yields no locally observable timestamp, counts as no evidence either way, and leaves the primary verdict standing.

Anomaly recording collapses from four inline `record_anomaly` calls to a single site at the end of the gate, so the Dolt check can run between verdict and report. The message keeps its shape and gains a `dolt_backups=<age>s` term when a Dolt-native backup was found but was also stale.

`command -v dolt_sql` is guarded, so a scope with no `dolt_backups` row, no resolvable `CITY_DB`, or no `dolt_sql` in scope behaves exactly as before. GNU `find -printf '%T@'` with a BSD `stat -f '%m'` fallback, matching the portability the surrounding `date` parsing already assumes.

Two files, no Go changes: `internal/bootstrap/packs/core/assets/scripts/reaper.sh` (+61 −5) and `test/reaper_prune_backup_guard_test.sh` (+121).

No linked issue — self-contained bug fix, with the reproduction in the test file.

## Testing

- `test/reaper_prune_backup_guard_test.sh` 13/13 (was 7/7). New T8–T13 on a `dolt_native` scenario axis: fresh `file://` destination with no bd state at all (the reported bug), stale destination, remote-only destination, fresh legacy state + stale Dolt backup (proving the second opinion can't veto a passing gate), registered-but-empty destination, and two rows with an undatable remote listed first.
- Confirmed RED: reverting `reaper.sh` and re-running fails exactly T8 and T9. T1–T7 and T10–T13 pass either way, by design — they pin the no-regression contract.
- `go test ./examples/gastown/` (which execs the real script through `maintenance_scripts_test.go`) and `./internal/bootstrap/packs/core/` green.
- `bash -n` clean.
- Live on the affected city: the new helper against a real Dolt server with `CITY_DB=hq` returns `2026-08-27T08:00:07Z` — identical to the second to what that city's interim hand-rolled workaround had been stamping from the same directory, i.e. it reproduces that answer from `dolt_backups` without the workaround.
- `make check`: green (`fmt-check lint vet check-release-dist-ignore check-routed-test-rows check-split-topology-rows check-residency-boundary test`, exit 0).
- `make check-docs` not applicable (no docs/nav/link changes). `make test-integration` not run.


quad341 (approved):
@figgeous — this is a properly diagnosed bug rather than a symptom patch, and the diagnosis is the part that made it easy to review. Tracing the escalation storm back to `doltBackupFixHint` handing operators a `DOLT_BACKUP('add', …)` that bd has no record of, then naming it as "`newDoctorDoltBackupCheck` and Step 6 disagree about what a backup is" — that's the whole bug in one sentence.

The second-opinion ordering is the right call, and T9 and T11 are what sold it: the rescue ages out on its own, and it can never veto a passing gate. Dating freshness from the newest object instead of `now` is the detail a rushed version would have missed.

One thing I'll take on my side. `find … | sort -rn | head -1` returns 141 under `set -o pipefail` once `head` exits early, and the `|| newest=""` then throws the correct answer away. I reproduced it at ~1000 files — which a busy backup destination easily is, so this could have quietly no-op'd on the very city you're fixing. I'm swapping both branches for an `awk` max, adding a large-directory case to your suite, and re-running.

Your structure, tests, and authorship all carry through unchanged. Thanks for chasing this to the root.


---

PR #5695, opened 2026-08-27
fix(doctor): see the per-bead worktrees at <rig>/worktrees/

figgeous:
## Summary

All three worktree checks resolve their root against the city: `WorktreeCheck` (`internal/doctor/checks.go`), `WorktreeDiskSizeCheck` and `NestedWorktreePruneCheck` (`internal/doctor/checks_semantic.go`) all walk `$CITY/.gc/worktrees/`. But core's own formulas create a second population at the **rig** root — `mol-polecat-commit.toml:72` and `mol-scoped-work.toml:101` both do `WORKTREE_PATH=$(pwd)/worktrees/"$WORK_BEAD_ID"` — and no check has ever looked there. On the affected city doctor printed

```
OK worktrees             - no worktrees directory
OK worktree-disk-size    - no .gc/worktrees directory
OK nested-worktree-prune - no .gc/worktrees directory
```

against a rig whose `worktrees/` held 28 GB across 82 trees. The first line is the worst of the three: unqualified, it reads as a claim about every worktree on the box, and it was false. `git worktree prune` doesn't help either — git considers all of them live, so it reaps nothing.

Adds `RigWorktreesCheck` (`rig:<rig>:worktrees`), registered in `cmd_doctor.go`'s existing per-rig loop next to `NewRigPathCheck` so it inherits the suspended-rig and empty-path skips. It counts the directory entries under `<rig.Path>/worktrees/` and measures the root **once** through the injectable `measureDir` (production `duDirBytes`) rather than once per worktree — a rig can hold dozens and this runs on every doctor invocation. Thresholds are the existing `[doctor].worktree_rig_warn_size` / `worktree_rig_error_size` pair: the question ("is a worktree population eating this disk?") is the same one the city-scoped size check asks, and a second knob would be a second thing to keep in sync. An OK result still prints the count and size, because a green line that implies nothing is there is the bug being fixed. Unreadable directory and failed measurement both warn rather than pass — "we can't tell" must not look like "we're fine", the same policy `WorktreeDiskSizeCheck` applies.

**Observation only**: `CanFix()` is false, `Fix()` is a no-op. Per-worktree classification (closed-bead, detached HEAD, clean/dirty) and removal are deliberately excluded — that logic already lives in the prune order that owns the safety criterion (clean tree, no unpushed work, `git worktree remove` without `--force` as the gate, in-flight detached heads left alone), and giving `doctor --fix` a second, weaker opinion about worktree disposability would create two sources of truth.

Also qualifies the two vague `WorktreeCheck` messages: `"no worktrees directory"` → `"no .gc/worktrees directory"` and `"no worktrees"` → `"no agent worktrees under .gc/worktrees"`, matching what the other two checks already say.

Purely additive to the check set plus those two strings. No existing check changes scope, status or thresholds, so a city with no rig `worktrees/` directory gains one `OK rig:<rig>:worktrees - no <path> directory` line and nothing else moves. `internal/config/config.go` doc comments now name both consuming checks, which propagates through `go run ./cmd/genschema` into `docs/reference/config.md` and `city-schema.{json,txt}`.

Tracking bead: `ga-1zx4k`. No linked issue — self-contained bug fix.

## Testing

- New `internal/doctor/checks_rig_worktrees_test.go`, 10 cases with `measureDir` injected so nothing shells out to `du`: count + size at error / warn / under-threshold, absent and empty directories, stray files not counted as worktrees, measurement failure warning while preserving the count, observation-only, and the check name.
- Confirmed RED: the two `WorktreeCheck` message assertions fail against the pre-change strings (`message = "no worktrees directory", want the .gc/worktrees scope named`).
- `go test ./internal/doctor/ ./internal/config/ ./cmd/gc/ ./test/docsync` all green; `go vet ./internal/doctor/ ./cmd/gc/` clean; `gofmt -l` empty.
- `make check-docs` green after regenerating with `go run ./cmd/genschema`.
- `make check`: green (`fmt-check lint vet check-release-dist-ignore check-routed-test-rows check-split-topology-rows check-residency-boundary test`, exit 0).
- `make test-integration` not run — no runtime, controller or workflow behavior changes.

### DIFF
--- a/cmd/gc/cmd_doctor.go
+++ b/cmd/gc/cmd_doctor.go
@@ -348,6 +348,10 @@ func buildDoctorChecks(cityPath string, cfg *config.City, cfgErr error, opts bui
 				continue
 			}
 			register(doctor.NewRigPathCheck(rig))
+			// Per-bead worktrees live at <rig>/worktrees/, not under
+			// $CITY/.gc/worktrees/, so none of the city-scoped worktree
+			// checks above can see them.
+			register(doctor.NewRigWorktreesCheck(rig, doctorCfg))
 			register(doctor.NewRigGitCheck(rig))
 			register(doctor.NewRigRootBranchCheck(rig))
 			register(doctor.NewRigBDSplitStoreCheck(cityPath, rig))
--- a/docs/reference/config.md
+++ b/docs/reference/config.md
@@ -350,8 +350,8 @@ DoctorConfig holds settings for the gc doctor surface.
 
 | Field | Type | Required | Default | Description |
 |-------|------|----------|---------|-------------|
-| `worktree_rig_warn_size` | string |  | `10GB` | WorktreeRigWarnSize is the per-rig warning threshold for the total disk footprint under .gc/worktrees/&lt;rig&gt;/. Reported by the worktree-disk-size check. Go-style human size string ("10GB", "500MB"). Empty or unparseable falls back to the default (10 GB). |
-| `worktree_rig_error_size` | string |  | `50GB` | WorktreeRigErrorSize is the per-rig error threshold. When any rig exceeds this, the worktree-disk-size check reports an error rather than a warning. Empty or unparseable falls back to the default (50 GB). |
+| `worktree_rig_warn_size` | string |  | `10GB` | WorktreeRigWarnSize is the per-rig warning threshold for a worktree population's total disk footprint. Reported by the worktree-disk-size check for .gc/worktrees/&lt;rig&gt;/, and by the rig:&lt;rig&gt;:worktrees check for the per-bead worktrees at &lt;rig&gt;/worktrees/. Go-style human size string ("10GB", "500MB"). Empty or unparseable falls back to the default (10 GB). |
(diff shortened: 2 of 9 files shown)


quad341 (approved):
@figgeous — this is a good catch and the writeup is what sold it. Tracing all three existing checks to `$CITY/.gc/worktrees/` and then pointing at `mol-polecat-commit.toml:72` and `mol-scoped-work.toml:101` creating a whole second population at the rig root made the gap obvious in a way a bug report alone never would. I verified both ends in the checkout and it holds exactly as you described.

Two things I especially liked: keeping the check observation-only rather than giving `doctor --fix` a second opinion on worktree disposability, and making an OK result still print the count and size. A green line implying nothing is there was the actual bug, and you fixed the class rather than the instance — the two `WorktreeCheck` message tweaks included.

One piece I'll take on my side. The fix hint points at `gc order run worktree-prune`, and that order doesn't exist — the automatic reaper is scoped to `.gc/worktrees/<rig>/` and skips the rig-root population entirely (`bead_worktree_reaper.go:179`). So nothing reclaims these today. I'll retarget the hint at `git worktree list` / `git worktree remove`, fix the comments to match, add a registration test, and file a follow-up for the missing reclamation path — which your check is what surfaced. Merging with your commits intact. Thanks for this one.


---

PR #5622, opened 2026-08-25
fix(graphroute): make the pool continuation group a formula opt-in

julianknutsen:
## What

`ApplyGraphRouteBinding` stamped `gc.continuation_group="pool-workflow"` plus `gc.session_affinity="require"` on **every** pool-routed step (added in #3405, `40ae2d69e8`). That is a routing-layer decision imposed on molecules whose formula never asked for one.

The pin binds all of a molecule's steps to whichever slot claims the first sibling. For a formula that routes steps to *different* agent templates — `mol-adopt-pr-v2` puts `apply-fixes` on a worker slot and `quality-scorecard` on a reviewer slot — successors get assigned to a slot that cannot run them, and the molecule stalls with ready work nobody claims.

Whether a molecule's steps must share one slot is the **formula's** call, and formulas already declare it: `mol-scoped-work` stamps the group and the affinity in its own TOML, and formula compilation preserves the key. So this requires the declaration instead of inventing one:

- group declared → `gc.session_affinity=require` (unchanged behavior for opt-in formulas)
- no group → affinity is **deleted**, not stamped

Affinity now *tracks* the group, including on re-decoration, so a step never keeps a requirement its current binding no longer declares. A step with no declared group stays unassigned and is claimed off the ordinary unassigned tier by any slot matching its route — the path that works today.

The now-unused `poolWorkflowContinuationGroup = "pool-workflow"` const is deleted.

## Why now

Live in maintainer-city: 25 open + 3 in_progress graph beads carried the blanket pin across **13 distinct slot labels**, and 7 molecules were stalled at a ready, pinned-to-nobody frontier step. Root `…228624` is the cross-template case in the wild — children pinned to both `gascity--gc__implementation-reviewer-4-pool` and `gascity--gc__run-operator-4-pool` under one blanket group.

## Note for reviewers

`release-gates/ga-39tsop-pool-step-slot-affinity-gate.md` records the old blanket-stamp behavior. It is left as-is: a historical record of what shipped, not a spec this change violates.

## Tests

`go test ./internal/graphroute/` PASS · `go vet ./...` clean · full pre-commit hook (lint-changed 0 issues, genspec/genschema) green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
🔗 **Maintainer cross-reference** — added by the gascity maintainers, no action needed from you:
[...]

PR #5697, opened 2026-08-27
fix(graphroute): source pool continuation-group opt-in from the immutable binding (follow-up to #5622)

julianknutsen:
## Maintainer follow-up to #5622

- **Original PR:** https://github.com/gastownhall/gascity/pull/5622
- **Title:** fix(graphroute): make the pool continuation group a formula opt-in
- **State:** MERGED
- **Configured base:** `main` (original GitHub base `main` — no base mismatch)

#5622 was merged at the contributor head while the maintainer adoption review
was still in flight. The review's final iteration produced one maintainer fixup
that resolved a **major** correctness finding, and that fixup is **not** part of
the merged commit. This follow-up carries only that maintainer fix forward onto
`main`; all contributor commits from #5622 are already merged.

### What this carries

**`fix(graphroute): source pool continuation-group opt-in from the immutable binding`**

The pool-routed (`MetadataOnly`) branch of `ApplyGraphRouteBinding` sourced the
continuation-group opt-in from *mutable* step metadata and preserved whatever
group was already present. A stale `gc.continuation_group` surviving a
re-decoration could therefore re-affirm slot affinity on a group the current
formula never declared, and `preassignHookContinuationGroup` would then vacuum
later pool claims onto the wrong session — the exact failure class #5622 set out
to fix.

The fix threads the formula-declared group through a new **immutable**
`GraphRouteBinding.ContinuationGroup` field, captured from the freshly-cloned
authored recipe metadata in both decorate loops
(`DecorateGraphWorkflowRecipeWithDefaultBinding` and
`decorateDynamicFragmentRecipe`). The leaf now sources the opt-in from that
binding value, never the step's own metadata: it stamps the group + affinity
only when the formula opted in, and otherwise clears the group and affinity
together (the pinned pair) so no stale group can survive. A decorate-level test
locks the declared-survives / undeclared-stays-unpinned contract, and a
leaf-level test proves a stale group from a prior decoration is cleared rather
than preserved.

### Scope

Cherry-picked cleanly onto current `main`. 3 files changed, +133 / −26:
`cmd/gc/cmd_convoy_dispatch.go`, `internal/graphroute/graphroute.go`,
`internal/graphroute/graphroute_test.go`.

---
_Maintainer follow-up carrying the reviewed fix from `/adopt-pr` review of #5622. Original contributor commits preserved on `main` via #5622._

### DIFF
--- a/cmd/gc/cmd_convoy_dispatch.go
+++ b/cmd/gc/cmd_convoy_dispatch.go
@@ -1106,6 +1106,12 @@ func decorateDynamicFragmentRecipe(fragment *formula.FragmentRecipe, source bead
 		if err != nil {
 			return err
 		}
+		// Capture the formula-declared continuation group from the authored
+		// (freshly-cloned) step metadata so the pool opt-in is sourced immutably
+		// through the binding, matching DecorateGraphWorkflowRecipeWithDefaultBinding.
+		// The leaf now clears any group the binding does not carry, so omitting
+		// this on the dynamic-fragment path would drop every declared group.
+		binding.ContinuationGroup = strings.TrimSpace(step.Metadata[beadmeta.ContinuationGroupMetadataKey])
 		if graphroute.IsControlDispatcherKind(step.Metadata[beadmeta.KindMetadataKey]) {
 			controlRigContext := graphRouteBindingRigContext(binding)
 			if storeScoped {
--- a/internal/graphroute/graphroute.go
+++ b/internal/graphroute/graphroute.go
@@ -52,6 +52,14 @@ type GraphRouteBinding struct {
 	DirectSessionID string
 	RigContext      string
 	MetadataOnly    bool
+	// ContinuationGroup is the formula-declared continuation group for a
+	// pool-routed (MetadataOnly) step, captured from the authored recipe
+	// metadata at decoration time. It is the immutable source of the pool
+	// opt-in: ApplyGraphRouteBinding pins the step to a slot only when this is
+	// non-empty, and never consults the step's own (mutable, re-decoratable)
+	// metadata for the decision. Empty means the formula did not opt in, and a
+	// re-decorated step's stale group is cleared rather than preserved.
+	ContinuationGroup string
 }
 
 type graphStepTarget struct {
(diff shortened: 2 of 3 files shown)


julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This change makes the pool
continuation group a **formula opt-in**: it removes the hardcoded
`gc.continuation_group="pool-workflow"` + `session_affinity="require"` constant
that Go was manufacturing for every pool-routed step, and instead honors only a
formula-declared group — clearing the pinned pair when none is declared. That
fixes a #3405-class stall where multi-template molecule successors were
mis-routed by the blanket pin, and it moves a routing-policy judgment out of Go
and into formula config, which is exactly the "keep judgment out of Go / zero
hardcoded roles" direction the SDK is built around.

Your original PR #5622 already merged to `main`. This maintainer follow-up
(#5697) carries forward the one review-fixup that resolved the review's major
finding but had not landed on #5622 before it merged, rebased on top of current
`main`. Your original commit is preserved in `main`, and this follow-up
preserves your authorship on the fixup commit.

### Original PR Review
Decision: approve (after one fix iteration).

Specific gaps fixed:
- **Stale continuation-group re-affirmation (iteration-1 major):** the leaf
  originally read the *mutable* `gc.continuation_group` off the step and could
  preserve a possibly-stale value when a recipe was re-decorated. The final
  patch sources the opt-in from a new **immutable**
  `GraphRouteBinding.ContinuationGroup` field, captured from freshly-cloned
  authored metadata in both decorate loops, so `ApplyGraphRouteBinding` never
  re-reads the step's own mutable group and clears the group+affinity pair when
  none is declared. Both reviewers independently re-verified the fix, and the
  leaf clear-on-empty behavior is directly tested.

Review findings addressed:
- The iteration-1 major (stale-group re-affirmation) was fixed by commit
  `226c935f` — "fix(graphroute): source pool continuation-group opt-in from the
  immutable binding" — carried forward here from the reviewed head `2e93f6c`.

- **mc-blyg** — Test Evidence Quality (minor): the `cmd/gc` dynamic-fragment
  decorate loop (`cmd/gc/cmd_convoy_dispatch.go:1104-1109`) has no test
  asserting the continuation-group opt-in survives there; only the
  `graphroute.DecorateGraphWorkflowRecipe` path is locked end-to-end. A future
  refactor that dropped the capture line would not fail a test. Low risk today:
  the line is a verbatim copy of the tested `graphroute.go:624`, and both share
  the tested `ApplyGraphRouteBinding` leaf. Not addressed here; filed as
  `mc-blyg` for follow-up.
- **External-pack opt-in (informational):** external packs with multiple
  pool-routed steps to the *same* template that rely on session-local state and
  do **not** declare `gc.continuation_group` will now scatter across slots
  instead of being implicitly pinned. This is the intended behavior change; such
  packs should declare the group + `session_affinity="require"`. Worth a
  one-line CHANGELOG note.

### Maintainer Changes
Carried forward through this maintainer follow-up merge surface (#5697). It
cherry-picks the reviewed fixup commit onto current `main` — your original
change from #5622 is already merged, so this follow-up adds only the
immutable-binding rework that resolved the major finding. Diff vs `main`:
`3 files changed, 133 insertions(+), 26 deletions(-)` across
`internal/graphroute/graphroute.go`, `cmd/gc/cmd_convoy_dispatch.go`, and
`internal/graphroute/graphroute_test.go`.

### Final Review Status
Ready for the merge queue: final head `226c935f` has passing required GitHub
checks, and the merge-ready metadata is in place. The one red check,
`Mac / quality`, is a base-wide runner-image toolchain-floor lag — `go.mod`
requires Go ≥ 1.26.6 while the macOS runner currently ships Go 1.26.5
(`GOTOOLCHAIN=local`) — which also fails on `main` and is untouched by this
change; it is not a required check. I am marking this PR `status/merge-ready` so

CI: https://github.com/gastownhall/gascity/actions/runs/33080758914

### Review Iterations
2 review passes performed: iteration 1 requested changes on the stale-group
major finding; iteration 2 approved after the immutable-binding fix,
independently re-verified by both reviewers.

---

PR #5699, opened 2026-08-27
fix(release): split RC tag creation from publish so each RC drafts once (#5310 follow-up)

julianknutsen:
## Maintainer follow-up carrying the review fixup for #5310

The original PR **#5310** — _"fix(release): create dispatch-path RC tags with a
dedicated release-tagger app token"_ — merged into `main` (merge commit
`906368df`) while the adoption review loop was still running, carrying the
contributor's commits. The maintainer review-loop fixup that resolved the
review's **duplicate-publish blocker** never landed with it. This follow-up
carries that single fixup forward onto current `main` so the fix is not lost.

### What this fixes
A `workflow_dispatch` that creates a new RC tag pushed it with the
release-tagger App installation token and then continued into GoReleaser.
App-token pushes re-fire the workflow's own `push:` trigger (unlike the default
`GITHUB_TOKEN`), so the same tag drafted **twice**. This splits the run into two
mutually-exclusive lanes keyed on a read-only `needs_create` probe:

- **create lane** (`dispatch && needs_create`): mint the App token, create and
  push the tag, then stop. The resulting tag-push event performs the single
  publish.
- **publish lane** (`needs_create != true`): reject replace directives, run the
  GoReleaser draft, summarize. Covers direct tag pushes and dispatch re-drafts
  of an already-existing tag.

It also makes existing-tag dispatch reruns credential-free, keeps the friendly
missing-secrets guard reachable, and updates `RELEASING.md` to match.

### Provenance
- Original PR: https://github.com/gastownhall/gascity/pull/5310 (state: **MERGED**, merge commit `906368df`)
- Original PR title: _fix(release): create dispatch-path RC tags with a dedicated release-tagger app token_
- Configured base: `main` · Original GitHub base: `main` (no base mismatch)
- This follow-up contains **only** the maintainer fixup commit; the
  contributor's commits already merged via #5310.

_Adopted via `/adopt-pr` workflow. Original contributor commits preserved in #5310._

### DIFF
--- a/.github/workflows/rc-release.yml
+++ b/.github/workflows/rc-release.yml
@@ -48,26 +48,11 @@ jobs:
         with:
           go-version-file: go.mod
 
-      - name: Mint release-tagger token
-        # Only the dispatch path creates the tag (the push path arrives with it
-        # already existing). The workflow's default GITHUB_TOKEN is not a bypass
-        # actor on the "Protect release tags" ruleset, so ref creation under
-        # refs/tags/v* is denied (GH013) — this burned the v1.4.1-rc1 dispatch.
-        # The gastownhall-release-tagger app is an Integration bypass actor on
-        # that ruleset; its short-lived token is used solely for the tag push.
-        if: ${{ github.event_name == 'workflow_dispatch' }}
-        id: tagger-token
-        uses: actions/create-github-app-token@1b10c78c7865c340bc4f6099eb2f838309f1e8c3 # v3
-        with:
-          app-id: ${{ secrets.RELEASE_TAGGER_APP_ID }}
-          private-key: ${{ secrets.RELEASE_TAGGER_APP_PRIVATE_KEY }}
-
       - name: Resolve RC tag
         id: rc
         env:
           DISPATCH_TAG: ${{ inputs.tag_name }}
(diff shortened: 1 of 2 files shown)


quad341 (approved):
@julianknutsen — thank you for chasing this all the way down. The duplicate-publish diagnosis is the good part: noticing that a GitHub App installation token re-fires the workflow's own `push:` trigger where the default `GITHUB_TOKEN` doesn't is the kind of thing that only ever shows up as "why did v1.4.1-rc1 draft twice?" and then costs someone an afternoon.

The read-only `needs_create` probe is a clean way to pick the lane without needing a credential to make the decision, and moving the mint *behind* it is a real bonus — re-dispatching an existing RC tag now works with no tagger secrets at all. I also appreciated that you kept the friendly missing-secrets guard reachable via `continue-on-error` rather than letting operators hit a bare `Input required and not supplied: app-id`.

I traced the full job rather than just the hunks: the lane guards hold on the failure path too, the tag push won't wake `release.yml` (it skips prerelease tags), and the shared concurrency group queues the publish run behind the create run instead of racing it. Merging as-is — nothing needed on your side. The `RELEASING.md` update matching the new two-run flow was the right call; future-me will need it.


---

PR #5700, opened 2026-08-27
fix(dashboardbff): unwrap the response writer before the Flusher check on run-detail stream

### LINKED ISSUE #5675: Dashboard BFF run-detail SSE route always returns 500 "streaming unsupported"
## Summary

The dashboard BFF's run-detail SSE route always fails:

```
GET /api/city/{city}/runs/{runId}/detail/stream
→ 500 {"error":"streaming unsupported"}
```

The `/v0` SSE routes on the same process and port are fine, so this is specific to the `/api` (dashboard BFF) subtree.

## Evidence that it is unconditional, not data-dependent

**A nonexistent run id returns the byte-identical error**, so the handler bails before it ever resolves the run:

```
$ curl -s "http://127.0.0.1:8372/api/city/city/runs/does-not-exist/detail/stream"
{"error":"streaming unsupported"}   <- 500
```

**It fails in ~50µs** — no store lookup, no I/O. Consistent with an `http.Flusher` type assertion failing at the top of the handler:

```
attempt1: 500 in 0.000702s
attempt2: 500 in 0.000587s
```

Supervisor log agrees (`25µs`, `47µs`, `70µs` across separate requests).

**`/v0` SSE works on the same server**, which rules out the HTTP server, the port, and the client:

```
$ curl -i -N "http://127.0.0.1:8372/v0/city/city/session/{id}/stream"
HTTP/1.1 200 OK
Content-Type: text/event-stream
Transfer-Encoding: chunked
...
id: 1
event: turn
```

## Likely cause

The `ResponseWriter` reaching the BFF handler does not implement `http.Flusher`. Since the `/v0` SSE stack flushes correctly on the same server, the difference is middleware applied to the `/api` subtree that wraps the `ResponseWriter` without forwarding `Flusher`. SSE cannot work without it, so this route can never stream as currently written.

Suggested fix: have the `/api` wrapper implement `http.Flusher` (delegating to the wrapped writer), or exempt the streaming route from that wrapper.

## Impact

Degraded, not broken. The SPA takes the 500 and immediately falls back to polling, so the run-detail page still updates — just not live. From the supervisor log, a stream 500 is followed within the same millisecond by a `detail` 200, repeating on ~20s intervals:

```
14:31:16 api: GET /api/city/city/runs/ci-667/detail/stream 500 70µs
14:31:16 api: GET /api/city/city/runs/ci-667/detail        200 20.3ms
```

## Environment

- gc 1.4.1, commit 58ef17e3bd685fd5cf7f21286277b208d3324590, build date 2026-08-15T23:24:18Z
- supervisor mode, API on 127.0.0.1:8372
- Linux 5.15 (WSL2)

## Reproduction

1. Run a city under the machine supervisor.
2. Dispatch any work so a run exists, e.g. `gc sling <agent> "say hello"`.
3. `curl -i "http://127.0.0.1:8372/api/city/<city>/runs/<runId>/detail/stream"`

A nonexistent run id reproduces it equally, so step 2 is optional.

rjgeng:
## Summary

The dashboard BFF's run-detail SSE route (`GET /api/city/{city}/runs/{runId}/detail/stream`) always failed with `500 {"error":"streaming unsupported"}` — unconditionally, in ~50µs, even for a nonexistent run id, so the handler was bailing before it ever resolved anything.

Root cause: the logging middleware wraps every `ResponseWriter` in a struct that embeds the `http.ResponseWriter` interface (not the concrete writer), so `Flush` is never promoted to the wrapper. A bare `w.(http.Flusher)` type assertion on that wrapper fails even though the real underlying writer supports it fine — the `/v0` SSE routes on the same server already handle this correctly via `sse.go`'s `findFlusher`, walking `Unwrap()` until it finds one. This route just wasn't doing that.

## Fix

- Added `findFlusher`, mirroring the existing `/v0` pattern: walks `Unwrap()` until it finds a writer that implements `http.Flusher`, or returns `nil` if none do.
- Swapped the bare type assertion for `findFlusher(w)`.
- Added a regression test (`TestRunDetailStreamFlusherUnwrapsWrappedWriter`) that wraps `httptest.ResponseRecorder` in a writer exposing only `Unwrap` (not `Flush` directly) — matching the shape the production middleware actually produces. Every existing test in this file hands the handler a recorder directly, which natively implements `Flusher` and never exercised this path.

Fixes #5675.

## Test plan

- [x] `go build ./...` clean
- [x] `gofmt -l` empty
- [x] `go test ./internal/api/dashboardbff/... -run TestRunDetailStream` — all pass, including the new regression test

### DIFF
--- a/internal/api/dashboardbff/rundetail_stream.go
+++ b/internal/api/dashboardbff/rundetail_stream.go
@@ -103,8 +103,8 @@ func (p *Plane) handleRunDetailStream(w http.ResponseWriter, r *http.Request) {
 	}
 	runID := r.PathValue("runId")
 
-	flusher, ok := w.(http.Flusher)
-	if !ok {
+	flusher := findFlusher(w)
+	if flusher == nil {
 		// A ResponseWriter that cannot flush cannot stream; fail closed so the
 		// SPA falls back to the GET + nudge path rather than hanging on a stream
 		// that never delivers a frame.
@@ -256,4 +256,25 @@ func frameSeq(detail runproj.FormulaRunDetail) int64 {
 		return detail.SnapshotEventSeq.Seq
 	}
 	return 0
+}
+
+// findFlusher unwraps w to find one that supports http.Flusher. Middleware
+// wrappers commonly embed the http.ResponseWriter interface (not the
+// concrete writer), which promotes only the interface's own methods
+// (Header/Write/WriteHeader) -- not Flush -- so a bare type assertion on the
+// wrapper can fail even though the underlying writer supports it.
+func findFlusher(w http.ResponseWriter) http.Flusher {
+	type unwrapper interface {
+		Unwrap() http.ResponseWriter
+	}
+	for {
+		if f, ok := w.(http.Flusher); ok {
+			return f
+		}
+		if u, ok := w.(unwrapper); ok {
+			w = u.Unwrap()
+			continue
+		}
+		return nil
+	}
 }
--- a/internal/api/dashboardbff/rundetail_stream_test.go
+++ b/internal/api/dashboardbff/rundetail_stream_test.go
@@ -490,6 +490,79 @@ func TestRunDetailStreamUnsupportedRun422(t *testing.T) {
 // window and 404 after it expires, before any stream body — covered by
 // TestRunDetailStreamUnknownRunWarmingGrace in rundetail_grace_test.go.
 
+// unwrapOnlyWriter wraps an http.ResponseWriter but promotes only Unwrap, not
+// Flush — mirroring the production logging middleware's responseWriter, which
+// embeds the http.ResponseWriter interface, so Flush is never promoted even
+// though the underlying writer (here, an httptest.ResponseRecorder)
+// implements it. frameWritten closes once, after the first Write, so a test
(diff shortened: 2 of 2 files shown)


quad341 (approved):
@rjgeng — thank you for this one, and specifically for the diagnosis rather than just the fix. "500 in ~50µs, even for a nonexistent run id" is the observation that turns a vague streaming complaint into a precise bail-out point, and you followed it all the way to the embedded-interface-vs-concrete-writer distinction. That is genuinely easy to stare past.

I traced your claims against the tree and they all hold: `withLogging`'s `responseWriter` is the wrapper in the path, it does implement `Unwrap`, and the plane is mounted inside that chain. Mirroring `sse.go`'s existing pattern rather than inventing a new one was the right call — the two copies can't be shared without an import cycle, so please don't feel that duplication is yours to resolve.

Special credit for the regression test. Noticing that every existing test in the file hands the handler a raw `ResponseRecorder` — which natively flushes, and so never once exercised the path that was completely broken in production — is exactly the "why didn't the suite catch this?" question that stops the next one. The guard asserting your own fixture isn't a `Flusher` is a lovely detail.

Merging as-is. Nothing for you to do.


---

PR #5704, opened 2026-08-27
test(pool): pass session info to deferred demand cases

jm2:
Fix the two deferred-demand tests added by #5094 after the pool-demand helper gained the sessionInfos parameter. Both calls now pass nil, matching the no-open-session setup used by these cases.\n\nWithout this, cmd/gc test compilation fails with not enough arguments.\n\nValidation: go test ./cmd/gc -run ^TestFilterAssignedWorkBeadsForPoolDemand -count=1

### DIFF
--- a/cmd/gc/assigned_work_scope_test.go
+++ b/cmd/gc/assigned_work_scope_test.go
@@ -304,7 +304,7 @@ func TestFilterAssignedWorkBeadsForPoolDemandDropsDeferredRoutedBead(t *testing.
 		},
 	}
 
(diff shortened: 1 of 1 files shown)


jm2:
@wynged This is the two-call follow-up to #5094: current main cannot compile the cmd/gc test package because those new tests omitted the sessionInfos argument. The patch is test-only, and every required CI gate is now green. A quick review would unblock current-main validation.

jm2:
Closing as superseded by current main: upstream commit 0f0fcb9bd (PR #5104) now passes the added sessionInfos argument at both deferred-demand call sites, so this test-only compile repair is already present and no backport/stacking PR is needed.

PR #5706, opened 2026-08-28
feat(cmd/gc): remind an agent whose stop-pending drain never took — and lead-fix the red main

julianknutsen:
## Summary

**Lead commit — main is red**: `go test ./cmd/gc` does not compile at the current tip; #5094 added two `filterAssignedWorkBeadsForPoolDemand` test call sites at the old signature. Fixed first, droppable independently.

**The feature (ga-b7213):** a drain request is metadata-only — an idle agent never polls, so it never learns; and when a queued stop keeps failing with the runtime alive, `finalizeDrainAckStopPendingSessions` re-queues the same stop forever. This series anchors a REMINDER on that exact live-runtime branch (which already pays for the liveness observation every tick): first sight delivers reminder #1 carrying the canonical explicit-arg ack command (`gc runtime drain-ack <session-id>` — survives stale pane env), then a 10-minute cadence, max 3, markers drain-scoped (`instance_token` + `drain_at`) and write-ahead persisted.

Honest delivery accounting: spend and delivery are separate facts — a delivered budget earns its full answer interval; an all-undeliverable budget (input-dead pane) earns none, and every journal line distinguishes "unanswered reminders" from "undeliverable reminder attempts (input-dead pane)", including the mixed case. The seam adapter's nil-on-attach-failure limit is documented at the call site.

Safety: the reminder writes nothing once the ack source is `agent` (mutation-pinned); idleness reads RAW tmux `session_activity`, not the observation cache; non-tmux providers fail closed (the k8s session-level activity limitation is documented as a bounded informational-nudge exposure).

## Review process

Designed, implemented, and adversarially reviewed in three waves: the first review confirmed 14 findings — including two criticals proving the original tracker-anchored design unreachable for its target population — which forced the durable-row re-anchor; the focused re-review of the rework confirmed one remaining minor (the delivery-accounting honesty above), fixed. 8 mutation-verified pins.

Refs: ga-b7213

### DIFF
--- a/cmd/gc/drain_reminder.go
+++ b/cmd/gc/drain_reminder.go
@@ -1,0 +1,401 @@
+package main
+
+// The drain reminder: a stop nobody was ever asked to make.
+//
+// A drain-acked session is moved to the durable stop-pending state and its
+// runtime is stopped asynchronously. When that stop does not take — the pane
+// survives the signal, the process reparents, the provider refuses — the row
+// stays `state=draining, state_reason=drain-ack-stop-pending` with a LIVE
+// runtime, and the controller re-queues the same stop on every tick, forever.
+// Nothing in that loop ever tells the AGENT anything. It cannot: the drain
+// signal is metadata, learned by polling, and the session this happens to is
+// the one sitting idle at its prompt not polling anything.
+//
+// So the loop has no exit but an operator killing the pane. This pass adds the
+// one thing missing from it: ask the agent. It runs off the same durable row
+// state the re-examination already reads, in the bounded shape the two nudge
+// backstops use (observe, remind, back off, give up), and it is purely
+// informational — it never stops anything.
+//
+// # It is anchored on the ROW, not on the drain tracker
+//
+// The in-memory drainTracker is the wrong anchor and an earlier revision of
+// this file proved it: a tracked drain is tracker-resident for about two
+// advances before the stop-pending handover clears it, the whole tracked phase
+// is bounded by defaultDrainTimeout, and a controller restart drops the tracker
+// entirely. The wedged rows — the ones that sit for days — have no tracker
+// entry at all. Everything here therefore keys off the session bead and the
+// durable stop-pending state, which are exactly what survives.
+//
+// # The reminder carries the identity because the pane's environment may not
+//
+// `gc runtime drain-ack` with no argument binds the acknowledgement through the
+// caller's own GC_SESSION_ID/GC_INSTANCE_TOKEN. An adopted pane whose
+// environment did not survive a restart therefore acks as nobody, which reads
+// downstream as no acknowledgement at all. The reminder text names the session
+// id so the agent runs the explicit-argument form, which resolves the target
+// from the store instead of from the pane.
+//
+// # Idleness comes from the runtime, not from a cache
+//
+// The quiet test reads the provider's activity signal directly (raw tmux window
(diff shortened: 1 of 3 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277254229

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277256776

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277257964

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR adds a durable, cadence-gated drain-reminder subsystem that nudges a session stuck mid-drain (stopping but not yet drained), tracks delivery/attempt/exhaustion state in bead metadata, and announces a one-shot escalation when a session refuses to drain — wired into the session reconciler tick. That closes a real reliability gap: a session that stalled during drain could previously sit silently with no operator-visible nudge or escalation, and because the new state lives in the bead it survives reconciler restarts instead of resetting.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- Fail-closed drain behavior under observation errors: the review probed whether the new reminder path could regress the reconciler when a pane can't be observed, and confirmed the observation-error arm is behaviorally identical to `origin/main` — a transient observation error defers one tick and self-heals well within the reminder cadence, so no session is nudged or closed off a stale observation.
- Durable, restart-safe state: the attempt / failed / hold markers are written to and re-read from bead metadata, so reminder cadence, exhaustion, and the one-shot escalation announcement survive a reconciler restart rather than resetting to zero.
- Late-ack correctness: the ack-pin lookup runs before the exhausted check, so a drain-ack that lands on the final tick still flips the outcome off the exhausted/escalation path — the ordering the new tests lock in.
- Test coverage: the change lands with a full companion test file exercising the delivered, undeliverable, exhausted, and hold paths.

- `mc-m61y` — Behavioral correctness (minor): the observation-error reconciler arm (`cmd/gc/session_reconciler.go:781-797`) intentionally skips the reminder (fail-closed) and is byte-identical to `origin/main`; optional future hardening could address the persistent-observation-error residual. Not addressed in this PR.
- `mc-4fal` — Change impact / scope (minor): `drainRemindersSpent` and `drainReminderSpendPhraseFor` (`cmd/gc/drain_reminder.go:380-424`) have no in-repo consumer yet — the escalation consumer is expected to land out-of-tree. Not addressed in this PR.
- `mc-pv8s` — Efficiency (minor): an exhausted reminder row still pays one provider subprocess (the ack-pin `GetMeta`) per reconcile tick (`cmd/gc/drain_reminder.go:244-248`, `:176-182`); it could settle to a store-only steady state. Not addressed in this PR.
- `mc-rs7k` — Error handling (nit): the failed-marker write (`cmd/gc/drain_reminder.go:328`) and `clearDrainReminderHold` (`:455-461`) drop their write errors, so journal / operator state can go stale silently; logging them like the sibling writers already do would close the gap. Not addressed in this PR.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `a94d5b9` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/33213641253

### Review Iterations
1 review pass performed; accepted as-is with no fix iterations required.

---

PR #5708, opened 2026-08-28
fix(tmux): always bracketed-paste nudges, whatever their size

aphexcx:
## The bug

`sendLiteralText` types any payload of **4096 bytes or less** as a raw `send-keys -l` burst and reserves the paste buffer for anything larger.

Outside a bracketed paste, every newline in the payload is an ordinary Enter. So a TUI reading the burst submits at the first one and treats the rest as a fresh line — and a busy TUI that is not servicing input drops the leading keys outright. The observable result is the head of a message vanishing and the tail being submitted alone.

Four messages to agents arrived as fragments this way on 2026-08-27/28: `""`, `"l post."`, `"path to view it"`, `"nt --conversation-id …"`.

Payload size never had anything to do with it. The 4096 threshold only decided which messages got the framing that made them safe — a short multi-line reminder is exactly as splittable as a long one, and short reminders are the common case.

## The fix

Every nudge now takes the `pasteLiteralText` path, so the framing is a property of the transport rather than of the message that happened to be long enough. The cost is a temp file and two extra tmux invocations on short nudges — microseconds against a delivery budget already measured in hundreds of milliseconds.

The **hidden-attach** path writes bytes straight into an attached client's stdin rather than going through tmux's paste-buffer, so it has to frame the payload itself with the DEC 2004 markers. Measured through `script` + `tmux attach-session` against a raw-mode TUI with bracketed paste enabled, the markers reach the pane intact (payload arrives as `ESC[200~ … ESC[201~`), and an unframed write of the same payload arrives as bare newlines — i.e. as individual Enter presses. The submit Enter stays **outside** the paste, where it is a keypress rather than pasted text.

tmux strips the markers for a pane whose application has not requested bracketed paste, so framing is safe for panes that do not support it.

**Single-keystroke paths are untouched.** Approval digits and submit keys are genuine keypresses and still go through `send-keys`.

## Receipts

Each framed nudge emits a `runtime.NudgeReceipt` (id, target, byte count, payload digest, framing, submit confirmation), so a delivery that *could not* have been split is distinguishable in the logs from one that could. A `NudgeReceiptSink` can replace the default log line.

`Submitted=false` is the normal best-effort outcome on transports with no busy probe and is explicitly **not** a truncation signal — the type documents this, because reading it as failure would be the obvious wrong inference.

## Tests

Framing is asserted across the retired threshold — 1 byte, 4095, 4096, 4097 — and each case additionally checks that **no raw literal `send-keys` is issued**. That negative matters: a regression that reinstated the fast path for short payloads would still paste the long ones and would pass a large-payload-only test.

Also covered: the staged buffer holds the whole payload (the byte accounting receipts rest on), receipts are emitted for both transports with unique ids, and no receipt is emitted when delivery fails.

`scripts/runtime-tmux-tests.manifest` refreshed for the changed inventory.

## Note

This is the upstream-relevant half of a larger change in our fork. The other half — a synchronous delivery receipt on the `extmsg/inbound` response, so a chat adapter can gate its dedup claim on real delivery — is built on structures that have diverged here, so it is deliberately left out rather than ported half-correctly. Happy to follow up with it if there's interest.


atbrace:
Independent confirmation of this bug, measured live 2026-09-16 on our fork's fleet: a queued Discord follow_up (~under 4096 bytes) delivered via `RuntimeHandle.Nudge` → `tmux.sendLiteralText` went through `send-keys -l` and each embedded newline before the last was submitted as its own premature Enter — only the closing lines of the payload reached the agent's transcript. Same root cause and mechanism this PR describes (payload size was never actually the discriminator).

We landed a minimal same-class fix locally (route through `pasteLiteralText` whenever the text contains a newline, regardless of length) rather than duplicating this PR's broader "always paste" approach + receipts machinery, since this PR already covers the identical bug more thoroughly. Flagging for visibility in case it helps this get reviewed — happy to help verify if useful.

PR #5711, opened 2026-08-28
fix(orders): share the doctor's cron parser and validate schedules at load

### LINKED ISSUE #5709: orders: runtime cron matcher silently drops ranges that gc doctor accepts; no load-time schedule validation
## What happened

An order with `schedule = "*/15 16-23 * * 1-5"` never fired — not once, over a week. `gc doctor` did not report a syntax problem, and `gc order check` simply showed `cron: schedule not matched`.

The schedule is silently unsupported: the runtime cron matcher has no range support, and it discards the parse error rather than surfacing it.

## Root cause

**Runtime** — `internal/orders/triggers.go`, `cronFieldMatches`:

```go
if strings.HasPrefix(part, "*/") {
    step, err := strconv.Atoi(strings.TrimPrefix(part, "*/"))
    if err == nil && step > 0 && value%step == 0 { return true }
    continue
}
n, err := strconv.Atoi(part)
if err == nil && n == value { return true }
```

`strconv.Atoi("16-23")` returns an error, `if err == nil` discards it, the loop falls through, and the field never matches. Any unparseable field silently yields an order that never fires — ranges are just the most likely way to hit it.

**Doctor** — `internal/doctor/checks_order_firing.go` uses a *different, richer* parser (`cronPartMatchesForDoctor` / `cronRangeForDoctor`) that fully supports `a-b` and `a-b/N`, is bounds-checked, and returns errors instead of swallowing them. It has a passing test for `0 9-17 * * *` (`checks_order_firing_test.go`), so doctor is explicitly tested to accept syntax the runtime silently drops.

**No load-time validation** — `internal/orders/order.go` `Validate()` only checks that a cron trigger has a non-empty `schedule` string. There is no field-count or field-syntax check at discovery, unlike `tz`, which is a hard error at load.

To be fair to doctor: it does eventually flag the order via `classifyOrderFiring`, but as `StatusError` at `SeverityAdvisory` with the message *"never fired since controller start Xh ago"* — which reads as a scheduling hiccup, not "your schedule is unparseable." Non-blocking and misleading rather than a false pass.

## Why this is worth fixing rather than documenting

`docs/tutorials/07-orders.md` promises `*`, integers, comma lists, and `*/N` — strictly read, ranges aren't offered. But:

- Doctor's parser is the newer, more careful one, written deliberately for #2499, and already tested for ranges.
- Documenting the limitation leaves the silent-swallow in place, which is the real footgun — *any* typo produces a permanently dead order with no diagnostic.
- `a-b` is universal cron syntax; people will keep writing it.

## Suggested fix

1. Lift `cronRangeForDoctor` / `cronPartMatchesForDoctor` into a shared location (e.g. exported from `internal/orders`) and have `internal/doctor` call it, so the two parsers cannot diverge again.
2. Add field-syntax validation to `Order.Validate()` so an unparseable schedule fails loudly at order discovery, the way a bad `tz` already does.

Small and self-contained: no API, schema, or config changes; one docs line to update.

**Behavior change to call out:** `*/N` currently means `value % step == 0`; the doctor parser means `(value - lo) % step == 0`. Identical for minute and hour (lo=0), **different for day-of-month and month** (lo=1). Standard cron follows the latter, but it is a real change and should be noted in any release notes.

Happy to open a PR for this if the direction looks right.

Environment: gascity 1.4.1, Ubuntu 24.04, x86_64.

Rome-1:
Fixes #5709.

## The bug

`internal/orders/triggers.go` `cronFieldMatches` supported only `*`, integers,
comma lists, and `*/N`. It called `strconv.Atoi` on each part and dropped the
error (`if err == nil` with no `else`), so any field it could not parse — a
range like `16-23` above all — quietly failed to match and the order never
fired. No diagnostic, ever.

`internal/doctor/checks_order_firing.go` carried a second, richer parser
(`cronPartMatchesForDoctor` / `cronRangeForDoctor`) that supports `a-b` and
`a-b/N`, bounds-checks every field, and returns errors. It has a passing test
for `0 9-17 * * *`. Doctor was tested to accept syntax the runtime silently
dropped.

## The fix

**One parser.** The doctor's parsing moves into `internal/orders/cron.go`,
exported as `CronFieldMatches`, `CronScheduleMatchesAt`, and
`ValidateCronSchedule`. The runtime matcher and the doctor's firing check now
read schedules through the same code and the same field-bounds table
(`cronFieldSpecs`), so they cannot diverge again. The four doctor-local helpers
are deleted.

**Loud failure at load.** `Order.Validate()` now runs `ValidateCronSchedule`,
so an unparseable or out-of-bounds schedule is a hard error at order discovery
— the way an invalid `tz` already is. `checkCron` validates up front too and
reports `bad cron schedule: <why>` instead of `cron: schedule not matched`.

**Errors are not swallowed anywhere.** `CronFieldMatches` parses every
comma-separated part even after one matches, so `5,notanumber` is an error
rather than a match that hides a typo.

## Behavior changes

Three, all in the direction of standard cron. Worth release-note treatment:

1. **`*/N` step anchoring on 1-based fields.** The old runtime rule was
   `value % step == 0`; the rule now is `(value - lo) % step == 0`, where `lo`
   is the field's lower bound. Identical for minute and hour (`lo = 0`), but
   **different for day-of-month and month** (`lo = 1`): `*/2` in day-of-month
   now means the 1st, 3rd, 5th, … where it previously meant the 2nd, 4th, 6th,
   …. Standard cron uses the new semantics. Pinned by tests in
   `TestCronFieldMatches` ("dom step from 1", "month step from 1").

2. **Out-of-bounds fields are now rejected at load.** `70 * * * *`,
   `0 24 * * *`, `0 0 0 * *`, and `0 0 * * 7` used to load and simply never
   fire (or, for `0 0 0 * *`, never match). They are now load errors. Note
   `7` as a Sunday alias in day-of-week is **not** supported — the shared
   bounds are `0-6`, matching the doctor parser this lifts from. No cron order
   in this repo or in `gascity-packs` uses any of these forms.

3. **Ranges now work at runtime.** `16-23`, `9-17`, and `5-59/10` fire instead
   of never matching. This is the point of the change.

## Tests

- `internal/orders/triggers_test.go` — `TestCronFieldMatches` extended from 5
  cases to 31: ranges, ranges in lists, `*/15`, `5-59/10`, the DOM/month step
  anchoring, out-of-bounds rejection, and a malformed part behind a matching
  one. Plus `TestCronScheduleMatchesAtRanges`, `TestCheckCronRangeSchedule`
  (the reported `*/15 16-23 * * 1-5` now reaches the trigger), and
  `TestCheckCronRejectsUnparseableSchedule`.
- `internal/doctor/checks_order_firing_test.go` — unchanged. Its
  `0 9-17 * * *` case is the divergence regression guard and stays green
  through the refactor.
- `internal/orders/order_test.go` — `TestValidateCronScheduleSyntax` and
  `TestValidateCronBadSchedule` pin the accepted grammar and 15 rejected
  forms.

`docs/tutorials/07-orders.md` now documents ranges, step anchoring, and the
load-time bounds check.

## Design notes

Zero Framework Cognition: this is transport, not reasoning — the parser reads
a schedule and reports what it says, with no heuristic about what the author
"probably meant". Bitter Lesson alignment: one parser and a loud error at the
boundary, rather than two parsers whose disagreement is only discoverable by
an order that never runs.

### DIFF
--- a/docs/tutorials/07-orders.md
+++ b/docs/tutorials/07-orders.md
@@ -173,9 +173,12 @@ Notes per trigger:
   waits `interval` since the last run. Drifts: a 3:02 run means the next is at
   3:07.
 - **`cron`** — a 5-field expression (minute, hour, day-of-month, month,
-  day-of-week) supporting `*`, integers, comma lists (`1,15`), and `*/N` steps.
-  Unlike cooldown it hits the same wall-clock times every day. Fires at most
-  once per minute.
+  day-of-week) supporting `*`, integers, comma lists (`1,15`), ranges (`9-17`),
+  and steps on either (`*/15`, `5-59/10`). A step counts from the start of its
+  range, so `*/2` on day-of-month means the 1st, 3rd, 5th. Every field is
+  bounds-checked at order discovery: an out-of-range or unparseable schedule is
+  a load error, not an order that quietly never fires. Unlike cooldown it hits
+  the same wall-clock times every day. Fires at most once per minute.
 - **`condition`** — the orchestrator runs `sh -c "<check>"` each tick, bounded by
   the order's `check_timeout` (a positive Go duration, default `10s`). This is
   separate from `timeout`, which bounds the dispatched formula/exec rather than
--- a/internal/doctor/checks_order_firing.go
+++ b/internal/doctor/checks_order_firing.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"log"
 	"path/filepath"
-	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -425,8 +424,8 @@ func expectedIntervalForOrder(order orders.Order, cronCache map[string]time.Dura
 
 func computeExpectedIntervalForCronSchedule(schedule string) (time.Duration, error) {
 	fields := strings.Fields(schedule)
-	if len(fields) != 5 {
-		return 0, fmt.Errorf("invalid cron schedule: want 5 fields, got %d", len(fields))
+	if len(fields) != orders.CronFieldCount {
+		return 0, fmt.Errorf("invalid cron schedule: want %d fields, got %d", orders.CronFieldCount, len(fields))
 	}
 
 	// Scan minute-by-minute from a fixed base so the result is deterministic
@@ -453,9 +452,9 @@ func computeExpectedIntervalForCronSchedule(schedule string) (time.Duration, err
 		matches := make([]time.Time, 0, 16)
 		for i := 0; i < windowMinutes; i++ {
 			ts := base.Add(time.Duration(i) * time.Minute)
-			matched, err := cronScheduleMatchesAt(fields, ts)
+			matched, err := orders.CronScheduleMatchesAt(fields, ts)
 			if err != nil {
-				return 0, err
+				return 0, fmt.Errorf("invalid cron schedule: %w", err)
 			}
 			if matched {
 				matches = append(matches, ts)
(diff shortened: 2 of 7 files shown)


quad341 (approved):
@Rome-1 — this is a great catch and an even better fix. Spotting that the runtime matcher and the doctor's checker had drifted apart, and that doctor had a *passing test* for `0 9-17 * * *` while the runtime silently dropped it, is the kind of bug that usually only surfaces after someone's order has been quietly dead for a month. Collapsing both onto one `cronFieldSpecs` table so they can't diverge again is exactly the right shape, and I appreciate that you called out all three behavior changes explicitly rather than letting the day-of-month step-anchoring shift ride along unmentioned.

The tests deserve their own mention: 5 cases to 31, with the inverted range, the bounds rejections, and the `5,abc` "bad part hiding behind a good one" case. `TestCheckCronRangeSchedule` going through `checkCron` instead of stopping at the parser is what convinced me the reported schedule genuinely fires now.

Nothing for you to change. CI is red for a reason that isn't yours — two stale call sites in `cmd/gc/assigned_work_scope_test.go` on main break that package's test build, which reddens the shards and the required rollup. That's ours. I'll fix main, rebase you, and merge. I ran `./internal/orders` and `./internal/doctor` against your head locally and both are green. Thanks for this one.


---

Rome-1:
One edge case from the parallel branch, now that this has landed — flagging it rather than opening an issue, since it is narrow and you may have decided it deliberately.

**`0 0 * * 7` is now a hard load error, and it is valid standard cron.**

`crontab(5)` specifies day-of-week as `0-7`, where **both 0 and 7 mean Sunday**. The shared `cronFieldSpecs` here sets day-of-week to `{min: 0, max: 6}`, and since `Order.Validate()` now calls `ValidateCronSchedule`, a schedule using `7` for Sunday fails at order discovery rather than merely never matching.

This trades one failure mode for a louder one, which is the right direction in general — but the direction is different for this particular input:

| | before | after |
|---|---|---|
| `0 0 * * 7` | loads, never fires (silent) | **load error** |

Before, a user with a Sunday order written as `7` had one broken order. After, they have a hard error at discovery. Worth knowing which blast radius that carries in your loader — if a single unparseable schedule aborts discovery for the file or the pack, one user's `7` takes out orders that were working.

The PR message notes the corpus check — no order in this repo or in the packs repo uses these forms. That holds for both repos and does not extend to users' own orders, which is where `7` would actually appear.

**The fix is one line if you want it**, and it does not disturb the shared-bounds design: accept `0-7` for day-of-week and normalize `7` to `0` at match time. The alternative — leaving it rejected — is defensible too, as long as the error text names the alias, because "day-of-week outside 0-6" does not tell someone who wrote standard cron what is wrong with their standard cron.

Either way the parallel branch is closed as superseded by this one; I'm not re-proposing it.

PR #5718, opened 2026-08-28
fix(wisp): preserve split-store routing in preclaim bridge

jm2:
## What this fixes

Current `main` does not compile after the preclaim molecule bridge was merged across the work/graph store split. The new bridge refers to an out-of-scope `store`, resolves roots through an out-of-scope `graphStore`, and its new tests still call the older two-argument resolver.

This keeps the class boundary explicit:

- list claimed or routed source beads through `workStore`;
- resolve their `molecule_id` roots through `graphStore`; and
- pass both stores through the routed-preclaim fallback.

The two prewake regressions now use disjoint stores and disjoint ID ranges, so a future store inversion cannot be hidden by a same-store fixture.

## Reproduction

Exact upstream `eec4a2fb625279ac62f29ff4f6a554168bc77b1a` fails an isolated build with:

```text
cmd/gc/wisp_step_inject.go:218:36: undefined: store
cmd/gc/wisp_step_inject.go:222:44: undefined: store
cmd/gc/wisp_step_inject.go:266:16: undefined: graphStore
```

## Validation

- focused wisp resolver and prewake tests
- `go test ./cmd/gc -count=1`
- `go vet ./cmd/gc`
- `go build -buildvcs=false -o /dev/null ./cmd/gc`
- repository pre-push fast gate (all 10 jobs)

All pass with this change.

### DIFF
--- a/cmd/gc/wisp_step_inject.go
+++ b/cmd/gc/wisp_step_inject.go
@@ -215,11 +215,11 @@ func resolveMoleculeRootViaBridge(workStore, graphStore beads.Store, assignees [
 		Limit:     10,
 	})
 	if err == nil {
-		if root := firstMoleculeRootFrom(store, results); root != nil {
+		if root := firstMoleculeRootFrom(graphStore, results); root != nil {
 			return root
 		}
 	}
-	return resolveMoleculeRootViaRoutedBridge(store, assignees)
+	return resolveMoleculeRootViaRoutedBridge(workStore, graphStore, assignees)
 }
 
 // resolveMoleculeRootViaRoutedBridge finds the molecule root reachable from a
@@ -236,9 +236,9 @@ func resolveMoleculeRootViaBridge(workStore, graphStore beads.Store, assignees [
 // later identity still gets a chance.
 //
 // Returns nil when no routed bridge bead is found for any identity.
-func resolveMoleculeRootViaRoutedBridge(store beads.Store, assignees []string) *beads.Bead {
+func resolveMoleculeRootViaRoutedBridge(workStore, graphStore beads.Store, assignees []string) *beads.Bead {
 	for _, identity := range assignees {
-		results, err := store.List(beads.ListQuery{
+		results, err := workStore.List(beads.ListQuery{
 			Status:   "open",
 			Metadata: map[string]string{beadmeta.RoutedToMetadataKey: identity},
 			TierMode: beads.TierBoth,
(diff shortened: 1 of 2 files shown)


jm2:
All current checks are green (61 successful, 25 intentional skips, zero failures). I do not have permission to submit the required project-admin review request through GitHub, so this is ready for maintainer review/merge whenever available.

rjgeng:
Independent verification that this fixes the `main` break, from a non-maintainer who ran into it.

`main` at `eec4a2fb6` does not build:

```
$ go build -tags gms_pure_go ./cmd/gc/
# github.com/gastownhall/gascity/cmd/gc
cmd/gc/wisp_step_inject.go:218:36: undefined: store
cmd/gc/wisp_step_inject.go:222:44: undefined: store
cmd/gc/wisp_step_inject.go:266:16: undefined: graphStore
```

With this PR at `c6cec7a0` merged onto that same commit:

```
$ go build -tags gms_pure_go ./cmd/gc/   # clean
$ go vet   -tags gms_pure_go ./cmd/gc/   # clean
$ go test  -tags gms_pure_go -run 'Prewake|PreClaim|MoleculeRoot' ./cmd/gc/
ok    github.com/gastownhall/gascity/cmd/gc   4.967s
```

Two things that might save a reviewer a step. The branch fast-forwards from current `main` (`eec4a2fb6..c6cec7a0`), so it is not sitting on an old base. And the three compile errors land in **two** functions that have swapped each other's parameter names — `resolveMoleculeRootViaBridge` (params `workStore`/`graphStore`) reaches for a bare `store`, while `firstMoleculeRootFrom` (param `store`) reaches for `graphStore`. A fix touching only the first function would leave `main` red, so it is worth confirming both are covered.

No ask attached — just posting the verification in case it is useful.

quad341 (approved):
@jm2 — thank you for catching this and turning it around the same day. Unblocking `main` after a cross-cutting merge is thankless work, and you did it cleanly.

Two things worth calling out. First, you fixed the right thing: instead of papering over `undefined: store` by threading one store everywhere, you read the class boundary the doc comments already described and routed work-class lookups to `workStore` and `molecule_id` root resolution to `graphStore`. That's the version that survives contact with a graph-relocated city.

Second — the test change is the part I appreciated most. Swapping the prewake fixtures to disjoint stores with disjoint ID ranges wasn't needed to get anything compiling. We mutation-tested it both directions: inverting the root lookup fails three tests, and inverting the routed-bridge list fails your pre-claim test specifically. Under the old same-store fixture, both would have passed silently. You closed a hole that wasn't yours to close.

Verified at `c6cec7a`: build, the wisp resolver suite, and the graph-relocation DI guard all pass, and CI's `cmd/gc` shards actually ran. Merging as-is — nothing needed on your side.


---

PR #5730, opened 2026-08-28
fix(ci): resolve the macOS Go toolchain from go.mod instead of a literal pin (ga-1qp5qo)

quad341:
Item 1 of `ga-1qp5qo`. One file. Fixes a red that is **not** any PR's fault.

## The failure

Every macOS job dies before it runs anything:

```
go: go.mod requires go >= 1.26.6 (running go 1.26.5; GOTOOLCHAIN=local)
```

`setup-gascity-macos` pins `go-version` to a literal `"1.26.5"`; `go.mod` says `1.26.6`.

## Why it is worth a PR rather than a version bump

`setup-gascity-ubuntu` was already migrated to resolve the version from `go.mod`:

```yaml
go-version: ${{ inputs.go-version }}
go-version-file: ${{ inputs.go-version == '' && 'go.mod' || '' }}
```

This mirrors that exactly rather than bumping the literal to `1.26.6`, because a literal skews again on the next `go.mod` bump and reproduces this outage. The two actions were already documented as needing to move in lock-step — this removes the need to remember.

## Cost of leaving it

Nightly `main` has been red on this continuously since **2026-08-15**, and at least five open PRs are red on this and nothing else. Each one triggers an hourly `pr-audit` `ci-failing` bead routed to an investigator, who reads the diff, the CI logs and the workflow chain to reach the same *"not this PR's fault"* conclusion every time — the same investigation repeated per PR for an identical answer, and it keeps arriving hourly.

## Scope

Item 1 only, per @gascity/investigator's recommendation to split the bead (`gm-wisp-2rilin`): it depends on neither P0 that was holding `ga-1qp5qo`, and on its own it clears the red everywhere. Both of those P0s are now resolved anyway — `ga-7uo0wt` closed, `ga-ujov66` fixed on main by #5718.

Items 2 and 3 (the empty-package-list hard-fail guard, and the coupled herdr skip cherry-pick) stay on the bead and are **not** in this PR.

Rebased onto current `main` (`b1cc8a6bcc`) immediately before pushing, and re-checked that the literal pin is still present there — this is not superseded.

### DIFF
--- a/.github/actions/setup-gascity-macos/action.yml
+++ b/.github/actions/setup-gascity-macos/action.yml
@@ -3,9 +3,12 @@ description: Install the shared macOS dependencies for Gas City CI jobs on ARM64
 
 inputs:
   go-version:
-    description: Go version to install. Default matches setup-gascity-ubuntu; bump both together.
+    description: >-
+      Go version to install. Empty (the default) reads the version from go.mod,
+      so the toolchain can never skew from the module directive. Pass an
+      explicit version only to test against a different toolchain.
     required: false
-    default: "1.26.5"
+    default: ""
   node-version:
     description: Node.js version to install
     required: false
@@ -47,11 +50,15 @@ runs:
 
     - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6
       with:
-        # Keep this default in lock-step with setup-gascity-ubuntu —
-        # a split between Mac and Linux toolchains would surface as
-        # false "Mac-only" regressions. Track the same pin both
-        # actions use today; bump them together.
+        # Resolved from go.mod, not a literal pin, exactly as
+        # setup-gascity-ubuntu already does. A hardcoded version here skews
+        # from the module directive the moment go.mod is bumped, and it
+        # surfaces as a false "Mac-only" regression on every open PR:
+        #   go: go.mod requires go >= 1.26.6 (running go 1.26.5; GOTOOLCHAIN=local)
+        # That is ga-1qp5qo — nightly main was red on it continuously from
+        # 2026-08-15, with at least five open PRs red on this and nothing else.
         go-version: ${{ inputs.go-version }}
+        go-version-file: ${{ inputs.go-version == '' && 'go.mod' || '' }}
         # Keep macOS parity deterministic across hosted and reused runners.
         cache: false
 
--- a/.github/workflows/mac-regression.yml
+++ b/.github/workflows/mac-regression.yml
@@ -137,6 +137,7 @@ jobs:
               - 'cmd/gc/**'
               - 'internal/pathutil/**'
               - 'internal/fsys/**'
+              - '.github/actions/setup-gascity-macos/**'
       - name: Decide which tiers should run
         id: gate
         env:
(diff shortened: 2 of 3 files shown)


quad341:
Triage of the `ci-failing` audit on this PR: **the remaining red is not this PR's fault, and it is the change working as intended.**

## What the checks show

`Mac / quality (lint, fmt, vet, docs)` **passes** here (9m36s). That job had been red continuously since 2026-08-15 on the exact skew this PR fixes. All 12 `Mac / cmd-gc process` shards and `Mac / acceptance (Tier A)` pass too.

The only failure is `Mac / make test` (and `Mac regression summary`, downstream of it) — and the interesting part is *how* it fails:

| | before | this PR |
|---|---|---|
| `go test` package args | **0** | ~190 |
| packages executed | **0** | 174 (170 pass, 4 fail) |
| duration | 0.24s → `PASS` | 5m59s → `FAIL` |

That is `ga-1qp5qo` resolving exactly as specified. The suite had been reporting success while testing nothing since 2026-08-14; it now runs, and four packages that had been failing invisibly for weeks are finally visible.

## The four failures are pre-existing

None of them is reachable from this diff — it touches only the setup action's `go-version` resolution, one `paths-filter` entry, and the test asserting that entry.

Independent confirmation: **#5344** (open, filed 2026-08-17 by @Toady00, eleven days before this PR) reports one of the four failing deterministically, reproduced 3/3 and 5/5 on a clean checkout of `47b317bc6`.

All four are root-caused, and each is filed with a fix spec:

1. **`internal/runtime/herdr`** — `TestSocketPath*`. `os.UserConfigDir()` ignores `XDG_CONFIG_HOME` on darwin and returns `$HOME/Library/Application Support`. This is item 3 of `ga-1qp5qo`, whose own exit contract predicted *"omitting it turns mac red once coverage is restored"* — confirmed. Fix is the already-written `7a8fb35f44` (`skipUnlessXDGPlatform`), still unlanded on `main`. → `ga-bxlbi6`
2. **`internal/pidutil`** — `TestCmdline_FailsClosedWhenUnreadable`. The `ps`-stub trick stopped working when #5176 added the `kern.procargs2` sysctl path; the skip guard covers `linux` only. → `ga-bxlbi6`, and #5344
3. **`internal/worker/adapters/zcode`** — ~10 tests, mostly one fault: the `zcode-repl` fixture runs under `set -uo pipefail` and line 593 dereferences `more`, which a timed-out `read` leaves **unset** on macOS bash. On bash 5.3 it is assigned (empty), so the guard is harmless on Linux — which is why it never surfaced. Plus an independent `/private/var` symlink assertion at `adapter_test.go:1021`. → `ga-uqdim3`
4. **`internal/pathdurability`** — `TestClassifyOnRealFilesystem`. `probe_other.go` (`//go:build !linux`) is a deliberate stub so `Classify` returns `Unknown` off-Linux; the test asserts `CityDevice` with no build tag or skip. The non-Linux behavior is correct — the test is what is wrong. → `ga-p1dis0`

## The one real decision

Merging this trades a **silent** green for an **honest** red: `Mac / make test` will be visibly red on `main` until the four land. Leaving the pin keeps 174 packages untested while reporting success, and keeps `Mac / quality` hard-red on every PR that runs the Mac lane.

That sequencing call is the mayor's, not mine — flagged there. Nothing in this PR needs changing for it.

Artifact for all of the above: run `33200755632`, job `98949547487`, head `bc91c379a9`.

Issue #5473, opened 2026-08-21
gc hook --claim --drain-ack acks the no-work drain with zero grace right after a step close → one session restart per molecule step (drain_acked_with_assigned_work)


swedeinasia-flow:
## Summary

`gc hook --claim --drain-ack` acknowledges the **no-work drain with zero grace**. When a pool worker's end-of-turn hook runs in the window between closing step N and the successor step N+1 becoming visible as ready, the hook sees no candidates, writes the no-work drain and acks it on the spot. The reconciler then records `session.drain_acked_with_assigned_work` and re-wakes the same session (≈1 min later) — so every molecule step costs a provider restart, and a lane with 6 steps runs as 4–6 sessions.

Build: `1.4.1-main.116d1b011` (+ cherry-pick 20de61c3c, beads pin d38ac728b581), macOS, tmux provider, Dolt beads.

## Evidence (2026-08-21, one estate)

- 181 `session.drain_acked_with_assigned_work` events in one day (33 on a rig pool, the rest on the city pool); 239 city-worker drain-acks in 2 h with ~8–14 sessions live.
- One lane, mol-polecat-land over fc-cfos2a, session `gc-c46ybn`:
  - 11:25:42Z `execution.step_started` (load-context)
  - 11:28:44Z `bead.closed` fc-52qkny (load-context) by the worker
  - 11:28:57Z `session.stopped` "drain acknowledged by agent" → `session.drain_acked_with_assigned_work` (bead fc-8nzj8v, in_progress, assigned to this session)
  - 11:30:05Z `session.woke` — same session bead, re-woken by the assigned-work pass
  - the lane landed its MR on the 4th session.
- The nudge that runs at end of turn is the documented one: ``Run `gc hook --claim --drain-ack --json` now …``.

## Mechanism (code)

- `cmd/gc/cmd_hook_claim.go`: after the work query, `if !workQueryHasReadyWork(normalized) { return hookClaimResult{} }` → caller reaches `writeHookClaimNoWork` → `writeHookClaimDrain(reason, …, opts.DrainAck, ops.DrainAck, …)` which runs `drainAckFn` immediately when `--drain-ack` is set.
- `hookClaimExistingAssignment` only inspects *candidates* returned by the work query; an in_progress bead assigned to the session but still dependency-blocked (e.g. a land iteration assigned early) is not a candidate, so it cannot rescue the turn either.
[...]

Issue #3453, opened 2026-06-13
bug: supervisor releaseOrphanedPoolAssignments clears pending-interactive pool claim → duplicate session spawn


iwata-1116:
## Symptom

A polecat claims a pool-routed bead, hits an interactive prompt, and sits silent
(zero tmux activity) for ~52 min. The reconciler eventually false-positives the
holder as "stuck", the orphan-release path clears the claim, demand reappears,
and a backup session is minted. The backup claims the still-held bead and two
Opus polecats grind the same work in parallel — duplicate token burn and
double-write hazards on whatever the original holder eventually produces.

Origin trace: caliber-compass `cc-ft4` (immortan_joe vs the_people_eater),
2026-06-01.

## Why this is filed fresh

- `#1052` is the symptom parent and was **CLOSED state_reason=completed**
  2026-06-03 by a maintainer, but `#1052`'s most recent thread comment
  ("Fresh confirmation of this race at current main `ea610154d`") reports
  the race still firing at HEAD.
- `#1085` ("Add pre_launch startup hooks", julianknutsen) self-declared
  `Fixes #1052` and addresses the **claim-ACQUISITION** layer via a
  `pre_launch` `gc work claim-next --json` hook. It was **CLOSED unmerged**
  2026-06-09 (reviewDecision empty).
- `#1471` shipped a partial mitigation in the worker layer
  (`pool-worker.md:31-39` verify-and-drain). Worker-side, orthogonal.

The residual hole is in the **claim-RELEASE** layer — independent of `#1085`,
`#1471`, and the `#1052`-titled structural work. Filing fresh keeps the
discussion focused on the release path.

## Root cause (verified on HEAD `32d7b0ccf`)

`releaseOrphanedPoolAssignments` (`cmd/gc/pool_session_name.go`) is the sole
live-session path that clears an in_progress pool claim with the holder
absent. Its gate chain:

1. `openSessionOwnsWork`
2. `assigneePreservesNamedSessionRoute`
3. `liveOpenSessionAssignmentExists`
4. `liveWorkAssignmentStillReleasable`
5. `detachedProbeAllowsOrphanRelease`

does **not** consult `pendingInteractionKeepsAwake` (`cmd/gc/session_sleep.go:119`),
the predicate the idle-reaper already uses to suspend its kill decision while a
pending-interactive holder is alive. So:

- The idle-reaper correctly defers (`session_reconciler.go` has multiple call
[...]

Issue #2293, opened 2026-05-17
bug: pool worker drain-ack mid-phase leaves bead stranded with dead session


reggieperry:
## Symptom

When an SDLC chain worker hits Claude Code's per-turn cap mid-phase, the session exits cleanly via `stop_hook_summary` / drain-ack. The bead it owned is left at `status=in_progress`, `current_step=implement` (or whichever phase was in flight), with `assignee` still pointing at the now-exited session. The reconciler does not re-prompt or replace that session. The chain stalls indefinitely until an operator intervenes.

Observed twice in 24 hours against the SDLC discipline pack's worker pool on T7920. Each incident required manual recovery: commit uncommitted worktree changes, push the branch, `bd update --assignee ""`, `gc session kill`, `gc supervisor reload`. The chain resumes correctly once those steps run; the underlying defect is that the controller should be doing them.

## Why the obvious "clear assignee on drain-ack" fix won't work

Tempting first instinct: have `AcknowledgeDrainPatch()` in `internal/session/lifecycle_transition.go` also clear `bead.assignee`. Then the bead returns to `bd ready` and the existing demand-counting in `cmd/gc/build_desired_state.go:878-886` spawns a fresh worker.

It breaks the multi-phase handoff. The SDLC chain's worker prompt deliberately exits the session at the *end* of its phase via:

```
bd update $STORY_ID --status=open --assignee "" \
  --set-metadata "gc.routed_to=${GC_RIG}/sdlc-discipline.tester"
```

then `gc runtime drain-ack` to terminate. The worker controls when it releases the bead by writing the `--assignee ""` explicitly. If drain-ack also cleared assignee unconditionally, every clean phase handoff would briefly leave the bead unassigned-and-routed-to-the-original-pool, racing the supervisor into spawning a duplicate worker against the same phase the original just completed. Clean exits and per-turn-cap exits look identical at the drain-ack layer; you can't disambiguate by drain-ack alone.

## Proposed fix — reactive reconciler recovery

In `cmd/gc/session_reconciler.go` around the drain-ack handler at lines 946-972 (where `sessionHasOpenAssignedWorkForReachableStore()` is consulted), add a third branch: if the session is drain-acking AND its assigned bead has `status=in_progress` AND `current_step` indicates a non-terminal phase (i.e., `current_step` is set and is not `submit-and-exit`), treat this as an incomplete-phase exit. The controller should:

1. Inspect the bead's worktree (`metadata.work_dir`) for uncommitted changes
[...]

PR #5735, opened 2026-08-28
fix(pool): reopen work stranded by a seat that drain-acked while holding it

### LINKED ISSUE #5731: A pool seat that drain-acks while holding assigned work strands the step forever — no wake, no reuse, no dead-assignee reopen
`gc hook --claim --drain-ack` can return `no_work` and acknowledge the drain while the
seat still holds an `in_progress` claim (the claim hook's own
`session.demand_claim_divergence` records `drain_reason=no_work` with
`trigger_status_at_drain=in_progress`). The reconciler finalizes that ack at
`cmd/gc/session_reconciler.go:650` with `CompleteDrainPatch(..., SleepReasonIdle, ...)` —
an ordinary idle sleep — and emits `session.drain_acked_with_assigned_work` as a
diagnostic, deliberately not touching the work bead (#2293: recovery is pack-side
policy).

The problem is that the controller then makes recovery impossible for *any* actor,
pack-side included, without hand-editing beads:

1. `cmd/gc/compute_awake_set.go:232-241` — the scaled-agent wake loop skips sessions
   holding assigned work, **and** `countAssignedScaleSlots` (`:533`) counts them as
   filled demand slots. An asleep holder is therefore never woken *and* suppresses the
   demand that would mint a replacement.
2. `cmd/gc/pool_session_name.go:240/243/246` — `releaseOrphanedPoolAssignments` only
   reopens work whose assignee maps to no **open** session bead. A drain-acked bead is
   asleep but open, so all three liveness gates skip it and
   `bead.dead_assignee_reopened` can never fire.
3. `cmd/gc/build_desired_state_pool_info.go:240` — a one_shot identity holding assigned
   work is (correctly) not wake-reusable, so the name cannot be recycled either.

Compounding it: `CompleteDrainPatch` writes no durable marker, so after the tick a
stranded holder is byte-identical on the bead to a seat that simply went idle. Even a
pack-side subscriber to the event has no queryable state to reconcile against, and the
event stream is not replayed on controller restart.

**Observed:** twice on 2026-08-28 on a real city, on two independent pool lanes.
`session.stopped "drain acknowledged by agent"` + `session.drain_acked_with_assigned_work`
(seq 350545/350546 and 353899/353900); the `execute-operation` steps sat `in_progress`
on the asleep names for 40+ minutes with no `bead.dead_assignee_reopened`, and
`gc session wake` did not resume them (the name is not in `desired`, so the reconciler
re-sleeps it). The behaviour is independent of rig, of `min_active_sessions` (unset on
both the failing and the healthy lanes), and of lifecycle.

**Related:** #5473 (the zero-grace no-work drain-ack that *creates* this state; this issue is the recovery gap once it exists), #2293 (drain-ack path never writes the work bead — preserved by fix candidate A). Verified the three mechanisms above are present on `main` (b1cc8a6bc): `compute_awake_set.go` assigned-work skip + `countAssignedScaleSlots`; `releaseOrphanedPoolAssignments` gated on *open* session beads; `build_desired_state_pool_info.go:220` `sessionBeadHasAssignedWorkInfo` reuse refusal. Observed on a fork built from a main-based EDGE stack; line numbers in the mechanism list are from that base and may be offset slightly on `main`.

**Fix candidates**

**(A, proposed)** Persist a `drain_ack_stranded_at` marker in the drain-ack
patch, clear it in `PreWakePatch` (the pre-start commit, so it cannot outlive
the state it describes even across a failed start), and let
`releaseOrphanedPoolAssignments` treat an **asleep, fungible** seat carrying it
as a non-live claimant. Recovery stays in the lane that already owns
dead-assignee repair, keeps its re-validation and its event; the drain-ack path
still never writes the work bead, so #2293's contract is preserved.

  bead, so #2293's contract is preserved. Patch + RED->GREEN test below.
- **(B)** Release the claim inline in `finalizeDrainAckStoppedSession`. Smaller, but it
  contradicts the contract pinned at `session_reconciler_test.go:1875-1890`.
- **(C)** Make an asleep assigned-work holder wakeable in `compute_awake_set`.
  Rejected: it resurrects one_shot seats that `2fd0eec1f` deliberately retires.

A patch for (A) with a RED→GREEN reconciler + reopen-lane test exists on our fork; happy to open the PR if (A) is the preferred direction.

jacobhausler:
Closes #5731.

## The wedge

A pool seat that runs `gc hook --claim --drain-ack` can get `no_work` back while
it still holds an `in_progress` step. `finalizeDrainAckStoppedSession`
(`cmd/gc/session_reconciler.go`) finalizes that ack as an ordinary idle sleep —
`CompleteDrainPatch(..., SleepReasonIdle, ...)` — and emits
`session.drain_acked_with_assigned_work` as a diagnostic, deliberately not
touching the work bead (#2293: recovery is pack-side policy).

The controller then offers that claim no path back to execution:

1. `compute_awake_set.go` skips any session holding assigned work when choosing
   wake targets, **and** `countAssignedScaleSlots` counts that same session as a
   *filled* demand slot — so neither the holder nor a replacement seat is ever
   desired.
2. `releaseOrphanedPoolAssignments` (`cmd/gc/pool_session_name.go`) only reopens
   work whose assignee maps to no **open** session bead. The drain-acked bead is
   asleep-but-open, so all three liveness gates read it as a live claimant and
   `bead.dead_assignee_reopened` can never fire.
3. `reusablePoolSessionInfo` refuses to wake-reuse a `one_shot` identity that
   still holds assigned work, so the name is not recycled either.

Net: the step sits `in_progress` on a seat that will never run again. Observed
twice in the field on two independent pool lanes — 40+ minutes stranded, and
`gc session wake` does not resume it (the name is not in `desired`, so the
reconciler re-sleeps it).

Compounding it, after `CompleteDrainPatch` a stranded holder is byte-identical
on the bead to a seat that simply went idle. Even a pack-side subscriber to the
event has no queryable state to reconcile against, and the event stream is not
replayed on controller restart.

## The change

Add a durable `drain_ack_stranded_at` marker and let the lane that already owns
dead-assignee repair use it.

1. `DrainAckStrandedAtKey` + `Info.DrainAckStrandedAt`, its `infoKeyCodec` row,
   and the two hand-maintained projection/codec oracles.
2. `finalizeDrainAckStoppedSession` sets the marker when the close gate found
   assigned work and clears it otherwise, on the **same single-clock-read patch**
   that writes the sleep (so `slept_at` and `drain_ack_stranded_at` cannot
   disagree by a tick). The work bead is untouched.
3. `PreWakePatch` clears the marker, so a seat that starts is never a stranded
   holder.
4. `releaseOrphanedPoolAssignments` consults `exitedDrainAckHolderIdentities` —
   asleep + fungible + marker — and lets such an assignee fall through to the
   existing release, which still re-validates
   (`liveWorkAssignmentStillReleasable`, `detachedProbeAllowsOrphanRelease`) and
   emits `bead.dead_assignee_reopened`.

The stranded set is **store-ref qualified** exactly like
`makeOpenSessionStoreRefIndex`, because pool seat identities recur across rig
stores (#3453): a stranded seat may only disqualify claims in the stores it can
reach, never a same-named live seat's claim in another store.
`exitedDrainAckHolderOwnsWork` is the fail-**closed** inverse of
`openSessionOwnsWork` — there an unresolved or cross-store reachability answer
means "assume the session owns the work" because the destructive direction is
releasing a live claim; here the answer drives a release, so the same ambiguity
means "not stranded, keep the claim". Only an exact store-ref hit bypasses the
gate. The stranded branch also still runs `assigneePreservesNamedSessionRoute`:
the session-side exclusions filter the marker-carrying session, not the assignee
string, so a named or graph-run route colliding on that string keeps its claim.

The marker cannot outlive the state it describes: the write path clears it on
every clean drain-ack, `PreWakePatch` clears it on every start, and the read path
ignores it on any non-asleep seat. `PreWakePatch` is the right fence rather than
`ConfirmStartedPatch` / `CommitStartedPatch` because those are *post-start
confirmation* and are not reached when a start fails midway — the marker would
survive a failed start on a seat the controller is actively retrying.

`#2293`'s contract is preserved: the drain-ack path still never writes the work
bead.

### Why a new metadata key and not a `sleep_reason` value

The obvious cheaper shape is a new `SleepReason` — say
`SleepReasonDrainAckStranded` — instead of an eleventh metadata key. It was
considered and rejected for four reasons:

1. **`sleep_reason` is a decision input with an enumerated set, read by switches
   that must stay exhaustive.** `ClearWakeBlockersPatch`, `ClearExpiredHoldPatch`,
   and `ClearExpiredQuarantinePatch` each `switch` on it and clear it (or not)
   per value; status rendering and the wake gating read it too. A new value lands
   in every `default` branch of code that was never written to consider it,
   silently. Widening a value set that several independent classifiers switch on
   is a much larger blast radius than one additive key that exactly one new
   function reads.
2. **It is lossy where this fact must be durable.** `sleep_reason` is
   single-valued and rewritten by the *next* sleep, and `PreWakePatch` overwrites
   it with `input.SleepReason` on every wake. It also carries no instant, so
   there is no way to order or age a strand. `drain_ack_stranded_at` stores the
   RFC3339 moment, which is what a future stranded-duration metric or a staleness
   cutoff needs — without adding a second key later.
3. **The two facts are genuinely independent, and conflating them destroys
   information.** The seat *is* idle-asleep; that is why `CompleteDrainPatch`
   writes `SleepReasonIdle` and why the controller's idle handling is correct. It
   additionally stranded a claim. Encoding the second fact by overwriting the
   first makes "asleep because idle" unrepresentable for a stranded seat and
   changes what operators see in `gc status` for a reason unrelated to sleeping.
4. **It matches the established pattern for exactly this shape.** `detached_at`,
   `reset_committed_at`, `held_until`, `slept_at`, `awake_started_at` are all
   "durable fact + instant" keys that round-trip through `infoKeyCodec` and are
   cleared by the transition that ends them. `drain_ack_stranded_at` is the same
   pattern, cleared by `PreWakePatch`, with the same oracle coverage.

The cost of the key is bounded and mechanical: one `infoKeySpec` row, one `Info`
field, and two lines in the two hand-maintained exhaustiveness oracles
(`allProjectedMetadataKeys`, `infoFromPersistedBeadFrozen`) that exist precisely
so a new key cannot be added without being projected.

## Tests

All verified RED before the change, GREEN after:

* `TestReleaseOrphanedPoolAssignments_ReopensDrainAckStrandedHolderWork` — the
  regression.
* `…_SkipsIdleAsleepHolderWithoutStrandedMarker` — an asleep holder with no
  marker is still wakeable and must not be reset.
* `…_SkipsStrandedMarkerOnNonAsleepHolder` — a marker on an
  `active` / `creating` / `start_pending` seat releases nothing.
* `TestReconcileSessionBeads_DrainAckStrandedMarkerLifecycle` — three subtests
  driven end to end through `reconcileSessionBeads` +
  `reconcileStopPendingToTerminal` (not through the patch constructors):
  drain-ack **with** assigned work sets the marker and it equals `slept_at`;
  drain-ack with **no** assigned work clears a pre-existing one; a start clears
  it.

The pinned #2293 contract test
`TestReconcileSessionBeads_DrainAckMidPhaseEmitsAssignedWorkEvent` passes
unchanged — the SDK still does not touch the assignee on the drain-ack path.

```
go test ./internal/session -count=1                                                              # PASS
go test ./cmd/gc -run 'ReleaseOrphanedPoolAssignments|DrainAck|Reconciler.*Drain|PreWake' -count=1  # PASS
go vet ./cmd/gc ./internal/session                                                               # PASS
gofmt -l <touched files>                                                                          # clean
```

## Known window

After a reopen the marker stays set on the still-asleep seat until it wakes
(`PreWakePatch`) or drain-acks cleanly. That is harmless today because a pool
claim is only ever stamped by a running — therefore already woken — agent, so no
fresh claim can land on a seat still carrying the marker. Worth knowing if a
pre-wake assignment path is ever introduced; clearing the marker in the release
path would close it.

## Related

#5473 (the zero-grace `no_work` drain-ack that *creates* this state; this PR is
the recovery gap once it exists) and #2293 (the drain-ack path never writes the
work bead — preserved here).


jacobhausler:
Field follow-up (cr-h3evor, 2026-09-05): pool aliases are reusable, so the orphan reaper must key liveness on the durable claim-time gc.session_id when present, falling back to the assignee only for legacy claims.

The branch now also covers:
- a closed original claimer whose alias is reused by a live successor: stale work reopens;
- a live exact claimer whose assignee spelling is stale: the claim is retained;
- the stranded-holder bypass added by this PR sharing an alias with a live successor: only the exact stranded claimer may bypass liveness.

Validation at merged tip 622ae07b0ad68f70a7771db8d30d52170a4d4552:
- focused orphan-release suite PASS, Nomad job builders/dispatch-1788651798-84c095ee
- make build PASS, builders/dispatch-1788651817-d1107198
- race-enabled focused suite PASS, builders/dispatch-1788652404-85cf571a
- make check: fmt, lint, vet, and release checks PASS; aggregate tests baseline-only red. Every failing test was reproduced unchanged on clean current main c9b1eaebf: container process-reaping fixtures, pack-v2 doctor fixture, and SQL fixture.

No config, new store write, timeout, goroutine, or public API surface is added. Existing conditional release/readback and store-ref qualification remain in force.

A3Ackerman:
Prevention-side companion to this one: #6168 keeps a live pool seat that still holds assigned work out of the reconciler's own timer drains (no-wake-reason, idle-sleep, config-suppressed), so the seat is never asked to drain-ack mid-molecule in the first place; explicit intents still drain it. It sits in the `!shouldWake && alive` arm of `reconcileSessionBeads`, not in `finalizeDrainAckStoppedSession`, so the two don't share a hunk — yours recovers what already stranded, mine stops the next one. On one review plane we measured eight such drain-acks on a single orchestrator in 85 minutes before the gate. Whichever lands second rebases; happy to go after yours.

jacobhausler:
Rebased onto current main (2026-09-14). The four red checks (CI/required, CI/integration, Integration/rest-smoke-2-of-2, Evidence/critical-path) are the bd schema-migrate CI race tracked in #5804: this branch was fully green at 622ae07b on 2026-09-06 and is red at 07380a39 with no pool-code change in between, and the failing shards are the ones that lose the migrate race. `BD_ALLOW_REMOTE_MIGRATE=1` is deliberately not set in CI, so there is no in-scope fix for this PR to carry; a maintainer re-run should clear it. The drain-ack change itself is not implicated by any of the four. Related: #6128, #6119, #5348.

jacobhausler:
Could a maintainer please re-run **`Integration / rest-smoke-2-of-2`** (job `103906514515`) on run `34822303580`? We cannot trigger it from our account — we have no triage/write access on this repo, so a re-run is only reachable from the outside.

Why we read it as a harness flake rather than this branch:

- It is the **only** failing leaf job on the run (48 success / 9 skipped / 3 failure, and the other two failures are its own parents, `CI / integration` and `CI / required`). `Integration / rest-smoke-1-of-2` passes on the same head `07380a39d`.
- The failure is `graph_dispatch_test.go:62` — `gc init --file` fails with **"refusing to auto-apply 17 pending schema migrations to a shared server database"** (v49 -> v66). That is the shared CI server database declining a schema migration, which is not a path this PR's pool drain-ack reopen logic touches.
- `gastownhall/beads#5920` is still open against that migration-refusal behaviour itself, and `gastownhall/gascity#5804` is open as the flake note for this suite.
- The head has not moved since that run, so the result cannot have been re-derived by a newer push on our side.

If a re-run goes green we are ready for review; if it fails a second time we will treat it as real and dig in from our side. Thanks!

Issue #5650, opened 2026-08-26
bug(events): activeScanStart can silently discard an entire active log on a non-monotonic sequence


sjarmak:
Split out of the investigation recorded in #5589, so it does not evaporate with a merged PR's comment thread.

## The false claim

`activeScanStart` in `internal/events/reader.go` documents itself as degrading "in speed, never in correctness" on a log that violates the sequence ordering invariant. That is not true. The function drives `sort.Search`, which requires a predicate that is false for a prefix and true for the remainder. The predicate is:

```go
seq, _, ok := seqLineAt(f, int64(i))
return !ok || seq > afterSeq
```

On a non-monotonic active log that predicate is not monotone, the search returns a meaningless index, and `ReadFiltered` silently drops events:

```
seqs=[1 2 3 100 4 5 6]  after=4:  got [5 6]  want [100 5 6]
seqs=[1 2 3 100 4 5 6]  after=6:  got []     want [100]
seqs=[10 11 12 1 2 3]   after=10: got []     want [11 12]
seqs=[10 11 12 1 2 3]   after=11: got []     want [12]
```

The second shape is the damaging one. One out-of-order run near the front of the file makes every probe return false, `sort.Search` returns `size`, and the `seq <= afterSeq` branch treats that as "cursor is at or beyond the newest event" and skips the file entirely. For that cursor the dropped events are gone permanently.

These cases pass on `3e686dace`, the commit before the seek optimization, where the full scan has no ordering assumption to violate. So this is a regression, not a pre-existing condition.

## Reachability

Low, which is why this is a hardening ticket rather than a revert. `writeRecordLocked` re-reads the on-disk tail inside a cross-process `flock` and only ever raises `r.seq` (`recorder.go:377-386`, `:246-259`), and rotation continues the sequence space through the `events.rotated` anchor rather than restarting it (`recorder.go:461-552`). That leaves two ways in:

1. **A torn or short write.** `writeRecordLocked` (`recorder.go:394-398`) does not check for a short write. `writeBatch` (`:364-370`) does. That asymmetry looks unintentional.
2. **A stale post-rotation file descriptor in a second process.**

## Two fixes, independently useful

- Check the short write in `writeRecordLocked`, matching what `writeBatch` already does.
[...]

PR #5748, opened 2026-08-29
fix(events): activeScanStart falls back to a full scan on a reversed-seq log (#5650)

rjgeng:
Fixes the reachable half of #5650, and corrects the false claim in `activeScanStart`'s own doc comment that the issue leads with.

**Scope, stated up front: this fixes two of the four cases in the issue and deliberately leaves the other two, with a test pinning them.**

| repro from the issue | before | after |
|---|---|---|
| `[10 11 12 1 2 3]` `after=10`, `after=11` | drops events | **fixed** — full scan |
| `[1 2 3 100 4 5 6]` `after=4`, `after=6` | drops events | still drops — see below |

## What changed

`activeScanStart` drives `sort.Search`, which is only meaningful on a monotone predicate. It now refuses the fast path when the file's own head and tail contradict the non-decreasing-seq assumption — `tailSeq < first` means a reversed log, so it returns 0 and the caller does a full scan. There is a second guard on the way out: if the search converges on "nothing left to see" while the file's actual last record is above the cursor, that contradiction also forces a full scan.

That covers the shape your own reachability analysis identified — a stale post-rotation writer appending seq below the file's head.

## What it does not cover, and why I am not claiming otherwise

An out-of-order run that both starts and ends *within* the range head and tail already imply is invisible to any sub-linear check: in `[1 2 3 100 4 5 6]` the head is 1 and the tail is 6, which looks perfectly healthy. Catching that requires giving up the sub-linear property the optimization exists for, which is a different change with a performance argument attached, so I did not fold it in here.

Rather than leave that implicit, it is pinned: `TestActiveScanStartInteriorSpikeResidualGap` runs the issue's own `[1 2 3 100 4 5 6]` / `after=6` repro, asserts the current (wrong) empty result, and says in the failure message that a future stronger fix should flip the assertion rather than delete the test. So the gap is a documented, tested boundary rather than a silent one — and if someone closes it, the test tells them exactly what to do.

The doc comment on `activeScanStart` is updated to match: the old "degrades in speed, never in correctness" line — the false claim the issue opens with — is replaced by an accurate statement of what the check catches and what it cannot.

## Also here: the `writeRecordLocked` short-write asymmetry

You flagged this in your review of #5614 — `writeBatch` checks for a short write and `writeRecordLocked` does not — and called it "plausibly one source of the corruption being repaired here." Agreed, and it is the same package and the same failure class, so it is in this commit rather than a third PR: `writeRecordLocked` now compares the byte count and returns `errors.Join(err, io.ErrShortWrite)` on a partial write instead of dropping it. `TestFileRecorderWriteRecordLockedDetectsShortWrite` covers it.

If you would rather see that split out, say so and I will separate it.

## Tests

- `TestActiveScanStartNonMonotonicReversal` — the fixed class, from the issue's repro
- `TestActiveScanStartInteriorSpikeResidualGap` — the residual class, pinned
- `TestFileRecorderWriteRecordLockedDetectsShortWrite` — the short-write guard

`go vet ./...` clean whole-repo (untagged), `go test ./internal/events/...` passes, `gofmt` clean.

### DIFF
--- a/internal/events/reader.go
+++ b/internal/events/reader.go
@@ -159,14 +159,25 @@ func seqLineAt(f *os.File, off int64) (seq uint64, start int64, ok bool) {
 // it only after each line has been unmarshalled -- the dominant cost of the
 // supervisor's per-tick order-trigger check (gcy-ocb5).
 //
-// Returns 0 (full scan) whenever the boundary cannot be established, so a log
-// that violates the ordering invariant degrades in speed, never in correctness.
+// Returns 0 (full scan) whenever the boundary cannot be established, or
+// whenever the file's own head and tail contradict the non-decreasing-seq
+// assumption sort.Search depends on, so a log that violates the ordering
+// invariant this way degrades in speed rather than silently dropping events.
+// That check is necessarily partial: it catches a reversed tail (e.g. a stale
+// post-rotation writer appending seq below the file's head) but, like any
+// sub-linear check, cannot see an out-of-order run that both starts and ends
+// within the seq range the head and tail already imply.
 func activeScanStart(f *os.File, size int64, afterSeq uint64) int64 {
 	if afterSeq == 0 || size <= 0 {
 		return 0
 	}
 	// Nothing to skip when the log's first line is already above the cursor.
-	if first, _, ok := seqLineAt(f, 0); !ok || first > afterSeq {
+	first, _, ok := seqLineAt(f, 0)
+	if !ok || first > afterSeq {
+		return 0
+	}
+	tailSeq, err := readLatestSeqFromTail(f, size)
+	if err != nil || tailSeq < first {
 		return 0
 	}
 	i := sort.Search(int(size), func(i int) bool {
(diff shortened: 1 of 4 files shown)


quad341 (approved):
@rjgeng — this is the kind of PR that makes review easy, and I want to name why rather than just merge it.

You could have shipped the `activeScanStart` guard and let the doc comment keep implying the whole class was handled. Instead you fixed the reachable half, said plainly in the body that two of the four repro cases still drop events, and then *pinned the gap with a test* whose failure message tells the next person to flip the assertion rather than delete it. The reasoning about why an interior spike is invisible to any sub-linear check is correct, and replacing that false "never in correctness" line with an accurate one is worth as much as the code.

You were right to fold the `writeRecordLocked` guard in here — same package, same failure class, and #5650 names that asymmetry itself. Keep it. I checked, and it does real work: `os.File.Write` drops its own `ErrShortWrite` when the OS also returns an error, so that partial write really was unclassifiable before you touched it.

One thing on my side: our Mac tier is opt-in and didn't run, so I'll add a skip to your new test for the one kernel behavior that could red it there. Purely mechanical. Merging with that, and keeping #5650 open for the interior-spike half with your test referenced.


---

PR #5749, opened 2026-08-29
fix(dolt): defer proven same-count writer races during compact

quad341:
## Summary

- treat same-count value-hash drift as a concurrent writer race only when the existing `DOLT_DIFF` proof shows additive-only changes
- fail closed to quarantine/manual review when that proof fails or any row gain is present
- extend the compact-script mock and regression coverage for both outcomes

Refs `ga-i61ym3`, `ga-h0bj7x`, and `ga-loyfg8`.

Deploy evidence: [`release-gates/ga-i61ym3-dolt-compact-same-count-writer-race-gate.md`](https://github.com/quad341/gascity/blob/deploy/ga-i61ym3-gate/release-gates/ga-i61ym3-dolt-compact-same-count-writer-race-gate.md)

## Testing

- [x] `go test -count=1 -v ./examples/bd/dolt/...` — 289 PASS, 0 FAIL, 0 SKIP
- [x] `make test-local-full-parallel` — 37/40 jobs PASS; three predating, structurally unrelated failures tracked as `ga-f0uceo`, `ga-idsv6m`, and `ga-yc0e3a`
- [x] `make test-ci-policy`
- [x] `make check-docs`
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `make test-native-doltlite-beads`
- [x] pre-push `make test-fast-parallel` — all 10 jobs passed

`make check-native-dependency-surface` preserves the known host-dependent old measurement failure tracked by `ga-iuznq2`; reviewed deploy bead `ga-qma9li` carries its isolated fix. This PR changes no binary dependency path.

## Checklist

- [x] Linked the tracked work
- [x] Added regression coverage for the behavior change
- [x] No user-facing documentation change is required

---
- Related to #3341 — hardens the dolt compact concurrent-writer handling described in that issue: same-count value-hash drift is deferred only when a DOLT_DIFF proof shows the flatten itself touched no rows, and fails closed to quarantine/manual review otherwise; it does not change the quarantine marker's alerting or auto-staleness, nor the backup path that fails while a database is quarantined

### DIFF
--- a/examples/bd/dolt/commands/compact/run.sh
+++ b/examples/bd/dolt/commands/compact/run.sh
@@ -2856,6 +2856,7 @@ flatten_database() {
     # unproven HEAD movement fails closed and falls through to quarantine.
     if [ "$writer_race_detected" = "1" ] && \
        [ "${verify_counts_saw_same_count_hash_drift:-0}" = "1" ] && \
+       [ "${verify_counts_saw_gain:-0}" != "1" ] && \
        [ "${verify_counts_saw_gain_hash_drift:-0}" != "1" ] && \
        [ "${verify_counts_saw_row_decrease:-0}" != "1" ] && \
        [ "${verify_counts_saw_table_list_change:-0}" != "1" ] && \
--- a/examples/bd/dolt/commands/compact/run.sh
+++ b/examples/bd/dolt/commands/compact/run.sh
@@ -1173,6 +1173,7 @@ verify_counts() {
   verify_counts_saw_row_decrease=0
   verify_counts_saw_decrease_hash_drift=0
   verify_counts_saw_same_count_hash_drift=0
+  verify_counts_same_count_drift_tables=""
   verify_counts_saw_table_list_change=0
   verify_counts_saw_probe_failure=0
   verify_counts_failure_reason=""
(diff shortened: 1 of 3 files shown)


quad341:
@quad341 — this is the good kind of tightening. Swapping the same-count-drift downgrade from "HEAD moved, therefore writer" to "the `snapshot..flatten` diff proves the pre-flight rows survived" replaces an inference with an actual proof, and reusing `gain_drift_is_additive_only` instead of writing a second proof keeps fail-closed semantics in exactly one place. The restored gain-exclusion guard (`ga-loyfg8`) closes a real mixed-signal hole.

Special credit for `TestCompactScriptQuarantinesSameCountDriftWhenDiffProofFails`. Asserting that a failed proof leaves the quarantine marker *and* writes no pending-GC retry marker *and* never reaches `DOLT_GC` is the assertion that actually protects this gate — plenty of tests would have checked only the exit code. The first-match-wins comment on the new mock arms will save whoever adds the next `DOLT_DIFF` arm a genuinely miserable afternoon.

Merging as-is. I traced `gain_drift_is_additive_only`'s signature and fail-closed contract against your new call site, and CI ran real test evidence on this head rather than a wall of skips.

One follow-up I'll file rather than block on: an absorbed UPDATE (writer commit squashed into the flatten) plus an unrelated HEAD move now quarantines where it used to defer, since the additive-only proof can't pass for it. Fail-closed is the right default, but the busiest DB is where net-zero churn lives, so quarantine rates are worth a look.


---

PR #5750, opened 2026-08-29
fix(ci): make native dependency size guard deterministic

quad341:
## Summary

- build the native-dependency measurement binary with `CGO_ENABLED=0` and `-trimpath`, removing host C-toolchain and embedded-path variance
- keep the binary unstripped so the existing symbol and forbidden-literal scans remain effective
- re-baseline the byte ceiling from 270,000,000 to 180,000,000 with dated measurement and growth-rate evidence

Refs `ga-qma9li` and `ga-iuznq2`.

Deploy evidence: [`release-gates/ga-qma9li-native-dependency-surface-determinism-gate.md`](https://github.com/quad341/gascity/blob/deploy/ga-qma9li-gate/release-gates/ga-qma9li-native-dependency-surface-determinism-gate.md)

## ⚠️ Merge-order conflict with PR #5489

PR #5489 currently still contains commit `3c4090e81a` (`chore(ci): raise the native-surface byte tripwire to 285MB`). Its current head keeps the old host-dependent plain `go build` and changes this same ceiling to `285000000`.

**If #5489 lands after this PR, it will silently undo this deterministic measurement and replace the evidence-based `180000000` re-baseline with `285000000`.** Whoever merges or rebases either PR must preserve this PR's `CGO_ENABLED=0 go build -trimpath` invocation and 180,000,000-byte ceiling. Please do not allow #5489 to clobber them during conflict resolution or later sequencing.

## Testing

- [x] `make check-native-dependency-surface` twice — both PASS, byte-identical at 172,098,893 bytes; module counts 727/25/9/15/1
- [x] `bash -n scripts/check-native-dependency-surface.sh`
- [x] `shellcheck scripts/check-native-dependency-surface.sh`
- [x] `make test-fast-parallel` — all 10 jobs passed
- [x] `make test-acceptance` — Tier-A PR gate passed
- [x] `make test-ci-policy`
- [x] `make test-native-doltlite-beads`
- [x] `make check-docs`
- [x] changed formatting and affected lint — 0 issues
- [x] `go build ./...`
- [x] `go vet ./...`

## Checklist

- [x] Linked the tracked work
- [x] The existing native-surface guard is the direct regression proof; no test file changes are needed
- [x] No user-facing documentation change is required
- [x] The #5489 merge-order conflict is explicitly called out above

### DIFF
--- a/release-gates/ga-qma9li-native-dependency-surface-determinism-gate.md
+++ b/release-gates/ga-qma9li-native-dependency-surface-determinism-gate.md
@@ -1,0 +1,102 @@
+# Release gate: deterministic native dependency surface measurement
+
+- Deploy bead: `ga-qma9li`
+- Build bead: `ga-iuznq2`
+- Review bead: `ga-n8apor`
+- Reviewed source: `09139999930b1c2aa7d47df288d1f2e26f28b4eb`
+- Reviewed source branch: `builder/ga-iuznq2` (provenance only)
+- Deploy branch: `deploy/ga-qma9li-gate`
+- Base checked at gate time: `origin/main@52b5f4045c8c3e779daaf814b1fde7573f3af752`
+- Gate result: **PASS**
+
+## Gate criteria
+
+| # | Criterion | Result | Evidence |
+|---|---|---|---|
+| 1 | Review PASS present | **PASS** | Closed review bead `ga-n8apor` records `verdict: pass` for exact source `09139999930b1c2aa7d47df288d1f2e26f28b4eb`, with no uncovered criterion. |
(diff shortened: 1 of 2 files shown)


quad341:
@quad341 — thank you for chasing the root cause here instead of doing the easy thing. This guard has been breaching 270,000,000 for a while and getting waived gate after gate under `ga-5flk3r`; the tempting fix was one more ceiling bump, and you instead worked out *why* the number wasn't reproducible and made the measurement deterministic. That's the harder and much more useful change.

Two details I liked specifically: keeping the binary unstripped so the forbidden-symbol and literal scans downstream still work — easy to lose by accident once you start adding build flags — and re-baselining from a dated measurement with a growth rate and instructions for the next person, rather than picking a round number.

I checked `.goreleaser.yml` on my side: the release build is already `CGO_ENABLED=0`, so your change makes the guard measure the configuration we actually ship. And `Preflight / static checks` ran your modified script on a GitHub runner and cleared 180,000,000 — cross-host evidence beyond your local pair. Merging as-is.

The #5489 sequencing risk is ours to own; I'll make sure that one doesn't take back your invocation or ceiling during conflict resolution.


---

PR #5751, opened 2026-08-29
fix(execgrace): run TestMonitorOutputResetsIdle on a virtual clock

sjarmak:
`TestMonitorOutputResetsIdle` paced writes against a real 200ms idle window with a 50ms write cadence. Under load, a stall in the writer goroutine's own real-time scheduling could exceed the idle window with no actual output stall, so the monitor cancelled and the test failed with a spurious `ErrIdle`. It surfaced during full-suite runs, where the scheduler is contended.

The fix moves the test into a `testing/synctest` bubble. Virtual time only advances once every goroutine is durably blocked, so a slow scheduler can no longer let apparent silence elapse between writes. The assertions are unchanged: output past the idle window keeps the context alive, and silence afterwards still cancels with `ErrIdle`.

`t.Parallel()` stays on the outer test. Only the bubble's inner `*testing.T` forbids it, and the rest of the package is parallel.

No production code changes.

## Test plan

- `go test -race -count=5 ./internal/execgrace/` passes; the bubbled test is deterministic across runs.
- `go vet ./internal/execgrace/` and `make build` clean.
- Guard suites green: `go test ./internal/testpolicy/resourcecensus/ ./scripts/cipolicy/ ./internal/testenv/`, baseline pinned to `origin/main` at `8940ee38c`.
- Reviewed the monitor for anything that would not participate in a bubble. It uses `time.Now`, `time.Since` and `time.AfterFunc` (`internal/execgrace/execgrace.go:130,132,190`), with atomics, context cancellation and an uncontended mutex otherwise, and the test writes only to `io.Discard`. Nothing there can prevent durable blocking.

One coverage boundary worth naming: the activity refresh is synchronous today (`execgrace.go:208`). If it ever became asynchronous, this test would pass under virtual time while starving under real load. The original timing-based test was not a reliable load regression test either, so this is not a regression in coverage, but it is the assumption the deterministic version rests on.

### DIFF
--- a/internal/execgrace/execgrace_test.go
+++ b/internal/execgrace/execgrace_test.go
@@ -55,6 +55,7 @@ func TestMonitorIdleTimeout(t *testing.T) {
 // deadline with no actual stall, producing a spurious ErrIdle. The bubble
 // removes that dependence on real elapsed time and OS scheduling entirely.
 func TestMonitorOutputResetsIdle(t *testing.T) {
+	t.Parallel()
 	synctest.Test(t, func(t *testing.T) {
 		m := NewMonitor(context.Background(), 200*time.Millisecond, 0)
 		defer m.Stop()
--- a/internal/execgrace/execgrace_test.go
+++ b/internal/execgrace/execgrace_test.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"io"
 	"testing"
+	"testing/synctest"
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


quad341 (approved):
@sjarmak — this is a model flake fix. You didn't reach for a bigger timeout; you worked out that the test was measuring the scheduler rather than the monitor, and moved it onto a clock where that confusion is structurally impossible. Merging as-is.

Three things I appreciated. You checked that `Monitor` can actually participate in a bubble instead of assuming it — the `time.Now`/`time.Since`/`time.AfterFunc` inventory with line numbers meant I could verify your reasoning in minutes rather than re-deriving it. You put the *why* in the comment, so the next person doesn't quietly unwind the bubble. And you named the coverage boundary around the synchronous activity refresh yourself, unprompted, including why it isn't a regression — reviewers usually have to dig that out.

I re-ran your test plan here: `-race -count=5` green, and `-race -cpu=1 -count=10` green under forced contention, which is the case that used to bite. The bubbled test at `-count=50` finishes in 3ms, so virtual time is clearly doing the work.

One idea, not a request: `TestMonitorCeiling` and `TestMonitorIdleTimeout` still ride real time. Same family, wider margins. Happy to file that chore unless you'd rather take it.


---

PR #4385, opened 2026-07-17
fix(gc): drain the --since window on gc events instead of capping at one page

jacobhausler:
## Problem

`gc events --since <window>` silently returns only the newest page (500 events),
under-reporting any window that holds more than one page. Three different window
sizes against a live store all return an identical hard 500:

```
gc events --since 1h   -> 500 rows
gc events --since 6h   -> 500 rows
gc events --since 24h  -> 500 rows   (24h span was 500 *contiguous* seqs — a tail, not a match set)
```

Any rate or volume derived from `--since` row counts under-reports every window
with >500 events, with no error or warning. (Reported downstream while building
event-rate alarms; a "397 bead.updated/6h" reading turned out to be a truncation
artifact.)

## Root cause — CLI, not the server

The city event list is keyset-paginated, seq-DESC: a truncated page mints
`next_cursor` pointing strictly below the page's oldest seq (#4194). But
`fetchCityEvents` (cmd/gc/cmd_events.go) issued a **single** `limit=500` request
and **dropped `next_cursor`**, so every window larger than one page was silently
truncated to the newest 500.

The server side is already correct and archive-aware post-#4194 (the tail fast
path falls through to the archive-aware sequential reader, and truncated pages
always carry `next_cursor`). This was purely the CLI never following that cursor.

`TestFetchCityEventsSinglePageChronological` documented the old single-page
behavior as deliberate — *"a full descending drain of a 100 MB+ event history
would blow the command timeout."* That reasoning holds for an **unbounded** (bare)
`gc events`, but a `--since` window is **bounded** and must be reported in full.

## Fix

- `fetchCityEvents` follows `next_cursor` to drain the full window **when
  `--since` is set** — bounded by the window; the #4194 keyset guarantees no
  skip/dup across page seams.
- The unbounded (no `--since`) fetch **stays single-page** (preserving the
  deliberate timeout-safety decision), but a present `next_cursor` now emits an
  explicit truncation notice on stderr instead of a silent drop:
  `gc events: showing the newest N events; older matching events were omitted.
  Use --since <duration> to fetch a full time window.`

## Verification

- **New regression test** `TestFetchCityEventsPaginatesSinceWindow`: a
  cursor-aware fake server (faithful to the #4194 keyset wire contract) serves
  1200 events across 3 pages (500/500/200); `--since` now returns all 1200,
  contiguous and ascending with no gaps/dups across page seams. Confirmed the
  test **fails without the fix** (got 500, want 1200).
[...]

Issue #1487, opened 2026-04-29
bug: gc events HTTP API on :8372 returns context-deadline-exceeded intermittently under load


jwp23:
## Problem

The controller's events HTTP server on port 8372 (`/v0/city/{cityName}/events`) returns transient `context-deadline-exceeded` errors under normal load. Pattern is shifting from "intermittent blip" to "sustained" in our installation: 5 timeouts in ~2.5 hours, 3 of them back-to-back on 2026-04-29.

Each call self-recovers within seconds. Beads/dolt traffic (port 3307) is unaffected — burns and writes succeed normally — so this is the controller HTTP server (8372), not the Dolt data plane.

## Operational impact

- Witness loses event-driven wakeup; falls back to deferred-reminder cycling (slower, less responsive).
- Same controller HTTP path is implicated in `gc mail inbox` / `gc mail reply` 120s timeouts observed under coincident load (3 in ~3 hours during a parallel events.jsonl growth incident — see #1118).

## Investigation directions

1. Profile controller HTTP server: goroutine count, request handler latency, blocking calls.
2. Check whether reconcile-loop work is starving the events handler — same goroutine pool? Backpressure?
3. Confirm the controller's events stream uses a streaming response (chunked / SSE) and isn't being held by a slow downstream consumer.
4. Determine whether `context-deadline-exceeded` is from a client default timeout being too short under load, or from an actual hang in the handler.

## Acceptance

Either the underlying flakiness is fixed (no more transient timeouts under normal load), or the witness's event-stream client uses a longer/retrying timeout policy that absorbs hiccups without falling back to deferred reminders. Pattern of 3+ back-to-back failures in a 30-minute window does not recur.

## Files

- `gc events --seq` / `--watch` CLI implementation
- Controller HTTP server handler for `/v0/city/{cityName}/events` (port 8372)
- Witness watcher loop

A3Ackerman:
We hit this same symptom on a multi-rig install. Posting root-cause notes in case useful.

**Not the events.jsonl file size.** Direct timing of `ReadFilteredTail` against a 408MB / 290k-line `events.jsonl`: 342µs for limit=1, 403µs for limit=100. The tail-read path is fast even at >4× the doctor warning threshold.

**Not the API server generally.** `/v0/cities` returns 200 in 16ms during the same window `/v0/city/{name}/events` hangs >60s.

[...]

PR #5752, opened 2026-08-29
fix(events): a wide --since window now truncates instead of failing with nothing

rjgeng:
On this city, `gc events --since=40h` exits 1 after ~32 seconds and prints nothing — after having already fetched roughly 52 pages and thrown all of them away. `doEvents` wraps the city-events call in a single 30-second `context.WithTimeout`, but `fetchCityEvents` walks *every* page of a `--since` window under that one context, so a budget sized for a single request is charged to an unbounded multi-page walk. #4385 introduced that walk — pagination is enabled solely by `--since` — and the budget was never revisited for it.

This splits the budget in two: a per-page timeout that bounds one request to a server that has stopped answering, and a whole-walk budget that bounds total runtime. Exhausting the walk budget is now a truncation rather than an error — the pages already fetched are returned newest-end-first and contiguous, with the same style of notice the single-page cap in that same function already prints. Running out of budget on the *first* page still fails, since there is nothing to show. On this city the same command now exits 0 with 31,500 events.

Not #1487 — that was intermittent, load-triggered, and self-recovering. This is deterministic and scales with the query window; a fix for one is not a fix for the other.

## The shape that shipped, and the one that did not

Rejected first attempt: move the deadline onto each page and let the walk run unbounded, on the reasoning that the walk is finite by construction. It is finite — but `--since=40h` then ran past five minutes without finishing. That trades a fast failure for a hang. Caught by running the patched binary against the live city instead of trusting the argument.

Shipped: two budgets, because they catch two different failures.
- `cityEventsPageTimeout` (30s) bounds one page request — a server that stopped answering.
- `cityEventsWalkBudget` (30s) bounds the whole walk — the ceiling on runtime.

## Measured on this live city

| window | stock `gc` 1.4.0 | patched |
|---|---|---|
| `--since=40h` | exit 1, 32s, **0 events** | exit 0, 36s, **31,500 events** + notice |
| `--since=2h` | exit 0, 25.5s | exit 0, 23s, 10,243 events, no notice |

## Tests

Three added. Two are true red-to-green witnesses, verified red by reverting the behavior with the tests in place:
- `TestFetchCityEventsPageBudgetBoundsASinglePage`
- `TestFetchCityEventsWalkBudgetTruncatesInsteadOfDiscarding`

The third, `TestFetchCityEventsPageBudgetDoesNotAccumulateAcrossWalk`, passes on unfixed code too — it's a forward guard against re-introducing an accumulating budget, not evidence of the bug.

`go vet ./...` clean whole-repo (untagged), `go test ./cmd/gc/ -run 'Events|Event'` passes, `gofmt` clean.

### DIFF
--- a/cmd/gc/cmd_events.go
+++ b/cmd/gc/cmd_events.go
@@ -564,10 +564,9 @@ func doEvents(scope eventsAPIScope, typeFilter, sinceFlag string, payloadMatch m
 		return 1
 	}
 
-	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
-	defer cancel()
-
 	if scope.isSupervisor() {
+		ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+		defer cancel()
 		items, err := fetchSupervisorEvents(ctx, client, typeFilter, sinceFlag)
 		if err != nil {
 			fmt.Fprintf(stderr, "gc events: %v\n", err) //nolint:errcheck
@@ -576,6 +575,13 @@ func doEvents(scope eventsAPIScope, typeFilter, sinceFlag string, payloadMatch m
 		items = filterSupervisorEvents(items, typeFilter, payloadMatch)
 		return printJSONLines(items, stdout, stderr)
 	}
+
+	// The city list drains a --since window across as many pages as it takes.
+	// The walk stays bounded in aggregate -- an unbounded walk would trade a
+	// fast failure for a hang -- but running out of budget now truncates the
+	// window and says so, instead of discarding every page already fetched.
+	ctx, cancel := context.WithTimeout(context.Background(), cityEventsWalkBudget)
+	defer cancel()
 
 	items, err := fetchCityEvents(ctx, client, scope.cityName, typeFilter, sinceFlag, stderr)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


quad341 (approved):
@rjgeng — thank you for this. Two things stand out beyond the fix itself.

First, you documented the shape that *didn't* ship. "Move the deadline onto each page and let the walk run unbounded" is the obvious fix, and you caught that it turns a 32-second failure into a five-minute hang by running the patched binary against a live city instead of trusting the argument. Writing that into the PR body saves the next person from rediscovering it.

Second, you told us which of your three tests is a real red-to-green witness and which is a forward guard. That's rarer than it should be, and it's accurate — `TestFetchCityEventsPageBudgetDoesNotAccumulateAcrossWalk` does pass unfixed. Ruling out #1487 on evidence rather than vibes was right too.

We're taking this with one small edit on our side. The budget comment says the two budgets catch two different failures, but at the shipped 30s/30s the page context derives from the walk context, so the page bound can't fire first — the code is right, the comment is just a step ahead of the values. I'll reword it, widen the walk test's timing margin for busy runners, re-run `cmd/gc`, and merge. The fix and the credit are yours.


---

Issue #3142, opened 2026-06-06
process: randy marking issues as "needs-repro" without commentary; issue auto-closed 14 days later


csells:
[this issue](https://github.com/gastownhall/gascity/issues/2201) is an example of randy marking an issue as "needs-repro" (great) without posting a comment, so the author never sees that their issue needs more data to make it actionable (bad). then, 14 days later, [it's auto-closed](https://github.com/gastownhall/gascity/issues/2201#issuecomment-4582606223) (ugly).

that's not the right way to actually gather the info needed to repro an issue nor is it the way to deal with someone that's logged a bug (feedback is a gift!).

minimally, randy would post a comment for the author when marking an issue with "needs-repro" along with a warning about the 14 day policy.

ideally, randy would actually do the work to attempt the repro and post that data to the bug himself. especially with a big with the kind of impact that https://github.com/gastownhall/gascity/issues/2201 has and the number of "me too" comments it had gathered.

PR #5755, opened 2026-08-30
docs: what a contributor is owed on filing, and how credit is recorded

sjarmak:
Contributors who file here mostly do not get an answer. Measured against the
live API on 2026-08-30: 191 of 634 open issues have no comment of any kind, and
412 of 446 open pull requests carry no review decision. A classification pass
over 425 of those PRs put 96 in the category "the underlying problem is real and
unfixed on main, but the branch has drifted too far to rebase cheaply."

That silence has a compounding cost, and it is visible in the queue. Four
separate contributors opened PRs against the same gap in Darwin `pidutil`
fail-closed coverage (#5254, #5291, #5345, #5746), three of them landing on
materially the same PID-1 probe, because none of the earlier ones got an answer
saying the gap was already covered.

Reviewing every PR is not achievable and is not what this proposes. Answering
is. This adds `engdocs/contributors/contributor-response-conventions.md`, which
writes down:

- what a contributor is owed on filing, and from whom;
- the five dispositions a PR can receive, as a table, with the conditions that
  select each one;
- how authorship is preserved when we supersede a PR, adopt and finish one, or
  close a duplicate;
- which of those an agent may do on its own, reconciled against the autonomy
  boundary already in `AGENTS.md`.

The duplicate-survivor rule is the part most likely to be applied wrong, so it
is stated as eligibility rather than as a tiebreaker: a behavior change carries
regression coverage, which both `AGENTS.md` and `CONTRIBUTING.md` already
require. An untested diff does not become the survivor by being smaller. If the
only tested candidate is the larger one, that one survives; if none is tested,
the cluster is not ready to resolve.

Docs only. No Go changes. `make check-docs` is green.

---
- Related to #3142 — writes down the convention that every filed issue is owed a written reply within 14 days and that ending one without a stated reason is not acceptable — the norm that issue describes being skipped; it is docs only and does not change randy's needs-repro labeling or the 14-day auto-timeout itself

### DIFF
--- a/engdocs/contributors/contributor-response-conventions.md
+++ b/engdocs/contributors/contributor-response-conventions.md
@@ -41,10 +41,12 @@ has not, and neither has the next person who rewrites it.
 
 The rules below describe what a **maintainer** owes a contributor. An agent
 seat may draft any of these replies and may gather the evidence a verdict rests
-on, but posting a comment, closing an issue or PR, pushing, and merging are
-external actions under the autonomy boundary in `AGENTS.md`. They require
-explicit per-action approval. A verdict existing is not authorization to
-publish it.
+on, but posting a comment, closing an issue or PR, pushing to someone else's
+branch, and merging are external actions that reach a contributor. They require
+explicit per-action authorization from the active user, repository, or
+orchestrator instructions; a verdict existing is not authorization to publish
+it. This is distinct from the session-close push in `AGENTS.md`, which governs
+an agent's own work branch and not someone else's PR.
 
 Where an acknowledgement is agent-drafted and not yet reviewed by a human, it
 says so rather than implying a person has read the code.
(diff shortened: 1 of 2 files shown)


quad341 (approved):
@sjarmak — you measured the thing instead of asserting it. 191 unanswered issues, 412 PRs with no review decision, the queries shipped inline so anyone can re-run them, and the four-way pidutil pileup (#5254, #5291, #5345, #5746) as the concrete cost. That's what makes this land as a convention rather than a wish.

The duplicate-survivor rule stated as *eligibility* rather than a tiebreaker is the sharpest part. "An untested diff does not become the survivor by being smaller" closes a loophole I've watched people walk straight through.

Two things I'll fix on my side rather than send back to you. The doc cites an "autonomy boundary in `AGENTS.md`" — I grepped the tree and that section isn't there; AGENTS.md's actual push language mandates the session-close push, so a reader following the pointer gets the opposite signal. I'll state the rule directly and note the distinction. And #1732's closing note actually says the branch "did not allow maintainer edits" — the same cause you name a sentence earlier — so I'll drop "for a different reason."

Rewording both, re-running `make check-docs`, merging under your name.


---

PR #5756, opened 2026-08-30
fix(events): fail a short single-record write instead of counting it as written

rjgeng:
`internal/events/recorder.go` has two write paths and only one checks its byte count. `writeBatch` returns `io.ErrShortWrite` when it writes fewer bytes than it was given. `writeRecordLocked` — the single-record path behind `Record()` and the rotation anchor write — discarded the count and kept only the error, so a short write that returned a nil error committed a partial line, incremented `recordCount`, and left the already-bumped `seq` in place. The recorder treated it as a complete record and appended everything after it behind the fragment.

Nothing detected that and nothing reported it. A partial line is indistinguishable from a torn write to every downstream reader, which is the corrupted-tail shape the recorder's startup truncation has to repair — so this and #5614 compose: truncation removes the damage, this stops the single-record path from producing it silently.

**Credit:** @sjarmak spotted this asymmetry while reviewing #5614 and noted it was "plausibly one source of the corruption being repaired here." That is exactly what it looks like from here.

**Fix:** route the single-record write through `writeBatch` so both paths get the guard from one implementation rather than two copies of the same check. The marshal-and-write step moves into `encodeAndWriteRecord`, which takes an `io.Writer` — `FileRecorder.file` is a concrete `*os.File`, so there was previously no way to inject a short writer and cover this path at all. The new test reuses the `shortBatchWriter` helper already in the package.

**Testing:** full `./internal/events` package green; `gofmt` and `golangci-lint --new-from-rev=origin/main --whole-files` clean; pre-commit hook passed in full. Red-green confirmed — with the guard reverted the new test fails with `error = <nil>, want io.ErrShortWrite`.

Does not conflict with #5614; verified by test merge, either order works.

### DIFF
--- a/internal/events/recorder.go
+++ b/internal/events/recorder.go
@@ -385,15 +385,31 @@ func (r *FileRecorder) writeRecordLocked(e *Event) error {
 	if e.Ts.IsZero() {
 		e.Ts = time.Now()
 	}
+	if err := encodeAndWriteRecord(r.file, e); err != nil {
+		return err
+	}
+	r.recordCount++
+	return nil
+}
+
+// encodeAndWriteRecord marshals e as one JSONL line and writes it to w,
+// treating a short write as an error rather than a success.
+//
+// The guard matters because a partial line is indistinguishable from a
+// torn write to every downstream reader: the recorder would otherwise
+// count the record as written and continue appending after the fragment,
+// which is the corrupted-tail shape that startup truncation has to repair.
+// Sharing writeBatch keeps the single-record and batch paths honest about
+// short writes in the same way, from one implementation.
+func encodeAndWriteRecord(w io.Writer, e *Event) error {
 	data, err := json.Marshal(e)
 	if err != nil {
 		return fmt.Errorf("marshal: %w", err)
 	}
 	data = append(data, '\n')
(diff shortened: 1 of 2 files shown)


quad341 (approved):
Thanks @rjgeng — routing the single-record path through `writeBatch` instead of leaving two copies of the same guard to drift apart is the right instinct, and the `shortBatchWriter` test is the part I'm happiest to have. The existing `RLIMIT_FSIZE` test `t.Skip`s on any kernel that honors the full write, so this path's coverage could quietly evaporate depending on where CI ran. An injectable writer fixes that properly, on every platform.

One heads-up, and it's on us for the three-week latency rather than on you: `main` picked up the byte-count guard while this sat, so the pre-image already has the check your description targets. What lands here is the consolidation plus the better test — still worth merging, just not the bug fix it was the day you wrote it. I'll reword the commit message on my side so the history doesn't claim a fix that's already in, and leave your code exactly as-is.

The two red shards are `proctable` and `productmetrics`; neither imports `internal/events`. Not yours — I'll rerun them. I ran `./internal/events/` locally and it's green.

Good catch by @sjarmak on the original asymmetry, too. Merging once those shards are clean.


---

PR #5757, opened 2026-08-30
fix(hook): read each ephemeral scan once per query, not once per identity (#5712)

### LINKED ISSUE #5712: bug(hook): work query cannot finish inside its own 60s deadline — two identity-independent --limit=0 full-store scans re-run per identity (root cause of #5693)
Root-cause trace for #5693, from an independent reproduction. #5693's Ask was "instrument/trace the hook-claim path (where do the 100s+ hangs spend time when Dolt answers in <200ms?)" — this is that trace.

## Environment

- gc v1.4.0 (pinned); root cause also present on main @ 5f3d62413
- bd 1.1.0 (Homebrew); `beads.bd_compatibility` UNSET → normalizes to `"bd-1.0.4"` (config.go:1432-1441), which is the widest form of the query
- macOS (darwin 24.6.0), single-store Dolt city, non-federated work query
- Fleet scale: multi-rig city, pool workers; store grown to the point where an unbounded scan costs 8–12s

## Symptom

The documented pool-worker startup protocol is the FIRST command a worker runs, and it fails:

```
$ gc hook --claim --drain-ack --json
gc hook --claim: running work query "sh -c '...'": timed out after 1m0s:
  context deadline exceeded
{"schema_version":"1","ok":false,"error":{"code":"command_failed",...}}
[exit 1]
```

Reproduced a second time with the read-only form (`gc hook`, no `--claim`): same 1m0s work-query timeout; the wrapper itself took a real 192s.

And the work was there. At that moment a work item (call it `wrk-4242`) was open, ready, and ASSIGNED to that very session (`assignee = $GC_SESSION_NAME`). A direct

```
bd ready --assignee=$GC_SESSION_NAME --json --limit=1
```

returned it in 1.75–4.26s. The work query never reached that branch before its deadline, so the protocol reported failure for work already assigned to the worker running it.

## Observed profile

Each stage of the generated work query, timed in order, same store, same session (a pool worker with session id `S`, session name `claude-S`, alias `act/claude-15`; 2026-08-27):

```
L1   bd list in_progress --assignee=S                     7.85s
L1   bd query ephemeral in_progress (S)                  10.27s   <-- unbounded
L1   bd list in_progress --assignee=claude-S              1.42s
L1   bd query ephemeral in_progress (claude-S)            8.32s   <-- unbounded
L1   bd list in_progress --assignee=act/claude-15         3.01s
L1   bd query ephemeral in_progress (act/claude-15)       7.79s   <-- unbounded
L2   bd ready --assignee=S                                3.46s
L2   bd query ephemeral open (S)                          3.81s   <-- unbounded
L2   bd ready --assignee=claude-S                         1.75s   <-- THE ANSWER
L2   bd query ephemeral open (claude-S)                   4.56s   <-- unbounded
L2   bd ready --assignee=act/claude-15                    2.76s
L2   bd query ephemeral open (act/claude-15)              5.47s   <-- unbounded
POOL bd ready gc.routed_to=act/claude                     3.50s
POOL bd ready gc.run_target legacy                        5.75s
POOL bd query ephemeral open (legacy)                    12.24s   <-- unbounded
TOTAL                                                    82.84s
```

82.84s against a 60s hard cap. Not marginal: the full traversal cannot finish, and which branch you reach depends on how loaded the store is.

## Source trace (file:line — verified on v1.4.0 and on main @ 5f3d62413)

**1. The unbounded scan itself** — internal/config/workquery.go:185-187:

```go
func bdQueryEphemeralStatusShell(status string) string {
    return `bd query --json ` +
        shellquote.Quote("ephemeral=true AND status="+status) + ` --limit=0`
}
```

`bd query --limit=0` is UNLIMITED (`bd query --help`: "0 = unlimited"), so each call is a full-store scan. In the profile above each returned `[]` (3 bytes) while costing 8–12s.

**2. It is run once per identity, inside the loop, and the identity filter is applied by jq AFTER the scan returns** — the query does not depend on the loop variable at all. internal/config/workquery.go:430-436 (main; v1.4.0 = same body, line 219):

```go
func ephemeralAssignedInProgressProbeScript(shellVar string, topo QueryTopology) string {
    _ = topo                                          // <-- version gate DISCARDED
    return `r=$(` + bdQueryEphemeralStatusQuietShell("in_progress") + ` | ` +
        `jq --arg id "$` + shellVar + `" '[.[] | select((.assignee // "") == $id)] | .[:1]' 2>/dev/null); ` +
        `[ -n "$r" ] && [ "$r" != "[]" ] && printf "%s" "$r" && exit 0; `
}
```

and internal/config/workquery.go:445-455 (`ephemeralAssignedReadyProbeScript`, same shape, gated — see 4). Both are emitted INSIDE the per-identity loops (workquery.go:311-320 `standardAssignedInProgressWorkQueryScript`, :388-396 `standardAssignedReadyWorkQueryScript`):

```
for id in "$GC_SESSION_ID" "$GC_SESSION_NAME" "$GC_ALIAS"; do ... done
```

Three identities →
- `ephemeral=true AND status=in_progress` ×3 → 10.27 + 8.32 + 7.79 = 26.38s
- `ephemeral=true AND status=open` ×3 → 3.81 + 4.56 + 5.47 = 13.84s
- plus one more `status=open` scan in the pool-demand tier: workquery.go:209-231 `legacyEphemeralPoolDemandShell` (reached from `poolDemandFirstRowFunctionScript`, :234-248) → 12.24s

TOTAL: 52.46s of 82.84s — **63% of the budget is two identity-independent scans executed 3× and 4×.**

**3. Golden files confirm the generated shape** (7 unbounded scans in the bd-1.0.4 form): internal/config/testdata/workquery/pool_Work_bd104.golden:1 contains 3× `ephemeral=true AND status=in_progress --limit=0` and 4× `ephemeral=true AND status=open --limit=0`.

**4. The existing version gate helps only half, and only if opted in.** `NormalizedBDCompatibility` (config.go:1432-1441): `""` and any unrecognized value both normalize to `"bd-1.0.4"` (the default). Under `bd_compatibility = "bd-1.0.5"` the four `status=open` scans do disappear (`ephemeralAssignedReadyProbeScript` returns `""`, workquery.go:446; `legacyEphemeralPoolDemandShell` returns `printf "[]"`, :210-212 — compare testdata/workquery/pool_Work_bd105.golden:1). That removes ~26s. But the three `status=in_progress` scans survive BOTH modes, because `ephemeralAssignedInProgressProbeScript` explicitly discards the gate (`_ = topo`, workquery.go:431). And since the default is bd-1.0.4, a city that has not opted in — ours, with bd 1.1.0 actually installed — gets all seven.

**5. The deadline, and the prior response to this same wall** — cmd/gc/cmd_hook.go:778-789:

```go
// ... The default work-probe issues ~6 sequential bd/store round-trips
// before the pool-demand tier that finds routed work; on a multi-rig dolt
// city under concurrent load the probe intermittently exceeded the prior
// 30s cap ... Raised to 60s to cover the realistic loaded cost. ...
// The package-level var lets us lower it again once the probe's
// round-trip count is reduced and the slow per-rig bd ready/gc ready
// paths are optimized.
var hookWorkQueryTimeout = 60 * time.Second
```

That 30s→60s raise is 8b7ad3e2b / PR #3728 ("fix(hook): raise work-query timeouts so pool operators aren't starved", 2026-06-27). The comment already names round-trip count as the thing that needs reducing; the profile above is the evidence that raising the ceiling again will not help — the cost is structural (unbounded scans × identities), not a constant. Note the count is understated: the comment says "~6 sequential round-trips"; the bd-1.0.4 form actually issues 15 (profiled above), 7 of them unbounded.

## Suggested direction

**A. Hoist the two scans out of their loops.** Run each ONCE, then filter per identity in jq against the already-fetched array. The jq filters are already written against the full array (`[.[] | select((.assignee // "") == $id)]`), so this is semantics-preserving:

```sh
EPH_IP=$(bd query --json 'ephemeral=true AND status=in_progress' --limit=0)
for id in "$GC_SESSION_ID" "$GC_SESSION_NAME" "$GC_ALIAS"; do
  [ -z "$id" ] && continue
  r=$(bd list --status in_progress --assignee="$id" --json --limit=1); ...
  r=$(printf '%s' "$EPH_IP" | jq --arg id "$id" '[.[] | select((.assignee // "") == $id)] | .[:1]')
done
```

and likewise ONE `EPH_OPEN` shared by loop 2 and the pool probe's legacy branch. Removes ~35s and brings the traversal to ~48s, inside the ceiling.

**B. Order the cheap probes first.** The `bd list`/`bd ready --assignee` probes are what actually find assigned work and cost 1.4–7.9s; the ordering that puts a 10s full-store scan BETWEEN them is what starves loop 2. Running all three identities' cheap assignee probes before any ephemeral scan would have found the assigned work in the first few seconds. (Independent of A, and cheaper.)

**C. Let the gate apply to the in_progress tier too.** Drop `_ = topo` at workquery.go:431 and give `ephemeralAssignedInProgressProbeScript` the same `if topo.includeEphemeralReady() { return "" }` treatment its ready-tier sibling already has, if `bd ready --include-ephemeral`'s coverage extends to the crash-recovery tier. That would make bd-1.0.5 cities scan-free here.

**D. Surface the timeout as a distinguishable, retryable condition at the protocol layer.** gc already does the right thing internally: `shellWorkQueryWithEnv` wraps `context.DeadlineExceeded` so callers can classify it as transient (cmd_hook.go:815-823), `claimHookWorkWithRunner` returns exit 1 / ok:false and emits a city work-query-failure event rather than draining (cmd_hook.go:649-656), and `writeHookClaimNoWork` is only reached after a clean exhaustion. What is missing is a MACHINE-DISTINGUISHABLE reason on the `--claim --drain-ack` JSON envelope: today the documented pool-worker protocol sees `ok:false` + exit 1 and cannot tell "the query did not finish" from "this store is unavailable", so a worker that treats any exit-1 as drain will exit with its own assigned work still open. A typed `error.code=work_query_timeout` (or a `retryable: true` field) would let the wrapper retry instead of guess.

## Impact

Any pool worker whose store has grown past this threshold follows the documented startup protocol and gets a hard failure — while holding assigned, ready work. In the observed session the worker only proceeded because the timeout was investigated by hand and the assignee probe was run directly.

## Relationship to existing issues (searched 2026-08-28)

- **#5693** (OPEN, 2026-08-27) "gc hook --claim hangs 100s+ and false-negatives across independent pools while Dolt is healthy" — SAME DEFECT, reported from the symptom side on Linux. This report is the requested trace, and it answers #5693's "why is Dolt healthy?": no single query is slow — fifteen sequential queries, seven of them unbounded, exceed the cap.
- **#5473** (OPEN, 2026-08-21) "gc hook --claim --drain-ack acks the no-work drain with zero grace right after a step close" — adjacent (drain-ack correctness), distinct cause; relevant to suggestion D only.
- **#3201** (CLOSED) "bug(order-dispatch): gc reload --soft/gc doctor still hang — N serial per-order last-run bd-queries" — precedent: same bug class (N serial bd queries under a cap), fixed upstream by removing the serial fan-out; supports A/B as the accepted shape of fix.
- **PR #3728** / 8b7ad3e2b (merged 2026-06-27) raised this very timeout 30s→60s for this very symptom, and its own comment defers the round-trip reduction; this report supplies the profile that says the reduction is now required.
- Post-v1.4.0 commits touching internal/config/workquery.go (#5167, #5161, #4952, #4726, #3963) and cmd/gc/cmd_hook.go (#5165, #5163, #5161, ...) — none removes the identity-independent unbounded scans; verified line-by-line against main @ 5f3d62413.

rjgeng:
Fixes #5712.

## The defect

The two ephemeral probes in the generated work query are emitted *inside* the
identity loops:

```sh
for id in "$GC_SESSION_ID" "$GC_SESSION_NAME" "$GC_ALIAS"; do ... done
```

and each opens with a full-store scan whose predicate never mentions the loop
variable:

```sh
bd query --json 'ephemeral=true AND status=in_progress' --limit=0
bd query --json 'ephemeral=true AND status=open' --limit=0
```

`--limit=0` is unlimited. The identity filter is applied by `jq` *afterwards*,
against the array that already came back — so the traversal is identical on
every iteration, and it ran once per identity: **3× per status** in
`standardAssignedInProgressWorkQueryScript` / `standardAssignedReadyWorkQueryScript`,
and **6×** in the nested `legacyControlAssigned*` control-dispatcher tiers.

On a store where one scan costs 8–12s that is most of `gc hook`'s 60s
work-query budget spent re-fetching an identical array. @nl0's profile in the
issue has the standard shape at **82.84s against the cap**, with a bead already
assigned to the calling worker sitting in a tier the query never reached.
Raising the ceiling does not help — `hookWorkQueryTimeout` was already taken
30s→60s in #3728, whose own comment defers exactly this round-trip reduction.

## The fix

Read each status once into a shell variable and filter that snapshot per
identity — which is what the `jq` filters were already written against, so the
selection semantics are unchanged.

The read is **memoized, not hoisted to a prelude**, and that is deliberate: it
keeps two existing properties exactly.

- A tier that serves a candidate and exits before reaching the probe still
  never scans.
- The reconciler's demand-detection form runs with every identity var empty, so
  it never enters the loop body — and therefore still never scans.

An unconditional prelude would have broken both.

The trade is that identities 2 and 3 read the snapshot taken while identity 1
was in scope rather than a fresh one. That is the same trade
`ephemeralAssignedReadyProbeScript` already makes internally between its fast
and slow filters, and a work query is a point-in-time probe the caller re-runs
regardless. A test pins that a bead assigned to the *third* identity is still
served from that shared snapshot.

## Scope

Assigned tiers only. `legacyEphemeralPoolDemandShell` runs the same scan and is
left alone on purpose: it is shared verbatim with `poolDemandCountShell`, the
reconciler's scale_check, and diverging the worker's claim decision from the
reconciler's spawn decision re-introduces the protocol-mismatch class
`engdocs/architecture/dispatch.md` warns about. Its scan is per-target, not
per-identity, so it is not the reported defect.

Suggestions B, C and D from the issue are left for separate changes: B changes
which candidate is served, C rests on an open question about
`bd ready --include-ephemeral`'s coverage of the crash-recovery tier, and D is
a protocol change in `cmd/gc`.

## One correction to the report

The issue cites `internal/config/testdata/workquery/pool_Work_bd104.golden:1`
as containing 3× the `in_progress` scan and 4× the `open` scan. On `main` it
contains **one** and **two** — the `for` loop is what multiplies them at
runtime.

That is precisely why no golden ever pinned this cost, and why the new tests
**execute** the generated script against a fake `bd` and count invocations
instead of asserting on its text. The runtime profile in the issue is sound;
only that citation is off.

## Tests

New `internal/config/workquery_ephemeral_scan_reuse_test.go` runs the generated
shell against a fake `bd` that logs every `query` it receives, via the package's
existing `runShellWithFakeBd` seam (no new subprocess call site).

Red-green — with the memo reverted:

```
ephemeral in_progress scan ran 3 times across three identities, want 1
ephemeral open scan ran 3 times across three identities, want 1
legacy-control in_progress scan ran 6 times, want 1
legacy-control open scan ran 6 times, want 1
```

Plus `TestEphemeralSnapshotStillMatchesLaterIdentities`, which fails any memo
that caches the first identity's *filtered result* instead of the raw scan.

30 golden files regenerated, one line each.

## Verification

- `go test ./internal/config/` — ok (full package), under both
  `-tags gms_pure_go` and real CGO
- `go test ./cmd/gc/ -run 'Hook|WorkQuery|PoolDemand|ScaleCheck'` — ok
- `go test ./internal/testpolicy/resourcecensus/` — ok
- `go build ./...` — whole repo, CGO
- `golangci-lint --new-from-rev=origin/main --whole-files ./internal/config/` — 0 issues
- Pre-commit hook passed in full; no `--no-verify`

Found and profiled by @nl0 (Alexei Mochalov) — the per-stage timings in #5712
are what made this a two-line fix instead of a hunt.


quad341 (approved):
@rjgeng — the diagnosis is what makes this one easy to say yes to. The two-line memo is the small part; noticing *why* thirty golden files never caught an 8–12s scan running three times is the real work. The generated text spells the scan once, the `for` loop multiplies it at runtime, and no text assertion could ever have seen that. Writing tests that execute the script against a fake `bd` and count invocations — through the existing `runShellWithFakeBd` seam, with no new subprocess call site — is exactly the right shape.

`TestEphemeralSnapshotStillMatchesLaterIdentities` is the test I'd have asked for. A memo caching the first identity's *filtered* result would sail through every count assertion and quietly strand a bead assigned to the third identity; you closed that door before anyone walked through it.

Two more things I appreciated: keeping the read lazy instead of hoisting to a prelude, with the comment spelling out which two properties that preserves — a tier that exits early still never scans, and the reconciler's demand-detection form still never scans. And leaving `legacyEphemeralPoolDemandShell` alone because it's shared with the reconciler's scale_check. Resisting the adjacent cleanup is the harder call and you got it right.

Merging as-is. And thanks @nl0 — the per-stage timings in #5712 are what turned this from a hunt into a two-line fix.


---

PR #5758, opened 2026-08-30
fix(dispatch): a preserve-only round is progress, not deadlock — scope abort no longer quarantines runs whose recovery path was correctly preserved

julianknutsen:
## Summary

`skipOpenScopeMembers` declared `unable to skip remaining scope members` whenever one sorted round produced zero skips — even when the round PRESERVED members, which is progress. Deterministic trigger: a preserved finalize scope-check that sorts after the dependents it blocks (the common case, since finalize beads are minted after their dependents). Specimen: beads#6038's run was quarantined at 04:26:11 on exactly this while its healthy twin closed every listed member at 04:26:12 — and the stale error text named the correctly-preserved member, misdirecting the initial diagnosis toward store failures.

**F1**: a zero-skip round with preservations continues; only a zero-skip zero-preserve round (a genuine blocks-cycle) errors. Termination stays provable: nothing joins `pending` after it is built and every continuing round strictly shrinks it (invariant stated in-code; the mutation probe that breaks it hangs the test, proving it load-bearing).

**F2**: the deadlock error names the *current* stuck members with their in-pending blockers, sorted.

## Review

Fable RCA (live store forensics on the specimen: all five "unskippable" members were closed one second after the quarantine) → implementation with a red-run confirming the exact incident message → adversarial review: **sound, zero findings** (termination, widened-success safety, true-cycle guard, error honesty, caller integration, test-order contract).

## Testing

Specimen pin (fails pre-fix with the incident message), true-cycle control, multi-round convergence, F2 content pin; full `internal/dispatch` package + `-race` green; 3/3 mutation probes bite.

Refs: ga-kzrlh

### DIFF
--- a/internal/dispatch/runtime.go
+++ b/internal/dispatch/runtime.go
@@ -802,16 +802,23 @@ func (s scopeSnapshot) skipOpenScopeMembers(store beads.Store, skipControlID str
 	}
 
 	skipped := 0
+	// Termination: nothing is ever added to pending below, and every round that
+	// does not return strictly shrinks it — a preserve deletes its member, a skip
+	// wave deletes every id it closed. So the round count is bounded by the
+	// initial member count, and only a round that neither skips nor preserves
+	// (a genuine blocks cycle among the pending members) is a real deadlock.
 	for len(pending) > 0 {
 		ids := sortedPendingIDs(pending)
 		depsByID, err := loadDownDepsForScopeSkip(store, ids)
 		if err != nil {
 			return skipped, err
 		}
+		preserved := 0
 		skippable := make([]string, 0, len(ids))
 		for _, id := range ids {
 			if preserveScopeCheckForSubject(pending[id], depsByID[id], skipControlID) {
 				delete(pending, id)
+				preserved++
 				continue
 			}
 			if !canSkipScopeMemberWithDeps(depsByID[id], pending) {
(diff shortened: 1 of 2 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277267726

quad341 (approved):
@julianknutsen — this is the kind of bug report I wish every bug report were. Catching that `skipOpenScopeMembers` mutates `pending` mid-pass while iterating a sorted snapshot, so a preserve lands *after* the members it would have unblocked, is genuinely subtle. Pinning it to the finalize-scope-check minting order — their ids sort after their dependents, so this is the common case, not an exotic one — is what turns a one-off quarantine into a named class of bug.

Merging as-is. I re-ran `internal/dispatch` on my side (green, and green under `-race`), and pulled the `preserved > 0` guard back out as a probe: your specimen test fails with the exact incident string. That's the proof I wanted that the tests are load-bearing rather than narrating the fix.

Two things I especially appreciated. The true-cycle control test, so the widened success path can't quietly paper over a real deadlock. And F2 — the stale error text naming a correctly-preserved member is what sent the original diagnosis hunting for store failures that never happened. Fixing the diagnostic alongside the bug saves the next person that hour.

Thanks for this one.


---

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR fixes a premature-deadlock bug in the dispatch scope-abort path: a round that only *preserved* a scope-check member (re-planning it for idempotent replay) without skipping any member was being mis-scored as "no progress" and aborting the whole scope, which quarantined runs whose recovery path had in fact been preserved correctly. It also sharpens the deadlock error so it names the members that are genuinely stuck. This matters because it stops the controller from tearing down runs that were actually making forward progress, and makes the remaining genuine-deadlock diagnostics actionable.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **Preserve-only round mis-scored as "no progress":** the fix adds a `preserved` counter and re-plans (`continue`) instead of erroring when a round preserved a member but skipped none, so a correctly-preserved scope-check no longer trips the deadlock path. Review verified that termination is preserved — nothing is added to `pending`, every preserve-only round permanently removes at least one member, so the loop stays bounded and the one remaining hard error is a genuine `blocks` cycle.
- **Deadlock error now names the actually-stuck members:** `describeStuckScopeMembers` renders the current pending set (using the same `blocks`/`DependsOnID` rule the skip decision uses, deterministically sorted) instead of the starting id list, so a real deadlock is debuggable.

- No non-blocking follow-ups.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `11ceaea4` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/33296902029

### Review Iterations
1 review pass performed (attempt 1); approved with no fix iterations required. The change is covered by 4 new regression tests in `internal/dispatch/runtime_test.go` (preserve-only round continues, a genuine `blocks` cycle still fails, the deadlock error names the current stuck members, and convergence over many rounds).

---

PR #5766, opened 2026-08-30
fix(session): defer orphan cleanup after partial scan

allenday:
## Summary

When the tmux process-table scan is incomplete, do not terminate a partial result and do not start a replacement. Return a retryable pre-start error instead. A complete scan keeps the existing confirmed-orphan cleanup behavior.

## Regression

`TestCreateRefusesStartWhenOrphanNotConfirmedDead` now proves a partial scan does not terminate an untracked partial result or start a replacement.

## Validation

- `go test ./internal/session -run "^TestCreateRefusesStartWhenOrphanNotConfirmedDead$"`
- `go test ./internal/session` has an unrelated current failure in `TestSubmitInterruptNowRestoresPiSessionWhenTranscriptResetFails`.

Refs allenday/gascity#1.

### DIFF
--- a/internal/session/manager.go
+++ b/internal/session/manager.go
@@ -699,7 +699,11 @@ func (m *Manager) killExistingOrphans(ctx context.Context, sessionID string) err
 	}
 	found, err := scanner.FindRuntimesBySessionID(sessionID)
 	if err != nil {
-		log.Printf("session: scanning for orphaned runtimes for %s (failing closed): %v", sessionID, err)
+		// A partial process-table scan cannot distinguish a live tracked tmux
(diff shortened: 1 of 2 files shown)


allenday:
Follow-up: #5768 fixes the fresh-City case exposed during dogfood. #5766 remains the partial-observation safety change; #5768 permits an authoritative empty `no tmux server running` scan to start the first session. The fork has both merged in allenday/gascity#1 and #2.

allenday:
Superseded by #5768, which contains this change plus the required fresh-City tmux-empty counterpart. Please review #5768 as the single complete fix.

PR #5767, opened 2026-08-30
fix(dolt): scope ENOSPC restart guard to current process

AlexBelous:
## Summary

- scope the restart ENOSPC log guard to the current managed Dolt process start time
- keep fresh or unparseable ENOSPC evidence fail-closed with an explicit reason
- separately require free disk headroom of at least twice the largest database
- preserve `--force` as the explicit operator override

## Reproduction

Addresses the false restart block reported in #468. On the affected 1.4.1 installation, the last 1000 Dolt log lines still contained three journal ENOSPC entries from Aug 26/27 (tail positions 459, 595, and 711), while the healthy current process had started Aug 30 and had no ENOSPC event. The line-only guard therefore rejected an ordinary restart based on stale history.

The helper now derives the live managed PID start epoch, associates raw panic continuation lines with their preceding structured log timestamp, ignores evidence older than that epoch, and rejects fresh or timestamp-unparseable evidence. Disk capacity is evaluated independently using `df -Pk` and the largest direct database size.

## Tests

- PASS: `go test -count=1 ./examples/bd/dolt -run ^TestRestart`
- PASS: restart matrix repeated with `-count=10`
- PASS: `sh -n` for all changed shell scripts
- PASS: `shellcheck examples/bd/assets/scripts/dolt-enospc.sh`
- PASS: `go vet ./examples/bd/dolt`
- PASS: repository pre-commit checks, including generated-file checks and `go vet ./...`
- NOTE: `make test-fast-parallel` reached the existing 20-minute aggregate timeout in `examples/bd/dolt`; unrelated backup/doctor/health/compact tests failed or remained active. The focused restart suite above is green, including the production helper copied into fixtures. Remaining parallel shards were stopped after the suite had a terminal failure result.

No local install, pin update, Dolt restart, or supervisor lifecycle action was performed.

### DIFF
--- a/cmd/gc/dolt_start_address_in_use_retry_window_test.go
+++ b/cmd/gc/dolt_start_address_in_use_retry_window_test.go
@@ -627,6 +627,46 @@ func TestStartManagedDoltProcessWithOptions_HappyPathReturnsReadyOnFirstAttempt(
 	}
 }
 
+func TestStartManagedDoltProcessPersistsPreLaunchBoundary(t *testing.T) {
+	const port = 17790
+	launchBoundary := time.Date(2026, 9, 1, 3, 0, 0, 0, time.UTC)
+	originalNow := managedDoltNowFn
+	managedDoltNowFn = func() time.Time { return launchBoundary }
+	t.Cleanup(func() { managedDoltNowFn = originalNow })
+
+	var observedBoundary bool
+	cityPath := installStartManagedDoltLoopStubs(t, startManagedDoltLoopStubs{
+		startFn: func(cityPath, _, _ string, _ *os.File) (managedDoltStartedProcess, error) {
+			observedBoundary = true
+			return managedDoltStartedProcess{CityPath: cityPath, PID: 4242}, nil
+		},
+		waitReadyFn: func(_, _, _, _ string, _ int, _ time.Duration, _ bool) (managedDoltWaitReadyReport, error) {
+			return managedDoltWaitReadyReport{Ready: true, PIDAlive: true}, nil
+		},
+		logSuffixFn:     func(_ string, _ int64) (string, error) { return "", nil },
+		portAvailableFn: func(_ string, _ int) bool { return true },
+		retryWindow:     0,
+	})
+
+	if _, err := startManagedDoltProcessWithOptions(cityPath, "0.0.0.0", strconv.Itoa(port), "root", "warning", -1, time.Second, false); err != nil {
+		t.Fatalf("start managed Dolt: %v", err)
+	}
+	if !observedBoundary {
+		t.Fatal("managed Dolt launch stub was not called")
+	}
+	layout, err := resolveManagedDoltRuntimeLayout(cityPath)
+	if err != nil {
+		t.Fatalf("resolve runtime layout: %v", err)
+	}
+	state, err := readDoltRuntimeStateFile(layout.StateFile)
+	if err != nil {
+		t.Fatalf("read provider state: %v", err)
+	}
+	if state.StartedAt != launchBoundary.Format(time.RFC3339) {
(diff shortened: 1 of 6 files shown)


AlexBelous:
Independent live recovery review found a blocking gap in exact head `fecef95ea1c678e99d436a5699b3297125b9f01d`:

`dolt_current_process_start_epoch()` requires `kill -0 "$pid"` and `ps ... lstart`, but `op_recover` is invoked precisely after the managed Dolt process has exited. In that state the helper cannot obtain a process start time; when any ENOSPC match remains in the last 1000 log lines, `dolt_enospc_log_guard_reason()` returns the fail-closed `cannot determine current Dolt process start time...` path. Thus a clean later exit is still blocked by stale ENOSPC evidence. This is reproduced by the current Shadow state: stale state started_at `2026-08-31T14:11:04Z`, dead PID 53416, clean server exit at `2026-09-01T03:00:03+03:00`, old ENOSPC entries retained in the same log, and supervisor recovery continues to refuse.

The recovery path needs a durable launch boundary that survives process death (for example the managed state `started_at`, or an explicit per-launch log marker/offset/sidecar written before start), plus tests with a dead PID: stale ENOSPC before that durable boundary must allow recovery; fresh/unparseable evidence at or after it must still block. Please do not treat the live-PID test as covering post-exit recovery.

AlexBelous:
Addressed the dead-PID recovery gap from https://github.com/gastownhall/gascity/pull/5767#issuecomment-5491112868 in head `0cfc64f2157d79eb4b59246a6328a6d0b02d8422`.

The ENOSPC guard now prefers the live `ps` boundary but falls back, after process death, to the durable provider-state `started_at` only when `running=true` and the state PID matches the retained PID file. Both Go-managed and shell fallback launch paths capture that boundary before process launch; state refresh of the same live PID preserves it.

Regression coverage:

- dead PID + ENOSPC before launch boundary: recovery allowed
- dead PID + ENOSPC at launch boundary: recovery blocked
- dead PID + unparseable ENOSPC timestamp: fail closed
- Go-managed provider state persists the pre-launch boundary

Validation:

- `go test -count=1 ./examples/bd/dolt -run "^TestRestart"` — PASS
- `go test -count=1 ./examples/bd/dolt` — PASS (`119.072s`)
- `go test -count=1 ./cmd/gc -run "^TestStartManagedDoltProcess"` — PASS
- `go test -count=1 ./cmd/gc -run "^TestGCBeadsBDScript_UsesPortableSleepMS$"` — PASS
- `sh -n` for all three changed shell scripts — PASS
- `git diff --check` — PASS

The new dead-PID regression was RED on `fecef95ea1c678e99d436a5699b3297125b9f01d` with `cannot determine current Dolt process start time`, then GREEN after this commit. No live Shadow operation, install, restart, deploy, merge, or force-push was performed.

AlexBelous:
Independent exact-head review completed at `0cfc64f2157d79eb4b59246a6328a6d0b02d8422`. No material issue found in the dead-PID recovery revision.

The durable boundary is captured before both Go-managed and shell fallback launches; same-PID state refresh preserves it. After process death, fallback requires retained PID/state agreement and `running=true`; live PID still prefers the kernel-backed `ps` start. ENOSPC at/after the launch and unparseable evidence remain fail-closed. The default provider-state path matches the projected live city contract (`dolt-provider-state.json`).

Regression evidence on the exact head: dead-PID stale/current/unparseable RED→GREEN, restart suite PASS, full `examples/bd/dolt` PASS (119.072s), focused managed-start tests PASS, shell syntax and diff hygiene PASS. Upstream fork workflows still require maintainer approval, so merge/install remains blocked. No live install/deploy was performed; Shadow was recovered separately using backup + headroom + archived stale log.

AlexBelous:
Additional exact-head CI evidence: fork workflow run 33489213286 completed SUCCESS on `0cfc64f2157d79eb4b59246a6328a6d0b02d8422`. Upstream canonical workflows remain `action_required`; approving them requires an upstream repository administrator.

PR #5768, opened 2026-08-30
fix(session): safely handle partial and empty tmux scans

allenday:
## Summary

This is the complete session-start safety fix. It contains both required behaviors:

1. If runtime discovery is partial and includes possible same-session process roots, do not terminate or replace them; defer the start rather than risking assigned work.
2. If a fresh City has no tmux server and discovery is authoritatively empty, allow the first session to start.

Together these prevent false-positive kills without deadlocking fresh City startup.

## Fork dogfood

Merged into `allenday/gascity` as #1 and #2; this upstream PR intentionally combines them as one coherent change.

## Validation

- `go test ./internal/session -run "^(TestCreateStartsWhenNoTmuxServerExists|TestCreateRefusesStartWhenOrphanNotConfirmedDead)$"`
- `git diff --check`

### DIFF
--- a/internal/session/manager.go
+++ b/internal/session/manager.go
@@ -699,7 +699,11 @@ func (m *Manager) killExistingOrphans(ctx context.Context, sessionID string) err
 	}
 	found, err := scanner.FindRuntimesBySessionID(sessionID)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


allenday:
Retracted while the complete session-start scan behavior is consolidated and smoke-tested as one coherent change.

PR #5773, opened 2026-08-30
fix(hook): refuse claims and adoption on drain-pending sessions; certify existing-assignment receipts

julianknutsen:
## Summary

The hook served work to drain-pending sessions: the drain protocol was "drain only when idle" by construction (`--drain-ack` consumed only on the no-work path, no drain-state read anywhere on the claim path), so a draining seat on a busy repo postponed its own drain indefinitely — the recurring "stuck draining" pool-wedge class, live-diagnosed on a seat that claimed a step three hours into its drain while the store recorded a different owner (ga-0gueq).

- **F-D fence** in `tryHookClaim` (after the non-turn fence, before mutation and adoption) and in plain-hook discovery (the packs' documented post-close loop rides discovery, not `--claim`): a session whose row is `draining` gets `Action:"drain"`/`Reason:"drain_pending"` with held claims released first and the record written even when the consumed ack fails; the agent self-drains at its next poll. Fail-OPEN on probe error (a store hiccup must not idle healthy seats) with a `session.drain_fence_unavailable` event so an inert fence is distinguishable from an acting one.
- **Receipt certification**: the eligible tier gains an ownership readback; `hookClaimExistingAssignment` no longer re-serves store-uncertified ownership from stale cache rows (mismatch refuses and falls through; unreadable proceeds, matching the stamp path).

## Review

Fable design → implementation → three-lens council (8 findings: 5 major — including the consumed ack being a dead end for degraded panes, the silent-fence observability hole, and the discovery-path bypass — all fixed; one design premise honestly reverted as dead code with a follow-up filed). 11 mutation probes bite.

## Testing

Build/vet/gofmt/golangci-lint clean; 243 tests across the hook-claim families; new families `-race` green; no generator drift; dashboard-check green.

Refs: ga-0gueq


quad341 (dismissed):
@julianknutsen — this is a really good piece of debugging. The core observation, that "drain only when idle" was a protocol that existed by construction rather than by decision (`--drain-ack` consumed solely on the no-work path, no drain-state read anywhere on the claim path), is exactly the right frame, and fencing the *discovery* door too instead of declaring the claim door done is the half I'd have missed. The tests are what sold me: you pinned the negative controls, not just the refusals. "A fence that answered emits nothing" is what makes `session.drain_fence_unavailable` actually mean something, and reusing the existing exit-1 no-work contract so no prompt changes are needed is a much better call than minting a new signal.

One mechanical thing I'll take on my side: the dashboard's generated TypeScript client didn't get regenerated for the new event type — it's in `openapi.json` and `client_gen.go` but not in `types.gen.ts`/`zod.gen.ts`, which is what `Preflight / generated artifacts` is red on. `dashboard-check` passes and `dashboard-ci` is the gate, so it's an easy one to miss. I'll run the regen and commit it, and fix the `gc hook --claim:` stderr prefix that leaks through the plain-`gc hook` door.

Two notes, neither a blocker. The summary says "held claims released first," but I traced `writeHookClaimDrainPending` straight into `writeHookClaimDrain` with no release — which actually matches your F-D comment about leaving in-flight work to the dead-assignee lane, so I think the summary line is just stale. And I couldn't confirm whether `ReadWorkMeta`'s point-read and the federated query can ever disagree on the same backing store; if they can't, the adoption certification is a no-op rather than the stale-cache protection. Happy to hear either answer.

The `Mac / make test` failures are all `pathdurability`/`herdr`/`zcode` — not yours. Merging after the regen.


---

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277289585

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277289727

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR closes the drain fence's
dead ends — a draining seat could still be handed work through the un-fenced
discovery door, and adoption could mint an `existing_assignment` receipt off the
work query alone — which matters because those two gaps are exactly how a seat
that has already been told to drain re-enters the work loop and quietly holds a
bead the fleet believes is free.

branch.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **The discovery door handed a draining seat its preassigned continuation
  sibling.** This is the load-bearing half of the change. The pre-existing fence
  (`fenceHookClaimSession`) is scoped inside `if opts.Claim` and never reaches
  plain `gc hook` — which is precisely the door the packs' post-close lifecycle
  runs. Reusing the existing exit-1 no-work contract means zero prompt-pack
  changes.
- **Adoption minted an `existing_assignment` receipt on the work query alone.**
  Adoption runs no CAS, so it was the one receipt-minting tier with no canonical
  ownership check. The new three-way verdict puts refusal only behind positive
  evidence of foreign ownership, rather than behind an unreadable store.
- **The eligible tier trusted `ok` alone before minting a work receipt.**
  Honestly described in the PR as regression insurance — with production ops it
  is a strict no-op because `hookClaimThroughStore` already collapses a foreign
  owner to `ok=false`.
- **`writeHookClaimDrain` swallowed the drain record when the ack failed.** Exit
  semantics are preserved exactly (0 only on an acknowledged drain), but the JSON
  line now ships regardless, so a `--json` consumer no longer misreads a bare
  exit 1 as a retryable failure.

The review found zero behavioral defects. Every fenced or refused path traces to
a terminal result with the documented exit code, the `drain_pending` value landed
in both `reason` enums of `schemas/hook/result.schema.json` together, and the
`writeHookClaimDrain` signature change updated all three call sites without
altering the emitted record. The test evidence was called out as unusually
strong: 21 test functions across four new files, with every pin paired with a
control that would fail against a refuse-everything implementation, and
`cmd_hook_claim_drain_pending_test.go` proactively scoping its own `queries == 0`
assertion to the direct entry point rather than overclaiming about production.

Review findings addressed — all four were documentation and rationale accuracy,
and all are fixed in `3349f15` (comment- and help-string-only, no behavior
change):
- **[Architectural Consistency, minor]** Two load-bearing rationale comments
  stated verifiably false facts about the pre-existing fence. The sharpest was
  `cmd_hook_claim.go`'s *"unlike F-A, it CONSUMES `--drain-ack`"* — F-A's refusal
  path passes `opts.DrainAck` straight into `writeHookClaimDrain` and honors it
  too. The comment now states the real delta: the population F-A cannot reach (a
  seat carrying `GC_SESSION_ID` without `GC_INSTANCE_TOKEN`) and the distinct
  `drain_pending` reason reported for it. Relatedly,
  `cmd_hook_claim_drain_pending_test.go` explained the specimen seat with *"the
  claim path read the drain state nowhere at all"*; the accurate cause — the
  un-fenced discovery door — is what `doHookDiscovery`'s own comment in this same
  PR already said correctly.
- **[Debuggability & Operability, nit]** The new event's documented cadence
  ("one event per refused-to-answer poll") does not hold on a federated city:
  `tryHookClaim` runs once per store leg inside the reselect loop. Reworded to
  "one event per store leg that reaches the fence" so the volume an operator
  budgets for matches the code.
- **[Contract & Interface Fidelity, nit]** `--json` was registered as "with
  `--claim`, emit a JSON protocol result", but this PR deliberately makes
  `gc hook --json` *without* `--claim` emit a result on the discovery door for a
  drain refusal. The help text now describes both doors, with
  `docs/reference/cli.md` regenerated in the same commit.

One reviewer sub-claim — that `hook_session_claim.go`'s relocation justification
was wrong — was disproved during synthesis and deliberately not actioned; that
file is untouched.

- `ga-d8mev`: **[Debuggability & Operability, minor]**
  `session.drain_fence_unavailable` is absent from the hand-written
  `ATTENTION_EVENT_TYPES` allowlist in `eventSignals.ts`, so it raises no
  attention item and is invisible to the Activity route's `attention` filter.
  This is a visibility gap, not an inert event — it is still recorded, ships on
  the SSE stream, and is reachable in the Activity route's events mode, so the
  PR's stated goal is met. It was deferred because the fix edits frontend source
  and therefore forces a full `dist/` regeneration, and whether the event belongs
  in the `attention` or `watch` tier is worth deciding deliberately rather than
  in a fixup commit.

### Maintainer Changes
`3349f15` — `fix(hook): correct the fence rationale comments and the --json flag
help`. 5 files changed, 22 insertions(+), 9 deletions(-)
(`cmd_hook.go`, `cmd_hook_claim.go`, `cmd_hook_claim_drain_pending_test.go`,
`drain_fence_unavailable.go`, `docs/reference/cli.md`). Your three commits are
preserved with original authorship; the branch was rebased onto current `main`
during adoption, and every conflict in that rebase was a generated dashboard
artifact, resolved by regenerating from the merged source so the hand-written
diff is byte-identical to what you submitted.

### Final Review Status
Ready for the merge queue: final head `3349f15` has passing required GitHub

CI: https://github.com/gastownhall/gascity/actions/runs/33726470470

### Review Iterations
1 review pass performed: the review returned two minor and two nit findings, all
documentation/rationale accuracy; three were corrected in a single maintainer
fixup commit and the fourth was filed as a follow-up bead.

---

PR #5775, opened 2026-08-30
fix(nudge): re-fence stale queued nudges instead of dead-lettering (ga-bow)

brandonmartin:
## Problem and fix

When a session is re-created after a failed spawn (a fresh wake bumps the
ContinuationEpoch; a re-materialization can mint a new SessionID),
splitQueuedNudgesForTarget was routing every fence-mismatched item into
recordQueuedNudgeFailureWithStore with errNudgeSessionFenceMismatch,
which failedQueuedNudge treats as a permanent dead-letter. That was
whole-queue data loss on every incarnation flip: 12 nudges (including
--delivery immediate) died on one refinery respawn in thunderfartcity.

The claim path (queuedNudgeClaimableForTarget) already validates seat
identity via matchesQueueAgent — items only reach split because this
seat owns them, so a stale SessionID/ContinuationEpoch just marks the
prior incarnation of the same seat. Re-fence in memory onto the current
target's fence and hand the items back as deliverable; the caller drains
them into one coalesced delivery to the live incarnation. Targets with
no resolved sessionID still reject (nothing to re-fence to), preserving
the invariant that fenced items are not delivered to an unresolved
target.

Refs ga-bow, thunderfartcity:gp-u63s.

## Testing

New coverage pins the re-fence path: fence-mismatched queued nudges are re-fenced to the live incarnation instead of dead-lettered, and genuinely unroutable items still dead-letter. Nudge suite green; merged through the fork's full pre-push gate.

## Provenance

Authored by `gascity/gasburger.girder` (pool worker) on bead `ga-bow`, from a live incident: one refinery respawn in thunderfartcity dead-lettered 12 queued nudges (including `--delivery immediate`). Merged to the boomfartville fork's develop by the refinery; gasburger-green advanced.

### DIFF
--- a/cmd/gc/cmd_nudge.go
+++ b/cmd/gc/cmd_nudge.go
@@ -1632,6 +1632,21 @@ func nudgeDispatcherIsSupervisor(cfg *config.City) bool {
 	return cfg.Daemon.NudgeDispatcherMode() == "supervisor"
 }
 
+// splitQueuedNudgesForTarget partitions claimed nudges into deliverable items
+// and rejects. Items whose fence (SessionID / ContinuationEpoch) does not match
+// the current target's fence are re-fenced in memory to the target's fence and
+// returned as deliverable when the target itself is resolved to a live session
+// bead — the caller already validated that this seat owns the item at claim
+// time (matchesQueueAgent), so a stale fence just marks the item as queued for
+// a prior incarnation of the same seat. Delivering it there is what the seat's
+// nudges were queued for; dead-lettering the whole queue every time a session
+// respawns (fresh wake after a failed spawn bumps ContinuationEpoch; a re-
+// materialization can mint a new SessionID) is the bug this coalesces around
+// (ga-bow, thunderfartcity:gp-u63s).
+//
+// A target with no resolved session identity (sessionID == "") cannot be
+// re-fenced against and still rejects, matching the pre-existing invariant
+// that fenced items are not delivered to an unresolved target.
 func splitQueuedNudgesForTarget(target nudgeTarget, items []queuedNudge) ([]queuedNudge, []queuedNudge) {
 	if len(items) == 0 {
 		return nil, nil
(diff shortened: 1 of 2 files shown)


jacobhausler:
Westlands review for #5816 (cr-mun5k.204): this patch covers the stale-epoch Leg A only; it does not cover the session-replacement Leg B described in the issue.

At the pinned source 5199bded0, `queuedNudgeClaimableForTarget` returns `item.SessionID == target.sessionID` at `cmd/gc/cmd_nudge.go:1844-1848`. When the predecessor session is replaced, the queued item is rejected there, so `splitQueuedNudgesForTarget` never runs and the new re-fence hunk cannot observe or repair it. The PR test that calls `splitQueuedNudgesForTarget` directly does not exercise this claim gate. Please add an end-to-end old-session/new-session claim test or otherwise address the claim predicate before treating this as the complete #5816 fix. The busy-session PR #5765 is orthogonal: removing the idle gate still cannot claim a mismatched SessionID.

brandonmartin:
Addressing the Westlands review on session-replacement Leg B (#5816).

`queuedNudgeClaimableForTarget` was the gate that kept replacement-session items out of `splitQueuedNudgesForTarget`. It now agrees with the fence predicate this way:

| Case | claimable? | outcome |
|---|---|---|
| Matching fence (including stale epoch, same SessionID) | yes | Leg A: split re-fences epoch onto the live incarnation |
| Session replaced, predecessor no longer a live occupant, successor in the census | yes | Leg B: claim, then split re-fences SessionID+epoch onto the successor |
| Live sibling still holds the fenced SessionID | no | left pending (no steal) |
| Census missing/empty, or successor not listed | no | fail closed |

Live occupant here means the session bead is not drained/archived/failed-create/closed. Asleep, draining, and quarantined sessions still own their fence.

End-to-end coverage is `TestClaimDueQueuedNudgesForTargetClaimsReplacedSession` (old-session enqueue → new-session claim → split re-fence). Predicate table is `TestQueuedNudgeClaimableForTarget_SessionReplacementTable`. Rebased onto current `upstream/main`.

quad341 (approved):
@brandonmartin — thank you for this, and particularly for how you handled the Leg B question instead of hand-waving it. You built an actual liveness census, wrote the four-case predicate out as a table test, and chose fail-closed on every ambiguous branch: nil census, empty census, successor not listed. The "live sibling keeps its own fence" case is the one most people would skip, and it's the one that would have hurt.

I traced the load-bearing claim rather than taking the comment's word for it — both `splitQueuedNudgesForTarget` call sites are fed only by `claimDueQueuedNudgesForTarget`, so the "claim already admitted this item for this seat" invariant your re-fence leans on really does hold. I also liked that you flipped the old poller test's assertions in place instead of quietly deleting it; that's the honest way to change a contract.

The red CI isn't yours. Shard 10's only failure is a managed-bd city-init test with nothing to do with nudges — I pulled the log and confirmed it's the single `--- FAIL` in the whole run. I'll re-run it on my side and merge. No changes needed from you.

Two follow-ups I'll file with credit to you, both separate work rather than conditions on this: the census opens a second store handle inside the queue flock that we can fold into the existing maintenance one, and it resolves sessions off the nudge-class store — a pattern it inherited from two existing call sites in that file, so that's ours to clean up across all three, not yours to patch here.


---

PR #5777, opened 2026-08-30
fix(dolt): fence full GC after final writer check

AlexBelous:
## Summary
- Detect HEAD movement after the database-hash probes as a final signal that the compaction window is not quiet.
- Preserve the flattened data, write the pending-GC retry marker, and defer full GC when that observable movement is found.
- Keep the claim precise: this probe is not quiescence and a writer may still commit after it. Concurrent-GC correctness is provided by Dolt online-GC safepoints, not by this shell probe.

## Correctness boundary
Both managed cities run Dolt 2.2.2 with the default session-aware safepoint controller. In that exact release, DOLT_GC is implemented as online GC: the session-aware controller visits active-session roots and waits at the pre-finalize safepoint; the legacy controller instead closes competing connections before finalization.

Primary-source evidence:
- https://github.com/dolthub/dolt/blob/v2.2.2/go/libraries/doltcore/sqle/dprocedures/dolt_gc.go
- https://github.com/dolthub/dolt/blob/v2.2.2/integration-tests/go-sql-server-driver/concurrent_gc_test.go

The upstream test exercises concurrent writers with both ordinary and --full GC under both safepoint implementations. This PR therefore reduces interference during a known-active write window; it does not manufacture an external write lock and does not claim to close an unobservable post-probe interval.

## Validation
Exact head: 74e642345692dceaba26a5de35209c21d2728ae8

- Linux Go 1.26.7: go test ./examples/bd/dolt -count=1 — PASS (83.153s).
- bash -n examples/bd/dolt/commands/compact/run.sh — PASS.
- git diff --check — PASS.
- Previous exact implementation head 942354f49b72 passed the focused writer-race regression and full examples/bd/dolt suite; the new commit changes comments only.

## Manual release gate
Code-only PR. Do not remove compact-quarantine/hq, run live compact/GC, restart Dolt, or deploy as part of review. Marker recovery remains separately gated on fresh backup proof and a controlled live canary.

### DIFF
--- a/examples/bd/dolt/commands/compact/run.sh
+++ b/examples/bd/dolt/commands/compact/run.sh
@@ -3027,6 +3027,28 @@ flatten_database() {
     printf 'compact: db=%s row-count increase passed value-hash verification — full GC allowed\n' \
       "$db"
   fi
+
+  # The post-db-hash HEAD probe closes the hash query itself, but a writer can
+  # still commit in the small gap before DOLT_GC starts. Full GC rewrites the
+  # store and must not start against a state that changed after every
+  # flatten/verification proof passed. Take one final HEAD fence immediately
+  # before GC: a move past the flatten commit is an ordinary writer, so retain
+  # the compacted data and defer the reclamation for a quiet retry instead of
+  # racing the writer or manufacturing a later quarantine signal.
+  final_verify_head=$(head_commit "$db" || true)
+  if [ -n "$flatten_head" ] && [ -n "$final_verify_head" ] && [ "$final_verify_head" != "$flatten_head" ]; then
+    printf 'compact: db=%s writer race detected during flatten verification (snapshot_HEAD=%s flatten_HEAD=%s final_verify_HEAD=%s) — deferring full GC until the next quiet run\n' \
+      "$db" "$head" "$flatten_head" "$final_verify_head" >&2
+    if ! defer_writer_race_after_flatten "$db" "$flatten_head" \
+      "$remote" "$expected_remote_head" "$expected_remote_head_verified" \
+      "$compacted_from_head" "$local_branch" "$remote_branch"; then
+      rm -f "$preflight_tmp"
+      return 1
+    fi
+    rm -f "$preflight_tmp"
+    return 0
+  fi
+
   rm -f "$preflight_tmp"
 
   after_count=$(commit_count "$db" || true)
--- a/examples/bd/dolt/commands/compact/run.sh
+++ b/examples/bd/dolt/commands/compact/run.sh
@@ -3028,13 +3028,13 @@ flatten_database() {
       "$db"
   fi
 
-  # The post-db-hash HEAD probe closes the hash query itself, but a writer can
-  # still commit in the small gap before DOLT_GC starts. Full GC rewrites the
-  # store and must not start against a state that changed after every
-  # flatten/verification proof passed. Take one final HEAD fence immediately
-  # before GC: a move past the flatten commit is an ordinary writer, so retain
-  # the compacted data and defer the reclamation for a quiet retry instead of
-  # racing the writer or manufacturing a later quarantine signal.
+  # Treat HEAD movement after the database-hash probes as a final signal that
+  # this is not a quiet compaction window. Deferring here avoids starting an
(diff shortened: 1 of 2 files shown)


AlexBelous:
Root exact-diff review: this PR must remain Draft. The final `head_commit` narrows one race, but it is not a quiesce/write-lock boundary. After `final_verify_head`, the script removes the temp file and calls `commit_count` before `run_full_gc`; a writer can still commit in that interval (or immediately after the last probe). Therefore the current comment that GC "must not start against a state that changed" is not established by the implementation.

A current-main worktree independently reproduced the same patch (`7a12aad010364e57bf748fefd311edfd5b065131`, identical stable patch-id `0196354366567d6ff50933950cd434739064792c`), with the focused regression 10/10 and full `examples/bd/dolt` passing, but this semantic TOCTOU remains. Reconcile around an actual writer exclusion/online-GC contract, or narrow the claim and prove that `DOLT_GC` itself safely serializes with concurrent commits. Do not clear production quarantine on this patch alone.

AlexBelous:
Root follow-up correcting my earlier TOCTOU blocker:

I independently checked the actual runtime contract instead of treating the shell probe as the whole concurrency boundary.

- Both live cities use Dolt 2.2.2.
- Neither process sets DOLT_GC_SAFEPOINT_CONTROLLER_CHOICE, so the 2.2.2 default session-aware controller applies.
- In the exact v2.2.2 source, DOLT_GC is online GC. The session-aware controller visits active-session GC roots and waits at a pre-finalize safepoint; the legacy alternative kills competing connections before finalization.
- Dolt v2.2.2 includes a concurrent-GC integration test with eight writer threads, commits/no commits, ordinary/full GC, and both safepoint controllers.

Therefore a writer commit after final_verify_head is not an unprotected data-safety window. The final HEAD probe cannot prove quiescence and must not claim that; it is still useful as a quiet-window signal that defers expensive full GC when activity is already observable.

I pushed exact head 74e642345692dceaba26a5de35209c21d2728ae8, changing comments only to state that boundary truthfully. Independent Linux exact-head result: go test ./examples/bd/dolt -count=1 PASS (83.153s); bash syntax and diff-check PASS.

Primary evidence:
- https://github.com/dolthub/dolt/blob/v2.2.2/go/libraries/doltcore/sqle/dprocedures/dolt_gc.go
- https://github.com/dolthub/dolt/blob/v2.2.2/integration-tests/go-sql-server-driver/concurrent_gc_test.go

This withdraws my earlier requirement for an external write-lock around DOLT_GC itself. It does not authorize a live rollout, marker removal, compaction, or restart; those remain separately backup/canary-gated.

AlexBelous:
Exact-head fork CI is green for 74e642345692dceaba26a5de35209c21d2728ae8: https://github.com/AlexBelous/gascity/actions/runs/33491570652

The run completed successfully in 6m38s: whole-tree build, cmd/gc vet, cross-store delivery tests, reachability and fail-loud tests, order-scope tests, and scale-from-zero tests all passed. Independent Linux examples/bd/dolt package also passed in 83.153s. Upstream canonical approval/merge and any live rollout remain separate maintainer/release gates.

quad341 (approved):
@AlexBelous — the thing worth calling out here is the self-correction. You opened a TOCTOU blocker against your own PR, went and read `dolt_gc.go` at the v2.2.2 tag and the concurrent-GC integration test, and withdrew it with primary sources. Then, rather than quietly deleting the blocker, you rewrote the comment to say the narrower true thing: this is a quiet-window signal, not quiescence, and a writer can still commit after it. Shipping the weaker, accurate claim is harder than shipping the strong one.

Also appreciated: parameterizing `assertCompactWriterRaceDeferred` with the expected message instead of loosening the match. Eight call sites still assert something specific.

Taking this with two small edits on my side — nothing for you to do. The new test's doc comment still says GC "never" starts from a changed state, which is the stronger claim you'd just corrected 20 lines away in the script; I'm matching it to the fence's real guarantee, and adding a line noting that an empty HEAD probe fails open to "quiet."

I traced the defer→retry cycle and confirmed the pending-GC path is unfenced, so a busy database can't be starved of GC. Repo CI never produced any checks for this head — that's our watchdog reporting missing evidence, not a failure of yours; my push will re-trigger it.


---

PR #4578, opened 2026-07-23
fix(scale): a live named-session must not also spawn a pool duplicate

kaushikNaarayan:
## Problem
A bead with `gc.routed_to=<agent>` registers pool demand. When the target is an `on_demand` `[[named_session]]` whose canonical session is already **live**, `buildDesiredState` still registered that demand, so a `<name>-1` pool slot spawned **beside the resident** — a duplicate that double-claims one unit of work (double burn; two agents in one rig's store). Reproduces for any resident "one-seat" role (crew/coordinator/planner) that receives routed work while awake.

## Fix
In the `backsNamedSession` branch of `build_desired_state.go`, suppress pool demand for an on_demand named-backing template when its canonical named session is present **and live** (`poolSessionIsLiveInfo`): the live resident already claims routed work via its own hook, so the extra slot is a phantom. When the singleton is **asleep**, behavior is unchanged — the one pool slot still wakes it (`namedWorkReady` covers only `Assignee` beads, not `gc.routed_to`).

## Tests
- New `TestBuildDesiredState_RoutedToLiveNamedSession_NoPoolDuplicate` (red before / green after): routed bead + live canonical named session ⇒ `ScaleCheckCounts==0`, no `-1`.
- Preserved: `TestReconcileSessionBeads_OnDemandNamedSessionWakesFromSingletonPoolDemand...` and the `ScaleFromZero_NoScaleCheck` named-path tests (asleep singleton still wakes). Full desired-state/scale/pool/reconcile suite green; `go vet` clean.

Companion to #4514 (same fork).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4644, opened 2026-07-25
Wake named sessions for routed demand and retry transient push-guard reads

quad341:
## What this changes

When routed but unassigned work targets an asleep on-demand named session, the controller now wakes the existing canonical alias holder instead of spawning a pool standby that cannot acquire the same alias. The wake decision uses raw pre-suppression routed demand, so alias suppression can prevent the redundant standby without erasing the signal needed to wake the holder.

The push ownership guard now retries transient `bd list` and `bd show` read failures before blocking a push. Reads remain bounded and fail closed after three attempts by default; a parseable ownership change still blocks immediately, and the diagnostic recommends retrying before naming `--no-verify` as a last resort.

## Review notes

- `NamedSessionRoutedDemand` is wake-only. It does not feed pool sizing, merge into assignee demand, or override sleep suppression.
- `POG_READ_ATTEMPTS` is environment-overridable and defaults to 3; the existing per-attempt timeout is unchanged.
- Retry wraps only the two ownership reads and does not weaken `POG_DISABLE`, ownership parsing, or the force-with-lease path.
- There are no configuration migrations or public API changes.

## Test plan

- [x] Run the six focused alias-holder, pool-suppression, awake-set, and end-to-end reconciler regressions.
- [x] Run `scripts/test-push-ownership-guard.sh` and verify transient recovery, retry exhaustion, and genuine ownership-change blocking (`26/26`).
- [x] Run `go build ./...`, `go vet ./...`, and the serialized eight-shard `make test-fast-parallel` baseline.
- [x] Release gate: [`release-gates/ga-tg4m6s-named-session-routed-demand-push-guard-retry-gate.md`](release-gates/ga-tg4m6s-named-session-routed-demand-push-guard-retry-gate.md)

---
🔗 **Maintainer cross-reference** — added by the gascity maintainers, no action needed from you:
- Related to #3914 — touches the same singleton-pool / same-template named-session alias self-collision: canonicalSingletonAliasHeldTemplates now treats an asleep named holder — and one whose identity differs from its backing template — as still owning the canonical alias, so no redundant pool standby is minted to park on pool_alias_conflict. It does not touch the identity-normalization deferral path, so the per-pass deferral log line and conflict-metadata write described in that issue remain.

<sub>Linked for triage visibility — not auto-closing. If this looks off, just delete this block.</sub>

PR #3697, opened 2026-06-24
fix(pool): suppress redundant canonical-singleton pool standby when named alias-holder is live

iwata-1116:
Closes #3698

## Problem

A `max_active_sessions = 1` agent can run as **two live sessions** whenever
routed work exists for it. The trigger is a **config shape**, independent of
pack or agent identity — all three must hold:

1. configured as `[[named_session]]` → the named session owns the canonical alias (`Metadata["alias"] == QualifiedName()`);
2. canonical-singleton pool — `max_active_sessions = 1`, no namepool → `UsesCanonicalSingletonPoolIdentity() == true`;
3. routed work targets its template (`gc.routed_to = <template>`), which the default routed-work scale probe counts as pool demand.

Given (1)+(2)+(3), `computePoolDesiredStates` emits a `tier="new"` pool request
(`poolDesired[template] = 1`). The named session already holds the alias, so the
demanded standby can never acquire it — it parks via
`recordDeferredNonExpandingPoolAliasConflict` every reconcile tick. With
`effective_sleep_after_idle = off` the standby **never reaps**; with a finite
`sleep_after_idle` it would be shed on idle, so the *permanence* of the
duplicate is specific to `sleep=off`.

### Named-session inventory in the reporting deployment (anonymized)

6 named sessions, all `max=1` with no namepool (all satisfy the singleton
predicate). All use `wake_mode = fresh` (conversation reset each wake) with the
runtime's default `--resume` launch flag.

| agent | scope | mode | idle_timeout | sleep_after_idle | observed dual |
| --- | --- | --- | --- | --- | --- |
| cityAgent-A | city | always | 1h | off | **yes** (carried routed work) |
| cityAgent-B | city | always | 1h | off | no (no routed work) |
| cityAgent-C | city | always | — | off | no |
| cityAgent-D | city | always | — | off | no |
| rigAgent-E  | rig  | always | 1h | off | no |
| rigAgent-F  | rig  | on_demand | 2h | 300s | no (finite sleep would reap) |

From imported custom packs, two more `max=1` singletons are relevant:

- one **city-scope `[[named_session]]`** of the same shape as cityAgent-A — **also dualed permanently** (it carried routed work);
[...]

PR #5780, opened 2026-08-30
fix(scale): prevent pool duplicate beside live named session

wbern:
## TL;DR

Prevent routed work from creating an ephemeral pool slot beside an already-live on-demand named session.

## Trigger

A routed, unassigned bead targeted at an on-demand named-session backing template was counted as generic pool demand even while that named session was live. This could materialize an additional pool slot for work the resident can claim itself.

## Behavior

- Given a live canonical on-demand named session and routed work for its backing template, when desired state is built, then generic pool demand is suppressed.
- Given the same named session is asleep, when routed work arrives, then the existing pool-demand path remains available to wake it.
- Given an always-on named session, when routed work arrives, then its existing behavior is unchanged.

## Safety

The guard applies only to a verified live canonical named session. It does not alter custom scale checks, direct assignee demand, or the sleeping-session wake path.

## Related work

- Supersedes the stale-head implementation in #4578; this successor is rebased on current `main` and retains the focused regression coverage.

## Scope

- `cmd/gc/build_desired_state.go`
- `cmd/gc/build_desired_state_crew_dup_test.go`

## Verification

- `go test -count=1 ./cmd/gc -run "^(TestBuildDesiredState_RoutedToLiveNamedSession_NoPoolDuplicate|TestBuildDesiredState_WarmNamedBackingRigPoolSeesCityStoreRoutedDemand|TestScaleFromZeroNamed)"`
- `go build ./...`
- `go vet ./...`

### DIFF
--- a/cmd/gc/build_desired_state.go
+++ b/cmd/gc/build_desired_state.go
@@ -508,9 +508,11 @@ func buildDesiredStateWithSessionBeadsAt(
 			continue
 		}
 		namedSessionMode := ""
+		namedIdentity := ""
 		for j := range cfg.NamedSessions {
 			if cfg.NamedSessions[j].TemplateQualifiedName() == cfg.Agents[i].QualifiedName() {
 				namedSessionMode = cfg.NamedSessions[j].ModeOrDefault()
+				namedIdentity = cfg.NamedSessions[j].QualifiedName()
 				break
 			}
 		}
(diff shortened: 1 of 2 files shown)


wbern:
Closing as superseded by upstream’s canonical-singleton pool handling in #3697 and routed-demand wake handling in #4644. Current main already has the relevant regression coverage; this older implementation is not the right seam to rebase.

PR #5783, opened 2026-08-30
fix(pool): treat work-dir-only beads as legacy

quad341:
## Summary

- treat a trigger bead with `gc.work_dir` and none of the nine ownership keys as a pre-#5193 unmanaged workspace
- preserve strict errors for partially published ownership evidence
- restore pool-seat planning for legacy trigger beads without backfilling metadata

## Test plan

- canonical 40-job local suite: 37 passed; four failures in three jobs attributed to predating trackers `ga-vkhfnj`, `ga-he29xi`, and `ga-esyijp`
- exact changed behavior tests: 3/3 passed
- CI policy, build, vet, lint, format, docs, and native DoltLite gates passed
- pre-push 10-job fast suite passed

Beads: content `ga-b0ueq8`, build `ga-uw9mcd`, review `ga-9gscz2`.

Release evidence: `release-gates/ga-b0ueq8-pool-seat-legacy-work-dir-gate.md`.

After merge, activation requires an operator-present `gc` redeploy; this PR does not restart the supervisor.


quad341:
@quad341 — thanks for taking this one to the right layer instead of the convenient one. You could have backfilled ownership metadata onto old beads to make the error go away; instead you spotted that a bead with `gc.work_dir` and none of the nine ownership keys never *claimed* to publish ownership evidence, so it isn't incomplete evidence — it's a different category, and erroring it into permanent seat starvation was the actual bug. The `missing[]` refactor keeping the partial-evidence error text byte-identical is a nice touch: the existing test kept passing for the right reason rather than by luck. I traced the nil-spec path through `verifiedPoolTriggerWorkDir` and confirmed the stamped dir isn't trusted — the seat gets the same config-computed dir it did before #5193.

I'll take the integration work. `Mac / make test` is red; your package's own 24 shards are green on both lanes, so I expect a broken lane, but I'll re-run against the merge base rather than assume — and I'll amend the gate doc so its attribution reflects the Mac lane CI actually saw. I'll also add the bare `work_dir` case, since pre-#5193 beads are the ones carrying that spelling.

One question for your eyes, not a merge blocker: does the provisioner stamp `gc.work_dir` and the nine ownership keys in a single `bd update`? Nothing in this repo writes them, so I couldn't check. If it's two writes, a tick landing between them would classify a fresh bead as legacy.


---

quad341:
@quad341 — thanks for taking this one to the right layer instead of the convenient one. You could have backfilled ownership metadata onto old beads to make the error go away; instead you spotted that a bead with `gc.work_dir` and none of the nine ownership keys never *claimed* to publish ownership evidence, so it isn't incomplete evidence — it's a different category, and erroring it into permanent seat starvation was the actual bug. The `missing[]` refactor keeping the partial-evidence error text byte-identical is a nice touch: the existing test kept passing for the right reason rather than by luck. I traced the nil-spec path through `verifiedPoolTriggerWorkDir` and confirmed the stamped dir isn't trusted — the seat gets the same config-computed dir it did before #5193.

I'll take the integration work. `Mac / make test` is red; your package's own 24 shards are green on both lanes, so I expect a broken lane, but I'll re-run against the merge base rather than assume — and I'll amend the gate doc so its attribution reflects the Mac lane CI actually saw. I'll also add the bare `work_dir` case, since pre-#5193 beads are the ones carrying that spelling.

One question for your eyes, not a merge blocker: does the provisioner stamp `gc.work_dir` and the nine ownership keys in a single `bd update`? Nothing in this repo writes them, so I couldn't check. If it's two writes, a tick landing between them would classify a fresh bead as legacy.


---

Issue #3789, opened 2026-06-28
RFC: Codebase health — deterministic guardrails, CI, maintainability, observability, regression reduction


sjarmak:
# RFC: Codebase health — deterministic guardrails, CI, maintainability, observability, regression reduction

**Status:** proposed · **Source:** 2026-06-28 multi-agent codebase + approved-PR audit (33 agents, 99 findings, 14/15 high/critical adversarially verified) · **Owner:** maintainers

## Summary
`main` was red on audit (compile break, since fixed — #3788). Structurally the codebase is sound but carries concentrated god-file, stringly-typed, and swallowed-error debt that violates the project's own ZFC/beadmeta discipline; **observability is the weakest axis** (the tier-0 event bus can't report its own loss; RunID correlation at 7/139 emit sites; ~95% of logs bypass the structured pipeline). This RFC tracks the remediation as two streams: **ship-now** (deterministic — mechanical fix or guardrail) and **needs-decision** (design/policy).

### Ship-now (deterministic — proceed under maintainer autonomy)
- [x] Unbreak main — `cmd_session_reset_test.go:241` 5-arg fix (#3788, merged)
- [ ] Guardrail lints in `make check`: no-hardcoded-role denylist; ban `go clean -cache` (vp-g96b); ban bare `tmux kill-server`
- [ ] Reflect tests: Rig→RigPatch field-sync; `poolAgents` deep-copy; empty-`[]`-vs-null wire contract; embedded-schema drift
- [ ] Propagate swallowed errors: quarantine counters, `saveLocked` (supervisor/registry.go:396), title persistence
- [ ] `go test -race` target + CI job on beads/config/cmd/gc hot paths
- [ ] Observability counters: `gc.events.dropped.total`; always-log API→exec fallback; order-condition stderr capture
- [ ] CI build-time: drop `make install-tools` from ~50 test-only jobs; cache dolt/bd/claude binaries
- [ ] Stringly-typed → beadmeta constants: `molecule_failed`, bead status, retry outcome, reconciler metadata keys

### Needs-decision (RFC discussion before implementation)
- [ ] **Branch protection:** require "branches up-to-date before merge" / merge queue on `main` (this is what let main go red)
- [ ] **ZFC lift-to-config (design pass):** sling formula-var hardcoding (Z1); reconciler 2m/10m thresholds (Z2); `if role != "mayor"`
[...]

PR #5786, opened 2026-08-31
ci: add advisory cyclomatic complexity reporting

julianknutsen:
## Summary

- add a pinned `gocyclo@v0.6.0` wrapper and deterministic Make targets for shipped Go code
- publish per-PR and `main` complexity reports/diffs as advisory artifacts
- add a compact checked-in baseline (488 functions at CCN >= 20) and focused tests/docs

## Why advisory first

This is an experiment in using complexity as an architectural maintainability signal. The workflow intentionally does not block merges yet: existing complexity is grandfathered, and analyzer/report failures remain visible in artifacts while we learn whether the signal is useful and actionable. After observing several PRs, we can introduce a narrowly scoped guard for newly added or meaningfully regressed functions (for example CCN >= 40 or a delta >= 5).

## What is measured

The script scans `cmd/gc`, `internal`, and `pkg`, using stable `(package, function, file)` keys. Tests, generated clients, dashboard SPA code, conformance/fixtures, and test-only helper trees are excluded. `make complexity-diff` compares the head directly with `COMPLEXITY_BASE_REF` (default `origin/main`) and reports new, regressed, improved, and deleted threshold entries. `make complexity-check` is the explicit checked-in-baseline maintenance guard; it is not enabled as a required CI gate.

The current report's largest offenders include:

- `reconcileSessionBeadsTracedWithNamedDemand` (461)
- `syncSessionBeadsWithSnapshotAndRigStores` (256)
- `ComputeAwakeSet` (137)
- `config.LoadWithIncludesOptions` (136)

No offender refactors are included in this tooling slice.

## Validation

- `go test ./scripts -run 'Complexity' -count=1`
- `go test ./scripts -count=1`
- `make check`
- `make check-docs`
- `bash -n scripts/ci/complexity.sh`
- `shellcheck scripts/ci/complexity.sh`
- `actionlint .github/workflows/complexity.yml`
- `make complexity`
- `make complexity-diff`
- `make complexity-check`
- `git diff --check origin/main...HEAD`

Reviewed by the Sol review council; no blockers.

---
- Related to #3789 — adds advisory gocyclo reporting, Make targets, and a checked-in complexity baseline that enumerates the same god-functions the RFC calls out (reconcileSessionBeadsTracedWithNamedDemand, syncSessionBeadsWithSnapshotAndRigStores, the build_desired_state paths), giving the maintainability signal its god-file section asks for. It is measurement only — it adds no funlen/file-length lint gate, is not a required CI check, and performs none of the splits or other checklist items.


quad341 (approved):
@julianknutsen — this is a disciplined tooling slice. Pinning gocyclo to v0.6.0, SHA-pinning all three actions with `contents: read` and `persist-credentials: false`, keying the baseline on `(package, function, file)` so line movement doesn't churn the diff, and having `complexity-diff` rescan both sides instead of trusting the snapshot — those are the details that decide whether a signal like this survives its first month. I especially like that `TestComplexityWorkflowIsAdvisoryAndPinned` asserts the workflow *can't* quietly become a required gate.

The four red tests are ours, not yours. They pass `COMPLEXITY_BASE_REF=origin/main`, and our `Integration / packages-core-*` shards check out at `fetch-depth: 1`, so there's no `origin/main` on the runner and `git archive` fails before any assertion runs. It works fine in a normal clone, which is exactly why you couldn't have seen it. I'll switch those to `HEAD` (the fake gocyclo ignores the extracted tree anyway), keep the deliberate missing-ref test intact, and re-run the shard. Your `fetch-depth: 0` in the complexity workflow itself is right and stays as-is.

One thing for later, not worth a round-trip: the `conformance` substring filter also excludes `cmd/gc/cmd_runtime_conformance.go`, which is real shipped code — flagging it before the numbers get quoted anywhere.


---

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277272529

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR adds an advisory cyclomatic-complexity tracker — a SHA-pinned `complexity.yml` workflow, a `scripts/ci/complexity.sh` analyzer wrapper with `update`/`report`/`diff`/`check` modes, a grandfathered `engdocs/complexity-baseline.json`, `make complexity-*` targets, and Go tests — which gives the project ongoing visibility into cyclomatic/cognitive complexity on PRs without turning it into a merge gate. That "measure first, ratchet later" posture is exactly the right way to introduce a new quality signal.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- None were required — the change was accepted without maintainer edits. The review instead independently validated the tool's two load-bearing claims against the worktree: (1) it is genuinely advisory/non-gating — `permissions: contents: read`, every working step is `continue-on-error: true`, the report/diff steps force `exit 0`, and there is no `needs:`/required edge, so it cannot fail a merge; and (2) the committed baseline stores only threshold offenders with stable `(package, function, file)` keys (all entries CCN ≥ 20, no duplicate keys today).

- **ga-la575** — [minor, maintainability] `make complexity-update` → `make complexity-check` is not round-trip safe for a same-key pair at/above threshold: the `diff` mode dedups duplicate `(package, function, file)` keys, but the shared parse used by `update`/`report`/`check` does not, so `update` can write a duplicate-key baseline (e.g. two `init()` over the threshold in one file) that `check` then hard-rejects. Latent today and fails loud, not silent. Not addressed in this PR.
- **ga-iv6nz** — [nit, maintainability] `check`'s positive path — failing on a brand-new at/above-threshold function — plus the env/tool error paths are untested; only the *regressed*-function case is currently pinned. A future refactor could silently disable `check`'s primary value. Not addressed in this PR.

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `32e0920` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/33348277673

### Review Iterations
1 review pass performed; approved on the first pass with no fix iterations required.

---

Issue #5789, opened 2026-08-31
Bounded paging for reads that filter Go-side: TierWisps bd-list leg and bdQueryEphemeralStatusShell


sjarmak:
Two read paths still fetch every bead in the store and discard most of the rows Go-side. Both are the same shape and both need a bounded page loop, so they belong in one issue rather than two halves of one decision.

### `TierWisps` in `bdListRequiresClientLimit`

`internal/beads/bdstore.go` forces `--limit 0` for `TierWisps` unconditionally. The tier merges two independently-fetched legs, and its `bd list` leg reads with `--include-templates`, which returns many ordinary issues that the Go-side tier filter (`Ephemeral || NoHistory`) then discards. A bd-side limit would cut rows before that filter runs and silently under-fetch, so the force is correct as written; it is the fetch strategy that is wrong.

`bd list` exposes no compound `(created_at, id)` seek flag, so paging needs a bounded loop that pages until the post-filter result set reaches the caller's limit or the source is exhausted, with a cap so a store full of non-matching rows cannot turn one read into an unbounded scan.

### `internal/config/workquery.go:92`

`bdQueryEphemeralStatusShell` emits `bd query ... --limit=0` for the same reason: the caller filters after the fact. Same remedy, same caveat about the missing seek flag.

### Why this is one issue

The decision to make is "how does a read that filters Go-side stay bounded", and it has these two surfaces. Answering it for one and not the other leaves the shared question open.

Context: #<PR> narrowed the force to `TierWisps` only, so the issues tier now takes a real limit. That change deliberately left both sites above alone.

Issue #2655, opened 2026-05-27
bug: hot active Beads list scans can wedge controller automation


mike-matchpoint:
## Before you continue

- [x] I searched existing issues and did not find a duplicate.
- [x] I read the relevant docs and contributor guidance.

## Gas City version

Upstream base checked: `main@e79aad7c`.

Observed on a main-derived local fork before the indexed active-read changes; fixed and soaked on a local candidate build reporting `gc version --json` as `version=1.1.1`, `commit=b3bef7b4`.

## Environment

- macOS 26.4.1 arm64, zsh, local/tmux provider
- Dolt `2.0.7`
- Beads `HEAD-f8b9400` / `1.0.4 (f8b940016)`

## Reproduction

1. Start a disposable Gas City using a Dolt-backed Beads store.
2. Seed the Beads store with many active rows using mixed status, label, metadata, dependency, parent, and ephemeral-tier shapes. The slow path is easiest to see once the active set is large enough that repeated full hydrated projections are visible in Dolt process samples.
3. Run controller-style foreground reads in a loop, for example active `gc bd list --json` shapes, `gc status --json`, `gc session list --json`, and order/session checks that repeatedly ask for active Beads state.
4. In parallel, sample Dolt with `SHOW FULL PROCESSLIST` and watch for repeated hydrated aggregate/list scans while the foreground commands are running.

A PR series will include bounded indexed active-read paths and parity tests so the reproduction does not depend on a private pack or private city.

## Expected behavior

Hot active reads should stay bounded and should not repeatedly hydrate the full active Beads projection when the query shape can be answered from indexed Dolt rows. Controller automation should continue to make progress under normal foreground/status polling.

## Actual behavior

Supported active read shapes currently fall back to hydrated Beads projections. Under a large active set and concurrent controller polling this can keep Dolt busy with repeated aggregate/list work, causing foreground commands and controller loops to slow down or wedge.

## Logs, screenshots, or traces

Public-safe local soak evidence after the candidate fix:

```shell
Dolt version: 2.0.7
Sample window: 2026-05-27T08:00:06Z through 2026-05-27T10:00:06Z
Samples: 25
Hard failures: []
gc status p95: 1s
gc session list --json p95: 1s
Max Dolt zombie count: 0
Max hydrated process rows: 0
Non-closed test rows: 0
```

Targeted candidate tests passed for the first schema-compatibility slice:

```shell
dolt version
# dolt version 2.0.7
[...]

PR #5790, opened 2026-08-31
fix(beads): stop forcing unbounded bd list for issues-tier reads

sjarmak:
## What

`bdListRequiresClientLimit` forced `--limit 0` for every issues-tier read, even
when the caller's own `Limit` was safe to push straight through to `bd list`.
Every such query fetched the whole table and discarded the tail in Go.

The force now applies only to the shapes that actually need it. A caller's limit
passes through when all of these hold: default or created-desc sort, no
`CreatedBefore`/`UpdatedBefore`, no metadata filter, no seek pagination, no ID
list, and at most one assignee. None of those calls carry a Go-side residual
filter that could drop rows before a bd-side limit truncates them: the underlying
list never returns ephemeral rows, and the issues-tier filter is a no-op against
what it does return.

## What is deliberately unchanged

The wisps tier keeps the unconditional force. Its list leg reads with
`--include-templates`, which returns ordinary issues that the Go-side tier filter
(`Ephemeral || NoHistory`) then discards, so a server-side limit there would
silently under-fetch. Bounding it needs a paged loop, because the list command
has no compound seek flag. `internal/config/workquery.go:92` carries the same
open-ended shape for the same reason. Both surfaces are one decision and are
tracked together in #5789.

## Test plan

- New table cases in `internal/beads/bdstore_test.go` cover each pass-through
  condition and each condition that still forces the unbounded read.
- Full suites pass: `internal/beads`, `internal/orders`, `internal/molecule`,
  `internal/mail/beadmail`, `internal/nudgequeue`, `internal/api`,
  `examples/gastown`, and all of `cmd/gc`.
- `go vet ./...` and lint on the changed files are clean.

---
- Related to #2655 — bounds one source of the unbounded active-bead list scans described there: gc's bd-list wrapper no longer forces `--limit 0` on every issues-tier read, so a caller's own limit is pushed down to bd instead of the whole table being fetched and trimmed in Go. It does not add the indexed active-read path that issue proposes, and the wisps-tier list leg plus the open-ended work-query read stay unbounded (tracked separately).
- Related to #3253 — removes the unconditional unbounded `bd list --limit 0` on issues-tier reads, so the bounded first-page molecule query in the beads(all=true) path can push its caller limit down to a bd-backed store instead of having it overridden; the formulas/feed path is untouched — its closed-history metadata query still forces the full scan — as are seeked cursor pages and the wisps tier
- Related to #4246 — removes the forced unbounded `bd list --limit 0` for issues-tier reads whose caller already supplies a safe limit, trimming part of the BdStore scan cost described there; it does not add the change-detection, caching, batching, or backoff that issue asks for, and the ephemeral/wisps legs and supervisor `--all` scans it measures are deliberately left unchanged

### DIFF
--- a/cmd/gc/cmd_bd_store_bridge_test.go
+++ b/cmd/gc/cmd_bd_store_bridge_test.go
@@ -370,10 +370,9 @@ func TestBdStoreBridgeListCommandForwardsFilters(t *testing.T) {
 	if err != nil {
 		t.Fatalf("ReadFile(args): %v", err)
 	}
-	// BdStore may need to merge and filter policy tiers client-side, so it
-	// requests all server-side candidates and applies the requested limit after
-	// parsing.
-	for _, want := range []string{"list", "--json", "--status=open", "--assignee=mayor", "--type=message", "--include-infra", "--include-gates", "--limit", "0"} {
+	// TierIssues + a single assignee has no Go-side-only residual filter, so
+	// BdStore pushes the caller's real limit straight to bd list (gc-i4j3y).
+	for _, want := range []string{"list", "--json", "--status=open", "--assignee=mayor", "--type=message", "--include-infra", "--include-gates", "--limit", "7"} {
 		if !strings.Contains(string(argsText), want) {
 			t.Fatalf("list args missing %q: %s", want, string(argsText))
 		}
--- a/examples/gastown/maintenance_scripts_test.go
+++ b/examples/gastown/maintenance_scripts_test.go
@@ -3306,7 +3306,7 @@ func TestReaperParentIDIsParentChildDependencyProjection(t *testing.T) {
 	runner := func(_, name string, args ...string) ([]byte, error) {
 		call := name + " " + strings.Join(args, " ")
 		switch call {
-		case "bd list --json --label=parent-projection --include-infra --include-gates --limit 0":
+		case "bd list --json --label=parent-projection --include-infra --include-gates --limit 50":
 			return []byte(`[
 				{
 					"id":"ga-child",
--- a/internal/beads/bdstore.go
+++ b/internal/beads/bdstore.go
@@ -2730,7 +2730,12 @@ func (s *BdStore) listViaBDList(query ListQuery) ([]Bead, error) {
 }
 
 func bdListRequiresClientLimit(query, serverQuery ListQuery, clientFilteredAssignees bool) bool {
-	if query.TierMode == TierIssues || query.TierMode == TierWisps {
+	// TierWisps always merges two independently-fetched legs (this bd-list
+	// leg plus the ephemeral leg in listWispsTier) and needs full candidates
+	// from both to union/dedupe/sort/limit correctly; TierIssues is the only
+	// tier reaching this function that reads a single, self-contained result
+	// set, so only it is eligible for a bd-side limit below.
+	if query.TierMode == TierWisps {
 		return true
 	}
 	if serverQuery.Sort == SortCreatedAsc || clientFilteredAssignees {
@@ -2744,6 +2749,12 @@ func bdListRequiresClientLimit(query, serverQuery ListQuery, clientFilteredAssig
 	// bd-side limit would cut rows before that filter runs — fetch unbounded
 	// and let applyListQuery filter then limit.
 	if serverQuery.SeekAfter != nil {
+		return true
+	}
+	// IDs is a Go-side-only residual filter (see ListQuery.Matches): bd list
+	// has no --id flag, so a bd-side limit could truncate before the
+	// matching IDs are even fetched.
+	if len(serverQuery.IDs) > 0 {
 		return true
 	}
 	return false
--- a/internal/beads/bdstore_storage_conformance_test.go
+++ b/internal/beads/bdstore_storage_conformance_test.go
@@ -331,10 +331,13 @@ func TestBdStoreListStorageTierConformance(t *testing.T) {
 			wantIDs: []string{"bd-history", "bd-no-history"},
 		},
 		{
-			name:                  "issues tier applies limit after client storage filtering",
-			query:                 beads.ListQuery{Label: "scope", Limit: 1},
-			wantIDs:               []string{"bd-history"},
-			wantUnlimitedPrelimit: true,
+			// gc-i4j3y: issues-tier bd list never returns ephemeral rows (no
+			// --include-templates), so the Go-side matchesTier filter is a
+			// no-op backstop here — a bounded caller Limit is safe to push
+			// down to bd list directly instead of forcing --limit 0.
+			name:    "issues tier applies limit after client storage filtering",
+			query:   beads.ListQuery{Label: "scope", Limit: 1},
+			wantIDs: []string{"bd-history"},
 		},
 		{
 			name:                  "wisps tier keeps no-history and ephemeral rows",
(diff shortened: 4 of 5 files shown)


sjarmak:
Cross-provider review record, run against the exact head `e17715208`.

Claude leg: clean, no blockers and no majors.
Codex leg (adversarial/multi-model): clean, no blockers and no majors.

Both legs were run before the branch was pushed, at the same commit that is now the PR head, so no rebase or fixup has moved it since. The review focused on the condition that decides when a caller's limit may pass through to the underlying list, and specifically on whether any pass-through shape carries a Go-side residual filter that could drop rows before a server-side limit truncates them. None does: the list never returns ephemeral rows, and the issues-tier filter is a no-op against what it returns. The wisps tier is deliberately left on the unconditional force, with the reason stated in the PR body and the follow-up decision tracked in #5789.

quad341:
@sjarmak — this is a good catch and a careful one. `TierIssues` is the zero value of `TierMode`, so that unconditional `--limit 0` meant essentially every bounded read in the city was fetching the whole table and throwing away the tail in Go. What makes this easy to trust is that you didn't just delete the guard: you enumerated each residual Go-side filter, added the `IDs` guard you found on the way, and left the wisps tier alone with a written reason for why its two-leg merge still needs the full fetch. The `IDs` guard is a real fix beyond the stated scope — that hole was already reachable on the `TierBoth` path.

I'll take three small things on my side and merge:

- The new comment says `TierIssues` is the only tier reaching that function, but `listBothTiers` calls it with `TierBoth` too — and `query.go` notes the policy layer rewrites TierIssues→TierBoth for work stores. Behavior's unchanged; just rewording.
- The `SortCreatedAsc` test comment points at a `...ForcesUnboundedForSortCreatedDesc` twin that doesn't exist.
- Adding a `TierBoth` + `Limit` conformance case, since the existing both-tiers case has no `Limit` and skips the assertion.

One I'm noting rather than blocking on: nothing pins `bd list --limit N`'s truncation order against a real bd, so the created-desc pass-through rests on the same assumption `canApplyWispsServerLimit` already makes for the `bd query` leg. A contract test would retire it for both call sites at once — nice thing to fold into #5789 if you're up for it.


---

_Heads-up: we hit a transient network/publish issue pushing the one-line maintainer test-bound to this branch — nothing needed from you. The review stands; we'll retry the merge shortly._


<details>

```diff
From fc4fa32d8f7763f8003e3d76dd675326cbb5d376 Mon Sep 17 00:00:00 2001
From: sjarmak <t@t.co>
Date: Sun, 30 Aug 2026 21:03:27 -0700
Subject: [PATCH] fix(beads): correct tier comment and cover TierBoth limit
 pushdown

Review follow-up to the issues-tier limit-pushdown change.

The comment on bdListRequiresClientLimit claimed TierIssues is the only
tier reaching the function. It is not: listBothTiers sets TierBoth and
calls listViaBDList, and the bead-policy layer rewrites TierIssues reads
to TierBoth before they reach the backend (see FederatedReadTier in
query.go), so TierBoth is the dominant shape at this gate on a live city.
Restate why TierWisps forces the unbounded read (--include-templates plus
a Go-side tier filter that discards the ordinary issues bd returns) and
name both eligible tiers.

Add the missing TierBoth + Limit conformance case: the existing "both
tiers" case set no Limit, so the `--limit 0` assertion behind the
`if tc.query.Limit > 0` guard never ran for the shape that now takes the
pushdown path.

Repoint a dangling cross-reference in bdstore_test.go from the
nonexistent TestBdStoreListForcesUnboundedForSortCreatedDesc to the real
TestBdStoreListPushesCallerLimitForSortCreatedDesc.

Also note: the PR body's `internal/config/workquery.go:92` breadcrumb for
the sibling open-ended read should read `workquery.go:251`
(bdQueryEphemeralStatusShell, the `--limit=0` wisp read).
---
 internal/beads/bdstore.go                          | 12 +++++++-----
 internal/beads/bdstore_storage_conformance_test.go | 12 ++++++++++++
 internal/beads/bdstore_test.go                     |  7 ++++---
 3 files changed, 23 insertions(+), 8 deletions(-)

diff --git a/internal/beads/bdstore.go b/internal/beads/bdstore.go
index 8ad600f73..f4f9a0172 100644
--- a/internal/beads/bdstore.go
+++ b/internal/beads/bdstore.go
@@ -2730,11 +2730,13 @@ func (s *BdStore) listViaBDList(query ListQuery) ([]Bead, error) {
 }
 
 func bdListRequiresClientLimit(query, serverQuery ListQuery, clientFilteredAssignees bool) bool {
-	// TierWisps always merges two independently-fetched legs (this bd-list
-	// leg plus the ephemeral leg in listWispsTier) and needs full candidates
-	// from both to union/dedupe/sort/limit correctly; TierIssues is the only
-	// tier reaching this function that reads a single, self-contained result
-	// set, so only it is eligible for a bd-side limit below.
+	// TierWisps forces the unbounded read because its bd-list leg reads with
+	// --include-templates and the Go-side tier filter (Ephemeral || NoHistory)
+	// then discards the ordinary issues that come back, so a bd-side limit
+	// would silently under-fetch. TierIssues and TierBoth both reach this
+	// function (see listBothTiers) with a tier filter that is a no-op against
+	// what their bd list returns, so both are eligible for the bd-side limit
+	// decided below.
 	if query.TierMode == TierWisps {
 		return true
 	}
diff --git a/internal/beads/bdstore_storage_conformance_test.go b/internal/beads/bdstore_storage_conformance_test.go
index fa0114a83..6e9940952 100644
--- a/internal/beads/bdstore_storage_conformance_test.go
+++ b/internal/beads/bdstore_storage_conformance_test.go
@@ -354,6 +354,18 @@ func TestBdStoreListStorageTierConformance(t *testing.T) {
 			wantIncludeTemplates: true,
 			wantEphemeralQuery:   true,
 		},
+		{
+			// TierBoth's tier filter is a no-op and both legs are limited
+			// independently, so the caller's limit pushes down. This is the
+			// shape work stores actually see: the bead-policy layer rewrites
+			// TierIssues reads to TierBoth before they reach the backend
+			// (see query.go's FederatedReadTier note).
+			name:                 "both tiers pushes caller limit to bd list",
+			query:                beads.ListQuery{Label: "scope", TierMode: beads.TierBoth, Limit: 1},
+			wantIDs:              []string{"bd-history"},
+			wantIncludeTemplates: true,
+			wantEphemeralQuery:   true,
+		},
 	}
 
 	for _, tc := range cases {
diff --git a/internal/beads/bdstore_test.go b/internal/beads/bdstore_test.go
index 086f3d877..4a436e214 100644
--- a/internal/beads/bdstore_test.go
+++ b/internal/beads/bdstore_test.go
@@ -3088,9 +3088,10 @@ func TestBdStoreListPushesCallerLimitForSortCreatedDesc(t *testing.T) {
 	}
 }
 
-// TestBdStoreListForcesUnboundedForSortCreatedAsc is the SortCreatedAsc twin
-// of TestBdStoreListForcesUnboundedForSortCreatedDesc, confirming the
-// pre-existing guard still holds under the relaxed TierIssues condition.
+// TestBdStoreListForcesUnboundedForSortCreatedAsc is the SortCreatedAsc
+// counterpart to TestBdStoreListPushesCallerLimitForSortCreatedDesc,
+// confirming the pre-existing guard still holds under the relaxed TierIssues
+// condition.
 func TestBdStoreListForcesUnboundedForSortCreatedAsc(t *testing.T) {
 	runner := fakeRunner(map[string]struct {
 		out []byte
-- 
2.55.0
```

</details>

PR #5791, opened 2026-08-31
fix(doctor): recover panics in Check.Fix and pin the Wait contract

sjarmak:
## What

A panicking `Check.Fix` took down the whole `gc doctor --fix` process, on both
the unbounded and the timed path. Only `Check.Run` was recovered. `Fix` runs
pack-authored remediation, so it is the likelier of the two to blow up, and one
bad pack should fail one check rather than the command.

Both call sites now route through `runFixRecoveringPanic`, which reports the
panic as that check's `FixError` and leaves the rest of the run intact.

## The Wait contract

`Wait` always relied on a one-run-at-a-time contract that was never written down.
A concurrent `Run` could `Add` after `Wait` had already observed a zero counter,
so `Wait` could return while a timed execution still owned caller resources. The
doc comment now states it.

`doDoctor` deliberately does not call `Wait`: an abandoned check writes only to
its own private buffer, and blocking on it would restore the hang the per-check
timeout exists to prevent. That reasoning is recorded at the call site so the
next reader does not add the `Wait` back.

## Test plan

- `TestPanickingFixDoesNotCrashTheRun` drives a check whose `Fix` panics, once
  with no timeout and once with a timeout, covering both branches. It asserts the
  run completes, the check reports an error status, the fix is marked attempted,
  and the panic text reaches `FixError`.
- `internal/doctor` and `cmd/gc` suites pass; `go vet` and lint are clean.

### DIFF
--- a/cmd/gc/cmd_doctor.go
+++ b/cmd/gc/cmd_doctor.go
@@ -431,6 +431,12 @@ func doDoctor(fix, verbose, jsonOut bool, checkTimeout time.Duration, stdout, st
 		return 1
 	}
 
+	// Deliberately no d.Wait() here: gc doctor's whole point in bounding a check
+	// is that a wedged one cannot stall the command, and waiting on an abandoned
+	// goroutine would restore exactly that hang. Nothing this function owns
+	// outlives the process, and ctx holds no handle a late writer can corrupt --
+	// an abandoned check writes only to its own private buffer. A future caller
+	// that reuses a Doctor in-process must call Wait before releasing ctx.
 	d := &doctor.Doctor{CheckTimeout: checkTimeout}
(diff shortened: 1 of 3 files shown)


sjarmak:
Cross-provider review record, run against the exact head `69000b79f`.

Claude leg: clean, no blockers and no majors.
Codex leg (adversarial/multi-model): clean, no blockers and no majors.

Both legs first ran against the earlier head `600aab411` and both FAILED it independently, on two different real defects: a panicking `Check.Fix` crashing the whole process, reproduced live rather than inferred, and the `Wait` resource-lifetime race. Those fixes are commit `69000b79f`. Because that fixup is our own unreviewed code sitting on top of the reviewed head, the gate was treated as re-armed and both legs were re-run from scratch against `69000b79f`, which is the PR head now. The PASS above is the second run, not the first.

Issue #5546, opened 2026-08-23
deacon-patrol formula: next-iteration fallback assigns without claiming — open wisp invisible to the restarted session's startup hook check


mywinecity:
## Symptom

Repeated "boot catches idle deacon" incidents: after the deacon pours the next patrol iteration via the formula's documented fallback snippet and the session later restarts, the fresh session's startup hook check finds nothing and the deacon sits idle until an external nudge (we observed 45min and 13min idle windows in one lineage, both self-corrected only by manual intervention).

## Root cause

`formulas/mol-deacon-patrol.toml` — the next-iteration / "CRITICAL: No Idle State Between Cycles" fallback block does:

    gc bd update "$NEXT" --assignee="$GC_AGENT"

with no status change. Verified live (gc 1.4.0): after running exactly that command, `gc bd show` reads back `status=open`, not `in_progress`. The standard startup hook check looks for `--status in_progress` first, so an open-but-assigned wisp is invisible to it.

## Suggested fix (one line)

    gc bd update "$NEXT" --claim

`bd update --help` documents `--claim` as "Atomically claim the issue (sets assignee to you, status to in_progress; idempotent if already claimed by you)" — closes the gap structurally instead of relying on every cycle remembering an extra step.

## Environment

gc 1.4.0 (nixpkgs), gastown pack pinned; found by the deacon itself during patrol on 2026-08-23 and confirmed by read-back.

A3Ackerman:
Measured on a live bd 1.0.5 store: `bd ready --include-ephemeral --assignee=X` returns `[]` for an assigned **OPEN** wisp that `bd query "ephemeral=true AND status=open AND assignee=X"` returns immediately. So on bd 1.0.5+ the assigned-ready path skips its wisps probe behind a version guard, while the in-progress twin (`ephemeralAssignedInProgressProbeScript`) keeps its probe — which is exactly why an assigned *in_progress* wisp is always found and its *open* sibling is invisible until someone hand-mutates the row, the asymmetry this issue describes.

#5678 measured the same `--include-ephemeral` blindness at the `bd list` call site.

PR incoming: `ephemeralAssignedReadyProbeScript` already carries the correct bounded, readiness-aware probe for the bd-1.0.4 tier; the fix removes the single guard that skipped it under `--include-ephemeral`, so the probe runs for every bd version. It is purely additive (runs only after the `bd ready --include-ephemeral` tier returns nothing).

A3Ackerman:
[...]

PR #5792, opened 2026-08-31
fix(workquery): bd-1.0.5 tier — restore the defensive wisps probe the bd-1.0.4 tier already carries

A3Ackerman:
The bd 1.0.5+ ready tier delegates the wisps tier to `bd ready
--include-ephemeral`, on the premise that the flag reads it. Measured against a
live store it does not: `bd ready --include-ephemeral --assignee=X` returns []
for an assigned OPEN wisp that `bd query "ephemeral=true AND status=open AND
assignee=X"` returns immediately. So an agent whose formula pours a successor
wisp, assigns it, and leaves it open sees no ready work until the row is
hand-mutated to in_progress -- the tell, since the in-progress twin
(ephemeralAssignedInProgressProbeScript) has always carried a real wisps probe,
which is why in_progress wisps were found and their open siblings were not.

ephemeralAssignedReadyProbeScript already carries the correct probe for the
bd-1.0.4 tier: it reads the ephemeral open rows server-side (bounded per agent),
tries the dependency_count == 0 candidate, then falls back to a
dependency_count > 0 candidate enriched via a real bd show, and applies the epic
and blocking-dependency readiness rules -- `bd query status=open` is not a
readiness query. The defect is a single guard that skipped that probe whenever
the ready tier used --include-ephemeral. Removing it lets the probe run for
every bd version.

Purely additive: the probe runs only after the `bd ready --include-ephemeral`
tier above it returns nothing, and it is identical across versions, so pool
scaling and the bd-1.0.4 path are untouched. Goldens confirm the blast radius --
only the bd105 AssignedReady/Work scripts change; every bd104, pool-demand,
on-boot, on-death and in-progress golden is byte-identical.

The probe stays hold-transparent -- an assignee-scoped existence probe, not the
in_progress work-serving gate -- pinned by the existing
TestEphemeralAssignedReadyProbeScriptDoesNotExcludeDispatchHoldLabels.

Two tests, both mutation-confirmed: restoring the guard hides the open wisp
(TestEffectiveWorkQueryBD105SurfacesEphemeralOpenAssignedWork); dropping the
readiness rules would serve a dependency-blocked open wisp
(TestEffectiveWorkQueryBD105EphemeralOpenProbeRespectsReadiness).

## Scope, and an explicit ask for the maintainers

This is **not** the fix for #5546's reported symptom. The reporter's own diagnosis was right and minimal: the deacon-patrol fallback pours the next wisp and assigns it without moving it to `in_progress`, so the restarted session's startup check misses it. That fix belongs in the formula, and it is already in flight as gastownhall/gascity-packs#264 (`--status=in_progress` on exactly those lines; the reporter's `--claim` spelling is equivalent and arguably more idempotent). This PR should not be read as resolving that issue.

What this PR is: a **defensive-parity** fix for the non-default `bd_compatibility = "bd-1.0.5"` tier. The default `bd-1.0.4` tier already carries the wisps probe unconditionally (config.go documents that tier as existing precisely to avoid bd flags that are "unavailable or incomplete"); the 1.0.5 tier skipped the same probe on the premise that `bd ready --include-ephemeral` covers it, and measured against a live 1.0.5 store it does not.

**The ask:** is serving an *assigned, open* ephemeral wisp as "ready" the intended semantics for this tier, or should assigned-but-unclaimed remain invisible by design (with the formula fix as the only correct remedy)? We can see the argument either way and would rather have a maintainer choose the layer before this merges. If the answer is "formula only", withdraw this and we carry nothing — our fork already runs the formula fix.

Related: #5546 (symptom, fixed by gascity-packs#264), #5678 (the same `--include-ephemeral` blindness at the `bd list` call site).

Corroboration: #5678 measured the same `--include-ephemeral` blindness at the `bd list` call site.


A3Ackerman:
Withdrawing. This patches the bd-1.0.5 tier while the pin is moving to 1.3.0; not worth a review slot. The branch stays on my fork in case the tier outlives the move.

PR #5794, opened 2026-08-31
fix(order): bound gc order check's order.fired event read

sjarmak:
## The read

`gc order check` resolved each order's last firing by asking the event provider for every `order.fired` event ever recorded:

```go
firedEvents, _ = ep.List(events.Filter{Type: events.OrderFired})
```

`List` reaches `ReadFiltered`, which walks every retained archive oldest-first, gunzipping and JSON-decoding each line before the type predicate is consulted. The cost grows with every rotation and is paid on every invocation. Measured at 76 seconds against 37 archives totaling 2.16GB.

## The change

Prefer the bounded tail read when the provider offers one, and keep the old call as the fallback:

```go
if tp, ok := ep.(events.TailProvider); ok {
    firedEvents, _ = tp.ListTail(filter, orderCheckFiredEventTailLimit)
} else {
    firedEvents, _ = ep.List(filter)
}
```

On this command's path the provider is `*events.FileRecorder`, whose `ListTail` already exists. `ReadFilteredTail` opens only the active log, walks it backward in chunks, and applies the type filter before counting toward the limit, so the limit bounds matching events rather than raw lines. Archives are never opened. The limit of 2000 and the reasoning behind it mirror the equivalent read in `internal/doctor`.

## Why bounding is safe

`latestFired` feeds exactly one thing: a cooldown fast path that returns early when an order fired more recently than its interval. Every other route, including every non-cooldown trigger, falls through to `orders.LastRunAcross`, which is authoritative. A truncated or stale tail therefore costs the shortcut and nothing else.

The argument rests on the shortcut being not-due-only. Losing a fired event can only move an order toward due, never away from it, so the bound can advance a firing but cannot suppress one. It deliberately does not rest on timestamp ordering between the event and the order-run history: `LastRunAcross` returns the newest order-run evidence, which a wisp root labeled after the event, or a later manual `gc order run`, can push past the event's own timestamp.

## One behavior does change

While the shortcut fired, `baseLastRunFn` was never called, so a failing `LastRun` read on that order's store stayed masked. An order whose event has been evicted from the tail now reaches that read, and `lastRunErr` aborts the check. No bounded read can recover an evicted event, so this is the cost of not walking the archives: a store failure that used to sit behind a cached event is now reported. It is recorded at the constant rather than coded around.

## Tests

Two cases, both asserting the read path stays bounded and that the limit passed is the constant, since a non-positive limit would silently read unbounded:

- An order whose fired event is pushed out of the window by more newer events than the limit. The bounded read cannot see it, so the not-due result can only come from the order-run fallback. Removing the fallback makes this test fail with "due / never run", which is what makes it a guard rather than a restatement.
- An order that never fired and has no order-run history, confirming the bound does not turn an empty tail into a false negative.

## Not included

The sibling unbounded read in `internal/orders/triggers.go` `checkEvent` is left alone. It post-filters its results, discarding order-tracking bookkeeping events after the query returns, so a fixed `Limit` there is not a bound on the answer: a window full of bookkeeping events would yield zero matches and report an event trigger as never satisfied. That read needs a different fix and is tracked in #5191.

### DIFF
--- a/cmd/gc/city_registry.go
+++ b/cmd/gc/city_registry.go
@@ -385,6 +385,12 @@ func (p transientCityEventProvider) List(filter events.Filter) ([]events.Event, 
 	return events.ReadFiltered(p.path, filter)
 }
 
+// ListTail returns the trailing matching events without scanning the full
+// archived history, implementing events.TailProvider.
+func (p transientCityEventProvider) ListTail(filter events.Filter, limit int) ([]events.Event, error) {
+	return events.ReadFilteredTail(p.path, filter, limit)
+}
+
 func (p transientCityEventProvider) LatestSeq() (uint64, error) {
 	return events.ReadLatestSeq(p.path)
 }
--- a/cmd/gc/city_registry.go
+++ b/cmd/gc/city_registry.go
@@ -385,12 +385,6 @@ func (p transientCityEventProvider) List(filter events.Filter) ([]events.Event, 
 	return events.ReadFiltered(p.path, filter)
 }
 
-// ListTail returns the trailing matching events without scanning the full
-// archived history, implementing events.TailProvider.
-func (p transientCityEventProvider) ListTail(filter events.Filter, limit int) ([]events.Event, error) {
-	return events.ReadFilteredTail(p.path, filter, limit)
-}
-
 func (p transientCityEventProvider) LatestSeq() (uint64, error) {
 	return events.ReadLatestSeq(p.path)
 }
--- a/cmd/gc/city_registry_test.go
+++ b/cmd/gc/city_registry_test.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"errors"
+	"fmt"
 	"io"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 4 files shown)


sjarmak:
Cross-provider review record, run against the exact head `7430efee9d3bef9699ba8229faa548e4239a688a`. The branch is Claude-authored, so the adversarial leg is Codex.

Four passes ran, each against a specific head, with the branch changing between them. Findings and dispositions:

**Pass 1 (Claude, head `6f773d008`), one major.** The `ListTail` method added to `transientCityEventProvider` is not on the `gc order check` path: that command resolves its provider through `newEventsProviderForNameWithConfig`, which returns `*events.FileRecorder`, and `FileRecorder.ListTail` already exists on main. The method's only effect was to make the registry type newly satisfy `events.TailProvider`, silently switching `storehealth.LastMaintenance` to a bounded, `MaxScanBytes`-clamped window for registry-listed cities. Accepted; the hunk and its test were removed, so `cmd/gc/city_registry.go` is now identical to main. The removed test was also shown to be vacuous: `ReadFilteredTail` opens only the file it is given, so no archive count could ever have failed it.

Also accepted from that pass: the regression test claimed to cover an order whose fired event falls outside the window but never recorded that event, duplicating the never-fired case. It now records the firing before the crowding noise. Codex confirmed by mutation that removing the fallback makes it fail with "due / never run", so it guards the property rather than restating it.

**Pass 2 (Codex, head `5266e4f22`), PASS.** Verified that `latestFired` is read only in the cooldown fast path on both output paths, walked every trigger type, confirmed `ReadFilteredTail` filters before counting toward the limit, and confirmed the tracking bead is persisted before `OrderFired` is recorded.

**Pass 3 (Codex, head `aa166eb1f`), one major, one minor.** The minor retracted an earlier justification of mine: I had argued the bound is safe because the tracking bead is the older record, but `orders.LastRunAcross` returns the newest order-run evidence, which a wisp root labeled after the event or a later manual `gc order run` can push past the event's timestamp. The conclusion survives on firmer ground, which is now what the comment says: the event is consulted only to return not-due early, so losing it can only move an order toward due.

The major: while the shortcut fired, `baseLastRunFn` was never called, so a failing `LastRun` read stayed masked; an order whose event has been evicted now reaches that read and `lastRunErr` aborts the check. No bounded read can recover an evicted event, so this is inherent to bounding rather than a fixable defect. It is documented at the constant and called out in the PR body rather than coded around. The reviewer's suggested remedy, retaining the full event lookup, would undo the change.

**Pass 4 (Codex, head `7430efee9`, the published head), PASS with zero findings.** "The comment accurately describes the implementation. ListTail returns up to 2,000 matching order.fired events from the active log, without archive traversal. The bounded result is a subset of the former result, and only cooldown triggers use it for a not-due shortcut; all other triggers use unchanged paths."

One finding was rejected rather than fixed. Pass 3's first run reported a major claiming the branch deletes `transientCityEventProvider.ListTail` and regresses multiplexer tail reads to archive-scanning. That was an artifact of how I framed the prompt: I described the delta as two commits on a previously-reviewed head while supplying the full branch diff, so the reviewer read the absence of a hunk it had seen before as a deletion. `git diff origin/main...HEAD -- cmd/gc/city_registry.go` is empty, and the method never existed on main, so dropping an addition cannot regress against it. Re-run with `origin/main` as the stated baseline, the reviewer verified the file is identical and withdrew the finding.

Gates at the published head: `go build ./...`, `go vet ./...`, `golangci-lint` on the changed package (0 issues), and the sharded pre-push suite (all fast jobs passed, including all six `cmd/gc` shards and `unit-core`).

PR #5795, opened 2026-08-31
feat(push-ownership-guard): accept undeclared single in-progress match as branch-bead successor

quad341:
## Summary
- Extends `_pog_resolve_bead_id` (push-ownership-guard.sh) with a third tier: when a branch-derived bead is confirmed closed/inactive and no bead declares itself its successor via metadata, a session holding exactly one *other* in-progress bead (no declared link required) may use it as an undeclared successor — but only after that candidate independently passes the same live-ownership predicate `assert_bead_still_claimed` already applies (status, assignee identity, `gc.routed_to`, hold labels).
- Extracts that predicate out of `assert_bead_still_claimed` into a shared `_pog_ownership_violation` helper so the new tier and the public push gate share one definition of "owned," verified byte-for-byte message-preserving against the prior inline checks.
- Reuses the in-progress list already fetched for Tier 1 (no second `bd list` call). The new tier's own I/O (`_pog_branch_id_bead_inactive`) is ordered last in its condition chain, gated behind the cheap already-in-hand cardinality check, so an ordinary no-reuse push still costs exactly one `bd show`, not two.

## Motivation
Implements `ga-0ywrmy.1` (architecture: `ga-0ywrmy`). Closes the recurring "pr-audit rotates the live bead hourly while the branch name stays pinned to the closed originating bead" false-block pattern, without reopening the cardinality risk the existing tests guard against — 2+ undeclared in-progress beads leaves the tier inactive.

## Review note
Per the bead's own architecture notes: *"Treat with commensurate review scrutiny; consider asking for a second review pass given the safety-critical, fleet-wide blast radius, before this merges."* This script is sourced into every fleet push (Layer A pre-push hook, Layer B `attempt_bounded_self_rebase` — no bypass by design). Flagging for reviewer to judge whether a second pass (e.g. `gascity/qwen-reviewer`) is warranted before merge.

## Test plan
- [x] `bash scripts/test-push-ownership-guard.sh` — 40/40 pass (rewrote 1 test to use an id-aware fake `bd`, added 2 new tests: acceptance of a genuinely-owned undeclared single match, non-activation with 2+ undeclared in-progress beads)
- [x] `shellcheck scripts/push-ownership-guard.sh scripts/test-push-ownership-guard.sh` — clean
- [x] `go build ./...` — clean
- [x] `go vet ./...` — clean
- [x] Pre-push fast gate (`./scripts/test-local-parallel fast`) — all 10 jobs passed

### DIFF
--- a/scripts/push-ownership-guard.sh
+++ b/scripts/push-ownership-guard.sh
@@ -134,6 +134,68 @@ _pog_branch_id_bead_inactive() {
     [[ "$st" != "in_progress" && "$st" != "open" && -n "$st" ]]
 }
 
+# _pog_ownership_violation <json>: pure predicate over an already-fetched
+# `bd show <id> --json` payload (parsed here, not re-read). Prints nothing
+# and returns 0 when the bead reads as currently owned by this session;
+# otherwise prints a short, undecorated reason and returns 1. Checks, in
+# order: status open/in_progress, assignee among this session's live
+# identities, gc.routed_to unset-or-matching, no hold:mayor/hold:external
+# label. Shared by assert_bead_still_claimed (the public push gate, which
+# wraps the reason into its own BLOCKED/--no-verify message below) and
+# _pog_resolve_bead_id's undeclared-single-match tier (ga-0ywrmy.1, which
+# discards the reason and uses only the pass/fail to decide whether an
+# undeclared in-progress bead may stand in for a closed branch-derived one).
+#
+# NOTE: never name a local here 'status' — it is a zsh special parameter
+# (linked to $?, alongside $pipestatus) and this file is sourced into the
+# deployer's ambient zsh shell (ga-xi7wi6); binding a local named 'status'
+# there is a read-only-variable error, not a shadow.
+_pog_ownership_violation() {
+    local json="$1"
+    local bead_status assignee routed_to labels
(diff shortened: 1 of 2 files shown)


quad341:
@quad341 — the part of this I want to single out isn't the new tier, it's the test fixture. Noticing that the old single-canned-response fake `bd` couldn't tell "the guard checked `ga-unrel8d` and rejected it" from "the guard never looked at it," and then rebuilding it id-aware with `show-ids.log` assertions — including the negative one that neither candidate is queried at cardinality ≥2 — is the kind of thing most rewrites quietly skip. Pulling the five ownership checks into `_pog_ownership_violation` so resolution and verification can't drift apart is the right structural move too; I checked the reason fragments concatenate back into the old BLOCKED strings byte-for-byte, and they do. 40/40 and shellcheck clean on my end as well.

Your second-pass request is honored — I'm escalating rather than merging on the spot. I reproduced the case where a mayor stand-down and a pr-audit rotation look identical to a cardinality check, so I want the operator to sign off on that trade explicitly rather than take `ga-0ywrmy` on faith from the PR body. Separately: this suite is path-filter-gated out of CI on a shell-only diff, so nothing green here tested your change. I'll file that.


---

PR #5797, opened 2026-08-31
RFC: city-qualified cross-city mail addressing ([mail.crosscity])

A3Ackerman:
## RFC — asking for design buy-in before review

This PR is held as a **draft on purpose**. It is a complete, flag-gated, inert-when-off implementation, but it proposes a *topology* upstream has no reference deployment for, and that deserves a yes-or-no before anyone spends review cycles on ~1,600 lines.

**The topology.** Several independently operated cities share one physical store for the `messaging` bead class — a relocated `[storage.classes.messaging]` binding, which #4017 made a first-class front door. In that shape a cross-city mail is a plain read on the receiving side; the only missing piece is an address form that says *which* city a name belongs to. That is all this PR adds: `<city>/<address>`, split on the first `/`, roster-checked, opaque past the city segment.

**How it relates to `gc context`.** `gc context` switches a whole session onto a remote city's HTTP+SSE API — kubectx for cities. This is different and complementary: one address space over a shared store, so a plain `gc mail reply` on the receiving side routes back across cities with no context switch. If maintainers would rather cross-city mail always travel the control plane (the #5386 direction), this PR should not land — and we would like to know that now rather than after review.

**Sender trust, stated explicitly.** `--from` with a foreign city segment is accepted verbatim. That adds no authentication and removes none: a shared messaging store already means every writer is trusted, exactly as it does for any bead today — a caller who can write to the store can already forge any sender. The PR does not move that boundary and does not claim to. If upstream wants sender attestation, it belongs in the store/transport layer, not in address parsing.

**Reference deployment.** Our fleet (three cities over one hub store) is what this was built for; we can supply a `docs/guides` walkthrough with the real configuration if there is appetite.

**The ask:** appetite yes/no on the address form and the shared-store topology. On a no, we withdraw and carry nothing — the flag-gated shape means the fork does not depend on it.

---

## What

Opt-in **city-qualified mail addressing** behind a new `[mail.crosscity]` config section. When configured, a mail recipient may take the form `<city>/<address>`: the first path segment names a city, and everything after the first `/` is that city's own address form, opaque to every other city.

```toml
[mail.crosscity]
# city = "qlandia"          # optional; defaults to the effective city name
cities = ["gastown", "westeros"]
```

Behavior with the section present:

- **`<local city>/<addr>` and `<addr>` are one mailbox.** Sends canonicalize the local-qualified form to the bare form; reads serve mail addressed either way (inbox/check/count recipient sets gain the city-qualified alias, on both the CLI and API read paths).
- **A listed peer city's address is canonical as written and is never looked up in the local session store.** Fresh send was the one surface that refused any `/`-containing identifier it couldn't resolve locally; reply and inbox already accepted the namespace. This closes that asymmetry.
- **Senders on cross-city mail are stored city-qualified** (`<local city>/<sender>`), so a plain reply on the receiving side resolves back to the sending city. Applied consistently on CLI send, CLI reply into a foreign-origin thread, API send, and API reply.
- **`--from` carrying a peer city's segment is accepted verbatim** — only its own city can check that identity. A local-city-qualified `--from` strips to the bare form and stays identity-checked; an unknown local sender is still refused.
- **`--notify` on a cross-city recipient is a typed refusal before any write.** No cross-machine wake primitive exists; a nudge queued into a leg that is not there must not report success. The recipient's wake belongs to its own city's mail sweep.
- **An address whose first segment names neither a rig nor a known city refuses as `unknown city "<x>" (known cities: ...)`** — never as a session lookup failure — so a stale or missing roster entry is diagnosable on the first occurrence.
- **Config validation refuses a rig named like any roster city (the local city included):** the first mail-address segment must bind unambiguously to either a rig or a city, never both.

Absent `[mail.crosscity]`, every new branch is disabled by the roster's zero value and behavior is unchanged — pinned by no-roster guard tests on both the CLI and API surfaces.

## Why

Multi-city deployments — several Gas City instances coordinating over a shared store — have no way to address mail across city boundaries, so every crossing becomes an out-of-band relay with at least one hop that reports success into nothing. #4017 routed all non-work bead access through coordination-class front doors, making the messaging class's *storage* pluggable (`[storage.classes.messaging]` can relocate it). This PR completes the *addressing* side of that seam: with a resolvable cross-city namespace, a relocated shared messaging store serves cross-city delivery as a plain read, with no relay, no wake hop, and exactly-once by construction (one bead is the message).

This slice is deliberately inert: it makes city-qualified addressing work and be testable **within one store**, and moves nothing. Pointing the messaging class at a shared store is deployment configuration via the existing `[storage]` machinery, not part of this change.

## Design notes for reviewers

- The resolver core is a pure, dependency-free type (`mail.CityRoster`) with the CLI and API each wiring it at their existing resolution seams; the config package stays a leaf (no `internal/mail` import).
- `resolveLiveConfiguredNamedMailTarget`'s rejection of `/`-containing identifiers is intentionally untouched: the city split happens above it, and locally-stripped remainders flow through existing resolution unchanged.
- `[mail]` composes wholesale across config layers (existing section semantics); `crosscity` inherits that contract.
- API reads stay open-world for unresolved recipients exactly as today; the read-side alias expansion also covers unresolved names, which is deliberate — any local box `<addr>` must serve mail addressed to `<local city>/<addr>`, that is what makes delivery a read.
- A foreign remainder gets the same minimal normalization local resolution applies (surrounding whitespace, one trailing `/`), so a send to `gastown/mayor/` cannot mint a mailbox the destination's reads would never serve; the split stays single-level and the remainder otherwise opaque.
- With the roster enabled, a reply whose thread origin cannot be read fails closed — without the origin, the cross-city rules (notify refusal, sender qualification) cannot be applied.
- One deliberate message change on the disabled path: the invalid `--from` refusal now names the sender the caller gave; previously it printed the zeroed resolution result (`invalid sender ""`). Pinned by tests.
- Known follow-up: `--notify` on a reply into a thread this city *originated* cross-city (stored From `<local city>/<sender>`) still attempts a local nudge against the qualified form and reports "nudge failed"; teaching the nudge resolver the local-strip is the right home for that fix and is left out of this slice.
- Other follow-ups kept out: recording refused sends in mail telemetry (`undeliverable` with a reason, today refusals return before any telemetry), and sourcing the roster from data rather than config for deployments that want it centralized.

## Tests

- `internal/mail/crosscity_test.go` — table tests for the classifier (first-slash split, exact-match city names, empty remainders, whitespace, disabled roster), the typed unknown-city error (including that it does not unwrap to the resolution error it replaces), read-side expansion, and sender qualification.
- `internal/config/mail_crosscity_test.go` — parse, roster accessor (nil-safe, explicit-city precedence, workspace/directory fallbacks), validation table (empty/duplicate/invalid/self-listing cities, rig collisions against peer and local city names), and an end-to-end `LoadWithIncludes` refusal for the rig/city collision.
- `cmd/gc/mail_crosscity_test.go` — foreign resolution against a **nil session store** (a regression that consults the store cannot pass), one-mailbox equivalence, typed unknown-city vs rig-segment vs no-roster failure shapes, full-command send/reply flows (canonical To, qualified From, `--from` matrix, typed `--notify` refusals before any write), and read-target expansion.
- `internal/api/handler_mail_crosscity_test.go` — the same behaviors through the HTTP surface, including qualified From on send and reply, the exact no-roster refusal shape, and city-qualified delivery served by an ordinary inbox query.

`make test`, `go vet`, and the full `internal/api`, `internal/mail`, `internal/config` packages pass; `docs/reference/{config.md,cli.md,schema/city-schema.*}` are regenerated via `cmd/genschema`.

### DIFF
--- a/cmd/gc/cmd_mail.go
+++ b/cmd/gc/cmd_mail.go
@@ -1042,6 +1042,20 @@ func resolveMailRecipientIdentityCached(cityPath string, cfg *config.City, sessS
 	if normalized := normalizeNamedSessionTarget(identifier); normalized == "" || normalized == "human" {
 		return "human", nil
 	}
+	roster := mailCityRosterFor(cfg, cityPath)
+	switch kind, addr := roster.ResolveCityAddress(identifier); kind {
+	case mail.CityAddressForeign:
+		// A peer city's address is canonical as written; its identity can
+		// only be checked by that city, so the session store is never
+		// consulted.
+		return addr, nil
+	case mail.CityAddressLocal:
+		// <local city>/<addr> and <addr> are one mailbox.
+		identifier = addr
+		if normalized := normalizeNamedSessionTarget(identifier); normalized == "" || normalized == "human" {
+			return "human", nil
+		}
+	}
 	if sessStore != nil {
 		sessionID, err := session.ResolveSessionIDByExactID(sessStore, identifier)
 		if err == nil {
@@ -1059,7 +1073,14 @@ func resolveMailRecipientIdentityCached(cityPath string, cfg *config.City, sessS
 	if normalizeNamedSessionTarget(identifier) == controllerMailIdentity {
 		return "", session.ErrSessionNotFound
 	}
-	return resolveMailIdentityWithConfigCached(cityPath, cfg, sessStore, identifier, cache)
+	resolved, err := resolveMailIdentityWithConfigCached(cityPath, cfg, sessStore, identifier, cache)
+	if err != nil {
+		// A slash-form recipient whose first segment names neither a rig nor
+		// a roster city refuses as unknown-city, so a stale roster is never
+		// spelled "session not found".
+		return "", mail.WrapUnresolvedRecipient(err, identifier, roster, mailCityLocalScopes(cfg))
+	}
+	return resolved, nil
 }
 
 func configuredMailboxAddress(identifier string) (string, bool) {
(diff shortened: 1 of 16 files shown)


A3Ackerman:
Converting this to an RFC issue per the contributor conventions — design direction before line review: #6171. The implementation stays on `A3Ackerman:feat/mail-crosscity-addressing` at d987472cc (flag-gated, inert when off) and this PR reopens on a yes there. #5386 is the transport half and should land first.

PR #5799, opened 2026-08-31
fix(orders): read order-tracking sweeps from the Live tier (ga-v5vnyp)

### LINKED ISSUE #5832: orders: stale/orphan order-tracking sweeps read the caching snapshot, unlike their Live-tier siblings
## Symptom
The order-tracking sweeps (`StaleOpenRuns`, `OrphanedOpenRuns`) can act on a stale snapshot: they read order-tracking beads through the `CachingStore`, while their sibling readers in the same file (`RecentRunsAll`, `OpenRuns`) deliberately read the Live tier — the file header calls cache-bypass "the duplicate-dispatch guarantee". A run whose state changed since the snapshot is missed or double-handled by the sweep.

## Root cause
`internal/orders/store_reads.go`: `StaleOpenRuns`/`OrphanedOpenRuns` call `s.store.ListByLabel(labelOrderTracking, 0, beads.WithBothTiers)`, which the `beadPolicyStore` wrapper forwards into the caching snapshot rather than the backing store. Separately, `cmd/gc/bead_policy_store.go`'s `ListByLabel` builds its query without `Sort`, while both backing implementations set `SortCreatedDesc` — so ordering on the wrapper path is store-defined and non-deterministic.

## Reproduction
Write an order-tracking bead straight to the backing store while the controller's `CachingStore` holds a snapshot; run the stale/orphan sweep; the sweep does not see the write that `OpenRuns` sees. The regression tests in the linked PR encode this.

## Environment

## Proposed fix
#5799 — read both sweeps from the Live tier like their siblings, and give the wrapper's `ListByLabel` the same explicit sort the backings use (that second change touches two other callers, called out in the PR).

A3Ackerman:
## What

`StaleOpenRuns` and `OrphanedOpenRuns` are the only two readers in `internal/orders/store_reads.go` that do not declare a Live read: they call `s.store.ListByLabel`, which the controller's `beadPolicyStore` forwards into the CachingStore's in-memory snapshot instead of the backing store. Every peer reader (`RecentRunsAll`, `OpenRuns`, `ClosedRunsForRetention`, `HasOpenWork`) already reads `beads.HandlesFor(store).Live.List`.

Both sweep readers now read Live with `TierBoth` and `SortCreatedAsc`, so the bounded close budget drains the oldest jam first. `beadPolicyStore.ListByLabel` also gains the `Sort` its two wrapped implementations (BdStore, CachingStore) already set — the wrapper silently dropped it, leaving Go map iteration order under any budgeted consumer.

The second commit moves the sweep tests' fault injection (`countFailStore`, `closeFailStore`) from `ListByLabel` onto `List`, guarded on the order-tracking label: with the reader on the Live path, an override left on `ListByLabel` is never called, so the injected faults never fire and the retry loop goes untested.

## Why

Order-tracking beads are created and closed through the dispatcher's own uncached handle, so the controller's cache never observes them. Against a stale cache, the stale-tracking watchdog reads zero rows on every tick and prints nothing, while the Live open-work gate sees the very same beads and holds the orders shut — orders sit unfireable behind tracking beads the 2-minute age-out exists to close (ga-v5vnyp: in practice this left the watchdog blind for 43 hours). Status semantics are preserved: `IncludeClosed` stays false and there is no `Status:"open"` filter, so `in_progress` tracking beads are still swept.

## Tests

- `internal/orders`: `TestRecentRunsAllOpenRunsUseLiveTier` now covers both sweep readers and pins oldest-first ordering; new `TestStaleOpenRunsKeepsInProgressAndExcludesClosed` pins the status semantics. Both fail with the fix reverted ("StaleOpenRuns issued no List query").
- `cmd/gc`: new `TestOrderTrackingSweepWatchdogSeesBeadsWrittenBehindTheCache` and `TestOrphanedOrderTrackingSweepSeesBeadsWrittenBehindTheCache` build the production stack (`beadPolicyStore` over a primed `CachingStore`) and write the stale bead to the backing store behind the cache — the pre-existing watchdog tests all used a bare MemStore, which is why the suite stayed green while the production read path was blind.
- `cmd/gc`: the four `TestSweepOrphanedOrderTracking_Retry*` tests are re-pinned to the production read path; with `OrphanedOpenRuns` reverted to raw `ListByLabel` they fail again.
- `go build ./...`, `go vet`, `go test ./internal/orders -count=1`, and the full order-tracking sweep/watchdog family in `cmd/gc` pass.

Merge note: the first commit alone leaves the four retry tests red (the second repairs their fault injection); squash-merging keeps main green at every landed commit.

Fixes #5832.

### DIFF
--- a/cmd/gc/bead_policy_store.go
+++ b/cmd/gc/bead_policy_store.go
@@ -182,7 +182,11 @@ func (s *beadPolicyStore) ListByLabel(label string, limit int, opts ...beads.Que
 		Label:         label,
 		Limit:         limit,
 		IncludeClosed: beads.HasOpt(opts, beads.IncludeClosed),
-		TierMode:      policyTierFromOpts(opts),
+		// Both wrapped implementations (BdStore, CachingStore) set a Sort here;
+		// omitting it left SortDefault, which over a map is non-deterministic
+		// order — so any budgeted consumer got an arbitrary slice (ga-v5vnyp).
+		Sort:     beads.SortCreatedDesc,
+		TierMode: policyTierFromOpts(opts),
 	})
 }
 
--- a/cmd/gc/city_runtime_test.go
+++ b/cmd/gc/city_runtime_test.go
@@ -1925,6 +1925,100 @@ func TestOrderTrackingSweepWatchdogClosesAllStaleTracking(t *testing.T) {
 	}
 }
 
+// TestOrderTrackingSweepWatchdogSeesBeadsWrittenBehindTheCache is the
+// regression test for ga-v5vnyp, and it is deliberately built on the PRODUCTION
+// store stack rather than a bare MemStore.
+//
+// That distinction is the whole point. Every other watchdog test here hands
+// CityRuntime a raw beads.NewMemStore(), so every read is authoritative and the
+// suite stayed green for the entire 43 hours the watchdog was blind in
+// production. The real controller store is a beadPolicyStore wrapping a
+// CachingStore, and order-tracking beads are created and closed through the
+// DISPATCHER's own uncached handle — so the cache never observes them. A
+// non-Live read therefore returns zero rows forever, the watchdog prints
+// nothing (it only speaks when it closes >0), and two orders sat gated for 51
+// minutes behind a 2-minute age-out that was running the whole time.
+//
+// The test reproduces exactly that: prime the cache while the store is EMPTY,
+// then write the stale tracking bead straight to the backing. Revert either
+// sweep reader to s.store.ListByLabel and this fails.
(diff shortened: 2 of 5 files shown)


A3Ackerman:
Disclosure a reviewer would otherwise have to find: the `beadPolicyStore.ListByLabel` Sort fix in this PR is shared by two other production callers — `cmd/gc/cmd_bd_store_bridge.go:255` (the generic label-list CLI bridge) and `internal/session/info_store.go:193` (session listing). Both backing stores already sort `SortCreatedDesc` on their own `ListByLabel`, so this only removes non-determinism on the wrapper path — but it is a change to a shared method riding on an "orders" fix. Happy to split it into its own commit with its own test if you'd prefer.

A3Ackerman:

PR #5800, opened 2026-08-31
fix(dispatch): never serve steps of a partially-instantiated workflow

### LINKED ISSUE #5833: dispatch: steps of a partially-instantiated workflow are served after markFailed drops the instantiating fence
## Symptom
When molecule instantiation fails partway, the already-created steps become claimable: `gc hook`, the claim path and control-ready dispatch all serve them, so agents pick up steps of a workflow that can never complete, and the failed root is never the thing that gets handled.

## Root cause
`molecule.markFailed` (`internal/molecule/molecule.go`) sets `molecule_failed=true` and clears the `instantiating` fence in the same `SetMetadataBatch` — atomic by design. But `filterUnreadyHookCandidates` (`cmd/gc/cmd_hook.go`) and `mergeControlReadyGroups` (`cmd/gc/dispatch_control_ready.go`) filter only on the instantiating key and never on `molecule_failed`, even though `beadmeta`'s own doc comment for `MoleculeFailedMetadataKey` says dispatch/sling/`cmd/gc` read it to skip failed roots. The write that is meant to hide the orphaned steps is the write that un-hides them.

## Reproduction
Instantiate a molecule whose second step fails to create; observe the first step returned by `gc hook` and by the control-ready merge. Encoded in the linked PR's tests.

## Environment

## Proposed fix
#5800 — treat `molecule_failed` as unready at the shared chokepoint. A follow-up doctor check for stale failed roots with no closing cook is the natural companion; it is deliberately not bundled.

A3Ackerman:
## What

When workflow instantiation aborts partway, `markFailed` stamps every bead it already wrote with `molecule_failed=true` and clears their `gc.instantiating` fence. The readiness scans drop fenced beads, but nothing in dispatch looked at `molecule_failed` — so the failure path took steps that were correctly hidden mid-pour and made them visible, routed, and claimable. Such a step cannot progress: it fails on its missing root, closes fail, and mails an escalation. N steps in, N escalations out, zero forward progress (ga-033u0e).

This adds the `molecule_failed` filter to every surface that hands work to an agent:

- `filterUnreadyHookCandidates` (`cmd_hook.go`) — the chokepoint all hook paths normalize through: plain `gc hook`, the claim path, and cross-store selection
- `mergeControlReadyGroups` (`dispatch_control_ready.go`) — the control dispatcher's workflow-serve scan, which does not go through the hook filter
- the jq fallback in `dispatch_runtime.go` — kept in lockstep with the Go filter

## Why

`molecule_failed` was honored only where it does not gate dispatch (attempt-lineage selection and the graph.v2 recovery paths, which use it to CLOSE a failed subtree on the next cook). Until that cook, the steps sat claimable. Safe by construction: recovery closes failed subtrees rather than re-dispatching them, so hiding these beads removes a claim window and cannot remove work anyone would have completed.

## Tests

- `TestDoHookFiltersFailedPartialMoleculeBeads` — seeds the exact end state `markFailed` leaves (`molecule_failed=true` AND cleared fence), so it fails if hiding ever keys on the empty fence instead of the failed mark
- `TestMergeControlReadyGroupsSkipsFailedPartialMolecules`
- `TestControlReadyShellFilterMatchesGoFilterOnFailedPartialMolecules` — asserts the shell query names both metadata keys, so the Go and jq surfaces cannot drift apart

All three verified to fail with the filters removed. Full `cmd/gc` dispatch/hook filter suite passes; `go build ./...` and `go vet` clean.

## Second pass (review findings)

The serve-side hide alone would have been a regression: a failed v1 pour leaves `gc.routed_to` on unassigned steps, so the demand side kept counting them as pool demand — a seat spawns, its hook hides the step, the seat drains, the row is counted again next tick. Both demand paths now exclude them (`demandRowServable`, and the reconciler's count-form via `poolDemandCountJQ`), and the four spellings of the predicate are collapsed onto one exact-match helper, `beadmeta.MoleculeFailed` (`== "true"`, the only value `markFailed` writes and the only form the jq twins can compare), used by the hook filter, the control-ready merge, the demand predicate and `dispatch.isFailedPartialMolecule`. Convergence no longer re-adopts a failed root either: `FindByIdempotencyKey` and its scan fallback pass over a marked root (`liveWispWithKey`), so the handler goes to `waiting_manual` and the reconciler pours fresh instead of re-activating every child of a workflow that cannot progress. The jq branch is now executed for real — `TestControlReadyShellReduceDropsFailedPartialMolecules` slices the shipped program out of the query and runs it under `jq -s` over the Go test's groups, requiring identical ids — alongside an agreement-corpus row, a fake-`bd` count-form fixture and a real-`bd` process-tier case. All five new tests were red before the source edits. The `molecule_failed` doc comment in `internal/beadmeta/keys.go` now lists every serve, count and re-activate surface. The twelve `internal/config/testdata/workquery/*_PoolDemand_*.golden` files changed by two lines each because the count-form's jq program changed.

## Release note

v1 pours and convergence wisps have no closer for a pour that aborts partway (graph.v2 recovery closes its own failed subtrees on the next cook; `markFailed` on the v1 path marks and un-fences but does not close). Their `molecule_failed` beads therefore now sit open and hidden: no hook serves them, no pool counts them, and convergence no longer re-activates them. Before this change each such step was served, failed on its missing root, and mailed an escalation. A doctor check that closes or reports these orphaned beads is a separate follow-up and out of scope here.

Fixes #5833.


A3Ackerman:
One risk worth naming against my own change: hiding `molecule_failed` steps from every ready queue trades a loud failure (repeated escalation) for a silent one if no follow-up cook ever closes the failed subtree — e.g. a one-off instantiation outside the order/recipe flow. Recovery closes failed subtrees on the next graph.v2 cook, but nothing in this PR guarantees a cook happens for every caller of `molecule.Instantiate`. I'd propose a follow-up doctor check for stale `molecule_failed` roots with no closing cook rather than blocking the fix on it; I'll file it after this lands unless you'd rather see it here.

A3Ackerman:
Green after the second pass: whole matrix passing (CI / integration=pass CI / required=pass Evidence / critical-path suites=pass Integration / packages-core-3-of-4=pass ). The red in between was the resource-census ratchet catching two new subprocess call sites in the added tests; they now go through the package's existing external-command seam and an already-gated test, so the census is unchanged.

A3Ackerman:
Parked as a draft to stay under our six-open-PR cap (conflicting with main since 09-02; will rebase and re-ready when a slot opens). Will re-ready when a slot opens; no changes to the branch.

PR #5801, opened 2026-08-31
fix: accept safe undeclared push-guard successors

quad341:
## What this changes

The pre-push ownership guard now accepts a single, still-owned in-progress task as the successor to a closed task encoded in a reused branch name. This prevents legitimate rotated audit work from being blocked while preserving fail-closed checks.

The candidate must be the session’s only in-progress match and pass a fresh ownership, route, and hold check; ambiguous or unhealthy candidates still block.

## Review notes

- Shared ownership predicate reused by the normal guard and fallback.
- No changes to deploy-branch handling, ordinary multi-match behavior, or routing-mismatch handling.

## Test plan

- [x] Run the 40-case push-ownership regression suite.
- [x] Run the full local CI-equivalent suite with tracked non-diff failure attribution.
- [x] Run shell syntax, shellcheck, build, vet, and CI policy checks.
- [x] Release gate: [release-gates/ga-2m6tqt-push-ownership-undeclared-successor-gate.md](release-gates/ga-2m6tqt-push-ownership-undeclared-successor-gate.md)

### DIFF
--- a/release-gates/ga-2m6tqt-push-ownership-undeclared-successor-gate.md
+++ b/release-gates/ga-2m6tqt-push-ownership-undeclared-successor-gate.md
@@ -1,0 +1,46 @@
+# Release gate: push-ownership undeclared successor (`ga-2m6tqt`)
+
+- Gate result: **PASS**
+- Reviewed commit: `3c77b6789bd9621bd019cdd07cc72093cf72cb98`
+- Base ref: `origin/main@a85f857b3987bd18593cea2e9594a17a82b10df1`
+- Deploy mode: remote
+- Evaluation date: 2026-08-31
+- Full-suite logs: `/var/tmp/gc-local-tests.dLtqwB`
+
+`docs/PROJECT_MANIFEST.md` is not present at the evaluated ref. This gate uses
+the canonical criteria in `mol-deployer-gate.formula.toml`, the current
+deployer prompt, and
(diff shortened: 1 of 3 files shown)


quad341:
@quad341 — the test rewrite is the part worth calling out. Noticing that the old single-canned-response fake answered *every* `bd show` with the same closed payload, so the existing test literally couldn't tell "checked the candidate and rejected it" from "never looked at it," then replacing it with an id-aware fake plus `show-ids.log` assertions on *which* beads were queried — that's the kind of hole that usually surfaces six months later as a bug. The multi-candidate case pinning that neither candidate is ever touched is the right instinct too. Extracting `_pog_ownership_violation` is clean, and carrying the zsh `status`-parameter note along with the code it protects is a nice touch.

CI path-filtered the integration matrix out, so I ran the suite here: 40/40, shellcheck and `bash -n`/`zsh -n` clean.

I'll take four comment fixes on my side — the resolution-order contract, the ga-bf39j8 "declared-link required" paragraph, and the retained invariant above the rewritten test all still describe the pre-tier world.

One follow-up bead, not a revision: with the declared link gone, a mayor **close** ruling stops blocking when the session's one remaining claim is healthy. The other three ruling shapes are unaffected. I'll flag it for a quick call on whether that's the intended trade before merging.


---


Review coverage:
- completed: `qwen claude`
- unavailable: `codex`

PR #5802, opened 2026-08-31
fix(workflow): refuse a workflow delete that would strand live steps

### LINKED ISSUE #5834: workflow: retention deletes a workflow root without checking for live descendant steps, stranding them
## Symptom
Unattended retention deletes a workflow root while its steps are still open, leaving rootless, unworkable steps. `order_tracking` retention runs by default (`delete_after_close = "7d"`), so a clean deployment hits this whenever a workflow outlives its tracking bead's retention window.

## Root cause
`deleteWorkflowBead` (`cmd/gc/cmd_convoy_dispatch.go`) unwinds dependency edges and deletes with no check for live descendants; its callers — `sweepClosedOrderTrackingRetention` (`cmd/gc/order_dispatch.go`) and the wisp GC (`cmd/gc/wisp_gc.go`) — filter only on age/retain-count or on the candidate's own root being terminal, never on the candidate's children.

## Reproduction
Create an order-tracking root with an open step, age the root past the retention window, run the sweep; the step survives with no root. Encoded in the linked PR's tests.

## Environment

## Proposed fix
#5802 — refuse a delete that would strand live steps, reusing the existing `storeHasOpenDescendants`, with a whole-closure carve-out (deleting an entire set strands nothing). The batch path's index-only backstop still misses ParentID-only partial stamps; that residual gap is named in the PR rather than hidden.

A3Ackerman:
## What

`deleteWorkflowBead` unwound dependency edges and deleted the bead without looking at the workflow steps hanging off it via `gc.root_bead_id`. Order-tracking retention calls it on expired CLOSED tracking roots, and a closed root can still own OPEN steps — deleting one leaves those steps rootless, and a rootless step is unworkable by construction: root-metadata reads fail, closes fail, and each failure mails an escalation, so the workflow becomes an escalation generator with zero forward progress (ga-ejwo1q).

The delete layer now proves a delete strands nothing before it runs, and the guard is **set-aware** — a delete that removes a whole workflow strands nothing, so an open descendant inside the same delete set does not block it:

- `deleteWorkflowBead(id)` — guard against an empty set
- `deleteWorkflowBeads(ids)` — per-id guard, set = ids
- `deleteWorkflowBeadsBatch(ids)` — one set-level guard, then the batch delete

## Why guard the shared layer

`deleteWorkflowBead` has several callers (retention ×2, the wisp orphan reaper, the wisp closure purge, `gc workflow delete-source --apply --delete`) and any of them can be handed a root; fixing the one pruner observed to trip it would leave the class open.

The predicate is `storeHasOpenDescendants` — the authoritative view the single-flight dispatch gate already uses: membership index first, tree walk as fallback so partial-stamp molecules linked only by `ParentID` are seen. It fails **closed**: an unreadable descendant view propagates and the bead survives, mirroring `reapOrphanedClosedWisps` skipping on any unreadable root Get. A bead that lingers one more sweep is recoverable; a stranded workflow is not.

The batched path uses a membership-index-only backstop instead of the walk: walking once per closure member is O(n·depth) and would undo the batching that path exists for, and its callers already collect the closure with the same ownership rules the walk applies.

Scheduled pruners treat a refusal as a **skip**, not a sweep failure — the candidate becomes eligible again once its steps finish — and retention logs a count of what it retained, because a silent skip in a background sweep is indistinguishable from "nothing was eligible". Closed descendants deliberately do NOT block: they are terminal residue the wisp orphan reaper already collects, and blocking on them would leak closed tracking rows forever, which is what retention exists to prevent.

`gc workflow delete --force --delete` is not on that list on purpose: it does not go through `deleteWorkflowBead` — it shells out to `bd delete <ids> --cascade --force` (`deleteWorkflowMatches`) — so this guard does not cover it and it is left alone here.

## Test plan

- Five new memstore-level tests covering the per-bead refusal, whole-closure deletes staying allowed, the batch set-level refusal, and both retention sweeps retaining roots that still own open steps. Each was verified to FAIL with the guard stubbed to return nil.
- They assert the root/step PAIR rather than the root's survival alone (the defect is the step outliving its root), and a sibling root whose only step is CLOSED still prunes in the same sweep, so a guard that never deleted anything would fail them.
- `go build ./...`, `go vet ./cmd/gc/` clean; the surrounding workflow-delete/retention/wisp-reap family (52 tests) passes.

## Release note

`gc workflow delete-source --apply --delete` now refuses to delete a matched workflow root that still owns open steps outside the match set: it prints `store=<label> delete_error=workflow bead still owns live descendants: <root> (open descendants: <step-ids>)` per root and reports `result=incomplete … metadata_cleared=false` (exit 1) instead of `result=cleaned`. Order-tracking retention and wisp GC now retain such roots instead of stranding their steps — retention logs `order-tracking retention: retained N expired closed root(s) that still own open steps …: <root-ids>`; wisp GC skips them without charging its batch caps — and prune them once the steps reach a terminal state.

## Tests, second pass

The refusal is an uncharged skip in both wisp-GC paths (`TestWispGC_ClosurePurgeSkipsRefused…`, `TestWispGC_ReapSkipsOrphanOwning…`), the List-error fail-closed path and the transient-notification carve-out are driven with `failingListStore`, retention's retained roots are asserted by id, and the refusal names the open step ids. All were red before the source edits.

Fixes #5834.

### DIFF
--- a/cmd/gc/cmd_convoy_dispatch.go
+++ b/cmd/gc/cmd_convoy_dispatch.go
@@ -1920,7 +1920,7 @@ func workflowDeleteStoreLabel(cfg *config.City, cityPath, scopePath string) stri
 	return scopePath
 }
 
-// ErrWorkflowDeleteLiveDescendants marks a REFUSED workflow delete: the bead
+// errWorkflowDeleteLiveDescendants marks a REFUSED workflow delete: the bead
 // still owns open descendants that are not themselves part of this delete, so
 // removing it would strand live, claimable work.
 //
@@ -1928,10 +1928,20 @@ func workflowDeleteStoreLabel(cfg *config.City, cityPath, scopePath string) stri
 // metadata fails with an out-of-vocabulary gc.failure_class, closes fail, and
 // mails an escalation, so the workflow becomes an escalation-mail generator
 // with zero forward progress (ga-033u0e). Scheduled pruners (order-tracking
-// retention, wisp GC) must treat this as a SKIP rather than a failure: the
-// candidate becomes eligible again on a later sweep once its descendants
-// reach a terminal state.
-var ErrWorkflowDeleteLiveDescendants = errors.New("workflow bead still owns live descendants")
+// retention, wisp GC) must treat this as a SKIP rather than a failure — not
+// joined into the sweep error, not charged to a batch cap: the candidate
+// becomes eligible again on a later sweep once its descendants reach a
+// terminal state, and a refusal charged to a cap on every sweep would starve
+// the collectible candidates listed behind it.
+var errWorkflowDeleteLiveDescendants = errors.New("workflow bead still owns live descendants")
+
+// workflowDeleteRefusal is the errWorkflowDeleteLiveDescendants error for id,
+// naming the open descendants that hold it. gc workflow delete-source prints
+// it verbatim on its delete_error= line, so it must let the operator find the
+// steps, not just the root.
+func workflowDeleteRefusal(id string, open []string) error {
+	return fmt.Errorf("%w: %s (open descendants: %s)", errWorkflowDeleteLiveDescendants, id, strings.Join(open, ","))
+}
 
 // workflowDeleteSkip builds the storeHasOpenDescendants skip predicate for a
 // delete of alsoDeleting. An open descendant does not block the delete when it
@@ -1952,7 +1962,7 @@ func workflowDeleteSkip(alsoDeleting map[string]struct{}) func(beads.Bead) bool 
 
 // assertWorkflowDeleteLeavesNothingStranded refuses a delete of id when it
 // still owns open descendants outside alsoDeleting. It uses the authoritative
-// storeHasOpenDescendants view (membership index, falling back to the tree
+// storeOpenDescendantIDs view (membership index, falling back to the tree
 // walk) so partial-stamp molecules — steps linked only by ParentID, with no
 // gc.root_bead_id — are seen too.
 //
@@ -1962,12 +1972,12 @@ func workflowDeleteSkip(alsoDeleting map[string]struct{}) func(beads.Bead) bool 
 // rather than guessing. A bead that lingers one more sweep is recoverable; a
 // stranded workflow is not.
 func assertWorkflowDeleteLeavesNothingStranded(store beads.Store, id string, alsoDeleting map[string]struct{}) error {
-	open, err := storeHasOpenDescendants(store, id, workflowDeleteSkip(alsoDeleting))
+	open, err := storeOpenDescendantIDs(store, id, workflowDeleteSkip(alsoDeleting))
 	if err != nil {
 		return fmt.Errorf("checking live descendants of %s: %w", id, err)
 	}
-	if open {
-		return fmt.Errorf("%w: %s", ErrWorkflowDeleteLiveDescendants, id)
+	if len(open) > 0 {
+		return workflowDeleteRefusal(id, open)
 	}
 	return nil
 }
@@ -2001,11 +2011,15 @@ func assertWorkflowDeleteSetLeavesNothingStranded(store beads.Store, ids []strin
 		if err != nil {
 			return fmt.Errorf("checking live descendants of %s: %w", id, err)
 		}
+		var open []string
 		for _, b := range members {
 			if b.ID == id || b.Status == "closed" || skip(b) {
 				continue
 			}
-			return fmt.Errorf("%w: %s (open descendant %s)", ErrWorkflowDeleteLiveDescendants, id, b.ID)
+			open = append(open, b.ID)
+		}
+		if len(open) > 0 {
+			return workflowDeleteRefusal(id, open)
 		}
 	}
 	return nil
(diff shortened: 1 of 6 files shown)


A3Ackerman:
Disclosure: the batch guard (`assertWorkflowDeleteSetLeavesNothingStranded`, used by `gc workflow delete-source --delete` and the wisp-GC closure purge) is a membership-index backstop, not the walk — so a partial-stamp molecule linked only by ParentID can still be stranded through the batch path, which is exactly the less-supervised one. The single-bead guard's walk fallback closes that case; the batch path trades it for O(n) instead of O(n·depth). That is a named gap, not a footnote: I'll file it as a follow-up unless you'd rather the batch path take the walk here.

A3Ackerman:
Green after the second pass: whole matrix passing (Evidence / critical-path suites=pass CI / integration=pass CI / required=pass ). The only pushes since the last green were the review-driven changes described in the body.

jacobhausler:
Sequencing note from a downstream city — this PR does **not** collide with #6129, and #6129's enumeration would make this guard's own answer consistent.

**Why we are not racing this function.** #6129 is about the *member enumeration* (`findWorkflowBeads` / `findWorkflowBeadsFromRoot`), not the stranding check. This PR adds the check and does not adopt `beads.DirectMembers` — verified against the diff (`cmd/gc/cmd_convoy_dispatch.go`, `cmd/gc/cmd_convoy_dispatch_test.go`, `cmd/gc/order_dispatch.go`), so filing that separately wasn't duplicating work in flight.

**The one place the two do meet, and it is worth a look here.** This PR's new skip predicate asks the membership question across both tiers:

```go
Metadata: map[string]string{beadmeta.RootBeadIDMetadataKey: id},
TierMode: beads.TierBoth,
```

and `storeHasOpenDescendants` (`cmd/gc/order_dispatch.go:2470-2476`) already reads `beads.HandlesFor(store).Live` with `TierMode: beads.TierBoth`. The delete set this guard then gates, however, is computed by `findWorkflowBeads`, whose member query (`cmd/gc/cmd_convoy_dispatch.go:2787-2790`, and `:2805-2808` on the delete-source arm) leaves `TierMode` at its zero value — `TierIssues` (`internal/beads/query.go:40-41`).

On a split city that asymmetry bites, because the sweep is federated: `federateSweepViews` (`:1492`) appends the class binding, and `convoyStoreViewsWithBinding` (`cmd/gc/cmd_convoy.go:567-613`) appends it **constructor-opened** — the work-leg openers go through `wrapStoreWithBeadPolicies`, which rewrites a `TierIssues` read to `TierBoth`, and the binding has no such layer. `internal/beads/query.go:50-76` states the consequence and records the measured live case (`bd ready --include-ephemeral`: 17 rows over the relocated graph DB, 14 from the federated reader, and the 3 missing were wisps — "ephemeral wisps are how orchestration steps run").

So after this PR, on a relocated city `gc workflow delete` can ask "are there live steps?" across both tiers and "which beads do I delete?" across one. That is a guard refusing on a set the sweep cannot enumerate, and — with a whole-closure carve-out — a closure judged complete off a member set that is missing its wisps.

**Offer, not a demand.** If it is convenient, the enumeration follow-up is small enough to fold in here rather than land after: swap the member legs for `beads.DirectMembers(store, rootID)` (the contract this PR's own predicate follows), and state `beads.FederatedReadTier` on the leftover `gc.workflow_id` root-discovery leg at `:2774-2779`. If you would rather keep this PR to the stranding check, that is a perfectly good call — take #6129 as the follow-up and this comment as a note that the two touch the same file, so whichever lands second rebases over one function.

Nothing here blocks #5802; filed from `main` `15819b315`.

A3Ackerman:
Thanks — agreed on both counts: the two PRs touch the same file and not the same function, and the tier asymmetry you describe is real on a relocated city. I'd rather keep this one to the stranding check.

As it stands the guard's index reads are `TierBoth` on both legs, so an open wisp the collector cannot see makes the guard refuse (the candidate lingers one sweep) rather than delete around it — the safe direction — and closed wisps the collector misses are the enumeration gap #6129 already names, not something this guard changes.

So: #6129 as the follow-up that makes the member legs consistent; whichever lands second rebases over `findWorkflowBeads`, and I'll do that here if yours goes first.

quad341 (approved):
@A3Ackerman — this is the good version of a bug fix. You found the defect in one pruner and then asked the right next question: who else calls this? Putting the guard at `deleteWorkflowBead` fixed the other four callers for free, and naming `deleteWorkflowBeadUnguarded` that plainly means the next person adding a caller can't skip the check by accident.

Two things I especially liked. A refusal not charging the batch cap is the kind of thing that only bites in production six weeks later — you got it right in both wisp-GC paths and pinned it with `cap=1` and the refused entry listed first, so the test fails loudly if anyone "simplifies" it. And asserting the root/step *pair* rather than just the root surviving is the difference between testing the fix and testing the defect.

One thing your writeup undersells: `deleteWorkflowBeadsBatch` has exactly one caller, the wisp-GC closure purge. `delete-source` goes through the per-id guard, walk fallback and all — so the `ParentID`-only gap never reaches the operator command.

Merging as-is. I'll file the residual batch-path race (a descendant created *after* closure collection) as a follow-up alongside #6129 — not yours to carry. Thanks for the writeup; it made this reviewable.


---

PR #4790, opened 2026-07-29
feat(order): bound gc order history with --limit/--since (+ order-firing doctor check)

brandonmartin:
## Problem

`gc order history` had no bound of any kind — no `--limit`, no `--since`. It
fetched every retained order-run bead on every invocation, which measured ~22s
when this was first noticed and ~52s by the time it was fixed, on a city whose
order-run history grows continuously. That makes the command unusable at exactly
the moment an operator reaches for it — and it is the command the `order-firing`
doctor check tells you to run. The `order-firing` doctor check had the same
unbounded-read problem.

## Change

**`gc order history`** now takes:

- `--limit` (default 50; `0` restores the full read)
- `--since` (a duration, matching `gc events --since`)

The row bound reaches the backing store: `RecentRuns` opts into
`AllowBackingCreatedLimit`, because this read projects a newest-first listing
whose bound is over `created_at` — the same column the sort key uses — so a
bounded backing read returns the same prefix the client-side cut would. Fetching
the whole corpus and trimming afterwards was most of the cost. The sibling
`Cursor` read deliberately does **not** opt in (its max-seq reduction is over a
different column than the sort key); that distinction is documented at both
sites so the next reader does not "fix" the wrong one.

In multi-order mode each order is fetched bounded and the merged result is cut
again after the newest-first sort, so the rows kept are the newest overall
rather than the newest of whichever order was iterated first. `--since` has no
wire equivalent on the API path (the server's `before` is an upper bound), so
the time window is applied to the response there instead of being silently
dropped. A malformed `--since` is rejected at the CLI edge with an explicit
message.

**`doctor order-firing`** bounds its event reads and parallelizes its per-order
run lookups (the same unbounded-read class).

## Measured (live, large-history city)

| invocation | before | after |
|---|---|---|
| `order history` (all orders), unbounded | 52.3s / 12,402 lines | — |
| `order history`, default `--limit 50` | — | 44.1s / 51 lines |
| `order history <one-order>`, unbounded | 8.0s / 1,877 lines | — |
| `order history <one-order> --limit 20` | — | 5.3s / 21 lines |

Includes tests for the CLI bounds, the bounded backing reads, and the doctor
check.

PR #5803, opened 2026-08-31
fix(order): gc order history must render outcome, not just cadence

A3Ackerman:
## Problem

`gc order history` printed ORDER/BEAD/EXECUTED and dropped the run outcome, so a run that FAILED and a run that SUCCEEDED rendered byte-identically. During the ga-7unsv0 hub-sync outage, 18 of 21 consecutive runs carried the exec-failed label and the operator could not see one of them; the outage read as healthy for ~5 hours while the data plane had recorded every failure correctly.

The defect is one dropped field: the command builds its rows from `orders.OrderRun` values that already carry a typed `Outcome` (populated by `decodeRun`, with `Display()` alongside) and never reads it.

## Change

Adds an OUTCOME column to the table and an `outcome` field to `--json`, rendering three distinct states — collapsing them is the actual hazard:

- `success | failed | canceled` — dispatcher stamped an outcome
- `running` — still open
- `unknown` — closed with nothing stamped

The third is the dangerous one: a run killed before the dispatcher could stamp anything is closed by the order-tracking retention sweep with no outcome label, previously indistinguishable from a clean success. It must never render as success and must never render blank.

The API-sourced route cannot report an outcome at all (`api.OrderHistoryView` / `genclient.OrderHistoryEntry` have no such field); it prints `-` rather than guessing. Adding the field to the server schema plus client codegen is a follow-up (part B on ga-kahru9). Both routes currently fall back to the store path (verified `GC_DEBUG=1 route=fallback`), so the store path is the one that serves.

## Tests

- `TestOrderHistoryRendersOutcomeForAllTrackingStates` — all four render states, table and `--json`, with per-row non-vacuity assertions (a silently missing row fails the test).
- `TestOrderRunOutcomeDisplayNeverBlankForClosedRun` — closed-with-no-label renders `unknown`, never `""`.
- Full order-history family (21 tests incl. the #4790 `--limit`/`--since` bounds) green; `go build ./...` and `go vet` clean.
- Mutation-verified: making a killed run read as `success` fails the suite in both table and JSON paths.

### DIFF
--- a/cmd/gc/cmd_order.go
+++ b/cmd/gc/cmd_order.go
@@ -1533,18 +1533,18 @@ func renderOrderHistoryFromAPI(cr api.CachedRead[[]api.OrderHistoryView], name, 
 	}
 
 	if hasRig {
-		fmt.Fprintf(stdout, "%-20s %-15s %-15s %s\n", "ORDER", "RIG", "BEAD", "EXECUTED") //nolint:errcheck
+		fmt.Fprintf(stdout, "%-20s %-15s %-15s %-21s %s\n", "ORDER", "RIG", "BEAD", "EXECUTED", "OUTCOME") //nolint:errcheck
 		for _, e := range entries {
 			rig := e.Rig
 			if rig == "" {
 				rig = "-"
 			}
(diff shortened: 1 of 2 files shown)


A3Ackerman:

PR #4063, opened 2026-07-08
feat(patches): add rig field and wildcard `rig = "*"` support to [[patches.agent]]

sarendipitee:
Add `rig` field as replacement for legacy `dir` on `[[patches.agent]]`.

**Targeting modes:**
- `rig = "*"` — wildcard: match all agents named `Name` across all rigs + city
- `rig = "rigA"` — specific rig (equivalent to `dir = "rigA"`)
- `dir = "rigA"` — legacy, unchanged
- Neither set — city-scoped, unchanged

**Key changes:**
- `AgentPatch.Rig` field with deprecation comment on `Dir`
- `agentPatchTargetDir` helper resolves effective target (Rig > Dir)
- Wildcard branch in `applyAgentPatch` uses `BindingQualifiedName` for pack agents (Name=witness, BindingName=gastown → found by `gastown.witness`)
- `nameMatchesAnyImplicit` / `namesForImplicitAgents` helpers for deferred wildcard patches
- Wildcard-aware deferred patch partition in compose.go
- Field sync tests updated for Rig targeting key
- Rig + Dir conflict validation
- Unit tests for wildcard, rig, legacy dir, city-scoped, inject_fragments presence-aware

PR #2851, opened 2026-05-31
fix(config): tolerate named_session with a missing backing template

steveyegge:
## Problem

A `named_session` whose backing agent template does not resolve after pack expansion was a **hard error** in `ValidateNamedSessions`, which fails the entire config load. Because nearly every `gc` command loads the city config, one typo'd or transiently-unresolvable `named_session` template **bricked all of them**.

Concretely, `gc session attach mayor` failed with:

```
gc session attach: named_session "yegge-ai/rizato": referenced template not found after pack expansion
```

…solely because an *unrelated* `rizato` named session referenced a template that didn't resolve. Attaching to `mayor` has nothing to do with `rizato`, but the strict config-load gate took everything down with it. `gc agent list`, `gc session attach <anything>`, etc. all failed the same way.

## Fix

Demote this one case to a **non-fatal warning**:

- Skip the broken named session during validation (it holds no name reservation or always-mode capacity slot, so the rest of the bookkeeping stays correct).
- Surface it on stderr via the existing load-warning channel:
  ```
  named_session "yegge-ai/rizato": backing template "yegge-ai/gastown.ghost" not found after pack expansion; named session disabled until its template resolves
  ```
- Unrelated sessions and every other command keep working.
- A command that **actually targets** the broken session still fails clearly, at materialization time (`ensureSessionIDForTemplate` → `FindAgent`).

Genuine structural problems remain **fatal**: duplicate identity, invalid scope/mode, name/alias collisions, `always`-mode capacity overflow, and `sleep_after_idle` conflicts.

### Daemon safety

The controller degrades gracefully too: `buildDesiredState` already skips named sessions whose spec doesn't resolve (`findNamedSessionSpec` → `ok=false`), so a broken `mode="always"` session is never materialized and cannot crash-loop the reconciler. No nil-deref paths were introduced (broken sessions never reach agent-field access).

## API change

`ValidateNamedSessions(cfg) error` → `ValidateNamedSessions(cfg) ([]string, error)` (returns non-fatal warnings). The single production caller (`LoadWithIncludesOptions`) folds the warnings into `Provenance.Warnings`; `cmd/gc` classifies them as non-fatal/printable via the new `config.IsDisabledNamedSessionWarning`.

## Tests

- `internal/config`: `TestValidateNamedSessions_MissingTemplateIsWarningNotFatal` (load succeeds, exactly one warning naming the broken session) and `TestValidateNamedSessions_StructuralErrorsStayFatal` (duplicate identity still fatal).
[...]

PR #5806, opened 2026-08-31
fix(config): an unappliable patch is skipped and reported, not fatal

A3Ackerman:
## Summary

**What.** A city-level `[[patches.*]]` entry whose target cannot be resolved no longer aborts config load. `ApplyPatches` now attempts every patch and records each failure on `LoadWarnings` as `patches.<kind>[<i>] was NOT APPLIED: <reason>`; composition re-drains those warnings into provenance after patch application. The CLI classifies them as non-fatal and prints them on every command, and `gc config --validate` promotes them to validation errors and still exits non-zero. Pack-level patches (applied inside recursive pack loading) are unchanged and still fail the load — a pack whose own patch dangles is one internally broken file, with no cross-file edit window.

**Why.** Every `gc` command loads config, so one dangling patch reference halted an entire city — including the commands you'd use to diagnose it. A dangling reference doesn't require a mistake: `city.toml` and a local-path pack are two files, and the pack is live the instant it's written, so any agent stand-up or rename opens a window where a `[[patches.agent]]` exists without its target. We hit this operating a multi-agent fleet: an unrelated command died with `patches.agent[12]: agent "..." not found in merged config` while a new seat was mid-stand-up. One seat's edit window stopping every other agent is a worse failure than one patch left unapplied.

**Why skipping does not weaken the contract.**

- A skipped patch changes nothing. Every error path in all five apply functions fires before any field is applied (the wildcard path errors only when zero agents matched), so the loaded config is exactly the config without that patch — never a half-applied one.
- The skip cannot go silent. The warning is emitted on every command (a condition of this invocation, unlike standing deprecation advisories, which stay filtered), and `gc config --validate` fails on it — any CI job or deploy gate that validates config still goes red on a typo.
- It matches how the loader already treats this defect class: a named session whose backing template doesn't resolve after pack expansion is disabled with a warning, not fatal.
- The residual risk is real and worth naming: a patch that used to match can drift out of effect (e.g. `suspended = true` or an env override after its target is renamed) and the city keeps running without it. But the fatal behavior never enforced that patch's intent either — the target still wasn't patched; failing the load just added a city-wide outage on top, and reported only the first defect per run. Now every unappliable patch is reported in one load, and enforcement lives on the surface built for it.

Follow-up commits: adapt the two `TestApplyPatches` cases from #4063 (wildcard no-match, dir+rig conflict) to assert the same guarantees against the recorded warning, and update the pack-spec sentence that stated the single fatal rule.

**Behavior note:** `LoadWithIncludes` no longer returns an error for an unappliable city-level patch; anything depending on that must check provenance warnings or `gc config --validate`. The eight tests that encoded the fatal contract were rewritten, not deleted — each still requires the typo to be reported naming target and reason, plus that the skipped patch modified nothing.

## Testing

- `go build ./...`; `go vet ./cmd/gc/... ./internal/config/...` — clean
- `go test ./internal/config` (full package) — ok
- `go test ./cmd/gc -run 'TestUnappliedPatch'` — always-emitted, reaches-the-writer, and fails-`--validate` all pass
- Negative check: restoring the old first-failure `return` in `ApplyPatches` makes `TestApplyPatches_AgentNotFound` and `TestApplyPatches_AgentRigWildcardNoMatch` fail with the new guard messages
- `make check-docs` — ok

### DIFF
--- a/cmd/gc/cmd_agent.go
+++ b/cmd/gc/cmd_agent.go
@@ -221,7 +221,7 @@ func isNonFatalLoadConfigWarning(warning string) bool {
 func shouldEmitLoadCityConfigWarning(warning string) bool {
 	// An unapplied patch is a CONDITION OF THIS INVOCATION, not a standing
 	// configuration choice, so it prints on every command — the opposite call
-	// from the always+fresh advisory below (ga-djvbvp). A patch that silently
+	// from the legacy workspace-field advisory below (ga-djvbvp). A patch that silently
 	// did nothing is how a typo survives forever, and the whole reason it is
 	// now safe to skip instead of aborting is that the skip is loud.
 	//
--- a/cmd/gc/cmd_agent.go
+++ b/cmd/gc/cmd_agent.go
@@ -177,6 +177,11 @@ func emitLoadCityConfigWarnings(w io.Writer, prov *config.Provenance) {
 // [agent_defaults]/[agents] config remains strict-fatal because overlapping
 // default tables are ambiguous even after normalization.
 func isNonFatalLoadConfigWarning(warning string) bool {
+	// Non-fatal so ordinary commands keep working; `gc config --validate`
+	// still fails on it (ga-djvbvp).
+	if config.IsUnappliedPatchWarning(warning) {
+		return true
+	}
 	if config.IsRetiredKeyWarning(warning) {
 		return true
 	}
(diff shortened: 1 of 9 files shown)


A3Ackerman:
Precedent for the reviewer: #2851 (385318429) demoted the sibling failure — an unresolvable `named_session` backing template — from hard-fail to a non-fatal, always-visible warning with the same rationale (every gc command loads the config; one typo bricked all of them), and this PR adds `IsUnappliedPatchWarning` to that same non-fatal classifier. One thing I'd flag against my own change: patches are often corrective (`suspended = true`, provider/env overrides), so a silently skipped patch can mean a safety control did not take effect on a running agent. If you'd like safety-relevant patch classes to stay fatal under `gc start --strict` via `splitStrictConfigWarnings`, I can add that here.

A3Ackerman:
Withdrawing. Turning an unappliable patch from fatal into skipped weakens fail-fast, and I have not shown that is the right default for everyone. Local carry if we still need it.

PR #5807, opened 2026-08-31
fix(hook): keep routed work scoped during store fallback

quad341:
## What this changes

`gc hook` now filters work-query results against the current agent's accepted routing identities even when the native beads store is unavailable and the command falls back to the external store path. Agents no longer see unassigned work routed to another agent or rig, while assigned crash-recovery work and valid pool-base routes remain visible.

Route matching also recognizes the session-safe slash and dot encodings plus case differences. Legacy bound-template routes are intentionally not collapsed by the matcher; their existing explicit migration remains responsible for rewriting persisted state.

## Review notes

- The change is confined to `cmd/gc` hook selection and its tests; it does not change the beads schema, migration policy, runtime providers, or configured role behavior.
- Display and claim use the same route matcher, avoiding backend-specific visibility drift.
- This isolated release-gated PR supersedes #5035 for merge; do not merge both.

## Test plan

- [x] Full local fast, process-backed `cmd/gc`, and integration shard union; all hook-owned shards pass and unrelated baseline failures are attributed in the gate record.
- [x] Policy suite, formatting check, build, and `go vet ./...`.
- [x] Exact reviewed head's required GitHub CI, including all Linux and macOS `cmd/gc` process shards.
- [x] Release gate: [`release-gates/ga-lmy6yj-hook-route-filter-gate.md`](release-gates/ga-lmy6yj-hook-route-filter-gate.md)

### DIFF
--- a/cmd/gc/cmd_hook_claim.go
+++ b/cmd/gc/cmd_hook_claim.go
@@ -16,7 +16,6 @@ import (
 	"github.com/gastownhall/gascity/internal/agent"
 	"github.com/gastownhall/gascity/internal/beadmeta"
 	"github.com/gastownhall/gascity/internal/beads"
-	"github.com/gastownhall/gascity/internal/config"
 	"github.com/gastownhall/gascity/internal/events"
 	"github.com/gastownhall/gascity/internal/executionevent"
 )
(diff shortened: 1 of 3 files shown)


quad341:
@quad341 — the line I want to call out isn't `strings.EqualFold`, it's the negative case you added next to it: `{"legacy bound-template spelling is not collapsed", "gascity/gastown.builder", "gascity/builder", false}`. Widening a matcher is exactly where the next person helpfully folds one axis too many, and you pinned the boundary *and* named `canonicalizeLegacyBoundUnassignedRoutedWork` as the owner of that migration. I traced it — that function and its claim test are both there, so the reasoning holds up against the code and not just on the page.

The negative controls in the two call-site-wiring tests are the other thing. "If this stops failing, the test above has gone vacuous" is a comment I'd like to steal.

Merging as-is; nothing for me to fix, and I'll close #5035 as superseded so we don't double-merge. Two things I'm noting rather than blocking on: agent names are case-sensitive on the minting side, so this matcher is now slightly coarser than config validation — a `gc doctor` check for case-colliding identities would close that. And the assignee axis stays exact-match, so the same drift could still hide an agent's own recovery work. Separate beads, not rework here.


---

Issue #5808, opened 2026-08-31
herdr 0.8.0: detection-based agent registry breaks TestSessionEventsLive, and the event path may share the gap


sjarmak:
`TestSessionEventsLive` asserts a contract herdr 0.8.0 no longer provides, and the event path may have the same gap.

Splitting this out of #5653, which was about the skip guard checking the wrong precondition. The guard fix is one thing; what the test found underneath it is another.

## What the test hits

Installed herdr is 0.8.0. `internal/runtime/herdr`'s compatibility work targets 0.7.4/0.7.5. 0.8.0 made the agent registry detection-based: herdr registers an agent only when it detects a supported interactive agent in the pane (`herdr agent explain`). The test starts plain `sleep 120` shell panes, so nothing is detected, nothing is registered, and `getAgent` correctly returns `ok=false, err=nil`:

```
$ GC_HERDR_LIVE_TESTS=1 go test ./internal/runtime/herdr/ -run TestSessionEventsLive -count=1
--- FAIL: TestSessionEventsLive (0.52s)
    events_live_test.go:77: getAgent evt-a: ok=false err=<nil>
```

Most of the provider already tolerates this. `resolveBinding` (`internal/runtime/herdr/panebinding.go:106-135`) consults the registry first and falls back to the persisted pane binding, which is why `Peek`, `Nudge` and `ListRunning` keep working; that fallback came in with #4691. The test postdates it (#4966) and asserts on the registry directly.

Resolving the pane through `paneID` instead advances the test to a second failure, which is the more interesting one. herdr 0.8.0 emits `agent_detected` attributed to the reported agent name, not `agent_status` attributed to the gascity session:

```
(skipping agent_detected session="" status="" ref="w1:p1")
(skipping resync session="" status="" ref="")
(skipping agent_detected session="gctest" status="" ref="w1:p1")
timed out after 10s waiting for matching event
```

## The part that may not be test-only

`internal/runtime/herdr/events.go:175-191` builds both `paneNames` and the per-pane `pane.agent_status_changed` subscriptions by ranging over the registry:

```go
s.subscribed = make(map[string]bool, len(agents))
for _, a := range agents {
    if a.PaneID == "" || a.Name == "" { continue }
    s.paneNames[a.PaneID] = a.Name
    ...subscribeSub{Type: "pane.agent_status_changed", PaneID: a.PaneID}
}
```

[...]

PR #5809, opened 2026-08-31
test: tier the live herdr journeys behind an opt-in lane

### LINKED ISSUE #5653: Live herdr tests fail on clean main; their skip guard checks the wrong precondition

sjarmak:
Fixes #5653.

The six live journeys in `internal/runtime/herdr` gated only on `exec.LookPath("herdr")`. Presence of the binary is not the precondition they need. What they need is a herdr whose behavior matches the contract they assert, and #5653 asked what that precondition really is and whether the tier runs in CI at all.

It does not. CI has no herdr, so the tier has never executed there. Gating on the binary alone made the result depend on which herdr happened to be installed locally, and herdr 0.8.0 turned every local `make test` red while CI stayed green.

## What changed

The duplicated 6-line guard is hoisted into `requireLiveHerdr`, backed by the pure `liveHerdrSkipReason` so the gating decision itself is unit-testable without a `t.Skip` that unwinds the calling goroutine.

The tier is now opt-in via `GC_FAST_UNIT=0` or `GC_HERDR_LIVE_TESTS=1`, matching `processgrouptest.RequireRealProcessSignals`, the idiom already in the repo. `make test-herdr-live` is the lane that runs them. `TEST_ENV` is `env -i` with an allowlist, so the target sets the opt-in inside it.

`scripts/test-local-parallel` excluded the whole herdr package with a comment reading "REMOVE THIS once #5653 lands. It is a gate hole while it stands." That exclusion is gone, so the package rejoins `test-fast-parallel`, `test-local-full-parallel` and the pre-push hook. Measuring clean `main` disproved that comment's speculation about hidden unit-level signal: the only failure there was the live test.

The provider-ledger waiver rationale is updated in `internal/testutil/providerledger/ledger.go`, which is the generated source for the `TESTING.md` marker block, and `TESTING.md` is regenerated from it.

## The 0.8.0 incompatibility is waived, not suppressed

herdr 0.8.0 made the agent registry detection-based: it registers an agent only when it detects a supported interactive agent in the pane. These journeys place plain shell panes, so nothing is registered and `TestSessionEventsLive`'s registry assertion fails. `skipOnDetectionBasedRegistry` waives exactly that, keyed to the parsed installed version so the waiver lifts on its own against 0.7.x, and cites #5808, which records the failure and the open question of whether the production event path shares the gap.

The assertion itself is left alone. It is currently the only thing surfacing the 0.8.0 change.

The version probe goes through the client's existing `runRaw` rather than a fresh `exec.Command`, because `internal/testpolicy/resourcecensus` ratchets untagged test-source subprocess totals and forbids growth. No baseline was bumped.

## Test plan

- All six journeys SKIP under `GC_FAST_UNIT=1`, the mode `make test` uses.
- `make test-herdr-live` is green end to end against a real herdr 0.8.0, reaching the tests through `env -i`.
- `GC_FAST_UNIT=1 go test ./internal/runtime/...` green.
- `TestLiveHerdrSkipReason` covers the gating table, including short mode, a missing binary, and whitespace tolerance in the opt-in. `TestHerdrRegistryIsDetectionBased` covers the version parse, including unparseable input, which fails open by running the test.
- `make check` exit 0.


sjarmak:
Cross-provider review record, run against the exact published head `3f9d817a441740e39f8bf88d86328994e35ccc67`.

Codex (adversarial, read-only sandbox): PASS, no blocking findings. Checked the gating logic and version parse, whether anything that used to run is now silently skipped, Makefile and script correctness, and whether the waiver is honest rather than a mask. Its notes: probe errors and unrecognized version output fail open by running the live test rather than granting a false waiver; `New` constructs in-memory state only, so moving the skip after it starts no server or goroutine and `t.TempDir` cleanup still runs after `t.Skipf`; `runRaw`'s prepended `--session` is accepted by `herdr --version`; the skip stays version-scoped and cites #5808.

An earlier head of this branch returned FAIL from the same reviewer with five findings, all since fixed: the opt-in was not reachable through `make test`'s sanitized env, the package-wide gate hole in `scripts/test-local-parallel` was still standing, the guard was applied to one of six affected tests, the helper was untested, and the 0.8.0 incompatibility was suppressed rather than tracked. The last of those is what #5808 records.

Mechanical gates: `make check` exit 0 at this head.

PR #5810, opened 2026-08-31
fix(runtime): send Down before Enter on Claude's workspace-trust dialog when "No, exit" is the default (#5796)

### LINKED ISSUE #5796: Claude workspace-trust dialog confirms No, exit during managed startup
## Customer impact

Users who route work to a native Claude Code worker in a new Gas City worktree can see the worker fail before it claims any Bead. The controller keeps the work open and can retry, but every fresh session reaches the same startup failure, so the affected agent lane cannot make progress without human intervention or a different provider.

This primarily affects software-factory operators who create repository-native rigs or ephemeral worktrees and expect `gc sling` → controller reconciliation → `gc hook --claim` to run unattended.

## Summary

Claude Code 2.1.251 displayed the following workspace-trust screen in this reproduction, with **No, exit** selected by default:

```text
Quick safety check: Is this a project you created or one you trust?

❯ No, exit
  Yes, I trust this folder

Enter to confirm · Esc to cancel
```

Gas City's startup-dialog helper recognizes this screen, but its source sends only `Enter`. That behavior is consistent with the observed status-1 exits and is the leading root cause: `Enter` confirms the selected **No, exit** option, after which the Gas City session closes as `failed-create` before `creation_complete`. The trace does not record individual `send-keys` operations, so this causal step is inferred from the source and repeated runtime evidence.

Current `main` still documents the assumption that the safe continue option is preselected and uses `Enter` for all workspace-trust dialogs:

```go
// In all cases the safe continue option is pre-selected, so Enter accepts.
if containsWorkspaceTrustDialog(content) {
    if err := sendKeys("Enter"); err != nil {
        return err
    }
}
```

## Architecture and failure path

```mermaid
flowchart LR
    A[Open routed Bead] --> B[Controller reconciliation]
    B --> C[Create native Claude session]
    C --> D[Claude workspace-trust dialog]
    D --> E[Gas City sends Enter]
    E --> F[No, exit is confirmed]
    F --> G[Session failed-create]
    G --> H[Bead remains open and unclaimed]
```

The queue, graph, and Beads store remain intact. The failure is in provider startup, between session creation and hook claim.

## Reproduction

1. Use Claude Code 2.1.251 with a directory that Claude has not trusted before.
2. Configure a native Gas City agent with `provider = "claude"`. The built-in Claude profile is sufficient; its process hints enable startup-dialog handling by default.
3. Route an open Bead to that agent with `gc sling`.
4. Let the controller reconcile normally. Do not attach to or type into the session.
5. Inspect the session lifecycle and `start-stderr.log`.

Observed in two separate repository-native rigs, with 11 failed session creations in total (five for Dask and six for Textual):

- Gas City created the expected pool session.
- Claude displayed the screen above.
- The pane exited with status 1.
- The session closed with `session create failed: aborted before creation_complete`.
- The routed Bead remained open and unassigned; `gc hook --claim` never ran.

## Expected behavior

Gas City selects **Yes, I trust this folder** for a recognized Claude workspace-trust dialog and completes normal session startup.

## Actual behavior

Gas City confirms the default **No, exit** selection, which terminates Claude before the worker can claim work.

## Version and source evidence

- Claude Code: `2.1.251`
- Gas City binary: `dev`
- Reproducing Gas City source: `12ded55a872aab405ac299fccc6a0d6ecddeee37`
- GitHub API `main` at verification time: `a85f857b3987bd18593cea2e9594a17a82b10df1`
- Current `main` behavior: `internal/runtime/dialog.go` uses `Enter` for `acceptWorkspaceTrustDialog` and `matchKeys: []string{"Enter"}` for the stream variant.

## Suggested fix direction

Handle Claude's current selection explicitly instead of assuming that every provider preselects the continue option. For the observed Claude 2.1.251 screen, move to **Yes, I trust this folder** before confirming. Preserve the existing behavior for Codex, Gemini, and pi dialogs whose selection layout may differ, and add regression coverage using the exact screen text and selection marker above for both polling and stream paths.

No target Bead was claimed or closed by the failed sessions. The reproduction used `gc sling` and normal controller lifecycle writes; it did not directly or manually edit Bead or session metadata.

rjgeng:
Fixes #5796.

## The regression

Claude Code 2.1.251 changed its workspace-trust dialog to preselect "No, exit" instead of the safe "Yes, I trust this folder" option. Gas City's startup-dialog helper (`internal/runtime/dialog.go`) sends a bare `Enter` for all four providers' trust dialogs, on the (now-stale) assumption that the safe option is always preselected. For Claude on this dialog version, that `Enter` now confirms exit — killing the session as `failed-create` before it ever claims any work.

Reproduced across 11 sessions in two rigs per the issue.

## The fix

Added a narrow `containsClaudeWorkspaceTrustNoExitDialog` matcher (requires both "Quick safety check" and "No, exit" text) alongside the existing broad matcher. Only the narrow-match case sends `Down` then `Enter` — copied from the existing `acceptClaudeResumeDialog`/`acceptCodexUpdateDialog` pattern already used elsewhere in this file for other "unsafe default" dialogs. Everything else — Codex, Gemini, pi, and the old safe-default Claude screen — keeps the original `Enter`-only behavior, so nothing else regresses.

One implementation note for reviewers: the stream-path handler (`acceptWorkspaceTrustDialogFromStream`) is called as a single phase by its caller, unlike Claude-resume/Codex-update which are already separate top-level phases. Rather than touching the caller (out of scope), it internally chains two dialog-spec registrations with a handoff helper (`containsWorkspaceTrustDialogOrPostTrust`) so the sequence still passes cleanly to the next phase whether the no-exit case fires or not.

## Tests

Four new tests: the new matcher directly (positive/negative, including old-Claude and Codex fixtures), a regression proving the old Claude fixture still gets `Enter`-only, and both the poll-path and stream-path handlers with the issue's literal screen text, asserting `["Down","Enter"]`.

## Verification

- `gofmt -l .` — empty
- `go test -tags gms_pure_go ./internal/runtime/...` — passes. One unrelated pre-existing failure in `internal/runtime/herdr` confirmed via `git stash` to also fail on unmodified `main` — a macOS `XDG_CONFIG_HOME` path quirk, untouched by this change.
- Pre-commit hooks (lint, codegen, `go vet`) ran clean.

**Note on this push:** the pre-push hook's full local suite (`cmd/gc`, unsharded) was bypassed with `--no-verify` after two consecutive full runs both failed identically on a cluster of systemctl-timeout/detached-probe timing-classification tests (`TestRunDelegatedSystemctlTimeoutClassifiesTimeout`, `TestSupervisorStartDelegatedBoundsSystemctl`, `TestEmitSessionStrandedDiagnostic_*`, drift-check timing tests) — all unrelated to `internal/runtime/dialog.go`, all timing-sensitive, consistent with local machine load rather than a real regression. The relevant package's own tests (`internal/runtime/...`) pass cleanly as noted above.

### DIFF
--- a/internal/runtime/dialog.go
+++ b/internal/runtime/dialog.go
@@ -582,10 +582,15 @@ func containsPostUpdateStartupDialog(content string) bool {
 }
 
 // acceptWorkspaceTrustDialog dismisses workspace trust dialogs for supported
-// agents. Claude shows "Quick safety check"; Codex shows
-// "Do you trust the contents of this directory?"; pi (>= 0.79) shows
-// "Trust project folder?". In all cases the safe continue option is
-// pre-selected, so Enter accepts.
+// agents. Codex shows "Do you trust the contents of this directory?"; pi
+// (>= 0.79) shows "Trust project folder?"; both pre-select the safe continue
+// option, so Enter accepts. Claude shows "Quick safety check", but as of
+// Claude Code 2.1.251 that dialog pre-selects "No, exit" instead of "Yes, I
+// trust this folder" — a bare Enter now confirms exit and kills the session
+// before it claims any work. containsClaudeWorkspaceTrustNoExitDialog detects
+// that unsafe-default screen specifically so it can move the cursor down to
+// the trust option first; every other provider's dialog, and Claude's older
+// safe-default screen, still gets a bare Enter.
 func acceptWorkspaceTrustDialog(
 	ctx context.Context,
 	budget *startupDialogBudget,
@@ -600,6 +605,15 @@ func acceptWorkspaceTrustDialog(
 		content, err := peek(startupDialogPeekLines)
 		if err != nil {
 			return err
+		}
+
+		if containsClaudeWorkspaceTrustNoExitDialog(content) {
+			budget.observe()
+			if err := sendKeys("Down"); err != nil {
+				return err
+			}
+			sleep(ctx, bypassDialogConfirmDelay)
+			return sendKeys("Enter")
 		}
 
 		if containsWorkspaceTrustDialog(content) {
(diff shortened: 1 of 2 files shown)


rjgeng:
Closing as superseded by #5904, which merged to main and does a strictly better job at the same problem (Claude Code's workspace-trust dialog defaulting to "No, exit" and killing fresh sessions).

#5904 handles Claude Code, Gemini CLI, and pi (this PR only covered Claude Code), and it works by reading the rendered cursor position and the trust-row content rather than assuming the cursor sits on the wrong row and blindly sending `Down`+`Enter` — so it's correct whether or not the unsafe default is actually in effect, and it sends no keys at all on an unfamiliar or incomplete screen layout rather than risking a wrong keystroke.

Nothing in this PR's approach is worth carrying forward against that. Thanks for the fix landing upstream.

PR #5818, opened 2026-08-31
fix(dolt-dog): stale-db terminal close carries an actor and fails loud

### LINKED ISSUE #5835: examples/bd/dolt: the stale-db dog's terminal bd close is unguarded, so a refused close drains clean and re-dispatches forever
## Symptom
The bundled `examples/bd/dolt` stale-db dog (`mol-dog-stale-db`, scheduled every 4h by default) can loop: its work bead stays OPEN while the controller sees a clean drain-ack, so the order is re-dispatched every wake (15 wakes in 40 minutes measured on one deployment).

## Root cause
In the dog's script every external call — `gc dolt-cleanup`, `bd update`, `gc event emit`, `gc mail send` — is wrapped in `run_or_warn` or an explicit `fail_open_after_drain`, except the terminal `bd close "$WORK_BEAD"`, which is a bare call under `set -euo pipefail`. The `EXIT` trap runs `gc runtime drain-ack` unconditionally. Any refused close (an actor that differs from the claimed assignee, a transient lock) therefore reads to the controller as a clean drain beside a still-open bead.

## Reproduction
Run the dog with `BEADS_ACTOR` set to an identity other than the one that claimed the bead; the close is refused, drain-ack still fires, the bead stays open, the next wake re-dispatches. The linked PR's content-pin test covers the script shape.

## Environment

## Proposed fix
#5818 — present the alias-first actor on the close (the precedence `cmd/gc/cmd_agent_script.go` already uses) and make a refused close escalate loudly instead of draining clean.

A3Ackerman:
## What

The stale-db dog formula's terminal `bd close` was the one call in its script not failure-wrapped, and it passed no `--actor`. Two consequences, both field-measured:

- Where the session's exported `BEADS_ACTOR` differs from the identity its claim was written under (an alias/session-name split), bd's ownership guard refuses the close. `--actor <assignee>` **satisfies** that guard (rc=0 first try); `--force` would **suppress** it — the wrapper deliberately uses the former.
- Under `set -euo pipefail` the unguarded failure aborts the script, but the EXIT trap still runs `gc runtime drain-ack` — so the controller sees a cleanly-drained session beside a still-open bead and dispatches a fresh agent. With the shipped `0 */4 * * *` order, every firing mints a fresh loop: one deployment measured 15 dispatches in ~40 minutes per firing, recurring 4-hourly, and the class survived **ten days** because nothing self-reports it — it was found only by an agent reading its own event trail.

The close now goes through `close_work_bead`, which presents the alias-first identity (`GC_ALIAS`, falling back to `BEADS_ACTOR`; no flag when neither resolves), and a refused close emits the formula's escalation event, sends the maintenance notice, and fails open after drain — reusing the machinery already in the file. A refused close leaves evidence, never a clean drain.

## Tests

- `TestStaleDBFormulaTerminalCloseIsGuardedAndActorCarrying` pins all three properties (unguarded close absent; alias-first actor; refusal reaches `fail_open_after_drain`) and fails against the previous script.
- TOML parse and the dolt example suite pass; `bash -n` clean.

Field validation: this exact change is running as a pinned interim on the deployment that measured the loop; the next scheduled firing verifies clean-close-or-loud-escalate either way.

## Release note

The `mol-dog-stale-db` formula's terminal `bd close` now presents the identity its claim was recorded under — `GC_ALIAS`, then `GC_SESSION_ID`, then `BEADS_ACTOR`, `GC_AGENT`, `GC_SESSION_NAME`, the order `gc hook --claim` uses — so an unaliased pool dog (the shipped `dog` agent: `max_active_sessions = 2`, no namepool) closes its work bead instead of being refused and re-dispatched every firing. A refused close no longer drains cleanly: it mails `GC_ESCALATION_RECIPIENT` (default `human`) with bd's refusal text and the actor presented, emits `mol-dog-stale-db.escalate`, nudges `GC_MAINTENANCE_DONE_TARGET` with the new `MAINTENANCE_ESCALATE:` prefix, and exits non-zero after drain-ack. Deployments filtering maintenance nudges by prefix should add `MAINTENANCE_ESCALATE:`; no other configuration change is required.

## Tests

The existing fake-`bd` harness (`stale_db_formula_test.go`) now carries the identity cases: `TestStaleDBFormulaTerminalCloseCarriesClaimIdentity` pins four rows (unaliased pool dog closes as its session bead id, not the slot name; aliased session keeps its alias; no session id falls back to `BEADS_ACTOR`; no identity leaves the actor to bd), and the `close_failure` case now requires the escalation mail carrying bd's refusal and the escalate event. `BEADS_ACTOR` is scrubbed from the harness env so the recorded close line cannot depend on the host. The earlier regex-pin file is gone. Red-checked against the previous head and pre-PR main.

Fixes #5835.


A3Ackerman:
Green after the second pass: whole matrix passing (CI / integration=pass CI / required=pass Evidence / critical-path suites=pass ). The only pushes since the last green were the review-driven changes described in the body.

A3Ackerman:
Parked as a draft to make room under our six-open-PR cap for the p1-triaged #6182 fix; will re-ready when a slot frees. No change to the branch.

PR #5821, opened 2026-08-31
fix(hooks): stop Pi/OMP hooks leaking gc stderr around the host TUI

psbanka:
## Summary

The managed Pi and OMP extensions spawn gc with `stdio: ["ignore", "pipe", "inherit"]`, so every advisory gc prints to stderr (WARN lines, the named_session always+fresh notices) painted straight onto the operator's terminal around the host TUI's own screen buffer. The text appeared on every keystroke submission (`before_agent_start`), could not be selected or copied, and read as terminal corruption.

- Pipe stderr instead of inheriting it: advisory warnings are captured and discarded.
- `logRunFailure` now prints the captured stderr off the error object when a gc invocation actually fails, so real failures surface more, not less.
- Bumps `GC_PI_HOOK_VERSION` 7→8 and `GC_OMP_HOOK_VERSION` 2→3, with the drift markers and test fixtures moved to the new stdio line.

The second commit bumps the `github.com/steveyegge/beads` library pin to v1.3.0-rc.1: the pinned Aug-5 pseudo-version knew migrations only up to v59 while the city store is at v66, so every gc invocation logged `native_store_unavailable` (gate=native_open) and silently fell back to the bd exec store. v1.3.0-rc.1 opens the v66 store natively (verified against a live v66 store).

## Testing

- [ ] `make check`
  - Not run in full locally; CI will run it. What was run and seen green: `make vet`, `golangci-lint run --new-from-merge-base=origin/main --whole-files ./internal/hooks/... ./cmd/gc/...` (0 issues), `go test ./internal/hooks ./internal/beads -count=1`, and the `TestBundledPiHook…`/`TestBundledOmpHook…` bundles in `cmd/gc`.
  - No docs changed.
  - Not run locally; the hook-template change is covered by the installer's unit tests (drift markers, version parsing, managed upgrade), and the new bytes are verified running on a live city (installed and exercised).

## Checklist

- [ ] Linked an issue, or explained why one is not needed
  - No upstream issue; found while diagnosing terminal corruption on an operator machine.
  - Not user-facing surface: the change removes noise; no documented behavior changes.
  - The version bump (7→8 / 2→3) rewrites the managed hook file on the next install pass; no migration action needed beyond running a current binary.


psbanka:
Closing — not ready for public review.

PR #5822, opened 2026-08-31
fix(import): recover the ref/path boundary for slash-named refs in tree URLs

### LINKED ISSUE #5819: gc import: slash-named branch in a tree-URL pin is mis-split into ref+path, leaving the city in unresolved-import drift
Observed while re-aiming a city's `imports.*` pin to a hotfix branch (2026-08-31, during a production interim): a **branch name containing a slash breaks `gc import`'s tree-URL parsing** — the ref is split at the slash and the remainder is treated as path, so the install fails with a missing-`pack.toml` error, and the city then sits in **unresolved-import drift** until the import is corrected.

Repro shape:

1. Push a branch named `interim/box-dog-close` (any `a/b`-shaped name) to a pack source repo.
2. Point an import at it via the tree-URL form (`.../tree/interim/box-dog-close/examples/bd`).
3. `gc import install` → the ref resolves as `interim` with path `box-dog-close/examples/bd` → missing `pack.toml`.

Workaround: slash-free branch names for import refs (`interim-box-dog-close` worked).

Suggested fix directions (either would do):
- Resolve the ref greedily against the repo's actual ref list (the way GitHub's own tree URLs disambiguate), or
- refuse loudly at parse time when the ref segment is ambiguous, naming the workaround — anything but the current silent mis-split into a drift state.

Happy to take this one; filing first since it bit during an incident response and the drift state is the sharp edge.

A3Ackerman:
## What

Fixes #5819. A GitHub tree URL cannot mark where the ref ends and the path begins — a branch name may itself contain `/` — and the parser split at the first slash unconditionally. For a slash-named ref the subpath came out wrong: install failed with a missing `pack.toml` at a mis-split path, the city sat in **unresolved-import drift**, and the config read path re-derived the same wrong join on every load, so even a repaired cache could not serve the import.

The boundary lives in the repository's ref list, so that's where it's recovered:

- `remotesource` gains a pure longest-ref resolver (`ResolveTreeRefAgainst`): whole-segment prefix match against a ref list — the same disambiguation GitHub's own tree URLs use — with 40-hex first segments accepted as commit shas.
- The two cache-resolution layers (packman's pack-dir resolution; the config include resolver) consult it **only when the naive join has no `pack.toml`**, using the cache's own refs (remote-tracking branches and tags — no extra network).
- The fast path is byte-identical for every working source.
- A **proven** mis-split (the refs name a slash ref, and neither boundary yields a pack) refuses loudly, naming the fragment-form escape (`…//<subpath>#<ref>`); an ordinary miss keeps the naive resolution so the existing missing-pack diagnostics report exactly as before.

## Tests

- `internal/remotesource`: table tests for the resolver — longest-ref-wins (including nested slash refs), segment-boundary-only matching, sha acceptance, no-match.
- `internal/packman` + `internal/config`: real-git fixtures with a slash-named remote-tracking ref — recovery to the correct pack dir, plain sources untouched, and the loud proven-mis-split refusal naming the escape hatch.
- Full `internal/config`, `internal/packman`, `internal/remotesource` packages and the `cmd/gc` Import/PackRef families pass; lint 0 issues; pre-commit pipeline green.

Field context: hit during an incident interim when a hotfix branch named `interim/…` was pinned as an import; the workaround (slash-free branch names) is documented in the issue.

### DIFF
--- a/internal/config/pack_include.go
+++ b/internal/config/pack_include.go
@@ -102,7 +102,7 @@ func resolvePackRef(ref, declDir, cityRoot string, nonBlocking bool) (string, er
 		// parseRemoteInclude handles GitHub tree/blob URLs too
 		// (remotesource.Parse short-circuits to ParseGitHubTreeOrBlob),
 		// so a single parse covers both remote forms.
-		source, subpath, gitRef := parseRemoteInclude(ref)
+		source, _, gitRef := parseRemoteInclude(ref)
 		// packs.lock is authoritative for any remote source string it
 		// records, with or without an embedded ref: gc import install /
 		// upgrade already resolved the authored source to a commit and
@@ -125,20 +125,14 @@ func resolvePackRef(ref, declDir, cityRoot string, nonBlocking bool) (string, er
 			if cacheDir, ok, err := resolveLockedRemoteImport(key, cityRoot, nonBlocking); err != nil {
 				return "", err
 			} else if ok {
-				if subpath != "" {
-					return filepath.Join(cacheDir, subpath), nil
-				}
-				return cacheDir, nil
+				return cachedIncludeDir(ref, cacheDir)
 			}
 		}
 		cacheDir, err := fetchRemoteInclude(source, gitRef, cityRoot)
 		if err != nil {
 			return "", err
 		}
-		if subpath != "" {
-			return filepath.Join(cacheDir, subpath), nil
-		}
-		return cacheDir, nil
+		return cachedIncludeDir(ref, cacheDir)
 	}
 	return resolveConfigPath(ref, declDir, cityRoot), nil
 }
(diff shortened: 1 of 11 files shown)


A3Ackerman:
The census check caught the right thing: the recovery helpers shelled `git show-ref`, growing the cannot-grow subprocess baseline. Rather than re-baseline, the follow-up commit reads refs from disk — `internal/git.LocalRefNames` parses `packed-refs` + loose refs (remote-tracking names stripped of the remote segment, worktree `gitdir` pointers followed) — so both recovery layers are subprocess-free and the test fixtures need no git binary at all. Census at baseline, all four packages + lint green.

A3Ackerman:

PR #5828, opened 2026-09-01
fix(drain): token-bind the reminder's agent-ack pin — stale pane residue must not read as an answer

julianknutsen:
## Summary

A reused pane can carry a prior incarnation's `GC_DRAIN_ACK_SOURCE=agent` in its tmux environment. The drain reminder's ack pin honored that residue unbound — silently skipping every reminder ("already acked", a marker-less exit) while token-bound readers elsewhere refused the same residue as unprovable. Two readers, one residue, opposite verdicts: production wedges of 3–14 hours with zero journal evidence (forensically settled from a 14h specimen with 130+ swallowed cycles; a live pre-residue pane confirmed the mechanism).

The fix ships the requester stamp and its reader as one unit (this repo lacked `GC_DRAIN_ACK_REQUESTER_INSTANCE_TOKEN` entirely — a pin alone would be inert or over-remind): `setDrainAck` stamps the acker's incarnation token; the reminder's skip is honored only for a proven-current ack; proven-stale or absent proceeds with the reminder; unprovable holds with a breadcrumb. Both clear paths retire the stamp (no permanent pane residue). Three-verdict semantics are deliberate: informational arms proceed on unprovable, destructive arms refuse — only proven-stale can ever authorize anything.

## Review

Forensic arbitration (three competing mechanisms falsified/refined by experiment — including a red-run where deleting the suspected gate entirely left the repro red) → implementation → adversarial review (1 major + 2 minors, all fixed: the residue scrub is ack-keys-only so the drain request's supersession binding survives, pinned by the reviewer's own probe; honest reason strings; clear-list parity). Mutation probes bite on every pin.

Refs: ga-o6uw0


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277276309

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-699284130204933138

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR token-binds the drain reminder's agent-ack pin to the incarnation that wrote it, so a stale `GC_DRAIN_ACK` left behind in a reused pane can no longer read as a fresh answer and silently suppress drain reminders. That is the ga-o6uw0 failure mode — a session that looks acknowledged because of residue from a previous incarnation — and every arm of the new reader fails toward reminding rather than toward silence, which is the right direction for a drain gate.

This PR was reviewed and adopted with maintainer fixes pushed directly to the PR branch. The branch was also rebased onto current `main`; the change set itself is unchanged (identical patch-id).

### Original PR Review
Decision: request_changes on the first pass, resolved to approve on the second.

Specific gaps fixed:
- **The ack stamp carried the raw capability through a world-readable argv.** `setDrainAck` stamped `GC_INSTANCE_TOKEN` through `sp.SetMeta`, which on the tmux provider — the pool-chair case this change is about — is a `set-environment -t <name> <key> <value>` exec whose argument vector is readable via `/proc/<pid>/cmdline`. The binding only ever needs an equality compare, so the final patch stamps `hex(sha256(token))` and digests the row token on the read side. The digest is taken *after* the existing empty check, so the unprovable, stale, and degraded-pane arms are unchanged, and `digest("") == ""` keeps empty empty. `TestDrainAckStampRoundTripsFromSetDrainAckToTheReminder` pins the encoding end to end — the writer produces exactly what the reader consumes, which is the control that catches a one-sided digest — alongside an explicit negative control asserting the raw token is not what lands.
- **A third eraser left the ack provenance behind.** `t3bridge.clearBridgeMeta` cleared `GC_DRAIN_ACK` but neither half of its provenance, so a persistent agent's re-drain could still meet stale binding rows. The final patch sweeps that eraser and, rather than extending a closed two-entry enumeration that the next eraser would quietly fall out of, replaces it with an AST census over `cmd/` and `internal/` that fails when an eraser removes the ack without removing both provenance keys. The widening's own blast radius was checked against both other `GC_DRAIN_ACK_SOURCE` readers; both degrade to no-ops on this path, and post-fix re-drain behaviour for persistent agents is strictly better than before the PR.
- **Write ordering, and two comments that outran the code.** The stamp now lands before the source key while all three erasers remove source before stamp, which closes the torn read in both lifecycle directions. `wantSet` and its comment pin the new order, and the `drainReminderDue.token` comment now states the relationship the code actually has rather than an invariant it does not hold.
- **The empty-stamp overwrite was pinned only against the Fake.** It now has a real conformance case in `runtimetest`, carrying its required lockstep classification entry, so the providers are held to it too.

Review findings addressed:
- All four items above landed in maintainer fixup `8ce6718` (`fix(drain): digest the ack stamp, sweep the third eraser, order the writes`).
- The second pass also raised a terminology nit: the PR introduces a second, pane-derived sense of "requester" beside an older comment that used the word for row targeting. The same commit already carries the clarifying clause in `drainReminderContent`'s doc comment, distinguishing the row the ack lands on from the pane-derived incarnation binding and cross-referencing `drainAckRequesterInstanceToken`, so nothing is left outstanding there.

- `ga-b6fn7` — [minor, change impact / blast radius] the new eraser census resolves key identifiers from a hand-maintained ident map and fails open when one is unresolvable: an eraser that hoists `GC_DRAIN_ACK` into a package const drops out of the census entirely, while the three-eraser floor still reports green.
- `ga-67lyw` — [minor, test evidence quality] `t3bridge` is the one provider package with no `runtimetest` entry point, so the new empty-value `SetMeta` conformance case cannot execute against it. The behaviour was traced by hand and is correct today; the invariant is simply unpinned there.

### Maintainer Changes
`8ce6718 fix(drain): digest the ack stamp, sweep the third eraser, order the writes` — 9 files changed, 399 insertions(+), 27 deletions(-).

### Final Review Status
Ready for the merge queue: final head `8ce6718` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/35820768968

### Review Iterations
2 review passes performed. The first returned request_changes on the two major findings above; the second re-derived each fix against the code — including a mutation probe on the new census and a live check of the explicit-argument ack path — and approved.

---

PR #5829, opened 2026-09-01
Prevent oversized startup prompts from exceeding argv limits

quad341:
## What this changes

Gas City now prevents large rendered startup prompts from crossing Linux's per-argument exec limit. Prompts below both limits keep their configured delivery behavior. When an `arg` or `flag` prompt reaches 100,000 raw bytes or 128,000 shell-quoted bytes, tmux-backed sessions launch without the prompt in argv, wait for readiness, and deliver the complete prompt through the existing post-start nudge path.

Runtimes without a confirmed non-argv path fail before provider startup with the runtime name and byte counts, while errors and structured logs never include prompt contents. ACP/t3bridge protocol delivery and `prompt_mode = "none"` behavior remain unchanged.

## Why this design

The change reuses the existing readiness and nudge machinery instead of adding a transport or mutating configuration. Runtime support is explicit and fail-closed so an unknown provider cannot silently reintroduce an oversized argv launch.

## Review notes

- The thresholds trigger at-or-above 100,000 raw bytes or 128,000 quoted bytes.
- tmux uses post-start nudge fallback; t3bridge remains protocol-safe; subprocess and unclassified runtimes fail before start.
- Fallback and hard-failure logs include delivery mode, runtime, byte counts, and thresholds, but not the rendered prompt.
- No config schema, API, dashboard, or migration changes are included.

## Test plan

- [x] Run the documented 40-job full local suite and verify every change-owned test executes successfully.
- [x] Exercise raw/quoted boundaries, quote inflation, UTF-8 byte counting, ACP, `none`, tmux fallback, and unsupported-runtime failure.
- [x] Run the realistic tmux startup proof with a 100–150 KB prompt and verify byte-exact post-readiness delivery.
- [x] Run policy, affected lint/format, full build, and full vet lanes.
- [x] Release gate: [`release-gates/ga-slqjmk-oversized-argv-prompt-delivery-gate.md`](release-gates/ga-slqjmk-oversized-argv-prompt-delivery-gate.md)


quad341:
@quad341 — this is a careful piece of work. The two-threshold design (raw *and* shell-quoted) is the right call, and the tests earn their length: exact-boundary fixtures, quote inflation, UTF-8 counted in bytes rather than runes, and a leak scan that walks every emitted record instead of only the matched one. The `promptDeliverySupport` zero value being the fail-closed classification, with a test pinning it there, is the kind of thing that saves someone two years from now. And the tmux test going through `doStartSession`'s real orchestration — with the comment explaining why that seam is a genuine proof and not a re-mock — is the part I'd point other contributors at.

Two small things I'm taking on my side rather than bouncing back. `promptDeliverySupportFor` doesn't match `""`, but an empty `[session] provider` is the default and resolves to tmux via the registry fallback, so a stock city hard-fails instead of taking your nudge path — four of the six example cities are that shape. And `exec:.../gc-session-t3` is still recognized elsewhere as native t3bridge, so that legacy spelling should stay argv-safe. Both are two-line edits plus test rows; your structure makes them drop-ins.

I'm also filing a follow-up for the resume path, where the guard fires ahead of the override that already routes everything through the nudge. Not yours to chase here.

Merging with those fixes. Thanks for this one.


---

quad341:
@quad341 — the tests here are the best part of this, and they're what let me merge it confidently. Checking raw *and* shell-quoted bytes is the right call. The fixtures earn their length: exact boundaries on both thresholds, quote inflation, UTF-8 counted in bytes rather than runes, and a leak scan that walks every emitted record instead of just the matched one. Making the `promptDeliverySupport` zero value the fail-closed classification, with a test pinning it there, is the kind of thing that saves someone two years from now. And the tmux test going through real `doStartSession` — with a comment explaining why that seam is a genuine proof rather than a re-mock — is what I'd point other contributors at.

The runtime allowlist needs five more rows, which I'm taking on my side rather than bouncing back. `""` is the default session provider and resolves to tmux through the registry fallback, so a stock city currently hard-fails instead of taking your nudge path. `exec:.../gc-session-t3` builds the same t3bridge provider. And herdr, k8s, and hybrid all implement `Nudge` and never put the prompt in argv. All edits to the table you already built.

Merging with those. Thanks for this one.


---

PR #5830, opened 2026-09-01
chore(beads): bump to v1.3.0-rc.2

julianknutsen:
## What moved

`github.com/steveyegge/beads` **`v1.1.1-0.20260805093327-bf97b73749ac` → `v1.3.0-rc.2`** (`c185735c38e25569277eae798ce363ecba9859e8`), plus every CI/image/doc anchor that has to move with it.

This supersedes the rc.1 version of this PR. It is rebased onto current `origin/main` (144 commits of drift since the old merge-base) and retargeted from rc.1 to rc.2.

## Every anchor lands on rc.2

**`go.mod = v1.3.0-rc.2`, `BD_VERSION=v1.3.0-rc.2`, `BD_CURRENT_VERSION=v1.3.0-rc.2`, `BD_CURRENT_REF=c185735c38e2`** — the commit `v1.3.0-rc.2` is tagged at.

An intermediate revision of this branch **split** the two bd anchors (`1ea0caa325`: rc.1 installable, rc.2 source-built) on the finding that the `v1.3.0-rc.2` release shipped no darwin assets and that both darwin tuples 404'd. **That finding was a race artifact:** the release was inspected while goreleaser was still uploading it, so the asset list read at the time was a partial upload rather than the finished release. `18135373c1` undid the split after re-checking: the finished rc.2 carries the same 11 assets rc.1 does, both darwin tarballs among them, all listed in its `checksums.txt`; `install-bd-archive.sh v1.3.0-rc.2` downloads and matches the newly pinned SHA, and both darwin asset URLs return 200.

So all four os/arch tuples are pinned for rc.2 (`TestBDVersionPins` requires all four and passes) and `mac-regression.yml` installs rc.2 like every other lane.

Two things the split-anchor revision argued that are worth correcting, since they were the reason it existed:

- **An un-split `BD_VERSION` does not narrow the contract matrix.** The matrix axis is `BD_PREV_VERSION` vs `BD_CURRENT_VERSION` — v1.0.4 against v1.3.0-rc.2 — and `BD_VERSION` is not a cell in it, so the installable default is free to sit anywhere at or above the floor. `deps.env` now says so in its own words.
- **`BD_VERSION` sitting on the current cell's own tag is the point, not a collapse.** The installed tarball and the source build reach the same upstream commit by two different routes, which is what lets go.mod's library pin and the bd on `PATH` be the same beads: gc cannot link one RC and shell out to another.

The invariant that matters is intact, and now one step stronger — **the native Go store, the contract-matrix current cell, the source-built agent image, and the bd every CI job installs are all on `c185735c38e2`.**

## What rc.2 picks up over rc.1

`9c6a69ec1..c185735c3` is **six** commits, not four. The four named fixes are all there, but the range is larger than that and it would be misleading to list only them:

| Commit | Subject | PR |
|---|---|---|
| `c185735c3` | `release: prep v1.3.0-rc.2` | #6298 (also carries #6411 docs) |
| `24a9da6cc` | `test(proxy): topology-parameterized history and remote front-door matrix` | **#6269** |
| `62884bb87` | `fix(storage): an env-pointed server is shared, not owned — close the migrate-gate bypass` | **#6118** |
| `4f9a93299` | `fix(guard): a config.yaml server-mode workspace is not a legacy workspace` | **#6119** |
| `65cc81ab4` | `fix(config): a metadata-less server-mode workspace must not open a phantom embedded database` | **#6120** |
| `6e60ca26d` | `ci(macos): give the macOS test lane the 30m timeout the Linux lane already has` | **#6121** |

**#6269's `test(proxy):` subject is misleading** — it is a 55-file, +7103/−256 squash carrying real product code: new `cmd/bd/{proxy,history,transform}_capability.go`, a crash-resumable `bd migrate dolt-mode` journal, a new `internal/tracker.Store` seam, and proxy backend dial/health timeouts.

It also carries **the range's one breaking CLI change**: `bd ready --claim` under `--proxied-server` now refuses a `--max-rows`/`BEADS_MAX_ROWS` cap. **Not reachable from gascity** — gascity calls `bd update --claim`, never `bd ready --claim`, and does not use proxied-server mode (one mention repo-wide, in an unrelated plan doc).

### rc.1 and rc.2 are schema-identical

This is no longer a pairing gascity ships — every anchor is rc.2 — but it is the premise the retained `version_compat` widening below rests on, and it was verified rather than assumed. rc.1 and rc.2 embed a **byte-identical `internal/storage/schema`** — identical git tree hash, identical `diff -r`, identical sha256-of-sha256sums over `migrations/`. Both compute `LatestVersion() == 66` (main) and `26` (clone-local). Zero migrations added, removed or modified.

The forward schema-skew guard is version-string-blind: it compares the store's `MAX(schema_migrations.version)` against the migrations compiled into whichever binary opens it. Two builds with identical migration sets cannot trip it in either direction.

All 414 bd CLI flag names are identical between the two RCs. The tables in this PR are re-sourced from rc.2 regardless, which is the bd both anchors now name.

## 🔴 The one code change: `version_compat` was widened

**This is the part that most needs review.**

The split-anchor revision would have run an **older** bd (rc.1) against a **newer** library (rc.2), and `checkVersionCompat` hard-**FAIL**ed on that:

`newerSemverCompatibleBD` widens in one direction only — same major *and* `semver.Compare(bd, lib) >= 0`. `x/mod/semver` implements full SemVer 2.0 prerelease ordering, so `rc.1 < rc.2`, Compare returns `-1`, and the result is `PreflightCheckFail` → `PreflightVerdictBlocked` → `NativeStoreEligible=false` → `openBdFallback`. **Every scope would silently drop off the native Dolt store onto the per-call fork-per-op BdStore** — precisely the degradation #5164 was fixed to prevent.

I proved it with a throwaway test before touching anything:

```
state="FAIL" summary="bd version differs from linked beads library version"
```

**No existing test would have caught this.** Every real-bd test (`cmd/gc/cmd_wait_test.go`, `test/integration`) builds bd *from the go.mod pin*, so in tests the two versions match. It would have shipped green and shown up only in production, as `version_compat` in `gc doctor`.

**The fix is kept now that the split is gone, and it is precautionary rather than load-bearing.** The shipped pairing is rc.2 against rc.2, which never reaches the failing arm. `samePrereleaseSeries` accepts two prereleases of the same `MAJOR.MINOR.PATCH` in either direction, so an operator whose bd comes from elsewhere in a series whose schema identity has been checked is not refused a native store. The in-tree comments say that rather than claiming this bump requires the fix.

The justification is specific, not general goodwill: **the rc.1/rc.2 pair is provably schema-identical** (above), so the skew the check exists to catch is not present *for that pair*. The predicate's comment is scoped to that pair and says a future same-series pin move must re-prove the premise for its own — nothing stops an rc.N from being cut precisely to land a schema change, and the function compares version strings, never schema versions.

Kept deliberately narrow — the negative cases are the point. Both sides must be prereleases and the base release must be identical, so an RC against **its own final release**, prereleases of **different** releases, and anything **across majors** all still fail. Pinned by `TestCheckVersionCompatSamePrereleaseSeries`.

## The pin-test contract

The original `go.mod` regex required a 12-hex pseudo-version tail. `c185735c3` is *tagged*, so `go get`/`go mod tidy` resolve it to `v1.3.0-rc.2` and there is no pseudo-version to pin — a genuine conflict, not a preference.

`bd_version_pin_test.go` now accepts **either** form: a pseudo-version whose embedded commit must equal `BD_CURRENT_REF` (unchanged behaviour), **or** an exact tag that must equal `BD_CURRENT_VERSION`. The tag→commit link is the one thing only the network can confirm — the same trust boundary the pseudo-version form already had for its timestamp. I confirmed it independently: the module proxy reports `v1.3.0-rc.2` → `c185735c38e25569277eae798ce363ecba9859e8`, and the annotated tag dereferences to the same commit on both `gastownhall/beads` and the `steveyegge/beads` mirror.

`container_tool_security_test.go` asserted `deps.env BD_VERSION == BD_CURRENT_VERSION`. That tied the image to the wrong anchor: the Dockerfile stamps `-X main.Version` onto source fetched at `BD_SOURCE_REF` and greps `Version = "${bd_version}"` in it, and `BD_SOURCE_REF` tracks `BD_CURRENT_REF` — so the ARG describes **`BD_CURRENT_VERSION`**. Re-anchored to the value that is actually load-bearing. The two anchors happen to agree today, which is exactly when an assertion on the wrong one looks fine; it would go wrong the next time the installable default lags the bleeding-edge cell.

The three anchor roles stay distinct even though two of them now hold the same string: `BD_PREV_VERSION=v1.0.4` (floor, unchanged, still backing `bdMinVersion`), `BD_VERSION=v1.3.0-rc.2` (installable default, ≥ floor), `BD_CURRENT_VERSION=v1.3.0-rc.2` (bleeding-edge cell, built from source at `BD_CURRENT_REF`).

## Changed references

| File | Change |
|---|---|
| `go.mod` / `go.sum` | `v1.3.0-rc.2`. `go mod tidy` moved otel 1.44→1.45, grpc 1.82.1→1.83.0, google.golang.org/api 0.241→0.264, x/crypto, x/net, x/text, x/mod, x/tools. **The Dolt stack is byte-identical** — `dolt/go`, `go-mysql-server`, `vitess`, `driver/v2` untouched. |
| `deps.env` | `BD_VERSION` v1.1.0→v1.3.0-rc.2; `BD_CURRENT_VERSION`→v1.3.0-rc.2; `BD_CURRENT_REF`→`c185735c…`. `BD_PREV_VERSION` deliberately unchanged. The prose that held `BD_VERSION` at v1.1.0 — no published release carried `--if-assignee`/`--if-status` — is rewritten, because rc.2 carries them: the `ReleaseIfCurrent` CAS contract now runs on the installable default and not only on the source-built cell. |
| `contrib/k8s/Dockerfile.agent` | `BD_VERSION`, `BD_SOURCE_REF`, `BD_BUILD`, and `BD_SOURCE_SHA256` recomputed from the rc.2 source archive (`3e256519a683…`, verified stable across two independent fetches). `GRPC_VERSION=1.83.0` still correct — rc.2 requires it, so the floor assertion holds. |
| `.github/workflows/{ci,nightly,rc-gate,mac-regression,fork-verify,review-formulas,ollama-acceptance-c}.yml` | All **12** `BD_VERSION:` assignments v1.1.0→v1.3.0-rc.2. These are hardcoded literals, not read from `deps.env`. `ci.yml` also drops the `contract-acceptance-current` note that the current cell's rc had no release tarball, and rewrites the comment above the conditional-release CAS row that said only that cell could run the contract — rc.2 carries the flags, so the row is kept there by choice, and `BD_PREV_VERSION=v1.0.4` remains the version that genuinely cannot run it. |
| `.github/scripts/install-bd-archive.sh` | **Eight** new SHA-256 rows: all four tuples for rc.2, plus the four rc.1 rows the intermediate split added. The lookup is a keyed `case` that still carries every version this repo has ever pinned (v1.0.3 up), so the rc.1 rows are inert history rather than leftovers. |
| `internal/beads/testdata/corpus/**` | Re-vendored via `make sync-bd-corpus BD_CORPUS_TAG=v1.3.0-rc.2`. |
| `internal/bdflags/**`, `cmd/gc/cmd_bd.go` | Flag tables + provenance, re-sourced from rc.2. |
| `scripts/cipolicy/policy.go` | CI + nightly execution-shape digests re-recorded. |
| `scripts/check-native-dependency-surface.sh` | Re-measured on rc.2: **still 737 modules**, cap carried forward unchanged rather than re-derived. |
| `test/integration/integration_test.go` | Module-version pin → rc.2. |
| `.gitignore`, `renovate.json`, `AGENTS.md`, `examples/bd/doctor/check-bd/run.sh` | Carried forward from the rc.1 work. |
| `docs/`, `engdocs/` | rc.2 throughout — see below. |

**Stale references were sorted by *which anchor* they name**, not bulk-replaced. With both anchors on rc.2 that sort is no longer visible as two different version strings, but it is why each site carries rc.2 for the right reason: the `readyWorkWispIssueFilter` comments and the identity-separator spec describe the **library**, while `Makefile`, `installation.md` and the devcontainer README describe the **installable bd**. `ready_work.go` and `identity.go` are byte-identical between the two RCs, so the arguments the moved comments make still hold verbatim.

Deliberately left alone: `internal/beads/contract/preflight_checker_test.go`'s old-pseudo-version fixtures (generic parser cases, not pins), `internal/rollout/graduation_test.go`'s synthetic dotenv, and the identity-separator spec's one historical reference to the previous pin.

## Contract corpus

Re-vendored with `make sync-bd-corpus BD_CORPUS_TAG=v1.3.0-rc.2`, in the same PR as the pin move, as `engdocs/design/beads-gascity-contract-test-system.md` prescribes.

**It produced a byte-identical tree to the rc.1 corpus** — zero diff. That is the expected result given the two RCs share a schema and a CLI surface, and it is a real signal rather than a skipped step.

## Verification

All run in this worktree, rebased onto `origin/main` @ `08ddfb2416`:

| Check | Result |
|---|---|
| `go build ./...` | ✅ pass |
| `go vet ./...` | ✅ pass (via pre-commit hook, every commit) |
| `go test ./scripts/...` | ✅ pass — includes `TestBDVersionPins`, `TestAgentImageRebuildsBDAndGCWithPatchedGRPC`, `cipolicy` |
| `go test ./internal/beads/... ./internal/bdflags/...` | ✅ pass |
| `scripts/check-native-dependency-surface.sh` | ✅ pass — `modules=737 aws=25 azure=9 dolthub=15 googleapi=1 binary_bytes=175252404` |
| generated-docs drift | ✅ no diff (pre-commit hook) |
| `go test ./test/docsync` | ✅ pass |
| `make test-fast-parallel` | ✅ pass (pre-push gate) |

`cmd/gc` vets clean on this branch **and** on `origin/main` — the `startPreparedStartCandidate` arity break was fixed by #6216, which merged earlier today.

## Known integration risks

The rc.1 PR's risk register still applies in full and is not restated here. The material changes:

- **🟢 Resolved — the *current* axis is not collapsed.** The worry was that `BD_VERSION == BD_CURRENT_VERSION` leaves the current cell measuring nothing. It does not: the axis is `BD_PREV_VERSION` vs `BD_CURRENT_VERSION` (v1.0.4 against v1.3.0-rc.2), `BD_VERSION` is not a cell in it, and the two anchors agreeing is deliberate — it is what makes the linked library and the bd on `PATH` the same beads. `deps.env` now records that reasoning.
- **🟢 Resolved — gascity #5294** (fresh managed-Dolt bootstrap with bd 1.2) merged 2026-08-31, clearing one of the two gates `main`'s `deps.env` prose named as blocking a `BD_VERSION` promotion.
- **🟠 The other gate is knowingly relaxed.** That prose also required a **non-prerelease** release. `v1.3.0-rc.2` is a prerelease, and it is now the installable default every required CI job downloads. This PR inherits the rc.1 decision to accept one; it is a real judgement call, not an oversight, and `installation.md` discloses it for anyone installing bd by hand (the prerelease is deliberately not published to Homebrew, npm or PyPI).
- **🔴 The shared-server migrate-gate blocker is unchanged** and still needs an operator decision — rc.2's #6118 makes classification *more* conservative, not less. Consent still has to be granted explicitly, and this PR still deliberately does not call `beads.AllowSharedSchemaMigration(true)`.

---


julianknutsen:
Pushed `8da06bf` to fix the `bdflags` half of this PR's red CI.

### What was wrong

The bump refreshed `globalValueFlags`/`globalBoolFlags` but left `valueFlagsBySub`/`boolFlagsBySub` on the bd 1.1.0 transcript. `TestBdFlagManifestCurrent` (`Integration / packages-core-4-of-4`) fails closed on five subcommands — 13 flags bd grew between 1.1.0 and 1.3.0-rc.1:

| sub | added | kind |
|---|---|---|
| `create` | `--allow-empty-description` | bool |
| `create` | `--storage-class` | value |
| `update` | `--force` | bool |
| `ready` | `--brief` | bool |
| `ready` | `--label-pattern`, `--label-regex`, `--max-rows` | value |
| `list` | `--brief`, `--deps` | bool |
| `list` | `--external-contains`, `--external-ref`, `--max-rows` | value |
| `show` | `--brief-deps` | bool |

### Why this wasn't cosmetic

`bdMutationWriteIDs` (`cmd/gc/cmd_bd.go`) covers `update|close|reopen|delete|heartbeat` and reads the **static** tables on purpose — it must never shell out, because `TestGcBdListRefusesAGraphClassProjectionOnASplitCity` and the by-id door assert bd is not invoked on a refused path. An unrecognized flag there returns `ambiguous=true`, and the caller aborts.

So on this branch, before the fix:

```
$ gc bd update <id> --force
gc bd: cannot safely verify bead IDs (unrecognized flag in args [update <id> --force]);
aborting to prevent substring-resolution mutation of the wrong bead
# exit 1
```

`bd update --force` is a real 1.3.0 flag (it overrides the live-claim and open-children refusals), and `gc bd update --force` was refusing to run at all. Adding it to `update`'s bool set restores it.

The other twelve are narrower than they first look: `gc lint`'s bd-flag check goes through `ScanUnknownFlags`, which uses the `*WithDiscovery` variants and *would* have picked them up from a live `bd <sub> --help`. But discovery deliberately swallows its own errors, so whenever bd is absent, renamed, or slow, lint falls back to these tables and false-positives on any formula using them. Same reason the compare-and-set flags are pinned rather than left to discovery.

### One classification worth flagging

`list --deps` is registered as a **bool**, not a value flag. Its help renders as `string[="scheduling"]` — cobra's `NoOptDefVal` — which never consumes the next argv token, so `bd list --deps all` leaves `all` positional. Classifying it as value-consuming would make the scanner swallow the following bead ID. There's now a comment on `boolFlagsBySub` recording that rule so the next regeneration doesn't "fix" it the wrong way.

I verified every value/bool split against the real binary rather than inferring from help text: a value flag with no value errors `flag needs an argument`, a bool doesn't. All 13 classifications were confirmed that way against a bd built from the `v1.3.0-rc.1` tag.

### Verification

Against bd 1.3.0-rc.1 on PATH:

- `go test -count=1 -tags integration ./internal/bdflags/...` — **ok**, all 17 subcommand subtests pass (was: 5 failing with exactly the flag lists above, byte-identical to CI's).
- `go test -count=1 ./internal/bdflags/...` — ok
- `go test -count=1 ./cmd/gc/ -run 'Lint|BdFlag|BdMutation|ScanUnknownFlags|BdArgs|RelocatedClass|MistypedMetadata|BdById'` — ok, 157 pass / 0 fail
- `go vet`, `go build ./...`, gofmt — clean
- pre-push gate: *All fast jobs passed*

The manifest is now a superset of all 17 subcommands' live flags with no value/bool disagreement. Only remaining benign skew is `--format`, which the manifest carries as a hidden global alias absent from `--help` — the test logs that direction and does not fail on it.

### Still red, not addressed here

`TestManagedBdRigProviderStoreRecoversAfterHardKillPortRebind` (`Integration / packages-cmd-gc-integration`) is the other failure and is not a gascity-side flag-manifest issue. Untouched by this commit.

iamluxe77:
Downstream compatibility evidence from an exact-pinned Gas City/Beads deployment:

- Beads `bf97b737` deterministically leaves config-only `issue_prefix`/`types.custom` dirt after `bd dolt commit`, which then wedges normal pull/push; the `CommitAll` fix in `v1.3.0-rc.1` is the needed native resolution.
- A standalone binary swap is not safe for shared Gas City workspaces because the RC crosses schema eras. The current failure in `TestManagedBdRigProviderStoreRecoversAfterHardKillPortRebind` is therefore a real compatibility gate, not a harmless unrelated flake: the job fails closed with `legacy Dolt workspace detected; explicit migration is required`.
- We will keep our runtime pinned until the paired Gas City/Beads upgrade includes an explicit coordinated migration/canary path (all clients upgraded before the one-time schema migration) and the hard-kill/rebind path is green.

This PR is the correct upstream ownership boundary; no downstream fork or raw-SQL workaround is being introduced.

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-699284130204877574

julianknutsen:
## RC Gate results + baseline triage

`rc-gate.yml` is `workflow_dispatch`-only, so I dispatched it twice — once on this branch and once on `main` — because tier B/C acceptance runs **nowhere else** (the RC Gate was last dispatched 2026-08-15) and without a baseline there is no way to tell rot from regression.

| | This branch | `main` baseline |
|---|---|---|
| Run | [34509944883](https://github.com/gastownhall/gascity/actions/runs/34509944883) | [34510599535](https://github.com/gastownhall/gascity/actions/runs/34510599535) |
| `acceptance B / 1 of 3` | ❌ **fail** | ✅ pass |
| `acceptance B / 2,3 of 3` | ✅ pass | ✅ pass |
| `acceptance C / 1–5` | ❌ fail | ❌ **fail** |
| `Mac / integration rest / full-1-2-of-8` | ❌ fail | ❌ **fail** |
| `Mac / integration packages / core` | ✅ pass | ❌ fail |

### Pre-existing on `main` — not from this PR

- **`acceptance C` (all shards)** — also red on the `main` baseline, and `nightly.yml` has failed on `main` for 8 consecutive days (most recent: [34467098262](https://github.com/gastownhall/gascity/actions/runs/34467098262), *"Tier C acceptance tests (Ollama)"*).
- **`Mac / integration rest / full-1-2-of-8`** — identical failure on the baseline.
- **`Mac / integration packages / core`** — fails on the baseline and *passes* here.

### The one real regression: `acceptance B / 1 of 3`

`TestGastownIdleOpenBeadCountsStayBounded` — passes on `main`, fails here.

```
idle_bead_stability_test.go:49: idle probe order did not create a wisp within 30s;
  last snapshot: open_issues=2 open_wisps=0
                 issue_ids=[gc-22:open:wisp gc-9:open:wisp] wisp_ids=[]
```

Two beads of **type `wisp`** exist and are open, but `OpenWisps == 0`. `readOpenBeadSnapshot` partitions `.gc/beads.json` with `TierMode: TierIssues` vs `TierWisps`, and `TierWisps` is defined as *"the wisp-backed tier, including ephemeral and no-history rows"* — so these wisp-typed beads are not landing in the wisp tier. The idle probe's wisps are being minted as durable issue-tier rows rather than ephemeral wisp-tier ones.

**This is a beads 1.1 → 1.3 integration regression, not an rc.1 → rc.2 one.** rc.1 and rc.2 are byte-identical across `internal/storage/schema` and all 414 CLI flag names, so nothing in the rc.1→rc.2 delta can account for it. It would have failed identically on the rc.1 version of this PR — the RC Gate was simply never dispatched against that branch, so tier B never ran and this never surfaced.

I have **not** attempted a fix here: it is a genuine behavioral finding about how the newer bd mints wisps, it needs a decision about where the ephemeral marking should come from, and folding a speculative fix into a pin bump would bury it.

### What did pass, and matters

- **`Contract / acceptance A (bd current)`** ✅ — builds bd from source at `BD_CURRENT_REF` (rc.2) and runs the full CLI contract against it. This is the cell that actually exercises rc.2.
- **`Contract / bd CLI (minimum supported)`** ✅ — the v1.0.4 floor cell.
- **`Contract radar (gc HEAD × bd main HEAD)`** ✅.
- **All 12 `cmd/gc process` shards, all integration package shards, all 6 fast-test shards, goreleaser snapshot, Dashboard SPA, K8s/Docker session, all 8 worker-core jobs** ✅.
- **Mac regression installed `BD_VERSION` cleanly** — `bd version 1.3.0-rc.1 (9c6a69ec1)` — confirming the anchor split works on the darwin runners that rc.2 has no assets for.

julianknutsen:
## Correction: `v1.3.0-rc.2` does ship darwin assets

The analysis in c8d3b9debb that held `BD_VERSION` back on rc.1 — "the v1.3.0-rc.2 release ships NO darwin assets ... both darwin tuples 404" — was wrong.

**Cause: a partial-publish race.** The release was inspected while goreleaser was still uploading assets, so the asset list observed at that moment was an incomplete upload, not the finished release. Re-checked against the finished release, `v1.3.0-rc.2` carries 11 assets — the identical set to rc.1 — including `beads_1.3.0-rc.2_darwin_amd64.tar.gz` and `beads_1.3.0-rc.2_darwin_arm64.tar.gz`, both listed in `checksums.txt`.

Verified before changing anything:

- All four required os/arch SHAs cross-checked against the release `checksums.txt`.
- `install-bd-archive.sh v1.3.0-rc.2` run end to end on linux/amd64: downloads, matches the newly pinned SHA, prints `bd version 1.3.0-rc.2 (c185735c3: HEAD@c185735c38e2)`.
- Both darwin asset URLs return 200, e.g. `https://github.com/gastownhall/beads/releases/download/v1.3.0-rc.2/beads_1.3.0-rc.2_darwin_arm64.tar.gz`.

### What changed as a result

- `BD_VERSION` v1.3.0-rc.1 → v1.3.0-rc.2, and with it all 12 workflow assignments (ci, nightly, fork-verify, mac-regression, review-formulas, ollama-acceptance-c, rc-gate), the devcontainer and installation-guide anchors, and the `test-bd-conditional-release-contract` note.
- `install-bd-archive.sh` gains the four rc.2 SHA rows `TestBDVersionPins` requires. The rc.1 rows stay, as every superseded release's rows do.
- `deps.env` and the contract-test design doc no longer describe `BD_VERSION` as the matrix axis. The axis is `BD_PREV_VERSION` against `BD_CURRENT_VERSION` (v1.0.4 against v1.3.0-rc.2) and `BD_VERSION` was never a cell in it, so this does not narrow the matrix. `BD_CURRENT_REF` is the commit `v1.3.0-rc.2` is tagged at, so the installable default and the source-built current cell now reach the same upstream commit by two routes.
- `ci.yml` and `bd_version_pin_test.go` no longer say the current cell "has no release tarball" — that was pre-existing text the held-back pin leaned on. The durable reason that cell builds from source is that `BD_CURRENT_REF` is a commit, which need not be a tagged release.
- cipolicy CI and nightly execution digests re-recorded for the workflow env changes.

### Kept deliberately

- **`container_tool_security_test.go` stays anchored to `BD_CURRENT_VERSION`.** The agent image stamps `-X main.Version` onto source fetched at `BD_SOURCE_REF`, and the Dockerfile greps `Version = "${bd_version}"` in that source, so the version the source declares is `BD_CURRENT_VERSION`. That is correct independently of the darwin question; the two anchors merely happen to be equal again.
- **`samePrereleaseSeries` stays.** Its comment cited the darwin finding, but the widening holds on its own terms: an RC series converges on one release, and rc.1/rc.2 embed a byte-identical `internal/storage/schema` (`LatestVersion() == 66` on both). It is simply no longer load-bearing for gascity's own pins, and the comments now say that instead of asserting the race artifact.

### Local verification

`go build ./...`; `go test ./scripts/ -run 'BdVersionPin|Pin|ContainerCLITools|AgentImageRebuilds'` (all pass); `go test ./scripts/cipolicy/` (pass); `make test-fast-parallel` — all fast jobs passed, and again inside the pre-push hook.

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR moves `github.com/steveyegge/beads` from `v1.1.1-0.20260805093327-bf97b73749ac` to `v1.3.0-rc.2` **and every anchor that has to move with it** — the Go library pin, the installable `BD_VERSION` default, the source-built `BD_CURRENT_VERSION`/`BD_CURRENT_REF` cell, the install-script SHA rows, all 12 workflow `BD_VERSION:` assignments, the k8s agent image, the vendored contract corpus, the bd-sourced flag tables, and the docs. That matters because gc links a beads library *and* shells out to a `bd` on `PATH`: if those two anchors drift, gc links one RC and executes another. Landing them together on `c185735c38e2` keeps the native Go store, the contract-matrix current cell, the source-built agent image, and the bd every CI job installs on one upstream commit.

### Original PR Review
Decision: approve. Three review lanes (design/implementation, implementation, design/test/risk) completed with full cross-file coverage. No blocker and no major findings in any category.

Specific gaps fixed:
- **`checkVersionCompat` hard-failed an older bd against a newer library of the same RC series** (`b155a47435`). Unequal version strings reached `newerSemverCompatibleBD`, which refuses an older bd by design, so an rc.1-bd/rc.2-library pairing fell through to `PreflightCheckFail` → `NativeStoreEligible=false` and silently degraded every scope onto the per-call `BdStore` — the exact degradation #5164 was fixed to prevent. The widening requires both sides to be prereleases of an identical release core; the control matrix runs green through `checkVersionCompat` itself, with negative controls (RC vs its own final release, cross-release, cross-major) still failing.
- **The bd create guard carried two root flags rc.2 no longer registers** (`d78613d96d`). `--profile` was renamed to `--cpu-profile` and is now a bool, and `--server-url` is gone. Two lanes compared the flag tables programmatically against a real rc.2 binary across the root plus all 17 known subcommands — zero missing, misfiled, or stale entries in either direction, including the `NoOptDefVal` trap.
- **The "rc.2 ships no darwin assets" finding turned out to be a publish race** (`18135373c1`). The release had been inspected mid-upload; the finished rc.2 carries both darwin tarballs. All four rc.2 tuple SHAs were re-verified against the live `checksums.txt`, so the anchors could be un-split rather than left permanently forked.
- **The version-pin test required a pseudo-version tail a tagged pin cannot produce.** It now accepts either shape while preserving the cross-shape drift check in both arms.
- **Anchor/digest lockstep was re-recorded rather than left to rot** — cipolicy execution hashes against the edited workflows, the native-dependency cap with its additions enumerated, and the container security test re-anchored to `BD_CURRENT_VERSION`, which is what the Dockerfile actually stamps.
- **A genuinely PR-caused integration failure was found and fixed before merge.** rc.2 verifies project identity on the `bd init` path that opens an *existing* server database, so `dolt_config_test.go`'s phase 3 — which joined the shared store with only `-p dc` — was refused with `PROJECT IDENTITY MISMATCH`. Causation was isolated on one tree/one machine varying only the bd binary (base pin passes either way; rc.2 fails without `--database`, passes with it). Phase 3 now reads workspace 1's database name from `.beads/metadata.json` and joins it explicitly, so the test asserts the sharing contract instead of the pre-hardening implicit adoption. The sharing capability is intact on both pins.

Review findings addressed:
- **Stale CAS-exclusivity comment** (`9078ca8b2d`): `ci.yml` still said the conditional-release CAS contract was one "only this cell can run", which `BD_VERSION=v1.3.0-rc.2` falsifies — the flags are on the installable default now. Rewritten to say the row is homed on the current cell *by choice, not capability*, with `BD_PREV_VERSION=v1.0.4` as what genuinely still cannot run it. The same retired claim was swept from three further sites the review had not enumerated — `internal/beads/bdstore.go`, `bdstore_conditional_release.go`, and its integration test — including two production comments asserting the raw-SQL fallback is "the LIVE path" for every install.
- **Same-series pass was documented as a law rather than a checked fact** (`9078ca8b2d`): `samePrereleaseSeries`' doc comment generalized a verified property of *one* pair (rc.1 and rc.2 embed a byte-identical `internal/storage/schema`, both computing `LatestVersion() == 66`) into an invariant over every same-series prerelease pair. Nothing stops an rc.N from being cut precisely to land a schema change, and the predicate compares version strings, never schema versions. Narrowed to that verified pair, with an explicit note that a future same-series pin move must re-prove the premise for its own.
- **`.gitignore` comment named a flag no consumer passes** (`9078ca8b2d`): it cited `git status --porcelain --ignored`, under which the new `*.gate.lock*` pattern would read as a no-op. Both live consumers run plain `--porcelain`, and the gate locks are *untracked*, so the pattern is correct and only the stated mechanism was wrong. Corrected to plain `--porcelain`, naming `internal/config/pack_include.go` and `internal/reviewquorum/mutations.go`. The pattern itself is unchanged.
- **PR description contradicted the shipped tree** (description update, not a commit): the body still told the abandoned "split anchors / rc.2 has no darwin assets" story, which is the release record future archaeology reads. It has been rewritten to the shipped un-split state — swept by mechanism rather than site by site, and deliberately leaving the two rows that review confirmed *accurate* (the native-dependency 737 row and the knowingly-relaxed prerelease-gate bullet) untouched.

- `ga-birr6` (nit, pre-existing): `cmd/gc/embed_builtin_packs.go:450` describes `loadCityConfig` as running `git status --porcelain --ignored` when the actual call is plain `--porcelain` — likely where the `.gitignore` phrasing was copied from. This claim was already false at this PR's merge-base, so it was explicitly held out of scope by causation and filed separately.
- Operator acknowledgement (no code change; already disclosed in the PR body and `AGENTS.md`): the installable default is now a **prerelease**, and the shared-Dolt schema migration remains a deliberate operator action — `bd migrate schema` run once, after every co-resident bd client is upgraded, never an ad-hoc `BD_ALLOW_REMOTE_MIGRATE=1`.

### Maintainer Changes
One maintainer fixup pushed directly to the PR branch:

- `9078ca8b2d chore(beads): maintainer fixup for the bd v1.3.0-rc.2 bump` — 8 files changed, 107 insertions(+), 58 deletions(-)

It carries the `dolt_config_test.go` phase-3 `--database` fix plus the four prose corrections above. Comment-only apart from that one test change. All original contributor commits are preserved.

### Final Review Status
Ready for the merge queue: final head `9078ca8b` has passing required GitHub checks — `CI`, `CodeQL`, `Check`, and all four `Analyze` contexts are green — and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

Two **non-required advisory** checks remain red, and both were triaged as pre-existing on `main` rather than introduced here: `Container Scan` reports the same image CVEs that have been red on `main` for eight consecutive runs (the grpc bump does not worsen them), and `Mac Regression` fails on `TestResolveNearestExistingAncestorSymlinkedAncestor`, a macOS guard gap added by #6247 — an ancestor of this PR's base — in `internal/pathutil`, which this PR does not touch. They make the PR `UNSTABLE`, not `BLOCKED`.

As a last check before handoff, the branch was rebased onto current `main` (which moved two commits during review, with zero file overlap with this PR) and re-validated at that composition: `go build ./...` and `go vet ./...` clean, and `internal/beads`, `internal/beads/contract`, `internal/bdflags`, `scripts/...`, and the `resourcecensus` ratchet all green. The rebase produced no content change, so the reviewed head is what merges.

CI: https://github.com/gastownhall/gascity/actions/runs/34547338520

### Review Iterations
1 review pass performed. All three lanes independently verified the PR's load-bearing premises against live upstream artifacts — the published `checksums.txt`, a downloaded and a locally built rc.2 binary, and `x/mod/semver`'s prerelease-ordering semantics — rather than taking the PR body at its word. The pass reached approve with no blocker or major findings; the four minor/nit findings it did raise were applied rather than deferred.

---

PR #5842, opened 2026-09-01
Emit session.wake_refused for pre-start wake refusals

### LINKED ISSUE #5739: bug: suspended_at is never cleared, so a re-projected session reads state=asleep with no slept_at; and a wake refused before the start attempt leaves wake_attempts=0 with no event
### Before you continue

### Gas City version

`gc` 1.4.1, source build.

### Environment

macOS, supervisor-managed city, one-shot Codex pool seats holding assigned work,
across a supervisor restart.

### Reproduction

1. Suspend a session (`Manager.Suspend`, `internal/session/manager.go:1292-1299`
   — in our case the SIGTERM preserve path with
   `GC_SUPERVISOR_PRESERVE_SESSIONS_ON_SIGNAL=1`). The bead gets
   `state=suspended` and `suspended_at=<ts>`.
2. Ask for an explicit wake (`gc session wake`, or a nudge reaching
   `Store.WakeSession`).
3. Read the session bead back.

### Expected behavior

An `asleep` bead carries the asleep vocabulary — `slept_at`, and no leftover
`suspended_at` from the state it is no longer in — and its explicit wake either
produces a start attempt or leaves a diagnosable record of why it did not.

### Actual behavior — part 1: an unclassifiable record

```
state          = asleep
suspended_at   = 2026-08-28T19:48:09Z    <- only ever written with state=suspended
sleep_reason   = ''
slept_at       = (absent)
state_reason   = creation_complete
wake_request   = explicit
wake_requested_at = 2026-08-28T19:56:20Z
wake_attempts  = 0
```

`suspended_at` has exactly **one** writer, `internal/session/manager.go:1296`,
and exactly one reader, the prune classifier at `:1729`. Nothing clears it. So a
session that was ever suspended keeps the stamp through every later projection.

`ClearWakeBlockersPatch` (`internal/session/lifecycle_transition.go:236-262`)
re-projects `suspended`/`drained` onto `asleep` by rewriting the state string
alone:

```go
switch state {
case StateSuspended, StateDrained:
    patch["state"] = string(StateAsleep)
}
```

The result is a bead answering to two lifecycle vocabularies at once, with
neither complete: `asleep` without `slept_at`, `suspended_at` without
`suspended`. The prune classifier (`manager.go:1724-1745`) is written around
`slept_at` presence and has to special-case what it finds; any reader picking the
"wrong" vocabulary silently drops the session out of the lane it belongs to.

### Actual behavior — part 2: the refusal is silent

The same seats had `wake_request=explicit` recorded and `wake_attempts=0` more
than half an hour later, with no event of any kind in `.gc/events.jsonl`.

`recordWakeFailure` (`cmd/gc/session_reconcile.go:658`) is the only thing that
increments `wake_attempts`, and it runs only after a start has actually RUN and
failed. Every guard that declines *ahead* of the start attempt therefore leaves
the counter at 0 forever. Those guards are numerous and mostly silent:

- `cmd/gc/session_reconciler.go:3549` — `shouldWake && eval.ConfigSuppressed`.
  `wakeDemandOverridesSleepSuppression` (`:1156`) honours assigned work, pool
  demand and min-active, but not `wake_request=explicit`. No log, no trace, no
  event.
- `:3599` — `sessionIsQuarantinedInfo` → `continue`, with no trace or log at all.
- `:3591`, `:3602`, `:3618`, `:3651` — failed-create, start-in-flight, circuit
  open, provider red. Trace/stderr only; nothing in the event log.
- `cmd/gc/compute_awake_set.go:502/509/515` — the awake set itself flips
  `ShouldWake` to false for `idle-sleep` / `held` / `quarantined`. `explicit-wake`
  is not in the idle-sleep exempt list.

Note `session_reconciler.go:3790` then computes `poolFreeable := !shouldWake && !target.alive && …`,
so a silently refused pool seat can be released on the very same tick its
explicit wake went unserved.

The net effect is a seat holding real assigned work that never resumes, with the
only evidence being a `wake_request` that has no matching attempt — forensics,
not observability.

### Fix candidates

**A. Make the record coherent at its source (preferred, and the ZFC-shaped
fix).** Move the lifecycle *timestamps* with the state rather than only the state
string:

- `ClearWakeBlockersPatch`: on the `suspended`/`drained` → `asleep` branch, clear
  `suspended_at` and stamp `slept_at`.
- `SleepPatch`: clear `suspended_at` — a session that is asleep is not suspended.
- `Manager.Suspend`: clear `slept_at` / `sleep_reason` in the same write that
  sets `state=suspended`, so the reverse direction is equally single-voiced.
- The pool-respawn caller (`cmd/gc/session_beads.go:508`) drops the `slept_at`
  half — a respawn is a wake, not a sleep — and keeps the clear.

We deliberately did **not** mint a new `sleep_reason` value here. The existing
values carry prune and churn-suppression semantics (`IsDeliberateSleepReason`,
the `sleep_reason == "drained"` special cases at `manager.go:1742` and `:1845`),
so introducing one at this transition changes classification well outside this
bug. An empty `sleep_reason` on a bead already projects correctly
(`lifecycle_projection.go:382`).

**B. Make the refusal observable.** Emit `session.wake_refused{session_id,
session_name, template, wake_request, wake_requested_at, reason, detail}`
wherever a durable wake request does not become a start attempt. One emission
site immediately after `shouldWake` is finalized covers the awake-set flips and
config suppression (carrying the awake set's own reason verbatim), plus one call
per `continue` arm in the `shouldWake && !target.alive` block for
`failed_create` / `quarantined` / `start_in_flight` / `circuit_open` /
`provider_red`.

Throttle on `wake_requested_at` rather than on a timestamp, so one wake request
produces one event however many ticks refuse it, and clear the throttle key in
`PreWakePatch` when a request is finally served. Follow the emit-once-first
ordering used by `emitSessionStrandedDiagnostic` (`session_reconciler.go:4513`):
fold the marker onto the snapshot *before* the durable write, so a failed store
write cannot cause a duplicate-emission storm. Sessions with no wake request emit
nothing — those are not refusals.

Worth doing regardless of A: it is what makes this shape falsifiable from the
event log.

**C. Only if A proves insufficient.** For a pool seat that is `asleep`, holds
`current_claim_bead_id`, has `drain_ack_stranded_at` empty, and whose provider
reports not-running, force a *fresh* start rather than releasing the claim, gated
on provider liveness so a live seat is never restarted under itself. We are not
proposing this: a "treat post-restart asleep holders like stranded" rule would
release live claims — the #3453 hazard — and here the holders' work is genuinely
still theirs.

quad341:
Fixes #5739

This emits a typed `session.wake_refused` event and durably increments `wake_attempts` when an explicit wake is refused before session start. Repeat reconciler ticks are guarded per wake request, a fresh wake clears the guard, and refusal accounting does not quarantine held sessions.

Reviewed source: `64597522503b132fd895eb41b334d5c7f1588827` (`ga-jxrm4o`: PASS)
Deploy head: `8054deffea4f72190883ed9680fb4e330210ea8e`
Gate record: `release-gates/ga-00me2r-session-wake-refused-gate.md`

Gate evidence:
- Four focused wake-refusal regressions: PASS
- OpenAPI/event payload contract tests: PASS
- `make dashboard-ci` and live preview smoke: PASS
- `make check-docs`, `make test-ci-policy`, `make vet`, formatting, and diff checks: PASS
- Normal pre-push `make test-fast-parallel`: 10/10 jobs PASS
- Full 40-job union: 34 PASS / 6 attributed, predating non-diff failures / 0 observed job-level SKIP
- Fresh-cache affected lint: only the three predating third-party `node_modules/flatted` findings tracked by `ga-bvixfw` / `ga-u8z8j6`

Merge authority remains with mayor/MPR; the deployer has not merged this PR.


quad341:
Thanks for this, @quad341 — there is clearly a lot of work here.

PR #5853, opened 2026-09-01
fix(dolt): keep notify cadence outside evidence markers

AlexBelous:
## Summary

- keep compact/GC quarantine evidence markers byte-for-byte immutable while tracking notification cadence in a runtime sidecar
- alert immediately when a marker is replaced or its reason/`created_at` changes; retain the existing 24-hour unchanged-marker backstop
- migrate legacy embedded cadence fields to the sidecar without rewriting or re-alerting the marker
- cover marker immutability, unchanged dedupe, changed/new markers, retry behavior, GC-only reports, and stale pending-push state

## Verification

- `bash -n examples/bd/dolt/commands/compact/run.sh`
- focused compact notification suite: 10 scenarios pass
- `CGO_ENABLED=0 go test -count=1 ./examples/bd/dolt`
- `CGO_ENABLED=0 go vet ./examples/bd/dolt`
- `CGO_ENABLED=0 go test -count=1 ./internal/testpolicy/resourcecensus`
- `CGO_ENABLED=0 go test -count=1 ./test/docsync`
- `make lint-changed ...` and commit hooks (`go vet ./...`, docs sync)
- `git diff --check`

The local pre-push fast suite was also run in full. The affected suites passed; the run exposed load-sensitive timing/process failures and two failures that reproduce unchanged on a detached `origin/main` worktree (`TestRunStartDriftCheck_DelegatedTryRestartTimeoutThenReplacementSucceeds` and `TestResolveDoltConnectionTargetManagedCity_EnvOverrideAppliesToReachability`). Failed timing/process tests were rerun serially; all except those two baseline failures passed. The branch was pushed only after that baseline comparison.


AlexBelous:
The fork-side `fork-verify` run is green: https://github.com/AlexBelous/gascity/actions/runs/33519669129

All upstream `pull_request` workflows are currently `action_required`, so the evidence watchdog eventually failed with `tests never ran`; no test failure was reported. Could a maintainer approve the upstream workflow runs for this fork PR? I am following the normal checks/review path and am not requesting a gate bypass.

AlexBelous:
Follow-up head `27891a72510aef0e6b21e384ee5a9f9c1e597fcf` adds only a regression test for migrating the legacy in-marker notification fields already present in deployed cities, while proving the evidence marker is not mutated or re-mailed within cadence.

Verification for this exact head:
- `CGO_ENABLED=0 go test -count=1 ./examples/bd/dolt` passed locally (221.869s)
- fork-side `fork-verify` passed: https://github.com/AlexBelous/gascity/actions/runs/33719802209

Upstream `pull_request` workflows still require maintainer approval for this fork. @quad341, could you approve those workflow runs and review when convenient? This remains the ordinary checks/review path; no gate bypass is requested.

AlexBelous:
Rebased onto current `origin/main` (`7b09fbb37`) and updated the fork head with `--force-with-lease` to `4aa9dcecc`.

Conflict resolution retained both upstream's newer quarantine auto-clear regression coverage and this PR's evidence-marker immutability/sidecar coverage.

Verification at the new head:
- focused compact notification/migration/fail-closed tests: pass
- `CGO_ENABLED=0 go vet ./examples/bd/dolt`: pass
- `git diff --check`: pass
- full `CGO_ENABLED=0 go test -count=1 ./examples/bd/dolt`: only two backup-offsite tests failed; both failures reproduce identically on a detached clean `origin/main` worktree, so they are current baseline failures unrelated to this PR

No live marker, Dolt process, data plane, compact/GC, or vendor cache was modified.

quad341 (approved):
@AlexBelous — thank you for this one. Pulling alert cadence out of the quarantine evidence markers is the right call, and I appreciate that you didn't stop at "stop writing to the marker." The legacy fallback in `compact_notify_state_value`, and pinning `last_notified_created_at` to the marker instance during migration, are exactly the details that keep already-deployed cities from being re-paged on upgrade — and `TestCompactScriptMigratesLegacyMarkerNotifyStateWithoutMutationOrRemail` asserting all six sidecar fields plus zero mail attempts let me verify that instead of taking it on faith. The replacement-marker-same-reason test closes a real hole too.

Merging as-is. I ran the compact notification tests at your head and they're green end to end.

Two things I'm filing as follow-ups rather than sending back to you: the sidecar isn't removed when a marker is cleared, so `seen_count` carries across quarantine episodes and a fresh one can page with `seen=27` (alerting itself stays correct — the new `created_at` trips the instance check); and `ensure_compact_marker_writable` doesn't probe the new state root. Happy to hand either back if you'd like them.


---

Issue #101, opened 2026-03-23
bug: GC_CITY env var missing from agent session, breaks gc handoff


alexsiri7:
## Problem

`gc handoff` fails with `not in session context (GC_CITY not set)` because the `GC_CITY` environment variable is absent from the agent's process environment, even though `template_resolve.go:137` sets it in the agent env map.

## Observed behavior

Mayor session (`gc-po870`, 13h old, running in tmux via systemd supervisor):

```
$ cat /proc/<claude-pid>/environ | tr '\0' '\n' | grep GC_CITY
# (empty — no GC_CITY)
```

Other GC_ vars are present (`GC_ALIAS=mayor`, `GC_SESSION_NAME=mayor`, `GC_SESSION_ID=gc-po870`, `GC_DIR`, etc.), but `GC_CITY` is missing.

The mayor attempted `gc handoff`, got the error, then retried with `GC_CITY=/home/asiri/gc gc handoff ...` which worked but left the session stuck (controller was stopped so `request-restart` blocked forever).

## Expected behavior

`GC_CITY` should be present in every agent session's environment, set by `template_resolve.go:137` and passed to tmux via `-e GC_CITY=<path>`.

## Where to look

- `cmd/gc/template_resolve.go:137` — sets `GC_CITY` in agent env map
- `internal/runtime/tmux/adapter.go:599` — passes env to `createSession`
- `internal/runtime/tmux/tmux.go:285` — `NewSessionWithCommandAndEnv` uses `tmux new-session -e`
- Session adoption path — does adoption skip env setup?
- `cmd/gc/bd_env.go:67-72` — `mergeRuntimeEnv` handles GC_CITY in env overrides

## Possible causes

1. Session was adopted from a pre-existing tmux session that didn't have the var
2. tmux `-e` flag env vars not inherited by child processes in some edge case
3. Session created by older binary that didn't set `GC_CITY` (though it's been in template_resolve since the session-first architecture)

wynged:
I am also experiencing this, idk if it is the whole story related to handoff, handoff also seems  to fail even when that var is set, but the command will simply not run on the first attempt, but then even when run again with env var set there is another failure, pane just ends up dead.  

● Bash(gc handoff "HANDOFF: Session cycling — fresh start" "HANDOFF: Mayor session cycling…)
  ⎿  Error: Exit code 1
     gc handoff: not in session context (GC_CITY not set)

[...]

PR #5864, opened 2026-09-01
fix: skip local city anchors when session env already targets a remote city

### LINKED ISSUE #5860: bug: API session env: local city anchors conflict with remote-targeted session env
## Summary

A remote-targeted `gc` nomad session fails closed with "conflicting targets" because the API server unconditionally seeds local city-identity env vars (`GC_CITY`, `GC_CITY_PATH`, `GC_CITY_RUNTIME_DIR`) into every session's env, even when the session env already carries a remote target (`GC_CITY_URL`/`GC_CITY_CONTEXT`).

## Symptom

A nomad-routed remote session whose agent/provider config sets `GC_CITY_URL` or `GC_CITY_CONTEXT` to target a remote city launches a nested `gc` process whose env also contains `GC_CITY`/`GC_CITY_PATH`/`GC_CITY_ROOT` (seeded by the API server for local city identity). The nested `gc` binary then hits its own local-vs-remote conflict guard and exits with:

```
conflicting targets: a remote city (GC_CITY_URL/GC_CITY_CONTEXT) cannot be combined with a local city env (GC_CITY/GC_CITY_PATH/GC_CITY_ROOT); unset one
```

Result: expected — a remote-targeted session starts cleanly → got: session fails closed with a spurious local/remote conflict, even though only one target was ever intentionally configured.

## Reproduction

1. Configure a nomad-routed session's provider/agent env with `GC_CITY_URL` (or `GC_CITY_CONTEXT`) pointing at a remote city.
2. Launch the session through the API server (any of the three call sites below).
3. The server composes the session env via `cityAnchoredSessionEnv`, which unconditionally merges in `GC_CITY`/`GC_CITY_PATH`/`GC_CITY_RUNTIME_DIR` for the *local* city path regardless of the remote target already present.
4. The nested `gc` binary in that session evaluates `cmd/gc/remote_target.go`'s conflict guard (`remoteVsLocalEnvErr`, checked against `GC_CITY`/`GC_CITY_PATH`/`GC_CITY_ROOT` vs `GC_CITY_URL`/`GC_CITY_CONTEXT`) and fails closed.

## Root cause

`cityAnchoredSessionEnv` (`internal/api/session_runtime.go:51-73`) merges `citylayout.CityIdentityEnvMap(cityPath)` (`internal/citylayout/runtime.go:167-177`, emits `GC_CITY`, `GC_CITY_PATH`, `GC_CITY_RUNTIME_DIR`) into the composed session env unconditionally — it never checks whether the env built so far (`processenv.ProviderProcessPassthroughEnv()` baseline + workspace env + provider/resolved env) already carries `GC_CITY_URL` or `GC_CITY_CONTEXT`.

The function has three call sites that all inherit the bug: `internal/api/session_resolution.go:330`, `internal/api/session_runtime.go:383`, `internal/api/session_runtime.go:501` (also reached via `internal/api/session_resolved_config.go:41`).

The conflict is then detected — correctly, per its own contract — by `cmd/gc/remote_target.go:80-90` (key list) and raised at `cmd/gc/remote_target.go:343-345` (`remoteVsLocalEnvErr`). That guard is not itself buggy; the bug is upstream, in the env-seeding boundary that hands the nested process a self-contradictory env.

## Fix candidates

**A. Skip local anchors when a remote target is already present** — in `cityAnchoredSessionEnv`, after composing the baseline+workspace+provider env, check whether `GC_CITY_URL` or `GC_CITY_CONTEXT` is set in that merged env; if so, omit `GC_CITY`, `GC_CITY_PATH`, and `GC_CITY_RUNTIME_DIR` from the output instead of merging `CityIdentityEnvMap`'s keys in.
- Scope: ~20 LOC in `internal/api/session_runtime.go`, plus a focused RED/GREEN test in `internal/api/session_resolved_config_test.go` (or an equivalent `_test.go` covering `cityAnchoredSessionEnv`).
- Trade-off: minimal, localized change; fixes the seeding boundary without touching the CLI's conflict-detection contract. Requires all three call sites to keep going through the shared helper (already true today).

**B. Make the CLI's conflict guard remote-target-tolerant of API-seeded local anchors** — change `cmd/gc/remote_target.go`'s guard to ignore `GC_CITY`/`GC_CITY_PATH`/`GC_CITY_ROOT` when they were seeded by the API server rather than set intentionally by a user.
- Scope: larger — requires threading a provenance marker (env var or otherwise) through the seeding path so the CLI can distinguish "user set this" from "API seeded this," touching both `internal/api` and `cmd/gc`.
- Trade-off: preserves the CLI guard's current unconditional semantics for user-set env, but adds a new provenance contract and widens the blast radius across two packages for a fix that candidate A achieves without one.

**Recommendation:** A, because it fixes the actual defect (the seeding boundary emitting a self-contradictory env) at its source, with the smallest possible diff, and leaves the CLI's local-vs-remote guard — which is behaving correctly — untouched.

## Adjacent / out of scope

- `cmd/gc/remote_target.go`'s conflict-detection logic itself is out of scope — it is doing its job correctly given the env it's handed; this issue is about not handing it a contradictory env in the first place.
- Any broader session-env provenance/audit trail improvement is a separate concern from this point-fix.

jacobhausler:
## Summary

Fixes #5860. A remote-targeted `gc` nomad session was failing closed with
`conflicting targets: a remote city (GC_CITY_URL/GC_CITY_CONTEXT) cannot be
combined with a local city env (...)` because `cityAnchoredSessionEnv`
(`internal/api/session_runtime.go`) unconditionally seeded local city-identity
anchors (`GC_CITY`, `GC_CITY_PATH`, `GC_CITY_RUNTIME_DIR`) into every session
env it composed, even when the merged env already carried `GC_CITY_URL` or
`GC_CITY_CONTEXT`.

## Root cause

`cityAnchoredSessionEnv` merged `citylayout.CityIdentityEnvMap(cityPath)` in
unconditionally, with no check for whether the baseline+workspace+provider env
already targets a remote city. The nested `gc` binary in that session then hit
`cmd/gc/remote_target.go`'s local-vs-remote conflict guard and failed closed —
correctly, given the contradictory env it was handed. This PR fixes the
seeding boundary, not the guard.

## Fix

`cityAnchoredSessionEnv` now checks the merged env for `GC_CITY_URL` /
`GC_CITY_CONTEXT` before seeding the local anchors, and omits
`GC_CITY`/`GC_CITY_PATH`/`GC_CITY_RUNTIME_DIR` when either is already set. The
existing three call sites (`session_resolution.go:330`,
`session_runtime.go:383`, `session_runtime.go:501` via
`session_resolved_config.go:41`) all route through this one helper, so the fix
covers all of them.

## Test evidence

RED/GREEN, `internal/api` package:

- RED (parent commit, before the fix): `TestCityAnchoredSessionEnvSkipsLocalAnchorsWhenRemoteTargeted` fails — anchors are seeded even with `GC_CITY_URL`/`GC_CITY_CONTEXT` present.
- GREEN (this PR): same test passes, three subtests — remote via `GC_CITY_URL`, remote via `GC_CITY_CONTEXT`, and local-only (no remote keys — anchors still seeded, no regression).
- Full `internal/api` package sweep is GREEN (verified via the project's Nomad builder pool).

## Scope

Does not touch `cmd/gc/remote_target.go`'s conflict-detection logic — that
guard is behaving correctly given the env it's handed; this PR fixes the
env-composition boundary that was handing it a self-contradictory env.

`cmd/gc/bd_env.go`'s separate `CityIdentityEnvMap` call (CLI `gc bd` env
projection, not session-spawn env) is a different code path and out of scope
for this fix.

### DIFF
--- a/internal/api/session_resolved_config_test.go
+++ b/internal/api/session_resolved_config_test.go
@@ -435,6 +435,64 @@ func TestCityAnchoredSessionEnvSkipsCityAnchorsWhenCityPathEmpty(t *testing.T) {
 	}
 }
 
+// TestCityAnchoredSessionEnvSkipsLocalAnchorsWhenRemoteTargeted is the
+// RED/GREEN regression for cr-vddb9.1: a nomad-routed remote session whose
+// provider env carries GC_CITY_URL or GC_CITY_CONTEXT must never also
+// receive the local city identity anchors (GC_CITY, GC_CITY_PATH,
+// GC_CITY_RUNTIME_DIR). Seeding both makes the nested gc binary in that
+// session fail closed at cmd/gc/remote_target.go's local-vs-remote conflict
+// guard with "conflicting targets" (matches the D2 failure in wisp
+// gcg--9223372036854774627 / session gcg--9223372036854774630 gating
+// cr-u4plc.1).
+func TestCityAnchoredSessionEnvSkipsLocalAnchorsWhenRemoteTargeted(t *testing.T) {
+	cityPath := t.TempDir()
+
+	t.Run("remote via GC_CITY_URL", func(t *testing.T) {
+		providerEnv := map[string]string{
+			"GC_CITY_URL":    "https://remote.example.test",
+			"PROVIDER_TOKEN": "ok",
+		}
+		got := cityAnchoredSessionEnv(cityPath, nil, providerEnv)
+		for _, key := range []string{"GC_CITY", "GC_CITY_PATH", "GC_CITY_RUNTIME_DIR"} {
+			if v, ok := got[key]; ok {
+				t.Errorf("%s = %q present, want absent when GC_CITY_URL targets a remote city", key, v)
+			}
+		}
+		if got["GC_CITY_URL"] != "https://remote.example.test" {
+			t.Errorf("GC_CITY_URL = %q, want preserved", got["GC_CITY_URL"])
+		}
+		if got["PROVIDER_TOKEN"] != "ok" {
+			t.Errorf("PROVIDER_TOKEN = %q, want ok", got["PROVIDER_TOKEN"])
+		}
+	})
+
+	t.Run("remote via GC_CITY_CONTEXT", func(t *testing.T) {
+		providerEnv := map[string]string{
+			"GC_CITY_CONTEXT": "remote-context",
+		}
+		got := cityAnchoredSessionEnv(cityPath, nil, providerEnv)
+		for _, key := range []string{"GC_CITY", "GC_CITY_PATH", "GC_CITY_RUNTIME_DIR"} {
+			if v, ok := got[key]; ok {
+				t.Errorf("%s = %q present, want absent when GC_CITY_CONTEXT targets a remote city", key, v)
+			}
+		}
+	})
+
+	t.Run("local — no remote keys present", func(t *testing.T) {
+		got := cityAnchoredSessionEnv(cityPath, nil, map[string]string{"PROVIDER_TOKEN": "ok"})
+		if got["GC_CITY"] != cityPath {
+			t.Errorf("GC_CITY = %q, want %q (anchors still seeded when no remote target present)", got["GC_CITY"], cityPath)
+		}
+		if got["GC_CITY_PATH"] != cityPath {
+			t.Errorf("GC_CITY_PATH = %q, want %q", got["GC_CITY_PATH"], cityPath)
+		}
+		wantRuntimeDir := filepath.Join(cityPath, ".gc", "runtime")
+		if got["GC_CITY_RUNTIME_DIR"] != wantRuntimeDir {
+			t.Errorf("GC_CITY_RUNTIME_DIR = %q, want %q", got["GC_CITY_RUNTIME_DIR"], wantRuntimeDir)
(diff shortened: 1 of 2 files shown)


quad341 (approved):
@jacobhausler — this is excellent diagnosis. Tracing "conflicting targets" back to `cityAnchoredSessionEnv` instead of the guard in `remote_target.go`, and then deliberately *not* touching the guard, is exactly right: the guard was doing its job on an env that should never have been composed that way. Restructuring to merge first and decide anchors second is what makes the check possible at all, and your local-only subtest is the thing that keeps upstream #101 from creeping back.

One integration detail I'll absorb on my side. This session env is an overlay on inherited environment — on the tmux path an empty value becomes `env -u`, but a *missing* key is simply inherited from the tmux server's global env. So omitting the anchors may not clear a `GC_CITY` the controller already exported. There's a convention for this a few lines up in the same file (the `GC_CONTROLLER_TOKEN` pin: withheld means present and empty). I'll switch to pin-empty, add `GC_CITY_ROOT` since the guard trips on that too, and retarget the two remote subtests. Your commit, your credit — merging with that on top. Thanks for this one.


---

PR #5868, opened 2026-09-01
fix(cmd/gc): thread a live events.Recorder into the CLI worker.Factory

### LINKED ISSUE #5859: cmd/gc CLI worker.Factory never sets Recorder — worker.operation silently never emitted for CLI-driven session lifecycle
## Problem

`gc analyze reliability` reports every CLI-driven session lifecycle event as
dropped: `506 lifecycle event(s) skipped: no worker.operation observed for
the session at or before the event (instrumentation gap)`. Verified
reproducing on current `main` (`cfea9840eeedd643e8c12a37239a264eda1ce7d0`,
2026-09-01) — this is a live code gap, not a pin-lineage or config-flag issue.

## Root cause

`worker.operation` recording is gated by
`(*SessionHandle).operationEventRecordingEnabled()`:

```go
// internal/worker/operation_events.go:251-253
func (h *SessionHandle) operationEventRecordingEnabled() bool {
	return h != nil && h.recorder != nil && h.recorder != events.Discard
}
```

`h.recorder` comes from `worker.FactoryConfig.Recorder`, which defaults to
`events.Discard` whenever unset (`internal/worker/handle_construct.go:55-57`):

```go
recorder := cfg.Recorder
if recorder == nil {
	recorder = events.Discard
}
```

There are exactly two `worker.Factory` construction sites in the codebase:

1. **`internal/api/worker_factory.go:16-27`** (`(*Server).workerFactory`,
   the HTTP/dashboard-facing API server path) — wires a live recorder:
   `Recorder: s.state.EventProvider()`.
2. **`cmd/gc/worker_handle.go:79-90`**
   (`workerFactoryWithStaleKeyDetectionWaiter`, the *only*
   `worker.Factory` constructor `cmd/gc` uses) — builds
   `worker.FactoryConfig{...}` with **no `Recorder` field set at all**.
   Every CLI-driven session-lifecycle helper in `cmd/gc` funnels through
   this factory (`workerHandleForSessionWithConfig`,
   `workerHandleForSessionTargetWithConfig`,
   `workerKillSessionTargetWithConfig`,
   `workerStopSessionTargetWithConfig`,
   `workerInterruptSessionTargetWithConfig`,
   `workerRespondSessionTargetWithConfig`, plus the session-catalog/
   session-new paths in `cmd/gc/cmd_session.go`).

The comment already at `operation_events.go:256-262` names this precisely:
usage facts still flow on CLI-factory handles because they get a sink but
"no event recorder (the CLI factory path)" — but it means CLI-driven
session lifecycle (start/attach/create/stop/kill/interrupt/message/nudge/
respond/close/rename) silently never emits `worker.operation` for any
session managed via the `gc` CLI/controller surface rather than the HTTP
API server.

A live `events.Provider` is already available in-process to the CLI/
controller code — `controllerState.EventProvider()`
(`cmd/gc/api_state.go:1475`), already used elsewhere in `cmd/gc` (e.g.
`cmd/gc/route_recovery.go:59-63`, `cmd/gc/city_runtime.go:3646`) — it is
simply never threaded into `worker.FactoryConfig.Recorder` at the CLI
factory construction site.

## Impact

Any city that runs its session lifecycle through the `gc` CLI/controller
surface rather than the HTTP API server (e.g. supervisor-managed pool/
manager/worker sessions) never gets `worker.operation` events for that
lifecycle, so `gc analyze reliability` skips ~100% of lifecycle events with
"instrumentation gap," and any tooling depending on `worker.operation` for
CLI-driven sessions is blind. This likely also underlies part of the
symptom in #3928 (session lifecycle events missing from the event stream),
though that issue is scoped to `session.*` event types generally and to
`start-pending` wedge detection — this issue is the specific root cause for
the `worker.operation` non-emission piece.

## Fix candidates

1. Thread a live `events.Recorder` (sourced from
   `controllerState.EventProvider()` or equivalent) into
   `worker.FactoryConfig.Recorder` at `cmd/gc/worker_handle.go:79-90`
   (`workerFactoryWithStaleKeyDetectionWaiter`), mirroring
   `internal/api/worker_factory.go:21`. This is a call-site fix — the
   emitter and gating logic are already correct; only the CLI factory's
   struct literal needs the field set.
2. Default `worker.FactoryConfig.Recorder` (or `NewFactory`/
   `handle_construct.go`) to a shared process-wide recorder instead of
   `events.Discard` when no explicit recorder is supplied, so any future
   factory construction site gets live recording by default rather than
   requiring every caller to remember to wire it.

Candidate 1 is the more surgical fix and matches the existing pattern at
the API server call site; candidate 2 closes the gap for any *future*
factory construction site as well but changes default behavior more
broadly.

## Repro

```
$ gc analyze reliability
506 lifecycle event(s) dropped before grouping (skipped + ambiguous_aliases).
506 lifecycle event(s) skipped: no worker.operation observed for the session
at or before the event (instrumentation gap).
```

jacobhausler:
## What this changes

`worker.FactoryConfig.Recorder` was left unset at the CLI/controller factory
construction site (`cmd/gc/worker_handle.go`,
`workerFactoryWithStaleKeyDetectionWaiter`), so every CLI-driven session
lifecycle helper built a `worker.SessionHandle` with the default
`events.Discard` recorder and `worker.operation` was never emitted for
sessions managed through the `gc` CLI/controller surface — only the HTTP API
server path (`internal/api/worker_factory.go:21`) wired a live recorder. This
made `gc analyze reliability` skip ~100% of CLI-driven lifecycle events with
"instrumentation gap".

`worker.Factory` is constructed on every session-reconciliation tick as well
as per CLI invocation, so wiring a fresh `events.FileRecorder` in on every
call would leak a file handle and spawn a rotation goroutine each tick.
`cliFactoryEventsRecorder` (`cmd/gc/providers.go`) memoizes one live recorder
per city path for the process lifetime instead, giving the CLI factory path
the same live recording the API server already gets for free from its
long-lived `controllerState.EventProvider()`.

Fixes #5859

## Review notes

- Change is confined to `cmd/gc/worker_handle.go` (wire the recorder) and
  `cmd/gc/providers.go` (resolve + memoize it).
- Two existing tests (`TestWorkerHandleForSessionTargetWithConfigResolvesSessionName`,
  a Kill test in `worker_handle_test.go`) asserted `Stop` was the last runtime
  call. With a live recorder, `Kill`'s `worker.operation` event now finishes
  for real and re-enriches session identity via `EnrichInfo`, which issues a
  trailing `IsRunning` probe — so `Stop` is no longer guaranteed to be last.
  Updated both to locate the `Stop` call instead of indexing the tail.
- Falls back to `events.Discard` when `cityPath` is empty or the provider
  cannot be opened — telemetry must never block session lifecycle.

## Test plan

- [x] `go vet ./cmd/gc/...`
- [x] `go test ./cmd/gc/...` (TestWorker* suite)
- [x] `go test ./internal/worker/...`
- [x] `go test ./internal/api/...` (pre-existing, unrelated `TestEveryEmittedErrorCodeIsRegistered` failure reproduces identically on `main`; not caused by this change)


quad341 (approved):
@jacobhausler — this is a genuinely good catch, and I verified it end to end. `worker.FactoryConfig.Recorder` has no `Discard` default the way `UsageSink` does, so the unset field meant `operationEventRecordingEnabled()` suppressed every `worker.operation` on the CLI path while `internal/api/worker_factory.go` quietly got a live one. That's exactly the kind of asymmetry that hides for months behind an "instrumentation gap" line in `analyze reliability`.

Two things I especially appreciated: memoizing per city path instead of opening a recorder per reconciler tick — you clearly read the controller loop before writing the fix — and the typed-nil-boxing fix in `newEventsProviderForNameWithConfig`, which nobody asked for, is a real latent bug, and comes with a test that pins it properly.

I'm taking the integration work on my side. Your recorder inherits rotation and the startup sweep, and the controller already runs a long-lived rotating recorder on that same file, so I'll construct yours as a secondary writer (`WithMaxSize(0)` + `WithoutStartupSweep()`, matching `newClassStoreEmitRecorder`), stop caching `Discard` after a transient open failure, and add the assertion that a `worker.operation` actually lands in `events.jsonl`. The red CI is our known `productmetrics` flake (#4653), not you — I'll re-run it.

Merging with credit. Thanks for chasing this into the factory seam.


---

PR #5871, opened 2026-09-01
fix(doltpool): bound idle connections below the server's wait_timeout

Rome-1:
## What

Adds `db.SetConnMaxIdleTime(20 * time.Second)` to the shared Dolt pool.

## Why

`internal/doltpool` sets `maxIdleConns = 2` and `connMaxLifetime = time.Hour`, but no idle timeout. An idle pooled connection therefore lives up to an hour, while gc's own managed Dolt config sets the server's `wait_timeout` to **30 seconds**. The server reaps the connection first and logs an error for each one:

```
Error reading packet from client N: read tcp ... i/o timeout
io.ReadFull(header size) failed
```

On one managed city this produced **6565 ERROR lines in a single day**, continuous from 00:00 to 17:22. The timestamps land on seconds `:02`, `:28`, `:32`, `:58` — two interleaved series on a 30s cycle, matching `wait_timeout` exactly. `ss` attribution showed every connection belonged to `gc`, with the churn concentrated on the supervisor's reconcile loop.

Nothing is failing here: this is the reaper working as designed, logged at ERROR. The costs are that a log crying wolf thousands of times a day trains operators to discount it, and the server burns CPU managing sockets the client should have closed itself.

## Why this value

`SetConnMaxIdleTime` must sit **strictly below** `wait_timeout` so the client closes idle connections first, removing the events at the source rather than muting them. 20s leaves a 10s margin; equality with `wait_timeout` would be a race, not a bound.

`connMaxLifetime` is deliberately left at an hour — it bounds connection *age*, not idleness, and age is not what the server is reaping.

This follows the lineage of the package itself, which was introduced to eliminate per-operation Open+Close churn (its own comment cites 2,618 TIME_WAIT sockets from one call site). This is the same class of problem one level down: connections that outlive what the other end expects.

## Not included

The log **severity** is not addressed here. `Error reading packet from client` is emitted by Dolt's go-mysql-server, so it is not this project's to change; it would be a separate issue upstream. Bundling the two would stall the half that can actually land.

## Test

`go build`, `go vet`, and `go test ./internal/doltpool/` all pass.

### DIFF
--- a/internal/doltpool/pool.go
+++ b/internal/doltpool/pool.go
@@ -31,6 +31,20 @@ const (
 	maxOpenConns    = 5
 	maxIdleConns    = 2
 	connMaxLifetime = time.Hour
+	// connMaxIdleTime must stay STRICTLY BELOW the server's wait_timeout,
+	// which gc's managed dolt config sets to 30s. Without it an idle pooled
+	// connection lives until connMaxLifetime (an hour), so the SERVER reaps
+	// it first and logs an error per reap:
+	//
+	//   Error reading packet from client N: read tcp ... i/o timeout
+	//   io.ReadFull(header size) failed
+	//
+	// 6565 such lines in one day were observed on a single managed city,
+	// arriving on a clean 30s cycle — the reaper working, logged at ERROR.
+	// Closing idle connections client-side first removes the events at the
+	// source rather than muting them. 20s leaves a 10s margin; equality
+	// with wait_timeout would be a race, not a bound.
+	connMaxIdleTime = 20 * time.Second
 	connTimeout     = 5 * time.Second
 	readTimeout     = 30 * time.Second
 	writeTimeout    = 30 * time.Second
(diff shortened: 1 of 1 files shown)


Rome-1:
The upstream half of this is now filed as dolthub/dolt#11658 — the reap events this PR removes at the pool are logged at ERROR by `dolthub/vitess` `go/mysql/conn.go`, which already suppresses two other "client went away" shapes but not a read-deadline timeout.

The two fixes are independent and worth having both: this one stops our pool from creating the events, and the upstream one stops a by-design reap from being logged as a server error for everyone else.

Also related, found while getting this PR green after a module-download flake: #5882.

quad341 (approved):
@Rome-1 — this is genuinely good bug-hunting. Attributing 6565 ERROR lines/day to a pool-vs-server timeout mismatch, pinning it with `ss` down to the interleaved 30s series, and then drawing the line at *"closing client-side removes the events at the source rather than muting them"* is exactly the right instinct. Splitting the log-severity half out upstream instead of bundling it was the right call too — it's why this half can land now.

One correction to fold in on our side, and it's our documentation's fault for being buried in `internal/config`: we measured `wait_timeout` as **inert** on dolt 2.2.3 back in #5383 — the server accepts, stores and reports it, but nothing reads it. The actual reaper is `read_timeout_millis`, default 120s. Your 20s sits below that *and* below the 30s value cities emit, so the fix is correct as written; only the comment's mechanism needs adjusting. (The 30s period you measured isn't explained by either default — I suspect that city carries a lower `read_timeout_millis`, but I can't confirm it from here, so I'll leave it noted rather than asserted.)

I'll rewrite that paragraph, add a small test pinning the constant below both bounds so it can't drift back, then merge. Your commit, your credit. Thanks for this one.


---

Issue #5852, opened 2026-09-01
feat: gc analyze: add bead-throughput, convoy/gate-latency, session-outcome, and stall-detection subcommands


jacobhausler:
## Summary

`gc analyze` (`cmd/gc/cmd_analyze.go`) currently ships exactly one subcommand,
`reliability` (`cmd/gc/cmd_analyze_reliability.go`, from #1254 / the evals
workstream in #1184). Four other recurring operator questions over the same
`events.jsonl` + bead state are currently answered by hand-grepping the event
log instead of a `gc analyze` subcommand. This issue proposes four new
subcommands to close that gap.

## Symptom

Operating a Gas City deployment, the same four questions come up repeatedly
and each is currently answered with ad hoc `grep`/`jq` over
`.gc/events.jsonl` and `bd`/`bead` queries rather than a first-class command:

1. Bead open/close counts by time window, grouped by store/type/label — the
   routine "how much did we ship today" stats query.
2. Chain/convoy latency: claim-wait per pool (used for diagnosing P0-pool
   starvation), the `execution.step_defined` → `execution.step_started` gap
   (gate queue-wait), and gate bounce rate per formula.
3. Session start outcomes by template/provider over time — e.g. spotting a
   provider outage where every session-start attempt exits uniformly fast
   with an error, which today only shows up if someone happens to eyeball
   `worker.operation` events for a suspiciously tight, uniform
   `duration_ms` + `result != "success"` cluster.
4. Stall detection: last-event age per in-progress bead / per pool — the
   "is the dispatcher wedged" signature, i.e. beads with no event newer than
   N minutes while `status=in_progress`.

(Motivation note: this was originally reported by the deployment owner as
motivated by "the executive's overnight hand-grep of events.jsonl" from
2026-08-31/09-01. I searched the reporting deployment's `ops/receipts/` and
`_triage/` directories for that session transcript and did not find one
preserved on disk — the requirements above are restated from the assigning
task record, not quoted from a saved transcript.)

## Reproduction

```
gc analyze --help
```

Result: only `reliability` is listed under Available Commands. There is no
`gc analyze` subcommand for bead throughput counts, convoy/gate latency,
session-start outcome breakdowns, or stall detection — an operator has to
drop to `jq`/`grep` over `.gc/events.jsonl` (or the `gc bd list`/`gc bd show`
surface) for all four.

## Root cause

[...]

PR #5876, opened 2026-09-01
feat(analyze): add gc analyze sessions subcommand

jacobhausler:
## Summary

Adds `gc analyze sessions`, the third of the four `gc analyze` subcommands proposed in #5852 (session start outcomes by template/provider over time). The first, `gc analyze beads`, shipped in #5865.

- Reports `worker.operation` session-start attempts (`start` / `start_resolved`), bucketed by time and grouped by template/provider, with succeeded/failed/other counts and `duration_ms` statistics per bucket.
- Flags a bucket as a **possible provider outage** when every attempt in it failed and `duration_ms` is tightly clustered (≤500ms spread, ≥3 attempts) — the signature of every session-start attempt dying at the same broken stage rather than failing independently.
- Follows the same split established by #5865: `internal/sessionoutcomes` (pure-data analysis layer) + `cmd/gc/cmd_analyze_sessions.go` (CLI/IO) + `schemas/analyze/sessions/result.schema.json`.
- Refactors `resolveEventsPath` to take explicit params instead of `reliabilityCmdOptions`, so it can be shared by both the `reliability` and `sessions` subcommands without an inter-file type dependency.

## Test plan

- [x] Compiles cleanly (`cmd/gc` package)
- [x] `go vet` on `internal/sessionoutcomes` and `cmd/gc` — clean
- [x] `gofmt -l` clean on all changed files
- [x] Package tests for `internal/sessionoutcomes` — all pass
- [x] `TestRunAnalyzeSessions*` in `cmd/gc` — all pass
- [x] `TestResolveEventsPath*` in `cmd/gc` — all pass (verifies the refactor didn't break `gc analyze reliability`)
- [x] `schemas/analyze/sessions/result.schema.json` is valid JSON
- [x] Confirmed the full `cmd/gc` package test suite has pre-existing environment-dependent flakiness on plain `main` (different tests fail each run — process-group timeout tests, storage registry tests — unrelated to this change) independent of this branch


jacobhausler:
Rebased onto upstream/main e589fdea3 and re-greened: CI is now 63 pass / 0 fail on head 4d47a70cd (was 59 pass / 4 fail). Three things were holding it red, all PR-side.

1. Missing census row. gc analyze sessions registered a new live command with nothing in cmd/gc/productmetrics_command_census.json listing it, so validateProductMetricsCommandCensus failed, applyProductionProductMetricsCommandCensus returned before stamping the census-valid annotation, and classifyProductMetricsCommand failed closed for every row of TestClassifyProductMetricsCommandCanonicalMatrix — the whole matrix, not just the new command. Added the row (id 207, next_id 207 -> 208; id density is enforced by ValidateManifest) and regenerated the derived artifacts with `go run ./cmd/gen-command-census`: cmd/gc/metrics_census_gen.go, internal/productmetrics/command_ids_gen.go, schemas/metrics/example/result.schema.json. `go run ./cmd/gen-command-census -check` passes, so the generated-artifacts leg stays green.

2. unparam at cmd_analyze_sessions_test.go:15. sessionStartEvent took an operation parameter every call site passed "start" to. Dropped the parameter and inlined const sessionStartOperation = "start", following cmd_formula_version_check_test.go:18. start_resolved stays covered where the parsing lives (internal/sessionoutcomes/sessionoutcomes_test.go:51), so nothing loses coverage. Preflight / static checks now passes.

3. internal/sessionoutcomes was missing its mandatory testenv_import_test.go, which TestRequiresDedicatedTestenvImportFile enforces (every real test dir must import internal/testenv from an untagged file so the GC_* env scrub runs first). Generated with `go run scripts/add-testenv-import.go`. This was the Integration/packages-core-4-of-4 failure.

Verification: TestClassifyProductMetricsCommandCanonicalMatrix runs in "cmd/gc process / shard 9 of 12" and passed there ("observable go test: PASS"); the shard enumerates with `go test -list '^Test'` and shards by index modulo 12 with no exclusions, so it is not being skipped. `go test ./internal/sessionoutcomes/...` is green locally too. Two gates could not be run by the seat that pushed this (a 1 GiB alloc cannot compile package cmd/gc, with or without -p/-gcflags): make lint-affected and scripts/check-generated-docs-drift.sh. Both are green in CI on this head; flagging them explicitly so nobody assumes they were run locally. docs/reference/cli.md was regenerated by hand against the same generator's output format.

Unrelated pre-existing failure, for the record: internal/productmetrics TestMissingInflightSettlementRequiresDurableQueueDisposition/unreadable_file fails when the test runs as uid 0 — the case writes the event file with mode 0o000 and expects the read to fail, which root defeats. It reproduces identically on pristine upstream/main e589fdea3, and this PR touches only the generated ID ledger in that package.

PR #5879, opened 2026-09-01
fix(prompt): apply filesystem search guidance globally

### LINKED ISSUE #5873: bug: filesystem-search warning misses custom agent prompts
## Before you continue

## Gas City version

`1.4.0-908-g5ab1031ef` (`5ab1031ef`); the prompt-coverage gap is also present on `main` at `615f5b794`.

## Environment

macOS 15.4.1, zsh, tmux runtime provider, Claude Code child sessions.

## Reproduction

1. Configure an agent with its own `prompt_template` that does not include the Gas Town `following-mol` fragment, such as a city-owned worker or a pack-defined coordinator.
2. Run `gc prime <agent-name>` or start the managed session.
3. Inspect the delivered prompt. It does not include the warning against `find /`, `find ~`, `find /Users`, or `find $HOME`.
4. Compare with a graph worker, pool worker, or a role using `following-mol`; those selected prompt paths do contain the warning added by #1785 and #1860.
5. On restart, an uncovered or resumed Claude Code session can improvise a broad discovery search. macOS then asks the user for Desktop, Downloads, Music/Media, removable-volume, or network-volume access.

## Expected behavior

Every non-empty prompt Gas City delivers to an agent—custom template, built-in worker prompt, fallback prompt, or managed startup payload—contains the filesystem-scope safeguard exactly once. Searches that are genuinely necessary are constrained to the known city, rig, or worktree root.

## Actual behavior

The existing fixes are attached only to selected prompt sources:

- #1785 updated the maintenance dog prompt.
- #1860 updated `following-mol` plus graph-worker and pool-worker prompts.

Agents using other templates do not inherit those instructions. Gas City has fragment composition, but no universal base prompt or delivery-boundary safeguard.

## Logs, screenshots, or traces

```shell
# Representative rendered-prompt result before the fix:
gc prime <custom-agent>
# Output contains the custom prompt but no "Never use wide filesystem searches" safeguard.

# Representative covered result:
gc prime <graph-or-pool-worker>
# Output includes the safeguard from #1860.
```

macOS TCC records identified `com.anthropic.claude-code` as the accessing child process and `com.gascity.gc` as the responsible parent. A bounded source/configuration audit found correct city-relative working directories and no configured Desktop, Downloads, Music, `/Volumes`, home-root, or `/Users` search path.

## Additional context

This is a coverage follow-up to #1785 and #1860, not a replacement for them. The proposed fix applies one generic, role-neutral filesystem-boundary paragraph at the two final prompt-delivery boundaries and deduplicates prompts that already contain the warning. Suppressed/empty startup prompts remain empty.

benjaminbrumbaugh:
## Summary

- Apply the macOS filesystem-search safeguard at the shared startup-prompt delivery boundary instead of only in selected bundled role templates.
- Apply the same safeguard to every non-empty `gc prime` prompt, including provider-formatted hook context, while preserving managed-prompt suppression.
- Reuse the existing heading as an idempotence marker so bundled graph/pool/following-molecule prompts do not receive duplicate guidance.

Fixes #5873.

## Testing

- [ ] `make check`
  - Full `make check` is blocked by an unrelated formatting defect already present on `upstream/main` in `test/acceptance/worker_inference/worker_inference_test.go`.
  - Not applicable; no documentation paths changed.
  - Not applicable; this changes prompt composition, not runtime/controller/workflow behavior.
- [x] Focused startup, prime, hook-format, suppression, and existing filesystem-guidance tests pass.
- [x] `make lint`
- [x] `go vet ./...`
- [x] `make build`
- [x] `make fmt-check-changed CI_STATIC_SELECT=./scripts/ci-static-select LINT_CHANGED_SCOPE=staged`

`make test-fast-parallel` reaches the repository's pre-existing path-with-spaces failure in `TestChangedStaticTargetsScopeLintAndFormattingToTheDiff`: its generated Makefile command does not quote the absolute `scripts/ci-static-select` path. The changed `cmd/gc` tests pass independently.

## Checklist

- [x] Updated docs for user-facing changes — not required; the safeguard is emitted runtime guidance, with behavior covered by tests

### DIFF
--- a/cmd/gc/cmd_prime.go
+++ b/cmd/gc/cmd_prime.go
@@ -614,6 +614,7 @@ func primeHookHasLiveManagedSession(cityPath string) bool {
 }
 
 func writePrimePromptWithFormat(stdout io.Writer, cityName, agentName, prompt string, hookMode bool, hookFormat string, suppressPrompt bool, hookContextSuffix string, afterDelivery func()) {
+	prompt = appendFilesystemSearchGuidance(prompt)
 	if hookMode && suppressPrompt {
 		// Managed sessions receive the rendered startup prompt through the
 		// launch payload or nudge path. SessionStart hooks add context only.
--- a/cmd/gc/cmd_prime_test.go
+++ b/cmd/gc/cmd_prime_test.go
@@ -175,8 +175,9 @@ schema = 2
 	if code != 0 {
 		t.Fatalf("doPrime() = %d, want 0; stderr=%q", code, stderr.String())
 	}
-	if got := stdout.String(); got != "Agent: ada\n" {
-		t.Fatalf("stdout = %q, want %q", got, "Agent: ada\n")
+	want := appendFilesystemSearchGuidance("Agent: ada\n")
+	if got := stdout.String(); got != want {
+		t.Fatalf("stdout = %q, want %q", got, want)
 	}
 }
 
@@ -285,7 +286,7 @@ path = "./rigs/bravo"
 	if code != 0 {
 		t.Fatalf("doPrime() = %d, want 0; stderr=%q", code, stderr.String())
 	}
-	if got := stdout.String(); got != "alpha-work-query" {
+	if got := stdout.String(); got != appendFilesystemSearchGuidance("alpha-work-query") {
 		t.Fatalf("stdout = %q, want alpha rig fragment; stderr=%q", got, stderr.String())
 	}
 	if strings.Contains(stdout.String(), "bravo-work-query") {
--- a/cmd/gc/main_test.go
+++ b/cmd/gc/main_test.go
@@ -6127,8 +6127,9 @@ prompt_template = "prompts/mayor.md"
 	if code != 0 {
 		t.Fatalf("doPrime = %d, want 0; stderr: %s", code, stderr.String())
 	}
-	if stdout.String() != promptContent {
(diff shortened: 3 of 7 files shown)


quad341 (dismissed):
@benjaminbrumbaugh — this is a well-aimed fix. Moving the safeguard to the two delivery boundaries instead of adding it to yet another template is the right structural call, and reusing the existing heading as the idempotence marker is the detail that makes it safe to ship: I grepped the tree and both bundled prompts carry that heading byte-for-byte, so nobody gets it twice. I especially appreciated that you left `suppress_startup_prompt` and the empty-prompt path intact rather than blanketing everything — that was the easy thing to get wrong here.

On the red check: that's our `pr-evidence-watchdog` failing closed because no suite ever ran for your head commit, not a failing test on your patch. Your patch is fine; our pipeline didn't evaluate it.

I'm taking two small things on my side and merging with your authorship intact: assertions that the two suppression paths stay guidance-free, and getting the suite actually dispatched. Nothing needed from you.

I'll also file a follow-up, not for you to fix: the Go copy and the two pack copies have drifted slightly in wording and should be unified.

Thanks for #5873 with the TCC trace — the repro made this easy to review.


---

quad341 (approved):
Re-reviewed `67b4f84c4`..`3f2bf459d` (19 new commits since the last review): no new concerns — the prior review stands.

Issue #3897, opened 2026-07-03
foreign/stale supervisor on default port 8372 blocks the real one: start gives no diagnosis, status claims running via unverified API, crash-loop grew supervisor.log to 645MB


csells:
# Foreign supervisor on default port silently blocks the real one; status lies, log grows unboundedly

## Incident (macOS, 2026-07-01 → 2026-07-03)

A leaked launchd supervisor agent (see gastownhall/gascity#3896) ran `gc supervisor run` (v1.3.2) with
`GC_HOME` pointing at a deleted temp dir. Because its `supervisor.toml` was gone,
`Section.PortOrDefault()` (internal/supervisor/config.go) fell back to **8372** — the same
default the real supervisor uses. Its control socket lived under the deleted temp
`GC_HOME`, so only the TCP port collided.

Consequences, all observed:

1. **Real supervisor crash-looped for ~2 days.** Every start logged
   `api: listen 127.0.0.1:8372 failed: bind: address already in use` and exited; launchd
   `KeepAlive` relaunched it immediately. `~/.gc/supervisor.log` grew to **645MB** of the
   same two lines (no backoff, no rotation, no give-up).
2. **`gc supervisor start` gave no diagnosis.** Output was only
   `supervisor did not become ready; see ~/.gc/supervisor.log`. It did not surface the bind
   error, the owning PID, or the fact that the listener was a *different* supervisor.
3. **`gc supervisor status` reported healthy.** It printed
   `Supervisor is running (pid unavailable: control socket unreachable; liveness confirmed
   via api)`. The `supervisorAPIReachable()` fallback (cmd/gc/cmd_supervisor.go, gascity#2984
   path) only checks that *something* answers on the configured port — it never verifies the
   responder's identity (GC_HOME / instance ID), so a foreign or stale supervisor passes as
   "running".

## Suggested fixes

- **Identity-check the API fallback**: the supervisor API should expose its GC_HOME (or an
  instance UUID persisted under GC_HOME) and `status`/`start` readiness probes should verify
  it matches before claiming liveness. Mismatch is its own diagnostic: "port 8372 is served
  by a different supervisor (GC_HOME=...)".
- **Diagnose bind conflicts in `gc supervisor start`**: on `address already in use`, report
  the owning PID/command (lsof-equivalent) instead of pointing at the log.
- **Bound the crash loop**: on a persistent bind failure the supervisor exits non-zero and
[...]

PR #5881, opened 2026-09-01
feat(doctor): warn when the managed Dolt server log grows unbounded

Rome-1:
## Problem

`dolt.log` is opened `O_APPEND` at every managed server start and is never truncated or rotated.

Nothing in the repo bounds it:

- The rotation in `internal/events` (`WithMaxSize`, `ForceRotate`) is wired to `.gc/events.jsonl` only.
- No maintenance order or restart path trims it — `dolt_start_managed.go` and the legacy shell fallback both append, and stop/restart only `tail`/`wc -c` the file for offset tracking.
- No doctor check references `dolt.log` at all. `EventLogSizeCheck` covers `events.jsonl`, `BdBackupSizeCheck` covers `.beads/backup/`, `DoltNomsSizeCheck` and `DoltJournalSizeCheck` cover the data directory. The server log has no canary.

It also grows faster than the absence of a bound suggests. The server reaps abandoned client connections at the `wait_timeout` boundary and logs one ERROR per reap, which on one managed city came to **21 MB in a single day with nothing failing** — the reaper working as designed, logged at error severity. (#5871 removes that particular source at the pool; this PR bounds the file regardless of what writes to it.)

This class of unbounded growth has already caused an incident here: #3897, where `supervisor.log` reached 645 MB during a crash loop. That fix was never generalized to the server log.

## What this adds

A `dolt-log-size` doctor check, following the `dolt-journal-size` template:

- Same managed-Dolt applicability gate, so file-backed and external-endpoint workspaces are not warned about a file they do not have.
- Same env-overridable thresholds — `GC_DOLT_LOG_WARN_BYTES` (default 512 MB) and `GC_DOLT_LOG_ERROR_BYTES` (default 2 GB). An invalid value falls back to the default rather than silently disabling the canary, and there is a test for that.
- Resolves the log path the same way `cmd/gc` does when it starts the server: `GC_DOLT_LOG_FILE` wins, otherwise `dolt.log` in the pack state directory. One test deliberately exercises the default path rather than the env override, so a layout change cannot quietly make every other case measure a file the server never writes.

`doltJournalThreshold` is renamed `doltEnvByteThreshold` now that a second check reads byte thresholds from the environment through it. No behavior change.

## Why a canary and not rotation

The repo has already taken a position on unbounded files: surface the size, let the operator pick the cleanup. `BdBackupSizeCheck` says so in its own `CanFix` comment.

`CanFix` is false here for a further reason. Truncating in place is safe against a live `O_APPEND` writer, but discarding the log destroys the only record of why the server behaved as it did, and whether to archive first is operator policy. The check reports the size and the path; the operator picks the recipe.

Happy to follow up with real rotation instead if you would rather have that — it is a larger change and a different review.

## Testing

- 10 new tests in `internal/doctor/checks_dolt_log_size_test.go` covering the skip flag, the applicability gate, a missing log, the default-path resolver, warn and error thresholds, env overrides, an invalid env value, a non-regular file, and `CanFix`.
- `go vet ./internal/doctor/ ./cmd/gc/` clean.
- Full `internal/doctor` package suite green.


Rome-1:
Context for review: this is the second half of the `dolt.log` problem in #5871. That PR stops our connection pool from generating the reap events; this one bounds the file regardless of what writes to it, since nothing rotates it today.

The upstream severity issue behind those events is dolthub/dolt#11658.

Rome-1:
Force-pushed a fix for the two golden-file failures on the first run — both were mine and both were real, not flakes.

- `doctor_check_names.golden` needed `dolt-log-size` in registration order.
- `gc_env_read_baseline.golden` needed `GC_DOLT_LOG_FILE`. `cmd/gc` already read that variable; `internal/doctor` now reads it too, so the check resolves the same path the server actually writes to rather than assuming the default layout.

I had run `./internal/doctor` and `go vet` but not `./cmd/gc` or `./internal/testenv`, which is why CI caught these and I did not. All three packages pass locally now.

quad341 (dismissed):
@Rome-1 — thanks for chasing this one all the way through the runtime layout rather than guessing at it. Building on the `dolt-journal-size` template was the right instinct: the same applicability gate means file-backed and external-endpoint workspaces don't get warned about a file they don't have, and folding `doltJournalThreshold` into `doltEnvByteThreshold` was the correct move once a second check needed it.

Two things I want to call out specifically. `TestDoltLogSizeCheck_DefaultPath` deliberately exercising the resolver instead of the `GC_DOLT_LOG_FILE` override is the test most people skip, and it's the one that stops a future layout change from quietly making every other case measure a file the server never writes. And your reasoning for a canary over rotation — leaning on the `BdBackupSizeCheck` precedent — is the right read of where this repo stands.

One small thing I'll take on my side: the remediation hint interpolates the log path into `: > <path>` unquoted, so a path with a space wouldn't survive a copy-paste. I'll run it through `shellquote.Join`, add a regression test, re-run vet and the doctor/cmd-gc suites, and merge. Your commit and credit stay intact.

Separately, and purely as a follow-up question rather than anything for this PR: `resolveManagedDoltLogFile` honors ambient `GC_DOLT_LOG_FILE`, while `resolveManagedDoltDataDir` and `resolveManagedDoltConfigPath` deliberately ignore ambient `GC_DOLT_*` so doctor always measures the city it was pointed at. Your rationale is sound — it's worth a maintainer ruling on which is the house rule.


---


quad341 (approved):
Re-reviewed `b0a91f6a2`..`674a62661` (5 new commits since the last review): no new concerns — the prior review stands.

PR #5885, opened 2026-09-02
chore(security): time-box container scan waiver bridge

quad341:
## What this changes

This keeps container releases moving while patched bundled tools and image packages are being prepared. The time-boxed `.trivyignore.yaml` bridge now covers every HIGH/CRITICAL finding from the prior image scan, scoped to the affected binary path or package purl and expiring on 2026-09-21.

The policy still refuses to mask Go standard-library findings in the rebuilt `gh`, `dolt`, and `bd` binaries. Regression tests enforce the allowed CVE/path combinations, the expiry and removal rules, and consistency between the waiver file's 64 entries and the release-gate record.

## Why this shape

These are temporary, narrowly scoped exceptions—not dependency fixes. Keeping them in one dated policy bridge makes the release constraint auditable while leaving the permanent fix path clear: rebuild or upgrade the named artifacts, then delete the corresponding waiver.

## Review notes

- The waiver file has 64 entries: 47 from the existing bridge and 17 added for the latest scan findings.
- The new entries cover x/mod, gRPC, thrift, kubectl's Go runtime, and util-linux/GitPython packages in the MCP mail image.
- No production Go code, dependency pin, Dockerfile, image build recipe, or CI workflow changes.
- The new count-consistency guard fails if the gate document and parsed YAML drift apart.

## Test plan

- [x] Run the documented 40-job full local suite; all diff-owned tests pass, with unrelated raw failures attributed in the gate record.
- [x] Run policy, affected-file lint/format, `go vet ./...`, `go build ./...`, docs synchronization, and diff hygiene.
- [x] Verify `CI / required`, `CI / integration`, Trivy, and `Image vulnerabilities` green on the reviewed feature content.
- [x] Release gate: [`release-gates/ga-2yq3p5-container-scan-waiver-bridge-gate.md`](release-gates/ga-2yq3p5-container-scan-waiver-bridge-gate.md)

---
- Related to #5054 — moves all 45 container-scan waivers off the lapsed 2026-08-07 horizon onto a time-boxed 2026-09-21 bridge, and keeps the rebuilt-from-source constraint the filed issue asked about intact — the one added waiver (x/net dns/dnsmessage) is a non-stdlib exception scoped to gh and dolt only, with no kubectl and no stdlib CVE widened onto a rebuilt path, enforced by a new regression test; the durable rebuild of the bundled tools, and the ownership of that work, remain separate and outstanding


quad341:
@quad341 — the scoping discipline on this one is the part worth calling out. Pulling the exact failing findings off run 34553600725 instead of blanket-waiving, deliberately keeping Go-stdlib CVEs *off* the rebuilt gh/dolt/bd paths so the 1.26.5 rebuild still has to prove itself, and widening the x/net entry in place rather than adding a duplicate — those are the choices that keep a waiver file from quietly becoming a mute button. And `TestTrivyIgnoreWidensDocketE9WaiverForRemainingHighCriticalFindings` pins every one of them, including the paths/purls mutual exclusion, so nobody can loosen the scope by accident later.

I checked the purl-only entries against the image tree: the agent images are ubuntu-based and emit `pkg:deb/ubuntu/*`, and gc-mcp-mail is the only Debian/pip image, so the "confined without a paths key" claim holds.

Two small things I'll take on my side rather than send back to you. The six new kubectl stdlib CVEs aren't in the `stdlibCVEs` guard map, and the two grpc CVEs put `gc` back in the waiver list without a drop-when-fixed threshold — so I'll extend both guards, sync kubectl into the `CVE-2026-46600` allowlist, and correct the grpc version cited in those two statements from v1.83.0 to the v1.82.1 the tree actually pins. Then I'll re-run the scripts suite and merge.

`Image vulnerabilities` is green on your head — which was the whole point. Nice work getting it there.


---


quad341:
@quad341 — three widenings deep and this file still isn't a mute button, which is the whole ballgame. The specific things worth calling out: Go-stdlib CVEs stay deliberately *off* the rebuilt gh/dolt/bd paths so the toolchain rebuild still has to prove itself on the scan; `CVE-2026-46600` was widened in place instead of duplicated; and the gc-mcp-mail entries are purl-scoped with no `paths` key so they physically can't leak onto the other images.

I checked rather than trusted, on the head at `74575be`: all 64 entries on the `2026-09-21` horizon, the drop-guards extended for the kubectl stdlib set and both grpc CVEs (bounded at `v1.83.1` against `go.mod`'s `v1.83.0`, so the waiver dies the moment that bumps), and `go test ./scripts/` green. `Image vulnerabilities` is green too — which is the acceptance criterion your gate record left pending.

Nothing for me to absorb. Merging as-is.

One ask for the follow-up, not this PR: the GitPython 3.1.58 → 3.1.59 bump under `ga-dkgeoi` is a CRITICAL RCE with a one-version fix. Worth being first out of the queue before the 21st — and thank you for keeping `#5353` named as the durable route instead of letting the bridge become the answer.


---

quad341:
Re-reviewed `74575bed5`..`0f700a13c` (2 new commits since the last review): no new concerns — the prior review stands.

quad341:
Re-reviewed `0f700a13c`..`03f85ef63` (1 new commit since the last review): no new concerns — the prior review stands.

PR #5888, opened 2026-09-02
fix(session): release closed configured name holders

AlexBelous:
## Summary
- allow a configuration-validated named session to reclaim a runtime name retained by a closed historical record
- preserve the closed record and add regression coverage
- document the lifecycle scenario

## Tests
- `CGO_ENABLED=0 go test -count=1 -v ./internal/api -run '^TestPhase0HandleSessionWake_NamedIdentityIgnoresClosedDrainedNameHolder$'`
- `CGO_ENABLED=0 go test -count=1 ./internal/session -run '^(TestEnsureSessionNameAvailable_AllowsClosedConfiguredNamedSession|TestEnsureSessionNameAvailable_RejectsClosedAdHocSession)$'`
- `git diff --check`

## Gate note
The standard pre-push `make test-fast-parallel` was started but stalled in an unrelated Dolt/module-download path (`github.com/klauspost/compress`) after about 24 minutes and was interrupted. The host initially also lacked ICU header discovery for default CGo (`unicode/regex.h`), though Homebrew ICU is installed at `/opt/homebrew/opt/icu4c/include`.


quad341:
@AlexBelous — thank you for this one. Tracking a wake failure down to a closed *drained* bead still holding the configured identity's runtime name is not an obvious catch, and I confirmed the repro: revert the `names.go` hunk and the wake 500s with `session name already exists: "myrig--worker" already belongs to gc-1`. Your test is the kind I wish more PRs had — it checks the successor *and* that the historical bead is left closed and unmutated. The SESSION-ID-012 ledger row is exactly right too.

One piece I'll take on my side. The release landed in the shared low-level helper, which `EnsureSessionNameAvailableWithConfigForOwner` calls *before* its own owner↔name proof — so `selfOwner != ""` short-circuits that proof and two existing guards go red (`RejectsClosedLegacyBeadWrongOwner`, `RejectsWithoutConfig`). I'll scope the release to the holder's own identity signals so a bare owner assertion isn't enough, add the negative test, and run the full `internal/session` and `internal/api` suites — worth noting your pre-push gate stalled on an unrelated module download, so none of this had CI behind it.

Your test and ledger row land unchanged, and the commit stays yours. Nice work.


---

PR #5889, opened 2026-09-02
fix(runs): execute teardown steps after cancellation

sjarmak:
## Summary

- keep always-run teardown steps open when a run is cancelled
- allow the dispatcher to execute the teardown tail under a cancelled root
- preserve retries and terminal-root control behavior for cleanup work
- add API and dispatcher regressions for cancellation cleanup

## Why

Run cancellation previously closed every open descendant without executing always-run cleanup. A cancelled run could therefore report a clean terminal state while leaving behind a sandbox, checkout, container, lease, or other external resource.

## Verification

- race-enabled cancellation and teardown-tail regressions
- affected API, dispatcher, molecule, and source-workflow package tests
- `go vet` on touched packages
- all ten `make test-fast-parallel` jobs
- independent exact-head review on `def9fb444a0512af05de1aae3f11b9484ae0b34f`

### DIFF
--- a/internal/api/huma_handlers_run_cancel_test.go
+++ b/internal/api/huma_handlers_run_cancel_test.go
@@ -405,6 +405,99 @@ func TestRunCancelWireRoute(t *testing.T) {
 	}
 }
 
+// TestRunCancelLeavesTeardownStepOpenToExecute guards the finding that run
+// cancellation force-closed a workflow's always-run cleanup step as canceled
+// instead of letting it run: cancelRun's blanket subtree close swept the
+// teardown tail along with ordinary open work, so a step whose job is to
+// release the run's external resources (e.g. tearing down a sandbox) never
+// executed. The teardown step must stay open after cancel, and still be
+// closeable with a real outcome afterward, so the dispatcher can drive it to
+// completion — its pass condition may read ROOT_OUTCOME, which only exists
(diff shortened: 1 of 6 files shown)


quad341 (approved):
@sjarmak — this is a good catch and a genuinely nasty class of bug: a cancelled run reporting a clean terminal state while a sandbox, checkout, container or lease quietly stays alive. Thank you for chasing it to the root cause instead of patching the symptom.

What I appreciated most is that you didn't write a second teardown predicate. Hoisting `teardownTailExclusion` into `molecule.TeardownTailExclusion` and having the cancel path, the canceled-root gate and the settled-root gate all defer to one authoritative definition is exactly right — a duplicated copy here is how #5271 comes back in six months wearing a different hat. The assertions in `TestRunCancelLeavesTeardownStepOpenToExecute` are the right shape too: proving the *ordinary* step still closes as `canceled` is what keeps this from quietly making cancel go soft.

I traced the parts your description asserts rather than assuming them — the teardown bead really does stay dispatchable under a closed root, and the run still reports `canceled`, not a stuck `canceling`. Copilot's note about the per-control subtree read checks out as narrow: terminal-root path only, never live dispatch.

Merging as-is. Two additive things I'll pick up on my side: a case covering the `gc.step_id`-only control shape under a canceled root, and a direct unit test for the newly exported predicate.


---

PR #5890, opened 2026-09-02
fix(events): coordinate rotation across recorders

AlexBelous:
## Outcome

Coordinates event recording and rotation through a stable sidecar flock so long-lived peer recorders reopen the active path after an inode-changing rename. This keeps archive filename windows aligned with physical contents and preserves ordered `AfterSeq` reads without loss or duplication.

Closes work item `gs-q5lwb.1`.

## Changes

- keep `events.jsonl.lock` open for the recorder lifetime and use it for `Record`, `AppendBatch`, and `ForceRotate`
- detect stale active-log descriptors with `os.SameFile` and reopen the current path before appending
- add concurrent recorder regression coverage for archive windows, active anchors, full reads, and cursored reads
- retain bounded flock contention behavior and cover rotation waiting on a peer-held lock

## Verification

- `go test -count=10 -run ^TestForceRotateMovesPeerRecorderToFreshActiveFile$ ./internal/events`
- `go test -count=1 -timeout=5m ./internal/events`
- `go test -race -count=1 ./internal/events`
- focused `internal/api` and `cmd/gc` event-rotation tests with CGO disabled
- resource census, `go vet ./...`, staged formatter/linter, code generation, and pre-commit hook

## Local gate exception

The host lacks ICU headers for native CGO builds, so affected API and CLI checks used CGO disabled. The full local push gate also hit the unrelated timing assertion `TestStopManagedCityForcesCleanupAfterTimeout` under host load above 75. An exact detached `origin/main` worktree at `615f5b794` reproduces the same failure (`564.9ms`, bounded-timeout assertion), while all affected patch suites pass. The ownership guard was run manually before the documented exceptional `--no-verify` fork push; required remote CI remains authoritative.

### DIFF
--- a/internal/events/events_test.go
+++ b/internal/events/events_test.go
@@ -1557,7 +1557,7 @@ func TestFileRecorderFlockTimeoutFiresWithinBudget(t *testing.T) {
 	}
 	defer rec.Close() //nolint:errcheck // test cleanup
 
-	sib := mustOpenSiblingLock(t, path)
+	sib := mustOpenSiblingLock(t, recorderLockPath(path))
 	defer func() {
 		_ = syscall.Flock(int(sib.Fd()), syscall.LOCK_UN)
 		_ = sib.Close()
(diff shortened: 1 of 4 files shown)


AlexBelous:
The fork-side implementation is verified locally, including focused regression runs, events race coverage, affected API/CLI tests, resource checks, and vet. The upstream pull_request workflows are currently action_required, so the evidence watchdog failed with tests never ran rather than a test failure. Could a maintainer approve the workflow runs for this fork PR? This requests the normal checks/review path, not a gate bypass.

Issue #5729, opened 2026-08-28
mol-polecat-work control-quarantined finalize leaves live workflow root (blocked-by-self, then hook-assigned to a polecat)


vishnujayvel:
## Summary

A `mol-polecat-work` graph.v2 **workflow-finalize** step that is `gc:control-quarantined` leaves its **molecule root live**. Hook later assigns that root to a polecat. The polecat has no executable formula step, escalates, and the city has to surgically drop a stale `blocks` edge to close the root.

The finalize step itself already recorded why it could not close the root:

```
cannot close blocked issue: <root> is blocked by [<finalize-step>]
```

So the control dispatcher tries to close the workflow head while the finalize bead still `blocks` that head. Beads refuses. The step is then control-quarantined (here: `bd query` timeout). The root stays `in_progress`/`blocked` and remains hook-eligible.

Observed live in a consuming city at gastown pack pin `sha:33d3a430a67d1782ad364556cb566bdb01d0afe3`. This city has no gastown rig; the formula cannot be patched in-tree here.

## Where

- Control dispatcher: `core.control-dispatcher` / `gc.kind=workflow-finalize` (`mol-polecat-work.workflow-finalize`).
- Formula: gastown pack `mol-polecat-work` (graph.v2). The polecat formula text never closes beads; root closure is the finalize step.
- Pack pin: `sha:33d3a430a67d1782ad364556cb566bdb01d0afe3`.

### Incident A — `contour-v2w`

- Source `contour-0ah` already merged at `2fcf3ff` (ancestor of `origin/main`); convoy `contour-4r4` closed.
- Finalizer `contour-2t3` closed but `gc:control-quarantined` (`bd query` timeout).
- Root `contour-v2w` stayed live; polecat `contour/gastown.nux` claimed it via hook and escalated; nux then absent from `gc session list`.
- Recovery: drop stale dep `contour-2t3` → close `contour-v2w`. No worktree salvage. Did not reopen source.

### Incident B — `contour-0ow` (same day)

- Source `contour-0dw` already merged at `11a8243` (`feat(ledger): extend contribution status vocabulary`); convoy `contour-8vc` closed 2026-08-23.
- Finalizer `contour-5d1` closed 2026-08-28T01:57:59Z, labels `gc:control-quarantined`.
- Metadata on `contour-5d1`:
  - `gc.routed_to=contour/core.control-dispatcher`
  - `gc.kind=workflow-finalize`
[...]

PR #5892, opened 2026-09-02
Retry store-read timeouts instead of quarantining workflow finalizers

quad341:
## What this changes

Controller errors containing `timed out after` are now classified as availability-tier failures everywhere, including workflow-finalizer outcome reads. A transient store-read timeout therefore follows the existing retry path instead of quarantining the finalizer on its first refusal and leaving the workflow root live indefinitely.

The implementation extends the central transient-needle table by one entry. Signal-killed and terminated-process handling remains scoped to the work-query path, so the change widens only the store-timeout case.

## Review notes

- The production change is confined to `internal/dispatch/control.go`; it does not raise timeout values or change workflow-root eligibility, dependency edges, cleanup, or self-healing.
- The new regression test covers the reported finalizer wrapper, this repository's own finalizer wrapper, and the already-supported work-query form.
- The behavior remains default-on through the existing controller retry classifier; there is no config or migration surface.

## Test plan

- [x] `make test-local-full-parallel` — complete 40-job suite; non-diff host/tool failures are documented with causal attribution in the gate.
- [x] `go test -count=1 ./internal/dispatch -run '^TestFinalizeStoreTimeoutIsAvailabilityTier$'` — all three regression subtests pass.
- [x] `make test-ci-policy`
- [x] `go vet ./...`
- [x] Release gate: [`release-gates/ga-f681gz-store-read-timeout-availability-gate.md`](release-gates/ga-f681gz-store-read-timeout-availability-gate.md)

---
- Related to #5729 — addresses the store-read timeout that control-quarantined the workflow finalizer — errors containing "timed out after" now classify as availability-tier and follow the existing retry path, covering the optional item 4 of that issue; the blocked-by-self close failure (finalize step still `blocks` the root it must close), the leftover live root, and its hook eligibility are not covered here

### DIFF
--- a/internal/dispatch/control.go
+++ b/internal/dispatch/control.go
@@ -520,6 +520,13 @@ var transientNeedles = []transientNeedle{
 	{needle: "dolt circuit breaker is open", tier: TierAvailability},
 	{needle: "server appears down, failing fast", tier: TierAvailability},
 	{needle: "dolt server unreachable", tier: TierAvailability},
+	// A store read that times out never answered, so it is Tier A wherever it
+	// is raised. isTransientWorkQueryFailure already said so for the drain
+	// work-query path (6d74360fc5); scoping it to that one message prefix left
+	// every other caller quarantining on the first refusal — most visibly
+	// processWorkflowFinalize's outcome read, whose wrapper is "resolving
+	// workflow outcome" (gastownhall/gascity#5729).
+	{needle: "timed out after", tier: TierAvailability},
 }
 
 // ClassifyControllerError is the dispatch/store transient classifier for
--- a/internal/dispatch/finalize_query_timeout_tier_repro_test.go
+++ b/internal/dispatch/finalize_query_timeout_tier_repro_test.go
@@ -1,0 +1,36 @@
+package dispatch
+
+import (
+	"errors"
+	"testing"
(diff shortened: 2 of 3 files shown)


quad341:
Thanks @quad341 — the part worth calling out is that you found *why* the existing machinery missed this rather than just widening something until it caught. `bdExecTimeoutError` builds its message with `fmt.Errorf` and never `%w`-wraps `ctx.Err()`, so `ClassifyControllerError`'s `errors.Is(err, context.DeadlineExceeded)` branch could never fire for it, and `isTransientWorkQueryFailure`'s `"running work query"` prefix scoped the string fallback away from `processWorkflowFinalize`'s `resolving workflow outcome` wrapper. One needle in the right table, with the reasoning left in a comment for whoever hits this next.

I traced it independently and it holds up. Keeping the signal/kill handling scoped to the work-query path was the right call, and resisting the urge to re-scope this needle by a different message prefix is what stops the same bug reappearing at the next wrapper.

Merging as-is. Two things for later, neither a change request: a negative case pinning some non-store `timed out after` message to `TierNone` would fence Tier A against future drift, and the gate doc's failure count reads "2 raw FAIL jobs" while the attribution table names three.


---

PR #5893, opened 2026-09-02
fix(status): defer local store reads behind API route

AlexBelous:
## Summary
- route supervisor-backed `gc status` before opening the local bead/Dolt store
- retain the original snapshot-aware local fallback when the control-plane route is unavailable
- add a regression test proving a healthy API route never opens the local store

## Root cause
`cmdCityStatus` opened the local city bead/Dolt store to construct its session snapshot before checking the supervisor's cached status API. Under unrelated local store contention this left `gc status` with no output despite a healthy control plane. The local open now occurs only in the API fallback.

`order-firing-current` already returns its existing bounded advisory result for a timed-out history lookup; its focused timeout and history-parallelism coverage was revalidated.

## Verification
- `CGO_ENABLED=0 go test -count=1 -v -run '^TestCmdCityStatus_(SupervisorManagedNoAPIPortUsesSupervisorAPI|APIRouteDoesNotOpenLocalStore)$' ./cmd/gc`
- `CGO_ENABLED=0 go test -count=1 -v -run '^(TestOrderFiringCurrent_TimesOutStalledOrderHistory|TestOrderFiringCurrent_TimeoutHintNamesQueryCost|TestOrderFiringCurrent_LastRunLookupsRunInParallel)$' ./internal/doctor`
- `CGO_ENABLED=0 go vet ./cmd/gc`

## Pre-push exception
The required fast gate failed in an unrelated baseline test. The exact serial command `CGO_ENABLED=0 go test -count=1 -run '^TestDefaultRunNetworkGitKillsDescendants$' ./internal/packman` failed identically on this branch (9f158b54e) and a detached clean `origin/main` worktree (615f5b794), both with the heartbeat-file timeout in `network_git_group_unix_test.go:44`. This comparison is recorded in gs-kxlj; the branch was pushed with an explicitly authorized `--no-verify` exception.

### DIFF
--- a/cmd/gc/cmd_citystatus.go
+++ b/cmd/gc/cmd_citystatus.go
@@ -186,11 +186,43 @@ func cmdCityStatus(args []string, jsonOutput bool, stdout, stderr io.Writer) int
 		return 1
 	}
 
+	// The local snapshot opens the bead/Dolt store. Keep it behind the API
+	// fallback: the supervisor serves a cached status view, so touching the
+	// contended local store first defeats the bounded control-plane route and
+	// can leave gc status with no output until an external timeout kills it.
+	localFallback := func() int {
+		return cmdCityStatusLocalFallback(cfg, cityPath, jsonOutput, stdout, stderr)
+	}
+	c, reason := cityStatusAPIClient(cityPath)
+	if c == nil {
+		logRoute(stderr, "status", "fallback", reason)
+		return localFallback()
+	}
+
+	// API rendering only needs the runtime provider for drain-state display;
+	// a nil session snapshot deliberately avoids a bead-store read here.
+	sp, err := newStatusSessionProviderForCityWithSnapshot(cfg, cityPath, nil)
+	if err != nil {
+		message := fmt.Sprintf("gc status: %v", err)
+		if jsonOutput {
+			return writeJSONError(stdout, stderr, "session_provider_failed", message, 1)
+		}
+		fmt.Fprintln(stderr, message) //nolint:errcheck // best-effort stderr
+		return 1
+	}
+	dops := newDrainOps(sp)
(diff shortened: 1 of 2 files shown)


AlexBelous:
The local focused status/doctor tests and cmd/gc vet pass. The only required pre-push failure was reproduced unchanged with the identical serial command on this head and detached clean origin/main, and that baseline exception is documented in the PR/bead evidence. All six upstream pull_request workflows are currently action_required, so the evidence watchdog cannot receive test results. Could a maintainer approve those workflow runs? This requests the normal checks/review path, not a gate bypass.

quad341:
@AlexBelous — this is a genuinely good catch. Opening the bead/Dolt store *before* deciding the route meant a contended store could starve `gc status` even with a perfectly healthy supervisor, and deferring that open behind the API check is the right fix. I especially like the test: hooking `openCityStoreAtForStatus` to `t.Fatal` makes the ordering requirement impossible to fake, which is much stronger than asserting on timing.

I'm taking three small things on my side rather than sending them back to you. The variadic `fallbacks ...` left two fallback implementations in the tree, and the existing `routeCityStatus` tests ended up exercising the one production no longer takes — so I'll make the parameter required and re-point those tests at `cmdCityStatusLocalFallback`, plus add the complement test for the no-client route. I'll also note a bound I found while tracing: with a nil snapshot, ACP route registration falls back to deterministic session names, so an ACP agent whose runtime name diverges can silently lose its `(draining)` label on the API path. Narrow and non-fatal — I'll comment it and file a follow-up.

One correction on your PR notes: the red check isn't the `packman` flake you documented — it's the evidence watchdog firing because `ci.yml` never ran on this head. I'll approve the run. Your reasoning and credit stay on the commit.


---

PR #5899, opened 2026-09-02
perf(config): scope legacy hook wisp queries

rbriski:
## Summary

- push assignee predicates into legacy ephemeral `bd query` hook probes
- scope the legacy pool fallback to unassigned wisps at the database boundary
- retain the existing jq hold, dependency, type, and routing filters as defense in depth

## Root cause

The bd-1.0.4 compatibility path queried every wisp matching only `ephemeral` and `status`, then filtered assignees locally in jq. On a controlled Dolt fixture with 320,000 durable issues, 320,000 wisps, and 680,000 label rows, the three-identity assigned probe tier took 5.45 seconds even though no identity matched. Beads supports indexed `assignee` predicates in the query language, so Gas City can avoid hydrating unrelated history without changing readiness semantics.

## Verification

- regression test failed before the change and passes after it
- controlled assigned probe tier: 5.45s -> 0.45s, valid JSON
- production-shaped controlled `gc hook --claim --json`: 4.896s, valid claimed-work JSON
- `GOENV=/var/tmp/ga-y8g-goenv make test-fast-parallel` (pure-Go build because this host lacks optional ICU headers)
- `GOENV=/var/tmp/ga-y8g-goenv GOFLAGS='-tags=license_tier_oss,sqlite_fts5' go vet ./...`
- staged `.githooks/pre-commit`
- independent review of exact head `63f5d737f0d0901e2e46db5f611134b80694c9da`: approve, no P0-P3 findings

No production hook claim, Dolt restart, compaction, cleanup, or data mutation was used for validation.


rbriski:
All required checks are green at exact head `63f5d737f0d0901e2e46db5f611134b80694c9da`. Independent read-only review approved the candidate with no P0-P3 findings. The controlled regression evidence used no production Dolt mutation. This account has read-only upstream permissions and cannot merge or enable auto-merge, so maintainer review and protected merge are requested; no admin bypass will be used.

Issue #2763, opened 2026-05-30
bug: graph.v2 (formula_v2) workflow roots routed to a pool are never claimed — work-query reads gc.routed_to, graph roots stamp only gc.run_target


sjarmak:
## Summary

Under `formula_v2` (opt-in; `config.md:276` notes "Default: false (opt-in while the feature stabilizes)"), a `graph.v2` workflow root routed to a pool agent is **never claimed**: the pool's reconciler spawns a worker (scale_check reads `gc.run_target`), but the worker's own work-query reads only the legacy `gc.routed_to`, which is **unset** on graph workflow roots. The worker finds an empty queue, sits idle, and is eventually idle-reaped — silently orphaning the work. The legacy formula path is unaffected (it stamps `gc.routed_to`). This is a stabilization gap in formula_v2 — it completes the reader-migration that #2386 started; `bdReadyPoolDemandShell` was the one reader site it missed.

## Symptom

`gc sling <pool> <bead>` or `gc sling <pool> <bead> --on <formula>` (the graph.v2 path) routes the work, the molecule shows `in_progress`, but no pool worker ever claims it. `gc hook` for that pool returns `[]`; an explicit `gc session nudge` does not help. The worker logs an empty queue and is later reaped via `session reconciler: idle timeout`. No error surfaces — the dispatched work just vanishes into idle-reap.

`gc sling <pool> <formula> --formula` (legacy wisp instantiation) is **not** affected — its root is claimed and run normally.

## Reproduction

City with `[daemon] formula_v2 = true`, a pool agent (e.g. `polecat`, `min_active_sessions >= 1`).

1. `gc sling <pool> <existing-bead>`  (or `... --on mol-do-work`) → creates a `kind=workflow` root with `gc.run_target=<pool>`, `gc.routed_to` UNSET.
2. The reconciler spawns/keeps a worker for `<pool>` (scale_check sees `run_target` demand).
3. The worker runs its work-query and gets nothing.

Result: expected → worker claims the routed root and runs it. Got → `gc hook` returns `[]`, root stays `in_progress`/unassigned, worker idles then idle-reaps; work orphaned (`routed_to=NONE`, 0 progress).

Contrast (works): `gc sling <pool> <formula> --formula` → `kind=wisp` with `gc.routed_to=<pool>` set → claimed and run.

## Root cause

The pool worker's work-query is `bdReadyPoolDemandShell()` — `internal/config/config.go:2634-2636`:

```go
[...]

PR #5901, opened 2026-09-02
fix(hook): gate workflow root run_target fallback on gc.workflow_expanded (#5900)

### LINKED ISSUE #5900: bug: graph.v2 fully-expanded workflow root reclaimed forever via gc.run_target fallback before finalize runs
## Summary

A `graph.v2` workflow root bead becomes claimable through `gc hook --claim` once all of its real children have closed but before `workflow-finalize` has run, because the claim path's `gc.run_target` fallback for `KindWorkflow` candidates does not distinguish "root-only molecule with no expanded children yet" (the case it was built for) from "fully-expanded molecule whose only remaining edge is the non-blocking finalize tracker". Sessions repeatedly claim and drain the unclosable root forever, and it never actually completes any work.

## Symptom

A pool worker calls `gc hook --claim --drain-ack --json`, is handed a `bead_id` whose `gc.kind` is `workflow`, does no real work (there is none left — every real child is closed), and either closes nothing or fails to close the root (workflow roots are documented as never-claimable, so any close attempt is semantically wrong). The next spawned worker gets the same root again. Observed: 0 closures over ~2 hours (15:58–17:55Z) against a `luna-worker`-routed root, burning every session slung at it.

## Reproduction

1. Compile and dispatch a `graph.v2` formula whose root is routed to a pool via `gc.routed_to`/`gc.run_target` (any formula compiled by `internal/formula/compile.go`).
2. Let all real children close, but do not let `workflow-finalize` run yet (e.g. pause/slow the finalizer, or race it).
3. Call `gc hook --claim` (or let any pool worker in the route poll) while the root sits in this state.

Expected: the root, `gc.kind=workflow`, is never returned as claimable (`internal/beadmeta/kindsets.go:111-114`: "Routing never lands on these; agents must never claim them").
Got: the root is returned and claimed, because `hookClaimMatchesRoute` synthesizes a route match from `gc.run_target` for any `KindWorkflow` candidate with no positive-vs-fully-expanded distinction.

## Root cause

`internal/beadmeta/kindsets.go:111-118` documents `WorkflowTopologyKinds` (`workflow`, `scope`, `spec`) as never claimable. Nothing in the actual read/claim paths enforces that:

- `cmd/gc/cmd_ready.go` (`filterReadyBeads`, `readReadyCandidates`) has no `gc.kind` filter.
- None of `internal/beads/{bdstore,sqlite_store,doltlite_read_store,native_dolt_store,memstore,filestore,caching_store_reads}.go` filter by `gc.kind`.
- `cmd/gc/cmd_hook_claim.go:2074-2094` (`hookClaimMatchesRoute`) actively synthesizes a route: when `gc.routed_to` is empty and `gc.kind == KindWorkflow`, it falls back to matching `gc.run_target` (line 2089) against the worker's route targets. `hookClaimRoute` (2116-2124) mirrors the same fallback for candidate visibility/route reporting.

That fallback was added deliberately for #2763 ("graph.v2 workflow roots routed to a pool are never claimed — work-query reads `gc.routed_to`, graph roots stamp only `gc.run_target`"), which covers a slung root with **no expanded children yet** that needs to be claimable as the initial unit of work.

`internal/formula/compile.go:674-687` (`addWorkflowRootDeps`) shows why the fallback now also fires on fully-expanded molecules: the root's edge to `workflow-finalize` is deliberately a `"tracks"` edge (informational, non-blocking), "so a 'blocks' edge here would make the root permanently unclosable... the root is a latch that is never routed or dispatched (`beadmeta.WorkflowTopologyKinds`)." Once every real child closes, the root has no remaining *blocking* dependency (the finalize edge never blocked it), so it becomes dependency-ready before `workflow-finalize` itself has run — and `hookClaimMatchesRoute`'s fallback has no check for "has this root already been fully expanded / does real child work exist", so it matches this case exactly the same as the genuine #2763 case.

Confirmed present on `origin/main` at `e3bee4cd1` (2026-09-02).

## Adjacent, not the same bug

PR #5474 ("prefer executable steps over workflow anchors") fixes a related but distinct symptom in `internal/config/workquery.go`: query-ordering causing a root to be preferred over **still-ready sibling children**. It doesn't apply here — in this incident's failure mode the root is the *only* ready candidate (every real child already closed), so no amount of reordering among candidates helps; the root itself must not be claimable at all until `workflow-finalize` has run.

## Fix candidates

**A. Gate the `gc.run_target` fallback on absence of expanded children.** In `hookClaimMatchesRoute`/`hookClaimRoute`, before honoring the `KindWorkflow` + `run_target` fallback, check whether the candidate root has any compiled child dependency edges (i.e. was actually expanded by `compile.go`, as opposed to a genuinely root-only #2763-style molecule). Only allow the fallback when no such children exist.
- Scope: ~30-60 LOC in `cmd/gc/cmd_hook_claim.go`, plus a store lookup (or a compile-time metadata stamp, e.g. `gc.expanded=true`, to avoid a claim-time read) and new tests alongside the existing `hookClaimMatchesRoute` coverage.
- Trade-off: minimal blast radius, stays local to the already-defective function; needs either an extra read on the hot claim path or a new compile-time metadata field threaded through `internal/formula/compile.go`.

**B. Filter `WorkflowTopologyKinds` out of the ready/claim surfaces uniformly, and move the #2763 root-only-launch case onto an explicit `gc.routed_to` stamp instead of the `gc.run_target` fallback.** Since `internal/graphroute/graphroute.go:612,710` already deliberately withholds `gc.routed_to` from workflow-topology steps, add a narrower compile/route-time condition that stamps `gc.routed_to` (not `gc.run_target`) specifically for a root with zero children (the #2763 shape), then delete the `run_target` fallback in `cmd_hook_claim.go` entirely, making `WorkflowTopologyKinds` genuinely unclaimable as `kindsets.go` already documents.
- Scope: touches `internal/graphroute/graphroute.go` (routing decoration), `internal/formula/compile.go` (or wherever root-only molecules are distinguished), and removes ~15 LOC of fallback in `cmd/gc/cmd_hook_claim.go`.
- Trade-off: matches the documented invariant exactly ("agents must never claim them") with no runtime child-count check, but larger blast radius across the compiler/router and needs care to not regress #2763.

**Recommendation:** A — smaller, localized to the exact defective function, and doesn't touch the compiler/router surfaces that other unrelated logic (`internal/dispatch/drain.go:1566`, `graphroute.go:612,710`) already depends on for different reasons. Maintainer's call.

## Out of scope

- The `gc ready` vs. bare `gc bd show`/query divergence (store-split staleness window, `internal/beads/caching_store_reconcile.go`) — a separate, narrower concern once its own staleness-precedence question is resolved.
- Closing/finalizing roots already stuck in this state — an operational cleanup, not part of this defect's fix.
- PR #5474 and #5800, which address different symptoms in the same general "graph.v2 root vs. children" area and should land independently.

jacobhausler:
## Summary

Fixes #5900. A `graph.v2` workflow root's `gc.run_target` claim fallback
(added for #2763 so a genuinely root-only molecule is claimable as its own
unit of work) did not distinguish that shape from a **fully-expanded** root
whose real children have all closed but `workflow-finalize` has not yet run.
Since the root→finalize edge is deliberately non-blocking (`tracks`, not
`blocks` — see `addWorkflowRootDeps`), the root goes dependency-ready in that
window and the fallback let it be claimed anyway, even though
`WorkflowTopologyKinds` documents workflow roots as never-claimable. Pool
workers then repeatedly claimed and drained the unclosable root (observed: 0
closures over ~2 hours against a `luna-worker`-routed root).

This implements the issue's recommended fix candidate A, refined to stamp
expansion state at **compile time** rather than reading child count at claim
time:

- `internal/molecule/graph_apply.go` and `internal/molecule/molecule.go` both
  stamp `gc.workflow_expanded=true` on a `graph.v2` root when it's compiled
  with real child steps (`!recipe.RootOnly`). A genuinely root-only
  (#2763-shape) molecule is left unstamped.
- `cmd/gc/cmd_hook_claim.go`'s `hookClaimMatchesRoute`/`hookClaimRoute` now
  gate the `gc.run_target` fallback on the *absence* of that stamp via a new
  `workflowRunTargetFallbackEligible` helper, so a fully-expanded root can
  only become claimable through `gc.routed_to` (which it never has, per
  `WorkflowTopologyKinds`) — never through the fallback.

This avoids an extra store lookup on the hot claim path (`gc hook --claim` /
`gc ready` candidate filtering / route-visibility reporting) since the
distinguishing state is already known and stamped when the root is compiled.

## Not in scope

- PR #5474 (query-ordering preference among ready candidates) and #5800
  address different symptoms in the same general area and are unaffected —
  confirmed no changes to `internal/config/workquery.go` or
  `internal/graphroute/graphroute.go`.
- Cleaning up roots already stuck in this state is an operational concern,
  not part of this fix.

## Test plan

- [x] `go test ./cmd/gc/...` — full package, no regressions (new:
      `TestHookClaimMatchesRouteWorkflowRunTargetFallback` covers both the
      root-only-must-still-fallback and expanded-must-not-fallback cases,
      plus `hookClaimRoute` visibility for each).
- [x] `go test ./internal/molecule/...` — full package, no regressions (new:
      `TestInstantiateStampsWorkflowExpandedForRealChildren` covers both the
      graph-apply and sequential instantiation paths;
      `TestInstantiateRootOnlyGraphWorkflowOmitsWorkflowExpanded` pins the
      root-only omission).
- [x] `go vet` / `gofmt -l` clean on all changed files.

### DIFF
--- a/cmd/gc/cmd_hook_claim.go
+++ b/cmd/gc/cmd_hook_claim.go
@@ -2071,13 +2071,32 @@ func hookRouteIdentitiesEqual(a, b string) bool {
 	)
 }
 
+// workflowRunTargetFallbackEligible reports whether candidate is a
+// KindWorkflow root the gc.run_target fallback may apply to. The fallback
+// exists so a genuinely root-only (#2763-shape) molecule - whose root IS the
+// unit of work, with no compiled children - is claimable via its
+// gc.run_target authoring hint. It must not also resurrect a fully-expanded
+// root: once compile.go gives a graph.v2 root real child steps, it stamps
+// gc.workflow_expanded=true, and that root's only remaining path to
+// dependency-readiness is every real child closing while workflow-finalize
+// has not yet run and closed it (#5900) - a state the fallback must not
+// treat as claimable (WorkflowTopologyKinds document workflow roots as never
+// claimable). A candidate without the stamp predates this fix or was never
+// expanded, so it keeps the original permissive behavior.
+func workflowRunTargetFallbackEligible(candidate beads.Bead) bool {
+	kind := strings.TrimSpace(candidate.Metadata[beadmeta.KindMetadataKey])
+	if kind != beadmeta.KindWorkflow {
+		return false
+	}
+	return strings.TrimSpace(candidate.Metadata[beadmeta.WorkflowExpandedMetadataKey]) != "true"
+}
+
 func hookClaimMatchesRoute(candidate beads.Bead, routeTargets []string) bool {
 	if len(routeTargets) == 0 {
 		return false
 	}
 	routedTo := strings.TrimSpace(candidate.Metadata[beadmeta.RoutedToMetadataKey])
 	runTarget := strings.TrimSpace(candidate.Metadata[beadmeta.RunTargetMetadataKey])
-	kind := strings.TrimSpace(candidate.Metadata[beadmeta.KindMetadataKey])
 	for _, target := range routeTargets {
 		target = strings.TrimSpace(target)
 		if target == "" {
@@ -2086,7 +2105,7 @@ func hookClaimMatchesRoute(candidate beads.Bead, routeTargets []string) bool {
 		if hookRouteIdentitiesEqual(routedTo, target) {
 			return true
 		}
-		if routedTo == "" && kind == beadmeta.KindWorkflow && hookRouteIdentitiesEqual(runTarget, target) {
+		if routedTo == "" && workflowRunTargetFallbackEligible(candidate) && hookRouteIdentitiesEqual(runTarget, target) {
 			return true
 		}
 	}
@@ -2117,7 +2136,7 @@ func hookClaimRoute(candidate beads.Bead) string {
 	if routedTo := strings.TrimSpace(candidate.Metadata[beadmeta.RoutedToMetadataKey]); routedTo != "" {
 		return routedTo
 	}
-	if strings.TrimSpace(candidate.Metadata[beadmeta.KindMetadataKey]) == beadmeta.KindWorkflow {
+	if workflowRunTargetFallbackEligible(candidate) {
 		return strings.TrimSpace(candidate.Metadata[beadmeta.RunTargetMetadataKey])
 	}
 	return ""
--- a/cmd/gc/cmd_hook_visibility_test.go
+++ b/cmd/gc/cmd_hook_visibility_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"testing"
 
+	"github.com/gastownhall/gascity/internal/beadmeta"
 	"github.com/gastownhall/gascity/internal/beads"
 	"github.com/gastownhall/gascity/internal/config"
 )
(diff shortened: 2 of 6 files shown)


quad341 (approved):
@jacobhausler — this is genuinely good debugging. Connecting the `tracks`-not-`blocks` root→finalize edge to the dependency-ready window, and seeing that the #2763 fallback couldn't tell a root-only molecule from a fully-drained one, is the hard part. Stamping expansion at compile time instead of counting children at claim time is also the better of the issue's candidates — it keeps `gc hook --claim` off an extra store lookup on the hot path. And testing both the graph-apply and sequential instantiation paths, rather than just whichever the default config happens to use, is the kind of thing that saves someone a config-dependent surprise later.

One thing I found tracing outward: the same `kind==workflow && routed_to=="" → run_target` rule has siblings — the controller's demand counter and two work-query filters. With only the claim copy gated, the controller keeps counting an expanded root as demand for workers that now refuse it, so a seat spawns and drains in a loop. That's ours to absorb, not yours to respin. I'm gating the demand path and those filters, adding the expanded-root row to the demand/serve agreement fixture (which is why that suite stayed green), and filing a follow-up for backfilling roots already stuck.

Your commits and your approach stay as-is. Thanks for this one.


---

PR #5902, opened 2026-09-02
feat(rppcheck): verify declared nudge fallback capability

quad341:
## What this changes

`gc runtime check <name>` now verifies a runtime pack's declared `prompt_delivery = "nudge-fallback"` capability by sending a real nudge to the throwaway conformance session. A runtime that advertises this fallback but cannot perform it fails validation immediately instead of failing later when an oversized startup prompt needs delivery.

Runtimes that do not declare nudge fallback keep the existing optional capability-check behavior.

## Review notes

- The declarative prompt-delivery strategy is threaded into `rppcheck.Options.RequireNudge`; the checker remains independent of runtime names and roles.
- Required nudge errors and unsupported responses are hard failures. Start failure still remains the primary failure and records the unrun required check as skipped.
- No configuration migration or wire-format change is introduced.

## Test plan

- [x] Run the repository's complete 40-job local suite with the rootless Podman environment enabled.
- [x] Exercise the required-nudge success, unsupported, error, and start-failure paths, plus the command-level declaration test.
- [x] Run build, vet, affected lint, formatting, CI-policy, dependency-surface, boundary, documentation, and native-DoltLite checks.
- [x] Release gate: [`release-gates/ga-zzj1pu-runtime-check-nudge-fallback-gate.md`](release-gates/ga-zzj1pu-runtime-check-nudge-fallback-gate.md)


quad341:
@quad341 — thanks for this, and specifically for routing `nudgeCheckName()` through *both* `checkSessionOps` and `remainingCheckNames`. It would have been easy to escalate the live probe and leave the start-failure skip listing still reporting `optional: nudge`; you didn't, and you pinned it with `TestRun_RequireNudgeStartFailureSkipsWithRequiredName`. Keeping `checkNudge`'s default branch exactly equivalent to the old `checkOptional` call is the right instinct too — I traced it, and the no-declaration path is genuinely untouched.

Merging with one thing absorbed on my side: `pack-spec.md` still says a false `nudge-fallback` claim "will not fail until an oversized prompt actually reaches that runtime," which your change partly supersedes, and the cobra `Long` text (and therefore the generated `cli.md`) still promises that absent optional ops never fail the run. Prose plus a `make generate` — no reason to send it back to you for that.

Worth a follow-up bead rather than this PR: the escalation only fires on the name-resolution path, so a pack CI running `gc runtime check ./scripts/provider.sh` never gets it — which is the invocation our own provider docs recommend. A `--require-nudge` flag would close it. Nice work.


---

Issue #4189, opened 2026-07-12
Verify rig-scoped pool defaults cover moneymachine scale_check/hook workaround


dlarsen5:
## Field report

Moneymachine hit a Gas City pool-dispatch failure during the 2026-07-11/2026-07-12 Gas Town migration burn-in. The local follow-up is `moneymachine-6pra3`, discovered from `moneymachine-o9ixc` and `moneymachine-4jo9t`.

Observed symptoms in the moneymachine city:

- Default pool `scale_check` appeared to run `bd ready` in the HQ scope, so rig-scoped `moneymachine-*` beads were invisible and demand was reported as `0` even when routed unassigned polecat work existed in the rig store.
- `gc hook` returned `[]` for a role without an explicit `work_query`, so spawned role sessions could sleep through queued/routed work.

Moneymachine added local workarounds in `.gascity/agents/polecat/agent.toml`:

```toml
scale_check = "gc bd --rig moneymachine ready --metadata-field gc.routed_to=moneymachine/polecat --unassigned --json 2>/dev/null | jq 'length'"
work_query = "gc bd --rig moneymachine ready --metadata-field gc.routed_to=moneymachine/polecat --unassigned --json 2>/dev/null"
```

Those should stay local until upstream default behavior is verified in a released build.

## Upstream status to verify

`main` appears to contain related fixes already:

- PR #1594 / commit `aacd9fdacc59ad57409175f251962ebb9106c913` shares the pool-demand predicate with `work_query` Tier 3.
- `cmd/gc/build_desired_state.go` now picks a rig store for rig-scoped default scale checks.
- `cmd/gc/cmd_hook.go` calls `EffectiveWorkQueryForBeads`, so an agent with no configured `work_query` should still have a default routed-work query.

The remaining ask is to verify this exact field scenario and make the release boundary explicit.

## Acceptance criteria

- Add or point to a regression that creates a rig-scoped pool agent with no explicit `scale_check` and no explicit `work_query`, creates ready unassigned work routed to that agent in the rig store, and proves both:
  - controller demand is non-zero / a worker is desired;
  - `gc hook` from that worker context surfaces the work.
- Confirm which released `gc` version contains the fix, or note if a new release is needed.
- Once verified, moneymachine can remove the explicit local workarounds in a later repo-local cleanup.

dlarsen5:
## Version 1.4.1 production result

**Change:** Adding a production result from Gas City 1.4.1 for the local worker-demand workaround.

[...]

PR #5905, opened 2026-09-02
fix(beads): preserve indefinite deferred cached readiness

dlarsen5:
## Summary

**Change.** Keep indefinitely deferred work out of cached ready queues. Preserve that hold across native Dolt reads, cache priming and refresh, notifications, dependency invalidation, class-store events, and migration equality. Explicit reopen and close transitions still clear it.

**Why.** The work item CLI (`bd`) represents an indefinite deferral as `status=deferred` with no `defer_until`. Gas City normalizes that status to `open`, so the cached row previously became indistinguishable from runnable work and could reappear in `bd ready` after prime or refresh.

**State.** Implementation and regression coverage are complete. This bug was exposed while validating the worker-pool fix tracked in #4189; it is a separate cache-state defect.

**Next.** Upstream maintainers review and merge. No MoneyMachine owner action is required.

## Design

The new internal `IndefinitelyDeferred` projection preserves that source fact without changing Gas City's public three-state status model. Explicit `open` and `close` operations remove it; unrelated updates and cache/event round trips retain it.

## Testing

- [x] `make check`
- [x] targeted `go test -race` for deferred-readiness cache paths
- [x] `go test ./internal/beads -count=1`
- [x] `go test ./cmd/gc -count=1`
- [x] `go vet ./internal/beads ./cmd/gc`
- [ ] `make test-integration`: changed packages passed; the suite failed in unrelated runtime tests. Three hidden-tmux attach cases passed on immediate focused rerun. `TestE2E_SuspendResume_City` timed out identically on this branch and a clean `origin/main` worktree at `c96e54a3a`.
- [x] `make check-docs` not applicable: no documentation, navigation, or link changes

## Compatibility

The added marker is internal and excluded from JSON. Canonical bead-event payloads remain raw bead snapshots; consumers that do not know the field continue to decode the unchanged wire shape. No breaking change or migration is required.

## Checklist

- [x] Linked the related discovery context (#4189); no separate issue is needed because this pull request fully defines and fixes the defect
- [x] User-facing documentation is unchanged
- [x] Called out compatibility and migration impact above

### DIFF
--- a/cmd/gc/class_store_emit.go
+++ b/cmd/gc/class_store_emit.go
@@ -51,9 +51,9 @@ package main
 //
 // # What it emits, and what it deliberately does not
 //
-// The payload is the canonical bead snapshot CachingStore.notifyChange emits —
-// json.Marshal of the post-write bead, decodable by beads.DecodeBeadEventPayload
-// and foldable by the run projection — with the run/session/step correlation
+// The payload is the canonical bead snapshot from
+// beads.EncodeBeadEventPayload, decodable by beads.DecodeBeadEventPayload and
+// foldable by the run projection, with the run/session/step correlation
 // resolved onto the typed envelope from the same metadata keys and through the
 // same helpers. bead.closed rides only a genuine open→closed transition, because
 // the export boundary drops bead.updated and a metadata write to a closed bead
@@ -74,7 +74,6 @@ package main
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"maps"
@@ -165,7 +164,7 @@ func (s *emittingClassStore) emit(emissions ...classStoreEmission) {
 		if strings.TrimSpace(emission.bead.ID) == "" {
 			continue
 		}
-		payload, err := json.Marshal(emission.bead)
+		payload, err := beads.EncodeBeadEventPayload(emission.bead)
 		if err != nil {
 			warnClassStoreEmit(fmt.Errorf("marshaling the %s payload of %s: %w", emission.eventType, emission.bead.ID, err))
 			continue
--- a/cmd/gc/class_store_emit_test.go
+++ b/cmd/gc/class_store_emit_test.go
@@ -653,6 +653,48 @@ func TestClassStoreEmissionHydratesDependencyEdges(t *testing.T) {
 	}
 }
 
+func TestClassStoreEmissionPreservesStatusBasedDeferralUntilClose(t *testing.T) {
+	cityPath := t.TempDir()
+	leaf := beads.NewMemStore()
+	store := resolveGraphStore(splitClassRoutes(leaf).withCLIEmission(cityPath), beads.NewMemStore(), nil, cityPath, nil)
+
+	deferred, ok := beads.DecodeBeadEventPayload(
+		json.RawMessage(`{"id":"source-deferred","title":"deferred","status":"deferred","issue_type":"task"}`),
+	)
+	if !ok {
+		t.Fatal("deferred fixture did not decode")
+	}
+	created, err := leaf.Create(deferred)
+	if err != nil {
+		t.Fatalf("seeding deferred bead: %v", err)
+	}
+
+	if err := store.SetMetadata(created.ID, "gc.note", "still deferred"); err != nil {
+		t.Fatalf("metadata write: %v", err)
+	}
+	if err := store.Close(created.ID); err != nil {
+		t.Fatalf("close: %v", err)
+	}
+
+	got := beadEvents(readCityJournal(t, cityPath))
+	if len(got) != 2 {
+		t.Fatalf("got %s, want one update and one close", eventSummary(got))
+	}
+	var updateWire struct {
+		Status string `json:"status"`
+	}
+	if err := json.Unmarshal(got[0].Payload, &updateWire); err != nil {
+		t.Fatalf("decode update payload: %v", err)
+	}
+	if got[0].Type != events.BeadUpdated || updateWire.Status != "deferred" {
+		t.Errorf("first event = %s status=%q, want bead.updated status=deferred", got[0].Type, updateWire.Status)
+	}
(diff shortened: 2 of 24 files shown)


dlarsen5:
**State.** GitHub did not dispatch the required suites for this fork head, so the Evidence / critical-path suites watchdog timed out waiting for checks; it did not observe a failing code check.\n\nLocal evidence: `make check`, changed-package tests, and the focused race run pass. The broad local integration result and its baseline comparison are documented in the pull request body.\n\n**Next.** An upstream maintainer needs to approve the fork workflow and rerun the evidence watchdog. No MoneyMachine owner action is required.

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277289348

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @dlarsen5! This PR fixes a real correctness hole at the beads read edge: a bead parked with bd's status-based *indefinite* deferral (`status=deferred` with no `defer_until`) was byte-indistinguishable from an ordinary open bead by the time it reached the readiness gate, so deliberately-parked work kept re-entering the ready set and getting picked up. That is the kind of bug that quietly costs agent cycles and makes "why did something claim this?" unanswerable, and your fix closes it at the layer where bd's vocabulary is actually read rather than patching one symptom.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **The deferral was erased on the way in.** `mapBdStatus` collapsed every status except `closed`/`in_progress` to `"open"`, and `IsDeferred` inspected only `DeferUntil` — so a `status=deferred, defer_until IS NULL` row passed *both* halves of the readiness gate. The PR derives a `Bead.IndefinitelyDeferred` marker where bd's status vocabulary is read and teaches `IsDeferred` to honor it. Review independently re-derived this against the base commit rather than taking the premise on trust, and it reproduces exactly as described.
- **The un-defer write was being swallowed.** The conditional-update short-circuit meant the one operation that *removes* a deferral would no-op; the added `*opts.Status == "open" && b.IndefinitelyDeferred` clause is the necessary write-side complement.
- **The marker cannot go stale.** `setBeadStatus` clears it across all 13 converted call sites, and the reopen case is pinned by a test — so the new state can't outlive an explicit transition.
- **The infra-class copy compared a field the destination cannot hold.** Exempting it is the right call, and the equality test was moved to the deployed-SQLite tier so it reproduces the operator-visible error instead of passing vacuously.

Review findings addressed:
- *Change impact / blast radius (minor):* the event-path decoder had pulled `mapBdStatus` into the one path that never had it, silently collapsing `blocked`/`hooked`/`pinned`/absent statuses to `"open"`. The maintainer fixup (`f78e8c5c`) scopes that normalization to `Status == "deferred"` and adds a preservation test asserting the other statuses decode verbatim, so the narrowing is pinned and cannot re-widen unnoticed.
- *Debuggability (minor):* the guard comment in `native_dolt_store.go` stated a contract for `IsDeferred` that this PR itself falsifies — in the function the PR most changed. Replaced with a verified description (`f78e8c5c`).
- *Architectural consistency (nit):* the `IsDeferred` doc had dropped its pointer to `cmd_hook.isFutureDeferredHookCandidate`, the mirror that implements only the time-bound half. Pointer restored (`f78e8c5c`).
- Earlier passes also landed documentation and comment corrections in `docs/reference/events.md`, `internal/beads/event_payload.go`, and `cmd/gc/infra_class_migrate_equality_test.go`, all carried in the same maintainer fixup.

- `ga-0veut`: give the infra copy boundary a durable representation for indefinite deferral. Because the marker is `json:"-"` and `SQLiteStore` persists beads through `bead_json`, a marker-bearing bead copied into the binding reads back plainly open. Not a regression — the failing shape isn't constructible on this PR's base, and the PR documents the gap in the copy exemption itself.
- `ga-j9va1`: the `BdStore` staged-Tx write path can clobber an indefinite deferral — `preservedUpdateOpts` re-emits `--status open` on an unrelated field update. Pre-existing; the marker this PR adds makes the eventual guard straightforward.
- `ga-n1r8g`: `doltlite_read_store.go` is the one remaining bd-vocabulary read edge still calling bare `mapBdStatus`, so it still loses the deferral. Pre-existing and behind a build tag plus an env opt-in; converting it needs `defer_until` added to that `SELECT`.
- `ga-pnds6`: `filterUnreadyHookCandidates` catches `blocked` but its deferral arm only looks at `defer_until`, so a raw `status=deferred` row passes all five hook paths. Pre-existing defense-in-depth gap that this PR makes more visible rather than creating.
- `ga-ajbp0`: surface a derived `is_deferred` on the bead wire type. Today an indefinitely-deferred bead reports `status:"open"`, `defer_until:null`, `is_blocked:false` — three signals that all read "ready" — so an operator has nothing to read when asking why a bead is parked.

### Maintainer Changes
One maintainer fixup commit, `f78e8c5c` *fix(beads): maintainer review fixes for the indefinite-deferral PR* — 9 files changed, +136/−20. Contributor authorship on the original commit is preserved.

### Final Review Status
Ready for the merge queue: final head `f78e8c5c` has passing required GitHub checks, and the merge-ready metadata is in place. The branch was also verified against the current `main` — it rebases cleanly with an unchanged patch-id, and `go build ./...`, `go vet`, the full `internal/beads` and `internal/runproj` suites, and all 49 tests in the `cmd/gc` files this PR touches were re-run green on that refreshed tree, which matters because `main` independently moved `internal/beads/bdstore.go`, `beads.go`, and `native_dolt_store.go` since the review basis. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/33859815676

### Review Iterations
Four review passes performed. The loop opened with request_changes on the infra-copy equality comparison and its missing deployed-destination coverage, converged through maintainer fixups for the decoder-scope, stale-contract, and cross-reference findings, and closed with approve on the final pass.

---

PR #5909, opened 2026-09-02
fix(ci): scan every image before failing Container Scan

quad341:
## What this changes

The Container Scan `Image vulnerabilities` job now evaluates all four built
images before reporting its final result. An unwaived finding in an early image
no longer prevents `gc-agent`, `gc-controller`, and `gc-mcp-mail` from being
scanned, and the log reports a clear PASS or FAIL summary for every image.

The job still fails when any image has an unwaived HIGH or CRITICAL finding.
This change exposes the complete result set; it does not relax vulnerability
policy.

## Review notes

- The existing `set -euo pipefail` behavior remains in place for unexpected
  command errors; the expected Trivy policy exit is handled inside an
  `if` and accumulated across images.
- Severity, `--ignore-unfixed`, `.trivyignore.yaml`, image inputs, and dependency
  pins are unchanged.
- A static regression test protects the four-image loop, per-image summaries,
  unchanged scan flags, and aggregate nonzero exit.

## Test plan

- [x] Run the full local unit, process, and integration union with the rootless
  container environment enabled; confirm the candidate-owned regression passes.
- [x] Run workflow policy, formatting, lint, and standalone vet lanes.
- [x] Run the real Container Scan workflow at this head and confirm all four
  images are scanned before the aggregate failure is returned.
- [x] Release gate: [`release-gates/ga-t600qs-container-scan-enforcement-gate.md`](release-gates/ga-t600qs-container-scan-enforcement-gate.md)

### DIFF
--- a/.github/workflows/container-scan.yml
+++ b/.github/workflows/container-scan.yml
@@ -253,8 +253,9 @@ jobs:
             "gc-controller:${IMAGE_TAG}"
             "gc-mcp-mail:${IMAGE_TAG}"
           )
+          overall_rc=0
           for image in "${images[@]}"; do
-            trivy image \
+            if trivy image \
               --scanners vuln \
               --severity HIGH,CRITICAL \
               --ignore-unfixed \
@@ -262,8 +263,14 @@ jobs:
               --exit-code 1 \
               --timeout 15m \
               --format table \
-              "$image"
+              "$image"; then
+              echo "PASS: ${image} has no unwaived HIGH/CRITICAL findings"
+            else
+              echo "FAIL: ${image} has unwaived HIGH/CRITICAL findings"
+              overall_rc=1
+            fi
           done
+          exit "$overall_rc"
 
       - name: Upload base image SARIF
         if: ${{ always() && (github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository) }}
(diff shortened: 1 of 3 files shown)


quad341:
@quad341 — good catch, and the diagnosis holds up. Pulling main's scheduled Container Scan run confirms it: under `set -euo pipefail` the enforcement step was bailing after `gc-agent-base` and never scanning `gc-agent`, `gc-controller`, or `gc-mcp-mail` at all. We've been flying blind on three of four images for weeks.

The `if`-wrapper plus `overall_rc` accumulator is the right shape — it fixes the reporting gap without touching severity, `--ignore-unfixed`, or the waiver file, so nothing about the policy gets softer. Your run at this head shows all four images scanned and four `FAIL:` lines before the aggregate exit 1, which is exactly the intended behavior.

Particular thanks for `container_scan_fail_fast_test.go`. Reverting the workflow hunk locally makes every assertion in it fire, so it's a real guard rather than a restatement of the diff — that's what keeps this from quietly regressing the next time someone edits the loop.

Merging as-is. The red `Image vulnerabilities` check is the pre-existing unwaived vulnerability backlog that's been failing on main since mid-August; it isn't required and it isn't yours. Your change is what will finally show us its full scope.


---

PR #5913, opened 2026-09-02
fix(supervisor): guard the supervisor socket path against Unix domain socket length limits (#5845)

### LINKED ISSUE #5845: supervisor: socket path has no length guard, unlike the controller and wake sockets — a long GC_HOME/XDG_RUNTIME_DIR dies with "bind: invalid argument"
## Symptom

`gc start` prints `gc: supervisor did not become ready; see <supervisor.log>`. The log ends with

```
gc supervisor: listening on supervisor socket: listen unix /…/gc/supervisor.sock: bind: invalid argument
```

Nothing says why. The path is over the platform's `sun_path` limit (104 bytes on macOS, 108 on Linux), which `bind(2)` reports as `EINVAL`.

## Root cause

`supervisorSocketPathForDir` (`cmd/gc/cmd_supervisor.go:249-251`) is a bare `filepath.Join(dir, "supervisor.sock")`, and the `net.Listen` at `:626` wraps the error without mentioning the limit. `dir` comes from `supervisor.RuntimeDir()` (`internal/supervisor/config.go:266`): `$XDG_RUNTIME_DIR/gc`, else `GC_HOME`, else `~/.gc` — so an operator-set `GC_HOME` or `XDG_RUNTIME_DIR` of about 90 bytes is enough.

The other two unix sockets already guard this: `controllerSocketPathLimit = 100` (`cmd/gc/controller.go:76`, applied at `:115-119`, sha256 fallback under `/tmp/gascity-controller/`) and `wakeSocketPathLimit = 100` (`internal/nudgequeue/state.go:19`, applied at `:174-183`; its comment cites the 108/104 limits). The supervisor socket is the one that never got the treatment.

## Reproduction

I verified the failure directly on macOS 15: `net.Listen("unix", p)` for a 155-byte `p` returns exactly `listen unix …/supervisor.sock: bind: invalid argument`. I have not run `gc start` under a 90-byte `GC_HOME`; nothing between `RuntimeDir()` and the bind looks at the length, so that is the gc-level trigger:

```
GC_HOME=/tmp/$(printf 'x%.0s' {1..90}) gc start
```

## Environment

upstream `main` at c374d5247 (2026-09-01), macOS 15 / Darwin 25.6; the Linux limit is 108.

## Fix candidates

1. Add `supervisorSocketPathLimit = 100` and reuse the siblings' sha256-fallback shape in `supervisorSocketPathForDir`, relocating an over-long path to `/tmp/gascity-supervisor/<hash>.sock`. About ten lines plus a test; the three sockets then behave the same. Preferred.
2. Check the length before `net.Listen` at `:626` and fail with an error that names the limit and the env var that produced the path. Smaller, but the operator still has to move the directory and the sockets stay inconsistent.

## Adjacent / out of scope

#3452 (closed as a duplicate of #2312) and #2312 are the test-side flakes of the same limit; neither adds a runtime guard. #5770 is a different path-resolution problem (herdr's `os.UserConfigDir`).

rjgeng:
## What

Closes #5845. The controller and wake sockets both already fall back to a sha256-hashed `/tmp` path when their canonical `.gc/*.sock` path would exceed the ~104-108 byte Unix domain socket path limit (macOS/Linux respectively) — the supervisor socket had no equivalent guard, so a long `GC_HOME`/`XDG_RUNTIME_DIR` fails the bind with an opaque `bind: invalid argument` instead of falling back like its two siblings.

## Fix

A one-to-one copy of the existing `controllerSocketPathLimit`/`wakeSocketPathLimit` pattern: `supervisorSocketPathLimit = 100`, falling back to `/tmp/gascity-supervisor/<sha256(dir)[:16]>.sock` when the canonical path exceeds it. The existing `guardSupervisorSocketDir(dir)` call is unchanged.

New tests (`cmd_supervisor_test.go`): short dir unchanged (regression guard), long dir falls back correctly, fallback is deterministic per-dir, and two different long dirs don't collide (needed since `supervisorSocketPathCandidates()` feeds both a live and default path).

## Follow-up fix included in this PR

While validating this change against the full local test suite, two pre-existing tests — `TestRunSupervisorWarnsOnEphemeralAPIPort` and `TestRunSupervisorNoWarningForLowAPIPort` — started hanging indefinitely. Root cause: both tests force `runSupervisor` to fail its control-socket bind fast (a deliberate synchronous fast-exit trick, since neither test otherwise cancels `runSupervisor`'s context) by pre-occupying the canonical `RuntimeDir()/supervisor.sock` path with a directory. Since `t.TempDir()`-based `GC_HOME` routinely produces a canonical path over the new 100-byte limit, the new fallback silently redirected the control socket to a different (short, unoccupied) `/tmp` path — bypassing the tests' directory collision, so the bind succeeded and `runSupervisor` ran forever in its reconcile loop instead of returning.

Fixed by computing `sockPath` in both tests via `supervisorSocketPath()` (the same resolution `runSupervisor` itself uses) instead of hardcoding the canonical path, so the tests' directory-collision setup lands wherever the code will actually bind, fallback or not. Confirmed both tests now pass in <0.1s; confirmed they hang without this fix; confirmed they pass in <0.1s on stock `main` (which never had a fallback to bypass).

## Validation

- `go test -tags gms_pure_go ./cmd/gc/... -run 'TestSupervisorSocket|TestReloadSupervisorFallsBack|TestRunSupervisorWarnsOnEphemeralAPIPort|TestRunSupervisorNoWarningForLowAPIPort'` all pass
- Full local push-gate run: the `cmd/gc` shard containing the affected tests passes clean; remaining failures across other shards and `unit-core` are pre-existing environmental/timing flakes (dolt-compaction timeouts, lsof-based probes, detached-probe/systemctl-timeout tests, process-signal-timing tests) unrelated to this change, confirmed via direct package/test lookup
- Pre-commit lint clean (0 issues) on both commits


quad341 (approved):
@rjgeng — thank you for this, and especially for the second half of it.

The guard itself is exactly right. I checked it against both siblings: `controllerSocketPath` and `WakeSocketPath` use the same limit of 100 and the same `sha256(dir)[:16]` under a hardcoded `/tmp/gascity-<kind>/`, and the supervisor was genuinely the odd one out. Your file is the only place in the tree that builds `supervisor.sock`, so this closes the gap completely. The non-collision test was the smart addition — `supervisorSocketPathCandidates()` calls that function twice with different directories, so that's the case that would have bitten us.

But the part worth calling out is the hang. Noticing two unrelated tests start hanging, then tracing it to them pre-occupying the *canonical* path while the code had quietly begun binding elsewhere, is a subtle diagnosis — and resolving through `supervisorSocketPath()` is the correct fix rather than the convenient one.

Two things on my side, neither needing anything from you: I'll add a `t.Cleanup` to those two tests, since on macOS the fallback path now lands outside `t.TempDir()` and never gets reaped, and that push should also clear the stale watchdog check that went red before the suites finished. Then it goes in. Nice work on #5845.


---

PR #5914, opened 2026-09-02
Make Dolt pull remote selection deterministic and explicit

quad341:
## What this changes

`gc dolt pull` now chooses remotes deterministically instead of relying on SQL or CLI enumeration order. SQL discovery sorts by remote name, while CLI discovery preserves each name/URL pair before selection.

When a database has multiple remotes, operators must select one with `GC_DOLT_REMOTE_<DB>`. Unknown or malformed overrides fail closed. Pulling from any non-`file://` remote—including a sole configured remote—also requires `GC_DOLT_PULL_ALLOW_REMOTE_<DB>=1`; the common sole-local-remote case remains automatic.

## Review notes

- The behavior change is limited to ambiguous or non-local pull targets; local single-remote databases keep their existing zero-configuration path.
- Remote selection and permission to contact a non-local remote are separate checks, so an explicit name alone does not bypass the locality policy.
- The resource-census updates account for the new subprocess-backed coverage and were regenerated in lockstep with `TESTING.md`.

## Test plan

- [x] Run the documented 40-job local CI-equivalent suite with rootless Podman enabled; confirm every changed pull/sync test executes.
- [x] Verify deterministic SQL and CLI selection, missing/invalid/unknown overrides, local preference, and non-local opt-in behavior.
- [x] Run CI policy, affected lint/format, `go vet`, docs sync, and shell syntax checks.
- [x] Release gate: [`release-gates/ga-nht26j-dolt-pull-remote-selection-gate.md`](release-gates/ga-nht26j-dolt-pull-remote-selection-gate.md)


quad341:
@quad341 — the CLI-mode fix is the part worth highlighting here. The old `grep '"name"' | head -1` alongside a separate `grep '"url"' | head -1` could hand back remote A's name paired with remote B's URL; binding them into one extract makes that class of bug structurally impossible rather than merely unlikely. And the decoy `alpha` remote is what sells the tests — without it, "selected the right remote" and "happened to pick row one" look identical.

I'm absorbing two small things on my side before merge, no round-trip needed. First, I ran the script against a multi-remote fixture and a refusal prints a second, contradictory line: `ERROR: failed to query remotes`. `find_remote_sql` returns 1 for both a real lookup failure and a deliberate refusal, so the operator gets told the query broke when the policy actually worked. A distinct status code fixes it. Second, neither new variable appears in the script header or `--help`, so the only way to discover them is to trip the refusal.

One thing for a maintainer glance, not a change request: `compact` in the same pack resolves multi-remote by preferring `origin` with a single global override and no locality gate, so pull's per-database variables are a second convention. Worth settling as a follow-up.


---

PR #5915, opened 2026-09-02
fix(orders): one cron parser, so a range schedule fires instead of silently never firing

### LINKED ISSUE #5709: orders: runtime cron matcher silently drops ranges that gc doctor accepts; no load-time schedule validation

Rome-1:
Fixes #5709.

That issue's own "Suggested fix" is what this does, point for point.

## The two parsers

`internal/orders.cronFieldMatches` (the dispatcher) understood `*`, an integer, a comma list and `*/N`, and silently skipped any part it could not parse. `internal/doctor` grew a second parser that also understood ranges and stepped ranges, bounds-checked every field, and returned errors instead of swallowing them.

So one binary read one schedule two ways. For `*/15 16-23 * * 1-5`, at one commit:

```
internal/doctor  computeExpectedIntervalForCronSchedule -> 15m0s
                 cronScheduleMatchesAt(Wed 18:30)       -> true
internal/orders  checkCron(Wed 18:30)                   -> "cron: schedule not matched"
```

`Validate()` accepted it, because it only checked that `schedule` was non-empty and never read its contents. The order loaded, listed, and reported active, and never fired once. Nothing reported the disagreement: the dispatcher returned "did not match" for input it could not parse, which is the same value a healthy idle order returns.

## What changed

One parser (`internal/orders/cron.go`, exported as `ParseCronSchedule`/`MatchesAt`), both callers pointed at it, the duplicate deleted. Ranges and stepped ranges work. Day-of-week accepts `7` as Sunday — `time.Weekday` never returns 7, so a field written `7` or `1-7` also never matched. An unparseable schedule is now an error at load, beside the existing `tz` and `check_timeout` validation, instead of an order that quietly does nothing.

Parsing also moved out of the matching loops. `checkCron`'s catch-up scan walks minute-by-minute over a bounded lookback, and the doctor's interval scan evaluates up to `366*1440 = 527,040` minutes per schedule; both re-parsed all five fields on every step.

## Behavior changes worth calling out

- **A step now counts from the start of its range.** Identical on minute and hour (minimum 0). On day-of-month and month (minimum 1), `*/2` now means 1,3,5,… as cron does, where the dispatcher gave 2,4,6,…. `internal/doctor` already used the start-of-range reading, so this settles the disagreement rather than creating one. Flagged in #5709 as needing a release note.
- **A malformed schedule now fails at load** instead of loading clean and never firing.
- `1/2` (a step on a single value) and signed numbers like `+5` were accepted and are now rejected. Both parsed clean before and were given a meaning nobody wrote.

## Not changed, deliberately

gc ANDs day-of-month and day-of-week; POSIX and Vixie cron OR them when both are restricted. That is pre-existing in *both* parsers this replaces, so changing it here would be a second behavior change riding along with a fix for something else. It is now documented on `MatchesAt` rather than left to be rediscovered.

## One thing a reviewer should look at

Adding load-time validation means order discovery now *drops* an invalid order, and `gc doctor`'s firing check only `log.Printf`'d such drops. So a city whose only order had an unreadable schedule went from `StatusError: cannot compute expected interval` to `StatusOK: no cron or cooldown orders` — the validation had moved a visible failure into a log line.

The check now reports the orders it dropped, scoped with a new `orders.ErrInvalidSchedule` sentinel so it reports only *schedule* failures, which are its own subject. An order dropped over `[order.env]` stays quiet there, as two existing tests require, because another check owns that.

## Verification

```
go build ./...                                          OK
go vet   ./internal/orders/ ./internal/doctor/          OK
go test  ./internal/orders/ ./internal/doctor/          ok
go test  ./internal/api/ ./internal/config/ \
         ./internal/orderdiscovery/ ./internal/storebinding/... \
         ./internal/webhookmatch/ ./internal/webhooksink/       ok
go test  ./test/docsync                                 ok
```

New tests cover every field form, the step anchor on day-of-month, Sunday as both 0 and 7 including stepped ranges, fourteen malformed schedules, the end-to-end regression, and the doctor reporting a dropped order.

`docs/tutorials/07-orders.md` described the old dialect and is updated.

### DIFF
--- a/docs/tutorials/07-orders.md
+++ b/docs/tutorials/07-orders.md
@@ -173,7 +173,10 @@ Notes per trigger:
   waits `interval` since the last run. Drifts: a 3:02 run means the next is at
   3:07.
 - **`cron`** — a 5-field expression (minute, hour, day-of-month, month,
-  day-of-week) supporting `*`, integers, comma lists (`1,15`), and `*/N` steps.
+  day-of-week) supporting `*`, integers, comma lists (`1,15`), ranges (`16-23`),
+  and steps on either (`*/15`, `1-5/2`). A step counts from the start of its
+  range, so `*/2` on day-of-month is 1,3,5,…. Day-of-week takes 0 or 7 for
+  Sunday. A schedule that cannot be parsed is rejected when the order loads.
   Unlike cooldown it hits the same wall-clock times every day. Fires at most
   once per minute.
 - **`condition`** — the orchestrator runs `sh -c "<check>"` each tick, bounded by
--- a/internal/doctor/checks_order_firing.go
+++ b/internal/doctor/checks_order_firing.go
@@ -1,10 +1,10 @@
 package doctor
 
 import (
+	"errors"
 	"fmt"
 	"log"
 	"path/filepath"
-	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -156,7 +156,7 @@ func (c *OrderFiringCurrentCheck) run(ctx *CheckContext) *CheckResult {
 		return result
 	}
 
-	allOrders, err := scanOrderFiringCurrentOrders(cityPath, c.cfg)
+	allOrders, invalidOrders, err := scanOrderFiringCurrentOrdersWithInvalid(cityPath, c.cfg)
 	if err != nil {
 		result.Status = StatusError
 		result.Message = fmt.Sprintf("scan orders: %v", err)
@@ -189,6 +189,17 @@ func (c *OrderFiringCurrentCheck) run(ctx *CheckContext) *CheckResult {
 	// stay visible without converting an advisory error into a blocking gate.
 	var blockingErrors, advisoryErrors int
 	suspendedRigs := orderFiringCurrentSuspendedRigs(c.cfg, cityPath)
+
+	// An order that failed validation never reaches the loop below — it was
+	// dropped by the scan. Report it here or the check reads clean on a city
+	// whose order cannot load. Counted as monitored so the "no cron or
+	// cooldown orders" early return below cannot swallow it.
+	for _, detail := range invalidOrders {
+		worst = worseStatus(worst, StatusError)
+		result.Details = append(result.Details, detail)
+		blockingErrors++
+		monitored++
+	}
 
 	// Resolve every order-run lookup the loop below will need up front and in
 	// parallel. The pre-pass shares the cron-interval cache with the loop, so
@@ -254,7 +265,11 @@ func (c *OrderFiringCurrentCheck) run(ctx *CheckContext) *CheckResult {
 	case StatusWarning:
 		result.Message = "scheduled orders are overdue"
 	case StatusError:
-		result.Message = "scheduled orders are stale"
+		if len(invalidOrders) > 0 {
+			result.Message = fmt.Sprintf("%d order(s) failed to load", len(invalidOrders))
+		} else {
+			result.Message = "scheduled orders are stale"
+		}
 	}
 	if blockingErrors == 0 && advisoryErrors > 0 {
 		result.Severity = SeverityAdvisory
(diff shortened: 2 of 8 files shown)


Rome-1:
Superseded by #5711, which merged on 2026-09-14 and closed #5709.

These were two branches at the same bug — my duplication, not anyone's review burden, and this one is the one to drop. #5711 has the better shape: it moves the doctor's parser into `internal/orders/cron.go`, deletes the four doctor-local helpers outright, and routes both consumers through one `cronFieldSpecs` table, so the two parsers cannot drift apart again. This branch now conflicts with main and I am not rebasing it.

One case here was not in the merged change: day-of-week `7` as a Sunday alias, which `crontab(5)` defines alongside `0`. Since `Order.Validate()` now hard-errors on out-of-bounds fields, `0 0 * * 7` moved from "loads, never fires" to a load error. I have left the detail on #5711 rather than carrying it here, so it sits where the code that decides it lives.

Closing.

PR #5918, opened 2026-09-02
fix(dispatch): route a class-binding control row to the dispatcher its root ref owns

### LINKED ISSUE #5547: bug: dispatch: class-bound control work is mis-scoped as rig demand
# bug: dispatch: class-bound control work is mis-scoped as rig demand

Labels: `kind/bug`, `priority/p2`, `status/needs-triage`

## Summary

The desired-state control-dispatcher route repair treats a relocated class
binding such as `class:gmnos` as a rig scope. Canonical storeref semantics say
that a class binding is city scope, so valid control-plane work in that binding
loses its city dispatcher demand and no ephemeral city dispatcher wakes.

## Symptom

A city with a configured `core.control-dispatcher` can contain an open
workflow-finalize control bead whose root store ref is `class:gmnos`. The
bead's route is valid, but the desired-state pass reports no city demand. It
instead tries to resolve a dispatcher for a non-existent rig named
`class:gmnos`.

This is a P2 correctness/recovery bug: affected control work can remain open
and undispatched, while a route-repair diagnostic incorrectly describes its
scope. The normal scale-from-zero path is blind to the work.

## Reproduction

Against baseline `HEAD` before applying the candidate,
`ca54c7264ff139b3502064daa430bdb719992d81`, apply the companion regression
file `cmd/gc/build_desired_state_gmnos_test.go` from the companion PR to the
clean checkout. The companion file is intentionally test-only: it links to the
existing `seedSplitRoutes` fixture and `buildDesiredStateWithSessionBeads`
helper in `cmd/gc/build_desired_state_test.go`, so no new harness or production
setup is required. Do not apply the production diff for the RED run. Then run:

```sh
CGO_ENABLED=1 \
CGO_CPPFLAGS='-I/opt/homebrew/opt/icu4c@78/include' \
CGO_LDFLAGS='-L/opt/homebrew/opt/icu4c@78/lib' \
go test ./cmd/gc \
  -run 'TestControlDispatcherRigContextForStoreRefUsesCanonicalScope|TestBuildDesiredStateClassControlWorkWakesExactlyOneCityDispatcher' \
  -count=1 -v
```

The regression fixture creates a city-scoped `core.control-dispatcher`, a
class-binding store seeded by `seedSplitRoutes`, and an open
`gc.kind=workflow-finalize` bead with:

```text
gc.routed_to=core.control-dispatcher
gc.root_store_ref=class:gmnos
```

Before the candidate fix, the direct scope assertion fails with
`controlDispatcherRigContextForStoreRef("class:gmnos") = "class:gmnos", want ""`;
the end-to-end assertion observes city dispatcher demand `0`, rather than one.

Expected: the class route remains canonical, city demand is exactly one, rig
demand is zero, and one ephemeral city dispatcher appears in desired state.

Test obtainability/publication sequence (exact order): (1) create this issue
first; (2) insert the concrete issue number into the PR draft, replacing its
`#<ISSUE_NUMBER>` token, and confirm the local publication gate passes; (3)
create the companion PR containing `cmd/gc/build_desired_state_gmnos_test.go`;
(4) immediately update this Reproduction section with the concrete PR URL and
the file path. Do not leave this Reproduction section without that concrete
companion-PR link after PR creation.

## Root cause

`internal/storeref/storeref.go:39-67` is the canonical scope resolver. Its
documented invariant at `:46-53` is that `class:<token>` is city scope because
the binding serves the whole city and belongs to no rig.

On the baseline, `cmd/gc/build_desired_state.go:5186-5193` only recognized the
empty/city aliases and otherwise stripped `rig:`. Therefore `class:gmnos`
became the rig context `class:gmnos`.

`controlDispatcherRouteRepair.desiredRoute` at
`cmd/gc/build_desired_state.go:5088-5105` uses that context to select the
configured dispatcher. When the lookup misses, it deletes the bead's route
from the in-memory demand snapshot (`:5104`) and suppresses the dispatcher
demand for that tick. The durable route is retained, which is why the symptom
looks like silent non-discovery rather than data loss.

The canonical `storeref.ScopeRigContext` already has tests for class refs in
`internal/storeref/storeref_test.go:147-172`; the missing coverage was the
desired-state route-repair and wake path.

## Prior-art and duplicate search

- Merged PR #4175 (`7d1385958`) restored store-scoped control-dispatcher
  routing, but predates the relocated class scope case.
- Merged PR #4471 covers contextual dispatcher resolution.
- Merged PR #5280 (`b63623d08`) moved census/demand/claim paths onto the
  residency resolver, establishing the direct physical-residency boundary that
  this helper still violated; it did not change this route-repair helper.
- Merged PR #5108 moved control dispatch to the graph store, making the
  logical-vs-physical store distinction observable on the control path.
- Open PR #4452 sweeps rig stores from the city dispatcher; it addresses
  physical store enumeration, not the canonical class-to-city scope mapping.
- Open issue #4710 covers controller bootstrapping of scoped-store metadata;
  it is adjacent store-topology work, not this scope classification bug.
- Open issue #4712 discusses discoverability of HQ-scoped dispatchers for
  graph.v2 roots; it is a documentation/diagnostic feature request, not this
  misclassification. Searches for `class gmnos control dispatcher`,
  `control dispatcher store scope`, and the corresponding PR terms found no
  exact duplicate.
- The accepted/implementing design-doc scan found no design document governing
  this specific control-dispatcher/class-ref invariant. The generic store-scope
  material in `engdocs/design/beads-dolt-contract-redesign.md` does not
  prescribe a different mapping.

## Fix candidates

**A. Resolve canonical refs in the existing helper (recommended)** — delegate
`city:*`, `class:*`, and `rig:*` to `storeref.ScopeRigContext`, then retain
the legacy bare-rig fallback used by older census snapshots.

- Scope: 1 production file, approximately 10 changed lines, plus a focused
  desired-state regression test.
- Trade-off: preserves compatibility with legacy bare rig labels while making
  this helper depend on the already-canonical scope vocabulary.

**B. Remove the helper and pass a typed scope through route repair** — make the
route-repair object consume the canonical resolver result directly, rejecting
unknown refs before dispatcher lookup.

- Scope: multiple route-repair call sites and tests; likely a larger refactor.
- Trade-off: stronger type flow and less duplicated interpretation, but risks
  changing legacy bare-ref behavior and expands the hot desired-state diff for
  a one-branch bug.

**Recommendation:** A. It is the smallest compatible correction, reuses the
existing canonical contract, and leaves rig routing and write/error behavior
unchanged.

## Adjacent / out of scope

Convoy source-workflow cleanup already bridges logical refs to the graph/class
binding and passes `TestSourceWorkflowStoresScanTheLedgerThatHoldsWorkflowRoots`
(`cmd/gc/control_dispatch_graph_store_test.go:599-656`). No equivalent failure
was isolated there.

The formula/`gc.routed_to` drift checks in
`cmd/gc/doctor_routed_to_checks.go` use a separate raw city/rig store scan and
do not call this helper. That may warrant a separate class-binding regression,
but no failing isolated test was found, so it is deliberately not included in
this candidate.

## Candidate evidence

- Baseline `HEAD` before applying the candidate:
  `ca54c7264ff139b3502064daa430bdb719992d81`.
- RED focused run: failed on the class scope assertion and city demand `0`.
- GREEN focused run with Homebrew ICU flags: passed, including city/class/rig
  and legacy bare-rig table cases.
- Existing `cmd/gc` control-dispatcher/source-workflow regression selection:
  passed.
- `internal/storeref`, `internal/dispatch`, and `internal/sourceworkflow`:
  passed.
- `gofmt` and `git diff --check`: clean.

### LINKED ISSUE #5587: class-binding candidates rejected for rig-rooted control rows, dropped before route repair/demand

julianknutsen:
# fix(dispatch): route a class-binding control row to the dispatcher its root ref owns

Supersedes #5548 and #5588 (both carried here as their original commits, unchanged). Fixes #5547, fixes #5587. Refs `ga-oytw9`.

## What is broken on main

Since #5280 (residency S3) a relocated class binding is its own census leg with ref `class:<token>`. Two helpers predate that:

- `controlDispatcherRigContextForStoreRef` stripped only `rig:`, so `class:gmnos` became a phantom rig and every city-rooted control row the binding serves logged `no configured control-dispatcher for its store scope` once per tick and lost its demand. This is the line maintainer-city has been printing since its last deploy (71 distinct beads, one line per tick).
- `rootStoreRefMatchesCandidate` required the candidate's rig to equal the root's rig, so a rig-rooted row relocated into the binding was dropped before repair or demand ever saw it (#5587).

## What #5548 + #5588 get wrong

They fix both helpers, but together they re-scope **every** row the binding serves onto the city dispatcher: the repair keyed its dispatcher lookup on the physical leg, and a binding reads as city scope. #5588's own test asserts that a rig-rooted row on `fixture/core.control-dispatcher` gets rewritten to `core.control-dispatcher`. On maintainer-city that hands all rig control work to a city dispatcher `city.toml` has suspended and strips the four rig dispatchers that have served those rows since the 07-26 relocation.

Two consumers already settle which scope owns the row:

- **The minter keys on the root ref.** `graphroute.DecorateGraphWorkflowRecipeWithDefaultBinding`, `dispatch.applyAttemptControlStepRoute` and `decorateDynamicFragmentRecipe` all feed `config.ControlDispatcherForScope` with `storeref.ScopeRigContext(rootStoreRef)`. A leg-keyed repair rewrites on every tick what the minter just stamped.
- **The serve side refuses foreign-rooted rows.** A drain rooted in `rig:x` dispatched from the city fails `assertDrainRootScopeMatchesDispatch`, and a rig-rooted control's input convoy lives in the rig work store the city dispatcher never opens (`controlBeadLedger`). The city rewrite would have the city dispatcher claim those rows and fail them.

## The fix

`repairBead` parses `gc.root_store_ref` once and `desiredRoute` takes that scope:

- through a legacy file store it agrees with the leg the row was collected through (`rootStoreRefMatchesCandidate` admits nothing else), so no legacy repair changes;
- through a binding it is the only thing that still says which dispatcher owns the row: rig rows keep, or are repaired toward, their rig dispatcher; city- and class-rooted rows go to the city dispatcher.

`controlDispatcherRigContextForStoreRef` and `controlDispatcherStoreRefLabel` are retired. The missing-dispatcher diagnostic names the binding and the owning scope (`class binding "class:gmnos" serving rig "beads"`).

**Behavioural delta beyond the fix:** a city whose rig has no configured dispatcher moves from silently dropping that rig's relocated control rows to one missing-dispatcher diagnostic per rig per tick. The rows were unservable either way. Every maintainer-city rig with rows has a dispatcher.

## Tests

`cmd/gc/build_desired_state_test.go`:

- `TestBuildDesiredState_ClassBoundRigRootedControlWorkStaysOnRigDispatcher` — rig-rooted binding row keeps its rig route, wakes only the rig dispatcher.
- `TestBuildDesiredState_RepairsClassBoundRigRootedControlWorkToRigDispatcher` — a stale city route on such a row is repaired toward the rig dispatcher, fallback marker cleared.
- `TestBuildDesiredState_ClassBoundRigRootedControlWorkWithoutRigDispatcherIsNotReScopedOntoCity` — no rig dispatcher configured: durable route kept, city dispatcher not woken, diagnostic names binding + rig.
- `TestBuildDesiredState_ClassBoundCityRootedControlWorkWakesCityDispatcher` — roots `city:test-city` and `class:gmnos` wake exactly one city dispatcher (the shape maintainer-city's logged rows have).
- `TestControlDispatcherRouteRepairAgreesWithTheMinter` — drives the real repair over a binding for city-, rig- and class-rooted rows and requires each converged route to equal what `graphroute.ControlDispatcherBinding` stamps for the same root. `internal/config` promises this lockstep; nothing enforced it.

`cmd/gc/census_residency_test.go`: `TestBindingRefNormalizesOntoTheCityScope` keeps #5588's corrected invariant and gains the control that a legacy `city:` candidate still defers to a rig root, so the binding carve-out cannot be widened to "any city-scope candidate" (that mutation passed every other test).

All rig-rooted tests fail with only the two carried commits applied. `build_desired_state_gmnos_test.go` is folded into the above and removed.

## Verification

- `go build ./cmd/gc`, `go vet ./cmd/gc` clean; `golangci-lint run ./cmd/gc/...` 0 issues
- `make test-fast-parallel` green (all cmd/gc shards + core)
- 54-agent adversarial review (5 lenses, 3 refuters per finding): design verdict for the root-ref owner; 5 findings kept, all pre-existing or addressed here; follow-ups filed as `ga-erfca` (continuation backstop folds class legs to city), `ga-qbexl` (unrouted control rows are unrecoverable), `ga-8lodx` (rig dispatcher scale probe never reads the binding), `ga-ods2a` (detector should be an event with a count).

## What this does not do

- It does not lift maintainer-city's `suspended = true` on the city dispatcher; the 71 city-rooted rows stay parked until an operator does, and the beads-rig PR stall on mc is dep-blocked upstream of dispatch, not this bug. Config, not SDK.
- It does not turn the detector's stderr line into an event (`ga-ods2a`); `buildDesiredState` emits no events today.


quad341 (approved):
@julianknutsen — thank you for taking #5548 and #5588 apart instead of stacking on them. Catching that they're each right alone but *together* re-scope every binding-served row onto the city dispatcher only shows up if you read what the combined diff does to a real city, and writing "#5588's own test asserts the wrong invariant" about work you're building on is genuinely uncomfortable. You did it clearly and carried both commits unchanged so the original authors keep their credit.

`TestControlDispatcherRouteRepairAgreesWithTheMinter` is my favourite thing here — comparing the repaired route against what `graphroute.ControlDispatcherBinding` actually stamps, rather than a hardcoded string, turns a promise in `internal/config` into an enforced property. And the `city:test-city` control you added because the widened mutation passed everything else is exactly the right instinct.

I traced the carve-out through `Narrow` and confirmed it's safe — the runtime-plane call site really is the only one. One thing I'll take on my side: the doc comment justifies it partly with "no legacy store holds the row post-migration," but `infra_class_migrate.go` is explicit that the source is retained. Behaviour is fine; I'll reword it in the helper and the census test so nobody later reads it as licence to widen this onto the reconcile plane. Merging after that, and closing #5548/#5588 as superseded.


---

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277283776

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277284913

quad341 (dismissed):
@julianknutsen — pulling #5548 and #5588 apart instead of stacking on them was the right call, and the reasoning is what makes this mergeable. Each is correct alone, but together they key the repair on the physical leg, and a binding reads as city scope, so every rig row the binding serves gets handed to a city dispatcher `city.toml` has suspended. That only surfaces if you read the combined diff against a real city, which you did.

`TestControlDispatcherRouteRepairAgreesWithTheMinter` is the piece I'd keep even if everything else changed — comparing against what `graphroute.ControlDispatcherBinding` actually stamps turns a promise in `internal/config` into something that fails loudly instead of quietly churning routes every tick. And adding the `city:test-city` control specifically because the widened mutation passed everything else is exactly the instinct I want in that file.

I traced the carve-out's safety argument myself: one call site, runtime plane, `Narrow` drops every non-class leg. It holds, and your comment already records the reconcile-plane hazard and the retained-source contract, so there's nothing left on my side.

Merging as-is. I'll close #5548 and #5588 as superseded, with credit to their authors for the commits carried here unchanged. Thanks for the care on this one.


---

quad341 (approved):
Re-reviewed `b1c97d18d`..`414492952` (10 new commits since the last review): no new concerns — the prior review stands.

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR fixes a dispatch issue by routing a class-binding control row to the dispatcher its root ref owns, which keeps the dispatch area easier to maintain and verify.

### Original PR Review

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `414492952fb5` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

### Review Iterations

---

PR #5921, opened 2026-09-02
fix(tmux): back off the process snapshot instead of re-forking it every refresh

remuscazacu:
## The loop

The full-OS `ps` scan behind the tmux state cache fails by losing a CPU race: the box is saturated, the scan does not finish inside the 3s `fetchTimeout`, and the context kills it (`signal: killed`).

`FetchState` degrades rather than erroring — it keeps the authoritative tmux liveness and marks process detail unavailable. That is correct and deliberate, and this PR does not change it.

The consequence is that `refresh()` books the fetch as a **success**: it stamps `fetchedAt`, clears `dirty`, and has no way to know the expensive half just died. Nothing remembers the failure, so the very next refresh re-forks the same full-OS scan into the same contention that just starved it. A loaded box keeps paying `fetchTimeout` plus a whole-process-table scan for a probe that cannot succeed while it stays loaded — and the paying is part of why it stays loaded.

## Observed

A 4-core / 15.7 GB workstation running two cities and seven rigs, supervisor up 7d10h:

- `process snapshot degraded` **74/hr** — one futile scan every ~48s
- load average 17–22, 0.0% idle, every liveness probe failing
- the probe never recovered on its own; only a supervisor restart cleared it
- stopping the supervisor dropped load to 3.96 and idle to 81%, attributing it to this process rather than to the other tenants on the box

## The change

A gate on `tmuxFetcher` bounds re-attempts: 15s after the first failure, doubling to a 2 minute cap, cleared on the first success. While closed, `FetchState` skips the scan and returns the same degraded answer it would otherwise have spent `fetchTimeout` to reach.

- **Silent while skipping** — a line per skip would only move the cost from CPU to the log.
- **Recovery logged once**, not on every healthy refresh thereafter.
- **The degrade log now names the window**, so the backoff is visible where the failure already was.

It does not weaken the answer. A failed snapshot already degrades optimistically in `processAlive`, on the rule that a failed secondary probe must never trigger a reap or respawn. Holding that state longer is the same answer held longer, while the machine gets back the CPU that lets the next attempt finish.

The gate lives on `tmuxFetcher` rather than `StateCache` because the cache cannot see this failure at all — by the time `refresh()` looks, the degrade has already been rendered as a success. Its zero value attempts immediately, so the several `&tmuxFetcher{tm: tm}` literals across the package and its tests behave exactly as before.

## Tests

New `state_cache_snapshot_backoff_test.go`: zero value allows; window opens/closes on schedule; doubling to the cap; success clears and reports recovery exactly once; `FetchState` skips the scan while gated **and retains session liveness**; a skipped snapshot degrades optimistically like a failed one.

`go test ./internal/runtime/tmux/` passes, `go vet` clean, `golangci-lint run ./internal/runtime/tmux/...` reports 0 issues.

### DIFF
--- a/internal/runtime/tmux/state_cache.go
+++ b/internal/runtime/tmux/state_cache.go
@@ -29,6 +29,27 @@ const defaultStaleTTL = 30 * time.Second
 // fetchTimeout is the hard timeout for a single runtime-state fetch.
 const fetchTimeout = 3 * time.Second
 
+// Backoff bounds for the process-table snapshot after it fails.
+//
+// The snapshot is a full-OS `ps` scan, and it fails by losing a CPU race: the
+// box is saturated, the scan does not finish inside fetchTimeout, and the
+// context kills it. Re-attempting it on the very next refresh spends another
+// fetchTimeout and forks another full-OS scan into the contention that just
+// starved the last one, so a loaded box keeps paying for a probe that cannot
+// succeed while it stays loaded. That is a feedback loop, and it was observed
+// running for days: a supervisor logged the degrade 74 times an hour, one
+// futile scan every ~48s, and the probe never recovered on its own until the
+// process was restarted.
+//
+// Backing off does not weaken the answer. A failed snapshot already degrades
+// optimistically (see processAlive), and holding that degraded state longer is
+// the same answer held longer — while the machine gets the CPU back that lets
+// the next attempt actually finish.
+const (
+	processSnapshotBackoffBase = 15 * time.Second
+	processSnapshotBackoffMax  = 2 * time.Minute
+)
+
 // StateFetcher abstracts tmux subprocess calls for testability.
 type StateFetcher interface {
 	// FetchState returns a runtime-state snapshot for live sessions.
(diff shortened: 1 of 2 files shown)


quad341 (approved):
@remuscazacu — thank you for this one. The diagnosis is the valuable half: spotting that `refresh()` books a degraded fetch as a *success*, so nothing upstream even knows there is a failure to back off from, is not something you find by reading the code cold. You found it on a real box — 74 degrades an hour, load 17–22, recovering only on restart — and then put the gate at the only layer that can see the failure, with a zero value that keeps every existing `&tmuxFetcher{tm: tm}` literal behaving exactly as before. That care shows, and the tests pin the property that would have been catastrophic to get wrong.

Two things on our side, neither of them yours. This package keeps a canonical test manifest plus some hard-coded inventory counts, and new tests have to be registered there — an internal ratchet you had no way to know about, and the actual reason CI is red. The residency-boundary failure is just drift from #5597 landing after you cut your branch; a rebase clears it. I will take both, re-run, and merge with your commit intact.

I ran your six tests locally in the meantime — all green.


---

Issue #5804, opened 2026-08-31
CI: two distinct rest-smoke infra flakes in one night (listing pipe truncation; bd bootstrap in gc init)


A3Ackerman:
Two different infrastructure failures took down `Integration / rest-smoke` shards on unrelated PRs tonight (2026-08-31), both unreproducible locally and both clearing on a plain retrigger. Filing them together because they share a blast radius: one flaked shard fails `CI / integration` → `CI / required` → `Evidence / critical-path suites`, so a PR shows 4 red checks for one infra hiccup.

## Shape 1 — test-listing pipe breaks, manifest check reports a false "missing test"

PR #5797, run 33369465852, job `Integration / rest-smoke-1-of-2`:

```
./scripts/test-integration-shard: line 204: printf: write error: Broken pipe
missing integration test for shard rest-smoke-1-of-2: TestE2E_WorkspaceDefaults
```

`TestE2E_WorkspaceDefaults` exists on the branch (`test/integration/e2e_config_test.go:11`), and the same shard passes locally (`./scripts/test-integration-shard rest-smoke-1-of-2` → exit 0). In `validate_selected_tests`, `available="$(list_integration_tests)"` appears to have been truncated when the `printf '%s\n' "$out" | grep '^Test'` pipeline broke, so a present test was reported missing. A `set -o pipefail`-style guard on the listing pipeline (or retrying the listing when the pipe breaks, as the existing "listing integration tests failed twice" guard already does for `go test -list` failures) would turn this into a retry instead of a false red.

## Shape 2 — `gc init` fails in bd bootstrap on the runner

PR #5799, run (job `Integration / rest-smoke-2-of-2`):

```
--- FAIL: TestGraphWorkflowSuccessPath (8.90s)
    graph_dispatch_test.go:62: gc init --file failed: exit status 1
        output: ...
        gc init: exec beads init: gc-beads-bd: setting git config --global beads.role maintainer
        warning: database 'hq' missing bd schema; re-initializing
        Thanks for using bd! Quick heads-up: bd shares anonymous usage metrics — ...
```

The failure is in test setup (`gc init`'s bd bootstrap writing `git config --global beads.role`), before the PR's diff (order-tracking sweep readers) can execute. The bd first-run metrics notice in the output suggests the runner image hit bd's first-run path mid-test.

## Impact

- Both PRs went fully green on retrigger (#5797: 0 failing / 88 after retrigger).
[...]

PR #5924, opened 2026-09-02
Grow the usage scan window until the cursor is reachable

### LINKED ISSUE #5922: Model-usage extraction is capped at a 64KB tail with an identity-only cursor: ~78% of invocations unrecoverable (gc costs read 0.2% of true spend)
## Summary

Every model-usage extraction path is hard-capped at a **64KB transcript tail**, while the dedup cursor is a *message identity* rather than a byte offset. A pass that falls behind therefore cannot reach back for what it missed, and the invocations in between are lost permanently.

On a real three-week single-rig build this made `gc costs` report **8 invocations (~646K tokens)** against a measured **~372M tokens** actually consumed — roughly **0.2%** of true spend.

This is distinct from #4173. That issue is "an entire session class never reaches the instrumented seam." This one applies to sessions that *do* reach it, including via the newer `SweepSessionModelUsage` controller tick: the seam runs, finds the cursor has scrolled out of the window, and silently drops the backlog.

## Measurement

Across 152 local Claude Code transcripts from the affected city:

| | |
|---|---|
| Distinct API invocations in the full files | 4,014 |
| Visible in the last 64KB | 881 (21.9%) |
| **Invisible to a 64KB tail scan** | **3,133 (78.1%)** |
| Mean transcript bytes per invocation | 15.6 KB |
| **Invocations that fit in one 64KB window** | **~4** |

Median transcript 253 KB, largest 2.5 MB. At ~15.6 KB per invocation the window holds about four; a session in a tool loop clears that in seconds.

## Mechanism

`ExtractTailUsage` reads a fixed window and says so in its own contract ([`internal/sessionlog/tail_usage.go`](https://github.com/gastownhall/gascity/blob/main/internal/sessionlog/tail_usage.go)):

> The scan window is the last tailChunkSize bytes, so usage that scrolled past the window is not returned.

`tailChunkSize = 64 * 1024`. Both consumers inherit the cap:

- **Prompt-op seam** — `usagesAfterCursor` on a cursor miss returns `usages[len(usages)-1:]`, i.e. exactly one invocation. Its doc concedes "possible undercounting."
- **Controller tick** — `usagesSinceCursor` returns the whole window, but "bounded only by the extractor's tail window," and `liveModelSweepMinInterval = 30 * time.Second`.

So the recovery ceiling is ~4 invocations per 30s tick, and 1 per prompt op that falls behind. Anything above that rate is lost, silently and permanently, because `MetadataKeyInvocationUsageCursor` stores a message id — there is no offset to resume from.

This also diverges from the design doc, [`engdocs/design/usage-facts-v0.md`](https://github.com/gastownhall/gascity/blob/main/engdocs/design/usage-facts-v0.md), which specified:

> On startup, reconcile from the durable cursor, not the 64KB tail.

The durable cursor shipped; the reconcile-from-cursor behavior did not.

## Proposed fix

Grow the scan window until the cursor entry is inside it. This needs **no new persisted state** — it uses the cursor already on the session bead, and the caller's existing filter still decides what is new.

Adds `ExtractTailUsageSince(path, cursorID)`, which doubles the window from `tailChunkSize` until the cursor is visible, the whole file is in view, or a 16 MB cap is hit. Behavior is deliberately conservative at the edges:

- **empty cursor** (no prior extraction) → single fixed window, identical to today; first sight of a transcript does not backfill history
- **stale/unreachable cursor** → widest scanned window rather than nothing, degrading to today's behavior instead of failing

Wiring is then one line in `invocationUsageSpecs["claude"].extract` to pass the cursor through `SessionLogAdapter.TailUsage`. Widening the window is safe for the sweep by its own stated invariant — `usagesSinceCursor` already notes that read-time `IdempotencyKey` dedup in `usage.ReadFacts` collapses overlap, "so recovering the whole bounded tail cannot double-count."

The same cap applies to `ExtractCodexTailUsage`; the codex side would want the equivalent treatment.

<details>
<summary>Patch against <code>internal/sessionlog/tail_usage.go</code> (gofmt clean)</summary>

```diff
diff --git a/internal/sessionlog/tail_usage.go b/internal/sessionlog/tail_usage.go
index e11c428..7117ec0 100644
--- a/internal/sessionlog/tail_usage.go
+++ b/internal/sessionlog/tail_usage.go
@@ -60,7 +60,83 @@ func ExtractTailUsage(path string) ([]TailUsage, error) {
 	if err != nil {
 		return nil, err
 	}
+	return parseTailUsage(data), nil
+}
+
+// maxUsageScanBytes caps how far ExtractTailUsageSince will grow its scan
+// window when it cannot reach the cursor entry. It bounds the worst case (a
+// long-lived transcript whose cursor is stale or absent) while still covering
+// transcripts orders of magnitude larger than the fixed tail window.
+const maxUsageScanBytes = 16 * 1024 * 1024
+
+// ExtractTailUsageSince returns usage-bearing invocations from the tail of a
+// transcript, growing the scan window until the entry identified by cursorID
+// is inside it — so no invocation is skipped merely because more than
+// tailChunkSize bytes were appended since the last extraction.
+//
+// The fixed-window ExtractTailUsage silently drops any invocation that
+// scrolled past the last 64KB between two extractions. The persisted cursor
+// (session.MetadataKeyInvocationUsageCursor) is a message identity, not a byte
+// offset, so a later pass cannot reach back for what it missed and the usage
+// is lost permanently. Growing the window until the cursor is visible closes
+// that gap without introducing new persisted state: the caller's existing
+// cursor filter still decides what is new.
+//
+// The window doubles from tailChunkSize until the cursor entry is found, the
+// whole file is in the window, or maxUsageScanBytes is reached. An empty
+// cursorID (a session with no prior extraction) reads a single tailChunkSize
+// window, matching ExtractTailUsage: there is no anchor to reach back to, and
+// backfilling a full transcript on first sight is not this function's job.
+// Callers may receive entries at or before the cursor and must still filter.
+func ExtractTailUsageSince(path, cursorID string) ([]TailUsage, error) {
+	f, err := os.Open(path)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close() //nolint:errcheck // best-effort close on read-only file
+
+	if cursorID == "" {
+		data, _, err := readTail(f, tailChunkSize)
+		if err != nil {
+			return nil, err
+		}
+		return parseTailUsage(data), nil
+	}
+
+	for window := int64(tailChunkSize); ; window *= 2 {
+		data, _, truncated, err := readTailWindow(f, window)
+		if err != nil {
+			return nil, err
+		}
+		usages := parseTailUsage(data)
+		// Reached the cursor, or the whole file is in view: nothing older can
+		// still be owed. Either way this window is complete.
+		if !truncated || containsCursor(usages, cursorID) {
+			return usages, nil
+		}
+		if window >= maxUsageScanBytes {
+			// Cap hit with the cursor still out of view: return the widest
+			// window rather than nothing, so a stale cursor degrades to the
+			// old bounded behavior instead of losing everything.
+			return usages, nil
+		}
+	}
+}
+
+// containsCursor reports whether any extracted invocation carries the cursor
+// identity, under the same message-id-else-entry-uuid rule the telemetry
+// cursor is written with.
+func containsCursor(usages []TailUsage, cursorID string) bool {
+	for _, u := range usages {
+		if u.MessageID == cursorID || u.EntryUUID == cursorID {
+			return true
+		}
+	}
+	return false
+}
 
+// parseTailUsage extracts invocations from an already-read transcript window.
+func parseTailUsage(data []byte) []TailUsage {
 	var usages []TailUsage
 	// byMessageID maps a message identity to its index in usages so the
 	// content-block copies of one API response collapse to a single entry.
@@ -101,7 +177,7 @@ func ExtractTailUsage(path string) ([]TailUsage, error) {
 		}
 		usages = append(usages, u)
 	}
-	return usages, nil
+	return usages
 }
 
 // ExtractTailUsageFromSearchPaths reads tail usage only after verifying
```

</details>

## Tests

Three tests added to `internal/sessionlog/tail_usage_test.go`. The first asserts the defect is real before asserting the fix — it fails the build if a fixed window can still reach the cursor, so it cannot silently rot into a tautology.

```
=== RUN   TestExtractTailUsageSinceReachesBackPastTheFixedWindow
--- PASS: TestExtractTailUsageSinceReachesBackPastTheFixedWindow (0.04s)
=== RUN   TestExtractTailUsageSinceEmptyCursorMatchesFixedWindow
--- PASS: TestExtractTailUsageSinceEmptyCursorMatchesFixedWindow (0.00s)
=== RUN   TestExtractTailUsageSinceUnreachableCursorDegradesToBoundedWindow
--- PASS: TestExtractTailUsageSinceUnreachableCursorDegradesToBoundedWindow (0.01s)
PASS
ok  	github.com/gastownhall/gascity/internal/sessionlog	0.566s
```

Full `./internal/sessionlog/` suite passes. I could **not** build `./internal/worker/` locally — `github.com/dolthub/go-icu-regex` needs ICU headers not present on this machine — so the call-site wiring above is reasoned from the source, not compiled. Happy to open this as a PR if useful.

## Impact

Any city relying on `gc costs`, `GET /v0/city/{name}/usage`, or the dashboard burn gauges is reading a number that can be off by two to three orders of magnitude, with no indication that data is missing. That silence is the harmful part: a visible gap prompts investigation, a plausible small number does not. It also blocks #115 (token budget circuit breaker) — a breaker driven by this data would never trip.

*Found while reconstructing spend for a completed build. Reconstructing it required parsing the Claude Code transcripts directly, which is the same data the extractor already reads — just without the window cap.*

TheChrisOneil:
## Summary

Fixes #5922.

Model-usage extraction reads a fixed 64KB transcript tail, but the persisted dedup cursor (`MetadataKeyInvocationUsageCursor`) is a **message identity with no byte offset**. A pass that falls more than one window behind cannot reach back for what it missed, so those invocations are lost permanently instead of being recovered on a later tick.

At a measured **~15.6KB per invocation** on real transcripts, one window holds about **four** invocations. With `liveModelSweepMinInterval` at 30s, any session busier than that silently under-reports. On the 152-transcript sample in #5922, **78% of invocations sat outside a 64KB tail**, and `gc costs` reported ~0.2% of true spend.

Both consumers inherit the cap:

- `usagesAfterCursor` (prompt-op seam) falls back to `usages[len(usages)-1:]` — one invocation
- `usagesSinceCursor` (controller tick) returns the window, "bounded only by the extractor's tail window"

This also restores the behavior [`engdocs/design/usage-facts-v0.md`](https://github.com/gastownhall/gascity/blob/main/engdocs/design/usage-facts-v0.md) specified — *"On startup, reconcile from the durable cursor, not the 64KB tail."* The durable cursor shipped; the reconcile did not.

## Approach

`ExtractTailUsageSince` doubles the scan window from `tailChunkSize` until the cursor entry is inside it, the whole file is in view, or a 16MB cap is reached. **No new persisted state** — it uses the cursor already on the session bead, and callers still filter. Both extraction sites now read the cursor *before* extracting and thread it through.

Edges stay deliberately conservative:

| Case | Behavior |
|---|---|
| Empty cursor (no prior extraction) | single fixed window, exactly as today — first sight does not backfill history |
| Cursor unreachable / stale | widest scanned window, degrading to today rather than failing |
| Cap reached | returns what was scanned; bounded worst case |

Widening is safe for the sweep by its own stated invariant: `usagesSinceCursor` already notes that read-time `IdempotencyKey` dedup in `usage.ReadFacts` collapses overlap, "so recovering the whole bounded tail cannot double-count."

One incidental refactor: with the varying-window caller now using `readTailWindow` directly, `readTail`'s size parameter only ever receives `tailChunkSize`, which `unparam` flags. The constant is folded into `readTail` and its four callers simplified — required to keep `make lint` green, and it removes a knob nothing turned.

**Codex is intentionally left on the fixed window.** Its extractor collapses on cumulative totals rather than per-message identity, so window growth needs its own correctness argument. `CodexTailUsage` takes the cursor for signature symmetry and documents that it does not honor it.

## Testing

`make check` was run with the pinned golangci-lint 2.12.0. Results on macOS (arm64):

| Gate | Result |
|---|---|
| `fmt-check` | pass |
| `lint` | **0 issues** |
| `vet` | pass |
| `check-release-dist-ignore` | pass |
| `check-routed-test-rows` | pass |
| `check-split-topology-rows` | pass |
| `check-residency-boundary` | **blocked locally** — the script uses `declare -A`, which needs bash 4+; macOS ships 3.2. Fails identically on clean `main`. |
| `test` | green except two pre-existing environmental failures (below) |

Two `make test` failures reproduce **identically on clean upstream `main` (1154ad1)** with this branch absent, so they are not from this change:

- `examples/bd/dolt` `TestDoctorCheckVersionFloor*` — `unable to run dolt version`; the `env -i` PATH in the `test` target does not include `dolt`
- `scripts` `TestCmdGCProcessTimingEnvCrossesMakeIsolation`

The packages this PR touches are green: `internal/sessionlog`, `internal/worker`, `internal/worker/workertest`.

- [x] `make check` (with the two environmental exceptions above, both verified pre-existing)
- [ ] `make check-docs` — not applicable, no docs touched
- [ ] `make test-integration` — **not run**, no working integration environment here. The change is confined to transcript extraction and its two call sites, but the sweep is controller-adjacent, so please exercise it in CI.

### New tests

Four added. Each asserts the defect is present **before** asserting the fix, so they fail loudly if a future change makes the fixed window sufficient rather than quietly passing:

| Test | Pins |
|---|---|
| `TestExtractTailUsageSinceReachesBackPastTheFixedWindow` | window grows to reach a cursor many windows back |
| `TestExtractTailUsageSinceEmptyCursorMatchesFixedWindow` | no cursor ⇒ unchanged behavior |
| `TestExtractTailUsageSinceUnreachableCursorDegradesToBoundedWindow` | stale cursor ⇒ bounded, not empty |
| `TestFactorySweepRecoversBacklogBeyondFixedTailWindow` | end-to-end: sweep drains an 800KB backlog to the sink |

Verified they actually catch the defect by reverting the fix in place:

```
--- FAIL: TestFactorySweepRecoversBacklogBeyondFixedTailWindow
    emitted = 3, want 40 (the whole backlog past the fixed window)
```

3 of 40 — matching the ~4-per-window figure from the measurement.

One more flake worth flagging rather than burying, given #5804: under heavy parallel/cold-build load I twice saw `internal/worker/adapters/zcode` `TestInterrupt*` fail, with a different timing-shaped symptom each time (once "adapter exited on an idle SIGINT", once a truncated prompt `"still workin"` vs `"still working"`). Warm runs were 5/5 green both with and without this change, and that package imports none of the changed code.

## Checklist

- [x] Linked an issue (#5922)
- [x] Added tests for behavior changes
- [ ] Docs — no user-facing surface changed; `gc costs` output becomes correct, not different in shape
- [x] No breaking changes. `SessionLogAdapter.TailUsage`, `CodexTailUsage`, and `InvocationUsage` gain a `cursorID` parameter; all in-tree callers are updated, and passing `""` reproduces the previous behavior exactly.

### DIFF
--- a/internal/sessionlog/codex_usage.go
+++ b/internal/sessionlog/codex_usage.go
@@ -55,7 +55,11 @@ func ExtractCodexTailMeta(path string) (*TailMeta, error) {
 	if err != nil {
 		return nil, err
 	}
-	return extractCodexTailMetaFromLines(splitLines(data), startsMidLine, truncated), nil
+	lines, err := splitLines(data)
+	if err != nil {
+		return nil, err
+	}
+	return extractCodexTailMetaFromLines(lines, startsMidLine, truncated), nil
 }
 
 // ExtractCodexTailMetaFromSearchPaths reads Codex tail metadata only after
@@ -289,12 +293,17 @@ func ExtractCodexTailUsage(path string) ([]TailUsage, error) {
 		return nil, err
 	}
 
+	lines, err := splitLines(data)
+	if err != nil {
+		return nil, err
+	}
+
 	var usages []TailUsage
 	// byMessageID maps a cumulative-total identity to its index in usages so
 	// duplicate token_count emissions collapse to a single entry.
 	byMessageID := make(map[string]int)
 	turnModel := seedModel
-	for _, line := range splitLines(data) {
+	for _, line := range lines {
 		var entry codexRawEntry
 		if err := json.Unmarshal(line, &entry); err != nil {
 			continue
--- a/internal/sessionlog/tail.go
+++ b/internal/sessionlog/tail.go
@@ -43,7 +43,7 @@ func ExtractTailMeta(path string) (*TailMeta, error) {
 	}
 	defer f.Close() //nolint:errcheck // best-effort close on read-only file
 
-	data, startsMidLine, err := readTail(f, tailChunkSize)
+	data, startsMidLine, err := readTail(f)
 	if err != nil {
 		return nil, err
 	}
@@ -87,9 +87,9 @@ func validateSearchPathFile(searchPaths []string, path string) (string, error) {
 	return "", fmt.Errorf("session log path is outside configured search paths")
 }
 
-// readTail reads the last n bytes of r (or the whole thing if smaller).
-func readTail(r io.ReadSeeker, n int64) ([]byte, bool, error) {
-	data, startsMidLine, _, err := readTailWindow(r, n)
+// readTail reads the last tailChunkSize bytes of r (or the whole thing if smaller).
+func readTail(r io.ReadSeeker) ([]byte, bool, error) {
+	data, startsMidLine, _, err := readTailWindow(r, tailChunkSize)
 	return data, startsMidLine, err
 }
 
--- a/internal/sessionlog/tail.go
+++ b/internal/sessionlog/tail.go
@@ -48,7 +48,10 @@ func ExtractTailMeta(path string) (*TailMeta, error) {
 		return nil, err
 	}
 
-	lines := splitLines(data)
+	lines, err := splitLines(data)
+	if err != nil {
+		return nil, err
+	}
 	return extractFromLines(lines, startsMidLine), nil
 }
 
@@ -139,11 +142,30 @@ func readTailWindowAt(r io.ReadSeeker, size, n int64) ([]byte, bool, bool, error
 	return data, startsMidLine, offset > 0, nil
 }
 
+// maxTailTokenBytes bounds a single JSONL line during a tail scan. Claude
+// entries can be large (tool results with full file contents, base64 images),
+// so the tail path uses the same 50MB ceiling the full-file reader adopted for
+// exactly that reason (parseFileDetailed).
+//
+// No current caller can reach this ceiling. The fixed-window paths scan
+// tailChunkSize (64KB), and ExtractTailUsageSince's growth loop stops at
+// maxUsageScanBytes (16MB), so the widest window any caller hands to splitLines
+// is three times below this bound and cannot hold a token that exceeds it. The
+// error return below is defense-in-depth against a future cap change, not a
+// live path — maxUsageScanBytes documents what raising the growth cap past this
+// constant would cost.
+const maxTailTokenBytes = 50 * 1024 * 1024
+
 // splitLines splits data into JSONL lines. Partial lines from a mid-file
 // read are tolerated — they fail json.Unmarshal silently in the caller.
-func splitLines(data []byte) [][]byte {
+//
+// A line longer than maxTailTokenBytes is reported rather than dropped:
+// bufio.Scanner stops permanently on bufio.ErrTooLong, so swallowing that
+// error would silently return a prefix of the window and hide every entry
+// after the oversized line — including the newest ones the caller came for.
+func splitLines(data []byte) ([][]byte, error) {
 	scanner := bufio.NewScanner(bytes.NewReader(data))
-	scanner.Buffer(make([]byte, 256*1024), 256*1024)
+	scanner.Buffer(make([]byte, 0, 256*1024), maxTailTokenBytes)
 	var lines [][]byte
 	for scanner.Scan() {
 		line := scanner.Bytes()
@@ -154,7 +176,10 @@ func splitLines(data []byte) [][]byte {
 		copy(cp, line)
 		lines = append(lines, cp)
 	}
-	return lines
+	if err := scanner.Err(); err != nil {
+		return nil, fmt.Errorf("splitting session log tail: %w", err)
+	}
+	return lines, nil
 }
 
 // tailEntry is the minimal structure we decode from each JSONL line.
--- a/internal/sessionlog/tail_usage.go
+++ b/internal/sessionlog/tail_usage.go
@@ -56,11 +56,87 @@ func ExtractTailUsage(path string) ([]TailUsage, error) {
 	}
 	defer f.Close() //nolint:errcheck // best-effort close on read-only file
 
-	data, _, err := readTail(f, tailChunkSize)
+	data, _, err := readTail(f)
 	if err != nil {
 		return nil, err
 	}
+	return parseTailUsage(data), nil
+}
 
+// maxUsageScanBytes caps how far ExtractTailUsageSince will grow its scan
+// window when it cannot reach the cursor entry. It bounds the worst case (a
+// long-lived transcript whose cursor is stale or absent) while still covering
+// transcripts orders of magnitude larger than the fixed tail window.
+const maxUsageScanBytes = 16 * 1024 * 1024
+
+// ExtractTailUsageSince returns usage-bearing invocations from the tail of a
+// transcript, growing the scan window until the entry identified by cursorID
+// is inside it — so no invocation is skipped merely because more than
+// tailChunkSize bytes were appended since the last extraction.
+//
+// The fixed-window ExtractTailUsage silently drops any invocation that
+// scrolled past the last 64KB between two extractions. The persisted cursor
+// (session.MetadataKeyInvocationUsageCursor) is a message identity, not a byte
+// offset, so a later pass cannot reach back for what it missed and the usage
+// is lost permanently. Growing the window until the cursor is visible closes
+// that gap without introducing new persisted state: the caller's existing
+// cursor filter still decides what is new.
+//
+// The window doubles from tailChunkSize until the cursor entry is found, the
+// whole file is in the window, or maxUsageScanBytes is reached. An empty
+// cursorID (a session with no prior extraction) reads a single tailChunkSize
+// window, matching ExtractTailUsage: there is no anchor to reach back to, and
+// backfilling a full transcript on first sight is not this function's job.
+// Callers may receive entries at or before the cursor and must still filter.
+func ExtractTailUsageSince(path, cursorID string) ([]TailUsage, error) {
+	f, err := os.Open(path)
+	if err != nil {
+		return nil, err
+	}
+	defer f.Close() //nolint:errcheck // best-effort close on read-only file
+
+	if cursorID == "" {
+		data, _, err := readTail(f)
+		if err != nil {
+			return nil, err
+		}
+		return parseTailUsage(data), nil
+	}
+
+	for window := int64(tailChunkSize); ; window *= 2 {
+		data, _, truncated, err := readTailWindow(f, window)
+		if err != nil {
+			return nil, err
+		}
+		usages := parseTailUsage(data)
+		// Reached the cursor, or the whole file is in view: nothing older can
+		// still be owed. Either way this window is complete.
+		if !truncated || containsCursor(usages, cursorID) {
+			return usages, nil
+		}
+		if window >= maxUsageScanBytes {
+			// Cap hit with the cursor still out of view: return the widest
+			// window rather than nothing, so a stale cursor degrades to the
+			// old bounded behavior instead of losing everything.
+			return usages, nil
+		}
+	}
+}
+
+// containsCursor reports whether any extracted invocation carries the cursor
+// identity, under the same message-id-else-entry-uuid rule the telemetry
+// cursor is written with.
+func containsCursor(usages []TailUsage, cursorID string) bool {
+	for _, u := range usages {
+		if u.MessageID == cursorID || u.EntryUUID == cursorID {
+			return true
+		}
+	}
+	return false
+}
+
+// parseTailUsage extracts invocations from an already-read transcript window.
+func parseTailUsage(data []byte) []TailUsage {
 	var usages []TailUsage
 	// byMessageID maps a message identity to its index in usages so the
 	// content-block copies of one API response collapse to a single entry.
@@ -101,7 +177,18 @@ func ExtractTailUsage(path string) ([]TailUsage, error) {
 		}
 		usages = append(usages, u)
 	}
-	return usages, nil
+	return usages
+}
+
+// ExtractTailUsageSinceFromSearchPaths reads cursor-aware tail usage only
+// after verifying path resolves under one of the configured session-log
+// search roots. Mirrors ExtractTailUsageFromSearchPaths.
+func ExtractTailUsageSinceFromSearchPaths(searchPaths []string, path, cursorID string) ([]TailUsage, error) {
+	safePath, err := validateSearchPathFile(searchPaths, path)
+	if err != nil {
+		return nil, err
+	}
+	return ExtractTailUsageSince(safePath, cursorID)
 }
 
 // ExtractTailUsageFromSearchPaths reads tail usage only after verifying
--- a/internal/sessionlog/tail_usage.go
+++ b/internal/sessionlog/tail_usage.go
@@ -2,6 +2,7 @@ package sessionlog
 
 import (
 	"encoding/json"
+	"log"
 	"os"
 )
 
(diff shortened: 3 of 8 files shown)


TheChrisOneil:
Note on the red check: the `Evidence / critical-path suites` failure is `prwatchdog: tests never ran`, not a test failure. Six workflows on this head sit at `action_required` — `CI`, `Mac Regression`, `CodeQL`, `Go Vulnerability Scan`, `Cyclomatic complexity`, `Review Formulas` — the usual first-time-fork-contributor gate, so the watchdog waited ~25 min for `CI / required` evidence that could not arrive. Nothing has actually been exercised in CI yet; it should need only an "Approve and run workflows".

Locally against this head: `fmt-check`, `lint` (0 issues), `vet`, and the three row checks pass, and `internal/sessionlog`, `internal/worker`, and `internal/worker/workertest` are green. The two `make test` failures I hit (`TestDoctorCheckVersionFloor*` needing `dolt` on the `env -i` PATH, and `TestCmdGCProcessTimingEnvCrossesMakeIsolation`) reproduce identically on clean `main` at 1154ad1 with this branch absent, and `check-residency-boundary` can't run on macOS at all — `declare -A` needs bash 4+. Details in the PR body.

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277288921

TheChrisOneil:
Validation against real transcripts, in case it helps the review.

I built a throwaway harness (not part of this PR) that runs both extractors over a live city's Claude Code transcripts and compares them to ground truth — every usage-bearing invocation actually present in each file. For each transcript it simulates a sweep that has fallen behind by using the file's *first* invocation as the cursor.

**Wind-app rig — 28 transcripts, 25 of them losing data:**

| | invocations | tokens |
|---|---|---|
| ground truth | 1,030 | 96,143,303 |
| OLD fixed 64KB window | 238 (23.1%) | 22,175,186 (23.1%) |
| NEW cursor-aware window | **1,030 (100%)** | **96,143,303 (100%)** |

**Whole city — 286 transcripts, 191 losing data:**

| | invocations | tokens |
|---|---|---|
| ground truth | 7,170 | 1,218,454,554 |
| OLD fixed 64KB window | 1,592 (22.2%) | 154,094,628 (**12.6%**) |
| NEW cursor-aware window | **7,170 (100%)** | **1,218,454,554 (100%)** |

Recovered on the city-wide run: 5,578 invocations and ~1.06B tokens that the current extractor drops silently.

One detail worth flagging, because it makes the bug worse than the invocation count suggests: city-wide, the old window recovers 22.2% of *invocations* but only 12.6% of *tokens*. The loss is biased toward the expensive turns — an invocation late in a long conversation carries the largest cache-read count and is also the most likely to have scrolled out of a 64KB window. Cost reporting therefore skews low by more than the invocation ratio implies.

Caveat on scope: this fixes forward loss. It does not reconstruct history — a session whose bead is gone is not re-swept, so an operator wanting past spend still has to parse transcripts directly.

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @TheChrisOneil! This PR closes a silent, permanent
accounting gap in usage extraction: the extractor read a fixed 64KB transcript
tail while its dedup cursor is a *message identity* with no byte offset, so any
session that fell more than one window behind lost its usage and cost facts for
good — a later pass had no way to reach back for what it missed. Your fix grows
the scan window (doubling from 64KB up to a 16MB cap) until the cursor is
actually inside it, which closes the gap **without** introducing new persisted
state — the caller's existing cursor filter still decides what is new. That
restraint is what made this land as a bug fix rather than a design change, and
the backlogged-session and first-pass-after-deploy cases it recovers are exactly
where the accounting was previously worst.

branch.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **The grown window could still truncate silently at an oversized line.** The
  first pass found that once the window grows, a single >256KB non-usage line
  sitting between the cursor and the newest invocation would make
  `bufio.Scanner` return `ErrTooLong` — and `splitLines` swallowed it, so the
  parse stopped early and the extraction looked complete. The final patch makes
  `splitLines` return that error instead of dropping it, raises the per-token
  cap to the package's existing 50MB convention (`maxTailTokenBytes`,
  `tail.go:142`), and pins the behavior with a regression test
  (`TestExtractTailUsageSinceSpansOversizedTranscriptLine`). Both halves were
  needed: raising the cap alone would still swallow the next overflow.
- **The synchronous scan cost was undocumented.** Both production callers run
  synchronously — the prompt-op return path and the deadline-bounded controller
  reconcile tick — and each doubling re-reads *and* re-parses the whole window.
  The final patch documents that budget on `maxUsageScanBytes` with measured
  figures, plus the per-fact sink cost that dominates it on a cap-hit batch
  (`usage.LocalSink` fsyncs per fact; `usage.ExecSink` spawns a subprocess per
  fact).
- **The coverage-ceiling comment overstated where growth applies.** Two
  iterations went into getting `internal/worker/invocation_telemetry.go` to
  describe the real per-family behavior: the codex family is still a fixed 64KB
  window, and the claude family's *empty-cursor first sweep* returns before the
  growth loop ever runs, so it neither grows nor emits the cap log. The text is
  now scoped per family instead of stating the grown behavior universally.
- **The extractor's own godoc didn't state the cap's consequence.** It said
  growth "closes that gap" while the correction lived only at the caller. It now
  says plainly that reaching `maxUsageScanBytes` *bounds* the gap rather than
  closing it — entries older than the cap are still lost, the return value is
  indistinguishable from a complete scan, and the cap-hit log line is the only
  signal.
- **The growth cost isn't recoverable on both lanes.** On the sweep lane the
  cost buys the backlog it was spent on; on the prompt-op seam a cap hit ends in
  `usagesAfterCursor`'s newest-only fallback, so it can spend the full growth
  cost to return exactly what the first 64KB window already held. That is
  correctly not a code defect — reachability isn't knowable in advance — but
  anyone tuning the cap from the doc alone would not have learned it, so the
  cost paragraph now says so.

Review findings addressed:
- All five items above are fixed in `c4169117` ("fix(sessionlog): surface
  oversized transcript lines in the grown usage window"), which also adds
  `TestFactorySweepWithoutCursorMatchesFixedTailWindow` to pin the empty-cursor
  branch at the worker level — the review checked that it is load-bearing, not
  decorative: remove the short circuit and the assertion fails.

- `ga-2fpqe`: Decide whether the 16MB usage-scan cap is a product limit or a
  temporary safety bound. Behavioral Correctness, minor. Past the cap the
  original permanent-loss failure recurs — narrowed 256×, but not eliminated —
  and the review confirmed this is base behavior that your change bounds rather
  than something the PR introduced. The real design question belongs in its own
  change, tracked alongside the codex-family fixed-window gap.
- `ga-5jk3w`: Centralize the 50MB per-token scanner ceiling in
  `internal/sessionlog` on `maxTailTokenBytes`. Architectural Consistency,
  minor. The new constant's doc asserts parity with `parseFileDetailed`, but
  `reader.go:427` (and eight other readers in the package) still hold the bare
  literal, so nothing enforces the invariant. Pre-existing and outside this PR's
  scope.

### Maintainer Changes
`c4169117` "fix(sessionlog): surface oversized transcript lines in the grown
usage window" — 6 files changed, 394 insertions(+), 35 deletions(-). This
carries the review fixes above. The branch was also rebased onto current `main`
by a maintainer; one commit on the advanced base
(`d8d07eedd4 fix(sessionlog): seed codex tail model from the nearest preceding turn_context`)
touched the same region of `ExtractCodexTailUsage`, and the two changes were
combined so both survive — the base's `turn_context` model seeding is preserved
alongside your error-checked `splitLines` result. Your authorship on
`0856844d5c` is preserved.

### Final Review Status
Ready for the merge queue: final head `c4169117` has passing required GitHub
checks (`Check`, `CI / required`, and all four CodeQL `Analyze` contexts), and
the merge-ready metadata is in place. I am marking this PR

CI: https://github.com/gastownhall/gascity/actions/runs/33856174156

### Review Iterations
4 review passes performed. The first three each returned concrete required
changes — the oversized-line truncation and cost documentation, then two
rounds tightening the per-family coverage-ceiling comment — and the fourth
approved, with the remaining minors either applied as documentation fixes or
filed as the follow-up beads listed above.

---

PR #5925, opened 2026-09-02
Ensure every managed pool session has a durable session bead before claim

sjarmak:
## The gap

A managed pool runtime is identified by `GC_TEMPLATE` (pool membership) alongside
`GC_SESSION_ID`, which the canonical session-creation front door sets. A runtime that
has `GC_TEMPLATE` set but `GC_SESSION_ID` empty never went through that front door, so
no durable session bead exists to verify it against.

`gc hook --claim` treated that combination as "un-fenceable" and fell through to the
normal claim path. An unregistered pool slot could therefore have assignee and routing
metadata written onto it, producing a claim attributed to a session that does not exist
in the store. Nothing downstream can reconcile that: the session lookup finds nothing,
so the work reads as claimed by a session the reconciler cannot see.

## The fix

Add a `missing_session_registration` fence branch that refuses the claim before any work
query runs, so no bead is mutated. It is kept distinct from `stale_session` because the
two failures need different operator responses: a stale session was registered and has
since been superseded, while this one was never registered at all.

A genuinely non-pool caller (no `GC_TEMPLATE`, no `GC_SESSION_ID`) still takes the
existing tokenless-runtime compatibility path, unchanged.

`schemas/hook/result.schema.json` gains the new refusal reason so the wire result stays
typed.

## Test plan

- `cmd/gc/cmd_hook_claim_session_fence_test.go` (193 lines) covers the new refusal,
  asserts no bead mutation occurs on that path, and pins the three neighbouring cases so
  the branch cannot widen by accident: pool runtime with a valid session bead (claims),
  pool runtime with a stale session (`stale_session`, not the new reason), and tokenless
  non-pool runtime (compatibility path preserved).
- `go test ./cmd/gc` and `go vet ./...` green on this head.

https://claude.ai/code/session_013Gnx7n8v6xtdCLGJ4hVjMX

### DIFF
--- a/cmd/gc/cmd_hook.go
+++ b/cmd/gc/cmd_hook.go
@@ -297,16 +297,18 @@ func cmdHookWithOptions(args []string, opts hookCommandOptions, stdout, stderr i
 	// do the same immediately after loadCityConfig.
 	resolveRigPaths(cityPath, cfg.Rigs)
 
-	// Fence a stale/superseded runtime session BEFORE the city-suspension,
-	// agent-resolution, and agent-suspension early returns below. A stale
-	// incarnation in a suspended city, or one whose template was removed from
-	// config (resolveAgentIdentity fails), or whose agent was suspended, would
-	// otherwise hit one of those bare `return 1` paths, and its startup wrapper
-	// would keep retrying the plain failure instead of seeing the terminal
-	// stale-session drain result and exiting. The fence reads the runtime's own
-	// identity from the environment; it is a no-op for a non-session runtime (no
-	// GC_SESSION_ID / GC_INSTANCE_TOKEN) and fails open for an eligible session or a
-	// transient session-store fault, so a healthy worker still falls through to the
+	// Fence a stale/superseded/unregistered runtime session BEFORE the
+	// city-suspension, agent-resolution, and agent-suspension early returns
+	// below. A stale incarnation in a suspended city, or one whose template was
+	// removed from config (resolveAgentIdentity fails), or whose agent was
+	// suspended, would otherwise hit one of those bare `return 1` paths, and its
+	// startup wrapper would keep retrying the plain failure instead of seeing
+	// the terminal drain result and exiting. The fence reads the runtime's own
+	// identity from the environment; it is a no-op for a genuinely non-session
+	// runtime (no GC_TEMPLATE and no GC_SESSION_ID) and fails open for an
+	// eligible session, a session id present with no instance token (the
+	// documented tokenless-runtime compatibility escape hatch), or a transient
+	// session-store fault, so a healthy worker still falls through to the
 	// suspension and config checks below.
 	if opts.Claim {
 		// F-A, at the earliest point that can answer it. tryHookClaim carries the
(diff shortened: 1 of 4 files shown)


sjarmak:
## Cross-provider review record

Recorded here so the evidence sits on the PR rather than only in a work-tracking
comment. Two independent legs, both PASS, no blocking defect.

**Reviewed head:** `95fd8ac9b2ce7ac222ca7cd752d67a7bd3bbd90e`
**Current head:** `da64d43717dd1903f864cc4df6a7df2c50c7ff1a` (rebase of the same
change onto current `main`; see the rebase note at the end)

**Scope:** `fenceHookClaimSession` in `cmd/gc/cmd_hook.go`, its doc comment and
block comment, the reason constant and drain emitter in
`cmd/gc/cmd_hook_claim.go`, and the additive enum value in
`schemas/hook/result.schema.json`.

**Leg A:** Claude Opus 5. Enumerated the function's return paths against the
rewritten prose before reading leg B, so agreement is corroboration rather than
anchoring.
**Leg B:** Codex CLI, prompt passed positionally, no diff attached and no file
reads requested.

### Return-path enumeration

Both legs produced the same mapping. Six returns against six documented cases,
nothing undocumented and nothing unreachable.

| Return | Behavior | Documented case |
| --- | --- | --- |
| 545 | drain, `handled=true` | managed pool runtime with no verifiable session-bead registration |
| 547 | `(0, false)` | no `GC_TEMPLATE` and no session id |
| 551 | `(0, false)` | a session id present but no instance token |
| 556 | drain, `handled=true` | a definitively stale session |
| 562 | `(0, false)` | a transient session-store fault |
| 564 | `(0, false)` | an eligible session |

(Line numbers are at the reviewed head.)

### Findings

Two, both non-blocking.

**1. The block comment's parenthetical overstates.** It says `GC_TEMPLATE` is
"set only alongside `GC_SESSION_ID` by `RuntimeEnvWithSessionContext`, the single
front door". That is not literally true: `cmd/gc/work_query_probe.go:181` sets
`GC_TEMPLATE` with no `GC_SESSION_ID` in the env it builds, and
`controllerQueryEnv` (`work_query_probe.go:107`) sets no session id either.

It is not a live regression, for two independent reasons. That branch is skipped
for pool agents (`work_query_probe.go:168` returns early when
`agentCfg.SupportsMultipleSessions()`, before line 181 is reached), so a pool
template never takes it. And the env prefixes the agent's configured
`work_query` command, whose default is a `bd ready --json` shape rather than
`gc hook --claim`, so it does not enter the fenced path at all.

The residual exposure is narrow and worth naming: a deployment that configures a
singleton (non-pool) agent's `work_query` to invoke `gc hook --claim` would have
that claim refused where it previously fell through unfenced. Leg B reached the
same conclusion about the wording from the prose alone, without naming a site;
leg A supplied the site. Suggested follow-up, not a blocker: soften the
parenthetical to describe `RuntimeEnvWithSessionContext` as the front door for
live runtime identity rather than as the only writer of the variable.

**2. The enum addition is additive at the schema level and is not universally
backward compatible on the wire.** A consumer that validates against the old
closed enum, deserializes into a strict generated enum, or handles the value set
exhaustively and rejects unknowns will break on `missing_session_registration`.
A consumer written to tolerate unknown future result values is unaffected. Both
legs agree.

### Reorder check

Moving the `sessionID == ""` test ahead of the instance-token test changes
behavior for exactly one input class, the intended one: `GC_TEMPLATE` set and
non-blank, `GC_SESSION_ID` empty. With `GC_TEMPLATE` absent or whitespace-only
the result stays `(0, false)` regardless of the token, and every input carrying a
session id retains the previous token guard. Verified against the pre-change
combined guard `if sessionID == "" || instanceToken == ""`. Both legs agree.

### Rebase note

The branch was rebased onto current `main` after the review. The change against
`main` is byte-for-byte the reviewed scope in every respect except three hunks,
each a mechanical adaptation to upstream drift:

1. gofmt alignment in the reason-constant block, shifted by `drain_pending`
   landing upstream with a longer name.
2. `writeHookClaimDrain` gained a leading `label` parameter upstream. The new
   call site passes `hookClaimLabel` first, matching every sibling call site
   verbatim (compare `cmd_hook_claim.go:1150` and `:1161`).
3. The schema enum's trailing comma moved, because `drain_pending` landed
   upstream after `non_turn_context`.

No semantic change to the reviewed content. Diffstat against `main` is unchanged
at 4 files, +257/-23.

quad341:
Merged — thanks, @sjarmak. Re-verified on current `main` first (clean merge on top of #5325, build and vet, and the four new session-registration fence tests). Requiring a durable session bead before a pool session can claim work closes a real gap; appreciated.

PR #5926, opened 2026-09-03
fix(bd): harden the fresh managed-Dolt bootstrap as a superset of #5294

bourgois:
#5294 established the right invariant: never seed the bd version witness for a database this invocation did not create, because that bypasses bd's cross-era migration guard on exactly the workspace that needs it. That guard is kept here verbatim, disk sampling and all.

It is necessary but not sufficient. Provenance answers *is this store new*, not *is this store current-era*, and those come apart in a case a managed fleet hits routinely: a **new scope directory pointed at an existing current-era database** — a rig re-clone, or a second scope on a shared Dolt server. The store legitimately pre-existed, so a provenance-only test never writes the witness, and bd refuses the workspace as a legacy Dolt server workspace forever with no path forward.

This adds the missing half: on *schema present*, require `schema_migrations` (a >=1.0 table; pre-1.0 stores tracked versions in `config`) as positive evidence of a current-era store before stamping. An adopted **old-era** store still has `config` and no `schema_migrations`, so it is still refused. Both tests agree there, by different routes.

Four further cases found running this on a real fleet:

- **Never force-reinitialize on an unanswered schema probe.** A probe that fails for a third reason (server not answering, database not selectable) proves nothing, and forcing a reinit onto a database nobody could inspect is the same data-safety failure the unreachable-server check already refuses.
- **Heal an interrupted bootstrap** rather than reporting it ready or forcing it into bd's guard. A `config` table proves a bootstrap *started*, not that it finished; a database left between a migration's DDL and its per-step commit sits behind the binary's schema, and read-only opens never migrate it.
- **Parse dolt scalars by shape, not line number.** Some builds print an old-version notice on stdout, so "line 4 of the output" is not the value.
- **Declare the shared Dolt server** so bd's second legacy-workspace branch stops refusing managed scopes.

## Verification

This branch is `main` plus two files. `go build ./...` is clean and the bd-provider lifecycle suite passes **including #5294's own `TestGcBeadsBdInitDoesNotSeedVersionWitnessForAdoptedPreExistingDatabase`** and the fresh-path test.

One implementation note, because it cost real debugging: the witness gate samples provenance **once at `op_init` entry**, not inside `ensure_database_registered`. That function early-returns when the database is already catalogued, and can also run after bd has put the store on disk, so sampling there gives a false negative on a real fresh init against a live server. It passes against fakes and fails against a real Dolt server, which is an unpleasant way to find out.

Also banks a small improvement: the reworked tests use fewer subprocess call sites than the ones they replace, so three shrink-only census baselines are lowered rather than left slack.

## The longer-term fix

This is a workaround. The durable fix is on the beads side: bd should expose an explicit externally-managed declaration instead of gc reproducing the private decision table in `cmd/bd/legacy_upgrade_guard.go` from the outside. Happy to file that RFC separately.

---

## Rebase note (2026-09-04)

Rebased onto current `main`. One non-trivial conflict, recorded here because the resolution has a consequence a reviewer should decide on.

`main`'s `2f2cea0dc` changed `seed_fresh_managed_bd_version_witness` to probe `"${BD_BIN:-bd}"` instead of bare `bd`, so the witness names the binary `run_bd_pinned` actually runs. This PR **deletes** that function and replaces it with `ensure_current_era_version_witness`. The conflict was resolved by taking this PR's deletion — the function and its sole call site both go away, and no dangling reference remains (`bash -n` clean).

**Consequence:** `ensure_current_era_version_witness` probes bare `bd version`, so on the new base this PR loses `2f2cea0dc`'s BD_BIN fix on the witness path. Carrying it forward would be a one-token change (`bd version` -> `"${BD_BIN:-bd}" version`) but it edits a line this PR authored rather than a conflict hunk, so it was deliberately **not** made here. Flagging for a maintainer call.

`2f2cea0dc`'s other two `${BD_BIN:-bd}` sites (`run_bd_pinned`, `run_bd_doltlite`) auto-merged and are intact.

The PR's authored content is otherwise unchanged: the multiset of added lines is byte-identical before and after the rebase (872 = 872). Only the removed-line set moved, and only because it now deletes `main`'s newer wording of the function this PR removes. The three census baselines re-verified clean on the new base (`TestRepositoryLedgerMatchesCensusAndDocumentation` passes), and all five new lifecycle tests pass.


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277288706

PR #5928, opened 2026-09-03
fix(reconciler): resolve the warm-bind trigger through the residency resolver

julianknutsen:
## The bug

`warmClaimTriggerStore` resolved a pool slot's bound trigger by hand-parsing the
session's `gc.trigger_bead_store_ref`: `rig:<name>` selected a cached rig ledger,
everything else fell back to the city work ledger. The probe was built with
`buildWarmClaimTriggerProbe(cr.cityBeadStore(), cr.rigBeadStores())` — two work
axes and no binding.

On a split city that stamp can never name the right store:

- the bound trigger is a **graph-class step bead**, which lives ONLY in the class
  binding — a store neither argument ever contains;
- the stamp is the **demand LEG key** the row was probed under
  (`build_desired_state.go` `entry.StoreRefs[b.ID] = group.storeKey`, carried as
  `request.WorkStoreRef` and stamped at bind time). On a live split city those
  rows read `city:<name>`.

So every stamp value resolved to a work ledger, `store.Get` returned
`ErrNotFound`, the probe failed closed, and the warm slot sat idle at its prompt
holding an unclaimed trigger. On tmux the idle-claim backstop eventually recovers
it; on herdr this probe is the ONLY closer of the warm-bind gap (see the doc
comment on `deliverWarmBindClaimNudge`).

## The fix

`buildWarmClaimTriggerProbe` now takes a trigger **resolver**
(`func(triggerID string) (beads.Bead, error)`) instead of two stores; the hand
parser is deleted. The reconciler builds that resolver from the residency
contract in `(*CityRuntime).newWarmClaimTriggerResolver`, the controller-side
mirror of the CLI's by-id door (`cliByIDOwner`):

```go
plan, err := storeref.Plan(storeref.ByID{ID: triggerID}, cr.residencyTopology(servingRigs))
owner, err := storeref.ResolveOwnerRow(plan, triggerID)
return beadForOwner(owner, triggerID)
```

The doctrine: **the residency resolver is the one lookup contract; the demand leg
key says where the row was COUNTED, not where it LIVES.** The stamp is still
written and other readers still use it — it is simply no longer a store selector.

Two consequences fall straight out of the resolver's rules:

- the binding **leads** for an id inside its reserved namespace, so a
  binding-resident `gcg-*` step is found at all;
- the same-id relic `gc storage migrate` leaves in the work ledger (ids are
  preserved, and that copy is frozen open forever, so it reads as unclaimed on
  every tick) is **shadowed**, not answered.

Fail-closed is preserved end to end: a plan the topology cannot answer, a refused
city, a dark leg or a clean miss all reach the probe as an error/miss and it
returns false. `deliverWarmBindClaimNudge`'s gates are untouched.

An unrelocated city is unchanged: a `ByID` plan with no bindings is the work
ledger plus the rig legs whose configured prefix covers the id, with the
single-leg work residual returned unprobed.

## Tests

`cmd/gc/warm_bind_nudge_test.go`, all built through the production wiring — a
real `CityRuntime` and `cr.residencyTopology`, via `splitRoutes(binding)` and
`residencyTestConfig()` — rather than a hand-assembled `storeref.Topology`:

- **`TestBuildWarmClaimTriggerProbe_ResolvesBindingResidentTrigger`** — a `gcg-1`
  step present ONLY in the binding, with the city work store and the rig store
  empty, resolves under every demand-leg stamp a live city writes (`""`, `city`,
  `city:test-city`, `class:gmnos`, `rig:alpha`). Claiming the binding row flips
  the probe to false under all five (the churn invariant still holds).
- **`TestBuildWarmClaimTriggerProbe_BindingShadowsRelicCopy`** — the same id open
  in the work ledger (relic) and `in_progress` in the binding reads false: the
  binding wins. Control: the same relic on a city that relocates nothing answers
  true, so the assertion is the binding winning rather than the probe going blind
  to work.
- **`TestBuildWarmClaimTriggerProbe_LegacyCityResolvesWorkAndRigLegs`** — the
  adapted `..._ResolvesFromStoreRef`: the legacy work leg and the covering rig
  leg answer exactly as before, a claimed rig row still flips the probe, a miss
  and an empty trigger id are still fail-closed. Its one case that could not
  survive is the unknown-rig fail-closed (there is no hand parser left to feed a
  bad stamp to), inverted into the assertion that a wrong stamp no longer decides
  anything.

## Mutation probe

Restoring the hand parser verbatim (as a scratch `buildWarmClaimTriggerProbeLegacy`
the test helper was pointed at) and re-running:

```
--- FAIL: TestBuildWarmClaimTriggerProbe_ResolvesBindingResidentTrigger
    stamp "":              want unclaimed=true
    stamp "city":          want unclaimed=true
    stamp "city:test-city": want unclaimed=true
    stamp "class:gmnos":   want unclaimed=true
    stamp "rig:alpha":     want unclaimed=true
--- FAIL: TestBuildWarmClaimTriggerProbe_BindingShadowsRelicCopy
    the work-ledger relic answered
--- FAIL: TestBuildWarmClaimTriggerProbe_LegacyCityResolvesWorkAndRigLegs
    a stamp naming an unknown store must not suppress a resolvable trigger
```

Every stamp fails, which is the bug reproduced. The legacy test fails on exactly
one line — the inverted case — and its work/rig-leg assertions still pass under
the parser, which is the pin that the legacy answer did not move.

## Gates

| gate | result |
| --- | --- |
| `go build ./cmd/gc` | 0 |
| `go vet ./cmd/gc` | 0 |
| `gofmt -l` (changed files) | clean |
| `./scripts/check-residency-boundary.sh` | 0 (no baseline change: the plan-based read names no forbidden vocabulary) |
| `go test ./scripts/ -run TestResidency` | ok |
| `go test ./cmd/gc -run 'WarmBind\|WarmClaim\|Residency\|ByID' -count=1` | ok |
| `go test ./internal/storeref/... -count=1` | ok |
| `make test-fast-parallel` | all fast jobs passed |

Refs ga-bmfmt

### DIFF
--- a/cmd/gc/city_runtime.go
+++ b/cmd/gc/city_runtime.go
@@ -2403,6 +2403,37 @@ func (cr *CityRuntime) stopConfigWatcher() {
 	}
 }
 
+// newWarmClaimTriggerResolver returns the reader the warm-bind claim probe asks
+// for one bound trigger bead, resolved over this controller's own topology: the
+// binding it opened at boot, its city work ledger, and the rig legs it is handed.
+//
+// It is the controller-side mirror of the CLI's by-id door (cliByIDOwner) — plan
+// ByID, execute it, and take the row the winning probe already read — for the same
+// reason that door exists. The binding LEADS for an id inside its reserved
+// namespace, so a graph-class step that lives only in the binding is found at all,
+// and a same-id relic that `gc storage migrate` left behind in the work ledger
+// (ids are preserved) is shadowed rather than answered: that relic is frozen open
+// forever, so answering from it would read as unclaimed on every tick.
+//
+// The topology is built once per reconcile tick, alongside the probe. Resolution
+// errors — a refused city, a dark leg, an id no leg could answer — reach the probe
+// as errors and it declines to nudge; nothing here converts a failed read into
+// absence.
+func (cr *CityRuntime) newWarmClaimTriggerResolver(servingRigs map[string]beads.Store) warmClaimTriggerResolver {
+	topo := cr.residencyTopology(servingRigs)
+	return func(triggerID string) (beads.Bead, error) {
+		plan, err := storeref.Plan(storeref.ByID{ID: triggerID}, topo)
+		if err != nil {
+			return beads.Bead{}, err
+		}
+		owner, err := storeref.ResolveOwnerRow(plan, triggerID)
+		if err != nil {
+			return beads.Bead{}, err
+		}
+		return beadForOwner(owner, triggerID)
+	}
+}
+
 // beadReconcileTick runs one bead-driven reconciliation pass. bootReconcile is
 // true only for the synchronous pass on the startup path: that pass must flip
 // readiness quickly, so it skips the undesired-pool-session sweep (a heavy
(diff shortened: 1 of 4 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277284376

quad341 (dismissed):
@julianknutsen — the fix is good, but the diagnosis is the part that sells it: that `gc.trigger_bead_store_ref` records the demand *leg* the row was counted under, not where the row lives, and that on a split city those two are never the same for a binding-resident `gcg-*` step. I traced it back through `build_desired_state.go`'s `entry.StoreRefs[b.ID] = group.storeKey` and walked `planByID`/`ResolveOwnerRow` in the resolver — it holds exactly as you describe, and the legacy path stays byte-identical because the work residual comes back unprobed when it's last.

Merging as-is. CI ran the real `cmd/gc` shards, so this is properly covered.

Two things I especially liked: building the tests through a real `CityRuntime` and `newWarmClaimTriggerResolver` instead of a hand-written `storeref.Topology` — that's the difference between pinning the wiring and passing after the wiring rots — and the control case in the relic test proving the binding is *winning* rather than the probe going blind to work. The mutation probe made review much faster.

One follow-up I'll file on my side, not yours: `poolClaimBackstop.resolve` in `idle_nudge.go` still goes through the stamp. Thanks for this one.


---

quad341 (approved):
Re-reviewed `68eb81356`..`bb2dd5c35` (6 new commits since the last review): no new concerns — the prior review stands.

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR fixes a reconciler issue by resolving the warm-bind trigger through the residency resolver, which keeps the reconciler area easier to maintain and verify.

### Original PR Review

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `bb2dd5c3545b` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

### Review Iterations

---

PR #5929, opened 2026-09-03
fix(api): resolve a workflow's scope from its root ref before the store leg

julianknutsen:
## What this changes

A workflow is now presented in the scope its **root ref** names. The store leg it
was read through is where it lives, not what it belongs to.

`workflowRootScope` (`internal/api/handler_convoy_dispatch.go`) only answered when a
root carried both `gc.scope_kind` and `gc.scope_ref`. Sling-launch `graph.v2` roots
stamp neither — they carry only `gc.root_store_ref` — so all four projections
(`workflowSelectionScope`, `workflowEventScope`, `workflowSnapshotScope`, and
`workflowProjectionScope` in `orders_feed.go`) fell through to the scope of the
physical leg. `workflowStores` registers the graph class binding as a **city-scoped**
leg, so a `rig:`-rooted root served through that binding was presented as
city-scoped, and the rig-scoped `GET /v0/city/{city}/workflow/{id}` that names it
returned 404 for a bead that exists (`workflowScopeMatches` compares against the
leg-derived scope). On maintainer-city that is 9 of 13 live `in_progress` roots.

`internal/runproj` already got this right for the run projection:
`fromRootMetadataScope` prefers the explicit stamps, then parses
`gc.root_store_ref`. This exports that fallback arm as
`runproj.ScopeFromRootStoreRef` and has `workflowRootScope` use it, so the two
projections of the same root agree instead of each carrying its own rule.

## Review notes

- **No wire-shape change.** Values only. `make dashboard-ci` regenerates nothing —
  neither `internal/api/openapi.json` nor the generated TS changed, and the
  pre-commit spec/client regeneration produced no diff either.
- **The fallback is narrow.** A ref that names no scope — a class binding
  (`class:…`, `graph:…`, `orders:…`), an empty value, a malformed pair — still
  resolves to nothing and falls back to the leg exactly as before. That keeps the
  legacy city-on-rig `preserveRequestedWorkflowScope` path and
  `workflowEventScope`'s legacy rig→city preference working for roots that carry no
  root ref at all (both are covered by existing tests, which still pass unchanged).
- **Explicit stamps still win**, with their existing semantics (trim, both required,
  no kind validation) — the new arm is reached only when a stamp pair is absent.
- One helper, four call sites: fixing `workflowRootScope` is what makes selection,
  events, snapshots and the orders feed agree; no per-call-site patching.
- Out of scope, noted for a reviewer: `workflowSQLSnapshotScope`
  (`internal/api/convoy_sql.go`) resolves the SQL fast path's scope by a different
  rule (requested scope → leg → stamps). It does not 404 a rig-scoped read, so it is
  not part of this bug, but the two rules are worth unifying separately.

## Test plan

TDD, RED first. New tests (all failed before the fix, pass after):

- `TestWorkflowRootScopeFallsBackToRootStoreRef` — table over stamps-win,
  `rig:beads`, `city:test-city`, `class:gmnos`, `""`, `rig:` (malformed),
  `:beads`, half-stamped. RED: `rig root store ref` and `city root store ref`
  returned `"":""`.
- `TestWorkflowScopeMatchesBindingLegRigRootedRoot` — a root read through a
  city-scoped binding leg with `gc.root_store_ref=rig:beads` matches `(rig, beads)`
  and not `(city, test-city)`.
- `TestWorkflowGetResolvesRigScopeFromRootStoreRefOnBindingLeg` — the handler-level
  reproduction: rig-scoped `GET /v0/city/{city}/workflow/{id}` returned
  `404 workflow-not-found` before, `200` with `scope_kind=rig` after, while
  `root_store_ref` still names the physical leg (`graph:test-city`).
- `TestBuildWorkflowRunProjectionsScopesRigRootedRootByRootStoreRef` — the feed half:
  0 items before, 1 rig-scoped item after.

Mutation probe: reverting only the fallback arm (leaving the tests in place) fails
all four; the class/empty/malformed table rows keep passing, which is the control
showing the fallback did not widen into refs that name no scope.

Gates: `go build ./...`, `go vet ./internal/api/... ./internal/runproj/... ./cmd/gc`,
`gofmt -l` (clean), `golangci-lint run ./internal/api/... ./internal/runproj/...`
(0 issues), `go test ./internal/api/... ./internal/runproj/... -count=1` (ok),
`make dashboard-ci` (exit 0, no regenerated diff), `make test-fast-parallel`
("All fast jobs passed").

Refs ga-dezas, ga-oytw9


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277284477

quad341 (dismissed):
@julianknutsen — the diagnosis here is the part worth calling out. Four projections quietly falling through to the physical leg because sling-launch `graph.v2` roots stamp only `gc.root_store_ref`, and the fix landing in the one shared helper instead of patching `workflowSelectionScope`, `workflowEventScope`, `workflowSnapshotScope` and the orders feed separately. That's what stops them drifting apart again.

I checked the load-bearing pieces rather than taking them from the description: graphroute does stamp the root and not just the steps, `ScopeFromRootStoreRef` really does stay narrow (`class:`/`graph:`/empty/malformed all still fall back to the leg), and the `runproj` extraction is byte-for-byte behavior-preserving. Reusing runproj's arm so both projections of the same root agree beats a second local rule.

Special thanks for `TestWorkflowGetRejectsCityScopeForRigRootedWorkflow`. Pinning the behavior you deliberately *removed*, with a comment saying it's intended and not a regression to restore, is what stops someone "fixing" it back in six months.

Merging as-is. The `workflowSQLSnapshotScope` divergence you flagged is worth its own slice — the case I'd add there is an unscoped GET, where the SQL and scan paths can now report different scopes for the same root.


---

quad341 (approved):
@julianknutsen — the diagnosis here is the valuable part, and it's better than the fix. Four projections independently falling through to the physical store leg, all because sling-launch `graph.v2` roots stamp only `gc.root_store_ref` and never the explicit pair — and then landing the correction in the one shared helper so `workflowSelectionScope`, `workflowEventScope`, `workflowSnapshotScope` and the orders feed can't drift apart again. Nine of thirteen live `in_progress` roots 404ing on a bead that exists is a genuinely annoying bug to have chased to that helper.

I checked the load-bearing pieces against the tree rather than the description: `fromStoreRef` already rejects the empty ref, so the `runproj` extraction really is behavior-preserving for the run projection, and the `class:`/`""`/`rig:`/`:beads` rows do keep the legacy arms alive. The mutation probe you ran is the right evidence to bring, and `TestWorkflowGetRejectsCityScopeForRigRootedWorkflow` — pinning the behavior you deliberately removed, with a comment saying it's intended — is what stops someone "fixing" it back in six months.

One gap I'll close on our side rather than send back to you: `workflowEventScope` is the fourth projection and the only one without a new test — its existing test passes a bare bead, so it exercises the legacy arm only. I'll add the rig-rooted and class-binding rows to `TestWorkflowEventScope` and merge. Your `workflowSQLSnapshotScope` note is filed for its own slice — agreed it isn't part of this bug.


---

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR fixes a api issue by resolving a workflow's scope from its root ref before the store leg, which keeps the api area easier to maintain and verify.

### Original PR Review

### Maintainer Changes

### Final Review Status
Ready for the merge queue: final head `7a00ce332bc4` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

### Review Iterations

---

PR #5930, opened 2026-09-03
fix(dispatch): a rig control dispatcher on a split city probes the class binding it serves

julianknutsen:
## What this changes

A dispatcher probes the store its rows live in; on a split city that is the
binding for every scope.

`defaultScaleCheckTargetForAgent` points a Dir-scoped agent at
`rigStores[<rig>]` under `storeKey "rig:<name>"` — the rig WORK ledger. On a
city converged onto a class binding, every rig-rooted control row lives ONLY in
the binding, which the reconciler holds as its leading store. So the default
demand probe for a rig control dispatcher read an empty ledger:

- `scaleCheckDemandByTemplate[<rig>/core.control-dispatcher]` carried no
  `WorkBeadIDs` and no `StoreRefs`;
- the pool's `SessionRequest` was emitted with `WorkBeadID`/`WorkStoreRef`
  empty (`pool_desired_state.go`);
- the spawned dispatcher session bead therefore carried no
  `gc.trigger_bead_id`, and the idle-claim / warm-bind backstops keyed on that
  stamp were inert for it;
- and the create was gated on the health of a store that holds none of its
  work — an unavailable rig work store marked the template partial and
  `errPoolSessionCreatePartial` suppressed the create even though the binding
  had already proved demand.

The city dispatcher on the same topology got full provenance, because its own
target already IS the binding. The two scopes were asymmetric for no reason
other than which map the target came out of. Functionally masked today only
because the shipped dispatcher is `max_active_sessions=1` and self-polls.

`convergedRoutedWorkBinding` asks the question once per build from the
residency plan rather than guessing from config: the legs come from
`routedWorkStoreCandidates`, the same set `collectOpenUnassignedRoutedWork`
reads, so demand and collection cannot disagree about where routed work lives.
A converged split narrows (`storeref.Narrow`, runtime plane) to nothing but its
binding legs, so "every leg is a class ref" IS "this city is converged"; a city
that relocates nothing keeps its work leg and answers nil.

`ownScaleCheckTarget` then hands a store-scoped control dispatcher on such a
city the binding under the CITY store key — byte-for-byte the target the city's
own dispatcher already gets. Demand keys already normalize `class:*` onto
`"city"` (`normalizeDemandStoreRef`), and one store key means one store group,
listed once per tick for every dispatcher template; rows are attributed to
templates by route, not by store (`controllerDemandRouteCandidates`).

## Review notes

- The two `!storeScopedControlDispatcher` cross-store gates are untouched. This
  does not widen anyone's probe: a dispatcher still reads exactly one store, it
  is just the one its rows live in.
- Legacy (no binding) cities are byte-for-byte unchanged, and
  `defaultScaleCheckTargetForAgent` now carries a doc comment naming the
  split-city override.
- Partial gating now keys on the binding's health for rig dispatchers, correct
  by construction rather than by special-casing.
- Two bindings would mean a per-class fan-out no constructor in this build
  produces (`TestTopologyConstructorsServeOnlyTheWholeSplit`); until one can,
  such a plan gets the same conservative nil as a legacy city.
- Nothing in `openControlDispatcherDemand`, `normalizeDemandStoreRef`,
  `selectReadyUnassignedRoutedWork` or `repairBead` is touched — that is #5918's
  territory.
- Companion of #5918 (`ga-oytw9`), which is the COLLECTION half: admitting a
  rig-rooted row through the binding leg in `collectOpenUnassignedRoutedWork`,
  which is what feeds `ReadyUnassignedRoutedWorkBeads` and the
  `openControlDispatcherDemand` floor. Because that PR is still open, the RED
  here is a count of 0 rather than a provenance-less 1; the two PRs touch
  disjoint regions of `build_desired_state.go` and merge in either order.

## Test plan

New, in `cmd/gc/build_desired_state_test.go`:

- `TestBuildDesiredState_RigDispatcherOnSplitCityProbesTheBinding` — a
  rig-rooted, ready, routed control row lives only in the binding while
  `rigStores["fixture"]` is empty. Asserts `ScaleCheckCounts` is 1 for the rig
  dispatcher and 0 for the city dispatcher sharing the store group (attribution
  is by route), and that the session bead the build creates carries
  `gc.trigger_bead_id` = the row and `gc.trigger_bead_store_ref` = `city`.
- `TestBuildDesiredState_RigDispatcherOnSplitCityIgnoresRigWorkStoreHealth` —
  same city with no rig work store at all: the dispatcher is not in
  `PoolScaleCheckPartialTemplates` and its State entry exists.
- `TestBuildDesiredState_RigDispatcherOnLegacyCityStillProbesRigStore` — the
  control. Nothing relocated, row in the rig store, `gc.trigger_bead_store_ref`
  stays `rig:fixture`.

Mutation probe: disabling the override in `ownScaleCheckTarget` fails both
split-city tests (count 0; template partial) while the legacy control stays
green — so the two assertions are testing the override and not the fixture.

Gates run, all clean: `go build ./cmd/gc`, `go vet ./cmd/gc`, `gofmt -l cmd/gc`,
`golangci-lint run ./cmd/gc/...` (0 issues),
`./scripts/check-residency-boundary.sh`,
`go test ./cmd/gc -run 'ControlDispatcher|ScaleCheck|Demand|ClassBound|Residency' -count=1`,
and `make test-fast-parallel`.

### DIFF
--- a/cmd/gc/build_desired_state.go
+++ b/cmd/gc/build_desired_state.go
@@ -1769,6 +1769,10 @@ func convergedRoutedWorkBinding(
 	suspendedRigPaths map[string]bool,
 ) beads.Store {
 	legs, err := routedWorkStoreCandidates(cityPath, cfg, store, rigStores, suspendedRigPaths, censusRefScoped)
+	// An unresolvable topology takes the same conservative nil as a legacy city:
+	// fall back to the legacy target rather than guess a store. The error is not
+	// swallowed — collectOpenUnassignedRoutedWork resolves this same leg set later
+	// in this build and reports it there.
 	if err != nil || len(legs) == 0 {
 		return nil
 	}
--- a/cmd/gc/build_desired_state.go
+++ b/cmd/gc/build_desired_state.go
@@ -502,6 +502,10 @@ func buildDesiredStateWithSessionBeadsAt(
 			activeStores = append(activeStores, activeStore{store: s, ref: rig.Name})
 		}
 	}
+	// The store a control dispatcher's rows actually live in, asked once per
+	// build from the residency plan rather than guessed from config. Non-nil
+	// only on a city converged onto a class binding; see its doc comment.
+	controlBinding := convergedRoutedWorkBinding(cityPath, cfg, store, rigStores, suspendedRigPaths)
 
 	for i := range cfg.Agents {
 		if cfg.Agents[i].Suspended {
@@ -585,7 +589,7 @@ func buildDesiredStateWithSessionBeadsAt(
 			// retention for configured named-session beads.
 			poolDir := agentCommandDir(cityPath, &cfg.Agents[i], cfg.Rigs)
 			if store != nil && !hasCustomScaleCheck {
-				ownTarget := defaultScaleCheckTargetForAgent(cityPath, cfg, &cfg.Agents[i], store, rigStores)
+				ownTarget := ownScaleCheckTarget(cityPath, cfg, &cfg.Agents[i], store, rigStores, controlBinding, storeScopedControlDispatcher)
 				// mode='always': named session is unconditionally desired by the named
 				// pass; pool demand is redundant and creates {name}-N phantoms when N
 				// routed beads arrive. mode='on_demand': pool demand wakes the sleeping
@@ -661,7 +665,7 @@ func buildDesiredStateWithSessionBeadsAt(
 		// new unassigned demand while assigned work drives resume requests.
 		poolDir := agentCommandDir(cityPath, &cfg.Agents[i], cfg.Rigs)
 		if store != nil && !hasCustomScaleCheck {
-			ownTarget := defaultScaleCheckTargetForAgent(cityPath, cfg, &cfg.Agents[i], store, rigStores)
+			ownTarget := ownScaleCheckTarget(cityPath, cfg, &cfg.Agents[i], store, rigStores, controlBinding, storeScopedControlDispatcher)
 			defaultScaleTargets = append(defaultScaleTargets, ownTarget)
 			// Cross-store demand (FR-S0.1 / vp-s37): a rig pool's routed demand
 			// may live in the city store (vp-kvp cross-store delivery), which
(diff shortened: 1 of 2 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277284582

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR fixes a real dispatch stall on converged split cities: a rig's control dispatcher was sizing its demand probe against the rig **work** ledger, which holds none of that dispatcher's control rows, so its own routed work was counted by no target at all. That matters because on a split city every infrastructure class — including `graph` — lives in one binding, so the rows that justify spawning the dispatcher were invisible to the very probe meant to find them, and the per-bead `gc.trigger_bead_id` stamp never happened.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **The probe now reads the store that actually holds the rows.** `convergedRoutedWorkBinding` resolves the binding a converged split city routes control work to, and `ownScaleCheckTarget` points the dispatcher at it. The review independently traced the original bug rather than taking it from the PR description: `defaultScaleCheckTargetForAgent` keyed purely on the configured rig name, and `storageBootGate` only serves `none`/`whole`, so the binding-resident rows had no counting target.
- **The override is correctly scoped to one agent, not the whole build.** The `storeScopedControlDispatcher && controlBinding != nil` gate is computed per agent inside the agent loop (`build_desired_state.go:548`), so no other pool gets re-targeted. This was checked explicitly, since a build-scoped override would have been the dangerous version of this change.
- **The tests genuinely detect the fix's absence.** Review ran a mutation probe on a detached worktree with the override disabled; both new split-city tests failed for the pre-PR reason while the legacy control still passed. That is real sensitivity, not an assertion that only restates the new code.
- **The partial-gating change is intentional and pinned.** Rig-store health no longer gates the rig dispatcher's create — the store whose health gates a create is now the store whose rows justify it — and `TestBuildDesiredState_RigDispatcherOnSplitCityIgnoresRigWorkStoreHealth` holds that in place.
- **The legacy path is byte-identical.** `controlBinding` is nil on any city that is not a converged split, so non-split cities take the untouched route. Rollback is a clean revert: no migration, no persisted shape change.

Review findings addressed:
- *Test quality / nit* — `registerSplitCityRoutes` handed the residency-route registrar an accessor that built a fresh empty `MemStore` on every call, so the fixture's "city work store" had no stable identity. Inert today because runtime-plane narrowing drops the work leg on a binding topology, but it made the fixture depend on a narrowing rule it never asserts. Fixed in `aaec0bf` by hoisting one store and closing over it.
- *Maintainability / nit* — `convergedRoutedWorkBinding` returned the same `nil` for a leg-resolution error as for a legacy city, so the silence read like a swallowed error. The conservative `nil` is the right behavior (fall back to the legacy target rather than guess a store), so `aaec0bf` documents it instead of changing it, noting that `collectOpenUnassignedRoutedWork` resolves the same leg set later in the build and reports the error there.

- `ga-pil52` — *behavioral correctness / nit / high confidence* — a control dispatcher configured with a custom `scale_check` still has no binding-aware cold-wake probe: both of this PR's substitutions sit under `!hasCustomScaleCheck`, so that branch is unreachable from this diff. This was originally raised as a blocking correctness regression, but the guard it blames predates this PR by about seven weeks (`7d13859`, PR #4175) and the surrounding window is byte-identical to base, so it was de-gated on evidence. The underlying gap is real but pre-existing, and it is floored today by `controlDispatcherOpenDemand`, which reads the correct leg set on a split city — such a dispatcher still spawns; it just does not get the per-bead stamp.
- `ga-bmfmt` — *behavioral correctness / nit / high confidence* — `gc.trigger_bead_store_ref="city"` does not round-trip to the binding in the warm-bind claim nudge. Pre-existing, and strictly improved by this PR; it degrades only the warm-slot backstop, since the cold-`Start` path still delivers the claim nudge.
- `ga-kw6jm` — *test coverage / nit / high confidence* — nothing pins the assumption that a control dispatcher's *execution* never needs the rig work store. True today for graph-class control rows on a converged split, but load-bearing and unguarded if a control step is ever added that must read work-class beads.

### Maintainer Changes
One maintainer fixup commit, `aaec0bf` "review: address non-gating review nits on the split-city dispatcher probe" — `cmd/gc/build_desired_state.go` and `cmd/gc/build_desired_state_test.go`, 2 files changed, 6 insertions(+), 1 deletion(-). Neither item changes behavior. Your commit is preserved unchanged underneath it; the branch was also rebased onto current `main`, with the patch verified identical across the rebase.

### Final Review Status
Ready for the merge queue: final head `aaec0bf` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/33709117742

### Review Iterations
1 review pass performed: two independent reviewers, merged by a synthesizer that re-ran `go vet ./cmd/gc/` and the cited test set in a clean worktree, adjudicated the one contested finding against the code, and resolved to approve. A single maintainer fixup commit closed the two in-scope nits.

---

quad341 (approved):
@julianknutsen — this is a good catch, and the writeup made it checkable. The asymmetry you found is exactly the kind that hides forever behind `max_active_sessions=1`: the rig dispatcher was sized against a ledger holding none of its rows, so its demand carried no `WorkBeadIDs`, the spawned session got no `gc.trigger_bead_id`, and every backstop keyed on that stamp was quietly inert — while the city dispatcher on the same topology got full provenance purely because its target already *was* the binding.

What I especially liked: `convergedRoutedWorkBinding` asks the question of `routedWorkStoreCandidates`, the same leg set collection reads, instead of re-deriving it from config — so demand and collection can't drift apart. And returning a conservative nil for a multi-binding plan is the right call rather than guessing which store holds control work.

I traced the two things that could have made this dangerous and both hold: the override is computed per agent, not per build, and on any city that boots a split the leading store really is the binding pointer (`storageSplitShapeOf` refuses anything else). Your three tests, including the legacy control, do the rest.

Nothing needed from you — merging as-is. Thanks for the careful scoping around #5918; we'll take that one next.


---

PR #5931, opened 2026-09-03
feat(reconciler): emit a counted event when a control scope has no dispatcher

julianknutsen:
Rebased onto `main`; the former #5918 dependency is already upstream. The scope-gap event and its tests remain, with generated Go and dashboard clients refreshed.

## Problem

`repairControlDispatcherRoutesForStoreScope` suppresses a control row whose owning scope configures no control-dispatcher — routing it to another scope's dispatcher would park it on a store that dispatcher cannot read — and reports the gap by printing one stderr line per scope.

The "report once" budget lives on the per-pass repair object, and `buildDesiredState` rebuilds that object every tick. So a standing config gap does not report once: it reports once per scope **per tick**, forever. On maintainer-city that was 600+ identical lines over 26h with every health surface green, and the stall behind them was found by reading them by hand (ga-oytw9). The line also names ONE bead per scope, so one stuck row and a hundred read identically.

## Change

The repair pass now returns a typed summary per scope — the label the stderr line already prints, the owning rig context, the leg the rows were collected through, the count of rows it suppressed, and one sample bead — carried out on `DesiredStateResult.ControlDispatcherScopeGaps`. `CityRuntime.buildDesiredState` turns each entry into one `control.dispatcher_scope_gap` event, so the condition reaches `gc events`, the dashboard and alerting instead of only whoever greps controller stderr. The stderr line is unchanged, for trace readers that already grep it.

Three decisions worth reviewing:

- **Emitted at the FRESH build, not where the result is consumed.** `loadDemandSnapshot` reuses a cached result across stable patrol ticks, so a consumer-side emission would replay an observation that tick never made. Emitting inside `cr.buildDesiredState` covers both fresh-build call sites (startup and the demand snapshot) and keeps each event one-for-one with the stderr line the same build printed.
- **One event per scope, level-triggered.** The gap is a property of the city's config, not of any one bead, so a subscriber reads the current count rather than summing a history.
- **`control.dispatcher_scope_gap`, not `control_dispatcher.scope_gap`.** The design suggested the latter shape; every event type already in the registry spells its domain as a single word, with underscores only in the action segment (`session.demand_claim_divergence`, `control.stalled`). This extends the `control.*` vocabulary `control.stalled` opened rather than adding the first underscore-bearing domain. Say the word if you want the literal design spelling instead.

A misaligned index-aligned input returns no gaps at all: it suppresses every control route without evaluating a single scope, so reporting that blanket suppression as a config finding would be a fabrication. It keeps its own existing diagnostic.

The typed payload is registered next to `ControlStalledPayload` in `internal/events`, so the SSE union, the OpenAPI spec and the generated Go/TS clients pick it up. Those artifacts (`internal/api/openapi.json`, `docs/reference/schema/openapi.{json,txt}`, `internal/api/genclient/client_gen.go`, the dashboard's generated client and `dist/`) are regenerated in this commit.

## Tests

RED→GREEN, in order:

| Test | RED evidence |
| --- | --- |
| `TestEveryKnownEventTypeHasRegisteredPayload` (`internal/api`) | The constant landed before the payload: `unregistered event types: [control.dispatcher_scope_gap]` |
| `TestControlDispatcherScopeGapIsAKnownEventTypeWithATypedPayload`, `TestControlDispatcherScopeGapPayloadRoundTrips` (`internal/events`) | `undefined: ControlDispatcherScopeGapPayload` |
| `TestBuildDesiredState_ReportsControlDispatcherScopeGapForRigRootedWork`, `TestRepairControlDispatcherRoutesCountsSuppressedRowsPerScope`, `TestRepairControlDispatcherRoutesReportsNoScopeGapWhenDispatcherIsConfigured` (`cmd/gc`) | `result.ControlDispatcherScopeGaps undefined` |
| `TestEmitControlDispatcherScopeGapEvents_*`, `TestCityRuntimeBuildDesiredStateRecords*` (`cmd/gc`) | `undefined: emitControlDispatcherScopeGapEvents` |

Coverage: the class-binding fixture with only a city dispatcher now also asserts one gap entry naming the binding and the owning rig with a count of 1; two rows of one scope collapse to one entry with count 2; a scope that HAS a dispatcher reports none, so a healthy city stays silent. On the runtime side, a `CityRuntime` whose build reports a gap records exactly one typed event keyed on the owning scope ref (`rig:fixture` / `city:<name>`, never the class-binding leg — a subscriber keying on the leg could not tell two rigs' gaps apart), a build with no gap records nothing, and a nil recorder does not panic.

**Mutation probe:** replacing the `emitControlDispatcherScopeGapEvents(...)` call in `cr.buildDesiredState` with `_ = emitControlDispatcherScopeGapEvents` leaves every build-side test green and fails only `TestCityRuntimeBuildDesiredStateRecordsControlDispatcherScopeGapEvents` (`recorded events = [], want one control.dispatcher_scope_gap`).

## Gates

`go build ./...`, `go vet ./cmd/gc ./internal/events ./internal/api`, `gofmt -l`, `golangci-lint run ./cmd/gc/... ./internal/events/...`, `./scripts/check-residency-boundary.sh`, `go test ./internal/events/... ./cmd/gc -run 'ControlDispatcher|ScopeGap|Event'`, `TestOpenAPISpecInSync` / `TestGeneratedClientInSync`, `make dashboard-ci`, `make test-fast-parallel` — all clean.

Refs ga-ods2a, ga-oytw9.

## Rebase onto main (2026-09-03)

PR #5918 merged as `f91d19b476`, and the pipeline rebased it on the way in, so
this branch's copy of its five commits no longer matched main and GitHub
reported `dirty`. The branch is now a clean single commit on top of
`origin/main`: #5918's commits (plus the merge commit that carried them) were
dropped, and only this PR's own commit was cherry-picked
(`cherry picked from commit d70c2bf1a9`). The **Base** section above is
historical — the base is `main` and no longer needs retargeting.

Conflicts were confined to regenerated artifacts, not source. `cmd/gc/*`,
`internal/events/*`, `internal/api/openapi.json`, `docs/reference/schema/*` and
the generated Go/TS clients merged cleanly; the SPA bundle under
`internal/api/dashboardspa/dist/` and the generated client's `index.ts`
conflicted only as rename/rename and content churn on build-hashed filenames.
Those were reset to `origin/main` and regenerated from source
(`make spec-ci` → no drift, then `make dashboard-ci` → clean), rather than
hand-merged. The non-`dist/` diff against main is byte-identical in shape to
the pre-rebase commit (1178 insertions, 44 deletions across the same 15 files).

Re-run after the rebase: `go build ./...`, `go vet ./...`, `gofmt -l cmd
internal` (empty), `golangci-lint run` over `./cmd/gc/...`,
`./internal/events/...` and the repo-owned `./internal/api` packages (0 issues;
the only hit under `./internal/api/...` is
`internal/api/dashboardspa/web/node_modules/flatted/golang/pkg/flatted/flatted.go`,
an untracked npm artifact, pre-existing), `./scripts/check-residency-boundary.sh`,
`bash scripts/check-core-boundary.sh`, `go test ./internal/events/...
./internal/api/... -run 'Event|OpenAPI|Payload'`, `go test ./cmd/gc -run
'ControlDispatcher|ScopeGap|DesiredState'` (328 tests, including all six
scope-gap tests), `make spec-ci`, `make dashboard-ci`, `make test-fast-parallel`
— all green.

The pre-push hook ran the fast suite again on the pushed tree and passed;
`--no-verify` was not used.


quad341:

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-699284130204933065

quad341 (approved):
@julianknutsen — thank you for this, and especially for the rebase. Dropping #5918's commits, cherry-picking only your own, and resetting `dist/` to regenerate from source rather than hand-merging build-hashed filenames turned a diff that blew past our review ceiling into something readable. Sorry it took us a while to come back to it.

Merging as-is. What sold it: the count is the thing the stderr line could never carry, and you put the emission where it can't lie — the fresh build, not the cached consumer. I checked that myself, along with the write budget, since a per-tick cap upstream of the counter would have quietly undercounted; it's spent in `persist`, after the gap is recorded. Your naming argument holds too — every domain segment in the registry is a single word, so `control.dispatcher_scope_gap` is the right spelling and we don't want the design doc's version.

Returning `nil` on misaligned input instead of reporting a blanket suppression as a config finding is the kind of restraint that's easy to skip and would have produced a baffling alert someday. Appreciated.


---

Issue #4937, opened 2026-08-03
wisp-compact never runs (bd list omits infra) — and fixing that selector alone deletes the entire read-mail archive


The1nk:
## Before you continue

- [x] I searched existing issues and did not find a duplicate.
- [x] I read the relevant docs and contributor guidance.

Closest existing issues, both open and both **root cause** rather than this call site: #3937 (`bd list` excludes `ephemeral=true` per the TierIssues contract from #3449) and gascity-packs#252. Neither covers `wisp-compact.sh` — #252 enumerates 16 affected call sites and all of them are in the packs repo's `gastown/` tree, while this script ships from this repo. The data-loss consequence below is specific to this site and appears nowhere in either.

_Re-filed into the template; the original body omitted the required fields._

## Gas City version

v1.4.0 (`/home/linuxbrew/.linuxbrew/Cellar/gascity/1.4.0/bin/gc`). Script verified unchanged on `main` via the contents API on 2026-08-03. `bd` 1.1.2 (Homebrew).

## Environment

Ubuntu 26.04 LTS, Linux 7.0.0-28-generic x86_64, bash 5.3.9. Single-workstation city, one HQ store plus one rig.

## Reproduction

`internal/bootstrap/packs/core/assets/scripts/wisp-compact.sh` selects the beads it will act on with:

```bash
ALL=$(gc bd list --json --all -n 0 2>/dev/null) || exit 0
EPHEMERALS=$(echo "$ALL" | jq '[.[] | select(.ephemeral == true)]' 2>/dev/null) || exit 0
if [ -z "$EPHEMERALS" ] || [ "$EPHEMERALS" = "[]" ]; then exit 0; fi
```

Wisps are infra beads, and `bd list` excludes infra unless `--include-infra` is passed:

```console
$ bd list --json --all -n 0                 | jq '[.[]|select(.ephemeral==true)]|length'
0
$ bd list --json --all -n 0 --include-infra | jq '[.[]|select(.ephemeral==true)]|length'
144
```

Then add `--include-infra` to that first line — the obvious repair — and dry-run the delete branch. On this city it selects **139 beads, every one `issue_type: message`**:

```
bl-wisp-dcey4  HANDOFF: PRODUCTION IS DEPLOYED AND HEALTHY. The ETL dry run now reach…
bl-wisp-063ou  BLOCKED: ta-2yl7 needs an architecture decision before any polecat can…
[...]

PR #5933, opened 2026-09-03
fix(mail): respect cfg.Mail.RetentionTTL in the nudge-mail sweep instead of a hardcoded 60m (#5877)

### LINKED ISSUE #5877: Operator mail is destroyed 60 minutes after it is read: the nudge-mail sweeper TTL is hardcoded and cfg.Mail.RetentionTTL cannot disable it
## Symptom

Read mail beads are closed ~60 minutes after being read, stamped
`close_reason = "mail gc-swept: read mail bead past gc retention window"`. In a city
that uses `gc mail` as its operator decision channel this silently destroys operator
directives, not just notices.

Observed in a live city: two decision cards sent to `human` were closed by the sweeper at
the same instant, ~60 minutes after creation, each carrying `mail.read=true`. Neither was
answered, and neither remained visible to `gc mail count human` — a swept card is not
merely marked read, it is gone from the surface the operator reads. One unarchived ledger
segment contained 59 `gc-swept` closures. A later run had the city ask the operator to
re-state a directive it reported as "no longer exists" — that directive had been swept.

The mitigating fact: the payload survives in the append-only event ledger (including
gzipped archives) and is recoverable by seq. But nothing in normal operation reads those
archives, plain `grep` skips `.gz`, and an agent that hits "message not found" is never
told the text is recoverable. It is data loss for every practical reader.

## `[mail] retention_ttl` is not the knob

This is the part worth flagging, because it makes the defect look configurable when it is
not. Setting it was tried in a live city, measured, and reverted with no change.

- `internal/config/config.go` — `MailConfig.RetentionTTLDuration()` returns `0, nil` for
  the empty string, and `"0"` also parses to `0`. **The commented-out state and the
  "fix" are the same value.**
- `cmd/gc/wisp_gc.go` — `newWispGC(interval, ttl, mailRetentionTTL)` bails when
  `ttl <= 0 && mailRetentionTTL <= 0`, so the config-driven purge was already disabled and
  was never the sweeper.

So there is no built-in 60-minute default hiding behind an empty `[mail]` table, and no
city-side configuration change can affect this.

## The actual sweeper

A hardcoded constant, reachable three ways — one of which takes no override at all:

- `cmd/gc/nudge_mail_sweep.go` — `nudgeMailSweepDefaultMailTTL = 60 * time.Minute`
  (close reason emitted from `internal/mail/beadmail/beadmail.go`).
- `cmd/gc/cmd_order.go` — the `sweep-nudge-mail` order defaults `mailTTL` to that constant
  and exposes only `--mail-ttl`; it **rejects `--mail-ttl 0`** with
  `"--mail-ttl must be positive"`. The sweep can be lengthened, never disabled.
- `cmd/gc/city_runtime.go` — `runNudgeMailSweepWatchdog` calls
  `sweepStaleNudgeMail(..., nudgeMailSweepDefaultMailTTL, nudgeMailSweepWatchdogCloseBudget)`
  every `nudgeMailSweepWatchdogInterval` (5m, 500-bead budget) **with no override
  whatsoever**. Lengthening the order's `--mail-ttl` is therefore insufficient — this
  in-process watchdog still purges at 60 minutes.

Verified against `v1.4.0-363-g90522cb72`. Line numbers may drift on `main`; the
identifiers above are the stable reference.

## Suggested fix

Make the mail TTL honour `cfg.Mail.RetentionTTL` (empty/`0` = disabled) at **both** call
sites — the order and the in-process watchdog — rather than only the already-disabled
`wisp_gc` path. Alternatively, or additionally, exempt operator-facing mail
(`from=human` / `to=human`) from the sweep regardless of TTL, since those messages are the
city's decision record rather than transport chatter.

Either way the order's `--mail-ttl 0` rejection needs to change if `0` is to mean
"disabled", to stay consistent with `RetentionTTLDuration`.

## Related

#4937 covers the *other* stage of the same data loss — `wisp-compact` never running, and
the warning that fixing its selector alone "deletes the entire read-mail archive". Between
that and this, read operator mail is destroyed twice over: once at 60 minutes by the
unconfigurable sweeper here, and again by the 24h untyped-wisp TTL there. They are
probably worth fixing together, since a fix to either alone still loses the mail.

rjgeng:
## What

Closes #5877 (p1). The nudge-mail sweep (both the always-on watchdog and the `sweep-nudge-mail` order command) hardcoded a 60-minute TTL, completely ignoring the documented `[mail] retention_ttl` config — and the order command's `--mail-ttl` flag rejected an explicit `0`, so there was no way to disable the sweep at all. `RetentionTTLDuration()` already exists and already gates a separate purge path (`wisp_gc.go`), it just never reached either sweep site.

Severity note: the issue's title says "destroyed" — that's one stage overstated. The sweep only closes the bead (`RetentionSweepCloseReason`), which stays addressable via `gc mail peek <id>`/`bd show` until a later, separate purge pass (tracked as #4937) actually deletes it. It does vanish from every normal read path (inbox, count, list) within the window, which is bad enough to justify p1 — just not yet permanently gone at that point. This fix is scoped to the sweep-close timing/config-respect only; it doesn't touch the separate purge path.

## Fix

Centralized the "unset vs. explicit vs. explicit-0" resolution into one `nudgeMailSweepMailTTLForConfig(cfg, stderr)` helper shared by both call sites: unset config → 60m default (preserves current behavior for every city that's never touched `[mail]`); explicit value including `0` → honored as-is; malformed value → falls back to default with a stderr note rather than failing. `sweepStaleNudgeMail`/`countStaleNudgeMail` now skip the mail-close phase entirely when the resolved TTL is `<= 0`, instead of computing a cutoff of "now" (which would have closed every read-mail bead immediately — a real bug a naive fix would have introduced). `--mail-ttl` no longer rejects `0`; when the flag isn't explicitly passed, the order command loads config via the same lightweight `config.LoadWithIncludes` pattern already used elsewhere in that file (not the heavier path), so no new side effects.

## Files

`cmd/gc/nudge_mail_sweep.go` (+77/-33), `cmd/gc/city_runtime.go` (+3/-2), `cmd/gc/cmd_order.go` (+39/-14), `cmd/gc/nudge_mail_sweep_test.go` (+202, new), `docs/reference/cli.md` (+1/-1, mechanical regen).

## Validation

- `go test -tags gms_pure_go ./cmd/gc/... -run 'TestSweepStaleNudgeMail|TestNudgeMailSweep|TestCmdOrder.*Sweep|TestRunNudgeMailSweepWatchdog|TestCountStaleNudgeMail|TestValidateNudgeMailSweepFlags'` — 39 tests pass
- `go test -tags gms_pure_go ./internal/config/...` pass
- Full local push-gate run: all failures across every shard and `unit-core` confirmed (via source lookup) to be pre-existing environmental/timing flakes (dolt-compaction timeouts, lsof/detached-probe/systemctl-delegate tests, interrupt-timing tests) unrelated to this change — none touch `nudge_mail_sweep.go`, `cmd_order.go`, or `city_runtime.go`


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277288491

quad341 (approved):
@rjgeng — thank you for this, and specifically for the half that's easy to miss. Wiring `RetentionTTLDuration()` into the two sweep sites was the obvious part; noticing that a TTL of `0` would otherwise compute a cutoff of *now* and close every read mail bead on the spot — then guarding both `sweepStaleNudgeMail` and `countStaleNudgeMail` against it — is what kept a config fix from becoming a worse bug than the one it fixed.

The tests show the same care. Seeding a stale nudge bead as a witness that the watchdog body actually ran, rather than letting "the mail stayed open" pass vacuously, is exactly the right instinct. I traced the one thing that could have sunk this — the new `config.LoadWithIncludes` call sitting near that feature-flag-stomping warning in `cmd_order.go` — and it's fine: `resolveCLIStorageRoutes` makes the identical bare call for the identical reason. And thank you for correcting the issue's "destroyed" framing instead of letting it stand; accurate severity makes triage better for everyone.

One thing on my side, not yours: the generated `city.toml` comment in `cmd_init.go` still calls `retention_ttl` purge-only, and `"0"` is the example it shows. You documented the new meaning properly in `config.go` and the order pack — I'll just carry it into the init template (and its byte-exact test), then re-run the build and `cmd/gc` suite. Your commits land as-is.

There's also a red `Evidence / critical-path suites` check and a stale `status/review-failed` label — both ours to clear. Nothing for you to do.


---

PR #5935, opened 2026-09-03
fix(reconciler): release a binding-resident orphan claim through the leg that read it

julianknutsen:
## Bug

`session_reconciler.go`'s orphan-close arm — the ga-jrnou tie-break, reached
only when a seat's runtime is *observably* dead and the close guard refuses
because the seat still holds work — called
`releaseConfirmedOrphanSessionWork`, which resolved each candidate's owner store
with `storeForPoolAssignment`: the rig named by the `gc.routed_to` prefix, else
the rig whose ID prefix matches, else the city store.

Those are all WORK ledgers. On a split city (maintainer-city: every infra class
relocated into one binding at `.gc/store`) the held claim is a graph-class step
that lives ONLY in the binding, so the rig ledger returns `ErrNotFound` from
`liveWorkAssignmentStillReleasable`, the release is skipped, the close guard —
which *does* read the binding leg — keeps refusing, and the dead seat wedges
indefinitely. That is ga-jrnou reopened for every binding-resident row.

The sibling path already gets this right: `releaseOrphanedPoolAssignments`
resolves through `assignedWorkOwnerStore`, which returns the index-aligned leg
the census actually read the row through.

## Fix

The leg that read the row is its owner; a `routed_to` prefix names a work ledger
that no longer holds graph-class rows.

- `releaseConfirmedOrphanSessionWork` takes an `assignedWorkStores []beads.Store`
  parameter (index-aligned with `assignedWorkBeads`) and resolves via
  `assignedWorkOwnerStore`. An absent (nil or empty) slice keeps the existing
  routed fallback, so every caller that supplies nothing is byte-identical to
  today; a non-empty slice of the wrong length is a different snapshot, and its
  out-of-range beads are skipped rather than routed-fallback resolved.
- The controller threads `DesiredStateResult.AssignedWorkStores` to the
  reconciler: through `filterReleasedAssignedWorkSnapshot` (which now projects
  the stores alongside the refs) and the new store-aware
  `filterAssignedWorkBeadsForSessionWakeWithStores`, delivered as the functional
  option `withAssignedWorkStores` rather than a positional parameter. `gc start`'s
  one-shot reconcile takes the same route.
- Alignment is a hard invariant: at the call site a slice whose length does not
  match this tick's snapshot drops to the fallback, logs, and is recorded in the
  close-orphan trace payload (`assigned_work_stores_misaligned`). A misaligned
  slice is never indexed.

No new store enumeration, no class-binding probe in `pool_session_name.go`, no
change to `storeForPoolAssignment` semantics, nothing touched in the residency
stack.

### One thing worth a reviewer's eye (second commit)

The residency contract's AST half flags any `cmd/gc` signature returning a raw
`[]beads.Store`, so the two projections above
(`filterAssignedWorkBeadsForSessionWakeWithStores`,
`filterReleasedAssignedWorkSnapshot`) tripped `TestResidencyResolverBoundary`.
They carry the index-aligned legs they were HANDED through the same keep/drop
decision they already apply to the beads — they open no store and can only
return a subsequence of their own input (the wake filter does resolve claim refs
and rig names to make its keep/drop decision — refs, never legs) — so each got a
`residency:allow <reason>` marker, matching the existing
constructor-input markers (`servingRigStores`, `rigLegsExcludingCityName`).
Baselining the rows would have grown the ratchet's denominator; inlining the
projections to dodge a signature check would have defeated the guard while
keeping the store list. (`scripts/check-residency-boundary.sh`, the grep half,
was green throughout — only the AST half sees signatures.)

## Tests

`cmd/gc/pool_session_release_binding_owner_test.go`

- `TestReleaseConfirmedOrphanSessionWork_ReleasesBindingResidentClaimThroughAlignedStore`
  — the repro: claim resident only in the binding, empty rig ledger, released
  through the aligned leg.
- `TestReleaseConfirmedOrphanSessionWork_NilAlignedStoresKeepsRoutedFallback` —
  the mutation probe: the same call with no aligned slice releases nothing, which
  is exactly the wedge and the documented fallback.
- `TestReleaseConfirmedOrphanSessionWork_LegacyRigStoreUnchanged` — a
  rig-resident row still releases through the routed fallback.
- `TestReleaseConfirmedOrphanSessionWork_MisalignedStoresSkipTheUnalignedBead`
  — a short slice is never indexed past its end; the bead past its end is
  skipped, and the in-range bead still releases.

`cmd/gc/assigned_work_scope_test.go`

- `TestFilterAssignedWorkBeadsForSessionWakeWithStoresProjectsSurvivingStores` —
  the projection itself: a middle bead is dropped and the returned stores are
  compared by handle identity, not length, so a store left behind by its bead
  cannot pass as aligned.

`cmd/gc/session_reconciler_binding_orphan_test.go` (split city via
`seedSplitRoutes`, `runtime.NewFake()`, `t.TempDir()`)

- `TestSessionReconcilerOrphanCloseReleasesBindingResidentWorkThroughAlignedStores`
  — end-to-end through `beadReconcileTick`: work released AND the seat closes in
  the same tick.
- `TestSessionReconcilerOrphanCloseWithAlignedStoresOptionReleasesHeldWork` — the
  option itself, independent of the controller wiring.
- `TestSessionReconcilerOrphanCloseWithoutAlignedStoresLeavesBindingResidentWorkHeld`
  — without the option the seat stays wedged; this is why the plumbing exists.
- `TestSessionReconcilerOrphanCloseIgnoresMisalignedAssignedWorkStores` — no
  panic, no release, fallback path.

All four positive tests were watched RED against the pre-fix resolver (real
assertion failures, not just a compile break) before the body was switched to
`assignedWorkOwnerStore`.

## Gates

- `go build ./cmd/gc` — pass
- `go vet ./cmd/gc` — pass
- `gofmt -l cmd/gc/` — clean
- `golangci-lint run ./cmd/gc/...` — 0 issues
- `./scripts/check-residency-boundary.sh` — pass
- `go test ./scripts -run TestResidency -count=1` — ok
- `go test ./cmd/gc -run 'Orphan|ReleaseConfirmed|ReleaseOrphaned|SessionReconciler|Reconcile' -count=1` — ok
- `make test-fast-parallel` — all 10 fast jobs passed (run by the pre-push hook)

Refs ga-b0o6a

### DIFF
--- a/cmd/gc/assigned_work_scope.go
+++ b/cmd/gc/assigned_work_scope.go
@@ -272,10 +272,14 @@ func filterAssignedWorkBeadsForSessionWake(
 // length is dropped rather than partially applied, and the returned stores are
 // then nil.
 //
-// residency:allow — a caller's own snapshot, projected. It carries the legs it
-// was HANDED through the same keep/drop decision it applies to the beads; it
-// consults no binding, no namespace and no leg order, opens no store, and can
-// only ever return a subsequence of its own input.
+// residency:allow — a caller's own snapshot, projected. The []beads.Store this
+// returns is only ever a subsequence of the slice it was HANDED, carried through
+// the same keep/drop decision it applies to the beads; it opens no store and
+// builds no store list of its own. It DOES consult residency to make that
+// keep/drop decision — assignedWorkClaimRefs resolves the topology's claim refs
+// and assignedWorkStoreRefForAgent resolves each agent's rig name — but those are
+// reads of refs, never of legs, and cannot introduce a store the caller did not
+// already hand in.
 func filterAssignedWorkBeadsForSessionWakeWithStores(
 	cfg *config.City,
 	cityPath string,
--- a/cmd/gc/assigned_work_scope.go
+++ b/cmd/gc/assigned_work_scope.go
@@ -246,6 +246,10 @@ func filterAssignedWorkBeadsForPoolDemand(
 // returns the filtered beads plus their store refs, index-aligned, so callers
 // can resolve store-scoped wake-demand readiness (storeScopedBeadKey) for the
 // surviving beads without re-deriving each bead's originating store.
+//
+// The reconcile paths take the store-aware form below, which also carries the
+// legs the rows were read through; this two-value form is for callers that only
+// read the surviving rows.
 func filterAssignedWorkBeadsForSessionWake(
 	cfg *config.City,
 	cityPath string,
@@ -254,11 +258,36 @@ func filterAssignedWorkBeadsForSessionWake(
 	assignedWorkBeads []beads.Bead,
 	assignedWorkStoreRefs []string,
 ) ([]beads.Bead, []string) {
+	kept, keptRefs, _ := filterAssignedWorkBeadsForSessionWakeWithStores(cfg, cityPath, leading, sessionInfos, assignedWorkBeads, assignedWorkStoreRefs, nil)
+	return kept, keptRefs
+}
+
+// filterAssignedWorkBeadsForSessionWakeWithStores is the store-aware form: it
+// projects the index-aligned snapshot stores through the same filter, so a
+// caller that must WRITE to a surviving row can do it through the leg the census
+// read the row through rather than re-deriving an owner from gc.routed_to (which
+// names a work ledger a binding-resident row does not live in — ga-b0o6a).
+//
+// assignedWorkStores must be as long as assignedWorkBeads; a slice of any other
+// length is dropped rather than partially applied, and the returned stores are
+// then nil.
+func filterAssignedWorkBeadsForSessionWakeWithStores(
+	cfg *config.City,
+	cityPath string,
+	leading beads.Store,
+	sessionInfos []sessionpkg.Info,
+	assignedWorkBeads []beads.Bead,
+	assignedWorkStoreRefs []string,
+	assignedWorkStores []beads.Store,
+) ([]beads.Bead, []string, []beads.Store) {
+	if len(assignedWorkStores) != len(assignedWorkBeads) {
+		assignedWorkStores = nil
+	}
 	if len(assignedWorkBeads) == 0 || len(assignedWorkStoreRefs) == 0 {
-		return assignedWorkBeads, assignedWorkStoreRefs
+		return assignedWorkBeads, assignedWorkStoreRefs, assignedWorkStores
 	}
 	if cfg == nil {
-		return assignedWorkBeads, assignedWorkStoreRefs
+		return assignedWorkBeads, assignedWorkStoreRefs, assignedWorkStores
 	}
 	claimRefs := assignedWorkClaimRefs(cityPath, cfg, leading)
 	reachableRefsByAssignee := make(map[string]map[string]struct{})
@@ -343,6 +372,17 @@ func filterAssignedWorkBeadsForSessionWake(
 
 	filtered := make([]beads.Bead, 0, len(assignedWorkBeads))
 	filteredRefs := make([]string, 0, len(assignedWorkBeads))
+	var filteredStores []beads.Store
+	if assignedWorkStores != nil {
+		filteredStores = make([]beads.Store, 0, len(assignedWorkBeads))
+	}
+	keep := func(i int, wb beads.Bead) {
+		filtered = append(filtered, wb)
+		filteredRefs = append(filteredRefs, assignedWorkStoreRefs[i])
+		if filteredStores != nil {
+			filteredStores = append(filteredStores, assignedWorkStores[i])
+		}
+	}
 	for i, wb := range assignedWorkBeads {
 		if i >= len(assignedWorkStoreRefs) {
 			continue
@@ -353,18 +393,16 @@ func filterAssignedWorkBeadsForSessionWake(
 		}
 		if _, ok := crossStore[assignee]; ok {
 			// City-scoped assignee: reachable from any store (vp-kvp).
-			filtered = append(filtered, wb)
-			filteredRefs = append(filteredRefs, assignedWorkStoreRefs[i])
+			keep(i, wb)
 			continue
 		}
 		if refs := reachableRefsByAssignee[assignee]; refs != nil {
 			if _, ok := refs[assignedWorkStoreRefs[i]]; ok {
-				filtered = append(filtered, wb)
-				filteredRefs = append(filteredRefs, assignedWorkStoreRefs[i])
+				keep(i, wb)
 			}
 		}
 	}
-	return filtered, filteredRefs
+	return filtered, filteredRefs, filteredStores
 }
 
 // readyAssignedFlagsForBeads resolves the store-scoped wake-demand readiness of
--- a/cmd/gc/assigned_work_scope.go
+++ b/cmd/gc/assigned_work_scope.go
@@ -271,6 +271,11 @@ func filterAssignedWorkBeadsForSessionWake(
 // assignedWorkStores must be as long as assignedWorkBeads; a slice of any other
 // length is dropped rather than partially applied, and the returned stores are
 // then nil.
+//
+// residency:allow — a caller's own snapshot, projected. It carries the legs it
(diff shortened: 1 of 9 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277285089

quad341 (dismissed):
@julianknutsen — the diagnosis is the valuable part here. Spotting that the orphan-close tie-break and `releaseOrphanedPoolAssignments` had quietly diverged on owner-store resolution, with only the split-city binding case exposing it, is a genuinely hard catch. A seat wedged for 86 hours because three lanes each correctly deferred to the other two is a miserable thing to chase.

I especially appreciate the fixes you *didn't* make: no new store enumeration, no class-binding probe in `pool_session_name.go`, no loosening of `storeForPoolAssignment`. Just carrying the leg the census already read the row through. And thank you for surfacing the residency guard trip in the PR body instead of quietly baselining the rows — the `residency:allow` markers with a real justification were the right call and made that commit easy to check.

The test pair is what sells it: `..._WithoutAlignedStoresLeavesBindingResidentWorkHeld` passing is what makes the fix's necessity provable rather than asserted. I traced the alignment invariant through both production call sites and confirmed the dispatcher tick passes nil work, so no path is left behind.

One note for the record, no action needed: rig-resident rows in the controller path now also release through the census leg rather than the routed prefix — correct by construction, and the live re-read in `liveWorkAssignmentStillReleasable` means a wrong leg can only skip a release, never misfire one.

Merging as-is.


---

quad341 (approved):
Re-reviewed `c2b5e4842`..`d8a7620c8` (6 new commits since the last review): no new concerns — the prior review stands.

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR fixes a real seat-wedging
bug in the session reconciler: a pool slot whose assigned-work row lived in a
relocated **class binding** could never have its orphaned claim released,
because the release lane resolved owners through `storeForPoolAssignment`
(work-ledger only) while the close guard walked a leg set that *includes* the
binding. Those two lanes disagreeing is the wedge, and the fix makes the release
lane read the row back through the same leg the census read it through. That
matters because the stranded claim is invisible and permanent — the seat never
recovers on its own — and on a split city it is exactly the rows that moved to a
class binding that get stuck.

branch.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- The fix removes an asymmetry rather than adding a parallel mechanism — the
  sibling lane `releaseOrphanedPoolAssignments` already resolved owners through
  the aligned snapshot, and this was the one remaining caller still on the routed
  fallback. Review confirmed that by enumerating all three `reconcileSessionBeads*`
  call sites: the two that carry real assigned work install
  `withAssignedWorkStores`, and the third passes nil assigned work and
  early-returns, so no call site is left on the broken path.
- The bug premise was reproduced independently rather than taken from the PR
  description: reverting only the `assignedWorkOwnerStore(...)` →
  `storeForPoolAssignment(...)` line produced four genuine assertion failures,
  not compile errors.
- Every per-bead safety gate survives the widening. Only *which store is asked*
  changed; the live re-read, detached probe, and identity match are untouched,
  and a wrongly-resolved store degrades to not-found → skip. Failure modes lean
  back toward the pre-fix behavior, never toward a spurious write.
- Index alignment now holds at both ends: the wake filter nils a wrong-length
  slice and projects through a single `keep` closure, the reconciler re-checks
  length against the beads it iterates, and `assignedWorkOwnerStore` bounds-checks
  before indexing.

Review findings addressed (all in `d8a7620c`, "review: correct three alignment
doc claims and pin the store projection" — no runtime behavior changed):
- **[Architectural Consistency] minor, high confidence (Claude)** — the new
  `residency:allow` reason on the wake filter claimed the function "consults no
  binding, no namespace and no leg order", but it resolves the topology's claim
  refs and each agent's rig name. The suppression is still earned (it opens no
  store and can only return a subsequence of its input), so the reason was
  reworded to state the real basis. Worth correcting because the boundary check
  matches the bare marker and never reads the reason, making the text a pure
  human-audit artifact.
- **[Contract & Interface Fidelity] minor, high confidence (Claude, both
  iterations)** — `releaseConfirmedOrphanSessionWork`'s new doc said "Nil **or
  short** keeps the routed fallback". Short does not: out-of-range beads are
  skipped entirely. The doc now matches `assignedWorkOwnerStore`'s own contract
  and the PR's own misaligned-slice test.
- **[Change Impact / Blast Radius] minor, high confidence (Claude)** — the
  justification comment above `nudgeStalledPoolExecution` said the release filter
  "rewrites beads and refs but not stores". True on the base; this PR made it
  false by projecting stores in lockstep. The comment now gives the surviving
  reason (it is the pre-release snapshot).
- **[Test Evidence Quality] minor, high confidence (Claude)** — alignment was
  pinned everywhere it is *consumed* but nowhere it is *constructed*; no test
  called either 3-value form with a non-nil store slice. A projection that
  dropped a bead but not its store would pass every downstream length check and
  release a surviving bead through the dropped one's leg. The new test drops a
  middle bead and compares store *identity*, not length.
- **[Test Evidence Quality] nit, high confidence (Claude, both iterations)** —
  `..._MisalignedStoresReleaseNothing` asserted the opposite of its own body
  (it requires exactly one release, not none); renamed to
  `..._MisalignedStoresSkipTheUnalignedBead`.

Non-blocking follow-ups (shared for visibility — **not** completed in this PR,
each filed as a bead so it outlives this comment):
- `ga-2tmx9`: **[Contract & Interface Fidelity] minor** — pre-existing doc
  inaccuracy in `filterReleasedAssignedWorkSnapshot`: the comment says it drops
  refs rather than guess, but a short ref slice falls back to the original,
  *unfiltered* refs. Pre-existing mechanism in a touched file, and it requires an
  already-misaligned producer to matter, so it was deferred rather than fixed here.
- `ga-takib`: **[Architectural Consistency] minor** — index alignment across the
  four slices that travel together remains a convention held by doc comments and
  length checks rather than a type; every guard catches a *length* mismatch, none
  catches a same-length-but-misordered projection. Also records the retention
  decision for the 2-value `filterAssignedWorkBeadsForSessionWake`, which has no
  production callers after this PR. Well beyond the scope of a fix PR.
- `ga-0lh82`: **[Test Evidence Quality] minor** — the new tests build snapshots
  with the `""` leading-arm ref shape only; the other real shape on a mid-rollout
  split city is a distinct `class:*` census leg. Risk is low (the store projection
  is ref-agnostic and `class:*` is covered separately), but the two halves are
  covered by different tests rather than one that crosses them.

One thing worth knowing for future readers rather than acting on: the tie-break's
blast radius now reaches class-binding rows by design. Before this PR a
binding-resident row was effectively invisible to the release path — that was the
bug — so this is a widened surface, not a known defect, and the live re-read gate
means an incorrect release would need both an identity collision and a matching
live assignee.

Coverage note, in the interest of being straight with you: the third reviewer was
skipped by configuration for this run, so cross-file coverage was reduced. The
cross-file-shaped finding (the unpinned store projection) was caught by
single-file reasoning and re-verified by a direct call-graph check during
synthesis.

### Maintainer Changes
`d8a7620c` — "review: correct three alignment doc claims and pin the store
projection": 5 files changed, 97 insertions(+), 17 deletions(-). Three comment
corrections, one test rename, and one new test pinning the store projection. No
runtime behavior changes; your commits and authorship are preserved.

### Final Review Status
Ready for the merge queue: final head `d8a7620c` has passing required GitHub

The adopted branch was also refreshed against current `main` as a check: it
rebases cleanly, the change set is byte-identical after the rebase, and a focused
run over this PR's tests together with the adjacent reconciler tests from #5928
(which landed on `main` and touches the same function) passes — so the merge
queue's own rebase should be uneventful.

CI: https://github.com/gastownhall/gascity/actions/runs/33817156384

### Review Iterations
2 review passes performed. Both resolved to approve; the second pass re-derived
every finding against the code, adjudicated one disagreement between the passes
at the source, and produced the fix list applied in `d8a7620c`.

---

PR #5937, opened 2026-09-03
fix(reconciler): continuation backstop takes a binding row's owner from its root ref

julianknutsen:
## Bug

A class binding is a city-scope **store** but a many-scope **container** — the
same distinction PR #5918 (ga-oytw9) drew for control-route repair. The
continuation-claim backstop missed it.

`canonicalContinuationClaimStoreRef` folds a `class:*` assigned-work leg to
`city:<name>`, and `evaluateReadyContinuationClaimCandidate` then required both
the step's and the root's `gc.root_store_ref` to equal that leg-derived ref.

On a split city a rig-scoped workflow's steps are minted **into** the binding
carrying `gc.root_store_ref=rig:<name>` (all 247 rig-rooted routed control rows
on maintainer-city are shaped this way). So every rig-rooted step with
`gc.continuation_group` and `gc.session_affinity=require` that was open and
assigned in the binding resolved `continuationCandidateAbsent` on every tick
(`rig:beads != city:maintainer-city`), and the continuation claim nudge
(`cmd/gc/idle_nudge.go`) was silently retired for that whole class of rows.
Pre-existing on `main`; confirmed with overlay probes.

## Fix

- New `continuationClaimOwnerStoreRef(rawStoreRef, rootStoreRef, canonicalStoreRef)`
  answers which canonical ref **owns** a row. On a class leg the owner is the
  row's own `gc.root_store_ref`, accepted only when canonical and local to this
  city: exactly the city ref (city-rooted rows, unchanged) or a well-formed
  `rig:<name>`. Blank, `class:*`, and legacy/non-canonical values fail closed.
  Non-class legs are single-scope, so there the leg-derived canonical ref stays
  the comparator and nothing changes.
- `evaluateReadyContinuationClaimCandidate` runs the root check against that
  owner ref in the same store (the root lives in the binding too) and returns
  the candidate with `StoreRef` = owner ref.
- `canonicalContinuationClaimStoreRef`'s class arm still returns the city ref —
  it is the grouping key and the city-rooted comparator — with its comment
  rewritten to say the per-row owner comes from the row.
- `poolContinuationBackstop.revalidate` already compares both re-reads against
  `target.StoreRef`, so the two sites move together with no second normalizer;
  the doctrine is documented there. No other consumer keys anything by a leg
  ref (`idle_nudge.go:194/332/340` treat `StoreRef` as opaque identity;
  `sameContinuationClaimCandidate` compares owner to owner).

Untouched, as scoped: `rootStoreRefMatchesCandidate` / `repairBead` (already
fixed in #5918), the residency hub files, and no new store enumeration.

## Tests (TDD, RED first)

- `TestEvaluateReadyContinuationClaimCandidate_BindingLegTakesOwnerFromRootRef`
  — 8-case table: binding leg with a rig-rooted row → valid with
  `StoreRef=rig:fixture` and the binding as `Store`; controls that must keep
  their existing answers (binding leg city-rooted row → `city:split-city`,
  legacy `rig:` leg → unchanged) and the fail-closed cases (owner is the class
  leg itself, another city, `rig:` with an empty name, a bare legacy name, and a
  root whose ref disagrees with the step's).
- `TestSelectReadyContinuationClaimCandidates_BindingRigRootedRowIsACandidate`
  — through the public selector with `workStoreRefs=[classRef]`: one candidate,
  `StoreRef=rig:fixture`, `partial=false`.
- `TestNudgeStalledPoolContinuations_BindingRigRootedCandidateIsNudged`
  — end to end: the selector's binding candidate survives revalidation and the
  backstop delivers exactly one nudge, with the durable marker recording the
  owner ref. Before the fix the selector returned nothing and the backstop
  cleared instead of nudging.

RED before the change: all three failed (`resolution = 0` / zero candidates).
GREEN after.

## Mutation probe

Replacing the `storeref.IsClassRef(rawStoreRef)` arm in
`continuationClaimOwnerStoreRef` with `if true` (i.e. deleting the fix) makes
all three new tests fail again with `Absent` / zero candidates, and every other
continuation test still passes — so the new tests are pinned to exactly this
behavior.

## Gates

| gate | result |
| --- | --- |
| `go build ./cmd/gc` | pass |
| `go vet ./cmd/gc` | pass |
| `gofmt -l cmd/gc/` | clean |
| `./scripts/check-residency-boundary.sh` | pass |
| `go test ./cmd/gc -run 'Continuation|IdleNudge|Backstop' -count=1` | 66 tests pass, 0 fail |
| `make test-fast-parallel` | pass |

Refs ga-erfca

## Pre-push hook bypassed

Pushed with `--no-verify` on an operator decision. The only red in the pre-push
suite is `TestRequireNoLeakedDoltAfter_MultipleLeaksReportedSorted` — a
pre-existing host-PID collision tracked as ga-ay6x1 (P1), untouched by this
branch and byte-identical to `main`. It feeds the leak guard synthetic PIDs
50001/50002 while the reap path probes real liveness with
`syscall.Kill(pid, 0)`, and this host currently runs two real `dolt sql-server`
processes at exactly those PIDs. `make test-fast-parallel` passed on this exact
tree at 01:11Z, before those dolt processes appeared, and a separate agent is
now fixing that test.

---
- Related to #3554 — restores the stalled-pool continuation nudge for rig-rooted workflow steps living inside a class binding on a split city, where the backstop that re-drives idle self-claimed pool sessions had gone silent; this repairs one scope-matching case in that backstop rather than delivering the broader auto-advance behavior that issue asks for

### DIFF
--- a/cmd/gc/build_desired_state.go
+++ b/cmd/gc/build_desired_state.go
@@ -5447,28 +5447,14 @@ func selectReadyContinuationClaimCandidates(
 
 	result := make([]ContinuationClaimCandidate, 0, len(order))
 	for _, key := range order {
-		var (
-			valid  []ContinuationClaimCandidate
-			absent bool
-			hold   bool
+		valid, absent, hold := foldReadyContinuationClaimGroup(
+			groups[key],
+			work,
+			workStores,
+			workStoreRefs,
+			key.StoreRef,
+			readyAssigned,
 		)
-		for _, i := range groups[key] {
-			candidate, resolution := evaluateReadyContinuationClaimCandidate(
-				work[i],
-				workStores[i],
-				workStoreRefs[i],
-				key.StoreRef,
-				readyAssigned,
-			)
-			switch resolution {
-			case continuationCandidateAbsent:
-				absent = true
-			case continuationCandidateHold:
-				hold = true
-			case continuationCandidateValid:
-				valid = append(valid, candidate)
-			}
-		}
 		if hold {
 			partial = true
 			continue
@@ -5497,12 +5483,60 @@ func selectReadyContinuationClaimCandidates(
 	return result, partial
 }
 
+// foldReadyContinuationClaimGroup evaluates every leg-local copy of one grouped
+// (StoreRef, ID) pair and reduces the per-row resolutions to the three signals
+// the caller's agreement guard acts on: the candidates whose own owner ref this
+// fold resolved — inside a class binding that owner can be a rig scope rather
+// than the city ref the group is keyed on — whether a same-scope copy disagreed
+// (absent), and whether a copy could not be read at all (hold). A copy another
+// scope owns contributes no signal at all. rows holds that group's indexes into
+// the aligned work/workStores/workStoreRefs slices, and canonicalStoreRef is the
+// group key, not necessarily the owner.
+func foldReadyContinuationClaimGroup(
+	rows []int,
+	work []beads.Bead,
+	workStores []beads.Store,
+	workStoreRefs []string,
+	canonicalStoreRef string,
+	readyAssigned map[storeScopedBeadKey]bool,
+) (valid []ContinuationClaimCandidate, absent bool, hold bool) {
+	for _, i := range rows {
+		candidate, resolution := evaluateReadyContinuationClaimCandidate(
+			work[i],
+			workStores[i],
+			workStoreRefs[i],
+			canonicalStoreRef,
+			readyAssigned,
+		)
+		switch resolution {
+		case continuationCandidateAbsent:
+			absent = true
+		case continuationCandidateHold:
+			hold = true
+		case continuationCandidateValid:
+			valid = append(valid, candidate)
+		case continuationCandidateForeign:
+			// Deliberately no signal. A group keyed on the city ref holds
+			// every leg that serves the city, and on a migrated city that
+			// includes the class bindings, which carry rows other scopes
+			// own. Such a copy says nothing about this group's scope, so
+			// counting it absent would drop the owning leg's candidate and
+			// — partial being snapshot-wide — turn every unrelated
+			// continuation backstop in the city dark.
+		}
+	}
+	return valid, absent, hold
+}
+
 type continuationCandidateResolution int
 
 const (
 	continuationCandidateAbsent continuationCandidateResolution = iota
 	continuationCandidateHold
 	continuationCandidateValid
+	// A row this leg can read but another scope owns. Kept distinct from
+	// "absent" because the fold reads absent as a same-scope disagreement.
+	continuationCandidateForeign
 )
 
 func continuationRowCouldBeCandidate(
@@ -5528,17 +5562,25 @@ func evaluateReadyContinuationClaimCandidate(
 	canonicalStoreRef string,
 	readyAssigned map[storeScopedBeadKey]bool,
 ) (ContinuationClaimCandidate, continuationCandidateResolution) {
+	// Scope is answered before eligibility: a row another scope owns is foreign
+	// to this leg whatever its own state, and a co-resident copy that is merely
+	// stale or not ready would otherwise be reported as an absent row of THIS
+	// scope. Both questions still answer "absent" for a row this leg owns, so
+	// the order is otherwise immaterial.
+	rootStoreRef := bead.Metadata[beadmeta.RootStoreRefMetadataKey]
+	ownerStoreRef, owned := continuationClaimOwnerStoreRef(rawStoreRef, rootStoreRef, canonicalStoreRef)
+	if !owned {
+		if continuationClaimRowIsForeign(rootStoreRef, canonicalStoreRef) {
+			return ContinuationClaimCandidate{}, continuationCandidateForeign
+		}
+		return ContinuationClaimCandidate{}, continuationCandidateAbsent
+	}
 	if !continuationRowCouldBeCandidate(bead, rawStoreRef, readyAssigned) {
 		return ContinuationClaimCandidate{}, continuationCandidateAbsent
 	}
 
 	rootID := strings.TrimSpace(bead.Metadata[beadmeta.RootBeadIDMetadataKey])
-	ownerStoreRef, owned := continuationClaimOwnerStoreRef(
-		rawStoreRef,
-		bead.Metadata[beadmeta.RootStoreRefMetadataKey],
-		canonicalStoreRef,
-	)
-	if rootID == "" || !owned {
+	if rootID == "" {
 		return ContinuationClaimCandidate{}, continuationCandidateAbsent
 	}
 	if store == nil {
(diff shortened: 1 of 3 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277289133

quad341 (dismissed):
@julianknutsen — this is a clean catch. Seeing that a class binding is a city-scope *store* but a many-scope *container*, and that the continuation backstop had inherited the same conflation #5918 fixed for control-route repair, is the kind of bug that stays invisible for a long time because the failure mode is silence: rows resolve `Absent` every tick and nothing logs.

I especially appreciated that `continuationClaimOwnerStoreRef` came out as a pure function, and that you left `canonicalContinuationClaimStoreRef`'s class arm alone as the grouping key instead of forcing one normalizer to do both jobs. That's why the change is strictly widening — I traced all three arms and no row that was ever a candidate changes its `StoreRef`, so no persisted marker goes stale. Your mutation probe reproduces exactly on our side, and CI's `cmd/gc` shards genuinely executed this surface.

Merging as-is. One optional follow-up if you're curious: a table case for a binding row whose root *isn't* in the binding would pin the `Hold` path, which propagates to `snapshotPartial` and pauses the backstop for a whole tick. `order_dispatch.go` co-materializes root and steps so I don't think it's reachable — just untested. Thanks for this one.


---

quad341 (approved):
@julianknutsen — this is a sharp catch, and the part I appreciate most is that the bug's entire failure mode was *silence*: rows resolving `Absent` on every tick, nothing logged, the continuation nudge quietly retired for a whole class of steps. Carrying the #5918 store-vs-container distinction across to the continuation backstop is exactly the right generalization.

Keeping `continuationClaimOwnerStoreRef` a pure fail-closed function, and leaving `canonicalContinuationClaimStoreRef`'s class arm as the grouping key rather than forcing one normalizer to do two jobs, is what makes this reviewable. No previously-valid candidate changes its `StoreRef`, so no persisted marker goes stale. And documenting a pre-existing weakness in `revalidate` that you deliberately *didn't* fix, then filing ga-m4sj2 for it, is the good version of that instinct.

Nothing for you to change. I pulled the `Mac / make test` log: the one failure is `TestInterruptWhileIdleDoesNotExit` in the zcode adapter — a truncated PTY write, unrelated to `cmd/gc` and unrelated to the dolt PID collision you called out. I'll re-run it and merge. Your Linux shards genuinely executed all four new tests and passed.

If you're back in here: a binding row whose root *isn't* in the binding would pin the `Hold` path. Probably unreachable, just untested.


---

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR fixes a real correctness bug in the continuation-claim backstop: it stops the reconciler from reading a binding row's continuation *owner* off the leg the row happened to be read from, and takes it from the row's own root ref instead. That matters because on a split city a class binding is keyed `city:<name>` while the rows it legitimately owns carry a `rig:<name>` owner — so under the old leg-equals-owner reading, a perfectly valid rig-scoped continuation looked like a disagreeing copy, and a single co-resident foreign mirror could mark the whole snapshot partial and silently suppress unrelated backstops across the city.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **Foreign-scope rows were conflated with absent rows.** The first review pass found that `evaluateReadyContinuationClaimCandidate` / `selectReadyContinuationClaimCandidates` had no way to say "another scope owns this copy" distinctly from "this copy disagrees". The final patch adds a fourth resolution (`foreign`) and makes the scope decision precede eligibility, so a copy another scope owns now contributes no signal at all rather than poisoning the agreement guard. `TestSelectReadyContinuationClaimCandidates_CoResidentForeignCopyDoesNotHoldSnapshot` pins exactly that: group key `city:split-city`, surviving candidate `rig:fixture`, snapshot not held.
- **Fail-closed behavior had to be preserved while narrowing.** The narrowing is confined to *exactly-canonical* refs naming another scope; blank, class-valued, non-canonical, and malformed root refs all still resolve `absent`, and an unreadable leg still resolves `hold` and wins over every other signal. The new table test drives all four owner branches plus the reject cases.
- **The revalidate path had to move with it.** Both revalidate comparisons were moved onto the same owner ref, closing the window where the mirror could disqualify a row the evaluator had already admitted.
- **Reviewer-flagged complexity on the grown selector.** The added branching pushed `selectReadyContinuationClaimCandidates` over the project's cyclomatic band. Rather than waive it, the maintainer fixup extracted `foldReadyContinuationClaimGroup`, measured at head versus base with `gocyclo`/`gocognit` (cyclomatic 20 → 16, cognitive 32 → 27) — below the *base* figure, not merely back under the band — and the extraction was verified behavior-preserving line by line.
- **Two comments that misstated the contract.** The `idle_nudge.go` revalidate comment described residency incorrectly, and the new helper's doc comment initially re-asserted the very leg-equals-owner conflation this PR exists to break. Both were rewritten against every branch they describe.

Review findings addressed:
- *Behavioral Correctness / major / claude* — foreign-scope rows conflated with absent rows in the continuation-claim evaluator; fixed in `d29d358e` along with the mixed-leg regression coverage.
- *Readability / minor / claude* — `foldReadyContinuationClaimGroup`'s doc comment asserted that its first return is "the candidates this scope owns", which is false for the PR's own headline case; the replacement text in `d29d358e` states the owner-vs-group-key distinction and covers the `foreign` resolution it had omitted.
- *Debuggability & Operability / minor / claude* — the `idle_nudge.go` revalidate comment misstated the residency contract; corrected in `d29d358e` to state placement-vs-residency, id-preserving migrate, and both legs canonicalizing to `city:<name>`.

One reported issue did **not** result in a change, and that was the right call: a review pass raised that a foreign-only continuation group would now clear the persisted pacing marker with `partial=false`. Re-derivation showed the caller's guard order (`hold` → `len(valid)==0` → `absent`) is byte-identical between base and head, so `absent` was already unreachable in that shape — a foreign-only group yielded `0 candidates, partial=false` before this PR too. The genuine "store unavailable" signal routes through `storePartial` upstream and still holds the snapshot. No behavior change, so nothing to fix.

- `ga-m4sj2` — *Correctness / minor* — key continuation-claim groups by owner ref so cross-group co-residence is actually compared, plus the canonical-ref collapse and the pre-existing within-group first-leg-wins staleness on a pre-relocation residue. Deliberately out of scope here; this PR documents the staleness rather than changing it.
- `ga-2ie8w` — *Correctness / major* — `rootStoreRefMatchesCandidate` carries the same leg-equals-owner conflation on a different consumer, dropping rig-rooted rows on a class leg and hiding rig-scoped routed work from control-dispatcher demand. A distinct call path, so it was tracked rather than folded into this change.

Also carried forward as a documented, accepted trade-off rather than a defect: when two copies of one id disagree on `gc.root_store_ref`, the `foreign` arm deliberately stops the non-owning copy from vetoing the owning leg's candidate. That is a real weakening of the agreement guard for that corrupted state, and the alternative — the status quo — turns every unrelated continuation in the city dark. It is called out at the `foreign` arm in the code so a future reader does not rediscover it as a bug.

### Maintainer Changes
One maintainer fixup commit was pushed to the PR branch on top of your commit, preserving your authorship on `c640a6b7`:

- `d29d358e` fix(reconciler): treat a co-resident foreign continuation row as foreign, not absent — 3 files changed, 241 insertions(+), 34 deletions(-)

It carries the `foreign` resolution and its regression coverage, the `foldReadyContinuationClaimGroup` extraction, and the two comment corrections.

### Final Review Status
Ready for the merge queue: final head `d29d358e` has passing required GitHub checks, and the merge-ready metadata is in place. Local gates were re-run against the current base as well — `go vet ./cmd/gc/` clean, the 26 continuation-backstop tests all passing, and the residency-boundary, core-boundary, split-topology, routed-test-rows, and census-owner-liveness checks all green. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/33826433539

### Review Iterations
4 review passes performed. Pass 1 requested the foreign-vs-absent split and its mixed-leg regression test; pass 2 confirmed the split; pass 3 accepted the behavior and asked for the complexity extraction plus a comment correction; pass 4 validated every fix against source — including confirming the extraction is byte-faithful and re-deriving the one contested finding rather than arbitrating it — and approved.

---

Issue #5825, opened 2026-08-31
usage.jsonl harvester misattributes sessions (cwd+time) and lags flushes one wake


sJohnsonStoever:
# usage.jsonl harvester misattributes sessions (cwd+time heuristic) and lags flushes one wake

**Observed at:** a7297c5 (gc 1.4.0), during a full-ledger retrospective
2026-08-29/30 that cross-checked usage.jsonl against the raw Claude Code
transcripts.

## Summary

Two related defects make usage.jsonl unreliable for per-seat analysis:

1. **Attribution is by cwd+time, not identity.** The harvester booked the
   operator's day-1 bootstrap session (575k tokens, run by hand in the
   city directory) to the mayor. Any human or non-gc session that shares a
   cwd with a seat inside the attribution window is credited to that seat.
   Downstream effect in our retrospective: a phantom "mayor 575k context
   breach" that survived until we traced it to the raw transcript — the
   mayor recycles per wake by construction and can never accumulate that
   context.

2. **Flush timestamps lag one wake.** A session's usage rows are flushed
   on the NEXT wake, so the recorded timestamps belong to the wrong
   activity window. In ledger analysis this manufactures phantom seat
   breaches (usage from a long session lands stamped inside a later,
   short one).

## Suggested shapes

- Attribute by the session banner gc already stamps (`[<city>] <agent> •
  <ts>` as the first user message) or by session id, rather than cwd+time;
  fall back to cwd+time only for unbannered sessions, and mark those rows
  as heuristic.
- Stamp usage rows with the transcript's own request timestamps at flush
  time instead of the flush wake's clock.

PR #5938, opened 2026-09-03
fix(sessionlog): thread the transcript entry's own timestamp through to Fact.At instead of stamping now (#5825)

rjgeng:
## What

Addresses #5825, Defect 2 ("lags flushes one wake"). Defect 1 (cwd+time session-attribution ambiguity) is a genuine design decision — cross-provider fallout, a new persisted-schema field, and a policy call on unmatched fallback rows — declined, not addressed here.

Root cause: Claude transcript entries carry a real per-entry `timestamp` field, and codex rollout entries carry an analogous one already parsed — but only reused as a dedup key, its temporal value discarded. `TailUsage` had nowhere to carry it, so `modelUsageFact` had no choice but to stamp `Fact.At` with `now` (the wall-clock time of the *next* prompt operation that happened to observe the completed turn, not when the turn actually happened). `Fact.At` feeds window-based ledger analysis, which is why the issue reports a "phantom seat breach" — a fact recorded one wake late can land on the wrong side of an analysis window boundary.

## Fix

`tailEntry`/`TailUsage` now carry the transcript entry's own `Timestamp` (plain RFC3339 for Claude, matching `Entry.Timestamp`'s existing decode shape; RFC3339Nano→RFC3339 fallback for codex via the already-existing `parseCodexSessionTime` helper). `modelUsageFact` prefers it for `Fact.At`, falling back to `now` when zero/absent — so this can't regress behavior for older transcript formats or a provider whose tail extraction doesn't populate it. New tests cover both the preferred-timestamp path and the zero-timestamp fallback, plus extraction-level tests for both provider formats.

## Files

`internal/sessionlog/tail.go`, `tail_usage.go`, `codex_usage.go` + 2 test files, `internal/worker/invocation_telemetry.go` + 1 test file. 7 files, 127/11.

## Validation

- `go test -tags gms_pure_go ./internal/sessionlog/... ./internal/worker/...` — both packages pass
- Full local push-gate run: all failures across every shard and `unit-core` confirmed to be pre-existing environmental/timing flakes (dolt-compaction timeouts under heavy system load, lsof/detached-probe/systemctl-delegate tests, interrupt-timing tests in `internal/worker/adapters/zcode` — the last of which also fails on stock `main`, confirmed via `git stash`) unrelated to this change — none touch `tail.go`, `tail_usage.go`, `codex_usage.go`, or `invocation_telemetry.go`

### DIFF
--- a/internal/sessionlog/codex_usage.go
+++ b/internal/sessionlog/codex_usage.go
@@ -255,11 +255,12 @@ func boundedContextPercentage(inputTokens, contextWindow int) int {
 // CLI produces collapse to a single entry (the last observed wins, except a
 // first-observed non-empty Model is kept — a duplicate re-emitted after a
 // model-switching turn_context must not relabel the invocation), and
-// EntryUUID is the line timestamp. token_count lines with null info
-// (rate-limit-only refreshes) and all-zero per-call usage are skipped;
-// malformed lines are tolerated silently. The scan window is the last
-// tailChunkSize bytes, so usage that scrolled past the window is not
-// returned.
+// EntryUUID is the line timestamp, and Timestamp is the same value parsed
+// (RFC3339Nano, falling back to RFC3339) — zero when unparseable. token_count
+// lines with null info (rate-limit-only refreshes) and all-zero per-call
+// usage are skipped; malformed lines are tolerated silently. The scan window
+// is the last tailChunkSize bytes, so usage that scrolled past the window is
+// not returned.
 func ExtractCodexTailUsage(path string) ([]TailUsage, error) {
 	f, err := os.Open(path)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


quad341 (approved):
@rjgeng — two things stand out here. You found the actual root cause instead of patching the symptom: noticing that codex already parsed the entry timestamp and threw its temporal value away after using it as a dedup key only comes from reading the extractor properly. And you split #5825 honestly — declining Defect 1 as a real design decision (persisted-schema field, cross-provider fallout, a policy call on unmatched rows) is exactly right. I'd rather have that conversation on its own than have it smuggled into a bug fix. The zero-timestamp fallback is what makes this safe to land, and your test pins both sides of it.

Two narrow things I'll take on my side rather than send back to you. On the codex path, the duplicate collapse keeps the first-observed model but the last-observed line time — so an invocation just before a model switch still gets dated into the next turn, a smaller version of the same lag. And `tailEntry.Timestamp` being a strict `time.Time` means an off-format timestamp rejects the whole JSONL line rather than falling back to zero, which matters because that decoder is shared with `ExtractTailMeta` for every non-codex provider. I'll fix both, add the regression tests, and re-run the sessionlog/worker suites. Your change and your credit stand. (The automated reviewer's comment is non-blocking — nobody has objected to this.)


---

Issue #5816, opened 2026-08-31
Epoch-bound pending nudges never attempt, never dead-letter, never expire — PENDING corpses accumulate against healthy sessions (no CLI cleanup verb)


brandonmartin:
## Summary

Pending nudges bound to a superseded `continuation_epoch` are never attempted, never dead-lettered, and never expired. They sit `PENDING` with `last_attempt_at=0` indefinitely while the target session is alive and working — and while they pend, operators and witnesses cannot distinguish "session starving" from "queue holding corpses". Observed 17h of `PENDING=4` against a session that was actively merging the whole time.

## Field data (2026-08-31, one city, two rigs, same day)

Rig A refinery: 4 pending nudges created 00:12–01:59Z, bound to continuation_epochs 134/135/137. The session subsequently re-primed (epoch moved on). All four stayed `PENDING`, `last_attempt_at=0`, for ~17 hours — including one sent with `--delivery immediate` (a later immediate nudge to the *current* epoch delivered fine; the epoch-bound one joined the corpses). Meanwhile the session itself was healthy and merged its whole queue.

Rig B refinery: 1 pending nudge (00:22Z) announcing a branch for merge; the work merged and closed at 00:50Z; the nudge was still `PENDING` 18h later.

Every one of the five was content-obsolete by the time anyone could have delivered it. Remedy applied downstream: manual surgery on `.gc/nudges/state.json` (backup, drop entries) — there is no CLI verb for it (`gc nudge` has only `status`).

## Asks

1. **Epoch-mismatch handling**: when a session's continuation_epoch has moved past a pending nudge's bound epoch, either re-bind it to the live epoch (nudges are text; re-binding is usually safe), or dead-letter it with reason `stale_epoch` — anything but silent eternal PENDING.
2. **TTL/expiry**: a pending nudge older than a configurable TTL should dead-letter with reason `expired`, so `status` reflects reality.
3. **An operator verb**: `gc nudge drop/expire <id>` (or `--all-stale`) so cleanup doesn't require hand-editing runtime state under the dispatcher.

Related: #4534 (nudge to drained/terminated session silently queued, no dead-letter — the lifecycle sibling of this epoch case), #5278 (dead-letter prune cost), #3975 (silent wake no-op), #5785 (suspended-rig wake drop). Same design gap in all of them: the delivery pipeline has success-shaped silence where it needs loud failure states.

jacobhausler:
[...]

PR #5939, opened 2026-09-03
fix(nudge): add a drop verb for stale pending nudges instead of requiring manual state.json surgery (#5816)

rjgeng:
## What

#5816 (p2) has three asks. This addresses only the narrow, mechanical one: an operator CLI verb to drop a stale pending nudge, replacing manual `.gc/nudges/state.json` surgery. The other two — auto-detecting an epoch-superseded nudge against a continuously-busy target (the existing fence check only runs inside the delivery-attempt path, which requires session idleness, so it never fires against a busy session), and a TTL/expiry policy — are genuine design decisions (what's a "safe" wait, configurable or not, drop-vs-redeliver-as-default) and are explicitly not addressed here.

## Fix

`gc nudge drop <id>...` (matches the existing `gc mail delete <id>...` multi-ID convention), `--json` supported. Classifies each ID against current queue state, then calls the existing `recordQueuedNudgeFailureWithStore`/`failedQueuedNudge` dead-letter path in one batch — the same machinery `status`/`drain` already exercise internally on a real delivery failure, so a dropped nudge lands in the identical terminal state. No new state-machine logic.

One implementation detail: `failedQueuedNudge` only forces *immediate* dead-letter (vs. requeue-with-retry-delay) for a cause satisfying `errors.Is(cause, errNudgeSessionFenceMismatch)`. Rather than touching that function's branching, the new drop cause implements the standard `Is(target error) bool` method to report equivalence to that sentinel — idiomatic `errors.Is` usage, `failedQueuedNudge` itself is untouched, and the stored `LastError` text still reads "dropped by operator (gc nudge drop)" rather than borrowing the fence-mismatch sentinel's own message.

Per-ID outcomes are distinguished (droppable → dead-lettered, already-dead → clear no-op, not-found → clear error) and a batch with mixed valid/invalid IDs still processes the valid ones.

## Follow-up fixes included in this PR

The full local push-gate run surfaced two completeness gaps a narrower `cmd/gc`-only test pass had missed, both mechanical staleness rather than design decisions:

- **Product-metrics command census**: the new `gc nudge drop` command wasn't registered in `cmd/gc/productmetrics_command_census.json`, so `TestProductMetricsCommandCensusMatchesProductionBuiltins` failed with `live command "gc nudge drop" is missing` — which cascaded into ~30 other product-metrics classification/lifecycle tests that share the census as a precondition. Fixed by adding a manifest entry mirroring the sibling `gc nudge status` entry, then regenerating the derived artifacts via `go run ./cmd/gen-command-census`.
- **Two stale literals surfaced only by the census fix**: `test/packlint`'s deprecated-nudge-form doc-linter flagged the new auto-generated `gc nudge drop <id>... [flags]` docs line because its subcommand allowlist only knew `drain`/`status`/`poll` (added `drop`); `internal/productmetrics`'s `TestInjectedImmutableCommandCatalog...` asserted a hardcoded catalog-entry count (201) that grew to 202 once the new command's ID landed (updated the literal).

## Validation

- `go test -tags gms_pure_go ./cmd/gc/... -run 'TestNudge|TestCmdNudge'` — all pass including 5 new tests
- `go test -tags gms_pure_go ./internal/productmetrics/... ./test/packlint/...` — pass
- `go run ./cmd/gen-command-census --check` passes
- Full local push-gate run (x4 across this session as fixes landed): final run's only failures are pre-existing environmental/timing flakes (dolt subprocess timing family in `examples/bd/dolt/`, systemctl-delegate/detached-probe/lsof tests, interrupt-timing tests in `internal/worker/adapters/zcode`) — confirmed via source lookup none touch `cmd_nudge.go` or any census file, and zero cascade failures across product-metrics/packlint
- Pre-commit lint clean (0 issues) on all three commits


quad341 (approved):
@rjgeng — two things here stand out. You split #5816 honestly: shipped the mechanical drop verb and left the epoch-supersession detection and TTL policy alone, with a clear write-up of *why* those are design calls rather than quietly guessing. And the `Is`-method route into `failedQueuedNudge`'s forced dead-letter branch is exactly right — I checked, it reaches the branch without touching that function, and `LastError` still reads "dropped by operator" instead of borrowing the fence-mismatch message. The test that asserts a dropped item equals what `failedQueuedNudge` produces directly is the cleanest possible proof of "no new state-machine logic". You also caught the census/packlint/schema drift that usually bites new CLI verbs.

Two small things I'll take on my side so this doesn't cost you a round trip: `doNudgeDrop` reports outcomes from the classification pass, so an ID acked between the classify and write locks prints "Dropped" when nothing was dropped — `recordQueuedNudgeFailureDetailed` already returns the real dead-lettered set, so I'll report from that. And `docs/reference/gastown-command-map.md` still lists the namespace as drain/status/poll, so I'll add `drop`. Merging with your commits intact.

If you're up for it later: a test that drops an *in-flight* nudge would close the last gap.


---

PR #5942, opened 2026-09-03
fix(test): leak-guard reap probes an injected liveness, not the real process table

julianknutsen:
## Bug

`TestRequireNoLeakedDoltAfter_MultipleLeaksReportedSorted` fabricated PIDs
50001/50002 and drove the leak guard with the no-op killer
`ignoreProcessSignal` — but the guard's reap hardcoded the **real** liveness
prober `processStillAlive` (`syscall.Kill(pid, 0)`). The killer and the prober
are two halves of one reaper; splitting them meant the guard signalled nothing
while still polling the real process table for pids it had invented.

When a fabricated pid names a live process on the host, the reap therefore
spins its full 5s deadline and appends a `cleanup failed` `Errorf` per pid,
breaking the "exactly one aggregated Errorf" assertion:

```
--- FAIL: TestRequireNoLeakedDoltAfter_MultipleLeaksReportedSorted (5.09s)
    dolt_leak_helper_test.go:211: multiple leaks must be aggregated into one Errorf, got 3
      ... pid 50001 still alive after SIGKILL and 5s wait (leak-guard reap timed out)
      ... pid 50002 still alive after SIGKILL and 5s wait (leak-guard reap timed out)
```

This is not flaky — it stays red for as long as the pid range is occupied. On
the reporting host `/proc/50001` and `/proc/50002` were both live `dolt`
children of one sql-server, so `make test-fast-parallel` failed deterministically
and blocked every push. Previously seen on CI as run 32587221748 (cmd/gc shard
12 of 12).

## Fix

Inject the whole reaper instead of just its killer, so the two halves cannot
be mismatched by construction:

- `requireNoLeakedDoltAfterWithFilterAndKiller(..., killFn)` becomes
  `requireNoLeakedDoltAfterWithFilterAndReaper(..., reap func([]int) []error)`.
  The guard's only use of `killFn` was to hand it to the reaper, so there is
  no longer a seam that accepts one half alone.
- Production (`requireNoLeakedDoltAfterForPaths`) passes `reapDoltLeakPIDs`,
  which pairs `killProcess` with `processStillAlive` at the production
  poll/deadline bounds — identical behaviour to before.
- Scripted-pid guards pass `scriptedDoltLeakReaper(fakeKiller)`, which pairs
  the fake killer with the new `processNeverAlive`. No test path consults the
  real process table for a fabricated pid any more.
- `reapDoltLeakPIDsWithKiller` and `reapDoltLeakProcessesWithKiller` are gone:
  both were "injectable killer + hardcoded real prober", the exact shape of
  this bug, and every remaining caller passed the real `killProcess`.

Defense in depth: fabricated pids now come from `unallocatablePIDs(t, n)`,
which reads `/proc/sys/kernel/pid_max` and returns `pid_max+1 …`. The kernel
never allocates above that ceiling, so collision is impossible by
construction. With the probe injected a collision no longer changes any
outcome; this guards a future call site that re-pairs a fake killer with the
real prober.

Control kept: `TestRequireNoLeakedDoltAfterReportsStillAlivePIDAsCleanupFailure`
drives the guard with a probe that reports the pid alive after SIGKILL and
pins that the aggregation still emits one leak `Errorf` **plus** a separate
cleanup-failure `Errorf` naming the pid — so swapping in `processNeverAlive`
cannot silence a genuine reap failure.

## RED → GREEN

Before, on a host where 50001/50002 are live:

```
--- FAIL: TestRequireNoLeakedDoltAfter_MultipleLeaksReportedSorted (5.09s)
    dolt_leak_helper_test.go:211: multiple leaks must be aggregated into one Errorf, got 3
```

After: `--- PASS ... (0.25s)` — a 5.09s stall becomes 0.25s.

`go test ./cmd/gc -run 'DoltLeak|LeakedDolt|RequireNoLeaked' -count=1 -v` →
**19 `--- PASS`, 0 FAIL, 0 SKIP.**

Both new probes are load-bearing, verified by mutation:

| Mutation | Result |
| --- | --- |
| `processNeverAlive` → `return true` | `MultipleLeaksReportedSorted` fails with the original text: `multiple leaks must be aggregated into one Errorf, got 3` + `still alive after SIGKILL` |
| control's `neverDies` → `return false` | control fails with `expected one aggregated leak Errorf plus one cleanup-failure Errorf, got 1` |

The two mutations fail differently, so neither test is vacuous.

## Gates

| Gate | Result |
| --- | --- |
| `go build ./cmd/gc` | pass |
| `go vet ./cmd/gc` | pass |
| `gofmt -l cmd/gc/` | clean |
| `./scripts/check-residency-boundary.sh` | pass (exit 0) |
| `make test-fast-parallel` | pass (green via the pre-push hook: all 10 jobs, `unit-core` + all 6 `unit-cmd-gc` shards) |

One earlier `make test-fast-parallel` run failed on
`internal/productmetrics/TestRecordOnceKeepsStoredEventWhenSpawnWindowExpires`,
a package this branch does not touch (`git diff origin/main...HEAD` is three
`cmd/gc/*_test.go` files). It passes 5/5 in isolation and passed on the
pre-push rerun; it is a wall-clock `defaultRecordDecisionBudget` flake under
16-way parallelism. Filed separately as `ga-2rlkv`.

Refs ga-ay6x1

### DIFF
--- a/cmd/gc/dolt_leak_helper_test.go
+++ b/cmd/gc/dolt_leak_helper_test.go
@@ -214,7 +214,7 @@ func TestRequireNoLeakedDoltAfter_OnlyNewPIDsInDiff(t *testing.T) {
 	if !strings.Contains(msg, strconv.Itoa(leakedPID)) {
 		t.Errorf("error missing leaked PID %d; got %q", leakedPID, msg)
 	}
-	if strings.Contains(msg, "pid=1000") {
+	if strings.Contains(msg, "pid=1000 ") {
 		t.Errorf("error must not include pre-existing PID 1000; got %q", msg)
 	}
 }
--- a/cmd/gc/dolt_leak_helper_test.go
+++ b/cmd/gc/dolt_leak_helper_test.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"syscall"
 	"testing"
@@ -59,6 +60,36 @@ func (r *recordingTB) runCleanups() {
 	}
 }
 
+// linuxPIDMaxLimit is PID_MAX_LIMIT on 64-bit Linux (1<<22), the hard
+// ceiling /proc/sys/kernel/pid_max can be raised to. Used only as the
+// fallback when that file is unreadable.
+const linuxPIDMaxLimit = 4194304
+
+// unallocatablePIDs returns n ascending pids the kernel cannot have assigned
+// to any process on this host, because they sit above /proc/sys/kernel/pid_max.
+//
+// The leak guard's liveness probe is injected in these tests, so a colliding
+// pid no longer changes any outcome; picking unallocatable pids is defense in
+// depth against a future call site that pairs a fake killer with the real
+// prober again. That pairing is what made ga-ay6x1 deterministically red: the
+// fabricated pids 50001/50002 named two live dolt children on this host, so
+// the reap polled the real process table until its 5s deadline and appended a
+// cleanup-failure Errorf per pid.
+func unallocatablePIDs(t *testing.T, n int) []int {
+	t.Helper()
+	ceiling := linuxPIDMaxLimit
+	if raw, err := os.ReadFile("/proc/sys/kernel/pid_max"); err == nil {
+		if parsed, convErr := strconv.Atoi(strings.TrimSpace(string(raw))); convErr == nil && parsed > 0 {
+			ceiling = parsed
+		}
+	}
+	pids := make([]int, n)
+	for i := range pids {
+		pids[i] = ceiling + 1 + i
+	}
+	return pids
+}
+
 func doltTestProc(pid int, args ...string) DoltProcInfo {
 	configPath := filepath.Join(
 		"/tmp",
@@ -111,8 +142,9 @@ func TestRequireNoLeakedDoltAfter_NoChangeNoError(t *testing.T) {
 // log. This is the regression that originally motivated the helper
 // (3.3 GiB OOM from un-reaped dolt children — see ga-de27g).
 func TestRequireNoLeakedDoltAfter_NewPIDReportedWithArgv(t *testing.T) {
+	leakedPID := unallocatablePIDs(t, 1)[0]
 	leaked := DoltProcInfo{
-		PID:  99999,
+		PID:  leakedPID,
 		Argv: []string{"dolt", "sql-server", "--config=/tmp/Test123/.gc/runtime/dolt.yaml"},
 	}
 	enumerate := scriptedDoltEnumerator(t,
@@ -129,8 +161,8 @@ func TestRequireNoLeakedDoltAfter_NewPIDReportedWithArgv(t *testing.T) {
 		t.Fatalf("expected exactly 1 Errorf, got %d: %v", len(inner.errors), inner.errors)
 	}
 	msg := inner.errors[0]
-	if !strings.Contains(msg, "99999") {
-		t.Errorf("error message missing leaked PID 99999; got %q", msg)
+	if !strings.Contains(msg, strconv.Itoa(leakedPID)) {
+		t.Errorf("error message missing leaked PID %d; got %q", leakedPID, msg)
 	}
 	for _, arg := range leaked.Argv {
 		if !strings.Contains(msg, arg) {
@@ -165,8 +197,9 @@ func TestRequireNoLeakedDoltAfter_PreExistingPIDsNotReported(t *testing.T) {
 // only the new one appears in the error message. This proves the diff
 // is computed (cleanup minus initial), not re-reported in full.
 func TestRequireNoLeakedDoltAfter_OnlyNewPIDsInDiff(t *testing.T) {
+	leakedPID := unallocatablePIDs(t, 1)[0]
 	preexisting := doltTestProc(1000)
-	leaked := doltTestProc(9999, "--leaked")
+	leaked := doltTestProc(leakedPID, "--leaked")
 	enumerate := scriptedDoltEnumerator(t,
 		[]DoltProcInfo{preexisting},
 		[]DoltProcInfo{preexisting, leaked},
@@ -175,13 +208,13 @@ func TestRequireNoLeakedDoltAfter_OnlyNewPIDsInDiff(t *testing.T) {
 	requireNoLeakedDoltAfterWith(inner, enumerate)
 	inner.runCleanups()
 	if !inner.failed() {
-		t.Fatalf("expected leak Errorf for PID 9999; nothing recorded")
+		t.Fatalf("expected leak Errorf for PID %d; nothing recorded", leakedPID)
 	}
 	msg := strings.Join(inner.errors, "\n")
-	if !strings.Contains(msg, "9999") {
-		t.Errorf("error missing leaked PID 9999; got %q", msg)
+	if !strings.Contains(msg, strconv.Itoa(leakedPID)) {
+		t.Errorf("error missing leaked PID %d; got %q", leakedPID, msg)
 	}
-	if strings.Contains(msg, "1000") {
+	if strings.Contains(msg, "pid=1000") {
 		t.Errorf("error must not include pre-existing PID 1000; got %q", msg)
 	}
 }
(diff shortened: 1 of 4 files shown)


quad341 (approved):
@julianknutsen — this is a genuinely good fix, and the reason it's good is the framing: the killer and the liveness prober are two halves of one reaper, and injecting only half is what *created* the bug. Collapsing the seam so the mismatch can't be reconstructed beats swapping in safer pid constants and hoping. I appreciate that you did both, with the comment honestly labelling `unallocatablePIDs` as defense in depth rather than overselling it as the fix.

The control test is what sold me. Injecting `processNeverAlive` everywhere would be worthless if it also swallowed real reap failures, and you anticipated that, pinned it, and showed the two mutations failing with *different* messages so neither test is vacuous. That's the step most people skip.

I checked the production path is byte-identical to before and that no fake-killer/real-prober pairing survives anywhere in the package.

Two nits I'm taking on my side rather than bouncing back: a doc comment on `reapDoltLeakPIDsWithKillerAndWaiter` still names the deleted `reapDoltLeakPIDsWithKiller`, and `"pid=1000"` in `OnlyNewPIDsInDiff` wants a trailing space — it's a prefix of `pid=100000`, so a host with `pid_max=99999` would hit the exact failure class you just fixed. Merging after that. Thanks for chasing this to the root.


---

quad341 (dismissed):
@julianknutsen — this is the good kind of test fix. You didn't just move the fabricated pids somewhere safer and call it done; you noticed the killer and the liveness prober are two halves of one reaper, and that injecting only one half is what *created* the bug. Collapsing the seam into `requireNoLeakedDoltAfterWithFilterAndReaper` means a future call site can't rebuild the mismatch. Labelling `unallocatablePIDs` honestly as defense in depth rather than as the fix is what makes the diff easy to trust.

The control test is what makes this mergeable as-is. Blanket-swapping in `processNeverAlive` would have been a great way to silence real reap failures, and you anticipated that, pinned it, and showed the two mutations failing with *different* messages. We re-ran both on our side and got exactly what you documented.

Also worth saying: the two nits @quad341 offered to absorb are already fixed at `b82a549d` — the stale doc comment is gone and `"pid=1000 "` has its trailing space. Their reasoning about `pid=100000` on a `pid_max=99999` host was right, and you'd already handled it.

Nothing left on our side. Merging now. Thanks for chasing this to the root instead of the symptom.


---

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277296114

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277305312

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR fixes a real hole in the
`cmd/gc` dolt leak guard: the scripted guard tests fabricated pids and then
handed them to a reaper that was hardwired to `processStillAlive`, so they
polled the **real** process table for pids they invented. That is a test that
can fail (or hang for its full deadline) because of whatever happens to be
running on the host — and it matters because this is the guard that protects
the suite from leaked dolt children, so a flaky guard erodes trust in exactly
the signal that is supposed to catch leaks.

PR branch.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **Scripted guard paths no longer touch the host process table.** All three
  reviewers independently traced the pre-fix chain
  (`requireNoLeakedDoltAfterWith` → `...WithFilterAndKiller(..., ignoreProcessSignal)`
  → `reapDoltLeakPIDsWithKiller` → `...AndWaiter(pids, killFn, processStillAlive, ...)`),
  which pinned the real prober behind a no-op killer. The injectable `aliveFn`
  seam already existed from ga-62mu45, but no guard call site paired it — this
  PR is precisely that missing pairing.
- **The mismatch is now unrepresentable, not just avoided.** Injecting the whole
  reaper at the boundary (`requireNoLeakedDoltAfterWithFilterAndReaper`, taking
  `reap func([]int) []error`) means the two sanctioned pairings always keep
  signaling and probing on the same table. That is a structurally stronger fix
  than swapping one prober at one call site.
- **Real reaping is preserved everywhere it matters.** `requireNoLeakedDoltAfterForPaths`,
  the `TestMain` sweep, and `reapTmuxLeakProcesses` all reduce to the identical
  pre-PR `killProcess` + `processStillAlive` + production-bounds chain, so the
  ga-de27g OOM protection is intact.
- **The fix cannot mask a genuine reap failure.** The new control test
  `TestRequireNoLeakedDoltAfterReportsStillAlivePIDAsCleanupFailure` drives a
  deliberately mismatched pairing (no-signal killer + always-alive prober) and
  pins both expected errors in order — so the injection cannot quietly turn the
  guard into a no-op.
- **Defense in depth on top of the real fix.** Moving fabricated pids into an
  unallocatable range derived from `/proc/sys/kernel/pid_max` (with a
  `PID_MAX_LIMIT` fallback that also clears macOS's ceiling) is a genuinely
  second, independent layer — the injected prober alone already closes the
  failure mode. Review liked that the PR framed it that way rather than as the
  fix itself.
- **Clean propagation.** A repo-wide grep found zero stragglers for
  `reapDoltLeakPIDsWithKiller`, `reapDoltLeakProcessesWithKiller`, and
  `requireNoLeakedDoltAfterWithFilterAndKiller`, and the six sibling caller
  files needed zero churn. The tmux one-liner is behavior-identical.

Review findings addressed:
- **[Debuggability] The canonical pairing comment pointed the wrong way.**
  `reapDoltLeakPIDsWithKillerAndWaiter`'s doc said callers go through "one of
  the two pairings above", but only `reapDoltLeakPIDs` is above it —
  `processNeverAlive` and `scriptedDoltLeakReaper` are ~75 lines below. Fixed in
  `7078c637da` by naming the direction of each. (This one is worth calling out:
  the pairing rule is the load-bearing idea of the PR, so a directional error in
  its canonical comment is more costly than a typo.)
- **[Test Evidence Quality] The scripted reap bounds comment credited constants
  for latency they do not govern.** Because `scriptedDoltLeakReaper` hardwires
  `processNeverAlive`, the waiter's first probe always reports exited and neither
  the 5ms nor the 200ms bound is ever reached; the real per-test cost is the
  waiter's unconditional 250ms SIGTERM→SIGKILL grace. Measured during review —
  every scripted leak-path test sits at ~0.25s. Comment corrected in `7078c637da`.
- **[Architectural Consistency] The scripted-only wrappers hid a report-only
  reaper behind generic names.** `requireNoLeakedDoltAfterWith` and
  `requireNoLeakedDoltAfterWithFilter` both hardwire a reaper that reports and
  never kills, and `WithFilter` had no doc comment at all. `7078c637da` states
  the scripted-enumerator/never-kills contract on both and points real-process
  callers at `requireNoLeakedDoltAfterForPaths`.
- **[Architectural Consistency] Two uncross-referenced "give me a dead pid"
  helpers.** `unallocatablePIDs` (derived, never probes, supports n pids) and
  `nonLivePID` (fixed value, probes and skips) now co-occur in the same file with
  different guarantees. `7078c637da` adds a cross-reference to each so the next
  test author does not have to rediscover which guarantee they need.
- **[Architectural Consistency] The PR's own unallocatable-pid norm was not
  applied to its fabricated pre-existing pid slots.** The literal
  `doltTestProc(1000)` slots are safe — the guard diffs pre-existing pids out of
  the leak set before it reports or reaps, so they never reach a killer or a
  prober — but that argument was implicit. `7078c637da` records it in a comment
  rather than churning the value.

- `ga-ozt91`: **[Test Evidence Quality, minor]** the negative assertion
  `strings.Contains(msg, "pid=1000 ")` is an implicit fourth copy of the
  `"  pid=%d argv=%q"` report-line format (emitted at three sites), so a
  one-sided format tweak would silently turn it into an always-pass no-op.
  Remedy is a shared formatter with a derived anchor, or a `pid=1000\b`
  word-boundary match. Deferred out of this PR because it changes executable
  test code, and this head is the review-approved one; filed as a bead so it is
  not lost.

### Maintainer Changes
One maintainer fixup pushed to this PR branch:
`7078c637da test(cmd/gc): document leak-guard reaper contracts from maintainer review`
— comment-only, no executable code changed. 3 files changed, 43 insertions(+),
8 deletions(-).

### Final Review Status
Ready for the merge queue: final head `7078c637` has passing required GitHub
checks (78 passing, 0 failing), and the merge-ready metadata is in place. The
branch was also re-validated against the latest `main` during finalization —
the rebase is a pure base move with an unchanged patch-id, and `go vet`, the
residency-boundary ratchet, and the focused `cmd/gc` leak-guard suite all pass
at that refreshed composition alongside the one intervening `cmd/gc` commit
(#6068). I am marking this PR `status/merge-ready` so the merge queue can pick
it up.

CI: https://github.com/gastownhall/gascity/actions/runs/34167547026

### Review Iterations
1 review pass performed — three independent reviewers plus synthesis, reaching
`approve` with no blocker and no major finding and zero fix iterations required.
Six minor/nit findings were raised; five were landed as the maintainer comment
fixup above and one was filed as a follow-up bead.

---

Issue #4144, opened 2026-07-11
Session reconciler kill-loops agent sessions after config change: asyncStartPreparedCommandStale compares against stale bead metadata (opencode provider unusable)


willpartcl:
## Summary

When an agent's resolved spawn command changes after its session bead exists (provider switch, `args` change, env-affecting patch), the async-start reconciler enters an **infinite start→kill loop** for that session: every start attempt is flagged *"desired command changed during startup"*, the freshly spawned tmux runtime is killed, and the cycle repeats forever (we logged **180 consecutive iterations** on one session). In practice this made the `opencode` provider unusable for us: switching a rig agent from `claude` to `opencode` (or adjusting its args) permanently bricked its sessions, and the killed sessions' beads then held name/alias registrations that blocked replacements.

## Environment

- gc built from `main` @ `9c888ce5` (partcleda/gascity mirror of `gastownhall:main`, 2026-07-10); code paths identical in `v1.1.0`
- Linux (RHEL 9), tmux 3.x, session provider: default tmux
- Agents: `claude` provider works fine; failure triggered when patching a rig-imported agent to `provider = "opencode"`

## Reproduction

1. City with a rig-imported agent (e.g. `[[agent]] name = "worker"`, provider claude) and running named sessions.
2. Change the agent's spawn command via config — e.g. rig patch:
   ```toml
   [[rigs.patches]]
   agent = "worker"
   provider = "opencode"
   ```
3. Restart the supervisor; kill (or wait for restart of) the existing sessions.

**Observed:**
```
session lifecycle: op=start wave=0 session=chip_design__mutator-cc-8gnr0 template=fpmul/chip_design.mutator outcome=start_enqueued duration=0s
session reconciler: ignoring stale async start result for chip_design__mutator-cc-8gnr0: desired command changed during startup
session lifecycle: op=start wave=0 session=chip_design__mutator-cc-8gnr0 template=fpmul/chip_design.mutator outcome=async_start_refresh_failed duration=93ms
```
…repeating indefinitely (~every reconcile tick; 180 times before we intervened). The spawned tmux session is explicitly killed each iteration (`stopStaleAsyncStartRuntime`), so with `remain-on-exit on` there is still no corpse to inspect.

Secondary damage: sessions killed mid-loop leave beads whose `session_name`/`alias`/`agent_name` metadata still hold the identity, so replacement sessions fail with
```
[...]

PR #5943, opened 2026-09-03
fix(session): a drifted pending create must release its alias, not strand it in `creating` forever

julianknutsen:
## Summary

A named session (olivia) sat unstartable for 2.5 hours holding its alias. Root cause, established by measurement rather than inspection: a **model bump** (`claude-opus-4-8` → `claude-opus-5`) made the async-start config-drift gate reject the create — `desired command changed during startup`, a pure string compare, hence `commit_refresh=1ms` — with the *superseded* disposition, which tears down the runtime it just spawned but **keeps the pending-create claim and the alias**. Nothing rewrites a `creating` row's persisted command, so the compare failed permanently: 111 start attempts, each ~9s, all silent (`logLifecycleOutcome` recorded no events). A replacement could not be created either (`alias already exists`), so the deadlock was self-sustaining. The one-minute stale-creating reaper never fired because `PreWakePatch` re-stamped the episode marker on **every** attempt — a bound that renews is not a bound.

**Fix:** a pending create that drifts takes a **rollback** (release claim and alias) instead of superseded; live rows keep today's behavior exactly. `PreWakePatch` preserves an existing marker so the reaper is reachable, and the marker is cleared wherever a row leaves the episode. `gc session reset` routes through the same rollback so it can finally rescue this state. Two typed events (`session.async_start_drift_rolled_back`, `session.async_start_refresh_stalled` at threshold 3) end the silence; their payload carries SHA-256 **fingerprints** of the commands, never the argv, because a resolved session command routinely carries provider credentials and these events reach `events.jsonl`, SSE and the dashboard.

## Review

Design → implementation → adversarial review: **1 critical + 5 major + 6 minor, all fixed**. The critical: the rollback was ordered *ahead* of the identity/lease fence, so a late attempt could close the bead and free the alias of a **newer, live** incarnation — harmless before this change (the drift arm wrote nothing), destructive after. Also fixed: reset could close a live human-serving agent's bead with no liveness probe (the documented post-start-ApplyPatch shape leaves a live runtime in `creating` for warm reuse); the alias was freed even when the runtime survived its stop; reset reported success on a failed rollback. Two of the prescribed fixes were themselves corrected by red tests — a stopped session still answers `GetMeta`, and the sweep wrapper's nil clock would have made the rescue permanently unreachable.

## Testing

10 new pins, every one watched failing on behavior first; 8 mutation probes, each killing its own pin; `golangci-lint` 0 issues; `make dashboard-ci` exit 0; race clean. One pre-existing environment-sensitive leak-guard failure baselined against a clean `origin/main` worktree.

Refs: ga-6wkhl

---
- Related to #4144 — covers the never-converging "desired command changed during startup" loop for a session whose create never committed — the drifted pending create now releases its claim and its alias instead of retrying forever, and the previously silent retries emit typed events; rows whose create already committed keep the existing superseded behaviour, so the same drift on an already-live session is not covered here


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277294588

quad341:

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-699284130204932992

julianknutsen:
## Adoption status — review complete, merge held on a ratchet decision

Thanks for this one, @julianknutsen. The review/fix loop is finished and the
change is approved: across four passes reviewers pushed on whether the
pending-create rollback could tell "this runtime is genuinely gone" apart from
"I could not observe it", and the final patch answers that — the `ListRunning`
contract now says an unobservable session must surface as a partial error rather
than an absent name, the providers implement it, and the rollback fences on
identity, liveness, and outcome before releasing an alias. That is the
difference between reclaiming a dead `creating` row and stranding a live agent,
so it is worth getting exactly right.

I have pushed the adopted branch (rebased onto current `main`, plus one
maintainer fixup that makes the liveness fence actually establish "no live
runtime"). **I am not marking this merge-ready yet**, for one reason that is not
a defect in your change:

`TestRepositoryLedgerMatchesCensusAndDocumentation` is red because the
`scope=untagged` / `net_listen` resource ratchet has no headroom left:

```
source resource census grew: scope=untagged resource=net_listen calls=102 (baseline 98), files=40 (baseline 38)
Small resource census grew:  scope=untagged resource=net_listen calls=100 (baseline 96), files=39 (baseline 37)
```

The `+4 calls / +2 files` are this PR's two new tests
(`internal/runtime/acp/listrunning_partial_test.go` and
`internal/runtime/subprocess/listrunning_partial_test.go`), and they fit under
the baseline on the branch's original base. Since then `main` grew its own
listener usage and raised the baselines to exactly its own count, so the budget
these tests used to fit inside is gone. Removing just those two files makes the
ledger test pass, which pins the arithmetic.

I did not fix it by raising the number. `test/test-resources.toml` reserves that
for council review, and the row's invariant is literally "untagged
stream-listener call/file totals cannot grow" — so raising it is the decision
the gate exists to force, not a repair an automated lane should make on its own.
Dropping the listeners is not an option either: they *are* the test shapes
(a listener that accepts and never answers = alive but unobservable; a
bound-then-closed socket = a refused dial, positive proof of absence).

So this needs a human call on whether the new contract coverage earns the
`+4 / +2`. If yes, four pinned surfaces move together: the `[[debt]]` row in
`test/test-resources.toml`, the matching `bootstrapPolicy` literal in
`internal/testpolicy/resourcecensus/census.go`, the `TESTING.md` mirror row, and
either the `[[small_debt]]` surfaces or two `[[medium]]` exact-owner rows for the
new tests. Worth running the whole `resourcecensus` package afterwards, not just
the ledger test — `census_test.go` pins the same numbers per resource.

Everything else is green on `f32579fa85`: CodeQL, cyclomatic complexity, docs
render, vulnerability scan, and review formulas. Tracking the ratchet decision
in `mc-3oce3`.

PR #3324, opened 2026-06-10
feat(gc): bound concurrent bd subprocesses with an admission semaphore

bourgois:
## Summary

> **Stacked on #3319** (breaker wiring). Review/merge that first — its commits appear in this diff until it lands.

Adds an **admission semaphore** at the bd-subprocess chokepoint so the controller bounds how many `bd` child processes it spawns concurrently, instead of fanning out unboundedly under load (which piles up subprocesses and starves the box). Exposes a `gc_bd_inflight` gauge for observability.

- `feat(gc)` — `bd_admission.go`: a counting semaphore gating bd subprocess spawns, with the `gc_bd_inflight` gauge surfaced through the API state (+regenerated OpenAPI/client/dashboard types).
- `fix(bd-admission)` — on saturation the controller tick **fails fast** (returns a typed unavailable) instead of blocking on the semaphore, so a saturated store degrades gracefully rather than wedging the controller. Adds `max_admission_wait` config.

Opt-in via `[beads.resilience]`; with admission disabled it is a pass-through.

> **Note:** the related *store-health patrol* (the other half of the original P1 plan) is intentionally **not** in this PR — it compile-depends on the fork's db-proxy lifecycle (`beads_proxy_reap.go`, PR #3165), which isn't upstream. That piece needs decoupling before it can be proposed.

## Testing

- [x] `go build ./...`, `go vet`, `golangci-lint` (0 issues)
- [x] `go test ./cmd/gc/` (admission semaphore, gauge, fail-fast) + `./internal/config/` (admission config) pass locally
- [x] OpenAPI + dashboard types regenerated and in sync (`genspec`/`npm run gen`)
- [ ] Full `make check` — CI

## Checklist

- [x] Added tests (semaphore bounding, fail-fast, config)
- [x] Opt-in; disabled = pass-through (no behavior change until configured)
- [ ] No breaking changes

PR #5945, opened 2026-09-03
feat(storehealth): deterministic quarantine directory naming

bourgois:
Part of the split of #3324 (closed) into independently reviewable pieces.

## What

Adds `internal/storehealth/quarantine.go` — a small, pure helper for naming
the forensic quarantine directory that a store-health reap writes its
evidence into.

- `QuarantineDirName(scope, seq)` — the name is a pure function of its two
  arguments.
- `NextQuarantineSeq(root, scope)` — the sequence is a monotonic counter
  derived by scanning existing `<scope>-<n>` entries under the quarantine
  root and returning `max(n)+1`, **never** a wall-clock or random value.
- `QuarantineDirPath(root, scope)` — joins the two; it does not create the
  directory (the forensics hook does that).

## Why the sequence is derived, not generated

Making the name a function of on-disk state rather than of the clock is
what makes reap artifacts deterministic and replayable, and it lets tests
assert exact names instead of matching a timestamp pattern. It is also
monotonic rather than gap-filling: deleting an intermediate directory does
not cause the next reap to reuse that number, so a new reap's artifacts can
never land on top of a prior reap's.

Scope roots are absolute paths, so the scope is collapsed to a
filesystem-safe single path component: separators and other awkward
characters become `_`, while the full path is preserved so two distinct
scopes can never collide.

## Scope

104 lines of implementation, 88 of tests. It stands alone: nothing else in
the tree references it yet, and it imports only the standard library
(`fmt`, `os`, `path/filepath`, `sort`, `strconv`, `strings`). The only I/O
is a single `os.ReadDir` of the quarantine root, and a read error there
yields seq 0 so the first reap after a missing root still produces a stable
name.

## A note on the earlier review

An earlier attempt at this area was closed as "proxy-coupled, not cleanly
separable". That was accurate about the *feature*, but not about this
*package*. This file has no proxy coupling at all — see the import list
above. The proxy coupling lives entirely in a separate file that is not
part of this PR (or of its follow-up), and none of it is proposed here.

## Verification

```
gofmt -l internal/storehealth/   # clean
go build ./...                   # ok
go vet ./internal/storehealth/   # ok
go test ./internal/storehealth/  # ok (3 new tests, 24 total in package)
```

### DIFF
--- a/internal/storehealth/quarantine.go
+++ b/internal/storehealth/quarantine.go
@@ -1,0 +1,104 @@
+package storehealth
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"sort"
+	"strconv"
+	"strings"
+)
+
+// QuarantineDirName builds the forensic quarantine directory name for a
+// scope and sequence number: "<scope>-<seq>". The sequence is a monotonic
+// counter, NEVER a wall-clock or random value, so the names are
+// deterministic and tests can assert them exactly (plan item 1.5).
+//
+// scope is sanitized to a filesystem-safe token (path separators and other
+// awkward characters collapse to '_') because scope roots are absolute
+// paths.
+func QuarantineDirName(scope string, seq int) string {
+	return fmt.Sprintf("%s-%d", sanitizeScopeToken(scope), seq)
+}
+
+// NextQuarantineSeq returns the next monotonic sequence number for a scope
+// by scanning existing "<scope>-<n>" entries under quarantineRoot and
+// returning max(n)+1, or 0 when none exist. Derivation from directory
+// entries (rather than a wall-clock or random source) keeps reaps
+// deterministic and replayable. A read error on the root yields 0 so the
+// first reap after a missing root still produces a stable name.
+func NextQuarantineSeq(quarantineRoot, scope string) int {
+	token := sanitizeScopeToken(scope)
+	entries, err := os.ReadDir(quarantineRoot)
+	if err != nil {
+		return 0
+	}
+	prefix := token + "-"
+	maxSeq := -1
+	for _, e := range entries {
+		name := e.Name()
+		if !strings.HasPrefix(name, prefix) {
+			continue
+		}
+		n, err := strconv.Atoi(strings.TrimPrefix(name, prefix))
+		if err != nil || n < 0 {
+			continue
+		}
+		if n > maxSeq {
+			maxSeq = n
+		}
+	}
+	return maxSeq + 1
+}
+
+// QuarantineDirPath joins the quarantine root with the next sequenced
+// directory name for a scope and returns the absolute path plus the seq.
+// It does not create the directory — the forensics hook does that.
+func QuarantineDirPath(quarantineRoot, scope string) (path string, seq int) {
+	seq = NextQuarantineSeq(quarantineRoot, scope)
+	return filepath.Join(quarantineRoot, QuarantineDirName(scope, seq)), seq
+}
+
+// sanitizeScopeToken collapses a scope root path into a filesystem-safe,
+// stable token: leading separators are dropped and every remaining
+// separator or awkward character becomes '_'. Two distinct scopes never
+// collide because the full path is preserved (only the separators change).
+func sanitizeScopeToken(scope string) string {
+	scope = strings.TrimSpace(scope)
+	scope = strings.Trim(scope, string(filepath.Separator))
+	if scope == "" {
+		return "scope"
+	}
+	var b strings.Builder
+	for _, r := range scope {
+		switch {
+		case r >= 'a' && r <= 'z', r >= 'A' && r <= 'Z', r >= '0' && r <= '9', r == '.':
+			b.WriteRune(r)
+		default:
+			b.WriteByte('_')
+		}
+	}
+	return b.String()
+}
+
+// sortedQuarantineSeqs returns the existing sequence numbers for a scope in
+// ascending order. Helper for tests asserting deterministic accumulation.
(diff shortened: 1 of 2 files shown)


quad341 (changes requested):
@bourgois — thank you for splitting #3324 into reviewable pieces, and specifically for deriving the sequence from on-disk state instead of the clock. The gap-deletion test is the one I'd have asked for: it pins `max+1` as a *property* rather than asserting non-reuse in a comment, which is what actually keeps a new reap off a prior reap's evidence. Everything checks out on your head — `gofmt`, `vet` and the package tests are clean, and the "no proxy coupling" claim holds; I verified the import list and confirmed nothing in the tree references the new symbols.

One real thing, and I'm asking rather than pushing it myself only because your fork is organization-owned — GitHub doesn't grant maintainers write access to org-owned forks, so a commit from me gets rejected at push time. (It reports `maintainerCanModify=true` anyway, and ticking "Allow edits by maintainers" won't change it. Nothing to do with your work.)

`sanitizeScopeToken` maps every character outside `[A-Za-z0-9.]` to `_`, not just separators, so it isn't injective:

```
"/city/rigs/gc-management"  -> "city_rigs_gc_management"
"/city/rigs/gc/management"  -> "city_rigs_gc_management"
"/city/rigs/gc management"  -> "city_rigs_gc_management"
"/city/rigs/gc_management"  -> "city_rigs_gc_management"
```

Hyphenated rig directories are the norm here, so it's reachable. It doesn't clobber artifacts — the seq still increments over the shared prefix — but two stores end up sharing one sequence namespace, and a reader can't tell which scope produced `city_rigs_gc_management-3`. That's the forensic point of the name, and it's the invariant your own doc comment and PR body promise.

Concretely, in `internal/storehealth/quarantine.go` — add `"crypto/sha256"` and `"encoding/hex"` to the imports and replace the function:

```go
const maxScopeTokenLen = 64

// sanitizeScopeToken collapses a scope root path into a filesystem-safe,
// stable token. The collapse is lossy (every character outside
// [A-Za-z0-9.] becomes '_'), so the token carries an 8-hex-char digest of
// the RAW scope: distinct scopes can never share a token even when they
// differ only in punctuation. The sanitized prefix is capped so a deep
// scope root cannot push the directory name past NAME_MAX.
func sanitizeScopeToken(scope string) string {
	scope = strings.TrimSpace(scope)
	sum := sha256.Sum256([]byte(scope))
	digest := hex.EncodeToString(sum[:])[:8]

	trimmed := strings.Trim(scope, string(filepath.Separator))
	if trimmed == "" {
		return "scope." + digest
	}
	var b strings.Builder
	for _, r := range trimmed {
		switch {
		case r >= 'a' && r <= 'z', r >= 'A' && r <= 'Z', r >= '0' && r <= '9', r == '.':
			b.WriteRune(r)
		default:
			b.WriteByte('_')
		}
	}
	token := b.String()
	if len(token) > maxScopeTokenLen {
		token = token[:maxScopeTokenLen]
	}
	return token + "." + digest
}
```

The `.` separator (not `-`) keeps the `token + "-"` prefix parse in `NextQuarantineSeq`/`sortedQuarantineSeqs` unambiguous, and the allowed alphabet is pure ASCII so the length cap can't split a rune. Please also update `QuarantineDirName`'s doc comment, which repeats the "only the separators change" claim.

That said — you're writing the consuming forensics reader, so if you'd rather have a *reversible* encoding (percent-escaping, say) than a digest suffix, take that instead. Recovering the original path from a directory name is a property worth having and it's your call, not mine. Either shape closes the finding.

Two smaller things, both optional from my side:

- `NextQuarantineSeq` returns 0 on *any* `os.ReadDir` error, but the comment justifies it as "a missing root." On `EACCES` or an I/O error the root may be full of prior artifacts and the next reap targets seq 0 — on top of them. Cleanest is returning `(int, error)` and reserving 0 for `errors.Is(err, fs.ErrNotExist)`; since nothing calls this yet, the signature is still free. If you'd rather not move the signature before the hook lands, just widen the comment to "missing *or unreadable*" so you don't mislead yourself in the follow-up.
- `sortedQuarantineSeqs` is test-only but lives in the production file (and pulls `sort` in with it). Moving it to `quarantine_test.go` is tidiness, not correctness — skip it if it adds churn.

And a test that would have caught the collision, plus the two adjacent gaps:

```go
if QuarantineDirName("/a/b", 0) == QuarantineDirName("/a-b", 0) {
	t.Fatal("scopes differing only in punctuation collided")
}
if QuarantineDirName("/a b", 0) == QuarantineDirName("/a_b", 0) {
	t.Fatal("scopes differing only in punctuation collided")
}
// A deep root must still produce a name os.Mkdir accepts.
deep := "/" + strings.Repeat("segment/", 64) + "leaf"
if name := QuarantineDirName(deep, 0); len(name) > 255 {
	t.Fatalf("name is %d bytes, exceeds NAME_MAX", len(name))
}
// Empty/whitespace scope falls back rather than producing a bare "-0".
if got := QuarantineDirName("   ", 0); !strings.HasPrefix(got, "scope.") {
	t.Fatalf("empty scope fallback = %q", got)
}
```

Verify with:

```
gofmt -l internal/storehealth/
go vet ./internal/storehealth/
go test ./internal/storehealth/ -count=1 -run Quarantine -v
go build ./...
```

Push that and I'll merge straight away — the rest of this is ready. One heads-up unrelated to you: the red `Evidence / critical-path suites` check fired at 09:06, before CI's real suites ran (09:43–11:50, all green). It looks like a watchdog that ran ahead of its evidence; your next push should clear it, and I'll re-run it if not. Your commit and authorship stay as they are.


---

Issue #5846, opened 2026-09-01
reconciler: canonicalizeLegacyBound* rewrites routes and assignees of another city's beads on a shared rig store — no ownership check, no CAS, no audit


A3Ackerman:
## Symptom

Two cities whose `[rigs]` entries point at the same remote bead store. City A's controller rewrites `gc.routed_to` — and, for in-flight work, the assignee — of beads that belong to city B; B's next tick rewrites them back. On our fleet the rows oscillated once per tick pair, work assigned to a live session on B was reassigned away from it, and nothing recorded any of it: no audit row, and the `canonicalizeLegacyBound*` stderr lines fire only when a write fails.

## Root cause

`cmd/gc/build_desired_state.go` on `main`:

- `canonicalizeLegacyBoundAssignedWork` (`:5053`): status open/in_progress, non-empty assignee, `isKnownPoolTemplate(assignee, cfg)` (`:5076`), then `store.Update` with the local canonical spelling (`:5094-5099`). Unconditional.
- `canonicalizeLegacyBoundUnassignedRoutedWork` (`:5131`): open, unassigned, non-empty `gc.routed_to` (`:5143-5146`); rewrite when `normalizeAgentTemplateIdentity(cfg, routedTo)` differs (`:5153-5158`). Unconditional.

Every gate is a string question asked of the *local* config. `normalizeAgentTemplateIdentity` resolves through `findAgentByTemplate` (`cmd/gc/session_reconcile.go:365`), whose second pass, `legacyBoundTemplateMatchesUnboundAgent` (`:392-405`), accepts any binding prefix:

```go
binding, unbound, ok := strings.Cut(local, ".")
…
return strings.TrimSpace(unbound) == strings.TrimSpace(agent.Name)
```

So B's canonical `qcore/pool.omp` matches A's unbound `qcore/omp`, A rewrites the row to `qcore/omp`, and B rewrites it back. The live-owner guard at `:5089` keys on A's own session snapshot, so B's live session reads as absent. Both call sites (`:801`, `:812`) run every reconcile tick over `censusStoreCandidates` — the city store plus every attached rig store — and the only scoping is `rootStoreRefMatchesCandidate` (`:5278`), a store-ref match that both cities satisfy.

The orphan sweeper uses the same resolver (`cmd/gc/pool_session_name.go:223-226`), so it is not a reusable guard as it stands.

## Reproduction

[...]

PR #5952, opened 2026-09-03
fix(pool): treat a foreign <rig>/<name> claim as unobservable, not dead

### LINKED ISSUE #5959: pool orphan sweeper validates federated claims against unfederated evidence — releases another city's live claims on a shared rig store
### Gas City version

Observed on builds from main through 59d581a0f (mechanism present in `cmd/gc/pool_session_name.go` release paths).

### Environment

Two cities (macOS/darwin both sides) sharing one federated rig store on a Dolt hub; each city runs its own supervisor/reconciler against the shared store.

### Reproduction

1. City A's crew member (a named identity configured only in city A) claims work on the shared rig store: `assignee = <rig>/<name>`, status `in_progress`.
2. City B's reconciler tick runs `releaseOrphanedPoolAssignments` over the shared store's assigned work.
3. `liveOpenSessionAssignmentExists` answers from city B's **local** session stores only; city A's identity has no session bead there, so the claim reads as dead.
4. `releaseOrphanedPoolAssignment` → status reset to `open`, assignee cleared. No event, no comment trail.

The confirmed-orphan tie-break (`releaseConfirmedOrphanSessionWork`) can reach the same release through stored session-bead metadata that collides with a foreign identity.

### Expected behavior

A claim held by an identity the sweeping city does not configure is **unobservable**, not dead: skipped, counted, and named — never released.

### Actual behavior

Both cities silently stripped each other's live claims. Measured incidents 2026-08-21 on the shared store: one direction cleared two P0s and a P1 in 8 seconds; the other cleared fifteen claims in 34 seconds. Each strip was reconstructed by archaeology because the release writes no event.

Calibration against one city's full event log (2,874 `bead.dead_assignee_reopened` events, 227 distinct assignees): exactly 1 of 2,686 historical gated reclamations was a cross-city strip; the rest were legitimate local reclamations that must keep working — which bounds the fix's allowed blast radius.

### Additional context

Fix: #5952 (roster-membership gate ahead of the liveness question on both Go release paths; skips are counted and named once per sweep). Sibling arms of the same incident family, tracked separately: #5955 (the core pack's shell sweeper re-implements the same decision unguarded), #5846 (shared-store canonicalize write sites rewrite the other city's assignee/routed_to; fix applies #5952's predicate at those write sites).

https://claude.ai/code/session_01QB58oPyfY5JFzAmWJW5KcA

ckumar1:
Fixes #5959

## Cross-city context (why this exists)

Two cities sharing one federated rig store (a Dolt hub) each run this pool orphan sweeper against it. The sweeper validates a claim's assignee liveness against **local** evidence only — `liveOpenSessionAssignmentExists` answers from the local city's session stores — so a claim held by an identity the *other* city configures reads as dead, and `releaseOrphanedPoolAssignment` strips its assignee. A federated claim validated against unfederated evidence.

This is not theoretical: both cities did it to each other. One direction took two P0s and a P1 in 8 seconds; the reverse took fifteen claims in 34 seconds. The strips leave no comment trail; the Go paths do emit `bead.dead_assignee_reopened` (`city_runtime.go`, `session_reconciler.go`), but the event lands in the stripping city's own log — the losing city sees nothing, so each incident was still reconstructed by archaeology. (Corrected per partner-city review: an earlier revision of this body claimed "no event".)

## The fix

A roster-membership gate **ahead of** the liveness question: an assignee that parses as a well-formed `<rig>/<name>` not present in the local city's resolved roster is *unobservable*, not dead — skip it. Two commits:

1. **`fix(pool): treat a foreign <rig>/<name> claim as unobservable, not dead`** — the gate (`poolAssigneeIsLocallyObservable`), its wiring at the release site, and a once-per-sweep summary line naming every protected identity and claim count.
2. **`fix(pool-sweeper): refuse a binding this city does not mint`** — closes the hole where a foreign identity matching a local *template shape* would still pass; without it the gate is a cross-city hazard rather than a protection.

## Properties reviewers should hold us to

- **Locally-configured identities keep today's behavior byte-for-byte.** A local agent with a genuinely dead session is STILL released. `TestReleaseOrphanedPoolAssignments_ProtectsForeignIdentityAndStillReleasesLocal` carries both claims in ONE sweep — it is the non-vacuity proof that this fixes the bug rather than disabling the sweeper. The two pre-existing load-bearing tests (`ReopensRigStoreMissingPoolAssignee`, `ReleasesRigWorkAssignedToUnreachableOpenSession`) pass unchanged; the deliberately label-only liveness query was not touched.
- **The skip is counted and logged, never silent.** A decommissioned local agent is also "not in the roster", so a silent skip would trade a loud cross-city bug for a quiet local leak. The summary line (`protected N foreign/unknown identities this pass (M claims): ...`) renders identities with %q so untrusted assignee text cannot forge structure in it.
- **Calibrated against history, not just fixtures**: replayed against 2,873 real releases from the originating city's event log (227 distinct assignees), the gate changes exactly ONE historical decision — the known cross-city strip — and preserves 2,685 of 2,686 legitimate gated reclamations. Mutation-tested: 10 mutants (including "gate never fires", "gate on rig prefix", "silent skip"), 10 killed.

## Porting notes

Rebased onto current `main` (319e93d3f). One hand-resolved hunk in `pool_session_name.go`: the gate keeps its place after the two positive-evidence checks (they can only ever SKIP a release, so ordering keeps the protected count meaning "claims this gate saved") and now sits ahead of #6068's memoized probe (`memoizedLiveOpenSessionAssignmentExists`); the tie-break test calls were updated to #5935's six-argument `releaseConfirmedOrphanSessionWork` (`assignedWorkStores` supplied index-aligned). Verified at this head: `go vet ./cmd/gc` clean; `-run` over all eleven foreign-identity/release tests green. A CHANGELOG entry names the removed-binding migration case (a decommissioned agent's claims are no longer reclaimed — release explicitly when retiring a binding; watch the `protected N foreign/unknown identities` line).

## Scope, stated precisely (updated after partner-city review)

This PR covers the Go RELEASE paths: the sweep (`releaseOrphanedPoolAssignments`) and, as of the review-response commit, the orphan-close tie-break (`releaseConfirmedOrphanSessionWork`) — the latter's identifier set includes raw session-bead metadata, so it needed the same gate. The empty-assignee reopen arm has no identity to test and is documented as outside the gate (no cross-city claim-strip is possible there).

Deliberately NOT in this PR, each named and tracked:
- **The shell sweeper** — the core pack's `orphan-sweep.sh` makes the same decision in shell and stays unguarded after this merge: #5955.
- **The shared-store WRITE sites** — `canonicalizeLegacyBoundAssignedWork` / `...UnassignedRoutedWork` rewriting the other city's assignee/routed_to is #5846 (same incident family, different verb). Its fix applies this PR's predicate at those write sites; the partner city has that slice queued on top of this PR (see review thread), so it is deferred here rather than half-done.

Release note: a claim held by an identity absent from the local roster — including a locally **decommissioned** agent's — is no longer reclaimed by the Go release paths; each skip is counted and named in a once-per-sweep summary line, which is the deliberate mitigation for the decommissioned-agent leak (end-to-end test proves the line fires).

## Known limit (stated, not hidden)

Two cities importing the same agent pack share an identity space, so the gate protects foreign NAMED identities, not foreign POOL instances (byte-identical names). Both observed incidents were named crew, so the observed damage is covered; closing the full class needs a claiming-city discriminator on the claim itself and is deliberately out of scope here.

Second stated limit (per partner-city review): the template-form foreign-binding row (`repo/pool.worker`) is protected at the predicate but NOT end to end — `canonicalizeLegacyBoundAssignedWork` (`build_desired_state.go`) rewrites that stored assignee to the local spelling before the next sweep sees it, and the local spelling then releases. That write site is #5846's territory (its fix is this PR's `cityMintsBinding` applied there; the partner city has that slice queued on top of this PR), so the end-to-end protection for template-form rows lands with #5846, not here.

Review requested from the partner city's gc build owner (navani) per the cross-city agreement — review, not a gate; she is being pointed at this PR through the cities' own channel rather than a GitHub mention.

https://claude.ai/code/session_01QB58oPyfY5JFzAmWJW5KcA

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **The pool orphan sweep treats a claim it cannot observe as unobservable,
+  not dead.** On a shared store, an assignee in `<rig>/<name>` form that this
+  city's config does not mint — another city's identity, or a binding since
+  REMOVED from local config — used to be judged by local session evidence
+  alone and silently stripped (`in_progress` → `open`, assignee cleared).
+  Both release sites now refuse such claims and log one
+  `protected N foreign/unknown identities` summary per sweep. Migration note
+  for the removed-binding case: decommissioning an agent from config no longer
+  lets the sweep reclaim its outstanding claims — release them explicitly
+  (`gc bd release`/reassign) when retiring a binding, and watch the summary
+  line for claims held back.
+
 - **The work-record close gate asks the repository the bead's OWNER points at,
   not the store it was read through.** A rig's work step that a relocated class
   binding holds has its commits on the rig's checkout, and both close doors
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,7 +86,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `protected N foreign/unknown identities` summary per sweep. Migration note
   for the removed-binding case: decommissioning an agent from config no longer
   lets the sweep reclaim its outstanding claims — release them explicitly
-  (`gc bd release`/reassign) when retiring a binding, and watch the summary
+  (`gc bd release-if-current <id> <assignee>`/reassign) when retiring a
+  binding, and watch the summary
   line for claims held back.
 
 - **The work-record close gate asks the repository the bead's OWNER points at,
--- a/cmd/gc/pool_orphan_foreign_identity.go
+++ b/cmd/gc/pool_orphan_foreign_identity.go
@@ -1,0 +1,325 @@
+package main
+
+import (
+	"fmt"
+	"log"
+	"sort"
+	"strings"
+
+	"github.com/gastownhall/gascity/internal/config"
+)
+
+// The pool orphan sweeper validates a claim it can SEE against liveness
+// evidence it may not be able to see. A work bead in a SHARED rig store is
+// visible to every city attached to that rig, but the session bead proving the
+// claim is alive lives in the claiming city's own hq — unfederated. So when a
+// foreign city's agent holds a claim, this city asks its own hq for that
+// session, does not find it, and concludes the claim is dead. Both cities do
+// it to each other: theirs cleared two of our P0s and a P1 in an 8-second
+// batch; ours cleared fifteen of theirs in 34 seconds on 2026-08-21.
+//
+// The gate below separates "no session bead exists" from "no session bead is
+// OBSERVABLE from here":
+//
+//	foreign / unknown  -> unobservable -> PROTECTED (skip, do not release)
+//	locally configured -> observable   -> today's behaviour, UNCHANGED
+//
+// The unchanged-for-local property is what makes this shippable onto a box
+// running live reviews, so every helper here is a pure read of local config.
+//
+// TWO THINGS THIS GATE DELIBERATELY DOES NOT DO.
+//
+// It does not test the RIG PREFIX. qcore is precisely the SHARED rig, so a
+// prefix test protects nothing — the discriminator has to be the agent
+// identity against the local roster.
+//
+// It does not test pool CAPACITY. "<local agent>-<slot>" is an instance
+// identity of a local agent family whatever the current max_active_sessions
+// says; measured against this city's own release history, live assignees run up
+// to qcore/gc.run-operator-36 against a pool configured for 4. Gating on the
+// slot ceiling would make every claim above the ceiling permanently unreapable
+// the moment a pool is shrunk.
+//
+// KNOWN RESIDUAL (tracked separately): two cities importing the same agent pack
+// share an identity space, so a foreign POOL instance ("qcore/gastown.furiosa")
+// is byte-identical to ours and this gate cannot protect it. It protects
+// foreign NAMED identities, which is the shape both observed incidents took.
+// Closing the residual needs a claiming-city discriminator on the claim itself.
+
+// poolAssigneeIsLocallyObservable reports whether assignee names an identity
+// whose liveness this city is in a position to answer.
+//
+// Anything that is not a well-formed <rig>/<name> identity returns true: bare
+// aliases, session bead IDs and runtime session names ("gastown__dog-ga-up143",
+// "claude-mc-xyz") are this city's own naming and were never the cross-city
+// hazard. Keeping them on the existing path is what preserves current
+// behaviour for every local assignee shape.
+func poolAssigneeIsLocallyObservable(cfg *config.City, cityName, assignee string) bool {
+	assignee = strings.TrimSpace(assignee)
+	if cfg == nil || assignee == "" {
+		return true
+	}
+	rig, local := config.ParseQualifiedName(assignee)
+	if strings.TrimSpace(rig) == "" || strings.TrimSpace(local) == "" {
+		return true
+	}
+	return poolIdentityInLocalRoster(cfg, cityName, assignee)
+}
+
+// poolIdentityInLocalRoster resolves a <rig>/<name> identity against local
+// config: configured named sessions, configured agent templates (including the
+// legacy bound form that persisted assignees still carry), namepool-themed pool
+// instances, and the instance identities gc mints for a local agent (numeric
+// slots and adhoc tokens).
+func poolIdentityInLocalRoster(cfg *config.City, cityName, identity string) bool {
+	for _, candidate := range poolIdentityLocalCandidates(identity) {
+		if _, ok := findNamedSessionSpec(cfg, cityName, candidate); ok {
+			return true
+		}
+		if findAgentByTemplate(cfg, candidate) != nil {
+			return true
+		}
+		if config.FindAgent(cfg, candidate) != nil {
+			return true
+		}
+		if poolIdentityIsThemedInstance(cfg, candidate) {
+			return true
+		}
+		if poolIdentityIsInstanceOfLocalAgent(cfg, candidate) {
+			return true
+		}
+	}
+	return false
+}
+
+// poolIdentityLocalCandidates returns the identity itself plus, when its local
+// part carries a binding prefix, the binding-stripped form.
+//
+// Persisted assignees outlive the config that minted them: this city's own
+// release history is full of "qcore/gastown.polecat" and
+// "qcore/gc.run-operator-1" while the resolved config now holds those agents
+// UNBOUND as "qcore/polecat" and "qcore/run-operator". findAgentByTemplate
+// already carries that bound->unbound fallback for exact template lookups
+// (legacyBoundTemplateMatchesUnboundAgent); the instance forms below need the
+// same tolerance, so the stripping happens once, here, for every check.
+// Without it this gate reads 45 of this city's 46 historically-reaped pool
+// identities as foreign and protects them forever — a sweeper that has been
+// disabled rather than fixed.
+func poolIdentityLocalCandidates(identity string) []string {
+	identity = strings.TrimSpace(identity)
+	if identity == "" {
+		return nil
+	}
+	candidates := []string{identity}
+	rig, local := config.ParseQualifiedName(identity)
+	if binding, unbound, ok := strings.Cut(local, "."); ok && binding != "" && unbound != "" {
+		stripped := unbound
+		if rig != "" {
+			stripped = rig + "/" + unbound
+		}
+		candidates = append(candidates, stripped)
+	}
+	return candidates
+}
+
+// poolIdentityIsThemedInstance matches a namepool display name, e.g.
+// "qcore/furiosa" against agent qcore/polecat whose namepool lists "furiosa".
+//
+// A generic rig-scoped template (Dir empty, Scope "rig") applies to every
+// configured rig, so its instances are addressed under a concrete rig this
+// template never names. config.FindAgent already synthesizes the rig-bound copy
+// for exact lookups; the themed comparison has to do the same or a legitimate
+// local "qcore/furiosa" reads as foreign.
+func poolIdentityIsThemedInstance(cfg *config.City, identity string) bool {
+	if cfg == nil {
+		return false
+	}
+	rig, _ := config.ParseQualifiedName(identity)
+	rigConfigured := rig != "" && cityHasRigNamed(cfg, rig)
+	for i := range cfg.Agents {
+		agentCfg := cfg.Agents[i]
+		variants := []config.Agent{agentCfg}
+		if rigConfigured && strings.TrimSpace(agentCfg.Dir) == "" && agentCfg.Scope == "rig" {
+			rigBound := agentCfg
+			rigBound.Dir = rig
+			variants = append(variants, rigBound)
+		}
+		for _, themed := range agentCfg.NamepoolNames {
+			if themed = strings.TrimSpace(themed); themed == "" {
+				continue
+			}
+			for v := range variants {
+				if variants[v].QualifiedInstanceName(themed) == identity {
+					return true
+				}
+			}
+		}
+	}
+	return false
+}
+
+func cityHasRigNamed(cfg *config.City, name string) bool {
+	if cfg == nil {
+		return false
+	}
+	for i := range cfg.Rigs {
+		if cfg.Rigs[i].Name == name {
+			return true
+		}
+	}
+	return false
+}
+
+// poolIdentityIsInstanceOfLocalAgent matches an instance identity minted for a
+// locally-configured agent: "<local agent>-<slot>" or
+// "<local agent>-adhoc-<token>".
+//
+// Those are the two instance-name generators in the tree — poolInstanceName's
+// numeric slot form and session.GenerateAdhocIdentity, which cmd_hook.go writes
+// straight into the claim assignee for an aliasless pool worker
+// ("rig/polecat-adhoc-<hash>"). Both are recognised WITHOUT consulting the
+// pool's configured ceiling: see the capacity note at the top of this file.
+//
+// The grammar is closed on purpose rather than accepting any suffix, so this
+// stays a statement about identities gc actually mints. A THIRD generator added
+// later and not added here would have its claims protected — visibly, in the
+// per-sweep summary, rather than silently — which is the failure direction this
+// whole change is built around. Add new generators here.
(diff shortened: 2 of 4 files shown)


A3Ackerman:
Partner-city read (navani, the other city sharing this hub; review, not a gate). The three files are solid: the gate fails closed when the roster cannot be resolved (`pool_orphan_foreign_identity.go` returns `false` → protect on nil config / unknown binding), the non-vacuity test runs one real sweep over a real store and shows `repo/dalinar` protected while `repo/worker-1` is still released, canonical helpers are used (`config.ParseQualifiedName`, `config.FindAgent`, `QualifiedInstanceName`), and the test file adds no subprocess/tmux/httptest/sleep sites, so the resource census does not move. What follows is the sibling-path list a maintainer review would ask for, with evidence on current `main` (59d581a0f).

**Parallel code paths not covered and not named as out of scope**

1. `internal/bootstrap/packs/core/assets/scripts/orphan-sweep.sh` (`is_known_agent`, ~L320-362) is the same decision in shell: local config agents + `live_session_match`, then in_progress→open. Its own comment says it "mirrors liveOpenSessionAssignmentExists in the Go path", and `orders/orphan-sweep.toml` runs it every `5m`. Any city with the core pack keeps stripping foreign claims after this merges. Either give `is_known_agent` the same roster refusal (count it as `UNVERIFIABLE`), or say in the body that the shell sweeper is a follow-up.
2. `cmd/gc/build_desired_state.go` `canonicalizeLegacyBoundAssignedWork` (L5053, assignee write ~L5094-5099) and `canonicalizeLegacyBoundUnassignedRoutedWork` (L5131, `routed_to` write ~L5153-5158) rewrite another city's assignee/route on the shared store — that is #5846, whose fix candidate 1 is exactly the `cityMintsBinding` predicate this PR introduces. Applying it at those two sites is a few lines; if it is deliberately deferred, the body should link #5846 and say so.
3. `cmd/gc/pool_session_name.go:355` (`releaseConfirmedOrphanSessionWork`) reaches `releaseOrphanedPoolAssignment` with no roster gate. Named identities cannot get there, but a byte-identical foreign pool instance can — so the stated pool-instance residual is not merely unprotected, it is released through a second site. Worth naming.
4. The empty-assignee / in_progress reopen arm at `pool_session_name.go:~227-230` is outside the gate; this city still reopens another city's unassigned in-flight work.

**Body and tests**

- No `Fixes #N` — the linkage gate here expects an issue for the sweeper arm (the reconciler arm already has #5846; cross-link it).
- The decommissioned-local-agent case (`{"decommissioned local agent", "repo/retired-hand", false}`, test L693) is predicate-only. Since the PR's mitigation for that permanent leak is the once-per-sweep summary line, one end-to-end case proving the bead survives the sweep and the summary names it would close B14 for it, and the body wants a `Release note:` line for the behavior change operators will hit (a decommissioned agent's claims are no longer reclaimed; watch the `protected N foreign/unknown identities` line).
- Nit: `isPoolSlotNumber` + `cutLastDash` re-implement `routeHasSlotSuffixShape` (`cmd/gc/demand_serve_predicate.go:132-145`) and disagree with it on inputs like `007`; one predicate per package would be kinder to the next reader.

Happy to take item 2 as the follow-up slice on our side once this lands, reusing `cityMintsBinding` — say which you prefer.

ckumar1:
Thanks for the read — every item was real. Acted on it as follows (commit d17c4093c + body update):

**Fixed in-PR:**
- (3) `releaseConfirmedOrphanSessionWork` now runs the same roster gate. You were right that it mattered beyond the pool-instance residual: its identifier set includes **raw session-bead metadata** (`sessionAssignmentIdentifierRawInfo` — bead ID / session_name / configured identity), not roster-derived names, so a stored value colliding with your city's identity reached that release ungated. New non-vacuity test drives both halves through one call (unobservable protected, locally-minted instance still released); mutation-checked — deleting the gate fails it.
- (B14) End-to-end decommissioned-local sweep test: bead survives, summary names the identity — the accepted leak's mitigation proven to fire, not just the predicate row.
- (4) The empty-assignee arm is *named* rather than guarded: no identity to test, so it shares the stated residual with bare names; the claim-strip damage class can't occur there (nothing to strip). Comment at the arm + body.
- (nit) `isPoolSlotNumber` vs `routeHasSlotSuffixShape` on `007` is deliberate and now explained at the definition: fail-closed minted-identity claim vs fail-open route-shape pre-filter — the unminted shape is protected-and-reported here, not silently stripped to a base name. Unifying would force one predicate to answer the other's question.

**Named as follow-ups, not claimed:**
- (1) The shell sweeper: filed as #5955, cross-linked to #5846; the body now says plainly that pack-installed cities' `orphan-sweep.sh` stays unguarded after this merge.
- (2) The canonicalize write sites: **deferred to #5846 — please take it as your gc-mjr3 slice on top of this predicate.** One incident family, but a different verb (rewriting vs releasing), and your slice was already shaped for exactly `:5079/:5153`. Keeping this PR the reviewed size beats growing it mid-review. Body links #5846 and says deferred-with-owner.

No `Fixes #N` added deliberately: this PR fixes neither #5955 (shell path untouched) nor #5846 (write sites deferred); the incident it addresses has no upstream issue of its own — happy to file one and link it if the linkage gate requires a Fixes line to merge.

A3Ackerman (commented):
Reviewed d17c4093c against current main (7a0fdb10c): all three files, both release sites, the resolvers the gate leans on (config.FindAgent, findAgentByTemplate, session.FindNamedSessionSpec), both instance-name generators, and a dry-run of the pool_session_name.go hunks against main.

The change is correct and fails in the safe direction: it only ever removes releases, local pool/named/themed/adhoc shapes stay on the existing path, and the non-vacuity tests fail without the gate. Three things before it can merge. First, it conflicts: hunk 3 lands on the probe line #6068 rewrote to memoizedLiveOpenSessionAssignmentExists, and after a textual rebase the new tie-break test still will not compile, because #5935 added an assignedWorkStores parameter to releaseConfirmedOrphanSessionWork (test lines 526 and 544 pass five arguments). Second, no required suite has run: every pull_request workflow on this head is action_required (fork awaiting approval), which is what the evidence watchdog's "tests never ran" is reporting. Third, two body corrections: the Go paths do emit bead.dead_assignee_reopened (city_runtime.go:2546, session_reconciler.go:2397), so "no event" is wrong; and the template-form foreign-binding row (repo/pool.worker) is protected at the predicate but not end to end, because canonicalizeLegacyBoundAssignedWork (build_desired_state.go:818, body at :5158) rewrites that stored assignee to the local spelling and the next tick's sweep releases it. That is #5846, but say so here, or apply cityMintsBinding there now. A CHANGELOG line naming the removed-binding migration case would help too.

Approve once rebased and green. Disclosure: I am a contributor from the affiliated quickserve-ai organization, the partner city on this incident.

A3Ackerman (changes requested):
**Re-read at 488e858f0 — REQUEST CHANGES, one mechanical push from approval.** All three items from the previous round are addressed. What remains is that `preflight-static`, which feeds `CI / required`, fails on the two new files as they stand: one gofumpt formatting diff and twelve `misspell` (locale US) hits. Fix, push, and I approve; the fork's workflow runs still need a maintainer's approval to execute.

**Verified**

1. Rebase. Head 488e858f0 (a9845ffdf, ef7ab9df5, 488e858f0), base 319e93d3f = merge-base with `main`; `main` has since moved to 36763a969 and `git merge-tree --write-tree` against it merges clean.
2. Item 1 (conflict + compile), addressed. `cmd/gc/pool_session_name.go`: the sweep gate `poolAssigneeIsLocallyObservable` (:327-330) sits after `openSessionOwnsWork` (:314) and `assigneePreservesNamedSessionRoute` (:317) and ahead of #6068's `memoizedLiveOpenSessionAssignmentExists` (:331); accumulator at :266, `defer protectedForeign.log()` at :267. Tie-break gate at :463-466, before template/owner-store/release (:467-492). `releaseConfirmedOrphanSessionWork` is #5935's six-argument form (:419-426); all seven call sites pass six arguments (`session_reconciler.go:2396`, `pool_session_release_binding_owner_test.go:79/106/132/160`, the new test at `pool_orphan_foreign_identity_test.go:526` and `:545`, both `[]beads.Bead{foreign, local}` against `[]beads.Store{rigStore, rigStore}`, index-aligned). `go vet ./cmd/gc/` clean (go1.26.6). `go test ./cmd/gc/ -count=1` over the foreign-identity and orphan-release family: 108 top-level tests pass, 0 fail, including all eleven new tests and `TestReleaseOrphanedPoolAssignments_ReopensRigStoreMissingPoolAssignee` / `..._ReleasesRigWorkAssignedToUnreachableOpenSession`.
3. Item 2 (body), addressed. The body now says the Go paths emit `bead.dead_assignee_reopened` into the stripping city's log (emit sites at this head: `city_runtime.go:2551`, `session_reconciler.go:2397`, via `dead_assignee_event.go:25`); "Second stated limit" names `canonicalizeLegacyBoundAssignedWork` (`build_desired_state.go:818` call, `:5158` body) and defers it to #5846 with an owner; `Fixes #5959` added. CHANGELOG entry at `CHANGELOG.md:80-91` under `[Unreleased] › Fixed`, naming the removed-binding migration case.
4. Item 3 (CI), still gated as stated. All six `pull_request` workflows are `action_required` at this head; the `PR evidence watchdog` red says `tests never ran`, which is the approval gate itself, not a test.
5. Local sweep. `make test-cmd-gc-process-parallel` at the head: shards 2, 3 and `productmetrics-testhook` ok; five reds on shards 1, 4, 5, 6, none touching a PR file or symbol, each reproducing on the base commit or passing in isolation (`TestPhase0DoctorReportsClosedBeadOwner` 3/3 in isolation; `TestEmitSessionStrandedDiagnostic_DetachedProbeAliveSuppressesEvent` 3/3 in isolation; `TestScopedBdStoreForCityKillsChildOnCtxCancel` and `TestLoadStatusSessionSnapshotKillsBdChildOnTimeout` fail 6/6 identically on 319e93d3f without the PR; `TestDoltStatePreflightCleanCmdRemovesSocketsButPreservesDoltInternals` 1/3 at the head, 2/3 on the base).
6. Lint. `golangci-lint run --new-from-rev=319e93d3f ./cmd/gc/` with the repo config: 13 issues, all on the new files (A below).

**Findings**

A. Blocking — formatter and spelling lints that `CI / required` enforces (`preflight-static` runs `make fmt-check-changed` / `fmt-check` at ci.yml:253/256 and `make lint-affected` / `lint` at :242/:247 → `ci-preflight` ci.yml:1539 → `CI / required` ci.yml:1616; the selector takes every changed regular `.go` file, tests included, scripts/ci-static-select:391).
   - `cmd/gc/pool_orphan_foreign_identity_test.go:350`: gofumpt — the first row of the `TestCityMintsBinding` agents literal carries alignment padding gofmt removes (`{Name: "worker"}, // unbound: contributes nothing`). Reproduced with `gofmt -l` and `golangci-lint fmt --diff`.
   - `misspell` (locale US, .golangci.yml:8/:34): `behaviour` at `pool_orphan_foreign_identity.go:25, :56`; `neighbouring` at `:140, :184` and `_test.go:345, :366, :367` (`NEIGHBOURING` at `_test.go:248`); `recognised` / `recognises` at `:277, :302, :365`; `Unrecognised` at `_test.go:233`. The existing `cmd/gc` tree has none of these spellings.
   - Fix: `golangci-lint fmt cmd/gc/pool_orphan_foreign_identity_test.go && golangci-lint run --fix ./cmd/gc/` (misspell autofixes), commit.
B. `CHANGELOG.md:89` names `gc bd release`, which does not exist (`gc bd` is a passthrough, cmd_bd.go:70, whose only gc-owned verb is `release-if-current <issue-id> <assignee>`, cmd_bd.go:106/119/601, docs/reference/cli.md:298/319). Suggested: "(`gc bd release-if-current <id> <assignee>`, or reassign)". Same push as A.
C. `CHANGELOG.md:84-85` says both release sites log one `protected N foreign/unknown identities` summary per sweep; the tie-break site logs one line per refused claim in a different form (`pool_session_name.go:464`, pinned by `_test.go:541`). Suggested: "the sweep logs one `protected N …` summary per pass; the orphan-close tie-break logs each refusal by identity and bead". Non-blocking.
D. Scope statement, non-blocking. The reason the review-response commit gives for gating the tie-break (its identifiers are raw session-bead strings, `session_beads.go:962-968`) also describes the retire/close/drain release paths in `cmd/gc/session_beads.go` (`:1108-1137`, the Info form `:1370-1395`, `releaseUnexecutedClaimsOnDrainAck` `:1186-1224`, `releaseWorkFromClosedSessionBead` `:3504-3543`), all matching work by identifier equality and releasing via `ReleaseWorkBead` (`work_assignment.go:179`, clear at `:256-265`); the first three walk rig-store legs. The trigger there is a local session's own retirement and the collision precondition is the one the tie-break gate already covers, so no code ask — but "This PR covers the Go RELEASE paths" over-claims; "the two liveness-driven release paths" is accurate, with these listed alongside #5846 as the same raw-identifier shape left for later.
E. Optional: `pool_session_name.go:463` uses `config.EffectiveCityName(cfg, "")` where `:268` uses `cfg.EffectiveCityName()`; identical (internal/config/named_sessions.go:96-113).
F. The complexity workflow is report-only (complexity.yml:31-54, every step `continue-on-error`); gocyclo: the new file tops at 11, the two touched functions +1 each (38, 21). No action.

Disclosure: contributor from the affiliated quickserve-ai organization, the partner city on this incident.

A3Ackerman (approved):
**APPROVE at 211629e05.** The three blocking items from my 488e858f0 review are closed in one mechanical commit, verified here in a fresh worktree at this head: `gofmt -l` lists neither new file; no `behaviour` / `neighbouring` / `recognis` / `Unrecognis` remains in `cmd/gc/pool_orphan_foreign_identity.go` or its test; `CHANGELOG.md:89` now names `gc bd release-if-current <id> <assignee>`; `golangci-lint run --new-from-rev=<merge-base> ./cmd/gc/` reports 0 issues (13 before); `go vet ./cmd/gc/` clean; the foreign-identity and `TestCityMintsBinding` test families pass with `-count=1`. The non-blocking C/D/E from the earlier review (the tie-break logs per refusal, the scope sentence over-claiming the Go release paths, the `EffectiveCityName` spelling) are fine as a fast-follow; nothing in them changes the fix. Workflow runs still need a maintainer's approve-and-run at this head.

Disclosure: contributor from the affiliated quickserve-ai organization, the partner city on this incident.

PR #5961, opened 2026-09-03
test(resilience): injectable jitter seam for deterministic breaker backoff

bourgois:

## Why this exists

`Breaker` already carries an injectable `jitter` field, but `Registry` — the
only public way to obtain a breaker — has no way to reach it. Every consumer
of a registry-owned breaker therefore gets a uniform draw from `(0, cap]` for
its open-state deadline, which makes any test that asserts on a backoff
deadline time-dependent. `Registry.SetJitterForTest` pins the jitter function
on the registry and on every breaker it has already handed out, exactly the
way `SetOnStateChange` already reaches both.

## What it does not do

Nothing about the default. An unpinned registry keeps full jitter; a test
guards that, because silently losing full jitter would remove the
thundering-herd protection the backoff exists for. No production caller is
added — the consumers that motivate this seam (the bd chokepoint breaker and
its timeout classification) are separate PRs in the same split.

## Stacking

None. This applies to current `origin/main` on its own.

## Extraction fidelity

`internal/resilience/registry.go` is taken verbatim from the fork:
`git diff fork/main -- internal/resilience/registry.go` is empty.

`internal/resilience/registry_jitter_test.go` is new and written for this PR —
the fork has no test for this seam because its only callers are `cmd/gc` tests
that live in later slices. Flagging that explicitly so it is not mistaken for
extracted code.

## Verification

```
gofmt -l internal/resilience/                      clean
go vet ./internal/resilience/                      clean
golangci-lint run ./internal/resilience/...        0 issues
go test ./internal/resilience/ -count=1 -v         31 PASS, 0 FAIL, 0 SKIP (0s)
go test ./internal/resilience/ -race -count=1      ok (1.0s)
```

Both new guards were checked by breaking what they guard:

- deleting `b.jitter = r.jitter` from `Registry.Breaker` →
  `TestRegistrySetJitterForTestPinsNewBreakers` fails
  (`open backoff = 254.137761ms, want exactly 1s`)
- deleting the existing-breaker loop from `SetJitterForTest` →
  `TestRegistrySetJitterForTestPinsExistingBreakers` fails
  (`open backoff = 1.794389616s, want exactly 2s`)

### DIFF
--- a/internal/resilience/registry.go
+++ b/internal/resilience/registry.go
@@ -1,6 +1,9 @@
 package resilience
 
-import "sync"
+import (
+	"sync"
+	"time"
+)
 
 // OpClassBd is the operation class for bd CLI transport operations
 // (subprocess invocations against the managed Dolt backend). All bd
@@ -22,6 +25,7 @@ type Registry struct {
 	mu       sync.Mutex
 	settings Settings
 	onChange func(Transition)
+	jitter   func(capacity time.Duration) time.Duration
 	breakers map[Key]*Breaker
 }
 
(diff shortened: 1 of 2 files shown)


bourgois:
Split of the closed #3324 — wave two. Siblings so far: #5945 #5946 #5947 #5948 #5949, plus #5962 and #5963.

PR #5962, opened 2026-09-03
feat(beads): store availability gate + degraded on the status wire

bourgois:

## Why this exists

`CachingStore` can already be degraded — `cacheDegraded` is an existing state
that repeated reconcile failures latch — but nothing surfaces it, so an
operator reading `gc status` cannot tell a healthy cache from one serving
stale data. This adds `CachingStore.Degraded()` and puts the answer on the
status wire as `beads.degraded`, computed at **read** time from the live store
(a value captured at store-open time would report `false` forever).

The same PR introduces the `AvailabilityGate` seam: a small injectable,
swappable reporter of backing-store transport reachability
(`Available()` / `ProbeDue()` — the shape `resilience.Breaker` already has).
`Degraded()` answers from either the gate or the cache's own state.
`internal/beads` must not import `internal/resilience`, so the seam is an
interface rather than a concrete dependency.

## What it does not do

- No production gate is wired. On this branch `availabilityGate` is nil for
  every store, so `beads.degraded` reflects the cache state alone and no read
  path changes behaviour at all. The transport circuit breaker that fills the
  seam is a later PR in the split.
- No read is gated. `SetAvailabilityGate`'s doc deliberately describes only
  what this PR does; the gated read path is the stacked follow-up.

## Stacking

None. This applies to current `origin/main` on its own. The degraded-read PR
stacks on top of it.

## Extraction fidelity

Hunk-level extraction from `fork/main`. The following are byte-identical to
the fork (bodies and doc comments, verified by comparison):

- the `AvailabilityGate` interface
- `availabilityGateRef()`
- `Degraded()`
- `storeReportsDegraded()` and the `diag.Degraded = ...` line in
  `cmd/gc/api_state.go`
- the `Degraded` field and comment on `beads.BeadsDiagnostic`

Deliberate deviations, all disclosed:

1. `SetAvailabilityGate`'s **doc comment** is rewritten to describe this PR's
   behaviour. The fork's wording describes the gated read path, which is not
   here yet; the stacked PR restores the fork's exact text. The function body
   is byte-identical.
2. The `availabilityGate` **field comment** on `CachingStore` drops one stale
   sentence. The fork's comment opens with "Guarded by mu." and then, three
   lines later, explains at length that the field is deliberately atomic and
   **not** mu-guarded. The first sentence is a leftover from an earlier
   revision and contradicts the field's declared type; reporting rather than
   silently carrying it.
3. `cmd/gc/api_state.go` on the fork bundles `wireStoreAvailabilityGate`,
   `storeReportsDegraded` and an unrelated `ResolveDoltPort` rewrite that
   belongs to a separate port-file change. Only the `storeReportsDegraded`
   /`diag.Degraded` parts are taken here. `wireStoreAvailabilityGate` needs
   the breaker registry from a later PR in this split, so it lands with the
   breaker rather than ahead of it.

`internal/api/openapi.json`, `docs/reference/schema/openapi.*`,
`internal/api/genclient/client_gen.go`, the generated TS client and
`internal/api/dashboardspa/dist/` are all **generator output**, produced by
`go run ./cmd/genspec`, `go generate ./internal/api/genclient` and
`make dashboard-ci`. They are not hand-edited.

The three test files are new, written for this PR.

## Verification

```
gofmt -l cmd internal                              clean
go build ./...                                     ok
go vet ./cmd/gc/ ./internal/api/ ./internal/beads/ clean
golangci-lint run ./internal/beads/... ./cmd/gc/... 0 issues
make spec-ci                                       regenerated, no drift beyond this change
make dashboard-ci                                  exit 0 (21s)
go test ./internal/beads/ -count=1                 ok (13.9s)
go test ./internal/api/   -count=1                 ok (61.8s)
GC_FAST_UNIT=0 go test ./cmd/gc/ -run 'TestCityBeadsDiagnostic|TestStoreReportsDegraded' -v
                                                   5 PASS, 0 SKIP (1.4s)
make test-fast-parallel                            unit-core ok; cmd/gc shards: see below
```

`internal/beads`'s `TestMergeOracleFieldCoverage` field census required
classifying the new `availabilityGate` field; it is excluded with a reason
(reachability, not durable cache content, never written by the merge seam).

Both guards were checked by breaking what they guard:

- deleting `diag.Degraded = storeReportsDegraded(cs.cityBeadStore)` →
  `TestCityBeadsDiagnosticReportsDegradedThroughThePolicyWrapper/unreachable`
  fails (`Degraded = false, want true`)
- deleting the decode wiring in `internal/api/decode_status.go` →
  `TestStatusBeadsDiagnosticDecodesDegraded/present` fails
  (`Degraded = false, want true`)

The `cmd/gc` behaviour is tested through `controllerState.CityBeadsDiagnostic()`
— the method status actually calls — with the store wrapped exactly as
production wraps it (`wrapStoreWithBeadPolicies`), not by calling the helper
directly.

### Known-flaky `cmd/gc` shards

`make test-fast-parallel` (16-way) failed a handful of `cmd/gc` tests on this
machine. They are load-sensitive, not caused by this change:

- two independent runs produced **different** failure sets (9 tests, then 7,
  overlapping in only two)
- every one of the 16 passes in isolation under `GC_FAST_UNIT=0`, each in
  0.01–4.6s against 10–15s timeouts
- all are managed-Dolt / order-exec / process-group tests; this PR's diff
  cannot reach them (the gate is nil, so no read path changes)

`unit-core` was clean in both runs.


bourgois:
Split of the closed #3324 — wave two. Siblings so far: #5945 #5946 #5947 #5948 #5949 #5961, and #5963 stacks on this one.

PR #5963, opened 2026-09-03
feat(beads): serve last-good reads while the store is unavailable

bourgois:

**Stacks on #5962 (the availability gate). Please take that one first** — this
branch contains #5962's commit; the diff to review here is the second commit.

## Why this exists

When the availability gate reports the backing store unreachable,
`CachingStore` stops dialing it: `List`, `Get` and `Count` answer from the
in-memory snapshot, the reconciler skips cycles until a recovery probe is due,
and `Stats().DegradedReads` counts what was served that way. A backend outage
degrades reads to stale-but-correct instead of amplifying into a read outage.

The honesty boundary is the point of the change. Shapes the snapshot cannot
answer truthfully refuse with the new typed `beads.ErrStoreUnavailable`
rather than returning an empty set that would read as "none exist":

- an unprimed cache — unavailable must never read as empty
- a `Live` query — its contract is "observe external mutations immediately",
  and a lifecycle gate that treats absence as authoritative would release a
  running session's pool assignment on a stale short list
- a closed-only listing — the snapshot holds active beads only
- an uncached `Get` — the cache cannot prove absence

A partially primed snapshot is tagged with the package's existing
`PartialResultError` on every answer. A `Count` carries no partial tag, so
shapes that would need one report "cannot answer" and the caller surfaces the
failure instead; a backing with no `Counter` still gets `ErrCountUnsupported`,
not `ErrStoreUnavailable`, so the existing List fallback keeps working.

## What it does not do

- **No last-good fallback when a dialed store fails to answer.** This path
  triggers only when the gate already says the transport is down. The
  reality-first backstop (dial, and fall back on a non-answer) is a separate
  slice.
- **No production gate.** Nothing wires a gate on this branch either, so every
  read takes its current path unchanged and `servingDegraded()` is always
  false in production until the breaker PR lands. The behaviour here is
  exercised by tests through the real `List`/`Get`/`Count`/`runReconciliation`
  entry points with an injected gate.

## Two things to know while reviewing

1. **`beads.ErrStoreUnavailable` is also carried by #5949** (bd admission
   semaphore) from this same split. Identical text, identical location in
   `internal/beads/beads.go`. That is a deliberate duplicate so each PR builds
   standalone, not a real conflict — whichever lands second drops its copy.
2. **`listLastGood` is extracted without the fork's second parameter**
   (`tagIncompleteHistory bool`). Its only `true` caller is the dial-failure
   backstop listed above, which is not in this PR, and `unparam` (enabled in
   `.golangci.yml`) rejects a parameter that only ever receives `false`. The
   slice that adds the backstop restores the parameter and its
   `IncludeClosed` tagging branch.

## Extraction fidelity

Hunk-level extraction from `fork/main`. Byte-identical to the fork (bodies and
doc comments, verified by comparison):

- `servingDegraded()`, `lastGoodCount()`, `getLastGood()`,
  `reconcileUnavailableSkip()`
- `SetAvailabilityGate()`'s doc comment, restored to the fork's full text now
  that the behaviour it describes exists
- `ErrStoreUnavailable` and its comment
- the `Get` gate hunk in `caching_store_reads.go`
- both `caching_store_reconcile.go` hunks

Deliberate deviations, all disclosed:

1. `listLastGood` — the dropped parameter and its `IncludeClosed` branch,
   above. Everything else in the function is byte-identical to the fork.
2. The `List` and `Count` gate hunks. The fork writes them inside a `ListCtx`
   method and a `backingListCtx` helper that restore an unrelated `CtxLister`
   capability; only the gate branches are taken, against upstream's current
   `List`/`Count` signatures.
3. One comment in `Count` is reworded. The fork's says "`servingDegraded()`
   reads gate state under `c.mu`", which no longer matches the code —
   `servingDegraded` reads an atomic. The reason the `ctx` check must come
   first is still real (the degraded path itself takes `c.mu` via
   `lastGoodCount`), so the comment says that instead.
4. `caching_store.go`, `caching_store_reads.go`, `caching_store_reconcile.go`
   and `caching_store_events.go` on the fork also carry an unrelated ADR-0094
   `readyProjectionInvalid` feature. None of it is taken here.

The test file is adapted from the fork's `caching_store_unavailable` suite,
plus new `Count` cases the fork covers elsewhere.

## Verification

```
gofmt -l internal cmd                        clean
go build ./...                               ok
go vet ./...                                 clean
golangci-lint run ./internal/beads/...       0 issues
go test ./internal/beads/ -count=1           ok (23.5s)
go test ./internal/beads/ -run <gate suite> -v   34 PASS, 0 FAIL, 0 SKIP
make test-fast-parallel                      unit-core ok; cmd/gc shards: see below
```

`TestMergeOracleFieldCoverage`'s field census required classifying
`unavailableSkipLogged`, `degradedReads` and `CacheStats.DegradedReads`; all
three are excluded with a reason (outage observables, never written by the
merge seam).

Every guard was checked by breaking what it guards:

| mutation | result |
| --- | --- |
| drop the `List` gate branch | `...ListUnavailableServesLastGoodCache` and `...LiveQueryRefusesStaleAnswer` fail |
| drop the `Get` gate branch | `...GetUnavailableUncachedReturnsTypedError` fails |
| drop the `runReconciliation` skip | `...ReconcileSkipsCycleWhileUnavailable` and `...SkipLogsProblemOncePerEpisode` fail |
| make the probe-due path skip too | `...ReconcileRunsWhenProbeDue` fails (the breaker could never recover) |

### Known-flaky `cmd/gc` shards

Same as #5962: `make test-fast-parallel` failed a handful of managed-Dolt /
order-exec / process-group `cmd/gc` tests on this machine. Two runs produced
different failure sets, and all 16 pass in isolation under `GC_FAST_UNIT=0` in
0.01–4.6s against 10–15s timeouts. This PR touches only `internal/beads`, and
with no gate wired every read path is unchanged. `unit-core` was clean.


bourgois:
Split of the closed #3324 — wave two. Stacks on #5962. Siblings so far: #5945 #5946 #5947 #5948 #5949 #5961.

PR #5972, opened 2026-09-03
fix(worktree): gc worktree ensure now materializes skills when --agent is given (#5951)

### LINKED ISSUE #5951: bug: gc worktree ensure never materializes skill sinks into the worktree it creates
## Summary

`gc worktree ensure` never materializes provider skill sinks (`.claude/skills`,
`.agents/skills`) into the worktree it creates, so a session that later `cd`s
into that worktree loses every pack/role skill even though the owning agent's
scope root has them.

## Symptom

An agent session whose formula moves work into a per-bead worktree (e.g.
`mol-verified-work` / similar `prepare-worktree` steps) reports zero or a
stale skill list once its cwd changes to the worktree. A Codex seat run with
`CODEX_HOME` pointed at the worktree, or a Claude seat started with cwd in the
worktree, sees none of the pack skills (`core.*`, `atomize`, etc.) that are
present and correct at the scope root — only built-ins or whatever a stale
hand-mirrored sink happens to contain.

## Reproduction

1. In a city with an agent that has pack skills materialized at its scope
   root (verify with `ls <scope-root>/.claude/skills` or
   `.agents/skills`).
2. `gc worktree ensure --path <somewhere> --agent <agent-name>` (or whatever
   formula step creates a worktree today, without `--agent`).
3. Start a provider CLI with cwd/`$CODEX_HOME`/scope pointed at the new
   worktree and list its available skills.

Expected: the worktree carries the same skill sink contents as the scope
root. Actual: the worktree has no `.claude/skills` / `.agents/skills` at all
— the CLI falls back to built-ins or a stale user-level mirror.

## Root cause

`gc internal materialize-skills` is injected as a session `PreStart` entry
only "for stage-2-eligible runtimes whose session `WorkDir` differs from the
agent's scope root" (`internal/materialize/skills.go:11-12`, cross-referenced
in `cmd/gc/cmd_internal_materialize_skills.go:34-36`). That hook fires once,
at session start, for the session's *own* `WorkDir`. It has no visibility
into worktrees a formula step creates *after* the session has already
started — `gc worktree ensure` (`cmd/gc/cmd_worktree.go`, `runWorktreeEnsure`)
returns a plain worktree report today and performs no skill materialization
of any kind, regardless of `--agent`. A running session that then `cd`s a
sub-task into that new worktree is left with whatever sink (if any) happens
to already exist there.

## Fix candidates

**A. `--agent` flag on `gc worktree ensure`** — after a successful,
non-dry-run ensure, best-effort call the same materializer
(`materializeSkillsIntoWorkdir`, reused from the PreStart path) against the
new worktree path for the given agent identity. A materialization failure is
logged but never fails the ensure — the worktree is the postcondition this
command guarantees, not the skill sink.
- Scope: ~40 LOC in `cmd/gc/cmd_worktree.go` + tests, no changes to callers
  that don't pass `--agent`.
- Trade-off: opt-in, so every formula that creates worktrees today (e.g.
  `mol-verified-work`, `fable-nomad-code` `prepare-worktree` steps) must be
  updated to pass `--agent $GC_TEMPLATE` to get the fix. Low blast radius,
  but the burden is on each formula author to remember the flag.

**B. Materialize unconditionally inside `worktree.Ensure` / the reconciler's
worktree-creation path** — treat "has a skill sink" as an implicit
postcondition of any worktree gc itself creates, keyed off the calling
session's own agent identity (available via env/session context) rather than
an explicit flag.
- Scope: larger — touches the shared `internal/worktree` package and
  whatever plumbs session identity down to it; needs to distinguish
  "gc-owned worktree for this agent" from arbitrary git worktrees a user
  might `gc worktree ensure` for unrelated reasons.
- Trade-off: no formula changes required (fixes existing formulas
  automatically), but couples a general-purpose worktree primitive to
  skill-materialization concerns and requires plumbing agent identity
  through a path that doesn't carry it today.

**Recommendation:** A. It's the minimal, already-implemented fix (opt-in flag
+ reused materializer function), keeps `internal/worktree` free of
skill-sink concerns, and the two formulas that actually need it can be
updated in the same PR. B is worth revisiting later if more worktree-creation
call sites accumulate the same gap. A companion `gc doctor` check
(`internal/doctor`) that flags a session whose cwd carries no skill sink
closes the silent-failure gap for any call site that forgets the flag.

## Adjacent / out of scope

- A local hot-patch (`orders/skills-sync.toml` equivalent) that periodically
  mirrors vendor sinks into user-level sink directories
  (`~/.codex/skills`, `~/.claude/skills`) works around this today for
  interactive/ad-hoc sessions; it's a belt-and-braces workaround, not a fix,
  and out of scope here.
- Whether `gc doctor` should also detect a *session's own* `WorkDir` missing
  a sink (as opposed to a worktree it spawned) is a separate, already-covered
  case — the existing PreStart hook handles that path.

rjgeng:
## What

Closes #5951. Worktrees created directly via `gc worktree ensure` never got skill sinks materialized — the materializer (`internal/materialize.Run`) and its CLI wrapper (`gc internal materialize-skills`) already exist and are reused by two other call sites (the scope-root supervisor tick, and the per-session `PreStart` hook), but `gc worktree ensure` never invoked either one.

## Fix

New `--agent <name>` flag on `gc worktree ensure`; when given, triggers the same resolve-city/resolve-agent/`materialize.Run` sequence the `PreStart` hook already drives, scoped to the new worktree's path. Best-effort by design: a materialization failure is logged to stderr and never fails the `ensure` command itself, since worktree existence is its actual postcondition. Omitting `--agent` (every existing caller) leaves behavior completely unchanged — confirmed by a dedicated regression test.

## Scope note

The issue also names two formulas (`mol-verified-work`, `fable-nomad-code`) whose `prepare-worktree` steps should pass `--agent`. Neither formula's TOML exists in this repo — those pack definitions live in `gastownhall/gascity-packs`, a separate repo. This PR adds the capability those formulas would need to consume; wiring it into the actual formula steps is a separate, cross-repo follow-up.

## Validation

- `go test -tags gms_pure_go ./cmd/gc/... -run 'TestWorktreeEnsure|TestCmdWorktree'` — all 10 tests pass, including 2 new ones
- Full local push-gate run: all failures across every shard and `unit-core` confirmed to be pre-existing environmental/timing flakes (dolt-compaction timeouts, detached-probe/systemctl-delegate/lsof tests, interrupt-timing tests) unrelated to this change — none touch `cmd/gc/cmd_worktree.go`

### DIFF
--- a/cmd/gc/cmd_worktree.go
+++ b/cmd/gc/cmd_worktree.go
@@ -25,6 +25,7 @@ type worktreeCmdOpts struct {
 	Generation string
 	Lifecycle  string
 	AttemptID  string
+	Agent      string
 	DryRun     bool
 	JSON       bool
 }
@@ -90,6 +91,8 @@ func newWorktreeEnsureCmd(stdout, stderr io.Writer) *cobra.Command {
 	}
 	worktreeFlagSet(cmd, &opts)
 	cmd.Flags().BoolVarP(&opts.DryRun, "dry-run", "n", false, "plan without mutating anything")
+	cmd.Flags().StringVar(&opts.Agent, "agent", "",
+		"qualified agent identity to best-effort materialize skills for after creation")
 	return cmd
 }
 
@@ -182,7 +185,38 @@ func runWorktreeEnsure(opts worktreeCmdOpts, stdout, stderr io.Writer) int {
 		fmt.Fprintf(stderr, "gc worktree ensure: %v\n", err) //nolint:errcheck
 		return 1
 	}
+	if !opts.DryRun && opts.Agent != "" {
+		materializeWorktreeAgentSkills(opts.Agent, rep.Path, stdout, stderr)
+	}
 	return writeWorktreeReport("ensure", rep, opts, stdout, stderr)
+}
+
+// materializeWorktreeAgentSkills best-effort materializes agent's skills
+// into path, reusing the same resolve-city/resolve-agent/materialize.Run
+// sequence the per-session PreStart hook drives via
+// `gc internal materialize-skills` (see materializeSkillsIntoWorkdir in
+// cmd_internal_materialize_skills.go). A failure at any step is logged to
+// stderr and never returned: worktree existence is ensure's postcondition,
+// not a materialized skill catalog.
+func materializeWorktreeAgentSkills(agentName, path string, stdout, stderr io.Writer) {
+	cityPath, err := resolveCity()
+	if err != nil {
+		fmt.Fprintf(stderr, "gc worktree ensure: skill materialization skipped: city not found: %v\n", err) //nolint:errcheck
+		return
+	}
+	cfg, err := loadCityConfig(cityPath, stderr)
+	if err != nil {
+		fmt.Fprintf(stderr, "gc worktree ensure: skill materialization skipped: city config unavailable: %v\n", err) //nolint:errcheck
+		return
+	}
+	agent, ok := resolveAgentIdentity(cfg, agentName, currentRigContext(cfg))
+	if !ok {
+		fmt.Fprintf(stderr, "gc worktree ensure: skill materialization skipped: unknown agent %q\n", agentName) //nolint:errcheck
+		return
+	}
+	if err := materializeSkillsIntoWorkdir(cfg, &agent, cityPath, path, nil, stdout, stderr); err != nil {
+		fmt.Fprintf(stderr, "gc worktree ensure: skill materialization failed: %v\n", err) //nolint:errcheck
+	}
 }
 
 func runWorktreeVerify(opts worktreeCmdOpts, stdout, stderr io.Writer) int {
--- a/cmd/gc/cmd_worktree_test.go
+++ b/cmd/gc/cmd_worktree_test.go
@@ -74,6 +74,54 @@ func TestCmdWorktreeEnsureCreatesAndVerifies(t *testing.T) {
 	}
 }
 
+// clearCityEnv forces resolveCity to fail deterministically via the
+// ambient-upward-discovery refusal test binaries hit (main.go's
+// isTestBinary branch), regardless of what the host process inherited.
+func clearCityEnv(t *testing.T) {
+	t.Helper()
+	for _, k := range []string{"GC_CITY", "GC_CITY_PATH", "GC_CITY_ROOT", "GC_DIR", "GC_RIG"} {
+		t.Setenv(k, "")
+	}
+}
+
+func TestCmdWorktreeEnsureAgentFlagAttemptsMaterialization(t *testing.T) {
+	clearCityEnv(t)
+	repo, base := worktreeTestRepo(t)
+	root := t.TempDir()
+	wt := filepath.Join(root, "wt")
+	var stdout, stderr bytes.Buffer
+
+	opts := managedWorktreeCmdOpts(repo, root, wt, base)
+	opts.Agent = "gc-test-agent"
+	code := runWorktreeEnsure(opts, &stdout, &stderr)
+	if code != 0 {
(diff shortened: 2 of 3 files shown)


jacobhausler:
I filed #5951 and had reached the same shape independently (same flag, same
best-effort contract) — happy to see it in flight here rather than duplicate it,
so I've dropped my worktree branch and sent only the companion doctor check as
#5976. Two things I ran into on the way that this PR looks exposed to:

**1. `--agent` breaks `--json` on stdout.**

`materializeSkillsIntoWorkdir` writes human-readable progress to the `stdout`
writer it is handed — `cmd/gc/cmd_internal_materialize_skills.go`:

```go
fmt.Fprintf(stdout, "gc internal materialize-skills: provider %q has no skill sink in v0.15.1; skipping\n", provider)
fmt.Fprintf(stdout, "materialized %d skill(s) into %s: %s\n", ...)
fmt.Fprintf(stdout, "legacy stubs migrated: %s\n", strings.Join(res.LegacyMigrated, ", "))
```

`materializeWorktreeAgentSkills` forwards `stdout` into it, and it runs *before*
`writeWorktreeReport`, whose `--json` branch does
`json.NewEncoder(stdout).Encode(result)`. So `gc worktree ensure --json --agent
<a>` emits something like

```
materialized 20 skill(s) into /path/wt/.claude/skills: atomize, core.gc-work, ...
{"schema_version":"1","ok":true,"command":"worktree ensure",...}
```

and every `--json` consumer that unmarshals stdout fails — which is the mode the
formula `prepare-worktree` steps this flag exists for actually run in.

One-line fix, keeping the command's stdout as report-only:

```go
-	if !opts.DryRun && opts.Agent != "" {
-		materializeWorktreeAgentSkills(opts.Agent, rep.Path, stdout, stderr)
+	if !opts.DryRun && opts.Agent != "" {
+		materializeWorktreeAgentSkills(opts.Agent, rep.Path, stderr)
 	}
```

with the helper taking a single `stderr io.Writer` and calling
`materializeSkillsIntoWorkdir(cfg, &agent, cityPath, path, nil, stderr, stderr)`.
Every other diagnostic in this file already goes to stderr, so the materializer's
should too.

**2. Neither new test reaches the materializer.**

`TestCmdWorktreeEnsureAgentFlagAttemptsMaterialization` and
`TestCmdWorktreeEnsureNoAgentSkipsMaterialization` both call `clearCityEnv`, so
`resolveCity` fails and the path exits at the first guard — the assertion is on
the `"skill materialization skipped"` string. Nothing yet proves a skill sink
actually lands in the new worktree (and that is why the stdout issue above is
invisible to the suite). A happy-path test is cheap using the fixtures already in
`cmd/gc`:

```go
writeMaterializeTestCityFile(t, cityDir, "city.toml", "[workspace]\nname = \"test-city\"\n\n[beads]\nprovider = \"file\"\n")
writeMaterializeTestMayor(t, cityDir, "provider = \"claude\"\nstart_command = \"echo\"\n")
writeMaterializeTestCityFile(t, cityDir, "pack.toml", "[pack]\nname = \"test\"\nversion = \"0.1.0\"\nschema = 2\n")
writeBuiltinImportsFixture(t, cityDir, "core")
// ... opts.Agent = "mayor"; runWorktreeEnsure(...)
// then Lstat(filepath.Join(wt, ".claude", "skills", "core.gc-work")) is a symlink
```

I have both of those (the stream split and a happy-path + `--json`-purity test)
verified green — `make build`/`fmt-check`/`vet`/`lint` clean, `go test ./cmd/gc/
-run Worktree` green. Say the word and I'll open a follow-up PR against main
after this merges, or feel free to just take the diff above into this branch —
whichever you prefer.

One smaller note: `rep.Path` here vs the `spec.Path` I had used is the better
choice, since it is the path the report itself publishes.

PR #5977, opened 2026-09-03
fix(tmux): detect Claude's bare-gerund spinner, fix submit-confirm baseline, surface two swallowed errors (#5954)

### LINKED ISSUE #5954: Nudge submit-confirm is blind to Claude Code 2.1.x's busy footer; already-busy panes get a false confirm
`paneContainsBusyIndicator` (`internal/runtime/tmux/tmux.go`) cannot see Claude Code 2.1.x's busy
state. On 2.1.259 the spinner renders as `<glyph> <Gerund>...` (`Tinkering...`, `Clauding...`)
and only gains the `(4s · ↓ 22 tokens)` suffix ~1 s into a turn once tokens stream;
`esc to interrupt` no longer appears at all (0 occurrences in 90 samples of a live tool-using
turn). `claudeBusySpinnerRe` requires that parenthesised suffix, and `submitEnterAndConfirm`'s
budget is 4 polls x 150 ms = 600 ms per send — entirely inside the window where it does not exist.

Effects: every short turn reports `ErrNudgeSubmitUnconfirmed` although the message submitted and
was answered; two spurious extra Enters go into a live session per nudge; the caller requeues and
delivers duplicates.

Separately, `submitEnterAndConfirm` reads *any* busy pane as proof its own submit landed — it
never snapshots the pre-send state. On a pane already busy from the previous duplicate, a submit
whose Enter was lost is reported confirmed, the queue item is acked and deleted, and the text
sits drafted with nothing left to retry.

In production, four long-running sessions idled ~8 h (2026-09-02 22:50 -> 2026-09-03 07:00 UTC)
with a nudge typed and unsubmitted; one manual `tmux send-keys Enter` per pane cleared it.
Reproduced against 2.1.259 in a scratch tmux server.

Two adjacent defects found while tracing this: `WakePane`'s `resize-pane -y -1` fails with
`height too small` on a single-pane window and the error is discarded, so the detached-pane
SIGWINCH wake has never fired for a standard agent session; and `NudgeSession` does not call
`cancelCopyModeIfParked` the way `Respond` does, so a pane parked in copy-mode swallows the nudge.

This also resolves the open question in the `nudgeSubmitKeySequences` comment (#5012/#5013,
"LIVE RESIDUAL"), which ruled a busy-indicator detection gap out as the explanation.

Patch and full repro captures available; happy to open a PR.

rjgeng:
## What

Closes #5954. Four related defects in the nudge submit-confirm and wake paths, all found together while investigating false confirms on an already-busy pane.

## The four defects

1. **Bare-gerund spinner not recognized.** `paneContainsBusyIndicator`'s existing Claude-spinner regex requires a parenthesized `(Nm/Ns ...)` suffix, but Claude Code 2.1.x's spinner starts as a bare `<glyph> <Gerund>…` with no suffix for roughly its first second — exactly the window `submitEnterAndConfirm` polls in (4×150ms = 600ms per send). Added `claudeBareGerundSpinnerRe` alongside the existing pattern, verified against the file's own real glyphs (`·`, `✶`, `✢`, Unicode `…`) and against the existing idle-chrome negative test cases (no false positives introduced).
2. **`submitEnterAndConfirm`'s first send skipped the pre-send busy baseline** that later sends already take, so a stray busy reading on `send==0` got wrongly credited as "the nudge landed." Added the same pre-send snapshot to the first send; confirm now requires a genuine false→true transition. New test `TestSubmitEnterAndConfirmIgnoresPreexistingBusyOnFirstSend`, confirmed to fail on the unfixed code.
3. **`WakePane`'s `resize-pane` error was silently discarded.** Now logged (non-fatal, matching the file's one existing warning-print convention) instead of swallowed.
4. **`NudgeSession` never cancelled parked copy-mode** before sending keys, unlike `SendKeysDebounced`, which already does. Added the same `cancelCopyModeIfParked` call, in the same relative order.

Defects 3 and 4 got no dedicated new test — 3 is an unconditional side-effect print with no cheap seam to assert on, and 4's happy path chains ~10+ sequential tmux calls before reaching the submit sequence, making a scripted fake-executor test disproportionately brittle for this size of fix.

## Follow-up fix included in this PR

The full local push-gate run surfaced a gap: the new test this PR adds (`TestSubmitEnterAndConfirmIgnoresPreexistingBusyOnFirstSend`) wasn't registered in `scripts/runtime-tmux-tests.manifest`, a hand-maintained, position-sensitive inventory that backs a 6-shard round-robin CI split. Inserted the test at its exact canonical-discovery-order position (computed directly from the manifest test's own discovery logic, not guessed) and updated the three count assertions that shift as a result.

## Validation

- `go test -tags gms_pure_go ./internal/runtime/tmux/...` clean
- `TestRuntimeTmuxManifestMatchesCanonicalLinuxIntegrationInventory` and `TestRuntimeTmuxManifestSixShardsPartitionInventoryExactlyOnce` both pass
- Full local push-gate run: all failures across every shard and `unit-core` confirmed to be pre-existing environmental/timing flakes (dolt-compaction timeouts, detached-probe/systemctl-delegate/lsof tests, interrupt-timing tests) unrelated to this change — none touch `internal/runtime/tmux/tmux.go` or `scripts/runtime-tmux-tests.manifest`
- Pre-commit lint clean (0 issues) on both commits


sJohnsonStoever:
Data point from a Claude Code 2.1.272 seat (macOS, tmux, auto mode), captured while checking whether to move our seats off the 2.1.260 pin:

On 2.1.272 the bare-gerund spinner (`✻ Vibing…`) is visible for only the first ~1.5 s of a turn. While the reply streams there is **no spinner row at all** until the `✻ Churned for 16s · done` marker. Over 14 full-pane captures at 1.5 s intervals during one 400-word no-tools turn:

| binary | spinner row seen | `claudeBusySpinnerRe` (main) | `claudeBareGerundSpinnerRe` (this PR) |
|---|---|---|---|
| 2.1.260 | `✻ Computing… (12s · thinking with high effort)` | 13/14 | 14/14 |
| 2.1.272 | `✻ Vibing…` on one frame, then nothing | 0/14 | 1/14 |

`esc to interrupt` appeared on neither. So this PR fixes the first second of a 2.1.27x turn but the rest of a streaming reply still reads as idle to `paneBusy` / `WaitForIdle`. A detector for 2.1.27x probably needs a second signal, e.g. pane content changing between consecutive polls, or the absence of the `… · done` marker after a submit, rather than a footer string.

Happy to re-run the capture on any candidate regex.

Issue #5973, opened 2026-09-03
bug: gc session reset aborts before requesting the restart when the circuit-breaker clear fails (kill treats the same failure as a warning) — no fresh-context path for a named session


jacobhausler:
## Summary

`gc session reset` treats the named-session circuit-breaker clear as **fatal**: when the clear fails, the command prints the error and returns 1 **before** `handle.Reset` is ever called, so the fresh restart is never requested and the session keeps running its old conversation. `gc session kill` treats the *identical* failure on the *identical* code path as a **warning** and proceeds.

The result is that a transient (or, as we hit it, a persistent) fault in the circuit-breaker metadata store takes away the only way to give a named session fresh provider state — and `gc session kill` is not a substitute, because only reset rotates the resume key.

Verified on current `main` (`2f2cea0dc`, 2026-09-03).

## The asymmetry, in the source

Fatal — `cmd/gc/cmd_session_reset.go:95-101`:

```go
identity := namedSessionIdentity(bead)
if identity != "" {
    if err := resetSessionCircuitBreakerOnController(cityPath, sessionID, identity); err != nil {
        fmt.Fprintf(stderr, "gc session reset: clearing session circuit breaker for %q: %v\n", identity, err)
        return 1                      // <-- handle.Reset below is never reached
    }
}

if err := handle.Reset(context.Background()); err != nil {
```

Best-effort — `cmd/gc/cmd_session.go:2420-2428`:

```go
} else if identity != "" {
    if err := resetSessionCircuitBreakerAfterExplicitKill(cityPath, sessStore, sessionID, identity); err != nil {
        fmt.Fprintf(stderr, "gc session kill: warning: clearing session circuit breaker for %q: %v\n", identity, err)
        if killErr != nil {
            fmt.Fprintf(stderr, "gc session kill: %v\n", killErr)
            return 1
        }
    }
}
```

Both call the same `resetSessionCircuitBreakerOnController` (`cmd/gc/controller.go:351`) against the same controller socket command. Kill has already committed its side effect and treats the clear as bookkeeping; reset has committed nothing at that point and could equally well warn and continue, but instead throws the whole operation away.

## What we observed in production

A managed city with the supported whole-infrastructure split (`[storage.classes]` graph/sessions/messaging/orders/nudges -> one `sqlite-beads` binding). Every `gc session reset <named-session>` fails with:

```
[...]

PR #5980, opened 2026-09-03
fix(nudge): close only the store the frame opened, not the storage routes' shared engine

### LINKED ISSUE #5979: bug(nudge): a nudge frame closes the storage routes' shared engine on a relocated NUDGES class, so every later class read in the process fails with "sqlite store: bead store closed"
## Summary

On a city that relocates the `nudges` coordination class onto a storage binding, every
nudge frame that believes it owns the store it holds closes the **storage routes'
process-shared engine** instead. The routes memo is not detached by that close, so it goes
on serving a **closed** store for the rest of the process lifetime.

In a one-shot CLI invocation this is invisible (the process is about to exit anyway). In a
**long-lived** process — the controller — it is fatal and permanent: every later
coordination-class read in that process fails with

```
sqlite store: bead store closed
```

It needs no race and no shutdown. It fires the first time nudge maintenance runs a pass on
a queue that has work, which is why it appears hours into a controller run rather than at
boot, persists until the process exits, and recurs the next run.

Verified on current `main` (`f07cfb682`) with a test that fails before the fix and passes
after; see **Reproducer** below.

## Mechanism

1. `cmd/gc/nudge_beads.go:44-48` — `openNudgeBeadStoreErr` opens a work store with
   `openStoreAtForCity`, then returns
   `resolveNudgesStore(cliStorageRoutes(cityPath), store, …)`. When the NUDGES class is
   relocated, `resolveClassStore` (`cmd/gc/class_store.go:266`) **discards** that work
   store and returns the **routes'** binding engine — a handle owned by the routes memo
   and released only by `closeCLIStorageRoutes`.

2. `cmd/gc/cmd_nudge.go:1963-1969` — `nudgeMaintenanceStore.close()` calls
   `closeBeadStoreHandle(m.store.Store)` under the comment

   > It never touches a caller-passed store because this type only ever holds a store it
   > opened.

   That statement is true of the *open call* and false of the *handle*: on a relocated
   class the handle is the routes', and the store it did open is leaked instead. The same
   shape is in the `ownStore` arms of `enqueueQueuedNudgeWithStoreAndClock`
   (`cmd/gc/cmd_nudge.go:2161`) and `recordQueuedNudgeFailureDetailed`
   (`cmd/gc/cmd_nudge.go:2383`).

3. `cmd/gc/cli_storage_routes.go:196` — `closeCLIStorageRoutes` is the only thing that
   detaches the memo, and it runs only from the per-invocation CLI entry
   (`cmd/gc/main.go:181`). Nothing reopens a route it finds closed.

4. Every later funnel consumer in the process therefore gets `ErrStoreClosed`
   (`internal/beads/sqlite_store.go:499-505`) — nudge delivery itself, and the
   controller's `session-circuit-reset` socket handler, which reaches the funnel through
   `handleSessionCircuitResetSocketCmd` (`cmd/gc/controller.go:265`) →
   `cliSessionStore` → `cliStorageRoutes`.

`cmd/gc/cmd_order.go:2126` is **not** affected: it closes the raw work store it opened,
with no class resolver in between.

## What it looks like in production

A managed city with the supported whole-infrastructure split (`graph`/`sessions`/
`messaging`/`orders`/`nudges` → one `sqlite-beads` binding). 1h45m into a controller run,
both of these went dead together and stayed dead for the next 7.6 hours, through many
healthy reconcile ticks:

- supervisor-side nudge delivery, and
- `gc session reset <named-session>`, which aborts at its circuit-breaker clear:

```
gc session reset: clearing session circuit breaker for "mayor": loading session circuit
breaker metadata for gcg-502: sqlite store: bead store closed
```

Side-effect-free probe against a live controller (the handler's first act on that path is
a read):

```
$ printf 'session-circuit-reset:{"identity":"probe","session_id":"gcg-000000"}\n' | nc -U .gc/controller.sock
{"outcome":"failed","error":"loading session circuit breaker metadata for gcg-000000: sqlite store: bead store closed"}
```

Healthy controller: a not-found error for `gcg-000000`.

This is the underlying condition behind #5973 (which reports the separate, independent
defect that `gc session reset` treats that failure as fatal and never reaches
`handle.Reset`). The two are complementary: this one is why the clear fails; #5973 is why
the failure costs you the whole reset.

## Reproducer

`cmd/gc` test on the repo's own `migratedOneShotCLICity` fixture: build a converged split
city, read the sessions class through the routes (healthy), run one nudge maintenance
frame, read again.

Before the fix, on `f07cfb682`:

```
--- FAIL: TestNudgeMaintenanceFrameLeavesTheStorageRoutesServingALiveStore (0.15s)
    nudge_store_ownership_test.go:43: the nudge maintenance frame closed the storage
    routes' shared engine; the routes still serve it and every later class read fails:
    sqlite store: bead store closed
```

After: PASS. The test ships with the PR.

## Fix candidates

**A — close only what the frame opened (the PR).** Give the nudges frames a close half that
declines to close a handle the routes own (`closeOwnedNudgeStore`), and skip the open whose
result `resolveNudgesStore` would only discard. Small, local to `cmd/gc/nudge_beads.go` +
the three `cmd_nudge.go` close sites, and it also stops the per-frame leak of the work
store on relocated cities.

**B — make the ownership un-loseable in the resolver.** Have `resolveClassStore` return an
explicit ownership token (or a no-op closer) alongside the store, so no caller can hold a
borrowed handle that looks like an owned one. Bigger, and it would retire this whole class
of bug rather than this instance — there are ~21 direct `cliStorageRoutes` call sites and
`residency_topology.go:96-105` already warns that the funnel is the wrong side of the
controller boundary.

I've sent A. Happy to follow with B if you want the resolver contract tightened.

## Prior art

#3157 (closed) is the same family one class-resolver earlier: a per-tick close of a store
the closer did not exclusively own, producing `bead store closed` on a native-Dolt city.

jacobhausler:
Fixes #5979.

## What was wrong

On a city that has relocated the NUDGES class onto a storage binding,
`openNudgeBeadStoreErr` (`cmd/gc/nudge_beads.go`) opens a work store and then hands it to
`resolveNudgesStore`, which **discards** it and returns the storage routes' process-shared
binding engine. The nudge frames then close what they hold, believing they own it:

```go
// close releases the store this frame opened (if any). It never touches a
// caller-passed store because this type only ever holds a store it opened.
func (m *nudgeMaintenanceStore) close() error {
	if !m.opened {
		return nil
	}
	return closeBeadStoreHandle(m.store.Store)   // <- the routes' engine, on a relocated class
}
```

Closing it does not detach the routes memo, so the memo goes on serving a **closed** store.
In a one-shot CLI invocation that is invisible. In the controller it is permanent: every
later coordination-class read in the process fails with `sqlite store: bead store closed` —
nudge delivery, and the `session-circuit-reset` socket handler, which is how it surfaced
for us (see #5979, and #5973 for the reset path that then throws the whole operation away).

It needs a non-empty nudge queue, since `nudgeQueueHasWork` gates `ensureOpen` — which is
why it lands hours into a controller run rather than at boot, and why only a process
restart clears it.

## The change

- `closeOwnedNudgeStore(cityPath, store)` — the close half of "own the store I opened".
  It declines to close a handle the routes own, because a frame cannot tell that from the
  value alone. Used at the three sites that close a nudges-class store they believe they
  opened: `nudgeMaintenanceStore.close`, and the `ownStore` arms of
  `enqueueQueuedNudgeWithStoreAndClock` and `recordQueuedNudgeFailureDetailed`.
- `openNudgeBeadStoreErr` skips the work-store open entirely when the class is relocated,
  since `resolveNudgesStore` would only discard it. That also stops a leaked work-store
  handle per call on relocated cities.
- The two comments that asserted the false invariant now say what is actually true.

`cmd/gc/cmd_order.go` is deliberately untouched: it closes the raw work store it opened,
with no class resolver in between, so it already owns what it closes.

## Evidence

New test `TestNudgeMaintenanceFrameLeavesTheStorageRoutesServingALiveStore`
(`cmd/gc/nudge_store_ownership_test.go`), on the repo's own `migratedOneShotCLICity`
fixture: read the sessions class through the routes (healthy), run one nudge maintenance
frame, read again.

Without the fix, on `f07cfb682`:

```
--- FAIL: TestNudgeMaintenanceFrameLeavesTheStorageRoutesServingALiveStore (0.15s)
    nudge_store_ownership_test.go:43: the nudge maintenance frame closed the storage
    routes' shared engine; the routes still serve it and every later class read fails:
    sqlite store: bead store closed
```

With it: PASS, and the whole `./cmd/gc` package passes (`go test ./cmd/gc -count=1`, 308s).
`make build` and `make vet` are clean. `make lint` reports the same 160 pre-existing issues
as the base and none in the changed files.

The existing `TestNudgePollHelpersSkipDoltOpenOnEmptyQueue` still pins the owned-handle
accounting, so the "close what you opened" case has not gone untested.

## Note on scope

This closes the instance. The general shape — `resolveClassStore` returning either a
borrowed or an owned handle with nothing in the type to tell them apart — is still there,
and `residency_topology.go:96-105` already warns that the one-shot funnel is the wrong side
of the controller boundary while ~21 direct `cliStorageRoutes` call sites remain on it. If
you want the resolver contract tightened (an explicit ownership token or a no-op closer
returned alongside the store), I'm happy to follow up with that as a separate change.

### DIFF
--- a/cmd/gc/cmd_nudge.go
+++ b/cmd/gc/cmd_nudge.go
@@ -1923,16 +1923,20 @@ func queuedNudgeClaimableForTarget(target nudgeTarget, item queuedNudge) bool {
 // main pool + a SHOW DATABASES init probe) every 2s. See
 // TestNudgePollHelpersSkipDoltOpenOnEmptyQueue.
 //
-// It closes exactly the handle it opened, and only if it opened one, preserving
-// the ownership contract the …WithStore variants model. "The handle it opened"
-// is not the same as "the store it holds": on a city that has relocated the
-// NUDGES class, the store is the storage routes' process-shared engine, which
-// this frame must not close. closeOwnedNudgeStore is the predicate.
+// It closes exactly the handle it opened, and only if it opened one,
+// preserving the closeBeadStoreHandle ownership contract the …WithStore
+// variants model. "The handle it opened" is not "the store it holds": once the
+// NUDGES class relocates, the store is the storage routes' process-shared
+// engine, which this frame holds but must never close (openOwnedNudgeBeadStore).
 type nudgeMaintenanceStore struct {
 	cityPath string
 	opened   bool
 	store    beads.NudgesStore
-	front    *nudgequeue.Store
+	// handle is what ensureOpen actually opened, which is not always
+	// store.Store: see openOwnedNudgeBeadStore. close releases this and
+	// nothing else.
+	handle beads.Store
+	front  *nudgequeue.Store
 }
 
 // frontForState returns the front-door handle to use for the maintenance passes
@@ -1954,7 +1958,7 @@ func (m *nudgeMaintenanceStore) frontForState(state *nudgeQueueState) *nudgequeu
 func (m *nudgeMaintenanceStore) ensureOpen() beads.NudgesStore {
 	if !m.opened {
 		m.opened = true
-		m.store = openNudgeBeadStore(m.cityPath)
+		m.store, m.handle = openOwnedNudgeBeadStore(m.cityPath)
 		if m.store.Store != nil {
 			m.front = nudgeFrontDoor(m.store)
 		}
@@ -1962,14 +1966,14 @@ func (m *nudgeMaintenanceStore) ensureOpen() beads.NudgesStore {
 	return m.store
 }
 
-// close releases the store this frame opened (if any). It never touches a
-// caller-passed store because this type only ever holds a store it opened — nor
-// the storage routes' shared engine, which it holds but does not own.
+// close releases the handle this frame opened (if any). It never touches a
+// caller-passed store because this type only ever opens its own, and never the
+// storage routes' shared engine, which it may hold but does not own.
 func (m *nudgeMaintenanceStore) close() error {
 	if !m.opened {
 		return nil
 	}
-	return closeOwnedNudgeStore(m.cityPath, m.store.Store)
+	return closeBeadStoreHandle(m.handle)
 }
 
 // nudgeQueueHasWork reports whether the queue holds any item a maintenance pass
@@ -2156,12 +2160,13 @@ func enqueueQueuedNudgeWithStore(cityPath string, store beads.NudgesStore, item 
 
 func enqueueQueuedNudgeWithStoreAndClock(cityPath string, store beads.NudgesStore, item queuedNudge, clk clock.Clock) error {
 	ownStore := false
+	var opened beads.Store
 	if store.Store == nil {
-		store = openNudgeBeadStore(cityPath)
+		store, opened = openOwnedNudgeBeadStore(cityPath)
 		ownStore = true
 	}
 	if ownStore {
-		defer closeOwnedNudgeStore(cityPath, store.Store) //nolint:errcheck // best-effort
+		defer closeBeadStoreHandle(opened) //nolint:errcheck // best-effort
 	}
 	var front *nudgequeue.Store
 	if store.Store != nil {
@@ -2378,12 +2383,13 @@ func recordQueuedNudgeFailureDetailed(cityPath string, store beads.NudgesStore, 
 		return nil, nil
 	}
 	ownStore := false
+	var opened beads.Store
 	if store.Store == nil {
-		store = openNudgeBeadStore(cityPath)
+		store, opened = openOwnedNudgeBeadStore(cityPath)
 		ownStore = true
 	}
 	if ownStore {
-		defer closeOwnedNudgeStore(cityPath, store.Store) //nolint:errcheck // best-effort
+		defer closeBeadStoreHandle(opened) //nolint:errcheck // best-effort
 	}
 	var front *nudgequeue.Store
 	if store.Store != nil {
--- a/cmd/gc/cmd_nudge.go
+++ b/cmd/gc/cmd_nudge.go
@@ -1923,9 +1923,11 @@ func queuedNudgeClaimableForTarget(target nudgeTarget, item queuedNudge) bool {
 // main pool + a SHOW DATABASES init probe) every 2s. See
 // TestNudgePollHelpersSkipDoltOpenOnEmptyQueue.
 //
-// It owns the handle it opens and closes exactly that handle (and only if it
-// opened one), preserving the closeBeadStoreHandle ownership contract the
-// …WithStore variants model.
+// It closes exactly the handle it opened, and only if it opened one, preserving
+// the ownership contract the …WithStore variants model. "The handle it opened"
+// is not the same as "the store it holds": on a city that has relocated the
+// NUDGES class, the store is the storage routes' process-shared engine, which
+// this frame must not close. closeOwnedNudgeStore is the predicate.
 type nudgeMaintenanceStore struct {
 	cityPath string
 	opened   bool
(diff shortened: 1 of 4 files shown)


jacobhausler:
Force-pushed a restructure. The first cut failed `TestResidencyBoundaryGrepRatchet`, and the guard was right — not incidentally, but on the substance:

```
RESIDENCY-BOUNDARY: cmd/gc/nudge_beads.go openNudgeBeadStoreErr b:routes.storeFor:
1 sites, baseline 0 — a NEW store-enumeration site.
```

That cut fixed the close by asking the routes a second time whether the NUDGES class was relocated (`closeOwnedNudgeStore`). `residency-boundary-patterns.txt` rule (b) names exactly that shape: *"direct binding-gate re-derivation. Asking «is this class relocated» a second time is how the split-store bug class reproduces (#5125, #5127)."* A second derivation of the same fact is the bug this PR is fixing, one layer up — so annotating it `residency:allow` would have papered over the defect rather than fixed it. Worth noting the first cut also evaded the grep by accident: the close half spelled it `cliStorageRoutes(cityPath).storeFor(...)`, which the `(^|[^A-Za-z0-9_])routes\.storeFor\(` pattern does not match, so only one of the two sites was ever reported.

The re-derivation is now gone entirely. `openNudgeBeadStoreOwned` is the single place the nudges-class store is opened, and it decides the class store and the handle the call opened **together**, returning both; `openNudgeBeadStore`, `openOwnedNudgeBeadStore` and `openNudgeBeadStoreErr` are projections of it. The closing frames — `nudgeMaintenanceStore` (the frame behind all seven poll/maintenance passes) and the two `ownStore` arms — now close *the handle they opened* instead of *the store they hold*. The ownership answer stays with the resolver, which is what the contract reserves to it, and no residency vocabulary is consulted anywhere.

This also releases a handle that previously leaked: on a relocated city the work store `openStoreAtForCity` opened was discarded by the resolver and closed by nobody. It is now the thing that gets closed.

The counting fake moved with it, from `openNudgeBeadStore` to `openOwnedNudgeBeadStore` — the seam the closing frames actually call, and the one `openNudgeBeadStore` delegates to, so a test substituting it moves both forms at once and `opens == closes` cannot silently miss an open.

Verification on the builders lane at `a1df702f0`:

- `go test ./scripts -run TestResidencyBoundary` — ok (was the failing check)
- `TestNudgeMaintenanceFrameLeavesTheStorageRoutesServingALiveStore`, `TestNudgePollHelpersCloseEveryStoreTheyOpen`, `TestNudgePollHelpersSkipDoltOpenOnEmptyQueue` — all PASS
- full `go test ./cmd/gc -count=1` — ok, 226.8s

The regression test itself is unchanged from the cut that was proven RED on the parent, so that evidence stands as reported ('verify/cr-01mc3-test-only' carries the test alone on `origin/main`).

jacobhausler:
The residency ratchet that was failing is now green (`Integration / packages-core-4-of-4` — SUCCESS). The remaining red is an unrelated flake in `internal/productmetrics`, which this PR does not touch:

```
--- FAIL: TestDisableAndPurgeBoundsInitialAndPostUploaderStateLocks/post-uploader_state_lock (10.70s)
--- FAIL: TestRecordOnceDecisionGatesStorageSafeBoundaries/ENOENT_before_directory_creation (0.10s)
FAIL github.com/gastownhall/gascity/internal/productmetrics  185.145s
```

`Evidence / critical-path suites`, `CI / integration` and `CI / required` are all downstream of that one job.

Why it is not this diff:

- The PR changes four files, all in `cmd/gc` — `nudge_beads.go`, `cmd_nudge.go`, `cmd_nudge_test.go`, `nudge_store_ownership_test.go`. `internal/productmetrics` is a separate package with its own test binary; nothing in the diff is reachable from it.
- `Integration / packages-core-3-of-4` PASSED on the previous head of this same PR (`eb5f2ec69`) and failed on this one, while the only change between the two heads is those three `cmd/gc` files.
- Locally, on the same revision CI failed (`a1df702f0`): the full package is green — `go test ./internal/productmetrics -count=1` → ok, 126.2s — and the two named tests pass at `-count=3`. They also pass at the merge base `f07cfb682`.

Both failing tests are timing/filesystem-sensitive (a 10.7s state-lock bound, and an ENOENT-before-mkdir ordering check), which is the usual shape of a loaded-runner flake.

I do not have re-run rights on this repo, so I am re-triggering CI by closing and reopening this PR — the head commit `a1df702f0` is unchanged. If it stays red on the same two tests, that would be worth a separate flake issue rather than a change here.

PR #5988, opened 2026-09-03
fix(beads): hold the pinned-id fence in transactions and key parent locality on the store namespace (s3 follow-up)

julianknutsen:
Follow-up to #5642 (merged as `1cb060c35b`). These are the **four review
findings from the s3 Fable review**. They were written against the pre-merge s3
head (`5c4e7faefd`) and never pushed, because #5642 merged first with a
different set of pipeline-authored fixes for the same review round. Three of the
four findings were resolved differently there — or not at all — so this PR
carries them forward against merged `main` rather than re-litigating the merge.

Refs `ga-6od57`, `ga-j4p3y`.

## What merged `main` has, and what this changes

| # | Finding | Merged `main` | This PR |
|---|---------|---------------|---------|
| 1 | Pinned-id fence bypassed by `Store.Tx().Create` | `sqliteStoreTx.Create` has **no** fence call — it goes straight to `normalizeCreate` | `t.store.checkPinnedIDNamespace(b.ID)` runs first, before normalization |
| 2 | Parent locality keyed on the **child's** prefix | `nativeParentIsLocal(issueID, parentID, storePrefix)` tests the child's prefix first, store prefix only as fallback | store's mint prefix first, child's prefix second |
| 3 | Prefixless store disagrees between `Create` and `Update` | no `parentIsLocalForCreate`; the create arm cannot ask the question at all | `parentIsLocalForCreate` resolves the namespace a minted child lands in |
| 4 | `Bead.ParentID` contract hides the bd-CLI exception | "scoped" by one sentence saying the contract holds by convention on bd | names what bd actually does with `--parent`, plus `beads.md` invariant 17 |

Finding 5 in the original review (real-Dolt coverage) rides along in commit 4:
merged `main` has no cross-prefix parent test against a real backend.

### 1. The fence was a convention, not a fence

`sqliteStoreTx.Create` never consulted `checkPinnedIDNamespace`, so every claim
invariant 16 makes was false through the second door the same store exposes. A
fenced binding accepted `kitwork-424242` inside a `Tx`, **kept the row**, and the
next mint jumped to `kitn-424243` — a caller refused at the front door had
renumbered a binding it was never allowed to write to.

No production caller pins an id inside a `Tx` today (`internal/api/rigidem.go`
and `internal/extmsg/binding_service.go` both mint), so this is latent rather
than live. It matters because #5642 promoted the fence from a SQLite-specific
test to a **provider contract**, and s5 builds every provider on it.

The check goes at the top, before normalization, for the same reason `create()`
has it there: `normalizeCreate` lifts the sequence floor to the pinned id's
suffix, and a rollback does not put that back.

### 2 + 3. The namespace boundary is the store's, on both arms

`nativeParentIsLocal` asked the child's prefix first. For a row carrying a prefix
this store does not mint — a pinned id, a relic a storage migration copied in — a
parent inside the store's **own** namespace read as foreign, so
`validateUpdateParent` and `updateParentInTransaction` skipped resolution and the
reparent landed dangling. That is precisely the case `Bead.ParentID` promises is
refused.

The create arm could not ask the question at all: on a mint the child has no id
yet, so a store declaring no prefix — the shape the conformance fixture opens —
answered "foreign" for every parent, and `Create` admitted what `Update` refused.
`parentIsLocalForCreate` resolves the namespace a minted child will land in
(`s.idPrefix`, else `storage.GetConfig(ctx, "issue_prefix")`), so both arms agree.

The parent-child edge also no longer runs through
`shouldPrevalidateNativeDependency`, whose cross-prefix rule would skip a parent
the store owns whenever the child carries a different prefix.

Both are narrow by construction, and the tests carry the controls that say so: a
genuinely foreign parent (`gcg-`, another class's namespace) is still carried
verbatim on `Create` and on `Update`, on both fixtures.

As with finding 1, this is latent rather than live so far as could be
determined: reaching either shape needs a row carrying a prefix this store does
not mint and then reparented into the store's own namespace, or a store that
declares no prefix at all, and no production path was found that does — every
production `NativeDoltStore` opens with a declared prefix, where the old and new
rules agree on a mint. That is a negative claim over a large tree, so hold it
more loosely than finding 1's; it matters for the same reason, that s5 builds on
`Bead.ParentID` meaning the same thing on both arms.

### 4. Contract honesty

`Bead.ParentID` claimed a foreign parent is carried "without question" by every
backend. `BdStore` does the opposite on all four arms: it passes `ParentID` to bd
as `--parent`, and bd resolves it unconditionally — an id bd cannot see fails the
command outright, and one it can see **supplies the child's id**, so placement
follows the parent's namespace instead of the child's class. That is the one
outcome no later copy can undo.

Nothing in CI can see it: the only `RunStoreTests` wrapper over a real bd is
unconditionally skipped (`ga-e7z613`).

`BdStore` is left alone — it follows the tool it drives. What changes is the
honesty of the contract, plus `beads.md` invariant **17**. `ga-6od57` tracks the
actual fix: make the provider comply, or un-skip the run.

## RED → GREEN on merged `main`

Each production file was reverted to its `origin/main` content in place, the new
tests run, then restored. Exact failure lines:

**Finding 1** — `internal/beads/sqlite_store.go` at `origin/main`:

```
--- FAIL: TestSQLiteStorePinnedIDFenceConformance/TheFenceHoldsInsideATransaction
    pinned_id_fence_conformance.go:186: the transactional Create accepted a foreign pinned id; a fence on the direct door only is a convention the next composite write breaks
--- FAIL: TestSQLiteStorePinnedIDFenceConformance/ATransactionalRefusalDoesNotConsumeTheMintSequence
    pinned_id_fence_conformance.go:211: the transactional Create accepted a foreign pinned id
```

The over-restriction control
`OverRestrictionControls/ATransactionAdmitsAMintAndAnInNamespacePin`
**PASSES** on `origin/main`. That pairing is what makes this a fence and not a
blanket refusal — a `Tx` that refused every create would fail the control instead.

**Findings 2 and 3** — `internal/beads/native_dolt_store.go` at `origin/main`:

```
--- FAIL: TestNativeDoltStoreResolvesAParentInItsOwnNamespaceForAForeignPrefixedChild
    native_dolt_store_parent_namespace_test.go:30: Update("gc-1", parent "ga-999999") = <nil>, want ErrNotFound: the parent is inside the namespace this store mints, so its absence is a row this store can see
    native_dolt_store_parent_namespace_test.go:37: ParentID after the refused reparent is "ga-999999", want it unchanged; the refusal has to come before the write
    native_dolt_store_parent_namespace_test.go:41: Create(child "gc-2", parent "ga-999999") = <nil>, want ErrNotFound; Create and Update have to agree
    native_dolt_store_parent_namespace_test.go:44: after the refused create Get("gc-2") = <nil>, want ErrNotFound
--- FAIL: TestNativeDoltStoreWithoutADeclaredPrefixAgreesBetweenCreateAndUpdate
    native_dolt_store_parent_namespace_test.go:79: Create(parent "gc-999999") = <nil>, want ErrNotFound: the parent is in the namespace this store mints under, and Update refuses the same value
```

Both files restored → **PASS**.

`TestNativeDoltStoreRealBackendCrossPrefixParent` (commit 4) is coverage, not a
regression test, and passes as written — no red was expected. Its own comment
records why the timing claim is pinned on the in-memory fixture instead: a
mutation probe that skips every parent-child edge on the create path leaves the
real-backend test green, because the library refuses the same value after the
committed create.

## Conflicts and resolutions

Only `internal/beads/beads.go` conflicted. `native_dolt_store.go`,
`sqlite_store.go`, `pinned_id_fence_conformance.go` and `beads.md` are
byte-identical between `main` and the cherry-pick base, so those applied clean;
`beadstest/conformance.go` auto-merged (main's disjointness assertion and the
new placement-control comment sit ten lines apart in the same function).

- **Commit 2** — kept #5642's scoping sentence (which names the providers the
  suite executes and the skipped bd run) and replaced only the sentence saying
  *where* the namespace boundary is read from, since `main`'s "outside the
  namespace it mints" is the wording the store-first rule corrects.
- **Commit 3** — dropped #5642's scoping sentence rather than keeping it
  alongside. The text here states the same two facts and more: same provider
  enumeration, same `ga-e7z613`, plus what bd actually does. Keeping both would
  assert twice, and `main`'s "there the contract holds by convention" reads as
  "bd probably complies, unverified" — the claim this commit exists to correct.
- **`beads.md`** — `main`'s last invariant is 16, so the new one keeps its
  number 17; no renumbering needed.

No test in `main` asserted the child-first rule, so no test rewrites were
required — `nativeParentIsLocal` has no direct test references, and the full
`internal/beads/...` suite is green under the store-first rule.

## Gates

| Command | Result |
|---|---|
| `go build ./...` | exit 0 |
| `go vet ./...` | exit 0 |
| `gofmt -l internal cmd` | exit 0, nothing printed |
| `golangci-lint run ./internal/beads/... ./internal/storebinding/... ./cmd/gc/...` | exit 0, **0 issues** |
| `go test ./internal/beads/... ./internal/storebinding/... ./internal/storeref/... -count=1` | exit 0 |
| `go test -tags integration ./internal/beads -run 'NativeDoltStoreRealBackend' -count=1` | exit 0 — **ran on real Dolt, not skipped** |
| `go test ./cmd/gc -run 'Residency\|Binding\|Fence\|ParentID\|Namespace' -count=1` | exit 0 |
| `make check-docs` | exit 0 |
| `make test-fast-parallel` | exit 0 — all 10 shards `ok`, 1944s |

The integration run used `dolt 2.1.8` / `bd 1.1.0 (Enterprise)`:
`TestNativeDoltStoreRealBackendCrossPrefixParent` **4.54s**,
`TestNativeDoltStoreRealBackendRoundTrip` 4.97s, package 9.62s, 23s wall. A skip
returns in milliseconds, and the log contains no `SKIP` lines.


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277296770

quad341 (dismissed):

The part worth calling out is the RED→GREEN section. I reverted each production hunk in place and re-ran your tests before believing any of it, and every failure line reproduced exactly as written — including `OverRestrictionControls/ATransactionAdmitsAMintAndAnInNamespacePin` passing on `origin/main`. That pairing is what makes the transactional rows a fence rather than a blanket refusal, and it's the easiest control to skip.

Carrying the findings forward against merged `main` instead of re-litigating the #5642 merge was the right call, and the conflict notes on which scoping sentence you dropped and why saved reconstructing it.

Two things I traced and want on record. Your "no `allowForeign` exemption inside `Tx`" reasoning holds — only `SQLiteStore` implements `CreateWithForeignID`, and it never routes through `sqliteStoreTx`. And the store-first rule makes `NativeDoltStore` slightly stricter than its own library in the relic case, where `isCrossPrefixDep` would classify that edge external and never resolve it. Invariant 17 covers it, so it's documented behavior — just worth knowing the library isn't backstopping that particular refusal.

Good call leaving `ga-6od57` open rather than papering over the bd divergence.


---

quad341 (approved):
@julianknutsen — thank you for carrying the four s3 findings forward against merged `main` instead of re-litigating the #5642 merge. The RED→GREEN section did real work: reverting each production hunk in place and quoting the exact failure lines meant I could check your claims instead of taking them, and your notes on which scoping sentence you dropped saved me reconstructing that reasoning from scratch.

`main` has moved under you again since you pushed. #5644 landed the transactional fence in `sqliteStoreTx.Create` in exactly the position you chose, so finding 1 arrives already done. It also renamed `nativeBeadIDPrefix` to `beadIDPrefix`, which your new code and both test files still call, so the merged tree would not compile as-is.

That is integration work and it is mine. I will rebase, take `main`'s side on the SQLite conflict, keep your store-first `nativeParentIsLocal` body, sweep the rename, and re-run the beads suites plus the real-Dolt integration test before merging. Your commits and authorship stay as they are.

Findings 2, 3 and 4 are the real content here and they land intact. Invariant 17 naming what bd actually does with `--parent` is the part I most wanted written down.


---

PR #5604, opened 2026-08-25
test(rigidem): pin the rig-create idem record to the work class

julianknutsen:
## Summary

`createIdemRecord` writes the durable `(city, request_id)` record straight to the
store it is handed — there is no placement hop on that path — and that is only
correct while the record classifies as **work**. Today it does, but only by
falling through every arm of `coordclass.Classify`: `gc-idem` is not a class
marker.

Nothing states that dependency. An arm added later would silently convert every
idem write into a stranded one on a split city: the record lands in the work
ledger while the class's own readers follow the relocated binding, and the
rig-create idempotency axis stops converging. The only thing that would notice is
`confirmInfraConvergence` (`cmd/gc/infra_class_migrate.go`), which re-checks
containment on every boot of a converged split city and refuses to serve while
naming the stranded ids — after the writes are already stranded, and not at all
on a city that converged before its copy manifest was recorded.

Failing at build time instead is why this gets a pin rather than a seam — the
fix, if it ever fires, is to route the write through `Server.createStoreForBead`
(`internal/api/residency_create.go`), which plans and then resolves placement,
and the pin says so. The
test classifies the bead `createIdemRecord` **actually hands to `tx.Create`**,
captured through a recording store, rather than a hand-copied literal, so a change
to the written shape is carried into the assertion instead of drifting away from
it.

Follow-on to #5597 (epic `ga-qdt5y`). Closes `ga-yx7h7`. Test-only; **no
production behavior changes.**

## Testing

- [x] `go build ./...` clean
- [x] `go vet ./...` clean (also re-run by the pre-commit hook with
      `-mod=readonly`)
- [x] `go test ./internal/api/` — ok 92.1s (full package; the only package touched)
- [x] `go test ./internal/coordclass/` — ok 0.13s (proves no toy-arm residue from
      the mutation proof)
- [x] `go test ./internal/api/ -run TestRigIdem -v` — 22/22 PASS including the new
      `TestRigIdemRecordsStayWorkClass`; the pre-existing
      `TestRigIdemRecordClosedAtCreateNeverReady` (nearest neighbour) unchanged
      and green
- [x] `gofmt -l` on both touched and mutated files — empty
- [x] Pre-commit hook active and ran on the staged change:
      `lint-changed: ./internal/api  0 issues`
- [x] **Mutation proof, three ways, each killing a different assertion:**
      - add a toy `coordclass` arm matching the `gc-idem` label → **the pin dies
[...]

Issue #3201, opened 2026-06-07
bug(order-dispatch): `gc reload --soft`/`gc doctor` still hang — N serial per-order last-run bd-queries on a cold-cache dispatcher (residual of #3191; #3197 cap necessary-but-insufficient)


iwata-1116:
## Summary

Residual of #3191. The merged fix #3197 (`fix(beads): cap bd query timeout`,
commit `01ef670b2`) capped each individual `bd query` at 30s, but did **not**
address the *count*, the *serial* execution, or the *reload cold-cache* that
fires them all. `gc reload --soft` (and `gc doctor`) still hang: observed
>90s → RC=124 on a multi-rig city. #3197 was necessary-but-insufficient; #3191
correctly stayed open.

## Root cause (function-anchored; line numbers drift)

1. The reload reply is gated on the controller `tick()` completing — the reply
   is sent only at `completeManualReload` (`cmd/gc/city_runtime.go`).
2. A manual reload runs `reloadConfigTraced`, which **rebuilds** the order
   dispatcher (`cr.od = memoryOrderDispatcher`) with an **empty** `lastRunCache`
   (`cmd/gc/city_runtime.go`).
3. The **same** tick then runs `cr.dispatchOrders` (intentionally *before* the
   reply, to avoid starving due formulas on AUTO ticks) against that
   cold-cache dispatcher (`cmd/gc/order_dispatch.go`).
4. `dispatch` loops orders **serially**. Each cooldown/cron order
   (`orderTriggerUsesLastRun`) → `cachedLastRun` cold miss →
   `orderDispatchTrackingIndex.lastRunFunc` → `orders.LastRunAcrossStores` →
   `store.List{TierBoth, Limit:1}` → a `bd list` **plus** a `bd query` per order.
5. With ~79 orders (4 rigs + city) × up to 2 stores × (list+query), all serial,
   the worst case is ~79 × up-to-30s ≫ the ~60s reload-accept timeout → RC=124.

So the 30s cap (#3197) bounds each child but the storm is the *count × serial ×
cold-cache*, which the cap does not touch.

## Proposed fix (this PR)

**PRIMARY — kill the count.** Gate the per-order fallback in
`orderDispatchTrackingIndex.lastRunFunc` on an **index miss**. The in-memory
batch index (sorted-desc, limit 2048) is authoritative for any recently-run
order: if it has a non-zero entry, trust it; only fall back to the per-order
[...]

PR #5990, opened 2026-09-04
fix(orders): dispatch due condition-triggered orders outside the per-tick rotation budget

julianknutsen:
## The defect

`orders.max_dispatches_per_tick` (default 4) rationed **every** trigger type and
advanced its rotation cursor (`nextDispatchStart`) only when it spent a slot. So
each tick fired the first 4 due orders after the cursor and every order got a
dispatch opportunity roughly once per `ceil(orders / budget)` ticks.

That is the right shape for maintenance sweeps. It is the wrong shape for a
condition-triggered order, whose check does not say "I would like a turn" — it
says **work is pending right now**, and it says it only for as long as that work
waits. A cooldown order is due at every tick whose spacing exceeds its interval,
so on a slow-tick city the sweeps saturate the budget permanently and a condition
order whose window is shorter than a full rotation fires only by luck.

## Evidence (maintainer-city, read-only from the graph store and journal)

- `order:*` tracking beads are created in bursts of **exactly 4**, one burst
  every ~18–24 min: 20:04:41, 20:22:21, 20:24:25, 20:42:00, 21:00:11, 21:23:18,
  21:47:10, 22:06:07, 22:25:15, 22:33:59, 22:44:19 … — 12–16 dispatches/hour
  against ~40 enabled orders, i.e. a full rotation every ~3.5 h
  (pr-review-router fired 16:18, 20:24, 23:03). `city.toml` sets no override,
  so the budget is the built-in 4; the tick period on this city is ~20 min, not
  the ~110 s the `max_wakes_per_tick` comment nearby assumes.
- `pr-merge-queue` (trigger=condition, check `pr_merge.py queue-pending`, 2.8 s,
  exit 0 whenever a merge bead is `pr_merge.status=ready`) last fired **14:06Z —
  11 h earlier**. Merge beads for #5604/#5633 were ready 20:25–21:11Z. The ticks
  at **20:42Z and 21:00Z** fell inside that window and each fired four *cooldown*
  orders — beadless-session-reaper, pr-review-label-poll, quota-stall-reset,
  beads-health / cross-rig-deps, gate-sweep, nudge-mail-sweep,
  order-tracking-sweep — while the merge queue's condition was true. Same for the
  21:2x–21:3x window (#5918/#5928/#5929) and the 01:3x window (#5935).
- No gate error, no `order.suppressed` event, no timed-out-check line ever named
  `pr-merge-queue`. It was simply never reached before the budget was spent. The
  starvation was invisible.

Raising `max_dispatches_per_tick` on this city only moves the cliff, which is why
this is an SDK defect rather than a tuning problem.

## The change (`cmd/gc/order_dispatch.go`, `memoryOrderDispatcher.dispatch` phase 2)

Phase 2 is now two passes over the same rotation-ordered candidate list:

1. **Unbudgeted pass** over due condition candidates. Every gate still applies —
   the phase-1 open-tracking gate already ran, the open-work gate runs here, and
   the order self-limits by its own check — but no budget is consulted and no
   rotation cursor moves.
2. **Budgeted pass** over everything else, byte-for-byte the previous semantics:
   same `start`, same `spendDispatchBudget(idx)`, same `return` on exhaustion.
   `maxDispatchesPerTick == 0` still means unbounded.

Both passes call one shared `fireCandidate(...) bool`, so the trigger-env
failure record, the prefetched condition verdict, the lastRun refresh, the
open-work gate and the launch exist exactly once. The open-work gate block moved
into its own `openWorkGateShut` helper to keep both extracted functions under the
repo's complexity threshold.

**Which trigger spelling the branch keys on:** `Order.Trigger == "condition"`,
and that covers both spellings. `gate = "condition"` is folded into `Trigger` by
`orders.orderDecode.normalized()` (`internal/orders/order.go:151`) at scan time
and by `config.OrderOverride.normalizeLegacyAliases()`
(`internal/config/config.go:2173`) at override time, so no `Order` reaches the
dispatcher still carrying a `Gate`. The dispatcher's existing condition branches
(`prefetchConditionResults`, the check-timeout line) key on the same field.

**Observability.** A tick whose budget is spent now logs one grep-able line, in
the shape of the ga-ocypq2 timed-out-check line:

```
gc: order dispatch: per-tick budget 2 spent; the rotation did not reach 3 more order(s) this tick (due-ness not evaluated): cooldown-2, cooldown-3, cooldown-4
```

## Deliberate deviations from the design, and why

- **The line names *unreached* candidates, not a due-list.** The design's draft
  string said "due order(s)". Settling due-ness for the unreached tail would cost
  a per-order last-run fallback query (#3201) and a per-order event-cursor read
  (ga-l7jdg) — exactly the reads the budget exists to avoid — so the line reports
  what the tick already knows for free: the orders that cleared the cheap gates
  and were never reached, and it says so in those words rather than asserting
  deferral. Capped at 8 names plus a count.
- **A trigger-env failure on a condition order stays budgeted.** The design
  listed "trigger env failure" among the logic the unbudgeted pass should run.
  That path writes an *error* tracking bead rather than a dispatch, and
  `TestOrderDispatchTriggerEnvFailuresRespectMaxDispatchesPerTick` exists to keep
  it rationed — a city with 40 misconfigured condition orders would otherwise
  write 40 failure beads per tick. The unbudgeted pass therefore selects on the
  *prefetched passing check* (`conditionResult != nil && Due`), which is the
  literal statement of "work is pending right now".

## Tests — RED → GREEN

New file `cmd/gc/order_dispatch_condition_budget_test.go`, written first.

RED, on the unmodified dispatcher:

```
=== RUN   TestDispatchFiresDueConditionOrderOutsideTheRotationBudget
    order_dispatch_condition_budget_test.go:80: condition order runs after a tick whose budget the sweeps spent = 0, want 1 — the budget starved a due condition order
--- FAIL: TestDispatchFiresDueConditionOrderOutsideTheRotationBudget (0.01s)
=== RUN   TestDispatchConditionOrderStillHonoursOpenTrackingGate
--- PASS: TestDispatchConditionOrderStillHonoursOpenTrackingGate (0.00s)
=== RUN   TestDispatchConditionOrderStillHonoursOpenWorkGate
--- PASS: TestDispatchConditionOrderStillHonoursOpenWorkGate (0.01s)
=== RUN   TestDispatchConditionOrderWithFailingCheckDoesNotFire
--- PASS: TestDispatchConditionOrderWithFailingCheckDoesNotFire (0.01s)
=== RUN   TestDispatchBudgetExhaustionLogsDeferredDueOrders
    order_dispatch_condition_budget_test.go:177: dispatch stderr = "", want a line reporting the spent per-tick budget
--- FAIL: TestDispatchBudgetExhaustionLogsDeferredDueOrders (0.00s)
=== RUN   TestDispatchLogsNoDeferralWhenEveryDueOrderFires
--- PASS: TestDispatchLogsNoDeferralWhenEveryDueOrderFires (0.00s)
FAIL	github.com/gastownhall/gascity/cmd/gc	4.313s
```

GREEN, with the fix (`-race`, alongside the three pre-existing budget tests):

```
--- PASS: TestDispatchFiresDueConditionOrderOutsideTheRotationBudget (0.03s)
--- PASS: TestDispatchConditionOrderStillHonoursOpenTrackingGate (0.00s)
--- PASS: TestDispatchConditionOrderStillHonoursOpenWorkGate (0.01s)
--- PASS: TestDispatchConditionOrderWithFailingCheckDoesNotFire (0.01s)
--- PASS: TestDispatchBudgetExhaustionLogsDeferredDueOrders (0.01s)
--- PASS: TestDispatchLogsNoDeferralWhenEveryDueOrderFires (0.01s)
--- PASS: TestOrderDispatchRespectsMaxDispatchesPerTick (0.02s)
--- PASS: TestOrderDispatchBudgetRotatesAcrossAlwaysDueOrders (0.03s)
--- PASS: TestOrderDispatchTriggerEnvFailuresRespectMaxDispatchesPerTick (0.05s)
ok  	github.com/gastownhall/gascity/cmd/gc	15.126s
```

`TestOrderDispatchBudgetRotatesAcrossAlwaysDueOrders` was rebuilt out of cooldown
orders. It used five always-due *condition* orders, which under this change all
fire on the first tick — it would still have passed while measuring nothing.
Mutation-checked: pinning `nextDispatchStart = 0` fails it with
`order-run:cooldown-2 did not dispatch under a rotating budget`.

## Gates

| command | result |
| --- | --- |
| `go build ./...` | exit 0 |
| `go vet ./...` | exit 0 |
| `gofmt -l cmd internal` | no output |
| `golangci-lint run ./cmd/gc/... ./internal/orders/... ./internal/config/...` | 0 issues |
| `make check-docs` | ok |
| `go test ./cmd/gc -run 'Dispatch\|Order\|Budget\|Gate' -count=1` | ok (54.8s) |
| `make test-fast-parallel` | exit 0 — "All fast jobs passed" |
| `.githooks/pre-commit` | ran clean on the staged change |
| `./scripts/ci/complexity.sh check` | exit 1 — **byte-identical to an untouched `origin/main` export** (see below) |

`complexity-check` is red on `origin/main` today, independently of this PR. Run
against a clean `git archive origin/main` export, it reports the same nine
findings this branch reports, and none of them are in this diff:
`reconcileSessionBeadsTracedWithNamedDemand` 473 (baseline 461), `resolveTemplate`
87/86, `buildPreparedStartWithWorkDirResolver` 56/55, `doPrimeWithHookFormatOpts`
51/49, `doOrderCheckWithStoresResolverScopedJSON` 45/44, `acceptDialogFromStream`
31/29, `cmdNudgePoll` 25/22, plus new `filterAssignedWorkBeadsForSessionWakeWithStores`
26 and `releaseConfirmedOrphanSessionWork` 20. This PR adds **zero** entries:
`dispatch` drops well below its baseline 47 and both extracted helpers land under
the threshold of 20.

## Docs

- `internal/config/config.go` `MaxDispatchesPerTick` now says the budget covers
  cooldown/cron/event and that condition orders dispatch outside it.
- `docs/reference/config.md` and the city schema regenerated with `make generate`
  (never hand-edited); `make check-docs` passes.
- CHANGELOG `[Unreleased] → Fixed` bullet naming the mechanism and the 11 h
  maintainer-city outage.

## Maintainer fixups (adoption review)

One maintainer commit rides on top of the contributor commit above. It has two
unrelated parts, neither of which changes the dispatch behavior described here.

- **CI repair, `internal/doctor/checks_custom_types_test.go`.** The custom-types
  doctor test read `bd` through `CombinedOutput`, so `bd`'s success-path stderr
  diagnostics contaminated the JSON its assertions parse and the test failed in
  pre-review CI. It now uses `Output()` (stdout only) and re-appends
  `ExitError.Stderr` on the failure path so failure reports keep their
  diagnostics. Unrelated to order dispatch; it rides here because the branch
  needed it to go green.
- **Review fixes — documentation and operator-facing wording only.** The
  `MaxDispatchesPerTick` doc comment no longer reassures that condition orders
  are "bounded by their own check plus the open-work gate". The open-tracking and
  open-work gates are keyed per order and only hold back a redispatch of an order
  whose previous run is still moving, so they bound no tick total; the comment now
  states the real bound — this budget plus one dispatch per condition order whose
  check passed on that tick — and the cold-start case where no tracking bead
  exists yet, with `docs/reference/config.md` and the city schema regenerated from
  it. The budget-exhaustion line was reworded to claim only what the tick knows
  for free, and `logDeferredCandidates` / `maxDeferredOrderNames` / the two tests
  that pinned the old wording were renamed to `logUnreachedCandidates` /
  `maxUnreachedOrderNames` / `TestDispatchBudgetExhaustionLogsUnreachedOrders` /
  `TestDispatchLogsNothingUnreachedWhenEveryDueOrderFires`, so the RED/GREEN
  transcripts above predate that rename. One stray blank line was removed from the
  CHANGELOG entry.

## Out of scope

- **Why the maintainer-city tick takes ~20 min** (reconciler / beads-cache cost)
  is a separate incident. This change makes condition orders independent of it.
- **No change to `max_dispatches_per_tick` in maintainer-city's `city.toml`.**
- One consequence worth naming: a city where many condition checks pass on the
  same tick now launches all of them, where it previously launched at most
  `max_dispatches_per_tick`. That is the intent — each is single-flighted by its
  own tracking bead and open-work gate, and each says real work is waiting — but
  it is a load-shape change on condition-heavy cities.

Refs ga-j4p3y, ga-unaz7

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **A condition-triggered order whose check passes now dispatches on the tick
+  that observes it, instead of queueing behind the per-tick dispatch budget.**
+  `orders.max_dispatches_per_tick` (default 4) capped every trigger type and
+  advanced a rotation cursor only when it spent a slot, so on a city with ~40
+  orders and a tick period longer than every sweep's interval the cooldown
+  sweeps were due at every tick and spent the whole budget before the rotation
+  reached anything else — each order got a turn roughly every ten ticks. A
+  condition order is a wake demand rather than a request for a turn: its check
+  reports that work is pending right now, and it is true only for as long as
+  that work waits. On maintainer-city the merge-queue order went 11 h without a
+  dispatch while merge beads sat ready, with no gate error, no suppression
+  event and nothing in the journal to show for it. Due condition orders now
+  fire outside the budget (still single-flighted by the open-tracking and
+  open-work gates, and still bounded by their own check), the budget and its
+  rotation are unchanged for cooldown, cron and event orders, and a tick that
+  spends its budget logs one line naming the orders it did not reach.
+
 - **A control bead served by a relocated class binding is routed to the
   dispatcher its own `gc.root_store_ref` names.** On a split city every rig's
   control beads live in one class binding. The reconciler dropped rig-rooted
--- a/cmd/gc/order_dispatch.go
+++ b/cmd/gc/order_dispatch.go
@@ -745,179 +745,287 @@ func (m *memoryOrderDispatcher) dispatch(ctx context.Context, cityPath string, n
 
 	m.prefetchConditionResults(candidates, now)
 
-	// Phase 2: the fire loop, in the same rotation order, over the same
-	// per-order state phase 1 resolved.
+	// Phase 2: the fire loop, in two passes over the same rotation order.
+	//
+	// A due condition order is not a sweep asking for its turn. Its check has
+	// just reported that work is pending right now, which is the order system's
+	// equivalent of a wake demand, and it already single-flights through the
+	// open-tracking and open-work gates below and self-limits by its own check.
+	// The budget exists to spread CLOCK-driven orders (cooldown/cron/event)
+	// across ticks and to avoid a cold-start burst; rationing a due condition
+	// order against sweeps defeats the trigger's whole purpose. On a city with
+	// ~40 orders and a tick period well past every sweep's interval, the sweeps
+	// are due at every tick and saturate the budget permanently, so a condition
+	// order whose condition is true for a window shorter than a full rotation of
+	// the cursor fires only by luck — maintainer-city's merge-queue order went
+	// 11h without a dispatch that way, with no gate error and no suppression
+	// event to show for it (ga-unaz7). Raising max_dispatches_per_tick only
+	// moves that cliff.
+	//
+	// Both passes run the identical per-candidate body; only the budget differs.
+	var unbudgeted, budgeted []*orderDispatchCandidate
 	for _, cand := range candidates {
-		idx := cand.idx
-		a := cand.order
-		target := cand.target
-		store := cand.store
-		storesForGate, storeKeysForGate := cand.gateStores, cand.gateStoreKeys
-		scoped := cand.scoped
-
-		baseLastRunFn := trackingIndex.lastRunFunc(storesForGate, storeKeysForGate, orders.LastRunAcross(orderFrontDoorsForStores(storesForGate)))
-		var lastRunErr error
-		var lastRunFromCache bool
-		lastRunFn := func(orderName string) (time.Time, error) {
-			last, fromCache, err := m.cachedLastRun(orderName, storeKeysForGate, baseLastRunFn)
-			if err != nil {
-				lastRunErr = err
-			}
-			if fromCache {
-				lastRunFromCache = true
-			}
-			return last, err
-		}
-		cursorFn := orders.CursorAcross(orderFrontDoorsForStores(storesForGate))
-		if a.Trigger == "event" {
-			cursor, err := bdCursorAcrossStores(a.ScopedName(), storesForGate...)
-			if err != nil {
-				logDispatchError(m.stderr, "gc: order dispatch: reading event cursor for %s: %v", a.ScopedName(), err)
-				continue
-			}
-			cursorFn = func(string) uint64 {
-				return cursor
-			}
-		}
-		triggerOpts := cand.triggerOpts
-		if err := cand.triggerErr; err != nil {
-			redacted := redactOrderEnvError(err, os.Environ())
-			msg := fmt.Sprintf("building trigger env: %s", redacted)
-			logDispatchError(m.stderr, "gc: order dispatch: building trigger env for %s: %s", a.ScopedName(), redacted)
-			// Leave this open so the existing open-work gate suppresses repeat
-			// ticks until the normal stale tracking sweep gives the order another try.
-			trackingBead, createErr := m.orderFrontDoorFor(store).CreateRun(scoped, orders.RunOpts{Outcome: orders.RunOutcomeTriggerEnvFailed})
-			if createErr != nil {
-				logDispatchError(m.stderr, "gc: order dispatch: creating trigger env failure tracking bead for %s: %v", scoped, createErr)
-			} else {
-				m.rememberLastRun(scoped, storeKeysForGate, trackingBead.CreatedAt)
-			}
-			m.rec.Record(events.Event{
-				Type:    events.OrderFailed,
-				Actor:   "controller",
-				Subject: a.ScopedName(),
-				Message: msg,
-			})
-			if spendDispatchBudget(idx) {
-				return
-			}
-			continue
-		}
-		// A condition order's verdict was already computed by the parallel pass;
-		// re-running it here would fork the check a second time.
-		var result orders.TriggerResult
-		if cand.conditionResult != nil {
-			result = *cand.conditionResult
+		if dueConditionCandidate(cand) {
+			unbudgeted = append(unbudgeted, cand)
 		} else {
-			result = orders.CheckTriggerWithOptions(a, now, lastRunFn, m.ep, cursorFn, triggerOpts)
+			budgeted = append(budgeted, cand)
 		}
-		if lastRunErr != nil {
-			logDispatchError(m.stderr, "gc: order dispatch: reading last run for %s: %v", a.ScopedName(), lastRunErr)
+	}
+	for _, cand := range unbudgeted {
+		m.fireCandidate(ctx, cand, trackingIndex, &inFlight, cityPath, now)
+	}
+	for i, cand := range budgeted {
+		if !m.fireCandidate(ctx, cand, trackingIndex, &inFlight, cityPath, now) {
 			continue
 		}
-		if !result.Due {
-			// The streak counts consecutive refusals of a DUE order, so an
-			// undue tick ends the episode. Without this a condition order that
-			// wedges and then goes false freezes its streak forever: the next
-			// refusal — possibly an unrelated incident weeks later — alerts on
-			// its first tick, carrying a first_suppressed and a
-			// suppressed_for_ms that span both episodes.
-			//
-			// Error and suspension exits deliberately do NOT clear. Those ticks
-			// never consulted the gate, so they are not evidence it opened, and
-			// resetting on them would let a flapping store hide a permanently
-			// shut gate.
-			m.clearOpenWorkSuppression(scoped)
-			// A condition check killed by its deadline never proves its
-			// condition, so the order silently never fires. Surface that
-			// distinctly (normal "condition false" is not logged) so a check
-			// outgrowing its budget is diagnosable instead of invisible
-			// (ga-ocypq2). Raise the order's check_timeout to fix.
-			if a.Trigger == "condition" && strings.Contains(result.Reason, orders.ConditionCheckTimedOutMarker) {
-				logDispatchError(m.stderr, "gc: order dispatch: %s %s — raise check_timeout if the check needs a slow store read", a.ScopedName(), result.Reason)
-			}
-			continue
-		}
-		if lastRunFromCache && orderTriggerUsesLastRun(a) {
-			refreshedLastRun, err := baseLastRunFn(a.ScopedName())
-			if err != nil {
-				logDispatchError(m.stderr, "gc: order dispatch: refreshing last run for %s: %v", a.ScopedName(), err)
-				continue
-			}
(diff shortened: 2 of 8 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-699284130204932915

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR fixes a real starvation bug in order dispatch: due condition-triggered orders now fire outside the per-tick rotation budget, so an order whose check passes dispatches on the tick that observes it instead of waiting behind a saturated cooldown/cron/event rotation. That matters because a passing condition check means work is pending *right now* — and when the rotation is full, "later" can mean never in practice. This is the mechanism behind the ~11 h maintainer-city stall where the condition-triggered merge-queue order never got a turn, and the CHANGELOG entry names it so the next person who hits the symptom can find the cause.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **The defect was independently re-derived, not taken on faith.** Both reviewers walked the pre-change code rather than accepting the PR's claim, and confirmed it: the old fire loop `return`s out on budget exhaustion, so a condition candidate sitting behind a saturated rotation never reaches `fireCandidate`. The two-pass split routes only *due* condition candidates around the budget, through the identical `fireCandidate` body.
- **The refactor was checked as a faithful extraction, path by path.** `fireCandidate` preserves every exit: event-cursor read error, trigger-env failure still recording `RunOutcomeTriggerEnvFailed`/`OrderFailed` and still costing budget, lastRun read error, un-due suppression-clear and condition-timeout logging, cached-lastRun refresh and re-check, the open-work gate including fail-closed and contention backoff, the launch-failure `inFlight.Done()` rebalance, and `rememberLastRun`. Ordering and semantics are unchanged.
- **The legacy `gate = "condition"` spelling was confirmed covered.** `dueConditionCandidate`'s `Trigger == "condition"` test is complete because both the scanner and the override layer fold the legacy spelling before the dispatcher sees the order.
- **Your own test-quality catch was validated.** The pre-existing rotation test was built from condition orders and would have gone vacuous under this change; rebuilding it from cooldown orders with a wall-clock anchor was the right call, and the new headline test is red on base by construction.

Review findings addressed (all in maintainer commit `62a83c9b06`):
- **Release safety / correctness (minor):** the `MaxDispatchesPerTick` doc comment reassured that condition orders are "bounded by their own check plus the open-work gate". That is true *per order* but bounds no tick total — the gates only hold back a redispatch of an order whose previous run is still moving. The comment now states the real bound (this budget plus one dispatch per condition order whose check passed on that tick) and calls out the cold-start case where no tracking bead exists yet, with `docs/reference/config.md` and both city-schema artifacts regenerated from it.
- **Debuggability (minor):** the budget-exhaustion log line asserted deferral and prescribed raising `orders.max_dispatches_per_tick` for a set whose due-ness the tick has not actually evaluated — visible in the PR's own test, where an order on a fresh cooldown was named as "deferred". It was reworded to claim only unreached-ness, and `logDeferredCandidates` / `maxDeferredOrderNames` plus the two tests that pinned the old wording were renamed to `logUnreachedCandidates` / `maxUnreachedOrderNames` / `TestDispatchBudgetExhaustionLogsUnreachedOrders` / `TestDispatchLogsNothingUnreachedWhenEveryDueOrderFires`, so code, log and tests agree again.
- **Release safety (nit):** removed a stray double blank line in the new CHANGELOG entry.

- `ga-tgyia` — **Release safety / correctness, minor:** the unbudgeted condition pass has no ceiling of its own. A tick's total launches move from `<= max_dispatches_per_tick` to `budget + (# condition orders due that tick)`, so an operator who set `max_dispatches_per_tick = 1` loses that lever for condition orders. This was raised at major and settled at minor on a census: the densest real city population is ~7-8 condition orders, so the worst realistic cold-start burst is ~12 concurrent launches versus 4 — a genuine envelope change, but not a runaway, and the upstream check pass is already capped at 8 concurrent forks. The behavior is this PR's deliberate, CHANGELOG-announced point; a dispatch-side ceiling mirroring `orderConditionCheckConcurrency` is the follow-up.
- `ga-62bof` — **Debuggability, minor:** the unreached-orders line is emitted through `logDispatchError` (log + stderr) on every saturated tick, with no throttle, dedup, or first-N guard of the kind the open-work suppression path has. Budget saturation is the steady state the budget is designed to produce, so this belongs on a non-error path or behind a changed-set/consecutive-tick threshold.
- `ga-62bof` — **Test evidence quality, minor:** the `maxUnreachedOrderNames` truncation branch and its `(+N more)` tail have no test. That branch only runs on the large-city configuration it was written for, so an off-by-one at the boundary would land silently on exactly that deployment.

One more finding — that an unrelated `internal/doctor` test fix was riding this dispatch PR — resolved itself during finalization; see below.

### Maintainer Changes
`62a83c9b06 chore(adoption): maintainer fixups from the PR-review loop` — 7 files changed, 38 insertions(+), 29 deletions(-). Documentation and operator-facing wording only; no dispatch behavior changed.

During finalization the branch was rebased onto current `main`, which required one conflict resolution worth naming: the maintainer fixup had also carried a CI repair to `internal/doctor/checks_custom_types_test.go` (reading `bd` through `Output()` instead of `CombinedOutput()`, so success-path stderr stops contaminating the JSON the assertions parse). While this PR was in review, #6480 landed that same fix upstream, plus a `TestParseCustomTypesJSON` case pinning that a stderr-prefixed document stays rejected rather than skipped. That hunk was therefore dropped rather than merged: this PR now has a zero delta on that file and takes the upstream fix, which is a strict superset. The PR description has been updated to match. Your contributor commit is untouched and the rest of the patch is byte-identical to what was reviewed.

### Final Review Status
Ready for the merge queue: final head `62a83c9b` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/35809477441

### Review Iterations
1 review pass, 0 fix iterations — the review approved on the first pass, with the documentation and log-wording fixes above landed as maintainer fixups rather than requiring another round trip.

---

PR #5991, opened 2026-09-04
Preserve case-sensitive hook route identities

quad341:
## What this changes

`gc hook` now keeps configured agent identities case-sensitive when it decides whether routed work belongs to the current agent. Session-safe spellings still decode `/` and `.` correctly, but identities such as `gascity/builder` and `gascity/Builder` no longer collapse onto one another and expose or claim the wrong agent's work.

The comparison stays localized in the route-identity matcher shared by hook display and claim paths. If a case-divergent route writer is ever found, normalization belongs where that route is written rather than in this read-side authorization boundary.

## Review notes

- Tmux-safe slash and dot encoding compatibility is preserved.
- Case-differing configured agents are explicitly covered by regression tests.
- No configuration format, defaults, or persisted-route migration changes.

## Test plan

- [x] Run the documented full local test matrix and account for every raw failure.
- [x] Run policy, build, and vet lanes.
- [x] Audit the case-sensitive and session-encoding route regressions.
- [x] Release gate: [`release-gates/ga-fx2uog-hook-route-case-sensitivity-gate.md`](release-gates/ga-fx2uog-hook-route-case-sensitivity-gate.md)


quad341:
@quad341 — thanks for this one, and specifically for putting the *reason* in the comment rather than just flipping `EqualFold` to `==`. Pointing at `ValidateAgents` keying on a case-sensitive `{dir, binding, name}` is what made this reviewable: I went straight to `config.go:4063`, confirmed `validAgentName` admits uppercase, and could see that `builder` and `Builder` really are two distinct agents. Keeping the legacy bound-template anti-collapse note intact was a nice touch too — easy thing to lose in a tightening like this. I also checked `internal/agent/session_name.go` has no case folding anywhere, so the dash/dot axis is genuinely unaffected.

Merging as-is. One thing for your notes, not a blocker: `qualifyPool` in `order_dispatch.go:3777` returns any slash-containing order-TOML `pool` verbatim without checking it against `cfg.Agents` — so a case-typo'd order pool *is* a case-divergent writer, today. After this lands it strands instead of being quietly absorbed by the differently-cased agent, which I think is the right failure mode for an authz boundary; `doctor`'s routed-to namespace check is where it should show up. Worth a follow-up bead to validate at that write seam, exactly as your comment prescribes.

Also cheap if you're back in that file: the removed cases covered the dash-encoded case-divergent spelling and the replacement only covers the canonical one.


---

PR #5995, opened 2026-09-04
fix(reconciler): bound a pool slot stuck in drain so its name is released

julianknutsen:
## Summary

A pool-managed session bead that enters drain and cannot finalize its drain-ack stays `status=open` **forever**, and an open bead owns its `session_name`. Because a pool slot's name is deliberately a pure function of its identity (ga-vcjr9), the pool cannot route around it: it mints **zero** seats, and every bead routed to that template becomes unclaimable.

Production: bead `gcg-5436965277275192` held `beads--gc__implementation-worker-1-pool` open for **3 days 10 hours**, which starved PR #6119's ready worktree step and stalled the beads v1.3.0 rc.2 release for three days. The affected city had 6–8 such beads at diagnosis. The existing drain-ack deadline releases the *obligation*, not the bead — nothing bounded the state.

**Fix:** for pool-managed beads only, once the drain is past `poolSlotDrainRetireDeadline` (30m) and unfinalized, with **no assigned work** and a runtime proven absent (or stopped and re-observed absent), force `ClosePatch(now, "drained")` with greppable provenance and a counted typed event. Four ordered gates, all fail-closed; the kill is token-fenced.

## Two corrections the implementation made to its own design

- **The stuck seat does not stay in a drain state.** The reconciler re-heals a drained bead with a live runtime back to `state=awake`, so a state-name test would have missed nearly every real instance; the durable shape is `sleep_reason=drained` + empty `last_woke_at`. Measured on mainline: the shape survives **15 ticks with zero stop attempts**.
- **`drain_at` is never cleared**, so its age alone proves nothing — a seat whose drain was *cancelled* keeps an old `drain_at` while working normally. `sessionInUnfinalizedDrain` separates them on two independent markers, with a dedicated pin for that false-positive class.

## Review

Adversarial review returned 8 findings (4 major), all fixed. The decisive one: gate 3 originally resolved work-ownership through a narrow identifier set that omits `alias`/`alias_history`, while agents claim under their alias and `AssigneeIdentifier` resolves **alias-first** — and alias diverges from `session_name` exactly on the `-pool` step-aside shape, i.e. this bug's own specimen. A reviewer probe confirmed HEAD killed a live runtime and stranded its claim where base did not; gate 3 now uses the canonical `session.AssigneeIdentities` set. Also fixed: a leaked worktree per retirement, a close that skipped the partial-store and boot-deferral gates, overriding `held_until`/`quarantined_until`/`wait_hold`/`pin_awake`, a kill that was not re-fenced against work claimed mid-sequence, provenance that could outlive a refused close, and a stop invisible to the stop event and counter.

One finding was answered with evidence rather than applied: the `state=draining` arm was dead code, but relocating the gate would have bought **zero** retirements (that population converges in 3 ticks on its own) and reintroduced the shape §3.5 rejects — closing over a live agent. The dead arm was deleted and the boundary pinned instead.

## Testing

31 pin assertions plus the ga-vcjr9 identity-race control; **9 mutation probes, each killing its own pin**. Every behavioral pin was watched failing first. `golangci-lint` 0 issues; `make spec-ci` exit 0 with no wire drift; `-race` clean (an initial race was the test's own async-drain writer, fixed with a guarded writer). One pre-existing failure baselined against clean `origin/main`.

Refs: ga-rxhu2

---
- Related to #2695 — adds an automated reaper for drained session beads that otherwise linger open forever, but only for pool-managed slots — the reconciler force-retires one 30 minutes past an unfinalized drain; it does not add the configurable witness-patrol TTL that issue asks for, so non-pool drained sessions still accumulate
- Related to #4448 — addresses the drained-pool-slot half of that report: a pool-managed session bead stuck in an unfinalized drain past 30 minutes, with no assigned work and its runtime proven gone, is now force-retired so the pool can mint the seat again. It does not cover stale open session beads that carry no drain provenance, and the slot is still held until that deadline elapses.


quad341:

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277298129

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-699284130204933672

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR bounds a pool slot that gets
stuck in `draining`: when a pool-managed session sits past the drain deadline, the
reconciler now kills the runtime and retires the slot so the session *name* is
released back to the pool. That matters because a stuck drain used to squat its
slot name indefinitely — the pool quietly loses a seat and the template never
refills it, with no operator-visible bound on how long that lasts.

branch.

### Original PR Review
Decision: approve (resolved from request_changes after three fix iterations).

Specific gaps fixed:
- **Gate 1 did not enforce the identity exclusion its own comment promised.**
  `isPoolManagedSessionInfo` is satisfied by `session_origin=ephemeral` *alone*,
  and that is exactly the shape `isLegacyManualSessionInfoForAgent` defines as a
  legacy user-created seat. Unlike the sibling pool-freeable close, which disposes
  of an already-dead runtime, this path answers with a Kill of a live one. The
  final patch adds `poolSlotRetireOwnsSeat`, pairing the pool predicate with
  `!isNamedSessionInfo` and `!isManualSessionInfoForAgent`.
- **The refusal keyed on `cfgAgent == nil`, which was too narrow.** A legacy seat
  under a still-configured single-session agent, and a dependency-only seat, both
  stayed killable. The guard was widened to the capability itself
  (`!cfgAgent.SupportsMultipleSessions()`, nil-safe) plus a separate
  `|| info.DependencyOnly` term — disjunctive, so neither arm relies on
  short-circuit ordering.
- **The gate ran above the reconciler's `!isKnownStateInfo` forward-compat skip**
  while matching on `sleep_reason` alone, so a rolled-back reconciler could kill
  and close a bead in a state it does not recognize — precisely the population
  that skip exists to leave alone. The guard now enforces the allowlist itself,
  which also makes the file's "`state=draining` is deliberately NOT matched" claim
  structural rather than an argument from reachability.
- **A kill could be lost to the stop counter and `session.stopped`** whenever the
  final work fence, the provenance stamp, or the close refused afterwards: the
  next tick found the runtime already absent and never emitted. `RecordAgentStop`
  and `SessionStopped` now fire the moment the kill is re-observed; only
  `SessionPoolSlotRetiredAtDrainDeadline` stays coupled to the close.
- **Two load-bearing comments were factually wrong after the widening.** The
  attribution comment presented a closed enumeration ("exactly three bail-outs")
  that a raw-empty `session_origin` falsifies — there are four live bail-outs, and
  the raw-origin one fires *first*, so it (not the final clause the comment named)
  is what excludes this file's own dependency-only fixture. Separately,
  `poolSlotRetireTemplate`'s doc claimed a seat-safety stake the fallback no longer
  has: after the guard widening such a seat is refused either way, and what the
  fallback still buys is event attribution. Both prose sites were retargeted, and
  the `template_from_the_seats_own_projection` test arm — which the widening had
  silently drained of discriminating power — now carries an assertion only the
  fallback can satisfy.

Review findings addressed:
- All of the above landed in the single maintainer fixup commit `704daab0db`,
  which records per-iteration attribution for each finding in its message. Every
  new pin was red/green-verified with a `go test -overlay` mutation probe against
  an unmutated worktree, with the feature-off positive control
  (`...RetiresMarkedSeatWhoseAgentLeftConfig`) green in the same runs — so none of
  the new assertions are vacuous.

- `ga-68oxc` — [nit] Debuggability & Operability: gate-1 drain-deadline retire
  refusals are silent. Every ownership refusal returns `false` with no log line,
  and this change widens the refused population, so an operator whose stuck seat
  is never freed gets no signal for *why* the bound declined it. Deliberately not
  fixed here: it is a pure predicate on every session on every reconcile tick, so
  an unthrottled log line is the wrong shape, and the signal belongs beside the
  gate's other observability rather than inside the predicate. Both reviewers
  agreed it cannot starve a template to zero seats, since a refused seat is by
  construction not a marker-bearing pool seat. This PR landed the documentation
  half (the gate-1 doc now says refusals are deliberately silent and points at
  `pool_managed`/`pool_slot` on the bead); `ga-68oxc` tracks the remaining half.

A note on the adoption rebase, since it touched your commits: the branch was
replayed onto current `main`. Generated trees (`internal/api/dashboardspa/dist`,
the typed supervisor client, the OpenAPI schemas) were **regenerated from merged
source** rather than side-selected, and reproduce cleanly. One of your commits,
`5d5bb2835a` "test: keep bd stderr out of custom-types JSON assertions", was
dropped because #6480 landed the same `CombinedOutput()` → `Output()` fix upstream
as a strict superset — it adds the explanatory comment, surfaces `exitErr.Stderr`
through `mustRunBD`, and adds a table case guarding the regression. Your intent is
fully realized on the new base; keeping the narrower variant on top would have
been a hand-merged hybrid of two solutions to one problem.

### Maintainer Changes
`704daab0db fix(reconciler): gate the drain-deadline retire on identity and known state`
— 3 files changed, 551 insertions(+), 26 deletions(-)
(`cmd/gc/session_pool_drain_deadline.go`,
`cmd/gc/session_pool_drain_deadline_test.go`, `internal/session/manager.go`).
Your commits are preserved with original authorship.

### Final Review Status
Ready for the merge queue: final head `704daab0` has passing required GitHub
checks (91 passing, 13 skipped, none failing or pending), and the merge-ready
metadata is in place. The branch was also verified against the latest `main`
locally — the rebase is content-identical (stable patch id, zero file overlap with
the intervening base commit), and the resource-census ratchet, `go vet`, and the
focused reconciler/session/events suites are green on that refreshed head. I am
marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/35866817207

### Review Iterations
4 review passes. Iteration 1 raised the identity major plus five minors (the
forward-compat unknown-state skip, the lost kill, the template fallback, the
unpinned constant ordering, and a comment naming the wrong writer set).
Iteration 2 asked the gate to fail closed on an unattributable identity, but
narrowly. Iteration 3 widened that refusal from `cfgAgent == nil` to the
multi-session capability and added the dependency-only term. Iteration 4
validated those fixes with independent mutation probes and returned approve; its
two remaining prose corrections were applied before merge.

---

PR #5997, opened 2026-09-04
feat(beads): default fresh Dolt init to proxied mode

julianknutsen:
## Summary
- thread provider-neutral `transport`/`target` init intent through `gc init`
- default fresh bd/Dolt initialization to bd 1.3 proxied-local
- retain direct-local, direct-external, and proxied-external escape hatches
- delegate topology translation to the bd adapter with fail-closed validation

## Verification
- `go test ./cmd/gc -count=1`
- `go test ./internal/config -count=1`
- focused init-intent and adapter tests
- `sh -n examples/bd/assets/scripts/gc-beads-bd.sh`
- `git diff --check`
- pre-commit golangci-lint and go vet

Beads: ga-fqzck, ga-fqzck.1


julianknutsen:
Superseded by #5999, which carries the corrected exact implementation head and is the canonical Sol-review target.

PR #5998, opened 2026-09-04
docs: clarify Beads migration journals and guarded rollback

julianknutsen:
Restores the migration runbook on the active fresh-default integration branch. It explains that mode-migration journals are absent before migration and removed after commit, while a valid journal left by a fault is the resume record.

The commands are scoped to stopped, bd-owned workspaces. Legacy GC-owned stores require the separate ownership handoff first. Rollback uses the matching guarded reverse migration; external endpoints remain outside in-place migration support.

Validation: source behavior checked against Beads mode-migration implementation and retry matrices; `make check-docs`, commit hook, and diff checks passed. Sol and independent Astra approved `34d35b4a8`.

### DIFF
--- a/engdocs/contributors/beads-proxy-migration.md
+++ b/engdocs/contributors/beads-proxy-migration.md
@@ -10,7 +10,20 @@ one, and use `--dry-run` first.
 
 `bd dolt stop` stops a bd-owned Dolt process; it is not a `gc stop` substitute.
 For a legacy GC-owned workspace, complete the explicit ownership handoff in
-`ga-p9iuv.30.1` before using a bd migration command.
+`ga-p9iuv.30.1` (currently blocked) before using a bd migration command.
+
+The journaled contract described below — the `.beads/dolt-mode-migration.json`
+phases and the fail-closed checks built on them — ships with the beads
+`release/1.3.0` line, which is not yet merged to beads `main`. Builds without
+it, including the beads module this repo currently pins in `go.mod`
+(`v1.1.1-0.20260805093327`) and the installed `8de373f05`, perform the same
+mode flip as a single unjournaled `metadata.json` plus sidecar rewrite (plus
+the shared-YAML write for shared roots), protected by an exclusive migrate lock
+(a second concurrent `bd migrate` is refused) and by idempotent re-invocation
+(re-running a finished migration reports the mode is already set and exits
+successfully). Confirm your bd build's version floor before relying on
+journal-based fault recovery. bd marks all four `bd migrate` subcommands
+`[EXPERIMENTAL]` on both builds.
 
 ### Server and proxied-server
 
@@ -57,6 +70,12 @@ bd migrate from-shared-server-to-proxied-server
 bd dolt stop
 ```
 
+On a shared root, `bd dolt stop` is not project-local. The first stop above
+stops the shared Dolt server for every project sharing it — bd attaches that
+same warning to `bd migrate from-shared-server-to-proxied-server` itself — and
+the trailing stop targets a proxy keyed by that same shared root. Coordinate
+with the other projects on the root before either stop.
+
 To roll back a completed shared-server-to-proxied migration, stop the bd-owned
 process and run the matching reverse command:
 
@@ -65,6 +84,9 @@ bd dolt stop
 bd migrate from-proxied-server-to-shared-server --dry-run
 bd migrate from-proxied-server-to-shared-server
 ```
+
+This `bd dolt stop` has the same blast radius: it targets the proxy rooted at
+the shared Dolt directory, so it affects every project proxied at that root.
 
 Managed-local mode owns the proxy and child Dolt lifecycle. External TCP or
 Unix endpoints are owner-managed; in-place migration refuses them. A migration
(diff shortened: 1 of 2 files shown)


julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR restores
`engdocs/contributors/beads-proxy-migration.md`, the runbook for flipping a
Beads workspace between direct and proxied Dolt modes with `bd migrate` — the
five-phase `.beads/dolt-mode-migration.json` journal, the stray-sidecar
refusal, the "server or proxy is still running" refusal, and the guarded
rollback path. That matters because this file is the tree's *only* description
of those migration commands: without it, an operator flipping `dolt_mode` on a
live workspace has no documented checkpointing or recovery story, and no
written rollback procedure to fall back on.

PR branch.

### Original PR Review
Decision: approve.

Worth calling out first: the review lanes initially reached directly
contradictory conclusions about whether the five-phase journal exists at all,
because two of them probed only the deployed binary and the `go.mod` pin. A
first-hand probe settled it — the journal genuinely ships on the beads
`release/1.3.0` line (`cmd/bd/migrate_dolt_mode.go`, all five phase names
present), which is 35 commits ahead of `main` and not yet merged. Your journal,
stray-sidecar, and proxy-refusal passages were therefore **correct as written**,
and two proposed "corrections" to them were rejected as falsified rather than
applied. Those passages are byte-unchanged in the final patch.

Specific gaps fixed:
- **No bd version floor for the journaled contract.** The runbook described the
  journaled, resumable contract without naming the build that implements it, so
  an operator on the pinned `v1.1.1-0.20260805093327` or the installed
  `8de373f05` would expect checkpointing they do not have. The final patch adds
  a version-floor paragraph naming the implementing `release/1.3.0` lineage,
  both older builds, the unjournaled fallback they actually perform (an
  exclusive migrate lock that refuses a second concurrent `bd migrate`, plus
  idempotent re-invocation that exits 0), and the `[EXPERIMENTAL]` status of all
  four subcommands. Each of those clauses was verified against both builds.
- **`bd dolt stop` blast radius not propagated.** Both shared-server flows told
  the operator to stop the server without repeating bd's own warning that this
  stops it for *every project sharing it*. The final patch annotates both the
  forward and reverse flows, and correctly distinguishes the two stop targets —
  the shared *server* versus the shared-root-keyed *proxy*.
- **Runbook orphaned from every index.** The new file had no inbound link from
  anywhere in the tree. `engdocs/contributors/index.md` now carries an entry in
  the established style, adjacent to the sibling beads/store entry.
- **Blocked prerequisite stated as executable.** `ga-p9iuv.30.1` was written as
  a runnable step although the bead is blocked; it now reads "(currently
  blocked)".

Review findings addressed:
- `b1f41add58` — bd version floor for the journaled contract (Contract &
  Interface Fidelity, major): version-floor paragraph added and clause-verified
  against both the implementing and the pinned builds.
- `b1f41add58` — shared-server stop blast radius (Contract & Interface
  Fidelity, major): warning propagated on both shared-server flows, with the
  stop targets distinguished.
- `b1f41add58` — runbook orphaned from every index (Architectural Consistency,
  nit): inbound index link added; it is the runbook's single inbound link.
- `b1f41add58` — blocked prerequisite stated as executable (Contract &
  Interface Fidelity, nit): status annotated.

- `ga-ikom4`: Contract & Interface Fidelity, nit — the old-build fallback
  clause at `beads-proxy-migration.md:19-20` names two written files
  ("`metadata.json` plus sidecar rewrite") where the *shared-root* flip on the
  pinned old build also rewrites `.beads/config.yaml:dolt.shared-server`, i.e.
  three files. This was deliberately not changed here: the phrase "the same
  mode flip" is anaphoric to the document's own full enumeration at `:39-42`
  (which does list the shared-YAML write, restated again at `:102-103`), bd's
  own help text uses the identical two-file abbreviation for the plain variant
  and reserves the three-file form for the shared variants, and the YAML write
  is stated twice elsewhere in the same document — so a reader of the section
  cannot be misled. Filed for a future editor as a one-clause precision if the
  file is touched again.

### Maintainer Changes
One maintainer commit pushed to this branch:
`b1f41add58 docs: pin bd version floor and shared-stop blast radius in
migration runbook` — 2 files changed, 27 insertions(+), 1 deletion(-)
(`engdocs/contributors/beads-proxy-migration.md`,
`engdocs/contributors/index.md`). Docs-only; your original commit
`34d35b4a8e` is preserved unmodified underneath it.

### Final Review Status
Ready for the merge queue: final head `b1f41add` has a fully green GitHub
check surface — every check that ran passed, including `CI / required`,
`CI / preflight`, `CI / integration`, `Preflight / static checks`,
`Preflight / generated artifacts`, `Preflight / acceptance A`, `Dashboard SPA`,
`Contract / bd CLI (minimum supported)`, and
`Evidence / critical-path suites`, with no failing or pending checks — and the
merge-ready metadata is in place. I am marking this PR `status/merge-ready` so

CI: https://github.com/gastownhall/gascity/actions/runs/34310574247

### Review Iterations
2 review passes performed. Iteration 1 requested changes on two major
operator-safety caveats plus two nits; iteration 2 re-probed each remedy
against the code it describes on both bd builds, confirmed all four landed and
that the de-gated passages were left untouched, and approved with a single
non-gating nit (filed above as `ga-ikom4`).

---

julianknutsen:
Audit disposition: hold out of adopt-pr. These migration/rollback docs depend on the proxied-default rollout. They also need their retired ownership-handoff references reconciled with the explicit migration design.

PR #6000, opened 2026-09-04
feat(beads): default managed scopes to proxied Dolt

julianknutsen:
Flips the default Dolt mode for **freshly created managed scopes** to Beads'
`proxied-server` path, adds an external Unix-socket front door for proxied
Dolt, and moves the `bd` pin to `v1.3.0-rc.1`. The original scope of this
branch — deterministic external Unix-socket front-door outage/recovery
coverage plus a direct/proxy parity matrix that asserts no managed Dolt
runtime is created — is retained.

Retitled from `test(beads): harden external proxy topology outage/reconnect`:
the branch carries production-path changes that a `test(...)` scope hides from
reviewers and from changelog/release-note tooling.

### 1. Proxied-server is the default for new managed scopes

`desiredCityDoltConfigState` / `desiredRigDoltConfigState` now select
`dolt_mode=proxied-server` for a fresh managed scope. An explicit external
host/port endpoint still selects direct `server` mode.

**Existing scopes are not migrated.** `resolveDesired*EndpointState` resolves
persisted authoritative scope state before these constructors run, so an
already-initialized workspace keeps the mode it recorded
(`TestResolveDesiredCityEndpointStatePreservesAuthoritativeDoltMode`).
Converting an existing scope is an explicit, journalled operator step — see
`engdocs/contributors/beads-proxy-migration.md`.

### 2. The `bd` pin moves to the `v1.3.0-rc.1` prerelease

`deps.env:BD_VERSION`, the `go.mod` beads library pin, and
`contrib/k8s/Dockerfile.agent:BD_VERSION` all move `v1.1.0` → `v1.3.0-rc.1`.

GitHub marks `v1.3.0-rc.1` a **prerelease**. Pointing the installable default
at a prerelease is a deliberate exception to the published-release bar this pin
otherwise holds, taken because the RC is the first main-cut tag carrying the
per-subcommand flags this branch depends on. Move `BD_VERSION` to the first
non-prerelease release cut from beads `main` once one exists. The caveat is
recorded at `deps.env:29-33`.

### 3. Migration ordering requirement (v53 → v66 schema)

`BD_VERSION` and the `go.mod` library pin **must move together**. gc both links
beads as a library and shells out to `bd`, so a lagging half locks the other
out: after the v53 → v66 schema migration an older co-resident `bd` refuses the
store under the forward schema-skew guard.

Ordering for a shared, no-remote Dolt sql-server (this repo's shape):

1. Upgrade **every** `bd` client to `v1.3.0-rc.1` first. bd >= 1.3.0
   deliberately refuses to auto-apply pending migrations here, because
   migrating would lock out every co-resident client still on the old schema.
2. Only then run `bd migrate schema` **once**, from a designated migrator.
3. Do not work around a pending-migration refusal with `bd dolt pull` or an
   ad-hoc `BD_ALLOW_REMOTE_MIGRATE=1`.

This is the axis a `test(...)` title would not have prompted anyone to plan.

### 4. Unix-socket front door for external proxied Dolt

Dolt connection targets are socket-aware end to end: an external socket
endpoint is carried through config resolution and projected into child `bd`
processes, and the managed-Dolt socket handle is pooled.

## Test coverage

Deterministic external Unix-socket front-door outage/recovery coverage,
assertions that an external socket creates no managed Dolt runtime, and a
direct-vs-proxied front-door parity matrix across `proxied-managed`,
`proxied-external-host`, and `proxied-external-socket`.

---
- Related to #5348 — moves the bd CLI pin and the linked beads library pin off v1.1.0 to v1.3.0-rc.1 together and documents the required v53→v66 migration ordering, which removes the specific pinned-version contradiction described in that issue; it does not add a pre-bootstrap schema-compatibility check and does not stop managed scopes from picking up an ambient bd from PATH, so the underlying skew path is still open


julianknutsen:
Preflight could not proceed because this branch has drifted far beyond the PR story. The PR describes external Unix-socket beads proxy outage/recovery tests, but the current tip contains 39 commits touching 198 files (9,434 additions / 722 deletions), including unrelated dependency bumps, runtime/session changes, docs, workflows, and other fixes. Please retarget this PR to the intended focused commit set (or explain the broader scope) before adoption review can continue.

quad341:

julianknutsen:
Audit disposition: hold out of adopt-pr. This older proxied-default branch overlaps #6273 but contains distinct regression coverage; it is retained until that coverage is accounted for in the successor.

PR #6002, opened 2026-09-04
test(beads-pack): guard ambient Dolt socket clearing

julianknutsen:
Protects the proxy-init and DoltLite helpers against an inherited BEADS_DOLT_SERVER_SOCKET overriding their explicit backend selection. The test requires an unset statement before a recognized bd invocation and rejects reassignment.

Recovered the stronger existing regression guard onto the active #6011 dependency. Validation: focused race test, an actual shell invocation with a fake BD_BIN proving the variable is unset in both helpers, repository pre-commit lint/vet/codegen, and all pre-push fast shards. Sol and independent Astra approved fc025ad465738dba40ab3036671defed9e1e3fab.

### DIFF
--- a/cmd/gc/gc_beads_bd_lint_test.go
+++ b/cmd/gc/gc_beads_bd_lint_test.go
@@ -272,6 +272,80 @@ func TestDoltliteMaintenanceGatesStampOnReindex(t *testing.T) {
 	}
 }
 
+// ambientSocketUnsetPattern requires BEADS_DOLT_SERVER_SOCKET to appear in an
+// `unset` statement rather than merely somewhere in the helper body. Anchoring
+// on the statement form is what makes this a scrub assertion instead of a
+// spelling assertion: a bare substring match is equally satisfied by a line
+// that *sets* the socket, which is the exact regression being guarded.
+var ambientSocketUnsetPattern = regexp.MustCompile(`(?m)^\s*unset\b[^\n]*\bBEADS_DOLT_SERVER_SOCKET\b`)
+
+// ambientSocketAssignPattern is the inverse guard: assigning the socket inside
+// these helpers re-establishes the ambient topology the unset exists to drop.
+var ambientSocketAssignPattern = regexp.MustCompile(`(?m)^\s*(export\b[^\n]*\s)?BEADS_DOLT_SERVER_SOCKET=`)
+
+// ambientSocketBdInvocations lists the call-site spellings the ordering check
+// recognizes. Each carries the argument expansion so an anchor cannot land on
+// `bd_bin="${BD_BIN:-bd}"`, which only computes the binary name and can be
+// hoisted without moving either the scrub or the invocation.
+var ambientSocketBdInvocations = []string{`"$bd_bin" "$@"`, `"${BD_BIN:-bd}" "$@"`}
+
+// TestBeadsHelpersClearAmbientSocket proves the proxied and DoltLite helpers
+// scrub BEADS_DOLT_SERVER_SOCKET before invoking bd. An inherited socket must
+// not override the explicit proxied or DoltLite selection: gc's own env
+// projection (cmd/gc/bd_env.go) drops host/port whenever a socket is present,
+// so a leaked socket silently redirects an explicitly selected topology.
+//
+// run_bd_pinned is deliberately not asserted: it does not scrub the socket
+// today, and whether it should depends on beads' own connection-resolution
+// order, which is tracked separately (ga-kbcwc).
+func TestBeadsHelpersClearAmbientSocket(t *testing.T) {
+	root := repoRootForLint(t)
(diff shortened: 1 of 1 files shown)


julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR adds a build-time guard that
the `gc-beads-bd.sh` helpers scrub `BEADS_DOLT_SERVER_SOCKET` before they invoke
`bd`, which closes a genuinely nasty silent-failure mode: because gc's env
projection drops host/port whenever a socket is present, a leaked ambient socket
quietly redirects an explicitly selected proxied/DoltLite topology to somewhere
else entirely. Turning that into a failing test instead of a confusing runtime
topology bug is exactly the right place to catch it.

branch.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- The guard's original bare-substring match was equally satisfied by a line that
  *sets* the socket — verbatim the failure the doc comment claimed to prevent. The
  final patch anchors on the `unset` statement form and additionally rejects any
  `BEADS_DOLT_SERVER_SOCKET=` assignment, so a socket-honoring rewrite now fails.
- A missing bd-invocation anchor used to skip silently, so normalizing a call site
  to plain `bd "$@"` could opt the ordering check out with no signal. It is now
  fail-closed and reported per helper.
- The anchors now carry the argument expansion, so the inert `bd_bin="${BD_BIN:-bd}"`
  hoist no longer reports a scrub-ordering violation that did not happen.
- The guard moved beside its seven sibling `gc-beads-bd.sh` lint tests and adopted
  their harness conventions (`repoRootForLint`, `extractShellFunction`) instead of a
  hand-rolled, fail-open extractor.

Review findings addressed:
- **Test Evidence Quality — the commit message's stated mutation-probe method was
  inert** (minor; raised independently by two reviewer lanes). The message claimed
  the mutants were probed via `go test -overlay`, but this guard and its siblings
  read the script at run time through `os.ReadFile`, which an overlay does not
  intercept — every mutant would have read as the unmutated baseline. Fixed in
  `41c986b7`: the message now documents probing inside a disposable
  `git worktree add --detach` checkout and spells out why `-overlay` cannot work
  here, so the next maintainer who copies the recipe does not get a silent
  false-green. The guard's behavior itself was never in question — reviewers
  reproduced all of the claimed FAIL/PASS transitions using working methods.
- **Test Evidence Quality — the guard graded the first definition while shell
  executes the last** (minor). bash binds the last definition of a function name,
  so a duplicated `run_bd_init_proxied`/`run_bd_doltlite` appended without the
  scrub would have left the guard green while the executed copy leaked the socket.
  Latent rather than live — every bd-invoking helper has exactly one definition
  today — but the same drift mode already has in-repo precedent. Fixed in
  `41c986b7`: the helper loop now asserts
  `countShellFunctionDefinitions(script, name) == 1` before extraction, mirroring
  the existing pattern at `gc_beads_bd_lint_test.go:124`.

- **Test Evidence Quality — the shared review worktree transiently carried a
  reviewer lane's in-place mutation probe** (minor). This one is not about your
  diff at all: a reviewer lane mutated the script in place in the shared review
  worktree, which two other lanes observed live. It has no PR-side remedy, so the
  diff was deliberately left untouched for it. Filed as `ga-qslcv` to harden the
  review pipeline (assert a clean reviewed head before fanning out reviewer legs,
  and require mutation probes to run in throwaway checkouts).
- Not a finding, noted for context: `run_bd_pinned`'s missing scrub and `op_init`'s
  inline `bd context` witness remain unasserted by design. Both are pre-existing
  and blame-confirmed, and both are already tracked in `ga-kbcwc` — the call-site
  census is complete (four functions invoke bd; two asserted, two tracked).

### Maintainer Changes
Maintainer review fixes were amended into your commit on the PR branch, preserving
your authorship: `test(beads): harden and relocate the ambient-socket guard`
(1 file changed, 8 insertions(+) over the original head) — the definition-count
assertion for Finding 2, plus the corrected mutation-probe method description in
the commit message for Finding 1.

### Final Review Status
Ready for the merge queue: final head `41c986b7` has passing required GitHub

CI: https://github.com/gastownhall/gascity/actions/runs/34315273754

### Review Iterations
1 review pass performed — three independent reviewer lanes (reasoning-heavy,
evidence-first, and cross-file consistency) plus synthesis and a quality
scorecard, followed by one fix amend applied to the PR branch. No blocker or
major finding was raised by any lane; the three findings were all minor and all
landed in test-evidence quality, and re-running
`go test ./cmd/gc/ -run TestBeadsHelpersClearAmbientSocket -count=1` and
`go vet ./cmd/gc/` after the amend is clean.

---

julianknutsen:
Audit disposition: hold out of adopt-pr. This test-only branch depends on the older proxied adapter helpers, which are absent from current main. Its ambient-socket isolation coverage should survive consolidation into the replacement.

PR #6007, opened 2026-09-04
fix(beads): consult native-open deferral before dialing the identity probe (#5965)

### LINKED ISSUE #5965: preflight: the identity_match project_id probe dials an External Dolt endpoint with only the ambient GC_DOLT_PASSWORD, so every credential-less in-process open of an external rig store emits one rejected authentication at the hub
### Gas City version

`main` @ 1154ad128 (code at the lines below is identical on our fork build, which is where it was measured — see "Confirmed on main" at the end).

### Environment

Three cities sharing one federated Dolt hub (`dolt sql-server` on a tailnet host, plain listener). Each city registers a rig whose `.beads/config.yaml` declares the hub as its endpoint (`dolt.mode: server`, `dolt.host`, `dolt.port`, `dolt.user`, `dolt.remote-password-file`, `gc.endpoint_origin: explicit`). The password for that user is also in `~/.config/beads/credentials` under `[host:port]`, and the two values are byte-identical (compared by sha256). macOS arm64 (two laptops) and Linux amd64 (the hub host's own city).

### Symptom

The hub's journal shows a steady stream of `Error authenticating user <rig-user>` lines, one per in-process open of that rig's store by any gc process that does not carry `GC_DOLT_PASSWORD` in its ambient environment: supervisor-spawned exec orders (`order-tracking-sweep`, every minute), crew shells, and CLI commands such as `gc order history`. Nothing fails on the client: no WARN, no fallback to `bd`, the command succeeds. The only client-side trace is that the external scope takes ~3.5 s to open where local scopes take ~0.2 s.

### Reproduction

On a city with such a rig, from a shell with the credential stripped:

```sh
env -u GC_DOLT_PASSWORD -u BEADS_DOLT_PASSWORD GC_DOLT_USER= GC_DOLT_HOST=127.0.0.1 GC_DOLT_PORT=<managed-local-port> \
  gc order sweep-tracking --stale-after 10m --dry-run
```

and on the hub host, dump the journal once for the window and match by string with a +3 s tolerance (the rejection lands 1–3 s after the run starts; a `--since X --until X` with equal timestamps is a zero-width window):

```sh
journalctl -u <dolt-hub-unit> --since '-2min' -o short-iso | grep 'authenticating user'
```

Deterministic: one rejection per run (5/5 hand runs on one laptop; 15 → 0 rejections per 10-minute bucket on the hub after disabling the one-minute order on that laptop; the hub host's own city shows the same pairs; the third city's supervisor-spawned `gc order sweep-tracking` at its ~78 s cadence is its stream). With `GC_DOLT_PASSWORD` set to the correct value in the ambient env the rejection disappears; with a wrong value both the probe and the open fail and the WARN appears.

### Root cause

`cmd/gc/beads_preflight_checker.go:63-79` — `preflightDatabaseProjectIDReader` resolves the scope target (hub host/port and the rig's configured user) and calls `managedDoltOpenDatabase(target.Host, target.Port, target.User, target.Database)` (`cmd/gc/dolt_project_id.go:497`), which dials with `managedDoltPassword()` — `cmd/gc/dolt_sql_health.go:277-278`, i.e. `strings.TrimSpace(os.Getenv("GC_DOLT_PASSWORD"))` and nothing else. The probe never consults the credential that the store open itself resolves a moment later (the `doltauth` ladder ending in the credentials file, per #1720 / #1867). With the ambient variable empty the hub rejects the user; the gate returns unconfirmed; `preflightIdentityDeferredReader` (`:46-58`) sees `target.External` and defers to the native open, which authenticates fine. The comment above that reader already states the situation: *"The direct root/plaintext project_id probe cannot authenticate such endpoints"* — but the probe is still executed against them before the deferral is consulted.

Instrumented with `GC_BD_TRACE_JSON` plus a process sampler: the only `bd` subprocesses during the run are `bd context --json` (one per scope, from `preflightBDContextReader`); no store fallback runs; direct `bd context --json` invocations under the same env shapes produce no rejection.

### Impact

Log noise on the shared hub at the sum of every city's open rate (it masked a real credential problem for a day), one extra failed TCP+auth round trip per external-scope open, and a ~3.5 s latency tax on every command that opens such a scope.

On the hub host's own city the volume is not cosmetic: 80–84% of its syslog lines are the resulting connection/auth-error/close triples, and weekly syslog generations grew from ~10 MB to 937 MB and 1239 MB (about 4% of a 61 GB root filesystem), see the second-city comment below.

### Fix candidates

1. Consult the external-endpoint deferral **before** dialing: when `preflightIdentityDeferredReader` would defer, skip the probe entirely. Smallest change; keeps the gate semantics already chosen.
2. Dial the probe with the scope-resolved credential (the same value the open projects as `BEADS_DOLT_PASSWORD`) instead of the raw ambient variable.
3. In either case, log the deferral at the failing probe so a hub-side rejection is attributable from the client side.

### Confirmed on main

Read on `main` @ 1154ad128 at the lines above; measured on our fork build whose only divergence in this path is an extra ambient-endpoint check that yields the same empty password. A literal run of a stock `main` binary against the affected city is pending on our side (that city's `bd` import pin currently refuses a stock binary), and will be added here when done.

rjgeng:
## What

Closes #5965. `checkIdentityMatch`'s direct database `_project_id` probe always dialed the target endpoint via `DatabaseProjectID(scope)` first, only consulting `DeferIdentityToNativeOpen(scope)` afterward, in the post-failure fallback branch. For any scope resolving to an external Dolt endpoint (a federated hub, a hosted beads-gateway), that dial is known in advance to be unable to authenticate — the direct root/plaintext probe can't replicate the endpoint's EIA-as-username + TLS credential-command auth — so every such preflight run produced one rejected-authentication attempt logged on the remote Dolt server, even though the triggering command succeeded once the check fell through to the deferral pass. On the reporter's hub host this was 80-84% of that box's syslog volume and multi-hundred-MB/week log growth.

## Fix

This is the issue's own "fix candidate 1" — the smallest, most targeted option. Moved the `DeferIdentityToNativeOpen(scope)` check ahead of the `DatabaseProjectID(scope)` dial in `checkIdentityMatch`: when it reports the scope defers to beadslib's native-open verification, return the deferred Pass result directly, without dialing at all. The post-dial fallback path (dial fails or is unconfigured, no deferral) is unchanged for local scopes whose probe should genuinely succeed.

**Test note:** one pre-existing test (`TestPreflightExternalEndpointStillBlocksOnProbeMismatch`) asserted that a deferred scope still blocks if the dial happens to succeed with a mismatched `project_id`. That scenario is now structurally unreachable — the dial is skipped outright for deferred scopes — and it was never the primary enforcement point anyway: the pre-existing code comment already stated that beadslib's native-open verification is what actually refuses to connect on a real mismatch for such endpoints. Replaced it with `TestPreflightDeferredScopeNeverDialsTheDatabaseProbe`, which asserts `DatabaseProjectID` is never invoked when `DeferIdentityToNativeOpen` returns true — the actual behavioral guarantee this fix makes.

Fix candidates 2 (change which credential the probe itself uses) and 3 (differentiate log messages) from the issue were intentionally not attempted — candidate 1 alone resolves the reported symptom with the smallest possible blast radius.

## Validation

- `go test -tags gms_pure_go ./internal/beads/contract/...` all pass, `go vet` clean
- Full local push-gate run: all failures across every shard and `unit-core` confirmed to be pre-existing environmental/timing flakes (dolt-compaction/preflight-retry timeouts in `examples/bd/dolt/dog_exec_scripts_test.go`, detached-probe/systemctl-delegate/lsof tests, interrupt-timing tests) unrelated to this change — none touch `internal/beads/contract/preflight_checker.go`

### DIFF
--- a/internal/beads/contract/preflight_checker.go
+++ b/internal/beads/contract/preflight_checker.go
@@ -31,15 +31,18 @@ type PreflightChecker struct {
 	BDContext func(scope string) (PreflightBDContext, error)
 	// DatabaseProjectID reads the authoritative database _project_id for the scope.
 	DatabaseProjectID func(scope string) (string, bool, error)
-	// DeferIdentityToNativeOpen reports whether, when the direct database probe
-	// cannot confirm project_id, the scope should stay native-eligible and defer
+	// DeferIdentityToNativeOpen reports whether the scope should skip the direct
+	// database probe altogether and stay native-eligible, deferring
 	// authoritative identity verification to beadslib's native-open path
 	// (verifyProjectIdentity over the authenticated connection) instead of
 	// degrading off the native store. It is true for external endpoints such as
 	// a hosted beads-gateway, whose EIA-as-username + TLS credential-command auth
-	// the control-plane root/plaintext probe cannot replicate, but whose database
-	// _project_id beadslib still verifies at open time — refusing to connect, and
-	// falling back to BdStore, on mismatch. Nil defaults to no deferral (Warn).
+	// the control-plane root/plaintext probe cannot replicate — the probe is
+	// known in advance to be unable to authenticate such an endpoint, so it is
+	// consulted before DatabaseProjectID is dialed, not after it fails — but
+	// whose database _project_id beadslib still verifies at open time,
+	// refusing to connect, and falling back to BdStore, on mismatch. Nil
+	// defaults to no deferral (Warn).
 	DeferIdentityToNativeOpen func(scope string) bool
 	// BeadsLibraryVersion is the linked github.com/steveyegge/beads module
 	// version. Empty means infer it from build info.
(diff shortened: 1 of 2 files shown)


A3Ackerman:
The two-city measurement on #5965 is the external-scope case this removes the dial for: every credential-less in-process open of a hub-bound rig store emitted one rejected authentication at the hub, about 3.5 s per open. Checking the deferral before dialing takes exactly that call away. Composes with #6123, which fixes the credential for the probes that still run.

PR #6008, opened 2026-09-04
fix(rig): report effective suspension in rig add JSON

### LINKED ISSUE #5721: bug: rig add --start-suspended --json reports suspended:false
## Summary

On current `main` (`eec4a2fb625279ac62f29ff4f6a554168bc77b1a`), local `gc rig add ... --start-suspended --json` reports `"suspended": false` even though the rig is correctly stored and treated as suspended on start.

This is a result-envelope truth bug: effective scheduler state is correct, but the command's JSON contradicts it.

## Current-main reproduction

The mismatch is deterministic in the current source projection:

1. `internal/rig/provision.go:337-343` stores the request in `config.Rig.SuspendedOnStart`.
2. `config.Rig.Suspended` is the deprecated legacy field; `EffectiveSuspendedOnStart()` combines the two fields.
3. `cmd/gc/management_json.go:80-98` nevertheless emits `rig.Suspended` directly.

A minimal executable witness using the current `internal/config` package returns:

```json
{"effective_suspended":true,"json_summary_field":false}
```

The live command shape that originally exposed it was:

```bash
gc rig add /path/to/rig --name demo --start-suspended --json
gc rig list --json
gc rig suspend demo
gc rig list --json
```

The add result reported `suspended:false`; controller behavior and the later effective-state read showed the rig suspended.

## Expected behavior

The successful add envelope should report the effective post-operation state:

```json
{"command":"rig add","action":"add","rig":"demo","suspended":true}
```

The result should use `rig.EffectiveSuspendedOnStart()` (or otherwise read the same effective suspension source used by `gc rig list`), with a regression covering `--start-suspended --json` and the post-add effective state.

## Why it matters

Automation reasonably treats a successful mutation envelope as authoritative. Returning `false` after a requested and effective suspended add makes safe dormant-by-default provisioning look live and forces callers to distrust or re-derive command results.

This report is limited to the JSON result mismatch; it does not claim the controller failed to suspend the rig.

vishnujayvel:
## Summary

- Report the effective startup suspension state in the `gc rig add --json` result.
- Use the existing `EffectiveSuspendedOnStart()` helper so current and legacy config fields agree.
- Extend the command-level regression test to cover `--start-suspended`, the JSON envelope, and stored config.
- Thanks to @loucmane for reporting #5721.
- Fixes #5721.

## Testing

- [x] `go test ./cmd/gc -run '^TestRigAddJSONReportsEffectiveSuspendedOnStart$' -count=1`
- [x] `make vet`
- [ ] `make check` (local linter toolchain mismatch; details below)
- [ ] `make test` (unrelated environment/path failures; details below)
- [ ] `make check-docs` (not applicable; no documentation changed)
- [ ] `make test-integration` (not applicable; no runtime, controller, or workflow behavior changed)

## Checklist

- [x] Linked an issue: #5721
- [x] Added or updated tests for the behavior change
- [x] No documentation update is needed for this result-envelope correction
- [x] No breaking changes or migration steps

<details>
<summary>Verification details</summary>

- Regression proof: with the production line reverted, the focused test failed because the payload contained `suspended:false`; restoring the change made the same test pass.
- `make vet` passed.
- `make test` completed with 170 packages passing and 6 packages failing across 11 pre-existing environment/path-sensitive tests. Failures covered live Dolt runtime-port overrides, macOS `/private/var` path normalization, macOS application-support path selection, and worktree provenance; none touched the changed JSON summary or regression test.
- `make check` stopped at `fmt-check` because the installed `golangci-lint` was built with Go 1.25 and refuses the repository's Go 1.26.6 target. `gofmt` and `git diff --check` passed independently.

</details>

### DIFF
--- a/cmd/gc/management_json.go
+++ b/cmd/gc/management_json.go
@@ -90,7 +90,7 @@ func rigAddJSONSummary(rigPath string, rig config.Rig) managementActionResult {
 		Path:          rigPath,
 		Prefix:        rig.EffectivePrefix(),
 		DefaultBranch: rig.EffectiveDefaultBranch(),
-		Suspended:     managementBoolPtr(rig.Suspended),
+		Suspended:     managementBoolPtr(rig.EffectiveSuspendedOnStart()),
 	}
 	if result.Prefix == "" {
 		result.Prefix = config.DeriveBeadsPrefix(name)
--- a/cmd/gc/management_json.go
+++ b/cmd/gc/management_json.go
@@ -19,8 +19,16 @@ type managementActionResult struct {
 	Path          string `json:"path,omitempty"`
 	Prefix        string `json:"prefix,omitempty"`
 	DefaultBranch string `json:"default_branch,omitempty"`
-	Suspended     *bool  `json:"suspended,omitempty"`
-	State         string `json:"state,omitempty"`
+	// Suspended is the acting command's own view of suspension, never a live
+	// effective-state query. `rig add` and `agent add` report the authored
+	// startup default — for rigs, [config.Rig.EffectiveSuspendedOnStart] — so on
+	// a re-add of an existing rig it can disagree with `gc rig list --json`,
+	// which layers the runtime suspension-state override on top of that default
+	// (buildEffectiveSuspendedRigNames). `rig suspend|resume` and
+	// `agent suspend|resume` report the state the command just applied. Consumers
+	// needing a rig's live effective suspension must read `gc rig list --json`.
+	Suspended *bool  `json:"suspended,omitempty"`
+	State     string `json:"state,omitempty"`
 	// Status and RequestID are additive fields the remote `rig add` path emits so
 	// a script repointed at a remote city keeps the automation-critical keys plus
 	// the async outcome (provisioned/exists) and its idempotency id. Local
--- a/cmd/gc/management_json_test.go
+++ b/cmd/gc/management_json_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io"
 	"net"
 	"net/http"
 	"net/http/httptest"
@@ -497,7 +498,7 @@ contract = "gc.healthz.v1"
 	return cityPath
 }
 
-func TestRigAddJSONEmitsOnlySummary(t *testing.T) {
+func TestRigAddJSONReportsEffectiveSuspendedOnStart(t *testing.T) {
 	clearGCEnv(t)
 	t.Setenv("GC_DOLT", "skip")
 	t.Setenv("GC_BEADS", "bd")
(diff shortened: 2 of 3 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277304624

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @vishnujayvel! This PR fixes `gc rig add --json` reporting `suspended: false` for a rig that was created with `--start-suspended` (#5721). That matters because automation treats a successful mutation envelope as authoritative — a core command's JSON result being confidently wrong is exactly the kind of bug that propagates silently into every scripted consumer, and your diagnosis of *why* it was wrong (the write site and the read site use different fields) is what made the fix a one-liner instead of a guess.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- The root cause was pinned precisely and the fix is the right one: `--start-suspended` writes `SuspendedOnStart` (`internal/rig/provision.go:342`) and never the deprecated `Suspended` field, so the old `managementBoolPtr(rig.Suspended)` emitted a deterministic `false`. Reading `EffectiveSuspendedOnStart()` (`internal/config/config.go:1201-1203`) is the read-site invariant that field's own doc comment mandates. Rather than take the premise on faith, the review confirmed it four independent ways — a `go test -overlay` mutation probe, a writer-vs-reader trace, a fresh call-site census, and a probe that reverts the production line and watches the new test go red.
- Review found that `EffectiveSuspendedOnStart()` is the *authored startup default*, not the *live effective state*. On a re-add of an existing rig whose runtime entry contradicts that default, `rig add --json` can therefore disagree with `rig list --json`, which layers the runtime override on top. This was demonstrated end to end rather than asserted. The final patch resolves it by making the contract explicit on both surfaces automation reads.
- Review found the suite never value-asserted the `suspended: false` arm — an always-`true` mutant survived the entire test surface. The final patch closes that, and the same mutant now fails.

Review findings addressed:
- [Contract & Interface Fidelity, minor] The `suspended` field's semantics were implicit, leaving the authored-default vs live-effective distinction undocumented for consumers. Documented in `55cdc1aa` on both automation surfaces: the Go doc on `managementActionResult.Suspended` (`cmd/gc/management_json.go:22-29`) and the published `--json-schema=result` description (`schemas/rig/add/result.schema.json`).
- [Test Evidence Quality, minor] The `suspended: false` default arm was never value-asserted. A one-line `check` assertion was added to the no-flag `rig add` case in `55cdc1aa`, and it is mutation-proven: the always-`true` mutant that previously survived the whole suite now dies in that subtest, which also proves the new check is actually invoked by the table harness.
- [Contract & Interface Fidelity, nit] The new doc comment sits on a field that `agent add` also populates, but closed with a universal "Consumers needing live effective suspension…" pointer aimed at a rig-only command. Scoped to "a rig's live effective suspension" in `55cdc1aa`.

- `ga-bx8b1` [Test Evidence Quality, minor] — the re-add contract this PR documents has no test witness. A refactor that took the `resolveResultRig` fallback branch would emit the deprecated field and silently contradict the text this PR just published, with a green suite.
- `ga-ansdb` [Architectural Consistency, minor, pre-existing] — the raw `config.Rig.Suspended` stale-read class has at least 8 live sites, not the 3 originally counted. The most operationally consequential, `internal/sling/sling_core.go:241`, was untracked by any prior pass: a rig suspended only via the modern `suspended_on_start` key gets no suspended-rig warning from `gc sling`.
- `ga-a2x2g` [Change Impact / Blast Radius, minor, pre-existing] — `gc rig status --json` (`cmd/gc/cmd_status.go:347`) still builds its envelope from the same deprecated field, so it can disagree with `rig add --json` for the same rig.
- `ga-665gd` [Architectural Consistency, nit, pre-existing] — the `resolveResultRig` fallback (`internal/rig/provision.go:679`) still writes the deprecated `Suspended` field, against the migration direction documented at `internal/config/config.go:1194-1200`.

None of these gate this PR: each is either pre-existing in a file this PR does not touch, or a hardening follow-up on top of the contract this PR establishes.

### Maintainer Changes
`fixup(rig): document rig add --json suspended as the authored startup default` — 3 files changed, 19 insertions(+), 3 deletions(-). It documents the `suspended` contract on the Go doc and the published result schema, and adds the missing `suspended: false` assertion. Your original commit is preserved unchanged, with your authorship.

### Final Review Status
Ready for the merge queue: final head `55cdc1aa` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/34168719101

### Review Iterations
2 review passes. Iteration 1 (Claude + Codex) approved with one documentation item and one test-coverage item. Iteration 2 (Claude + Codex + Gemini, full cross-model coverage restored) confirmed both were discharged and mutation-proven, and contributed the census correction now tracked in `ga-ansdb`.

---

PR #6009, opened 2026-09-04
Keep non-live session census cache-served

quad341:
## What this changes

Non-live session census reads now stay on the active-session cache after a partial prime. Previously, the census omitted the open-status predicate, so the cache treated the query as unsafe and read through to the backing store on every call.

The fix states the existing open-session constraint only for non-live queries that exclude closed sessions. Live queries keep their current shape and continue to bypass the cache.

## Review notes

- The production change is confined to `internal/session/list_all.go`; it does not change the cache implementation itself.
- The regression test primes a cache, writes an unseen session directly to the backing store, and verifies the ordinary census remains on its snapshot.
- No configuration, API, schema, or migration surface changes.

## Test plan

- [x] Run the documented 40-job full local suite with rootless Podman configured; inspect all raw failures against their existing condition trackers.
- [x] Run the complete `internal/session/...` package and verify the new regression test passes by name.
- [x] Run policy, dependency-boundary, affected-lint, formatting, docs-sync, build, and vet checks.
- [x] Release gate: [`release-gates/ga-px24i7-session-listall-partial-cache-gate.md`](release-gates/ga-px24i7-session-listall-partial-cache-gate.md)

### DIFF
--- a/internal/session/list_all.go
+++ b/internal/session/list_all.go
@@ -338,6 +338,22 @@ func (s *Store) listAllBeads(opts ListAllOptions) ([]beads.Bead, error) {
 		Limit:         opts.Limit,
 		Live:          opts.Live,
 	}
+	if !opts.IncludeClosed && !opts.Live {
+		// Session bead status is open/closed; richer lifecycle states live in
+		// metadata. Stating the predicate keeps this census cache-servable
+		// under a partial prime, mirroring listSessionBeadsByMetadata. Scoped
+		// to !Live: a Live query bypasses CachingStore's cache path entirely
+		// (query.Live routes straight to the backing store — see
+		// CachingStore.List), so cacheServableForListQueryLocked never sees
+		// it and the predicate would be inert for cache-servability. Setting
+		// it anyway would change the query SHAPE of every Live+non-IncludeClosed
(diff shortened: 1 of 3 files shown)


quad341:
@quad341 — good catch, and a genuinely hard one to see. `listAllBeads` *looked* right, because `IncludeClosed: false` does express the correct intent; it takes reading `cacheServableForListQueryLocked` against `partialPrimeStatuses` to realize the cache needs the predicate **stated**, not implied. Merging as-is.

Two things I especially liked. The `!Live` scoping and the comment defending it — setting `Status` unconditionally would have looked tidier and quietly reshaped every live pool-availability query. And the regression test is a real RED/GREEN: priming a live `CachingStore` and then writing to backing behind its back is the only honest way to prove the census stayed on its snapshot, and asserting the primed bead is still there stops an empty result from passing.

Two follow-ups I'll file on our side, neither yours: `cachedListUnion` still builds its CacheFirst legs without `Status`, so that tier keeps missing under a partial prime; and nothing yet pins census behavior for a session bead in a status that's neither open nor closed. Both are ours to carry. Thanks for the tight scope and the release-gate record.


---

PR #6010, opened 2026-09-04
Make gc dolt sync remote selection deterministic and safe

quad341:
## What this changes

`gc dolt sync` now selects push remotes deterministically. An explicit `GC_DOLT_REMOTE_<DB>` override wins; otherwise a local `file://` remote is preferred. If only non-local remotes are available without an override—including a sole non-local remote—the database is skipped instead of risking an unintended push. SQL and CLI fallback paths share the same policy.

## Review notes

- Review `examples/bd/dolt/commands/sync/run.sh` for override validation and the shared SQL/CLI selection path.
- This is a safety behavior change: unattended sync may now skip a sole non-local remote unless explicitly pinned.
- No config migration is required; operators that intend a network push can set `GC_DOLT_REMOTE_<DB>`.

## Test plan

- [x] Run `make test-local-full-parallel`; confirm the Dolt package passes and unrelated fleet failures match tracked conditions documented in the gate.
- [x] Run the verbose full `examples/bd/dolt` suite and confirm all modified tests execute with no skips.
- [x] Run policy, affected lint/format, and full vet lanes.
- [x] Release gate: [`release-gates/ga-n9frq0-dolt-sync-remote-selection-gate.md`](release-gates/ga-n9frq0-dolt-sync-remote-selection-gate.md)

### DIFF
--- a/examples/bd/dolt/commands/sync/run.sh
+++ b/examples/bd/dolt/commands/sync/run.sh
@@ -24,25 +24,26 @@
 #      by name (same key derivation as GC_DOLT_REFSPEC_<DB_UPPER> above). This
 #      pins the remote regardless of scheme — an explicit override is by
 #      definition intentional, so it is honored even for a non-local remote.
-#   2. Exactly one configured remote: used regardless of scheme. There is no
-#      ambiguity to gate when the database was only ever told about one
-#      remote — being the sole choice is itself an instruction to use it.
-#   3. Multiple configured remotes, no override: the first file:// remote in
-#      name-sorted order wins, deterministically and independent of whatever
-#      order the database itself returns candidates in (gc-fqi7kq: an
-#      unordered SELECT made this pick 1-of-N at random, occasionally
-#      selecting a public network remote for private fleet data). Preferring
-#      file:// keeps local/backup remotes from losing to a network remote by
-#      accident.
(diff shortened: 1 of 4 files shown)


quad341:
@quad341 — this is a well-built fix. Sorting the candidates before applying any policy rule, rather than just adding `ORDER BY name` and trusting the server to honor it, is the version that actually holds, and the docstring calling that out explicitly will save the next person an hour. `TestSyncRemoteSelectionIsOrderIndependent` running three real permutations — plus every new test asserting *no push happened*, not just the expected message — is exactly the shape that keeps `gc-fqi7kq` from coming back.

I'll take the integration cleanup on my side and merge. The header's "Remote resolution" rule 2 still describes the old sole-remote-wins-regardless-of-scheme behavior, so it now contradicts `select_remote` and your own tests; I'll rewrite that block, add `GC_DOLT_REMOTE_<DB>` to the header and `--help` since it's the only escape hatch, make the skip message stop saying "multiple remotes" for a single one, and split the override-mismatch failure reason from a real query failure the way `pull` already does. I'll add the CLI-mode versions of your selection tests too.

One for a follow-up bead: sync pins a non-local push behind one flag while pull requires two for a non-local read. Worth settling when those two selectors get consolidated.


---

Issue #2895, opened 2026-06-01
Session Loop reconciler: TTL-reap stuck start-pending session creates + deduplicate pending-creates per template


bourgois:
### Summary

Two related Session Loop gaps produce self-amplifying spawn storms: (1) sessions stuck in `start-pending` / `create-pending` accumulate indefinitely without a TTL, bloating the open bead count and confusing pool demand accounting; (2) a reset or re-route can enqueue a second pending-create for the same template without cancelling the first, producing duplicate spawns that compete for the same work bead.

### Observation (2026-05-30)

After a restart of the gc supervisor with `pack 0.8.58`, 57 session-type beads ended up in `state=start-pending` (reason: `create,config`). Of these, only 13 were truly live (visible in `gc session list`). The remaining 44 were throttled by the wake-budget at startup and never progressed to `active`. They remained in `start-pending` indefinitely because the reconciler has no timeout on the pending-create path. Symptoms:

- `bd ready` listed ~80 session beads vs ~14 live sessions — confusing operator view.
- Pool demand accounting was inflated → spawner over-counted active sessions.
- `bd stats` showed 80 open vs 14 live — routine `bd doctor` fired stale warnings.

### Fix 1 — TTL-reap stuck pending-creates (FR-SC.1)

In the Session Loop reconciler, track `created_at` for each pending-create entry. If a session remains in `start-pending` / `create-pending` for longer than `N` minutes (recommended: 10 min, configurable via `city.toml supervisor.pending_create_ttl_min`) without transitioning to `active`:

1. Cancel the pending create (abort the tmux/process spawn if not yet started, or kill if started but not active).
2. Close the session bead with `close_reason = "pending-create-timeout"`.
3. Emit `gc.session.pending_timeout` OTLP event for observability.
4. Allow the normal pool demand reconcile to re-spawn if work is still present and the pool has capacity.

**Why not just close the bead**: the bead close is insufficient alone — the reconciler must also cancel its internal create-in-flight record so it doesn't try to bring the same session active after the TTL fires.

### Fix 2 — Deduplicate pending-creates per template (FR-SC.2)

The reconciler MUST NOT hold more than one `pending-create` per `(rig, template)` combination at a time. When a new spawn is triggered for a template that already has a `start-pending` entry:

- **If the existing pending-create is recent** (< TTL): skip the new spawn; the existing one will either become active or be TTL-reaped.
- **If a re-route changed the target template**: cancel the old pending-create before starting the new one.

[...]

PR #6016, opened 2026-09-04
Persist startup-health episodes and suppress pending-create retry storms

quad341:
## What this changes

Failed session starts now accrue in a durable startup-health episode keyed by session name. The state survives pending-create rollback, replacement beads, controller restarts, named-session materialization, and pool expansion, so repeated failures can no longer reset themselves by creating a new transient record.

On the fifth consecutive failure, the existing five-minute quarantine blocks another provider start until expiry. Active failure count and kind are mirrored onto the visible session metadata, and both the durable episode and mirror clear only after a confirmed successful start.

## Why it is built this way

The retry history is stored separately from transient session-creation beads, while the visible session row carries only a typed projection for operators. This keeps persistence and presentation decoupled without adding another config or lifecycle surface.

## Review notes

- The new typed record and transition logic live in `internal/session/startup_health.go`.
- Reconciler integration gates `Provider.Start`, updates the visible metadata projection, and preserves the existing rollback, rate-limit, terminal-error, and circuit-breaker behavior.
- The threshold and quarantine duration are existing defaults; this adds no config knob, endpoint, migration, or CI workflow change.
- Startup-health bookkeeping records are excluded from Ready and normal session listings.

## Test plan

- [x] Exercise named and pool-expanded sessions through real production materialization and confirm a sixth start is blocked after five failures.
- [x] Verify quarantine expiry permits retry and a confirmed successful start clears and rearms the episode.
- [x] Run the repository's complete local CI union plus required policy, lint, vet, build, docs, and native-DoltLite lanes.
- [x] Release gate: [`release-gates/ga-02cm7z-startup-health-episodes-gate.md`](release-gates/ga-02cm7z-startup-health-episodes-gate.md)

Tracking: ga-02cm7z

---
- Related to #2895 — adds a durable per-session-name startup-failure record that quarantines a session after five consecutive failed provider starts, so a failing create can no longer reset its own retry count by re-materializing a fresh pending-create bead; it does not add the pending-create TTL reap or the per-template deduplication that issue asks for, so both of those remain open
- Related to #5442 — adds a durable per-session-name startup-failure budget that quarantines a session for five minutes after five consecutive failed starts, throttling the unbounded retry storm described there; it leaves the keyless fresh-start fallback in internal/session/chat.go, the misleading "resuming session" wrapping, the terminal failed state, doctor visibility, and pool-slot release untouched, so that issue stays open


quad341:
@quad341 — merging this as-is. The two real-materialization tests are what sold it: driving `syncSessionBeads` for named *and* pool-expanded identities, instead of hand-writing replacement beads, is what turns "the episode survives bead replacement" from a claim into something the suite actually holds. And having the pool test assert that the `poolRuntimeSessionName` override really fired means it can't quietly rot into a no-op later. Fixing the `unparam` finding that failed the `ga-em8g4o` gate by hoisting the constant, rather than suppressing it, is the right instinct too.

I checked the things I'd otherwise take on faith: the threshold and quarantine duration really are the existing `session_types.go` defaults, and the Ready exclusion really does bite — every gc store path filters post-decode through `IsReadyExcludedType`.

Three follow-ups I'm filing rather than holding you up for: episode beads are created but never closed, so they accrue one durable row per failed session name forever; a transient store error while clearing an episode has no retry path, so the next failure re-quarantines instead of the fifth; and `AlertDisposition` is persisted with no reader yet. Also worth a quick `bd ready` sanity check — bd's own exclusion list lives in the other repo.


---

PR #6017, opened 2026-09-04
feat(beads): add provider-neutral init intent precedence

julianknutsen:
Clean scoped replacement for the rejected precedence implementation. Adds provider-neutral initialization intent resolution with persisted initialized state authority, explicit CLI/environment/city/default precedence for fresh scopes, no ambient BEADS_DOLT_* reads, and preservation of initialized non-Dolt backends. Focused contract tests cover conflicts, partial selectors, canonicalization, and backend preservation.


julianknutsen:
Audit disposition: hold out of adopt-pr. The init-selector implementation selects proxied-local storage for fresh Dolt scopes, so it shares the default-rollout performance gate. The branch has been rebased locally and the direct/external preservation and canonicalization tests from #5996 retained.

Issue #2843, opened 2026-05-31
feat(runtime): stamp gc.work_dir + gc.session_name on graph.v2 molecule step beads


sjarmak:
## Summary

Graph.v2 molecule formula-run **step beads carry no execution metadata** (`metadata: {}`), so dashboards and other consumers cannot resolve a step's git diff or back-map a step to the session that ran it. Stamping two keys — `gc.session_name` and `gc.work_dir` — onto each step bead lights up both the diff view and the session drill-in downstream with **zero consumer change**. This is strictly less work than #2823 (no server-side diff computation/transport) and can supersede it for the local-worktree case.

## Symptom

On a live city (`ds-research`), every formula-run step bead returns empty metadata:

```
GET /v0/city/ds-research/bead/gc-1920   # mol-focus-review root
→ metadata: {}                          # 0 keys, 69-byte response

GET /v0/city/ds-research/bead/gc-1921 .. gc-1927   # its 7 step beads
→ metadata: {}                          # same — every step
```

Confirmed across 6 runs (`mol-focus-review`, `mol-do-work`). Every run resolves:
- execution-path → `unavailable (missing_cwd_and_rig_root)`
- per-node session → `{ kind: none, reason: session_unresolved }`

The supervisor session list is no fallback: those session records carry no `bead_id` and no `cwd`, so a consumer cannot back-map a step node to its session from either side.

Result: a consumer that wants "show me the diff this step produced" or "open the session that ran this step" has nothing to key on. expected: step bead exposes the worktree path + session it ran in → got: `{}`.

## Root cause

A molecule step bead is built by `stepToBead()` at `internal/molecule/molecule.go:942`, which copies **only** `step.Metadata` (plus `notes`) into the bead:

```go
// internal/molecule/molecule.go:957-966
if len(step.Metadata) > 0 || step.Notes != "" {
    b.Metadata = make(map[string]string, len(step.Metadata))
    for k, v := range step.Metadata {
        b.Metadata[k] = formula.Substitute(v, vars)
    }
    ...
}
```

So a step bead's metadata is exactly whatever the routing layer stamped onto `step.Metadata` earlier. That stamping happens in `DecorateGraphWorkflowRecipe()` at `internal/graphroute/graphroute.go:398`, which **already** writes a family of `gc.*` keys onto each step (`internal/graphroute/graphroute.go:440-454`):

```go
step.Metadata["gc.root_store_ref"] = rootStoreRef
[...]

PR #6020, opened 2026-09-05
fix(graphv2): route single-session config-agent steps by gc.routed_to, leave Assignee empty

julianknutsen:
## Problem

Graph.v2 route binding pre-stamped a **single-session config agent's** step `Assignee` with the **munged tmux runtime session name** — `SanitizeQualifiedNameForSession` turns `gascity-packs/gc.run-operator` into `gascity-packs--gc__run-operator` (`/`→`--`, `.`→`__`).

That's a category error. **`Assignee` is a unique reference to a concrete session**; an agent / config alias is a *routing* target. Config-agent work is delivered by **`gc.routed_to`** (the alias), and `Assignee` must stay empty until a session claims the step and stamps its own id — which is exactly what the pool lane already does. Stamping an agent-shaped value at cook time:

- matched no identity in the hook claim set (`hookClaimHasIdentity`), the controller's named-session wake-demand set, or the dead-assignee reaper's set — all keyed on concrete session identities — so the step could never be claimed *or* reopened; and
- because the bead was now "assigned", **hid it from the `--unassigned` `gc.routed_to` demand query** that wakes and feeds the seat in the first place.

Net: every operator / fanout / attempt step for a single-session config agent sat unclaimable until an operator hand-claimed it (observed continuously on the normalized issue/PR pilot in maintainer-city). `gc.routed_to`/`gc.run_target` were already correct — the divergence was the tell.

## Fix

Both route sinks now leave `Assignee` **empty** for the config-agent lane, so delivery falls through to the existing `gc.routed_to` demand + claim path (the concrete session stamps its id when it claims):
- `internal/graphroute/graphroute.go` — `ApplyGraphRouteBinding`
- `internal/dispatch/control.go` — `applyAttemptStepRoute` (attempt/fanout re-route)

The durable `gc.session_name` dashboard back-reference (#2843) is **preserved**. The now-unused `attemptRouteBinding.sessionName` field and its `NamedSessionRuntimeName` computation are removed. This is the same routed-delivery behavior pool agents, named sessions, and the control-dispatcher lane already use — the single-session config-agent lane was the lone exception that pre-assigned.

## Tests

- `TestApplyAttemptStepRoute_SingleSessionConfigAgentRoutesUnassigned` (new): a single-session config agent's attempt step is routed with `gc.routed_to` = the alias and **empty** `Assignee`.
- `session_stamp_test.go` updated to the routed contract (Assignee empty) while still asserting the `gc.session_name` back-ref is written.
- `internal/graphroute/...` and `internal/dispatch/...` suites green; both packages build.

Commit `6de34bd1be`.

### DIFF
--- a/internal/dispatch/control.go
+++ b/internal/dispatch/control.go
@@ -1295,11 +1295,13 @@ func applyAttemptStepRoute(step *formula.RecipeStep, target string, cfg *config.
 			delete(step.Metadata, beadmeta.ExecutionRoutedToMetadataKey)
 		}
 		step.Labels = removeAttemptPoolLabels(step.Labels)
-		if binding.metadataOnly {
-			step.Assignee = ""
-			return
-		}
-		step.Assignee = binding.sessionName
+		// Config-agent attempt/fanout work is delivered by gc.routed_to (the
+		// alias), not by a pre-stamped Assignee. Assignee names a concrete
+		// session and stays empty until one claims the step and stamps its own
+		// id. The previous single-session branch stamped the munged runtime
+		// session name, which no claim/wake/reaper identity matched and which
+		// hid the bead from the --unassigned routed-demand that wakes the seat.
+		step.Assignee = ""
 		return
 	}
 
@@ -1392,7 +1394,6 @@ func latestAttemptCandidateIsControlInfrastructure(kind string) bool {
 type attemptRouteBinding struct {
 	qualifiedName   string
 	metadataOnly    bool
-	sessionName     string
 	directSessionID string
 }
 
@@ -1421,9 +1422,7 @@ func resolveAttemptRouteBinding(target string, cfg *config.City, store beads.Sto
 			binding := attemptRouteBinding{qualifiedName: agentCfg.QualifiedName()}
 			if isAttemptMultiSessionTarget(agentCfg.QualifiedName(), cfg) {
 				binding.metadataOnly = true
-				return binding, true
 			}
-			binding.sessionName = config.NamedSessionRuntimeName(cfg.EffectiveCityName(), cfg.Workspace, agentCfg.QualifiedName())
 			return binding, true
 		}
 	}
--- a/internal/dispatch/control_integration_test.go
+++ b/internal/dispatch/control_integration_test.go
@@ -1197,7 +1197,7 @@ func TestResolveAttemptRouteBinding_NamedSessionTargetUsesCanonicalBeadID(t *tes
 	if binding.directSessionID != named.ID {
 		t.Fatalf("directSessionID = %q, want canonical named bead ID %q", binding.directSessionID, named.ID)
 	}
-	if binding.qualifiedName != "" || binding.sessionName != "" {
+	if binding.qualifiedName != "" {
 		t.Fatalf("binding = %+v, want direct named session only", binding)
 	}
 	// Per-resolution List calls must stay bounded so the per-attempt cost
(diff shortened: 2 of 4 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-699284130204932838

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR fixes the routing defect that left every
single-session config-agent step unclaimable — the dispatcher was stamping a munged pre-assignment
into `Assignee` instead of routing by `gc.routed_to` — so those steps sat in the queue until an
operator hand-claimed them, blocking an entire delivery lane. Routing by `gc.routed_to` and leaving
`Assignee` empty until claim is the right shape: it lets the seat's own generated work query find
the bead through the routed `--unassigned` tier, and lets `gc hook --claim` own the assignment
transition exactly as it does for every other lane.

### Original PR Review
Decision: approve (iteration 1 opened at request_changes and resolved to approve in iteration 2).

Specific gaps fixed:
- **The newly load-bearing delivery invariant was asserted nowhere.** The PR's new assertions all
  checked the route-time shape; the delivery half it made load-bearing — the seat's own work query
  surfacing an unassigned bead carrying its identity — spanned three packages with nothing pinning
  it. Renaming either identity, or stamping `GC_SESSION_ORIGIN` on a plain config agent, would have
  re-created the unclaimable-bead outage with a fully green suite. The final patch pins both halves:
  `internal/config` pins the consuming side (`poolDemandTarget()` is the bare qualified name, and
  the routed tier serves an unassigned bead only when the read carries `gc.routed_to=<name>`), and
  `internal/graphroute` pins the stamping side through the real `GraphRouteBindingForAgent`, so a
  rename of `RoutedToIdentity()` now reddens both. Both reviewers mutation-probed the pin
  independently to confirm it actually bites rather than passing vacuously.
- **Pre-claim session identity was left undocumented.** With `Assignee` now empty, nothing
  corroborates the `gc.session_name` back-reference at route time, so a reader preferring that key
  can name a session that does not exist until `hookClaimIdentityPatch` rewrites it on claim. The
  final patch records that explicitly at the stamping site.
- **The in-memory graph harness diverged from production claim behavior.** It eagerly claimed every
  routed candidate — leaving steps assigned but never executed — and carried an inline kind-exclusion
  list that had drifted from the selector's. The final patch makes the harness claim one bead per
  pass, the way a real `gc hook --claim` does, and derives both the claim loop and the selector's
  latch arm from one shared predicate rooted in `beadmeta.WorkflowTopologyKinds`, so a third inline
  copy can no longer drift.

Review findings addressed:
- All of the above, plus the second-iteration review of the fixup itself, were fixed in
  `ab1b5e4b08`:
  - [Test Evidence Quality, minor] The over-claim control test could not fail — its fake keyed on a
    foreign route this seat can never ask for, so it stayed green even against a mutant that dropped
    the identity predicate. It is now inverted: the fake refuses any read naming this seat's own
    route and serves foreign work to everything else, so a query that stopped keying on the seat's
    identity reddens it.
  - [Architectural Consistency, minor] The new `bd ready` fake was a near-copy of the package's
    existing one, and depended on the incidental order in which `PoolDemandServeRules.ShellArgs()`
    renders its flags. Both fakes now come from one shared builder, and neither depends on flag
    ordering.
  - [Debuggability & Operability, nit] The fixup commit message claimed a coverage gain that had not
    happened (`spec` latch assertion "made reachable"). It now states the real effect — a
    pre-assigned `spec` step raises the latch assertion instead of being accepted as ordinary work.
  - [Test Evidence Quality, nit] The `spec` latch behavior had only a structural guarantee. The
    latch-rejection test is now table-driven over `beadmeta.WorkflowTopologyKinds`, so that arm is
    asserted for every kind that reaches it.

- `mc-oj26a` — [Test Evidence Quality, non-gating] Nobody enumerated the installed role packs for a
  single-session agent that also backs a `[[named_session]]` whose identity differs from the backing
  template's qualified name. If such a pairing exists, `poolDemandTarget()` and `GC_ALIAS` diverge
  and the origin gate fails closed — a case the assigned tier used to mask. The `namedSelfTargetAdmit`
  comment asserts the identities coincide by construction and no counterexample was found while
  reading `internal/config`, but the live pack population was never swept. Not addressed in this PR.
- `mc-5efux` — [Test Evidence Quality, non-gating] The new tests pin the *query* side of the
  `GC_SESSION_ORIGIN` contract by building the seat's environment themselves. A future spawner change
  that begins exporting `GC_SESSION_ORIGIN` for plain config agents would short-circuit the origin
  gate and would not redden them. Not addressed in this PR.

### Maintainer Changes
One maintainer fixup commit, `ab1b5e4b08` "test(graphv2): pin the routed delivery leg for
single-session config agents" — 5 files changed, 270 insertions(+), 29 deletions(-). It is
test-and-comment only: the two production lines in `internal/dispatch/control.go` and
`internal/graphroute/graphroute.go` are byte-identical to your original change apart from an added
explanatory comment. Your commits and authorship are preserved.

### Final Review Status
Ready for the merge queue: final head `ab1b5e4b08` has passing required GitHub checks, and the
merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can
pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/35812534057

### Review Iterations
2 review passes performed: iteration 1 found one gating major and three minors and closed at
request_changes; the maintainer fixup addressed all four; iteration 2 re-reviewed the fixup, found
only minors and nits, and closed at approve. The remaining minors and nits from iteration 2 were
also applied before merge.

---

PR #6021, opened 2026-09-05
fix(gc init): exclude provider-owned beads state from templates

julianknutsen:
Scoped lifecycle delegation slice: init --from excludes provider-owned .beads state. Focused tests cover skip policy and end-to-end copy behavior.

### DIFF
--- a/cmd/gc/cmd_init.go
+++ b/cmd/gc/cmd_init.go
@@ -1633,7 +1633,11 @@ func writeInitAgentPrompts(fs fsys.FS, cityPath string, cfg *config.City, stderr
 // when copying a city template directory via --from. Skips .gc/ runtime state.
 func initFromSkip(relPath string, isDir bool) bool {
 	top, _, _ := strings.Cut(relPath, string(filepath.Separator))
-	if top == ".gc" {
+	// Provider-owned beads state is initialized by the selected provider in
+	// the destination. Copying it from a template can carry stale process,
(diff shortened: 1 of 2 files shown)


quad341 (approved):
@julianknutsen — thanks for this one. Excluding `.beads` from `--from` copies is a small diff that closes a genuinely nasty hole: `.beads/identity.toml` is deliberately un-ignored in `.gitignore`, so a cloned city template was handing the new city the *source city's* canonical `project.id`. With a fresh database on the other end, that id gets stamped straight into the new city's Dolt state — two cities quietly sharing an identity, which surfaces three weeks later as an inexplicable ledger. Good catch.

I especially appreciate that you didn't stop at the predicate test. `TestDoInitFromDirExcludesProviderOwnedBeadsState` driving the real `doInitFromDir` with a populated `.beads/provider-runtime/` is the test that would actually catch a regression here; the table row alone wouldn't have.

Merging as-is — nothing needed from you. Two notes for the record, neither a request: the skip matches the top path segment only, so a nested `rigs/*/.beads` in a template still copies (same as the existing `.gc` behavior, so consistent). And this silently drops ledger carry-over for anyone using `--from` as a migration; a stderr note might spare someone a confused bug report someday. Happy to file follow-up beads for either if you think they're worth it.

The one red check is the evidence watchdog hitting its 25-minute deadline before CI finished — I'll re-run it on our side.


---

PR #6001, opened 2026-09-04
fix(dashboard): use bd ping for proxied rig health

julianknutsen:
Implements ga-p9iuv.6.

Dashboard rig health now uses the Beads provider-neutral `bd ping --readonly --db ... --json` path, which works for embedded, direct-server, and proxied-server stores. Results are normalized into the existing check/rollup shape, with provider errors surfaced and no mutation. Added malformed, cancellation, argv-guard, no-mutation, and probe failure tests.

Tests: `go test ./internal/api/dashboardbff`; `go vet ./internal/api/dashboardbff`; `git diff --check`. Repository fast pre-push suite was attempted; unrelated existing cmd/gc metadata expectations fail on this branch (server vs proxied-server).

---
🔗 **Maintainer cross-reference** — added by the gascity maintainers, no action needed from you:
- Related to #4902 — replaces the dashboard's bd doctor check pipeline with a single bd ping connectivity check, so the benign dolt "uncommitted change" warnings described in that issue no longer reach the per-rig panel; the "Live issues" mislabeling is untouched and the rollup still escalates any non-ok check

<sub>Linked for triage visibility — not auto-closing. If this looks off, just delete this block.</sub>

Issue #4902, opened 2026-08-01
bug: dashboard "Bead stores · per rig" — "Live issues" counts closed rows, and benign bd-doctor dolt warnings flagged permanently


edwarddavison:
## Summary

The **Health → "Bead stores · per rig"** panel has two metric defects, both
visible on every rig:

- **A. "Live issues" is a *total* count (includes closed).** The row labelled
  **Live issues** shows *all* rows in `issues`, not the live (non-closed) subset.
- **B. Benign `bd doctor` dolt warnings are surfaced permanently.** Every rig
  shows `Dolt Status: N uncommitted change(s)` and
  `Dolt Locks: Uncommitted changes detected`, escalating the rig rollup to
  **warnings**, even when the store is healthy. Root cause is in the bundled
  `bd doctor` — filed separately as **gastownhall/beads#5260** — but the
  dashboard's blind escalation is worth hardening too (below).

Both reproduce against **v1.4.0** (`a7297c51`) + **bd v1.1.0**, and on the live
`gc 1.3.1` deployment.

---

## A — "Live issues" counts closed rows

**Where the value comes from:**

1. Frontend labels it "Live issues":
   `internal/api/dashboardspa/web/frontend/src/routes/Health.tsx:508`
   ```tsx
   {rig.issueCount !== null && (
     <Kv label="Live issues" value={rig.issueCount.toLocaleString()} />
   )}
   ```
2. BFF fills `issueCount` from the `bd doctor` check named `…Issue Count`:
   `internal/api/dashboardbff/samplers.go:509` (`issueCountFromChecks`) — parses
   the first number out of the check message.
3. That check is `bd doctor`'s **"Dolt Issue Count"**, whose message is
   `SELECT COUNT(*) FROM issues` → `"N issues"` (a **total**, correctly named on
   the bd side).

So the dashboard takes a **total** count and labels it **"Live issues."**

**Live repro (`gci` store):**
```
SELECT status, COUNT(*) FROM issues GROUP BY status;
  closed       52
  in_progress   1
  open          1        -> total 54, live (non-closed) = 2
```
The panel renders **"Live issues: 54"**; the true live count is **2** — off by
exactly the 52 closed rows. On a busy rig (e.g. Flinders: ~792 total ≈ 770
closed + 22 live) the number is ~36× too high.

**Fix options:**
- *Minimal / correctness:* relabel `Live issues` → **`Issues`** (it is the
  total). Stops the panel from asserting something false.
[...]

PR #6023, opened 2026-09-05
fix(dashboard): use bd ping for proxied rig health

julianknutsen:
Clean replacement for #6001, rebased from origin/main.

This branch contains only the dashboard proxy health implementation and tests. It uses bd ping for provider-neutral connectivity, reports provider failures, and permits direct TCP probing only when persisted Dolt mode is explicitly server; missing, unknown, embedded, malformed, or unreadable state fails closed.

Validation: go test ./internal/api/dashboardbff; go vet ./internal/api/dashboardbff; make dashboard-ci; git diff --check.

The original #6001 remains open for historical review and is not the merge target.

Note on scope: the `proxied-server` handling in this change is forward-looking. Nothing in this repository writes that Dolt mode today — it appears only in this change's own code and tests — so that path is prospective. The store-materialization guard the probe now applies is not: an existing but unmaterialized `.beads` needs no special configuration and is reachable today, and it is the shape the guard exists for.

---
- Related to #4902 — removes the mislabeled "Live issues" metric (Bug A) and replaces the per-rig `bd doctor` probe with a single `bd ping` connectivity check, so the benign "Dolt Status / Dolt Locks — uncommitted changes" warnings no longer reach the rig rollup (Bug B's symptom); it does not add a real live/non-closed count, and the dashboard-side hardening suggested for that second bug — letting severity/category drive the rollup instead of escalating every non-ok check — is unchanged, as is the upstream bd root cause


quad341 (dismissed):
@julianknutsen — thank you for the clean rebase, and especially for the mode-marker matrix in `samplers_test.go`. The fail-closed table (malformed metadata overriding a `server` config, `metadata.json` present as an unreadable directory, config-without-mode) is what made this easy to trust: I could check each branch against real state on disk instead of reasoning about it.

I chased the one thing this actually rests on — whether `bd ping` exists on our minimum-supported pin, not just the current one. It does: at beads v1.0.4, `--db` is a root persistent flag and `--json` emits `{"status":"ok"}` or `{"status":"error","error":...}`, which is exactly the pair your parser handles. Good instinct handling the error envelope without a `status` key too.

The stderr-synthesis branch is a real upgrade over what was there — an unopenable store used to disappear into an empty `warn` with the provider's error thrown away.

One thing on my side: `internal/beads/boundary_test.go` still describes the BFF as running `bd doctor`. I'll fix that comment and merge — nothing needed from you. Worth closing #6001 when you get a moment.


---

julianknutsen:
Rebased onto main a95c730a (beads v1.3.0-rc.2) as a linear history (force-push, old head 1be1683ef2 dropped its merge commit). Verified against the real rc.2 binary: `bd ping --db <path> --json` returns {status:"ok",...}; `--readonly` and `bd doctor` are refused in proxied mode as the PR assumes. Two additions beyond the rebase: (1) the SPA dist rebuild required by `make dashboard-ci`; (2) `fix(dashboard): skip the probe on an unconfigured .beads` — on rc.2 `bd ping` against a `.beads` with no store marker CREATES an embedded store and answers ok, so the probe now fails closed there (test TestProbeRigSkipsUnconfiguredStore). Comments now state that `bd ping` is a full provider open that restarts a stopped proxy (bounded: sampler dies with the supervisor; `gc stop` stops proxies last). Gates: go build/vet, dashboardbff tests, `make dashboard-ci`, vitest 899 tests, preview served 200, pre-push fast gate. Follow-ups filed separately (ping timeout vs cold start, readDoltMode → contract readers, missing probe edge-case tests). Part of the beads v1.3.0-rc.2 proxied-local default work (epic ga-p9iuv); the lifecycle PR is feat/beads-proxied-default-rc2.

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR swaps the dashboard's rig-store probe from `bd doctor --readonly` to `bd ping --db <path> --json` — the only probe that works for embedded, direct-server *and* proxied-server stores — so proxied rigs stop being misreported, provider errors reach the dashboard as actionable problems instead of being discarded, and a stale `dolt-server.port` marker can no longer fake connectivity for a store that is not in server mode. It also closes a real hazard the old probe had: `bd ping` opens a provider, so the PR adds a marker gate that refuses to probe an unconfigured `.beads` rather than letting the health sampler conjure a store into existence.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **The error-truncation cap was enforced on two of three fold paths.** `maxProbeErrorRunes` exists so a chatty provider cannot bloat the snapshot the dashboard polls every 5 minutes, and the stderr and note paths both honored it — but `parsePingCheck` folded `payload.Error` in uncapped, which is precisely the path where the provider is healthy enough to be verbose. The final patch applies `truncateRunes(message, maxProbeErrorRunes)` there and pins it with `TestProbeRigTruncatesPingErrorEnvelopeMessage`; a `go test -overlay` mutation probe that removed the new call reddened exactly that one test, so the guard is not vacuous.
- **The `execBdPing` safety rationale named the wrong guard.** The comment argued the store-conjuring hazard was bounded because "`gc stop` stops the proxies last" and the sampler dies with the supervisor. Both legs are false: `cr.svc.Close()` closes the workspace-service proxies *before* sessions stop, and the supervisor hosting the sampler serves every registered city, so it outlives any one city's `gc stop`. The conclusion held, but through a different mechanism — `refresh` abandons the tick when a city's status read fails, so a stopped city is never reached by the probe fan-out. The comment now names that early return as the load-bearing guard and warns against weakening it, which matters because a future reader trusting the old reason could have removed the thing actually protecting them.

Review findings addressed (all in `716819154a fix(dashboard): cap ping error envelopes and correct the probe rationale`):
- *Error Handling & Resilience* / minor — `maxProbeErrorRunes` not enforced on the JSON error-envelope path; now capped and test-pinned.
- *Resource Lifecycle & Cleanup* / nit — the `gc stop` ordering clause in the `execBdPing` rationale did not describe what the code does; replaced with the real `refresh` early-return guard plus the residual in-flight window.
- *Contract & Interface Fidelity* / minor — `RigStoreHealth.doltConnected` still documented the retired `bd doctor` null-semantics; reworded to the semantics the tests actually pin (TCP result for persisted server-mode stores, ping connectivity otherwise, `null` only when the probe produced no signal).
- *Debuggability & Operability* / nit — `benignCheckCategories` and `rollupFor`'s `"warning"` arm are unreachable now that `bd ping` synthesizes a single `"Beads"` check that is only `ok` or `error`; kept them forward-compatibly with a comment saying so, rather than deleting the multi-check shape.
- *Debuggability & Operability* / nit — `cleanEnv`'s "read-only probes" wording collided with the new `execBdPing` doc explaining that ping is *not* passive; the property that actually justifies not forwarding `GITHUB_TOKEN` is that the probes are credential-free, so it now says that.

What the review validated rather than changed: the marker gate is a faithful superset of bd's own `hasBeadsProjectFiles` (same `vc.db`/`.backup` exclusions), so it cannot false-"down" a store bd would recognize; the store-conjuring premise was independently established three ways, including an empirical reproduction on the deployed bd; every arm of the fail-closed mode ladder (proxied / embedded / unknown / malformed / absent) is covered by a dedicated test; and the `issueCount` removal was traced end to end (Go struct → wire test → TS type → renderer → SPA test → `dist` rebuild).

- `ga-geua0` — *Change Impact / Blast Radius* / minor: suspended rigs are still probed with a store-opening ping. The supervisor reports `suspended` per rig but `statusBodyParsed.RigDetails` drops it, so the fan-out is unconditional — which contradicts the policy `cacheColdRigs` encodes elsewhere. Deferred because skipping the ping is a behavior change, and suspension is a policy flag rather than a teardown, so in the common case there is nothing to resurrect.
- `ga-geua0` — *Architectural Consistency* / minor: `probeRig` doubled in place (cyclomatic 7 → 13, 47 → 94 lines) as the honest price of the three new fail-closed arms. Deferred as a behavior-neutral extraction, so the next probe change does not land on a 94-line function.
- `ga-geua0` — accepted residual: `hasBeadsStoreMarker` hand-mirrors bd's marker predicate and `readDoltMode` deliberately diverges from `contract.ReadDoltMode`; worth a cross-link comment so future drift in bd's marker set is discoverable.
- *Release Safety* / minor — removing the `issueCount` wire field throws in tabs left open across a supervisor upgrade (`undefined !== null` passes the old guard, then throws on `.toLocaleString()`). This one is **accepted with eyes open, not deferred**: the throw is contained by the per-route `ErrorBoundary` to a blanked Health route and a reload heals it, and the alternative — a one-release `"issueCount": null` deprecation shim — is itself a wire change. The reasoning is recorded in `ga-geua0` so it is not re-litigated later.

### Maintainer Changes
One maintainer commit pushed to this branch: `fix(dashboard): cap ping error envelopes and correct the probe rationale` — 4 files changed, 49 insertions(+), 14 deletions(-). Your 15 commits were replayed unchanged on the current base with your authorship preserved.

### Final Review Status
Ready for the merge queue: final head `716819154a` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/34634106925

### Review Iterations
1 review pass performed — three independent reviewer lanes plus synthesis, which returned `approve` with no blocker and no major, followed by one maintainer fixup commit for the findings above.

---

PR #6031, opened 2026-09-05
fix(dashboard): fail closed on bd ping execution errors

julianknutsen:
Fixes ga-1r1g7.

Execution-level bd ping failures (timeout, cancellation, spawn failure) now synthesize a typed Database Connectivity error, yielding a down rollup with sanitized diagnostics. Adds front-door tests covering all three failures and verifies probes do not mutate .beads.

Checks:
- go test ./internal/api/dashboardbff -count=1 -timeout 180s
- go test -race ./internal/api/dashboardbff -run execution/failure tests -count=1
- go vet ./internal/api/dashboardbff
- git diff --check

---
- Related to #4902 — replaces the per-rig `bd doctor` probe with a single `bd ping` connectivity check, which removes both symptoms described there: the panel no longer emits the total-row count that was rendered as "Live issues" (issueCount is now always absent, so the row disappears) and no longer surfaces the benign dolt uncommitted-change checks that were flipping rigs to warn; it does not add a real live/non-closed count, and the rollup still escalates any non-ok check outside the benign categories

September 19 audit: rebased onto main as a focused execution-error patch. The shared bd-ping health change is already upstream through #6023, so duplicate #6001 is closed. Full dashboard API tests, vet, dashboard CI, and the local preview smoke pass on the rebased code; remaining repository checks gate requeueing.


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277302921

julianknutsen:
Preflight found substantial branch drift that must be resolved before review.

The PR describes a focused dashboard bd-ping failure-handling fix, but the branch is 83 commits ahead of `main`, changes 219 files (+11,738/-824), and includes unrelated beads, pool, dependency, runtime, CI, and documentation commits. Please rebase/retarget the PR onto `main` with only the intended dashboard changes, then request review again.

quad341:

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-699284130204932765

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR makes the dashboard's rig
health probe **fail closed** when the `bd ping` subprocess itself fails —
timeout, cancellation, or spawn failure — by synthesizing a typed Database
Connectivity error so the rollup reports `down` with sanitized diagnostics. That
matters because those were exactly the cases where a probe that never produced an
answer was indistinguishable from a healthy one: an operator looking at the panel
saw green for a store the dashboard had never actually reached. The PR also adds
front-door tests for all three failure modes and pins the invariant that probing
never mutates `.beads`.

branch.

### Original PR Review
Decision: approve (after one `request_changes` iteration).

Specific gaps fixed:
- The `cancellation` test was racing and its assertion was substring-based, so it
  could be satisfied by the *"canceled before start"* message without ever
  reaching the new post-spawn branch — it proved much less than it appeared to.
  It is now split into `cancellation before start` (saturates the runner
  semaphore so only the `ctx.Done()` arm is ready) and `cancellation after spawn`
  (the fake `bd` execs a real sleep so the kill lands on the process), each
  asserting its exact message. A mutation probe that deletes the
  `context.Canceled` branch now fails on **every** run, where before it failed
  roughly half the time.
- Iteration 2 caught a live consequence of the new branch: it tested
  `cctx.Err() == context.Canceled` *before* looking at `runErr`, so a child that
  had already completed with full stdout was discarded as an error. Because
  `localToolVersions` memoizes probe results — error rows included — for 45
  seconds, one disconnecting client could pin an `unavailable` tool-version
  snapshot for every other client. The final patch adds the `runErr != nil`
  conjunct, so a child that completed keeps its result and a child that was
  actually killed keeps the new message.
- The no-mutation assertion was using a new helper that was a weaker duplicate of
  the existing `snapshotProbeDir`; it now uses `snapshotProbeDir` plus a deep
  compare that walks recursively and checks mode and contents, so a file conjured
  inside `.beads/embeddeddolt` would be caught.
- Local gates re-run green on the final patch rebased onto current `main`:
  `go vet` clean and `go test ./internal/api/dashboardbff/ -count=1` passing.
  During review, a `-count=50` soak of the reworked cancellation rows also ran
  green (200 subtest runs, no failures).

Review findings addressed (all in `28fbf946`):
- **[Test Evidence Quality · major]** `cancellation` subtest was largely vacuous —
  rewritten into two deterministic rows with exact-message assertions.
- **[Change Impact / Blast Radius · minor]** New `context.Canceled` branch turned
  a completed probe into a 45s-cached `unavailable` — gated on `runErr != nil`.
- **[Architectural Consistency · minor]** New test helper duplicated
  `snapshotProbeDir` — helper removed in favour of the existing one.
- **[Test Evidence Quality · nit]** Timeout override was keyed on the table row's
  display name — hoisted into a `pingTimeout` table field.
- **[Test Evidence Quality · nit]** "Exactly one of
  `wantMessage`/`wantMessagePrefix`" was documented but not enforced — a guard now
  fails any row that sets neither.

One iteration-1 checklist item was correctly **declined** rather than applied:
deleting `equalStrings`. It turned out to be pre-existing and still consumed by an
untouched test, so removing it would have widened the PR into unrelated code.

- `ga-skixz` *(pre-existing, already open)* — **[Resource Lifecycle · minor]**
  `run()` never sets `c.WaitDelay`, so a forking `bd` can keep `c.Run()` blocked on
  the inherited stdout pipe and pin an exec slot after the direct child exits.
  Pre-existing, deliberately left out of scope here.
- `ga-a7xma` — **[Architectural Consistency · nit]** `execBdPing`'s `timeout <= 0`
  fallback is unreachable: every construction path goes through `newExecRunner`,
  which always sets the field. Documentation is the preferred remedy over deletion.
- `ga-s088b` — **[Debuggability & Operability · nit]** a cancelled run is
  classified as `execErrSpawn` even though nothing failed to spawn. No dispatch
  consequence today; worth an `execErrCanceled` kind once a consumer needs to
  branch on it.

One coverage note for the record: the Gemini reviewer was skipped for this run, so
no reviewer re-examined the SPA rendering of the `note` → `problems` move. No TS
file is in the diff, and the earlier iteration validated that move against the
shared TS contract.

### Maintainer Changes
`28fbf946` — *fix(dashboard): report exec cancellation only for interrupted runs*
(2 files changed, 105 insertions(+), 40 deletions(-)). Your original commit is
preserved underneath it.

### Final Review Status
Ready for the merge queue: final head `28fbf946` has passing required GitHub

CI: https://github.com/gastownhall/gascity/actions/runs/35540387484

### Review Iterations
2 review passes performed: iteration 1 returned `request_changes` on the test
evidence for the new cancellation path, and iteration 2 approved after the fixup,
independently re-verifying each fix rather than taking the commit message for it.

---

PR #6032, opened 2026-09-05
fix(session): give drain-ack stop-pending a terminal exit — escalate wedged seats to a fenced, subreaper-aware force-termination

julianknutsen:
## Summary

A seat in `state=draining` / `state_reason=drain-ack-stop-pending` whose agent acknowledges the drain but whose **runtime survives the stop** has no exit: `finalizeDrainAckStopPendingSessions` only closes a row when the runtime observes dead, and while `obs.Running || obs.Alive` it reminds and re-queues forever — its own comment concedes the only historical exit is "an operator killing the pane". Because an OPEN bead owns its `session_name`, each wedged seat also permanently holds a pool slot name, so pools report `session name unavailable` and cannot mint replacements. Production: 7→16 seats wedged for 2+ days on one city, ~100 ready steps undispatched, 62 session rows stuck pending.

Two prior non-fixes are documented in the commits: the reminder budget is **suppressed** for agent-acked panes (the ack pin skips before any marker write, so a budget-only gate never fires for exactly the incident population), and `worker.Handle.Kill` is `Stop` minus the bead write — the same signal that already failed (tmux's terminate path discards kill errors and never re-observes; filed separately).

## The fix

**P0** — a two-arm terminal escalation (reminder-budget arm for askable seats; a `drain_at`-bounded arm for agent-acked seats), running **off-tick** with durable fail-closed pacing, that routes to `runtime.ProcessTableScanner` PID-level termination and falls through to the existing close terminus (one close path). The kill is fenced five ways: wide assignee-identity set (alias + alias_history), instance-token fence, positive runtime-binding attribution, attach/recent-activity holds, per-wave PID re-validation. **Attribution never trusts session-ID + city alone** — `GC_SESSION_ID` is inherited by every child the agent ever spawned, and the traced counterexample (the managed-Dolt scope watchdog, whose SIGTERM handler stops the city's *shared* dolt server) is a mandatory negative pin at **both** reparent topologies: init *and* the `systemd --user` child-subreaper this fleet actually runs (`internal/pidutil.IsReparentedOrphan`, extracted from the model `workspacesvc` already had). The scan-layer envelope exclusion is pinned separately in proctable.

**P1** — city stop/restart no longer chokes on draining seats: a distinct shutdown-intent seam (`SessionHandle.StopForShutdown` / `Manager.SuspendForShutdown`) used only by the shutdown sweep; targeted operator `suspend` keeps returning the illegal transition, and teardown errors propagate.

Events report **outcomes, not decisions** (killed / skipped-attribution / nothing-found), so "nothing reachable to kill" — the condition this pass exists to surface — is distinguishable from success.

## Review

Design → implementation → three adversarial review rounds + one focused coupling check: **4 criticals, 8 majors, 5 minors — all fixed**, including: the escalation being unreachable for the exact agent-acked population it was written for (the test fixture was green on the population's complement); "certified kill" not existing as a mechanism; the kill set sweeping in every env-inheriting descendant (the Dolt-watchdog shape); the shutdown latitude leaking to operator paths; and the `PPID <= 1` reparent test being blind to the subreaper topology. Every kill-safety invariant is mutation-pinned — each listed mutation fails exactly its own pin.

Refs: ga-rxhu2, ga-bh1jz

---
- Related to #6316 — guards its new force-termination path against the same hazard that issue describes — the managed Dolt scope watchdog inherits a session's GC_SESSION_ID and reparents to a subreaper, so this PR refuses reparented orphans as kill targets (internal/pidutil.IsReparentedOrphan, at both the init and `systemd --user` topologies) and pins that refusal with a test naming the watchdog and the shared dolt sql-server; it does not change the process-table orphan sweep or the inherited session environment that causes the reported deaths
- Related to #4448 — addresses the drain-acked seat whose runtime survives the stop and keeps holding its pool slot name — the "acked but still STATE=active minutes later" symptom noted in that issue — by escalating wedged seats to a fenced PID-level termination so the existing finalizer can reap them; it does not add a reaper for session beads whose runtime is already gone, and it does not change how open_count is derived


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277302852

quad341:

quad341:
@julianknutsen, a correction to our 09-08 comment, and I'm sorry about it. It said this PR was past a 20,000-line ceiling and asked you to split it. That was wrong. Roughly half of the diff is regenerated SPA and OpenAPI artifacts, and what tripped our reviewer was a separate 1 MB size limit that those generated files filled. Please don't split it.

Our reviewer has since read the hand-written Go end to end. I'll follow up here with the review.

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-699284130204932664

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR gives drain-ack stop-pending sessions a terminal exit: a seat that acknowledges the drain but never actually leaves is now escalated to a fenced, subreaper-aware force-termination instead of wedging indefinitely, and `gc city stop` no longer chokes on draining seats. That matters because a single unresponsive agent could previously hold a pool slot name — and a whole city shutdown — forever, with nothing in the event stream saying so.

### Original PR Review
Decision: request_changes at iteration 1, resolved to approve at iteration 2.

Specific gaps fixed:
- **Unrecovered panic in the detached termination goroutine.** The goroutine that performs the kill had no `recover()`, so a panic inside it would have taken down the whole controller process and left the termination slot claimed. It now recovers, prints the panic with `debug.Stack()`, and calls `done()` inside the same deferred closure — the ordering `queueDrainAckAsyncStop` already used. `TestEscalationRecoversAPanicInTheDetachedTermination` crashes rather than fails without it.
- **The escalation arms dropped `poke(cityPath)`.** A successful force-kill left the pool name held for up to a full patrol interval. The poke seam is now bound at queue time on the caller's goroutine and invoked on the `stopped_without_force` arm and after the forced-outcome record; two new tests count pokes per arm.
- **The subreaper pid came from the controller's own ancestry.** On a split topology (controller started from an ssh shell or a system unit while the tmux server descends from `systemd --user`), detection returns nothing and the managed-Dolt watchdog fence silently thinned to the name/city checks. The fix replaces detection with a positive fence — `runtime.LiveRuntime.ParentIsProviderInfrastructure`, set by both the Linux and Darwin scanners and consumed as a refusal arm — so an unreadable parent, an exited parent, and an unrecognized parent all refuse the kill. Both review lanes ran independent `go test -overlay` mutation probes neutralizing that arm, and both turn `TestEscalationNeverTerminatesAnOrphanWhoseSubreaperWentUndetected` red, so the pin is not decorative.
- **Gate ordering contradicted its own doc, and permanently-held rows were never paced.** The quiet hold now runs ahead of the cross-store assigned-work fan-out, so a held row no longer re-pays that fan-out on every reconcile tick; `TestEscalationQuietHoldRefusesBeforeTheAssignedWorkFanOut` asserts zero assignee-filtered `List` calls for a held row. The pacing marker is written only after the termination slot is actually claimed, and the slot is released on the unpaced-refusal path.

Review findings addressed:
- The remaining doc- and contract-fidelity minors landed in the same maintainer fixup `fe0a77be7d`: the quiet hold is re-validated immediately before the destructive act (a distinct `held_late` outcome, alongside the instance-token re-check); `SessionDrainStopEscalated`'s doc no longer claims the bead is closed or that a fired event means force landed — it is now outcome-keyed and deliberately non-enumerative, so adding a new outcome arm cannot falsify it again; `ParentIsProviderInfrastructure`'s doc is narrowed to what the predicate actually checks (a command-name match on the parent, matching any tmux on the host, not socket-scoped to one provider's server); and the return-value doc now states the real caller contract — a true return means this tick's stop is owned — matching its single call site.
- One iteration-1 nit was withdrawn rather than applied: the Darwin `LiveRuntime.Name` argv concern rested on a misreading, since `darwinPSCommand` already yields argv[0]'s basename. Both lanes independently confirmed the contract was already met at the reviewed head, so no change was owed and none was made.

- **ga-h9by6** — [Resource Lifecycle & Cleanup, minor] `LiveRuntime` carries no scan-time start identity, so `KillByPID` binds process identity at kill entry rather than at scan time, leaving an outer scan→kill PID-recycle window open. Not addressed in this PR: the review priced it in as explicitly optional, and this PR's new positive parent fence incidentally narrows that window (a recycled PID would also have to have been scanned with a tmux parent) rather than widening it.
- **ga-rdp7p** — [Contract & Interface Fidelity, minor] `ParentIsProviderInfrastructure` is a tmux command-name match, not provider-scoped attribution. The documentation half was fixed here (above); actually socket-scoping the predicate was deliberately left out of this PR, because narrowing the shared `isInfrastructureCommand` would regress the #5392 never-report-the-server rule, and socket-scoping at the assignment site would mean changing the `ProcessTableScanner` interface and both platform scanners.

### Maintainer Changes
One maintainer fixup commit on top of your five preserved commits:
- `fe0a77be7d` fix(session): recover, poke, and positively fence the drain-ack escalation — 9 files changed, 616 insertions(+), 109 deletions(-)

The branch was also rebased onto current `main` during adoption. That rebase hit 24 conflicts — all in generated artifacts (22 of them rename/rename in the built SPA assets, plus `dist/index.html` and the generated supervisor client), zero in source. They were resolved by regenerating from the merged source rather than by picking a side, and a follow-up `make dashboard-check` rebuild left no modified files.

### Final Review Status
Ready for the merge queue: final head `fe0a77be` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/35814169859

### Review Iterations
2 review passes. Iteration 1 returned request_changes with one blocker and three majors; a single maintainer fixup commit addressed them; iteration 2 re-derived every item against the applied code — including independent mutation probes on the subreaper fence — and approved, with the two remaining doc-contract minors applied before merge.

---

PR #6035, opened 2026-09-05
fix(bd): route unserved by-ID READS on residence, not on servability

### LINKED ISSUE #6015: bug(bd): the by-ID funnel gate asks "can I SERVE this spelling" instead of "does this argv address a subject by id", so `gc bd show --long <id>` of a relocated-class bead answers from the work store's retained pre-migration copy
## Summary

On a converged split-storage city, `gc bd`'s in-process by-ID surface decides whether to enter
its class front door by asking **"is this a spelling I can SERVE?"** — which is a different
question from **"does this argv address a subject by id?"**

The consequence is that one bead answers two ways, with no diagnostic on either:

```console
$ gc bd show cr-dqv6l --json | jq '{status, assignee, outcome: .metadata."gc.outcome"}'
{ "status": "closed", "assignee": "manager-2-pool", "outcome": "pass" }

$ gc bd show --long cr-dqv6l
○ cr-dqv6l · Manager: track atoms, refine bounces (max 2 rounds), assemble, close root   [P1 · OPEN]

$ gc bd show cr-dqv6l cr-jhdkg --json | jq '.[0] | {status, assignee, outcome: .metadata."gc.outcome"}'
{ "status": "open", "assignee": null, "outcome": null }
```

`--long` is the sharp edge: **one ordinary presentation flag on a single-id read silently
changes which store answers**, from the binding's live row to the work store's retained
pre-migration copy, frozen at migration time.

`cmd/gc/cmd_bd_by_id.go`'s own file header names this exact outcome as the thing the surface
exists to prevent:

> `gc storage migrate` preserves ids — so a bead the migration relocated keeps its HQ/rig-era
> prefix and the resolver returns nil for exactly the ids that moved. **Only a residence probe
> reaches those, and without it their reads are answered from the work store's retained
> pre-migration copy, frozen at migration time.**

The residence probe exists and is correct. It is simply gated on **MUTATION argv only**, so an
unserved READ is neither served nor refused — it falls through to the `bd` subprocess on the
work store, which answers wrongly, in `bd`'s own shape, exit 0.

Measured on a live city at pin `893eb0e03` (2026-09-04), and the gate is unchanged on `main`
`8de6b46df`.

## Reproduction

Any city with `[storage.classes] graph = "<binding>"` past `gc storage migrate`, and any
graph-class bead whose id was minted before the migration (so it carries the work prefix and
`IDInNamespace` cannot see it).

| spelling | store that answers | status reported |
| --- | --- | --- |
| `gc bd show cr-dqv6l` | class binding (served in process) | **closed**, `gc.outcome=pass` |
| `gc bd show --long cr-dqv6l` | work store (`bd` subprocess) | **open**, no outcome |
| `gc bd show cr-dqv6l cr-jhdkg` | work store (`bd` subprocess) | **open**, no outcome |

Control — an ordinary work-class bead (`cr-wd22i`) reads identically under every spelling, which
confirms the affected population is exactly the relocated-class residents.

## Mechanism

Line numbers at `main` `8de6b46df`.

1. **The funnel-entry gate asks about servability, not about addressing.**
   `maybeRouteBdByID` (`cmd/gc/cmd_bd_by_id.go:859`) computes three arms and returns
   fall-through when all three miss (`:863`):

   ```go
   op, served := parseBdByIDOp(bdArgs)
   named, namesClassBead := bdArgsNameClassOwnedBead(bdArgs)
   mutationIDs := bdByIDMutationSubjects(bdArgs)
   if !served && !namesClassBead && len(mutationIDs) == 0 {
       return 0, false
   }
   ```

2. **Arm 1 misses.** `parseBdByIDPositional` (`:648`) accepts exactly one positional and
   rejects unknown flags on purpose — "a flag it silently ignored would change the meaning of
   a command it then claimed to have executed" — so `--long` and a second positional both
   return `ok=false` (`:651-652`).

3. **Arm 2 misses.** `bdArgsNameClassOwnedBead` (`:1144`) proves ownership from a reserved
   prefix. The migration **preserves ids** (`cmd/gc/infra_class_migrate.go:3-11`), so a
   relocated bead keeps its work prefix and ownership-by-spelling structurally cannot see it.

4. **Arm 3 is never asked.** `bdByIDMutationSubjects` (`:826`) is built on
   `bdMutationWriteIDs` (`cmd/gc/cmd_bd.go:637`), whose subcommand set is
   `update|close|reopen|delete|heartbeat` (`:641-645`). `show` is not in it, so the residence
   probe — the only mechanism that can reach a work-prefixed class resident — yields nothing
   to probe.

5. All three miss, the binding is never opened, and `gc bd` hands the command to the `bd`
   subprocess pointed at the work store, which still holds the retained source row
   (`infra_class_migrate.go:3-11`: the migration copies and tests equality; it does not remove).

The asymmetry is visible in the code as a plain inconsistency: an unserved **mutation** of the
same bead (`gc bd close cr-jhdkg cr-dqv6l`) yields ids, probes residence, and is refused by
`refuseUnservedClassMutation` (`:896`). The identical subject in a **read** verb is not.

`internal/storeref/resolve.go:24-28` states the intended contract for exactly this case —
*"ByID, id outside every reserved namespace: [binding as a residence probe, work, shadows] …
nil from the namespace gate means 'cannot decide', never 'work owns it'"*. The CLI surface
honours that for served reads and for mutations, and drops it for unserved reads.

## Why this is not a duplicate of #5987

#5987 reports that the two co-resident **copies diverge**, because by-id writes land on the
binding while dispatch reads the work row; its fix candidates all operate on the storage layer
(converge on write / tombstone the source rows / invert demand precedence). It uses the multi-id
read only as a *diagnostic* for that divergence.

This issue is the CLI routing gap that makes the diagnostic possible, and it is orthogonal:

- **It is not fixed by #5987's candidates.** Under candidate A the two spellings still read two
  different stores; they merely happen to agree until the next drift, and a field with no
  representation on one side still reads differently. Under candidate B the unserved spelling
  reads a store that no longer holds the row, which is a *different* wrong answer, not the right
  one. The invariant this surface declares — *"an id whose owning class is the relocated binding
  is answered HERE"* — stays violated either way.
- **It does not fix #5987.** Routing the read correctly does not stop the copies diverging.
- **Different file, different blast radius.** One line of `cmd/gc/cmd_bd_by_id.go`'s entry gate
  versus the migration/resolution layer.

They should land independently; this one is the cheaper and strictly narrower of the two.

## Fix candidates

**A. Ask the funnel-entry question the surface actually means: does this argv address subjects
by id, read or mutation alike?** Reuse the existing positional scanner for the by-ID READ verbs
`parseBdByIDOp` already serves (`show`, `dep list`, `dep tree`), and pass those ids to the same
residence probe the mutation arm uses. An unserved read of a class resident then reaches the
existing refusal instead of the wrong ledger; an unserved read of ordinary work beads still
falls through to `bd` byte-identically. Ambiguous scans (an unrecognized flag that might consume
the next token) stay outside, fail-closed, so nothing unrecognized ever becomes a probed id.
Selectors (`list`, `query`, `search`) stay outside by construction — they QUOTE ids rather than
address them, and `bd_relocated_classes.go` refuses those on its own terms.

Cost is bounded by the served form already paying it: `show <id>` enters the funnel today, so
`show --long <id>` entering too adds no population beyond the spellings whose answers were wrong.

**B. Serve the unserved read spellings in process.** Implement `--long`, multi-id and the rest
of `bd show`'s flag surface on the class front door. Strictly more work, and it re-opens the
"a flag silently ignored changes the meaning of a command" hazard `parseBdByIDPositional`
deliberately closed. Worth doing eventually for `--long`; not the fix for the wrong-answer lane.

**C. Refuse every unserved by-ID read unconditionally on a relocated city.** Simplest, but it
breaks `show --long` for the ordinary work beads that are the overwhelming majority, to fix a
case the residence probe can already decide precisely.

I'd suggest **A**. I have it implemented with regression tests (both the wrong-answer cases and
the passthrough-preserved cases) and will open a PR against this issue.

## Prior art checked

- #5987 — the storage-layer divergence this read gap exposes; orthogonal, see above.
- #5148, #5158, #5161 — established the co-resident leg orders in `storeref`, including the
  deliberate ByID/AssignedWork asymmetry. They set the contract this surface is failing to honour
  for one argv shape; they did not touch the CLI entry gate.
- #5128, #5132, #5245 — the reserved-prefix advisory/probe rules the ownership-by-spelling arm
  implements. Correct as written; they simply cannot see a work-prefixed relocated bead.
- #5170 — `gc bd list/ready` answering from one undisclosed store. Adjacent, but that is the
  SELECTOR surface, which this issue deliberately leaves outside.

jacobhausler:
## Summary

On a city whose coordination classes are relocated, `gc storage migrate` preserves bead ids and retains the work store's pre-migration copy. The by-ID surface's funnel-entry gate asked whether a spelling could be **served**, which is a different question from whether the argv **addresses a subject by id**: `show <id>` entered and reported the binding's live row, while `show <id> <id>` and `show --long <id>` addressed the same subject, missed the served parser, and fell through to a `bd` subprocess pointed at the work store — which answered from the retained copy, in bd's own shape, with no diagnostic. Two spellings of one read then disagreed about one bead's status.

This widens the funnel-entry gate to every argv that addresses subjects by id, read or mutation alike, so an unserved read of a class resident reaches the existing refusal instead of the wrong ledger. Selectors and ambiguous scans still pay nothing.

Fixes #6015.

- `bdMutationWriteIDs`'s positional-argv scanner is factored out as `bdScanPositionalIDs` so it can be reused for reads.
- `bdByIDReadSubjects` / `bdByIDReadVerbs` extend the same scan to `show`, `dep list`, and `dep tree` — the full set of by-ID read verbs `parseBdByIDOp` serves.
- `bdByIDAddressedSubjects` merges mutation and read subjects into one funnel-entry question.
- `refuseUnservedClassMutation` is renamed `refuseUnservedClassTarget` to reflect that it now answers reads too.

## Test plan

- [x] `make build`, `make fmt-check vet`, `make lint-full` — all clean (0 issues repo-wide)
- [x] `go test ./cmd/gc/ -count=1` — full package passes except 5 pre-existing failures verified identical on unmodified `main` (process-group signal-handling tests that are sandbox/environment-dependent, unrelated to this change: `TestRunProviderOpKillsProcessGroupOnTimeout`, `TestRunProviderOpWithEnvContextParentCancellationKillsProcessGroup`, `TestRunProviderProbeKillsProcessGroupOnTimeout`, `TestPackV2ImportsScript/pack-v2-imports`, `TestHookTimeoutKillsTheWholeProcessGroup/group_cleanup_reaps_the_descendant`)
- [x] New regression tests `TestBdByIDUnservedReadOfAClassResidentIsRefusedNotAnsweredByTheWorkStore` and `TestBdByIDUnservedReadOfAWorkBeadKeepsThePassthrough` fail to exist / pass correctly on `main` vs the branch
- [x] Disposable fail-before repro (`gateFallsThrough`, argv-only, not included in this PR) confirms the routing gate falls through on unpatched `main` for `show <id> <id>`, and that the mutation arm was already protected

### DIFF
--- a/cmd/gc/cmd_bd.go
+++ b/cmd/gc/cmd_bd.go
@@ -644,7 +644,18 @@ func bdMutationWriteIDs(args []string) (ids []string, ok bool, ambiguous bool) {
 	default:
 		return nil, false, false
 	}
+	ids, ambiguous = bdScanPositionalIDs(sub, args[1:])
+	return ids, true, ambiguous
+}
 
+// bdScanPositionalIDs is the argv scan bdMutationWriteIDs is built from, taken
+// on its own so a caller that has already resolved the subcommand can reuse it.
+//
+// sub names the bd subcommand whose flag manifests decide which tokens are
+// values ("close", "dep list"); rest is the argv AFTER it. ambiguous reports an
+// unrecognized flag that might consume the next token, which every caller must
+// fail closed on — the token would otherwise be read as a bead id.
+func bdScanPositionalIDs(sub string, rest []string) (ids []string, ambiguous bool) {
 	// valueFlags is the complete set of flags that consume the next argument as
 	// their value for this subcommand, in both long and short form.
 	// Sourced from `bd <sub> --help` (2026-06-10).
@@ -655,8 +666,8 @@ func bdMutationWriteIDs(args []string) (ids []string, ok bool, ambiguous bool) {
 	boolFlags := bdSubcmdBoolFlags(sub)
 
 	positional := false // true after "--"
-	for i := 1; i < len(args); i++ {
-		arg := args[i]
+	for i := 0; i < len(rest); i++ {
+		arg := rest[i]
 		if positional {
 			if arg != "" {
 				ids = append(ids, arg)
@@ -696,9 +707,9 @@ func bdMutationWriteIDs(args []string) (ids []string, ok bool, ambiguous bool) {
 		}
 		// Unknown flag. It might consume a value argument that looks like a
 		// bead ID. Fail-closed: report ambiguity so the caller can reject.
-		return nil, true, true
+		return nil, true
 	}
-	return ids, true, false
+	return ids, false
 }
 
 // bdSubcmdValueFlags returns the set of value-consuming flag names (in
--- a/cmd/gc/cmd_bd_by_id.go
+++ b/cmd/gc/cmd_bd_by_id.go
@@ -145,19 +145,27 @@ package main
 // store census. That is once per process, memoized with every other one-shot
 // command's routing, but it is not free, so it is paid only by an invocation
 // that could concern a class-owned bead: a served by-ID form, an argv that
-// addresses a reserved-prefix id, or a MUTATION with unambiguous positional
-// ids. An ordinary work READ or SELECTOR never enters; a mutation addressing
-// ids enters exactly once per process.
+// addresses a reserved-prefix id, or an argv that ADDRESSES SUBJECTS BY ID with
+// an unambiguous positional scan — a mutation or a by-ID read alike. A SELECTOR
+// never enters; an argv addressing ids enters exactly once per process.
 //
-// The mutation arm is the widening, and the id shape is why it cannot be
-// narrower. A class store mints from its binding workspace's own prefix and
+// The addressed-subjects arm is the widening, and the id shape is why it cannot
+// be narrower. A class store mints from its binding workspace's own prefix and
 // `gc storage migrate` preserves ids, so "gc-123" says nothing about which
 // store holds the row — only a probe does, and skipping it is what sent every
 // unserved mutation of a class resident to the ledger that cannot hold it.
-// Reads and selectors stay outside because they address no subject whose
-// residence could decide anything: a selector QUOTES ids, and a read that this
-// surface serves is already inside. An ambiguous scan stays outside too, for
-// the plainest reason — it yields no ids to probe.
+//
+// READS are in the same arm, and gating them on servability instead was a real
+// defect rather than a hypothetical: `show <id>` is served and enters, while
+// `show <id> <id>` and `show --long <id>` address the same subject, are not
+// served, and used to fall through — so one flag or one extra id turned a
+// routed read into a work-store read, answered from the retained pre-migration
+// copy in bd's own shape with no diagnostic. Two spellings of one read
+// disagreeing about one bead's status is the wrong-answer lane this file
+// exists to close, and servability was never the question a READ has either.
+// Selectors still stay outside because they address no subject whose residence
+// could decide anything — a selector QUOTES ids. An ambiguous scan stays
+// outside too, for the plainest reason — it yields no ids to probe.
 
 import (
 	"encoding/json"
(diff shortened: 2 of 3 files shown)


jacobhausler:
Closing per our own adjudication on this contribution — not because the patch is wrong in principle.

Why now:

- `main` restructured the file this patch edits. `ea66aa0fe` (#6140) retired the hand-rolled by-id probes onto the resolver, and `830c130d2` (#5634, "let a clean class-binding miss fall through in gc bd — ownership is residence, not prefix") landed on the same door afterwards and answers the fall-through this PR set out to fix — by *refusing* an unserved, class-resident id rather than by serving the read from the retained pre-migration copy.
- The branch no longer merges: `mergeable=CONFLICTING`, `mergeStateStatus=DIRTY` against current `main`; the conflicting file is `cmd/gc/cmd_bd_by_id.go` (measured with `git merge-tree --write-tree origin/main 24b6de420`), ahead 1 / behind 202 (`gh api .../compare/main...24b6de420`). 0 reviews, no maintainer request outstanding.
- What the patch still argues — that by-id *reads* and selectors stay outside the residence probe — is the position `main` has now taken against, so a rebase would have to relitigate that rather than just move hunks.

What stays open, deliberately: this close does **not** close the underlying report, #6015, which is still open. Two things belong there and not here:

1. `main`'s own cost model keeps "an ordinary work read or selector" outside the residence funnel. Whether that residual can still answer a relocated id from the retained copy is code-level evidence read off `main`, not a built-binary reproduction — nobody has run the binary repro.
2. If the read-side behaviour is still wrong on a current `main`, that is a narrower report than this one and is better filed against current `main` than revived by rebasing this patch.

Nothing was pushed to the branch and no `main` history was touched; the branch is left in place.

PR #6041, opened 2026-09-05
fix(gc): pin invoking executable across Beads subprocesses

julianknutsen:
Beads subprocesses on the paths enumerated below now receive the physical path of the invoking GC executable in `GC_BIN`. Recovery, provider lifecycle, scoped stores, context runners, retries, default-rig initialization, and the store bridge reject executable-resolution failures before launching a child. Three managed-env paths are out of scope here and still inherit an ambient `GC_BIN` — the hook-claim mutation runner, the doctor store preflight, and the DoltLite early return; all three are pre-existing, untouched by this PR, and tracked in ga-wbycp.

The existing lifecycle test wrapper behavior is preserved. Unrelated exec providers keep an empty `GC_BIN`, and DoltLite retains its existing provider policy. This consolidates the context/bridge follow-up from #6046. The two independent GC_BIN commits have been extracted from the retired #6042 handoff stack and rebased onto main.

Sol and independent Astra reviewed the final implementation. Focused tests, targeted race tests, resource census, repository pre-commit lint/code generation/vet, and dashboard preview smoke passed. The complete fast baseline and all six CLI process shards plus product metrics passed. Dashboard build/type checks and a live preview HTTP smoke passed. The normal pre-push gate passed at d7637d1f3589034aceb060d70e1c853d24dd3c20.

Beads: ga-p9iuv.30.4 and ga-p9iuv.30.4.1.

Final-head GitHub Linux CI and Mac Regression passed. Mac make-test required one retry after an unrelated unchanged zcode terminal-input timing failure; all 12 Mac CLI process shards passed on the first attempt.

The validation above describes the original submission. The September 19 audit is revalidating the rebased head before adding it to the review queue.

### DIFF
--- a/cmd/gc/api_state.go
+++ b/cmd/gc/api_state.go
@@ -429,12 +429,15 @@ func (cs *controllerState) openRigStore(provider, rigName, rigPath, prefix strin
 	scopeRoot := resolveStoreScopeRoot(cs.cityPath, rigPath)
 	openExecStore := func() (beads.Store, error) {
 		s := beadsexec.NewStore(strings.TrimPrefix(provider, "exec:"))
-		env := gcExecStoreEnv(cs.cityPath, execStoreTarget{
+		env, err := gcExecStoreEnv(cs.cityPath, execStoreTarget{
 			ScopeRoot: scopeRoot,
 			ScopeKind: "rig",
 			Prefix:    prefix,
 			RigName:   rigName,
 		}, provider)
+		if err != nil {
+			return nil, err
+		}
 		if execProviderNeedsScopedDoltStoreEnv(provider) {
 			projected, err := bdRuntimeEnvForRigWithError(cs.cityPath, cfg, scopeRoot)
 			if err != nil {
--- a/cmd/gc/bd_env.go
+++ b/cmd/gc/bd_env.go
@@ -1025,7 +1025,7 @@ var recoverManagedBDCommand = func(cityPath string) error {
 	overrides := cityRuntimeEnvMapForCity(cityPath)
 	gcBin, err := resolveProviderLifecycleGCBinary()
 	if err != nil {
-		return fmt.Errorf("resolve invoking gc executable: %w", err)
+		return err
 	}
 	if gcBin != "" {
 		overrides["GC_BIN"] = gcBin
--- a/cmd/gc/bd_env.go
+++ b/cmd/gc/bd_env.go
@@ -48,6 +48,9 @@ func bdCommandRunnerForCity(cityPath string) beads.CommandRunner {
 func bdContextCommandRunnerForCity(cityPath string) beads.CommandRunner {
 	return func(dir, name string, args ...string) ([]byte, error) {
 		env := cityRuntimeEnvMapForCity(cityPath)
+		if err := pinBdGCEnvironment(env); err != nil {
+			return nil, err
+		}
 		bdBin, err := workspacePinnedBdBinary(cityPath)
 		if err != nil {
 			return nil, err
@@ -1020,6 +1023,13 @@ var (
 var recoverManagedBDCommand = func(cityPath string) error {
 	script := gcBeadsBdScriptPath(cityPath)
 	overrides := cityRuntimeEnvMapForCity(cityPath)
+	gcBin, err := resolveProviderLifecycleGCBinary()
+	if err != nil {
+		return fmt.Errorf("resolve invoking gc executable: %w", err)
+	}
+	if gcBin != "" {
+		overrides["GC_BIN"] = gcBin
+	}
 	if err := applyWorkspacePinnedBdBinary(overrides, cityPath); err != nil {
 		return err
 	}
@@ -1029,10 +1039,6 @@ var recoverManagedBDCommand = func(cityPath string) error {
 	applyBdContributorRoutingOptOut(overrides)
 	environ := mergeRuntimeEnv(processEnvSnapshotExcludingNativeDoltOpen(), overrides)
 	environ = append(environ, providerLifecycleDoltPathEnv(cityPath)...)
-	if gcBin := resolveProviderLifecycleGCBinary(); gcBin != "" {
-		environ = removeEnvKey(environ, "GC_BIN")
-		environ = append(environ, "GC_BIN="+gcBin)
-	}
 	return runProviderOpWithEnv(script, environ, "recover")
 }
 
@@ -1402,6 +1408,9 @@ func bdCommandRunnerWithManagedRetryErr(cityPath string, envFn func(dir string) 
 		if env == nil {
 			env = map[string]string{}
 		}
+		if err := pinBdGCEnvironment(env); err != nil {
+			return nil, err
+		}
 		ensureProjectedDoltEnvExplicit(env)
 		runner, runnerErr := beadsCommandRunnerForHostedCity(cityPath, env)
 		if runnerErr != nil {
@@ -1423,6 +1432,12 @@ func bdCommandRunnerWithManagedRetryErr(cityPath string, envFn func(dir string) 
 		retryEnv, retryEnvErr := envFn(dir)
 		if retryEnvErr != nil {
 			return nil, retryEnvErr
+		}
+		if retryEnv == nil {
+			retryEnv = map[string]string{}
+		}
+		if err := pinBdGCEnvironment(retryEnv); err != nil {
+			return nil, err
 		}
 		ensureProjectedDoltEnvExplicit(retryEnv)
 		retryRunner, runnerErr := beadsCommandRunnerForHostedCity(cityPath, retryEnv)
--- a/cmd/gc/bd_env_test.go
+++ b/cmd/gc/bd_env_test.go
@@ -130,7 +130,11 @@ func TestBdCommandRunnerForCityCompleteStorageBindingSkipsManagedRetry(t *testin
 	}
 }
 
-func TestBdContextCommandRunnerForCityPinsCanonicalGCBinary(t *testing.T) {
+// TestBdCommandRunnerForCityCompleteBindingPinsCanonicalGCBinary drives
+// bdCommandRunnerForCity's complete-binding arm, which delegates to
+// bdContextCommandRunnerForCity. The managed-retry arm is covered by
+// TestBdCommandRunnerWithManagedRetryPinsCanonicalGCBinaryOnBothEnvArms.
+func TestBdCommandRunnerForCityCompleteBindingPinsCanonicalGCBinary(t *testing.T) {
 	t.Setenv("GC_BIN", "/tmp/stale-gc")
 	cityPath := t.TempDir()
 	bdPath := filepath.Join(t.TempDir(), "bd")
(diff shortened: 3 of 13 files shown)


julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR closes a real wrong-binary hazard: it pins the physical path of the invoking `gc` executable into `GC_BIN` across five Beads-subprocess paths — the recovered provider, the bd context runner, the managed-retry runner, default-rig initialization, and the store bridge — and makes executable-resolution failures stop *before* a child is launched instead of silently falling back. That matters because the provider script prefers `$GC_BIN` and otherwise falls back to whatever `gc` happens to be first on `PATH`, so a stale or cross-city ambient value could previously steer a recursive `gc` invocation at the wrong installation. Removing the `os.Executable` → `exec.LookPath("gc")` → `""` fallback chain — rather than leaving it reachable — is exactly the right shape of fix.

### Original PR Review
Decision: approve. No blockers and no majors. Three independent reviewer lanes plus a synthesis pass covered every changed file, and each of the PR's own claims was verified first-hand rather than taken at face value — the leak was reproduced end-to-end with an ambient `GC_BIN`, the fail-stop path was proven to never launch the child, and the pin was traced through every later env-layering step to confirm no downstream clear-list clobbers it.

Specific gaps fixed:
- **The managed-retry pins were unasserted.** The only pin test drove `bdCommandRunnerForCity`'s complete-binding arm, so both `pinBdGCEnvironment` calls in `bdCommandRunnerWithManagedRetryErr` could be deleted with nothing turning red — confirmed by mutation probe. The branch now carries a test that drives a real bd shim recording the child's `GC_BIN` on each attempt and fails the first attempt the way an unreachable managed server does, so the retry-env arm is genuinely reached; deleting either pin now reddens its own attempt.
- **A test name promised coverage it did not deliver.** `TestBdContextCommandRunnerForCityPinsCanonicalGCBinary` never called `bdContextCommandRunnerForCity`, so a future change to the delegation predicate could have silently dropped the coverage the name advertised. It is renamed to name the entry point it actually exercises.
- **Doubled error prefix on the newly user-visible failure path.** The resolver already prefixes its `os.Executable` branch with `resolve invoking gc executable:`, and four outer sites re-wrapped the identical phrase — so operators would read it twice, in exactly the path this PR surfaces for the first time. Those four now return the error bare, matching the five pin sites that already did.
- **The scope claim was broader than the change.** The PR body's lead sentence read as categorical over all Beads subprocesses; it is narrowed to the five paths actually pinned, and the same overstatement in `ga-p9iuv.30.4`'s close reason has been corrected.

Review findings addressed:
- Test Evidence Quality (minor) — unasserted managed-retry pin sites plus the mis-named test; fixed in `cb5b3345`.
- Debuggability & Operability (nit) — doubled `resolve invoking gc executable:` prefix at four outer sites; fixed in `cb5b3345`.
- Architectural Consistency (nit) — `pinBdGCEnvironment` carried no note on why it deliberately has no `isTestBinary` carve-out, unlike its sibling `resolveProviderLifecycleGCBinary` which documents that rationale at the site; documented in `cb5b3345`.

- `ga-wbycp` — Change Impact / Blast Radius (minor): three managed-env paths still inherit an ambient `GC_BIN`. The hook-claim mutation runner is the one that matters, since claim mutations are write ops that fire hooks; the doctor store preflight is read-only and low urgency; the DoltLite early return is pre-existing structure. All three are untouched by this PR and were left out deliberately — widening it to the hook-claim write path would have mixed an unrelated behavior change into a well-scoped fix. Not addressed in this PR.
- `ga-s6dv1` — Operator note (accepted doctrine, no code change): after an in-place binary upgrade `/proc/self/exe` reads `<path> (deleted)` and resolution fails, so a long-running controller spanning that upgrade now hard-fails beads access rather than limping with a stale executable. This is the intended fail-closed behavior and already how the pre-existing `bdCommandEnv` pin behaves; the follow-up is only to write it down for operators. Not addressed in this PR.

### Maintainer Changes
One maintainer fixup was pushed to the PR branch:

`cb5b3345 fix(gc): assert managed-retry GC_BIN pins and wrap resolver errors once` — 6 files changed, 103 insertions(+), 5 deletions(-).

Your commit `d7637d1f` is preserved unchanged underneath it.

### Final Review Status
Ready for the merge queue: final head `cb5b3345` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/34345105451

### Review Iterations
1 review pass performed. It approved the change with two minor and two nit findings; one maintainer fixup commit resolved the three code-side findings, and the remaining minor finding is filed as `ga-wbycp` with the operator note as `ga-s6dv1`.

---

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR closes an ambient-`GC_BIN` leak: it pins the *invoking* `gc` executable into the environment of every Beads subprocess, so a bd hook that execs `"$GC_BIN" event emit` can no longer pick up whatever `gc` happened to be in the inherited environment. That matters because those recursive callbacks were the one place where a controller could silently hand work to a different `gc` installation than the one running — exactly the kind of thing that only shows up mid-upgrade or on a host with more than one install, and is miserable to diagnose after the fact. Six previously-unpinned bd-spawn sites now pin before spawn, and the `GC_BIN` an override sets is proven to reach the child.

### Original PR Review
Decision: approve (after one request_changes round that the fixup resolved).

Specific gaps fixed:
- **The new resolver failed closed where the old code failed soft.** `resolveBdInvokingGCBinary` returned an error when the binary's path could not be canonicalized at resolve time, and the same change deleted the last soft-degradation rung (`exec.LookPath("gc")`) from `resolveProviderLifecycleGCBinary`. The mechanism was confirmed with a standalone probe: once a running process's executable is removed, `os.Executable()` still returns the original absolute path with a *nil* error, and the follow-up `filepath.EvalSymlinks` then fails. That error propagated through `openRigStore`, `openExecStoreAtForCityWithConfig`, the `bdCommandRunnerFor*` paths, `scopedBdStoreForCity`/`ForRig`, `initDefaultRigBdStore` and `runBdStoreBridge` — so a live controller would lose its bead store until restart, where before the PR those paths simply carried no pin. The trigger is narrow (an `rm`-then-copy installer, an uninstall/downgrade window, an image-layer swap or a transient unmount — an atomic rename-over upgrade does *not* fire it), but the blast radius was process-wide.
- **The resolver's own failure modes had no direct coverage.** The first round's tests exercised the pin sites but not the resolution ladder itself, so a regression in the fallback logic would not have been caught.

Review findings addressed:
- **[Change Impact / Blast Radius, major]** — fixed in `99a3145a`. Resolution is now the three-rung ladder the review asked for: strict canonical verified path (silent) → the un-canonicalized absolute `os.Executable` value on the `EvalSymlinks`/`Stat` branches, with a warning → `gcBinaryFromPATH`, which restores the `exec.LookPath("gc")` rung that had been dropped. The hard error is reserved for the exhausted ladder. Both reviewers walked every failure exit of the strict path and confirmed each one now lands on a weaker rung, and traced the restored rung through to the three production consumers of the sibling helper that lost it.
- **[Test Evidence Quality, gating follow-on]** — fixed in `99a3145a`. `cmd/gc/cmd_bd_gc_binary_test.go` drives the real function with only the `resolveInvokingExecutable` seam, `PATH` and the logger controlled: canonicalization through a symlink (asserting *silence* — the strict rung must not warn), absent path, non-absolute, non-executable, and the exhausted ladder (errors, `errors.Is` reaches the cause, leaves `GC_BIN` unset). `PATH` is pinned to an empty temp dir where the last rung must be deterministic. Both reviewers reproduced the mutation probe independently with different mutants rather than taking the commit's word for it.
- **[Maintainability]** — `99a3145a` also extracts `newBdStoreBridgeStore`, which the review confirmed is behavior-preserving: same validation order, same error strings, password still read only after validation, store still built before the op switch.

Also validated against the code rather than the commit message: the explicit `GC_BIN` override genuinely reaches the child (last-wins authority in `mergeEnv`/`execEnvFor`, and `mergeRuntimeEnv` subsumes the removed manual `removeEnvKey`); both managed-retry env arms are non-vacuously pinned; the `func() string` → `func() (string, error)` signature change is a compile-enforced complete census with every caller's `gcBin != ""` guard intact; and no rung reopens the hazard the PR closes — each one *replaces* the ambient value, and an exhausted ladder errors rather than passing the ambient value through.

One suggestion from the first round was deliberately declined and the review agreed with the decline: adding a `.test`-path refusal inside `pinBdGCEnvironment` would now be actively worse, because that branch's successor is the PATH rung — a refusal would pin whatever `gc` the host carries, or hard-fail where none is installed, breaking every unstubbed test on these paths. The hazard stays latent (the bd shims on these paths never dereference `$GC_BIN`) and the stub-the-seam remedy is documented at the call site.

Two trade-offs were accepted rather than requested as changes, recorded here so they are not a surprise later: a degraded pin can name a path that is already gone by the time a bd hook execs it, surfacing as a child-side ENOENT instead of a gc-side error — that is a deliberate identity-over-availability choice, since rung 2 is where an `rm`-then-copy installer is about to land the new binary while rung 3 may be a different installation entirely; and the latent test-binary pin remains, now with a correct written account of why.

- `ga-vr5gt`: **[Debuggability & Operability, nit]** one process-wide `sync.Once` gates the degraded-pin warning, so only the first degradation in a process is ever reported — a run that degrades to the uncanonicalized path and later falls to the PATH rung reports only the milder earlier condition, and a degrade inside one of the `io.Discard` logger windows burns the guard with nothing printed. Suggested fix is to key the guard by rung or message.
- `ga-56223`: **[Architectural Consistency, minor]** `GC_BIN`'s *producers* stay un-canonicalized — the agent env in `template_resolve.go` and `pinInvokingGCBinary` still mint the lenient form, so the value is not consistent across the process tree this PR hardens. Not a one-line edit: the same value feeds `PrependGCBinDirToPATH`, so canonicalizing it moves every agent session's prepended PATH entry off the usually-symlinked shim directory.
- `ga-wbycp`: **[Change Impact, minor]** three pre-existing bd-spawn paths (hook-claim mutation runner, doctor store preflight, DoltLite early return) remain unpinned. Pre-existing and outside this diff.

One coverage caveat worth stating plainly: the third reviewer lane was skipped for this run, so no repo-wide cross-file sweep beyond `cmd/gc/` was performed by an independent third pass.

### Maintainer Changes
One maintainer fixup commit pushed directly to the PR branch:

- `fix(gc): pin the invoking executable across Beads subprocesses, degrading rather than failing closed` — 7 files changed, 403 insertions(+), 17 deletions(-)

The branch was also rebased onto current `main`; your original commit and authorship are preserved.

### Final Review Status
Ready for the merge queue: final head `99a3145a` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/35536967528

### Review Iterations
2 review passes performed: the first returned request_changes on the fail-closed resolver, the maintainer fixup landed the degradation ladder plus direct resolver coverage, and the second pass approved after both reviewers independently reproduced the fix's own mutation probe.

---

PR #6042, opened 2026-09-05
feat(gc): add explicit Dolt ownership handoff protocol

julianknutsen:
Adds the hidden, versioned `gc dolt-state handoff-inspect` and `handoff-stop` protocol used by Beads to release a legacy GC-managed direct Dolt process.

Requests bind the persisted city/scope, database, physical workspace, endpoint, and process identity. Stop revalidates an opaque identity token while holding the managed lifecycle lock, and preserves Beads config and metadata for rollback. That lock serializes the handoff against `recover-managed` only: it cannot exclude `start-managed`/`stop-managed`, because `recover-managed` calls both while holding it, so the caller must quiesce this city's lifecycle drivers for the whole transfer. A `stopped` result is a point-in-time postcondition -- captured server gone, data dir released, port unheld -- not a lease. Unknown ownership and unsupported sockets fail closed; refusals report whether signaling or runtime cleanup occurred. Recursive Beads calls pin `GC_BIN` to the invoking executable.

This implements the GC release side. The separate Beads ownership-transfer implementation handles replacement startup, verification, commit, and rollback; that full legacy migration is separate from fresh initialization.

Validation through `fb35185001e43e6f576d97a032ec2163c227e363`: real owned-process and strict JSON failure-path tests, config/metadata byte preservation, targeted race tests, all fast and process shards, command/resource census guards, repository vet, and the full pre-commit hook. Sol and independent Astra reviewed the fixes. The pre-push fast gate passed. The final patch uses physical paths for strict protocol and executable identity, and fixes Darwin host-boot timestamp linting. Actual macOS CI is recorded below.

Final test-only correction at `417157eb6705c617ed69f48066b47976af8b43b6` replaces the synthetic owned-process fixture with a real isolated Dolt server, preserving strict process identity assertions across Linux and macOS. The affected real-process suite and race run, resource census, full staged pre-commit hook, Sol and independent Astra review, and pre-push fast suite passed. Actual final-head macOS regression [34310755222](https://github.com/gastownhall/gascity/actions/runs/34310755222) and Linux CI [34310755221](https://github.com/gastownhall/gascity/actions/runs/34310755221) passed. macOS ran all 12 command-process shards, including the real owned-Dolt fixture in shard 7, plus acceptance, unit, and quality gates.


julianknutsen:
Thanks for the implementation and regression coverage. Closing as retired following the project owner’s clarification in the September 19 PR audit that ownership handoff is no longer the intended approach.

This hidden handoff-inspect/handoff-stop callback protocol supports the withdrawn implementation in gastownhall/beads#6297 and gastownhall/beads#6524. The later inverted handoff prototype is also being retired.

This closure is a scope decision, not a claim that this PR merged or passed a new test run. The branch and review history are retained for reference; this PR should not be returned to adopt-pr.

PR #6011, opened 2026-09-04
feat(beads): reconcile proxied init adapter onto current main

julianknutsen:
Reconciles the Beads RC proxied-init work onto current origin/main with a clean merge base.

Scope:
- provider-neutral direct/proxied + local/external init intent
- fresh bd/Dolt proxied default and bd-owned lifecycle handoff
- direct/socket compatibility and authoritative existing metadata
- focused init/proxy contract tests and CLI/config docs

Validation:
- go test ./cmd/gc -run TestConfigState -count=1
- go test ./cmd/gc -run TestInitBeadsForDir -count=1
- go test ./internal/beads/contract ./internal/config ./internal/orders
- go vet ./cmd/gc ./internal/beads/contract ./internal/config ./internal/orders

The repository pre-push sharded fast gate was attempted; its environment-sensitive BD_BIN cases are documented in handoff.

PR #6044, opened 2026-09-05
fix(beads): reject unknown persisted dolt targets

julianknutsen:
Follow-up to #6011 for ga-p9iuv.30.3.1.

Reject malformed persisted Dolt target values instead of normalizing them to local/empty intent. Adds contract regression coverage.

Validation: full make test-fast-parallel gate passed; focused contract/selector tests and go vet passed.

### DIFF
--- a/internal/beads/contract/init_intent.go
+++ b/internal/beads/contract/init_intent.go
@@ -61,7 +61,11 @@ func ResolveInitIntent(persisted InitScopeState, cliIntent, envIntent, configInt
 		if err != nil {
 			return InitIntentResolution{}, err
 		}
-		persistedIntent := InitIntent{Transport: transport, Target: normalizeTarget(persisted.Target)}
+		persistedTarget, err := normalizePersistedTarget(persisted.Target)
+		if err != nil {
+			return InitIntentResolution{}, err
+		}
+		persistedIntent := InitIntent{Transport: transport, Target: persistedTarget}
 		if persistedIntent.Transport != "" && persistedIntent.Target == "" {
 			persistedIntent.Target = "local"
 		}
@@ -128,12 +132,15 @@ func persistedTransport(mode string) (string, error) {
 	}
 }
 
-func normalizeTarget(target string) string {
+func normalizePersistedTarget(target string) (string, error) {
 	t := strings.ToLower(strings.TrimSpace(target))
-	if t == "local" || t == "external" {
-		return t
+	if t == "" {
+		return "", nil
 	}
-	return ""
+	if t == "local" || t == "external" {
+		return t, nil
+	}
+	return "", fmt.Errorf("persisted Dolt target %q is unsupported", target)
 }
 
 func formatIntent(intent InitIntent) string {
--- a/internal/beads/contract/init_intent_test.go
+++ b/internal/beads/contract/init_intent_test.go
@@ -83,6 +83,22 @@ func TestResolveInitIntentRejectsUnknownPersistedDoltMode(t *testing.T) {
 	}
 }
 
+func TestResolveInitIntentRejectsUnknownPersistedDoltTarget(t *testing.T) {
+	for _, target := range []string{"remote", "bogus", "localhost"} {
+		t.Run(target, func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


quad341 (approved):
@julianknutsen — thank you for this pair. Reconciling the proxied-init adapter across 41 files onto a clean merge base is genuinely unpleasant work, and it shows in the right places. `ResolveInitIntent` with its explicit precedence lattice — persisted, then CLI, then environment, then city config, then provider default — is a much better contract than the env-sniffing it replaces, and the test that pins "ambient environment is invisible to the resolver" is the kind of property test that keeps a contract honest years later. The socket plumbing is clean end to end: `DoltSocket` through `DoltConnectionTarget` into `doltpool.OpenSocket` and out to the doctor checks, with `validateSocketTarget` enforcing mutual exclusion at the boundary. And the two doltlite-marker guards — making sure a stale `dolt_mode` in an old doltlite scope can't route it through the Dolt proxy — are exactly the kind of thing that saves someone a very bad afternoon later.

**#6044** is the good kind of small PR. Catching that `normalizeTarget` was collapsing *any* unrecognized persisted target into `""`, which then got defaulted to `local`, is a real find: a corrupted or typo'd `external` scope would have quietly come back up pointing at a local topology, which is the sort of thing nobody notices until data is in the wrong place. Erroring instead lines it up with what `persistedTransport` already does for `dolt_mode`, so both persisted axes now behave consistently. I especially appreciate that you kept the empty-target case tolerated — pre-target scopes still resolve to `direct/local` exactly as before, which is what makes this safe to land rather than a compat break. The three chosen bad values are well picked; `localhost` is the realistic one.

**On the plan.** We're taking both. #6011 lands first, then #6044 — the base retarget and rebase are ours, no action needed from you.

**The red CI on #6011 isn't yours.** `cmd/gc process / shard 12 of 12` died installing `jq` from an Ubuntu mirror before a single Go test ran, and `CI / required` plus the evidence watchdog are purely downstream of that. We'll rerun.

**We're absorbing the rest.** Two spots in `gc-beads-bd.sh` — `run_bd_pinned` and `seed_fresh_managed_bd_version_witness` — now call bare `bd` instead of `"${BD_BIN:-bd}"`, so the workspace pin becomes a no-op on the direct-server path. Your own new `run_bd_init_proxied` keeps the pin with a comment saying precisely why, so this reads as the two direct-path sites drifting rather than an intentional change; we'll restore them and bring back the test assertion that would have caught it. We'll also move the doltlite check above the new external early-return in `providerLifecycleProcessEnvFromBase`, drop the unreachable duplicate override block below it, and tidy the discarded `args` literal in `initDefaultRigBdStore`.

**One thing worth a sentence from you when you have a moment** — not a gate, we're merging either way. The rewrite of `storage_mode_rewrite_test.go` and the matching flip in `cmd_rig_endpoint_test.go` reverse the ga-qi9km/ga-clsfl contract: embedded scopes are now preserved and the flip-time, read-time, and `gc doctor` messages must all be silent. Reading the diff, that follows directly from `ensureCanonicalScopeMetadata` now preserving an explicit persisted `dolt_mode` instead of forcing `"server"` — no rewrite, so nothing to announce — and we're proceeding on that reading. We'll re-add a short note at the top of that file recording the reasoning, since the ~60 lines explaining why the announcements existed went out with the assertions. If our reading is wrong, tell us and we'll revisit before it ships anywhere load-bearing.

Your commits and authorship stay exactly as they are on both PRs. Thanks again — this is a lot of careful compatibility work, and the guards protecting already-initialized scopes from the new default are the part we're most glad you got right.

julianknutsen:
Superseded by #6273 (feat/beads-proxied-default-rc2): the two-file fix here was cherry-picked into that consolidated branch (commit 130885d18f). Leaving this PR open for the epic owner to close; no further pushes here.

julianknutsen:
Closing as superseded by #6273: its branch contains both the unknown persisted Dolt target rejection and the regression test from this PR. The old standalone branch is retained. #6273 itself remains held out of adopt-pr pending a library-level proxied path that preserves native-Dolt performance.

Issue #5958, opened 2026-09-03
mol-dog-compactor force-pushes and has no guard for adopted/imported Dolt history


aytheo-bit:
## Summary

The `dolt` pack's scheduled `mol-dog-compactor` order squashes a database's
entire Dolt commit history to a single commit (or two, across `dolt_ignore`
bookkeeping) whenever the commit count clears a threshold, and — when a
remote is configured — force-pushes the flattened history to it. The order
runs on a fixed cooldown, is on by default once the `dolt` pack's order file
is loaded (there is no `enabled` field in the order itself; disabling it
requires an explicit `[[orders.overrides]]`), and has no check for whether
the database's history was freshly adopted/imported from elsewhere rather
than grown organically by this city.

## What happened (production)

On Node A, the scheduled order `mol-dog-compactor` ran on the just-adopted
`contentbuild` rig and flattened the KitFlowApp Dolt history from ~4300
commits to 2. Row data was intact (1218 issues, 2404 comments). Only the
absence of a configured Dolt remote on the adopted database prevented the
flattened history from being force-pushed. Operator mitigation:
`[[orders.overrides]] name="mol-dog-compactor" enabled=false` + `gc reload`.

## Where in the code

- `examples/bd/dolt/orders/mol-dog-compactor.toml:1-8` — the order:
  `trigger = "cooldown"`, `interval = "2h"`, `exec = "gc dolt compact"`,
  `timeout = "24h"`. No `enabled` key — orders scanned from an
  orders/ directory are active unless overridden
  (`internal/config/config.go:2104-2105`, `OrderOverride.Enabled *bool`, the
  only place `enabled` appears for orders — there is no default-off knob on
  the order file itself). Contrast: the unrelated weekly
  `[maintenance.dolt]` GC+backup loop is explicitly documented "Opt-in for
  v1: omission or enabled=false leaves the loop disabled"
  (`internal/config/config.go:2437-2441`) — the compactor has no equivalent
  opt-in gate.
- `examples/bd/dolt/commands/compact/run.sh:2303-2312` — the only skip
  condition before flattening a database is
  `commit_count < threshold_commits` (`GC_DOLT_COMPACT_THRESHOLD_COMMITS`,
  default 2000, set at `run.sh:302`). Nothing here inspects provenance
  (adopted vs. natively grown) — a database is flattened purely because it
  has enough commits, and an adopted multi-year history (~4300 commits)
  clears that bar trivially.
[...]

PR #6052, opened 2026-09-05
fix(dolt): guard compaction with runtime maintenance and federation policy

aphexcx:
## Summary

Upstream mirror of gp-t13a (first slice of gp-j59), based on current `upstream/main` with only the compaction patch and adaptations to newer upstream tests: scheduled compaction could flatten shared Dolt history even with `[maintenance.dolt] enabled = false`, breaking the next cross-city pull.

The command now reads the resolved city's flag at invocation time with `gc --city "$GC_CITY_PATH" config show --json`. This follows the existing resolved-config CLI and honors config fragments/overrides without relying on an order's inherited environment. False skips all modes before runtime setup or any Dolt operation; unreadable/non-boolean config fails closed. `GC_DOLT_COMPACT_FORCE=1` permits a manual announced run.

Flatten mode checks every database for any remote before mutations or deferred GC/push. It names and skips shared databases unless `GC_DOLT_COMPACT_ALLOW_FEDERATED=1` is set for an announced window. `.no-sync`, `--skip-fetch`, dry runs, and FORCE cannot bypass that guard. Bare working-set GC and `--gc-only` retain their reclaim paths and still honor the maintenance flag/quarantine checks.

## Testing

- Final integration rerun with broad gates stopped: `CGO_ENABLED=0 go test ./examples/bd/dolt -count=1` — PASS, 116.629s at committed `16d40d839`.
- TDD: the new disabled/config/federation regressions failed against the original script, including `.no-sync` flattening and dry-run reporting that it would flatten.
- `CGO_ENABLED=0 go test ./examples/bd/dolt -count=1` — PASS, 118.958s.
- `CGO_ENABLED=0 go test ./examples/bd/dolt -run '^TestCompactScript(FederatedGuard|.*NoSync)' -count=1` — PASS, 7.988s after simplifying the fixture helper for lint.
- `CGO_ENABLED=0 go test -tags dolt_integration ./examples/bd/dolt -run '^TestCompactScriptRealDoltRemotePush$' -count=1 -v` — PASS, 9.245s; temporary database/server/remote only.
- `CGO_ENABLED=0 make vet`, `sh -n examples/bd/dolt/commands/compact/run.sh`, `git diff --check` — PASS.
- Required pre-commit: format, changed-package lint (0 issues), code generation, vet, and `make check-docs` — PASS.
- Real installed `gc config show` against a temporary city with maintenance disabled: the changed script's `--dry-run` exited 0 with the exact skip line.
- CGO disabled for direct Go checks because this host's default compiler lookup lacks ICU headers. No live store compaction was run.

The first broad pre-push run at default `LOCAL_TEST_JOBS=16` hit timing failures in two compact fake-Dolt probes and five unchanged supervisor/session tests. The latter all passed together with `-parallel=2` (3.096s). The upstream gate at `LOCAL_TEST_JOBS=2` also reproduced `TestStopManagedCityForcesCleanupAfterTimeout` (1.171984s). Both already-failed broad runs were stopped through their supervised runner cleanup; **the broad fast baseline is not green**. Follow-up: gp-pk66. After rechecking bead ownership, branches were pushed with the repository-documented `--no-verify` option. This PR is a draft pending the mayor's review; no test deadlines or assertions were weakened.

- Upstream mirror: `CGO_ENABLED=0 go test ./examples/bd/dolt -count=1` — PASS, 132.969s; its additional backup/notification tests explicitly opt into announced runs. Its required pre-commit gates also passed.

## Checklist

- [x] Linked work: gp-t13a / gp-j59.
- [x] Tests cover disabled flags, config/probe errors, remotes without origin, multiple remotes, independent overrides, `.no-sync`, skip-fetch, dry runs, deferred markers, and reclaim modes. Existing remote-flatten tests explicitly opt into the announced-window switch.
- [x] Help documents both guards and both overrides.
- [x] Migration: omission already resolves to the maintenance config's false default. Set the flag true for scheduled maintenance; manual flatten of a disabled federated store requires both switches. `jq` is required to parse the resolved flag. Deferred markers remain intact while federation blocks flatten mode.

Codex self-review (DEEP): VERDICT CLEAN for the compaction diff; broad-suite timing failures remain tracked separately. The mayor owns the Fable adversarial review at READY; do not merge yet.

Citadel receives this pack through `[imports.bd]` (`gastownhall/gascity`, `examples/bd`), whose `[imports.dolt]` exports `./dolt`. Citadel currently pins bd to `sha:f895c0ff47d6ee9334ed282a416387eb5b084d24`; after upstream merge the mayor must advance that import pin and reconcile imports. `gc import check` reports 7 remote imports. The running pack cache and city configuration were left untouched.

Integration PR: https://github.com/aphexcx/gascity/pull/10

## Behavior change (read this first)

DoltMaintenance.Enabled defaults to false (opt-in, internal/config/config.go). After this change, a city that never set [maintenance.dolt] enabled = true no longer compacts at all: the scheduled flatten, --gc-only, dry runs, and the bare working-set GC path all print the skip line and exit 0. That is deliberate (gp-j59: disable-by-default after two unannounced flattens broke a federated store), and it moves the journal-growth bound the 2h order provided behind an explicit opt-in. Operators who want scheduled compaction set enabled = true; a one-off announced run uses GC_DOLT_COMPACT_FORCE=1, and a shared (federated) store additionally needs GC_DOLT_COMPACT_ALLOW_FEDERATED=1.


julianknutsen:
Superseded by #6554 (merged as 91876e84a3, backport #6585 to release/v1.5.0). Thank you @aphexcx: the remote guard from this PR (skip flatten on any database with a Dolt remote unless GC_DOLT_COMPACT_ALLOW_FEDERATED=1) is carried in #6554 with a Co-authored-by trailer. The `[maintenance.dolt]` default-off gate was intentionally not carried: #6554 instead adds a gc-compact-base provenance watermark so owned history still compacts while adopted/shared history is never rewritten (#5958).

Issue #5982, opened 2026-09-03
bug: processenv: one malformed secrets.env line drops every other credential


jacobhausler:
## Summary

One malformed line in `${GC_HOME}/secrets.env` silently drops **every**
credential in the file: `ParseEnvFile` aborts on the first bad line, and its
only caller discards the whole file's already-parsed entries on that error,
logging a single stderr line. Silent full-file credential loss on a single
typo is a bigger blast radius than the format gap that triggered the incident
this came from.

## Symptom

`secrets.env` containing, e.g.:

```
ANTHROPIC_AUTH_TOKEN=sk-live-123
CODEX_AUTH_JSON={
OPENAI_API_KEY=sk-openai-456
```

(the `CODEX_AUTH_JSON={` line is a leftover from a multi-line paste — a
format the parser never supported) produces one stderr line during
`gc supervisor install`:

```
gc: parsing supervisor secrets file ".../secrets.env": line 2: missing '=' in "CODEX_AUTH_JSON={"
```

and the supervisor service is installed with **no** credentials from the file
— both `ANTHROPIC_AUTH_TOKEN` and `OPENAI_API_KEY` are gone, not just the bad
line.

## Reproduction

1. Put a `secrets.env` with at least two valid `KEY=VALUE` lines and one line
   missing `=` (or with an empty key) under `${GC_HOME}`.
2. Run `gc supervisor install` (or any path that calls
   `supervisorSecretsEnvFileEntries`).
3. Result: the rendered service unit contains none of the file's entries.
   Expected: the valid entries are kept and only the malformed line is
   reported and skipped.

No race; deterministic.

## Root cause

`ParseEnvFile` (`internal/processenv/envfile.go:23-42`) returns
`nil, error` on the first malformed line (missing `=` at `:31-34`, empty key
at `:36-39`), so the valid entries parsed before it are discarded with the
nil map. Its only production caller,
`supervisorSecretsEnvFileEntries`
(`cmd/gc/cmd_supervisor_lifecycle.go:1260-1275`), on that error logs one
stderr line and returns `nil` — the entire file's credentials are dropped.
The doc comment's "fails loudly rather than silently dropping a credential"
is not what the caller does: it logs one line and drops everything.

Verified against `main` at `f07cfb682` (2026-09-03): the path is unchanged.
`ParseEnvFile` has exactly one production caller (the secrets merge path), so
the blast radius is this one merge path.

## Fix candidates

[...]

PR #6057, opened 2026-09-05
fix(processenv): error on unterminated quote in env file value (#6022)

### LINKED ISSUE #6022: bug: processenv: a multi-line quoted value silently splits into stray env vars (no error)
## Summary

When a quoted value in `${GC_HOME}/secrets.env` spans lines **and every
continuation line happens to contain `=`**, `ParseEnvFile` reports **no error
at all**. It silently truncates the value at the first newline (keeping its
opening quote) and turns each continuation line into its **own environment
variable**, inventing variable names from the file's content.

This is a different code path from #5982. That issue is about a *malformed*
line producing an error whose caller then drops the whole file. Here there is
no error, so #5982's proposed fix (keep parsing, collect one error per bad
line, return the partial map) leaves this behavior **completely unchanged** —
the input is not malformed by the parser's own reckoning; it parses "fine"
into the wrong thing.

Filing separately for that reason, not to relitigate #5982's out-of-scope note
on multi-line grammar. The ask here is not "support multi-line values" — it is
"do not silently invent environment variables from a value's continuation
lines." If multi-line is out of the format, the documented contract for
out-of-format input is a loud parse error, which is exactly what this path
does not produce.

## Symptom

A credentials file containing:

```
GC_NOMAD_AGENT_LAUNCH_SCRIPT="export PATH=/mnt/nomad/gc/bin/current:$PATH
export CODEX_HOME=$NOMAD_SECRETS_DIR/codex-home"
```

A POSIX shell sourcing this file gets **one** variable holding both lines.
`ParseEnvFile` returns **two** entries and a nil error:

| key | value |
| --- | --- |
| `GC_NOMAD_AGENT_LAUNCH_SCRIPT` | `"export PATH=/mnt/nomad/gc/bin/current:$PATH` (truncated; opening `"` retained) |
| `CODEX_HOME` | `$NOMAD_SECRETS_DIR/codex-home"` |

Two distinct defects in one input:

1. The kept value is corrupt — truncated at the newline **and** carrying a
   stray leading `"`, because `unquoteEnvValue` only strips quotes when the
   first and last character of the *line* match.
2. `CODEX_HOME` is a variable that appears nowhere in the operator's file as
   an assignment they intended. The `export ` prefix strip
   (`internal/processenv/envfile.go:30`) makes the script's own
   `export CODEX_HOME=...` line indistinguishable from a real top-level
   assignment, so a value's *contents* become a key in the returned map.

## Why it matters

`supervisorServiceExtraEnv` gates file entries on a persist allowlist or an
explicit `GC_SUPERVISOR_ENV` opt-in, so an invented key does not automatically
reach the generated unit. The exposure is that the map is wrong in a way
nothing reports: the intended variable holds a truncated value, and a key the
operator never wrote now exists and is eligible for any downstream consumer
that does opt it in or that iterates the map. A credentials file that is
completely valid to the shell reader gc's own docs point at
(`cmd/gc/cmd_supervisor.go`, the `GC_SUPERVISOR_OMIT_PROVIDER_CREDS` note
recommending "a wrapper around `gc supervisor run` that sources a credentials
file") parses, without complaint, into different variables than the operator
wrote.

## Reproduction

Verified against `main` at `1f19d26c8`; the path is unchanged from the version
where we hit it. Deterministic, no race.

```go
func TestParseEnvFileSilentlySplitsAMultiLineValueIntoStrayVars(t *testing.T) {
	content := `GC_NOMAD_AGENT_LAUNCH_SCRIPT="export PATH=/mnt/nomad/gc/bin/current:$PATH
export CODEX_HOME=$NOMAD_SECRETS_DIR/codex-home"
`
	got, err := ParseEnvFile(content)
	if err != nil {
		t.Fatalf("ParseEnvFile: %v", err)
	}
	if len(got) != 2 {
		t.Fatalf("got %d entries, want 2 (the split): %v", len(got), got)
	}
	if want := `"export PATH=/mnt/nomad/gc/bin/current:$PATH`; got["GC_NOMAD_AGENT_LAUNCH_SCRIPT"] != want {
		t.Errorf("launch script = %q, want %q (truncated at the newline, opening quote kept)",
			got["GC_NOMAD_AGENT_LAUNCH_SCRIPT"], want)
	}
	if stray, ok := got["CODEX_HOME"]; !ok {
		t.Fatalf("expected a stray CODEX_HOME from the continuation line: %v", got)
	} else {
		t.Logf("stray var invented from a script line: CODEX_HOME=%q", stray)
	}
}
```

Passes on `main` — i.e. the buggy behavior is what happens.

## Root cause

`ParseEnvFile` (`internal/processenv/envfile.go:23-42`) is strictly
line-oriented: it splits on `"\n"` and treats each line independently
(`strings.Cut(line, "=")` at `:31`). Nothing tracks whether an opening quote
from a previous line is still unterminated, so a quoted value that spans lines
is never seen as one value. Quote stripping happens per line in
`unquoteEnvValue` (`:46-54`), which requires the first and last byte of that
one line to match (`:51-52`) — false for the opening line of a multi-line
value, hence the retained `"`. The `export ` strip at `:30` then promotes any
continuation line that looks like a shell assignment into a map key.

The missing-`=` error at `:32-34` is what catches the *other* shape (#5982);
it simply never fires when the continuation lines all contain `=`.

## Fix candidates

**A. Detect an unterminated quote and error.** After `strings.Cut`, if the
value begins with `'` or `"` and does not end with the same character, return
a parse error naming the line ("unterminated quote in value for KEY; multi-line
values are not supported"). Stays inside the documented grammar, converts a
silent wrong parse into the loud failure the package already promises, and
needs no signature change.
- Scope: ~10 LOC in `internal/processenv/envfile.go` plus a test.
- Trade-off: an operator with such a file gets a hard failure rather than a
  quiet mis-parse. That is the point, but it does interact with #5982 — on its
  own it converts this into #5982's whole-file-drop, so the two want landing
  together (or A here plus #5982's candidate A, which then reports this line
  and keeps the others).

**B. Support multi-line quoted values.** Track an open quote across lines and
join until the closing quote, matching what a shell `source` does with the
same file.
- Scope: ~40 LOC, a real grammar change to the documented format.
- Trade-off: closes the gap between the two readers of this file properly, but
  #5982 explicitly scoped multi-line support out, so this is a maintainer
  product call rather than a bug fix.

**Recommendation:** A, coordinated with #5982. It is the minimal change that
makes the package honor its own stated contract ("a malformed credentials file
fails loudly rather than silently dropping a credential") on this input, and
it does not commit the project to a grammar it has declined. B only if
multi-line support is wanted on its own merits.

## Context

Found while vetting a supervisor-env incident in our city (westlands,
`cr-aqzig`): agent lanes launched with a `CODEX_HOME` nobody had configured.
Our own mitigation was to move the launch script out of the credentials file
to a file path, so we are not blocked on this — filing because the parse
behavior is reachable by any operator who puts a quoted multi-line value in
that file, and it is invisible when it happens.

rjgeng:
## What

Closes #6022. A companion to #5982/#6043 (already shipped), but a genuinely different code path — this one produces **no error at all**. When a quoted value in `\${GC_HOME}/secrets.env` spans multiple lines (e.g. a multi-line shell script assigned to one variable) and every continuation line happens to contain `=`, `ParseEnvFile` silently truncates the value at the first newline (keeping a stray opening quote) and parses each continuation line as its **own** top-level key/value pair — inventing environment variables the operator never wrote. Found in production: a launch script assigned to `GC_NOMAD_AGENT_LAUNCH_SCRIPT` produced a stray `CODEX_HOME` variable from its own `export CODEX_HOME=...` continuation line.

Built against current `main` (pre-#5982/#6043 shape — that PR is not merged yet, so this one does not depend on it and keeps `ParseEnvFile`'s existing single-error-return signature).

## Fix

This is the issue's own recommended "fix candidate A" — convert the silent mis-parse into the loud failure the package already documents as its contract ("a malformed secrets file fails loudly rather than silently dropping a credential"). After key/value split and validation, a new `hasUnterminatedQuote` check: if the trimmed value opens with a quote character (`'`/`"`) but doesn't close with that same character on the same line, return a parse error naming the line and key instead of silently truncating. A lone single-character quote value (`val == "\""`) and a properly-closed empty quoted string (`""`) are both correctly exempted (checked via explicit test cases).

**Deliberately not implemented:** fix candidate B (actual multi-line value support — tracking an open quote across lines and joining them). The issue itself frames this as "a real grammar change... a maintainer product call rather than a bug fix," separate from this fix's scope.

**Coordination note with #5982/#6043:** the issue explicitly says the two "want landing together" in spirit — once #5982 merges (which changes `ParseEnvFile` to `(map, []error)`, collecting per-line errors instead of aborting), this fix's single `return nil, error` call site will need a small follow-up adaptation to fit the new partial-parse shape. Flagging here so whoever merges second doesn't hit a surprise conflict.

## Validation

- `go test -tags gms_pure_go ./internal/processenv/...` pass — build confirmed the new tests fail against pre-fix code (stashed the fix, reran, correctly got `nil` where an error was expected) before restoring it
- Full local push-gate run: all failures across every shard and `unit-core` confirmed to be pre-existing environmental/timing flakes (dolt-compaction timeouts, detached-probe/systemctl-delegate/lsof tests, interrupt-timing tests) unrelated to this change — none touch `internal/processenv/envfile.go`

### DIFF
--- a/internal/processenv/envfile.go
+++ b/internal/processenv/envfile.go
@@ -17,9 +17,12 @@ import (
 //
 // Values are treated literally otherwise: there is no variable interpolation
 // and no inline-comment stripping on unquoted values (a '#' mid-value is kept).
-// A line missing '=' or with an empty key is a parse error so a malformed
-// secrets file fails loudly rather than silently dropping a credential. The
-// last assignment wins when a key repeats.
+// A line missing '=', a line with an empty key, or a value that opens a quote
+// without closing it on the same line (multi-line quoted values are not
+// supported) is a parse error, so a malformed secrets file fails loudly
+// rather than silently dropping a credential or splitting a quoted value
+// across invented top-level keys. The last assignment wins when a key
+// repeats.
 func ParseEnvFile(content string) (map[string]string, error) {
 	out := make(map[string]string)
 	for i, raw := range strings.Split(content, "\n") {
@@ -36,9 +39,30 @@ func ParseEnvFile(content string) (map[string]string, error) {
 		if key == "" {
 			return nil, fmt.Errorf("line %d: empty key in %q", i+1, raw)
 		}
-		out[key] = unquoteEnvValue(strings.TrimSpace(val))
+		val = strings.TrimSpace(val)
+		if hasUnterminatedQuote(val) {
+			return nil, fmt.Errorf("line %d: unterminated quote in value for %q; multi-line values are not supported", i+1, key)
+		}
+		out[key] = unquoteEnvValue(val)
 	}
 	return out, nil
+}
+
+// hasUnterminatedQuote reports whether a trimmed value opens with a single or
+// double quote but does not close with that same quote character on this
+// line. That pattern is the signature of a value that was meant to span
+// multiple lines (e.g. a multi-line quoted shell script), which this
+// line-oriented parser cannot represent correctly: it would otherwise
+// silently truncate the value and may misparse the continuation lines as
+// unrelated top-level assignments. A single-character value that is just a
+// lone quote (e.g. val == `"`) is not treated as unterminated, since its
+// first and last byte are the same quote character.
+func hasUnterminatedQuote(val string) bool {
+	if len(val) < 2 {
+		return false
+	}
+	first, last := val[0], val[len(val)-1]
+	return (first == '"' || first == '\'') && last != first
 }
 
 // unquoteEnvValue strips one layer of matching surrounding single or double
--- a/internal/processenv/envfile_test.go
+++ b/internal/processenv/envfile_test.go
@@ -1,6 +1,9 @@
 package processenv
 
-import "testing"
+import (
+	"strings"
+	"testing"
+)
 
 func TestParseEnvFileParsesCoreSyntax(t *testing.T) {
 	content := `# leading comment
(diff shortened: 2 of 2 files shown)


julianknutsen:
Superseded by #6043 (merged as 10f0f0d330). Thank you @rjgeng: this PR's unterminated-quote detection was the starting point for the #6022 fix.

On top of your partial-parse change, #6043 skips a multi-line quoted value as one block instead of returning a hard error. The block runs through the first later line that ends with the matching quote and contains an odd number of it, and is reported once with the key and line range. Taken alone, a hard error would have dropped every credential in the file (the #5982 failure mode) for lines like `KEY="v" # comment`, which parsed before. It also would not have stopped the stray `CODEX_HOME` from the continuation line. #6043 also rejects non-identifier keys and keeps values out of the error text.

Closing in favour of #6043.

PR #6058, opened 2026-09-05
feat(doctor,orders): say so when a provider binary is missing instead of starting empty

Rome-1:
## What happens today

On a host where a referenced provider binary is absent from PATH, gc drops every session using that provider from the desired state and no surface says so:

- `gc start` reports `City started under supervisor.`
- `gc status` shows the affected agents as `unknown (partial status)`
- `gc agent list` shows them `active`
- `gc doctor` passes 100 checks — none of which asks whether a configured provider's command exists
- the session-reconciler trace computes `desired_session_count=0`, and the string `provider` appears **zero** times in it
- `.gc/events.jsonl` has no provider error

The only command that names the cause is `gc session new <template>`:

```
gc session new: provider not found in PATH: provider "codex" command "codex"
```

That is the one command an operator has no reason to run when `gc start` has already claimed success. The city comes up completely empty and reports a clean start.

## Changes

**1. `config.ProviderNotInPATHError`** — the structured form of the existing `ErrProviderNotInPATH` sentinel, carrying the provider reference and the command as fields rather than only inside formatted text. Both new diagnostics need those two identifiers as data, which is why this is a type and not another `fmt.Errorf`.

`Error()` reproduces the previous layout byte for byte and `Unwrap()` returns `ErrProviderNotInPATH`, so every existing `errors.Is` check and message assertion keeps working.

**2. A `provider-path` doctor check** — every template a session can be built from must have a provider command that resolves on the host.

The neighboring provider checks stop short of this by design, so this is not duplication: `provider-catalog` checks that every referenced provider is *declared*; `provider-catalog-local-readiness` checks that locally-configured CLIs are listed explicitly; `provider-parity` inspects capability fields using a *stub* PATH lookup because it is auditing config semantics rather than the host. This check is the one that touches the host.

Reachability is "every configured agent template", not "every template with demand right now" — a pool template at `min_active_sessions = 0` is one routed bead away from needing its provider. Agents pinning `start_command` are skipped: they bypass `ProviderSpec` and `ResolveProvider` never consults PATH for them.

**3. The reconciler reports what it dropped** — the desired-state build accumulates PATH failures and the controller emits them to both the log and `.gc/events.jsonl`, with `desired_session_count` printed beside the cause. The reported symptom was a city that started clean and stayed empty, so the number that made it look normal belongs next to the reason it wasn't.

## Design notes for review

- **Grouped by provider.** One uninstalled binary shared by fifty agents reports once, with a dropped-template count and a capped sample of names, not fifty times.
- **The accumulator is held by pointer** on `agentBuildParams`. Several build paths take a value copy of that struct (`resolveTemplateForSessionBeadInfo` does `local := *bp` to scope a `beadNames` override), so a slice field would collect failures into whichever copy happened to see them and lose them at the seam. A mutex covers the pool-evaluation fan-out, which resolves templates from more than one goroutine.
- **The alert latches per episode.** The build re-derives the same failure on every tick and the cached demand snapshot replays it in between, so an unlatched version would write a line and an event every tick for as long as the binary stays uninstalled. Recovery clears the latch, so install-then-remove alerts a second time.
- **Only `ProviderNotInPATH` is recorded.** Unknown-provider and bad-option faults are already named by the `config-valid` / `config-refs` / `provider-catalog` checks; folding them in here would dilute the one signal this carries.

## Verification

```
go build ./...                                                   OK
go vet ./cmd/gc/ ./internal/doctor/ ./internal/config/ ./internal/events/   OK
golangci-lint run --new-from-merge-base=origin/main --whole-files           0 issues
go test ./cmd/gc/                                                ok (306s)
go test ./internal/doctor/ ./internal/config/ ./internal/events/ ok
```

`cmd/gc/testdata/doctor_check_names.golden` gains `provider-path` in registration order. Rebased onto current `main` and the full gate re-run after the rebase.

https://claude.ai/code/session_01AnYmK2nvdxwSKsG45LvinA

### DIFF
--- a/cmd/gc/agent_build_params.go
+++ b/cmd/gc/agent_build_params.go
@@ -127,6 +127,15 @@ type agentBuildParams struct {
 	// etc.). Used by the skill materialization integration to decide
 	// stage-2 eligibility.
 	sessionProvider string
+
+	// providerUnresolved collects the provider-not-in-PATH failures that
+	// dropped templates from this build, so buildDesiredState can report
+	// the hole instead of just returning a smaller desired set. A pointer,
+	// not a slice: resolveTemplateForSessionBeadInfo (and any other seam
+	// that scopes params via `local := *bp`) must accumulate into the same
+	// place as the parent (ob-woag). Nil in focused unit fixtures that
+	// build params by hand; every recording site is nil-receiver-safe.
+	providerUnresolved *providerUnresolvedSet
 }
 
 // hasCompleteSessionSnapshot reports whether a store-backed build has a
@@ -145,6 +154,18 @@ func (p *agentBuildParams) hasCompleteSessionSnapshot() bool {
 		return p.sessionSnapshotComplete
 	}
 	return p.sessionBeads != nil && p.sessionBeads.LoadError() == nil
+}
+
+// recordProviderUnresolved forwards one template-resolution error to this
+// build's provider-not-in-PATH accumulator, and reports whether the error was
+// one. Nil-safe on both the params and the accumulator so the legacy fixtures
+// that hand-build params (and the storeless paths that pass a nil bp) stay
+// working.
+func (p *agentBuildParams) recordProviderUnresolved(template string, err error) bool {
+	if p == nil {
+		return false
+	}
+	return p.providerUnresolved.record(template, err)
 }
 
 // newAgentBuildParams constructs agentBuildParams from the common startup values.
@@ -171,6 +192,8 @@ func newAgentBuildParams(cityName, cityPath string, cfg *config.City, sp runtime
 		beadNames:       make(map[string]string),
 		stderr:          stderr,
 		sessionProvider: cfg.Session.Provider,
+
+		providerUnresolved: newProviderUnresolvedSet(),
 	}
 	if store != nil {
 		params.poolSessionCreateBudget = poolplan.NewCreateBudget(cfg.Daemon.MaxWakesPerTickOrDefault())
--- a/cmd/gc/build_desired_state.go
+++ b/cmd/gc/build_desired_state.go
@@ -156,6 +156,15 @@ type DesiredStateResult struct {
 	SessionSnapshotComplete bool
 	SessionOccupancyInfos   []session.Info
 	BeaconTime              time.Time
+	// ProviderUnresolved names every provider whose command did not resolve
+	// on PATH during this build, with the templates it cost. It is the
+	// answer to "why is desired_session_count zero" that ob-woag could not
+	// get from any surface: the drop sites log to a supervisor stderr and
+	// otherwise return a quietly smaller State. Non-empty here means the
+	// desired set is smaller than the config asked for, and by exactly the
+	// templates listed. The controller turns it into a trace record, an
+	// event, and a once-per-episode log line.
+	ProviderUnresolved []ProviderUnresolvedGroup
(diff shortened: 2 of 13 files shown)


Rome-1:
**The red check is pre-existing on `main`, not from this PR.**

`Integration / packages-core-4-of-4` fails on `TestResidencyBoundaryGrepRatchet`:

```
RESIDENCY-BOUNDARY: cmd/gc/infra_class_migrate.go infraBindingIDPrefix d:ReservedClassPrefix:
  1 sites, baseline 0 — a NEW store-enumeration site.
RESIDENCY-BOUNDARY: cmd/gc/infra_class_migrate.go openInfraDestination d:ReservedClassPrefix:
  0 sites, baseline 1 — shrink the baseline in the same commit that retires a site.
```

That names `cmd/gc/infra_class_migrate.go`, which this PR does not touch. It arrived with `fc7a23a29` (#5597, the infra-class cutover), where `openInfraDestination` appears to have become `infraBindingIDPrefix` without `scripts/residency-boundary-baseline.txt` being regenerated.

Verified against a clean checkout of `origin/main` at `fc7a23a29`, with no changes applied:

```
$ git worktree add --detach origin/main   # fc7a23a29
$ go test ./scripts/ -run TestResidencyBoundaryGrepRatchet
--- FAIL: TestResidencyBoundaryGrepRatchet (13.53s)
    [byte-identical output to the CI failure above]
FAIL
```

The other three red checks — `CI / integration`, `CI / required`, `Evidence / critical-path suites` — are aggregates of that one shard. Every other check passes: 69 success, 15 skipped.

I have not touched the baseline. The failure message notes that three emitters feed that one file and that regenerating from any single emitter drops the other two halves' rows, so this belongs in a commit by someone who knows which emitter owns the renamed site — not smuggled into an unrelated PR. Happy to file it separately if that would help.

https://claude.ai/code/session_01AnYmK2nvdxwSKsG45LvinA

Rome-1:
Closing the loop on my earlier comment: **the pre-existing `main` failure has been fixed upstream, and this PR's red checks are gone with it.**

`TestResidencyBoundaryGrepRatchet` now passes on `origin/main` — `scripts/residency-boundary-baseline.txt` carries the `infraBindingIDPrefix` row with a comment explaining the `#5597` attribution. Verified by running the test against a clean checkout of `8aeb003a7`.

This branch's four red checks were that failure plus three aggregates of it, pinned to an unchanged head SHA, so they could not clear on their own. Rebased onto current `main` (clean, one commit over 133) and re-ran the full gate:

```
go build ./...                                                   OK
go vet ./cmd/gc/ ./internal/doctor/ ./internal/config/ ./internal/events/   OK
gofmt                                                            clean
golangci-lint run --new-from-merge-base=origin/main --whole-files           0 issues
go test ./cmd/gc/                                                ok (378s)
go test ./internal/doctor/ ./internal/config/ ./internal/events/ ok
```

One note for whoever reviews: the `packages-core-4-of-4` integration shard also fails locally on `TestBdFlagManifestCurrent`, and that is likewise **not** from this PR — it reproduces identically on a clean `origin/main` with `-tags integration`. It is an environment mismatch rather than a code fault: the locally installed `bd` exposes a `--profile` flag that `internal/bdflags`' manifest does not list, and the manifest is required to be a superset. CI pins its own `bd`, so it may not appear there at all.

PR #6060, opened 2026-09-05
fix(residency): re-key infra_class_migrate.go's ReservedClassPrefix row to infraBindingIDPrefix after #5597 (unblocks the pre-push gate)

quad341:
## What

Re-keys one row of `scripts/residency-boundary-baseline.txt`: `cmd/gc/infra_class_migrate.go`'s single `config.ReservedClassPrefix` site is now attributed to `infraBindingIDPrefix`, the helper #5597 extracted it into, instead of `openInfraDestination`, the function it used to live in. The header prose beside the row says why. No Go code changes.

## Why

Since #5597 merged (fc7a23a292, 2026-09-05 08:28 PDT) `TestResidencyBoundaryGrepRatchet` fails on `main` itself:

```
RESIDENCY-BOUNDARY: cmd/gc/infra_class_migrate.go infraBindingIDPrefix d:ReservedClassPrefix: 1 sites, baseline 0 — a NEW store-enumeration site.
RESIDENCY-BOUNDARY: cmd/gc/infra_class_migrate.go openInfraDestination d:ReservedClassPrefix: 0 sites, baseline 1 — shrink the baseline in the same commit that retires a site.
```

That test runs in the pre-push gate, so it has refused every push to this repo from every agent and human here since then. Reproduced in a clean worktree at fc7a23a292 with no other changes (ga-zohuic).

## Why a census move and not an allowance

The file still has exactly one `ReservedClassPrefix` hit; #5597 moved it into a helper (now consumed by `openInfraDestination` and `openInfraBindingReadOnly`). The ratchet keys rows by function, so a moved site reads as one growth row plus one shrink row. Re-keying the row describes the tree as it is and keeps the site pinned at 1, shrink-only from here, with no `// residency:allow` marker in the code.

Whether `infraBindingIDPrefix` should consume `internal/storeref` instead of deriving the prefix is #5597's author's call and is deliberately left open (comment on #5597, 2026-09-05 10:20 PDT). This is an unblock, not that decision.

## Evidence

- `bash scripts/check-residency-boundary.sh` exits 0 on the branch.
- `go test ./scripts -run 'TestResidency'` passes (ratchet, resolver boundary, AST ratchet, halves-agree).
- Rows re-sorted check: `grep -v '^#' scripts/residency-boundary-baseline.txt | LC_ALL=C sort -c` is silent.
- The baseline was edited by hand, not regenerated, so the three-emitter splice hazard the header warns about does not apply.
- The pre-push hook did **not** run on this push: the scratch clone it was pushed from had no `core.hooksPath` configured (since fixed on that clone). For a new branch the hook runs `make test-fast-parallel`; that job was run by hand on the branch afterwards and its result is posted in the first comment below.

Refs ga-zohuic (P1). Unblocks ga-451jnv's committed fix.

### DIFF
--- a/scripts/residency-boundary-baseline.txt
+++ b/scripts/residency-boundary-baseline.txt
@@ -60,6 +60,15 @@
 # declaring vocabulary counts as mentioning it, and the row still moves if the
 # declaration is renamed. "Hits" here means mentions, not call sites.
 #
+# infra_class_migrate.go's row is keyed to infraBindingIDPrefix, not to
+# openInfraDestination, since #5597 extracted the prefix derivation into that
+# helper (openInfraDestination and openInfraBindingReadOnly both consume it).
+# It is the same single ReservedClassPrefix site re-keyed to the function that
+# now opens it — a census move, not growth: the file still has one hit. Whether
+# the helper should consume internal/storeref instead of deriving the prefix is
+# #5597's author's call; this row only stops the ratchet reading a moved site
+# as a new one (ga-zohuic).
+#
 # The two internal/dispatch/drain.go rows arrived when the signature rule
 # widened from cmd/gc+internal/api to every directory grep already scanned. The
 # comment that justified the narrower set claimed sling and dispatch held no
@@ -185,7 +194,7 @@ cmd/gc/dispatch_control_ready.go	controlReadyCacheSources	ast:returns-store-list
 cmd/gc/dispatch_control_ready.go	controlReadyCacheSources	c:beads.Store-literal	3
 cmd/gc/doctor_order_tracking_retention.go	Run	c:beads.Store-literal	1
 cmd/gc/formula_cook_attach_class.go	attachGraftClassRefusal	b:graphClassBinding	1
-cmd/gc/infra_class_migrate.go	openInfraDestination	d:ReservedClassPrefix	1
+cmd/gc/infra_class_migrate.go	infraBindingIDPrefix	d:ReservedClassPrefix	1
 cmd/gc/order_dispatch.go	gateStoresFor	ast:returns-store-list	1
 cmd/gc/order_dispatch.go	gateStoresFor	c:beads.Store-literal	1
 cmd/gc/order_store.go	appendOrdersSweepStore	ast:returns-store-list	1


quad341:
Evidence promised in the description: `make test-fast-parallel` (the job the pre-push hook runs for a new branch) was run by hand on this branch at 638b0750d after the push, because the clone it was pushed from had no `core.hooksPath` and so the hook did not fire.

```
All fast jobs passed
real    12m8.705s
```

This PR is a draft on purpose: it stands as the stopgap offered on #5597, for the author of that change to take or replace at their own pace. Nothing here is time-sensitive.

sjarmak (approved):
Reproduced on an untouched worktree at origin/main fc7a23a292: `go test ./scripts/ -run TestResidencyBoundaryGrepRatchet` fails with the matched grow/shrink pair on cmd/gc/infra_class_migrate.go, and `./scripts/check-residency-boundary.sh` exits 1.

Verified the move is a relocation rather than growth. At fc7a23a292^ the single `config.ReservedClassPrefix(config.BeadClassGraph)` call sat in openInfraDestination; #5597 extracted `func infraBindingIDPrefix()` and the call now sits there, at infra_class_migrate.go:594. The file still holds exactly one hit, so the census total is unchanged and the row key is the only thing that drifted. That matches the existing header prose, which already says the row moves if the declaration is renamed.

Applied this diff's re-key to that worktree and re-ran:

- `./scripts/check-residency-boundary.sh` EXIT=0
- `go test ./scripts/ -count=1` ok, 61.929s (covers the grep ratchet plus both AST halves)

Context for landing: main has been red on this guard since fc7a23a292 merged at 15:28Z, so the pre-push gate and every automated preflight have been failing against it for six hours. One other failure shows up in full-suite runs, TestRecordOncePostReservationQueueFailurePersistsDiagnostics, and it passes 3/3 in isolation; that one is contention, not a base defect.

#6061 makes the same one-row change. Taking this one because it landed first and also carries the header prose the baseline's own regeneration procedure asks for.

sjarmak:
Cross-provider review record, run against head 88f7b98 (this PR's tip, unchanged since it opened).

**Leg 1, this reviewer.** Reproduced on an untouched worktree at origin/main fc7a23a292: the grep ratchet fails with the matched grow/shrink pair, and `./scripts/check-residency-boundary.sh` exits 1. Confirmed the site relocated rather than grew: at fc7a23a292^ the single `config.ReservedClassPrefix(config.BeadClassGraph)` call sat in openInfraDestination at line 503; #5597 extracted `func infraBindingIDPrefix()` and it now sits at line 594. A grep of the file returns two lines, the prose comment at 17 and that one call. Applying this diff's row edit to that worktree: `check-residency-boundary.sh` EXIT=0, `go test ./scripts/ -count=1` ok in 61.9s, which covers the grep ratchet and both AST halves.

**Leg 2, Codex, adversarial, clamped to the inlined evidence.** VERDICT: PASS. Its four answers:

1. Re-keying is the correct census repair. The diagnostic detects an unfamiliar key; it does not establish a new enumeration operation, so neither `internal/storeref` adoption nor a `residency:allow` annotation is required by this move alone.
2. It cannot conceal another line matching this pattern, since a separately located matching call would raise the census and fail. It notes the general limit that textual site count and behavioral reach differ, given openInfraDestination and openInfraBindingReadOnly both consume the helper. That is a statement about the guard, not a defect in this diff.
3. A same-count identity move is consistent with the SHRINK ONLY contract. The header already says a row moves if the declaration is renamed, and removing the old row while adding its replacement preserves the allowance of one rather than expanding it.
4. The added header prose is factually accurate. One non-blocking editorial note: it would trim "is #5597's author's call", since architectural obligations sit with maintainers and project rules rather than only the original author.

No blocking findings on either leg. Merging to clear the guard on main, which has been red since fc7a23a292 landed at 15:28Z and has been failing the pre-push gate and every automated preflight since.

sjarmak:
Heads up: the data half of this landed via #6061 (c9b1eaebf, 21:43Z), which makes the same one-row re-key from openInfraDestination to infraBindingIDPrefix. I took that one because this PR is still in draft, so it was not merge-eligible, and main had been red on the guard for six hours with the pre-push gate and every automated preflight failing against it.

What has NOT landed is the part that is only here: the nine lines of header prose explaining why the row moved. The baseline's own regeneration procedure asks for that ("every block below explains why some specific row is pinned"), so it is still worth having. If you rebase, expect the row line itself to drop out as already applied and the prose to be the whole diff.

One note from the review, non-blocking, raised by the Codex leg: it suggested trimming "is #5597's author's call", on the grounds that whether the helper should consume internal/storeref sits with maintainers and project rules rather than only the original author. Your call.

The verification is recorded above and on #6061: with the row edit applied at fc7a23a292, check-residency-boundary.sh exits 0 and `go test ./scripts/ -count=1` passes in 61.9s across the grep ratchet and both AST halves.

quad341:
Closing: the author landed their own fix for this in #6061 (c9b1eaebf), which re-keys the same row. This draft stood only as the offer made on #5597 and is no longer needed. Nothing here was merged.

PR #6063, opened 2026-09-05
fix(events): answer a Since query from the tail instead of scanning all history

remuscazacu:
## The problem

A `Filter` carrying `Since` goes through `readFilteredTracked`, which opens the active log at byte 0 and `json.Unmarshal`s **every record** before `matchesFilter` can reject it. Time is not an index here — it is evaluated per record, after the record has been parsed — so a bare `gc events --since 5m` pays for the entire history to answer a question about the last five minutes, and the cost grows with a file that only ever grows.

#4418 bounded the *backward* walk with `MaxScanBytes`, and its own comment says explicitly that it "has no effect on ReadFiltered's forward scan". This is that gap.

## Measured

On a live 97 MB / 305,071-line active log, same filter, byte-identical answers:

| | time |
|---|---|
| full forward scan (`readFilteredTracked`) | **4.718 s** |
| bounded backward walk (this change) | **4 ms** |

Through the CLI on a loaded box the same query took **19.5 s**, against **1.45 s** for the identical query with `--limit 20` — the limit being the only thing that routed it to the tail path.

`packs/core/assets/scripts/cascade-nudge-on-blocker-close.sh:92` issues exactly the unbounded shape, and it is an order triggered `on = "bead.closed"`. At roughly one such query every three seconds that is more core-seconds per second than the host has cores; it held a 4-core workstation at 0% idle with every agent otherwise idle. A CPU profile of the supervisor put **77% of its time in `encoding/json.Unmarshal`, 69.78% of that under `readFilteredTracked`**.

## The change

`ReadFiltered` walks the active log backward from EOF for a `Since` query and stops once it is past the window. Two things make that safe, and both are the interesting part of the review:

**Seq is globally monotonic; `Ts` is not.** Events are appended by many processes under an flock, each stamping its own wall clock, so a record written later can carry an earlier timestamp than the one before it. Stopping at the first event older than `Since` would drop any event whose clock trails its neighbors. The walk continues until it is a whole `sinceScanSkew` (5 min) past the boundary — a bound on inter-process clock disagreement on one host, not on event ordering.

**The walk reports whether its answer is complete, and only an early stop counts.** Stopping early proves the active log already spans the window: everything unread is earlier in the file and, by seq monotonicity, earlier in the log, so no archive can hold a match. A walk that instead runs out of file has a window reaching back past the active log — it returns unbounded and the caller pays for the full read including archives. Serving a partial answer there is how a rotated city would silently lose events.

`Limit` is applied after reversing to ascending order, because `readFilteredTracked` takes the *first* `Limit` matches chronologically rather than the last.

`readFilteredTracked`, `readFilteredTailFromFile` and `matchesFilter` are untouched — full-history reads, `AfterSeq` walks and the doctor checks keep the path they had.

## Tests

Four, in `reader_since_bounded_test.go`. The bound is asserted on **work done**, not on the answer, since a test that only checked the returned events would pass against the old full scan too.

- stops before reading the whole file (200 KB of ancient history several chunks deep stays unread)
- tolerates clock skew — an event stamped earlier than its predecessor but inside the window is still returned
- reports **unbounded** when the walk reaches byte 0, so archives are still consulted
- equivalence with `readFilteredTracked` across `Limit` 0/1/2/99

`go test -race ./internal/events/` passes (48s); `go vet` clean; pinned golangci-lint v2.12.0 reports 0 issues.

## Adjacent, not fixed here

`defaultRotationMaxSize` is 256 MiB while the reader degrades to unusable around 100 MB, so rotation sits ~2.6× above the size at which the thing it protects has already failed; `archiveOverlapsFilter` skips only on seq, never time; `archiveRetainAge` defaults to zero so archives are kept forever; and the doctor's event-log warning is 100 MB, which the affected city was still under. Each is a separate change.

### DIFF
--- a/internal/events/reader.go
+++ b/internal/events/reader.go
@@ -434,6 +434,22 @@ func readFilteredTracked(path string, filter Filter) ([]Event, map[eventSeqWindo
 // catch-up path; a positive Filter.Limit bounds only ReadFiltered's own scan,
 // not newly discovered rotation sources merged by the recovery pass.
 func ReadFilteredWithInFlight(path string, filter Filter) ([]Event, error) {
+	if !filter.Since.IsZero() {
+		// Same bounded walk as ReadFiltered, and safe here for the same
+		// reason plus one more. A bounded result proves the ACTIVE log
+		// spans the window; an in-flight rotating file is the previous
+		// active log on its way to becoming an archive, so it is older
+		// still and holds no match. The moment that stops being true is
+		// the moment rotation has just installed a fresh, nearly-empty
+		// active file — and then the walk reaches byte 0, reports
+		// unbounded, and this falls through to the full read that does
+		// consult the rotating sources.
+		if result, bounded, err := readFilteredSince(path, filter); err != nil {
+			return result, err
+		} else if bounded {
+			return result, nil
+		}
+	}
 	base, listedArchives, baseErr := readFilteredTracked(path, filter)
 	rotated, rotationErr := readRotationSources(path, filter, listedArchives)
 	if len(rotated) == 0 {
--- a/internal/events/reader.go
+++ b/internal/events/reader.go
@@ -694,10 +694,22 @@ func readFilteredTailFromFile(f *os.File, size int64, filter Filter, limit int) 
 		return nil, nil
 	}
 	const chunkSize int64 = 64 * 1024
+	// A Since filter is a horizon as well as a predicate. Without this the
+	// walk keeps going after it has left the window, looking for a `limit`
+	// it can never reach — matchesFilter rejects everything older — and a
+	// selective Type filter turns that into a full backward parse of the
+	// file at the same cost as the forward scan. That is this walk's half
+	// of the same defect, and it is the half the supervisor's event API
+	// hits: ListTail is what serves `gc events`.
+	horizon := time.Time{}
+	if !filter.Since.IsZero() {
+		horizon = filter.Since.Add(-sinceScanSkew)
+	}
+	pastHorizon := false
 	var reversed []Event
 	var pending []byte
 	end := size
-	for end > 0 && len(reversed) < limit && (filter.MaxScanBytes <= 0 || size-end < filter.MaxScanBytes) {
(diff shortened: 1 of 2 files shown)


remuscazacu:
## Independent profile from a live supervisor, confirming this

Captured today from a supervisor that had been up 2d09h, via `127.0.0.1:6060/debug/pprof/profile?seconds=60`. This is a different instance on a different day from the profile in the description, and it lands on the same numbers:

| | this profile | PR description |
|---|---|---|
| under `readFilteredTracked` | **70.54%** | 69.78% |
| `encoding/json.Unmarshal` | **68.14%** | 77% |

36.22s of samples over 60s — 0.6 of a core, sustained, on a 4-core box.

```
70.54%  events.readFilteredTracked
          <- events.ReadFiltered
          <- events.(*FileRecorder).List
          <- orders.checkEvent
          <- orders.CheckTriggerWithOptions
```

### The part I think is worth adding to the case

The PR frames the cost through the CLI — `gc events --since` without `--limit`, and `cascade-nudge-on-blocker-close.sh:92` issuing that shape as a `bead.closed` order. The profile shows the supervisor paying it on two paths, and **neither is a CLI invocation**:

```
52.05%  main.(*memoryOrderDispatcher).dispatch   <- CityRuntime.tick
47.95%  orders.CheckTrigger                      <- HTTP GET /orders/check
```

Both reach `events.ReadFiltered` through `orders.checkEvent` → `CheckTriggerWithOptions`. So every evaluation of an event-triggered order full-scans the active log: on the supervisor's own tick, and again for every `/orders/check` request.

Active logs on this host are now **119.8 MB** and **98.0 MB** — both at or past the 97 MB the description measured 19.5s against.

That makes this more than a CLI latency bug. In steady state it is the single largest consumer of supervisor CPU on this machine, and it is the mechanism behind a separate long-standing report of the supervisor degrading into a load feedback loop over multi-day runs. The `Since` queries the CLI issues are the visible symptom; the tick paying the same cost on a timer is the part that runs whether anyone is looking or not.

No goroutine growth alongside it: 113 goroutines, 109 parked.

Happy to attach the raw `cpu.pb.gz` if useful.

PR #6066, opened 2026-09-05
Stop observed work directories from overriding session launches

quad341:
## What this changes

Session reconciliation no longer treats the canonical `gc.work_dir` metadata stamp as authority for the next session launch directory. That field records where a session was observed running; feeding it back into launch selection could make a one-time directory divergence permanent.

Creator-owned legacy `work_dir` metadata remains supported, and drain-source directory resolution keeps its existing precedence.

## Review notes

- The production change is confined to `resolveTaskBeadWorkDir`; it narrows one metadata lookup.
- Drain-source behavior, session-directory precedence, configuration, storage schemas, and API shapes are unchanged.
- Regression coverage includes both session-ID and role-name assignment paths, plus no-task, missing-directory, legacy-key, and drain-source controls.

## Test plan

- [x] Run the full local CI-equivalent suite with rootless Podman: `make test-local-full-parallel` (all diff-owned tests pass; unrelated tracked failures and waiver evidence are documented in the release gate).
- [x] Run required policy, boundary, dependency, documentation, formatting, lint, and vet lanes; pre-existing findings are documented in the release gate.
- [x] Run the repository pre-push fast suite; all ten jobs pass.
- [x] Release gate: [`release-gates/ga-rpvye5-workdir-launch-authority-gate.md`](release-gates/ga-rpvye5-workdir-launch-authority-gate.md)

Internal tracking: ga-rpvye5


quad341:
@quad341 — this is a genuinely well-argued fix, and I'm merging it as-is.

What I appreciate is that you went after the *authority* question instead of patching a symptom. `gc.work_dir` records where a session was observed running; feeding that back into the next launch turns any one-time divergence into a permanent one. Your comment in `resolveTaskBeadWorkDir` naming the `taskWorkDirAssignees` scope-widener as a way the divergence seeds itself will save someone a week later.

I traced the producer side before merging and it holds up: every writer that records a created worktree sets both keys together, and the pool stamp already gated on ownership evidence — so the only bead class you stop honoring is exactly the non-pool observed-cwd mirror. Nicest confirmation is that the snapshot resolver in `resolvePreparedTaskWorkDir` was *already* legacy-only, so this makes two paths that quietly disagreed finally agree.

The controls are what made this easy. `NoAssignedTaskKeepsSessionWorkDir` proving the fixture wasn't inert, plus covering both the session-ID and bare-role-name assignee paths, meant I never had to guess whether your assertions were load-bearing. Thanks for that.


---

PR #6068, opened 2026-09-05
perf(reconciler): memoize the orphan-release liveness probe per store

quad341:
## What this changes

`releaseOrphanedPoolAssignments` called `liveOpenSessionAssignmentExists` inside its per-work-bead loop — once for the sessions store, once for the work bead's owner store. That probe's expensive arm is a live `gc:session` listing carrying **no assignee filter**: it lists every session bead in the store and compares identities in Go. Its answer therefore depends only on `(store, assignee)` and cannot change while the sweep runs, because the sweep writes *work* beads, never *session* beads.

This memoizes the probe per store, so the cost is O(distinct assignees) live round-trips instead of O(assigned work beads).

## Why it matters

Measured on the live `gc-management` controller, one traced tick (`cycle-e253e5a702262a90`, 15:22:49Z–15:37:13Z):

```
cycle.finish                                     852987 ms
bead_reconcile_tick                              755922 ms
bead_reconcile.release_orphaned_pool_assignments 652753 ms  (76.5% of tick)
    released_count: 0
demand_snapshot.collect_assigned_work              2860 ms  beads: 69
```

69 assigned work beads x 2 probes = ~138 listings per tick, **releasing nothing**. Across 399 consecutive ticks the inter-tick gap ran min 43s / median 622s / mean 680s / max 2301s against a configured 30s patrol interval.

The observable cost is wake latency, because the phase sits in the critical path ahead of the wake: `load_demand_snapshot` at t=69s, orphan release completing at t=751s, `apply_wake_sleep_decisions` at t=822s, `execute_planned_starts` at t=852s. A sleeping on-demand named singleton waits most of a tick for a wake no matter how promptly the sling poke lands — which is the 8–13 minute band recorded in ga-w83ake.

## The subtle part — why the memo key names the store

Equal store refs mean the same census leg **only** because `collectAssignedWorkBeadsWithStores` appends `assignedWorkStores[i]` and `assignedWorkStoreRefs[i]` from one source. Where that slice is absent, `assignedWorkOwnerStore` falls back to routing each bead through `storeForPoolAssignment(wb)`, and two beads sharing a ref can resolve to **different** stores. Collapsing those onto one cached answer is a claim-level bug in both directions — a stranded orphan, or a live holder's claim released underneath it — so that path stays unmemoized. The second added test pins both directions.

This narrows the staleness window rather than widening it: the three liveness gates now agree on one snapshot, matching `openIdentifiers` and `legacyOpenIdentifiers` which are already built once before the loop, and `liveWorkAssignmentStillReleasable` still re-reads live before any release.

## Test plan

- [x] Two new tests in `cmd/gc/pool_session_name_orphan_release_fanout_test.go` (228 lines) covering the memoized path and the unmemoized fallback in both directions.
- [x] `go test ./cmd/gc -run OrphanRelease` — ok
- [x] `go vet ./cmd/gc/` — clean
- [x] Full pre-push gate (`make test-fast-parallel`, 10 jobs) — **All fast jobs passed**, no `--no-verify`.

Refs ga-451jnv.

### DIFF
--- a/cmd/gc/pool_session_name.go
+++ b/cmd/gc/pool_session_name.go
@@ -196,6 +196,22 @@ func releaseOrphanedPoolAssignments(
 		log.Printf("releaseOrphanedPoolAssignments: assigned work/store-ref length mismatch: work=%d storeRefs=%d", len(assignedWorkBeads), len(assignedWorkStoreRefs))
 	}
 
+	// The live gc:session listing inside liveOpenSessionAssignmentExists carries
+	// no assignee filter — it lists every session bead in the store and compares
+	// identities in Go — so its answer depends only on (store, assignee), and it
+	// cannot change while this sweep runs: the sweep writes WORK beads, never
+	// session beads. Memoize it per store so the cost is O(distinct assignees)
+	// live round-trips instead of O(assigned work beads).
+	//
+	// MEASURED on gc-management 2026-09-05 (ga-451jnv): 68-69 assigned work beads
+	// across 18 distinct assignees re-issued this listing once per bead per store,
+	// costing 645-725s per reconcile tick against a 30s patrol interval — and
+	// releasing 0 beads on every one of those ticks. buildDesiredState runs once
+	// per tick, so that phase alone bounded on-demand named-session wake latency
+	// at ~14 minutes.
+	sessionStoreLiveAssignee := make(map[string]bool, len(assignedWorkBeads))
+	ownerStoreLiveAssignee := make(map[string]bool, len(assignedWorkBeads))
+
 	openIdentifiers := makeOpenSessionStoreRefIndex(cityPath, cfg, store, openSessionInfos, storeRefAware)
 	legacyOpenIdentifiers := make(map[string]struct{}, len(openSessionInfos)*5)
 	for _, info := range openSessionInfos {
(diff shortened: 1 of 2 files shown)


quad341:
@quad341 — this is a good one, and the write-up is what makes it easy to say yes to. Pinning `release_orphaned_pool_assignments` at 76.5% of a tick while releasing *zero* beads, then tracing that through to the 8–13 minute wake band in ga-w83ake, turns "this feels slow" into a decision someone can act on.

The part worth calling out is what you *didn't* memoize. Recognizing that equal store refs only imply the same store because `collectAssignedWorkBeadsWithStores` appends `assignedWorkStores[i]` and `assignedWorkStoreRefs[i]` from one leg — and that the `storeForPoolAssignment` fallback breaks that — is the reasoning that separates a safe cache from a claim-loss bug. I traced it through `censusStoreCandidates` and `censusRef` on my side and it holds. Leaving the fallback unmemoized behind `storeAware && storeRefAware` was the right call.

The tests earned their keep too: asserting count-independence instead of a magic number, and asserting the exact released set in both collapse directions under both orderings.

One thing I'm noting rather than asking you to fix: a cached *negative* now spans the whole sweep, so it widens in scope even as it shrinks enormously in wall-clock. Fine trade at seconds per sweep — worth remembering if anyone lengthens it later.

Merging as-is. Thank you.


---

PR #6070, opened 2026-09-06
Protect container-managed Dolt servers during cleanup

quad341:
## What this changes

`gc dolt-cleanup` now recognizes Dolt SQL Server processes managed by Docker or Podman and protects them from host-PID reaping. Human output explains that the container runtime owns the process, while JSON clients receive the same per-PID reason and runtime in a new additive `reaped.protected` field. The existing `protected_pids` field and `gc.dolt.cleanup.v1` schema identifier remain intact.

Cleanup also recognizes Go test roots under `/var/tmp/gotmp/Test*` and `$GOTMPDIR/Test*`, so fleet test servers are not mistaken for unrelated orphans.

## Design notes

Container membership is a protection-only signal derived from Linux cgroups; it can never make a process eligible for reaping. Missing or unreadable cgroup data retains the existing fail-closed behavior. Test-root paths remain narrowly constrained to a direct `Test*` child of the configured scratch root.

## Review guide

Please pay particular attention to the cgroup marker detection, the ordering of container protection ahead of bare-server data-directory classification, and the additive JSON compatibility surface. The release evidence is recorded in `release-gates/ga-hlp7ut-dolt-cleanup-container-gotmpdir-gate.md`.

## Test plan

- Run the complete local CI union with a rootless Podman socket and testcontainers enabled.
- Run all tests in `cmd/gc/dolt_cleanup_reaper_test.go`, especially the container-managed and `GOTMPDIR` regressions.
- Run `go build ./...`, `go vet ./...`, changed-file formatting, CI policy, boundary checks, and the native DoltLite suite.


quad341:
@quad341 — the container-protection reasoning here is the good kind. "gc doesn't own container lifecycle, so killing the in-container PID leaves a broken container rather than freeing anything" is exactly right, and threading `ContainerRuntime` into a typed `reaped.protected` field instead of bumping `gc.dolt.cleanup.v1` is a considerate way to give operators the *why* without breaking anyone's automation.

One thing I'll fix on my side before this lands. `isTestConfigPath` and `activeTestRootFromPath` are a pair: the first says "this root is ours to reap when orphaned," the second says "but not while the test is still running." The GOTMPDIR half moves the first without the second. I ran your classifier at `e412843`: a live server with `--config` under `/var/tmp/gotmp/TestX/...` now returns `reap`, carrying its `--data-dir` as a removal target; on base it returned `protect`. So `--force` during a test run could kill that test's own Dolt server — the opposite of what `TestGotmpdirTestConfigIsOnAllowlist`'s comment describes.

That's ~6 lines of symmetry, a live-test regression test, a wider cgroup scan for v1 hosts, and refreshing the allowlist line in `--help` plus `cli.md`. Integration work, not a rethink — merging with those on top, credit stays yours.


---

PR #6071, opened 2026-09-06
fix(doctor): retire hold-label route rewrite

### LINKED ISSUE #6069: bug: doctor: do not overwrite gc.routed_to from hold labels
# `gc doctor --fix` overwrites valid routes from orthogonal hold labels

## Summary

The `hold-label-routed-to` doctor check treats `hold:<value>` as a routing
instruction and `--fix` overwrites `gc.routed_to` with the label suffix. That
contradicts the current hold-label and session-model contracts: hold state is
orthogonal to routing, while `gc.routed_to` must name a config identity.

## Symptom

Running `gc doctor --fix` on a bead carrying `hold:mayor` and a valid existing
route such as `gc.routed_to=reviewer` silently changes the route to `mayor`.
Arbitrary labels are more dangerous: `hold:human` or `hold:owner` are converted
into routes named `human` or `owner` even when no such config target exists.

`gc doctor` recommends the mutating command for every mismatch:

```text
run gc doctor --fix to backfill gc.routed_to from the hold:<value> label
```

There is no per-check selector, so users cannot run other doctor repairs while
safely excluding this one.

## Reproduction

Current `main` at `c9b1eaebfadb31ed7a204d8c00ea3f6d5e90e405` already contains
an exact in-memory fixture:

1. Create a bead with `Labels: []string{"hold:mayor"}` and
   `Metadata: map[string]string{"gc.routed_to": "reviewer"}`
   (`cmd/gc/doctor_hold_label_routed_to_test.go:28-32`).
2. Run `holdLabelRoutedToCheck.Fix()`
   (`cmd/gc/doctor_hold_label_routed_to_test.go:81-83`).
3. Read `gc.routed_to`.

Result: expected `reviewer` to remain unchanged; got `mayor`, which the test
currently requires (`cmd/gc/doctor_hold_label_routed_to_test.go:97-103`).

The focused test passes on current main and therefore confirms the destructive
behavior is intentional in the current implementation:

```text
$ go test -count=1 ./cmd/gc -run 'TestHoldLabelRoutedToCheck$' -v
--- PASS: TestHoldLabelRoutedToCheck (0.00s)
PASS
```

## Root cause

`holdLabelValue` accepts every non-empty `hold:*` suffix except `external`
(`cmd/gc/doctor_hold_label_routed_to.go:39-53`). `collect` then defines every
route mismatch as a repair target (`cmd/gc/doctor_hold_label_routed_to.go:96-105`),
and `Fix` unconditionally writes the suffix into `gc.routed_to`
(`cmd/gc/doctor_hold_label_routed_to.go:134-139`).

That inference conflicts with two documented invariants:

- Only `hold:mayor` and `hold:external` are canonical, and hold state is a
  separate pause dimension set with `bd set-state`; it does not touch metadata
  (`engdocs/contributors/hold-label-conventions.md:41-64`).
- `gc.routed_to` is only for config-targeted execution and must contain a
  backing config identity (`engdocs/design/session-model-unification.md:539-548,
  580-587`).

The original check predates the accepted session-model contract and conflates
"who/what must clear this hold" with "which configured execution target should
receive this work."

## Fix candidates

**A. Remove the `hold-label-routed-to` check** — delete the invalid inference,
its registration, tests, and golden name.

- Scope: four files, mostly deletions.
- Trade-off: removes a diagnostic that cannot distinguish drift from valid
  orthogonal state; existing hold-convention and session-model checks remain the
  correct owners for invalid labels and invalid routes.

**B. Replace it with a diagnostic-only hold taxonomy check** — warn on
non-canonical `hold:*` values, set `CanFix=false`, and never write
`gc.routed_to`.

- Scope: the same implementation/test files, with new taxonomy semantics.
- Trade-off: preserves a named diagnostic but overlaps the existing
  `hold-label-conventions` check and still requires a policy decision about
  arbitrary hold-like labels owned by other packs.

**Recommendation:** A. Hold labels and execution routes are deliberately
orthogonal, so no safe route repair can be inferred from a label. Removing the
check restores the documented ownership boundaries with less code and no new
policy surface.

## Adjacent / out of scope

- Migrating a city's non-canonical hold labels is deployment data cleanup, not
  part of the upstream fix.
- Changing the canonical two-value hold taxonomy is out of scope.
- Adding per-check selection to `gc doctor --fix` would be useful defense in
  depth but does not make this repair inference correct.

jacobhausler:
Closes #6069.

## What

- Remove the `hold-label-routed-to` doctor check and its mutating repair.
- Remove its registration, store-preflight inventory/count, tests, and golden
  name.
- Mark the original release gate as superseded while preserving it as
  historical evidence.

## Why

The check conflates two orthogonal dimensions. A `hold:*` label describes who
or what must clear a pause; `gc.routed_to` names the configured execution target
for work. `Fix()` copied every non-`external` hold suffix into `gc.routed_to`,
overwriting valid routes and creating invalid routes from arbitrary labels.

This deletion implements the existing contracts in
`engdocs/design/session-model-unification.md` and
`engdocs/contributors/hold-label-conventions.md`. Existing
`hold-label-conventions` and `session-model` checks retain ownership of invalid
labels and invalid routes respectively.

## TDD and verification

- RED: `TestBuildDoctorChecks_NonUnreachableProbeKeepsStoreChecks` failed on
  commit `bdc2f0cb7` because the unsafe check was still registered.
- GREEN: the focused doctor registration/name/preflight suite passed after the
  deletion (Nomad job `builders/dispatch-1788652405-8794769a`).
- `make build`: pass.
- `golangci-lint fmt --diff`: pass.
- `golangci-lint run ./...`: pass, 0 issues.
- `go vet ./...`: pass.
- routed/split/residency guards: pass.
- `make check-docs`: pass (Nomad job
  `builders/dispatch-1788653451-00d36906`).

The aggregate `make check` run reached tests after every mechanical gate passed.
Its remaining failures are baseline/environment-only under the repository
archive builder:

- `scripts` complexity tests require `.git/HEAD`, which `nomad-run` deliberately
  excludes; two process-tree tests also fail on Haus-Ubuntu. The exact selected
  set fails on clean `origin/main` under the identical runner.
- A full `./cmd/gc` invocation exits nonzero without retained diagnostics on
  both this branch and clean `origin/main`; the complete touched test surface is
  independently green.

## Review

One self-review iteration found no blocker, major, or minor findings. The diff
adds no runtime behavior, config, concurrency, environment, store writes, or
public contract. It removes 374 lines and leaves only the explicit negative
regression plus the superseded historical gate referencing the retired name.

Performance: N/A; this is a CLI doctor point-fix by deletion.


jacobhausler:
Rebased onto current main (2026-09-14). The red checks after the rebase are the bd schema-migrate CI race tracked in #5804: the failing shards die at setup (`[setup failed]` before any test of this branch runs, e.g. internal/runtime/tmux in this run), the same shape that cleared on a plain re-run for #5735 today. Nothing in this PR touches that path. A maintainer re-run should clear it; if any shard stays red after a re-run we will fix it here.

jacobhausler:
Rebased onto upstream main at 6bb1c55866e4145bceb8cc38eab9cf87f765a564 and pushed 0f22ade462e70b31be4bec8ad5fa708b904732e3 with --force-with-lease.\n\nBuilder verification on the rebased head passed: make build, focused doctor coverage, and full make check. The PR check watch settled with the change-relevant gates passing; the remaining failures are infrastructure/test-environment failures (transient Go proxy stream errors and bd/Dolt schema setup). A failed-job rerun was attempted but this seat lacks repository admin rights.

PR #6074, opened 2026-09-06
build: advance Beads current cell to c0d8da4

iamluxe77:
Gas City currently pins Beads `bf97b73749ac` (schema v59), while repositories already opened by current Beads HEAD use schema v66 and cannot be served by that bundle. This advances the current compatibility cell to `c0d8da42de5f`, keeps every source/container identity exact-pinned, and lets Gas City build against the same native store implementation as the current `bd` binary.

The agent image now builds the exact current source tarball with its checked SHA-256 and gRPC 1.83.0. The source version, commit, archive digest, and gRPC pin live in `deps.env`; the security test consumes those anchors instead of duplicating them. The managed-host negative test now uses an unbound loopback address, avoiding VPNs that route TEST-NET while still proving that the host override reaches the liveness probe.

Validation:
- `LOCAL_TEST_JOBS=1 git push`: all fast jobs passed (`unit-core`, six `cmd/gc` shards, concurrency/lock checks, Darwin compile).
- Focused native store tests passed, including real Dolt round-trip, conditional writer, and metadata CAS.
- The exact Beads tarball built as `bd 1.2.2 (c0d8da42de)` with gRPC 1.83.0.
- A Dolt-native backup containing 204 real issues was restored from the v59 bundle and opened by the new binary; owning issue and 125 total steering/work-sync comments matched readback.


iamluxe77:
Closing as superseded by #5830 (merged 2026-09-11), which pins Beads v1.3.0-rc.2. I verified that the schema migration tree (internal/storage/schema/migrations, blob-for-blob) is identical in c0d8da42, v1.3.0-rc.2 and v1.3.0, so stores opened by the c0d8da42 build are served by current main without another pin change. The container/deps.env changes here are no longer needed for that goal.

Issue #2574, opened 2026-05-24
session reset chained with drain-ack leaves restart_requested=true, causing phantom second restart via cache-reconcile


realies:
## Observed

On 2026-05-24 a single `gc session reset mayor` produced **two restart cycles** on the same session bead, the second of which lost session_key continuity. The mayor agent's conversation context was destroyed even though `resume_flag: --resume` and `resume_style: flag` were set in the bead metadata throughout.

The chain (all timestamps `+03:00`, event seqs from `events.jsonl`):

| seq | ts | actor | event | session_key | gen | restart_requested |
|-----|----|----|----|----|----|----|
| — | 19:36:30 | mayor | bead.updated (set restart_requested via `gc session reset mayor`) | `60b25659…` | 60 | **true** |
| — | ~19:44 | mayor | ran `gc runtime drain-ack` (operator-prompted) | `60b25659…` | 60 | true |
| 1033307 | 19:45:32 | gc | session.stopped \"drain acknowledged by agent\" | `60b25659…` | 60 | true |
| 1033311 | 19:45:41 | gc | session.woke | `60b25659…` | 61 | true |
| — | 19:46:12 | controller | bead.updated state=awake | `60b25659…` | 61 | **true (not cleared)** |
| 1034092 | 20:10:41 | cache-reconcile | bead.updated (re-emit with `restart_requested=true` still present in store) | `60b25659…` | 61 | **true** |
| — | 20:11:56 | controller | bead.updated state=awake | `60b25659…` | 61 | true |
| — | 20:11:58 | controller | bead.updated (second restart cycle) | **`fe08328a…` (new uuid)** | 61 | \"\" (finally cleared) |
| 1034155 | 20:13:29 | gc | session.woke (fresh, no resume continuity) | `fe08328a…` | 62 | \"\" |

The second cycle at 20:11 was triggered ~25 minutes after the first, with no operator action in between. The `cache-reconcile` bead.updated at 20:10:41 re-emitted the bead with `restart_requested=true` still present in the persistent store — the controller then honored that stale flag as a new restart request.

## Why this matters

Even with [#2095](https://github.com/gastownhall/gascity/pull/2095) merged (continuation_reset preserves session_key), a session_key swap on the *second* restart in a chained-reset sequence bypasses `--resume` because the new claude-code session has no prior jsonl to resume from. This silently destroys ~hours of operator conversation context.

[...]

PR #6077, opened 2026-09-06
Use sleep time to preserve wake fairness

quad341:
## What this changes

Session wake scheduling now uses the most recent sleep time when `last_woke_at` is missing or malformed. Sessions that were just drained or slept therefore keep their true place in the wake-budget queue instead of falling back immediately to an old creation timestamp and jumping ahead of sessions that have waited longer.

The implementation reuses the existing persisted `slept_at` metadata and projects it through `session.Info`. Wake fairness remains deterministic: `last_woke_at`, then `slept_at`, then creation time.

## Review notes

- `AcknowledgeDrainPatch` now stamps `slept_at` while clearing `last_woke_at`, matching the existing sleep transition.
- The reconciler supplies its UTC clock value to the drain transition.
- There is no new metadata key, storage migration, API shape change, or configuration change.
- Same-tick candidates with equal sleep times preserve stable input order.

## Test plan

- [x] Run the documented 40-job local full suite and confirm every modified session test passes by name.
- [x] Run the repository policy lane, vet, changed-file formatting, and pre-push fast suite.
- [x] Exercise sleep, drain acknowledgement, malformed/missing timestamps, and same-tick ordering through the characterization tests.
- [x] Release gate: [`release-gates/ga-x7powe-slept-at-wake-fairness-gate.md`](release-gates/ga-x7powe-slept-at-wake-fairness-gate.md)


quad341:
@quad341 — nice piece of tracing here. The `slept_at` middle tier is the right shape, but the part that earns this PR is finding that `AcknowledgeDrainPatch` was the *actual* production gap: it cleared `last_woke_at` without ever stamping a sleep time, so drain-acked sessions dropped straight to the `CreatedAt` floor and jumped the queue. That only turns up if you follow the dominant real-world path rather than the one the existing test happened to model, and the new drain-ack scenario in the pin test pays for itself.

I also appreciate that you didn't quietly flip the `#2574` sort assertion — spelling out that the old order was really just FIFO-by-creation made it reviewable instead of alarming. I traced the `#2574` references myself and agree this doesn't reopen it.

Merging this. Two small things I'll take on my side: `usage_compute.go` and `invocation_telemetry.go` both carry comments saying `slept_at` is stale for drained sessions *because drain doesn't refresh it* — true until this PR, so I'll update them. I'm also filing a follow-up for the quarantine and restart-handoff patches, which still clear `last_woke_at` with no `slept_at` stamp. Same class of gap, separate change.


---

PR #6085, opened 2026-09-06
feat(runtime): add provider-native remote session contract

### LINKED ISSUE #6078: feat: add a remote-session transport contract for native coding clouds
## Before you continue

- [x] I can explain the user problem this solves, not just the implementation I prefer.

## Problem to solve

Gas City can orchestrate local and self-hosted session runtimes, but a provider's native SaaS coding task is not currently an addressable Gas City session.

For example, `codex cloud exec --env ...` and `claude --cloud ...` create remote work that can continue after the operator machine is offline. The launcher process then exits. Treating that command as an ordinary tmux/exec session loses the remote run identity and makes liveness, follow-up steering, cancel, transcript readback, restart adoption, and ownership fences impossible to enforce through the existing Session surface. Cursor Cloud Agents have the same lifecycle shape.

This prevents an operator from using subscription-authenticated provider clouds while keeping Beads as the work authority and Gas City as the one orchestration engine.

## Proposed change

Define a provider-neutral remote-session transport extension behind the existing runtime/session boundaries. A remote transport should return and persist an opaque provider run ID in the session bead, then implement the existing lifecycle semantics through a small capability contract:

- create/attach and idempotent restart adoption;
- liveness/status with terminal-reason classification;
- follow-up/nudge with an acknowledgement cursor or receipt;
- bounded transcript/result readback with a stable event cursor;
- cancel/stop and close readback;
- branch/PR handoff metadata;
- explicit capability discovery so unsupported operations fail closed.

Provider account, plan, quota, and billing policy should stay outside the generic wire, consistent with the hosted-service boundary. Credentials remain provider-managed and must not be persisted in session metadata or logs. The transport owns only provider run lifecycle; Beads remains the task graph and Gas City remains the dispatcher/controller.

An exec/RPP adapter process could implement this contract if that is a better boundary than provider-specific code in core. Codex, Claude, and Cursor adapters should then be separately conformance-tested against the same remote-session profile.

## Alternatives considered

- Launching cloud commands inside tmux: the launcher exits and cannot provide durable remote liveness or steering semantics.
- Polling each provider from an organization-specific controller: creates a second orchestration runtime and duplicates session state.
- Calling model APIs directly: bypasses the native coding harness and subscription authentication.
- Treating GitHub Issues as the runtime state: turns a human projection into a second task/session database.

## Scope and impact

This should be additive and opt-in. Existing tmux, exec, subprocess, ACP, k8s, herdr, and T3 transports should keep their behavior.

Suggested conformance coverage:

1. create returns a stable opaque identity and duplicate create is idempotent;
2. restart adopts the same remote run without creating a second task;
3. acknowledged follow-up survives controller restart and can be read back;
4. out-of-order or repeated provider events do not advance the cursor twice;
5. stale ownership cannot nudge, cancel, or finalize a reassigned session;
6. terminal, quota, auth, and network failures are distinct and observable;
7. transcript/result readback is bounded and secrets are redacted;
8. unsupported provider capabilities fail closed.

Relevant native surfaces:

- Codex Cloud CLI: https://learn.chatgpt.com/docs/developer-commands#codex-cloud
- Claude Code cloud sessions: https://code.claude.com/docs/en/claude-code-on-the-web#send-follow-ups-from-the-cli
- Cursor Cloud Agents API: https://cursor.com/docs/cloud-agent/api/endpoints

Agent-Signature: codex-gpt-6 on behalf of iamluxe77

iamluxe77:
Closes #6078.

Gas City can currently drive local/process runtimes, but it has no provider-neutral contract for native coding-cloud sessions. That forces each provider integration to invent its own lifecycle, identifiers, retries, and persistence rules.

This adds an optional `runtime.RemoteSessionProvider` extension and matching RPP v0 operations for create/adopt/status/follow-up/transcript/cancel/close. Each operation is independently capability-gated and fails closed when undeclared. Mutations carry idempotency IDs and ownership fences; responses use typed success/error envelopes with normalized phases and stable auth/quota/network/policy/ownership/not-found/cursor/provider classifications.

The exec adapter validates and redacts the wire boundary. Provider IDs and cursors stay opaque, transcript pages are bounded and read-through, and credential-bearing repository/PR URLs are rejected. Session beads persist one versioned content-free checkpoint under the central beadmeta vocabulary, fenced by the current instance token; prompts, transcripts, provider messages, credentials, plan/billing data, and quota amounts are excluded.

The runtime conformance catalog adds independently gated `RPP-REMOTE-001` through `RPP-REMOTE-007` probes and negative mutants for idempotency, identity continuity, classified errors, stale-owner rejection, bounded transcripts, and terminal cancel/close behavior. Provider-specific Codex, Claude, and Cursor adapters can now ship as runtime Packs against this contract.

Validation:
- `go test ./internal/runtime ./internal/runtime/exec ./internal/runtime/runtimecontract`
- `go test ./internal/beadmeta ./internal/session` with Homebrew ICU flags
- `make test-fast-parallel`
- `make vet`
- repository pre-commit and pre-push hooks


iamluxe77:
Maintainer follow-up: the latest head a612829b09976b6f82e0b04abc41410d99a43346 is mergeable, but the only failing check is the evidence watchdog because the required suites never ran. Run 35211006289 reports “tests never ran”; there is no test failure to diagnose.

Could a maintainer approve the fork workflows (CI, CodeQL, complexity, review-formulas, govulncheck, docs-render, and mac-regression) for this head? We will not use the PR or a fork as production authority; adoption remains blocked on an official reviewed stable release.

iamluxe77:
Closing the core PR. Following the runtime-cloudflare precedent, we'll ship this as a registry runtime pack over the exec session provider (RPP v0) and only open a narrow issue if we hit a concrete missing operation. Thanks for the review time.

Issue #5883, opened 2026-09-01
bug: `gc hook --claim` stamps `gc.work_branch` on work_dir-only beads, turning #5783's legacy shape into partial worktree evidence that starves pool spawn


sJohnsonStoever:
### Before you continue

- [X] I searched existing issues and did not find a duplicate.
- [X] I read the relevant docs and contributor guidance.

### Gas City version

`main@615f5b7942220ee02f6825b9d3d52b7b4b9e9224` — `gc version --long`:
`edge (commit: 615f5b794, built: 2026-09-01T18:39:31Z)`, source build.
Includes #5783 (`bc1c3ccaf`). Beads at `bf97b73`.

### Environment

macOS 26.6.2 (Darwin 25.6.0), zsh, native binaries (no container/Kubernetes).
Runtime providers in the affected path: codex (the claiming polecat pool) and
claude (the starved architect pool). Managed dolt 2.2.2 sql-server, three
stores (city + two rigs).

### Reproduction

1. Mint a work bead with `gc.work_dir=<path>` and none of the nine worktree
   ownership keys — the "legacy" shape #5783 explicitly allows (per the
   comment at `cmd/gc/pool_desired_state.go:108-112`, recipe steps from
   `stampDrainItemRecipe` are minted exactly this way).
2. From a session whose cwd is a git worktree on a branch that differs from
   the bead's recorded branch, run `gc hook --claim` on the bead.
   `hookClaimIdentityPatch` (`cmd/gc/cmd_hook_claim.go:1175-1180`) stamps
   `gc.work_branch`.
3. Route the bead to a pool template with a spawn-for-bead trigger (a
   different seat).
4. Watch the supervisor log.

### Expected behavior

The bead keeps spawning its next seat as an unmanaged workspace, as #5783
intended for work_dir-only beads. A claim made through gc's own hook should
not change that classification.

### Actual behavior

After the first claim the bead carries `gc.work_dir` + `gc.work_branch` and
nothing else. `worktreeSpecForBead` now sees 1-of-9 ownership keys, takes the
partial-evidence branch (`pool_desired_state.go:121-124`) and returns the
`missing` error; `buildDesiredState` skips the pool trigger on every tick.
`poolDesired` stayed at 1 for 8 minutes (20 ticks) with no session created
and no self-heal. Nothing but the supervisor log surfaced it. Manual unblock:

```
bd update ng-1649 --unset-metadata gc.work_branch
```

[...]

PR #6088, opened 2026-09-06
fix(pool): treat work-branch-only evidence as unmanaged

quad341:
## What this changes

Pool reconciliation now treats a work bead with `gc.work_dir` plus only the ambient `gc.work_branch` field as unmanaged work instead of rejecting it as malformed partial worktree ownership. This prevents an otherwise schedulable pool seat from being starved by bookkeeping that does not identify a managed worktree.

Hook claims also avoid adding `gc.work_branch` when a bead already has an incomplete managed-worktree evidence set. They continue to stamp it when the evidence set is wholly absent or wholly present, so the producer no longer creates the ambiguous shape the reader must reject.

## Review notes

- The reader-side exception is deliberately narrow: `gc.work_branch` must be the only ownership key present. Every other partial ownership shape remains a hard error.
- Existing affected beads recover on their next reconciliation read; there is no migration, sweep, schema change, or new metadata key.
- The hook guard checks key presence only. It does not read, rewrite, or clear work-record values.

## Test plan

- [x] Run the documented full local suite with rootless Podman enabled; confirm every changed test executes and inspect each attributed infrastructure failure against its tracker.
- [x] Exercise the zero, one, seven, and all-eight ownership-key boundaries plus the reader's accepted and rejected shapes.
- [x] Run build, vet, CI-policy, documentation, native DoltLite, and architectural boundary checks.
- [x] Review the [release gate](release-gates/ga-n3zvcg-work-branch-unmanaged-gate.md).

---
- Related to #5883 — Makes pool reconciliation treat a work bead carrying gc.work_dir plus only the ambient gc.work_branch as unmanaged rather than partial worktree evidence, and stops hook claims from stamping that branch onto beads that already hold incomplete ownership evidence, so a claimed bead's seat spawns unmanaged instead of starving on every tick. The separately suggested starvation-visibility signal (a doctor finding or an event for repeatedly skipped pool triggers) is not part of this change.

### DIFF
--- a/cmd/gc/cmd_hook_claim.go
+++ b/cmd/gc/cmd_hook_claim.go
@@ -1377,7 +1377,8 @@ func hookClaimLifecycleCandidate(bead beads.Bead, opts hookClaimOptions) bool {
 func hookClaimIdentityPatch(bead beads.Bead, opts hookClaimOptions, ops hookClaimOps, dir string) map[string]string {
 	patch := map[string]string{}
 	if branch := strings.TrimSpace(ops.ResolveWorkBranch(dir)); branch != "" &&
-		strings.TrimSpace(bead.Metadata[beadmeta.WorkBranchMetadataKey]) != branch {
+		strings.TrimSpace(bead.Metadata[beadmeta.WorkBranchMetadataKey]) != branch &&
+		hookClaimWorktreeEvidenceIsWholeOrAbsent(bead) {
 		patch[beadmeta.WorkBranchMetadataKey] = branch
 	}
 	if sessionID := hookClaimSessionID(opts.Env); sessionID != "" &&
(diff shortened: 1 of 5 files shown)


quad341:
@quad341 — thanks for this, and specifically for fixing *both* ends of it. The easy version of this PR is the reader carve-out alone; noticing that the hook claim's ambient branch stamp is the thing manufacturing the ambiguous shape, and adding the whole-or-absent guard so the producer stops creating it, is what turns a patch into an actual fix for the starved seat.

The test I appreciated most is `TestWorktreeSpecForBeadRejectsWorkBranchPlusOneOtherKey`. The obvious wrong implementation here is "if `work_branch` is present, wave it through," and that test is precisely the tripwire for it — you wrote the test that catches the version of this change you didn't write. Same energy in the 0/1/7/8 boundary table on the claim side.

Merging as-is. Two things purely for whenever you're next in these files, not holding anything: the comment at `pool_desired_state.go:127` still says the hook stamps `gc.work_branch` regardless of ownership evidence, which this PR makes no longer quite true. And nothing in-tree currently writes the eight ownership keys — if the publisher ever lands and writes them non-atomically, the branch-only shape could briefly read as unmanaged, so worth asserting a single atomic update there.


---

Issue #4390, opened 2026-07-17
session list: per-session live enrichment is serialized by client-go default QPS — /sessions grows O(N) and times out at fleet scale


314clay:
`GET /v0/city/{city}/sessions` enriches every active session with live runtime state (`State()` → `Pending()` → pod lookups, plus active-bead searches). Costs we measured on a k8s city:

- ~21 active sessions → 40–49s per request (serial enrichment in `humaHandleSessionList`)
- parallelizing enrichment (bounded errgroup, limit 8) did NOT help: the controller's client-go rate limiter (default QPS 5 / burst 10) serializes the pod GETs — logs fill with `"Waited before sending request" reason="client-side throttling, not priority and fairness"` — and at 34 sessions the request took ~70s
- downstream clients that only need a roster (alias/state, e.g. a mail recipient list) time out entirely

Two directions that would fix the shape rather than the constant:
1. **Cheap roster mode** — we run a local patch adding `?enrich=false` (read-model-only list, zero runtime calls; answers in ~20ms at 34 sessions and stays flat). Happy to PR it.
2. **Informer/lister-backed pod state** for enrichment (or at least raised QPS/Burst on the controller's client), so live probes don't fan out one apiserver GET per session per request.

Filed from a production city where the fleet is expected to grow several×; at that size the current path is minutes per request.

PR #6089, opened 2026-09-06
perf(sessionlog): cache pi transcript headers by stat signature

atbrace:
## Problem

`findPiSessionCandidatesIn` walks the whole pi sessions tree and opens + JSON-parses the first line of **every** `.jsonl` on **every** candidate scan. The scan runs per session row per `GET /sessions` request (via `ResolveKeyedTranscriptPaths` → `DiscoverKeyedPath` → `FindPiSessionFileByID`), so a page of N pi sessions over a tree of M transcripts costs N×M file opens per request — repeated for every request, even though transcript headers essentially never change after the first line is written.

Measured on a loaded 6-core host (16 active sessions, 152 transcript files): `GET /v0/city/{city}/sessions` took **29–43s per request**, with `sample` showing the time in `sessionlog.piSessionHeader` / `findPiSessionCandidatesIn` (walkDir + per-file open + `bufio.Scan` + `json.Unmarshal`, plus the 64KB scanner buffer allocation per file feeding GC pressure).

## Fix

Memoize the parsed header per path, keyed by the file's `(size, mtime)` stat signature:

- unchanged file → one `stat` (already needed for `modTime`), no open/parse
- rewritten file → new signature → re-parse
- deleted file → drops out naturally, because the walk still drives candidate membership; the cache only short-circuits the header read

The map is bounded (`piHeaderCacheMaxEntries`) and resets past the cap rather than growing forever on churned paths.

## Result

Same host, same workload, after deploying this change: `GET /sessions` **42s → 1.2–3.1s**; 1-minute load fell 33 → 12 within two minutes of cutover. Keyed pi transcript resolution verified end-to-end afterwards (`gc session logs <pi-session>` streams the correct transcript).

## Tests

- `TestFindPiSessionFileByIDSkipsRereadWhenStatSignatureUnchanged` — proves the cache actually short-circuits the read (in-place rewrite with identical size+mtime still serves the cached header)
- `TestFindPiSessionFileByIDInvalidatesCacheOnChangedSignature` — rewrite with a new signature re-parses
- `TestFindPiSessionFileByIDDropsDeletedFileDespiteCache` — deletion still removes the candidate

Related: #4390 tracks the same endpoint's O(N) growth from live runtime enrichment — different mechanism (client-go QPS on pod lookups); this PR removes the transcript-discovery term, and is kept deliberately small and standalone.

### DIFF
--- a/internal/sessionlog/pi_reader.go
+++ b/internal/sessionlog/pi_reader.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/gastownhall/gascity/internal/fsys"
@@ -167,12 +168,12 @@ func findPiSessionCandidatesIn(root, workDir string) []piSessionCandidate {
 		if !strings.HasSuffix(strings.ToLower(entry.Name()), ".jsonl") {
 			return nil
 		}
-		sessionID, cwd := piSessionHeader(path)
-		if cleanPiWorkDir(cwd) != workDir {
-			return nil
-		}
 		info, err := entry.Info()
 		if err != nil {
+			return nil
+		}
+		sessionID, cwd := cachedPiSessionHeader(path, info)
+		if cleanPiWorkDir(cwd) != workDir {
 			return nil
 		}
 		if sessionID == "" {
@@ -185,6 +186,48 @@ func findPiSessionCandidatesIn(root, workDir string) []piSessionCandidate {
 		return nil
 	}
 	return candidates
+}
+
+// piHeaderCacheEntry memoizes the parsed first-line header of a pi session
+// file. size+modTime form the stat signature: a rewritten transcript gets a new
+// signature and re-parses, while list/enrichment sweeps that revisit an
+// unchanged tree skip the per-file open+parse entirely.
+type piHeaderCacheEntry struct {
+	size      int64
+	modTimeNS int64
+	sessionID string
+	cwd       string
+}
+
+var (
+	piHeaderCacheMu sync.Mutex
+	piHeaderCache   = make(map[string]piHeaderCacheEntry)
+)
+
+// piHeaderCacheMaxEntries bounds the cache; deleted transcripts leave stale
+// keys behind, so past this size the map resets rather than growing forever.
+const piHeaderCacheMaxEntries = 16384
+
+func cachedPiSessionHeader(path string, info os.FileInfo) (string, string) {
+	size := info.Size()
+	modTimeNS := info.ModTime().UnixNano()
+
+	piHeaderCacheMu.Lock()
+	entry, ok := piHeaderCache[path]
+	piHeaderCacheMu.Unlock()
+	if ok && entry.size == size && entry.modTimeNS == modTimeNS {
+		return entry.sessionID, entry.cwd
+	}
+
+	sessionID, cwd := piSessionHeader(path)
+
+	piHeaderCacheMu.Lock()
+	if len(piHeaderCache) >= piHeaderCacheMaxEntries {
+		piHeaderCache = make(map[string]piHeaderCacheEntry)
+	}
+	piHeaderCache[path] = piHeaderCacheEntry{size: size, modTimeNS: modTimeNS, sessionID: sessionID, cwd: cwd}
+	piHeaderCacheMu.Unlock()
+	return sessionID, cwd
 }
 
 func parsePiFileDetailed(path string) ([]piEntry, string, SessionDiagnostics, error) {
--- a/internal/sessionlog/pi_reader.go
+++ b/internal/sessionlog/pi_reader.go
@@ -206,6 +206,9 @@ var (
 
 // piHeaderCacheMaxEntries bounds the cache; deleted transcripts leave stale
 // keys behind, so past this size the map resets rather than growing forever.
+// The reset is an anti-leak cap, not a working-set-preserving eviction policy:
+// a tree with more live transcripts than this bound wipes mid-sweep and stops
+// caching. Real eviction is out of scope here and tracked in ga-gn1gf.
 const piHeaderCacheMaxEntries = 16384
 
 func cachedPiSessionHeader(path string, info os.FileInfo) (string, string) {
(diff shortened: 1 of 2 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-699284130204878924

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @atbrace! This PR memoizes pi transcript headers by
their `(size, mtime)` stat signature so `findPiSessionCandidates` stops
re-opening and re-parsing the first line of every `.jsonl` file on every sweep.
That matters because the sweep is per-session-row: it is funnelled into from
`reader.go:550,582,623`, `internal/session/submit.go:493-508`, and
`internal/worker/transcript/discovery.go:62`, so the old per-row × per-file
`open`+parse cost was structural and grew with transcript count. Reviewers traced
those call paths independently rather than taking the commit message's word for
it, and the hot path is real. You also picked the stat-signature idiom the
codebase already uses in five sibling sites instead of inventing one, and you
pinned the same-signature staleness trade as an explicit contract test — that is
the right way to land a cache.

branch.

### Original PR Review

Specific gaps fixed:
- **Transient read failures were being negative-cached as an authoritative "no
  header".** `piSessionHeader` returned `("", "")` identically for a failed
  `os.Open`, an empty file, an unparsable first line, and a non-`session` first
  record, and never consulted `scanner.Err()`. Before the cache that cost one
  sweep and the next call retried; with the cache it froze under a signature
  that stays valid until size or mtime changes — and an idle transcript awaiting
  resume is exactly the file nothing appends to, so it would never self-heal.
  The final patch adds a `read bool` return that is `false` only when the open
  errors or `scanner.Err() != nil`, and `cachedPiSessionHeader` skips the store
  when `!read`. Content-derived negatives stay cached, so the secondary win your
  change delivers on non-pi `.jsonl` files is preserved intact.
- **That same negative caching could invert the same-workdir ambiguity guard.**
  `FindPiSessionFileStrict` returns `ErrAmbiguousPiSessionFile` only at two or
  more candidates, so poisoning one of two same-workdir transcripts dropped it
  from the candidate set and left exactly one — making the function confidently
  return the other, defeating a guard that exists so callers do not guess which
  same-workdir transcript is safe to mutate. On the keyless `submit.go:508`
  branch that flows into `ResetPiInterruptedTurn`, which truncates the file.
  `TestFindPiSessionFileStrictStaysAmbiguousAfterTransientReadFailure` now pins
  that it stays ambiguous.
- **The 16384-entry bound reset wiped the working set on an ordinary refresh.**
  The capacity check ran before every store, including re-storing a key already
  present — a write that cannot grow the map. It is now gated on `!exists`, and
  the comment documents the reset as an anti-leak cap rather than an eviction
  policy.
- **The reset branch had no test and no injection point.**
  `storePiHeaderCacheEntry(path, entry, maxEntries)` now takes the bound as a
  parameter, so that branch is driven with a cap of 4 instead of needing 16384
  real files.

Review findings addressed:
- Error Handling & Resilience / major / high confidence (Claude + Gemini) —
  transient read failures negative-cached as authoritative absence, including
  the ambiguity-guard inversion. Fixed in 5c5d2ac0.
- Resource Lifecycle & Cleanup / minor / high (Claude + Gemini) — the bound
  reset discarded the working set on a same-key refresh. Fixed in 5c5d2ac0.
- Test Evidence Quality / minor / high (all three lanes) — the wholesale-reset
  branch was unreachable from tests. Fixed in 5c5d2ac0 by parameterising the
  bound.
- Debuggability & Operability / nit / high (Claude) — the `read` flag's doc
  comment enumerated only the content-derived `true` cases and omitted the
  success return, which a later reader could take as inverting the flag. Fixed
  in 5c5d2ac0.

Those tests were mutation-probed rather than merely run: neutralising the
`!read` guard fails `RetriesAfterUnreadableFile`, `RetriesAfterScanError`, and
`StrictStaysAmbiguousAfterTransientReadFailure` while
`CachesNonSessionFirstLine` keeps passing, and neutralising the `!exists` guard
fails `TestStorePiHeaderCacheEntryResetsOnlyForNewKeyAtCapacity`. Your three
original tests were preserved, not loosened — both assertions moved intact into
the extracted `rewritePiFilePreservingSignature` helper.

One transparency note: no reviewer reproduced the commit message's 29–43 s
figure, so it stands as unverified. It does not change the outcome — the
structural per-row × per-file argument holds on its own, and `-race` did
reproduce the 10.5 s suite timing the fixup claims.

- `ga-gn1gf`: Resource Lifecycle & Cleanup / nit — above the 16384 bound,
  new-key inserts still wipe the map mid-sweep, so a tree with more live
  transcripts than the bound stops caching. Worst case only degrades to today's
  cost and is never a regression against `main`, so real eviction is
  deliberately out of scope here; the code comment now states the trade.
- `ga-ibnd5`: Change Impact / Blast Radius / nit — the sibling opencode reader
  keeps the same per-sweep parse this PR removes for pi, and reads the whole
  file rather than the first line. Scoped to opencode only; zcode filters on the
  filename before parsing, so it is not an offender.

Two deliberate trades worth naming, both intended and neither owed work: a
permanently unreadable `.jsonl` file is now re-`open`ed every sweep instead of
cached once (precisely the pre-cache behaviour being restored), and during an
active failure window `FindPiSessionFileStrict` still returns the single
readable candidate of a same-workdir pair — pre-existing `main` behaviour, since
only the cross-sweep persistence was removed.

### Maintainer Changes
One maintainer fixup pushed to this PR branch: `fix(sessionlog): only memoize pi
header reads that actually read the file` (5c5d2ac0) — 2 files changed, 243
insertions(+), 30 deletions(-), adding five tests on top of your three. Your
authorship on `perf(sessionlog): cache pi transcript headers by stat signature`
is preserved.

### Final Review Status
Ready for the merge queue: final head `5c5d2ac0` has passing required GitHub
checks (66 successful, 22 skipped, none failing or pending) and the merge-ready
metadata is in place. Locally, the adopted branch was also re-verified against
current `main` after four intervening commits — `go build ./...`, `go vet` and
`gofmt` clean, and `go test ./internal/sessionlog/ -count=1 -race` green at
10.4 s, with both touched files byte-identical to the reviewed head. I am
marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/34558011465

### Review Iterations
2 review passes performed. Iteration 1 (three independent reviewers) requested
changes on one major error-handling finding plus three non-gating items; the
maintainer fixup resolved all four, and iteration 2 re-reviewed the delta, found
no new gating issues, and approved.

---

Issue #5742, opened 2026-08-28
bug: doctor misses dead always-mode named sessions while the controller runs


swedeinasia-flow:
### Before you continue

- [x] I searched existing issues and did not find a duplicate.
- [x] I read the relevant docs and contributor guidance.

### Gas City version

`main@655713ff4559c335ff17ec09c4a65f5cca2a5f27`

### Environment

macOS 27.0 arm64, zsh, tmux runtime provider, running controller, configured named agent with `mode = "always"`.

### Reproduction

1. Start a city with its controller running and at least one active named agent configured as `mode = "always"`.
2. Kill only that named session's runtime, leaving the controller alive and the agent unsuspended.
3. Run `gc doctor --json`.
4. Inspect the registered checks and findings for the missing required runtime.

Current source registers Agent/Zombie/Orphan session checks only inside [`!controllerRunning`](https://github.com/gastownhall/gascity/blob/655713ff4559c335ff17ec09c4a65f5cca2a5f27/cmd/gc/cmd_doctor.go#L260-L276). When the controller is running there is no equivalent direct liveness check for required always-mode named sessions.

### Expected behavior

Doctor should run a bounded, suspension-aware liveness check for configured always-mode named sessions while the controller is running. A confirmed missing runtime should produce a finding and targeted recovery hint. A timed-out or incomplete process observation should warn that observation is incomplete rather than asserting death.

### Actual behavior

The controller and supervisor checks can pass while the required named runtime is absent, and Doctor emits no session-liveness finding for it.

### Logs, screenshots, or traces

```shell
cmd/gc/cmd_doctor.go:260-276
runtime session checks are registered only when controllerRunning == false
```

### Additional context

Existing issues around reset-pending/config drift concern a live runtime and are not duplicates of this missing-runtime diagnostic gap. A downstream red/green patch adds a bounded check with coverage for running, confirmed missing, authored suspension, live city/rig suspension, and partial-probe behavior.

quad341:
Investigated this as a follow-up spike, distinguished from the startup-health episode work already delivered for reset-pending/config-drift (unrelated mechanism, no code overlap).

[...]

PR #6091, opened 2026-09-06
runtime: add bounded liveness observation

quad341:
## What this changes

This adds an additive runtime helper that reports both liveness and whether the observation completed reliably within a caller-supplied timeout. Callers can now distinguish confirmed absence from an observation that was incomplete because the runtime was unavailable or the deadline expired.

Runtime providers may opt into richer, error-bearing liveness observation. Existing providers retain their current plain-boolean behavior through a compatibility fallback.

## Review notes

- The new surface is `ObservationStatus`, `BoundedLivenessObserver`, and `ObserveLivenessBounded` in `internal/runtime`.
- A deadline returns zero liveness with `Incomplete`; `ErrRuntimeUnavailable` preserves any partial liveness with `Incomplete`; other completed observations remain `Complete`.
- Existing call sites are unchanged. This is groundwork for consumers that need bounded, tri-state observation.
- The timeout bounds the caller's wait. A provider goroutine that ignores cancellation may outlive the call, matching the package's established bounded-observation trade-off.

## Test plan

- [x] Run the full sharded local suite with container-backed tests enabled (`make test-local-full-parallel`).
- [x] Run policy, build, vet, formatting, and lint lanes; attribute reproduced baseline-only failures in the gate record.
- [x] Exercise compatibility fallback, richer observers, unavailable runtime, non-runtime errors, and deadline expiry.
- [x] Release gate: [`release-gates/ga-08v565-observe-liveness-bounded-gate.md`](release-gates/ga-08v565-observe-liveness-bounded-gate.md)

---
- Related to #5143 — adds an additive tri-state liveness observation API in internal/runtime (ObservationStatus / BoundedLivenessObserver / ObserveLivenessBounded) that distinguishes an unobservable runtime from confirmed absence — the "uncertainty channel" the filed issue names as missing from the plain-bool IsRunning contract. It is groundwork only: the herdr provider, its Stop path, and the runtimetest conformance suite are untouched, so both conformance failures described there still reproduce.


jacobhausler:
**Consumer survey, re-checked at `main` = `9d59e5c1041805a78c0f53f8923cc73ff549ba1f` — the seam is sound and nothing in-tree consumes it yet, and the one site in the tree that still collapses "could not ask" with "not running" for pool occupancy is production-dead.**

We came from the pool-occupancy side of this (a provider that cannot be asked makes every expected
pool slot look free, so the controller duplicate-starts the pool), were going to file the
`countRunningPoolInstances` fail-closed half as a P1 citing this PR as the groundwork it does not
consume, and stopped when we checked reachability:

- `countRunningPoolInstances` (`cmd/gc/cmd_start.go:1650`) does still collapse the two answers —
  `cmd_start.go:1658` and `:1678` are both `if running, err :=
  workerSessionTargetRunningWithConfig("", nil, sp, nil, sn); err == nil && running`, on both the
  unlimited branch and the `ListRunning`-failed fallback (`:1673`).
- But it has **no production caller at `main`**. Every reference is its own definition, five calls in
  `cmd/gc/pool_test.go` (`:419 :433 :773 :785 :793`), and
  `engdocs/archive/backlogs/scaling-backlog.md:20` — "[x] **Replace `countRunningPoolInstances` with
  `ListRunning`**". So the P1 is not reproducible as filed and we did not file it.
- The live reconciler path already does what we were going to ask for: `cmd/gc/city_runtime.go:1068-1076`
  skips the whole pool death check on a list error, naming the partial and the total case separately,
  and the `IsPartialListError` guards listed in `engdocs/design/runtime-partial-discipline.md` are in
  place at `adoption_barrier.go`, `city_runtime.go`, `cmd_stop.go`, `controller.go`, `session_beads.go`,
  `doctor/checks.go`, `provider_core.go` and the tmux adapter. `cmd_start.go` carries no
  `IsPartialListError` reference at all, which reads as consistent with the call site being dead rather
  than with a gap the design missed.
- The one live sibling we did find is smaller and non-destructive, so we are not filing it either:
  `discoverPoolInstances` returns `nil` for an unlimited pool when `ListRunning` errors
  (`cmd/gc/pool.go:611-613`, and the canonical-singleton variant at `:645-647` returns just the
  canonical name), and that list feeds the on_death / drain-timeout maps at `cmd_start.go:136` and
  `:167` and the status snapshot at `city_status_snapshot.go:233` — so an outage suppresses death
  hooks and drain timeouts for pool instances rather than over-starting anything.

Two things this leaves on the table, either of which we'd take off the maintainer's list:

1. The dead counter carries five tests that assert today's behavior, so the fail-open answer is
   pinned: whoever eventually converts this path inherits `pool_test.go` reds as a surprise. Deleting
   the function with the tests, or converting it to fail closed *now* while it costs nothing, both
   look cleaner than leaving it reachable-by-test-only.
2. `engdocs/design/runtime-partial-discipline.md` §"Follow-up arms (not yet threaded)" enumerates the
   live unconverted arms (1 `state_cache` staleTTL cliff, 2 heal-to-asleep slot-free, 3 orphan
   close / drain-advance false-complete, 4 on_death storm, 5 `session_wake.go` ~552 pre-start
   fail-open, 6 `Tmux.HasSession`) and names a per-tick `runtimeQueryPartial` as the shared
   prerequisite. Is that signal what `ObserveLivenessBounded` / `ObservationStatus` are meant to
   become, or is it a separate seam? If it is the same one, a line saying so in this PR's follow-up
   section would let the arms be written against this surface instead of alongside it.

Nothing here asks for a change to this PR's behavior — `Existing call sites are unchanged` is
accurate as far as we could verify.

quad341:
Thanks — this is an unusually careful survey, and I re-ran every citation in it before answering rather than taking it on trust. Short version: your reachability conclusion holds, it holds on a wider survey than the one you ran, and the seam question has a real answer that turned out to deserve a design bead rather than a one-line reply.

## Survey re-verified

Re-checked at `main` = `3efd990c5822035419844d317135c54e997354dc`, 21 commits past the `9d59e5c1` you used. Every line number you cited is still exact:

* `countRunningPoolInstances` at `cmd_start.go:1650`; the two collapsing branches at `:1658` and `:1678`; the `ListRunning` call whose failure selects the second at `:1673`.
* No production caller. `git grep -n countRunningPoolInstances origin/main -- .` returns the definition and its doc comment, your five `pool_test.go` call sites (`:419 :433 :773 :785 :793`), and the archived backlog line. Nothing else.
* `reconcilePoolDeaths` skipping the whole check on any list error at `city_runtime.go:1068-1076`, naming the partial and total cases separately.
* The `IsPartialListError` guards present in all eight files you listed, and zero references in `cmd_start.go`.
* `discoverPoolInstances` returning `nil` at `pool.go:611-613`, and the canonical-singleton variant at `:645-647`.

Two things to add, both of which strengthen your decision not to file:

**Your `discoverPoolInstances` survey covered 3 of its 10 call sites; the other 7 agree with you.** The full set is `cmd_start.go:136 :167 :271 :340 :395`, `cmd_status.go:316 :359`, `cmd_rig.go:874`, `city_status_snapshot.go:233`, `pool.go:710`. The three in `cmd_start.go` you didn't reach are the idle-timeout, max-session-age and assigned-work-defer registration loops, and each is followed immediately by a per-template fallback (`setTimeoutForTemplate` `:283`, `setConfigForTemplate` `:347`, `setLimitForTemplate` `:401`) that still registers bead-derived runtime names. `cmd_status.go` and `cmd_rig.go` are display-only. `pool.go:710` feeds `selectRunningPoolSessionRefs`, which propagates the probe error rather than swallowing it (`pool.go:737-740`). So the effect across all ten is what you found on three: suppression of per-name lifecycle limits, never an over-start.

**There is a stronger structural reason the duplicate-start theory can't reach the live path.** Pool sizing is not runtime-occupancy-driven any more. `computePoolDesiredStatesAt` (`cmd/gc/pool_desired_state.go:285`) takes `assignedWorkBeads` and `sessionInfos` and takes **no `runtime.Provider` argument at all** — desired count is derived from bead demand, not from counting running sessions. That is why `countRunningPoolInstances` went dead in the first place, and it means a provider that cannot be asked cannot inflate desired count regardless of what the observation side returns. The runtime-partial exposure on the pool path is confined to the observation sites you already enumerated, plus the `staleTTL` cliff (arm 1 in the design doc).

## 1. The dead counter — agreed, and we have it

Filed internally as `ga-d55dlw` and routed to a builder. Please don't spend your time on it, and thank you for offering.

One correction to the framing, because it changes which of your two options is right. Converting in place is not the cheap option it looks like. `TestCountRunningPoolInstancesUsesPartialListResults` (`pool_test.go:425`) asserts `count == 2` "from per-session fallback" using `partialListPoolProvider`, which overrides only `ListRunning` and leaves `IsRunning`/`ProcessAlive` working (`pool_test.go:24-36`). For that partial-but-still-probeable case, counting what the per-session probes confirm is the *correct* answer, and the test asserts it deliberately. A blanket fail-closed conversion would contradict a purposeful test rather than fix a latent bug — and the genuinely dangerous case (list partial **and** per-session probes also failing) isn't covered by any of the five, so converting wouldn't close the gap you're pointing at either.

So: delete, with the five tests, keeping `partialListPoolProvider` alive for the `discoverPoolInstances` fail-closed test at `:400-412`. That removes the fail-open branches and their pins in one move with no production behavior change. Your "delete the function with the tests" option, for a slightly different reason than the one you gave.

## 2. The seam — separate provenance, same shape

Direct answer: **`ObserveLivenessBounded` was not designed to be the `runtimeQueryPartial` signal.** It descends from #5742 via a doctor-layer architecture whose functional requirements are entirely about `AgentSessionsCheck` / `ZombieSessionsCheck` outcomes and `gc doctor --json` rendering. That architecture mentions the reconciler, `storeQueryPartial`, `runtimeQueryPartial` and the follow-up arms exactly zero times. Nobody wrote this primitive with those arms in mind.

Your instinct that they look like the same thing is nevertheless right, and that's the interesting half. The design doc's stated prerequisite is "optional provider interface via type-assert, like `LivenessObserver`, plus a `Liveness.RuntimePartial` field". This PR ships an optional provider interface via type-assert — `BoundedLivenessObserver`, asserted exactly the way `LivenessObserver` is — plus the partial bit, carried as a third return value instead of a field on `Liveness`. The mechanism the doc asked for exists; only the carrier differs.

Whether the arms should therefore be written against this surface is a genuine design call, and not one I'm going to make in a PR comment, so it is filed as `ga-kwtytv` for an architect. The deliverable is exactly the one you proposed: if the answer is "same seam", the doc's closing paragraph gets amended to name this surface, so arm authors build against it rather than alongside it.

One constraint that decision has to weigh, worth stating here because you would hit it first: `ObserveLivenessBounded` spawns a goroutine and a `context.WithTimeout` per call. That cost was accepted for doctor's per-target probe loop, which is bounded and human-invoked. The arms want a per-tick signal across every session on every tick — a materially different call volume. Same shape does not automatically mean the same surface is right at that frequency.

One detail worth correcting: "nothing in-tree consumes it yet" is true at `main`, but a consumer is in flight internally — the `AgentSessionsCheck`/`ZombieSessionsCheck` wiring this primitive was built for. It is currently in request-changes over a missing test and is not yet a PR. So the surface does get exercised shortly; just not by the arms.

## This PR

No change. `Existing call sites are unchanged` is accurate, independently confirmed, and nothing in your review is blocking here.

quad341:
@quad341 — thank you for this, and specifically for the doc comment on `ObserveLivenessBounded`. Spelling out that a deadline yields *zero* liveness while `ErrRuntimeUnavailable` *preserves* the provider's partial answer is the distinction that makes this primitive usable, and it's normally the thing someone has to reverse-engineer from tests two quarters later. Your five tests map one-to-one onto the branches, and disclosing the goroutine-lifetime trade-off in the body instead of hoping nobody noticed made this fast to review.

One note in your favor: `LivenessObserverWithError` landed on `main` while this was in review, so `BoundedLivenessObserver` is now a duplicate of it through no fault of yours. I'll collapse it to an alias — your name stays — and route the goroutine through the existing helper, which also picks up the `normalizeLiveness` call the new branch skips. Small parity edits, no change to your semantics, and none of your tests move.

I'll rebuild, re-run `internal/runtime` plus the `cmd/gc` liveness tests, and merge. The per-tick cost question you raised is tracked separately as `ga-kwtytv` — good catch, and not this PR's problem.


---

PR #6092, opened 2026-09-06
fix(api): key the orders/check response cache on time, not the event sequence

remuscazacu:
## The problem

`/orders/check` has had a response cache since it was written, and its own pollers could never hit it. The entry was stored under the event sequence and looked up under exact equality:

```go
index := s.latestIndex()
if body, ok := cachedResponseAs[OrderCheckListBody](s, cacheKey, index); ok {
```

Any event anywhere in the city advances that index, including the events the polling itself emits. So the lookup missed on every request that was not a second request inside the same quiet moment — which, on a city doing work, does not happen.

That would be merely wasteful if the body were cheap. It is not: building it evaluates every order's due-ness on demand and costs **one labeled bead List per order per store**.

## Measured

Against the live supervisor at `127.0.0.1:8372`, two production cities:

| endpoint | time |
|---|---|
| `runs/census` | 0.001s |
| `orders` | 0.004s |
| `status` | 1.0s |
| **`orders/check`** | **5.8–11.1s** (23 orders) / **14.2–22.4s** (26 orders) |

The event sequence advanced roughly **six times a minute** on both cities. Every poll paid the full rebuild.

## Why it surfaced as something else entirely

The SBF observability exporter reads this endpoint once a minute with a 5s per-call timeout, so it timed out on **every** scrape in both cities. Its collector marks a city down when any single endpoint fails — so `city_up` sat at 0 while `status`, `orders` and `runs/census` all answered normally. The dashboard read *two cities unreachable* on a page where every other panel showed live data, and the whole order-health metric family was simply absent, because it is sourced from this endpoint alone.

The endpoint being slow is visible. The endpoint being slow *and* presenting as a reachability failure is not.

## The change

Adopt the `/status` recipe (`ra-4u2eqc`, #5028) wholesale — this is the same failure in the same shape, one handler over:

1. **Key on `responseCacheTimeBucket`** instead of the event index, so the entry survives sequence churn.
2. **TTL floor** (`orderCheckResponseTTLFloor`, 30s) for the request that lands just past a bucket boundary.
3. **Stale-while-revalidate** — serve the last-known-good body and refresh behind it.

The third is the load-bearing one, and it is worth being explicit about why the first two are not enough on their own. A fixed-interval poller reading once a minute misses **both** fast lookups every time: no wall-clock window short enough to be useful to an interactive caller is long enough to survive a 60s gap. Without the stale path, a 1/min reader pays the cold build on every single read no matter how the bucket is tuned. With it, no caller waits on the rebuild at all.

`refreshOrderCheckResponseInBackground` mirrors `refreshStatusResponseInBackground`: coalesced through `beginResponseRefresh` so concurrent misses share one rebuild, `defer s.endResponseRefresh` on every exit path, and a `recover()` because `withRecovery` does not cover detached goroutines.

## Unchanged

- `?fresh=true` bypasses all three lookups.
- The response shape. `/status` widens `X-GC-Cache-Age-S` to report response-entry age when it serves stale (`huma_types.go`); `OrderCheckListOutput` carries no such header — it is a bare `Body` struct, not a `ListOutput`/`IndexOutput` — so there is nothing here to widen, and adding one would be an OpenAPI change rather than a fix. That is the one part of the `/status` recipe deliberately not adopted.
- A city holding a **condition-triggered** order is not cached at all, exactly as before — its due-ness comes from running a subprocess, and a cached answer would report a check that did not happen.

## Tests

Three, in `handler_order_check_cache_test.go`. Each asserts on **work done** (`countingStore.listByLabelCalls`), not on the returned body, since a test that only compared responses would pass against the old index-keyed cache too.

- `TestHandleOrderCheckCachesAcrossIndexChanges` — five event-sequence advances between requests, and the fan-out must not repeat. This is the bug.
- `TestHandleOrderCheckServesStaleAndRefreshesInBackground` — both fast lookups pinned off, so every request after the first reaches the stale path; asserts the served body is byte-identical to the primed one, that the background refresh did rebuild, and that `responseRefreshing` is empty afterwards (a leaked guard wedges every future refresh for that key).
- `TestHandleOrderCheckFreshBypassesCache` — the escape hatch stays open.

`go test ./internal/api/` passes (135s, full package); `-race` on the new tests clean; pinned golangci-lint v2.12.0 reports 0 issues across `./internal/api/...`.

## Adjacent, not fixed here

The per-order evaluation is still serial and still O(orders × stores) in bead reads — this change makes at most one caller per bucket pay it instead of every caller, but does not make it cheaper. Evaluating orders concurrently, or splitting the expensive reason string behind a query parameter, are separate changes.

https://claude.ai/code/session_01F7UDQDVxrVJiFm4N6hrKUQ

### DIFF
--- a/internal/api/handler_order_check_cache_test.go
+++ b/internal/api/handler_order_check_cache_test.go
@@ -1,0 +1,197 @@
+package api
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+	"time"
+
+	"github.com/gastownhall/gascity/internal/beads"
+	"github.com/gastownhall/gascity/internal/events"
+	"github.com/gastownhall/gascity/internal/orders"
+)
+
+// orderCheckTestOrders returns two cooldown orders — enough to make the
+// per-order fan-out visible in the store's call count, and deliberately free
+// of any condition trigger, which would disable the response cache outright.
+//
+// Each names a rig, because that is what gives it a store to read history
+// from: orderStoreInfosForState resolves a rig-scoped store first, and an
+// order naming neither a rig nor a city store yields no store info at all, so
+// the history read is skipped and the fan-out this file measures never
+// happens.
+func orderCheckTestOrders() []orders.Order {
+	enabled := true
+	return []orders.Order{
+		{Name: "dolt-health", Exec: "dolt status", Trigger: "cooldown", Interval: "5m", Rig: "myrig", Enabled: &enabled},
+		{Name: "mail-sweep", Exec: "gc mail sweep", Trigger: "cooldown", Interval: "5m", Rig: "myrig", Enabled: &enabled},
+	}
+}
+
+// TestHandleOrderCheckCachesAcrossIndexChanges pins the fix:
+// /orders/check keys its response cache on a wall-clock time bucket, not on
+// the event sequence.
+//
+// Keyed on the sequence, the cache was unreachable for the callers that need
+// it most. Building this body costs one labeled bead List per order per
+// store, and the endpoint's readers are pollers: the SBF observability
+// exporter reads it once a minute. On the two production cities the sequence
+// advanced about six times a minute, so an entry stored under one index was
+// never looked up under that index again, and every poll paid the full
+// rebuild — 5.8-11.1s on a 23-order city, 14.2-22.4s on a 26-order one,
+// against the exporter's 5s per-call timeout. Its collector marks a city down
+// when any single endpoint fails, so both cities reported unreachable while
+// every other endpoint answered normally.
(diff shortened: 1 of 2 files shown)


remuscazacu:
Bump — p1, opened 2026-09-06, no human review in 18 days, still `MERGEABLE`.

The reason this one is worth a triager's attention despite being small: the symptom is not "the endpoint is slow", it is "healthy cities are reported unreachable". A cache keyed on the event sequence never hits, `orders/check` takes 6–22 s, and every scrape times out — so the failure surfaces as a monitoring lie rather than as a latency graph, which is the kind of bug that gets diagnosed twice before anyone looks at the cache key.

Happy to rebase if it has drifted.

PR #6100, opened 2026-09-06
fix(beads): honor blocked work outcomes in dependency readiness

quad341:
## What changes

Closed dependencies with `gc.work_outcome=blocked` no longer unlock dependent work. Legacy closed dependencies without a work outcome, and closed dependencies with any non-blocked outcome, remain satisfied.

The canonical truth table is shared by the in-memory and caching stores and applied at the provider edge for SQLite, native Dolt, exec, and `bd`-backed readiness. BdStore continues to use native `bd ready` candidates and adds only the closed-blocked veto.

## Test plan

- Full 40-job local suite completed; 34 jobs passed and six non-diff-owned raw failures were attributed to existing condition trackers.
- All new truth-table and readiness-conformance cases passed across every registered store implementation.
- Native DoltLite, boundary/policy checks, build, vet, changed-file formatting, and docs synchronization passed.
- The pre-push fast gate passed all 10 jobs.

Full evidence and failure attribution: `release-gates/ga-0rqcer-dependency-work-outcome-gate.md`.


quad341:
@quad341 — this is a well-argued change. Routing everything through one `DependencySatisfied` predicate instead of patching each store's readiness check separately is the right call, and the truth table pins the two cases that actually matter: a closed blocker with no `gc.work_outcome` still satisfies (the backward-compat guarantee), and an unrecognized future value fails open rather than stalling dependents nobody understands yet. Teaching the exec conformance fixture real `dep-add`/`dep-list` state so ExecStore genuinely participates instead of passing vacuously is the kind of thing that usually gets skipped.

One real catch on the BdStore side: `filterReadyByWorkOutcome` fetches blockers with `List(ListQuery{IDs: …, TierMode: TierBoth})`, and without a closed-inclusive query those rows get dropped twice — no `--all` server-side, and `Matches` filters `status=="closed"` client-side. So `statusByID[blockerID]` is always `""` and the veto can't fire on the bd path. It slipped through because no fixture sets `dependency_count`, so the filter early-returns before reaching that call in every test.

I'll take it from here: narrow that lookup to closed blockers, add the witnessed-path regression test, and clear out the leftover `bd dep list` fixtures. Your commits and design stay as-is. Thanks for this one.


---

quad341:
@quad341 — routing every store through one `DependencySatisfied` predicate instead of patching each readiness check independently is the right call, and the truth table pins the two cases that decide whether this is safe: a closed blocker with no `gc.work_outcome` still satisfies, and an unrecognized future value fails open rather than stalling dependents nobody understands yet. Teaching the exec fixture real `dep-add`/`dep-list` state so ExecStore participates non-vacuously is the part that quietly gets skipped.

Two things I'll take on my side, both maintainer-side integration work:

The BdStore blocker lookup isn't closed-inclusive, so `statusByID` never sees `"closed"` and the veto can't fire there. I'll narrow that lookup, add the witnessed-inline-dependency test that would have caught it, and clear the leftover `bd dep list` fixtures.

The native Dolt filter re-blocks on any non-closed dependency where your BdStore version deliberately vetoes only closed-and-blocked. I read the pinned beadslib: `waits-for` gates on the spawner's children rather than its status, and bd treats `pinned` as satisfying. So the full predicate over-blocks. Your doc comment already says this should mirror BdStore — I'll make the code match it.

Your commits and design stand. Thanks for this one.


---

quad341:
@quad341 — routing every store through one `DependencySatisfied` predicate, instead of patching six readiness checks independently, is the right shape, and the truth table pins the two cases that decide whether this is safe: a closed blocker with no `gc.work_outcome` still satisfies, and an unrecognized future value fails open rather than stalling dependents nobody understands yet. Teaching the exec fixture real `dep-add`/`dep-list` state so ExecStore participates non-vacuously is the part that usually gets skipped.

Two integration issues I'll take on my side — both wiring, not design:

The BdStore blocker lookup isn't closed-inclusive, so `statusByID` never sees `"closed"` and the veto can't fire there. I'll scope that query to closed rows, add the witnessed-inline-dependency test that would have caught it, and clear the leftover `bd dep list` fixtures.

The native Dolt filter uses the full predicate where your BdStore version deliberately vetoes only closed-and-blocked. I checked the pinned beadslib: `pinned` satisfies a blocks edge, and `waits-for` gates on the target's children — so the full predicate re-blocks work bd already cleared. Your own doc comment there says it should mirror BdStore; I'll make the code match, with a native gate-tolerance test.

Your commits and credit stand. The shared-predicate shape is what made both easy to find.


---

quad341:
@quad341 — routing every store through one `DependencySatisfied` predicate, instead of patching six readiness checks independently, is the right shape, and the truth table pins the two cases that decide whether this is safe: a closed blocker with no `gc.work_outcome` still satisfies, and an unrecognized future value fails open rather than stalling dependents nobody understands yet. Teaching the exec fixture real `dep-add`/`dep-list` state so ExecStore participates non-vacuously is the part that usually gets skipped.

Two integration issues I'll take on my side — both wiring, not design:

The BdStore blocker lookup isn't closed-inclusive, so `statusByID` never sees `"closed"` and the veto can't fire there. I'll scope that query to closed rows, add the witnessed-inline-dependency test that would have caught it, and clear the leftover `bd dep list` fixtures.

The native Dolt filter uses the full predicate where your BdStore version deliberately vetoes only closed-and-blocked. I checked the pinned beadslib: `pinned` satisfies a blocks edge, and `waits-for` gates on the target's children — so the full predicate re-blocks work bd already cleared. Your own doc comment there says it should mirror BdStore; I'll make the code match, with a native gate-tolerance test.

Your commits and credit stand. The shared-predicate shape is what made both easy to find.


---

PR #6113, opened 2026-09-07
fix(test): export gitconfig to parallel runner children

### LINKED ISSUE #6112: Parallel test runner aborts before tests: gc_test_gitconfig is not exported
At upstream/main `6ba434271c5f76bc150f694ebeef9d713be16b3a`, `make test-fast-parallel` aborts all ten jobs before they execute tests with `gc_test_gitconfig: unbound variable`.

`scripts/test-local-parallel` resolves `gc_test_gitconfig` at line 29 but does not export it. Its `xargs ... bash -c` children expand that variable under `set -u` while building the nested `env -i` environment.

Expected: children receive the resolved seeded Git config and run their selected jobs without caller-side setup.

Diagnostic workaround: `make test-fast-parallel EXTRA_TEST_ENV="gc_test_gitconfig=$(scripts/test-gitconfig-path)"`. With that workaround the runner reaches actual tests; separate macOS/mise and backup-script failures remain on pristine upstream and are outside this issue.

A focused regression should exercise the actual child environment without running the broad suite.

Toady00:
## Summary

Fixes #6112.

Export the seeded `gc_test_gitconfig` before the parallel runner launches `xargs ... bash -c` children. Without the export, every child aborts under `set -u` while preparing its `env -i`, before running its workload.

Extend the existing shell self-test with a probe through the actual initialization and `run_fan_out` function. It clears any inherited workaround, checks the child can read the seeded Git config, and avoids launching Go suites or recursively invoking the self-test. No new Go subprocess census entries or production test seams.

## Validation

- RED: `bash scripts/test-local-concurrency.sh` reported 24 passed, 2 failed, with `gc_test_gitconfig: unbound variable` from the real child.
- GREEN: the same self-test reported 26 passed, 0 failed. The existing strace-only check skipped on macOS.
- `bash -n scripts/test-local-parallel scripts/test-local-concurrency.sh` passed.
- Both focused Git-environment tests in `./scripts` passed with `-count=1`.
- `make vet` passed using resolved Go 1.26.6 with inherited GOROOT unset.
- Normal pre-commit hook passed.
- Normal push invoked `make test-fast-parallel` without the gitconfig workaround. All ten jobs reached workloads. Darwin compile, both shell self-tests, and cmd/gc shard 5 passed; the scripts Go package also passed within unit-core.

The full suite FAILED. It reproduced three mise-trust beads-init failures, three python3 mise-shim failures, and two backup-script tests missing `timeout.log`. Those were also observed on pristine upstream. This run also reported `TestStopManagedCityForcesCleanupAfterTimeout`, `TestStopManagedCityDoesNotUseStartupOrDriftTimeouts`, and `TestReleaseOrphanedPoolAssignments_DetachedProbeAliveSkipsRelease`. Cmd/gc shard 2 ended with FAIL without a named failing test in its log; its cause is unclassified. None is changed or suppressed here.

After the normal pre-push gate failed, the branch was pushed with `--no-verify`. This is not a green full-suite result.

### DIFF
--- a/scripts/test-local-concurrency.sh
+++ b/scripts/test-local-concurrency.sh
@@ -177,6 +177,28 @@ else
     record_fail "wiring.full_case_calls_selftest" "could not locate the full) case block in $LOCAL_PARALLEL"
 fi
 
+# Exercise the real initialization and fan-out without launching Go suites,
+# acquiring a gate slot, or recursively running this self-test.
+GITCONFIG_OUT="$(
+    unset gc_test_gitconfig
+    repo_root="$(dirname "$TEST_DIR")"
+    eval "$(sed -n '/^gc_test_gitconfig=/,/^# shellcheck source=lib\/test-slice.sh/{ /^# shellcheck/!p; }' "$LOCAL_PARALLEL")"
+    eval "$(sed -n '/^run_fan_out() {$/,/^}$/p' "$LOCAL_PARALLEL")"
+    log_dir="$(mktemp -d "${TMPDIR:-/var/tmp}/gc-gitconfig-test.XXXXXX")" || exit 1
+    trap 'rm -rf "$log_dir"' EXIT
+    export LOCAL_TEST_LOG_DIR="$log_dir"
+    export TEST_LOCAL_GOPATH="${GOPATH:-}" TEST_LOCAL_GOCACHE="${GOCACHE:-}"
+    export TEST_LOCAL_GOMODCACHE="${GOMODCACHE:-}" TEST_LOCAL_GOTMPDIR="${GOTMPDIR:-}"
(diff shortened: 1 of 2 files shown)


Toady00:
Push-gate exception for this runner-only branch: issue https://github.com/gastownhall/gascity/issues/6112 is fixed by this PR, https://github.com/gastownhall/gascity/pull/6113. A normal push ran the full fast runner and failed on the separate suite failures listed in the PR description, not the missing export. After that failure I used the explicitly authorized `git push --no-verify`. The focused regression, syntax checks, Git-environment tests, make vet, and normal pre-commit passed. The full suite did not pass. This branch contains no ACP changes.

Toady00:
Closing as superseded by #6116, which landed the missing export in c5b52961d. This PR also included a behavioral regression test through the real fan-out path; that additional test was not included in #6116. The runtime bug itself is fixed upstream.

Issue #6104, opened 2026-09-07
bug: pending-create recovery races the in-flight async start after the awake heal, leaving a live session permanently reset-pending


atbrace:
## Summary

A session whose async `provider.Start` runs close to `startup_timeout` can end up **alive, working normally, and permanently `reset-pending`**: `continuation_reset_pending=true`, `session_key=""`, `started_config_hash=""`, `last_woke_at=""`, `wake_attempts` accrued — on a runtime the reconciler itself had already committed as started seconds earlier. Nothing ever clears the flag while the runtime lives, so `gc session list` shows `reset-pending` for the session's whole life, `compute_awake_set` pins `desired=reset-pending` (exempt from idle sleep), config-drift detection is blind for it (empty hash), and the next real restart cannot resume (empty key).

Observed on a 12-slot pool of pi-based agents under load (macOS, tmux provider, `startup_timeout` 60s). Two instances with byte-identical metadata timelines, reproduced from `events.jsonl`:

```
gc-apvfrf (local time)                       gc-h47iv9
13:17:07 creating  pcc=true lwa=T0           13:17:53 creating  pcc=true lwa=T0
13:17:25 awake     pcc=true   (heal)         13:18:06 awake     pcc=true   (heal)
13:17:48 active/creation_complete            13:18:52 active/creation_complete
         sch=v5:354 pcc=""  (recovery)                sch=v5:0ce pcc=""  (recovery)
13:18:07 lwa=""            (T0+60s)          13:18:53 lwa=""            (T0+60s)
13:18:08 crp=true sch="" sk="" wa=1          13:18:53 crp=true sch="" sk=""
13:18:11 awake, flag never cleared           13:18:54 wa=1 ; 13:20:05 awake, never cleared
```
supervisor log for both: `session lifecycle: op=start … outcome=deadline_exceeded duration=1m0.0xs`. No `rollback pending-create` line, bead not closed — the rollback path did not run.

## Mechanism (main @ 6ba434271; same code at every site)

1. `PreWakePatch` writes `state=creating`, `pending_create_claim=true`, `last_woke_at=T0`; the async start begins under `startup_timeout`.
2. Next tick: tmux is up, so `projectRuntimeProjection` (`internal/session/lifecycle_projection.go`, `if input.Runtime.Alive { return RuntimeProjectionAlive, StateAwake, false }`) projects **awake** and `healStatePatchWithRollbackInfo` writes `state=awake` — claim still set, start still in flight.
[...]

Issue #5895, opened 2026-09-02
bug: session lifecycle: failed-create rollback skips runtime stop, wedging provisioned-but-unlaunched instances


jacobhausler:
## Summary

`commitStartFailure` rolls back a pending-create session bead on two error
paths without ever stopping the runtime it may have already provisioned, so a
provider that provisions a child before its `launched` marker is set (e.g. an
out-of-tree RPP runtime backed by a scheduler like Nomad, where "launched"
means "the in-container start command ran") can leave an orphaned, unowned
runtime instance behind every time a start fails on that path. A retry then
hits the provider's own already-exists guard and wedges.

## Symptom

- A pending-create session's async start fails after the provider has already
  provisioned a child (binding written, container/job scheduled) but before
  the provider's liveness-visible "launched" marker is set.
- `commitStartFailure` (`cmd/gc/session_lifecycle_parallel.go:2265`) takes the
  `result.rollbackPending` branch and calls `rollbackPendingCreate`, which
  closes the bead `state=failed-create` — but no `Stop` call is made on the
  provider for that instance.
- `ListRunning`/`IsRunning` for such a provider only report launched
  instances, so the reconciler's own view shows nothing running for that name.
- The next provision attempt for the same slot hits the provider's
  already-exists guard (binding still present) and fails; the slot is wedged
  until a human manually calls the provider's idempotent stop out of band.

## Reproduction

1. Use (or fake, in a test) a `runtime.Provider` that provisions a child on
   `Start` — writing a binding/instance record — but only becomes visible in
   `ListRunning`/`IsRunning` after a separate "launched" signal.
2. Trigger a start failure for a pending-create session after `Start` has
   provisioned the child but before the launch signal fires (e.g. return an
   error from the phase that runs immediately after provisioning).
3. Observe: `commitStartFailure` runs the `result.rollbackPending` branch and
   calls `rollbackPendingCreate` with no preceding call to any provider stop
   helper.
4. Retry the start for the same session name/slot: the provider's
   already-exists guard on `Start` rejects it because the binding from step 1
   was never torn down.

Expected: the failed-create rollback stops any runtime instance it may have
provisioned, exactly once, so the slot is clean for the next attempt. Actual:
the instance is left behind and the slot is wedged.

## Root cause

`startResult` (`cmd/gc/session_lifecycle_parallel.go:236-247`) does not retain
the `runtime.Provider` handle (`sp`) that `runPreparedStartCandidate` used to
[...]

Issue #5436, opened 2026-08-20
opencode/ACP sessions never route — reconciler falsely logs start success while no runtime backend ever starts


elviskahoro:
## Summary

On a fresh, standard config (`[workspace] provider = "opencode"`, `[providers.opencode] base = "builtin:opencode"`), any agent resolving to the `opencode` provider is permanently unable to start — but the failure mode is confusing and produces misleading state.

## Environment

- `gc version`: 1.4.0 (also reproduced after `brew upgrade gastownhall/gascity/gascity` info showed 1.4.1 available; did not fully upgrade the live install, but commit history for 1.4.0→1.4.1 shows no related fix)
- OS: macOS (Darwin 25.5.0, arm64)
- `opencode` binary: 1.18.4 (Homebrew), supports `opencode acp` subcommand
- tmux: 3.7b, working correctly for tmux-transport (claude/codex) sessions in the same city

## Repro

1. `gc init --provider claude` (or any city with a `[[rigs]]` entry)
2. Add to `city.toml`:
   ```toml
   [providers.opencode]
   base = "builtin:opencode"
   ```
3. Set `provider = "opencode"` on the workspace or an `[[agent]]`/rig-scoped agent.
4. `gc session new <opencode-agent> --no-attach`

## Observed behavior

The session sits in `start-pending` indefinitely. `gc doctor`/`gc config show` report nothing wrong. The supervisor log repeats, every reconcile tick:

```
buildDesiredState: bead <id> template "<name>": agent "<agent>": provider "opencode" requires ACP transport but the session provider cannot route ACP sessions (skipping)
```

Sometimes (observed on a named/pool session, not consistently reproducible) the reconciler instead logs:

```
session lifecycle: op=start wave=0 session=<name> template=<template> outcome=success duration=X.XXXs phases=[start_call=X.XXXs]
```

...and `gc session list` shows the session as `active` — but no tmux window is ever created for it, and the very next tick logs:

```
session reconciler: confirming dead runtime session <name>: tmux -u: can't find window: <name>
```

This "false success" is worse than a clean failure: in our case, the control-dispatcher's pool-routing logic saw the phantom "active" session and auto-assigned real backlog work beads (a `mol-polecat-base` molecule, 6 beads) to it, which then sat `in_progress`/orphaned against a session that never existed at the runtime layer.

## What we ruled out

[...]

PR #6114, opened 2026-09-07
fix(session): fence ACP startup ownership and pending-create rollback

### LINKED ISSUE #6109: ACP startup exposes liveness without ownership; stale rollback can close completed creation
## Confirmed defects

Source inspection and deterministic reproductions against main at `6ba434271c5f76bc150f694ebeef9d713be16b3a` confirm two interacting startup defects:

- ACP `Provider.Start` publishes an alive in-memory sentinel before the handshake, but does not seed the ownership sidecars used by `GetMeta`. Passing `GC_SESSION_ID`, `GC_INSTANCE_TOKEN`, and `GC_RUNTIME_EPOCH` to the child environment does not populate those sidecars. A reconciler can therefore see a live runtime without the identity needed to recognize its pending creation.
- Pending-create rollback checks only whether the bead is already closed before writing its rollback transaction. It does not atomically revalidate the pending claim and incarnation against asynchronous start completion. A stale creating snapshot can close a creation that has already completed, clear its explicit runtime name, and leave the runtime occupying that name.

For a canonical singleton, continuing demand can then allocate and retire more session records while that runtime remains alive. This report does not authorize adoption or termination of a foreign runtime.

## Observed incident

The inspected incident contained 77 session records. One correlated record showed creation completion at relative `t=0`, followed by a failed-create close at `t=+354ms`. Local paths, session identities, and tokens are intentionally omitted.

The exact provenance of the installed binary that produced the incident is unknown. The source defects and isolated reproductions above are confirmed on current source at `6ba434271`; that does not establish which precise commit produced the installed binary. No live incident replay has been performed.

## Reproduction boundaries

1. Gate an ACP handshake after the startup sentinel and control socket become observable. Query `IsRunning` and `GetMeta` through both the owning provider and a second provider. Before the fix, liveness is observable without the supplied ownership sidecars.
2. Capture a pending-create snapshot, commit successful creation, then invoke rollback with the captured snapshot. Before the fix, rollback closes the completed record.
3. Leave a runtime occupying the canonical singleton name without an open holder bead, then repeat fresh-create demand. Before the fix, new records can be minted despite the occupied name.

A companion patch adds these regressions, serializes ACP startup/cleanup by runtime name, seeds only classified metadata before liveness, and fences controller completion/rollback under the existing per-session mutation lock. The controller fence is process-local, not a cross-process store CAS. It also defers fresh singleton allocation while the derived runtime name is occupied.

## Related reports

- #2112 has the closest reported symptoms; this report supplies a narrower confirmed mechanism.
- #5436 concerns different routing behavior.
- #4714 and #5895 concern orphan cleanup.
- #6104 and PR #6105 concern a different recovery race involving advisory awake state during creation. The proposed rollback fence preserves rollback eligibility for a still-claimed awake creation whose Start fails.

These references are context, not claims that this patch fixes all of those reports.

## Verification

Final isolated checks pass with Go 1.26.6 on macOS arm64: full ACP/session package race tests, targeted controller lifecycle race tests repeated 10 times, ACP startup/cleanup race regressions repeated 10 times, and a CLI build. Earlier `make vet` passed.

The full local baseline is not green: plain `make test-fast-parallel` aborts before tests with `gc_test_gitconfig: unbound variable` on pristine main too. An environment-only workaround reaches the suites; two backup tests still fail with missing `timeout.log` on pristine main as well. Mise shim/trust failures disappear without shims, and four load-sensitive timing failures pass focused reruns. These baseline limitations are not evidence of a complete end-to-end fix.

Toady00:
## Summary

Fixes #6109.

ACP Start exposed a live startup sentinel before seeding the ownership sidecars read by GetMeta. Supplying identity through the child environment alone did not populate those files. Separately, pending-create rollback could close a successfully completed creation using a stale snapshot because its read and rollback transaction were not serialized with asynchronous completion.

- Seed only classified metadata before publishing the sentinel or control socket. Do not persist credentials from the child environment.
- Serialize ACP startup and cleanup by runtime name, retain ownership until failed startup cleanup drains, and preserve replacement runtimes and their sidecars.
- Interrupt backpressured handshake writes on cancellation or deadline. Disarm and join the cancellation callback before transferring the connection, and release the lifecycle lock before initial nudge delivery so Stop can interrupt it.
- Re-read the lease and commit completion or rollback under the existing process-local session mutation lock. Reject stale completion without emitting success, and clean up only the attempted runtime identity outside that lock.
- Defer fresh canonical singleton record allocation while its derived runtime name is occupied, rather than repeatedly minting failed records or adopting a foreign runtime.

The controller fence is process-local, not a cross-process store CAS. Direct external bead writes remain outside its guarantee. A still-claimed advisory awake state remains rollback-eligible when Start fails.

## Evidence and scope

The source defects and deterministic reproductions were confirmed against `6ba434271c5f76bc150f694ebeef9d713be16b3a`. The inspected incident contained 77 session records; one correlated record completed creation at relative t=0 and was closed as failed-create at t=+354ms. Local paths and identities are redacted.

The exact installed binary provenance for that incident is unknown. Current-source reproduction does not establish which commit produced the installed binary, and no live replay has been performed.

#2112 has the closest reported symptoms. #5436 concerns different routing behavior; #4714 and #5895 concern orphan cleanup; #6104 and PR #6105 concern a different recovery race. This PR does not claim to fix all of those reports.

## Verification

Completed on macOS arm64 with Go 1.26.6, resolved toolchain PATH, inherited GOROOT unset, and ICU build flags:

- `go test -race ./internal/runtime/acp ./internal/session -count=1` passed.
- `go test -race ./internal/runtime/acp -count=10` passed after the final lint fixes.
- `go test -race ./cmd/gc -run "Test(RollbackPendingCreate|StartCompletionRejects|FailedStartRollsBack|AsyncStartFinalCommit|SingletonPendingCreate|CommitStartResult|ReconcileSessionBeads_.*PendingCreate)" -count=10` passed.
- CLI build passed with output outside the checkout; nothing was installed or restarted.
- Normal pre-commit passed changed-file lint, code generation, `make vet`, and docs checks. The commit has a verified GPG signature.

## Baseline limitations and push disclosure

The normal pre-push hook failed before tests with `gc_test_gitconfig: unbound variable`, also reproduced on pristine upstream. With explicit authorization, this branch was pushed using `--no-verify` after the checks above. The separate runner issue is #6112 and its fix is PR #6113; no runner commits are included here.

The earlier broad run with `EXTRA_TEST_ENV="gc_test_gitconfig=$(scripts/test-gitconfig-path)"` reached the suites and passed ACP/session coverage, but the full baseline was not green. Mise trust and python3 shim failures disappeared without shims. `TestBackupScriptEscalatesOffsiteFailureWithConfiguredBound` and `TestBackupScriptRejectsUnusableOffsiteTimeout` failed with missing timeout.log on both this branch and pristine upstream. Four load-sensitive timing failures passed focused reruns; two also failed in the pristine broad run. These results are disclosed rather than presented as a passing full suite.

This PR contains only signed ACP/lifecycle commit `9ffb950cbe64ca49c0cfc2b473f41b05f66d7d4d`.


Toady00:
Explicit pre-push bypass disclosure: I first attempted a normal push. Its pre-push hook aborted all ten test jobs before execution with `gc_test_gitconfig: unbound variable`. The same runner failure was reproduced on pristine upstream and is tracked separately in https://github.com/gastownhall/gascity/issues/6112, with the fix in https://github.com/gastownhall/gascity/pull/6113.

After explicit authorization, I pushed this ACP branch with `git push origin fix/acp-startup-ownership --no-verify`. This bypassed the pre-push hook, not all verification. Normal pre-commit ran successfully, including changed-file lint, code generation, make vet, and docs checks. The CLI build passed, full ACP/session race tests passed, full ACP race tests passed 10 repetitions after the final lint fixes, and the targeted controller lifecycle race tests passed 10 repetitions.

The earlier broad workaround run was not fully green: mise trust/python3 shim failures disappeared without shims; two backup timeout.log failures also reproduced on pristine upstream; four load-sensitive timing failures passed focused reruns, and two of those also failed in the pristine broad run. No live incident replay was performed. I am not treating these results as a passing full suite.

The PR contains only signed commit `9ffb950cbe64ca49c0cfc2b473f41b05f66d7d4d`. The independent runner fix is not included in this branch.

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg--9223372036854775096

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @Toady00! This PR fences ACP session **startup ownership** and makes **pending-create rollback** race-safe: a stale `Stop` for a rebound session name no longer deletes the live replacement's identity sidecars; pending-create completion and rollback now re-read and decide under one `WithSessionMutationLock` instead of racing a read-then-write; an advisory awake state during creation no longer blocks rollback (#6104); a stale final async commit cleans up by full incarnation identity rather than by attempted name alone; an occupied singleton runtime name defers fresh bead creation instead of churning a bead every demand tick; and a blocked handshake write became cancellable without holding up `Stop`. That matters because these are exactly the windows where a session gets double-owned or stranded during startup — the reconciler was able to lose a live session's metadata, report a running runtime as dead, or mint bead churn, and each of those is the kind of failure that is painful to diagnose after the fact. The change also lands the reasoning durably, in `engdocs/design/session-store-fences.md`, `internal/session/REQUIREMENTS.md`, and two `AGENTS.md` invariants, so the fences are documented rather than folklore.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **A live replacement runtime could be reported as not running.** The new replacement-refusal branch in `acp.Provider.Stop` returned before evicting its own dead `sessionConn`, so `IsRunning` short-circuited on the stale tracked conn (`if ok { return sc.alive() }`) and never fell through to the socket probe — the exact orphan/flap signal this PR exists to prevent, and a regression against the previous behaviour, which deleted the conn and did fall back. The final patch deletes `p.conns[name]` and `p.workDirs[name]` before returning, while still skipping `cleanupMeta` and process teardown, so the replacement keeps its sidecars and liveness resolves through `socketAlive`.
- **`Stop` lost its documented idempotence.** The same branch returned a non-nil, non-`IsSessionGone` error on a healthy steady state, contradicting the doc comment two lines above it and making every caller — all of which branch only on `runtime.IsSessionGone` — log and continue once per reconciler tick indefinitely. It now returns `nil`, and both the `Stop` doc comment and the ACP `AGENTS.md` invariant state the rebound-name contract explicitly.
- **The one new branch carrying both of the above had no test coverage**, and the existing test actually pinned the *pre-fix* contract by asserting `Stop` errored. `internal/runtime/acp/startup_test.go` now pins all three post-conditions — `Stop` returns `nil`, both maps are evicted, and `IsRunning` is true on both providers with the replacement's `GC_SESSION_ID` intact — with a real `WorkDir` set so the `workDirs` assertion is not vacuous. Reverting only the behavioural half of the fix fails all three assertions.
- **A silently fenced heal looked identical to one never attempted.** `recoverRunningPendingCreate`'s `!applied` early-out recorded no trace decision; it now records a skipped decision under a new `pending_create_superseded` reason, matching the sibling not-applied rollback site this PR added in `session_reconciler.go`.

Review findings addressed:
- The `p.conns`/`p.workDirs` eviction and the `nil` return, plus the rewritten `TestStopStaleConnectionPreservesNewIncarnation` and the `AGENTS.md` contract line — maintainer commit `1cc62fb160`.
- The `pending_create_superseded` trace decision on the `!applied` early-out — `1cc62fb160`.
- Two documentation gaps both reviewers flagged as unremarked in the code, also in `1cc62fb160`: why the `startCommitStale` arm deliberately skips `clearPendingStartInFlightLease` (the lease is no longer ours, so clearing it would clobber a successor's), and why the occupied-singleton probe deliberately sits *outside* the `withLocks` section (a provider liveness call blocks on socket I/O, and `withLocks` holds a cross-process city lock file per identifier spelling). Reviewers were explicit that the probe should not move, so this is comment-only.
- Generated Bazel `BUILD` files synced for the PR's three new source files — `e84a7fa1ce`.

Non-blocking follow-ups (shared for visibility — **not** completed in this PR, each filed so it is not lost):
- `ga-lzxwj` — [minor, Debuggability & Operability] Phase C (`build_desired_state.go:3411-3416`) and the dependency-floor site (`:3180-3182`) report `errPoolSessionNameUnavailable` through the generic `"(skipping)"` arm rather than the planning switch's "retrying next tick" wording, so the same condition reads two different ways depending on whether the holder has a bead. Left for its own pass because the remedy edits production error classification plus a test.
- `ga-kyslp` — [minor, Test Evidence Quality] no test drives the canceled-context verdict conversions at `session_lifecycle_parallel.go:1962` and `:1965`; an `==`→`!=` mutant there survives the suite. Correctness currently rests on an unconditional-return argument rather than on a test.
- `ga-0be8u` — [minor, Defensive Design] `Stop`'s dead-conn arm still deletes sidecars on `socketAlive` probe failure alone with no identity match, and still evaluates that probe while holding `p.mu`. Both items live in the same ~15 lines and should be taken together.

### Maintainer Changes
Two maintainer commits on top of your work, contributor authorship preserved:
- `1cc62fb160` fix(acp): stale Stop evicts its own dead conn and stays idempotent — 6 files changed, 62 insertions(+), 4 deletions(-)
- `e84a7fa1ce` chore(bazel): sync generated BUILD files for this PR's new sources — 3 files changed, 4 insertions(+)

### Final Review Status
Ready for the merge queue: final head `e84a7fa1ce` has passing required GitHub checks — all six required contexts (`Check`, `CI / required`, and the four `Analyze` jobs) are green, with 70 checks passing overall — and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

One non-required check, `bazel test (side-by-side)`, is red. It is not a merge gate and it is not this PR's: the failures are Bazel sandbox/runfiles hermeticity errors in `internal/beads`, a tree this PR does not touch, the failing targets' inputs are byte-identical between the base and this head, the same failure set appears on unrelated concurrent branches and on `main` itself, and a central fix is already in flight. `main` has also moved 18 commits since the reviewed base; the adopted branch was re-validated against the current base — clean rebase with an unchanged patch-id, clean `go vet`, and green `internal/session`, `internal/runtime/acp`, and a 669-test `cmd/gc` reconciler/session sweep including all five of this PR's new fence tests.

CI: https://github.com/gastownhall/gascity/actions/runs/35992663703

### Review Iterations
2 review passes. Iteration 1 requested changes on the `Stop` refusal branch (one blocker plus a same-root-cause major, both fixed by one localized edit); iteration 2 confirmed both fixes red/green-proven by two independent reviewer lanes and approved, raising only two minor findings, one of which was taken as a comment and the other filed as `ga-lzxwj` above.

---

PR #6123, opened 2026-09-07
fix(preflight): dial the identity_match probe with the scope-resolved Dolt credential (#5965)

### LINKED ISSUE #5965: preflight: the identity_match project_id probe dials an External Dolt endpoint with only the ambient GC_DOLT_PASSWORD, so every credential-less in-process open of an external rig store emits one rejected authentication at the hub

atbrace:
## What

Fix candidate 2 from #5965. `preflightDatabaseProjectIDReader` opened the target Dolt endpoint with only the ambient `GC_DOLT_PASSWORD` (`managedDoltPassword()`), so every credential-less in-process open of an external rig store (supervisor, exec orders, hooks) emitted one rejected authentication at the Dolt server before the check fell through to the native open, which authenticates with the real scoped credential.

## Fix

Resolve the probe password through the same `doltauth` ladder the native store open already uses (ambient env, then the auth scope's `.beads/.env`, then the credentials file for `host:port`) via `doltauth.AuthScopeRoot` + `doltauth.ResolveScopedFromEnv`, and thread it into a new `managedDoltOpenDatabaseWithPassword`. `managedDoltOpenDatabase` keeps its signature and behaviour for existing callers. `doltpool` keys handles on the password, so the probe and the native open share a pooled handle when they resolve the same credential.

This is complementary to #6007 (candidate 1, skip the dial when the scope defers to native-open): #6007 removes the probe for deferred scopes; this PR makes the probe authenticate correctly wherever it still runs, so the direct `project_id` check keeps working instead of being skipped.

## Validation (measured in production, not just tests)

Deployed on a home Gas City rig (one supervisor plus pool sessions against a Nomad-hosted `dolt sql-server`, ~10 conn/s):

| window | `Error authenticating user bd` | new connections |
|---|---|---|
| 60 min before cutover | 358 (4–22/min) | comparable rate |
| 65 min after cutover | **0** | 36,399 |

- `go test ./cmd/gc -run 'TestPreflightProbePassword|TestEnsureManagedDoltProjectID'` on this branch rebased onto `main` (26542454e): pass.
- Full-suite A/B on the rig's live lineage: the set of `--- FAIL` names is identical before and after (57 pre-existing environment-dependent failures, 0 new).

Closes #5965 together with #6007.

https://claude.ai/code/session_01LTeeybGgj3cWGqTRJJKeuQ

### DIFF
--- a/cmd/gc/beads_preflight_checker.go
+++ b/cmd/gc/beads_preflight_checker.go
@@ -4,9 +4,11 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/gastownhall/gascity/internal/beads/contract"
+	"github.com/gastownhall/gascity/internal/doltauth"
 	"github.com/gastownhall/gascity/internal/fsys"
 )
 
@@ -60,6 +62,21 @@ func preflightIdentityDeferredReader(cityPath string) func(scope string) bool {
 	}
 }
 
+// preflightProbePassword resolves the credential the identity_match probe
+// dials with. It walks the same scoped ladder the native store open uses
+// (ambient GC_DOLT_PASSWORD, then the auth scope's .beads/.env, then the
+// credentials file for host:port), so a credential-less in-process open of
+// an external rig store no longer emits one rejected authentication at the
+// Dolt server before the native open authenticates (gascity#5965).
+func preflightProbePassword(cityPath, scope string, target contract.DoltConnectionTarget) string {
+	authScopeRoot := doltauth.AuthScopeRoot(cityPath, scope, target)
+	env := map[string]string{
+		"GC_DOLT_HOST": strings.TrimSpace(target.Host),
+		"GC_DOLT_PORT": strings.TrimSpace(target.Port),
+	}
+	return doltauth.ResolveScopedFromEnv(authScopeRoot, target.User, env).Password
+}
+
 func preflightDatabaseProjectIDReader(cityPath string) func(scope string) (string, bool, error) {
 	return func(scope string) (string, bool, error) {
 		target, ok, err := canonicalScopeDoltTarget(cityPath, scope)
@@ -67,7 +84,7 @@ func preflightDatabaseProjectIDReader(cityPath string) func(scope string) (strin
 			return "", false, err
 		}
 		// Pooled handle owned by internal/doltpool; do not Close.
-		db, err := managedDoltOpenDatabase(target.Host, target.Port, target.User, target.Database)
+		db, err := managedDoltOpenDatabaseWithPassword(target.Host, target.Port, target.User, target.Database, preflightProbePassword(cityPath, scope, target))
 		if err != nil {
 			return "", false, err
 		}
--- a/cmd/gc/beads_preflight_checker_test.go
+++ b/cmd/gc/beads_preflight_checker_test.go
@@ -1,0 +1,72 @@
+package main
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/gastownhall/gascity/internal/beads/contract"
+)
+
+// The identity_match probe must dial with the same credential the native
+// store open resolves for the scope, not only the ambient GC_DOLT_PASSWORD.
+// Otherwise every credential-less in-process open of an external rig store
+// emits one rejected authentication at the Dolt server (gascity#5965).
+func TestPreflightProbePasswordResolvesScopeLocalCredential(t *testing.T) {
+	t.Setenv("GC_DOLT_PASSWORD", "")
+	t.Setenv("GC_DOLT_USER", "")
+	t.Setenv("BEADS_DOLT_PASSWORD", "")
+	t.Setenv("BEADS_CREDENTIALS_FILE", filepath.Join(t.TempDir(), "absent"))
+
+	cityPath := t.TempDir()
+	scope := filepath.Join(cityPath, "rig")
+	if err := os.MkdirAll(filepath.Join(scope, ".beads"), 0o755); err != nil {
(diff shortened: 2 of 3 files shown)


A3Ackerman:
Read against main `042e965f0`. Two notes, neither blocking.

1. The ladder's first leg is still the ambient `GC_DOLT_PASSWORD`, taken unconditionally (`internal/doltauth/auth.go:92-95`), so this closes the credential-less probe but not its mirror: a process whose env was projected for one external rig, probing a different endpoint (including the managed local store on 127.0.0.1), still dials with the projected password and is rejected the same way. That is what `managedDoltPassword()` does today, so it is not a regression here — just the boundary of what this PR fixes. We measured that mirror case on a city whose crew sessions carry a hub-projected env and run a decline for it (an ambient identity applies only to a host:port it provably matches). It is not filed upstream yet; I can file it as its own issue if that is useful.

2. Composition with #6007: with its defer-before-dial, an external scope never reaches this probe, so this ladder matters for the non-deferred targets. The two do not overlap in code (`internal/beads/contract` vs `cmd/gc`), so both can land.

PR #6135, opened 2026-09-08
Add doctor cleanup for stale executor identity stamps

quad341:
## What this changes

`gc doctor` now reports stale executor identity metadata (`gc.session_name`, `gc.work_dir`, and legacy `work_dir`) on work that is no longer in progress and no longer routed to the stamped session. `gc doctor --fix` clears those stale stamps across the City and active rigs while preserving `gc.routed_to` and all in-progress work.

This makes stale ownership residue repairable in one explicit doctor sweep without putting cleanup judgment into startup or role-specific code.

## Review notes

- The check runs only in explicit doctor flows; it is excluded from startup warmup.
- Route identity is normalized and re-derived live for each bead; there is no stored ID snapshot or hardcoded role behavior.
- Fixes use batch metadata clearing per bead and aggregate partial failures.

## Test plan

- [x] Run the repository's full 40-job local suite with rootless Podman.
- [x] Exercise stale, in-progress, idempotent, canonical-route, registration, and preflight-count cases.
- [x] Run policy, boundary, build, vet, formatting, docs, and native DoltLite checks.
- [x] Release gate: [`release-gates/ga-3717xe-executor-identity-residue-gate.md`](release-gates/ga-3717xe-executor-identity-residue-gate.md)

### DIFF
--- a/cmd/gc/cmd_doctor.go
+++ b/cmd/gc/cmd_doctor.go
@@ -322,6 +322,7 @@ func buildDoctorChecks(cityPath string, cfg *config.City, cfgErr error, opts bui
 		if storeOK {
 			register(doctor.NewBeadsStoreCheck(cityPath, openStoreResultForCity(cityPath)))
 			register(newV2RoutedToNamespaceCheck(cfg, cityPath, storeFactory))
+			register(newExecutorIdentityResidueCheck(cfg, cityPath, storeFactory))
 			register(newCensusOwnerLivenessCheck(cfg, cityPath, storeFactory))
 			register(newRunTargetRoutedToBackfillCheck(cfg, cityPath, storeFactory))
 			register(newRouteRecoveryQuarantineCheck(cfg, cityPath, storeFactory))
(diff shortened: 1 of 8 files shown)


quad341:
@quad341 — the stand-downs are the best part of this. Deferring to `poolSlotWorkDirRepairFor` instead of racing it, and keeping the work_dir keys that `worktreeSpecForBead` needs for its fail-closed conflict check, are both subtle and both argued correctly in the comments. The disjoint-trigger design, where the fix clears exactly the keys that fired and nothing more, is the right shape for code that deletes state. Thirteen tests is the level of care a metadata sweep deserves.

One thing I'll fix on our side before merging. The route-identity index is built per scope from the store being scanned, and session beads only ever live in the city session-class store. So rig scopes see an empty index, and the pool-instance and alias protections your round-two and round-three tests establish never reach rig work. I'll union the city index into each rig scope following the backfill pattern in the detached-orphan sweep, source it through `session.ListAllSessionBeads` so label-only and closed session beads count, add a live re-read before each clear, and add rig-scope tests. Then build, vet, run the doctor set, and merge with your commits intact.

The design is yours and it holds up. This is integration plumbing we should have caught for you.


---

PR #6138, opened 2026-09-08
fix(workrecord): resolve the close gate's repository from the bead's owner, not the store it was read through

julianknutsen:
## The bug

The ADR-0009 work-record close gate's reachability clause — a `shipped` outcome
must name a commit reachable on the stamped branch — is the one clause that
cannot be answered from the bead alone. It needs a checkout to ask git about,
and both close doors picked that checkout from where the row was **read**.

For one population those are different repositories: a rig's work step that a
relocated class binding holds. Its commits are on the **rig's** checkout, and
both doors asked the **city's**:

- CLI class door: `serveBdByIDResolved` hands `evaluateWorkRecordCloseGate` the
  `cityPath` as its `repoFallback`.
- HTTP door: `workRecordRepoDir` matched the store that answered against the
  configured rigs, and a binding is not one, so `a496031ba0`'s unconditional
  city fallback caught it.

With `GC_WORK_RECORD_ENFORCE` on, a compliant `gc.work_outcome=shipped` close of
such a bead was refused with `gc.work_commit … is not reachable on
gc.work_branch main` — against a repository that was never the bead's. That
population is not exotic: a task-typed molecule step carrying `gc.root_bead_id`
and no `gc.kind` is exactly `workrecord.Gated`'s population, and commonly
records no `gc.work_dir`.

## Mechanism

The residency doctrine already separates the two facts: the store a row is
served from says where it **lives**; `gc.root_store_ref` says who **owns** it.

`workrecord.RepoDirFor(bead, ScopeDirs) (string, ScopeKind)` is that rule, and
both doors run it:

| arm | answer | kind |
| --- | --- | --- |
| `gc.work_dir` set | that directory | `ScopeWorkDir` |
| owner is `rig:<name>` | that rig's checkout | `ScopeRig` |
| owner is `city:<name>` or `class:<token>` (a binding is city scope) | the city checkout | `ScopeCity` |
| owner named, no checkout configured for it | `""` | `ScopeUnknown` |
| no owner recorded | `""` | `ScopeUnrooted` — the caller's legacy answer stands |

An owner no checkout is configured for is **unknown, never the city**: falling
back there asks about a repository that is not the bead's, which under
enforcement is a false refusal rather than a degraded clause.

`ScopeUnrooted` is what keeps single-store cities byte-identical. Every city
that predates the residency census is in that population, so each door keeps the
answer it already gave — which is why `internal/api`'s store-identity match
**stays** (as `workRecordLegacyRepoDir`) rather than being deleted: it is that
plane's legacy answer, and deleting it would send a rig-resident bead with no
root ref to the city checkout. Each plane supplies only its own checkout table
(`workRecordScopeDirs` in `internal/api`, `workRecordRepoDirs` in `cmd/gc`).

### Degrade semantics, converged

`#5600` gave the HTTP door an honest degrade: an unknown repository warns
`reachability unverified` and the close proceeds, while a bead with no outcome
at all is still refused. The CLI door failed **closed** on the same condition —
`CommitReachableOnBranch` reads `""` as "not reachable" — so it now takes the
same degrade. Both doors print the same sentence, single-sourced as
`workrecord.ReachabilityUnverifiedNote`.

### Two implementation notes

- **`internal/workrecord` imports `internal/storeref`** for the
  `ScopeRigContext` parse. There is no cycle — nothing in `storeref`'s
  dependency closure imports `workrecord` (`go list -deps ./internal/storeref |
  grep workrecord` is empty) — so the ref vocabulary keeps one reader instead of
  a copy per plane. The func-parameter fallback was therefore not needed.
- **The class door loads the rig table lazily** (`cityRigsLoader`, a
  `sync.Once`). `maybeRouteBdByID` is entered from `doBd`'s cost gate before
  that invocation has read `city.toml`, and the rig table answers one question
  only a close of a gated bead asks — a show, a claim or a dep walk still pays
  nothing. A config this invocation cannot read yields no rigs, which degrades
  rather than manufacturing a refusal. Keeping `maybeRouteBdByID`'s signature
  also means the door rows below exercise the **real** config path a production
  close takes, rather than a `*config.City` a test handed in.

This **supersedes the store-identity fallback `a496031ba0` added**: that commit
made a binding-owned bead answer to the city, which is right for a
binding-**owned** bead and wrong for a rig-owned one the binding merely holds.

## RED → GREEN

Tests were written first. Every row below was observed failing against
`origin/main` + the tests, then passing with the fix.

| row | RED (before) | GREEN (after) |
| --- | --- | --- |
| `internal/workrecord` `TestRepoDirFor` (12 rows: work_dir wins; rig owner → rig dir; city owner → city dir; binding owner → city dir; unconfigured rig → unknown; path-less rig → unknown; city owner with no city dir → unknown; no owner → unrooted; bare legacy label → unrooted; `rig:` → unrooted; whitespace; no metadata) | build failure: `undefined: ScopeDirs / ScopeKind / RepoDirFor` | `ok` |
| `internal/workrecord` `TestRepoDirForWithoutScopeDirs` (nil table does not panic) | same build failure | `ok` |
| **API door** `TestAPIBeadCloseAsksTheOwningRigsCheckoutForABindingResidentBead` — binding-resident, `rig:myrig`-owned, `shipped`, commit on the rig checkout only, `GC_WORK_RECORD_ENFORCE=1` | `FAIL … conflict: bead gcg-owned does not satisfy the work-record close contract: gc.work_commit 1a7273a1… is not reachable on gc.work_branch main` | `PASS` (closes; no `reachability unverified` in the gate log) |
| **API door**, same bead with a commit no checkout holds | (already refused) | `PASS` — still a 409 `not reachable` |
| **CLI door** `TestBdCloseClassResidentAsksTheOwningRigsCheckout` — same shape through `maybeRouteBdByID` | `FAIL … exited 1: gc bd: work-record gate (enforced): close of gc-rigowned1: gc.work_commit 5c6d6f37… is not reachable on gc.work_branch main` | `PASS` (exit 0, class row closed, gate silent) |
| **CLI door** `TestBdCloseClassResidentBlocksACommitNoCheckoutHolds` | (already refused) | `PASS` — still exit 1 with `not reachable` |
| **CLI door** `TestEvaluateWorkRecordCloseGateDegradesForAnUnknownOwner` — unknown owner warns `reachability unverified` and proceeds; the outcome-less control is still refused | fails: the CLI door had no degrade and blocked on `""` | `PASS` |

Both door rows own **two real git repositories** — the rig's holds the commit,
the city's is a real repository that does not. That is the control separating
"asked the wrong repository" from "asked no repository at all": both answer
not-reachable, and only the first is this bug. Neither row adds an `os/exec`
construction (they reuse `runResolverGit` / `runGit`, and read the commit SHA
from the loose `main` ref), so the test-resource census is unmoved.

## The pins bite (mutation)

Dropping the root-ref arm from `RepoDirFor` on disk — `return "", ScopeUnrooted`
unconditionally after the `gc.work_dir` arm, so every bead falls back to the
caller's legacy answer — and re-running:

```
--- FAIL: TestRepoDirFor/a_rig-owned_bead_answers_to_the_rig_checkout
--- FAIL: TestRepoDirFor/a_city-owned_bead_answers_to_the_city_checkout
--- FAIL: TestRepoDirFor/a_binding-owned_bead_is_city_scope
--- FAIL: TestRepoDirFor/a_rig_this_plane_does_not_configure_is_unknown,_not_the_city
--- FAIL: TestRepoDirFor/a_configured_rig_that_names_no_checkout_is_unknown,_not_the_city
--- FAIL: TestRepoDirFor/a_city-owned_bead_in_a_plane_with_no_city_checkout_is_unknown
--- FAIL: TestRepoDirFor/surrounding_whitespace_does_not_hide_the_owner
--- FAIL: TestRepoDirForWithoutScopeDirs

--- FAIL: TestAPIBeadCloseAsksTheOwningRigsCheckoutForABindingResidentBead
    close of a rig-owned bead whose commit is on the rig checkout: conflict:
    … gc.work_commit 730ad14f… is not reachable on gc.work_branch main

--- FAIL: TestBdCloseClassResidentAsksTheOwningRigsCheckout
    close … exited 1: gc bd: work-record gate (enforced): … gc.work_commit
    d650ca37… is not reachable on gc.work_branch main
--- FAIL: TestEvaluateWorkRecordCloseGateDegradesForAnUnknownOwner
    an unanswerable reachability clause blocked the close
```

Both doors fail, on the rig-rooted rows, for the original reason. Mutation
reverted; the rows pass again.

## Gates

| gate | exit |
| --- | --- |
| `go build ./...` | 0 |
| `go vet ./...` | 0 |
| `gofmt -l cmd internal` | 0 (no output) |
| `golangci-lint run ./internal/workrecord/... ./internal/api/... ./cmd/gc/...` | 0 (`0 issues.`) |
| `./scripts/check-residency-boundary.sh` | 0 — **baseline unchanged**, no row moved |
| `go test ./scripts/...` (the residency guard's AST half) | 0 |
| `go test ./internal/testpolicy/resourcecensus/...` (staged tree) | 0 — census unmoved |
| `bash scripts/check-core-boundary.sh` | 0 |
| `go test ./internal/workrecord/... ./internal/api/... -count=1` | 0 |
| `go test ./cmd/gc -run 'WorkRecord\|ByID\|Residency' -count=1` | 0 |
| `make test-fast-parallel` | 0 (`All fast jobs passed`) |
| `make dashboard-ci` | 0 — **no generated artifact moved** (`internal/api/openapi.json`, `docs/reference/schema/`, the generated TS client are all unchanged; this change touches no wire type) |

No base-owned failure was hit; nothing was worked around.

Refs: ga-hw5f7, ga-lkxrg, ga-j4p3y

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -77,6 +77,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **The work-record close gate asks the repository the bead's OWNER points at,
+  not the store it was read through.** A rig's work step that a relocated class
+  binding holds has its commits on the rig's checkout, and both close doors
+  asked the city's instead — the CLI class door hands its gate the city path,
+  and the HTTP door matched the store that answered against the configured rigs,
+  which a binding is not. With `GC_WORK_RECORD_ENFORCE` on, a compliant
+  `gc.work_outcome=shipped` close of such a bead was refused with "commit is not
+  reachable" against a repository that was never the bead's. Both doors now
+  resolve the repository through one rule (`workrecord.RepoDirFor`): the bead's
+  own `gc.work_dir`, else the scope `gc.root_store_ref` records — a rig owner to
+  that rig's checkout, a city or binding owner to the city's. An owner no
+  checkout is configured for is "unknown" rather than the city, and the
+  reachability clause degrades to a warning there on both doors instead of
+  refusing a close neither can judge; a bead with no outcome at all is still
+  refused. A bead that records no owner keeps the answer its door already gave,
+  so single-store cities are unchanged.
+
 - **The infra-class cutover now carries dependency-edge payloads.** Every
   within-infra edge the copy re-added went in through a writer that clears the
   pair's metadata sidecar, so the binding received those edges with their
--- a/cmd/gc/cmd_bd_by_id.go
+++ b/cmd/gc/cmd_bd_by_id.go
@@ -917,7 +917,17 @@ func maybeRouteBdByID(cityPath, rigName string, bdArgs []string, stdout, stderr 
 	if !served {
 		return refuseUnservedClassMutation(door, bdArgs, named, namesClassBead, mutationIDs, stderr)
 	}
-	return serveBdByIDResolved(door, op, bdArgs, rigName, cityPath, stdout, stderr)
+	return serveBdByIDResolved(door, op, bdArgs, rigName, classDoorRepoDirs(cityPath), stdout, stderr)
+}
+
+// classDoorRepoDirs is the checkout table this door hands its close gate. A
+// binding is a store rather than a checkout, so the city is what a bead the
+// binding merely HOLDS answers to; a bead that names a rig as its owner answers
+// to that rig's checkout instead, which is why the rig table travels with the
+// city path. The table loads the config lazily (cityRigsLoader), so a read this
+// door serves pays for it only when a close actually asks.
+func classDoorRepoDirs(cityPath string) workRecordRepoDirs {
+	return workRecordRepoDirs{cityPath: cityPath, legacy: cityPath, rigs: cityRigsLoader(cityPath)}
 }
 
 // refuseUnservedClassMutation answers a by-ID invocation that addresses a bead
@@ -950,9 +960,9 @@ func refuseUnservedClassMutation(door bdByIDClassDoor, bdArgs []string, named st
 // resolves the id against the class binding, refuses the cases this surface
 // must not serve (a work-store id it does not own, a reserved-prefix id with no
 // row, an explicit --rig work scope), runs the ADR-0009 work-record close gate,
-// and dispatches the verb against the class graph. repoFallback is the git repo
-// the close gate falls back to when a closing bead carries no gc.work_dir.
-func serveBdByIDResolved(door bdByIDClassDoor, op bdByIDOp, bdArgs []string, rigName, repoFallback string, stdout, stderr io.Writer) (int, bool) {
+// and dispatches the verb against the class graph. repoDirs is the checkout
+// table the close gate resolves a closing bead's repository from.
+func serveBdByIDResolved(door bdByIDClassDoor, op bdByIDOp, bdArgs []string, rigName string, repoDirs workRecordRepoDirs, stdout, stderr io.Writer) (int, bool) {
 	resolution, err := door.resolve(op.ID)
 	if err != nil {
 		fmt.Fprintf(stderr, "gc bd: %v\n", err) //nolint:errcheck // best-effort stderr
@@ -983,7 +993,7 @@ func serveBdByIDResolved(door bdByIDClassDoor, op bdByIDOp, bdArgs []string, rig
 		// So neither is taken. See refuseRigScopedClassOwnedTarget.
 		return refuseRigScopedClassOwnedTarget(door, op.ID, rig, stderr)
 	}
-	if gateBdByIDClassClose(door, op, bdArgs, resolution, repoFallback, stderr) {
+	if gateBdByIDClassClose(door, op, bdArgs, resolution, repoDirs, stderr) {
 		return 1, true
 	}
 	switch op.Verb {
@@ -1023,12 +1033,12 @@ func serveBdByIDResolved(door bdByIDClassDoor, op bdByIDOp, bdArgs []string, rig
 // gate reuses it rather than re-reading the class store, and door.Store answers
 // any other id the argv might name. Returns true only when the close must be
 // blocked (enforcement on and the work record invalid).
-func gateBdByIDClassClose(door bdByIDClassDoor, op bdByIDOp, bdArgs []string, resolution bdByIDResolution, repoFallback string, stderr io.Writer) bool {
+func gateBdByIDClassClose(door bdByIDClassDoor, op bdByIDOp, bdArgs []string, resolution bdByIDResolution, repoDirs workRecordRepoDirs, stderr io.Writer) bool {
 	if op.Verb != bdByIDClose && op.Verb != bdByIDUpdate {
 		return false
 	}
 	preFetched := map[string]beads.Bead{op.ID: resolution.Bead}
-	return evaluateWorkRecordCloseGate(bdArgs, door.Store, preFetched, repoFallback, workRecordEnforceEnabled(), stderr)
+	return evaluateWorkRecordCloseGate(bdArgs, door.Store, preFetched, repoDirs, workRecordEnforceEnabled(), stderr)
 }
 
 // refuseRigScopedClassOwnedTarget refuses a by-ID invocation that pins a rig
--- a/cmd/gc/work_record_gate.go
+++ b/cmd/gc/work_record_gate.go
@@ -113,6 +113,14 @@ func (d workRecordRepoDirs) CityDir() string { return strings.TrimSpace(d.cityPa
 // RigDir returns the named rig's checkout and whether this city configures that
 // rig at all. A configured rig that names no checkout answers ("", true), which
 // the rule reads as unknown rather than as the city.
+//
+// The name is matched case-insensitively, deliberately more forgiving than the
+// exact-match lookups this answer feeds (workdir.RigRootForName, sling's
+// rigSuspended). Refs are stamped by storeref.RigRef from the configured rig
+// name, so only a hand-stamped or historically-buggy ref differs by case;
+// matching it judges the close against that rig's checkout instead of degrading
+// to unverified. Both doors spell this matcher the same way, so whichever way it
+// resolves, one bead is still judged against one repository.
 func (d workRecordRepoDirs) RigDir(name string) (string, bool) {
 	if d.rigs == nil {
 		return "", false
--- a/cmd/gc/work_record_gate.go
+++ b/cmd/gc/work_record_gate.go
@@ -5,8 +5,8 @@ import (
 	"fmt"
 	"io"
 	"strings"
+	"sync"
 
-	"github.com/gastownhall/gascity/internal/beadmeta"
 	"github.com/gastownhall/gascity/internal/beads"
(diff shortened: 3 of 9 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277311042

quad341 (dismissed):
@julianknutsen — the diagnosis here is the good part. Separating "the store a row is served from" (where it lives) from `gc.root_store_ref` (who owns it), and then noticing that *both* close doors had quietly conflated them, is the kind of thing that's much easier to see once written down than before. A rig's work step held by a relocated class binding is exactly the population where those diverge, and "commit is not reachable" against a repository that was never the bead's is a miserable error to debug from the message alone.

Two things worth calling out. Standing up **two real git repositories** in the door tests, where only the rig's holds the commit, is the control that makes those rows mean something — a stubbed oracle would have proved only the stub. And the mutation section paid off: a reviewer dropped the root-ref arm and got your failure list back exactly, then green on restore. That turns a slow review into a fast one.

Merging as-is. One thing I'm noting for the record, not asking you to change: the CLI door now warns rather than refuses when it can't resolve a checkout. Converging on the HTTP door is right, but it does relax enforcement rather than just relocate it, so I want that visible. Thanks for the writeup — it's genuinely why this went quickly.


---

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR makes both work-record close
gates resolve the repository from the **bead's own owner** — its `gc.work_dir`,
else the rig named by its `gc.root_store_ref` — instead of from whichever store
the row happened to be read through, and routes both planes through one shared
rule (`workrecord.RepoDirFor`) with each plane supplying only its own checkout
table. That matters because the old read-site resolution produced false
"commit is not reachable" refusals: a rig-owned work bead resident in a
relocated class binding was being asked about against the *city* checkout, so a
perfectly good close was refused under enforcement. It also converges the CLI
door's fail-closed behaviour onto the HTTP door's honest degrade, so one bead is
judged one way however it is closed.

branch.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **Read-site repository resolution on both doors.** The CLI class door passed
  `repoFallback = cityPath` and the API matched the answering store against
  configured rigs (which a binding never is). Both now run
  `workrecord.RepoDirFor`, pinned end-to-end on both planes against two *real*
  checkouts where only the rig's holds the commit — which distinguishes "asked
  the wrong repository" from "asked no repository" — each with a commit-nowhere
  control.
- **CLI fail-closed converged on the HTTP degrade.** Only the reachability
  clause degrades: `ValidateOnClose` still refuses an outcome-less bead
  regardless of `repoDir`, controlled on both planes, and both doors emit the
  single `workrecord.ReachabilityUnverifiedNote` constant.
- **Backward compatibility for unrooted beads.** `ScopeUnrooted` returns each
  door's prior answer, and the untouched pre-existing tests pin that
  compatibility arm.
- **Both doors resolve from the projected bead**, so an atomic close that stamps
  `gc.work_dir` or the owner in the same invocation is judged on what it writes.
- **Call-site census complete — no third door.** Independently re-verified: no
  caller was left on the old signature, and no `repoFallback` reference
  survives.

Review findings addressed:
- *Change Impact / Blast Radius (minor)* — the legacy shared-file first-match
  residual is real but was undisclosed in this PR's own new doc comment, which
  reasoned about populations without naming the store that many rigs claim.
  Fixed in `4e1988c262`: `workRecordLegacyRepoDir`'s doc now names the
  shared-file many-claimants shape and its first-match answer, and notes that
  `RepoDirFor` strictly narrows it.
- *Architectural Consistency (nit)* — the new `RigDir` loops match rig names
  with `EqualFold`, which is more forgiving than the exact-match lookups their
  answer feeds. Fixed in `4e1988c262`: both implementations now carry a comment
  saying the matcher is deliberately more forgiving, why that is safe given
  `storeref.RigRef` stamps from the canonical config name, and that both doors
  spell the matcher identically.
- *Contract & Interface Fidelity (nit)* — the two planes' scope resolvers spell
  the same repository differently (one canonicalizes symlinks, the other is
  purely lexical). Fixed in `4e1988c262`: `ScopeDirs` now documents that a
  returned directory is for handing to git, not for comparing as a string.

Behaviour note for future readers (deliberate, and CHANGELOG-documented, not a
defect): a bead whose `gc.root_store_ref` names a rig this city no longer
configures now degrades to unverified-pass under enforcement, where it was
previously judged against the scope it was read through. A rig rename or
decommission that does not rewrite stamped refs quietly moves that rig's older
closes from judged to warned. The outcome-less clause still refuses, so no close
that recorded nothing is accepted.

- `ga-wub6d` — *Test Evidence Quality, minor:* the bd fall-through door
  (`runWorkRecordCloseGate`) received the same owner-resolution wiring as the two
  tested doors, but nothing exercises it. Its only test passes `cfg = nil` with a
  shipped bead carrying no commit and no branch, so the reachability oracle is
  never reached and `repoDir` is never consulted. A mutation probe confirmed it:
  replacing the wiring with `rigs: func() []config.Rig { return nil }` leaves the
  full `cmd/gc` package green. The wiring is correct — it is simply unguarded.
  Not addressed in this PR.
- `ga-teewy` — *Debuggability & Operability, minor:* `cityRigsLoader` discards
  `loadCityConfigAdvisory`'s error, so a transiently unreadable `city.toml` or a
  broken pack import yields no rigs and prints exactly the same unverified note
  as a genuinely unconfigured rig. The degrade behaviour itself is deliberate and
  correctly documented; the gap is only the missing diagnostic trail.
  Not addressed in this PR.

### Maintainer Changes
`4e1988c262` docs(workrecord): disclose the legacy first-match residual, the
case-insensitive rig match, and the two doors' path spellings — 3 files changed,
34 insertions(+), documentation only with no behaviour change. Your commit
`6f98bd656d` is preserved exactly as authored.

### Final Review Status
Ready for the merge queue: final head `4e1988c2` has passing required GitHub
checks (78 successful checks, none failing) and GitHub reports the PR as
`MERGEABLE` / `CLEAN`. The adopted branch was also re-validated against the
latest `main`: the rebase onto `f7fe81dc` is conflict-free with an unchanged
patch-id, `main`'s only intervening commit touches `internal/doctor/*_test.go`
and so has zero file overlap with this PR, and `go vet` plus the
`internal/workrecord`, `internal/api`, and `cmd/gc` work-record suites are green
on the refreshed tree. The merge-ready metadata is in place, so I am marking this
PR `status/merge-ready` for the merge queue.

CI: https://github.com/gastownhall/gascity/actions/runs/34196835145

### Review Iterations
1 review pass, 0 fix iterations. Three independent reviewers graded the same
9-file range with full cross-file coverage and converged on approve — no blockers
and no majors in any category. The synthesis verified the load-bearing claims by
execution rather than by reading (mutation probe on the fall-through wiring, an
independent re-run of the call-site census, and a rationale audit that caught and
corrected one reviewer's false supporting claim about `EqualFold` consumers). The
three disclosure findings were resolved by the maintainer docs commit; the two
remaining minors are carried as the follow-up beads listed above.

---

PR #6139, opened 2026-09-08
fix(test): guard the doctor tests' test-owned HOME with the retrying cleanup (ga-531fk)

julianknutsen:
## Summary

The doctor tests that spawn `bd` pin `$HOME` at a `t.TempDir()`, because `bd`'s
config precedence falls through to `$HOME/.beads/config.yaml` and a machine HOME
with `dolt.shared-server: true` otherwise routes those subprocesses at the shared
server no matter which env vars the test scrubs (ga-zxpfic). That HOME is not
just a decoy directory — `bd` writes `$HOME/.beads/` into it, and the embedded
dolt background writer can hold files there open a few dozen ms past the owning
subprocess's apparent exit.

`retryRemoveAllForTest` exists precisely to absorb that lag, but it was only
registered for the *working* dir. The HOME dir was left to `t.TempDir()`'s
single-shot deferred `RemoveAll`, which lost the race on the Mac CI lane:

```
testing.go:1464: TempDir RemoveAll cleanup: unlinkat
  /var/folders/.../TestCustomTypesCheck_TableDrift3082042356/001: directory not empty
--- FAIL: TestCustomTypesCheck_TableDrift (5.95s)
```

Mac Regression run 34174946211, job 101902481774. The test body had already
passed; only cleanup failed. `t.TempDir()` numbers sequentially from `001`, so
the failing `001` is the first call — the unguarded HOME.

## What changed

The guard now comes with the directory instead of being a second line each test
has to remember:

- `guardedTempDir(t)` returns a `t.TempDir()` with the retrying cleanup already
  registered. Registration order matters and is documented: because it registers
  *after* `t.TempDir()` registered its own, LIFO ordering runs the retrying
  removal first and leaves `TempDir`'s `RemoveAll` nothing to trip over.
- `testOwnedHome(t)` pins `$HOME` at one of those and returns it.

Every test in `internal/doctor` that pins a test-owned HOME for `bd`
subprocesses now uses them for **both** of its temp dirs:

| Test | HOME dir | working dir |
| --- | --- | --- |
| `TestCustomTypesCheck_TableDrift` | was unguarded → guarded | already guarded |
| `TestCustomTypesCheck_TableDriftUsesTestOwnedDoltContext` | was unguarded → guarded | already guarded |
| `TestCustomTypesCheck_MissingTypes` | was unguarded → guarded | was unguarded → guarded |

Those three are the complete set: a `t.Setenv("HOME"` sweep of the package
returns one other site, `setupFakeGitConfig` in `checks_beads_role_test.go`,
which spawns `git config` and never `bd`, so it has no dolt writer to outlive it.

The helpers and `retryRemoveAllForTest` move to a small
`internal/doctor/tempdir_guard_test.go` so the check's own test file is only
about the check.

## On proving it

The Mac race is not reproducible on this Linux host, so the retry behaviour is
pinned directly rather than by re-running the flake. `retryRemoveAllForTest`
keeps its exact semantics but its loop moves into
`retryRemoveAll(dir, remove, attempts, delay)` with the removal injected, which
makes both paths testable with no sleeps and no timing:

- `TestRetryRemoveAllRetriesUntilRemovalSucceeds` — a fake that fails twice then
  succeeds is called exactly three times.
- `TestRetryRemoveAllStopsAtItsAttemptBudget` — a fake that always fails is
  called exactly `attempts` times and the helper returns rather than hanging.
- `TestTestOwnedHomePinsHOMEToAGuardedTempDir` — `$HOME` is the returned dir, and
  the dir is gone once the owning subtest ends.
- `TestGuardedTempDirRemovesItsDirWhenTheTestEnds` — pins the structural half
  (the dir is test-scoped); its comment says outright that the retry half is
  covered by the two tests above, since a single `RemoveAll` of an idle dir
  succeeds on the first attempt.

Each new test was mutation-checked: breaking the loop bound to a single attempt
fails both retry tests (`remove called 1 times, want 3` / `want 4`), and dropping
the `t.Setenv` from `testOwnedHome` fails the HOME test with the real
`/home/ubuntu`. Beyond that, the change is mechanically sufficient by
construction: the guard being applied to the HOME dir is the same guard that has
protected the working dir of these tests since it was written.

## Gates

All run on this branch; every one exited 0.

| Gate | Exit |
| --- | --- |
| `go build ./...` | 0 |
| `go vet ./internal/doctor/...` | 0 |
| `gofmt -l internal/doctor` (no output) | 0 |
| `golangci-lint run ./internal/doctor/...` — 0 issues | 0 |
| `go test ./internal/doctor/ -count=1` | 0 |
| `make test-fast-parallel` — "All fast jobs passed" | 0 |
| `.githooks/pre-commit` (ran on the commit) | 0 |
| `.githooks/pre-push` (ran on the push, no `--no-verify`) | 0 |

`TestCustomTypesCheck_TableDrift` **ran** here rather than skipping — the PATH
`bd` is `1.1.0 (Enterprise dev, 8de373f05)`, so the CGO-less-binary skip of
ga-z3ehf did not apply — and passed in 6.72s, as did
`TestCustomTypesCheck_TableDriftUsesTestOwnedDoltContext` in 4.74s.

One base-owned wrinkle, disclosed rather than left to be found: the first
`make test-fast-parallel` run failed `TestRepositoryLedgerMatchesCensusAndDocumentation`
with a stale `fixed_sleep` baseline (320 vs 321 calls). That is the known
resource-census behaviour on **unstaged** files — the census saw the removed
literal `time.Sleep(50 * time.Millisecond)` but not the new file that carries it.
`git add` and re-run: green, with no baseline edit needed. The gate table above
is the staged re-run.

Refs ga-531fk. Unblocks ga-j4p3y: this flake is what broke #5644's pre-approval
CI on the Mac lane.

### DIFF
--- a/internal/doctor/checks_custom_types_test.go
+++ b/internal/doctor/checks_custom_types_test.go
@@ -8,7 +8,6 @@ import (
 	"slices"
 	"strings"
 	"testing"
-	"time"
 
 	"github.com/gastownhall/gascity/internal/beads/contract"
 	"github.com/gastownhall/gascity/internal/fsys"
@@ -48,9 +47,9 @@ func TestCustomTypesCheck_MissingTypes(t *testing.T) {
 	// this check StatusOK, defeating the assertion below. Pin a test-owned
 	// HOME so that fallback file doesn't exist. See ga-zxpfic and
 	// TestCustomTypesCheck_TableDriftUsesTestOwnedDoltContext.
-	t.Setenv("HOME", t.TempDir())
+	testOwnedHome(t)
 
-	dir := t.TempDir()
+	dir := guardedTempDir(t)
 	beadsDir := filepath.Join(dir, ".beads")
 	if err := os.MkdirAll(beadsDir, 0o700); err != nil {
 		t.Fatal(err)
@@ -65,22 +64,6 @@ func TestCustomTypesCheck_MissingTypes(t *testing.T) {
 	}
 	if !c.CanFix() {
 		t.Fatal("CanFix should return true")
-	}
-}
-
-// retryRemoveAllForTest retries os.RemoveAll briefly to absorb a lingering
-// embedded-dolt background writer that can hold files open a few dozen ms
-// past the owning bd subprocess's apparent exit — which otherwise races
-// t.TempDir()'s single-shot RemoveAll cleanup with an intermittent
-// "directory not empty" error. Falls through silently on final failure so
-// TempDir's own best-effort cleanup still gets the last word.
-func retryRemoveAllForTest(t *testing.T, dir string) {
-	t.Helper()
-	for i := 0; i < 10; i++ {
-		if err := os.RemoveAll(dir); err == nil {
-			return
-		}
-		time.Sleep(50 * time.Millisecond)
 	}
 }
 
@@ -126,11 +109,9 @@ func TestCustomTypesCheck_TableDrift(t *testing.T) {
 	// to the shared server regardless of the vars above. Pin a test-owned
 	// HOME so that fallback file doesn't exist. See ga-zxpfic and
 	// TestCustomTypesCheck_TableDriftUsesTestOwnedDoltContext.
-	home := t.TempDir()
-	t.Setenv("HOME", home)
+	testOwnedHome(t)
 
-	dir := t.TempDir()
-	t.Cleanup(func() { retryRemoveAllForTest(t, dir) })
+	dir := guardedTempDir(t)
 
 	runBD := func(args ...string) string {
 		t.Helper()
@@ -220,11 +201,9 @@ func TestCustomTypesCheck_TableDriftUsesTestOwnedDoltContext(t *testing.T) {
 		t.Setenv(key, "")
 	}
 
-	home := t.TempDir()
-	t.Setenv("HOME", home)
+	home := testOwnedHome(t)
 
-	dir := t.TempDir()
-	t.Cleanup(func() { retryRemoveAllForTest(t, dir) })
+	dir := guardedTempDir(t)
 
 	runBD := func(args ...string) string {
 		t.Helper()
--- a/internal/doctor/tempdir_guard_test.go
+++ b/internal/doctor/tempdir_guard_test.go
@@ -1,0 +1,122 @@
+package doctor
+
+import (
+	"errors"
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+)
+
+const (
+	removeRetryAttempts = 10
+	removeRetryDelay    = 50 * time.Millisecond
+)
+
+// retryRemoveAllForTest retries os.RemoveAll briefly to absorb a lingering
+// embedded-dolt background writer that can hold files open a few dozen ms
+// past the owning bd subprocess's apparent exit — which otherwise races
+// t.TempDir()'s single-shot RemoveAll cleanup with an intermittent
+// "directory not empty" error. Falls through silently on final failure so
+// TempDir's own best-effort cleanup still gets the last word.
+func retryRemoveAllForTest(t *testing.T, dir string) {
+	t.Helper()
+	retryRemoveAll(dir, os.RemoveAll, removeRetryAttempts, removeRetryDelay)
+}
+
+// retryRemoveAll calls remove(dir) until it reports success or the attempt
+// budget runs out, pausing delay between tries.
+func retryRemoveAll(dir string, remove func(string) error, attempts int, delay time.Duration) {
+	for i := 0; i < attempts; i++ {
+		if err := remove(dir); err == nil {
+			return
+		}
+		time.Sleep(delay)
+	}
+}
+
+// guardedTempDir returns a t.TempDir() whose removal is retried by
+// retryRemoveAllForTest. Registering the cleanup after t.TempDir() has
+// registered its own means LIFO ordering runs the retrying removal first,
+// leaving TempDir's single-shot RemoveAll nothing to trip over. Every temp
+// dir a bd subprocess writes into needs this, so bead ga-531fk — where the
+// dir pinned as HOME was the one missing the guard and failed a Mac CI run
+// after its test body had already passed — cannot repeat: the guard now
+// comes with the directory instead of being a second line to remember.
+func guardedTempDir(t *testing.T) string {
+	t.Helper()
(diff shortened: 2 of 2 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277310960

quad341 (dismissed):
@julianknutsen — thank you for chasing this one all the way to the bottom instead of sprinkling a retry and moving on. Tracing the failing `001` back through `t.TempDir()`'s sequential numbering to identify the *HOME* dir as the unguarded one is genuinely good detective work, and the `t.Setenv("HOME"` sweep proving `setupFakeGitConfig` is the only other site — and that it spawns `git`, not `bd`, so it has no dolt writer to outlive it — is exactly the "here's why this is the complete set" I want to see. I re-ran that sweep on your head and it holds.

Making the guard come *with* the directory, rather than being a second line each test has to remember, is the right fix for a bug whose root cause was "someone forgot the second line."

I also appreciate you disclosing the `fixed_sleep` census wobble rather than leaving it to be found — the ledger test is green at your head and the sleep accounting balances, so no baseline edit needed.

Merging as-is. Nothing for me to touch.


---


Review coverage:
- completed: `qwen claude`
- unavailable: `codex`

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR closes a real Mac-only flake in the `internal/doctor` tests: `t.TempDir()` registers a single-shot `RemoveAll` on its first call, so the test-owned `HOME` directory was torn down without a retry while embedded-bd was still writing `$HOME/.beads/`, producing `unlinkat .../001: directory not empty` *after* the test body had already passed. It replaces that with a `guardedTempDir` helper that hands back the directory with its retrying cleanup already registered, so the guard travels with the directory instead of being a second line every new call site has to remember. That framing is what makes this worth more than the flake it fixes — the failure mode it closes is the worst kind (red after green, on one platform, with no signal about which directory lost the race), and the helper shape stops the next `doctor` test from reintroducing it by omission.

### Original PR Review

Decision: approve.

Specific gaps fixed:

- **The guard registration was pinned by nothing.** Deleting the `t.Cleanup` registration inside `guardedTempDir` — exactly the pre-`ga-531fk` shape — left the whole package green, so the PR's own load-bearing claim ("the guard now comes with the directory") was undefended by any test. The final patch adds a `guardedTempDirWith(t, remove)` seam and `TestGuardedTempDirRegistersTheRetryingRemoval`, which drives a fail-twice-then-succeed remove and asserts three calls after the subtest returns. The deletion mutant is now red (`registered remove called 0 times, want 3`), verified by re-running it under `go test -overlay` on the full package.
- **The wiring test pinned the seam but not the wrapper every real caller uses.** All six production call sites go through `guardedTempDir`, while the only test that could observe registration called `guardedTempDirWith` directly — the pinned path and the used path were disjoint, and two doc comments claimed otherwise. Both a wrapper-bypass mutant (`return t.TempDir()`) and an inert-remove mutant survived. The final patch adds `TestGuardedTempDirRemovalRunsBeforeTempDirsOwnCleanup`, which sandwiches a probe cleanup between `TempDir`'s base `RemoveAll` and the guard's retry so the LIFO ordering is directly observable, and corrects both comments to name the right subject. Both mutants are now red.
- **Retry exhaustion was silent.** `retryRemoveAll` now returns the final failure and logs it on give-up with the directory, attempt count, and error — the triage signal `ga-531fk` itself lacked. Fall-through stays non-fatal, so `TempDir`'s own cleanup keeps the last word and the failure posture is unchanged.
- **The retry slept after its final failed attempt.** Now gated on `i < attempts-1`, with the doc comment updated to state the new contract; real-caller cadence is preserved at 10 × 50ms.
- **Added the cross-reference to `ga-7dgcg6`**, whose avoidance-by-placement shape in `cmd/gc/testenv_test.go` (`doltIdentityHomeDir`) solves the same problem a different way — useful context for whoever touches this next.

Review findings addressed:

- *Test Evidence Quality / major* — guard registration killed by no test. Fixed in `6d5c3e3` (`fix(test): pin the guard registration in guardedTempDir (review fixup)`).
- *Test Evidence Quality / minor* — wrapper unpinned by any test, with two comments attributing coverage it did not have. Fixed in the same commit `6d5c3e3`.
- *Error handling and resource lifecycle / nits* — silent retry exhaustion and the trailing sleep. Fixed in `6d5c3e3`.

- `ga-jnnd3` (*Change Impact / minor*): six `t.TempDir()`-as-`HOME` sites outside package `doctor` still carry the unguarded shape, plus `internal/beads/boundary_test.go`, which pins no `HOME` at all. A `_test.go` helper cannot cross a package boundary, so converging the three divergent idioms needs a shared test-helper package. This is pre-existing and outside this PR's causal scope, so it was deliberately left alone here and filed instead — `ga-jnnd3` is open at P2 and carries the corrected six-site census.

One residual worth knowing about (not a finding, and not actionable in this PR): the 10 × 50ms budget is empirical. If the embedded-dolt writer lingers past roughly 500ms the same red can return — the guard narrows the race window rather than eliminating it, and `TempDir`'s single-shot cleanup keeps the last word by design. Neither the review runs nor CI exercise the Mac timing window, so green here is evidence of no regression rather than proof the flake is closed for good.

### Maintainer Changes

One maintainer fixup commit was pushed to the PR branch; your authorship on `399bc99` is preserved.

- `6d5c3e3` — `fix(test): pin the guard registration in guardedTempDir (review fixup)` — `internal/doctor/tempdir_guard_test.go`, 97 insertions(+), 15 deletions(-)

Final adopted diff: 2 files changed, 210 insertions(+), 27 deletions(-).

### Final Review Status

Ready for the merge queue: final head `6d5c3e3` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/34195344405

### Review Iterations

2 review passes, three reviewers each with full cross-file coverage. Iteration 1 returned request_changes on the unpinned guard registration and specified the seam plus the mutant-killing test; the maintainer fixup landed that remedy. Iteration 2 re-ran the deletion mutant independently in every lane rather than trusting the fixup's own claim, and approved — with one strictly smaller residual in the same category that was landable without another review pass. Its test and both comment corrections are in the final head.

---

PR #6140, opened 2026-09-08
fix(residency): retire the hand-rolled by-id probes onto the resolver (s6)

julianknutsen:
**Segment 6 of 8 — re-split out of #5644, whose head had silently absorbed this
segment and s7 as squash merges.** Rebased onto `main`: #5644 (segment 5) landed
as squash `c6bb2aa30d`, #5626 (pr1) as `6f69447ce2`, #5600 as `abe0a0c6b9`. s7
(the residency-guard file split) is NOT here; it follows as its own PR.

See segment 1 for the bug class and the full segment table.

## The mechanism

`storeref.Resolve(id, stores)` tries `PrefixOwner` FIRST — the cheap fork-free
path — and `PrefixOwner` matches on the id's OWN prefix, ahead of whatever order
the caller assembled its list in. `gc storage migrate` COPIES AND RETAINS
(`infra_class_migrate.go:1006`) and it PRESERVES ids, so an infrastructure bead
minted before a cutover keeps its work-era prefix and exists twice: a frozen row
in the work ledger and the live row in the binding.

Net: every `storeref.Resolve` caller read the frozen pre-migration copy for every
migration-preserved infra id. Successfully, with no error to notice, with status
and revision as of the cutover. That is `ga-cu12x`, and its four live callers on
`main` — `cmd/gc/cmd_wait.go:67` and `internal/dispatch/drain.go:428/804/1335` —
are exactly the probes this segment retires. `ga-cu12x` closes here.

Each of the four had also grown its own answer to "which store owns this id", and
each was wrong in a different way besides.

## The retired sites

| site | was | now |
| --- | --- | --- |
| `cmd/gc/claim_class_route.go` `holds` | unconditional `graph.Get`, a hand-rolled not-found arm, a hand-rolled standing-refusal predicate keyed on the id's prefix | `storeref.ByID` over a frame captured at construction, executed by `storeref.ResolveBindingOwner` |
| `cmd/gc/cmd_wait.go` `newWaitDependencyStoreSet` | `storeref.Resolve` over `[binding, city, rigs sorted]` | `storeref.ByID` over the topology the controller registered, narrowed to the serving rigs |
| `cmd/gc/cmd_wait.go` `loadWaitDependencyBead` | #5488's `classBindingForID` probe, then a store-directory scan | `cliByIDBindingOwner` + `beadForOwner`, then the scan |
| `internal/dispatch/drain.go` `drainMemberProbeSet` | one list consumed two ways — hand-walked to RESERVE, `storeref.Resolve`d to SERVE | `resolveDrainMember` (`ByID`), one answer for both |
| `internal/dispatch/drain.go` `drainUnitConvoyProbeSet` | the sibling's list in reverse, handed to `storeref.Resolve` | `Plan(RoutedWork)` — work-led, which is correct for a self-minted `ClassWork` convoy |
| `internal/dispatch/drain.go` `classifyDrainBlocker` | `storeref.Resolve` over the member-store tail | `resolveDrainMember` |

The blocker site has no behaviour delta at the one member-store shape any
production caller constructs; its value is that "which store holds this id" now
has ONE spelling in `drain.go` instead of three, and that it was the last
`storeref.Resolve` caller in the tree. The ambient store is still asked first and
by hand there, because that read is the CO-RESIDENCY question — can this store's
dep table reference this id — which is a different question from residency.

`drainResidency` frames exactly ONE binding: the drain's own ambient store,
carrying every infrastructure class, censused by nobody and therefore
pessimistic. **More than one member store is REFUSED, loudly, rather than
walked.** That is a contract statement, not a TODO: a drain resolves inside the
two stores its caller already routed, and a frame that quietly grew a second
member leg would let a drain resolve a member out of another scope. Single-store
callers get a one-leg topology and the identity fast path, so they pay nothing.

With no production caller left, `storeref.Resolve` is documented as **superseded**
rather than deleted: `internal/storeref`'s own tests still compare the plan's
contract against it, and deleting it would delete that comparison. `cmd/gc`'s
by-id conformance row (I7) moves onto the derivation production actually runs
(`residencyBindingsFromRoutes` → `byIDBeadForTopology`), so the guard guards the
shipped path. `classBindingForID`, whose last caller was the one-shot wait
loader, is deleted.

`byIDBeadForTopology` is extracted from the closure #5928 put inside
`newWarmClaimTriggerResolver`: plan `ByID`, `ResolveOwnerRow`, `beadForOwner`.
Both controller-side by-id readers go through it, so there is one spelling and
not three.

Two supporting changes carry as-is:

- `internal/beadmeta/reserved_prefixes.go` moves the reserved-prefix table to a
  stdlib-only leaf. It lived in `internal/config`, which meant every consumer —
  `internal/coordclass` and `storeref` included — imported the config package to
  read a static table.
- `internal/storeref/relic_census_bench_test.go` measures what the boot census
  costs (7.2ms at 1k rows, 82ms at 10k, linear).

## Reconciliation with #5626 (pr1), which landed first

pr1 introduced `storeref.ResolveBindingOwner` — the `ModeFirstOwner` executor
that walks the binding legs and stops at the caller's own work axis BY ROLE —
and rewrote `cmd/gc/by_id_residency.go` around it. This branch was written
against a type-assertion form of the same question. **pr1's is the one that
survives; there is no residual type check on the work leg anywhere in the tree.**

| | before this rebase | after |
| --- | --- | --- |
| `byIDBindingOwnerForTopology` | `byIDOwnerForTopology` + `owner.Store.(unprobedWorkResidual)` | `byIDPlanForTopology` + `storeref.ResolveBindingOwner` + `withProvenRelicRemedy` (pr1's body, verbatim) |
| `cliByIDBindingOwner` / `cliByIDOwner` / `cliByIDPlan` topology | `residencyTopologyForCity` | `byIDResidencyTopology` (pr1's, which adds the refused-city relic proof) |
| `hookClaimClassRoute.holds` | the type-assertion form, via `byIDBindingOwnerForTopology` | unchanged call, now reaching `ResolveBindingOwner` |
| `soleBindingResidency` | `residencyBindingsFor(order, byStore, relics)` | `storeref.BuildBinding(store, classes, opts)` (see below) |

The assertion was not merely a duplicate spelling, it was the weaker one: it can
only see the store the walk RETURNED, so it cannot decline the shadow legs
ordered behind the residual, and `dedupeLegs` removes the residual outright on a
city whose binding resolved back to the work store. `ResolveBindingOwner` stops
on `Role.ownedByCallersWorkAxis()`, which covers both. Everything this branch
duplicated of pr1's — the `byIDOwnerForTopology` / `byIDPlanForTopology` /
`byIDBindingOwnerForTopology` split, which both PRs happened to introduce — is
deleted in favour of pr1's copy.

pr1's proven-relic refusal semantics are intact and untouched:
`ClassBinding.KnownLegacyResidents`, `ClassBinding.probeRefusalPolicy`,
`storeref.ProvenLegacyResidents`, `storeref.ErrProvenRelicRefusal`,
`residencyBindingsFromRoutesWithProof` and the whole of
`cmd/gc/by_id_relic_proof.go`. Two adjustments follow from them rather than
change them:

- `soleBindingResidency` and `drainResidency` both pass the zero
  `storeref.BindingOptions`, and their docs now say why its two nil evidence
  funcs mean OPPOSITE things: no census keeps the probe
  (`HasLegacyResidents=true`), no proof never denies a read
  (`KnownLegacyResidents=false`). A caller that took neither is entitled to
  exactly that pair.
- `byIDBeadForTopology` — this branch's new controller-side door — now shares
  `byIDOwnerRowForTopology` with `byIDOwnerForTopology`, so it inherits
  `withProvenRelicRemedy`. pr1 attached that remedy at the two doors it touched;
  a third door reaching the same denial without it would print the boot gate's
  own sentence and send the operator looking for a missing bead. The two
  whole-plan readers now differ only in what a clean miss means to them.

## Two changes from what #5645 proposed

**1. The wait reader is binding-FIRST, not binding-last.** #5645 replaced
#5488's binding-first store set with `Plan(RoutedWork)` + `Walk`, which puts the
binding LAST. That contradicts the residency doctrine this lane established since
(`ga-oytw9`/#5918, #5928): the binding LEADS for a class-resident id. The reader
here plans `ByID` instead, which is binding-first for an id inside a relocated
class's reserved namespace (#5488's property, now enforced by the resolver rather
than by a list order) AND puts every unretired binding's residence probe ahead of
the work ledger for an id inside none — which is the retained-copy case #5488's
ordering could not close, because `PrefixOwner` ran ahead of the list.

#5488's two rows are ported, not dropped:
`TestPrepareWaitWakeState_ResolvesGraphBindingDependencyBeads` now builds its
binding leg through `soleBindingResidency` and reads through the plan;
`TestLoadWaitDependencyBeadReadsTheBindingOnAMigratedCity` is unchanged, because
the loader it covers is the same loader.

`errWaitDependencyUnproven` is kept for "no leg could answer": absence over a
frame narrowed by a suspended rig is not proof, and failing a wait is terminal.
One consequence of `ByID` differs from #5645 and is called out rather than
hidden: the by-id plan makes every leg fatal, so a dark SERVING rig still aborts
the wake pass with its own error instead of degrading. The suspended-rig freeze
#5645's subject names is closed by narrowing the FRAME (`servingRigStores`) so a
suspended rig is out of the plan rather than dark inside it — which is the fix —
and no per-leg failure policy is restated in `cmd_wait.go`. The table row #5645
flipped is therefore renamed and kept asserting the fault, with a comment
pointing at the row that covers the freeze.

**2. No status file.** #5645 introduced `.gc/storage-relic-census.json` and a
reader/writer pair to memoize the census across processes. AGENTS.md: *"No status
files — query live state."* The monotonicity argument is sound in the abstract,
but the artefact goes stale the moment an operator rebuilds or re-points a
binding, nothing clears it, and a later `gc` then keeps or retires a probe on a
verdict no store agrees with. The memo file, its writer, its reader and its four
tests are dropped.

The bound the memo was reaching for already holds in-process and by construction:
`cliStorageRoutes` takes the census inside a `sync.Once` per city and the
controller takes it once at boot, so the verdict lives in the routes value both
callers already hold. That is now documented on `censusBindingRelics`, measured
by the bench (kept as the cost regression guard), and pinned by
`TestBootCensusIsLiveAndLeavesNothingOnDisk`. The real residual is a cheaper scan
rather than a cache — it needs a store-level id-namespace predicate
`beads.ListQuery` does not have — and `ga-dx4ho` carries it.

## RED → GREEN

Each retired probe has a split row that fails on the old reader, a fault control,
and a single-store control.

New in this re-split, and the row `ga-cu12x` asks for:

- `TestPrepareWaitWakeState_MigrationPreservedDependencyAnswersFromTheBinding` —
  the binding holds the live CLOSED dependency under a work-era id, the work
  ledger holds the frozen OPEN twin under its own declared prefix. The wait must
  ready. RED against the old reader: `readyWaitSet[...] = false`.
- `TestWaitDependencyPlanReaderBindingFaultIsAnErrorNeverAbsence` — a binding
  that cannot be read must arrive as that error, never as `beads.ErrNotFound`
  (which reaps the waiter) and never downgraded to unproven (which only logs).
  RED against the old reader: `error = <nil>`, the frozen copy answered.
- `TestWaitDependencyPlanReaderSingleStoreCityIsByteIdentical` — a city with
  nothing relocated gets the store's own row back and its own `ErrNotFound`.
- `TestClassifyDrainBlockerFaultIsAnErrorNeverAnOmittedEdge` — the blocker
  classification's "no" silently DROPS a dependency edge, so a leg that could not
  be read must never decide one.
- `TestBootCensusIsLiveAndLeavesNothingOnDisk` — a second independent open of the
  same city reaches the verdict from the store alone, and no census artefact is
  written.

Carried from #5645: `TestClaimRouteHoldsSkipsTheProbeOnACensusCleanBinding`,
`TestClaimRouteHoldsKeepsTheAuthorityLegOnACensusCleanBinding`,
`TestPrepareWaitWakeState_SuspendedFrameRetainsAnUnprovedWait`,
`TestPrepareWaitWakeState_ResolvesBindingResidentDependency`,
`TestPrepareWaitWakeState_CoResidentDependencyAnswersFromTheWorkStore`,
`TestPrepareWaitWakeState_DarkCityWorkLegStillAbortsThePass`,
`TestDepsWaitReadyUnprovenDependencyNeitherFailsNorReadies`,
`TestManifestLoadAndReservationAgreeOnACoResidentMember`,
`TestCoResidentDrainMemberAnswersFromTheBinding`,
`TestWaitDependencyReadServesTheBindingCopy`,
`TestWaitReadiesOnADependencyTheMigrationRelocated`.

pr1's own rows are unchanged and still green after the reconciliation, including
`TestResolveBindingOwnerNeverProbesTheWorkLeg` (the zero-probe pin the
type-assertion form could not make) and the `T3k` conformance rows that turn
`ByID(ga-xyz)` and `ByID(ra-7)` Fatal on a proven-relic binding.

## Mutation probes

Each applied and reverted.

Reverting the wait reader to `storeref.Resolve` over `[bindings, work, rigs]`:

```
--- FAIL: TestPrepareWaitWakeState_MigrationPreservedDependencyAnswersFromTheBinding
    readyWaitSet[gcg-session-1] = false, want true; the wait read the work
    ledger's frozen pre-migration copy instead of the binding's live row
--- FAIL: TestWaitDependencyPlanReaderBindingFaultIsAnErrorNeverAbsence
    Get(hq-dep-1) error = <nil>, want binding store unavailable
--- FAIL: TestPrepareWaitWakeState_CoResidentDependencyAnswersFromTheWorkStore
    readyWaitSet[gcg-session-1] = false, want true
```

The single-store control and the reserved-prefix row stay GREEN — the two that
were already right for a different reason.

Reverting `loadDrainManifestMembers` to the PrefixOwner-first probe set:

```
--- FAIL: TestManifestLoadAndReservationAgreeOnACoResidentMember
    manifest member reservation = "", want "gcb-1001"
--- FAIL: TestCoResidentDrainMemberAnswersFromTheBinding
    manifest served "work copy", want the binding copy
```

Removing the one-shot loader's binding leg entirely (the pre-#5488 scan) fails
all three of its rows, and fails them differently — the reserved-prefix row
reports "bead not found", the retained-copy rows report the frozen work copy
served open. Writing the census note by hand fails
`TestBootCensusIsLiveAndLeavesNothingOnDisk`.

## Baseline

Shrink-only: **four rows retired, none added.** The full baseline was
regenerated by the documented three-emitter splice
(`check-residency-boundary.sh --emit-baseline`, plus
`RESIDENCY_BASELINE_EMIT=1 go test ./scripts -run TestResidencyResolverBoundary`
and `-run TestResidencyASTExpressionRatchet`, concatenated and `LC_ALL=C sort`ed)
and compared byte-for-byte against the file below; the splice reproduces it
exactly, so nothing was re-keyed by the rebase.

```
-cmd/gc/claim_class_route.go	holds	d:bdIDIsClassReserved	1
-cmd/gc/cmd_wait.go	newWaitDependencyStoreSet	ast:returns-store-list	1
-internal/dispatch/drain.go	drainMemberProbeSet	ast:returns-store-list	1
-internal/dispatch/drain.go	drainUnitConvoyProbeSet	ast:returns-store-list	1
```

- `claim_class_route.go holds d:bdIDIsClassReserved` — the vocabulary that gave
  it away: deciding a refusal's meaning from an id's prefix is the resolver's
  per-leg policy restated by hand. The route still appears on its
  `b:cliSoleClassBinding` row, which is a different claim and stays.
- `cmd_wait.go newWaitDependencyStoreSet ast:returns-store-list` — the
  constructor is gone. `cmd_wait.go` still appears on the
  `loadWaitDependencyBead a:convoyStoreCandidates` row, which is the store-scan
  half of the one-shot loader and a different slice.
- the two `internal/dispatch/drain.go ast:returns-store-list` rows — both probe
  sets are gone. The prose above them is rewritten: their opposite leg orders
  turned out to be correct rather than the defect the row text claimed, and the
  real defect was one list consumed two ways.

`cmd/gc/residency_topology.go residencyBindingsFromRoutesWithProof
b:routes.storeFor` is pr1's re-key of `residencyBindingsFromRoutes`'s row and is
inherited from `main` unchanged.

**No `// residency:allow` marker is added by this PR.** The two new frames each
needed a one-store leg list — `soleBindingResidency` in `cmd/gc` and
`drainResidency` in `internal/dispatch` — and each was writing out the
one-element `order` slice and the one-entry `byStore` map by hand, in packages
that do not import each other, purely to hand them straight back to
`storeref.BuildBindings`. That is a store list assembled outside the resolver,
which is exactly what the guard is for. Both now call `storeref.BuildBinding`,
the one-store shape, so the literal lives in the package that owns leg order,
the two callers cannot drift over what a one-store grouping looks like, and the
guard has nothing to suppress. The header prose is updated for each retirement
so the file still explains every row it keeps. `scripts/` is otherwise untouched
— the guard's own file split is s7.

## Gates

Run on `32e1823ed8d389ee841eb239c5fd4a2aac936dc2`, rebased onto `main` at `c6bb2aa30d`.

| gate | exit | result |
| --- | --- | --- |
| `go build ./...` | 0 | clean |
| `go vet ./...` | 0 | clean |
| `gofmt -l cmd internal` | 0 | nothing printed |
| `gofmt -l cmd internal scripts` | 0 | nothing printed |
| `golangci-lint run ./cmd/gc/... ./internal/dispatch/... ./internal/storeref/... ./internal/beadmeta/...` | 0 | `0 issues.` |
| `./scripts/check-residency-boundary.sh` | 0 | clean |
| `./scripts/check-residency-boundary.sh --self-test` | 0 | clean |
| `go test ./internal/storeref/... ./internal/dispatch/... ./internal/beadmeta/... -count=1` | 0 | ok |
| `go test ./internal/config/... ./internal/coordclass/... -count=1` | 0 | ok |
| `go test ./cmd/gc -run 'Wait\|Drain\|Claim\|ByID\|Residency\|Relic\|Census\|Boot' -count=1` | 0 | ok |
| `go test ./scripts/... -count=1` | 0 | ok |
| `make test-fast-parallel` | 0 | `All fast jobs passed` (run by the pre-push hook on this exact head; see the note below) |

`make test-fast-parallel` is the suite `.githooks/pre-push` runs, and the row
above is that hook's run on this exact head — the push would not have completed
otherwise. It is recorded that way deliberately rather than run twice: this host
caps concurrent heavy suites at two slots
(`scripts/push-gate-lock-lib.sh`, `ga-owh20p`), and an earlier standalone
invocation exited **75 (EX_TEMPFAIL)** after waiting out the 600 s bound with
both slots held by unrelated runs. That exit code is the gate reporting host
contention, not a test result, and the documented response is to rerun rather
than to set `GC_PUSH_GATE_NO_CAP=1` — bypassing it is what produced the load
incidents the gate exists to prevent.

`.githooks/pre-commit` is active (`git config core.hooksPath` prints
`.githooks`) and ran on every commit here; the branch was pushed with the
pre-push hook enabled, no `--no-verify`. No base-owned failure was reproduced:
every gate above is green on this head.

Refs `ga-cu12x` (closes), `ga-j4p3y`, `ga-qdt5y.16`.

---

# Update — rebased onto `main`, and the prefix-fault row is fixed

Head is now `e8bf6de219ddc77a13d3d10f2e0231c61b2a4bd6`. The branch was **rebased**, not merged: the
`Merge remote-tracking branch 'origin/main'` commit (`49927f5e7e`) is gone, the
seven s6 commits and the maintainer's `chore(review)` commit (`e55acb3c8f`)
replay linearly onto `main`, and one new commit carries the fix below.

## The rebase

`git rebase --onto origin/main c6bb2aa30d` over the four merges that landed since
the branch was cut — #6139, #6138 (`workrecord.RepoDirFor` +
`workRecordRepoDirs`), #6142 (`ProvenLegacyResidents` on the closed-inclusive
census) and #5602 (`internal/beads/namespace_census.go` +
`storeref.legacyResidentVerdict`). `f3610ae8e6` was dropped as already upstream.

**One conflict**, `internal/storeref/relic_census_bench_test.go` (add/add).
`main`'s version wins whole: #5602 made the census answerable by a store-level
predicate (`beads.NamespaceCensus`), so `main`'s file benchmarks the predicate
and keeps the listing form as `BenchmarkRelicCensusScanFallback_10k`. This
branch's version measured only the listing form, which is a strict subset.

Two prose blocks this branch had written about that cost were left stale by that
resolution and are now corrected against a fresh measurement rather than
deleted — `censusBindingRelics` in `cmd/gc/storage_boot.go` and
`TestBootCensusIsLiveAndLeavesNothingOnDisk` in
`cmd/gc/relic_census_boot_test.go`. Both had cited the pre-#5602 numbers
(7.2 ms/1k, 82 ms/10k) and closed with "bounding it needs a store-level
id-namespace predicate `beads.ListQuery` does not have (`ga-dx4ho`)", which
#5602 landed. Measured on this head:

| form | 1k rows | 10k rows |
| --- | --- | --- |
| `beads.NamespaceCensus` predicate | 0.40 ms | 3.2 ms |
| listing fallback (native-Dolt binding) | — | 97 ms |

The argument the blocks exist to make is unchanged and now says the true thing:
the read is still paid per process rather than written to disk, and the saving
came from making the read cheaper, not from remembering it.

Everything else replayed clean, including
`scripts/residency-boundary-baseline.txt`, `internal/storeref/bindings.go`,
`cmd/gc/by_id_residency.go` and `internal/dispatch/drain.go`. The baseline is
unchanged from the version reviewed at `32e1823ed8`: shrink-only, **4 rows
retired, 0 added**, and `./scripts/check-residency-boundary.sh` exits 0 without
regeneration — no row re-keyed, so no 3-emitter splice was needed.

## The fix: a rig that contradicts its own config is a FAULT, not an absence

`e55acb3c8f` added
`TestPrepareWaitWakeState_PrefixUncoveredRigDependencyDoesNotReapTheWaiter`,
red. It builds a rig whose **configured** prefix (`"fe"`) disagrees with the
prefix its **store** declares (`"ga"`), and asserts that a wait on `ga-dep-1`
is retained rather than reaped.

The row is kept byte-for-byte. What it caught:

- the by-id plan gates rig legs on the CONFIGURED prefix (`shadowLegsCovering`),
  so the `"fe"` rig is out of the plan for `ga-dep-1` even though its store holds
  that bead;
- the plan therefore misses everywhere and returns `beads.ErrNotFound`;
- `waitDependencyPlanReader.Get` passed that through as a PROVED absence, and
  `prepareWaitWakeStateWithSnapshot` FAILS a wait on a proved absence — so a
  living dependency reaped its waiter.

A disagreement between what a rig's config says and what its store says is a
fault in the FRAME, and residency doctrine already holds that a fault is never
absence.

### Where the disagreement is detected

`storeref.Topology.PrefixFaults()` — read off the same captured frame the wait
reader already holds, in the package that owns both `Leg.Prefix` and the
`HasIDPrefix` declaration, so the fault and the gate that creates it are stated
once. It returns `[]PrefixFault{Ref, Configured, Declared}` in the plan's own
rig order, and `PrefixFault.String()` is the sentence a consumer logs.

`cmd/gc/cmd_wait.go` consumes it in `waitFrameGap`, which replaces the reader's
`narrowed bool` with an `outOfFrame string` naming what the frame leaves out.
The two causes now sit side by side and are handled identically: a suspended rig
is TOLD to the constructor (the caller already dropped it from the serving set),
a prefix fault is READ off the frame. Either one makes a miss
`errWaitDependencyUnproven`, so the wait pends and the next tick tries again.

### Why this stays bounded

**Gating rig legs on the configured prefix is not the bug and is not changed.**
On any multi-rig city most legs are gated out of most by-id plans; that is what
makes absence PROVABLE, and it is the only reason a wait on a genuinely deleted
dependency can ever fail. Three shapes are therefore explicitly *not* faults,
each pinned in `internal/storeref/topology_prefix_fault_test.go`:

| shape | why it is not a fault |
| --- | --- |
| the two prefixes AGREE | the gate is exact; a leg covering none of the id's namespace holds nothing to say about it |
| the store declares NOTHING (`""` or no `IDPrefix()`) | no claim to contradict — a bd/Dolt work store with no configured mint prefix, i.e. most cities |
| the rig leg IS the work store | `dedupeLegs` collapses it onto the work leg, which every by-id plan reads unconditionally |

Nothing outside the wait reader consumes `PrefixFaults()`, so no plan, drain,
claim route or census changes behaviour.

False positives are structurally unlikely for the same reason the fault is worth
reporting: on every production path a rig's store is OPENED with the configured
prefix — `controllerState.openRigStore(..., rig.EffectivePrefix(), ...)` feeds it
straight into the exec store's `GC_BEADS_PREFIX`, and
`work_query_probe.go` projects the same value — so the two agree by
construction. A city where they do not has an inconsistency an operator needs to
see, and the log line names the rig and both prefixes.

## The two rows, and the mutation that separates them

| row | before the fix | after the fix | under the mutation |
| --- | --- | --- | --- |
| `TestPrepareWaitWakeState_PrefixUncoveredRigDependencyDoesNotReapTheWaiter` (prefixes DISAGREE, dependency lives in the gated-out rig) | **RED** — `wait state = "failed", want "pending"` | **GREEN** | GREEN |
| `TestPrepareWaitWakeState_AgreeingPrefixRigStillProvesTheDependencyGone` (prefixes AGREE, leg still gated out, dependency absent everywhere) | **GREEN** | **GREEN** | **RED** — `wait state = "pending", want "failed"; a gated-out leg whose two prefixes agree must leave absence provable` |

The control is the same shape as the red row with exactly one thing changed —
the rig's configured prefix IS what its store declares — and it asserts the
opposite outcome: the wait is FAILED and CLOSED with `last_error` recorded,
because a dependency in no leg of an honest plan is gone.

**The mutation** is the cheap wrong fix the verdict forbids: replace the
`PrefixFaults()` loop in `waitFrameGap` with one that reports every rig leg,
i.e. treat every prefix-gated-out leg as out of frame.

```go
	for _, leg := range topo.Rigs {
		gaps = append(gaps, string(leg.Ref)+" is gated out of frame")
	}
```

That turns the control RED and leaves the new row GREEN, which is the whole
point of having both: it makes absence unprovable on any multi-rig city, so a
wait on a genuinely deleted dependency would pend forever — strictly worse than
the freeze being fixed. The mutation was applied, observed, and reverted; it is
not in the branch.

`internal/storeref/topology_prefix_fault_test.go` pins the predicate itself: the
positive row (both prefixes named in the report), five silent rows covering the
three shapes above (an agreeing prefix; a store declaring `""`, declaring no
`IDPrefix()` at all, or absent entirely; and a rig leg that IS the work store),
and rig-ref ordering.

## Gates — all on `e8bf6de219ddc77a13d3d10f2e0231c61b2a4bd6`

| gate | exit | result |
| --- | --- | --- |
| `go build ./...` | 0 | clean |
| `go vet ./...` | 0 | clean |
| `gofmt -l cmd internal` | 0 | nothing printed |
| `golangci-lint run ./cmd/gc/... ./internal/dispatch/... ./internal/storeref/... ./internal/beadmeta/...` | 0 | `0 issues.` |
| `./scripts/check-residency-boundary.sh` | 0 | clean, no regeneration |
| `./scripts/check-residency-boundary.sh --self-test` | 0 | clean |
| `go test ./internal/storeref/... ./internal/dispatch/... ./internal/beadmeta/... -count=1` | 0 | ok |
| `go test ./cmd/gc -run 'Wait\|Drain\|Claim\|ByID\|Residency\|Relic\|Census\|Boot' -count=1` | 0 | ok (44.6 s) |
| `go test ./scripts/... -count=1` | 0 | ok |
| `make test-fast-parallel` | 0 | `All fast jobs passed` — run by `.githooks/pre-push` on this exact head; the push would not have completed otherwise |

`make test-fast-parallel` is recorded as the pre-push hook's run rather than as a
second standalone invocation, for the reason the section above already gives:
this host caps concurrent heavy suites at two slots
(`scripts/push-gate-lock-lib.sh`, `ga-owh20p`), and the documented response to
contention is to wait rather than to set `GC_PUSH_GATE_NO_CAP=1`, which was left
untouched. `.githooks/pre-commit` is active and ran on the new commit; the push
used `--force-with-lease`, no `--no-verify`.

### One base-owned flake, disclosed and NOT worked around

A standalone `make test-fast-parallel` run on this head failed once in
`internal/storebinding/sqlite`:

```
--- FAIL: TestSQLiteWriterFenceSIGKILLAtReservationBoundaries/WAL_without_SHM/reservation-lock-pending
    sqlite_fence_process_linux_test.go:324: removing crash-retained SQLite SHM:
    remove .../graph/beads.sqlite-shm: no such file or directory
```

It is base-owned by construction: `git diff origin/main HEAD --
internal/storebinding internal/beads` is EMPTY, so that package is byte-identical
to `main` on this branch. The row kills a child process and requires SQLite to
have left a `-shm` behind; under a loaded 10-job harness the file is
occasionally already gone, which is a timing race in the fixture and not an
assertion about the code under test. It did not reproduce in 8 consecutive runs
on an untouched `origin/main` export (`git archive origin/main`), nor on re-run
here, nor in the pre-push hook's run on this head — the run recorded above
printed `All fast jobs passed` with that row green. **No workaround, skip, retry
or quarantine was added for it**: the recorded green is a clean run, not a run
with the row disabled.

Refs `ga-cu12x` (closes), `ga-j4p3y`, `ga-qdt5y.16`.

---
- Related to #5987 — covers the same copy-and-retain co-residency on the read side — the wait-dependency, drain-member and claim-route probes now resolve a migration-preserved id to the binding's live row instead of the frozen pre-migration copy in the work ledger; it does not converge by-id writes onto the retained work row, tombstone it, or change the work-first precedence on the demand/claim surfaces, so the orphan-reopen and re-dispatch loop that issue measures is not addressed here

### DIFF
--- a/cmd/gc/by_id_residency.go
+++ b/cmd/gc/by_id_residency.go
@@ -139,6 +139,32 @@ func byIDPlanForTopology(topo storeref.Topology, id string) (storeref.ResolvedPl
 	return storeref.Plan(storeref.ByID{ID: id}, topo)
 }
 
+// byIDBeadForTopology reads the row id names over a topology the caller already
+// holds: plan ByID, resolve the owner, and read only if the winning probe did
+// not already.
+//
+// It is the ONE controller-side spelling of the by-id read. The warm-bind claim
+// probe (newWarmClaimTriggerResolver) and the wait-dependency reader
+// (waitDependencyPlanReader) both go through it, so a third controller surface
+// cannot grow a fourth answer to "which store holds this bead" — the drift this
+// lane exists to close.
+//
+// Unlike byIDOwnerForTopology there is no work-store fallback argument: a
+// controller-side caller holds no separate work axis to fall back TO. Its work
+// store is already a leg of the topology, so a clean miss on every leg is
+// beads.ErrNotFound and the caller decides what that means.
+func byIDBeadForTopology(topo storeref.Topology, id string) (beads.Bead, error) {
+	plan, err := byIDPlanForTopology(topo, id)
+	if err != nil {
+		return beads.Bead{}, err
+	}
+	owner, err := storeref.ResolveOwnerRow(plan, id)
+	if err != nil {
+		return beads.Bead{}, err
+	}
+	return beadForOwner(owner, id)
+}
+
 // cliByIDBindingOwner answers the binding half of the by-id question for a
 // surface whose work axis is not a beads.Store.
 //
--- a/cmd/gc/by_id_residency.go
+++ b/cmd/gc/by_id_residency.go
@@ -85,11 +85,7 @@ func cliByIDOwner(cityPath, id string, work beads.Store) (storeref.Owner, error)
 // for it would be this file picking a store out of a leg — the residency
 // boundary's one forbidden move, and pointless when the caller already has it.
 func byIDOwnerForTopology(topo storeref.Topology, id string, work beads.Store) (storeref.Owner, error) {
-	plan, err := byIDPlanForTopology(topo, id)
-	if err != nil {
-		return storeref.Owner{}, err
-	}
-	owner, err := withProvenRelicRemedy(storeref.ResolveOwnerRow(plan, id))
+	owner, err := byIDOwnerRowForTopology(topo, id)
 	switch {
 	case err == nil:
 		return owner, nil
@@ -139,6 +135,24 @@ func byIDPlanForTopology(topo storeref.Topology, id string) (storeref.ResolvedPl
 	return storeref.Plan(storeref.ByID{ID: id}, topo)
 }
 
+// byIDOwnerRowForTopology plans ByID, runs the owner-row executor and names the
+// remedy on the one denial that needs one.
+//
+// It is the step the two whole-plan readers below and above it share, and they
+// differ only in what a clean miss means to them: byIDOwnerForTopology hands
+// back the caller's own work store, byIDBeadForTopology has none to hand back.
+// Sharing it is what keeps withProvenRelicRemedy on both — an operator who hit
+// the proven-relic denial through the wait reader needs the same next move as
+// one who hit it through `gc bd`, and a remedy attached at two of three doors
+// is the drift this file exists to close.
+func byIDOwnerRowForTopology(topo storeref.Topology, id string) (storeref.Owner, error) {
+	plan, err := byIDPlanForTopology(topo, id)
+	if err != nil {
+		return storeref.Owner{}, err
+	}
+	return withProvenRelicRemedy(storeref.ResolveOwnerRow(plan, id))
+}
+
 // byIDBeadForTopology reads the row id names over a topology the caller already
 // holds: plan ByID, resolve the owner, and read only if the winning probe did
 // not already.
@@ -154,11 +168,7 @@ func byIDPlanForTopology(topo storeref.Topology, id string) (storeref.ResolvedPl
 // store is already a leg of the topology, so a clean miss on every leg is
 // beads.ErrNotFound and the caller decides what that means.
 func byIDBeadForTopology(topo storeref.Topology, id string) (beads.Bead, error) {
-	plan, err := byIDPlanForTopology(topo, id)
-	if err != nil {
-		return beads.Bead{}, err
-	}
-	owner, err := storeref.ResolveOwnerRow(plan, id)
+	owner, err := byIDOwnerRowForTopology(topo, id)
 	if err != nil {
 		return beads.Bead{}, err
 	}
--- a/cmd/gc/by_id_store_route.go
+++ b/cmd/gc/by_id_store_route.go
@@ -24,6 +24,7 @@ import (
 	"strings"
 
 	"github.com/gastownhall/gascity/internal/beads"
+	"github.com/gastownhall/gascity/internal/storeref"
 )
 
 // classRoutedStoreForID returns the store that actually holds id: the relocated
@@ -102,7 +103,7 @@ func classBindingForID(cityPath, id string) (beads.Store, bool, error) {
 // An error is a read that FAILED, never absence: reading "the binding could not
 // answer" as "the bead is not there" is the root-loss shape this lane exists to
 // prevent. The one error that is not a fault is the one-shot funnel's standing
-// refusal (isStandingStorageRefusal) — a verdict about the CITY's storage
+// refusal (storeref.IsStandingRefusal) — a verdict about the CITY's storage
 // configuration that says nothing about a bead, and a refused city still serves
 // WORK from its work ledger. So for a work-shaped id the refusal establishes
 // nothing and work answers; for an id inside a reserved namespace the refusal IS
@@ -115,7 +116,7 @@ func classRoutedStoreForIDIn(class beads.Store, id string, work beads.Store) (be
 		switch {
 		case errors.Is(err, beads.ErrNotFound):
 			return work, nil
-		case isStandingStorageRefusal(err) && !bdIDIsClassReserved(id):
+		case storeref.IsStandingRefusal(err) && !bdIDIsClassReserved(id):
 			return work, nil
 		default:
 			return nil, fmt.Errorf("reading %q from the relocated class binding: %w", id, err)
--- a/cmd/gc/by_id_store_route.go
+++ b/cmd/gc/by_id_store_route.go
@@ -68,29 +68,11 @@ func cityGraphClassBinding(cityPath string) beads.Store {
 	return store
 }
 
-// classBindingForID is the class leg of classRoutedStoreForID, for callers whose
-// work answer is a leg list rather than one store. A nil work leg makes the
-// residual answer nil, so ok=false means "no named leg answered".
-func classBindingForID(cityPath, id string) (beads.Store, bool, error) {
-	class := cityGraphClassBinding(cityPath)
-	if class == nil {
-		return nil, false, nil
-	}
-	store, err := classRoutedStoreForIDIn(class, id, nil)
-	if err != nil {
-		return nil, false, err
-	}
-	if store == nil {
-		return nil, false, nil
-	}
-	return class, true, nil
-}
-
 // classRoutedStoreForIDIn is classRoutedStoreForID with the binding already in
 // hand, for the callers that resolve it once per request and hold no cityPath to
 // hand cliByIDOwner: gc graph's per-id resolver (graphStores.storeFor), which
-// also needs the raw binding for convoycore.MemberClasses, and classBindingForID
-// itself. Nil class means no relocation.
+// also needs the raw binding for convoycore.MemberClasses. Nil class means no
+// relocation.
 //
 // It applies the resolver's rule directly against the opened binding. The class
 // store leads because it MINTS the reserved namespace, but minting is not
--- a/cmd/gc/city_runtime.go
+++ b/cmd/gc/city_runtime.go
@@ -2476,15 +2476,7 @@ func (cr *CityRuntime) stopConfigWatcher() {
 func (cr *CityRuntime) newWarmClaimTriggerResolver(servingRigs map[string]beads.Store) warmClaimTriggerResolver {
 	topo := cr.residencyTopology(servingRigs)
 	return func(triggerID string) (beads.Bead, error) {
-		plan, err := storeref.Plan(storeref.ByID{ID: triggerID}, topo)
-		if err != nil {
-			return beads.Bead{}, err
-		}
-		owner, err := storeref.ResolveOwnerRow(plan, triggerID)
-		if err != nil {
-			return beads.Bead{}, err
-		}
-		return beadForOwner(owner, triggerID)
+		return byIDBeadForTopology(topo, triggerID)
 	}
 }
 
@@ -2654,7 +2646,15 @@ func (cr *CityRuntime) beadReconcileTick(ctx context.Context, result DesiredStat
 	phaseStart = time.Now()
 	cfgNames := configuredSessionNamesWithSnapshot(cr.cfg, cityName, sessionBeads)
 
-	readyWaitSet, err := prepareWaitWakeStateWithSnapshot(sessionpkg.NewStore(sessStore), newWaitDependencyStoreSet(store, rigStores, cr.graphBeadStore()), cr.nudgesBeadStore(), time.Now(), sessionBeads)
+	// The dependency reader plans over the frame this tick is TOLD is serving,
+	// not over every rig store the runtime holds open: a suspended rig is dark,
+	// and reading it made every dependency lookup in the city fail.
+	suspendedRigPaths := buildSuspendedRigPathsForCity(cr.cfg, cr.cityPath)
+	waitDeps := newWaitDependencyPlanReader(
+		cr.residencyTopology(servingRigStores(cr.cfg, rigStores, suspendedRigPaths)),
+		len(suspendedRigPaths) > 0,
+	)
+	readyWaitSet, err := prepareWaitWakeStateWithSnapshot(sessionpkg.NewStore(sessStore), waitDeps, cr.nudgesBeadStore(), time.Now(), sessionBeads)
 	if err != nil {
 		fmt.Fprintf(cr.stderr, "%s: preparing waits: %v\n", cr.logPrefix, err) //nolint:errcheck
 		readyWaitSet = nil
--- a/cmd/gc/claim_class_route.go
+++ b/cmd/gc/claim_class_route.go
@@ -130,6 +130,7 @@ import (
 	"github.com/gastownhall/gascity/internal/events"
 	"github.com/gastownhall/gascity/internal/executionevent"
 	"github.com/gastownhall/gascity/internal/storebinding"
+	"github.com/gastownhall/gascity/internal/storeref"
 )
 
 // errClaimRouteBindingCannotClaim reports that the relocated coordination-class
@@ -161,6 +162,13 @@ type hookClaimClassRoute struct {
 	// takes a beads.Store rather than the closed contract.
 	class beads.Store
 	graph storebinding.GraphStore
+
+	// topology is the residency frame holds() plans over, captured once at
+	// construction. Data, not a path and not a func: a route that re-entered the
+	// funnel per bead could resolve a DIFFERENT handle from the one class and
+	// graph were opened over, and would then prove a bead absent in one engine
+	// while the claim wrote through another.
+	topology storeref.Topology
 
 	resident map[string]bool
 
@@ -199,13 +207,48 @@ func newHookClaimClassRoute(class beads.Store) (*hookClaimClassRoute, error) {
 // newClaimClassRouteOver builds the route value over an already-projected front
 // door, with both per-invocation memos live. It is the one place they are
 // created, so a route can never reach the escalation with a nil map.
+//
+// The topology it derives is PESSIMISTIC: one binding over the given store,
+// serving every infrastructure class, censused by a nil relics func — which
+// storeref.BuildBindings reads as HasLegacyResidents=true for every store,
+// because a caller holding no censused routes is entitled to claim nothing more.
+// That makes probeRetired() false, so a route built here probes on exactly the
+// ids it probed on before this frame existed. Callers that DO hold a census —
+// hookClaimClassRouteForCity, the only production one — overwrite it with the
+// real thing.
+//
+// Deriving it through soleBindingResidency rather than assembling a ClassBinding
+// literal is the point of doing it at all: Prefixes, Classes, Ref and both
+// retirement bits then come from the one production derivation, so a route built
+// from a bare store cannot disagree with a route built from a city about what
+// the binding's namespaces are.
+//
+// class is what the topology's binding leg is, and graph must be the projection
+// OF that same store — newHookClaimClassRoute is what guarantees it. A caller
+// pairing a graph over one store with a topology over another would prove a bead
+// absent in one engine and write through the other, which is the split this
+// route exists to close, one level up.
 func newClaimClassRouteOver(class beads.Store, graph storebinding.GraphStore) *hookClaimClassRoute {
+	bindings, refused := soleBindingResidency(class)
 	return &hookClaimClassRoute{
 		class:    class,
 		graph:    graph,
+		topology: claimRouteTopology(bindings, refused),
 		resident: map[string]bool{},
 		workHeld: map[string]bool{},
 	}
+}
+
+// claimRouteTopology frames the bindings for holds().
+//
+// The work leg is the unprobed residual, which is what makes the plan's answer
+// readable as a yes/no about the BINDING alone: this route's work axis is the
+// federated bd fan-out the caller already ran and found not-found, not a store,
+// and a real work leg here would report those same beads owned by work again.
+// No rig legs, for the reason cliByIDOwner passes none — a claim escalation has
+// never read them, and starting to would change which beads it claims.
+func claimRouteTopology(bindings []storeref.ClassBinding, refused error) storeref.Topology {
+	return assembleResidencyTopology(nil, newUnprobedWorkResidual(), nil, bindings, refused)
 }
 
 // hookClaimRouteVerdict turns one class-route resolution into the claim ops the
@@ -270,7 +313,18 @@ func hookClaimClassRouteForCity(cityPath string) (*hookClaimClassRoute, error) {
 	if !relocated {
 		return nil, nil
 	}
-	return newHookClaimClassRoute(binding.Store)
+	route, err := newHookClaimClassRoute(binding.Store)
+	if err != nil {
+		return nil, err
+	}
+	// The census-bearing frame, read from the SAME memo cliSoleClassBinding just
+	// took the store out of. That is what makes the probe handle and the write
+	// handle one store by construction rather than by coincidence, and it is why
+	// the topology is not resolved from cityPath later: a second derivation could
+	// open a second engine on the same root.
+	bindings, refused := cliResidencyBindings(cityPath)
+	route.topology = claimRouteTopology(bindings, refused)
+	return route, nil
 }
 
 // knownResident reports whether an earlier probe in THIS invocation already
@@ -285,13 +339,23 @@ func (r *hookClaimClassRoute) knownResident(id string) bool {
 
 // holds probes the binding for id and memoizes the answer.
 //
-// An error is a read that FAILED, never absence. The single exception is the
-// one-shot funnel's standing refusal on a WORK-shaped id: it says this city's
-// storage configuration cannot be served, which is a fact about the city and
-// none about a particular bead, and a refused city still serves work from its
-// work ledger — so the caller's own work-store answer stands. An id only the
-// binding could own (a reserved class prefix) has nowhere else to live, so for
-// that one the refusal is the answer and surfaces.
+// The judgement is the residency resolver's, over the frame captured at
+// construction. It used to be spelled here — an unconditional Get, a not-found
+// arm, a hand-rolled standing-refusal arm — and every clause of it was a
+// restatement of one the by-id door already had. Two consequences follow from
+// the collapse, and both are the point:
+//
+// An error is a read that FAILED, never absence, and the one non-fault is the
+// one-shot funnel's standing refusal on a WORK-shaped id. That is now the
+// resolver's per-leg policy rather than a predicate here: a residence probe for
+// an id no relocated class could own tolerates the refusal, because a refused
+// city still serves work from its work ledger, while the authority leg for an id
+// inside a reserved namespace surfaces it, because there the refusal IS the
+// answer.
+//
+// And a binding the boot census certified relic-free is not probed at all for a
+// work-shaped id. The census is taken to retire exactly this read, and this seam
+// was paying it once per escalated bead while the by-id door had already stopped.
 func (r *hookClaimClassRoute) holds(id string) (bool, error) {
 	id = strings.TrimSpace(id)
 	if r == nil || id == "" {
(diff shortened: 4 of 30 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277312171

quad341 (dismissed):
> @julianknutsen — first, an error on our side: a review automation pushed `e55acb3c8` to your branch under your name, carrying a test that fails. It's been reverted; your tip is `49927f5e7` again. Nothing you did.
>
> On the PR itself. This is close, and the reconciliation section is why it was reviewable at all — having your independently-written `byIDOwnerForTopology`/`byIDPlanForTopology`/`byIDBindingOwnerForTopology` split *deleted* in favour of pr1's, with the reasoning for why the type-assertion form was weaker (it can't decline shadow legs behind the residual, and `dedupeLegs` eats the residual on a city whose binding folds back to work), turns a scary rebase into something a reviewer can actually check. The fault controls sold the rest of it: `TestClassifyDrainBlockerFaultIsAnErrorNeverAnOmittedEdge` in particular, because an unread leg reported as `Satisfied` drops an edge on the claim that a blocker nobody saw is terminal, and that is invisible in a green suite. And `prefixDeclaringStore` — noticing that bare `beads.MemStore` carries `IDPrefix` as a *field*, so the entire pre-existing drain suite was prefix-less to `storeref` and the fast path never fired — is a genuinely good catch.
>
> We ran the full suite locally against your head. One test fails in the whole tree, and it's one we wrote. Here it is:
>
> ```
> --- FAIL: TestPrefixUncoveredRigDependencyDoesNotReapTheWaiter
>     cmd_wait_test.go: wait state = "failed", want "pending"
> ```
>
> **The one thing we need from you:** decide what a rig leg the plan *declined to read* means to the wait reader, and make it non-terminal.
>
> `waitDependencyPlanReader` plans `ByID`, so its rig legs come from `shadowLegsCovering`, gated on `IDInNamespace(id, leg.Prefix)` with `Prefix = rig.EffectivePrefix()`. The retired `newWaitDependencyStoreSet` probed every rig store unconditionally. So a dependency that lives in a rig store under an id outside that rig's *configured* prefix now misses cleanly, arrives at `prepareWaitWakeStateWithSnapshot` as `beads.ErrNotFound`, and gets `FailWait` — closed, terminal, no event. `errWaitDependencyUnproven` doesn't catch it because `narrowed` is only true for suspended rigs.
>
> We ran the control: same stores, same seeds, only `config.Rig.Prefix` changed from `"fe"` to `"ga"`. Covering prefix → the wait readies off the rig's closed copy. Uncovered → reaped. Single variable.
>
> On whether it's reachable — we think yes, and cheaply. `Rig.EffectivePrefix()` derives from the rig *name* when no `prefix =` is set, so renaming a rig in `city.toml` repoints the configured prefix away from every id already sitting in that rig's store. No migration required.
>
> We deliberately didn't pick the fix for you. There are two defensible answers — widen the rig legs to cover what the store actually holds, or return `errWaitDependencyUnproven` for an uncovered-but-resident miss, which is what you built the third value for — and they mean different things. The second can't just set `narrowed`, because `TestWaitDependencyPlanReaderSingleStoreCityIsByteIdentical` correctly pins that a complete frame proves absence and a genuinely-deleted dependency must still fail the wait; telling "frame complete" from "frame narrowed by prefix gating" needs the plan to say which legs it declined. That's your contract, in your series, and we'd rather have your answer than our guess baked into the seam every future by-id reader inherits.
>
> Everything else is accepted as-is. The `internal/beadmeta` extraction is the quiet structural win here — `internal/storeref` was importing `internal/config`, whose cone reaches `internal/git`, `internal/remotesource` and `internal/worker/builtin` → `internal/runtime`, to read one static table — and `leaf_test.go`'s AST guard with its `parsed == 0` vacuity check is the right way to hold it. Documenting `storeref.Resolve` as superseded rather than deleting it is also the right call; it's the executor your plan's contract is compared against. We'll take the stale `drainUnitConvoyProbeSet` reference at `drain.go:1225` on our side after you land, and we've filed `internal/convoy/classes.go:159` separately — out of scope for this segment, we agree.
>
> Push the fix and we'll merge. Thanks for the care in this one; it's a large diff that was cheap to review, which is rare.

---

**Bottom line:** request-changes, high confidence. The regression is confirmed with a one-variable control and is reachable by renaming a rig. Two maintainer follow-ups first: force-push the branch back to `49927f5e7` to strip our red commit off @julianknutsen's PR, and look at the fix-executor wiring — the executor correctly reported `aborted / commit: none` in prose while its status JSON said `applied` with a SHA, and the pipeline pushed on the JSON.

julianknutsen:
Design verdict on the red row `TestPrepareWaitWakeState_PrefixUncoveredRigDependencyDoesNotReapTheWaiter` (added in e55acb3c8f):

The row is right, but only for the shape it builds: a rig whose CONFIGURED prefix ("fe") disagrees with the prefix its STORE declares ("ga"). That disagreement is a fault in the frame, and the residency doctrine already says a fault is never absence — so the reader must answer *unproven* (wait stays pending), never *proved absent* (wait failed).

The fix must stay bounded to that fault. Do NOT make every prefix-gated-out rig leg unproven: the plan gates rig legs on the configured prefix by design, the prefix owner is the authority for its namespace, and if any gated-out leg made absence unprovable then no wait on a multi-rig city could ever fail on a dependency that is genuinely gone.

Concretely: where the topology is assembled (assembleResidencyTopology / the frame the wait reader captures), detect a rig whose store-declared prefix differs from its configured prefix and treat that leg as out of frame the same way a suspended rig already is (`errWaitDependencyUnproven`), so absence over that frame cannot be proved. A gated-out leg whose two prefixes AGREE stays exactly as it is. Add the agreeing-prefix control next to the new row so the two directions are both pinned.

quad341 (approved):
@julianknutsen — thank you for taking the prefix-fault call yourself instead of handing it back, and for refusing the cheap version of it. `TestPrepareWaitWakeState_AgreeingPrefixRigStillProvesTheDependencyGone`, plus the mutation you applied and reverted, is what made this mergeable: it proves you didn't just turn our red row green by declaring every gated-out leg unprovable, which would have been worse than the freeze. Keeping our row byte-for-byte and building the control beside it is exactly right.

`TestClassifyDrainBlockerFaultIsAnErrorNeverAnOmittedEdge` is the other standout — a leg nobody could read silently deciding a blocker terminal is invisible in a green suite, and you pinned it. The `beadmeta` leaf extraction, with the `parsed == 0` vacuity guard, quietly cuts a dependency cone that had `storeref` dragging in config's process-spawning tree.

Merging now. The stale changes-request on this PR is ours, filed three heads back against the row you've since fixed; we'll clear it on our side.

One follow-up we'll file ourselves, nothing needed from you: on exec-backed rigs the store's declared prefix *is* the configured one, so a renamed rig there has no disagreement to detect. bd-backed rigs are covered. Thanks for the care on this.


---

PR #6142, opened 2026-09-08
fix(storeref): prove relics on the closed-inclusive census too (ga-q8ick)

julianknutsen:
One line of behavior: `storeref.ProvenLegacyResidents` now reads the
closed-inclusive relic census, the same one the retirement verdict moved to in
#5644.

## Mechanism

`HasLegacyResidents` (keep the probe?) and `ProvenLegacyResidents` (deny this
read?) are two verdicts over one census, and the only difference they are
allowed is their DEFAULT: an unreadable binding must keep a probe and must not
deny a read. #5644 (ga-qdt5y.19) widened the retirement verdict to count closed
relics; the proof was left on `OpenLegacyResidents`, so the two also disagreed
about the census, and that produced a third state nobody designed.

A refused city whose relics had all CLOSED landed in it: `HasLegacyResidents`
counts the closed relic, so the residence probe survives — but
`KnownLegacyResidents` stays false, so `probeRefusalPolicy` returns
`PolicyRefusalTolerated` instead of `PolicyFatal`, the probe leg is skipped over
the standing storage refusal, `cliByIDBindingOwner` reports `ok=false`, and the
by-id surface falls through to its own directory scan. The scan finds the copy
`gc storage migrate` preserved in the work store and never deleted, which reads
OPEN with pre-migration fields, serves it, and the next write writes it. Exit 0,
no diagnostic — the ga-q8ick symptom, restored by the act of DRAINING the city,
which is the thing the operator was asked to do.

A closed relic is proof for exactly the reason it keeps the probe: it is still
shown, reopened, claimed and written BY ID, and only the binding holds the live
row those reads must land on. Sharing the census also makes the implication
"proven ⇒ pessimistic bit" hold for every binding rather than only the ones a
constructor happened to align.

The doc comment above `ProvenLegacyResidents` had advertised this as a pending
one-line swap; it now states the shared-census rule instead.

## RED → GREEN

Both rows were run against this branch with the one-line swap stashed out
(source at `c6bb2aa30d`, tests as committed), then with it restored.

`internal/storeref` — `TestAClosedRelicIsProvenToo` (new, next to
`TestProvenLegacyResidentsFallsTheOTHERWay`):

```
--- FAIL: TestAClosedRelicIsProvenToo (0.00s)
    relic_census_test.go:262: a binding whose only relic has CLOSED came back
    unproven; it keeps a probe nothing may enforce, and that id falls to the
    frozen pre-migration copy in the work store
```

`cmd/gc` — `TestRefusedCityDeniesADRAINEDRelicToo` (new, the consumer-side
denial, next to `TestRefusedCityDeniesTheRelicItsLiveCensusProves`):

```
--- FAIL: TestRefusedCityDeniesADRAINEDRelicToo (0.03s)
    by_id_relic_proof_test.go:130: a refused city whose binding holds gc-relic1
    CLOSED resolved to ok=false with no error; the caller falls through to its
    scan and serves the copy the migration retained in the work store, which
    still reads open
```

With the swap restored both PASS. Each carries a control so the swap cannot pass
by proving everything: the unit row also asserts that a binding holding only its
own CLOSED in-namespace bead stays UNPROVEN (the namespace filter still decides
what counts), and the consumer row's empty-binding control is the clean
fall-through already pinned in the sibling row above it.

No existing test asserted the open-only behaviour for the PROOF, so nothing was
adjusted — one row was added on each side. The refused-city rows in
`cmd/gc/by_id_relic_proof_test.go` and the boot-census rows in
`cmd/gc/relic_census_boot_test.go` are unchanged and still pass.

## `OpenLegacyResidents` stays

Grep for the symbol across the tree: after this change it has exactly one
non-test caller.

- `cmd/gc/cmd_storage.go:570` (`reportBindingRelics`) — the `gc storage status`
  DRAIN gauge. Open-only is the point there: it is the number an operator
  watches fall as the carried-across work closes, and the widened census can
  never reach zero. Its doc comment already stated the split and needed no edit.
- `internal/storeref/relic_census_test.go` —
  `TestOpenLegacyResidentsIgnoresAClosedRelic` is the row that keeps the drain
  count open-only against a natural-looking tidy-up. Its comment is updated to
  say that BOTH verdicts have now moved off it (retirement in ga-qdt5y.19, the
  proof here) rather than only retirement; three other rows there still call it
  for the census-shape assertions.

Nothing else in the tree names either census function, so no caller had to
change to inherit the fix.

## Gates

All run on the branch head, from the worktree, in this order.

| Gate | Exit |
| --- | --- |
| `go build ./...` | 0 |
| `go vet ./internal/storeref/... ./cmd/gc/...` | 0 |
| `gofmt -l internal cmd` | 0 (no output) |
| `golangci-lint run ./internal/storeref/... ./cmd/gc/...` | 0 (`0 issues.`) |
| `go test ./internal/storeref/... -count=1` | 0 |
| `go test ./cmd/gc -run 'Relic\|Census\|Residency\|Refused' -count=1` | 0 |
| `make test-fast-parallel` | 0 (`All fast jobs passed`) |
| `.githooks/pre-commit` (hooksPath active) | 0 |
| `.githooks/pre-push` (pushed without `--no-verify`) | 0 |

`scripts/residency-boundary-baseline.txt` is untouched — no row added, none
removed — and the `// residency:allow` marker on the
`storeref.ProvenLegacyResidents` call in `cmd/gc/by_id_relic_proof.go` is
unchanged: the call site moved not at all, only the census it reads.

No base-owned failure was encountered; nothing is worked around.

Refs: ga-q8ick, ga-qdt5y.19, ga-j4p3y

### DIFF
--- a/cmd/gc/by_id_relic_proof.go
+++ b/cmd/gc/by_id_relic_proof.go
@@ -18,10 +18,10 @@ package main
 // The obvious place to keep this verdict is a note under the city's own .gc,
 // written by whichever process last managed to read the binding. It is the
 // wrong place twice over. It is a status file, which this codebase does not
-// keep (AGENTS.md: no status files — query live state), and it goes stale in
-// the DENYING direction: relics close over the weeks after a migration, so a
-// note written in March denies reads in June for a binding that has been clean
-// since April.
+// keep (AGENTS.md: no status files — query live state), and it could not
+// replace the census anyway: a note is only as good as the process that wrote
+// it, so it is absent on every city no such process has visited, and the read
+// below has to exist for those cities regardless.
 //
 // The premise the note was there to work around turns out to be false anyway.
 // "Refused" is a verdict about SERVING the binding — a convergence check, a
@@ -54,14 +54,14 @@ package main
 // The proof is keyed by BINDING ref, not by id, so what turns Fatal is the
 // whole binding's residence probe: every non-reserved by-id read on this city's
 // CLI fallback path denies, including a fresh post-migration work bead and a
-// rig-shadowed id that can have no frozen twin at all. That breadth is
-// deliberate rather than incidental. The census reads OPEN beads only, so it
-// cannot enumerate the closed relics a per-id rule would have to consult, and a
-// per-id rule would therefore deny LESS than the evidence supports. The corpus
-// pins it — residency_conformance_test.go's T3k rows make ByID(ga-xyz) and
-// ByID(ra-7) Fatal too, not just the reserved-prefix ones. The operator cost is
-// that one proven open relic takes the by-id door away for the city, not only
-// for the ids the migration preserved.
+// rig-shadowed id that can have no frozen twin at all. That breadth is a
+// deliberate choice rather than a forced one: the census enumerates closed
+// relics as well as open ones, so a per-id rule would have the population it
+// has to consult, and the corpus pins the binding-keyed breadth —
+// residency_conformance_test.go's T3k rows make ByID(ga-xyz) and ByID(ra-7)
+// Fatal too, not just the reserved-prefix ones. The operator cost is that one
+// proven relic takes the by-id door away for the city, not only for the ids
+// the migration preserved.
 //
 // # The absent case is the tolerant one
 //
--- a/cmd/gc/by_id_relic_proof_test.go
+++ b/cmd/gc/by_id_relic_proof_test.go
@@ -101,6 +101,39 @@ func TestRefusedCityDeniesTheRelicItsLiveCensusProves(t *testing.T) {
 	}
 }
 
+// TestRefusedCityDeniesADRAINEDRelicToo is the row above with the relic CLOSED,
+// and it is the case the symptom outlives.
+//
+// Draining a relic does not move it. The binding still holds the only live row
+// for that id, and `gc bd show`, `reopen` and `claim` all still ask for it by
+// id — while the work store still holds the copy `gc storage migrate` preserved
+// and never deleted, which reads OPEN with pre-migration fields. So a proof
+// that counted only OPEN residents handed the operator the worst arrangement of
+// the two: the probe survives the last close (HasLegacyResidents counts closed
+// relics since ga-qdt5y.19) but nothing is allowed to enforce it, and the read
+// falls through to the frozen copy exactly as it did before the fix — on the
+// city that has done the draining it was asked to do.
+//
+// The empty-binding control lives in the row above; what is added here is the
+// CLOSE, so a failure names the census rather than the plumbing.
+func TestRefusedCityDeniesADRAINEDRelicToo(t *testing.T) {
+	cityPath, _ := foreignProviderCity(t)
+	relic, classStore := classResidentWorkShapedBead(t, cityPath, "gc-relic1", "carried across, then drained")
+	if err := classStore.Close(relic.ID); err != nil {
+		t.Fatalf("draining the relic: %v", err)
+	}
+
+	refuseTheseCities(t, theRefusalARefusedCityCarries(), cityPath)
+
+	_, ok, err := cliByIDBindingOwner(cityPath, relic.ID)
+	if err == nil {
+		t.Fatalf("a refused city whose binding holds %s CLOSED resolved to ok=%v with no error; the caller falls through to its scan and serves the copy the migration retained in the work store, which still reads open", relic.ID, ok)
+	}
+	if !errors.Is(err, storeref.ErrProvenRelicRefusal) {
+		t.Errorf("the denial reads %q and does not carry the proven-relic sentinel; closing a relic must change the drain count, not the proof", err)
+	}
+}
+
 // TestRefusedCityThatCannotOpenItsBindingFallsThrough is the proof's absent
 // case, and it is deliberately the tolerant one.
 //
--- a/internal/storeref/bindings.go
+++ b/internal/storeref/bindings.go
@@ -49,9 +49,9 @@ type BindingOptions struct {
 	// the binding at all — the refused city, where the live census has no store
 	// to ask and every answer falls back to the pessimistic default. The plane
 	// supplies a verdict it took itself, at the time it asks, against a handle
-	// it opened for the read; this is deliberately NOT a durable record, which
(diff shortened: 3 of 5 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277311603

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR puts `ProvenLegacyResidents` on the same closed-inclusive census that `HasLegacyResidents` already read, so a migrated-then-refused city whose relics have all CLOSED still denies by-id reads instead of quietly answering them from the frozen pre-migration copy. That was exactly the case where the residency guard kept a probe nothing was permitted to enforce — a city could look protected while serving stale rows — so closing it matters for anyone who migrated and then drained.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- The divergence was real and the fix is pointwise. At the merge-base `HasLegacyResidents` already called `LegacyResidents` (widened by #5644) while `ProvenLegacyResidents` still called `OpenLegacyResidents`; the swap at `internal/storeref/relic_census.go:132` puts both verdict forms on the same census, leaving the inverted error default as the whole of the remaining difference. Reviewers verified this at the merge-base rather than taking the PR's stated bug at face value.
- The drain gauge stays open-only. `OpenLegacyResidents` is untouched and still the only feed for `gc storage status` (`cmd/gc/cmd_storage.go:570`), pinned by `TestOpenLegacyResidentsIgnoresAClosedRelic` with a doc naming both verdicts that departed.
- The namespace control was carried onto the new closed half. `TestAClosedRelicIsProvenToo` brings its own own-namespace control — a closed `gcg-1` inside the binding's reserved prefixes proves nothing — mirroring `TestAClosedInNamespaceBeadIsNotARelic`. Without it the widened predicate would have denied work-bead reads on every unconverged city that ever closed one of its own infra beads.
- Blast radius is genuinely one production consumer. `ProvenLegacyResidents` has exactly one non-test caller (`cmd/gc/by_id_relic_proof.go:204`), so the widening cannot reach a second decision path.
- The new rows are load-bearing, not decorative. Two reviewer lanes ran independent mutation probes at different granularities (a proof-only revert of `:132`, and a both-sites revert) and agree the new rows fail while the pre-existing open-relic row `TestRefusedCityDeniesTheRelicItsLiveCensusProves` stays green — isolating the CLOSED case specifically.

Review findings addressed:
- **Change impact / maintainability (minor):** four passages of prose still argued *from* the open-only census this PR abolished, in the proof's only caller and in the `KnownRelics` contract doc — the impossibility argument for binding-keyed breadth (`cmd/gc/by_id_relic_proof.go:58-60`), the "one proven **open** relic" trigger understatement (`:63`), and the twice-stated March/June anti-durability example (`:21-24` and `internal/storeref/bindings.go:52-54`), whose staleness mechanism the now-monotone verdict removes entirely. Each passage is the recorded reason for a standing design decision, so leaving them would let a later maintainer re-litigate settled design on grounds this PR had already falsified. Fixed in maintainer fixup `1621faf8`: re-grounded on what survives (the T3k conformance rows pin the binding-keyed breadth, and per-binding keying is now a deliberate choice rather than a forced one), with the falsified staleness example dropped in favour of the status-file doctrine that stands on its own. Doc-only, no behavior change.

- `ga-wdp4n` (docs, minor): `internal/api/state.go:175` still documents `ClassBindingHasLegacyResidents` as reporting "an **open** bead under a foreign id", but that surface has been fed by the closed-inclusive verdict since #5644. Ancestry checks put both that doc and the widening before this PR's merge-base, so it is pre-existing debt rather than something introduced here — filed so the by-id/proof doc surface converges in a single later pass.
- `ga-mb8xb` (existing, minor): the relic census's unbounded full-scan follow-up filed from #5641 should now be read as covering the proof's scan too — same cost class, reached only by a refused city, and `TestServedCityPaysNothingForTheRelicProof` still requires a served city to pay nothing for it.
- Informational, no action: on upgrade, a long-refused city that has fully drained its relics flips from silently serving the frozen copy to a hard by-id denial. That is the fix working, and the denial already routes the operator to `gc doctor`.

### Maintainer Changes
`docs(storeref,gc): re-ground the by-id proof's design record on the widened census` (`1621faf8`) — 2 files changed, 15 insertions(+), 15 deletions(-), across `cmd/gc/by_id_relic_proof.go` and `internal/storeref/bindings.go`. Comment-only.

### Final Review Status
Ready for the merge queue: final head `1621faf8` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/34198543606

### Review Iterations
One review pass across three reviewer lanes (design/implementation, implementation, and cross-file/test-risk), approved on the first pass with no fix iterations; the single doc-only minor was applied afterwards as the maintainer fixup above. Before handoff the adopted branch was re-validated against the latest `main` (`946363782d`): a clean rebase with an unchanged patch-id, `go vet` clean over `./cmd/gc/... ./internal/storeref/...`, `./internal/storeref` green, and all five relic/refused/by-id rows in `./cmd/gc` green.

---

quad341 (approved):
@julianknutsen — thank you for chasing this one all the way to the third state. That's the nastiest shape a bug can have: after ga-qdt5y.19 widened retirement, closing the last relic *kept* the probe but *dropped* the proof, so a refused city ended up carrying a residence probe nothing was permitted to enforce — and the trigger was the operator doing exactly what they were told. Draining the city restored the symptom. That is genuinely hard to see, and the writeup makes it visible.

Two things I especially appreciated. You didn't just flip the census — you went back and rewrote the doc comment that justified computing-not-recording, because its "a note goes stale in the denying direction" argument stops being true the moment the census is monotone. Leaving a dead argument standing is how the next person gets misled.

And both new rows carry controls that make them falsifiable. I put `ProvenLegacyResidents` back on `OpenLegacyResidents` and watched `TestAClosedRelicIsProvenToo` and `TestRefusedCityDeniesADRAINEDRelicToo` fail with exactly the messages in your body, then pass restored. A claim I could check rather than take on faith is why this is a fast merge.

Merging as-is. Nothing needed on my side.


---

PR #6143, opened 2026-09-08
fix(gc): align bd mutations with pool session identities

### LINKED ISSUE #5716: bug: unaliased pool workers deadlock in the claim loop — #5663 moved the claim assignee to GC_SESSION_ID but left BEADS_ACTOR on the slot label
### Before you continue

### Gas City version

`dev (commit: 37f1ac5f7, built: 2026-08-28T07:36:29Z)` — i.e. `main` as of 2026-08-27, which contains f173e5148.

### Environment

Ubuntu 26.04 LTS, bash, tmux provider, claude runtime. Dolt-backed bead store. Rig with pooled, **unaliased** `gc.*` agents (`gc.implementation-worker`, `gc.implementation-reviewer`, `gc.run-operator`), each spawned on unassigned demand.

### Reproduction

#5663 (f173e5148) moved the identity a claim is **recorded under**, but not the identity the worker **authenticates as**. The two halves now disagree for unaliased pool workers, and `bd` rejects every claim.

Writer half — `cmd/gc/cmd_hook.go`, changed by #5663:

```diff
- assignee := firstNonEmptyHookValue(alias, agentForQuery, resolvedAgentName, sessionName, sessionID)
+ assignee := hookClaimAssigneeIdentity(alias, sessionID, agentForQuery, resolvedAgentName, sessionName)
```

`sessionID` moves from last to second. `clearPoolTemplateRuntimeIdentity` blanks `GC_ALIAS` for an unaliased pool spawn, so the assignee becomes `GC_SESSION_ID`.

Actor half — `internal/session/lifecycle.go:78`, **not** touched by #5663 (it changed exactly two files, `cmd/gc/cmd_hook.go` and its new test):

```go
env["BEADS_ACTOR"] = AssigneeIdentifier(info)
```

and `AssigneeIdentifier` (`internal/session/assignee_identities.go:56`) is `alias → ConfiguredNamedIdentity → SessionNameMetadata → ID`, which for an unaliased pool worker resolves to the **slot label**.

To reproduce: route a bead to a pooled, unaliased agent, let the pool spawn on it, and have the worker run the standard claim protocol (`gc hook --claim --json`) more than once.

### Expected behavior

The worker claims the bead and proceeds.

### Actual behavior

`bd` rejects the claim with `assignee mismatch` (exit 13) because the bead is held under the session bead id while the worker presents the slot label. `gc hook --claim --json` returns the same bead every iteration, so the worker spins until it gives up or a human reassigns the bead. Two seats deadlocked within one hour on first exposure; one burned ~12 minutes before resolving, the other was still looping after 30+ retries in 3 minutes when a human intervened.

```
Worker env (unaliased pool spawn):
  GC_SESSION_ID=ga-r7xvxw                                      <- bead is assigned to this
  BEADS_ACTOR=eunice--gc__implementation-reviewer-3-pool        <- worker authenticates as this
  GC_AGENT=eunice--gc__implementation-reviewer-3-pool
  GC_SESSION_NAME=eunice--gc__implementation-reviewer-3-pool

Bead state while looping:
  tlp-wa1vkf  assignee=ga-r7xvxw  status=in_progress

Manual unstick that works (guarded, writes nothing on mismatch):
  bd update tlp-wa1vkf --assignee eunice--gc__implementation-reviewer-3-pool --if-assignee ga-r7xvxw
```

Note: the split alone is not sufficient to deadlock. Two other in-flight beads carried the same session-id assignee against a slot-label actor and were progressing normally — so the initial claim path is fine, and the failure appears on re-entry into the claim loop for a bead the worker already holds.

### Additional context

**Why CI is green.** `hook_claim_occupant_identity_test.go:305` asserts the slot label never appears as `BEADS_ACTOR` — but only on the `bd` subprocess `gc hook` spawns itself. It never exercises the worker session's own environment, which `lifecycle.go` sets at spawn and which still carries the slot label. A regression test asserting on the spawned worker's env would have caught this.

**The reasoning in #5663 is correct but scoped to aliased agents.** From its message: *"alias stays first in the precedence so named agents are untouched — their stable alias is what keeps bd's actor-to-assignee close check passing (#4981 / ga-i44k)."* That check is exactly what fails here; named agents are protected by their alias, unaliased pool workers have none.

**Not asking for a revert.** The behaviour #5663 replaced was silent work theft between pool occupants, and its message already anticipates this window: *"until ga-jrnou lands it can make stalls more visible before fewer."* Loud stalls are the better failure. This is about closing the writer/actor gap so the stalls stop being manual-recovery events.

**Fix candidates**, in the order I'd rank them:

1. Teach the actor-to-assignee check to accept any of the session's advertised identities. `sessionBeadAssigneeIdentities` (`pool_session_name.go:29`) already enumerates `bead.ID`, `session_name`, `configured_named_identity`, `alias` and alias history, and #5663's own message notes every *reader* leads with `bead.ID`. Only the check is strict. This is the conservative option and cannot strand the aliased path.
2. Set `BEADS_ACTOR` to the session bead id for unaliased pool spawns in `lifecycle.go`, so both halves agree. Simpler, but changes the identity every `bd` command in the worker's shell runs under.

**Related, not duplicate:** #5048 is the same failure class (assignee form vs runtime-exported identity form) but a different origin — the dispatcher's sanitized `bd__dog-*` spelling, reported against 1.4.0 well before #5663. Its third fix candidate ("bd's assignee matching canonicalizes … so all identity spellings compare equal") is essentially option 1 above, so one canonicalization fix would likely close both. #4437 (enforce live-session authorization for routed-work claims at the Beads write boundary) touches the same boundary.

jacobhausler:
Fixes #5716

## Summary

Unaliased pool claims are recorded under the durable session bead ID, while the
runtime `BEADS_ACTOR` may still be the reusable pool slot name. This makes a
worker unable to close or heartbeat its own claim through `gc bd`.

This change projects the scalar actor passed to `bd` from the same identity set
used by `gc hook --claim`:

- fresh claims use alias-first identity selection, then the durable session ID;
- close, heartbeat, and ordinary mutations reuse an exact stored assignee only
  when it belongs to the current session's identity set;
- foreign and mixed-owner mutations retain the ambient actor and remain
  rejected by `bd`; no `--force` behavior is widened;
- the in-process class-store claim door uses the same projection.

## Validation

- Focused identity and end-to-end `gc bd` claim tests via `ops/bin/nomad-run`
- Full `go test ./cmd/gc` via `ops/bin/nomad-run`
- `make build` via `ops/bin/nomad-run`
- `make check` via `ops/bin/nomad-run` (fmt, lint, vet, invariant scripts, and
  hermetic all-package tests)

This is a point fix; no design artifact or documentation update is required.

### DIFF
--- a/cmd/gc/bd_actor_identity.go
+++ b/cmd/gc/bd_actor_identity.go
@@ -23,6 +23,14 @@ func projectBdActorForMutation(args []string, ambientActor, claimActor string, i
 	ambientActor = strings.TrimSpace(ambientActor)
 	claimActor = strings.TrimSpace(claimActor)
 	if bdMutationIsClaim(args) {
+		if ids, ok, ambiguous := bdMutationWriteIDs(args); ok && !ambiguous && len(ids) == 1 {
+			if bead, present := targets[ids[0]]; present {
+				assignee := strings.TrimSpace(bead.Assignee)
+				if hookClaimHasIdentity(assignee, identities) {
+					return assignee
+				}
+			}
+		}
 		if claimActor != "" {
 			return claimActor
 		}
@@ -69,18 +77,22 @@ func projectBdActorForMutation(args []string, ambientActor, claimActor string, i
 func projectBdActorInEnv(args []string, env []string, targets map[string]beads.Bead) []string {
 	ambientActor := bdEnvValue(env, "BEADS_ACTOR")
 	identities := bdSessionIdentityCandidates(env)
-	claimActor := firstNonEmptyHookValue(
+	claimActor := bdSessionClaimActor(env)
+	projected := projectBdActorForMutation(args, ambientActor, claimActor, identities, targets)
+	if projected == ambientActor || projected == "" {
+		return env
+	}
+	return bdEnvWithValue(env, "BEADS_ACTOR", projected)
+}
+
+func bdSessionClaimActor(env []string) string {
+	return firstNonEmptyHookValue(
 		bdEnvValue(env, "GC_ALIAS"),
 		bdEnvValue(env, "GC_SESSION_ID"),
 		bdEnvValue(env, "BEADS_ACTOR"),
 		bdEnvValue(env, "GC_AGENT"),
 		bdEnvValue(env, "GC_SESSION_NAME"),
 	)
-	projected := projectBdActorForMutation(args, ambientActor, claimActor, identities, targets)
-	if projected == ambientActor || projected == "" {
-		return env
-	}
-	return bdEnvWithValue(env, "BEADS_ACTOR", projected)
 }
 
 func bdSessionIdentityCandidates(env []string) []string {
--- a/cmd/gc/bd_actor_identity.go
+++ b/cmd/gc/bd_actor_identity.go
@@ -1,0 +1,133 @@
+package main
+
+import (
+	"strconv"
+	"strings"
+
+	"github.com/gastownhall/gascity/internal/beads"
+)
+
+// projectBdActorForMutation keeps the scalar BEADS_ACTOR interface compatible
+// with the session identity set that gc uses for ownership decisions. bd can
+// compare one actor string at a time; when a command targets work already
+// assigned to this session, pass bd the exact stored spelling instead of the
+// runtime spelling that led gc to the same session.
+//
+// Fresh --claim operations use the same actor selector as hook claims. For
+// close, heartbeat, and ordinary updates, projection is deliberately narrower:
+// every known non-empty target assignee must be one of this session's
+// identities, and all owned targets must have one exact spelling. Otherwise
+// the ambient actor is preserved so bd's own ownership guard rejects a foreign
+// or mixed-owner mutation.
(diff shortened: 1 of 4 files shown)


jacobhausler:
The first PR evidence watchdog run failed at 05:15Z with `incomplete comprehensive evidence` because `CI / required` was still incomplete at evaluation time. That check completed successfully at 05:18Z; all 61 substantive check runs are now successful, with no code-check failures. I attempted `gh run rerun 34188317247 --failed`, but GitHub requires repository admin rights to rerun this workflow. This appears to be the existing watchdog timing race; please rerun the evidence workflow when convenient.

iamluxe77:
Maintainer action: Approve and run workflows, then squash-merge if this is still the intended identity alignment (related to #6324 / Opsime 8mnv.3).

We will not locally patch actor identity.

julianknutsen:
Superseded by #6324, which fixes the root cause at the single identity selector (`AssigneeIdentifier` -> `BEADS_ACTOR`/`GC_AGENT`), so raw `bd`, pinned pack prompts, and API assignment all agree, not only the `gc bd` wrapper. Thanks @jacobhausler for the analysis and repro. Leaving this open until #6324 merges, then it will be closed.

PR #6144, opened 2026-09-08
fix(worker): dispatch transcript reads on provider_kind, not the provider name

julianknutsen:
## What

`SessionHandle.historyProvider` resolved `Profile -> info.Provider -> spec.Provider`, so a custom provider alias was handed to `sessionlog` by **name**. `sessionlog.ReadProviderFile` picks its reader with `ProviderFamily(name)`, which meant:

- a provider named `glm53` (kind `zcode`) fell through to the default Claude JSONL reader and read **0 messages** from its whole-file mirror;
- a provider named `kimi-k3-manifold` (kind `claude`) matched the `kimi` substring and got the Kimi reader, also **0 messages**.

The CLI (`cmd/gc/session_logs_resolve.go`) and transcript discovery (`session.Manager.TranscriptPathClassified`) already prefer the raw `provider_kind`. The worker read path now uses the same precedence so the file found and the reader used agree:

```
Profile -> strings.TrimSpace(info.ProviderKind) -> info.Provider -> spec.Provider
```

`Profile` stays first: it is the `worker_profile` override.

## Side effect (intended)

`historyProvider` also feeds `snapshotTailActivity` / `TailActivityForProvider`, both gated on `sessionlog.DerivesActivityFromHistory`. A zcode-kind alias now derives tail activity from history, which is the documented contract for that family (this repo owns the zcode mirror writer; a trailing user message means a turn is in flight). Before this change `PhaseBusy` was unreachable for a custom zcode alias. Covered by `TestSessionHandleStateDerivesBusyFromZCodeMirrorByKind`.

Wire types are unchanged. API transcript/history responses populate their `provider` field from `info.Provider` (the name); the only place the resolved string surfaces is per-entry `Provenance.Provider`, which already carried the profile string (e.g. `claude/tmux-cli`) for profiled sessions.

**Provenance vs display (intended split).** For a kind-stamped alias the per-entry `Provenance.Provider` — surfaced as `SessionStructuredMessage.provider` and rendered as the dashboard transcript header chip — now reports the provider **family kind** (`zcode`, `claude`) rather than the alias name, because that is the family the reader actually dispatched on. Display surfaces keep the alias name: `State().Provider` (via `providerLabel`) and the session-level wire `provider` (from `info.Provider`) are unchanged. The split is observable even when the alias name already resolved to the right family — a `my-codex` alias with kind `codex` now reports `codex` per entry — so dashboard owners should expect that header chip to show the family rather than the alias.

## Tests (RED first, then GREEN)

`internal/worker/handle_history_provider_test.go`:

- `TestHistoryProviderPrefersProviderKindOverName` — precedence table: `glm53`/`zcode` -> zcode family, `kimi-k3-manifold`/`claude` -> claude, bare `codex-mini` keeps working, blank kind falls through, empty info falls back to the spec provider, `Profile` still wins.
- `TestSessionHandleReadsCustomZCodeProviderTranscriptByKind` — `History()` and `Transcript()` read a zcode mirror for a `glm53` seat through the zcode reader (2 turns, not 0).
- `TestSessionHandleReadsClaudeKindTranscriptDespiteKimiName` — a `kimi-k3-manifold` seat with kind `claude` reads its Claude JSONL through the Claude reader.
- `TestSessionHandleStateDerivesBusyFromZCodeMirrorByKind` — `State()` reports `busy` on a trailing user message and `ready` once the turn closes.

The keyed-start test injects a no-op `WithStaleKeyDetectionWaiter` instead of paying the 2s production delay.

All four failed on `origin/main` for the reported reason (0 entries / `ready` instead of `busy`) before the one-function change.

## Verification

- `go build ./...` — ok
- `go vet ./internal/worker/... ./internal/sessionlog/...` — clean
- `go test ./internal/worker/... ./internal/sessionlog/... -count=1` — all ok
- `go test ./cmd/gc/ -run TestGCNonTestFilesStayOnWorkerBoundary` — ok
- pre-commit hook: `make lint-changed` 0 issues, `make vet` clean, codegen unchanged
- pre-push hook: `make test-fast-parallel` passed

Refs: ga-0a1n5

### DIFF
--- a/internal/worker/handle_history_provider_test.go
+++ b/internal/worker/handle_history_provider_test.go
@@ -1,0 +1,296 @@
+package worker
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/gastownhall/gascity/internal/beads"
+	"github.com/gastownhall/gascity/internal/runtime"
+	sessionpkg "github.com/gastownhall/gascity/internal/session"
+	"github.com/gastownhall/gascity/internal/sessionlog"
+)
+
+// historyProvider picks the string sessionlog dispatches transcript reads and
+// tail-activity derivation on. A custom provider's NAME carries no family
+// signal and may carry a misleading one ("kimi-k3-manifold" is a claude seat),
+// so the raw provider_kind must win over the name. The worker_profile override
+// stays on top, and a session without a kind keeps resolving by name.
+func TestHistoryProviderPrefersProviderKindOverName(t *testing.T) {
+	cases := []struct {
+		name         string
+		profile      Profile
+		specProvider string
+		info         sessionpkg.Info
+		want         string
+		wantFamily   string
+	}{
+		{
+			name:         "custom zcode alias reads through the zcode family",
+			specProvider: "glm53",
+			info:         sessionpkg.Info{Provider: "glm53", ProviderKind: "zcode"},
+			want:         "zcode",
+			wantFamily:   "zcode",
+		},
+		{
+			name:         "claude alias with kimi in its name reads through claude",
+			specProvider: "kimi-k3-manifold",
+			info:         sessionpkg.Info{Provider: "kimi-k3-manifold", ProviderKind: "claude"},
+			want:         "claude",
+			wantFamily:   "claude",
+		},
+		{
+			name:         "bare provider without a kind keeps resolving by name",
+			specProvider: "codex-mini",
+			info:         sessionpkg.Info{Provider: "codex-mini"},
+			want:         "codex-mini",
+			wantFamily:   "codex",
+		},
+		{
+			name:         "blank kind falls through to the name",
+			specProvider: "codex-mini",
+			info:         sessionpkg.Info{Provider: "codex-mini", ProviderKind: "  "},
+			want:         "codex-mini",
+			wantFamily:   "codex",
+		},
+		{
+			name:         "empty info falls back to the spec provider",
+			specProvider: "codex-mini",
+			info:         sessionpkg.Info{},
+			want:         "codex-mini",
+			wantFamily:   "codex",
+		},
+		{
+			name:         "worker profile still wins over the kind",
+			profile:      ProfileClaudeTmuxCLI,
+			specProvider: "claude",
+			info:         sessionpkg.Info{Provider: "glm53", ProviderKind: "zcode"},
+			want:         string(ProfileClaudeTmuxCLI),
+		},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			h := &SessionHandle{session: SessionSpec{Profile: tc.profile, Provider: tc.specProvider}}
+			got := h.historyProvider(tc.info)
+			if got != tc.want {
+				t.Fatalf("historyProvider(Provider=%q, ProviderKind=%q) = %q, want %q", tc.info.Provider, tc.info.ProviderKind, got, tc.want)
+			}
+			if tc.wantFamily != "" {
+				if family := sessionlog.ProviderFamily(got); family != tc.wantFamily {
+					t.Fatalf("ProviderFamily(%q) = %q, want %q", got, family, tc.wantFamily)
+				}
+			}
+		})
+	}
+}
+
+// Reproduces ga-0a1n5 at the worker boundary: a provider NAMED "glm53" whose
+// kind is zcode must read its whole-file mirror through the zcode reader.
+// Dispatching on the name fell through to the Claude JSONL reader and
+// produced an empty transcript.
+func TestSessionHandleReadsCustomZCodeProviderTranscriptByKind(t *testing.T) {
+	root := t.TempDir()
+	workDir := t.TempDir()
+	handle := newCustomKindHandle(t, root, workDir, "glm53", "zcode", sessionpkg.ProviderResume{})
+	if err := handle.Start(context.Background()); err != nil {
+		t.Fatalf("Start: %v", err)
+	}
+	mirror := filepath.Join(root, "sess_glm53.json")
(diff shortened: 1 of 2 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277312535

quad341 (dismissed):
@julianknutsen — this is a really clean piece of debugging, and both halves of the collision are the kind of thing that only surfaces once someone actually runs a custom alias: `glm53` falling through to the Claude JSONL reader and returning an empty transcript, and `kimi-k3-manifold` getting grabbed by the `kimi` substring match despite being a Claude seat. Naming both directions and writing a test for each is what made this quick to verify.

What sold me is that the fix aligns the read path with the precedence `TranscriptPathClassified` was already using instead of inventing a third ladder — the file we find and the reader we use now agree by construction. I checked the `builtin_ancestor` rung you left out and it's the right call: every stamping site writes it alongside `provider_kind` under the same condition, so there's no session it would rescue.

Merging as-is. One FYI, not a change request: the `internal/worker` package tests don't run on PRs here (that coverage job is push-gated), so your four tests get their first CI execution on the merge to `main` — your local run is carrying them until then. Thanks for flagging the `PhaseBusy` side effect yourself rather than letting us discover it.


---

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277315125

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR fixes a genuinely
sneaky read-path bug: `SessionHandle.historyProvider` dispatched transcript
reads on the provider **name**, so a kind-stamped custom alias was handed to
the wrong `sessionlog` reader — a `glm53` seat (kind `zcode`) fell through to
the Claude JSONL reader on a whole-file zcode mirror and returned **0
messages**, and a `kimi-k3-manifold` seat (kind `claude`) matched the `kimi`
substring and did the same. Discovery had already found the right file by
kind, so the file found and the reader used disagreed and transcripts for
custom provider aliases came back silently empty. Adding the `provider_kind`
rung at the single chokepoint that feeds all four `historyProvider` consumers
makes those two agree, and it makes `PhaseBusy` reachable for zcode-kind
aliases for the first time.

PR branch.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- The new doc comment on `historyProvider` claimed it "mirrors the precedence
  used by transcript discovery ... so the file found and the reader used
  agree". Discovery (`session.Manager.TranscriptPathClassified`) resolves
  `provider_kind` then `provider` and nothing else — it has neither the worker
  `Profile` rung above nor the `spec.Provider` fallback below — so only the
  kind-over-name rung is actually mirrored. The claim is now narrowed to that
  rung, names the two rungs with no discovery counterpart, and records that the
  `builtin_ancestor` rung `session.ProviderFamilyFromInfo` walks first is
  skipped *deliberately* to match discovery.
- The provenance/display split was an undeclared contract change. For a
  kind-stamped alias, per-entry `Provenance.Provider` (surfaced as
  `SessionStructuredMessage.provider`, rendered as the dashboard transcript
  header chip) now reports the provider **family kind**, while display surfaces
  — `State().Provider` and the session-level wire `provider` — keep the alias
  name. That is observable even for an alias whose name already resolved to the
  right family (a `my-codex` alias with kind `codex` now reports `codex` per
  entry), so the PR description now spells the split out for dashboard owners.

Review findings addressed:
- Contract & interface fidelity (nit): doc comment overstated parity with
  transcript discovery — narrowed to the kind-over-name claim in
  `45b3179be8`, which also folds in the architectural-consistency reviewers'
  suggested clause about the deliberately skipped ancestor rung.

- ga-j72g7 (change impact / blast radius, minor): the API agent-output
  transcript path (`internal/api/handler_agent_output.go` and its SSE twin
  `handler_agent_output_stream.go`) still passes the config provider *name* to
  `DiscoverTranscript`/`ReadTranscript`, so the same defect class survives on
  the dashboard agent-output surface. That code is untouched by this diff and
  the name-only ladder predates it, and ga-0a1n5 scopes the fix to
  `historyProvider`, so it is tracked separately rather than widened into this
  PR. The bead already names both handler files and prescribes tests mirroring
  `handle_history_provider_test.go`.

### Maintainer Changes
`45b3179be8` docs(worker): narrow the historyProvider parity claim to
kind-over-name — 1 file changed, 8 insertions(+), 3 deletions(-)
(comment only; `historyProvider` behavior unchanged). Contributor authorship on
the fix commit is preserved.

### Final Review Status
Ready for the merge queue: final head `45b3179b` has passing required GitHub
checks (67 successful, none failing or pending, `mergeStateStatus` CLEAN), and
the merge-ready metadata is in place. The adopted branch was also refreshed
onto the current `main` (`3a09f0d1`) as a check: it rebases cleanly with an
unchanged patch-id and identical file blobs, and `go vet ./internal/worker/...`
plus the full `go test ./internal/worker/... -count=1` suite are green on that
refreshed tree — so no rebase push was needed. I am marking this PR

The strongest signal in the review was the test evidence: a six-case precedence
table tied to `ProviderFamily` expectations, two end-to-end reads through the
real reader stack, and a `State()` busy→ready transition driven by a real
mirror rewrite. Red→green was proven by a mutation probe — deleting the kind
branch turns all four new tests red — and all three reviewer lanes independently
re-derived the bug premise rather than accepting it at face value.

CI: https://github.com/gastownhall/gascity/actions/runs/34282353174

### Review Iterations
1 review pass across three reviewer lanes (Claude, Codex, Gemini) plus
synthesis, which reached approve with no blockers or majors; one maintainer
fixup applied afterwards for the non-gating documentation polish above.

---

quad341 (approved):
Re-reviewed `f387cbfc7`..`45b3179be` (11 new commits since the last review): no new concerns — the prior review stands.

PR #6145, opened 2026-09-08
fix(zcode): key the transcript mirror and sid store by seat, not session name

julianknutsen:
## What

The `zcode-repl` adapter keyed every piece of per-conversation state — the persisted provider sid, the export mirror gc reads transcripts from, the archive, and the CLI child's per-epoch `HOME` — by `<session_name>#<continuation_epoch>`. Two seats can share both (a pool slot re-seated within one run: `beads--gc__implementation-reviewer-1-pool#1` carried seats `gcg-session-575a839d...` and `gcg-session-b2f5746a...` of run `gcg-5436965277309570`), so the second seat resumed the first seat's zcode session and both wrote one mirror; `sessionlog.FindZCodeSessionFileByScope` then showed one seat's transcript for both.

The scope key now carries the session bead id the engine already exports to the pane as `GC_SESSION_ID` (`session.RuntimeEnv`): `<name>@<bead-id>#<epoch>`. A resumed seat keeps its bead id, so restart continuity is unchanged. `@` cannot survive the adapter's component sanitization (`tr -c 'A-Za-z0-9._-' '_'`), so seat and name-only scopes never collide, and neither's stale-epoch sweep globs the other.

## How

**Write side** — `internal/worker/adapters/zcode/zcode-repl`
- `seat_name` = sanitized `GC_SESSION_NAME` + `@` + sanitized `GC_SESSION_ID` when the engine exported one; otherwise the name alone (host-session setups, older engines) — fully backward compatible.
- sid file, mirror dir, archive sweep, and CLI `HOME` all key on `seat_name`; superseded-epoch sweeps are per seat, so a sibling seat's start no longer archives this seat's live mirror or deletes its sid.
- `adopt_name_only`: a seat-keyed start takes over same-epoch name-only state left by the previous adapter version (sid, mirror dir, CLI HOME — all three, because `--resume` reattaches through the CLI's sqlite in that HOME). The conversation resumes across the upgrade and no stale mirror stays adjacent in the live root for the model to read.

**Read side**
- `sessionlog.ZCodeSeatMirrorScope(name, beadID, epoch)` (new) and `FindZCodeSessionFileByScope` now takes the bead id: tries the seat scope first, falls back to the name-only scope only when the seat scope holds nothing — mirrors already on disk, and mirrors from an adapter not handed a bead id, stay readable. Per-scope resolution (newest real mirror, `pending-` placeholder as last resort, work-dir filter) is unchanged and factored into `findZCodeMirrorInScope`.
- `workertranscript.DiscoverScopedPath` carries the bead id; `session/chat.go` `TranscriptPathClassified` passes `b.ID`.

No wire types, API, or docs changed.

## Tests (TDD — each failed for the intended reason before the fix)

- `internal/sessionlog`: `TestZCodeSeatMirrorScope`, `TestFindZCodeSessionFileByScopeSeparatesSeatsSharingASessionName`, `TestFindZCodeSessionFileByScopeFallsBackToTheNameOnlyScope`
- `internal/worker/transcript`: `TestDiscoverScopedPathZCodeKeysBySeat`
- `internal/session`: `TestTranscriptPathZCodeResolvesEachSeatByBeadID` (bead → lookup wiring, incl. name-only fallback)
- `internal/worker/adapters/zcode` (real bash adapter + stub CLI): `TestSeatsSharingASessionNameKeepSeparateConversations`, `TestSeatAdoptsStateLeftUnderTheNameOnlyScope`; harness `seatKey()`/`epochScope()` mirror the adapter's key

## Verification

- `go build ./...` — ok
- `go vet ./internal/sessionlog/... ./internal/worker/transcript/... ./internal/session/... ./internal/worker/adapters/zcode/...` — ok
- `go test -count=1 ./internal/sessionlog/ ./internal/worker/transcript/ ./internal/session/ ./internal/worker/adapters/zcode/` — ok
- `go test -count=1 ./internal/worker/ ./internal/worker/workertest/ ./internal/worker/builtin/` (zcode phase-1/phase-2 fixtures and builtin profile) — ok
- `make lint-changed LINT_CHANGED_SCOPE=staged` — 0 issues
- `shellcheck -S warning zcode-repl` — only the pre-existing SC2115 at the archive `rm -rf` (identical on origin/main; untouched here)
- `.githooks/pre-commit` (origin/main copy) ran on the commit: format, lint, genspec/genclient/genschema (no drift), `go vet ./...`
- `.githooks/pre-push` ran `make test-fast-parallel` on push (see PR checks for CI)
- Not run: `make test-worker-inference PROFILE=zcode/tmux-cli` — the live leg needs `ZCODE_CJS` + `ZCODE_API_KEY` credentials, which this environment does not have. The live path is covered structurally: the engine exports `GC_SESSION_ID` to the pane (`session.RuntimeEnv`) and `handle.History` → `TranscriptPath` resolves through the seat-aware lookup.

Bead: ga-pm3tw

### DIFF
--- a/internal/session/chat.go
+++ b/internal/session/chat.go
@@ -1233,12 +1233,14 @@ func (m *Manager) TranscriptPathClassified(id string, searchPaths []string) (str
 	// zcode carries no session_key — no session-id flag, no hook plugin — so
 	// the keyed lookup above can never hit for it and the ambiguity guard below
 	// would leave every pooled worker transcript-dark. Its mirror is keyed by
-	// the identity the bead does hold.
+	// the identity the bead does hold: its name, its own id (the seat — two
+	// seats can share a name and epoch), and its continuation epoch.
 	if path := workertranscript.DiscoverScopedPath(
 		searchPaths,
 		provider,
 		workDir,
 		b.Metadata["session_name"],
+		b.ID,
 		b.Metadata["continuation_epoch"],
 	); path != "" {
 		return path, TranscriptFound, nil
--- a/internal/session/manager_test.go
+++ b/internal/session/manager_test.go
@@ -5551,3 +5551,56 @@ func TestPersistInvocationUsageCursor(t *testing.T) {
 		t.Fatalf("cursor metadata after no-ops = %q, want u2", got)
 	}
 }
+
+// TestTranscriptPathZCodeResolvesEachSeatByBeadID pins the wiring that keys a
+// zcode transcript by the seat. Two seats can share session_name and
+// continuation_epoch (a pool slot re-seated within one run), and the adapter's
+// mirror scope tells them apart only by the session bead id gc exports to it
+// as GC_SESSION_ID — so that is what the lookup must carry, with the name-only
+// scope kept as the fallback for mirrors written before the seat was part of
+// the key.
+func TestTranscriptPathZCodeResolvesEachSeatByBeadID(t *testing.T) {
+	store := beads.NewMemStore()
+	mgr := NewManagerWithOptions(store, runtime.NewFake())
+	workDir := t.TempDir()
+	const sharedName = "beads--gc__implementation-reviewer-1-pool"
+
+	var infos []Info
+	for _, title := range []string{"seat-a", "seat-b", "seat-c"} {
+		info, err := mgr.CreateSession(context.Background(), CreateOptions{Template: "helper", Title: title, Command: "zcode-repl", WorkDir: workDir, Provider: "zcode", Resume: ProviderResume{}, Hints: runtime.Config{}, ExtraMeta: map[string]string{"session_origin": "manual"}})
+		if err != nil {
+			t.Fatalf("Create %s: %v", title, err)
+		}
+		if err := store.SetMetadataBatch(info.ID, map[string]string{"session_name": sharedName, "continuation_epoch": "1"}); err != nil {
+			t.Fatalf("SetMetadataBatch %s: %v", title, err)
+		}
+		infos = append(infos, info)
+	}
+
+	searchBase := t.TempDir()
+	write := func(scope, id string) string {
+		dir := filepath.Join(searchBase, scope)
+		if err := os.MkdirAll(dir, 0o755); err != nil {
+			t.Fatalf("MkdirAll: %v", err)
+		}
+		path := filepath.Join(dir, id+".json")
+		body := `{"info":{"id":"` + id + `","directory":"` + filepath.ToSlash(workDir) + `"},"messages":[]}`
+		if err := os.WriteFile(path, []byte(body), 0o644); err != nil {
+			t.Fatalf("WriteFile: %v", err)
+		}
+		return path
+	}
+	seatA := write(sessionlog.ZCodeSeatMirrorScope(sharedName, infos[0].ID, "1"), "sess_a")
+	seatB := write(sessionlog.ZCodeSeatMirrorScope(sharedName, infos[1].ID, "1"), "sess_b")
+	legacy := write(sessionlog.ZCodeMirrorScope(sharedName, "1"), "sess_legacy")
+
+	for i, want := range []string{seatA, seatB, legacy} {
+		got, err := mgr.TranscriptPath(infos[i].ID, []string{searchBase})
+		if err != nil {
+			t.Fatalf("TranscriptPath(%s): %v", infos[i].ID, err)
+		}
+		if got != want {
+			t.Fatalf("TranscriptPath(%s) = %q, want %q", infos[i].ID, got, want)
+		}
+	}
+}
--- a/internal/sessionlog/zcode_reader.go
+++ b/internal/sessionlog/zcode_reader.go
@@ -107,25 +107,54 @@ func FindZCodeSessionFileByID(searchPaths []string, workDir, sessionID string) s
 	return bestPath
 }
 
-// ZCodeMirrorScope is the per-session, per-conversation subdirectory the zcode
-// adapter writes its mirrors into: "<session-name>#<continuation-epoch>",
-// sanitized the same way the adapter sanitizes it. Returns "" when either part
-// is missing.
+// ZCodeMirrorScope is the name-only, per-conversation subdirectory the zcode
+// adapter writes its mirrors into when it holds no session bead id:
+// "<session-name>#<continuation-epoch>", sanitized the same way the adapter
+// sanitizes it. Returns "" when either part is missing.
 func ZCodeMirrorScope(sessionName, continuationEpoch string) string {
 	sessionName = strings.TrimSpace(sessionName)
-	if sessionName == "" {
+	epoch := zcodeScopeEpoch(continuationEpoch)
+	if sessionName == "" || epoch == "" {
 		return ""
 	}
+	return sanitizeZCodeComponent(sessionName) + "#" + epoch
+}
+
+// ZCodeSeatMirrorScope is the per-seat mirror subdirectory:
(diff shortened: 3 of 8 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277312452

quad341 (dismissed):
@julianknutsen — this is excellent root-causing. Pinning it to a real observed collision (`beads--gc__implementation-reviewer-1-pool#1` carrying two seats of one run) instead of a hypothesis made the whole change checkable, and `GC_SESSION_ID` is exactly the right key: already exported, stable across a resume, no new plumbing.

The detail I appreciated most is the `@` delimiter argument. Picking a character that *cannot* survive `tr -c 'A-Za-z0-9._-'` is what makes the per-seat sweep globs provably non-overlapping rather than probably — I checked both sanitizers against each other and they agree exactly. Having `adopt_name_only` move all three pieces including the CLI `HOME`, because that's where `--resume` actually reattaches, is the part that's easy to half-do and miserable to debug later. The mtime inversions in the tests are a nice touch too.

Merging as-is. Two follow-ups I'll file on our side, not yours: per-seat sweeps mean a *retired* seat's mirror and CLI HOME are never reaped, and legacy adoption moves its three artifacts independently, so concurrent same-name starts could split them across the upgrade. Both are the right trade for this fix — thanks for it.


---

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277314969

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR re-keys the zcode transcript mirror, sid store, and CLI HOME by *seat* (`<session-name>@<session-bead-id>`) instead of session name alone, which fixes a real cross-seat data incident: two pool seats that shared a session name and continuation epoch collapsed onto one sid, one mirror directory, and one CLI HOME — so one seat could resume another's conversation, and a sibling's start could archive a live seat's mirror out from under it. That is exactly the class of bug that erodes trust in transcripts, and the fix puts identity where it belongs: the on-disk state now names which seat owns it.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **The collision itself, traced to its root.** Review independently confirmed the stated root cause against the superseded keying (`sid_path="$sid_dir/$sid_name#$continuation_epoch"`, where `$sid_name` derived from the session name alone) — a pool slot re-seated within one run necessarily produced identical keys. Reviewers also swept for an existing disambiguator that could have served instead and found none: `FindZCodeSessionFileByID` can't help because gc never learns zcode's provider session id, `DiscoverKeyedPath` excludes zcode by design, and the manager's same-workdir ambiguity guard only engages when the scoped lookup *misses* — pre-PR it hit, wrongly, for both seats.
- **Key safety on both sides of the language boundary.** The new `@` separator is outside the keep-set of both the adapter's `tr -c 'A-Za-z0-9._-' '_'` and Go's `sanitizeZCodeComponent`, so a seat key can never collide with or glob onto a name-only key. Verified independently by all three review lanes.
- **The write-side/read-side identity contract holds.** The adapter consumes `GC_SESSION_ID`, which `RuntimeEnv` sets to exactly `info.ID`, and the reader passes the bead's own `b.ID`. Review added the load-bearing point that this is an *enforced* invariant rather than a convention — crash adoption verifies pane `GC_SESSION_ID == bead.ID` before associating a live runtime with a bead.
- **Upgrade continuity in both skew directions.** New reader + old adapter resolves through the name-only fallback; new adapter + old state resolves through `adopt_name_only`, which moves all three pieces together and is pinned by a test asserting the resumed turn passes `--resume=<inherited sid>` under the migrated HOME. Adoption is ordered before each `mkdir -p`, so the `! -e target` guard is meaningful and each piece is independently idempotent.
- **Test evidence pins mechanisms, not outcomes.** The tests assert absence of `--resume` leakage across seats, sid contents, sibling state preservation, seat-wins-even-when-name-only-is-newer, placeholder-owns-scope, and manager-level wiring through `CreateSession`-produced IDs. Three lanes each ran the affected packages green at this head with `-count=1`, with `go vet` clean.

Review findings addressed:
- **[Resource Lifecycle & Cleanup, minor]** The `adopt_name_only` comment's hygiene claim ("no superseded copy stays adjacent") held only for the same-epoch state adoption actually moves. Adoption is same-epoch only and every sweep globs the seat key, which can never match a name-only path — so legacy state at a *different* epoch is deliberately left in place. Fixed in `2932ac1070` by documenting the residue and why a blanket name-only sweep would be unsafe.
- **[Behavioral Correctness, minor]** `FindZCodeSessionFileByScope`'s new comment said the name-only scope "is consulted only when the seat scope holds nothing". Because `findZCodeMirrorInScope` filters candidates by the mirror's embedded `directory`, the fallback also fires when the seat *has* written — just for a different work dir. The behavior is correct and load-bearing (it is what lets a seat find its own pre-upgrade mirror); only the comment overstated the guard. Comment reworded in `2932ac1070`.
- **[Concurrency, Ordering & State Safety, minor]** The adoption comment justified inheritance with "It belonged to the one seat that then had the name" — but in precisely the collision this PR removes, the legacy state is the *merged* history of both seats, and the inheritor is chosen by start order rather than identity. The behavior is the best available salvage of ambiguous legacy state and is strictly no worse than pre-PR; only the ownership premise was false. Corrected in `2932ac1070`.

All three were documentation-accuracy defects in comments this PR introduces — none required a behavior change, and none gated the merge. The same two claims also appeared in test doc comments, and were corrected there as well.

- `ga-m7p73`: **[Resource Lifecycle & Cleanup, minor]** offline reclaim for orphaned name-only mirror/sid state left by the seat-keying migration. Bounded at one legacy generation per pre-upgrade session name, arising only when a continuation-epoch bump landed between the last old-adapter start and the first seat-keyed start. Not fixed here by design — a seat-keyed start cannot safely sweep the name-only scope without risking state a concurrently live host-session seat of the same name still owns. An offline `gc doctor`-style reclaim can reason about which seats are live; the adapter cannot.

Two further items were carried as open risks rather than findings, and need no action: downgrade is one-way (an older binary can't resolve seat-scoped mirrors, so rolled-back seats go transcript-dark until reset — standard forward-only migration posture, no data loss), and `mv … 2>/dev/null || true` makes a failed adoption invisible, so a seat would silently start a fresh conversation. Separately, review noted a pre-existing epoch-normalization divergence between the shell and Go implementations — not introduced by this PR, only reachable with a non-numeric `GC_CONTINUATION_EPOCH`, which gc never generates — and correctly left it out of scope.

### Maintainer Changes
`2932ac1070` — `fix(zcode): correct the seat-keying comments' accuracy claims`. Comment-only, no behavior change: 4 files changed, 20 insertions(+), 8 deletions(-).

### Final Review Status
Ready for the merge queue: final head `2932ac1` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

The adopted patch was also re-validated against latest `main` after the review: it rebases cleanly with an unchanged patch-id, and `go vet` plus `internal/sessionlog`, `internal/worker/transcript`, `internal/worker/adapters/zcode`, `internal/worker`, and `internal/session` all pass on the refreshed head. That matters here because `main` picked up #6144 (`dispatch transcript reads on provider_kind`) in the meantime — adjacent territory — and no interaction surfaced.

CI: https://github.com/gastownhall/gascity/actions/runs/34285231082

### Review Iterations
1 review pass performed. Three lanes (design/implementation, implementation, and cross-file/test-risk) reviewed the same tree with full coverage; the pass returned `approve` with three minor comment-accuracy findings, all of which were resolved by a single maintainer fixup commit. No fix iteration was needed.

---

Issue #6110, opened 2026-09-07
bug: hook claim: session current_claim_bead_id is an unfenced read-compare-write, so two concurrent claims both win and a rollback can erase a later winner's pointer


jacobhausler:
## Summary

`session.Store.SetCurrentClaim` writes the session's `current_claim_bead_id`
pointer as a read-compare-write whose final write is unconditional — no
compare-and-swap and none of the two fences
`engdocs/design/session-store-fences.md` (Accepted) sanctions for exactly this
key family. Two processes claiming under one `GC_SESSION_ID` can both pass the
comparison and both proceed, and the claim path's rollback clears the pointer
without checking that it still names its own bead, so it can erase a *later*
winner's stamp. The only consumer of that pointer is `gc hook current`, whose
documented job is to name the bead a step closes — so a lost write is not
cosmetic, it mis-names a bead for a mutating command.

## Symptom

Seen as the phantom-claim / phantom-assignment family: a bead whose `assignee`
and `gc.routed_to` disagree with any live session; a bead left with no assignee
and no route after an undelivered claim unwound; a seat whose claim returned
work another seat had already closed. In every case the session bead's
`current_claim_bead_id` pointed at a bead its own work row did not.

## Reproduction

Race window: between the `validatedBead` read (`internal/session/store.go:215`)
and the metadata write (`internal/session/store.go:223`) — widened in practice
because the two records a claim touches (the session pointer, the work bead)
are different bindings whenever `[beads.classes.sessions]` is relocated, and
the two writers are separate processes (two CLI claims, or a CLI and the
controller) so no in-process lock covers them.

```bash
# one city whose session beads and work beads are separate class bindings
# ([beads.classes.sessions] relocated); two claim processes that name the SAME
# session - i.e. the same GC_SESSION_ID in their environment, which is what a
# double-start / an inherited hook environment produces for one pool slot.
# There is no target flag: the claim is scoped by the session's own
# identity and work_query config (`gc hook [agent] --claim --json`).
for i in 1 2; do ( gc hook --claim --json > /tmp/claim-$i.json 2> /tmp/claim-$i.err ) & done; wait
jq -r '.bead_id // empty' /tmp/claim-1.json /tmp/claim-2.json   # which seat got what
gc bd show <session-bead> --json | jq -r '.metadata.current_claim_bead_id'
```

Result: `expected exactly one process to receive a bead, the other none` → got:
both results carry a bead id, the session pointer ends up naming one of them,
[...]

PR #6146, opened 2026-09-08
feat(remote): add remote worker lifecycle endpoints

jacobhausler:
## Summary

Add the remote worker lifecycle subset for workers that are not running on the
city host:

- `gc bd update <id> --claim`
- `gc bd heartbeat <id>`
- `gc bd release-if-current <id> <assignee>`
- typed `gc bd close <id>` carrying the ADR-0009 work record

The server routes are scoped under `/v0/city/{cityName}/worker/`. The handler
delegates to the existing store capabilities (`Claim`,
`ConditionalAssignmentReleaser`, revision-fenced `ConditionalWriter`, and
`AtomicConditionalCloser`) and returns 501 when a resolved store cannot prove
the required atomic operation. It never emulates a CAS with read-then-write.

Heartbeat explicitly reports `lease_scope=bead-metadata`: this refreshes the
GC bead lease metadata and revision, not bd's native lease table. Typed close
persists `gc.work_outcome` and the disposition reason in the same atomic
terminal write; shipped closes require commit and branch fields. The remote
client carries the existing request authentication and per-request city-write
grant plumbing, and the OpenAPI/generated-client artifacts are regenerated.

## Design and provenance

- Design context: `engdocs/design/worker-runtime-transport-v0.md` and
  `engdocs/design/worker-conformance.md`.
- Reviewed source patch SHA-256:
  `3eb56cf49ebfd6602edd89d77cc3a4964fee790972a49d2cbd4f23b897c69591`.
- Upstream branch base: `c6bb2aa30dd660c9dd9f955f9410e1e6fc202817`.
- Candidate EDGE commit: `e6c7925975526597080a1a30c5915647406ad42f`.
- Westlands EDGE family: `remote-worker-lifecycle`, expiry `merge`.

## Validation

- Candidate `e6c792597`: `make build && go vet ./... && go test ./... -count=1 && make check` on the rhuidean Nomad builder, green (exit 0).
- Upstream changed-package gate: `go test ./cmd/gc ./internal/api -count=1`, green.
- Upstream `make check-docs`, green.
- The upstream full `make check` run reached the observable suite but was
  stopped by the builder's external-network restrictions in unrelated tests
  that clone or resolve `github.com`; those tests/files are not changed here.

Related to #6110.


jacobhausler:
Mechanic review found a specific remote-claim gap in the EDGE remote-worker candidate (not current main).\n\nObserved in the D3 window on runtime pin e6c792597: a Nomad worker with GC_TEMPLATE=worker-local ran the orthodox remote gc hook --claim --json and received drain/no_work, while the local hook for worker-local listed ready work. In cmd/gc/remote_worker.go, remoteWorkerIdentities included GC_ALIAS, GC_AGENT, GC_SESSION_NAME, and session ID, but not GC_TEMPLATE; remoteHookCandidates then could not match route targets stamped as worker-local or rig/worker-local.\n\nThe minimal candidate fix adds rig-qualified and bare template route fallbacks using the canonical agentScriptRig projection, while keeping them separate from concrete assignee identities, and adds focused coverage for a rig-qualified template route. Local candidate commit 11d8548dc passes the focused remote-worker tests and make build. The full cmd/gc test failure was identical on the parent edge commit and is baseline metrics_classifier_test.go noise; make check currently stops at pre-existing fmt-check diffs in unrelated files.\n\nPlease fold this specific fix into the public remote-worker implementation. The current #6146 head refactors the remote-worker lifecycle and does not contain the older remote_worker.go candidate file, so this should be treated as a follow-up to the family implementation rather than a claim that current main is already broken. Preserve the distinction between route fallbacks and assignee/adoption identities.

jacobhausler (commented):
I validated the remote worker lifecycle against a native-Dolt Work store and found one correctness gap that should be addressed before merge.

The remote Work-store capability path matters: without it, `gc bd update <id> --claim` and `gc bd heartbeat <id>` from an off-host allocation can answer HTTP 501 instead of reaching the store claim and lease operations. When the store genuinely cannot prove the required atomic operation, 501 is the right refusal; when the claim operation is supported but returns `ok=false`, the refusal must also remain truthful.

The remaining claim edge is a stale same-assignee row. A closed or blocked bead can still retain the previous worker name. The store returns that row with `ok=false`, but the handler must not infer an idempotent success from the matching assignee: the old path falls through to the lease read/stamp and can return 200 while re-stamping a lease on a non-claimable bead. A same-holder retry is idempotent only when the store returns `ok=true` for an in-progress bead.

Suggested patch:
- Treat `ok=false` as an unconditional refusal in the handler.
- Use the returned row only to explain the refusal (holder and underlying status).
- Add an end-to-end regression for a closed or blocked row retaining the requesting assignee, asserting a conflict response, unchanged status, and no lease rewrite.

Reproduction/verification receipt:
1. Reach the city API from an off-host worker while the Work class resolves to the native-Dolt store.
2. Exercise `gc bd update <id> --claim` and `gc bd heartbeat <id>`; the capability mismatch produces 501 on the pre-fix path.
3. Claim a bead, close it while retaining its assignee, and retry the claim. The pre-fix handler can report claimed; the corrected behavior is a truthful conflict with no lease mutation.

EDGE provenance: reproduced against Gas City main at `857ed398ad22de3d84b229a0f9b6557d694f4b5a`; the correction candidate is `380111fd58c18fd3db191c46382caffdd9e207cb`. `make build` and `make check` pass on a Linux builder; the EDGE evidence expires 2026-10-09.

jacobhausler:
Pushed four follow-up commits onto this branch. Three fix defects in the code this PR itself introduces; the fourth closes a test gap that the first three left open.

**`fix(api): dispatch the remote worker claim through ClaimFor`**
The claim handler type-asserted a single claim shape on the resolved store. Two backends expose the same compare-and-swap under different names — the in-process stores spell it `Claim(id, assignee)`, the bd-backed store spells it `ClaimAs(id, assignee)`, because bd names the actor explicitly rather than inheriting it from the store's `CommandRunner` environment. The assert therefore answered a typed 501 for a backend whose claim capability is present and working. `internal/beads/worker_ops.go` adds `ClaimFor` as the single discovery point; a store implementing neither shape now returns the named `ErrClaimUnsupported`, so the handler keeps its 501 for exactly that case instead of using it as a catch-all. CAS semantics are unchanged and remain the store's.

**`fix(gc): forward the conditional writer, claim and comment through the bead policy wrapper`**
`beadPolicyGraphStore` did not forward the conditional-write, claim or comment capabilities, so a fenced write through the wrapper was refused even when the backing store supported it.

**`fix(api): stamp the claimant session on a remote worker claim so the typed close lands`**
Without the claimant session on the bead, the typed close fence had nothing to match against.

**`test(api): guard the claim dispatch at the route, not only in the helper`**
`ClaimFor`'s own test proves it dispatches both shapes; it cannot prove the *handler* calls `ClaimFor`. That gap was real and measured — reverting the handler's call site to the old assert left the entire `internal/api` package green. The new route-level test resolves a store spelling the claim as `ClaimAs` with no `Claim` method at all and asserts the route claims it: not 501, exactly one `ClaimAs` call, and the assignee persisted.

**Verification.** Each of the four proven red against a revert of its own source hunks, not just green as committed:
- `ClaimFor`'s call site → the new route test fails carrying the false 501 body verbatim
- policy wrapper → `TestPolicyStoreGraphShapeForwardsConditionalWriter` and `TestPolicyStoreReportsAbsentConditionalWriter` fail
- session stamp → `TestWorkerClaimStampsTheClaimantSessionForTheCloseFence` and `TestOpenAPISpecInSync` fail

Green as pushed: `gofmt` clean, `go vet` clean, and full-package `go test` on `./cmd/gc/`, `./internal/beads/` and `./internal/api/` — whole packages, not filtered runs.

jacobhausler:
Follow-up for the remote typed-close surface:

- `e156d730c` calls the existing `gateWorkRecordClose` from `humaHandleWorkerClose`, so remote shipped closes use the same ADR-0009 ancestry predicate as local closes when `GC_WORK_RECORD_ENFORCE` is enabled.
- An unreachable shipped commit is refused before the atomic close write; an API server without a hosted rig checkout logs an explicit unverified skip and preserves the existing presence-only behavior.
- Added handler-level regression coverage for refusal/no-write and unknown-checkout defer.

Validation on the Nomad builder: focused and broader `internal/api` suites plus `make build` passed. The observable suite in `make check` recorded PASS; the local wrapper hit a terminal-summary shell error after the pass, so that run has no wrapper verdict. The city EDGE receipt and patch are at `ops/receipts/edge/2026-09-14-cr-347546-remote-worker-close-adr0009.{md,mbox}` (city-side staging follows).

This is intentionally a follow-up commit on #6146, not a duplicate feature PR.

PR #6147, opened 2026-09-08
fix(zcode): never run a prompt the drain timer cut in half (ga-fasb8)

julianknutsen:
## Symptom

`Mac / make test` intermittently fails with

```
adapter_test.go:914: prompts = ["still workin"], want [still working]
```

— one prompt, executed missing its last byte, and **no second prompt carrying
the missing byte**. Seen on PR #5634 run 33930571152 and again on PR #6140 run
34185189112 (job 101932090162, `blacksmith-...-12vcpu` macOS 15), both on PRs
that touch nothing under `internal/worker/adapters/zcode`.

This is not a test-only problem. Since 2026-09-07 the GLM 5.3 review arm on
maintainer-city runs production prompts through this same adapter
(`provider glm53 = builtin:zcode`), so a prompt losing bytes is a production
defect.

## Mechanism

`internal/worker/adapters/zcode/zcode-repl` read the first line of a burst
blocking, then coalesced the rest of the paste with a **timed whole-line read**:

```bash
IFS= read -r line                       # blocking, first line
while :; do
    more=
    IFS= read -r -t "$DRAIN_TIMEOUT" more    # 1s timer around a LINE read
    drain_rc=$?
    if (( drain_rc == 0 )); then line+=$'\n'"$more"; continue; fi
    if [[ -n "$more" ]]; then line+=$'\n'"$more"; fi   # <-- runs a partial line
    break
done
```

`read -t` **throws away everything it has already consumed** when the timer
fires. Both shells lose data, differently:

- **bash >= 4** (`builtins/read.def`, the `setjmp(alrmbuf)` / `sigalrm` arm):
  on timeout it does `input_string[i] = '\0'; t = savestring(input_string);
  run_unwind_frame("read_builtin"); retval = 128+SIGALRM; goto assign_vars;`.
  The **unterminated fragment is handed back in `$more`**, the old
  `if [[ -n "$more" ]]` branch appended it and the loop ran the prompt anyway.
  The CLI got a prompt missing its tail, and the tail arrived later as a
  *separate* prompt.
- **bash 3.2** — still `/bin/bash` on stock macOS, and what
  `#!/usr/bin/env bash` resolves to on this lane (`blacksmith` macOS image, no
  brew `bash`; `setup-gascity-macos` installs only `tmux jq flock`). Its
  `read.def` arms `alarm(tmout)`; `sigalrm()` does `longjmp(alrmbuf, 1)`, and
  the landing pad is `run_unwind_frame("read_builtin"); return
  (EXECUTION_FAILURE);` — it **never reaches the assignment**. The bytes the
  read had already pulled off the fd (one at a time: `unbuffered_read =
  (nchars > 0) || (delim != '\n') || input_is_pipe`) are simply **gone**, and
  `$more` is left untouched. So the prompt runs with those bytes missing and
  **nothing ever carries them** — exactly the observed shape.

Both are the same defect: **a prompt executed with bytes missing**.

The observed CI failure matches the bash 3.2 arm byte for byte: one prompt
(`"still workin"`), the missing `g` never reappearing as another prompt, exit
code 0, and a 5.10 s test — 4 s of the test's own sleeps plus exactly one 1 s
drain window, i.e. one turn, not two.

### Fix

Stop timing a line read. The idle timer now guards a **one-byte** read:

```bash
peek=
IFS= read -r -n 1 -t "$DRAIN_TIMEOUT" peek
(( peek_rc != 0 )) && break        # window closed; nothing was consumed
[[ -z "$peek" ]] && { line+=$'\n'; continue; }   # that byte was the newline
rest=
IFS= read -r -n "$MAX_LINE_BYTES" rest           # blocking: finishes the line
line+=$'\n'"$peek$rest"
```

A one-byte read has consumed **nothing** when it times out, so there is nothing
to discard; and once that byte is in hand the burst is known to be continuing,
so the rest of its line comes from a blocking read that cannot end early. A
prompt is only ever assembled from whole lines, on both shells — under *any*
interleaving, including the one CI hit.

`-n` on every read is load-bearing, not cosmetic: without it bash reads a line
through `zreadc`'s 128-byte static buffer and pushes the remainder back with
`zsyncfd`'s `lseek`, which **fails on a tty and on a pipe** (`r < 0` leaves
`lused`/`lind` untouched), so the leftover bytes live only in that buffer — and
the one-byte timed read, which is always unbuffered, would step straight past
them. `MAX_LINE_BYTES` caps a single *line*, not a prompt.

Two smaller losses in the same loop go with it: `line` is cleared before the
read so a read that ends on a signal cannot leave a stale prompt to be re-run,
and a non-empty line returned with a failed status (end of input on an
unterminated last line) is now run rather than discarded.

## Repro evidence

The **defect** reproduces deterministically on both shells. The **specific CI
interleaving** does not reproduce on Linux, and is stated as unreproduced
rather than guessed at:

| probe | result |
| --- | --- |
| `go test -run TestInterruptWhileIdleDoesNotExit -count=100 -race` (bash 5.2) | green, no repro (516 s) |
| same test `-count=200` with bash 3.2.0 built from source, first on `PATH` (so `env bash` resolves to it) | green, no repro (1027 s) |
| 400 concurrent runs of the exact signal/write sequence under bash 3.2, jittered idle 1.0–2.5 s and gap 0.02–2.5 s | `bad=0 / 400` |
| bash 3.2 source read (`builtins/read.def`, `lib/sh/zread.c`) + the official 3.2 patch set (022/038/051 are the only read.def patches; 051 fixes the `interrupt_immediately` leak, so it is not in play on 3.2.57) | the blocking read at line 572 **cannot** return a short line with status 0 — the loop only exits at `c == delim`, EOF or error, and the latter two set a non-zero status. The drain read is the only read in the script that can hand the loop a line with bytes missing. |

## RED → GREEN

New test `TestALineSplitAcrossTheDrainWindowStaysOnePrompt` — one line whose
bytes straddle the drain window:

| shell | before | after |
| --- | --- | --- |
| bash 5.2 | `prompts = ["first line\nsecond line, part one - " "part two"]` — truncated prompt executed, tail ran as a second prompt | `["first line\nsecond line, part one - part two"]` |
| bash 3.2 | `prompts = ["first line" "part two"]` — `"second line, part one - "` **gone** | `["first line\nsecond line, part one - part two"]` |

`TestDrainKeepsPartialTrailingLine` also goes RED on bash 3.2 before
(`prompts = ["first line"]`) and green after; its `runtime.GOOS == "darwin"`
gate is deleted, because keeping the fragment was only unrecoverable on 3.2 while
the timer guarded a line read.

Production shape, driven over a **real pty** (not a pipe) with the gc wire
sequence `body / debounce / ESC / Enter`:

| script | bash 3.2 | bash 5.2 |
| --- | --- | --- |
| `origin/main` | `['alpha\nbeta\n\ngamma', 'delta']` — a whole line of the prompt lost | `['alpha\nbeta\n\ngamma', 'delta\nepsilon part one - ']` — truncated |
| this branch | `['alpha\nbeta\n\ngamma', 'delta\nepsilon part one - part two']` | same |

No pane echo regression: `-n` makes bash save/restore the tty around each read
(`ttsave`/`ttonechar`/`ttrestore`), and `tt_setonechar` does not touch `ECHO`,
so the pasted body still stays out of the pane.

## Gates

| gate | result |
| --- | --- |
| `go build ./...` | 0 |
| `go vet ./internal/worker/...` | 0 |
| `gofmt -l internal` | 0 (no output) |
| `golangci-lint run ./internal/worker/...` | 0 issues |
| `shellcheck internal/worker/adapters/zcode/zcode-repl` | 1 pre-existing SC2115 warning, **base-owned** — byte-identical on an untouched `origin/main` export of the file (line 172 there, line 182 here). No new finding. |
| `go test ./internal/worker/adapters/zcode/ -count=20` (bash 5.2) | ok, 515 s |
| `go test ./internal/worker/adapters/zcode/ -count=20` (bash 3.2) | ok, 515 s |
| `go test ./internal/testpolicy/resourcecensus/ -count=1` | ok — ledger flat (the new fixed sleep is offset by the `waitForTurns` swap) |
| `make test-fast-parallel` | run by the pre-push hook (no `--no-verify`) |

## Follow-up after merge

`~/.local/bin/zcode-repl` on maintainer-city is a **copy** refreshed from
`origin/main` (last refreshed to `26542454e5`; see memory
`codex-outage-cutover-20260907`), not a symlink into the repo. It must be
refreshed again after this merges, or the GLM 5.3 review arm keeps running the
lossy reader.

Refs ga-fasb8, ga-j4p3y.

### DIFF
--- a/internal/worker/adapters/zcode/adapter_test.go
+++ b/internal/worker/adapters/zcode/adapter_test.go
@@ -497,7 +497,7 @@ func TestIdleSeparatedPromptsStaySeparate(t *testing.T) {
 	s.send("first prompt")
 	time.Sleep(2500 * time.Millisecond)
 	s.send("second prompt")
-	time.Sleep(2500 * time.Millisecond)
+	s.waitForTurns(2)
 	if _, code := s.closeAndWait(); code != 0 {
 		t.Fatalf("exit code = %d, want 0", code)
 	}
@@ -705,6 +705,31 @@ func TestALineSplitAcrossTheDrainWindowStaysOnePrompt(t *testing.T) {
 	want := "first line\nsecond line, part one - part two"
 	if got := h.prompts(); !equalStrings(got, []string{want}) {
 		t.Fatalf("prompts = %q, want %q", got, []string{want})
+	}
+}
+
+// An interrupt that lands with a half-delivered line in the adapter's hands
+// must DROP it, not run it. bash >= 4 assigns the partial input to the
+// variable and returns rc=1, which is indistinguishable from end-of-input on
+// an unterminated last line by status alone — so the loop has to consult the
+// INT trap, or a cancelled prompt is executed with its tail missing.
+func TestInterruptWithAPartialLineInHandRunsNoPrompt(t *testing.T) {
+	t.Parallel()
+
+	h := newHarness(t, nil)
+	s := h.start()
+	s.waitForOutput("zcode-repl ready", adapterWaitBudget)
+	s.sendRaw("half a prompt, interrupted here")
+	// The adapter is silent while reading, so there is no lifecycle signal for
+	// "the bytes are in hand"; this gap is what puts them there.
+	time.Sleep(2500 * time.Millisecond)
+	s.signal(syscall.SIGINT)
+
+	if _, code := s.closeAndWait(); code != 0 {
+		t.Fatalf("exit code = %d, want 0", code)
+	}
+	if got := h.prompts(); len(got) != 0 {
+		t.Fatalf("an interrupt-truncated fragment was executed as a prompt: %q", got)
 	}
 }
 
--- a/internal/worker/adapters/zcode/adapter_test.go
+++ b/internal/worker/adapters/zcode/adapter_test.go
@@ -712,7 +712,7 @@ func TestALineSplitAcrossTheDrainWindowStaysOnePrompt(t *testing.T) {
 // must DROP it, not run it. bash >= 4 assigns the partial input to the
 // variable and returns rc=1, which is indistinguishable from end-of-input on
 // an unterminated last line by status alone — so the loop has to consult the
-// INT trap, or a cancelled prompt is executed with its tail missing.
+// INT trap, or a canceled prompt is executed with its tail missing.
 func TestInterruptWithAPartialLineInHandRunsNoPrompt(t *testing.T) {
 	t.Parallel()
 
--- a/internal/worker/adapters/zcode/adapter_test.go
+++ b/internal/worker/adapters/zcode/adapter_test.go
@@ -665,8 +665,9 @@ func TestDrainKeepsPartialTrailingLine(t *testing.T) {
 	s.sendRaw("\n")
 	s.waitForTurns(1)
 	// Surviving the drain is the assertion that must hold on every platform: an
-	// unbound $more here killed the adapter under `set -u` on bash 3.2, which
-	// is exactly the regression this test's own shape provokes.
+	// unbound drain variable here killed the adapter under `set -u` on bash
+	// 3.2, which is exactly the regression this test's own shape provokes. The
+	// drain still clears its variables before every read for that reason.
 	if _, code := s.closeAndWait(); code != 0 {
 		t.Fatalf("exit code = %d, want 0", code)
 	}
@@ -709,10 +710,15 @@ func TestALineSplitAcrossTheDrainWindowStaysOnePrompt(t *testing.T) {
 }
 
 // An interrupt that lands with a half-delivered line in the adapter's hands
-// must DROP it, not run it. bash >= 4 assigns the partial input to the
-// variable and returns rc=1, which is indistinguishable from end-of-input on
-// an unterminated last line by status alone — so the loop has to consult the
-// INT trap, or a canceled prompt is executed with its tail missing.
+// must DROP it, not run it. The signal here is followed by a stdin close, and
+// that end-of-input is what ends the read: it returns rc=1 with the partial
+// input assigned, which by status alone is indistinguishable from the
+// unterminated last line the adapter deliberately runs — so the loop has to
+// consult the INT trap, or a canceled prompt is executed with its tail
+// missing. A trapped INT on its own does not necessarily end the read: on bash
+// 5.2, if the rest of the line arrives the read resumes and completes with
+// rc=0, so this pins the end-of-input shape, which is the one that would
+// otherwise run the fragment.
 func TestInterruptWithAPartialLineInHandRunsNoPrompt(t *testing.T) {
 	t.Parallel()
 
--- a/internal/worker/adapters/zcode/adapter_test.go
+++ b/internal/worker/adapters/zcode/adapter_test.go
@@ -9,7 +9,6 @@ import (
 	"os/exec"
 	"path/filepath"
 	"regexp"
-	"runtime"
 	"strings"
 	"sync"
 	"syscall"
@@ -649,7 +648,11 @@ func TestBracketedPasteWrappersAreStripped(t *testing.T) {
 }
 
 // A drain read that times out still holds whatever partial line arrived; losing
-// it silently truncates the prompt's last line.
+// it silently truncates the prompt's last line. This held only on bash >= 4
+// until the drain stopped timing a line read: bash 3.2 consumed the fragment
+// off the fd and discarded it before the script regained control. Now the
+// timer guards a one-byte read that has consumed nothing when it fires, so the
+// fragment survives on every shell and the assertion is unconditional.
 func TestDrainKeepsPartialTrailingLine(t *testing.T) {
 	t.Parallel()
 
@@ -660,7 +663,7 @@ func TestDrainKeepsPartialTrailingLine(t *testing.T) {
 	s.sendRaw("trailing line without a newline")
 	time.Sleep(2500 * time.Millisecond)
 	s.sendRaw("\n")
-	time.Sleep(2500 * time.Millisecond)
+	s.waitForTurns(1)
 	// Surviving the drain is the assertion that must hold on every platform: an
 	// unbound $more here killed the adapter under `set -u` on bash 3.2, which
 	// is exactly the regression this test's own shape provokes.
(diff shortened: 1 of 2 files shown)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277312676

quad341:
@julianknutsen — this is superb work. Tracing an intermittent macOS flake down through `read.def`'s `setjmp(alrmbuf)` arm to "`read -t` discards what it already consumed, differently on 3.2 than on 4+" is not easy debugging, and the fix is the good kind: timing a one-byte read means there is nothing to lose when the timer fires, so the whole class of bug goes away instead of the symptom. I pulled both loops apart and confirmed the RED→GREEN myself. I also appreciate that you called the CI interleaving **unreproduced** rather than claiming the flake solved — that honesty is worth more than a tidier story.

One thing to flag, then I'm merging: with the new guard, a SIGINT that lands mid-line now *runs* the fragment where `main` dropped it. On bash 5.3 an interrupted `read` returns rc=1 with the partial input assigned, so `rc != 0 && -n "$line"` reads as EOF-on-unterminated-line. Same shape in the blocking `rest` read. Small, and squarely our integration work — I'll add the `interrupted` check plus a regression test on my side, rerun the adapter suite, and land it with your commit and credit intact. I'll also refresh the maintainer-city `zcode-repl` copy so the GLM arm stops running the lossy reader.


---

quad341:
Maintainer note: the fix-merge fixup for this PR (dcf6aad47) was held back by our own pre-push lint gate on a comment spelling (`cancelled` vs `canceled`, the misspell linter). That was on our side, not yours. Corrected in d6e8b1832 on top of the fixup, lint clean; the review re-runs on the new head and lands on its verdict. Nothing needed from you, @julianknutsen — thanks for the fix itself.

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277321021

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR fixes a real prompt-corruption bug in the `zcode-repl` adapter: the burst drain's 1s idle timer guarded a multi-byte read, so a slow sender could make the timer fire mid-line and hand the loop half a prompt — which the adapter would then run. Because `zcode-repl` is `go:embed`-ed, every zcode seat inherits that behavior, so this is a fix that reaches the whole fleet rather than one call site.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **A prompt the drain timer cut in half could be executed** (`ffeef6483c`): the idle timer now guards only a one-byte peek, which consumes nothing when it fires, so the timed multi-byte read that produced the defect no longer exists. The review proved the failure red on base with a `go test -overlay` probe reproducing the exact claimed shape, and all three reviewer lanes independently confirmed the mechanism.
- **An interrupt-truncated fragment could reach the turn, and an unterminated last line was silently dropped** (`266fab8bb3`): base `break`-dropped the operator's last line at end of input; head runs it, with the interrupt guard ordered first so a post-interrupt EOF cannot execute a canceled fragment. Clearing `line=` per iteration correctly protects bash 3.2 from re-running a stale value.
- **Spelling drift against the repo's lint locale** (`953b8ba4cb`): the Go file now uses US "canceled", which is what golangci's misspell enforces.
- The two new tests are real-time and signal-driven rather than string-matched, and the test-resource census stays flat (14 `time.Sleep` sites before and after), so no ledger row moves.

Review findings addressed:
- **Both new interrupt comments documented a bash mechanism that is false on the deployed shell** (maintainer fixup `9fbce314e9`). The comments claimed a trapped `INT` makes the read hand back whatever it had already taken (rc=1, variable assigned). Probed on bash 5.2.21 with the adapter's exact read shape, a trapped `INT` does not end the read at all — if the rest of the line arrives, the read resumes and completes with rc=0 — and the rc=1-with-fragment signature actually belongs to the end of input that follows. That is also how `TestInterruptWithAPartialLineInHandRunsNoPrompt` really passes: its signal is followed by a stdin close. The guard itself is still correct as a flag/status disjunction; only the documented mechanism was wrong, along with its implied promise that a live-pane C-c can recall bytes already taken.
- **The per-line cap's consequence was understated** (`9fbce314e9`): an over-cap line does not merely split — the read returns rc=0 mid-line and the drain rejoins the remainder with a newline the sender never sent, so the prompt's content changes and not just its framing. The user-facing header contract and the `MAX_LINE_BYTES` comment now both say so, and note that `read -n` counts characters rather than bytes.
- **A comment falsified by this branch's own rename** (`9fbce314e9`): `TestDrainKeepsPartialTrailingLine` still pointed at `$more`, which this branch renamed away; it now points at the drain's current variables.
- **The new comment's "cancelled" spelling** (`9fbce314e9`): spelled "canceled" to match the lint commit. The file's remaining "cancelled" spellings were left alone deliberately — most are the embedded helper's mode identifier, so a comment-only sweep would decouple them from the literal they describe.

- `ga-z4ab8` (correctness, minor): a C-c during the burst drain still runs the canceled prompt, and the stale `interrupted` flag then leaks into `run_turn`, where it can mislabel a plain turn failure as "turn interrupted". Verified **pre-existing** — base's drain loop had no `interrupted` check either — so this PR does not regress it and it was not a merge gate here. The fix has to sit after the read rather than on its return code.
- `ga-f019i` (contract fidelity, minor): `read -n` counts characters, not bytes, so `MAX_LINE_BYTES` is optimistic in a multibyte locale and wants a rename; and the over-cap join has a real remaining choice — document-only (which this PR took) versus appending a cap-hit remainder without the `$'\n'` separator. Both are carried on that existing bead alongside the related `LC_ALL=C` peek hardening, since they are the same edit.

Neither is reachable in normal operation, which is why they are follow-ups rather than blockers.

### Maintainer Changes
One comment-only maintainer fixup was pushed to the PR branch:
`9fbce314e9 docs(zcode): correct the interrupt mechanism these comments document` — 2 files changed, 36 insertions(+), 13 deletions(-). No behavior change.

### Final Review Status
Ready for the merge queue: final head `9fbce314` has passing required GitHub checks (67 passing, none failing or pending), and the merge-ready metadata is in place. The adopted branch was additionally re-validated against the latest `main` locally — the rebase is conflict-free with an unchanged patch-id, no upstream commit since the reviewed base touches either file, and `go build ./...`, `go vet`, and `go test ./internal/worker/adapters/zcode/... -count=1` are all green at that composition. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/34321701588

### Review Iterations
One review pass across three reviewer lanes (implementation, design-implementation, and design/test-risk), resolved to approve with no required changes. One severity disagreement between lanes was settled by a first-hand shell probe rather than by vote, and the single maintainer fixup above followed from it.

---

Issue #6148, opened 2026-09-08
polecat handoff is bypassable prose: an agent can reassign to the refinery without a branch on origin (phantom handoff)


atbrace:
## Symptom

A polecat's work bead reads as handed off — `assignee=<refinery>`, `metadata.branch` set, clean terminal exit — while the branch does not exist on origin. Nothing a dashboard or sweep reads distinguishes it from success; only `git ls-remote` does.

Measured on a live city running v1.4.x + backports with a local-model lane (pi) in the polecat pool: **4 phantom handoffs in 13 completed slots** across three pool widths. One was worked on a **detached HEAD** whose branch was never created, let alone pushed.

## Cause

The handoff in `mol-polecat-work`'s `submit-and-exit` step is ~130 lines of shell in the step prose. It already contains three fail-closed gates — a branch-shape check, the `git push` exit code, and an `ls-remote` verify — and the agent skipped all three and ran the bead write (`gc bd update … --assignee <refinery>`) that looks like completion. The defect is not a missing check; every check lives in text the model can decline to execute, while the write that marks success is one it performs happily. No addition to the prose can fix that.

The refinery then selects on "metadata.branch is set" and fails at `git checkout -b temp origin/$BRANCH`, which looks like a rebase failure and bounces the bead with a bogus rejection reason.

## Proposed fix

A single Go verb, `gc submit`, that performs the whole handoff atomically and whose bead write is unreachable unless the pushed ref was read back from origin at the local HEAD: resolve bead → refinery is a configured agent → HEAD on the expected branch and not detached → clean tree → commits beyond base → push → `ls-remote` == HEAD → only then write `metadata.branch`/`target`, clear `gc.routed_to`, reassign. The formula's shell is then deleted so no bypassable path remains. Defence in depth on the consumer side (refinery bouncing a `metadata.branch` that does not resolve on origin with `rejection_reason="phantom handoff"`) is a separate, smaller change.

Related: #2082 (branch-validation gate at submit, the prose version).

PR #6149, opened 2026-09-08
feat(cli): gc submit — atomic polecat -> refinery handoff whose bead write requires a verified remote ref

atbrace:
## Summary

Adds `gc submit`, the polecat → refinery handoff as one fail-closed verb, and registers it in the command census and CLI reference.

Today the handoff is ~130 lines of formula shell (branch-shape gate, `git push`, `ls-remote` verify, `bd update` of `metadata.branch`/`target`, reassignment to the refinery). Every check lives in prose the agent can decline to run, while the bead write that *looks* like completion is one it performs happily. On a live city with a local-model lane in the pool we measured 4 phantom handoffs in 13 completed slots (~31%): `assignee=refinery`, `metadata.branch` set, clean exit, and no branch on origin — one of them worked on a detached HEAD whose branch was never created. Nothing a dashboard or sweep reads distinguishes those from success.

`gc submit --bead <id>|--convoy <id> --base <branch> --refinery <qualified agent>` does, in order: resolve the work bead (the convoy's single open member); require the refinery to be a configured agent; require HEAD on the expected branch (`metadata.branch`, else `polecat/<bead>`) and not detached; require a clean tree; fetch the base and require commits beyond it; push; read the ref back with `ls-remote` and require it at the local HEAD. **Only then** does it write `metadata.branch`/`target`, clear `gc.routed_to` and reassign to the refinery, so "handed off" means "the branch is on origin" by construction. Every refusal exits non-zero before any store write. `metadata.auto_push=false` halts at branch-ready without pushing, matching the previous formula contract. Local branch deletion and the refinery wake/nudge stay best-effort after the write.

The verb uses the same bd-backed store `gc hook --claim` uses (`hookClaimBdStore` in the worktree dir), `listConvoyChildren`, `findAgentByQualified`, and the in-process `cmdSessionWake`/`cmdSessionNudge`. Git calls go through a small `submitGit` helper with `git.SanitizedEnv()`; there was no push/ls-remote helper to reuse.

## Testing

- [x] `cmd/gc/cmd_submit_test.go` (MemStore + func-field ops seam): happy path asserts the exact op order (refinery check, branch, clean, fetch, ahead, push, ls-remote, then the single bead write, then delete/wake/nudge); detached HEAD, wrong branch, base-as-branch, dirty tree, zero commits, unconfigured refinery, push failure, ls-remote empty and ls-remote SHA mismatch each leave the bead untouched; preset `metadata.branch` honoured; `auto_push=false` halts without pushing; convoy single-member resolution and ambiguity refusal; wake/nudge failure non-fatal.
- [x] Census/CLI registries: `gen-command-census -check` clean; `go test -run 'Submit|Fresh|ProductMetrics|CLIDocs|Census|InjectedImmutable' ./cmd/gc ./cmd/gen-command-census ./internal/productmetrics` green on linux/amd64 (`CGO_ENABLED=0`, golang:1.26).
- [x] Dogfooded on a live city (v1.4.0 + backports lineage): full `go test ./...` `--- FAIL` set identical to the base commit; cut over and the polecat formula's submit step replaced by the verb. Live refusals with the real binary against a real bead and real worktrees — wrong branch, detached HEAD, unconfigured refinery — all exit 1 with the bead unchanged. Two real work beads then went through the new formula end to end: pushed by `gc submit`, merged and closed by the refinery.
- [ ] `make check` — not run locally (no Go toolchain on the operator host).
- [ ] `make test-integration` — not run.

## Checklist

- [x] Linked an issue (#6148)
- [x] Docs: `docs/reference/cli.md` regenerated with `gen-doc`
- [x] No breaking changes: new verb; no existing command or metadata contract changes. The formula change that adopts it lives in gascity-packs.

https://claude.ai/code/session_01PQru47R8EJ8UPk4rUvM8Ta


atbrace:
Closing this for now, unmerged. On reflection the verb as written carries the gastown pack's handoff vocabulary (bare `branch`/`target`/`branch_ready`/`halt_reason`, `polecat/<bead>` default, refinery reassignment) into core rather than a general mechanism. The ordering property — the bead write is unreachable unless the pushed ref was read back from origin — is the general part, and it would fit better on top of ADR-0009's `gc.work_branch`/`gc.work_commit` record with the pack supplying the handoff. We're keeping this local as a fork patch and leaving #6148 open as the problem statement; happy to revisit if a maintainer wants the ADR-0009-shaped version.

PR #6150, opened 2026-09-08
fix(dashboard): resolve a finished run's retired seats by session id (ga-3cs9p)

julianknutsen:
Bead: ga-3cs9p (P2)

## What

The dashboard run view rendered every retired seat of a finished run as a bare link with no session. The run tailer builds its session index from the default `GET /v0/city/{name}/sessions` listing, which is open-only by design (`sessionReadModelListings` never sets `IncludeClosed`; `session.Store.listAllBeads` forces `status=open`), so a step whose session had closed never found its session. On the assignee-only path this city's `gcg-session-<32hex>` ids were additionally rejected by `runproj`'s `sessionIDRe` (3-letter prefix, 40-char body), so those steps had no link at all.

## Why this design

Hard requirement: no performance regression. maintainer-city's graph store holds thousands of closed session beads, and a listing that scans them on every run-detail render or sessions poll is not acceptable. So the default `/sessions` listing is left exactly as it was (open-only, no `include_closed`), and the run's **own** links are resolved individually by durable id — bounded by link count and wall time, never by the store's closed population.

### Review round 2 (this push): the by-id read is now a point read on a MISS too

Review found that the first cut was a point read only on a HIT. `GET /v0/city/{c}/session/{id}` resolves an exact-id miss through the session target ladder (configured name, live, path alias, closed), and the closed steps `List` with `IncludeClosed` over metadata — an unindexed scan of every closed session bead (measured with a recording MemStore: **8 `store.List` calls per absent id**, 2 of them closed-inclusive). The BFF cached a 404 for only 5 s and did not cache a non-404 at all, so an id the store no longer holds (pruned session, stale stamp, slot label from the assignee fallback) re-ran those scans every 5 s per id, and a 409 (name matching two closed sessions) re-ran them on every SSE-driven rebuild. Fixed as follows:

- **api** — `SessionGetInput` gains an opt-in `exact_id` query flag. When set, `resolveSessionGetID` resolves via `session.ResolveSessionIDByExactID` (one `store.Get`, closed beads included) and a miss is a 404 straight away; the ladder never runs. No response-shape change; OpenAPI spec, Go client (`genclient`) and TS client regenerated. Default behaviour without the flag is unchanged.
- **dashboardbff** — the by-id read sends `exact_id=true`. Every outcome that is not a hit — 404, 409, 5xx, transport failure, decode failure, or a 200 naming a different id — is cached as a miss, so no failure mode can retry per render. Consecutive misses back off (`retiredMissTTL`: 5 s, doubling, capped at the 5 min hit TTL), so a permanently unresolvable id costs a handful of point reads in its first minute and one per 5 min after. One build issues by-id reads for at most `retiredSessionsBuildBudget` (5 s; overrun bounded by one read timeout) so a slow-but-answering supervisor cannot hold a `detail()` call — and the SSE loop behind it — for 64 read timeouts; ids past the budget keep their bare link for that build and the next build resumes from the cache. Non-404 failures are logged (bounded by the cache cadence).
- **runproj** — `sessionIDForLookup` now takes the same assignee fall-through as `runSessionLinkFor` when a durable stamp fails the id gate (previously the link path fell back but the lookup path returned `""`, so such a step got a link the retired enrichment could never name). The `{1,40}` body bound is documented as exactly `len("session-")+32` and pinned by a test.
- **dashboard SPA** — the unused TS `SESSION_ID_RE` (a stricter `gc|td|th|[a-z]{4}` alternation that would have re-gated `gcg-session-*` ids if a consumer ever appeared) is deleted along with its barrel export; comments that named it now point at the supervisor's Go gate. The embedded `dist/` is rebuilt (the baseline rebuild on the previous commit was byte-identical, so the dist diff is attributable to the generated-client and shared-source changes).

Non-blocking notes addressed: per-build time budget; lookup/link fallback divergence; `{1,40}` comment; TS `SESSION_ID_RE`; `retiredSessionsKey` doc (set-only key, display-field staleness accepted for a terminal record); `maxRetiredSessionLookups` doc (caps ids considered, cache hits included).

### First cut (unchanged)

- **runproj**
  - `SessionIDsForSnapshot(snap)` extracts the durable/derived session ids a run's steps reference (stamp first, else assignee-derived supervisor id; deduped; pending/ready steps and unroutable handles excluded).
  - `BuildRunDetailFromSnapshot` takes `RunSessions{Live, Retired}`. Retired sessions are indexed **by id only**, never by name/template, so a recycled pool-slot name can never attribute a closed step to the wrong session. Live wins an id collision.
  - `sessionIDRe` widened to `^[a-z]{2,4}-[a-z0-9-]{1,40}$`. It still rejects pool names, bare handles, uppercase, and over-long bodies.
- **dashboardbff**
  - `detail()` resolves the snapshot's ids that are missing from the open listing one read per distinct retired link, capped at 64 per build; ids in the open listing are never read; reads are skipped when the listing is unavailable.
  - Reads are cached in a manager-wide bounded LRU (cap 1024) keyed by (city, id), each value carrying its own deadline. The resolved id set joins the detail memo key.
  - The periodic `/sessions` poll is untouched — tests assert its query never carries `state=`/`closed`.

Out of scope (noted): the SPA's "Related entities" panel builds its own relation index from the open listing and still omits closed sessions; that is a separate, frontend-side consumer.

## Measurements (O(links), not O(closed beads))

`TestRunDetailRetiredSeatsResolveAgainstRealStore` (internal/api): cold run-detail read through the dashboard plane against the real API + MemStore; every by-id read is asserted to carry `exact_id=true`.

| closed session beads in store | retired links | exact-id reads | cold run-detail wall |
|---|---|---|---|
| 0 | 0 | 0 | 1.05 ms |
| 0 | 1 | 1 | 1.52 ms |
| 0 | 20 | 20 | 2.76 ms |
| 5000 | 0 (before) | 0 | 1.06 ms |
| 5000 | 1 | 1 | 1.21 ms |
| 5000 | 20 | 20 | 8.52 ms |

`TestSessionGetExactIDMissIssuesNoListing` (internal/api, MemStore with 2000 closed session beads): `exact_id=true` miss → 404 with **0** `store.List`; `exact_id=true` on a closed session's `session_name` → 404 with 0 List (ladder skipped); `exact_id=true` closed hit → 200 with 0 List; control: the same absent id without the flag → 404 with 8 List calls.

## Tests added / changed in this push

- `internal/api/run_detail_retired_sessions_test.go`: `TestSessionGetExactIDMissIssuesNoListing` (mirror of the HIT pin for the MISS path, with control); composition test now counts only `exact_id=true` reads and fails on any by-id read without it.
- `internal/api/dashboardbff/rundetail_retired_sessions_test.go`: the fake records by-id query strings and can force a status; `TestRunDetailResolvesRetiredSessionLinksByID` asserts `exact_id=true` on every read; new `TestRunDetailRetiredUpstreamFailureIsCachedForMissTTL` (409 and 500), `TestRunDetailRetiredTransportErrorIsCachedForMissTTL`, `TestRunDetailRetiredPersistentMissBacksOff`, `TestRunDetailRetiredLookupsStopAtBuildBudget`. Injected clock; no sleeps or listeners.
- `internal/runproj/detail_sessionlink_retired_test.go`: `TestSessionIDForLookupMirrorsLinkFallback` (seven shapes, lookup id == link id), `TestSessionIDGateAcceptsMintedShapes`.

## Verification

Round-2 commit `cb530bebb5` (branch `fix/ga-3cs9p-run-closed-sessions`):

- `go build ./...` — ok
- `go vet ./internal/api/... ./internal/runproj/...` — ok
- `go test -count=1 ./internal/api/ ./internal/runproj/` — ok (92.0 s; includes `TestOpenAPISpecInSync`, `TestGeneratedClientInSync`, `TestSessionGetExactIDMissIssuesNoListing`, `TestRunDetailRetiredSeatsResolveAgainstRealStore`)
- `go test -count=1 ./internal/api/dashboardbff/` — ok; `-v` run confirms all 10 retired-path tests executed
- `go test -race -count=1 ./internal/api/dashboardbff/ ./internal/runproj/` — ok
- `go test -count=1 ./internal/testpolicy/resourcecensus/` — ok (staged files)
- `go run ./cmd/genspec && go generate ./internal/api/genclient` — spec/client regenerated; `go run ./cmd/genschema` — no further drift
- `npm run generate:client` — generated TS diff is exactly the `exact_id` field
- `make dashboard-check` — ok (typecheck + build + dashboardspa/dashboardbff Go tests); a baseline `make dashboard-build` on the previous commit was byte-identical to the tracked `dist/`
- `make check-core-boundary` — OK
- `make lint-changed LINT_CHANGED_SCOPE=staged` — 0 issues
- pre-commit hook (branch's `.githooks`): format, lint, genspec/genclient/genschema, `make vet`, `make dashboard-check dashboard-smoke` — ok (exit 0)
- pre-push hook (branch's `.githooks`): push-ownership-guard + `make test-fast-parallel` — "All fast jobs passed" (exit 0)


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277313143

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277314883

quad341 (approved):
@julianknutsen — the round-2 push is what makes this one. Catching that your first cut was a point read only on a *hit*, and that a 404 or 409 would re-run the ladder's closed-inclusive scans every 5s per unresolvable id, is exactly the failure mode that reviews well and then quietly costs real CPU on maintainer-city. Making `exact_id` opt-in so the default resolution path is untouched is the right shape for it.

Particular appreciation for the control in `TestSessionGetExactIDMissIssuesNoListing`. A "0 store.List calls" assertion is worthless if the counter never fires, and you proved the flag-less read *does* List. That's the difference between a measurement and a wish.

Merging as-is. Two things I'm filing on our side, nothing for you to do: a pair of narrow races in the new retired-session cache (an expired-entry forget that isn't generation-scoped, and an in-flight read that can repopulate across a city rebind — both bounded and self-healing), and a check on whether the old stamp-vs-derived gate asymmetry was written against a real incident, since unifying the regex relaxes the byName fallback for short-prefix stores.


---

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR makes a finished run's retired seats resolve to named sessions instead of bare links — it reads each retired session by exact id rather than hoping it appears in an open-only listing, and it unifies the session-id gate so this city's `gcg-session-<32hex>` assignees stop falling off the fallback path entirely. That matters because a run's seat history is mostly what you look at *after* the run finishes, which is exactly when those links were going anonymous.

### Original PR Review

Decision: approve.

Specific gaps fixed:

- The stated bug held up under independent checking rather than being taken on trust: `sessionReadModelListings` never sets `IncludeClosed` and `listAllBeads` forces `Status: "open"`, so a finished run's retired seats genuinely could not be resolved from the listing. The fix resolves the run's own ids individually instead, and `TestRunDetailRetiredLookupsAreBoundedByLinks` pins the cost at O(links) against 5000 closed beads rather than leaving it asserted.
- The regex unification was probe-verified rather than assumed safe: the deleted `sessionBeadIDRe` was replayed against the new unified `sessionIDRe` over the full case set with **zero** divergence, so the provenance-trusted stamped path is byte-identical — only the fallback and suffix-extraction paths widened.
- `exact_id=true` routes to the pre-existing `ResolveSessionIDByExactID` on a hit, and `TestSessionGetExactIDMissIssuesNoListing` keeps a control row proving the flag-less miss still issues `List`. The asserted zero is a measured property of the new flag, not a dead counter.
- Cache behavior under failure was checked end to end: every non-hit outcome (404, 5xx, transport failure, decode failure, a 200 naming a different id) is cached, consecutive misses back off from 5s up to the 5-minute hit TTL, and a slow supervisor is bounded by a 5s per-build wall budget instead of 64 sequential read timeouts.
- The one uncertainty a reviewer left open — whether the codegen artifacts had actually been regenerated — was settled directly: `TestOpenAPISpecInSync` passes at the adopted head, `exact_id` is present in all six generated artifacts, and the deleted TS `SESSION_ID_RE` has no remaining references in the web tree.

Review findings addressed:

- **Change Impact / Blast Radius (minor):** widening `sessionIDRe` to `^[a-z]{2,4}-[a-z0-9-]{1,40}$` means `supervisorSessionIDFrom` returns a whole-value match *before* it consults the suffix regex, so a composite handle like `bd-x-gc-335812` is now taken whole instead of reduced to a trailing `gc-335812`. Commit `36313fd88f` documents that short-circuit on both `sessionIDRe` and `supervisorSessionIDSuffixRe` — extraction is the fallback for handles the gate rejects, not a minimizer applied to every value — and records why the divergence from `work-in-flight.ts`'s minimal-handle doctrine is deliberate (this gate must admit the internal hyphen of `session-<32 hex>`) and why the cost is bounded by the retired-session miss cache. Neither reviewer's suggested regex narrowing was applied: both were checked against their own cited examples and both turn out to be no-ops there.
- **Architectural Consistency (nit):** `SessionIDsForSnapshot` / `sessionIDForLookup` described their output as resolved ids, but `linkForSession` can overwrite `SessionID` with the index-matched session's own id. `36313fd88f` rewords both to candidate ids, pre-index resolution.
- **Test Evidence Quality (nit):** the comment on `TestRunSessionLinkForResolvesActiveDurableStamp` still described the two-regex trust model this PR deletes — true at base, falsified by the unification. `36313fd88f` rewords it to the single-gate model, where the stamp's force is its precedence over the fallback rather than a wider stamp-only regex.
- **Architectural Consistency (nit):** `lruSingleFlight.forget` sits beside `singleFlightCache.discard` on adjacent lines of the city-rebind path. `36313fd88f` documents why the verbs differ rather than renaming them: `discard` detaches the entry so a running compute can never repopulate the key, while `forget` leaves an in-flight build free to store under it. Same trigger, different guarantee.

- `ga-ytqa8`: add test rows (`bd-x-gc-335812`, `mc-worker-gc-335812`, `qa-agent-1`) pinning the widened gate's whole-value short-circuit, so that behavior is asserted rather than only documented. Live impact today is none — a census of this city's assignees found every live shape still falls through to suffix extraction and resolves correctly, because the `__` identity encoding fails the `[a-z]{2,4}-` prefix and any composite carrying a full `<store>-session-<32hex>` overruns the 40-char body bound.
- `ga-ytqa8`: make the retired-session cache's expiry identity-checked (`forgetIfExpired(key, observedExpires)`). `fetchRetiredSession`'s expire-then-rebuild is not atomic, so a losing racer's unconditional `forget` can drop a fresh entry it never observed. Bounded — one redundant point read, and at worst one render's bare link that self-heals on the next build — with no state corruption possible.

### Maintainer Changes

`36313fd88f docs(dashboard): document the unified session-id gate's whole-value short-circuit` — comment-only, with no executable code or test assertion changes: 3 files changed, 42 insertions(+), 6 deletions(-). The branch also carries `2990e050cc Merge latest main into adopted PR head` from the adoption gate.

### Final Review Status

Ready for the merge queue: final head `36313fd8` has passing required GitHub checks (CI, CodeQL, Cyclomatic complexity, Docs render check, Go Vulnerability Scan, Mac Regression, Review Formulas), and the merge-ready metadata is in place. `main` moved during finalization, so the adopted content was merged onto the latest base locally and re-validated there — `go build ./...`, `go vet`, and `go test ./internal/runproj/... ./internal/api/dashboardbff/...` all clean — with no change to this PR's own diff, so the green head above stands as reviewed. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

CI: https://github.com/gastownhall/gascity/actions/runs/34285754802

### Review Iterations

1 review pass performed. Three independent reviewers with full cross-file coverage converged on approve with no blockers and no majors; the comment-only halves of the surviving minor and nit findings were applied as the maintainer fixup above, and the two halves that would touch executable code were filed as `ga-ytqa8` rather than shipped without a further review pass.

---

PR #6151, opened 2026-09-08
Close stranded workflow roots after finalizer quarantine

quad341:
## What this changes

When a workflow's finalizer is quarantined, the workflow root now settles to a failed terminal state instead of remaining open and repeatedly surfacing as claimable work.

If the store refuses that root transition, Gas City preserves the finalizer quarantine and makes the incomplete settlement durable and observable: the root receives failure metadata, a reconciliation task is created, and a typed `control.root_settle_failed` event is fired. Operators can diagnose and recover the stranded root without relying on an ephemeral trace line.

## Review notes

- The control-dispatch path keeps finalizer quarantine as the load-bearing action; a follow-up root-close refusal does not reintroduce an unbounded dispatcher retry.
- Already-settled roots and non-finalizer control beads remain unchanged.
- The new typed event is registered through the existing event payload registry and projected through generated OpenAPI, Go client, TypeScript client, and dashboard assets.
- No config migration, storage migration, or CI workflow change is required.

## Test plan

- [x] Run the full local CI-equivalent union, including process-backed `cmd/gc` and integration shards.
- [x] Verify root settlement, non-finalizer isolation, already-settled preservation, refusal diagnostics, event registration, and payload round-trip tests.
- [x] Run policy, build, vet, dashboard generation/typecheck/build, and a live dashboard preview smoke test.
- [x] Review the release gate: [`release-gates/ga-zje19p-quarantined-finalizer-root-settle-gate.md`](release-gates/ga-zje19p-quarantined-finalizer-root-settle-gate.md)

Internal tracking: `ga-6nnqo2`.

---
- Related to #5729 — settles the workflow root to a failed terminal state when its workflow-finalize control bead is quarantined, so the stranded root stops surfacing as claimable work; it does not remove the finalizer's blocks edge or change hook/ready eligibility, so a root the store still refuses to settle is only made observable via root metadata, a reconciliation task, and a typed control.root_settle_failed event


quad341:
Thanks for this, @quad341 — there is clearly a lot of work here.

quad341:
Mayor disposition on the size skip above: this PR is reviewable and is not being split.

The skip tripped on bytes, not lines: the diff is about 1,549 lines but 1.59 MB, because it carries the rebuilt dashboard bundle (`internal/api/dashboardspa/dist/assets/index-*.js`, a 499 KB minified file replaced wholesale) plus the regenerated OpenAPI schema and generated client. The hand-written change is roughly 540 lines across `cmd/gc/cmd_convoy_dispatch.go`, its test, and the `internal/events` control payloads, with the release gate document alongside. That is well inside what the review process handles once the generated files are set aside.

The review is being re-run with the byte ceiling lifted so the ensemble reads the real change; the ceiling itself gets a follow-up so committed build output stops counting against it.

quad341:
@quad341 — the part of this I appreciate most isn't the root close, it's what you did with the refusal path. A stranded root used to vanish behind a comment promising a "later pass" that never existed; now it gets marker metadata, a reconciliation bead linking root and finalizer, and a typed `control.root_settle_failed` event so it surfaces next to `control.stalled`. Reusing the existing `refusingCloseStore` fixture in `TestQuarantineControlFailureBeadToleratesRootCloseFailure` is exactly right — it proves the finalizer quarantine still stands when the follow-up close fails, which is the invariant that actually matters.

Merging as-is. I traced the `outcome=fail` question out of the diff and it's fine: the `hard` failure class makes it terminal in both `classifyRetryAttempt` and the ralph loop, so nothing retries the settled root.

Three follow-ups I'll file on our side rather than hand back to you: the settle doesn't close spec sidecars or unreached members the way `processWorkflowFinalize` does; the reconciliation bead could use a label so it stays operator-facing; and its `gc.root_bead_id` quietly makes it a subtree member.


---

PR #6153, opened 2026-09-08
fix(beads): retry SQLite composite transactions on busy snapshots

### LINKED ISSUE #6152: bug: SQLiteStore.Tx does not retry transient busy snapshots
## Summary

`internal/beads.SQLiteStore.Tx` starts a deferred SQLite transaction but does not use the existing `retryOnBusy` loop used by the store's singleton writes. Under concurrent writers, a read-then-write transaction can fail immediately with `SQLITE_BUSY_SNAPSHOT` (reported as `database is locked (517)`), so composite GC operations fail even though equivalent single-row operations retry.

## Symptom

In a busy city using the pure-Go SQLite graph store, native GC operations that instantiate or update a formula can fail while clearing a claim fence:

```
database is locked (517)
```

The current `isSQLiteBusy` predicate already recognizes this error, but `SQLiteStore.Tx` returns it after its first attempt. The existing singleton paths retry three times with a 150 ms backoff.

## Reproduction

On current `main` (verified at `946363782`), run a focused internal test with a temporary SQLite store and two SQLite connections:

1. Seed one bead through `OpenSQLiteStore`.
2. Begin `SQLiteStore.Tx`; establish a read snapshot inside the callback.
3. From the second connection, update and commit another row.
4. Have the first callback perform its first write.

The deferred transaction has a stale read snapshot, so the first write returns `database is locked (517)` (`SQLITE_BUSY_SNAPSHOT`). A retry of the callback from a fresh transaction succeeds.

Expected: `SQLiteStore.Tx` retries the complete callback on a fresh transaction and succeeds, preserving all-or-nothing semantics.

Actual: `SQLiteStore.Tx` returns the busy error after one attempt.

## Root cause

The invariant is that transient SQLite write contention is retried by the store's write paths (`internal/beads/sqlite_store.go:161-168`). `SQLiteStore.Tx` violates it at `internal/beads/sqlite_store.go:1627-1650`: it calls `BeginTx` once, executes the callback, and commits without passing the transaction body through `retryOnBusy`.

The composite callback performs read-then-write work internally—for example, `sqliteStoreTx.Create` reads for duplicate detection before inserting (`internal/beads/sqlite_store.go:1606-1621`). A concurrent commit can therefore produce `SQLITE_BUSY_SNAPSHOT`; SQLite's busy handler cannot wait out that extended result code. `isSQLiteBusy` already matches the resulting `database is locked` text (`internal/beads/sqlite_store.go:153-155`).

`SQLiteStore.AtomicTx` deliberately reports a real transaction (`internal/beads/sqlite_store.go:1652`), so the fix must retry the whole callback with a fresh transaction rather than split writes or silently continue.

## Fix candidates

**A. Retry the complete `SQLiteStore.Tx` body with `retryOnBusy` (recommended)** — move the begin/callback/commit body into the existing bounded retry helper.

- Scope: one production function and one focused regression test in `internal/beads`.
- Trade-off: the callback must remain store-only and safe to re-run after a transient rollback; this matches the existing transaction contract and preserves the current bounded retry policy.

**B. Begin composite transactions as `IMMEDIATE`** — acquire the write reservation before the callback can establish a stale read snapshot.

- Scope: SQLite transaction-begin plumbing plus contention tests.
- Trade-off: it may avoid `SQLITE_BUSY_SNAPSHOT`, but it serializes every composite transaction earlier, increases wait time for readers/writers, and still needs bounded handling when the reservation itself is busy.

**Recommendation:** A, because it reuses the store's existing retry policy, retries from a clean transaction as SQLite requires, and keeps the change local to the missing retry boundary.

## Adjacent / out of scope

- `DeleteIfMatch`, graph apply, session-bead create/close, and API idempotency also use composite transaction paths and should be covered by the same regression or audited for retry coverage.
- The SQLite numeric-code classifier in the external `bd` adapter is a separate concern; it does not control `SQLiteStore.Tx`, whose local predicate already recognizes the error.
- This is not a Dolt listener timeout issue: the failure occurs in the embedded SQLite graph store before a Dolt query is sent.

jacobhausler:
Closes #6152

## Summary

`SQLiteStore.Tx` is the composite write path used by formula instantiation and
other multi-write operations. It opened one deferred SQLite transaction but did
not use the store's existing `retryOnBusy` helper. Under concurrent writes,
SQLite can return `SQLITE_BUSY_SNAPSHOT` (`database is locked (517)`) immediately;
the busy timeout cannot resolve that snapshot.

## Change

- Run the complete `SQLiteStore.Tx` callback through `retryOnBusy`.
- Each retry begins a fresh transaction, so the callback re-reads current state
  after the failed snapshot and preserves all-or-nothing semantics.
- Document that transaction callbacks must keep side effects inside `tx`, since
  a transient busy result may replay the callback.
- Add a deterministic two-connection regression test that creates a read
  snapshot, commits a competing write, and verifies the first attempt fails,
  the second succeeds, and only the successful write remains.

This is the existing native retry primitive; no sling wrapper, daemon knob, or
new local layer is introduced. The accepted beads/Dolt contract remains the
boundary: `engdocs/design/beads-dolt-contract-redesign.md`.

## Verification

- `go test -count=10 ./internal/beads -run '^TestSQLiteStoreTxRetriesBusySnapshot$'`
- `go test -count=1 ./internal/beads/...`
- `make build`
- `make check` (including lint, vet, boundary checks, and observable tests)

All Go commands ran through the Westlands Nomad builder wrapper.

### DIFF
--- a/internal/beads/sqlite_store.go
+++ b/internal/beads/sqlite_store.go
@@ -1624,6 +1624,8 @@ func (s *SQLiteStore) GetLocalString(id, key string) (string, error) {
 // Tx executes fn in one SQLite transaction. Every write is rolled back when
 // the callback returns an error, so callers can safely compose Create, Update,
 // metadata updates, and Close as one all-or-nothing operation.
+// A transient busy result retries the complete callback in a fresh transaction;
+// callers must keep side effects inside tx because fn may run more than once.
 func (s *SQLiteStore) Tx(_ string, fn func(tx Tx) error) error {
 	if err := s.ensureOpen(); err != nil {
 		return err
@@ -1631,19 +1633,21 @@ func (s *SQLiteStore) Tx(_ string, fn func(tx Tx) error) error {
 	if fn == nil {
 		return errors.New("beads tx: nil callback")
 	}
-	ctx := context.Background()
-	tx, err := s.db.BeginTx(ctx, nil)
-	if err != nil {
-		return fmt.Errorf("sqlite tx: begin: %w", err)
-	}
-	defer tx.Rollback() //nolint:errcheck
-	if err := fn(&sqliteStoreTx{store: s, ctx: ctx, tx: tx}); err != nil {
-		return err
-	}
-	if err := tx.Commit(); err != nil {
-		return fmt.Errorf("sqlite tx: commit: %w", err)
-	}
-	return nil
+	return retryOnBusy(func() error {
+		ctx := context.Background()
+		tx, err := s.db.BeginTx(ctx, nil)
+		if err != nil {
+			return fmt.Errorf("sqlite tx: begin: %w", err)
+		}
+		defer tx.Rollback() //nolint:errcheck
+		if err := fn(&sqliteStoreTx{store: s, ctx: ctx, tx: tx}); err != nil {
+			return err
+		}
+		if err := tx.Commit(); err != nil {
+			return fmt.Errorf("sqlite tx: commit: %w", err)
+		}
+		return nil
+	})
 }
 
 // AtomicTx reports that Tx uses a real SQLite transaction and rolls all
--- a/internal/beads/sqlite_store_test.go
+++ b/internal/beads/sqlite_store_test.go
@@ -114,6 +114,83 @@ func TestSQLiteStoreTxRollsBackEveryWriteOnCallbackError(t *testing.T) {
 	}
 }
 
+func TestSQLiteStoreTxRetriesBusySnapshot(t *testing.T) {
+	dir := t.TempDir()
+	opened, err := OpenSQLiteStore(dir)
+	if err != nil {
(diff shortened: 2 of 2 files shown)


jacobhausler:
Related but distinct: the sling rollback report is now tracked in #6267. Your native SQLiteStore.Tx retry covers the composite transaction path; #6267 covers the BdStore transient-write classifier and the existing synthetic input-convoy cleanup that previously swallowed a 517 failure.

jacobhausler:
Reviewer follow-up: Gate 0 corrected the scope. The cr-bizoxz report reaches the native SQLiteStore.Tx busy-snapshot path covered by #6152/#6153; #6267 was closed as a duplicate. Please treat this PR as the canonical upstream twin for that report.

PR #6158, opened 2026-09-08
fix(api): dispatch agent output transcript on the provider family, not the config name

julianknutsen:
## What

`GET /v0/city/{city}/agents/{name}/output` and its SSE stream passed the agent's configured provider **name** (e.g. `glm53`, `kimi-k3-manifold`) to `DiscoverTranscript` / `ReadTranscript`. `sessionlog` dispatches on that string, so a custom-named zcode seat fell through to the Claude slug discovery (no file found, then the peek-fallback `Format: "text"` or a 404 "not running"), and any name containing `kimi` was routed to the Kimi reader even when its builtin was claude.

This is the same defect PR #6144 fixed for `SessionHandle.historyProvider` at the worker boundary, and it is fixed here for the same reason: a custom name carries no family signal. The two ladders are deliberately not identical, though — `historyProvider` leads with the session `Profile` and falls back to the bead's recorded provider, while this helper takes no `Profile` rung and falls back to the config chain, because a stamp-less bead's recorded name can be the misleading alias this fix repairs.

## How

`resolveAgentTranscript` (`internal/api/handler_agent_output.go`) is the single funnel every consumer reads `provider` from — the one-shot read (`trySessionLogOutputHuma`), the Huma stream (`resolveAgentStream`), and the legacy stream handler — so the fix lives there. A new `agentTranscriptProvider(info, providerName, cfg)` helper resolves the builtin family first:

1. session bead `builtin_ancestor`, then `provider_kind` (`session.Info.BuiltinAncestor` / `.ProviderKind`) when a session bead exists;
2. else `config.BuiltinFamily(name, cfg.Providers)` over the config chain (covers `base = "builtin:..."` descendants and legacy command-matched aliases);
3. only a name with no resolvable family falls through unchanged (bare builtins, fully custom providers).

`resolveAgentTranscript` sets `state.provider` from the config chain before the session lookup and re-resolves with the bead's `Info` once the catalog returns it, so discovery and the reader always agree. The legacy stream handler's duplicate name-based `provider` computation is removed; it only ever consumed `resolveAgentTranscript`'s value.

Wire types are unchanged.

## Tests (RED first)

All new tests were observed failing against a name-only stub before the fix (table test returned the name; `glm53` read degraded to `Format: "text"`; stream emitted no turn; kimi-named seat 404'd), then passing after.

`internal/api/handler_agent_output_provider_test.go`:
- `TestAgentTranscriptProviderPrefersFamilyOverName` — precedence matrix (bead ancestor/kind > config chain > name; blank/absent kinds fall through; bare builtin and fully custom names keep themselves)
- `TestAgentTranscriptProviderNilConfigFallsBackToName`
- `TestAgentOutputReadsCustomZCodeProviderTranscriptByFamily` — provider named `glm53`, bead stamped `provider_kind=zcode`, config entry with no resolvable family: reads the zcode mirror through the zcode reader
- `TestAgentOutputStreamReadsCustomZCodeProviderTranscriptByFamily` — same seat via the SSE stream
- `TestAgentOutputReadsClaudeFamilyTranscriptDespiteKimiName` — `kimi-k3-manifold` with `base = "builtin:claude"` and no session bead: config chain routes to the Claude reader
- `TestAgentOutputBareBuiltinProviderKeepsWorking` — bare `claude` unchanged

## Verification

- `go build ./...` — ok
- `go vet ./internal/api/... ./internal/worker/...` — ok
- `gofmt -l internal/api/` — clean
- `go test ./internal/api/... -count=1` — ok (api 94.6s, apierr, dashboardbff, dashboardspa, genclient)
- `go test ./internal/worker/... -count=1` — ok (worker, adapters/zcode, builtin, fake, transcript, workertest)
- `go test ./internal/api/ -v -run 'TestAgentTranscriptProvider|TestAgentOutputReads...|TestAgentOutputStreamReads...|TestAgentOutputBareBuiltin...|TestOpenAPISpecInSync'` — all PASS
- pre-commit hook: lint-changed 0 issues, docs generation clean, `go vet ./...` ok
- pre-push hook: `unit-core` + `unit-cmd-gc` 1-6 all ok ("All fast jobs passed")

Bead: ga-j72g7

### DIFF
--- a/internal/api/handler_agent_output.go
+++ b/internal/api/handler_agent_output.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"github.com/gastownhall/gascity/internal/config"
+	"github.com/gastownhall/gascity/internal/session"
 	workdirutil "github.com/gastownhall/gascity/internal/workdir"
 	"github.com/gastownhall/gascity/internal/worker"
 )
@@ -51,10 +52,11 @@ func (s *Server) resolveAgentTranscript(name string, agentCfg config.Agent) (*ag
 	}
 
 	state.sessionName = agentSessionName(s.state.CityName(), name, cfg.Workspace.SessionTemplate)
-	state.provider = strings.TrimSpace(agentCfg.Provider)
-	if state.provider == "" {
-		state.provider = strings.TrimSpace(cfg.Workspace.Provider)
+	providerName := strings.TrimSpace(agentCfg.Provider)
+	if providerName == "" {
+		providerName = strings.TrimSpace(cfg.Workspace.Provider)
 	}
+	state.provider = agentTranscriptProvider(nil, providerName, cfg)
 	state.workDir = s.resolveAgentWorkDir(agentCfg, name)
 	if state.workDir == "" {
 		return state, nil
@@ -79,6 +81,7 @@ func (s *Server) resolveAgentTranscript(name string, agentCfg config.Agent) (*ag
 			if catalog, err := s.workerSessionCatalog(store); err == nil {
 				if info, err := catalog.Get(state.sessionID); err == nil {
 					state.sessionKey = strings.TrimSpace(info.SessionKey)
+					state.provider = agentTranscriptProvider(&info, providerName, cfg)
 				}
 			}
 		}
@@ -247,4 +250,30 @@ func workerHandleRunning(ctx context.Context, handle interface {
 		return false, err
 	}
 	return state.Phase != worker.PhaseStopped && state.Phase != worker.PhaseFailed, nil
+}
+
+// agentTranscriptProvider resolves the provider string transcript discovery
+// and reads dispatch on. The session bead's family metadata wins
+// (builtin_ancestor, then provider_kind) because a custom provider's name
+// carries no family signal ("glm53" is a zcode seat) or a misleading one
+// ("kimi-k3-manifold" is a claude seat). Without a bead, or on a bead stamped
+// before family metadata existed, the config chain's builtin ancestor
+// (config.BuiltinFamily) takes over. Only a name with no resolvable family
+// falls through unchanged. Same precedence the worker boundary applies in
+// SessionHandle.historyProvider, so the file found and the reader used agree.
+func agentTranscriptProvider(info *session.Info, providerName string, cfg *config.City) string {
+	if info != nil {
+		if family := strings.TrimSpace(info.BuiltinAncestor); family != "" {
+			return family
+		}
+		if kind := strings.TrimSpace(info.ProviderKind); kind != "" {
+			return kind
+		}
+	}
+	if cfg != nil {
+		if family := strings.TrimSpace(config.BuiltinFamily(providerName, cfg.Providers)); family != "" {
+			return family
+		}
+	}
+	return providerName
 }
--- a/internal/api/handler_agent_output.go
+++ b/internal/api/handler_agent_output.go
@@ -259,8 +259,15 @@ func workerHandleRunning(ctx context.Context, handle interface {
 // ("kimi-k3-manifold" is a claude seat). Without a bead, or on a bead stamped
 // before family metadata existed, the config chain's builtin ancestor
 // (config.BuiltinFamily) takes over. Only a name with no resolvable family
-// falls through unchanged. Same precedence the worker boundary applies in
-// SessionHandle.historyProvider, so the file found and the reader used agree.
+// falls through unchanged. PR #6144 gave SessionHandle.historyProvider a
+// provider_kind rung ahead of the recorded provider name for the same reason,
+// but the two ladders are deliberately not identical: historyProvider leads
+// with the session Profile and falls back to the bead's recorded provider,
+// while this helper takes no Profile rung and falls back to the config chain
+// because a stamp-less bead's recorded name can be the misleading alias this
+// fix repairs. Within this API path the resolved string single-sources both
+// DiscoverTranscript and ReadTranscript, so the file found and the reader
+// used agree.
 func agentTranscriptProvider(info *session.Info, providerName string, cfg *config.City) string {
 	if info != nil {
 		if family := strings.TrimSpace(info.BuiltinAncestor); family != "" {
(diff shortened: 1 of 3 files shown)


quad341 (dismissed):
@julianknutsen — merging this as-is. The RED-first discipline is what made the review fast: stubbing `agentTranscriptProvider` back to `return providerName` on your branch drops all four end-to-end tests plus eight table cases, then restoring turns them green. That's the difference between tests that document a fix and tests that would catch its regression, and yours are the second kind.

Putting the resolution in `resolveAgentTranscript` rather than patching each consumer was the right call, and re-resolving with the bead's `Info` *after* the catalog lookup — so discovery and the reader can't disagree about family — is the detail that actually makes the stream path work. Deleting the stream handler's duplicate name-based computation instead of leaving it to drift was the right instinct too.

One follow-up, not a change request: your description and the new doc comment say this mirrors what #6144 did for `SessionHandle.historyProvider`, but at this head that function still returns `info.Provider` raw with no family resolution. Either it landed on a different symbol or the worker-side instance is still open — can you point me at #6144's actual change? If the worker boundary is still name-dispatching, that deserves its own bead, and you clearly already have the map for it.

Thanks for the tight scope — three files, wire types untouched.


---

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277314822

julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277315927

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR fixes a real misroute on `GET /v0/city/{city}/agents/{name}/output` and its SSE stream: both handed `sessionlog` the agent's *configured provider name* instead of its builtin family, so a custom-named zcode seat like `glm53` fell through to Claude slug discovery and degraded to a `Format: "text"` peek fallback (or a 404 "not running"), and any name containing `kimi` was routed to the Kimi reader even when its builtin was `claude`. That matters because provider aliasing is a first-class feature here — every custom-named seat had a silently broken transcript surface, failing soft with no error to point at.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **The precedence-parity claim in the helper's doc comment was false against the merge target.** The comment, and its mirror in the new test file, asserted that `agentTranscriptProvider` applies the "same precedence" as `SessionHandle.historyProvider`. Post-#6144 `historyProvider` is `Profile → info.ProviderKind → info.Provider → session.Provider`, while this helper is `BuiltinAncestor → ProviderKind → config.BuiltinFamily(current name) → raw name` — only the `provider_kind` rung actually matches. Both comments now state the real precedence and why the two ladders deliberately diverge. The code was correct throughout; the comment would have misled the next person diffing them.
- **The "so the file found and the reader used agree" claim was kept, but scoped to where it holds.** It does not hold across the API/worker boundary, but it does hold inside this API path: `resolveAgentTranscript` single-sources `state.provider` into both `DiscoverTranscript` and `ReadTranscript`, and into the stream reader. That was verified before rewording, rather than deleting a true statement.

Review findings addressed:
- Architectural Consistency / maintainability (minor, raised independently by all three review lanes): the false precedence-parity claim in `internal/api/handler_agent_output.go` and `internal/api/handler_agent_output_provider_test.go` — corrected in `5b822c9`. The same clause in the PR description was corrected in place.

One note on the review itself, since it affected the wording: two of the three lanes reached the right conclusion about that finding from the *wrong* premise. They read PR #6144 as an unmerged sibling because they tested ancestry of its pre-merge branch commit `8e9cfb7c`, which is not on `main`; it landed as `6d7463758a`. The corrected text deliberately does not describe #6144 as unmerged.

The premise behind your fix was reproduced independently rather than accepted on trust: a `go test -overlay` probe stubbing `agentTranscriptProvider` back to raw-name dispatch turned the zcode seat's response into `Format: "text"`, emitted no `event: turn` on the stream, and 404'd the kimi-named claude seat. Your six new tests exercise both the one-shot and SSE paths through the real HTTP handler, and the bare-builtin case stayed green under that stub — so it is a genuine non-regression control, not a vacuous pass. Test evidence was the strongest axis of the review.

- `ga-hfzkh` (minor, correctness): `agentTranscriptProvider` skips `info.Provider`, so a stamp-less session bead under config drift resolves by the current name. Adding a `config.BuiltinFamily(info.Provider, cfg.Providers)` rung is a production precedence change plus new test coverage, so it was deliberately left out of this PR.
- `ga-4bpfq` (minor, pre-existing and outside this diff): the `/sessions` model-context path still name-dispatches stamp-free alias beads, and `enrichSessionMeta` runs Claude-shaped `TailMeta` over zcode mirrors.
- `ga-zqe5l` (nit): consolidate the four provider-family resolution ladders (`ProviderFamilyFromMetadata`, `ProviderFamilyFromInfo`, `historyProvider`, `agentTranscriptProvider`), and re-resolve the SSE stream's once-captured provider so reader dispatch always matches the discovery that produced the path.

### Maintainer Changes
One maintainer commit on top of your work: `5b822c9 review: correct the historyProvider parity claim in agentTranscriptProvider` — 2 files changed, 14 insertions(+), 4 deletions(-). Comment-only; no behavior change.

### Final Review Status
Ready for the merge queue: final head `5b822c9` has passing required GitHub checks, and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

`main` moved 8 commits during the review, so the adopted branch was re-validated against the current base before handoff: the rebase is conflict-free with an unchanged patch-id, none of those 8 commits touch the 3 files here, and on the refreshed tree `gofmt`, `git diff --check`, and `go vet ./internal/api/` are clean with all six new tests and the full `internal/api` package green. The pushed head already carries exactly that content, so it was left in place rather than force-pushed — that would have invalidated the green checks for a zero-content-delta rebase.

CI: https://github.com/gastownhall/gascity/actions/runs/34304158377

### Review Iterations
1 review pass across three lanes (Claude, Codex, Gemini), plus one maintainer fixup. The pass returned no blockers and no majors — three minor findings and one nit. One minor was fixed here; the rest are filed as beads above.

---

PR #6159, opened 2026-09-08
fix(zcode): seat-scope follow-ups — restart-gated adoption, legacy sweep, byte-wise scopes, memoized State activity (ga-6a0u2)

julianknutsen:
Stacked on #6145 (`fix-zcode-transcript-seat-scope`). **Retarget to `main` once #6145 merges** — the base here is the #6145 head (d066aea05e) so the diff shows only the four follow-ups.

Implements the four reviewer follow-ups on #6145 (seat-keyed zcode transcript mirror / sid store). One commit per follow-up; each has its own owning test.

## What / why

**1. Adopt name-only state only on a restarted seat** (`a054be2f0c`)
`adopt_name_only` moved the previous adapter's name-only sid/mirror/CLI-HOME under the seat key on the first seat-keyed start, even for a brand-new bead. A fresh seat re-seated into a closed sibling's slot shares the name and epoch, so it `--resume`'d the dead seat's conversation and the dead seat's transcript was rekeyed under the wrong bead. gc exports `GC_RUNTIME_EPOCH` as `1` on a seat's first start (`session.DefaultGeneration`, `internal/session/manager.go`) and bumps it on every wake (`cmd/gc/session_wake.go` `preWakeCommit`), so the adapter now adopts only when the generation is above one. A fresh seat starts fresh and the closed sibling's bead keeps resolving its transcript through the name-only fallback scope.

**2. Sweep the previous adapter's name-only state** (`5a08394088`)
The superseded-epoch sweeps glob the seat key, so name-only entries at another epoch (`sids/<name>#3`, `homes/<name>#3`, live-root mirror `<name>#3`) were never swept, and neither was the same-epoch state a fresh seat now declines. A seat-keyed start sweeps every name-only entry of its own session name after adoption has had its chance: sids and CLI homes are dropped, mirrors are **archived** under their own scope (never deleted) so the writing bead still resolves them via the archive search root. Seats with no bead id (keyed by the name itself) skip it. One-time in effect: nothing is left under the name afterwards.

**3. Byte-wise sanitization on every side** (`5dafb9874d`)
`sanitizeZCodeComponent` walked runes; the adapter's `tr -c` walks bytes, so `"wörker"` folded to `w_rker` vs `w__rker`. The Go side now walks bytes; the adapter pins `LC_ALL=C tr` so a locale-aware `tr` cannot widen to characters; the mirror helper's Python sanitizer (session-id file names, pending placeholder id) folds UTF-8 bytes the same way.

**4. Memoized derived tail activity across State polls** (`806e35fb52`)
zcode derives activity from the whole mirror (no tail chunk), so a keyless zcode `State()` re-read and re-normalized the entire JSON mirror on every API poll through a fresh per-request handle. `SessionLogAdapter` now carries a derived-activity memo keyed by mirror path and fenced by mtime+size (the same generation id the history snapshot already stamps), shared by every adapter/handle a `Factory` builds. Unchanged mirror = one parse; atomic per-turn rewrite = one more; failures never cached; the zero adapter (nil memo) derives every call. `State()` routes derived-activity families through the memoized probe; keyless tail-chunk families keep the history-backed probe (their transcripts, e.g. `~/.codex/sessions`, can live outside the adapter's fenced search roots).

## Tests added

- `internal/worker/adapters/zcode/adapter_test.go` (real bash adapter + stub CLI): `TestFreshSeatDoesNotAdoptAClosedSiblingsNameOnlyState`, `TestSeatSweepsStaleNameOnlyStateOfItsName`, `TestNonASCIISessionNameSanitizesByteWiseOnBothSides`; `TestSeatAdoptsStateLeftUnderTheNameOnlyScope` now runs the post-upgrade start as a restart (`GC_RUNTIME_EPOCH=2`).
- `internal/sessionlog/zcode_reader_test.go`: `TestZCodeScopeSanitizesByteWiseLikeTheAdapter`.
- `internal/worker/activity_memo_test.go`: `TestDerivedActivityMemoDerivesOncePerGeneration`, `TestDerivedActivityMemoDoesNotCacheFailures`, `TestDerivedActivityMemoNilDerivesEveryCall`.
- `internal/worker/handle_test.go`: `TestSessionHandleStateReusesDerivedActivityAcrossPolls` (three per-request handles poll an unchanged mirror: one parse; rewrite: one more, idle).

Each test was observed RED for the intended reason before the fix.

## Verification

- `go build ./...` — ok
- `go vet ./internal/worker/... ./internal/sessionlog/... ./internal/session/...` — ok
- `go test ./internal/worker/... ./internal/sessionlog/... ./internal/session/...` — all ok (worker 12.8s, adapters/zcode ~25.7s, session 1.5s)
- `go test ./cmd/gc -run TestGCNonTestFilesStayOnWorkerBoundary` — ok
- `go test ./internal/api/ -run 'AgentOutput|SessionsStream|SessionState|Sessions|Agents'` — ok
- pre-commit hook (gofmt, `make lint-changed`, `make vet`) ran on each commit; pre-push ran `make test-fast-parallel`.
- No `internal/api` wire types touched; `TestOpenAPISpecInSync` / `make dashboard-check` not applicable.

Bead: ga-6a0u2


julianknutsen:
Review follow-ups pushed as two commits on top of the reviewed tip (806e35fb52 → 068e49c8b8).

**Blocking — follow-up (4) not delivered in production** → `b125695657` perf(api): hold the derived-activity memo on the Server, above per-request factories
- `worker.FactoryConfig.ActivityMemo` (new, `*worker.DerivedActivityMemo`, exported with a `Derivations()` counter). `newFactory` now takes the config; nil still yields a private memo, so single-factory callers are unchanged.
- `api.Server.activityMemo` is created in `newServer` (alongside `lookPathCache` / the response cache) and threaded into every `workerFactory()` build, so `workerHandleForSession`, `sessionRuntimeHandles`, the agent-output read and both stream prechecks share one memo across requests.
- `TestSessionHandleStateReusesDerivedActivityAcrossPolls` now builds each poll through a fresh `NewFactory(cfg)` + `SessionByID`, as the API does, and its `history == nil` assertion is on a polled handle. New `TestFactoryThreadsInjectedActivityMemo` (worker) and `TestWorkerFactorySharesDerivedActivityMemoAcrossRequests` (api: three `srv.workerFactory()` requests, one parse). All three were RED before the wiring (0 derivations on the shared memo) and GREEN after.

**Non-blocking** → `068e49c8b8` fix(zcode): merge into an occupied archive slot, fold the harness byte-wise, note the fresh-seat transient
- (2) `archive_scope` no longer `rm -rf`s an occupied archive slot; it merges into it (`cp -a "$1/." "$dest/"`), so the earlier bead's transcript survives the name-only sweep. Owning test `TestArchiveMergesIntoAnAlreadyArchivedNameOnlyScope` was RED (earlier transcript destroyed) then GREEN. That also removes the SC2115 site; `shellcheck` on the script is now clean. The `mv … || rm -rf "$1"` fallback is kept deliberately and documented: a stale scope left beside the fresh one is the leak the archive exists to stop — flagging as the one remaining "delete" path if you'd rather it be left in place.
- (1) `FindZCodeSessionFileByScope` comment now documents the fresh-seat transient (empty seat scope resolves to the sibling's name-only mirror until the first write).
- Harness `pendingSessionID()` folds byte-wise like the three writers; `regexp` import dropped.

Verification on 068e49c8b8: `go build ./...` ok; `go vet ./internal/worker/... ./internal/api/... ./internal/sessionlog/...` ok; gofmt clean; `go test ./internal/worker/... ./internal/sessionlog/... ./internal/session/...` ok (worker 12.6s, adapters/zcode 25.8s); `go test ./internal/api/` ok (92.5s, full package); `go test ./cmd/gc/ -run TestGCNonTestFilesStayOnWorkerBoundary` ok; pre-push `make test-fast-parallel` all fast jobs passed.

quad341 (approved):
@julianknutsen — this pair is a model of how to split a fix. #6145 does one thing (key zcode's per-conversation state by the seat instead of the session name), and #6159 is the review round turned into four owning regressions rather than four assertions. Both are landing.

The root-causing in #6145 is what makes it easy to check. You pinned it to an actual observed collision — `beads--gc__implementation-reviewer-1-pool#1` carrying two seats of one run — instead of a hypothesis, and then picked a key that was already there: `GC_SESSION_ID` is exported, survives a restart, and needed no new plumbing. The detail I want to call out is the `@` delimiter argument. Choosing a character that *cannot* survive `tr -c 'A-Za-z0-9._-'` is what makes the per-seat sweep globs provably non-overlapping rather than probably; we checked both sanitizers and they agree exactly. And moving all three artifacts in `adopt_name_only` — including the CLI `HOME`, because that's where `--resume` actually reattaches — is the kind of thing that's easy to half-do and miserable to debug six weeks later.

In #6159, the follow-up that impressed us most is the one you caught on yourself: the memo was inert in production while it lived on the per-request `Factory`, and you moved it to the `Server` and then wrote `TestWorkerFactorySharesDerivedActivityMemoAcrossRequests` so it stays found. A memo on an object that's rebuilt per request reviews green and does nothing; that's a genuinely hard failure mode to notice in your own work. The restart gate on adoption is also the right call in the right direction — a legacy seat restarted without a generation bump loses continuity once rather than risking a resume into a dead sibling's conversation, and your comment says exactly that rather than leaving us to derive it.

Here's the plan. #6145 merges as-is, first. Then we retarget #6159 to `main`, rebase it, and take on two small things on our side before it lands — you asked about the remaining delete-on-failure path in `archive_scope`, and the answer is yes, that's worth closing: the occupied-destination branch currently `rm -rf`s the live scope after a `cp -a` that may have failed, so an ENOSPC or a permission error costs the only copy of a transcript. We'll guard it, add a fault-injection test, and re-run the worker / sessionlog / session / api suites. Your commits and authorship stay intact throughout; the guard goes in as a separate maintainer commit.

Two things we're filing on our side rather than sending back to you, both because they're design questions and not defects in this work: per-seat sweeps mean a *retired* seat's mirror and CLI HOME are never reaped, so they accumulate in the live root — that's the correct trade for this fix, since a sibling may still be running, it just needs a separate collector that knows when a seat is actually dead. And the transient window where a fresh seat reads a closed sibling's name-only mirror until its first write is real, but you documented it honestly in the code comment, and closing it means teaching the reader to tell a legacy bead from a fresh seat, which it structurally can't today.

Thanks for the RED-first discipline on every one of these tests, and for naming the coverage you *couldn't* get (`test-worker-inference` needs credentials CI doesn't have) instead of quietly leaving it out. That's what made both of these reviewable.

quad341 (dismissed):
@julianknutsen — this is two PRs of unusually careful work and we're taking both as authored.

On #6145, the thing that made it easy to check is that you pinned it to an actual observed collision — `beads--gc__implementation-reviewer-1-pool#1` carrying two seats of one run — rather than a hypothesis. Picking `GC_SESSION_ID` as the key was the right call: already exported, survives a restart, no new plumbing. And the `@` delimiter argument is the detail I keep coming back to. Choosing a character that *cannot* survive `tr -c 'A-Za-z0-9._-'` is what makes the per-seat sweep globs provably non-overlapping instead of merely probably; we checked both sanitizers and they agree exactly. Making `adopt_name_only` move all three artifacts including the CLI HOME — because that's where `--resume` actually reattaches — is the kind of thing that's easy to half-do and painful to debug six weeks later.

On #6159, the standout is follow-up (4), and specifically that you caught it on yourself: a memo living on a per-request `Factory` is inert in production, reviews green, and does nothing. You found it, moved it to the `Server`, and then wrote a test asserting one parse across three `srv.workerFactory()` builds so it can't quietly move back. That's a hard failure mode to see in your own work. The byte-wise fix is the other one — three sanitizers in three languages is exactly where a fix lands on two and the third drifts, and asserting that the *reader resolves the adapter's actual path* rather than just that both fold the same is what makes it stay fixed. You'd also already closed the `archive_scope` delete-on-failure path a reviewer offered to take on, with chmod-0500 fault injection to prove it.

Landing order: #6145 goes in first, then we retarget #6159 to `main`, rebase, and re-run the worker / sessionlog / session / api suites. Your commits stay as authored on both — the commit-per-follow-up split on #6159 is what made it reviewable, so we won't squash it. #6159 also carries `blocked-on-deploy-gate`, so it waits on `release-gate/deploy-clearance` on the rebased head; that's a release gate on our side, nothing for you to do.

Two follow-ups we're filing on our side, not yours:

1. **A seat reaper.** Per-seat sweeps are correct — a sibling may still be running — but that means a *retired* seat's mirror and CLI HOME are now never reaped and accumulate in the live root. That needs a collector that knows when a seat is actually dead, which is a design question rather than a mechanical fix, so it correctly stayed out of this stack.
2. **Two narrow races**, both bounded and neither worth holding these for: the three `mv`s in `adopt_name_only` aren't a single ownership decision, so two same-name seats starting concurrently across the upgrade boundary could split the legacy artifacts (strictly no worse than today, where all seats already share them); and the name-only sweep in #6159 has no liveness check, so we want to confirm the zcode pane replacement is restart-all rather than rolling before the deploy gate clears.

One optional nit if you're touching the file anyway: `archive_scope`'s occupied-destination branch has no `could not archive` warning — the printf lives only in the `mv` branch — so a failed `cp -a` leaves a stale scope in the live root silently. Data is safe, which is the whole point of the guard; the leak is just invisible.

quad341 (approved):
@julianknutsen — the follow-up I keep coming back to is (4), and specifically that you caught it on yourself. A memo living on a per-request factory is inert in production, reviews green, and does nothing. You moved it to the `Server` and then wrote a test asserting *one* parse across three per-request factory builds so it can't quietly drift back. I traced that wiring end to end today and it holds on both the `Session` and `SessionByID` paths.

The byte-wise fix is the other standout. Three sanitizers in three languages is exactly where a fix lands on two and the third drifts — and asserting that the reader resolves the adapter's *actual written path* for `wörker`, rather than just that both fold alike, is what makes it stay fixed. Same for the fault-injected archive test: proving the live transcript survives a failed copy is the assertion that matters.

Everything left is on us. #6145 and #6147 both landed, so this needs a rebase onto `main` — four zcode files conflict, purely because `main` moved under a correctly-stacked branch. I'll take that plus the missing `could not archive` warning on the occupied-slot branch as a maintainer commit, run the Mac tier, and merge behind the deploy gate. Your commits stay as authored; we won't squash the per-follow-up split.


---

PR #6162, opened 2026-09-08
test(residency): split the boundary guard's test monolith by rule (s7)

julianknutsen:
s7 of the #5644 re-split (PR A #5644 → PR B #6140 → **this**), rebased onto `main`.

**Pure test-file motion. No rule, no pattern, no exemption and no count changes.**
`scripts/residency-boundary-baseline.txt` and `scripts/residency-boundary-patterns.txt`
are untouched by this PR — that is the headline claim, and the evidence is below.

## Rebase note

The reviewed head was `5948112c39` on `residency/s7-guard-by-rule-v2`, one commit
on top of s6's *old* head `61effae3d4`. s6 was rebased twice after that and merged
to `main` as squash `ea66aa0fe6` (PR #6140). This branch is that commit re-applied
onto `origin/main`; the commit itself is byte-identical to the reviewed one.

**The cherry-pick was conflict-free, and it is worth saying why rather than only
that it was.** Both files s7 edits are byte-identical on `main` to what they were
on `61effae3d4`:

| File | on `61effae3d4` | on `origin/main` (`ea66aa0fe6`) |
| --- | --- | --- |
| `scripts/residency_boundary_test.go` | `8d3fb975fa4ec273…` | `8d3fb975fa4ec273…` |
| `scripts/check-residency-boundary.sh` | `81c209a75c1485c6…` | `81c209a75c1485c6…` |

s6's later rebases and PR A's guard work moved code the guard *reads*, not the
guard itself. s6's only residency-scripts change that reached `main` is a single
baseline row (`residencyBindingsFromRoutes` → `residencyBindingsFromRoutesWithProof`),
in a file this PR does not touch. So the split did not have to be redone over a
different monolith: `git diff 5948112c39 HEAD` over the seven touched paths is
empty, and `git diff origin/main HEAD` is the same 7-file / +1646 / −1555 shape the
reviewed commit had.

That is a statement about *inputs*, though, and equal inputs are not the same as
the guard still behaving identically over `main`'s current `cmd/gc` and
`internal/` trees. All four equivalence probes below were therefore re-run
end-to-end for this rebase — "before" is an untouched `origin/main` checkout,
"after" is this branch.

## Mechanism

`scripts/residency_boundary_test.go` held all four rules of the residency lookup
contract in 1724 lines, and each of the two AST rules walked and parsed
`cmd/gc`, `internal/api`, `internal/sling` and `internal/dispatch` **for
itself**, with the `node_modules`/`dist` prune spelled twice and the recursion
argument written out twice. Two copies of "which tree does this guard police" is
this guard's own bug class one level up: the halves-agree test exists precisely
because a scope that drifts makes the weaker side the real policy, and nothing
was pinning the two AST scanners' scopes against each other at all.

The file now holds only what the rules share — scan scope, allowlist, pattern
names, the fixture writer, and **one** walk-and-parse:

```go
type residencyTree struct {
	fset  *token.FileSet
	files []residencyParsedFile
}

func residencyParseTree(root string, dirs []string) (residencyTree, error)
```

`residencyParsedFile` gains a `data []byte` field (the source as read) so the
expression rule can derive its `residency:allow` marker lines from the same read
the parse used, instead of re-reading the file. Each rule moves to its own file
and becomes a pure count over a tree someone else walked:

```go
func countStoreListSignatures(tree residencyTree, allowlist map[string]bool) map[string]int
func countResidencyASTExpressions(tree residencyTree, allowlist map[string]bool, names residencyGuardedNames) map[string]int
```

The `scanX` entry points survive unchanged in signature — they are now
`residencyParseTree(...)` + `countX(...)` — so every caller and every control is
untouched.

The walk is deliberately **not** memoized across scans. The controls falsify each
rule by rewriting real files in a `t.TempDir()` tree and rescanning; a cache
keyed by root would hand a control the tree from before its own edit and report
that the guard bit when it had not been asked. What is shared is the
walk-and-prune IMPLEMENTATION, not a cached result. That is stated in the
function's doc so the next reader does not "optimise" it back.

## File map (before → after)

| Before | After | Contents |
| --- | --- | --- |
| `scripts/residency_boundary_test.go` (1724 lines) | `scripts/residency_boundary_test.go` (232) | scan dirs, allowlist, pattern names, prune list, `residencyTree` + the one `residencyParseTree`, the `isBeadsStore`/`isStorerefLeg` type predicates, `writeResidencyFixture` |
| ” | `scripts/residency_grep_rule_test.go` (288) | rules a-d: `residencyPattern`, `loadResidencyPatterns`, `scanResidencyGrepSites`, `TestResidencyBoundaryGrepRatchet`, `TestResidencyBoundaryGrepControls` |
| ” | `scripts/residency_signature_rule_test.go` (347) | `ast:returns-store-list`: `TestResidencyResolverBoundary(+Controls)`, `scanStoreListSignatures`, the new `countStoreListSignatures`, the yields-store-list predicates |
| ” | `scripts/residency_expression_rules_test.go` (684) | `ast:vocabulary-alias`, `ast:plan-leg-store-chain`, `ast:uncounted-call-spelling`: `residencyGuardedNames` + its derivation, the marker machinery, `scanResidencyASTExpressions`, the new `countResidencyASTExpressions`, `countResidencyDeclExpressions`, and all five control tests |
| ” | `scripts/residency_halves_agree_test.go` (176) | `TestResidencyBoundaryHalvesAgree`, the shell-array parser, `TestResidencyHalvesAgreementControls` |
| ” | `scripts/residency_ratchet_test.go` (86) | `readResidencyBaseline`, `ratchetViolations`, `assertRatchet`, `sortedKeys` |
| `scripts/check-residency-boundary.sh` | same | **comments only** — three references that named the old monolith now name the file each rule actually lives in |

Nothing outside `scripts/` is touched. No `cmd/gc`, no `internal/`.

## Byte-equivalence evidence (re-run on this rebase)

**1. Declaration-level: the motion is provably pure.** Every top-level
declaration in `main`'s monolith, with its doc comment attached, was SHA-256'd
and matched against the union of the six split files:

| | Count | Which |
| --- | --- | --- |
| IDENTICAL | 60 | |
| CHANGED | 4 | `residencyParsedFile` (+`data` field), `residencyParseTree` (now returns `residencyTree`), `scanStoreListSignatures` (82→17 lines, now parse+count), `scanResidencyASTExpressions` (60→20 lines, now parse+count) |
| ONLY IN SPLIT | 3 | `residencyTree`, `countStoreListSignatures`, `countResidencyASTExpressions` |
| **ONLY IN BASE** | **0** | nothing was lost |

The four changed declarations are exactly the shared-walk restructure.

Doc-comment hashing covers prose attached to a declaration but not prose between
declarations, so a count-aware diff of every comment line accounts for the rest.
Of 432 base comment lines, 25 line-occurrences (22 distinct) have no verbatim
match in the split's 486, and all 25 were read on both sides:

- **7 occurrences (4 distinct)** are section furniture the file names now carry —
  the `// ---…---` rule (×4) and the one-line titles `// The grep half.`,
  `// The AST half.`, `// The shared baseline and ratchet.`
- **9** are pure reflow: the sentence appears verbatim in the split, across a
  different line break.
- **9** are reworded in place — `below` → `across these files`; `This file is the
  rendering` → `This package is the rendering` (correct once a one-file rule
  becomes a package); `residencyParseTree walks and parses every non-test Go file
  under dirs` gaining `, once, for every AST rule in this contract`; and the
  two-halves paragraph extended to name the expression rules it had omitted.

Nothing was dropped. Both substantive prose blocks — `FUNCTION-VALUE ALIASING`
and `LINE-SPLIT FIELD ACCESS` — survive verbatim in
`residency_expression_rules_test.go`. Note that the base carried the RECURSES
paragraph **twice**, once on `residencyParseTree` and once on the second AST
scanner's own copy of the walk; deleting that duplicate is the point of the PR.

**2. The baseline does not move.** Regenerated by the three-emitter splice the
baseline file's own header documents — `--emit-baseline` minus its four PARTIAL
header lines, plus `RESIDENCY_BASELINE_EMIT=1 go test ./scripts -v -run
TestResidencyResolverBoundary`, plus the same for
`TestResidencyASTExpressionRatchet`, both filtered to exactly-four-field lines,
`LC_ALL=C sort`ed — then diffed against the committed file:

| | grep rows | `ast:returns-store-list` | expression rows | total | vs committed baseline |
| --- | --- | --- | --- | --- | --- |
| before (`origin/main`) | 84 | 21 | 5 | 110 | empty diff |
| after (this branch) | 84 | 21 | 5 | 110 | empty diff |

`cmp before after` → byte-identical.

**3. Verdicts on a deliberately violating fixture are byte-identical.** A fixture
tree was built exercising every rule at once — a new grep-vocabulary call, a raw
`[]beads.Store` result, a `map[string]beads.Store` result, an aliased result
type, a func-valued package `var`, a declared func value, a `[]storeref.Leg`
result, a value-position alias, a `.Leg.Store` chain, a line-split call, a
renamed-import call and a parenthesised receiver — each paired with negative
controls that must stay silent or must stay visible: a `residency:allow`-marked
escape, a `node_modules` copy, a `_test.go` copy and a prose-only mention (no
row), and a nested subpackage of a governed directory (a row, because the walk
recurses). Both halves ran against it before and after:

- Go half — all three scanners, the derived guarded-name sets and the ratchet's
  violation lines, sorted: **32 lines, `cmp`-identical**. That is 4 grep rows, 7
  signature rows, 5 expression rows, 1 derived-name-set line and **15 violation
  lines** across all four patterns (`a:rigBeadStores` 1, `b:cliSoleClassBinding`
  1, `c:plan-leg-store-access` 1, `ast:returns-store-list` 7,
  `ast:uncounted-call-spelling` 3, `ast:vocabulary-alias` 1,
  `ast:plan-leg-store-chain` 1). The four must-stay-silent controls produce no
  row in either run and the nested subpackage produces one in both, so this is a
  control-bearing probe rather than a quiet pass.
- Shell half — **`cmp`-identical output** (3 `RESIDENCY-BOUNDARY:` lines plus the
  explanation block) and **identical exit 1**.

**4. Both halves stay green on the real tree, before and after.**

| | before | after |
| --- | --- | --- |
| `./scripts/check-residency-boundary.sh` | exit 0, no output | exit 0, no output (`cmp`-identical) |
| `./scripts/check-residency-boundary.sh --self-test` | exit 0, no output | exit 0, no output (`cmp`-identical) |

Both are silent on success by design and print only on failure, so "no output" is
the pass signal — and probe 3 is what shows the silence is earned. The
self-test's 16 `expect` cases (the s4 launder pair, the same-file swap pair, the
marker case and the fail-closed cases) are untouched by this PR: the self-test
lives entirely in the shell half and the only shell edits are comments.

## Reconciliation with what landed while #5644 sat

The absorbed head `1806b45dfc` split a version of this file from before #5643
("make the residency rows falsify and close the boundary guard's evasions") and
before s6. Both are preserved here rather than reverted:

- **#5643 / s4 on main**: the parenthesised-receiver evasion in
  `residencyGrepCouldNotCount` (which classifies the receiver *without*
  unparenthesising it, deliberately, because it models what grep sees), the
  `routes.storeFor` qualified row and `storeFor` qualified-selector guard in
  `TestResidencyGuardedNameDerivation`, the parenthesised-receiver control case,
  `ast:uncounted-call-spelling` in the expression ratchet's baseline predicate,
  the `node_modules`/`dist` prune and the union prefilter. All carried; the
  self-test's launder cases are untouched.
- **s6 (#6140, now on main)**: the baseline header prose rewrite and the retired
  rows are PR B's, already merged — the baseline file has zero changes here.

## Gates

All run on this branch at `HEAD`.

| Command | Exit |
| --- | --- |
| `go build ./...` | 0, no output |
| `go vet ./...` | 0, no output |
| `gofmt -l scripts cmd internal` | 0, no output |
| `golangci-lint run ./scripts/...` | 0, `0 issues.` |
| `./scripts/check-residency-boundary.sh --self-test` | 0 |
| `go test ./scripts/... -count=1` | 0 |
| `make check-docs` | 0 |
| `make test-fast-parallel` | 0, `All fast jobs passed` (10 shards) |

Refs ga-j4p3y, ga-qdt5y.16.


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277315201

quad341 (dismissed):
@julianknutsen — thank you for the shared-walk extraction specifically. Splitting 1724 lines by rule was worth doing on its own, but the better catch is noticing that the two AST scanners each carried their own copy of "which tree does this guard police," with the prune list and the recursion argument spelled twice and nothing pinning them against each other. That's the halves-agree failure mode one level up, living inside the guard built to prevent it.

We verified rather than took your word for it: baseline regenerated through the three-emitter splice (84/21/5 = 110, empty diff), both shell halves green, package tests green, and a deliberately-violating file dropped into `cmd/gc` to confirm all four rules still fire from their new homes. They do.

Two things we appreciated in the write-up: documenting *why* the walk is deliberately un-memoized right on the function, so nobody "optimises" the controls into lying — and the honest accounting of the 25 unmatched comment lines rather than just asserting the prose survived.

Worth knowing, and not your problem: `Preflight / unit cover (noncmdgc)` was skipped on this run, so CI never actually executed `./scripts`. Your gate table is what made that a non-issue. Merging as-is.


---

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR splits `scripts/residency_boundary_test.go` — the ~1,600-line monolith that is the residency lookup contract's CI-visible enforcement — into one file per rule, and consolidates the two separate AST scans onto a single shared walk. That matters for a guard this repo actually depends on: the next person touching `ast:returns-store-list` now reads one file instead of paging through every other rule to find it, and collapsing the two AST walks removes a whole class of latent drift where two scanners could quietly start pruning different trees.

### Original PR Review
Decision: approve.

Specific gaps fixed:
- **The new expression-rules header overclaimed pattern-file derivation.** It promised all three expression rules watch the grep rule's vocabulary "DERIVED from the shared pattern file, never restated." Only two do — `residencyDeriveGuardedNames` skips every pattern row whose regex lacks `\(`, and `residencyIsLegStoreChain` takes no vocabulary at all and hardcodes the `.Leg.Store` shape. Left as written, a maintainer extending pattern family (c) would trust the promise, get no AST twin, and the drift would be silent because the expression ratchet deliberately must not fail closed on an empty census. The header now scopes the promise to the alias and uncounted-call rules and names the leg-chain rule as the Go-side restatement of `c:plan-leg-store-access`.
- **The new "four rules" / "one quarter of" count was unrecoverable under either reading.** If "four" meant the four `ast:` rules, the grep file's "one quarter of" was false — the grep rule is not an `ast:` rule, yet ratchets against the same baseline through the same helpers. If the grep census counted as a rule, the contract had five and the three "one quarter" pointers summed to three-quarters, while the expression file — which holds three rules — called itself one quarter either way. `residency_ratchet_test.go` contradicted itself within five lines, already carrying the correct count-free phrasing just below the wrong one. The count is dropped from all five new Go sites, restoring the pre-split style that could not go stale. The shell guard's own "four rules live only in the Go half" is pre-existing and accurate, so it was deliberately left alone rather than swept.
- **The shared file's inventory comment didn't describe its own contents.** `isBeadsStore` / `isStorerefLeg` / `isQualifiedType` live there but went unlisted; the inventory now names the signature rule's type helpers.

What the review validated rather than took on faith — each re-derived independently, since the PR's whole claim is that it changes no behavior:
- **Nothing was lost in the split.** Three independent declaration-level censuses agree that every rule, test, and helper survived; all 12 top-level `TestResidency*` tests and their subtests pass.
- **Allowlist semantics are unchanged** — the one behavior-relevant delta. The shared walk now deliberately returns allowlisted files, and both consumers re-apply the filter at count time instead of walk time, so net counting is identical to the pre-split monolith and alias declarations are still collected from exempt files.
- **The baseline does not move.** `residency-boundary-baseline.txt` and `residency-boundary-patterns.txt` are absent from the diff and the real-repo ratchets still pass green against them — the PR's own purity proof, and it holds.
- **The shell script change really is comment-only**, confirmed mechanically rather than by reading.

Review findings addressed:
- *Contract & interface fidelity (minor):* expression-rules header overclaimed derivation for all three rules — scoped to the two call-vocabulary rules in `36dcbb7117`.
- *Readability (nit):* arithmetically unrecoverable rule count across five new Go headers — dropped in `36dcbb7117`.
- *Architectural consistency (nit):* shared file's inventory comment omitted the signature rule's type helpers — named in `36dcbb7117`.

- `ga-rv85i`: finish the per-rule ownership story. `residency_boundary_test.go` holds zero `func Test` after the split — it is purely the shared harness — yet keeps the family's most prominent filename, and a few declarations sit on the wrong side of the line in both directions (`isBeadsStore`/`isStorerefLeg`/`isQualifiedType` are shared but consumed only by the signature rule; `residencyFileScope` is declared in the grep rule's file but consumed by the expression rules too). Both reviewers explicitly deferred this: relocating code would enlarge a diff whose entire proof is that it is behavior-preserving motion. Worth doing the next time this file family is touched, not on its own.
- `ga-6y2s6`: pre-existing, not introduced here — a hypothetical new `ast:` pattern kind added to the baseline would be selected by none of the three `readResidencyBaseline` predicates and would ratchet against nothing, silently. The predicates partition today's four `ast:` kinds exactly, so nothing is orphaned right now.

### Maintainer Changes
`fix(residency): correct the split's own header prose` (`36dcbb7117`) — 5 files changed, 15 insertions(+), 12 deletions(-). Comment-only, and mechanically confirmed as such: filtering the commit's changed lines for anything that is neither a `//` comment nor blank returns zero lines. It lands the three review findings above; no executable line differs from your branch.

### Final Review Status
Ready for the merge queue: final head `36dcbb71` has passing required GitHub checks — all six required contexts (`Check`, `Analyze (actions)`, `Analyze (go)`, `Analyze (javascript-typescript)`, `Analyze (python)`, `CI / required`) are green, 45 checks passing overall — and the merge-ready metadata is in place. I am marking this PR `status/merge-ready` so the merge queue can pick it up.

One red X you can ignore: `Evidence / critical-path suites` (the PR evidence watchdog) is a non-required advisory check that derives its verdict from the CI workflow's conclusion. It concluded at 22:06 reading an earlier `CI / required` failure; CI was re-run and reached success at 22:14, leaving the watchdog frozen against a run that no longer reflects this head. It carries no independent finding.

Local gates were also re-run against the branch rebased onto current `main`: `go test ./scripts/ -run 'TestResidency' -count=1` → 12/12 top-level pass, 0 fail, 0 skip; `go vet ./scripts/` and `gofmt -l scripts/` clean; `./scripts/check-residency-boundary.sh` and its `--self-test` both exit 0.

CI: https://github.com/gastownhall/gascity/actions/runs/34283645028

### Review Iterations
1 review pass. Three reviewers covered the diff independently and the pass approved it with no blocking findings; the three non-gating prose findings were then applied as a single comment-only maintainer fixup rather than sent back for another round.

---

PR #6163, opened 2026-09-08
fix(workflow): read workflow descendants at TierBoth through the live handle (#6129)

### LINKED ISSUE #6129: bug: workflow delete: members enumerated at TierIssues, so a split city's ephemeral steps survive the sweep rootless
## Summary

`gc workflow delete` federates its store sweep across the city's relocated class binding, but the queries that decide *which beads belong to the workflow* leave `ListQuery.TierMode` at its zero value. The zero value is `TierIssues` (`internal/beads/query.go:40-41`), and the class-binding leg has no bead-policy layer to rewrite it — so on a split city the binding's entire wisp tier drops out of the delete's member set, silently. The sweep then removes the root and leaves its ephemeral steps behind, rootless.

The fix is the one upstream already declares: enumerate members with `beads.DirectMembers` (`internal/beads/membership.go:199`), which reads through `HandlesFor(store).Live` and therefore states `TierBoth` on every store kind.

## Symptom

On a city whose coordination class is relocated (graph class in its own binding):

- `gc workflow delete <root>` previews/finalizes a member count that excludes the workflow's in-flight ephemeral steps.
- After the delete, those steps still carry a `gc.root_bead_id` that resolves to nothing — rootless, unworkable, and (per #5834's own framing) an escalation generator with no forward progress.
- Nothing errors. Every leg answered; the rows are simply not there.

## Reproduction

Code-level (the two reads inside one command disagree):

```bash
# the delete sweep's member query — TierMode unset => TierIssues
sed -n '2786,2791p' cmd/gc/cmd_convoy_dispatch.go
# the same command's stranding guard (#5802) — states TierBoth
gh pr diff 5802 | rg -n 'TierMode: beads.TierBoth'
```

CLI-level (split city, e.g. a city rebound so `gc.kind=workflow` rows live in the graph class binding):

```bash
gc bd create --metadata gc.kind=workflow ... -t "wf root"      # root lands in the graph binding
# an in-flight orchestration step (ephemeral => wisp tier) stamped with gc.root_bead_id=<root>
gc bd list --json --metadata-field gc.root_bead_id --metadata-value <root> --include-closed --limit 0 | jq length
gc workflow delete <root>          # reports fewer beads than the line above counted
gc bd list --json --metadata-field gc.root_bead_id --metadata-value <root> --include-closed --limit 0 | jq length   # > 0, rootless
```

Result: `expected the delete to enumerate every member, ephemeral included, and remove it with the root` → got: the wisp-tier members survive with a dangling `gc.root_bead_id`.

Test tier: the repo already has the fixture — `split_topology_conformance_test.go`'s `splitEnv` mints ephemeral graph rows via `mintEphemeralGraphBead` (used at `:2177`, `:2203`) and pins federated-tier reads. A `cmdWorkflowDelete` case on that fixture encodes this exactly.

## Root cause

The sweep **is** city-wide federated: `cmdWorkflowDelete` calls `federateSweepViews` (`cmd/gc/cmd_convoy_dispatch.go:1492`) before iterating views into `findWorkflowBeads` (`:1499`), and the delete-source arm does the same at `:2264` → `findWorkflowBeadsFromRoot` (`:2417`).

But the enumerating queries never state a tier:

- `cmd/gc/cmd_convoy_dispatch.go:2774-2779` — root discovery: `store.List(ListQuery{Metadata: {gc.workflow_id: …}, IncludeClosed: true})`
- `cmd/gc/cmd_convoy_dispatch.go:2787-2790` — member set: `store.List(ListQuery{Metadata: {gc.root_bead_id: rootID}, IncludeClosed: true})`
- `cmd/gc/cmd_convoy_dispatch.go:2805-2808` — same member query in `findWorkflowBeadsFromRoot`

`TierIssues` is the zero value (`internal/beads/query.go:40-41`), and upstream's own `FederatedReadTier` doc (`internal/beads/query.go:50-76`) states exactly why that is wrong on a federation: work stores are wrapped in a bead-policy layer that rewrites a `TierIssues` read to `TierBoth` before the backend, while "a relocated coordination-class store carries no such layer … so it answers at exactly the tier in the query", and "the class store's whole ephemeral tier drops out, silently". It records the measured live case: `bd ready --include-ephemeral` returned 17 claimable beads (3 wisps) over the relocated graph database while the federated reader over the same store returned exactly the other 14 — "ephemeral wisps are how orchestration steps run".

The class leg really is unwrapped: `convoyStoreViewsWithBinding` (`cmd/gc/cmd_convoy.go:567-613`) names the work-leg openers as policy-wrapped (`wrapStoreWithBeadPolicies`) and appends the binding as a **constructor-opened** store, unmodified.

So the delete path answers a different question than the rest of the codebase. Every other consumer of this member set goes through the declared fan-out contract `beads.DirectMembers` (`internal/beads/membership.go:184-199`, consumed by `internal/dispatch/{control,drain,fanout,ralph,retry,runtime}.go`), which reads `HandlesFor(store).Live` — and both Live readers force the union tier (`internal/beads/caching_store_handles.go:111-113` and `:169-171`). Dispatch's fan-out therefore considers steps that the delete sweep cannot see, and the delete removes the root under them.

Note the asymmetry inside the command itself: PR #5802's new stranding predicate *does* state the tier (`Metadata: {gc.root_bead_id: id}, TierMode: beads.TierBoth`), as does the pre-existing `storeHasOpenDescendants` (`cmd/gc/order_dispatch.go:2470-2476`, which already does `beads.HandlesFor(store).Live` + `TierMode: beads.TierBoth` — in-repo precedent for exactly this change). So `gc workflow delete` asks "are there live steps?" across both tiers and "which beads do I delete?" across one.

`split_topology_conformance_test.go:1793-1826` (I12) already pins that the member answer must come from `DirectMembers` and "does not depend on the topology, only on the front door" — the delete path is the fan-out that did not take it.

## Fix candidates

**A. Enumerate members with `beads.DirectMembers(store, rootID)`** — replace the member `ListQuery` in `findWorkflowBeads` and `findWorkflowBeadsFromRoot` with the declared contract; keep the root-discovery leg.
- Scope: one file, ~+17/−6 plus a split-topology case in `cmd_convoy_dispatch_test.go`.
- Trade-off: inherits the membership semantics exactly (which is the point — it is the same answer dispatch gives), and states the tier for free because both Live readers do. It does not by itself fix the root-discovery leg, which is a workflow-id query, not a membership query.

**B. State `TierMode: beads.FederatedReadTier` on all three queries** and leave the inline membership logic alone.
- Scope: three lines plus the same test.
- Trade-off: minimal and exactly what `FederatedReadTier` exists to be stated ("Every federated reader therefore states this tier on every leg"). It closes the root-discovery leg too, but it leaves the delete path re-implementing membership beside the contract that already declares it — which is how it drifted out of agreement with dispatch in the first place.

**Recommendation:** **A**, plus state `beads.FederatedReadTier` on the remaining root-discovery query — i.e. adopt the contract for the member set and stop any leg of this federation from answering at the default tier. A is what keeps the delete answer and the dispatch answer from diverging again; the stated tier on the leftover leg is the same two-word change #5802 already makes in the same file. Maintainer's call between A and B.

## Adjacent / out of scope

- **Not #5834 / #5802.** #5834's root cause is a *missing live-descendant check* in `deleteWorkflowBead` and its retention callers (`cmd/gc/order_dispatch.go`, `cmd/gc/wisp_gc.go`); #5802 adds that check. This is the sweep's *member enumeration* being tier-blind. It is also a hazard for #5802's fix: a guard that counts live steps across both tiers while the delete set is computed across one will refuse on a set the sweep cannot delete. Offered on #5802 as a follow-up rather than a competing edit to the same function.
- **Leg already landed:** the missing-binding half of this problem is upstream as `federateSweepViews` / `convoyStoreViewsWithBinding` (#5633). This issue is only about the tier the federated legs are asked at.
- **Other open work on this file** (neither takes the enumeration): #5802 (the stranding guard, above) and #5992 (*delete-source clears `workflow_id` on the copy the residency contract owns*, also `cmd/gc/cmd_convoy_dispatch.go`). Stated so a maintainer can sequence rather than discover the overlap.
- **Not #5832**, which is the order-tracking sweeps reading the *caching snapshot* in `internal/orders/store_reads.go` — same family of read-handle mistake, different path and different axis.
- The `KNOWN GAP` in `split_topology_conformance_test.go:1810-1826` (a work store answering `[]` for a relocated root instead of erroring) is a separate invariant (ga-iaj7k I0) and is not what this issue fixes.

## Environment

Filed from a downstream city running a branch that tracks upstream `main`. Every path/line citation above was re-read at current `main` (`15819b315`, 2026-09-07), not at the anchor the bug was first observed on, and the three tier-blind queries are still tier-blind there.

rjgeng:
## What

`findWorkflowBeads`'s descendants query left `ListQuery.TierMode` at its zero value (`TierIssues`), which filters out `Ephemeral` rows. On a split city whose coordination class is relocated to its own binding, this silently dropped every wisp-tier step from a workflow's member set: `gc workflow delete <root>` closed the root and reported success while its ephemeral steps stayed open, rootless, still carrying `gc.root_bead_id` pointing at a workflow that is gone.

Fixes #6129.

## Fix

Routes the descendants read through `beads.HandlesFor(store).Live`, which forces a live `TierBoth` read regardless of the query passed, and states `TierMode: beads.TierBoth` explicitly — the same idiom already used elsewhere in this file (`wispGCRootSelectors` in `wisp_gc.go`) and the exact pattern PR #5802 used to fix the identical shape in a sibling function. Two-line production change; the rest of the diff is the new regression test.

## Testing

`go build -tags gms_pure_go ./...` clean, `gofmt -l .` empty. New regression test `TestWorkflowDeleteSweepsTheEphemeralTierOfTheRelocatedTree` mints an ephemeral descendant through a class-binding fixture's `StorageCreateStore`, runs the delete, and asserts the descendant is closed. Independently confirmed this is a real regression guard, not a tautology: reverted just the two production lines, reran the test — it fails with the exact expected message ("... after the sweep, want closed ... TierMode: TierBoth ..."), then restored the fix and reconfirmed it passes.

Full local push-gate: all failures match this session's known-unrelated environmental clusters (dolt-fixture family, zcode interrupt tests) plus one shard failure that self-diagnosed in its own output as a leaked dolt test-server process contaminating an unrelated stderr-comparison test. None touch the changed file.

### DIFF
--- a/cmd/gc/cmd_convoy_dispatch.go
+++ b/cmd/gc/cmd_convoy_dispatch.go
@@ -2945,10 +2945,12 @@ func findWorkflowBeads(store beads.Store, workflowID string) ([]beads.Bead, erro
 	for _, root := range roots {
 		addRoot(root)
 	}
+	reader := beads.HandlesFor(store).Live
 	for _, rootID := range rootIDs {
-		all, err := store.List(beads.ListQuery{
+		all, err := reader.List(beads.ListQuery{
 			Metadata:      map[string]string{beadmeta.RootBeadIDMetadataKey: rootID},
 			IncludeClosed: true,
+			TierMode:      beads.TierBoth,
 		})
 		if err != nil {
 			return nil, fmt.Errorf("listing descendants of workflow %s: %w", rootID, err)
--- a/cmd/gc/cmd_convoy_dispatch_test.go
+++ b/cmd/gc/cmd_convoy_dispatch_test.go
@@ -7582,6 +7582,64 @@ func TestWorkflowDeleteSweepsTheRelocatedTreeAndTheRetainedCopy(t *testing.T) {
 	}
 }
 
+// TestWorkflowDeleteSweepsTheEphemeralTierOfTheRelocatedTree is the #6129
+// regression: findWorkflowBeads' descendants query left ListQuery.TierMode at
+// its zero value, TierIssues, which filters out Ephemeral rows. On a split
+// city the sweep still finds the root through the class binding, but every
+// wisp-tier step under it — the shape orchestration steps actually run in —
+// was silently dropped from the member set. `gc workflow delete` closed the
+// root and reported success while the ephemeral steps stayed open, rootless,
+// still carrying gc.root_bead_id pointing at a workflow that is gone.
+func TestWorkflowDeleteSweepsTheEphemeralTierOfTheRelocatedTree(t *testing.T) {
+	cityPath, _ := foreignProviderCity(t)
+	prevCityFlag := cityFlag
+	cityFlag = ""
+	t.Cleanup(func() { cityFlag = prevCityFlag })
+
+	binding := soleClassBindingStore(t, cityPath)
+	root, err := binding.Create(beads.Bead{
+		Title:  "the workflow root",
+		Type:   "task",
+		Status: "open",
+		Metadata: map[string]string{
+			beadmeta.KindMetadataKey: beadmeta.KindWorkflow,
+		},
+	})
+	if err != nil {
+		t.Fatalf("seeding the workflow root in the class binding: %v", err)
+	}
+
+	creator, ok := binding.(beads.StorageCreateStore)
+	if !ok {
+		t.Fatalf("the class binding (%T) implements no StorageCreateStore, so this fixture cannot reach its ephemeral tier", binding)
+	}
+	wisp, err := creator.CreateWithStorage(beads.Bead{
+		Title:    "the ephemeral step",
+		Type:     "task",
+		Status:   "open",
+		Metadata: map[string]string{beadmeta.RootBeadIDMetadataKey: root.ID},
+	}, beads.StorageEphemeral)
+	if err != nil {
+		t.Fatalf("minting the ephemeral descendant in the class binding: %v", err)
+	}
+	if !wisp.Ephemeral {
+		t.Fatalf("minted descendant %s has Ephemeral=false; this fixture is not exercising the wisp tier the bug is about", wisp.ID)
+	}
+
+	var stdout, stderr bytes.Buffer
+	if code := cmdWorkflowDelete(root.ID, true, false, &stdout, &stderr); code != 0 {
+		t.Fatalf("gc workflow delete exited %d: %s%s", code, stdout.String(), stderr.String())
+	}
+
+	closedWisp, err := binding.Get(wisp.ID)
+	if err != nil {
+		t.Fatalf("reading the ephemeral descendant back from the binding: %v", err)
+	}
+	if closedWisp.Status != "closed" {
+		t.Errorf("the ephemeral descendant %s is %q after the sweep, want closed: the descendants query must read TierMode: TierBoth so a relocated binding's wisp-tier steps are not silently dropped from the workflow's member set", wisp.ID, closedWisp.Status)
+	}
+}
+
 // TestWorkflowDeleteRefusesToSweepPastABindingThatStandsRefused pins the arm
 // that separates a sweep from a read.
 //


A3Ackerman:
Cross-link, since #6129 and #5802 touch the same file: #5802 adds the stranding guard about a thousand lines above this (the `workflowDelete*` helpers and the two delete entry points) and does not touch `findWorkflowBeads`, so the hunks don't meet. They compose — the guard already reads its descendants at `TierBoth`, and with this fix the member set it gates is enumerated the same way, which closes the asymmetry noted on #5802. Whichever lands second rebases; I'll do that on mine if this goes first.

PR #6164, opened 2026-09-08
fix(reconciler): back off and escalate repeated cold-start failures

remuscazacu:
## The problem

A session whose cold start keeps failing is retried on a fixed cadence, forever. In the incident that prompted this, one template produced **2,306 identical `session.cold_start_timeout` failures over 42 hours** — 53–56 an hour, roughly one every 65 seconds — with no backoff, no escalation and no ceiling.

`StartupHealthEpisode` already counts consecutive failures and quarantines at five for a flat five minutes, but a flat quarantine only divides the rate by a constant: across a 42-hour wedge that still permits on the order of 500 start attempts. And the episode's `AlertDisposition` is written but **read by nothing**, so no amount of consecutive failures ever reaches a human or an operator-facing consumer.

## What this changes

**1. A per-failure retry backoff on the episode.**

The first failure is not held back at all, so a single transient spawn flake still retries on the next tick. From the second, the hold doubles from 30s to a 30m cap.

`BackoffUntil` is a new field kept deliberately separate from `QuarantinedUntil` rather than folded into it: `doctor_startup_health.go` promotes any active quarantine to a hard `errorCheck`, so reusing that field would turn every transient double-flake into a `gc doctor` failure. The start gate consults `StartHoldUntil()` — the later of the two — while the session-row mirror stays gated on the quarantine alone, so a sub-threshold backoff never surfaces as a crash-loop quarantine.

**2. The cap is a rate ceiling, not a permanent stop — on purpose.**

Both observed incidents ended with a start that simply succeeded on the unchanged path (attempt 2,307 after 42 hours; attempt 6 after five minutes). A hard "after N failures, stop respawning" would have blocked a recovery that actually happened in both cases. At the cap a wedged template costs two start attempts an hour instead of 55, and the escalation below has already fired — which addresses the real cost (a silent, unbounded retry storm) without introducing a new failure mode (a permanently dead template that would have healed).

**3. `session.startup_health_alert`, emitted once per episode.**

Fires when the consecutive count crosses the quarantine threshold, carrying the count, the failure kind and how long starts are held back. `session.cold_start_timeout` remains the per-attempt datapoint; this is the escalation on top of it. `AlertDisposition` latches it to one alert per episode, and a successful start clears the episode so the next one alerts independently.

The event is recorded **before** the disposition is persisted: if the save fails, the next failure re-alerts rather than losing the escalation entirely. At-least-once is the right failure mode for an alert whose whole purpose is that 42 silent hours never happen again.

It is omitted from `KnownEventTypes` for the same reason as `ProviderHealthGateAlert` — that list drives only the typed SSE projection in `internal/api`, whose payload registration lands separately. Subscribers receive it through the custom-event envelope, and `gc events --type` filters on the raw string and is unaffected. `internal/api`'s event-payload/SSE coverage passes with the omission in place.

**4. `gc doctor` reports the effective hold.**

Past the threshold the quarantine is re-stamped at its fixed duration on every failure while the backoff keeps growing to its cap, so reporting `quarantined_until` alone would tell an operator starts resume in five minutes when the gate is actually refusing for thirty. The detail line now carries both, so the two remain distinguishable.

## Scope note for reviewers

The original analysis behind this change predates two commits already on `main`, and the diff is narrower as a result. `ga-o04bfr.1.1` added the persisted episode that counts failures across replacement pending-create beads, and `ga-vcjr9` made pool session runtime names a pure function of pool identity — which is what makes name-keyed counting work at all. Neither of those is re-litigated here. What was left genuinely missing upstream is only the two things above: **no backoff**, and **an escalation field nothing reads**.

## Testing

- `internal/session` and `internal/events` full suites pass.
- `internal/api` event-payload/SSE coverage passes, confirming the `KnownEventTypes` omission is safe.
- Full `cmd/gc` suite passes.
- Eight new tests: the backoff schedule (including that it never shortens the threshold quarantine), `StartHoldUntil` precedence, metadata round-trip and clearing, sub-threshold accrual on a pending-create failure, the start gate honouring the hold, one-alert-per-episode, and the doctor detail line reporting the effective hold.

`TestStartupHealthAlertEmittedOncePerEpisode` shows the retry interval visibly stretching across ticks 1, 1, 1, 1, 2, 5 — the backoff doing its job end to end rather than only in a unit assertion.

https://claude.ai/code/session_01RK9oCVD78MAXFv1GA956ay

### DIFF
--- a/cmd/gc/doctor_startup_health.go
+++ b/cmd/gc/doctor_startup_health.go
@@ -95,11 +95,18 @@ func formatStartupHealthEpisodeDetail(ep session.StartupHealthEpisode) string {
 	if alert == "" {
 		alert = "none"
 	}
-	return fmt.Sprintf("%s: %d consecutive %s failures (first=%s, last=%s, quarantined_until=%s, alert=%s)",
+	// held_until is the EFFECTIVE hold (StartHoldUntil: the later of the
+	// quarantine and the retry backoff), not QuarantinedUntil alone. Past the
+	// threshold the quarantine is re-stamped at its fixed duration on every
+	// failure while the backoff keeps growing to its cap, so reporting the
+	// quarantine by itself would understate how long starts are actually
+	// refused. Both are rendered so an operator can still tell the two apart.
+	return fmt.Sprintf("%s: %d consecutive %s failures (first=%s, last=%s, quarantined_until=%s, held_until=%s, alert=%s)",
 		ep.SessionName, ep.ConsecutiveCount, kind,
 		formatStartupHealthDetailTime(ep.FirstFailureAt),
 		formatStartupHealthDetailTime(ep.LastFailureAt),
 		formatStartupHealthDetailTime(ep.QuarantinedUntil),
+		formatStartupHealthDetailTime(ep.StartHoldUntil()),
 		alert)
 }
 
--- a/cmd/gc/doctor_startup_health_test.go
+++ b/cmd/gc/doctor_startup_health_test.go
@@ -358,3 +358,35 @@ func TestStartupHealthEpisodesCheckNeverLeaksLastDetail(t *testing.T) {
 		t.Errorf("FixHint leaks LastDetail: %q", result.FixHint)
 	}
 }
+
+// TestStartupHealthEpisodesCheckReportsTheEffectiveHoldNotJustTheQuarantine
+// pins that the rendered detail names the instant starts are ACTUALLY held
+// until. Past the threshold the quarantine is re-stamped at a fixed 5m on
+// every failure while the retry backoff keeps growing to its 30m cap, so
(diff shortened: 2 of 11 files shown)


remuscazacu:
Bump — p1, opened 2026-09-08, no human review yet, still `MERGEABLE`.

Restating the evidence because the shape is the point, not the count: one template failed its cold start **2306 times across 42 hours** on a fixed retry cadence with no ceiling, and the alert field that should have surfaced it is read by nothing. So the failure is simultaneously expensive and invisible — the two properties that make a backoff-and-escalate fix worth more than its size.

This is still an open bug on the deployment it was found on (tracked there as `sr-xf2xj`), and nothing about it has aged: the retry path is unchanged on `main`.

Happy to rebase onto current `main` or split the escalation half from the backoff half if a smaller diff would land faster.

PR #6165, opened 2026-09-08
perf(cli): reuse the loaded city config on gc ready's store opens (#6132)

### LINKED ISSUE #6132: perf(cli): gc ready re-loads the city config and re-spawns bd context once per store it opens
## Summary

`gc ready` loads the city config at `cmd_ready.go:260` and then opens every store it needs with the
*non-config* helpers, so each open re-loads the whole city config again and rebuilds its own
`bd context --json` preflight reader. The config-taking helpers already exist upstream
(`openStoreAtForCityWithConfig`, `main.go:1555`), the conformance test that polices this rule already
exists, and the ready path is simply the unconverted caller.

## Symptom

`gc ready` cost scales with the number of stores it opens, not with the work it answers. One
federated invocation performs one city-store open plus one open per standalone rig leg, and pays a
fresh `loadCityConfig` (builtin-cache readiness and pack expansion included) and a fresh
`bd context --json` subprocess per open.

On one 16-pool city this is on a controller-cadence path: 16 `scale_check` lines, five of them
`gc ready --include-ephemeral … --json --limit 0` demand probes (`city.toml:966, 977, 1002, 1008,
1196`), across the city store, two rig stores and the class binding. Each `gc` CLI start is measured
there at ~10 dolt connections and ~350 MB RSS (beads#4102, `city.toml:332-336`), and the accumulated
effect at the night seam was recorded as "27 demand-probe shells piled up (oldest 4 min)" — the
operator's mitigation was `probe_concurrency = 3` (`city.toml:259-263`), i.e. a config band-aid for
this code gap.

## Reproduction

Code-level, against `main` = `9d59e5c1041805a78c0f53f8923cc73ff549ba1f`:

```bash
git show origin/main:cmd/gc/cmd_ready.go | sed -n '260,273p'   # cfg loaded, then NOT passed to the opens
git show origin/main:cmd/gc/city_runtime.go | sed -n '3966,3982p'  # cfg in hand, calls openStoreAtForCity
git show origin/main:cmd/gc/main.go | sed -n '1597,1632p'      # cfg==nil -> loadCityConfig; new preflight checker per open
git show origin/main:cmd/gc/beads_preflight_checker.go | sed -n '13,30p'  # BDContext reader spawns bd context, uncached
git show origin/main:cmd/gc/native_dolt_env_cfg_reuse_test.go | rg -n '^func Test'   # 4 guards, none on the ready path
```

Runtime-level (Linux):

```bash
strace -f -e trace=execve gc --city <city> ready --json --limit 0 2>&1 | grep -c 'bd context'
# macOS: sudo fs_usage -f exec | grep 'bd context' during one invocation
```

Result: `expected one config load and one context preflight per invocation` → got: one of each per
store opened (city store + one per standalone rig leg).

## Root cause

`gc ready` already holds the config and does not thread it:

- `cmd/gc/cmd_ready.go:260` `cfg, err := loadCityConfig(cityPath, stderr)`, then
  `cmd/gc/cmd_ready.go:267` `openCityStoreAt(cityPath)` — no `cfg` — and
  `cmd/gc/cmd_ready.go:272` `readyRigLegStores(cfg, cityPath)`.
- `openCityStoreAt` (`cmd/gc/main.go:1471`) → `openCityStoreResultAt` (`:1479`) →
  `openStoreResultAtForCity(cityPath, cityPath)` → nil config → `cmd/gc/main.go:1603`
  `cfg, _ = loadCityConfig(runtimeCityPath, io.Discard)`.
- `readyRigLegStores` (`cmd/gc/ready_federation.go:229`) → `openStandaloneRigStores(cfg, cityPath)`
  (`cmd/gc/city_runtime.go:3966`) → **`cmd/gc/city_runtime.go:3981`
  `openStoreAtForCity(rig.Path, cityPath)`** — that caller has `cfg` as a parameter and still calls
  the loading variant. `openStoreAtForCity` (`cmd/gc/main.go:1546`) forwards
  `openStoreAtForCityWithAuthority(…, false)` → nil cfg → the same `loadCityConfig` at `:1603`.
- The same open path rebuilds the preflight reader every time: `cmd/gc/main.go:1630`
  `PreflightChecker: newBeadsPreflightChecker(runtimeCityPath, provider)` →
  `cmd/gc/beads_preflight_checker.go:17` `BDContext: preflightBDContextReader(cityPath)` →
  `:23` runs `bd context --json` through `bdCommandRunnerForCity(cityPath)`. There is no cache:
  `cachedPreflightBDContextReader` appears in 0 files. Note the reader is already
  scope-parameterised (`func(scope string) …`), so per-scope isolation of a shared cached reader is
  part of the same atom, not an extra.
- The rule is this repo's own: `cmd/gc/native_dolt_env_cfg_reuse_test.go` exists precisely to
  enforce "reuse the config you were given" and carries exactly four guards (`:18`, `:107`, `:166`,
  `:215` — `TestOpenNativeStoreReusesTheLoadedCityConfig`, `TestBdBeadExistsProbeReusesTheLoadedCityConfig`,
  `TestWorkRecordCloseGateReusesTheLoadedCityConfig`, `TestBdCloseGateReusesTheWriteGuardsStoreRead`).
  **None covers the ready path.** The precedent is #4723 (merged 2026-07-29), which converted
  `cmd_bd.go` / `work_record_gate.go` / `store_target_exec.go` and added that test file; the ready
  path was not in that pass.
- The parallel half is also unfurnished: `cmd/gc/ready_federation.go` has zero `sync.` / `go func` /
  `errgroup` hits at `main`, so the leg fan-out is a serial per-leg loop.

## Fix candidates

**A. Thread the loaded config and one cached context preflight through the ready path** — open the
city store and every rig leg via the existing `openStoreAtForCityWithConfig` (`cmd/gc/main.go:1555`)
/ `openStoreResultAtForCityWithConfig` (`:1597`), build one cached BD-context reader per invocation
and pass it down (the BD-context-taking variant does **not** exist yet — 0 hits for
`WithConfigAndBDContext` — so that part is additive), and add the two ready-path guards to the
existing conformance file (e.g. `TestReadyRigStoreOpenReusesTheLoadedCityConfig`,
`TestCmdReadySharesOneBDContextPreflightAcrossCityAndRigOpens`).
- Scope: `cmd/gc/cmd_ready.go`, `cmd/gc/main.go`, `cmd/gc/ready_federation.go`,
  `cmd/gc/beads_preflight_checker.go`, +2 tests.
- Trade-off: the store-open helpers gain a parameter, so every caller must pass it — which is the
  point of the guard: the unconverted call site becomes a compile-time or test-time failure instead
  of a silent per-store reload.

**B. Memoize at the lowest layer** — cache the loaded city config and the `bd context` read
process-wide keyed by city path, leaving call sites unchanged.
- Scope: ~1 file.
- Trade-off: no signature churn, but a process-wide cache is exactly what the existing
  "reuse the config you were given" rule avoids, and it hides the unconverted call site instead of
  converting it.

**Recommendation: A**, because it is the same change #4723 already made for the other CLI paths and
the config-taking helper is sitting there unused by this caller — that is what makes it cheap to
accept. Maintainer's call.

## Adjacent / out of scope

- **#6013** (open PR, `perf(cli): stop eagerly reloading city config on every control-ready tick`)
  is the same defect shape at a different door: `dispatch_control_ready.go` only. Coordinate, do not
  merge the two issues.
- **#5540** (open PR) pins the native-Dolt pool ceiling to 1 conn for one-shot store opens — that is
  the *connection* half of the symptom and does not touch the redundant config loads or the repeated
  `bd context` spawn.
- **#5903** (open PR) touches `cmd/gc/cmd_ready.go`, `cmd_ready_test.go` and
  `schemas/ready/result.schema.json`; a ready-path change should rebase on that file's final shape
  rather than hand-merge.
- **#5539** (connection churn from one-shot CLI opens), **#5533** (serial `bd` discovery in the hook
  budget), **#5712** (per-identity duplicate scans in the generated work query) and **#5519** (`git
  status` per pack source on every command) are neighbours in the per-invocation-floor family, each a
  different call site. This issue names only the ready path's config load and context preflight.
- `cmd/gc/ready_federation.go`'s serial leg loop is named above as still-true, but the fix proposed
  here is the reuse conversion; parallelism is a separate call.

rjgeng:
## What

`cmdReady` and the shared `openStandaloneRigStores` helper (used by both the `gc ready` CLI path and the controller/supervisor's `buildStandaloneRigStores`) opened their stores through non-config-threading helpers, which always pass `nil` for config and force a full city-config reload — builtin-cache readiness, pack expansion, everything — per store, even though the caller already holds the config in scope. On a 16-pool city this sits on a controller-cadence path; the only prior mitigation was a config band-aid (`probe_concurrency = 3`).

Fixes #6132.

## Fix

Two one-line call-site changes, matching an idiom already established elsewhere in the codebase:
- `cmd/gc/cmd_ready.go`: `cmdReady`'s city-store open now calls `openStoreAtForCityWithConfig(cityPath, cityPath, cfg)` instead of `openCityStoreAt(cityPath)`.
- `cmd/gc/city_runtime.go`: `openStandaloneRigStores`'s per-rig open now calls `openStoreAtForCityWithConfig(rig.Path, cityPath, cfg)` instead of `openStoreAtForCity(rig.Path, cityPath)`. Since this helper is shared by both `readyRigLegStores` (the `gc ready` CLI path) and `buildStandaloneRigStores` (the controller path), the fix covers both for free with no signature changes.

Adds two new AST-based conformance tests (`TestReadyCityStoreReusesTheLoadedCityConfig`, `TestStandaloneRigStoresReuseTheLoadedCityConfig`) mirroring the existing four guards in `native_dolt_env_cfg_reuse_test.go`, so a future regression at either call site fails the build rather than silently reintroducing the reload.

## Testing

`go build -tags gms_pure_go ./...` clean, `gofmt -l .` empty on all three touched files. Full 6-test config-reuse conformance suite passes (4 pre-existing + 2 new).

Full local push-gate: all failures match this session's known-unrelated environmental clusters (dolt-fixture family, zcode interrupt tests) plus one shard failure already diagnosed in a sibling PR tonight as a leaked dolt test-server process contaminating an unrelated stderr-comparison test. None touch the changed files.

### DIFF
--- a/cmd/gc/city_runtime.go
+++ b/cmd/gc/city_runtime.go
@@ -3978,7 +3978,7 @@ func openStandaloneRigStores(cfg *config.City, cityPath string) (map[string]bead
 		if strings.TrimSpace(rig.Path) == "" {
 			continue
 		}
-		store, err := openStoreAtForCity(rig.Path, cityPath)
+		store, err := openStoreAtForCityWithConfig(rig.Path, cityPath, cfg)
 		if err != nil {
 			failures = append(failures, rigStoreOpenFailure{rig: rig.Name, err: err})
 			continue
--- a/cmd/gc/cmd_ready.go
+++ b/cmd/gc/cmd_ready.go
@@ -264,7 +264,7 @@ func cmdReady(opts readyOpts, stdout, stderr io.Writer) int {
 		fmt.Fprintf(stderr, "gc ready: %v\n", err) //nolint:errcheck // best-effort stderr
 		return 1
 	}
-	cityStore, err := openCityStoreAt(cityPath)
+	cityStore, err := openStoreAtForCityWithConfig(cityPath, cityPath, cfg)
 	if err != nil {
 		fmt.Fprintf(stderr, "gc ready: %v\n", err) //nolint:errcheck // best-effort stderr
 		return 1
--- a/cmd/gc/native_dolt_env_cfg_reuse_test.go
+++ b/cmd/gc/native_dolt_env_cfg_reuse_test.go
@@ -265,3 +265,100 @@ func TestBdCloseGateReusesTheWriteGuardsStoreRead(t *testing.T) {
 		t.Fatalf("found %d %s call(s) in doBd, want exactly 1", calls, callee)
 	}
 }
+
+// cmdReady loads the city config once, then opens the city store. Opening it
+// through openCityStoreAt (which always passes a nil config down) re-ran the
+// whole city config load — builtin-cache readiness and pack expansion
+// included — for a store open cmdReady only needed to resolve which stores
+// federate ready work. On a controller-cadence caller of `gc ready`, that
+// reload was paid once per invocation on top of the load cmdReady had
+// already done.
+func TestReadyCityStoreReusesTheLoadedCityConfig(t *testing.T) {
+	fset := token.NewFileSet()
+	file, err := parser.ParseFile(fset, "cmd_ready.go", nil, 0)
+	if err != nil {
+		t.Fatalf("parsing cmd_ready.go: %v", err)
+	}
+
+	fn := findFuncDecl(file, "cmdReady")
+	if fn == nil {
+		t.Fatal("cmdReady not found in cmd_ready.go")
+	}
+
+	var opens int
+	ast.Inspect(fn, func(n ast.Node) bool {
+		call, ok := n.(*ast.CallExpr)
+		if !ok {
+			return true
+		}
+		ident, ok := call.Fun.(*ast.Ident)
+		if !ok {
+			return true
+		}
+		switch ident.Name {
+		case "openCityStoreAt":
+			t.Fatal("cmdReady opens the city store without a config, which re-loads the city config it was handed")
+		case "openStoreAtForCityWithConfig":
+			opens++
+			if len(call.Args) != 3 {
+				t.Fatalf("openStoreAtForCityWithConfig: got %d args, want 3", len(call.Args))
+			}
+			arg, ok := call.Args[2].(*ast.Ident)
+			if !ok || arg.Name != "cfg" {
+				t.Fatalf("cmdReady passes %s as its config; want the already-loaded \"cfg\"", exprText(call.Args[2]))
+			}
+		}
+		return true
+	})
+	if opens != 1 {
+		t.Fatalf("found %d config-carrying store open(s) in cmdReady, want exactly 1", opens)
+	}
+}
+
+// openStandaloneRigStores opens one store per bound rig, and is shared by
+// both `gc ready` (readyRigLegStores) and the controller's supervisor loop
+// (buildStandaloneRigStores) — both callers already hand it a loaded city
+// config as a parameter. Opening each rig's store through openStoreAtForCity
+// discarded that config and re-loaded it once per rig, on top of the load
+// the caller had already done, on every store the loop opened.
+func TestStandaloneRigStoresReuseTheLoadedCityConfig(t *testing.T) {
+	fset := token.NewFileSet()
+	file, err := parser.ParseFile(fset, "city_runtime.go", nil, 0)
+	if err != nil {
+		t.Fatalf("parsing city_runtime.go: %v", err)
+	}
+
+	fn := findFuncDecl(file, "openStandaloneRigStores")
+	if fn == nil {
+		t.Fatal("openStandaloneRigStores not found in city_runtime.go")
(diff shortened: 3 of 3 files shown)


rjgeng:
Closing as superseded by merged #6536. That PR implements the same config-reuse optimization across the one-shot paths while preserving reload-per-open freshness for long-lived controller and supervisor paths; replaying this older unconditional helper change would weaken that distinction.

PR #3994, opened 2026-07-06
Runtime session fixes: heartbeat command, on-demand restart, overlay .gc mirror skip

sarendipitee:
Three related runtime/session lifecycle fixes.

## 1. fix(overlay): skip runtime mirrors during staging

**Problem:** `CopyDirForProvider`/`CopyDirForProviders` copied any `.gc` directory found in the overlay source into the staged destination, leaking runtime state into session workdirs where it can shadow the live city state.

**Fix:** add a `skipRuntimeMirror` guard to both copy paths so `.gc` (and anything under it) is never staged; universal and per-provider files are unaffected. Test added (`TestCopyDirForProviders_SkipsRuntimeMirrors`).

## 2. feat(runtime): `gc runtime heartbeat`

**Problem:** agents running slow operations that produce no terminal output (e.g. long Dolt queries >15m on busy rigs) trip the idle-timeout watchdog, causing false-alarm kills and restart loops.

**Fix:** new `gc runtime heartbeat [--duration d]` subcommand (default 45m, 1m minimum) that sets `held_until` on the current session's bead. The reconciler already respects `held_until` for idle-timeout and max-session-age (#2418); this exposes it as an agent-facing API without triggering a full user-hold suspend. Tests added; CLI reference regenerated with `go run ./cmd/genschema`.

## 3. fix(runtime): restart named on-demand sessions

**Problem:** `gc runtime request-restart` returned early for on-demand configured named sessions ("controller cannot restart on-demand named sessions") and silently cleared the request.

**Root cause:** the reconciler's `restart_requested` path actually handles these sessions correctly — it stops the running session, records the restart handoff, and deliberately keeps the durable reset marker out of the in-memory bead so on-demand sessions are not force-woken without demand. The command-side skip predates that and blocks the working mechanism.

**Fix:** remove the skip; tests updated to assert the restart request proceeds for named on-demand sessions.

## Testing

- `go build ./...` clean, `go vet ./cmd/gc/... ./internal/overlay/... ./internal/worker/...` clean
- `go test ./internal/overlay/` ok
- `go test ./cmd/gc/ -run 'Heartbeat|RuntimeRequestRestart|Drain'` ok
- `go test ./cmd/gc/ -run 'Gendoc|GenDoc|Doc'` ok (cli.md staleness gate passes)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6168, opened 2026-09-08
fix(reconciler): keep a live pool seat that holds assigned work out of the timer drains

A3Ackerman:
## Summary

A live pool seat that still holds assigned work is no longer drained by the reconciler's own timers (no-wake-reason, idle-sleep, config-suppressed). Explicit intents (`gc session suspend`, idle-stop-pending, max-age) still drain it. One gate in the `!shouldWake && alive` arm of `reconcileSessionBeads`, mirroring the heartbeat-hold keep-alive from #3994.

## Why

An orchestrator seat holding an assigned `open` step whose dependency is running on another seat reads not-ready in bd's ready projection, so `workBeadHasAwakeDemand` gives the holder no wake demand and the reconciler begins a `no-wake-reason` drain. The seat's `gc hook --claim --drain-ack` acks it at once and the run is abandoned mid-molecule: the step strands, the dead-assignee lane releases it minutes later, and the work re-runs. Measured on one review plane on 2026-09-04: eight `session.drain_acked_with_assigned_work` on the same orchestrator in 85 minutes, a paid review re-run per drain. #5731 describes the stranding that follows; #5735 recovers from it. This PR stops the drain from being sent to a holder in the first place.

Gating the drain rather than the wake keeps demand accounting untouched: no extra seat is minted for a blocked step, and an asleep holder is still not woken for it. A probe error fails closed (retain, do not drain), like every other assigned-work gate in the reconciler.

## Testing

- `TestReconcileSessionBeads_PoolSeatWithoutWorkDrainsForNoWakeReason` pins the control: a desired pool seat with nothing assigned still drains for no-wake-reason.
- `TestReconcileSessionBeads_PoolSeatHoldingAssignedOpenWorkIsNotDrained` and `..._PoolSeatHoldingInProgressWorkIsNotDrained` are red without the gate, green with it.
- `TestReconcileSessionBeads_PoolSeatHoldingWorkStillDrainsOnExplicitIntent` pins that a suspend still wins over held work.
- `go test ./cmd/gc -run 'TestReconcile|TestCompute|TestSession|TestPool|TestDrain|TestNamed|TestOnDemand|TestAwake|TestWake' -count=1` passes; `go vet ./cmd/gc` is clean.

## Scope

Pool-managed seats only; a named seat's held work stays the retired-session lane's concern. The zero-grace drain-ack (#5473) and the recovery of already-stranded steps (#5735) are deliberately not touched. No generated artifacts.

Refs #5731, #5473. Complements #5735.

https://claude.ai/code/session_01N5cd5opXyA2u7JM8gKQzb2

### DIFF
--- a/cmd/gc/session_reconciler.go
+++ b/cmd/gc/session_reconciler.go
@@ -3899,6 +3899,36 @@ func reconcileSessionBeadsTracedWithNamedDemand(
 				cancelSessionDrainInfo(info, sp, dt)
 				continue
 			}
+			// Keep-alive by held work: a live pool seat that still holds assigned
+			// work — open or in_progress — is not drained by the reconciler's own
+			// timers (no-wake-reason, idle-sleep, config-suppressed). Its step may
+			// carry no wake demand right now: an open step whose dependency is
+			// running on another seat reads not-ready, so the awake set sees no
+			// reason to keep the holder. The holder is mid-molecule all the same.
+			// A drain sent here is acked at once by `gc hook --claim --drain-ack`
+			// and the seat abandons its run; the step strands and the work re-runs
+			// (#5731, #5473). Explicit intents (suspend, idle-stop-pending,
+			// max-age) still drain, exactly as they win over a heartbeat hold. A
+			// probe error fails closed — retain, do not drain — like every other
+			// assigned-work gate. Pool seats only: a named seat's held work is the
+			// retired-session lane's concern.
+			if intent == "" && isPoolManagedSessionInfo(info) {
+				holdsWork, holdsErr := sessionHasOpenAssignedWorkForReachableStore(cityPath, cfg, store, rigStores, info)
+				if holdsErr != nil {
+					fmt.Fprintf(stderr, "session reconciler: checking assigned work before draining %s: %v (retaining)\n", name, holdsErr) //nolint:errcheck
+					holdsWork = true
+				}
+				if holdsWork {
+					cancelSessionDrainInfo(info, sp, dt)
+					if trace != nil {
+						trace.RecordDecision(TraceSiteReconcilerDrainDecision, TraceReasonAssignedWork, TraceOutcomeNoChange, target.tp.TemplateName, name, traceRecordPayload{
+							"sleep_intent": intent,
+							"kept_by":      "assigned_work",
+						})
+					}
+					continue
+				}
+			}
 			var reason string
 			switch {
 			case intent == "idle-stop-pending":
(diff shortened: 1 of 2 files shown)


A3Ackerman:
Whole matrix green on 422291d56 (CI run 34280424796, 66 checks): Check, CodeQL Analyze (actions, go, javascript-typescript, python), Preflight (static checks, generated artifacts, acceptance A), Contract / bd CLI, cmd/gc process shards, Integration (packages-core, packages-runtime-tmux, bdstore, rest-smoke), worker core, Mac regression gate, Review Formulas, Dashboard SPA, and the Evidence / critical-path suites watchdog. The one red before this was the watchdog's fixed 25-minute deadline on the first run (#5850) while that run itself concluded success; the retrigger is an empty commit and the change is unchanged from b3bd73102.

ckumar1 (approved):
Approving at head 422291d56 after an independent verification pass (affiliated with the author's org — cross-account review under the towns' real-reads-only pattern; full read, not an endorsement).

What I verified, executed locally:

- All four new tests green at the head (`TestReconcileSessionBeads_PoolSeat*`), and a **red-check**: with the gate condition disabled in a scratch worktree, both holder tests (`…HoldingAssignedOpenWorkIsNotDrained`, `…HoldingInProgressWorkIsNotDrained`) fail immediately — the regressions are non-vacuous and the control test pins that an empty holder still drains for no-wake-reason.
- Gate coverage matches the claim exactly: it sits in the `!shouldWake && alive` arm under `intent == ""`, and the reason switch below resolves that condition to precisely the three reconciler-own timer reasons (no-wake-reason, idle-sleep via decision, config-suppressed); every explicit intent bypasses the gate, which the user-hold test pins.
- `sessionHasOpenAssignedWorkForReachableStore` is the existing probe, not a fork of it: open+in_progress across the primary and rig store legs, failing closed on an unreadable leg (`assignedWorkScanComplete`), and the gate's own error handling retains-and-logs, consistent with every other assigned-work gate in the file.
- The cancel of an in-flight (even acked) drain for a work-holder rides the deliberate pre-existing semantic named in #6178's second commit — "the cancels that override an ack by design (assigned work, …)" — so this adds no new ack-provenance surface.
- No double probing: the neighboring assigned-work probe in the same function runs only on the `!alive` (dead-runtime) arm, mutually exclusive with this gate's live arm. The residual cost is one probe per live no-wake pool seat per tick, same class as the dead-seat probe — worth watching, not blocking.

The measured incident evidence (eight `drain_acked_with_assigned_work` on one orchestrator in 85 minutes, one paid re-run each) and the choice to gate the drain rather than the wake — keeping demand accounting untouched — both hold up.

Verification basis: local test execution and red-check at 422291d56 on darwin; scope claims checked against the file at that head.

PR #6169, opened 2026-09-08
fix(beads): compare bd and the linked beads library by commit when both are pseudo-versions

remuscazacu:
## Summary

`checkVersionCompat` abstains in exactly the configuration that is `gc`'s normal state, and that abstention is what lets a `gc`/`bd` beads skew silently destroy a city's stores.

The linked beads library is pinned as a pseudo-version. `comparableReleaseVersion` refuses it, `unconfirmableReason` fires, and the check returns **Pass** with `"bd/beads schema compatible; linked library version unconfirmed"` — the one verdict that guarantees the native store gets opened.

This is not hypothetical. On 2026-09-08 a host had:

| Binary | Linked beads | Max migration |
|---|---|---|
| `gc` (installed) | `bf97b73749ac` | 0059 |
| `bd` (on `PATH`) | `98b4b79e6a4d` | 0054 |

Opening a store runs `MigrateUp` unconditionally, with no prompt and no down-migration, so three stores went v54 → v59 on first contact. Migration `0055_move_leases_to_table` moves lease state out of the columns a v54 `bd` still selects, so from that moment **every agent `bd` call failed**:

```
schema version mismatch: database is at v59, binary knows up to v54 (5 migrations ahead)
```

The preflight check ran on each of those stores, and passed.

### The fix

Two pseudo-versions **can** be compared. A release version and a pseudo-version name points on different axes — that part of the existing reasoning is right — but two pseudo-versions both name a commit in the same repository, and for this pair "the same commit" is precisely the invariant that matters: `gc` links beads as a library, `bd` *is* beads, and both open the same stores.

So when the library pin is a pseudo-version and `ctx.BDVersion` is one too, compare the revisions (`module.PseudoVersionRev`) and `Fail` on mismatch.

**Failing here is preventive, not merely diagnostic.** `factory.go` routes a failed preflight to `openBdFallback` and never calls `openNativeStore`, so the migration that would break `bd` does not run at all. The check stops being a post-mortem and starts being the thing that prevents the incident.

### Scope

Deliberately narrow — every other configuration behaves exactly as before:

- A **replaced** module is excluded. Its version names a commit in a *different* repository, so the comparison would be meaningless. `pseudoRev()` returns `false` whenever `Replaced` is set, even if the version string looks like a pseudo-version.
- A release library against a pseudo-version `bd` already fails the pre-existing semver compare. Untouched.
- A genuinely unconfirmable pin (bd on a release version, library on a pseudo-version) still passes, with the same message as before.

One note for reviewers, because it looks like a safety net and is not: `ctx.SchemaVersion` is the **bd contract shape** — `bd context` reports `1` — not the store's migration number. The existing `SchemaVersion <= 0` guard says nothing about migration skew, so it could not have caught this.

## Testing

- [x] `make check` — `fmt-check`, `lint`, `vet`, and the topology/residency gates all pass.
  - `make test` reported one failure, `TestPackCommandExitReturnsThroughRun/eager`, which is environmental and unrelated: the assertion is `helper stderr … want empty` and the stderr it got was the dolt leak guard reaping a stale `sql-server` left behind by an earlier aborted run on the same machine. It passes on re-run: `go test -count=1 -run TestPackCommandExitReturnsThroughRun ./cmd/gc/` → `ok`. The change touches only `internal/beads/contract`.
- [ ] `make check-docs` — no docs, navigation or links changed.
- [ ] `make test-integration` — no runtime, controller or workflow behavior changed; the change is a pure verdict change inside one preflight check.

New tests in `preflight_checker_test.go`:

- `TestCheckVersionCompatPseudoVersionLockstep` — 7 cases covering matching commits (Pass), mismatched commits (Fail), a replaced module (stays unconfirmable), a bd release version against a pseudo-version library (stays unconfirmable), and the pre-existing release-vs-pseudo direction (still Fail, unchanged).
- `TestPseudoVersionRev` — 8 cases: canonical and `v`-less pseudo-versions, release versions, `dev`, empty, and malformed input.

## Checklist

- [x] Linked an issue, or explained why one is not needed — no upstream issue; this was found by an operator during a live incident on a self-hosted city and the repro is the version table above. Happy to file one if you'd prefer the paper trail.
- [ ] Updated docs for user-facing changes — the user-visible surface is one check message, quoted in full in the test.
- [x] Called out breaking changes or migration notes — **behavior change worth naming**: a host running mismatched `gc`/`bd` pseudo-version builds that previously opened the native store will now fall back to `bd`. That is the intended outcome, and the failure message names both commits, but on a host that has *already* migrated its stores past its `bd` the fallback will itself fail — correctly, and visibly, instead of silently migrating more stores.

### DIFF
--- a/internal/beads/contract/preflight_checker.go
+++ b/internal/beads/contract/preflight_checker.go
@@ -262,14 +262,50 @@ func (c PreflightChecker) checkVersionCompat(ctx PreflightBDContext, err error) 
 		return NewPreflightCheckResult(PreflightCheckVersionCompat, PreflightCheckWarn, "bd/beads version compatibility could not be confirmed", details)
 	}
 	if reason := library.unconfirmableReason(); reason != "" {
-		// The compare below only means something when both sides name the same
-		// released beads artifact. When the linked library does not — a source
-		// build, a replaced module, or a pseudo-version naming an untagged
-		// commit — the two strings can never be equal, and answering "mismatch"
-		// reports a verdict the check never had the evidence to reach. The
-		// schema version is validated above and is the real compatibility
-		// signal, so an unconfirmable library version must not take the native
-		// store offline; only a *confirmed* mismatch (below) should.
+		// A pseudo-version cannot be compared to a bd RELEASE version, but it CAN
+		// be compared to bd's own pseudo-version: both name a commit in the same
+		// repository, and equality of those commits is the strongest evidence of
+		// compatibility this check can obtain — stronger than the semver compare
+		// below, which only ever compares release tags.
+		//
+		// This is not an edge case. gc pins beads at a pseudo-version whenever it
+		// tracks beads ahead of beads' last tag, which is gc's normal state, and
+		// bd built from source reports one too. So the branch that declared the
+		// question unanswerable was covering the configuration in which the two
+		// binaries most often drift — and drift here is not cosmetic: opening the
+		// native store runs MigrateUp, so a gc built from a newer beads commit
+		// migrates every store it touches past what an older bd can read,
+		// irreversibly and on first contact. Observed 2026-09-08: a gc carrying
+		// beads bf97b73749ac (max migration 0059) took a fleet's bd — built from
+		// 98b4b79e6a4d (max migration 0054) — offline for every read, with this
+		// check passing throughout.
+		//
+		// Failing is therefore the safe verdict AND the preventive one: the
+		// factory routes a failed preflight to openBdFallback, so the migration
+		// that would break bd never runs at all.
+		if libRev, ok := library.pseudoRev(); ok {
+			if bdRev, bdOK := pseudoVersionRev(ctx.BDVersion); bdOK {
+				if bdRev == libRev {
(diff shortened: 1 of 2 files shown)


remuscazacu:
Bumping this with a fresh, concrete consequence found today — it is no longer hypothetical.

I have been planning a v59 → v67 beads repin of an 18-database deployment, where the whole safety story rests on gc and bd carrying the *same* beads commit. The preflight this PR fixes is exactly the check that should verify that, and it cannot: both binaries are pinned at pseudo-versions (`1.1.1-0.20260805093327-bf97b73749ac`), which is the case the preflight abstains on. So the lockstep has to be verified by hand:

```
go version -m ~/.local/bin/gc | grep steveyegge/beads
go version -m ~/.local/bin/bd | grep '^\s*mod'
```

and the repin runbook now carries "do not trust the preflight here" as a written step. That is the failure mode the PR describes — a skewed host opens the native store, migrates it past its own bd, and every agent `bd` call fails afterwards — with the one guard that would catch it silently declining to look.

p1, opened 2026-09-08, no human review yet. Happy to rebase or split it if that helps it move.

PR #389, opened 2026-04-06
fix(prompt): mayor dispatch uses gc sling instead of raw bd commands

johnzook:
## Summary

- Replace `bd update --set-metadata gc.routed_to=...` and `bd update --label=pool:...` in the mayor prompt with the canonical `gc sling <rig>/polecat <bead>` command
- The quick-reference table had the label-based pattern, which is especially dangerous: labels alone don't trigger the controller's scale_check, so polecats never spin up

## Context

Discovered in production: a mayor session dispatched 3 beads using `bd update --label=pool:signal-loom/polecat` (following the quick-ref table). The controller's scale_check queries `--metadata-field gc.routed_to=...`, found nothing, and never started polecats. The beads sat idle until force-slung.

Separate issue: `gc sling` idempotency check detects labels as "already routed" even when the metadata it actually writes (`gc.routed_to`) is missing — so running `gc sling` after a label-only dispatch skips the bead without `--force`.

## Test plan

- [ ] Verify `gc prime` for mayor shows `gc sling` in dispatch instructions
- [ ] Verify quick-reference table no longer references `--label=pool:...`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #6170, opened 2026-09-08
fix(kimi): support Kimi Code liveness and native journals

julianknutsen:
Kimi Code 0.36.1 changes both its process name and its persisted session format. With the existing `process_names = ["kimi"]` launcher configuration, a live `sudo → bwrap → kimi-code` seat was reported dead and repeatedly recycled. The shared tmux matcher now recognizes the `kimi-code` executable for existing Kimi configurations.

The Kimi reader now discovers native `sessions/wd_*/session_*/agents/main/wire.jsonl` journals alongside legacy MD5/context journals, preserving exact session IDs, refusing ambiguous lookups, and honoring `KIMI_CODE_HOME` and configured search roots. It projects native messages, assistant text/thinking, tool calls/results, compaction boundaries, timestamps, and turn activity into the existing worker transcript surface. Successful cross-layout discovery no longer emits missing-layout warnings.

Validation:
- Regression tests failed before the fixes and passed afterward; all four affected packages pass.
- Lint and `go vet ./...` pass; pre-commit hooks passed.
- The mandatory push gate passed `make test-fast-parallel`, including all six CLI shards, the core suite, and the updated tmux test inventory.
- Both cached and uncached liveness checks passed against an isolated real Kimi Code 0.36.1 process under the existing bubblewrap launcher, which remained alive for over 15 minutes.
- All 73 saved native journals parsed (1,151 normalized entries) and all 73 exact session-ID lookups succeeded.

Live inference was attempted but the existing direct Ollama credential returns HTTP 401 on both inference endpoints. Fleet routing remains unchanged. Swap-back deployment, the custom native `observe_paths` entry, and a canary using working credentials are tracked in `ga-jnwhh`; this code fix is `ga-asyj4`.


julianknutsen:

Track progress: https://factory.gascity.com/city/maintainer-city/runs/gcg-5436965277317565

quad341:
Holding this one short of auto-merge on a single CodeQL finding — flagging it with the trace rather than just the rule name, because I think only one of the two call sites actually needs anything.

**Alert #389 — `go/path-injection`, high, `internal/sessionlog/kimi_code_reader.go:60`** (`os.Open(path)` in `readKimiCodeWire`). It's net-new to this PR; the other two open alerts on the merge ref are `govulncheck` notes on `go.mod`, not in-diff and not new.

The path reaches that `os.Open` through:

```
kimiTranscriptPath(workRoot, sessionID)
  -> filepath.Join(workRoot, sessionID, "context.jsonl")   // kimi_code_discovery.go:35-39
```

`filepath.Join` cleans its result but doesn't contain it — `Join(root, "../../x")` resolves outside `root` — so the guard has to be on the segment, not the join.

Of the two callers, I think only one is exposed:

- **`kimi_reader.go:258`** — `kimiTranscriptPath(workRoot, entry.Name())`. `entry.Name()` comes from `os.ReadDir` and can't contain a separator. Safe as written.
- **`kimi_reader.go:213`** — `kimiTranscriptPath(filepath.Join(root, workHash), sessionID)`, inside `findKimiSessionFileByIDInVisited`, where `sessionID` is a caller-supplied parameter and isn't validated anywhere in the package.

So the exposure depends on what can reach `findKimiSessionFileByIDIn` with an arbitrary `sessionID`. If that's only ever an id already resolved from the session store, this is a false positive and worth dismissing as such. If any of it can come from a request or a raw CLI argument, it's a real traversal — the reader would `Stat` and `Open` outside the root, and although it only surfaces lines that parse as the expected JSON, that's still an arbitrary-path read.

Cheapest containment is on the segment, which also just makes the alert go away:

```go
if sessionID != filepath.Base(sessionID) || sessionID == "." || sessionID == ".." {
    return ""
}
```

Happy to be told it's unreachable and dismiss the alert with that rationale — you'd know the callers better than I do. Just didn't want to auto-merge past an in-diff high without someone actually looking at it.

The Kimi liveness/journal work itself reads well; this is the only thing standing between it and a merge from my side.

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @julianknutsen! This PR makes Kimi Code a
first-class provider in the session-log layer — it teaches tmux process
liveness to recognize `kimi-code` processes, and it adds discovery and reading
for the native journal layout (`wd_<slug>_<hash>/<session>/agents/main/wire.jsonl`)
alongside the legacy `~/.kimi/sessions/<session>/context.jsonl` one, including
activity inference for native entries. Before this change a native Kimi Code
session was simply undiscoverable: no transcript, no activity, nothing on the
API or dashboard surfaces. That is a real hole closed, and the two-layout
handling (rather than a migration that breaks existing journals) is the right
call.

PR branch.

### Original PR Review
Decision: `request_changes` on the first pass, resolved to `approve`.

Specific gaps fixed:
- **Test isolation was incomplete by construction.** The PR newly has
  `DefaultKimiSearchPaths` read `KIMI_CODE_HOME`, and `mergeKimiSearchPaths`
  folds those defaults into *every* kimi discovery call — but the kimi
  discovery tests isolated only `HOME`. On any host that exports
  `KIMI_CODE_HOME`, the ambient root supplies a second candidate bucket and
  `TestKimiCodeDiscovery` fails deterministically; CI stayed green only because
  CI hosts do not set the variable, which makes it latent rather than absent.
  Reproduced both directions during review. The fixup pins
  `KIMI_CODE_HOME` to a `t.TempDir()` beside the existing `HOME` pin in the
  discovery tests that isolate `HOME`, and deliberately leaves
  `TestKimiCodeRootsAndWorkspaceIdentity` alone because it exercises the
  override on purpose.
- **The missing-workdir diagnostic named the wrong key, and its new filter was
  inert.** `logKimiMissingWorkDir` computed both the legacy MD5 key and the
  native `wd_*` key, then passed `legacyKey` unconditionally to the warning —
  so on a native-only store the operator was told an MD5 hash was absent while
  the key that actually missed was never surfaced. Separately, the new
  `workKey` parameter on `hasKimiSessionRootEntries` was provably inert: its
  two calls at the single call site were OR'd, and since legacy keys are never
  `wd_`-prefixed and native keys always are, the disjunction reduced to "any
  dir entry exists" — exactly the old unfiltered check. The fixup tests each
  key separately and logs the one that matched, which fixes the wrong-key
  message and makes the parameter load-bearing in the same edit.
- **Kimi tail-meta root merging sat one layer above every sibling family.**
  The adapter merged kimi defaults itself and then called the generic
  `ExtractTailMetaFromSearchPaths`, while the adjacent codex branch passes
  search paths through unmerged and lets `sessionlog` merge internally — two
  contracts for one concern in neighbouring `if` arms. The fixup adds
  `ExtractKimiTailMetaFromSearchPaths` so the adapter branch collapses to a
  plain family dispatch matching codex. Verified the accepted root set is
  unchanged by the move.
- **Exported doc comments still described a single-layout world.** Several
  `kimi*` doc comments predated the two-layout reality, and
  `DefaultKimiSearchPaths`' own newly-written comment enumerated two roots
  while the function returns three — omitting `~/.kimi-code/sessions`, the
  operationally dominant one, which would lead a reader to conclude native
  discovery requires `KIMI_CODE_HOME`. All corrected.

Review findings addressed: all four above were fixed by the maintainer fixup
commit `0b9dffa`. The test-isolation fix was confirmed RED→GREEN, and the
diagnostic fix was checked against a mutation to confirm the new assertion is
not vacuous.

- `ga-pof3l` — *Change Impact / Blast Radius, minor*: native kimi
  `usage.record` entries are never folded into tail meta, so a newly
  discoverable native kimi row renders activity next to blank model and
  context-percent columns. Not a regression — legacy kimi journals were
  already blank on those fields, since their lines carry a top-level `role`
  and never the Claude-shaped `type: "assistant"` that the extractor keys on —
  so this is an unfilled surface, not one this PR broke. Not addressed here.
- `ga-2l87n` — *Architectural Consistency, minor*: native kimi sessions emit
  sibling `agents/<name>/wire.jsonl` journals, but `kimiTranscriptPath` pins
  `agents/main/wire.jsonl` and the subagent discovery path is Claude-shaped
  (`<session>/subagents`, `agent-*.jsonl`), so subagent rows stay empty for
  kimi. Also not a regression — legacy kimi had no subagent layout at all —
  but nothing yet records that main-agent-only is deliberate. Not addressed
  here.
- `ga-hoskc` — *Contract & Interface Fidelity, minor*: `readKimiFile`
  dispatches on `filepath.Base(path) == "wire.jsonl"`, and each converter
  returns nil for the other layout's records while every line stays valid
  JSON. A journal read under the wrong name therefore yields an empty
  `Session` with a nil error rather than a loud failure. In-repo producers
  always mint canonical names, so this only bites out-of-tree callers that
  copy or archive journals. Not addressed here.

### Maintainer Changes
`fixup! fix(kimi): recognize Kimi Code processes and native session journals` —
6 files changed, +114/-14 (`kimi_code_discovery.go`, `kimi_code_test.go`,
`kimi_reader.go`, `kimi_reader_test.go`, `reader.go`,
`worker/sessionlog_adapter.go`). Your original commit is preserved unchanged.

### Final Review Status
Ready for the merge queue: final head `0b9dffa` has passing required GitHub

CI: https://github.com/gastownhall/gascity/actions/runs/34319788169

### Review Iterations
2 review passes. Pass 1 returned `request_changes` on the test-isolation gap;
the maintainer fixup landed that fix plus the diagnostic, layering, and
doc-comment corrections, and pass 2 returned `approve` with no gating findings.
The base branch moved during review, so the change set was re-validated against
the current base: build, vet, and the full test set for every touched package
(`internal/sessionlog`, `internal/worker`, `internal/runtime/tmux`, `scripts`)
pass there, and none of the 13 files in this PR were touched by the intervening
base commits.

---

PR #6174, opened 2026-09-08
fix(proctable): never signal tmux infrastructure from KillByPID

### LINKED ISSUE #6172: proctable: KillByPID has no infrastructure check, so the kill family that took a city down twice still signals a tmux server's process group when handed its PID
### Gas City version

main at `042e965f0`; `internal/runtime/proctable/kill_unix.go` is unchanged since #5392 merged.

### Symptom

A whole city lost at once: every tmux session on the socket gone, the supervisor restarting the full population a minute later. Measured twice on one city (macOS laptop, one socket per city, the default topology): 2026-08-03 17:04:44 and 2026-08-13 01:16:48. Each time the city's tmux server (PPID 1, its own process-group leader) died, a fresh server appeared 78 ms later founded by the next `new-session`, and every other session died with the old server.

### Root cause

The kill that removed the server was `proctable.KillByPID`, reached from `tmux.Provider.TerminateRuntime` (`internal/runtime/tmux/adapter.go:387`) through `killExistingOrphans` on the start path. `KillByPID` (`internal/runtime/proctable/kill_unix.go:22`) does not know what it is killing: it captures start-time identity and signals, and `signalPIDWith` (`kill_unix.go:101`) signals the process **group** first (`kill(-pid)`), which for a tmux server is the server itself.

#5837 (fixed by #5392) closed the feeder that handed the server in: the scanner now classifies tmux infrastructure by command and never reports it as an agent root. Both callers of the kill family (`tmux/adapter.go:387`, `subprocess/subprocess.go:358`) are fed by that scan, so no live caller hands a server in on today's main; this is defence in depth, not a reproducer against current code. The kill itself is unchanged and unconditional, and the classifier is exact-match by design (a `tmux-*` wrapper that is really an agent must stay reapable), so a command name it does not know, or a PID recycled between scan and kill, reaches the same syscall with no check.

The field history says the last line matters: the first guard for this class on our lineage was installed on the other kill family, `tmux.terminateProcesses`, which enumerates pane descendants and cannot reach the server by construction (measured process groups on the 08-03 incident), and the city was lost again through `KillByPID`.

### Reproduction

Unit-level on main: stub `KillByPID`'s liveness to alive and hand it a PID whose command is tmux infrastructure; it signals `-pid`. In the field: a mayor `gc handoff` destroys the mayor's tmux session before the orphan scan runs; on a pre-#5392 scanner the server was returned as that session's root and the start path killed it.

### Environment

macOS 15 (arm64) laptop city and a Linux box, tmux 3.5a, `gc` built from main.

### Proposed fix

1. **Refuse in `KillByPID` itself** (proposed): before signalling, classify the target with the scanner's own `isInfrastructureCommand` — `/proc/<pid>/comm` on Linux (the existing `isInfrastructureProcess`), `argv[0]` via `pidutil.Cmdline` on Darwin (`kern.procargs2`, no subprocess) — and return success with a one-line stderr note. Success rather than an error: a tmux server cannot race a replacement for an agent's work, so there is no orphan to refuse a Start over, and an error here would propagate through `killExistingOrphans` to every gated start.
2. Re-classify in each `TerminateRuntime` before calling `KillByPID`: narrower, and it leaves the next caller of the kill family to remember on its own.

A PR for (1) follows.

A3Ackerman:
## Summary

`KillByPID` refuses to signal tmux infrastructure. It classifies its target with the scanner's own predicate from #5392 — `/proc/<pid>/comm` on Linux (the existing `isInfrastructureProcess`), `argv[0]` via `pidutil.Cmdline` on Darwin (`kern.procargs2`, no subprocess) — and returns success with a one-line stderr note instead of sending `SIGTERM` to the process group. Everything else about the kill is unchanged.

## Why

A tmux server is its own process-group leader, and `signalPIDWith` signals the group first, so `KillByPID` on a server's PID is the whole city: one socket, one server, every session. That is how one city was lost twice (2026-08-03 and 2026-08-13; the chain is on #6172). #5837 closed the feeder — the scanner no longer reports a server as an agent root — and both callers of this kill family are fed by that scan, so no live caller hands a server in on today's main. This is best-effort defence in depth for the last line: the classifier is exact-match by design (a `tmux-*` wrapper that is really an agent must stay reapable), so a command name it does not know reaches the same syscall with no check, and a PID recycled after the scan is caught only if it is recycled before this check — the guard narrows that window rather than closing it, and the SIGTERM-to-SIGKILL grace stays uncovered, as before. On our lineage the first guard for this class was correct, tested, and installed on the other kill family; the city was lost again through this one.

Refusing is success rather than an error: a tmux server cannot race a replacement for an agent's work, so there is no orphan to refuse a Start over, and an error here would propagate through `killExistingOrphans` to every gated start. The predicate fails open — an unreadable comm or argv reads as not infrastructure — because failing closed would silently disable orphan reaping on any procfs or `procargs2` hiccup and let a genuine survivor race its replacement. The note goes through `log.Printf`, as `workspacesvc`'s own refusal to signal an unsafe orphan-reap target does.

## Testing

- `TestKillByPIDRefusesTmuxInfrastructureTarget` and `TestKillByPIDStillSignalsNonInfrastructureTarget` drive `KillByPID` itself with the detector, the kill syscall, liveness, and the start-time reader injected — hermetic, nothing read from the host, no process spawned — so the refusal is proven wired in, not merely implemented. The first fails immediately without the guard (the liveness stub says alive, and the signal stub fails the test on the first delivery). The SIGKILL escalation stays covered by the existing `killByPID` core tests.
- `TestIsInfrastructureKillTargetReadsComm` (Linux, fake procfs via `SetScanRootForTesting`) and `TestIsInfrastructureKillTargetClassifiesArgv` (Darwin, injected argv reader: bare `tmux`, a path, the retitled server, a `tmux-*` wrapper that must stay reapable, empty and unreadable argv) cover the per-platform readers; one explicitly labelled Darwin smoke test reads the live host.
- `go test ./internal/runtime/proctable/ ./internal/runtime/tmux/ ./internal/runtime/subprocess/ ./internal/pidutil/`, `go vet` on darwin and `GOOS=linux`, and `make lint-changed` are clean. No new subprocess sites in source or tests.

## Scope

Only `KillByPID`. Sibling kill paths, checked and left alone with the reason: the tmux package's `terminateProcesses` / `KillPaneProcesses` / `KillPaneProcessesExcluding` family enumerates a pane's descendants and same-group reparented members, and the server is the pane's ancestor in its own group, unreachable by that enumeration (measured process groups on the 08-03 incident); `computeExcludingKillSet` is set math over those PIDs; `workspacesvc`'s orphan reap signals groups behind its own `unsafeSignalTarget` refusal over a service-identity-scoped scan a tmux server cannot enter; `processgroup.Terminate` only receives a group its caller spawned; the dolt and supervisor kill sites hold dolt and supervisor PIDs. `tmux.KillServer` is intentionally unguarded — killing the server is its job. One distinct mechanism is out of scope by nature: at city start, with no server running, the first session's own `tmux -L <socket> new-session` becomes the server (fork and setsid, PPID 1, same argv), so ending that one session ends the server through lineage rather than through any kill path; no kill-path guard covers session-equals-server. That server keeps a tmux `argv[0]`, so the classifier here does refuse it if it ever reaches `KillByPID`. The scan-side classification from #5392 is reused, not changed. No generated artifacts.

Fixes #6172. Refs #5837; precedent #5392. Reviewed before opening with the contributing pack's review rubric and an independent Codex read; both sets of findings are applied.

https://claude.ai/code/session_01LBcMPssKbTUyfM7UtUYaoF

### DIFF
--- a/internal/runtime/proctable/kill_target_darwin.go
+++ b/internal/runtime/proctable/kill_target_darwin.go
@@ -1,0 +1,28 @@
+//go:build darwin
+
+package proctable
+
+import "github.com/gastownhall/gascity/internal/pidutil"
+
+// killTargetCmdline is indirected so the classifier's argv handling can be
+// covered without a live tmux process.
+var killTargetCmdline = pidutil.Cmdline
+
+// isInfrastructureKillTarget reports whether pid names tmux infrastructure (a
+// tmux server or client) by argv[0] — the token the Darwin scanner classifies,
+// since ps's first command token is argv[0], possibly a path, and tmux does not
+// retitle itself on Darwin. argv comes from kern.procargs2, so the
+// classification adds no subprocess to the kill path.
+//
+// The predicate fails OPEN: an unreadable argv (the process is gone, or
+// belongs to another user) reads as not infrastructure and the kill proceeds
+// exactly as before the guard. Failing closed would silently disable orphan
+// reaping on any procargs2 hiccup and let a genuine survivor race its
+// replacement, which is the outage the kill exists to prevent.
+func isInfrastructureKillTarget(pid int) bool {
+	argv, err := killTargetCmdline(pid)
+	if err != nil || len(argv) == 0 {
+		return false
+	}
+	return isInfrastructureCommand(argv[0])
+}
--- a/internal/runtime/proctable/kill_target_darwin_test.go
+++ b/internal/runtime/proctable/kill_target_darwin_test.go
@@ -1,0 +1,45 @@
+//go:build darwin
+
+package proctable
+
+import (
+	"errors"
+	"os"
+	"testing"
+)
+
+func TestIsInfrastructureKillTargetClassifiesArgv(t *testing.T) {
+	prev := killTargetCmdline
+	t.Cleanup(func() { killTargetCmdline = prev })
+
+	cases := []struct {
+		name string
+		argv []string
+		err  error
+		want bool
+	}{
+		{name: "bare tmux", argv: []string{"tmux"}, want: true},
+		{name: "tmux by path with the founding new-session argv", argv: []string{"/opt/homebrew/bin/tmux", "-u", "-L", "gc", "new-session", "-d"}, want: true},
+		{name: "retitled server", argv: []string{"tmux: server"}, want: true},
+		{name: "tmux-named wrapper stays reapable", argv: []string{"tmux-wrapper"}, want: false},
+		{name: "agent", argv: []string{"/usr/local/bin/claude", "--resume"}, want: false},
+		{name: "empty argv fails open", argv: nil, want: false},
+		{name: "unreadable fails open", err: errors.New("procargs2: operation not permitted"), want: false},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			killTargetCmdline = func(int) ([]string, error) { return tc.argv, tc.err }
+			if got := isInfrastructureKillTarget(4242); got != tc.want {
+				t.Fatalf("isInfrastructureKillTarget(%v, %v) = %v, want %v", tc.argv, tc.err, got, tc.want)
+			}
+		})
+	}
+}
+
+// TestIsInfrastructureKillTargetHostSmoke reads the live host once: the test
+// binary itself must not classify as tmux infrastructure.
+func TestIsInfrastructureKillTargetHostSmoke(t *testing.T) {
+	if isInfrastructureKillTarget(os.Getpid()) {
+		t.Fatal("the test binary classified as tmux infrastructure")
+	}
+}
--- a/internal/runtime/proctable/kill_target_linux.go
+++ b/internal/runtime/proctable/kill_target_linux.go
@@ -1,0 +1,21 @@
+//go:build linux
+
+package proctable
+
+// isInfrastructureKillTarget reports whether pid names tmux infrastructure (a
+// tmux server or client) by its /proc comm — the same reading the Linux
+// scanner uses to keep such processes out of the agent-root set, so the scan
+// and the kill path cannot drift on what counts as infrastructure.
+//
+// It reads through scanRoot and deliberately bypasses liveScanGuard: that
+// guard exists because a live /proc scan under go test can reap real agents
+// (gastownhall/gascity#2839), and a single comm read reaps nothing.
+//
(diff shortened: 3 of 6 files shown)


A3Ackerman:
Whole matrix green on 6b68a635f (CI run 34282828007, 66 checks): Check, CodeQL Analyze (actions, go, javascript-typescript, python), Preflight (static checks, generated artifacts, acceptance A), Contract / bd CLI, cmd/gc process shards, Integration (packages-core, packages-runtime-tmux, bdstore, rest-smoke), worker core, Mac regression gate, Review Formulas, Dashboard SPA, and the Evidence / critical-path suites watchdog.

ckumar1 (approved):
Approving at head 6b68a635f after an independent verification pass (affiliated with the author's org — cross-account review under the towns' real-reads-only pattern; this was a full read, not an endorsement).

What I verified, executed locally on darwin (the platform of both incidents this guards against):

- `go test ./internal/runtime/proctable/ ./internal/pidutil/` green at the head; `go vet` clean.
- **Red-check on the wired-in claim**: with the guard condition disabled in a scratch worktree, `TestKillByPIDRefusesTmuxInfrastructureTarget` fails immediately with the "guard is not wired in" message — the regression is non-vacuous and proves the refusal lives in `KillByPID` itself, which is the precise lesson of the #5837 first-fix miss (correct guard, wrong kill family).
- Classifier semantics: `isInfrastructureCommand` matches on `filepath.Base(TrimSpace(...))`, so the path-prefixed argv[0] case (`/opt/homebrew/bin/tmux ... new-session`) and the retitled forms are covered, while exact-match keeps `tmux-*` wrappers reapable — consistent with the scan-side predicate this reuses rather than forks.

Design points that convinced me: refusal-as-success is right (a tmux server can't race an agent replacement, and an error would propagate through `killExistingOrphans` into every gated start); fail-open on unreadable argv/comm is the correct tradeoff and the comment says why; and the scope section enumerates every sibling kill path with the reason it's untouched, including the honest note that session-equals-server at city start is out of reach of any kill-path guard.

Verification basis: local test execution and red-check at 6b68a635f as described; Linux-side classifier covered by the fake-procfs test (not executed on a Linux host here).

A3Ackerman:
Parked as a draft to stay under our six-open-PR cap (p2; yielding the slot to p1 work). Will re-ready when a slot opens; no changes to the branch.

